Issue #7515, opened 2018-12-07
Generalize interface on hp-adaptive methods


marcfehling:
@davydden, @bangerth and I are currently working on how to simplify the user interface for hp-adaptive methods, so that it is more easily accessible.

We agreed on a workflow that users have to follow in the following order:
- Setting cell refinement flags as usual (e.g. using KellyErrorEstimator).
- Decide whether h- or p-refinement should be imposed.
- Perform refinement and coarsening.

---

This requires the implementation of the following missing features:

- [ ] Set flags regarding p-adaptivity.
  - [ ] Supply `p_refine` and `p_coarsen` flags.
This involves adding flags to `internal::hp::DoFLevel` and corresponding `get` and `set` functions for `DoFCellAccessor`.
  - [ ] Rename `cell_(refine|coarsen)_flag` to `h_(refine|coarsen)_flag`, and deprecate the former. h-adaptivity will be imposed by default.
- [ ] Perform p-refinement and p-coarsening on `hp::DoFHandler`.
  - [ ] Provide `hp::FECollection` with hierarchy.
Provide two `std::function` objects as members that determine which `active_fe_index` should follow the current one on refinement / coarsening. For example with a function `hp::FECollection::set_hierarchy(refinement_criterion, coarsening_criterion)`. Default functions should return the succeeding / preceeding index respectively, or return the same index if we'd get out of bounds. This may be done elegantly with a C-style conditional assignment.
  - [ ] Provide decision criteria, whether h- or p-refinement shall be performed.
We thought about three different options in a namespace `hpRefinement` or `hp::GridRefinement`:
    - [ ] `pure_p_adaptivity()`
Converts all h_flags to p_flags.
    - [ ] `decide_hp_on_regularity()`
Compare polynomial degree of current FE with estimated Sobolev index [Houston2005].
    - [ ] `decide_hp_on_criteria()`
Compare criteria vector to a different one, e.g. predicted errors [Melenk2001].
    - [ ] Also consider p coarsening !!!
- [ ] Provide `CellDataTransfer` classes.
  - [ ] Provide a class `CellDataTransfer` similar to `SolutionTransfer` that is able to transfer cell attached data across refinement. Necessary to transfer predicted errors.
  - [ ] Provide a class `p::d::CellDataTransfer` similar to `p::d::SolutionTransfer`.
[...]

Issue #3511, opened 2016-11-09
Make hp::DoFHandler work with parallel::distributed::Triangulation


bangerth:
#1366 already asks for extending `hp::DoFHandler` for shared triangulations. But we also want it to work for completely parallel triangulations, of course.

I've created a project for this here: https://github.com/dealii/dealii/projects/5 . However, I find the "project" format awkward to use because it does not allow referencing projects from pull requests, and the GUI is also awkward because one gets into trouble with project that want more space than is visible on the screen in any given column. Consequently, here is my TODO list for getting there (help of course appreciated!):

 
- [x]  Go through the internal::DoFHandler::Policy namespace and understand the structure there. This includes which operations are available in the PolicyBase class and how work is dispatched between the sequential, shared, and parallel cases.

- [x]  As appropriate clean up code.

- [x]  Create a similar structure for hp::DoFHandler where the major operations are dispatched to a separate set of classes. At first, simply create the base class and a class for the sequential case.

- [x]  Move existing code into the derived policy class for the sequential case.

- [x]  Create parallel, derived classes for the shared and distributed case. It may be easiest to deal with the shared case first (#1366, but will also require #1367). 

- [x]  Following the structure for `parallel::distributed::Triangulation` implemented for `::DoFHandler`, also implement something similar for the corresponding `hp::DoFHandler` policy. The issue here is going to be that one has to transfer variable amounts of data between cells because cells no longer have the same number of DoFs. The easiest way for the moment will be to just allocate buffers that have the size of the maximal amount of data, and just pad with zeros as appropriate. We have, for the moment, no other choice anyway because p4est only allows us to attach a fixed sized buffer to cells.

- [x] Support `parallel::distributed::SolutionTransfer`

- [ ] Provide a weighting mechanism for cells upon partitioning

- [x] Write tests!

davydden:
> However, I find the "project" format awkward to use because it does not allow referencing projects from pull requests

You need to add PR/issue as a card into the project and then you will see on the right of the web interface (same as in this issue), which project is the PR/issue assigned to.

offtopic: to me the project are a better way because one can move the card into `In Progress` column which would indicate that one of the developers (or more) are working on it. One can also add more columns which better reflect the project needs. With issues it is also possible to indicate that someone is working on this using `Assigness`, but IMHO projects give a better overview of all relevant PRs and issues and Notes.

drwells:
Should we have a milestone for this?

bangerth:
[...]

PR #8325, opened 2019-06-20
Introduced hp::Refinement namespace.

marcfehling:
As pointed out in step-27 for hp adaptive methods, we need decision criteria whether h or p adaptation should be supplied. The tutorial presents one particular decision algorithm and implements it from scratch., but we could provide a selection of different algorithms to the use

This PR introduces a namespace with a selection of different decision algorithms. We tried to provide this functionality as general as possible, so that it could be easily adopted to individual needs. Please see the namespace documentation for a detailed description.

Part of #3511 and #7515.

@bangerth @davydden -- FYI

### DIFF
--- a/include/deal.II/hp/refinement.h
+++ b/include/deal.II/hp/refinement.h
@@ -1,0 +1,355 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_hp_refinement_h
+#define dealii_hp_refinement_h
+
+
+#include <deal.II/base/config.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+// forward declarations
+template <typename Number>
+class Vector;
+
+namespace hp
+{
+  template <int dim, int spacedim>
+  class DoFHandler;
+}
+
+namespace hp
+{
+  /**
+   * We supply adaptive methods to align computational ressources with the
+   * complexity of the numerical solution. Error estimates are an appropriate
+   * means of determining where adjustments need to be made.
+   *
+   * However with hp adaptivity, we have two ways to realise these adjustments:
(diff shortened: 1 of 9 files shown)


marcfehling:
Many of the provided implementations contain similar syntax.

It would be worth a thought to reduce duplicates: Each function `p_adaptivity_from_threshold()`, `p_adaptivity_from_regularity()` and `p_adaptivity_from_prediction` could prepare a container of flags itself and forward it to the function `p_adaptivity_from_flags()`. This would add another loop over all cells, but would make the code easier to read.

I'll prepare a corresponding commit. You may decide which path to take ;)

marcfehling:
Namespace and function names are open for discussion. :)

Forgot to mention that the error prediction algorithm by Melenk corresponding to `p_adaptivity_from_prediction()` will be provided in a follow-up PR.

marcfehling:
ping

include/deal.II/hp/refinement.h
```diff
@@ -0,0 +1,355 @@
...
+}
+
+namespace hp
+{
+  /**
+   * We supply adaptive methods to align computational ressources with the
+   * complexity of the numerical solution. Error estimates are an appropriate
+   * means of determining where adjustments need to be made.
+   *
+   * However with hp adaptivity, we have two ways to realise these adjustments:
```
kronbichler:
```suggestion
   * However with hp adaptivity, we have two ways to realize these adjustments:
```
(We try to consistently use American English spelling in deal.II.)

include/deal.II/hp/refinement.h
```diff
@@ -0,0 +1,355 @@
...
+   * For irregular solutions, h adaptive methods which dynamically assign cell
+   * sizes tend to reduce the approximation error, while for smooth solutions p
+   * adaptive methods are better suited in which function spaces will be
+   * selected dynamically. This namespace offers all tools to decide which type
```
kronbichler:
```suggestion
   * selected dynamically. This namespace collects tools to decide which type
```

include/deal.II/hp/refinement.h
```diff
@@ -0,0 +1,355 @@
...
+      const double                         p_refine_fraction  = 0.5,
+      const double                         p_coarsen_fraction = 0.5);
+
+    /**
+     * Adapt the finite element on cells based on the regularity of the
+     * (unknown) analytical solution.
+     *
+     * With an approximation of the local Sobolev regularity index $k_K$,
+     * we may assess to which finite element space our local solution on cell
+     * $K$ belongs. Since the regularity index is only an etimate, we won't
```
kronbichler:
```suggestion
     * $K$ belongs. Since the regularity index is only an estimate, we won't
```

PR #8302, opened 2019-06-04
Changed extract_dofs so that it works in parallel

mathsen:
Current implementation of extract_dofs for hp::DoFHandler doesn't work in parallel, since the call to "get_fe()" throws an exception:

> The violated condition was: 
>     (dynamic_cast<const dealii::DoFHandler<DoFHandlerType::dimension, DoFHandlerType::space_dimension> *>( this->dof_handler) != nullptr) || (this->is_locally_owned() || this->is_ghost())
> Additional information: 
>     You can only query active_fe_index information on cells that are either locally owned or (after distributing degrees of freedom) are ghost cells.

I propose the small attached changes, which make the function run in parallel.

PR #8326, opened 2019-06-20
Indent comments.

bangerth:
It's the little things that make software great, right? ;-)

Follow-up to #8302.


masterleinad:
The indentation is correct and it's just documentation. There is no reason to bother the CI tester.

PR #8327, opened 2019-06-23
Some updates to the introduction of step-65.

bangerth:
While there, also leave a few more links to step-65.


examples/step-65/doc/intro.dox
```diff
@@ -69,18 +78,39 @@ suitable averages of the surrounding points in the Cartesian coordinate system
...
 
 At this point, one might assume that curved volume descriptions are the way to
-go. However, this becomes impossible for as simple a case as the
-two-dimensional disk because the polar manifold degenerates in the origin and
-would not produce reasonable new points. A similar thing happens at the origin
-of the three-dimensional ball when one tries to attach a spherical manifold to
-the whole volume &endash; in this case, the computation of new manifold points
-would abort with an exception. These two simple examples make it clear that
+go. This is generally not wrong, though it is sometimes not so easy to
+describe how exactly this should work. Here are couple examples:
```
kronbichler:
```suggestion
describe how exactly this should work. Here are a couple of examples:
```

examples/step-65/doc/intro.dox
```diff
@@ -317,7 +372,7 @@ curved mesh elements is the same as the degree of the polynomials for the
 numerical solution. If the degree of the geometry is higher or lower than the
 solution, one calls that a super- or sub-parametric geometry representation,
 respectively. In deal.II, the standard class for polynomial representation is
-MappingQGeneric. If this class is used with polynomial degree $4$ in 3D, a
+MappingQGeneric. If, for example, this class is used with polynomial degree $4$ in 3D, a
 total of 125 ($=(4+1)^3$) points are needed for the tri-cubic
```
kronbichler:
```suggestion
total of 125 ($=(4+1)^3$) points are needed for the
```
(I know this is a mistake entirely in my original text, but since you are so close it would be great if you could remove it. I was confused by degree 3 that is the degree of the polynomials in the actual code, but the mapping uses one degree higher.)

bangerth:
Updated as requested :-)

PR #8328, opened 2019-06-24
Improve vectorized transpose with AVX-512

kronbichler:
The functions `vectorized_load_and_transpose()` and `vectorized_transpose_and_store()` for AVX-512 were rewritten to better utilize execution units with two load units, 1-2 store units, and one shuffle/swizzle unit as common on recent Intel CPUs.

I also marked all vectorized load functions as `DEAL_II_ALWAYS_INLINE` to ensure that the compiler aggregates it with surrounding code (or puts separate functions there).

I verified the implementation with the pre-existing test `base/vectorization_05.cc` that really helps with all corner cases.

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -876,9 +876,9 @@ vectorized_load_and_transpose(const unsigned int       n_entries,
 {
   // do not do full transpose because the code is long and will most
   // likely not pay off because many processors have two load units
-  // (for the top 16 instructions) but only 1 permute unit (for the 8
-  // shuffle/unpack instructions). rather do the transposition on the
-  // vectorized array of half the size, m256d
+  // (for the top 8 instructions) but only 1 permute unit (for the 8
+  // shuffle/unpack instructions). rather start the transposition on the
+  // vectorized array of half the size with 256 bits
   const unsigned int n_chunks = n_entries / 4;
   for (unsigned int i = 0; i < n_chunks; ++i)
     {
@@ -902,6 +902,7 @@ vectorized_load_and_transpose(const unsigned int       n_entries,
       out[4 * i + 2].data = _mm512_unpacklo_pd(v1, v3);
       out[4 * i + 3].data = _mm512_unpackhi_pd(v1, v3);
     }
+  // remainder loop of work that does not divide by 4
   for (unsigned int i = 4 * n_chunks; i < n_entries; ++i)
     out[i].gather(in + i, offsets);
 }
@@ -919,9 +920,8 @@ vectorized_transpose_and_store(const bool                     add_into,
                                const unsigned int *           offsets,
                                double *                       out)
 {
-  // do not do full transpose because the code is long and will most
-  // likely not pay off. rather do the transposition on the vectorized
-  // array of half the size, m256d
+  // as for the load, we split the store operations into 256 bit units to
+  // better balance between code size, shuffle instructions, and stores
   const unsigned int n_chunks = n_entries / 4;
   __m512i mask1 = _mm512_set_epi64(0xd, 0xc, 0x5, 0x4, 0x9, 0x8, 0x1, 0x0);
   __m512i mask2 = _mm512_set_epi64(0xf, 0xe, 0x7, 0x6, 0xb, 0xa, 0x3, 0x2);
@@ -978,6 +978,8 @@ vectorized_transpose_and_store(const bool                     add_into,
           _mm256_storeu_pd(out + 4 * i + offsets[7], res7);
         }
     }
+
+  // remainder loop of work that does not divide by 4
   if (add_into)
     for (unsigned int i = 4 * n_chunks; i < n_entries; ++i)
       for (unsigned int v = 0; v < 8; ++v)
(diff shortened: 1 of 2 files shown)


include/deal.II/base/vectorization.h
```diff
@@ -1278,57 +1279,53 @@ class VectorizedArray<float>
...
-          *reinterpret_cast<__m256 *>(
-            reinterpret_cast<float *>(&out[4 * i + 3].data) + outer) =
-            _mm256_shuffle_ps(v1, v3, 0xdd);
-        }
-      for (unsigned int i = 4 * n_chunks; i < n_entries; ++i)
-        for (unsigned int v = 0; v < 8; ++v)
-          out[i][v + outer] = in[offsets[v + outer] + i];
+      // To avoid warnings about uninitialized variables, need to initialize
+      // one variable with some old value.
+      __m512 t0, t1, t2, t3 = out[0].data;
```
tjhei: Does it make sense to read a value here instead of initializing to ``{}`` or something?

include/deal.II/base/vectorization.h
```diff
@@ -868,47 +868,42 @@ class VectorizedArray<double>
...
+      __m512d v0          = _mm512_shuffle_f64x2(t0, t2, 0x88);
+      __m512d v1          = _mm512_shuffle_f64x2(t0, t2, 0xdd);
+      __m512d v2          = _mm512_shuffle_f64x2(t1, t3, 0x88);
+      __m512d v3          = _mm512_shuffle_f64x2(t1, t3, 0xdd);
+      out[4 * i + 0].data = _mm512_unpacklo_pd(v0, v2);
+      out[4 * i + 1].data = _mm512_unpackhi_pd(v0, v2);
+      out[4 * i + 2].data = _mm512_unpacklo_pd(v1, v3);
+      out[4 * i + 3].data = _mm512_unpackhi_pd(v1, v3);
     }
+  for (unsigned int i = 4 * n_chunks; i < n_entries; ++i)
```
tjhei: here (and in other places): can you please leave a comment, that these are the "left-over" pieces done manually?

include/deal.II/base/vectorization.h
```diff
@@ -1278,57 +1279,53 @@ class VectorizedArray<float>
...
+      t0 = _mm512_insertf32x4(t0, _mm_loadu_ps(in + offsets[0] + 4 * i), 0);
+      t0 = _mm512_insertf32x4(t0, _mm_loadu_ps(in + offsets[4] + 4 * i), 1);
+      t0 = _mm512_insertf32x4(t0, _mm_loadu_ps(in + offsets[8] + 4 * i), 2);
+      t0 = _mm512_insertf32x4(t0, _mm_loadu_ps(in + offsets[12] + 4 * i), 3);
+      t1 = _mm512_insertf32x4(t3, _mm_loadu_ps(in + offsets[1] + 4 * i), 0);
+      t1 = _mm512_insertf32x4(t1, _mm_loadu_ps(in + offsets[5] + 4 * i), 1);
+      t1 = _mm512_insertf32x4(t1, _mm_loadu_ps(in + offsets[9] + 4 * i), 2);
+      t1 = _mm512_insertf32x4(t1, _mm_loadu_ps(in + offsets[13] + 4 * i), 3);
+      t2 = _mm512_insertf32x4(t3, _mm_loadu_ps(in + offsets[2] + 4 * i), 0);
+      t2 = _mm512_insertf32x4(t2, _mm_loadu_ps(in + offsets[6] + 4 * i), 1);
```
tjhei: I have to admit that I have no idea what is going on here with the permutations in the indexes. :-)
 (don't worry and don't bother explaining).

include/deal.II/base/vectorization.h
kronbichler: Out of the 250 lines I changed you spot the one that is critical. :smile: 

I would have liked to use `t3 = {}` like I did for the other parts but it turns out that gcc-9 generates wrong code (at least to the extent that `vectorization_05.release` fails because it wrongly assumes alignment of the `in + offsets[v]` points to 16 byte which cannot be guaranteed. It works with all other compilers I have tried on godbolt. I am in the process of filing a compiler bug report once I have the case reduced and verified that I'm not doing something obviously wrong first.

In any case, I will add a short remark here.

include/deal.II/base/vectorization.h
kronbichler: Yes.

include/deal.II/base/vectorization.h
tjhei: Alright. Compiler bugs are scary.

Can you move the declaration of these temporaries outside the loop? Or does that break performance?

kronbichler: That is a good idea I should have had myself during the last 3 hours of debugging. It indeed works and it will not alter performance notably, judging from the assembly. In any case, it will be better than the spurious load. I will push it.

bangerth:
On 6/24/19 5:14 AM, Martin Kronbichler wrote:
> The functions |vectorized_load_and_transpose()| and 
> |vectorized_transpose_and_store()| for AVX-512 were rewritten to better 
> utilize execution units with two load units, 1-2 store units, and one 
> shuffle/swizzle unit as common on recent Intel CPUs.

A "swizzle"? That's not a real word!? :-)

kronbichler:
> Can you confirm that this still works with AVX2 and AVX512?

I ran all `base` and `matrix_free` tests with AVX2 and AVX-512, respectively (on two different machines), and they all pass.

kronbichler:
> A "swizzle"? That's not a real word!? :-)

Do you want me to drop it? I just took it from some Intel documents
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#cats=Swizzle

I guess the main reason is some marketing buzzwords. We could call it "shuffle/permute" if you prefer which are at least words I have seen in other contexts.

jppelteret:
> A "swizzle"? That's not a real word!? :-)

Actually [it is](https://www.diffordsguide.com/encyclopedia/317/cocktails/swizzles-and-how-to-swizzle-a-cocktail) :-D I know it in the sense of to "mix/churn something up".

bangerth:
On 6/24/19 8:29 AM, Martin Kronbichler wrote:
> <https://software.intel.com/sites/landingpage/IntrinsicsGuide/#cats=Swizzle>
> 
> I guess the main reason is some marketing buzzwords. We could call it 
> "shuffle/permute" if you prefer which are at least words I have seen in 
> other contexts.

Oh, I have no preference, I was just amused by the use of a word I 
didn't know :-) If that's how Intel calls it, we should leave it as is!

kronbichler:
> > A "swizzle"? That's not a real word!? :-)
>
> Actually it is (kind of) :-D I know it in the sense of to "mix/churn something up".

This is exactly what those kind of instructions are supposed to do: You have a vector of 2-16 elements and permute the entries within that vector or among two such vectors. So I guess that term is not too bad a fit.

include/deal.II/base/vectorization.h
kronbichler: It turned out that it is not enough - if the compiler is able to see that `n_entries` is 4 and thus there is no loop, it might inline the whole thing and again emit the wrong instruction. I've added the load back (now outside the loop) and linked to the bugzilla of gcc that I opened this morning so we keep track of it.

PR #8331, opened 2019-06-26
Introduced CellDataTransfer.

marcfehling:
This PR introduces the `CellDataTransfer` class for non-distributed triangulations. The distributed pendant has been introduced in #7537.

Similarly to the `SolutionTransfer` class, we store `active_cell_indices` on the unrefined mesh and use them to transfer data from the initial to the refined mesh.

We use the `CoarseningStrategy` struct to decide how to handle data during coarsening. However for `CellDataTransfer`, we need the values on all coarsened cells not before, but after coarsening, when the cells no longer exist. Thus, some changes were necessary.

For more flexibility, we moved the `CoarseningStrategy` struct out of the scope of the `p::d::CellDataTransfer` class. It is now an independent struct in the `dealii` namespace defined in `deal.II/numerics/cell_data_transfer.h`. Shall we introduce a separate header for this struct, e.g. `deal.II/numerics/coarsening_strategy.h`?
EDIT: Moved it to `grid/grid_tools.h` now.

Requested by #7515.

### DIFF
--- a/include/deal.II/distributed/cell_data_transfer.templates.h
+++ b/include/deal.II/distributed/cell_data_transfer.templates.h
@@ -37,23 +37,33 @@
 DEAL_II_NAMESPACE_OPEN
 
 
-namespace
+namespace internal
 {
-  template <typename VectorType>
-  void
-  post_refinement_action(std::vector<VectorType *> &all_out)
+  namespace parallel
   {
-    for (auto &out : all_out)
-      out->compress(::dealii::VectorOperation::insert);
-  }
+    namespace distributed
+    {
+      namespace CellDataTransferImplementation
+      {
+        template <typename VectorType>
+        void
+        post_unpack_action(std::vector<VectorType *> &all_out)
+        {
+          for (auto &out : all_out)
+            out->compress(::dealii::VectorOperation::insert);
+        }
 
-  template <typename ValueType>
-  void
-  post_refinement_action(std::vector<std::vector<ValueType> *> &)
-  {
-    // Do nothing for std::vector as VectorType.
-  }
-} // namespace
+        template <typename value_type>
+        void
+        post_unpack_action(std::vector<std::vector<value_type> *> &)
+        {
+          // Do nothing for std::vector as VectorType.
+        }
+      } // namespace CellDataTransferImplementation
+    }   // namespace distributed
+  }     // namespace parallel
+} // namespace internal
+
 
 
 namespace parallel
@@ -212,7 +222,8 @@ namespace parallel
                   std::placeholders::_3,
                   std::ref(all_out)));
 
-      post_refinement_action(all_out);
+      dealii::internal::parallel::distributed::CellDataTransferImplementation::
+        post_unpack_action(all_out);
 
       input_vectors.clear();
     }
--- a/include/deal.II/distributed/cell_data_transfer.templates.h
+++ b/include/deal.II/distributed/cell_data_transfer.templates.h
@@ -63,17 +63,60 @@ namespace parallel
     template <int dim, int spacedim, typename VectorType>
     CellDataTransfer<dim, spacedim, VectorType>::CellDataTransfer(
       const parallel::distributed::Triangulation<dim, spacedim> &triangulation,
-      const bool                         transfer_variable_size_data,
-      const std::function<typename VectorType::value_type(
-        const typename parallel::distributed::Triangulation<dim, spacedim>::
-          cell_iterator & parent,
-        const VectorType &input_vector)> coarsening_strategy)
+      const bool transfer_variable_size_data,
+      const std::function<value_type(
+        const std::vector<value_type> &children_values)> coarsening_strategy)
       : triangulation(&triangulation, typeid(*this).name())
       , transfer_variable_size_data(transfer_variable_size_data)
       , coarsening_strategy(coarsening_strategy)
       , handle(numbers::invalid_unsigned_int)
     {}
 
+    template <int dim, int spacedim, typename VectorType>
+    DEAL_II_DEPRECATED
+    CellDataTransfer<dim, spacedim, VectorType>::CellDataTransfer(
+      const parallel::distributed::Triangulation<dim, spacedim> &triangulation,
+      const bool transfer_variable_size_data,
+      const std::function<
+        value_type(const typename parallel::distributed::
+                     Triangulation<dim, spacedim>::cell_iterator &parent,
+                   const VectorType &input_vector)> coarsening_strategy)
+      : triangulation(&triangulation, typeid(*this).name())
+      , transfer_variable_size_data(transfer_variable_size_data)
+      , handle(numbers::invalid_unsigned_int)
+    {
+      value_type (*const *old_strategy)(
+        const typename parallel::distributed::Triangulation<dim, spacedim>::
+          cell_iterator &,
+        const VectorType &) =
+        coarsening_strategy.template target<
+          value_type (*)(const typename parallel::distributed::
+                           Triangulation<dim, spacedim>::cell_iterator &,
+                         const VectorType &)>();
+
+      if (*old_strategy == CoarseningStrategies::check_equality)
+        const_cast<
+          std::function<value_type(const std::vector<value_type> &)> &>(
+          this->coarsening_strategy) =
+          &dealii::CoarseningStrategies::check_equality<value_type>;
+      else if (*old_strategy == CoarseningStrategies::sum)
+        const_cast<
+          std::function<value_type(const std::vector<value_type> &)> &>(
+          this->coarsening_strategy) =
+          &dealii::CoarseningStrategies::sum<value_type>;
+      else if (*old_strategy == CoarseningStrategies::mean)
+        const_cast<
+          std::function<value_type(const std::vector<value_type> &)> &>(
+          this->coarsening_strategy) =
+          &dealii::CoarseningStrategies::mean<value_type>;
+      else
+        Assert(
+          false,
+          ExcMessage(
+            "The constructor using the former function type of the "
+            "coarsening_strategy parameter is no longer supported. Please use "
+            "the latest function type instead"));
+    }
 
 
     // Interface for packing
@@ -226,8 +269,7 @@ namespace parallel
                                                           spacedim>::CellStatus
         status)
     {
-      std::vector<typename VectorType::value_type> cell_data(
-        input_vectors.size());
+      std::vector<value_type> cell_data(input_vectors.size());
 
       // Extract data from input_vectors for this particular cell.
       auto it_input  = input_vectors.cbegin();
@@ -247,10 +289,26 @@ namespace parallel
 
               case parallel::distributed::Triangulation<dim,
                                                         spacedim>::CELL_COARSEN:
-                // Cell is parent whose children will get coarsened to.
-                // Decide data to store on parent by provided strategy.
-                *it_output = coarsening_strategy(cell, *(*it_input));
-                break;
+                {
+                  // Cell is parent whose children will get coarsened to.
+                  // Decide data to store on parent by provided strategy.
+                  std::vector<value_type> children_values(cell->n_children());
+                  for (unsigned int child_index = 0;
+                       child_index < cell->n_children();
+                       ++child_index)
+                    {
+                      const auto child = cell->child(child_index);
+                      Assert(child->active() && child->coarsen_flag_set(),
+                             typename dealii::Triangulation<
+                               dim>::ExcInconsistentCoarseningFlags());
+
+                      children_values[child_index] =
+                        (**it_input)[child->active_cell_index()];
+                    }
+
+                  *it_output = coarsening_strategy(children_values);
+                  break;
+                }
 
               default:
                 Assert(false, ExcInternalError());
@@ -281,21 +339,20 @@ namespace parallel
         &                        data_range,
       std::vector<VectorType *> &all_out)
     {
-      std::vector<typename VectorType::value_type> cell_data;
+      std::vector<value_type> cell_data;
 
       // We have to unpack the corresponding datatype that has been packed
       // beforehand.
       if (all_out.size() == 1)
-        cell_data.push_back(Utilities::unpack<typename VectorType::value_type>(
+        cell_data.push_back(Utilities::unpack<value_type>(
           data_range.begin(),
           data_range.end(),
           /*allow_compression=*/transfer_variable_size_data));
       else
-        cell_data =
-          Utilities::unpack<std::vector<typename VectorType::value_type>>(
-            data_range.begin(),
-            data_range.end(),
-            /*allow_compression=*/transfer_variable_size_data);
+        cell_data = Utilities::unpack<std::vector<value_type>>(
+          data_range.begin(),
+          data_range.end(),
+          /*allow_compression=*/transfer_variable_size_data);
 
       // Check if sizes match.
       Assert(cell_data.size() == all_out.size(), ExcInternalError());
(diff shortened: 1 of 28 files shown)


marcfehling:
Do we need a `DEPRECATED` version of the old `CoarseningStrategies` struct?

marcfehling:
ping

jppelteret:
Hi @marcfehling. Sorry that its taken so long for one of us to respond. It seems like we're all preoccupied with other things at the moment -- not ideal at all, especially given that we're really interested in, and grateful for, your contributions. I think that this is cool, and this sort of mechanisms to transfer data would be helpful to facilitate that transfer of data in the `CellDataStorage` class, which I believe might only be possible in the distributed case right now. Thanks a lot!

> It is now an independent struct in the dealii namespace defined in deal.II/numerics/cell_data_transfer.h. Shall we introduce a separate header for this struct, e.g. deal.II/numerics/coarsening_strategy.h?

Since `CoarseningStrategy` is referenced in both `deal.II/distributed/cell_data_transfer.h` and `deal.II/numerics/cell_data_transfer.h `, I think that putting it in a some common header is a good idea. I see that you've now put it in `GridTools`. That's quite a big header to drag into `cell_data_transfer.h`, so I think that your original idea of putting it in `deal.II/numerics/coarsening_strategy.h` might be best. Since you've also put it in its own `doxygen` section, its clear that its intended use is quite context specific.

> Do we need a DEPRECATED version of the old CoarseningStrategies struct and the old p::d::CellDataTransfer constructor?
> EDIT: Flagging parts of this class as DEPRECATED is not that easy: We store an std::function which takes a different set of parameters with this PR. Thus, we need to deprecate the old pointer and introduce the new pointer at the same time. And the callback function would have to call the corresponding one which is not a nullptr.
> Shall we follow this approach? Or shall we accept this incompatibility?

Hmm... its always best for us to deprecate functionality if its at all possible. But this does seem like a bit of a difficult case to handle. If I understand correctly, you're taking about this function:
OLD:
```c++
const std::function<typename VectorType::value_type(
  const typename parallel::distributed::Triangulation<dim, spacedim>::
  cell_iterator & parent,
  const VectorType &input_vector)>
coarsening_strategy;
```
NEW:
```c++
const std::function<value_type(
  const std::vector<value_type> &children_values)>
coarsening_strategy;
```
We could keep and deprecate the old `CoarseningStrategy` struct in the distributed `CellDataTransfer`, and in the (old/deprecated) constructor just translate its meaning by forwarding it on to the new constructor. Something like
```c++
CellDataTransfer(
  const parallel::distributed::Triangulation<dim, spacedim> &triangulation,
  const bool transfer_variable_size_data = false
  const std::function<typename VectorType::value_type(
    const typename parallel::distributed::Triangulation<dim, spacedim>::
    cell_iterator & parent,
    const VectorType &input_vector)>
  coarsening_strategy = &CoarseningStrategies::check_equality)
: CellDataTransfer (
  triangulation, transfer_variable_size_data,
  (coarsening_strategy == &CoarseningStrategies::check_equality 
    ? &GridTools::CoarseningStrategies<value_type>::check_equality 
    : ...<check all options; maybe do with with some internal helper struct>... )
)
```
Do you think that this could work?

drwells:
We have a ton of open PRs now and I tagged myself with the intent of reviewing this early this week. @jppelteret are you willing to take this over?

jppelteret:
Sorry @drwells, I wish I could but I'm not able say with certainty when my next free chunk of time will be :-( I had a bit of a unicorn moment today when I had some time to devote to browse through the open PRs, and so I added my 2c here! If the opportunity arises then I will come back and look at this, but (if possible) that would likely only be next weekend. But I'd certainly appreciate it if you're able to give it some scrutiny when you have a free moment.

drwells:
I will review this later today; @jppelteret's notes are a good start.

marcfehling:
Moved `CoarseningStrategies` to a separate header.
Thank you @jppelteret for the hints on deprecating the constructor. The implementation turned out to be a bit whacky, but serves its purpose well I guess. However I could not find a way to make this work for general functions with the old set of parameters and will throw an exception instead.

include/deal.II/numerics/cell_data_transfer.h
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * @code
+ * Vector<double> data_to_transfer (triangulation.n_active_cells());
+ * //[fill data_to_transfer with cell-wise values...]
+ *
+ * PETScWrappers::MPI::Vector
+ * distributed_data_to_transfer (mpi_communicator,
+ *                               triangulation.n_active_cells(),
+ *                               triangulation.n_locally_owned_active_cells());
+ * for (const auto cell& : triangulation.active_cell_iterators ())
```
drwells:
```suggestion
 * for (const auto &cell : triangulation.active_cell_iterators())
```

include/deal.II/numerics/cell_data_transfer.h
```diff
@@ -0,0 +1,212 @@
...
+  /**
+   * Constructor.
+   *
+   * @param[in] triangulation The triangulation on which all operations will
+   *   happen. At the time when this constructor is called, the refinement
+   *   in question has not happened yet.
+   * @param[in] transfer_variable_size_data Specify whether your VectorType
+   *   container stores values that differ in size. A varying amount of data
+   *   may be packed per cell, if for example the underlying ValueType of
+   *   the VectorType container is a container itself.
```
drwells: It doesn't look like `transfer_variable_size_data` is in this constructor. Should it be?

include/deal.II/numerics/cell_data_transfer.h
```diff
@@ -0,0 +1,212 @@
...
+#ifdef DEBUG
+  /**
+   * Number of active cells on the initial triangulation that has not been
+   * refined yet.
+   *
+   * It will be set in prepare_for_coarsening_and_refinement() and used to
+   * validate user inputs after refinement happened.
+   */
+  unsigned int n_active_cells_pre;
+#endif
```
drwells: I think it would be better to just write
```cpp
(void)n_active_cells_pre;
```
instead of changing the class members in different modes.

include/deal.II/numerics/cell_data_transfer.templates.h
```diff
@@ -0,0 +1,196 @@
...
+    out.compress(::dealii::VectorOperation::insert);
+  }
+
+  template <typename value_type>
+  void
+  post_unpack_action(std::vector<value_type> &)
+  {
+    // Do nothing for std::vector as VectorType.
+  }
+} // namespace
```
drwells: Having anonymous namespaces in headers is a bit risky: I recommend putting these in `namespace internal` instead.

include/deal.II/numerics/coarsening_strategies.h
```diff
@@ -0,0 +1,94 @@
...
+ * words, which data should be stored in the corresponding parent cell.
+ *
+ * In this namespace, we offer a few strategies that cope with this
+ * problem. Such strategies can be passed to the CellDataTransfer and
+ * parallel::distributed::CellDataTransfer constructors.
+ *
+ * @author Marc Fehling, 2019
+ */
+template <typename value_type>
+struct CoarseningStrategies
```
drwells: Would this work if we used a namespace instead, or would that screw up taking addresses of functions?

include/deal.II/numerics/coarsening_strategies.h
```diff
@@ -0,0 +1,94 @@
...
+ * how to handle data of child cells which will be coarsened. Or in other
+ * words, which data should be stored in the corresponding parent cell.
+ *
+ * In this namespace, we offer a few strategies that cope with this
+ * problem. Such strategies can be passed to the CellDataTransfer and
+ * parallel::distributed::CellDataTransfer constructors.
+ *
+ * @author Marc Fehling, 2019
+ */
+template <typename value_type>
```
drwells: Do we have some template type check we could use here to ensure that `median` would work correctly?

include/deal.II/numerics/coarsening_strategies.h
```diff
@@ -0,0 +1,94 @@
...
+#include <vector>
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * When data is transferred during coarsening, it is not trivial to decide
+ * how to handle data of child cells which will be coarsened. Or in other
+ * words, which data should be stored in the corresponding parent cell.
+ *
+ * In this namespace, we offer a few strategies that cope with this
```
drwells:
```suggestion
 * In this struct, we offer a few strategies that cope with this
```

include/deal.II/numerics/coarsening_strategies.h
```diff
@@ -0,0 +1,94 @@
...
+    Assert(children_values.size() > 0, ExcInternalError());
+    const value_type value = children_values[0];
+
+    for (unsigned int index = 1; index < children_values.size(); ++index)
+      Assert(value == children_values[index],
+             ExcMessage(
+               "Values on cells that will be coarsened are not equal!"));
+
+    return value;
+  }
```
drwells: It doesn't make a big difference, but we usually separate function definitions and declarations (even if they are in the same file). We should do this.

include/deal.II/numerics/cell_data_transfer.templates.h
```diff
@@ -0,0 +1,196 @@
...
+
+  // Transfer data of the parent cell to all of its children that it has been
+  // refined to.
+  for (const auto &refined : refined_cells_active_index)
+    for (unsigned int child_index = 0;
+         child_index < refined.first->n_children();
+         ++child_index)
+      {
+        const auto child = refined.first->child(child_index);
+        Assert(child->active(), ExcInternalError());
```
drwells: Is this always true? Couldn't I do the following:
```
+-----------+
|           |
|           |
|           |
|           |
|           |
+-----------+
```
to
```
+--+--+-----+
|__|__|     |
|  |  |     |
+--+--+-----+
|     |     |
|     |     |
+-----+-----+
```
then a refined cell could have grandchildren.

include/deal.II/numerics/cell_data_transfer.templates.h
```diff
@@ -0,0 +1,196 @@
...
+          if (coarsened_cells_active_index.find(parent) ==
+              coarsened_cells_active_index.end())
+            {
+              std::set<unsigned int> indices_children;
+              for (unsigned int child_index = 0;
+                   child_index < parent->n_children();
+                   ++child_index)
+                {
+                  const auto sibling = parent->child(child_index);
+                  Assert(sibling->active(), ExcInternalError());
```
drwells: Like my refinement example: could we coarsen more than one level at once? I think we can but perhaps there is some other precondition enforced that I don't know about.

include/deal.II/numerics/cell_data_transfer.templates.h
```diff
@@ -0,0 +1,196 @@
...
+void
+CellDataTransfer<dim, spacedim, VectorType>::unpack(const VectorType &in,
+                                                    VectorType &      out)
+{
+#ifdef DEBUG
+  Assert(in.size() == n_active_cells_pre,
+         ExcDimensionMismatch(in.size(), n_active_cells_pre));
+  Assert(out.size() == triangulation->n_active_cells(),
+         ExcDimensionMismatch(out.size(), triangulation->n_active_cells()));
+#endif
```
drwells:
```suggestion
#else
  (void)n_active_cells_pre;
#endif
```

include/deal.II/numerics/cell_data_transfer.h
marcfehling: Woops, I blindly copied that from the `p::d::CellDataTransfer` class. Thanks for noticing!

include/deal.II/numerics/coarsening_strategies.h
marcfehling: I just tried it out: Works like a charm!

include/deal.II/numerics/coarsening_strategies.h
marcfehling: Were you referring to the `mean` function? I guess the code wouldn't compile if `value_type` has no zero value and no `operator+=`.

Otherwise I don't know what you mean by `median`. Would you elaborate on that?

include/deal.II/numerics/cell_data_transfer.templates.h
marcfehling: This is indeed possible.

However, `Triangulation::prepare_coarsening_and_refinement()` cleans up all coarsening flags:  We only keep these flags whenever all siblings are active and flagged for coarsening as well. We require this specific function to be called beforehand.

This means that the exception thrown should not be `ExcInternalError`, but a custom one giving users a hint towards the `prepare()` function. Thank you for pointing that out!

marcfehling:
@drwells, thank your for your extensive review! Addressed all suggestions but one. Would you give me more information on the remaining one.

I will squash all commits into one once you approve this PR. I cleaned up some other code on the last commit since there were exceptions that recur. Let me know what you think of it!

include/deal.II/numerics/coarsening_strategies.h
marcfehling: Are you suggesting a new `median` function that returns the median of all child values? (The number of children is always even)

drwells: Sorry, I meant `mean`.

marcfehling: We may use the  concepts feature that has been introduced in [C++20](https://en.cppreference.com/w/cpp/language/constraints) or [Boost 1.4](https://www.boost.org/doc/libs/1_70_0/libs/concept_check/concept_check.htm) for this.

But I think using [type traits](https://en.cppreference.com/w/cpp/types) to check for arithmetic types should be sufficient for now until someone demands a more versatile implementation. But I don't know how to sum/mean `bool` variables - we may just exclude them.

include/deal.II/distributed/cell_data_transfer.templates.h
```diff
@@ -247,10 +300,26 @@ namespace parallel
...
+                  // Decide data to store on parent by provided strategy.
+                  std::vector<value_type> children_values(cell->n_children());
+                  for (unsigned int child_index = 0;
+                       child_index < cell->n_children();
+                       ++child_index)
+                    {
+                      const auto child = cell->child(child_index);
+                      Assert(child->active() && child->coarsen_flag_set(),
+                             typename dealii::Triangulation<
+                               dim>::ExcObscureCoarseningFlags());
```
marcfehling: Do you think placing this particular one in `Triangulation<dim>` is reasonable?

source/distributed/cell_weights.cc
```diff
@@ -145,18 +145,23 @@ namespace parallel
...
               dof_handler->get_fe_collection().find_dominated_fe_extended(
                 fe_indices_children, /*codim=*/0);
 
             Assert(fe_index != numbers::invalid_unsigned_int,
-                   ExcMessage(
-                     "No FiniteElement has been found in your FECollection "
-                     "that is dominated by all children of a cell you are "
-                     "trying to coarsen!"));
+                   typename dealii::hp::FECollection<
+                     dim>::ExcNoDominatedFiniteElementAmongstChildren());
```
marcfehling: Do you think placing this particular one in `hp::FECollection<dim>` is reasonable?

drwells:
Yes, the exceptions look good, but I think the name `ExcObscureCoarseningFlags` is misleading; the flags are not obscure (though this error surely is :)), they are inconsistent. How about `ExcInconsistentCoarseningFlags`?

Otherwise :+1:

marcfehling:
Did some cleanup and squashed the commits into four separate ones. Addressed all suggestions.

Ready for final review and/or testing!

PR #8332, opened 2019-06-27
Revert #8312: Replace boost random

masterleinad:
Given the test current failures, I propose to just revert #8312. We could of course just add more alternative output files for each compiler, but that is pretty annoying and brittle.

Trying to use `std::shuffle` with `::boost::mt19937` doesn't compile since `::boost::mt19937::max` is not `constexpr`. We can neither use `boost::range::random_shuffle` since that uses `std::random_shuffle` that is removed in `C++17`.


drwells:
I thought about it some more, and I think using boost here is better than the standard library since it will use the same algorithms across all compilers.

PR #8337, opened 2019-07-02
 Fix a bug in CUDA matrix-free with adapted mesh when using Volta

Rombur:
On Volta the code would deadlock. I am not 100% sure why it would deadlock but Volta introduced a new thread scheduling and `__syncthreads()`, which is a barrier, has also been changed. This PR just moves the code around so that `__syncthreads()` is not called inside a `if()`.

I have also changed the CUDA tester to test two different architecture (Kepler and Volta) but the Kepler cards are reaching end-of-life so I will probably have to switch to full Volta soon.


doc/news/changes/minor/20190702BrunoTurcksin
```diff
@@ -0,0 +1,4 @@
+Fixed: CUDAWrappers::MatrixFree::cell_loop() would deadlock on adaptively
+refined mesh when using Volta GPU.
```
masterleinad:
```suggestion
refined meshes when using Volta GPU.
```

include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
```diff
@@ -967,26 +967,28 @@ namespace CUDAWrappers
      */
     template <int dim, int fe_degree, bool transpose, typename Number>
     __device__ void
-    resolve_hanging_nodes(const unsigned int constr, Number *values)
+    resolve_hanging_nodes(const unsigned int constraint_mask, Number *values)
     {
+      printf("resolve_hanging_nodes: dim %i\n", dim);
```
masterleinad:
```suggestion
```

include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
```diff
@@ -758,80 +758,82 @@ namespace CUDAWrappers
...
         const unsigned int y_idx = threadIdx.y;
 
         const unsigned int this_type =
           (direction == 0) ? internal::constr_type_x : internal::constr_type_y;
 
-        if (constr & (((direction == 0) ? internal::constr_face_y : 0) |
-                      ((direction == 1) ? internal::constr_face_x : 0)))
+        const unsigned int interp_idx = (direction == 0) ? x_idx : y_idx;
+
+        Number t = 0;
```
masterleinad: What is this variable for?

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -238,9 +238,8 @@ namespace CUDAWrappers
     // Use the read-only data cache.
     values[idx] = __ldg(&src[src_idx]);
 
-    if (constraint_mask)
```
masterleinad: So this might not be consistent across all threads?

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -258,9 +257,8 @@ namespace CUDAWrappers
   {
     static_assert(n_components_ == 1, "This function only supports FE with one \
                   components");
-    if (constraint_mask)
```
masterleinad: See above.

include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
```diff
@@ -758,80 +758,82 @@ namespace CUDAWrappers
...
+                                  (y_idx == fe_degree))) ||
+          ((direction == 1) && ((constraint_mask & internal::constr_type_x) ?
+                                  (x_idx == 0) :
+                                  (x_idx == fe_degree)));
+        if (constrained_face && constrained_dof)
           {
-            const unsigned int interp_idx = (direction == 0) ? x_idx : y_idx;
+            const bool type = constraint_mask & this_type;
 
-            __syncthreads();
```
masterleinad: So you are essentially just removing this synchronization point...

include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
```diff
@@ -758,80 +758,82 @@ namespace CUDAWrappers
...
+                                                     fe_degree - interp_idx] :
+                        internal::constraint_weights[(fe_degree - interp_idx) *
+                                                       (fe_degree + 1) +
+                                                     fe_degree - i];
+                    t += w * values[real_idx];
                   }
               }
+          }
 
-            __syncthreads();
```
masterleinad: ... and make sure every thread synchronizes here and the rest is cosmetics?

Rombur: Yeah exactly

include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
Rombur: This is a variable that we use to compute the constrained value. It needs to be defined here because it is used in the two `if` that surround the `__synchtreads()`. Note that this is not a new variable, I have just moved it up.

Rombur:
Done

include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
masterleinad: I know that it is not new, but I still don't understand what `t` is an abbreviation for. Anyway, not crucial to me.

masterleinad:
Did you also test on a `Kepler` system?

Rombur:
> Did you also test on a Kepler system?

No, it's annoying to do with my current setup that's why I have set one of the tester to Kepler and one to Volta.

Rombur:
I have added a comment

PR #8064, opened 2019-05-10
Constexpr Tensor

rezarastak:
Following #8011 and being careful about #8037, this is the second step towards #8003.
1. Defined the macro `DEAL_II_CONSTEXPR` that represents `constexpr` in c++14-enabled compilers and disappears in older compilers with c++11. Let me know if I defined it the correct way.
1. Added `constexpr` to simple functions and `DEAL_II_CONSTEXPR` to complicated functions in `tensor.h` and `tensor_accessors.h`.
1. Removed `inline` from functions that are `constexpr` because `constexpr` implies `inline` based on the standard.
1.  Removed `DEAL_II_ALWAYS_INLINE` from functions that are `constexpr`. My reason is that any function that has a single return statement (satisfies c++11 `constexpr` requirements) will be inlined by the compiler in almost all circumstances.
1. Functions decorated with `DEAL_II_CONSTEXPR` still include `inline` and `DEAL_II_ALWAYS_INLINE` attributes.
1. Changed the constructors for `Tensor<0, dim, Number>` and `Tensor<rank, dim, Number>`. The constructors are now `constexpr` (c++11). I had to use variadic templates to make it work. It looks relatively elegant but I needed to add a private helper constructor to `Tensor<0, dim, Number>`.
1. Because of the aforementioned changes to the constructors, I had to change the constructors of `VectorizedArray<Number>` as well.

The result of all of these changes is that we can compile the following code in a portable way and make use `constexpr` features.
```
constexpr double tensor_0_dim_constexpr(){
  constexpr Tensor<0,2> a{7.6};
  return a;
}

DEAL_II_CONSTEXPR double tensor_2_dim_calc(){
  DEAL_II_CONSTEXPR const double a[3][3] =
    {{4., 2., 0.}, {2., 3., 0.},{0., 0., 5.}};
  DEAL_II_CONSTEXPR const Tensor<2,3> A(a);
  DEAL_II_CONSTEXPR const auto det =  determinant(A);  // 40
  DEAL_II_CONSTEXPR const auto tr = trace(transpose(A) * A); // 58
  return det + tr;
}

void use_constexpr_functions(){
  constexpr auto x = tensor_0_dim_constexpr();   // returns 7.6
  DEAL_II_CONSTEXPR auto y = tensor_2_dim_calc();  // returns 98.0
}
```
If c++14 is enabled, the assembly code created for the function `use_constexpr_functions()` in debug mode is:
```
   1dcaa:       55                      push   %rbp
[...]

PR #8338, opened 2019-07-03
CUDA device annotations for Tensor

masterleinad:
Based on #8064, this pull request adds `CUDA` annotations for all member functions and free functions related to Tensor that can be used in `CUDA` code so far.

Only the last two commits contain the real changes here.

### DIFF
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -940,8 +940,7 @@ template <typename OtherNumber>
 DEAL_II_CONSTEXPR inline bool
 Tensor<0, dim, Number>::operator==(const Tensor<0, dim, OtherNumber> &p) const
 {
-#if defined(DEAL_II_ADOLC_WITH_ADVANCED_BRANCHING) && \
-  !defined(DEAL_II_COMPILER_CUDA_AWARE)
+#if defined(DEAL_II_ADOLC_WITH_ADVANCED_BRANCHING)
   Assert(!(std::is_same<Number, adouble>::value ||
            std::is_same<OtherNumber, adouble>::value),
          ExcMessage(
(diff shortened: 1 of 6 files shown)


include/deal.II/base/tensor.h
```diff
@@ -687,11 +728,13 @@ class Tensor
...
-   *  to create constexpr constructors for Tensor<rank, dim, Number>
+   * to create constexpr constructors for Tensor<rank, dim, Number>
+   *
+   * @note This function can also be used in CUDA device code.
    */
   template <typename ArrayLike, std::size_t... Indices>
-  constexpr Tensor(const ArrayLike &initializer,
-                   std_cxx14::index_sequence<Indices...>);
+  constexpr DEAL_II_CUDA_HOST_DEV
+  Tensor(const ArrayLike &initializer, std_cxx14::index_sequence<Indices...>);
```
Rombur: Does that work with CUDA 8.0? There is no support for c++14 in CUDA 8.0 and we can't fall back on boost...

tests/cuda/cuda_tensor_02.cu
```diff
@@ -0,0 +1,252 @@
...
+  Tensor<rank, dim, Number>                      dummy_1(array);
+  Tensor<rank, dim, Number>                      dummy_2;
+  Tensor<rank, dim, Number>                      dummy_3 = dummy_2;
+
+  // access
+  Tensor<rank + 1, dim, Number> initializer_1;
+  const auto                    dummy_5 = initializer_1[0];
+
+  // assignment
+  dummy_2 = dummy_3;
```
Rombur: This test is really weak. We should check the result.

include/deal.II/base/tensor.h
masterleinad: We can test that.

tests/cuda/cuda_tensor_02.cu
masterleinad: Yes. The only problems I have seen lately is compiling and running. Hence, I didn't bother testing the result to be identical with the host result so far.

include/deal.II/base/tensor.h
masterleinad: The test also passes using `CUDA 8.0`.

masterleinad: In the end, `std_cxx14::index_sequence` doesn't even use `boost`.

masterleinad:
I addressed the comments and rebased.

masterleinad:
This should also fix https://cdash.43-1.org/testDetails.php?test=474777&build=95.

PR #8128, opened 2019-05-13
Disable intrinsics includes if DEAL_II_VECTORIZATION_LEVEL=0

kronbichler:
This disables all includes to intrinsics headers in case `DEAL_II_VECTORIZATION_LEVEL=0` as suggested by https://github.com/dealii/dealii/pull/8125#issuecomment-491659211

This supersedes #8125 
This fixes #8075.

PR #8340, opened 2019-07-04
Fix include to intrinsics

kronbichler:
We previously switched from version-specific intrinsics headers to `#include <x86intrin.h>` in #8128 but we forgot one spot. This makes the selection (and whether we enable any of these features) consistent with a single header.


masterleinad:
Looks good.

PR #3967, opened 2017-02-12
Vectorization of SymmetricTensor

MathiasMen:
Dear all,

I am currently trying to solve a vector-valued equation using MatrixFree and my weak form requires the double contraction of symmetric tensors. While trying to double contract `SymmetricTensor<4,3,VectorizedArray<double>>` with `SymmetricTensor<2,3,VectorizedArray<double>>`  I encountered a compile time-error because internally `perform_double_contraction` (line 1210 within [symmetric_tensor.h](https://github.com/dealii/dealii/tree/master/include/deal.II/base/symmetric_tensor.h)) is called which would require a constructor of sort `VectorizedArray<double>::VectorizedArray(some_scalar)`. I implemented the constructor for the various specializations and added a unit test. 

Thanks in advance for your feedback.

Mathias

PR #8342, opened 2019-07-04
 Extend VectorizedArray with a second template argument to control vector length

peterrum:
This PR adds an optional second template argument to the class `VectorizedArray`. This template
parameter controls the number of vectorization lanes to be used, i.e. which instruction set architecture extension to be used. 

This means that `VectorizedArray::n_array_elements` is not fixed at compile time but rather an upper limit is set (depending on the given hardware and compiler flags).  The evaluation of the template argument is postponed until the compilation of the application, enabling `VectorizedArray` with different number of lanes in the same application. 

The maximal number of lanes are given by:
```cpp
internal::VectorizedArrayWidthSpecifier<Number>::width;
```
and is used as the default value if the user write - as until now - `VectorizedArray<Number>`.

Adding the second parameter, lets the user explicitly control the used ISA extension. For example on Intel Skylake, you have the following options:
```cpp
VectorizedArray<double, 1> // no vectorization (auto-optimization)
VectorizedArray<double, 2> // SSE2
VectorizedArray<double, 4> // AVX2
VectorizedArray<double, 8> // AVX-512
```
Not supported lane numbers lead to a static assert.

This PR is part of the effort to introduce new vectorization strategies in `MatrixFree`: no vectorization, vectorization over quadrature points, vectorization over elements. The first version will internally use `VectorizedArray<Number, 1>` and the latter two as usual `VectorizedArray<Number>`.


include/deal.II/base/numbers.h
```diff
@@ -37,8 +37,55 @@
...
+#endif
+  };
+
+
+} // namespace internal
+
 // forward declarations to support abs or sqrt operations on VectorizedArray
-template <typename Number>
+template <typename Number,
+          unsigned int N =
```
kronbichler:
```suggestion
          int width =
```
(We always use `int` in template argument type even in case the underlying type is strictly unsigned like for the dimension.)

include/deal.II/base/numbers.h
```diff
@@ -37,8 +37,55 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+namespace internal
+{
+  template <typename Number>
+  struct VectorizedArrayWidthSpecifier
+  {
+    static const int width = 1;
```
kronbichler:
```suggestion
    static const unsigned int max_width = 1;
```

include/deal.II/base/numbers.h
```diff
@@ -37,8 +37,55 @@
...
+  };
+
+  template <>
+  struct VectorizedArrayWidthSpecifier<double>
+  {
+    static const int width =
```
kronbichler: Same here: `width` -> `max_width`.

include/deal.II/base/vectorization.h
```diff
@@ -111,14 +111,12 @@ namespace internal
...
-template <typename Number>
-struct EnableIfScalar<VectorizedArray<Number>>
+template <typename Number, unsigned int width>
+struct EnableIfScalar<VectorizedArray<Number, width>>
 {
   using type = VectorizedArray<typename EnableIfScalar<Number>::type>;
 };
 
-
-
```
kronbichler: Please keep these two empty lines in.

include/deal.II/base/vectorization.h
```diff
@@ -180,6 +178,10 @@ class VectorizedArray
    */
   static const unsigned int n_array_elements = 1;
 
+  static_assert(width == n_array_elements,
+                "There seems to be something wrong with "
+                "th specified number of elements!");
```
```suggestion
                "The base variant of this class only supports a single entry in the array, i.e., width=1!");
```

include/deal.II/base/vectorization.h
```diff
@@ -437,43 +439,47 @@ class VectorizedArray
...
+  friend VectorizedArray<Number2, width2>
+  std::min(const VectorizedArray<Number2, width2> &,
+           const VectorizedArray<Number2, width2> &);
 };
 
 // We need to have a separate declaration for static const members
-template <typename Number>
-const unsigned int VectorizedArray<Number>::n_array_elements;
-
-
```
kronbichler: Please keep these two empty lines to provide three empty lines between the declaration of the member and the `make_vectorized_array()` function. Ideally, also insert two additional lines before this declaration, to comply with the coding style of deal.II: https://www.dealii.org/developer/doxygen/deal.II/CodingConventions.html

include/deal.II/base/vectorization.h
```diff
@@ -221,7 +223,7 @@ class VectorizedArray
    */
   DEAL_II_ALWAYS_INLINE
   VectorizedArray &
-  operator+=(const VectorizedArray<Number> &vec)
+  operator+=(const VectorizedArray<Number, width> &vec)
```
kronbichler: For consistency with the specialized classes further down, can you please write this as follows:
```suggestion
  operator+=(const VectorizedArray &vec)
```
here and for the following functions? I know this is pre-existing, but it's good to be consistent.

include/deal.II/base/vectorization.h
```diff
@@ -437,43 +439,47 @@ class VectorizedArray
...
-  VectorizedArray<Number> result;
+  VectorizedArray<Number, width> result;
   result = u;
   return result;
 }
 
 
 
+#if true
+
```
kronbichler: Can you please remove these two lines?

include/deal.II/base/vectorization.h
```diff
@@ -551,27 +559,29 @@ vectorized_load_and_transpose(const unsigned int       n_entries,
...
     for (unsigned int i = 0; i < n_entries; ++i)
-      for (unsigned int v = 0; v < VectorizedArray<Number>::n_array_elements;
+      for (unsigned int v = 0;
+           v < VectorizedArray<Number, width>::n_array_elements;
            ++v)
         out[offsets[v] + i] = in[i][v];
 }
 
-
+#endif
```
kronbichler: Please remove.

include/deal.II/base/vectorization.h
```diff
@@ -1453,15 +1467,15 @@ vectorized_transpose_and_store(const bool                    add_into,
         out[offsets[v] + i] = in[i][v];
 }
 
+#endif
 
-
-#elif DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 2 && defined(__AVX__)
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 2 && defined(__AVX__)
```
kronbichler: Nice!

tests/base/vectorization_10.cc
```diff
@@ -0,0 +1,181 @@
...
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// test transpose operations of vectorized array using the array+offset method
+// (otherwise the same as vectorization_05)
```
kronbichler:
```suggestion
// for the set of all supported vectorization widths (otherwise the same as vectorization_05)
```

include/deal.II/base/vectorization.h
```diff
@@ -2854,10 +2874,10 @@ class VectorizedArray<float>
  */
```
kronbichler: Please adjust friend declarations in lines 2856-2867 above (and in the other specializations if appropriate). The CI tests of clang-tidy and on OSX complain about a private `get_abs()` function.

tjhei:
This is great! I was thinking that it would be neat to have this instead of several deal.II installations to test vectorization performance.

include/deal.II/base/vectorization.h
```diff
@@ -180,6 +180,11 @@ class VectorizedArray
    */
   static const unsigned int n_array_elements = 1;
 
+  static_assert(
+    width == n_array_elements,
+    "There seems to be something wrong with "
+    "The base variant of this class only supports a single entry in the array, i.e., width=1!");
```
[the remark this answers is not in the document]
peterrum: I will extend the doxygen class documentation during the day.

To your questions. `VectorizedArray<Number, width>` is specilized for 8 cases:
```cpp
VectorizedArray<double, 1>
VectorizedArray<double, 2>
VectorizedArray<double, 4>
VectorizedArray<double, 8>
VectorizedArray<float, 1>
VectorizedArray<float, 4>
VectorizedArray<float, 8>
VectorizedArray<float, 16>
```
If you specify any other `width`, the compiler will go to the default implementation (which has `n_array_elements=1`) and will through a static error since `width == n_array_elements` is not fulfilled.

include/deal.II/base/vectorization.h
```diff
@@ -180,6 +180,11 @@ class VectorizedArray
    */
   static const unsigned int n_array_elements = 1;
 
+  static_assert(
+    width == n_array_elements,
+    "There seems to be something wrong with "
```
kronbichler: Remove this line.

include/deal.II/base/vectorization.h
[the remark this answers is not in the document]
tjhei: Ok. Then we should update the error message above to be more helpful (as in: you specified an illegal width that is not supported).

peterrum:
I have found some issues related `memory_consumption` and `NumberType`. I first have to fix these.

peterrum:
When I was integrating the modified `VectorizedArray` in `MatrixFree` (see #8348 ), I have experienced some issues with `memory_constumtion()` and `NumberType`.

Writing it like this does - for some reason - not work:
```cpp
template <typename T, int width>
inline std::size_t
memory_consumption(const VectorizedArray<T, width> &);


template <int rank, int dim, typename T, int width>
struct NumberType<Tensor<rank, dim, VectorizedArray<T, width>>>;


template <typename T, unsigned int width>
struct NumberType<VectorizedArray<T, width>>;
```
In my last commit 0f93b8293910bf6a8010ab05c456ec2d30f43e5e, I explicitly specialized the classes/functions for float/double and 1/2/4/8/16 such that I was able to compile and successfully run programs with multiple `MatrixFree`s with different `VectorizedArray<Number, width>` in the same application.

I guess this issue is somehow related to #3967.

I guess we should solve this problem in this PR?

masterleinad:
> When I was integrating the modified `VectorizedArray` in `MatrixFree` (see #8348 ), I have experienced some issues with `memory_constumtion()` and `NumberType`.
> 
> Writing it like this does - for some reason - not work:
> [...]

Out of curiosity, what kind of problems did you see?

peterrum:
@masterleinad: for some reason the compiler did not find the specialization for `VectorizedArray`
```cpp
template <typename T, unsigned int width>
struct NumberType<VectorizedArray<T, width>>;
```
but used the standard implementation:
```cpp
template <typename T>
struct NumberType
```
leading to the circumstance that I could not pass `double` to `NumberType<VectorizedArray<double>>`.

bangerth:
That should have worked. Rather than trying to find a different way to achieve this, see if you can figure out why the compiler didn't recognize the specialization. (Was it defined in a file in which the compiler can't see it, for example?)

peterrum:
@bangerth: Thanks for you answer. I have finished the extension of `MatrixFree` with a second template argument in #8348. Now that is working, I have had a second look at the issue: and it turns out that it is actually working. Sorry for the confusion.

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
  * similar to std::vector otherwise but always makes sure that data is
  * correctly aligned.
  *
+ * The user can explicitly control the used instruction set architecture (ISA)
+ *  extension by specifying the vector width/number of lanes via  the second
+ * template parameter of this wrapper class. For example on Intel Skylake,
+ * you have the following options for the data type double:
+ *  - VectorizedArray<double, 1> // no vectorization (auto-optimization)
+ *  - VectorizedArray<double, 2> // SSE2
+ *  - VectorizedArray<double, 4> // AVX2
```
```suggestion
 *  - VectorizedArray<double, 4> // AVX
```
The width has been introduced by AVX two years before AVX2 and we should honor the main innovation, which was AVX. We hardly use AVX2-specific stuff, only for gather and implicitly via the compiler to transform MUL + ADD into FMA. And for the latter, Intel typically considers FMA a separate feature besides AVX2.

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
...
+ *  - VectorizedArray<double, 8> // AVX-512 (default)
+ *
+ * and for Intel SandyBridge, Haswell, and Broadwell:
+ *  - VectorizedArray<double, 1> // no vectorization (auto-optimization)
+ *  - VectorizedArray<double, 2> // SSE2
+ *  - VectorizedArray<double, 4> // AVX2 (default)
```
```suggestion
 *  - VectorizedArray<double, 4> // AVX (default)
```

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
...
+ * The user can explicitly control the used instruction set architecture (ISA)
+ *  extension by specifying the vector width/number of lanes via  the second
+ * template parameter of this wrapper class. For example on Intel Skylake,
+ * you have the following options for the data type double:
+ *  - VectorizedArray<double, 1> // no vectorization (auto-optimization)
+ *  - VectorizedArray<double, 2> // SSE2
+ *  - VectorizedArray<double, 4> // AVX2
+ *  - VectorizedArray<double, 8> // AVX-512 (default)
+ *
+ * and for Intel SandyBridge, Haswell, and Broadwell:
```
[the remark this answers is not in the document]
kronbichler: Furthermore, please also add AMD Bulldozer and Zen/Ryzen to this list.

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
...
+ *  - VectorizedArray<double, 1> // no vectorization (auto-optimization)
+ *  - VectorizedArray<double, 2> // SSE2
+ *  - VectorizedArray<double, 4> // AVX2 (default)
+ *
```
kronbichler: Please add an additional category:

```
for older x86 processors or in case no processor-specific compilation flags
were added (i.e., without `-D CMAKE_CXX_FLAGS=-march=native` or similar
flags):
- VectorizedArray<double, 1> // no vectorization (auto-optimization)
- VectorizedArray<double, 2> // SSE2
```

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
...
+ * Similar considerations apply also for the data type float.
+ *
+ * Wrongly selecting the width i.e width=3 or width>8 for Skylake leads to
+ * static assert.
+ *
+ * @tparam Number underlying data type
+ * @tparam width  vector length
+ *
  * @author Katharina Kormann, Martin Kronbichler, 2010, 2011
```
```suggestion
  * @author Katharina Kormann, Martin Kronbichler, Peter Munch, 2010, 2011, 2019
```

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
...
+ *  - VectorizedArray<double, 4> // AVX2 (default)
+ *
+ * and for processors with AltiVec support:
+ *  - VectorizedArray<double, 1>
+ *  - VectorizedArray<double, 2>
+ *
+ * Similar considerations apply also for the data type float.
+ *
+ * Wrongly selecting the width i.e width=3 or width>8 for Skylake leads to
+ * static assert.
```
```suggestion
 * a static assert.
```

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
...
+ * and for Intel SandyBridge, Haswell, and Broadwell:
+ *  - VectorizedArray<double, 1> // no vectorization (auto-optimization)
+ *  - VectorizedArray<double, 2> // SSE2
+ *  - VectorizedArray<double, 4> // AVX2 (default)
+ *
+ * and for processors with AltiVec support:
+ *  - VectorizedArray<double, 1>
+ *  - VectorizedArray<double, 2>
+ *
+ * Similar considerations apply also for the data type float.
```
```suggestion
 * Similar considerations also apply to the data type `float`.
```

include/deal.II/base/vectorization.h
```diff
@@ -167,9 +167,35 @@ struct EnableIfScalar<VectorizedArray<Number>>
...
+ *
+ * Wrongly selecting the width i.e width=3 or width>8 for Skylake leads to
+ * static assert.
+ *
+ * @tparam Number underlying data type
+ * @tparam width  vector length
```
```suggestion
 * @tparam width  vector length (optional; if not set, the maximal width of the 
 *                architecture is used)
```

peterrum: @kronbichler: thanks for the comments!

include/deal.II/base/numbers.h
```diff
@@ -37,8 +37,55 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+namespace internal
+{
+  template <typename Number>
+  struct VectorizedArrayWidthSpecifier
```
bangerth: Can you add documentation to this class and to the specializations below?

That's because in some sense, this class may be in an internal namespace, but it's in the public interface as it provides the default template argument for the `VectorizedArray` class, and so should be documented.

include/deal.II/base/numbers.h
```diff
@@ -37,8 +37,55 @@
...
+  {
+    static const unsigned int max_width = 1;
```
bangerth: This and the corresponding variables in the specializations could be made `constexpr`.

peterrum:
@bangerth: I will implement the changes in a separate PR. Is that fine?

kronbichler:
> I will implement the changes in a separate PR. Is that fine?

Yes, this is what we always do.

bangerth:
Indeed. Thanks for offering to do so!

peterrum:
@bangerth @kronbichler I have created the PR #8424 for this issues.

PR #8345, opened 2019-07-05
Add maxwell equations tutorial

dangars:
This tutorial shows how to:
* Calculate the eigenvalues and eigenvectors of an electromagnetic cavity.
* Calculate the quality factor using an eigenvalue calculation
* Solve the maxwell equations using `NedelecSZ`
* Use parallel/MPI `std::complex` SLEPc`
* Do an spectral transformation to accelarate the eigenvalue calculation

This tutorial is complementary to step-62. Both approaches lead to the same result. An eigenvalue calculation with an spectral transformation is an order of magnitude faster than the approach taken in step-62, but step-62 allows to calculate the bandgap. I'll discuss all this in the tutorial

Todo list:
- [x] Write the code
- [ ] Write the documentation

### DIFF
--- a/cmake/checks/check_02_compiler_features.cmake
+++ b/cmake/checks/check_02_compiler_features.cmake
@@ -70,10 +70,7 @@ ENDIF()
 #
 CHECK_CXX_SOURCE_COMPILES(
   "
-  #include <emmintrin.h>
-#ifdef __AVX512F__
-  #include <immintrin.h>
-#endif
+  #include <x86intrin.h>
   int main()
   {
     __m128d a, b;


bangerth:
@dangars I hope you didn't close this PR because you couldn't get feedback for it. I apologize for not getting to patch reviews lately, in particular to the very large ones. I do hope that we get your program eventually!

dangars:
@bangerth Thanks for your message. I closed it because I have too many projects in parallel and had to prioritize. In France it is grant writing season.

I also wanted to finish  #8419 Which I finished today. It works and the test gives the expected result but I'm not sure I'm taking the best approach, you can take a look ;)

I'm going to give a final push to this PR a re-open it very soon!

Issue #8351, opened 2019-07-07
FE_PolyTensor does not support mapping_covariant for shape functions


GrahamBenHarper:
If I were to take a derived class of `FE_PolyTensor` and replace the mapping type with `mapping_covariant`, then the library builds just fine, but running simulations using said derived class does not work. More specifically, it throws an exception (which I can paste in a comment below) when asked for the value or gradient of a function mapped by the covariant transform. This happens exactly when FEValues calls `reinit(cell)`.

This physically makes sense because the covariant transform most frequently appears when taking the pullback of the gradient of some finite element vector field.

I thought I'd raise this as an issue anyway because it is probably worth having a discussion on whether mapping shape function vector fields by the covariant transform should be allowed in the future or not.

GrahamBenHarper:
Here's a sample output of when it fails:
```
--------------------------------------------------------
An error occurred in line <3735> of file </home/graham/Programming/cpp/Deal.II/deal.II-master/source/fe/mapping_q_generic.cc> in function
    void dealii::internal::MappingQGenericImplementation::{anonymous}::transform_differential_forms(const dealii::ArrayView<const dealii::DerivativeForm<rank, dim, spacedim> >&, dealii::MappingType, const typename dealii::Mapping<dim, spacedim>::InternalDataBase&, const dealii::ArrayView<dealii::Tensor<(rank + 1), spacedim> >&) [with int dim = 3; int spacedim = 3; int rank = 1; typename dealii::Mapping<dim, spacedim>::InternalDataBase = dealii::Mapping<3, 3>::InternalDataBase]
The violated condition was: 
    false
Additional information: 
    You are trying to use functionality in deal.II that is currently not implemented. In many cases, this indicates that there simply didn't appear much of a need for it, or that the author of the original code did not have the time to implement a particular case. If you hit this exception, it is therefore worth the time to look into the code to find out whether you may be able to implement the missing functionality. If you do, please consider providing a patch to the deal.II development sources (see the deal.II website on how to contribute).

Stacktrace:
-----------
#0  /usr/local/lib/libdeal_II.g.so.9.2.0-pre: 
[...]

Issue #7970, opened 2019-04-26
Issue with sign_change in FEPolyTensor


bangerth:
A student in my department (hi @GrahamBenHarper!) is working on some modifications in `fe_poly_tensor.cc` and asked a question that led me down a rabbit hole at the end of which was a huge "Oh sh**t!" sign.

Here's the issue: `FEPolyTensor` is the base class for the Raviart-Thomas and Nedelec elements. The shape functions for these elements are vectors, and they need to agree between neighboring cells. If one has a rectangular mesh, then (to pick just one example) the shape functions of the RT element on the left and right faces of each cell point to the right. This also happens to be with the coordinate directions in each cell separately. But it is not difficult to construct meshes where the local coordinate directions on a face are different between the two adjacent cells disagrees, and in this case one has to pick one direction and on one of the two adjacent cells, one will have to multiply the shape function by `-1`.

`FEPolyTensor` does this in its `fill_fe_*_values()` functions, which in essence look like this:
```
FE_PolyTensor<PolynomialType, dim, spacedim>::fill_fe_values(...)
{
  // Create table with sign changes, due to the special structure of the RT
  // elements.
  // TODO: Preliminary hack to demonstrate the overall prinicple!

  // Compute eventual sign changes depending on the neighborhood
  // between two faces.
  std::fill(fe_data.sign_change.begin(), fe_data.sign_change.end(), 1.0);

  if (mapping_type == mapping_raviart_thomas)
    internal::FE_PolyTensor::get_face_sign_change_rt(cell,
                                                     this->dofs_per_face,
                                                     fe_data.sign_change);
  else if (mapping_type == mapping_nedelec)
    internal::FE_PolyTensor::get_face_sign_change_nedelec(cell,
                                                          this->dofs_per_face,
                                                          fe_data.sign_change);

  for (unsigned int i = 0; i < this->dofs_per_cell; ++i)
    {
      ...
          switch (mapping_type)
            {
              case mapping_none:
                {
                  ...
                }

              case mapping_covariant:
              case mapping_contravariant:
                {
                  ...
                }

              case mapping_raviart_thomas:
              case mapping_piola:
                {
                  mapping.transform(make_array_view(fe_data.shape_values, i),
                                    mapping_piola,
                                    mapping_internal,
                                    make_array_view(
                                      fe_data.transformed_shape_values));
                  for (unsigned int k = 0; k < n_q_points; ++k)
                    for (unsigned int d = 0; d < dim; ++d)
[...]

PR #8352, opened 2019-07-07
Add functionality to support multiple mappings per element in FE_PolyTensor

GrahamBenHarper:
These changes to `FE_PolyTensor` generalize it so that derived classes may support different mapping types for each shape function.

The motivation for this is the existence of many different kinds of "enriched" finite elements, where some vector fields may be mapped with no mapping (such as the Q_1^d vector field), but others may be mapped so that the vector field remains normal to the element (such as bubble functions). This way future implementation of finite element classes will not be restricted.

I've run a build and most of the tests and everything comes back clean so far. I also experimented with the `FE_BernardiRaugel` class and changing the mappings to
```
this->mapping_type = 
{mapping_none, mapping_none, mapping_none, mapping_none,
 mapping_none, mapping_none, mapping_none, mapping_none,
 mapping_piola, mapping_piola, mapping_piola, mapping_piola}
```
in 2D (3D case used 24 mapping_nones and 6 mapping_piolas). The shape functions evaluate correctly when this is done, and also when `mapping_contravariant` is used instead of `mapping_piola`.

The only case that didn't pass my own testing is `mapping_covariant`, but I'm fairly certain that `FE_PolyTensor` in its current form does not support mapping vector fields with `mapping_covariant` in the first place. I raised an issue so that can also be discussed as needed #8351.

I tried squashing all the commits down as much as I could, but then I forgot a couple minor things. Please remind me later, and I can squash commits down again, but I would need to be reminded which git commands I need to use in order to do so.

I look forward to your feedback!

### DIFF
--- a/include/deal.II/fe/fe_poly_tensor.h
+++ b/include/deal.II/fe/fe_poly_tensor.h
@@ -216,6 +216,18 @@ protected:
    */
   std::vector<MappingType> mapping_type;
 
+  /**
+   * Returns a boolean that is true when the finite element uses a single
+   * mapping and false when the finite element uses multiple mappings.
+   */
+  bool
+  single_mapping() const;
+
+  /**
+   * Returns MappingType @p i for the finite element.
+   */
+  MappingType
+  get_mapping_type(unsigned int i) const;
 
   /* NOTE: The following function has its definition inlined into the class
      declaration because we otherwise run into a compiler error with MS Visual
@@ -254,19 +266,17 @@ protected:
     // allocate memory
 
     const bool update_transformed_shape_values =
-      (std::find_if(this->mapping_type.begin(),
-                    this->mapping_type.end(),
-                    [](const MappingType t) { return t != mapping_none; }) !=
-       this->mapping_type.end());
+      std::any_of(this->mapping_type.begin(),
+                  this->mapping_type.end(),
+                  [](const MappingType t) { return t != mapping_none; });
 
     const bool update_transformed_shape_grads =
-      (std::find_if(this->mapping_type.begin(),
-                    this->mapping_type.end(),
-                    [](const MappingType t) {
-                      return (t == mapping_raviart_thomas ||
-                              t == mapping_piola || t == mapping_nedelec ||
-                              t == mapping_contravariant);
-                    }) != this->mapping_type.end());
+      std::any_of(this->mapping_type.begin(),
+                  this->mapping_type.end(),
+                  [](const MappingType t) {
+                    return (t == mapping_raviart_thomas || t == mapping_piola ||
+                            t == mapping_nedelec || t == mapping_contravariant);
+                  });
 
     const bool update_transformed_shape_hessian_tensors =
       update_transformed_shape_values;
--- a/include/deal.II/fe/fe_poly_tensor.h
+++ b/include/deal.II/fe/fe_poly_tensor.h
@@ -129,11 +129,13 @@ DEAL_II_NAMESPACE_OPEN
  * transformations can be selected from the set MappingType and stored in
  * #mapping_type. Therefore, each constructor should contain a line like:
  * @code
- * this->mapping_type = mapping_none;
+ * this->mapping_type = {mapping_none};
  * @endcode
  * (in case no mapping is required) or using whatever value among
  * the ones defined in MappingType is appropriate for the element you
- * are implementing.
+ * are implementing. If each shape function may be mapped by different
+ * mappings, then @p mapping_type may be a vector with the same number
+ * of elements as there are shape functions.
  *
  * @see PolynomialsBDM, PolynomialsRaviartThomas
  * @ingroup febase
--- a/include/deal.II/fe/fe_poly_tensor.h
+++ b/include/deal.II/fe/fe_poly_tensor.h
@@ -207,9 +207,12 @@ public:
 protected:
   /**
    * The mapping type to be used to map shape functions from the reference
-   * cell to the mesh cell.
+   * cell to the mesh cell. If this vector is length one, the same mapping
+   * will be applied to all shape functions. If the vector size is equal to
+   * the finite element dofs per cell, then each shape function will be mapped
+   * according to the corresponding entry in the vector.
    */
-  MappingType mapping_type;
+  std::vector<MappingType> mapping_type;
 
 
   /* NOTE: The following function has its definition inlined into the class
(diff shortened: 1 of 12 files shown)


source/fe/fe_poly_tensor.cc
```diff
@@ -84,66 +83,82 @@ namespace internal
...
                     {
-                      Assert(f * dofs_per_face + j < face_sign.size(),
+                      const unsigned int cell_j = fe.face_to_cell_index(j, f);
+                      // something in FiniteElement or FiniteElementData base
+                      // classes
+
+                      Assert(f * fe.dofs_per_face + j < face_sign.size(),
+                             ExcInternalError());
+                      Assert(mapping_type.size() == 1 ||
+                               cell_j < mapping_type.size(),
```
drwells: I think this check and the one below would be clearer if we just added
```cpp
const bool one_mapping = mapping_type.size() == 1;
```
(or some other similar name) at the top of the function.

source/fe/fe_poly_tensor.cc
```diff
@@ -49,30 +49,29 @@ namespace internal
...
+        const FiniteElement<2, 2> &                    fe,
+        const std::vector<MappingType> &               mapping_type,
         std::vector<double> &                          face_sign)
       {
         const unsigned int dim      = 2;
         const unsigned int spacedim = 2;
 
-        // Default is no sign
-        // change. I.e. multiply by one.
-        std::fill(face_sign.begin(), face_sign.end(), 1.0);
```
drwells: Shouldn't we still fill with ones by default? I don't understand why this is no longer needed.

source/fe/fe_poly_tensor.cc
```diff
@@ -84,66 +83,82 @@ namespace internal
...
+        const std::vector<MappingType> &mapping_type,
+        std::vector<double> &           face_sign)
       {
         const unsigned int dim = 3;
-        std::fill(face_sign.begin(), face_sign.end(), 1.0);
         // TODO: This is probably only going to work for those elements for
         // which all dofs are face dofs
         for (unsigned int l = 0; l < GeometryInfo<dim>::lines_per_cell; ++l)
-          if (!(cell->line_orientation(l)))
+          if (!(cell->line_orientation(l)) & mapping_type[0] == mapping_nedelec)
```
drwells: Did you mean to use `&` instead of `&&` here? I think it will produce the same result since both are `bool`s but I would prefer to use the logical operators.

source/fe/fe_poly_tensor.cc
```diff
@@ -364,18 +379,23 @@ FE_PolyTensor<PolynomialType, dim, spacedim>::fill_fe_values(
...
+
+  internal::FE_PolyTensor::get_face_sign_change_nedelec(cell,
+                                                        *this,
+                                                        this->mapping_type,
+                                                        fe_data.sign_change);
 
 
   for (unsigned int i = 0; i < this->dofs_per_cell; ++i)
     {
+      MappingType mapping_type =
```
drwells:
```suggestion
      const MappingType mapping_type =
```

source/fe/fe_poly_tensor.cc
```diff
@@ -2140,88 +2168,95 @@ FE_PolyTensor<PolynomialType, dim, spacedim>::requires_update_flags(
...
-            out |= update_gradients | update_values |
-                   update_jacobian_pushed_forward_grads;
-
-          if (flags & update_hessians)
-            out |= update_hessians | update_values | update_gradients |
-                   update_jacobian_pushed_forward_grads |
-                   update_jacobian_pushed_forward_2nd_derivatives;
-          break;
-        }
+      MappingType mapping_type =
```
drwells:
```suggestion
      const MappingType mapping_type =
```

GrahamBenHarper: I think I got all of them. Lines 395, 975, 1607, 2173 should all have this and also use the convenience function now.

source/fe/fe_poly_tensor.cc
GrahamBenHarper: Whoops. I got it, thanks!

source/fe/fe_poly_tensor.cc
GrahamBenHarper: Lines 380, 961, 1595 each fill this with ones and then call `get_face_sign_change_rt` and `get_face_sign_change_nedelec` consecutively, so there's no need for the redundant fill with ones.

The idea is to stack possible sign changes for both rt and nedelec in the same vector. Filling with ones inside `get_face_sign_change_rt` or `get_face_sign_change_nedelec` could overwrite sign change information from the other function, and may not generalize as well should there be a need for another sign change function in the future.

source/fe/fe_poly_tensor.cc
GrahamBenHarper: Good plan. These convenience functions- I suppose they should be `const` functions and `protected` privacy, right? Let me know if I did anything silly with this part.

Unfortunately, however, since these sign change functions are not part of the FE_PolyTensor class itself, the convenience functions will not work in this context.

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -247,11 +252,30 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
     // initialize fields only if really
     // necessary. otherwise, don't
     // allocate memory
+
+    const bool update_transformed_shape_values =
+      (std::find_if(this->mapping_type.begin(),
+                    this->mapping_type.end(),
+                    [](const MappingType t) { return t != mapping_none; }) !=
+       this->mapping_type.end());
```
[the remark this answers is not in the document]
GrahamBenHarper: I didn't know about those! Should I use `std::any_of` since updating needs to happen if any of the mapping types are nontrivial?

I applied the same change to the `std::find_if` below this as well.

source/fe/fe_poly_tensor.cc
GrahamBenHarper: Good point!

source/fe/fe_poly_tensor.cc
drwells: Good point. I should read the whole diff.

GrahamBenHarper:
Sounds good! I have a changelog entry written.

drwells:
:+1:

include/deal.II/fe/fe_poly_tensor.h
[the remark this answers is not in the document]
bangerth: Nice -- I had no idea about `std::any_of` either -- which is why I had suggested the `std::find_if` version above to @GrahamBenHarper :-)

bangerth:
Thanks all for shepherding this patch through! Good work, @GrahamBenHarper !

bangerth:
This is also in relation to #7970.

PR #8353, opened 2019-07-08
Add a space between sentences in ExcNumberNotFinite

GrahamBenHarper:
In my adventures I came across a space between words that was missing. I figured it was worth a quick pull request, even though it's an extremely small detail.

Previously ExcNumberNotFinite had a paragraph that was missing a space between two sentences:
> ..."before the place where this error is produced.One way to check for finiteness is to use the 'AssertIsFinite' macro"...

Now it should read:
> ,,,"before the place where this error is produced. One way to check for finiteness is to use the 'AssertIsFinite' macro"...


kronbichler:
Given the small change, I take the risk to not spin up the testers.

PR #8358, opened 2019-07-09
Remove unused code in CUDAWrappers::internal::HangingNodes

masterleinad:
This code line has no effect and seems to be duplicated from https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/cuda_hanging_nodes_internal.h#L890 while restructuring the code in https://github.com/dealii/dealii/pull/8337.


drwells:
I am really surprised that none of the static analysis tools complained about dead code!

PR #8360, opened 2019-07-09
fixed misrepresentation of "*" in documentation of tutorials

rezarastak:
For some reason that I don't know, any line containing a single asterisk within a code block within doxygen is not displayed in the documentation. I searched the tutorial files and I found 3 lines of code that must be slightly reformatted to properly show the multiplication sign `*`


drwells:
Thanks for the fix! I never noticed the problem before now.

I think doxygen assumes such lines are empty lines in a comment block.

drwells:
Since this only changes documentation files I will merge as-is.

PR #8362, opened 2019-07-09
Fix a few gcc-9 warninigs in tests

masterleinad:
Running the test suite with `-Werror` only really gave these two warnings.


kronbichler:
The tester won't run over these, so I'm merging without it

Issue #8371, opened 2019-07-10
intel 19.0.4 AVX512 error


tjhei:
I am getting
```
/scratch1/02578/theister/deal-git/include/deal.II/base/vectorization.h(1355): error: a value of type "dealii::VectorizedArray<float, 16>" cannot be assigned to an entity of type "const __m512 &"
      t3 = out[0];
```

tjhei:
@kronbichler do we need to do a load using instrinsics?

peterrum:
I have fixed this as part of #8348.

tjhei:
Did you discuss with @kronbichler ? I thought ``= {}`` causes a bug in gcc.

peterrum:
@tjhei: you are completely right! I forgot that. I guess @kronbichler wanted to write:
```cpp
  __m512 t0, t1, t2, t3;
  if (n_chunks > 0)
    t3 = out[0].data;
```
@kronbichler's test program was working directly on `__m512*` and not on on `VectorizedArray<float>`.

peterrum:
@tjhei: I would update  #8348 accordingly. Or should it be a separate PR?

kronbichler:
@peterrum thinking about, let us set up a separate PR so it can get in quickly.

peterrum:
I'll do it right away.

tjhei:
yep, that would be great!

tjhei:
closed via #8373

PR #8373, opened 2019-07-10
Fix VectorizedArray::vectorized_load_and_transpose specialization

peterrum:
References #8371.


tjhei:
I can confirm that this works for me. :+1:

kronbichler:
The tester won't see this code, so let's merge.

PR #8375, opened 2019-07-10
update URLs to 43-1.org and use https instead of http

tamiko:


drwells:
Its just documentation so lets merge.

PR #8381, opened 2019-07-12
Improve hermite like interpolation

kronbichler:
The additional roots of the HermiteLikeInterpolation with degree greater than four have been switched to the roots of the Jacobi polynomial `P^{4,4}` rather than `P^{2,2}`, making the interior bubble functions L2 orthogonal and improving the conditioning of interpolation somewhat.


kronbichler:
We have only a few tests with polynomial degrees 5 and higher, which is where this matters. And those tests probably simply compare some matrix-free result with some reference computations. But yes, we should compute the 1D mass matrix for this basis and print it, just for the sake of what we claim. Let me add an additional test to this PR.

kronbichler:
Done, we now have at least 2 tests that depend on the coefficients. :wink:

masterleinad:
OK with me.

Issue #8366, opened 2019-07-10
Cannot compile 9.1.1 with Trilinos and PETSc (macOS)


fmilicchio:
Dear all,

I am trying to compile deal.II 9.1.1 on macOS 10.14.5 with Apple LLVM version 10.0.1 (clang-1001.0.46.4) enabling both Trilinos (12.12.1) and PETSc (3.11.2). Both libraries are compiled from sources, without Fortran or MPI.

First, if I compile with only PETSc it works just fine. 

With **both** Trilinos and PETSc, I have the following error:

```
% make     
[  0%] Built target expand_instantiations_exe
[  0%] Built target doxygen_headers
...
Scanning dependencies of target obj_numerics_debug
[ 43%] Building CXX object source/numerics/CMakeFiles/obj_numerics_debug.dir/data_out.cc.o
In file included from /Users/sensei/Downloads/src/dealii-9.1.1/source/numerics/data_out.cc:23:
In file included from /Users/sensei/Downloads/src/dealii-9.1.1/include/deal.II/fe/fe_values.h:51:
In file included from /Users/sensei/Documents/Projects/libraries/petsc/include/petsc.h:5:
In file included from /Users/sensei/Documents/Projects/libraries/petsc/include/petscbag.h:4:
In file included from /Users/sensei/Documents/Projects/libraries/petsc/include/petscsys.h:132:
/Users/sensei/Documents/Projects/libraries/petsc/include/petsc/mpiuni/mpi.h:142:14: error: typedef redefinition with different types
      ('long' vs 'size_t' (aka 'unsigned long'))
typedef long MPI_Aint;
             ^
/Users/sensei/Documents/Projects/libraries/trilinos/include/mpi.h:193:16: note: previous definition is here
typedef size_t MPI_Aint;
               ^
In file included from /Users/sensei/Downloads/src/dealii-9.1.1/source/numerics/data_out.cc:23:
In file included from /Users/sensei/Downloads/src/dealii-9.1.1/include/deal.II/fe/fe_values.h:51:
In file included from /Users/sensei/Documents/Projects/libraries/petsc/include/petsc.h:5:
In file included from /Users/sensei/Documents/Projects/libraries/petsc/include/petscbag.h:4:
In file included from /Users/sensei/Documents/Projects/libraries/petsc/include/petscsys.h:132:
/Users/sensei/Documents/Projects/libraries/petsc/include/petsc/mpiuni/mpi.h:199:52: error: typedef redefinition with different types
[...]

PR #8382, opened 2019-07-12
Checking if MPI is enabled to allow both Trilinos and PETSc.

fmilicchio:
Dear all,

I am not skilled in large CMake projects, but I've done my best to check the incompatibility between PETSc + Trilinos and not allowing MPI.

I hope this helps.

Cheers,
    Franco

### DIFF
--- a/cmake/configure/configure_2_trilinos.cmake
+++ b/cmake/configure/configure_2_trilinos.cmake
@@ -104,6 +104,25 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
     ENDIF()
 
     #
+    # deal.II has to be configured with MPI if both Trilinos and PETSc are
+    # enabled.
+    #
+    IF( DEAL_II_WITH_TRILINOS AND DEAL_II_WITH_PETSC AND NOT DEAL_II_WITH_MPI )
+      MESSAGE(STATUS "Incompatible configuration settings: "
+        "MPI must be enabled to use both Trilinos and PETSc."
+        )
+      SET(TRILINOS_ADDITIONAL_ERROR_STRING
+        ${TRILINOS_ADDITIONAL_ERROR_STRING}
+        "To enable both Trilinos and PETSc, deal.II must be configured with MPI "
+        "or enable one between Trilinos and PETSc removing MPI, but found:\n"
+        "  DEAL_II_WITH_MPI = ${DEAL_II_WITH_MPI}\n"
+        "  DEAL_II_WITH_TRILINOS = ${DEAL_II_WITH_TRILINOS}\n"
+        "  DEAL_II_WITH_PETSC = ${DEAL_II_WITH_PETSC}\n"
+        )
+      SET(${var} FALSE)
+    ENDIF()
+
+    #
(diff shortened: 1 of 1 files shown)


drwells:
This looks good!

I have a minor request: could you extend the error message to mention that the incompatibility stems from Trilinos and PETSc providing mutually incompatible MPI stubs?

drwells:
Fixes #8366.

fmilicchio:
@drwells I hope you can see this last commit, as requested.

Thanks!

cmake/configure/configure_2_trilinos.cmake
```diff
@@ -103,6 +103,27 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
       SET(${var} FALSE)
     ENDIF()
 
+    #
+    # deal.II has to be configured with MPI if both Trilinos and PETSc are
+    # enabled.
+    #
+    IF( DEAL_II_WITH_TRILINOS AND DEAL_II_WITH_PETSC AND NOT DEAL_II_WITH_MPI )
```
tamiko: Would you mind removing the spaces between the argument and the opening and closing parenthesis?

cmake/configure/configure_2_trilinos.cmake
```diff
@@ -103,6 +103,27 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
...
+	"provide mutually incompatible MPI stubs."
+        )
+      SET(TRILINOS_ADDITIONAL_ERROR_STRING
+        ${TRILINOS_ADDITIONAL_ERROR_STRING}
+        "To enable both Trilinos and PETSc, deal.II must be configured with MPI "
+        "or enable one between Trilinos and PETSc removing MPI. Both Trilinos and PETSc "
+	"provide mutually incompatible MPI stubs, but found:\n"
+        "  DEAL_II_WITH_MPI = ${DEAL_II_WITH_MPI}\n"
+        "  DEAL_II_WITH_TRILINOS = ${DEAL_II_WITH_TRILINOS}\n"
+        "  DEAL_II_WITH_PETSC = ${DEAL_II_WITH_PETSC}\n"
```
tamiko: What about
```
     SET(TRILINOS_ADDITIONAL_ERROR_STRING
        ${TRILINOS_ADDITIONAL_ERROR_STRING}
        "Incompatible Trilinos and PETSc libraries found. Both libraries were"
        "configured without MPI support and cannot be used at the same time due"
        "to incompatible mpi stub files. Either reconfigure deal.II, Trilinos,"
        and PETSc with MPI support, or disable one of the libraries.\n"
        )
```

tamiko: I think this makes the problem (and what one has to do) a bit clearer.

fmilicchio: @tamiko I've included your changes and removed the spurious spaces in the IF() call.

bangerth:
@tamiko ?

masterleinad:
@tamiko ping

tamiko:
@bangerth Ups. You can also "dismiss" the stale review in order to merge :-)

PR #8384, opened 2019-07-15
Complex periodic constraints

dangars:
This PR adds `std::complex make_periodicity_constraints()`


include/deal.II/dofs/dof_tools.h
```diff
@@ -1153,12 +1153,12 @@ namespace DoFTools
...
    * @author Matthias Maier, 2012 - 2015
    */
-  template <typename FaceIterator>
+  template <typename FaceIterator, typename number>
   void
   make_periodicity_constraints(
     const FaceIterator &                         face_1,
     const typename identity<FaceIterator>::type &face_2,
-    AffineConstraints<double> &                  constraints,
+    AffineConstraints<number> &                  constraints,
```
masterleinad: Don't you want to use `FullMatrix<number>` instead of `FullMatrix<double>` below?

masterleinad:
If I understand you correctly, you don't really want periodicity constraints with complex values, but just want to be able to initialize an `AffineConstraints<std::complex<double>>` object, correct?

Otherwise, it might make sense to also touch modify `matrix` in `DoFTools::make_hanging_node_constraints` and `matrix` and `offset` in `GridTools::collect_periodic_faces` to be complex-valued.

include/deal.II/dofs/dof_tools.h
tamiko: As this is an item on my (ever growing) todo list: I would very much like to see full complex number support in `make_periodicity_constraints` - including a phase shift over periodic boundaries. The hard part for this (and why this was not in my initial patch for `AffineConstraints` is the fact that we have to transform all the interpolation matrices and rotation matrices that are getting generated recursively also to complex valued entities.

@dangars Would you be open to try to tackle this as part of this PR?

dangars: @masterleinad That is right for the particular application I'm working on right now, this patch is enough because I just need to initialize an `AffineConstraints<std::complex<double>>` object.

@tamiko In the near future I want to compute the bandstructure of a phononic and phononic device. For that I need to impose a phase shift over periodic boundaries. I guess that is the application that you have in mind. Yes, let's do the phase shift support in this PR!

On my todo list there is a tutorial on how to calculate the bandgap using a periodic shift. I'll start to work on that after this PR is finished.

@tamiko I started to update the rotation matrix and I see all the dependencies. As @masterleinad suggested I'll update as well `matrix` in `DoFTools::make_hanging_node_constraints` and `matrix` and `offset` in `GridTools::collect_periodic_faces` to be complex-valued.

Can you add the WIP label? I'll let you know when it is ready for review :)

include/deal.II/fe/fe.h
```diff
@@ -1308,6 +1308,16 @@ class FiniteElement : public Subscriptor, public FiniteElementData<dim>
...
+
+
+  /**
+   * Same function as above with `std::complex<double>`.
+   */
+  virtual void
+  get_subface_interpolation_matrix(
+    const FiniteElement<dim, spacedim> &source,
+    const unsigned int                  subface,
+    FullMatrix<std::complex<double>> &  matrix) const;
```
dangars: In the file `dofs_tools_constrains.cc` line 1861, I need the `std::complex` version of
```c++
fe.get_subface_interpolation_matrix(fe, c, subface_interpolation);
```
But `get_subface_interpolation_matrix` is virtual, therefore I can not template it. What would you suggest? I did a proof of concept for the class `FE_Q_Base`, see below. Do you have a better solution?

include/deal.II/fe/fe_q_base.h
```diff
@@ -92,6 +92,27 @@ class FE_Q_Base : public FE_Poly<PolynomialType, dim, spacedim>
...
+private:
+  /**
+   * Template implementation of get_subface_interpolation_matrix().
+   */
+  template <typename number>
+  void
+  get_subface_interpolation_matrix_implemetation(
+    const FiniteElement<dim, spacedim> &source,
+    const unsigned int                  subface,
+    FullMatrix<number> &                matrix) const;
```
dangars: I did a private template implementation.

source/fe/fe_q_base.cc
```diff
@@ -605,13 +605,38 @@ FE_Q_Base<PolynomialType, dim, spacedim>::get_face_interpolation_matrix(
...
+template <class PolynomialType, int dim, int spacedim>
+void
+FE_Q_Base<PolynomialType, dim, spacedim>::get_subface_interpolation_matrix(
+  const FiniteElement<dim, spacedim> &x_source_fe,
+  const unsigned int                  subface,
+  FullMatrix<std::complex<double>> &  interpolation_matrix) const
+{
+  get_subface_interpolation_matrix_implemetation(x_source_fe,
+                                                 subface,
+                                                 interpolation_matrix);
```
dangars: `get_subface_interpolation_matrix()` calls the private template implementation. If this is good I can do it for the rest of the elements in deal.II. Otherwise, let me know if you have a better suggestion.

masterleinad:
> @masterleinad It seems that it is not necessary to template `DoFTools::make_hanging_node_constraints()` which does not have a double parameter. I happy to touch that function, but what should I template?

Oh, that was a typo. I meant `DoFTools::make_periodicity_constraints()`.

source/grid/grid_tools_dof_handlers.cc
```diff
@@ -2261,27 +2261,34 @@ namespace GridTools
...
     else
-      distance = point1;
+      {
+        if (distance.dimension >= 1)
+          distance(0) = point1(0);
+        if (distance.dimension >= 2)
+          distance(1) = point1(1);
+        if (distance.dimension == 3)
+          distance(2) = point1(2);
+      }
```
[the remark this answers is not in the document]
masterleinad: You are using copy assignment operators here, not constructors. :wink:
The compiler implicitly generates a copy constructor for an object of the same type, but not for one with a different number type. In general, it is not clear how to convert from one number type to another one. 

I would simply use a `for`-loop here.

source/dofs/dof_tools_constraints.cc
```diff
@@ -1848,8 +1848,8 @@ namespace DoFTools
 
           const unsigned int dofs_per_face =
             face_1->get_fe(face_1->nth_active_fe_index(0)).dofs_per_face;
-          FullMatrix<double> child_transformation(dofs_per_face, dofs_per_face);
-          FullMatrix<double> subface_interpolation(dofs_per_face,
+          FullMatrix<number> child_transformation(dofs_per_face, dofs_per_face);
+          FullMatrix<number> subface_interpolation(dofs_per_face,
```
masterleinad: Can't you leave this as `FullMatrix<double>` and drop your modifications to `get_subface_interpolation_matrix`?

source/grid/grid_tools_dof_handlers.cc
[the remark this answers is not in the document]
dangars: That is right :) I'm not using constructors.

The class `Tensor` defines a copy assignment operator for a different number type. It seems that you have to use `using` if you want to use that particular copy assignment operator (see below).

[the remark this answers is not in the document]
dangars: Now that I added the following code to `Point`, there is no need to use a for loop
```c++
using Tensor<1, dim, Number>::operator=;
```

source/dofs/dof_tools_constraints.cc
dangars: That is right!  `subface_interpolation` can be `FullMatrix<double> ` and `child_transformation` can be `FullMatrix<number>`. I had to add a few instantiations for `FullMatrix`, notably `FullMatrix<S1>::mmult<S2>`.

include/deal.II/base/point.h
```diff
@@ -195,6 +195,11 @@ class Point : public Tensor<1, dim, Number>
   DEAL_II_CUDA_HOST_DEV Number &
                         operator()(const unsigned int index);
 
+  /**
+   * Assignment operator inherited from Tensor.
+   */
+  using Tensor<1, dim, Number>::operator=;
```
dangars: This is necessary in order to use `operator=` of Tensor for a different number. The link for [operator=](https://www.dealii.org/current/doxygen/deal.II/classTensor.html#aa4aae50e15d4cc42b73c779b484366c7)

https://en.cppreference.com/w/cpp/language/using_declaration

dangars: I compiled the documentation with Doxygen and _as expected_ this statement does not appear in the documentation.

I wonder if it is more clear to overload `operator=` and then call the base class ` Tensor<1, dim, Number>::operator=` from `Point<dim,number>::operator=`?

Which option do you prefer: `using `or overloading `operator=`?

masterleinad: I would prefer to implement
```
template <typename Number2>
Pointer& operator(Pointer<dim, Number2> &other_point);
```

The `using` declaration potentially slices the object, see https://godbolt.org/z/7W6ekT
(even if `Point` doesn't have any non-inherited member variables). Please add a changelog entry anyway.

source/grid/grid_tools_dof_handlers.cc
```diff
@@ -2261,20 +2261,20 @@ namespace GridTools
...
-  orthogonal_equality(const Point<spacedim> &    point1,
-                      const Point<spacedim> &    point2,
-                      const int                  direction,
-                      const Tensor<1, spacedim> &offset,
-                      const FullMatrix<double> & matrix)
+  orthogonal_equality(const Point<spacedim> &            point1,
+                      const Point<spacedim> &            point2,
+                      const int                          direction,
+                      const Tensor<1, spacedim, number> &offset,
+                      const FullMatrix<number> &         matrix)
```
dangars: `point1` and `point2` are real, therefore the transformation between `point1` and `point2` has to be real.

masterleinad:
I agree that `offset` and `matrix` in `GridTools::collect_periodic_faces` should always be real. Sorry for requesting you to change that!

dangars:
No worries :) It is better that we realized this before merging. I had a copy of the old branch. I reverted the PR to the original patch.

I'm also going to work on the phase shift with complex numbers, but I think that it is better to do it in a different PR, in order to keep it simple. In the phase shift PR, probably I will add a new parameter for the phase shift.

This patch is ready for review!

With this patch it is possible to do periodic boundary conditions with complex numbers without phase shift.

masterleinad:
Let's see if @tamiko has more to say.

dangars:
I added the PR #8419 for the phase shift condition. PR #8419 builds on this PR. I think that it is easier to review if they are two separate PRs.

dangars:
Thanks @masterleinad :)

PR #8389, opened 2019-07-17
Add assignment operator between different MemorySpace

Rombur:
This PR allows the assignment of a `LA::d::Vector` of one `MemorySpace` to another one. This allows to simplify some code since we don't need to use a `ReadWriteVector` all the time to copy the solution from the device to the host.


masterleinad:
The indentation is off.

doc/news/changes/minor/20190717BrunoTurcksin
```diff
@@ -0,0 +1,4 @@
+New: Add assignment operator to LinearAlgebra::distributed::Vector() when the
+two vector have different MemorySpace.
```
kronbichler:
```suggestion
two vectors live in a different MemorySpace.
```

include/deal.II/lac/la_parallel_vector.templates.h
```diff
@@ -629,79 +631,43 @@ namespace LinearAlgebra
...
-    operator=(const Vector<Number2, MemorySpaceType> &c)
+    operator=(const Vector<Number, MemorySpaceType2> &c)
     {
-      Assert(c.partitioner.get() != nullptr, ExcNotInitialized());
-
       // we update ghost values whenever one of the input or output vector
       // already held ghost values or when we import data from a vector with
       // the same local range but different ghost layout
-      bool must_update_ghost_values = c.vector_is_ghosted;
+      assignment(c, c.vector_is_ghosted);
```
kronbichler: I don't think you can pass in `c.vector_is_ghosted` as an argument because we do not want to change the ghosted value of the other constant vector `c` (you can only change it because the variable is `mutable`). We should create a separate value. Ideally, we should simply return the `is_ghosted` information from the `assignment` function and resolve the boolean variable inside.

include/deal.II/lac/la_parallel_vector.templates.h
```diff
@@ -629,79 +631,43 @@ namespace LinearAlgebra
...
+    {
+      // we update ghost values whenever one of the input or output vector
+      // already held ghost values or when we import data from a vector with
+      // the same local range but different ghost layout
+      assignment(c, c.vector_is_ghosted);
 
-      if (must_update_ghost_values)
-        update_ghost_values();
-      else
-        zero_out_ghosts();
```
kronbichler: I don't think you want to remove these four lines.

Rombur:
@kronbichler  you are right. I wrote this patch a while ago and went back to it yesterday I added the bugs...

include/deal.II/lac/la_parallel_vector.templates.h
```diff
@@ -629,79 +631,43 @@ namespace LinearAlgebra
...
-        {
-          dealii::internal::VectorOperations::
-            functions<Number, Number2, MemorySpaceType>::copy(
-              thread_loop_partitioner, this_size, c.data, data);
-        }
+    template <typename Number, typename MemorySpaceType>
+    template <typename Number2>
+    inline Vector<Number, MemorySpaceType> &
+    Vector<Number, MemorySpaceType>::
+    operator=(const Vector<Number2, MemorySpaceType> &c)
```
masterleinad: Can't you just have a
```
operator=(const Vector<Number2, MemorySpaceType2> &c)
```
signature for all the combinations and `static_assert` that the `MemorySpace` is `Host` if we don't compile with `CUDA` support?

Rombur: No that doesn't compile. I need at least two functions `operator=(const Vector<Number, MemorySpaceType> &c)` (otherwise I can a bunch of error that the operator is deleted) and `operator=(const Vector<Number2, MemorySpaceType2> &c)`. I think it's clearer this way too since I do not allow `Number != Number2` and `MemorySpaceType != MemorySpaceType2` at the same time.

tjhei:
I am not sure this applies necessarily here, but I think we made a wrong design decision about using ``operator=`` in the past when we used to to mean ``update ghost values in ghosted vector from this other vector``. We should be cautious in adding non-obvious semantics to ``operator=``.  In this case, does ``a=b;`` imply ``a==b``?

Rombur:
@tjhei good point. After the assignment the two vectors are still different because for one vector the data is on the host and for the other the data is on the device. Would it be better to create a new function?

tjhei:
> Would it be better to create a new function?

I am not sure, I just wanted you to think about this question. :-)
I guess one could argue that after ``a=b;`` both contain the same data here.
What would you call this function, ``import(other)``?

kronbichler:
> What would you call this function, `import(other)`?

I would prefer the `import()` function indeed (and I am also annoyed by the ambiguous semantics we attached to `operator=`).

Rombur:
OK let me close this PR. I'll reopen another one once I have the `import()` function.

PR #8391, opened 2019-07-18
Add alternative test output for some roundoff-sensitive tests

kronbichler:
With AVX-512 and release configuration, we have a different roundoff behavior that shows off for single-precision inner products, see e.g. here for one of the tests:
https://cdash.43-1.org/testDetails.php?test=1681002&build=244
This PR adds the output generated on such a machine.


kronbichler:
I agree, I will change these three tests and the others that output the value of the inner product to check for a tolerance instead. This should give robust results and allow me to eliminate a lot of duplicate output files for variants.

kronbichler:
I have changed the code now and I check the correctness in the program by a tolerance and print "correct" in case we are fine, and "wrong" with the two results in case we are not. This should make these tests much more stable I hope. (I still have to see whether my tolerance is stable to give good results on all systems, but we will see when this is merged or even earlier by the CI test.)

PR #8393, opened 2019-07-18
Provide AVX-512 output for four additional tests

kronbichler:
Due to different roundoff effects with AVX-512 vectorization (or FMA), we get different test output on a few more tests. This fixes four additional simple cases where we need another output. Note that 3 of the 4 tests for which I added an AVX-512 variant already had two variants, so this is nothing unexpected, while the `data_out_base_vtu_04` tests did not but it is influenced by AVX, so I think it also make sense.

Together with #8392 and #8391, this should bring down the list of failures on my AVX-512 machine with `-O3` compilation,
https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=244
to the `step-51` tests where there seems to be a subtle race condition (no idea yet how to avoid it without disabling threads) and two more tests `numerics/project_q_hierarchical_02.release` and `grid/grid_out_in_vtk_01.debug` that run into an FPE somewhere and I won't fix now.


masterleinad:
The indentation is off.

kronbichler:
Sorry, I fixed that now.

PR #8107, opened 2019-05-12
use auto & cell instead of auto cell

tamiko:
Unify the range-based for loops over active cells as discussed in #8102

PR #8396, opened 2019-07-19
ranged-for documentation fix

rezarastak:
Since we seek to unify the style of range-based cell loops (#8107), the documentation should be fixed (especially `c++11.h`)

Related question: Has anyone measured the speed difference between
```c++
for (auto cell : triangulation.active_cell_iterators())
```
and
```c++
```
?

### DIFF
--- a/doc/doxygen/headers/c++11.h
+++ b/doc/doxygen/headers/c++11.h
@@ -40,7 +40,7 @@
  * @code
  *   Triangulation<dim> triangulation;
  *   ...
- *   for (auto cell : triangulation.active_cell_iterators())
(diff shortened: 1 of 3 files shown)


tjhei:
> Has anyone measured the speed difference

I do not expect a speed difference between const and non-const iteration, but I haven't checked.

bangerth:
The `const` will make no speed difference -- it's really just a way to let the compiler help find bugs.

The reference may or may not. I suspect that in most cases, the compiler will be able to optimize out all copies anyway for objects as small as cell iterators. But it never hurts to get in the habit to do things in a way that also make it more efficient in other cases.

bangerth:
Looks good, thank you!

PR #8400, opened 2019-07-20
Add new make_vectorized_array function

peterrum:
During using the features form PR #8348, I realized that it would be nice to have a new utility function:
```cpp
template <typename VectorizedArrayType>
VectorizedArrayType
make_vectorized_array(const typename VectorizedArrayType::value_type &u);
```

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -522,6 +522,29 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
 
 
 /**
+ * Create a vectorized array of given type and broadcast the scalar value
+ * to all array elements.
+ *
+ *  @relatesalso VectorizedArray
+ */
+template <typename VectorizedArrayType>
+inline DEAL_II_ALWAYS_INLINE VectorizedArrayType
+                             make_vectorized_array(const typename VectorizedArrayType::value_type &u)
+{
+  static_assert(
+    std::is_same<VectorizedArrayType,
+                 VectorizedArray<typename VectorizedArrayType::value_type,
+                                 VectorizedArrayType::n_array_elements>>::value,
+    "VectorizedArrayType is not a VectorizedArray.");
+
+  VectorizedArrayType result;
+  result = u;
+  return result;
+}
+
+
+
+/**
  * This method loads VectorizedArray::n_array_elements data streams from the
  * given array @p in. The offsets to the input array are given by the array @p
  * offsets. From each stream, n_entries are read. The data is then transposed
--- a/tests/base/vectorization_make_vectorized_array_01.cc
+++ b/tests/base/vectorization_make_vectorized_array_01.cc
@@ -1,0 +1,107 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2015 - 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// test make_vectorized_array function for all possible value_types and vector
+// lengths
+
+#include <deal.II/base/vectorization.h>
+
+#include <limits>
+
+#include "../tests.h"
+
+
+template <typename VectorizedArrayType>
+void
+do_test(const VectorizedArrayType                      array,
+        const typename VectorizedArrayType::value_type number)
+{
+  deallog << "  test " << VectorizedArrayType::n_array_elements
+          << " array elements" << std::endl;
+  for (unsigned int i = 0; i < VectorizedArrayType::n_array_elements; i++)
+    if (array[i] != number)
+      deallog << "  problem in element " << i << std::endl;
+}
+
+
+template <typename Number>
+struct Tester
+{
+  static void
+  test()
+  {}
+};
+
+template <>
+struct Tester<double>
+{
+  static void
+  test()
+  {
+    do_test(make_vectorized_array<double>(2.0), 2.0);
+
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 3 && defined(__AVX512F__)
+    do_test(make_vectorized_array<VectorizedArray<double, 8>>(2.0), 2.0);
+#endif
+
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 2 && defined(__AVX__)
+    do_test(make_vectorized_array<VectorizedArray<double, 4>>(2.0), 2.0);
+#endif
+
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 1 && defined(__SSE2__)
+    do_test(make_vectorized_array<VectorizedArray<double, 2>>(2.0), 2.0);
+#endif
+
+    do_test(make_vectorized_array<VectorizedArray<double, 1>>(2.0), 2.0);
+  }
+};
+
+template <>
+struct Tester<float>
+{
+  static void
+  test()
+  {
+    do_test(make_vectorized_array<float>(2.0), 2.0);
+
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 3 && defined(__AVX512F__)
+    do_test(make_vectorized_array<VectorizedArray<float, 16>>(2.0), 2.0);
+#endif
+
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 2 && defined(__AVX__)
+    do_test(make_vectorized_array<VectorizedArray<float, 8>>(2.0), 2.0);
+#endif
+
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 1 && defined(__SSE2__)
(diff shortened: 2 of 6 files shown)


kronbichler:
Can you please add a test?

masterleinad:
I guess we could discuss if we want to have a constructor instead.

tjhei:
An alternative would be something like ``void VectorizedArray::load_single_value(double value)`` (or ``load_all``, ...). It would mirror the SIMD ``set1_*`` etc..
I kind of like to be explicit with these operations as reasoning about vectorized code is hard enough...

include/deal.II/base/vectorization.h
```diff
@@ -521,6 +521,23 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
...
 
+/**
+ * Create a vectorized array that sets all entries in the array to the given
+ * scalar.
+ *
+ *  @relatesalso VectorizedArray
+ */
+template <typename VectorizedArrayType>
+inline DEAL_II_ALWAYS_INLINE VectorizedArrayType
+                             make_vectorized_array(const typename VectorizedArrayType::value_type &u)
```
bangerth: This requires you to write "vectorized" and "array" twice, once in the function name and once in the template argument (in the form of the class name) you need to specify explicitly here. I think you might want to use this instead:
```
VectorizedArray<Number,width>
make_vectorized_array (const Number &u) {...}
```
instead.

peterrum:
@kronbichler: I have added a test.

peterrum:
@masterleinad, @tjhei, @bangerth: Thanks for your feedback!

Let's ask @kronbichler about the copy constructor and alternative names. I like the idea with the copy constructor. But I guess there have been reasons against it, e.g. clear separation between `Number` and `VectorizedArray<Number>`.

include/deal.II/base/vectorization.h
peterrum: I see your point @bangerth. You are thinking about such a case: you have given the value type `Number` and the vector length `width`:
```cpp
auto array = make_vectorized_array<VectorizedArray<Number, width>>(value);
```
indeed this case can be simply written (as used to be) as:
```cpp
auto array = make_vectorized_array<Number, width>(value);
```

However, I am tackling in this PR another case: the type `VectorizedArrayType` is given (instead of `Number` and `width`). If I would use the old `make_vectorized_array<Number, width>`,  I would have to write:
```cpp
auto array = make_vectorized_array<VectorizedArrayType::value_type, 
                                   VectorizedArrayType::n_array_elements>(value);
```
If I would use the new proposed function, I could simply write:
```cpp
auto array = make_vectorized_array<VectorizedArrayType>(value);
```
This makes the life of the user somewhat easier, since he/she does not have to think of the vector length explicitly.

kronbichler:
> Let's ask @kronbichler about the copy constructor and alternative names. I like the idea with the copy constructor. But I guess there have been reasons against it, e.g. clear separation between `Number` and `VectorizedArray<Number>`.

First, the name copy constructor is misleading because it does not actually copy but creates a genuinely new object. The biggest reason to introduce a constructor `explicit VectorizedArray<Number>(const Number)` is for compatibility in a large variety of classes, most prominently `Point` and `Tensor` where we have some kind of issue requiring a new workaround at least once per year. We would really like to write
```c++
VectorizedArray<Number> a = Number(0.);
```
for compatibility with `double` or also the automatic differentiation types. The reason why we hesitated is the POD property and other simple setups like the default constructor not doing anything. By only introducing the conversion constructor - which does a broadcast into all SIMD lanes, we would otherwise keep the expected `std::is_standard_layout` and `std::is_trivial` that is necessary for `memcpy` and friends. So I am definitely in favor of introducing that constructor, as also mentioned in #8064. In any case, this does not eliminate the need for the `make_vectorized_array` functionality.

Regarding the member function `void VectorizedArray::load_single_value(double value)` proposed by @tjhei: I like having such a member function, but I would suggest to call it `broadcast_from_scalar` or something like that because I think the broadcast name exactly matches what happens in the `_mm_set1_pd` functions.

peterrum:
If I understand it correctly, we want these 4 functions/methods:
```cpp
// 1
VectorizedArray<Number,width>
make_vectorized_array (const Number u);

// 2
template <typename VectorizedArrayType>
VectorizedArrayType
make_vectorized_array (const VectorizedArrayType::value_type u);

class VectorizedArray
{
  // 3
  explicit VectorizedArray<Number>(const Number);

  // 4
  void broadcast_from_scalar(const Number);
}
```
We already have the first one. The second is part of this PR. I can add the third and forth in a separate PR.

bangerth:
On 7/21/19 12:07 AM, peterrum wrote:
> However, I am tackling in this PR another case: the type |VectorizedArrayType| 
> is given (instead of |Number| and |width|). If I would use the old 
> |make_vectorized_array<Number, width>|, I would have to write:
> 
> auto  array = make_vectorized_array<VectorizedArrayType::value_type,
>                                     VectorizedArrayType::n_array_elements>(value);
> 
> If I would use the new proposed function, I could simply write:
> 
> auto  array = make_vectorized_array<VectorizedArrayType>(value);
> 
> This makes the life of the user somewhat easier, since he/she does not have to 
> think of the vector length explicitly.

I still think that being explicit is useful. One of the issues I see is that 
with the signature of the function you propose, it's possible to write code 
such as

```
   auto array = make_vectorized_array<std::vector<double>> (value);

```

which is definitely not what you want. You make an implicit assumption in your 
function for what kinds of classes are valid template arguments, but you are 
not expressing this assumption in the code itself.

peterrum:
@bangerth: is a static assert an option?
```cpp
template <typename VectorizedArrayType>
inline DEAL_II_ALWAYS_INLINE VectorizedArrayType
                             make_vectorized_array(const typename VectorizedArrayType::value_type &u)
{
  static_assert(
    std::is_same<VectorizedArrayType, VectorizedArray<VectorizedArrayType::type_name, VectorizedArrayType::n_array_elements>>::value,
    "VectorizedArrayType is not a VectorizedArray.");

  VectorizedArrayType result;
  result = u;
  return result;
}
```

It would be really nice to have a solution, in which the user does not have not care with the vector length at too many instances (maybe only at one place in a whole application)...

bangerth:
Yes, I think the static_assert would be a nice approach to the issue.

peterrum:
Is there already an issue regarding creating the two methods, which we discussed above, around?
```cpp
class VectorizedArray
{
  // 3
  explicit VectorizedArray<Number>(const Number);

  // 4
  void broadcast_from_scalar(const Number);
}
```
If not, I can create one and assign it to myself.

kronbichler:
There is no such issue yet, feel free to open one.

tests/base/vectorization_make_vectorized_array_01.cc
```diff
@@ -0,0 +1,107 @@
...
+  }
+};
+
+template <>
+struct Tester<float>
+{
+  static void
+  test()
+  {
+    do_test(make_vectorized_array<float>(2.0), 2.0);
```
bangerth: will this not now fail your static_assert?

include/deal.II/base/vectorization.h
```diff
@@ -521,6 +521,29 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
...
+{
+  static_assert(
+    std::is_same<VectorizedArrayType,
+                 VectorizedArray<typename VectorizedArrayType::value_type,
+                                 VectorizedArrayType::n_array_elements>>::value,
+    "VectorizedArrayType is not a VectorizedArray.");
```
bangerth: Can you reflect this assertion in the documentation of the function?

peterrum: I will update the documentation.

tests/base/vectorization_make_vectorized_array_01.cc
peterrum: Why should it? There is still the old `make_vectorized_array<Number>`:
```cpp
template <
  typename Number,
  int width = internal::VectorizedArrayWidthSpecifier<Number>::max_width>
inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
                             make_vectorized_array(const Number &u);
```
Or am missing something?

bangerth: Hm. How does the compiler distinguish between the two functions if you write `make_vectorized_array<float>(...)`?

kronbichler: Does this probably get resolved due to SFINAE-like handling because one runs into the static assert whereas the other does compile fine?

masterleinad: It's not because of the `static_assert`, but because you of using `VectorizedArrayType::value_type` in the signature, see https://godbolt.org/z/92xiyE.
This is a plain SFINAE case.

bangerth: Yes. Selecting a function is done purely by looking at its signature, not its implementation. But the `VectorizedArrayType::value_type` makes a difference. The compiler would complain about an ambiguity, though, if you said
```
  make_vectorized_array<std::vector<double>> (...);
```
I suspect we never want to create such a vectorized array, so that's ok. Otherwise, we could play with `std::enable_if`.

kronbichler: I guess we can indeed live with the ambiguity for `VectorizedArray<ComplicatedType>` where `ComplicatedType` defines `value_type` such that the lookup would succeed. Since `VectorizedArray` is really only non-trivial for `double` and `float` types at this point, we should be pretty safe. As you say, we could fix it by `enable_if` in case we feel for it.

masterleinad:
@bangerth needs to approve we can merge.

PR #8401, opened 2019-07-21
Fixing the linker error of StandardTensors with clang 4.0.1

rezarastak:
I'm relatively confident this fixes the issue in [this](https://cdash.43-1.org/buildSummary.php?buildid=338), although I don't have clang 4.0.1 on my machine.

### DIFF
--- a/source/physics/elasticity/standard_tensors.cc
+++ b/source/physics/elasticity/standard_tensors.cc
@@ -20,32 +20,46 @@
 DEAL_II_NAMESPACE_OPEN
 
 #ifndef DOXYGEN
+
+template <int dim>
+DEAL_II_CONSTEXPR const SymmetricTensor<2, dim>
+                        Physics::Elasticity::StandardTensors<dim>::I
 #  ifndef DEAL_II_HAVE_CXX14_CONSTEXPR_CAN_CALL_NONCONSTEXPR
-
-template <int dim>
-const SymmetricTensor<2, dim>
-  Physics::Elasticity::StandardTensors<dim>::I = unit_symmetric_tensor<dim>();
-
-
-
-template <int dim>
-const SymmetricTensor<4, dim>
-  Physics::Elasticity::StandardTensors<dim>::S = identity_tensor<dim>();
+  = unit_symmetric_tensor<dim>()
+#  endif
+  ;
 
 
 
 template <int dim>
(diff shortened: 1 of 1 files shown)


masterleinad:
Yes, I tested that this works. Thanks!

/rebuild

source/physics/elasticity/standard_tensors.cc
```diff
@@ -20,32 +20,46 @@
...
-#  ifndef DEAL_II_HAVE_CXX14_CONSTEXPR_CAN_CALL_NONCONSTEXPR
 
 template <int dim>
-const SymmetricTensor<2, dim>
-  Physics::Elasticity::StandardTensors<dim>::I = unit_symmetric_tensor<dim>();
+DEAL_II_CONSTEXPR const SymmetricTensor<2, dim>
+                        Physics::Elasticity::StandardTensors<dim>::I
+#  ifndef DEAL_II_HAVE_CXX14_CONSTEXPR_CAN_CALL_NONCONSTEXPR
+  = unit_symmetric_tensor<dim>()
+#  endif
```
bangerth: Where does the initialization of the variable end up happening if the `#if` is not satisfied? In the `.h` file?

If so, you could just put the whole declaration into `#if` -- without the initialization, all you're doing here is to re-declare a variable you had previously already declared. That's not wrong, but unnecessary.

masterleinad: That is what the code before this pull request was doing.

Before C++17 we need the additional declaration for `static constexpr` member variables when it is ODR-used (https://en.cppreference.com/w/cpp/language/definition). That's why `clang-4` (and lower versions) were complaining.

rezarastak: As @masterleinad mentioned it is necessary to redeclare these variables for some compilers.

bangerth: Ah, how bizarre. If these were member variables, I would understand. But it doesn't make any sense for namespace level global variables.

But it's not wrong either...

masterleinad: In fact, `StandardTensors` is a class.

bangerth: Oh, I missed that part. Yes, then this makes perfect sense.

PR #8403, opened 2019-07-22
Enable MappingFEField on multigrid levels

kronbichler:
This commit enables `MappingFEField` on the multigrid levels of a triangulation.

The vector can be generated either by `MGTransferMatrixFree::interpolate_to_mg` (see the `_04` and `_05` tests) or filled manually on all levels (see the `_06` test). This works both in serial and in parallel.

### DIFF
--- a/include/deal.II/fe/mapping_fe_field.h
+++ b/include/deal.II/fe/mapping_fe_field.h
@@ -19,6 +19,7 @@
 
 #include <deal.II/base/config.h>
 
+#include <deal.II/base/mg_level_object.h>
 #include <deal.II/base/thread_management.h>
 
 #include <deal.II/dofs/dof_handler.h>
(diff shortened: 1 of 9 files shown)


include/deal.II/fe/mapping_fe_field.h
```diff
@@ -118,6 +119,31 @@ class MappingFEField : public Mapping<dim, spacedim>
                  const VectorType &    euler_vector,
                  const ComponentMask & mask = ComponentMask());
 
+  /**
+   * Constructor taking vectors on the multigrid levels rather than the active
+   * cells only. The vector of vectors is expected to have as many entries as
+   * there are levels in the triangulation and provide valid data on each
```
tjhei: what is n_levels != n_global_levels? What is the correct size?

include/deal.II/fe/mapping_fe_field.h
```diff
@@ -118,6 +119,31 @@ class MappingFEField : public Mapping<dim, spacedim>
...
+   * constructor need to be provided.
+   */
+  MappingFEField(const DoFHandlerType &         euler_dof_handler,
+                 const std::vector<VectorType> &euler_vector,
+                 const ComponentMask &          mask = ComponentMask());
+
+  /**
+   * Constructor taking vectors on the multigrid levels rather than the active
+   * cells only, variant with MGLevelObject instead of std::vector. The vector
+   * of vectors is expected to have as many entries as there are levels in the
```
tjhei: this needs to be updated as it is no "vector of vectors"

source/fe/mapping_fe_field.cc
```diff
@@ -262,10 +263,92 @@ MappingFEField<dim, spacedim, VectorType, DoFHandlerType>::MappingFEField(
...
+        fe_to_real[i] = size++;
+    }
+  AssertDimension(size, spacedim);
+
+  Assert(euler_dof_handler.has_level_dofs(),
+         ExcMessage("The underlying did not call distribute_mg_dofs(). "
+                    "In this case, the construction via level vectors does not "
+                    "make sense."));
+  AssertDimension(euler_vector.size(),
+                  euler_dof_handler.get_triangulation().n_global_levels());
```
tjhei: aha, n_global_levels it is. :-)

include/deal.II/fe/mapping_fe_field.h
tjhei: or replace the whole documentation block with "like above but with an MGLevelObject"

kronbichler:
@tjhei fixed!

source/fe/mapping_fe_field.cc
```diff
@@ -262,10 +263,92 @@ MappingFEField<dim, spacedim, VectorType, DoFHandlerType>::MappingFEField(
...
+      if (fe_mask[i])
+        fe_to_real[i] = size++;
+    }
+  AssertDimension(size, spacedim);
+
+  Assert(euler_dof_handler.has_level_dofs(),
+         ExcMessage("The underlying did not call distribute_mg_dofs(). "
+                    "In this case, the construction via level vectors does not "
+                    "make sense."));
+  AssertDimension(euler_vector.max_level() + 1 - euler_vector.min_level(),
```
tjhei: do you also want to require that min_level()==0? Otherwise it would make sense to only check ``max_level()+1==n_global`` and allow the user not to hand in stuff for lvl<min_lvl?

kronbichler: It is right, essentially I could be happy with `min_level>0` as well (my changes make more sense if you, for some reason, have a shift in the levels in the `MGLevelVector`). It is up to the user to give us the right level vectors that he/she later intends to use. Let me change the code to what you suggest.

kronbichler:
@masterleinad fixed!

PR #8405, opened 2019-07-23
Fix a bug in LA::d::Vector::compress on the host

Rombur:
`compress()` would fail if it was called in a CUDA file, for a `LA::d::Vector<Number, MemorySpace::Host>` and `deal.II/la/la_parallel_vector.templates.h` was included. This is because we entirely relied on preprocessor macro to switch between the `Host` and the `CUDA` path. This means that in a CUDA file, the `CUDA` path was always chosen even for a `Host` vector.


doc/news/changes/minor/20190723BrunoTurcksin
```diff
@@ -0,0 +1,6 @@
+Fixed: When compiling with CUDA support, the call to
+LinearAlgebra::distributed::Vector::compress() would failed if the vector's
```
masterleinad:
```suggestion
LinearAlgebra::distributed::Vector::compress() would have failed if the vector's
```

doc/news/changes/minor/20190723BrunoTurcksin
```diff
@@ -0,0 +1,6 @@
...
+MemorySpace was Host, the header deal.II/lac/la_parallel_vector.templates.h was
+included, and the call was done in .cu file.
```
masterleinad:
```suggestion
included, and the call was done in a .cu file.
```

doc/news/changes/minor/20190723BrunoTurcksin
```diff
@@ -0,0 +1,6 @@
+Fixed: When compiling with CUDA support, the call to
+LinearAlgebra::distributed::Vector::compress() would failed if the vector's
+MemorySpace was Host, the header deal.II/lac/la_parallel_vector.templates.h was
```
masterleinad:
```suggestion
memory space was MemorySpace::Host, the header deal.II/lac/la_parallel_vector.templates.h was
```
So doxygen can create more links.

tests/cuda/parallel_vector_26.cu
```diff
@@ -0,0 +1,82 @@
...
+// ---------------------------------------------------------------------
+
+
+// Check for a bug where compress cannot be called for a LA::d::Vector<Host> in
+// a CUDA file
+
+#include <deal.II/base/index_set.h>
+#include <deal.II/base/utilities.h>
+
+#include <deal.II/lac/la_parallel_vector.h>
```
masterleinad: You should be able to drop this header. Otherwise, we are doing something wrong.
```suggestion
```

Rombur:
Fixed

PR #8407, opened 2019-07-23
Point assignment operator

dangars:
Following up on the comment https://github.com/dealii/dealii/pull/8384#pullrequestreview-265055555

I added a point assignment operator between two points with different underlying type.


include/deal.II/base/point.h
```diff
@@ -195,6 +195,17 @@ class Point : public Tensor<1, dim, Number>
...
 
+  /**
+   * Assignment operator from Tensor<1, dim, Number> with different underlying
+   * scalar type. This obviously requires that the @p OtherNumber type is
+   * convertible to @p Number.
+   *
+   * @note This function can also be used in CUDA device code.
+   */
+  template <typename OtherNumber>
+  DEAL_II_CUDA_HOST_DEV Point<dim, Number> &
```
masterleinad:
```suggestion
Point<dim, Number> &
```
Let's be cautious here. Assigning between different Number types in `CUDA` code doesn't turn out to work well and leads to hard to understand errors.

masterleinad:
Thanks! You also need to run the indentation script.

dangars:
Now I think it should be ok :)

masterleinad:
Thanks!

include/deal.II/base/point.h
```diff
@@ -195,6 +195,17 @@ class Point : public Tensor<1, dim, Number>
   DEAL_II_CUDA_HOST_DEV Number &
                         operator()(const unsigned int index);
 
+  /**
+   * Assignment operator from Tensor<1, dim, Number> with different underlying
+   * scalar type. This obviously requires that the @p OtherNumber type is
+   * convertible to @p Number.
+   *
```
masterleinad:
```suggestion
```

include/deal.II/base/point.h
```diff
@@ -195,6 +195,17 @@ class Point : public Tensor<1, dim, Number>
...
+   * @note This function can also be used in CUDA device code.
```
masterleinad: Ah, I missed that there is still this comment. Would you removing it?
```suggestion
```

dangars: Sorry, I forgot to remove that comment. Now it is done!

PR #8414, opened 2019-07-25
Minor clean-ups.

bangerth:
Follow-up to #8352.

@GrahamBenHarper -- FYI.

### DIFF
--- a/include/deal.II/fe/fe_poly_tensor.h
+++ b/include/deal.II/fe/fe_poly_tensor.h
@@ -221,13 +221,13 @@ protected:
    * mapping and false when the finite element uses multiple mappings.
    */
   bool
-  single_mapping() const;
+  single_mapping_type() const;
 
   /**
    * Returns MappingType @p i for the finite element.
    */
   MappingType
-  get_mapping_type(unsigned int i) const;
+  get_mapping_type(const unsigned int i) const;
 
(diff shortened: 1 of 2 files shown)


include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -221,13 +221,13 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
    * mapping and false when the finite element uses multiple mappings.
    */
   bool
-  single_mapping() const;
+  single_mapping_type() const;
```
masterleinad: Hmm... Wouldn't `use_single_mapping` be better? It seems a bit strange that a type is either true or false.

bangerth: It doesn't refer to a type. It refers to the `mapping_type` variable, which could either be a single value or a vector of length greater than 1. 

I think it would have been nicer if we had called this variable `mapping_kind`, but that's an unrelated question.

bangerth: @GrahamBenHarper -- would you be willing to maybe do this? I.e., rename the `mapping_type` variable to `mapping_kind` and similarly for the functions you added? These are all internal, so there shouldn't be a problem with compatibility. This would make for a nice small pull request, I imagine.

Issue #7988, opened 2019-05-01
Reading in a 4M cell coarse grid into parallel::distributed::Triangulation is slow


tamiko:
@ragusa and @loganharbour reported an interesting issue with the `parallel::distributed::Triangulation`.

Namely reading in [1] a large coarse grid with 4M cells takes _significantly longer_ (in wall time and memory consumption) than with the serial `Triangulation`---all happening just on a single core. We are aware of the fact that one should ideally not read in such a large coarse grid. But the increase in computational time is still a bit surprising (we haven't found the time to investigate much further than this).

I will attach a minimal POC (creating a 4M cell triangulation and copying it over to a distributed one) a bit later today.

@tjhei @kronbichler *ping*

[1] This also happens when reading into a serial triangulation first and then copying over to a distributed one.

bangerth:
For a self-reproducing testcase, just start with a mesh with a single cell, refine it a bunch of times, do `GridGenerator::flatten`, and copy the result of this into the parallel triangulation.

bangerth:
@cburstedde -- FYI

When we first wrote the p4est interfaces, we did hit on bottlenecks in p4est that had quadratic behavior in the number of coarse cells. Carsten et al. fixed these, and consequently we now have tests in the testsuite that check that things work sufficiently quickly with 300,000 cells in the coarse mesh. (That's one of the stored external meshes we have in the testsuite.) But we all know that if you increase the size of a problem by a factor of ten, you'll see another bottleneck :-)

kronbichler:
I would say that a factor of 2-3 in increased time is expected because we do certain operations several times plus have p4est do something similar. The interesting question is whether there is actually an increase in complexity when moving to larger grids, which would then be worth investigating.

I think the largest meshes I've worked with are around 1M-3M coarse cells - but these are painful because you cannot really run them in parallel when the mesh on each MPI rank takes 5G or so for the mesh only. That said, together with @peterrum we are working on a concept similar to #3956 (which will be combined with or replace that PR depending on the wishes of @tjhei).

cburstedde:
I agree. If the run time becomes nonlinear in the number of trees, we may want to look at it.

tjhei:
> That said, together with @peterrum we are working on a concept similar to #3956 (which will be combined with or replace that PR depending on the wishes of @tjhei).

[...]

PR #8418, opened 2019-07-26
Add new fully distributed triangulation.

peterrum:
This PR introduces a new distributed parallel triangualation framework with the name 
`parallel::fullydistributed::Triangulation` (short: `PFT`).  

**Note:** This PR is accompanied by a seperate [repo](https://github.com/peterrum/dealii-pft). It gives more detailed information on the concepts and provides 8 tutorials clarifying the usage of the `PFT` framework.

Key features of the new framework are:

- the multi-level graph-based partitioning exploiting hardware properties
  regarding the NUMA-domain, the node and the island sizes (`dealii::GridTools::shared_partition_triangulation()`) and
- a distributed coarse grid, which makes `PFT` in a sense fully distributed. 

In this sense the PR is closely related to PR #3956, which presented `parallel::split::Triangulation`. 

The class `parallel::distributed::Triangulation` stores on every process a copy
the coarse grid, which might be too expensive from a memory point of view if a non-negligible 
number of coarse cells (>10,000) is considered: The need to partition the coarse grid is indeed essential and has been discussed in issue #7988.

An (incomplete) list of requirements - beside the need to partition the coarse grid - is:
- [x] extract from `dealii::Triangulation` (serial triangulation) 
- [x] extract from `dealii::parallel::distributed::Triangulation` (parallel triangulation) 
- [x] static mesh
- [x] hanging nodes
- [x] geometric multigrid
- [x] periodicity
- [x] 2D/3D and 1D (also in parallel!)
- [x] I/O from/to `*.pft`-files
- [ ] adaptive mesh refinement (AMR)

The ticks indicate that the given features have been implemented to the best of the author's knowledge. The features were tested for a small set of triangulations (hypercube, subdivided hypercube, hyper-L, and hyper-shell with a large number of different of refinement, subdivision and process counts). 

AMR has been not tackled yet. AMR-capabilities should be added to `PFT` by the generalization of the  oracle principle employed in `PDT`.

To be able to construct a fully partitioned triangulation that fulfills the requirements, we need following ingredients:

1. a locally relevant coarse-grid triangulation (vertices, cells with material and manifold IDs, boundary IDs; including a layer of ghost cells)
2. a mapping of the locally relevant coarse-grid triangulation into the global coarse-grid triangulation
3. information about which cell should be refined as well as information regarding the subdomain ID and the level subdomain ID of each cell.

The ingredients listed above are bundled in `parallel::fullydistributed::ConstructionData`. The user has to fill this data structure - in a pre-processing step - before actually creating the triangulation. I provide a front-end to serialize and deserialize triangulations as well as to convert serial and distributed meshes (collected in `parallel::fullydistributed::util`). E.g.:
```cpp
// create triangulation object
parallel::fullydistributed::Triangulation<dim> tria_pft(comm);

// extract relevant information form serial triangulation
auto construction_data =
  parallel::fullydistributed::Utilities::copy_from_triangulation(basetria, tria_pft);

// actually create triangulation
tria_pft.reinit(construction_data);
```

For tackling really huge coarse grids (4M - as described in #7988), it is possible to create a serial triangulation on selected master processes (e.g. 1 master process per compute node), partition the mesh, and finally distribute the mesh information.

**Note:** The reason for the choice of the name `parallel::fullydistributed::Triangulation` is the need to distinguish this class from `parallel::distributed::Triangulation` and to emphasize that also the coarse mesh is partitioned in this case. Please feel free to suggest alternative names!

**Note:** This PR is not ready to be merged. I am still adding comments and 
refactoring the code to increase the reusability of the components. 
deal.II-tests still have to be added. Please consider this PR as an opportunity 
to discuss the concepts (also in person at the deal.II workshop in a week). 

@tjhei, @tamiko, @bangerth, @kronbichler *ping*

### DIFF
--- a/include/deal.II/distributed/tria.h
+++ b/include/deal.II/distributed/tria.h
@@ -1424,33 +1424,33 @@ namespace parallel
 
 #endif
 
-struct Part_
-{
-  Part_()
-  {}
-
-  Part_(CellId::binary_type index,
-        unsigned int        subdomain_id,
-        unsigned int        level_subdomain_id)
-    : index(index)
-    , subdomain_id(subdomain_id)
-    , level_subdomain_id(level_subdomain_id){};
-
-  CellId::binary_type index;
-  unsigned int        subdomain_id;
-  unsigned int        level_subdomain_id;
-};
-
-class Part
-{
-public:
-  std::vector<Part_> cells;
-};
 
 namespace parallel
 {
   namespace fullydistributed
   {
+    struct Part_
+    {
+      Part_()
+      {}
+
+      Part_(CellId::binary_type index,
+            unsigned int        subdomain_id,
+            unsigned int        level_subdomain_id)
+        : index(index)
+        , subdomain_id(subdomain_id)
+        , level_subdomain_id(level_subdomain_id){};
+
+      CellId::binary_type index;
+      unsigned int        subdomain_id;
+      unsigned int        level_subdomain_id;
+    };
+
+    class Part
+    {
+    public:
+      std::vector<Part_> cells;
+    };
     template <int dim, int spacedim>
     struct ConstructionData
     {
--- a/include/deal.II/distributed/tria.h
+++ b/include/deal.II/distributed/tria.h
@@ -22,6 +22,7 @@
 #include <deal.II/base/smartpointer.h>
 #include <deal.II/base/subscriptor.h>
 #include <deal.II/base/template_constraints.h>
+#include <deal.II/base/timer.h>
 
 #include <deal.II/distributed/p4est_wrappers.h>
 #include <deal.II/distributed/tria_base.h>
@@ -64,8 +65,10 @@ namespace internal
     {
       template <typename>
       class ParallelDistributed;
-    }
-  } // namespace DoFHandlerImplementation
+      template <typename>
+      class ParallelFullyDistributed;
+    } // namespace Policy
+  }   // namespace DoFHandlerImplementation
 } // namespace internal
 
 namespace FETools
@@ -1388,7 +1391,6 @@ namespace parallel
   } // namespace distributed
 } // namespace parallel
 
-
 #else // DEAL_II_WITH_P4EST
 
 namespace parallel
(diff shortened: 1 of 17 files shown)


source/grid/tria.cc
```diff
@@ -1825,9 +1855,10 @@ namespace internal
           // nodes (for example a
           // ring of cells -- no end
           // points at all)
-          AssertThrow(((spacedim == 1) && (boundary_nodes == 2)) ||
-                        (spacedim > 1),
-                      ExcMessage("The Triangulation has too many end points"));
+          // AssertThrow(((spacedim == 1) && (boundary_nodes == 2)) ||
+          //              (spacedim > 1),
+          //            ExcMessage("The Triangulation has too many end
+          //            points"));
```
kronbichler: Can you explain why it is removed? I suppose this will get enabled again at some point?

peterrum:
@kronbichler Thanks for the comments! I will make the modifications. Regarding the asserts: I had to disable some because many asserts assume that the periodicity is applied on the coarse grid, which is not valid any more. I will have a look at these asserts again and adjust them for the new use case.

source/grid/tria.cc
```diff
@@ -968,6 +969,35 @@ namespace
     Assert(periodic_face_map.count(cell_face_1) == 0, ExcInternalError());
     periodic_face_map.insert(periodic_faces);
 
+    if (dim == 1)
+      {
+        if (cell_1->has_children())
```
bangerth: I think this part of the patch is not specific to your work, right? Could you submit this as a separate PR?

(If so, it's really difficult to read code of the form
```
  if (dim==1)
    {
      ...
      return;
    }

  something else
```
because it's so easy to miss the `return` statement. It's substantially easier to read it if you write
```
  if (dim==1)
  {
    // 1d is special because X, Y, and Z. So treat this case separately
    ...
  }
  else  // dim > 1
  {
    ...
  }
```

source/grid/tria.cc
```diff
@@ -950,8 +950,9 @@ namespace
     Assert(face_1->at_boundary() && face_2->at_boundary(),
            ExcMessage("Periodic faces must be on the boundary"));
 
+    if (std::abs(cell_1->level() - cell_2->level()) > 1)
+      return;
```
bangerth: This case can only happen in 1d. It might be worth asserting this. But regardless, can you explain why in that case there is nothing to do in the function?

source/grid/tria.cc
peterrum: @bangerth you are right! I'll create a separate PR. Is there any test I could use as template for a new test?

bangerth: If I understand correctly, you're here adding periodicity in 1d. Did that just not work at all before?

There are numerous tests for periodicity in `bits/`. It may be simple enough to adapt one of them to run in 1d.

peterrum: @bangerth: thanks! I have created a new PR #8423.

source/grid/tria.cc
peterrum: See also #8423. I have moved the assert into the else-statement of the above if-else construct!

I must I do not have an explanation for the different behavior of 1D and 2/3D. Do you have an answer?

include/deal.II/distributed/tria.h
```diff
@@ -1422,6 +1424,162 @@ namespace parallel
...
+
+      // information
+      std::map<int, int> coarse_lid_to_gid;
+
+      // information describing how to constuct the levels
+      std::vector<Part> parts;
+    };
+
+    template <int dim, int spacedim = dim>
+    class Triangulation : public dealii::parallel::Triangulation<dim, spacedim>
```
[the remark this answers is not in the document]
peterrum: OK! I'll postpone this.

source/grid/grid_tools_dof_handlers.cc
```diff
@@ -2063,8 +2063,8 @@ namespace GridTools
     Assert(0 <= direction && direction < space_dim,
            ExcIndexRange(direction, 0, space_dim));
 
-    Assert(pairs1.size() == pairs2.size(),
-           ExcMessage("Unmatched faces on periodic boundaries"));
+    // Assert(pairs1.size() == pairs2.size(),
+    //        ExcMessage("Unmatched faces on periodic boundaries"));
```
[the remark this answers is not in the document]
peterrum: I have commented this and the next assert, because both cases might happen for periodic distributed coarse grids. 

**Note**:  [step-5](https://github.com/peterrum/dealii-pft/blob/master/step-5/src/main.cpp) deals with periodic boundaries. I am not very happy with the fact that two times `add_periodicity()` has to be called. I might be able to make the periodicity part of `ConstructionData`...

source/grid/tria.cc
bangerth: You mean why this can only happen in 1d? In 2d/3d, we enforce the requirement that each edge can only have at most one hanging node, and as a consequence neighboring cells can differ by at most one refinement level. But in 1d, there are no hanging nodes and so we allow neighboring cells to differ by more than one refinement level.

Since this wasn't clear to you, it may also not be clear to others -- maybe it's worthwhile paraphrasing what I wrote above into a comment at the location where you added the code?

peterrum: Thank you for the explanation! I was also thinking about that... It just felt natural that there is always limitations in the level differences. Limitation is useful for the hanging nodes (p4est, space-filling curves, etc.) and for multigrid...

I have slightly modified your comments and added those in PR #8423.

bangerth: Thanks. I agree that there are good reasons to limit the level difference between neighboring cells. We generally do that through the smoothing flags given to the triangulation and enforced upon mesh refinement, but I don't recall whether there is a way to enforce this kind of difference in 1d.

tjhei:
First, thanks for tackling a difficult project like this one! I think it is an important step to support a) more flexible partitioning than p4est offers, b) efficient support for large coarse meshes (p4est is not good at partitioning coarse meshes as the order in the z-curve is fixed).

I agree with @kronbichler that we need to be careful about API. A few comments (without having looked closely at the code):
- I think it was a mistake to have ``Triangulation`` and ``parallel::distributed::Triangulation`` as names and using inheritance (deriving for functionality and API is bad practice). Using the same class name is also not a great idea, for example doxygen is being tripped up by this (there are warnings about it in the logs) and it is confusing in IDEs. My vote would be to have distinctive names and not use long nested namespaces.
- I would also offer the suggestion that we rename the existing ``p::d::Triangulation`` in this process.
- Taking a step back, it would be great to allow various ways to partition, modify, and represent meshes in the future. I don't know how the partitioning is implemented here, but it would be great if we can also support Zoltan, Parmetis, or whatever else the user wants in the future. Ultimately, if the end product is a ``dealii::Triangulation`` (a list of locally "interesting" cells on each processor), the rest doesn't really matter, right? Would it make sense to split the ``mesh partitioner`` (p4est, your PR, etc.) from the ``Triangulation``? The interaction would mostly be "make Triangulation from partitioner" and "take refinement flags from Triangulation, refine, repartition". I think this closely resembles the p4est model as well. I see no reason in having the partitioning be so closely coupled to the triangulation (and I think what we did for p::d::Tria was a mistake)
- What is the difficulty in supporting AMR in this class? I am asking because this stops it from being a drop-in replacement.

Anyways, great progress!

tjhei:
> lease consider this PR as an opportunity
> to discuss the concepts (also in person at the deal.II workshop in a week).

I'll take you up on that. :-) I think this is a great thing to have discussions about in person.

peterrum:
@tjhei: The idea is, indeed, to split up responsibilities: 
1. triangulation (cell-loops, ...), 
2. generation of the initial mesh and partitioning (I call it front-end/`ConstructionData` at the moment; with Metis and p4est at the moment; IO), and 
3. adaptive re-partitioning (AMR, Parmetis, p4est).

I have mainly dealt with the first two points (it is by no mean as mature as `p:d:t`). On the field of AMR, I am unfortunately not an expert (yet): I think it is doable without too much work as long as coarse cell do not have to move from process to process. As far as I am aware, changing coarse grids are not taken care in `dealii::Triangulation`.

bangerth:
On 7/28/19 8:55 AM, Timo Heister wrote:
> (p4est is not good 
> at partitioning coarse meshes as the order in the z-curve is fixed).

Which is why we pre-sort the coarse mesh cells :-)

bangerth:
On 7/28/19 8:55 AM, Timo Heister wrote:
> # I think it was a mistake to have |Triangulation| and 
> |parallel::distributed::Triangulation| as names and using inheritance 
> (deriving for functionality and API is bad practice). Using the same 
> class name is also not a great idea, for example doxygen is being 
> tripped up by this (there are warnings about it in the logs) and it is 
> confusing in IDEs. My vote would be to have distinctive names and not 
> use long nested namespaces.
> # I would also offer the suggestion that we rename the existing 
> |p::d::Triangulation| in this process.

I disagree with this. The derivation from the base class has allowed us 
to avoid a large amount of code duplication because essentially all 
functionality is in the base class and most derived functionality 
doesn't care -- it only needs to know whether cells are locally owned or 
ghost or artificial, and then these functions can work on sequential, 
shared, or distributed triangulations. Furthermore, we can add any 
number of additional triangulation types without upsetting this balance.

If we had to redo things, we should have had a triangulation base class 
that all triangulation implementations derive from, rather than letting 
the parallel triangulations derive from the sequential triangulation. 
But that's really the only issue I see with the current structure. I 
certainly don't think that we ought to templatize on the triangulation 
type, for example.

bangerth:
On 7/28/19 8:55 AM, Timo Heister wrote:
> Taking a step back, it would be great to allow various ways to 
> partition, modify, and represent meshes in the future. I don't know how 
> the partitioning is implemented here, but it would be great if we can 
> also support Zoltan, Parmetis, or whatever else the user wants in the 
> future.

@peterrum: You may want to look at how the shared::Triangulation does 
exactly this.

tjhei:
> (p4est is not good at partitioning coarse meshes as the order in the z-curve is fixed).
> Which is why we pre-sort the coarse mesh cells :-)

Which is still not a great way to partition the mesh if you have more than a handful of coarse cells, in my opinion.

> it only needs to know whether cells are locally owned or ghost or artificial, and then these functions can work on sequential, shared, or distributed triangulations

And my point is that all this could be achieved without having access to private things inside the Triangulation, which means deriving from the serial Triangulation would not be necessary.

> If we had to redo things, we should have had a triangulation base class that all triangulation implementations derive from, rather than letting the parallel triangulations derive from the sequential triangulation.

This is what I am saying: using inheritance for defining an interface _and_ to re-use implementation at the same time is rarely a good idea. I think the code is more complicated than it need to be. Anyways, it might be too late to change this now even though I like thinking about what it would take.

peterrum:
@bangerth all suggestions are welcome!

PR #8419, opened 2019-07-26
Bloch periodic conditions (a.k.a. phase shift periodic conditions)

dangars:
Hi, following up on the discussion https://github.com/dealii/dealii/pull/8384#pullrequestreview-265567370 This PR adds a factor to `make_periodicity_constraints()` which can be used to implement periodic boundary conditions with a phase shift.

If the domain is periodic, the solution of the wave equation satisfies `u(d) = e(i*omega*t - k*d)*u(0)` where omega is the angular frequency, `k` the wavevector and `d` the spatial periodicity. In the harmonic domain you can use this PR to apply the condition `u(d) = e(- k*d)*u(0)`

@tamiko Could you let know what you think?

I still need have to test it.


source/dofs/dof_tools_constraints.cc
```diff
@@ -2517,18 +2529,21 @@ namespace DoFTools
 
     make_periodicity_constraints<DoFHandlerType>(matched_faces,
                                                  constraints,
-                                                 component_mask);
+                                                 component_mask,
+                                                 std::vector<unsigned int>(),
+                                                 periodicity_factor);
```
dangars: Here and below. I need to pass the parameter `periodicity_factor` to `make_periodicity_constraints()`. The fourth parameter takes the default value. In order to pass `periodicity_factor` to `make_periodicity_constraints()` I added the default value to the fourth parameter'

tamiko:
Looking over it now :-)

include/deal.II/dofs/dof_tools.h
```diff
@@ -1153,19 +1153,20 @@ namespace DoFTools
...
+    AffineConstraints<number> &                  constraints,
     const ComponentMask &            component_mask   = ComponentMask(),
     const bool                       face_orientation = true,
     const bool                       face_flip        = false,
     const bool                       face_rotation    = false,
     const FullMatrix<double> &       matrix           = FullMatrix<double>(),
     const std::vector<unsigned int> &first_vector_components =
-      std::vector<unsigned int>());
+      std::vector<unsigned int>(),
+    const number periodicity_factor = 1.);
```
tamiko: I like this interface. I think I always thought about applying a phase shift in a too complicated manner (complex valued phase-shift rotation matrix instead of splitting the complex phase shift and the matrix-valued rotation). This seems to be a very sensible and pragmatic alternative.

tamiko:
@masterleinad Do you have some comments on this?

source/dofs/dof_tools_constraints.cc
```diff
@@ -2098,23 +2100,25 @@ namespace DoFTools
...
-              if (std::abs(cycle_factor - 1.) > eps)
+              if (std::abs(cycle_constraint_factor - 1.) > eps)
                 affine_constraints.add_line(dof_left);
             }
           else
             {
               affine_constraints.add_line(dof_left);
-              affine_constraints.add_entry(dof_left, dof_right, factor);
+              affine_constraints.add_entry(
+                dof_left, dof_right, periodicity_factor * constraint_factor);
```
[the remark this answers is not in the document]
masterleinad: The build fails because you are missing `constraint_factor` in the signature.

masterleinad:
Let's merge #8384 first.

tamiko:
@dangars As discussed we need a small testcase for this feature as well :-)

tests/bits/step-62_periodic.cc
```diff
@@ -0,0 +1,942 @@
...
+    std::ifstream file;
+    file.open("/tmp/dofs");
+    for (unsigned int idx = 0; idx < 340; ++idx)
+      {
+        file >> dof_left;  // reads in the double value
+        file >> dof_right; // reads in the double value
+        std::cout << dof_left << " i " << dof_right
+                  << std::endl; // for test purposes
+        constraints.add_line(dof_left);
+        constraints.add_entry(dof_left, dof_right, periodicity_factor);
```
dangars: This is a workaround. If I apply the constraints by hand. The test works. As you can see I commented `make_periodicity_constraints()`

source/dofs/dof_tools_constraints.cc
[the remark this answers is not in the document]
dangars: For some reason I have the feeling that after `affine_constraints.add_entry()` has been applied `make_periodicity_constraints()` makes some dofs equal to zero.

dangars:
I solved the issue :) The test gives the expected result. But let's wait until PR #8878 is merged, as I'm using the `std::complex` version of `VectorTools::integrate_difference()`

source/dofs/dof_tools_constraints.cc
```diff
@@ -2094,23 +2096,26 @@ namespace DoFTools
...
-          //    cycle_factor != 1 occurs in situations such as
+          //    cycle_constraint_factor != 1 occurs in situations such as
           //    x1 == x2 and x2 == -1. * x1. This system is only solved by
           //    x_1 = x_2 = 0.
 
           if (constraints_are_cyclic)
             {
-              if (std::abs(cycle_factor - 1.) > eps)
+              if (std::abs(cycle_constraint_factor / periodicity_factor - 1.) >
+                  eps)
```
dangars: This is the solution that I took. It works, but I have the feeling that this could break in a corner case with an unstructured mesh. Is there a way to know if `dof_left` belongs to `face_1` and `dof_right` belongs to face_2?

bangerth:
Question 1: What lives in `tests/bits/`? Essentially everything that hasn't found a more descriptive home. So your test is fine there. Or you could put it into `tests/lac/`, where a lot of the constraint tests currently already live.

Question 2: About whether to constrain x1 in terms of x2 or the other way around? Thing about the situation where your constraint looks like `a*x1 + b*x2 = 0`. You could just always pick `x1 = b/a*x2`, but in practice this is not so stable if `a` could be a small number -- intended to be zero, but just very small due to roundoff. Of course, constraining `x2` in terms of `x1` has the same problem. So one chooses `x1 = b/a*x2` if `|b|<|a|`, and  `x2 = a/b*x1` if `|a|<|b|`.

source/dofs/dof_tools_constraints.cc
```diff
@@ -1823,7 +1823,8 @@ namespace DoFTools
       const ComponentMask &                        component_mask,
       const bool                                   face_orientation,
       const bool                                   face_flip,
-      const bool                                   face_rotation)
+      const bool                                   face_rotation,
+      number                                       periodicity_factor)
```
bangerth: make it `const`

bangerth:
@tamiko -- would you mind looking at this? The actual changes are pretty small.

dangars:
@bangerth Regarding your answer. That is a very good point, I didn't think about that. I will add an `Assert` that requires that `|periodicity_factor|=>1`. I'll add it as well to the documentation.

My point was different, I think that @tamiko  could answer to this because he wrote the code. I used a debugger and I realized that the function `make_periodicity_constraints` loops over all the constraints twice (or maybe more than twice). That is why there is the safeguard to avoid cyclic constraints. I modified the safeguard, but in order to be in the safe side I would need to know if `x1` belongs to `face_1` or `face_2`, otherwise I may apply the wrong constraint.

source/dofs/dof_tools_constraints.cc
tamiko: I do not understand how the mesh structure should play a role here. :-)

A bit above:
```
2045           auto dof_left  = dofs_1[j];                                            
2046           auto dof_right = dofs_2[i];
```
and the only place where this changes is
```
2055               std::swap(dof_left, dof_right);
```
So you could simply check whether `dof_left == dofs_1[j]`.

(I am about to hit the pillows, sorry for no coherent answer). The one thing I see that we have to check is the following: Let's say for a square domain with bloch periodicity on top and bottom and regular periodicity left and right, do we correctly identify the left and right point and set the phase shift between the upper and lower points? (without forcing the degree of freedom to 0.)

source/dofs/dof_tools_constraints.cc
```diff
@@ -1966,19 +1968,19 @@ namespace DoFTools
...
           // of the row of the transformation matrix and check whether we
           // find exactly one nonzero entry. If this is the case, set
           // "is_identity_constrained" to true and record the corresponding
-          // index and factor.
+          // index and constraint_factor.
 
           bool         is_identity_constrained = false;
           unsigned int target                  = numbers::invalid_unsigned_int;
-          double       factor                  = 1.;
+          number       constraint_factor       = periodicity_factor;
```
dangars: Now I added `periodicity_factor` here. This is more consistent with code that @tamiko wrote. Now the code below remains almost the same.

source/dofs/dof_tools_constraints.cc
```diff
@@ -2094,23 +2096,50 @@ namespace DoFTools
...
+              // it would be necessary to rewrite this function and the dof
+              // ordering method should be modified.
+              // Let's take the following constraint a*x1 + b*x2 = 0. You could
+              // just always pick x1 = b/a*x2, but in practice this is not so
+              // stable if a could be a small number -- intended to be zero, but
+              // just very small due to roundoff. Of course, constraining x2 in
+              // terms of x1 has the same problem. So one chooses x1 = b/a*x2 if
+              // |b|<|a|, and x2 = a/b*x1 if |a|<|b|.
+              Assert(
+                std::abs(constraint_factor) < 1e10,
```
dangars: This code is only valid for real wavenumbers. If the wavenumber is real then the absolute value of `periodic_factor` is one, there is only a phase shift.

In case we want to add support for imaginary wavevectors. I wrote in this comment what it has to be done.

I added an assert in case imaginary wavectors are used.

dangars:
This PR is ready :)

I modified the test and now it is compatible with the Bloch conditions in 2D
* I applied Bloch conditions in 2 dimensions. One for x (left and right) and another one for y (bottom and top) . This is what @tamiko  suggested to check.
* I checked that there are no degrees of freedom forced to 0 and that the constraints are correctly applied.
* In order to test the general case, the conditon on x is different than the condition on y. The wavevector is not parallel to x or y, it points to somewhere in the middle.
* The fabricated solution is compatible with the Bloch conditions.
* I checked that the simulated solution is equal to the exact solution.

include/deal.II/dofs/dof_tools.h
```diff
@@ -1150,10 +1150,36 @@ namespace DoFTools
    * This function makes sure that identity constraints don't create cycles
    * in @p constraints.
    *
+   * @p periodicity_factor can be used to to implement Bloch periodic conditions
+   * (a.k.a. phase shift periodic conditions) of the form
+   * $\psi(\mathbf{r})=e^{-i\mathbf{k}\cdot\mathbf{r}}u(\mathbf{r})$
+   * where $u$ is periodic with the same periodicity as the crystal lattice and
+   * and $k$ is the wavevector, see
+   * [Photonic Crystals: Molding the Flow of Light by J.D. Joannopoulos et
+   * al](https://press.princeton.edu/books/hardcover/9780691124568/photonic-crystals).
```
masterleinad: I would consider referring to https://en.wikipedia.org/wiki/Bloch_wave instead since that is freely available.

source/dofs/dof_tools_constraints.cc
dangars: > Why not just constrain one of the dofs to zero if the constraint_factor is unreasonably small/large?

@masterleinad I agree with the rest of the comments. But I would prefer to place an assert here. If I set the dofs to zero when the constraint_factor is unreasonably small/large, then if somebody runs into this situation they will have a hard time to find out why their program is not working. Whereas if there is an assert here they can identify the source of the problem, and that leaves the door open to modify this function in order to deal properly with those situations.

Let me know what is best, the assert or set the dofs to zero.

dangars:
> Thanks! Looks good to me.

Great! I addressed your comments and wrote and more simple test

dangars:
Should I address other comments/issues?

dangars:
@masterleinad thanks :)

PR #8421, opened 2019-07-26
 New tests: Parallel distributed version of step-27. 

marcfehling:
Right on time for the upcoming deal.II workshop, we prepared parallel distributed versions of the hp-tutorial `step-27` using both PETSc and Trilinos.

We end up with a slightly different refinement behavior depending on the amount of processors assigned, and that's with each library. This may be due to the symmetry of the problem.

Further, PETSc seems to converge with lesser iterations than Trilinos, but the refinement history seems to be less dependent on the number of processors with Trilinos.

A separate tutorial step will follow once the remaining suggestions of #7515 are implemented.

Part of #3511.


marcfehling:
@bangerth -- FYI

PR #8423, opened 2019-07-27
Enable periodicity for 1D

peterrum:
This PR enabled periodicity for 1D triangulations. It works if following functions are called:
- `GridTools::collect_periodic_faces()`
- `Triangulation::add_periodicity()`

It is tested by: tests/grid/periodicity_01.cc


peterrum:
The test tests/bits/periodicity_01.cc uses the function `DoFTools::make_periodicity_constraints()`, which calls `DoFAccessor::n_active_fe_indices()`  and `DoFAccessor::nth_active_fe_index()`. These functions are not yet supported `DoFAccessor<0,...>`.

Implementing `DoFAccessor<0,..>::n_active_fe_indices()` is easy (just return 1), however, I am not sure how to fill  `DoFAccessor<0,..>::nth_active_fe_index()`.

source/grid/tria.cc
```diff
@@ -950,8 +950,6 @@ namespace
     Assert(face_1->at_boundary() && face_2->at_boundary(),
            ExcMessage("Periodic faces must be on the boundary"));
 
-    Assert(std::abs(cell_1->level() - cell_2->level()) < 2, ExcInternalError());
```
kronbichler: Shouldn't we leave this assertion in and instead check
```
Assert(dim == 1 || std::abs(cell_1->level() - cell_2->level()) < 2, ExcInternalError());
```

peterrum: @kronbichler I didn't remove it; I only moved it down into the else-statement. But I can keep it here; then it better fits to the @bangerth's explanation, which I will paraphrase.

peterrum:
The test `tests/bits/periodicity_01.cc` is working. However, I am not completely sure if the changes I made in dof_accessor.templates.h are valid.

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -2565,8 +2565,7 @@ inline unsigned int
 DoFAccessor<0, DoFHandlerType<1, spacedim>, level_dof_access>::
   n_active_fe_indices() const
 {
-  Assert(false, ExcNotImplemented());
-  return numbers::invalid_unsigned_int;
+  return 1;
```
bangerth: This (and the other function you have below) are correct if `DoFHandlerType` is `::DoFHandler`, but not if it is a `hp::DoFHandler`. Do you know whether there are specializations? You might want to take a look at how this function is implemented in the case of `dim>1`.

Alternatively, if that is too complicated, you could add an `Assert` checking that `std::same_type<DoFHandlerType,DoFHandler<1,spacedim>>::value == true`, saying that the other cases are not implemented.

peterrum: @bangerth: thanks! I have chosen the approach with the asserts. 

I have tried to implement the `hp::DoFHandler`-version, however, was not able to find an equivalent of `this->present_index` for `DoFAccessor<0,..>`. There is only `global_vertex_index`, which is obviously not equivalent...

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -2565,8 +2566,11 @@ inline unsigned int
 DoFAccessor<0, DoFHandlerType<1, spacedim>, level_dof_access>::
   n_active_fe_indices() const
 {
-  Assert(false, ExcNotImplemented());
-  return numbers::invalid_unsigned_int;
+  const bool check = std::is_same<DoFHandlerType<1, spacedim>,
+                                  dealii::DoFHandler<1, spacedim>>::value;
+  Assert(check == true, ExcNotImplemented());
```
kronbichler: Can you move the `std::is_same...` into the assert? This avoids introducing the variable `check` that is unused in release mode.

peterrum: No, I cannot. I have tried it. But the compiler complained:
```bash
dof_accessor.templates.h:2570:103: error: macro "Assert" passed 5 arguments, but takes just 2
                                   dealii::DoFHandler<1, spacedim>>::value == true, ExcNotImplemented());
                                                                                                       ^
```
Do you know a workaround?

kronbichler: Try to put it in braces, `(std::is_same<...>)`.

tamiko: There is also the following idiom:
```
#ifdef DEBUG
    // ...
    Assert(...);
#endif
```

peterrum: @kronbichler: That worked! Stupid that I did not think of that myself...

bangerth:
FYI, the issue with the compiler error comes from the following: You were trying to write
```
  Assert (std::is_same<X,Y>::value, ExcNotImplemented());
```
To a human reader, it's clear that `<X,Y>` is simply a template list, and the compiler itself also sees it that way. But `Assert` is a macro that takes two arguments, and the C++ preprocessor does its work before the compiler is run. So it sees that you're trying to use the macro with three arguments: `std::is_same<X`, `Y>::value`, and `ExcNotImplemented()`. In the first and second argument, the template brackets are interpreted as less-than and larger-than operators. The point is that using the preprocessor is complicated with template arguments. You can avoid all of this using the parentheses:
```
  Assert ((std::is_same<X,Y>::value), ExcNotImplemented());
```
This makes it clear to the preprocessor that the first comma does not separate macro arguments, but either separates the elements of a template or function argument list, or is in fact a comma operator.

I know it's stupid, and we've all stumbled over this issue at one point...

PR #8426, opened 2019-07-28
Add two new methods to VectorizedArray

peterrum:
In PR #8400, we expressed to add two new methods to `VectorizedArray`:
- constructor:
```cpp
VectorizedArray(const Number scalar);
```
- broadcast function:
```cpp
void broadcast_from_scalar(const Number scalar);
```
This PR intends to add these two functions and remove existing workarounds that were needed due to the lack of these (e.g. #8064).

@masterleinad  @tjhei, @bangerth, @kronbichler ping

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -225,6 +225,20 @@ public:
   // and copy functions (the standard is somewhat relaxed in C++2011, though).
 
   /**
+   * Default empty constructor, leaving the data in an uninitialized state
+   * similar to float/double.
+   */
+  VectorizedArray() = default;
+
+  /**
+   * Construct an array with the given scalar broadcast to all lanes
+   */
+  VectorizedArray(const Number scalar)
+  {
+    this->operator=(scalar);
+  }
+
+  /**
    * This function assigns a scalar to this class.
    */
   DEAL_II_ALWAYS_INLINE
@@ -667,6 +681,20 @@ public:
   static const unsigned int n_array_elements = 8;
 
   /**
+   * Default empty constructor, leaving the data in an uninitialized state
+   * similar to float/double.
+   */
+  VectorizedArray() = default;
+
+  /**
+   * Construct an array with the given scalar broadcast to all lanes
+   */
+  VectorizedArray(const double scalar)
+  {
+    this->operator=(scalar);
+  }
+
+  /**
    * This function can be used to set all data fields to a given scalar.
    */
   DEAL_II_ALWAYS_INLINE
@@ -1084,6 +1112,20 @@ public:
    * This gives the number of vectors collected in this class.
    */
   static const unsigned int n_array_elements = 16;
+
+  /**
+   * Default empty constructor, leaving the data in an uninitialized state
+   * similar to float/double.
+   */
+  VectorizedArray() = default;
+
+  /**
+   * Construct an array with the given scalar broadcast to all lanes
+   */
+  VectorizedArray(const float scalar)
+  {
+    this->operator=(scalar);
+  }
 
   /**
    * This function can be used to set all data fields to a given scalar.
@@ -1556,6 +1598,20 @@ public:
   static const unsigned int n_array_elements = 4;
 
   /**
+   * Default empty constructor, leaving the data in an uninitialized state
+   * similar to float/double.
+   */
+  VectorizedArray() = default;
+
+  /**
+   * Construct an array with the given scalar broadcast to all lanes
+   */
+  VectorizedArray(const double scalar)
+  {
+    this->operator=(scalar);
+  }
+
+  /**
    * This function can be used to set all data fields to a given scalar.
    */
   DEAL_II_ALWAYS_INLINE
@@ -1942,6 +1998,20 @@ public:
    * This gives the number of vectors collected in this class.
    */
   static const unsigned int n_array_elements = 8;
+
+  /**
+   * Default empty constructor, leaving the data in an uninitialized state
+   * similar to float/double.
+   */
+  VectorizedArray() = default;
+
+  /**
+   * Construct an array with the given scalar broadcast to all lanes
+   */
+  VectorizedArray(const float scalar)
+  {
+    this->operator=(scalar);
+  }
 
   /**
    * This function can be used to set all data fields to a given scalar.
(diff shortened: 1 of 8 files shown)


peterrum:
**Note:** I will need your help since I am not that much familiar with POD and not aware of all discussions related to this topic. Also it might be useful to setup a very simple code and have a look at the compiler output, to make sure that the new constructor does not introduce any unwanted overhead.

include/deal.II/base/vectorization.h
```diff
@@ -224,6 +224,20 @@ class VectorizedArray
   // POD means that there should be no user-defined constructors, destructors
   // and copy functions (the standard is somewhat relaxed in C++2011, though).
 
+  VectorizedArray()
```
masterleinad: I would prefer ` = default` here and in the other places below. Then, you don't need to modify the `static_assertion`s below, i.e. the class is still `trivial` and has `standard_layout`, see https://godbolt.org/z/L9Q1hN.

include/deal.II/base/vectorization.h
```diff
@@ -224,6 +224,20 @@ class VectorizedArray
...
+  {}
+
+  VectorizedArray(const Number scalar)
+  {
+    this->operator=(scalar);
+  }
+
+  void
+  broadcast_from_scalar(const Number scalar)
```
masterleinad: @kronbichler @tjhei Do we want to deprecate `operator=(const Number)` then? I don't see a reason for having to functions doing exactly the same (both conceptually and behaviorwise).

source/base/vectorization.cc
```diff
@@ -29,6 +30,7 @@ static_assert(std::is_standard_layout<VectorizedArray<double>>::value &&
 static_assert(std::is_standard_layout<VectorizedArray<float>>::value &&
                 std::is_trivial<VectorizedArray<float>>::value,
               "VectorizedArray<float> must be a POD type");
+ */
```
masterleinad: See above.

include/deal.II/base/vectorization.h
kronbichler: We need to provide `operator=` for compatibility in the various tensor classes where we use it to set to zero. So that part is fundamental to the class and the compatibility we want to achieve. The question is whether we should omit this `broadcast_from_scalar` then?

include/deal.II/base/vectorization.h
peterrum: You are right. I have updated the code.

include/deal.II/base/vectorization.h
```diff
@@ -224,6 +224,19 @@ class VectorizedArray
   // POD means that there should be no user-defined constructors, destructors
   // and copy functions (the standard is somewhat relaxed in C++2011, though).
 
+  VectorizedArray() = default;
+
+  VectorizedArray(const Number scalar)
```
kronbichler: Can you please add a short documentation - "Construct an array with the given scalar broadcast to all lanes."?

peterrum: I have added the documentation. I'll wait with the documentation of `broadcast_from_scalar` until we have decides, if we actually want it.

masterleinad:
For comparison in terms of assembly between the version with a defaulted constructor and the current code see https://godbolt.org/z/Y_vwt9.

kronbichler:
The things to look over to see whether they can be removed:
- https://github.com/dealii/dealii/blob/master/include/deal.II/base/numbers.h#L682-L785 - here we have other functionality, so we likely cannot fully remove it
- https://github.com/dealii/dealii/blob/master/include/deal.II/base/tensor.h#L743-L801
- https://github.com/dealii/dealii/blob/master/include/deal.II/base/symmetric_tensor.h#L3539-L3548

peterrum:
I have modified `symmetric_tensor.h`. Changes in `tensor.h` and `numbers.h` lead to a lot of compiler errors...

kronbichler:
> Changes in tensor.h and numbers.h lead to a lot of compiler errors...

It's not clear whether we can remove them all, my list was rather meant to list some workarounds we had done over the time.

masterleinad:
> > Changes in tensor.h and numbers.h lead to a lot of compiler errors...
> 
> It's not clear whether we can remove them all, my list was rather meant to list some workarounds we had done over the time.

Also, we don't need to clean up everything in this pull request.

peterrum:
@masterleinad: OK! Could you start the tests, so that I can see if there are any issues?

masterleinad:
Apart from adding a test, a changelog entry and discussing `broadcast_from_scalar` this looks good to me.

masterleinad:
I would suggest to postpone `braodcast_from_scalar` (or discussing it) and merge this already (with a changelog entry and a test).

peterrum:
@masterleinad I have removed `braodcast_from_scalar` for new (but if @tjhei still wants, lets add it back in a next PR) as well as added the test and the missing change log.

include/deal.II/base/vectorization.h
```diff
@@ -224,6 +224,14 @@ class VectorizedArray
   // POD means that there should be no user-defined constructors, destructors
   // and copy functions (the standard is somewhat relaxed in C++2011, though).
 
+  VectorizedArray() = default;
+
+  // Construct an array with the given scalar broadcast to all lanes
```
masterleinad: This is not a `doxygen` comment. Can you please do
```suggestion
  /**
   * Construct an array with the given scalar broadcast to all lanes.
   */
```
here and below?

tests/base/vectorization_11.cc
```diff
@@ -52,20 +52,25 @@ struct Tester<double>
   test()
   {
```
masterleinad: Is there a reason to rename the file?

include/deal.II/base/vectorization.h
peterrum: I have changed these lines.

tests/base/vectorization_11.cc
peterrum: The test used to only test `VectorizedArray::make_vectorized_array`. Now it also tests the constructor. This is the reason for the name change.

masterleinad: OK.

include/deal.II/base/vectorization.h
```diff
@@ -3285,6 +3365,14 @@ class VectorizedArray<float, 4>
    */
   static const unsigned int n_array_elements = 4;
 
+  VectorizedArray() = default;
+
+  // Construct an array with the given scalar broadcast to all lanes
```
masterleinad: Looks like you missed one place.

peterrum: Thanks!

include/deal.II/base/vectorization.h
```diff
@@ -224,6 +224,16 @@ class VectorizedArray
   // POD means that there should be no user-defined constructors, destructors
   // and copy functions (the standard is somewhat relaxed in C++2011, though).
 
+  VectorizedArray() = default;
```
kronbichler: I suggest we add a short comment `Default empty constructor, leaving the data in an uninitialized state similar to float/double.` to this new constructor as well.

peterrum: @kronbichler: It is pushed. @masterleinad: Hope I did miss any this time!

Issue #8376, opened 2019-07-11
Communicate active_fe_indices after refinement?


marcfehling:
Reviewing the recently added feature of an automatic transfer of `active_fe_indices` after refinement, I noticed that we are communicating these indices on ghost (and artificial) cells in case of a `parallel::shared::Triangulation` ...
https://github.com/dealii/dealii/blob/59a12e7559c7710ae4125632c9fea25be61d727d/source/hp/dof_handler.cc#L2119
... but skip it for `parallel::distributed::Triangulation`. Thus, they are only available on locally owned cells in the latter case.
https://github.com/dealii/dealii/blob/59a12e7559c7710ae4125632c9fea25be61d727d/source/hp/dof_handler.cc#L2148

---

In applications, we usually have to call `distribute_dofs()` which does the job of communicating the indices, but it may lead to confusion whenever a user wants to access an `active_fe_index` on a ghost cell immediately after refinement happened.

What do you think? Should we include the communication, or leave it out completely? I'll tend towards including the communication to avoid confusion of users.

I'll prepare a PR according to the result of this discussion :)

bangerth:
Yes, I think that is reasonable. The added work is rather minimal, I imagine.

PR #8429, opened 2019-07-29
Parallel distributed: Communicate active_fe_indices after deserializa…

marcfehling:
Follow-up to #8422. I just realized that we need to update the indices after deserialization as well to be consistent with the requirements in #8376. Sorry for any inconvenience.

### DIFF
--- a/source/hp/dof_handler.cc
+++ b/source/hp/dof_handler.cc
@@ -2281,6 +2281,10 @@ namespace hp
         // Update all locally owned active_fe_indices.
         set_active_fe_indices(active_fe_index_transfer->active_fe_indices);
 
+        // Update active_fe_indices on ghost cells.
+        dealii::internal::hp::DoFHandlerImplementation::Implementation::
+          communicate_active_fe_indices(*this);
+
         // Free memory.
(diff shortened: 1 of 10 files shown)


drwells:
I am surprised that we did not catch this before. Do we have any tests that serialize and deserialize an `hp::DoFHandler` in parallel?

marcfehling:
The test `mpi/p4est_save_06` deals with parallel `hp::DoFHandler` serialization. In these tests however, we call `distribute_dofs()` immediately afterwards, which also takes care of communicating the `active_fe_indices` on ghost cells. In real applications, this would be the usual workflow.

We want to avoid confusion for users by communicating those during the refinement/seriailzation process and not only by calling `distribute_dofs()`. The amount of work added should be negligible.

bangerth:
While the existing tests checks the "common" situation, can you clone the test and adjust the copy so that it verifies the condition you implement here?

marcfehling:
Enhanced all tests related to the transfer of `active_fe_indices` to verify ghost cells as well.

marcfehling:
ping -- most of the lines changed result from adding ghost related ones in the output file

PR #8430, opened 2019-07-29
Reduce floating-point division in operator/

masterleinad:
Profiling some code that was using `Tensor::operator/` we noticed that we repeatedly do floating-point division when we can just multiply by the inverse instead. I looked through the entire code base, but could only find a few more places.


peterrum:
The reason why some tests have failed is that `OtherNumber` might be `int`. A fix would be to use at those places `Number` (once you have rebased with master):
```diff
diff --git a/include/deal.II/base/tensor.h b/include/deal.II/base/tensor.h
index c57fad6..b603909 100644
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -1356,7 +1356,7 @@ template <typename OtherNumber>
 DEAL_II_CONSTEXPR inline DEAL_II_CUDA_HOST_DEV Tensor<rank_, dim, Number> &
 Tensor<rank_, dim, Number>::operator/=(const OtherNumber &s)
 {
-  OtherNumber inverse_factor = OtherNumber(1.) / s;
+  Number inverse_factor = Number(1.) / s;
   for (unsigned int i = 0; i < dim; ++i)
     values[i] *= inverse_factor;
   return *this;
@@ -1759,7 +1759,7 @@ DEAL_II_CONSTEXPR DEAL_II_CUDA_HOST_DEV inline DEAL_II_ALWAYS_INLINE
 {
   // recurse over the base objects
   Tensor<rank, dim, typename ProductType<Number, OtherNumber>::type> tt;
-  const OtherNumber inverse_factor = OtherNumber(1.) / factor;
+  const Number inverse_factor = Number(1.) / factor;
   for (unsigned int d = 0; d < dim; ++d)
     tt[d] = t[d] * inverse_factor;
   return tt;
```
This fixes all but one test. Test `data_out/data_out_base_vtu_04.debug` still failes, but I guess because there VTU binary files are compared.

include/deal.II/base/point.h
```diff
@@ -549,8 +549,9 @@ inline DEAL_II_CUDA_HOST_DEV
   Point<dim, Number>::operator/(const OtherNumber factor) const
 {
   Point<dim, typename ProductType<Number, OtherNumber>::type> tmp;
+  const Number inverse_factor = Number(1.) / factor;
```
bangerth: Here and in the place below, I tried to think through whether we want to use `Number` or `OtherNumber` as type. I came to the conclusion that what we really want (I think!) is
```
  const decltype(this->operator[](0)/factor)  inverse_factor = ...;
```

include/deal.II/base/tensor.h
```diff
@@ -1362,8 +1362,9 @@ template <typename OtherNumber>
 DEAL_II_CONSTEXPR inline DEAL_II_CUDA_HOST_DEV Tensor<rank_, dim, Number> &
 Tensor<rank_, dim, Number>::operator/=(const OtherNumber &s)
 {
+  OtherNumber inverse_factor = OtherNumber(1.) / s;
```
bangerth: With the same argument as above, the right hand side should then be `Number(1.0) / s`.

include/deal.II/base/tensor.h
```diff
@@ -1764,8 +1765,9 @@ DEAL_II_CONSTEXPR DEAL_II_CUDA_HOST_DEV inline DEAL_II_ALWAYS_INLINE
 {
   // recurse over the base objects
   Tensor<rank, dim, typename ProductType<Number, OtherNumber>::type> tt;
+  const OtherNumber inverse_factor = OtherNumber(1.) / factor;
```
bangerth: and similarly here

masterleinad:
We should probably also handle the case of an integral underlying type via a different code path. Otherwise, we would again multiply by 0 in case we want to divide, e.g. an object of `Tensor<1, dim, int>`, by an `int` scalar.

include/deal.II/base/tensor.h
jppelteret: Drive by comment: If this causes problems with the AD tests, then the fix will probably be to use `internal::NumberType<OtherNumber>::value(1.)` here and below, and similarly in the point class. 

Also, this should be `const OtherNumber inverse_factor` :-)

masterleinad:
That should do the trick.

include/deal.II/base/tensor.h
```diff
@@ -1739,6 +1783,57 @@ DEAL_II_CUDA_HOST_DEV constexpr DEAL_II_ALWAYS_INLINE
...
+                                        const OtherNumber &              factor)
+    {
+      Tensor<rank, dim, typename ProductType<Number, OtherNumber>::type> tt;
+      // recurse over the base objects
+      for (unsigned int d = 0; d < dim; ++d)
+        tt[d] = t[d] / factor;
+      return tt;
+    }
+  } // namespace TensorImplementation
+} // namespace internal
```
kronbichler: Can't you express the member function `/=` in terms of this function (or the other way around) to only have one implementation class? For example, you could pass in `*this` as the operand in this function? This puts some duty on the compiler to properly handle return types but I think it will be fine.

masterleinad: I could unify the two by having a 
```
division_operator(const Tensor &input, const Number scalar, Tensor output)
{
  for (i ...)
    output[i] = input[i]/scalar;
}
```
interface and
```
operator/(const Tensor &t, const Number scalar)
{
  Tensor tmp;
  division_operator(t, scalar, tmp);
 return tmp;
}

Tensor::operator/=(const Number scalar)
{
  division_operator(t, scalar, t);
}
```

I am just not sure if we copy unnecessarily for `operator/=` then.

masterleinad:
I am kind of at a loss here. Locally, everything works (also with `Adol-C`).

masterleinad:
That seemed to have worked. Regarding joining the two cases, https://godbolt.org/z/Kx6_2y shows what I had in mind. It seems like the compiler is not able to generate equivalent assembly so I would rather have two implementations.

include/deal.II/base/point.h
```diff
@@ -560,6 +560,50 @@ inline DEAL_II_CUDA_HOST_DEV
...
+                std::is_integral<
+                  typename ProductType<Number, OtherNumber>::type>::value,
+                int>::type = 0>
+    inline DEAL_II_CUDA_HOST_DEV
+      Point<dim, typename ProductType<Number, OtherNumber>::type>
+      division_operator(const Point<dim, Number> &p, const OtherNumber factor)
+    {
+      Point<dim, typename ProductType<Number, OtherNumber>::type> tmp;
+      for (unsigned int i = 0; i < dim; ++i)
+        tmp[i] = p[i] / factor;
```
bangerth: What's the reason to not pull out the inverse of the factor here? Is it more efficient to divide by an integer than to multiply by a floating point number?

jppelteret: I had been wondering the same thing wrt. all of the refactoring that's been done here, but I thought that I must be missing something. Couldn't we forward all of the operations for `Point::operator/=` and `Tensor::operator/=` to their respective `operator*=`?

masterleinad: This is the case where both the `ProductType` for `Number` and `OtherNumber` is integral. Hence, `1/factor` is zero and it seems that we can't use the same implementation.

tamiko:
@bangerth @masterleinad *ping* What's the status on this one?

include/deal.II/base/complex_overloads.h
```diff
@@ -45,6 +45,21 @@ operator*(const std::complex<T> &left, const std::complex<U> &right)
...
+/**
+ * Provide an <tt>operator/</tt> that operates on mixed complex floating point
+ * types. Annoyingly, the standard library does not provide such an
+ * operator...
+ *
+ * @relatesalso ProductType
+ */
+template <typename T, typename U>
+typename ProductType<std::complex<T>, std::complex<U>>::type inline
+operator/(const std::complex<T> &left, const std::complex<U> &right)
```
bangerth: the `inline` should precede the return type

include/deal.II/base/complex_overloads.h
```diff
@@ -63,6 +78,22 @@ operator*(const std::complex<T> &left, const U &right)
...
+ * Provide an <tt>operator/</tt> for a scalar division of a complex
+ * floating point type with a different real floating point type. Annoyingly,
+ * the standard library does not provide such an operator...
+ *
+ * @relatesalso EnableIfScalar
+ * @relatesalso ProductType
+ */
+template <typename T, typename U>
+typename ProductType<std::complex<T>,
+                     typename EnableIfScalar<U>::type>::type inline
```
bangerth: same here

bangerth: is `EnableIfScalar` false for complex types? I tend to think of complex types as scalars as well...

include/deal.II/base/complex_overloads.h
```diff
@@ -80,6 +111,23 @@ operator*(const T &left, const std::complex<U> &right)
...
+ * Provide an <tt>operator/</tt> for a scalar division of a real
+ * floating point type with a different complex floating point type.
+ * Annoyingly, the standard library does not provide such an operator...
+ *
+ * @relatesalso EnableIfScalar
+ * @relatesalso ProductType
+ */
+template <typename T, typename U>
+typename ProductType<typename EnableIfScalar<T>::type,
+                     std::complex<U>>::type inline
```
bangerth: and here

tests/data_out/data_out_base_vtu_04.with_zlib=on.output
```diff
@@ -9,7 +9,7 @@
...
...d9W91uoD9YPWnx74q+0+QW2o3lX3WqsP1A/RR/un2tkn+Vvp/7/qIHIXXag+Mh7XzXrCJpjJMQsXci7yw87x19X1hE0wk2MWLuRchCM5x4W7ziZsgpkcs3Ah5yIcyTkiNsFMjlm4kHMRjpT1Pxhf1IbqXXWvtfpA/aD1vR/UbfcJakP1rrrXWn2gftD66z8+aWefZG2o3lX3zbdd6Kgfmp5n29kng/F97sblhE0w145ZuJBzEY7kHI992UvYBHPtmIULORfhSM6x1TqdsAnm2jELF3IuwpFA/0kbMjPZZ6ac7LNU3nOe/1kbMjPZZ6ac7LNU3nOe/1kbMjPZZ6ac7LNU3nOe//lbyQwhnyFyss8Wec8w/xM20Tv7nSU+YL/LhK/2tEtsonf2O0t8wH6XiT9SP2ITvbPfWeID9rtM/JH6EZv8/7PPcNEF+2wXvaT+QfmiNmRmks9MOdlnqbznPP+zNmRmks9MOclnqbznPP+zNmRmks9MOclnqbwnmP8D8UVsonf2O0t8wH6XiT/475k/J0psonf2O0t8wH6XiT/4zhPTEyU20Tv5nSU+IL/LVPPSM1FqwzRDoCUCjXGpDdMMgZYINEalNkwzBFoi0BiV38q+IcG3JfjmG7CZZxm8zOBxLrGZZxm8zOBxLrGZZxm8zOBxLrGZhgm0zaB5HoYvasM0U4OWCDRGpTZMMzVoiUBjVGrDNFODlgg0RsPwRWzmWQIvM3icS2zmWQIvE3icS2zmWQIvE3icSm3YzsCwSxj+yZ75pE8btjMw7BKx3jPT6ZlP+rRhOwPDLhHrKycO98wnfd/K7lCGu9X13DPdyM94wmY7G8Mux7DjxXqmPpKw2c7GsMsx7Hix3rL2ZMJmOxvDLsew423AZjsMw27DsPMMxRe1YTsDwy6R+tUnpTZsZ2DYJVK/+qTUhu0MBLsEZTytofiOfbspYbOdjWGXY9jxYn3g3O0Jm+1sDLscw45n/9bFWcdmOxvBLkew41GpDcuMDFmSIWNu0IZlRoYsyZAxqdSGZUaGLMmQMan8VpahGLIVQeaiEptldoIs73xTpkNsltkJsjyVGR+xWWYnyPJUZnzEZhmWINtSmXkH5YvasMzIkCUJMiaV2rDMSJAlCTLmBm1YZiTIkgQZk4bhi9gssxNk+brM+IjNMjtBlq/LjI/YLLMTZPm6zPj/A1hg9+k=
```
bangerth: This is funny -- why does your patch change a .vtu file?

masterleinad: The test passes without modifications to the output file when `deal.II` is configured without `ZLIB` support. Hence, there should only be a small difference that results in a different output here.

kronbichler: The fact that this changes by going from `a/b` to `a * (1/b)` shows why compilers do not do this optimization automatically: It changes in floating point arithmetic (even though only slightly in the last one or two bits for most reasonable numbers).

include/deal.II/base/complex_overloads.h
masterleinad: It actually is: https://github.com/dealii/dealii/blob/master/include/deal.II/base/template_constraints.h#L567-L571, but 
```
operator/(const std::complex<T>&, const std::complex<U>&)
```
is a better match: https://godbolt.org/z/W4ikrT

bangerth: I see, thank you!

masterleinad:
@bangerth Is there anything else you want me to address?

PR #8432, opened 2019-07-29
Implement output operator for VectorizedArray 

masterleinad:
It always annoyed me that it is hard to view the content of a `VectorizedArray` object and I ended up writing the same simple code for it over and over. Let's just define an output operator in the library and be done with it.

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -3841,9 +3841,24 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
   return VectorizedArray<Number, width>() - u;
 }
 
+/**
+ * Output operator for vectorized array.
+ *
+ * @relatesalso VectorizedArray
+ */
+template <typename Number, int width>
+inline std::ostream &
+operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
+{
+  constexpr unsigned int n = VectorizedArray<Number, width>::n_array_elements;
+  for (unsigned int i = 0; i < n - 1; ++i)
+    out << p[i] << ' ';
+  out << p[n - 1];
+
+  return out;
+}
 
 DEAL_II_NAMESPACE_CLOSE
-
 
 /**
  * Implementation of functions from cmath on VectorizedArray. These functions
--- a/doc/news/changes/minor/20190729DanielArndt
+++ b/doc/news/changes/minor/20190729DanielArndt
@@ -1,0 +1,4 @@
+New: The content of a VectorizedArray object can be printed
+using the output operator.
+<br>
+(Daniel Arndt, 2019/07/29)
--- a/tests/base/vectorization_12.cc
+++ b/tests/base/vectorization_12.cc
@@ -1,0 +1,65 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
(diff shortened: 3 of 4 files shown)


peterrum:
This is indeed useful! Thanks @masterleinad !

include/deal.II/base/vectorization.h
```diff
@@ -3841,9 +3841,25 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
...
 
+/**
+ * Output operator for vectorized array.
+ *
+ * @relatesalso VectorizedArray
+ */
+template <typename Number>
+inline std::ostream &
+operator<<(std::ostream &out, const dealii::VectorizedArray<Number> &p)
+{
```
peterrum: Please generalize it, such that it works for arbitrary `VectorizedArray<Number, width>`:
```cpp
inline std::ostream &
operator<<(std::ostream &out, const dealii::VectorizedArray<Number, width> &p)
```

include/deal.II/base/vectorization.h
```diff
@@ -3841,9 +3841,25 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
...
 }
 
+/**
+ * Output operator for vectorized array.
+ *
+ * @relatesalso VectorizedArray
+ */
+template <typename Number>
+inline std::ostream &
+operator<<(std::ostream &out, const dealii::VectorizedArray<Number> &p)
```
peterrum: When is `operator<<` and `print()` used in deal.II? 

I find the `print` methods quite useful because you can add additional parameters like formatting. Furthermore, it would be quite useful if we could specify which/how many entries we want to print. E.g. `Matrixfree::n_active_entries_per_cell_batch()` and  `Matrixfree::n_active_entries_per_face_batch()` provides us the information about how many lanes are actually relevant.

peterrum: I can implement it in a follow up PR if want this feature.

include/deal.II/base/vectorization.h
```diff
@@ -3841,9 +3841,25 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArray<Number, width>
...
+{
+  for (unsigned int i = 0;
+       i < dealii::VectorizedArray<Number>::n_array_elements;
+       ++i)
+    out << p[i] << ' ';
+  out << '\n';
```
kronbichler: The question is whether we want to skip the newline for consistency with the output in the `Point` and `Tensor` classes, see e.g.
https://github.com/dealii/dealii/blob/master/include/deal.II/base/point.h#L669-L673
While I would agree that 16 floats definitely fill a line, I think it is better to let the user decide how to wrap things.

masterleinad: Sure, that's fine with me.

include/deal.II/base/vectorization.h
masterleinad: My motivation here was to also be able to print the content where `VectorizedArray` is a nested type, e.g. `Tensor<1, dim, VectorizedArray<double>>`. In also see, `VectorizedArray` more like a `Tensor` object with a fixed and rather small number of entries (where we provide `operator<<`) than like a matrix or a vector (where we typically have a `print()` interface).

masterleinad:
I addressed both comments.

PR #8434, opened 2019-07-30
Doc: Reviewed documentation for hp::Refinement.

marcfehling:
The `KellyErrorEstimator` returns each cell's actual local error estimate, and not its squared value.
https://github.com/dealii/dealii/blob/732cd1b9c7594c14af36309fd3933d07676b9aa5/include/deal.II/numerics/error_estimator.templates.h#L1370

We actually considered this during the preparation for the `hp` decision algorithms, but somehow left the documentation at an earlier stage (where I was still figuring out what Melenk et al. did in their publication and adapted their notation). This PR corrects this issue.

### DIFF
--- a/include/deal.II/hp/refinement.h
+++ b/include/deal.II/hp/refinement.h
@@ -40,16 +40,16 @@ namespace hp
    * complexity of the numerical solution. Error estimates are an appropriate
    * means of determining where adjustments need to be made.
    *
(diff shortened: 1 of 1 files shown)


marcfehling:
I just realized there were some more issues with the documentation which I fixed on the fly.

Plus, any occurrence of the `h` or `p` suffix is now connected to the corresponding word by hyphen in that namespace.

PR #8436, opened 2019-07-30
Add a note to the count_lines.sh script.

bangerth:


drwells:
I hope no one lost data here.

rezarastak:
For counting the lines in one of my git projects, I found the utility [cloc](https://github.com/AlDanial/cloc) very useful because it distinguishes lines of code versus blank lines and comment lines. We can think about using `cloc` instead of `wc` in this script (although I love the simplicity of `wc`). `cloc` can be easily installed from ubuntu repositories so we don't need to bundle it.

rezarastak:
Oh I seems that you have discussed about it before in #6959. Never mind then :) 
I still think that separating the number of lines of code and lines of documentation(~ lines of comment) would be informative.

bangerth:
I think using `cloc` could be useful, though of course much more expensive. If you wanted to give it a try to use that, we could make that an option in the script!

Issue #5982, opened 2018-03-01
In the tutorials, get rid of EPS output


bangerth:

bangerth:
Oh, hit ENTER to fast.

Rationale: EPS is sort of an outdated format to begin with. In addition, we don't support it in 1d, and you can't rotate it in 3d. I'd much rather we don't advertise this as good practice, but instead output as VTK or gnuplot.

kronbichler:
I cannot agree more, I think we really should focus on teaching people to use the powerful tools with vtk (ParaView, VisIt). I think all other formats/tools are for advanced users with particular workflows, the basic user is fine with the former.

bangerth:
We still have this:
```
step-12/step-12.cc:      grid_out.write_eps(triangulation, eps_output);
step-14/step-14.cc:      GridOut().write_eps(dof_handler.get_triangulation(), out);
step-1/step-1.cc:  grid_out.write_eps(triangulation, out);
step-1/step-1.cc:  grid_out.write_eps(triangulation, out);
step-30/step-30.cc:    grid_out.write_eps(triangulation, eps_output);
step-49/step-49.cc:  grid_out.write_eps(triangulation, out);
step-5/step-5.cc:  data_out.write_eps(output);
step-6/step-6.cc:    grid_out.write_eps(triangulation, output);
```

krishnakumarg1984:
This does not seem to be completely agreeable to me. For instance, it looks like the `write_svg` function (SVG writer) does not know how to print curved cells and can only display quadrilaterals.
See #9296 .

For a newcomer to the library, this is off-putting when the visualization does not produce what you expect. In the case of a `hyper_ball` or `hyper_shell`, we need to see a curved outer boundary for sure.

bangerth:
Well, this bug is simply about the fact that EPS is an outdated file format we should not be using any more.

As for your statement above: It is not clear what is correct. We use a manifold object that describes the outer boundary as the (curved) perimeter of a circle. But when you do computations, you generally use a mapping and then the actual domain you are solving on has a piecewise polynomial boundary (unless you use special mappings). So, for example, if you use a Q1 mapping, then the domain you are solving on really *is* one with piecewise straight boundaries. As a consequence, it is not clear to me that what you see is actually *different* than what you are doing. 

As for `write_svg` not supporting curved boundaries: If you take a look at what `write_eps` does, then it shouldn't be very difficult to implement the same in `write_svg`.

bangerth:
[...]

PR #8447, opened 2019-08-04
Let step-1 output SVG instead of EPS.

bangerth:
EPS is just an outdated format.

I've already put these pictures online, but they look like this:
step-1 grid-1-r9 2
step-1 grid-2r2

### DIFF
--- a/doc/news/changes/minor/20190803Bangerth-1
+++ b/doc/news/changes/minor/20190803Bangerth-1
@@ -1,0 +1,4 @@
+Changed: The step-1 program no longer uses EPS to output the meshes,
+but the far more modern SVG format.
+<br>
+(Wolfgang Bangerth, 2019/08/03)
--- a/examples/step-1/doc/results.dox
+++ b/examples/step-1/doc/results.dox
@@ -1,27 +1,37 @@
 <h1>Results</h1>
 
-Running the program produces graphics of two grids (grid-1.eps and grid-2.eps). They look like this:
+Running the program produces graphics of two grids (grid-1.svg and grid-2.svg). 
+You can open these with most every web browser -- in the simplest case,
+just open the current directory in your file system explorer and click
+on the file. If you like working on the command line, you call your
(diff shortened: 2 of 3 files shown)


bangerth:
Fail. The second picture should have been this one:
step-1 grid-2-r9 2

tjhei:
While I agree that eps is a terrible format to output into by default, I am also not sure if the default settings for svg are very useful. Especially the cell index might be confusing?

bangerth:
Would it be more useful if I defaulted the SVG writer to not output cell numbers?

tjhei:
> Would it be more useful if I defaulted the SVG writer to not output cell numbers?

What is your opinion? We can also leave it like you proposed with the paragraph you wrote about the SVG settings.

bangerth:
I would remove the cell numbers from the plot but leave the colors. In fact, I find the cell numbers not useful as a default.

There are currently two pending patches for the SVG writer. If you're willing to merge this, then I'll remove the cell number output in a follow-up patch that changes the default.

bangerth:
Part of #5982.

tjhei:
The documentation you wrote as part of this PR reference the "cell number" though. Do you still want this PR to be merged as is?

bangerth:
Yes. Let's keep things in sync.

peterrum:
@bangerth: I have the feeling that the figures here and in PR#8448 are too big. When I open the tutorials of developer version, I cannot see 10% on the right in my browser (100% zoom; 1920 x 1080 monitor) and have to scroll right. I think the sizes as shown here on github would be better...

bangerth:
@peterrum -- good point. Is this better?

krishnakumarg1984:
@bangerth 
This does not seem to be completely agreeable to me. For instance, it looks like the `write_svg` function (SVG writer) does not know how to print curved cells and can only display quadrilaterals.
See #9296.

For a newcomer to the library, this is off-putting when the visualization does not produce what you expect. In the case of a `hyper_ball` or `hyper_shell`, the user needs to see a curved outer boundary for sure.

PR #8448, opened 2019-08-04
Use VTU format in step-5.

bangerth:
Same as before. These are the pictures:
step-5 solution-0-r9 2
step-5 solution-1-r9 2
step-5 solution-3-r9 2
step-5 solution-4-r9 2
step-5 solution-5-r9 2

I used a better color scale while there.

### DIFF
--- a/doc/news/changes/minor/20190803Bangerth-2
+++ b/doc/news/changes/minor/20190803Bangerth-2
@@ -1,0 +1,4 @@
+Changed: The step-5 program no longer uses EPS to output the solutions,
+but the far more modern VTK/VTU format.
+<br>
+(Wolfgang Bangerth, 2019/08/03)
--- a/examples/step-5/doc/results.dox
+++ b/examples/step-5/doc/results.dox
@@ -39,34 +39,32 @@ Cycle 5:
 
 In each cycle, the number of cells quadruples and the number of CG
 iterations roughly doubles.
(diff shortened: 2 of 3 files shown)


jppelteret:
Thanks for modernising the output in these early tutorials! With the grid lines in place I feel like the centre region of the most discretised solution end up looking very dark, at first glance suggesting a solution ~0. But then again, the original images had the same problem... Have you by any chance tried some of the other colour schemes (e.g. blue-white-red) to see if they help make the solution in the central region "pop" a bit more?

jppelteret:
I do recognise though that its probably going to be very difficult to win here, so if you say that this is the best visualisation that you can achieve then I'm happy for us to go with this.

bangerth:
I tried to go for one of the color scales that are not "divergent" because the solution is strictly positive. A blue-white-red color scale is divergent and so is best used for cases where you have both positive and negative values. 

I recognize that the pictures are maybe not perfect, but I feel like you'd get something similar for every non-divergent color scale. I do think the pictures are vastly better than what's here: https://www.dealii.org/developer/doxygen/deal.II/step_5.html#Results

jppelteret:
Cool, makes sense. And I agree that these pics are a great improvement on those which are going to be replaced!

bangerth:
Part of #5982.

PR #8451, opened 2019-08-04
Slight restructure the SVG grid output.

bangerth:
In particular:
* Make the svg_project_point() a function in an anonymous namespace
  in the .cc file, rather than a static member function of the class.
* Make those of its arguments that represent directions into tensors,
  rather than Points.
* Move a couple of variable declarations next to their initialization.

As far as functionality is concerned, represent cells by
'level.index' rather than 'level,index', since the former is what you
get when you pass a cell to a std::ostream. Also just use 'level'
in the SVG caption, rather than 'level_number'.

### DIFF
--- a/include/deal.II/grid/grid_out.h
+++ b/include/deal.II/grid/grid_out.h
@@ -1724,22 +1724,6 @@ private:
 
 
   /**
-   * This function projects a three-dimensional point (Point<3> point) onto a
-   * two-dimensional image plane, specified by the position of the camera
-   * viewing system (Point<3> camera_position), camera direction (Point<3>
-   * camera_position), camera horizontal (Point<3> camera_horizontal,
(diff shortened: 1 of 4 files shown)


bangerth:
I should mention that in order to make the `svg_project_point()` function a local one in an anonymous namespace, I had to also move it above the first use. This obscures the rest of the diffs by a significant degree :-(

bangerth:
Ah, I only checked every test that had `svg` in its name. Among the ones you have, only `grid/get_finest_common_cells_04.mpirun=2.debug` fails on my machine, but that's because I don't have Zoltan. So I *think* it should all be ok.

tjhei:
> So I _think_ it should all be ok.

I agree. That test doesn't print the output.

PR #8458, opened 2019-08-06
Fix for step-20 tutorial: changed point subtraction to tensor distance

kirana-bergstrom:
Subtraction of points is now a tensor instead of a point, points do not have a "square" function. Changed to use distance for tensors instead.

### DIFF
--- a/examples/step-20/doc/results.dox
+++ b/examples/step-20/doc/results.dox
@@ -265,8 +265,7 @@ KInverse<dim>::value_list (const std::vector<Point<dim> > &points,
 
       double permeability = 0;
       for (unsigned int i=0; i<centers.size(); ++i)
-        permeability += std::exp(-(points[p]-centers[i]).square()
-                                 / (0.1 * 0.1));
+        permeability += std::exp(-(points[p] - centers[i]).norm_square() / (0.1 * 0.1));
 
       const double normalized_permeability
         = std::max(permeability, 0.005);


examples/step-20/doc/results.dox
```diff
@@ -265,8 +265,8 @@ KInverse<dim>::value_list (const std::vector<Point<dim> > &points,
 
       double permeability = 0;
       for (unsigned int i=0; i<centers.size(); ++i)
-        permeability += std::exp(-(points[p]-centers[i]).square()
-                                 / (0.1 * 0.1));
```
masterleinad: Thanks for spotting this! I guess you are missing some braces, though. Currently, updating `permeability` is not inside the body of the for-loop.
Alternatively, you could probably also use `Tensor::norm_square()`.

bangerth:
@kirana-bergstrom  -- nice job figuring this out, and great first pull request! If you need help with @masterleinad 's comment, please find any of us to talk it through!

examples/step-20/doc/results.dox
kirana-bergstrom: ### Thank you! Sorry for the missing braces - bad merge on my part, thank you for catching it!

bangerth:
It's in a doc file, so merging -- nice job!

bangerth:
@kirana-bergstrom: Would you mind creating a follow-up pull request in which you put an entry into the changelog? They are in `doc/news/changes/minor` -- just copy an existing file and edit as appropriate!

PR #8471, opened 2019-08-06
Null pointer check added to CellDagaStorage::get_data()

rezarastak:
When we initialize the cell history data using 
```c++
cell_data_storage.template initialize<T>(cell, n_data_points_per_cell)
```
and retrieve the data using
```c++
std::vector<std::shared_ptr<T>> data_vector = cell_data_storage.template get_data<T>(cell);
```
The type T used to initialize the cell data must be the same T that is used to retrieve it. If there is a type mismatch, `get_data` returns a vector of null pointers and it leads to catastrophic segfault when it is dereferenced.

I added an assert statement that checks this conditions and provides an understandable error message.


bangerth:
Do you think you could add an entry to `doc/news/changes/minor` that explains your change?

PR #8472, opened 2019-08-06
Change all uses of MappingType and mapping_type

GrahamBenHarper:
I've heard the use of MappingType in mapping.h is a little confusing since it's not actually a type.

To fix that, I've changed all instances of MappingType to MappingKind, all instances of mapping_type to mapping_kind, and then updated as much documentation as I could find to accordingly match. Please let me know if I missed anything.

I'm running a build on my end now to verify nothing was broken but it may take awhile on my laptop :cry:

### DIFF
--- a/doc/news/changes/incompatibilities/20190806GrahamHarper
+++ b/doc/news/changes/incompatibilities/20190806GrahamHarper
@@ -1,0 +1,7 @@
+Changed: mapping.h referred to different kinds of mappings in an enum 
+MappingType; however, since it's an enum, MappingType is a misleading name. 
+All instances of MappingType and mapping_type have been changed to 
+MappingKind and mapping_kind, respectively. The documentation has been 
+updated in order to reflect these changes as well.
+<br>
+(Graham Harper, 2019/08/06)
--- a/include/deal.II/fe/fe.h
+++ b/include/deal.II/fe/fe.h
@@ -2289,7 +2289,7 @@ public:
    *   the node functionals of the element applied to the given function.
    *
    * @note It is safe to call this function for (transformed) values on the
-   * real cell only for elements with trivial MappingType. For all other
+   * real cell only for elements with trivial MappingKind. For all other
    * elements (for example for H(curl), or H(div) conforming elements)
    * vector values have to be transformed to the reference cell first.
    *
--- a/include/deal.II/fe/fe_dg_vector.h
+++ b/include/deal.II/fe/fe_dg_vector.h
@@ -58,7 +58,7 @@ public:
   /**
    * Constructor for the vector element of degree @p p.
    */
-  FE_DGVector(const unsigned int p, MappingType m);
+  FE_DGVector(const unsigned int p, MappingKind m);
 
   /**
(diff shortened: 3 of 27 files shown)


bangerth:
Do you think you could add an entry to `doc/news/changes/minor` that explains your change?

GrahamBenHarper:
I have added a changelog entry.

Also, I'd like feedback: the file `bundled/boost-1.70.0/include/boost/geometry/strategies/geographic/mapping_ssf.hpp` has its own enum named `mapping_type`. I did not modify it because it doesn't use camel case, which is used for types, but I can change it also if you'd like.

tjhei:
> I did not modify it because it doesn't use camel case, which is used for types, but I can change it also if you'd like.

This should stay as is.  We do not develop/improve things in ``bundled/`` (these are basically fixed, imported versions).

masterleinad:
Is this backward-compatible?

GrahamBenHarper:
Backward compatible in which sense?

I only renamed the enum and all references to the enum, so I imagine it should be.

bangerth:
It's not backward compatible: The `enum MappingType/Kind` is a public one in the mapping and finite element classes. At the same time, it's only used internally in these kinds of classes, so I suspect that the likelihood of anyone using the type in user classes is, essentially, zero.

@GrahamBenHarper -- can you move the changelog entry to the `doc/news/changes/incompatible` directory?

GrahamBenHarper:
Oh I understand what you mean now! I never would have thought of that.

The changelog entry should be in the right place now :)

doc/news/changes/incompatibilities/20190806GrahamHarper
```diff
@@ -0,0 +1,7 @@
+Changed: Mapping.h referred to different kinds of mappings in an enum
```
masterleinad:
```suggestion
Changed: mapping.h referred to different kinds of mappings in an enum 
```

doc/news/changes/minor/20190806GrahamHarper
```diff
@@ -0,0 +1,3 @@
+Changed: Mapping.h referred to different kinds of mappings in an enum MappingType; however, since it's an enum, MappingType is a misleading name. All instances of MappingType and mapping_type have been changed to MappingKind and mapping_kind, respectively. The documentation has been updated in order to reflect these changes as well.
+<br>
+(Graham Harper, 2019/08/06)
```
masterleinad: Just delete this file.

GrahamBenHarper:
Whoops!

I also took the liberty of squashing those multiple egregious commits. The changelog entry should be in the right place now.

doc/news/changes/incompatibilities/20190806GrahamHarper
GrahamBenHarper: Got the typo!

doc/news/changes/minor/20190806GrahamHarper
GrahamBenHarper: Deleted!

PR #8473, opened 2019-08-06
Avoid deprecated features in step-54.

drwells:
We renamed these classes recently and deprecated the names ending in `Boundary`. I noticed this during @luca-heltai's talk.


bangerth:
Looks good to me, but I'll let @luca-heltai pull the trigger.

/rebuild

PR #8474, opened 2019-08-06
Cleanup use of DEAL_II_CONSTEXPR

masterleinad:
This pull request tries to test all the `Tensor` and `SymmetricTensor` functions we added `DEAL_II_CONSTEXPR` annotations to and removes it for functions I couldn't get working in a `constexpr` context.

@rezarastak FYI


tests/tensors/constexpr_symmetric_tensor.cc
```diff
@@ -111,6 +111,37 @@ main()
...
+  }
+  {
+    DEAL_II_CONSTEXPR double a_init[3][3] = {{1., 2., 3.},
+                                             {2., 1., 2.},
+                                             {3., 2., 1.}};
+    DEAL_II_CONSTEXPR Tensor<2, 3> dummy_a{a_init};
+    DEAL_II_CONSTEXPR SymmetricTensor<2, 3> a          = symmetrize(dummy_a);
+    DEAL_II_CONSTEXPR auto                  transposed = transpose(a);
+    Assert(transposed == a, ExcInternalError());
+    constexpr auto dummy   = scalar_product(a, a);
```
rezarastak:
```suggestion
    DEAL_II_CONSTEXPR const auto dummy   = scalar_product(a, a);
```
We need to make sure these lines work within pre-c++14 compilers. For example, in c++11, `a` is not constexpr and this line fails. There are other similar lines that we should fix before merging this.

tests/tensors/constexpr_symmetric_tensor.cc
```diff
@@ -111,6 +111,37 @@ main()
   DEAL_II_CONSTEXPR const auto B = get_tensor_4();
   deallog << "SymmetricTensor<4,2> = " << B << std::endl;
 
+  {
+    DEAL_II_CONSTEXPR double a_init[3][3] = {{1., 0., 0.},
```
rezarastak:
```suggestion
    constexpr double a_init[3][3] = {{1., 0., 0.},
```

tests/tensors/constexpr_symmetric_tensor.cc
```diff
@@ -111,6 +111,37 @@ main()
...
+                                             {2., 1., 0.},
+                                             {3., 2., 1.}};
+    DEAL_II_CONSTEXPR Tensor<2, 3> dummy_a{a_init};
```
rezarastak:
```suggestion
    constexpr Tensor<2, 3> dummy_a{a_init};
```
I believe that all of the constructors of `Tensor` are `constexpr` and compatible with `c++11`

tests/tensors/constexpr_symmetric_tensor.cc
```diff
@@ -111,6 +111,37 @@ main()
...
+    DEAL_II_CONSTEXPR SymmetricTensor<2, 3> a = symmetrize(dummy_a);
```
rezarastak:
```suggestion
    DEAL_II_CONSTEXPR const SymmetricTensor<2, 3> a = symmetrize(dummy_a);
```
Using `const` here is not very critical but it's nice to write. It enforces const-ness even when `DEAL_II_CONSTEXPR` is empty.

tests/tensors/constexpr_tensor.cc
```diff
@@ -21,15 +21,24 @@
 
 template <int rank, int dim, typename Number>
 void
-test_consexpr_tensor_constructors()
+test_constexpr_tensor_constructors()
```
rezarastak: Thanks for fixing the type.

include/deal.II/base/symmetric_tensor.h
```diff
@@ -593,10 +593,12 @@ class SymmetricTensor
    * is symmetric only up to round-off, then you may want to call the
    * <tt>symmetrize</tt> function first. If you aren't sure, it is good
    * practice to check before calling <tt>symmetrize</tt>.
+   *
+   * If you want to use this function in a `constexpr` context,
+   * use symmetrize() instead.
```
bangerth: Would you mind elaborating why this is so? I suspect saying that "Because conversion constructors are not allowed in `constexpr` expressions, you will have to use the `symmetrize()` function in such contexts." would be enough. (I'm assuming that this is the reason.)

include/deal.II/base/tensor.h
```diff
@@ -322,17 +322,17 @@ class Tensor<0, dim, Number>
...
   /**
    * Return the square of the Frobenius-norm of a tensor, i.e. the sum of the
    * absolute squares of all entries.
    *
    * @note This function can also be used in CUDA device code.
    */
-  DEAL_II_CUDA_HOST_DEV real_type
-                        norm_square() const;
+  DEAL_II_CONSTEXPR DEAL_II_CUDA_HOST_DEV real_type
+                                          norm_square() const;
```
bangerth: What might be the reason that the `norm()` function can't be `constexpr`, but the `norm_square()` function can be? It seems a bit difficult at the moment to see the consistency in this `constexpr` business, and that will make it difficult to maintain all of this going forward :-(

include/deal.II/base/tensor.h
```diff
@@ -1008,7 +1006,7 @@ DEAL_II_CONSTEXPR inline DEAL_II_ALWAYS_INLINE
 
 
 template <int dim, typename Number>
-DEAL_II_CUDA_HOST_DEV inline DEAL_II_ALWAYS_INLINE
+DEAL_II_CONSTEXPR DEAL_II_CUDA_HOST_DEV inline
```
bangerth: Is the removal of the `always_inline` attribute going to upset anything here?

include/deal.II/base/tensor.h
```diff
@@ -1235,8 +1233,9 @@ template <typename OtherNumber>
 DEAL_II_CONSTEXPR inline DEAL_II_ALWAYS_INLINE Tensor<rank_, dim, Number> &
 Tensor<rank_, dim, Number>::operator=(const Tensor<rank_, dim, OtherNumber> &t)
 {
-  if (dim > 0)
-    std::copy(&t.values[0], &t.values[0] + dim, &values[0]);
+  // std::copy is constexpr only from C++20 on.
+  for (unsigned int i = 0; i < dim; ++i)
+    values[i] = t.values[i];
```
bangerth: The comment is going to be hard to read without the context of this particular patch. The comment doesn't reference anything in the code, but rather something that's not there. I think you might want to say something like this: "The following loop could be written more concisely using `std::copy`, but that function is only `constexpr` from C++20 on."

include/deal.II/base/tensor.h
```diff
@@ -1432,12 +1431,12 @@ namespace internal
   inline unsigned int
   mod<0>(const unsigned int x)
   {
-    Assert(false, ExcInternalError());
+    // Assert(false, ExcInternalError());
     return x;
```
bangerth: Nah, I think you want to leave this in.

include/deal.II/base/tensor.h
kronbichler: I don't think it's critical here, except that the function is cheap enough that it would not hurt.

include/deal.II/base/tensor.h
masterleinad: `std::sqrt` and `std::abs` are not required to be `constexpr`.

include/deal.II/base/tensor.h
masterleinad: That's a rebase artifact. `DEAL_II_CONSTEXPR` should be here.

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+    Tensor<2, 3>::unrolled_to_component_indices(0);
+  DEAL_II_CONSTEXPR const auto index =
+    Tensor<2, 3>::component_to_unrolled_index(TableIndices<2>{});
+  Assert(index == 0, ExcInternalError());
+
+  DEAL_II_CONSTEXPR const auto used_memory = Tensor<2, 3>::memory_consumption();
+  deallog << "Used memory: " << used_memory << std::endl;
+
+  {
+    DEAL_II_CONSTEXPR const double a_init[3][3] = {{1., 0., 0.},
```
rezarastak:
```suggestion
    constexpr double a_init[3][3] = {{1., 0., 0.},
```

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+                                                   {2., 1., 0.},
+                                                   {3., 2., 1.}};
+    DEAL_II_CONSTEXPR const Tensor<2, 3> a{a_init};
```
rezarastak:
```suggestion
    constexpr Tensor<2, 3> a{a_init};
```

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+    DEAL_II_CONSTEXPR const auto         inverted       = invert(a);
+    DEAL_II_CONSTEXPR const double       ref_init[3][3] = {{1., 0., 0.},
```
rezarastak:
```suggestion
    constexpr double       ref_init[3][3] = {{1., 0., 0.},
```

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+                                                     {-2., 1., 0.},
+                                                     {1., -2., 1}};
+    DEAL_II_CONSTEXPR const Tensor<2, 3> ref{ref_init};
```
rezarastak:
```suggestion
    constexpr Tensor<2, 3> ref{ref_init};
```

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+    Assert(inverted == ref, ExcInternalError());
+  }
+  {
+    DEAL_II_CONSTEXPR const double a_init[3][3] = {{1., 0., 0.},
```
rezarastak:
```suggestion
    constexpr double a_init[3][3] = {{1., 0., 0.},
```

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+                                                   {2., 1., 0.},
+                                                   {3., 2., 1.}};
+    DEAL_II_CONSTEXPR const Tensor<2, 3> a{a_init};
+    DEAL_II_CONSTEXPR const auto         transposed     = transpose(a);
+    DEAL_II_CONSTEXPR const double       ref_init[3][3] = {{1., 2., 3.},
```
rezarastak:
```suggestion
    constexpr double       ref_init[3][3] = {{1., 2., 3.},
```

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+                                                     {-2., 1., 0.},
+                                                     {1., -2., 1}};
+    DEAL_II_CONSTEXPR const Tensor<2, 3> ref{ref_init};
+    Assert(inverted == ref, ExcInternalError());
+  }
+  {
+    DEAL_II_CONSTEXPR const double a_init[3][3] = {{1., 0., 0.},
+                                                   {2., 1., 0.},
+                                                   {3., 2., 1.}};
+    DEAL_II_CONSTEXPR const Tensor<2, 3> a{a_init};
```
rezarastak:
```suggestion
    constexpr Tensor<2, 3> a{a_init};
```

tests/tensors/constexpr_tensor.cc
```diff
@@ -90,6 +118,75 @@ main()
...
+    DEAL_II_CONSTEXPR const auto         transposed     = transpose(a);
+    DEAL_II_CONSTEXPR const double       ref_init[3][3] = {{1., 2., 3.},
+                                                     {0., 1., 2.},
+                                                     {0., 0., 1}};
+    DEAL_II_CONSTEXPR const Tensor<2, 3> ref{ref_init};
```
rezarastak:
```suggestion
    constexpr Tensor<2, 3> ref{ref_init};
```

masterleinad:
@bangerth Are you happy with the explanations and updates to the documentation here?

masterleinad:
Thanks!

PR #8477, opened 2019-08-06
Fixed markdown in a step-1 tutorial comment.

mnkajan:
Most function names in the comments of step 1 are written with backticks so that they appear in a typewriter font on the tutorial page, but one line uses a different format. I'm pretty sure this is an unintentional inconsistency.


bangerth:
Good find and nice first patch! Can you add an entry to `doc/news/changes/minor` in a follow-up pull request?

mnkajan:
@bangerth Will do!

PR #8481, opened 2019-08-07
Rearrange measure

drwells:
I found this while looking at something else (a second PR is on its way; I just want to check a few more things).

### DIFF
--- a/source/grid/tria_accessor.cc
+++ b/source/grid/tria_accessor.cc
@@ -1264,52 +1264,12 @@ namespace
   double
   measure(const TriaAccessor<2, 2, 2> &accessor)
   {
-    // the evaluation of the formulae
-    // is a bit tricky when done dimension
-    // independently, so we write this function
-    // for 2D and 3D separately
-    /*
-      Get the computation of the measure by this little Maple script. We
-      use the blinear mapping of the unit quad to the real quad. However,
-      every transformation mapping the unit faces to straight lines should
-      do.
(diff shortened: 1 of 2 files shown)


source/grid/tria_accessor.cc
```diff
@@ -1264,52 +1264,12 @@ namespace
...
-                         accessor.vertex(1)(0),
-                         accessor.vertex(2)(0),
-                         accessor.vertex(3)(0)};
-    const double y[4] = {accessor.vertex(0)(1),
-                         accessor.vertex(1)(1),
-                         accessor.vertex(2)(1),
-                         accessor.vertex(3)(1)};
+    unsigned int vertex_indices[GeometryInfo<2>::vertices_per_cell];
+    for (unsigned int i = 0; i < GeometryInfo<2>::vertices_per_cell; ++i)
+      vertex_indices[i] = accessor.vertex_index(i);
```
masterleinad: Can't you use this implementation dimension-independent?

drwells: I am not sure how to do this since we have a total of five template definitions and we already have a catch-all case.

masterleinad: Oh, I see. We have to treat the codimension one case special.

masterleinad:
You still need to run the indentation script. :wink:

drwells:
Ooops. Fixed.

PR #8484, opened 2019-08-07
Cell id is the key

rezarastak:
CellDataStorage() uses `CellIteratorType` as the key of its internal `std::map` data structure. The issue is that when local refinement and coarsening is applied, the identity of cell iterators may change. The data assigned to a cell may be wrongfully assigned to a different cell.

To demonstrate that the current master branch contains this error, a test is added. We use two different materials with two different stored data structures. A simple refinement and coarsening step and a subsequent reinitialization of cell data structures introduces the error. Basically instead of initializing a data structures for newly refined cells, it assumes that these already exist and no initialization is performed. Now if we try to retrieve this data using the `get_data()` method, the exception `ExcCellDataTypeMismatch` is thrown (I proposed #8471 to appropriately present this obscure error). Note that this error is only thrown when we have two different material ids and two material data structures. Last week I found this bug and it took me around 12-16 hours to track the source of it because my code would randomly fail after 10 levels of refinement with a segfault.

To fix this error we change the key of the private data map to `CellId` because it uniquely represents cells even after refinement and coarsening steps.

@peterrum helped me considerably to reproduce this bug in a test and to solve it.

### DIFF
--- a/doc/news/changes/minor/20190807RezaRastakPeterMunch
+++ b/doc/news/changes/minor/20190807RezaRastakPeterMunch
@@ -1,0 +1,4 @@
+Fixed: Bug in CellDataStorage that produced error when applying adaptive
+refinement in multi-material grids.
+<br>
+(Reza Rastak, Peter Munch, 2019/08/07)
--- a/include/deal.II/base/quadrature_point_data.h
+++ b/include/deal.II/base/quadrature_point_data.h
@@ -91,6 +91,10 @@ public:
    * may reflect, for example, different constitutive models of continuum
    * mechanics in different parts of the domain.
    *
+   * @note The first time this method is called, it stores a SmartPointer to the
+   * Triangulation object that owns the cell. The future invocations of this
+   * method expects the cell to be from the same stored triangulation.
+   *
    * @pre The type @p T needs to either equal @p DataType, or be a class derived
    * from @p DataType. @p T needs to be default constructible.
    */
@@ -138,6 +142,9 @@ public:
    * cell-by-cell basis.
    *
    * @pre The type @p T needs to match the class provided to initialize() .
+   *
+   * @pre @p cell must be from the same Triangulation that is used to
+   * initialize() the cell data.
    */
   template <typename T = DataType>
   std::vector<std::shared_ptr<T>>
@@ -153,6 +160,9 @@ public:
    * cell-by-cell basis.
    *
    * @pre The type @p T needs to match the class provided to initialize() .
+   *
+   * @pre @p cell must be from the same Triangulation that is used to
+   * initialize() the cell data.
    */
   template <typename T = DataType>
   std::vector<std::shared_ptr<const T>>
@@ -160,31 +170,32 @@ public:
 
 private:
   /**
-   * Depending on @p CellIteratorType we need to define a triangulation type
+   * Number of dimensions
    */
-  using triangulation_type =
-    Triangulation<CellIteratorType::AccessorType::dimension,
-                  CellIteratorType::AccessorType::space_dimension>;
+  static constexpr unsigned int dimension =
+    CellIteratorType::AccessorType::dimension;
 
   /**
-   * The key is used as identified for a cell data structure.
-   * We need to use CellId as part of the key because it remains unique during
-   * adaptive refinement. We need to use a pointer to the triangulation in the
-   * key because we want to allow the data storage class to keep information on
-   * multiple meshes simultaneously.
+   * Number of space dimensions
    */
-  using key_type = std::pair<const triangulation_type *, CellId>;
+  static constexpr unsigned int space_dimension =
+    CellIteratorType::AccessorType::space_dimension;
+
+  /**
+   * To ensure that all the cells in the CellDataStorage come from the same
+   * Triangulation, we need to store a reference to that Triangulation within
+   * the class.
+   */
+  SmartPointer<const Triangulation<dimension, space_dimension>,
+               CellDataStorage<CellIteratorType, DataType>>
+    tria;
 
   /**
    * A map to store a vector of data on each cell.
+   * We need to use CellId as the key because it remains unique during
+   * adaptive refinement.
    */
-  std::map<key_type, std::vector<std::shared_ptr<DataType>>> map;
-
-  /**
-   * Computes the key from a cell iterator
-   */
-  static key_type
-  get_key(const CellIteratorType &cell);
+  std::map<CellId, std::vector<std::shared_ptr<DataType>>> map;
 
   /**
    * @addtogroup Exceptions
@@ -192,6 +203,13 @@ private:
   DeclExceptionMsg(
     ExcCellDataTypeMismatch,
     "Cell data is being retrieved with a type which is different than the type used to initialize it");
+
+  /**
+   * @addtogroup Exceptions
+   */
+  DeclExceptionMsg(
+    ExcTriangulationMismatch,
+    "The provided cell iterator does not belong to the triangulation that corresponds to the CellDataStorage object.");
 };
 
 
@@ -545,14 +563,6 @@ namespace parallel
 //--------------------------------------------------------------------
 
 template <typename CellIteratorType, typename DataType>
-typename CellDataStorage<CellIteratorType, DataType>::key_type
-CellDataStorage<CellIteratorType, DataType>::get_key(
-  const CellIteratorType &cell)
-{
-  return {&cell->get_triangulation(), cell->id()};
-}
-
-template <typename CellIteratorType, typename DataType>
 template <typename T>
 void
 CellDataStorage<CellIteratorType, DataType>::initialize(
@@ -561,8 +571,13 @@ CellDataStorage<CellIteratorType, DataType>::initialize(
 {
   static_assert(std::is_base_of<DataType, T>::value,
                 "User's T class should be derived from user's DataType class");
+  // The first time this method is called, it has to initialize the reference
+  // to the triangulation object
+  if (!tria)
+    tria = &cell->get_triangulation();
+  Assert(&cell->get_triangulation() == tria, ExcTriangulationMismatch());
 
-  const auto key = get_key(cell);
+  const auto key = cell->id();
   if (map.find(key) == map.end())
     {
       map[key] = std::vector<std::shared_ptr<DataType>>(n_q_points);
@@ -596,10 +611,11 @@ template <typename CellIteratorType, typename DataType>
 bool
 CellDataStorage<CellIteratorType, DataType>::erase(const CellIteratorType &cell)
 {
-  const auto key = get_key(cell);
+  const auto key = cell->id();
   const auto it  = map.find(key);
   if (it == map.end())
     return false;
+  Assert(&cell->get_triangulation() == tria, ExcTriangulationMismatch());
   for (unsigned int i = 0; i < it->second.size(); i++)
     {
       Assert(
@@ -645,8 +661,9 @@ CellDataStorage<CellIteratorType, DataType>::get_data(
 {
   static_assert(std::is_base_of<DataType, T>::value,
                 "User's T class should be derived from user's DataType class");
+  Assert(&cell->get_triangulation() == tria, ExcTriangulationMismatch());
 
-  const auto it = map.find(get_key(cell));
+  const auto it = map.find(cell->id());
   Assert(it != map.end(), ExcMessage("Could not find data for the cell"));
 
   // It would be nice to have a specialized version of this function for
@@ -673,8 +690,9 @@ CellDataStorage<CellIteratorType, DataType>::get_data(
 {
   static_assert(std::is_base_of<DataType, T>::value,
                 "User's T class should be derived from user's DataType class");
+  Assert(&cell->get_triangulation() == tria, ExcTriangulationMismatch());
 
-  const auto it = map.find(get_key(cell));
+  const auto it = map.find(cell->id());
   Assert(it != map.end(), ExcMessage("Could not find QP data for the cell"));
 
   // Cast base class to the desired class. This has to be done irrespectively of
--- a/include/deal.II/base/quadrature_point_data.h
+++ b/include/deal.II/base/quadrature_point_data.h
@@ -160,9 +160,31 @@ public:
 
 private:
   /**
-   * A map to store a vector of data on a cell.
+   * Depending on @p CellIteratorType we need to define a triangulation type
    */
-  std::map<CellIteratorType, std::vector<std::shared_ptr<DataType>>> map;
+  using triangulation_type =
+    Triangulation<CellIteratorType::AccessorType::dimension,
+                  CellIteratorType::AccessorType::space_dimension>;
+
+  /**
+   * The key is used as identified for a cell data structure.
+   * We need to use CellId as part of the key because it remains unique during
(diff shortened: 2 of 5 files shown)


jppelteret:
I *think* that we used cell iterators because they are bound to a
triangulation. So if you have two triangulations and one qph (i.e.
celldatastorage) then you can only access the qph from the triangulation
from which it was initialised. Does this hold for CellId too?

Although it's not explicitly stated in the documentation, what you should
do is clear() and rebuild the qph upon refinement. I believe that it how we
intended for it to be used.

jppelteret:
But if my thinking isn't correct then I'm also happy with this change.
Thanks for the patch!

drwells:
I think that this fix is still okay, but I think I need to read the implementation a few more times to convince myself.

tjhei:
Can you please add a changelog entry with @peterrum and your name.

drwells:
@tjhei there is a changelog already?

doc/news/changes/minor/20190807RezaRastakPeterMunch
```diff
@@ -0,0 +1,4 @@
+Fixed: Bug in CellDataStorage that produced error when applying adaptive
+refinement in multi-material grids.
+<br>
+Reza Rastak, Peter Munch (2019/08/07)
```
tjhei:
```suggestion
(Reza Rastak, Peter Munch, 2019/08/07)
```

tjhei:
> @tjhei there is a changelog already?

Oops, I somehow missed that.

jppelteret:
@tjhei Can you quickly ease my mind on this?
```c++
Triangulation<2> tria_1;
Triangulation<2> tria_2;

GridGenerator::hyper_cube(tria_1);
GridGenerator::hyper_cube(tria_2);

const auto cell_id_1 = tria_1.begin_active()->cell_id();
const auto cell_id_2 = tria_2.begin_active()->cell_id();

// Is cell_id_1 == cell_id_2 ?
```

peterrum:
@jppelteret : yes it is the same

drwells:
One possible fix would be to use `std::pair<Triangulation<dim, spacedim> *, CellID>` as the key. To be clear: I am not yet convinced that this is necessary (or that it is unnecessary).

peterrum:
@drwells: I do not really think this is necessary. This case would have failed in the old implementation, since it used to compare only level and index on that level...

jppelteret:
> @jppelteret : yes it is the same

Thanks for the confirmation @peterrum.

Hmm... Then I don't think that this exact approach is necessarily the way that we want to go. I think that I'd prefer for @drwells's suggestion to be adopted. I would like any alterations made here to be as water-tight a solution as possible. Either that or the original behaviour to be documented better, and for it to be highlighted that upon mesh adaptation a call to `clear()` be made.

jppelteret:
But then again, I may misunderstand how the iterator comparison is implemented. I'm quickly digging through the source code to see how its done.

peterrum:
@jppelteret: see https://www.dealii.org/developer/doxygen/deal.II/classTriaRawIterator.html#a3b368e27fe7701f4f9d62de184cf2717

jppelteret:
@peterrum More specifically, https://github.com/dealii/dealii/blob/master/include/deal.II/grid/tria_iterator.h#L1065-L1066

jppelteret:
(The underlying triangulations are checked.)

peterrum:
@jppelteret: Oh, you are right! Sorry about the misinformation: in debug mode it is, indeed, checked. I am fine with pairs. How about you @rezarastak  and @drwells?

jppelteret:
A final confirmation of my supposition:
```c++

#include <iostream>

int
main()
{
  Triangulation<2> tria_1;
  Triangulation<2> tria_2;

  GridGenerator::hyper_cube(tria_1);
  GridGenerator::hyper_cube(tria_2);

  const auto cell_id_1 = tria_1.begin_active()->id();
  const auto cell_id_2 = tria_2.begin_active()->id();

  std::cout << "CELL_ID: " << std::boolalpha << (cell_id_1 == cell_id_2) << std::endl;
  std::cout << "ITERATOR: " << std::boolalpha << (tria_1.begin_active() == tria_2.begin_active()) << std::endl;
}
```
Debug mode output:
```
CELL_ID: true
ITERATOR:
An error occurred in line <124> of file <PATH/deal.II/dealii-dev/include/deal.II/grid/tria_accessor.templates.h> in function
    bool dealii::TriaAccessorBase<2, 2, 2>::operator==(const TriaAccessorBase<, , > &) const [structdim = 2, dim = 2, spacedim = 2]
The violated condition was:
   tria == a.tria || tria == nullptr || a.tria == nullptr
Additional information:
   Iterators can only be compared if they point to the same triangulation, or if neither of them are associated with a triangulation.
```
Release mode output:
```
CELL_ID: true
ITERATOR: false
```

jppelteret:
> Sorry about the misinformation

No worries :-) I'm feeling a bit rusty so I thought I was missing something. Thanks to you and @rezarastak for figuring out this bug and for the patch. 

A cool follow-up patch would be to implement a `has_data<T>()` function, which checks if there exists data for the current cell. This would tell you if you need to call `initialize()` on the cell or not after refinement. I guess we didn't create such a function because it was originally expected that a call to clear() would necessarily mean that you did need to `initialize()` for each cell. We could always open an issue for this and come back to it later.

rezarastak:
@jppelteret Thank you for your comments about considering multiple triangulations scenarios. Maybe we need to make sure that `CellDataStorage` corresponds to a unique `Triangulation` object (perhaps by keeping a constant `SmartPointer` to that triangulation).

 Please take a look at how `operator<()` is implemented in `TriaIterator` classes vs `CellId`. `CellId` provides a similar but richer comparison operator that allows consistency between successive refinements.

I strongly vote against needing to `clear()` between refinements because of my specific use case. I am running a nonlinear time dependent problem where initially the mesh is coarse. As the time advances, the error estimate grows and the mesh needs to be adaptively refined. During the refinement, we can't afford to lose any information by calling `clear()` because we don't want to rerun the simulation from `t = 0`. Also, we don't want to needlessly reconstruct and copy the data structures for all cells. We just need to initialize the data on a small percentage of cell that have been just created, then interpolate the values from the parent cell to the newly created child cells. Then we continue with our simulation through time.

jppelteret:
BTW. Why you might want to clear the data storage object after refinement is that you could end up with a whole lot of memory allocated to QPH data ("owned" by coarse cells) that is no longer used / accessible.

rezarastak:
Yes we can selectively `erase()` those coarse cells after we interpolate their values to the children. We don't need to call `clear()`.

jppelteret:
> Yes we can selectively erase() those coarse cells after we interpolate their values to the children. We don't need to call clear().

Fair enough :-)

> Maybe we need to make sure that CellDataStorage corresponds to a unique Triangulation object (perhaps by keeping a constant SmartPointer to that triangulation).

I'm not sure that the ownership model needs to be that strong. With the current implementation you can have one data storage object that is used to store data associated with multiple triangulations (since the cell iterators are unique for different Trias). So what needs to be clear is that for each triangulation that gives data to the `CellDataStorage` object there remains a 1-1 correspondence between each `cell` and the returned data. I believe that suggestion that @drwells gave provides this solution.

rezarastak:
It seems that there is consensus to use `std::pair<Triangulation<dim, spacedim> *, CellID>` as the  key of the map, and it's fine with me. Should I make the modification in the current PR or should I prepare another PR for that after we merge this?

jppelteret:
> Please take a look at how operator<() is implemented in TriaIterator classes vs CellId. CellId provides a similar but richer comparison operator that allows consistency between successive refinements.

Apart from this one minor issue that I've raised, you've won me over with using `CellId`. So lets try to make your implementation a hint more robust then we can move forward!

> It seems that there is consensus to use std::pair<Triangulation<dim, spacedim> *, CellID> as the key of the map, and it's fine with me. Should I make the modification in the current PR or should I prepare another PR for that after we merge this?

Please amend the current PR :-)

jppelteret:
@rezarastak @peterrum Thanks again for working on this. Great catch for a horrible deficiency / bug.

@drwells @tjhei I leave this in your safe hands -- I'm off to club duvet.

drwells:
Its late, so I may be missing something, but it looks like this PR has exposed something else that we need to check: the `Triangulation` should live longer than this object. To guarantee that the pointers remain valid we should explicitly store a `SmartPointer<Triangulation<dim,spacedim>>`, especially if we store raw pointers to cells.

Since we store cell ids, we should also check (at least in debug mode) that all provided iterators point to the same `Triangulation` that we store.

I would like to review this more carefully in the morning to make sure that both this problem and the suggested fix above are sufficient, so please don't merge until I take a closer look.

rezarastak:
@drwells Thanks for pointing this out and looking more into it tomorrow.

> To guarantee that the pointers remain valid we should explicitly store a SmartPointer<Triangulation<dim,spacedim>>

It doesn't seem to be necessary since we never dereference this pointer. We only store the address of the triangulation as an identifier. If the triangulation related to a piece of cell data goes out of scope, we might have difficulty accessing the old (most likely useless) data, but memory corruption or invalid memory read is not possible.

> Since we store cell ids, we should also check (at least in debug mode) that all provided iterators point to the same Triangulation that we store.

We might have cell_data_storage that stores data corresponding to multiple triangulation objects. We don't need to and don't want to restrict it to only one triangulation object.

peterrum:
@drwells @rezarastak: I cannot really imagine a case where anyone might want create multiple triangulation objects (except in very very special cases). The class `ContinuousQuadratureDataTransfer` cannot even handle multiple triangulations.

So if one should actually have multiple tringulation objects, then the user should create an two `CellDataStorage`, one for each triangulation, and communicate in two steps.

rezarastak:
I thought @jppelteret wants to allow multiple triangulations to store data on the same `CellDataStorage` object. That's the reason we store the pointer to the triangulation as part of the key. If there is only a unique triangulation corresponding to a `CellDataStorage` object, why duplicate the same pointer in all the keys?

jppelteret:
> The class ContinuousQuadratureDataTransfer cannot even handle multiple triangulations.

Yes, this is what just came to mind when I glossed over #8506. I'd forgotten about `ContinuousQuadratureDataTransfer`, so maybe my suggestion was in fact overly broad :-(

peterrum:
I think we should definetly check that a CellDataStorage corresponds to one Triangulation. Maybe not in the form of pairs as it is done right now, but rather a pointer to the Triangulation is set as a class variable the first time the initializer is called....

rezarastak:
After talking with @drwells we suggest the following plan:

1. Require that `CellDataStorage` relates to a single unique existing triangulation.
1. Store a `SmatrPointer` to the triangulation.
1. Add a new constructor `CellDataStorage(const triangulation_type &tria);`
1. Deprecate the old default constructor `CellDataStorage() DEAL_II_DEPRECATED;`
1. Key of the internal `std::map` becomes `CellId`
1. If `CellDataStorage` is default constructed, then the pointer to triangulation is set when `initialize()` method is called for the first time.
1. When calling `initialize()`, `get_data()`, or `erase()` we assert that the provided cell belongs to the triangulation.

Feel free to suggest modifications to this plan.

jppelteret:
Yip, sounds good to me. Thanks for giving this some more thought and for being willing to follow through with the modifications.

tjhei:
> * Add a new constructor `CellDataStorage(const triangulation_type &tria);`
> * Deprecate the old default constructor `CellDataStorage() DEAL_II_DEPRECATED;`

Would it be possible to leave the current constructor in place and lazily set the SmartPtr the first time a cell iterator is handed in? Then you don't need to set the Triangulation.

drwells:
That would also work (I assumed that we would set the pointer that way to preserve backwards compatibility anyway): so would an explicit `initialize` method.

@tjhei which initialization method do you think would be the most idiomatic in deal.II?

drwells:
whoops, wrong button!

doc/news/changes/minor/20190808RezaRastak
```diff
@@ -0,0 +1,5 @@
+Improved: CellDataStorage stores a reference to a single unique triangulation
+to ensure that cells from other triangulations cannot store any data on the
+current object.
+<br>
+(Reza Rastak, 2019/08/08)
```
drwells: This is technically an incompatibility. Could you move the changelog entry to that folder?

include/deal.II/base/quadrature_point_data.h
```diff
@@ -138,6 +142,9 @@ class CellDataStorage : public Subscriptor
    * cell-by-cell basis.
    *
    * @pre The type @p T needs to match the class provided to initialize() .
+   *
+   * @pre @p cell must be from the same triangulation that is used to
```
drwells:
```suggestion
   * @pre @p cell must be from the same Triangulation that is used to
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -153,23 +160,51 @@ class CellDataStorage : public Subscriptor
...
```
drwells:
```suggestion
   * @pre @p cell must be from the same Triangulation that is used to
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -153,23 +160,51 @@ class CellDataStorage : public Subscriptor
...
   get_data(const CellIteratorType &cell) const;
 
 private:
   /**
-   * A map to store a vector of data on a cell.
+   * Depending on @p CellIteratorType we need to define a triangulation type
+   */
+  using triangulation_type =
+    Triangulation<CellIteratorType::AccessorType::dimension,
+                  CellIteratorType::AccessorType::space_dimension>;
```
drwells: The normal way we do that is to use
```
static const unsigned int dimension = CellIteratorType::AccessorType::dimension;
```
etc. Could you do that for `dimension` and `space_dimension` and then remove this using statement? They do the same thing, but we usually save the spatial dimensions in this way.

include/deal.II/base/quadrature_point_data.h
```diff
@@ -153,23 +160,51 @@ class CellDataStorage : public Subscriptor
...
+
+  /**
+   * To ensure that all the cells in the CellDataStorage come from the same
+   * triangulation, we need to store a reference to that triangulation within
```
drwells:
```suggestion
   * Triangulation, we need to store a reference to that Triangulation within
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -91,6 +91,10 @@ class CellDataStorage : public Subscriptor
    * may reflect, for example, different constitutive models of continuum
    * mechanics in different parts of the domain.
    *
+   * @note The first time this method is called, it stores a reference to the
```
drwells:
```suggestion
   * @note The first time this method is called, it stores a SmartPointer to the
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -91,6 +91,10 @@ class CellDataStorage : public Subscriptor
...
+   * triangulation object that owns the cell. The future invocations of this
```
drwells:
```suggestion
   * Triangulation object that owns the cell. The future invocations of this
```

rezarastak:
I addressed the comments.

drwells:
The fixes address all comments made by myself, @tjhei, and @jppelteret, so lets merge.

PR #8485, opened 2019-08-07
Add a possible extension in Step-2 tutorial

User-zwj:
An extension in step-2 tutorial: Use Gnuplot to print out the mesh with locations and numbering of the support points.


bangerth:
@tjhei -- I think this is in reference to your recent issue.

examples/step-2/doc/results.dox
```diff
@@ -73,3 +73,5 @@ Terminal type set to 'x11'
 gnuplot> set style data points
 gnuplot> plot "sparsity_pattern.1"
 @endcode
+
+Another practice based on <ahref="http://www.gnuplot.info/">GNUPLOT</a> is trying to print out the mesh with locations and numberings of the support points where you might need to use GridOut and MappingQ1.
```
bangerth: Can you break this line to roughly 80 characters? You will also have to put a space between `<a` and `href`. There may be more things I can't currently see because they scroll out to the right :-)

User-zwj: Sorry about about the long line, I just divided it into several short lines, hope it will work. Also added the code in the end.

tjhei:
A few things went wrong with your pull request. I think it would be easiest if you talk to one of us about it. As I made the suggestion for the change, maybe talk to me?

User-zwj:
Replaced by #8514

PR #8465, opened 2019-08-06
Add frobenius_norm method to BlockMatrixBase

class4kayaker:
Fix to #7629

I could not find any tests in `tests/lac/` which used the `BlockMatrixBase::matrix_norm_square` function, so while I would like to add a test for this new functionality, I do not see an obvious place to insert it.

Issue #7629, opened 2019-01-23
add dealii::BlockSparseMatrix::frobenius_norm()


davydden:
subj. Should be trivial but useful.

class4kayaker:
I am working on this

class4kayaker:
The relevant PR has been merged.

PR #8487, opened 2019-08-07
Add L1 and LInfty norm methods to BlockMatrixBase

class4kayaker:
Logical extension of issue #7629, and work in #8465.

As suggested, both the l1_norm and linfty_norm method tests are located in separate test files.


drwells:
Looks good. We need to change one minor thing before proceeding: we need to define `BlockMatrixBase::real_type` (see `vector.h`) so that these norms will return a real type when we use complex-valued matrices.

class4kayaker:
Altered to include a real_type in BlockMatrixBase, and use it as the return type.

drwells:
We can now address the merge conflict here.

class4kayaker:
Rebased onto master with merge conflict fixed.

drwells:
I thought about this some more (I should have done this before turning on the CI): this implementation is different from the one used in the Frobenius norm PR in that it no longer dispatches calculation of the norms to the blocks themselves. That approach is fine. This approach, where we implement the 1 and infinity norms ourselves, isn't going to work with the parallel sparse matrix classes.

I recommend that we add some tests for either `PETScWrappers::MPI::BlockSparseMatrix` or `TrilinosWrappers::BlockSparseMatrix` (which both use `BlockMatrixBase`) that run with multiple MPI ranks to verify that this will work correctly when not all rows or columns are locally available.

class4kayaker:
You are correct, and I do not see a way to correct this given the differences in implementation between PETSc and Trilinos which both dispatch to internal functions for the `SparseMatrix` templates. I believe that closing this PR may be best.

drwells:
It might be possible to fix this. Lets mark it as WIP for now.

drwells:
Unfortunately, I don't think that it is possible to implement this in a way that will work with all derived classes. Consider the following: suppose we have a 3x3 block matrix of PETSc matrices. There is no way to compute the 1 norm of this block matrix since we can only compute the max row sums of individual blocks: (and the max of the sums is not the sum of the maxes). The same problem holds for the infinity norm.

class4kayaker:
Yeah, that was what I concluded when you noted that there were MPI cases that I overlooked, once I had checked and did not see an obvious way to do the row/column sums in a general manner.

PR #8497, opened 2019-08-07
Update step 33 (fix UpdateFlags)

mzaguilar8:
Chane update_q_points to update_quadrature_points.

### DIFF
--- a/doc/news/changes/minor/20190806KiranaBergstrom
+++ b/doc/news/changes/minor/20190806KiranaBergstrom
@@ -1,0 +1,4 @@
+Fixed: step-20 tutorial permeability computation now uses Tensor::norm_square instead
+of Point::square on a difference of points.
+<br>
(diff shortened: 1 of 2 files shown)


bangerth:
Ah, you triggered the line length check :-( Can you talk to one of us about running the indentation script?

drwells:
Thanks a lot for the fix!

Could you add a changelog entry describing the fix so that we can add you to the list of deal.II authors? You should add a file in `doc/new/changes/minor/` describing this change: the easiest way to do this is to copy and paste one of the files in this directory, rename it with your name and date, and then change its text to describe this fix.

drwells:
@mzaguilar8 Let us know if you need a hand in wrapping this up.

drwells:
Since all we needed to do was run `make indent` and fix authorship information I went ahead and did it. Closed in favor of #8642.

PR #8500, opened 2019-08-07
Fix typos

adriana6:
Fixed typos in two documents.

### DIFF
--- a/examples/step-11/step-11.cc
+++ b/examples/step-11/step-11.cc
@@ -188,7 +188,7 @@ namespace Step11
     // beforehand the maximal number of entries per row, either for all rows
     // or for each row separately. There are functions in the library which
     // can tell you this number in case you just have hanging node constraints
-    // (namely <code>DoFHandler::max_coupling_between_dofs</code>), but how is
+    // (namely DoFHandler::max_couplings_between_dofs), but how is
     // this for the present case? The difficulty arises because the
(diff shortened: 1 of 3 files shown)


examples/step-11/step-11.cc
```diff
@@ -188,7 +188,7 @@ namespace Step11
     // beforehand the maximal number of entries per row, either for all rows
     // or for each row separately. There are functions in the library which
     // can tell you this number in case you just have hanging node constraints
-    // (namely <code>DoFHandler::max_coupling_between_dofs</code>), but how is
+    // (namely <code>DoFHandler::max_couplings_between_dofs</code>), but how is
```
drwells:
```suggestion
    // (namely DoFHandler::max_couplings_between_dofs), but how is
```
This problem was already here, but since its going to be a hyperlink we don't need the HTML `<code>` and `</code>` delimiters.

drwells:
Could you add a changelog entry describing the fix so that we can add  you to the list of deal.II authors? You should add a file in `doc/new/changes/minor/` describing this change: the easiest way to do this is to copy and paste one of the files in this directory, rename it with your name and date,  and then change its text to describe this fix.

drwells:
Looks good :+1:

Since this is a small change, would you mind also rebasing all of your commits together? Feel free to ask if you are not sure how to do this.

PR #8502, opened 2019-08-07
Clarified instructions for install directory

genevaporter:
Clarify some options for installation of deal.II.

(Edited by @drwells)

### DIFF
--- a/doc/readme.html
+++ b/doc/readme.html
@@ -161,7 +161,7 @@
 
     <p>
         <acronym>deal.II</acronym> uses the
-        <a href="http://www.cmake.org/" target="_top">CMake</a> integrated configuration and build system. Unpacking will create a subdirectory <tt>deal.II/</tt> in the current directory. Then do the following steps to configure the build process:</p>
+        <a href="http://www.cmake.org/" target="_top">CMake</a> integrated configuration and build system. Unpacking will create a subdirectory <tt>deal.II/</tt> in the current directory. Then do the following steps to configure the build process (make sure the installation directory is not the same as the location where you unpacked <tt>deal.II/</tt>):</p>
 
     <pre>
   mkdir build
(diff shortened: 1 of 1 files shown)


doc/readme.html
```diff
@@ -183,7 +183,7 @@ <h3>Configuring and building the library</h3>
         <b>Note:</b></p>
     <ul>
         <li> <code>/path/to/install/dir</code> is the directory which deal.II should be installed into. This can be a directory in your home directory (e.g., <code>~/bin/deal.II</code>) or a directory such as <code>/usr/local</code> if you have root privileges.
-            Another option is to use something like <code>`pwd`/../installed/</code> (note the backticks). Make sure the installation directory is not the same as the location where you unpacked <tt>deal.II/</tt>.
+            Another option is to use something like <code>`pwd`/../installed/</code> (note the backticks).
```
drwells: Backticks are popular but hard to read and understand. I think it would be better to use the environment variable, i.e, `something like <code>$PWD/../installed/</code>`

See
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08

drwells:
Furthermore, would you please change the commit log entry?

doc/readme.html
tjhei: This is unrelated to this pull request, but I agree that this would be a useful change.

tjhei:
Your commit message contains some random text. That is why this PR looks really weird if you look at this page. Can you please find one of us to rewrite your commit to have a good commit message?

doc/readme.html
drwells: Good point: I should have read the full line change. Lets skip this.

genevaporter:
Yes, will find one of you today!

masterleinad:
@drwells We fixed the changelog entry.

drwells:
@masterleinad @genevaporter Are you sure? I don't see a changelog entry :)

masterleinad:
Oh, I meant the commit message. Do you want to have a changelog entry as well?

drwells:
Oh, sorry, I should have been more explicit. The commit log entry is fine. I meant to ask (and did not previously) for y'all to also add a changelog entry (like the other PRs from the workshop).

drwells:
Based on the discussion the developers had elsewhere, lets not worry about adding the changelog here: we can deal with that later on our own.

Thanks a lot for the fix @genevaporter!

PR #8519, opened 2019-08-08
Grid tools debug

### LINKED ISSUE #8517: GridTools::Cache does not in fact cache things sometimes :-)
Looking at #8516, here is some code in `GridTools::Cache` (in `source/grid/grid_tools_cache.cc`):
```
  const RTree<
    std::pair<BoundingBox<spacedim>,
  Cache<dim, spacedim>::get_cell_bounding_boxes_rtree() const
  {
    if (update_flags & update_cell_bounding_boxes_rtree)
      {
        std::vector<std::pair<
          BoundingBox<spacedim>,
          typename Triangulation<dim, spacedim>::active_cell_iterator>>
                     boxes(tria->n_active_cells());
        unsigned int i = 0;
        for (const auto &cell : tria->active_cell_iterators())
          boxes[i++] = std::make_pair(mapping->get_bounding_box(cell), cell);

        cell_bounding_boxes_rtree = pack_rtree(boxes);
      }
    return cell_bounding_boxes_rtree;
  }
```
If the flag is set, it recomputes the bounding box and then returns it. The *idea* of course is that once computed, the cache will know about it the next time the function is called -- but it forgets to clear the flag in the update flags. Here is a function that does it correctly:
```
  const KDTree<spacedim> &
  Cache<dim, spacedim>::get_vertex_kdtree() const
  {
    if (update_flags & update_vertex_kdtree)
      {
        vertex_kdtree.set_points(tria->get_vertices());
        update_flags = update_flags & ~update_vertex_kdtree;     // *** clearing the flag ***
      }
    return vertex_kdtree;
  }
```

We should go through all of the `get_()` functions of this class to see whether they really do clear the respective flag in the `if()` block.

@mjayayadharan -- want to give this a try?

mjayadharan:
Fixes #8517.

### DIFF
--- a/doc/news/changes/minor/20190808MJayadharan
+++ b/doc/news/changes/minor/20190808MJayadharan
@@ -1,0 +1,3 @@
+Debug: Fixed two functions in GridTools::Cache class (in source/grid/grid_tools_cache.cc), namely get_used_vertices_rtree() and get_cell_bounding_boxes_rtree where the update flag were not cleared. Fixed that by adding a line to both functions at the end of if loop checking the flag.
+<br>
+(Manu Jayadharan, 2019/08/08)
--- a/source/grid/grid_tools_cache.cc
+++ b/source/grid/grid_tools_cache.cc
@@ -114,7 +114,7 @@ namespace GridTools
         for (const auto &it : used_vertices)
           vertices[i++] = std::make_pair(it.second, it.first);
         used_vertices_rtree = pack_rtree(vertices);
-        update_flags = update_flags & ~update_used_vertices_rtree;
+        update_flags        = update_flags & ~update_used_vertices_rtree;
       }
     return used_vertices_rtree;
   }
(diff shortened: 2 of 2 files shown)


masterleinad:
Please create two separate pull requests if you are fixing unrelated issues.

Could you also add a test case that would fail prior to this pull request?

tjhei:
And if you need help with anything @masterleinad said, talk to one of us!

bangerth:
@mjayadharan -- come see me tomorrow and we'll fix the git issues.

@dealii/dealii : There's no test that can check this. What was happening is that we wanted to cache things, but forgot to set the flag that we had already computed something. The result wasn't wrong -- it was just inefficient: *Every* time we were looking up the information we re-computed it.

mjayadharan:
I have separated two issues with this pull request. Now please take at look.

bangerth:
/rebuild

OK to merge if it passes the tester (which I strongly suspect it will). Thanks for the patch, @mjayadharan !

bangerth:
Fixes #8517.

mjayadharan:
Glad that it worked out.

luca-heltai:
Thanks for fixing my mistake!

:)

Issue #1973, opened 2015-12-10
FEValues cleanup: Remove polynomial space template argument from FE_Poly and FE_PolyTensor


bangerth:
FE_Poly and FE_PolyTensor get a classname for the polynomial space as a template argument. This leads to some awkward duplication of code, see for example the explicit specializations in fe_poly.cc. I think that in practice, this could be avoided by not passing the name of the polynomial space class to `FE_Poly` and `FE_PolyTensor` but instead passing an object and then storing a pointer to it in the class. This should also significantly reduce compile time because the same class doesn't have to be instantiated that many times (and we could move the contents of the `.templates.h` file into the `.cc` file because we then know what template arguments to instantiate).

Related to #1198.

edit: Add the tasks
1. [x] Implement the TensorPolynomialsBase class
2. [x] Change tensor polynomial classes to derive from it
3. [x] Remove the class template argument from FE_PolyTensor
4. [x] Implement the ScalarPolynomialsBase class
5. [ ] Change scalar polynomial classes to derive from it
6. [ ] Remove the class template argument from FE_Poly

kronbichler:
We won't be able to do this now on a short notice.

masterleinad:
Depends on #8711.

PR #8528, opened 2019-08-09
Implement the PolynomialsBase classes

GrahamBenHarper:
The goal of this is the first step of many required to address #1973.

This class will use clone functions in a similar fashion to the `FiniteElement` classes, and then use that to pass them to `FE_PolyTensor`, which will have a member `PolynomialsBase poly_space` which is created when the constructor for the derived classes, such as `FE_RaviartThomas` are called.

I'm looking for all of your feedback since this will be a multi-step process.

### DIFF
--- a/include/deal.II/base/tensor_polynomials_base.h
+++ b/include/deal.II/base/tensor_polynomials_base.h
@@ -1,0 +1,174 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2005 - 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_tensor_polynomials_base_h
+#define dealii_tensor_polynomials_base_h
+
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/base/exceptions.h>
+#include <deal.II/base/point.h>
+#include <deal.II/base/tensor.h>
+
+#include <vector>
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * This class provides a framework for the finite element polynomial
+ * classes for use with finite element classes that are derived from
+ * @p FE_Poly and @p FE_PolyTensor. An object of this type is stored
+ * as a member variable in the finite element class.
+ *
+ * <h3>Deriving classes</h3>
+ *
+ * Any derived class must provide the most basic properties for shape
+ * functions evaluated on the reference cell. This includes, but is not
+ * limited to, implementing the <tt>compute</tt>, <tt>name</tt>, and
+ * <tt>clone</tt> member functions. These functions are necessary to store the
+ * most basic information of how the polynomials in the derived class evaluate
+ * at a given point on the reference cell. More information on each function can
+ * be found in the corresponding function's documentation.
+ *
+ * Some classes that derive from this class include
+ * <ul>
+ *   <li> <tt>PolynomialsABF</tt>
+ *   <li> <tt>PolynomialsBDM</tt>
+ *   <li> <tt>PolynomialsBernardiRaugel</tt>
+ *   <li> <tt>PolynomialsNedelec</tt>
+ *   <li> <tt>PolynomialsRaviartThomas</tt>
+ *   <li> <tt>PolynomialsRT_Bubbles</tt>
+ * </ul>
+ *
+ * @ingroup Polynomials
+ * @author Graham Harper
+ * @date 2019
+ */
+template <int dim>
+class TensorPolynomialsBase
+{
+public:
+  /**
+   * Constructor. This takes the degree of the space, @p deg from the finite element
+   * class, and @p n, the number of polynomials for the space.
+   */
+  TensorPolynomialsBase(const unsigned int deg,
+                        const unsigned int n_polynomials);
+
+  /**
+   * Move constructor.
+   */
+  TensorPolynomialsBase(TensorPolynomialsBase<dim> &&) = default; // NOLINT
+
+  /**
+   * Copy constructor.
+   */
+  TensorPolynomialsBase(const TensorPolynomialsBase<dim> &) = default;
+
+  /**
+   * Virtual destructor. Makes sure that pointers to this class are deleted
+   * properly.
+   */
+  virtual ~TensorPolynomialsBase() = default;
+
+  /**
+   * Compute the value and the derivatives of the polynomials at
+   * @p unit_point.
+   *
+   * The size of the vectors must either be zero or equal <tt>n()</tt>.  In
+   * the first case, the function will not compute these values.
+   *
+   * If you need values or derivatives of all polynomials then use this
+   * function, rather than using any of the <tt>compute_value</tt>,
+   * <tt>compute_grad</tt> or <tt>compute_grad_grad</tt> functions, see below,
+   * in a loop over all tensor product polynomials.
+   */
+  virtual void
+  compute(const Point<dim> &           unit_point,
+          std::vector<Tensor<1, dim>> &values,
+          std::vector<Tensor<2, dim>> &grads,
+          std::vector<Tensor<3, dim>> &grad_grads,
+          std::vector<Tensor<4, dim>> &third_derivatives,
+          std::vector<Tensor<5, dim>> &fourth_derivatives) const = 0;
+
+  /**
+   * Return the number of polynomials.
+   */
+  unsigned int
+  n() const;
+
+  /**
+   * Return the highest polynomial degree of polynomials represented by this
+   * class. A derived class may override this if its value is different from
+   * @p my_degree.
+   */
+  unsigned int
+  degree() const;
+
+  /**
+   * A sort of virtual copy constructor, this function returns a copy of
+   * the polynomial space object. Derived classes need to override the function
+   * here in this base class and return an object of the same type as the
+   * derived class.
+   *
+   * Some places in the library, for example the constructors of FE_PolyTensor,
+   * need to make copies of polynomial spaces without knowing their exact type.
+   * They do so through this function.
+   */
+  virtual std::unique_ptr<TensorPolynomialsBase<dim>>
+  clone() const = 0;
+
+  /**
+   * Return the name of the space.
+   */
+  virtual std::string
+  name() const = 0;
+
+private:
+  /**
+   * The highest polynomial degree of this functions represented by this object.
+   */
+  const unsigned int polynomial_degree;
+
+  /**
+   * The number of polynomials represented by this object.
+   */
+  const unsigned int n_pols;
+};
+
+
+
+template <int dim>
+inline unsigned int
+TensorPolynomialsBase<dim>::n() const
+{
+  return n_pols;
+}
+
+
+
+template <int dim>
+inline unsigned int
+TensorPolynomialsBase<dim>::degree() const
+{
+  return polynomial_degree;
+}
+
+
+
+DEAL_II_NAMESPACE_CLOSE
+
+#endif
--- a/source/base/CMakeLists.txt
+++ b/source/base/CMakeLists.txt
@@ -59,6 +59,7 @@ SET(_unity_include_src
   polynomial.cc
   polynomials_abf.cc
   polynomials_adini.cc
+  tensor_polynomials_base.cc
   polynomials_bernstein.cc
   polynomials_bdm.cc
   polynomials_bernardi_raugel.cc
--- a/source/base/tensor_polynomials_base.cc
+++ b/source/base/tensor_polynomials_base.cc
@@ -1,0 +1,42 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2005 - 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+#include <deal.II/base/quadrature_lib.h>
+#include <deal.II/base/tensor_polynomials_base.h>
+#include <deal.II/base/thread_management.h>
+
+#include <iomanip>
+#include <iostream>
+
+DEAL_II_NAMESPACE_OPEN
+
+template <int dim>
+TensorPolynomialsBase<dim>::TensorPolynomialsBase(
+  const unsigned int deg,
+  const unsigned int n_polynomials)
+  : polynomial_degree(deg)
+  , n_pols(n_polynomials)
+{
+  // nothing to do here for now
+}
+
+
+
+template class TensorPolynomialsBase<1>;
+template class TensorPolynomialsBase<2>;
+template class TensorPolynomialsBase<3>;
+
+DEAL_II_NAMESPACE_CLOSE


masterleinad:
It's really difficult to say anything about this class if it's not used anywhere and doesn't do anything.

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+          std::vector<Tensor<2, dim>> &grads,
+          std::vector<Tensor<3, dim>> &grad_grads,
+          std::vector<Tensor<4, dim>> &third_derivatives,
+          std::vector<Tensor<5, dim>> &fourth_derivatives) const = 0;
+
+  /**
+   * Return the number of polynomials.
+   */
+  virtual unsigned int
+  n() const;
```
bangerth: How would the base class implement this and the next function? In classes such as this, it's rare that functions should be both `virtual` and not abstract at the same time, because either the case class can do the right thing (no need to make it `virtual`) or the base class doesn't know anything (`virtual` but also abstract).

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * Return the degree of the space.
```
bangerth:
```suggestion
   * Return the polynomial degree of the space.
```

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+   */
+  virtual std::string
+  get_name() const = 0;
+
+  /**
+   * Return the number of polynomials in the space. This must be overridden by
+   * the derived class.
+   */
+  virtual unsigned int
+  compute_n_pols(const unsigned int k) const;
```
bangerth: If it must be overridden, then it should be abstract.

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+
+private:
+  /**
+   * The degree of this object as given to the constructor.
+   */
+  const unsigned int my_degree;
```
bangerth:
```suggestion
  const unsigned int polynomial_degree;
```

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Return the number of polynomials in the space. This must be overridden by
+   * the derived class.
+   */
+  virtual unsigned int
+  compute_n_pols(const unsigned int k) const;
+
+private:
+  /**
+   * The degree of this object as given to the constructor.
```
bangerth:
```suggestion
   * The polynomial degree of this functions represented by this object,
   * as given to the constructor.
```

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// NOTE: The required calls for FE_PolyTensor seem to be
+// poly_space.compute(), poly_space.n(), and poly_space.degree()
+// poly_space.n() will be filled by the function compute_n_pols(k).
```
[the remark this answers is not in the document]
GrahamBenHarper: Oh yes! Those were notes to myself to make sure I implement those functions in the base class. I'll remove those.

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+
+#ifndef dealii_polynomials_base_h
+#define dealii_polynomials_base_h
```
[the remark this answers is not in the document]
GrahamBenHarper: I was following the example the other polynomials files, which use
`#define dealii_polynomials_raviart_thomas_h`
or
`#define dealii_polynomials_bernardi_raugel_h`
Should I change this?

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * This class provides a framework for the finite element polynomial
+ * classes for use with finite element classes that are derived from
+ * @p FE_Poly and @p FE_PolyTensor. An object of this type is stored
+ * as a member variable in the finite element class.
+ *
+ * <h3>Deriving classes</h3>
+ *
+ * Any derived class must provide a the most basic properties for shape
```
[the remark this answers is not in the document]
GrahamBenHarper: It might be useful to have a section about writing derived classes when others try to implement their own finite element classes. Is there a way I can make this section more meaningful?

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+ * @author Graham Harper
+ * @date 2019
+ */
+template <int dim>
+class PolynomialsBase
+{
+  /**
+   * Constructor.
+   */
+  PolynomialsBase(const unsigned int k);
```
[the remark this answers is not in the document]
GrahamBenHarper: Got it!

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+   * the first case, the function will not compute these values.
+   *
+   * If you need values or derivatives of all polynomials then use this
+   * function, rather than using any of the <tt>compute_value</tt>,
+   * <tt>compute_grad</tt> or <tt>compute_grad_grad</tt> functions, see below,
+   * in a loop over all tensor product polynomials.
+   */
+  virtual void
+  compute(const Point<dim> &           unit_point,
+          std::vector<Tensor<1, dim>> &values,
```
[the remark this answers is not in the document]
GrahamBenHarper: That's a good point. I didn't think about that. I think I will change the name to `TensorPolynomialsBase` and also write a `ScalarPolynomialsBase` if that sounds reasonable. I think there wouldn't be a need to abstract those any further since any polynomial space must be derived from one of those two directly.

include/deal.II/base/polynomials_base.h
GrahamBenHarper: Got it!

include/deal.II/base/polynomials_base.h
GrahamBenHarper: @bangerth I originally had it the way you suggested, but I was running into trouble with a warning. This function is called when the class is constructed
``
template <int dim>
PolynomialsBase<dim>::PolynomialsBase(const unsigned int k)
  : my_degree(k)
  , n_pols(compute_n_pols(k))
{
  // nothing to do here for now
}
``
If I make it abstract, then I get a warning that an abstract function is called by the constructor, so I chose instead to implement `compute_n_pols` with
``
template <int dim>
unsigned int
PolynomialsBase<dim>::compute_n_pols(unsigned int /*k*/) const
{
  return 0;
}
``
but use an ExcNotImplemented() that it must be overridden. What do you think?

include/deal.II/base/polynomials_base.h
GrahamBenHarper: Sometimes these spaces hold a degree that is not necessarily equal to the polynomial degree. For example, RT_0 is a subspace of Q_1. Is there a way to avoid confusion between those two quantities?

include/deal.II/base/polynomials_base.h
GrahamBenHarper: Good point. These are just getters, so they shouldn't be virtual.

include/deal.II/base/polynomials_base.h
```diff
@@ -0,0 +1,169 @@
...
+ * Any derived class must provide a the most basic properties for shape
+ * functions evaluated on the reference cell.
+ *
+ * @ingroup Polynomials
+ * @author Graham Harper
+ * @date 2019
+ */
+template <int dim>
+class PolynomialsBase
+{
```
masterleinad:
```suggestion
{
public:
```

include/deal.II/base/polynomials_base.h
masterleinad: A constructor call never resolves virtual function calls, see https://godbolt.org/z/YKO2JP.
In other words, you have to initialize `n_pols` in the derived class constructor.

GrahamBenHarper: Oh! So it would call the unimplemented function anyway. I think the two arguments route is probably the best way to go at the moment.

include/deal.II/base/polynomials_base.h
GrahamBenHarper: Whoops! Thanks!

GrahamBenHarper:
I have a follow-up question:
In all instances of polynomial files, the name getter function is implemented as `::name()`, but in all instances of finite element files, the name getter function is implementaed as `::get_name()`.

Should I call this continue this convention or should I change the polynomial files to `::get_name()`?

include/deal.II/base/polynomials_base.h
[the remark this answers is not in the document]
bangerth: No, no worries. I had hoped that we put the directory name in there as well, such as `dealii_base_polynomial_raviart_thomas_h`. But we don't seem to do that, so don't worry.

include/deal.II/base/polynomials_base.h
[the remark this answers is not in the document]
bangerth: It's useful in general, but just saying that derived classes have to implement abstract virtual functions is pretty obvious :-) If you want to, provide an overview of what kinds of operations (and in which functions) derived classes have to implement.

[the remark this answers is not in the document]
bangerth: You might also point out which classes use this class, and what functions they use.

include/deal.II/base/polynomials_base.h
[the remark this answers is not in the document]
bangerth: Correct. But one at a time -- create this one class and convert `FE_PolyTensor`. The other class can come as a second step.

include/deal.II/base/polynomials_base.h
bangerth: Then document it as the *highest* degree of the polynomials represented by this class.

bangerth:
It's an incompatible change to change the name. Leave it as is for now, but if you want to, open a github issue about it that can be changed in a separate pull request later on. It's useful to only do one thing at a time.

include/deal.II/base/polynomials_base.h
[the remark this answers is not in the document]
GrahamBenHarper: Okay. Let me know how this update looks. I tried to improve some of the wording too.

include/deal.II/base/polynomials_base.h
GrahamBenHarper: Got it!

include/deal.II/base/polynomials_base.h
GrahamBenHarper: I'm thinking that `my_degree` should be the value passed during construction. Going back to the RT_0 example, `my_degree` would be assigned to 0 on construction. However, I will make `degree()` virtual and overriding it occurs when `my_degree` and the highest polynomial degree of the space disagree.

include/deal.II/base/tensor_polynomials_base.h
```diff
@@ -0,0 +1,181 @@
...
+#include <deal.II/base/tensor.h>
+
+#include <vector>
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * This class provides a framework for the finite element polynomial
+ * classes for use with finite element classes that are derived from
+ * @p FE_Poly and @p FE_PolyTensor. An object of this type is stored
```
bangerth: Remove the reference to `FE_Poly` here, since you're not supporting it with this class.

include/deal.II/base/tensor_polynomials_base.h
```diff
@@ -0,0 +1,181 @@
...
+ * @date 2019
+ */
+template <int dim>
+class TensorPolynomialsBase
+{
+public:
+  /**
+   * Constructor. This takes a integer @p k from the finite element class,
+   * which assigned to @p my_degree on construction. @p n is assigned to @p n_pols
+   * on construction.
```
bangerth: Drop the references to the names of member variables -- these are internal to the class, and not of interest to the reader.

bangerth: Also, can you use better names for `n` and `k`? You might want to say what they *represent*, rather than what variables they are assigned to.

include/deal.II/base/tensor_polynomials_base.h
```diff
@@ -0,0 +1,181 @@
...
+  unsigned int
+  n() const;
+
+  /**
+   * Return the highest polynomial degree of polynomials represented by this
+   * class. A derived class may override this if its value is different from
+   * @p my_degree.
+   */
+  virtual unsigned int
+  degree() const;
```
bangerth: I think you should either:
* Pass the degree to the constrtuctor and return it here via a non-`virtual` function
* Make the function `=0` and let derived classes implement it.

Having both is not useful: The value passed to the constructor does, after all, come from the derived class. The derived class *knows* what the proper value is, so there is no reason for the constructor of the derived class to pass something down to the constructor of the base class, but then later return something *different* in its own implementation of the virtual function.

My preference would be the first option above.

include/deal.II/base/tensor_polynomials_base.h
GrahamBenHarper: Got it!

include/deal.II/base/tensor_polynomials_base.h
GrahamBenHarper: How about `deg` and `n`? I know `n` isn't super helpful, but I'm unsure what else would be a fitting name.

include/deal.II/base/tensor_polynomials_base.h
GrahamBenHarper: My worry is based on how different the implementations of the `degree()` function is for each polynomial space. For example, looking at BDM, BernardiRaugel, and RaviartThomas shows each is implemented totally differently.
```
PolynomialsBDM<dim>::degree() const
{
  return polynomial_space.degree() - 1;
}
```

```
PolynomialsBernardiRaugel<dim>::degree() const
{
  return 2;
}
```

```
PolynomialsRaviartThomas<dim>::degree() const
{
  return my_degree;
}
```
My belief is they should be virtual so that the base class knows they are there, but not abstract since the base implementation shouldn't always be overridden.

Or, should the class contain two member variables? One for the degree of the space (the 0 in RT_0) and one for the polynomial degree of the space (1 for RT_0), and then `degree()` always returns the polynomial degree? I apologize if I'm making this more complicated than necessary, but I just want to make sure it's compatible with all the classes that will derive from it.

include/deal.II/base/tensor_polynomials_base.h
bangerth: We generally spell things out. How about `max_polynomial_degree` and `n_polynomials`?

include/deal.II/base/tensor_polynomials_base.h
bangerth: These are all values known at the time the constructor is called in the derived class. These derived class constructors may as well pass this value down to the constructor of the base class:
```
  PolynomialsBDM::PolynomialsBDG (const uint degree) :
    TensorPolynomials (...something that can be computed from degree..., ...)
  {}
```

bangerth: And even simpler in the other two cases!

include/deal.II/base/tensor_polynomials_base.h
GrahamBenHarper: `n_polynomials` sounds good to me!

include/deal.II/base/tensor_polynomials_base.h
GrahamBenHarper: I see. That works! So this would change the implementations but keep the base class simple.

GrahamBenHarper:
I'd like to take a minute to ask a forward-thinking question before finishing these changes. Further down the road, there are changes to `FE_PolyTensor` that remove the `<class PolynomialType>` template argument. I have much of this done already on a separate branch, but i'm still bug-chasing some things.

Because of this template argument deletion, most derived classes of `FE_PolyTensor` will need to take a polynomials class as an argument. For example, the constructor of `FE_RaviartThomas` would look something like
```
FE_RaviartThomas<dim>::FE_RaviartThomas(const unsigned int deg)
  : FE_PolyTensor<dim>(
      PolynomialsRaviartThomas<dim>(deg),
      deg,
      FiniteElementData<dim>(get_dpo_vector(deg),
                             dim,
                             deg + 1,
                             FiniteElementData<dim>::Hdiv),
      std::vector<bool>(PolynomialsRaviartThomas<dim>::compute_n_pols(deg),
                        true),
      std::vector<ComponentMask>(PolynomialsRaviartThomas<dim>::compute_n_pols(
                                   deg),
                                 std::vector<bool>(dim, true)))
```
where the `PolynomialsRaviartThomas<dim>(deg)` argument uses `clone()` to initialize a const pointer object in `FE_PolyTensor`.

Most FE constructors depend on this static `compute_n_pols` from the polynomials class to initialize the size of certain variables in the FE class. Currently this PR has `compute_n_pols` as an abstract function in `TensorPolynomialsBase`.

How would I reconcile these two approaches down the road?

include/deal.II/base/tensor_polynomials_base.h
```diff
@@ -0,0 +1,180 @@
...
+   */
+  virtual std::string
+  name() const = 0;
+
+  /**
+   * Return the number of polynomials in the space based on the polynomial space
+   * degree @p deg. This must be overridden by the derived class.
+   */
+  virtual unsigned int
+  compute_n_pols(const unsigned int deg) const = 0;
```
bangerth: Can't you get rid of this function now? Wouldn't derived classes simply return the same thing that the `n()` function above already returns?

bangerth:
As for your broader question: Because you know the actual data type (here, `PolynomialsRaviartThomas<dim>`), the call to `PolynomialsRaviartThomas<dim>::compute_n_pols(deg)` does not in fact call a virtual function but a concrete implementation of `compute_n_pols()`, assuming the derived class in fact implements it.

But as mentioned elsewhere in the patch, there is really no need to actually make `compute_n_pols()` available in the base class. I think it's in fact superfluous (duplicating what `n()` does, if I see this right) and also not necessary as an interface in the base class -- it could just be a function available in the derived class because `FE_PolyTensor` does not need to call it. (If the user of the pointers-to-base-class doesn't call a function, then that function doesn't need to be virtual in the base class.)

include/deal.II/base/tensor_polynomials_base.h
GrahamBenHarper: Yes I can remove that. Is there any way that we can require derived classes to implement a `static unsigned int compute_n_pols(const unsigned int deg)`?

GrahamBenHarper:
@bangerth I see. Since we have the two argument constructor for `TensorPolynomialsBase` a derived class would look something like
```
PolynomialsRaviartThomas<dim>::PolynomialsRaviartThomas(const unsigned int deg)
  : TensorPolynomialsBase<dim>(deg, compute_n_pols(deg))
  , ...
```
From there on out, there's no need to call `compute_n_pols` with an initialized object since both the derived class and the base class will have access to `n()`. I'm following now!

GrahamBenHarper:
If everything looks good, I'll squash the commits tomorrow!

bangerth:
You'll also have to run the indentation script once again: https://travis-ci.org/dealii/dealii/builds/571723510?utm_source=github_status&utm_medium=notification

include/deal.II/base/tensor_polynomials_base.h
masterleinad: No, the only way for doing that are pure `virtual` functions, but `virtual` functions can't be `static`. You have to be explicit about the class anyway if you want to use this function for initializing a `FiniteElement`.

GrahamBenHarper:
Okay everything should be good to go now!

bangerth:
Great. I'll merge once the last tester is through.

I guess the next step would be to derive the existing polynomials classes from this base class and mark their respective functions as `virtual`. Want to submit a patch for that next?

PR #8530, opened 2019-08-09
Mutable bind implementation.

luca-heltai:
First step towards `ParsedGridGenerator`


include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,298 @@
...
+   * @authors Luca Heltai, Matthias Maier, 2019.
+   */
+  template <typename ReturnType, class... FunctionArgs>
+  struct MutableBind
+  {
+    /**
+     * A typedef to the stored std::tuple type.
+     */
+    using TupleType = std::tuple<typename std::remove_cv<
+      typename std::remove_reference<FunctionArgs>::type>::type...>;
```
jppelteret: Sorry for the potentially stupid question: Does the `std::remove_reference` mean that objects are always transferred as copies into `arguments` of the bound function? If so, then I believe that this not work for objects with no copy constructor (e.g. `Triangulation`)? Maybe this should be documented.

Otherwise... Cool class :-)

luca-heltai: Yes, that's precisely what it means. I'll add a note in the documentation. :)

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+
+    /**
+     * An std::function that stores the the original function.
```
bangerth:
```suggestion
     * An std::function that stores the original function.
```

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+   * bound_cube.parse_parameters("0: 1");
+   * bound_cube(); // calls GridGenerator::hyper_cube(tria, 0, 1);
+   * @endcode
+   *
+   * @authors Luca Heltai, Matthias Maier, 2019.
+   */
+  template <typename ReturnType, class... FunctionArgs>
+  struct MutableBind
+  {
+    /**
```
bangerth: Is there a particular reason why this is a `struct` instead of a `class`, and why all members are `public`?

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+
+    /**
+     * Currently stored arguments. These are forwarded to the function object
+     * above, when calling operator()().
+     */
+    TupleType arguments;
+
+    /**
+     * Call the original function, passing as arguments the elements of the
+     * stored TupleType.
```
bangerth:
```suggestion
     * tuple of bound arguments.
```

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+     */
+    ReturnType
+    operator()() const;
+
+    /**
+     * Set the arguments to use in @p function, for next time
+     * operator()() is called.
+     */
+    void
+    set_arguments(const TupleType &value);
```
bangerth: You could probably optimize a few things if you also added the following function:
```
    void
    set_arguments(TupleType &&value) { arguments = std::move(value); }
```

bangerth: Also, I would call the function argument `arguments` to make their intent clear.

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+
+    /**
+     * Parse the arguments to use in @p function from a string, for next time
+     * operator()() is called.
+     */
+    void
+    parse_arguments(const std::string &                           value_string,
+                    const std::unique_ptr<Patterns::PatternBase> &pattern =
+                      Patterns::Tools::Convert<TupleType>::to_pattern());
```
bangerth: Ah, that's neat!

Can you document the purpose of the two arguments?

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+
+    /**
+     * Set the arguments to use in @p function, for next time
+     * operator()() is called.
+     */
+    void
+    set_arguments(FunctionArgs... args);
```
bangerth: move this function next to the existing `set_arguments` function, i.e., above `parse_arguments`. You can probably optimize the same way by offering a variant that takes rvalue references.

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+   * auto bound = mutable_bind(my_function);
+   *
+   * bound.set_parameters(2, 3.0);
+   * bound(); // will execute my_function(2, 3.0);
+   *
+   * bound.parse_parameters("3: 4.0");
+   * bound(); // will execute my_function(3, 4.0);
+   * @endcode
+   *
+   * Notice that, if you do not call one of the MutableBind::set_parameters()
```
bangerth:
```suggestion
   * Notice that if you do not call one of the MutableBind::set_parameters()
```

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+   * methods, or the MutableBind::parse_parameters(), the arguments passed to
```
bangerth:
```suggestion
   * methods, or the MutableBind::parse_parameters() function, on the returned object, then the arguments passed to
```

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+   * the function object will be initialized with the values coming from the
+   * arguments defalt constructors.
```
bangerth:
```suggestion
   * arguments' default constructors.
```

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+   * bound.set_parameters(2, 3.0);
+   * bound(); // will execute my_function(2, 3.0);
+   *
+   * bound.parse_parameters("3: 4.0");
+   * bound(); // will execute my_function(3, 4.0);
+   * @endcode
+   *
+   * Notice that, if you do not call one of the MutableBind::set_parameters()
+   * methods, or the MutableBind::parse_parameters(), the arguments passed to
+   * the function object will be initialized with the values coming from the
```
bangerth:
```suggestion
   * the function object will be initialized with the values coming from each of the
```

include/deal.II/base/mutable_bind.h
bangerth: Maybe also say that this can only work if each argument type has a default constructor.

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+   * bound.parse_parameters("3: 4.0");
+   * bound(); // will execute my_function(3, 4.0);
+   * @endcode
+   *
+   * @authors Luca Heltai, Matthias Maier, 2019.
+   */
+  template <typename ReturnType, class... FunctionArgs>
+  MutableBind<ReturnType, FunctionArgs...>
+  mutable_bind(std::function<ReturnType(FunctionArgs...)>,
+               typename identity<FunctionArgs>::type... arguments);
```
bangerth: Here and in the other places, you could also do the rvalue reference thingy. It's going to lead to a proliferation of functions, though.

It might be easier if you didn't duplicate the code snippet for every function, but only said "Like the previous function, except ..."

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+  {
+    return std_cxx17::apply(function, arguments);
+  }
+
+
+
+  template <typename ReturnType, class... FunctionArgs>
+  void
+  MutableBind<ReturnType, FunctionArgs...>::set_arguments(const TupleType &args)
+  {
```
bangerth: move it next to the other `set_arguments` function.

include/deal.II/base/mutable_bind.h
```diff
@@ -0,0 +1,320 @@
...
+    return {function,
+            typename MutableBind<ReturnType, FunctionArgs...>::TupleType()};
+  }
+
+
+} // namespace Utilities
+
+DEAL_II_NAMESPACE_CLOSE
+
+#endif
```
bangerth: needs a newline

tests/base/mutable_bind_01.cc
```diff
@@ -0,0 +1,51 @@
...
+  exp.parse_arguments("2.0 , 2.0 : 1.0 : 4");
+  exp();
+
+  exp.set_arguments(Point<2>(.3, .3), 2.0, 5);
+  exp();
+
+  auto exp1 =
+    Utilities::mutable_bind(example_function, Point<2>(.8, .8), 4.0, 10);
+  exp1();
+}
```
bangerth: newline

tests/base/mutable_bind_02.cc
```diff
@@ -0,0 +1,55 @@
...
+
+  exp.parse_arguments("1.0 : 4");
+  exp();
+
+  exp.set_arguments(2.0, 5);
+  exp();
+
+  auto exp1 = Utilities::mutable_bind(reduced_function, 4.0, 10);
+  exp1();
+}
```
bangerth: newline

include/deal.II/base/mutable_bind.h
luca-heltai: This you should ask @tamiko :D

bangerth: If there isn't a good reason, then I'd suggest making it a `class` and make most of its member `private` :-)

luca-heltai:
> I think the code example about the bound reference argument is too complicated -- one could just use `std::bind(...)` to bind the reference argument with `std::ref(tria)` and leave the other two arguments open as `std::placeholders::_1/2`. Or maybe show that as an alternative?

Sure. I guess this is a matter of taste. I personally find the use of `std::bind` much less clear than lambda expressions, but that's just me.

> I did think about whether it's possible to just layer your work on top of what `std::bind` does, because that would allow to use `std::ref` and `std::cref` when creating the argument tuple. I think that could actually work if instead of calling your own `apply` function, you'd do something like this:

I'm not sure I get what you want to say here.

Suppose I have the following: 
```
void some_function(const Triangulation<dim> &tria, const Point<dim> &p);
```
My objective here would be to block the first argument (to tria), and make the second accessible to the `mutable_bind` class. I do *not* want to block p to be a reference (or a const reference) to something *already existing*. 

What I wanted  was to create a compatible tuple (in this case containing only a single element: a `Point<dim>`), store it in the simplest possible way (a struct is a good choice in this case), and make sure I can parse a string to fill the tuple. 

What would you exactly want to do in allowing the tuple to store a `std::ref` or a `std::cref`? Maybe make it such that when I parse the tuple, I actually change an *existing* point, instead of creating one (by copy) in the tuple? This is in principle ok for the `std::ref`, however it would not be currently supported for references in our `Convert<T>::to_value("some string");`, since we return `T` by value (this was your call during the review process of the `Patterns::Tools::Convert<T>` class. In the original implementation I had `Convert<T>::to_value("some string", T& val);` which would have allowed this for `std::ref` objects. However, this would not work at all for `std::cref`.

Moreover, when examining the function signature, I have no way of figuring out if the point `const Point<dim> &p` should be treated as a `std::ref` or a `std::cref`, and the user has to make a choice explicitly. Even when the user correctly set something to be a `std::cref`, its presence in the tuple is useless, as I would not be able to parse anything into it. I'd have to scan all arguments, and selectively eliminate those that are `std::crefs` from the tuple.

The last point, is that I have no idea how to make the `mutable_bind` function work with std::bind (as you suggested):
```
auto bound_function = mutable_bind(std::bind(some_function, std::cref(tria), std::place_holder::_1);
```

The reason is that I have found no way to deduce the arguments of the resulting type (unspecified, according to the standard). See also here: 

https://stackoverflow.com/questions/29259433/deducing-parameter-types-from-bind-expression

The conclusion of that article was: don't use `std::bind`.

I can, however, do the following:
```
  void
  example_function(const Point<2> &p, const double &d, const unsigned int i = 3);
  const Point<2> p(1, 2);

  Utilities::MutableBind<void, double, unsigned int> exp = {
    std::bind(example_function,
              std::cref(p),
              std::placeholders::_1,
              std::placeholders::_2),
    {}};

  exp.parse_arguments("3.0 : 4");
  exp(); // Calls example_function(p, 3.0, 4);
```
and this now replaces the version with the lambda (wiriting a bit more in the `MutableBind` explicit construction, which now needs to know its template arguments explicitly). Incidentally, and here you also see why I'd like to keep this as a struct: very easy to assign to it `exp = {f, {}}`.

include/deal.II/base/mutable_bind.h
luca-heltai: The reason is practical: this is a very simple object, that holds a function and a tuple. As a struct, it is very easy to assign to it: 
```
void f(double d);
MutableBind<void, double> exp = {f, {}};
```
I'd like to keep this possibility.

bangerth: Hm, but that assumes that you know the layout. And it means that you can never *change* the layout either. If you like this assignment type, it would be easiest to just write a constructor for it.

masterleinad:
```
/jenkins/workspace/dealii_PR-8530/tests/base/mutable_bind_01.cc: In function ‘int main()’:
/jenkins/workspace/dealii_PR-8530/tests/base/mutable_bind_01.cc:49:72: error: call of overloaded ‘mutable_bind(void (&)(const dealii::Point<2>&, const double&, unsigned int), dealii::Point<2>, double, int)’ is ambiguous
Utilities::mutable_bind(example_function, Point<2>(.8, .8), 4.0, 10);
^
In file included from /jenkins/workspace/dealii_PR-8530/tests/base/mutable_bind_01.cc:19:0:
/jenkins/workspace/dealii_PR-8530/include/deal.II/base/mutable_bind.h:306:3: note: candidate: dealii::Utilities::MutableBind<ReturnType, FunctionArgs ...> dealii::Utilities::mutable_bind(ReturnType (*)(FunctionArgs ...), typename dealii::identity<FunctionArgs>::type ...) [with ReturnType = void; FunctionArgs = {const dealii::Point<2, double>&, const double&, unsigned int}]
mutable_bind(ReturnType (*function)(FunctionArgs...),
^
/jenkins/workspace/dealii_PR-8530/include/deal.II/base/mutable_bind.h:316:3: note: candidate: dealii::Utilities::MutableBind<ReturnType, FunctionArgs ...> dealii::Utilities::mutable_bind(ReturnType (*)(FunctionArgs ...), typename dealii::identity<FunctionArgs>::type&& ...) [with ReturnType = void; FunctionArgs = {const dealii::Point<2, double>&, const double&, unsigned int}]
mutable_bind(ReturnType (*function)(FunctionArgs...),
^
```

```
/jenkins/workspace/dealii_PR-8530/include/deal.II/base/mutable_bind.h:336:3: note: candidate: dealii::Utilities::MutableBind<ReturnType, FunctionArgs ...> dealii::Utilities::mutable_bind(std::function<_Res(_ArgTypes ...)>, typename dealii::identity<FunctionArgs>::type ...) [with ReturnType = void; FunctionArgs = {double, unsigned int}]
mutable_bind(std::function<ReturnType(FunctionArgs...)> function,
^
/jenkins/workspace/dealii_PR-8530/include/deal.II/base/mutable_bind.h:223:3: note: candidate: dealii::Utilities::MutableBind<ReturnType, FunctionArgs ...> dealii::Utilities::mutable_bind(std::function<_Res(_ArgTypes ...)>, typename dealii::identity<FunctionArgs>::type&& ...) [with ReturnType = void; FunctionArgs = {double, unsigned int}]
mutable_bind(std::function<ReturnType(FunctionArgs...)>,
^
```

include/deal.II/base/mutable_bind.h
luca-heltai: Done. Left three constructors.

bangerth:
I think where I was going with this is the following: In your example with the function
```
  void some_function(const Triangulation<dim> &tria, const Point<dim> &p);
```
it is currently not possible to use your `MutableBind` class since only copyable objects can be used as arguments. But `Triangulation` is not copyable. The trick behing `std::ref` is that `std::ref` objects *are* copyable, even if the referenced objects are not. So you can do
```
  std::bind (&some_function, std::ref(my_triangulation), my_point);
```
which creates an object where `my_triangulation` is only stored as a reference and `my_point` is actually copied. I was hoping that one could do something similar with your class.

I *think* that you can in essence achieve this in the following way:
```
  class MutableBind {
     std::function<ReturnType (Args...)> unbound_function;
     std::function<ReturnType ()>           bound_function;

     // usual constructor. create the fully bound object with all arguments set;
     // use std::bind so that we can allow for some of the args... to be std::ref/cref;
     // also store the unbound function because we want to be able to re-bind
     // arguments later on
     MutableBind (std::function<ReturnType (Args...)> &f,
                           Args... args)
       : unbound_function(f),
         bound_function (std::bind(f, ...expand args...))
      {}

      // operator() simply calls the fully bound object
      ReturnType operator() { return bound_function(); }

      // rebinding arguments: re-uses the unbound function and simply
      // creates a newly bound object; again uses std::bind to
      // allow using std::ref/cref
      void set_arguments (Args... args) {
         bound_function = std::bind (unbound_function, ...expand args...);
  };
```

This way you can do
```
  // create a bound function with set arguments, one of which happens to be a reference
  auto f = mutable_bind (&some_function, std::ref(my_triangulation), my_point);

  // ah, never mind, I'll change my mind:
  f.set_arguments (std::ref(some_other_triangulation), some_other_point);
```
In both cases, the second argument is copied, not taken by reference.


As I said before, I don't know whether this is all necessary, but it's worth thinking things through before we commit to a design :-)

luca-heltai:
This you can already do. I added another test that shows exactly your point:

```
  const Point<2> p(1, 2);
  const Point<2> p2(2, 3);

  Utilities::MutableBind<void,
                         std::reference_wrapper<const Point<2>>,
                         double,
                         unsigned int>
    exp(std::bind(example_function,
                  std::placeholders::_1,
                  std::placeholders::_2,
                  std::placeholders::_3),
        std::cref(p),
        3.0,
        4);

  exp.set_arguments(std::cref(p), 3.0, 4);
  exp();
  exp.set_arguments(std::cref(p2), 4.0, 5);
```
The only problem here, is that there is no way to **parse** the arguments at this point, since I have no way, from a text file, to figure out what reference you want. You can still use mutable bind as mutable wrapper around `std::bind`, just as it is done in test `mutable_bind_04.cc`. Parsing will fail.

PR #8533, opened 2019-08-09
Variable size transfer of quadrature point data implemented

### LINKED ISSUE #8506: ContinuousQuadratureDataTransfer with non-fixed-size quadrature data structures
After today's talk about hp-adaptivity at the workshop, I had a conversion with @peterrum and @marcfehling about using `ContinuousQuadratureDataTransfer` where the data type stored at quadrature points might be different from a cell to another. However, the current implementation of packing and unpacking functions within `ContinuousQuadratureDataTransfer` enforces that all the cells in the triangulation have data structures of the same fixed size (returned by `DataType::number_of_values()`). I assume `ContinuousQuadratureDataTransfer` was written at a time when variable size cell data transfer between subdomains were not fully implemented. However, @marcfehling believes, because of recent advancements in data transfer algorithms made for hp parallel algorithms, we can have variable sized data transfer without any noticeable performance overhead.

What do you think, @jppelteret ? Do you think this is something possible and desirable?

In my structural FE simulations, a lot of times most of the mesh contains elastic materials which don't need any history data accosiated with them. But a small portion of the mesh contains a more complicated material with plasticity/damage/etc that requires complicated storage of many history variables per quadrature point. Currently, using the helpful facilities in `ContinuousQuadratureDataTransfer` and `CellDataStorage`, I have to make sure all cells in the mesh, even the elastic ones, are initialized with history data types that all return the same value from the method `number_of_values()`. The elastic cells, at every refinement step, transfer and interpolate multiple trivial zeros to initialize child members. I have to make these unnecessary data transfers and use strange workaround to allow my code to use multiple materials across an adaptively refined mesh. My objective is to allow any cell in the mesh:

- [ ]  to have no structure attached to it.
- [ ]  to have a structure attached to it which may differ in size between cell to cell.
- [ ]  to have a structure that may grow in time. For example, a cell might be pristine for most of the simulation with elastic response and suddenly cracks after the mechanical stress reaches a critical value. Now the cell needs to store damage and crack opening data at every quadrature point.

rezarastak:
Fixes #8506 

This PR is not complete right now, it seems to work though. Looking forward to your feedback.

Progress:
- [x] `ContinuousQuadratureDataTransfer` transfers variables size quadrature data
- [x] Passes all existing tests.
- [ ] Properly handle the case where some cells don't need to store any data ([this comment](https://github.com/dealii/dealii/issues/8506#issuecomment-519935629))
- [ ] Add a new test.

 @jppelteret @davydden @marcfehling @peterrum


include/deal.II/base/quadrature_point_data.h
```diff
@@ -732,18 +732,17 @@ pack_cell_data(const CellIteratorType &                           cell,
   const std::vector<std::shared_ptr<const DataType>> qpd =
     data_storage->get_data(cell);
 
-  const unsigned int n = matrix_data.n();
+  const auto m = qpd.size();
+  const auto n = qpd[0]->number_of_values();
```
davydden: should assert that m>0

marcfehling: ...or change both variables `m` and `n` to be of type `const unsigned int`.

I would suggest to refrain from using `auto` here since `unsigned int` makes the code more easily readable.

marcfehling:
Nice job! There were really just a few lines that had to changed :-)
I'm curious about the test you are about to submit!

include/deal.II/base/quadrature_point_data.h
jppelteret: I believe that what @davydden is suggesting is that `qpd[0]` actually exists. So `Assert(m>0)` does seem appropriate here.

rezarastak:
This PR is ready to be reviewed. It has all the required new functionality and has appropriate new tests.

tests/base/quadrature_point_data_04.cc
```diff
@@ -0,0 +1,247 @@
...
+    {
+      const auto center = cell->center();
+      if (center[0] < 0.5 && center[1] < 0.5)
+        cell->recursively_set_material_id(0);
+      else if (center[0] >= 0.5 && center[1] < 0.5)
+        cell->recursively_set_material_id(1);
+      else if (center[0] < 0.5 && center[1] >= 0.5)
+        cell->recursively_set_material_id(2);
+      else
+        // material_id 3 does not have an associated data structure
```
peterrum: You can remove the comment here.

tests/base/quadrature_point_data_04.cc
```diff
@@ -0,0 +1,247 @@
...
+    if (cell->is_locally_owned())
+      {
+        // It does not modify the data on already initialized cell data
+        if (cell->material_id() == 0)
+          storage.template initialize<Mat0>(cell, n_data_points_per_cell);
+        else if (cell->material_id() == 1)
+          storage.template initialize<Mat1>(cell, n_data_points_per_cell);
+        else if (cell->material_id() == 2)
+          storage.template initialize<Mat2>(cell, n_data_points_per_cell);
+      }
```
peterrum: Move the comment `// material_id 3 does not have an associated data structure` here.

tests/base/quadrature_point_data_04.cc
```diff
@@ -0,0 +1,247 @@
...
+        // material_id 3 does not have an associated data structure
+        cell->recursively_set_material_id(3);
+    }
+
+  // create a history structure and populate it
+  QuadratureStorage storage;
+  initialize_data(tria, storage);
+
+  // Assign some value to the cell data
+  FE_Q<2>     fe(1);
```
peterrum: Why is `degree` not  a constexpr, but `n_quadrature_per_dim`?

tests/base/quadrature_point_data_04.cc
```diff
@@ -0,0 +1,247 @@
...
+void
+initialize_data(const Triangulation<2> &tria, QuadratureStorage &storage)
+{
+  deallog << "Initializing quadrature cell data" << std::endl;
+  for (auto cell : tria.active_cell_iterators())
+    if (cell->is_locally_owned())
+      {
+        // It does not modify the data on already initialized cell data
+        if (cell->material_id() == 0)
+          storage.template initialize<Mat0>(cell, n_data_points_per_cell);
```
peterrum: I would instead of using `n_data_points_per_cell` use `quad->size()`

tests/base/quadrature_point_data_04.cc
```diff
@@ -0,0 +1,247 @@
...
+    AssertDimension(values.size(), number_of_values());
+  }
+  virtual void
+  unpack_values(const std::vector<double> &values) final
+  {
+    AssertDimension(values.size(), number_of_values());
+  }
+};
+
+constexpr unsigned int n_quadrature_per_dim = 3;
```
peterrum: I would move this line after the includes.

tests/base/quadrature_point_data_04.cc
```diff
@@ -0,0 +1,247 @@
...
+constexpr unsigned int n_data_points_per_cell =
```
peterrum: I would remove this line; see also other comments.

include/deal.II/base/quadrature_point_data.h
```diff
@@ -707,6 +713,14 @@ CellDataStorage<CellIteratorType, DataType>::get_data(
   return res;
 }
 
+template <typename CellIteratorType, typename DataType>
+bool
+CellDataStorage<CellIteratorType, DataType>::stores_data(
+  const CellIteratorType &cell) const
```
bangerth: You might want to mark the function as `inline` since it lives in a header file.

include/deal.II/base/quadrature_point_data.h
```diff
@@ -168,6 +168,12 @@ class CellDataStorage : public Subscriptor
   std::vector<std::shared_ptr<const T>>
   get_data(const CellIteratorType &cell) const;
 
+  /**
+   * Queries whether a data belonging to @p cell exists.
+   */
+  bool
+  stores_data(const CellIteratorType &cell) const;
```
bangerth: It's a shame that we didn't anticipate the fact that some cells might not have data associated with them, or we could have let `get_data()` return a `boost::optional` object.

include/deal.II/base/quadrature_point_data.h
```diff
@@ -729,21 +743,27 @@ pack_cell_data(const CellIteratorType &                           cell,
     std::is_base_of<TransferableQuadraturePointData, DataType>::value,
     "User's DataType class should be derived from QPData");
 
+  if (!data_storage->stores_data(cell))
```
bangerth: In most places, we use this instead:
```suggestion
  if (data_storage->stores_data(cell) == false)
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -729,21 +743,27 @@ pack_cell_data(const CellIteratorType &                           cell,
...
+    {
+      matrix_data = FullMatrix<double>();
```
bangerth:
```suggestion
      matrix_data.clear();
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -729,21 +743,27 @@ pack_cell_data(const CellIteratorType &                           cell,
...
-  for (unsigned int q = 0; q < qpd.size(); q++)
+  for (unsigned int q = 0; q < m; ++q)
     {
       qpd[q]->pack_values(single_qp_data);
-      Assert(single_qp_data.size() == n,
-             ExcDimensionMismatch(single_qp_data.size(), n));
+      AssertDimension(single_qp_data.size(), n);
 
-      for (unsigned int i = 0; i < n; i++)
+      for (unsigned int i = 0; i < n; ++i)
```
bangerth: :-)

include/deal.II/base/quadrature_point_data.h
```diff
@@ -763,17 +783,20 @@ unpack_to_cell_data(const CellIteratorType &                     cell,
     std::is_base_of<TransferableQuadraturePointData, DataType>::value,
     "User's DataType class should be derived from QPData");
 
+  if (!data_storage->stores_data(cell))
```
bangerth: same here

include/deal.II/base/quadrature_point_data.h
rezarastak: I also think changing the output of `get_data()` to give additional info on whether a data exists would have been a much better design. We discussed it in #8506 and decided to go with this design (adding member function `stores_data()` instead).

I prefer writing
```c++
auto qpd = data_storage(cell);
if (!qpd)
  return;
```
instead of the current expression
```c++
if (!data_storage->stores_data(cell))
    return;
auto qpd = data_storage->get_data(cell);
```
because the latter requires two map lookups instead of just one lookup.
However, backward compatibility limits us a bit here because it forces us to choose the second design.

include/deal.II/base/quadrature_point_data.h
rezarastak: All of the function definitions in this file don't explicitly contain the `inline` keyword but they are all implicitly inline because they are templated.

include/deal.II/base/quadrature_point_data.h
bangerth: Or do `if (auto data = data_storage->get_data) { ... }` :-)

Like I mentioned elsewhere, we always have the option of deprecating `get_data()` and invent a new function that has the right interface.

include/deal.II/base/quadrature_point_data.h
bangerth: That's not correct. `template` functions have weak linkage, but they're not automatically `inline`.

rezarastak: Okay! Do you think other function definitions in this header file need `inline` keyword as well? or is it just this `stores_data()` method because it is small?

include/deal.II/base/quadrature_point_data.h
rezarastak: I couldn't find the `clear()` method in the `FullMatrix<Number>` interface. I used the `reinit({0, 0})` method instead.

tests/base/quadrature_point_data_04.cc
rezarastak: I removed the consexpr global variable and made it more consistent.

include/deal.II/base/quadrature_point_data.h
bangerth: In principle, every function that is small enough that inlining makes sense. That requires inspection of every function defined in the .h file.

include/deal.II/base/quadrature_point_data.h
bangerth: Yes, that makes sense.

include/deal.II/base/quadrature_point_data.h
jppelteret: I'm happy if you want to follow through on @bangerth's suggestion if it seems like the better solution to the majority of people. @bangerth Maybe you wouldn't mind reviewing this again once its been updated -- I'm still short on time ATM.

bangerth: @jppelteret -- yes, gladly!
@rezarastak -- let me know once you want me to take another look!

include/deal.II/base/quadrature_point_data.h
```diff
@@ -897,10 +895,11 @@ namespace parallel
...
       // project to FE
-      project_to_fe_matrix.mmult(matrix_dofs, matrix_quadrature);
+      const unsigned int number_of_values = matrix_quadrature.n();
+      matrix_dofs.reinit(project_to_fe_matrix.m(), number_of_values);
+      if (number_of_values > 0)
+        project_to_fe_matrix.mmult(matrix_dofs, matrix_quadrature);
 
-      // to get consistent data sizes on each cell for the fixed size transfer,
-      // we won't allow compression
       return Utilities::pack(matrix_dofs, /*allow_compression=*/false);
```
marcfehling: Please consider introducing an optimized pack function.

include/deal.II/base/quadrature_point_data.h
```diff
@@ -925,11 +924,18 @@ namespace parallel
         Utilities::unpack<FullMatrix<double>>(data_range.begin(),
```
marcfehling: Please consider introducing an optimized unpack function.

rezarastak:
@marcfehling I suggest performing this pack/unpack optimization at a separate PR in the future. We first have to use this variable quadrature data transfer feature and benchmark its performance before deciding whether we want to specialize the pack/unpack functions.

bangerth:
I agree with the approach of doing one thing at a time.

rezarastak:
Sorry @bangerth, I haven't yet implemented the modified `get_data()` method that returns an optional yet. Give me another day or two to finish that before merging this.

masterleinad:
Just tell us when you are ready and we should have another look.

rezarastak:
I added the function `try_get_data()` that returns an optional if data is not available. I decided to not deprecate `get_data()` because it is useful as it provides a simpler interface for the cases where we are sure that data exists for all cells. For example, step-44 would have become unnecessarily complicated if I replaced `get_data()` with `try_get_data()`.

This PR is ready to be reviewed/tested.

doc/news/changes/minor/20190821RezaRastak
```diff
@@ -0,0 +1,5 @@
+New: method try_get_data() added to CellDataStorage in both const and
+non-const forms. It returns an optional which determines whether a
```
masterleinad:
```suggestion
non-const forms. It returns an std_cxx17::optional which determines whether a
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -168,6 +169,44 @@ class CellDataStorage : public Subscriptor
   std::vector<std::shared_ptr<const T>>
   get_data(const CellIteratorType &cell) const;
 
+  /**
+   * Returns an optional indicating whether @p cell contains an associated
```
masterleinad:
```suggestion
   * Returns an std_cxx17::optional indicating whether @p cell contains an associated
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -168,6 +169,44 @@ class CellDataStorage : public Subscriptor
...
+   * data or not. If data is available, dereferencing the optional
```
masterleinad:
```suggestion
   * data or not. If data is available, dereferencing the std_cxx17::optional
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -168,6 +169,44 @@ class CellDataStorage : public Subscriptor
...
+   * @pre The type @p T needs to match the class provided to initialize().
+   * @pre @p cell must be from the same Triangulation that is used to
+   * initialize() the cell data.
+   */
+  template <typename T = DataType>
+  std_cxx17::optional<std::vector<std::shared_ptr<T>>>
+  try_get_data(const CellIteratorType &cell);
+
+  /**
+   * Returns an optional indicating whether @p cell contains an associated
```
masterleinad:
```suggestion
   * Returns an std_cxx17::optional indicating whether @p cell contains an associated
```

include/deal.II/base/quadrature_point_data.h
```diff
@@ -168,6 +169,44 @@ class CellDataStorage : public Subscriptor
...
+   * data or not. If data is available, dereferencing the optional reveals
```
masterleinad:
```suggestion
   * data or not. If data is available, dereferencing the std_cxx17::optional reveals
```

PR #8536, opened 2019-08-09
Make particle iterators const references in range-based for loops.

bangerth:
I didn't convert them all so we can continue to test both ways to write a loop.


bangerth:
@gassmoeller -- FYI.

masterleinad:
> I didn't convert them all so we can continue to test both ways to write a loop.

Can you clarify this comment? Do you actually want to measure the difference in terms of performance between the two loop variants?

bangerth:
I meant to say that there are other tests that I didn't touch. I have no doubt that the difference is really immeasurable.

PR #8537, opened 2019-08-09
Initialize particle locations with invalid numbers.

bangerth:
The previous version just put the particle onto the origin, but that seems as arbitrary as any other way. So just give it an invalid location and throw a SIGFPE if it is accessed.

I also like the idea of instantiating a particle in nirvana, with no actual location. Setting the location then yields actually instantiates the particle in real life.

@gassmoeller -- FYI.

### DIFF
--- a/source/particles/particle.cc
+++ b/source/particles/particle.cc
@@ -13,6 +13,8 @@
 //
 // ---------------------------------------------------------------------
 
+#include <deal.II/base/signaling_nan.h>
+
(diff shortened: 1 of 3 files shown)


masterleinad:
> I will report back once I had the chance to test it.

Let's put a label for not merging then.

gassmoeller:
Actually the tests just passed (5 min ago). Ready to go from my side :smile:.

PR #8539, opened 2019-08-09
Improve resolution in step 8.

chronictectonic:
Improved resolution to generate better looking figures (attached here).
step-8 x
step-8 y

### DIFF
--- a/examples/step-8/step-8.cc
+++ b/examples/step-8/step-8.cc
@@ -622,7 +622,7 @@ namespace Step8
         if (cycle == 0)
           {
             GridGenerator::hyper_cube(triangulation, -1, 1);
-            triangulation.refine_global(2);
+            triangulation.refine_global(4);
           }
         else
           refine_grid();


drwells:
Looks good! Would you please crop the images so that we do not see the coordinates axes?

chronictectonic:
Thanks! I noticed that shortly after uploading.
step-8 x
step-8 y

chronictectonic:
And the vector field
step-8 vectors

bangerth:
This increases run-time of this program from 4 to 40 seconds -- which seems entirely reasonable.

@chronictectonic -- do you think you could come up with a small changelog entry that, in one line, explains what you changed? Let us know once you've got it added and we can merge!

bangerth:
I've updated the pictures on the website with the ones from here.

bangerth:
I've used the pictures in #8557, by the way. Thanks for creating them!

PR #8542, opened 2019-08-09
Added the function GridGenerator::subdivided_hyper_L().

maemar:
~~Still need to add results to output file for testing.~~
Only works in 2D and 3D.
Subdivided_hyper_L is a generalization of the hyper_L domain. Essentially allows the user to create a subdivided hyper rectangle of arbitrary size, and remove a select number of cells from that domain. The user must provide a triangulation, points representing the lower left and upper right corners of the domain, a vector of unsigned integers representing the number of subdivisions in each direction, and a vector of integers where the ith coordinate represents how many cells to cut away in the ith direction. A positive number denotes cutting in the positive orientation: i.e. cutting left to right, bottom to top, and front to back; a negative number represents cutting right to left, top to bottom, and back to front.
Useful to generate the backward facing step, a common benchmark in fluid dynamics.
backward_facing_step3d


marcfehling:
Would you add a changelog entry to `doc/news/changes/minor/`? Just copy one of the files you find there and adjust them to your needs. These entries will be published [here](https://www.dealii.org/current/doxygen/deal.II/changes_between_8_5_0_and_9_0_0.html) for the next release.

If you want to, you can squash your commits into one afterwards with `git rebase -i`.

marcfehling:
Since documentation and the test output is provided now, you may also edit your description above. You may use `~~text~~` to dash previous text like this ~~text~~.

source/grid/grid_generator.cc
```diff
@@ -2891,6 +2891,81 @@ namespace GridGenerator
...
+        h.push_back(
+          (top_right[d] - bottom_left[d]) /
+          repetitions.at(
+            d)); // mesh spacing in each direction in cartesian coordinates
+        // left to right, bottom to top, front to back
+        if (num_cells_to_cut.at(d) >= 0)
+          {
+            // cartesian coordinates of vertex location
+            cut_step[d] =
+              h.at(d) * std::fabs(num_cells_to_cut.at(d)) + bottom_left[d];
```
masterleinad: Since we already know the size of `h`, I see no reason for using `std::vector::at` instead of `std::vector::operator[]` here and below.

source/grid/grid_generator.cc
```diff
@@ -2891,6 +2891,81 @@ namespace GridGenerator
 
 
 
+  template <int dim, int spacedim>
+  void
+  subdivided_hyper_L(Triangulation<dim, spacedim> &   tria,
+                     const Point<dim> &               bottom_left,
+                     const Point<dim> &               top_right,
+                     const std::vector<unsigned int> &repetitions,
```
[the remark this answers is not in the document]
marcfehling: I think we kept it like that to keep the parameter types similar to the `subdivided_hyper_rectangle()` function, which also uses `std::vector`.

We should probably change all affected functions of the whole `GridGenerator` namespace to use `std::array`, but in a separate PR.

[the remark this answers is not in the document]
maemar: Exactly what Marc said--I was just trying to make it consistent with the rest of the GridGenerator functions, which use vectors.

source/grid/grid_generator.cc
maemar: Just new to C++ and wasn't sure what the common practice was, but I wasn't even being consistent with my own code and had a mix of .at(d) and [d]. Changed it all to [d], thanks for pointing this out!

source/grid/grid_generator.cc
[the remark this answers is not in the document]
masterleinad: Fair enough. Let's discuss using `std::array` in all the `GridGenerator` functions somewhere else.

marcfehling:
@luca-heltai -- FYI -- Do you have any comments on this new `GridGenerator` function?

doc/news/changes/minor/20190810MaeMarkowski
```diff
@@ -0,0 +1,3 @@
+New: Added the function subdivided_hyper_L to GridGenerator.
```
marcfehling: This way, `doxygen` will be able to point directly to the new function.
```suggestion
New: Added the function GridGenerator::subdivided_hyper_L().
```

include/deal.II/grid/grid_generator.h
```diff
@@ -873,6 +873,45 @@ namespace GridGenerator
...
+   * dimensions, but throws an error if called in 1D.
+   *
+   * @author Mae Markowski, 2019
+   */
+  template <int dim, int spacedim>
+  void
+  subdivided_hyper_L(Triangulation<dim, spacedim> &   tria,
+                     const Point<dim> &               bottom_left,
+                     const Point<dim> &               top_right,
+                     const std::vector<unsigned int> &repetitions,
```
marcfehling: I just realized that the parameter order of this function is different to all other `GridGenerator::subdivided_*` functions: They all start with `repetitions` followed by the start and end points, while this function does it the other way around.

Do you mind changing the order of the function parameters? It is up to you where you want to place `num_cells_to_cut`, though, but I would suggest to keep it as the last parameter. For this, you would need to update the `.cc`, `.inst.in` files and the test as well.

maemar: changed the test, source, header, and inst files. good catch--thanks for the suggestion

doc/news/changes/minor/20190810MaeMarkowski
marcfehling: @maemar Would you also take care of this before merging?

source/grid/grid_generator.cc
```diff
@@ -2891,6 +2891,77 @@ namespace GridGenerator
...
+                     const std::vector<int> &         num_cells_to_cut)
+  {
+    Assert(dim > 1, ExcNotImplemented());
+    // Check the consistency of the dimensions provided.
+    AssertDimension(repetitions.size(), dim);
+    AssertDimension(num_cells_to_cut.size(), dim);
+    for (unsigned int d = 0; d < dim; ++d)
+      {
+        Assert(std::fabs(num_cells_to_cut[d]) <= repetitions[d],
+               ExcInternalError("Attempting to cut away too many cells."));
```
marcfehling:
```suggestion
               ExcMessage("Attempting to cut away too many cells."));
```

doc/news/changes/minor/20190810MaeMarkowski
maemar: Ah yes, sorry--I thought this was referring to the name of the PR on github. This is in the log file right?

marcfehling: Exactly. This suggestion referred to your changelog entry :)

include/deal.II/grid/grid_generator.h
```diff
@@ -873,6 +873,45 @@ namespace GridGenerator
...
+   *
+   * @author Mae Markowski, 2019
+   */
+  template <int dim, int spacedim>
+  void
+  subdivided_hyper_L(Triangulation<dim, spacedim> &   tria,
+                     const std::vector<unsigned int> &repetitions,
+                     const Point<dim> &               bottom_left,
+                     const Point<dim> &               top_right,
+                     const std::vector<int> &         num_cells_to_cut);
```
bangerth: I would call this argument `n_cells_to_remove`. The `n_` prefix is what we usually use, and "remove" just seems like an easier to understand "verb" than "cut" for non-native speakers in this context.

source/grid/grid_generator.cc
```diff
@@ -2891,6 +2891,77 @@ namespace GridGenerator
...
+      }
+    // Create the domain to be cut
+    Triangulation<dim, spacedim> rectangle;
+    GridGenerator::subdivided_hyper_rectangle(rectangle,
+                                              repetitions,
+                                              bottom_left,
+                                              top_right);
+    // compute the vertex of the cut step, we will cut according to the
+    // location of the cartesian coordinates of the cell centers
+    std::vector<double> h;
```
bangerth: I think you could just make this a `std::array<double,dim> h` to make its dimensions clear.

source/grid/grid_generator.cc
```diff
@@ -2891,6 +2891,77 @@ namespace GridGenerator
...
+    Point<dim>          cut_step;
+    for (unsigned int d = 0; d < dim; ++d)
+      {
+        // mesh spacing in each direction in cartesian coordinates
+        h.push_back((top_right[d] - bottom_left[d]) / repetitions[d]);
```
bangerth: and then this would just be `h[d] = ...`

source/grid/grid_generator.cc
```diff
@@ -2891,6 +2891,77 @@ namespace GridGenerator
...
+
+
+    // compute cells to remove
+    std::set<typename Triangulation<dim, spacedim>::active_cell_iterator>
+      cells_to_remove;
+    std::copy_if(
+      rectangle.active_cell_iterators().begin(),
+      rectangle.active_cell_iterators().end(),
+      std::inserter(cells_to_remove, cells_to_remove.end()),
+      [&](
```
bangerth: Oh, this is an elegant way to do this! Well done!

maemar: Big thanks to @tamiko and @marcfehling !

luca-heltai:
Oh, I like this! :)

PR #8543, opened 2019-08-09
Fixed documentation for Triangulation<...>::set_manifold(...)

### LINKED ISSUE #8309: Documentation of Triangulation::set_manifold is incorrect
The documentation of Triangulation::set_manifold states:

The manifold_object is not copied and MUST persist until the triangulation is destroyed. This is also true for triangulations generated from this one by copy_triangulation.

However, the grid generator routines create a lot of manifolds from temporary objects and the implementation of set_manifold (https://www.dealii.org/current/doxygen/deal.II/grid_2tria_8cc_source.html#l08955) shows that this is not the case:

`manifold[m_number] = manifold_object.clone();`

alexrobomind:
Fixes  #8309

The documentation previously indicated that set_manifold will not create a copy and the lifetime of the manifold must be managed manually. However, this is false. Instead, a copy is created using Manifold::clone() and stored.

### DIFF
--- a/include/deal.II/grid/tria.h
+++ b/include/deal.II/grid/tria.h
@@ -1780,9 +1780,8 @@ public:
    * for non-linear (i.e.: non-Q1) transformations of cells to the unit cell
    * in shape function calculations.
    *
-   * The @p manifold_object is not copied and MUST persist until the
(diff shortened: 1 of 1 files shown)


bangerth:
Great, thanks for fixing this!

PR #8545, opened 2019-08-10
Fix several typos in CMake documentation

brownraming:
Found couple typos in the CMake documentation and fixed them.


doc/users/cmakelists.html
```diff
@@ -560,7 +560,7 @@ <h3>External libraries</h3>
 FIND_PACKAGE(foo 8.0 REQUIRED)
 </pre>
 Alternatively, the version number and <code>REQUIRED</code> keyword can be
-omitted. (Depending on the external library) the project configuration or
```
masterleinad: This looks actually correct. You might remove the parentheses, though.

brownraming: You are right. I deleted the parentheses and modified the change.

PR #8546, opened 2019-08-10
Remove diff from testsuite

drwells:
The test suite now officially requires numdiff. I discussed this with @tamiko.

I also fixed a minor configuration bug.


drwells:
The large diff comes from converting our single binary output test to use base64 output.

masterleinad:
Hmm.. Looks like the test you changed doesn't pass anymore. Does it work locally?

drwells:
I got different output files on my laptop and workstation. I added the new file.

masterleinad:
Hmm.. still failing and locally for me as well. I get 
[data_out_base_dx.txt](https://github.com/dealii/dealii/files/3508034/data_out_base_dx.txt)

drwells:
Hm. I don't understand DX output so I am going to guess that this is somehow not reproducible between different machines.

drwells:
AFAICT OpenDX is not as popular as it used to be (it looks like the last substantial commit on sourceforge was ten years ago). Would anyone mind if I solved this problem by deprecating our OpenDX code and then deleting the test?

bangerth:
That seems reasonable though drastic. It also doesn't quite fix the issue here unless you decide that we don't need tests for deprecated features any more.

drwells:
I am still not sure why the binary output is suddenly machine-dependent when translated to base64. There should be a way to fix this.

drwells:
@bangerth 
> It also doesn't quite fix the issue here unless you decide that we don't need tests for deprecated features any more.

Ordinarily I support keeping tests for deprecated features, but I very strongly doubt that anyone has used DX for deal.II in the last five years, so I don't mind removing them in this particular case. I would still prefer to fix the test!

drwells:
I fixed the issue: it turns out that DX has a line in the output for the time and date, which (duh) produces different results when converted to base64. I amended the test to delete that line so we should now be set.

PR #8547, opened 2019-08-11
Fix bug in MatrixFree ghost exchange with faces and FE_Q

kronbichler:
This actually fixes two independent bugs:
- `MatrixFree::loop()` forgot to enforce that some face integrals scheduled for elements in contact with other MPI ranks would happen before calling `LA::d::Vector::compress_start()` (overlapped with the remaining computations). This would only affect `FE_Q` type elements with unknowns on locally owned cells that are ghosted and in case the face loop information is built.
- The more tight MPI partitioners of `MatrixFree` used e.g. for `MatrixFree::cell_loop` that get used alongside the actual vector partitioner for cheaper sends forgot to collect ghosted dofs on the first locally owned active cell.

@tjhei FYI


masterleinad:
We need to rebase on top of #8550  for the indentation script to succeed.

kronbichler:
Thanks @masterleinad.

kronbichler:
I ran the testsuite locally and all matrix-free related tests are passing.

PR #8555, opened 2019-08-12
New test: Chains of constraints on parallel hp-adaptive applications.

marcfehling:
During the last deal.II workshop, @tjhei has brought to my attention that, in case of parallel hp-adaptivity, there may be cases of chains of constraints that reach over multiple domains via constrained DoFs on cell faces. This may become an issue: If one of the locally owned DoFs is constrained to a different one which is not a locally relevant one (i.e. neither locally owned nor ghost).

I tried to come up with a scenario that may be prone to this issue, but can not produce a failing test. I supplied a T-shaped domain with a `Q2` element on the center cell, and `Q1` on the others. The `AffineConstraints` matrix is built with hanging node constraints and no flux constraints, if desired. But in all of these cases, I do not get any inconsistencies in the `AffineConstraints` object at all.

This may be a good sign -- but I assume that we haven't found that specific corner case yet.

@tjhei -- Do you have a particular scenario in mind that will surely fail? Feel free to use this PR as a sandbox...

@bangerth -- FYI


tjhei:
Can you talk with @bangerth about it? He knows more about the hp paper and how constraints might form chains.

bangerth:
@tjhei -- the problem is that I've never come up with a case where these constraints have chains that extend beyond one neighboring cell. You've mentioned many times to me that you think that that case can happen, and so I had always believed that you had a case in mind -- I've tried many times to think of one, but always took your word for a fact that these cases exist :-)

So if you can't come up with a case either, then let's at least for the moment ignore the issue. If such cases exist, we will eventually learn of one!

tjhei:
> the problem is that I've never come up with a case where these constraints have chains that extend beyond one neighboring cell. You've mentioned many times to me that you think that that case can happen, and so I had always believed that you had a case in mind -- I've tried many times to think of one, but always took your word for a fact that these cases exist :-)

Haha, I always thought you had a case in mind. :-)
I am basing it on this paragraph in the p:d paper:
```
Dealing with Chains of Constraints. These considerations are of only theoretical in-
terest if constraints can be against degrees of freedom that are themselves constrained,
that is, if constraints form chains. This frequently happens in at least two situations.
First, it is common in hp-adaptive methods; (see that is, Bangerth and Kayser-Herold
[2009]. In that case, it is even conceivable that chains of constraints extend to the
boundary between ghost and artificial cells.
```

bangerth:
Yes, there can be chains, but only around an edge in 3d, for example. Think about an edge with N neighbor cells on which you have FE_Q(1)...FE_Q(N). The FE_Q(N) is constrained against the FE_Q(N-1), which is constrained against the FE_Q(N-2), etc., all the way down to the FE_Q(1) if they are arranged in the right order. But because they're all connected along one edge, they're all at worst ghost cells to each other. So yes there are chains, but I've never come across a case where these constraints extend past the ghost region.

(That is not quite true -- I've written a paper on the DSSY element, for which hanging node constraints can reach arbitrarily far. This would indeed be a problem. But nobody uses the DSSY element, and if someone did, we would have to address this somehow. I don't consider that a crittical issue at this moment.)

PR #8556, opened 2019-08-12
Fix a funny comment.

bangerth:
@tamiko pointed this out. The text was introduced in 6a7840fd733. I am glad that I did not author that commit :-)

### DIFF
--- a/source/grid/grid_reordering.cc
+++ b/source/grid/grid_reordering.cc
@@ -1166,12 +1166,9 @@ GridReordering<2>::invert_all_cells_of_negative_grid(
           ++n_negative_cells;
           std::swap(cell.vertices[1], cell.vertices[3]);
 
-          // check whether the
-          // resulting cell is now ok.
-          // if not, then the grid is
-          // seriously broken and
-          // should be sticked into the
-          // bin
+          // Check whether the resulting cell is now ok.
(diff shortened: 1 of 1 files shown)


bangerth:
Or was it @drwells who pointed it out? I don't remember -- @drwells mentioned in on slack...

drwells:
I thought the consensus here was to write
```cpp
          for (unsigned int i = 0; i < GeometryInfo<3>::vertices_per_cell; ++i)
            vertices_lex[GeometryInfo<3>::ucd_to_deal[i]] = cell.vertices[i];
          AssertThrow(GridTools::cell_measure<3>(all_vertices, vertices_lex) >
                        0,
                        0,
                      ExcMessage("the grid is seriously broken and should be sticked into the bin"));
```
:)

bangerth:
@drwells: While funny, in practice no user will actually get the joke because they don't know the backstory. It is sad in some sense, but I think we ought to just fix all of these grammar problems -- though I suspect that the smiley at the end of your last comment implies that you agree :-)

For a similar case that's just plain aggravating to everyone else who is not inordinately proud of their own, 20 year old joke, see here: https://lwn.net/Articles/753646/

drwells:
Yes; to be clear, I do not advocate actually changing the error message to say that!

PR #8557, opened 2019-08-12
Update the description in step-8 that uses the new pictures.

bangerth:
Updates some text to reference the new versions of the pictures I just added.


masterleinad:
Hmm... looking at https://www.dealii.org/8.5.0/doxygen/deal.II/step_8.html we now have the vector field twice in the old documentation. Would it make sense to use different file names?

bangerth:
Ah, yes, I copied the new picture to the wrong filename. I've restored the old mesh and copied it to the right file. The patch is also updated.

masterleinad:
Thanks! That looks reasonable now.

PR #8560, opened 2019-08-13
Fix an assertion that was missing a semicolon.

bangerth:


masterleinad:
We need to rebuild on top of #8561.

bangerth:
Rebased.

PR #8564, opened 2019-08-13
Change GridOut::write_svg() defaults

bangerth:
The GridOut::write_svg() function now no longer outputs the
cell level and number by default; this also obviates the need for the
colorbar and the caption. This is achieved by changing the defaults in
the constructor of the GridOutFlags::Svg structure. This likely comes
closer to what most users want to see from this function -- namely,
just the mesh for inclusion into text.

It might be easiest to read the four commits individually.

### DIFF
--- a/include/deal.II/grid/grid_out.h
+++ b/include/deal.II/grid/grid_out.h
@@ -672,42 +672,67 @@ namespace GridOutFlags
    */
   struct Svg
   {
-    /// Height of the plot in SVG units, computed from width if zero. Defaults
-    /// to 1000
+    /**
+     *  Height of the plot in SVG units, computed from width if zero. Defaults
+     *  to 1000.
+     */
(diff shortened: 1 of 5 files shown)


bangerth:
Ping?

include/deal.II/grid/grid_out.h
```diff
@@ -669,41 +669,65 @@ namespace GridOutFlags
...
+     * The style used for the background of the mesh.
+     */
     Background background;
 
     // View angles for the perspective view of the grid; Default is 0, 0 (top
     // view).
     /**
      * The azimuth angle measured from ??? in degrees. Default is 0.
      */
+
```
tjhei: remove?

bangerth:
I've addressed @tjhei's one comment.

PR #8575, opened 2019-08-15
Extend read_vtk to add FIELD data

OpenFCST:
This PR extends the functionality of read_vtk to include FIELD data which can be used to add further information to each cell beyond material ID. In OpenFCST this is used for example to store pore size in each cell. To access the field data, please use the function get_field_data which returns a map with a key (in the form of a string) and then a map with cellID and value.


include/deal.II/grid/grid_in.h
```diff
@@ -663,8 +663,25 @@ class GridIn
 
 
   DeclException0(ExcGmshNoCellInformation);
+  
+  /**
+   * 
+   */
+  std::map< std::string, std::map<int,double> > get_field_data()
```
bangerth: can you add documentation here?

include/deal.II/grid/grid_in.h
```diff
@@ -663,8 +663,25 @@ class GridIn
...
+  {
+
```
bangerth:
```suggestion
```

include/deal.II/grid/grid_in.h
```diff
@@ -663,8 +663,25 @@ class GridIn
...
   DeclException0(ExcGmshNoCellInformation);
+  
+  /**
+   * 
+   */
+  std::map< std::string, std::map<int,double> > get_field_data()
+  {
+      
+      return this->field_data;
+  }
```
bangerth: Is there a good reason that this function returns a non-`const` reference object? If it returned a `const` reference, you could also make the function itself `const`.

include/deal.II/grid/grid_in.h
```diff
@@ -663,8 +663,25 @@ class GridIn
...
 
 protected:
+    
+  /**
+  * Data member that stores cell data. The format is as follows:
+  * std::string stores identified
```
bangerth: I don't understand this half sentence. Can you expand on it?

doxygen will not render the line breaks like you have it here. If you want this and the following line to be a bullet point list, start the line with " - ".

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
-      while (in >> keyword)
-        if (keyword == "CELL_DATA")
+      /////////////////////Processing the CELL_DATA
+      /// section////////////////////////
+      in >> keyword;
+
+      AssertThrow(
+        keyword == "CELL_DATA",
+        ExcMessage(std::string(
+          "While reading VTK file, missing CELL_DATA section. Found <" +
```
bangerth: This implies that *every* file you want to read in now has to have a `CELL_DATA` section where before it was optional. Can you explain why this is now necessary?

bangerth: I've tried to read the part below that corresponds to the existing code. It's a bit complicated to read because the indentation changed, but I don't actually see that you changed anything in the part that reads the cell data -- or did you?

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+            in >> keyword;
+        }
+      }
+      
+      //////////////////
+      //// Addition of FIELD DATA:
+      //////////////////      
+      if(keyword == "FIELD")
+      {
+        std::cout<<std::endl<<"Processing FIELD_DATA...."<<std::endl;
```
bangerth: We try not to output any debug output to `std::cout`.

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+        unsigned int no_fields;
+        /*char word[10];                                     //FieldData in the Field Section
```
bangerth: Just remove the commented out part if you don't need it.

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+        }
+      }
+      
+      //////////////////
+      //// Addition of FIELD DATA:
+      //////////////////      
+      if(keyword == "FIELD")
+      {
+        std::cout<<std::endl<<"Processing FIELD_DATA...."<<std::endl;
+        unsigned int no_fields;
```
bangerth: Our usual coding style is to name these variables `n_fields`, to ensure that it's not misread as "[There are] no fields", which would then have to be a `bool` variable.

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+        /*char word[10];                                     //FieldData in the Field Section
+        in>>word;
+        */
+       in>>keyword;
```
bangerth: I don't think this matches our usual indentation styles (there ought to be spaces around `>>`). Can you run the `./contrib/utilities/indent` script (or just say `make indent`)?

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+        AssertThrow(
+          keyword == "FieldData",
+          ExcMessage("While reading VTK file, missing keyword FieldData"));
+        
+        //if (strcmp(word,"FieldData")==0)
```
bangerth: same here -- just remove the code that is no longer needed

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+          in>>no_fields;
+        
+        for(unsigned int i = 0;i < no_fields;i++)
```
bangerth:
```suggestion
        for(unsigned int i = 0; i < no_fields; ++i)
```

source/grid/grid_in.cc
```diff
@@ -340,139 +340,181 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+          for (unsigned int j=0;j<n_ids;j++)
+          {
+            in>>data;
+            if (j<cells.size())
+              temp_data[j]=data;
           }
+          this->field_data[section_name]=temp_data;
+        }
+      }
+      std::cout<<std::endl<<"Processing FIELD section complete."<<std::endl;
```
bangerth:
```suggestion
```

bangerth:
Let us know once you want us to take another look at this!

OpenFCST:
A test has been added and the requested changes have been made.

Regarding the indentation, we replaced a while (keyword<<in) with keyword<<in, thereby changing the indentation. We could not get the code to work with while (keyword<<in)  and, honestly, we do not really understand its purpose very well. Let us know if you think it is necessary.

masterleinad:
@bangerth I believe all the changes you were requesting are made.

/rebuild

source/grid/grid_in.cc
```diff
@@ -340,139 +340,172 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+                        Utilities::int_to_string(cells.size()) + ") plus the number of quads (" +
+                        Utilities::int_to_string(subcelldata.boundary_quads.size()) +
+                        " in 3d or the number of lines (" + Utilities::int_to_string(
+                          subcelldata.boundary_lines.size()) + ") in 2d."));
+          in>>data_type;
+          std::map<int,double> temp_data;
+          for (unsigned int j=0;j<n_ids;j++)
+          {
+            in>>data;
+            if (j<cells.size())
```
bangerth: You are making an implicit assumption here that the `FIELD` you are reading is tied to cells, and you are truncating everything that extends beyond the number of cells. But how do you know that this is true? What if someone has field data that consists of something else? I don't think that's generally true, and if it were cell-associated data, why is it not in a `CELL_DATA` section like, for example, material and manifold ids? At least I can't infer any cell relationship from the documentation at https://lorensen.github.io/VTKExamples/site/VTKFileFormats/#legacy-file-examples

source/grid/grid_in.cc
```diff
@@ -340,139 +340,172 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+              temp_data[j]=data;
```
bangerth: This is interesting, because you're looping over `j` and then you create a map from `j` to `data`. But `j` consists of indices 0, 1, 2, ... so that map is really quite simple: It consists of data `(0,a0), (1,a1), (2,a2), ...`. This can be stored in an easier way: You don't need the expensive map, you could just store it as a vector `(a0,a1,a2,...)`.

include/deal.II/grid/grid_in.h
```diff
@@ -663,8 +663,26 @@ class GridIn
...
+  }
 
 protected:
+    
+  /**
+  * Data member that stores cell data. The format is as follows:
+  * - std::string stores the name of the field data (identifier) as specified in the external mesh
+  * - std::map<int, double> stores cell_id and value for the given identifier
+  */
+  std::map< std::string, std::map<int,double> > field_data;
```
bangerth: This is the place where you store the map that I'm commenting on below. I think it would be easier to write this as `std::map< std::string, std::vector<double> >`.

include/deal.II/grid/grid_in.h
bangerth: This is still an open comment. What you're doing here with this function is return a *copy* of the object. That's an expensive operation. An easier approach would be to return a `const` reference (i.e., in essence a pointer to the object you're storing).

bangerth:
@OpenFCST -- any progress?

OpenFCST:
I believe we have address all of the comments in the new update. Please let us know if there are further changes needed.

Regarding the comment of why we use FIELD to include the data, the main reason is because the libraries we use to generate the VTK file, i.e. tvtk.UnstructuredGrid(points=points) in python, puts any additional cell_data in field data. Further, the documentation specifies "Field data is a general format without topological and geometric structure, and without a particular dimensionality. Typically field data is associated with the points or cells of a dataset." I agree that FIELD could include many things and at some point it could be generalized, but in our case it is always cell_data. We have an assertion to make sure people are aware that a field with cell data size is used.

bangerth:
We apparently stepped on the changes here in the meantime and there is now a merge conflict. Would you mind rebasing on the current master?

include/deal.II/grid/grid_in.h
```diff
@@ -663,8 +663,26 @@ class GridIn
 
 
   DeclException0(ExcGmshNoCellInformation);
+  
+  /**
+   * Return a map containing field data. The format of the returned map is as follows: 
+   * - std::string stores the name of the field data (identifier) as specified in the external mesh
+   * - std::vector<double > stores value for the given identifier in each cell. To access the value use field_data[name_field][cell_id].
```
bangerth, line 670:
```suggestion
   * - std::vector<double> stores value for the given identifier in each cell. To access the value use field_data[name_field][cell_id].
```

include/deal.II/grid/grid_in.h
bangerth: This is still an open comment.

source/grid/grid_in.cc
```diff
@@ -181,15 +181,15 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
   unsigned int n_ints;
 
   if (keyword == "CELLS")
-    {
-      in >> n_geometric_objects;
-      in >> n_ints; // Ignore this, since we don't need it.
+  {
+    in >> n_geometric_objects;
+    in >> n_ints; // Ignore this, since we don't need it.
```
bangerth, line 186: There are a lot of whitespace changes in this patch that make it really difficult to review. Please run `make indent` in your build directory to normalize whitespace. This is going to make it substantially easier to review what *really* is different.

include/deal.II/grid/grid_in.h
```diff
@@ -663,8 +663,26 @@ class GridIn
...
+  }
 
 protected:
+    
+  /**
+  * Data member that stores cell data. The format is as follows:
+  * - std::string stores the name of the field data (identifier) as specified in the external mesh
+  * - std::vector<double> stores value for the given identifier in each cell_id. To access the value use field_data[name_field][cell_id].
+  */
+  std::map< std::string, std::vector<double> > field_data;
```
bangerth, line 684: This field is `protected:`. But I think it could just as well be `private:`, right? If so, please make it `private:`.

source/grid/grid_in.cc
```diff
@@ -336,143 +336,183 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
-                      for (const auto &set_cmp : data_sets)
-                        if (keyword == set_cmp)
-                          {
-                            set = keyword;
-                            break;
-                          }
-                      if (set.empty())
-                        // keep ignoring everything until the next SCALARS
+        /////////////////////Processing the CELL_DATA and FIELD_DATA
+        /// section////////////////////////
```
bangerth, line 344: This comment looks a bit funnily formatted. Please break it into something more easily readable.

source/grid/grid_in.cc
```diff
@@ -336,143 +336,183 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+                                    "that lists a total of " + Utilities::int_to_string(n_ids) +
+                                    " cell data objects, but this needs to equal the number of cells (which is " +
+                                    Utilities::int_to_string(cells.size()) + ") plus the number of quads (" +
+                                    Utilities::int_to_string(subcelldata.boundary_quads.size()) +
+                                    " in 3d or the number of lines (" + Utilities::int_to_string(
+                                    subcelldata.boundary_lines.size()) + ") in 2d."));
+                    in>>data_type;
+                    std::vector<double> temp_data;
+                    temp_data.resize(n_ids);
+                    for (unsigned int j=0;j<n_ids;j++)
```
bangerth, line 500:
```suggestion
                    for (unsigned int j=0;j<n_ids;++j)
```

source/grid/grid_in.cc
```diff
@@ -336,143 +336,183 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+                    {
+                        in>>data;
+                        if (j<cells.size())
+                            temp_data[j]=data;
                     }
-              }
-          }
-
+                    this->field_data[section_name]=temp_data;
```
bangerth, line 506: You can make this a bit more efficient if you write
```suggestion
                    this->field_data[section_name]=std::move(temp_data);
```

masterleinad:
Closed in favor of #10649.

PR #8585, opened 2019-08-16
added #ifndef DOXYGEN to forward declarations

rezarastak:
Forward declarations cause doxygen to report an incorrect location for the definition of a class/struct.   For example, at the end of the documentation of `SymmetricTensor` it says
> Definition at line 93 of file mpi.h.

instead of
> Definition at line 555 of file symmetric_tensor.h.

This PR attempts to fix it. My first commit is a proof of concept that fixes this issue for the documentation of `Tensor` and `SymmetricTensor`. If you agree with these changes, I will add apply the same modifications to all forward declarations of all header files.

In my observation, the issue is seen only for forward-declared classes/structs. doxygen does not have a problem with forward-declared functions.

### DIFF
--- a/doc/news/changes/minor/20190816RezaRastak
+++ b/doc/news/changes/minor/20190816RezaRastak
@@ -1,0 +1,4 @@
+Fixed: Forward declarations in header files are prevented from confusing
+doxygen.
+<br>
+(Reza Rastak, 2019/08/16)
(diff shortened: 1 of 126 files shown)


masterleinad:
Sure, go ahead. https://github.com/dealii/dealii/pull/8566 was recently going in a similar direction.

bangerth:
Do you want us to merge this as-is and follow up with potential other places separately (which I'm ok with), or do you want to include it all in this PR by pushing more changes?

rezarastak:
I am going to push more changes to the this PR.

masterleinad:
I have no idea why this is no tested. Let's try again.
/rebuild

PR #8567, opened 2019-08-14
Modify the definition of CellId

peterrum:
This PR is part of the effort to introduce the new `parallel::fullydistributed::Triangulation` (see #8558).

In this PR, the definition of a `CellId` is modified in the sense that it depends now on the coarse-cell id, in contrast to the coarse-cell (level) index. Both have been the same till now, however, cannot be the same in the fully distributed case.

I have introduced in the `Triangulation` class two new methods to translate the ids and the indices:
```cpp
  virtual unsigned int
  translate_coarse_cell_id_to_coarse_cell_index(
    const unsigned int coarse_cell_id) const;

  virtual unsigned int
  translate_coarse_cell_index_to_coarse_cell_id(
    const unsigned int coarse_cell_index) const;
```
These methods are at the moment very boring: they simply return the input value... However, will be overridden in coming PRs. In the case of `p:d:t`, they will access `coarse_cell_to_p4est_tree_permutation` and `p4est_tree_to_coarse_cell_permutation`. In the case of `p:f:t`, id will become global id and index a local id.

Issue #8558, opened 2019-08-13
Introduction of parallel::fullydistributed::Triangulation


peterrum:
At the 7th deal.II users and developers workshop, we have agreed on merging the new `parallel::fullydistributed::Triangulation` (short `p:f:t`, see also PR #8418) in at least 5 stages. The name `parallel::fullydistributed::Triangulation` has been accepted. The stages are:

**Step 1:** modify class hierarchy
- [x] rename `parallel::Triangulation` to `parallelTriangulationBase`
- [x]  introduce a new `parallel::DistributedTrinagulationBase`, which will be the base class for `p:d:t` and `p:f:t`. This way we can handle `p:d:t` and `p:f:t` at many places in the library the same way.

**Step 2:** generalization
- [x]  redefinition of `CellId`: `coarse_cell_id`s should define uniquely coarse cells (even in the fully distributed case)
- [x] let the user access `CellId`'s  `coarse_cell_id` directly
- [x]  rewrite the `DoFHandlerPolicy` of `p:d:t` such that it takes a reference to `parallel::DistributedTrinagulationBase`, only uses the redefined `CellID`s and does not use the p4est-mappings
- [x] Enable `DoFHandlerPolicy` of `p:d:t` for 1D
- [x] Implement `DoFAccessor::set/get_mg_dof_indices` for 1D

**Step 3:**
- [x]  introduce the new `parallel::fullydistributed::Triangulation` with a single method to create the `ConstructionData`

**Step 4:**
- [x]  introduce (de)serialization capabilities
- [x] perform partitioning by a subgroup of processes  

**Step 5:**
- [ ]  enable global coarsening and refinement

For the future we leave:
* introduce advanced multilevel portioning algorithms in  `parallel::fullydistributed::util`
* re-partitioning algorithms (in the form of an oracle)

*ping:* @bangerth, @tjhei, @tamiko, @luca-heltai, @kronbichler

peterrum:
* In what way is #8535 related to the issue?
* How do we want to call the `ConstructionData`. We discussed the term `MeshPartitionerData`... I prefer the first term. Which term *loves* @bangerth *and* @tjhei?

peterrum:
@bangerth @tjhei  I think the 6 PRs #8567, #8586, #8579, #8587, #8592, and #8593 contain all necessary generalizations of deal.II that are required by `p:f:t` to be able to run simulations on fully distributed triangulations (also for 1D with MG). Let's get these merged step by step!

[...]

PR #8586, opened 2019-08-16
Return coarse-cell id by CellId

peterrum:
This PR is part of the effort to introduce the new parallel::fullydistributed::Triangulation (see #8558) and is a follow-up to PR #8567.

Ping: @bangerth @tjhei

### DIFF
--- a/include/deal.II/base/types.h
+++ b/include/deal.II/base/types.h
@@ -95,6 +95,18 @@ namespace types
 #  define DEAL_II_DOF_INDEX_MPI_TYPE MPI_UNSIGNED
 #endif
 
+#ifdef DEAL_II_WITH_64BIT_INDICES
+  /**
+   * The type used for coarse-cell ids.
+   */
+  using coarse_cell_id = unsigned long long int;
+#else
+  /**
+   * The type used for coarse-cell ids.
+   */
+  using coarse_cell_id = unsigned int;
+#endif
+
   /**
    * The type used to denote boundary indicators associated with every piece
    * of the boundary and, in the case of meshes that describe manifolds in
@@ -187,6 +199,12 @@ namespace numbers
    */
   const types::global_dof_index invalid_dof_index =
     static_cast<types::global_dof_index>(-1);
+
+  /**
+   * An invalid value for coarse-cell ids.
+   */
+  const types::coarse_cell_id invalid_coarse_cell_id =
+    static_cast<types::coarse_cell_id>(-1);
 
   /**
    * Invalid material_id which we need in several places as a default value.
--- a/include/deal.II/grid/cell_id.h
+++ b/include/deal.II/grid/cell_id.h
@@ -85,7 +85,7 @@ public:
    * and the number of children of a cell in the current space dimension (i.e.,
    * GeometryInfo<dim>::max_children_per_cell).
    */
-  CellId(const unsigned int               coarse_cell_id,
+  CellId(const types::coarse_cell_id      coarse_cell_id,
          const std::vector<std::uint8_t> &child_indices);
 
   /**
@@ -99,9 +99,9 @@ public:
    * GeometryInfo<dim>::max_children_per_cell). The array
    * @p child_indices must have at least @p n_child_indices valid entries.
    */
-  CellId(const unsigned int  coarse_cell_id,
-         const unsigned int  n_child_indices,
-         const std::uint8_t *child_indices);
+  CellId(const types::coarse_cell_id coarse_cell_id,
+         const unsigned int          n_child_indices,
+         const std::uint8_t *        child_indices);
 
   /**
    * Construct a CellId object with a given binary representation that was
@@ -161,12 +161,18 @@ public:
   void
   serialize(Archive &ar, const unsigned int version);
 
+  /**
+   * Return the id of the coarse cell.
+   */
+  types::coarse_cell_id
+  get_coarse_cell_id() const;
+
 private:
   /**
    * The number of the coarse cell within whose tree the cell
    * represented by the current object is located.
    */
-  unsigned int coarse_cell_id;
+  types::coarse_cell_id coarse_cell_id;
 
   /**
    * The number of child indices stored in the child_indices array. This is
@@ -307,6 +313,14 @@ CellId::operator<(const CellId &other) const
     return false;
   return true; // other.id is longer
 }
+
+
+inline types::coarse_cell_id
+CellId::get_coarse_cell_id() const
+{
+  return coarse_cell_id;
+}
+
 
 DEAL_II_NAMESPACE_CLOSE
 
--- a/include/deal.II/grid/tria.h
+++ b/include/deal.II/grid/tria.h
@@ -3483,6 +3483,34 @@ public:
     std::pair<std::pair<cell_iterator, unsigned int>, std::bitset<3>>> &
   get_periodic_face_map() const;
 
+  /**
+   * Translate the unique id of a coarse cell to its index.
+   *
+   * @note For serial and shared triangulation both id and index are the same.
+   *       For distributed triangulations setting both might differ, since the
+   *       id might correspond to a global id and the index to a local id.
+   *
+   * @param coarse_cell_id Unique id of the coarse cell.
+   * @return Index of the coarse cell within the current triangulation.
+   */
+  virtual unsigned int
+  coarse_cell_id_to_coarse_cell_index(
+    const types::coarse_cell_id coarse_cell_id) const;
+
+
+  /**
+   * Translate the index of coarse cell to its unique id.
+   *
+   * @note: See the note of the method
+   * translate_coarse_cell_id_to_coarse_cell_index().
+   *
+   * @param coarse_cell_index Index of the coarse cell.
+   * @return Id of the coarse cell.
+   */
+  virtual types::coarse_cell_id
+  coarse_cell_index_to_coarse_cell_id(
+    const unsigned int coarse_cell_index) const;
+
 
   BOOST_SERIALIZATION_SPLIT_MEMBER()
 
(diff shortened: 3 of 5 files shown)


doc/news/changes/minor/20190813PeterMunch
```diff
@@ -0,0 +1,4 @@
+New: The definition of CellID has been modified so that it depends on the 
+unique coarse-cell id and anymore on the index.
```
tjhei:
```suggestion
unique coarse-cell id and no longer on the index.
```

include/deal.II/grid/cell_id.h
```diff
@@ -158,6 +158,12 @@ class CellId
   void
   serialize(Archive &ar, const unsigned int version);
 
+  /**
+   * Return the id of the coarse cell.
+   */
+  unsigned int
+  get_coarse_cell_id() const;
```
tjhei: Are we okay of this being an int or do we ever expect to have more than 4b cells?

source/grid/tria.cc
```diff
@@ -13296,6 +13296,26 @@ Triangulation<dim, spacedim>::get_periodic_face_map() const
 
 
 
+template <int dim, int spacedim>
+unsigned int
+Triangulation<dim, spacedim>::translate_coarse_cell_id_to_coarse_cell_index(
+  const unsigned int coarse_cell_id) const
+{
+  return coarse_cell_id;
```
tjhei: Would these be good candidates to inline instead?

include/deal.II/grid/tria.h
```diff
@@ -3482,6 +3482,34 @@ class Triangulation : public Subscriptor
...
+   *  @note: For serial and shared triangulation both id and index are the same.
+   *         For distributed triangulations setting both might differ, since the
+   *         id might correspond to a global id and the index a local id.
+   *
+   * @param coarse_cell_id unique id of the coarse cell
+   * @return index of the coarse cell
+   */
+  virtual unsigned int
+  translate_coarse_cell_id_to_coarse_cell_index(
+    const unsigned int coarse_cell_id) const;
```
tjhei: I would remove "translate_" from these function names.

include/deal.II/grid/cell_id.h
kronbichler: I don't think you want to read a 4b cell mesh from disk (or any parallel file system) now or in the next ten years. @peterrum would there be any internal inefficiency to make it 64 bits in any case?

peterrum: So: I modified the type of `cell_id` to `unsigned long int` (and I have introduced `numbers::invalid_unsigned_long_int`). Is that fine or should I use some deal.II alias?

It is nice that you are exited about `p:f:t` (which hopefully can handle so big meshes) and you are thinking about the long term future (Agenda 2030), but my short term goal is that @tamiko is happy and he is happy if he can run simulations on meshes with more than 10 millions coarse cells efficiently ;)

Note: in an upcoming PR we have to update the binary representation of `CellId` such that it can handle `unsigned long int` (and it is easily modifiable with simple C-shifts and -logical operations!!!).

peterrum:
I have made the changes. Once PR #8567 is merged, I will update this branch and then this PR is ready!

include/deal.II/grid/cell_id.h
peterrum: Furthermore, we might need to use other data structures in the DoFHandlerPolicy. Currently it relies on the quadrant the p4est quadrants...

include/deal.II/grid/cell_id.h
```diff
@@ -82,7 +82,7 @@ class CellId
    * and the number of children of a cell in the current space dimension (i.e.,
    * GeometryInfo<dim>::max_children_per_cell).
    */
-  CellId(const unsigned int               coarse_cell_id,
+  CellId(const unsigned long int          coarse_cell_id,
```
bangerth: Rather than picking a type here, let's give it a name in `types.h`: `types::coarse_cell_id`. This way, the type is also self-documenting, and easy to change. Indeed, I think you should make the type 32bit by default, and only 64bit if we have `DEAL_II_WITH_64BIT_INDICES`: It doesn't make sense to have more than 4B coarse mesh cells if you don't also allow more than 4B DoFs.

include/deal.II/base/types.h
```diff
@@ -173,6 +173,14 @@ namespace numbers
   static const unsigned int invalid_unsigned_int =
     static_cast<unsigned int>(-1);
 
+  /**
+   * Representation of the largest number that can be put into an unsigned
+   * long integer.
+   */
+  static const unsigned long int invalid_unsigned_long_int =
+    static_cast<unsigned long int>(-1);
```
bangerth: Then you can name this here `invalid_coarse_cell_id` and define it in the same place as `invalid_subdomain_id` and the other `invalid_...` numbers.

include/deal.II/grid/tria.h
```diff
@@ -3482,6 +3482,34 @@ class Triangulation : public Subscriptor
...
+   * @note: For serial and shared triangulation both id and index are the same.
+   *        For distributed triangulations setting both might differ, since the
+   *        id might correspond to a global id and the index a local id.
+   *
+   * @param coarse_cell_id Unique id of the coarse cell.
+   * @return Index of the coarse cell.
+   */
+  virtual unsigned int
+  coarse_cell_id_to_coarse_cell_index(
+    const unsigned long int coarse_cell_id) const;
```
bangerth: Only one of these two would then be `types::coarse_cell_id`, and similarly for the function below.

peterrum: Thanks @bangerth! This is what I was asking above. I have modified the code accordingly. I still have to add some comments.

include/deal.II/grid/tria.h
```diff
@@ -4281,6 +4309,30 @@ Triangulation<dim, spacedim>::load(Archive &ar, const unsigned int)
 }
 
 
+<<<<<<< HEAD
+=======
```
bangerth: merge mistake :-)

include/deal.II/grid/tria.h
```diff
@@ -4281,6 +4309,30 @@ Triangulation<dim, spacedim>::load(Archive &ar, const unsigned int)
...
+inline types::coarse_cell_id
+Triangulation<dim, spacedim>::coarse_cell_index_to_coarse_cell_id(
+  const unsigned int coarse_cell_index) const
+{
+  return coarse_cell_index;
+}
+
+
+
+>>>>>>> c1ae70e... Add get_coarse_cell_id to CellID
```
bangerth: and here

include/deal.II/base/types.h
```diff
@@ -95,6 +95,12 @@ namespace types
 #  define DEAL_II_DOF_INDEX_MPI_TYPE MPI_UNSIGNED
 #endif
 
+#ifdef DEAL_II_WITH_64BIT_INDICES
+  using coarse_cell_id = unsigned long long int;
+#else
+  using coarse_cell_id = unsigned int;
```
bangerth: This looks good if you add some documentation.

include/deal.II/grid/tria.h
```diff
@@ -3482,6 +3482,34 @@ class Triangulation : public Subscriptor
...
+   * @note: For serial and shared triangulation both id and index are the same.
+   *        For distributed triangulations setting both might differ, since the
+   *        id might correspond to a global id and the index a local id.
+   *
+   * @param coarse_cell_id Unique id of the coarse cell.
+   * @return Index of the coarse cell.
+   */
+  virtual unsigned int
+  coarse_cell_id_to_coarse_cell_index(
+    const types::coarse_cell_id coarse_cell_id) const;
```
bangerth: This is where you had renamed things :-)

peterrum: Thanks @bangerth! I have updated the PR.

source/grid/cell_id.cc
```diff
@@ -168,6 +168,14 @@ CellId::to_cell(const Triangulation<dim, spacedim> &tria) const
   return cell;
 }
 
+
+
+types::coarse_cell_id
```
tjhei: I think this now needs to be removed as you inlined it.

include/deal.II/grid/tria.h
```diff
@@ -3483,6 +3483,34 @@ class Triangulation : public Subscriptor
     std::pair<std::pair<cell_iterator, unsigned int>, std::bitset<3>>> &
   get_periodic_face_map() const;
 
+  /**
+   * Translate the unique id of a coarse cell to its index.
+   *
+   * @note: For serial and shared triangulation both id and index are the same.
```
bangerth: doxygen will render this funny with the colon:
```suggestion
   * @note For serial and shared triangulation both id and index are the same.
```

include/deal.II/grid/tria.h
```diff
@@ -3483,6 +3483,34 @@ class Triangulation : public Subscriptor
...
+   *        For distributed triangulations setting both might differ, since the
+   *        id might correspond to a global id and the index to a local id.
+   *
+   * @param coarse_cell_id Unique id of the coarse cell.
+   * @return Index of the coarse cell.
```
bangerth:
```suggestion
   * @return Index of the coarse cell within the current triangulation.
```

peterrum:
One test did timeout. How is that possible?

tjhei:
> One test did timeout. How is that possible?

bad luck? :-) I am rerunning it hoping that will fix it...

source/grid/cell_id.cc
peterrum: I have inlined it.

include/deal.II/grid/tria.h
peterrum: OK!

include/deal.II/grid/tria.h
peterrum: OK!

bangerth:
OK once the testers are through and happy.

PR #8588, opened 2019-08-16
Change Polynomial Classes to Derive from TensorPolynomialsBase

GrahamBenHarper:
This is step 2 in a multi-step process for addressing #1973.

This changes 6 polynomial classes to derive from the `TensorPolynomialsBase` class that was implemented in #8528.

The next thing I need to do (either as part of this PR or the next PR) is remove some function implementations from these derived classes so that they use the base class's functions and variables.

### DIFF
--- a/include/deal.II/base/polynomials_abf.h
+++ b/include/deal.II/base/polynomials_abf.h
@@ -24,6 +24,7 @@
 #include <deal.II/base/polynomial.h>
 #include <deal.II/base/polynomial_space.h>
 #include <deal.II/base/tensor.h>
+#include <deal.II/base/tensor_polynomials_base.h>
 #include <deal.II/base/tensor_product_polynomials.h>
 #include <deal.II/base/thread_management.h>
 
@@ -51,7 +52,7 @@ DEAL_II_NAMESPACE_OPEN
  * @date 2006
  */
 template <int dim>
-class PolynomialsABF
+class PolynomialsABF : public TensorPolynomialsBase<dim>
 {
 public:
   /**
@@ -82,26 +83,13 @@ public:
           std::vector<Tensor<2, dim>> &grads,
           std::vector<Tensor<3, dim>> &grad_grads,
           std::vector<Tensor<4, dim>> &third_derivatives,
-          std::vector<Tensor<5, dim>> &fourth_derivatives) const;
-
-  /**
-   * Return the number of ABF polynomials.
-   */
-  unsigned int
-  n() const;
-
-  /**
-   * Return the degree of the ABF space, which is two less than the highest
-   * polynomial degree.
-   */
-  unsigned int
-  degree() const;
+          std::vector<Tensor<5, dim>> &fourth_derivatives) const override;
 
   /**
    * Return the name of the space, which is <tt>ABF</tt>.
    */
   std::string
-  name() const;
+  name() const override;
 
   /**
    * Return the number of polynomials in the space <tt>RT(degree)</tt> without
@@ -111,23 +99,19 @@ public:
   static unsigned int
   compute_n_pols(unsigned int degree);
 
-private:
   /**
-   * The degree of this object as given to the constructor.
+   * @copydoc TensorPolynomialsBase<dim>::clone()
    */
-  const unsigned int my_degree;
+  virtual std::unique_ptr<TensorPolynomialsBase<dim>>
+  clone() const override;
 
+private:
   /**
    * An object representing the polynomial space for a single component. We
    * can re-use it for the other vector components by rotating the
    * coordinates of the evaluation point.
    */
   const AnisotropicPolynomials<dim> polynomial_space;
-
-  /**
-   * Number of Raviart-Thomas polynomials.
-   */
-  unsigned int n_pols;
 
   /**
    * A mutex that guards the following scratch arrays.
@@ -159,22 +143,6 @@ private:
    */
   mutable std::vector<Tensor<4, dim>> p_fourth_derivatives;
 };
-
-
-template <int dim>
-inline unsigned int
-PolynomialsABF<dim>::n() const
-{
-  return n_pols;
-}
-
-
-template <int dim>
-inline unsigned int
-PolynomialsABF<dim>::degree() const
-{
-  return my_degree;
-}
 
 
 template <int dim>
--- a/include/deal.II/base/polynomials_bdm.h
+++ b/include/deal.II/base/polynomials_bdm.h
@@ -24,6 +24,7 @@
 #include <deal.II/base/polynomial.h>
 #include <deal.II/base/polynomial_space.h>
 #include <deal.II/base/tensor.h>
+#include <deal.II/base/tensor_polynomials_base.h>
 #include <deal.II/base/thread_management.h>
 
 #include <vector>
@@ -97,7 +98,7 @@ DEAL_II_NAMESPACE_OPEN
  * @date 2003, 2005, 2009
  */
 template <int dim>
-class PolynomialsBDM
+class PolynomialsBDM : public TensorPolynomialsBase<dim>
 {
 public:
   /**
@@ -128,13 +129,7 @@ public:
           std::vector<Tensor<2, dim>> &grads,
           std::vector<Tensor<3, dim>> &grad_grads,
           std::vector<Tensor<4, dim>> &third_derivatives,
-          std::vector<Tensor<5, dim>> &fourth_derivatives) const;
-
-  /**
-   * Return the number of BDM polynomials.
-   */
-  unsigned int
-  n() const;
+          std::vector<Tensor<5, dim>> &fourth_derivatives) const override;
 
   /**
    * Return the degree of the BDM space, which is one less than the highest
@@ -147,7 +142,7 @@ public:
    * Return the name of the space, which is <tt>BDM</tt>.
    */
   std::string
-  name() const;
+  name() const override;
 
   /**
    * Return the number of polynomials in the space <tt>BDM(degree)</tt>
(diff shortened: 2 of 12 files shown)


include/deal.II/base/polynomials_abf.h
```diff
@@ -111,6 +112,10 @@ class PolynomialsABF
   static unsigned int
   compute_n_pols(unsigned int degree);
 
+  // documentation inherited from base class
```
masterleinad:
```suggestion
  /**
   * @copydoc TensorPolynomialsBase<dim>::clone()
   */
```

include/deal.II/base/polynomials_bdm.h
```diff
@@ -157,6 +158,10 @@ class PolynomialsBDM
   static unsigned int
   compute_n_pols(unsigned int degree);
 
+  // documentation inherited from the base class
```
masterleinad:
```suggestion
  /**
   * @copydoc TensorPolynomialsBase<dim>::clone()
   */
```

include/deal.II/base/polynomials_bernardi_raugel.h
```diff
@@ -142,6 +143,10 @@ class PolynomialsBernardiRaugel
   static unsigned int
   compute_n_pols(const unsigned int k);
 
+  // documentation inherited from base class
```
masterleinad:
```suggestion
  /**
   * @copydoc TensorPolynomialsBase<dim>::clone()
   */
```

include/deal.II/base/polynomials_nedelec.h
```diff
@@ -108,6 +109,10 @@ class PolynomialsNedelec
   static unsigned int
   compute_n_pols(unsigned int degree);
 
+  // documentation inherited from the base class
```
masterleinad:
```suggestion
  /**
   * @copydoc TensorPolynomialsBase<dim>::clone()
   */
```

include/deal.II/base/polynomials_raviart_thomas.h
```diff
@@ -106,6 +107,10 @@ class PolynomialsRaviartThomas
   static unsigned int
   compute_n_pols(unsigned int degree);
 
+  // documentation inherited from base class
```
masterleinad:
```suggestion
  /**
   * @copydoc TensorPolynomialsBase<dim>::clone()
   */
```

include/deal.II/base/polynomials_rt_bubbles.h
```diff
@@ -140,6 +141,10 @@ class PolynomialsRT_Bubbles
   static unsigned int
   compute_n_pols(const unsigned int degree);
 
+  // documentation inherited from the base class
```
masterleinad:
```suggestion
  /**
   * @copydoc TensorPolynomialsBase<dim>::clone()
   */
```

GrahamBenHarper:
I got all 6 instances. Neat! I didn't know about the `@copydoc` command before.

source/base/polynomials_abf.cc
```diff
@@ -47,7 +48,8 @@ namespace
 
 template <int dim>
 PolynomialsABF<dim>::PolynomialsABF(const unsigned int k)
-  : my_degree(k)
+  : TensorPolynomialsBase<dim>(k, compute_n_pols(k))
+  , my_degree(k)
   , polynomial_space(get_abf_polynomials<dim>(k))
   , n_pols(compute_n_pols(k))
```
bangerth: This suggests that you now have duplicate data in the base class and the derived class. You could get rid of the `n_pols` variable by writing `this->n()` everywhere. I suspect that the same is true for `my_degree`.

source/base/polynomials_bdm.cc
```diff
@@ -27,7 +28,8 @@ DEAL_II_NAMESPACE_OPEN
 
 template <int dim>
 PolynomialsBDM<dim>::PolynomialsBDM(const unsigned int k)
-  : polynomial_space(Polynomials::Legendre::generate_complete_basis(k))
+  : TensorPolynomialsBase<dim>(k, compute_n_pols(k))
+  , polynomial_space(Polynomials::Legendre::generate_complete_basis(k))
   , monomials((dim == 2) ? (1) : (k + 2))
   , n_pols(compute_n_pols(k))
```
bangerth: same here

source/base/polynomials_bernardi_raugel.cc
```diff
@@ -15,13 +15,15 @@
...
 
 DEAL_II_NAMESPACE_OPEN
 
 
 template <int dim>
 PolynomialsBernardiRaugel<dim>::PolynomialsBernardiRaugel(const unsigned int k)
-  : my_degree(k)
+  : TensorPolynomialsBase<dim>(k, compute_n_pols(k))
+  , my_degree(k)
   , n_pols(compute_n_pols(k))
```
bangerth: and here

source/base/polynomials_raviart_thomas.cc
```diff
@@ -35,7 +36,8 @@ DEAL_II_NAMESPACE_OPEN
 
 template <int dim>
 PolynomialsRaviartThomas<dim>::PolynomialsRaviartThomas(const unsigned int k)
-  : my_degree(k)
+  : TensorPolynomialsBase<dim>(k, compute_n_pols(k))
+  , my_degree(k)
   , polynomial_space(create_polynomials(k))
   , n_pols(compute_n_pols(k))
```
bangerth: and here

source/base/polynomials_rt_bubbles.cc
```diff
@@ -27,7 +28,8 @@ DEAL_II_NAMESPACE_OPEN
 
 template <int dim>
 PolynomialsRT_Bubbles<dim>::PolynomialsRT_Bubbles(const unsigned int k)
-  : my_degree(k)
+  : TensorPolynomialsBase<dim>(k, compute_n_pols(k))
+  , my_degree(k)
```
bangerth: and here

source/base/polynomials_rt_bubbles.cc
```diff
@@ -27,7 +28,8 @@ DEAL_II_NAMESPACE_OPEN
...
   , raviart_thomas_space(k - 1)
   , monomials(k + 2)
   , n_pols(compute_n_pols(k))
```
bangerth: and here

GrahamBenHarper:
I'll remove some of the duplicate information here! I was concerned about breaking classes, but since we have the /rebuild command, we'll see if I accidentally break a class.

I'll be cautious and avoid submitting a ton of commits so that I don't kill the testing machines. :smiley:

bangerth:
If you have the test suite set up, you can also just do
```
  ctest -R fe.*debug
```
to only run tests that both have `fe` in the name and run only debug mode tests. That should cut down time to test drastically. `ctest` also understands the `-j4` flag (or using a different number) to run tests in parallel.

source/base/polynomials_abf.cc
GrahamBenHarper: This one works!

source/base/polynomials_bdm.cc
GrahamBenHarper: There's a concern with `PolynomialsBDM`. At the moment, it has its own function `degree()` which computes the degree as `polynomial_space.degree() - 1`. At the time that `TensorPolynomialsBase` is constructed, the quantity is unknown.

Should I leave that function as-is for now?

bangerth: Yes. Make a note about that to figure this out at a later time -- I usually use github issues for these notes so they don't get forgotten, and you should feel free to do the same :-)

GrahamBenHarper:
I have a question about the testers actually:

If I push 6 commits all at once (I'm keeping them separate so I can go back and change things as needed if something breaks), will the testers then proceed to test all 6 commits? Can I avoid having them do that?

peterrum:
If you push 6 commits, the tester will run only once, i.e. with all committed changes. Furthermore, if you push will a test is running, that test is aborted and the newest version of your PR is tested.

source/base/polynomials_bernardi_raugel.cc
GrahamBenHarper: Got it!

source/base/polynomials_raviart_thomas.cc
GrahamBenHarper: Got it!

source/base/polynomials_rt_bubbles.cc
GrahamBenHarper: Got it!

source/base/polynomials_rt_bubbles.cc
GrahamBenHarper: Got it!

GrahamBenHarper:
Thank you @peterrum!

source/base/polynomials_bdm.cc
GrahamBenHarper: Okay #8591 is there to remind us about it. :smiley:

masterleinad:
```
lib/libdeal_II.g.so.9.2.0-pre: error: undefined reference to 'dealii::PolynomialsABF<1>::n() const'
lib/libdeal_II.g.so.9.2.0-pre: error: undefined reference to 'dealii::PolynomialsABF<2>::n() const'
lib/libdeal_II.g.so.9.2.0-pre: error: undefined reference to 'dealii::PolynomialsABF<3>::n() const'
```
Seems like you missed to remove the declaration.

GrahamBenHarper:
Whoops! I fixed those declarations.

I only built the library on my laptop until around 50% when I saw the individual files compiled, so I didn't catch the linker error.

bangerth:
Since I see that you're working on this, can you also squash the commits into one, or into one per class?

GrahamBenHarper:
Yes I'll squash the commits into one commit in the next 10-30 minutes for sure. I'm running a build all the way on my end to make sure I don't miss removing any other declarations.

PR #8587, opened 2019-08-16
Enable DoFHandlerPolicy of p:d:t for 1D

peterrum:
This PR is part of the effort to introduce the new parallel::fullydistributed::Triangulation (see #8558) and is a follow-up to PR #8579.

Ping: @bangerth @tjhei

PR #8592, opened 2019-08-17
Introduce parallel::DistributedTriangulationBase

peterrum:
This PR is part of the effort to introduce the new parallel::fullydistributed::Triangulation (see #8558) and is a follow-up to PR #8587.

This PR introduces a new virtual class `parallel::DistributedTrinagulationBase` between `parallel::Trinagulation` and `parallel::distributed::Trinagulation`. This class will be also the parent class of `p:f:t`.

*Note*: Please only consider the last commit, since all other commits will be removed from this PR once  PR #8587 is merged.

*Ping*: @bangerth @tjhei

### DIFF
--- a/doc/news/changes/minor/20190817PeterMunch
+++ b/doc/news/changes/minor/20190817PeterMunch
@@ -1,0 +1,4 @@
+New: Introduce new class parallel::DistributedTriangulationBase 
+between parallel::TriangulationBase and parallel::distributed::Triangulation.
+<br>
+(Peter Munch, 2019/08/17)
--- a/include/deal.II/distributed/tria.h
+++ b/include/deal.II/distributed/tria.h
@@ -249,7 +249,7 @@ namespace parallel
      */
     template <int dim, int spacedim = dim>
     class Triangulation
-      : public dealii::parallel::TriangulationBase<dim, spacedim>
+      : public dealii::parallel::DistributedTriangulationBase<dim, spacedim>
     {
     public:
       /**
@@ -382,6 +382,12 @@ namespace parallel
        */
       virtual void
       clear() override;
+
+      /**
+       * Return if multilevel hierarchy is supported and has been constructed.
+       */
+      bool
+      is_multilevel_hierarchy_constructed() const override;
 
       /**
        * Implementation of the same function as in the base class.
@@ -1247,7 +1253,7 @@ namespace parallel
      */
     template <int spacedim>
     class Triangulation<1, spacedim>
-      : public dealii::parallel::TriangulationBase<1, spacedim>
+      : public dealii::parallel::DistributedTriangulationBase<1, spacedim>
     {
     public:
       /**
@@ -1334,6 +1340,9 @@ namespace parallel
        */
       void
       save(const std::string &filename) const;
+
+      bool
+      is_multilevel_hierarchy_constructed() const override;
 
       /**
        * This function is not implemented, but needs to be present for the
--- a/include/deal.II/distributed/tria_base.h
+++ b/include/deal.II/distributed/tria_base.h
@@ -246,6 +246,32 @@ namespace parallel
   template <int dim, int spacedim = dim>
   using Triangulation DEAL_II_DEPRECATED = TriangulationBase<dim, spacedim>;
 
+  /**
+   * A base class for distributed triangulations. It can be used to test
+   * (via <tt>dynamic_cast</tt>) whether a particular triangulation object
+   * is able to distribute meshes between processes.
+   */
+  template <int dim, int spacedim = dim>
+  class DistributedTriangulationBase
+    : public dealii::parallel::TriangulationBase<dim, spacedim>
+  {
+  public:
+    /**
+     * Constructor.
+     */
+    DistributedTriangulationBase(
+      MPI_Comm mpi_communicator,
+      const typename dealii::Triangulation<dim, spacedim>::MeshSmoothing
+                 smooth_grid = (dealii::Triangulation<dim, spacedim>::none),
+      const bool check_for_distorted_cells = false);
+
+    /**
+     * Return if multilevel hierarchy is supported and has been constructed.
+     */
+    virtual bool
+    is_multilevel_hierarchy_constructed() const = 0;
+  };
+
 } // namespace parallel
 
 DEAL_II_NAMESPACE_CLOSE
--- a/include/deal.II/dofs/dof_handler_policy.h
+++ b/include/deal.II/dofs/dof_handler_policy.h
@@ -215,7 +215,7 @@ namespace internal
 
       /**
        * This class implements the policy for operations when we use a
-       * parallel::distributed::Triangulation object.
+       * parallel::DistributedTriangulationBase object.
        */
       template <class DoFHandlerType>
       class ParallelDistributed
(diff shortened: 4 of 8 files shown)


include/deal.II/distributed/tria.h
```diff
@@ -249,7 +249,7 @@ namespace parallel
      */
     template <int dim, int spacedim = dim>
     class Triangulation
-      : public dealii::parallel::TriangulationBase<dim, spacedim>
+      : public dealii::parallel::DistributedTrinagulationBase<dim, spacedim>
```
bangerth: There's a typo in `Triangulation` here and everywhere else.

peterrum: Thanks! Ups. I have updated theses...

doc/news/changes/minor/20190817PeterMunch
```diff
@@ -0,0 +1,4 @@
+New: Intdroduce new class parallel::DistributedTrinagulationBase
```
kronbichler: There is one more typo here:
```suggestion
New: Intdroduce new class parallel::DistributedTriangulationBase 
```

jppelteret:
And in the word "introduce" in the changelog

include/deal.II/distributed/tria_base.h
```diff
@@ -246,6 +246,24 @@ namespace parallel
...
+     * Constructor.
+     */
+    DistributedTriangulationBase(
+      MPI_Comm mpi_communicator,
+      const typename dealii::Triangulation<dim, spacedim>::MeshSmoothing
+                 smooth_grid = (dealii::Triangulation<dim, spacedim>::none),
+      const bool check_for_distorted_cells = false);
+
+    virtual bool
+    do_construct_multigrid_hierarchy() const = 0;
```
tjhei: this function needs documentation. I actually have no idea what it does from the name. "do_...." means that it does something?

peterrum:
@tjhei: you are right, I have forgotten the documentation.

The motivation of this method is that in the DoFHanlderPolicy the following lines of code check if multigrid has been constructed by the triangulation or not:
```cpp
AssertThrow((triangulation->settings &
                     parallel::distributed::Triangulation<dim, spacedim>::
                       construct_multigrid_hierarchy),
                    ExcMessage(
                      "Multigrid DoFs can only be distributed on a parallel "
                      "Triangulation if the flag construct_multigrid_hierarchy "
```

In this form it cannot stay, if we want to generalize the DoFHandlerPolicy for `parallel::DistributedTriangulationBase`. I propose to introduce a method for that. The name is not fixed...

bangerth:
I think the addition of the new function is an overdesign. Right now, no other class derived from the new intermediate base class can handle MG algorithms (I suspect), and so the easiest would probably be to just modify the assertion:
```
p::d::T *p_d_tria = dynamic_cast<p::d::T*>(triangulation);
AssertThrow((p_d_tria != nullptr
                       &&
                     p_d_tria->settings &
                     parallel::distributed::Triangulation<dim, spacedim>::
                       construct_multigrid_hierarchy),
                    ExcMessage(
                      "Multigrid DoFs can only be distributed on a parallel "
                      "Triangulation if the flag construct_multigrid_hierarchy "
```
This way you will be kicked out if someone uses a different kind of triangulation, which makes sense since we likely have to revisit the whole algorithm again anyway if a different triangulation gains this kind of functionality.

peterrum:
@bangerth Thanks for the response. But you have missed that also `p:f:t` (this non an open PR yet) will take this path because of its mg capabilities. 

I really do not like to cast within this function (two times!) because it makes the class less general. If there should be in the future a distributed triangulation that does not support multigrid it would simply implement `do_construct_multigrid_hierarchy` (or a method with another name) and return `false`.

tjhei:
> ping @tjhei

Yes, it makes sense to have a function like this, but I would call it ``is_multilevel_hierarchy_constructed`` or ``is_multigrid_available`` or ``supports_multigrid`` or something like that. ``do_*`` makes it sound like the function does something.

bangerth:
I see. I didn't realize the the new class already has MG capabilities.

include/deal.II/distributed/tria_base.h
peterrum: @tjhei @bangerth I have renamed the method to `is_multilevel_hierarchy_constructed` as propsed by @tjhei  and have added some documentation.

peterrum:
@bangerth What for other users is adaptivitiy, is in our case multigrid. The consequence was that the first thing `p:f:t` was capable of was multigrid...

bangerth:
Ha, fair enough :-)

include/deal.II/distributed/tria.h
```diff
@@ -383,6 +383,9 @@ namespace parallel
       virtual void
       clear() override;
 
+      bool
+      is_multilevel_hierarchy_constructed() const override;
```
bangerth: This looks good to me, but can you add documentation to the function?

include/deal.II/distributed/tria_base.h
```diff
@@ -246,6 +246,27 @@ namespace parallel
...
+     * Constructor.
+     */
+    DistributedTriangulationBase(
+      MPI_Comm mpi_communicator,
+      const typename dealii::Triangulation<dim, spacedim>::MeshSmoothing
+                 smooth_grid = (dealii::Triangulation<dim, spacedim>::none),
+      const bool check_for_distorted_cells = false);
+
+    /**
+     * Return if multilevel hierarchy is supported or has been constructed.
```
bangerth: or -> and?

include/deal.II/distributed/tria.h
bangerth: Ah, never mind -- this is the derived class, which inherits the documentation.

peterrum: Done.

include/deal.II/distributed/tria_base.h
peterrum: Done.

peterrum:
@bangerth @tjhei  After @bangerth merged #8587, I have rebased this branch. Hopefully now all tests are passed!

include/deal.II/distributed/tria_base.h
```diff
@@ -246,6 +246,27 @@ namespace parallel
   template <int dim, int spacedim = dim>
   using Triangulation DEAL_II_DEPRECATED = TriangulationBase<dim, spacedim>;
 
+  template <int dim, int spacedim = dim>
+  class DistributedTriangulationBase
+    : public dealii::parallel::TriangulationBase<dim, spacedim>
```
bangerth: This class is missing documentation. Can you add something short that this is an intermediate class that can be used to test (via `dynamic_cast`) whether a particular triangulation object is able to distribute meshes between processors?

doc/news/changes/minor/20190817PeterMunch
```diff
@@ -0,0 +1,4 @@
+New: Introduce new class parallel::DistributedTriangulationBase 
+between parallel::Triangulation and parallel::distributed::Triangulation.
```
masterleinad:
```suggestion
between parallel::TriangulationBase and parallel::distributed::Triangulation.
```

include/deal.II/distributed/tria_base.h
peterrum: Done!

doc/news/changes/minor/20190817PeterMunch
peterrum: You are right. This is the old documentation...

PR #8593, opened 2019-08-17
Implement DoFAccessor::set/get_mg_dof_indices for 1D

peterrum:
This PR is part of the effort to introduce the new `parallel::fullydistributed::Triangulation` (see #8558) and is a follow-up to PR #8592.

This PR implements `DoFAccessor::set_mg_dof_indices` and `DoFAccessor::get_mg_dof_indices` for 1D. We need this, since `p:f:t` works for 2D, 3D, and 1D(!) for GMG.

Note: Please only consider the last commit, since all other commits will be removed from this PR once PR #8592 is merged. I still have to add tests. 

Ping: @bangerth @tjhei

### DIFF
--- a/doc/news/changes/minor/20190818PeterMunch
+++ b/doc/news/changes/minor/20190818PeterMunch
@@ -1,0 +1,3 @@
+New: Implement DoFAccessor::set_mg_dof_indices and get_mg_dof_indices for 1D.
+<br>
+(Peter Munch, 2019/08/18)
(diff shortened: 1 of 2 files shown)


bangerth:
The last commit is ok. If you do
```
  git checkout dofaccessor_mgdofindices_1D
  git rebase -i master
```
and then drop all of the commits except the last one, you will end up with a branch that has only this last commit on it. That's the one we want to review and merge. (And of course this has no influence on all of your other branches: They continue to contain all of the commits you had on there before!)

peterrum:
@bangerth I have updated all PRs so that each of them has exactly one PR.

bangerth:
Thanks, that is much better!

PR #8594, opened 2019-08-17
Replace boost::optional by std_cxx17::optional.

### LINKED ISSUE #8590: Introduce std_cxx17::optional
C++17 includes `std::optional`. We currently use `boost::optional` but what we should probably do for these locations is to introduce a namespace `std_cxx17` that contains either `std::optional` if the compiler has it, or uses `boost::optional` as an alias if it does not.

masterleinad:
Fixes #8590.

### DIFF
--- a/include/deal.II/base/std_cxx17/optional.h
+++ b/include/deal.II/base/std_cxx17/optional.h
@@ -1,0 +1,37 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+#ifndef dealii_cxx17_optional_h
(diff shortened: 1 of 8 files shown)


masterleinad:
I replaced the `DEAL_II_WITH_CXX17` check by `__cpp_lib_optional`. That should be more reliable.

include/deal.II/base/std_cxx17/optional.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef dealii_cxx17_optional_h
+#define dealii_cxx17_optional_h
+
+#include <deal.II/base/config.h>
+
+#ifdef __cpp_lib_optional
+#  include <optional>
+#else
+#  include <boost/optional.hpp>
+#endif // DEAL_II_WITH_CXX17
```
bangerth: might want to replace it here as well

include/deal.II/base/std_cxx17/optional.h
```diff
@@ -0,0 +1,37 @@
...
+
+DEAL_II_NAMESPACE_OPEN
+namespace std_cxx17
+{
+#ifdef __cpp_lib_optional
+  using std::optional;
+#else
+  using boost::optional;
+#endif // DEAL_II_WITH_CXX17
```
bangerth: and here

masterleinad:
> Is `__cpp_lib_optional` one of the C++17 feature macros?

Yes, https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B17

PR #8599, opened 2019-08-19
Add a test.

bangerth:
#8574 is failing for reasons nobody seems to quite understand. @masterleinad found out that if one switches the order of operands, that things magically works, but why is not clear.

I though that maybe the contraction of tensors uses the wrong index, so I wrote a test. The test succeeds, unfortunately, so is not helping at all. But I thought it's worthwhile having more tests, so here it is.


luca-heltai:
I think you need to reindent. A part from this, looks good.

tests/tensors/tensor_30.cc
```diff
@@ -0,0 +1,76 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2017 - 2018 by the deal.II authors
```
luca-heltai:
```suggestion
// Copyright (C) 2019 by the deal.II authors
```

tests/tensors/tensor_30.cc
```diff
@@ -0,0 +1,76 @@
...
+  // which in 2d equates to
+  //          = (i + 2j) + 3
+  // and in 3d equates to
+  //          = 3(i + 2j) + 3*5
+  //
+  // In other words, in 2d this is the 2x2 matrix
+  //    [[3,4],[4,6]]
+  // and in 3d is the 3x3 matrix
+  //    [[15,21,27],[18,24,30],[21,27,33]]
+  Tensor<2,dim> c = a*b;
```
luca-heltai:
```suggestion
  Tensor<2, dim> c = a*b;
```

tests/tensors/tensor_30.cc
```diff
@@ -0,0 +1,76 @@
...
+#include <deal.II/base/tensor.h>
+
+#include "../tests.h"
+
+using namespace dealii;
+
+template <int dim>
+void test ()
+{
+  Tensor<3,dim> a;
```
luca-heltai:
```suggestion
  Tensor<3, dim> a;
```

tests/tensors/tensor_30.cc
```diff
@@ -0,0 +1,76 @@
...
+  Tensor<1,dim> b;
```
luca-heltai:
```suggestion
  Tensor<1, dim> b;
```

bangerth:
Indentation is fixed. The copyright years will be adjusted automatically next time the corresponding script is run.

masterleinad:
> The copyright years will be adjusted automatically next time the corresponding script is run.

But that is only true for the year the file was last touched, not for the year it reports to be created, isn't it?

bangerth:
Maybe. But I copied the `main()` function from another test that had those years, so strictly speaking the years aren't wrong. I'll fix it if you want me to, but it seems like unnecessary work :-)

masterleinad:
No, I was just asking for confirmation. It might be a good idea to use `git` for the first year, too. Then, we never need to think about these dates anymore when creating/copying files.

bangerth:
I suspect that it's substantially more expensive to figure out which is the first year a file was touched. But it's also inaccurate: git doesn't actually keep track of files that were moved, and certainly not if a file was split into two, or if material was copied from one file to another -- both of which need to be reflected in the beginning date of the copyright years...

Issue #8535, opened 2019-08-09
Rename parallel::Triangulation to parallel::TriangulationBase


bangerth:
@tjhei *loves* the idea.

peterrum:
I did understand the result of our discussion as following, expressed as a class hierarchy (see also the discussion in PR#8558):
* Tringulation
  * parallel::Triangulation
    * parallel::shared::Triangulation
    * **parallel::DistributedTrinagulationBase** (?)
      * parallel::distributed::Trianguation
      * parallel::fullydistributed::Trianguation  

In this circumstance, we would not have to rename `parallel::Triangulation`...

bangerth:
@peterrum Your naming scheme is fine with me. But I also don't see much of a reason to not *also* rename `parallel::Triangulation` to `parallel::TriangulationBase`: it's a rather internal class, so I'm not too worried about it being incompatible.

peterrum:
@bangerth @tjhei  The updated class hierarchy is as follows:
* Tringulation
  * **parallel::TriangulationBase**
    * parallel::shared::Triangulation
    * **parallel::DistributedTrinagulationBase**
      * parallel::distributed::Trianguation
      * parallel::fullydistributed::Trianguation

peterrum:
We can close this issue. It has been resolved by PR #8600.

masterleinad:
You know that you can close issues automatically by saying "Fixes #..." ? https://help.github.com/en/articles/closing-issues-using-keywords

PR #8600, opened 2019-08-19
Rename parallel::Triangulation to parallel::TriangulationBase

peterrum:
references: issue #8535

ping: @bangerth @tjhei

### DIFF
--- a/doc/news/changes/incompatibilities/20190819PeterMunch
+++ b/doc/news/changes/incompatibilities/20190819PeterMunch
@@ -1,0 +1,5 @@
+Changed: The parallel::Triangulation class has been renamed to parallel::TriangulationBase
+to better reflect its purpose of a base to a series of parallel triangulations such as
+parallel::distributed::Triangulation or parallel::shared::Triangulation.
+<br>
+(Peter Munch, 2019/08/19)
--- a/include/deal.II/distributed/tria_base.h
+++ b/include/deal.II/distributed/tria_base.h
@@ -44,13 +44,13 @@ namespace parallel
    * parallel::shared::Triangulation.
    */
   template <int dim, int spacedim = dim>
-  class Triangulation : public dealii::Triangulation<dim, spacedim>
+  class TriangulationBase : public dealii::Triangulation<dim, spacedim>
   {
   public:
     /**
      * Constructor.
      */
-    Triangulation(
+    TriangulationBase(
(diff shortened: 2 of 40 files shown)


masterleinad:
Would you please also provide a changelog entry in "incompatibilities" for this?

peterrum:
@masterleinad: Done!

kronbichler:
> I assume this type is not used in user codes or should we introduce a deprecated `using Triangulation = TriangulationBase` for a while?

I am definitely in favor of providing a `using` statement with deprecation status. While I agree that it is mostly used internally (and it is used a lot as can be seen by the changes), I am sure that some projects have this in user code - for example, I have used the idiom to downcast from `Triangulation` to `parallel::Triangulation` quite a lot because that gives you access to the MPI communicator. There are certainly others doing similar things, so it would create a big glitch if we did  to not have at least one year with a deprecation warning.

doc/news/changes/incompatibilities/20190819PeterMunch
```diff
@@ -0,0 +1,3 @@
+Improvement: Rename parallel::Triangulation to parallel::TriangulationBase.
```
kronbichler: Please be a bit more verbose and avoid "improvement" because it is not really that from a user perspective. It much better describes the intent of the class - some kind of interface class. I would write:
```
Changed: The parallel::Triangulation class has been renamed to parallel::TriangulationBase
to better reflect its purpose of a base to a series of parallel triangulations such as
parallel::distributed::Triangulation or parallel::shared::Triangulation.
```

peterrum: @kronbichler: done!

peterrum:
@tjhei @kronbichler I have added the deprecated alias. Could you have a look if it is fine this way...

include/deal.II/distributed/tria_base.h
```diff
@@ -239,6 +239,10 @@ namespace parallel
     fill_level_ghost_owners();
   };
 
+
```
tjhei: Can you please add something like:
```
/**
 * Using directive for backwards-compatibility.
 * @deprecated Use TriangulationBase instead of Triangulation.
 */
```

peterrum: Done!

PR #8601, opened 2019-08-19
Add assertion to PETScWrappers::MatrixBase::begin() and end().

### LINKED ISSUE #8571: begin() and end() of PETScWrappers::MPI::SparseMatrix segfaults
This little example segfaults if run on more than one processor:
```cpp
#include <vector>
#include <iostream>

#include <deal.II/base/mpi.h>
#include <deal.II/lac/petsc_sparse_matrix.h>
#include <deal.II/lac/dynamic_sparsity_pattern.h>

using namespace dealii;
using namespace std;

int main(int argc, char **argv)
{

	int n_procs;
	MPI_Comm_size(MPI_COMM_WORLD, &n_procs);

	int this_proc;
	MPI_Comm_rank(MPI_COMM_WORLD, &this_proc);

	IndexSet local_indices(n_procs);
	local_indices.add_index(this_proc);

	vector<unsigned int> n_rows_per_proc(n_procs, 1);

	DynamicSparsityPattern dsp(local_indices);

	PETScWrappers::MPI::SparseMatrix K(MPI_COMM_WORLD, dsp, n_rows_per_proc, n_rows_per_proc, this_proc);

	K = 0.0;
	K.compress(VectorOperation::insert);

	// this fails if run with more than one processor
	// the reason is that begin() will always start from the first global line,
	// no matter what the local range is; similarly end() will always point past
	// the last global line, no matter what the local range is
	if(this_proc == 1)
		cout << K.begin()->column() << endl;
}
```

The reason seems to be that  `begin()` and `end()` of the base class `PETScWrappers::MatrixBase` don't check for the local range. So, my suggestion would be to make sure that  `begin()` and `end()` take the local range into consideration. If there's nothing opposed to this approach, I can write a patch for this.

starki0815:
Add assertion to make sure that begin() and end() can only be called on a processor owning the entire matrix. Also make sure that begin() works in case the first row(s) of the matrix is(are) emtpy.
Fixes #8571.

### DIFF
--- a/doc/news/changes/minor/20190819SebastianStark
+++ b/doc/news/changes/minor/20190819SebastianStark
@@ -1,0 +1,3 @@
+Fixed: Make sure that begin() and end() of PETScWrappers::MatrixBase can only be called on a processor owning all rows of the matrix. Also fix a bug which caused begin() to fail when the first row(s) of the matrix is(are) empty.
+<br>
+(Sebastian Stark, 2019/08/19)
--- a/include/deal.II/lac/petsc_matrix_base.h
+++ b/include/deal.II/lac/petsc_matrix_base.h
@@ -837,13 +837,17 @@ namespace PETScWrappers
     residual(VectorBase &dst, const VectorBase &x, const VectorBase &b) const;
 
     /**
-     * Iterator starting at the first entry.
+     * Iterator starting at the first entry. This can only be called on a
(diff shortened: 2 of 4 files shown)


masterleinad:
Would you please also add a test that would have failed before?

starki0815:
I have added a test for the case that the first row(s) of the matrix is(are) empty. I think adding a test checking that the assertions work is overkill.

PR #8602, opened 2019-08-19
Fix Symengine test

masterleinad:
Fixes https://cdash.43-1.org/testDetails.php?test=7384172&build=1019. Our overloads for multiplications/division where an object of type `std::complex` is involved are really greedy. In particular, we need to disable them if we have another overload defined that should have higher precedence.

### DIFF
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -45,6 +45,13 @@ template <int rank_, int dim, typename Number = double>
 class Tensor;
 template <typename Number>
 class Vector;
+namespace Differentiation
+{
+  namespace SD
+  {
+    class Expression;
+  }
+} // namespace Differentiation
 #endif
 
 #ifndef DOXYGEN
@@ -1341,7 +1348,8 @@ namespace internal
               typename OtherNumber,
               typename std::enable_if<
                 !std::is_integral<
-                  typename ProductType<Number, OtherNumber>::type>::value,
+                  typename ProductType<Number, OtherNumber>::type>::value &&
+                  !std::is_same<Number, Differentiation::SD::Expression>::value,
                 int>::type = 0>
     DEAL_II_CONSTEXPR DEAL_II_CUDA_HOST_DEV inline DEAL_II_ALWAYS_INLINE void
                       division_operator(Tensor<rank, dim, Number> (&t)[dim],
@@ -1360,7 +1368,8 @@ namespace internal
               typename OtherNumber,
               typename std::enable_if<
                 std::is_integral<
-                  typename ProductType<Number, OtherNumber>::type>::value,
+                  typename ProductType<Number, OtherNumber>::type>::value ||
+                  std::is_same<Number, Differentiation::SD::Expression>::value,
                 int>::type = 0>
     DEAL_II_CONSTEXPR DEAL_II_CUDA_HOST_DEV inline DEAL_II_ALWAYS_INLINE void
                       division_operator(dealii::Tensor<rank, dim, Number> (&t)[dim],
--- a/tests/symengine/symengine_scalar_operations_04.cc
+++ b/tests/symengine/symengine_scalar_operations_04.cc
@@ -1,0 +1,83 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE at
+// the top level of the deal.II distribution.
+//
+// ---------------------------------------------------------------------
+
+
+// Check that the std::complex overload for the division and multiplication
+// operator works.
+
+#include <deal.II/differentiation/sd.h>
+
+#include "../tests.h"
+
+using namespace dealii;
+namespace SD = Differentiation::SD;
+
+template <typename Number1, typename Number2>
+void
+test()
+{
+  using SD_number_t = SD::Expression;
+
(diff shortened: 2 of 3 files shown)


tests/symengine/symengine_tensor_operations_02.output
```diff
@@ -29,26 +29,26 @@ DEAL:Integer:Copy constructor::cs: 5 7 7 6
 DEAL:Integer:Math operators:Tensor::c+=b: 14 16 18 20
 DEAL:Integer:Math operators:Tensor::c*=s: 28 32 36 40
 DEAL:Integer:Math operators:Tensor::c-=b: 19 22 25 28
-DEAL:Integer:Math operators:Tensor::c/=s: 19/2 11 25/2 14
+DEAL:Integer:Math operators:Tensor::c/=s: 9.5 11.0 12.5 14.0
```
bangerth: ha, that's actually quite funny that that worked!

masterleinad:
I agree that the greedy approach is not nice. However, whitelisting types instead is not backward-compatible. That's why I choose to blacklist the problematic types instead so far.

bangerth:
What do others think?

Rombur:
I think that whitelisting is the way to go. If someone needs to add another overload, she shouldn't have to blacklist the new overload.

tests/symengine/symengine_tensor_operations_02.output
jppelteret: No, I think that this output is now incorrect. Division where the numerator and demoninator are both integers should result in rational type.

jppelteret: Note: You're not seeing arithmetic types being printed here. These are SymEngine types being printed with their own overloaded operator<<.

tests/symengine/symengine_tensor_operations_02.output
```diff
@@ -92,13 +92,13 @@ DEAL:Float:Math operators:Tensor::c+=x: 9.5 + x_00 11.0 + x_01 12.5 + x_10 14.0
 DEAL:Float:Math operators:Tensor::c*=y: y*(9.5 + x_00) y*(11.0 + x_01) y*(12.5 + x_10) y*(14.0 + x_11)
 DEAL:Float:Math operators:Tensor::c*=s: 2.0*y*(9.5 + x_00) 2.0*y*(11.0 + x_01) 2.0*y*(12.5 + x_10) 2.0*y*(14.0 + x_11)
 DEAL:Float:Math operators:Tensor::c-=x: -x_00 + 2.0*y*(9.5 + x_00) -x_01 + 2.0*y*(11.0 + x_01) -x_10 + 2.0*y*(12.5 + x_10) -x_11 + 2.0*y*(14.0 + x_11)
-DEAL:Float:Math operators:Tensor::c/=y: (-x_00 + 2.0*y*(9.5 + x_00))/y (-x_01 + 2.0*y*(11.0 + x_01))/y (-x_10 + 2.0*y*(12.5 + x_10))/y (-x_11 + 2.0*y*(14.0 + x_11))/y
+DEAL:Float:Math operators:Tensor::c/=y: 1.0*(-x_00 + 2.0*y*(9.5 + x_00))/y 1.0*(-x_01 + 2.0*y*(11.0 + x_01))/y 1.0*(-x_10 + 2.0*y*(12.5 + x_10))/y 1.0*(-x_11 + 2.0*y*(14.0 + x_11))/y
```
jppelteret: Has this been generated with a new version of SymEngine? They're forever changing their printed output -- its impossible to maintain these tests!

tests/symengine/symengine_tensor_operations_02.output
jppelteret: Drats, so this implies that the `Number(1.)` in
```
const Number inverse_factor = Number(1.) / factor;
```
needs to change to `Number(1)` else when `Number` is a `SD::Expression`

jppelteret:
@masterleinad Thanks for fixing the test, and sorry for the drive-by comments.

tests/symengine/symengine_tensor_operations_02.output
masterleinad: We should be able to just use the old implementation for `Differentiation::SD::Expression`.

jppelteret: Sounds fine with me.

jppelteret:
With the second commit now in place, is the first one even necessary?

masterleinad:
I would prefer if we address whitelisting/blacklisting in a separate PR/issue.

include/deal.II/base/complex_overloads.h
```diff
@@ -28,6 +28,14 @@ struct EnableIfScalar;
 template <typename T, typename U>
 struct ProductType;
 
+namespace Differentiation
```
rezarastak: Please add `#ifndef DOXYGEN` when adding forward declaration :)
I have already added this preprocessor directive to this file in #8585. So it is probably helpful if you rebase this PR on the latest `upstream/master`.

jppelteret:
> With the second commit now in place, is the first one even necessary?

I ask because at a glance it looks like we should never hit the offending line with the latest fix in place.

masterleinad:
> I ask because at a glance it looks like we should never hit the offending line with the latest fix in place.

The compiler complained about an ambiguous call to `operator/` with operand types `dealii::Differentiation::SD::Expression` and `const std::__1::complex<float>`. Why should we never hit this problem?

jppelteret:
> Why should we never hit this problem?

I clearly did not impress on you how quick that glance was :-D
I misread this, split over two lines.
``` c++
  std::is_integral<typename ProductType<Number, OtherNumber>::type>::value
```

Thanks for the fix :-)

masterleinad:
I rebased onto upstream/master and cleaned up the commit history.

masterleinad:
I created #8607 for whitelisting types instead.

masterleinad:
I rebased the pull request to only add a test and fix the currently failing test.

PR #8613, opened 2019-08-21
Fixed small bug in OCC::interpolation_curve.

luca-heltai:
With this fix:

image

Without this fix:

Screenshot 2019-08-21 at 14 52 46


tjhei:
looks good! :+1:

PR #8614, opened 2019-08-21
Only allow floating point types in complex_overloads.h

### LINKED ISSUE #8607: Whitelist types usuable for std::complex multiplication/devision overloads
Currently, we allow any type marked with `EnableIfScalar` to be used in the `std::complex` overloads for multiplication and division in `include/deal.II/base/complex_overloads.h`. As #8602 shows, this is problematic in case these types define `operator*` or `operator/` on their own.
We should make the implementation in `include/deal.II/base/complex_overloads.h` less greedy and whitelist suitable types instead.

masterleinad:
Fixes #8607. This appears simpler than I expected. At least, I could compile and all the `symengine` (apart from the one addressed in #8602), `tensor` and `complex` tests pass.

### DIFF
--- a/doc/news/changes/incompatibilities/20190821DanielArndt
+++ b/doc/news/changes/incompatibilities/20190821DanielArndt
@@ -1,0 +1,4 @@
+Changed: The complex operator overloads for multiplication and division are
+restricted to floating point types.
+<br>
+(Daniel Arndt, 2019/08/21)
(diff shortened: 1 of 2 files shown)


masterleinad:
> I can't wait till we can write these things with constraints/concepts, rather than using the clunky `enable_if`...
 
Well, we could start to support C++20 features already. It will just be a really long time until we don't need `std::enable_if` for older compilers anymore.

PR #8617, opened 2019-08-21
Add std::complex support to DataPostprocessor

dangars:
Hi, I would like to add support for a vector valued `std::complex` DataPostprocessor but I think that I took the wrong approach.

Could you give me a hint :) ?

I think that it should be feasible because the machinery to write a vector valued `std::complex` solution is already in place.

The members in `DataPostprocessor` are `double` and I have the feeling that `add_data_vector()` converts any `Vector<number>` to `Vector<double>`. That is a problem because `std::complex` can not be converted to double.

### DIFF
--- a/include/deal.II/numerics/data_out_dof_data.h
+++ b/include/deal.II/numerics/data_out_dof_data.h
@@ -213,7 +213,7 @@ namespace internal
      *
      * @author Wolfgang Bangerth, 2004
      */
-    template <typename DoFHandlerType>
+    template <typename DoFHandlerType, typename Number = double>
     class DataEntryBase
     {
     public:
@@ -233,9 +233,9 @@ namespace internal
        * case, the names and vector declarations are going to be acquired from
        * the postprocessor.
        */
-      DataEntryBase(const DoFHandlerType *dofs,
-                    const DataPostprocessor<DoFHandlerType::space_dimension>
-                      *data_postprocessor);
+      DataEntryBase(const DoFHandlerType *           dofs,
+                    const DataPostprocessor<DoFHandlerType::space_dimension,
+                                            Number> *data_postprocessor);
 
       /**
        * Destructor made virtual.
@@ -366,8 +366,8 @@ namespace internal
        * Pointer to a DataPostprocessing object which shall be applied to this
        * data vector.
        */
-      SmartPointer<
-        const dealii::DataPostprocessor<DoFHandlerType::space_dimension>>
+      SmartPointer<const dealii::
+                     DataPostprocessor<DoFHandlerType::space_dimension, Number>>
         postprocessor;
 
       /**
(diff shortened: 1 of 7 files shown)


include/deal.II/numerics/data_postprocessor.h
```diff
@@ -250,7 +250,7 @@ namespace DataPostprocessorInputs
      * or DataPostprocessorTensor may pass this flag to the constructor of
      * these three classes.
      */
-    std::vector<Tensor<1, spacedim>> solution_gradients;
+    std::vector<Tensor<1, spacedim, Number>> solution_gradients;
```
dangars: This is `double` and I have the feeling that `DataPostprocessor` will convert any `number` to `double`. This wouldn't work for `std::complex`

include/deal.II/numerics/data_out_dof_data.h
```diff
@@ -1094,12 +1113,17 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::add_data_vector(
...
 template <typename VectorType>
 void
-DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::add_data_vector(
-  const VectorType &                                        vec,
-  const DataPostprocessor<DoFHandlerType::space_dimension> &data_postprocessor)
+DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim, Number>::
+  add_data_vector(const VectorType &vec,
+                  const DataPostprocessor<DoFHandlerType::space_dimension,
+                                          typename VectorType::value_type>
+                    &data_postprocessor)
```
dangars: I made the type of `DataPostprocessor` equal to the type of the `Vector`. But I have the feeling that you do a conversion. For example `Vector` can be `float` and you convert it to `double`. This wouldn't work for `std::complex`

Why you convert `float` to `double`? It is possible to store `float` in the vtk/vtu file, there is no need for a conversion.

bangerth:
`DataOut` and friends already support complex numbers: solution vectors are just separated into real and imaginary parts to create two scalar fields out of things. I implemented this as part of the (still unfinished) step-58. The two scalar parts will then also be passed as two solution components to the `DataPostprocessor`. So in some sense everything you want already works.

Why did I do it this way, rather than making everything templated on the scalar type? Because output formats generally do not understand anything other than real-valued types. So at some point one has to take complex-valued types apart and put them into two components. This currently happens just before we call the `DataPostprocessor` and/or lower the solution vectors to the `Patch` format.

In other words, I think that everything you need is already there. I'm not sure we need to make things more complicated beyond this :-)

dangars:
Thanks! I understand now. It almost works :)

The current implementation of `DataPostprocessor` works for a `std::complex` scalar but I think that it fails for a `std::complex` vector (unless it is a bug in my program)

I close this PR and open the issue #8625 We can continue the conversation in the issue.

PR #8619, opened 2019-08-21
Rename TensorPolynomials::compute() to ::evaluate().

bangerth:
We've used this kind of terminology in a number of other places. The point is
that we don't 'compute' the polynomial at that point; its form is fixed. Rather,
we 'evaluate' the polynomial for a particular value of its argument.

@GrahamBenHarper -- FYI.

Part of #1973.


Rombur:
The format is wrong

bangerth:
Khkhkh. Let's try this again...

bangerth:
Seems to pass the testers now.

Issue #7821, opened 2019-03-17
AMD GPU support


Rombur:
Someone asked me about deal.II support for AMD GPU recently, so I looked how we could support both type of GPUs. The good news is that HIP C++ (AMD's language) is extremely close to CUDA. Many CUDA functions have an equivalent in HIP C++ for example: `cudaMalloc` and `hipMalloc` and all of our kernels are mostly compatible with both languages (the only thing that is not compatible is that `threadIdx.x` becomes `hipThreadIdx_x`). There are a few things that HIP C++ doesn't support that CUDA supports, the big one is UVM but we don't use it. So I think that at least for `MatrixFree` there is a way to support both GPUs with some small refactors. For the `SparseMatrix`, we use CUDA libraries so it would require more work. I see two ways forward: 

1. We encapsulate the CUDA code and provide an HIP C++ version of the code
2. AMD provides a script that automatically transform CUDA code in HIP C++. So we could just keep most of our CUDA code unchanged and just make sure that the script is able to convert all the CUDA code.

Personally, I prefer solution 2. I think it will be less work and since I don't have access to an AMD GPU, there is no way to check that the code we would write is correct. Of course there is still the problem of the `#if DEAL_II_WITH_CUDA`, some should be kept some should be changed but at least it is a step in the right direction. What do you all think?

masterleinad:
The second approach sounds more favorable for me as well. From playing around with `HIP` a bit, it seems that the code generated by `hipify` can still be compiled by `nvcc` provided the `rocm` headers can found in the include directories. We then only have to make sure that these always exist when we compile with `CUDA` support, i.e. we should bundle `HIP`. Otherwise, we would break `CUDA` support for people only having `nvcc`.

We still need to try if all the code can be converted, though.

Rombur:
The way I see it we wouldn't change much for CUDA. If someone wants to use AMD, she would have to run `hipify` and then compile. Our job would be to make sure that the `HIP` conversion step works. I don't want everybody to have to use `hipcc` if they don't need to. It's already a pain to get `cmake` and CUDA to work on every platform, I don't want to make that more complicated than it is.

masterleinad:
I just tried. At least, it doesn't work out of the box to compile using `hipcc` or `nvcc` after running `hipify`. I guess we need to write some kind of tutorial once someone has successfully done that.

jppelteret:
[...]

PR #8621, opened 2019-08-22
Fourier::calculate only works with double

Rombur:
I don't know why it works with most compiler but if you look at [this line](https://github.com/dealii/dealii/compare/master...Rombur:fe_series?expand=1#diff-ca3f80c49e696f0c490c9e4dd93bcfd3R211) `matrix[i][j]` is of type `std::complex<double>` and `local_dof_values[j]` is of type `Number` but this only works if `Number` is a `double`

Part of #7821

### DIFF
--- a/source/fe/fe_series_fourier.cc
+++ b/source/fe/fe_series_fourier.cc
@@ -209,7 +209,8 @@ namespace FESeries
 
     for (unsigned int i = 0; i < unrolled_coefficients.size(); i++)
       for (unsigned int j = 0; j < local_dof_values.size(); j++)
-        unrolled_coefficients[i] += matrix[i][j] * local_dof_values[j];
+        unrolled_coefficients[i] +=
+          matrix[i][j] * static_cast<double>(local_dof_values[j]);
 
     fourier_coefficients.fill(unrolled_coefficients.begin());
   }


masterleinad:
What exactly is the problem here? What is the error message? Aren't we just multiplying an `std::complex<double>` object with a floating-point number object?

Rombur:
>  Aren't we just multiplying an std::complex<double> object with a floating-point number object?

Here is the error message: `error: invalid operands to binary expression ('const dealii::AlignedVector<std::complex<double> >::value_type' (aka 'const std::complex<double>') 
and 'float')`

If you look [here](https://en.cppreference.com/w/cpp/numeric/complex/operator_arith3), you can see that you cannot mix `std::complex<double>` and `float`

masterleinad:
That is why we have `complex_overloads.h`, isn't it? As long as the underlying types are floating-point types we should have an implementation that defines the product.

Rombur:
@marcfehling I was not starting a poll ;-) There are many other problems with `hipcc` that I still don't understand. I have a limited amount of time that I can spend on making deal.II work on AMD GPU and so I won't spend more time on this bug right now. Of course you are more than welcome to find  a better fix.

Rombur:
This is not necessary anymore using HIP 2.9

PR #8626, opened 2019-08-23
Clarify something in the readme file that seems to perpetually perplex people

bangerth:


bangerth:
No reason for the line breaks -- all lines in this file are overly long, and I just tried to make the minimal number of changes to simplify the review.

Re compiler basics: I think it's in reality a language issue. For you and me, `/path/to/some/directory` is easy to read as a shorthand that needs to be replaced. For someone struggling with English, that might not be the case because the words don't mean much to them.

tjhei:
> Re compiler basics: I think it's in reality a language issue. For you and me, `/path/to/some/directory` is easy to read as a shorthand that needs to be replaced. For someone struggling with English, that might not be the case because the words don't mean much to them.

I am sure people will still do things incorrectly. Maybe we should bail out with an error message if somebody tries to use "/path/*" ? :-)
Alternatively, maybe we should just put "~/deal-ii-installation/" there?

bangerth:
On 8/23/19 11:50 AM, Timo Heister wrote:
> I am sure people will still do things incorrectly. Maybe we should bail out 
> with an error message if somebody tries to use "/path/*" ? :-)

Not a silly suggestion.


> Alternatively, maybe we should just put "~/deal-ii-installation/" there?

I don't think you can win this. People will then ask where cmake put their 
deal.II installation.

PR #8627, opened 2019-08-23
Rename settings_ to settings in distributed/tria.cc

peterrum:
FYI @tjhei: to be consistent

### DIFF
--- a/source/distributed/tria.cc
+++ b/source/distributed/tria.cc
@@ -2123,20 +2123,20 @@ namespace parallel
       MPI_Comm mpi_communicator,
       const typename dealii::Triangulation<dim, spacedim>::MeshSmoothing
                      smooth_grid,
-      const Settings settings_)
+      const Settings settings)
       : // Do not check for distorted cells.
         // For multigrid, we need limit_level_difference_at_vertices
         // to make sure the transfer operators only need to consider two levels.
       dealii::parallel::DistributedTriangulationBase<dim, spacedim>(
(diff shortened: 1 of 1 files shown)


bangerth:
YES! I've never been a fan of these underscores :-)

masterleinad:
Honestly, I don't like using the same name for a function argument and a member variable at all. I would try to avoid shadowing as much as possible but that might just be me.

tjhei:
> Honestly, I don't like using the same name for a function argument and a member variable at all. I would try to avoid shadowing as much as possible but that might just be me.

I think this is the only place where shadowing is okay. The alternative is a) that the member is named awkwardly, b) the argument is named awkwardly (which shows up in the documentation), or c) they have different names and it is not obvious to me that they are the same thing.
Maybe adopting something like bla_ is not a bad idea, but we don't do that consistently...

bangerth:
On 8/23/19 11:49 AM, Timo Heister wrote:
> I think this is the only place where shadowing is okay. The alternative is a) 
> that the member is named awkwardly, b) the argument is named awkwardly (which 
> shows up in the documentation), or c) they have different names and it is not 
> obvious to me that they are the same thing.

To be clear, I only advocate for this style if the two are identical -- i.e., 
if the member variable is initialized with the function argument of same name. 
In this case, there is no question what *value* is being used if the name 
appears in the remainder of the function (even though there is a legitimate 
question as to what *variable* is being used.

PR #8633, opened 2019-08-23
Add initializing constructor to DiagonalMatrix

masterleinad:
So far constructing a  `DiagonalMatrix` object required to use intermediate which seems to be really unnecessary. This pull request introduces a constructor that immediately initializes the underlying vector.

### DIFF
--- a/include/deal.II/lac/diagonal_matrix.h
+++ b/include/deal.II/lac/diagonal_matrix.h
@@ -52,9 +52,15 @@ public:
   using size_type  = typename VectorType::size_type;
 
   /**
-   * Constructor.
+   * Default constructor. The object needs still to be reinitialized to be
+   * usable.
    */
   DiagonalMatrix() = default;
+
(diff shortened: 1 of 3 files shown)


include/deal.II/lac/diagonal_matrix.h
```diff
@@ -52,10 +52,16 @@ class DiagonalMatrix : public Subscriptor
...
-   * Constructor.
+   * Default constructor. The object needs still to be reinitialized to be
+   * usable.
    */
   DiagonalMatrix() = default;
 
+  /**
+   * Constructor immediately initializing the object properly.
+   */
+  explicit DiagonalMatrix(const VectorType &vec);
```
bangerth: Can you document *what* the object is initialized to? Namely, to a matrix of size nxn where n is the size of the vector, and with diagonal entries equal to the elements of vec.

PR #8579, opened 2019-08-15
Generalize DoFHandlerPolicy of p:d:t

peterrum:
This PR is part of the effort to introduce the new `parallel::fullydistributed::Triangulation` (see #8558) and is a follow-up to PR #8586.

In this PR, I eliminate the direct usage of `coarse_cell_to_p4est_tree_permutation` and `p4est_tree_to_coarse_cell_permutation` int the `DoFHandlerPolicy` of `p:d:t`. The translation is moved inside of `translate_coarse_cell_id_to_coarse_cell_index` and `translate_coarse_cell_index_to_coarse_cell_id` of `p:d:t`, since it is a `p:d:t`-specific thing and is independent from the policy.

Ping: @bangerth @tjhei

PR #8635, opened 2019-08-24
Fixed: tests/mpi/hp_constraints_consistent*

### LINKED ISSUE #8616: New test fails
The new test that came in via #8555 fails on my system with the following error message. @marcfehling -- any idea? Does the test succeed on your system?

```
7388: --------------------------------------------------------
7388: An error occurred in line <112> of file </home/fac/f/bangerth/p/deal.II/1/dealii/tests/mpi/hp_constraints_consistent_03.cc> in function
7388:     void test(unsigned int, unsigned int, bool) [with int dim = 2]
7388: The violated condition was: 
7388:     cell->center()[0] < 0
7388: Additional information: 
7388:     This exception -- which is used in many places in the library -- usually indicates that some condition which the author of the code thought must be satisfied at a certain point in an algorithm, is not fulfilled. An example would be that the first part of an algorithm sorts elements of an array in ascending order, and a second part of the algorithm later encounters an element that is not larger than the previous one.
7388: 
7388: There is usually not very much you can do if you encounter such an exception since it indicates an error in deal.II, not in your own program. Try to come up with the smallest possible program that still demonstrates the error and contact the deal.II mailing lists with it to obtain help.
7388: --------------------------------------------------------
7388: 
7388: [up:65196] *** Process received signal ***
7388: [up:65196] Signal: Aborted (6)
7388: [up:65196] Signal code:  (-6)
7388: [up:65196] [ 0] /lib64/libpthread.so.0(+0xf5d0)[0x7f56df0f95d0]
7388: [up:65196] [ 1] /lib64/libc.so.6(gsignal+0x37)[0x7f56dda0c2c7]
7388: [up:65196] [ 2] /lib64/libc.so.6(abort+0x148)[0x7f56dda0d9b8]
7388: [up:65196] [ 3] /home/fac/f/bangerth/p/deal.II/1/build/lib/libdeal_II.g.so.9.2.0-pre(_ZN6dealii18deal_II_exceptions9internals22do_issue_error_nothrowERKNS_13ExceptionBaseE+0x0)[0x7f5703ccfef1]
7388: [up:65196] [ 4] /home/fac/f/bangerth/p/deal.II/1/build/tests/mpi/hp_constraints_consistent_03.debug/hp_constraints_consistent_03.debug(_ZN6dealii18deal_II_exceptions9internals20issue_error_noreturnINS_18StandardExceptions16ExcInternalErrorEEEvNS1_17ExceptionHandlingEPKciS7_S7_S7_T_+0x71)[0x47210e]
7388: [up:65196] [ 5] /home/fac/f/bangerth/p/deal.II/1/build/tests/mpi/hp_constraints_consistent_03.debug/hp_constraints_consistent_03.debug(_Z4testILi2EEvjjb+0x6f7)[0x474340]
7388: [up:65196] [ 6] /home/fac/f/bangerth/p/deal.II/1/build/tests/mpi/hp_constraints_consistent_03.debug/hp_constraints_consistent_03.debug(main+0xbf)[0x467beb]
7388: [up:65196] [ 7] /lib64/libc.so.6(__libc_start_main+0xf5)[0x7f56dd9f8495]
7388: [up:65196] [ 8] /home/fac/f/bangerth/p/deal.II/1/build/tests/mpi/hp_constraints_consistent_03.debug/hp_constraints_consistent_03.debug[0x465fa9]
7388: [up:65196] *** End of error message ***
```

marcfehling:
Closes #8616.

I didn't verify these changes on my local testsuite yet (still occupied by checking a different branch), but these changes should suffice. I'll let you know once my testsuite passes.


tjhei:
I would prefer a check of the cellid over a floating point check with ==. Do you know why #8616 is an issue?

marcfehling:
It appears like `GridGenerator::hyper_cross` enumerates cells differently on @bangerth's machine, and I have no idea why. On my machine, the leftmost cell was given the ID `1_0:` which doesn't seem to be the case for @bangerth.

bangerth:
Really? This seems like a rather unlikely problem. I can check that tomorrow, but I'm really surprised that that would be it. Does the test work for everyone else?

masterleinad:
> It appears like `GridGenerator::hyper_cross` enumerates cells differently on @bangerth's machine, and I have no idea why. On my machine, the leftmost cell was given the ID `1_0:` which doesn't seem to be the case for @bangerth.

Just for the record: This test does not work for any configuration we test for (https://cdash.43-1.org/testSummary.php?project=1&name=mpi%2Fhp_constraints_consistent_03.mpirun%3D1.debug&date=2019-08-24). It seems to have start failing between these commits: https://github.com/dealii/dealii/compare/a96482633608a8816389ded0632ca5e17bb92adf...d172f149393106a5621899ffe7c36863a73ba7b1

peterrum:
@bangerth @marcfehling: are you working both on master? The definition of `CellID` has changed recently (PR #8567 and PR #8579):

- https://github.com/dealii/dealii/blob/master/source/grid/cell_id.cc#L157-L168
- https://github.com/dealii/dealii/blob/master/source/distributed/tria.cc#L4564-L4580

In the case of `p:d:t` (as used in this test) this might lead to a different `CellID`. Working directly with `cell->id().to_string()` should be not allowed (except for debugging reasons), since the definition of the `CellID` is arbitrary: the only requirement - from point of the user - is that it is unique.

marcfehling:
I'll try to find the exact commit responsible with a `git bisect` session.

peterrum:
> I'll try to find the exact commit responsible with a `git bisect` session.

@marcfehling It is with high probability e76451737dc011a7c87957cdb817dfd80a7089b7. But as I stated above, I don't think I have introduced there a bug. See also the doxygen documentation of `CellID` which states:

> A class to represent a unique ID for a cell in a Triangulation. 

> Note: How this data is internally represented is not of importance (and not exposed on purpose).

As I read these comments: the ID is guaranteed to be unique, however, user should not rely on the actual value... This requirement is - as far as I know - still fulfilled.

marcfehling:
> @marcfehling It is with high probability [e764517](https://github.com/dealii/dealii/commit/e76451737dc011a7c87957cdb817dfd80a7089b7).

@peterrum It is indeed that particular commit. I didn't want to accuse you of introducing a bug -- I just wasn't aware that the `CellId` class has been recently changed. Thank you for your hint!

peterrum:
@marcfehling No worries. But, this issue might a indicator that there might be people that use or misuse the `CellId`s in such a way that their code might break...

tjhei:
> As I read these comments: the ID is guaranteed to be unique, however, user should not rely on the actual value... This requirement is - as far as I know - still fulfilled.

well, I used this in tests all over the place and I don't think how and why this should not work.

tjhei:
> misuse the `CellId`s in such a way that their code might break...

By the way, I introduced the cellid precisely for that reason (to be able to refine a mesh in a deterministic and simple manner in unit tests).

peterrum:
@tjhei: for `p:d:t` the coarse-cell id is - as it introduced by PR #8579 - the p4est tree and not the coarse-cell index any more. Since the coarse-cell id is used in `CellID::to_string()`, this method might return a different string.

tjhei:
> might return a different string.

well, that just means the test needs updating to the current master. The PR was probably overlapping with your PR. The cellid is still deterministic, though.

peterrum:
> well, that just means the test needs updating to the current master. 

Yes. (I have updated 2 old tests in that PR.)

> The cellid is still deterministic, though.

Yes, if `p4est_tree_to_coarse_cell_permutation` and `coarse_cell_to_p4est_tree_permutation` are deterministic.

marcfehling:
> > might return a different string.
> 
> well, that just means the test needs updating to the current master. The PR was probably overlapping with your PR. The cellid is still deterministic, though.

Alright! I'll adjust this fix to take into account the new `CellId` implementation and refrain from checking the position of each cell's center.

marcfehling:
Checking for `CellID` now. Ready for merge once testers are happy.

PR #8637, opened 2019-08-25
hp::DoFHandler: Provide own CoarseningStrategy while transferring active_fe_indices on p::d::Triangulations.

marcfehling:
A [discussion on our mailing list](https://groups.google.com/forum/#!msg/dealii/T_Ykn8KXxME/qUPBAt6GDgAJ) revealed a bug in our transfer algorithm for `active_fe_indices` in case a `p::d::Triangulation` has been used.

Currently, we set and send `active_fe_indices` based on the refinement flags on the `p::d::Triangulation` object. However, `p4est` has the last word on deciding which cells will be refined -- and `p4est` makes use of it in the specific scenario provided as a test.

This PR resolves this issue.

### DIFF
--- a/tests/mpi/solution_transfer_05.cc
+++ b/tests/mpi/solution_transfer_05.cc
@@ -1,0 +1,127 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 1998 - 2018 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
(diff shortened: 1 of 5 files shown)


tests/mpi/solution_transfer_05.cc
```diff
@@ -0,0 +1,127 @@
...
+
+#include <iostream>
+#include <vector>
+
+#include "../tests.h"
+
+
+template <int dim>
+void
+transfer(std::ostream &out)
```
marcfehling: Just realized that we don't need the `ostream` here. I'll remove it and rename the function to `test`.

PR #8641, opened 2019-08-25
p::d::SolutionTransfer: Provide optimized pack/unpack functions.

marcfehling:
As discussed in #8513, boost serialization is potentially slow and we should thus rely on optimized pack & unpack functions for data types that we know, i.e. as in #8522.

This PR provides such optimized functions for `p::d::SolutionTransfer`.


marcfehling:
Addressed suggestions. Renamed all occurrences of `dofvalues` to `dof_values`.
Ready for final review/merge once testers are happy.

marcfehling:
Aah different question: We introduced these optimized functions in an anonymous namespace (in this PR as well as #8522).

Should we introduce those in an `internal` namespace instead, i.e. `dealii::internal::parallel::distributed::SolutionTransferImplementation::Implementation`?

bangerth:
On 8/27/19 1:47 AM, marcfehling wrote:
> Should we introduce those in an |internal| namespace instead, i.e. 
> |dealii::internal::parallel::distributed::SolutionTransferImplementation::Implementation|?

If there is no reason to call these functions from anywhere outside this one 
file, then an anonymous namespace is the right place.

PR #8646, opened 2019-08-26
Fix AffineConstraints::set_zero for MemorySpace::CUDA

masterleinad:
We need to make sure not to ask the partitioner for the local dof index of global dofs that are not stored. This happens in particular for distributed vectors when the `AffineConstraints` object also contains locally relevant dofs.


kronbichler:
Excellent.

Issue #8625, opened 2019-08-22
DataPostprocessor fails with std::complex Vector


dangars:
Following up on #8617, in step-68 I calculate the eigenmodes of an electromagnetic cavity. The variable is the complex vector `electric_field`. To calculate `h_field= i*curl(E)/(omega*mu)` I use `DataPostprocessor` which seems to work fine except that it repeats names.

The problem is that `h_field` and `electric_field` are std::complex vectors. In the case of `electric_field` I can choose `DataComponentInterpretation::component_is_scalar` then give the appropriate names. In the case of `h_field` I don't have the option to give the names, therefore the names of `h_field` will be repeated. `h_field` has 6 fields 3x2, 3 for the dimensions a 2 for real and imaginary.

This is the output of the debugger for the `VTK` names, as you can see the names are repeated for `h_field`, therefore the program crashes.
```
$1 = std::vector of length 12, capacity 16 = {"electric_field_x_re",
"electric_field_x_im", "electric_field_y_re", "electric_field_y_im",
"electric_field_z_re", "electric_field_z_im", "h_field_re", "h_field_im",
"h_field_re", "h_field_im", "h_field_re", "h_field_im"}
```

I use the following code:

In the output function:
``` c++
HfieldPostprocessor<dim> h_field;
data_out.add_data_vector(locally_relevant_eigenfunctions, h_field);
```

The class code:
```c++
template <int dim> class HfieldPostprocessor : public DataPostprocessorVector<dim> {
public:
  HfieldPostprocessor() : DataPostprocessorVector<dim>("h_field", update_gradients) {}
  virtual void evaluate_vector_field(const DataPostprocessorInputs::Vector<dim> &input_data,
                                     std::vector<Vector<double>> &computed_quantities) const;
};

template <int dim>
void HfieldPostprocessor<dim>::evaluate_vector_field(const DataPostprocessorInputs::Vector<dim> &input_data,
                                                     std::vector<Vector<double>> &computed_quantities) const {
  // Ensure that there really are as many output slots as there are points at which DataOut provides the curl values:
  AssertDimension(input_data.solution_gradients.size(), computed_quantities.size());
  for (unsigned int p = 0; p < input_data.solution_gradients.size(); ++p) {
    // Ensure that each output slot has exactly 'dim' components (as should be expected, given that we want to create
    // vector-valued outputs), and copy the curl of the solution at the evaluation points into the output slots:
[...]

PR #8648, opened 2019-08-27
Make postprocessors work when using complex-valued solutions.

bangerth:
It turned out that I had this patch sitting in my repository for more than a year now. It is necessary for the step-58 program that I have yet to finish.

Passes all of my tests. I have a feeling that someone is going to ask me for a test but wanted to put this out there for the discussion in #8625. Would appreciate feedback.


bangerth:
I've re-thought this patch. I think my original approach was wrong, and what is really the case is that we have never actually tested complex-valued solutions with `DataOut` postprocessors at all. So I've rewritten the patch so that it simply rules that out with an assertion. This passes all tests (because we don't have any tests that complex-valued solutions, `DataOut`, and postprocessors at the same time), and makes sure that we exit with an error early on, rather than doing something silly downstream.

bangerth:
I'll try to fix the actual omission of the functionality in a follow-up patch. 
(FYI: @dangars)

/rebuild

source/numerics/data_out.cc
```diff
@@ -200,35 +210,231 @@ DataOut<dim, DoFHandlerType>::build_one_patch(
...
-                      scratch_data.patch_values_system.solution_gradients);
-                  if (update_flags & update_hessians)
-                    this->dof_data[dataset]->get_function_hessians(
-                      this_fe_patch_values,
-                      internal::DataOutImplementation::ComponentExtractor::
-                        real_part,
-                      scratch_data.patch_values_system.solution_hessians);
+                  scratch_data.resize_system_vectors(
+                    n_components *
+                    (this->dof_data[dataset]->is_complex_valued() ? 2 : 1));
```
dangars: @bangerth If I understand correctly. If the solution is a scalar value, I will get 2 components in DataPostprocessor (real and imag). If the solution is a complex vector with 3 components, I will get 6 components in DataPostprocessor that contain the real and the imaginary parts. Is this correct?

That is why you mentioned in #8625 that is possible to calculate complex derive quantities because you have access to all the real and imaginary components at the same time in DataPostprocessor. I realized that I misunderstood :)

bangerth:
On 9/5/19 3:15 AM, Daniel Garcia-Sanchez wrote:
> @bangerth <https://github.com/bangerth> If I understand correctly. If the 
> solution is a scalar value, I will get 2 components in DataPostprocessor (real 
> and imag). If the solution is a complex vector with 3 components, I will get 6 
> components in DataPostprocessor that contain the real and the imaginary parts. 
> Is this correct?

Yes, correct. Except that I haven't implemented all of this yet, but that's 
the plan at least.

dangars:
> Yes, correct. Except that I haven't implemented all of this yet, but that's the plan at least.

This sounds great! thanks :)

bangerth:
I've updated this patch and it works now for postprocessors and complex-valued solutions -- for the moment only for scalar solutions, but that will be fixed in follow-ups.

The key to understanding most of the patch is to realize the following two issues:
* When using a postprocessor, *all* results are real-valued, regardless of whether the inputs are real- or complex-valued. To fix the fact that I had not originally accounted for this required the addition of a fair number of `&& (this->dof_data[dataset]->postprocessor == nullptr)` conditions when we ask whether the input is complex-valued.
* We need to have code that extracts both the real and imaginary parts of a complex-valued solution, and puts these things into the `DataPostprocessorInputs` structure. This is repetitive, but I don't see a better solution to the problem -- we have to somehow pick the various parts apart. The way I currently do this would clearly be very onerous if anyone ever thought about using octonions as the elements of some input vector, but I feel that this is not likely to happen anytime soon.

I've added a test that verifies that everything we do here works as intended -- which, indeed, it does!

bangerth:
Ping?

bangerth:
Ping?

bangerth:
@kronbichler -- what, not a single comment? Have I started writing clean code finally? ;-)

kronbichler:
I am maybe not the most critical reviewer, but the code seemed very clean to me. :wink:

bangerth:
I take my compliments where I can get them ;-)

PR #8650, opened 2019-08-27
Simplify and clarify the use of std::complex in DataPostprocessor.

bangerth:


examples/step-29/step-29.cc
```diff
@@ -333,19 +333,26 @@ namespace Step29
...
-    // $v$ and $w$:
+    // The computation itself is straightforward: We iterate over each
+    // entry in the output vector and compute $|u|$ from the
+    // corresponding values of $v$ and $w$. We do this by creating a
+    // complex number $u$ and then calling `std::abs()` on the
+    // result. (One may be tempted to call `std::norm()`, but in a
+    // historical quirk, the C++ committee decided that `std::norm()`
+    // should return the <i>square</i> of the absolute value --
+    // thereby not satisfying the properties mathematicians require of
+    // something called a "norm".)
```
dangars: This is very unfortunate. This means that `std::norm` can lead to subtle errors if you don't know that it is not a norm.

PR #8657, opened 2019-08-27
Add test for std::complex DataPostprocessorVector

dangars:
This tests `std::complex DataPostprocessorVector`. This is related to #8652 

Note that the test fails and shouldn't be merged before #8625 is fixed.

This test is based on `tests/data_out/data_out_postprocessor_vector_01-new.cc`


dangars:
> I think that might take a bit of work, but at least I know what needs to be done!

Thanks @bangerth !

bangerth:
I've put this test in #8769.

PR #8662, opened 2019-08-27
Allow using PreconditionChebyshev for parallel::distributed::Vector with MemorySpace::CUDA

masterleinad:
Based on #8646. The test is a modfied version of `cuda/matrix_free_matrix_vector_10`. We need to set the constrained degrees of freedom in the initial vector to zero to get the same results as with the matrix-based implementation.


kronbichler:
Can you please look at the `step-59` case with double affine constraints and float vectors? It seems we are missing that instantiation of the new function.

kronbichler:
Do you want to squash/rebase or merge as is?

masterleinad:
I squashed the commits and rebased.

PR #8667, opened 2019-08-29
Doc updates in a variety of places.

bangerth:
While reading up on places as I write the finite element section of the paper.


bangerth:
Change made.

PR #8671, opened 2019-08-30
parameter handling for a couple of classes improved

rezarastak:
Found a missing `param.get_integer()` call in one of the classes. Also added some bounds on the parameter values.

Question: are we planning to rewrite the parameter handling of all classes to follow the style of `ParameterAcceptor`?

### DIFF
--- a/include/deal.II/algorithms/newton.templates.h
+++ b/include/deal.II/algorithms/newton.templates.h
@@ -50,9 +50,9 @@ namespace Algorithms
   {
     param.enter_subsection("Newton");
     ReductionControl::declare_parameters(param);
-    param.declare_entry("Assemble threshold", "0.", Patterns::Double());
-    param.declare_entry("Stepsize iterations", "21", Patterns::Integer());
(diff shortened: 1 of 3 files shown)


include/deal.II/algorithms/newton.templates.h
```diff
@@ -65,6 +65,7 @@ namespace Algorithms
     control.parse_parameters(param);
     assemble_threshold    = param.get_double("Assemble threshold");
     n_stepsize_iterations = param.get_integer("Stepsize iterations");
+    debug                 = param.get_integer("Debug level");
```
bangerth: This might warrant a changelog entry.

bangerth:
Ah, we missed out on the changelog entry. Feel free to add it in a separate PR.

As for your question: I'm not sure about any plans to change to `ParameterAcceptor`. As for so many things, it takes volunteers. In this case, I'm not sure it's worth it, but maybe @luca-heltai has a different opinion.

PR #8672, opened 2019-08-30
Introduced DoFCellAccessor:get_future_fe().

marcfehling:
This function does exactly what `get_fe()` does for `active_fe_indices` and is thus basically a copy of it. This will be useful for an upcoming rework of the `p::d::CellWeights` class.

We don't check if an active/future index is valid until we actually access the corresponding finite element via the `dof_handler` object. Thus I removed the `try-catch` block in the corresponding test.

### DIFF
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -3974,6 +3974,7 @@ inline const FiniteElement<DoFHandlerType::dimension,
                            DoFHandlerType::space_dimension> &
 DoFCellAccessor<DoFHandlerType, level_dof_access>::get_fe() const
 {
+  Assert(this->dof_handler != nullptr, typename BaseClass::ExcInvalidObject());
   Assert(
     (dynamic_cast<const dealii::DoFHandler<DoFHandlerType::dimension,
                                            DoFHandlerType::space_dimension> *>(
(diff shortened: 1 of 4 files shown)


include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -4048,6 +4048,26 @@ DoFCellAccessor<DoFHandlerType, level_dof_access>::set_active_fe_index(
...
+{
+  Assert(
+    (dynamic_cast<const dealii::DoFHandler<DoFHandlerType::dimension,
+                                           DoFHandlerType::space_dimension> *>(
+       this->dof_handler) != nullptr) ||
+      (this->has_children() == false),
+    ExcMessage("In hp::DoFHandler objects, finite elements are only associated "
+               "with active cells. Consequently, you can not ask for the "
+               "future finite element on cells with children."));
+  Assert(this->dof_handler != nullptr, typename BaseClass::ExcInvalidObject());
```
bangerth: I think you ought to move this check above the previous one. You'd probably get funny error messages otherwise where the text doesn't match the expected condition.

marcfehling: Agree! We should do the same for the `get_fe` function, too.

bangerth:
Yes, thanks for making the change. Now go to bed! ;-)

Ready to merge!

Issue #8563, opened 2019-08-13
Use cell->face_iterators() in tutorials


bangerth:
The idiom we now use everywhere for cells is of the form
```
  for (const auto &cell : triangulation.active_cell_iterators())
```

It would be nice if we could also do this:
```
  for (const auto &cell : triangulation.active_cell_iterators())
    for (const auto &face : cell->face_iterators())
```
or similar. That should not be terribly hard to achieve, and would make a *lot* of code a lot simpler.

masterleinad:
It looks like you would need to collect all the global indices of the faces, create a container storing the corresponding face iterators and implement the LegacyForwardIterator interface for using it with `IteratorRange`.

peterrum:
I am currently working on a draft. What I was thinking about is to introduce a new Iterator, which stores a `CellAccessor`/`TriaCellIterator<TriaAccessor<dim, dim, spacedim>` and a counter [0,2*dim[. The new iterator could simply access the face (`TriaAccessor<dim - 1, dim, spacedim>`) by calling `cell->face(counter)`. 

This would be one very simple implementation. The version described by @masterleinad would be more flexible since it is not per-se connected to a single cell.

bangerth:
I had thought that implementing it via a cell iterator and an index is the easiest way, but it is true that if one just returned a `std::array<face_iterator,2*dim>`, then one doesn't even need to do anything else since that can already serve as the argument to a range-based for.

peterrum:
The version with `std::array` makes the implementation really short (19 added loc; see #8569)...

masterleinad:
Yes, no need to use a customer container if `std::array` does the job.

bangerth:
@peterrum fixes this in record time (thank you very much!).

We ought to also make the simplifying changes to the tutorials. (Like for the corresponding cell changes, the entire library is too large a code base to effectively change; that's ok -- but we should try to teach the current best practice in the tutorials.)

PR #8676, opened 2019-08-31
Permit reinitialization with iterators instead of (sub)face numbers.

drwells:
This is convenient when paired with the new cell->face_iterators() function.

### DIFF
--- a/include/deal.II/fe/fe_values.h
+++ b/include/deal.II/fe/fe_values.h
@@ -3758,6 +3758,18 @@ public:
          const unsigned int                                     face_no);
 
   /**
+   * Reinitialize the gradients, Jacobi determinants, etc for face @p face
+   * and cell @p cell.
+   *
+   * @note @p face must be one of @p cell's face iterators.
+   */
+  template <template <int, int> class DoFHandlerType, bool level_dof_access>
+  void
+  reinit(const TriaIterator<DoFCellAccessor<DoFHandlerType<dim, spacedim>,
+                                            level_dof_access>> &     cell,
+         const typename Triangulation<dim, spacedim>::face_iterator &face);
+
+  /**
    * Reinitialize the gradients, Jacobi determinants, etc for the given face
    * on a given cell of type "iterator into a Triangulation object", and the
    * given finite element. Since iterators into a triangulation alone only
@@ -3773,6 +3785,25 @@ public:
   void
   reinit(const typename Triangulation<dim, spacedim>::cell_iterator &cell,
          const unsigned int                                          face_no);
+
+  /*
+   * Reinitialize the gradients, Jacobi determinants, etc for the given face
+   * on a given cell of type "iterator into a Triangulation object", and the
+   * given finite element. Since iterators into a triangulation alone only
+   * convey information about the geometry of a cell, but not about degrees of
+   * freedom possibly associated with this cell, you will not be able to call
+   * some functions of this class if they need information about degrees of
+   * freedom. These functions are, above all, the
+   * <tt>get_function_value/gradients/hessians/third_derivatives</tt>
+   * functions. If you want to call these functions, you have to call the @p
+   * reinit variants that take iterators into DoFHandler or other DoF handler
+   * type objects.
+   *
+   * @note @p face must be one of @p cell's face iterators.
+   */
+  void
+  reinit(const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+         const typename Triangulation<dim, spacedim>::face_iterator &face);
 
   /**
    * Return a reference to this very object.
@@ -3878,6 +3909,17 @@ public:
          const unsigned int                                     subface_no);
 
   /**
+   * Alternative reinitialization function that takes, as arguments, iterators
+   * to the face and subface instead of their numbers.
+   */
+  template <template <int, int> class DoFHandlerType, bool level_dof_access>
+  void
+  reinit(const TriaIterator<DoFCellAccessor<DoFHandlerType<dim, spacedim>,
+                                            level_dof_access>> &     cell,
+         const typename Triangulation<dim, spacedim>::face_iterator &face,
+         const typename Triangulation<dim, spacedim>::face_iterator &subface);
+
+  /**
    * Reinitialize the gradients, Jacobi determinants, etc for the given
    * subface on a given cell of type "iterator into a Triangulation object", and
    * the given finite element. Since iterators into a triangulation alone only
@@ -3894,6 +3936,30 @@ public:
   reinit(const typename Triangulation<dim, spacedim>::cell_iterator &cell,
          const unsigned int                                          face_no,
          const unsigned int subface_no);
+
+  /**
+   * Reinitialize the gradients, Jacobi determinants, etc for the given
+   * subface on a given cell of type "iterator into a Triangulation object", and
+   * the given finite element. Since iterators into a triangulation alone only
+   * convey information about the geometry of a cell, but not about degrees of
+   * freedom possibly associated with this cell, you will not be able to call
+   * some functions of this class if they need information about degrees of
+   * freedom. These functions are, above all, the
+   * <tt>get_function_value/gradients/hessians/third_derivatives</tt>
+   * functions. If you want to call these functions, you have to call the @p
+   * reinit variants that take iterators into DoFHandler or other DoF handler
+   * type objects.
+   *
+   * This does the same thing as the previous function but takes iterators
+   * instead of numbers as arguments.
+   *
+   * @note @p face and @p subface must correspond to a face (and a subface of
+   * that face) of @p cell.
+   */
+  void
+  reinit(const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+         const typename Triangulation<dim, spacedim>::face_iterator &face,
+         const typename Triangulation<dim, spacedim>::face_iterator &subface);
 
   /**
    * Return a reference to this very object.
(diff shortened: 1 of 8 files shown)


drwells:
Part of #8563: a lot of the loops cannot be converted unless we provide this kind of reinitialization.

source/fe/fe_values.cc
```diff
@@ -5009,6 +5074,32 @@ FESubfaceValues<dim, spacedim>::reinit(
...
+template <int dim, int spacedim>
+void
+FESubfaceValues<dim, spacedim>::reinit(
+  const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+  const TriaIterator<TriaAccessor<dim - 1, dim, spacedim>> &  face,
+  const TriaIterator<TriaAccessor<dim - 1, dim, spacedim>> &  subface)
+{
+  const auto faces   = cell->face_iterators();
+  const auto face_it = std::find(faces.begin(), faces.end(), face);
+  Assert(face_it != faces.end(),
```
tjhei: You are repeating these 3+ lines of code many times. Would it make sense to add a function that does this? Maybe ``cell->face_iterator_to_index(face)`` or ``face->my_child_index()`` or something? 

I am also not sure if we really need a linear search here, is there no identifying info inside the iterator we can use?

drwells: Good idea: I should add that (I will repeat this search even more times when I write the `hp` version of this PR).

There is no such identifying information in the iterator since it could belong to two different cells.

include/deal.II/fe/fe_values.h
```diff
@@ -3774,6 +3784,14 @@ class FEFaceValues : public FEFaceValuesBase<dim, spacedim>
   reinit(const typename Triangulation<dim, spacedim>::cell_iterator &cell,
          const unsigned int                                          face_no);
 
+  /*
+   * Alternative reinitialization function that also only provides information
+   * that does not depend on the global DoF numbering.
```
bangerth: I don't think this sentence is understandable if you don't already know what this function does. Instead of writing it in the negative (what it doesn't provide), I'd say that this function reinitializes the current object using a reference to a cell and face; "given that the cell iterator only points into a triangulation, the same restrictions apply as for the previous function".

include/deal.II/fe/fe_values.h
```diff
@@ -3774,6 +3784,14 @@ class FEFaceValues : public FEFaceValuesBase<dim, spacedim>
...
+   */
+  void
+  reinit(const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+         const TriaIterator<TriaAccessor<dim - 1, dim, spacedim>> &  face);
```
bangerth: Why not use a `typename Triangulation<dim, spacedim>::face_iterator` here?

include/deal.II/fe/fe_values.h
```diff
@@ -3757,6 +3757,16 @@ class FEFaceValues : public FEFaceValuesBase<dim, spacedim>
...
+   * current face instead of its number.
+   */
+  template <template <int, int> class DoFHandlerType, bool level_dof_access>
+  void
+  reinit(const TriaIterator<DoFCellAccessor<DoFHandlerType<dim, spacedim>,
+                                            level_dof_access>> &   cell,
+         const TriaIterator<TriaAccessor<dim - 1, dim, spacedim>> &face);
+
   /**
    * Reinitialize the gradients, Jacobi determinants, etc for the given face
```
bangerth: While you're here anyway, would you indulge me in making the following changes:
```suggestion
   * Reinitialize the gradients, Jacobi determinants, etc., for the given face
```

include/deal.II/fe/fe_values.h
```diff
@@ -3757,6 +3757,16 @@ class FEFaceValues : public FEFaceValuesBase<dim, spacedim>
...
    * on given cell of type "iterator into a Triangulation object", and the
```
bangerth:
```suggestion
   * on the given cell of type "iterator into a Triangulation object", and the
```

include/deal.II/fe/fe_values.h
```diff
@@ -3757,6 +3757,16 @@ class FEFaceValues : public FEFaceValuesBase<dim, spacedim>
...
          const unsigned int                                     face_no);
 
+  /**
+   * Alternative reinitialization function that takes an iterator to the
+   * current face instead of its number.
+   */
+  template <template <int, int> class DoFHandlerType, bool level_dof_access>
+  void
+  reinit(const TriaIterator<DoFCellAccessor<DoFHandlerType<dim, spacedim>,
+                                            level_dof_access>> &   cell,
```
bangerth: Why not use a `typename Triangulation<...>::cell_iterator` here, and a `face_iterator` for the next argument?

include/deal.II/fe/fe_values.h
```diff
@@ -3877,6 +3895,17 @@ class FESubfaceValues : public FEFaceValuesBase<dim, spacedim>
...
          const unsigned int                                     subface_no);
 
+  /**
+   * Alternative reinitialization function that takes, as arguments, iterators
+   * to the face and subface instead of their numbers.
+   */
+  template <template <int, int> class DoFHandlerType, bool level_dof_access>
+  void
+  reinit(const TriaIterator<DoFCellAccessor<DoFHandlerType<dim, spacedim>,
+                                            level_dof_access>> &   cell,
```
bangerth: same questions about argument types here.

include/deal.II/fe/fe_values.h
```diff
@@ -3877,6 +3895,17 @@ class FESubfaceValues : public FEFaceValuesBase<dim, spacedim>
          const unsigned int                                     face_no,
          const unsigned int                                     subface_no);
 
+  /**
+   * Alternative reinitialization function that takes, as arguments, iterators
+   * to the face and subface instead of their numbers.
```
bangerth: It might be useful to point out that `subface` must be one of the `face->child(...)` iterators and that it must be the face iterator for a neighboring cell.

include/deal.II/fe/fe_values.h
```diff
@@ -3895,6 +3924,15 @@ class FESubfaceValues : public FEFaceValuesBase<dim, spacedim>
          const unsigned int                                          face_no,
          const unsigned int subface_no);
 
+  /**
+   * Alternative reinitialization function that also only provides information
+   * that does not depend on the global DoF numbering.
```
bangerth: same as above

source/fe/fe_values.cc
```diff
@@ -4788,6 +4788,23 @@ FEFaceValues<dim, spacedim>::reinit(
...
+FEFaceValues<dim, spacedim>::reinit(
+  const TriaIterator<DoFCellAccessor<DoFHandlerType<dim, spacedim>, lda>> &cell,
+  const TriaIterator<TriaAccessor<dim - 1, dim, spacedim>> &               face)
+{
+  const auto faces   = cell->face_iterators();
+  const auto face_it = std::find(faces.begin(), faces.end(), face);
+  Assert(face_it != faces.end(),
+         ExcMessage("The provided face iterator does not correspond to a face "
+                    "owned by the provided cell."));
+  reinit(cell, face_it - faces.begin());
```
bangerth: This requires more work than I think is necessary as it builds all N face iterators even though on average you'll only need to check N/2. It would be easier if you just wrote
```
  for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
    if (cell->face(f) == face)
      {
         reinit (cell, f);
         return;
      }

  Assert (false, ...);
```
Same for the other functions below.

source/fe/fe_values.cc
```diff
@@ -4984,6 +5017,38 @@ FESubfaceValues<dim, spacedim>::reinit(
...
+         ExcMessage("The face used in FESubfaceValues should have children."));
+
+  unsigned int subface_n     = 0;
+  bool         found_subface = false;
+  (void)found_subface;
+  for (; subface_n < face->n_children(); ++subface_n)
+    if (face->child(subface_n) == subface)
+      {
+        found_subface = true;
+        break;
```
bangerth: Just call the `reinit()` function here and then `return`.

drwells:
@bangerth Thanks for the thorough read-through. I'll try to get this addressed later this week.

bangerth:
@drwells -- ping?

drwells:
I let this get away from me. I'll try to find time to finish it this week.

include/deal.II/fe/fe_values.h
drwells: I don't think the first one will work since we do need a `DoFCellAccessor` here in the subsequent `reinit` call. We can make the change for the face iterator, though.

bangerth: I misread. I really meant `DoFHandler<...>::cell_accessor. Why would that not work here?

bangerth:
@drwells ?

include/deal.II/fe/fe_values.h
drwells: I don't think that it would work since we have a boolean for `level_dof_access` in this function: this is, I think, always false, for `DoFHandler::cell_accessor`.

source/fe/fe_values.cc
drwells: I added two such index functions in f791b78827fa0444093b31dcb2f48ba6e1a32d82.

drwells:
It took awhile to address everything but I think I have now resolved all review comments.

PR #8509, opened 2019-08-08
hyper_L documentation fix

LiMaIG:

PR #8677, opened 2019-09-01
Add another .mailmap entry

masterleinad:
@Rombur This is what needs to be done for commits without a proper author name.


tjhei:
thanks! I will take a look while this wasn't a problem in #8509

tjhei:
oh, got it. The PR was based on an older commit before we added the new indent check. It looks like travis checks the tip of the branch and not the merged result...

PR #8679, opened 2019-09-01
Update PolynomialsBDM degree and pass the degree to the base class constructor

### LINKED ISSUE #8591: Update PolynomialsBDM.degree()
While making changes under #8588, I noticed that `PolynomialsBDM` defines its `degree()` function to return `polynomial_space.degree() - 1`.

The goal is to assign a degree to the base class during construction before the `polynomial_space` is constructed.

I imagine the best way to tackle this is to write down what `polynomial_space` is by hand and see if there is a formula for the polynomial degree in terms of the argument for the constructor, `k`.

GrahamBenHarper:
Based on the fact that P_k is a subset of BDM_k which is a subset of P_{k+1}, the polynomial degree of this class is k+1.

I've changed the base class constructor accordingly, but it might be a good idea to run the testers just in case. This resolves #8591

### DIFF
--- a/include/deal.II/base/polynomials_bdm.h
+++ b/include/deal.II/base/polynomials_bdm.h
@@ -132,13 +132,6 @@ public:
            std::vector<Tensor<5, dim>> &fourth_derivatives) const override;
 
   /**
-   * Return the degree of the BDM space, which is one less than the highest
-   * polynomial degree.
-   */
-  unsigned int
-  degree() const;
-
-  /**
    * Return the name of the space, which is <tt>BDM</tt>.
    */
   std::string
@@ -201,14 +194,6 @@ private:
    */
   mutable std::vector<Tensor<4, dim>> p_fourth_derivatives;
 };
-
-
-template <int dim>
-inline unsigned int
-PolynomialsBDM<dim>::degree() const
-{
-  return polynomial_space.degree() - 1;
-}
 
 
 template <int dim>
--- a/source/base/polynomials_bdm.cc
+++ b/source/base/polynomials_bdm.cc
@@ -28,7 +28,7 @@ DEAL_II_NAMESPACE_OPEN
 
 template <int dim>
 PolynomialsBDM<dim>::PolynomialsBDM(const unsigned int k)
-  : TensorPolynomialsBase<dim>(k, n_polynomials(k))
+  : TensorPolynomialsBase<dim>(k + 1, n_polynomials(k))
   , polynomial_space(Polynomials::Legendre::generate_complete_basis(k))
   , monomials((dim == 2) ? (1) : (k + 2))
   , p_values(polynomial_space.n())
@@ -193,7 +193,8 @@ PolynomialsBDM<dim>::evaluate(
               // p(t) = t^(i+1)
               monomials[i + 1].value(unit_point(d), monovali[d]);
               // q(t) = t^(k-i)
-              monomials[degree() - i].value(unit_point(d), monovalk[d]);
+              monomials[this->degree() - 1 - i].value(unit_point(d),
+                                                      monovalk[d]);
             }
           if (values.size() != 0)
             {
--- a/doc/news/changes/minor/20190902Harper
+++ b/doc/news/changes/minor/20190902Harper
@@ -1,0 +1,4 @@
+Bugfix: In PolynomialsBDM, the degree() function is corrected to return a 
+polynomial degree of k+1, instead of its previous value of k.
+<br>
+(Graham Harper, 2019/09/02)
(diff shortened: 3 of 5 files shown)


bangerth:
Was this a bug? I can't quite follow the changes -- they are incompatible, but I can't tell whether the code was wrong before or whether it just returned something that is semantically different.

In any case, every incompatible change makes for a useful target for a testcase because clearly something was wrong before and we should make sure it remains correct after.

/rebuild

masterleinad:
Please rebase to make the indentation script happy and run the indentation script.

GrahamBenHarper:
The changes addressing #1973 defined `TensorPolynomialsBase::degree()` to return the highest order polynomial degree of the space. This was not totally consistent across the polynomial classes.

For example, the BDM_k element previously returned `polynomial_space.degree() - 1`. `polynomial_space.degree()` does not return the polynomial degree, but rather the number of polynomials in the space. BDM_k fills `polynomial_space` with a complete degree k Legendre basis, which contains k+1 polynomials. This was an extremely roundabout way of returning just `k`, which isn't quite correct by our definition. We instead want it to be `k+1` for the highest polynomial degree.

As for the indent script, I'm seeing an odd error when I attempt to run it...
```
  invalid author 'LiMaIG' without firstname and lastname
```
Did I never assign my name on my Github? What's the command again?

GrahamBenHarper:
No, my name is correct... The indent script seems to have gone rogue?

GrahamBenHarper:
I just checked and #8509 seems to be creating the same problems for me that @Rombur commented about yesterday. Is there a workaround for this?

GrahamBenHarper:
Oh oh oh oh I was 4 commits behind master, and one of them happened to be #8677. Does this work?

masterleinad:
```
1016 - base/polynomials_bdm_01.debug (Failed)
1117 - base/polynomials_bdm_02.debug (Failed)
```
I guess these test failures are expected but please check yourself.

GrahamBenHarper:
Okay yes the two tests used `poly.degree()` to define the quadrature and outputs. Should I make an incompatibilities change log entry about this?

bangerth:
If I read the description above right, you've essentially just fixed a bug and not introduced incompatible behavior (e.g., changing the semantics of a function that returned the then-correct behavior before and the now-correct but different behavior now). In that case, it's a judgement call where you want to put the changelog entry. I'd probably put it into "minor", given that few people will notice the issue.

source/base/polynomials_bdm.cc
```diff
@@ -28,7 +28,7 @@ DEAL_II_NAMESPACE_OPEN
 
 template <int dim>
 PolynomialsBDM<dim>::PolynomialsBDM(const unsigned int k)
-  : TensorPolynomialsBase<dim>(k, n_polynomials(k))
+  : TensorPolynomialsBase<dim>(k + 1, n_polynomials(k))
```
bangerth: So just to be clear, this was the bug you're fixing, right?

include/deal.II/base/polynomials_bdm.h
```diff
@@ -203,14 +196,6 @@ class PolynomialsBDM : public TensorPolynomialsBase<dim>
 };
 
 
-template <int dim>
-inline unsigned int
-PolynomialsBDM<dim>::degree() const
-{
-  return polynomial_space.degree() - 1;
```
bangerth: And the removal of this function is essentially just a clean-up to make sure the class doesn't have a function that was awkwardly named? (Because it confuses the "polynomial degree" with the "degree of the element".)

This makes sense to me. It's unfortunate that we say things like "the RT element of degree 0" (or the "RT(0) element") as that has polynomial degree 1.

bangerth:
OK to merge if you fix the indentation issues.

source/base/polynomials_bdm.cc
GrahamBenHarper: Yes this is the bug I'm fixing. This is because in the base class, we have
```
  /**
   * The highest polynomial degree of this functions represented by this object.
   */
  const unsigned int polynomial_degree;
```

include/deal.II/base/polynomials_bdm.h
GrahamBenHarper: Yes and I didn't clean it up initially because I was uncertain what this `PolynomialsBDM<dim>::degree()` function did. `polynomial_space.degree() - 1` did not make sense until I took a few minutes to read how `polynomial_space` was constructed and how `degree()` was defined for that a `polynomial_space` object.

It is a little unfortunate. Even the BR1 element is a polynomial degree 2 space, but I believe this is okay because the finite element will always keep track of the space degree `k`, even if the polynomial space has a different degree like `k+1`.

GrahamBenHarper:
At first, I was going to avoid a changelog entry; however, I want anybody who is using the PolynomialsBDM class to hopefully see this change. I added a changelog entry for this, labeled as a bugfix.

PR #8680, opened 2019-09-02
Fix a bug with cuda and hanging nodes

Rombur:
Right now we can't use cuda matrix-free on an adapted mesh if `dim=2` and `fe_degree` is even. There is also a test failing on the cuda tester when using atomics in release mode. Going back to Karl's code, I realized that I was missing some synchronization barriers.

### DIFF
--- a/include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
+++ b/include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
@@ -177,11 +177,6 @@ namespace CUDAWrappers
       , fe_degree(fe_degree)
       , dof_handler(dof_handler)
     {
-      AssertThrow(
-        (dim == 3) || ((fe_degree % 2) == 1),
-        ExcMessage(
-          "This function is not implemented when dim = 2 and fe_degree is even."));
-
       // Set up line-to-cell mapping for edge constraints (only if dim = 3)
       setup_line_to_cell();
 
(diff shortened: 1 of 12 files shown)


drwells:
Nice job. I am glad that we are finally rid of this weird bug.

Is there a simple explanation for why these barriers are necessary for this one case and apparently not necessary for the others?

include/deal.II/matrix_free/cuda_hanging_nodes_internal.h
```diff
@@ -787,6 +782,7 @@ namespace CUDAWrappers
           ((direction == 1) && ((constraint_mask & internal::constr_type_x) ?
                                   (x_idx == 0) :
                                   (x_idx == fe_degree)));
+        __syncthreads();
```
masterleinad: I have to admit that it is not clear to me what could happen between this `__syncthreads` and the next one. It seems that we are only reading values into local variables... Is it possible that the synchronization problem happens before this function is called?

kronbichler: I have not yet checked in detail what happens before, but couldn't the boolean operations (masking) that enable/disable certain lanes lead to some clash that manifests for some sizes, given that we have different threads (lanes) on different local dofs? That would at least be my interpretation, but then it has been a while since I last looked into how these boolean flags are translated into masked operations for the various lanes.

Rombur: @masterleinad Yes it is possible. Honestly I use Karl implementation as a reference and I was missing these `__synchtreads`. Adding them fix the problem but I am not sure why.

masterleinad:
I thoroughly verified that this works as expected, but I think we can reorganize the synchronization a little bit better. Let me do that in a separate pull request.

PR #8681, opened 2019-09-02
Rename variable

peterrum:
ping @kronbichler 

The parameter name `fe_component` confused me. Normally, we use `dof_handler_index` or short `dof_index`. I associate `fe_component` with a component within `FESystem`...

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -1307,7 +1307,7 @@ public:
   typename DoFHandler<dim>::cell_iterator
   get_cell_iterator(const unsigned int macro_cell_number,
                     const unsigned int vector_number,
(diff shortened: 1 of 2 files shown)


kronbichler:
Can you please do the same in
https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/matrix_free.h#L1395
and update the name also here:
https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/matrix_free.templates.h#L159
Thanks!

peterrum:
Done!

I keep this instance of `dof_index` since it is not related to the identification of the `DoFHandler`:
https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/matrix_free.templates.h#L1046-L1047

PR #8687, opened 2019-09-03
Refactored parallel::CellWeights.

marcfehling:
We generalized the interface of the `parallel::CellWeights` class in such a way that we introduced static member functions. Those let the user manage connecting and disconnecting callback functions manually if desired without instantiating an object of this class.

The legacy interface has been deprecated. Corresponding tests have been updated, revealing a bug in `hp::DoFHandler`: All its internal data structures need to be cleared when calling `initialize()`.

Part of #7515.

@bangerth -- FYI

### DIFF
--- a/doc/news/changes/incompatibilities/20190903Fehling
+++ b/doc/news/changes/incompatibilities/20190903Fehling
@@ -1,0 +1,6 @@
+Changed: The interface to the parallel::CellWeights class was generalized
+by introducing static member functions. This way, users are capable of
+connecting and disconnecting callback functions manually if desired without
+instantiating an object of this class. The legacy interface has been deprecated.
+<br>
+(Marc Fehling, 2019/09/03)
--- a/include/deal.II/distributed/cell_weights.h
+++ b/include/deal.II/distributed/cell_weights.h
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 2018 by the deal.II authors
+// Copyright (C) 2018 - 2019 by the deal.II authors
 //
 // This file is part of the deal.II library.
 //
@@ -38,80 +38,180 @@ namespace parallel
    * This class allows computing these weights for load balancing by
    * consulting the FiniteElement that is associated with each cell of
    * a hp::DoFHandler. One can choose from predefined weighting
-   * algorithms provided by this class or provide a custom one. The
-   * chosen weighting function will be connected to the corresponding
-   * signal of the linked parallel::TriangulationBase via callback.
+   * algorithms provided by this class or provide a custom one.
    *
-   * An object of this class needs to exist for every DoFHandler associated
-   * with the Triangulation we work on to achieve satisfying work balancing
-   * results.
+   * This class offers two different ways of connecting the chosen weighting
+   * function to the corresponding signal of the linked
+   * parallel::TriangulationBase. The recommended way involves creating an
+   * object of this class which will automatically take care of registering the
+   * weighting function upon creation and de-registering it once destroyed. An
+   * object of this class needs to exist for every DoFHandler associated with
+   * the Triangulation we work on to achieve satisfying work balancing results.
+   * The connected weighting function may be changed anytime using the
+   * CellWeights::reinit() function. The following code snippet demonstrates how
+   * to achieve each cell being weighted by its current number of degrees of
+   * freedom. We chose a factor of `1000` that corresponds to the initial weight
+   * each cell is assigned to upon creation.
+   * @code
+   * parallel::CellWeights<dim, spacedim> cell_weights(
+   *   hp_dof_handler,
+   *   parallel::CellWeights<dim, spacedim>::ndofs_weighting({1000, 1}));
+   * @endcode
    *
-   * A small code snippet follows explaining how to achieve each cell
-   * being weighted by its current number of degrees of freedom.
+   * On the other hand, you are also able to take care of handling the signal
+   * connection manually by using the static member function of this class. In
+   * this case, an analogous code example looks as follows.
    * @code
-   * parallel::CellWeights<dim, spacedim> cell_weights(hp_dof_handler);
-   * cell_weights.register_ndofs_weighting();
+   * boost::signals2::connection connection =
+   *   hp_dof_handler.get_triangulation().signals.cell_weight.connect(
+   *     parallel::CellWeights<dim, spacedim>::make_weighting_callback(
+   *       hp_dof_handler,
+   *       parallel::CellWeights<dim, spacedim>::ndofs_weighting(
+   *         {1000, 1}));
    * @endcode
-   * The weighting function can be changed anytime. Even more ambitious
-   * approaches are possible by submitting customized functions, e.g.
-   * @code
-   * cell_weights.register_custom_weighting(
-   *   [](const FiniteElement<dim, spacedim> &active_fe,
-   *      const typename hp::DoFHandler<dim, spacedim>::cell_iterator &)
-   *   -> unsigned int {
-   *   return 1000 * std::pow(active_fe.dofs_per_cell, 1.6);
-   * });
-   * @endcode
-   * The returned value has to be an unsigned integer and is thus limited by
-   * some large number. It is interpreted as the additional computational load
-   * of each cell. See Triangulation::Signals::cell_weight for a discussion on
-   * this topic.
+   *
+   * @note See Triangulation::Signals::cell_weight for more information on
+   * weighting and load balancing.
    *
    * @note Be aware that this class connects the weight function to the
-   * Triangulation during its construction. If the Triangulation
+   * Triangulation during this class's constructor. If the Triangulation
    * associated with the DoFHandler changes during the lifetime of the
-   * latter, an assertion will be triggered in the weight_callback() function.
-   * It is recommended to create a separate object in this case and to destroy
-   * the previous one.
+   * latter via hp::DoFHandler::initialize(), an assertion will be triggered in
+   * the weight_callback() function. Use CellWeights::reinit() to deregister the
+   * weighting function on the old Triangulation and connect it to the new one.
    *
    * @ingroup distributed
-   * @author Marc Fehling, 2018
+   * @author Marc Fehling, 2018, 2019
    */
   template <int dim, int spacedim = dim>
   class CellWeights
   {
   public:
     /**
+     * An alias that defines the characteristics of a function that can be used
+     * for weighting cells during load balancing.
+     *
+     * Such weighting functions take as arguments an iterator to a cell and the
+     * future finite element that will be assigned to it after repartitioning.
+     * They return an unsigned integer which is interpreted as the cell's
+     * weight or, in other words, the additional computational load associated
+     * with it.
+     */
+    using WeightingFunction = std::function<unsigned int(
+      const typename hp::DoFHandler<dim, spacedim>::cell_iterator &,
+      const FiniteElement<dim, spacedim> &)>;
+
+    /**
+     * Constructor.
+     *
+     * @param[in] dof_handler The hp::DoFHandler which will be used to
+     *    determine each cell's finite element.
+     * @param[in] weighting_function The function that determines each
+     *    cell's weight during load balancing.
+     */
+    CellWeights(const dealii::hp::DoFHandler<dim, spacedim> &dof_handler,
+                const WeightingFunction &weighting_function);
+
+    /**
+     * Destructor.
+     *
+     * Disconnects the function previously connected to the weighting signal.
+     */
+    ~CellWeights();
+
+    /**
+     * Connect a different @p weighting_function to the Triangulation
+     * associated with the @p dof_handler.
+     *
+     * Disconnects the function previously connected to the weighting signal.
+     */
+    void
+    reinit(const hp::DoFHandler<dim, spacedim> &dof_handler,
+           const WeightingFunction &            weighting_function);
+
+    /**
+     * Converts a @p weighting_function to a different type that qualifies as
+     * a callback function, which can be connected to a weighting signal of a
+     * Triangulation.
+     *
+     * This function does <b>not</b> connect the converted function to the
+     * Triangulation associated with the @p dof_handler.
+     */
+    static std::function<unsigned int(
+      const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+      const typename Triangulation<dim, spacedim>::CellStatus     status)>
+    make_weighting_callback(const hp::DoFHandler<dim, spacedim> &dof_handler,
+                            const WeightingFunction &weighting_function);
+
+    /**
+     * @name Selection of weighting functions
+     * @{
+     */
+
+    /**
+     * Choose a constant weight @p factor on each cell.
+     */
+    static WeightingFunction
+    constant_weighting(const unsigned int factor = 1000);
+
+    /**
+     * The pair of floating point numbers $(a,b)$ provided via
+     * @p coefficients determines the weight $w_K$ of each cell $K$ with
+     * $n_K$ degrees of freedom in the following way: \f[ w_K =
+     * a \, n_K^b \f]
+     *
+     * The right hand side will be rounded to the nearest integer since cell
+     * weights are required to be integers.
+     */
+    static WeightingFunction
+    ndofs_weighting(const std::pair<float, float> &coefficients);
+
+    /**
+     * The container @p coefficients provides pairs of floating point numbers
+     * $(a_i, b_i)$ that determine the weight $w_K$ of each cell
+     * $K$ with $n_K$ degrees of freedom in the following way: \f[ w_K =
+     * \sum_i a_i \, n_K^{b_i} \f]
+     *
+     * The right hand side will be rounded to the nearest integer since cell
+     * weights are required to be integers.
+     */
+    static WeightingFunction
+    ndofs_weighting(const std::vector<std::pair<float, float>> &coefficients);
+
+    /**
+     * @}
+     */
+
+    /**
+     * @name Deprecated functions
+     * @{
+     */
+
+    /**
      * Constructor.
      *
      * @param[in] dof_handler The hp::DoFHandler which will be used to
      *    determine each cell's finite element.
      */
+    DEAL_II_DEPRECATED
     CellWeights(const dealii::hp::DoFHandler<dim, spacedim> &dof_handler);
 
     /**
-     * Destructor.
+     * @copydoc CellWeights::constant_weighting()
      */
-    ~CellWeights();
-
-    /**
-     * Choose a constant weight @p factor on each cell.
-     */
-    void
+    DEAL_II_DEPRECATED void
     register_constant_weighting(const unsigned int factor = 1000);
 
     /**
-     * Choose a weight for each cell that is proportional to its number of
-     * degrees of freedom with a factor @p factor.
+     * @copydoc CellWeights::ndofs_weighting()
      */
-    void
+    DEAL_II_DEPRECATED void
     register_ndofs_weighting(const unsigned int factor = 1000);
 
     /**
-     * Choose a weight for each cell that is proportional to its number of
-     * degrees of freedom <i>squared</i> with a factor @p factor.
+     * @copydoc CellWeights::ndofs_squared_weighting()
      */
-    void
+    DEAL_II_DEPRECATED void
     register_ndofs_squared_weighting(const unsigned int factor = 1000);
 
     /**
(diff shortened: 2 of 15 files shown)


marcfehling:
When refactoring this class, we actually didn't use the feature introduced in #8672. It may be useful in the future, but is unused now.

Shall we leave it in or remove it?

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+   * boost::signals2::connection listener =
+   *   dof_handler.get_triangulation().signals.cell_weight.connect(
+   *     parallel::CellWeights<dim, spacedim>::make_weighting_callback(
+   *       dof_handler,
+   *       parallel::CellWeights<dim, spacedim>::ndofs_weighting());
    * @endcode
-   * The returned value has to be an unsigned integer and is thus limited by
-   * some large number. It is interpreted as the additional computational load
-   * of each cell. See Triangulation::Signals::cell_weight for a discussion on
-   * this topic.
```
bangerth: Would it be useful to preserve these four lines somehow, or at least link to a place where it is explained how the weighting works?

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+   *
+   * @note See Triangulation::Signals::cell_weight for more information.
    *
    * @note Be aware that this class connects the weight function to the
    * Triangulation during its construction. If the Triangulation
    * associated with the DoFHandler changes during the lifetime of the
    * latter, an assertion will be triggered in the weight_callback() function.
-   * It is recommended to create a separate object in this case and to destroy
-   * the previous one.
+   * Use CellWeights::reinit() to update dependencies.
```
bangerth: It's not clear to me what "dependencies" you refer to. Can you elaborate?

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
-   * @author Marc Fehling, 2018
+   * @author Marc Fehling, 2018, 2019
    */
   template <int dim, int spacedim = dim>
   class CellWeights
   {
   public:
+    /**
+     * An alias that defines the characteristics of a function that can be used
+     * for weighting cells during load balancing.
```
bangerth:
```suggestion
     * for weighting cells during load balancing. Such weighting functions
     * take as arguments a finite element and an iterator to a cell
     * on which this element is to be used. They return an
     * unsigned integer (the weight).
```

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+     */
+    using weighting_function = std::function<unsigned int(
```
bangerth: Can we call it `WeightingFunction` to stick with typical `CameCase` for types?

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+      const FiniteElement<dim, spacedim> &,
+      const typename hp::DoFHandler<dim, spacedim>::cell_iterator &)>;
```
bangerth: I suspect that the fe is to be used on the cell; so the fe is *subordinate* to the cell. This would suggest that one should reverse the order of arguments, unless that leads to an incompatible change.

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+    ndofs_squared_weighting(const unsigned int factor = 1000);
+
+    /**
+     * The pair of floating point numbers \f$(a,b)\f$ provided via
+     * @p coefficients determines the weight \f$w_K\f$ of each cell \f$K\f$ with
+     * \f$n_K\f$ degrees of freedom in the following way: \f[ w_K =
+     * \mathrm{truncate}\left( a \, n_K^b \right) \f]
+     */
+    static weighting_function
+    ndofs_exponential_weighting(std::pair<double, double> coefficients);
```
bangerth:
```suggestion
    ndofs_exponential_weighting(const std::pair<double, double> &coefficients);
```

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+     * Triangulation.
+     *
+     * This function does <b>not</b> connect the converted function to the
+     * Triangulation associated with the @p dof_handler.
+     */
+    static std::function<unsigned int(
+      const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+      const typename Triangulation<dim, spacedim>::CellStatus     status)>
+    make_weighting_callback(const hp::DoFHandler<dim, spacedim> &dof_handler,
+                            const weighting_function weighting_function);
```
bangerth:
```suggestion
                            const weighting_function &weighting_function);
```
Same elsewhere

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+
+    /**
+     * The container @p coefficients provides pairs of floating point numbers
+     * \f$(a_i, b_i)\f$ that determine the weight \f$w_K\f$ of each cell
+     * \f$K\f$ with \f$n_K\f$ degrees of freedom in the following way: \f[ w_K =
+     * \mathrm{truncate}\left( \sum_i a_i \, n_K^{b_i} \right) \f]
+     */
+    static weighting_function
+    ndofs_exponential_weighting(
+      std::vector<std::pair<double, double>> coefficients);
```
bangerth:
```suggestion
      const std::vector<std::pair<double, double>> &coefficients);
```

include/deal.II/distributed/cell_weights.h
```diff
@@ -147,40 +255,33 @@ namespace parallel
...
-    unsigned int
-    weight_callback(
+    static unsigned int
+    weighting_callback(
       const typename Triangulation<dim, spacedim>::cell_iterator &cell,
-      const typename Triangulation<dim, spacedim>::CellStatus     status);
+      const typename Triangulation<dim, spacedim>::CellStatus     status,
+      const hp::DoFHandler<dim, spacedim> &                       dof_handler,
+      const parallel::TriangulationBase<dim, spacedim> &          triangulation,
+      const weighting_function weighting_function);
```
bangerth:
```suggestion
      const weighting_function &weighting_function);
```

bangerth:
On 9/3/19 6:28 AM, marcfehling wrote:
> Shall we leave it in or remove it?

Leave it.

include/deal.II/distributed/cell_weights.h
marcfehling: Are you okay with moving the explanation to the `WeightingFunction` alias (as you suggested below) and leaving a `@note` referring to `Triangulation::Signals::cell_weight` in the general class documentation?

include/deal.II/distributed/cell_weights.h
marcfehling: Agreed! The `FiniteElement` is the future finite element that will be associated to the cell after repartitioning.

In fact, we never use the `cell_iterator` in all of these weighting functions, but I wanted to leave it in case someone comes up with some fancy `weighting_function` on his own.

marcfehling:
Would it make sense to also consider `hp::QCollection` objects in the weighting process?

This would involve major changes to the interface of this class, and there may be no demand for this feature at all. I would leave it out for now and probably come back to that idea once someone requests it.

bangerth:
Yes, one thing at a time!

marcfehling:
I would like to move the static member functions from the `CellWeights` class into a separate namespace `CellWeightingFunctions`. The current interface looks rather ugly with these `template` specifiers and I think this would clean up things at least a bit:
```
parallel::CellWeights<dim, spacedim> cell_weights(
  hp_dof_handler,
  parallel::CellWeights<dim, spacedim>::template ndofs_weighting<
    unsigned int, unsigned int>({1000, 1}));
```

marcfehling:
Further: Would it make sense to provide functors instead of function pointers here? This should give us a performance boost.

marcfehling:
Some remarks on my previous comments:
- Introducing a new namespace does not yield to any benefit: We still need the `dim` and `spacedim` templates. We may rely on class template argument deduction introduced in `C++17` to clean out these lines :)
- Functors would work with `boost::signals`, but I think I'm over-engineering this. I'd be fine to stick to the function pointers for now.

doc/news/changes/incompatibilities/20190903Fehling
```diff
@@ -0,0 +1,3 @@
+Changed: The interface to the parallel::CellWeights class was changed.
```
bangerth: It might be useful to be a little bit more explicit.

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+   * parallel::TriangulationBase. The recommended way involves creating an
+   * object of this class which will automatically take care of registering the
+   * weighting function upon creation and de-registering it once destroyed. An
+   * object of this class needs to exist for every DoFHandler associated with
+   * the Triangulation we work on to achieve satisfying work balancing results.
+   * The connected weighting function may be changed anytime using the
+   * CellWeights::reinit() function. The following code snippet demonstrates how
+   * to achieve each cell being weighted by its current number of degrees of
+   * freedom. We chose a factor of `1000` that corresponds to the initial weight
+   * each cell is assigned to upon creation.
```
bangerth: Don't we mention this factor of 1000 somewhere in the documentation of the triangulation? I think it would be good to reference that place here to explain that factor of 1000.

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
    * @code
-   * parallel::CellWeights<dim, spacedim> cell_weights(hp_dof_handler);
-   * cell_weights.register_ndofs_weighting();
+   * parallel::CellWeights<dim, spacedim> cell_weights(
```
bangerth: could it be `const`?

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+   * connection manually by using the static member function of this class. In
+   * this case, an analogous code example looks as follows.
    * @code
-   * cell_weights.register_custom_weighting(
-   *   [](const FiniteElement<dim, spacedim> &active_fe,
-   *      const typename hp::DoFHandler<dim, spacedim>::cell_iterator &)
-   *   -> unsigned int {
-   *   return 1000 * std::pow(active_fe.dofs_per_cell, 1.6);
-   * });
+   * boost::signals2::connection listener =
```
bangerth:
```suggestion
   * boost::signals2::connection connection =
```

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
-   * The returned value has to be an unsigned integer and is thus limited by
-   * some large number. It is interpreted as the additional computational load
-   * of each cell. See Triangulation::Signals::cell_weight for a discussion on
-   * this topic.
+   *
+   * @note See Triangulation::Signals::cell_weight for more information on
+   * weighting and load balancing.
    *
    * @note Be aware that this class connects the weight function to the
    * Triangulation during its construction. If the Triangulation
```
bangerth:
```suggestion
   * Triangulation during this class's constructor. If the Triangulation
```

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+     */
+
+    /**
+     * Choose a constant weight @p factor on each cell.
+     */
+    static WeightingFunction
+    constant_weighting(const unsigned int factor = 1000);
+
+    /**
+     * The pair of floating point numbers \f$(a,b)\f$ provided via
```
bangerth: Here and below, remove the `\f` -- we have a script that allows us to write just `$...$`.
```suggestion
     * The pair of floating point numbers $(a,b)$ provided via
```

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+     * @p coefficients determines the weight \f$w_K\f$ of each cell \f$K\f$ with
+     * \f$n_K\f$ degrees of freedom in the following way: \f[ w_K =
+     * \mathrm{truncate}\left( a \, n_K^b \right) \f]
```
bangerth: can you define what the truncate function actually does? I think you just want to omit that term in the formula and instead simply say "where the right hand side is rounded to the nearest integer since cell weights must be integers".

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+     */
+    template <typename FactorType, typename ExponentType>
+    static WeightingFunction
+    ndofs_weighting(const std::pair<FactorType, ExponentType> &coefficients);
```
bangerth: I looked at the code example above, and it's awkward to read because of the template arguments. Why not just let both of these be floating point numbers?

include/deal.II/distributed/cell_weights.h
```diff
@@ -38,80 +38,178 @@ namespace parallel
...
+    /**
+     * The container @p coefficients provides pairs of floating point numbers
+     * \f$(a_i, b_i)\f$ that determine the weight \f$w_K\f$ of each cell
+     * \f$K\f$ with \f$n_K\f$ degrees of freedom in the following way: \f[ w_K =
+     * \mathrm{truncate}\left( \sum_i a_i \, n_K^{b_i} \right) \f]
+     */
+    template <typename FactorType, typename ExponentType>
+    static WeightingFunction
+    ndofs_weighting(
+      const std::vector<std::pair<FactorType, ExponentType>> &coefficients);
```
bangerth: same issues here

include/deal.II/distributed/cell_weights.h
```diff
@@ -147,40 +255,33 @@ namespace parallel
...
      */
     boost::signals2::connection tria_listener;
 
     /**
-     * A callback function that will be attached to the cell_weight signal of
-     * the Triangulation, that is a member of the DoFHandler. Ultimately
-     * returns the weight for each cell, determined by the weighting_function
-     * member.
+     * A callback function that will be connected to the cell_weight signal of
+     * the @p triangulation, which is attached to the @p dof_handler. Ultimately
```
bangerth:
```suggestion
     * the @p triangulation, to which the @p dof_handler is attached. Ultimately
```

source/distributed/cell_weights.cc
```diff
@@ -63,71 +44,186 @@ namespace parallel
...
+  CellWeights<dim, spacedim>::ndofs_weighting(
+    const std::pair<FactorType, ExponentType> &coefficients)
   {
-    weighting_function =
-      [factor](const FiniteElement<dim, spacedim> &active_fe,
-               const typename hp::DoFHandler<dim, spacedim>::cell_iterator &)
-      -> unsigned int {
-      return factor * active_fe.dofs_per_cell * active_fe.dofs_per_cell;
-    };
+    if (std::is_same<ExponentType, unsigned int>::value)
```
bangerth: you could save yourself some work by just picking one type for `ExponentType` :-)

source/distributed/cell_weights.cc
```diff
@@ -63,71 +44,186 @@ namespace parallel
...
+                 -> unsigned int {
+          const typename std::common_type<FactorType, unsigned int>::type
+            result = 0;
+          for (const auto &pair : coefficients)
+            result +=
+              pair.first * Utilities::pow(future_fe.dofs_per_cell, pair.second);
+          result = std::trunc(result);
+
+          Assert(
+            result <= std::numeric_limits<unsigned int>::max(),
```
bangerth: `results` also needs to be non-negative, by the way!

source/distributed/cell_weights.cc
```diff
@@ -63,71 +44,186 @@ namespace parallel
...
+    const parallel::TriangulationBase<dim, spacedim> *tria =
+      dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
+        &(dof_handler.get_triangulation()));
+
+    Assert(
+      tria != nullptr,
+      ExcMessage(
+        "parallel::CellWeights requires a parallel::TriangulationBase object."));
+
+    return std::bind(&CellWeights<dim, spacedim>::weighting_callback,
```
bangerth: @masterleinad has been working hard to banish `std::bind` from the code base by using lambda functions instead -- he's going to stop liking you if you introduce new ones ;-)

source/distributed/cell_weights.cc
```diff
@@ -171,7 +267,113 @@ namespace parallel
...
+  // ---------- deprecated functions ----------
+
+  template <int dim, int spacedim>
+  CellWeights<dim, spacedim>::CellWeights(
+    const hp::DoFHandler<dim, spacedim> &dof_handler)
+    : dof_handler(&dof_handler)
+  {
+    triangulation =
+      dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
+        &(this->dof_handler->get_triangulation()));
```
bangerth: Couldn't you move this into the member initializer list above already?

source/distributed/cell_weights.cc
```diff
@@ -171,7 +267,113 @@ namespace parallel
...
+
+  template <int dim, int spacedim>
+  void
+  CellWeights<dim, spacedim>::register_constant_weighting(
+    const unsigned int factor)
+  {
+    tria_listener.disconnect();
+
+    tria_listener = triangulation->signals.cell_weight.connect(
+      std::bind(&CellWeights<dim, spacedim>::weighting_callback,
```
bangerth: same here as above

source/distributed/cell_weights.cc
```diff
@@ -171,7 +267,113 @@ namespace parallel
...
+
+  template <int dim, int spacedim>
+  void
+  CellWeights<dim, spacedim>::register_ndofs_weighting(
+    const unsigned int factor)
+  {
+    tria_listener.disconnect();
+
+    tria_listener = triangulation->signals.cell_weight.connect(
+      std::bind(&CellWeights<dim, spacedim>::weighting_callback,
```
bangerth: and here

source/distributed/cell_weights.cc
```diff
@@ -171,7 +267,113 @@ namespace parallel
...
+
+  template <int dim, int spacedim>
+  void
+  CellWeights<dim, spacedim>::register_ndofs_squared_weighting(
+    const unsigned int factor)
+  {
+    tria_listener.disconnect();
+
+    tria_listener = triangulation->signals.cell_weight.connect(
+      std::bind(&CellWeights<dim, spacedim>::weighting_callback,
```
bangerth: and here

source/distributed/cell_weights.cc
```diff
@@ -171,7 +267,113 @@ namespace parallel
...
+  {
+    tria_listener.disconnect();
+
+    const std::function<unsigned int(
+      const typename hp::DoFHandler<dim, spacedim>::cell_iterator &,
+      const FiniteElement<dim, spacedim> &)>
+      converted_function =
+        [&custom_function](
+          const typename hp::DoFHandler<dim, spacedim>::cell_iterator &cell,
+          const FiniteElement<dim, spacedim> &future_fe) -> unsigned int {
```
bangerth: I don't think you actually need to specify the return type explicitly. It doesn't hurt, but also doesn't help.

source/distributed/cell_weights.cc
```diff
@@ -171,7 +267,113 @@ namespace parallel
...
+      return custom_function(future_fe, cell);
+    };
+
+    tria_listener = triangulation->signals.cell_weight.connect(
+      std::bind(&CellWeights<dim, spacedim>::weighting_callback,
```
bangerth: and here

include/deal.II/distributed/cell_weights.h
marcfehling: We could let them be just floating point numbers. I thought we'd get some performance benefits from this since we'll work with `std::pow` on `floats`, but realize now that it would be negligibly small.

I'll remove the template parameters so we'll stick to floats.

include/deal.II/distributed/cell_weights.h
marcfehling: Unfortunately not.

`register_ndofs_weighting()` is a non-const function and changes class members, i.e. `connection`.

include/deal.II/distributed/cell_weights.h
marcfehling: We actually do mention the section you are referring to in lines 74,75 (see https://github.com/dealii/dealii/pull/8687/commits/1d5d6132a74be88e4badc728d5481f2df9ef3635).

Do you think this is sufficient?

marcfehling:
Addressed all suggestions. Once you reviewed these changes, I'll squash them together as usual (which will include rebasing on the master branch because of some merge conflicts).

source/distributed/cell_weights.cc
marcfehling: I'd prefer to be explicit about this :-)

include/deal.II/distributed/cell_weights.h
bangerth: Yes, that sounds good. I think the performance benefits are not going to be noticeable.

source/distributed/cell_weights.cc
bangerth: I appreciate this -- I think we share that trait ;-)

bangerth:
Looks good, please squash.

/rebuild

marcfehling:
Testers have passed. Commits are squashed. Ready for merge.

PR #8692, opened 2019-09-03
Remove outdated documentation on largest allowed number of dofs.

simonsticko:
In the module on distributed parallel computing the largest allowed
number of dofs is stated to be the largest attainable value of int.
This no longer true, remove the statement.


bangerth:
Excellent, thank you!

PR #8693, opened 2019-09-03
Reorganize __syncthreads for hanging nodes

masterleinad:
`cuda-memcheck --tool racecheck` does a great job for telling where race conditions related to `__shared__` data occur. It turns out that we need to synchronize
- after assigning values from the `src` vector to the `__shared__` data before we can use it,
- before writing values to the `__shared__` data in `interpolate_boundary_2d/3d`, and
- after writing values to the `__shared__` data in `interpolate_boundary_2d/3d`.

The current code already does the exact same synchronizations, but moving them around should make a little more sense.


Rombur:
Since we are only touching cuda files there is no point in running the tester.

Issue #1894, opened 2015-11-21
Extend DataOut to complex-valued vectors


bangerth:
The problem with DataOut and complex valued solution vectors is that (i) it somehow needs to rip real and imaginary parts apart, given that output formats all only support outputting real data; (ii) that DataOut stores the solution vectors it is supposed to output in opaque types that hide whether they are deal.II, PETSc or Trilinos vectors, and whether their elements are floats, doubles, or complex-valued.

To address this, here are three necessary steps:
- [x] The opaque data types that store data vectors need to be able to tell whether they store a real or complex-valued vector. To this end, let's add a virtual function `is_complex_valued()` to class `internal::DataOut::DataEntryBase` that returns whether a data vector is complex-valued. (Fixed by #5333.)
- [x] Let's add a boolean argument to the various `internal::DataOut::DataEntryBase::get_*` functions that indicates whether we are asking for the real or imaginary parts. (Fixed by #5335.)
- [X] In `DataOut::build_one_patch`, where we call this, we need to ask whether the vector is complex-valued or not (using the function introduced above) and if so, treat real and imaginary parts separately, using the mechanism outlined in part 2 above. (Fixed by #5419.)
- [x] In all of these classes, one also has to deal with the case of the postprocessor and figure out how to sort components in that case. (Fixed by #8648, #8650, #8769, #8770.)
- [x] Add instantiations of `DataOut_DoFData` for complex-valued vectors. (Fixed by #5398.)
- [x] Change the fact that currently, we output real and imaginary components immediately following each other. This is ok for scalar components, but not if an output corresponds to a vector. In that case, we first want to output the real x/y/z components, and then the imaginary x/y/z components. Same for tensors, of course. (Fixed by #9815.)
- [ ] The same has to be applied to `DataOutFaces`, `DataOutRotation`, and `DataOutStack`.
- [X] Add a changelog entry (Fixed by #10059.)

bangerth:
I should add that all of the work above is necessary because we have to separate real and imaginary parts of complex numbers once we want to create graphical output -- no graphical output file I know of understands complex numbers, so we will have to output them as a pair of real values.

davydden:
in reference to https://github.com/dealii/dealii/issues/2033

bangerth:
Here's a testcase:
```
// ---------------------------------------------------------------------
//
// Copyright (C) 2003 - 2017 by the deal.II authors
//
// This file is part of the deal.II library.
//
[...]

PR #8694, opened 2019-09-04
Indent a block properly.

bangerth:
We had a situation here where we do
```
  if (...)
  {
    ...
  }
  else
     if (...)
     { ... }
     else
     {...}
```
That's bad style, but it was made worse by the fact that clang-format *outdented* the last `else` by two spaces, bringing it to the same level as the initial `if` and the middle `else` :-(

Fix this by using the appropriate braces. Best viewed using `?w=1`.


bangerth:
Part of #1894.

PR #8695, opened 2019-09-04
Make hp::FEvaluesBase non-internal.

### LINKED ISSUE #8691: doxygen: broken base class for hp::FEValues
It seems doxygen doesn't understand the base class, see https://www.dealii.org/developer/doxygen/deal.II/classhp_1_1FEValues.html so functions like get_present_fe_values() don't show up.

drwells:
Fixes #8691 by moving `hp::internal::FEValuesBase` to `hp::FEValuesBase`. We publicly inherit from this class so I think its better (regardless of the doxygen status) to not make it `internal`.

### DIFF
--- a/include/deal.II/hp/fe_values.h
+++ b/include/deal.II/hp/fe_values.h
@@ -37,154 +37,147 @@ class FiniteElement;
 #endif
 
 
-namespace internal
+namespace hp
 {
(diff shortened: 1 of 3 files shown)


tjhei:
thanks! :+1:

PR #8698, opened 2019-09-04
SIMD: Add std::pow int-exponent variant for vectorized array

tamiko:


tamiko:
@kronbichler The avx512 variant is untested.

peterrum:
Stupid question: why cannot we write the following:
```cpp
  template <typename Number, int width, typename Number2>
  inline ::dealii::VectorizedArray<Number, width>
  pow(const ::dealii::VectorizedArray<Number, width> &x, const Number2 p);
```

include/deal.II/base/vectorization.h
```diff
@@ -4110,6 +4110,27 @@ namespace std
...
+   */
+  template <typename Number, int width>
+  inline ::dealii::VectorizedArray<Number, width>
+  pow(const ::dealii::VectorizedArray<Number, width> &x, const int p)
+  {
+    Number values[::dealii::VectorizedArray<Number, width>::n_array_elements];
+    for (unsigned int i = 0;
+         i < dealii::VectorizedArray<Number, width>::n_array_elements;
+         ++i)
+      values[i] = std::pow(x[i], p);
```
bangerth: Are there no intrinsics that can do all of this at once?

tamiko: @bangerth There are no portable intrinsics that can do all of this at once. We are currently investigating more efficient implementations of `std::pow` for `VectorizedArray` as well.
For the time being, I'd really like to see this variant merged.

include/deal.II/base/vectorization.h
```diff
@@ -4110,6 +4110,27 @@ namespace std
...
 
 
+  /**
+   * Variant of above function that takes an integer as second argument.
+   *
+   * @relatesalso VectorizedArray
+   */
+  template <typename Number, int width>
+  inline ::dealii::VectorizedArray<Number, width>
+  pow(const ::dealii::VectorizedArray<Number, width> &x, const int p)
```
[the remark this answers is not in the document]
tamiko: Yes, because this `pow` is in the `std` namespace :-)

tamiko:
@peterrum This is not a stupid question. That variant created some undefined references for me for some arcane reason but I think I tracked it down.

include/deal.II/base/vectorization.h
bangerth: Thanks for the confirmation!

include/deal.II/base/vectorization.h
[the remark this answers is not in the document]
bangerth: Ah, good point :-)

masterleinad:
> @peterrum This is not a stupid question. That variant created some undefined references for me for some arcane reason but I think I tracked it down. (nope)

Interestingly, these problems don't appear if you define the function in a different namespace.

masterleinad:
According to https://en.cppreference.com/w/cpp/language/extending_std, overloading in `namespace std` is undefined behavior.

peterrum:
> @kronbichler The avx512 variant is untested.

The test passes.

peterrum:
> @peterrum This is not a stupid question. That variant created some undefined references for me for some arcane reason but I think I tracked it down. (nope)

Coule we have an internal function which is called by `pow(VectorizedArray<Number>, Number)` and `pow(VectorizedArray<Number>, int)`?

include/deal.II/base/vectorization.h
kronbichler: Since this is an integer power, wouldn't it make sense with some specific integer code path? I know this probably duplicates some code that might be contained in the standard library, but it is reasonably simple nonetheless and, if inlined, should remove all loops and conditional branches.
```C++
  inline ::dealii::VectorizedArray<Number, width>
  pow(const ::dealii::VectorizedArray<Number, width> &x, const int p)
  {
    unsigned int positive_exponent = std::abs(p);
    VectorizedArray<Number> result = Number(1);
    VectorizedArray<Number> x_to_p = x;
    while (true)
      {
        // exponent is odd -> this happens at least once if exponent>0,
        // so the result is ok
        if (positive_exponent & 1)
          result *= x_to_p;
        positive_exponent /= 2;
        if (positive_exponent == 0)
         break;
        x_to_p *= x_to_p;
      }
    if (p > 0)
      return result;
    else
      return Number(1)/result;
  }
```
```

Comment (kronbichler) [reply include/deal.II/base/vectorization.h]: I should add that this way we do fully vectorized computations, so they should potentially be much faster.

Comment (tamiko) [reply include/deal.II/base/vectorization.h]: Good point!

Comment (bangerth) [reply include/deal.II/base/vectorization.h]: Don't we already have some `int_power` function somewhere that does essentially this, just for scalars?

Comment (masterleinad) [reply include/deal.II/base/vectorization.h]: Yes, we have `Utilities::pow` which currently only works for non-negative exponents.

Comment (kronbichler) [reply include/deal.II/base/vectorization.h]: @tamiko how much more functionality do you need? @masterleinad I guess we could switch to a templated `Utilities::pow` function, where we currently only support `int` types. I would see it work also for negative integer exponents at the expense of yet another `operator?:`. At the same time, the function body here is not too long, and it would give us the integer `pow` in the `std` namespace`.

Comment (bangerth) [reply include/deal.II/base/vectorization.h]: I didn't mean to suggest to unify code paths, but just to copy the code that has already worked well in another location. (Maybe with a reference to where the code was copied from.)

Comment (tamiko) [reply include/deal.II/base/vectorization.h]: @kronbichler Only the posted functionality for the time being.

**Comment by tamiko** [comment]
*mhm* *mhm* *mhm*

So let's see what happens:
```
#include <deal.II/base/vectorization.h>

using namespace dealii;

VectorizedArray<float>
foo(const VectorizedArray<float> a)
{
  return std::pow(a, 17);
}

VectorizedArray<float>
bar(const VectorizedArray<float> a, const int b)
{
  return std::pow(a, b);
}
```
My expectation is that a compiler has to reduce `foo` to 4 (packed single) SIMD instructions and `bar` to a tight loop. `foo` looks like
```
_Z3fooN6dealii15VectorizedArrayIfLi8EEE:
        vmulps  %ymm0, %ymm0, %ymm1
        vmulps  %ymm1, %ymm1, %ymm1
        vmulps  %ymm1, %ymm1, %ymm1
        vmulps  %ymm1, %ymm1, %ymm1
        vmulps  %ymm0, %ymm1, %ymm0
        ret
```
whereas `bar` looks like
```
_Z3barN6dealii15VectorizedArrayIfLi8EEEi:
.LFB23168:
        pushq   %rbp
.LCFI593:
        movq    %rsp, %rbp
.LCFI594:
        pushq   %r15
        pushq   %r14
        pushq   %r13
        pushq   %r12
        pushq   %rbx
        andq    $-32, %rsp
        subq    $288, %rsp
.LCFI595:
        movq    %fs:40, %rax
        movq    %rax, 280(%rsp)
        xorl    %eax, %eax
        testl   %edi, %edi
        je      .L819
        vmulps  %ymm0, %ymm0, %ymm3
        vmovaps %ymm0, %ymm1
        movl    %edi, %ebx
        js      .L862
        movl    %edi, %r12d
        sarl    %r12d
        je      .L796
        movl    %edi, %r13d
        sarl    $2, %r13d
        je      .L825
        movl    %edi, %r14d
        sarl    $3, %r14d
        vmulps  %ymm3, %ymm3, %ymm2
        je      .L826
        movl    %edi, %r15d
        sarl    $4, %r15d
        vmulps  %ymm2, %ymm2, %ymm4
        je      .L827
        movl    %edi, %eax
        sarl    $5, %eax
        vmulps  %ymm4, %ymm4, %ymm5
        movl    %eax, 192(%rsp)
        je      .L828
        vmulps  %ymm5, %ymm5, %ymm6
        vmovaps %ymm0, (%rsp)
        movl    %edi, %esi
        sarl    $6, %esi
        leaq    224(%rsp), %rdi
        vmulps  %ymm6, %ymm6, %ymm0
        vmovaps %ymm4, 32(%rsp)
        vmovaps %ymm2, 64(%rsp)
        vmovaps %ymm3, 96(%rsp)
        vmovaps %ymm5, 128(%rsp)
        vmovaps %ymm6, 160(%rsp)
        vmovaps %ymm0, 224(%rsp)
        vzeroupper
        call    _ZSt3powIfLi8EEN6dealii15VectorizedArrayIT_XT0_EEERKS3_i
        movl    192(%rsp), %eax
        vmovaps 160(%rsp), %ymm6
        testb   $1, %al
        vmovaps 128(%rsp), %ymm5
        vmovaps 96(%rsp), %ymm3
        vmovaps 64(%rsp), %ymm2
        vmovaps 32(%rsp), %ymm4
        vmovaps (%rsp), %ymm1
        jne     .L812
        vmovaps .LC16(%rip), %ymm6
.L812:
        vmulps  %ymm6, %ymm0, %ymm0
.L811:
        andl    $1, %r15d
        je      .L810
        vmulps  %ymm5, %ymm0, %ymm0
.L810:
        andl    $1, %r14d
        je      .L809
        vmulps  %ymm4, %ymm0, %ymm0
        .p2align 4,,10
        .p2align 3
.L809:
        andl    $1, %r13d
        je      .L808
        vmulps  %ymm2, %ymm0, %ymm0
        jmp     .L808
        .p2align 4,,10
        .p2align 3
.L819:
        vmovaps .LC16(%rip), %ymm0
.L796:
        movq    280(%rsp), %rax
        xorq    %fs:40, %rax
        jne     .L863
        leaq    -40(%rbp), %rsp
        popq    %rbx
        popq    %r12
        popq    %r13
        popq    %r14
        popq    %r15
        popq    %rbp
.LCFI596:
        ret
```
nope nope nope. I'll try @kronbichler's variant instead.

**Comment by tamiko** [comment]
I played around with some different implementations and none yields a convincing result in terms of emitted code. clang is doing significantly worse than gcc. What about we drop the `int p` parameter and make it a compile-time constant:
```
template <int p, typename Number, int width>
inline ::dealii::VectorizedArray<Number, width>
pow(const ::dealii::VectorizedArray<Number, width> &x)
{
  if (p == 0)
    return ::dealii::VectorizedArray<Number, width>(1.);
  else if (p < 0)
    return ::dealii::VectorizedArray<Number, width>(1.) / my_power<-p>(x);
  else
    return ((p % 2 == 1) ? x : ::dealii::VectorizedArray<Number, width>(1.)) *
           my_power<p / 2>(x * x);
}
```
This produces optimal code for both gcc and clang already at moderate optimization (`-O2` instead of `-Oomg-i-am-flying-so-fast`). And this happens to be exactly what I need.

**Comment by tamiko** [comment]
For example
```
VectorizedArray<double> foobar(const VectorizedArray<double> &a)
{
  return my_power<-5661>(a);
}
```
is happily reduced to
```
.LCPI1_0:
        .quad   4607182418800017408     # double 1
_Z6foobarRKN6dealii15VectorizedArrayIdLi4EEE: # @_Z6foobarRKN6dealii15VectorizedArrayIdLi4EEE
# %bb.0:
        vmovapd (%rdi), %ymm0
        vmulpd  %ymm0, %ymm0, %ymm1
        vmulpd  %ymm1, %ymm1, %ymm1
        vmulpd  %ymm1, %ymm1, %ymm2
        vmulpd  %ymm2, %ymm2, %ymm3
        vmulpd  %ymm3, %ymm3, %ymm4
        vmulpd  %ymm4, %ymm4, %ymm4
        vmulpd  %ymm4, %ymm4, %ymm4
        vmulpd  %ymm4, %ymm4, %ymm4
        vmulpd  %ymm4, %ymm4, %ymm4
        vmulpd  %ymm4, %ymm4, %ymm5
        vmulpd  %ymm5, %ymm5, %ymm6
        vmulpd  %ymm6, %ymm6, %ymm6
        vmulpd  %ymm6, %ymm5, %ymm5
        vmulpd  %ymm5, %ymm4, %ymm4
        vmulpd  %ymm4, %ymm3, %ymm3
        vmulpd  %ymm3, %ymm2, %ymm2
        vmulpd  %ymm2, %ymm1, %ymm1
        vmulpd  %ymm1, %ymm0, %ymm0
        vbroadcastsd    .LCPI1_0(%rip), %ymm1 # ymm1 = [1.0E+0,1.0E+0,1.0E+0,1.0E+0]
        vdivpd  %ymm0, %ymm1, %ymm0
        retq
```

**Comment by masterleinad** [comment]
> What about we drop the `int p` parameter and make it a compile-time constant:

Then it no longer matches the `std::pow` signature and it might be a good idea to put it into a different namespace. :slightly_smiling_face:

**Comment by tamiko** [comment]
@masterleinad Absolutely.

**Comment by masterleinad** [comment]
Essentially, this is what `Utilities::fixed_power` currently does.

**Comment by tamiko** [comment]
@masterleinad From a function signature, yes. Impementation wise not quite :-)

```
_Z6foobarRKN6dealii15VectorizedArrayIdLi4EEE:
.LFB25523:
        vmovapd (%rdi), %ymm1
        movl    $5660, %eax
        vmovapd %ymm1, %ymm0
        .p2align 4,,10
        .p2align 3
.L723:
        vmulpd  %ymm1, %ymm0, %ymm0
        decl    %eax
        jne     .L723
        ret
```

bangerth:
If it does what `fixed_power` does, then that's how it should be named ;-)

tamiko:
@bangerth Why? I thought @masterleinad pointed out (correctly) that the above implementation is simply redundant - we already have a function in the library.

bangerth:
Oh, so you're saying that you can call `Utilities::fixed_power` with vector types? I had missed that part.

tamiko:
@bangerth Jup it actually worked. I'll have a pull request to make it performant in just a minute.
You'll like it. It adds a test :-D

kronbichler:
@tamiko do you agree that we simply close this one, now that #8705 is merged and provides a good solution?

tamiko:
@kronbichler Yes. I would probably never use the could in here at all (currently looking into some "fast" vectorized `pow` implementations).

PR #8699, opened 2019-09-04
GeometryInfo<dim>::unit_normal_vector 

rezarastak:
Added a useful constant to `GeometryInfo` and used it to simplify a function in `mapping_cartesian.cc`. It also showcases the usefulness of `consexpr Tensor`.

### DIFF
--- a/source/fe/mapping_cartesian.cc
+++ b/source/fe/mapping_cartesian.cc
@@ -279,19 +279,13 @@ MappingCartesian<dim, spacedim>::compute_fill(
     }
 
 
-  // compute normal vectors. since
-  // cells are aligned to coordinate
-  // axes, they are simply vectors
-  // with exactly one entry equal to
-  // 1 or -1. Furthermore, all
-  // normals on a face have the same
-  // value
+  // compute normal vectors. All normals on a face have the same value.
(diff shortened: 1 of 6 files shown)


source/fe/mapping_cartesian.cc
```diff
@@ -289,50 +289,9 @@ MappingCartesian<dim, spacedim>::compute_fill(
...
-                        normals[face_no]);
-              break;
-            }
-
-          default:
-            Assert(false, ExcNotImplemented());
-        }
+      std::fill(normal_vectors.begin(),
+                normal_vectors.end(),
+                Point<dim>{GeometryInfo<dim>::unit_normal_vector[face_no]});
```
bangerth: I *think* that we changed the type of the `normal_vectors` array to `Tensor<1,dim>` at some point in the past, even though this was not reflected in the initializers in the old code you are deleting here. If true, you should be able to get away without the cast in the last line here.

rezarastak: Yes, you are right. I fixed it.

bangerth: Thanks for checking!

PR #8700, opened 2019-09-04
VectorizedArray: Some minor doxygen markup improvements

tamiko:


tamiko:
I would like to group the related functions of a `VectorizedArray` a little bit because I am about to introduce a bit more :-)

include/deal.II/base/vectorization.h
```diff
@@ -627,10 +633,14 @@ vectorized_transpose_and_store(const bool                            add_into,
 }
 
 
+//@}
+
+#ifndef DOXYGEN
```
bangerth: You exclude all of the specializations of the class from the doxygen generation. Is this on purpose?

tamiko: Yes - we never generated any official documentation where these variants get picked up, see for example [1]. There is also no additional information in the documentation of the potentially 8 specializations.

[1] https://www.dealii.org/developer/doxygen/deal.II/classVectorizedArray.html

bangerth: OK. Good to merge then!

PR #8701, opened 2019-09-04
VectorizedArray: Implement a compare_and_apply_mask function

tamiko:


include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,167 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+
+/**
+ * @name Ternary operations on VectorizedArray
+ */
+//@{
+
+/**
+ * TODO
+ */
+enum SIMDComparison : int
```
tamiko: I have created an enum with more legible names for the most useful variants.
I have chosen to only use `ordered` `non-signaling` variants here.

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,167 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+
+  return mask ? true_value : false_value;
+}
+
+//@}
+
+#ifndef DOXYGEN
+#  if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 3 && defined(__AVX512F__)
+
+template <int predicate>
```
tamiko: It is for that reason that the SIMD overloads actually don't take an enum member as template argument, but an int. This allows to use one of the `unordered`, or `signalling` variants by simply providing the macro directly. For example `_CMP_LE_US` for the unorderd signalling variant instead of `SIMDComparison::less_than_or_equal`. (All of this makes a difference only if a `NaN` is encountered in the `VectorizedArrays`s.)

include/deal.II/base/vectorization.h
bangerth: Want to make it an `enum class` instead?

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,167 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+  less_than_or_equal    = _CMP_LE_OQ,
+  greater_than          = _CMP_GT_OQ,
+  greater_than_or_equal = _CMP_GE_OQ
+};
+
+
+/**
+ * TODO
+ */
+template <SIMDComparison predicate, typename Number>
```
bangerth: why use a template parameter instead of a regular run-time argument?

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,167 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+                       const Number &true_value,
+                       const Number &false_value)
+{
+  bool mask;
+
+  switch (predicate)
+    {
+      case SIMDComparison::equal:
+        mask = left == right;
+      case SIMDComparison::not_equal:
```
bangerth: I'd put parentheses around the condition to help with readability. In any case, you're missing a `break` statement ;-)

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,167 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+template <int predicate>
+DEAL_II_ALWAYS_INLINE inline VectorizedArray<float, 16>
+compare_and_apply_mask(const VectorizedArray<float, 16> &left,
+                       const VectorizedArray<float, 16> &right,
+                       const VectorizedArray<float, 16> &true_values,
+                       const VectorizedArray<float, 16> &false_values)
+{
+  const auto mask = _mm512_cmp_ps(left.data, right.data, predicate);
+  VectorizedArray<float, 16> result;
+  result.data = _mm512_or_ps(_mm512_and_ps(mask, true_values.data),
```
bangerth: Could you in principle write this as
```
  VectorizedArray<...> result {{_mm512_or_ps(...)}};
  return result;
```
possibly with a different number of curly braces? I suspect that the `.data` member of `result` isn't initialized above anyway, so there is no duplicate work -- but this would make it really clear.

include/deal.II/base/vectorization.h
tamiko: Not without writing a lot of case statements. (Even though it doesn't look like from the signature,) the mask operations require a compile-time constant expression because the mask operations are reduced to different machine instructions (depending on the int).

include/deal.II/base/vectorization.h
tamiko: Good point done. The `break` statements got in after I wrote a test :-D

include/deal.II/base/vectorization.h
tamiko: We currently have no constructor that would support this syntax.

This would be a nice improvement for the class, though :-) @kronbichler *ping*

peterrum:
@tamiko Out of curiosity: in what context do you need this function?

include/deal.II/base/vectorization.h
peterrum: @kronbichler I am not sure about this. We can have private constructors which take `__m512` (for AVX512 and other intrinsics types for other ISAs). The reason why I think it should be private is that users should not be encouraged to use it (since it makes the user code not portable any more - which is actually one of the goals of `VectorizedArray`).

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,211 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+#  if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 3 && defined(__AVX512F__)
+
+template <int predicate>
+DEAL_II_ALWAYS_INLINE inline VectorizedArray<float, 16>
+compare_and_apply_mask(const VectorizedArray<float, 16> &left,
+                       const VectorizedArray<float, 16> &right,
+                       const VectorizedArray<float, 16> &true_values,
+                       const VectorizedArray<float, 16> &false_values)
+{
+  const auto mask = _mm512_cmp_ps(left.data, right.data, predicate);
```
kronbichler: I must admit that I had to check up the type behind `auto`, and found that it is not as easy in the AVX512 case. It seems Intel changed the name to `_mm512_cmp_ps_mask` and returns `__mmask16`. The next difficulty is that the `_mm512_and_ps` and `_mm512_or_ps` functions are not in AVX512-F, the common set available in all supported machines, but only in the AVX512-DQ. Market segmentation of Intel - so annoying.

kronbichler: I will look into how to implement those later, I guess we can use some functions operating on the masks directly.

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,211 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+template <int predicate>
+DEAL_II_ALWAYS_INLINE inline VectorizedArray<double, 8>
+compare_and_apply_mask(const VectorizedArray<double, 8> &left,
+                       const VectorizedArray<double, 8> &right,
+                       const VectorizedArray<double, 8> &true_values,
+                       const VectorizedArray<double, 8> &false_values)
+{
+  const auto mask = _mm512_cmp_pd(left.data, right.data, predicate);
+  VectorizedArray<double, 8> result;
+  result.data = _mm256_or_ps(_mm512_and_ps(mask, true_values.data),
```
```suggestion
  result.data = _mm512_or_ps(_mm512_and_pd(mask, true_values.data),
```
(but as I said above, this would need adjustments anyway).

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,211 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+DEAL_II_ALWAYS_INLINE inline VectorizedArray<float, 16>
+compare_and_apply_mask(const VectorizedArray<float, 16> &left,
+                       const VectorizedArray<float, 16> &right,
+                       const VectorizedArray<float, 16> &true_values,
+                       const VectorizedArray<float, 16> &false_values)
+{
+  const auto mask = _mm512_cmp_ps(left.data, right.data, predicate);
+  VectorizedArray<float, 16> result;
+  result.data = _mm512_or_ps(_mm512_and_ps(mask, true_values.data),
+                             _mm512_andnot_ps(mask, false_values.data));
```
kronbichler: I checked the case for AVX-512. I think we can simply write the function like this:
```C++
const __mmask16 mask = _mm512_cmp_ps(left.data, right.data, predicate);
VectorizedArray<float, 16> result;
result.data = _mm512_mask_mov_ps(false_values.data, mask, true_values.data);
return result;
```
I have not verified it yet, but this seems to be even simpler than the SSE/AVX case.

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,211 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+#  if DEAL_II_COMPILER_VECTORIZATION_LEVEL >= 1 && defined(__SSE2__)
+
+template <int predicate>
+DEAL_II_ALWAYS_INLINE inline VectorizedArray<float, 4>
+compare_and_apply_mask(const VectorizedArray<float, 4> &left,
+                       const VectorizedArray<float, 4> &right,
+                       const VectorizedArray<float, 4> &true_values,
+                       const VectorizedArray<float, 4> &false_values)
+{
+  const auto mask = _mm_cmp_ps(left.data, right.data, predicate);
```
kronbichler: It looks like the `predicate` variable is an AVX thing, so for the SSE part we need the switch statement with the six cases separately, `_mm_cmpeq_ps, _mm_cmplt_ps, _mm_cmple_ps, _mm_cmpgt_ps, _mm_cmpge_ps, _mm_cmpneq_ps`. Same in the double case below.

include/deal.II/base/vectorization.h
bangerth: I don't think I quite understand why the switch/case statements are necessary. You pass the `predicate` value to `_mm512_cmp_ps`, which as far as I can see takes an `int` as last argument. Why does it have to be a compile-time constant?

tamiko: The underlying problem is that the `_mm256_cmp_ps` intrinsic is a bit of a fake function that reduces the last argument to a dedicated machine instruction.

So, considering the following:
```
#include <x86intrin.h>
inline __m256 foo(const int pred, const __m256 left, const __m256 right) {
  return _mm256_cmp_ps(left, right, pred);
}
```

Clang does not like it at all:
```
test.cc:4:10: error: argument to '__builtin_ia32_cmpps256' must be a constant integer
  return _mm256_cmp_ps(left, right, pred);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/llvm/8/bin/../../../../lib/clang/8.0.1/include/avxintrin.h:1840:11: note: expanded from macro '_mm256_cmp_ps'
  (__m256)__builtin_ia32_cmpps256((__v8sf)(__m256)(a), \
          ^
1 error generated.
```

And gcc (after triggering the actual use):
```
test.cc: In function ‘__m256 foo(int, __m256, __m256)’:
test.cc:4:10: error: the last argument must be a 5-bit immediate
    4 |   return _mm256_cmp_ps(left, right, pred);
      |          ^~~~~~~~~~~~~
```

include/deal.II/base/vectorization.h
```diff
@@ -3941,6 +3941,211 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+        break;
+      case SIMDComparison::less_than_or_equal:
+        mask = (left.data <= right.data);
+        break;
+      case SIMDComparison::greater_than:
+        mask = (left.data > right.data);
+        break;
+      case SIMDComparison::greater_than_or_equal:
+        mask = (left.data >= right.data);
+        break;
```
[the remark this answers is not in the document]
tamiko: done

include/deal.II/base/vectorization.h
tamiko: done

tamiko:
@peterrum I want to compute SIMD vectorized nonlinear updates and limiters that require some limited form of control flow (ternary operators to select the right branch).

tests/base/vectorization_14.output.avx512
```diff
@@ -0,0 +1,57 @@
+
+DEAL::  test 8 array elements
+DEAL::-1.00000 -1.00000 1.00000 -1.00000 -1.00000 -1.00000 -1.00000 -1.00000
+DEAL::1.00000 1.00000 -1.00000 1.00000 1.00000 1.00000 1.00000 1.00000
+DEAL::1.00000 1.00000 -1.00000 -1.00000 -1.00000 -1.00000 -1.00000 -1.00000
+DEAL::1.00000 1.00000 1.00000 -1.00000 -1.00000 -1.00000 -1.00000 -1.00000
+DEAL::-1.00000 -1.00000 -1.00000 1.00000 1.00000 1.00000 1.00000 1.00000
+DEAL::-1.00000 -1.00000 1.00000 1.00000 -1.00000 -1.00000 -1.00000 -1.00000
```
kronbichler: I just ran the test and this is the only output that is wrong (copy-paste issue?):
```suggestion
DEAL::-1.00000 -1.00000 1.00000 1.00000 1.00000 1.00000 1.00000 1.00000
```

include/deal.II/base/vectorization.h
kronbichler: I can confirm @tamiko's statement here - there are a considerable number of intrinics which take an `int` argument when it actually needs to be an 8 bit immediate, i.e., compile time constant.

bangerth: I see. How unfortunate. If you don't mind, explain this in two sentences in the (still to be written) documentation so that nobody gets the idea that this could be simplified in the future.

tests/base/vectorization_14.output.avx512
tamiko: copy-paste issue. Your version is correct.

include/deal.II/base/vectorization.h
tamiko: @peterrum  You're right - let's not expose implementation details.

tamiko:
@kronbichler Thanks for the fix!
@bangerth @peterrum @kronbichler I have finished up the PR.

include/deal.II/base/vectorization.h
tamiko: I have changed this to an `enum class`.

include/deal.II/base/vectorization.h
```diff
@@ -3957,9 +3957,307 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+/**
+ * enum class encoding binary operations for a component-wise comparison of
+ * VectorizedArray data types.
+ *
+ * @note In case of SIMD vecorization (sse, avx, av512) we select the
+ * corresponding ordered, non-signalling (<code>OQ</code>) variants.
+ */
+enum class SIMDComparison : int
+{
+#if DEAL_II_COMPILER_VECTORIZATION_LEVEL > 0
```
kronbichler: This needs to be `#if DEAL_II_COMPILER_VECTORIZATION_LEVEL > 1 && defined(__AVX__)` because the relevant fields `_CMP_XX_OQ` are in `avxintrin.h`.

include/deal.II/base/vectorization.h
```diff
@@ -3957,9 +3957,307 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+/**
+ * Computes the vectorized equivalent of the following ternary operation:
+ * @code
+ *   (left OP right) ? true_value : false_value
+ * @endcode
+ * where <code>OP</code> is a binary operator (such as <code>=</code>,
+ * <code>!=</code>, <code><</code>, <code><=</code>, <code>></code>, and
+ * <code>>=</code>).
+ *
+ * Such a compuational idiom is useful as an alternative to branching
```
```suggestion
 * Such a computational idiom is useful as an alternative to branching
```

include/deal.II/base/vectorization.h
```diff
@@ -3957,9 +3957,307 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+ * example, in case of a scalar data type the statement
+ * <code>(left < right) ? true_value : false_value</code>
+ * could have been also implementd using an <code>if</code>-statement:
+ * @code
+ * if (left < right)
+ *     result = true_value;
+ * else
+ *     result = false_value;
+ * @endcode
+ * This, however, is fundamentally impossible in case of vectorization and
```
```suggestion
 * This, however, is fundamentally impossible in case of vectorization
 * because different decisions will be necessary on different vector entries
 * (lanes) and
```

include/deal.II/base/vectorization.h
```diff
@@ -3957,9 +3957,307 @@ operator<<(std::ostream &out, const VectorizedArray<Number, width> &p)
...
+ * Some more examples:
+ * @code
+ *   VectorizedArray<double> left;
+ *   VectorizedArray<double> right;
+ *
+ *   // std::max
+ *   const auto maximum = compare_and_apply_mask<SIMDComparison::greater_than>
+ *     (left, right, left, right);
+ *
+ *   // std::abs (warning: less efficient than the dedicated std::max variant)
```
kronbichler: do you want to move the text in parentheses up to the `std::max` case? (Because even in the max case we have a dedicated function, just as we have for abs.)

peterrum:
@tamiko Thanks! Makes sense! In that instance I guess there is indeed no alternative. I am looking forward to see this function (and the limiter) in practice! 

It would be so nice if C++ would allow us to misuse the `?`-operator for `VectorizedArray`s...

include/deal.II/base/vectorization.h
tamiko: fixed

include/deal.II/base/vectorization.h
tamiko: done

tamiko:
@kronbichler I have made all suggested changes, rebased and squashed.

doc/news/changes/minor/20190905MatthiasMaier
```diff
@@ -0,0 +1,8 @@
+New: compare_and_apply_mask() allows to use a ternary operater idiom with
+VectorizedArray that is based on generating a mask via component-wise
+comparison. Such a compuational idiom is useful as an alternative to
```
kronbichler:
```suggestion
comparison. Such a computational idiom is useful as an alternative to
```

tamiko: Fixed :-)

PR #8703, opened 2019-09-05
Fix compiling with CUDA support

masterleinad:
Fixes https://cdash.43-1.org/viewBuildError.php?buildid=1395. 
For some reason `nvcc` doesn't like `static constexpr std::array<Tensor<1, dim>, 2*dim>`. The only workaround I found was to use `static const` instead.

### DIFF
--- a/include/deal.II/base/geometry_info.h
+++ b/include/deal.II/base/geometry_info.h
@@ -41,209 +41,300 @@ namespace internal
     template <>
     struct Initializers<1>
     {
-      static constexpr std::array<unsigned int, 2> ucd_to_deal{{0, 1}};
+      static constexpr std::array<unsigned int, 2>
+      ucd_to_deal()
+      {
+        return {{0, 1}};
+      }
 
-      static constexpr std::array<unsigned int, 2> unit_normal_direction{
-        {0, 0}};
+      static constexpr std::array<unsigned int, 2>
+      unit_normal_direction()
+      {
+        return {{0, 0}};
+      }
 
-      static constexpr std::array<int, 2> unit_normal_orientation{{-1, 1}};
+      static constexpr std::array<int, 2>
+      unit_normal_orientation()
+      {
+        return {{-1, 1}};
+      }
 
-      static constexpr std::array<Tensor<1, 1>, 2> unit_normal_vector{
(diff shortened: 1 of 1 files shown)


bangerth:
Hm, these changes are really very unfortunate -- they make optimization and loop unrolling quite difficult because the actual values are not exposed outside the .cc file. Have you tried to initialize the list using objects of the correct type, rather than just the numbers?

rezarastak:
Have you tried something like [this](https://godbolt.org/z/INYN_1)?

masterleinad:
I pushed a new fix. Converting the `static constexpr` variables to `static constexpr` functions seems to work. For consistency, I also converted all the other variables.

masterleinad:
It would be nice to move forward with this pull request such that compiling with `CUDA` support works again. Note that the changes here are not directly related to `CUDA` functionalities.

bangerth:
But the conversion of existing functions breaks backward compatibility, doesn't it? Did you try @rezarastak 's suggestion? Are there other ways of circumventing the current style of initialization, e.g., by writing
```
  class ABC
  {
     constexpr int q = []() -> int { return 1; }();
  }
```
instead of
```
  class ABC
  {
     constexpr int q = 1;
  }
```
or maybe even
```
  namespace internal {
    constexpr int init_i { return 1; }
  };

  class ABC
  {
     constexpr int q = internal::init_i();
  }
```

Any of these options might allow us to keep these as variables instead of functions, and we could put the workaround into an `#ifdef` after the initialization `=`.

masterleinad:
> But the conversion of existing functions breaks backward compatibility, doesn't it?

No, it doesn't. The changes are only in namespace `internal`.

> Did you try @rezarastak 's suggestion? Are there other ways of circumventing the current style of initialization, e.g., by writing
> 
> ```
>   class ABC
>   {
>      constexpr int q = []() -> int { return 1; }();
>   }
> ```
Lambdas are not `constexpr` before C++17.

> or maybe even
> 
> ```
>   namespace internal {
>     constexpr int init_i { return 1; }
>   };
> 
>   class ABC
>   {
>      constexpr int q = internal::init_i();
>   }
> ```
This is what we are currently already (and continue) doing.

masterleinad:
Thanks!

PR #8704, opened 2019-09-05
Fix undefined references for set_zero_all

masterleinad:


include/deal.II/lac/la_vector.h
```diff
@@ -160,6 +160,15 @@ namespace LinearAlgebra
...
            const bool omit_zeroing_entries = false) override;
 
+    /**
+     * Returns `false` as this is a serial block vector.
+     *
+     * This functionality only needs to be called if using MPI based vectors and
+     * exists in other objects for compatibility.
+     */
+    bool
+    has_ghost_elements() const;
```
bangerth: Is this from a different part of your work? It doesn't match the PR subject.

masterleinad: No, we need that for being able to instantiate the function. It seems that this is the only vector class that is missing that function.

include/deal.II/lac/la_vector.h
```diff
@@ -160,6 +160,15 @@ namespace LinearAlgebra
     reinit(const VectorSpaceVector<Number> &V,
            const bool omit_zeroing_entries = false) override;
 
+    /**
+     * Returns `false` as this is a serial block vector.
```
drwells:
```suggestion
     * Returns `false` as this is a serial vector.
```

doc/news/changes/minor/20190905DanielArndt
```diff
@@ -0,0 +1,4 @@
+Fixed: Implementing LinearAlgebra::has_ghost_elements() improves compatibility
```
kronbichler:
```suggestion
Fixed: Implementing LinearAlgebra::Vector::has_ghost_elements() improves compatibility
```

masterleinad:
We should merge this soonish so the testers work again.

kronbichler:
> We should merge this soonish so the testers work again.

I agree. If nobody objects within the next two hours, I will merge.

drwells:
This patch seems completely uncontroversial so lets merge.

PR #8705, opened 2019-09-05
Utilities::fixed_power - use exponentiation by squaring

tamiko:
Also add a SIMD test variant

In reference to #8698

@bangerth @masterleinad @kronbichler FYI


include/deal.II/base/utilities.h
```diff
@@ -1030,27 +1030,15 @@ namespace Utilities
 {
   template <int N, typename T>
   inline T
-  fixed_power(const T n)
+  fixed_power(const T x)
   {
-    Assert(N >= 0, ExcNotImplemented());
```
bangerth: I think it would be useful to leave this in if `T` is an integral type. There's a type traits in `std::` for this somewhere.

include/deal.II/base/utilities.h
```diff
@@ -1030,27 +1030,15 @@ namespace Utilities
...
-            result *= n;
-          return result;
-      }
+    if (N == 0)
+      return T(1.);
+    else if (N < 0)
+      return T(1.) / fixed_power<-N>(x);
+    else
+      // Use exponentiation by squaring:
+      return ((N % 2 == 1) ? x : T(1.)) * fixed_power<N / 2>(x * x);
```
bangerth: You can avoid one half multiplication (on average) if you rewrote this as
```suggestion
      return ((N % 2 == 1) ? x * fixed_power<N / 2>(x * x) : fixed_power<N / 2>(x * x));
```

tamiko: @bangerth Both (clang and gcc) seem to happily elid the multiplication with `T(1.)` when unrolling (also for `VectorizedArray`.

tamiko: But I like your variant much more - its more readable.

include/deal.II/base/utilities.h
```diff
@@ -1030,27 +1030,20 @@ namespace Utilities
...
-          return n * n * n * n;
-        default:
-          T result = n;
-          for (int d = 1; d < N; ++d)
-            result *= n;
-          return result;
-      }
+    static_assert(
+      !std::is_integral<T>::value || (N > 0),
+      "The non-type template parameter N must be a non-negative integer for integral type T");
```
tamiko: @bangerth I have added a `static_assert`.

bangerth: Even better!

bangerth: Well, it could be zero if an integer.

tamiko: It's getting late. Fixed.

masterleinad: The testers confirm that we still need `Assert` here since we have a branch where we call the function with `-N` as template parameter (and we can't use `if constexpr`).

tamiko: @masterleinad I would really prefer to have the `static_assert` in place. What do you think of the current variant?

include/deal.II/base/utilities.h
```diff
@@ -1030,27 +1030,21 @@ namespace Utilities
...
-      }
+    Assert(
+      !std::is_integral<T>::value || (N >= 0),
+      ExcMessage(
+        "The non-type template parameter N must be a non-negative integer for integral type T"));
+
+    if (N == 0)
+      return T(1.);
+    else if (N < 0)
+      return T(1.) / fixed_power<-N>(x);
```
bangerth: For reference, you could have stuck with the `static_assert` if you had made sure that the function here is never called with a negative argument. This line is executed if `N<0`, so what you're really calling here is `fixed_power<abs(N)>`, which you can write as `fixed_power<(N>=0 ? N : -N>`. This way, no code paths explicitly call `fixed_power` with a negative argument, and you can revert to the `static_assert`.

If you're up to it, feel free to propose the change as a follow-up (with a comment explaining the funny conditional); if not, let me know and I'll take care of it!

masterleinad: Good suggestion! https://godbolt.org/z/qZHi5d confirms that this works.

PR #8709, opened 2019-09-08
Add maxwell singularity test

dangars:
This tests the maxwell singularities that can be found in the [Fichera Corner](https://en.wikipedia.org/wiki/Gaetano_Fichera) or [hyper_L](https://www.dealii.org/current/doxygen/deal.II/namespaceGridGenerator.html#a3b2a4ad8296c2b72a11d23b5969e8cc0) 

Note that the test fails for `Nedelec` and `NedelecSZ`.

The equation is `curl(cur E)- omega^2*epsilon_0*mu_0 E = -i*omega*J`

The manufactured solution has a singularity for `curl(curl E)` at `x=y=0`. The simulated solution for `E_z` is correct, but the simulated solution is discontinuous for `E_x` and `E_y`. Below you can find the results for `z=0`:

e_z
e_x
e_y

### DIFF
--- a/tests/fe/fe_nedelec_singularity_01.cc
+++ b/tests/fe/fe_nedelec_singularity_01.cc
@@ -1,0 +1,798 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+//
+// By Daniel Garcia-Sanchez, CNRS
+//
+// Test a maxwell singularity in 3D. Maxwell singularities are common in sharp
+// metallic edges such as the Fichera corner. Here we test the elements Nedelec
+// and NedelecSZ using the L2 norm and the continuity of the solution.
+//
+// This test solves the complex valued curl-curl equation in 3D:
+//
+// curl((1/mu_r)curl(E))
+//   -omega^2*epsilon_0*mu_0(epsilon_r-(i*sigma/(omega*epsilon_0)))E
+//                                                        = RightHandSide
+//
+// The manufactured solution is:
+//
+// E_x = (((x^2 / sqrt(x^2 + y^2)) * (x^2 - (dimension_x / 2)^2) *
+//                 (y^2 - (dimension_y / 2)^2) * (z^2 - (dimension_z / 2)^2)) /
+//        ((dimension_x / 2)^3 * (dimension_y / 2)^2 * (dimension_z / 2)^2))
+// E_y = ( ((y^2 / sqrt(x^2 + y^2)) * (x^2 - (dimension_x / 2)^2) *
+//                 (y^2 - (dimension_y / 2)^2) * (z^2 - (dimension_z / 2)^2)) /
+//        ((dimension_x / 2)^2 * (dimension_y / 2)^3 * (dimension_z / 2)^2))
+// E_z =  10 * (x * (x^2 - (dimension_x / 2)^2) * (y^2 - (dimension_y / 2)^2) /
+//                      ((dimension_x / 2)^2 * (dimension_y / 2)^2))
+//
+// This manufactured solution has a singularity at x = y = 0
+//
+// The right hand side can be calculated with a symbolic math package such as
+// sympy.
+
+#include <deal.II/base/conditional_ostream.h>
+#include <deal.II/base/function.h>
+#include <deal.II/base/index_set.h>
+#include <deal.II/base/quadrature_lib.h>
+#include <deal.II/base/tensor.h>
+#include <deal.II/base/timer.h>
+#include <deal.II/base/utilities.h>
+
+#include <deal.II/distributed/tria.h>
+
+#include <deal.II/dofs/dof_accessor.h>
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/fe/fe_nedelec.h>
+#include <deal.II/fe/fe_nedelec_sz.h>
+#include <deal.II/fe/fe_q.h>
+#include <deal.II/fe/fe_system.h>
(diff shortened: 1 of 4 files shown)


masterleinad:
What is your intention (or the status) here? After all the discussion in #8710, do you want to update the quadrature formula and then submit it as a test?

dangars:
Yes, I'll update the quadrature formula and then I'll submit this as a test. We can leave this as WIP.

What is the policy regarding the tests? How much "computing time" is reasonable for a test. I would prefer if this test is in 3D instead of 2D. If I have a coarse mesh, the test could take just a few minutes.

I could also have two versions of the test, one in 3D with a coarse mesh and another one in 2D with a finer mesh.

bangerth:
Tests should generally run in under a minute if at all possible. Having a 2d test separate from the 3d test is useful.

But why use such a fine mesh. Do you suspect that you see something fundamentally different on a fine mesh than you would see on a coarse mesh? The pictures are of course not quite as nice, but the fundamental behavior should be the same -- a test doesn't have to compute something to high accuracy to be useful, it just needs to demonstrate a particular behavior that we want to make sure remains unchanged over time.

dangars:
> But why use such a fine mesh. Do you suspect that you see something fundamentally different on a fine mesh than you would see on a coarse mesh? The pictures are of course not quite as nice, but the fundamental behavior should be the same -- a test doesn't have to compute something to high accuracy to be useful, it just needs to demonstrate a particular behavior that we want to make sure remains unchanged over time.

Yes, there is no need for a fine mesh. I can clearly see in the coarse mesh if the test fails. I will make it run in less than a minute.

@bangerth @masterleinad The test succeeds now with `QGauss<dim>(order * 2 + 3)`. According to your experience. Is it useful to write a quadrature formula that is specific for Nedelec and NedelecSZ? My intuition is that it should be possible to write a quadrature formula tailored specifically for Nedelec that should work better than `QGauss`, but I'm not 100% sure.

If it is possible to write a specific quadrature formula for Nedelec, then I'll write it and use this PR as test for the new quadrature formula. If it is not possible to write a specific quadrature formula for Nedelec then I will submit the test now with minor modifications to make it work in 1 minute.

masterleinad:
> @bangerth @masterleinad The test succeeds now with `QGauss<dim>(order * 2 + 3)`. According to your experience. Is it useful to write a quadrature formula that is specific for Nedelec and NedelecSZ? My intuition is that it should be possible to write a quadrature formula tailored specifically for Nedelec that should work better than `QGauss`, but I'm not 100% sure.

In the end, you have tensor product polynomials where the degree in one dimension is one less than for the remaining ones and this different for every component. The common polynomial space is still given by tensor product polynomials of degree `order+1`. Hence, I would expect `QGauss` not to be too bad.

bangerth:
@masterleinad states this correctly, but just to be clear: If you have a finite element with polynomials of degree k (in each coordinate direction), then you expect to get O(h^k) in the energy norm for the error if you did exact integration. Now, if you do quadrature instead, then in order to not spoil the O(h^k) convergence order, you need a quadrature formula that is accurate enough to integrate polynomials of degree 2k exactly. Since a Gauss formula with p points in each coordinate direction integrates polynomials of degree 2p-1 exactly, you need p=k+1.

This guarantees the correct convergence order regardless of the choice of right hand side, but it does not imply that you can't get *more accurate* if you used a higher order quadrature formula. But at least you will not get a higher *convergence order*.

bangerth:
I think the conclusion of all of this is that there is no need for a specialized quadrature formula. In particular, any other choice will in general have to have a larger number of quadrature points to achieve the same convergence order.

dangars:
@bangerth @masterleinad You are right! As I mentioned in https://github.com/dealii/dealii/pull/8720#discussion_r322851928, the quadrature formula `QGauss(order+2)` is enough for Nedelec and gives the right solution :)
There is no need for another quadrature formula. I'm going to review a little bit the tests to make it work in less than a minute. I'll let you know when they are ready for review

tests/fe/fe_nedelec_singularity_01.cc
```diff
@@ -0,0 +1,798 @@
...
+      VectorTools::integrate_difference(dof_handler,
+                                        locally_relevant_solution,
+                                        exact_solution_function,
+                                        difference_per_cell,
+                                        QGauss<dim>(fe_order + 2),
+                                        VectorTools::L2_norm);
+      const double L2_error =
+        VectorTools::compute_global_error(triangulation,
+                                          difference_per_cell,
+                                          VectorTools::L2_norm);
```
dangars: I copied this pattern from other tests. How does the `L2_norm` work? The result of `integrate_difference()` combined with `compute_global_error()` is `\int_Omega\sqrt(simulated_solution - exact_solution)^2` ? What does it happen if one cell is smaller than the rest of the cells. Does it take into account the size of the cell? The `L2_norm` is normalize by the total volume of the domain?

What does it happen if you run the test and the `L2_error` is smaller? The test would fail in the case of a smaller `L2_error`. It should always give roughly the same error because the mesh has the same size, right?

tests/fe/fe_nedelec_singularity_01.cc
```diff
@@ -0,0 +1,798 @@
...
+      }
+      // Only one process has the solution_a or/and solution_b. This is a simple
+      // approach to send solution_a and solution_b to all the processes.
+      Utilities::MPI::sum(solution_a, mpi_communicator, solution_a);
+      Utilities::MPI::sum(solution_b, mpi_communicator, solution_b);
+      deallog << " Solution(point_a) : " << solution_a << std::endl;
+      deallog << " Solution(point_b) : " << solution_b << std::endl;
+      // Vector does not provide operator-
+      deallog << " Solution(point_b) - solution (point_a): "
+              << (solution_b -= solution_a) << std::endl;
```
dangars: `point_a` and `point_b` are very close, but they are in two different elements. The difference between the solution at those two points should be small.

tests/fe/fe_nedelec_singularity_01.cc
bangerth: Yes, it's the integral, so small cells are weighted less than large cells. (Each entry of `difference_per_cell` has a value of `sqrt(\int_K |u-u_h|^2)`. The `compute_global_error()` function then takes the sum of squares of these values, and takes the square root -- which corresponds to the global error: `sqrt(\int_\Omega |u-u_h|^2)`.)

tests/fe/fe_nedelec_singularity_01.cc
bangerth: Yes.

PR #8713, opened 2019-09-09
Put implementation details of FEFaceEvaluation into separate file

kronbichler:
We already have most of the implementation of the tensor product/sum factorization routines of the matrix-free evaluator `FEEvaluation` in a separate file `evaluation_kernels.h`, whereas significant parts of `FEFaceEvaluation` were still in the main `fe_evaluation.h` header file. To make the file a bit more compact, this PR also moves the face-specific routines into the kernel file `evaluation_kernels.h`.

Furthermore, I also added a separate header file for organizing the access to the vectors, rather than keeping that inline in `FEEvaluation`.

Even after this change, we still have around 7650 lines in `fe_evaluation.h`, so this change is really a step in the right direction.

There should not be any code changes except for providing an appropriate interface to call for `FEFaceEvaluation`, it is just moving existing code around and making sure we have access to all variables that are used in `FEFaceEvaluation`.

@peterrum FYI


include/deal.II/matrix_free/operators.h
```diff
@@ -1091,35 +1042,14 @@ namespace MatrixFreeOperators
...
-            evaluator.template hessians<0, true, false>(out, out);
-          }
-        if (dim == 1)
-          evaluator.template hessians<0, true, false>(in, out);
-      }
+    internal::CellwiseInverseMassMatrixImpl<dim,
+                                            fe_degree,
+                                            n_components,
+                                            VectorizedArrayType>::
+      transform_from_q_points_to_basis(inverse_shape,
```
peterrum: Nice!

include/deal.II/matrix_free/vector_access_internal.h
```diff
@@ -0,0 +1,683 @@
...
+
+  // Below, three classes (VectorReader, VectorSetter,
+  // VectorDistributorLocalToGlobal) implement the same interface and can be
+  // used to to read from vector, set elements of a vector and add to elements
+  // of the vector.
+
+  // 1. A class to read data from vector
+  template <typename Number, typename VectorizedArrayType>
+  struct VectorReader
+  {
```
peterrum: I like these new classes! I will have a second look a these later, since I think I might be able to use it and be able to replace a class (700 locs) from our application code (FYI @kkormann).

kronbichler: They are not new, though, I just moved them from the `fe_evaluation.h` to this one.

kronbichler:
*ping*

masterleinad:
> There should not be any code changes except for providing an appropriate interface to call for `FEFaceEvaluation`, it is just moving existing code around and making sure we have access to all variables that are used in `FEFaceEvaluation`.

Can you please point out where you add the interface for calling `FEFaceEvaluation`? I guess we are all fine with moving the code, but the actual change is then hard to spot.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -8235,502 +7474,43 @@ FEFaceEvaluation<
...
+                        evaluate_gradients,
+                        this->active_fe_index,
+                        this->first_selected_component,
+                        this->cell,
+                        this->face_no,
+                        this->subface_index,
+                        this->dof_access_index,
+                        this->face_orientation,
+                        this->mapping_data->descriptor[this->active_fe_index]
+                          .face_orientations))
```
kronbichler: This spot from lines 7477-7508 is one of the large blocks where I outsourced to an implementation class.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -8146,73 +7433,25 @@ FEFaceEvaluation<dim,
...
+                                    this->begin_gradients(),
+                                    this->scratch_data,
+                                    integrate_values,
+                                    integrate_gradients,
+                                    this->face_no,
+                                    this->subface_index,
+                                    this->face_orientation,
+                                    this->mapping_data
+                                      ->descriptor[this->active_fe_index]
+                                      .face_orientations);
```
kronbichler: Here is one spot of the simpler kind I moved to a single implementation file. This is similar to `SelectEvaluator` in `FEEvaluation` but now for the face variant. If you @masterleinad want I can pick a name a bit closer to the cell selection class. I don't know, maybe `SelectEvaluatorFace` maybe. (But in the present we do not have any of the recursive parts we could do for the cell part).

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -8759,559 +7539,51 @@ FEFaceEvaluation<
...
+                          integrate_gradients,
+                          this->active_fe_index,
+                          this->first_selected_component,
+                          this->cell,
+                          this->face_no,
+                          this->subface_index,
+                          this->dof_access_index,
+                          this->face_orientation,
+                          this->mapping_data->descriptor[this->active_fe_index]
+                            .face_orientations))
```
kronbichler: And this, lines 7542-7575, is the second one of these big spots.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -8759,559 +7539,51 @@ FEFaceEvaluation<
...
-          typename Number,
-          typename VectorizedArrayType>
-inline void
-FEFaceEvaluation<
-  dim,
-  fe_degree,
-  n_q_points_1d,
-  n_components,
-  Number,
-  VectorizedArrayType>::adjust_for_face_orientation(const bool integrate,
```
kronbichler: This also moved to the implementation class.

include/deal.II/matrix_free/fe_evaluation.h
masterleinad: This is an internal class anyway so we don't need to overly picky with the name. I am fine with that.

Issue #8710, opened 2019-09-08
Fichera corner Maxwell simulation fails


dangars:
@bangerth , @tamiko and @tjhei During the workshop I showed you some Maxwell simulations in which the solution was discontinuous. I've been working on this and I've been able to isolate the problem. I find these discontinuities when I have sharp metallic corners.

I find the same problem for `Nedelec` and for `NedelecSZ`. I don't know if the problem is related to the finite element. @rosskynch any ideas?

The Nedelec element is supposed to be able to deal with the singularities of the Maxwell equations. An example is the Fichera Corner. I found that `deal.II` provides the [Fichera Corner geometry (hyper_L)](https://www.dealii.org/current/doxygen/deal.II/namespaceGridGenerator.html#a3b2a4ad8296c2b72a11d23b5969e8cc0) and there is a reference to Fichera in the `deal.II` documentation, which makes think me that somebody studied these problems with `deal.II` in the past.

The test #8709 reproduces the problem. The manufactured solution has a singularity for `curl(curl E)` at `x=y=0`. The solution for `E_z` is correct, but the solution is discontinuous for `E_x` and `E_y`. Below you can find the results for `z=0`:

![e_z](https://user-images.githubusercontent.com/25910642/64487191-8bae4900-d237-11e9-834d-1b92598afdc9.png)
![e_x](https://user-images.githubusercontent.com/25910642/64487187-7cc79680-d237-11e9-81b9-b01e44a440ac.png)
![e_y](https://user-images.githubusercontent.com/25910642/64487189-851fd180-d237-11e9-9736-1221239157e7.png)

rosskynch:
How do things behave if you try to isolate the singularity with h refinement?

dangars:
@rosskynch Thanks for the quick reply!

I'll try h refinement. I just tried a global refinement, but it does not help as you can see below. In "average" the solution is good, but there are discontinuities. I'm using `std::complex`, I don't think that the origin is the `std::complex` algebra, but I can do the same test for a `real` simulation.

I din't use the `hyper_L` geometry for the test. In order to rule out the mesh, I used a cube that has the dimensions `0.04x0.04x0.04mm` and a manufactured solution with a singularity in `curl(curl(E))`. I had to play a lot with different manufactured solutions in order to find one that reproduces the problem. Below you can find the vector field of the exact manufactured solution.

[...]

PR #8720, opened 2019-09-09
Clarify the degree of the elements NedelecSZ and Nedelec

dangars:
Here I summarize the outcome of the discussion in #8710

The NedelecSZ class used `degree` for the constructor, but it was really the 'order' of the element. The polynomial degree as used in the base class is 'order+1'.

See commit 121ecf772dccea4527ee6ef4247b10ca7d079dfd and #3459

I put the same text in the constructor of Nedelec and NedelecSZ in order to be consistent. Both elements have the same number of elements for `order=0` and `oder=1`.

### DIFF
--- a/include/deal.II/fe/fe_nedelec.h
+++ b/include/deal.II/fe/fe_nedelec.h
@@ -114,10 +114,16 @@ class FE_Nedelec : public FE_PolyTensor<dim>
 {
 public:
   /**
-   * Constructor for the N&eacute;d&eacute;lec element of given @p order.
-   * The maximal polynomial degree of the shape functions is
-   * <code>order+1</code> (in each variable; the total polynomial degree
-   * may be higher).
+   * Constructor for the Nedelec element of given @p order. The maximal
+   * polynomial degree of the shape functions is `order+1` (in each variable;
+   * the total polynomial degree may be higher). If `order = 0`, the element is
+   * linear and has degrees of freedom only on the edges. If `order >=1` the
+   * element has degrees of freedom on the edges, faces and volume. For example
+   * the 3D version of FE_Nedelec has 12 degrees of freedom for `order = 0`
+   * and 54 for `degree = 1`. It is important to have enough quadrature points
+   * in order to perform the quadrature with sufficient accuracy.
+   * For example [QGauss<dim>(order + 2)](@ref QGauss) can be used for the
+   * quadrature formula, where `order` is the order of FE_Nedelec.
    */
   FE_Nedelec(const unsigned int order);
 
--- a/include/deal.II/fe/fe_nedelec_sz.h
+++ b/include/deal.II/fe/fe_nedelec_sz.h
@@ -77,9 +77,18 @@ public:
                 "FE_NedelecSZ is only implemented for dim==spacedim!");
 
   /**
-   * Constructor for an element of given @p degree.
+   * Constructor for the NedelecSZ element of given @p order. The maximal
+   * polynomial degree of the shape functions is `order+1` (in each variable;
+   * the total polynomial degree may be higher). If `order = 0`, the element is
+   * linear and has degrees of freedom only on the edges. If `order >=1` the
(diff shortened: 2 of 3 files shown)


include/deal.II/fe/fe_nedelec_sz.h
```diff
@@ -77,9 +84,15 @@ class FE_NedelecSZ : public FiniteElement<dim, dim>
...
+   * Constructor for the NedelecSZ element of given @p order.
+   * The maximal polynomial degree of the shape functions is `order+1` (in each
+   * variable; the total polynomial degree may be higher). If `order = 0`, the
+   * element is linear and has degrees of freedom only on the edges. If
+   * `order >=1` the element has degrees of freedom on the edges, faces and
+   * volume. For example the 3D version of FE_NedelecSZ has 12 degrees of
+   * freedom for `order = 0` and 54 for `degree = 1`.
    */
-  FE_NedelecSZ(const unsigned int degree);
+  FE_NedelecSZ(const unsigned int order);
```
bangerth: yes, nice change!

include/deal.II/fe/fe_nedelec.h
```diff
@@ -106,6 +106,13 @@ DEAL_II_NAMESPACE_OPEN
  * QGauss<sub>k+2</sub> on each face and in the interior of the cell (or none
  * for N<sub>1</sub>).
  *
+ * It is important to have enough quadrature points in order to perform the
+ * quadrature with sufficient accuracy. For example
+ * [QGauss<dim>(2 * order + 3)](@ref QGauss) can be used for the quadrature
```
bangerth: That's too much. The general rule is that the number of quadrature points must be `poldegree + 1` in each coordinate direction. So a Nedelec element of order 0 has polynomial degree 1, and would need a 2-point Gauss quadrature. Does that not work in your case?

Also, interesting approach to use `[...](@ref QGauss)`. Does this get rendered as you hoped in the doxygen HTML output?

dangars: > That's too much. The general rule is that the number of quadrature points must be `poldegree + 1` in each coordinate direction. So a Nedelec element of order 0 has polynomial degree 1, and would need a 2-point Gauss quadrature. Does that not work in your case?

`poldegree + 1`  works when you don't have a singularity, but when you have a singularity you really need lot of quadrature points. This makes me think that it should be possible to write a better quadrature formula for Nedelec. Maybe a a quadrature formula that uses the basis functions of Nedelec?

In #8710 @rosskynch mentioned that he needed the quadrature to be high.

Having many quadrature points is inconvenient because that increases the calculation time in the assemble step. How about we write now in the documentation `2 * order + 3` for singularites (as I wrote) and at the same time I try to figure out if there is better quadrature formula for Nedelec.

> Also, interesting approach to use `[...](@ref QGauss)`. Does this get rendered as you hoped in the doxygen HTML output?

I found it in the [doxygen markdown documentation](http://www.doxygen.nl/manual/markdown.html) :)

masterleinad: > `poldegree + 1` works when you don't have a singularity, but when you have a singularity you really need lot of quadrature points. This makes me think that it should be possible to write a better quadrature formula for Nedelec. Maybe a a quadrature formula that uses the basis functions of Nedelec?

But the singularity is not related to the Nedelec elemet but rather to the equation you are trying to solve. Unless your coefficients are non-constant, assembling the system matrix (for Maxwell) should be exact using `QGauss(order+2)`.

dangars: That is right! I tested it and `QGauss(order+2)` gives the right solution :) It is important not to use `QGauss(order+1)` with `Nedelec`. I modified a little bit the text. I placed the clarification in the constructor, because that is where people are going to look. Let me know if the text is ok or you have suggestions.

bangerth: Ah, nice to know!

PR #8711, opened 2019-09-08
[Post 9.2] Remove PolynomialType template argument from FE_Poly

GrahamBenHarper:
This is a WIP, so I'm going to ask for some preliminary feedback since my build stops due to an error.

`FE_DGP_Monomial` calls for member functions of `poly_space` that don't exist in the `ScalarPolynomialsBase` class currently. Namely, the `directional_degrees()` call in `fe_dgp_monomial.cc` on line 458, 480 and the `compute_value()` call in `fe_dgp_monomial.cc` on line 238.
I assume this means I should add declarations of these two functions to `ScalarPolynomialsBase` so that `FE_DGP_Monomial` can properly look them up, but I'd like to make sure I'm not missing an easier approach.

This PR completes step 6/6, so I believe this closes #1973 
1. [x] Implement the `TensorPolynomialsBase` class #8528 
2. [x] Change tensor polynomial classes to derive from it #8588 
3. [x] Remove the class template argument from `FE_PolyTensor` #8597 
4. [x] Implement the `ScalarPolynomialsBase` class #8651 
5. [x] Change scalar polynomial classes to derive from it #8678 
6. [x] Remove the class template argument from `FE_Poly`

PR #8678, opened 2019-09-01
Change polynomial classes to derive from ScalarPolynomialsBase

GrahamBenHarper:
This is a WIP at the moment. I did a bit of work in related files to make all of this consistent with #8619 as well.

I believe it compiles in its current form, but some of these classes have their own derived classes and I need to check that things are inherited and implemented properly in the derived classes, so we shouldn't test it just yet.

As far as answering #1973, this will complete step 5/6
1. [x] Implement the TensorPolynomialsBase class
2. [x] Change tensor polynomial classes to derive from it
3. [x] Remove the class template argument from FE_PolyTensor
4. [x] Implement the ScalarPolynomialsBase class
5. [x] Change scalar polynomial classes to derive from it
6. [ ] Remove the class template argument from FE_Poly

PR #8723, opened 2019-09-10
Change TensorProductPolynomials and AnisotropicPolynomials to derive from ScalarPolynomialsBase

### LINKED ISSUE #8682: Rename AnisotropicPolynomials::compute() to evaluate()
This is a follow-up to #8619 to make all the polynomials classes consistent; however, this is used in many test files and also referred to by classes like `PolynomialsBernardiRaugel`, so the changes for this will likely span many files.

This was made obvious in  #8678 while changing the test `anisotropic_1.cc`. Now there are two polynomial spaces side-by-side, one is `TensorProductPolynomials<dim>`, and one is `AnisotropicPolynomials<dim>`, and the former calls `evaluate()` but the latter calls `compute()`.

GrahamBenHarper:
While working on #8711, I realized that these two classes need to derive from `ScalarPolynomialsBase` as well since there are FE classes which are constructed using `TensorProductPolynomials` or `AnisotropicPolynomials`. I will have to rebase #8711 onto this once everything is polished. Think of this as a follow-up to #8678 for working on #1973 as well.

To do: Currently the base classes store duplicate information for the number of polynomials, but I was unsure what value should be assigned to the base class for `degree`, so I used a placeholder.

This passes the `base/` and `fe/` tests on my end at the moment.

This closes #8682 as well because I rename all instances of `compute()` to `evaluate()`.

### DIFF
--- a/include/deal.II/base/tensor_product_polynomials.h
+++ b/include/deal.II/base/tensor_product_polynomials.h
@@ -22,6 +22,7 @@
 #include <deal.II/base/exceptions.h>
 #include <deal.II/base/point.h>
 #include <deal.II/base/polynomial.h>
+#include <deal.II/base/scalar_polynomials_base.h>
 #include <deal.II/base/tensor.h>
 #include <deal.II/base/utilities.h>
 
@@ -65,7 +66,7 @@ class TensorProductPolynomialsBubbles;
  * 2003
  */
 template <int dim, typename PolynomialType = Polynomials::Polynomial<double>>
-class TensorProductPolynomials
+class TensorProductPolynomials : public ScalarPolynomialsBase<dim>
 {
 public:
   /**
@@ -126,7 +127,7 @@ public:
            std::vector<Tensor<1, dim>> &grads,
            std::vector<Tensor<2, dim>> &grad_grads,
            std::vector<Tensor<3, dim>> &third_derivatives,
-           std::vector<Tensor<4, dim>> &fourth_derivatives) const;
+           std::vector<Tensor<4, dim>> &fourth_derivatives) const override;
 
   /**
    * Compute the value of the <tt>i</tt>th tensor product polynomial at
@@ -192,23 +193,22 @@ public:
   compute_grad_grad(const unsigned int i, const Point<dim> &p) const;
 
   /**
-   * Return the number of tensor product polynomials. For <i>n</i> 1d
-   * polynomials this is <i>n<sup>dim</sup></i>.
+   * Return the name of the space, which is <tt>TensorProductPolynomials</tt>.
    */
-  unsigned int
-  n() const;
+  std::string
+  name() const override;
 
+  /**
+   * @copydoc ScalarPolynomialsBase<dim>::clone()
+   */
+  virtual std::unique_ptr<ScalarPolynomialsBase<dim>>
+  clone() const override;
 
 protected:
   /**
    * Copy of the vector <tt>pols</tt> of polynomials given to the constructor.
    */
   std::vector<PolynomialType> polynomials;
-
-  /**
-   * Number of tensor product polynomials. See n().
-   */
-  unsigned int n_tensor_pols;
 
   /**
    * Index map for reordering the polynomials.
@@ -268,7 +268,7 @@ protected:
  * @author Wolfgang Bangerth 2003
  */
 template <int dim>
-class AnisotropicPolynomials
+class AnisotropicPolynomials : public ScalarPolynomialsBase<dim>
 {
 public:
   /**
@@ -282,6 +282,9 @@ public:
    * Since we want to build <i>anisotropic</i> polynomials, the @p dim
    * sets of polynomials passed in as arguments may of course be
    * different, and may also vary in number.
+   *
+   * The number of tensor product polynomials is <tt>Nx*Ny*Nz</tt>, or with
+   * terms dropped if the number of space dimensions is less than 3.
    */
   AnisotropicPolynomials(
     const std::vector<std::vector<Polynomials::Polynomial<double>>>
@@ -292,7 +295,7 @@ public:
    * product polynomial at <tt>unit_point</tt>.
    *
    * The size of the vectors must either be equal <tt>0</tt> or equal
-   * <tt>n_tensor_pols</tt>.  In the first case, the function will not compute
+   * <tt>this->n()</tt>.  In the first case, the function will not compute
    * these values.
    *
    * If you need values or derivatives of all tensor product polynomials then
@@ -301,12 +304,12 @@ public:
    * in a loop over all tensor product polynomials.
    */
   void
-  compute(const Point<dim> &           unit_point,
-          std::vector<double> &        values,
-          std::vector<Tensor<1, dim>> &grads,
-          std::vector<Tensor<2, dim>> &grad_grads,
-          std::vector<Tensor<3, dim>> &third_derivatives,
-          std::vector<Tensor<4, dim>> &fourth_derivatives) const;
+  evaluate(const Point<dim> &           unit_point,
+           std::vector<double> &        values,
+           std::vector<Tensor<1, dim>> &grads,
+           std::vector<Tensor<2, dim>> &grad_grads,
+           std::vector<Tensor<3, dim>> &third_derivatives,
+           std::vector<Tensor<4, dim>> &fourth_derivatives) const override;
 
   /**
    * Compute the value of the <tt>i</tt>th tensor product polynomial at
@@ -317,7 +320,7 @@ public:
    * polynomials is not efficient, because then each point value of the
    * underlying (one-dimensional) polynomials is (unnecessarily) computed
    * several times.  Instead use the <tt>compute</tt> function, see above,
-   * with <tt>values.size()==n_tensor_pols</tt> to get the point values of all
+   * with <tt>values.size()==this->n()</tt> to get the point values of all
    * tensor polynomials all at once and in a much more efficient way.
    */
   double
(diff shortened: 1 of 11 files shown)


include/deal.II/base/tensor_product_polynomials.h
```diff
@@ -62,7 +63,7 @@ DEAL_II_NAMESPACE_OPEN
  * 2003
  */
 template <int dim, typename PolynomialType = Polynomials::Polynomial<double>>
-class TensorProductPolynomials
+class TensorProductPolynomials : ScalarPolynomialsBase<dim>
```
bangerth: You'll have to do the derivation as `public:`

include/deal.II/base/tensor_product_polynomials.h
```diff
@@ -259,7 +272,7 @@ class TensorProductPolynomials
  * @author Wolfgang Bangerth 2003
  */
 template <int dim>
-class AnisotropicPolynomials
+class AnisotropicPolynomials : ScalarPolynomialsBase<dim>
```
bangerth: same here

source/base/tensor_product_polynomials.cc
```diff
@@ -405,13 +406,26 @@ TensorProductPolynomials<dim, PolynomialType>::evaluate(
...
 
 
+template <int dim, typename PolynomialType>
+std::unique_ptr<ScalarPolynomialsBase<dim>>
+TensorProductPolynomials<dim, PolynomialType>::clone() const
+{
+  Assert(false, ExcNotImplemented());
+
+  return 0; // std_cxx14::make_unique<TensorProductPolynomials<dim,
+            // PolynomialType>>(*this);
```
bangerth: Do you mean on leaving this as a TODO for later?

source/base/tensor_product_polynomials.cc
```diff
@@ -700,6 +714,16 @@ AnisotropicPolynomials<dim>::get_n_tensor_pols(
 }
 
 
+template <int dim>
+std::unique_ptr<ScalarPolynomialsBase<dim>>
+AnisotropicPolynomials<dim>::clone() const
+{
+  Assert(false, ExcNotImplemented());
+
+  return 0; // std_cxx14::make_unique<AnisotropicPolynomials<dim>>(*this);
```
bangerth: same here?

include/deal.II/base/tensor_product_polynomials.h
GrahamBenHarper: I just saw that on my end! I was thinking "what do you mean the base class is inaccessible?" :P

source/base/tensor_product_polynomials.cc
GrahamBenHarper: I was running into an error with this. Thanks for reminding me about this. I need to fix it but I'm unsure how.

include/deal.II/base/tensor_product_polynomials.h
GrahamBenHarper: Got it!

source/base/tensor_product_polynomials.cc
GrahamBenHarper: yes same here. I'm rerunning a build now to see what the errors looked like

source/base/tensor_product_polynomials.cc
bangerth: Post the error and we'll help you :-)

GrahamBenHarper: Oh, it looks like I don't have that error now that the base class is accessible. That's nice!

GrahamBenHarper:
The only thing that's left is to decide what `degree` should be assigned to the base class since `TensorProductPolynomials` and `AnisotropicPolynomials` don't originally have a degree member variable. Also the indentation issues, but I'll do those after I run tests on my end and check if any tests need modifications to call `this->n()` instead of `this->n_tensor_pols`.

I could take the list of `Polynomials::Polynomial` and assign a degree based on the `degree()` value. For example, `AnisotropicPolynomials` degree would be the max of `pols[i].degree()`? I think I would need a lambda expression for that.

GrahamBenHarper:
Or an alternative would be to define a static function and do something similar to the way the `n_polys` is computed on construction through a static function call.

include/deal.II/base/tensor_product_polynomials_bubbles.h
```diff
@@ -176,7 +176,7 @@ template <int dim>
 inline unsigned int
 TensorProductPolynomialsBubbles<dim>::n() const
 {
-  return this->n_tensor_pols + dim;
+  return TensorProductPolynomials<dim>::n() + dim;
```
bangerth: Oh, this is awkward -- you are in essence overloading a member function from the base class but giving it a different meaning. This suggests that the "is-a" relationship of `TensorProductPolynomialsBubbles` and `TensorProductPolynomials` classes that is usually implied when you derive from a base class is not the case here, but that maybe a "has-a" relationship of using `TensorProductPolynomials` as a member variable would be more appropriate?

source/base/tensor_product_polynomials.cc
```diff
@@ -700,6 +704,16 @@ AnisotropicPolynomials<dim>::get_n_tensor_pols(
 }
 
 
+template <int dim>
+std::unique_ptr<ScalarPolynomialsBase<dim>>
+AnisotropicPolynomials<dim>::clone() const
+{
+  Assert(false, ExcNotImplemented());
+
+  return std_cxx14::make_unique<AnisotropicPolynomials<dim>>(*this);
```
bangerth: why is it not implemented? Does the implementation in the last line not work?

include/deal.II/base/tensor_product_polynomials_bubbles.h
GrahamBenHarper: Well the two commits marked as (broken) were originally going to try to replace all instances of `n_tensor_pols` in the derived class with `this->n()`; however, the number of polynomials for the derived class is actually `n_tensor_pols + dim`. I ran into a bunch of trouble with this and ended up reverting four files then redoing the broken commits.

At the moment, I have replaced all references to `n_tensor_pols` with the `TensorProductPolynomials::n()` and left the derived `n()` as `n_tensor_pols + dim`.

Should I instead just make a `n_tensor_pols` const member of this derived class to make this easier?

source/base/tensor_product_polynomials.cc
GrahamBenHarper: Great question. I forgot to remove that assert while addressing your previous comments!

GrahamBenHarper:
Every `ScalarPolynomialsBase<dim>` has a `const` member `polynomial_degree`; however, `TensorProductPolynomials<dim>` never defines a member called `degree`. Thus, I'm unsure what value to pass to the base class constructor.

include/deal.II/base/tensor_product_polynomials_bubbles.h
GrahamBenHarper: I think giving the derived class a `n_tensor_pols` member might be an easier way to preserve the "is-a" relationship instead of turning it into a "has-a" relationship.

GrahamBenHarper: Oh, but the base constructor is always called without giving it some notion of the additional functions, be them the +1 from const or +dim from bubbles. That's why this is a little tricky.

My goal with this PR is to preserve the original setup of this class without drastically changing it, since I'm only trying to add that base class.

bangerth: HM, but that doesn't sound right. What I'm saying is that if the number of polynomials is `TPP::n()+dim`, then someone has lied to the base class when calling the base class constructor. Nothing good can come of this: If you pass on object of this type via a reference-to-base, and the place where that argument is used calls `n()` on the object they have, then they will return the wrong number for the number of polynomials because this calls the `n()` function of the *base* class, not the *derived* class. 

The original sin here is to say that the `TPPBubbles` object *is a* TBB object, when in reality it is of course not: It's a polynomial space alright, but the shape functions are not constructed by taking tensor products. So the derivation may have seemed convenient at the time, but incorrect. That's what should be fixed first.

If that's a step too far for you, let me know and I can take care of it.

GrahamBenHarper: I see. So perhaps we should have
```
template <int dim>
TensorProductPolynomialsBubbles : public ScalarPolynomialsBase<dim>
{
  ...
private:
  const TensorProductPolynomials<dim> tensor_polys;
}
```
and then we initialize this in the constructor with something like this?
```
TensorProductPolynomialsBubbles<dim>::TensorProductPolynomialsBubbles(const std::vector<Pol> &pols)
: ScalarPolynomialsBase<dim>(???, num_tensor_pols + dim)
, tensor_polys(pols)
```
I could handle that and make it a separate PR if that looks right to you.

GrahamBenHarper: That particular implementation may also require `TensorProductPolynomials` and `TensorProductPolynomialsBubbles` to be friend classes?

bangerth: That's exactly what I was thinking of. This kind of derivation makes a lot more sense to me.

If you need something to be `friend`, mark it as such. It's often possible to work around this (and that also makes the code cleaner) but sometimes it just makes sense to do it with a `friend` declaration.

GrahamBenHarper: Okay great! I was thinking of using a `friend` since `TensorProductPolynomials` has protected members `polynomials`, `index_map`, and `index_map_inverse`. This way I can rely on those instead of storing them a second time in `TensorProductPolynomialsBubbles` since it "has-a" `TensorProductPolynomials`.

Note to self: change those protect members to private members

GrahamBenHarper:
Whoops it looks like I messed this up while rebasing onto master and the other PR... how do I fix this?

Anyway, it looks like the `base` and `fe` tests are fine at this state.

bangerth:
Do
```
  git checkout master
  git pull upstream master
  git checkout <...thisbranch...>
  git rebase master
```
and I think the additional commits should all go away.

GrahamBenHarper:
There we go! Rebase was definitely not happy with conflicts from #8742 but this should be working a little better now.

PR #8724, opened 2019-09-10
Generalize an if-statement for all parallel triangulations

peterrum:
FYI @bangerth 

Should I replace in this PR also all other `dynamic_cast<const parallel::distributed::Triangulation<>>`s?

### DIFF
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -1293,7 +1293,7 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
   // only if this is a sequential
   // triangulation. it doesn't work
   // correctly yet if it is parallel
-  if (dynamic_cast<const parallel::distributed::Triangulation<dim, spacedim> *>(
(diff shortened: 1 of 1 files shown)


source/dofs/dof_handler.cc
```diff
@@ -1293,8 +1293,8 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
   // only if this is a sequential
   // triangulation. it doesn't work
   // correctly yet if it is parallel
-  if (dynamic_cast<const parallel::distributed::Triangulation<dim, spacedim> *>(
-        &*tria) == nullptr)
+  if (dynamic_cast<const parallel::DistributedTriangulationBase<dim, spacedim>
+                     *>(&*tria) == nullptr)
```
bangerth: Thinking about this, the comment suggests that it only works for sequential triangulations. That would also rule out shared triangulations. Can you change this to the common base class for *all* parallel triangulations?

PR #8728, opened 2019-09-10
Embed vector operations in MatrixFree cell loops

kronbichler:
This PR enables `MatrixFree::cell_loop` to schedule an operation before the cell operation first touches an unknown, and another operation after it last touches an unknown. The goal is to do operations on vectors close to the access to the vectors in matrix-free loops to improve caching. This can be used e.g. for conjugate gradient solvers with simple preconditioners, or with PreconditionChebyshev.

For conjugate gradients with a diagonal preconditioners, one can get very nice performance from this setup (using some special variant that I still need to commit in a later PR in the coming weeks):
locality_loops

For large sizes (many dofs) at the right of the plot, I can improve performance of a conjugate gradient solver with matrix-free implementation (and special geometry approximations) by more than a factor of 2, compare the "fully merged" line with the "plain CG" line in the plot, where the former uses the proposed loop facilities. Now this is of course application-dependent, i.e., the cost of the matrix-vector products compared to access to vectors. The motivation of this contribution is that I have several cases where the matrix-vector product no longer dominates, so we need an algorithm-level optimization.

I will later also add one or two tests doing some generic stuff like update a vector before/after to show how this works, but the code is ready. I have some code [in a private branch](https://github.com/kronbichler/ceed_benchmarks_dealii/blob/experimental_merge_vector/common_code/poisson_operator.h#L361-L397) somewhere that generated the above graph, and work towards that in several PRs.

### DIFF
--- a/include/deal.II/matrix_free/dof_info.templates.h
+++ b/include/deal.II/matrix_free/dof_info.templates.h
@@ -1214,7 +1214,10 @@ namespace internal
       const unsigned int n_components = start_components.back();
       const unsigned int n_dofs       = vector_partitioner->local_size() +
                                   vector_partitioner->n_ghost_indices();
-      std::vector<unsigned int> touched_by(
+      std::vector<unsigned int> touched_first_by(
+        (n_dofs + chunk_size_zero_vector - 1) / chunk_size_zero_vector,
+        numbers::invalid_unsigned_int);
+      std::vector<unsigned int> touched_last_by(
         (n_dofs + chunk_size_zero_vector - 1) / chunk_size_zero_vector,
         numbers::invalid_unsigned_int);
       for (unsigned int part = 0;
@@ -1238,8 +1241,10 @@ namespace internal
                   {
                     const unsigned int myindex =
                       dof_indices[it] / chunk_size_zero_vector;
-                    if (touched_by[myindex] == numbers::invalid_unsigned_int)
-                      touched_by[myindex] = chunk;
+                    if (touched_first_by[myindex] ==
+                        numbers::invalid_unsigned_int)
+                      touched_first_by[myindex] = chunk;
+                    touched_last_by[myindex] = chunk;
                   }
               }
             if (faces.size() > 0)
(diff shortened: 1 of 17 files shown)


peterrum:
FYI: @arosch

include/deal.II/matrix_free/dof_info.templates.h
```diff
@@ -1217,6 +1217,9 @@ namespace internal
       std::vector<unsigned int> touched_by(
```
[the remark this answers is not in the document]
Rombur: That's a good idea

include/deal.II/matrix_free/dof_info.templates.h
```diff
@@ -1262,39 +1266,102 @@ namespace internal
...
         if (index == numbers::invalid_unsigned_int)
-          index = task_info.cell_partition_data.back() - 1;
+          index =
+            task_info
+              .partition_row_index[task_info.partition_row_index.size() - 2] -
+            1;
+
+      // lambda to convert from the map to a vector indexed by the
+      // `range_list_index` for straight access
+      auto convert_map_to_vector =
```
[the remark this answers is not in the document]
Rombur: So if I understand this function correctly, you transform a map of (partition, vector of dofs) to a vector of range and a vector of indices to know how many ranges belong to each partition. So unlike what the name and the comment imply you go from one map to multiple vectors not one, is that right?

include/deal.II/matrix_free/dof_info.templates.h
```diff
@@ -1262,39 +1266,102 @@ namespace internal
...
                   }
           }
+
       // ensure that all indices are touched at least during the last round
       for (auto &index : touched_by)
         if (index == numbers::invalid_unsigned_int)
-          index = task_info.cell_partition_data.back() - 1;
+          index =
+            task_info
+              .partition_row_index[task_info.partition_row_index.size() - 2] -
```
Rombur: The `-2` is because you are interested in the first index?

include/deal.II/matrix_free/dof_info.templates.h
```diff
@@ -1262,39 +1266,102 @@ namespace internal
...
+      touched_by.resize(
+        (vector_partitioner->local_size() + chunk_size_zero_vector - 1) /
+        chunk_size_zero_vector);
+      for (auto it : vector_partitioner->import_indices())
+        for (unsigned int i = it.first; i < it.second; ++i)
+          touched_by[i / chunk_size_zero_vector] = n_partitions;
+      std::map<unsigned int, std::vector<unsigned int>> chunk_must_do_pre;
+      for (unsigned int i = 0; i < touched_by.size(); ++i)
+        chunk_must_do_pre[touched_by[i]].push_back(i);
+      convert_map_to_vector(n_partitions + 1,
```
Rombur: Why is this `n_partitions +1` while for `chunk_must_zero_vector` the size is `n_partitions`?

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -802,6 +802,138 @@ class MatrixFree : public Subscriptor
...
+   *
+   * @param owning_class The object which provides the `cell_operation`
+   * call. To be compatible with this interface, the class must allow to call
+   * `owning_class->cell_operation(...)`.
+   *
+   * @param dst Destination vector holding the result. If the vector is of
+   * type LinearAlgebra::distributed::Vector (or composite objects thereof
+   * such as LinearAlgebra::distributed::BlockVector), the loop calls
+   * LinearAlgebra::distributed::Vector::compress() at the end of the call
+   * internally.
```
Rombur: So if I use a Epetra vector, I have to call `compress()` myself? Why is there a difference with `LA::d::Vector`?

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -802,6 +802,138 @@ class MatrixFree : public Subscriptor
...
+   * operation on entries of the `src` and `dst` vectors (or other vectors)
+   * after the operation on cells last touches a particular DoF according to
+   * the general description in the text above. This function is passed a
+   * range of the locally owned degrees of freedom on the selected
+   * `dof_handler_index_pre_post` (in MPI-local numbering).
+   *
+   * @param dof_handler_index_pre_post Set which of the underlying DoFHandler
+   * objects passed at initialization should be selected by the
+   * `operator_before_loop` and `operation_after_loop`. Defaults to the
+   * `dof_handler_index` 0.
```
Rombur: I don't understand this parameter.

include/deal.II/matrix_free/dof_info.templates.h
kronbichler: The last index is associated to cells corresponding to faces that are not worked on locally, which are not part of cell/face loops but needed in other instances.

include/deal.II/matrix_free/dof_info.templates.h
kronbichler: It is because we need to schedule work on the source *before* the ghost exchange starts, so use another index that is not used within the cell loop. Same for the operation after the loop: We need to indicate that it is done after everything else. (The actual index value does not matter, so I use the same in both cases.)

include/deal.II/matrix_free/matrix_free.h
kronbichler: The infrastructure does currently not support external vectors anyway because the indices are local in matrix-free, whereas we only have access to global indices in Epetra or PETSc vectors. I will add a comment that those vectors are not supported in parallel, so it also makes no sense to call a method on them.

include/deal.II/matrix_free/matrix_free.h
kronbichler: MatrixFree can be initialized with a list of `DoFHandler` objects, e.g. one for velocity and one for pressure in a Stokes system. However, the call-back only gives me a range on one vector. The alternative to choosing which `DoFHandler` to provide a vector range for would be to return ranges to all `DoFHandlers` (by separate call-backs). In that case, the user would be given a third index, say 'dof_handler_index', specifying to which range that belongs. However, that feels like a bit of a waste because a user will not often run block systems through a single loop (otherwise, I would put the components in the same DoFHandler and vector). What do you think?

kronbichler:
I fixed the review comments and added test cases.

PR #8732, opened 2019-09-10
Move the call to update_number_cache() after all other calls.

bangerth:
Conceptually, this function (as its name suggests) just updates a cache. We should
call it after all other functions that change things in the triangulation have
been called, even if the order does not matter in a particular context. This just
makes reading the code simpler.


peterrum:
I'll do the same in `p:f:t` once the tests have passed here.

bangerth:
@peterrum -- you're *on it*, I see ;-)

Issue #8735, opened 2019-09-11
doxygen: avoid Python dependency for setting canonical links


fvanmaele:
`set_canonical_doxygen.py` may lead to errors when using certain Python3 versions (`UnicodeDecodeError` on `/usr/bin/python3.6`, see issue #8714). 

Having the script in Perl (as most of the other `doxygen` scripts) avoids these issues and the additional Python dependency.

----
While rewriting the script, I noticed that `# Do not add canonical link twice` does not actually work, since the `canonical_link_added` is always set to `False` inside the loop. This should be fixed in the follow-up PR to this issue.

fvanmaele:
PR #8736

drwells:
Fixed by #8736.

PR #8736, opened 2019-09-11
doxygen: rewrite set_canonical_doxygen.py in Perl

fvanmaele:
Fixes issue #8735

---
The Perl script ends up slightly longer than the Python version. Using `Tie::File` would add brevity, it however has [issues with `UTF-8` encoded files](https://www.nntp.perl.org/group/perl.perl5.porters/2014/05/msg215836.html). Most `doxygen` files are `ascii` encoded, but on my system (Ubuntu 18.04) there were some `utf8` encoded as well.

The Perl version also correctly appends the canonical link precisely once; the Python version would not do this, due to setting the `canonical_link_added` inside the loop to `False` at every iteration.

I wasn't sure on style guidelines for Perl scripts, and as far as I know `clang-format` only applies to `C++` source files. As such, I tried following the style of the other Perl scripts in the `doc/doxygen/scripts` directory.

---
This is a draft pull request since I'm still building and comparing all doxygen created files between `set_canonical_doxygen.py` and `set_canonical_doxygen.pl` (and some remarks pointed out below).


doc/doxygen/scripts/set_canonical_doxygen.pl
```diff
@@ -0,0 +1,93 @@
...
+	  {
+	       push(@html, $File::Find::name);
+	  }
+     }, "./");
+chomp(@html);
+
+for my $filename (@html) {
+     if ($filename eq "./header.html")
+     {
+	  # there is no relevant content in the header
```
fvanmaele: I wasn't sure on what to add here - the original script uses `pass`. Is the intent to not modify `header.html`?

doc/doxygen/scripts/set_canonical_doxygen.pl
```diff
@@ -0,0 +1,93 @@
...
+##
+## ---------------------------------------------------------------------
+
+# This script sets the canonical webpage for all the webpages of the
+# doxygen documentation. For more information see
+# https://en.wikipedia.org/wiki/Canonical_link_element
+#
+# This script is invoked by CMake. To add the canonical link to the webpages of
+# the deal.ii repository you can use the script
+# contrib/utilities/set_canonical_webpages.py
```
fvanmaele: This script seems no longer present in the source tree (unless I'm missing something).

doc/doxygen/scripts/set_canonical_doxygen.pl
drwells: Yes, but it really should have said ‘continue’. Since, if I recall correctly, there is no match, using the wrong one didn’t matter.

fvanmaele: Added `next` in 33c9950. Alternatively, `if ($_ =~ /\.html$/ and $_ ne "header.html")` could be used, which is shorter (but perhaps less clear).

fvanmaele:
I've compared two copies of the `dealii` directory, one with this PR applied, and another from master (`dealii-orig`). 

Apart from header.html, the generated files are identical between `set_canonical_doxygen.py` and `set_canonical_doxygen.pl`. ([Full diff with path changes](https://gist.githubusercontent.com/fvanmaele/68d078f445507946da4fa015fe205447/raw/2d0d420f62835453812df4c9fd619d62cc2bb881/html.diff).)

drwells:
I am not fluent in perl but the script looks reasonable and produces the correct output. Since this is just a doc change lets merge.

fvanmaele:
Thanks!

PR #8739, opened 2019-09-12
add functions to access svd u and vt variables

MFraters:
This pull request adds two functions to the LAPACKFullMatrix class to give read access to the private variables `svd_u` and `svd_vt` generated by the `compute_svd` function.

### DIFF
--- a/include/deal.II/lac/lapack_full_matrix.h
+++ b/include/deal.II/lac/lapack_full_matrix.h
@@ -873,14 +873,14 @@ public:
   singular_value(const size_type i) const;
 
   /**
-   * Retrieves #svd_u matrix after compute_svd() or compute_inverse_svd() was
+   * Retrieve the matrix #svd_u after compute_svd() or compute_inverse_svd() was
    * called.
    */
   inline const LAPACKFullMatrix<number> &
   get_svd_u() const;
 
   /**
-   * Retrieves #svd_vt matrix after compute_svd() or compute_inverse_svd() was
+   * Retrieve the matrix #svd_vt after compute_svd() or compute_inverse_svd() was
    * called.
    */
   inline const LAPACKFullMatrix<number>&
--- a/include/deal.II/lac/lapack_full_matrix.h
+++ b/include/deal.II/lac/lapack_full_matrix.h
@@ -815,7 +815,17 @@ public:
    *
    * Requires that the #state is LAPACKSupport::matrix, fills the data members
    * #wr, #svd_u, and #svd_vt, and leaves the object in the #state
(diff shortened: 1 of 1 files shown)


tjhei:
Would you mind adding a little bit of info what u and vt are to the documentation of compute_svd? Maybe you can steal from the LAPACK help text:
```
The SVD is written

           A = U * SIGMA * transpose(V)

      where SIGMA is an M-by-N matrix which is zero except for its
      min(m,n) diagonal elements, U is an M-by-M orthogonal
      matrix, and V is an N-by-N orthogonal matrix.  The diagonal
      elements of SIGMA are the singular values of A; they are
      real and non-negative, and are returned in descending order.
      The first min(m,n) columns of U and V are the left and right
      singular vectors of A.

      Note that the routine returns V**T, not V.
```

include/deal.II/lac/lapack_full_matrix.h
```diff
@@ -862,6 +862,20 @@ class LAPACKFullMatrix : public TransposeTable<number>
   number
   singular_value(const size_type i) const;
 
+  /**
+   * Retrieves #svd_u matrix after compute_svd() or compute_inverse_svd() was
+   * called.
+   */
+  inline LAPACKFullMatrix<number>
```
tjhei: are you sure you want to return the matrix by value? I think a ``const &`` would be more appropriate,

MFraters: yes, you are right. I will change it to a `const &`

MFraters:
> Would you mind adding a little bit of info what u and vt are to the documentation of compute_svd? Maybe you can steal from the LAPACK help text:
> 
> ```
> The SVD is written
> 
>            A = U * SIGMA * transpose(V)
> 
>       where SIGMA is an M-by-N matrix which is zero except for its
>       min(m,n) diagonal elements, U is an M-by-M orthogonal
>       matrix, and V is an N-by-N orthogonal matrix.  The diagonal
>       elements of SIGMA are the singular values of A; they are
>       real and non-negative, and are returned in descending order.
>       The first min(m,n) columns of U and V are the left and right
>       singular vectors of A.
> 
>       Note that the routine returns V**T, not V.
> ```

Sure, I have added a bit more documentation based on that text and a bit of wikipedia :) Since I am not too familiar with the SVD, please read closely whether I made mistakes in my interpretation of the detail of the variables.

I tried to use `make indent` after this, but it gave an error: `invalid author 'MFraters' without firstname and lastname`. I don't know why my github name would have anything to do with indenting :s

Issue #8740, opened 2019-09-12
CellDataTransferBuffer in dof handler is super slow


kronbichler:
When re-evaluating #8293, I discovered that `DoFHandler::distribute_mg_dofs()` takes an unproportional amount of time. After putting some timers, I discovered the culprit: `CellDataTransferBuffer` and in particular the `pack_data` function.

When running on 50k MPI ranks (to make sure I do not underestimate the cost of MPI communication) with 270b dofs, the two `communicate_mg_ghost_cells` phases take 12.3 seconds with the current setting, and the full `distribute_mg_dofs` takes 13.7 seconds. When I change [here to `gzip::best_speed`](https://github.com/dealii/dealii/blame/master/source/dofs/dof_handler_policy.cc#L4312), the ghost communication takes only 0.75 seconds and the full `distribute_mg_dofs` takes 1.65 seconds. Looking at the code and the hardware properties, I am sure than one could go even lower by simply sending the raw data (dof indices and quadrants). Networks are slower than access to main memory, but not by much. Thus, fiddling around with data many times can easily eat any gain by saving network traffic. On top of that, the boost serialization infrastructure apart from the compression also scores pretty low on my performance ranking. Removing the boost stuff altogether would even simplify code.

These observations are very much in line with #8513. There is a reason why we do not compress data in the MPI ghost exchange for vectors - compression is at least 10 times slower than the network of decent machines, measured in MB/s, so there is no chance to win here.

However, I do not feel for modifying the sends now, so I will probably just go the `best_speed` route and probably never notice this again, unless someone explicitly wants me to ditch the slow boost infrastructure from this file.

PR #8747, opened 2019-09-13
Change gzip compression level to best_speed

kronbichler:
This change addresses one of three places where we use gzip. The second one, discussed in #8740, will be tackled by an upcoming PR. The third one in [Utilities::pack()](https://github.com/dealii/dealii/blob/master/include/deal.II/base/utilities.h#L1214) was already changed in #8513 to disable compression altogether. To reduce impact on the test suite, I have not touched it, but I can if one insists. From my experiments in #8740 for the `dof_handler_policy.cc` case the difference between `default_compression` and `best_speed` is maybe 20-50%, and `best_speed` in that case was around 2x slower than no compression at all on 50k MPI ranks (including the time to actually send the data).

### DIFF
--- a/source/dofs/number_cache.cc
+++ b/source/dofs/number_cache.cc
@@ -153,7 +153,7 @@ namespace internal
                 boost::iostreams::filtering_ostream out;
                 out.push(boost::iostreams::gzip_compressor(
                   boost::iostreams::gzip_params(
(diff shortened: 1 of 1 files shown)


drwells:
Its been my experience that `best_speed` is preferable in most circumstances, so :+1:

PR #8749, opened 2019-09-13
Improve performance of DoFHandler::distribute_mg_dofs

### LINKED ISSUE #8740: CellDataTransferBuffer in dof handler is super slow

kronbichler:
Fixes #8740.

Even though I though in #8740 that I would simply change the compression level, I decided that it was worth to delete around 150 lines of code by sending the uncompressed data. So this is a double gain - simpler code and faster implementation.

While there, I also used a slightly cheaper MPI routine, `MPI_Bcast` rather than `MPI_Allreduce` for counting the global number of dofs since we have that information on one processor already.

### DIFF
--- a/source/dofs/dof_handler_policy.cc
+++ b/source/dofs/dof_handler_policy.cc
@@ -5152,16 +5152,29 @@ namespace internal
             //* 3. communicate local dofcount and shift ids to make
             // them unique
             dealii::types::global_dof_index my_shift = 0;
-            const int                       ierr =
-              MPI_Exscan(DEAL_II_MPI_CONST_CAST(
-                           &level_number_cache.n_locally_owned_dofs),
-                         &my_shift,
-                         1,
-                         DEAL_II_DOF_INDEX_MPI_TYPE,
-                         MPI_SUM,
-                         triangulation->get_communicator());
+            int ierr = MPI_Exscan(DEAL_II_MPI_CONST_CAST(
+                                    &level_number_cache.n_locally_owned_dofs),
+                                  &my_shift,
+                                  1,
+                                  DEAL_II_DOF_INDEX_MPI_TYPE,
+                                  MPI_SUM,
+                                  triangulation->get_communicator());
             AssertThrowMPI(ierr);
 
+            // The last processor knows about the total number of dofs, so we
+            // can use a cheaper broadcast rather than an MPI_Allreduce via
+            // MPI::sum().
+            level_number_cache.n_global_dofs =
+              my_shift + level_number_cache.n_locally_owned_dofs;
+            ierr = MPI_Bcast(&level_number_cache.n_global_dofs,
+                             1,
+                             DEAL_II_DOF_INDEX_MPI_TYPE,
+                             Utilities::MPI::n_mpi_processes(
+                               triangulation->get_communicator()) -
+                               1,
+                             triangulation->get_communicator());
+
+            // shift indices
             for (types::global_dof_index &index : renumbering)
               if (index != numbers::invalid_dof_index)
                 index += my_shift;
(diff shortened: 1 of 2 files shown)


peterrum:
> /jenkins/workspace/dealii_PR-8749/source/dofs/dof_handler_policy.cc:5204:18: error: assignment of read-only variable ‘ierr’
>             ierr = MPI_Bcast(&level_number_cache.n_global_dofs,

This is why, I do not think it is a good idea that `ierr` is reused ;)

source/dofs/dof_handler_policy.cc
```diff
@@ -5317,16 +5186,29 @@ namespace internal
...
+                                  &my_shift,
+                                  1,
+                                  DEAL_II_DOF_INDEX_MPI_TYPE,
+                                  MPI_SUM,
+                                  triangulation->get_communicator());
             AssertThrowMPI(ierr);
 
+            // The last processor knows about the total number of dofs, so we
+            // can use a cheaper broadcast rather than all MPI_Allreduce via
+            // MPI::sum().
```
tjhei: clever!

source/dofs/dof_handler_policy.cc
```diff
@@ -4570,38 +4437,34 @@ namespace internal
...
-                                         &requests[idx]);
-              AssertThrowMPI(ierr);
-            }
+          std::vector<MPI_Request> requests(level_ghost_owners.size());
+          {
+            unsigned int idx = 0;
+            for (const auto &it : neighbor_cell_list)
+              {
+                // send the data in terms of the tree indices and the quadrant
+                // data. the data will be around till the packet has been
```
tjhei: this comment is a bit confusing, as this is now all stored in ``quadrants``.

source/dofs/dof_handler_policy.cc
```diff
@@ -4235,130 +4225,11 @@ namespace internal
         template <int dim>
         struct CellDataTransferBuffer
         {
-          std::vector<unsigned int> tree_indices;
-          std::vector<typename dealii::internal::p4est::types<dim>::quadrant>
+          std::vector<
```
tjhei: What is the use of this struct now? Either we should put some of the logic inside it or just remove it. We don't communicate the whole struct anymore anyways, right?

kronbichler: You are right, we could simply remove it. Let me try it (though not immediately).

kronbichler: I fixed it now, indeed the code is even cleaner.

kronbichler:
I ran all tests containing `multigrid`, `mg`, or `mpi` in the name and they are all passing.

PR #8746, opened 2019-09-13
Add Boost serialization function to CellData

peterrum:
This PR is part of the effort to introduce the new `parallel::fullydistributed::Triangulation` (see #8558). This triangulation will be optionally created in two steps: 
1. create the `p:f:ConstructionData` by a set of master processes and 
2. distributed it to the other processes. 

To pack the `ConstrcutionData`, I intend to use the Boost serialization capabilities. Since `ConstructionData` contains a list of `dealii::CellData`, this PR adds the `serialization`-function to `dealii::CellData`.

~~I think no test is needed.~~

PR #8751, opened 2019-09-13
Move operator== from a test to the CellData class itself.

bangerth:
This has the advantage that if that class ever gains a new member variable,
it is much easier to update the code for operator== since it's obvious
that one has to update it. If the code remained in the test, we would almost
certainly forget to update it.

Follow-up to #8746. @peterrum -- FYI.

### DIFF
--- a/source/grid/tria.cc
+++ b/source/grid/tria.cc
@@ -43,6 +43,40 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+
+template <int structdim>
+CellData<structdim>::CellData()
+  : material_id(0)
+  , manifold_id(numbers::flat_manifold_id)
+{
+  for (unsigned int i = 0; i < GeometryInfo<structdim>::vertices_per_cell; ++i)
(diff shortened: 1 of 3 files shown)


source/grid/tria.cc
```diff
@@ -43,6 +43,40 @@
...
 DEAL_II_NAMESPACE_OPEN
 
+
+template <int structdim>
+CellData<structdim>::CellData()
+  : material_id(0)
+  , manifold_id(numbers::flat_manifold_id)
+{
+  for (unsigned int i = 0; i < GeometryInfo<structdim>::vertices_per_cell; ++i)
+    vertices[i] = numbers::invalid_unsigned_int;
```
masterleinad: You could use `std::fill` here.

bangerth: Sure. `vertices` is a C-style array, so I can't use `vertices.begin()`, but `std::begin(vertices)` works.

bangerth:
Change made as requested. That actually added a line, but I agree that it's easier to read.

PR #8754, opened 2019-09-14
Disable compression in ghost exchange of DoFHandler

kronbichler:
Similarly to #8749, it is also better to not compress the ghost indices for DoFHandler. As observed in #8747, we already used the faster `default_compression` in `Utilities::pack`, but it is still slower than no compression at all. As also shown in #8513, the packing in terms of boost serialization still induces some overhead, but due to the generality of that function I refrain from further changes which would make the code more complicated to read.

To fix the case, I added a parameter to `GridTools::exchange_cell_data_to_ghosts` to enable/disable compression. Since I believe no compression is usually faster, I disabled it, but this is up to discussion with @bangerth, @marcfehling, @peterrum and @gassmoeller.

### DIFF
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -3892,7 +3892,7 @@ namespace GridTools
         // pack all the data into the buffer for this recipient and send it.
         // keep data around till we can make sure that the packet has been
         // received
-        sendbuffers[idx] = Utilities::pack(data);
+        sendbuffers[idx] = Utilities::pack(data, /*enable_compression*/ false);
         const int ierr   = MPI_Isend(sendbuffers[idx].data(),
                                    sendbuffers[idx].size(),
                                    MPI_BYTE,
@@ -3928,7 +3928,8 @@ namespace GridTools
         AssertThrowMPI(ierr);
 
         auto cellinfo =
-          Utilities::unpack<CellDataTransferBuffer<dim, DataType>>(receive);
+          Utilities::unpack<CellDataTransferBuffer<dim, DataType>>(
+            receive, /*enable_compression*/ false);
 
         DataType *data = cellinfo.data.data();
         for (unsigned int c = 0; c < cellinfo.cell_ids.size(); ++c, ++data)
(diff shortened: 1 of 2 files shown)


include/deal.II/grid/grid_tools.h
```diff
@@ -3793,7 +3798,8 @@ namespace GridTools
     const std::function<std_cxx17::optional<DataType>(
       const typename MeshType::active_cell_iterator &)> &pack,
     const std::function<void(const typename MeshType::active_cell_iterator &,
-                             const DataType &)> &        unpack)
+                             const DataType &)> &        unpack,
+    const bool                                           enable_compression)
   {
 #    ifndef DEAL_II_WITH_MPI
     (void)mesh;
```
peterrum: Add this here:
```cpp
(void)enable_compression;
```

source/dofs/dof_handler_policy.cc
```diff
@@ -4775,7 +4775,7 @@ namespace internal
 
           GridTools::exchange_cell_data_to_ghosts<
             std::vector<types::global_dof_index>,
-            DoFHandlerType>(dof_handler, pack, unpack);
+            DoFHandlerType>(dof_handler, pack, unpack, false);
```
tjhei: can you please add ``/*enable_compression*/ for the last argument?

tjhei:
> I disabled it, but this is up to discussion with @bangerth, @marcfehling, @peterrum and @gassmoeller.

You certainly have support from me here. But I will wait with merging to have someone from the list comment on it.

marcfehling:
We refrain from using compression since #8513. We even skip serialization if we know the data structure to send by providing optimized `pack/unpack` functions (e.g as introduced in #8641 and #8522).

In this case, I am fine with disabling compression as you proposed.

bangerth:
If you believe that compression is not useful here, then I see no reason to even offer it as an option. I'd just make that the only option.

kronbichler:
I believe compression is not useful when going over the network (unless it is done in hardware or MPI libraries, which is not our business), so indeed we can simply disable it. (On the other hand, it is useful when writing things to file, especially on parallel file systems.) Let me remove the option again and add a changelog.

tjhei:
There has been broad consensus for this change, so I will merge as is. Thanks, @kronbichler .

PR #8756, opened 2019-09-16
Fix a few typos in changelogs

kronbichler:
I noticed a few typos while tracking a recent change.


tjhei:
Thanks! Documentation only, so I will merge...

PR #8760, opened 2019-09-16
fixed inconsistent state in TimestepControl

rezarastak:
Calling `TimestepControl::parse_parameters()` is supposed to initialize the object. However, it does not update the internal variables such as `now_val`, `step_val`  etc, creating an inconsistent state. Calling the `restart()` method fixes this problem.


bangerth:
Yes, this looks alright.

/rebuild

PR #8761, opened 2019-09-16
Change TensorProductPolynomialsConst from "is-a" to "has-a" TensorProductPolynomials

GrahamBenHarper:
This is almost exactly the same as #8742 except for `TensorProductPolynomialsConst` instead.

Future work would involve modifying the classes to remove the need for friendship with `TensorProductPolynomials`. I'd like to push this through before #8723 so I may base that PR off this work and derive all 4 classes from `ScalarPolynomialsBase` in one PR.

All tests for `fe` and `base` are successful on my end.

### DIFF
--- a/include/deal.II/base/tensor_product_polynomials.h
+++ b/include/deal.II/base/tensor_product_polynomials.h
@@ -30,8 +30,13 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+// Forward declarations for friends
+// TODO: We may be able to modify these classes so they aren't
+// required to be friends
 template <int dim>
 class TensorProductPolynomialsBubbles;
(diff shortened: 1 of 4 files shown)


GrahamBenHarper:
On second thought, since it looks like #8723 has passed all tests, I can rebase this after #8723 merges. That may be easier.

bangerth:
Please rebase since there are currently some merge conflicts.

GrahamBenHarper:
All `base` and `fe` tests pass on my end!

GrahamBenHarper:
I'll add that as a comment in #8768. I'm not certain that the renumbering functionalities for TPPBubbles and TPPConst are used in the testsuite.

PR #8765, opened 2019-09-16
Avoid warnings as a result of an unused variable.

bangerth:
These were introduced in some of the recent patches addressing #1973. The warning only happens in release mode when the `Assert` that uses the variable is not evaluated.

@GrahamBenHarper -- FYI.


tjhei:
Would it make sense to make n_bubbles a (constexpr) member variable of the class instead?

GrahamBenHarper:
Thanks for catching this! Whoops- I did remove a `(void) n_bubbles` here and there because I didn't see the warnings in debug mode.

bangerth:
@tjhei: No you can't -- `n_bubbles` is (was) initialized this way:
```
const unsigned int n_bubbles     = ((q_degree <= 1) ? 1 : dim);
```
It's not a compile-time constant because `q_degree` is not.

tjhei:
> because `q_degree` is not.

ah, too bad.

PR #8766, opened 2019-09-16
Provide some explanation on 'virtual' and 'override'.

### LINKED ISSUE #8696: Step-4 documentation: Explanation of keywords virtual & override in RHS/BC template functions
Step-4 tutorial explains the idea of deriving the `RHS` and `BC` classes from the general `Function` class. However, there is a sudden mental jump for the reader upon encountering the actual `c++` code.

In particular, the line
`virtual double value(const Point<dim> & p, const unsigned int component=0) const override;` 

might be confusing to a newbie (it certainly is for me).  I don't know much about inheritance, but I can guess that the `virtual` and `override` have something to do with inheritance.  I know that this is a C++ question rather than a deal.ii question, however a general web search yields quite a bit of daunting information which can be off-putting for someone to quickly get through this tutorial (and not get stuck).

A one-liner explanation of why these additional keywords were suddenly introduced (with an optional link to further resources on the topic) may be considered if there is consensus.

PS:
I am only pointing this out because of the context in which I arrived at this juncture. So far, it has been a linear/smooth ride. There was even a detailed explanation of what templates are, in the form of an entire video lecture, in addition to the intro section. There is even a deal.ii FAQ section for templates. IMHO, templates seem conceptually easier to understand than inheritance. And, therefore, to help with the reading continuity, a line or two on inheritance is much appreciated. Perhaps the mailing list is more appropriate for this type of requests, but I am raising this here, since potentially this may involve an update to the tutorial.

bangerth:
Fixes #8696.

### DIFF
--- a/examples/step-4/step-4.cc
+++ b/examples/step-4/step-4.cc
@@ -102,11 +102,11 @@ private:
 // Function, which declares the common interface which all functions have to
 // follow. In particular, concrete classes have to overload the
 // <code>value</code> function, which takes a point in dim-dimensional space
-// as parameters and shall return the value at that point as a
+// as parameters and returns the value at that point as a
 // <code>double</code> variable.
 //
 // The <code>value</code> function takes a second argument, which we have here
-// named <code>component</code>: This is only meant for vector valued
+// named <code>component</code>: This is only meant for vector-valued
 // functions, where you may want to access a certain component of the vector
 // at the point <code>p</code>. However, our functions are scalar, so we need
 // not worry about this parameter and we will not use it in the implementation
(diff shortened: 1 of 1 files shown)


bangerth:
I took the opportunity in this patch to also remove the empty default constructor the classes derived from `Function<dim>`. The remainder of the tutorial programs are addressed in #8767.

examples/step-4/step-4.cc
```diff
@@ -148,17 +138,37 @@ template <int dim>
...
-
   virtual double value(const Point<dim> & p,
                        const unsigned int component = 0) const override;
 };
 
-
-
-// For this example, we choose as right hand side the function
+// If you are not familiar with what the keywords `virtual` and `override` in
+// function declarations above mean, you will probably want to take a look at
```
kronbichler:
```suggestion
// the function declarations above mean, you will probably want to take a look at
```

examples/step-4/step-4.cc
```diff
@@ -148,17 +138,37 @@ template <int dim>
...
+// it just says "this is how Function objects look like", but what kind of
+// function it actually is, is left to derived classes that implement
+// the `value()` function.
+//
+// Deriving one class from another is often called an "is-a" relationship
+// function. Here, the `RightHandSide` class "is a" Function class
+// because it implements the interface described by the Function base class.
+// (The actual implementation of the `value()` function is in the code block
+// below.) The `virtual` keyword then means "Yes, the
+// function here is one that is one that can be overridden by derived classes",
```
kronbichler:
```suggestion
// function here is one that can be overridden by derived classes",
```

examples/step-4/step-4.cc
```diff
@@ -148,17 +138,37 @@ template <int dim>
...
+// and the `override` keyword means "Yes, this is in fact a function we know
+// has been declared as part of the base class. The `override` keyword is not
```
kronbichler:
```suggestion
// has been declared as part of the base class". The `override` keyword is not
```

examples/step-4/step-4.cc
```diff
@@ -148,17 +138,37 @@ template <int dim>
...
+// strictly necessary, but is an insurance against typos: If we get the name
+// of the function or the type of one argument wrong, the compiler will warn
+// us by stating "You say that this function overrides one in a base class,
+// but I don't actually know any such function with this name and these
+// arguments.
```
kronbichler:
```suggestion
// arguments."
```

bangerth:
Thanks, @kronbichler -- I guess I should have proof-read that more carefully :-(

bangerth:
Incidentally, at one point you should add an entry to `doc/news/changes/major` when you're at a point where you think that it's worth announcing the new class!

PR #8769, opened 2019-09-16
Make DataPostprocessor work for complex-valued vector-valued problems.

bangerth:
This is the equivalent patch to #8648, but for vector-valued complex solutions. It fills the necessary fields of the `DataPostprocessor` input structures when the solution is both vector-valued and complex-valued. We just need to copy the real and imaginary parts to the correct place.

Part of #1894.

@dangars -- FYI.


bangerth:
I've added a second test that is a marginally edited version of the one that I believe I had originally gotten from @dangars (though I now no longer can't find where I had it from).

bangerth:
I found it: The test comes from #8657. It works with this patch. (There is still work in other areas, but at least it doesn't crash.)

tests/data_out/data_out_postprocessor_vector_complex_01.cc
```diff
@@ -0,0 +1,206 @@
...
+LaplaceProblem<dim>::output_results() const
+{
+  {
+    // Save the data in the log file
+    MyPostprocessor<dim> p;
+    DataOut<dim>         data_out;
+    data_out.attach_dof_handler(dof_handler);
+    data_out.add_data_vector(solution, p);
+    data_out.build_patches();
+    data_out.write_gnuplot(logfile);
```
dangars: I run this test and output the data  as `solution.vtu`. There is only one output which is `magnitudes_times_d`. This output has 3 components for x, y and z. There are no real and imaginary parts. Is this the expected behavior?

I used:
``` c++
std::ofstream output("solution.vtu");
data_out.write_vtu(output);
```

tests/data_out/data_out_postprocessor_vector_complex_01.cc
```diff
@@ -0,0 +1,206 @@
...
+      {
+        Assert(computed_quantities[q].size() == dim, ExcInternalError());
+
+        for (unsigned int d = 0; d < dim; ++d)
+          computed_quantities[q](d) =
+            input_data.solution_values[q](0) *
+              input_data.solution_values[q](0) +
+            input_data.solution_values[q](1) * input_data.solution_values[q](1);
+        AssertThrow(std::fabs(computed_quantities[q](0) - 1) < 1e-12,
+                    ExcInternalError());
```
dangars: Is `computed_quantities[q](idx)` supposed to be equal to `1` for `idx = 1,2,3`? If yes, wouldn't it make sense to put the assert inside the loop?
```c++
        for (unsigned int d = 0; d < dim; ++d){
          computed_quantities[q](d) =
            input_data.solution_values[q](0) *
              input_data.solution_values[q](0) +
            input_data.solution_values[q](1) * input_data.solution_values[q](1);
        AssertThrow(std::fabs(computed_quantities[q](d) - 1) < 1e-12,
                    ExcInternalError());
        }
```

bangerth:
On 9/17/19 11:31 AM, Daniel Garcia-Sanchez wrote:
> Is |computed_quantities[q](idx)| supposed to be equal to |1| for |idx = 
> 1,2,3|? If yes, wouldn't it make sense to put the assert inside the loop?
> 
>          for  (unsigned  int  d =0; d < dim; ++d){
>            computed_quantities[q](d) =
>              input_data.solution_values[q](0) *
>                input_data.solution_values[q](0) +
>              input_data.solution_values[q](1) * input_data.solution_values[q](1);
>          AssertThrow(std::fabs(computed_quantities[q](d) -1) <1e-12,
>                      ExcInternalError());
>          }

I don't know -- you wrote the test ;-)

dangars:
> I don't know -- you wrote the test ;-)

Hahahaha that is exactly what I thought after I pushed the button "comment".

I'm going to check if it makes sense to put the assert inside the loop!

dangars:
Yes, it does makes sense to put the assert inside the loop!

I wonder if it makes sense to put this test now inside this PR right. The output of the test should be `real = 1`, `imag=0`. So far it seems that the test only prints the real part.

I you update `deal.II` to support `vector` complex numbers by separating the real and the imaginary parts. Then the output of this test would be different and I guess that the loop:
``` c++
 for  (unsigned  int  d =0; d < dim; ++d){
```
should be replaced by
``` c++
 for  (unsigned  int  d =0; d < 2*dim; ++d){
```
Wouldn't it be better to merge this test once everything is done? Or you are planing to modify the test afterwards?

dangars:
If you separate in real and imaginary parts. I guess that the complete version of the loop would be something like:
``` c++
for (unsigned int d = 0; d < 2*dim; ++d)
{
computed_quantities[q](d) = input_data.solution_values[q](0) *
                          input_data.solution_values[q](0) +
                        input_data.solution_values[q](1) *
                          input_data.solution_values[q](1)+
                          input_data.solution_values[q](2) *
                          input_data.solution_values[q](2) +
                        input_data.solution_values[q](3) *
                          input_data.solution_values[q](3);
AssertThrow(std::fabs(computed_quantities[q](d) - 1) < 1e-12,
        ExcInternalError());
std::cout << "d = " << d<< " " <<computed_quantities[q](d)<< std::endl;
}
```
The components 0 and 1 are real. The components 2 and 3 are imaginary. If I put this loop in the test it fails.

In addition, the drawback of this particular test is that the imaginary numbers are equal to zero.

Once everything is done, if you ping me, I can write a better test in which the imaginary parts are not equal to zero.

tests/data_out/data_out_postprocessor_vector_complex_01.cc
bangerth: That's what you specify in the test: That it's a `DataPostprocessorVector` (which has `dim` output components), and you provide the name "magnitude_times_d" to this output vector. So yes, by virtue of deriving from this base class, you say that the output is a real-valued vector.

bangerth:
@dangars -- the output of data postprocessors are only real-valued fields, for the same reason as the output of `DataOut` are real-valued fields. If you wanted to output a vector with both real and imaginary parts, you'd have to either write one class derived from `DataPostprocessor` that computes two real-valued vectors, or you'd have two classes, each derived from `DataPostprocessorVector`, that compute the real and the imaginary parts, respectively.

@dealii/dealii: While the test we're debating here doesn't necessarily make sense, I've double checked that it really does the correct thing based on what I wanted the functionality to be. Would someone be so kind and review it? :-)

source/numerics/data_out.cc
```diff
@@ -429,8 +429,196 @@ DataOut<dim, DoFHandlerType>::build_one_patch(
...
                       else
                         {
-                          // This is the vector-valued, complex-valued case.
-                          Assert(false, ExcNotImplemented());
+                          scratch_data.resize_system_vectors(2 * n_components);
+
+                          // This is the vector-valued, complex-valued case. In
+                          // essence, we just need to do the same as above,
+                          // i.e., call the functions in this->dof_data[dataset]
+                          // to retrieve first the real and then the imagine
```
kronbichler:
```suggestion
                          // to retrieve first the real and then the imaginary
```

bangerth:
I've fixed the typo mentioned by @kronbichler .

PR #8770, opened 2019-09-17
Document DataPostprocessor for the case of complex-valued solutions.

bangerth:
Part of #1894. @dangars -- FYI.

### DIFF
--- a/include/deal.II/numerics/data_postprocessor.h
+++ b/include/deal.II/numerics/data_postprocessor.h
@@ -218,7 +218,12 @@ namespace DataPostprocessorInputs
    * A structure that is used to pass information to
    * DataPostprocessor::evaluate_scalar_field(). It contains
    * the values and (if requested) derivatives of a scalar solution
-   * variable at the evaluation points on a cell or face.
+   * variable at the evaluation points on a cell or face. (This class
+   * is not used if a scalar solution is complex-valued, however,
+   * since in that case the real and imaginary parts are treated
+   * separately -- resulting in vector-valued inputs to data
+   * postprocessors, which are then passed to
+   * DataPostprocessor::evaluate_vector_field() instead.)
    *
    * Through the fields in the CommonInputs base class, this class also
    * makes available access to the locations of evaluations points,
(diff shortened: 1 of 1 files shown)


include/deal.II/numerics/data_postprocessor.h
```diff
@@ -422,8 +450,47 @@ namespace DataPostprocessorInputs
...
+ * solves a problem whose solution at each point consists of a complex number
+ * represented by a `std::complex<double>` variable. (step-29 also solves such
+ * a problem, but there we choose to represent the solution by two real-valued
+ * fields.) In such cases, the vector that is handed to
+ * DataOut::build_patches() is of type `Vector<std::complex<double>>`, or
+ * something essentially equivalent to this. The issue with this, as also
+ * discussed in the documentation of DataOut itself, is that no visualization
+ * program we know of can actually represent complex quantities. The only thing
+ * that can reasonably be stored in data files for visualization tools are
+ * real-valued quantities.
```
dangars: I don't think this is completely accurate. The software based on the VTK library can not deal with complex numbers, but other visualization software, such as `python/numpy` have native support for complex numbers. HDF5 can store `std::complex` datasets. The `deal.II` interface of HDF5 supports `std::complex` datasets.

In step-62 I store complex-valued data in a HDF5 file and then visualize the data with `python`.

I would say that the VTK library does not support complex numbers, but I wouldn't say that the rest of the visualization tools don't support complex numbers.

bangerth: I agree that other file formats allow *storing* complex numbers, but I don't know any visualization file format that allows *visualizing* that data.

Certainly, to the best of my knowledge, none of the file formats that the functions in `DataOutBase` can write in, supports complex numbers.

I'm willing to make this sound more conciliatory, of course, but I think it's fundamentally true: All visualization software ultimately visualizes real-valued data :-)

dangars: > I'm willing to make this sound more conciliatory.

That sounds good :)

bangerth:
@dangars -- take another look if you don't mind.

bangerth:
Ping?

include/deal.II/numerics/data_postprocessor.h
```diff
@@ -422,8 +450,47 @@ namespace DataPostprocessorInputs
...
+ * fields.) In such cases, the vector that is handed to
+ * DataOut::build_patches() is of type `Vector<std::complex<double>>`, or
+ * something essentially equivalent to this. The issue with this, as also
+ * discussed in the documentation of DataOut itself, is that the most widely
+ * used file formats for visualization (notably, the VTK and VTU formats)
+ * can not actually represent complex quantities. The only thing
+ * that can be stored in these data files are real-valued quantities.
+ *
+ * As a consequence, DataOut is forced to take things apart into their real
+ * and complex parts, and both are output as separate quantities. This is the
```
kronbichler:
```suggestion
 * and imaginary parts, and both are output as separate quantities. This is the
```

include/deal.II/numerics/data_postprocessor.h
```diff
@@ -528,7 +600,8 @@ class DataPostprocessor : public Subscriptor
  *
  * All derived classes have to do is implement a constructor and overload
  * either DataPostprocessor::evaluate_scalar_field() or
- * DataPostprocessor::evaluate_vector_field().
+ * DataPostprocessor::evaluate_vector_field() as discussed in the
+ * DataPostprocessor class's document.
```
kronbichler: Don't you spell this like this:
```suggestion
 * DataPostprocessor class' document.
```

include/deal.II/numerics/data_postprocessor.h
```diff
@@ -609,7 +682,8 @@ class DataPostprocessorScalar : public DataPostprocessor<dim>
...
```
kronbichler:
```suggestion
 * DataPostprocessor class' document.
```

include/deal.II/numerics/data_postprocessor.h
```diff
@@ -873,7 +947,8 @@ class DataPostprocessorVector : public DataPostprocessor<dim>
...
```
kronbichler:
```suggestion
 * DataPostprocessor class' document.
```

include/deal.II/numerics/data_postprocessor.h
bangerth: No, in the English language the 's' after the apostrophe is only dropped if you the 's' before the apostrophe is a plural 's', not if it's part of the noun ("class"). See https://www.thepunctuationguide.com/apostrophe.html#possessives

Now, while you were fixing this, you completely missed the fact that it should have been this:
```suggestion
 * DataPostprocessor class's documentation.
```
:-)

include/deal.II/numerics/data_postprocessor.h
bangerth:
```suggestion
 * DataPostprocessor class's documentation.
```

bangerth:
I fixed the marked up places, with the exception of the ones related to the possessive "'s".

include/deal.II/numerics/data_postprocessor.h
kronbichler: You're completely right of course. Thanks for fixing it.

include/deal.II/numerics/data_postprocessor.h
kronbichler: Can you please fix `documentation` here as well?

bangerth:
Fixed the last issue -- I thought I had found all instances, but apparently missed one...

kronbichler:
Thanks @bangerth !

Issue #8293, opened 2019-05-28
Code optimizations for >100k MPI ranks


kronbichler:
I'm looking at a run of deal.II on 304,128 cores right now:
```
Cycle 34
Time create grid              25.407      [p169179] 25.43       25.459      [p17739 ]
Time distribute DG dofs       8.8281      [p199286] 8.8357      8.8419      [p117609]
Number of degrees of freedom  1,099,511,627,776 = (4,096 x 2,048 x 2,048) x 4^3
Time distribute fe dofs       7.5373      [p10692 ] 7.5441      7.5643      [p98357 ]
Time distribute fe mg dofs    8.7065      [p197882] 9.3725      9.9435      [p2837  ]
Compute numbering L0          0.22594     [p251129] 0.30643     0.37282     [p20196 ]
Apply new numbers L0          0.19688     [p195524] 0.25103     0.27644     [p26683 ]
Compute numbering L1          0.18782     [p129933] 0.22178     0.25139     [p182506]
Apply new numbers L1          0.1976      [p164685] 0.25189     0.28095     [p109296]
Compute numbering L2          0.19066     [p204755] 0.2251      0.24673     [p20984 ]
Apply new numbers L2          0.2033      [p7121  ] 0.25571     0.28955     [p166320]
Compute numbering L3          0.18837     [p260943] 0.22299     0.24845     [p44645 ]
Apply new numbers L3          0.19909     [p133311] 0.25184     0.2965      [p187072]
Compute numbering L4          0.18634     [p98239 ] 0.22364     0.24942     [p212268]
Apply new numbers L4          0.20318     [p195401] 0.25861     0.31113     [p16181 ]
Compute numbering L5          0.1919      [p274846] 0.22556     0.24382     [p99466 ]
Apply new numbers L5          0.27852     [p194310] 0.31275     0.33555     [p156624]
Compute numbering L6          0.20708     [p97629 ] 0.2337      0.26738     [p72537 ]
Apply new numbers L6          0.31305     [p10692 ] 0.34891     0.37427     [p41142 ]
Compute numbering L7          0.20731     [p171822] 0.23188     0.24859     [p48050 ]
Apply new numbers L7          0.31745     [p152361] 0.34565     0.3826      [p144677]
Compute numbering L8          0.2073      [p302909] 0.23454     0.247       [p216398]
Apply new numbers L8          0.31612     [p196019] 0.34979     0.39201     [p114101]
[...]

PR #8772, opened 2019-09-17
Eliminate MPI all-to-all communication in MPI::Partitioner

kronbichler:
This PR uses the sparse data exchange functionality of `Utilities::MPI::compute_index_owner` to eliminate the `MPI_Alltoall` and `MPI_Allgather` functions in `Utilities::MPI::Partitioner`, which do not scale to large processor counts.

To use the code, I refactored the `compute_index_owner` code a bit and moved some common code that I want to re-use and extend for the partitioner class to a new header file `mpi_compute_index_owner_internal.h`. To ease the review, I split moving existing code around into a first commit and the actual code changes and extensions into a [the second commit](https://github.com/dealii/dealii/pull/8772/commits/6763de1c6c622fe675363546e0a02cd61aaaf7ed).

What I did: `compute_index_owner` evaluates the owner the ghost "ghost indices" by looking them up in a dictionary. The `Partitioner` class further wants to cache the place where those indices came from at the owner of the index. So whenever I look up the dictionary, I do not only compute the owner of the index, but also store which process wanted to have that information. This allows me then to send that information via the dictionary partitioning to the owner.

Note that I also plan to introduce a minimum grain size in the `dofs_per_process` variable in the dictionary: Right now, we use a fair splitting `(size + n_procs - 1) / n_procs`, which obviously leads to very small ranges and lots of unnecessary communication in case `size` is only a small multiple of the number of processors. We should use at least 2048 unknowns in the dictionary per process. To ease reviewing and allow me separate benchmarking, I will not include that change in this PR but open a separate one once this is merged.

Part of #8293.


kronbichler:
@peterrum FYI.

One thing about the implementation I should add is that I currently determine the unknowns subject to ghost exchange at the index owner's site via the dictionary. The code would probably be a bit simpler if we only collected the ranks of the requests, sent the list of ranks to the index owner, and, given the knowledge of the number of affected ranks, send the ghost indices with the usual `MPI_Probe` mechanism. However, I only realized this was possible after I had the current solution, and the alternative solution involves more messages because right now I attach the ranges directly with the ranks, so they are in the same message.

peterrum:
I'll review it in the afternoon. I think you might have forgotten in some places `#ifdef DEAL_II_WITH_MPI`.

source/base/partitioner.cc
```diff
@@ -474,7 +413,7 @@ namespace Utilities
             }
           ghost_indices_subset_data = ghost_indices_subset;
         }
-    }
+    } // namespace MPI
```
Rombur: I don't think that this is the end of the namespace.

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -29,21 +29,29 @@ namespace Utilities
...
+         * Dictionary class with basic partitioning in terms of a single
+         * interval of fixed size known to all MPI ranks for two-stage index
+         * lookup.
+         */
         struct Dictionary
         {
           static const unsigned int tag_setup = 11;
 
           std::vector<unsigned int> actually_owning_ranks;
+          std::vector<unsigned int> actually_owning_rank_list;
```
Rombur: Can you add a comment about the difference between the two variables

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -192,6 +203,25 @@ namespace Utilities
...
+            if (actually_owning_rank_list[guess] == actual_rank)
+              return guess;
+            else
+              {
+                auto it = std::lower_bound(actually_owning_rank_list.begin(),
+                                           actually_owning_rank_list.end(),
+                                           actual_rank);
+                Assert(it != actually_owning_rank_list.end(),
+                       ExcInternalError());
+                Assert(*it = actual_rank, ExcInternalError());
```
Rombur: I don't think you want to do an assignment in the `Assert`

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -223,23 +260,31 @@ namespace Utilities
...
           const IndexSet &           indices_to_look_up;
           const MPI_Comm &           comm;
           const unsigned int         my_rank;
           const unsigned int         n_procs;
+          const bool                 track_index_requests;
           std::vector<unsigned int> &owning_ranks;
+          std::vector<std::vector<
+            std::pair<unsigned int,
+                      std::vector<std::pair<unsigned int, unsigned int>>>>>
+            requesters;
```
Rombur: Can you add a comment about what is the data stored here.

peterrum:
> What I did: `compute_index_owner` evaluates the owner the ghost "ghost indices" by looking them up in a dictionary. The `Partitioner` class further wants to cache the place where those indices came from at the owner of the index. So whenever I look up the dictionary, I do not only compute the owner of the index, but also store which process wanted to have that information. This allows me then to send that information via the dictionary partitioning to the owner.

This idea sound quite familiar ;) Glad that my implementation is in use!

include/deal.II/base/mpi_compute_index_owner_internal.h
kronbichler: I will add descriptions to all members in this struct.

source/base/partitioner.cc
kronbichler: You are right - that was inserted automatically by `clang-format` when I had a bug in the code (one closing brace too much), but apparently not removed again.

include/deal.II/base/mpi_compute_index_owner_internal.h
kronbichler: Done!

peterrum:
@kronbichler indentation ;)

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -0,0 +1,799 @@
...
+        };
+
+
+
+        /**
+         * Specialization of ConsensusAlgorithmProcess for the context of
+         * Utilities::MPI::compute_index_owner() and
+         * Utilities::MPI::Partitioner::set_ghost_indices().
+         */
+        class ConsensusAlgorithmProcess
```
tjhei: maybe this class should have a different name?

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -0,0 +1,799 @@
...
+          /**
+           * The indices which are "ghosts" on a given rank and should be
+           * looked up in terms of their owner rank from owned_indices.
+           */
+          const IndexSet &indices_to_look_up;
+
+          /**
+           * The underlying MPI communicator.
+           */
+          const MPI_Comm &comm;
```
tjhei: this is an int internally, right? I think we normally copy it.

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -0,0 +1,799 @@
...
+          virtual void
+          process_request(
+            const unsigned int                          other_rank,
+            const std::vector<types::global_dof_index> &buffer_recv,
+            std::vector<unsigned int> &                 request_buffer) override
+          {
+            Assert(buffer_recv.size() % 2 == 0, ExcInternalError());
+            unsigned int owner_index = 0;
+            for (unsigned int j = 0; j < buffer_recv.size(); j += 2)
+              for (auto i = buffer_recv[j]; i < buffer_recv[j + 1]; i++)
```
tjhei: ``buffer_recv`` conceptually is a vector of ``pair<dof_index,dof_index>``, is that possible to make that change?

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -0,0 +1,799 @@
...
+            // 3) collect indices for each process
+            {
+              unsigned int index = 0;
+              for (auto i : indices_to_look_up)
+                {
+                  unsigned int other_rank = dict.dof_to_dict_rank(i);
+                  if (other_rank != my_rank)
+                    {
+                      recv_indices[other_rank].push_back(index);
+                      temp[other_rank].push_back(i);
```
tjhei: I am not sure I like the name ``temp``.

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -0,0 +1,799 @@
...
+            // to make sure no memory-reallocation is needed while filling in
+            // the data and communicating, as we need to keep the buffers
+            // alive while MPI is active.
+            std::vector<unsigned int> send_data;
+            {
+              unsigned int n_integers = 0;
+              for (unsigned int i = 0; i < requesters.size(); ++i)
+                if (dict.actually_owning_rank_list[i] != my_rank)
+                  for (const auto &j : requesters[i])
+                    n_integers += 2 + 2 * j.second.size();
```
tjhei: I find packing data manually into a ``vector<unsigned int>`` hard to understand/debug/fix. I understand that you don't want the overhead of serializing a data structure. Would it be difficult to have an internal class that stores data contiguously internally?

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -0,0 +1,799 @@
...
+            // need to send (a) the MPI rank of the requester, (b) the number
+            // of intervals directed to this requester, and (c) a list of
+            // intervals, i.e., two integers per interval. The number of items
+            // sent in total can be deduced both via the MPI status message at
+            // the receiver site as well as be counting the buckets from
+            // different requesters. First we count how much data to reserve
+            // to make sure no memory-reallocation is needed while filling in
+            // the data and communicating, as we need to keep the buffers
+            // alive while MPI is active.
+            std::vector<unsigned int> send_data;
```
tjhei: Why do you put the buffers you send to different ranks in the same std::vector? I think you could skip the ``offset`` logic if you have one per message.

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -0,0 +1,799 @@
...
+                    {
+                      ++last;
+                      it.second[last] = it.second[check];
+                    }
+
+                AssertIndexRange(last, it.second.size());
+
+                // erase duplicates
+                if (last + 1 != it.second.size())
+                  it.second.erase(it.second.begin() + last + 1,
```
tjhei: would an IndexSet do all this logic of sorting/merging for you here?

kronbichler: I tried, but `IndexSet::add_range` implies quadratic complexity in the number of ranges; for a typical example of an `FE_Q`, we can have thousands of ranges. I tried to work around it by different index sets and extracting/compressing indices, but then the code is not nicer than what I have here. To see the "code overhead" of skipping the index set, have a look at the [fourth commit](https://github.com/dealii/dealii/pull/8772/commits/3f756a1c9cf951405bd1cb96933c707bfaa1346c).

tjhei: Alright, but we might want to rethink the implementation in IndexSet then. Implementing an ``add_ranges()`` function should be able to work in O(nlogn).

include/deal.II/base/mpi_compute_index_owner_internal.h
kronbichler: The question is whether the additional class is worth the effort, since it's a single place. A step into this direction would be to convert the `std::vector<unsinged int>` into an `std::vector<std::pair<unsigned int,unsigned int>>`. Then we still have the ambiguity that the very first entry interprets the pair as the number of "queued pairs for this process" plus the processor rank, but at least the other items are ranges as intended. At least that's what we do in the respective classes of `DoFHandlerPolicy`: We misuse the first integer in some global dof numbers to indicate how many dof numbers follow for the current element (to support hp). That's no excuse for the sin of course, but I would rather live with this "packing" problem given the local scope than an additional class that hides the packing.

include/deal.II/base/mpi_compute_index_owner_internal.h
kronbichler: OK, will do with the other changes.

include/deal.II/base/mpi_compute_index_owner_internal.h
tjhei: > Then we still have the ambiguity that the very first entry interprets the pair as the number of "queued pairs for this process" plus the processor rank, but at least the other items are ranges as intended. 
That doesn't sound better to me. :-)

I wish there was an easy way to have several std::vectors place their memory allocations contiguously after each other.

masterleinad: You can always write a custom allocator. :wink:

include/deal.II/base/mpi_compute_index_owner_internal.h
bangerth: It's not specified. It could be implemented as a struct, or a pointer to something. It's better to copy it.

include/deal.II/base/mpi_compute_index_owner_internal.h
bangerth: We already have `add_ranges()` -- it's the function that computes the union of two `IndexSet`s.

include/deal.II/base/mpi_compute_index_owner_internal.h
bangerth: If they're the same size, you can always use `Table<2,T>`.

kronbichler: No, unfortunately not. Let us stick with the current solution, given that I'm unhappy to move the mechanism to stick things together, used once here, to a separate class.

include/deal.II/base/mpi_compute_index_owner_internal.h
kronbichler: Right, how could I miss that. Well, maybe because `add_indices` is not the name I was looking for (rather `merge` or something like that). I will go back to the IndexSet solution, given that it makes code slightly simpler.

include/deal.II/base/mpi_compute_index_owner_internal.h
kronbichler: Not easily at least: The problem is that we use `mpi_type_id(T1)` in `ConsensusAlgorithm` for sending the MPI, and we do not have that for `pair<dof_index,dof_index>`. So we would either have to define an MPI type for a pair or use `MPI_BYTE` as a type times `sizeof(T1)`. I will temporarily go with the `sizeof(T1)` variant but if @peterrum strongly objects how I misuse the class we should think about an alternative.

peterrum: @tjhei It used to be a pair. I guess if you have look at the first commit regarding this matter you will see that. I moved from pair to an int because of the reasons @kronbichler wrote (when I generalized the class and introduced the template parameter). `sizeof` is somewhat an ugly approach, but I guess it will do the job ;)

kronbichler: I have that variant implemented. Let me run some local matrix-free/multigrid tests where this is stressed a lot before I force-push the changes.

kronbichler:
I should have fixed all comments, except for the one of encoding the length of vectors within the integer vector itself where I see no better alternative solution.

tjhei:
clang tidy says:
```
include/deal.II/base/mpi_compute_index_owner_internal.h:435:13: error: use range-based for loop instead [modernize-loop-convert,-warnings-as-errors]
```

kronbichler:
> clang tidy says: [...]

That's what I really like about the CI tests - it is a good suggestion that was not possible before but now that we have pairs.

tjhei:
Looks good to me. Did you run all mpi/multigrid tests?

kronbichler:
Yes, I ran all tests with `mpi` in the name, which includes all parallel multigrid tests as they have some `mpirun=...` in the name.

PR #8773, opened 2019-09-17
Add function MappingCartesian::fill_jacobian_derivatives(..)

simonsticko:
Presently all of the fill_fe_*values functions in MappingCartesian contain loops that fill the derivatives of the Jacobian with zeros. In order to remove duplication, add a function which does this instead.

### DIFF
--- a/source/fe/mapping_cartesian.cc
+++ b/source/fe/mapping_cartesian.cc
@@ -316,6 +316,58 @@ MappingCartesian<dim, spacedim>::maybe_update_normal_vectors(
 
 
 template <int dim, int spacedim>
+void
+MappingCartesian<dim, spacedim>::maybe_update_jacobian_derivatives(
+  const InternalData &             data,
+  const CellSimilarity::Similarity cell_similarity,
+  internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
+    &output_data) const
+{
+  if (cell_similarity != CellSimilarity::translation)
+    {
+      if (data.update_each & update_jacobian_grads)
+        for (unsigned int i = 0; i < output_data.jacobian_grads.size(); ++i)
+          output_data.jacobian_grads[i] = DerivativeForm<2, dim, spacedim>();
+
+      if (data.update_each & update_jacobian_pushed_forward_grads)
+        for (unsigned int i = 0;
+             i < output_data.jacobian_pushed_forward_grads.size();
+             ++i)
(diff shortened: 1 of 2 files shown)


simonsticko:
Differently sizes Cartesian cells does make sense. 

I didn't think much about this when I refactored. Do we even need this function I've added? The derivatives of the Jacobian are zero regardless of the cell size. They are set to zero when `MappingRelatedData` is constructed right? Seems like there is no need to set them to zero on each call to `reinit(..)'.

bangerth:
I don't know. It's possible that these values are set to `signaling_nan()` to make sure they're not used uninitialized. But then we should be able to inspect the update flags once at the very beginning and set the values to zeros -- and then never touch them again.

I'm an incrementalist. This patch is a step in the right direction, so let's get it merged. But if you have a minute, take a look at a follow-up patch to see whether it's possible to just move the initialization to zero to an earlier time where it only has to happen once.

source/fe/mapping_cartesian.cc
```diff
@@ -315,6 +315,59 @@ MappingCartesian<dim, spacedim>::maybe_update_normal_vectors(
...
+MappingCartesian<dim, spacedim>::maybe_update_jacobian_derivatives(
+  const InternalData &             data,
+  const CellSimilarity::Similarity cell_similarity,
+  internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
+    &output_data) const
+{
+  if (data.update_each & update_jacobian_grads)
+    if (cell_similarity != CellSimilarity::translation)
+      for (unsigned int i = 0; i < output_data.jacobian_grads.size(); ++i)
+        output_data.jacobian_grads[i] = DerivativeForm<2, dim, spacedim>();
```
bangerth: Hm, on second thought, I think this may be the best you can do. The `MappingCartesian` class doesn't own the `MappingRelatedData` (it's owned by the `FEValues` object) and we can't initialize it any earlier than when we see it the first time. I think there is nothing else you can do.

By the way, `cell_similarity != CellSimilarity::translation` also on the very first cell one encounters.

bangerth:
Given my last comment, the only possible improvement I see is to move the `if (cell_similarity != CellSimilarity::translation)` check out of the individual blocks into a surrounding `if`. No functional change, just making things a bit easier on the eye.

simonsticko:
Updated as suggested.

PR #8777, opened 2019-09-17
Remove doc stating FECollection is unused in hp::FE*Values classes.

simonsticko:
In the documentation of the constructors of the various hp::FE*Values classes it is stated that the FECollection parameter passed to the constructors isn't used inside the classed. This is no longer true. Remove the statement.


bangerth:
Thanks for fixing my mistakes from a long time ago!

PR #8779, opened 2019-09-17
Fix compiling with clang

masterleinad:
This should fix the failing `clang` builds.


bangerth:
This seems uncontroversial enough that I'm tempted to merge it even without running the testers. But the mandatory checks haven't completed yet...

PR #8782, opened 2019-09-18
Redefine TimestepControl::tolerance()

### LINKED ISSUE #8771: Definition of TimestepControl::tolerance
The variable `tolerance_val` and its corresponding get and set methods in `TimestepControl` do not provide a good documentation of exactly how it is used in the algorithm.

I assume that it determines how much the size of a time step can be increased at the last step in order to reach the final time. However, my assumption is not reflected in the implementation of the `advance()` method. So I am slightly confused about the meaning of `tolerance` in this context.

I can improve the documentation and the code once I know the meaning of this variable.

rezarastak:
fixes #8771 

The previous implementation used the variable `tolerance_val` in the following block in the method `advance()`
```c++
   if (now_val != start())
     {
       if (strategy_val == doubling && 2 * s <= tolerance_val)
         s *= 2;
       if (s > max_step_val)
         s = max_step_val;
     }
```
The implementation is confusing and implies `tolerance_val` and `max_step_val` provide the same functionality.

In my opinion, the best way we can define `tolerance` is the amount a time step can be increased at the final step. For example, is current time is 0.945, the step size is 0.5, and the tolerance is 0.1, the algorithm can increase the step size to 0.55 so that we reach the final time (t = 1) in a single step. Here we increase the step size by 0.05 which is smaller than the tolerance. I modified the documentation to explain this concept in a clear way.

This PR changes the behavior of the `advance()` in a tiny way and might break backward compatibility in rare cases. Since there were no tests for `TimestepControl`, I added one test to ensure its usability in the future.

### DIFF
--- a/source/algorithms/timestep_control.cc
+++ b/source/algorithms/timestep_control.cc
@@ -94,7 +94,7 @@ TimestepControl::advance()
   // first step.
   if (now_val != start())
     {
-      if (strategy_val == doubling && 2 * s <= tolerance_val)
+      if (strategy_val == doubling)
         s *= 2;
       if (s > max_step_val)
         s = max_step_val;
@@ -113,8 +113,7 @@ TimestepControl::advance()
   // shot over the final time, adjust
   // it so we hit the final time
   // exactly.
-  double s1 = .01 * s;
-  if (h > final_val - s1)
+  if (h > final_val - tolerance_val)
     {
       current_step_val = final_val - now_val;
       h                = final_val;
--- a/doc/news/changes/incompatibilities/20190917RezaRastak
+++ b/doc/news/changes/incompatibilities/20190917RezaRastak
@@ -1,0 +1,5 @@
+Modified: The definition of tolerance in TimestepControl is changed to
+represent the amount that the time step size can be increased in the last
+step.
+<br>
+(Reza Rastak, 2019/09/17)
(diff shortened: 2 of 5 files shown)


bangerth:
On 9/17/19 6:53 PM, Reza Rastak wrote:
> ||
> 
>     if  (now_val != start())
>       {
>         if  (strategy_val == doubling &&2  * s <= tolerance_val)
>           s *=2;
>         if  (s > max_step_val)
>           s = max_step_val;
>       }
> 
> The implementation is confusing and implies |tolerance_val| and 
> |max_step_val| provide the same functionality.

Yes, this makes no sense. In one use of 's', it is treated as a time, in 
the other use it is treated as an output variable that can be compared 
against a tolerance. These generally have different physical units...

source/algorithms/timestep_control.cc
```diff
@@ -94,7 +94,7 @@ TimestepControl::advance()
   // first step.
   if (now_val != start())
     {
-      if (strategy_val == doubling && 2 * s <= tolerance_val)
+      if (strategy_val == doubling)
         s *= 2;
       if (s > max_step_val)
```
bangerth: That can't have been the intent. This means that in each time step, you're going to double the time step length until you eventually hit the end time or the maximal time step size.

I strongly suspect that this strategy was never tested, and has always been irretrievably broken. I don't think your patch actually "fixes" it -- the whole scheme here makes no sense. Can you just remove `Strategies::doubling` altogether? It's not been tested, it doesn't work, and it's likely that nobody uses it, so let's just get rid of it.

rezarastak:
I agree. I will remove enum Strategy and will delete all of its related code.

bangerth:
Let's close this PR then to avoid confusion. We can reopen if you choose to reuse the same PR, or you can open a separate PR for the removal of the option.

PR #8783, opened 2019-09-18
Clean up compute_vertices_with_ghost_neighbors

### LINKED ISSUE #8734: Treat periodic neighbors correctly in  compute_vertices_with_ghost_neighbors
Inside of `source/distributed/tria_base.cc` we can find the problematic implementation: 
```cpp
  std::map<unsigned int, std::set<dealii::types::subdomain_id>>
  TriangulationBase<dim, spacedim>::compute_vertices_with_ghost_neighbors()
    const
  {
    // TODO: we are not treating periodic neighbors correctly here. If we do
    // we can remove the overriding implementation for p::d::Triangulation
    // that is currently using a p4est callback to get correct ghost neighbors
    // over periodic faces.
    Assert(this->get_periodic_face_map().size() == 0, ExcNotImplemented());


    std::vector<bool> vertex_of_own_cell(this->n_vertices(), false);

    for (const auto &cell : this->active_cell_iterators())
      if (cell->is_locally_owned())
        for (unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell; ++v)
          vertex_of_own_cell[cell->vertex_index(v)] = true;

    std::map<unsigned int, std::set<dealii::types::subdomain_id>> result;
    for (const auto &cell : this->active_cell_iterators())
      if (cell->is_ghost())
        {
          const types::subdomain_id owner = cell->subdomain_id();
          for (unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell;
               ++v)
            {
              if (vertex_of_own_cell[cell->vertex_index(v)])
                result[cell->vertex_index(v)].insert(owner);
            }
        }

    return result;
  }
```
where the first comment states that it does not work for periodicity properly in the case for distributed triangulations. As a workaround the job is passed over to p4est in the case of `p:d:t`. However, this is not possible in the case of `p:f:t`. @bangerth  and I think, it would be nice if we/I would fix this function so that I can use in `p:f:t` and we can eliminate the specialization in `p:d:t`.

peterrum:
Unify `compute_vertices_with_ghost_neighbors` in `parallel::TriangulationBase`, `parallel::distributed::Triangulation`, and `parallel::fullydistributed::Triangulation`. Hopefully it is possible.

In the first step, I have moved my implementation of `p:f:t` in the base class and removed the `p:d:t` specializations. This does not work yet. The following tests are failing (I still have to figure out the reason):
- ~~mpi/mg_ghost_dofs_periodic_04.mpirun=5.debug (Failed)~~
- ~~mpi/periodicity_04.mpirun=3.debug (Failed)~~ 
- ~~mpi/periodicity_04.mpirun=5.debug (Failed)~~
- ~~mpi/periodicity_04.mpirun=7.debug (Failed)~~
- ~~mpi/periodicity_06.mpirun=2.debug (Failed)~~
- ~~mpi/periodicity_07.mpirun=13.debug (Failed)~~

Closes #8734.


source/distributed/tria_base.cc
```diff
@@ -362,36 +362,125 @@ namespace parallel
...
+        for (auto &ss : s.second)
+          vertex_to_coinciding_vertices[ss] = s.second;
+    }
 
+    // 2) collect vertices belonging to local cells
+    std::vector<bool> vertex_of_own_cell(this->n_vertices(), false);
     for (const auto &cell : this->active_cell_iterators())
       if (cell->is_locally_owned())
-        for (unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell; ++v)
+        for (auto v = 0u; v < GeometryInfo<dim>::vertices_per_cell; ++v)
```
kronbichler: Can you please undo this change? I find the specified type much easier to read than `0u`, especially for inexperienced users.

source/distributed/tria_base.cc
```diff
@@ -362,36 +362,125 @@ namespace parallel
...
           vertex_of_own_cell[cell->vertex_index(v)] = true;
 
+    // 3) for for each vertex belonging to a locally owned cell all ghost
```
kronbichler:
```suggestion
    // 3) for each vertex belonging to a locally owned cell all ghost
```

source/distributed/tria_base.cc
```diff
@@ -362,36 +362,125 @@ namespace parallel
...
+    //    neighbors (including the periodic own)
     std::map<unsigned int, std::set<dealii::types::subdomain_id>> result;
+
+    // loop over all active ghost cells
     for (const auto &cell : this->active_cell_iterators())
       if (cell->is_ghost())
         {
-          const types::subdomain_id owner = cell->subdomain_id();
+          const auto owner = cell->subdomain_id();
```
kronbichler: Also here, I prefer to keep the type - right now, the type is explicit in line 453 via the `std::set` but implicit here. Since this is so little to type, the educational aspect of listing it here is more important than saving 15 characters in my opinion, at least for existing code. (As opposed to say an iterator that comes out of `find` in line 471 below.)

source/distributed/tria_base.cc
```diff
@@ -362,36 +362,125 @@ namespace parallel
...
+    // 1) collect for each vertex on periodic faces all vertices it coincides
+    //    with
+    std::map<unsigned int, std::set<unsigned int>>
+      vertex_to_coinciding_vertices;
+    {
+      // 1a) collect nodes coinciding due to periodicity
+      std::map<unsigned int, unsigned int> vertex_to_coinciding_vertex;
+      for (auto &cell : this->active_cell_iterators())
+        if (cell->is_locally_owned() || cell->is_ghost())
+          for (auto i = 0u; i < GeometryInfo<dim>::faces_per_cell; ++i)
```
kronbichler:
```suggestion
          for (unsigned int i = 0; i < GeometryInfo<dim>::faces_per_cell; ++i)
```
(let us spell out integers consistent with the rest of the library)

peterrum:
Has anyone implemented a function that collects all vertices coinciding due to periodicity? I guess my implementation has a bug...

kronbichler:
> Has anyone implemented a function that collects all vertices coinciding due to periodicity?

Didn't we have one such function either in the setup of the levels of the distributed triangulation or in `DoFHandlerPolicy`? I recall having done something like that, but I have not looked up where.

peterrum:
The reason for issue #8734 and this PR is that the distributed triangulation passes the job to `p4est`. See `source/distributed/tria_base.cc`:
```cpp
    // TODO: we are not treating periodic neighbors correctly here. If we do
    // we can remove the overriding implementation for p::d::Triangulation
    // that is currently using a p4est callback to get correct ghost neighbors
    // over periodic faces.
```
@bangerth and I do not want a seperate implementation of `compute_vertices_with_ghost_neighbors` for `p:s:t`, for `p:d:t`, and for ``p:f:t``.

peterrum:
@bangerth As we have discussed, I have unified this function for all parallel triangulations (unfortunately the number of loc has hardly decreased). It seems to work. I have run successfully 37 tests with ` ctest -j30 -R periodic`. 

In a separate - follow-up - PR, I would like to move the determination of coinciding vertices into `GridTools`.

source/distributed/tria_base.cc
```diff
@@ -362,36 +362,160 @@ namespace parallel
...
+            {0, 1, 2, 3}, // true        false false
+            {1, 3, 0, 2}  // true        false true
+          },
+          {
+            {3, 2, 1, 0}, // true        true  false
+            {2, 0, 3, 1}  // true        true  true
+          }}};
+
+      // loop over all periodic face pairs
+      for (auto pair : this->get_periodic_face_map())
```
kronbichler: We typically write
```suggestion
      for (const auto &pair : this->get_periodic_face_map())
```
(even though the entries are not super large and do not involve memory re-allocation, I would still capture by reference).

peterrum: Done. Ready to test!

masterleinad:
We have a few failing periodicity tests (https://cdash.43-1.org/viewTest.php?onlydelta&buildid=1891) and I believe this pull request is responsible.

PR #8784, opened 2019-09-18
Remove an auto from a lambda in pft

peterrum:
Thanks and FYI @kronbichler


kronbichler:
This should fix the build failure on gcc-4.9.4:
https://cdash.43-1.org/viewBuildError.php?buildid=1689

bangerth:
Ah, darn it -- I missed that one...

PR #8789, opened 2019-09-18
enum Strategy is removed from TimestepControl

rezarastak:
From our discussion in #8782 

Since people rarely use this enum and the implementation of the `doubling` strategy does not make much sense, I removed  `enum Streategy` from `TimestepControl`.


kronbichler:
The tests did run successfully but the machine got stuck in the end. I'm merging.

PR #8300, opened 2019-06-03
Mpi compute index owner

peterrum:
The function `Utilities::MPI::compute_index_owner()` provides the owner of the ghost degrees of freedom relative to a set of locally owned ones. This function is used `Utilities::MPI::Partitioner` to avoid global all-to-all communication, speeding up computations on many MPI ranks.

I will also apply this new function in `internal::MGTransfer::fill_copy_indices`. However, this is still work in progress.

Relates to #8293 and #8298 .

PR #8792, opened 2019-09-18
Remove obsolete global communication from MGTransfer

kronbichler:
This code was forgotten when @peterrum introduced the index owner computation in #8300.

@tjhei @tcclevenger FYI


peterrum:
Thanks @kronbichler ! I guess no one noticed it because it was not visible in the diff...

kronbichler:
tidy seems to be stuck - but since this is strictly removing code, I'm going to merge now.

PR #8794, opened 2019-09-19
Avoid dangling pointer in initializer_list

masterleinad:
See https://godbolt.org/z/_X7hQ-. We should avoid copying `std::initializer_list` objects.
The copy is shallow and the underlying data is destroyed when the original object goes out of scope.
We seem to be lucky if all the objects pointed to had the same type. In case of different types, we need to cast the pointers to a common base class pointer type and run into segmentation faults.


masterleinad:
Actually, just casting before is nicer.

PR #8795, opened 2019-09-19
Make a char[30] variable into a std::string.

### LINKED ISSUE #8764: Algorithms::TimestepControl has security problem
We have
```
  inline void
  TimestepControl::file_name_format(const char *fmt)
  {
    strcpy(format, fmt);
  }
```
but `format` is declared as
```
    /**
     * Output file name template.
     */
    char format[30];
```

So `strcpy` will in general not notice when it writes past the end. That's why the man page says this:
```
NAME
       strcpy, strncpy - copy a string

SYNOPSIS
       #include <string.h>

       char *strcpy(char *dest, const char *src);

       char *strncpy(char *dest, const char *src, size_t n);

DESCRIPTION
       The  strcpy()  function  copies the string pointed to by src, including
       the terminating null byte ('\0'), to the buffer  pointed  to  by  dest.
       The  strings  may  not overlap, and the destination string dest must be
       large enough to receive the copy.  Beware  of  buffer  overruns!   (See
       BUGS.)

[...]

BUGS
       If  the destination string of a strcpy() is not large enough, then any‐
       thing might happen.   Overflowing  fixed-length  string  buffers  is  a
       favorite  cracker technique for taking complete control of the machine.
       Any time a program reads or copies data  into  a  buffer,  the  program
       first  needs  to check that there's enough space.  This may be unneces‐
       sary if you can show that overflow is impossible, but be careful:  pro‐
       grams  can  get changed over time, in ways that may make the impossible
       possible.
```

bangerth:
This avoids the potential for buffer overflows.

Fixes #8764.

### DIFF
--- a/doc/news/changes/incompatibilities/20190918Bangerth
+++ b/doc/news/changes/incompatibilities/20190918Bangerth
@@ -1,0 +1,7 @@
+Changed: The type of the argument of the
+Algorithms::TimeStepControl::file_name_format() function that takes a
+string has been changed from `char *` to `std::string`. Similarly, the
+returned type of the function of same name that doesn't take an
+argument has undergone the same kind of change.
+<br>
+(Wolfgang Bangerth, 2019/09/18)
--- a/include/deal.II/algorithms/timestep_control.h
+++ b/include/deal.II/algorithms/timestep_control.h
@@ -190,12 +190,12 @@ namespace Algorithms
      * Set the output name template.
      */
     void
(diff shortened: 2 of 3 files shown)


bangerth:
I'll note that the change is, strictly speaking, incompatible. But since `char*` and `std::string` are easily converted into each other, I doubt that it's worth fussing too much over it. As we've seen with @rezarastak 's recent patches, this class can't have been used very much.

PR #8796, opened 2019-09-19
fixed the usage of deprecated std::uncaught_exception

### LINKED ISSUE #8793: std::uncaught_exception() is deprecated in c++17
[std::uncaught_exception() is deprecated in c++17 and should be replaced with std::uncaught_exceptions()](https://en.cppreference.com/w/cpp/error/uncaught_exception)

We didn't remove this deprecated function in #2882, but we can think about it now.

We use `std::uncaught_exception()` in three places inside the code.
```
$ grep uncaught_exception -n -r source/
source/base/mpi.cc:783:          if (std::uncaught_exception())
source/base/subscriptor.cc:77:      if (std::uncaught_exception() == false)
source/base/timer.cc:386:  if (std::uncaught_exception() && mpi_communicator != MPI_COMM_SELF)
```

The compiler should already give us a [warning](https://godbolt.org/z/R4TdNf) and I'm not sure why we don't see the warning when we build `deal.II`.

rezarastak:
fixes #8793

### DIFF
--- a/source/base/mpi.cc
+++ b/source/base/mpi.cc
@@ -780,7 +780,12 @@ namespace Utilities
 #ifdef DEAL_II_WITH_MPI
       if (job_supports_mpi() == true)
         {
-          if (std::uncaught_exception())
+#  if __cpp_lib_uncaught_exceptions >= 201411
+          // std::uncaught_exception() is deprecated in c++17
+          if (std::uncaught_exceptions() > 0)
+#  else
+          if (std::uncaught_exception() == true)
+#  endif
             {
               std::cerr
                 << "ERROR: Uncaught exception in MPI_InitFinalize on proc "
--- a/source/base/subscriptor.cc
+++ b/source/base/subscriptor.cc
@@ -74,7 +74,12 @@ Subscriptor::check_no_subscribers() const noexcept
   // just display a message and continue the program.
   if (counter != 0)
     {
+#  if __cpp_lib_uncaught_exceptions >= 201411
+      // std::uncaught_exception() is deprecated in c++17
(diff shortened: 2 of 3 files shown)


source/base/mpi.cc
```diff
@@ -780,7 +780,12 @@ namespace Utilities
 #ifdef DEAL_II_WITH_MPI
       if (job_supports_mpi() == true)
         {
-          if (std::uncaught_exception())
+#  if __cplusplus >= 201703L
```
masterleinad:
```suggestion
#  if __cpp_lib_uncaught_exceptions >= 201411
```

source/base/subscriptor.cc
```diff
@@ -74,7 +74,12 @@ Subscriptor::check_no_subscribers() const noexcept
   // just display a message and continue the program.
   if (counter != 0)
     {
+#  if __cplusplus >= 201703L
```
masterleinad:
```suggestion
#  if __cpp_lib_uncaught_exceptions >= 201411
```

source/base/timer.cc
```diff
@@ -383,7 +383,12 @@ TimerOutput::~TimerOutput()
   // avoid communicating with other processes if there is an uncaught
   // exception
 #ifdef DEAL_II_WITH_MPI
-  if (std::uncaught_exception() && mpi_communicator != MPI_COMM_SELF)
+#  if __cplusplus >= 201703L
```
masterleinad:
```suggestion
#  if __cpp_lib_uncaught_exceptions >= 201411
```

rezarastak:
Thank you @masterleinad. Fixed it.

PR #8800, opened 2019-09-19
Speed up IndexSet::add_indices()

kronbichler:
While there, I also moved `IndexSet::add_range` to the header file because it is a rather small function body, especially for the (hopefully) common case that we append indices at the end. Noticed while working on #8772.

### DIFF
--- a/include/deal.II/base/index_set.h
+++ b/include/deal.II/base/index_set.h
@@ -1693,6 +1693,11 @@ IndexSet::add_indices(const ForwardIterator &begin, const ForwardIterator &end)
 
       tmp_ranges.emplace_back(begin_index, end_index);
       p = q;
+
+      // if the starting index of the next go-around of the for loop is less
+      // than the end index of the one just identified, then we will have at
+      // least one pair of ranges that are not sorted, and consequently the
+      // whole collection of ranges is not sorted.
       if (p != end && *p < end_index)
         ranges_are_sorted = false;
     }
--- a/include/deal.II/base/index_set.h
+++ b/include/deal.II/base/index_set.h
@@ -1636,12 +1636,49 @@ IndexSet::add_index(const size_type index)
 
 
 
+inline void
+IndexSet::add_range(const size_type begin, const size_type end)
+{
+  Assert((begin < index_space_size) ||
+           ((begin == index_space_size) && (end == index_space_size)),
+         ExcIndexRangeType<size_type>(begin, 0, index_space_size));
+  Assert(end <= index_space_size,
+         ExcIndexRangeType<size_type>(end, 0, index_space_size + 1));
+  Assert(begin <= end, ExcIndexRangeType<size_type>(begin, 0, end));
+
+  if (begin != end)
+    {
+      const Range new_range(begin, end);
+
+      // the new index might be larger than the last index present in the
+      // ranges. Then we can skip the binary search
+      if (ranges.size() == 0 || begin > ranges.back().end)
+        ranges.push_back(new_range);
+      else
+        ranges.insert(Utilities::lower_bound(ranges.begin(),
+                                             ranges.end(),
+                                             new_range),
+                      new_range);
+      is_compressed = false;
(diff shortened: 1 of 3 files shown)


include/deal.II/base/index_set.h
```diff
@@ -1654,9 +1691,34 @@ IndexSet::add_indices(const ForwardIterator &begin, const ForwardIterator &end)
           ++q;
         }
 
-      add_range(begin_index, end_index);
+      tmp_ranges.emplace_back(begin_index, end_index);
       p = q;
+      if (p != end && *p < end_index)
+        ranges_are_sorted = false;
```
bangerth: Can you add a comment in front of the `if` here? I think what you're saying is that the starting index of the range we will identify in the next go-around of the `for` loop is less than the end index of the one just identified, then we will have at least one pair of ranges that are not sorted, and consequently the whole collection of ranges is not sorted.

include/deal.II/base/index_set.h
```diff
@@ -1654,9 +1691,34 @@ IndexSet::add_indices(const ForwardIterator &begin, const ForwardIterator &end)
           ++q;
         }
 
-      add_range(begin_index, end_index);
+      tmp_ranges.emplace_back(begin_index, end_index);
```
bangerth: Not wrong, but would be easier to read if you said
```suggestion
      tmp_ranges.emplace_back({begin_index, end_index});
```

include/deal.II/base/index_set.h
```diff
@@ -1654,9 +1691,34 @@ IndexSet::add_indices(const ForwardIterator &begin, const ForwardIterator &end)
...
       p = q;
+      if (p != end && *p < end_index)
+        ranges_are_sorted = false;
     }
+
+  if (!ranges_are_sorted)
+    std::sort(tmp_ranges.begin(), tmp_ranges.end());
```
bangerth: Interesting that this works. I guess it uses the lexicographic ordering of the pairs, which for non-overlapping ranges does the right thing...

kronbichler: It should also work for overlapping ranges because the lexicographic orderings goes by the first index. We are in an uncompressed state, so the only thing I could imagine is not sorted is if we have two ranges with the same starting index when the one with the higher ending index would be placed afterwards. But that is exactly a case for the merging in `compress()`.

include/deal.II/base/index_set.h
kronbichler: We have a pair of two integers which we construct from the two arguments - it won't work if we do `emplace_back` with an initializer list (the compiler complains about that). It might work to do `push_back` with the initializer list (I have not tried), but in that case I would prefer `emplace_back`. So I suggest to leave as is.

bangerth: Good point!

bangerth: One could also do `emplace_back (std::make_pair(...,...))`. But it's really an entirely unimportant point. I was just a bit confused about what is actually emplaced: Two integers, or a pair.

tjhei: Wolfgang, the whole point of emplace_back() is to forward the arguments and construct the object in-place. Why would you construct the pair first? I think of the arguments of emplace_back() as being "pasted" as the constructor arguments.
I think Martin was doing the right thing from the start.

tjhei: > I think of construct+move as about as good as construct-in-place ;-)

If you are happy with construct+move, you might as well just use push_back() or am I missing something? Saving the move makes a lot of sense to me and is the whole reason for emplace_back.

bangerth: What I was *trying* to say that the performance implications are zero in case of a POD type as we have here. So one might opt for the code that's easier to read -- which I thought might be the one that makes it clear that we're pushing a pair, not two integers.

Of course, if the object being emplaced is more complex, there is no question about any of your arguments!

PR #8802, opened 2019-09-19
Some bug fixed in p:f:t

peterrum:
I am integrating `p:f:t` into an application code and have noticed some bugs. ~~I have not finished yet...~~

Note: the output of some tests changed. I have ~~/will~~ added new tests...


source/distributed/fully_distributed_tria.cc
```diff
@@ -289,10 +289,10 @@ namespace parallel
     void
     Triangulation<dim, spacedim>::update_number_cache()
     {
+      parallel::Triangulation<dim, spacedim>::update_number_cache();
+
       if (settings & construct_multigrid_hierarchy)
         parallel::Triangulation<dim, spacedim>::fill_level_ghost_owners();
-
-      parallel::Triangulation<dim, spacedim>::update_number_cache();
```
bangerth: Ah, so you need to revert the order after all...

source/distributed/fully_distributed_tria.cc
```diff
@@ -511,8 +511,8 @@ namespace parallel
 
       const auto coarse_cell_id =
         coarse_cell_index_to_coarse_cell_id_vector[coarse_cell_index];
-      Assert(coarse_cell_id != numbers::invalid_coarse_cell_id,
-             ExcMessage("You are trying to access a dummy cell!"));
+      AssertThrow(coarse_cell_id != numbers::invalid_coarse_cell_id,
+                  ExcMessage("You are trying to access a dummy cell!"));
```
bangerth: Why do you need this change?

source/distributed/fully_distributed_tria_util.cc
```diff
@@ -444,6 +449,26 @@ namespace parallel
                     return false;
                   };
 
+                std::vector<bool> vertices_owned_by_locally_owned_cells_strong(
```
bangerth: You might want to add a comment to this variable and to the lambda just below that explains what "strong" refers to.

source/distributed/fully_distributed_tria_util.cc
```diff
@@ -494,12 +519,16 @@ namespace parallel
...
                       numbers::artificial_subdomain_id;
 
-                    if (is_locally_relevant(cell))
+                    if (cell->active() && is_locally_relevant_strong(cell))
+                      {
+                        cell_info.level_subdomain_id =
+                          cell->level_subdomain_id();
+                        cell_info.subdomain_id = cell->subdomain_id();
+                      }
+                    else if (is_locally_relevant(cell))
```
bangerth: Is this equivalent to `cell->is_locally_owned() || cell->is_ghost()`?

source/dofs/dof_handler.cc
```diff
@@ -855,7 +855,7 @@ DoFHandler<dim, spacedim>::DoFHandler(const Triangulation<dim, spacedim> &tria)
                                ParallelShared<DoFHandler<dim, spacedim>>>(
         *this);
   else if (dynamic_cast<
-             const parallel::distributed::Triangulation<dim, spacedim> *>(
+             const parallel::DistributedTriangulationBase<dim, spacedim> *>(
```
bangerth: I suspect that there are going to be many cases like this where we'll have to change the dynamic-cast. If you want, make this into a separate PR -- it's obviously correct -- and I'll approve this separately.

source/dofs/dof_handler_policy.cc
```diff
@@ -4405,9 +4405,19 @@ namespace internal
                 p4est_coarse_cell;
               internal::p4est::init_coarse_quadrant<dim>(p4est_coarse_cell);
 
+              types::coarse_cell_id coarse_cell_id = 0;
+              try
+                {
+                  coarse_cell_id = cell->id().get_coarse_cell_id();
+                }
+              catch (...)
```
bangerth: How can this fail? This probably deserves a comment.

peterrum:
@bangerth Thanks for the quick response!

source/dofs/dof_handler.cc
peterrum: It is quite annoying that there is not `else Assert(...)`. If this would have been there, I would have noticed my mistake a month ago... I think we have to look through all `dynamic_cast`s...

bangerth: Yes :-(

peterrum:
I have added matrixfree tests for `p:f:t`. FYI @kronbichler

peterrum:
@tjhei  @bangerth We have to return to our discussion regarding the dummy-cell (if one process has no locally relevant cell). In the following code (taken from source/dofs/dof_handler_policy.cc), we loop over all coarse cell and access the ID of each:

```cpp
          for (typename DoFHandlerType::level_cell_iterator cell =
                 dof_handler.begin(0);
               cell != dof_handler.end(0);
               ++cell)
            {
             // ...
              find_marked_mg_ghost_cells_recursively<dim, spacedim>(
                tria,
                cell->id().get_coarse_cell_id(),
                cell,
                p4est_coarse_cell,
                neighbor_cell_list);
             // ...
           }
```
Even of the dummy cell, which leads (as it is implemented right now) to an assert. To circumvent this, I have replaced the `Assert` with an `AssertThrow` and catch it here. This is not nice; but I do not see another solution if we want to keep the assert in `coarse_cell_index_to_coarse_cell_id`.

source/dofs/dof_handler.cc
peterrum: I would like to keep the bug-fixes of this PR together since they depend on each other (for example, one test crashed with these changes and was fixed after I have introduced the `catch` - see below).

peterrum:
@bangerth I have updated this PR. Thanks for your comments! Lets get this merged soon: these are quite significant bug fixes in `p:f:t` :(

tests/fullydistributed_grids/copy_serial_tria_04.mpirun=7.output
```diff
@@ -3,158 +3,158 @@ DEAL:0:1d::n_levels:                  1
...
+DEAL:1:2d::n_dofs:                             66049
+DEAL:1:2d::n_locally_owned_dofs:               9456
 DEAL:1:2d::
 DEAL:1:2d::n_levels:                  1
 DEAL:1:2d::n_cells:                   6556
 DEAL:1:2d::n_active_cells:            6556
 DEAL:1:2d::
-DEAL:1:2d::n_dofs:                             59555
-DEAL:1:2d::n_locally_owned_dofs:               59555
+DEAL:1:2d::n_dofs:                             274625
```
kronbichler: Without looking closely, this seems to be a rather big size for a test. Is this size necessary or is it enough with one level less?

tests/matrix_free/parallel_multigrid_fullydistributed.cc
```diff
@@ -0,0 +1,519 @@
...
+
+  MatrixFree<dim, number>                    data;
+  LinearAlgebra::distributed::Vector<number> inverse_diagonal_entries;
+};
+
+
+
+template <typename MatrixType>
+class MGTransferPrebuiltMF
+  : public MGTransferPrebuilt<LinearAlgebra::distributed::Vector<double>>
```
kronbichler: Can you derive this class from `MGTransferMatrixFree` which will make this run faster? Then you can also drop the Trilinos dependency in the output.

tests/matrix_free/parallel_multigrid_fullydistributed.with_mpi=true.with_p4est=true.mpirun=7.output
```diff
@@ -0,0 +1,289 @@
+
```
kronbichler: Is the 7 MPI rank case really necessary or is it enough to run it with 3 ranks? I'm asking because I would be conservative with tests on more than 5 MPI ranks or so. We have a few dozen, but should think a second before we add more, especially since the `parallel_multigrid_...` are quite heavy.

tests/fullydistributed_grids/copy_serial_tria_04.mpirun=7.output
peterrum: I have reduced the number of levels in some of the tests.

tests/matrix_free/parallel_multigrid_fullydistributed.cc
peterrum: Done.

tests/matrix_free/parallel_multigrid_fullydistributed.with_mpi=true.with_p4est=true.mpirun=7.output
peterrum: I have removed this test. I also have reduced the number of needed processes in the other test. There are 2 p:f:t tests left with more than 5 processes: one uses 9 processes (however is very cheap, since I am testing the corner case that there a processes that do not get any cells) and one uses 7 (to test the construction of `ConstructionData` in groups of size 3+3+1). I would like to keep these.

peterrum:
What does CI want to say with `Sending interrupt signal to process`?

tjhei:
Not sure, CI had some hiccups the last few days. This probably means a job got canceled.

PR #8804, opened 2019-09-19
Remove TimeStepControl::format altogether.

bangerth:
I had second thoughts about #8795 where I changed a `char[30]` variable to a `std::string` -- because it occurred to me that all I did was change getters and setters for the variable. I started to wonder whether that variable is actually used anywhere at all. It turns out that it's not -- the variable is just dead and can be removed.

I guess that class is just a mystery. Maybe we should just remove the whole class...

### DIFF
--- a/doc/news/changes/incompatibilities/20190918Bangerth
+++ b/doc/news/changes/incompatibilities/20190918Bangerth
@@ -1,7 +1,7 @@
-Changed: The type of the argument of the
-Algorithms::TimeStepControl::file_name_format() function that takes a
-string has been changed from `char *` to `std::string`. Similarly, the
-returned type of the function of same name that doesn't take an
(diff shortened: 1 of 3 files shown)


bangerth:
I'd favor removing (or, in a first step, deprecating) the class altogether and rebuilding it as a new class if that's something that you're interested in. That's because the class uses a naming scheme that we no longer use elsewhere. For example, using the same name for setters and getters is no longer in our current style. There are other pieces of the class that I think we would probably do differently -- like naming a function `print()` when it does not, in fact, print anything.

There's no reason not to just copy the class in a first step, but I'd like to start with a state where we think through the design and the naming of everything. It's hard to do that when you're tempted to keep things because you worry about people still using it.

PR #8750, opened 2019-09-13
Create p:f:ConstructionData by a selected group of processes

peterrum:
This PR is part of the effort to introduce the new `parallel::fullydistributed::Triangulation` (see #8558). This triangulation will be optionally created in two steps: 
1. create the `p:f:ConstructionData` by a set of master processes and 
2. distributed it to the other processes. 

The PR actually adds the function `parallel::fullydistributed::Utilities::create_construction_data_from_triangulation`, which uses the serialization functionalities introduced by PR #8746 and #8748.

PR #8806, opened 2019-09-19
Move comparator of CellData and ConstructionData out of the tests

peterrum:
As discussed with @bangerth. Once this is merged we can proceed with #8750.


peterrum:
Travis seems to be stuck...

peterrum:
Thanks @masterleinad !

PR #8807, opened 2019-09-20
Avoid calling the constructor for emplace_back

masterleinad:
Motivated by the comments in #8800. There really is no reason to call `emplace_back` constructing the variable of the type that shall be constructed in-place.


masterleinad:
https://godbolt.org/z/HgOwdD for the operators called for `push_back` and `emplace_back` if someone is interested.

PR #8808, opened 2019-09-20
Compute number of locally owned active cells per process on-demand

### LINKED ISSUE #8778: Remove parallel::TriangulationBase::NumberCache::n_locally_owned_active_cells
`parallel::TriangulationBase::NumberCache::n_locally_owned_active_cells` stores the number of locally owned active cells per MPI rank. This data structure does not scale with the number of processes.

The question is can we eliminate it? If not, we should only fill it need based. It would be nice to get rid of the `MPI_Allgather` in `parallelTriangulationBase::update_number_cache()`...

FYI @kronbichler

peterrum:
Do not update `number_cache.n_locally_owned_active_cells` in `TriangulationBase::update_number_cache()`. This data structure stores for every process the number of active cells...

If the data structure should be actually requested by the user (by calling `n_locally_owned_active_cells_per_processor`), the user should call `update_n_locally_owned_active_cells_per_processor()` first. Normally the data structure is only needed during debugging.

Once we agree that this is a useful change (and we want it), I will add some additional comments.

Closes #8778.

### DIFF
--- a/doc/news/changes/incompatibilities/20190920PeterMunch
+++ b/doc/news/changes/incompatibilities/20190920PeterMunch
@@ -1,0 +1,5 @@
+Changed: The parallel::TriangulationBase class does not store the number of active cells
+of all MPI processes any more. Instead, the information is computed on-demand when calling
+the function parallel::TriangulationBase::compute_n_locally_owned_active_cells_per_processor.
+<br>
+(Peter Munch, 2019/09/20)
--- a/include/deal.II/distributed/tria_base.h
+++ b/include/deal.II/distributed/tria_base.h
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 2008 - 2018 by the deal.II authors
+// Copyright (C) 2008 - 2019 by the deal.II authors
 //
 // This file is part of the deal.II library.
 //
@@ -79,10 +79,11 @@ namespace parallel
      * that contribute to this triangulation. The element of this vector
      * indexed by locally_owned_subdomain() equals the result of
      * n_locally_owned_active_cells().
+     *
+     * @note This function involves global communication!
      */
-    const std::vector<unsigned int> &
-    n_locally_owned_active_cells_per_processor() const;
-
+    std::vector<unsigned int>
+    compute_n_locally_owned_active_cells_per_processor() const;
 
     /**
      * Return the number of active cells in the triangulation that are locally
@@ -195,10 +196,9 @@ namespace parallel
     struct NumberCache
     {
       /**
-       * This vector stores the number of locally owned active cells per MPI
-       * rank.
+       * Number of locally owned active cells of this MPI rank.
        */
-      std::vector<unsigned int> n_locally_owned_active_cells;
+      unsigned int n_locally_owned_active_cells;
       /**
        * The total number of active cells (sum of @p
        * n_locally_owned_active_cells).
--- a/source/distributed/tria_base.cc
+++ b/source/distributed/tria_base.cc
@@ -56,7 +56,6 @@ namespace parallel
            ExcMessage("You compiled deal.II without MPI support, for "
                       "which parallel::TriangulationBase is not available."));
 #endif
-    number_cache.n_locally_owned_active_cells.resize(n_subdomains);
   }
 
 
@@ -99,8 +98,6 @@ namespace parallel
       MemoryConsumption::memory_consumption(mpi_communicator) +
       MemoryConsumption::memory_consumption(my_subdomain) +
       MemoryConsumption::memory_consumption(
-        number_cache.n_locally_owned_active_cells) +
-      MemoryConsumption::memory_consumption(
         number_cache.n_global_active_cells) +
       MemoryConsumption::memory_consumption(number_cache.n_global_levels);
     return mem;
@@ -117,7 +114,7 @@ namespace parallel
 
   template <int dim, int spacedim>
   TriangulationBase<dim, spacedim>::NumberCache::NumberCache()
-    : n_global_active_cells(0)
+    : n_locally_owned_active_cells(0)
     , n_global_levels(0)
   {}
 
@@ -125,7 +122,7 @@ namespace parallel
   unsigned int
   TriangulationBase<dim, spacedim>::n_locally_owned_active_cells() const
   {
-    return number_cache.n_locally_owned_active_cells[my_subdomain];
+    return number_cache.n_locally_owned_active_cells;
   }
 
   template <int dim, int spacedim>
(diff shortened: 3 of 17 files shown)


include/deal.II/distributed/tria_base.h
```diff
@@ -188,6 +191,13 @@ namespace parallel
      */
     types::subdomain_id n_subdomains;
 
+    /**
+     * Flag indicating if the vector n_locally_owned_active_cells has been
+     * filled. It is only filled need based since in the regular simulations
+     * this information is not needed.
+     */
+    bool n_locally_owned_active_cells_per_processor_has_been_updated;
```
kronbichler: Do you really need this variable? One should be able to query it from `n_locally_owned_active_cells.size()`

peterrum: That is a good idea!

peterrum:
Could you add `ready to test` label to see that I did not miss to update any tests...

kronbichler:
> Could you add ready to test label to see that I did not miss to update any tests...

I will highlight though only a small subset of the MPI tests is run (some MG transfer stuff IIRC). Better run all tests on our cluster.

include/deal.II/distributed/tria_base.h
```diff
@@ -199,6 +202,11 @@ namespace parallel
        * rank.
        */
       std::vector<unsigned int> n_locally_owned_active_cells;
+
+      /**
+       * Number of locally owned active cells of this MPI rank.
+       */
+      unsigned int my_n_locally_owned_active_cells;
       /**
```
bangerth:
```suggestion

      /**
```

include/deal.II/distributed/tria_base.h
```diff
@@ -83,6 +83,9 @@ namespace parallel
     const std::vector<unsigned int> &
     n_locally_owned_active_cells_per_processor() const;
 
+    void
+    update_n_locally_owned_active_cells_per_processor();
```
bangerth: Please add a comment and mention in the previous function that this function needs to be called to make the data available.

source/distributed/tria_base.cc
```diff
@@ -147,9 +146,61 @@ namespace parallel
...
     return number_cache.n_locally_owned_active_cells;
   }
 
+  template <int dim, int spacedim>
+  void
+  TriangulationBase<dim, spacedim>::
+    update_n_locally_owned_active_cells_per_processor()
+  {
+#ifdef DEAL_II_WITH_MPI
+    if (number_cache.n_locally_owned_active_cells.size() == 0)
```
bangerth: Is the size of this field reset to zero upon mesh refinement? Otherwise, you'd skip the computation if the field was still around from the previous refinement if that function had been called there.

source/distributed/tria_base.cc
```diff
@@ -162,16 +213,12 @@ namespace parallel
...
   {
-    Assert(number_cache.n_locally_owned_active_cells.size() ==
-             Utilities::MPI::n_mpi_processes(this->mpi_communicator),
-           ExcInternalError());
-
-    std::fill(number_cache.n_locally_owned_active_cells.begin(),
-              number_cache.n_locally_owned_active_cells.end(),
-              0);
-
+    number_cache.n_locally_owned_active_cells.clear();
```
bangerth: Ah yes, it is :-)

source/distributed/tria_base.cc
```diff
@@ -162,16 +213,12 @@ namespace parallel
...
     number_cache.ghost_owners.clear();
     number_cache.level_ghost_owners.clear();
+    number_cache.my_n_locally_owned_active_cells = 0;
+
+    // indicate that number_cache.n_locally_owned_active_cells is outdated
```
bangerth: what does this comment relate to?

source/distributed/tria_base.cc
```diff
@@ -147,9 +146,61 @@ namespace parallel
...
+  void
+  TriangulationBase<dim, spacedim>::
+    update_n_locally_owned_active_cells_per_processor()
+  {
+#ifdef DEAL_II_WITH_MPI
+    if (number_cache.n_locally_owned_active_cells.size() == 0)
+      {
+        // allocate memory
+        number_cache.n_locally_owned_active_cells.resize(
+          Utilities::MPI::n_mpi_processes(this->mpi_communicator));
```
bangerth: Is this code just copied from the original location?

peterrum:
@bangerth Thanks for the review!

> This is an incompatible change -- it should get an entry in the changelog as such.

I know ;) Before doing so, I was interested if people are fine with this  incompatible change. In the library, we only use it for debugging purposes...

bangerth:
Yes, I'm ok with the direction of this patch.

kronbichler:
One thing I want to remark is that we introduced `compute_locally_owned_dofs_per_processor()` that returns a vector of index sets in `DoFHandler` when we removed the storage of the full vector internally. The question is whether we see a reason to treat this class differently or if we also should go the route to let the user handle them and never store them. By the relatively small number of affected tests I think we might get around to not store them at all and have a `compute_...` function.

bangerth:
Ah, that's a good point. Consistency across classes is useful.

peterrum:
@bangerth @kronbichler OK then I remove the function and replace it with a `compute_`-function.

peterrum:
How do you like it this way? Much cleaner :)

doc/news/changes/incompatibilities/20190920PeterMunch
```diff
@@ -0,0 +1,5 @@
+Changed: The parallel::TriangulationBase class does not store the number of active cells
+of all MPI processes any more. Instead, the information is computed on-demand when calling
+the function parallel::TriangulationBase::compute_n_locally_owned_active_cells_per_processor.
```
bangerth: Ah, nice job sticking this into the base class!

include/deal.II/distributed/tria_base.h
```diff
@@ -79,10 +79,11 @@ namespace parallel
      * that contribute to this triangulation. The element of this vector
      * indexed by locally_owned_subdomain() equals the result of
      * n_locally_owned_active_cells().
+     *
+     * @note This function involves global communication!
      */
-    const std::vector<unsigned int> &
-    n_locally_owned_active_cells_per_processor() const;
-
+    const std::vector<unsigned int>
```
bangerth:
```suggestion
    std::vector<unsigned int>
```

source/distributed/tria_base.cc
```diff
@@ -143,11 +140,31 @@ namespace parallel
   }
 
   template <int dim, int spacedim>
-  const std::vector<unsigned int> &
-  TriangulationBase<dim, spacedim>::n_locally_owned_active_cells_per_processor()
-    const
+  const std::vector<unsigned int>
```
bangerth:
```suggestion
  std::vector<unsigned int>
```

source/distributed/tria_base.cc
```diff
@@ -143,11 +140,31 @@ namespace parallel
...
+                        n_locally_owned_active_cells_per_processor.data(),
+                        1,
+                        MPI_UNSIGNED,
+                        this->mpi_communicator);
+        AssertThrowMPI(ierr);
+      }
+
+    return n_locally_owned_active_cells_per_processor;
+#else
+    return std::vector<unsigned int>();
```
bangerth: This seems wrong -- we should return a vector of size 1 with `n_dofs()` as an element.

peterrum:
@bangerth Thanks for the review! I have added the changes.

Issue #8785, opened 2019-09-18
Speed up ComputeIndexOwner::Dictionary


kronbichler:
While working on #8772, I noticed two things we should improve for the dictionary class of `ComputeIndexOwner`:
1. `reinit()` is slower than it needs to be (and slow enough to be noticed in profiles) because it expands the indices of the locally owned range one by one, and computes the rank one by one. We should do this in terms of intervals. Unfortunately a bit of interval intersection stuff, but still manageable.
2. As already noted in #8772, we should introduce a minimal grain size in `dofs_per_process`.

Related to #8293.

peterrum:
The second point is easily made by replacing:
```cpp
dofs_per_process  = (size + n_procs - 1) / n_procs;
```
by:
```cpp
dofs_per_process  = std::max((size + n_procs - 1) / n_procs, min_dofs_per_process);
```
But in a second step, we should distributed more smarter the load among nodes/islands...

kronbichler:
> But in a second step, we should distributed more smarter the load among nodes/islands...

The question is how expensive that would be in terms of implementation and in terms of communication. My rationale is that it probably won't matter much even if communication is heavily skewed to the lower ranks: This is a setup routine and we won't need an optimal decision here; point-to-point is cheap also when 90% goes to rank 0 as in that case the global problem is small and not using too many ranks, either. But if you know something that is less than half an hour of work feel free to think about it.

peterrum:
> But if you know something that is less than half an hour of work feel free to think about it.

Essentially, we would need to inroduce `dofs_per_group < group_size * dofs_per_process`. Then it would look like this :
```cpp
          unsigned int
          dof_to_dict_rank(const types::global_dof_index i)
          {
            unsigned int i_group = i / dofs_per_group;
            unsigned int i_local = (i % dofs_per_group) / dofs_per_process;
            return i_group * group_size + i_local;
          }
```
Note: This took me 5 minutes (and I did not try it out)...

Note: I will introduce in one of my next PRs a function to determine the size a compute node.

kronbichler:
> Note: I will introduce in one of my next PRs a function to determine the size a compute node.

There is some infrastructure here:
https://github.com/dealii/dealii/blob/master/source/base/mpi.cc#L643-L682

You might want to look for something similar and/or adapt it so we can keep functionality in one place.

PR #8811, opened 2019-09-20
Avoid sorting a large array in MG fill_copy_indices

kronbichler:
This is one of the last performance improvements I plan for now, given an inspection of `step-37` and `step-59`, apart from the open issue #8785 and some of my open PRs.

In this PR, I replace adding some copy indices and later sorting them by adding them into an appropriate array with direct access, and later picking those numbers that were actually touched. This is for the case in the `copy_indices` where both the level dof and the global dof are locally owned, which is the most usual one with more than 95% of indices filled. The old solution of adding them at the back of a vector is still used for the rare indices global-mine/level-remote, where having a full vector does of course make less sense.

### DIFF
--- a/source/multigrid/mg_transfer_internal.cc
+++ b/source/multigrid/mg_transfer_internal.cc
@@ -98,16 +98,18 @@ namespace internal
       for (unsigned int level = 0; level < n_levels; ++level)
         {
           std::vector<bool> dof_touched(globally_relevant.n_elements(), false);
-          copy_indices[level].clear();
+
+          // for the most common case where copy_indices are locally owned
+          // both globally and on the level, we want to skip collecting pairs
+          // and later sorting them. instead, we insert these indices into a
+          // plain vector
+          std::vector<types::global_dof_index> unrolled_copy_indices;
+
           copy_indices_level_mine[level].clear();
           copy_indices_global_mine[level].clear();
 
-          typename dealii::DoFHandler<dim, spacedim>::active_cell_iterator
-            level_cell = mg_dof.begin_active(level);
-          const typename dealii::DoFHandler<dim, spacedim>::active_cell_iterator
-            level_end = mg_dof.end_active(level);
-
-          for (; level_cell != level_end; ++level_cell)
+          for (const auto &level_cell :
+               mg_dof.active_cell_iterators_on_level(level))
             {
               if (mg_dof.get_triangulation().locally_owned_subdomain() !=
                     numbers::invalid_subdomain_id &&
@@ -116,6 +118,10 @@ namespace internal
                    level_cell->subdomain_id() ==
                      numbers::artificial_subdomain_id))
                 continue;
+
+              unrolled_copy_indices.resize(
+                mg_dof.locally_owned_dofs().n_elements(),
+                numbers::invalid_dof_index);
 
               // get the dof numbers of this cell for the global and the
               // level-wise numbering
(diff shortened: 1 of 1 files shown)


source/multigrid/mg_transfer_internal.cc
```diff
@@ -98,16 +98,20 @@ namespace internal
...
-          const typename dealii::DoFHandler<dim, spacedim>::active_cell_iterator
-            level_end = mg_dof.end_active(level);
+          // for the most common case where copy_indices are locally owned
+          // both globally and on the level, we want to skip collecting pairs
+          // and later sorting them. instead, we insert these indices into a
+          // plain vector
+          std::vector<types::global_dof_index> unrolled_copy_indices;
 
-          for (; level_cell != level_end; ++level_cell)
+          copy_indices_level_mine[level]  = {};
```
masterleinad: I find `.clear();` much more often than `= {}`. Is there any reason for choosing this syntax?

source/multigrid/mg_transfer_internal.cc
```diff
@@ -117,6 +121,14 @@ namespace internal
                      numbers::artificial_subdomain_id))
                 continue;
 
+              if (has_active_cells_on_this_level == false)
```
masterleinad: Does this make any difference at all? I guess you could also remove this if-clause?

source/multigrid/mg_transfer_internal.cc
```diff
@@ -131,42 +143,67 @@ namespace internal
                         level, level_dof_indices[i]))
                     continue;
 
-                  types::global_dof_index global_idx =
-                    globally_relevant.index_within_set(global_dof_indices[i]);
-                  // skip if we did this global dof already (on this or a
-                  // coarser level)
-                  if (dof_touched[global_idx])
-                    continue;
```
masterleinad: Why do we need to move this?

source/multigrid/mg_transfer_internal.cc
```diff
@@ -131,42 +143,67 @@ namespace internal
...
+          // the expected format of a pair of indices.
+          if (has_active_cells_on_this_level)
+            {
+              copy_indices[level] = {};
+              if (copy_indices_global_mine[level].empty())
+                copy_indices[level].reserve(unrolled_copy_indices.size());
+              for (unsigned int i = 0; i < unrolled_copy_indices.size(); ++i)
+                if (unrolled_copy_indices[i] != numbers::invalid_dof_index)
+                  copy_indices[level].emplace_back(
+                    mg_dof.locally_owned_dofs().nth_index_in_set(i),
```
masterleinad: So `nth_index_in_set` here and `index_within_set` above are not overly expensive?

source/multigrid/mg_transfer_internal.cc
kronbichler: `std::vector::clear()` does not release previously allocated memory, so I could in principle imagine a case where we would hold more data from a previous round. This variable can be passed on from a member variable e.g. in `MGTransferPrebuilt`. However, that is not always the case, e.g. not for `MGTransferMatrixFree` which builds this from temporary data, so let us stick with the more common `clear()` and remove this micro optimization.

source/multigrid/mg_transfer_internal.cc
kronbichler: I removed the `if` clause and the variable as I can query the size only.

source/multigrid/mg_transfer_internal.cc
masterleinad: I expected `= {}` to not release memory either. At least in https://godbolt.org/z/qOH8-v, I see that the capacity is the same after the copy.

source/multigrid/mg_transfer_internal.cc
kronbichler: It is usually much cheaper to check the two cases `global_mine` and `level_mine` which go to the locally owned sets, which in most cases consists of a single range or at most a few only. The locally relevant dofs, on the other hand, contain many ranges due to the ghost layer. Therefore I decided to only do it in case we need to mark the indices.

kronbichler: I will soon push the code with a comment about this fact.

source/multigrid/mg_transfer_internal.cc
kronbichler: No, because they are on locally owned dofs; I added comments.

source/multigrid/mg_transfer_internal.cc
kronbichler: Alright, thanks for confirming. One more reason to stick to the old code.

PR #8812, opened 2019-09-20
Make TensorProductQuadrature a separate class.

simonsticko:
Presently a boolean is used internally to keep track of whether Quadrature<dim> is a tensor product or not. This makes Quadrature<dim> have two different states which suggests that the tensor product
functionality would be better off as a separate class. Add a new class TensorProductQuadrature and move the 1-dimensional base quadratures from Quadrture<dim> to this new class. Derive the different quadrature rules (QGauss, etc.) which are tensor products from TensorProductQuadrature.

### DIFF
--- a/include/deal.II/base/quadrature.h
+++ b/include/deal.II/base/quadrature.h
@@ -241,18 +241,116 @@ public:
   void
   serialize(Archive &ar, const unsigned int version);
 
+protected:
   /**
-   * This function returns true if the quadrature object is a tensor product
-   * of one-dimensional formulas and the quadrature points are sorted
-   * lexicographically.
+   * List of quadrature points. To be filled by the constructors of derived
+   * classes.
    */
-  bool
-  is_tensor_product() const;
+  std::vector<Point<dim>> quadrature_points;
 
   /**
-   * In case the quadrature formula is a tensor product, this function
-   * returns the @p dim one-dimensional basis objects.
-   * Otherwise, calling this function is not allowed.
+   * List of weights of the quadrature points.  To be filled by the
+   * constructors of derived classes.
+   */
+  std::vector<double> weights;
+};
+
+
+/**
+ * This function returns true if the incoming Quadrature<dim> is of type
+ * TensorProductQuadrature<dim>.
+ */
+template <int dim>
+bool
+is_tensor_product(const Quadrature<dim> &quadrature);
+
+
+/**
+ * This class represents a Quadrature<dim> which is the tensor product of a
+ * dim 1-dimensional quadratures. The main feature of this class is that it is
+ * possible to extract the base quadratures using the get_tensor_basis()
+ * function. Being able to access these base quadratures allows for
+ * optimizations which otherwise wouldn't be possible.
+ */
+template <int dim>
+class TensorProductQuadrature : public Quadrature<dim>
+{
+public:
+  /**
+   * Alias for a quadrature that acts on an object of one dimension
+   * less. Note that a SubTensorProductQuadrature is an SubQuadrature.
+   *
+   * Since a 1-dimensional quadrature can be considered a tensor product we
+   * let the SubTensorProductQuadrature for dim=2 to be an ordinary quadrature.
+   */
+  using SubTensorProductQuadrature =
+    typename std::conditional<(1 < dim),
+                              TensorProductQuadrature<dim - 1>,
+                              Quadrature<0>>::type;
+  /**
+   * Constructor.
+   *
+   * This constructor is marked as explicit to avoid involuntary accidents
+   * like in <code>hp::QCollection@<dim@> q_collection(3)</code> where
+   * <code>hp::QCollection@<dim@> q_collection(QGauss@<dim@>(3))</code> was
+   * meant.
+   */
+  explicit TensorProductQuadrature(const unsigned int n_quadrature_points = 0);
+
+  /**
+   * Build this quadrature formula as the tensor product of a formula in a
+   * dimension one less than the present and a formula in one dimension.
+   * This constructor assumes (and tests) that constant functions are integrated
+   * exactly, i.e. the sum of the quadrature weights is one.
+   *
+   * <tt>SubQuadrature<dim>::type</tt> expands to <tt>Quadrature<dim-1></tt>.
+   */
+  TensorProductQuadrature(const SubTensorProductQuadrature &,
+                          const Quadrature<1> &);
+
+  /**
+   * Build this quadrature formula as the <tt>dim</tt>-fold tensor product of
+   * a formula in one dimension.
+   *
+   * Assuming that the points in the one-dimensional rule are in ascending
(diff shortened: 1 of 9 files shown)


simonsticko:
Following the discussion in #8583.

One issue with this refactoring is that Quadrature<1> is no longer a tensor product quadrature.

include/deal.II/base/quadrature.h
```diff
@@ -241,18 +241,116 @@ class Quadrature : public Subscriptor
...
+   * less. Note that a SubTensorProductQuadrature is an SubQuadrature.
+   *
+   * Since a 1-dimensional quadrature can be considered a tensor product we
+   * let the SubTensorProductQuadrature for dim=2 to be an ordinary quadrature.
+   */
+  using SubTensorProductQuadrature =
+    typename std::conditional<(1 < dim),
+                              TensorProductQuadrature<dim - 1>,
+                              Quadrature<0>>::type;
+  /**
```
simonsticko: Could possibly let
Quadrature<2>::SubTensorProductQuadrature be Quadrature<1> ?
Not really sure what is the best.

include/deal.II/base/quadrature.h
```diff
@@ -241,18 +241,116 @@ class Quadrature : public Subscriptor
...
+   * Alias for a quadrature that acts on an object of one dimension
+   * less. Note that a SubTensorProductQuadrature is an SubQuadrature.
+   *
+   * Since a 1-dimensional quadrature can be considered a tensor product we
+   * let the SubTensorProductQuadrature for dim=2 to be an ordinary quadrature.
+   */
+  using SubTensorProductQuadrature =
+    typename std::conditional<(1 < dim),
+                              TensorProductQuadrature<dim - 1>,
+                              Quadrature<0>>::type;
```
simonsticko: Could possibly let
Quadrature<2>::SubTensorProductQuadrature be Quadrature<1> ?
Not really sure what is the best.

kronbichler:
I like making this state explicit, so there is support from my end. Of course, it would be nice if a 1d  quadrature formula could present itself as a tensor product formula - at the same time, whenever the information on tensor products is needed and one wants to do something with that, one should be fine to distinguish the 1D case. Do you have a particular case in mind where this would not be the case?

simonsticko:
Not sure if I should remove these constructors from `Quadrature` or let it remain...
``` c++
Quadrature(const SubQuadrature &, const Quadrature<1> &);

explicit Quadrature(const Quadrature<dim != 1 ? 1 : 0> &quadrature_1d);
```

simonsticko:
> Do you have a particular case in mind where this would not be the case?

No, I'm not sure it's a problem. It's just different from before.

include/deal.II/base/quadrature.h
```diff
@@ -241,18 +241,116 @@ class Quadrature : public Subscriptor
...
+};
+
+
+/**
+ * This function returns true if the incoming Quadrature<dim> is of type
+ * TensorProductQuadrature<dim>.
+ */
+template <int dim>
+bool
+is_tensor_product(const Quadrature<dim> &quadrature);
```
masterleinad: Can't you just write this as a member function for backward-compatibility similar to https://godbolt.org/z/W2VhZC or at least call this function.

simonsticko:
> Can't just do something like
> 
> ```
> template <int dim>
> class Quadrature;
> 
> template <int dim>
> class TensorProductQuadrature : public Quadrature<dim>
> {};
> 
> template<>
> class TensorProductQuadrature<1>
> {};
> 
> template<>
> class Quadrature<1> : public TensorProductQuadrature<1>
> {};
> ```

But then TensorProductQuadrature wouldn't be a Quadrature for dim==1. So the quadrature rules (QGauss etc.) that are tensor products would need to derive from TensorProductQuadrature<dim> when dim!=1 and from Quadrature<dim> when dim==1.

simonsticko:
I guess one could imagine something like the following to make sure Quadrature<1> is always a TensorproductQuadrature. But I'm not sure I like it. It's a bit too complicated...

```c++
  template <int dim>
  class Quadrature;

  /**
   * This is no longer a quadrature. Merely something that has a tensor basis.
   */
  template <int dim>
  class TensorProductQuadrature
  {
  public:
    const std::array<Quadrature<1>, dim> &
    get_tensor_basis() const;

    virtual ~TensorProductQuadrature() = default;
  };


  /**
   * Need an empty class to derive from for dim!=1.
   */
  class Void
  {};


  /**
   * Quadrature derives from TensorProductQuadrature if dim==1.
   */
  template <int dim>
  class Quadrature
    : public Subscriptor,
      virtual public std::
        conditional<(1 < dim), Void, TensorProductQuadrature<dim>>::type
  {
  public:
    virtual bool
    is_tensor_product()
    {
      return dynamic_cast<TensorProductQuadrature<dim> *>(this) != nullptr;
    }

    virtual ~Quadrature() = default;
  };


  /**
   * QGauss is a TensorProductQuadrature in addition to being a Quadrature.
   */
  template <int dim>
  class QGauss : public Quadrature<dim>,
                 virtual public TensorProductQuadrature<dim>
  {
  public:
  };

```

bangerth:
No, let's not use virtual base classes :-)

bangerth:
One possible way forward is to have class `TensorProductQuadrature` be derived from `Quadrature`, and then make sure that all current implementations of 1d quadrature classes are derived from `TensorProductQuadrature`. This way, while it isn't guaranteed that 1d quadratures show up as tensor products *by virtue of being 1d quadratures*, they are *as a matter of practice*. 

I tend to think that that achieves all practical aims of introducing this class, while also not being overly contorted.

bangerth:
@simonsticko -- any progress on this?

simonsticko:
Sorry for my late response. It seems like this solution will not work out after all.

Several classes (e.g. `FEValues`, `QCollection`) have member functions where an incoming `Quadrature<dim> &` is copied to a member variable `Quadrature<dim>`. This means that the tensor basis will be lost if the incoming object is a `TensorProductQuadrature`. 

It seems that one would need to rewrite a lot without gaining too much here... I'm leaning towards closing the pull request. Or what do you think @masterleinad, @bangerth?

bangerth:
Hm, yes, I see the problem. You may be right.

You had originally proposed this solution because there was an operation that led to an inconsistent state. Can you say again what that was? Could we, instead, change that operation so that it results in a consistent state again?

simonsticko:
1. Create a `Quadrature` which is a tensor product.
2. Call `initialize` with new points and weights.
Here, initialize should set `is_tensor_product_flag=false`. But right now it doesn't, so you can end up with a non-tensor product quadrature which has `is_tensor_product_flag=true`.

The original discussion is here: https://github.com/dealii/dealii/pull/8583#discussion_r318646014

The `initialize` function can also put the class `ImmersedSurfaceQuadrature` in an invalid state, where the vector of normals have different size from the points and weights.

PR #8818, opened 2019-09-20
Remove std::bind from hp

masterleinad:
It is widely agreed that there is no need to use `std::bind` ever from C++14 on. Performance-wise it is always inferior to lambdas. This pull request replaces all its uses in the `hp` namespace.

### DIFF
--- a/source/hp/dof_handler.cc
+++ b/source/hp/dof_handler.cc
@@ -1731,13 +1731,11 @@ namespace hp
   {
     // connect functions to signals of the underlying triangulation
     tria_listeners.push_back(this->tria->signals.pre_refinement.connect(
-      std::bind(&DoFHandler<dim, spacedim>::pre_refinement_action,
-                std::ref(*this))));
+      [&]() { this->pre_refinement_action(); }));
     tria_listeners.push_back(this->tria->signals.post_refinement.connect(
-      std::bind(&DoFHandler<dim, spacedim>::post_refinement_action,
-                std::ref(*this))));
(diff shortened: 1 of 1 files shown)


source/hp/dof_handler.cc
```diff
@@ -1731,13 +1731,11 @@ namespace hp
   {
     // connect functions to signals of the underlying triangulation
     tria_listeners.push_back(this->tria->signals.pre_refinement.connect(
-      std::bind(&DoFHandler<dim, spacedim>::pre_refinement_action,
-                std::ref(*this))));
+      [&]() { this->pre_refinement_action(); }));
```
bangerth: How is this actually implemented by compilers? Does the anonymous struct that is used to represent a lambda include *all* locally visible symbols, or just the ones actually referenced in the lambda itself?

I'm asking because this happens to be an object that has a long lifetime: Until the triangulation dies. If it stores references to all local variables, then that's a lot. I suspect that it doesn't, but it might be easier to just write `[this]` here.

masterleinad: No, the compiler can figure out which variables are actually needed.

bangerth: And I suspect it does, right?

bangerth: I mean not only does the standard allow it, but the compilers actually do that. (In reality, I can't imagine anything else -- the objects would be exceptionally large otherwise...)

masterleinad: Yes, but I agree that is better to be specific. Will change.

PR #8819, opened 2019-09-20
Constructors of Subscriptor simplified

rezarastak:


drwells:
I agree with @bangerth: if I recall correctly we discussed this as a possible style change when we moved to requiring C++11 but decided not to go through with it.

I will put together a quick PR to mention this in the code style document.

rezarastak:
Okay I understand. Thanks for the explanation.

PR #8821, opened 2019-09-20
Improve coding conventions

drwells:
Quick followup to #8819: I also fixed a few minor things in the coding conventions document while I was there.

### DIFF
--- a/doc/doxygen/headers/coding_conventions.h
+++ b/doc/doxygen/headers/coding_conventions.h
@@ -119,7 +119,36 @@ executable.
   then functions.
   <br>
   Exceptions shall be declared at the end of the public section
-  before the non-public sections start.</li>
+  before the non-public sections start.
+  <br>
+  We do not use the C++11-style class member initialization for member variables
+  that are neither <code>static const</code> nor <code>static constexpr</code>;
+  i.e., instead of
+@code
+  class Foo
+  {
+    int a = 42;
+    int *b = nullptr;
+  };
+@endcode
+  write
+@code
+  class Foo
+  {
+    Foo();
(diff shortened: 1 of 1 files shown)


doc/doxygen/headers/coding_conventions.h
```diff
@@ -119,7 +119,33 @@ executable.
...
+    int a = 42;
+    int *b = nullptr;
+  };
+@endcode
+  write
+@code
+  class Foo
+  {
+    Foo();
+  };
```
bangerth: add the member variables here, without their initializers

doc/doxygen/headers/coding_conventions.h
```diff
@@ -119,7 +119,33 @@ executable.
   then functions.
   <br>
   Exceptions shall be declared at the end of the public section
-  before the non-public sections start.</li>
+  before the non-public sections start.
+  <br>
+  We do not use the C++11-style class member initialization for member variables
+  that are neither <code>static const</code> nor <code>static constexpr</code>;
+  i.e., instead of
```
bangerth: do we provide rationales for the other rules, or do we just say that that's what it is? if the former, feel free to add a sentence explaining why.

doc/doxygen/headers/coding_conventions.h
drwells: whoops.

doc/doxygen/headers/coding_conventions.h
drwells: We just enumerate the rules here: nothing else under 'style issues' gives an explanation.

PR #8824, opened 2019-09-20
The class DiscreteTime implemented.

rezarastak:
following the conversation from #8804 

The class `Algorithms::DiscreteTime` is implemented with minimal public interface. It allows us to keep track of time in a time-dependent problem. The goal is that this class eventually replaces the `Algorithms::TimestepControl` class. Currently, the class `Algorithms::TimestepControl` is being used only within the class `Algorithms::ThetaTimestepping`. Thus, after adding a small number of new features, `Algorithms::DiscreteTime` can completely replace `Algorithms::TimestepControl`.

Let me know your suggestions about the naming of the class and its methods, and whether it is correct to be placed inside the namespace `Algorithms`.

### DIFF
--- a/include/deal.II/base/discrete_time.h
+++ b/include/deal.II/base/discrete_time.h
@@ -1,0 +1,196 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_discrete_time_h
+#define dealii_discrete_time_h
+
+#include <deal.II/base/config.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * Provides a means to keep track of the simulation time in a time-dependent
+ * simulation. It manages stepping forward from an initial time to a final
+ * time. It also allows adjusting the time step size during the simulation.
+ * It is guaranteed that at all times the current simulation time is in the
+ * closed interval between the start time and the end time.
+ *
+ * You can loop over all of the time steps by using a for loop
+ * @code
+ *   for (DiscreteTime time(0., 1., 0.3);
+ *        time.get_current_time() != time.get_end_time();
+ *        time.advance_time())
+ *   {
+ *     // Insert simulation code here
+ *   }
+ * @code
+ *
+ * In the above example the time starts at $t = 0$. Assuming the time step
+ * $dt = 0.3$ is not modified inside the loop, the time is advanced to
+ * $t = 0.3$, $t = 0.6$, $t = 0.9$ and finally it reaches the end time at
+ * $t = 1.0$. Note that the final step size is automatically reduced to
+ * $dt = 0.1$ in order to ensure that we finish the simulation exactly at
+ * the specified end time.
+ *
+ * @author Reza Rastak, 2019
+ */
+class DiscreteTime
+{
+public:
+  /**
+   * Constructor
+   */
+  DiscreteTime(const double start_time,
+               const double end_time,
+               const double start_step_size);
+
+  /**
+   * Return the current time.
+   */
+  double
+  get_current_time() const;
+
+  /**
+   * Return the start time.
+   */
+  double
+  get_start_time() const;
+
+  /**
+   * Return the end of the time interval.
+   * The final time step ends exactly at this point. This exact floating-point
+   * equality is very important because it allows us to use the expression
+   * <code>time.get_current_time() != time.get_end_time()</code> as the
+   * conditional statement in a for loop to check if the end time is reached.
+   */
+  double
+  get_end_time() const;
+
+  /**
+   * Return the size of the step from current time step to the next.
+   */
+  double
+  get_next_step_size() const;
+
+  /**
+   * Set the value of the next time step size. The next time advance_time()
+   * is called, the newly set @p time_step_size will be used to advance
+   * the simulation time. However, if the step is too large such that the next
+   * simulation time exceeds the end time, the step size is truncated.
+   * Additionally, if the step size is such that the next simulation time
+   * approximates the end time (but falls just slightly short of it), the step
+   * size is adjusted such that the next simulation time exactly matches the
+   * end time.
+   */
+  void
+  set_next_step_size(const double time_step_size);
+
+  /**
+   * Advance the current time based on the value of the current step.
+   * If you want to adjust the next time step size, call the method
+   * set_next_step_size() before calling this method.
+   * If you call this function repeatedly, the time
+   * is increased with the same step size until it reaches the end
+   * time. See the documentation of set_next_step_size() for explanation
+   * of the rules for automatic adjustment of the step size.
+   *
+   * @pre Current time must be smaller than the end time. The object cannot
+   * advance time if it is already at the end time. This rule is created to
+   * avoid the creation of an infinite loop when advance_time() is called
+   * inside a loop.
+   */
+  void
+  advance_time();
+
+  /**
+   * Set the current time equal to start time and set the step size to the
+   * initial step size.
+   */
+  void
+  restart();
+
+private:
+  /**
+   * The beginning of the time interval.
+   */
+  const double start_time;
+
+  /**
+   *The end of the time interval.
+   */
+  const double end_time;
+
+  /**
+   * The size of the first step.
+   */
+  const double start_step_size;
+
+  /**
+   * The current time.
+   */
+  double current_time;
+
+  /**
+   * The time at the next step.
+   *
+   * @note Internally, the next simulation time is stored instead of the
+   * current step size. For example, when the method set_next_step_size()
+   * is called, it computes the appropriate next simulation time and stores
+   * it. When advance_time() is called, the current_time is replaced by
+   * next_time. This choice for the internal state allows for simpler code
+   * and ensures than when we call advance_time() at the last step, the
+   * floating-point value of the time exactly matches the end time.
+   */
+  double next_time;
+};
+
+
+/*---------------------- Inline functions ------------------------------*/
+
+
+inline double
+DiscreteTime::get_start_time() const
+{
+  return start_time;
+}
+
+
+
+inline double
+DiscreteTime::get_end_time() const
+{
+  return end_time;
+}
+
+
+
+inline double
+DiscreteTime::get_next_step_size() const
+{
+  return next_time - current_time;
+}
+
+
+
+inline double
+DiscreteTime::get_current_time() const
+{
+  return current_time;
+}
+
+
+DEAL_II_NAMESPACE_CLOSE
+
+#endif
--- a/source/base/discrete_time.cc
+++ b/source/base/discrete_time.cc
@@ -1,0 +1,86 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+#include <deal.II/base/discrete_time.h>
+#include <deal.II/base/exceptions.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+namespace
+{
+  // Helper function that computes the next discrete time, adjusting it if:
+  //  - The next time exceeds the end time.
+  //  - The next time is smaller but very close to the end time.
+  double
+  get_next_time(const double current_time,
+                const double step_size,
+                const double end_time)
+  {
+    Assert(step_size >= 0., ExcMessage("Time step size must be non-negative"));
+    Assert(end_time >= current_time, ExcInternalError());
+    double           next_time          = current_time + step_size;
+    constexpr double relative_tolerance = 0.05;
+    const double     time_tolerance     = relative_tolerance * step_size;
+    if (next_time > end_time - time_tolerance)
+      next_time = end_time;
+    return next_time;
+  }
+} // namespace
+
+
+
+DiscreteTime::DiscreteTime(const double start_time,
+                           const double end_time,
+                           const double start_step_size)
+  : start_time{start_time}
+  , end_time{end_time}
+  , start_step_size{start_step_size}
+  , current_time{start_time}
+  , next_time{get_next_time(start_time, start_step_size, end_time)}
+{}
+
+
+
+void
+DiscreteTime::set_next_step_size(const double next_step_size)
+{
+  next_time = get_next_time(current_time, next_step_size, end_time);
+}
+
+
+
+void
+DiscreteTime::advance_time()
+{
+  Assert(next_time > current_time,
+         ExcMessage(
+           "You can't advance time further."
+           "Either dt == 0 or you are at the end of the simulation time."));
+  const double step_size = get_next_step_size();
+  current_time           = next_time;
+  next_time              = get_next_time(current_time, step_size, end_time);
+}
+
+
+
+void
+DiscreteTime::restart()
+{
+  current_time = start_time;
+  next_time    = get_next_time(current_time, start_step_size, end_time);
+}
+
+DEAL_II_NAMESPACE_CLOSE
--- a/doc/news/changes/minor/20190920RezaRastak
+++ b/doc/news/changes/minor/20190920RezaRastak
@@ -1,0 +1,4 @@
+New: The class DiscreteTime is created to keep track of time
+during a time-dependent simulation.
+<br>
+(Reza Rastak, 2019/09/20)
--- a/source/base/CMakeLists.txt
+++ b/source/base/CMakeLists.txt
@@ -25,6 +25,7 @@ SET(_unity_include_src
   bounding_box.cc
   conditional_ostream.cc
   convergence_table.cc
+  discrete_time.cc
   event.cc
   exceptions.cc
   flow_function.cc
--- a/tests/base/discrete_time_1.cc
+++ b/tests/base/discrete_time_1.cc
@@ -1,0 +1,84 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Basic tests for class DiscreteTime
+
+#include <deal.II/base/discrete_time.h>
+
+#include "../tests.h"
+
+void
+print_time(const DiscreteTime &time)
+{
+  deallog << "Current time = " << time.get_current_time()
+          << ", next step size = " << time.get_next_step_size() << std::endl;
+}
+
+void
+test_from_start_to_end()
+{
+  LogStream::Prefix p("Start to end");
+
+  DiscreteTime time(/*start_time*/ 0.,
+                    /*end_time*/ 1.5,
+                    /*start_step_size*/ 0.123);
+
+  deallog << "Start time = " << time.get_start_time() << std::endl;
+  deallog << "End time = " << time.get_end_time() << std::endl;
+  print_time(time);
+  while (time.get_current_time() != time.get_end_time())
+    {
+      time.advance_time();
+      print_time(time);
+    }
+  time.restart();
+  deallog << "Restarted" << std::endl;
+  print_time(time);
+  deallog << "OK" << std::endl;
+}
+
+void
+test_adjust_time_step_size()
+{
+  LogStream::Prefix p("Adjust time step size");
+
+  DiscreteTime time(/*start_time*/ 0.4,
+                    /*end_time*/ 2.1,
+                    /*start_step_size*/ 0.15);
+  print_time(time);
+  time.advance_time();
+  print_time(time);
+  time.set_next_step_size(0.36);
+  time.advance_time();
+  print_time(time);
+  time.advance_time();
+  print_time(time);
+  time.set_next_step_size(0.61);
+  time.advance_time();
+  print_time(time);
+  time.advance_time(); // Here we reach the end time.
+  print_time(time);
+  // If we call time.advance_time() one more time, it fails an assersion
+  deallog << "OK" << std::endl;
+}
+
+int
+main()
+{
+  initlog();
+  test_from_start_to_end();
+  test_adjust_time_step_size();
+  return 0;
+}
--- a/tests/base/discrete_time_1.output
+++ b/tests/base/discrete_time_1.output
@@ -1,0 +1,27 @@
+
+DEAL:Start to end::Start time = 0.00000
+DEAL:Start to end::End time = 1.50000
+DEAL:Start to end::Current time = 0.00000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.123000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.246000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.369000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.492000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.615000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.738000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.861000, next step size = 0.123000
+DEAL:Start to end::Current time = 0.984000, next step size = 0.123000
+DEAL:Start to end::Current time = 1.10700, next step size = 0.123000
+DEAL:Start to end::Current time = 1.23000, next step size = 0.123000
+DEAL:Start to end::Current time = 1.35300, next step size = 0.123000
+DEAL:Start to end::Current time = 1.47600, next step size = 0.0240000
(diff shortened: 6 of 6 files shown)


masterleinad:
How does this new class differ from `TimeStepControl`? I would likely be happier to deprecate functionality in there if it's just small differences.

rezarastak:
The old `TimestepControl` class does not have a single test, it uses obsolete method naming conventions, and has many unused and poorly documented private members. In #8804 , #8795, #8762, and #8789 we tried to improve it, but @bangerth [suggested](https://github.com/dealii/dealii/pull/8804#issuecomment-533260517) that we deprecate this class and create a new class with modern design.

For my own simulation needs I like some added functionality that the old `TimestepControl` cannot provide.

> 1. Allow step size to arbitrarily change during the simulation. We can add the method void step(double new_step)
> 1. Keep track of the time step number (step 1, step 2, step 3, ...)
> 1. Be able to go back in time by one step. This is useful if the convergence fails at some step, so we have to go back and try with a smaller time step size.

I like to get `DiscreteTime` to replace `TimestepControl` and then slowly add the new desired features.

masterleinad:
Fair enough. I can't take to the suggested name, though. What about `TimeTracking` or something that tells you that it is supposed to manage the time steps in a simulation?

include/deal.II/algorithms/discrete_time.h
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Constructor
+     */
+    DiscreteTime(const double start_time,
+                 const double final_time,
+                 const double start_step_size);
+
+    /**
+     * Return the end of the time interval.
+     * The final time step end exactly at this point.
```
bangerth:
```suggestion
     * The final time step ends exactly at this point.
```

include/deal.II/algorithms/discrete_time.h
```diff
@@ -0,0 +1,129 @@
...
+     * Return the size of the step from current time step to the next.
+     */
+    constexpr double
+    next_step_size() const;
+
+    /**
+     * Return the current time.
+     */
+    constexpr double
+    current_time() const;
```
bangerth: Move this function to before `final_time()`.

include/deal.II/algorithms/discrete_time.h
```diff
@@ -0,0 +1,129 @@
...
+    DiscreteTime(const double start_time,
+                 const double final_time,
+                 const double start_step_size);
+
+    /**
+     * Return the end of the time interval.
+     * The final time step end exactly at this point.
+     */
+    constexpr double
+    final_time() const;
```
bangerth: I think the common terminology would be "start time" and "end time". You might want to reflect this in the variable and function names.

include/deal.II/algorithms/discrete_time.h
```diff
@@ -0,0 +1,129 @@
...
+     * initial step size.
+     */
+    void
+    restart();
+
+  private:
+    /**
+     * The beginning of the time interval.
+     */
+    const double m_start_time;
```
bangerth: We don't generally use `m_` for member variables. Can we just name these variables `start_time` etc? You'll get into conflict with the member variables, so let's call these `get_xxx()` then.

source/algorithms/discrete_time.cc
```diff
@@ -0,0 +1,65 @@
...
+  DiscreteTime::DiscreteTime(const double start_time,
+                             const double final_time,
+                             const double start_step_size)
+    : m_start_time{start_time}
+    , m_final_time{final_time}
+    , m_start_step_size{start_step_size}
+    , m_current_time{start_time}
+    , m_next_time{start_time + start_step_size}
+  {
+    Assert(final_time > start_time, ExcInternalError());
```
bangerth:
```suggestion
    Assert(final_time >= start_time, ExcInternalError());
```

source/algorithms/discrete_time.cc
```diff
@@ -0,0 +1,65 @@
...
+    Assert(start_step_size <= final_time - start_time, ExcInternalError());
```
bangerth: Is this necessary? If there is only one time step to be had between start and end time, we'll just treat it like every other "last time step".

source/algorithms/discrete_time.cc
```diff
@@ -0,0 +1,65 @@
...
+
+    const double step_size = next_step_size();
+    m_current_time         = m_next_time;
+    m_next_time            = current_time() + step_size;
+
+    // update the next step size if needed
+    constexpr double relative_tolerance = 0.05;
+    const double     time_tolerance     = relative_tolerance * step_size;
+    if (m_next_time > final_time() - time_tolerance)
+      m_next_time = final_time();
```
bangerth: Why use the tolerance?

rezarastak:
Thank you for the feedback!
> What about TimeTracking or something that tells you that it is supposed to manage the time steps in a simulation?

@masterleinad I actually like the names `TimeControl` or `TimestepControl` more than `DiscreteTime`, I chose `DiscreteTime` just to be different from the old name `TimestepControl`. Since @bangerth suggested that we take it out of the `Algorithms` namespace, I can reuse the name `TimestepControl` without any name collision. How would that sound? Personally, I don't like the name `TimeTracking` because of the `ing` in the name :) 
>I like the class name, but I think it should just be in namespace dealii, not in Algorithms.

@bangerth Sure, I will change the namespace. Should I also move the file into the `base/` directory?
>I think it shouldn't be very hard to add a function that allows setting the time step size.

Ok I will add it in the same PR if that's fine.

include/deal.II/algorithms/discrete_time.h
rezarastak: Although I slightly dislike the verbosity of adding `get_` to the member functions, your suggestion makes perfect sense because of more readability and consistency. I will fix it.

bangerth:
Yes, I think moving the file into `base/` might actually be a nice idea as well. Alternatively, `numerics/` where a lot of this kind of stuff has lived in the past.

rezarastak:
I apologize for the long delay on this PR. In my latest commit I addressed all of the review comments. It is ready to be reviewed again.

source/algorithms/discrete_time.cc
rezarastak: Not it is not necessary. I fixed it.

source/algorithms/discrete_time.cc
rezarastak: If the calculated value of the next time is lower than end time but it is very close (closer than `time_tolerance`), then we directly advance to the end time.

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,132 @@
...
+   */
+  void
+  set_next_step_size(const double time_step_size);
+
+  /**
+   * Advance the current time based on the value of the current step.
+   * Afterward, it modifies the value of the time step size if needed.
+   */
+  void
+  advance_time();
```
bangerth: Can you say in which circumstances it modifies the time step size?

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,132 @@
...
+               const double start_step_size);
+
+  /**
+   * Return the current time.
+   */
+  double
+  get_current_time() const;
+
+  /**
+   * Return the ends of the time interval.
```
bangerth:
```suggestion
   * Return the end of the time interval.
```

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,132 @@
...
+
+#include <deal.II/base/exceptions.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * Provides a means to keep track of the simulation time in a time-dependent
+ * simulation. It manages stepping forward from an initial time to a final
+ * time. It only allows fixed time step size, although the step size may be
+ * automatically adjusted slightly to capture the final time.
```
bangerth: I don't think that last sentence is true any more, given that you have the `set_next_time_step()` function. Right?

source/base/discrete_time.cc
```diff
@@ -0,0 +1,78 @@
...
+  //  - The next time exceeds the end time.
+  //  - The next time is smaller but very close to the end time.
+  double
+  get_next_time(const double current_time,
+                const double step_size,
+                const double end_time)
+  {
+    Assert(end_time >= current_time, ExcInternalError());
+    double           next_time          = current_time + step_size;
+    constexpr double relative_tolerance = 0.05;
```
bangerth: I think the algorithm here should be documented somewhere in the documentation of the class.

source/base/discrete_time.cc
```diff
@@ -0,0 +1,78 @@
...
+    const double     time_tolerance     = relative_tolerance * step_size;
+    if (next_time > end_time - time_tolerance)
+      next_time = end_time;
+    return next_time;
+  }
+} // namespace internal
+
+DiscreteTime::DiscreteTime(const double start_time,
```
bangerth: Two more empty lines

source/base/discrete_time.cc
```diff
@@ -0,0 +1,78 @@
...
+  , start_step_size{start_step_size}
+  , current_time{start_time}
+  , next_time{internal::get_next_time(start_time, start_step_size, end_time)}
+{}
+
+
+void
+DiscreteTime::set_next_step_size(const double next_step_size)
+{
+  next_time = internal::get_next_time(current_time, next_step_size, end_time);
```
bangerth: That's a bit funny here because one would intuitively expect that setting the time step size is permanent. But with this algorithm, it really only applies to the very next time step and then is forgotten.

drwells:
@bangerth you have reviewed this more recently than I have. Can I assign you to it since you have a better idea of what is in this PR?

tjhei:
I haven't looked at the PR in detail but I have some questions.
1. We have various sets of timestepping related classes:
a) numerics/time_dependent.h: TimeStepBase, TimeDependent
b) base/time_stepping.h: TimeStepping::TimeStepping and Runge Kutta classes
c) algorithms/theta_timestepping.h: Algorithms::ThetaTimestepping
d) algorithms/timestep_control.h Algorithms::TimestepControl
I do not have an overview about what all of these do. Can you please investigate how old these are and if they are used/maintained/tested and how your class would fit in?

2. I know @Rombur worked on the Runge Kutta classes, so it would make sense to have his opinion about time stepping before we move forward with adding/removing functionality.

Rombur:
> For my own simulation needs I like some added functionality that the old TimestepControl cannot provide.
> 1.   Allow step size to arbitrarily change during the simulation. We can add the method void step(double new_step)
> 2.   Keep track of the time step number (step 1, step 2, step 3, ...)
> 3.   Be able to go back in time by one step. This is useful if the convergence fails at some step, so we have to go back and try with a smaller time step size.

This looks very close to what we are doing [here](https://dealii.org/developer/doxygen/deal.II/classTimeStepping_1_1EmbeddedExplicitRungeKutta.html). Did you take a look at [step-52](https://dealii.org/developer/doxygen/deal.II/step_52.html)? If something is missing, we could extend the class.

drwells:
Others have a better grasp on what we need in this class than I do so I am unassigning myself.

bangerth:
@rezarastak -- ping?

include/deal.II/base/discrete_time.h
rezarastak: I improved the documentation here.

include/deal.II/base/discrete_time.h
rezarastak: I fixed the documentation regarding this issue.

source/base/discrete_time.cc
rezarastak: I added more explanation on the algorithm.

source/base/discrete_time.cc
rezarastak: done

source/base/discrete_time.cc
rezarastak: Yes, that is slightly counter intuitive. That is why added some explanation to the member variable `next_time`.

rezarastak:
One question: how do I make sure that doxygen creates a new html page for the new class? I built the documentation but I was not able to search and find `DiscreteTime`.

rezarastak:
> We have various sets of timestepping related classes:
a) numerics/time_dependent.h: TimeStepBase, TimeDependent
b) base/time_stepping.h: TimeStepping::TimeStepping and Runge Kutta classes
c) algorithms/theta_timestepping.h: Algorithms::ThetaTimestepping
d) algorithms/timestep_control.h Algorithms::TimestepControl
I do not have an overview about what all of these do. Can you please investigate how old these are and if they are used/maintained/tested and how your class would fit in?

Thank you for asking this question! I had certainly studied these existing classes before proposing the new class `DiscreteTime`. One important thing to note is that these classes all have something to do with time and time-stepping; However, just by looking at their name, I (and maybe also others) don't have a clue what these classes do. I truly believe that we should do a better job of naming time-related classes.
I believe these existing classes can be divided into three categories:
1. `Algorithms::TimestepControl`: Manages how time is incremented. The proposed class `DiscreteTime` is a drop-in replacement for this class. We have had discussions about why `Algorithms::TimestepControl` has many unnecessary features, uses old naming standards and is not tested (see #8760, #8762, #8764, #8771 #8782, #8789, #8795, #8804) I suggest I replace all occurrence of `Algorithms::TimestepControl` with `DiscreteTime` and deprecate `Algorithms::TimestepControl`. The only advantage of `Algorithms::TimestepControl` over `DiscreteTime` is that is has a better, more descriptive class name.
2. The classes that are orthogonal to `DiscreteTime`: These classes are related to simulation time, but their role is totally orthogonal to the incrementation aspect of time, i. e. they do not perform $t_{n+1} = t_n + \Delta t_n$. For example:
    * `TimeDependent`: it manages the data that is stored for each time step. It does not decide on the size of the time step and it does not increment the time by a step, but it loops through a series of pre-determined times.
    * `TimeDependent::TimeSteppingData`: It determines it determines how many time steps ahead and how many time steps behind are needed to successfully solve the solution for the next step. This class does not increment time at all.
    * `TimeStepBase`: stores data for a single specified time and can solve primal or dual problem for that specific time. It does not increment time at all.
3. The classes that increment time in addition to performing other tasks. These classes do too many things and they can be simplified. Better separation of tasks can be achieved if they use `DiscreteTime` as a member variable or function input/output. For example:
    * `Algorithms::ThetaTimeStepping`: it stores `Algorithms::TimestepControl` as a member variable. We just have to replace it with a `DiscreteTime` object.
    * `TimeStepping::TimeStepping` and its subclasses including Runge-Kutta classes: These classes not only perform time-integration of functions, they also increment time. The time incrementation aspect can be delegated to `DiscreteTime` because `DiscreteTime` can handle adaptive time stepping better and can prevent us from overshooting the end time of the simulation. The usage of these classes in step 52:
```c++
for (unsigned int i = 0; i < n_time_steps; ++i)
  {
      time = explicit_runge_kutta.evolve_one_time_step(
        [this](const double time, const Vector<double> &y) {
          return this->evaluate_diffusion(time, y);
        },
        time,
        time_step,
        solution);
  }
```
can be replaced by 
```c++
for (DiscreteTime time(initial_time, final_time, time_step);
     time.get_current_time() != time.end_time();
     time.advance_time())
  {
      explicit_runge_kutta.evolve_one_time_step(
        [this](const double time, const Vector<double> &y) {
          return this->evaluate_diffusion(time, y);
        },
        time,
        solution);
  }
```
or
```c++
for (DiscreteTime time(initial_time, final_time, time_step);
     time.get_current_time() != time.end_time(); )
  {
      explicit_runge_kutta.evolve_one_time_step(
        [this](const double time, const Vector<double> &y) {
          return this->evaluate_diffusion(time, y);
        },
        time,
        solution); //performs time.advance_time() internally
  }
```

rezarastak:
> This looks very close to what we are doing here. Did you take a look at step-52? If something is missing, we could extend the class.

As I explained in my last comment, the concept of time incrementation (t = t + D t) in step 52 can be offloaded to `DiscreteTime`. However, this current PR only implements the bare-bone features for incrementing time. `DiscreteTime` cannot be immediately used in step-52 because it currently does not store the variable `i` (the number of steps so far). From the code:
```c++
  TimeStepping::ExplicitRungeKutta<Vector<double>> explicit_runge_kutta(
    method);
  output_results(0, method);
  for (unsigned int i = 0; i < n_time_steps; ++i)
    {
      time = explicit_runge_kutta.evolve_one_time_step(
        [this](const double time, const Vector<double> &y) {
          return this->evaluate_diffusion(time, y);
        },
        time,
        time_step,
        solution);
      constraint_matrix.distribute(solution);
      if ((i + 1) % 10 == 0)
        output_results(i + 1, method);
    }
```
you can see that we have to keep track of the variable `i` in order to decide when we can output the results. So, after this PR is approved and merged, I will submit another PR to make sure the step number is stored in `DiscreteTime`. Let me know if I should do it within this PR.

Rombur:
Thanks for the explanation. Like you said, the current time-stepping capabilities in deal.II are pretty confusing so we should make sure that anything that's added does not make the situation worse. The first thing is the name, `DiscreteTime` is not a great name. I think that it should be changed. Second, you have moved the class out of `algorithm`, what about putting it in `TimeStepping`? That way, it will be grouped with the other `TimeStepping` schemes in the doxygen. I think a better integration with `Runge-Kutta` would be really useful but that can be done in a different PR.

rezarastak:
I know that choosing the proper name for this class and putting it in a good namespace within a convenient directory is very important. In the numerous replies of this thread, there are multiple suggestions for the change of name and I am not sure which guidance to follow. So I like to ask for your opinion on the name and namespace and directory again. Please note that the name should be appropriate for the features that I have included in this PR as well as the following features that I will introduce in future PRs.
 * New self-explanatory methods that are not included here but I plan to add such as: `get_step_number()`, `get_previous_time()`, `get_previous_step_size()`, `is_at_start()`, `is_at_end()`.
 * Providing an interface where users of the library can attach custom **actions** to the objects of this class. These **actions** can be specified to automatically run at the beginning of each time step or at the end of each time step. An optional custom **condition** can also be provided. For example, we can attach the post-processing function that generates `vtk` files as an **action** to this class and provide a **condition** that says automatically run this every 10 steps. The **condition** in this example is `time.get_step_number() % 10 == 0`.
 * Providing an interface for automated adjustments of time step size. I might introduce a new class that to adjust `dt` of `DiscreteTime` objects. These `dt` adjustment objects decide, on the fly, whether to increase or decrease the size of the time step and they decide how much they want to change it. They can make these decisions by looking at convergence results of a solver or the value of a time-integration error estimate. They can also be used to ensure the stability of the explicit time-stepping methods that enforce a maximum allowable size for `dt`.

These future features make this class very extensible and customizable while still keeping it simple. It will be able to improve the readability of the code in many tutorials. Right now I am working on adapting step-21 to use `DiscreteTime`. We should choose the name and the namespace considering these future features and tutorial integrations.

rezarastak:
If you think `DiscreteTime` is not a good descriptive name, please specify which alternative names you are fine with. It is better if you specify multiple names.
Some suggested names other than `DiscreteTime`: `Time`, `TimeTracker`, `TimestepControl`, `TimeControl`, `TimeTracking`, `TimeStepper`
Feel free to suggest new ones not in the list.

If you think `dealii` is not a good namespace for this class, please specify which alternative namespaces you are fine with.
Some suggested namespaces other than `dealii` : `dealii::Algorithms`, `dealii::Timestepping`
Feel free to suggest new namespaces not on the list.

If you think `base/` is not a good directory, please specify the other directory locations you are fine with.
Some suggested directories other than `base/`: `numerics/`, `algorithms/`

Thank you.

rezarastak:
After thinking about the naming options, I think the best is to call this class `TimeStepping::Time` and include it within the file `base/time_stepping.h` next to other existing classes such as `TimeStepping::RungeKutta`. If everyone is fine with that, I will make this change.

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * Provides a means to keep track of the simulation time in a time-dependent
+ * simulation. It manages stepping forward from an initial time to a final
+ * time. It also allows adjusting the time step size during the simulation.
+ * It is guaranteed that at all times the current simulation time is in the
+ * closed interval between the start time and the end time.
+ *
+ * You can loop over all of the time steps by using a for loop
+ * @code
+ *   for (DiscreteTime time(0., 1., 0.1);
```
bangerth, line 32: Maybe make this more obvious (and mention it in a comment):
```suggestion
 *   for (DiscreteTime time(0., 1., 0.3);
```
The issue here is that stepping forward by 0.3 every time doesn't hit the end time, but this class will do the right thing and adjust time step sizes at the end so that we do get to exactly the end time.

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,190 @@
...
+  double
+  get_next_step_size() const;
+
+  /**
+   * Set the value of the next time step size. The next time advance_time()
+   * is called, the newly set @p time_step_size will be used to advance
+   * the simulation time. However, if the step is too large such that the next
+   * simulation time exceeds the end time, the step size is truncated.
+   * Additionally, if the step size is such that the next simulation time
+   * approximates the end time, the step size is adjusted such that the
```
bangerth, line 86:
```suggestion
   * approximates the end time (but falls just slightly short of it), the step size is adjusted such that the
```

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,190 @@
...
+   */
+  void
+  set_next_step_size(const double time_step_size);
+
+  /**
+   * Advance the current time based on the value of the current step.
+   * If we are already at the end time, this function does nothing.
+   * Otherwise if the current time is less than the end time, the time
+   * is advanced to the next simulation time. If you want to adjust the
+   * next simulation time, call the method set_next_step_size() before
```
bangerth, line 97:
```suggestion
   * next time step size, call the method set_next_step_size() before
```

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,190 @@
...
+   * calling this method. If you call this function repeatedly, the time
+   * is increases with the same step size until it reaches the end
```
bangerth, line 99:
```suggestion
   * is increased with the same step size until it reaches the end
```

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,190 @@
...
+   * time. See the documentation of set_next_step_size() for explanation
+   * of the rules for automatic adjustment of the step size.
+   *
+   * @pre Current time must be smaller than the end time. The object cannot
```
bangerth, line 103: That's in contradiction to the statement above where you say that if you are already at the end time, nothing happens.

source/base/discrete_time.cc
```diff
@@ -0,0 +1,85 @@
...
+//
+// ---------------------------------------------------------------------
+
+
+#include <deal.II/base/discrete_time.h>
+#include <deal.II/base/exceptions.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+namespace internal
```
bangerth, line 22: Just make it an anonymous namespace:
```suggestion
namespace
```

source/base/discrete_time.cc
```diff
@@ -0,0 +1,85 @@
...
+  , current_time{start_time}
+  , next_time{internal::get_next_time(start_time, start_step_size, end_time)}
+{}
+
+
+
+void
+DiscreteTime::set_next_step_size(const double next_step_size)
+{
+  next_time = internal::get_next_time(current_time, next_step_size, end_time);
```
bangerth, line 59:
```suggestion
  next_time = get_next_time(current_time, next_step_size, end_time);
```

source/base/discrete_time.cc
```diff
@@ -0,0 +1,85 @@
...
+void
+DiscreteTime::advance_time()
+{
+  Assert(next_time > current_time,
+         ExcMessage(
+           "You can't advance time further."
+           "Either dt == 0 or you are at the end of the simulation time."));
+  const double step_size = get_next_step_size();
+  current_time           = next_time;
+  next_time = internal::get_next_time(current_time, step_size, end_time);
```
bangerth, line 73:
```suggestion
  next_time = get_next_time(current_time, step_size, end_time);
```

source/base/discrete_time.cc
```diff
@@ -0,0 +1,85 @@
...
+}
+
+
+
+void
+DiscreteTime::restart()
+{
+  current_time = start_time;
+  next_time = internal::get_next_time(current_time, start_step_size, end_time);
```
bangerth, line 82:
```suggestion
  next_time = get_next_time(current_time, start_step_size, end_time);
```

bangerth:
Please also squash or fixup the commits into either one, or:
* one for the changes in .h/.cc files
* one for the changelog entry
* one for the tests

rezarastak:
Thank you so much @bangerth for reviewing this class. I will make these changes in the next couple of days.

include/deal.II/base/discrete_time.h
```diff
@@ -0,0 +1,190 @@
...
+   */
+  DiscreteTime(const double start_time,
+               const double end_time,
+               const double start_step_size);
+
+  /**
+   * Return the current time.
+   */
+  double
+  get_current_time() const;
```
rezarastak, line 56:
```suggestion
  get_time() const;
```
On a separate note, @bangerth , do you think `get_time()` would be a shorter and better name for here?

bangerth: No, let's not worry about it. I think the name you have so far is concise and in contrast to `get_next_time()` or `get_previous_time()` that I wouldn't be surprised we eventually want.

bangerth:
I forgot to say that yesterday, but I'm really excited about the directions you outlined above for future extensions with the actions and signals!

include/deal.II/base/discrete_time.h
bangerth: @jppelteret  Yes. I have some specific language in my head with which to update the documentation of this class once merged -- but I think this has been dragging on for long enough that it's time to get it merged and make other improvements in follow-ups.

rezarastak:
I made the requested changes. This is ready for final review/merging.

bangerth:
Yes, finally through all of the testers!

bangerth:
@rezarastak I'm writing the documentation for the new step-19 tutorial program (#10301) where I'm using the `DiscreteTime` class and wanted to say thanks again for this addition. In particular, I think that the documentation you wrote, specifically about the "update" and "consistent" stages, is conceptually such a nice approach to things. I don't think I was ever fully clear about these concepts, and it's just a pleasure to be able to say in the documentation "See the discussion about the update and consistent stages in a simulation in the documentation of the `DiscreteTime` class." Really nicely done -- thank you!

rezarastak:
@bangerth I am very excited see the step-19 and the widespread usage of particles in dealii. Hopefully I can learn about these new features of the library. I am glad the `DiscreteTime` class and its extensive documentation have been beneficial in your work.

PR #8825, opened 2019-09-21
Remove std::bind from source

masterleinad:
These are the remaining replacements of `std::bind` by lambdas in the `source` directory.

### DIFF
--- a/source/base/quadrature_lib.cc
+++ b/source/base/quadrature_lib.cc
@@ -774,10 +774,9 @@ QSorted<dim>::QSorted(const Quadrature<dim> &quad)
 
   std::sort(permutation.begin(),
             permutation.end(),
-            std::bind(&QSorted<dim>::compare_weights,
-                      std::ref(*this),
-                      std::placeholders::_1,
-                      std::placeholders::_2));
+            [this](const unsigned int x, const unsigned int y) {
+              return this->compare_weights(x, y);
+            });
 
   // At this point, the variable is_tensor_product_flag is set
   // to the respective value of the given Quadrature in the base
--- a/source/dofs/dof_tools_constraints.cc
+++ b/source/dofs/dof_tools_constraints.cc
@@ -2819,9 +2819,7 @@ namespace DoFTools
 #endif
           }
 
-        WorkStream::run(
-          coarse_grid.begin_active(),
-          coarse_grid.end(),
+        auto worker =
           [coarse_component,
            &coarse_grid,
            &coarse_to_fine_grid_map,
@@ -2836,7 +2834,9 @@ namespace DoFTools
                                                        coarse_grid.get_fe(),
                                                        coarse_to_fine_grid_map,
                                                        parameter_dofs);
-          },
+          };
+
+        auto copier =
           [coarse_component,
            &coarse_grid,
            &weight_mapping,
@@ -2848,9 +2848,14 @@ namespace DoFTools
                                                     weight_mapping,
                                                     is_called_in_parallel,
                                                     weights);
-          },
-          scratch,
-          copy_data);
+          };
+
+        WorkStream::run(coarse_grid.begin_active(),
+                        coarse_grid.end(),
+                        worker,
+                        copier,
+                        scratch,
+                        copy_data);
 
 #ifdef DEAL_II_WITH_MPI
         for (std::size_t i = 0;
--- a/source/dofs/dof_tools_constraints.cc
+++ b/source/dofs/dof_tools_constraints.cc
@@ -2819,25 +2819,38 @@ namespace DoFTools
 #endif
           }
 
-        WorkStream::run(coarse_grid.begin_active(),
-                        coarse_grid.end(),
-                        std::bind(&compute_intergrid_weights_3<dim, spacedim>,
-                                  std::placeholders::_1,
-                                  std::placeholders::_2,
-                                  std::placeholders::_3,
-                                  coarse_component,
-                                  std::cref(coarse_grid.get_fe()),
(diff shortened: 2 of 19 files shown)


source/base/quadrature_lib.cc
```diff
@@ -774,10 +774,9 @@ QSorted<dim>::QSorted(const Quadrature<dim> &quad)
 
   std::sort(permutation.begin(),
             permutation.end(),
-            std::bind(&QSorted<dim>::compare_weights,
-                      std::ref(*this),
-                      std::placeholders::_1,
-                      std::placeholders::_2));
+            [this](const unsigned int x, const unsigned int y) {
```
tjhei: question: Why do you capture ``this`` like this explicitly? I haven't seen that used in deal.II before. I always just use ``[&]`` if something needs to be captured.

source/dofs/dof_tools_constraints.cc
```diff
@@ -2819,25 +2819,38 @@ namespace DoFTools
...
-                                  std::cref(coarse_grid.get_fe()),
-                                  std::cref(weight_mapping),
-                                  is_called_in_parallel,
-                                  std::ref(weights)),
-                        scratch,
-                        copy_data);
+        WorkStream::run(
+          coarse_grid.begin_active(),
+          coarse_grid.end(),
+          [coarse_component,
```
tjhei: this looks quite ugly. Would you mind moving the definition of the lambdas before ``run()``?

source/base/quadrature_lib.cc
masterleinad: Mainly to be explicit what needs to be captured. @bangerth liked that better in https://github.com/dealii/dealii/pull/8818#discussion_r326796217 and we also don't need to always capture by reference (we don't use references for small types, but copy instead).

source/numerics/data_out.cc
```diff
@@ -812,15 +812,19 @@ DataOut<dim, DoFHandlerType>::build_patches(
...
-                this,
-                std::placeholders::_1,
-                std::placeholders::_2,
-                /* no std::placeholders::_3, since this function doesn't
-                   actually need a copy data object -- it just writes everything
-                   right into the output array */
-                n_subdivisions,
-                curved_cell_region),
+      [this, n_subdivisions, curved_cell_region](
+        const std::pair<cell_iterator, unsigned int> *cell_and_index,
```
tjhei: same here, maybe move it outside?

source/base/quadrature_lib.cc
tjhei: okay, I didn't see that conversation. Fine with me.

kronbichler: To me the question is not capturing `this` but some of the other examples you have below where you list several variables to be captured: While not wrong, I feel one loses some of the charm of lambdas if one still has to declare which variables where used. As you said in https://github.com/dealii/dealii/pull/8818#discussion_r326817434 the compiler will typically get the scope of what is needed right. Thus my question: Do we want to encourage listing what gets captured rather than simply `[&]`. I'm certainly guilty of breaking that rule, so I guess it would be good to make a style decision at this point.

peterrum: Personally I like to use `[&]`. However, sometimes I prefer to explicitly specify what data structures to capture (which is in general a few) to make it clear to myself (and the reader) what the actual inputs are to the function. I think what to do should be the decision of the programmer (on the user hand we could definitely replace `[this]` by `[&]`).

masterleinad: I could argue either way. `[&]` is short but not necessarily concise and listing everything might be precise, but overly verbose. I would be fine to encourage capture by reference by default, but also listing variables you want to copy explicitly. If the capture default is not used, I don't see any reason to use it, i.e., I prefer `[this]` over `[&]` if only `this` is captured.

For me, the capture list is not that much of a problem. They are typically much shorter than the declaration of the arguments.

Anyway, I am happy to follow any consensus we find.

kronbichler: I like the flexible definition, so if someone wants to specify more than `[&]` then that is find as to transport an intent, and at the same time not require the verbose way in all circumstances.

bangerth: For the record, I don't feel strongly either way. We've generally erred on the side of being explicit about stuff in deal.II, but I don't feel like we *need* to be. 

In cases where we really just capture one or two variables, it might be nice to say so, but I wouldn't advocate for listing 5 or 8 variables in cases where we have a substantial amount of code in a lambda.

Do we actually need a policy, or could we leave the current set of patches as they are and just leave it to the authors in future cases to choose whichever way they prefer? We could of course try to be consistent, but we don't have *that* many lambdas that it's really going to matter. (Or maybe that counts as "famous last words"?)

bangerth: For reference, because someone mentioned copying small objects above: When you write
```
  [this]() {...}
```
or
```
  [=this]() {...}
```
then in both cases is the `this` *pointer* copied -- which is equivalent to capturing the `*this` object by *reference*.

kronbichler:
I think the consensus on capturing variables for lambdas is to do a case-by-case decision, so I see no reason to hold this PR back.

source/base/quadrature_lib.cc
masterleinad: For the sake of completeness,
```
 [=, this]() {...}
```
is only supported from C++20 on. 
If implicitly captured, `(*this)` is captured by reference even if the capture default is `=`.
From C++17 on
```
 [*this]() {...}
```
can be used to copy the whole object.

PR #8828, opened 2019-09-21
Assertion checks the wrong item size.

dougshidong:
If the covariant is not required, its size will be zero and the assertion will fail.

### DIFF
--- a/source/fe/mapping_fe_field.cc
+++ b/source/fe/mapping_fe_field.cc
@@ -838,7 +838,8 @@ namespace internal
 
         if (update_flags & update_volume_elements)
           {
-            AssertDimension(data.covariant.size(), data.volume_elements.size());
+            AssertDimension(data.contravariant.size(),
+                            data.volume_elements.size());
             if (cell_similarity != CellSimilarity::translation)
               for (unsigned int point = 0; point < data.contravariant.size();
                    ++point)
--- a/tests/mappings/mapping_fe_field_01.cc
+++ b/tests/mappings/mapping_fe_field_01.cc
@@ -77,7 +77,8 @@ test(const unsigned int degree)
   FEValues<dim, spacedim> fe_values(map_fe,
                                     fe_sys,
                                     quadrature_formula,
-                                    update_values | update_JxW_values);
+                                    update_values | update_JxW_values |
+                                      update_volume_elements);
 
   typename DoFHandler<dim, spacedim>::active_cell_iterator cell =
                                                              dof_sys


drwells:
You are absolutely right. Thanks a lot for the fix!

drwells:
@dougshidong It looks like your commit author name is set to `dougshidong`, but we require (for attribution purposes) your given and family names to be separate. Could you please amend your commit to have a space, i.e., `Doug Shidong`, or whatever alternative you would prefer? Let me know if you do not know how to do this or need help!

bangerth:
Separately, how did you find this? If you had a small code snippet that shows the problem, I'd convert this into a test case for you!

dougshidong:
I git force pushed to amend my commit's author name. Just realised that force-pushing might not be encouraged in those codes, so let me know if there was a better way.

To generate the assertion, the easiest way is to slightly modify line 80 of
tests/mappings/mapping_fe_field_01.cc
https://github.com/dealii/dealii/blob/master/tests/mappings/mapping_fe_field_01.cc#L80 
from
    update_values | update_JxW_values);
to
    update_values | update_JxW_values | update_volume_elements);

and it will hit the assertion.

I think I was trying to print out the cell Jacobians to find some other bug in my code. Wasn't sure if querying JxW would generate the whole dependencies, so I turned on "update_volume_elements".

masterleinad:
> I git force pushed to amend my commit's author name. Just realised that force-pushing might not be encouraged in those codes, so let me know if there was a better way.

It is fine to force push to your own branches realted to pull requests, in particular for squashing commits. The only problem is that GitHub does not send any notifications. Therefore, you should let us explicitly know that you changed something.

> To generate the assertion, the easiest way is to slightly modify line 80 of
> tests/mappings/mapping_fe_field_01.cc
> https://github.com/dealii/dealii/blob/master/tests/mappings/mapping_fe_field_01.cc#L80
> from
> update_values | update_JxW_values);
> to
> update_values | update_JxW_values | update_volume_elements);
> 
> and it will hit the assertion.

I would be fine if you just do the modification there (in this pull request). The flags are not strictly necessary in that code anyway.

bangerth:
Indeed, force pushing is what we do all the time? Want to do it one more time? If you squash the two commits (which really do not need to be preserved separately), you end up with just one commit that you could force push. Let us know when you're done!

dougshidong:
I had heard that force pushing was a bad habit. Though, I just read some workflow that encourages squashing commits. I guess it depends on the project's preferences.

I just rebased it and force pushed. I included a few commits past mine in the rebase (but didn't squash them), hence why 2 of `tjhei`'s commits show up? Not sure how to fix that.

bangerth:
If you do
```
  git checkout master
  git pull upstream master
  git checkout <mybranch>    # substitute branch name
  git rebase master
```
then the unwanted commits should also disappear.

bangerth:
This looks good now -- thank you!

On the question of force pushes: force pushed are bad if they happen on branches that other people use, because then their history no longer matches what is in the repository. This leads to difficult to resolve situations one wants to avoid, and as a consequence, one must under no circumstances force push into the deal.II repository, for example. But we're talking about a side branch of your personal repository. Nobody other than you is using it, and so nothing terrible can happen.

dougshidong:
Thank you all for the support on this first pull request.

bangerth:
Quite gladly, @dougshidong ! We hope it won't be your last pull request!

PR #8553, opened 2019-08-12
Update a bunch of changelog entries.

bangerth:
This updates a bunch of changelog entries added last week.

PR #8830, opened 2019-09-21
Collate changelog entries

drwells:
We previously discussed collating all of the small (i.e., 'fix a typo in X') changelog entries from the workshop. This supercedes #8553 by combining aforementioned changelog entries into one single entry.

I have a second PR that copies the changelog entries from #8553 into a new PR (while skipping the deleted ones and adding a few more fixes).


doc/news/changes/minor/20190810Workshop
```diff
@@ -0,0 +1,8 @@
+Improved: Make a large number of prose, grammar, and various other improvements
+to the library.
+<br>
+(Ashna Aggarwal, Mario Zepeda Aguilar, Manaswinee Bezbaruah, Katherine Cosburn,
```
tjhei: I thought the format needs to be ```(name, name, name, date)``? Do our scripts handle newlines in the list of names?

tjhei:
I went through the comments and collating them seems fine! Thanks.

doc/news/changes/minor/20190810Workshop
drwells: They do: I generated the doxygen and it looks fine (everything is on one line).

drwells:
I moved the changes suggested  in #8831 here.

doc/news/changes/minor/20190810Workshop
bangerth: The script @tjhei mentioned is the one we use for the release checklist and the release papers. It's in a separate repository. We'll probably have to adjust a few things by hand, but I think we'll remember when we need to do that.

drwells: I'll take a look.

PR #8832, opened 2019-09-22
Remove std::bind from include (apart from documentation)

masterleinad:


masterleinad:
I removed the changes for `thread_management.h` and will submit them separately.

PR #8834, opened 2019-09-22
Add coarsening/refinement capabilities to p:f:t

peterrum:
This PR is part of the effort to introduce the new `parallel::fullydistributed::Triangulation` (see #8558). It adds coarsening/refinement capabilities.

~~Note: I still have to figure out, how to treat the flags of ghost cells.~~


peterrum:
@tjhei Could you give me your first impression on the method `parallel::fullydistributed::Triangulation::execute_coarsening_and_refinement()`. Currently, I have two working tests:

tests

The first tests has 4 coarse cells and is run with 4 processes. The second test has 4 coarse cells and is run with 5 processes: it is well visible - that without re-partitioning - the starting configuration and the final configuration is not identical (which is expected because during every coarsening step information of the initial partitioning is lost).

Thanks in advance!

source/distributed/fully_distributed_tria.cc
```diff
@@ -301,8 +313,140 @@ namespace parallel
...
+                return true;
+              }
+            else
+              return false;
+          },
+          [](const TriaIterator<CellAccessor<dim, spacedim>> &cell,
+             const unsigned int                               refinement_flag) {
+            if (refinement_flag == true)
+              {
+                cell->clear_coarsen_flag();
```
tjhei: how do you get by with just a single bool, shouldn't you have refine|coarsen|keep ?

source/distributed/fully_distributed_tria.cc
```diff
@@ -301,8 +313,140 @@ namespace parallel
...
+            {
+              cell->clear_refine_flag();
+              cell->clear_coarsen_flag();
+
+              // start assumption:
+              if (cell->level() > 0)
+                cell->set_coarsen_flag();
+            }
+
+        // update refinement/coarsening flags of non-local cells via
```
tjhei: I think you need to call prepare_coarsening_and_refinement first, before you send this information.

source/distributed/fully_distributed_tria.cc
```diff
@@ -301,8 +313,140 @@ namespace parallel
...
+             active_cells.end()))
+          {
+            // cell has been refined
+            cell->set_subdomain_id(cell->parent()->level_subdomain_id());
+            cell->set_level_subdomain_id(cell->parent()->level_subdomain_id());
+          }
+        else if (active_cells.find(cell->id().template to_binary<dim>()) !=
+                 active_cells.end())
+          {
+            // nothing has happened with this cell: nothing to do (?)
```
tjhei: correct.

source/distributed/fully_distributed_tria.cc
```diff
@@ -301,8 +313,140 @@ namespace parallel
...
+            if (vertices_owned_by_loclly_owned_cells[cell->vertex_index(v)])
+              return true;
+          return false;
+        };
+
+        // mark all active cells which are not connected via a vertex to
+        // an active local cells as artificial
+        for (auto cell : this->cell_iterators())
+          if (cell->active() && !is_locally_relevant(cell))
+            cell->set_subdomain_id(numbers::artificial_subdomain_id);
```
tjhei: you probably want to mark active cells which are artificial for coarsening and try to coarsen them?

tjhei (commented):
This is a good start, but I am not convinced that you will have the correct ghost layer after an arbitrary coarsening step.

Can you add an Assert in debug mode that checks that all cells neighboring own cells are marked as ghosts?

peterrum:
@tjhei Thanks for your quick response. I will add some further tests. Any recommendations?

peterrum:
I have added two additional tests. In the first test, nothing happens: no flags are set and as a consequence no cell is coarsened/refined. In the second case, only one quadrant (4 coarse cells and 4 processes) is refined:

tests

kronbichler:
@peterrum do you still have this on your priority list? I think this would be an interesting feature, and I'm wondering how much would be missing to make this work with the current capabilities. Feel free to close if you think it will not be done in the near future, say the next year or so, and if you think there are several major issues pending.

bangerth:
@peterrum This isn't going to make it into 9.8, but is this still something we should keep open?

PR #8835, opened 2019-09-22
Fix boost problems with MSVC

masterleinad:
There were some boost files missing required for compiling with `MSVC`. 
Also, newer boost versions set `BOOST_NO_AUTO_PTR` themselves in `config.hpp`. Hence, we should not define the macro first in that case.

### DIFF
--- a/cmake/configure/configure_2_boost.cmake
+++ b/cmake/configure/configure_2_boost.cmake
@@ -63,8 +63,14 @@ MACRO(FEATURE_BOOST_CONFIGURE_COMMON)
   ELSE()
     ADD_FLAGS(CMAKE_REQUIRED_FLAGS "-Werror")
   ENDIF()
+  # The configure function is called only once. In case an externally provided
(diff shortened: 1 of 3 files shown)


cmake/configure/configure_2_boost.cmake
```diff
@@ -63,15 +63,19 @@ MACRO(FEATURE_BOOST_CONFIGURE_COMMON)
   ELSE()
     ADD_FLAGS(CMAKE_REQUIRED_FLAGS "-Werror")
   ENDIF()
+  LIST(APPEND CMAKE_REQUIRED_INCLUDES ${BOOST_INCLUDE_DIRS} ${BOOST_BUNDLED_INCLUDE_DIRS})
```
drwells: This appears to be necessary but I think we might want to add a comment explaining that the first one will be empty if we use the bundled boost (and, in the opposite circumstance, we will pick headers from the first directory if both are present). It took me a minute to understand this.

PR #8836, opened 2019-09-22
Inline functions because of bugs in hipcc

Rombur:
With this PR, #8621, and renaming `lac/exceptions` to `lac/petsc_exceptions`, I can compile deal.II with hipcc. There is still quite a bit of work before we can run on AMD hardware but at least, it compiles.

### DIFF
--- a/include/deal.II/hp/dof_faces.h
+++ b/include/deal.II/hp/dof_faces.h
@@ -343,6 +343,15 @@ namespace internal
     {}
 
 
+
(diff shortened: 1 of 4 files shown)


Rombur:
It just needs to be inlined

masterleinad:
Would you mind providing a comment then such that we don't forget why this is defined in class?

Rombur:
I've moved the implementation outside the class

PR #8838, opened 2019-09-23
Eliminate q_point from functor and other GPU functions

peterrum:
This PR removes the `q_point` and `dof` input argument from following GPU matrix-free functions:
- `get_value()`, `get_dof_value()`, `submit_value(const value_type &val_in)`
- `get_gradient()`, `get_value()`, `submit_gradient(const gradient_type &grad_in)`
- and the functor that is passed in `apply_quad_point_operations`

In all these functions, `q_point` and `dof` can simply be recomputed (and hopefully the compiler eliminates this index computation).

The tutorial step-64 looks like this:
```cpp
  template <int dim, int fe_degree>
  __device__ void HelmholtzOperatorQuad<dim, fe_degree>::
                  operator()(CUDAWrappers::FEEvaluation<dim, fe_degree> *fe_eval,
             const unsigned int                          q) const
  {
    fe_eval->submit_value(coef * fe_eval->get_value(q), q);
    fe_eval->submit_gradient(fe_eval->get_gradient(q), q);
  }
```
and could look like this with these changes:
```cpp
  template <int dim, int fe_degree>
  __device__ void HelmholtzOperatorQuad<dim, fe_degree>::
                 operator()(CUDAWrappers::FEEvaluation<dim, fe_degree> *fe_eval) const
  {
    fe_eval->submit_value(coef * fe_eval->get_value());
    fe_eval->submit_gradient(fe_eval->get_gradient());
  }

```

@kronbichler @Rombur @masterleinad  Up for discussions! Any objections?

### DIFF
--- a/include/deal.II/matrix_free/cuda_fe_evaluation.h
+++ b/include/deal.II/matrix_free/cuda_fe_evaluation.h
@@ -38,6 +38,25 @@ DEAL_II_NAMESPACE_OPEN
  */
 namespace CUDAWrappers
 {
+  namespace internal
+  {
+    /**
+     * Compute the dof/quad index for a given thread id, dimension, and
+     * number of points in each space dimensions.
+     */
+    template <int dim, int n_points_1d>
+    __device__ inline unsigned int
+    compute_index()
+    {
+      return (dim == 1 ?
+                threadIdx.x % n_points_1d :
+                dim == 2 ?
+                threadIdx.x % n_points_1d + n_points_1d * threadIdx.y :
+                threadIdx.x % n_points_1d +
+                    n_points_1d * (threadIdx.y + n_points_1d * threadIdx.z));
+    }
+  } // namespace internal
+
   /**
    * This class provides all the functions necessary to evaluate functions at
    * quadrature points and cell integrations. In functionality, this class is
@@ -134,48 +153,103 @@ namespace CUDAWrappers
     /**
      * Return the value of a finite element function at quadrature point
      * number @p q_point after a call to @p evaluate(true,...).
+     *
+     * @deprecated Use the version without parameters instead.
+     */
+    DEAL_II_DEPRECATED __device__ value_type
+                                  get_value(const unsigned int q_point) const;
+
+    /**
+     * Same as above, except that the quadrature point is computed from thread
+     * id.
      */
     __device__ value_type
-               get_value(const unsigned int q_point) const;
+               get_value() const;
 
     /**
      * Return the value of a finite element function at degree of freedom
      * @p dof after a call to integrate() or before a call to evaluate().
+     *
+     * @deprecated Use the version without parameters instead.
+     */
+    DEAL_II_DEPRECATED __device__ value_type
+                                  get_dof_value(const unsigned int dof) const;
+
+    /**
+     * Same as above, except that the local dof index is computed from the
+     * thread id.
      */
     __device__ value_type
-               get_dof_value(const unsigned int dof) const;
+               get_dof_value() const;
 
     /**
      * Write a value to the field containing the values on quadrature points
      * with component @p q_point. Access to the same fields as through @p
      * get_value(). This specifies the value which is tested by all basis
      * function on the current cell and integrated over.
+     *
+     * @deprecated Use the version without parameters instead.
+     */
+    DEAL_II_DEPRECATED __device__ void
+                       submit_value(const value_type &val_in, const unsigned int q_point);
+
+    /**
+     * Same as above, except that the quadrature point is computed from the
+     * thread id.
      */
     __device__ void
-    submit_value(const value_type &val_in, const unsigned int q_point);
+    submit_value(const value_type &val_in);
 
     /**
      * Write a value to the field containing the values for the degree of
      * freedom with index @p dof after a call to integrate() or before
      * calling evaluate(). Access through the same fields as through
      * get_dof_value().
+     *
+     * @deprecated Use the version without parameters instead.
+     */
+    DEAL_II_DEPRECATED __device__ void
+                       submit_dof_value(const value_type &val_in, const unsigned int dof);
+
+    /**
+     * Same as above, except that the local dof index is computed from the
+     * thread id.
      */
     __device__ void
-    submit_dof_value(const value_type &val_in, const unsigned int dof);
+    submit_dof_value(const value_type &val_in);
 
     /**
      * Return the gradient of a finite element function at quadrature point
      * number @p q_point after a call to @p evaluate(...,true).
+     *
+     * @deprecated Use the version without parameters instead.
+     */
+    DEAL_II_DEPRECATED __device__ gradient_type
+                                  get_gradient(const unsigned int q_point) const;
+
+    /**
+     * Same as above, except that the quadrature point is computed from the
+     * thread id.
      */
     __device__ gradient_type
-               get_gradient(const unsigned int q_point) const;
+               get_gradient() const;
 
     /**
      * Write a contribution that is tested by the gradient to the field
-     * containing the values on quadrature points with component @p q_point
+     * containing the values on quadrature points with component @p q_point.
+     *
+     * @deprecated Use the version without parameters instead.
+     */
+    DEAL_II_DEPRECATED __device__ void
+                       submit_gradient(const gradient_type &grad_in, const unsigned int q_point);
+
+
+    /**
+     * Same as above, except that the quadrature point is computed from the
+     * thread id.
      */
     __device__ void
-    submit_gradient(const gradient_type &grad_in, const unsigned int q_point);
+    submit_gradient(const gradient_type &grad_in);
 
     // clang-format off
     /**
@@ -187,11 +261,29 @@ namespace CUDAWrappers
      *   CUDAWrappers::FEEvaluation<dim, fe_degree, n_q_points_1d, n_components, Number> *fe_eval,
      *   const unsigned int                                                               q_point) const;
      * \endcode
+     * 
+     * @deprecated Use apply_for_each_quad_point() instead.
+     */
+    // clang-format on
+    template <typename Functor>
+    DEAL_II_DEPRECATED __device__ void
+                       apply_quad_point_operations(const Functor &func);
+
+    // clang-format off
+    /**
+     * Same as above, except that the functor @func only takes a single input 
+     * argument (fe_eval) and computes the quadrature point from the thread id.
+     *
+     * @p func needs to define
+     * \code
+     * __device__ void operator()(
+     *   CUDAWrappers::FEEvaluation<dim, fe_degree, n_q_points_1d, n_components, Number> *fe_eval) const;
+     * \endcode
      */
     // clang-format on
     template <typename Functor>
     __device__ void
-    apply_quad_point_operations(const Functor &func);
+    apply_for_each_quad_point(const Functor &func);
 
   private:
     types::global_dof_index *local_to_global;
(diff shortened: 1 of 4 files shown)


masterleinad:
Hmm... I would be fine if we can prove that the generated instructions are actually the same.

peterrum:
@masterleinad I will investigate that!

masterleinad:
OK. Looking at https://godbolt.org/z/-hLkl- I think we should be fine.

peterrum:
@masterleinad Thanks a lot! I will try to compile the code tomorrow on a GPU and update the doxygen-documentation and the tutorial (anything else?)!

masterleinad:
> @masterleinad Thanks a lot! I will try to compile the code tomorrow on a GPU and update the doxygen-documentation and the tutorial (anything else?)!

Thanks for spotting this opportunity! We should try to be backward-compatible, though. Hence, either provide a default argument or duplicate the functions. With the latter choice we can deprecate the old behavior more easily, though.

peterrum:
> This look OK to me from a GPU point of view. My main question is whether we could also remove that index from the CPU FEEvaluation object somehow, to ensure as little adaptation in user interfaces. Or, more precisely, have at least a version that can work without the index q to ensure compatibility.

I don't think that is trivial. In the CPU version, a lane is not the same as q_point/dof. We would still need to set explicitly the qpoint or have an iterator over all q_points, which internally sets the current qpoint...

Just as a side note: I really dislike that you have to specify as a user explicitly the q-point `fe_eval->get_value(q)` and you have a loop over all q-points (in the CPU code - there is never a clear separation of operations you want to perform on elements and quadrature points). This is why I really like it that we can live without these in the GPU code!

For now, I do not see a problem with the fact that the CPU/GPU version slightly diverge...

kronbichler:
On a second thought: What if we support `n_q_points_1d != fe_degree+1`? I think we could still partition by the q-points and leave some lanes empty when we read out the dof values. In the MG transfer class Karl assigned between 1 and 2^d dofs per thread, so in that case there was a need to consider some q-index, see for example [here](https://github.com/kalj/dealii-cuda/blob/master/matrix_free_gpu/mg_transfer_matrix_free_gpu.cu#L288-L332). The question is if we want to support different parallelization schemes than 1 thread per quadrature point, in which case we would need some index.

tjhei:
> For now, I do not see a problem with the fact that the CPU/GPU version slightly diverge...

I would like a) to be able to also run the GPU version on CPU only (vectorization over quadrature point might be beneficial for small cell counts) and b) ideally have the same interface for both vectorization schemes. Letting them diverge might not be a great solution unless we can define one abstraction that sits slightly above the two schemes.

Rombur:
I think this PR is fine. We cannot remove the current function right now anyway so we will need both functions for the foreseeable future. We can revisit later if we need to keep the quadrature point in the functor. ThisPR simplifies user code, so I think this is a good idea.

peterrum:
@masterleinad @Rombur I have added the old functions again (and added DEAL_II_DEPRECATED to them). Also, I have added some doxygen-documentation. Should I update the tutorial in a next PR?

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -384,6 +440,32 @@ namespace CUDAWrappers
...
+  FEEvaluation<dim, fe_degree, n_q_points_1d, n_components_, Number>::
+    get_value() const
+  {
+    const unsigned int q_point =
+      (dim == 1 ?
+         threadIdx.x % n_q_points_1d :
+         dim == 2 ?
+         threadIdx.x % n_q_points_1d + n_q_points_1d * threadIdx.y :
+         threadIdx.x % n_q_points_1d +
+             n_q_points_1d * (threadIdx.y + n_q_points_1d * threadIdx.z));
```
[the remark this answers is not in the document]
Rombur: Also make the function public. It can be used to clean up some code when computing the diagonal.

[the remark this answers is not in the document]
peterrum: I have updated the tests!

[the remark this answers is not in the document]
peterrum: I have moved the index computation into a new function. Ready!

bangerth:
For the record, I believe that we suppress deprecation warnings in the test suite. Not 100% sure about this, but I believe so.

peterrum:
@masterleinad @Rombur anything missing here?

masterleinad:
> For the record, I believe that we suppress deprecation warnings in the test suite. Not 100% sure about this, but I believe so.

Correct.

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
      * Return the value of a finite element function at quadrature point
      * number @p q_point after a call to @p evaluate(true,...).
      */
+
```
masterleinad:
```suggestion
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
...
 
     /**
      * Write a value to the field containing the values for the degree of
      * freedom with index @p dof after a call to integrate() or before
      * calling evaluate(). Access through the same fields as through
      * get_dof_value().
      */
+    DEAL_II_DEPRECATED __device__ void
+                       submit_dof_value(const value_type &val_in, const unsigned int dof);
+
```
masterleinad:
```suggestion
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
      * Return the value of a finite element function at quadrature point
      * number @p q_point after a call to @p evaluate(true,...).
      */
```
masterleinad:
```suggestion
      * @deprecated Use the version without parameters instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
...
+     * id.
+     */
     __device__ value_type
-               get_value(const unsigned int q_point) const;
+               get_value() const;
 
     /**
      * Return the value of a finite element function at degree of freedom
      * @p dof after a call to integrate() or before a call to evaluate().
      */
```
masterleinad:
```suggestion
      * @deprecated Use the version without parameters instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
...
     __device__ value_type
-               get_dof_value(const unsigned int dof) const;
+               get_dof_value() const;
 
     /**
      * Write a value to the field containing the values on quadrature points
      * with component @p q_point. Access to the same fields as through @p
      * get_value(). This specifies the value which is tested by all basis
      * function on the current cell and integrated over.
      */
```
masterleinad:
```suggestion
      * @deprecated Use the version without parameters instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
...
     __device__ void
-    submit_value(const value_type &val_in, const unsigned int q_point);
+    submit_value(const value_type &val_in);
 
     /**
      * Write a value to the field containing the values for the degree of
      * freedom with index @p dof after a call to integrate() or before
      * calling evaluate(). Access through the same fields as through
      * get_dof_value().
      */
```
masterleinad:
```suggestion
      * @deprecated Use the version without quadrature point parameter instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
...
+     * thread id.
+     */
     __device__ void
-    submit_dof_value(const value_type &val_in, const unsigned int dof);
+    submit_dof_value(const value_type &val_in);
 
     /**
      * Return the gradient of a finite element function at quadrature point
      * number @p q_point after a call to @p evaluate(...,true).
      */
```
masterleinad:
```suggestion
      * @deprecated Use the version without parameters instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
...
+     * thread id.
+     */
     __device__ gradient_type
-               get_gradient(const unsigned int q_point) const;
+               get_gradient() const;
 
     /**
      * Write a contribution that is tested by the gradient to the field
      * containing the values on quadrature points with component @p q_point
      */
```
masterleinad:
```suggestion
      * @deprecated Use the version without quadrature point parameter instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -135,47 +151,87 @@ namespace CUDAWrappers
...
+    DEAL_II_DEPRECATED __device__ void
+                       submit_gradient(const gradient_type &grad_in, const unsigned int q_point);
+
     __device__ void
```
masterleinad: This function is missing documentation.

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -190,8 +246,24 @@ namespace CUDAWrappers
      */
     // clang-format on
     template <typename Functor>
+    DEAL_II_DEPRECATED __device__ void
+                       apply_quad_point_operations(const Functor &func);
```
masterleinad:
```suggestion
      * @deprecated Use apply_for_each_quad_point() instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -190,8 +246,24 @@ namespace CUDAWrappers
      */
```
masterleinad:
```suggestion
      * @deprecated Use apply_for_each_quad_point() instead.
      */
```

include/deal.II/matrix_free/cuda_fe_evaluation.h
```diff
@@ -38,6 +38,22 @@ DEAL_II_NAMESPACE_OPEN
  */
 namespace CUDAWrappers
 {
+  /**
+   * Compute the dof/quad index for a given thread id, dimension, and
+   * number of points in each space dimensions.
+   */
+  template <int dim, int n_points_1d>
+  __device__ inline unsigned int
+  compute_index()
```
Rombur: Thinking about this more, I think that we don't need this function to be public. Can you put it in `internal` namespace?

peterrum:
@masterleinad @Rombur done!

Rombur:
Great. Thanks @peterrum. For me this can be merged.

include/deal.II/matrix_free/cuda_fe_evaluation.h
masterleinad: You didn't delete this empty line between the documentation and the function. :wink:

peterrum: I definitely deleted a line, but I guess it was another ;) Now, its done!

masterleinad:
I will run the `CUDA` tests on it today and merge then.

masterleinad:
This pull request does not increase the number of failing `cuda` tests.

PR #8846, opened 2019-09-24
Avoid dynamic_cast in ghost/artificial cell queries

### LINKED ISSUE #8801: Definition of subdomain_id() on serial triangulation
While looking into #8772, I saw that `TriaAccessor::is_locally_owned()` etc all perform a `dynamic_cast` to see whether the triangulation is actually a more derived (parallel) class. This is reflected in the documentation https://www.dealii.org/current/doxygen/deal.II/classCellAccessor.html#a87a3ff6217f3b5b4537f6a101560f493 where we state that this is always true for serial triangulation. I find it confusing that we need the dynamic cast and why we do not have `cell->subdomain_id() == tria.locally_owned_subdomain()`. (I did a quick check on a serial case and realized that we have assigned 0 to the former and `-1` to the latter.)

I vaguely remember that we once mis-used the subdomain id also for other purposes, but thought we had stopped. Is there a reason to not simply have the two in a consistent state? Or does somebody remember what the design was?

As an alternative to the dynamic cast, I think the check for a serial triangulation would be cheaper if we simply asked `cell->tria->locally_owned_subdomain() == numbers::invalid_subdomain_id`. (`dynamic_cast` has to do some string operations to check for the actual class type which are non-negligible.)

kronbichler:
This replaces the `dynamic_cast` in the `CellAccessor::is_locally_owned` and friends functions by a query to the subdomain of the triangulation as discussed in #8801. While that is also a virtual function call, it is still much cheaper than finding out about the actual type of a class with virtual inheritance. As discussed in #8801, we do not intend to change the subdomain id of a serial triangulation at this point.

Closes #8801.

### DIFF
--- a/include/deal.II/grid/tria_accessor.templates.h
+++ b/include/deal.II/grid/tria_accessor.templates.h
@@ -22,8 +22,6 @@
 #include <deal.II/base/geometry_info.h>
 #include <deal.II/base/template_constraints.h>
 
-#include <deal.II/distributed/tria_base.h>
-
 #include <deal.II/grid/tria.h>
 #include <deal.II/grid/tria_accessor.h>
 #include <deal.II/grid/tria_faces.h>
@@ -34,15 +32,6 @@
 #include <cmath>
 
 DEAL_II_NAMESPACE_OPEN
-
-// Forward declaration
-#ifndef DOXYGEN
-namespace parallel
-{
-  template <int, int>
-  class TriangulationBase;
-}
-#endif
 
 
 /*--------------------- Functions: TriaAccessorBase -------------------------*/
--- a/include/deal.II/grid/tria_accessor.templates.h
+++ b/include/deal.II/grid/tria_accessor.templates.h
@@ -3566,17 +3566,10 @@ CellAccessor<dim, spacedim>::is_locally_owned() const
 #ifndef DEAL_II_WITH_MPI
   return true;
 #else
-  if (is_artificial())
-    return false;
 
-  const parallel::TriangulationBase<dim, spacedim> *pt =
-    dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
-      this->tria);
-
-  if (pt == nullptr)
-    return true;
-  else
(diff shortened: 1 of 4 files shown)


kronbichler:
I ran all tests with `mpi, grid, tria` in the name and they all passed. I will run the full testsuite now.

kronbichler:
All tests passing.

include/deal.II/grid/tria_accessor.templates.h
```diff
@@ -3566,17 +3566,10 @@ CellAccessor<dim, spacedim>::is_locally_owned() const
...
-  const parallel::TriangulationBase<dim, spacedim> *pt =
-    dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
-      this->tria);
 
-  if (pt == nullptr)
-    return true;
-  else
-    return (this->subdomain_id() == pt->locally_owned_subdomain());
+  return (this->tria->locally_owned_subdomain() ==
+            numbers::invalid_subdomain_id ||
```
bangerth: Can I ask you to prefix all of these comparisons with a comment that says that serial triangulations report `numbers::invalid_subdomain_id` for `tria->locally_owned_subdomain()`, and that consequently the first part of the condition indicates that the triangulation is serial where all cells are locally owned?

Same in the other places.

bangerth:
This looks good in principle, but I think a comment in all of the places of concern would help understand what the condition really tests.

kronbichler:
I added the suggested comments. Furthermore, I also removed the inclusion to the now obsolete header file `distributed/tria_base.h` in `tria_accessor.templates.h`. As a consequence, I had to add two includes elsewhere to make sure `parallel::TriangulationBase` (in `dof_handler.h`) and `ArrayView` (in `mapping.h`, that was previously transitively included via `tria_accessor.templates.h` -> `tria.h` -> `distributed/tria_base.h` -> `base/mpi.h` -> `base/array_view.h`). I'm re-running the test suite now because we might have depended on this include somewhere as well, given that we also did in the library.

PR #8853, opened 2019-09-25
Modify method accessibility in ShapeInfo

peterrum:
@kronbichler As far as I see it, we can do these changes.

### DIFF
--- a/include/deal.II/matrix_free/shape_info.h
+++ b/include/deal.II/matrix_free/shape_info.h
@@ -356,6 +356,7 @@ namespace internal
        */
       dealii::Table<2, unsigned int> face_to_cell_index_hermite;
 
+    private:
       /**
        * Check whether we have symmetries in the shape values. In that case,
        * also fill the shape_???_eo fields.
@@ -370,7 +371,7 @@ namespace internal
        * that save some operations in the evaluation.
        */
       bool
-      check_1d_shapes_collocation();
+      check_1d_shapes_collocation() const;
     };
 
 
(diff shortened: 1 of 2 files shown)


include/deal.II/matrix_free/shape_info.h
```diff
@@ -356,21 +356,22 @@ namespace internal
...
-
       /**
        * Check whether symmetric 1D basis functions are such that the shape
        * values form a diagonal matrix, i.e., the nodal points are collocated
        * with the quadrature points. This allows for specialized algorithms
        * that save some operations in the evaluation.
        */
       bool
-      check_1d_shapes_collocation();
+      check_1d_shapes_collocation() const;
```
kronbichler: Can you make this one private as well?

peterrum: Actually, I would like to use this function externally for a constant object (which is the reason why I stumbled into this function).

kronbichler: But wouldn't we then also possibly want to allow to look into the symmetric case with the `check_1d_shapes_symmetric()` call? It is on the same level of query. That would need some restructuring because we would need to separate the query about symmetric from filling additional `eo` fields (so the current naming is misleading already because it does not just check but also fill some fields).

But my actual question regarding public/private is that we have the result of those queries cached in the variable [element_type](https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/shape_info.h#L146) with e.g. `tensor_symmetric_collocation` for the case of collocation. Did you look for that part?

peterrum: I overlooked the `ElementType`; I will rather use that. Then, I can make `check_1d_shapes_collocation` also private!

PR #8854, opened 2019-09-25
Compute shape_gradients/hessians_collocation in ShapeInfo

peterrum:


peterrum:
@kronbichler Does this change require a change-log entry? It is internal struct...

peterrum:
@kronbichler now it should work. Please note, there will be a follow-up PR, where I introduce a lambda for the conversion to eo...

kronbichler:
> Does this change require a change-log entry? It is internal struct...

I don't think so - not at this point as long as it is internal.

include/deal.II/matrix_free/shape_info.h
```diff
@@ -196,14 +208,14 @@ namespace internal
       /**
        * Stores the shape gradients of the shape function space associated to
        * the quadrature (collocation), given by FE_DGQ<1>(Quadrature<1>). For
-       * faster evaluation only the even-odd format is necessary.
+       * faster evaluation it is stored in the even-odd format.
```
kronbichler: Maybe slightly reword this now that we have the same setup as with `shape_values`. How about writing the second sentence as "This array provides an alternative representation of the `shape_gradients_collocation` field in the even-odd format."

include/deal.II/matrix_free/shape_info.h
```diff
@@ -196,14 +208,14 @@ namespace internal
...
        */
       AlignedVector<Number> shape_gradients_collocation_eo;
 
       /**
        * Stores the shape hessians of the shape function space associated to
        * the quadrature (collocation), given by FE_DGQ<1>(Quadrature<1>). For
-       * faster evaluation only the even-odd format is necessary.
+       * faster evaluation it is stored in the even-odd format.
```
kronbichler: Same here.

kronbichler:
> How about you introduce the lambda right in this PR?

Just push a second commit with those changes.

peterrum:
@kronbichler  done!

peterrum:
@kronbichler Before I start to dig into the problem, do you have a hunch what might be the reason that one test fails....

include/deal.II/matrix_free/shape_info.templates.h
```diff
@@ -483,46 +456,50 @@ namespace internal
...
-            shape_gradients_eo[fe_degree / 2 * stride + q] =
-              shape_gradients[(fe_degree / 2) * n_q_points_1d + q];
-            shape_hessians_eo[fe_degree / 2 * stride + q] =
-              shape_hessians[(fe_degree / 2) * n_q_points_1d + q];
-          }
+      auto convert_to_eo = [](AlignedVector<Number> &      array_eo,
+                              const AlignedVector<Number> &array,
+                              const unsigned               n_points_dst,
+                              const unsigned               n_points_src) {
+        const unsigned int stride = (n_points_dst + 1) / 2;
```
kronbichler: You will need to rule out the case where `array.empty()` which is the case of `n_q_points_1d >= 200` at this stage. While here, I realized that we might want to change the lambda so it returns the `array_eo` instead? This will allow you to write `shape_values_eo = convert_to_eo(shape_values, n_q_points_1d, fe_degree+1)`.

PR #8857, opened 2019-09-25
Extend TensorProductMatrixSymmetricSum

jwitte08:
Expose the static lengths of arrays (set by a template parameter) and the underlying value type (similar to other matrix types).

### DIFF
--- a/include/deal.II/lac/tensor_product_matrix.h
+++ b/include/deal.II/lac/tensor_product_matrix.h
@@ -75,6 +75,17 @@ class TensorProductMatrixSymmetricSumBase
 {
 public:
   /**
+   * Type of matrix entries. This alias is analogous to <tt>value_type</tt>
+   * in the standard library containers.
+   */
+  using value_type = Number;
+
+  /**
+   * Compile-time array lengths given by the template parameter size.
+   */
+  static constexpr int static_size = size;
+
+  /**
    * Return the number of rows of the tensor product matrix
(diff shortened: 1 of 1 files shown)


jwitte08:
Actually, I am not satisfied with the name `size`. In the class `Tensor` there is the static constexpr member `n_independent_components` which provides nearly the meaning of the template parameter `size`. 

However, we have tensor products concerning matrices such that the number of independent variables is the square of `size`.

Setting the template `size` is only valid if the number of rows and columns coincide for all dimensions. Maybe renaming
`size` to `n_rows_1d`
`static_size` to `n_rows_1d_static`
is more descriptive.

kronbichler:
I agree, `n_rows_1d` would be a good name.

include/deal.II/lac/tensor_product_matrix.h
```diff
@@ -74,6 +74,17 @@ template <int dim, typename Number, int size = -1>
...
+  /**
+   * Type of matrix entries. This alias is analogous to <tt>value_type</tt>
+   * in the standard library containers.
+   */
+  using value_type = Number;
+
+  /**
+   * Compile-time array lengths given by the template parameter size.
+   */
+  static constexpr int static_size = size;
```
masterleinad:
```suggestion
  static constexpr int n_rows_1d_static = n_rows_1d;
```

PR #8859, opened 2019-09-26
Document direction parameter for periodic constraints

### LINKED ISSUE #8820: Better document 'direction' of DoFTools::make_periodicity_constraints
From the mailing list today:

> Also, I'm wondering if 
> someone can explain the meaning of 'direction' in the 
> DoFTools::make_periodicity_constraints function since it's not clear to 
> me what it does.
Good question -- I don't really know either. It's the same as in GridTools::collect_periodic_faces(), but the documentation there is also not much better. We may have to ask @tamiko (Matthias).

masterleinad:
Fixes #8820.


tamiko:
@masterleinad Thanks for taking care of that!

PR #8860, opened 2019-09-26
Fix SolutionTransfer when coarsening from FE_Nothing

masterleinad:
I was seeing some problems when trying to use adaptive refinement where some cells were using `FE_Nothing`. This pull request makes sure that coarsening works and the contributions from these cells are ignored.

### DIFF
--- a/source/dofs/dof_accessor_get.cc
+++ b/source/dofs/dof_accessor_get.cc
@@ -69,15 +69,23 @@ DoFCellAccessor<DoFHandlerType, lda>::get_interpolated_dof_values(
           // well, here we need to first get the values from the current
           // cell and then interpolate it to the element requested. this
           // can clearly only happen for hp::DoFHandler objects
-          Vector<number> tmp(this->get_fe().dofs_per_cell);
-          this->get_dof_values(values, tmp);
+          const unsigned int dofs_per_cell = this->get_fe().dofs_per_cell;
+          if (dofs_per_cell == 0)
+            {
+              interpolated_values = 0;
+            }
+          else
+            {
+              Vector<number> tmp(dofs_per_cell);
+              this->get_dof_values(values, tmp);
 
-          FullMatrix<double> interpolation(
-            this->dof_handler->get_fe(fe_index).dofs_per_cell,
-            this->get_fe().dofs_per_cell);
-          this->dof_handler->get_fe(fe_index).get_interpolation_matrix(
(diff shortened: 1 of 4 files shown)


source/numerics/solution_transfer.cc
```diff
@@ -511,36 +505,35 @@ SolutionTransfer<dim, VectorType, DoFHandlerType>::interpolate(
...
+                        interpolation_hp(active_fe_index, old_index);
+                      // The interpolation matrix might be empty when using
+                      // FE_Nothing.
+                      if (interpolation_matrix.empty())
+                        tmp.reinit(dofs_per_cell, false);
+                      else
+                        {
+                          tmp.reinit(dofs_per_cell, true);
+                          const unsigned int old_index =
+                            pointerstruct->second.active_fe_index;
```
tjhei: Isn't old_index already available above?

masterleinad: Yes, forgot to remove after restructuring. Fixed now.

tjhei:
Looks good! I am glad you found the bug. :-)

source/dofs/dof_accessor_get.cc
```diff
@@ -69,15 +69,23 @@ DoFCellAccessor<DoFHandlerType, lda>::get_interpolated_dof_values(
...
-          FullMatrix<double> interpolation(
-            this->dof_handler->get_fe(fe_index).dofs_per_cell,
-            this->get_fe().dofs_per_cell);
-          this->dof_handler->get_fe(fe_index).get_interpolation_matrix(
-            this->get_fe(), interpolation);
-          interpolation.vmult(interpolated_values, tmp);
+          const unsigned int dofs_per_cell = this->get_fe().dofs_per_cell;
+          if (dofs_per_cell == 0)
+            {
+              interpolated_values = 0;
```
bangerth: @masterleinad : I'm way late to this game, but isn't it the case that if you end in this branch, `interpolated_values` needs to have size zero anyway? If so, one could just skip the `if` and jump straight to the `else`. If you agree, I can write that patch.

masterleinad: No, in the test for this pull request the size of `interpolated_values` is 4 since the new element is a `FE_Q<2>(1)`.

bangerth: Oh, I see, it's the other way around. Thanks for confirming!

PR #8861, opened 2019-09-26
Fix Manifold get_new_point()

dougshidong:
The weights can be negative, we simply want to avoid dividing by 0.

### DIFF
--- a/source/grid/manifold.cc
+++ b/source/grid/manifold.cc
@@ -101,9 +101,15 @@ Manifold<dim, spacedim>::get_new_point(
         weight = w / (weights[permutation[i]] + w);
 
       if (std::abs(weight) > 1e-14)
-        p = get_intermediate_point(p,
-                                   surrounding_points[permutation[i]],
-                                   1.0 - weight);
+        {
+          p = get_intermediate_point(p,
+                                     surrounding_points[permutation[i]],
+                                     1.0 - weight);
+        }
+      else
+        {
+          p = surrounding_points[permutation[i]];
+        }
       w += weights[permutation[i]];
     }
 
--- a/source/grid/manifold.cc
+++ b/source/grid/manifold.cc
@@ -95,7 +95,7 @@ Manifold<dim, spacedim>::get_new_point(
   for (unsigned int i = 1; i < n_points; ++i)
     {
       double weight = 0.0;
-      if ((weights[permutation[i]] + w) < tol)
+      if (std::abs(weights[permutation[i]] + w) < tol)
         weight = 0.0;
       else
         weight = w / (weights[permutation[i]] + w);


dougshidong:
I found this by 
- deriving the Manifold class into my own
- implementing get_intermediate_point()
- attaching said manifold to a triangulation
- refining the grid

When refining the grid, the cell->center(const bool respect_manifold = true, const bool use_interpolation=true) will be called, which in turn will call get_new_point(). For a quad bi-linear element, the corners will be weighted at -0.25 and the edge midpoints will be weighted at 0.5.

The correct final position should be evaluated as seen below for a bilinear quad element.

  | x | y | weight | weighted x | weighted y
-- | -- | -- | -- | -- | --
Corner 1 | 10 | 0 | -0.25 | -2.5 | 0
Corner 2 | 0 | 10 | -0.25 | 0 | -2.5
Corner 3 | 1 | 0 | -0.25 | -0.25 | 0
Corner 4 | 0 | 1 | -0.25 | 0 | -0.25
Mid-edge 1 | 5.5 | 0 | 0.5 | 2.75 | 0
Mid-edge 2 | 0 | 5.5 | 0.5 | 0 | 2.75
Mid-edge 3 | 5 | 5 | 0.5 | 2.5 | 2.5
Mid-edge 4 | 0.5 | 0.5 | 0.5 | 0.25 | 0.25
  |   |   | Weighted position | 2.75 | 2.75

However, the original algorithm without a std::abs will not evaluate the negative weighted nodes.

dougshidong:
One more bug:
A performance-if ignores next point when current point weight = 0.

However, since the current point weight is 0, we should be replacing the current point with the next point, not skipping the next point.

Let me know if you want me to squash my 2 commits. They fix the same function, but fix 2 different bugs.

kronbichler:
It is also ok if the two commits stay separate.

source/grid/manifold.cc
```diff
@@ -101,9 +101,15 @@ Manifold<dim, spacedim>::get_new_point(
...
-                                   1.0 - weight);
+        {
+          p = get_intermediate_point(p,
+                                     surrounding_points[permutation[i]],
+                                     1.0 - weight);
+        }
+      else
+        {
+          p = surrounding_points[permutation[i]];
+        }
```
masterleinad: Actually, this optimization should not be necessary at all. In `get_intermediate_point` we do the same check with `tolerance=1.e-10`.

dougshidong: True, unless we derive the `Manifold` class and only implement `get_intermediate_point` as suggested by the documentation. Either way, this optimization doesn't change much. Let me know if you'd rather take it out.

masterleinad: It's fine.

bangerth:
@dougshidong -- this is a bug fix worth documenting, both for its own sake and to give you credit where credit is due. Want to write a follow-up pull request that adds an entry to `doc/news/changes/minor` in the same format as the other files in that directory?

PR #8863, opened 2019-09-26
Add missing test output

peterrum:


kronbichler:
We don't need the testers as they do not run this test anyway. @masterleinad you wrote this test initially, so I suggest you have a look before merging.

PR #8865, opened 2019-09-26
Fix failing CUDA tests

masterleinad:
Fixes most of the failing test in https://cdash.43-1.org/viewTest.php?onlydelta&buildid=1899.


peterrum:
Thanks!

PR #8871, opened 2019-09-27
Update step-64

peterrum:
Adapt step-64 to the changes introduced in #8838.


peterrum:
@masterleinad Done! Thanks for the review!

PR #8875, opened 2019-09-27
Bypass vector copy in MatrixFree::cell_loop with MPI

kronbichler:
The CUDA code with MPI would previously copy the vector in `CUDAWrappers::MatrixFree::cell_loop` for handling the ghost entries. However, if we have compatible partitioners, we can bypass that and simply let the vectors call `update_ghost_values()` and `compress()`.

### DIFF
--- a/doc/news/changes/minor/20190927MartinKronbichler
+++ b/doc/news/changes/minor/20190927MartinKronbichler
@@ -1,0 +1,6 @@
+Improved: CUDAWrappers::MatrixFree::cell_loop() can now detect whether the
+given vectors have the same parallel MPI partitioner as the matrix-free
+class, which circumvents vector copies and speeds up execution. This
+approach has also been applied to step-64.
+<br>
+(Martin Kronbichler, Peter Munch, 2019/09/27)
--- a/examples/step-64/step-64.cu
+++ b/examples/step-64/step-64.cu
@@ -236,6 +236,9 @@ namespace Step64
           const LinearAlgebra::distributed::Vector<double, MemorySpace::CUDA>
             &src) const;
 
+    void initialize_dof_vector(
+      LinearAlgebra::distributed::Vector<double, MemorySpace::CUDA> &vec) const;
+
   private:
     CUDAWrappers::MatrixFree<dim, double>       mf_data;
     LinearAlgebra::CUDAWrappers::Vector<double> coef;
@@ -301,6 +304,15 @@ namespace Step64
       coef.get_values());
     mf_data.cell_loop(helmholtz_operator, src, dst);
     mf_data.copy_constrained_values(src, dst);
+  }
+
+
+
+  template <int dim, int fe_degree>
+  void HelmholtzOperator<dim, fe_degree>::initialize_dof_vector(
+    LinearAlgebra::distributed::Vector<double, MemorySpace::CUDA> &vec) const
+  {
+    mf_data.initialize_dof_vector(vec);
   }
 
 
@@ -401,8 +413,8 @@ namespace Step64
     ghost_solution_host.reinit(locally_owned_dofs,
                                locally_relevant_dofs,
                                mpi_communicator);
-    solution_dev.reinit(locally_owned_dofs, mpi_communicator);
-    system_rhs_dev.reinit(locally_owned_dofs, mpi_communicator);
+    system_matrix_dev->initialize_dof_vector(solution_dev);
+    system_rhs_dev.reinit(solution_dev);
   }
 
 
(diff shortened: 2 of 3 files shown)


kronbichler:
FYI @peterrum

kronbichler:
When running with multiple GPUs the CG solver in step-64 does not converge any more - I have to investigate what happens, so putting the `do no merge` label on for now.

examples/step-64/step-64.cu
```diff
@@ -304,6 +307,15 @@ namespace Step64
   }
 
 
+
+  template <int dim, int fe_degree>
+  void HelmholtzOperator<dim, fe_degree>::initialize_dof_vector(
+    LinearAlgebra::distributed::Vector<double, MemorySpace::CUDA> &vec) const
+  {
+    mf_data.initialize_dof_vector(vec);
```
peterrum: As far as I see it, this function does not zero out the entries. Inside it calls `reinit(const std::shared_ptr<const Utilities::MPI::Partitioner> &partitioner)`, which does not  have the flag `const bool omit_zeroing_entries = false`. I do not know if this helps. Everything else looks good (from a standpoint that I do not have access to a compiler).

Rombur: This might be a problem. All the constrained entries need to be zero out.

kronbichler: The [function `reinit(Partitioner)`](https://github.com/dealii/dealii/blob/master/include/deal.II/lac/la_parallel_vector.templates.h#L511-L536) should indeed zero the vector, including the ghosts, as it calls `operator=(Number())`. I also checked that the `zero_out_ghosts()` function is doing the right thing.

kronbichler:
Note that I had `DEAL_II_MPI_WITH_CUDA_SUPPORT=ON` and it is unclear whether the RDMA works correctly (e.g. some synchronization missing in our code or some bug on the machine). When I disable this part and go through the host, my program runs fine (I tested on 2 and 5 GPUs). Thus, I'm removing the WIP flags; this should be ready for review.

kronbichler:
/rebuild

(This should only trigger on the CUDA machines and thus not the tester, but let's play safe given that are non-trivial changes.)

doc/news/changes/minor/20190927MartinKronbichler
```diff
@@ -0,0 +1,6 @@
+Improved: CUDAWrappers::MatrixFree::cell_loop() can now detect whether the
+given vectors have the same parallel MPI partitioners as the matrix-free
```
masterleinad:
```suggestion
given vectors have the same parallel MPI partitioner as the matrix-free
```

masterleinad:
I don't see more tests failing enabling CUDA-aware MPI support compared to not doing it. We need to investigate that separately.

kronbichler:
This likely depends on the MPI implementation. On the machine I'm testing on (Piz Daint), I have seen that to fix this we need
```
diff --git a/include/deal.II/base/partitioner.templates.h b/include/deal.II/base/partitioner.templates.h
index 167e01c45c..da662e10de 100644
--- a/include/deal.II/base/partitioner.templates.h
+++ b/include/deal.II/base/partitioner.templates.h
@@ -126,6 +126,7 @@ namespace Utilities
                   import_indices_plain_dev[i].first.get(),
                   locally_owned_array.data(),
                   chunk_size);
+              cudaDeviceSynchronize();
             }
           else
 #    endif
@@ -400,6 +401,11 @@ namespace Utilities
             ExcMessage("Index overflow: Maximum message size in MPI is 2GB. "
                        "The number of ghost entries times the size of 'Number' "
                        "exceeds this value. This is not supported."));
+#    if defined(DEAL_II_COMPILER_CUDA_AWARE) && \
+      defined(DEAL_II_MPI_WITH_CUDA_SUPPORT)
+          if (std::is_same<MemorySpaceType, MemorySpace::CUDA>::value)
+            cudaDeviceSynchronize();
+#    endif
           const int ierr =
             MPI_Isend(ghost_array_ptr,
                       ghost_targets_data[i].second * sizeof(Number),
```
I'm hesitating to add this now unless someone else confirms that this is necessary and it's not a "bug" in the MPI library. In my case, this is necessary because `MPI_Isend` does not wait for the CUDA kernels to finish, so the data to be exchanged might not be ready at the time MPI copies the data from one rank to the other. In any case, we have a solution to the problem.

masterleinad:
I agree with these changes. I also get different numbers of iterations in `step-64` with CUDA-aware MPI.

If you already have the changes in branch please turn them into a pull request. Otherwise, I can do that as well.

PR #8880, opened 2019-10-01
Synchronize CUDA device in MPI ghost exchange

kronbichler:
As discussed in #8875: we need to make sure the CUDA kernels are done before we try to send data over the network with CUDA-aware MPI.


masterleinad:
Thanks! I tested the pull request locally and it fixes a number of tests when running with CUDA-aware MPI. Hence, I am fine without another test.

PR #8881, opened 2019-10-01
Fix behavior for CUDA-aware MPI

masterleinad:
With this pull request, all the tests using CUDA-aware MPI are passing again.
- We must not use `std::memset` for device data. Instead use `.cudaMemset`.
- Ghost entries must not be actually inserted matching the behavior for the CPU code. There we assert that the values are the same, but this is difficult on the GPU so we omit it.

### DIFF
--- a/include/deal.II/base/partitioner.templates.h
+++ b/include/deal.II/base/partitioner.templates.h
@@ -513,18 +513,30 @@ namespace Utilities
                    "vector_operation argument was passed to "
                    "import_from_ghosted_array_start as is passed "
                    "to import_from_ghosted_array_finish."));
-#      ifdef DEAL_II_WITH_CXX17
-          if constexpr (std::is_trivial<Number>::value)
-#      else
-          if (std::is_trivial<Number>::value)
-#      endif
-            std::memset(ghost_array.data(),
-                        0,
-                        sizeof(Number) * ghost_array.size());
+
+#      if (defined(DEAL_II_COMPILER_CUDA_AWARE))
+          if (std::is_same<MemorySpaceType, MemorySpace::CUDA>::value)
+            {
+              cudaMemset(ghost_array.data(),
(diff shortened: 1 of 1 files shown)


include/deal.II/base/partitioner.templates.h
```diff
@@ -513,18 +513,30 @@ namespace Utilities
                    "vector_operation argument was passed to "
                    "import_from_ghosted_array_start as is passed "
                    "to import_from_ghosted_array_finish."));
+
+#      if (defined(DEAL_II_COMPILER_CUDA_AWARE))
```
kronbichler:
```suggestion
#      if defined(DEAL_II_COMPILER_CUDA_AWARE)
```
(I think we typically do not put braces here.)

include/deal.II/base/partitioner.templates.h
```diff
@@ -675,17 +687,9 @@ namespace Utilities
             {
               for (auto const &import_indices_plain : import_indices_plain_dev)
                 {
+                  // We can't easily assert here, so we just move the pointer
+                  // matching the host code.
```
Rombur: I don't get the comment. Compared to the CPU code, we removed an assert for the GPU code? You say we can't do the assert but we are not asserting anything right now.

kronbichler: In the insert case, nothing should happen with the ghosts as the owner is supposed to set them (and we do not keep track of individual set entries). So I think this is correct.

bangerth: I think what @Rombur was trying to point out is that the "We can't easily assert here" comment seems to refer to some other place in the code base where we *can* easily assert, but that that reference is lost on the casual reader. Maybe saying what one would *want* to assert would have helped, or an "In contrast to the same code in the xxx function". :-)

PR #8882, opened 2019-10-01
Overlap communication and computation in CUDA cell_loop [WIP]

peterrum:
This PR enables to the overlap of communication and computation in `CUDAWrapper::MatrixFree::cell_loop()`.

Sorry for the next [WIP] PR, however, I would like to share the changes already and get feedback already now. I have somewhat re/misused the coloring scheme ;)

ping @kronbichler @Rombur @masterleinad

### DIFF
--- a/include/deal.II/matrix_free/cuda_matrix_free.templates.h
+++ b/include/deal.II/matrix_free/cuda_matrix_free.templates.h
@@ -811,6 +811,7 @@ namespace CUDAWrappers
 
     this->parallelization_scheme = additional_data.parallelization_scheme;
     this->use_coloring           = additional_data.use_coloring;
+    this->use_ghost_coloring     = additional_data.use_ghost_coloring;
 
     // TODO: only free if we actually need arrays of different length
     free();
@@ -887,7 +888,41 @@ namespace CUDAWrappers
 
     if (begin != end)
       {
-        if (additional_data.use_coloring)
+        if(additional_data.use_ghost_coloring)
+          {
+            graph.resize(3, std::vector<CellFilter>());
+            
+            std::vector<bool> ghost_vertices(dof_handler.get_triangulation ().n_vertices (), false);
+            
+            for(const auto cell : dof_handler.get_triangulation ().active_cell_iterators ())
+              if(cell->is_ghost())
+                for(unsigned int i = 0; i < GeometryInfo<dim>::vertices_per_cell; i++)
+                  ghost_vertices[cell->vertex_index (i)] = true;
+              
+            std::vector<dealii::FilteredIterator<dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<dim, dim>, false>>>> inner_cells;
+              
+            for (auto cell = begin; cell != end; ++cell)
+            {
+              bool flag = false;
+                
+              for(unsigned int i = 0; i < GeometryInfo<dim>::vertices_per_cell; i++)
+                if(cell->vertex_index (i))
+                    flag = true;
+              
+              if(flag)
+                graph[1].emplace_back(cell);
+              else
+                inner_cells.emplace_back(cell);
+            }
+            
+            for(unsigned i = 0; i < inner_cells.size(); i++)
+                if(i < inner_cells.size() / 2)
+                  graph[0].emplace_back(inner_cells[i]);
+                else
+                  graph[2].emplace_back(inner_cells[i]);
+              
+          }
+        else if (additional_data.use_coloring)
           {
             const auto fun = [&](const CellFilter &filter) {
               return internal::get_conflict_indices<dim, Number>(filter,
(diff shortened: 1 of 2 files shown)


kronbichler:
This looks excellent from an execution point of view. (Obviously, we need to think about what to do with the existing functionality.)

include/deal.II/matrix_free/cuda_matrix_free.templates.h
```diff
@@ -887,7 +888,41 @@ namespace CUDAWrappers
...
+                  ghost_vertices[cell->vertex_index (i)] = true;
+              
+            std::vector<dealii::FilteredIterator<dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<dim, dim>, false>>>> inner_cells;
+              
+            for (auto cell = begin; cell != end; ++cell)
+            {
+              bool flag = false;
+                
+              for(unsigned int i = 0; i < GeometryInfo<dim>::vertices_per_cell; i++)
+                if(cell->vertex_index (i))
```
peterrum: This is wrong ;) I have to replace it with `if(ghost_vertices[cell->vertex_index (i)])`!

Rombur:
I like the idea. I quickly went over the code and my understanding is that `coloring` and `ghost_coloring` cannot be true at the same time. So we need to check that. Also I am not sure about the name `ghost_coloring` because the goal is very different of the other coloring.

Rombur:
@peterrum I'd like to get this merged before the release. What is left to do? I think that the `ghost_coloring` flag can go. If you use coloring nothing changes and if you don't (which is what you should do on modern architecture), then we should always uses the new path.

peterrum:
I would prefer if it would be like in the case of CPU version:
https://github.com/dealii/dealii/blob/2f9a789ebdb8b91b3bb4b7c85f9722c92eb2b306/include/deal.II/matrix_free/matrix_free.h#L227
so that there would be still the option to turn on and off *overlap computation and communication*.

To the state: I had some issues with the code which I posted here (which should be the same as here https://github.com/peterrum/dealii/blob/1278ba52648fc3e7949d376795133650e6aa47b2/include/deal.II/matrix_free/cuda_matrix_free.templates.h#L1086-L1116). I don't recall exactly what the problem was but the result was wrong under certain configurations. After some debugging, I limited the size of each color (see https://github.com/peterrum/dealii/blob/85bd6e8cacd59b949920a161dfad8588faabb9c5/include/deal.II/matrix_free/cuda_matrix_free.templates.h#L1107-L1146). This worked better, but I am not 100% sure if this fixed the problem. This is when my account at Piz Daint expired...

My guess is that the code ist 95% ready but there is a (minor) bug somewhere. Currently, I cannot work on this PR. However, if you have the time and resources to work on this PR, I would be happy to help and discuss aspects!

Rombur:
> but the result was wrong under certain configurations

Do you remember what kind of configurations? Did it depend on the number of GPU? The size of the problem?

peterrum:
I would say it was related to the number of refinements (for multiple GPUs - I think more than 2).

peterrum:
I have been using this code https://github.com/peterrum/deal-and-ceed-on-gpu/tree/fix.

Rombur:
Great thanks.

include/deal.II/matrix_free/cuda_matrix_free.templates.h
Rombur: Fixing this bug break step-64 :(

include/deal.II/matrix_free/cuda_matrix_free.templates.h
```diff
@@ -1032,17 +1067,44 @@ namespace CUDAWrappers
...
-            <<<grid_dim[i], block_dim[i]>>>(func,
-                                            get_data(i),
+            <<<grid_dim[0], block_dim[0]>>>(func,
+                                            get_data(0),
                                             src.get_values(),
                                             dst.get_values());
-        dst.compress(VectorOperation::add);
-        src.zero_out_ghosts();
+          src.update_ghost_values_finish();
+          internal::apply_kernel_shmem<dim, Number, Functor>
```
Rombur, line 1081: I think this kernel should be preceded and followed by `cudaDeviceSynchronize()` because of CUDA-aware MPI. How do you know that the data exchanged during the update_ghost/compress has been computed by the time the computation is done? If you don't use CUDA-aware MPI, we know that everything is using the same stream so it implicitly synchronize but we don't know when using CUDA-aware MPI. This also explains the problem you saw: you have a race condition and using more colors changes the time require to loop over the domain.

peterrum: @Rombur Thanks for the debugging! 

I am wondering which version of the code are you using? Are you using this version or the version on https://github.com/peterrum/dealii/tree/dealii-on-gpu? I would think 369a5b9ab5c203d651a1f435c3f01d152dcb3ef5 is working. However, what we observed at the workshop that the memcopy DtoH and HtoD was not overlapped, that is why we introduced the streams in 75988028d91e70095db24df61efebba395a2d664. I am afraid here we indeed need the synchronizations (before and after the second block - although I hoped that it is implicitly synchronized by the default stream).

Rombur: I found the bug. You can't see it if you use CUDA-aware MPI but funny enough you can see it in serial if you use regular MPI. In that case, we copy all the data to host, perform the communication, and then copy **all** the data back to the device. This means that if we change the data on the device, it is overwritten by the data from the host.

peterrum: Does this mean that the HtoD in `update_ghost_values_finish()` overrides values that have been computed in the first block? Do I understand it correctly? But how is that possible? Only the ghost values are copied around and that should be covered by the coloring (if it is correct), shouldn't it?

Rombur: No actually the problem is [here](https://github.com/dealii/dealii/blob/master/include/deal.II/lac/la_parallel_vector.templates.h#L961-L965) in `compress_finish`. The values computed in block 2 are not taken into account.

Rombur: So the problem is that when you don't use CUDA-aware MPI, the function `import_from_ghosted_array_finish` uses the data that has been copied to the host during the call to `compress_start`. If you update the data on the device after the call to `compress_start`, the host is not updated and you lose the new values. I need to double check but I think the solution is to copy the ghost data to the device, then do the call to `import_from_ghosted_array_finish`, and change `import_from_ghosted_array_finish` so that there is one CUDA path and one host path regardless of CUDA-aware MPI.

peterrum: This is the `if defined DEAL_II_COMPILER_CUDA_AWARE && !defined  DEAL_II_MPI_WITH_CUDA_SUPPORT` branch. In this configuration the whole vector is copied from and to the host. But that is not the case for `if defined DEAL_II_COMPILER_CUDA_AWARE && defined  DEAL_II_MPI_WITH_CUDA_SUPPORT`, isn't it?

peterrum: I think what I wanted to do [here](https://github.com/dealii/dealii/commit/369a5b9ab5c203d651a1f435c3f01d152dcb3ef5) was not write `#ifdef DEAL_II_COMPILER_CUDA_AWARE` but `#ifdef DEAL_II_MPI_WITH_CUDA_SUPPORT`. Sorry about this.

The whole purpose of the cude-aware MPI implementation is not to copy the whole vector.

Rombur: > This is the if defined DEAL_II_COMPILER_CUDA_AWARE && !defined DEAL_II_MPI_WITH_CUDA_SUPPORT branch. In this configuration the whole vector is copied from and to the host. But that is not the case for if defined DEAL_II_COMPILER_CUDA_AWARE && defined DEAL_II_MPI_WITH_CUDA_SUPPORT, isn't it?

That's right.

peterrum: > That's right.

I noticed your message to late...

PR #6873, opened 2018-07-03
Improve SolutionTransfer documentation

jppelteret:
The documentation for the SolutionTransfer class now includes a note on how to use with MPI partitioned grids. I've added a couple of tests that I used to teach myself how this is supposed to work.

PR #8886, opened 2019-10-03
Minor documentation fix in P::D::SolutionTransfer

rezarastak:
The original documentation used the method `refine_interpolate` which is not actually implemented in `parallel::distributed::SolutionTransfer`.

### DIFF
--- a/include/deal.II/distributed/solution_transfer.h
+++ b/include/deal.II/distributed/solution_transfer.h
@@ -113,12 +113,16 @@ namespace parallel
      *                     locally_relevant_dofs,
      *                     mpi_communicator);
      * old_solution = solution;
+     *
+     * // Initialize SolutionTransfer object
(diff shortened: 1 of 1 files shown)


bangerth:
I *think* that this looks good. @marcfehling ?

marcfehling:
These changes have been introduced in #6873. I think this section has been copied from the non-pd `SolutionTransfer` class, which is used for `p::s::Triangulation` objects.

This looks indeed good!

However, once you are there, I would suggest to move the section about `LinearAlgebra::distributed::Vector` below the section you just modified. It's better structured then in my opinion.

rezarastak:
> I would suggest to move the section about `LinearAlgebra::distributed::Vector` below the section you just modified. It's better structured then in my opinion.

I agree, it improves the flow of thought if we do this small reordering.

PR #8891, opened 2019-10-07
Remove std::bind from examples

masterleinad:
This pull request removes all the uses of `std::bind` from the examples. I am happy for suggestions for improving the documentation. Currently, I am removing all occurences of `std::bind` also in the documentation for the examples but would also be fine to mention it as long as we encourage using lambdas.

### DIFF
--- a/examples/step-13/step-13.cc
+++ b/examples/step-13/step-13.cc
@@ -788,7 +788,7 @@ namespace Step13
       // function</a> that wraps the function call including the individual
       // arguments with fixed values. The fixed values are passed into the
       // lambda function using the capture list
-      // ([...]). All the other arguments the wrapped function requires and are
+      // (`[...]`). All the other arguments the wrapped function requires and are
       // part of the outer function signature are specified as regular function
       // parameters in the lambda function.
       //
--- a/examples/step-13/step-13.cc
+++ b/examples/step-13/step-13.cc
@@ -737,22 +737,22 @@ namespace Step13
       Threads::Task<> rhs_task =
         Threads::new_task(&Solver<dim>::assemble_rhs, *this, linear_system.rhs);
 
-      WorkStream::run(dof_handler.begin_active(),
-                      dof_handler.end(),
-                      std::bind(&Solver<dim>::local_assemble_matrix,
-                                this,
-                                std::placeholders::_1,
-                                std::placeholders::_2,
-                                std::placeholders::_3),
-                      std::bind(&Solver<dim>::copy_local_to_global,
-                                this,
-                                std::placeholders::_1,
-                                std::ref(linear_system)),
-                      AssemblyScratchData(*fe, *quadrature),
-                      AssemblyCopyData());
+      WorkStream::run(
+        dof_handler.begin_active(),
+        dof_handler.end(),
+        [this](const typename DoFHandler<dim>::active_cell_iterator &cell,
+               AssemblyScratchData &scratch_data,
+               AssemblyCopyData &   copy_data) {
+          this->local_assemble_matrix(cell, scratch_data, copy_data);
+        },
+        [this, &linear_system](const AssemblyCopyData &copy_data) {
+          this->copy_local_to_global(copy_data, linear_system);
+        },
+        AssemblyScratchData(*fe, *quadrature),
+        AssemblyCopyData());
       linear_system.hanging_node_constraints.condense(linear_system.matrix);
 
-      // The syntax above using <code>std::bind</code> requires
+      // The syntax above using lambda functions requires
       // some explanation. There are multiple version of
       // WorkStream::run that expect different arguments. In step-9,
       // we used one version that took a pair of iterators, a pair of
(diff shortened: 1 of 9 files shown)


examples/step-13/step-13.cc
```diff
@@ -737,22 +737,22 @@ namespace Step13
...
-                      AssemblyCopyData());
+      WorkStream::run(
+        dof_handler.begin_active(),
+        dof_handler.end(),
+        [this](const typename DoFHandler<dim>::active_cell_iterator &cell,
+               AssemblyScratchData &scratch_data,
+               AssemblyCopyData &   copy_data) {
+          this->local_assemble_matrix(cell, scratch_data, copy_data);
+        },
+        [this, &linear_system](const AssemblyCopyData &copy_data) {
```
bangerth: Since this is in a tutorial program where simplicity is a concern, what would you think about omitting the explicit captures here?

examples/step-13/step-13.cc
```diff
@@ -783,58 +783,15 @@ namespace Step13
...
-      // to <code>*this</code>, a reference to a copy object, and a
-      // reference to a linear system, and binding the first and third
-      // of these, leaving something that has only one open argument
-      // that can then be filled by WorkStream::run().
+      // typical way to generate such function objects is using a
+      // <a href="http://en.wikipedia.org/wiki/Anonymous_function">lambda
+      // function</a> that wraps the function call including the individual
+      // arguments with fixed values. The fixed values are passed into the
+      // lambda function using the capture list
+      // ([...]). All the other arguments the wrapped function requires and are
```
bangerth:
```suggestion
      // (`[...]`). All the other arguments the wrapped function requires and are
```

examples/step-14/step-14.cc
```diff
@@ -522,19 +522,19 @@ namespace Step14
...
-                      AssemblyCopyData());
+      WorkStream::run(
+        dof_handler.begin_active(),
+        dof_handler.end(),
+        [this](const typename DoFHandler<dim>::active_cell_iterator &cell,
+               AssemblyScratchData &scratch_data,
+               AssemblyCopyData &   copy_data) {
+          this->local_assemble_matrix(cell, scratch_data, copy_data);
+        },
+        [this, &linear_system](const AssemblyCopyData &copy_data) {
```
bangerth: same here

examples/step-14/step-14.cc
```diff
@@ -2193,13 +2193,14 @@ namespace Step14
...
-        std::bind(&WeightedResidual<dim>::estimate_on_one_cell,
-                  this,
-                  std::placeholders::_1,
-                  std::placeholders::_2,
-                  std::placeholders::_3,
-                  std::ref(error_indicators),
-                  std::ref(face_integrals)),
+        [this,
+         &error_indicators,
+         &face_integrals](const active_cell_iterator & cell,
```
bangerth: and here

examples/step-32/step-32.cc
```diff
@@ -2150,16 +2148,14 @@ namespace Step32
...
-      std::bind(
-        &BoussinesqFlowProblem<dim>::copy_local_to_global_stokes_preconditioner,
-        this,
-        std::placeholders::_1),
+      [this](const typename DoFHandler<dim>::active_cell_iterator &cell,
+             Assembly::Scratch::StokesPreconditioner<dim> &        scratch,
+             Assembly::CopyData::StokesPreconditioner<dim> &       data) {
+        this->local_assemble_stokes_preconditioner(cell, scratch, data);
+      },
+      [this](const Assembly::CopyData::StokesPreconditioner<dim> &data) {
```
bangerth: and here

examples/step-32/step-32.cc
```diff
@@ -2721,18 +2716,20 @@ namespace Step32
...
-                maximal_velocity,
-                global_entropy_variation,
-                std::placeholders::_1,
-                std::placeholders::_2,
-                std::placeholders::_3),
-      std::bind(
-        &BoussinesqFlowProblem<dim>::copy_local_to_global_temperature_rhs,
-        this,
-        std::placeholders::_1),
+      [this, global_T_range, maximal_velocity, global_entropy_variation](
```
bangerth: and in particular here

examples/step-13/step-13.cc
tjhei: I don't feel that listing the captures is a big issue, but I find it a lot more readable if you move the definition of the lambda above the function call and give it a descriptive name.

masterleinad: My reasoning for the explicit capture list would be that we make it as clear as possible what the lambda is really doing (and which variables it is using). 
I don't have a strong opinion here, though. If someone strongly prefers one variant I am happy to change all of them in the examples.

I think that we should discuss both variants here in any way.

masterleinad: Also see http://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p0806r2.html and https://en.cppreference.com/w/cpp/language/lambda:
```
 The implicit capture of *this when the capture default is = is deprecated. (since C++20).
```

examples/step-13/step-13.cc
```diff
@@ -783,58 +787,19 @@ namespace Step13
...
-      // result in calling <code>f(2,1)</code>.
+      // typical way to generate such function objects is using a
+      // <a href="http://en.wikipedia.org/wiki/Anonymous_function">lambda
+      // function</a> that wraps the function call including the individual
+      // arguments with fixed values. All the arguments that are part of the
+      // outer function signature are specified as regular function arguments in
+      // the lambda function. The fixed values are passed into the lambda
+      // function using the capture list (`[...]`). It is possible to use a
+      // capture default or to list all the variables that are to be bound to
+      // the lambda explicitly. For the sake of clarity we decided to to omit
```
bangerth:
```suggestion
      // the lambda explicitly. For the sake of clarity we decided to omit
```

examples/step-13/step-13.cc
```diff
@@ -783,58 +787,19 @@ namespace Step13
...
+      // the capture default here, but that capture list could equally well be
+      // `[&]` meaning that all used variables are copied into the lambda by
```
bangerth:
```suggestion
      // `[&]`, meaning that all used variables are copied into the lambda by
```

examples/step-14/step-14.cc
```diff
@@ -2188,19 +2192,24 @@ namespace Step14
...
+      auto worker = [this,
+                     &error_indicators,
+                     &face_integrals](const active_cell_iterator & cell,
+                                      WeightedResidualScratchData &scratch_data,
+                                      WeightedResidualCopyData &   copy_data) {
+        this->estimate_on_one_cell(
+          cell, scratch_data, copy_data, error_indicators, face_integrals);
+      };
+
+      auto copier = std::function<void(const WeightedResidualCopyData &)>();
```
bangerth: Let's make it extra clear :-)
```suggestion
      auto do_nothing_copier = std::function<void(const WeightedResidualCopyData &)>();
```

examples/step-14/step-14.cc
```diff
@@ -2188,19 +2192,24 @@ namespace Step14
...
-        std::bind(&WeightedResidual<dim>::estimate_on_one_cell,
-                  this,
-                  std::placeholders::_1,
-                  std::placeholders::_2,
-                  std::placeholders::_3,
-                  std::ref(error_indicators),
-                  std::ref(face_integrals)),
-        std::function<void(const WeightedResidualCopyData &)>(),
+        worker,
+        copier,
```
bangerth:
```suggestion
        do_nothing_copier,
```

examples/step-32/step-32.cc
```diff
@@ -2716,26 +2716,31 @@ namespace Step32
...
+          Assembly::CopyData::TemperatureRHS<dim> &             data) {
+        this->local_assemble_temperature_rhs(global_T_range,
+                                             maximal_velocity,
+                                             global_entropy_variation,
+                                             cell,
+                                             scratch,
+                                             data);
+      };
+
+    auto copier = [=](const Assembly::CopyData::TemperatureRHS<dim> &data) {
```
bangerth: You really just copy `this` here, and you've explicitly listed it in other places as well, so let's go with that.
```suggestion
    auto copier = [this](const Assembly::CopyData::TemperatureRHS<dim> &data) {
```

examples/step-51/step-51.cc
```diff
@@ -1027,16 +1027,17 @@ namespace Step51
...
-                      0U);
+      WorkStream::run(
+        dof_handler_u_post.begin_active(),
+        dof_handler_u_post.end(),
+        [this](const typename DoFHandler<dim>::active_cell_iterator &cell,
+               PostProcessScratchData &                              scratch,
+               unsigned int &                                        data) {
+          this->postprocess_one_cell(cell, scratch, data);
+        },
+        [](const unsigned int &) {},
```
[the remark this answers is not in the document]
bangerth: Same issue here as before: you really do want the `do_nothing_copier` here.

masterleinad:
@bangerth Would you mind approving this "officially" since you already gave your consent in the comments? Pull request authors can't approve their own pull requests.

PR #8893, opened 2019-10-08
Add VectorizedArray::size() member function

kronbichler:
This adds a `size()` function to `VectorizedArray`. I find this more convenient than using the `VectorizedArray::n_array_elements` member variable when writing loops with STL in mind.

One thing to discuss is whether we want to add more interfaces familiar from `std::array`, such as `begin()/end()` that would also allow to use range-based for loops for example.

### DIFF
--- a/doc/news/changes/minor/20191008MartinKronbichler
+++ b/doc/news/changes/minor/20191008MartinKronbichler
@@ -1,0 +1,5 @@
+New: The class VectorizedArray now provides a member function
+VectorizedArray::size() to return the number of array elements, in analogy to
+the STL containers.
+<br>
+(Martin Kronbichler, 2019/10/08)
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -217,6 +217,16 @@ public:
   }
 
   /**
+   * Return the number of elements in the array stored in the variable
+   * n_array_elements.
+   */
+  static constexpr unsigned int
+  size()
+  {
+    return n_array_elements;
+  }
+
+  /**
    * Access operator (only valid with component 0 in the base class without
    * specialization).
    */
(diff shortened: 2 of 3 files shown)


kronbichler:
> Looks good! Why didn't you call it right away so? Could we deprecate the other one?

Before C++11 even static member functions were not `constexpr`, so we could not propagate the compile-time knowledge (if I remember correctly). In any case, we did not envision that the array would be used to widely, so we did not think about a possible implementation with STL-like interface.

Regarding deprecation: The array is used quite a lot, so this would create some major noise which me must carefully take into account. However, we could make `size()` a `static constexpr` function which would at least give the potential to replace it. I do not feel too strongly to provide the same information twice, so that is not a big topic.

> Regarding iterator: would be nice

Yes, we do not want to copy the body if we do all iterators. We would need to introduce a base class for making the code duplication acceptable. That would require a bigger action. I could do it if we feel strongly for it, but I don't think that's high priority.

masterleinad:
I would prefer to make the function `static` as well. Otherwise, this looks good to me.
I don't feel strongly about deprecating `n_array_elements`, though.

peterrum:
Regrading `n_array_elements`: I tend to forget the name and have to look it up always in Doxygen...

tamiko:
On Tue, Oct  8, 2019, at 02:21 CDT, Martin Kronbichler <notifications@github.com> wrote:

> This adds a `size()` function to `VectorizedArray`. I find this more
> convenient than using the `VectorizedArray::n_array_elements` member
> variable when writing loops with STL in mind.
>
> One thing to discuss is whether we want to add more interfaces
> familiar from `std::array`, such as `begin()/end()` that would also
> allow to use range-based for loops for example.

Support for range-based for loops is the one feature I am missing the
most :-)

kronbichler:
I made the function static and added that part to the test as well.

PR #8894, opened 2019-10-08
New functionality to compute maximum aspect ratio of a mesh

nfehn:
This pull request adds new functionality to compute the maximum aspect ratio of a mesh. A first function fills a vector with one aspect ratio value per cell. A second function computes the maximum over all cells. The aspect ratio definition used here relies on the ratio of the maximum to minimum singular value of the Jacobian. Available deal.II functionality is used if possible and a new test ensures a correct behavior for hyper_rectangle examples.

### DIFF
--- a/doc/news/changes/minor/20191009NiklasFehn
+++ b/doc/news/changes/minor/20191009NiklasFehn
@@ -1,0 +1,4 @@
+New: The function GridTools::compute_maximum_aspect_ratio() computes the
+maximum aspect ratio of a mesh with arbitrarily deformed elements.
+<br>
+(Niklas Fehn, 2019/10/08)
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -231,6 +231,55 @@ namespace GridTools
   cell_measure(const T &, ...);
 
   /**
+   * Computes an aspect ratio measure for all locally-owned active cells and
+   * fills a vector with one entry per cell, given a @p triangulation and
+   * @p mapping. The size of the vector that is returned equals the number of
+   * active cells. The vector contains zero for non locally-owned cells. The
+   * aspect ratio of a cell is defined as the ratio of the maximum to minimum
+   * singular value of the Jacobian, taking the maximum over all quadrature
+   * points of a quadrature rule specified via @p quadrature. For example, for
+   * the special case of rectangular elements in 2d with dimensions $a$ and $b$
+   * ($a \geq b$), this function returns the usual aspect ratio definition
+   * $a/b$. The above definition using singular values is a generalization to
+   * arbitrarily deformed elements. This function is intended to be used for
+   * $d=2,3$ space dimensions, but it can also be used for $d=1$ returning a
+   * value of 1.
+   *
+   * @note Inverted elements do not throw an exception. Instead, a value of inf
+   * is written into the vector in case of inverted elements.
+   *
+   * @note Make sure to use enough quadrature points for a precise calculation
+   * of the aspect ratio in case of deformed elements.
+   *
+   * @note In parallel computations the return value will have the length
+   * n_active_cells but the aspect ratio is only computed for the cells that
+   * are locally owned and placed at index CellAccessor::active_cell_index(),
+   * respectively. All other values are set to 0.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  template <int dim>
+  Vector<double>
+  compute_aspect_ratio_of_cells(const Triangulation<dim> &triangulation,
+                                const Mapping<dim> &      mapping,
+                                const Quadrature<dim> &   quadrature);
+
+  /**
+   * Computes the maximum aspect ratio by taking the maximum over all cells.
+   *
+   * @note When running in parallel with a Triangulation that supports MPI,
+   * this is a collective call and the return value is the maximum over all
+   * processors.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  template <int dim>
+  double
+  compute_maximum_aspect_ratio(const Triangulation<dim> &triangulation,
+                               const Mapping<dim> &      mapping,
+                               const Quadrature<dim> &   quadrature);
+
+  /**
    * Compute the smallest box containing the entire triangulation.
    *
    * If the input triangulation is a `parallel::distributed::Triangulation`,
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -52,6 +52,7 @@
 #include <deal.II/lac/vector_memory.h>
 
 #include <deal.II/numerics/matrix_tools.h>
+#include <deal.II/numerics/vector_tools.h>
 
 #include <boost/random/mersenne_twister.hpp>
 #include <boost/random/uniform_real_distribution.hpp>
@@ -413,6 +414,79 @@ namespace GridTools
     return (t34 + t64 + t95 + t125 + t156 + t181 + t207 + t228) / 12.;
   }
 
+
+  template <int dim>
+  Vector<double>
+  compute_aspect_ratio_of_cells(const Triangulation<dim> &triangulation,
+                                const Mapping<dim> &      mapping,
+                                const Quadrature<dim> &   quadrature)
+  {
+    FE_Nothing<dim> fe;
+    FEValues<dim>   fe_values(mapping, fe, quadrature, update_jacobians);
+
+    Vector<double> aspect_ratio_vector(triangulation.n_active_cells());
+
+    // loop over cells of processor
+    for (const auto &cell : triangulation.active_cell_iterators())
+      {
+        if (cell->is_locally_owned())
+          {
+            double aspect_ratio_cell = 0.0;
+
+            fe_values.reinit(cell);
+
+            // loop over quadrature points
+            for (unsigned int q = 0; q < quadrature.size(); ++q)
+              {
+                Tensor<2, dim, double> jacobian =
+                  Tensor<2, dim, double>(fe_values.jacobian(q));
+
+                LAPACKFullMatrix<double> J = LAPACKFullMatrix<double>(dim);
+                for (unsigned int i = 0; i < dim; i++)
+                  for (unsigned int j = 0; j < dim; j++)
+                    J(i, j) = jacobian[i][j];
+
+                // We intentionally do not want to throw an exception in case of
+                // inverted elements since this is not the task of this
+                // function. Instead, inf is written into the vector in case of
+                // inverted elements.
+                if (determinant(jacobian) <= 0)
+                  {
+                    aspect_ratio_cell = std::numeric_limits<double>::infinity();
+                  }
+                else
+                  {
+                    J.compute_svd();
+
+                    double const max_sv = J.singular_value(0);
+                    double const min_sv = J.singular_value(dim - 1);
+                    double const ar     = max_sv / min_sv;
+
+                    aspect_ratio_cell = std::max(aspect_ratio_cell, ar);
+                  }
+              }
+
+            // fill vector
+            aspect_ratio_vector(cell->active_cell_index()) = aspect_ratio_cell;
+          }
+      }
+
+    return aspect_ratio_vector;
+  }
+
+  template <int dim>
+  double
+  compute_maximum_aspect_ratio(const Triangulation<dim> &triangulation,
+                               const Mapping<dim> &      mapping,
+                               const Quadrature<dim> &   quadrature)
+  {
+    Vector<double> aspect_ratio_vector =
+      compute_aspect_ratio_of_cells(triangulation, mapping, quadrature);
+
+    return VectorTools::compute_global_error(triangulation,
+                                             aspect_ratio_vector,
+                                             VectorTools::Linfty_norm);
+  }
 
 
   template <int dim, int spacedim>
--- a/tests/grid/grid_tools_aspect_ratio.cc
+++ b/tests/grid/grid_tools_aspect_ratio.cc
@@ -1,0 +1,234 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Short test to validate GridTools::compute_maximum_aspect_ratio()
+
+#include <deal.II/base/exceptions.h>
+#include <deal.II/base/mpi.h>
+#include <deal.II/base/partitioner.h>
+#include <deal.II/base/quadrature_lib.h>
+#include <deal.II/base/utilities.h>
+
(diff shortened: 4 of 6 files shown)


source/grid/grid_tools.cc
```diff
@@ -414,6 +415,79 @@ namespace GridTools
...
+    FE_Nothing<dim> fe;
+    FEValues<dim>   fe_values(mapping, fe, quadrature, update_jacobians);
+
+    typedef typename Triangulation<dim>::active_cell_iterator cell_iter;
+    cell_iter cell = triangulation.begin_active(), endc = triangulation.end();
+
+    Vector<double> aspect_ratio_vector(triangulation.n_active_cells());
+
+    // loop over cells of processor
+    for (; cell != endc; ++cell)
```
kronbichler: Can you change this to a range-based for loop:
```suggestion
    for (const auto &cell : triangulation.active_cell_iterators())
```
This will allow you to remove lines 427-428.

tests/grid/grid_tools_aspect_ratio.cc
```diff
@@ -0,0 +1,188 @@
+#include <deal.II/base/exceptions.h>
```
kronbichler: Can you go to another test file and copy the standard deal.II header (copyright notice etc.) in here?

source/grid/grid_tools.cc
nfehn: done in https://github.com/dealii/dealii/pull/8894/commits/cd86f95e2cbfe49fee59b9e792b12e9b3921e5a6

tests/grid/grid_tools_aspect_ratio.cc
nfehn: done in https://github.com/dealii/dealii/pull/8894/commits/64e29053cb6ebb238940ffda402e6f043de45d51

doc/news/changes/minor/20191009NiklasFehn
```diff
@@ -0,0 +1,4 @@
+New: Write new function in GridTools to compute the maximum aspect ratio
```
tjhei:
```suggestion
New: The function GridTools::calculate_aspect_ratio_of_cells() computes the maximum aspect ratio 
```

include/deal.II/grid/grid_tools.h
```diff
@@ -230,6 +230,43 @@ namespace GridTools
...
+   * @note Inverted elements do not throw an exception. Instead, a value of inf
+   * is written into the vector in case of inverted elements.
+   *
+   * @note Make sure to use enough quadrature points for a precise calculation
+   * of the aspect ratio in case of deformed elements.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  template <int dim>
+  Vector<double>
```
tjhei: can you clarify the size of the return value? I assume this is of length n_active_cells()? What happens with ghost or artificial cells?

include/deal.II/grid/grid_tools.h
```diff
@@ -230,6 +230,43 @@ namespace GridTools
...
+                                  const Quadrature<dim> &   quadrature);
+
+  /**
+   * Computes the maximum aspect ratio by taking the maximum over all cells.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  template <int dim>
+  double
+  calculate_maximum_aspect_ratio(const Triangulation<dim> &triangulation,
```
tjhei: we call most of these functions compute_*, maybe this is worth changing to be a bit more consistent?

include/deal.II/grid/grid_tools.h
```diff
@@ -230,6 +230,43 @@ namespace GridTools
...
+   * is written into the vector in case of inverted elements.
+   *
+   * @note Make sure to use enough quadrature points for a precise calculation
+   * of the aspect ratio in case of deformed elements.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  template <int dim>
+  Vector<double>
+  calculate_aspect_ratio_of_cells(const Triangulation<dim> &triangulation,
```
tjhei:
```suggestion
  compute_cell_aspect_ratios(const Triangulation<dim> &triangulation,
```

source/grid/grid_tools.cc
```diff
@@ -414,6 +415,76 @@ namespace GridTools
...
+                                  const Quadrature<dim> &   quadrature)
+  {
+    FE_Nothing<dim> fe;
+    FEValues<dim>   fe_values(mapping, fe, quadrature, update_jacobians);
+
+    Vector<double> aspect_ratio_vector(triangulation.n_active_cells());
+
+    // loop over cells of processor
+    for (const auto &cell : triangulation.active_cell_iterators())
+      {
```
tjhei: I think you might want to skip artificial cells (does this even work right now?).

nfehn: Sorry, I need some background. Are you happy with the previous (non range-based) version, or is it the same for both versions?

doc/news/changes/minor/20191009NiklasFehn
nfehn: I can change this. I am somehow hesitating because the change of name as requested below will introduce inconsistencies if not remembering all occurrences. Should I change nevertheless?

tjhei: It is useful to have the name of the new function in the changelog and "Write new function ..." sounds a bit weird. Of course they will need updating before we merge the PR to make sure they are consistent. You are welcome to wait to address this comment until everything else is figured out.

nfehn: changed in https://github.com/dealii/dealii/pull/8894/commits/957deb80a0ed01b949663db78708c52e2f643f33

include/deal.II/grid/grid_tools.h
nfehn: changed in https://github.com/dealii/dealii/pull/8894/commits/a55ae94a1579df8590f06e8c7897b0b2401afa61

source/grid/grid_tools.cc
nfehn: I don't know what an ``artificial`` cell is. I agree with ``locally_owned``. What I found in the deal.II docu is that ``locally_owned`` and ``!artificial`` is clearly something different, so I am not sure if I really got your point (assume that this is just an example? What info does a processor have regarding ghost cells vs. artificial cells? Do you want to imply that the computation would be correct for a ghost cell?).

nfehn: changed in https://github.com/dealii/dealii/pull/8894/commits/b9ae00282d0a425fbcd7b8c6dc37af4bb40950d1

include/deal.II/grid/grid_tools.h
nfehn: I don't see why the size of the vector should be different than what you have written. Can you please share your concerns? Do you want to assert the size?

include/deal.II/grid/grid_tools.h
```diff
@@ -230,6 +230,43 @@ namespace GridTools
   double
   cell_measure(const T &, ...);
 
+  /**
+   * Computes an aspect ratio measure for all cells and fills a vector with one
```
peterrum: Just write: `Computes an aspect ratio measure for all locally owned active cells...`. That should please @tjhei!?

include/deal.II/grid/grid_tools.h
peterrum: It seems (active) ghost and artificial cells have an aspect ratio of 0.0. Is that what we want?

source/grid/grid_tools.cc
masterleinad: The glossary (https://www.dealii.org/current/doxygen/deal.II/DEALGlossary.html) says:
> Artificial cells
    If a mesh is distributed across multiple MPI processes using the parallel::distributed::Triangulation class, each processor stores only the cells it owns, one layer of adjacent cells that are owned by other processors (called ghost cells), all coarse level cells, and all cells that are necessary to maintain the invariant that adjacent cells must differ by at most one refinement level. The cells stored on each process that are not owned by this process and that are not ghost cells are called "artificial cells", and for these cells the predicate cell->is_artificial() returns true. Artificial cells are guaranteed to exist in the globally distributed mesh but they may be further refined on other processors. See the Distributed Computing paper for more information.
    The concept of artificial cells has no meaning for triangulations that store the entire mesh on each processor, i.e. the Triangulation class. 

Artifical cells don't necessarily carry the correct, i.e. globall consistent, geometric informations while this is true for locally owned and ghosted cells.

include/deal.II/grid/grid_tools.h
masterleinad: Sounds as `inf` to me. We should document that in any way.

tjhei: if you use inf for ghost/artificial then taking the max over the vector is no longer a sensible operation. :-)
So either denote 0.0 for cells where no information is available (ghost, artificial), or do not include them in the return value.

@nfehn, my comment here refers to the fact that you have to document the size of the return value. You have two options:
A) size is n_active_cells() and you need to decide what value is set for ghost and artificial cells.
B) size is the number of locally owned cells. (we currently don't even have a function that returns this number).

source/grid/grid_tools.cc
tjhei: > Do you want to imply that the computation would be correct for a ghost cell?).

Likely, unless one processor decides to move vertices of only their own cells around.

include/deal.II/grid/grid_tools.h
tjhei: yes, that is better.

include/deal.II/grid/grid_tools.h
peterrum: @tjhei do you mean `parallel::TriangulationBase::n_locally_owned_active_cells`?  We could move that to `Triangulation` and override it in `parallel::TriangulationBase`...

nfehn: What meaning does ``locally_owned`` have for a non-parallel ``Triangulation``? But I agree that some basic deal.II functionality should be extended. From an OO design perspective, it should not be necessary to discuss ghost and artificial cells in this context at all.

nfehn: Improved documentation regarding vector size in https://github.com/dealii/dealii/pull/8894/commits/28a740482db88d0cd605140b889e860f8e001730 .

I intentionally do not mention ghost or artificial cells (assume someone introduces a third type of non locally-owned cells and the inconsistencies this would introduce, see Open-Closed Principle).

source/grid/grid_tools.cc
kronbichler: I would consider wrong information for ghost cells a bug - DG methods taking derivatives on faces critically depend on the ghost cells to be consistent. (Hyperbolic terms should be fine if all vertices to the locally owned cells are consistent across processors because the normal vector and integration factors do only depend on the vertices on the face; whereas gradients need *all* vertices of a ghost cell.)

Nonetheless, I agree with @tjhei that we should only use the own cells - this is assumed by `VectorTools::compute_global_error` which is used here. Also, we should initialize the vector with `0` so artificial cells always get that value. The typical use case is to compute a max over all owned cells (where only owned cells are used) or to pass that vector to `DataOut::add_data_vector` where one also ignores non-owned cells, so the state is correct. And an aspect ratio of 0 signals that nothing has happened as the definition cannot be less than 1.

include/deal.II/grid/grid_tools.h
kronbichler: To be consistent with output writers, we need to have the size of the vector `n_active_cells()` rather than the number of locally owned active cells and skip non-owned cells.

source/grid/grid_tools.cc
nfehn: Initialization with 0 is done by the constructor, isn't it?

kronbichler: Yes.

tests/grid/grid_tools_aspect_ratio.cc
```diff
@@ -0,0 +1,205 @@
...
+template <int dim>
+double
+compute_aspect_ratio_hyper_rectangle(
+  Point<dim> const &               left,
+  Point<dim> const &               right,
+  std::vector<unsigned int> const &refinements,
+  unsigned int                     degree     = 1,
+  unsigned int                     n_q_points = 2,
+  bool                             deform     = false,
+  double                           factor     = 1.0)
```
kronbichler: Can you add a short comment for this function that discusses the purpose of the parameters, in particular `refinements`, `deform` and `factor`, so in case some else stumbles across this test immediately sees the purpose? While there, you might also want to output the case that is currently done (i.e., the value of these three parameters plus maybe the dimension that you can then remove from the main function) at the top of this function. Imagine a case where one test fails - one wants to immediately spot in the output file what configuration has failed, rather than counting the calls in the main function and compare to the output file.

Also, I think you can skip the `degree` parameter` because you only have a flat manifold where `MappingQGeneric` degenerates to a bi-/trilinear mapping in any case. The number of quadrature points is more interesting, and I would suggest that add two more calls to this function where you pick more than 2 points, say for the 2D cases with moderate deformed vertex (where I would expect a different result for the aspect ratio) and one without deformation (where I would expect the same result as previously.

nfehn: done in https://github.com/dealii/dealii/pull/8894/commits/309e90b0631f3c3a4a40bf7380addaab683dd01a

include/deal.II/grid/grid_tools.h
tjhei: Please add a sentence like
"@note In parallel computations the return value will have the length n_active_cells but the aspect ratio is only computed for the cells that are locally owned and placed at index CellAccessor::active_cell_index(), respectively. All other values are set to 0."

include/deal.II/grid/grid_tools.h
```diff
@@ -230,6 +230,46 @@ namespace GridTools
...
+  template <int dim>
+  Vector<double>
+  compute_aspect_ratio_of_cells(const Triangulation<dim> &triangulation,
+                                const Mapping<dim> &      mapping,
+                                const Quadrature<dim> &   quadrature);
+
+  /**
+   * Computes the maximum aspect ratio by taking the maximum over all cells.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
```
tjhei: please add something like
```
@note When running in parallel with a Triangulation that supports MPI, this is a collective call and the return value is the maximum over all processors.
```

nfehn: done in https://github.com/dealii/dealii/pull/8894/commits/da7e69142eb66d977e7c02e0959c31b0163d70c1

include/deal.II/grid/grid_tools.h
nfehn: done in https://github.com/dealii/dealii/pull/8894/commits/da7e69142eb66d977e7c02e0959c31b0163d70c1

tests/grid/grid_tools_aspect_ratio.cc
```diff
@@ -0,0 +1,234 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2001 - 2018 by the deal.II authors
```
peterrum:
```suggestion
// Copyright (C) 2019 by the deal.II authors
```

nfehn: changed in https://github.com/dealii/dealii/pull/8894/commits/c860e2131a5647fd37eb07272b1b88df0a75a1df

It seems we are converged now?

tjhei:
Looks good! @nfehn , would you mind squashing your commits down to 1 or 2?

PR #8895, opened 2019-10-08
Fix a bug in merge triangulations()

starki0815:
The following example produces a mesh for which there are no cell faces with `face->at_boundary() == true`, though there should be 10 boundary faces:
```cpp
#include <iostream>

#include <deal.II/grid/grid_tools.h>

using namespace std;
using namespace dealii;

int main()
{
	const unsigned int spacedim = 3;

	Triangulation<spacedim> tria_0, tria_1, tria_2;

	GridGenerator::hyper_cube(tria_0);
	GridGenerator::hyper_cube(tria_1);
	Tensor<1, spacedim> shift;
	shift[0] = 1.0;
	GridTools::shift(shift, tria_1);

	tria_0.set_all_manifold_ids(0);
	tria_1.set_all_manifold_ids(0);
	GridGenerator::merge_triangulations(tria_0, tria_1, tria_2, 1e-12, true);

	unsigned int boundary_face_count = 0;
	for(const auto& cell : tria_2.active_cell_iterators())
		for(unsigned int f = 0; f < GeometryInfo<spacedim>::faces_per_cell; ++f)
			if(cell->face(f)->at_boundary())
				++boundary_face_count;

	if(boundary_face_count != 10)
		cout << "Found " << boundary_face_count << " boundary faces. However, there should be 10 of them!" << endl;
	else
		cout << "OK!" << endl;
}
```
The reason is that `create_triangulation()` assigns internal boundary id's to the actual boundary. This patch should fix this.


source/grid/tria.cc
```diff
@@ -3166,8 +3166,11 @@ namespace internal
                             ExcInconsistentLineInfoOfLine(line_vertices.first,
                                                           line_vertices.second,
                                                           "boundary ids"));
-
-                line->set_boundary_id_internal(subcell_line.boundary_id);
+                // do not use the boundary id provided in subcell_line
+                // if it is an internal one
+                if (subcell_line.boundary_id !=
+                    numbers::internal_face_boundary_id)
+                  line->set_boundary_id_internal(subcell_line.boundary_id);
```
masterleinad: Let's use `TriaAccessor::set_boundary_id` then.
```suggestion
                  line->set_boundary_id(subcell_line.boundary_id);
```

source/grid/tria.cc
```diff
@@ -3317,8 +3320,11 @@ namespace internal
                                                           line[2]->index(),
                                                           line[3]->index(),
                                                           "boundary ids"));
-
-                quad->set_boundary_id_internal(subcell_quad.boundary_id);
+                // do not use the boundary id provided in subcell_quad
+                // if it is an invalid or internal one
+                if (subcell_quad.boundary_id !=
+                    numbers::internal_face_boundary_id)
+                  quad->set_boundary_id_internal(subcell_quad.boundary_id);
```
masterleinad: Let's use `TriaAccessor::set_boundary_id` then.
```suggestion
                  quad->set_boundary_id(subcell_quad.boundary_id);
```

source/grid/tria.cc
starki0815: Definitely makes sense.

bangerth:
Bizarre -- how did we miss this? I'm also a bit surprised: Why did we end up with the internal face boundary id in these arrays? Shouldn't `merge_triangulation()` have just copied the boundary_ids of the original triangulations (which were definitely not internal, given that these faces were at the boundary of those triangulations)?

masterleinad:
According to the documentation, boundary indicators are never copied.

starki0815:
Correct, the boundary ids are not copied at all. This is not a problem with `merge_triangulations()` as long as manifold ids are not copied over, because in this case the `SubCellData` data structure is left empty and `create_triangulation()` makes sure that the boundary ids are 0 on the external boundary and `internal_face_boundary_id` in the interior. But if manifold ids are copied over, the `SubCellData` data structure is created and something has to be assigned to the boundary ids. And this is the point, where the `internal_face_boundary_id` gets into these arrays as a default choice.

I think another, perhaps better approach to deal with the problem directly in `merge_triangulations()` (instead of my patch which does it in `create_triangulation()`) would be the following: Assign 0 to the boundary ids in the `SubCellData` data structure of the merged triangulation wherever the boundary id in the `SubCellData` data structure of the underlying triangulation is something other than `internal_face_boundary_id`. This way some boundary ids would be set to 0 instead of `internal_face_boundary_id` on internal faces and lines in the `SubCellData` data structure of the merged triangulation (everywhere, where the original triangulations are "glued" together). However, it seems to me that `create_triangulation()` does anyway only care about boundary ids present in the  `SubCellData` data structure which are related to lines and quads at the boundary. So this wouldn't be a problem. The advantage of this approach is that it would be possible to assert in `create_triangulation()` on the presence of `internal_face_boundary_id` in the `SubCellData` data structure whenever a cell is at the boundary.

bangerth:
On 10/17/19 2:27 AM, Sebastian Stark wrote:
> Correct, the boundary ids are not copied at all. This is not a problem with 
> |merge_triangulations()| as long as manifold ids are not copied over, because 
> in this case the |SubCellData| data structure is left empty and 
> |create_triangulation()| makes sure that the boundary ids are 0 on the 
> external boundary and |internal_face_boundary_id| in the interior. But if 
> manifold ids are copied over, the |SubCellData| data structure is created and 
> something has to be assigned to the boundary ids. And this is the point, where 
> the |internal_face_boundary_id| gets into these arrays as a default choice.

Ah yes, this is the part I was missing!


> I think another, perhaps better approach to deal with the problem directly in 
> |merge_triangulations()| (instead of my patch which does it in 
> |create_triangulation()|) would be the following: Assign 0 to the boundary ids 
> in the |SubCellData| data structure of the merged triangulation wherever the 
> boundary id in the |SubCellData| data structure of the underlying 
> triangulation is something other than |internal_face_boundary_id|. This way 
> some boundary ids would be set to 0 instead of |internal_face_boundary_id| on 
> internal faces and lines in the |SubCellData| data structure of the merged 
> triangulation (everywhere, where the original triangulations are "glued" 
> together). However, it seems to me that |create_triangulation()| does anyway 
> only care about boundary ids present in the |SubCellData| data structure which 
> are related to lines and quads at the boundary. So this wouldn't be a problem. 
> The advantage of this approach is that it would be possible to assert in 
> |create_triangulation()| on the presence of |internal_face_boundary_id| in the 
> |SubCellData| data structure whenever a cell is at the boundary.

That could work as well if you felt ambitious, though I think that the current 
approach is also reasonable. Instead of 0, I would choose an invalid value.

PR #8898, opened 2019-10-11
Step-61 Use FE_DGRT for Darcy velocity

sophy1029:
Use FE_DGRT to calculate Darcy velocity instead of FE_RT.

### DIFF
--- a/examples/step-61/step-61.cc
+++ b/examples/step-61/step-61.cc
@@ -43,6 +43,7 @@
 #include <deal.II/dofs/dof_tools.h>
 #include <deal.II/fe/fe_dgq.h>
 #include <deal.II/fe/fe_raviart_thomas.h>
+#include <deal.II/fe/fe_dg_vector.h>
 #include <deal.II/fe/fe_system.h>
 #include <deal.II/fe/fe_values.h>
 #include <deal.II/fe/fe_face.h>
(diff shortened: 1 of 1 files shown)


bangerth:
It looks like you need to run the indentation script. Do you remember how to do that?

bangerth:
Actually, it looks like you need to fix this error:
```
invalid author 'sophy1029' without firstname and lastname
```
This means that you need to set your first and last name in the git repository. Does anyone recall the exact commands for this?

masterleinad:
```
git commit --amend --author="Author Name <email@address.com>"
```
for fixing pushed commits and 
```
git config --global user.name "John Doe"
git config --global user.email "john@doe.org"
```
for setting these information globally.

sophy1029:
> It looks like you need to run the indentation script. Do you remember how to do that?

I tried to run the indent under step-61 branch, but I got the error:

bash-3.2$ ./contrib/utilities/download_clang_format
create folder /Applications/deal.II-9.0.0.app/dealii/contrib/utilities/programs
Downloading and installing clang-format-6 from https://github.com/dealii/dealii/releases/download/v9.0.0/clang-format-6-darwin.tar.gz
Using wget to download...
shasum: clang-format-6-darwin.tar.gz: 
clang-format-6-darwin.tar.gz: FAILED open or read
shasum: WARNING: 1 listed file could not be read
*** The downloaded file has the wrong SHA256 checksum!

How should I fix this?

masterleinad:
Does `wget` return successfully? AFAICT, the link is valid at least.
Can you manually extract the archive and move its content into "contrib/utilities/programs/clang-6/"?

PR #8899, opened 2019-10-11
ArpackSolver sets number of iterations in SolverControl after finishing calculations

anates:
Inspired by https://groups.google.com/forum/#!topic/dealii/jzGIgdUfrR8. SolverControl gets the correct number of iterations after Arpack has finished. Residual is set to 0.0 here, i.e. will be fixed in a future commit.


bangerth:
Yes, something like this. I think it would be good to have a test for it as well, though.

anates:
Then I just need to read a bit about the theory behind it. Not sure, though, if I can add a useful residual, Arpack returns a residual vector, not a single number.

bangerth:
I think just setting the number of iterations as you do here is good enough -- it's strictly progress compared to before. The test doesn't actually have to solve anything useful -- it's just that we should have a record in the test suite that that change was made and to make sure that something fails if the code is rewritten later on without preserving that behavior. It should be enough to just take one of the existing arpack tests and ensure that we also output the number of iterations.

anates:
Added a test. During testing, another internal bug/error came up I would like to fix before merging this branch, if that is ok?

anates:
Added an additional safety check (after running into that problem earlier). Can I add it to that commit, or should I create a new pull request?

bangerth:
In general, we try to keep each pull request restricted to one topic -- that makes reviewing simpler. So in the future, feel free to just add two pull requests.

That said, let's just move ahead with this here now! Thanks for the patch.

By the way, would you like to put an entry into `doc/news/changes`?

anates:
Can do that. If necessary/wished, I can split up both parts into separate requests, the commits are not squashed yet.

bangerth:
No, let's not worry about it here. But feel free to add the changelog entry -- you should get credit for your work!

anates:
I understand why the tests are failing, but not sure why that is happening after my changes...

doc/news/changes/minor/20191710Richter
```diff
@@ -0,0 +1,3 @@
+Changed: SolverControl returns number of iterations of ArpackSolver after solving for Eigenvalues. It also checks if the requested eigenvalues are supported by the solver, depending on symmetry.
```
masterleinad: Would you mind breaking the line at around 80 characters here as well?

bangerth:
These are existing tests. If they now produce different output (and if you have verified that the new output is correct!), then you could just adjust the stored `.output` file to the now correct values.

anates:
Apparently calling check() results in an additional output line with the number of iterations and the residual value, thus resulting in a diff failure. The calculated eigenvalues stay the same, therefore I assume a simple update of the test output files should be enough?

anates:
After some tests, I noticed that i get 8 iterations for the test step-36_ar_shift.release, but only four for the test step-36_ar_shift.debug. The eigenvalues are the same, but I can not use the same output file for tests (due to the different iteration numbers). Any solutions for that?

tests/arpack/step-36_ar_with_iterations.cc
```diff
@@ -0,0 +1,373 @@
+/* ---------------------------------------------------------------------
```
masterleinad: This file is an exact copy of `tests/arpack/step-36_ar.cc` with the only difference being https://github.com/dealii/dealii/pull/8899/files#diff-20c5b8e0204331b74f322770a09c0407R320. It should be pretty fine to just add this one line to the original test.

bangerth:
@masterleinad has a point, but let's not prolong the agony of this PR any longer and instead make sure that @anates gets hits patch merged!

PR #8902, opened 2019-10-11
Make output cells, faces and co-faces optional in GridOut::write_vtk

peterrum:
This PR enables us to output cells, faces and co-faces/edges in `GridOut::write_vtk` separately. (Or is there an alternative existing way to do this - some filters in Paraview or so?)

ping @luca-heltai 

FYI @nfehn


source/grid/grid_out.cc
```diff
@@ -3051,18 +3051,22 @@ GridOut::write_vtk(const Triangulation<dim, spacedim> &tria,
...
-  const int cells_size =
-    tria.n_active_cells() * (GeometryInfo<dim>::vertices_per_cell + 1) +
-    faces.size() * (GeometryInfo<dim>::vertices_per_face + 1) +
-    co_faces.size() * (3); // only in 3d, otherwise it is always zero.
+  const int cells_size = vtk_flags.draw_cells * tria.n_active_cells() *
+                           (GeometryInfo<dim>::vertices_per_cell + 1) +
+                         vtk_flags.draw_faces * faces.size() *
+                           (GeometryInfo<dim>::vertices_per_face + 1) +
+                         vtk_flags.draw_co_faces * co_faces.size() *
+                           (3); // only in 3d, otherwise it is always zero.
```
luca-heltai: Can you add an asserthrow here to make sure that cell_size is non zero? With the flags, one could in principle have no output at all.

peterrum: @luca-heltai Thanks for the review! I have added the assert and a test, which I I think matches your description.

include/deal.II/grid/grid_out.h
```diff
@@ -865,7 +865,31 @@ namespace GridOutFlags
...
+    Vtk()
+      : draw_cells(true)
+      , draw_faces(true)
+      , draw_co_faces(true)
+    {}
+
+    /**
+     * Draw cells.
+     */
+    bool draw_cells;
```
bangerth: I think `output_cells` etc would be a more appropriate name. Nothing is being drawn by this class :-)

source/grid/grid_out.cc
```diff
@@ -3051,19 +3051,24 @@ GridOut::write_vtk(const Triangulation<dim, spacedim> &tria,
   const auto co_faces = relevant_co_faces(tria);
 
   // Write cells preamble
-  const int n_cells = tria.n_active_cells() + faces.size() + co_faces.size();
+  const int n_cells = vtk_flags.draw_cells * tria.n_active_cells() +
```
bangerth: This and the next two lines make use of the implicit conversion of `bool` to `int`. That's hard to read. Make it explicit:
```suggestion
  const int n_cells = (vtk_flags.draw_cells ? tria.n_active_cells() : 0 ) +
```
and similarly for the next couple of lines

source/grid/grid_out.cc
```diff
@@ -3051,19 +3051,24 @@ GridOut::write_vtk(const Triangulation<dim, spacedim> &tria,
...
   // cell connectivity information. The connectivity information is read cell by
   // cell, first specifying how many vertices are required to describe the cell,
   // and then specifying the index of every vertex. This means that for every
   // deal.II object type, we always need n_vertices + 1 integer per cell.
   // Compute the total number here.
-  const int cells_size =
-    tria.n_active_cells() * (GeometryInfo<dim>::vertices_per_cell + 1) +
-    faces.size() * (GeometryInfo<dim>::vertices_per_face + 1) +
-    co_faces.size() * (3); // only in 3d, otherwise it is always zero.
+  const int cells_size = vtk_flags.draw_cells * tria.n_active_cells() *
```
bangerth: same here

include/deal.II/grid/grid_out.h
```diff
@@ -865,7 +865,31 @@ namespace GridOutFlags
    * @ingroup output
    */
   struct Vtk : public DataOutBase::VtkFlags
-  {};
+  {
+    /**
+     * Default constructor.
+     */
+    Vtk()
```
bangerth: We typically write these constructors in a way so that they take the initializer values for the member variables as arguments, with default values `=true` in this case.

source/grid/grid_out.cc
```diff
@@ -3051,19 +3051,24 @@ GridOut::write_vtk(const Triangulation<dim, spacedim> &tria,
...
+                           (GeometryInfo<dim>::vertices_per_cell + 1) +
+                         vtk_flags.draw_faces * faces.size() *
+                           (GeometryInfo<dim>::vertices_per_face + 1) +
+                         vtk_flags.draw_co_faces * co_faces.size() *
+                           (3); // only in 3d, otherwise it is always zero.
 
+  AssertThrow(cells_size > 0, ExcMessage("No cells given to be drawn!"));
```
bangerth:
```suggestion
  AssertThrow(cells_size > 0, ExcMessage("No cells given to be output!"));
```

source/grid/grid_out.cc
bangerth: I think one might also want to have an assertion somewhere that not all three of the flags are `false` at the same time.

bangerth:
@peterrum -- any news?

peterrum:
@bangerth Sorry for the delay! Thanks for the review! I have integrated your suggestions!

bangerth:
@luca-heltai still has to remove his block before this can be merged.

bangerth:
@luca-heltai ?

PR #8903, opened 2019-10-14
VectorizedArray: Update documentation POD -> StandardLayoutType

kronbichler:
When checking up some information about the array, I realized that we forgot to update the documentation of `VectorizedArray` when we switched from POD type to the standard layout type.

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -254,14 +254,15 @@ public:
  * This behavior of this class is made similar to the basic data types double
  * and float. The definition of a vectorized array does not initialize the
  * data field but rather leaves it undefined, as is the case for double and
- * float. However, when calling something like VectorizedArray<double> a =
- * VectorizedArray<double>(), it sets all numbers in this field to zero. In
- * other words, this class is a plain old data (POD) type which has an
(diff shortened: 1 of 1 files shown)


include/deal.II/base/vectorization.h
```diff
@@ -254,28 +254,28 @@ class VectorizedArrayBase
...
- * VectorizedArray<double>(), it sets all numbers in this field to zero. In
- * other words, this class is a plain old data (POD) type which has an
- * equivalent C representation and can e.g. be safely copied with std::memcpy.
- * This POD layout is also necessary for ensuring correct alignment of data
- * with address boundaries when collected in a vector (i.e., when the first
- * element in a vector is properly aligned, all subsequent elements will be
- * correctly aligned, too).
+ * float. However, when calling something like `VectorizedArray<double> a =
+ * VectorizedArray<double>()` or `VectorizedArray<double> a = 0.`, it sets all
+ * numbers in this field to zero. This class of standard layout type according
```
masterleinad:
```suggestion
 * numbers in this field to zero. This class has standard layout type according
```
or
```suggestion
 * numbers in this field to zero. This class is of standard layout type according
```
?

kronbichler: I will go with this one, thanks!

PR #8904, opened 2019-10-14
Simplify code for writing vtu/pvtu output files in application programs

nfehn:
Currently, no functionality is available that automatically adds filename extensions like counter, processor ID, and .vtu/.pvtu ending to the filenames in ``write_vtu`` and ``write_pvtu_record``. Instead, the same code is written again and again. Currently 10 tutorials (step-37/18/32/45/55/48/42/64/50/40) essentially copy-paste 20 lines of code from the step-40 tutorial (and probably every deal.II user has reimplemented this code more than once in other application code). 

This Pull Request provides a new interface in order to free the deal.II user from this kind of formatting. Unlike in step-40 where it is written that "PPPP refers to the processor number (enough for up to 10,000 processors, though we hope that nobody ever tries to generate this much data – you would likely overflow all file system quotas)", the number of digits reserved in the filename for the processor ID is determined automatically. The number of output files written (e.g., cycles or mesh refinement levels as in step-40, or time steps for unsteady simulations of time dependent problems) depends on the application and can be specified as parameter. Padding with leading zeros is omitted if the default value is used. Since the formatting in step-40 does not allow to identify which number is the counter and which one the processor ID by looking at the filenames only, the formatting has been changed as compared to step-40.

### DIFF
--- a/include/deal.II/base/data_out_base.h
+++ b/include/deal.II/base/data_out_base.h
@@ -3031,6 +3031,56 @@ public:
                     const std::vector<std::string> &piece_names) const;
 
   /**
+   * This function combines DataOutInterface::write_vtu(), or
+   * DataOutInterface::write_vtu_in_parallel(), with
+   * DataOutInterface::write_pvtu_record(). It automatically constructs
+   * the filename of the generated output files and writes a pvtu file.
+   * A specified @p directory and a @p filename_without_extension
+   * form the first part of the filename. The filename is then extended with
+   * a @p counter labeling the current timestep/iteration/etc., the processor ID,
+   * and finally the .vtu/.pvtu ending. Since the number of timesteps to be
+   * written depends on the application, the number of digits to be reserved in
+   * the filename can be specified as parameter @p n_digits_for_counter, and the number
+   * is not padded with leading zeros if this parameter is left at its default
+   * value numbers::invalid_unsigned_int. If more than one file identifier
+   * is needed (e.g. time step number and iteration counter of solver), the
+   * last identifier is used as @p counter, while all other identifiers have to be
+   * added to @p filename_without_extension when calling this function. In a
+   * parallel setting, several files are typically written per time step. The
+   * number of files written in parallel depends on the number of MPI processes
+   * (see parameter @p mpi_communicator with default value MPI_COMM_WORLD), and a
+   * specified number of @p n_groups with default value 0. The background is that
+   * VTU file output supports grouping files from several CPUs into a given
+   * number of files using MPI I/O when writing on a parallel filesystem. The
+   * default value of @p n_groups is 0, meaning that every MPI rank will write one
+   * file. A value of 1 will generate one big file containing the solution over
+   * the whole domain, while a larger value will create @p n_groups files (but not
+   * more than there are MPI ranks). Note that only one processor needs to
+   * generate the .pvtu file, where processor zero is chosen to take over this
+   * job.
+   *
+   * The return value is the filename of the master file for the pvtu record.
+   *
+   * @note The code simply combines the strings @p directory and
+   * @p filename_without_extension, i.e., the user has to make sure that
+   * @p directory contains a trailing character, e.g. "/", that separates the
+   * directory from the filename.
+   *
+   * @note Use an empty string "" for the first argument if output is to be
+   * written in the current working directory.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  std::string
+  write_vtu_with_pvtu_record(
+    const std::string &directory,
+    const std::string &filename_without_extension,
+    const unsigned int counter,
+    const unsigned int n_digits_for_counter = numbers::invalid_unsigned_int,
+    const MPI_Comm &   mpi_communicator     = MPI_COMM_WORLD,
+    const unsigned int n_groups             = 0) const;
+
+  /**
    * Obtain data through get_patches() and write it to <tt>out</tt> in SVG
    * format. See DataOutBase::write_svg.
    */
--- a/source/base/data_out_base.cc
+++ b/source/base/data_out_base.cc
@@ -7544,6 +7544,83 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
 }
 
 
+template <int dim, int spacedim>
+std::string
+DataOutInterface<dim, spacedim>::write_vtu_with_pvtu_record(
+  const std::string &directory,
+  const std::string &filename_without_extension,
+  const unsigned int counter,
+  const unsigned int n_digits_for_counter,
+  const MPI_Comm &   mpi_communicator,
+  const unsigned int n_groups) const
+{
+  const unsigned int rank = Utilities::MPI::this_mpi_process(mpi_communicator);
+  const unsigned int n_ranks =
+    Utilities::MPI::n_mpi_processes(mpi_communicator);
+  const unsigned int n_files_written =
+    (n_groups == 0 || n_groups > n_ranks) ? n_ranks : n_groups;
+
+  const unsigned int n_digits =
+    static_cast<int>(std::ceil(std::log10(std::fabs(n_files_written))));
+
+  const unsigned int color = rank % n_files_written;
+  const std::string  filename =
+    directory + filename_without_extension + "_" +
+    Utilities::int_to_string(counter, n_digits_for_counter) + "." +
+    Utilities::int_to_string(color, n_digits) + ".vtu";
+
+  if (n_groups == 0 || n_groups > n_ranks)
+    {
+      // every processor writes one file
+      std::ofstream output(filename.c_str());
+      this->write_vtu(output);
+    }
+  else if (n_groups == 1)
+    {
+      // write only a single data file in parallel
+      this->write_vtu_in_parallel(filename.c_str(), mpi_communicator);
+    }
+  else
+    {
+#ifdef DEAL_II_WITH_MPI
+      // write n_groups data files
+      MPI_Comm comm_group;
+      int ierr = MPI_Comm_split(mpi_communicator, color, rank, &comm_group);
+      AssertThrowMPI(ierr);
+      this->write_vtu_in_parallel(filename.c_str(), comm_group);
+      ierr = MPI_Comm_free(&comm_group);
+      AssertThrowMPI(ierr);
+#else
+      AssertThrow(false, ExcMessage("Logical error. Should not arrive here."));
+#endif
+    }
+
+  // write pvtu record
+  const std::string filename_master =
+    filename_without_extension + "_" +
+    Utilities::int_to_string(counter, n_digits_for_counter) + ".pvtu";
+
+  if (rank == 0)
+    {
+      std::vector<std::string> filename_vector;
+      for (unsigned int i = 0; i < n_files_written; ++i)
+        {
+          const std::string filename =
+            filename_without_extension + "_" +
+            Utilities::int_to_string(counter, n_digits_for_counter) + "." +
+            Utilities::int_to_string(i, n_digits) + ".vtu";
+
+          filename_vector.emplace_back(filename);
+        }
+
+      std::ofstream master_output((directory + filename_master).c_str());
+      this->write_pvtu_record(master_output, filename_vector);
+    }
+
+  return filename_master;
+}
+
+
 
 template <int dim, int spacedim>
 void
--- a/doc/news/changes/minor/20191014NiklasFehn
+++ b/doc/news/changes/minor/20191014NiklasFehn
@@ -1,0 +1,6 @@
+New: The function DataOutInterface::write_vtu_with_pvtu_record() combines
+write_vtu() and write_pvtu_record() into a single function and generates 
+the file extensions with counter, processor ID, and .vtu/.pvtu ending 
+automatically.
+<br>
+(Niklas Fehn, 2019/10/14)


include/deal.II/base/data_out_base.h
```diff
@@ -3030,6 +3030,35 @@ class DataOutInterface
...
+   * value numbers::invalid_unsigned_int. Note that only one processor needs to
+   * generate the .pvtu file, where processor zero is chosen to take over this
+   * job.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  void
+  write_vtu_with_pvtu_record(
+    const std::string &folder,
+    const std::string &filename_without_extension,
```
kronbichler: I think we can reduce redundancy by extracting the filename without folders from the string. On Unix-like systems, one can search whether `filename_without_extension` contains any `/` character and remove everything up to that one:
```c++
std::string filename_without_folder;
if (filename_without_extension.find('/') != std::string::npos)
  filename_without_folder = std::string(filename_without_extension.begin()+
                                        filename_without_extension.find_last_of('/')+1,
                                        filename_without_extension.end());
else
  filename_without_folder = filename_without_extension;
Assert(!filename_without_folder.empty(), ExcMessage("Cannot have empty file name"));
```
On Windows, that character would be `\`. Probably there is even a boost function for that, I have not checked.

tjhei:
> Currently 10 tutorials (step-37/18/32/45/55/48/42/64/50/40) essentially copy-paste 20 lines of code from the step-40 tutorial (and probably every deal.II user has reimplemented this code more than once in other application code).

Good observation and thanks for tackling problems like this!

Before I go into reviewing the code itself, a few comments:
1. The format ``name_<cycle>.<rank>.vtu`` in step-40 is done on purpose, so that sorting files by name groups the files by cycle.
2. Using write_vtu_in_parallel() is much more scalable (fewer files, which would break most file systems) and more efficient (using MPI I/O). While we are proposing a nice way to write output, maybe we should consider incorporating this here as well? In ASPECT we allow the user to specify how many files to write (you can make a sensible compromise between number of files as a single .vtu is also not desirable as it can become too big). This is quite easy to do:
https://github.com/geodynamics/aspect/blob/master/source/postprocess/visualization.cc#L494-L498
Here, ``group_files`` is an int specified by the user:
https://github.com/geodynamics/aspect/blob/master/source/postprocess/visualization.cc#L830-L838

include/deal.II/base/data_out_base.h
nfehn: In my opinion, this has some disadvantages and I want to share them before deciding for this variant. For me, ``folder`` + ``filename`` is more natural. On the one hand, the code has to separate into folder and filename anyway, so the implementation suggested above is more complicated and error prone. On the other hand, I think the suggested version is not self-documenting, as it is not clear what ``filename_without_extension`` is. Sure, one might define ``filename`` in a way that it always includes the folder. However, assume we have two parameters ``folder`` and ``filename``, I think this does not allow misunderstandings, as it is clear that one does not expect that ``filename`` also includes the folder in this case. Finally, in my experience an application code has separate parameters for folder and filename anyway as soon as it becomes more complicated (because the user has to specify several names for different output files in an input file that collects all parameters but does not want to rewrite the folder for all instances, and hence uses a separate parameter for the folder). I have the concern that in practice we first combine folder and filename while the code later separates them again, while it is the aim of this PR to simplify interfaces. All tutorials either write the output in the current working directory or use two separate parameters for folder and filename (see step-42). Hence, there is no compatibility problem in the other direction in the sense that the filename has to be separated into folder and filename manually by the user before calling the new function ``write_vtu_with_pvtu_record()``.

peterrum:
I agree with @tjhei regarding `write_vtu_in_parallel` and very much like the implementation in ASPECT! (In my opinion we should deprecate `write_vtu`.)

include/deal.II/base/data_out_base.h
nfehn: I think it would make more sense to write a free function that can separate a string into folder and filename_without_folder in case that some code uses this format (a single filename including the folder). It needs several lines of code to implement this and why should this very general functionality be implemented in ``write_vtu_with_pvtu_record``?

nfehn:
@kronbichler I have seen that ASPECT also separates folder and filename.

peterrum:
I would suggest to keep `write_vtu_in_parallel` out of this PR and proceed with this PR as it is. 

In an upcoming PR, we can tackle `write_vtu_in_parallel`. I think it would be very useful to have a `_write_pvtu_record` for that, too. However, for that I have a question/comment: why are there two functions `write_vtu` and `write_vtu_in_parallel`? The first one could be done with the second one (with group size one.)

To reformulate my comment from above: I would move the functionality from `write_vtu` into an internal namespace; move the content of `write_vtu_in_parallel` into `write_vtu` and extend the functionality with MPI groups (with an optional argument of group size, which is per default 1). For this refactored `write_vtu`, we can use `write_vtu_with_pvtu_record` from this PR with small modifications (without having to write code twice). @nfehn @tjhei do you agree?

I am willing to work on the follow up PR!

nfehn:
Yes, I agree with @peterrum . 

The name ``write_vtu_in_parallel()`` has probably historical reasons but should be implemented in what is currently called ``write_vtu()``, as it is the aim of deal.II do things efficiently in parallel, so we do not need to highlight this by function names? I think the current interface is prepared for the parallel-vtu version. We would extend ``write_vtu_with_pvtu_record()`` by another parameter ``n_groups`` as in ASPECT, taking a default value of 1 so that no compatibility problems arise related to the current PR.

nfehn:
@tjhei In ASPECT, you are using another ID for the nonlinear iterations in addition to the time step. The current interface is prepared for that. In this case, the user has to manually add the first ID (time step) to the ``filename`` string, and use the third argument ``counter`` for the nonlinear iterations. As it is conceivable to further extend this (e.g., timestep + nonlinear iteration + linear iteration) it does not make sense to blow up the interface in my opinion. Nevertheless, this is a clear argument to change the order of counter and processor ID, so that processor ID always comes last. I will change this according to your comment.

include/deal.II/base/data_out_base.h
kronbichler: I fully agree on the case that larger codes will put this into subfolders - my argument was that it is an optional parameter for simple programs, setting the first argument of this function to `""`.
On the other hand, I agree that `filename_without_extension` would be ambiguous in case it should contain a folder name, but that is just a matter of naming. Funny though that I initially thought your interface @nfehn makes it more easy to make small mistakes, so here is something more to be documented: when you have separate `folder` and `filename` strings the `folder` argument must include the trailing "/" character - not a big deal but easily forgotten when you are used to bash that does the magic for you.

> I have the concern that in practice we first combine folder and filename while the code later separates them again, while it is the aim of this PR to simplify interfaces.

But what I suggested would just make the internal implementation more complicated, not the interface.

In any case, I do not feel strongly about either way, as long as it is well documented what is required.

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,50 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+    {
+      std::vector<std::string> vector;
+      for (unsigned int i = 0; i < n_ranks; ++i)
+        {
+          std::string filename =
+            filename_without_extension + "_" +
+            Utilities::int_to_string(counter, n_digits_counter) + "." +
+            Utilities::int_to_string(i, n_digits) + ".vtu";
+
+          vector.push_back(filename.c_str());
```
peterrum:
```
clang-tidy-6.0 -header-filter=/jenkins/workspace/dea/jenkins/workspace/dealii-tidy_PR-8904/source/base/data_out_base.cc:7579:18: error: use emplace_back instead of push_back [modernize-use-emplace,-warnings-as-errors]
```

tjhei:
> The name `write_vtu_in_parallel()` has probably historical reasons but should be implemented in what is currently called `write_vtu()`, as it is the aim of deal.II do things efficiently in parallel, so we do not need to highlight this by function names?

My fault. It was made in a time where I (we?) were more hesitant to modify things.

> 1). For this refactored `write_vtu`, we can use `write_vtu_with_pvtu_record` from this PR with small modifications (without having to write code twice)

As long as the new write_vtu with group=1 uses a code path without MPI I/O (there have been bug reports in ASPECT where some clusters on some filesystems have a broken MPI I/O interface).

> I would suggest to keep `write_vtu_in_parallel` out of this PR and proceed with this PR as it is.

I feel we should have the conversation about this first, instead of having to modify the API introduced for this PR again later on. Whatever you introduce here (and consequently use in all tutorials), should include the grouping and MPI I/O support.

nfehn:
> I feel we should have the conversation about this first, instead of having to modify the API introduced for this PR again later on. Whatever you introduce here (and consequently use in all tutorials), should include the grouping and MPI I/O support.

@tjhei please have a look at the new commit. It includes the parameter ``n_groups`` and the ``write_vtu_in_parallel()`` case. Everything else is done automatically. A pvtu-file is written if more than one file is written in parallel (and a pvtu file is also written for the ``write_vtu_in_parallel()`` version used in ASPECT).

It could make sense, as suggest by @peterrum , to remove the function ``write_vtu_in_parallel()`` and include this functionality in ``write_vtu()``, but this is subject of a different PR.

include/deal.II/base/data_out_base.h
```diff
@@ -3030,6 +3030,55 @@ class DataOutInterface
...
+   * parallel setting, several files are typically written per time step. The
+   * number of files written in parallel depends on the number of MPI processes
+   * (see parameter @p mpi_communicator with default value MPI_COMM_WORLD), and a
+   * specified number of @p n_groups with default value 0. The background is that
+   * VTU file output supports grouping files from several CPUs into a given
+   * number of files using MPI I/O when writing on a parallel filesystem. The
+   * default value of @p n_groups is 0, meaning that every MPI rank will write one
+   * file. A value of 1 will generate one big file containing the solution over
+   * the whole domain, while a larger value will create @p n_groups files (but not
+   * more than there are MPI ranks). A.pvtu file is written if more than one
```
tjhei: I would always write the .pvtu file based on the name of this function. We typically want to generate a master record with all .pvtu files as well, which would be awkward if n_ranks==1 is a special case.

tjhei:
```suggestion
   * more than there are MPI ranks). A .pvtu file is written if more than one
```

include/deal.II/base/data_out_base.h
```diff
@@ -3030,6 +3030,55 @@ class DataOutInterface
...
+   * DataOutInterface::write_pvtu_record(). It automatically constructs
+   * the filename of the generated output files and writes a pvtu file if
+   * necessary. A specified @p folder and a @p filename_without_extension
+   * form the first part of the filename. The filename is then extended with
+   * a @p counter labeling the current timestep/iteration/etc., the processor ID,
+   * and finally the .vtu/.pvtu ending. Since the number of timesteps to be
+   * written depends on the application, the number of digits to be reserved in
+   * the filename can be specified as parameter @p n_digits_counter, and the number
+   * is not padded with leading zeros if this parameter is left at its default
+   * value numbers::invalid_unsigned_int. If more than one file identifier
```
tjhei: why don't you use 0 as the default for no padding?

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+  const unsigned int rank = Utilities::MPI::this_mpi_process(mpi_communicator);
+  const unsigned int n_ranks =
+    Utilities::MPI::n_mpi_processes(mpi_communicator);
+  const unsigned int n_files_written =
+    (n_groups == 0 || n_groups > n_ranks) ? n_ranks : n_groups;
+
+  const unsigned int n_digits =
+    std::ceil(std::log10(std::fabs(n_files_written)));
+
+  const unsigned int color =
```
tjhei: can't you use ``rank % n_files_written``?

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+    (n_groups == 0 || n_groups > n_ranks) ? rank : rank % n_groups;
+  std::string processor_id =
+    (n_files_written > 1) ? ("." + Utilities::int_to_string(color, n_digits)) :
```
tjhei: I find it surprising that the filename changes if I happen to run with ``mpirun -n 1``. This function is made for parallel output, after all.

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+                            "";
+  std::string filename = folder + filename_without_extension + "_" +
+                         Utilities::int_to_string(counter, n_digits_counter) +
+                         processor_id + ".vtu";
+  std::ofstream output(filename.c_str());
```
tjhei: only open the file if you use ``write_vtu``

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+
+  // every processor writes one file
```
tjhei: I would move these comments inside the { block

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+  // write n_groups data files
+  else
+    {
+      MPI_Comm comm_group;
+      MPI_Comm_split(mpi_communicator, color, rank, &comm_group);
+      this->write_vtu_in_parallel(filename.c_str(), comm_group);
+      MPI_Comm_free(&comm_group);
+    }
+
+  // pvtu record only needed if more than one file is written
```
tjhei: as commented above, I would always write this record.

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+  if (n_files_written > 1)
+    {
+      if (rank == 0)
+        {
+          std::vector<std::string> vector;
```
tjhei: please don't call this variable ``vector``

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+          for (unsigned int i = 0; i < n_files_written; ++i)
+            {
+              std::string filename =
+                filename_without_extension + "_" +
+                Utilities::int_to_string(counter, n_digits_counter) + "." +
+                Utilities::int_to_string(i, n_digits) + ".vtu";
+
+              vector.push_back(filename);
```
tjhei: emplace_back

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,80 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+            }
+
+          std::string master_name =
+            folder + filename_without_extension + "_" +
+            Utilities::int_to_string(counter, n_digits_counter) + ".pvtu";
+          std::ofstream master_output(master_name.c_str());
+          this->write_pvtu_record(master_output, vector);
```
tjhei: I think it would be very convenient if we can return the name of the master record back to the user (to build the master-master records later).

tjhei:
> It could make sense, as suggest by @peterrum , to remove the function `write_vtu_in_parallel()` and include this functionality in `write_vtu()`, but this is subject of a different PR.

yes and yes. :-)

source/base/data_out_base.cc
nfehn: yes, it should be the same

source/base/data_out_base.cc
nfehn: how would that look like exactly? Only processor 0 has the master_name?

include/deal.II/base/data_out_base.h
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/f07a8bf9bce24520305b079bc179599d78782753

nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/f07a8bf9bce24520305b079bc179599d78782753

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/f07a8bf9bce24520305b079bc179599d78782753

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/f07a8bf9bce24520305b079bc179599d78782753

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/f07a8bf9bce24520305b079bc179599d78782753

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/f07a8bf9bce24520305b079bc179599d78782753

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/f07a8bf9bce24520305b079bc179599d78782753

source/base/data_out_base.cc
nfehn: what is the return type?

tjhei: std::string with the value ``filename_without_extension + "_" +
            Utilities::int_to_string(counter, n_digits_counter) + ".pvtu"``

nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/7b261624b8d14ff7a1a6e68b29aeef2b520dd05e

peterrum:
`MPI_Comm_split` and `MPI_Comm_free` are not known if deal.II is compiled without MPI. You have to guard these function calls with `#ifdef DEAL_II_WITH_MPI`.

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,81 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+    {
+      // write n_groups data files
+      MPI_Comm comm_group;
+#ifdef DEAL_II_WITH_MPI
+      MPI_Comm_split(mpi_communicator, color, rank, &comm_group);
+#endif
+      this->write_vtu_in_parallel(filename.c_str(), comm_group);
+#ifdef DEAL_II_WITH_MPI
+      MPI_Comm_free(&comm_group);
+#endif
```
peterrum:
```suggestion
#ifdef DEAL_II_WITH_MPI
      MPI_Comm comm_group;
      MPI_Comm_split(mpi_communicator, color, rank, &comm_group);
      this->write_vtu_in_parallel(filename.c_str(), comm_group);
      MPI_Comm_free(&comm_group);
#endif
```

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,81 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+      MPI_Comm_split(mpi_communicator, color, rank, &comm_group);
+      this->write_vtu_in_parallel(filename.c_str(), comm_group);
+      MPI_Comm_free(&comm_group);
+#elif
+      AssertThrow(false, ExcMessage("Logical error. Should not arrive here."));
+#endif
+    }
+
+  // write pvtu record
+  std::string filename_master =
```
tjhei:
```suggestion
  const std::string filename_master =
```

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,81 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+    filename_without_extension + "_" +
+    Utilities::int_to_string(counter, n_digits_counter) + ".pvtu";
+
+  if (rank == 0)
+    {
+      std::vector<std::string> filename_vector;
+      for (unsigned int i = 0; i < n_files_written; ++i)
+        {
+          std::string filename =
```
tjhei:
```suggestion
          const std::string filename =
```

masterleinad:
```
/jenkins/workspace/dealii_PR-8904/source/base/data_out_base.cc:7591:6: error: #elif with no expression
```

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/192d009860b2e79bff73f6e636b71442aecfbabd

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/192d009860b2e79bff73f6e636b71442aecfbabd

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,81 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+  const unsigned int color = rank % n_files_written;
+  const std::string  filename =
+    folder + filename_without_extension + "_" +
+    Utilities::int_to_string(counter, n_digits_counter) + "." +
+    Utilities::int_to_string(color, n_digits) + ".vtu";
+
+  if (n_groups == 0 || n_groups > n_ranks)
+    {
+      // every processor writes one file
+      std::ofstream output(filename.c_str());
```
nfehn: previous comment by @tjhei  done in https://github.com/dealii/dealii/pull/8904/commits/734946826eada01551285a7d60270107fff476f0

tjhei:
thanks for being patient with my endless list of requests. ;-)

peterrum:
You could write `static_cast<int>(std::ceil(std::log10(std::fabs(n_files_written)));)`...

drwells:
```
/source/base/data_out_base.cc:7564:5: error: implicit conversion turns floating-point number into integer: 'double' to 'const unsigned int' [-Werror,-Wfloat-conversion]

    std::ceil(std::log10(std::fabs(n_files_written)));
```
It looks like we are missing an explicit cast. Only the mac build complained. Once we add it we should be done :)

include/deal.II/base/data_out_base.h
```diff
@@ -3030,6 +3030,56 @@ class DataOutInterface
...
+   * folder from the filename.
+   *
+   * @note Use an empty string "" for the first argument if output is to be
+   * written in the current working directory.
+   *
+   * @author Niklas Fehn, Martin Kronbichler, 2019
+   */
+  std::string
+  write_vtu_with_pvtu_record(
+    const std::string &folder,
```
bangerth: The standard naming on Unix is to use "directory" instead of "folder". The latter is a windows/mac thing.

include/deal.II/base/data_out_base.h
```diff
@@ -3030,6 +3030,56 @@ class DataOutInterface
...
+    const std::string &filename_without_extension,
+    const unsigned int counter,
+    const unsigned int n_digits_counter = numbers::invalid_unsigned_int,
```
bangerth: maybe `n_digits_for_counter`?

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,81 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+    {
+      // write only a single data file in parallel
+      this->write_vtu_in_parallel(filename.c_str(), mpi_communicator);
+    }
+  else
+    {
+#ifdef DEAL_II_WITH_MPI
+      // write n_groups data files
+      MPI_Comm comm_group;
+      MPI_Comm_split(mpi_communicator, color, rank, &comm_group);
```
bangerth: Here and a couple of lines below, we ought to check the return code. We have `AssertThrowMpi` for this.

include/deal.II/base/data_out_base.h
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/829b7f07d37f31b09d2b60079380446b5d7abdb2

nfehn: probably originates from the fact that my generation grew up with apple and microsoft ;)

include/deal.II/base/data_out_base.h
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/829b7f07d37f31b09d2b60079380446b5d7abdb2

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,81 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
```
peterrum:
```suggestion
      int ierr = MPI_Comm_split(mpi_communicator, color, rank, &comm_group);
      AssertThrowMPI(ierr);
```

source/base/data_out_base.cc
```diff
@@ -7544,6 +7544,81 @@ DataOutInterface<dim, spacedim>::write_pvtu_record(
...
+      this->write_vtu_in_parallel(filename.c_str(), comm_group);
+      MPI_Comm_free(&comm_group);
```
peterrum:
```suggestion
ierr = MPI_Comm_free(&comm_group);
AssertThrowMPI(ierr);
```

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/554bef32cf65c39b12892a61e937ffebf78e3c1a

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/554bef32cf65c39b12892a61e937ffebf78e3c1a . Thanks!

source/base/data_out_base.cc
nfehn: done in https://github.com/dealii/dealii/pull/8904/commits/554bef32cf65c39b12892a61e937ffebf78e3c1a

PR #8912, opened 2019-10-16
Fix VectorTools::interpolate() for p::d::Triangulation

dougshidong:
More specifically this following overloaded version.
void VectorTools::interpolate(
    const DoFHandler< dim, spacedim > & dof_1,
    const DoFHandler< dim, spacedim > & dof_2,
    const FullMatrix< double > & transfer,
    const InVector & data_1,
    OutVector & data_2)

The older version would traverse every cell, including ones that are not locally owned. Basically just added an if-statement.

Checked that it didn't break any existing test using `ctest -R interpolate` and obtained `100% tests passed, 0 tests failed out of 134`

### DIFF
--- a/tests/vector_tools/interpolate.cc
+++ b/tests/vector_tools/interpolate.cc
@@ -1,0 +1,223 @@
+#include <deal.II/base/utilities.h>
+
+#include <deal.II/grid/grid_generator.h>
+
+#include <deal.II/grid/tria.h>
+#include <deal.II/distributed/tria.h>
+
+#include <deal.II/numerics/vector_tools.h>
+
+#include <deal.II/fe/fe_q.h>
+#include <deal.II/fe/fe_system.h>
+
+#include <deal.II/dofs/dof_tools.h>
+
+#include "../tests.h"
+
+template<int dim>
+void run_1d()
+{
+    using Vector = dealii::Vector<double>;
+
+    // Create grid
+    dealii::Triangulation<dim> triangulation(
+        typename dealii::Triangulation<dim>::MeshSmoothing(
+        dealii::Triangulation<dim>::smoothing_on_refinement |
+        dealii::Triangulation<dim>::smoothing_on_coarsening));
+    const unsigned int n_1d_cells = 2;
+    dealii::GridGenerator::subdivided_hyper_cube(triangulation, n_1d_cells);
+
+    // Create low order system
+    dealii::DoFHandler<dim> dof_handler_coarse(triangulation);
+
+    const unsigned int degree_coarse = 2;
+    const dealii::FE_Q<dim> fe_q_coarse(degree_coarse);
+    const dealii::FESystem<dim> fe_system_coarse(fe_q_coarse, dim);
+    dof_handler_coarse.initialize(triangulation, fe_system_coarse);
+    dof_handler_coarse.distribute_dofs(fe_system_coarse);
+
+    Vector solution_coarse;
+    solution_coarse.reinit(dof_handler_coarse.n_dofs());
+    // Initialize dummy coarse solution
+    //solution_coarse.add(1.0);
+    for (unsigned int idof = 0; idof < dof_handler_coarse.n_dofs(); ++idof) {
+        solution_coarse[idof] = idof;
+    }
+
+    const unsigned int degree_fine = 3;
+
+    dealii::DoFHandler<dim> dof_handler_fine(triangulation);
+
+    const dealii::FE_Q<dim> fe_q_fine(degree_fine);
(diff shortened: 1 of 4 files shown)


dougshidong:
Any reason why add contributions from multiple elements and take an average? For continuous finite elements, wouldn't the solution be... continuous? For DG element, those DoFs would be different.

Maybe something to do with hanging nodes?

Anyways, I removed that part and only interpolate to the DoF once, by the first element that encounters it.

There were two more bugs discovered by the test.

First, an initial non-zero output vector would have values added onto it.

Second, DoFs that do not belong to the processors were averaged out over a count that is not MPI-summed over the processors.

masterleinad:
> Any reason why add contributions from multiple elements and take an average? For continuous finite elements, wouldn't the solution be... continuous? For DG element, those DoFs would be different.
> 
> Maybe something to do with hanging nodes?
> 
> Anyways, I removed that part and only interpolate to the DoF once, by the first element that encounters it.

This is simply what the function is doing. There are no requirements on the transfer matrix that would make sure that your changes don't change the output. The new DoF values on each cell are some linear combinations of the old values on each cell and in general, this gives different values for DoFs shared by multiple cells. Averaging the values make sure that the associated "error" is more evenly distributed.

dougshidong:
Oh I see. So the test I just pushed isn't verifying the correct purpose of this function. The test checks that the interpolated solution actually interpolates the old FiniteElement space at the new FiniteElement support points, which I assumed was the whole point of interpolate().

From what I understand of your explanation, this function could accept a zero interpolation matrix. If the goal is to provide generality, I don't see why it is called "interpolation", more so than some linear solution transfer.

Anyways, the latest 2 commits including the test assume that the function's intent is to actually interpolate. Let me know if you'd like me to revert those commits.

masterleinad:
Evaluating "the old FiniteElement space at the new FiniteElement support points" is one of the possible use cases of this function. It might even be the only behavior we use in the codebase and test for. On the other hand, the function allows for other uses and I see no reason for changing that (and possibly breaking user code).
"interpolate" might be a historical misnomer, but this is what we are at right now.

The correct way to extend (or fix) the functionality for distributed triangulations is to loop over all the locally relevant cells and compute the average for all the locally owned dofs.

tests/vector_tools/interpolate.cc
```diff
@@ -0,0 +1,223 @@
+#include <deal.II/base/utilities.h>
```
masterleinad: The copyright header is missing here.

tests/vector_tools/interpolate.output
```diff
@@ -0,0 +1,7 @@
+
```
masterleinad: This tests needs `MPI` and `p4est`. Hence, you should rename the output file to ` tests/vector_tools/interpolate.with_p4est=true.mpirun=1.output`

dougshidong:
> The correct way to extend (or fix) the functionality for distributed triangulations is to loop over all the locally relevant cells and compute the average for all the locally owned dofs.

One thing to keep in mind is that we would need to perform the collective call vector.compress(add) within the interpolate() function and we also need to MPI-sum the touch_counts to correctly average out those cells. If that is OK, I can easily make those changes. I would assume that no one previously used that function in parallel since it would not have previously worked, and therefore, this collective call would not break anyone's code.

masterleinad:
No, you don't need to call sum the contributions across MPI processes. All the information that is required for the locally owned degrees of freedom is contained in the locally relevant cells.

Yes, you need to call `vector.compress(VectorOperations::insert)` at the end of this function.

elauksap:
Hi,

i have just opened a [new pull request](https://github.com/dealii/dealii/pull/8946) that apparently solves the same issue solved by @dougshidong.

In the solution proposed by me, the variable `touch_count` is a parallel distributed vector itself: do you find reasonable to declare it as a `std::vector<short unsigned int> touch_count(dof_2.n_dofs(), 0);`, which is as long as the *total* number of dofs (_i.e._ accumulated over all the processors) rather than, for example, `n_locally_owned_dofs()`?

Thanks!

bangerth:
Just to give you an example of why we *add* instead of set: Think about starting with a piecewise constant field that we want to interpolate onto a piecewise linear one. In that case (and in 2d), the matrix that interpolates from the DGQ(0) to the Q(1) element is simply `[1;1;1;1]`. But you want to take the averages over all adjacent cells when computing the future values of the Q1 field, rather than letting the last cell you touch "win".

dougshidong:
Aaah, yeah ok, that's another great example. I had not thought of about DGQ to Q elements. I think I had taken the the word "interpolate" a bit too literally in this scenario.

I'll make the suggested changes on the other pull request #8946.

PR #8197, opened 2019-05-16
Improve GridGenerator::torus, volume case (dim=3)

kronbichler:
Together with @nfehn, I implemented a fix to the torus:

The function GridGenerator::torus() run in 3D (volume mesh) would previously only use a single cell to represent the poloidal shape, which leads to singular mappings similar to how a circle degenerates when meshed with a single cell. The poloidal shape is now represented with 5 cells just like the circle. Furthermore, the function GridGenerator::torus() has gained an optional argument to control the number of cell layers in the toroidal direction. The default manifold set to the torus has also been improved: Now, the TorusManifold is applied on the surface, a CylindricalManifold to the middle cells in toroidal coordinates, and a TransfiniteInterpolationManifold on the cells between the surface and the inner ring. This leads to an excellent mesh quality for all supported scenarios, as opposed to the previous setup that could lead to bad cells in case there were not enough cells in toroidal direction.

Example:
torus_mesh

Closes #6250.

PR #8913, opened 2019-10-16
fix bug in GridGenerator::torus<3,3>()

nfehn:
For some reasons, the code contained hard-coded values of 2.0 and 0.5 instead of the parameters R and r.


doc/news/changes/minor/20191016NiklasFehn
```diff
@@ -0,0 +1,5 @@
+BugFix: In GridGenerator::torus<3,3>() the radii set for the
+TorusManifold have been wrong, using some hard-coded values like 2.0
+and 0.5 instead of R and r.
+<br>
+(Niklas Fehn, 2019/10/16)
```
kronbichler: This is correcting a change within the current 9.2-pre version of deal.II, namely #8197, for which we usually do not write changelog entry as this is not a change against a previous release version that needs to show up on [this page](https://www.dealii.org/developer/doxygen/deal.II/recent_changes.html). So I think we can drop this file here (which does not mean that your change is not appreciated).

kronbichler:
Can you squash the two commits into one?

nfehn:
I wanted to extend the functionality towards an open torus, see #8916, but fixed the bug in a separate pull request in a first step to not mix up different things.

PR #8915, opened 2019-10-16
Save level_mg_handler inside of MatrixFree

peterrum:
FYI @kronbichler

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -1602,6 +1602,12 @@ public:
   mapping_initialized() const;
 
   /**
+   * Return the level of the mesh to be worked on.
+   */
+  unsigned int
+  get_level_mg_handler() const;
+
+  /**
    * Return an approximation of the memory consumption of this class in
    * bytes.
    */
@@ -1928,6 +1934,11 @@ private:
    */
   mutable std::list<std::pair<bool, AlignedVector<Number>>>
     scratch_pad_non_threadsafe;
+
+  /**
+   * Stored the level of the mesh to be worked on.
+   */
+  unsigned int level_mg_handler;
 };
 
 
@@ -2478,6 +2489,14 @@ inline bool
 MatrixFree<dim, Number, VectorizedArrayType>::mapping_initialized() const
 {
   return mapping_is_initialized;
+}
+
+
+template <int dim, typename Number, typename VectorizedArrayType>
+inline unsigned int
+MatrixFree<dim, Number, VectorizedArrayType>::get_level_mg_handler() const
+{
+  return level_mg_handler;
 }
 
 
--- a/include/deal.II/matrix_free/matrix_free.templates.h
+++ b/include/deal.II/matrix_free/matrix_free.templates.h
@@ -58,6 +58,7 @@ MatrixFree<dim, Number, VectorizedArrayType>::MatrixFree()
   : Subscriptor()
   , indices_are_initialized(false)
   , mapping_is_initialized(false)
+  , level_mg_handler(numbers::invalid_unsigned_int)
 {}
 
 
@@ -249,6 +250,7 @@ MatrixFree<dim, Number, VectorizedArrayType>::copy_from(
   task_info                 = v.task_info;
   indices_are_initialized   = v.indices_are_initialized;
   mapping_is_initialized    = v.mapping_is_initialized;
+  level_mg_handler          = v.level_mg_handler;
 }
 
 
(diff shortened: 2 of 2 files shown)


tjhei:
I find the naming unfortunate and would a function like this to return a "handler". Maybe ``get_multigrid_level()`` or something?

peterrum:
@tjhei Objection noted.

tjhei:
> @tjhei Objection noted.

Are you okay with changing it? I am happy to make a PR.

peterrum:
@tjhei  I agree with changing the method name. But personally, I am not that happy with `*_multigrid_*`. I regard it rather as the refinement level (or active level) within the triangulation. I try to distinguish between these two. What about simply `get_level()`?

tjhei:
> What about simply `get_level()`?

I thought you can either iterate over active cells (level is set to -1) or over a specific multigrid level. Given that information "multigrid level" seems sensible to me. But either option is better than the current choice. ;-)

peterrum:
>But either option is better than the current choice. ;-)

I agree with that. 

The documentation of `MatirxFree::AdditionalData::level_mg_handler` states the following (see also: https://www.dealii.org/developer/doxygen/deal.II/structMatrixFree_1_1AdditionalData.html#a1f2e3f4876fd66850ca89950674b21f6):


> This option can be used to define whether we work on a certain level of the mesh, and not the active cells. If set to invalid_unsigned_int (which is the default value), the active cells are gone through, otherwise the level given by this parameter. Note that if you specify to work on a level, its dofs must be distributed by using dof_handler.distribute_mg_dofs(fe);.

So it kinda sounds like `get_mesh_level` ;) Let's get a third opinion by @kronbichler. I am fine with all names...

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1601,6 +1601,12 @@ class MatrixFree : public Subscriptor
   bool
   mapping_initialized() const;
 
+  /**
+   * Return the level of the mesh to be worked on.
+   */
+  unsigned int
+  get_level_mg_handler() const;
```
bangerth: I'll agree with @tjhei that that is an unfortunate naming -- the function doesn't return a handler, but a level.

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1928,6 +1934,11 @@ class MatrixFree : public Subscriptor
    */
   mutable std::list<std::pair<bool, AlignedVector<Number>>>
     scratch_pad_non_threadsafe;
+
+  /**
+   * Stored the level of the mesh to be worked on.
+   */
+  unsigned int level_mg_handler;
```
bangerth: And the same is true here -- maybe `mg_dof_handler_level` or `mg_level` would have been easier to understand?

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1928,6 +1934,11 @@ class MatrixFree : public Subscriptor
    */
   mutable std::list<std::pair<bool, AlignedVector<Number>>>
     scratch_pad_non_threadsafe;
+
+  /**
+   * Stored the level of the mesh to be worked on.
```
bangerth: Stored -> Store

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -265,6 +267,9 @@ MatrixFree<dim, Number, VectorizedArrayType>::internal_reinit(
   const typename MatrixFree<dim, Number, VectorizedArrayType>::AdditionalData
     &additional_data)
 {
+  // Store the level of the mesh to be worked on.
+  this->level_mg_handler = additional_data.level_mg_handler;
```
bangerth: This suggests that the name is already awkward in the `AdditionalData` structure.

kronbichler: Indeed it is - the reason for the name is that, back in 2011 when this code was written, we had a separate `MGDoFHandler` object, so the name was at least tangentially relevant because it referred to the case of a level when we had an `MGDoFHandler` type. But yes, it is a bit confusion; we could select a new one and deprecate this one, but as all interface breaks this will incur some noise for the applications.

peterrum: @kronbichler I do not think changing the name would be a problem. We create some form of alias and deprecate this... But do you agree that the name of this variable and the method should be called consistently. If yes, what name do you prefer?

include/deal.II/matrix_free/matrix_free.h
kronbichler: I like `mg_level` as the variable name and `get_mg_level()` for the access function; it is good to imply that the `level` relates to multigrid functionality and the prefix `mg` in line with most of our other names elsewhere, like `DoFHandler::distribute_mg_dofs()`, `DoHandler::locally_owned_mg_dofs()`. On the other hand, we use the `level` word in a few spots as well, like `DoFTools::extract_level_dofs()` or `DoFTools::extract_locally_relevant_level_dofs()` (rather than `..._mg_dofs`), so in that sense this function would contain both parts and use a word that we have not used in the library before (I grepped for `mg_level` in `include` and `source` and only found the member variable `mg_levels` and some header file names). The name is better than `get_level()` alone given its ambiguity, and better than `get_multigrid_level()`, as we hardly spell out the `multigrid` word for level information.

PR #8916, opened 2019-10-16
Extend GridGenerator::torus<3,3>() so that one can generate a torus segment

nfehn:
In the current implementation, the torus is closed (angle phi = 2 * pi). It can be useful to be able to generate an open torus with an angle of 0 < phi <= 2 * pi, which is addressed in this Pull Request. One application of this new functionality would be to simulate only on a part of the torus to reduce computational costs (since symmetries can be exploited). Another use case would be to glue together several torus segments to form more complex pipe systems (although this has currently some other limitations discussed elsewhere).

To implement this new functionality, the existing ``GridGenerator::torus<3,3>()`` function is extended by an additional parameter with default argument, and some tricky modifications need to be done in the code to obtain the correct cell connectivity and manifolds for a full and an open torus, respectively.

A new test case has been added to verify correctness of the new implementation. This is done by comparing the maximum aspect ratio (implemented recently in https://github.com/dealii/dealii/pull/8894) of a full torus to an open torus (which should be the same if the number of elements per arc length is the same). The idea behind is that if the manifolds are messed up in the open torus case, this will show up in the aspect ratio, while at the same time tests based on the aspect ratio are attractive as the result is reduced to one double value and one does not produce lengthy output files.

### DIFF
--- a/include/deal.II/grid/grid_generator.h
+++ b/include/deal.II/grid/grid_generator.h
@@ -1188,6 +1188,13 @@ namespace GridGenerator
    * @param n_cells_toroidal Optional argument to set the number of cell
    * layers in toroidal direction. The default is 6 cell layers.
    *
+   * @param phi Optional argument to generate an open torus with angle
+   * $0 < \varphi <= 2 \pi$. The default value is $2 \pi$,
+   * in which case a closed torus is generated. If the torus is open,
+   * the torus is cut at two planes perpendicular to the torus centerline.
+   * The center of these two planes are located at $(x_1, y_1, z_1) = (R, 0, 0)$
+   * and $(x_2, y_2, z_2) = (R \cos(\varphi), 0, R \sin(\varphi))$.
+   *
    * @note Implemented for Triangulation<2,3> and Triangulation<3,3>.
    */
   template <int dim, int spacedim>
@@ -1195,7 +1202,8 @@ namespace GridGenerator
   torus(Triangulation<dim, spacedim> &tria,
         const double                  R,
         const double                  r,
-        const unsigned int            n_cells_toroidal = 6);
+        const unsigned int            n_cells_toroidal = 6,
+        const double                  phi              = 2.0 * numbers::PI);
 
   /**
    * This function produces a square in the <i>xy</i>-plane with a cylindrical
--- a/source/grid/grid_generator.cc
+++ b/source/grid/grid_generator.cc
@@ -592,7 +592,8 @@ namespace GridGenerator
   void torus<2, 3>(Triangulation<2, 3> &tria,
                    const double         R,
                    const double         r,
-                   const unsigned int)
+                   const unsigned int,
+                   const double)
   {
     Assert(R > r,
            ExcMessage("Outer radius R must be greater than the inner "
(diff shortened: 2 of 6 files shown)


source/grid/grid_generator.cc
```diff
@@ -743,11 +745,16 @@ namespace GridGenerator
...
+    AssertThrow(phi > 0.0 && phi < 2.0 * numbers::PI + 1.0e-15,
+                ExcMessage("Invalid angle phi specified."));
 
     // the first 8 vertices are in the x-y-plane
-    Point<3> const        p = Point<3>(R, 0.0, 0.0);
-    double const          a = 1. / (1 + std::sqrt(2.0));
-    std::vector<Point<3>> vertices(8 * n_cells_toroidal);
+    Point<3> const p                = Point<3>(R, 0.0, 0.0);
+    double const   a                = 1. / (1 + std::sqrt(2.0));
+    unsigned int   additional_layer = 0; // torus is closed (angle of 2*pi)
```
drwells:
```suggestion
    const unsigned int   additional_layer = phi < 2.0 * numbers::PI - 1.0e-15 ? 1 : 0;
```

drwells: Couple this with removal of the next two lines.

source/grid/grid_generator.cc
```diff
@@ -777,13 +784,15 @@ namespace GridGenerator
       {
         for (unsigned int j = 0; j < 2; ++j)
           {
-            unsigned int offset = (8 * (c + j)) % (8 * n_cells_toroidal);
+            unsigned int offset =
```
drwells:
```suggestion
            const unsigned int offset =
```

include/deal.II/grid/grid_generator.h
```diff
@@ -1188,14 +1188,22 @@ namespace GridGenerator
    * @param n_cells_toroidal Optional argument to set the number of cell
    * layers in toroidal direction. The default is 6 cell layers.
    *
+   * @param phi Optional argument to generate an open torus with angle
+   * $0 < \varphi <= 2 \pi$. The default value is $2 \pi$,
+   * in which case a closed torus is generated. If the torus is open,
+   * the torus is cut at two planes perpendicular to the torus centerline.
+   * The center of these two planes are located at $(x_1, y_1, z_1) = (R, 0, 0)$
+   * and $(x_2, y_2, z_2) = (R \cos(\varphi), 0, R \sin(\varphi))$.
```
drwells: We should document the boundary ids assigned to the three different boundary parts.

nfehn: The boundary IDs should all be 0 (which is the deal.II default value?), so I would not document

drwells: You're right: I misread the code that checks face numbers as checking boundary ids, so I thought the new boundaries had ids 4 and 5.

source/grid/grid_generator.cc
nfehn: done in https://github.com/dealii/dealii/pull/8916/commits/df2226367fedbca8a7dc1829f1db034db63b7554

source/grid/grid_generator.cc
nfehn: done in https://github.com/dealii/dealii/pull/8916/commits/df2226367fedbca8a7dc1829f1db034db63b7554

PR #8917, opened 2019-10-16
Use positive form for comments.

bangerth:
Negative comments are often difficult to read because they *refer* to something -- often a bug in a previous version. Rather try to describe what *is* supposed to happen.

Follow-up to #8895. @starki0815 -- FYI.


drwells:
As always, I admire your attention to detail.

bangerth:
It really speaks to the level of mental ability I had to muster this week in the midst of 8am-7pm meetings (including working lunches and dinners) reviewing the management structure of a $1.8B project...

PR #8924, opened 2019-10-18
Refactoring of Utilities::needed_digits()

nfehn:
Use functionality from standard library instead of naive implementation.

Call ``Utilities::needed_digits()`` in ``write_vtu_with_pvtu_record()``, see PR #8904.

Add more tests to test corner cases where rounding is essential.


source/base/data_out_base.cc
```diff
@@ -7560,8 +7560,11 @@ DataOutInterface<dim, spacedim>::write_vtu_with_pvtu_record(
   const unsigned int n_files_written =
     (n_groups == 0 || n_groups > n_ranks) ? n_ranks : n_groups;
 
+  AssertThrow(n_files_written >= 1, ExcMessage("Logical error."));
```
drwells: If this should be impossible to reach we usually write
```cpp
Assert(n_files_written >= 1, ExcInternalError());
```
Is that the case here?

source/base/utilities.cc
```diff
@@ -466,20 +466,13 @@ namespace Utilities
   unsigned int
   needed_digits(const unsigned int max_number)
   {
-    if (max_number < 10)
+    if (max_number > 0)
+      return static_cast<int>(
+        std::ceil(std::log10(std::fabs(max_number + 0.1))));
+    else
       return 1;
```
drwells: We can get rid of this `else` clause. I think clang-tidy will complain about it so this will placate one of the CI builds.

source/base/data_out_base.cc
nfehn: yes, I will change this

nfehn: done in https://github.com/dealii/dealii/pull/8924/commits/20f131f20cfe477e162501d34fece92307858667

source/base/utilities.cc
nfehn: done in https://github.com/dealii/dealii/pull/8924/commits/20f131f20cfe477e162501d34fece92307858667

doc/news/changes/minor/20191018NiklasFehn
```diff
@@ -0,0 +1,4 @@
+Refactor: improve implementation of Utilities::needed_digits() by using
+functionality from the standard library.
+<br>
+(Niklas Fehn, 2019/10/18)
```
bangerth: It doesn't hurt, but in general we only add changelog entries for things that change user-visible behavior. This one doesn't.

PR #8925, opened 2019-10-18
Document usage of empty copiers in WorkStream

masterleinad:
Motivated by https://github.com/dealii/dealii/pull/8891#discussion_r332079352, document how to best select an empty copier object in `WorkStream::run()`.


tjhei:
no need to run tests as this is documentation only...

PR #8927, opened 2019-10-18
new function Utilities::cutoff_digits()

nfehn:
This function can be used to cutoff floating point numbers after a specified number of digits of accuracy.

Background: This function originates from adaptive time stepping, where the time step size adjusted dynamically from one time step to the next depends on the solution field. When computing on a large number of cores, rounding errors cause a slightly different time step size, which means that in the next time step the solution is evaluated at a different time, again causing a different time step size. This way, rounding errors can accumulate, rendering testing on the application level by comparing numbers for serial and parallel computations very difficult. The solution is to cutoff the time step size after a certain number of digits, so that errors can not accumulate.

A new test has been added to verify the correctness of this function for double/float, positive/negative numbers, positive/negative exponents.

### DIFF
--- a/include/deal.II/base/utilities.h
+++ b/include/deal.II/base/utilities.h
@@ -174,9 +174,25 @@ namespace Utilities
   /**
    * Determine how many digits are needed to represent numbers at most as
    * large as the given number.
+   *
+   * @author Niklas Fehn, 2019
    */
   unsigned int
   needed_digits(const unsigned int max_number);
+
+  /**
+   * This function allows to cut off a floating point number @p number
+   * after @p n_digits of accuracy, i.e., after @p n_digits decimal places
+   * in scientific floating point notation. When interpreted as rounding
+   * operation, this function reduces the absolute value of a floating point
+   * number and always rounds towards zero, since decimal places are simply
+   * cut off.
+   *
+   * @author Niklas Fehn, 2019
+   */
+  template <typename Number>
+  Number
+  truncate_to_n_digits(const Number number, const unsigned int n_digits);
 
   /**
    * Given a string, convert it to an integer. Throw an assertion if that is
--- a/source/base/utilities.cc
+++ b/source/base/utilities.cc
@@ -484,6 +484,34 @@ namespace Utilities
 
 
 
+  template <typename Number>
+  Number
+  truncate_to_n_digits(const Number number, const unsigned int n_digits)
+  {
+    AssertThrow(n_digits >= 1, ExcMessage("invalid parameter."));
+
+    if (!(std::fabs(number) > std::numeric_limits<Number>::min()))
+      return number;
+
+    const int order =
+      static_cast<int>(std::floor(std::log10(std::fabs(number))));
+
+    const int shift = -order + static_cast<int>(n_digits) - 1;
+
+    Assert(shift <= static_cast<int>(std::floor(
+                      std::log10(std::numeric_limits<Number>::max()))),
+           ExcMessage(
+             "Overflow. Use a smaller value for n_digits and/or make sure "
+             "that the absolute value of 'number' does not become too small."));
+
+    const Number factor = std::pow(10.0, static_cast<Number>(shift));
+
+    const Number number_cutoff = std::trunc(number * factor) / factor;
+
+    return number_cutoff;
+  }
+
+
   int
   string_to_int(const std::string &s_)
   {
@@ -1227,6 +1255,11 @@ namespace Utilities
   to_string<double>(double, unsigned int);
   template std::string
   to_string<long double>(long double, unsigned int);
+
+  template double
+  truncate_to_n_digits(const double, const unsigned int);
+  template float
+  truncate_to_n_digits(const float, const unsigned int);
 
   template std::vector<std::array<std::uint64_t, 1>>
   inverse_Hilbert_space_filling_curve<1, double>(
--- a/tests/base/utilities_20.output
+++ b/tests/base/utilities_20.output
@@ -1,0 +1,21 @@
+
+DEAL::
+DEAL::number = 1.234568e+05
+DEAL::number_1digit = 1.000000e+05
+DEAL::number_2digit = 1.200000e+05
+DEAL::number_3digit = 1.230000e+05
+DEAL::
+DEAL::number = 1.000000e+00
+DEAL::number_1digit = 9.000000e-01
+DEAL::number_2digit = 9.900000e-01
+DEAL::number_3digit = 9.990000e-01
+DEAL::
+DEAL::number = 0.000000e+00
+DEAL::number_1digit = 0.000000e+00
+DEAL::number_2digit = 0.000000e+00
+DEAL::number_3digit = 0.000000e+00
+DEAL::
+DEAL::number = -9.876543e-06
+DEAL::number_1digit = -9.000000e-06
+DEAL::number_2digit = -9.800000e-06
+DEAL::number_3digit = -9.870000e-06
--- a/doc/news/changes/minor/20191018NiklasFehn-2
+++ b/doc/news/changes/minor/20191018NiklasFehn-2
@@ -1,0 +1,5 @@
+New: new function Utilities::truncate_to_n_digits() that can be used to
+cutoff floating point numbers after a specified number of digits of
+accuracy (decimal places) in scientific floating point notation.
+<br>
+(Niklas Fehn, 2019/10/18)
--- a/tests/base/utilities_20.cc
+++ b/tests/base/utilities_20.cc
@@ -1,0 +1,92 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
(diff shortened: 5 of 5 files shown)


include/deal.II/base/utilities.h
```diff
@@ -178,6 +178,14 @@ namespace Utilities
   unsigned int
   needed_digits(const unsigned int max_number);
 
+  /**
+   * This function allows to cutoff a floating point number @p number
+   * after @p n_digits of accuracy.
```
kronbichler: Can you please add a small example of what happens in that case. Furthermore, it would be good to state what happens for negative numbers - are they always truncated to zero or rounded down?

source/base/utilities.cc
```diff
@@ -484,6 +484,22 @@ namespace Utilities
...
 
+  template <typename Number>
+  Number
+  cutoff_digits(const Number number, const unsigned int n_digits)
+  {
+    AssertThrow(n_digits >= 1, ExcMessage("invalid parameter."));
+
+    const int order = std::floor(std::log10(std::fabs(number)));
+
+    const Number factor = std::pow(10.0, (Number)(-order + int(n_digits) - 1));
```
kronbichler: We typically spell out double <-> integer conversions like this:
```suggestion
    const Number factor = std::pow(10.0, (Number)(-order + static_casts<int>(n_digits) - 1));
```

source/base/utilities.cc
```diff
@@ -484,6 +484,22 @@ namespace Utilities
...
+
+    const Number number_cutoff = int(number * factor) / factor;
```
kronbichler: Same here.

masterleinad:
```
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-8927/source/base/utilities.cc:493:23: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
    const int order = std::floor(std::log10(std::fabs(number)));
```

source/base/utilities.cc
tjhei:
```suggestion
    const Number number_cutoff = std::floor(number * factor) / factor;
```

include/deal.II/base/utilities.h
```diff
@@ -178,6 +178,14 @@ namespace Utilities
   unsigned int
   needed_digits(const unsigned int max_number);
 
+  /**
+   * This function allows to cutoff a floating point number @p number
```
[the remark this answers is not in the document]
nfehn: Not sure if I got the point. Maybe there is a misunderstanding because you think of a fixed floating point notation, e.g., 0.00367. But I am using scientific notation, 3.67e-3, where there is no difference between decimal places before or after the separator. That‘s why I write digits of accuracy. 7 is the third digit of accuracy, not 3. So the „relative error“ caused by this function is bounded by n_digits in my implementation. Cutting off after some decimal places in fixed notation could lead to a relative error that tends to infinity, depending on the order of magnitude of the number. In this sense, the implementation is not that naive.

source/base/utilities.cc
nfehn: Are you aware of the fact that std::floor is not the same as int()?

std::floor(-5.7)=-6
int(-5.7)=-5

Run the test to see which influence your change will have.

nfehn: Has been changed to deal.ii standard

include/deal.II/base/utilities.h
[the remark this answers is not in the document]
tjhei: > Maybe there is a misunderstanding because you think of a fixed floating point notation,

Alright, can you please update the documentation of the function to help clear up this misunderstanding? I am sure I won't be the only one with this question. ;-)

source/base/utilities.cc
tjhei: OK. How about std::trunc()?

nfehn: Should be the same. Why do you want to change? The current version is already correct

tjhei: 1. We do not use ``int(x)`` as it is a c-style cast [1].
2. I don't read this code as you performing a cast here. Instead, logically this is a rounding/truncating operation. Being explicit about it is good.
3. It is much more likely for a reader to know the way ``trunc`` rounds compared to a cast.

[1] https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast

source/base/utilities.cc
```diff
@@ -484,6 +484,24 @@ namespace Utilities
...
+  Number
+  cutoff_digits(const Number number, const unsigned int n_digits)
+  {
+    AssertThrow(n_digits >= 1, ExcMessage("invalid parameter."));
+
+    const int order =
+      static_cast<int>(std::floor(std::log10(std::fabs(number))));
+
+    const Number factor =
+      std::pow(10.0, (Number)(-order + static_cast<int>(n_digits) - 1));
```
masterleinad: ...and here is another C-style cast.

include/deal.II/base/utilities.h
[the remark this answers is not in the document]
nfehn: the documentation has already been improved

include/deal.II/base/utilities.h
```diff
@@ -178,6 +178,14 @@ namespace Utilities
...
+   * after @p n_digits of accuracy.
+   */
+  template <typename Number>
+  Number
+  cutoff_digits(const Number number, const unsigned int n_digits);
```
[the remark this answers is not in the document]
nfehn: done in https://github.com/dealii/dealii/pull/8927/commits/58465502f0cd24ac8a500a369d68987fa3c39171

source/base/utilities.cc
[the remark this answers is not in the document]
nfehn: done in https://github.com/dealii/dealii/pull/8927/commits/58465502f0cd24ac8a500a369d68987fa3c39171

One can construct examples where the code now runs into this assert, but I think this has to be dealt with (for example catched) in an application code, e.g., where one specified a value below which the number is simply set to zero or to a small value (e.g. 1e-10), depending on how this function is used.

tests/base/utilities_20.output
```diff
@@ -0,0 +1,16 @@
...
+DEAL::
+DEAL::number = 1.000000e+00
+DEAL::number_1digit = 9.000000e-01
+DEAL::number_2digit = 9.900000e-01
+DEAL::number_3digit = 9.990000e-01
+DEAL::
+DEAL::number = -9.876543e-06
+DEAL::number_1digit = -9.000000e-06
+DEAL::number_2digit = -9.800000e-06
+DEAL::number_3digit = -9.870000e-06
```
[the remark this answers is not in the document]
nfehn: done in https://github.com/dealii/dealii/pull/8927/commits/58465502f0cd24ac8a500a369d68987fa3c39171

source/base/utilities.cc
```diff
@@ -484,6 +484,37 @@ namespace Utilities
...
+    const int max_order = static_cast<int>(
+      std::floor(std::log10(std::numeric_limits<Number>::max())));
+
+    const int shift = -order + static_cast<int>(n_digits) - 1;
+
+    AssertThrow(
+      shift <= max_order,
+      ExcMessage(
+        "Overflow. Use a smaller value for n_digits and/or make sure "
+        "that the absolute value of number does not become too small."));
```
bangerth: Make clear that you're talking about a particular variable:
```suggestion
        "that the absolute value of 'number' does not become too small."));
```

source/base/utilities.cc
```diff
@@ -484,6 +484,37 @@ namespace Utilities
...
+    const int order =
+      static_cast<int>(std::floor(std::log10(std::fabs(number))));
+
+    const int max_order = static_cast<int>(
+      std::floor(std::log10(std::numeric_limits<Number>::max())));
+
+    const int shift = -order + static_cast<int>(n_digits) - 1;
+
+    AssertThrow(
+      shift <= max_order,
```
bangerth: You'll get a warning about an unused variable `max_order` in release mode because the variable is only used inside the assertion. Just inline the variable's value into this line.

nfehn: sorry, I don't know what inlining means in this context. Is AssertThrow correct?

nfehn: How about https://github.com/dealii/dealii/pull/8927/commits/b6f371ffde891122e9eb3baf9ef367ba43224272 ?

peterrum:
```suggestion
    AssertThrow(
      shift <= static_cast<int>(
      std::floor(std::log10(std::numeric_limits<Number>::max()))),
```

bangerth: Instead of creating a named variable, just take the expression you use to define `max_order` and put it into the assertion here. Then there will simply be no variable called `max_order` at all.

source/base/utilities.cc
```diff
@@ -484,6 +484,35 @@ namespace Utilities
...
+      return number;
+
+    const int order =
+      static_cast<int>(std::floor(std::log10(std::fabs(number))));
+
+    const int shift = -order + static_cast<int>(n_digits) - 1;
+
+    AssertThrow(
+      shift <= static_cast<int>(
+                 std::floor(std::log10(std::numeric_limits<Number>::max()))),
```
bangerth: Ah yes, just like this!

source/base/utilities.cc
masterleinad: `AssertThrow` is also visible in Release mode. If you were using `Assert` instead we would need to be more careful.

bangerth: Oh, correct. But we should be using `Assert` here.

PR #8930, opened 2019-10-18
Fix ConsensusAlgorithm_NBX

### LINKED ISSUE #8929: SegFault in ConsensusAlgorithmProcess::process_request
I am seeing the following crash (intel 19.0.5 with 1792 MPI ranks and about 2 million DoFs):
```
dealii::Utilities::MPI::ComputeIndexOwner::ConsensusAlgorithmProcess::process_request
[c137-061:30925:0:30925] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x3f116d0)
==== backtrace (tid:  30925) ====
 0 0x000000000a29ccf0 dealii::Utilities::MPI::ComputeIndexOwner::ConsensusAlgorithmProcess::process_request()  ???:0
 1 0x000000000a297d04 dealii::Utilities::MPI::ConsensusAlgorithm_NBX<unsigned int, unsigned int>::process_requests()  ???:0
 2 0x000000000a297259 dealii::Utilities::MPI::ConsensusAlgorithm_NBX<unsigned int, unsigned int>::run()  ???:0
 3 0x000000000a294fb6 dealii::Utilities::MPI::compute_index_owner()  ???:0
 4 0x000000000b7096b2 dealii::internal::MGTransfer::fill_copy_indices<3, 3>()  ???:0
 5 0x000000000b68003c dealii::(anonymous namespace)::fill_internal<3, 3, double>()  mg_level_global_transfer.cc:0
 6 0x000000000b6a3130 dealii::MGLevelGlobalTransfer<dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> >::fill_and_communicate_copy_indices<3, 3>()  ???:0
 7 0x000000000b966f23 dealii::MGTransferMatrixFree<3, double>::build()  ???:0
 8 0x00000000012c6f90 aspect::StokesMatrixFreeHandler<3>::setup_dofs()  ???:0
 9 0x0000000001095022 aspect::Simulator<3>::setup_dofs()  ???:0
10 0x00000000010a5fd9 aspect::Simulator<3>::refine_mesh()  ???:0
11 0x000000000111b4b6 aspect::Simulator<3>::maybe_do_initial_refinement()  ???:0
12 0x000000000108fba5 aspect::Simulator<3>::run()  ???:0
13 0x0000000000ac3060 run_simulator<3>()  main.cc:0
14 0x0000000000aaa3ee main()  ???:0
15 0x0000000000022495 __libc_start_main()  ???:0
```
I find it scary to ``emplace_back()`` for ``request_buffers`` and ``request_requests`` followed by an ``MPI_Isend`` because this could cause an allocation invalidating the memory, or not?
Before I dig much deeper into the code, @peterrum @kronbichler do you have an idea what the problem might be?

FYI: @tcclevenger

peterrum:
part of #8929.

@tjhei could you try this patch out? I have my doubt that it will help. I would have imagined that `std::vector` on itself is something like a `std::shared_ptr`: the actual data is on the heap. When the size of the outer vector is resized the data of the inner vectors should not be moved...

FYI: @kronbichler @tcclevenger

### DIFF
--- a/include/deal.II/base/mpi.h
+++ b/include/deal.II/base/mpi.h
@@ -941,12 +941,12 @@ namespace Utilities
       /**
        * Buffers for sending answers to requests.
        */
-      std::vector<std::vector<T2>> request_buffers;
+      std::vector<std::unique_ptr<std::vector<T2>>> request_buffers;
 
       /**
        * Requests for sending answers to requests.
        */
-      std::vector<std::shared_ptr<MPI_Request>> request_requests;
+      std::vector<std::unique_ptr<MPI_Request>> request_requests;
 
       // request for barrier
       MPI_Request barrier_request;
--- a/source/base/mpi.cc
+++ b/source/base/mpi.cc
@@ -1011,11 +1011,12 @@ namespace Utilities
           AssertThrowMPI(ierr);
 
           // allocate memory for answer message
-          request_buffers.emplace_back();
-          request_requests.emplace_back(new MPI_Request);
+          request_buffers.emplace_back(
+            std_cxx14::make_unique<std::vector<T2>>());
+          request_requests.emplace_back(std_cxx14::make_unique<MPI_Request>());
 
           // process request
-          auto &request_buffer = request_buffers.back();
+          auto &request_buffer = *request_buffers.back();
           this->process.process_request(other_rank,
                                         buffer_recv,
                                         request_buffer);


tjhei:
> I would have imagined that `std::vector` on itself is something like a `std::shared_ptr`: the actual data is on the heap

That is correct. 

> When the size of the outer vector is resized the data of the inner vectors should not be moved...

Likely. This assumes that the outer vector uses move assignment and the move assignment operator for the inner vector does not copy. Maybe that is not the case here?!

tjhei:
> could you try this patch out?

it crashes in a different place (earlier):
```
 0 0x000000000a4f52bd dealii::Utilities::MPI::internal::ComputeIndexOwner::ConsensusAlgorithmPayload::get_requesters()  ???:0
 1 0x000000000a4f18f6 dealii::Utilities::MPI::Partitioner::set_ghost_indices()  ???:0
 2 0x000000000a4f0651 dealii::Utilities::MPI::Partitioner::Partitioner()  ???:0
 3 0x00000000094ad334 dealii::internal::DoFHandlerImplementation::Policy::ParallelDistributed<dealii::DoFHandler<3, 3> >::renumber_dofs()  ???:0
 4 0x00000000090212a4 dealii::DoFHandler<3, 3>::renumber_dofs()  ???:0
 5 0x0000000009526734 dealii::DoFRenumbering::hierarchical<dealii::DoFHandler<3, 3> >()  ???:0
 6 0x00000000012bc6b8 aspect::StokesMatrixFreeHandler<3>::setup_dofs()  ???:0
 7 0x000000000108e632 aspect::Simulator<3>::setup_dofs()  ???:0
 8 0x00000000010884d3 aspect::Simulator<3>::run()  ???:0
```

I will try to figure out how to reproduce it on a smaller run...

kronbichler:
> I would have imagined that std::vector on itself is something like a std::shared_ptr: the actual data is on the heap. When the size of the outer vector is resized the data of the inner vectors should not be moved...

At least prior to C++11 the vector did perform a copy (I had checked many years ago). However, I would also have guessed that the newer standard could replace this by a move.

masterleinad:
> At least prior to C++11 the vector did perform a copy (I had checked many years ago). However, I would also have guessed that the newer standard could replace this by a move.

If the move constructor is `noexcept` modern compiler use it, otherwise the copy constructor is invoked (https://godbolt.org/z/FBwR6g). I don't think that this behavior is hard-coded in the standard, though.

peterrum:
> I will try to figure out how to reproduce it on a smaller run...

Thanks @tjhei!

bangerth:
If you wonder whether the previous behavior was well defined, then it's probably worth replacing the code by something that's unambiguous -- regardless of whether the previous code *could* or *might* or *should* have worked :-)

tjhei:
I can confirm that this PR eliminates one of the crashes reported in #8929 but it does not fix the issue completely.

include/deal.II/base/mpi.h
```diff
@@ -941,7 +941,7 @@ namespace Utilities
       /**
        * Buffers for sending answers to requests.
        */
-      std::vector<std::vector<T2>> request_buffers;
+      std::vector<std::shared_ptr<std::vector<T2>>> request_buffers;
```
masterleinad: Any reason not using `std::unique_ptr` here?

source/base/mpi.cc
```diff
@@ -1011,11 +1011,11 @@ namespace Utilities
           AssertThrowMPI(ierr);
 
           // allocate memory for answer message
-          request_buffers.emplace_back();
+          request_buffers.emplace_back(new std::vector<T2>);
```
bangerth: We tried to avoid using bare `new`. Use `std_cxx14::make_unique`.

tjhei:
Let me check if unique_ptr also works for me.

peterrum:
@tjhei I wanted to ping you as soon as the test have passed. They have not finished yet but the MPI ones have all passed...

PR #8940, opened 2019-10-21
Improve and fix some doxygen things in step-3.

drwells:
- use `CellAccessor::face_iterators()`
- no space after function names
- add some more doxygen links
- fix a missing `@endcode`


tjhei:
:+1: doxygen changes only, so no need to run tests...

PR #8943, opened 2019-10-22
Fix coarse solver documentation in step-50

dangars:
Step-50 uses for the coarse solver `MGCoarseGridIterativeSolver`, not `MGCoarseGridHouseholder`. Add discussion about when it would be appropriated to use `MGCoarseGridHouseholder`.

I guess that in the past step-50 used `MGCoarseGridHouseholder`. Although, I did a little bit of git archaeology and I couldn't find when step-50 used `MGCoarseGridHouseholder`.


tjhei:
Hi @dangars , thanks but the text in step-50 is not written yet (this is copied from step-16 or something). I am working with @tcclevenger on the text and modifications including the coarse solver. Hopefully, this is done soon.

PR #7180, opened 2018-09-13
Implement TpetraWrappers::Vector

masterleinad:
This PR implements a wrapper class for `Tpetra::Vector` derived from `VectorSpaceVector` similar to `EpetraWrappers::Vector`.
The current version just sets all template arguments to there default value.
Currently, there is a discussion at https://github.com/trilinos/Trilinos/wiki/Tpetra-Information-Page and a corresponding PR https://github.com/trilinos/Trilinos/pull/3398 that removes some of the template arguments. We need to have a look if this new version will eventually replace the current `Tpetra` interfaces. It should be manageable to keep this branch up-to-date anyway, though.

This PR needs testing for the case when we use `CUDA` device memory instead of CPU memory.

With this PR, we also have to be careful with respect to #6856 and #6423.

PR #8945, opened 2019-10-22
Remove unnecessary trilinos options

dangars:
The options `-DTpetra_INST_COMPLEX_DOUBLE=ON` and `-DTpetra_INST_COMPLEX_FLOAT=ON`
are no longer necessary. `-DTrilinos_ENABLE_EXPLICIT_INSTANTIATION=ON` can be
used to enable explicit instantiations.

I built without `-DTpetra_INST_COMPLEX_DOUBLE=ON`, `-DTpetra_INST_COMPLEX_FLOAT=ON`,  `-DTrilinos_ENABLE_EXPLICIT_INSTANTIATION=ON` and everything works perfect :)

When I tried to use `-DTpetra_INST_COMPLEX_DOUBLE=ON` and `-DTpetra_INST_COMPLEX_FLOAT=ON` I got the following errors:

```
CMake Warning at packages/tpetra/CMakeLists.txt:864 (MESSAGE):
  Tpetra_INST_COMPLEX_DOUBLE=ON, but Trilinos_ENABLE_COMPLEX_DOUBLE=OFF.
  This will work, but it will make builds more expensive, because it turns
  off explicit instantiation for complex<double> in KokkosKernels.  If you
  want to use Scalar=complex<double> in Tpetra, just set
  Trilinos_ENABLE_COMPLEX_DOUBLE=ON.  You do not need to set
  Tpetra_INST_COMPLEX_DOUBLE explicitly any more.


CMake Error at packages/tpetra/CMakeLists.txt:884 (MESSAGE):
  Tpetra: Tpetra_INST_COMPLEX_DOUBLE is ON (meaning that you want to want to
  instantiate and/or test Tpetra classes with Scalar = std::complex<double>),
  but Teuchos_ENABLE_COMPLEX is OFF.  Please set
  Teuchos_ENABLE_COMPLEX:BOOL=ON, reconfigure, and rebuild.
```

I use Trilinos 12.14

### DIFF
--- a/doc/external-libs/trilinos.html
+++ b/doc/external-libs/trilinos.html
@@ -94,8 +94,8 @@
     -DTrilinos_ENABLE_ML=ON                          \
     -DTrilinos_ENABLE_ROL=ON                         \
     -DTrilinos_ENABLE_Tpetra=ON                      \
-    -DTpetra_INST_COMPLEX_DOUBLE=ON                  \
-    -DTpetra_INST_COMPLEX_FLOAT=ON                   \
+    -DTrilinos_ENABLE_COMPLEX_DOUBLE=ON              \
+    -DTrilinos_ENABLE_COMPLEX_FLOAT=ON               \
     -DTrilinos_ENABLE_Zoltan=ON                      \
     -DTrilinos_VERBOSE_CONFIGURE=OFF                 \
     -DTPL_ENABLE_MPI=ON                              \


masterleinad:
IIRC, we were running into trouble without these options for previous Trilinos releases. Of course, you can use ETI instead but that increases the build time for Trilinos tremendously.

dangars:
> IIRC, we were running into trouble without these options for previous Trilinos releases. Of course, you can use ETI instead but that increases the build time for Trilinos tremendously.

Which were the problems? I didn't use those options to compile Trilinos, I can check if the problems persist without those options in Trilinos 12.14.

What would you propose?
* Remove the options and add the following sentence: Old versions of Trilinos might require the options `-DTpetra_INST_COMPLEX_DOUBLE=ON` and `-DTpetra_INST_COMPLEX_FLOAT=ON`.
* Keep the options untouched and add the following sentence: New versions of Trilinos don't require the options `-DTpetra_INST_COMPLEX_DOUBLE=ON` and `-DTpetra_INST_COMPLEX_FLOAT=ON`.
* Close this PR

masterleinad:
You can have a look at the discussion starting at https://github.com/dealii/dealii/pull/7180#issuecomment-423212096. There was a problem with `Trilinos 12.4.2`.

I would likely just add `Trilinos_ENABLE_COMPLEX_DOUBLE=ON` and `Trilinos_ENABLE_COMPLEX_FLOAT=ON` (I didn't test it). I am not sure if the `ETI` option already existed in version 12.4.2.

dangars:
I included those options. Is this ok? I built Trilinos with those options, then built deal.II, then run step-50 and everything is ok.

What is the ETI option?

masterleinad:
> What is the ETI option? ETI is equal to `-DTrilinos_ENABLE_EXPLICIT_INSTANTIATION=ON` ?

Yes, explicit template instantiation. Let's go with this version then.

dangars:
Sounds good!

PR #8946, opened 2019-10-22
Make VectorTools::interpolate() working in parallel

elauksap:
Make VectorTools::interpolate working in parallel + update documentation.

Co-authored-by: Ivan Fumagalli <ivan.fumagalli@polimi.it> @ifumagalli


masterleinad:
This seems to be a duplicate of #8912. Did you communicate?

elauksap:
> This seems to be a duplicate of #8912. Did you communicate?

You are right, I just noticed that pull request #8912 concerns and solves the same issue. Here @ifumagalli and I propose a different implementation based on a `std::vector<bool>`, which is less memory-consuming than a `std::vector<short unsigned int>` and fixes the `TODO`.

Also, the function documentation has been updated.

include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -602,53 +602,47 @@ namespace VectorTools
...
 
-    std::vector<short unsigned int> touch_count(
-      dof_2.n_dofs(), 0); // TODO: check on datatype... kinda strange (UK)
+    // Prevent multiple processing of the same dof.
+    std::vector<bool> touched_dofs(dof_2.n_dofs(), false);
+
     std::vector<types::global_dof_index> local_dof_indices(
       dof_2.get_fe().dofs_per_cell);
 
-    typename DoFHandler<dim, spacedim>::active_cell_iterator h =
```
masterleinad: Why do you remove the second template argument here and below?

include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -602,53 +602,47 @@ namespace VectorTools
...
+                // If local_dof_indices[j] has not been processed yet.
+                if (touched_dofs[local_dof_indices[j]] == false)
+                  {
+                    ::dealii::internal::ElementAccess<OutVector>::set(
+                      cell_data_2(j), local_dof_indices[j], data_2);
 
-        ::dealii::internal::ElementAccess<OutVector>::set(
-          val / static_cast<value_type>(touch_count[i]), i, data_2);
+                    touched_dofs[local_dof_indices[j]] = true;
+                  }
```
masterleinad: Why do you change the implementation to not use the average of the contributions anymore? Also, see https://github.com/dealii/dealii/pull/8912#issuecomment-544790302.

include/deal.II/numerics/vector_tools.templates.h
elauksap: The typos have been fixed.

include/deal.II/numerics/vector_tools.templates.h
elauksap: The average has been restored as well and made efficient for parallel computations.

elauksap:
> 
> 
> Thanks for also looking into this!
> 
> I don't see a reason to make changes to this function that are not backward-compatible, though. We should keep averaging the contributions from different cells. Then, we can save some MPI communication by looping over all the locally relevant cells and only set values for the locally owned dofs.
> 
> It would be great if both pull requests could be merged. In addition to the changes in the codebase, we need a changelog and a test (such as the one in #8912).

I implemented the changes requested and merged the test proposed in #8912!

dougshidong:
Thanks for changes.

bangerth:
For completeness, the reason to average was discussed here: https://github.com/dealii/dealii/pull/8912#issuecomment-545699245

tests/vector_tools/interpolate.cc
```diff
@@ -0,0 +1,282 @@
+#include <deal.II/base/utilities.h>
```
bangerth: can you add the usual copyright header to this test?

doc/news/changes/minor/20191022PasqualeClaudioAfrica
```diff
@@ -0,0 +1,3 @@
+Improved: VectorTools::interpolate() now works for a parallel::distributed::Triangulation
+<br>
+(Pasquale Claudio Africa, 2019/10/22)
```
bangerth: Since I think you cooperated on this patch with @dougshidong , you might want to list both of you here :-)

tests/vector_tools/interpolate.cc
```diff
@@ -0,0 +1,282 @@
...
+
+#include <deal.II/distributed/tria.h>
```
bangerth: We typically place a description of what the test does right after the copyright header and before the include files start. This way, one can figure out what a test does without having to read through a couple hundred lines of code :-)

tests/vector_tools/interpolate.cc
```diff
@@ -0,0 +1,282 @@
...
+{
+  using Vector = dealii::LinearAlgebra::distributed::Vector<double>;
+  // Create grid
+  dealii::parallel::distributed::Triangulation<dim> triangulation(
+    MPI_COMM_WORLD,
+    typename dealii::Triangulation<dim>::MeshSmoothing(
+      dealii::Triangulation<dim>::smoothing_on_refinement |
+      dealii::Triangulation<dim>::smoothing_on_coarsening));
+  const unsigned int n_1d_cells = 2;
+  dealii::GridGenerator::subdivided_hyper_cube(triangulation, n_1d_cells);
```
bangerth: Is the code here in essence duplicated from the 1d function, with the exception of using a `p::d::Triangulation`? If so, the easier approach might simply be to construct the triangulation in `main()` (non-distributed in 1d, distributed for 2d and 3d), and then pass it as an argument to this function. The body of the function would then be the same for 1d, 2d, and 3d.

tests/vector_tools/interpolate.output
```diff
@@ -0,0 +1,7 @@
+
```
masterleinad: Please also rename this file to `tests/vector_tools/interpolate.with_p4est=true.mpirun=1.output`. This test requires `MPI` and `p4est` and would fail if executed withput it.

tests/vector_tools/interpolate.cc
elauksap: Copyright added!

tests/vector_tools/interpolate.cc
elauksap: The duplicated code has been removed and the rest modified as suggested.

elauksap:
All the changes requested have been done!
Please find the updated test case and function implementation in the last 4 commits.

tests/vector_tools/interpolate.cc
elauksap: Description added.

doc/news/changes/minor/20191022PasqualeClaudioAfrica
elauksap: Added @dougshidong as an author.

tests/vector_tools/interpolate.output
elauksap: The file has been renamed!

elauksap:
All the changes requested have been done!
Please find the updated test case and function implementation in the last 5 commits.

include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -602,56 +602,76 @@ namespace VectorTools
...
-                std::numeric_limits<decltype(touch_count)::value_type>::max(),
-              ExcInternalError());
-            ++touch_count[local_dof_indices[j]];
+            cell_2->get_dof_indices(local_dof_indices);
+
+            // Distribute cell vector.
+            for (unsigned int j = 0; j < dof_2.get_fe().dofs_per_cell; ++j)
+              {
+                if (data_2.locally_owned_elements().is_element(
+                      local_dof_indices[j]))
```
masterleinad: I don't think that this is correct anymore. Since the locally owned elements are non-overlapping, you are missing contributions for degrees of freedom of cells at the interface between MPI processes.
We might just choose a better test case like the one described in https://github.com/dealii/dealii/pull/8912#issuecomment-545699245 such that it makes a difference if you do the average correctly or not.

This should be fine again if you loop over all cells that are not `artificial`, i.e. the `ghost` and `locally_owned` cells` in line 620.

masterleinad: What was the error with the previous implementation where you were setting values for all dofs?

masterleinad:
Would you mind squashing the commits a bit?

dougshidong:
Thanks for all those changes. I was making those simultaneously, I guess you were quicker. 

I know Prof. Bangerth wanted to see a test description. If you want to add the following comment at the top describing the test:

```
// The test initializes a non-zero solution over a hypercube using
// finite element space FESystem(FE_Q(2),dim). The solution is then
// interpolated to the finite element space FESystem(FE_Q(3),dim) using
// the VectorTools::interpolate() function.
//
// Since the solution is continuous across the elements, the solution
// values at the new support points of FESystem(FE_Q(3),3) should be
// equal to the old solution defined in FESystem(FE_Q(2),dim) at those
// same support points.
```

include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -602,56 +602,70 @@ namespace VectorTools
...
-              touch_count[local_dof_indices[j]] <
-                std::numeric_limits<decltype(touch_count)::value_type>::max(),
-              ExcInternalError());
-            ++touch_count[local_dof_indices[j]];
+            cell_2->get_dof_indices(local_dof_indices);
+
+            // Distribute cell vector.
+            for (unsigned int j = 0; j < dof_2.get_fe().dofs_per_cell; ++j)
+              {
+                ::dealii::internal::ElementAccess<OutVector>::add(
```
dougshidong: We should zero-out data_2 before we starting adding to it.

tests/vector_tools/interpolate.cc
```diff
@@ -0,0 +1,229 @@
...
+  if (dim == 1)
+    {
+      solution_fine.reinit(locally_owned_dofs_fine, MPI_COMM_WORLD);
+    }
+  else
+    {
+      solution_fine.reinit(locally_owned_dofs_fine,
+                           ghost_dofs_fine,
+                           MPI_COMM_WORLD);
+    }
```
dougshidong: Adding 

```cpp
solution_fine.add(1.0);
```

will reveal the bug of non-zero initial `data_2`.

bangerth:
Yes, I had briefly thought about the zeroing out issue while looking through the patch as well.

Let's keep this patch to strictly the issue at hand, namely getting the function to work in parallel. If you want to propose a separate patch (and separate testcase) for the issue with not currently setting the output vector to zero, then feel free to propose a separate pull request -- you then all get double credit for two pull requests :-)

elauksap:
I've squashed the commits and force-pushed, making @dougshidong co-author of the squash commits (he is also listed in the changelog).

In this final version the vector `data_2` is zeroed-out before assignment as correctly reported by @dougshidong!

Thanks for all the suggestions!

bangerth:
Looks good and ready to merge! Thanks for the collaboration, you two!

PR #8950, opened 2019-10-23
Provide a link to the geometry paper.

bangerth:


doc/doxygen/headers/glossary.h
```diff
@@ -1033,6 +1033,31 @@
...
+ * <dd>The "geometry paper" is a paper by L. Heltai, W. Bangerth, M. Kronbichler,
+ * and A. Mola, titled
+ * "Using exact geometry information in finite element computations", that
+ * describes how deal.II describes the geometry of domains. In particular,
+ * it discusses the algorithmic foundations on which the Manifold class
+ * is based, and what kind of information it needs to provide for mesh
+ * refinement, the computation of normal vectors, and the many other places
+ * where geometry enters into finite element computations.
+ *
+ * The paper is currently available on arXiv at https://arxiv.org/abs/1910.09824 .
```
tjhei: do you want a link here?

doc/doxygen/headers/glossary.h
```diff
@@ -1033,6 +1033,31 @@
...
+ * The full reference for this paper is as follows:
+ * @code{.bib}
+@misc{heltai2019using,
+    title={Using exact geometry information in finite element computations},
+    author={Luca Heltai and Wolfgang Bangerth and Martin Kronbichler and Andrea Mola},
+    year={2019},
+    eprint={1910.09824},
+    archivePrefix={arXiv},
+    primaryClass={math.NA}
```
tjhei: no doi and no url?

doc/doxygen/headers/glossary.h
bangerth: doxygen recognizes this as the link.

doc/doxygen/headers/glossary.h
bangerth: Well, the URL is above already, and arXiv papers have no DOIs :-)

doc/doxygen/headers/glossary.h
tjhei: doh, of course. Somehow I thought this was html source.

doc/doxygen/headers/glossary.h
tjhei: I think it is silly that the arxiv citation window doesn't add the url into the bibtex. That would be useful for anybody copying this. Anyways, not important...

bangerth:
I'll update the html in the bibtex once the paper is actually accepted/published.

Issue #8951, opened 2019-10-24
compile error with std::vector<FunctionParser<dim>>


rezarastak:
I am getting a compile error when I `push_back` (or `emplace_back`) data into an object of type `std::vector<FunctionParser<dim>>`.

There are multiple compile errors. All of them are related to the following member variable.
```c++
mutable Threads::ThreadLocalStorage<std::vector<std::unique_ptr<mu::Parser>>> fp;
```
One error is caused due to the forward declaration of `mu::Parser`. However, even after fixing that, I saw more compile errors which I couldn't fix.

The necessary files for reproducing this error and the logs are attached here.
[CMakeLists.txt](https://github.com/dealii/dealii/files/3765446/CMakeLists.txt)
[function_parser_vec.cc.txt](https://github.com/dealii/dealii/files/3765444/function_parser_vec.cc.txt)
[error.txt](https://github.com/dealii/dealii/files/3765442/error.txt)
[summary.log](https://github.com/dealii/dealii/files/3765443/summary.log)

bangerth:
Yes, `FunctionParser` objects can't be copied, but `std::vector<T>` requires that `T` is copyable. So you can't have a vector of `FunctionParser` objects.

rezarastak:
Thank you for explaining that. I will store my expressions in `std::vector<std::string>` instead of `std::vector<FunctionParser<dim>>` and use a single `FunctionParser` to create the functions later from the expressions. So, my issue is solved. 

Even though `FunctionParser` is is not copiable, I think it should allow us to move it into a container or construct it in-place using `emplace_back`. Since c++11, `std::vector<T>` does not require `T` to be copiable. For example, we should be able to do something like
```c++
FunctionParser<2> f("x");
std::vector<FunctionParser<2>> v;
v.push_pack(std::move(f));
v.emplace_back("x + y");
```

bangerth:
Yes, that's possible.

You can of course also store *pointers* to `FunctionParser* objects in your vector.

PR #8952, opened 2019-10-24
Delete FunctionParse copy constructor.

bangerth:
In response to #8951. This just makes more obvious what is already the case.

### DIFF
--- a/include/deal.II/base/function_parser.h
+++ b/include/deal.II/base/function_parser.h
@@ -259,6 +259,13 @@ public:
   ~FunctionParser() override;
 
   /**
+   * Copy operator. Objects of this type can not be copied, and
+   * consequently this operator is deleted.
+   */
+  FunctionParser &
+  operator=(const FunctionParser &) = delete;
+
+  /**
    * Type for the constant map. Used by the initialize() method.
    */
   using ConstMap = std::map<std::string, double>;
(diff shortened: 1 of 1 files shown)


bangerth:
@rezarastak -- FYI.

include/deal.II/base/function_parser.h
```diff
@@ -246,6 +246,12 @@ class FunctionParser : public AutoDerivativeFunction<dim>
                                                      ",t",
                  const double h = 1e-8);
 
+  /**
+   * Copy constructor. Objects of this type can not be copied, and
+   * consequently this constructor is deleted.
+   */
+  FunctionParse(const FunctionParser &) = delete;
```
masterleinad:
```suggestion
  FunctionParser(const FunctionParser &) = delete;
```

bangerth: That's what happens if you're working on an unplugged laptop and don't want to compile the entire library...

bangerth:
I've now also made the copy operator `delete`d as that of course had the same problem. Let me think about move constructors and operators -- I think they could work, but if you allow me, I'll take care of that in a separate patch.

/rebuild

masterleinad:
Sure. Defining the copy constructor implies deleting the move constructor so you are already touching that implicitly.

bangerth:
It turns out that some of the TBB data structures we use in this class have no appropriate move constructors/operators, so the compiler falls back to the copy constructors/operators instead that then, however, turns out to be deleted. In other words, moving doesn't work for these kinds of objects -- and so I've deleted the move operators to make that clear.

PR #8962, opened 2019-10-25
Output all faces and edges to vtk

peterrum:
Follow-up to #8902. It enables to output all non-internal faces and edges to VTK.

ping @luca-heltai @bangerth 

FYI @nfehn


include/deal.II/grid/grid_out.h
```diff
@@ -869,12 +869,14 @@ namespace GridOutFlags
     /**
      * Default constructor.
      */
-    Vtk(const bool output_cells    = true,
-        const bool output_faces    = true,
-        const bool output_co_faces = true)
+    Vtk(const bool output_cells         = true,
+        const bool output_faces         = true,
+        const bool output_co_faces      = true,
```
bangerth: I should have caught this earlier, but we don't generally use the term "co_faces". We just say "edges" and then say that this flag is simply ignored in 1d and 2d.

include/deal.II/grid/grid_out.h
```diff
@@ -891,6 +893,12 @@ namespace GridOutFlags
      * Output co-faces/edges.
      */
     bool output_co_faces;
+
+    /**
+     * Output only faces/co-faces, which are relevant (and differ
```
bangerth: The English language is less regular about the comma -- it's really more of a tool to break a train of thought (except in enumerations) and not necessary here.
```suggestion
     * Output only faces/edges which are relevant (and differ
```

include/deal.II/grid/grid_out.h
```diff
@@ -891,6 +893,12 @@ namespace GridOutFlags
...
+     * from default settings).
```
bangerth: Can you say what "relevant" actually means?

source/grid/grid_out.cc
```diff
@@ -2951,6 +2951,46 @@ namespace
     return v;
   }
 
+  /**
+   * Return all lines of non-internal faces in three dimension.
+   */
+  std::vector<typename Triangulation<3, 3>::active_line_iterator>
+  all_co_faces(const Triangulation<3, 3> &tria)
```
bangerth:
```suggestion
  get_edges(const Triangulation<3, 3> &tria)
```

source/grid/grid_out.cc
```diff
@@ -2951,6 +2951,46 @@ namespace
...
+  {
+    std::vector<typename Triangulation<3, 3>::active_line_iterator> res;
+
+    std::vector<bool> flags;
+    tria.save_user_flags_line(flags);
+    const_cast<Triangulation<3, 3> &>(tria).clear_user_flags_line();
+
+    for (auto face = tria.begin_active_face(); face != tria.end_face(); ++face)
```
bangerth: I think you can now say this:
```suggestion
    for (auto face = tria.active_face_iterators())
```

source/grid/grid_out.cc
```diff
@@ -2951,6 +2951,46 @@ namespace
...
+      for (unsigned int l = 0; l < GeometryInfo<3>::lines_per_face; ++l)
+        {
+          auto line = face->line(l);
```
bangerth:
```suggestion
          const auto line = face->line(l);
```

source/grid/grid_out.cc
```diff
@@ -2951,6 +2951,46 @@ namespace
...
+          if (line->user_flag_set() || line->has_children())
+            continue;
+          else
+            line->set_user_flag();
+          if (line->boundary_id() != numbers::invalid_boundary_id)
+            res.emplace_back(line);
```
bangerth: doesn't this also catch internal edges?

source/grid/grid_out.cc
```diff
@@ -2951,6 +2951,46 @@ namespace
...
+  }
+
+
+
+  /**
+   *Same as above, for 1 and 2 dimensional grids. Does nothing.
+   */
+  template <int dim, int spacedim>
+  std::vector<typename Triangulation<dim, spacedim>::active_line_iterator>
+  all_co_faces(const Triangulation<dim, spacedim> &)
```
bangerth:
```suggestion
  get_edges(const Triangulation<dim, spacedim> &)
```

source/grid/grid_out.cc
```diff
@@ -2951,6 +2951,46 @@ namespace
...
+            res.emplace_back(line);
+        }
+    const_cast<Triangulation<3, 3> &>(tria).load_user_flags_line(flags);
+    return res;
+  }
+
+
+
+  /**
+   *Same as above, for 1 and 2 dimensional grids. Does nothing.
```
bangerth:
```suggestion
   * Same as above, for 1 and 2 dimensional grids. Does nothing.
```

source/grid/grid_out.cc
```diff
@@ -2951,6 +2951,46 @@ namespace
     return v;
   }
 
+  /**
+   * Return all lines of non-internal faces in three dimension.
```
bangerth: I can't see where this is in the file, but the function should be put into an anonymous namespace if it isn't already.

source/grid/grid_out.cc
bangerth: Or maybe, more clearly, `get_edge_iterators()`.

source/grid/grid_out.cc
```diff
@@ -2995,6 +3035,27 @@ namespace
...
+   */
+  template <int dim, int spacedim>
+  std::vector<typename Triangulation<dim, spacedim>::active_face_iterator>
+  all_faces(const Triangulation<dim, spacedim> &tria)
+  {
+    std::vector<typename Triangulation<dim, spacedim>::active_face_iterator>
+      res;
+    if (dim == 1)
+      return res;
+    for (auto face = tria.begin_active_face(); face != tria.end_face(); ++face)
```
bangerth: as above

source/grid/grid_out.cc
```diff
@@ -2995,6 +3035,27 @@ namespace
 
 
 
+  /**
+   * Return all non-internal faces of a triangulation.
+   */
+  template <int dim, int spacedim>
+  std::vector<typename Triangulation<dim, spacedim>::active_face_iterator>
+  all_faces(const Triangulation<dim, spacedim> &tria)
```
bangerth: `get_faces()`. But on second thought, I think you want to qualify this: you're not returning all faces or edges, but in fact only the ones on the boundary.

include/deal.II/grid/grid_out.h
peterrum: This is how it is called internally... Should I nevertheless change it?

bangerth: Let's stick with our usual terminology, which is edge.

source/grid/grid_out.cc
peterrum: Does it? How should I read than: `line->boundary_id() != numbers::invalid_boundary_id`?

source/grid/grid_out.cc
peterrum: It is already!

peterrum:
@bangerth Thank you for your review! I have integrated your suggestions!

source/grid/grid_out.cc
bangerth: I don't recall what `invalid_boundary_id` corresponds to -- are these precisely the interior edges and faces? This might be easier to read if you wrote `if (line->at_boundary())`, assuming that that is what you want.

bangerth:
Just the one issue I commented on. Other than that, I think this looks good!

source/grid/grid_out.cc
peterrum: @bangerth done!

PR #8963, opened 2019-10-25
Update tutorials to use write_vtu_with_pvtu_record

### LINKED ISSUE #8960: Update tutorials step-37/18/32/45/55/48/42/64/50/40 to recent changes in PR #8904
The code for generating both .vtu and .pvtu output files has been simplified in PR #8904. Since several tutorials (step-37/18/32/45/55/48/42/64/50/40) write .vtu output, the code could be simplified according to the "Don't repeat yourself" paradigm.

peterrum:
some of the comments still have to be updated...

FYI @nfehn 

closes #8960

### DIFF
--- a/examples/step-18/step-18.cc
+++ b/examples/step-18/step-18.cc
@@ -1271,59 +1271,14 @@ namespace Step18
     // all these solution and other data vectors:
     data_out.build_patches();
 
-
-    // Let us determine the name of the file we will want to write it to. We
-    // compose it of the prefix <code>solution-</code>, followed by the time
-    // step number, and finally the processor id (encoded as a three digit
-    // number):
-    std::string filename =
-      "solution-" + Utilities::int_to_string(timestep_no, 4) + "." +
-      Utilities::int_to_string(this_mpi_process, 3) + ".vtu";
-
-    // The following assertion makes sure that there are less than 1000
-    // processes (a very conservative check, but worth having anyway) as our
-    // scheme of generating process numbers would overflow if there were 1000
-    // processes or more. Note that we choose to use <code>AssertThrow</code>
-    // rather than <code>Assert</code> since the number of processes is a
-    // variable that depends on input files or the way the process is started,
-    // rather than static assumptions in the program code. Therefore, it is
-    // inappropriate to use <code>Assert</code> that is optimized away in
-    // optimized mode, whereas here we actually can assume that users will run
-    // the largest computations with the most processors in optimized mode,
-    // and we should check our assumptions in this particular case, and not
-    // only when running in debug mode:
-    AssertThrow(n_mpi_processes < 1000, ExcNotImplemented());
-
-    // With the so-completed filename, let us open a file and write the data
-    // we have generated into it:
-    std::ofstream output(filename);
-    data_out.write_vtu(output);
+    // Let us open a file and write the data we have generated into it:
+    const auto pvtu_master_filename = data_out.write_vtu_with_pvtu_record(
+      "./", "solution-", timestep_no, 4, mpi_communicator);
 
     // The record files must be written only once and not by each processor,
     // so we do this on processor 0:
     if (this_mpi_process == 0)
       {
-        // Here we collect all filenames of the current timestep (same format as
-        // above)
-        std::vector<std::string> filenames;
-        for (unsigned int i = 0; i < n_mpi_processes; ++i)
-          filenames.push_back("solution-" +
-                              Utilities::int_to_string(timestep_no, 4) + "." +
-                              Utilities::int_to_string(i, 3) + ".vtu");
-
-        // Now we write the .visit file. The naming is similar to the .vtu
-        // files, only that the file obviously doesn't contain a processor id.
-        const std::string visit_master_filename =
-          ("solution-" + Utilities::int_to_string(timestep_no, 4) + ".visit");
-        std::ofstream visit_master(visit_master_filename);
-        DataOutBase::write_visit_record(visit_master, filenames);
-
-        // Similarly, we write the paraview .pvtu:
-        const std::string pvtu_master_filename =
-          ("solution-" + Utilities::int_to_string(timestep_no, 4) + ".pvtu");
-        std::ofstream pvtu_master(pvtu_master_filename);
-        data_out.write_pvtu_record(pvtu_master, filenames);
-
         // Finally, we write the paraview record, that references all .pvtu
         // files and their respective time. Note that the variable
         // times_and_names is declared static, so it will retain the entries
--- a/examples/step-18/step-18.cc
+++ b/examples/step-18/step-18.cc
@@ -1271,9 +1271,20 @@ namespace Step18
     // all these solution and other data vectors:
     data_out.build_patches();
 
-    // Let us open a file and write the data we have generated into it:
-    const auto pvtu_master_filename = data_out.write_vtu_with_pvtu_record(
-      "./", "solution", timestep_no, 4, mpi_communicator);
+    // Let us call a function that opens the necessary output files and writes
+    // the data we have generated into them. The function automatically
+    // constructs the file names from the given directory name (the first
+    // argument) and file name base (second argument). It augments the resulting
+    // string by pieces that result from the time step number and a "piece
+    // number" that corresponds to a part of the overall domain that can consist
+    // of one or more subdomains.
+    //
+    // The function also writes a record files (with suffix `.pvd`) for Paraview
+    // that describes how all of these output files combine into the data for
+    // this single time step:
+    const std::string pvtu_master_filename =
+      data_out.write_vtu_with_pvtu_record(
+        "./", "solution", timestep_no, 4, mpi_communicator);
 
     // The record files must be written only once and not by each processor,
     // so we do this on processor 0:
@@ -1288,13 +1299,6 @@ namespace Step18
           std::pair<double, std::string>(present_time, pvtu_master_filename));
         std::ofstream pvd_output("solution.pvd");
         DataOutBase::write_pvd_record(pvd_output, times_and_names);
-
-        std::ofstream visit_output("solution.visit");
-        static std::vector<std::pair<double, std::vector<std::string>>>
-          times_and_pieces;
-        times_and_pieces.emplace_back(
-          present_time, std::vector<std::string>(1, pvtu_master_filename));
-        DataOutBase::write_visit_record(visit_output, times_and_pieces);
       }
   }
 
--- a/examples/step-18/step-18.cc
+++ b/examples/step-18/step-18.cc
@@ -1273,7 +1273,7 @@ namespace Step18
 
     // Let us open a file and write the data we have generated into it:
     const auto pvtu_master_filename = data_out.write_vtu_with_pvtu_record(
-      "./", "solution-", timestep_no, 4, mpi_communicator);
+      "./", "solution", timestep_no, 4, mpi_communicator);
 
     // The record files must be written only once and not by each processor,
     // so we do this on processor 0:
--- a/examples/step-18/step-18.cc
+++ b/examples/step-18/step-18.cc
@@ -1288,6 +1288,13 @@ namespace Step18
           std::pair<double, std::string>(present_time, pvtu_master_filename));
         std::ofstream pvd_output("solution.pvd");
         DataOutBase::write_pvd_record(pvd_output, times_and_names);
+
+        std::ofstream visit_output("solution.visit");
+        static std::vector<std::pair<double, std::vector<std::string>>>
+          times_and_pieces;
+        times_and_pieces.emplace_back(
+          present_time, std::vector<std::string>(1, pvtu_master_filename));
+        DataOutBase::write_visit_record(visit_output, times_and_pieces);
       }
   }
 
--- a/examples/step-32/step-32.cc
+++ b/examples/step-32/step-32.cc
@@ -3243,41 +3243,9 @@ namespace Step32
     data_out.add_data_vector(locally_relevant_joint_solution, postprocessor);
     data_out.build_patches();
 
-    static int        out_index = 0;
-    const std::string filename =
-      ("solution-" + Utilities::int_to_string(out_index, 5) + "." +
-       Utilities::int_to_string(triangulation.locally_owned_subdomain(), 4) +
-       ".vtu");
-    std::ofstream output(filename);
-    data_out.write_vtu(output);
-
-
-    // At this point, all processors have written their own files to disk. We
-    // could visualize them individually in Visit or Paraview, but in reality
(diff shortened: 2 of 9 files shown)


examples/step-42/step-42.cc
```diff
@@ -2046,32 +2046,10 @@ namespace Step42
...
-             ++i)
-          filenames.push_back(filename_base + "-" +
-                              Utilities::int_to_string(i, 4) + ".vtu");
-
-        std::ofstream pvtu_master_output(
-          (output_dir + filename_base + ".pvtu"));
-        data_out.write_pvtu_record(pvtu_master_output, filenames);
-
-        std::ofstream visit_master_output(
-          (output_dir + filename_base + ".visit"));
```
peterrum: What do we want to do with this `.visit` file?

tjhei:
you can drop step-50, see #8948

examples/step-18/step-18.cc
```diff
@@ -1271,69 +1271,9 @@ namespace Step18
...
-
-        // Finally, we write the paraview record, that references all .pvtu
-        // files and their respective time. Note that the variable
-        // times_and_names is declared static, so it will retain the entries
-        // from the previous timesteps.
-        static std::vector<std::pair<double, std::string>> times_and_names;
-        times_and_names.push_back(
-          std::pair<double, std::string>(present_time, pvtu_master_filename));
-        std::ofstream pvd_output("solution.pvd");
-        DataOutBase::write_pvd_record(pvd_output, times_and_names);
```
tjhei: you need to keep the pvd record.

examples/step-18/step-18.cc
```diff
@@ -1271,69 +1271,9 @@ namespace Step18
...
-      }
+    // Let us open a file and write the data we have generated into it:
+    data_out.write_vtu_with_pvtu_record(
```
tjhei: what happened to ``timestep_no``?

examples/step-32/step-32.cc
```diff
@@ -3243,41 +3243,9 @@ namespace Step32
...
-                              Utilities::int_to_string(i, 4) + ".vtu");
-        const std::string pvtu_master_filename =
-          ("solution-" + Utilities::int_to_string(out_index, 5) + ".pvtu");
-        std::ofstream pvtu_master(pvtu_master_filename);
-        data_out.write_pvtu_record(pvtu_master, filenames);
-
-        const std::string visit_master_filename =
-          ("solution-" + Utilities::int_to_string(out_index, 5) + ".visit");
-        std::ofstream visit_master(visit_master_filename);
-        DataOutBase::write_visit_record(visit_master, filenames);
```
tjhei: what about the visit record, do we care? @bangerth are you using .pvtu or .visit with visit?

examples/step-40/step-40.cc
```diff
@@ -572,40 +572,9 @@ namespace Step40
...
-          filenames.push_back("solution-" + Utilities::int_to_string(cycle, 2) +
-                              "." + Utilities::int_to_string(i, 4) + ".vtu");
-
-        std::ofstream master_output(
-          "solution-" + Utilities::int_to_string(cycle, 2) + ".pvtu");
-        data_out.write_pvtu_record(master_output, filenames);
-      }
+    // The next step is to write this data to disk.
+    data_out.write_vtu_with_pvtu_record(
+      "./", "solution-", cycle, 2, mpi_communicator);
```
tjhei: can we use MPI I/O here and explain it in a paragraph or so? Maybe use 8 or so?

peterrum:
@tjhei Thanks for your comments! I have integrated them. The only question is what happens with the visit record. In step-50, you have eliminated it...

tjhei:
I don't care about it, but I let Wolfgang comment if he needs them to use with visit. 😉

examples/step-18/step-18.cc
```diff
@@ -1335,470 +1290,469 @@ namespace Step18
...
       }
   }
+} // namespace Step18
 
 
 
-  // @sect4{TopLevel::do_initial_timestep}
-
-  // This and the next function handle the overall structure of the first and
-  // following timesteps, respectively. The first timestep is slightly more
```
tjhei: What happened here?

peterrum: I have accidentally added a closing bracket...

bangerth:
On 10/26/19 8:26 AM, Timo Heister wrote:
> I don't care about it, but I let Wolfgang comment if he needs them to use with 
> visit. 😉

Yes, please keep the .visit file. It's the equivalent of the .pvd file that 
Paraview uses.

peterrum:
@bangerth I find it quite annoying to keep the `.visit`-stuff because for that we have to collect the  `filenames` (which also happens within `write_vtu_with_pvtu_record`). The result is that the code will not be simpler as before...

Personally, I think there should be no `write_vtu_with_pvtu_record()` but there should be a `DataOutFlags::Vtu` (similarly as `GridOutFlags::Vtu`), which configures `write_vtu()` in regard of number of output files, `.pvtu`- and `.visit`-output (and other parameters). Or does anyone want to implement `write_vtu_with_visit_record()`?

tjhei:
> I find it quite annoying to keep the `.visit`-stuff because for that we have to collect t

okay, so the current .visit master file format lists all individual files at each timestep:
```
!TIME 0
!TIME 1
!TIME 2
!TIME 3
!NBLOCKS 4
solution/solution-00000.0000.vtu
solution/solution-00000.0001.vtu
solution/solution-00000.0002.vtu
solution/solution-00000.0003.vtu
solution/solution-00001.0000.vtu
solution/solution-00001.0001.vtu
solution/solution-00001.0002.vtu
solution/solution-00001.0003.vtu
solution/solution-00002.0000.vtu
solution/solution-00002.0001.vtu
solution/solution-00002.0002.vtu
solution/solution-00002.0003.vtu
solution/solution-00003.0000.vtu
solution/solution-00003.0001.vtu
solution/solution-00003.0002.vtu
solution/solution-00003.0003.vtu
```

I just played around with this, and this works for me in visit as well:
````
!TIME 0
!TIME 1
!TIME 2
!TIME 3
!NBLOCKS 1
solution/solution-00000.pvtu
solution/solution-00001.pvtu
solution/solution-00002.pvtu
solution/solution-00003.pvtu
```

This means we can make a new function to generate a visit master record (like the .pvd) that doesn't need the individual filenames but references the .pvtus!  Then write_vtu_with_pvtu_record can just also write a .visit file...

examples/step-18/step-18.cc
```diff
@@ -1333,6 +1288,8 @@ namespace Step18
           std::pair<double, std::string>(present_time, pvtu_master_filename));
         std::ofstream pvd_output("solution.pvd");
         DataOutBase::write_pvd_record(pvd_output, times_and_names);
+        std::ofstream visit_output("solution.visit");
+        DataOutBase::write_visit_record(visit_output, times_and_names);
```
peterrum: @tjhei Like this? In the other visit-cases this does not work. Here it is used as an equivalent for `.pvd`, there is is used as an equivalent for `.pvtu`.

peterrum:
> edit: @peterrum I can make these changes to the master record function, if you want. This means you can delete everything related to .visit files here for now.

@tjhei I have removed the `.visit` instances.

tjhei:
I pushed a couple of minor changes on top of yours.

To explain the situation, there are two different kind of records: 
1. group pieces into a single time step (.pvtu can be read by visit and paraview and a .visit that can only be read by visit)
   => my vote would be to only write the .pvtu because:
    1. fewer files
    2. .pvtu is a standard, .visit is not
    3. less code in tutorials
    4. it is confusing to have two different kind of .visit files (see below)
2. There are master records that contain several time steps with times (.pvd and .visit, can only be read by paraview and visit, respectively).
    => it turns out non of the tutorials were writing this file in visit format. I added it for step-18 here.

@bangerth are you okay with this?

peterrum:
@tjhei looks good! Thanks!

peterrum:
Is there a way to test the tutorials?

tjhei:
> Is there a way to test the tutorials?

No, I am not aware of a way to do that easily.

examples/step-18/step-18.cc
```diff
@@ -1271,59 +1271,14 @@ namespace Step18
...
-    // the largest computations with the most processors in optimized mode,
-    // and we should check our assumptions in this particular case, and not
-    // only when running in debug mode:
-    AssertThrow(n_mpi_processes < 1000, ExcNotImplemented());
-
-    // With the so-completed filename, let us open a file and write the data
-    // we have generated into it:
-    std::ofstream output(filename);
-    data_out.write_vtu(output);
+    // Let us open a file and write the data we have generated into it:
```
bangerth:
```suggestion
    // Let us call a function that opens the necessary output files and writes the 
    // data we have generated into them. The function automatically constructs
    // the file names from the given directory name (the first argument) and file
    // name base (second argument). It augments the resulting string by pieces
    // that result from the time step number and a "piece number" that corresponds
    // to a part of the overall domain that can consist of one or more subdomains.
    //
    // The function also writes a record files (with suffix `.pvd`) for Paraview
    // that describes how all of these output files combine into the data for
    // this single time step:
```

examples/step-18/step-18.cc
```diff
@@ -1271,59 +1271,14 @@ namespace Step18
...
+    const auto pvtu_master_filename = data_out.write_vtu_with_pvtu_record(
```
bangerth: Let's be explicit:
```suggestion
    const std::string pvtu_master_filename = data_out.write_vtu_with_pvtu_record(
```

examples/step-18/step-18.cc
```diff
@@ -1333,6 +1288,13 @@ namespace Step18
           std::pair<double, std::string>(present_time, pvtu_master_filename));
         std::ofstream pvd_output("solution.pvd");
         DataOutBase::write_pvd_record(pvd_output, times_and_names);
+
+        std::ofstream visit_output("solution.visit");
```
bangerth:
```suggestion
        // The final piece of this is to also write a master record for Visit, in the same way as the
        // call above already did for Paraview:
        std::ofstream visit_output("solution.visit");
```

examples/step-32/step-32.cc
bangerth: I'm using the .visit file. I *think* it is true that Visit can now read .pvd files, right?

examples/step-18/step-18.cc
```diff
@@ -1333,6 +1288,13 @@ namespace Step18
...
+        static std::vector<std::pair<double, std::vector<std::string>>>
+          times_and_pieces;
```
bangerth: The use of a `static` variable is really not very pleasant here because it means that you can't call the simulator twice in a row, or twice in parallel. I think it would be nicer to put this into a member variable instead.

But maybe we don't need this -- does Visit now read .pvd files?

examples/step-32/step-32.cc
```diff
@@ -3243,41 +3243,9 @@ namespace Step32
...
-      }
+    static int out_index = 0;
+    data_out.write_vtu_with_pvtu_record(
```
bangerth: Same here with the static variable -- I recognize that that's preexisting, but if it were me, I'd say let's just make that a member variable with a more descriptive name.

examples/step-40/step-40.cc
```diff
@@ -572,40 +572,11 @@ namespace Step40
...
-
-        std::ofstream master_output(
-          "solution-" + Utilities::int_to_string(cycle, 2) + ".pvtu");
-        data_out.write_pvtu_record(master_output, filenames);
-      }
+    // The next step is to write this data to disk. We write up to 8 VTU files
+    // in parallel with the help of MPI-IO. Additionally a PVTU record is
+    // generated, which groups the written VTU files.
+    data_out.write_vtu_with_pvtu_record(
+      "./", "solution", cycle, 2, mpi_communicator, 8);
```
bangerth: Is 8 a good number? We ran this program with up to 16k processors. I'd expect that with just 8 blocks we'd get into a rather bad bottleneck here!

peterrum: > just 8 blocks we'd get into a rather bad bottleneck here

It should not be a bottleneck. All 16k processes would write in parallel in 8 files via MPI-IO.

tjhei:
> I'm using the .visit file. I think it is true that Visit can now read .pvd files, right?

Yes, I verified that this works. Did you read my comment above? https://github.com/dealii/dealii/pull/8963#issuecomment-547284404

examples/step-32/step-32.cc
tjhei: can we do the static variables in a separate PR?

examples/step-40/step-40.cc
tjhei: There is several considerations:
1. Does writing the output scale? When I wrote the routines (a few years ago), I tested writing into a single output file from 1000+ cores and got excellent performance (10s of GB/s). I don't see a problem with this.
2. Is this sensible for visualization? If you do visualization in serial, it doesn't matter. If you visualize in parallel, it might be helpful to have more than one file. With 8, you can at least run on 8 nodes. Maybe it makes sense to increase this number slightly? Not that it matters that much...

examples/step-32/step-32.cc
bangerth: Yes.

examples/step-40/step-40.cc
bangerth: What if we used `max(8,ceil(n_procs/128))`? I mean, right now we write one file per processor; maybe we can find a compromise that at least makes sure we don't make this the one function that blocks everything if someone tried to run the program on 10k processors...

bangerth:
@tjhei: Thanks, I had missed that. Then let's not worry about .visit files any more from now on.

peterrum:
@tjhei @bangerth I have implemented the suggestions. However, I have decided against `max(8,ceil(n_procs/128))`. Personally, I think everything above 1 is not really useful. This should be handled by a suitable configuration of MPI I/O according to the GPFS at hand. We show here an example, where we set the value to an arbitrary value (`8`), but actually it should be by default 1.

tjhei:
> Personally, I think everything above 1 is not really useful.

It should be helpful if you use parallel visualization (paraview server for example).

> but actually it should be by default 1.

from the write performance point of view? Likely, yes.

> @bangerth: maybe we can find a compromise that at least makes sure we don't make this the one function that blocks everything if someone tried to run the program on 10k processors...

I have to admit that I haven't made a performance comparison with the number of files and maybe we should test before making any more guesses (especially complicated ones)? I am fairly certain that 1 file is way more performant than anything else. I don't see why 8 files would "block everything" in a large run.

peterrum:
>It should be helpful if you use parallel visualization (paraview server for example).

I see! Thanks for the info!

peterrum:
Anything missing?

Issue #8961, opened 2019-10-25
Failure to configure with python bindings on Ubuntu


agrayver:
I get this error when trying to compile with python bindings:

> -- Setting up python bindings
> -- Found PythonInterp: /home/ag/anaconda3/bin/python (found version "3.7.3") 
> -- Found PythonLibs: /home/ag/anaconda3/lib/libpython3.7m.so (found version "3.7.3") 
> CMake Warning at /usr/share/cmake-3.10/Modules/FindBoost.cmake:801 (message):
>   New Boost version may have incorrect or missing dependencies and imported
>   targets
> Call Stack (most recent call first):
>   /usr/share/cmake-3.10/Modules/FindBoost.cmake:907 (_Boost_COMPONENT_DEPENDENCIES)
>   /usr/share/cmake-3.10/Modules/FindBoost.cmake:1558 (_Boost_MISSING_DEPENDENCIES)
>   contrib/python-bindings/source/CMakeLists.txt:38 (_FIND_PACKAGE)
>
> -- Could NOT find Boost
> CMake Error at contrib/python-bindings/source/CMakeLists.txt:42 (MESSAGE):
>   DEAL_II_COMPONENT_PYTHON_BINDINGS has unmet configuration requirements: The
>   external boost library does not provide Boost.Python

This error comes from this snippet in '/dealii/contrib/python-bindings/source':

```
#
# Unset Boost_Found and run the low level FindBOOST.cmake module again to
# pick up libboost_python.so
#
LIST(REMOVE_ITEM CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)
SET(Boost_FOUND)
# Use the low level _FIND_PACKAGE function instead of our wrapper
_FIND_PACKAGE(Boost 1.54 COMPONENTS python)
LIST(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)

IF(NOT Boost_FOUND)
  MESSAGE(FATAL_ERROR
    "DEAL_II_COMPONENT_PYTHON_BINDINGS has unmet configuration requirements: "
    "The external boost library does not provide Boost.Python"
    )
ENDIF()
```
This indeed seems fragile... The external boost.python v. 1.71 is compiled and installed correctly and passes independent tests. Is there a better way to add boost.python to the configuration?

agrayver:
[...]

PR #8968, opened 2019-10-28
Fix python bindings configuration for new boost

agrayver:
In relation to #8961 

@masterleinad is this what you meant?


masterleinad:
Nearly. I was suggesting to bump the required version to 1.59 since this is what we require elsewhere, keep the current logic for <1.67 and only add `${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}` for version >=1.67

agrayver:
Done. I don't understand the purpose of "LIST(REMOVE_ITEM ...) ... LIST(APPEND ...)" so I kept it for the old branch only. The _FIND_PACKAGE line alone suffices for the CMake 3.10 and 3.14 and, I believe, provides the same behaviour if we add keyword REQUIRED.

agrayver:
ping

masterleinad:
Can you confirm that this works for boost version <1.67.0 when removing
```
  LIST(REMOVE_ITEM CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)
  SET(Boost_FOUND)
  # Use the low level _FIND_PACKAGE function instead of our wrapper
  LIST(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)
```
? This is what I wanted to check before merging.

agrayver:
> Can you confirm that this works for boost version <1.67.0 when removing

Yes, it does work:

```
-- Setting up python bindings
-- Found PythonInterp: /home/ag/anaconda3/envs/gofem/bin/python (found version "3.7.3") 
-- Found PythonLibs: /home/ag/anaconda3/envs/gofem/lib/libpython3.7m.so (found version "3.7.3") 
-- Boost 1.59 found.
-- Found Boost components:
   python
-- Setting up python bindings - Done
```


Shall I remove these lines?

masterleinad:
Yes, that appears to be OK.

agrayver:
Done.

PR #8971, opened 2019-10-28
Add copy assignment to MatrixFree::AdditionalData

peterrum:
FYI @kronbichler @nfehn


tjhei:
can we get a unit test please?

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -267,6 +267,35 @@ class MatrixFree : public Subscriptor
...
+      tasks_parallel_scheme = other.tasks_parallel_scheme;
+      tasks_block_size      = other.tasks_block_size;
+      mapping_update_flags  = other.mapping_update_flags;
+      mapping_update_flags_boundary_faces =
+        other.mapping_update_flags_boundary_faces;
+      mapping_update_flags_inner_faces = other.mapping_update_flags_inner_faces;
+      mapping_update_flags_faces_by_cells =
+        other.mapping_update_flags_faces_by_cells;
+      mg_level            = other.mg_level;
+      level_mg_handler    = this->mg_level;
```
tjhei: Can you leave a comment for this one here, please?

peterrum:
@tjhei The copy assignment in this PR and the copy constructor from one of the last PRs is a workaround for the `level_mg_handler` which is an alias for `mg_level`. The variable `level_mg_handler` is deprecated and will be removed. Once that happens, the copy assignment and the copy constructor can be removed. Do we need nevertheless a test?

peterrum:
@tjhei Added the test!

tjhei:
> @tjhei Added the test!

great, thank you. I think the test is useful, as we might rename or add some other flag in the future and if it is untested, it is likely to break again.

Issue #8969, opened 2019-10-28
Do not transfer MPI communicator for p::s::Tria::copy_tria


agrayver:
I use p::s::Tria::copy_triangulation and it turns to copy the MPI communicator:

https://github.com/dealii/dealii/blob/1634ecf9dd790e9a0d132711586f180c8c9f2f2e/source/distributed/tria_base.cc#L74-L80

This breaks my workflow where I create p::s::Tria on a sub-communicator and want to copy only the content of p::s::Tria, but not the communicator ownership. 

Can we make the line 80 above optional? I can submit a PR. 

On that note, the doc of this method seems somewhat ambiguous:

> This function also partitions triangulation based on the MPI communicator provided to the constructor.

What constructor is meant here? That of the argument?

masterleinad:
What is your use case here? Do you want to drop processors that contain no cells or do you simply want to duplicate the communicator?
I could see to make it an option as long as the default case is not changed.

agrayver:
> What is your use case here? Do you want to drop processors that contain no cells or do you simply want to duplicate the communicator?

I split communicator (MPI_Comm_split) and create a set of p::s::Tria for each subcommunicator. Smth like:

```
p::s::Tria tria(MPI_COMM_WORLD)
// work with tria / create / refine / etc

subcom = MPI_Comm_split(MPI_COMM_WORLD)
p::s::Tria tria_sub(subcom)
tria.copy_triangulation(tria_sub) // THIS WILL MAKE tria_sub belong to the MPI_COMM_WORLD
```

> I could see to make it an option as long as the default case is not changed.

Yes, essentially I am thinking of replacing the current method signature with (note optional bool):

```
TriangulationBase<dim, spacedim>::copy_triangulation(
    const dealii::Triangulation<dim, spacedim> &other_tria, bool transfer_communicator = true)
```

Depending on the flag, the communicator won't be copied.

bangerth:
On 10/31/19 3:04 AM, Alexander Grayver wrote:
> |subcom = MPI_Comm_split(MPI_COMM_WORLD) p::s::Tria tria_sub(subcom) 
> tria.copy_triangulation(tria_sub) // THIS WILL MAKE tria_sub belong to the 
> MPI_COMM_WORLD|

I don't think I understand this statement. Are you saying that `tria` now 
operates on `subcom` or `MPI_COMM_WORLD`? I think it should continue to work 
on subcom because that's what the triangulation was initialized with -- in 
[...]

PR #8979, opened 2019-10-31
Do not transfer the MPI communicator when copying the triangulation

agrayver:
Follow-up of the discussion in #8969 

@bangerth @masterleinad FYI

### DIFF
--- a/include/deal.II/distributed/tria_base.h
+++ b/include/deal.II/distributed/tria_base.h
@@ -98,6 +98,11 @@ namespace parallel
 
     /**
      * Implementation of the same function as in the base class.
+     *
+     * @note This function copies the cells, but not the communicator,
+     * of the source triangulation. In other words, the resulting
+     * triangulation will operate on the communicator it was constructed
+     * with.
      */
     virtual void
     copy_triangulation(
--- a/doc/news/changes/minor/20193110Grayver
+++ b/doc/news/changes/minor/20193110Grayver
@@ -1,0 +1,4 @@
+Changed: parallel::TriangulationBase::copy_triangulation does not
+copy the MPI communicator from the given source triangulation.
+<br>
+(Alexander Grayver, 2019/10/31)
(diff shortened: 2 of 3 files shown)


masterleinad:
```
invalid author 'agrayver' without firstname and lastname
```
Would you please fix your name for the commits?

include/deal.II/distributed/tria_base.h
```diff
@@ -98,6 +98,11 @@ namespace parallel
 
     /**
      * Implementation of the same function as in the base class.
+     *
+     * @note this functions copies the cells, but not the communicator,
```
bangerth:
```suggestion
     * @note This functions copies the cells, but not the communicator, 
```

Rombur: There is also an extra `s` add the end of `function`

agrayver:
Ready from my side.

masterleinad:
CI is still failing:
```
invalid author 'agrayver' without firstname and lastname
```
We require to have proper first and last name in the commits.

agrayver:
Well I am not sure how to fix this? Seemingly, it is all right:

> (base) ag@ag:~$ cat .gitconfig 
> [user]
> 	name = Alexander Grayver

peterrum:
@agrayver  The problem is that your first commit still has `agrayver` as author. If you `rebase -i c223f4e0c4256f9f9ca9816500bf07fb25989bb2`, squash both commits and force push the changes, everything should work!

masterleinad:
```
git commit --amend --author="John Doe <john@doe.org>"
```
should work in any way.

peterrum:
> should work in any way.

That will only operate on the last commit, will it not? The problem is with the commit before...

masterleinad:
That depends on the commit you are currently looking at. If you do an interactive rebase to fix a certain commit and mark it as `edit`, this is a way to change the author's information. I am not sure if calling `amend` without the flag is sufficient as well.

agrayver:
I did everything as suggested above, but this still fails. Any clue?

masterleinad:
I can take care of it.

masterleinad:
Here you go.

agrayver:
Thank you. Can we then merge it?

PR #8980, opened 2019-10-31
Fix a step-61 typo

GrahamBenHarper:
I saw a typo in the dox. Should it read more like this?

@sophy1029 FYI


bangerth:
Yes, please! If you find something that's not right, just fix it!

PR #8981, opened 2019-10-31
Add new read_cell_data to FEEvaluationBase for arbitrary data types

peterrum:
... this way we can access more complex data types from `FEEvaluation` and `FEFaceEvaluation` (not just `Number` and in this case `CellId`).

@kronbichler what do you think?

ping @lpsaavedra


peterrum:
@kronbichler As a side remark: in a next step, I intend to unify it with the other `read_cell_data` by introducing a *new* method `get_cell_ids`.

PR #8982, opened 2019-10-31
Fix create_construction_data_from_triangulation for periodic boundaries

peterrum:
... and create a function `collect_coinciding_vertices`, which I am now also using in `p:f:Util::create_construction_data_from_triangulation`.


source/grid/grid_tools.cc
```diff
@@ -5651,6 +5651,125 @@ namespace GridTools
...
+
+              if (it_b != vertex_to_coinciding_vertex_group.end())
+                it_b->second = temp;
+              else
+                vertex_to_coinciding_vertex_group[vertex_b] = temp;
+            }
+        }
+
+      // 1b) compress map: let vertices point to the coinciding vertex with
+      //     the smallest index
```
drwells: Its not very important, but these are no longer part of step 1 so we can get rid of `1b`, `1c`, etc.

peterrum:
@drwells Thanks for your review! I have updated the comments.

The problem was as follows: given is a hypercube (with periodic faces in all direction), which is partitioned in for parts as depicted here:

pbc1

In this configuration, the the ghost layer (locally owned cells) of rank 0 should look like this: 

pbc2

However, the top-right cell was missing...

source/distributed/fully_distributed_tria_util.cc
```diff
@@ -154,34 +154,34 @@ namespace parallel
...
+            std::vector<bool> &vertices_owned_by_locally_owned_cells) {
             // add local vertices
             for (unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell;
                  ++v)
-              vertices_owned_by_locally_owned_cells[cell->vertex_index(v)] =
-                true;
+              {
+                vertices_owned_by_locally_owned_cells[cell->vertex_index(v)] =
+                  true;
+                auto coinciding_vertex_group =
```
drwells:
```suggestion
                const auto coinciding_vertex_group =
```

source/distributed/fully_distributed_tria_util.cc
```diff
@@ -154,34 +154,34 @@ namespace parallel
...
+                  vertex_to_coinciding_vertex_group.find(cell->vertex_index(v));
+                if (coinciding_vertex_group !=
+                    vertex_to_coinciding_vertex_group.end())
+                  for (auto &co_vertex : coinciding_vertex_groups.at(
```
drwells:
```suggestion
                  for (const auto &co_vertex : coinciding_vertex_groups.at(
```

peterrum:
@drwells done!

masterleinad:
Looks like this broke some tests: https://cdash.43-1.org/viewTest.php?onlydelta&buildid=2600.
Does the test pass for you locally?

PR #8989, opened 2019-11-02
Added changes entry for pull request #8973

winnifried:
As requested by Wolfgang, here is the changes entry.


doc/news/changes/minor/20191102Wollner
```diff
@@ -0,0 +1,6 @@
+Changed: Fixed the computation of the FiniteElement.degree in
```
drwells:
```suggestion
Changed: Fixed the computation of the finite element degree in
```
I don't think doxygen will render this correctly so lets avoid code.

winnifried: Fine with me, I have updated the file.

drwells:
Looks good to me!

PR #8994, opened 2019-11-04
Slightly simplify MPI::ComputeIndexOwner::Dictionary range detection

kronbichler:
As discussed with @peterrum @tjhei @bangerth I have taken the chance to slightly simplify the code and remove a place where we made explicit use of the `dof<->rank` association in the dictionary of the compute index owner functionality.


tjhei:
@peterrum does this look correct to you?

tjhei:
I was successful in running a computation with block systems and 5 billion unknowns with this patch.

PR #8996, opened 2019-11-04
Use face_iterators in tutorials

masterleinad:
Part of #8563. There are still many places where we rely on the face number in the tutorials, but this should cover all the cases that can be converted up to now.
For this to work, I also had to implement a `DoFCellAccessor` version of `CellAccessor::face_iterators`.


examples/step-24/step-24.cc
```diff
@@ -341,12 +341,12 @@ namespace Step24
...
       for (const auto &cell : dof_handler.active_cell_iterators())
-        for (unsigned int f = 0; f < GeometryInfo<dim>::faces_per_cell; ++f)
-          if (cell->at_boundary(f))
+        for (const auto &face : cell->face_iterators())
+          if (face->at_boundary())
             {
               cell_matrix = 0;
 
-              fe_values.reinit(cell, f);
+              fe_values.reinit(cell, face);
```
bangerth: I think this way of initializing `FEFaceValues` is actually quite nice!

examples/step-44/step-44.cc
```diff
@@ -2333,10 +2332,8 @@ namespace Step44
     // Next we assemble the Neumann contribution. We first check to see it the
     // cell face exists on a boundary on which a traction is applied and add
     // the contribution if this is the case.
-    for (unsigned int face = 0; face < GeometryInfo<dim>::faces_per_cell;
-         ++face)
-      if (cell->face(face)->at_boundary() == true &&
-          cell->face(face)->boundary_id() == 6)
+    for (const auto &face : cell->face_iterators())
+      if (face->at_boundary() == true && face->boundary_id() == 6)
```
bangerth: In almost all other places in this patch, we use the following instead:
```suggestion
      if (face->at_boundary() && face->boundary_id() == 6)
```
I don't care one way or the other, so let's just go with the common style.

examples/step-61/step-61.cc
```diff
@@ -441,7 +441,7 @@ namespace Step61
     // @p face component of the shape functions.
     const FEValuesExtractors::Vector velocities(0);
     const FEValuesExtractors::Scalar interior(0);
-    const FEValuesExtractors::Scalar face(1);
+    const FEValuesExtractors::Scalar face_extractor(1);
```
bangerth: Let's find a better name for this than to duplicate `_extractor`. If you call the previous variable `p_interior`, then this could be called `p_interface`.

examples/step-61/step-61.cc
```diff
@@ -747,7 +745,7 @@ namespace Step61
     const FEValuesExtractors::Vector velocities(0);
     const FEValuesExtractors::Scalar pressure(dim);
     const FEValuesExtractors::Scalar interior(0);
-    const FEValuesExtractors::Scalar face(1);
+    const FEValuesExtractors::Scalar face_extractor(1);
```
bangerth: same here

bangerth:
Please address the two or three places I marked up and then merge yourself!

bangerth:
Nice, thank you!

PR #8998, opened 2019-11-04
Fix compilation of gmsh support

rumajo:
Use std::ofstream rather than just ofstream.

It's unclear to me how this ever compiled, unless one the of the includes does `using namespace std` or `using std::ofstream`?


drwells:
Nice catch: I have no idea how this worked either.

PR #8999, opened 2019-11-04
Step-61 Use FE_DGRT instead of FE_RT

### LINKED ISSUE #8978: Use FE_DGRT everywhere in step-61
step-61 still uses `FE_RaviartThomas` in a number of places. We should really be using `FE_DGRT` everywhere -- not because it makes a difference (all of the places do cell-local computations), but because it's the right thing to do (precisely because we're doing cell-local computations: All of the things we compute only live on a single cell, with no shape functions extending to neighboring cells.

@sophy1029 -- do you want to give this a try? I think you really just need to replace all of the places where we create a `FE_RaviartThomas` by the other element, remove the `fe_raviart_thomas.h` header file, and make sure that the result doesn't change.

sophy1029:
Fixes #8978.

### DIFF
--- a/examples/step-61/step-61.cc
+++ b/examples/step-61/step-61.cc
@@ -261,9 +261,7 @@ namespace Step61
   WGDarcyEquation<dim>::WGDarcyEquation(const unsigned int degree)
     : fe(FE_DGQ<dim>(degree), 1, FE_FaceQ<dim>(degree), 1)
     , dof_handler(triangulation)
-    ,
-
-    fe_dgrt(0)
+    , fe_dgrt(degree)
     , dof_handler_dgrt(triangulation)
 
   {}
@@ -311,7 +309,6 @@ namespace Step61
     solution.reinit(dof_handler.n_dofs());
     system_rhs.reinit(dof_handler.n_dofs());
 
-    darcy_velocity.reinit(dof_handler_dgrt.n_dofs());
 
     {
       constraints.clear();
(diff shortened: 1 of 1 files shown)


examples/step-61/step-61.cc
```diff
@@ -672,10 +669,14 @@ namespace Step61
   template <int dim>
   void WGDarcyEquation<dim>::compute_velocity_errors()
   {
-    const FE_RaviartThomas<dim> fe_rt(fe.base_element(0).degree);
+    const FE_DGRaviartThomas<dim> fe_dgrt(fe.base_element(0).degree);
+    DoFHandler<dim>               dof_handler_dgrt(triangulation);
+    dof_handler_dgrt.distribute_dofs(fe_dgrt);
+
+    darcy_velocity.reinit(dof_handler_dgrt.n_dofs());
```
bangerth: If you move the `dof_handler_dgrt` into this function, then you also need to move the declaration of the `darcy_velocity` vector here because it references the enumeration scheme of the `dof_handler_dgrt`. Is that possible, or are you using the vector in some other function as well?

sophy1029: I checked the convergence rates, they do not change.
I calculated darcy_velocity here, and I draw the Darcy velocity figures in output_results.

bangerth: Ah, then we can't move the `dof_handler_dgrt`, I'm afraid. Can you move it back into the class like you had before?

sophy1029: Is that possible we also define dof_handler_dgrt in output_results?

bangerth: That's what you are already doing right now.

What I mean is this: Imagine that you decide to use a different enumeration scheme for the `dof_handler_dgrt` in the two places where you define it right now, for example by calling one of the `DoFRenumbering` functions in one of the two places. Because you save the `darcy_velocity` in a member variable, that would mean that you interpret which degrees of freedom the individual components of `darcy_velocity` are associated with using the numbering scheme in the first of these functions, but that will no longer be correct in the second function because you have changed the numbering scheme in between. Does that make sense?

The solutions are either to move the `dof_handler_dgrt` back into a member variable as you had before, or to also make `darcy_velocity` a local variable in each of the two functions -- but then you have to re-compute `darcy_velocity` in `output_results()`. One could do that, but it requires copying a bunch of code and that seems wasteful.

sophy1029: Oh. I see. I will change it back to what we had before.

bangerth:
@sophy1029 -- do you know how to squash/fix-up the five commits you have here into one? If so, please do so!

PR #9000, opened 2019-11-04
Fix a bug with a combination Q2 x Nedelec element.

### LINKED ISSUE #8974: VectorTools::project_boundary_values_curl_conforming_l2 fails in 3d on FESystems containing FE_Nedelec and FE_Q
This is issue 2 described in detail in 
https://groups.google.com/forum/#!topic/dealii/taFA8q2ugsg

When trying to use VectorTools::project_boundary_values_curl_conforming_l2 to apply boundary values to the FE_Nedelec component of a FESystem containing also one FE_Q element the application fails with the error 

"Error: Unexpected number of 3D edge DoFs"

bangerth:
Might actually fix #8974. I don't understand the code well, I'm afraid. Let's see what the tests say.

/rebuild


masterleinad:
The indentation is off.

masterleinad:
```
360 - fe/nedelec_non_rect_face.debug (Failed)
533 - fe/fe_nedelec_sz_non_rect_face.debug (Failed)
882 - numerics/project_bv_curl_conf.debug (Failed)
911 - numerics/project_bv_curl_conf_complex.debug (Failed)
```

bangerth:
Yes, I found those too. More work necessary, but I guess that will take a couple of days to get to.

winnifried:
> Might actually fix #8974. 
I have written an additional test for #8974, unfortunately this fix currently breaks it earlier than dealii/master. Maybe the test is of interest?
[main.cc](https://github.com/dealii/dealii/files/3814542/main.txt)

winnifried:
I think I found the bug. Do you prefer it to be merged into bangerth:nedelec first or a new pull request directly with the current git as there have been more changes to the file vector_tools.templates.h in the meantime.

bangerth:
Let me try this.

bangerth:
@winnifried -- I've tried this out, but it still fails a number of tests:
```
        3588 - fe/fe_nedelec_singularity_02.mpirun=4.debug (Failed)
        3598 - fe/fe_nedelec_sz_non_rect_face.debug (Failed)
        3950 - fe/nedelec_non_rect_face.debug (Failed)
        8799 - numerics/project_bv_curl_conf.debug (Failed)
        8805 - numerics/project_bv_curl_conf_04.debug (Failed)
```
I did not look into why so far, but do these tests succeed for you?

I've had to resolve some merge conflicts by hand because your patch contained so many merge commits. It is possible that I made a mistake doing this (the results are now here in this PR), so it would be useful if you could try what you had on the tests mentioned above.

In any case, thanks already for the help!

winnifried:
@bangerth I checked numerics/project_bv_curl_conf.debug as it had failed in your previous version 
   and it worked with my installation. 
  numerics/project_bv_curl_conf_04.debug also worked; the output file came from my installation.

I will check the other examples tomorrow, and check whether I get different results after the merge.

winnifried:
@bangerth I have checked the code, it seems my fix was lost during the merge. I have created a new pull request to bangerth:nedelec with the necessary change.

bangerth:
Let's give this another spin:

/rebuild

bangerth:
This seems to test ok. Would someone be willing to review this patch?

kronbichler:
@bangerth do you still plan to squash some commits or should we merge as is?

include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -5440,38 +5454,69 @@ namespace VectorTools
...
-          if (((dynamic_cast<const FESystem<dim> *>(&fe) != nullptr) &&
-               (fe.system_to_base_index(cell_idx).first == base_indices) &&
-               (lower_bound <= fe.system_to_base_index(cell_idx).second) &&
-               (fe.system_to_base_index(cell_idx).second <= upper_bound)) ||
-              (((dynamic_cast<const FE_Nedelec<dim> *>(&fe) != nullptr) ||
-                (dynamic_cast<const FE_NedelecSZ<dim> *>(&fe) != nullptr)) &&
-               (line * (degree + 1) <= face_idx) &&
-               (face_idx <= (line + 1) * (degree + 1) - 1)))
+          Assert(cell->line_orientation(line),
+                 ExcMessage("Edge orientation doesnot meet expectation."));
```
agrayver: Does it mean the method itself cannot work on non-standard meshes? I thought there are some tests, which prove the opposite (it was the purpose of creating FE_NedelecSZ...). However, unless you have non-homogeneous BCs, this may still work by coincidence...

bangerth: Good question. I don't think any of us really understands how this function works any more (maybe @winnifried is the exception). I'm thinking that this patch fixes a real and known bug and that it is a step forward. I would advocate for merging it.

@agrayver -- first, many thanks for actually looking at this patch! If you think that this breaks something, do you think you could create a test that you think worked before and now doesn't any more?

bangerth:
@kronbichler -- I tried to rebase this PR, but it has unfortunately ingested a few cross-branch merges over its history and can no longer be rebased or squashed. It does seem to merge just fine here on github. Thus, I'd say let's just go with it as is.

bangerth:
(I did update the test @winnifried added to make it look more like the usual ones. It wrote to `std::cout` instead of `deallog` and didn't have a copyright header. But that is now fixed.)

include/deal.II/numerics/vector_tools.templates.h
agrayver: @bangerth yes, this sounds good -- I will try to look through existing tests and see if this situation is not covered.

bangerth: @agrayver -- I ran all tests that had 'curl' or 'nedelec' in their name, and they seemed to work.

bangerth:
The test failure seems spurious. @kronbichler?

bangerth:
Yay! Thanks, @winnifried, for the help with this, and my apologies for being slow with it :-(

PR #8900, opened 2019-10-11
Introduce VectorizedArrayBase and VectorizedArrayIterator

peterrum:
This is a first draft. ~~I am not completely satisfied especially due to the virtual functions, but hopefully someone has a better idea!~~

Closes #8896.

Ping @kronbichler @masterleinad

PR #9005, opened 2019-11-05
Allow change of SIMD type via template argument

jodlbauer:
This PR adds the SIMD type to the template argument list of the matrix-free operators and forwards it to the respective MatrixFree classes. Hence, one can create MF-Operators with different vector-widths (i.e. not always the maximum size via VectorizedArray<number>) to investigate the impact of vectorized operations.
The additional argument is added last in the template list with VectorizedArray<number> as default, which should not yield any incompatibilities.

As a side-effect, it may be possible to use SIMD classes other than the deal.II ones (although I somewhat doubt that ...)

### DIFF
--- a/doc/news/changes/minor/20191106DanielJodlbauer
+++ b/doc/news/changes/minor/20191106DanielJodlbauer
@@ -1,0 +1,3 @@
+New: template argument for type of vectorized array (vector width) in matrix-free operators.
+<br>
+(Daniel Jodlbauer, 2019/11/06)
--- a/include/deal.II/matrix_free/operators.h
+++ b/include/deal.II/matrix_free/operators.h
@@ -183,7 +183,9 @@ namespace MatrixFreeOperators
    * @author Denis Davydov, Daniel Arndt, Martin Kronbichler, 2016, 2017
    */
   template <int dim,
-            typename VectorType = LinearAlgebra::distributed::Vector<double>>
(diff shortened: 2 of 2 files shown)


peterrum:
@jodlbauer The test said:

> invalid author 'jodlbauer' without firstname and lastname

The name in your commit message is nor correct; it should be "Daniel Jodlbauer"

peterrum:
> As a side-effect, it may be possible to use SIMD classes other than the deal.II ones (although I somewhat doubt that ...)

If they support the same interface (`n_array_elements`,  `value_type`, ...), maybe...

jodlbauer:
Changed the author name now.




> If they support the same interface (`n_array_elements`, `value_type`, ...), maybe...

While on it, I could create a PR replacing `n_array_elements` by `size()`, if desired.

peterrum:
> While on it, I could create a PR replacing n_array_elements by size(), if desired.

We have already introduced std-like functionalities in PR #8893 and #8900.

jodlbauer:
I know, I was thinking more about replacing remaining occurrences of n_array_elements by size() in the deal.II sources.

masterleinad:
> I know, I was thinking more about replacing remaining occurrences of n_array_elements by size() in the deal.II sources.

Neither of these possibilities is deprecated so there is no need to do that.

PR #9011, opened 2019-11-06
Fix Boost.Python CMake

agrayver:
Apparently, it was a mistake to remove this line in #8968  -- we do need to null Boost_FOUND to get Boost.Python setup properly.


Rombur:
Merging this since we don't have any tester for python.

PR #9012, opened 2019-11-07
Fix test output for fullydistributed_grids/copy_serial_tria

masterleinad:
Fixes the newly failing tests in https://cdash.43-1.org/viewTest.php?onlydelta&buildid=2605. The test output is consistent across all regression testers. Related to #8982.


tjhei:
no reason to run the testers...

PR #8976, opened 2019-10-30
added stl read/write functions

nicola-giuliani:
I've added functions to read and write stl files using opencascades functions. Four tests check stl write to file, refinement using stl and NormalToMeshManifold and stl creation starting from iges files.

PR #9019, opened 2019-11-07
Fix Open Cascade incompatibility

dangars:
This is related to #8976 

There is an incompatibility with old cascade versions. I have opencascad 6.8. When I compile deal.II I get:
```
/opt/src/dealii/source/opencascade/utilities.cc:331:19: error: variable has incomplete type 'const void'
    const auto    error = writer.Write(shape_to_be_written, filename.c_str());  
                  ^                                                                     
/opt/src/dealii/source/opencascade/utilities.cc:332:26: error: use of undeclared identifier 'StlAPI_StatusOK'
    AssertThrow(error == StlAPI_StatusOK,
```
This PR fixes the incompatibility issues.

The problem is that `StlAPI_StatusOK` is defined in `StlAPI_ErrorStatus.hxx` which does not exit in opencascade 6.8

### DIFF
--- a/source/opencascade/utilities.cc
+++ b/source/opencascade/utilities.cc
@@ -328,9 +328,13 @@ namespace OpenCASCADE
       }
 
     StlAPI_Writer writer;
-    const auto    error = writer.Write(shape_to_be_written, filename.c_str());
+#  if (OCC_VERSION_MAJOR >= 7)
+    const auto error = writer.Write(shape_to_be_written, filename.c_str());
     AssertThrow(error == StlAPI_StatusOK,
                 ExcMessage("Error writing STL from shape."));
+#  else
+    writer.Write(shape_to_be_written, filename.c_str());
+#  endif
   }
 
   TopoDS_Shape


source/opencascade/utilities.cc
```diff
@@ -328,9 +328,13 @@ namespace OpenCASCADE
       }
 
     StlAPI_Writer writer;
-    const auto    error = writer.Write(shape_to_be_written, filename.c_str());
+#  if (OCC_VERSION_MAJOR >= 7)
```
tamiko: Is `OCC_VERSION_MAJOR` available for opencascade and the oce community edition?

dangars: I have the OpenCASCADE Community Edition CAE version 0.17 in my computer and in the detailed.log of deal.II I get.
```
OPENCASCADE_VERSION = 6.8.0
```
I checked and `OCC_VERSION_MAJOR` is available in the oce comunity edition library of my computer.

Do you know what is the difference between the OCE Comunity Edition and Open CASCADE Technology? Both libraries are available in Debian.

dangars:
#9026 fixes the issue
I close this PR

PR #9022, opened 2019-11-08
Do not hardcode the warp size

Rombur:
In a couple of places we have hardcoded the size of a warp to 32. This is only true for Nvidia GPUs, AMD gpus have a warp of 64 threads. In this PR we define a `warpsize` variable so we can change the size of the warp later on

Related to #7821

### DIFF
--- a/include/deal.II/base/cuda_size.h
+++ b/include/deal.II/base/cuda_size.h
@@ -33,6 +33,11 @@ namespace CUDAWrappers
    * changed depending on the architecture the code is running on.
    */
   constexpr int chunk_size = 1;
+
+  /**
+   * Define the number of threads in a warp.
+   */
+  constexpr int warp_size = 32;
 } // namespace CUDAWrappers
 
 DEAL_II_NAMESPACE_CLOSE
--- a/include/deal.II/lac/cuda_kernels.templates.h
+++ b/include/deal.II/lac/cuda_kernels.templates.h
@@ -191,45 +191,13 @@ namespace LinearAlgebra
 
       template <typename Number, typename Operation>
       __device__ void
-      reduce_within_warp(volatile Number *result_buffer, size_type local_idx)
-      {
-        if (block_size >= 64)
-          result_buffer[local_idx] =
-            Operation::reduction_op(result_buffer[local_idx],
-                                    result_buffer[local_idx + 32]);
-        if (block_size >= 32)
-          result_buffer[local_idx] =
-            Operation::reduction_op(result_buffer[local_idx],
-                                    result_buffer[local_idx + 16]);
-        if (block_size >= 16)
-          result_buffer[local_idx] =
-            Operation::reduction_op(result_buffer[local_idx],
-                                    result_buffer[local_idx + 8]);
-        if (block_size >= 8)
-          result_buffer[local_idx] =
-            Operation::reduction_op(result_buffer[local_idx],
-                                    result_buffer[local_idx + 4]);
-        if (block_size >= 4)
-          result_buffer[local_idx] =
-            Operation::reduction_op(result_buffer[local_idx],
-                                    result_buffer[local_idx + 2]);
-        if (block_size >= 2)
-          result_buffer[local_idx] =
-            Operation::reduction_op(result_buffer[local_idx],
-                                    result_buffer[local_idx + 1]);
-      }
-
-
-
-      template <typename Number, typename Operation>
-      __device__ void
       reduce(Number *        result,
              Number *        result_buffer,
              const size_type local_idx,
              const size_type global_idx,
              const size_type N)
       {
-        for (size_type s = block_size / 2; s > 32; s = s >> 1)
+        for (size_type s = block_size / 2; s > warp_size; s = s >> 1)
           {
             if (local_idx < s)
               result_buffer[local_idx] =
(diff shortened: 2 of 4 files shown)


drwells:
'threads in a warp' is not something I expected to see in this context but it is quite descriptive.

include/deal.II/matrix_free/cuda_matrix_free.h
```diff
@@ -602,13 +603,13 @@ namespace CUDAWrappers
...
-                     fe_degree==4 ? 4 :
+    return dim==2 ? (fe_degree==1 ? CUDAWrappers::warp_size :
+                     fe_degree==2 ? CUDAWrappers::warp_size/4 :
+                     fe_degree==3 ? CUDAWrappers::warp_size/8 :
+                     fe_degree==4 ? CUDAWrappers::warp_size/8 :
                      1) :
-           dim==3 ? (fe_degree==1 ? 8 :
-                     fe_degree==2 ? 2 :
+           dim==3 ? (fe_degree==1 ? CUDAWrappers::warp_size/4 :
+                     fe_degree==2 ? CUDAWrappers::warp_size/8 :
```
drwells: Should this be `warp_size/16` or was the original incorrect?

bangerth:
On 11/7/19 8:18 PM, David Wells wrote:
> 'threads in a warp' is not something I expected to see in this context but it 
> is quite descriptive.

Whichever computer scientist came up with this term had a knitter/weaver as a 
spouse!

include/deal.II/lac/cuda_kernels.templates.h
```diff
@@ -238,8 +206,15 @@ namespace LinearAlgebra
             __syncthreads();
           }
 
-        if (local_idx < 32)
-          reduce_within_warp<Number, Operation>(result_buffer, local_idx);
+        if (local_idx < warp_size)
```
peterrum: I am not sure if you want to replace the unrolled version with a loop... NVIDIA normally recommends to completely unroll it...

include/deal.II/base/cuda_size.h
```diff
@@ -33,6 +33,11 @@ namespace CUDAWrappers
    * changed depending on the architecture the code is running on.
    */
   constexpr int chunk_size = 1;
+
+  /**
+   * Define the number of threads in a warp.
+   */
+  constexpr int warp_size = 32;
```
peterrum: Can we move this to the cmake setup? We could configure it with some flags like `--D DEAL_II_WITH_AMD`...

include/deal.II/matrix_free/cuda_matrix_free.h
```diff
@@ -602,13 +603,13 @@ namespace CUDAWrappers
...
   {
     /* clang-format off */
-    return dim==2 ? (fe_degree==1 ? 32 :
-                     fe_degree==2 ? 8 :
-                     fe_degree==3 ? 4 :
-                     fe_degree==4 ? 4 :
+    return dim==2 ? (fe_degree==1 ? CUDAWrappers::warp_size :
+                     fe_degree==2 ? CUDAWrappers::warp_size/4 :
+                     fe_degree==3 ? CUDAWrappers::warp_size/8 :
+                     fe_degree==4 ? CUDAWrappers::warp_size/8 :
```
peterrum: Let't put here some comments:

```suggestion
    // we are trying to limit the number of threads according to the 
    // following formula:
    //    `threads = cells * (k+1)^d <= 4*CUDAWrappers::warp_size`
    return dim==2 ? (fe_degree==1 ? CUDAWrappers::warp_size :    // 128 <= 128 (for CUDA)
                     fe_degree==2 ? CUDAWrappers::warp_size/4 :  //  72 <= 128
                     fe_degree==3 ? CUDAWrappers::warp_size/8 :  //  64 <= 128
                     fe_degree==4 ? CUDAWrappers::warp_size/8 :  // 100 <= 128
```

Also, if @masterleinad and you have created an investigation regarding optimal number of cells maybe create an issue with those results and add a link here to that issue, so that we remember the reason for these (new?) numbers.

include/deal.II/base/cuda_size.h
Rombur: That's the plan but first I want to have everything working on AMD GPU and then I'll add the options in CMake.

include/deal.II/lac/cuda_kernels.templates.h
Rombur: Yeah but how do I unroll by hand if I don't know the size. The loop size is known at compile time so I would expect the compiler to be able to do it.

include/deal.II/matrix_free/cuda_matrix_free.h
Rombur: So we quickly looked at these numbers and they are OK but maybe not the best. @masterleinad was able to get a 10% increase of performance by tweaking the numbers but it didn't behave like we expected. So for now we stick with what we have.

include/deal.II/matrix_free/cuda_matrix_free.h
Rombur: These are the same numbers than before. I plan to revisit them when I start running real tests on AMD gpu.

include/deal.II/matrix_free/cuda_matrix_free.h
Rombur: Oops sorry. You are right it should be 16 not 8.

PR #9023, opened 2019-11-08
Extend cell python wrappers and add face wrappers

agrayver:
As it says, this extends existing cell wrappers and adds support for face iterators. 

It enables working with boundary ids, e.g.:

```
triangulation = dealii.Triangulation('2D')
triangulation.generate_hyper_cube()
triangulation.refine_global(1)

for cell in triangulation.active_cells():
    for face in cell.faces():
        if face.at_boundary():
            face.boundary_id = 1
            
triangulation.write('mesh.vtk', 'vtk') 
```

PR #9028, opened 2019-11-08
Make indent script work for python bindings

agrayver:
From discussion in #9023 

@bangerth FYI

### DIFF
--- a/contrib/.clang-format
+++ b/contrib/.clang-format
@@ -1,6 +1,0 @@
-#
-# The clang-format (Clang 6) style file used by deal.II.
-# This directory shall not be formatted.
-#
-
-DisableFormat: true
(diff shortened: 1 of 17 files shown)


bangerth:
Alternatively (and maybe easier), you can run your fixed indentation script and submit the changes in that directory as a separate patch that is merged before this one.

agrayver:
I suggest we postpone this until after #9023 is merged -- I will then update the indent of the entire bindings.

agrayver:
Coming back to this: what should I do? 

Remove contrib/.clang-format permanently and add indent changes to this branch?

masterleinad:
I am not opposed to just removing the file. We might have been a bit too restrictive when disallowing indenting in all these folders. On the other hand, it is a bit strange to have `python-bindings` in the `contrib` folder anyway, but this should not be a concern here.

agrayver:
Hm, I removed the file and ran indent script from the root directory, but it did not change a single file in the contrib folder. Any clue?

masterleinad:
> Hm, I removed the file and ran indent script from the root directory, but it did not change a single file in the contrib folder. Any clue?

Yes, you need to run `./contrib/utilities/indent-all` and also modify that file accordingly beforehand.

agrayver:
Ok, now this has worked!

Issue #8975, opened 2019-10-30
Failing build with Tpetra


winnifried:
> It is tested in deal v 9.1.1 - I could not test in the current deal git version since dealii/dealii/master fails to compile in debug mode at the moment.

Would you mind opening another issue for this? Our regression tests (https://cdash.43-1.org/index.php?project=deal.II) don't indicate any problem, though.

masterleinad:
Oh, I mean telling us the problem you are seeing.

winnifried:
I have configured the current deal.II git master branch   
(1d48c9c5d8c0d075d3b351ff2d94dcbd0833a8b1)
with p4est 2.0, petsc 3.9.4, scalapack 2.0.2, slepc 3.9.2, trilinos 12.12.1 
when compiling I get the following error 

In file included from [...]dealii-git-sources/source/numerics/vector_tools_integrate_difference.cc:22:0:
[...]dealii-git-build/source/numerics/vector_tools_integrate_difference.inst:31045:2: error: expected identifier before ‘&’ token

winnifried:
The compiler is the mpich wrapper for gcc 5.4.0. 
This setup worked well with the latest release.

masterleinad:
How do the lines around `dealii-git-build/source/numerics/vector_tools_integrate_difference.inst:31045:2` look like?

winnifried:
# ifdef DEAL_II_TRILINOS_WITH_TPETRA
 template void
 integrate_difference< 1 ,
 LinearAlgebra::TpetraWrappers::Vector<float>,
 Vector<float>,
  1 >(
 const DoFHandler< 1 ,  1 > &,
 const LinearAlgebra::TpetraWrappers::Vector<float>,
 &,
 const Function< 1 > &,
 Vector<float> &,
 const Quadrature< 1 > &,
 const NormType &,
 const Function< 1 > *,
 const double);
# endif

winnifried:
To me it looks like line 323 in sources/numerics/vector_tools_integrate_difference.inst.in creates this, and there a  `const LinearAlgebra::TpetraWrappers::Vector<float>, &,` should in fact be 
`const LinearAlgebra::TpetraWrappers::Vector<float> &,`
I will see if that is solves the issue and then open another pull request.

masterleinad:
Yes, that looks wrong.

winnifried:
[...]

PR #9030, opened 2019-11-09
configure trilinos: Ensure that we are linking against MPI for Tpetra configure check

tamiko:


tamiko:
In reference to #8975 - it turns out we never tested any Tpetra code path on the tester because this configure check always fails due to "DSO missing from command line", aka underlinkage.

PR #9032, opened 2019-11-10
Generalize ComputeIndexOwner::Dictionary

### LINKED ISSUE #9031: broken :ComputeIndexOwner::ConsensusAlgorithm if not all dofs are covered
The test /tests/mpi/parallel_vector_16.cc hangs since the new consensus algorithm is used to determine ghost ownership. The test is there to test 64 bit indices and has a vector where the first 2^32 indices are not available on either processor. The test (even if you run with 2 ranks hangs in):
```
#3  0x00007fffc8b899e9 in PMPI_Probe () from /usr/lib/libmpi.so.12
#4  0x00007ffff21d7a9c in dealii::Utilities::MPI::internal::ComputeIndexOwner::Dictionary::reinit (this=0x7fffffffb280, owned_indices=..., 
    comm=@0x9d90b8: 0x7fffc8de8680 <ompi_mpi_comm_world>) at ../include/deal.II/base/mpi_compute_index_owner_internal.h:216
#5  0x00007ffff21d6862 in dealii::Utilities::MPI::internal::ComputeIndexOwner::ConsensusAlgorithmPayload::ConsensusAlgorithmPayload (this=0x7fffffffb230, owned_indices=..., 
    indices_to_look_up=..., comm=@0x9d90b8: 0x7fffc8de8680 <ompi_mpi_comm_world>, owning_ranks=..., track_index_requests=<optimized out>)
    at ../include/deal.II/base/mpi_compute_index_owner_internal.h:410
#6  0x00007ffff2225816 in dealii::Utilities::MPI::Partitioner::set_ghost_indices (this=0x9d8f20, ghost_indices_in=..., larger_ghost_index_set=...)
    at ../source/base/partitioner.cc:223
#7  0x00007ffff2225299 in dealii::Utilities::MPI::Partitioner::Partitioner (this=0x9d8f20, locally_owned_indices=..., ghost_indices_in=..., communicator_in=<optimized out>)
    at ../source/base/partitioner.cc:75
#8  0x00007ffff1da3937 in dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host>::reinit (this=0x7fffffffccf0, locally_owned_indices=..., 
    ghost_indices=..., communicator=0x7fffb5f39008) at ../include/deal.II/lac/la_parallel_vector.templates.h:489
#9  0x00007ffff1da3894 in dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host>::Vector (this=0x7fffffffccf0, local_range=..., ghost_indices=..., 
    communicator=0x7fffb5f39008) at ../include/deal.II/lac/la_parallel_vector.templates.h:579
#10 0x000000000041069e in test () at /ssd/deal-git5/tests/mpi/parallel_vector_16.cc:55
#11 0x0000000000410d35 in main (argc=1, argv=0x7fffffffd3d8) at /ssd/deal-git5/tests/mpi/parallel_vector_16.cc:109
```
```
fr 3
#3  0x00007ffff21d7a9c in dealii::Utilities::MPI::internal::ComputeIndexOwner::Dictionary::reinit (this=0x7fffffffb280, owned_indices=..., comm=@0x9d6558: 0x7fffc8de8680 <ompi_mpi_comm_world>)
    at ../include/deal.II/base/mpi_compute_index_owner_internal.h:216
216                     auto ierr = MPI_Probe(MPI_ANY_SOURCE, tag_setup, comm, &status);
p dic_local_received
$3 = 84
p local_size
$4 = 2147483670
```

@kronbichler @peterrum ?

peterrum:
... now an assert is violated here:

https://github.com/dealii/dealii/blob/a3d367a339d46ba26ac0d165e8194175288734da/source/base/partitioner.cc#L339-L340

Any ideas about this one, @tjhei ? See also the new test below: I think the ghost owners and the targets are computed correctly.

Closes #9031

### DIFF
--- a/include/deal.II/base/mpi_compute_index_owner_internal.h
+++ b/include/deal.II/base/mpi_compute_index_owner_internal.h
@@ -35,6 +35,120 @@ namespace Utilities
       namespace ComputeIndexOwner
       {
         /**
+         * Specialization of ConsensusAlgorithmProcess for setting up the
+         * Dictionary even if there are ranges in the IndexSet space not owned
+         * by any processes.
+         *
+         * @note Only for internal usage.
+         */
+        class DictionaryPayLoad
+          : public ConsensusAlgorithmProcess<
+              std::pair<types::global_dof_index, types::global_dof_index>,
+              unsigned int>
+        {
+        public:
+          /**
+           * Constructor.
+           */
+          DictionaryPayLoad(
+            const std::map<unsigned int,
+                           std::vector<std::pair<types::global_dof_index,
+                                                 types::global_dof_index>>>
+              &                        buffers,
+            std::vector<unsigned int> &actually_owning_ranks,
+            const std::pair<types::global_dof_index, types::global_dof_index>
+              &                        local_range,
+            std::vector<unsigned int> &actually_owning_rank_list)
+            : buffers(buffers)
+            , actually_owning_ranks(actually_owning_ranks)
+            , local_range(local_range)
+            , actually_owning_rank_list(actually_owning_rank_list)
+          {}
+
+          /**
+           * Implementation of
+           * Utilities::MPI::ConsensusAlgorithmProcess::compute_targets().
+           */
+          virtual std::vector<unsigned int>
+          compute_targets() override
+          {
+            std::vector<unsigned int> targets;
+            for (const auto &rank_pair : buffers)
+              targets.push_back(rank_pair.first);
+
+            return targets;
+          }
+
+          /**
+           * Implementation of
+           * Utilities::MPI::ConsensusAlgorithmProcess::pack_recv_buffer().
+           */
+          virtual void
+          pack_recv_buffer(const int other_rank,
+                           std::vector<std::pair<types::global_dof_index,
+                                                 types::global_dof_index>>
+                             &send_buffer) override
+          {
+            send_buffer = this->buffers.at(other_rank);
+          }
+
+          /**
+           * Implementation of
+           * Utilities::MPI::ConsensusAlgorithmProcess::process_request().
+           */
+          virtual void
+          process_request(
+            const unsigned int                                     other_rank,
+            const std::vector<std::pair<types::global_dof_index,
+                                        types::global_dof_index>> &buffer_recv,
+            std::vector<unsigned int> &request_buffer) override
+          {
+            (void)request_buffer; // not needed
+
+
+            // process message: loop over all intervals
+            for (auto interval : buffer_recv)
+              {
+#ifdef DEBUG
+                for (types::global_dof_index i = interval.first;
+                     i < interval.second;
+                     i++)
+                  Assert(actually_owning_ranks[i - local_range.first] ==
+                           numbers::invalid_unsigned_int,
+                         ExcInternalError());
+                Assert(interval.first >= local_range.first &&
+                         interval.first < local_range.second,
+                       ExcInternalError());
+                Assert(interval.second > local_range.first &&
+                         interval.second <= local_range.second,
+                       ExcInternalError());
+#endif
+                std::fill(actually_owning_ranks.data() + interval.first -
+                            local_range.first,
+                          actually_owning_ranks.data() + interval.second -
+                            local_range.first,
+                          other_rank);
+              }
+            actually_owning_rank_list.push_back(other_rank);
+          }
+
+        private:
+          const std::map<unsigned int,
+                         std::vector<std::pair<types::global_dof_index,
+                                               types::global_dof_index>>>
+            &buffers;
+
+          std::vector<unsigned int> &actually_owning_ranks;
+
+          const std::pair<types::global_dof_index, types::global_dof_index>
+            &local_range;
+
+          std::vector<unsigned int> &actually_owning_rank_list;
+        };
+
+
+
+        /**
          * Dictionary class with basic partitioning in terms of a single
          * interval of fixed size known to all MPI ranks for two-stage index
          * lookup.
@@ -121,6 +235,10 @@ namespace Utilities
                                            types::global_dof_index>>>
               buffers;
 
+            std::fill(actually_owning_ranks.begin(),
+                      actually_owning_ranks.end(),
+                      numbers::invalid_subdomain_id);
+
             // 2) collect relevant processes and process local dict entries
             for (auto interval = owned_indices.begin_intervals();
                  interval != owned_indices.end_intervals();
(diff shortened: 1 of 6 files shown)


tjhei:
I have no idea. Hopefully @kronbichler can help... :-)

kronbichler:
Regarding the failing assertion: Can you try to post the violating indices, i.e., in which sense the two indices differ? The output is from a sanity check and even though I believe it looks right, it could be simply a defect in the test for non-onto maps that I do not see immediately.

And yes, it is embarrassing that we had this test failing the 64 bit configuration on the tester [1] for several months without noticing - I guess it is time to look at the regularly failing tests at some point and try to fix them so we notice failures like this more easily.

[1] https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=2771

kronbichler:
@peterrum can you explain how expensive you expect the new code to be, compared to the basic version we had before? In particular, do you think it would make sense to add a check of the form `if(Utilities::sum(locally_owned.n_elements(), comm) == locally_owned.size())` to keep the basic version and only otherwise do the more elaborate version? But if you believe it is easy compared to the rest of the index owner computation, I would go with the simpler variant.

peterrum:
> Regarding the failing assertion: Can you try to post the violating indices, i.e., in which sense the two indices differ? The output is from a sanity check and even though I believe it looks right, it could be simply a defect in the test for non-onto maps that I do not see immediately.

I suggest, you have a look at my (mini) test. The global size is 3:
- rank 0:  [1] / [2] (local/ghost)
- rank 2:  [2] / [1]  

It cannot be simpler...

peterrum:
> keep the basic version and only otherwise do the more elaborate version?

@kronbichler  I like your suggestion!

> @peterrum can you explain how expensive you expect the new code to be, compared to the basic version we had before?

Essentially the new implementation is a 3-way handshake (so in theory it should be 3 times slower - with an `IBarrier`). However, we should note that both implementations use somewhere the new `CollectiveMutex` (with two `Barriers`!) so that I  do not think there a big performance gap...

tjhei:
> However, we should note that both implementations use somewhere the new `CollectiveMutex` (with two `Barriers`!) so that I do not think there a big performance gap...

This workaround is hopefully only temporary.

I assume that the basic case is good for practically all applications, so I agree with Martin that we should think about not doing extra communication if we can easily avoid it.

peterrum:
Just as a note: `if(Utilities::sum(locally_owned.n_elements(), comm) == locally_owned.size())` also introduces a global communication...

tjhei:
> Just as a note: `if(Utilities::sum(locally_owned.n_elements(), comm) == locally_owned.size())` also introduces a global communication... 

You are correct.

> I would suggest we pass somewhere a bool that indicates if this case is possible or not.

Does the partitioner know if we cover the whole space? If not, it might make sense to store this information in the partitioner or the vector class.
But, I also wouldn't mind disallowing this situation altogether. I don't see any use case for it...

peterrum:
> But, I also wouldn't mind disallowing this situation altogether.

I am not in favor of this suggestion. I would like to keep the flexibility in the `ComputeIndexOwner` algorithm (I regard this independent of vector or partition). What I learnt in the last year is that you tend to only notice later on how you can use implementations in other contexts; and then you are happy!

peterrum:
I have added the if-statement. The new test passes if I disable the assert. Test `parallel_vector_16` still fails (segfault)...

kronbichler:
I have added a commit that fixes the assertion: There was still a bug in the partitioner where the indices are set. I have not had the time to look into `parallel_vector_16` yet, but the current code should be worth a try.

tests/mpi/compute_index_owner_01.cc
```diff
@@ -0,0 +1,101 @@
...
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+
+// check that cell->active_cell_index() works as advertised
```
kronbichler: Please update comment.

peterrum:
@kronbichler By the way, you have fixed `parallel_vector_16`! Thanks! I will rebase later on!

peterrum:
@tjhei Now, e.th. should work! Could you check if I have not deleted anything you have added recently (I had some merge conflicts). Nice collaboration @kronbichler!

I would suggest that we have a look on how to reduce the communication in this context (maybe move the info to the `Vector` or `Partitioner` class). Especially, I would like to introduce s.th. like: 
```cpp
            if (no_gaps_flag ||Utilities::MPI::sum(owned_indices.n_elements(), comm) ==
                owned_indices.size())
                  //  P2P 
               else
                  //  ConsenusAlgo
```

kronbichler:
> I would suggest we pass somewhere a bool that indicates if this case is possible or not.

I am not so sure about the bool - given how cheap an `MPI_Allreduce` is in this context compared to the other things we do in the partitioner setup, I would put this rather low on the priority list. Also, you do not know it at the level of the partitioner setup, and I am not a big fan of adding a bool parameter for a miniscule performance optimization to our interfaces to propagate it out of the partitioner and into the vector setup where you would know that an index set of locally owned dofs that comes from a `DoFHandler` is one-to-one. One could signal it to the index set, but then the index set has gained functionality that we want the partitioner to hold.

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -180,89 +276,109 @@ namespace Utilities
...
+                    // receive message
+                    Assert(number_amount % 2 == 0, ExcInternalError());
+                    std::vector<std::pair<types::global_dof_index,
+                                          types::global_dof_index>>
+                      buffer(number_amount / 2);
+                    ierr = MPI_Recv(buffer.data(),
+                                    number_amount,
+                                    DEAL_II_DOF_INDEX_MPI_TYPE,
+                                    other_rank,
+                                    mpi_tag,
```
kronbichler: IIRC these two lines were introduced by @tjhei, so I suggest you revert them.

tests/mpi/parallel_partitioner_09.cc
```diff
@@ -0,0 +1,53 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// test MPI::Partitioner setup for the case the owned indices are not onto,
+// i.e., we skip some index (related test: parallel_vector_16)
```
kronbichler:
```suggestion
// i.e., we skip some index (related tests: parallel_vector_16 and
// compute_index_owner_01)
```

peterrum:
@kronbichler Done! @tjhei Now it is your turn!

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -34,6 +34,98 @@ namespace Utilities
        */
       namespace ComputeIndexOwner
       {
+        class DictionaryPayLoad
```
tjhei: can you add a short doxygen comment here so that it appears in the documentation please (and fix the comment 4 lines up)?

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -180,89 +276,109 @@ namespace Utilities
...
+                    std::vector<std::pair<types::global_dof_index,
+                                          types::global_dof_index>>
+                      buffer(number_amount / 2);
+                    ierr = MPI_Recv(buffer.data(),
+                                    number_amount,
+                                    DEAL_II_DOF_INDEX_MPI_TYPE,
+                                    status.MPI_SOURCE,
+                                    status.MPI_TAG,
+                                    comm,
+                                    &status);
```
tjhei:
```suggestion
                                    MPI_STATUS_IGNORE);
```

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -180,89 +276,109 @@ namespace Utilities
...
-                                           rank_pair.first,
-                                           mpi_tag,
-                                           comm,
-                                           &request.back());
-                AssertThrowMPI(ierr);
-              }
-
-            // 4) receive messages until all dofs in dict are processed
-            while (this->local_size != dic_local_received)
+            if (Utilities::MPI::sum(owned_indices.n_elements(), comm) ==
```
tjhei: Please add a comment here that explains that we switch between two implementations based on the check if every element is owned by exactly one rank.

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -180,89 +276,109 @@ namespace Utilities
...
+                                  actually_owning_ranks.data() +
+                                    interval.second - local_range.first,
+                                  other_rank);
+                        dic_local_received += interval.second - interval.first;
+                      }
                   }
               }
+            else
+              {
+                // 3) send/recv messages with local dofs to the right dict
```
tjhei: this comment looks like it needs an update

source/base/partitioner.cc
```diff
@@ -199,7 +199,7 @@ namespace Utilities
       if (my_pid == 0)
         my_size += local_range_data.first;
 
-      types::global_dof_index my_shift = 0;
+      types::global_dof_index my_shift = local_range_data.first;
```
tjhei: this is to allow gaps between processors, right? please leave a comment.

peterrum: @kronbichler Could you explain how I should read this line? Why is the value `0` not correct? It is overridden in the following lines, isn't it? 

```cpp
      types::global_dof_index my_shift = local_range_data.first;
      {
        const int ierr = MPI_Exscan(&my_size,
                                    &my_shift,
                                    1,
                                    DEAL_II_DOF_INDEX_MPI_TYPE,
                                    MPI_SUM,
                                    communicator);
        AssertThrowMPI(ierr);
      }
```

kronbichler: `MPI_Exscan` computes the sum from the processor 0 up to (excluding) the current rank. Thus, on rank 0, we lack the knowledge on the shift.

include/deal.II/base/mpi_compute_index_owner_internal.h
```diff
@@ -180,89 +298,120 @@ namespace Utilities
...
-                                           DEAL_II_DOF_INDEX_MPI_TYPE,
-                                           rank_pair.first,
-                                           mpi_tag,
-                                           comm,
-                                           &request.back());
-                AssertThrowMPI(ierr);
-              }
+                // no gaps: setup is simple! Processes send their locally owned
+                // indices to the dictionary. The dictionary stores the sending
+                // rank for each index.The dictionary knows exactly
```
tjhei:
```suggestion
                // rank for each index. The dictionary knows exactly
```

source/base/partitioner.cc
tjhei: makes sense. @peterrum can you leave a comment that states this?

include/deal.II/base/mpi_compute_index_owner_internal.h
peterrum: Done!

source/base/partitioner.cc
peterrum: I have slightly rewritten the code. That MPI does not alter the value of rank 0 is as far I see it not documented. I would have expected it is set to zero, since it an output (and is the logical result of an exclusive scan). However, [this side](https://mpi.deino.net/mpi_functions/MPI_Exscan.html) states:

> The value in recvbuf on the process with rank 0 is undefined, and recvbuf is not signficant on process 0.

To be safe, I set the value on rank 0 **after** `MPI_Exscan`.

ping @tjhei and @kronbichler

source/base/partitioner.cc
```diff
@@ -194,12 +194,14 @@ namespace Utilities
...
-      // processors
+
+      // Allow non-zero start index for the vector. Part 1:
+      // Assume for now that the index set of rank 0 starts with 0
+      // and therefore has an increased size.
       if (my_pid == 0)
         my_size += local_range_data.first;
 
-      types::global_dof_index my_shift = 0;
+      types::global_dof_index my_shift;
```
kronbichler: I prefer that you keep the initialization to zero here to be on the safe side regarding reading undefined data when moving code around.

source/base/partitioner.cc
```diff
@@ -209,6 +211,15 @@ namespace Utilities
...
+
+      // Allow non-zero start index for the vector. Part 2:
+      // We correct the assumption made above and let the
+      // index set of rank 0 actually start from the
+      // correct value, i.e. we correct the shift to
+      // its start.
+      if (my_pid == 0)
+        my_shift = local_range_data.first;
+
       if (my_shift != local_range_data.first)
```
kronbichler: When we have gaps *inside* the index set, I believe this code is still wrong. The reason why we have this block is to fix the local range for the case we do not have any indices in the local range and hence reading a range from the associated index set is not possible. Otherwise, `local_range_data` should already be correct and the data desired by the user. To fix this, I suggest we write:
```
// Fix the index start in case the index set could not give us that information.
if (local_range_data.first == 0 && my_shift != 0)
```

peterrum: Looks good. Although, I am not really sure why support this gap (only) at the beginning of the vector...

peterrum:
1007 MPI tests have passed! Ready to merge @tjhei

source/base/partitioner.cc
kronbichler: The problem is that we make the assumption elsewhere in this class that the local range is contiguous. That could be relaxed, but it is not necessary in this PR.

tjhei: 1. Is there an Assert at the top for making sure the local range is contiguous? If not, we should add that.
2. @kronbichler , I think @peterrum is referring to the situation of having a gap between processors (with each processor having a contiguous range).

tjhei: 3. I am still not sure I understand this block of code. Is it for gaps between processors, the gap before proc 0, or for processors with 0 owned indices? The comment doesn't explain this.

Issue #8993, opened 2019-11-04
Assert for parallel::distributed::Triangulation and hp::DoFHandler


BenBrands:
I have created an example (extension of step-4) using both the hp::DoFHandler and the distributed Triangulation to illustrate the problem.

As soon as I use local refinement, the following assertion is thrown:

> An error occurred in line <1119> of file </home/benjamin/spack/var/spack/stage/dealii-develop-tddwd3d2gapfekuseackzafq6slbo6kk/spack-src/source/dofs/dof_tools_constraints.cc> in function
    void dealii::DoFTools::internal::make_hp_hanging_node_constraints(const DoFHandlerType&, dealii::AffineConstraints<number>&) [with DoFHandlerType = dealii::hp::DoFHandler<2, 2>; number = double]
The violated condition was: 
    cell->face(face)->child(c)->n_active_fe_indices() == 1
Additional information: 
    This exception -- which is used in many places in the library -- usually indicates that some condition which the author of the code thought must be satisfied at a certain point in an algorithm, is not fulfilled. An example would be that the first part of an algorithm sorts elements of an array in ascending order, and a second part of the algorithm later encounters an element that is not larger than the previous one.
There is usually not very much you can do if you encounter such an exception since it indicates an error in deal.II, not in your own program. Try to come up with the smallest possible program that still demonstrates the error and contact the deal.II mailing lists with it to obtain help.
/home/benjamin/spack/opt/spack/linux-linuxmint18-x86_64/gcc-7.2.0/dealii-develop-tddwd3d2gapfekuseackzafq6slbo6kk/lib/libdeal_II.g.so.9.2.0-pre: void dealii::DoFTools::internal::make_hp_hanging_node_constraints<dealii::hp::DoFHandler<2, 2>, double>(dealii::hp::DoFHandler<2, 2> const&, dealii::AffineConstraints<double>&)
/home/benjamin/spack/opt/spack/linux-linuxmint18-x86_64/gcc-7.2.0/dealii-develop-tddwd3d2gapfekuseackzafq6slbo6kk/lib/libdeal_II.g.so.9.2.0-pre: void dealii::DoFTools::make_hanging_node_constraints<dealii::hp::DoFHandler<2, 2>, double>(dealii::hp::DoFHandler<2, 2> const&, dealii::AffineConstraints<double>&)
./step-4: Step4<2>::setup_system()
./step-4: Step4<2>::run()
./step-4: main

[...]

PR #9036, opened 2019-11-10
Adding a test for hp::DoFHandler and parallel::distributed::Triangulation 

BenBrands:
Add a test derived from step-4 to check interplay of hp::DoFHandler and parallel::distributed::Triangulation for a locally refined mesh.
The test is inspired by #8993.

### DIFF
--- a/tests/mpi/hp_step-4.cc
+++ b/tests/mpi/hp_step-4.cc
@@ -1,0 +1,374 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// A lightly adapted version of the step-4 tutorial program. Compared
+// to step-4, this test uses hp::DoFHandler and
+// parallel::distributed::Triangulation. In the upper half of the domain
+// FENothing is used to save effort, as these elements would only be used later.
+// The test checks whether the hanging-node constraints are setup correctly.
+
+#include <deal.II/base/conditional_ostream.h>
+#include <deal.II/base/function.h>
+#include <deal.II/base/logstream.h>
+#include <deal.II/base/quadrature_lib.h>
+#include <deal.II/base/utilities.h>
+
+#include <deal.II/distributed/tria.h>
+
+#include <deal.II/dofs/dof_accessor.h>
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/fe/fe_nothing.h>
+#include <deal.II/fe/fe_q.h>
+#include <deal.II/fe/fe_values.h>
+
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/tria.h>
+#include <deal.II/grid/tria_accessor.h>
+#include <deal.II/grid/tria_iterator.h>
+
+#include <deal.II/hp/dof_handler.h>
+
+#include <deal.II/lac/sparsity_tools.h>
+#include <deal.II/lac/trilinos_precondition.h>
+#include <deal.II/lac/trilinos_solver.h>
+#include <deal.II/lac/trilinos_sparse_matrix.h>
+#include <deal.II/lac/trilinos_vector.h>
+
+#include <deal.II/numerics/data_out.h>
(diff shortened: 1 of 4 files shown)


BenBrands:
@bangerth and @marcfehling: As suggested!

tests/mpi/hp_step-4.cc
```diff
@@ -0,0 +1,352 @@
...
+
+    TrilinosWrappers::SolverDirect::AdditionalData add_data(false,
+                                                            "Amesos_Mumps");
+
+    TrilinosWrappers::SolverDirect direct_solver(solver_control, add_data);
+
+    TrilinosWrappers::MPI::Vector completely_distributed_solution(
+      locally_owned_dofs, communicator);
+
+    direct_solver.solve(system_matrix, completely_distributed_solution, rhs);
```
marcfehling: It would probably be a good idea to also check for the number of steps the solver takes. Depending on the version of Trilinos, this number may vary. We should thus check for a range of values.

There is a macro defined in `tests.h` that does exactly this and can be used like the following:
```
check_solver_within_range(direct_solver.solve(system_matrix,
                                              completely_distributed_solution,
                                              rhs),
                          solver_control.last_step(),
                          /*min_allowed=*/10,
                          /*max_allowed=*/80);
```

tests/mpi/hp_step-4.cc
```diff
@@ -0,0 +1,352 @@
...
+    locally_relevant_dofs.compress();
+
+    constraints.clear();
+
+    constraints.reinit(locally_relevant_dofs);
+
+    DoFTools::make_hanging_node_constraints(dof_handler, constraints);
+
+    DoFTools::make_zero_boundary_constraints(dof_handler, constraints);
+
```
marcfehling: You are working with `p::d::Triangulation` objects. We still don't know whether we will ever encounter a case in which constraints depend on degrees of freedom of ghost cells, which again are constrained to degrees of freedom of a different subdomains. We check for these 'nested' constraints in other tests. It would be a good idea to add this here, too!

Like this:
https://github.com/dealii/dealii/blob/22c4b6d5ead0a0afcc361bbdb7d46d23e72bbc74/tests/mpi/trilinos_step-27.cc#L232-L244

marcfehling:
@BenBrands -- ping

BenBrands:
I will have a look by the end of the week!

tests/mpi/hp_step-4.cc
```diff
@@ -0,0 +1,352 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2009 - 2018 by the deal.II authors
```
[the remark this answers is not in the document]
BenBrands: Done!

tests/mpi/hp_step-4.cc
```diff
@@ -0,0 +1,352 @@
...
+} // namespace Step4
+
+
+
+int
+main(int argc, char *argv[])
+{
+  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+
+  Step4::Step4<2> laplace_problem_2d;
```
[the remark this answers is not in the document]
BenBrands: Done!

tests/mpi/hp_step-4.cc
BenBrands: Is that necessary for the direct solver?

tests/mpi/hp_step-4.cc
BenBrands: Done!

BenBrands:
@marcfehling -- Please have a look!

tests/mpi/hp_step-4.cc
```diff
@@ -0,0 +1,374 @@
...
+
+#ifdef DEBUG
+    // We did not think about hp constraints on ghost cells yet.
+    // Thus, we are content with verifying their consistency for now.
+    IndexSet locally_active_dofs;
+    DoFTools::extract_locally_active_dofs(dof_handler, locally_active_dofs);
+    AssertThrow(constraints.is_consistent_in_parallel(
+                  dof_handler.locally_owned_dofs_per_processor(),
+                  locally_active_dofs,
+				  communicator,
```
marcfehling: The code indentation of this particular line looks odd. Please run the indentation script as described [here](https://www.dealii.org/current/doxygen/deal.II/CodingConventions.html). Otherwise testers of the CI will complain.

tests/mpi/hp_step-4.cc
marcfehling: You are correct, this only makes sense for iterative solvers. I got so used to iterative solvers and the `check_solver_within_range` function that I didn't even think of this :)

tests/mpi/hp_step-4.cc
marcfehling: I've just noticed that the indentation check of `travis-ci` passed. Hmm, that's strange, but I guess we are good as long as the tester does not complain...

BenBrands:
I reran the indentation script!

PR #9038, opened 2019-11-11
Make variable volatile to avoid incorrect optimization

Rombur:
See https://devtalk.nvidia.com/default/topic/965588/race-condition-within-warp/


Rombur:
> This seems reasonable to me, although of course I don't understand much about CUDA. Is this really the only place in the code base where this is necessary?

I think so. As far as I can tell this is the only place we are working at the warp level.

PR #9039, opened 2019-11-11
Add more choices to QuadratureSelector

jodlbauer:
Extends the class QuadratureSelector by some more choices.

Some points to discuss (probably):

- So far, I added all quadratures that could be constructed using an unsigned int argument, i.e. including those with default-valued additional arguments. Not sure if those should be selectable or not.
- Removed checks for valid order, this is done anyway by the called quadrature
- Naming of flags

PS: I think I screwed up something before and accidentally closed the previous PR #9037, but couldn't reopen it again ... . So here it is again.

### DIFF
--- a/doc/news/changes/minor/20191111DanielJodlbauer
+++ b/doc/news/changes/minor/20191111DanielJodlbauer
@@ -1,0 +1,3 @@
+New: add more quadrature choices to QuadratureSelector class.
+<br>
+(Daniel Jodlbauer, 2019/11/11)
--- a/source/base/quadrature_selector.cc
+++ b/source/base/quadrature_selector.cc
@@ -25,10 +25,15 @@ QuadratureSelector<dim>::create_quadrature(const std::string &s,
                                            const unsigned int order)
 {
   if (s == "gauss")
-    {
-      AssertThrow(order >= 1, ExcInvalidQGaussOrder(order));
-      return QGauss<dim>(order);
-    }
+    return QGauss<dim>(order);
+  else if (s == "gauss_lobatto")
+    return QGaussLobatto<dim>(order);
+  else if (s == "gauss_chebyshev")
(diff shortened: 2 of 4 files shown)


masterleinad:
Thanks! It would be great if you could extend tests/base`/quadrature_selector.cc` for checking that the new choices work as expected.

jodlbauer:
Added some tests, but haven't run them so far.

drwells:
This looks good.

I can think of one quick improvement, though: could you add something like
```cpp
  Assert(typeid(QuadratureSelector<dim>(name, order).get_points()) 
         == typeid(q.get_points()),
         ExcInternalError());
```
to the test to verify that we really do create the right type?

jodlbauer:
Had to remove the QGaussLog/R choices, as those quadratures only exist for dim = 1.

jodlbauer:
`typeid(QuadratureSelector<dim>(name, order)) == typeid(q)`

will not work, as the left one is still of type QuadratureSelector, but q is QGauss, ... etc.

masterleinad:
You should be able to tets via `dynamic_cast`, though.

jodlbauer:
The problem I see is that after construction, the selector class has no info about which quadrature it corresponds to. The returned quadrature class is only used to call the base Quadrature constructor with resp. points and weights.

drwells:
You're right: this isn't possible to implement.

drwells:
@masterleinad are you happy? If so lets merge :)

bangerth:
Oh, I should have waited for @masterleinad , I guess :-( 

@masterleinad -- if you have more concerns here, then let me know and either I or @jodlbauer will fix them up!

masterleinad:
That's fine with me. I was looking at `QuadratureSelector::create_quadrature` for the `dynamic_cast` check but just realized that this is `private`. The test here should be good enough.

PR #9040, opened 2019-11-11
added stl read/write functions with compatibility 

nicola-giuliani:
post comments from wolfgang

added changes

added IncrementalMesh


nicola-giuliani:
@tamiko , @luca-heltai , @bangerth , @masterleinad . I hope that this solves the compatibility issues, I changed the backend opencascade function and added a check on the version

source/opencascade/utilities.cc
```diff
@@ -267,6 +275,66 @@ namespace OpenCASCADE
...
+    std::vector<TopoDS_Edge>   edges;
+    std::vector<TopoDS_Face>   faces;
+    OpenCASCADE::extract_geometrical_shapes(shape, faces, edges, vertices);
+    bool mesh_is_present = true;
+    for (unsigned int i = 0; i < faces.size(); ++i)
+      {
+        Handle(Poly_Triangulation) theTriangulation =
+          BRep_Tool::Triangulation(faces[i], Loc);
+        if (theTriangulation.IsNull())
+          mesh_is_present = false;
```
luca-heltai: can you add a break after this? You don't need to check anymore faces if there is at least one face which is not meshed.

nicola-giuliani: done

tests/opencascade/stl_write.cc
```diff
@@ -0,0 +1,46 @@
...
+  TopoDS_Shape sh = read_STL(SOURCE_DIR "/stl_files/sphere_refined.stl");
+  write_STL(sh, "tmp.stl", 0.001, false, 1e-6, false, 0.001, false);
+  std::ifstream in("tmp.stl");
+  std::ofstream out("output");
+  std::string   line;
+  unsigned int  counter = 5;
+  while (counter--)
+    std::getline(in, line);
+  while (std::getline(in, line))
+    out << line << std::endl;
```
bangerth: Use the `cat_file` function in `tests.h` :-)

tests/opencascade/stl_refinement.cc
```diff
@@ -0,0 +1,55 @@
...
+  OpenCASCADE::NormalToMeshProjectionManifold<2, 3> manifold(sh, 1e-3);
+  // CachedNormalToMeshProjectionManifold manifold(shape,1e-3);
+  tria.set_all_manifold_ids(0);
+  tria.set_manifold(0, manifold);
+  tria.refine_global(2);
+
+  GridOut       gridout;
+  std::ofstream ofile("output");
+  gridout.write_ucd(tria, ofile);
+  return 0;
```
bangerth: no need for this line. same in the other tests

tests/opencascade/stl_creation_watertight.cc
```diff
@@ -0,0 +1,38 @@
...
+
+int
+main()
+{
+  const std::string closed_iges       = SOURCE_DIR "/iges_files/wing.iges";
+  const std::string output_closed_stl = "output";
+  auto              shape_fuse = OpenCASCADE::read_IGES(closed_iges, 1e-2);
+  write_STL(shape_fuse, output_closed_stl, 0.1, true, 1e-6, false, 0.1, false);
+
+  return 0;
```
bangerth:
```suggestion
```

source/opencascade/utilities.cc
```diff
@@ -267,6 +275,69 @@ namespace OpenCASCADE
...
+        if (sew_different_faces)
+          {
+            BRepBuilderAPI_Sewing sewer(sewer_tolerance);
+            sewer.Add(shape_to_be_written);
+            sewer.Perform();
+            shape_to_be_written = sewer.SewedShape();
+          }
+        else
+          shape_to_be_written = shape;
+        BRepMesh_IncrementalMesh mesh_im(shape_to_be_written,
```
nicola-giuliani: @bangerth @luca-heltai  This is the major change w.r.t. the previous version . This routine is still in OCC 7.4.0 while the one we used before disappeared in 7.2.0. It has more arguments so I left the possibility for the deal user to set them all providing the default arguments that OCC 7.4.0 suggests.

tests/opencascade/stl_creation_single_patch.cc
```diff
@@ -0,0 +1,37 @@
...
+using namespace OpenCASCADE;
+
+int
+main()
+{
+  const std::string open_iges =
+    SOURCE_DIR "/iges_files/DTMB-5415_bulbous_bow.iges";
+  const std::string output_open_stl = "output";
+  auto              shape_bow       = OpenCASCADE::read_IGES(open_iges, 1e-3);
+  write_STL(shape_bow, output_open_stl, 0.5, false, 1e-6, false, 0.5, false);
```
nicola-giuliani: @bangerth this now results in 3500 lines

source/opencascade/utilities.cc
```diff
@@ -267,6 +275,70 @@ namespace OpenCASCADE
...
+                                         deflection,
+                                         is_relative,
+                                         angular_deflection,
+                                         in_parallel);
+      }
+
+    StlAPI_Writer writer;
+    const auto    error = writer.Write(shape_to_be_written, filename.c_str());
+#  if ((OCC_VERSION_MAJOR * 100 + OCC_VERSION_MINOR * 10) >= 690)
```
nicola-giuliani: @bangerth @luca-heltai  this is the other change, following the discussion of last week we should check the version and this way seems much more general than the first one I uploaded yesterday.

source/opencascade/utilities.cc
```diff
@@ -267,6 +275,70 @@ namespace OpenCASCADE
...
+            BRepBuilderAPI_Sewing sewer(sewer_tolerance);
+            sewer.Add(shape_to_be_written);
+            sewer.Perform();
+            shape_to_be_written = sewer.SewedShape();
+          }
+        else
+          shape_to_be_written = shape;
+        BRepMesh_IncrementalMesh mesh_im(shape_to_be_written,
+                                         deflection,
+                                         is_relative,
```
bangerth: I am confused by this statement. You are creating an object of type `BRepMesh_IncrementalMesh` here in line 327, but this is the last statement in the block -- the next line after the creation of the object is a closing brace. So the object is dead -- it is destroyed again immediately after being created. What is it there for?

tamiko:
@nicola-giuliani Thanks a lot for giving this another shot! Just to make sure that everything is alright, I will let the tester run on this very pull request - we'll now in about 2h whether the tester is happy!

tamiko:
We still have a build error: https://cdash.43-1.org/viewBuildError.php?buildid=2801

nicola-giuliani:
Which OCC version Is the tester running?

source/opencascade/utilities.cc
nicola-giuliani: Occ automatically calls a perform method that creates the triangulation in the shape.

masterleinad:
> Which OCC version Is the tester running?

7.3.0

source/opencascade/utilities.cc
bangerth: So one of the constructor arguments here is an output argument? Can you please add a summary of this discussion as a comment so that it's clear to the next reader how this work? -- Thanks!

nicola-giuliani:
Appearently from 7.3.0 the write function simply returns a booleana which should be the error state. So either we don't chech the exit or se assume that 0 should be returns and that 1 triggers the asserthrow. Let me know and tomorrow i Will modify.

bangerth:
Can you just split the `#ifdef`s further so that you can check for both cases?

source/opencascade/utilities.cc
nicola-giuliani: Ok, done

nicola-giuliani:
Updated the ifdef part, commented the strange OCC line. I think we are ready for another tester trial.

bangerth:
This looks good to me. @tamiko -- any objections to merging?

tamiko:
Looking good: https://cdash.43-1.org/buildSummary.php?buildid=2822

tamiko:
@nicola-giuliani @luca-heltai *Umh* It's me again :-D

We fail with opencascade 6.8.0 [1]. This looks like a simple header file missing :-) I will investigate a bit.

[1] https://cdash.43-1.org/viewBuildError.php?buildid=2824

masterleinad:
`opencascade/stl_creation_watertight`, `stl_creation_single_patch` are failing on all configurations. Can you please have a look?

nicola-giuliani:
@masterleinad which branch? master?

masterleinad:
Yes, see https://cdash.43-1.org/testSummary.php?project=1&name=opencascade%2Fstl_creation_single_patch.debug&date=2019-12-09 and https://cdash.43-1.org/testSummary.php?project=1&name=opencascade%2Fstl_creation_watertight.debug&date=2019-12-09.

luca-heltai:
We have no way of reproducing. It works on all my machines. What version of opencascade is this with?
it's possible that we should simply add another version of the output files. Does @tamiko have access to those outputs? Can you share them here? They are STL files afterall...

masterleinad:
> What version of opencascade is this with?

https://cdash.43-1.org/buildSummary.php?buildid=3233 says it's version 7.3.0.

PR #9045, opened 2019-11-12
Add airfoil grid generator

eliasstudiert:
Resulting (2D) triangulations looks like this:

NACA2412


peterrum:
This PR is part of @eliasstudiert's Bachelor's thesis. Hope you like the new grid generator! It is currently limited to 2D and does not use manifolds. The documentation still has to be extended...

A first feedback is welcome!

drwells:
This is a nice picture and definitely something worth adding to the library!

@eliasstudiert do you have any plans to extend this to use manifolds?

eliasstudiert:
> This is a nice picture and definitely something worth adding to the library!
> 
> @eliasstudiert do you have any plans to extend this to use manifolds?

Thanks! I didn't think about how to incorporate Manifolds yet, but I'd be thankful for any ideas or hints.

blaisb:
@eliasstudiert this will be very useful for us! Nice addition.
It would be very interesting if you could add manifolds, I agree with that :)!

drwells:
@eliasstudiert I'll have to take a closer look, but since you have some description of the geometry already embedded in he source code (so that we can generate the grid around the airfoil) it should be possible to generalize that as a manifold.

bangerth:
Nice work! I only know for the NACA0012 airfoil, but for that the surface is given as an analytically known function that should be relatively easy to encode in a manifold. One would then use the `TransfiniteInterpolationManifold` class for the remainder of the mesh.

@eliasstudiert -- there are a number of TODOs or unfinished sentences still left in the documentation in the .h file. If you wanted to go through those first, we can save one round of reviews and the whole process is likely going to be a bit faster than if we have to mark all of these places up. Can you go through your text again and let us know when these are addressed?

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,173 @@ namespace GridGenerator
...
+
+      /**
+       * This Function gets the entry values out of the
+       * ParameterHandler object and assigns those to the corresponding
+       * parameters in class AdditionalData.
+       *
+       * @param prm Parameter handler.
+       */
+      void
+      get(ParameterHandler &prm);
```
luca-heltai: Is there a reason not to use the `add_parameter` feature of `ParameterHandler`? This would allow you to attach a parameter right at the time of the declaration, and would remove the above `get` function (simplifying the code, and removing the possibility of bugs when adding/removing parameters in future versions).

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1175 @@
...
+
+    void
+    AdditionalData::declaration(ParameterHandler &prm)
+    {
+      prm.enter_subsection("FarField");
+      {
+        prm.declare_entry("Height",
+                          std::to_string(height),
+                          Patterns::Double(),
+                          "CoarseGridBlockHight");
```
luca-heltai: For example, this would become:

`prm.add_entry("Height", height, "CoarseGridBlockHight");`

and you could remove the corresponding entry in the `get` function.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1175 @@
...
+        prm.declare_entry("LengthB2",
+                          std::to_string(length_b2),
+                          Patterns::Double(),
+                          "LengthBlock2");
```
luca-heltai: same here. And the same in all the other cases.

eliasstudiert:
> Nice work! I only know for the NACA0012 airfoil, but for that the surface is given as an analytically known function that should be relatively easy to encode in a manifold. One would then use the `TransfiniteInterpolationManifold` class for the remainder of the mesh.
> 
> @eliasstudiert -- there are a number of TODOs or unfinished sentences still left in the documentation in the .h file. If you wanted to go through those first, we can save one round of reviews and the whole process is likely going to be a bit faster than if we have to mark all of these places up. Can you go through your text again and let us know when these are addressed?

source/grid/grid_generator.cc
eliasstudiert: I removed function get() in AdditionalData as you suggested and used add_parameter() to attach the parameters instead. 
Somehow, the function `create_triangulation(tria)` and `create_triangulation(tria, periodic_faces)` which should produce a default mesh since no `AdditionalData data` object is passed as input, doesn't work without the inputfile (parse_input(inputfile)).

Is it possible that this change (using add_paramter instead declare_entry() and get()) caused that issue?

include/deal.II/grid/grid_generator.h
eliasstudiert: > Nice work! I only know for the NACA0012 airfoil, but for that the surface is given as an analytically known function that should be relatively easy to encode in a manifold. One would then use the `TransfiniteInterpolationManifold` class for the remainder of the mesh.
> 
> @eliasstudiert -- there are a number of TODOs or unfinished sentences still left in the documentation in the .h file. If you wanted to go through those first, we can save one round of reviews and the whole process is likely going to be a bit faster than if we have to mark all of these places up. Can you go through your text again and let us know when these are addressed?

I added the documentation and finished the TODOs. Also I simplified the function interpolate() in taking advantage of symmetry of the coarse grid. The code is ready to review. I didn't figure out so far, why `create_triangulation()` without `AdditionalData data` as inputparameter (default setup) doesn't work without providing the inputfile.

eliasstudiert:
@bangerth I added the documentation and finished the TODOs. Also I simplified the function interpolate() in taking advantage of symmetry of the coarse grid. The code is ready to review. I didn't figure out so far, why create_triangulation() without AdditionalData data as inputparameter (default setup) doesn't work without providing the inputfile.

drwells:
@eliasstudiert I don't think you meant to close this: we can keep it open and mark as WIP until its ready :)

source/grid/grid_generator.cc
```diff
@@ -13,10 +13,12 @@
 //
 // ---------------------------------------------------------------------
 
+#include <deal.II/distributed/fully_distributed_tria.h>
 #include <deal.II/distributed/shared_tria.h>
 #include <deal.II/distributed/tria.h>
 
 #include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/grid_out.h>
```
peterrum: This is not needed anymore.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+      }
+
+      /**
+       * Calculate airfoil points for 4-digit NACA airfoils according to
+       * following reference literature
+       *
+       * I.H. Abbott and A.E. von Doenhoff. Theory of Wing Sections: Including a
+       * Summary of Airfoil Data. New York: Dover Publications, 1959.
+       *
+       * @param NACA-serialnumber for different airfoilshapes (so far only 4-digit-series implemented)
```
peterrum:
```suggestion
       * @param NACA-serial number for different airfoil shapes (so far only 4-digit-series implemented)
```
the same below

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+        // conversion string (char * ) to int
+        const unsigned int digit_0 = int(serialnumber[0]) - 48;
+        const unsigned int digit_1 = int(serialnumber[1]) - 48;
+        const unsigned int digit_2 = int(serialnumber[2]) - 48;
+        const unsigned int digit_3 = int(serialnumber[3]) - 48;
+
+        const unsigned int digit_23 = 10 * digit_2 + digit_3;
+
+        const double t = static_cast<double>(digit_23) / 100.0;
+        // 1.0 * digit_23 / 100; // maximum thickness in percentag of the cord
```
peterrum: Move comment above.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+        double x, y_t;
```
peterrum: You can move these into the loops below and make them const.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+        return naca_points;
+      }
+
+
+
+      /**
+       * Set airfoil length (i.e. chord length) to a desired length.
+       * calls function set_airfoil_length() for each vector of the array
+       *
+       * @param array containing upper and lower vector
```
peterrum: I just noticed. The format is:
```suggestion
       * @param input Array containing upper and lower vector.
```
Please adjust the parameters in the other functions similarly.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+
+
+        // loop over all subsections
+        for (unsigned int j = 0, i = 1; j < n_points - 1; j++)
+          {
+            // get reference left and right end of this section
+            const auto Lj  = arclength_L[j];
+            const auto Ljp = arclength_L[j + 1];
+
+
```
peterrum: You can remove this line.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+        // calculate arclength
+        std::vector<double> arclength_L(non_equidistant_points.size(), 0);
+        for (unsigned int i = 0; i < non_equidistant_points.size() - 1; i++)
+          arclength_L[i + 1] =
+            arclength_L[i] +
+            non_equidistant_points[i + 1].distance(non_equidistant_points[i]);
+
+
+        double airfoil_length =
```
peterrum:
```suggestion
        const auto airfoil_length =
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+          arclength_L.back(); // arclength upper or lower side
+        double deltaX = airfoil_length / (number_points - 1);
```
peterrum:
```suggestion
        const auto deltaX = airfoil_length / (number_points - 1);
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+                double l = i * deltaX;
+
+                // perform linear interpolation
+                equidist[i] = Point<2>((l - Lj) / (Ljp - Lj) *
+                                         (non_equidistant_points[j + 1] -
+                                          non_equidistant_points[j]) +
+                                       non_equidistant_points[j]);
+                i++;
+              }
+          }
```
peterrum:
```suggestion
            while (Lj <= i * deltaX && i * deltaX <= Ljp &&
                   i < number_points - 1)
                equidist[i++] = Point<2>((i * deltaX - Lj) / (Ljp - Lj) *
                                           (non_equidistant_points[j + 1] -
                                            non_equidistant_points[j]) +
                                         non_equidistant_points[j]);
          }
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+
+
+
+      /**
+       * Create the coarse grid.
+       * Initializes a given triangulation with a coarse grid.
+       * Create 6 coarse grids based on points A-L (class fields) and merges
+       * them to one triangulation.
+       *
+       * @param tria
```
peterrum: Just remove this line.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+       * and set the boundary_ids depending on the material_id of the cell and
+       * the face number. The resulting boundary_ids are:
+       * - 0: inlet
+       * - 1: outlet
+       * - 2: upper airfoil surface (aka. suction side)
+       * - 3, lower airfoil surface (aka. pressure side),
+       * - 4: upper far-field side
+       * - 5: lower far-field side
+       *
+       * @param tria
```
peterrum: Just remove this line.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+            }
+      }
+
+
+      /*
+       * This function returns a bias factor 'alpha' which is used to make the
+       * mesh more tight in close distance of the airfoil.
+       *
+       * @param alpha
+       * @return alpha
```
peterrum: Just remove these lines.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+              else
+                {
+                  Assert(false, ExcIndexRange(cell.material_id(), 1, 6));
+                }
+            }
+      }
+
+
+      /*
+       * This function returns a bias factor 'alpha' which is used to make the
```
peterrum: You could mention that this is bijective function mapping from [0,1] onto [0,1]. Are values near 0 or near 1 made tighter?

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+                  // indices for interpolation (all x-nodes are positiv)
+                  // Move trapeze to be in first quadrant by addig
+                  // trapeze_offset
+                  Point<2, double> node_;
+                  if (cell.material_id() == 1)
+                    {
+                      rotation_matrix_1[0][0] = +std::cos(-gamma);
+                      rotation_matrix_1[0][1] = -std::sin(-gamma);
+                      rotation_matrix_1[1][0] = +std::sin(-gamma);
+                      rotation_matrix_1[1][1] = +std::cos(-gamma);
```
peterrum: You should setup the matrix only once at the top.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+                    }
+                  // step 1: rotate block 4 counterclockwise and move down so
+                  // that trapeze is located in fourth quadrant (subtracting
+                  // trapeze_offset)
+                  else if (cell.material_id() == 4)
+                    {
+                      rotation_matrix_1[0][0] = +std::cos(+gamma);
+                      rotation_matrix_1[0][1] = -std::sin(+gamma);
+                      rotation_matrix_1[1][0] = +std::sin(+gamma);
+                      rotation_matrix_1[1][1] = +std::cos(+gamma);
```
peterrum: The same here. You can save it into a new `rotation_matrix_2`.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+      void interpolate(Triangulation<2> &tria) const
+      {
+        // array storing the information if a vertex was processed
+        std::vector<bool> vertex_processed(tria.n_vertices(), false);
+
+        // rotation matrix for clockwise rotation of block 1 by angle gamma
+        Tensor<2, 2, double> rotation_matrix_1;
+
+        // horizontal offset in order to place coarse-grid node A in the origin
+        Point<2, double> horizontal_offset(A(0), 0.0);
```
peterrum:
```suggestion
        const Point<2, double> horizontal_offset(A(0), 0.0);
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+
+        // Move block 1 so that face BC coincides the x-axis
+        Point<2, double> trapeze_offset(0.0, std::sin(gamma) * edge_length);
```
peterrum:
```suggestion
        const Point<2, double> trapeze_offset(0.0, std::sin(gamma) * edge_length);
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+
+        tria_grid.refine_global(refinements);
+        this->interpolate(tria_grid);
+      }
+
+      /**
+       * Create a parallel::distributed::Triangulation.
+       */
+      void create_triangulation(
+        parallel::distributed::Triangulation<2> &parallel_grid,
```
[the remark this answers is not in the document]
eliasstudiert: so we can remove  function `void create_triangulation(parallel::distributed::Triangulation<2> &parallel_grid, ...)` since `void create_triangulation(Triangulation<2> &tria_grid, ...)` will work for serial and distributed triangulations, right?

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+        mesh_generator.create_triangulation(*parallel_tria, nullptr);
+      else if (auto parallel_tria = dynamic_cast<
+                 dealii::parallel::fullydistributed::Triangulation<2, 2> *>(
+                 &tria))
+        mesh_generator.create_triangulation(*parallel_tria, nullptr);
+      else
+        mesh_generator.create_triangulation(tria, nullptr);
+      // Assert(false, ExcMessage("Mesh has been created successfully!"));
+      std::cout << "Mesh has been created successfully!" << std::endl;
+    }
```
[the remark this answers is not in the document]
eliasstudiert: I changed this function as you suggested, but there is a compiling error which says no matching function to call, since 

no known conversion for argument 2 from ‘std::nullptr_t’ to ‘std::vector<dealii::GridTools::PeriodicFacePair<dealii::TriaIterator<dealii::CellAccessor<2, 2> > >, std::allocator<dealii::GridTools::PeriodicFacePair<dealii::TriaIterator<dealii::CellAccessor<2, 2> > > > >&’

How could I fix this ?

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+
+      // y-coordinate of last airfoil point in airfoil_1D[0] and airfoil_1D[1]
+      const double tail_y;
+
+      // x-coordinate of C,D,E,F indicating ending of blocks 1 and 4 or
+      // beginning of blocks 2 and 5, respectively
+      const double center_mesh;
+
+      // length of blocks 2 and 5
+      const double length_b1_x;
```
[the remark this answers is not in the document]
eliasstudiert: to be conform with `n_subdivision_x_2`. I use numbers 0 (=Blocks 1&4), 1 (=Blocks 2&5) and 2 (=Blocks 3&6). Should I use a different approach?

source/grid/grid_generator.cc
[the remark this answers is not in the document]
peterrum: I think it will work. You should try it out.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+       * @param circle points
+       * @return vector containing joukowski airfoil points
+       */
+      static std::vector<Point<2>>
+      joukowski_transform(const std::vector<Point<2>> &circle_points)
+      {
+        std::vector<Point<2>> joukowski_points(circle_points.size());
+
+        std::complex<double> z;
+        double               chi, eta;
```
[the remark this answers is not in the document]
eliasstudiert: you mean, I shouldn't define parameters chi and eta at all and directly use
`const std::complex<double> zeta(circle_points[i](0), circle_points[i](1));`

Or should I move the definition 
            `const double chi = circle_points[i](0);`
            `const double eta = circle_points[i](1);`
inside the for loop ?

source/grid/grid_generator.cc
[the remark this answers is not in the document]
peterrum: You are right. The function takes a reference not a pointer. I guess you have to create an internal function `internal_create_triangulation`, which takes a pointer and is called by both `create_triangulation` functions.

source/grid/grid_generator.cc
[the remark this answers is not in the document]
peterrum: It is somewhat inconsistent to start one from 0 and the other from 1.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+            }
+
+          // set material id of block
+          for (auto cell : tria.active_cell_iterators())
+            cell->set_material_id(material_id);
+        };
+
+        // create a subdivided quadrilateral for each block (see last number of
+        // block id)
+        make(trias[0], {A, B, D, C}, {n_subdivision_y, n_subdivision_x_0}, 1);
```
peterrum: Could you make all instances of block ids static constants. You could write at the top of the class:
```cpp
static const unsigned int id_block_1 = 1;
```
and write here:
```suggestion
        make(trias[0], {A, B, D, C}, {n_subdivision_y, n_subdivision_x_0}, id_block_1);
```
This would make it easier to distinguish block ids form actual numbers.

source/grid/grid_generator.cc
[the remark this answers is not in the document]
peterrum: Yes.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+        std::complex<double> z;
+        double               chi, eta;
+
+        for (unsigned int i = 0; i < circle_points.size();
+             i++) // transform each point
+          {
+            chi = circle_points[i](0);
+            eta = circle_points[i](1);
+            std::complex<double> zeta(chi, eta);
+            z = zeta + 1. / zeta;
```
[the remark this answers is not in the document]
eliasstudiert: You mean private variables inside a function? What is the intention of private variables inside a function?

source/grid/grid_generator.cc
eliasstudiert: so for each @param we use:
input <Type> <description>. 
E.g., @param input Double desired length. ?

source/grid/grid_generator.cc
eliasstudiert: how about parameters `deltaX` and `l` in the same function, can't they be constant as well?
`const double deltaX` and `const double l`

source/grid/grid_generator.cc
eliasstudiert: warning: operation on ‘i’ may be undefined [-Wsequence-point]
                 equidist[i++] = Point<2>((i * deltaX - Lj) / (Ljp - Lj) *

Shouldn't be a problem, right? Or can we remove that warning somehow?

source/grid/grid_generator.cc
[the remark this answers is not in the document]
peterrum: Like I have written in code snippet.

source/grid/grid_generator.cc
peterrum: Yes.

source/grid/grid_generator.cc
peterrum: Warnings are bad. Does it help to move the `i++` down?

source/grid/grid_generator.cc
eliasstudiert: at the top of the class means right after `void make_coarse_grid(Triangulation<2> &tria) const
      {` or as class lists of MeshGenerator class?

source/grid/grid_generator.cc
peterrum: Make as much as possible `const`.

peterrum: Statements at one place is true for all other places...

source/grid/grid_generator.cc
masterleinad: It's not clear if `i` is first incremented and then evaluated on the right-hand side or the other way around.

source/grid/grid_generator.cc
```diff
@@ -28,11 +30,1180 @@
...
+    }
+
+
+
+    /**
+     * This class actually creates the airfoil triangulation.
+     */
+    class MeshGenerator
+    {
+    public:
```
peterrum:
```cpp
static const unsigned int id_block_1 = 1;
```

source/grid/grid_generator.cc
eliasstudiert: in this function no more parameters can be const right? Sorry, but I don't understand what you mean by `Statements at one place is true for all other places...`

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,179 @@ namespace GridGenerator
...
+     * @note This function creates a refined mesh (number of global refinements
+     *       can be specified by the user). No manifold is attached. The
+     *       vertices in the final meshed are moved by this function to the
+     *       right position.
+     *
+     * @note This function is currently only implemented for 2D.
+     *
+     * @param tria The triangulation to be created. It needs to be empty upon
+     *             calling this function.
+     * @param additional_data Configuration of the mesh.
```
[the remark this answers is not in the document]
eliasstudiert: So a figure created in incscape showing a mesh schematically with the input parametes of class AdditionalData, right? I will add the figure till Friday evening.
Which format of the picture, png? And how to add a picture in the code? By the link where it is saved?

peterrum:
@eliasstudiert Looks good now! Let's wait for some feedback from the main developers!

eliasstudiert:
visualization

I don't know where is the right place to add this image. It should help to understand the input parameter.

peterrum:
Very nice! This should help the reviewer. @bangerth can we use this picture in the docu?

blaisb:
> visualization
> 
> I don't know where is the right place to add this image. It should help to understand the input parameter.

Amazing picture / amazing work. Love it

source/grid/grid_generator.cc
luca-heltai: This is totally unrelated. My guess is that the default object you create is invalid (i.e., maybe you forgot to initialize some of the members of AdditionalData), and they are initialized by the add_parameters method instead of being initialized at construction time. 

I'd move the initialization where you declare the objects, i.e.,

`
center = Point<2>(0,0);
`
instead of having them initialized in the constructor. This documents the default values in the same place where you declare the members, and ensures that no entries are left un-initialized.

The add_parameter method simply sets the values when it is called, or when a file is parsed.

eliasstudiert: The mentioned problem was caused, as Peter expected, by a mistake in my test program.

Should I still move the initialization of `center` as you suggested?

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1189 @@
...
+
+      prm.enter_subsection("NACA");
+      {
+        prm.add_parameter("NacaId", naca_id, "Naca ID");
+      }
+      prm.leave_subsection();
+
+      prm.enter_subsection("Joukowski");
+      {
+        prm.add_parameter("CenterX",
```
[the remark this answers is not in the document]
peterrum: In the parameter file, you could simply write than:
```
set Center = 0, 0
```

and we can remove `CenterX` and `CenterY`.

source/grid/grid_generator.cc
peterrum: @luca-heltai Initializing default values in the initializer list is how we normally do it, isn't it?

source/grid/grid_generator.cc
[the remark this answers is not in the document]
eliasstudiert: Thanks! This works.

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1186 @@
...
+      prm.enter_subsection("NACA");
+      {
+        prm.add_parameter("NacaId", naca_id, "Naca ID");
+      }
+      prm.leave_subsection();
+
+      prm.enter_subsection("Joukowski");
+      {
+        prm.add_parameter("Center",
+                          center,
```
masterleinad:
```suggestion
                          joukowski_center,
```

peterrum: @masterleinad Thanks! I have changed this!

peterrum:
@eliasstudiert I have squashed all your commits to a single commit. What I have noticed that your name and email address in git is not set up correctly. Could you set it up as described here: https://github.com/dealii/dealii/wiki/Indentation#commit-authorship

Thanks!

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,183 @@ namespace GridGenerator
...
+       */
+      unsigned int airfoil_sampling_factor;
+
+      /**
+       * Constructor.
+       */
+      AdditionalData();
+
+      /**
+       * This Function adds the ParameterHandler entries.
```
kronbichler:
```suggestion
       * This function adds the ParameterHandler entries.
```

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,183 @@ namespace GridGenerator
...
+       */
+      void
+      add_parameters(ParameterHandler &prm);
+    };
+
+
+
+    /**
+     * Initialize the given triangulation with a flow field around an airfoil,
+     * i.e. a Mesh of C-Type approximating Joukowski or NACA (4 digit) airfoils.
```
kronbichler:
```suggestion
     * i.e., a mesh of C-Type approximating Joukowski or NACA (4 digit) airfoils.
```

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,183 @@ namespace GridGenerator
...
+     *
+     * The user can specify the airfoil geometry and the mesh setup by providing
+     * input parametrs for class AdditionalData.
```
kronbichler:
```suggestion
     * input parameters for the struct AdditionalData.
```

include/deal.II/grid/grid_generator.h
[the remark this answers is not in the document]
kronbichler: Regarding the figure, I suggest we post it on the homepage (since it is 364 kB, so a bit large to include it in the git repo). I have uploaded the file to the homepage repo, the link is:
https://www.dealii.org/images/grids/airfoils_naca_joukowski.png

Given the resolution of the file, I suggest we insert the file with some relative width compared to screen. I'm not an expert on what markup is allowed by doxygen, but looking around in other places suggests we could do something like
```
\htmlonly <style>div.image
   * img[src="https://www.dealii.org/images/grids/airfoils_naca_joukowski.png"]{width:50%;}</style> \endhtmlonly
   * @image html https://www.dealii.org/images/grids/airfoils_naca_joukowski.png
```

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,179 @@ namespace GridGenerator
...
+      double length_b2;
+
+      /**
+       * Factor defining the inclination of the face that block 2 and 3
+       * or block 5 and 6 share.
+       */
+      double incline_factor;
+
+      /**
+       * Factor to receive a finer mesh around the airfoil by increasing it.
```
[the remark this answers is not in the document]
kronbichler: I agree that the function `tanh` should be specified.

kronbichler:
There are errors in the MacOSX build:
```
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:177:20: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
                   std::vector<Point<2>>{Point<2>(0), Point<2>(1)},
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:566:17: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
        return {make_points_equidistant(
                ^~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:994:39: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                    const int    ix = std::round((node_(0) - deltax) / dx);
                                 ~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:995:39: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                    const int    iy = std::round(std::abs(node_(1)) / dy);
                                 ~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:1006:42: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                    const int    ix    = std::round(node_(0) / dx);
                                 ~~      ^~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:1007:42: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                    const int    iy    = std::round(node_(1) / dy);
                                 ~~      ^~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:1038:38: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                  const int    iy  = std::round((std::abs(node(1)) - by) / dy);
                               ~~    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:1041:21: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                    std::round(std::abs(node(0) - center_mesh) / dx);
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:1067:37: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                  const int    ix = std::round(std::abs(node(0) - H(0)) / dx);
                               ~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9045/source/grid/grid_generator.cc:1068:37: error: implicit conversion turns floating-point number into integer: 'double' to 'const int' [-Werror,-Wfloat-conversion]
                  const int    iy = std::round(std::abs(node(1)) / dy);
                               ~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Apart from the braces which are probably clear, you need to explicitly cast to an integer, like for the last line:
```
const int    iy = static_cast<int>(std::round(std::abs(node(1)) / dy));
```

kronbichler:
And here the errors from `clang-tidy`:
```
clang-tidy-6.0 -header-filter=/jenkins/workspace/dea/jenkins/workspace/dealii-tidy_PR-9045/source/grid/grid_generator.cc:557:31: error: the const qualified parameter 'serialnumber' is copied for each invocation; consider making it a reference [performance-unnecessary-value-param,-warnings-as-errors]
/jenkins/workspace/dealii-tidy_PR-9045/source/grid/grid_generator.cc:586:45: error: the const qualified parameter 'serialnumber' is copied for each invocation; consider making it a reference [performance-unnecessary-value-param,-warnings-as-errors]
/jenkins/workspace/dealii-tidy_PR-9045/source/grid/grid_generator.cc:613:54: error: the const qualified parameter 'serialnumber' is copied for each invocation; consider making it a reference [performance-unnecessary-value-param,-warnings-as-errors]
/jenkins/workspace/dealii-tidy_PR-9045/source/grid/grid_generator.cc:1072:37: error: C-style casts are discouraged; use static_cast [google-readability-casting,-warnings-as-errors]
```

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,183 @@ namespace GridGenerator
    */
   ///@{
 
+  /**
+   * Namespace Airfoil contains classes and functions in order to create a
+   * C-type mesh for the (flow) field around Joukowski or NACA airfoils.
+   *
+   * @ingroup GridGenerator
+   */
+  namespace Airfoil
```
kronbichler: I realize one thing now: Can you move this code towards the end of the file? I think it makes sense to keep the basic grids like the hypercube at the top and the more specialized ones further down.

include/deal.II/grid/grid_generator.h
```diff
@@ -60,6 +61,183 @@ namespace GridGenerator
...
+      /**
+       * Type of the airfoil: either "NACA" or "Joukowksi" to choose airfoil
+       * geometry among NACA and Joukowski airfoil.
+       */
+      std::string airfoil_type;
+
+      /**
+       * NACA serial number defining the airfoil shape.
+       *
+       * @note Currently serial numbers with length 4 are supported.
```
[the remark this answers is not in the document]
eliasstudiert: Well, I've added a literature in grid_generator.cc where the 4-digit points are computed. 
(I.H. Abbott and A.E. von Doenhoff. Theory of Wing Sections: Including a Summary of Airfoil Data. New York: Dover Publications, 1959.)

I think, a good overview is given in Wikipedia, but is a link to Wikipedia okey? (https://en.wikipedia.org/wiki/NACA_airfoil)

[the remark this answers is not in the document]
peterrum: > but is a link to Wikipedia okey?

Yes it is okay.

eliasstudiert:
@kronbichler and @bangerth , I've fixed the edited your suggested modifications.
@kronbichler , I'm not sure whether I placed the code in .h at the right position and whether the image is included correctly.

include/deal.II/grid/grid_generator.h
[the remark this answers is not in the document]
bangerth: Yes, links are always ok. Just make sure it's linked to from the place where users want to know about it, i.e., the documentation of the function in the .h file (and not somewhere in the innards of the .cc file where a user wouldn't know to look).

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+            // copy point on upper side of airfoil
+            for (unsigned int i = tail_index; i < jouk_points.size(); i++)
+              upper_points.emplace_back(jouk_points[i]);
+            for (unsigned int i = 0; i <= nose_index; i++)
+              upper_points.emplace_back(jouk_points[i]);
+            std::reverse(upper_points.begin(), upper_points.end());
+
+            // copy point on lower side of airfoil
+            for (unsigned int i = nose_index; i <= tail_index; i++)
+              lower_points.emplace_back(jouk_points[i]);
```
bangerth: You can probably do this more efficiently by saying
```
  lower_points.insert (lower_points.end(), jouk_points.begin()+nose_index, jouk_points.begin()+tail_index+1);
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+         *                |y
+         *          .   . |
+         *      .         | .
+         *     .          |   .
+         *    .           |    .
+         *    .       x   |    .
+         * ---.----|------|----|------>x
+         *     .  -1      |   .1
+         *      .         | .
+         *         .    . |
```
bangerth: I love your little ASCII art pieces -- well done!

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+                         const unsigned int number_points)
+        {
+          std::vector<Point<2>> circle_points;
+
+          // Create Circle with number_points - points
+          // unsigned int number_points = 2 * points_per_side - 2;
+
+          // Calculate radius so that point (x=1|y=0) is enclosed - requirement
+          //  for Joukowski transform
+          double radius =
```
bangerth:
```suggestion
          const double radius =
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+            sqrt(center(1) * center(1) + (1 - center(0)) * (1 - center(0)));
+          double radius_test =
```
bangerth:
```suggestion
          const double radius_test =
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+            sqrt(center(1) * center(1) + (1 + center(0)) * (1 + center(0)));
+          // Make sure point (x=-1|y=0) is enclosed by the circle
+          AssertThrow(
```
bangerth: I suspect that you're going to get a warning about an unused variable `radius_test` in release mode. You probably want to add this:
```suggestion
          (void)radius_test;
          AssertThrow(
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+            radius_test < radius,
+            ExcMessage(
+              "Error creating lower circle: Circle for Joukowski-transform does"
+              " not enclose point zeta = -1! Choose different center "
+              "coordinate."));
+          // Create a full circle with radius 'radius' around Point 'center' of
+          // (number_points) equidistant points.
+          double theta = 2 * numbers::PI / number_points;
```
bangerth:
```suggestion
          const double theta = 2 * numbers::PI / number_points;
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+        {
+          std::vector<Point<2>> circle_points;
+
+          // Create Circle with number_points - points
+          // unsigned int number_points = 2 * points_per_side - 2;
+
+          // Calculate radius so that point (x=1|y=0) is enclosed - requirement
+          //  for Joukowski transform
+          double radius =
+            sqrt(center(1) * center(1) + (1 - center(0)) * (1 - center(0)));
```
bangerth:
```suggestion
            std::sqrt(center(1) * center(1) + (1 - center(0)) * (1 - center(0)));
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+          double radius_test =
+            sqrt(center(1) * center(1) + (1 + center(0)) * (1 + center(0)));
```
bangerth:
```suggestion
            std::sqrt(center(1) * center(1) + (1 + center(0)) * (1 + center(0)));
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+         * @return naca_points Vector containing (number_points)-upper or lower
+         *  airfoil points (non-equidistant).
+         */
+        static std::vector<Point<2>>
+        naca_create_points_4_digits(const std::string &serialnumber,
+                                    const unsigned int number_points,
+                                    const bool         is_upper)
+        {
+          // conversion string (char * ) to int
+          const unsigned int digit_0 = int(serialnumber[0]) - 48;
```
bangerth: I think this might be easier to read if you wrote it like this:
```suggestion
          const unsigned int digit_0 = (serialnumber[0] - '0');
```
which I believe is the same.

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+        static std::vector<Point<2>>
+        set_airfoil_length(const std::vector<Point<2>> &input,
+                           const double                 desired_len)
+        {
+          std::vector<Point<2>> output(input.size());
+
+          const auto scale = desired_len / input.front().distance(input.back());
+
+          for (unsigned i = 0; i < input.size(); i++)
+            output[i] = input[i] * scale;
```
bangerth: I think you can write this whole function simpler by saying
```
  std::vector<Point<2>> output = input;
  const double scale = ...;
  for (auto &x : output)
    x *= scale;
  return output;
```
It's the same length in # of lines, but it's a bit easier to read what is happening.

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+              const auto Ljp = arclength_L[j + 1];
+
+              while (Lj <= i * deltaX && i * deltaX <= Ljp &&
+                     i < number_points - 1)
+                {
+                  equidist[i] = Point<2>((i * deltaX - Lj) / (Ljp - Lj) *
+                                           (non_equidistant_points[j + 1] -
+                                            non_equidistant_points[j]) +
+                                         non_equidistant_points[j]);
+                  i++;
```
bangerth: We generally prefer the prefix form:
```suggestion
                  ++i;
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+
+        /**
+         * Create the coarse grid.
+         * Initializes a given triangulation with a coarse grid.
+         * Create 6 coarse grids based on points A-L (class fields) and merges
+         * them to one triangulation.
+         */
+        void make_coarse_grid(Triangulation<2> &tria) const
+        {
+          // create vector of serial triangulations for each block and temporal
```
bangerth:
```suggestion
          // create vector of serial triangulations for each block and temporary
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+
+          rotation_matrix_1[0][0] = +std::cos(-gamma);
+          rotation_matrix_1[0][1] = -std::sin(-gamma);
+          rotation_matrix_1[1][0] = +std::sin(-gamma);
+          rotation_matrix_1[1][1] = +std::cos(-gamma);
+
+          rotation_matrix_2[0][0] = +std::cos(+gamma);
+          rotation_matrix_2[0][1] = -std::sin(+gamma);
+          rotation_matrix_2[1][0] = +std::sin(+gamma);
+          rotation_matrix_2[1][1] = +std::cos(+gamma);
```
bangerth: I *think* that `rotation_matrix_2` is just the transpose of `rotation_matrix_1`, given that `cos` is an even and `sin` an odd function. If so, just write
```
  rotation_matrix_2 = transpose(rotation_matrix_1);
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+      };
+    } // namespace
+
+
+    /**
+     * Initialize the given triangulation an airfoil, i.e. a Mesh of C-Type
+     * approximating Joukowski or NACA (4 digit) airfoils.
+     *
+     * To specify the airfoil geometry and the mesh setup the provided
+     * inputparametrs from class AdditionalData is used. Thereby, different
```
bangerth:
```suggestion
     * input parameters from class AdditionalData is used. Thereby, different
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+     * types of Joukowski or NACA airfoils with variable chord length, far field
+     * size and mesh density can be created.
+     *
+     * By default, a triangulation for a NACA 2412 airfoil with chord
+     * lenght one is generated.
```
bangerth:
```suggestion
     * length one is generated.
```

source/grid/grid_generator.cc
```diff
@@ -28,11 +29,1219 @@
...
+        {
+          return std::tanh(bias_factor * alpha) / std::tanh(bias_factor);
+        }
+      };
+    } // namespace
+
+
+    /**
+     * Initialize the given triangulation an airfoil, i.e. a Mesh of C-Type
+     * approximating Joukowski or NACA (4 digit) airfoils.
```
bangerth: Is this text just a copy of the documentation you have in the .h file? If so, just delete it here -- it's in the .h file where doxygen can pick it up, and having duplication is always a prescription for letting things get out of sync.

doc/news/changes/minor/20191112EliasDejene
```diff
@@ -0,0 +1,4 @@
+New: Add new GridGenerator tool to create a C-type mesh around a NACA or
```
bangerth: By the way, I think this out to be in the "major" directory!

bangerth: out -> ought.

source/grid/grid_generator.cc
eliasstudiert: Cool! But I don't understand why for input last, i.e. `jouk_points.begin()+tail_index+1`, the +1 is required. The last input element should be `jouk_points[tail_index]`.

kronbichler: The `insert` functions, like many other C++ library functions and conventions in deal.II, work with half-open intervals, i.e., ranges where the `begin` index is included but the `end` index is excluded.

eliasstudiert: Thanks!

source/grid/grid_generator.cc
eliasstudiert: It's not a copy but it states exactly the same.  There is only this additional info in that documentation:
 `By default, a triangulation for a NACA 2412 airfoil with chord length one is generated.` 
Is this information necessary?

eliasstudiert:
@bangerth I've edited the suggested changes and moved `20191112EliasDejene`  to major directory.

eliasstudiert:
@bangerth I've edited the suggested changes and moved 20191112EliasDejene to major directory.

bangerth:
This looks good to me now! Can you please squash/fixup the four commits into one (or, if you want, three: one for all of the code, one for the test, one for the changelog)? Then we can merge!

peterrum:
@bangerth Ready to merge :)

bangerth:
Done! Good job everyone over there, what a nice addition!

peterrum:
@eliasstudiert Thanks for your very good work! @blaisb Hope you like it!

eliasstudiert:
Thanks for helping !

PR #9049, opened 2019-11-14
Allow initialization of MappingQCache by lambda

kronbichler:
This function allows to fill the `MappingQCache` by an external lambda function. The typical use case are computations on moving grids where we want to avoid calling into the expensive `MappingFEField` functions (which need to do a lot of work to evaluate the mappings and lead to huge slowdowns when done with high order shape functions) and instead fill the moved points exactly once.

While working on the given test case, I observed that we have a bug in `MappingQGeneric::transform_real_to_unit_cell()` where we go to the shortcut with exact formula in 1D/2D with linear mappings even though we might have excluded `MappingQGeneric::preserves_vertex_locations()`. Since the attached test case provides a check for this (and it is a bit work to create an additional test), I have included this one-line fix in the present PR, even though it is otherwise unrelated.


kronbichler:
@nfehn FYI

source/fe/mapping_q_cache.cc
```diff
@@ -77,7 +77,23 @@ MappingQCache<dim, spacedim>::initialize(
   const MappingQGeneric<dim, spacedim> &mapping)
 {
   AssertDimension(this->get_degree(), mapping.get_degree());
+  initialize(triangulation,
+             [&mapping](const typename Triangulation<dim>::cell_iterator &cell)
+               -> std::vector<Point<spacedim>> {
+               return mapping.compute_mapping_support_points(cell);
```
bangerth: This is a case where things would actually have been shorter with `std::bind` :-) But we're not going to bring that back...

masterleinad: Yes, I would have complained. :wink:

PR #9050, opened 2019-11-14
Fix two places where we used too large numbers in communication channel

kronbichler:
Follow-up to #9034: One place in the block vector was forgotten; furthermore, for the matrix-free code I think we can select yet another channel, placing it `20` within the range of `200` possible channels.

Previously, I had two failing asserts, `matrix_free/stokes_computation.mpirun=4.debug` and `matrix_free/matrix_vector_blocks.mpirun=5.debug`. I'm currently running all MPI tests to see if this is all. `matrix_free` is already through and correct.

### DIFF
--- a/include/deal.II/lac/la_parallel_block_vector.templates.h
+++ b/include/deal.II/lac/la_parallel_block_vector.templates.h
@@ -289,10 +289,10 @@ namespace LinearAlgebra
 
           // In order to avoid conflict with possible other ongoing
           // communication requests (from LA::distributed::Vector that supports
-          // unfinished requests), add an arbitrary number 9923 to the
-          // communication tag
+          // unfinished requests), add 100 to the communication tag (the first
+          // 100 can be used by normal vectors)
           for (unsigned int block = start; block < end; ++block)
-            this->block(block).update_ghost_values_start(block - start + 9923);
+            this->block(block).update_ghost_values_start(block - start + 100);
           for (unsigned int block = start; block < end; ++block)
             this->block(block).update_ghost_values_finish();
         }
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -2923,10 +2923,12 @@ namespace internal
   template <int dim, typename Number, typename VectorizedArrayType>
   struct VectorDataExchange
   {
-    // An arbitrary shift for communication to reduce the risk for accidental
+    // A shift for the MPI messages to reduce the risk for accidental
     // interaction with other open communications that a user program might
-    // set up
-    static constexpr unsigned int channel_shift = 103;
+    // set up (parallel vectors support unfinished communication). We let
+    // the other vectors use the first 20 assigned numbers and start the
+    // matrix-free communication.
+    static constexpr unsigned int channel_shift = 20;
 
 
 


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -2923,10 +2923,12 @@ namespace internal
...
   {
-    // An arbitrary shift for communication to reduce the risk for accidental
+    // A shift for the MPI messages to reduce the risk for accidental
     // interaction with other open communications that a user program might
-    // set up
-    static constexpr unsigned int channel_shift = 103;
+    // set up (parallel vectors support unfinished communication). We let
+    // the other vectors use the first 20 assigned numbers and start the
+    // matrix-free communication.
+    static constexpr unsigned int channel_shift = 20;
```
tjhei: Where is this shift used?

include/deal.II/lac/la_parallel_block_vector.templates.h
```diff
@@ -289,10 +289,10 @@ namespace LinearAlgebra
 
           // In order to avoid conflict with possible other ongoing
           // communication requests (from LA::distributed::Vector that supports
-          // unfinished requests), add an arbitrary number 9923 to the
-          // communication tag
+          // unfinished requests), add 100 to the communication tag (the first
+          // 100 can be used by normal vectors)
           for (unsigned int block = start; block < end; ++block)
-            this->block(block).update_ghost_values_start(block - start + 9923);
+            this->block(block).update_ghost_values_start(block - start + 100);
```
tjhei: ouch, sorry. Yes, I missed that.

tjhei:
Can you please check:
```
        7379 - mpi/fe_tools_extrapolate_06.mpirun=10.debug (Failed)
        7753 - mpi/parallel_block_vector_02.mpirun=10.debug (Failed)
        7755 - mpi/parallel_block_vector_02.mpirun=4.debug (Failed)
```

include/deal.II/matrix_free/matrix_free.h
tjhei: (I worry that somebody selecting 20 as their channel will conflict again)

peterrum:
> (I worry that somebody selecting 20 as their channel will conflict again)

I have run these tests. But they did not fail in my case...

tjhei:
I can confirm that this PR fixes the 3 tests I mention above. :+1:

peterrum:
> (I worry that somebody selecting 20 as their channel will conflict again)

A general remark: I am still not convinced about the way we handle the tags at the moment from a user perspective.  What we are doing is to ban certain tags from usage (since PR #9010 there is a global list with tags >= 300, which is better than before). But is this actually safe? Other libraries (and user codes) might user tags from that the banned tags with the same communicator (in standard case MPI_COMM_WORLD). Wouldn't is be safer to create at distinct places a duplicate of the communicator designated for internal usage only. One such place might be in the constructor of a `parallel::DistributedTriangulationBase` from where it might be passed down to `DoFHandler`, `Vector`, `Partitioner` and `MatrixFree`...

A problem we would face is that at the moment at many places the user explicitly passes the communicator...

include/deal.II/matrix_free/matrix_free.h
kronbichler: It is used for various `LinearAlgebra::distributed::Vector::update_ghost_values_start()` and `_ghost_values_end()` calls, i.e., similarly as we do for the block vector, just using `20` as a shift rather than `100`. The point is: No matter what we choose here, there is always a chance that other communications might go on at the same time. This call is in the range of allowed channels for the parallel vector, and it is unlikely that an exchange not coming from `MatrixFree` would be going on at the same time and use the same MPI tags. The only thing I can imagine, without the user explicitly trying to break it, would conflict is two `MatrixFree` object, which both have the same number of ghost entries to be exchanged, and use the same channel and communicator. That would likely be some multi-threading locally with two of those objects in some embarrasingly parallel approach. Not sure how to approach that - I think at that point we should rather aim to duplicate the MPI communicators as @peterrum suggests.

peterrum:
@kronbichler `/rebuild` seems to be missing...

include/deal.II/matrix_free/matrix_free.h
tjhei: Alright. This is not ideal but likely to work for now. I am okay with merging this PR as is.

Would it make sense to duplicate the communicator for the MatrixFree object?

tjhei:
Let's merge to fix master and think about duplicating communicators later.

Issue #8814, opened 2019-09-20
Function<dim> default constructor in examples and intel compiler


tjhei:
Intel 19.0.5 gives the following warnings:
```
/home/heister/deal-git/examples/step-4/step-4.cc(327): warning #854: const variable "right_hand_side" requires an initializer -- class "RightHandSide<2>" has no user-provided default constructor
    const RightHandSide<dim> right_hand_side;
```
I think this comes from removing it in #8767 
@bangerth thoughts?

bangerth:
The particular change you run into is in #8766. Do you get the same warning in the other tutorials where I removed the default constructor in #8767?

I think we had gotten past the point where the compiler warns about this issue :-(

tjhei:
> Do you get the same warning in the other tutorials where I removed the default constructor in #8767?

Yes.

> I think we had gotten past the point where the compiler warns about this issue :-(

Well, it is only a warning and we could try to silence it.

On the other hand, I don't mind having the constructor there to explicitly call the ``Function<dim>`` constructor: more verbose, but it makes it obvious what to change if you want more than one component.

bangerth:
So we'd have to revert #8767 and that part of #8766. That's a bummer. It's not wrong, it's just one more thing one has to explain.

Before we do this, can someone point out where in the standard it says that such classes should have user-defined constructors? Would an `=default` constructor do?

rezarastak:
Obeservations from [here](https://godbolt.org/z/TI96Ii)

* The warning only appears when we define `RightHandSide right_hand_side` with `const` specifier.
* Changing the constructors or changing the member initialization of `Function` or `FunctionTime` does not help.
* Setting `= default` works.

rezarastak:
Also, this one works too (without a need to add default constructor).
```c++
const RightHandSide<dim> right_hand_side{};
```

masterleinad:
This is DR35 (http://open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html) with C++17 status. It seems that Intel didn't implement the new behavior yet.

tjhei:
Okay, I will just disable the warning as this is considered a "defect".

bangerth:
Is DR35 really the right number? It's 20 years old and has been resolved since TC1. For reference, this link takes you directly to #35: http://open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#35

bangerth:
[...]

PR #8018, opened 2019-05-06
Fix some failing AD tests.

jppelteret:
This targets issues picked up by compiling with -Werror.

PR #7669, opened 2019-01-31
 Be explicit about initialization of static const Tensor variables

masterleinad:
Supersedes #7668. This is needed for older `clang` compilers.

PR #9051, opened 2019-11-14
Fix compiling with clang-3.7.1

masterleinad:
Fixes the errors in https://cdash.43-1.org/viewBuildError.php?buildid=2832.

### DIFF
--- a/examples/step-13/step-13.cc
+++ b/examples/step-13/step-13.cc
@@ -1357,10 +1357,10 @@ namespace Step13
     Triangulation<dim> triangulation;
     GridGenerator::hyper_cube(triangulation, -1, 1);
     triangulation.refine_global(2);
-    const FE_Q<dim>          fe(1);
-    const QGauss<dim>        quadrature(4);
-    const RightHandSide<dim> rhs_function{};
-    const Solution<dim>      boundary_values{};
+    const FE_Q<dim>    fe(1);
+    const QGauss<dim>  quadrature(4);
+    RightHandSide<dim> rhs_function;
+    Solution<dim>      boundary_values;
 
     // Create a solver object of the kind indicated by the argument to this
     // function. If the name is not recognized, throw an exception!
--- a/examples/step-13/step-13.cc
+++ b/examples/step-13/step-13.cc
@@ -1359,8 +1359,8 @@ namespace Step13
     triangulation.refine_global(2);
     const FE_Q<dim>          fe(1);
     const QGauss<dim>        quadrature(4);
-    const RightHandSide<dim> rhs_function;
-    const Solution<dim>      boundary_values;
+    const RightHandSide<dim> rhs_function{};
+    const Solution<dim>      boundary_values{};
 
     // Create a solver object of the kind indicated by the argument to this
     // function. If the name is not recognized, throw an exception!
--- a/examples/step-27/step-27.cc
+++ b/examples/step-27/step-27.cc
@@ -317,7 +317,7 @@ namespace Step27
                                      update_quadrature_points |
                                      update_JxW_values);
 
-    const RightHandSide<dim> rhs_function;
+    const RightHandSide<dim> rhs_function{};
 
     FullMatrix<double> cell_matrix;
     Vector<double>     cell_rhs;
--- a/examples/step-27/step-27.cc
+++ b/examples/step-27/step-27.cc
@@ -317,7 +317,7 @@ namespace Step27
                                      update_quadrature_points |
                                      update_JxW_values);
 
-    const RightHandSide<dim> rhs_function{};
+    RightHandSide<dim> rhs_function;
 
     FullMatrix<double> cell_matrix;
     Vector<double>     cell_rhs;
--- a/examples/step-38/step-38.cc
+++ b/examples/step-38/step-38.cc
@@ -370,7 +370,7 @@ namespace Step38
     std::vector<double>                  rhs_values(n_q_points);
     std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
 
-    const RightHandSide<spacedim> rhs;
+    const RightHandSide<spacedim> rhs{};
 
     for (const auto &cell : dof_handler.active_cell_iterators())
       {
--- a/examples/step-38/step-38.cc
+++ b/examples/step-38/step-38.cc
@@ -370,7 +370,7 @@ namespace Step38
     std::vector<double>                  rhs_values(n_q_points);
     std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
 
-    const RightHandSide<spacedim> rhs{};
+    RightHandSide<spacedim> rhs;
 
     for (const auto &cell : dof_handler.active_cell_iterators())
       {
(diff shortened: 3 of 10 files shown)


bangerth:
I think we had previously concluded that the requirement of the braces is a misunderstanding of the intent of the C++ standard. (We concluded that when we talked about the requirement of having a default constructor for the same classes for which you now have the default initializers.)

clang 3.7 is pretty old. Do you still get the warnings/error for newer versions? Is there a flag (e.g., `-fpermissive`) that one could use for this old compiler?

masterleinad:
You are likely referring to #8767. No, `fpermissive` doesn't help and I am not aware of another workaround. I don't know the minimum version that doesn't result in these errors.

We were dealing similarly with older `clang` compilers in (https://github.com/dealii/dealii/pull/7669 and https://github.com/dealii/dealii/pull/8018). https://github.com/dealii/dealii/issues/8814 is also related, but there we only had warnings we could suppress.

We still claim to support clang 3.3 (https://www.dealii.org/developer/readme.html) so we should either fix these errors or update the requirements. I would be fine with both. @tamiko brought this to our attention so he might have a case where he needs this old version.

tamiko:
@masterleinad Thanks for looking into this :-D

*mhm* I don't need clang-3.7.1 supports - it just happens to be the oldest version floating around on `simserv04`. So what shall we do? I can try to disable compiling example steps for the clang 3.7.1 configuration in case we do not want to touch the example steps.

But I sense the time has come to update our oldest supported compiler versions again.

Opinions?

tjhei:
> Opinions?

I wouldn't mind bumping to clang 3.8 (this is what ubuntu 16.04 ships with).

masterleinad:
If we also bump the `gcc` requirement to `gcc-5`(which ubuntu 16.04 ships with), we could almost require C++14. :smile:

bangerth:
I'll explain why I'm not happy with this patch: We use the tutorials as teaching tools, and as such, everything that is not blatantly obvious requires an explanation that distracts from the actual message we want to convey with each tutorial. Now, C++ is already a rather complex language our students have trouble with, and so the thought of having to explain why one needs `{}` in certain places (and maybe why using `()` is going to yield something entirely different, even though if one were to provide a nonzero number of constructor arguments the two are exactly the same) is just not an appealing thought. The whole point of #8767 was to simplify the code precisely because we had to explain why some classes needed a default constructor -- something that was just not to the message of the programs in which that happened.

We're going to release what we develop now in 2020. That would be four years after clang 3.8 was bundled with Ubuntu, and probably even more years after gcc stopped requiring the constructor. I think that's a reasonably long lag time.

bangerth:
Have we confirmed that clang 3.8 is actually fine with the code?

masterleinad:
No, we need clang 3.9: https://godbolt.org/z/qCCkJe.

tjhei:
> The whole point of #8767 was to simplify the code precisely because we had to explain why some classes needed a default constructor -- something that was just not to the message of the programs in which that happened.

I agree that having to explain ``{}`` is not elegant. In that case, I would rather undo #8767 and be explicit  about the constructor. I actually like this because I had questions on how to set the number of components to something other than 1 before, which is easy if the explicit call to the base is there.

By the way: Ubuntu 18.04 (the next stable) comes with clang-6. I don't mind bumping the minimum to that. It certainly would make testing easier. :-D

bangerth:
I agree that reverting #8767 would be preferable to me over the current patch. We could just re-apply it for 9.3 then.

@dealii/dealii: But what do others think about requiring clang 3.9? (I don't think it's necessary to *require* 6.0.)

drwells:
I think I prefer reverting #8767. Would clang 3.9 accept `= default;` instead?

masterleinad:
> I think I prefer reverting #8767. Would clang 3.9 accept `= default;` instead?

Yes, but `clang 3.8` does not: https://godbolt.org/z/YGJjjq

drwells:
Bummer.

It looks like clang 3.8 is only three years old so I think we should support it - we usually have a much longer support window than that.

Rombur:
> But what do others think about requiring clang 3.9?

I am always in favor of requiring newer compilers :wink:

drwells:
We don't all agree, so lets do the usual thing and vote:

:+1: drop support for clang 3.8 and clang 3.9 
:-1: keep support for clang 3.8 and clang 3.9

Rombur:
@drwells I modified your poll, we wouldn't drop support for 3.8 and 3.9 and 3.7 and 3.8

drwells:
Thanks. We should follow up with a second poll about 3.9 if we drop those.

masterleinad:
Why do you want to discuss 3.9?

drwells:
I thought support of 3.9 was the original question posed by @bangerth.

tamiko:
I currently have the following configurations on the tester:
```
testsuite-64bit_indices (gcc-9)
testsuite-clang-4.0
testsuite-clang-5.0
testsuite-clang-6.0
testsuite-clang-7.0
testsuite-clang-8.0
testsuite-clang-9.0
testsuite-gcc-4.9
testsuite-gcc-6
testsuite-gcc-7
testsuite-gcc-8
testsuite-gcc-9
testsuite-minimal (gcc-9)
testsuite-petsc_complex (gcc-9)
```
I wouldn't mind dropping some of the older compilers. That would in particular free some resources for adding four more flavors (last Ubuntu LTS, last Ubuntu release, last debian release, debian testing).

tamiko:
For context - I plan to migrate these guys here to the tester in a couple of days:
```
 -        bionic            shut off   ubuntu 18.04 lts
 -        bullseye          shut off   debian 11
 -        buster            shut off   debian 10
 -        disco             shut off   ubuntu 19.04
```

bangerth:
Alternative proposal: We only get the error if we mark the objects as `const`. That's conceptually the right thing to do, but I suspect nobody will notice if we "forget" the `const` for these objects. At the very least, I suspect we won't get questions we then need to explain. Would that be a compromise? It avoids having to declare the default constructors, and also avoids the need for `{}` initialization.

masterleinad:
I pushed a commit following that suggestion. I don't have a strong opinion on whether this is a better solution or not. 
We should nevertheless discuss updating the compiler requirements.

bangerth:
I'm ok with merging this as currently written -- I think it introduces minimal hassle. 

The current situation on the poll is:
* Drop support for clang 3.7/3.8: 4 votes
* Keep support: 1 vote.

So we *could* require clang 3.9 (if one wants to use clang to begin with), but we don't actually have to go that route yet. The current patch suffices to extend life support for these older compilers by a while.

masterleinad:
Someone needs to approve this.

bangerth:
I'm approving, but maybe we could have a second person do the same?

(I think it's worthwhile keeping both commits.)

masterleinad:
Then, let's remove the label for now.

drwells:
I think I am missing something - we voted to not support the compiler for which this patch was written. Doesn't that mean we should not merge?

tamiko:
Let's merge for the time being and drop support when something else shows up.

PR #9053, opened 2019-11-15
Bug in grid_tools compute_mesh_predicate_bounding_box for dim!=spacedim triangulations

blaisb:
With @luca-heltai, we have found a bug within the compute_mesh_predicate_bounding_box that leads to an assertion error when the dimension of the triangulation is not the same as the spatial dimension (e.g. <2,3>). This pull request fixes the single line error in the code that is responsible for this. It also adapts the tests with dim!=spacedim

**Importantly** The commented parts in the test  will still crash if you have dim<spacedim (e.g. <2,3>) AND if the maxbbox < number of cell in the refinement level. This is
because the bounding box merge does not work when the cells are part of
a Triangulation<dim,spacedim> with dim!=spacedim. This part should be
further investigated. It seems the algorithm that is responsible for the merge of bounding boxes is at the core responsible for this. A part of this is most likely due to the fact that all comparisons between bounding boxes are made in spacedim, but when dim<spacedim, the distance between the box in the additional dimension will always be zero. However, I have been unable to find an easy fix. Your comments / suggestions are most appreciated on this issue.

### DIFF
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -2021,7 +2021,7 @@ namespace GridTools
         for (; i < active_cells.size(); ++i)
           if (predicate(active_cells[i]))
             for (unsigned int v = 0;
-                 v < GeometryInfo<spacedim>::vertices_per_cell;
+                 v < GeometryInfo<MeshType::dimension>::vertices_per_cell;
                  ++v)
               for (unsigned int d = 0; d < spacedim; ++d)
                 {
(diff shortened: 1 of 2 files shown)


luca-heltai:
>  the distance between the box in the additional dimension will always be zero. However, I have been unable to find an easy fix. Your comments / suggestions are most appreciated on this issue

This will happen only on grids that are truly two dimensional, i.e., where the bounding boxes have zero third dimension. The standard case of a surface mesh in three dimensions should still work. 

The analysis of the code showed several flaws in the design. We're thinking on a solution based on boost::geometry that would be a complete refactoring of the function...

blaisb:
> This will happen only on grids that are truly two dimensional, i.e., where the bounding boxes have zero third dimension. The standard case of a surface mesh in three dimensions should still work.

Yeah you are right. Sorry that was poorly worded.

bangerth:
I think I remember, too,  that the bbox merging code was a bit brittle and could be written in a better way. Either way, this patch here is clearly correct.

PR #9054, opened 2019-11-18
Defaulted copy constructor for Function

masterleinad:
This should fix the warnings in https://cdash.43-1.org/viewBuildError.php?type=1&buildid=2901.

### DIFF
--- a/include/deal.II/base/function.h
+++ b/include/deal.II/base/function.h
@@ -178,6 +178,11 @@ public:
            const time_type    initial_time = 0.0);
 
   /**
+   * Copy constructor.
+   */
+  Function(const Function &f) = default;
+
+  /**
    * Virtual destructor; absolutely necessary in this case.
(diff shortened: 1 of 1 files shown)


drwells:
Deprecating implicit copies is a C++20 change, right?

masterleinad:
No, it's C++11 (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2011/n3242.pdf
> D.3  Implicit declaration of copy functions [depr.impldec]
>
> The implicit declaration of a copy constructor is deprecated if the class has a user-declared copy assignment operator or a user-declared destructor. The implicit declaration of a copy assignment operator is deprecated if the class has a user-declared copy constructor or a user-declared destructor (12.4,12.8).
```

drwells:
That's right - it's just the GCC warning that's new.

PR #9056, opened 2019-11-18
Clarify documentation.

bangerth:


bangerth:
@blaisb -- FYI.

blaisb:
I think it's a lot more clear now from a user point of view. Thank you for the addition :)!

PR #9057, opened 2019-11-19
Fix ADOL-C warnings

masterleinad:
This should fix the warnings in https://cdash.43-1.org/viewBuildError.php?type=1&onlydeltap&buildid=2900.


source/base/symmetric_tensor.cc
```diff
@@ -23,6 +23,145 @@
...
+template <>
+std::array<std::pair<adouble, Tensor<1, 3, adouble>>, 3>
+eigenvectors(const SymmetricTensor<2, 3, adouble> & /*T*/,
+             const SymmetricTensorEigenvectorMethod /*method*/)
+{
+  AssertThrow(false, ExcADOLCAdvancedBranching());
+  return std::array<std::pair<adouble, Tensor<1, 3, adouble>>, 3>();
+}
+#  else
+template std::array<adouble, 1>
```
bangerth: can you add a couple empty lines around the `#else`? It is easy to miss right now.

source/base/symmetric_tensor.cc
```diff
@@ -23,6 +23,145 @@
...
+
+template std::array<adouble, 2>
+eigenvalues(const SymmetricTensor<2, 2, adouble> &);
+
+template std::array<adouble, 3>
+eigenvalues(const SymmetricTensor<2, 3, adouble> &);
+
+template <>
+std::array<std::pair<adouble, Tensor<1, 1, adouble>>, 1>
+eigenvectors(const SymmetricTensor<2, 1, adouble> &,
```
bangerth: This is a forward declaration, not an instantiation. Is this what you want?

masterleinad: No, of course of not. Thanks!

PR #9058, opened 2019-11-19
Fix compiling python-bindings

masterleinad:
This should fix the failing build https://cdash.43-1.org/viewBuildError.php?buildid=2914.

### DIFF
--- a/contrib/python-bindings/include/cell_accessor_wrapper.h
+++ b/contrib/python-bindings/include/cell_accessor_wrapper.h
@@ -172,7 +172,7 @@ namespace python
                    << "acceptable is " << arg2 - 1);
 
   private:
-    template <int dim, int spacedim>
+    template <int dim_, int spacedim_>
     const CellAccessorWrapper
     construct_neighbor_wrapper(const int i) const;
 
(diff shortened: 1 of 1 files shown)


contrib/python-bindings/include/cell_accessor_wrapper.h
```diff
@@ -193,22 +193,22 @@ namespace python
   };
 
 
-  template <int dim, int spacedim>
+  template <int dim_, int spacedim_>
```
Rombur: Can you do the same change at line 175

contrib/python-bindings/include/cell_accessor_wrapper.h
```diff
@@ -193,22 +193,22 @@ namespace python
...
   const CellAccessorWrapper
   CellAccessorWrapper::construct_neighbor_wrapper(const int i) const
   {
-    CellAccessor<dim, spacedim> *cell =
-      static_cast<CellAccessor<dim, spacedim> *>(cell_accessor);
+    Assert(dim_ == dim && spacedim_ == spacedim, ExcInternalError());
```
bangerth: You could make this a `static_assert`.

masterleinad:
@Rombur Here you go!

contrib/python-bindings/include/cell_accessor_wrapper.h
masterleinad: No, `dim` and `spacedim` are member variables. Apart from shadowing, this is the source of the compile errors in the first place.

bangerth: Oh yes, of course!

PR #9059, opened 2019-11-19
Fix hp_cell_weights_03/04

masterleinad:
This should fix https://cdash.43-1.org/testSummary.php?project=1&name=mpi%2Fhp_cell_weights_03.mpirun%3D2.release&date=2019-11-19 and https://cdash.43-1.org/testSummary.php?project=1&name=mpi%2Fhp_cell_weights_04.mpirun%3D2.release&date=2019-11-19 similar to #9013.


marcfehling:
Oh I recall that I wanted to see if the exception was actually thrown. But I didn't think of the `release` build... My bad! :(

PR #9065, opened 2019-11-20
Use std::array in Tensor

masterleinad:


bangerth:
I know you marked this as a WIP, but I'm willing to merge it as is already assuming it passes the tests!

masterleinad:
I removed the `WIP` label. I just had it because I don't know if that helps MSVC or not.

bangerth:
It fails one of the testers (clang 6):
```
clang-tidy-6.0 -header-filter=/jenkins/workspace/dealii-tidy_error: too many errors emitted, stopping now [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/base/tensor.h:1393:3: error: no matching function for call to 'division_operator' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:6989:5: error: no matching function for call to 'compute_nonzero_normal_flux_constraints' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:7181:35: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:7321:24: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:7327:24: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:7520:29: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:7547:33: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:7598:5: error: no matching function for call to 'compute_nonzero_tangential_flux_constraints' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:7619:5: error: no matching function for call to 'compute_nonzero_normal_flux_constraints' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/base/tensor.h:1393:3: error: no matching function for call to 'division_operator' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/base/tensor.h:1393:3: error: no matching function for call to 'division_operator' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:4327:32: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:4504:40: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/numerics/vector_tools.templates.h:5532:22: error: no viable overloaded '/=' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/base/tensor.h:1393:3: error: no matching function for call to 'division_operator' [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/base/tensor.h:1393:3: error: no matching function for call to 'division_operator' [clang-diagnostic-error]
```

drwells:
Will this still work with `dim == 0`?

masterleinad:
> Will this still work with `dim == 0`?

Is there anything specific you are worried about? Otherwise, I guess we just trust in the testsuite.

tamiko:
I love this change. In particular, this should finally allow us to initialize a Tensor solely on nested initializer lists. :-)

bangerth:
Still not quite:
```
clang-tidy-6.0 -header-filter=/jenkins/workspace/dealii-tidy_PR-9065/include/* -p=. -quiet /jenkins/workspace/dealii-tidy/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/physics/elasticity/standard_tensors.h:70:62: error: constexpr variable 'I' must be initialized by a constant expression [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/physics/elasticity/standard_tensors.h:70:62: error: constexpr variable 'I' must be initialized by a constant expression [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/physics/elasticity/standard_tensors.h:103:62: error: constexpr variable 'S' must be initialized by a constant expression [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/physics/elasticity/standard_tensors.h:119:62: error: constexpr variable 'IxI' must be initialized by a constant expression [clang-diagnostic-error]

/jenkins/workspace/dealii-tidy_PR-9065/include/deal.II/physics/elasticity/standard_tensors.h:169:62: error: constexpr variable 'dev_P' must be initialized by a constant expression [clang-diagnostic-error]
```

masterleinad:
Hmm... All most of the modifying `std::array` member funtions are `constexpr` only from C++17 on. I am a little bit stuck here.

bangerth:
Hm, do you have any idea why this code fails at all after your changes? Is it because we are using `operator[]` somewhere where for a plain C-style array, that is allowed in `constexpr` expressions, but not when using ``std::array::operator[]`?

masterleinad:
Yes, precisely. the `non-const` `std::array::operator[]`is `constexpr` only from C++17 on (https://en.cppreference.com/w/cpp/container/array/operator_at).

bangerth:
Darn. So is `data()` and basically all other functions of that class, just like you say. So we use `operator[]` in one of the functions that initialize these variables `I`, `S`, ...?

masterleinad:
Yes, see https://github.com/dealii/dealii/blob/8236ec9b21bd7979076a62365c0770bdc2b5b804/include/deal.II/base/symmetric_tensor.h#L3181-L3204
Of course, we also lose the ability to call `Tensor::operator[]` in a `constexpr` function. This doesn't really seem to be worth dropping the workaround for me.

tamiko:
@bangerth @masterleinad Restricting support to gcc-9 and clang-9 only and requiring C++-20? :stuck_out_tongue:

masterleinad:
> @bangerth @masterleinad Restricting support to gcc-9 and clang-9 only and requiring C++-20? :stuck_out_tongue:

Enforcing C++17 would make things so much easier already. :wink:

bangerth:
I assume that we could specialize these functions so that we don't have to use `operator[]` but just assign an array to the `std::array` object. 

The alternative -- not terrible -- is to just not mark these functions as `constexpr` but go back to `const`. That's what we had before, and it worked just fine. That might be my preferred option for the moment moving forward. Presumably, the number of users caring about these objects being `constexpr` is quite small at the moment. It would also allow us to move forward with this patch for now -- we can iterate ideas about avoiding `operator[]` in a follow-up.

masterleinad:
If `operator[]` is not-`constexpr`, essentially every `non-const` `constexpr` function has to rewritten (and all the `const` ones that use a temporary object of the same type). For me, that doesn't seem to be worth it at all. If someone else wants to pick up the battle, that's fine with me.

bangerth:
That's just a shame :-(

bangerth:
I should add that I think I wasn't quite clear, though: What I really meant is that I wanted the *variables* to just be `const` instead of `constexpr`. The issue is about the various tensors in `StandardTensors` which we currently declare as `constexpr`, but the reality is that I can't really see why they need to be -- they could just be `const` because I can see no good reason why anyone would want to do compile-time arithmetic on them.

And if we went back and made them `const`, then we could still move forward with the patch.

masterleinad:
Sure, we could just fix these variables, but most of the `Tensor` member functions currently declared `constexpr` or `DEAL_II_CONSTEXPR` could still not be used in a `constexpr` context (without major refactoring).

bangerth:
I think my issue with `constexpr` is that I haven't quite seen yet what benefit there is to attaching this attribute to complex functions such as those in classes like `Tensor` and `SymmetricTensor`. It's a bit like the `throw(A,B,C)` statements of yore that were part of C++98 but then were deprecated again in C++11: It sounds like a good idea that, in principle, might enable cool stuff, but it's hard to imagine someone doing complicated *compile-time* stuff with `Tensor` and `SymmetricTensor` with the aim of generating constant expressions (as far as the compiler is concerned). Anything that can be done through regular compiler optimizations is already available in the existing, `const` functions. The issue with `throw()` attributes was that they weren't consistently applied to the standard library, and to really benefit from them, one had to apply them uniformly throughout an entire code base -- much like `constexpr`. What you're finding here is that the C++11 people didn't do that, and it results in a substantial pain.

But like I said, it's probably that I haven't yet completely understood the benefits and use cases.

@rezarastak -- what's your opinion on the issue?

rezarastak:
Short answer: It is highly useful if we keep `Tensor`, `SymmetricTensor`, and `Point` constexpr compatible. It is sad that `std::array` does not provide a `constexpr operator[]`. I am slightly in favor of keeping the `constexpr` functionality even if we have to keep using C-style arrays in `Tensor`.

Long explanation answer:
These are my opinions regarding using `constexpr` in dealii
* The best use case for the `constexpr` feature is to initialize constants. The constants, that we already know their value and they don't depend on any runtime conditions, should ideally be calculated at compile time instead of runtime. The benefits include:
  1. We save some cpu cycles during runtime for initializing the constants.
  2. Global variables and class static variables from different source files can be initialized in unpredictable order causing the notorious **static initialization order ‘fiasco’**. Using `constexpr` global constants creates deterministic and correct initialization behavior.
  3. During the compile-time computations of constants, no undefined behavior can happen. So the compiler will help us find errors such as out-of-bounds access.
  4. The tests used to check the implementations of the constants can replaced by `static_assert`. We haven't used this feature that much so far, but now we can think of having `static_assert` tests at the end of some of the `cc` files.
* The modules and classes that should use `constexpr` in my opinion:
  1. `SymmetricTensor`s defined in `Physics::Elasticity::StandardTensors`. These are heavily used in constitutive equations and knowing their value at compile time can (theoretically) provide better optimization. For example [this code](https://www.godbolt.org/z/EYFpDB) shows that defining a static tensor as `constexpr` can reduce the amount of assembly generated.
  2. All constants in the namespace `numbers`. These area already `constexpr`.
  3. The constants in `GeometryInfo`. Most of these are already `constexpr`.
  4. Within the subclasses of `FiniteElement`, many classes define many constants including coordinates of support points on the unit cell, coefficients of polynomials, etc. It would be great if these constants become `constexpr`, but currently it is not possible because these coefficients and support points are stored in `std::vector`. Additionally the parent class `Subscriptor` cannot currently become `constexpr`.
  5. The `Quadrature` classes have many constants such as quadrature points and weight coefficients. These have the potential to become constexpr but it is currently impossible because these values are stored in`std::vector`.
* Apart from the mentioned constants, we don't need `constexpr` for other parts of dealii, because it increases the noise in the source code. Also, except in certain cased, we should avoid `DEAL_II_CONSTEXPR` and `#ifdef DEAL_II_HAVE_CXX14_CONSTEXPR_CAN_CALL_NONCONSTEXPR` because they are ugly and poorly readable. However, they are needed in certain important parts of `Tensor` and `SymmetricTensor`. If, at some point in the future we decide to require `c++14`, we can remove these macros and improve the readability considerably.
* Users might want to define their own `constexpr` constants, for example they can define constant tensors in their own material models. I think it is important to allow the users of dealii to define `constexpr` variables of types `Tensor`, `SymmetricTensor`, and `Point`. So it is very useful if we provide `constexpr` constructors and member accessors for these classes.

bangerth:
@rezarastak -- thanks for chiming in. I understand the benefits in general, I'm just not convinced that in the case of these `StandardTensors` they are really worth it :-) But let's leave it as is -- it's probably not worth spending more time on the issue.

PR #9067, opened 2019-11-21
hp::Refinement::predict_error: Detangle assertions.

marcfehling:
Doesn't hurt to be more specific on this assertion.

### DIFF
--- a/source/hp/refinement.cc
+++ b/source/hp/refinement.cc
@@ -317,9 +317,14 @@ namespace hp
           {
             if (cell->future_fe_index_set()) // p adaptation
               {
-                Assert(!cell->refine_flag_set() && !cell->coarsen_flag_set(),
-                       ExcMessage("Cell has to be either flagged for h or p "
-                                  "adaptation, and not for both!"));
+                Assert(cell->future_fe_index_set() && !cell->refine_flag_set(),
+                       ExcMessage(
+                         "For error prediction, a cell marked for p-adaptation "
+                         "should not also be flagged for h-refinement!"));
+                Assert(cell->future_fe_index_set() && !cell->coarsen_flag_set(),
+                       ExcMessage(
+                         "For error prediction, a cell marked for p-adaptation "
+                         "should not also be flagged for h-coarsening!"));
 
                 const int degree_difference =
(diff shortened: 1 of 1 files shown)


masterleinad:
I am not sure if this an improvement or not.

marcfehling:
It helped me during a debugging process, and I thought it would be reasonable to integrate that in the library. I see that it may be pointless, but it's just another assertion. Decision is up to you :)

source/hp/refinement.cc
```diff
@@ -317,9 +317,12 @@ namespace hp
           {
             if (cell->future_fe_index_set()) // p adaptation
               {
-                Assert(!cell->refine_flag_set() && !cell->coarsen_flag_set(),
+                Assert(!cell->refine_flag_set(),
                        ExcMessage("Cell has to be either flagged for h or p "
-                                  "adaptation, and not for both!"));
+                                  "refinement, and not for both!"));
```
bangerth: The error text doesn't match the actual condition of the assertion. The conditition is that both calls need to return `false` (so that the negation is `true` and the `&&` implies that both calls need to return `false`). But the text corresponds to the condition `! (cell->refine_flag_set() && cell->coarsen_flag_set())` which is equivalent to `!cell->refine_flag_set() || !cell->coarsen_flag_set()`.

Before moving on with the actual change, can you clarify what it is you want the assertion to be?

marcfehling: I've been more verbose on the error message, and I think it would make sense to also include the other condition that has to be met. We check for this condition in the above `if`-statement anyways, but it won't hurt to include it in the assertion as well since the condition will be displayed to the user, too, once triggered.

```
if (cell->future_fe_index_set()) // p adaptation
  {
    Assert(cell->future_fe_index_set() && !cell->refine_flag_set(),
           ExcMessage(
             "For error prediction, a cell marked for p-adaptation "
             "should not also be flagged for h-refinement.!"));
    Assert(cell->future_fe_index_set() && !cell->coarsen_flag_set(),
           ExcMessage(
             "For error prediction, a cell marked for p-adaptation "
             "should not also be flagged for h-coarsening."));
```

PR #9069, opened 2019-11-21
Insert particles globally.

luca-heltai:
Create and insert a number of particles into the collection of
particles.
This function takes a list of positions and creates a set of particles
at these positions, which are then distributed and added to the local
particle collection of a procesor. Note that this function uses
GridTools::distributed_compute_point_locations. Consequently, it can
require intense communications between the processors.

Opening for discussions. We'll move the implementation to `particle_handler.cc` as soon as we settle discussions.


bangerth:
I would be interested in hearing @gassmoeller 's opinions.

luca-heltai:
\rebuild

luca-heltai:
> One high-level question: You decided to make the indices contiguous in the order the particles were initially handed over to this function (so process 1 creates particles 0,1,2 even if they get transferred to completely different ranks), was that on purpose? 

Yes. The application we have in mind is a fully distributed fluid structure interaction problem. The ids of the particles are "attached" to the dofs of the solid. We could extend this function (which is agnostic on how the numbering is done) to allow, with a switch, to number particles *after* communicating them. I'd do so, if required, in a subsequent PR.

> The only concern I have at the moment is about scalability of this function as it fills some maps entry by entry. Did you do any tests on performance so far?

Partly, we did. Most of the communication is one-to-one, and  in our experiments in step-70, this is still negligible w.r.t. assemblying and solving. Notice also that we keep communicating with the original owners of the particles at each time step, and this, at the moment, takes no more than 2% of the overall cost.

source/particles/particle_handler.cc
```diff
@@ -426,6 +426,203 @@ namespace Particles
...
+    this->insert_particles(particles);
+    for (auto &c : cpu_to_indices)
+      c.second.compress();
+
+    // Take care of properties, if the input vector contains them.
+    const auto global_n_properties =
+      Utilities::MPI::sum(properties.size(), triangulation->get_communicator());
+    if (global_n_properties > 0)
+      {
+        // [TODO]: fix this in some_to_some, to allow communication from
```
gassmoeller: Do I understand this correctly that if some_to_some would support communicating with itself (= simply copying from send to receive buffer) you could save lines 559 - 607 and a copy of all properties? How hard would the fix be?

source/particles/particle_handler.cc
```diff
@@ -426,6 +426,203 @@ namespace Particles
...
+            }
+        // Actually fill the property pool of each particle.
+        for (auto particle : *this)
+          {
+            particle.set_property_pool(get_property_pool());
+            const auto id = particle.get_id();
+            Assert(property_start.find(id) != property_start.end(),
+                   ExcInternalError());
+            const auto start = property_start[id];
+            particle.set_properties(
```
gassmoeller: Since the properties are stored by particle anyway, do you need the unified copy of all properties in lines 557-607? Or could you move the logic of where to read the properties from into this loop? This might reduce the number of index variables one has to keep track of.

tests/particles/particle_handler_10.cc
```diff
@@ -0,0 +1,99 @@
...
+  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+
+  MPILogInitAll all;
+
+  deallog.push("2d/2d");
+  test<2, 2>();
+  deallog.pop();
+  // deallog.push("2d/3d");
+  // test<2, 3>();
+  // deallog.pop();
```
gassmoeller: Please remove this for now, or document why you commented it out.

tests/particles/particle_handler_11.cc
```diff
@@ -0,0 +1,99 @@
...
+  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+
+  MPILogInitAll all;
+
+  deallog.push("2d/2d");
+  test<2, 2>();
+  deallog.pop();
+  // deallog.push("2d/3d");
+  // test<2, 3>();
+  // deallog.pop();
```
gassmoeller: same as above, and in the other tests as well

bangerth:
Thanks, @gassmoeller, for taking a look! 

I can see that keeping the original indices is useful because the order of particles may have meaning the particle handler doesn't know, and I see no good reason to destroy the ordering. If someone wants to re-order by processor, then we can figure out a way to do this before calling the current function, but it seems that doesn't have to happen under the hood.

@luca-heltai -- let me know once you've addressed Rene's comments and I'll take a look myself!

gassmoeller:
> Yes. The application we have in mind is a fully distributed fluid structure interaction problem. The ids of the particles are "attached" to the dofs of the solid. We could extend this function (which is agnostic on how the numbering is done) to allow, with a switch, to number particles after communicating them. I'd do so, if required, in a subsequent PR.

Ah, yes, I forgot to answer this. I am completely fine with the current way if it is more useful to you. Lets not worry about the switch until someone needs it.

include/deal.II/particles/particle_handler.h
```diff
@@ -231,6 +237,57 @@ namespace Particles
...
+     *
+     * @param[in] (Optional) a vector of properties associated with each
+     * local point. The size of the vector should be either zero (no
+     * properties will be transfered nor attached to the generated particles)
+     * or it should be `positions.size()*this->n_properties_per_particle()`.
+     * Notice that this function call will tranfer the properties from the
+     * local mpi process to the final mpi process that will own each of the
+     * particle, and it may therefore be communication intensive.
+     *
+     * @return A pair of maps from owner to IndexSet, that contains the local
```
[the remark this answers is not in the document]
luca-heltai: We originally returned both. However the second can be obtained by calling `second = some_to_some(first)`, and may not be used at all, so we decided to remove it later on (it can be obtained with a one liner, afterall). This part of the comment was a left over... I'm fixing it now, and improving the comment.

source/particles/particle_handler.cc
luca-heltai: Yes. 

The fix is trivial (and that's how the code was originally), but there were a lot of discussions on this exact functionality when we proposed the `some_to_some` function, and its functionality ended up being the way it is after some 250 comments... We can try again, but I was not feeling like proposing it together with this PR... :D

source/particles/particle_handler.cc
luca-heltai: With the some to some doing "the right thing", this would be unified anyway... 

I'd like to try changing the some_to_some in another PR, and then come back to this. It took some time to get this right...

source/particles/particle_handler.cc
```diff
@@ -426,6 +426,203 @@ namespace Particles
...
+          }
+      }
+
+    this->insert_particles(particles);
+    for (auto &c : cpu_to_indices)
+      c.second.compress();
+
+    // Take care of properties, if the input vector contains them.
+    const auto global_n_properties =
+      Utilities::MPI::sum(properties.size(), triangulation->get_communicator());
```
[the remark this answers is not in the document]
luca-heltai: It actually does. The second test checks this. If the sum of all properties is zero, then no properties were passed (on no processor). If the sum is not zero, then every one (even when they have no properties locally, because they have no particles locally) need to enter below and exchange properties. The initial assert guarantees that anyone having particles will have also properties (locally with the correct numbering).

tests/particles/particle_handler_10.cc
luca-heltai: It's documented in the PR that tries to address the bug. I'd rather keep them here, to make sure we remember to uncomment them once that bug is fixed. These are all scattered around the 8 PR we opened. I'll uncomment them out in a future PR. 

Added the comment.

luca-heltai:
@bangerth : feel free to comment!

luca-heltai:
If you prefer to wait for #9121 to be merged, I can simplify a bit the logic afterwards.

gassmoeller:
It looks like #9121 is essentially ready, so yes, waiting for that and simplifying the logic would be great.

bangerth:
Please rebase after the changes to `some_to_some()` and ping me again!

source/particles/particle_handler.cc
luca-heltai: Done.

luca-heltai:
@bangerth , I think I have addressed all your comments. @blaisb , the interface has changed: now the function expects a `std::vector<std::vector<double>>` instead of `std::vector<double>`.

This has simplified the implementation, but will propagate in all the other PR and in step-70...

luca-heltai:
Ping @bangerth ?

source/particles/particle_handler.cc
```diff
@@ -442,6 +442,204 @@ namespace Particles
...
+
+    // Unpack the information into several vectors:
+    // All cells that contain at least one particle
+    const auto &local_cells_containing_particles =
+      std::get<0>(cells_positions_and_index_maps);
+
+    // The reference position of every particle in the local part of the
+    // triangulation.
+    const auto &local_reference_positions =
+      std::get<1>(cells_positions_and_index_maps);
```
bangerth: I can't wait till we can finally have structured bindings to make the initialization of these local references simpler and easier to read...

PR #9070, opened 2019-11-21
Extract dofs per component

luca-heltai:
Extract the set of local, active or relevant DoF indices for each components within the mask that are owned by the current  processor. For components disabled by the mask, an empty IndexSet is returned.

For sequential and scalar DoFHandler objects, this function returns a single set, which  is the complete set with all DoF indices.


include/deal.II/dofs/dof_tools.h
```diff
@@ -1697,6 +1697,82 @@ namespace DoFTools
...
+    std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
+    for (const auto &cell : dof_handler.active_cell_iterators())
+      {
+        if (cell->is_locally_owned() ||
+            (cell->is_ghost() && ownership == OwnershipType::relevant))
+          {
+            cell->get_dof_indices(local_dof_indices);
+            for (unsigned int i = 0; i < dofs_per_cell; ++i)
+              {
+                unsigned int dof_comp = fe.system_to_component_index(i).first;
```
bangerth:
```suggestion
                const unsigned int dof_comp = fe.system_to_component_index(i).first;
```

include/deal.II/dofs/dof_tools.h
```diff
@@ -1697,6 +1697,82 @@ namespace DoFTools
   extract_locally_relevant_dofs(const DoFHandlerType &dof_handler,
                                 IndexSet &            dof_set);
 
+
+  /**
+   * Extract the set of local, active or relevant DoF indices for each
+   * components within the mask that are owned by the current  processor. For
+   * components disabled by the mask, an empty IndexSet is returned.
+   * For regular DoFHandler objects, this set is the complete set with all DoF
```
bangerth: what is a "regular" DoFHandler? I think you want to say "a DoFHandler built on a sequential triangulation".

include/deal.II/dofs/dof_tools.h
```diff
@@ -1697,6 +1697,82 @@ namespace DoFTools
...
+   * indices. In either case, it equals what DoFHandler::locally_owned_dofs()
+   * returns. The ownership type is fixed by a small enum that controls which
+   * type of ownership is asked off.
+   */
+
+  enum class OwnershipType
+  {
+    owned,
+    active,
```
bangerth: We've generally tried to avoid introducing these one-time enums. I'd favor us using three different functions (like the ones we already have), or if you must an argument that is simply an integer or a string -- but three different functions would still be my preference.

luca-heltai: This would lead to three identical implementation, with the exception of a check on what cells to ask dofs to. They are one-time enums, if no-one else uses them... but there are many other functions that may actually be converted to a single one taking such an enum: it is not a function specific enum.

If you still prefer three different functions, we'll replicate them.

luca-heltai:
@blaisb , I have changed the name, to reflect @bangerth comments. Since we currently only need the locally owned version, I've only provided the implementation for the `locally_owned_dofs_per_component`.

By skimming through the other implementation and documentation (the one filling a vector of bools), I changed a bit the implementation to make it as general as the other one.

luca-heltai:
@tjhei , I'm not sure why the mpi build is failing. I only see a linking failed message, but it seems unrelated to our changes...

tjhei:
> @tjhei , I'm not sure why the mpi build is failing. I only see a linking failed message, but it seems unrelated to our changes...

Not sure. I just restarted the build and will check in a bit.

luca-heltai:
ping @bangerth

luca-heltai:
@bangerth , in reference to your comments on #9073, we also need this one to be merged first (and this should be ready now)

include/deal.II/dofs/dof_tools.h
```diff
@@ -1697,8 +1697,23 @@ namespace DoFTools
   extract_locally_relevant_dofs(const DoFHandlerType &dof_handler,
                                 IndexSet &            dof_set);
 
+
   /**
+   * Extract the set of locally owned DoF indices for each components within the
+   * mask that are owned by the current  processor. For components disabled by
+   * the mask, an empty IndexSet is returned. For a scalar DoFHandler built on a
+   * sequential triangulation, the return vector contains a single complete
+   * IndexSet with all DoF indices.
```
bangerth:
```suggestion
   * IndexSet with all DoF indices. If the mask contains all components (which also
   * corresponds to the default value), then the union of the index sets returned
   * equals what DoFHandler::locally_owned_dofs() returns.
```

PR #9071, opened 2019-11-21
Map dofs to support points

luca-heltai:
Same function as the old ones, but returns an IndexSet
per component.


include/deal.II/dofs/dof_tools.h
```diff
@@ -2356,6 +2357,63 @@ namespace DoFTools
     const DoFHandler<dim, spacedim> &                   dof_handler,
     std::map<types::global_dof_index, Point<spacedim>> &support_points);
 
+  // Same as above but with support for a component mask
+
+  template <int dim, int spacedim>
```
bangerth: make this a regular doxygen-style comment

include/deal.II/dofs/dof_tools.h
```diff
@@ -2356,6 +2357,63 @@ namespace DoFTools
...
+  void
+  map_dofs_to_support_points(
+    const Mapping<dim, spacedim> &                      mapping,
+    const DoFHandler<dim, spacedim> &                   dof_handler,
+    std::map<types::global_dof_index, Point<spacedim>> &support_points,
+    const ComponentMask &                               mask)
```
bangerth: why not just give the existing function the additional argument and provide a default value?

blaisb:
@bangerth You are absolutely right. When Luca and I wrote that we had envisioned something else, and since we were running out of time before I left Italy we did not rewrite. Clearly, the best way to proceed is to add the mask to the existing function and have a default empty mask that is set to all true.
If you agree, let me fix that (most likely Tuesday) and push another commit. I think keeping the test is a good idea however.
@luca-heltai I'll take care of this one :).

blaisb:
Sorry I had not updated the main branch. @bangerth @luca-heltai  I refactored the new map_dof_to_support_points to include a mask that is by default set to all true. This way I could remove the additional code. Looking forward to your comments.

include/deal.II/dofs/dof_tools.h
```diff
@@ -26,6 +26,7 @@
 #include <deal.II/dofs/dof_handler.h>
 
 #include <deal.II/fe/component_mask.h>
+#include <deal.II/fe/fe_values.h>
```
bangerth: I think  you only want to `#include <deal.II/fe/component_mask.h>`.

include/deal.II/dofs/dof_tools.h
```diff
@@ -2309,7 +2310,8 @@ namespace DoFTools
   void
   map_dofs_to_support_points(const Mapping<dim, spacedim> &   mapping,
                              const DoFHandler<dim, spacedim> &dof_handler,
-                             std::vector<Point<spacedim>> &   support_points);
+                             std::vector<Point<spacedim>> &   support_points,
+                             const ComponentMask &mask = ComponentMask());
```
bangerth: Can you document the additional argument?

include/deal.II/dofs/dof_tools.h
```diff
@@ -2348,13 +2351,16 @@ namespace DoFTools
    * @param support_points A map that for every locally relevant DoF index
    * contains the corresponding location in real space coordinates. Previous
    * content of this object is deleted in this function.
+   * @param component_mask An optional component mask that restricts the
+   * components from which the support points are extracted
```
bangerth: Copy this sentence to the other places. Add a period to the end of the sentence.

source/dofs/dof_tools.cc
```diff
@@ -2126,8 +2133,14 @@ namespace DoFTools
               const std::vector<Point<spacedim>> &points =
                 fe_values.get_quadrature_points();
               for (unsigned int i = 0; i < cell->get_fe().dofs_per_cell; ++i)
-                // insert the values into the map
-                support_points[local_dof_indices[i]] = points[i];
+                {
+                  unsigned int dof_comp =
```
bangerth:
```suggestion
                  const unsigned int dof_comp =
```

blaisb:
@bangerth  Sorry for the delay. I applied all your comments. I was not so sure about the documentation of the first instance of that function since it previously did not contain any explanation of the parameters. So I went ahead and added it for all parameters (maybe it is easier to understand this way?). I am unsure if that was the right approach.

blaisb:
@bangerth  I think this will be ready to merge when you have the time to take a look :)!

bangerth:
Five of the six testers are ok, but 'tidy' is apparently stuck. I'll merge on the assumption that it won't find anything too grave...

PR #9077, opened 2019-11-21
Coupling between non matching parallel distributed objects

luca-heltai:
Handle all things needed to couple non matching parallel distributed
triangulations through particle handler objects.

Depends on #9069  and some other subsequent pull requests.


bangerth:
@luca-heltai: What's the status of this patch? It lacks documentation, but I think that, from a very cursory look at the patch, it's also not quite clear to me whether the main class wants to be a container storing data or wants to be a namespace for tool functions. The first few functions that generate sparsity patterns and similar could certainly be tool functions in some namespace -- instead of getting their data from member variables, they would then just get their inputs via regular arguments. For certain we have many similar functions that are all tool functions in some namespace.

Let me know how you want to proceed with this patch!

bangerth:
@luca-heltai -- ping?

luca-heltai:
I've discussed about this PR with @blaisb yesterday. I think we will scorporate some of the functions in this class, and put them in `non_matching/coupling`. The rest of this class, will be incorporated directly in step-70, as it is rather application specific.

PR #9080, opened 2019-11-22
Fix compiling with MSVC

### LINKED ISSUE #9060: Compilation on MSVS 2019 dealii.master
Hi, I'm try deal.ii for the first time. I have MSVS 2019 and I have managed to run cmake as instructed. However I get compilation errors. I know its experiemental but would appreciate some help.

C:\dealii\include\deal.II/base/symmetric_tensor.h(2428,1): error C2668: 'dealii::internal::SymmetricTensorImplementation::unrolled_to_component_indices': ambiguous call to overloaded function

C:\dealii\include\deal.II/base/tensor_product_polynomials.h(482,16): error C2466: cannot allocate an array of constant size 0

Error	C2133	'indices': unknown size	obj_base_release	C:\dealii\include\deal.II\base\tensor_product_polynomials.h	482

Error	C3615	constexpr function 'dealii::Tensor<1,6,Number>::Tensor' cannot result in a constant expression
        with
        [
            Number=double
        ]	obj_physics_elasticity_release	C:\dealii\include\deal.II\physics\elasticity\standard_tensors.h	72	

Error	C3615	constexpr function 'dealii::Tensor<1,6,Number>::Tensor' cannot result in a constant expression
        with
        [
            Number=double
        ]	obj_physics_elasticity_release	C:\dealii\include\deal.II\physics\elasticity\standard_tensors.h	121

masterleinad:
Fixes #9060.

### DIFF
--- a/cmake/checks/check_01_cxx_features.cmake
+++ b/cmake/checks/check_01_cxx_features.cmake
@@ -731,28 +731,30 @@ CHECK_CXX_SOURCE_COMPILES(
 # expression, C++14 allows to call non-constexpr functions from constexpr
 # functions. Unfortunately, not all compilers obey the standard in this regard.
 #
-CHECK_CXX_SOURCE_COMPILES(
-  "
-  #define Assert(x,y) if (!(x)) throw y;
-  void bar()
-  {}
+IF(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+  CHECK_CXX_SOURCE_COMPILES(
+    "
+    #define Assert(x,y) if (!(x)) throw y;
+    void bar()
+    {}
 
-  constexpr int
-  foo(const int n)
-  {
-    Assert(n>0, \"hello\");
-    if(!(n >= 0))
-      bar();
-    return n;
-  }
+    constexpr int
+    foo(const int n)
+    {
+      Assert(n>0, \"hello\");
+      if(!(n >= 0))
+        bar();
+      return n;
+    }
 
-  int main()
-  {
-    constexpr unsigned int n=foo(1);
-    return n;
-  }
-  "
-  DEAL_II_HAVE_CXX14_CONSTEXPR_CAN_CALL_NONCONSTEXPR)
+    int main()
+    {
+      constexpr unsigned int n=foo(1);
+      return n;
+    }
+    "
+    DEAL_II_HAVE_CXX14_CONSTEXPR_CAN_CALL_NONCONSTEXPR)
+ENDIF()
 
 #
 # The macro DEAL_II_CONSTEXPR allows using c++ constexpr features in a portable way.
--- a/contrib/ci/appveyor.yml
+++ b/contrib/ci/appveyor.yml
@@ -1,0 +1,9 @@
+image:
+  - Visual Studio 2019
+clone_folder: c:\projects\source
+build_script:
+- cmd: >-
+    mkdir build &&
+    cd build &&
+    cmake c:\projects\source -DCMAKE_BUILD_TYPE=Debug -G "Visual Studio 16 2019" &&
+    cmake --build . -- -m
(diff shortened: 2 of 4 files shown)


masterleinad:
See https://ci.appveyor.com/project/masterleinad/dealii/builds/29057545.

cmake/checks/check_01_cxx_features.cmake
```diff
@@ -731,28 +731,30 @@ CHECK_CXX_SOURCE_COMPILES(
 # expression, C++14 allows to call non-constexpr functions from constexpr
 # functions. Unfortunately, not all compilers obey the standard in this regard.
 #
-CHECK_CXX_SOURCE_COMPILES(
-  "
-  #define Assert(x,y) if (!(x)) throw y;
-  void bar()
-  {}
+IF(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
```
tjhei: what is the problem with MSVC here, does it pass this test but breaks at something more complicated?

tjhei:
Would you mind moving the appveyor.yml into contrib/ci/ ? I think we can easily configure appveyor to use that file instead.

tjhei:
also please fix indentation. :-)

cmake/checks/check_01_cxx_features.cmake
masterleinad: Yes, it passes the test but gives internal compiler errors if we actually use `constexpr` in the respective places.

masterleinad:
Done.
/rebuild

cmake/checks/check_01_cxx_features.cmake
bangerth: Ah, that's unfortunate because the way the macro works now is that MS VC will *never* work. It would be nice to actually encode the internal compiler error into a feature check itself, so that if MS comes around and fixes the issue, the cmake check will ultimately succeed. 

I think it's ok to do it this way for the moment just to make progress on the overarching issue, but it would be good if we could come back to it in the next few days and refine the macro. Do you have access to a compiler that shows these problems?

Separately: Can you explain the issue in the comment above?

masterleinad:
> Would you mind moving the appveyor.yml into contrib/ci/ ? I think we can easily configure appveyor to use that file instead.

Yes, you can set a custom path in `Settings->General`. We could also have the script just in the `appveyor` configuration, but I prefer to have it visible externally in a file.

cmake/checks/check_01_cxx_features.cmake
masterleinad: I discovered that `ddet_F_dC` (being a `constexpr` function) triggers the compiler error even if it just returns a `SymmetricTensor` object with all member functions and member variables deleted. At this point, I deactivated the macro already. Unfortunately, I couldn't reproduce that in a stand-alone test case, yet (and don't really intend to at the moment). Removing `constexpr` fixed the issue.

No, I don't have access to `MSVC` directly, but just through `appveyor`. I am happy if someone else looks into the issue separately and tries to find a reproducer we could use.

masterleinad: I am not sure if a comment says much more than the code here but I can put something (after the testsuite finished).

masterleinad:
See https://ci.appveyor.com/project/masterleinad/dealii.

cmake/checks/check_01_cxx_features.cmake
tjhei: would be good to have something like "MSVC 2019 crashes with an internal compiler error when in a constexpr function see #9080", or something.

masterleinad: @bangerth If you don't mind, I would actually prefer to add a comment in a follow-up PR to not run the tester again (if there aren't any other comments to address here).

tjhei: Okay, I will hit merge then.

bangerth: Yes, that's fine with me.

PR #9088, opened 2019-11-24
Suppress warning in python-bindings

masterleinad:
Suppresses the warnings in https://cdash.43-1.org/viewBuildError.php?type=1&buildid=2990. Also, see https://github.com/pybind/pybind11/issues/1893.


tjhei:
Can you please add a comment why this is necessary?

PR #9096, opened 2019-11-25
cleanup typedefs in DataOut

tamiko:
The goal is to support (interpolated) coarser output of a solution with
distributed data structures. The subtle problem here lies in the fact that
we can not simply iterate on a (local) coarser level and output the
solution - we have to communicate the ghost layer to do the interpolation
first; secondly, we also have to create a partition on the desired coarser
level. But that pretty much sounds like multigrid!

So my idea is as follows: Suppose you have a distributed solution
`solution`. We can then create a (matrix free) multrigrid transfer as
follows:
```
unsigned int coarsening_steps = 2;
vector_type solution_vector;

MGConstrainedDoFs constrained_dofs;
MGTransferMatrixFree<dim, Number> transfer;

MGLevelObject<vector_type> output_vector;

const auto max_level = triangulation.n_global_levels() - 1;
const auto output_level = max_level - std::min(coarsening_steps, max_level);

constrained_dofs.initialize(dof_handler);
transfer.initialize_constraints(constrained_dofs);
transfer.build(dof_handler);
transfer.interpolate_to_mg(dof_handler, output_vector, solution_vector);
```
So, the only step left over is to actually output said level. For example
via
```
MyDataOut<dim> data_out;
data_out.attach_dof_handler(dof_handler);
data_out.add_data_vector(output_vector);
```
where `MyDataOut` is derived from `dealii::DataOut` and overrides the
`first_cell()` and `next_cell()` functions to iterate over a fixed level.

With the current functionality I think we can integrate this quite nicely.
The following might work:
 - [x] remove unnecessary `active_cell_iterator`s from `DataOut` and `DataOut_DoFData`
 - [ ] improve the `first_locally_owned_cell` and `next_locally_owned_cell`
       member functions to use
       `cell->level_subdomain_id() == triangulation.locally_owned_subdomain`
 - [ ] improve `ParallelDataBase<dim, spacedim>::reinit_all_fe_values` to
       initialize with a `DoFHandler::level_cell_iterator` if the current
       cell is not active.
 - [ ] change `get_function_values` (and similar) to work on level dofs
       instead of dofs if the cell is not active.

@kronbichler *ping*

PR #9098, opened 2019-11-26
Implement data_out for mg levels (WIP)

peterrum:
Same as #9096 but implemented independently in parallel. 

My test program:
```cpp
#include <deal.II/distributed/tria.h>

#include <deal.II/numerics/data_out.h>

const int dim                = 2;
const int global_refinements = 2;
const int level              = 2;
const int degree             = 1;

using namespace dealii;

template <int dim, typename DoFHandlerType = DoFHandler<dim>>
class MGDataOut : public DataOut<dim, DoFHandlerType>
{
public:
  MGDataOut(unsigned int level)
    : level(level)
  {}

private:
  typename DataOut<dim, DoFHandlerType>::cell_iterator
  first_cell()
  {
    return this->triangulation->begin(level);
  }

  typename DataOut<dim, DoFHandlerType>::cell_iterator
  next_cell(const typename DataOut<dim, DoFHandlerType>::cell_iterator &cell)
  {
    typename Triangulation<DoFHandlerType::dimension,
                           DoFHandlerType::space_dimension>::cell_iterator
      active_cell = cell;

    ++active_cell;

    if (active_cell == this->triangulation->end(level))
      return this->dofs->end();

    return active_cell;
  }

  unsigned int level;
};



int
main(int argc, char **argv)
{

  parallel::distributed::Triangulation<dim> tria(
    MPI_COMM_WORLD,
    dealii::Triangulation<dim>::none,
    parallel::distributed::Triangulation<dim>::construct_multigrid_hierarchy);
  tria.refine_global(global_refinements);

  DoFHandler<dim> dof_handler(tria);
  FE_DGQ<dim>     fe(degree);
  dof_handler.distribute_dofs(fe);
  dof_handler.distribute_mg_dofs();


  for (unsigned int level = 0; level < global_refinements; level++)
    {
      MGDataOut<dim> data_out(level);
      data_out.attach_dof_handler(dof_handler);

      Vector<double> solution;
      solution.reinit(dof_handler.n_dofs(level));

      unsigned int i = 0;
      for (auto &v : solution)
        v = i++;
      // solution[3] = 1.0;

      solution.print(std::cout);

      data_out.add_data_vector(
        solution,
        "solution",
        MGDataOut<dim>::type_dof_data,
        std::vector<DataComponentInterpretation::DataComponentInterpretation>(),
        level);
      data_out.build_patches(1 /*degree*/);

      const std::string filename =
        "output/solution." + std::to_string(level) + ".vtu";
      data_out.write_vtu_in_parallel(filename.c_str(), MPI_COMM_WORLD);
    }
}
```

@tamiko Let's collaborate on this issue! @bangerth I had make changes (and remove some asserts) to make it work as it is. I would be thankful for some help regarding `FEValues`, ....

ping @lpsaavedra

### DIFF
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -3907,25 +3907,39 @@ DoFCellAccessor<DoFHandlerType, level_dof_access>::get_dof_values(
   ForwardIterator    local_values_end) const
 {
   (void)local_values_end;
-  Assert(this->is_artificial() == false,
-         ExcMessage("Can't ask for DoF indices on artificial cells."));
-  Assert(!this->has_children(), ExcMessage("Cell must be active."));
-  Assert(this->dof_handler != nullptr, typename BaseClass::ExcInvalidObject());
+  // Assert(this->is_artificial() == false,
+  //       ExcMessage("Can't ask for DoF indices on artificial cells."));
+  // Assert(!this->has_children(), ExcMessage("Cell must be active."));
+  // Assert(this->dof_handler != nullptr, typename
+  // BaseClass::ExcInvalidObject());
 
-  Assert(static_cast<unsigned int>(local_values_end - local_values_begin) ==
-           this->get_fe().dofs_per_cell,
-         typename DoFCellAccessor::ExcVectorDoesNotMatch());
-  Assert(values.size() == this->get_dof_handler().n_dofs(),
-         typename DoFCellAccessor::ExcVectorDoesNotMatch());
+  // Assert(static_cast<unsigned int>(local_values_end - local_values_begin) ==
+  //         this->get_fe().dofs_per_cell,
+  //       typename DoFCellAccessor::ExcVectorDoesNotMatch());
+  // Assert(values.size() == this->get_dof_handler().n_dofs(),
+  //       typename DoFCellAccessor::ExcVectorDoesNotMatch());
 
-  const types::global_dof_index *cache =
+  std::vector<types::global_dof_index> temp;
+  types::global_dof_index *            cache;
+  if (this->active())
     this->dof_handler->levels[this->present_level]->get_cell_cache_start(
       this->present_index, this->get_fe().dofs_per_cell);
+  else
+    {
+      temp.resize(this->get_fe().dofs_per_cell);
+      this->get_mg_dof_indices(temp);
+      cache = &temp[0];
+    }
+
   dealii::internal::DoFAccessorImplementation::Implementation::
     extract_subvector_to(values,
                          cache,
                          cache + this->get_fe().dofs_per_cell,
                          local_values_begin);
+
+  for (unsigned int i = 0; i < this->get_fe().dofs_per_cell; i++)
+    std::cout << local_values_begin[i] << std::endl;
+  std::cout << std::endl;
 }
 
 
--- a/include/deal.II/numerics/data_out_dof_data.templates.h
+++ b/include/deal.II/numerics/data_out_dof_data.templates.h
@@ -231,7 +231,7 @@ namespace internal
               duplicate = true;
           if (duplicate == false)
             {
-              typename DoFHandlerType::active_cell_iterator dh_cell(
+              typename DoFHandlerType::cell_iterator dh_cell(
                 &cell->get_triangulation(),
                 cell->level(),
                 cell->index(),
@@ -1042,8 +1042,10 @@ void
 DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::add_data_vector(
   const DoFHandlerType &                                    dof_handler,
   const VectorType &                                        vec,
-  const DataPostprocessor<DoFHandlerType::space_dimension> &data_postprocessor)
+  const DataPostprocessor<DoFHandlerType::space_dimension> &data_postprocessor,
+  const unsigned int                                        mg_level)
 {
+  (void)mg_level;
   // this is a specialized version of the other function where we have a
   // postprocessor. if we do, we know that we have type_dof_data, which makes
   // things a bit simpler, we also don't need to deal with some of the other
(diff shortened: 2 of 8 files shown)


include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -3907,25 +3907,39 @@ DoFCellAccessor<DoFHandlerType, level_dof_access>::get_dof_values(
...
+  //       ExcMessage("Can't ask for DoF indices on artificial cells."));
+  // Assert(!this->has_children(), ExcMessage("Cell must be active."));
+  // Assert(this->dof_handler != nullptr, typename
+  // BaseClass::ExcInvalidObject());
+
+  // Assert(static_cast<unsigned int>(local_values_end - local_values_begin) ==
+  //         this->get_fe().dofs_per_cell,
+  //       typename DoFCellAccessor::ExcVectorDoesNotMatch());
+  // Assert(values.size() == this->get_dof_handler().n_dofs(),
+  //       typename DoFCellAccessor::ExcVectorDoesNotMatch());
```
kronbichler: This is of course something we need to work on to improve - assuming that we still need to go over the `DoFCellAccessor` class in the end. One thing I'm thinking of is to enable the assertions in case `level_dof_access` is `false`, and in case it is true we could check
```c++
  Assert(this->level_subdomain_id() != numbers::artificial_subdomain_id,
         ExcMessage("Can't ask for DoF indices on artificial cells."));
  Assert(this->dof_handler != nullptr, typename BaseClass::ExcInvalidObject());
  Assert(static_cast<unsigned int>(local_values_end - local_values_begin) ==
           this->get_fe().dofs_per_cell,
         typename DoFCellAccessor::ExcVectorDoesNotMatch());
  Assert(values.size() == this->get_dof_handler().n_dofs(this->level()),
         typename DoFCellAccessor::ExcVectorDoesNotMatch());
```
(As you can see, I do think that we should be able to keep the third and fourth assertion.) Can you try if something along those lines works? For that to work, we would of course need to have level iterators for the level output, and active iterators for the standard output.

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -3907,25 +3907,39 @@ DoFCellAccessor<DoFHandlerType, level_dof_access>::get_dof_values(
...
+  if (this->active())
     this->dof_handler->levels[this->present_level]->get_cell_cache_start(
       this->present_index, this->get_fe().dofs_per_cell);
+  else
+    {
+      temp.resize(this->get_fe().dofs_per_cell);
+      this->get_mg_dof_indices(temp);
+      cache = &temp[0];
+    }
+
```
kronbichler: As we need this switch, I would suggest to rather duplicate the call to `extract_subvector_to` rather than using this additional pointer manipulation.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1170,7 +1173,7 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::
       case type_dof_data:
         Assert(dof_handler != nullptr,
                Exceptions::DataOutImplementation::ExcNoDoFHandlerSelected());
-        Assert(data_vector.size() == dof_handler->n_dofs(),
+        Assert(data_vector.size() == dof_handler->n_dofs(mg_level),
```
kronbichler: Without having looked into all mechanisms in detail yet, I do not believe this works in case `mg_level` is in the default `-1` state, but you have to switch at the level of the call info `n_dofs()` or `n_dofs(mg_level)`.

source/dofs/dof_accessor_get.cc
```diff
@@ -50,7 +50,7 @@ DoFCellAccessor<DoFHandlerType, lda>::get_interpolated_dof_values(
   Vector<number> &   interpolated_values,
   const unsigned int fe_index) const
 {
-  if (!this->has_children())
+  if (true || !this->has_children())
```
kronbichler: This is something we need to get rid of in the end.

peterrum:
@kronbichler This code is far from mergable: I was able to the data on the mg levels. The commit shows only what is necessary to change to make it work. Having now a working example, I will refactor the code.

kronbichler:
> This code is far from mergable

I know, I just wanted to give some hints where to refactor and how to possibly to some steps.

tamiko:
@peterrum Funny how similar ideas get attention at exactly the same moment in time :smile: Would you mind basing your current state on top of my commit on the other branch? I have just pushed some very minimal cleanup. I'd then just close my PR and we get this implemented together.

tamiko:
@peterrum One quick remark: In your test program you currently encode the `level` (that should be exported) twice. Once in the `MGDataOut` class that determines the cell level and once in your call
```
data_out.add_data_vector(
        solution,
        "solution",
        MGDataOut<dim>::type_dof_data,
        std::vector<DataComponentInterpretation::DataComponentInterpretation>(),
        level);
```
But here it shouldn't be really necessary to have the level, should it?

tjhei:
Oh no. I have had a "working" version for outputting multigrid solutions sitting around for a long time. Here is the branch: https://github.com/tjhei/dealii/tree/mg_data_out_clean

This is supposed to be part of step-50...

peterrum:
@tjhei You are kidding ;)

Rombur:
Is it too late for me to start my own PR? Since all the cool kids are doing it...

peterrum:
> Is it too late for me to start my own PR? Since all the cool kids are doing it...

@Rombur  Was this a joke? If not could you share your code so that we can select from 4 options with which we proceed...

Rombur:
Yes, it was a joke :wink:

tamiko:
Amazing :smiley: I will be coming back to this little thing later today (teaching duties + something urgent this afternoon). My suggestion is to simply go with the most complete implementation (mine is clearly losing..) and a thorough review.

tamiko:
@tjhei @peterrum Had a one minute look at Timo's branch. This is more or less what I had in mind.

tamiko:
@tjhei Would you mind if I rebase your changeset on top of my PR branch?

PR #9104, opened 2019-11-27
Allow using float shape functions

masterleinad:
Previously, we were not able to compile with `float` as number type since `global_co_shape_gradients`/`global_shape_gradients`/`global_shape_values` were `double` arrays.
This pull request fixes the issue by duplicating the storage for `float`.

### DIFF
--- a/include/deal.II/matrix_free/cuda_matrix_free.templates.h
+++ b/include/deal.II/matrix_free/cuda_matrix_free.templates.h
@@ -46,16 +46,80 @@ namespace CUDAWrappers
   namespace internal
   {
     // These variables are stored in the device constant memory.
-    // TODO: use a template parameter
     constexpr unsigned int max_elem_degree = 10;
+
+    template <typename NumberType>
+    using DataArray = NumberType[(max_elem_degree + 1) * (max_elem_degree + 1)];
+
     __constant__ double
-      global_shape_values[(max_elem_degree + 1) * (max_elem_degree + 1)];
+      global_shape_values_d[(max_elem_degree + 1) * (max_elem_degree + 1)];
+    __constant__ float
+      global_shape_values_f[(max_elem_degree + 1) * (max_elem_degree + 1)];
+
+    template <typename Number>
+    __host__ __device__ inline DataArray<Number> &
+             get_global_shape_values();
+
+    template <>
+    __host__ __device__ inline DataArray<double> &
+             get_global_shape_values<double>()
+    {
+      return global_shape_values_d;
+    }
+
+    template <>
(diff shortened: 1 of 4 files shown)


include/deal.II/matrix_free/cuda_matrix_free.templates.h
```diff
@@ -46,16 +46,76 @@ namespace CUDAWrappers
   namespace internal
   {
     // These variables are stored in the device constant memory.
-    // TODO: use a template parameter
     constexpr unsigned int max_elem_degree = 10;
     __constant__ double
-      global_shape_values[(max_elem_degree + 1) * (max_elem_degree + 1)];
+      global_shape_values_d[(max_elem_degree + 1) * (max_elem_degree + 1)];
+    __constant__ float
+      global_shape_values_f[(max_elem_degree + 1) * (max_elem_degree + 1)];
```
peterrum: This looks ugly. Any chance to create a struct like:
```cpp
template<typename T>
struct ShapeData
{
    T global_shape_values[(max_elem_degree + 1) * (max_elem_degree + 1)];
    T global_shape_gradients[(max_elem_degree + 1) * (max_elem_degree + 1)];
    T global_co_shape_gradients[(max_elem_degree + 1) * (max_elem_degree + 1)];
}
```

In the long term, it would be nice to be able to write ` __constant__ ShapeData<Number, degree, q_points>` or s.th. similar. It would be nice to have as much flexibility as in the case of the CPU version.

masterleinad: Not a bad suggestion even if helps only slightly. We still need separate global `__constant__` variables (function scope `static` variables are also not allowed) and this is also the reason that we can't templatize on `degree` and `n_q_points` in a meaningful way. These arrays have to exist independent of the polynomial degree. Of course, we could create separate arrays, but that would only waste space.

masterleinad:
I had to discard the `struct` again to make it actually work.

include/deal.II/matrix_free/cuda_matrix_free.templates.h
masterleinad: I give up on improving this. Accessing `__constant__` memory seems to be really brittle. If you pass a pointer instead of a reference to a C-style array or try to use a `struct` for structuring the data better, you always get an `invalid device symbol` error.

peterrum: > If you pass a pointer instead of a reference to a C-style

I also tried that out some time ago; I had the same conclusion :( Thanks for trying out the struct approach! Let's merge!

PR #9106, opened 2019-11-28
Make pft independent of p4est

peterrum:
... for this I have to make `Policy::ParallelDistributed` independent from `p4est`. As far as I can see, we might have to replace some data structures.


peterrum:
This one is ready to merged. `p:f:t` is working now with `-D DEAL_II_WITH_P4EST="OFF"`!

peterrum:
Since I only renamed some tests and these are not by CI, I guess we can directly merge...

peterrum:
Thanks @kronbichler !

PR #9072, opened 2019-11-21
Add simple output for particles.

luca-heltai:
This class manages the DataOut of a Particle Handler
It currently only supports writing the particle position
and their ID

PR #9110, opened 2019-11-29
Allow using serial triangulations with ParticleHandler

### LINKED ISSUE #9046: Inconsistent ParticleHandler numbering
While working with @blaisb, we found out that the `insert_particles` method is inconsistent w.r.t. numbering of the particles. In particular, lines 136 to 139 of particle_handler.cc show that if we have no particles, the next_free_particle_index will be 1, instead of 0 (the logic is correct whenever we have particles, but not if no particles are present yet), meaning that the indices of particles, when adding particles using `insert_particles` that takes a vector of locations will have their ids start at one. 

This is because in the `insert_particles` that takes a vector of locations the `update_cached_numbers` is called *both* at the beginning (line 314) and at the end (line 365), while in the `insert_particles` that takes a vector of particles it is only called at the end.

Is there a reason to call `update_cached_numbers` at the beginning of `insert_particles`? 

@gassmoeller

masterleinad:
Throughout the implementation of the ParticleHandler class, we have multiple checks for `DEAL_II_WITH_MPI`, but we guard the whole implementation with `DEAL_II_WITH_P4EST` making these checks superfluous. It appears that removing this restriction doesn't require that much work. Most of the member functions also work with a serial Triangulation object and the others are only required if running with multiple MPI processes.

Related to https://github.com/dealii/dealii/pull/9072#issuecomment-559133775.

Also fixes #9046.


source/particles/particle_handler.cc
```diff
@@ -189,16 +186,26 @@ namespace Particles
...
+            locally_highest_index, parallel_triangulation->get_communicator()) +
+          1;
+        global_max_particles_per_cell = dealii::Utilities::MPI::max(
+          local_max_particles_per_cell,
+          parallel_triangulation->get_communicator());
+      }
+    else
+      {
+        global_number_of_particles    = particles.size();
+        next_free_particle_index      = locally_highest_index + 1;
```
luca-heltai: make this `particles.size() == 0 ? 0 : locally_highest_index + 1;`

source/particles/particle_handler.cc
```diff
@@ -189,16 +186,26 @@ namespace Particles
...
-    global_max_particles_per_cell =
-      dealii::Utilities::MPI::max(local_max_particles_per_cell,
-                                  triangulation->get_communicator());
+    if (const auto parallel_triangulation =
+          dynamic_cast<const parallel::Triangulation<dim, spacedim> *>(
+            &*triangulation))
+      {
+        global_number_of_particles = dealii::Utilities::MPI::sum(
+          particles.size(), parallel_triangulation->get_communicator());
+        next_free_particle_index =
```
luca-heltai: `next_free_particle_index = (global_number_of_particles == 0 ? 0 : dealii::Utilities::MPI::max(
            locally_highest_index, parallel_triangulation->get_communicator()) + 1);`

luca-heltai:
Feel free to ignore my two comments above, or to insert them also in the `update_cached_numbers`. This could go into another PR, and would fix a bug @blaisb and myself discovered a couple of weeks ago.

masterleinad:
I rebased and removed the restrictions for `Generators` and `DataOut` as well.

PR #9115, opened 2019-11-30
Do not export bundled subdirectory if empty

### LINKED ISSUE #9114: include/deal.II/bundled added to DEAL_II_INCLUDE_DIRS in deal.IIConfig.cmake even when it doesn't exist
If deal.II is compiled with DEAL_II_ALLOW_BUNDLED=Off (as it is done by the spack package), the include/bundled directory is never created. It is however added to the list of include directories in DEAL_II_INCLUDE_DIRS. This is not a problem when setting up a target via dealii_setup_target, but when I create a dependent library that exports this directory in its INTERFACE_INCLUDE_DIRECTORIES property (which in my honest opinion the targets exported by deal.II already should do), CMake complains with an error message when using that derived library

```
CMake Error in CMakeLists.txt:
  Imported target "[derived library name]" includes non-existent path

    "[deal.II install directory]/include/deal.II/bundled"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

The easiest solution to this would probably to create an empty include/bundled folder just in case, but I don't know where to best put that in the cmake scripts.

tamiko:
Avoid recording a nonexistent include directory in case we do not
install any bundled header files.

Fixes #9114

### DIFF
--- a/cmake/config/CMakeLists.txt
+++ b/cmake/config/CMakeLists.txt
@@ -150,10 +150,20 @@ CONFIGURE_FILE(
 # For installation:
 #
 
+# Make sure that we only ever record the bundled subdirectory if it is in
+# fact created:
+IF("${DEAL_II_BUNDLED_INCLUDE_DIRS}" STREQUAL "")
+  SET(_installed_bundled "")
+ELSE()
+  SET(_installed_bundled
+    "\${DEAL_II_PATH}/\${DEAL_II_INCLUDE_RELDIR}/deal.II/bundled"
+    )
+ENDIF()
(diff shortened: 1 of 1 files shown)


masterleinad:
@alexrobomind Does this work for you?

drwells:
I verified that, with this patch, we no longer create `bundled/`.
/rebuild

PR #9121, opened 2019-12-02
Fix some to some.

luca-heltai:
When calling `Utilities::MPI::some_to_some`, if the map contained our own mpi process, an assert would be thrown stating that we cannot communicate with our own process.

This PR modifies this behaviour, and makes sure that anything that we try to send to our selves is actually simply copied in the output map, simplifying the logic behind many algorithms.

### DIFF
--- a/include/deal.II/base/mpi.h
+++ b/include/deal.II/base/mpi.h
@@ -1481,21 +1481,24 @@ namespace Utilities
     {
 #  ifndef DEAL_II_WITH_MPI
       (void)comm;
-      Assert(objects_to_send.size() == 0,
+      Assert(objects_to_send.size() < 2,
              ExcMessage("Cannot send to more than one processor."));
       Assert(objects_to_send.find(0) != objects_to_send.end() ||
                objects_to_send.size() == 0,
              ExcMessage("Can only send to myself or to nobody."));
       return objects_to_send;
 #  else
+      const auto my_proc = this_mpi_process(comm);
 
-      std::vector<unsigned int> send_to(objects_to_send.size());
-      {
-        unsigned int i = 0;
-        for (const auto &m : objects_to_send)
-          send_to[i++] = m.first;
-      }
-      AssertDimension(send_to.size(), objects_to_send.size());
+      std::vector<unsigned int> send_to;
+      send_to.reserve(objects_to_send.size());
+      for (const auto &m : objects_to_send)
+        if (m.first != my_proc)
+          send_to.emplace_back(m.first);
+
+      // Shortcut, when running in serial
+      if (send_to.size() == 0)
+        return objects_to_send;
 
       const auto receive_from =
         Utilities::MPI::compute_point_to_point_communication_pattern(comm,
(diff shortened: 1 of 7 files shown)


luca-heltai:
@blaisb , FYI. :)

blaisb:
Awesome. This will be very useful !

include/deal.II/base/mpi.h
```diff
@@ -1481,21 +1481,24 @@ namespace Utilities
...
-        unsigned int i = 0;
-        for (const auto &m : objects_to_send)
-          send_to[i++] = m.first;
-      }
-      AssertDimension(send_to.size(), objects_to_send.size());
+      std::vector<unsigned int> send_to;
+      send_to.reserve(objects_to_send.size());
+      for (const auto &m : objects_to_send)
+        if (m.first != my_proc)
+          send_to.emplace_back(m.first);
```
peterrum:
```suggestion
      for (const auto &m : objects_to_send)
        if (m.first==my_proc)
          received_objects[my_proc] = m.second;
        else
          send_to.emplace_back(m.first);
```

luca-heltai:
The failed test is unrelated to this pull request.

```

99% tests passed, 1 tests failed out of 605



Total Test time (real) = 1042.97 sec



The following tests FAILED:

	4111 - multigrid/transfer_04b.mpirun=2.debug (Timeout)

Errors while running CTest

script returned exit code 8

```

include/deal.II/base/mpi.h
```diff
@@ -1481,21 +1481,28 @@ namespace Utilities
...
+          received_objects[my_proc] = m.second;
+        else
+          send_to.emplace_back(m.first);
+
+      // Shortcut, when running in serial
+      if (send_to.size() == 0)
+        return received_objects;
 
       const auto receive_from =
         Utilities::MPI::compute_point_to_point_communication_pattern(comm,
```
masterleinad: Only the number of messages to receive is required from this function later on (l.1544). Hence, `Utilities::MPI::compute_n_point_to_point_communications` should suffice here.

luca-heltai:
> Looks good to me, but I couldn 't find the discussion you were referring to in #9069 (comment).

I could not find it either. Maybe I'm confusing this with other stuff we did in the same period... My bad.

include/deal.II/base/mpi.h
```diff
@@ -1481,25 +1481,31 @@ namespace Utilities
...
+      send_to.reserve(objects_to_send.size());
+      for (const auto &m : objects_to_send)
+        if (m.first == my_proc)
+          received_objects[my_proc] = m.second;
+        else
+          send_to.emplace_back(m.first);
+
+      // Shortcut, when running in serial
+      if (send_to.size() == 0)
+        return received_objects;
```
peterrum: This is the reason why  `multigrid/transfer_04b.mpirun=2.debug` fails. `compute_n_point_to_point_communications` is a  global communication step!

luca-heltai: I did not know we used some_to_some in multigrid transfer... :) I'm on it.

include/deal.II/base/mpi.h
```diff
@@ -1481,25 +1481,31 @@ namespace Utilities
...
+          send_to.emplace_back(m.first);
+
+      const auto n_point_point_communications =
+        Utilities::MPI::compute_n_point_to_point_communications(comm, send_to);
 
-      const auto receive_from =
-        Utilities::MPI::compute_point_to_point_communication_pattern(comm,
-                                                                     send_to);
+      // Shortcut, when we have nothing to send or nothing to receive
+      if (send_to.size() == 0 && n_point_point_communications == 0)
```
masterleinad:
```suggestion
      if (send_to.size() == 0 || n_point_point_communications == 0)
```
?

luca-heltai: Why? At least one of the two blocks below would not be visited with ||, leading to a deadlock.

masterleinad: Yes, makes sense. The sentence above confused me. I guess something like
"Shortcut, when we have neither anything to send nor anything to receive" or "Shortcut, when we have nothing to send and nothing to receive"

It's not important to me, though.

masterleinad: Thanks!

PR #9124, opened 2019-12-02
New set id method in particle.

luca-heltai:
First of two PR: next one will allow renumbering ids in particles.

### DIFF
--- a/doc/news/changes/minor/20191202LucaHeltai2
+++ b/doc/news/changes/minor/20191202LucaHeltai2
@@ -1,0 +1,3 @@
+New: Particles::Particle::set_id() allows to set a particle ID number.
+<br>
+(Luca Heltai, 2019/12/02)
--- a/include/deal.II/particles/particle.h
+++ b/include/deal.II/particles/particle.h
@@ -271,6 +271,14 @@ namespace Particles
     get_id() const;
 
     /**
+     * Set the ID number of this particle.
+     *
+     * @param[in] new_id The new ID number for this particle.
+     */
+    void
+    set_id(const types::particle_index &new_id);
+
+    /**
      * Tell the particle where to store its properties (even if it does not
      * own properties). Usually this is only done once per particle, but
      * since the particle does not know about the properties,
--- a/source/particles/particle.cc
+++ b/source/particles/particle.cc
@@ -274,6 +274,15 @@ namespace Particles
 
   template <int dim, int spacedim>
   void
+  Particle<dim, spacedim>::set_id(const types::particle_index &new_id)
+  {
+    id = new_id;
+  }
+
+
+
+  template <int dim, int spacedim>
+  void
   Particle<dim, spacedim>::set_property_pool(PropertyPool &new_property_pool)
   {
     property_pool = &new_property_pool;
--- a/tests/particles/particle_08.cc
+++ b/tests/particles/particle_08.cc
@@ -1,0 +1,53 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+
+// Test set/get id.
+
+#include <deal.II/particles/particle.h>
+
+#include "../tests.h"
+
+
+template <int dim>
+void
+test()
+{
+  {
+    Point<2> position;
+    position(0) = 0.3;
+    position(1) = 0.5;
+
+    Point<2> reference_position;
+    reference_position(0) = 0.2;
+    reference_position(1) = 0.4;
+
+    const types::particle_index index(7);
+
+    Particles::Particle<2> particle(position, reference_position, index);
+
+    deallog << "ID: " << particle.get_id() << std::endl;
+    particle.set_id(9);
+    deallog << "New ID: " << particle.get_id() << std::endl;
+  }
+}
+
+int
+main()
+{
+  initlog();
+  test<2>();
+}
--- a/tests/particles/particle_08.output
+++ b/tests/particles/particle_08.output
@@ -1,0 +1,3 @@
+
+DEAL::ID: 7
+DEAL::New ID: 9


luca-heltai:
@gassmoeller

bangerth:
Why would you ever want to do that? I've always thought of the id as a globally unique number that allows identifying particles and in that case, it seems counter-productive to allow changing the id at a later point. Everything that can be changed we thought of as a property.

Can you explain what you want to do with changing the id?

luca-heltai:
Well, in principle I would agree with you. On the other hand, we allow particles to "disappear" (we just drop 'em out when they are moved and we don't find them in out locally owned cells, or in our ghost cells). This breaks all algorithms that associate to each particle something that needs to live for the entire duration of the simulation. Allowing one to set the ids of existing particles, could in principle allow for a coherent renumbering. 

Now it is perfectly fine for `n_global_particles()` to be smaller than `next_free_particle_index()`. If, however, we assemble matrices that depend on these numbers to be same, we have to make sure that some renumbering (i.e., reindexing, in the case of particles) is possible. If we need to invert a matrix with the same size of the number of particles, we better guarantee that no lines have zero entries (entirely possible with our current implementation).

As you said, I actually think of the id associated to a particle as one of its properties. And we do allow users to change properties. Why not particle ids? We allow for dof renumbering. The ID of a particle is, in a sense, a degree of freedom associated to it. 

There are algorithms that make arbitrary decisions on what id to give particles. For example: `insert_particles` taking an `std::vector<Point<spacedim>>` associates ids based on their original numbering and on `next_free_particle_index()`. This is a very special numbering, and there is no reason why this should be the final numbering of the particles. Now the only way around this limitation is to store (in the properties) the id we want to associate to a particle. But why store an additional double for each particle, if we already have an ID? Why not just allowing users to change the ID?

For example, every once in a while, we could renumber particles following their global ordering in the ParticleHandler (making sure that you could use the particle id to construct coherent interpolation matrices).  At the moment, the only way this is doable, is by reconstructing from scratch every particle with a new id (expensive).

If you don't think this is useful, let's just drop this PR. It is not very important, after all, since there are (more expensive) ways around it. I see no harm in having it though, and it's the only setter we miss. 

BTW: if we don't want to set particle ids after construction, then the member `id` should be marked const.

bangerth:
Nice discussion!

I can see your use case for allowing *certain* kinds of changes. Since you have presumably already written this kind of function, how about adding a function `ParticleHandler::compress_particle_id_range()` that (i) enumerates particles so that they remain globally unique, (ii) eliminates all holes? I think such a function would remain on the safe side. If you need a mapping from old to new ids, the function could return that -- but then I don't know whether you actually need this kind of information.

bangerth:
There's also a merge conflict here.

luca-heltai:
I would also like to point out that I'm proposing to change the *Particle* interface. Not the *ParticleAccessor* interface, i.e., I still don't want the user to set ids externally through *ParticleHandler*. I agree that any change in all the particle ids should be done through methods like *compress_particle_ids* (in `ParticleHandler`, as you suggested) that would make all the sanity checks. I *don't* want users to do that by themselves. And at the moment, the user *cannot*, even with this change.

A *Particle*, at the moment, is agnostic of who owns it, what cell contains it, and even if there are any other particles around. Not being able to change its ID does not seem like a good design choice. I'm just specifying *how* this change should happen (i.e., through a setter, rather than only through the constructor). We already isolated *Particle* from users via the *ParticleAccessor*

BTW: I really think we should also remove the reference location from the *Particle* class. It does not belong there. A particle should really be just a position and ID. I know you added this to have a compact way to represent the data structure you needed in ASPECT, but this carries around (and forces you to **communicate**, if you send the entire particle) **dim more doubles per particle**, that very likely you'll have to compute again as soon as the particle has been communicated.

We isolated the user from the particle using the *ParticleAccessor*. This could be a good time to fix this issue (in another PR and discussion, though...).

gassmoeller:
> I would also like to point out that I'm proposing to change the Particle interface. Not the ParticleAccessor interface, i.e., I still don't want the user to set ids externally through ParticleHandler. I agree that any change in all the particle ids should be done through methods like compress_particle_ids (in ParticleHandler, as you suggested) that would make all the sanity checks. I don't want users to do that by themselves. And at the moment, the user cannot, even with this change.

Very good point, you are right. Let's merge this PR.

> BTW: I really think we should also remove the reference location from the Particle class. It does not belong there. A particle should really be just a position and ID. I know you added this to have a compact way to represent the data structure you needed in ASPECT, but this carries around (and forces you to communicate, if you send the entire particle) dim more doubles per particle, that very likely you'll have to compute again as soon as the particle has been communicated.

I am not opposed to making the reference position an optional property, and we can make that a separate PR, but I think it is a very useful property for many applications so maybe we can create some convenience functions to use it (ideally without breaking compatibility, but I am fine to make some adjustments to ASPECT as well). Also if the particle was in a ghost cell it will have the same reference location on the active cell it is transferred to, so I never transmit outdated data (although I understand that in your application that is different because particles start in artificial cells). In my scaling tests it was consistently cheaper to store the reference location and ship it around than to recompute it (except for MappingEulerian which was cheaper to compute; MappingQ1 in 2D was around the same cost as communicating if I remember correctly, and everything higher order is more expensive to compute than communicating).

luca-heltai:
> I am not opposed to making the reference position an optional property, and we can make that a separate PR, but I think it is a very useful property for many applications so maybe we can create some convenience functions to use it (ideally without breaking compatibility, but I am fine to make some adjustments to ASPECT as well). 

I agree it is a useful (fundamental!) property. But it does not belong to the Particle class, unless you add to the class also the cell it refers to. I understand that you keep the cell as the key in the multimap, and this is a nice and compact way to store things, but a Particle may exist on its own (outside of the multimap), and it has no concept of a triangulation, nor of a cell, therefore it should have no concept of a reference location.

Maybe a more coherent structure in ParticleHandler could be:

`std::multimap<internal::LevelInd, std::pair<Particle<dim, spacedim>, Point<dim>>`

where the reference position is associated to each entry of the map and not to the particle itself.

Alternatively, we could store the reference position in the particle properties by default.

luca-heltai:
Ping @bangerth ?

> Very good point, you are right. Let's merge this PR.

luca-heltai:
With a really trivial change (i.e., a tiny wrapper as in `https://github.com/dealii/dealii/blob/master/include/deal.II/boost_adaptors/point.h`) it is possible to feed vectors of particles to packing algorithms (i.e., boost::rtree) to optimize and decrease the sorting time of (many) particles into cells and processors. This can be done before any `Mapping::real_to_unit` occurs, and, combined with an index tree for the cells, would allow a (theoretical) cost of distribution of all particles in all cells at `O(N*log(N)+M*log(M) + log(N)*log(M))` operations, instead of the (theoretical) `O(N*log(N)+M*log(M) +N*log(M)+M*log(N))`, where `N` random particles are distributed in `M` cells.

Bringing around a reference location for every particle is just a waste (it basically halves the cache locality of vectors of particles).

bangerth:
Ah, very cool!

PR #9125, opened 2019-12-02
Feature tensor function parser

konsim83:
Wrote a parser to create a TensorFunction similar to FunctionParser. 

All functions (internal::mu_XXX() ) defined in the internal namespace and used in  base/function_parser.cc and in base/tensor_function_parser.cc are moved to a separate file base/mu_parser_internal.h (since other wise the linker see doubly defined symbols)

The new TensorFunctionParser class inherits from TensorFunction and overloads the value() and value_list() functions. 

I wrote doxygen code to document the usage that should be self contained (although it sometimes refers to the FunctionParser class) and 4 tests that all pass. All test for the FunctionParser class still pass after adding the feature.


include/deal.II/base/tensor_function.h
```diff
@@ -181,7 +181,7 @@ class ConstantTensorFunction : public TensorFunction<rank, dim, Number>
  * @ingroup functions
  * @author Matthias Maier, 2013
  */
-template <int rank, int dim, typename Number = double>
+template <int rank, int dim, typename Number = float>
```
drwells: Why was this changed?

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ *
+ * // TensorFunctionParser with 2+1 variables (space + time) in 2D of rank 2:
+ * TensorFunctionParser<2,2> tfp;
+ * tfp.initialize(variables,
+ *               expression,
+ *               constants,
+ *               true);       // This tells the parser that
+ *                            // it is a time-dependent function
+ *                            // and there is another variable
+ *                            // to be taken into account (t).
```
drwells: Could you move this comment up a line (i.e., right before the line with `true`) so that we don't have to split it across so many lines?

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ *         << result[0][0] << " " << result[0][1]
+ *         << result[1][0] << " " << result[1][1]
+ *         << std::endl;
+ * @endcode
+ *
+ * See also the documentation of the FunctionParser class.
+ *
+ * @note This class implements the list of functions just mentioned as
+ *   user-defined functions by extending muparser. This means, in particular,
+ *   that the `if(condition, then-value, else-value)` syntax evaluates all
```
drwells: If I recall correctly the triple-if statement was removed in a recent version of muParser: we should just use the `?` and `:` replacement instead.

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+  /**
+   * Destructor. Explicitly delete the TensorFunctionParser objects (there is one
+   * for each component of the function).
+   */
+  ~TensorFunctionParser() override;
+
+  /**
+   * Type for the constant map. Used by the initialize() method.
+   */
+  using ConstMap = std::map<std::string, double>;
```
drwells: I'd prefer to not introduce the extra type alias - it makes the other declarations harder to read.

Either way, I think that this should be templated on `Number`, right?

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,419 @@
...
+#  include <muParser.h>
+#else
+
+
+
+namespace fparser
+{
+  class TensorFunctionParser
+  {};
+} // namespace fparser
```
drwells: What is this class for?

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,419 @@
...
+  , initialized(false)
+  , n_vars(0)
+  , n_components(std::pow(dim,rank))
+{}
+
+
+template <int rank, int dim, typename Number>
+TensorFunctionParser<rank, dim, Number>::TensorFunctionParser(const std::string &expression,
+														const std::string &constants,
+														const std::string &variable_names)
```
drwells: I think you used tabs instead of spaces. running clang-format should fix this.

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,419 @@
...
+                                          "int",
+                                          "ceil",
+                                          "cot",
+                                          "csc",
+                                          "floor",
+                                          "sec",
+                                          "pow",
+                                          "erfc",
+                                          "rand",
+                                          "rand_seed"};
```
drwells: Could we move this array to the muparser header?

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,419 @@
...
+             Utilities::split_string_list(variable_names, ",").size() ==
+               dim + 1);
+}
+
+
+// We deliberately delay the definition of the default destructor
+// so that we don't need to include the definition of mu::Parser
+// in the header file.
+template <int rank, int dim, typename Number>
+TensorFunctionParser<rank, dim, Number>::~TensorFunctionParser() = default;
```
drwells: Why can't we just use the default destructor from the beginning (i.e., don't declare it at all)? I don't understand why this is necessary.

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,419 @@
...
+      tfp.get()[component]->DefineFun("ceil", internal::mu_ceil, true);
+      tfp.get()[component]->DefineFun("cot", internal::mu_cot, true);
+      tfp.get()[component]->DefineFun("csc", internal::mu_csc, true);
+      tfp.get()[component]->DefineFun("floor", internal::mu_floor, true);
+      tfp.get()[component]->DefineFun("sec", internal::mu_sec, true);
+      tfp.get()[component]->DefineFun("log", internal::mu_log, true);
+      tfp.get()[component]->DefineFun("pow", internal::mu_pow, true);
+      tfp.get()[component]->DefineFun("erfc", internal::mu_erfc, true);
+      tfp.get()[component]->DefineFun("rand_seed", internal::mu_rand_seed, true);
+      tfp.get()[component]->DefineFun("rand", internal::mu_rand, true);
```
drwells: Could we put a function that does this inside the internal muparser header?

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ *   enabled will abort the program with a floating point exception when
+ *   evaluated with a negative `x`. This is because the square root of
+ *   `x` is computed before the `if` statement's condition is considered
+ *   to determine whether the result should be the second or third
+ *   argument. If this kind of behavior is a problem, you can resort to
+ *   the muparser built-in syntax `(condition ? then-value : else-value)`,
+ *   using the ternary syntax familiar to C++ programmers. If this
+ *   syntax is used, muparser uses lazy evaluation in which only one of the
+ *   branches is evaluated, depending on whether the `condition` is
+ *   true or false.
```
drwells: I think it would be better to refer to the documentation of `FunctionParser` instead of copying and pasting this block.

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+
+  /**
+   * Type for the constant map. Used by the initialize() method.
+   */
+  using ConstMap = std::map<std::string, double>;
+
+  /**
+   * Iterator for the constants map. Used by the initialize() method.
+   */
+  using ConstMapIterator = ConstMap::iterator;
```
drwells: I don't think this is used anywhere.

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+   * then an exception is thrown.
+   */
+  TensorFunctionParser(const double initial_time = 0.0);
+
+  /**
+   * Constructor for parsed functions. This object needs to be initialized
+   * with the initialize() method before you can use it. If an attempt to
+   * use this function is made before the initialize() method has been called,
+   * then an exception is thrown.
+   * Takes a semi-colon separated list of expressions (one for each component
```
drwells:
```suggestion
   * Takes a semicolon separated list of expressions (one for each component
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+
+  /**
+   * Iterator for the constants map. Used by the initialize() method.
+   */
+  using ConstMapIterator = ConstMap::iterator;
+
+  /**
+   * Initialize the tensor function.  This methods accepts the following parameters:
+   *
+   * <b>vars</b>: a string with the variables that will be used by the
```
drwells: Please use doxygen's `@param[in]` syntax instead.

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+   */
+  using ConstMap = std::map<std::string, double>;
+
+  /**
+   * Iterator for the constants map. Used by the initialize() method.
+   */
+  using ConstMapIterator = ConstMap::iterator;
+
+  /**
+   * Initialize the tensor function.  This methods accepts the following parameters:
```
drwells:
```suggestion
   * Initialize the tensor function.  This constructor accepts the following parameters:
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+   * <b>vars</b>: a string with the variables that will be used by the
+   * expressions to be evaluated. Note that the variables can have any name
+   * (of course different from the function names defined above!), but the
+   * order IS important. The first variable will correspond to the first
+   * component of the point in which the function is evaluated, the second
+   * variable to the second component and so forth. If this function is also
+   * time dependent, then it is necessary to specify it by setting the
+   * <code>time_dependent</code> parameter to true.  An exception is thrown if
+   * the number of variables specified here is different from dim (if this
+   * function is not time-dependent) or from dim+1 (if it is time- dependent).
```
drwells:
```suggestion
   * function is not time-dependent) or from dim+1 (if it is time-dependent).
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+   *
+   * <b>expressions</b>: a list of strings containing the expressions that
```
drwells:
```suggestion
   * <b>expressions</b>: a vector of strings containing the expressions that
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+   * and this->get_time() is used to initialize it when evaluating the
+   * function. Naturally the number of variables parsed by the initialize()
+   * method in this case is dim+1. The value of this parameter defaults to
+   * false, i.e. do not consider time.
+   */
+  void
+  initialize(const std::string &             vars,
+             const std::vector<std::string> &expressions,
+             const ConstMap &                constants,
+             const bool                      time_dependent = false);
```
drwells: It seems odd that `t` is automatically added as a variable for the constructor but here we set `time_dependent = false`. We should set these consistently.

drwells:
It looks like a lot of the things I commented on are really issues with `FunctionParser`. I'll put up a little PR that implements some of the fixes there :)

include/deal.II/base/tensor_function_parser.h
drwells: Nevermind - `FunctionParser` also has this inconsistency. We should match that classes' behavior.

include/deal.II/base/tensor_function_parser.h
drwells: Since the documentation here is the same as that of FunctionParser it would be better to write something like "this has the same arguments as FunctionParser::initialize()" instead of copying and pasting.

source/base/tensor_function_parser.cc
drwells: I removed this in `function_parser.cc` - AFAICT it doesn't do anything there either.

include/deal.II/base/tensor_function_parser.h
drwells: I reread the code and this comment is correct, though we are a bit tricky here: we reimplement the old ternary if function ourselves to resolve the compatibility issue.

Either way, we should just refer to the documentation in function_parser.h.

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,94 @@
...
+#include <vector>
+
+
+DEAL_II_NAMESPACE_OPEN
+
+
+
+#ifdef DEAL_II_WITH_MUPARSER
+
+namespace internal
```
drwells: We should at least add a comment for this file (not the namespace) explaining why this header is necessary.

include/deal.II/base/tensor_function_parser.h
konsim83: Done.

include/deal.II/base/tensor_function_parser.h
konsim83: I removed the note from the header and refer to FunctionParser.

include/deal.II/base/tensor_function_parser.h
konsim83: Wait. I read the code in function_parser.cc that the input (see constructor) is being converted map. There Patterns:Double() is used. So I guess the use of double in the typedef is on purpose?

source/base/tensor_function_parser.cc
konsim83: I moved it to the header. I also changed that in function_parser.cc

source/base/tensor_function_parser.cc
konsim83: I left it as is since I saw that you opened another issue #9143

include/deal.II/base/tensor_function_parser.h
konsim83: Removed it.

source/base/tensor_function_parser.cc
konsim83: Done

source/base/tensor_function_parser.cc
konsim83: Did it.

source/base/tensor_function_parser.cc
konsim83: Can that be taken care of together with the clean up of function_parser.cc/h?
See issue #9143

include/deal.II/base/tensor_function_parser.h
konsim83: I removed the note completely

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,137 @@
...
+    "int",
+    "ceil",
+    "cot",
+    "csc",
+    "floor",
+    "sec",
+    "pow",
+    "erfc",
+    "rand",
+    "rand_seed"};
```
drwells: I am surprised that this works - I think that we need to have an `extern` declaration of the array here and to move the definition into the source file.

konsim83: Yes, I forgot. And it does not compile. OK, I think I solved the problem (but a slightly different): I created (in the internal namespace) a class to hold internal function names. Since objects of this class share the data it is declared as static const. The strategy to use the extern specifier turned out to be not suitable since it messes with range based for loops used in (tensor_)function_parser.cc. I see that tests on my machine pass.

source/base/mu_parser_internal.cc
```diff
@@ -0,0 +1,220 @@
...
+    static boost::random::uniform_real_distribution<> uniform_distribution(0,
+                                                                           1);
+    static boost::random::mt19937                     rng(
+                          static_cast<unsigned long>(std::time(nullptr)));
+    return uniform_distribution(rng);
+  }
+
+  // helper function to determine the past-the-end pointer of an array
+  template<typename T, std::size_t N>
+  T* past_the_end_ptr(T (&ra)[N])
```
masterleinad: There is `std::end()` defined in header `<iterator>` that provides the same functionality.

source/base/function_parser.cc
```diff
@@ -330,48 +207,12 @@ FunctionParser<dim>::init_muparser() const
...
-                                          "sec",
-                                          "pow",
-                                          "erfc",
-                                          "rand",
-                                          "rand_seed"};
-          for (const auto &function_name_c_string : function_names)
+          internal::MuParserFunctionNames function_names;
+          for (auto function_name_it = function_names.begin();
+        		  function_name_it != function_names.end();
+        		  ++function_name_it)
```
masterleinad: Is there any reason not to use a range-based for loop here (as was done previously)?

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,141 @@
...
+  // returns a random value in the range [0,1]
+  double
+  mu_rand();
+
+  /**
+   * This is only a helper function to determine the past-the-end pointer
+   * of an array of arrays of type T.
+   */
+  template<typename T, std::size_t N>
+  T* end(T (&ra)[N]);
```
masterleinad: `std::end()`.

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,390 @@
...
+          // compatible to the library we used to use before muparser
+          // (the tfparser library) but also makes no real sense.
+          // consequently, in the expressions we set, remove any space
+          // we may find after function names
+          std::string transformed_expression = expressions[component];
+
+          internal::MuParserFunctionNames function_names;
+          for (auto function_name_it = function_names.begin();
+			  function_name_it != function_names.end();
+			  ++function_name_it)
```
masterleinad: This also looks like a place a range-based for loop can be used.

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,336 @@
...
+  TensorFunctionParser(
+    const std::string &expression,
+    const std::string &constants      = "",
+    const std::string &variable_names = default_variable_names() + ",t");
+
+  /**
+   * Destructor. Explicitly delete the TensorFunctionParser objects (there is
+   * one for each component of the function).
+   */
+  ~TensorFunctionParser() override;
```
bangerth: We generally explicitly mark it as `virtual`, even if it's implied.

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * Object to hold a list of function names used internally
+   * and by MuParser. Const iterators can be used to iterate through
+   * the names. Modification is not allowed.
+   */
+  class MuParserFunctionNames
```
drwells: This works but I would prefer to simply mark the array as `extern` and put its definition in the source file instead of creating a new class.

konsim83: I am struggeling with it a bit since marking it as extern in  mu_parser_internal.h and defining it in the corresponding.cc (I hope I understood you correctly) will compile this translation unit but in (tensor_)function_parser.cc I get the error
<code> range-based ‘for’ expression of type ‘const char* []’ has incomplete type </code> in the range based for loop. The compiler seems to want to know the size of the array explicitly (?). With the class I could avoid that.

konsim83: I simply made the const `const char* const function_names_init[];` a `std::vector<std::string>`, mark it extern and initialize it in the source file. That seems to work and avoids the class.

include/deal.II/base/tensor_function_parser.h
konsim83: Done. Also in function_parser.h.

source/base/tensor_function_parser.cc
konsim83: OK, I am using a range based for loop now.

source/base/mu_parser_internal.cc
```diff
@@ -0,0 +1,220 @@
...
+      			"rand",
+      			"rand_seed"};
+
+  // Declaration must be here since member is static and non-integral
+  // and can not be done in a member initializer list.
+  const std::vector<std::string>
+  MuParserFunctionNames::fnct_names(function_names_init, past_the_end_ptr(function_names_init));
+
+  // Constructor
+  MuParserFunctionNames::MuParserFunctionNames ()
```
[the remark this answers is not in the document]
konsim83: Class removed. Found a simpler way.

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,141 @@
...
+	  /**
+	   * Use only one instance of function names since it is the same
+	   * for all objects using the interface to MuParser.
+	   */
+      static const char* const function_names_init[];
+
+      /**
+	   * List of function names to be initialized from the function_names_init list.
+	   */
+	  static const std::vector<std::string> fnct_names;
```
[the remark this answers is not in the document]
konsim83: Sorry, I forgot to run the formatting. Fixed

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,141 @@
...
+	  /**
+	   * Return a const_iterator to part-the-end of the list of function names.
+	   */
+	  std::vector<std::string>::const_iterator end() const;
+  private:
+	  /**
+	   * Use only one instance of function names since it is the same
+	   * for all objects using the interface to MuParser.
+	   */
+      static const char* const function_names_init[];
```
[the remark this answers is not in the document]
konsim83: Removed the class

include/deal.II/base/mu_parser_internal.h
konsim83: Thanks for the hint but I removed the class.

konsim83:
I applied the requested changes. Sorry for too many commits. There is a merge conflict that popped up for function_parser.cc since it seems that someone changed something on master (deleted the copy and move constructors). I applied the same to the in tensor_function_parser.cc and to function_parser.cc in my branch.

include/deal.II/base/tensor_function_parser.h
bangerth: Ah, thanks for looking there as well!

include/deal.II/base/mu_parser_internal.h
bangerth: Yes, I would have suggested that next ;-)

masterleinad:
> I applied the requested changes. Sorry for too many commits. There is a merge conflict that popped up for function_parser.cc since it seems that someone changed something on master (deleted the copy and move constructors).

Would you please fix the merge conflict? :wink:
Don't worry about the number of commits. It would be great if you could squash your commits in the end.

konsim83:
I fixed the merge conflict. Can you please let me know when you want me to squash the commits? (Never did that before)

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,390 @@
...
+  tfp.get().reserve(this->n_components);
+  vars.get().resize(var_names.size());
+  for (unsigned int component = 0; component < this->n_components; ++component)
+    {
+      tfp.get().emplace_back(new mu::Parser());
+
+      for (std::map<std::string, double>::const_iterator constant =
+             constants.begin();
+           constant != constants.end();
+           ++constant)
```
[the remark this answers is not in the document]
masterleinad: How is this resolved?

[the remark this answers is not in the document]
konsim83: Oh, sorry. Missed that. Fixed it to a range loop.

source/base/tensor_function_parser.cc
konsim83: Removed form tensor_function_parser.cc as well.

include/deal.II/base/tensor_function_parser.h
konsim83: Done.

include/deal.II/base/tensor_function_parser.h
konsim83: Done.

source/base/mu_parser_internal.cc
konsim83: Class removed.

source/base/function_parser.cc
konsim83: Changed it to range based loop.

source/base/function_parser.cc
```diff
@@ -309,48 +186,9 @@ FunctionParser<dim>::init_muparser() const
...
-                                          "sec",
-                                          "pow",
-                                          "erfc",
-                                          "rand",
-                                          "rand_seed"};
-          for (const auto &function_name_c_string : function_names)
+          for (const auto &current_function_name : internal::function_names)
             {
-              const std::string  function_name        = function_name_c_string;
+              const std::string  function_name(current_function_name);
```
masterleinad: `clang-tidy`: You don't need the copy here. `current_function_name` is already of type `std::string`.

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,375 @@
...
+          // parenthesis. this is awkward because it is not backward
+          // compatible to the library we used to use before muparser
+          // (the tfparser library) but also makes no real sense.
+          // consequently, in the expressions we set, remove any space
+          // we may find after function names
+          std::string transformed_expression = expressions[component];
+
+          for (const auto &current_function_name : internal::function_names)
+            {
+              const std::string  function_name(current_function_name);
```
masterleinad: the same here.

masterleinad:
It seems that your commit doesn't include your name properly:
```
invalid author 'konsim83' without firstname and lastname

```

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,375 @@
...
+
+
+
+template <int rank, int dim, typename Number>
+TensorFunctionParser<rank, dim, Number>::TensorFunctionParser(
+  const double initial_time)
+  : TensorFunction<rank, dim, Number>(initial_time)
+  , initialized(false)
+  , n_vars(0)
+  , n_components(std::pow(dim, rank))
```
masterleinad:
```suggestion
  , n_components(Utilities::pow(dim, rank))
```
to avoid the implicit conversion from floating-point number into integer (`clang` warns about that).

source/base/tensor_function_parser.cc
```diff
@@ -0,0 +1,375 @@
...
+
+template <int rank, int dim, typename Number>
+TensorFunctionParser<rank, dim, Number>::TensorFunctionParser(
+  const std::string &expression,
+  const std::string &constants,
+  const std::string &variable_names)
+  : TensorFunction<rank, dim, Number>()
+  , initialized(false)
+  , n_vars(0)
+  , n_components(std::pow(dim, rank))
```
masterleinad: The same here.

source/base/function_parser.cc
konsim83: Done.

source/base/tensor_function_parser.cc
konsim83: Done.

source/base/tensor_function_parser.cc
konsim83: Done.

source/base/tensor_function_parser.cc
konsim83: Done.

peterrum:
The problem is that you have old commits with your github name. The following command `git log --author="konsim83"` shows me:
```
commit afb9bea11307029d37408306a91f4a52e539b34c
Author: konsim83 <konrad.simon@uni-hamburg.de>
Date:   Fri Nov 29 12:38:11 2019 +0100

    Moved "internal" namespace to separate file since it is share with
    function_parser.cc. Initialization still fails.

commit f566726253497dc754ab0b6ffbecb845735c5c0a
Author: konsim83 <konrad.simon@uni-hamburg.de>
Date:   Thu Nov 28 19:40:22 2019 +0100

    Finction compiles, now test.

commit 39ce55ecd27f67a918c143faebc75885d86119f3
Author: konsim83 <konrad.simon@uni-hamburg.de>
Date:   Thu Nov 28 18:11:51 2019 +0100

    Find a way how to initialize an array from a C-style array of Numbers.

commit d23b7c82878e3e35e2c35367340a9ba68b6fa71a
Author: konsim83 <konrad.simon@uni-hamburg.de>
Date:   Wed Nov 27 22:05:44 2019 +0100

    Added tensor function parser and first test.

commit 59515663d927bd48a1ca2236eb4dadbd516bd9aa
Author: konsim83 <konrad.simon@uni-hamburg.de>
Date:   Mon Sep 16 23:15:01 2019 +0200

    Setup new feature branch for tensor function parser
```
My suggestion would be checkout the current version of master again and apply your changes via cherry-pick to that and especially to squash all commit to a single one.  Esp. never merge with master but rather rebase!

masterleinad:
The homepage we are linking to from the one the indenting script prints also contains information about how to fix the ownership information for past commits (https://www.git-tower.com/learn/git/faq/change-author-name-email).

It is also fine to squash the commits and then just fix this only commit.

konsim83:
Thank you @peterrum. I refreshed the source of my fork of dealii, made an interactive rebase for my feature branch and resoved the merge conflicts. After I squashed all commits I changed the author on the root of my branch. Looks clean to me now.

The `make indent` command does not trigger complaints anymore. Should I take care of something else before I push the changes to my repository? 

(Sorry for the git-questions. I just did not collaborate on a large project before. Your documentation is basically excellent but even there I do not find every answer and google often spits out bulls**t)

peterrum:
I would squash both commits and add a more useful commit message ;)

> (Sorry for the git-questions. I just did not collaborate on a large project before. Your documentation is basically excellent but even there I do not find every answer and google often spits out bulls**t)

You're welcome. Hope this is not your last PR!

masterleinad:
Please provide a changelog entry in `doc/news/changes/minor` as well.

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,100 @@
...
+  // returns a random value in the range [0,1] initializing the generator
+  // with the given seed
+  double
+  mu_rand_seed(double seed);
+
+  // returns a random value in the range [0,1]
+  double
+  mu_rand();
+
+  extern std::vector<std::string> function_names;
```
bangerth: The internal namespace is, in essence, a global namespace for functions and classes throughout deal.II. That's ok for the functions above, since they are all prefixed with `mu_` and so, in some sense, namespace themselves. But `function_names` is a different issue: It's a generic name without explanation which part of deal.II it belongs to. I think all of what you have here should actually be in a further nested namespace `internal::FunctionParser` to make things clearer.

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+template <typename>
+class Vector;
+#endif
+
+/**
+ * This class implements a tensor function object that gets its value by parsing
+ * a string describing this function. It is a wrapper class for the muparser
+ * library (see http://muparser.beltoforion.de/). This class is essentially an
+ * extension of the FunctionParser class to read in a TensorFunction. The class
+ * reads in an expression of length dim<sup>rank<\sup> (separated by a
```
bangerth:
```suggestion
 * reads in an expression of length dim<sup>rank</sup> (separated by a
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ *
+ * @note In contrast to the FunctionParser class the TensorFunctionParser class does not support
+ * 	 automatic differentiation.
+ *
+ * A minimal example for the usage of the class would be:
+ * @code
+ * // set up time dependent tensor function:
+ * std::string variables = "x,y,t";
+ * std::string expression =
+ * "exp(-t)*cos(x+y);-sin(pi*x*y-t);sin(pi*x*y-t);exp(t)*cos(x+y)";
```
bangerth:
```suggestion
 *       "exp(-t)*cos(x+y);-sin(pi*x*y-t);sin(pi*x*y-t);exp(t)*cos(x+y)";
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ * semicolon) where the components of the tensor function are filled according
+ * to the C++ convention (fastest index is the most right one).
+ *
+ * @note In contrast to the FunctionParser class the TensorFunctionParser class does not support
+ * 	 automatic differentiation.
+ *
+ * A minimal example for the usage of the class would be:
+ * @code
+ * // set up time dependent tensor function:
+ * std::string variables = "x,y,t";
```
bangerth:
```suggestion
 * const std::string variables = "x,y,t";
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ * std::string expression =
```
bangerth:
```suggestion
 * const std::string expression =
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ * "exp(-t)*cos(x+y);-sin(pi*x*y-t);sin(pi*x*y-t);exp(t)*cos(x+y)";
+ * std::map<std::string,double> constants;
+ * constants["pi"] = numbers::PI;
+ *
+ * // TensorFunctionParser with 2+1 variables (space + time) in 2D of rank 2.
+ * It is necessary to tell the parser that there is an additional variable
```
bangerth:
```suggestion
 * // It is necessary to tell the parser that there is an additional variable
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ * to be taken into account (t).
```
bangerth:
```suggestion
 * // to be taken into account (t).
```

include/deal.II/base/tensor_function_parser.h
```diff
@@ -0,0 +1,361 @@
...
+ *
+ * // evaluate the expression at 'point':
+ * double result = tfp.value(point);
+ *
+ * deallog << "Function '" << expression << "'"
+ *         << " @ " << point
+ *         << " is "
+ *         << std::endl
+ *         << result[0][0] << " " << result[0][1]
+ *         << result[1][0] << " " << result[1][1]
```
bangerth: you probably want a separator at the end of the previous line (a space, comma, semicolon)

include/deal.II/base/mu_parser_internal.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include <deal.II/base/thread_management.h>
+
+#include <boost/random.hpp>
+
+#include <cmath>
+#include <iterator>
+#include <map>
+#include <memory>
+#include <vector>
```
bangerth: I don't think you need any of these include files.

bangerth: The only one you definitely should keep is `config.h`. You also need `<string>`.

source/base/mu_parser_internal.cc
```diff
@@ -0,0 +1,190 @@
...
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#include <deal.II/base/mu_parser_internal.h>
```
bangerth: You'll also have to `#include` the muparser header files here.

include/deal.II/base/mu_parser_internal.h
konsim83: Done.

include/deal.II/base/mu_parser_internal.h
konsim83: Yes, but then I need <cmath> in the .cc. Changed it. Thank you for the hint.

source/base/mu_parser_internal.cc
konsim83: I assume you mean the muParser.h header. If so I don't see the reason. Everything compiles and the tests run successfully on my machine.

include/deal.II/base/tensor_function_parser.h
konsim83: Done.

source/base/mu_parser_internal.cc
konsim83: If you are fine with that I will squash my commits after the last points are resolved.

doc/news/changes/minor/20191217KonradSimon
```diff
@@ -0,0 +1,4 @@
+New: The TensorFunctionParser class allows to read 
+in a function similarly to the FunctionParser class using the MuParser.
+<br>
+(Timo Heister, 2019/11/04)
```
bangerth: I think you want to give yourself credit here :-) (Unless @konsim83 is really an alias for @tjhei , though I doubt that :-) )

Please also add a newline at the end of the file.

konsim83: Oops. Copy and paste is dangerous... :-) Corrected.

konsim83:
OK, I addressed the last requests made by Wolfgang and squashed everything into one commit with a hopefully elaborate enough commit message. Sorry if some of my commits on the way contained changes that were a bit confusing/unnecessary. I learned a lot about git and development processes on the way. Thank you guys for that one. In case of any other change requests, please let me know.

bangerth:
Thanks very much for the contribution! You'll learn that over time the process is going to get easier as you anticipate the style we like to see (and as you'll become a better a better programmer -- as we all become due to patch review!). We hope for many more contributions! :-)

PR #9130, opened 2019-12-03
Use a better way to create graphical output in step-61.

bangerth:
Specifically, rather than creating two output files for the cell pressures
and the Darcy velocities, put them all into a single file using a method
that @kronbichler introduced a while ago.


bangerth:
@sophy1029 -- FYI.

/rebuild

kronbichler:
Can you please run `clang-format` @bangerth?

bangerth:
Darn, same problem every time...

Now indented correctly!

PR #9133, opened 2019-12-03
Deprecate the files in base/std_cxx11 -- we no longer need them.

bangerth:


include/deal.II/base/std_cxx11/array.h
```diff
@@ -16,6 +16,7 @@
 #ifndef dealii_std_cxx11_array_h
 #define dealii_std_cxx11_array_h
 
+#warning "This file is deprecated. Simply use the corresponding C++11 header <array>."
```
drwells: I think that for MSVC compatibility we need to use `DEAL_II_WARNING` which is in `config.h`.

include/deal.II/base/std_cxx11/bind.h
```diff
@@ -17,6 +17,8 @@
 #define dealii_std_cxx11_bind_h
 
 
+#warning "This file is deprecated. Simply use the corresponding C++11 header <functional>."
```
drwells: etc.

bangerth:
Ah, I had forgotten about `DEAL_II_WARNING`. Thanks for reminding me!

The patch is updated as requested now!

PR #9139, opened 2019-12-04
Be consistent in capitalizing text. Other minor updates.

bangerth:
In the first column, all but one entry were capitalized. This greatly irks me. While there,
also capitalize all other places where text starts. Finally, there is no need to
abbreviate 'info' -- we have plenty of space in the table heading to use proper
words in their full glory.

Follow-up to #9091.


drwells:
As always I very much appreciate the attention to detail :)

PR #9140, opened 2019-12-04
Adjusted thresholds for 'refine_and_coarsen_fixed_number'.

marcfehling:
While working on a `fixed_number` refinement variant for the p-adaptation part, I read through the already existing implementation for h-adaptation to find some inspiration. However, I stumbled over something that might be an inconsistency in the serial implementation.

For the `fixed_number` algorithm, we forward a corresponding threshold to functions `refine` and `coarsen`. However, the idea of determining the thresholds for refinement and coarsening differs. Consider the following case of three cells, in which we want to refine the first and coarsen the last one (hypothetically).
```
refine|     |coarsen
+-----+-----+-----+
|     |     |     |
|  2  |  1  |  0  |  <- criteria
|     |     |     |
+-----+-----+-----+
   ^                 ^
   |                 |
 begin              end
                    begin+size

n_refine_cells = 1
n_coarsen_cells = 1
```
Right now, we use `std::nth_element` to sort a copy of the criteria vector. So far so good. For the refinement threshold, we pick the value of `begin + n_refine_cells` which corresponds to the first cell that will be left untouched. However for coarsening, we pick the value of `begin + size - n_coarsen_cells` which corresponds to the last cell to be coarsened. In my opinion, we pick these thresholds inconsistently.

Since functions `refine` and `coarsen` compare each cell's criterion with a `<=` or `>=` operation respectively, I would suggest to pick the threshold corresponding to the last cell that should be refined or coarsened. Hence my suggestion to use `begin + n_refine_cells - 1` for refinement.

---

Please inform me about your thoughts on this topic. There are a couple of tests whose output need to be adjusted, so please flag this as `WIP` if you like:
```
	1085 - bits/fe_q_constraints.debug (Failed)
	1151 - bits/max_n_cells_01.debug (Failed)
	1152 - bits/max_n_cells_02.debug (Failed)
	1245 - bits/static_condensation.debug (Failed)
	1249 - bits/step-12.debug (Failed)
	1250 - bits/step-13.debug (Failed)
	1264 - bits/step-7.debug (Failed)
	1265 - bits/step-8.debug (Failed)
	1464 - distributed_grids/3d_coarse_grid_02.debug (Failed)
	1614 - dofs/dof_tools_22a.mpirun=5.debug (Timeout)
	2090 - feinterface/step-12.debug (Failed)
	2594 - grid/refine_and_coarsen_fixed_number_02.debug (Failed)
	2649 - hp/crash_17.debug (Failed)
	2857 - hp/step-12.debug (Failed)
	2858 - hp/step-13.debug (Failed)
	2860 - hp/step-27.debug (Failed)
	2869 - hp/step-7.debug (Failed)
	2870 - hp/step-8.debug (Failed)
	3000 - lac/inhomogeneous_constraints.debug (Failed)
	3919 - mpi/multigrid_adaptive.mpirun=1.debug (Failed)
	3920 - mpi/multigrid_adaptive.mpirun=4.debug (Failed)
	4097 - mpi/periodicity_01.mpirun=5.debug (Timeout)
	4098 - mpi/periodicity_01.mpirun=7.debug (Timeout)
	4161 - mpi/refine_and_coarsen_fixed_number_04.mpirun=1.debug (Failed)
	4343 - multigrid/step-16-03.debug (Failed)
	4344 - multigrid/step-16-04.debug (Failed)
	4347 - multigrid/step-16-07.debug (Failed)
	4355 - multigrid/step-16.debug (Failed)
```


bangerth:
I suspect you're right, though the issue is of course unimportant in practice because whether or not we refine/coarsen an individual cell doesn't matter.

Can you point out where we use these inconsistent offsets? I assume this is where we determine which threshold we should use? Do we actually use a particular value plucked from the vector? A better choice might be to choose the mean value of the last value we want to use and the first one we don't want to use, because in that case it no longer matters whether the downstream functions use `<` or `<=`.

source/grid/grid_refinement.cc
```diff
@@ -231,10 +231,10 @@ GridRefinement::refine_and_coarsen_fixed_number(
           else
             {
               std::nth_element(tmp.begin(),
-                               tmp.begin() + refine_cells,
+                               tmp.begin() + refine_cells - 1,
                                tmp.end(),
                                std::greater<double>());
-              refine(tria, criteria, *(tmp.begin() + refine_cells));
+              refine(tria, criteria, *(tmp.begin() + refine_cells - 1));
```
bangerth: Certainly looks reasonable to me. I see now how it's difficult to take the average value of two values: We only do the partial sort, so determining the adjacent value would require another partial sort.

Incidentally, the existing code is going to blow up if `refine_cells=n_cells`. So what you have clearly looks better.

Can you update the failing tests?

marcfehling:
> I suspect you're right, though the issue is of course unimportant in practice because whether or not we refine/coarsen an individual cell doesn't matter.

Agreeing to its unimportance, but this sort of bothered some tests from a follow-up PR. Considering how much time I've spent in adjusting this, it was probably not worth the effort :(

> Can you point out where we use these inconsistent offsets? I assume this is where we determine which threshold we should use? Do we actually use a particular value plucked from the vector? A better choice might be to choose the mean value of the last value we want to use and the first one we don't want to use, because in that case it no longer matters whether the downstream functions use `<` or `<=`.

> Certainly looks reasonable to me. I see now how it's difficult to take the average value of two values: We only do the partial sort, so determining the adjacent value would require another partial sort.

For `GridRefinement::refine_and_coarsen_fixed_number()`, we pick a particular value from the vector, correct. For `GridRefinement::refine_and_coarsen_fixed_fraction()`, the mean value between the last flagged and first non-flagged cell will be determined as a threshold.

> Incidentally, the existing code is going to blow up if `refine_cells=n_cells`. So what you have clearly looks better.

Right above the partial sort, you'll find a check for exactly this case. So this code was 'safe' from the beginning. However, I am still confident that my change has its right to exist :)

> Can you update the failing tests?

I am just about to do so. Just to be safe, I'll run the testsuite once more on my workstation.

marcfehling:
Adjusting these thresholds, I found other inconsistencies:
- For `GridRefinement::refine_and_coarsen_fixed_fraction()`, we currently do not consider all cells for the determination of thresholds. There must have been some confusion with 'past-the-end' iterators.
- We might pick the wrong coarsening threshold for `parallel::distributed::GridRefinement::refine_and_coarsen_fixed_number()` compared to the serial variant. A cleverly placed `std::ceil` made both results consistent, ultimately leading to similar results. This produces the correct output for test `mpi/refine_and_coarsen_fixed_number_04`. So I consider this whole approach a success :)

I'll push these commits as soon as my testsuite completes.

Lots of test output has been altered, so it appears that this might be a profound change to the library. I have respect for that (and maybe also some jitters). But in the end seeing that both `GridRefinement` and `parallel::distributed::GridRefinement` yield the same results feels like a huge success!

bangerth:
I appreciate that you're willing to do things because they're right, even if they're inconvenient. I think we share that trait :-)

tests/bits/fe_q_constraints.output
```diff
@@ -1,16 +1,16 @@
 
 DEAL::---------------------------------------------------------
-DEAL::P-Order: 1  Number of degrees of freedom: 24070
-DEAL::L2-Norm of test function 0.00307079
+DEAL::P-Order: 1  Number of degrees of freedom: 23529
+DEAL::L2-Norm of test function 0.163112
```
bangerth: Do you know what causes such a large change?

bangerth:
I'll merge this because it seems like the right thing to do and because the tests succeed. I'm still curious about whether you have an idea why the difference is so large for this one test.

tests/bits/fe_q_constraints.output
marcfehling: I am not particularly sure about the impact of this PR on this test.

I assume that this is to the nature of the test scenario: We distribute random refinement indicators on the whole mesh. It might happen that with this PR, we now exclude crucial parts of the mesh that have a big impact on the global error. I think this will be the main cause for this, since refinement happens arbitrarily.
https://github.com/dealii/dealii/blob/0ad551ed25348288e5d7917a40c284a14bc74c5c/tests/bits/fe_q_constraints.cc#L142-L149

bangerth: Ah, good point. I should have just looked at the test. Thanks!

PR #9142, opened 2019-12-05
Added the colorize option to subdivided_hyper_cube grid generator

blaisb:
The subdivided_hyper_cube is the one of the only grid generator which
does not allow the boundaries to be colorized. This is a non-standard
behavior, because all of the other generators (hyper_cube,
hyper_rectangle, subdivided_hyper_rectangle) allow one to colorize the
boundaries. This small pull request addresses this by adding the
colorize option to the subdivided_hyper_cube grid. It also fixes the
grid_generator_from_name to make it coherent with the new
implementaiton.

### DIFF
--- a/tests/grid/grid_generator_subdivided_hyper_cube_01.cc
+++ b/tests/grid/grid_generator_subdivided_hyper_cube_01.cc
@@ -1,0 +1,53 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Test for GridGenerator::subdivided_hyper_cube() with colorized boundary
+// conditions
+
+#include <deal.II/base/tensor.h>
+
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/tria.h>
+
+#include "../tests.h"
+
+template <int dim>
+void
+test()
+{
+  deallog << "dim = " << dim << std::endl;
+  Triangulation<dim> tr;
+  GridGenerator::subdivided_hyper_cube(tr, 1, -1, 1, true);
+  for (const auto &cell : tr.active_cell_iterators())
+    {
+      deallog << "cell:" << std::endl;
+
+      for (const auto &face : cell->face_iterators())
+        {
+          if (face->at_boundary())
+            deallog << "boundary id = " << face->boundary_id()
+                    << " center = " << face->center()
+                    << " faceidx = " << face->index() << std::endl;
+        }
+    }
+}
+
+int
+main()
+{
+  initlog();
+  test<2>();
+  test<3>();
+}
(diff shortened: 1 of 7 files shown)


blaisb:
> Looks good to me, but we need a changelog entry and a test. slightly_smiling_face

Sorry this is the dumbest question, but how do I write a changelog entry in this case?
I'll write a simple test that uses the colorization

masterleinad:
Just something like:
> Improved: GridGenerator::subdivided_hyper_cube() gained an option to also set boundary ids. 
> `<br>`
> (Bruno Blais, 2019/12/05)

as  `doc/news/changes/minor/20191205BrunoBlais`.

tests/grid/grid_generator_subdivided_hyper_cube_01.cc
```diff
@@ -0,0 +1,56 @@
...
+template <int dim>
+void
+test()
+{
+  deallog << "dim = " << dim << std::endl;
+  Triangulation<dim> tr;
+  GridGenerator::subdivided_hyper_cube(tr, 1, -1, 1, true);
+  typename Triangulation<dim>::active_cell_iterator cell = tr.begin_active(),
+                                                    endc = tr.end();
+  for (; cell != endc; ++cell)
```
masterleinad: Personally, I would use a range-based for loop here via
```
for (const auto &cell: tr.active_cell_iterators())
```
...

tests/grid/grid_generator_subdivided_hyper_cube_01.cc
```diff
@@ -0,0 +1,56 @@
...
+    {
+      deallog << "cell:" << std::endl;
+
+      for (unsigned int face = 0; face < GeometryInfo<dim>::faces_per_cell;
+           ++face)
```
masterleinad: ... and
```
for (const auto& face: cell->face_iterators())
```
here,

tests/grid/grid_generator_subdivided_hyper_cube_01.cc
```diff
@@ -0,0 +1,56 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2016 - 2018 by the deal.II authors
```
masterleinad: You could also fix the year here.

blaisb: I applied all your comments. Might as well do things correctly the first time :)! Please tell me if you feel anything else should be changed.

tests/grid/grid_generator_subdivided_hyper_cube_01.cc
```diff
@@ -0,0 +1,55 @@
...
+
+template <int dim>
+void
+test()
+{
+  deallog << "dim = " << dim << std::endl;
+  Triangulation<dim> tr;
+  GridGenerator::subdivided_hyper_cube(tr, 1, -1, 1, true);
+  typename Triangulation<dim>::active_cell_iterator cell = tr.begin_active(),
+                                                    endc = tr.end();
```
masterleinad: You don't need these variables anymore.

blaisb: Indeed, was trying to go too fast. Fixed

tests/grid/grid_generator_subdivided_hyper_cube_01.cc
```diff
@@ -0,0 +1,53 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2016 - 2019 by the deal.II authors
```
masterleinad: Oh, the point here was to update the first year this file (or its content) was part of the repository. We update the second year automatically.
```suggestion
// Copyright (C) 2019 by the deal.II authors
```

blaisb:
> You might also want to squash all your commits into one or two.

I squashed everything into a single commit. Not point in having multiple ones here I think.

masterleinad:
Thanks!

blaisb:
> Thanks!

No problem. Thanks for taking all the time to review!
 :)!

PR #9143, opened 2019-12-05
Cleanup function parser

drwells:
Related to #9125: there were a few places in the documentation that needed to be converted to doxygen style comments, some dead code, and a few other small things.

### DIFF
--- a/include/deal.II/base/function_parser.h
+++ b/include/deal.II/base/function_parser.h
@@ -91,8 +91,8 @@ class Vector;
  * @code
  * // set up problem:
  * std::string variables = "x,y";
- * std::string expression = "cos(x)+sqrt(y)";
- * std::map<std::string,double> constants;
+ * std::string expression = "cos(x) + sqrt(y)";
+ * std::map<std::string, double> constants;
  *
  * // FunctionParser with 2 variables and 1 component:
  * FunctionParser<2> fp(1);
@@ -113,7 +113,7 @@ class Vector;
  * The second example is a bit more complex:
  * @code
  * // Define some constants that will be used by the function parser
- * std::map<std::string,double> constants;
+ * std::map<std::string, double> constants;
  * constants["pi"] = numbers::PI;
  *
  * // Define the variables that will be used inside the expressions
@@ -162,7 +162,7 @@ class Vector;
  * at http://muparser.beltoforion.de/ .
  *
  * For a wrapper of the FunctionParser class that supports ParameterHandler,
- * see ParsedFunction.
+ * see Functions::ParsedFunction.
  *
  * Vector-valued functions can either be declared using strings where the
  * function components are separated by semicolons, or using a vector of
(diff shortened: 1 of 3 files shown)


include/deal.II/base/function_parser.h
```diff
@@ -259,10 +256,13 @@ class FunctionParser : public AutoDerivativeFunction<dim>
   FunctionParser(FunctionParser &&) = delete;
 
   /**
-   * Destructor. Explicitly delete the FunctionParser objects (there is one
-   * for each component of the function).
+   * Destructor.
+   *
+   * @note We explicitly declare this destructor here (without the definition)
+   * since we do not expect muParser.h to be available in user projects: i.e.,
+   * the destructor must be defined in the source file.
```
bangerth: It took me twenty seconds to understand what you're trying to say here. I think it's of no importance to the user -- in fact nobody ever has to look up the documentation of a destructor because it takes no arguments and calling it is not a voluntary act.

I would move this comment to the implementation of the destructor to explain why we even have a user-defined destructor and why it's implementation is (presumably) empty.

source/base/function_parser.cc
```diff
@@ -88,9 +79,8 @@ FunctionParser<dim>::FunctionParser(const std::string &expression,
 }
 
 
-// We deliberately delay the definition of the default destructor
-// so that we don't need to include the definition of mu::Parser
-// in the header file.
+
+// See the note in the header for why this is here.
 template <int dim>
 FunctionParser<dim>::~FunctionParser() = default;
```
bangerth: This is where the comment about the destructor should go. Interesting that one can put the `=default` onto a definition -- I hadn't realized this.

drwells:
I fixed it.

source/base/function_parser.cc
```diff
@@ -88,9 +79,11 @@ FunctionParser<dim>::FunctionParser(const std::string &expression,
 }
 
 
-// We deliberately delay the definition of the default destructor
-// so that we don't need to include the definition of mu::Parser
-// in the header file.
+
+// note We explicitly define the constructor here (instead of silently using
```
rezarastak:
```suggestion
// note We explicitly define the destructor here (instead of silently using
```

drwells: Nice catch. Fixed :)

source/base/function_parser.cc
```diff
@@ -292,12 +274,9 @@ FunctionParser<dim>::init_muparser() const
     {
       fp.get().emplace_back(new mu::Parser());
 
-      for (std::map<std::string, double>::const_iterator constant =
-             constants.begin();
-           constant != constants.end();
-           ++constant)
+      for (const std::pair<std::string, double> &constant : constants)
```
masterleinad: `clang-tidy` complains that the correct type for the first template parameter is (`const std::pair<const std::__cxx11::basic_string<char>, double> &`) and suggests `const auto&` instead.

drwells: I think those are the same but lets use `auto &` to make it happy.

drwells:
Fixed!

PR #9144, opened 2019-12-05
Add to FEInterfaceValues what we already have in FEValues.

bangerth:
I'm already using this functionality in step-71, so forewent writing a test.

### DIFF
--- a/include/deal.II/fe/fe_interface_values.h
+++ b/include/deal.II/fe/fe_interface_values.h
@@ -52,6 +52,12 @@ class FEInterfaceValues
 {
 public:
   /**
+   * Number of quadrature points.
+   */
+  const unsigned int n_quadrature_points;
+
+
+  /**
    * Construct the FEInterfaceValues with a single FiniteElement (same on both
    * sides of the facet). The FEFaceValues objects will be initialized with
    * the given @p mapping, @p quadrature, and @p update_flags.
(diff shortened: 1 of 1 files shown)


include/deal.II/fe/fe_interface_values.h
```diff
@@ -162,6 +168,20 @@ class FEInterfaceValues
   bool
   at_boundary() const;
 
+  /**
+   * Mapped quadrature weight. This value equals the
+   * mapped surface element times the weight of the quadrature
+   * point.
+   *
+   * You can think of the quantity returned by this function as the
+   * surface element $dds$ in the integral that we implement here by
```
kronbichler: Why do you write `$dds$`? Did you mean `$ds$` as [here](https://github.com/dealii/dealii/blob/master/include/deal.II/fe/fe_values.h#L2943)?

bangerth: Copy-paste mistake...

bangerth:
Documentation mistake is now fixed!

include/deal.II/fe/fe_interface_values.h
```diff
@@ -51,6 +51,12 @@ template <int dim, int spacedim = dim>
 class FEInterfaceValues
 {
 public:
+  /**
```
tjhei: Do we put the public member variables at the top above the constructor? I would prefer them to be below the functions.

bangerth: I just copied the style from `FEValuesBase` :-)

bangerth:
Ping?

include/deal.II/fe/fe_interface_values.h
```diff
@@ -553,6 +575,17 @@ FEInterfaceValues<dim, spacedim>::reinit(const CellIteratorType &cell,
 
 
 
+template <int dim, int spacedim>
+inline double
```
masterleinad: We don't have `inline` for the other functions here, but it doesn't really matter.

bangerth: That too is just copied from `fe_values.h`. If you want me to, I'll commit to making all of the functions `inline` in a follow-up patch, but for the moment I'm eager to get this patch merged because I have an external project that relies on it.

masterleinad: No, it's fine with me.

PR #9151, opened 2019-12-07
Introduced hp::Refinement::p_adaptivity_fixed_number().

marcfehling:
To compare the impact of different p-adaptation strategies, I was looking for a restriction on all of them that made them adapt a predefined number of cells only. For this, we need a strategy for p-adaptation comparable to `GridRefinement::refine_and_coarsen_fixed_fraction()`. This PR provides such a strategy.

Since most of the functionality is similar to pre-existing code, I wanted to re-use some helper functions from `parallel::distributed::GridRefinement` namespace. The first commit basically moves the corresponding functions from an anonymous namespace to a globally accessible one. Please leave a note if you do not agree to this procedure.

The second commit provides the new functionality.


include/deal.II/hp/refinement.h
```diff
@@ -254,6 +254,49 @@ namespace hp
...
+     * For refinement, the threshold will be associated with the cell that has
+     * the @p p_refine_fraction times Triangulation::n_active_cells() largest
+     * indicator, while it is the cell with the @p p_refine_coarsen times
+     * Triangulation::n_active_cells() lowest indicator for coarsening.
+     *
+     * Each entry of the parameter @p criteria needs to correspond to an active
+     * cell. Parameters @p p_refine_fraction and @p p_coarsen_fraction need to be
+     * in the interval $[0,1]$.
+     *
+     * @note Preceeding calls of Triangulation::prepare_for_coarsening_and_refinement()
```
bangerth: Do you mean "Following calls" or "Later calls"?

Maybe suggest to call that function first. Or does the current function clear refine/coarsen flags on cells where we will use p-adaptivity, and consequently one *has* to call `prepare_c_and_r` again at a later time (maybe through `execute_c_and_r`)?

source/hp/refinement.cc
```diff
@@ -207,6 +228,182 @@ namespace hp
...
+      AssertDimension(dof_handler.get_triangulation().n_active_cells(),
+                      criteria.size());
+      Assert((p_refine_fraction >= 0) && (p_refine_fraction <= 1),
+             dealii::GridRefinement::ExcInvalidParameterValue());
+      Assert((p_coarsen_fraction >= 0) && (p_coarsen_fraction <= 1),
+             dealii::GridRefinement::ExcInvalidParameterValue());
+
+      // 1.) First extract from the vector of indicators the ones that
+      //     correspond to cells that we locally own.
+      unsigned int   n_flags_refinement(0), n_flags_coarsening(0);
```
bangerth: Put these on separate lines and initialize via `=0` instead of `(0)` for common style with many other places

source/hp/refinement.cc
```diff
@@ -207,6 +228,182 @@ namespace hp
...
+          {
+            if (cell->refine_flag_set())
+              indicators_refinement(n_flags_refinement++) =
+                criteria(cell->active_cell_index());
+            else if (cell->coarsen_flag_set())
+              indicators_coarsening(n_flags_coarsening++) =
+                criteria(cell->active_cell_index());
+          }
+      indicators_refinement.grow_or_shrink(n_flags_refinement);
+      indicators_coarsening.grow_or_shrink(n_flags_coarsening);
```
bangerth: ha, I didn't remember we had that function :-)

source/hp/refinement.cc
```diff
@@ -207,6 +228,182 @@ namespace hp
...
+
+      // 2.) Determine the number of cells for p-refinement and p-coarsening on
+      //     basis of the flagged cells.
+      //
+      // 3.) Find thresholds for p-refinment and p-coarsening on only those
+      //     cells flagged for adaptation.
+      //
+      //     For cases in which no or all cells flagged for refinement and/or
+      //     coarsening are subject to p-adaptation, we usually pick thresholds
+      //     that appeal to all or none of the cells at once. However here, we
```
bangerth: appeal -> apply?

source/hp/refinement.cc
```diff
@@ -207,6 +228,182 @@ namespace hp
...
+      //     do not know which threshold would suffice for this task because the
+      //     user could provide any comparison function. Thus if necessary, we
+      //     overwrite the user's choice with suitable functions simplying
+      //     returning 'true' and 'false' for any cell with reference wrappers.
+      //     Thus, no function object copies are stored.
+      //
+      // 4.) Perform p-adaptation with absolute thresholds.
+      Number threshold_refinement(0.), threshold_coarsening(0.);
+      auto   reference_compare_refine  = std::cref(compare_refine);
+      auto   reference_compare_coarsen = std::cref(compare_coarsen);
```
bangerth: It would be easier to read if you just inlined these variable's values wherever they are used.

source/hp/refinement.cc
```diff
@@ -207,6 +228,182 @@ namespace hp
...
+      if (const parallel::distributed::Triangulation<dim, spacedim>
+            *parallel_tria = dynamic_cast<
+              const parallel::distributed::Triangulation<dim, spacedim> *>(
```
bangerth: shouldn't this also apply to the fully distributed triangulation?

source/hp/refinement.cc
```diff
@@ -207,6 +228,182 @@ namespace hp
...
+          const std::pair<Number, Number> global_min_max_coarsening =
+            dealii::internal::parallel::distributed::GridRefinement::
+              compute_global_min_and_max_at_root(indicators_coarsening,
+                                                 mpi_communicator);
+
+          // 3.) Compute thresholds if necessary.
+          if (target_index_refinement == 0)
+            reference_compare_refine = std::cref(compare_false<Number>);
+          else if (target_index_refinement == n_global_flags_refinement)
+            reference_compare_refine = std::cref(compare_true<Number>);
```
bangerth: ah, I see how this can be used!

source/hp/refinement.cc
```diff
@@ -207,6 +228,182 @@ namespace hp
...
+                target_index_coarsening,
+                mpi_communicator);
+        }
+      else
+        {
+          //
+          // serial implementation (and parallel::shared implementation)
+          //
+
+          // 2.) Determine amount of cells scheduled for p-adaptation.
```
bangerth: amount -> number

tests/hp/p_adaptivity_flags.cc
```diff
@@ -181,6 +181,28 @@ test()
     validate(tria, dh);
   }
 
+  deallog << "p-adaptivity fixed number" << std::endl;
+  {
+    Triangulation<dim>  tria;
```
bangerth: Just put it into a separate test. this makes it easier to find bugs.

source/hp/refinement.cc
```diff
@@ -168,8 +188,9 @@ namespace hp
               }
           }
 
-      if (const parallel::TriangulationBase<dim, spacedim> *parallel_tria =
-            dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
+      if (const parallel::distributed::Triangulation<dim, spacedim>
+            *parallel_tria = dynamic_cast<
```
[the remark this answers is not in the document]
marcfehling: I noticed that we do not need a the succeeding communication step for the `parallel::shared::Triangulation` class, since we will use distributed vectors in that case (similar to `step-17`).

However, I totally forgot about the new `parallel::fullydistributed::Triangulation` class. I'll revert back to the general class, and add another check whether it's a `parallel::shared::Triangulation`.

source/hp/refinement.cc
marcfehling: Yes, it's the same situation that led me to the decision above. I will apply the same changes here as well.

tests/hp/p_adaptivity_flags.cc
marcfehling: I see. While there, I will divide this particular test into multiple ones, each checking one particular component.

source/hp/refinement.cc
marcfehling: I'm not sure what you mean by that. I need to place all of these variables in front of the conditional statement to keep them in the correct scope, so I'm not sure if there is any benefit if I move them somewhere else. Or should I be more elaborate in the comments on why we need these reference wrappers here?

include/deal.II/hp/refinement.h
marcfehling: This was meant as a warning that `prepare_c_and_r` will change `refine` and `coarsen` flags which is the decision basis for this function.

So preceeding or earlier calls of `prepare_c_and_r` might change the output of this function. Maybe I should be more specific like "Don't call `prepare_c_and_r` before this function unless you know what you are doing."

marcfehling:
I was unfocused and had some kind of accident during the last rebasing session... I fixed everything now and addressed all suggestions. Waiting until testers finish, but feel free to give it another look!

marcfehling:
Part of #7515.

Issue #9153, opened 2019-12-08
Introduce element-centric loops (ECL)


peterrum:
In collaboration with @kronbichler, I will integrate element-centric loops (ECL) into the matrix-free framework. ECL can be used for `FE_DGQ` and `FE_DGQHermite`; in contrast to the face-centric loop `MatrixFree::loop` (FCL), which processes cells, faces, and boundary faces independently, the new loop will process a cell and all its faces in one go. As a consequence, faces will be processed two times. However, the benefit of ECL is the improved cache locality and its suitability for shared memory 
(see also https://dl.acm.org/citation.cfm?id=3325864).

I will reference this issue by the upcoming PRs.

peterrum:
FYI @nfehn

peterrum:
I think these 5 PRs are enough for first working ECL implementation. They have added:
- the capability to access neighboring cells even if `FEFaceEvaluation::reinit(cell, face)` was celled
- a cell loop with the capability to update the ghost values of a vector and to compress (so that all ghost faces/cells are updated)
- the capability to read/write mapping information and values (even for neighbors)

peterrum:
While reviewing PR #8319, I was wondering in which tutorial we could present ECLs. The tutorial in this PR might be suitable since it uses explicit explicit time stepping and so we could merge simply both the advection and the inverse-mass matrix operator. What do you think @kronbichler? We probably need to extend ECLs for unstructured meshes so that it works for the cylinder, but that is either way planed...

kronbichler:
I do plan to switch this tutorial program to that variant once we are settled with the infrastructure, indeed. But there is no hurry and right now there is independent value in having this version, too.

peterrum:
@kronbichler While working on an improved version of step-67 and working on the simplification of the ECL support in hyper.deal (https://github.com/hyperdeal/hyperdeal/pull/30), I found that some common utility functions would be nice. Let's discuss next where to put these!

drwells:
I believe ECL is now mostly done - if not lets reopen this.

PR #9155, opened 2019-12-08
ECL: make mapping of neighbor available

peterrum:
Only the last commit is relevant.

@kronbichler:
- should we add a new parameter to `MatrixFree::AdditionalData` to request the setup of these data structures?
- how should we treat hanging nodes; we could ignore them for now; however, we should add an assert in that case!
- any suggestions for a suitable test?

Related to issue #9153.

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -7417,10 +7417,11 @@ FEFaceEvaluation<dim,
                             .normal_vectors[offsets];
   this->jacobian = &this->matrix_info->get_mapping_info()
                       .face_data_by_cells[this->quad_no]
-                      .jacobians[0][offsets];
-  this->normal_x_jacobian = &this->matrix_info->get_mapping_info()
-                               .face_data_by_cells[this->quad_no]
-                               .normals_times_jacobians[0][offsets];
+                      .jacobians[!this->is_minus_face][offsets];
+  this->normal_x_jacobian =
+    &this->matrix_info->get_mapping_info()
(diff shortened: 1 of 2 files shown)


kronbichler:
> should we add a new parameter to MatrixFree::AdditionalData to request the setup of these data structures?

I don't think so - the big spike already comes when the users sets `AdditionalData::mapping_update_flags_faces_by_cells`, so populating the remaining 40% or so will show up very little in memory consumption of that part, and even less for the rest.

> how should we treat hanging nodes; we could ignore them for now; however, we should add an assert in that case

Yes, let us ignore them for now and assert in case we are asked to set up this  - you could simply check whether the level of the neighbor is different from the one on the present cell (which would mean coarser). With that check, we are sure to find out whether that is the case.

> any suggestions for a suitable test?

We do not check those data structures for the rest either, so I suggest this waits until we compute some weak forms with it... :smile: 

A more long-term question is what the most appropriate data structure for this would be because we are duplicating the Jacobians and Jacobians times normals that we store from the exterior side - they are already present for some other cell index on the interior side. Ideally, one would keep a common storage of those fields. Particularly interesting is the Jacobian (not the inverse but the original value) because `dim-1` out of `dim` columns are the same with the interior side, coming from the tangential derivatives which are the same on both sides of a face. Not sure if we want to exploit those. In any case, I will need to think about what is best to do here, given that this is a case-by-case decision and sometimes we are memory-constrained, where loading too much hurts, whereas some other times we might not be limited by memory bandwidth so using the present approach would be preferable.

peterrum:
@kronbichler I have rebased!

PR #8319, opened 2019-06-13
step-67: explicit time integration with matrix-free implementations

kronbichler:
This tutorial implements an explicit time integrator with matrix-free implementation for DG elements. A Runge-Kutta scheme is used. A particular feature presented here is the [MatrixFreeOperators::CellwiseInverseMassMatrix class](https://www.dealii.org/developer/doxygen/deal.II/classMatrixFreeOperators_1_1CellwiseInverseMassMatrix.html) and a few other things that are not included in the Poisson-type problems we have for demonstrating matrix-free algorithms right now.

As equations, I implemented the Euler equations with some setup that does not lead to too many shocks in order to avoid implementing shock capturing, which should end up being maybe ~600-1000 lines of code (including comments)~[update: I wrote lots of comments, so we are now at around 1k lines of code only and 2300 lines of code including comments].

~Scalar advection would also be possible but is a bit basic, given that we have several simple Poisson operators for the matrix-free case already. Acoustic wave propagation would have been another possibility but I have another project [ExWave](https://github.com/kronbichler/exwave) already that explains this, so it feels redundant. Vote with :+1: if you agree on the Euler example or :confused: if you want something else.~

PR #9157, opened 2019-12-09
Add field for inverse/projection of 1D polynomials to MF::ShapeInfo

kronbichler:
We currently have a limited support for inverse mass matrix operations in the matrix-free context via the class `MatrixFreeOperators::CellwiseInverseMassMatrix`. That class previously computed the necessary transformation matrix (essentially the inverse of the interpolation matrix from dof values to quadrature points) in the constructor. This concept can be applied more generally also when going from a higher degree polynomial space to a lower degree polynomial space by a projection. Considering the case of a quadrature formula with more points than there are scalar polynomials, this projection is a reference-element operation to reduce the polynomial degree, as opposed to an actual mass matrix with over-integration that does the least squares fit in the transformed coordinates.

Given the generality of the concept, I have decided to pull this information into the internal `MatrixFreeFunctions::ShapeInfo` class to have it once and for all for a given quadrature formula / finite element combination.

The reason for introducing it is that I have the code for #8319 almost ready and want to simplify the inverse mass matrix application there.


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -3078,30 +3081,30 @@ namespace internal
           Number *      out = out_array + d * dofs_per_component;
           // Need to select 'apply' method with hessian slot because values
           // assume symmetries that do not exist in the inverse shapes
-          evaluator.template hessians<0, false, false>(in, out);
+          evaluator.template hessians<0, true, false>(in, out);
```
kronbichler: Note that the old implementation stored the transpose of the variant I provide now; the reason is that for generalization to the non-square (or pseudo inverse) case it is natural to combine the operations with the rows and columns switched.

kronbichler:
The simplification at this point is that we do not need to compute anything in the constructor. I plan a second simplification in terms of access functions without the explicit call to `fill_inverse_JxW_values` for the case when a user simply wants to use the mass matrix without additional scaling, but I would rather include that in a follow-up PR.

peterrum:
> without the explicit call to fill_inverse_JxW_values

That will be great ;)

Issue #8219, opened 2019-05-17
Redo tutorial images


bangerth:
A lot of the pictures shown in the tutorials are quite old and show an aesthetic and color scheme that is a bit outdated. In particular, we have many many pictures that were created with Visit's default "rainbow" color scale. But we know that the rainbow scale is a bad choice (why, or why, does Visit not change its default!) because it lumps half of the color range into the greens, and it is not a good choice for people with color blindness either.

We should redo essentially all of the pictures of the tutorial using better color schemes. For example, in Visit the "oranges", "blues", and "magma" scales are pretty good. I think this might make for a nice project at a hackathon, or if someone wanted to spend a few minutes at the end of a long day on being a bit of an artist.

Rombur:
The people of matplotlib did a lot of work on how to create good colormaps. Here is their [page](https://matplotlib.org/tutorials/colors/colormaps.html) about it and a nice [video](https://www.youtube.com/watch?v=xAoljeRJ3lU) about it

drwells:
See also #6948: we should tackle both at once.

bangerth:
Ah yes, @Rombur, thanks for the link. I had looked for it, but couldn't find it (because I was looking for python, not matplotlib).

marcfehling:
Here is another reference: "Ten Simple Rules for Better Figures"
https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1003833

jpthiele:
Just found this and I wholeheartedly support this. I just went through all current steps and here's the list which still has rainbow/jet plots. Also @bangerth it would be good to know where to "publish" these images so they end up at the right place.
Also which colorbar was used for the already changed ones, so it's consistent?
(Even though personally I'd prefer viridis or cividis)
- [ ] step-4
- [ ] step-7
- [ ] step-12
- [ ] step-16
- [ ] step-17
- [ ] step-18
- [ ] step-20
- [ ] step-21
- [ ] step-23
- [ ] step-24
- [ ] step-25
- [ ] step-29
- [ ] step-30
- [ ] step-31
- [ ] step-33
- [ ] step-34
- [ ] step-36
- [ ] step-38
- [ ] step-40
- [ ] step-41
- [ ] step-43
- [ ] step-51
- [ ] step-53
- [ ] step-61

jpthiele:
Cividis is a color vision deficiency optimized version of viridis as described here:
[...]

PR #8187, opened 2019-05-15
Fix bug in tutorial step-46

starki0815:
This fixes a bug in tutorial step-46, which caused the y-displacements (z in 3d) to be unsymmetric. As the results are affected, the plots in the documentation maybe need adjustment too ... but I don't know how to do this.

PR #9159, opened 2019-12-10
Update the results section of step-46.

bangerth:
This is a follow-up to #8187, updating the 2d figures. I took the opportunity to get rid of another instance of the rainbow color scale (see #8219). Running the 3d computations takes longer and I'll have to do that on a different machine.

@starki0815 -- FYI.

These are the pictures I'm linking to:

step-46 9 2 velocity-2d
step-46 9 2 displacement-2d
step-46 9 2 pressure-2d


drwells:
Thanks a lot for getting rid of rainbow :)

bangerth:
I also updated the 3d section now. The picture I put there is this one:
step-46 9 2 3d

PR #9160, opened 2019-12-10
ECL: add loop_cell_centric

peterrum:
@kronbichler Here you might also have to think about a proper naming. Do you agree that we have a new type of function in addition to `loop` and `cell_loop`? This type of loop is similar to `cell_loop`, just that you have the flags for communications...

Related to issue #9153.

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -1254,6 +1254,101 @@ public:
          DataAccessOnFaces::unspecified) const;
 
   /**
+   * This method runs the loop over all cells (in parallel) similarly as
+   * cell_loop() does. However, this function is intended to be used
+   * for the case if face and boundary integrals should be also
+   * evaluated. In contrast to loop(), the user provides only a single function
+   * that should contain the cell integral over a cell (or batch of cells when
+   * vectorizing) and the face and boundary integrals over all its faces. This
+   * is referred to in the literature as `element-centric loop` or `cell-centric
+   * loop`.
+   *
+   * To be able to evaluate all face integrals (with values or gradients
+   * from the neighboring cells), all ghost values from neighboring cells are
+   * updated. Use
+   * FEFaceEvalution::reinit(cell, face_no) to access quantities on arbitrary
+   * faces of a cell and the respective neighbors.
+   *
+   * @param cell_operation Pointer to member function of `CLASS` with the
+   * signature <tt>cell_operation (const MatrixFree<dim,Number> &, OutVector &,
+   * InVector &, std::pair<unsigned int,unsigned int> &)</tt> where the first
+   * argument passes the data of the calling class and the last argument
+   * defines the range of cells which should be worked on (typically more than
+   * one cell is passed in from the loop in order to reduce overheads).
+   *
+   * @param owning_class The object which provides the `cell_operation`
+   * call. To be compatible with this interface, the class must allow to call
+   * `owning_class->cell_operation(...)`.
+   *
+   * @param dst Destination vector holding the result. If the vector is of
+   * type LinearAlgebra::distributed::Vector (or composite objects thereof
+   * such as LinearAlgebra::distributed::BlockVector), the loop calls
+   * LinearAlgebra::distributed::Vector::compress() at the end of the call
+   * internally.
+   *
+   * @param src Input vector. If the vector is of type
+   * LinearAlgebra::distributed::Vector (or composite objects thereof such as
+   * LinearAlgebra::distributed::BlockVector), the loop calls
+   * LinearAlgebra::distributed::Vector::update_ghost_values() at the start of
+   * the call internally to make sure all necessary data is locally
+   * available. Note, however, that the vector is reset to its original state
+   * at the end of the loop, i.e., if the vector was not ghosted upon entry of
+   * the loop, it will not be ghosted upon finishing the loop.
+   *
+   * @param zero_dst_vector If this flag is set to `true`, the vector `dst`
+   * will be set to zero inside the loop. Use this case in case you perform a
+   * typical `vmult()` operation on a matrix object, as it will typically be
+   * faster than calling `dst = 0;` before the loop separately. This is
+   * because the vector entries are set to zero only on subranges of the
+   * vector, making sure that the vector entries stay in caches as much as
+   * possible.
+   *
+   * @param src_vector_face_access Set the type of access into the vector
+   * `src` that will happen inside the body of the @p cell_operation function
+   * during face integrals.
+   * As explained in the description of the DataAccessOnFaces
+   * struct, the purpose of this selection is to reduce the amount of data
+   * that must be exchanged over the MPI network (or via `memcpy` if within
+   * the shared memory region of a node) to gain performance. Note that there
+   * is no way to communicate this setting with the FEFaceEvaluation class,
+   * therefore this selection must be made at this site in addition to what is
+   * implemented inside the `face_operation` function. As a consequence, there
+   * is also no way to check that the setting passed to this call is
+   * consistent with what is later done by `FEFaceEvaluation`, and it is the
+   * user's responsibility to ensure correctness of data.
+   */
+  template <typename CLASS, typename OutVector, typename InVector>
+  void
+  loop_cell_centric(void (CLASS::*cell_operation)(
+                      const MatrixFree &,
+                      OutVector &,
+                      const InVector &,
+                      const std::pair<unsigned int, unsigned int> &) const,
+                    const CLASS *           owning_class,
+                    OutVector &             dst,
+                    const InVector &        src,
+                    const bool              zero_dst_vector = false,
+                    const DataAccessOnFaces src_vector_face_access =
+                      DataAccessOnFaces::unspecified) const;
+
+  /**
+   * Same as above, but for the class member function which is non-const.
+   */
+  template <typename CLASS, typename OutVector, typename InVector>
+  void
+  loop_cell_centric(void (CLASS::*cell_operation)(
+                      const MatrixFree &,
+                      OutVector &,
+                      const InVector &,
+                      const std::pair<unsigned int, unsigned int> &),
+                    CLASS *                 owning_class,
+                    OutVector &             dst,
+                    const InVector &        src,
+                    const bool              zero_dst_vector = false,
+                    const DataAccessOnFaces src_vector_face_access =
+                      DataAccessOnFaces::unspecified) const;
+
+  /**
    * In the hp adaptive case, a subrange of cells as computed during the cell
    * loop might contain elements of different degrees. Use this function to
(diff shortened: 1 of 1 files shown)


kronbichler:
I think the names make sense. My main question would be whether we need the boolean flag to control the zeroing of the vector. It probably makes sense to have it for consistent interfaces with `cell_loop`; on the other hand, for a typical DG setup it would make more sense to use `FEEvaluation::set_dof_values()` rather than zeroing first and then sum into the vector. On the other hand, if this kind of loop is applied to other discretizations that contain some continuity (I think about Raviart-Thomas, but it could also be things like the biharmonic equation), so I think it make sense to keep the interface general.

peterrum:
I do not have any preferences in this regard. It just would be consistent with the other loop-methods.

kronbichler:
Then we agree on the interface you're suggesting.

peterrum:
@kronbichler  We could start testing this one.

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,38 @@ class MatrixFree : public Subscriptor
        const DataAccessOnFaces src_vector_face_access =
          DataAccessOnFaces::unspecified) const;
 
+  template <typename CLASS, typename OutVector, typename InVector>
```
kronbichler: Even though this functionality is not yet finished, I suggest we still add some basic documentation of the interface before merging.

peterrum: OK. Let me do it in the afternoon.

peterrum: I have copied the documentation from `loop()`, changed the introduction paragraph, and made some changes regarding the parameters (to reflect that there are no face loops).

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
        const DataAccessOnFaces src_vector_face_access =
          DataAccessOnFaces::unspecified) const;
 
+  /**
+   * This method runs the loop over all cells (in parallel) similarly as
+   * cell_loop() does. However, this function is intended to be used
+   * similarly  as loop(): if face and boundary integrals should be also
```
kronbichler:
```suggestion
   * for the case face and boundary integrals should be also
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   * evaluated. In contrast to loop(), the user provides only a single function
+   * that should contain the cell integral over a cell (batch) but also the
```
kronbichler:
```suggestion
   * that should contain the cell integral over a cell (or batch of cells when vectorizing) and the
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   * face and boundary integral over all its faces. This is referred to in the
```
kronbichler:
```suggestion
   * face and boundary integrals over all its faces. This is referred to in the
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   * literature as `element-centric loop` or `cell-centric loop`.
+   *
+   * To be able to evaluate all face integrals (with values or gradients
+   * from the neighboring cells), all ghost values from neighboring cells are
+   * updated. Use
+   * FEFaceEvalution::reinit(cell, face_no) to access quantities on arbitrary
+   * faces of a cell (and its neighboring cell).
```
kronbichler:
```suggestion
   * faces of a cell and the respective neighbors.
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   *
+   * @param cell_operation Pointer to member function of `CLASS` with the
+   * signature <tt>cell_operation (const MatrixFree<dim,Number> &, OutVector &,
+   * InVector &, std::pair<unsigned int,unsigned int> &)</tt> where the first
+   * argument passes the data of the calling class and the last argument
+   * defines the range of cells which should be worked on (typically more than
+   * one cell should be worked on in order to reduce overheads).
```
kronbichler:
```suggestion
   * one cell is passed in from the loop in order to reduce overheads).
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   *
+   * @param owning_class The object which provides the `cell_operation`
+   * call. To be compatible with this interface, the class must allow to call
+   * `owning_class->cell_operation(...)`.
+   *
+   * @param dst Destination vector holding the result. If the vector is of
+   * type LinearAlgebra::distributed::Vector (or composite objects thereof
+   * such as LinearAlgebra::distributed::BlockVector), the loop calls
+   * LinearAlgebra::distributed::Vector::compress() at the end of the call
+   * internally.
```
kronbichler: Do we plan to call `compress()`? To be precise, we can do that if we use the vector partitioner associated to the cells, which in the DG case does not do anything.

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   * vector, making sure that the vector entries stay in caches as much as
+   * possible.
+   *
+   * @param dst_vector_face_access Set the type of access into the vector
+   * `dst` that will happen the face integrals. As explained in the description
+   * of the DataAccessOnFaces
+   * struct, the purpose of this selection is to reduce the amount of data
+   * that must be exchanged over the MPI network (or via `memcpy` if within
+   * the shared memory region of a node) to gain performance. Note that there
+   * is no way to communicate this setting with the FEFaceEvaluation class,
```
kronbichler: The way we intend to use the cell-centric loop, we should only touch the destination vector with an `FEEvaluation` call, whereas `FEFaceEvaluation` will only access the source.

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   * therefore this selection must be made at this site in addition to what is
+   * implemented inside the `face_operation` function. As a consequence, there
+   * is also no way to check that the setting passed to this call is
+   * consistent with what is later done by `FEFaceEvaluation`, and it is the
+   * user's responsibility to ensure correctness of data.
+   *
+   * @param src_vector_face_access Set the type of access into the vector
+   * `src` that will happen inside the body of the @p face_operation function,
```
kronbichler:
```suggestion
   * `src` that will happen inside the body of the @p cell_operation function when working on faces.
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   * in analogy to `dst_vector_face_access`.
```
kronbichler:
```suggestion
```

peterrum:
@kronbichler Thanks for the review! I have made the changes.

Regarding:

> Do we plan to call compress()? To be precise, we can do that if we use the vector partitioner associated to the cells, which in the DG case does not do anything.

and 

> The way we intend to use the cell-centric loop, we should only touch the destination vector with an FEEvaluation call, whereas FEFaceEvaluation will only access the source.

Indeed, in our use case we would use `DataAccessOnFaces:: values_all_faces` for the source vector (resulting in an `update_ghost_values()`) and `DataAccessOnFaces::none` for the destination vector (resulting in that `compress()` is skipped).

If you want we could remove `dst_vector_face_access`. This would make probably the orignal usage more clear: write to the destination vector only once, making palatalization with threads trivial.

On the other hand, as it is now, it is general: but I do not know if there are any use cases...

kronbichler:
We can leave the decision up to the user, but I would make the default argument `DataAccessOnFaces::none` to highlight that this is the general use in analogy to what gets set upon `cell_loop()`. If we have continuous elements and some face integrals attached to them (think of step-47), we would still do communication.

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1253,6 +1253,107 @@ class MatrixFree : public Subscriptor
...
+   * of the DataAccessOnFaces
+   * struct, the purpose of this selection is to reduce the amount of data
+   * that must be exchanged over the MPI network (or via `memcpy` if within
+   * the shared memory region of a node) to gain performance. Note that there
+   * is no way to communicate this setting with the FEFaceEvaluation class,
+   * therefore this selection must be made at this site in addition to what is
+   * implemented inside the `face_operation` function. As a consequence, there
+   * is also no way to check that the setting passed to this call is
+   * consistent with what is later done by `FEFaceEvaluation`, and it is the
+   * user's responsibility to ensure correctness of data.
```
kronbichler: Maybe you now want to slightly rephrase this text to specify why the default is `none`. Also avoid mentioning `face_operation` and `FEFaceEvaluation` as those should not write into the result vector with this type of loop (but rather provide some results to be merged with `FEEvaluation`.

peterrum: >  Also avoid mentioning face_operation and FEFaceEvaluation as those should not write into the result vector with this type of loop 

I don't get your comment. If `FEFaceEvaluation` should not write into the destination vector, this flag does not have any sense, does it?

> (but rather provide some results to be merged with FEEvaluation.

What kind of results do you mean? These results have be at the places where `FEFaceEvaluation` would have written into...

kronbichler: > If FEFaceEvaluation should not write into the destination vector, this flag does not have any sense, does it?

It depends on how we want to define the operation "write into the vector" and whether the face evaluation related to the interior of a cell writes to the vector or a temporary array. But the main topic we want to define is that `FEFaceEvaluation` must not write to more indices in the vector than the respective `FEEvaluation`. But indeed, it seems that the only selection that is possible here is to use the path that the `cell_loop` is going so, so we do not have a choice. If you agree, we remove the parameter, and put some of the documentation to the `src_vector_face_access`.

tamiko:
@kronbichler @peterrum Is this ready to be merged?

peterrum:
@tamiko I need to adjust the documentation slightly. I hope I get to that tomorrow in the afternoon or on the weekend!

include/deal.II/matrix_free/matrix_free.h
peterrum: I agree and have removed the parameter and updated the documentation accordingly.

PR #9074, opened 2019-11-21
Extract index sets from particle handlers

luca-heltai:
Extract an IndexSet with global dimensions equal to
<s>`n_comps*particles.get_next_free_particle_index()`</s>, `get_next_free_particle_index()` containing <s>for each
particle index, a set of `n_comps*` consecutive indices associated to
the particles that are locally owned.</s> the locally owned indices.

<s>
The indices associated to a particle with global index `n` will be the
half open range `[n_comps*n`, n_comps*(n+1))`.</s>

PR #9168, opened 2019-12-13
Contiguous tensor product IndexSet.

luca-heltai:
Followup to #9074


include/deal.II/base/index_set.h
```diff
@@ -352,6 +352,14 @@ class IndexSet
   void
   subtract_set(const IndexSet &other);
 
+  /**
+   * Return a new IndexSet, with global size equal to `n_comps*this->size()`,
+   * containing for every element `n` of this IndexSet, the half open range
+   * `[n*n_comps, (n+1)*n_comps)`.
+   */
```
bangerth:
```suggestion
   *
   * The name results from the perspective that one starts with an IndexSet and takes the tensor
   * product with a "mask" `[1,...,1]` that contains `n_comp` elements; this results in a matrix
   * of size `this->size()` times `n_comps` that has ones in exactly the rows for which the IndexSet
   * contained an index. This matrix is then "unrolled" again by going through each row one by one
   * and reindexing the entries of the matrix in consecutive order. A one in the matrix then corresponds
   * to an entry in the reindexed IndexSet that is returned by this function.
   */
```

luca-heltai:
> OK if you add something like my suggestion to the documentation. I think it would have been interesting to write the function in a way so that it takes another IndexSet or vector<bool> or ComponentMask, etc, as argument. But this will do.

That's simple enough. I'll write it as you suggest. :)

doc/news/changes/minor/20191212LucaHeltai
```diff
@@ -0,0 +1,5 @@
+New: IndexSet::contiguous_tensor_product() creates a new IndexSet with global size equal to
```
tjhei: update this entry please (function name, description)

PR #9171, opened 2019-12-14
Symbolic function.

luca-heltai:
A new `SymbolicFunction` class, that allows evaluation of gradients, laplacians, hessians and time derivatives leaveraging the Symengine library.

I wanted this class for a very long time... @jppelteret, thanks for all your work with the SD framework! It is very powerful!

Now we can use:

```
SymbolicFunction<dim> fun("cos(x)+y**2*tan(t*x); x-y; t**2");
SymbolicFunction<dim> fun_dt = fun.time_derivative();
```
and use both `fun` and `fun_dt` whenever a `Function` is required in deal.II.


include/deal.II/base/symbolic_function.templates.h
```diff
@@ -0,0 +1,227 @@
...
+    return out;
+  }
+
+
+
+  template <int dim, typename RangeNumberType>
+  Differentiation::SD::types::substitution_map
+  SymbolicFunction<dim, RangeNumberType>::substitute(const Point<dim> &p) const
+  {
+    return Differentiation::SD::make_substitution_map(
```
jppelteret: If my suggestion about caching the user-given substitution map is followed through on, then I would suggest wrapping this in a call to `resolve_explicit_dependencies()`.

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+   * auto df_dt = fun.time_derivative();
+   *
+   * auto c = df_dt.value(p, 0); // c = 0.0
+   * auto d = df_dt.value(p, 1); // d = 2.0
+   * @endcode
+   * where a Function with two components is defined using a string containing
+   * their expressions separated by semicolons.
+   *
+   * A more involved example, that explicitly uses
+   * Differentiation::SD::Expression() objects, is given by
```
jppelteret: `Differentiation::SD::Expression()` -> `Differentiation::SD::Expression`

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+   * auto fp = function.value(p);
+   * auto gradfp = function.gradient(p);
+   * auto lapp = function.laplacian(p);
+   *
+   * // Evaluate the time derivative of the function, its gradient, and its
+   * // Laplacian
+   * auto time_derivative = function.time_derivative();
+   * auto dt_fp = time_derivative.value(p);
+   * auto dt_gradfp = time_derivative.gradient(p);
+   * auto dt_lapp = time_derivative.laplacian(p);
```
jppelteret: `dt_lapp` -> `dt_lapfp`?

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+    Tensor<1, dim, Differentiation::SD::Expression> argument;
+
+    /**
+     * The symbolic time.
+     */
+    mutable Differentiation::SD::Expression time;
+  };
+
+  /**
+   * Allow output using <<.
```
jppelteret: Do we ever write this as << rather than saying the bitwise left shift operator?

tests/base/symbolic_function_01.cc
```diff
@@ -0,0 +1,78 @@
...
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Test the SymbolicFunction class
+
+#include <deal.II/base/point.h>
+#include <deal.II/base/symbolic_function.h>
+
+#include <deal.II/lac/vector.h>
```
jppelteret: Is this header needed?

tests/base/symbolic_function_01.with_symengine=true.output
```diff
@@ -0,0 +1,22 @@
...
+DEAL::SymbolicFunction<dim>: x, y, t -> x + y + cos(exp(t*x*y))
+DEAL::p: 1.00000 2.00000, f(p): 3.54030, grad(f)(p): 1.00000 1.00000, laplacian(f)(p): 0.00000, H(f)(p): 0.00000 0.00000 0.00000 0.00000
+DEAL::Time derivative of SymbolicFunction<dim>: x, y, t -> -x*y*exp(t*x*y)*sin(exp(t*x*y))
+DEAL::p: 1.00000 2.00000, f_t(p): -1.68294, grad(f_t)(p): -1.68294 -0.841471, laplacian(f_t)(p): 0.00000, H(f_t)(p): 0.00000 -0.841471 -0.841471 0.00000
+DEAL::==========================================================
+DEAL::dim = 3, Symengine
+DEAL::x: x y z, f: tan(t**2*z) + atan(y/x), df: -y/(x**2*(1 + y**2/x**2)) 1/(x*(1 + y**2/x**2)) t**2*(1 + tan(t**2*z)**2), H:-2*y**3/(x**5*(1 + y**2/x**2)**2) + 2*y/(x**3*(1 + y**2/x**2)) -1/(x**2*(1 + y**2/x**2)) + 2*y**2/(x**4*(1 + y**2/x**2)**2) 0 -1/(x**2*(1 + y**2/x**2)) + 2*y**2/(x**4*(1 + y**2/x**2)**2) -2*y/(x**3*(1 + y**2/x**2)**2) 0 0 0 2*t**4*tan(t**2*z)*(1 + tan(t**2*z)**2)
```
jppelteret: Are you sure that you no longer want to calculate this Hessian by hand? :-D

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+namespace Functions
+{
+  template <int dim, typename RangeNumberType>
+  class SymbolicFunction;
+}
+#endif
+
+namespace Functions
+{
+#ifdef DEAL_II_WITH_SYMENGINE
```
bangerth: Can you update the file `doc/doxygen/options.dox.in` to also always `#define` this variable so that the documentation shows up?

To make sure people get reasonable error messages when using this class but not having SymEngine installed, it's probably useful to have
```
  #else

  template <int dim>
  class SymbolicFunction {
    static_assert (false, "You can't use this class without having SymEngine installed and detected during "
                                    "configuration of deal.II");
  };
```

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+  /**
+   * A Function class that leaverages symbolic differentiation to compute
```
bangerth:
```suggestion
   * A Function class that leverages symbolic differentiation to compute
```

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+   * gradients, laplacians, hessians, and time derivatives.
```
bangerth: below you spell it "Hessians". See what we use throughout the library. I think I've tried to use upper-case "Hessian" in the documentation where possible, but I'm not sure we're consistent. I prefer the upper-case variant, and the same for "Laplacians".

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+   * be thrown. Additional symbols can be partially evaluated or substituted
+   * by calling apply_substitution_map().
+   *
+   * The simplest use case of this class is given in the following example:
+   * @code
+   * SymbolicFunction<2> fun("x^2+y; t*x*y");
+   * fun.set_time(3.0);
+   * Point<2> p(1.0, 2.0);
+   *
+   * auto a = fun.value(p, 0); // a = 3.0
```
bangerth: Maybe as a reminder for slow people like me:
```suggestion
   * auto a = fun.value(p, /*component=*/0); // a = 3.0
```

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+   * auto b = fun.value(p, 1); // b = 6.0
```
bangerth: same here and below (or not, your choice)

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,315 @@
...
+   * Differentiation::SD::Expression() objects, is given by
+   * @code
+   * // Create a Tensor<1,2,Differentiation::SD::Expression>
+   * // with symbols "x" and "y", and the symbol "t"
+   * const auto x = SymbolicFunction<2>::default_argument();
+   * const auto t = Differentiation::SD::make_symbol("t");
+   *
+   * // Use directly x[0] (the symbol "x"), x[1] (the symbol "y"), and t
+   * // (the symbol "t").
+   * auto f = std::sin(x[0])*std::cos(x[1])*std::sin(t);
```
bangerth: It's really just a `double`. Let's make that clear to prevent people from trying to think too hard:
```suggestion
   * const double f = std::sin(x[0])*std::cos(x[1])*std::sin(t);
```

bangerth: Or did you mean to enclose this in `"..."`?

include/deal.II/base/symbolic_function.h
luca-heltai: > Can you update the file doc/doxygen/options.dox.in to also always #define this variable so that the documentation shows up?

It is already there.... Is the documentation not showing up?

luca-heltai: As far as the static assert is concerned, we don't do this for any of our other classes, which are instead simply forward declared (as in this case, at the beginning of the header).

Moreover, this would break the include tests, which try to include each .h file and compile with it...

bangerth: Ah, I hadn't checked `options.dox.in` :-(

We have quite a lot of classes that have a stub for the case where a dependency is not met. `p::d::Triangulation` is an example. But I don't see why the `static_assert` breaks the test: you can compile a file, you just can't instantiate the class template. Or do I miss something?

include/deal.II/base/symbolic_function.templates.h
luca-heltai: That should not be necessary. The way I implemented your suggestion, is a hybrid version:

I keep a user substitution map around, which can be updated by the user. Whenever that happens, a mutable function (different from the one specified by the user), is generated, together with all its derivatives. If the user updates the map, then a new function is generated using the original function (which is left untouched from construction time), and the new substitution map.

This leaves the user with a consistent function, where you can replace parameters, and that it is only updated when necessary (i.e., I do not need to resolve all explicit dependencies at evaluation time, but just once, when the user supplies a substitution map, or at construction time).

tests/base/symbolic_function_01.with_symengine=true.output
luca-heltai: :D

include/deal.II/base/symbolic_function.h
luca-heltai: MWE:

```
class MyClass {
static_assert(false, "Error message");
}

int main () {
return 0;
}
```
does not compile for me:
```
../snippet.cc:2:1: error: static_assert failed "Error message"
static_assert(false, "Error message");
^             ~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
```

Same with putting the assert in the constructor.

luca-heltai: The file, with your suggestion, cannot be included.

include/deal.II/base/symbolic_function.h
luca-heltai: How do you put a comment like `/* ... */` in the doxygen body of a code snippet? Mumble mumble...

include/deal.II/base/symbolic_function.h
luca-heltai: It is *not* just a double. `x[0]` is of type `SD::Expression`, and so is `x[1]` and `t`. The result is a symbolic expression.

luca-heltai: Or, to be more precise, the result is a `Differentiation::SD::Expression` object.

luca-heltai:
I think I have addressed most of your comments. Ready for second round.

include/deal.II/base/symbolic_function.h
bangerth: Hm :-] Then maybe make it a regular assertion, or make the assertion template-argument dependent. Or just forget about it -- it's not that important.

include/deal.II/base/symbolic_function.h
bangerth: Oh, I see. Can you make that clear in the comment above (unless it already is and I just didn't read carefully enough).

include/deal.II/base/symbolic_function.h
bangerth: I think we've got places where we use `/ *` with a space in between. Visually, it's going to be nearly indistinguishable.

bangerth:
Will take a look tomorrow.

Shouldn't you be in bed? ;-)

luca-heltai:
@jppelteret, is there a better way to create a substitution map?

This works:
```
sub = make_substitution_map(std::make_pair(make_symbol("alpha"), 1.0));
```
and I thought it should have been the same as this:
```
sub = make_substitution_map("alpha", "1.0");
```
but the second substitution map always leads to `libc++abi.dylib: terminating with uncaught exception of type SymEngine::SymEngineException: Symbol cannot be evaluated.`, even though the expression seems ok when printed (and it looks identical to the one generated with the other map...)

luca-heltai:
>Will take a look tomorrow.

>Shouldn't you be in bed? ;-)

Yes. :D I couldn't sleep... and these 40 minutes of programming were refreshing! :)

include/deal.II/base/symbolic_function.h
luca-heltai: Done.

include/deal.II/base/symbolic_function.h
luca-heltai: Done.

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,377 @@
...
+     * function. This object does not include any user-defined arguments.
+     */
+    const Tensor<1, dim, Differentiation::SD::Expression> &
+    get_symbolic_position() const;
+
+    /**
+     * Get the actual symbolic time in use in this symbolic function.
+     */
+    const Differentiation::SD::Expression &
+    get_symbolic_time() const;
```
bangerth: maybe `get_time_symbol()`?

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,377 @@
...
+    /**
+     * Get the currently stored @p user_substitution_map.
+     */
+    const Differentiation::SD::types::substitution_map &
+    get_user_substitution_map() const;
+
+    /**
+     * Return a SymbolicFunction object that represents the time derivative of
+     * this function. The spatial argument, the symbolic time, and the currently
+     * stored user substitution map are forwarded to the new function.
```
bangerth: nice!

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,377 @@
...
+    time_derivative() const;
+
+    /**
+     * Return the value of the function at the given point. Unless there is only
+     * one component (i.e., the function is scalar), you should state the
+     * component you want to have evaluated; it defaults to zero, i.e., the
+     * first component.
+     */
+    virtual RangeNumberType
+    value(const Point<dim> &p, const unsigned int component = 0) const override;
```
bangerth: For this and the following few functions, you might as well omit the documentation altogether and just have
```
// documentation inherited from the base class
```

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,377 @@
...
+    StreamType &
+    print(StreamType &out) const;
+
+  private:
+    /**
+     * Return a substitution map that replaces the argument with the values of
+     * @p point, and the symbolic time with the value of this->get_time().
+     */
+    Differentiation::SD::types::substitution_map
+    substitute(const Point<dim> &point) const;
```
bangerth: if I understand correctly, then this function creates a substitution map that replaces coordinates by the values of `point`? If so, maybe call this function `create_substitution_map_for_point()`? That's because if the function is `const`, it can't actually *substitute* anything.

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,377 @@
...
+  class SymbolicFunction
+  {
+  public:
+    SymbolicFunction()
+    {
+      AssertThrow(
+        false,
+        ExcMessage(
+          "This class is not available if you did not enable SymEngine "
+          "when compiling deal.II."));
```
bangerth: ah, thanks!

luca-heltai:
> Separately: We often pass `Function` objects to other functions that work in parallel. I think you ought to give the class a mutex that is locked in the `value`, `gradient` etc functions.

Is this really necessary?

Each of the `value`, `gradient`, etc.  work in the following way:

```
 return Differentiation::SD::substitute_and_evaluate<RangeNumberType>(
      function[component], create_substitution_map_for_point(p));
```
i.e., `function` (`gradient`, `hessian`, etc.) is not touched, and a temporary is created with `create_substitution_map_for_point(p)`. My understanding is that whenever a thread calls this function, it will create a new temporary, and work on that. No mutex should be necessary.

jppelteret:
> @jppelteret, is there a better way to create a substitution map?
> 
> This works:
> 
> ```
> sub = make_substitution_map(std::make_pair(make_symbol("alpha"), 1.0));
> ```
> 
> and I thought it should have been the same as this:
> 
> ```
> sub = make_substitution_map("alpha", "1.0");
> ```
> 
> but the second substitution map always leads to `libc++abi.dylib: terminating with uncaught exception of type SymEngine::SymEngineException: Symbol cannot be evaluated.`, even though the expression seems ok when printed (and it looks identical to the one generated with the other map...)

This *should* work. It should call [this variant](https://dealii.org/developer/doxygen/deal.II/namespaceDifferentiation_1_1SD.html#a11c0ecc04f2bc13fbe1922457b04c7fe) of `make_substitution_map()`. Where I think that things might be going wrong is that [the constructor for `SD::Expression`](https://dealii.org/developer/doxygen/deal.II/classDifferentiation_1_1SD_1_1Expression.html#abd213bcc0e25ec24dfdf5631985ef3c6) that takes a `char*` might be being called preferentially to that taking the string. Could you try to see what happens when you remove that constructor, or be ridiculously explicit with 
`sub = make_substitution_map(Expression(std::string("alpha")), "1.0");`? I see that the goal is to invoke the [other constructor](https://dealii.org/developer/doxygen/deal.II/classDifferentiation_1_1SD_1_1Expression.html#a9b7f46103f32b947cfb74ebb61ef0cfb) that takes a symbolic expression that is, without the second argument specified, interpreted as a symbol.

include/deal.II/base/function_parser.h
```diff
@@ -211,6 +211,11 @@ class Vector;
  *                        constants);
  * @endcode
  *
+ * @note The difference between this class and the SymbolicFunction class is
+ * that the SymbolicFunction class allows to compute first and second order
+ * derivatives (in a symbolic way), while this class computes first order
+ * derivatives only, using finite difference. For complicated expressions,
```
jppelteret: finite difference -> finite differences

include/deal.II/base/function_parser.h
```diff
@@ -211,6 +211,11 @@ class Vector;
...
+ * this class is generally faster than SymbolicFunction.
```
jppelteret: Out of curiosity, do you notice a significant performance difference between the two for your use cases?

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,389 @@
...
+     * equal to two, and "x", "y", "z" for `dim` equal to three.
+     *
+     * @param time_symbol A symbolic variable representing time. It defaults
+     * to a symbolic variable named "t".
+     *
+     * @param user_substitution_map Any other symbol that may be contained in
+     * the symbolic function needs to be specified in this map. The map may be
+     * empty, and the functions may still contain unevaluated symbols, provided
+     * that you call update_user_substitution_map() and replace all symbols
+     * except
```
jppelteret: Nitpick: Unnecessary line break

include/deal.II/base/symbolic_function.templates.h
```diff
@@ -0,0 +1,243 @@
...
+    , time_symbol(Differentiation::SD::make_symbol("t"))
+  {
+    update_derivatives();
+  }
+
+
+
+  template <int dim, typename RangeNumberType>
+  void
+  SymbolicFunction<dim, RangeNumberType>::update_derivatives()
```
jppelteret: Thanks for splitting this up into a separate function.

include/deal.II/base/symbolic_function.templates.h
```diff
@@ -0,0 +1,243 @@
...
+  {
+    function.resize(user_function.size());
+    function_gradient.resize(user_function.size());
+    function_hessian.resize(user_function.size());
+    function_laplacian.resize(user_function.size());
+
+    for (unsigned int i = 0; i < user_function.size(); ++i)
+      {
+        function[i] = user_function[i].substitute(user_substitution_map);
```
jppelteret: I really don't think that this is necessary. I don't see why the `user_substitution_map` cannot be added to `create_substitution_map_for_point()`. If the user symbols evaluate to constants, or some polynomial function of another user symbol, then we're just doing all of the costly work of differentiation for no reason. If the user symbols depend on the time or evaluation coordinates, then we have a mechanism for them to state that dependency by declaring the symbol a symbolic function. We don't do any partial differentiation here, so there's certainly not an intermediate step that requires substitution before computing a total derivative.

Could you maybe describe a use case or give a short piece of pseudo-code so that I can better understand why your approach would be necessary/preferable to performing the differentiation only once per instance that a `SymbolicFunction` is created?

include/deal.II/base/symbolic_function.templates.h
jppelteret: Yeah, you're right. I'm overly defensive about this. Let's leave the responsibility of making this call to the user.

tests/base/symbolic_function_03.with_symengine=true.output
```diff
@@ -0,0 +1,11 @@
+
+DEAL::x, y, t -> x; 1.0*y; 1.0*t
```
jppelteret: In these cases, this substitution (and subsequent derivative computations) is totally unnecessary. There's nothing wrong (IMO) with retaining everything in symbolic form and having this always return `x, y, t -> x; alpha*y; alpha*t`. All of the substitutions can be done right at the end when the evaluation is done. This is optimal, and you'll see a significant performance benefit when not having to continuously recompute the derivatives on large expressions, and even larger performance benefits when invoking the (to be added) batch optimiser and not doing the evaluation through the symbolic code paths at all.

include/deal.II/base/function_parser.h
luca-heltai: Not really. They are substantially identical, but I have not tested complicated stuff.

include/deal.II/base/symbolic_function.templates.h
luca-heltai: My reasoning is the following: I'd like function/gradient/hessian evaluation to be as quick as possible.  For that evaluation to be quick, I don't want to have to compute derivatives *at every evaluation point*. Think of this case:

~~~
SymbolicFunction<2> my_function("r^3+g(theta)");
auto my_map = make_substitution_map("r", Expression("sqrt(x^2+y^2)", true));
my_function.update_user_substitution_map(my_map);
...
my_function.gradient(p);
~~~
with your approach, I'd have to compute the gradient on the fly, before substituting the value of the point p. We could compute the gradients and hessians only if required (this I can fix), but I still believe it is better *not* to compute them at every point evaluation. With your suggestion, we'd get a gradient computation for every evaluation...

tests/base/symbolic_function_03.with_symengine=true.output
luca-heltai: If you are worried about the output, I can return the `user_function` instead of the `function`, and it will always return `x, y, t -> x; alpha*y; alpha*t`. In the current implementation`function` (the one used for the evaluation) is substituted with the user map only once, instead of doing the substitution at every point evaluation. 

Why do you want to repeat the user substitution again and again and again (for every quadrature point in a three dimensional domain in parallel computations? Even just a few microseconds more will add up...). 

If you don't want me to perform unnecessary derivatives, then that can be easily changed. But I still think that substitution should happen only once for the user map...

include/deal.II/base/symbolic_function.templates.h
jppelteret: Ok, fair point. But I think that we can easily compromise on this. We have two member functions, something like `set_user_substitution_map()` that does what I want to do, and `perform_permanent_substitution()` that does what you want to do.  I'll explain why in the reply to https://github.com/dealii/dealii/pull/9171#discussion_r368151217.

jppelteret: As an aside, we could make a follow up issue / PR to add some functions that convert between Cartesian/cylindrical/spherical coordinates to the `GridTools` or `Physics` namespace, and make some member function that does this transformation for one (assuming some preset variables for the radius, angle etc. that can also be returned in the same way that `get_default_coordinate_symbols()` returns `x,y,z`).

jppelteret: As an aside, we could make a follow up issue / PR to add some functions that convert between Cartesian/cylindrical/spherical coordinates to the `GridTools` or `Physics` namespace, and make some member function that does this transformation for one (assuming some preset variables for the radius, angle etc. that can also be returned in the same way that `get_default_coordinate_symbols()` returns `x,y,z`).

luca-heltai: I think there is a misunderstanding on how things work now.

There are two objects. One is user_function, immutable, inserted at construction time, that contains all the symbols one wants. The other is function, which is the one that is actually evaluated. Since the only way to update the user_substitution_map is through the update_user_substitution_map, this function does not touch the user_function, but only the one that is used to evaluate the function, thus allowing you to perform other substitutions later on (by calling again update_user_substitution_map).
This puts together the advantages of both worlds. It does not change what the user inserted permanently, but only up to the next update_user_substitution_map. How this happen internally is now efficient, while with your suggestion it would be performed at every point evaluation.
Wasn't this what you suggested?

luca-heltai:
I have changed the interface a bit, to come towards your comments, @jppelteret. In particular, now we have

1. The user creates the function as 
~~~
SymbolicFunction<2> my_function("alpha**2*x+y*t);
// any evaluation now will fail
auto subs = make_substitution_map("alpha", 1.0);
my_function.update_user_substitution_map(subs);
// no computations are done yet

auto v = my_function.value(p1); // This triggers a substitution with subs
auto w = my_function.value(p2); // No more substitution with alpha here. We already made it internally, and nobody called update_user_substitution_map() another time.

auto subs = make_substitution_map("alpha", 2.0);
my_function.update_user_substitution_map(subs);
auto v = my_function.value(p1); // This triggers again a substitution with the new subs
auto w = my_function.value(p2); // Reuse the substitution we triggered before

auto T = my_function.gradient(p1); // This triggers computation of gradients (with alpha=2)
auto H = my_function.hessian(p1); // This triggers computation of Hessians (with alpha=2)
~~~
I think this is the optimal solution. No computation is performed unless needed. No substitution is performed if not necessary.

luca-heltai:
Oh, and the output contains the original user function (before substitution).

tests/base/symbolic_function_03.with_symengine=true.output
jppelteret: So, what I want to do as a follow up is to leverage the `BatchOptimizer` to make this evaluate quicker. For my tests with a function `psi = det(T)^2 + (V.V)^2 + s^2.2` parameterised in terms of a scalar, vector and tensor field, transforming this functions into a JIT compiled variant can yield between 2 and 3 orders of magnitude speed-up in 3d when evaluating all terms (function value, first and second derivatives) at one evaluation point 250000 times. In the end, the JIT compiled variant was only 3x slower than native code. As the expression get bigger and bigger (think multi-physics constitutive modelling), that overhead diminished even more.

The main compromise that is to be made is that invoking the optimiser can be time consuming, especially if the functions that it evaluates (which will be all of the functions and derivative, in this case) are lengthy. We would need to invoke the optimiser each time the function definition changes, or the derivatives are recomputed. So, in that case, performing substitutions like `alpha=1.0` during evaluation is so much less time consuming than it may seem. (In the batch optimised version of the function, these substituted variables get transformed to arguments to a tailor-made function and, in native SymEngine code, become disassociated from the substitution map.)

This unwanted overhead will compound itself when you have some function where the coefficients change with the evaluation point. This could be something simple like a boundary condition function with different coefficients on each boundary (but the same generic function). But position->value relationships need not be encoded into the symbolic expression itself, but could in theory be a random number (stupid example, a randomly perturbed coefficient) that's selected at each evaluation point. Each time you'd have to create a whole new `SymbolicExpression` if this coefficient differs from the last one created, which is just going to be very inefficient. In fact, in this case you would certainly like to retain the dictionary substitution that you use here, instead of using any form of symbolic optimiser. However, if one keeps the function as generalised as possible, then this is quite easily tractable. Moreover, we can serialise these expression (and the optimiser itself) and use the save/load capability to prevent having to recompute the derivatives in later simulations! There you win even more by keeping things generic. (I truly believe that the more one tries to exploit the generic nature of symbolic computations, the more there is to gain.) 

There is definitely a cross-over point where just using normal dictionary substitution without even considering a batch optimisation process is quicker that optimising, but its hard to say where it is. But, with this this in mind

> Why do you want to repeat the user substitution again and again and again (for every quadrature point in a three dimensional domain in parallel computations?

I'd say that more often than not you'd want to exploit some form of optimiser.

So, in summary, in order to best future-proof this class, I think that we should give it some functionality to add, at every point, the option to substitute a set of user-defined symbols or coefficients. This would allow us to use this BatchOptimizer thingy in the best way possible, and will really make this thing fly. I know that this is not a feature that's currently in deal.II, but my main objective is to make sure that whatever is developed here is compatible with that class. Your point of wanting to execute a permanent substitution is also equally valid and entirely relevant, so I propose that we meet in the middle and provide both options.

jppelteret: > If you are worried about the output, I can return the user_function instead of the function, and it will always return x, y, t -> x; alpha*y; alpha*t. In the current implementation function (the one used for the evaluation) is substituted with the user map only once, instead of doing the substitution at every point evaluation.

No, I think that what we print should always reflect the actual state of the `function` member variable.

> But I still think that substitution should happen only once for the user map...

The user map can (or should be able to) change from point-to-point. That is, at least, how I would want to exploit this.

luca-heltai: > The user map can (or should be able to) change from point-to-point. That is, at least, how I would want to exploit this.

What I'm not understanding is the following: is applying two substitution map one after the other  so much more costly w.r.t. to applying it all at once? If the cost is comparable, then you can already use this in the way you describe it.

Since we don't have yet a batch optimizer, do you really think it would make a difference? If so, we could implement both, but I fear we are a bit overthinking it.

luca-heltai:
Ping @jppelteret ?

Is this what you had in mind?

include/deal.II/base/symbolic_function.h
```diff
@@ -0,0 +1,470 @@
...
+  template <int dim, typename RangeNumberType>
+  template <typename StreamType>
+  StreamType &
+  SymbolicFunction<dim, RangeNumberType>::print(StreamType &out) const
+  {
+    for (unsigned int i = 0; i < dim; ++i)
+      out << coordinate_symbols[i] << ", ";
+    out << time_symbol << " -> " << user_function[0];
+    for (unsigned int i = 1; i < user_function.size(); ++i)
+      out << "; " << user_function[i];
```
jppelteret: I guess that, in order to match the description of the `print` function, we need to revert this to printing the `function` itself. Alternatively, we could print the keys in the `user_substitution_map` and `additional_function_arguments` (if not empty), which I think might actually be nicer (add to the list of coord symbols and time symbol before the `->`).

luca-heltai: I've adopted a strategy which is mixed. Arguments are treated in the same way of the coordinates and of time, while the user_subsitution_map is appended after a `#` symbol, providing the full picture of the function.

luca-heltai:
By the way, the testers do not have symengine installed. No need to wait for CI.

include/deal.II/base/symbolic_function.h
luca-heltai: Let me know what you think of the output of test n. 4

jppelteret: Looks great. Thank you :-)

jppelteret:
> By the way, the testers do not have symengine installed. No need to wait for CI.

I'll let you do the honours :-)

PR #9173, opened 2019-12-15
Fix particles/extract_index_set_01

masterleinad:
Fixes https://cdash.43-1.org/testDetails.php?test=24482254&build=3292.


tjhei:
Thanks, @masterleinad !

PR #9175, opened 2019-12-16
Improve GridGenerator::channel_with_cylinder

kronbichler:
While working on #8319, I noticed that the grid generated by `GridGenerator::channel_with_cylinder` is a bit pressed together in the transfinite interpolation part, i.e., where we switch from the cylinder mesh to the outer square mesh. I moved the four vertices on the sides of the inserted rectangle a bit, moving some of the deformation into the outer squares. The result can be seen in the three attached files, in particular channel_with_cylinder_2d.png, and compare that with the old version here: https://www.dealii.org/developer/doxygen/deal.II/namespaceGridGenerator.html#a9d0b8a34fe5ca58425aa1fcf0d6132f1


source/grid/grid_generator.cc
```diff
@@ -3625,6 +3625,25 @@ namespace GridGenerator
...
+    // Move the vertices in the middle of the faces of cylinder_tria slightly
+    // to give a better mesh quality. We choose to move them such that the
+    // distance between (0.05 + shell_region_width) and 0.1 is at least half
+    // as long as the remaining length to the boundary.
+    const double shifted_point =
+      std::min(0.05 + shell_region_width + 0.1 * std::sqrt(2) - 0.05 -
+                 shell_region_width,
+               0.1 * 4. / 3.);
+    for (const auto &cell : tria.active_cell_iterators())
+      for (unsigned int v = 0; v < GeometryInfo<2>::vertices_per_cell; ++v)
```
drwells: It would be slightly simpler to just loop over the vertices directly, I think, e.g.,
```cpp
for (Point<2> &v : tria.get_vertices())
  if (v.distance()...)
```

drwells: Nevermind: that is always a const reference.

source/grid/grid_generator.cc
```diff
@@ -3625,6 +3625,25 @@ namespace GridGenerator
...
+        if (cell->vertex(v).distance(Point<2>(0.1, 0.205)) < 1e-10)
+          cell->vertex(v) = Point<2>(0.2 - shifted_point, 0.205);
+        else if (cell->vertex(v).distance(Point<2>(0.3, 0.205)) < 1e-10)
+          cell->vertex(v) = Point<2>(0.2 + shifted_point, 0.205);
+        else if (cell->vertex(v).distance(Point<2>(0.2, 0.1025)) < 1e-10)
+          cell->vertex(v) = Point<2>(0.2, 0.2 - shifted_point);
+        else if (cell->vertex(v).distance(Point<2>(0.2, 0.3075)) < 1e-10)
+          cell->vertex(v) = Point<2>(0.2, 0.2 + shifted_point);
```
drwells: Thanks for putting these in lexical order :+1:

source/grid/grid_generator.cc
```diff
@@ -3625,6 +3625,25 @@ namespace GridGenerator
...
       tria_without_cylinder, cylinder_tria, tria, vertex_tolerance, true);
 
+    // Move the vertices in the middle of the faces of cylinder_tria slightly
+    // to give a better mesh quality. We choose to move them such that the
+    // distance between (0.05 + shell_region_width) and 0.1 is at least half
+    // as long as the remaining length to the boundary.
+    const double shifted_point =
+      std::min(0.05 + shell_region_width + 0.1 * std::sqrt(2) - 0.05 -
+                 shell_region_width,
+               0.1 * 4. / 3.);
```
drwells: I don't understand the point of this calculation. The first expression is just `0.1*sqrt(2)` so its just a comparison of two constants.

As far as I can tell all that this (and the code below) does is (since `0.1*4.0/3.0 < 0.1*sqrt(2)`) nudge the vertices out by `0.1/3.0` in the x direction and `0.1/3.0 - 0.0025` (or `0.1/3.0 - 0.0075`) in the y direction. Is this right? Did you mean for it to rely on the shell region width?

source/grid/grid_generator.cc
```diff
@@ -3625,6 +3625,25 @@ namespace GridGenerator
     GridGenerator::merge_triangulations(
       tria_without_cylinder, cylinder_tria, tria, vertex_tolerance, true);
 
+    // Move the vertices in the middle of the faces of cylinder_tria slightly
+    // to give a better mesh quality. We choose to move them such that the
+    // distance between (0.05 + shell_region_width) and 0.1 is at least half
+    // as long as the remaining length to the boundary.
+    const double shifted_point =
+      std::min(0.05 + shell_region_width + 0.1 * std::sqrt(2) - 0.05 -
```
bangerth: Dangerous...
```suggestion
      std::min(0.05 + shell_region_width + 0.1 * std::sqrt(2.) - 0.05 -
```

source/grid/grid_generator.cc
```diff
@@ -3625,6 +3625,25 @@ namespace GridGenerator
     GridGenerator::merge_triangulations(
       tria_without_cylinder, cylinder_tria, tria, vertex_tolerance, true);
 
+    // Move the vertices in the middle of the faces of cylinder_tria slightly
+    // to give a better mesh quality. We choose to move them such that the
+    // distance between (0.05 + shell_region_width) and 0.1 is at least half
+    // as long as the remaining length to the boundary.
+    const double shifted_point =
```
bangerth: This is an awkward naming of the variable since it's not actually a point, and it's not the shifted location either but really just the "shift" if I see correctly how you're using it below.

source/grid/grid_generator.cc
kronbichler: You are absolutely right - I missed the forest for the trees. The formula comes from how I constructed a better shape in radial distance by some angles I had drawn on paper. In the end, this turned out to just be a constant distance from the center in radial direction (`0.1 * sqrt(2)`). And you are right, once one realizes that things turn out constant, the `0.1 * 4./3.` is the smaller number of the two (making sure the outer elements are not too distorted), so we can simply put that number. I will add this describing text next to the place in the code so we remember how we arrived here.

kronbichler:
I updated the code, updated the output of the two affected tests, and added a changelog. Ready from my point of view.

kronbichler:
I realize that one of my own tests that I added recently was brittle - I updated it now.

PR #9176, opened 2019-12-16
Remove unnecessary assignment

peterrum:
ping @tjhei


drwells:
Its possible that we meant to use that and didn't. Lets not merge until an expert chimes in.

/rebuild

peterrum:
@drwells When looking back in history, I see it used to be used; but not any more, since code was strongly rewritten.

tjhei:
Yeah, this can be removed I would think.

bangerth:
4 approvals for a triviality :-)

Issue #9015, opened 2019-11-07
Python bindings for pre-/post-processing


agrayver:
@dealii

I have decided to turn this into a meta-issue that will track my efforts towards a better python support in deal.II. 

- [x] Add support for faces and boundary ids (fixed by #9023).
- [x] Enable indentation for python bindings (fixed by #9028).
- [x] Add interface for mapping objects (see #9055). 
- [x] Enable working with manifolds (see #9109 #9179)
- [x] Extend GridTools support (see #9148 #9178 #9429 #9510)
- [x] Add bindings for Quadrature (see #9428)
- [x] Implement tutorial programs (candidates: step-49, step-53) in python (see #9526 #9777)
- [x] Avoid duplication of the docs by using *copydoc* (see  #9668 )

If anyone has something to add, feel free.

agrayver:
@Rombur I see you already implemented merge/shift -- do you see any caveats for other functions?

Rombur:
Being able to do step-49 in python would be really great! I think that it should be straight forward. You should be able to follow the recipe used for the functions that have already been ported but let me know if you encounter any problem.

agrayver:
I would like to consult regarding the nature and place of the new python tutorials (step-49, step-53 initially). They will replicate the functionality of the original programs, using the python bindings and will be presented using the python notebooks.

Should they be placed next to the C++ programs, in the corresponding folders? 

@Rombur @masterleinad ping

Rombur:
I think they should go with the other python notebooks in `contrib/python-bindings/notebooks` and I would add a section in the C++ tutorials that links to the python notebooks.

PR #9178, opened 2019-12-17
Add a set of new python wrappers

agrayver:
Keep extending python wrappers. 

Part of #9015

### DIFF
--- a/contrib/python-bindings/include/cell_accessor_wrapper.h
+++ b/contrib/python-bindings/include/cell_accessor_wrapper.h
@@ -181,6 +181,47 @@ namespace python
     measure() const;
 
     /**
+     * Test whether the cell has children (this is the criterion for activity of
+     * a cell).
+     */
+    bool
+    active() const;
+
+    /**
+     * Return the level within the mesh hierarchy at which this cell is
+     * located.
+     */
+    int
+    level() const;
+
+    /**
+     * Return the index of the element presently pointed to on the present
+     * level.
+     */
+    int
+    index() const;
+
+    /**
+     * Return whether the neighbor is coarser then the present cell.
+     */
+    bool
+    neighbor_is_coarser(const unsigned int neighbor) const;
+
+    /**
+     * Return the how-many'th neighbor this cell is of cell->neighbor(neighbor),
+     * i.e. return the face_no such that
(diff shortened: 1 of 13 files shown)


agrayver:
@Rombur @masterleinad I would appreciate your review here. 

Apart from new methods, I have changed the save triangulation behavour to write in binary instead of the text. Since I already use python binds for real problems, text format appeared much slower and less efficient.

contrib/python-bindings/source/triangulation_wrapper.cc
```diff
@@ -1435,31 +1520,31 @@ namespace python
   void
   TriangulationWrapper::save(const std::string &filename) const
   {
-    std::ofstream                 ofs(filename);
-    boost::archive::text_oarchive oa(ofs);
+    std::ofstream                   ofs(filename);
+    boost::archive::binary_oarchive oa(ofs);
```
Rombur: Any reason for this change?

agrayver: Yes. The text format is slower and occupies much more space on the disk. This is especially apparent in 3-D (I am using this already for production runs...).

agrayver:
@Rombur I addressed the comments.

contrib/python-bindings/include/cell_accessor_wrapper.h
```diff
@@ -180,6 +180,47 @@ namespace python
...
 
+    /**
+     * Test whether the cell has children (this is the criterion for activity of
+     * a cell).
+     */
+    bool
+    active() const;
+
+    /**
+     * Returns the level within the mesh hierarchy at which this cell is
```
Rombur:
```suggestion
     * Return the level within the mesh hierarchy at which this cell is
```

contrib/python-bindings/source/export_triangulation.cc
```diff
@@ -112,6 +112,16 @@ namespace python
 
 
 
+  const char dim_docstring[] =
+    "Return the dimension of the Triangulation                              \n";
+
+
+
+  const char spacedim_docstring[] =
+    "Return the space dimension of the Triangulation                        \n";
```
Rombur: Do you want to add a final dot at the end of the comment.

agrayver:
@Rombur Done.

PR #9179, opened 2019-12-17
Add python interface for FunctionManifold

agrayver:
As title says, enable FunctionManifold usage from the python. Together with #9178, this is the last step before porting tutorial steps 49 and 53 to python. 

Part of #9015


agrayver:
Example

image

contrib/python-bindings/tests/manifold_wrapper.py
```diff
@@ -67,5 +67,47 @@ def test_manifold(self):
...
+        for cell in self.triangulation.active_cells():
+            cell.set_all_manifold_ids(0)
+
+        self.triangulation.refine_global(2)
+
+        volume = 0
+        for cell in self.triangulation.active_cells():
+            volume += cell.measure()
+
+        self.assertTrue(abs(volume - 1.) < 1e-8)
```
Rombur: Won't this test pass even if you don't refine? Would it be possible to have something like https://github.com/dealii/dealii/blob/master/tests/manifold/function_manifold_02.cc and compare with an output file?

include/deal.II/grid/manifold_lib.h
```diff
@@ -593,12 +593,16 @@ class FunctionManifold : public ChartManifold<dim, spacedim, chartdim>
...
+   * Use @p owns_functions to change this behaviour.
    */
   FunctionManifold(
     const Function<chartdim> & push_forward_function,
     const Function<spacedim> & pull_back_function,
-    const Tensor<1, chartdim> &periodicity = Tensor<1, chartdim>(),
-    const double               tolerance   = 1e-10);
+    const Tensor<1, chartdim> &periodicity    = Tensor<1, chartdim>(),
+    const double               tolerance      = 1e-10,
+    bool                       owns_functions = false);
```
Rombur: Oh OK so that's how you fixed the memory leak from python. This works with me but I don't really know anything about the `Manifold` interface. @dealii/dealii Can someone who knows more about the `Manifold` interface give an opinion on this.

agrayver:
I addressed your comments.

Regarding `Manifold` interface. What I did was to expose the option to transfer ownership of the passed Function objects to the FunctionManifold object (the default behaviour remains the same and does not affect any existing code). This functionality already existed internally for the purpose of handling two different cases, as commented below says:

https://github.com/dealii/dealii/blob/e3cb2d61587999d4a8ff80f846221f9af75ccd7d/source/grid/manifold_lib.cc#L1376-L1386

contrib/python-bindings/include/manifold_wrapper.h
```diff
@@ -26,6 +30,43 @@ DEAL_II_NAMESPACE_OPEN
 
 namespace python
 {
+  template <int dim>
+  class FunctionWrapper : public Function<dim, double>
```
masterleinad: I would create a separate file for this function since it is not tied to the `ManifoldWrapper` class and could be used elsewhere. Moving it later wouldn't be backward-compatible.

contrib/python-bindings/include/manifold_wrapper.h
```diff
@@ -45,23 +86,38 @@ namespace python
...
 
     /**
-     * Create CylindricalManifold
+     * Create CylindricalManifold.
      */
     void
     create_cylindrical(const int axis = 0, const double tolerance = 1e-10);
 
+    /**
+     * Create FunctionManifold with string expressions for push and pull
```
masterleinad:
```suggestion
     * Create FunctionManifold with string expressions for the push forward and pull back
```

contrib/python-bindings/include/manifold_wrapper.h
```diff
@@ -45,23 +86,38 @@ namespace python
...
+     * functions.
+     */
+    void
+    create_function_string(const std::string push_forward,
+                           const std::string pull_back);
+
+    /**
+     * Create FunctionManifold with python push and pull functions.
```
masterleinad:
```suggestion
     * Create FunctionManifold with python push forward and pull back functions.
```

contrib/python-bindings/source/export_manifold.cc
```diff
@@ -35,6 +35,15 @@ namespace python
     " Create polar manifold with a given center point           \n";
 
 
+  const char create_function_docstring[] =
+    " Create manifold with given python push and pull functions \n";
```
masterleinad:
```suggestion
    " Create manifold with given python push forward and pull back functions \n";
```

contrib/python-bindings/source/export_manifold.cc
```diff
@@ -35,6 +35,15 @@ namespace python
...
+
+
+  const char create_function_string_docstring[] =
+    " Create manifold with given string expression for push     \n"
```
masterleinad:
```suggestion
    " Create manifold with given string expression for push forward    \n"
```

contrib/python-bindings/source/export_manifold.cc
```diff
@@ -35,6 +35,15 @@ namespace python
...
+    " and pull functions.                                       \n";
```
masterleinad:
```suggestion
    " and pull back functions.                                       \n";
```

contrib/python-bindings/source/manifold_wrapper.cc
```diff
@@ -216,6 +245,56 @@ namespace python
...
+        manifold_ptr =
+          internal::create_function_manifold<2, 3>(push_forward, pull_back);
+      }
+    else if ((dim == 3) && (spacedim == 3))
+      {
+        manifold_ptr =
+          internal::create_function_manifold<3, 3>(push_forward, pull_back);
+      }
+    else
+      AssertThrow(false, ExcMessage("Wrong dim-spacedim combination."));
```
masterleinad: The combination might not be "wrong", but is certainly not implemented.

include/deal.II/grid/manifold_lib.h
```diff
@@ -593,12 +593,16 @@ class FunctionManifold : public ChartManifold<dim, spacedim, chartdim>
    *
    * The tolerance argument is used in debug mode to actually check that the
    * two functions are one the inverse of the other.
+   *
+   * By default, only pointers to the passed functions are stored.
+   * Use @p owns_functions to change this behaviour.
```
masterleinad: I would prefer to make it clearer that you are passing memory ownership by introducing a
```
 FunctionManifold(
    std::unique_ptr<Function<chartdim>> push_forward_function,    
    std::unique_ptr<Function<spacedim>> pull_back_function,
    ...
```
constructor.

agrayver: Hm, you mean by invoking std::move(push_forward_function) in the constructor and assigning the result to a member variable? BTW, this will require changing the member type from SmartPointer to std::unique_ptr, won't it?

masterleinad: > Hm, you mean by invoking std::move(push_forward_function) in the constructor and assigning the result to a member variable?

Yes.


> BTW, this will require changing the member type from SmartPointer to std::unique_ptr, won't it?

Not necessarily. We can do internally whatever we want, e.g. obtaining the underlying pointer via `std::unique_ptr::release()`.

agrayver:
@masterleinad I have updated the interface accordingly.

contrib/python-bindings/source/manifold_wrapper.cc
```diff
@@ -211,7 +249,63 @@ namespace python
...
+      AssertThrow(false,
+                  ExcMessage(
+                    "Given dim-spacedim combination is not implemented."));
+  }
+
+
+
+  void
+  ManifoldWrapper::create_function_string(const std::string push_forward,
+                                          const std::string pull_back)
```
bangerth: We generally pass strings by reference.

source/grid/manifold_lib.cc
```diff
@@ -1321,6 +1321,26 @@ FunctionManifold<dim, spacedim, chartdim>::FunctionManifold(
...
+template <int dim, int spacedim, int chartdim>
+FunctionManifold<dim, spacedim, chartdim>::FunctionManifold(
+  std::unique_ptr<Function<chartdim>> &push_forward_function,
+  std::unique_ptr<Function<spacedim>> &pull_back_function,
+  const Tensor<1, chartdim> &          periodicity,
+  const double                         tolerance)
+  : ChartManifold<dim, spacedim, chartdim>(periodicity)
+  , const_map()
+  , push_forward_function(push_forward_function.release())
+  , pull_back_function(pull_back_function.release())
```
bangerth: You can just copy them here -- no need to write `.release()`.

include/deal.II/grid/manifold_lib.h
```diff
@@ -600,6 +600,15 @@ class FunctionManifold : public ChartManifold<dim, spacedim, chartdim>
     const Tensor<1, chartdim> &periodicity = Tensor<1, chartdim>(),
     const double               tolerance   = 1e-10);
 
+  /**
+   * Same as previous, but takes pointers to the functions.
+   */
+  FunctionManifold(
+    std::unique_ptr<Function<chartdim>> &push_forward_function,
+    std::unique_ptr<Function<spacedim>> &pull_back_function,
```
bangerth: I think this makes sense. But it would be useful to add something along the lines of
```
  This function takes ownership of the Function objects passed as first and second argument,
  and is ultimately in charge of deleting the pointers when the FunctionManifold object is
  destroyed. 
```
Maybe also something like
```
  This constructor is useful because it allows creating function objects at the place
  of calling the constructor without having to name and later delete these objects.
  This allows the following idiom:
     FunctionManifold<dim> manifold (std_cxx14::make_unique<MyPushForward>(...),
                                                           std_cxx14::make_unique<MyPullBack>(...));
```

agrayver:
@bangerth thanks, I fixed those.

source/grid/manifold_lib.cc
agrayver: Hm... If I don't release them, won't they be destroyed then when unique_ptr goes out of scope? Or what do you mean by copying?

bangerth: If you write `pull_back_function(pull_back_function)` (with the second occurrence of the word being the function argument), then you are invoking the move constructor of `std::unique_ptr` since there is no copy constructor of that class. The move constructor moves ownership from the source to the destination objects. That's exactly what you are doing here, just in a more complicated way :-)

masterleinad: The member variable is a `SmartPointer`, not a `std::unique_ptr`. I guess this is correct as is.

bangerth: Oh oh, yes, then that is correct.

agrayver: This is then ready to merge, isn't?

bangerth:
I don't know enough about the Python interfaces. I'd rather have someone else do the merge.

bangerth:
You might want to squash the commits. Also feel free to add a changelog entry!

agrayver:
Done. This is ready from my side.

Rombur:
Looks good to me. @masterleinad can we merge this?

source/grid/manifold_lib.cc
```diff
@@ -1376,13 +1396,16 @@ FunctionManifold<dim, spacedim, chartdim>::clone() const
...
   // pointers point either to the same Function objects used to construct this
-  // function (owns_pointers == false) or that the newly generated manifold
-  // creates internally the push_forward and pull_back functions using the same
-  // expressions that were used to construct this class (own_pointers == true).
-  if (owns_pointers == true)
+  // function or that the newly generated manifold creates internally the
+  // push_forward and pull_back functions using the same expressions that were
+  // used to construct this class.
+  if (const FunctionParser<chartdim> *f =
+        dynamic_cast<const FunctionParser<chartdim> *>(&*push_forward_function))
```
masterleinad: What if one of the constructors taking `Function` objects (instead of `std::string` objects) is a `FunctionParser` object?
In this case, we are trying to create an object with an empty expression string. Hence, I would rather check for `push_forward_expression` and `pull_back_expression` being empty.

agrayver:
@masterleinad good points. I addressed them.

PR #9181, opened 2019-12-19
ECL: specialize read/write operations

peterrum:
In the case with ECL with positive faces, we cannot use the same `face_no` for reading/writing from the solution vector and from the internal data structures (containing mapping, ...). The first one needs actually the right `face_no` (of the positive side); the second one the `face_no` of the negative side.

I would introduce a new internal field in `FEEvaluationBase`:
```cpp
std::array<unsigned int, VectorizedArrayType::n_array_elements> face_no_data;
```
which stores the actual `face_no` for each lane. At the moment only the first entry is used; however the long-term goal is that each lane can have different `face_no` and during reading/writing the data is re-shuffled so that all lanes have the same `face_no` for `evaluation` and `integration`. 

@kronbichler  Any first thoughts?

Related to issue #9153.


kronbichler:
The question is whether we need this as a member variable or if it is enough to decide it when we read the respective data. My idea is that we need a separate `if` clause for the case we need to re-shuffle when not all exterior faces return the same number (or when some of the faces have a different orientation) anyway, so we could put those decisions there and not make it fill the information unconditionally. The main question is whether we need to make the decision only once in a typical evaluation setup or whether it repeats. As far as I can see, we would need it only once when we read from the source vector.

peterrum:
@kronbichler Let's discuss this PR in person since we need a solution for non-structured meshes (not just here).

peterrum:
@kronbichler Sorry for spamming this PR. I added a `loop_cell_centric`, which accepts `std::function`, made some bug fixes, and have finally added the (serial) ECL test.

The test does not contain all the fancy optimizations but is nevertheless better than nothing. Any opinions on the test?

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -7491,6 +7634,13 @@ FEFaceEvaluation<dim,
   if (!(evaluate_values + evaluate_gradients))
     return;
 
+  const unsigned int face_no =
+    (this->dof_access_index ==
+       internal::MatrixFreeFunctions::DoFInfo::dof_access_cell &&
+     this->is_interior_face == false) ?
+      compute_face_no_data()[0] :
+      this->face_no;
+
```
peterrum: @kronbichler I have modified the code so that the function `compute_face_no_data` is only called for exterior faces in the case of ECL. At the moment, I am only using the information of the first lane, but in a next PR I would also use the information of other lanes to be able to support unstructured meshes.

peterrum:
Closing this in favor of #11747.

PR #9182, opened 2019-12-20
Deprecate some versions of DoFTools::extract_dofs().

bangerth:
Ever since we started using C++11, it has bothered me that we have a lot of functions that return their results in output arguments (typically in last position, but not always -- e.g., if there are additional defaulted arguments). This can now be done more efficiently because we can move objects, and I've wanted to address this for at least some of the functions where this is the case.

Here is a start: A couple of functions in `DoFTools`.


source/dofs/dof_tools.cc
```diff
@@ -227,9 +227,7 @@ namespace DoFTools
 
       // then loop over all cells and do the work
       std::vector<types::global_dof_index> indices;
-      for (typename DoFHandlerType::active_cell_iterator c = dof.begin_active();
-           c != dof.end();
-           ++c)
+      for (const auto c : dof.active_cell_iterators())
```
drwells:
```suggestion
      for (const auto &c : dof.active_cell_iterators())
```

source/dofs/dof_tools.cc
```diff
@@ -443,6 +441,70 @@ namespace DoFTools
...
+          this_selected_dofs[i] =
+            component_mask[local_component_association[i]];
+
+        local_selected_dofs.emplace_back(std::move(this_selected_dofs));
+      }
+
+    // Then loop over all cells and do the work
+    IndexSet                             selected_dofs(dof.n_dofs());
+    std::vector<types::global_dof_index> local_dof_indices;
+    for (const auto c : dof.active_cell_iterators())
```
drwells:
```suggestion
    for (const auto &c : dof.active_cell_iterators())
```

bangerth:
Thanks, @drwells. I've updated the code based on your suggestions.

PR #9186, opened 2019-12-21
Avoid variable template in hp::Refinement

masterleinad:
Fixes the failing build in https://cdash.43-1.org/viewBuildError.php?onlydeltap&buildid=3370.
Generic lambdas are a C++14 feature.

### DIFF
--- a/source/hp/refinement.cc
+++ b/source/hp/refinement.cc
@@ -31,25 +31,6 @@
 
 DEAL_II_NAMESPACE_OPEN
 
-namespace
-{
-  /**
-   * ComparisonFunction returning 'true' or 'false' for any set of parameters.
-   *
-   * These will be used to overwrite user-provided comparison functions whenever
-   * no actual comparison is required in the decision process, i.e. when no or
-   * all cells will be refined or coarsened.
-   */
-  template <typename Number>
-  hp::Refinement::ComparisonFunction<Number> compare_false =
-    [](const Number &, const Number &) { return false; };
-  template <typename Number>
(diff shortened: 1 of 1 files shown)


source/hp/refinement.cc
```diff
@@ -249,6 +230,15 @@ namespace hp
...
              dealii::GridRefinement::ExcInvalidParameterValue());
 
+      // ComparisonFunction returning 'true' or 'false' for any set of
+      // parameters. These will be used to overwrite user-provided comparison
+      // functions whenever no actual comparison is required in the decision
+      // process, i.e. when no or all cells will be refined or coarsened.
+      ComparisonFunction<Number> compare_false =
+        [](const Number &, const Number &) { return false; };
+      ComparisonFunction<Number> compare_true =
+        [](const Number &, const Number &) { return true; };
```
bangerth: can you make these `const`?

bangerth:
As a complete aside, it's not that the thing is a generic lambda (which would be a structure with a `template <typename T> operator() (...arguments depending on T...)`), but it's the fact that the variable that the lambda is assigned to is a template: https://en.cppreference.com/w/cpp/language/variable_template This is not allowed in C++11, whereas the definition of the lambda itself had a clearly specified argument list (no template) that should have been ok.

Either way, the change is of course correct -- thanks for tackling it!

masterleinad:
Yes, you are right.

source/hp/refinement.cc
masterleinad: Fixed.

PR #9188, opened 2019-12-21
Include dof_accessor.h

peterrum:
I want to compile the following program:
```cpp

using namespace dealii;

int
main()
{
  FE_DGQ<2> fe(1);

  Triangulation<2> tria;

  DoFHandler<2> dof_handler(tria);
  dof_handler.distribute_dofs(fe);

    if (cell->active())
      {
        // do something
      }
}
```

without `#include <deal.II/dofs/dof_accessor.h>`. Currently, the compiler complaints if I do not include that file and says `invalid use of incomplete type ‘class dealii::DoFCellAccessor<dealii::DoFHandler<2>, false>’`. This PR fixes this.

### DIFF
--- a/include/deal.II/dofs/dof_handler.h
+++ b/include/deal.II/dofs/dof_handler.h
@@ -30,6 +30,7 @@
 
 #include <deal.II/dofs/block_info.h>
 #include <deal.II/dofs/deprecated_function_map.h>
+#include <deal.II/dofs/dof_accessor.h>
 #include <deal.II/dofs/dof_faces.h>
 #include <deal.II/dofs/dof_iterator_selector.h>
 #include <deal.II/dofs/dof_levels.h>


bangerth:
Well, how is this different than any other case where you want to use a class that has only been forward declared? If you want to use a class, you have to `#include` the appropriate header *in the file where you want to use it*.

peterrum:
I find it very odd, if as a user who wants to use `dof_handler.active_cell_iterators()`, the user has to include both `deal.II/dofs/dof_handler.h` and `deal.II/dofs/dof_accessor.h`. The latter because of the return type of the function. As a user you want to use whatever is returned (you don't want to explicitly care about the type - optimally you only write `auto`).

bangerth:
Does the error result from calling `dof_handler.active_cell_iterators()`, or from calling `if (cell->active())`?

peterrum:
@bangerth  It is coming from `dof_handler.active_cell_iterators()` (I have commented out `if (cell->active())`).

bangerth:
Hm. Then I think you might have a point. Let's see what happens...

/rebuild

peterrum:
@bangerth All tests have run (and passed), only OSX has not started...

peterrum:
Thanks @masterleinad. But your 

> /rebuild

did not work. Lets hope that pushing again helps...

masterleinad:
No, it doesn't. There is some problem with the `OSX` tester. Hence, I would just ignore that one for now.

PR #9193, opened 2019-12-24
Let step-61 only build on step-51, not step-4.

bangerth:
The tutorial graph is complicated enough as it is. We don't need to list dependencies
on some of the more basic tutorials.


tjhei:
yes, I think any indirect dependencies to basic tutorials don't need to be spelled out.

PR #9195, opened 2019-12-25
Create new dealii::Triangulation::create_triangulation()

### LINKED ISSUE #9192: Create/Load grid with local refinement
@nicola-giuliani 

In a project we are following, it has become necessary to have an input-output format that supports grids with local refinement.

While it is quite easy to store this information in a vtk file (we just save all cells, including inactive ones, maybe with information about what level they belong to in a cell-data field), it is not so trivial to create a triangulation from this information.

I would like to discuss here what would the best approach be. We were thinking of adding a Triangulation::create_triangulation that takes vector versions of the current arguments, one per level, and that internally figures out child-parent relationship, and fills in a grid that is locally refined.

The current implementation would then be simply calling the new one, with argument vectors containing a single level (the coarsest).

The following is the proposed interface:
```
virtual void create_triangulation (const std::vector< Point< spacedim >> &vertices, const std::vector< std::vector< CellData< dim >> > &cells_per_level, const std::vector< SubCellData > & subcelldata_per_level)
```

Before we start implementing stuff, we thought it would be a good idea to hear what others think.

peterrum:
This PR adds the method `dealii::Triangulation::create_triangulation()` (by moving part of the current implementation in `dealii::fullydistributed::Triangulation`).

We need to decide about the fate of `p::f::ConstructionData` and `p::f::Utilities` (latter might be useful now outside of the `p::f`-namespace).

ping @luca-heltai 

Closes #9192.


include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,253 @@ namespace internal
...
+       */
+      types::material_id manifold_id;
+
+      /**
+       * Manifold id of all lines of the cell.
+       *
+       * @note Only used for 2D and 3D.
+       */
+      std::array<types::manifold_id, GeometryInfo<dim>::lines_per_cell>
+        manifold_line_ids;
```
luca-heltai: This is better than the original version of CellData. My vote would be to simply use this one instead of the old one, with a constructor that does "the right thing" when initialized in the same way as it was before.

luca-heltai:
In the documentation of p::f::t you specifically say that after a Triangulation has been created with this method, it cannot be refined/coarsened anymore. Why? After all, you simply apply a few steps of refinements right after creating a coarse grid. What blocks you from further refinements/coarsenings afterwards?

peterrum:
@luca-heltai I will implement your suggestions tonight (namespace, documentation).

Regarding the 'smoothing': I have never used it and the reviewers did not point out that this is not out by the current `ConstructionData`. But I guess to add the described functionality should be so hard (as you described).  However, I would keep this out of this PR.

Regarding the dynamic coarsening and refinement in `p:f:t`: PR #8834 is addressing that, but it is still not finished. The practical issue is to have only a single layer of ghost cells and is to coarsen artificial cells as much as possible.

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,253 @@ namespace internal
 /*------------------------------------------------------------------------*/
 
 
+namespace parallel
+{
+  namespace fullydistributed
+  {
```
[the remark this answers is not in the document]
peterrum: I have renamed the namespace to `construction`, but we can still name it somehow differently...

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,253 @@ namespace internal
...
+       * communication.
+       */
+      construct_multigrid_hierarchy = 0x1
+    };
+
+    /**
+     * Information needed for each locally relevant cell, stored in
+     * ConstructionData and used during construction of a
+     * parallel::fullydistributed::Triangulation. This struct stores
+     * the cell id, the subdomain_id and the level_subdomain_id as well as
```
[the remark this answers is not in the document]
peterrum: I guess so. This `CellData` also contains information related to faces and edges...

peterrum:
I also moved the content of `distributed/fullydistributed_util.h` to  `grid/construction_util.h` so that we can use the functions contained there to setup `construction::ConstructionData`.

I'll update the documentation tomorrow.

peterrum:
@luca-heltai I have updated the documentation! Feel free to make changes directly.

include/deal.II/distributed/fully_distributed_tria.h
```diff
@@ -49,164 +49,25 @@ namespace parallel
...
-       */
-      default_setting = 0x0,
-      /**
-       * This flags needs to be set to use the geometric multigrid
-       * functionality. This option requires additional computation and
-       * communication.
-       */
-      construct_multigrid_hierarchy = 0x1
-    };
+    using Settings = construction::Settings;
```
luca-heltai: I'm not sure we want to keep this in. We had no releases yet with p::f::t, so there is no need to keep backward compatibility. But that's just my two cents. @kronbichler ?

include/deal.II/grid/construction_util.h
```diff
@@ -0,0 +1,120 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_grid_construction_util_h
+#define dealii_grid_construction_util_h
```
luca-heltai: I would not shorten the file name since it is not mnemonic. I would either use `utilities` or `construction_utilities`, with a slight preference for the second.

luca-heltai: It is true that we shorten `triangulation` in `tria` consistently, however this is only true for triangulation (and already troublesome, IMHO). Introduring other short names around makes life very difficult... we have `base/utilities`, it would make sense to have `grid/utilities`, or not to add anything else, and just add it to `GridTools`. For this, though, I'll let others comment.

source/distributed/fully_distributed_tria.cc
```diff
@@ -60,7 +60,8 @@ namespace parallel
       settings = construction_data.settings;
 
       // set the smoothing properties
-      if (settings & construct_multigrid_hierarchy)
+      if (settings &
+          parallel::fullydistributed::Settings::construct_multigrid_hierarchy)
```
luca-heltai: Make this`construction::Settings` (same thing everywhere else). Don't use deprecated things in the library.

include/deal.II/distributed/fully_distributed_tria.h
kronbichler: I agree, we don't need backwards compatibility at this point.

include/deal.II/grid/construction_util.h
peterrum: I have updated the code. I would prefer if we could keep a separate file for these functionality.

peterrum:
Any opinions on this PR? @luca-heltai and I would be glad about some feedback regarding the namespace/class/file names!

include/deal.II/distributed/tria.h
```diff
@@ -406,6 +406,15 @@ namespace parallel
...
                            const SubCellData &subcelldata) override;
 
+      /*
+       * @copydoc Triangulation::create_triangulation()
+       *
+       * @note Not inmplemented yet.
+       */
+      virtual void
+      create_triangulation(const construction::ConstructionData<dim, spacedim>
+                             &construction_data) override;
```
bangerth: This function shouldn't be very hard to implement, though. You'd just need to call the corresponding function of the base class and then just duplicate the logic of what happens after construction from the other `create_triangulation()` function of this class.

include/deal.II/grid/construction_util.h
```diff
@@ -0,0 +1,120 @@
...
+
+#include <deal.II/distributed/fully_distributed_tria.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * A namespace dedicated to the struct ConstructionData, which can be used in
+ * Triangulation::create_triangulation().
+ */
+namespace construction
```
bangerth: This is my chance to rectify something that I think we got wrong last time: I find the name `ConstructionData` just a bit too generic, because it doesn't give away what is actually being constructed here. I missed that in the original PRs, but have rued the mistake ever since :-)

The same is true for the current namespace: `construction` in the global namespace could literally mean anything ;-) So let's be better about this then. How about `namespace GridDescription` (sticking with the various `Grid*` classes), or `namespace MeshDescription` (I think we also use the word "mesh" in some places), or just `namespace TriangulationDescription`.

The main class in the namespace would then simply be `Description` or again `XxxDescription`.

include/deal.II/grid/construction_util.h
```diff
@@ -0,0 +1,120 @@
...
+     *       to be able to use this
+     *       feature in the case of parallel::fullydistributed::Triangulation,
+     *       the user has to enable multilevel hierarchy support in
+     *       parallel::fullydistributed::Triangulation.
+     *
+     * @author Peter Munch, 2019
+     */
+    template <int dim, int spacedim = dim>
+    ConstructionData<dim, spacedim>
+    create_construction_data_from_triangulation(
```
bangerth: the name of the function would then mirror the new name of the structure.

include/deal.II/grid/construction_util.h
```diff
@@ -0,0 +1,120 @@
...
+     * @param construct_multilevel_hierarchy Construct multigrid levels.
+     * @return ConstructionData to be used to setup a Triangulation.
+     *
+     * @author Peter Munch, 2019
+     */
+    template <int dim, int spacedim = dim>
+    ConstructionData<dim, spacedim>
+    create_construction_data_from_triangulation_in_groups(
+      std::function<void(dealii::Triangulation<dim, spacedim> &)>
+                                              serial_grid_generator,
```
bangerth: take this and the next argument as `const` references

include/deal.II/grid/construction_util.h
```diff
@@ -0,0 +1,120 @@
+// ---------------------------------------------------------------------
```
bangerth: I'd say just merge this file into the one where you declare the relevant data structures.

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
 /*------------------------------------------------------------------------*/
 
 
+namespace construction
```
bangerth: same comment as above for the namespace name, of course.

I'd just break this part out into a file of its own, for simplicity, and merge it with the separate file you have for the utilities

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+{
+  /**
+   * Configuration flags for fully distributed Triangulations.
```
bangerth: no longer just for these

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+   * @author Peter Munch, 2019
+   */
+  enum Settings
+  {
+    /**
+     * Default settings, other options are disabled.
+     */
+    default_setting = 0x0,
+    /**
+     * This flags needs to be set to use the geometric multigrid
```
bangerth:
```suggestion
     * This flag needs to be set to use the geometric multigrid
```

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+     * functionality. This option requires additional computation and
+     * communication.
+     */
+    construct_multigrid_hierarchy = 0x1
+  };
+
+  /**
+   * Information needed for each locally relevant cell, stored in
+   * ConstructionData and used during construction of a
+   * parallel::fullydistributed::Triangulation. This struct stores
```
bangerth: no longer just these

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+   * the cell id, the subdomain_id and the level_subdomain_id as well as
+   * information related to manifold_id and boundary_id.
+   *
+   * @author Peter Munch, 2019
+   */
+  template <int dim>
+  struct CellData
```
bangerth: Could this class be derived from `::CellData`? You'd then only have to implement whatever *else* you need for the current operation.

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+
+
+
+  /**
+   * Data used in Triangulation::create_triangulation().
+   *
+   * @author Peter Munch, 2019
+   */
+  template <int dim, int spacedim>
+  struct ConstructionData
```
bangerth: see name suggestion above

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+    /**
+     * Comparison operator.
+     */
+    bool
+    operator==(const ConstructionData<dim, spacedim> &other) const;
+
+    /**
+     * Cells of the locally-relevant coarse-grid triangulation.
+     */
+    std::vector<dealii::CellData<dim>> coarse_cells;
```
bangerth: If you derive the new `CellData` class from `::CellData`, then you don't need the separate `cell_infos` variable below. I think that would make things substantially simpler since the two objects currently don't really have clearly separated purposes.

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+
+    /**
+     * The MPI communicator used to create this struct. It will be compared
+     * to the communicator inside of parallel::fullydistributed::Triangulation
+     * and an assert is thrown if they do not match.
+     *
+     * @note Please note this is necessary since the communicator inside of
+     * parallel::TriangulationBase is const and cannot be changed after the
+     * constructor has been called.
+     *
```
bangerth:
```suggestion
```

include/deal.II/grid/tria.h
```diff
@@ -511,6 +513,167 @@ namespace internal
...
+     */
+    MPI_Comm comm;
+
+    /**
+     * @note settings See the description of the Settings enumerator.
```
bangerth:
```suggestion
     * Properties to be use in the construction of the triangulation.
```

include/deal.II/grid/tria.h
```diff
@@ -4181,6 +4356,88 @@ namespace internal
...
+
+
+  template <int dim>
+  bool
+  CellData<dim>::operator==(const CellData<dim> &other) const
+  {
+    if (this->id != other.id)
+      return false;
+    if (this->subdomain_id != other.subdomain_id)
+      return false;
```
bangerth: You can simplify this and the next function by just saying
```
  return ((this->id == other.id)
             &&
            (this->subdomain_id == other.subdomain_id)
            &&
            ...);
```
The compiler internally converts that into exactly what you wrote, i.e., the chain of evaluations is terminated once it encounters the first `false` comparison (because then the rest of the comparisons no longer matter).

include/deal.II/grid/tria.h
peterrum: The issue here that `ConstructionData` depends on `CellData` and `Triangulation` on `ConstructionData` and the `ConstructionData`-utility functions on `Triangulation`...

peterrum:
@bangerth First of all, thanks a lot for your 6h review marathon yesterday! I will tackle your suggestions later today!

bangerth:
On 1/7/20 1:22 AM, Peter Munch wrote:
> @bangerth <https://github.com/bangerth> First of all, thanks a lot for your 6h 
> review marathon yesterday!

Ha, yes, I guess I went through a lot of pull requests yesterday ;-)

include/deal.II/grid/tria.h
bangerth: Couldn't you get away with forward declarations in most cases?

include/deal.II/distributed/tria.h
peterrum: Future work...

include/deal.II/grid/construction_util.h
peterrum: I went with `TriangulationDescription::Description`. The corresponding files are called now `tria_description.h`/`tria_description.cc`.

The utility functions are called now:
* `TriangulationDescription::Utilities::create_description_from_triangulation`
* `TriangulationDescription::Utilities::create_description_from_triangulation_in_groups`

include/deal.II/grid/tria.h
peterrum: I am not really in favor of this. The only thins which is common `material_id` and `manifold_id`. `::CellData::vertices` is not common. However, we can change this in the future depending on how @luca-heltai intends to implement his features.

include/deal.II/grid/tria.h
peterrum: The same as above. This is the coarse-grid description, where we only need the `vertices` from `::CellData`.

include/deal.II/grid/tria.h
peterrum: I leave this for the future.

include/deal.II/grid/tria.h
peterrum: I have moved `TriangulationDescription::Description` and the other data structures to the new file. However, I have also moved `CellData` and `SubCellData` to that file. The file `tria.h` includes now `tria_description.h`; and `tria_description.h` includes `tria.h`; this cyclic dependency is a real mess.

> Couldn't you get away with forward declarations in most cases?

As a side remark,  I consider forward declarations as a pure evil "feature".

peterrum:
@luca-heltai @bangerth Could you have another look at the code! Thanks!

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+      const bool construct_multilevel_hierarchy)
+    {
+#ifndef DEAL_II_WITH_MPI
+      (void)serial_grid_generator;
+      (void)serial_grid_partitioner;
+      (void)comm;
+      (void)group_size;
+      (void)construct_multilevel_hierarchy;
+
+      return ConstructionData<dim, spacedim>();
```
masterleinad:
```suggestion
      return Description<dim, spacedim>();
```

peterrum: Thanks!

peterrum:
Could we proceed with this PR. All new `p:f:t` implementations will be in conflict with this one.

include/deal.II/grid/tria_description.h
```diff
@@ -0,0 +1,571 @@
...
+
+    /**
+     * level_subdomain_id of the cell.
+     */
+    types::subdomain_id level_subdomain_id;
+
+    /**
+     * Manifold id of the cell.
+     */
+    types::material_id manifold_id;
```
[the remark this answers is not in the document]
peterrum: Yes, it should. I have noticed that yesterday also https://github.com/dealii/dealii/issues/9353#issuecomment-575323216.

peterrum:
> I find some of the things in this PR very confusing. In order to understand how things work, I tried to implement a `write_vtk` function for `ConstructionData`, and this is what I understood:
> 
> * `TriangulationDescription::CellData` has some overlap with `dealii::CellData` (including the name). As far as I could understand, you interpret this as a collection of **some** of the ids pertinent to a cell (including **some** of the ids of the cell subobjects)

I do not get what you mean by **some of the ids pertinent to a cell*. For locally relevant cells, it contains a globally unique ID, which is needed to determine which cell should be refined. 

> * `TriangulationDescription::CellData`  does _not_ contain any information on connectivity

Yes. This information is here https://github.com/dealii/dealii/blob/663cd41dafd96892fe36e94200f23d9103bffc8d/include/deal.II/distributed/fully_distributed_tria.h#L169-L177

> * `TriangulationDescription::CellData`  does _not_ contain any information on material_ids

Not at the moment. But we can change that. Although, I this is not really relevant information to create the grid, is it?

> * `TriangulationDescription::CellData`  is _not_ intended to be used with objects of dimension < `dim`

Till now, I only used if for `dim==spacedim`. But we can extend it...

> 
> on the other hand, we have `dealii::CellData` that:
> 
> * Should really be called `dealii::TriaObjectData`, since it is used also for objects of dimension < dim
> * Contains connectivity information
> * Contains material ids/boundary ids information _only of the specific object_
> * Contains manifold id information _only of the specific CellData object_
> * Does not have a way to connect to its lower dimensional SubCell data structures, nor to its children (if any)

Yes.

> 
> In particular, I think you use `TriangulationDescription::CellData` to
> 
> 1. set the manifold ids of all subobjects of the coarse level to the correct value
> 2. figure out what cell on level `i` needs to be refined
> 3. refine all needing cells (hoping they get on the correct spot thanks to the manifolds)
> 4. repeat step 2 and 3
> 5. set boundary ids correctly (at the end of all loops)

Correct.

> 
> If I understood correctly, in Step 1 of your implementation, you do not create a coarse grid with the correct manifold ids, because you don't use SubCellData to figure out what manifold ids to use. Instead you create a corse grid without using SubCellData, and then correct the ids using `TriangulationDescription::CellData` on the coarse grid.
> 

Yes.

> Is there a reason you don't want to use SubCellData to impose manifold ids? This could render your `TriangulationDescription::CellData` unnecessary (upon refinement, deal.II makes sure children inherit most of the ids correctly from their parents), provided that you find a way to assign correctly the other required ids (subdomain_id, level_subdomain_id, etc.)
> 

There is no particular reason. I am not that familiar `SubCellData` with. The data structure and the setup functions were constructed with `p:f:t` in mind and should be usable for it also in the future.

> I think we should rethink a little these structure, so that we don't replicate existing information, and we don't create more and more classes with the same name but different functionality.
> 
> I have to go now, but I'll try to write a proposal here later this morning.

I am looking forward to your proposal.

luca-heltai:
> I do not get what you mean by "some of the ids pertinent to a cell". For locally relevant cells, it contains a globally unique ID, which is needed to determine which cell should be refined.

some of the ids = globally unique id, manifold id of the cell, manifold ids of subobjects, subdomain_id, level_subdomain_id

>  There is no particular reason. I am not that familiar SubCellData with. The data structure and the setup functions were constructed with p:f:t in mind and should be usable for it also in the future.

I'm not sure I grasp all that is needed, but my (long term) proposal is the following:

- We create a `TriaObject<structdim>` class, containing
  - Manifold id
  - The union of {Material id/boundary id} (as it is now in SubCellData)
  - connectivity
  - unique global id (this is what you need, and what you have in `TriangulationDescription::CellData`)

This is identical to the current `dealii::CellData<structdim>`, with the addition of the unique global id. To preserve backward compatibility, we can have the unique global id set to a default invalid value, so that the construction of (flat) grids using this class is not different from the previously working case.

- We add a new  `dealii::CellData<dim>`, making it derived from `TriaObject<dim>`, and containing in addition:
  - domain id
  - level_domain_id
  - a vector of indices pointing to non-default quad objects
  - a vector of indices pointing to non-default line objects

The last two vectors should match faces and lines in the corresponding entries in the `SubCellData` structure (that should have a vector of non default `TriaObject<1>` and one of non default `TriaObject<2>`).

The current logic inside the old `create_triangulation` did not require the last two vectors, since it reconstructed that information from the SubCellData itself, withouth the need to specify parent relationship when building the grid. The same could be done in other levels, so the last two vectors may not be necessary.

- For every level, we store a vector of CellData, and a SubCellData (containing all lines and quads with non default ids).

In the input to `create_triangulation` for the serial case, we would only need 
- a vector of points (all of them as taken from tria.get_vertices())
- a vector of vectors of CellData 
- a vector of SubCellData (one for each level) 

This is essentialy what you are storing. Notice that, your solution stores **all** manifold ids of **all** subobject of a cell, many time (in three dimensions, for every edge, you have its manifold id stored in all incident cells, and for every face, you have two cells storing its mainfold ids).
In your current code no check is performed to guarantee that all these are coherent, and last id wins.

The current solution in deal.II, instead, only stores subcelldata for objects containing non-default ids, and only once per object.

Do you fore see a problem with this structure in the code that generates `p::f::t`?

If we want to postpone these changes, I perfectly agree, provided that at least you rename the current `TriangulationDescription::CellData` into `TriangulationDescription::CellInfo` or similar. It should have a different name if it does a different thing.

peterrum:
@luca-heltai Thanks for your explanation. I am aware that saving the information of all edges and faces is far from optimal (and not safe). Your description is totally reasonable. 

Could further elaborate:
- a vector of indices pointing to non-default quad objects
- a vector of indices pointing to non-default line objects

Would you want to save for all faces and lines an index? I used to have - instead of the `std::array` - a `std::vector` non-default faces and edges but that was prevented by reviewers at that time...

Currently I do not see any issues from `p:f:t` perspective. But let me sleep one night over your proposal.

 I would definitely support your proposal; I am - however - not sure how much time I will have to rewrite the code adequately (I have some deadlines). However, if you want to begin I will support you.

As far as I see it once the new `create_triangulation`-function and the utility functions (converting the serial and distributed triangulation to the data structure you propose) work, `p:f:t` is fine. That the definition of the old `ConstructionData` changes is not a problem, since I don't think anyone creates this data structure on their own but use the utility functions.

I am very much looking forward to the changes!

luca-heltai:
Let me make an example for a 2d square, created in this way:
~~~
tria.begin_active()->face(0)->set_manifold_id(1);

tria.begin_active()->set_refine_flag();
tria.execute_coarsening_and_refinement();

tria.begin_active()->set_refine_flag();
tria.execute_coarsening_and_refinement();

GridTools::transform(transformation, tria);
~~~

Now, the transformation has moved all the vertices around a bit.

To create the coarsest level, we would need a SubCellData containing a single `CellData<1>` for face 0 on level 0.

Upon refinement, all parent faces of tria.begin(0)->face(0),  will inherit the manifold id 1, so there is no need in subsequent levels to set manifold ids to subobjects (unless you change them explicitly after some refinement steps, which you may want to do. In this case, we need a way to identify on level i the subobjects that have a manifold id different from their parents, and insert them in the SubCellData first, and attach the content of their id during the construction of the finer levels).

These three objects would be enough to regenerate the grid in a consistent way.
~~~
std::vector<Point<2>> vertices(13); // all vertices
std::vector<std::vector<CellData<2>>> level_cells = {{c0}, {c1, c2, c3, c4}, {c5, c6, c7}};
std::vector<SubCellData> level_sub_cells = {S0, SubCellData(),  SubCellData()};
~~~

Now `create_triangulation` with arguments  `level_cells[0]` and `level_sub_cells[0]` will create a consistent coarse triangulation, with manifold ids and boundary ids set by `S0` for lower dimensional objects, and by `level_cells[0]` for the co-dimension zero objects.

In 99% of the cases, simply identifying the cells to refine, and resetting the vertex position to the correct position would produce an identical mesh to the original one.

For this to work, we need *connectivity* on all levels, and we need *all vertices* (not just the coarse ones).

With this procedure, no additional data structure is necessary to recreate a Triangulation. Adding the CellId to CellData should make this work also for `p::f::t`.

If we want to support the modification of manifold ids on finer levels, we need to keep track of the non-default ids of each sub object.

A vector of `std::pair<subobject_id, manifold_id>`, as you have for the boundary id, would solve the issue without using any SubCellData on any level other than the coarse one.

I am in favour of merging this PR, and then move forward a better solution later on. I should have chimed in when you first proposed these functions in `p::f::t`, but did not have the time then. Sorry. :)

peterrum:
@luca-heltai @bangerth I have squashed all commits and rebased this PR.

I would appreciate it if we could merge this PR as it is now soon. This PR has introduced a new struct `TriangulationDescription::Description`, which is used in the new method `Triangulation::create_triangulation()`. The old `parallel::fullydistributed::Triangulation::create_triangulation()` has been modified  in such a way that it calls the implementation of the base class and only does some pre- and post-processing. 

The interface of `dealii::Triangulation::create_triangulation()` and `dealii::parallel::fullydistributed::Triangulation::create_triangulation()` (and the fact that the `p:f:t` implementation calls that of the base class) will not change in the near future and will continue to take a `TriangulationDescription::Description`. The `p:f:t` test will probably not be effected so that we a way to tests changes.

What will change - by the implementation of the proposal of @luca-heltai, which I support  - is only the definition of `TriangulationDescription::Description` and consequently the utility functions. @luca-heltai  and I should do the changes in coordinated way till the release of version 9.2. If this PR is merged, we can tackle this in small portion (when one of us has time). 

I am looking forward that `TriangulationDescription::Description` and the utility functions will be generally usable also for other triangulations. This makes the content of the `parallel::fullydistributed`-namespace very small and specialized to fully distributed triangulations, which apparently does not need a lot.

peterrum:
I have rebased this branch and removed all the merge conflicts. Does anyone fancy to add `ready to test label and /rebuild`?

luca-heltai:
I think we can merge this as is, and fill in the details later on. I don't want to hold @peterrum's work any longer. ;)

masterleinad:
```
In file included from /jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria_description.h:24:0,
                 from /jenkins/workspace/dealii_PR-9195/tests/all-headers/test_header.cc:17:
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1821:42: error: ‘CellData’ was not declared in this scope
                        const std::vector<CellData<dim>> &  cells,
                                          ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1821:51: error: template argument 1 is invalid
                        const std::vector<CellData<dim>> &  cells,
                                                   ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1821:51: error: template argument 2 is invalid
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1821:54: error: expected ‘,’ or ‘...’ before ‘>’ token
                        const std::vector<CellData<dim>> &  cells,
                                                      ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1835:11: error: ‘TriangulationDescription’ does not name a type
     const TriangulationDescription::Description<dim, spacedim>
           ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1835:48: error: expected unqualified-id before ‘<’ token
     const TriangulationDescription::Description<dim, spacedim>
                                                ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1835:48: error: expected ‘)’ before ‘<’ token
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1835:37: error: expected ‘;’ at end of member declaration
     const TriangulationDescription::Description<dim, spacedim>
                                     ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1835:48: error: expected unqualified-id before ‘<’ token
     const TriangulationDescription::Description<dim, spacedim>
                                                ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1849:23: error: ‘CellData’ was not declared in this scope
     const std::vector<CellData<dim>> &  cells,
                       ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1849:32: error: template argument 1 is invalid
     const std::vector<CellData<dim>> &  cells,
                                ^
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1849:32: error: template argument 2 is invalid
/jenkins/workspace/dealii_PR-9195/include/deal.II/grid/tria.h:1849:35: error: expected ‘,’ or ‘...’ before ‘>’ token
     const std::vector<CellData<dim>> &  cells,
                                   ^
make[3]: *** [CMakeFiles/grid-tria_description.h.debug.dir/test_header.cc.o] Error 1
make[2]: *** [CMakeFiles/grid-tria_description.h.debug.dir/all] Error 2
make[1]: *** [CMakeFiles/grid-tria_description.h.debug.build.dir/rule] Error 2
make: *** [grid-tria_description.h.debug.build] Error 2
```

peterrum:
@masterleinad Let's hope this works!

Rombur:
This cannot be merged because @bangerth needs to approve the PR.

include/deal.II/grid/tria_description.h
```diff
@@ -0,0 +1,571 @@
...
+     * where the partitioning is adopted unaltered.
+     *
+     * @param tria Partitioned input triangulation.
+     * @param comm MPI_Communicator to be used. In the case
+     *             of dealii::distributed::Triangulation, the communicators
+     *             have to match.
+     * @param construct_multilevel_hierarchy Signal if the multigrid levels
+     *        should be constructed.
+     * @param my_rank_in Construct Description for this rank (only
+     *                   working for serial triangulations).
```
bangerth: I have to admit that I don't understand this description. Can you elaborate?

include/deal.II/grid/tria_description.h
```diff
@@ -0,0 +1,571 @@
...
+    Description<dim, spacedim>
+    create_description_from_triangulation(
+      const dealii::Triangulation<dim, spacedim> &tria,
+      const MPI_Comm                              comm,
+      const bool         construct_multilevel_hierarchy = false,
+      const unsigned int my_rank_in = numbers::invalid_unsigned_int);
+
+
+    /**
+     * Construct a construction::Description. In contrast
```
bangerth: fix the class name

include/deal.II/grid/tria_description.h
```diff
@@ -0,0 +1,571 @@
...
+     * to the function above, this function is also responsible for creating
+     * a serial triangulation and for its partitioning (by calling the
+     * provided std::functions). Internally only selected processes (every
```
bangerth:
```suggestion
     * provided `std::function' objects). Internally only selected processes (every
```

include/deal.II/grid/tria_description.h
```diff
@@ -0,0 +1,571 @@
...
+     * n-th/each root of a group of size group_size) create a serial
+     * triangulation and the ConstructionData for all processes in its group,
+     * which is communicated.
+     *
+     * @note A reasonable group size is the size of a NUMA domain or the
+     * size of a compute node.
+     *
+     * @param serial_grid_generator A function, which creates a serial triangulation.
+     * @param serial_grid_partitioner A function, which can partition a serial
```
bangerth:
```suggestion
     * @param serial_grid_partitioner A function which can partition a serial
```

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+  namespace Utilities
+  {
+    namespace
+    {
+      /**
+       * Set the user_flag of a cell and of all its parent cells.
+       */
+      template <int dim, int spacedim>
+      void
+      set_user_flag_reverse(TriaIterator<CellAccessor<dim, spacedim>> cell)
```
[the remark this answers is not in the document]
bangerth: I also think that you mean `recursively`? But then you only do it for a cell and its immediate parent, so that also doesn't quite sound right.

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+                      cell->face(f)->boundary_id();
+                    if (boundary_ind != numbers::internal_face_boundary_id)
+                      cell_info.boundary_ids.emplace_back(f, boundary_ind);
+                  }
+
+                // d) compute a new coarse-cell id (by ignoring all parent
+                // level)
+                types::coarse_cell_id new_coarse_cell_id =
+                  convert_cell_id_binary_type_to_level_coarse_cell_id<dim>(
+                    cell->id().template to_binary<dim>());
```
bangerth: Please move the functionality you have in `convert_cell_id_binary_type_to_level_coarse_cell_id` to a member function of `CellId`. What you are doing here is first convert to a binary representation, and then using special knowledge of what's in that binary representation to extract information from the original `CellId`. That's backward and also dangerous because you assume information you shouldn't have. Just let the `CellId` class do this.

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+                    my_rank < dealii::Utilities::MPI::n_mpi_processes(comm),
+                  ExcMessage(
+                    "Rank has to be smaller than available processes."));
+
+      if (auto tria_pdt = dynamic_cast<
+            const parallel::distributed::Triangulation<dim, spacedim> *>(&tria))
+        {
+          AssertThrow(
+            my_rank == numbers::invalid_unsigned_int ||
+              my_rank == dealii::Utilities::MPI::this_mpi_process(comm),
```
bangerth: Shouldn't this be true also for a `shared` triangulation?

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+                       coinciding_vertex_group->second))
+                  vertices_owned_by_locally_owned_cells[co_vertex] = true;
+            }
+        };
+
+      // check if multilevel hierarchy should be constructed
+      if (construct_multilevel_hierarchy == false)
+        {
+          AssertThrow(
+            tria.has_hanging_nodes() == false,
```
bangerth: I'd make this a regular `Assert`

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+          // 1a) clear user_flags
+          const_cast<dealii::Triangulation<dim, spacedim> &>(tria)
+            .clear_user_flags();
+
+          // 1b) loop over levels (from fine to coarse) and mark on each level
+          //     the locally relevant cells
+          for (unsigned int level =
+                 tria.get_triangulation().n_global_levels() - 1;
+               level != numbers::invalid_unsigned_int;
+               level--)
```
bangerth:
```suggestion
               --level)
```

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+
+          // 1a) clear user_flags
+          const_cast<dealii::Triangulation<dim, spacedim> &>(tria)
+            .clear_user_flags();
+
+          // 1b) loop over levels (from fine to coarse) and mark on each level
+          //     the locally relevant cells
+          for (unsigned int level =
+                 tria.get_triangulation().n_global_levels() - 1;
+               level != numbers::invalid_unsigned_int;
```
bangerth: This uses special knowledge of what `invalid_unsigned_int` actually is. We tend to think of this as just some exceptional value like a `NaN` that lives outside the usual values. Just make `level` a `signed int` and use this as condition:
```suggestion
               level >= 0;
```

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+              // 3a) create construction data for other ranks
+              const auto construction_data =
+                create_description_from_triangulation(
+                  tria, comm, construct_multilevel_hierarchy, other_rank);
+              // 3b) pack
+              std::vector<char> buffer;
+              dealii::Utilities::pack(construction_data, buffer, false);
+
+              // 3c) send ConstructionData
+              const auto ierr = MPI_Send(buffer.data(),
```
bangerth: For a follow-up patch: You could be using an Isend here if you pack the data into a buffer that is part of an array that lives outside the loop.

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+                                         buffer.size(),
+                                         MPI_CHAR,
+                                         other_rank,
+                                         mpi_tag,
+                                         comm);
+              AssertThrowMPI(ierr);
+            }
+
```
bangerth: You would then just have to do an `MPI_Allwait` here.

peterrum:
@bangerth Thanks for you comments; I have tackled most of the comments in the a new PR #9432.

include/deal.II/grid/tria_description.h
```diff
@@ -0,0 +1,571 @@
...
+    std::array<types::manifold_id, GeometryInfo<dim>::lines_per_cell>
+      manifold_line_ids;
+
+    /**
+     * Manifold id of all face quads of the cell.
+     *
+     * @note Only used for 3D.
+     */
+    std::array<types::manifold_id,
+               dim == 1 ? 1 : GeometryInfo<3>::quads_per_cell>
```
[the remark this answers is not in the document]
peterrum: This is a workaround needed for `boost::serialization()` so I cannot remove it here.

source/grid/tria.cc
```diff
@@ -10744,6 +10744,125 @@ Triangulation<dim, spacedim>::create_triangulation(
...
+
+  // sort cell_infos on each level separately
+  for (auto &cell_info : cell_infos)
+    std::sort(
+      cell_info.begin(),
+      cell_info.end(),
+      [&](TriangulationDescription::CellData<dim> a,
+          TriangulationDescription::CellData<dim> b) {
+        const CellId a_id(a.id);
+        const CellId b_id(b.id);
```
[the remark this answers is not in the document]
peterrum: Actually not, `a.id` is type of binary id

source/grid/tria_description.cc
peterrum: As discussed in PR #9391, I would like to construct per default all multigrid levels and remove the flattening function here. As a consequence, this function will also not be called anymore and can be removed.

source/grid/tria_description.cc
peterrum: I have never tried out shared triangulations. But I guess this assert does not apply for it. My guess is that the next branch is taken and it should work fine if it is treated like a serial triangulation.

source/grid/tria_description.cc
```diff
@@ -0,0 +1,662 @@
...
+          TriaIterator<CellAccessor<dim, spacedim>> &cell,
+          std::vector<bool> &vertices_owned_by_locally_owned_cells) {
+          // add local vertices
+          for (unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell;
+               ++v)
+            {
+              vertices_owned_by_locally_owned_cells[cell->vertex_index(v)] =
+                true;
+              const auto coinciding_vertex_group =
+                vertex_to_coinciding_vertex_group.find(cell->vertex_index(v));
```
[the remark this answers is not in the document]
peterrum: The map `vertex_to_coinciding_vertex_group` should be quite small since it only contains vertices coinciding due to periodic boundary conditions. 

I have run the function for large triangulations. See https://github.com/dealii/dealii/issues/8558#issuecomment-532250763

source/grid/tria_description.cc
peterrum: This is done intentionally. I only want to create the next `Description` once the current one has been sent away (to reduce memory consumption). I have added a comment.

bangerth: Hm, is this really an issue? I don't think it's particularly important, but suspect that matrices, vectors, and all other objects take far more memory than you might be using here -- but then they also take more time, so I don't think that it matters much.

source/grid/tria_description.cc
bangerth: OK, I look forward to the function disappearing. Please open a github issue that reminds us before the next release to make sure the function is indeed gone!

PR #9198, opened 2019-12-29
A sample code of advection-diffusion equation

saeedbahrami91:
Calculation of temperature distribution by discretizing an advection-diffusion equation with the CDS method in the following two-dimensional figure below:

### DIFF
--- a/A sample code of advection-diffusion equation
+++ b/A sample code of advection-diffusion equation
@@ -1,0 +1,70 @@
+clear all
+clc
+n=100; % number of nodes in x axis
+disp('Number of nodes in x axis = ')
+disp(n)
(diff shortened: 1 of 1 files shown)


peterrum:
If I am not mistaken this is a MATLAB code. How is this related to deal.II?

jppelteret:
Indeed, this does not appear to be directly relevant to our core code base. @saeedbahrami91, is this somehow related to https://github.com/dealii/code-gallery/tree/master/cdr ?

jppelteret:
@saeedbahrami91, can you perhaps explain the intent of this PR? Maybe if we understood better what you're trying to do, then we could help some more.

PR #9202, opened 2019-12-30
Mark a bunch of input arguments as 'const'.

bangerth:


kronbichler:
Strange that the compiler accepted having the arguments marked as `const` in declaration the header but not in the definition.

bangerth:
That's because the `const` on non-reference non-pointers has no implications on the caller side. It's really just a thing of relevance in the implementation of the function, and consequently can differ between declaration and definition.

PR #9204, opened 2019-12-30
mu_parser functions modified to use std random number generators

### LINKED ISSUE #9177: FunctionParser still uses boost::random
#9125 alerts me to the fact that the `source/base/function_parser.cc` file still uses the BOOST random number facilities. #9125 moves this functionality to `source/base/mu_parser_internal.cc`, so we should wait till that PR has been merged, but after that, we should replace the BOOST random facilities by the corresponding C++11 facilities.

@konsim83 -- FYI.

rezarastak:
fixes #9177 

I had to change the output of a test because it depended on the details of the pseudo-random number generator.


rezarastak:
This PR is ready to be reviewed and merged

PR #9205, opened 2019-12-31
Use a range-based for loop.

bangerth:
/rebuild

### DIFF
--- a/include/deal.II/base/thread_management.h
+++ b/include/deal.II/base/thread_management.h
@@ -1833,10 +1833,8 @@ namespace Threads
     void
     join_all() const
     {
-      for (typename std::list<Task<RT>>::const_iterator t = tasks.begin();
-           t != tasks.end();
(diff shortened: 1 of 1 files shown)


include/deal.II/base/thread_management.h
```diff
@@ -1833,10 +1833,8 @@ namespace Threads
     void
     join_all() const
     {
-      for (typename std::list<Task<RT>>::const_iterator t = tasks.begin();
-           t != tasks.end();
-           ++t)
-        t->join();
+      for (auto t : tasks)
```
drwells: Shouldn't this be a reference type?

bangerth: Yes, although it happens to work because all task objects corresponding to the same task point to a single instance of a task base object :-)

bangerth:
I fixed the missing reference.

PR #9206, opened 2019-12-31
Provide TaskGroup::size().

bangerth:
/rebuild

### DIFF
--- a/include/deal.II/base/thread_management.h
+++ b/include/deal.II/base/thread_management.h
@@ -1824,6 +1824,21 @@ namespace Threads
       return *this;
     }
 
+
+    /**
+     * Return how many tasks have been put into this group. This
+     * function does not distinguish how many of these tasks have
+     * already run and have finished, are still waiting to be
+     * scheduled to a CPU resource, or are currently running. Tasks
+     * that have been joined already are also still counted.
+     */
+    std::size_t
+    size() const
+    {
+      return tasks.size();
+    }
+
+
     /**
      * Wait for all tasks in the collection to finish. It is not a problem if
      * some of them have already been waited for, i.e. you may call this
--- a/tests/base/task_group_size.with_threads=on.output
+++ b/tests/base/task_group_size.with_threads=on.output
@@ -1,0 +1,4 @@
+
+DEAL::1
+DEAL::2
+DEAL::2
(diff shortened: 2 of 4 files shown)


include/deal.II/base/thread_management.h
```diff
@@ -1824,6 +1824,21 @@ namespace Threads
...
 
+
+    /**
+     * Return how many tasks have been put into this group. This
+     * function does not distinguish how many of these tasks have
+     * already run and have finished, are still waiting to be
+     * scheduled to a CPU resource, or are currently running. Tasks
+     * that have been joined already are also still counted.
+     */
+    unsigned int
```
drwells: We are never going to have billions of tasks, but I think that this will probably give us a narrowing warning. Could we just use `std::size_t` instead?

tests/base/task_group_size.with_threads=on.output
```diff
@@ -0,0 +1,3 @@
+DEAL::1
+DEAL::2
+DEAL::2
```
drwells: I think there needs to be a blank line at the end:
```

***  File "output" has more lines than file "/jenkins/workspace/dealii_PR-9206/tests/base/task_group_size.with_threads=on.output",
***  line 4 is the last one read from file "output"
```

include/deal.II/base/thread_management.h
bangerth: Sure!

bangerth:
I think I resolved both issues.

drwells:
I misremembered what deallog does - we need an empty line at the beginning, I think. Could you just copy and paste the output file to fix it?

bangerth:
Should now finally work -- I had hoped to klduge my way through it without actually building the entire library with this patch on my laptop, but that clearly didn't work...

drwells:
The OSX tester failed for unrelated reasons:
```
Sending interrupt signal to process

clang: error: unable to execute command: Terminated: 15

clang: error: linker command failed due to signal (use -v to see invocation)

make[2]: *** [lib/libdeal_II.g.9.2.0-pre.dylib] Error 254

make[1]: *** [source/CMakeFiles/deal_II.g.dir/all] Error 2

make: *** [all] Error 2

script returned exit code 2
```

drwells:
The new test passed in other configurations, so unless someone objects I'll go ahead and merge by the end of the day.

tjhei:
> The OSX tester failed for unrelated reasons:

it was killed after 60 minutes. I updated the machine and the new xcode takes longer to compile. :-(

PR #9207, opened 2019-12-31
Implement the solution of complex linear systems using UMFPACK.

bangerth:
This required a bit of work because we only compiled a subset of the UMFPACK functionality. Consequently, I had to take a look at our cmake translation of the UMFPACK Makefile, which was a bit hard to understand. So I rewrote the whole damn thing by translating the Makefile again and staying closer to the original.

The remainder of the work was not overly complicated.


masterleinad:
You need to fix the indentation.

bangerth:
Indentation is fixed now. Review would be appreciated :-)

bangerth:
Ping?

drwells:
I'll take a look at this early next week.

bundled/umfpack/UMFPACK/Source/CMakeLists.txt
```diff
@@ -30,142 +293,86 @@ STRIP_FLAG(DEAL_II_CXX_FLAGS "-Wfloat-conversion")
...
   umfpack_tictoc.cc
   umfpack_timer.cc
   )
 
 
 #-------------------------------------------------------------------------------
-# compile each int and long routine (with no real/complex version)
+# Compile each of the following files as long int routines; UMFPACK also allows
+# compiling with int indices, but we don't use these and so let's not
+# bother -- it only takes compile time unnecessarily.
```
drwells: I guess we never ended up using these.

source/lac/sparse_direct.cc
```diff
@@ -268,24 +293,45 @@ SparseDirectUMFPACK::factorize(const Matrix &matrix)
...
   int status;
-  status = umfpack_dl_symbolic(N,
-                               N,
-                               Ap.data(),
-                               Ai.data(),
-                               Ax.data(),
-                               &symbolic_decomposition,
-                               control.data(),
-                               nullptr);
+  if (numbers::NumberTraits<number>::is_complex == false)
```
drwells: I'll approve this PR without this change (this is just a minor style request), but would you mind putting the positive check first here and below?

source/lac/sparse_direct.cc
```diff
@@ -402,6 +600,17 @@ SparseDirectUMFPACK::solve(Vector<double> &, const bool) const
 
 
 
+void
+SparseDirectUMFPACK::solve(Vector<std::complex<double>> &, const bool) const
+{
+  AssertThrow(
+    false,
+    ExcMessage(
+      "To call this function you need UMFPACK, but you configured deal.II without passing the necessary switch to 'cmake'. Please consult the installation instructions in doc/readme.html."));
```
drwells: Could we chop up this line to make it shorter?

source/lac/sparse_direct.cc
bangerth: Well, the way I thought about it is that this is actually the default case: The real-valued case. We just don't have a `NumberTraits::is_real` trait.

source/lac/sparse_direct.cc
bangerth: Sure.

bangerth:
I disabled all of the instantiations and tests if `DEAL_II_WITH_COMPLEX=OFF`.

bangerth:
I also verified that all of the tests continue to work with `DEAL_II_WITH_COMPLEX=OFF`.

drwells:
I think this is fine once we get rid of the unused boolean:
```
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9207/source/lac/sparse_direct.cc:385:39: error: unused parameter 'transpose' [-Werror,-Wunused-parameter]

                           const bool transpose /*=false*/) const
```

bangerth:
Fixed.

drwells:
There are some more unused booleans.

bangerth:
Well, let's see about now...

source/lac/sparse_direct.cc
```diff
@@ -326,6 +379,122 @@ SparseDirectUMFPACK::solve(Vector<double> &rhs_and_solution,
 }
 
 
+
+void
+SparseDirectUMFPACK::solve(Vector<std::complex<double>> &rhs_and_solution,
+                           const bool transpose /*=false*/) const
+{
+#  ifdef DEAL_II_WITH_COMPLEX
```
drwells:
```suggestion
#  ifdef DEAL_II_WITH_COMPLEX_VALUES
```

source/lac/sparse_direct.cc
```diff
@@ -326,6 +379,122 @@ SparseDirectUMFPACK::solve(Vector<double> &rhs_and_solution,
...
+    }
+
+#  else
+
+  (void)rhs_and_solution;
+  (void)transpose;
+  Assert(false,
+         ExcMessage(
+           "This function can't be called if deal.II has been configured "
+           "with DEAL_II_WITH_COMPLEX=FALSE."));
```
drwells:
```suggestion
           "with DEAL_II_WITH_COMPLEX_VALUES=OFF."));
```

source/lac/sparse_direct.cc
```diff
@@ -341,6 +510,31 @@ SparseDirectUMFPACK::solve(BlockVector<double> &rhs_and_solution,
 
 
 
+void
+SparseDirectUMFPACK::solve(BlockVector<std::complex<double>> &rhs_and_solution,
+                           const bool transpose /*=false*/) const
+{
+#  ifdef DEAL_II_WITH_COMPLEX
```
drwells: etc.

bangerth:
I can't seem to get this right, having to build two different configurations :-) Either way, let's see what happens now...

/rebuild

PR #9208, opened 2019-12-31
Fix our TBB version check.

drwells:
Since the regular expression we used was eager it matched the 202 in 2020, yielding a version number of 0.0. Note that this worked with older versions of TBB since we would get, e.g., version 8.0 instead of version 2018.0.

I am not sure why TBB released 2020 version software a few weeks ago, but here we are :)


cmake/modules/FindTBB.cmake
```diff
@@ -65,12 +65,12 @@ DEAL_II_FIND_PATH(TBB_INCLUDE_DIR tbb/tbb_stddef.h
 IF(EXISTS ${TBB_INCLUDE_DIR}/tbb/tbb_stddef.h)
   FILE(STRINGS "${TBB_INCLUDE_DIR}/tbb/tbb_stddef.h" TBB_VERSION_MAJOR_STRING
     REGEX "#define.*TBB_VERSION_MAJOR")
-  STRING(REGEX REPLACE "^.*TBB_VERSION_MAJOR.*([0-9]+).*" "\\1"
+  STRING(REGEX REPLACE "^.*TBB_VERSION_MAJOR *([0-9]+).*" "\\1"
```
bangerth: That's pre-existing, but I think we ought to expect there to be at least one space.
```suggestion
  STRING(REGEX REPLACE "^.*TBB_VERSION_MAJOR +([0-9]+).*" "\\1"
```
Same below.

drwells: Fixed!

tjhei:
> I am not sure why TBB released 2020 version software a few weeks ago, but here we are :)

I guess we had a y2k20 bug. :-)

PR #9209, opened 2019-12-31
Keep the 'dealii::' in 'using namespace dealii::'.

### LINKED ISSUE #9163: Typo on Step-40's webpage
I tried to run the code from Step-40's webpage,
https://www.dealii.org/current/doxygen/deal.II/step_40.html

However, I encountered errors: error: ‘LinearAlgebraPETSc’ is not a namespace-name
   using namespace ::LinearAlgebraPETSc;

So I compared the code from the locally installed file of Step-40 and found there is a slight difference:

Webpage:
```
namespace LA
{
#if defined(DEAL_II_WITH_PETSC) && !defined(DEAL_II_PETSC_WITH_COMPLEX) && \
!(defined(DEAL_II_WITH_TRILINOS) && defined(FORCE_USE_OF_TRILINOS))
using namespace ::LinearAlgebraPETSc;
#define USE_PETSC_LA
#elif defined(DEAL_II_WITH_TRILINOS)
using namespace ::LinearAlgebraTrilinos;
#else
#error DEAL_II_WITH_PETSC or DEAL_II_WITH_TRILINOS required
#endif
} // namespace LA
```

Locally installed source code:
```
namespace LA
{
#if defined(DEAL_II_WITH_PETSC) && !defined(DEAL_II_PETSC_WITH_COMPLEX) && \
  !(defined(DEAL_II_WITH_TRILINOS) && defined(FORCE_USE_OF_TRILINOS))
  using namespace dealii::LinearAlgebraPETSc;
#define USE_PETSC_LA
#elif defined(DEAL_II_WITH_TRILINOS)
  using namespace dealii::LinearAlgebraTrilinos;
#else
#error DEAL_II_WITH_PETSC or DEAL_II_WITH_TRILINOS required
#endif
} // namespace LA
```

The second one has namespace of "dealii", while the first doesn't.

After adding "dealii", the program from the webpage can be run successfully. Therefore, I think the webpage version should be updated to:
`using namespace dealii::LinearAlgebraPETSc`
instead of
`using namespace ::LinearAlgebraPETSc`

Thanks

Best regards,

Zhidong Brian Zhang

drwells:
Some of our example code on the website (e.g., step-40) doesn't work since we get rid of the preceding `dealii` and print

    using namespace ::LinearAlgebraPETSc;

when it should be

    using namespace dealii::LinearAlgebraPETSc;

instead. This change keeps the `dealii::` only in this case, which currently only occurs in step-40.cc, step-50.cc, step-55.cc, and local_integrators.h.

Fixes #9163.

### DIFF
--- a/doc/doxygen/scripts/filter
+++ b/doc/doxygen/scripts/filter
@@ -60,6 +60,11 @@ while (<>)
         s/::dealii::(.*)\\$/::\1        \\/g;
         s/dealii::(.*)\\$/\1        \\/g;
     }
+    elsif (m/using namespace dealii::/)
+    {
+        # namespace declarations (see, e.g., step-40) in source code don't work
+        # if we cut off the 'dealii::' part, so leave it in that case.
+    }
     else
     {
         s/(::)?dealii::/::/g;


drwells:
This does prevent doxygen from correctly forming a link but I think its better to show the right code than to have the right link.

PR #9210, opened 2019-12-31
Fix some doxygen namespace links.

drwells:
I noticed this while working on #9209.


drwells:
I didn't realize that #9195 will remove that namespace. Let me close this since a different fix is already in the pipeline.

peterrum:
@drwells Thanks! If you want you can join the discussion in https://github.com/dealii/dealii/pull/9195#pullrequestreview-337165896. We still have not agreed on the final name of the namespace.

PR #9215, opened 2020-01-02
Make DoFHandlerPolicy independent of p4est 

peterrum:
This PR makes `internal::DoFHandlerImplementation::Policy::ParallelDistributed` independent from `p4est`. The only reason why the dofhandler-policy was depending on `p4est` was because it used the data structure `dealii::internal::p4est::types<dim>::quadrant` to identify globally uniquely the cells (and some utility functions to compare ids). This goal can be also reached by using `CellId`.

This PR depends on PR #9214 (which adds `CellId::is_ancestor_of`).

Once this is merged, `p:f:t` will be working completely without `p4est` (see also PR #9106).

### DIFF
--- a/source/dofs/dof_handler_policy.cc
+++ b/source/dofs/dof_handler_policy.cc
@@ -4202,24 +4202,18 @@ namespace internal
 
 
       /* --------------------- class ParallelDistributed ---------------- */
-
-#ifdef DEAL_II_WITH_P4EST
-
       namespace
       {
         template <int dim, int spacedim>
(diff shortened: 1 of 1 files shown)


peterrum:
@kronbichler  Thanks for merging PR #9214. I have rebased this PR. I like the statistics: +27 and -71 locs...

peterrum:
@tjhei Any objections from your side?

bangerth:
Very nice work indeed!

PR #9216, opened 2020-01-02
Extend MatrixFree::get_dof_handler

Rombur:
This extends the current `MatrixFree::get_dof_handler()` function to `hp::DoFHandler`. I have made the function fit with the current interface but it's pretty ugly when you want the function to return an `hp::DoFHandler<dim>`. You get something like:
```c++
matrix_free.template get_dof_handler<hp::DoFHandler<dim>>();
```
Let me know now if you think that we should have a `get_hp_dof_handler()` function instead. I didn't do that because it would be the only function in the interface that would have a different name for the `hp` case.

### DIFF
--- a/doc/news/changes/minor/20200102BrunoTurcksin
+++ b/doc/news/changes/minor/20200102BrunoTurcksin
@@ -1,0 +1,4 @@
+New: The function MatrixFree::get_dof_handler() has been templated to enable
+to return either a DoFHandler<dim> or an hp::DoFHandler<dim>.
+<br>
+(Bruno Turcksin, 2020/01/02)
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -1520,9 +1520,14 @@ public:
 
   /**
    * Return the DoFHandler with the index as given to the respective
-   * `std::vector` argument in the reinit() function.
+   * `std::vector` argument in the reinit() function. Note that if you want to
+   * call this function with a template parameter different than the default
+   * one, you will need to use the `template` before the function call, i.e.,
+   * you will have something like `matrix_free.template
+   * get_dof_handler<hp::DoFHandler<dim>>()`.
    */
-  const DoFHandler<dim> &
+  template <typename DoFHandlerType = DoFHandler<dim>>
+  const DoFHandlerType &
   get_dof_handler(const unsigned int dof_handler_index = 0) const;
 
   /**
--- a/include/deal.II/matrix_free/matrix_free.templates.h
+++ b/include/deal.II/matrix_free/matrix_free.templates.h
@@ -72,6 +72,7 @@ namespace internal
     const unsigned int n_dof_handlers,
     const unsigned int dof_handler_index)
   {
+    (void)n_dof_handlers;
     AssertDimension(dof_handler.size(), n_dof_handlers);
     return *dof_handler[dof_handler_index];
   }
@@ -92,6 +93,7 @@ namespace internal
     const unsigned int n_dof_handlers,
     const unsigned int dof_handler_index)
   {
+    (void)n_dof_handlers;
     AssertDimension(hp_dof_handler.size(), n_dof_handlers);
     return *hp_dof_handler[dof_handler_index];
   }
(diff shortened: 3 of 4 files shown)


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -1500,9 +1500,13 @@ class MatrixFree : public Subscriptor
 
   /**
    * Return the DoFHandler with the index as given to the respective
-   * `std::vector` argument in the reinit() function.
+   * `std::vector` argument in the reinit() function. Note that if you want to
+   * call this function with a template parameter different than the default
+   * one, you will need to use the `template` before the function call, i.e.,
+   * you will have something like `matrix_free.template<hp::DoFHandler<dim>>()`.
```
peterrum:
```suggestion
   * you will have something like `matrix_free.template get_dof_handler<hp::DoFHandler<dim>>()`.
```

Rombur: Good catch, thanks

marcfehling:
That's nice of you to ask and to remember! :-)

Indeed, my plan was to add a `DoFHandlerType` template to the `MatrixFree` class. This way we could easily generalize both `get_dof_handler` and `get_cell_iterator` functions. If, somehow, we need to distinguish between `DoFHandler` and `hp::DoFHandler`, there is a static flag `is_hp_dof_handler` available at compile time.

I still prefer the `DoFHandlerType` template variant, since I can not think of any scenario where we need both `get_dof_handler` and `get_hp_dof_handler` functions simultaneously. This way, we would also refrain from duplicating code.

What do you think?

doc/news/changes/minor/20200102BrunoTurcksin
```diff
@@ -0,0 +1,4 @@
+New: Template the function MatrixFree::get_dof_handler() to return either a
+DoFHandler<dim> or an hp::DoFHandler<dim>
```
kronbichler: Maybe write:
```
New: The function MatrixFree::get_dof_handler() has been templated to enable
to return either a DoFHandler<dim> or an hp::DoFHandler<dim>.
```

peterrum: @Rombur I am curious in which context are you using this function? Either you are using `MatrixFree` as a container so that you do not have to pass the dof-handler explicitly as argument  to functions or you are accessing some functions for the dof-handler which are not provided by `MatrixFree`. In latter case, it might be an option - independent from this PR - to add that functionality...

peterrum: @kronbichler In a small side project, I am having a look into the hp-implementations. I'll keep you posted how user-friendly `MatrixFree` is for a matrix-free operator (once setup with a regular `DoFHandler` and once with `hp::DoFHandler`).

Rombur: @peterrum I am using it as a "container". I don't want to have to keep track of the `DoFHandler` everywhere.

peterrum: @Rombur If you add @kronbichler's suggestion, we can merge.

Rombur:
I have updated the changelog

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -53,6 +53,50 @@ DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
...
+      SmartPointer<const ::dealii::DoFHandler<DoFHandlerType::dimension>>>
+      &dof_handler,
+    const std::vector<
+      SmartPointer<const ::dealii::hp::DoFHandler<DoFHandlerType::dimension>>>
+      &,
+    const unsigned int n_dof_handlers,
+    const unsigned int dof_handler_index)
+  {
+    AssertDimension(dof_handler.size(), n_dof_handlers);
+    return *dof_handler[dof_handler_index];
```
masterleinad: It looks like `n_dof_handlers` is unused in `Release` mode. The same below.

tamiko:
@masterleinad I have pushed a quick fix for the release mode warning. Let's wait for the tester to pass and merge afterwards.

PR #9220, opened 2020-01-04
Test python bindings

masterleinad:
This pull request makes sure that the tests for the python bindings are run when calling `ctest` in the build directory.
This required moving some code from `python-bindings/source/CMakeLists.txt` to `python-bindings/CMakeLists.txt` which is in line with how we split `CMake` code between `CMakeLists.txt` and `source/CMakeLists.txt`.

### DIFF
--- a/contrib/python-bindings/tests/CMakeLists.txt
+++ b/contrib/python-bindings/tests/CMakeLists.txt
@@ -1,0 +1,30 @@
+## ---------------------------------------------------------------------
+##
+## Copyright (C) 2020 by the deal.II authors
+##
+## This file is part of the deal.II library.
+##
+## The deal.II library is free software; you can use it, redistribute
+## it, and/or modify it under the terms of the GNU Lesser General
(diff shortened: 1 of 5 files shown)


contrib/python-bindings/tests/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+## ---------------------------------------------------------------------
```
bangerth: My preference would be to put the python tests into `tests/python-bindings` to make them first-class citizens along with the other tests.

masterleinad:
> But have we reached a state where the python binds should be considered first-class citizens of the deal.II universe?

Let's discuss/address that somewhere else. I would be happier moving them.

masterleinad:
Can we move forward with this?

masterleinad:
I was thinking about doing that in a second step.

tamiko:
@masterleinad Well, then go ahead!

PR #9228, opened 2020-01-05
fixes a typo in step-3 tutorial doc (convergence topic in results/extensions)

krishnakumarg1984:

PR #9229, opened 2020-01-05
Return error code from indent-all in check_indentation.sh

masterleinad:
Travis reports success in https://github.com/dealii/dealii/pull/9228 although the author information requirements are not met. We just need to forward the exit code in case of failure to fix this.

### DIFF
--- a/contrib/utilities/check_indentation.sh
+++ b/contrib/utilities/check_indentation.sh
@@ -34,6 +34,6 @@ else
 	echo "Running indentation test on Pull Request #${TRAVIS_PULL_REQUEST}"
 fi
 
-./contrib/utilities/indent-all
+./contrib/utilities/indent-all || exit $?
 git diff
 git diff-files --quiet 


rezarastak:
It seems that your change ignores the error code returned by `git diff-files --quiet`

contrib/utilities/check_indentation.sh
```diff
@@ -35,5 +35,7 @@ else
 fi
 
 ./contrib/utilities/indent-all
+retval=$?
```
tjhei: how about you do ``./contrib/utilities/indent-all || exit 1``

masterleinad: That was my initial commit but I was concerned that we didn't return the diff. Realizing that we don't indent when the author information is off, this seems to be OK again.

masterleinad:
Fixed. That should work now.

tjhei:
why do you use return instead of exit? Isn't that usually used for returning from functions? I assume you checked that return works here?

masterleinad:
> I assume you checked that return works here?

Yes, I checked. I changed to `exit` for consistency now.

PR #9231, opened 2020-01-05
Simpler CellwiseInverseMassMatrix::apply() function

kronbichler:
The new function allows to use the cellwise inverse mass matrix without providing an explicit `inverse` field, using the implicit definition of `FEEvaluationBase::JxW()`. This leads to simpler code, even though it is less general than the old one that also allowed to add a coefficient from the user code.

This is the function I promised in https://github.com/dealii/dealii/pull/9157#issuecomment-568302128 and that I will add to the step-67 tutorial program in #8319. I will re-base #8319 once this is merged; the comments there are already adapted to this new function.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -3043,6 +3043,58 @@ namespace internal
   template <int dim, int fe_degree, int n_components, typename Number>
   struct CellwiseInverseMassMatrixImpl
   {
+    template <typename FEEvaluationType>
+    static void
+    apply(const FEEvaluationType &fe_eval,
+          const Number *          in_array,
+          Number *                out_array)
+    {
+      constexpr unsigned int dofs_per_component =
+        Utilities::pow(fe_degree + 1, dim);
+
+      Assert(dim >= 1 || dim <= 3, ExcNotImplemented());
+      Assert(fe_eval.get_shape_info().element_type <=
+               MatrixFreeFunctions::tensor_symmetric,
+             ExcNotImplemented());
+
+      internal::EvaluatorTensorProduct<internal::evaluate_evenodd,
+                                       dim,
+                                       fe_degree + 1,
+                                       fe_degree + 1,
+                                       Number>
+        evaluator(AlignedVector<Number>(),
+                  AlignedVector<Number>(),
+                  fe_eval.get_shape_info().inverse_shape_values_eo);
+
+      for (unsigned int d = 0; d < n_components; ++d)
+        {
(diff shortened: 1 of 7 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -3043,6 +3043,58 @@ namespace internal
...
+          // assume symmetries that do not exist in the inverse shapes
+          evaluator.template hessians<0, true, false>(in, out);
+          if (dim > 1)
+            evaluator.template hessians<1, true, false>(out, out);
+          if (dim > 2)
+            evaluator.template hessians<2, true, false>(out, out);
+        }
+      for (unsigned int q = 0; q < dofs_per_component; ++q)
+        {
+          const Number inverse_JxW_q = Number(1.) / fe_eval.JxW(q);
```
peterrum: From step-67:
> Another trick we apply is a separate variable for the inverse density $\frac{1}{\rho}$. This enables the compiler to only perform a single division for the flux, despite the division being used at several places. As divisions are around ten to twenty times as expensive as multiplications or additions, avoiding redundant divisions is crucial for performance.

Couldn't we add a method `fe_eval.inverted_JxW(q)` (we could configure the pre-computation via `AdditionalData`)? Or do you think that loading these inverted values are more expensive than inverting the non-inverted values (since these are probably in the cache - especially in the ECL case).

kronbichler: Yes, my idea was that the `JxW` is (maybe) in the cache (depends of course on the setup, it is not for what we show in step-67 at this stage). Having a separate variable that stores a pre-computed inverse JxW would require the interface in `FEEvaluationBase` and the storage in `MappingInfo`, so it is a bit of work for this niche topic. Add to this the fact that if you want to use compression on affine/Cartesian cells, we would also need to store the inverse of the quadrature weight to make an effect. Finally, I would point out that on typical hardware, we are a factor of 2 within the memory-bound regime by just accessing the vectors, so oftentimes the cost of the division should be hidden behind the memory transfer. Doing one division in vectorized format has a reciprocal throughput of 2 cycles/double (see e.g. Agner Fog's instruction tables and divide by the vectorization width), whereas accessing one byte from RAM is around 8 cycles/double, so there is around a factor of 2 of slack, assuming everything else is perfect.

peterrum: OK. Thanks for the info!

kronbichler:
*ping*

peterrum:
@masterleinad This PR is fine to be merged! Bulk of this PR are new tests.

peterrum:
Thanks @masterleinad!

Issue #9226, opened 2020-01-05
[documentation] Lack of clarity on simplex() function in the GridGenerator namespace


krishnakumarg1984:
I am a bit confused by the documentation of the `simplex()` function in the `GridGenerator` namespace [here](https://dealii.org/developer/doxygen/deal.II/namespaceGridGenerator.html#acea0cbcd68e52ce8113d1134b87de403).

* First, aren't the meshes handled by deal.ii all quadrilaterals in 2D?  I do not understand the triangles drawn in this function's documentation.

* Secondly, there is some really confusing wording for the function's description.  Improving this shall be very helpful to novice users of deal.II (I landed here following the "suggested steps" in the 'results' section of Step-3 tutorial). 

* Finally, there is some odd choice of words
>And I am not happy about the discrimination involved here.

Who is "I" here, and what "discrimination" are we talking about?

masterleinad:
> * First, aren't the meshes handled by deal.ii all quadrilaterals in 2D?  I do not understand the triangles drawn in this function's documentation.

This function meshes a simplex by introducing additional vertices at the face centers, line centers, and the midpoint.

> * Secondly, there is some really confusing wording for the function's description.  Improving this shall be very helpful to novice users of deal.II (I landed here following the "suggested steps" in the 'results' section of Step-3 tutorial).

We are happy to take any suggestions for improvements as pull requests.

>  * Finally, there is some odd choice of words 
> > And I am not happy about the discrimination involved here.

This can likely just be deleted.

krishnakumarg1984:
>This function meshes a simplex by introducing additional vertices at the face centers, line centers, and the midpoint.

I had forgotten my basics and had to look up `simplex` on [Wikipedia](https://en.wikipedia.org/wiki/Simplex)! According to Wikipedia,

>In geometry, a simplex (plural: simplexes or simplices) is a generalization of the notion of a triangle or tetrahedron to arbitrary dimensions.
For example,
    a 0-simplex is a point,
    a 1-simplex is a line segment,
    a 2-simplex is a triangle,
    a 3-simplex is a tetrahedron,
    a 4-simplex is a 5-cell.

Well, the 2-simplex  and 3-simplex cases seem to be in direct contrast with the [FAQ entry here](https://github.com/dealii/dealii/wiki/Frequently-Asked-Questions#can-i-useimplement-trianglestetrahedra-in-dealii) which says

>_Can I use/implement triangles/tetrahedra in deal.II?_
[...]

PR #9233, opened 2020-01-05
Remove cryptic comment

masterleinad:
Fixes https://github.com/dealii/dealii/issues/9226.


bangerth:
By the way, the comment expressed a sentiment about treating right-handed people with preference over left-handed people.

PR #9235, opened 2020-01-06
Augment the UMFPACK error message.

bangerth:
In response to a recent thread on the mailing list.


include/deal.II/lac/sparse_direct.h
```diff
@@ -304,9 +304,12 @@ class SparseDirectUMFPACK : public Subscriptor
         "want to check your assembly procedure. Similarly, a "
         "matrix can be rank deficient if you forgot to apply the "
         "appropriate boundary conditions. For example, the "
-        "Laplace equation without boundary conditions has a "
-        "single zero eigenvalue and its rank is therefore "
-        "deficient by one."
+        "Laplace equation for a problem where only Neumann boundary "
+        "conditions are posed (or where you forget to apply Dirichlet "
+        "boundary conditions) has a one zero eigenvalue and its rank is "
```
drwells:
```suggestion
        "boundary conditions) has exactly one eigenvalue equal to zero and its rank is "
```
"has a one zero" doesn't look right and I think its clearer to not write "one zero".

bangerth:
@drwells -- yes, good catch!

drwells:
This is quite picky, but are you willing to reflow the string literals so that they are all about the same length? If not then lets merge.

bangerth:
Sure, if it makes you happy I will do that :-) Done!

PR #9240, opened 2020-01-07
Turn all for loops using "begin_active" into ranged-based ones

masterleinad:


bangerth:
Actually, `grid/grid_out_svg_02.debug` fails.

source/grid/grid_out.cc
```diff
@@ -1398,10 +1383,7 @@ GridOut::write_xfig(const Triangulation<2> &tria,
   // on finer levels. Level 0
   // corresponds to a depth of 900,
   // each level subtracting 1
-  Triangulation<dim, spacedim>::cell_iterator       cell = tria.begin();
-  const Triangulation<dim, spacedim>::cell_iterator end  = tria.end();
-
-  for (; cell != end; ++cell)
+  for (const auto &cell : tria.active_cell_iterators())
```
bangerth: Here's your bug. You should loop over all, not just the active cells.

masterleinad: Thanks!

source/multigrid/mg_tools.cc
```diff
@@ -138,7 +135,7 @@ namespace MGTools
     // unknown number of cells may
     // share an object of dimension
     // smaller than dim-1.
-    for (cell = dofs.begin(level); cell != end; ++cell)
+    for (const auto &cell : dofs.active_cell_iterators_on_level(level))
```
rezarastak:
```suggestion
    for (const auto &cell : dofs.cell_iterators_on_level(level))
```

source/multigrid/mg_tools.cc
```diff
@@ -332,7 +326,7 @@ namespace MGTools
     // unknown number of cells may
     // share an object of dimension
     // smaller than dim-1.
-    for (cell = dofs.begin_active(); cell != end; ++cell)
+    for (const auto &cell : dofs.active_cell_iterators_on_level(level))
```
rezarastak: The original code is confusing, the for loop starts at `dofs.begin_active()` but ends at `dofs.end(level)`.

tjhei: I am not sure that the old code is correct. I would go from begin(level) to end(level).

masterleinad: Hmm... I am not sure either, but for me it sounds more reasonable to loop over active cells on each level if we want to find out about the number of coupling degrees of freedom on a particular level. Unfortunately, there is no test for this function.

tjhei: We are asking for ``get_mg_dof_indices``, so I don't see a reason for looking at active cells only.

masterleinad: Thinking some more about it, I guess you're right. I changed the pull request accordingly.

Issue #7228, opened 2018-09-26
Thread cleanup


tamiko:
With native language support for threads we could think about

- [x] <s>replacing `ThreadLocalStorage` by `thread_local` qualified objects in the library and deprecate the class</s> (this doesn't work)
- [x] replace our `Mutex` and `ScopedLock` by `std::mutex` and `std::lock_guard`/`std::unique_lock` in the library and deprecate both classes
- [x] same for `std::condition_variable` and `ConditionVariable`
- [x] deprecate `PosixThreadBarrier`
- [x] deprecate almost everything under `Thread::*`, we have `std::thread::get_id`, `std::thread::hardware_concurrency`, etc.
- [x] clean up the cmake thread detection (almost all of this is obsolete)

drwells:
We did not come to a conclusion regarding `std::mutex` vs `Mutex` earlier, since the former cannot be copied; see #7103 for the related discussion.

bangerth:
We don't use the `PosixThreadBarrier` class anywhere, and one can imagine that none of our users are doing so either -- so we could just deprecate it without any replacement. But then it's a useful class for which C++ doesn't appear to have a replacement -- or do you happen to know whether there is one?

masterleinad:
There is a stackoverflow question with various implementations https://stackoverflow.com/questions/24465533/implementing-boostbarrier-in-c11 and a TS https://en.cppreference.com/w/cpp/experimental/barrier.

drwells:
@bangerth I think that the timer class also uses POSIX functionality (for counting CPU time), so we are not completely rid of it if we just remove this class.

masterleinad:
We also have `posix_memalign` replaceable by `std::aligned_alloc` in C++17.

bangerth:
OK, but that'll be a while until we can use that unconditionally :-)

tamiko:
Almost there I would say :-)

bangerth:
I read up on `thread_local`, see https://en.cppreference.com/w/cpp/language/storage_duration . The problem is that C++11 specifies that it can only be used for `static` member variables, but we most frequently use it for non-`static` member variables. So we can't quite get around using the class we currently have and which is based on ` tbb::enumerable_thread_specific` (https://www.threadingbuildingblocks.org/docs/help/reference/thread_local_storage/enumerable_thread_specific_cls.html).

There is also this piece of code that wants to access not only the current thread's data, but also that of others:
```
std::stack<std::string> &
[...]

PR #9242, opened 2020-01-07
replace some instances of Threads::Mutex with std::mutex

rezarastak:


masterleinad:
> Nice work. I wonder whether we should just replace all occurrences of `Threads::Mutex` and deprecate the class.

See https://github.com/dealii/dealii/issues/7228. In particular, `std::mutex` is not copyable and we would need to find a way around that.

bangerth:
Ah yes, I remember now. But we should probably replace it where possible.

PR #9246, opened 2020-01-07
fixes a typo in function name. Replace make_grid_and_dofs() with just make_grid()

krishnakumarg1984:


bangerth:
It still fails with this error message:
```
Running indentation test on Pull Request #9246
invalid author 'Krishnakumar' without firstname and lastname
```
Do you have a first and last name? I recognize that we may have put cultural assumptions into the script.

examples/step-4/doc/intro.dox
```diff
@@ -90,7 +90,7 @@ The same can be done with member variables. Consider the following
 function, which might in turn call the above one:
 @code
   template <int dim>
-  void make_grid_and_dofs (Triangulation<dim> &triangulation)
+  void make_grid (Triangulation<dim> &triangulation)
```
bangerth: Actually, this isn't right. The text here reads like this:
```
The same can be done with member variables. Consider the following
function, which might in turn call the above one:
@code
  template <int dim>
  void make_grid_and_dofs (Triangulation<dim> &triangulation)
  {
    make_grid (triangulation);

    DoFHandler<dim> dof_handler(triangulation);
    ...
  };
@endcode
```
The function name should really be as originally stated.

krishnakumarg1984: Ah. I concur. We can just close this one, and not merge.

PR #9249, opened 2020-01-07
Replace Threads::mutex by std::mutex

masterleinad:
After #9242, this should replace all (replacable) occurences of `Threads::Mutex` by `std::mutex`.

### DIFF
--- a/include/deal.II/matrix_free/dof_info.templates.h
+++ b/include/deal.II/matrix_free/dof_info.templates.h
@@ -1442,11 +1442,11 @@ namespace internal
       static constexpr unsigned int bucket_size_threading = 256;
 
       void
-      compute_row_lengths(const unsigned int           begin,
-                          const unsigned int           end,
-                          const DoFInfo &              dof_info,
-                          std::vector<Threads::Mutex> &mutexes,
-                          std::vector<unsigned int> &  row_lengths)
+      compute_row_lengths(const unsigned int         begin,
+                          const unsigned int         end,
+                          const DoFInfo &            dof_info,
+                          std::vector<std::mutex> &  mutexes,
+                          std::vector<unsigned int> &row_lengths)
       {
         std::vector<unsigned int> scratch;
         const unsigned int n_components = dof_info.start_components.back();
@@ -1485,7 +1485,7 @@ namespace internal
                              const unsigned int               end,
                              const DoFInfo &                  dof_info,
                              const std::vector<unsigned int> &row_lengths,
-                             std::vector<Threads::Mutex> &    mutexes,
+                             std::vector<std::mutex> &        mutexes,
                              dealii::SparsityPattern &        connectivity_dof)
       {
         std::vector<unsigned int> scratch;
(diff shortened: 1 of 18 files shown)


include/deal.II/matrix_free/dof_info.templates.h
```diff
@@ -1568,9 +1568,9 @@ namespace internal
         ++n_rows;
 
       // first determine row lengths
-      std::vector<unsigned int>   row_lengths(n_rows);
-      std::vector<Threads::Mutex> mutexes(
-        n_rows / internal::bucket_size_threading + 1);
+      std::vector<unsigned int> row_lengths(n_rows);
+      std::vector<std::mutex> mutexes(n_rows / internal::bucket_size_threading +
+                                      1);
```
bangerth: I'm surprised this works, given that the mutex class is not copyable...

bangerth:
It currently fails with this message:
```


clang-tidy-6.0 -header-filter=/jenkins/worksp/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/stl_construct.h:75:38: error: call to implicitly-deleted copy constructor of 'dealii::TensorProductMatrixSymmetricSum<3, dealii::VectorizedArray<float, 4>, 9>' [clang-diagnostic-error]
```
I have no idea where this is created, but it shows one of the issues with the now deleted copy constructor.

include/deal.II/matrix_free/dof_info.templates.h
masterleinad: This call only requires a default constructor, see https://godbolt.org/z/wQevHQ.

masterleinad:
>  I guess that means that all of these classes become non-copyable. But they probably shouldn't be copied anyway. I wonder whether we should make their copy operators/copy constructors explicitly deleted to provide better error messages, though. What are your thoughts?

Yes, these classes become non-copyable (if they are not already). I am happy to discuss if this is an undesired requirement for some of these. If not, I am going forward and deleting the copy constructors explicitly. That's a good sugegstion.

include/deal.II/matrix_free/dof_info.templates.h
bangerth: I thought that `std::vector` had as one of its requirements that its template argument must be copyable. It may be that this particular call doesn't require it, though.

bangerth:
I'm unsure what we want with these classes. I suspect that someone has put them into a `std::vector` at some point. What do others think?

tjhei:
> What do others think?

- We already discourage/disable copying of heavy objects like matrices. 
- For something derived from Function<dim> is also doesn't make sense to copy or put into a std::vector (because one would want pointers to base instead).
- FiniteElements also don't need to be copyable, right?
Anything else affected by this change?

@masterleinad : ``fe/fe_move_02.debug: DIFF failed``

Rombur:
> We already discourage/disable copying of heavy objects like matrices.

I personally think that it was not our best idea. Vectors and matrices not being copyable are just annoying to deal with. I had to write workaround code in one of my project because I can't use the copy constructor.

bangerth:
On 1/9/20 7:08 AM, Bruno Turcksin wrote:
> I personally think that it was not our best idea. Vectors and matrices not
>  being copyable are just annoying to deal with. I had to write workaround
> code in one of my project because I can't use the copy constructor.

But you *can* copy vectors. It's really just matrices.

Rombur:
> But you *can* copy vectors. It's really just matrices.

Not with this patch. There is a mutex in `LinearAlgebra::distributed::Vector`.

bangerth:
I see. But shouldn't that be resolvable with a custom copy operator/constructor?

masterleinad:
I reverted all the changes for classes that don't already have a copy constructor.

PR #8774, opened 2019-09-17
Hdf5 documentation

jpthiele:
Addition to the possibilities for extensions of step-3
-Description of using the HDF5 library to output not only solution and grid, but also additional data    like the mean value and point value. 
- Description of using an R script to automatically plot grid, pseudocolor and convergence curves out of the resulting HDF5 files.

PR #9250, opened 2020-01-07
Update the discussion of HDF5 and R scripting in step-3.

bangerth:
This is a really nice introduction to what can be done by putting data into
hdf5 files and then postprocessing them in R. I'm really only updating
stylistic issues here.

This is a follow-up to #8774.


bangerth:
Ping?

drwells:
Looks good to me :+1:

Since its just a documentation fix lets merge without running the tester.

PR #9256, opened 2020-01-07
fixes a redundant paragraph in Step-4 tutorial (since vtk format was already introduced in step-3)

krishnakumarg1984:
Looks like in many of the tutorials, the old eps format code has been systematically replaced by vtk output code. Therefore, the paragraph introducing vtk is superfluous in step-4 (since it is already introduced in detail in step-3).

### DIFF
--- a/examples/step-4/step-4.cc
+++ b/examples/step-4/step-4.cc
@@ -469,14 +469,6 @@ void Step4<dim>::solve()
 // This function also does what the respective one did in step-3. No changes
 // here for dimension independence either.
 //
-// The only difference to the previous example is that we want to write output
-// in VTK format, rather than for gnuplot. VTK format is currently the most
-// widely used one and is supported by a number of visualization programs such
-// as Visit and Paraview (for ways to obtain these programs see the ReadMe
-// file of deal.II). To write data in this format, we simply replace the
-// <code>data_out.write_gnuplot</code> call by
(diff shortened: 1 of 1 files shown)


bangerth:
Yes, you have the history right. Thanks for the fix!

krishnakumarg1984:
Thank you for approving and your general support. I found one or two small instances talking about vtk/paraview. I did not remove these in my PR, since I thought it may be helpful to have a quick reminder to new users, and they are not long.

Also, I found duplicate references to the visualization lectures 11 & 32 in both step-3 and step-4. I did not delete these also especially because the video lectures are super useful, and the links to them from each of the two tutorials can turn out to be handy (rather than redundant).

bangerth:
On 1/7/20 5:19 PM, Krishnakumar Gopalakrishnan wrote:
> Thank you for approving and your general support. I found one or two 
> small instances talking about vtk/paraview. I did not remove these in my 
> PR, since I thought it may be helpful to have a quick reminder to new 
> users, and they are not long.
> 
> Also, I found duplicate references to the visualization lectures 11 & 32 
> in both step-3 and step-4. I did not delete these also especially 
> because the video lectures are super useful, and the links to them from 
> each of the two tutorials can turn out to be handy (rather than redundant).

Yes, I think these are all good decisions.

bangerth:
Nice, I'm glad we figure the name thing out!

PR #9258, opened 2020-01-08
shape_3rd_derivative* checks for update_3rd_derivatives

simonsticko:
The functions shape_3rd_derivative and shape_3rd_derivative_component
check that update_flags contains update_hessians, but they should check
that update_flags contain update_3rd_derivatives. Fix this.


tjhei:
yes, that makes a lot more sense. /rebuild

PR #9261, opened 2020-01-08
Fix two typos in documentation of FEInterfaceValues.

simonsticko:


include/deal.II/fe/fe_interface_values.h
```diff
@@ -96,9 +96,9 @@ class FEInterfaceValues
    * @param[in] face_no An integer identifying which face of the first cell the
    *   interface is on.
    * @param[in] sub_face_no An integer identifying the subface (child) of the
-   *   face identified by the previous two arguments the interface corresponds
-   *   to. If equal to numbers::invalid_unsigned_int, then the interface is
-   *   considered to be the entire face.
+   *   face identified by the previous two arguments. The interface corresponds
```
tjhei: I think the original sentence is a bit awkward but correct. Maybe:
```suggestion
   *   face (identified by the previous two arguments) the interface corresponds
```

simonsticko: Okey. Then I would like to add a "that":
"An integer identifying the subface (child) of the face (identified by the previous two arguments) **that** the interface corresponds to."
but it might not be needed. I can't really tell.

tjhei:
great, thanks. Documentation so I will merge...

PR #9263, opened 2020-01-08
ParameterHandler in unsorted order

elauksap:
Dear community,

the `ParameterHandler` class currently prints the parameters transversing the Boost property tree using **ordered** iterators by default (probably for compatibility with a previous version, as commented in the class implementation).

This behavior seemed counterintuitive to me, so I'm opening this pull request that updates the class using **unordered** iterators, so that sections and parameters are printed in the same order as they are declared.

Best regards,
Pasquale


tjhei:
Interesting, I didn't know that property_tree would store the elements in original order. While I like this change (it was annoying to me before), I wonder if we break some workflow (or at least some tests) with this.

rezarastak:
I like the idea of printing the parameters in the order they were declared. Usually when I like the parameter `beta` be printed before the parameter `alpha`, I declare them as `1-beta` and `2-alpha`. With this PR, I don't have to do that anymore.

I also suspect that it will create backward compatibility issues. Ideally, we can have an optional flag in the constructor of `ParameterHanlder` that controls the ordering of parameters, but it will make the code unnecessarily complex.

blaisb:
I also like this idea a lot to be honest. This would really clean our parameter files templates.

elauksap:
Thanks, I'm glad that you find this PR useful!

I'm not really sure about backward compatibility. I had already thought of adding an optional flag to control the way of sorting, but since there is no common interface between Boost `assoc_iterator`s and standard `iterator`s there would necessarily be a lot of duplicate code for the two versions (unless using generic lambdas or similar, which would be overly complicated for such a feature in my opinion).

Anyhow, if you are really concerned about backward compatibility issues I can manage to implement such optional flag!

tjhei:
> Anyhow, if you are concerned about backward compatibility issues I can manage to implement such optional flag!

Let's wait and see if there are strong opinions in favor of having this option first. I don't mind this change, for example.

elauksap:
> Let's wait and see if there are strong opinions in favor of having this option first. I don't mind this change, for example.

Ok then! I'll be waiting for further reviews, thank you! 👍

luca-heltai:
I don't see why this should have any backward compatibility issues. Why would the order in which parameters are parsed matter? The essential part is that, if two lines with the same parameter are present in a file, the last one wins. There will be of course differences in all tests that output parameters, but this is to be expected with such a change.

tjhei:
> Why would the order in which parameters are parsed matter?

That should not be a problem, yes.

> There will be of course differences in all tests that output parameters, but this is to be expected with such a change.

Agreed and this might affect projects using deal.II as well if they test the output.

Anyways, I think we should go ahead with this change.

tjhei:
@elauksap this change will require you to update expected test output for many tests. Let us know if you need help with that.

elauksap:
> 
> 
> @elauksap this change will require you to update expected test output for many tests. Let us know if you need help with that.

Thanks for approving the PR!

I guess all the tests that need to be modified can be found with a `grep -rl -e "print_parameters" -e "log_parameters" tests/`, right?

tjhei:
> right?

I would think so. I would start by running ``ctest -R parameter`` and update the output.

I would then fire of a complete test run for all tests to be sure. I can do that if you are done with fixing the obvious files.

elauksap:
Ok thanks, I will let you know as soon as I've updated the obvious files!

elauksap:
Hi @tjhei, I think it's done!

I've modified the output of the tests found with `ctest -R "parameter"` and with `grep -rl -e "print_parameters" -e "log_parameters" tests/`. All those tests succeed on my system now.

Could you please run the complete test pipeline?

source/base/parameter_handler.cc
```diff
@@ -1164,12 +1164,11 @@ ParameterHandler::recursively_print_parameters(
...
-          // print entries one by one. make sure they are sorted by using
-          // the appropriate iterators
+          // print entries one by one
           bool first_entry = true;
-          for (boost::property_tree::ptree::const_assoc_iterator p =
-                 current_section.ordered_begin();
-               p != current_section.not_found();
+          for (boost::property_tree::ptree::const_iterator p =
+                 current_section.begin();
+               p != current_section.end();
```
masterleinad: Can't you use `range-based` for loops in all these places?

elauksap: You are absolutely right, I've updated the modifications!

tjhei:
By looking at the changed test output I notice two things:
1. The log_parameters() function is probably only used in tests. Maybe it is not worth changing as it just changes output?
2. I did not think about writing latex and the order of output there. We use these in ASPECT for the manual. @gassmoeller  @bangerth - how do you feel about the order changing from alphabetical to as declared?

elauksap:
>     1. The log_parameters() function is probably only used in tests. Maybe it is not worth changing as it just changes output?

Yes, you are right. I decided to modify `log_parameters()` too just for consistency with the other methods in the class!

gassmoeller:
I am not very thrilled by this change for the following reasons:

- in ASPECT we have by now >950 input parameters sorted in many sections. The alphabetical order gives us a nice unique way to sort them in our manual. If we switch to as-declared we need to adapt our sorting in quite a number of places, otherwise we end up with a wild order that requires everyone to search for known parameters all the time.
- The order we declare our parameters in is not easy to understand. E.g. we have signals that allow users to declare additional parameters in sections that have already been declared before. Do we now see these sections twice in the output, or will later parameters appended to existing subsections in the place the sections where first declared? Another example is that the order of our declaration functions depends on the order in which their plugins are registered, which depends on when a static function is called, which happens at the end of its source file in a global scope. So does that mean the order of our parameters will depend on the order in which the build system processes the source files? Is that unique between different systems or will a Mac now build a different manual than a linux than a windows pc? We could probably work around that by sorting our plugins alphabetically internally, but it would require adjustments to all our plugin systems.

That being said I understand that this is a real improvement for many other people so I would be willing to adjust accordingly if necessary. But an alphabetical option would make this task a lot easier :-)

tjhei:
> * Do we now see these sections twice in the output, or will later parameters appended to existing subsections in the place the sections where first declared?

No, they would output in the order they first appear.

> * Another example is that the order of our declaration functions depends on the order in which their plugins are registered, which depends on when a static function is called, which happens at the end of its source file in a global scope.

I did not think of that. Yes, this is a problem.

elauksap:
Would having a `print_parameters_alphabetical` method (or an input flag to `print_parameters`) be an easier solution to you?

bangerth:
On 1/13/20 6:39 AM, Timo Heister wrote:
> #
> # I did not think about writing latex and the order of output there. We use 
> these in ASPECT for the manual. @gassmoeller <https://github.com/gassmoeller> 
> @bangerth <https://github.com/bangerth> - how do you feel about the order 
> changing from alphabetical to as declared?

Also the graphical front-end that lets you pick parameters in a GUI.

I don't particularly care one way or the other. I see @gassmoeller's point 
about being able to find parameters more easily. At the same time, presumably 
the person who wrote the various declarations put them into an order that made 
sense to them, and maybe that's worth something as well.

My understanding is that the only thing that is affected is the order *within 
a section*. If you go out of the section and back in later on, the parameters 
added the second time around are still in that section and will be printed in 
the same section, just after all of the other parameters.

tjhei:
> My understanding is that the only thing that is affected is the order *within a section*

and the order of the sections themselves (that could be random in ASPECT as pointed out by Rene)?

elauksap:
Right, both the order of the sections and of the parameters within each section are affected.

gassmoeller:
Hmm, so it would be an inconvenience for us, but not the end of the world. A flag to switch between the orderings would be nice, but if it entails duplicating more than 50-100 lines then feel free to just change the order (since I got the impression the majority seemed to favor the as-declared order). Thanks for listening to my concerns :-)

elauksap:
I've estimated that in the worst case about 540 lines of codes would be duplicated if using a flag to switch between the two orderings.

tjhei:
@elauksap I just realized that there is a ``sort()`` function that "Sorts the direct children of this node according to key order.". Maybe we can add a flag to the ``ParameterHandler`` (or individual print functions) that will ``sort()`` all entries recursively if requested? Can you try if this works?

elauksap:
Hi @tjhei, thanks for your suggestion!
I don't know if you like my solution, but I have implemented a public method `sort_parameters()` that optionally sorts all parameters and subsections whenever the users wants to. Its equivalent `sort_parameters_section()` does the same only for the current subsection.

Both methods invoke a private `recursively_sort_parameters()` method which calls `boost::property_tree::ptree::sort()` recursively.

I think this solution is a bit more versatile than an optional input flag, since a user can decide whether to sort:
- all the parameters;
- all the parameters declared so far but not those that are still to be declared;
- all the parameters belonging to a specified subsection but not others.

Do you think it can fit?
If this solution gets your approval I will implement a test for all the new methods.

tjhei:
> Do you think it can fit?
> If this solution gets your approval I will implement a test for all the new methods.

I would prefer a) adding a flag ``bool sort_alphabetical=true`` to print_*, or b) a property of ParameterHandler (either in constructor or by calling a function over your suggestion. The reason is quite simple: with your approach, you can not be backwards-compatible with the way it works currently. Or am I wrong?

elauksap:
You are right, my solution was more versatile but not backward compatible. Now I've added an optional boolean flag to `print_parameters()` and `log_parameters()` methods.

All test outputs have been restored to their original state, and I've added an additional test with the newly generated output.

tjhei:
> You are right, my solution was more versatile but not backward compatible.

yeah, I think you are experiencing that even a tiny change is difficult to do without breaking somebodies workflow.

include/deal.II/base/parameter_handler.h
```diff
@@ -1423,9 +1431,15 @@ class ParameterHandler : public Subscriptor
...
+   *
+   * Before printing, all current parameters and subsections are sorted
+   * alphabetically by default.
+   * This option can be disabled setting the last parameter @p sort_alphabetical
+   * to <kbd>false</kbd>: in this case entries are printed in the same order
+   * as they have been declared.
    */
   void
-  log_parameters(LogStream &out);
+  log_parameters(LogStream &out, const bool &sort_alphabetical = true);
```
tjhei: Do you need this functionality in log_parameters?

include/deal.II/base/parameter_handler.h
```diff
@@ -1393,7 +1399,9 @@ class ParameterHandler : public Subscriptor
    * @endcode
    */
   std::ostream &
-  print_parameters(std::ostream &out, const OutputStyle style) const;
+  print_parameters(std::ostream &    out,
+                   const OutputStyle style,
+                   const bool &      sort_alphabetical = true) const;
```
tjhei: one thing I just realized, is that this is also not intuitive. Imagine me doing:
```
print_parameters(...,...,true);
print_parameters(...,...,false);
```
Notice that print parameters now modifies the state of the object and the second call still keeps things sorted? Not sure how to best deal with this. Maybe it needs to be a class property after all.

include/deal.II/base/parameter_handler.h
elauksap: The only reason to have it there is backward compatibility!

include/deal.II/base/parameter_handler.h
elauksap: Yes, the object state is modified after the first call to `print_parameters`.

I don't know how to make it work even if making it a class property! Where would the sort happen? Whenever a parameter is declared?

elauksap:
> yeah, I think you are experiencing that even a tiny change is difficult to do without breaking somebodies workflow.

Exactly, I'm experiencing that even a small change breaks something elsewhere :sweat_smile:

include/deal.II/base/parameter_handler.h
bangerth: Just as a minor comment on top of all of the other things: We generally don't mark scalar arguments (like `bool`) as references.

elauksap:
@tjhei I've overcome the issue by creating (if needed) a backup of the entries tree before sorting and by restoring it by the end of the `print_*` methods. In this way the class state is never modified. What do you think?

@bangerth Thanks, I've removed the reference :wink:

source/base/parameter_handler.cc
```diff
@@ -1040,14 +1037,52 @@ ParameterHandler::set(const std::string &entry_string, const bool new_value)
...
   AssertThrow(out, ExcIO());
 
+  // Save backup of entries before sorting, so that
+  // it can be restored by the end of this method.
+  boost::property_tree::ptree entries_backup;
+
+  // Sort parameters alphabetically, if needed.
+  if (sort_alphabetical)
+    {
+      entries_backup = (*entries);
```
tjhei: Can you sort the backup and print the backup instead? This way you are not modifying the data structures stored at all.

elauksap: Done @tjhei! I've modified a few methods to make them modify the backup copy instead of the class state.

elauksap:
Hi @tjhei! I think we have reached a stable version (finally all tests pass), please let me know if other modifications are needed! Thanks :smile:

luca-heltai:
I'll let @bangerth comment, but otherwise I think this is ready to merge.

include/deal.II/base/parameter_handler.h
```diff
@@ -1332,10 +1333,15 @@ class ParameterHandler : public Subscriptor
...
   set(const std::string &entry_name, const bool new_value);
 
-
   /**
    * Print all parameters with the given style to <tt>out</tt>.
    *
+   * Before printing, all current parameters and subsections are printed sorted
+   * alphabetically by default.
+   * This option can be disabled setting the last parameter @p sort_alphabetical
+   * to <kbd>false</kbd>: in this case entries are printed in the same order
```
tjhei: what is <kbd>? Can you use ``@p false`` instead?

include/deal.II/base/parameter_handler.h
```diff
@@ -1406,6 +1414,12 @@ class ParameterHandler : public Subscriptor
    * to get a valid XML document and output starts with one root element
    * <tt>ParameterHandler</tt>.
    *
+   * Before printing, all parameters and subsections of the present subsection
+   * are printed sorted alphabetically by default.
+   * This option can be disabled setting the last parameter @p sort_alphabetical
+   * to <kbd>false</kbd>: in this case entries are printed in the same order
```
tjhei: same here and several places below.

include/deal.II/base/parameter_handler.h
elauksap: It is an alias for the deprecated `<tt>`, but it should be rendered likewise.
However it is fixed now!

tjhei: Interesting, let's hope <tt> will not disappear any time soon, considering we are using it all over the place.

tjhei:
One more question: why did the test output for json/xml tests change?

elauksap:
> One more question: why did the test output for json/xml tests change?

The reason is that their output is written by Boost `write_{json,xml}()` functions. In other output formats we are assumming that subsections are printed after parameters declared in the present section, which apparently is not the case for the Boost builtin functions.

tjhei:
> The reason is that their output is written by Boost `write_{json,xml}()` functions. In other output formats we are assumming that subsections are printed after parameters declared in the present section, which apparently is not the case for the Boost builtin functions.

What explains a change in the output order though? I was not expecting the normal output to change. Is this because you are making a copy of the property_tree now?

include/deal.II/base/parameter_handler.h
```diff
@@ -1332,10 +1333,15 @@ class ParameterHandler : public Subscriptor
   void
   set(const std::string &entry_name, const bool new_value);
 
-
   /**
    * Print all parameters with the given style to <tt>out</tt>.
    *
+   * Before printing, all current parameters and subsections are printed sorted
```
bangerth:
```suggestion
   * Before printing, all current parameters and subsections are sorted
```

include/deal.II/base/parameter_handler.h
```diff
@@ -1332,10 +1333,15 @@ class ParameterHandler : public Subscriptor
...
+   * alphabetically by default.
+   * This option can be disabled setting the last parameter @p sort_alphabetical
```
bangerth:
```suggestion
   * This behavior can be disabled setting the last parameter @p sort_alphabetical
```

include/deal.II/base/parameter_handler.h
```diff
@@ -1406,6 +1414,12 @@ class ParameterHandler : public Subscriptor
    * to get a valid XML document and output starts with one root element
    * <tt>ParameterHandler</tt>.
    *
+   * Before printing, all parameters and subsections of the present subsection
+   * are printed sorted alphabetically by default.
+   * This option can be disabled setting the last parameter @p sort_alphabetical
```
bangerth: same two comments here

include/deal.II/base/parameter_handler.h
```diff
@@ -1417,27 +1431,40 @@ class ParameterHandler : public Subscriptor
...
 
   /**
    * Print parameters to a logstream. This function allows to print all
    * parameters into a log-file. Sections will be indented in the usual log-
    * file style.
+   *
+   * All current parameters and subsections are printed sorted
+   * alphabetically by default.
+   * This option can be disabled setting the last parameter @p sort_alphabetical
+   * to @p false: in this case entries are printed in the same order
```
bangerth: and here

include/deal.II/base/parameter_handler.h
```diff
@@ -1417,27 +1431,40 @@ class ParameterHandler : public Subscriptor
...
   /**
    * Log parameters in the present subsection. The subsection is determined by
    * the <tt>subsection_path</tt> member variable. This variable is controlled
    * by entering and leaving subsections through the enter_subsection() and
    * leave_subsection() functions.
    *
+   * All current parameters and subsections are printed sorted
+   * alphabetically by default.
+   * This option can be disabled setting the last parameter @p sort_alphabetical
+   * to @p false: in this case entries are printed in the same order
```
bangerth: and here

include/deal.II/base/parameter_handler.h
```diff
@@ -1700,6 +1727,16 @@ class ParameterHandler : public Subscriptor
...
 
+  /**
+   * Sort all parameters of the subsection given by the
+   * @p target_subsection_path argument in alphabetical order,
+   * as well as all subsections within it recursively.
+   */
+  void
+  recursively_sort_parameters(
+    boost::property_tree::ptree &   tree,
+    const std::vector<std::string> &target_subsection_path) const;
```
bangerth: We typically have the input arguments first, then the output arguments. Would you mind reversing the order of the two arguments? 

You make it sound in the description of this function as if the function *only* operates on the given arguments. Could it then be made `static` (and then of course non-`const`)?

elauksap:
> What explains a change in the output order though? I was not expecting the normal output to change. Is this because you are making a copy of the property_tree now?

It is due to the sorting. By default `sort()` mixes up parameters and subsections. I'm currently investigating whether providing a custom comparator to the sorting function can fix the issue.

I will update you soon!

include/deal.II/base/parameter_handler.h
tjhei: It can also be moved into an anonymous namespace in the .cc instead, but I would suggest to move forward with the PR instead (after we figure out if the functionality is correct).

tjhei:
> I will update you soon!

easier option: we only do the copy&sorting if needed, for example:
```
  boost::property_tree::ptree sorted_entries;
  boost::property_tree::ptree* current_entries = &(*entries);

  // Sort parameters alphabetically, if needed.                                                                                             
  if (sort_alphabetical)
    {
      sorted_entries = *entries;
      current_entries = &sorted_entries;

      // Dive recursively into the subsections,                                                                                             
      // starting from the top level.                                                                                                       
      recursively_sort_parameters(sorted_entries, std::vector<std::string>());
    }
[...]
      write_json(out, *current_entries);
```

include/deal.II/base/parameter_handler.h
elauksap: The order of the two arguments has been reversed, and the method has been made static.

Currently it is not possible to move it into the anonymous namespace since it makes use of `collate_path_string` which is non-static and in turn refers to `path_separator`, which is a *private* static class member. This of course can be easily fixed, but since this goes beyond the scope of this PR I preferred not to implement here also these changes.

elauksap:
> 
> 
> > I will update you soon!
> 
> easier option: we only do the copy&sorting if needed, for example:
> 
> ```
>   boost::property_tree::ptree sorted_entries;
>   boost::property_tree::ptree* current_entries = &(*entries);
> 
>   // Sort parameters alphabetically, if needed.                                                                                             
>   if (sort_alphabetical)
>     {
>       sorted_entries = *entries;
>       current_entries = &sorted_entries;
> 
>       // Dive recursively into the subsections,                                                                                             
>       // starting from the top level.                                                                                                       
>       recursively_sort_parameters(sorted_entries, std::vector<std::string>());
>     }
> [...]
>       write_json(out, *current_entries);
> ```

Thanks! However I'm not sure if this solves the issue, since it lays inside the `recursively_sort_parameters` method!

Using a custom comparator actually does what expected (see the updated commit), and all the output are now sorted in alphabetical order (with parameters printed before nested subsections).

You can still find few differences in some tests output: I've found out that before this PR some parameters where not printed 100% alphabetically. I can't really tell why, but after my last commit XML and JSON output has become consistent with what done in Text, LaTeX and Description modes.

elauksap:
> easier option: we only do the copy&sorting if needed

Done, thanks for the suggestion!

elauksap:
I've rebased on top of 1b19a91cbc444375c99d5b7f34f3770c9c86c04a (pull request #9430).

tjhei:
I finally figured out why you had to update the test output: The old printing in xml and json format was printing the output as declared and you now changed it to be sorted by default, whereas the text output used to be sorted (and did not change with the current default).
Even though this is not 100% backwards compatible, I am okay with the change (xml and json are to be consumed by a machine anyways).

bangerth:
Excellent. Thanks, everyone, and welcome @elauksap to the list of deal.II authors!

You picked a complicated one as a first pull request, but I promise you that they're not all that complex to finally get approved. You'll also learn to anticipate many of the style issues and it will all get much easier over time! (Also, I hope we get to meet when I visit MOX on 2/28!)

elauksap:
Thank you all, it has been a pleasure to give my small contribution to this fantastic community!
Even though I had previously contributed with another PR (#8946), this one was for sure more challenging and rewarding!

See you at MOX next month @bangerth @luca-heltai 😉

PR #9264, opened 2020-01-08
added measure computation for distorted 2d cells

nicola-giuliani:


source/grid/tria_accessor.cc
```diff
@@ -1286,42 +1286,35 @@ namespace
...
-                 ((v03 * v03) * (v01 * v01) * (v02 * v02))) >= 1e-24)
-      {
-        Assert(
-          false,
-          ExcMessage(
-            "Computing the measure of a nonplanar face is not implemented!"));
-        return std::numeric_limits<double>::quiet_NaN();
-      }
+    // In general the area can be computed as
+    // 0.25*(v_0+v_1-v_2-v_3)*(v_0-v_1+v_2-v_3)
```
bangerth: That's interesting. Can you explain why this is so? I have to say that geometrically this formula seems unlikely to me because you are, in essence taking the cross product of the tangent vectors at the center.

Think about the planar face in 3d with coordinates
```
  0,0,0      1,0,0     0,1,0     1,1,0
```
and then the one where you lift opposing vertices above the xy plane and move the others below:
```
  0,0,1      1,0,-1     0,1,-1     1,1,1
```
Your formula predicts the same are, but that's clearly not the case.

source/grid/tria_accessor.cc
```diff
@@ -1286,42 +1286,35 @@ namespace
...
+                                 accessor.vertex(2) - accessor.vertex(3);
+    const Tensor<1, 3> piece_2 = accessor.vertex(0) - accessor.vertex(1) +
+                                 accessor.vertex(2) - accessor.vertex(3);
+
+    return 0.25 * cross_product_3d(piece_1, piece_2).norm();
   }
 
+  // // a 2d cell in 3d space
+  // double
+  // measure(const dealii::TriaAccessor<2, 2, 3> &accessor)
```
bangerth: If you remove a function, just remove it altogether rather than leaving it commented out.

nicola-giuliani: You are more than right ;)

source/grid/tria_accessor.cc
nicola-giuliani: You are absolutely right again. Thanks, I will look again tomorrow, and If I don't found the mistake I will close the PR.

bangerth: The area you compute is that of the projection onto the plane perpendicular to the normal vector of the face at the center of the face. I would not be surprised if there is a formula that expresses the true area in terms of a warp or twist relative to this projection.

But think about how you would compute things mathematically: You want to compute
```
  \int_F  1   ds_1 ds_2
```
which you can pull back to the reference face using
```
  \int_{\hat F}   det(ds/d\hat s)  dxi deta
  = \int_{\hat F}   det(J)  dxi deta
```
The Jacobian (of the mapping) can be computed from the coordinates of the face. Then the integrand will be of the form `polynomial of degree d` and should be computable. The only thing you'll have to pay attention to is that `J` is a `3x2` matrix, and I think you have to replace `det(J)` by `sqrt(det(J^T J))`. I'll leave the details for you to figure out :-)

nicola-giuliani: The surface in codim one should be defined as the integral of the norm of the cross product of the two tangential vectors. I have assumed a bi-linear patch, am I correct to assume that the proper result should be the same as the one computed integrating 1 on the patch using a MappingQ1 and a properly accurate QGauss?

bangerth: I think that's true. The norm of the cross product of the two properly scaled tangential vectors is simply the determinant of the Jacobian.

The current formula just makes use that under certain situations, the integrand is constant and so can be computed analytically.

nicola-giuliani:
@bangerth I didn't manage to come up with a proper analytical formula so I added a numerical quadrature of order 4 if the quadratical cell is non planar. If we manage in the future to find the proper analytical formula we can always change it. The test on the measure computation if a real non planar cell converges reasonably well.

source/grid/tria_accessor.cc
```diff
@@ -1308,11 +1311,92 @@ namespace
     if (std::abs((v03 * normal) * (v03 * normal) /
                  ((v03 * v03) * (v01 * v01) * (v02 * v02))) >= 1e-24)
       {
-        Assert(
-          false,
-          ExcMessage(
-            "Computing the measure of a nonplanar face is not implemented!"));
-        return std::numeric_limits<double>::quiet_NaN();
+        // If the vectors are non planar we integrate the norm of the normal
+        // vector using a numerical Gauss scheme of order 4.
```
bangerth: This is a bit too abstract even for me. Can you write out in formulas what we are integrating here?

tests/codim_one/measure_of_distorted_codim_one_face.cc
```diff
@@ -0,0 +1,101 @@
...
+
+void
+test()
+{
+  Triangulation<2, 3>      tria;
+  std::vector<Point<3>>    vertices;
+  std::vector<CellData<2>> cells;
+  SubCellData              subcelldata;
+
+  double tol = 1e-12;
```
bangerth:
```suggestion
  const double tol = 1e-12;
```

tests/codim_one/measure_of_distorted_codim_one_face.cc
```diff
@@ -0,0 +1,101 @@
...
+  // vertices.push_back(Point<3>{10, 56, 0});
+  // vertices.push_back(Point<3>{22, 1, 0});
+  // vertices.push_back(Point<3>{15, 44, 0});
+  // vertices.push_back(Point<3>{1, 1, 1});
```
bangerth: just remove these lines if you no longer need them.

tests/codim_one/measure_of_distorted_codim_one_face.cc
```diff
@@ -0,0 +1,101 @@
...
+  QGauss<2>       quadrature_formula(4);
+  MappingQ1<2, 3> mapping;
+  FEValues<2, 3>  fe_values(mapping, fe, quadrature_formula, update_JxW_values);
+
+  double sum_1    = 0;
+  double sum_2    = 0;
+  auto   cell     = dof_handler.begin_active();
+  auto   accessor = tria.begin_active();
+
+  sum_1 = cell->measure();
```
bangerth: just write
```suggestion
  const double sum_1 = cell->measure();
```
and then delete the declaration of the variable above.

nicola-giuliani: Ok

tests/codim_one/measure_of_distorted_codim_one_face.cc
nicola-giuliani: yes, thanks

source/grid/tria_accessor.cc
nicola-giuliani: Sure, I've added the description of the procedure.

bangerth:
Looks good. Would you mind adding an entry to `doc/news/changes/minor` so you get the credit you deserve?

nicola-giuliani:
@bangerth, @masterleinad   can you tag this PR as ready to test so that we can proceed with the testing/eventual merge?

masterleinad:
```
grid/measure_of_3d_face_01.debug: DIFF failed. ------ Diff:   /home/dealii/build/tests/grid/measure_of_3d_face_01.debug/failing_diff
grid/measure_of_3d_face_01.debug: DIFF failed. ------ First 20 lines of numdiff output:
----------------
##67      #:4   <== nan
##67      #:4   ==> 1.2809
```

doc/news/changes/minor/20200123NicolaGiuliani
```diff
@@ -0,0 +1,3 @@
+New: Add method to compute measure of distorted quads embedded in a three dimensional space.
+<br>
+(Nicola Giuliani, 2020/01/23)
```
bangerth: Please also end the file with a newline.

bangerth:
It's possible that the difference in the output file is actually correct. The function seemed to abort before and now yields anything useful. But I don't know whether it's correct -- the test seems to have computed the correct value analytically, but the value computed by the new function doesn't match. Can you take a look?

nicola-giuliani:
@bangerth sorry for the late answer. I looked into the code and the numerical integral of a FE_Q(1) over the distorted cell is around 1.281 (depending on the quadrature formula). That should be exactly what I am computing with the new measure. I don't know how the 5/3 of the test is actually computed. If you want I can commit an updated output file for the failing test.

luca-heltai:
@nicola-giuliani , can you fix the failing test output (https://jenkins.tjhei.info/blue/organizations/jenkins/dealii/detail/PR-9264/6/tests)? I think that the computation you are doing is now correct.

masterleinad:
I removed the `reviewed and ready to merge` label for now (until the test is fixed). We have a number of pull requests where the CI needs to be restarted and I am really irritated when I look at a pull request with a real test failure that has that label.

nicola-giuliani:
> @nicola-giuliani , can you fix the failing test output (https://jenkins.tjhei.info/blue/organizations/jenkins/dealii/detail/PR-9264/6/tests)? I think that the computation you are doing is now correct.

@luca-heltai done

luca-heltai:
> I removed the `reviewed and ready to merge` label for now (until the test is fixed). We have a number of pull requests where the CI needs to be restarted and I am really irritated when I look at a pull request with a real test failure that has that label.

Sorry if I irritated you. In my head, the "reviewed and ready to merge" refers to the actual code, and it comes without saying that no-one presses the merge button until all CI pass. The label does not say "Reviewed, all tests passing, ready to merge" after all. For that we have the testers... :)

masterleinad:
> Sorry if I irritated you. In my head, the "reviewed and ready to merge" refers to the actual code, and it comes without saying that no-one presses the merge button until all CI pass. The label does not say "Reviewed, all tests passing, ready to merge" after all. For that we have the testers... :)

That's totally fine. My point is that I regularly look for all the `Reviewed and ready to merge` which signals to me "no action required, the pull request is good to go". If it seems to be a problem with the CI itself, I rerun it (The pull request doesn't show as failing anymore) and if there are tests failing that require interaction I would notify the author and remove the label so I won't look at it again until it is (potentially) resolved. It normally doesn't make much of a difference unless there are like 10 pull requests that are potentially ready to be merged you need to look through.
Anyway, it's not a big deal and I certainly felt responsible for adding it myself again here.

PR #9266, opened 2020-01-08
CMake: Check whether Trilinos has been configured with third-party libraries.

### LINKED ISSUE #9252: Test mpi/hp_step_4 fails because of missing Trilinos component
This test fails for me with the error
```
7614: --------------------------------------------------------
7614: An error occurred in line <829> of file </home/fac/f/bangerth/p/deal.II/1/dealii/source/lac/trilinos_solver.cc> in function
7614:     void dealii::TrilinosWrappers::SolverDirect::do_solve()
7614: The violated condition was: 
7614:     Factory.Query(additional_data.solver_type.c_str())
7614: Additional information: 
7614:     You tried to select the solver type <Amesos_Mumps> but this solver is not supported by Trilinos either because it does not exist, or because Trilinos was not configured for its use.
7614: --------------------------------------------------------
```

I don't know what the appropriate `with_...=true` component to the output file name should be in this case, but we ought to predicate running the test on the existence of Amesos_Mumps.

@marcfehling -- FYI.

marcfehling:
Not entirely sure whether this fixes #9252, since I don't know in which format Trilinos actually stores its third-party libraries, i.e. `mumps|Mumps|MUMPS`.

@BenBrands -- Would you be so kind and check whether the tests `mpi/hp_step-4` show up in your `ctest -N | grep "hp_step-4"` after applying the update?

Fixes #9252


BenBrands:
No, after your modifications the tests `hp_step-4` are not picked up by ctest.

marcfehling:
Just noticed that I introduced a typo :(
Would you try again? Hopefully this is it!

BenBrands:
It is working!

cmake/configure/configure_2_trilinos.cmake
```diff
@@ -175,6 +181,20 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
           MESSAGE(STATUS "Module ${_optional_module} not found!")
       ENDIF()
       ENDFOREACH()
+
+      #
+      # Check for third-party libraries (tpl).
+      #
+      FOREACH(_optional_tpl MUMPS)
+      ITEM_MATCHES(_tpl_found ${_optional_tpl} ${Trilinos_TPL_LIST})
```
marcfehling: We may combine the check for third-party libraries with the module check above to save some lines of code. However, I prefer to treat them separately.

bangerth: I think that's alright as you do here. I've learned what TPL stands for this way :-)

Can you indent lines 189-196 by two spaces to make it easier to read? (I know you only copied the text from above, where that is also a problem -- if you feel ambitious, also indent lines 175-182.)

marcfehling:
There may be issues with the memory space for some testers  -- @tjhei
```
/jenkins/workspace/dealii_PR-9266/tests/fe/mass_matrix_arbitrary_nodes.cc:90:1: fatal error: error writing to /tmp/cctGcRue.s: No space left on device
```

tamiko:
Well, we configured correctly and that is what I wanted to se..

Issue #9271, opened 2020-01-09
DoFTools::count_dofs_per_block clearly wrong/inefficient for the hp case


bangerth:
I'm rewriting `DoFTools::count_dofs_per_{component,block}` for a separate PR, and I'm finding that `DoFTools::count_dofs_per_block()` can't possibly be doing what we think it's doing in the hp case. Specifically, the code currently looks like this:
```
void
  count_dofs_per_block(const DoFHandlerType &                dof_handler,
                       std::vector<types::global_dof_index> &dofs_per_block,
                       const std::vector<unsigned int> &     target_block_)
  {
    std::vector<unsigned int> target_block = target_block_;

    const dealii::hp::FECollection<DoFHandlerType::dimension,
                                   DoFHandlerType::space_dimension>
      &fe_collection = dof_handler.get_fe_collection();
    Assert(fe_collection.size() < 256, ExcNotImplemented());

    for (unsigned int this_fe = 0; this_fe < fe_collection.size(); ++this_fe)
      {
        const FiniteElement<DoFHandlerType::dimension,
                            DoFHandlerType::space_dimension> &fe =
          fe_collection[this_fe];
        std::fill(dofs_per_block.begin(),                              // ************************
                  dofs_per_block.end(),
                  types::global_dof_index(0));

        ...
```
Note how in the `std::fill` call, we set all elements of the output array `dofs_per_block` to zero, overwriting what we've computed for previous elements of the collection. The only other state that is carried from one loop iteration is the `target_block` array, but that turns out to also be overwritten in each iteration without regard to any previous content:
```
        // If the empty vector was given as default argument, set up this
        // vector as identity.
        if (target_block.size() == 0)
          {
            target_block.resize(fe.n_blocks());
            for (unsigned int i = 0; i < fe.n_blocks(); ++i)
              target_block[i] = i;
          }
        else
          Assert(target_block.size() == fe.n_blocks(),
                 ExcDimensionMismatch(target_block.size(), fe.n_blocks()));
```
(Interestingly, that means that it is only set up once, for the first element of the collection; for all following elements, we end up in the `else` case.)

In other words, the function really just looks at the last element of the collection. At best, this is inefficient; at worst it's wrong -- I haven't decided which one yet, but the function is for sure not written in a particularly clear way and/or as the author had intended it to.

kronbichler:
@bangerth any plans to still work on that for the release? If not, please adjust the target milestone.

peterrum:
[...]

PR #9272, opened 2020-01-09
Make DoFTools::count_dofs_per_{component,block} return its data.

bangerth:
Instead of doing so via an output argument. The existing functions are deprecated and just call the new versions.

See also #9271 for an issue that is mentioned in the patch. I haven't touched the tests and example programs yet because I wanted to be sure that there is no change in test results first.

/rebuild


bangerth:
Ping?

PR #9274, opened 2020-01-09
fixes a keyboard fat-fingering typo wherever 'he' in incorrectly used instead of 'the'

krishnakumarg1984:


bundled/muparser_v2_2_4/include/muParserTokenReader.h
```diff
@@ -1,161 +1,161 @@
-/*
```
bangerth: What's going on with this file? Did you change the line ending?

tjhei:
Please leave the code in ``bundled/`` as is (see comment in other PR).

bundled/muparser_v2_2_4/include/muParserTokenReader.h
krishnakumarg1984: Yes. Something incorrect had happened. I have restored this file to the former state now (I think).

krishnakumarg1984:
All code in `bundled/` is now restored to its original state.

PR #9268, opened 2020-01-09
Use AssertIndexRange some more

masterleinad:
Mostly generated by
```
find include/ source/ \( -name "*.cc" -o -name "*.h" \) -exec perl -0777 -i.original -pe 's/Assert\(([^<]*)[\s]*<([^,]*),[^E]*ExcIndexRange[^(]*\([\s]*\1,[\s]*0,[\s]*\2[\s]*\)[\s]*\);/AssertIndexRange(\1,\2);/igs' {} \;
```

PR #9283, opened 2020-01-10
Remove tensor.cc

masterleinad:
I don't see any reason to explicitly instantiate the `Tensor` class. The file was introduced to make  `Tensor::n_independent_components` ODR-usable. We can achieve this by moving the external declaration to the header file as well.
#9268 shows that we already have some cases where we are relying on implicit template instantiation where removing these results in undefined references to the respective `Tensor::n_independent_components` variables.


bangerth:
/rebuild

Let's see what happens!

PR #9284, opened 2020-01-11
Fix \coloneq again.

bangerth:
It turned out that my previous attempt at fixing the spacing in `\coloneq` in #9277 did not actually work -- but it's really hard to tell between browsers willing to cache nearly everything and doxygen having its own system of deciding which formulas need to be run through latex again.

Anyway, I verified that this finally works now: I just blew away the doxygen directories and forced it to re-generate everything.


drwells:
I rebuilt the documentation and checked that `\dealcoloneq` looks right.

PR #9286, opened 2020-01-11
Introduce \jump and \average commands to be used in the documentation.

bangerth:
This makes sure that we can be consistent in how we denote these operators: single or double brackets/braces; with or without spaces, etc.


bangerth:
This also satisfies a promise I made to @tjhei to be consistent between `FEInterfaceValues` and step-71. I will convert step-71 after that is merged (it's a collaboration with others, so I don't want to rebase that branch).

tjhei:
looks good. Please re-indent.

bangerth:
Indented. I keep forgetting that it also indents comments...

krishnakumarg1984:
if I remember  correctly, the video lectures also had this (minor) inconsistency with notation of the jump function in the gradient.

Issue #9099, opened 2019-11-26
Extend DataOut for multigrid levels


peterrum:
@tamiko, @tjhei, and me were working independently on the extension of `DataOut` for multigrid levels. In my opinion, we should be able to output the results on the active levels and on each (!) multigrid level, as shown in the following code:

```cpp
#include <deal.II/distributed/tria.h>

#include <deal.II/dofs/dof_handler.h>

#include <deal.II/fe/fe_dgq.h>

#include <deal.II/grid/grid_generator.h>

#include <deal.II/lac/la_parallel_vector.h>

#include <deal.II/multigrid/mg_constrained_dofs.h>
#include <deal.II/multigrid/mg_data_out.h>
#include <deal.II/multigrid/mg_transfer_matrix_free.h>

#include <deal.II/numerics/data_out.h>
#include <deal.II/numerics/vector_tools.h>

const int dim                = 2;
const int global_refinements = 3;
const int degree             = 2;

using namespace dealii;

using Number      = double;
using vector_type = LinearAlgebra::distributed::Vector<Number>;

template <int dim, typename Number = double>
class Solution : public Function<dim, Number>
{
  virtual Number
  value(const Point<dim> &p, const unsigned int component = 0) const
  {
    (void)component;
    return p.norm();
  }
};

int
main(int argc, char **argv)
{
  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);

  parallel::distributed::Triangulation<dim> tria(
    MPI_COMM_WORLD,
    dealii::Triangulation<dim>::none,
    parallel::distributed::Triangulation<dim>::construct_multigrid_hierarchy);
  GridGenerator::hyper_cube(tria);
  tria.refine_global(global_refinements);

  DoFHandler<dim> dof_handler(tria);
  FE_DGQ<dim>     fe(degree);
  dof_handler.distribute_dofs(fe);
  dof_handler.distribute_mg_dofs();

  MGConstrainedDoFs                 constrained_dofs;
  MGTransferMatrixFree<dim, Number> transfer;

  vector_type solution_vector;

  IndexSet locally_relevant_dofs;
  DoFTools::extract_locally_relevant_dofs(dof_handler, locally_relevant_dofs);

  solution_vector.reinit(dof_handler.locally_owned_dofs(),
                         locally_relevant_dofs,
                         MPI_COMM_WORLD);

  VectorTools::interpolate(dof_handler, Solution<dim>(), solution_vector);

[...]

PR #9287, opened 2020-01-11
Replace the Dataout::first_cell/next_cell() mechanism

bangerth:
Instead, use function objects that can be used with lambda functions. This addresses the idea mentioned in https://github.com/dealii/dealii/issues/9099#issuecomment-571775767

The current approach is backward compatible in that the new function objects by default just call the old function. As a consequence, all of the patches currently floating around (#9100, #9096) should continue to work -- but could be streamlined a bit by just calling `DataOut::set_cell_selection()` in their constructors instead of implementing several more functions.

/rebuild

### DIFF
--- a/doc/news/changes/minor/20200111Bangerth
+++ b/doc/news/changes/minor/20200111Bangerth
@@ -1,0 +1,6 @@
+New: The mechanism implemented by DataOut via the `first_cell()` and
+`next_cell()` virtual functions has been deprecated. In its place, the
+DataOut::set_cell_selection() function allows doing the same but
+without the need for deriving classes.
+<br>
+(Wolfgang Bangerth, 2020/01/11)
--- a/include/deal.II/numerics/data_out.h
+++ b/include/deal.II/numerics/data_out.h
@@ -20,6 +20,8 @@
 
 #include <deal.II/base/config.h>
 
+#include <deal.II/grid/filtered_iterator.h>
+
 #include <deal.II/numerics/data_out_dof_data.h>
 
 #include <memory>
@@ -111,40 +113,32 @@ namespace internal
  * <h3>Extensions</h3>
  *
  * By default, this class produces patches for all active cells. Sometimes,
- * this is not what you want, maybe because they are simply too many (and too
+ * this is not what you want, maybe because there are simply too many (and too
  * small to be seen individually) or because you only want to see a certain
- * region of the domain (for example in parallel programs such as the step-18
- * example program), or for some other reason.
+ * region of the domain (for example only in the fluid part of the domain in
+ * step-46), or for some other reason.
  *
  * For this, internally build_patches() does not generate the sequence of
- * cells to be converted into patches itself, but relies on the two functions
- * first_cell() and next_cell(). By default, they return the first active
- * cell, and the next active cell, respectively. Since they are @p virtual
- * functions, you can write your own class derived from DataOut in which you
- * overload these two functions to select other cells for output. This may,
+ * cells to be converted into patches itself, but relies on the two function
+ * that we'll call first_cell() and next_cell(). By default, they return the
+ * first active cell, and the next active cell, respectively. But this can
+ * be changed using the set_cell_selection() function that allows you to
+ * replace this behavior. What set_cell_selection() wants to know is how
+ * you want to pick out the first cell on which output should be generated,
+ * and how given one cell on which output is generated you want to pick the
+ * next cell.
+ *
+ * This may,
  * for example, include only cells that are in parts of a domain (e.g., if you
  * don't care about the solution elsewhere, think for example a buffer region
  * in which you attenuate outgoing waves in the Perfectly Matched Layer
  * method) or if you don't want output to be generated at all levels of an
  * adaptively refined mesh because this creates too much data (in this case,
- * the set of cells returned by your implementations of first_cell() and
- * next_cell() will include non-active cells, and DataOut::build_patches()
+ * the set of cells returned by your implementations of the `first_cell` and
+ * `next_cell` arguments to set_cell_selection() will include
+ * non-active cells, and DataOut::build_patches()
  * will simply take interpolated values of the solution instead of the exact
- * values on these cells children for output). Once you derive your own class,
- * you would just create an object of this type instead of an object of type
- * DataOut, and everything else will remain the same.
- *
- * The two functions are not constant, so you may store information within
- * your derived class about the last accessed cell. This is useful if the
- * information of the last cell which was accessed is not sufficient to
- * determine the next one.
- *
- * There is one caveat, however: if you have cell data (in contrast to nodal,
- * or dof, data) such as error indicators, then you must make sure that
- * first_cell() and next_cell() only walk over active cells, since cell data
- * cannot be interpolated to a coarser cell. If you do have cell data and use
- * this pair of functions and they return a non-active cell, then an exception
- * will be thrown.
+ * values on these cells children for output).
  *
  * @ingroup output
  * @author Wolfgang Bangerth, 1999
@@ -159,6 +153,8 @@ public:
                 "The dimension given explicitly as a template argument to "
                 "this class must match the dimension of the DoFHandler "
                 "template argument");
+  static constexpr unsigned int spacedim = DoFHandlerType::space_dimension;
+
   /**
    * Typedef to the iterator type of the dof handler class under
    * consideration.
@@ -211,6 +207,11 @@ public:
      */
     curved_inner_cells
   };
+
+  /**
+   * Constructor.
+   */
+  DataOut();
 
   /**
    * This is the central function of this class since it builds the list of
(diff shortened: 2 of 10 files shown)


doc/news/changes/minor/20200111Bangerth
```diff
@@ -0,0 +1,6 @@
+New: The mechanism implemented by Dataout via the `first_cell()` and
```
masterleinad: I am pretty sure that you need `DataOut::first_cell()` and `DataOut::next_cell()` for `doxygen` to pick it up.

bangerth: I specifically didn't want doxygen to pick these names up because I intend to remove these functions :-)

masterleinad: I see. But I guess you still want 
```suggestion
New: The mechanism implemented by DataOut via the `first_cell()` and
```

tjhei:
> I find that an implementation with `FilteredIterator` as suggested by @tjhei would be cleaner.

FilteredIterator also turns out to be a bit awkward to use in practice as the predicate is part of the type of begin and end.

Can we change set_cell_selection to take a single argument that allows passing ``IteratorRange< FilteredIterator< BaseIterator > >``
so we can use
``filter_iterators(triangulation.active_cell_iterators(),IteratorFilters::LocallyOwnedCell())``
and
``triangulation.cell_iterators_on_level(lvl)``
etc.? See https://www.dealii.org/developer/doxygen/deal.II/group__CPP11.html#ga3948dd66d9e66ac55d89dc8f75fa4855

include/deal.II/numerics/data_out.h
```diff
@@ -316,42 +313,117 @@ class DataOut : public DataOut_DoFData<DoFHandlerType,
...
+   * objects that can be used what the first cell on which output is
+   * generated is supposed to be, and what given one cell the next
+   * function is supposed to be. Through these function objects,
+   * it is possible to select a subset of cells on which output should
+   * be produced (e.g., only selecting those cells that belong to a
+   * part of the domain -- say, the fluid domain in a code such as step-46),
+   * or to completely change *where* output is produced (e.g., to produce
+   * output on non-active cells of a multigrid hierarchy or if the finest
+   * level of a mesh is so fine that generating graphical output would lead
+   * to an overwhelming amount of data.
```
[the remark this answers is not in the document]
bangerth: Let's not document things that aren't possible *yet* :-)

include/deal.II/numerics/data_out.h
```diff
@@ -316,42 +313,117 @@ class DataOut : public DataOut_DoFData<DoFHandlerType,
...
+   *
+   * @param[in] A function object that takes as argument the triangulation
+   *   this class works on and that should return the first cell on which
+   *   output should be generated.
+   * @param[in] A function object that takes as argument the last cell
+   *   on which output was generated and that should return the next
+   *   cell on which output should be generated. (In practice, the
+   *   function object may be triggered *before* any processing
```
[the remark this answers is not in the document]
bangerth: I've removed the sentence. I meant to say that *all* of the walking over cells might happen before the first output is generated. But that's really immaterial to the user to know.

include/deal.II/numerics/data_out.h
```diff
@@ -316,42 +313,117 @@ class DataOut : public DataOut_DoFData<DoFHandlerType,
...
+   *   actually happens, but you get the idea.)
```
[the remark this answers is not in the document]
bangerth: I documented both. Since the triangulation is so easy to obtain, I think there's no point in passing it explicitly (and most implementations wouldn't need it because they just do something like `++cell`.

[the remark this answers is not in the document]
bangerth: Actually, I change my mind.

source/numerics/data_out.cc
```diff
@@ -65,6 +65,24 @@ namespace internal
...
+template <int dim, typename DoFHandlerType>
+DataOut<dim, DoFHandlerType>::DataOut()
+{
+  // For the moment, just call the existing virtual functions. This
+  // preserves backward compatibility. When these deprecated functions are
+  // removed, we'll just inline their functionality into the lambda
+  // functions created here.
+  set_cell_selection(
+    [this](const Triangulation<dim, spacedim> &) {
+      return this->first_locally_owned_cell();
```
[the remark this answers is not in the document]
bangerth: First, no it doesn't trigger a deprecation warning: We disable these while compiling the library itself.

Second: For backward compatibility, this is exactly the behavior we want to implement: To call the existing functions. Only when we remove the deprecated functions should we inline their implementations into the lambda function here.

bangerth:
I've addressed the comments made before, and pushed another commit that implements a second version of the function. What do you think of the latter?

I'll caution that the new version does not select locally owned cells. I leave this to the provided filter -- or should I just filter on that as well?

tjhei:
> I've addressed the comments made before, and pushed another commit that implements a second version of the function. What do you think of the latter?

I think usability should be much better with the filtered iterators compared to writing the lambda functions. :+1:  It might not be a bad idea to keep both around for more flexibility.

> I'll caution that the new version does not select locally owned cells. I leave this to the provided filter -- or should I just filter on that as well?

1. Your documentation currently says that it filters for locally owned (even though it doesn't). I can think of one case where I would not want to filter out locally_owned_cells: to visualize what part of the mesh each processor sees, including ghost cells. Maybe this is not too important as we have done it with GridOut in the past. It gets more complicated to do the right thing if you allow non-active cells, though (owned vs owned on level). Maybe let the user deal with it?
2. You are currently using ``begin_active()`` but ``cell_iterator``. Does this mean you cover active and non-active cells but only starting from begin_active() right now? I would want this function to work on all cells (begin() to end()) by default.

bangerth:
Good points. I've addressed them both in the second to last commit. (I've changed the description of the function, and I've made the iterator start at `.begin()` instead of `.begin_active()` -- that was a mistake.)

tjhei:
I like it!
To make things testable, can we keep the first/next_cell() public functions and make them call the lambda functions. Then we can have simple tests like in  #9278 to check if the API works as intended (select a subset of cells, non-active ones, etc).

bangerth:
Well, I specifically wanted to get away from the virtual functions because I think that's a rather clunky interface. Right now, the lambdas by default call the virtual functions. But the test will be easy to adapt: You can keep the (virtual or not) function in your class and in the constructor you just call the `set_cell_selection()` function with two lambda functions that just call the functions in your class.

With this patch, we now have two interfaces -- one via the virtual functions (which the patch deprecates) and one via the function objects. I'd rather not keep both of these interfaces around indefinitely.

tjhei:
> Well, I specifically wanted to get away from the virtual functions because I think that's a rather clunky interface.

Alright. I don't want to keep them as a way to customize behavior by deriving, I suggested to have something to unit test the implementation of set_cell_selection().

> With this patch, we now have two interfaces -- one via the virtual functions (which the patch deprecates) and one via the function objects. I'd rather not keep both of these interfaces around indefinitely.

If the deprecated functions are removed, I don't know of a way to test that what you implemented in this PR actually works (the FilteredIterator logic for example). Would you mind writing some tests that show that what you implemented works or suggest how you would test it?

bangerth:
Sure, I just wanted to see whether everyone likes the design first. Two tests for the two functions are in the last commit now! (Turns out that there was also a bug in my code for the second `set_cell_selection()` before -- good to have checked!)

tests/data_out/data_out_08_set_cell_selection_01.cc
```diff
@@ -0,0 +1,144 @@
...
+  GridGenerator::hyper_cube(tria, 0., 1.);
+  tria.refine_global(1);
+
+  Vector<double> cell_data(4);
+  for (unsigned int i = 0; i < 4; ++i)
+    cell_data(i) = i * 1.0;
+
+  // this should skip the first cell
+  typename Triangulation<dim>::active_cell_iterator it = tria.begin_active();
+  //++it;
```
tjhei: please remove

tests/data_out/data_out_08_set_cell_selection_01.cc
```diff
@@ -0,0 +1,144 @@
...
+  it->set_subdomain_id(1);
+
+  FE_DGQ<dim>     fe(0);
+  DoFHandler<dim> dof_handler(tria);
+  dof_handler.distribute_dofs(fe);
+
+  //  DataOut<dim> data_out;
```
tjhei: please remove

tests/data_out/data_out_08_set_cell_selection_02.cc
```diff
@@ -0,0 +1,128 @@
...
+  GridGenerator::hyper_cube(tria, 0., 1.);
+  tria.refine_global(1);
+
+  Vector<double> cell_data(4);
+  for (unsigned int i = 0; i < 4; ++i)
+    cell_data(i) = i * 1.0;
+
+  // this should skip the first cell
+  typename Triangulation<dim>::active_cell_iterator it = tria.begin_active();
+  //++it;
```
tjhei: please remove

tests/data_out/data_out_08_set_cell_selection_02.cc
```diff
@@ -0,0 +1,128 @@
...
+  it->set_subdomain_id(1);
+
+  FE_DGQ<dim>     fe(0);
+  DoFHandler<dim> dof_handler(tria);
+  dof_handler.distribute_dofs(fe);
+
+  //  DataOut<dim> data_out;
```
tjhei: please remove

tjhei:
can you please rebase to master?

bangerth:
Fixed. I thought about this some more, and it turns out that one can convert lambda functions to `FilteredIterator` objects. Take a look at the last commit -- I think you're really going to like this :-)

peterrum:
>  I think you're really going to like this :-)

Yup ;)

tjhei:
awesome!

PR #9288, opened 2020-01-11
Update constraints documentation module to reflect recent updates to step-6

krishnakumarg1984:
Removes a stale sentence that incorrectly stated that the step-6 tutorial uses the `condense` followed by `distribute` approach to constraint-handling. It now appears that step-6 has been updated to use the distribute_local_to_global() approach.

The real questions to the developers are the following:

What are the plans for the simplistic (first) approach to handling constraints?  Will this functionality be marked for deprecation (as a legacy feature) in future releases of deal.II (since it doesn't possess any advantages that decisively outweigh the 2nd approach for the user's problems). Is there any reason to keep the 1st approach? If not, then the documentation on constraints can be simplified considerably and made more readable.

### DIFF
--- a/doc/doxygen/headers/constraints.h
+++ b/doc/doxygen/headers/constraints.h
@@ -179,7 +179,7 @@
  * see the discussion below.
  *
  *
- * <h5>Condensing vectors</h5>
+ * <h4>Condensing vectors</h4>
  *
  * Condensing vectors works exactly as described above for matrices. Note that
  * condensation is an idempotent operation, i.e. doing it more than once on a
--- a/doc/doxygen/headers/constraints.h
+++ b/doc/doxygen/headers/constraints.h
@@ -164,8 +164,7 @@
  * auxiliary vectors (e.g. methods using explicit orthogonalization
  * procedures).
  * Nevertheless, this process is more efficient due to its lower
- * memory consumption and is discussed in the first few programs
- * of the @ref Tutorial , for example in step-6.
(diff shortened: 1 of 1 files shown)


peterrum:
> What are the plans for the simplistic (first) approach to handling constraints? Will this functionality be marked for deprecation (as a legacy feature) in future releases of deal.II (since it doesn't possess any advantages that decisively outweigh the 2nd approach for the user's problems). Is there any reason to keep the 1st approach? If not, then the documentation on constraints can be simplified considerably and made more readable.

Why should it? The methods `distribute` and `distribute_local_to_global` have different uses. You can use `distribute` if you have solved your problem and you want to update the constrained values accordingly. The other function is used during assembly.

krishnakumarg1984:
@peterrum I was referring to the function `condense()` as used in the older versions of `step-6` tutorial, and in the video lecture.

drwells:
Thanks again for taking the time to fix this!

drwells:
@krishnakumarg1984 To answer your question: I don't think we have any plans for deprecating or removing the old way of doing things (that is, the explicit calls to `condense(matrix, vector)`). `distribute_local_to_global` is usually the right choice but using `condense()` was the only way for a long time so we should continue to support it. Its also useful to have as a reference - i.e., we can say that `distribute_local_to_global` does the same thing but piece by piece.

krishnakumarg1984:
Understood. Thank you for the explanation.

PR #9295, opened 2020-01-12
fixes stale documentation about boundary_object for hyper_ball function

### LINKED ISSUE #9294: [Step-6 documentation] Fix the outdated sentence mentioning the need to attach a boundary ID to the circular mesh
It looks like the `GridGenerator::hyber_ball` function has been massively improved over the last few releases,  but the documentation related to it in Step-6 has remained stale.

In the documentation of the `run()` function of the commented code, we have:

>The domain is again a circle, __which is why we have to provide a suitable boundary object as well__. 

It doesn't look like this is true anymore. We simply have:

```
 if (cycle == 0)
        {
          GridGenerator::hyper_ball(triangulation);
          triangulation.refine_global(1);
        }
      else
          .....
```

It looks like this documentation bug has been around for the last few releases. In fact, the last time, a boundary_id was attached was in version `8.5.0` of deal.II

```
      if (cycle == 0)
        {
          GridGenerator::hyper_ball (triangulation);
          static const SphericalManifold<dim> boundary;
          triangulation.set_all_manifold_ids_on_boundary(0);
          triangulation.set_manifold (0, boundary);
          triangulation.refine_global (1);
        }
      else
          ...

```

I shall submit a PR to remove this sentence from the `master` version of Step-6 tutorial's doc.

krishnakumarg1984:
Fixes #9294

### DIFF
--- a/examples/step-6/step-6.cc
+++ b/examples/step-6/step-6.cc
@@ -513,13 +513,12 @@ void Step6<dim>::output_results(const unsigned int cycle) const
 // disk, in that we adaptively instead of globally refine the mesh, and that
 // we output the solution on the final mesh in the present function.
 //
-// The first block in the main loop of the function deals with mesh
-// generation. If this is the first cycle of the program, instead of reading
-// the grid from a file on disk as in the previous example, we now again
-// create it using a library function. The domain is again a circle, which is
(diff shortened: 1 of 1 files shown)


drwells:
Nice catch!

PR #9298, opened 2020-01-13
Clarify GridGenerator::hyper_shell in 3D

masterleinad:
Fixes the warning in https://cdash.43-1.org/viewBuildError.php?type=1&buildid=3667.

### DIFF
--- a/include/deal.II/grid/grid_generator.h
+++ b/include/deal.II/grid/grid_generator.h
@@ -1056,7 +1056,8 @@ namespace GridGenerator
    * Produce a half hyper-shell, i.e. the space between two circles in two
    * space dimensions and the region between two spheres in 3d, with given
    * inner and outer radius and a given number of elements for this initial
-   * triangulation.  However, opposed to the previous function, it does not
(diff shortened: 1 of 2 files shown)


drwells:
Am I missing something, or does it seem like a bug that we take `n` as an argument but never use it?

masterleinad:
> Am I missing something, or does it seem like a bug that we take `n` as an argument but never use it?

I don't think it's a bug. It's just that we only implement the one option for 3d.

tjhei:
> I don't think it's a bug. It's just that we only implement the one option for 3d.

We should document that fact. #9301

masterleinad:
I think it's better to not allow something arbitrary for the 3d implementation.

bangerth:
This seems wrong. In 3d, the parameter is just ignored as far as I can see, and consequently we should not ensure that it has a particular value. But the documentation is wrong in suggesting that in 3d there are certain allowed values.

masterleinad:
I won't fight for it. If you like the other pull request better, go with that. My idea was just to argue similarly for both implementations.

bangerth:
Let's go with #9301 instead.

PR #9300, opened 2020-01-13
In step 6 results doc, move the header file inclusion sentence to reflect that we need it for Sparse_ILU and not Jacobi preconditioner

krishnakumarg1984:


examples/step-6/doc/results.dox
```diff
@@ -142,13 +142,13 @@ into
...
 we can try out different relaxation parameters for SSOR. By using
-(you have to also add the header file <code>lac/sparse_ilu.h</code> to
-the include list at the top of the file)
 @code
   PreconditionJacobi<> preconditioner;
   preconditioner.initialize(system_matrix);
 @endcode
 we can use Jacobi as a preconditioner. And by using
+(you have to also add the header file <code>lac/sparse_ilu.h</code> to
+the include list at the top of the file)
```
bangerth: It reads funny with the parenthetical remark in the middle of the sentence. How about moving it  to the end of the sentence that ends on line 157? (The name of the header file predates a lot of work over nearly 20 years. It should really be `deal.II/lac/sparse_ilu.h` now.)

krishnakumarg1984: Sure, that is possible.

PR #9303, opened 2020-01-13
In the doc of the point() function, replaces monospace l_2 with its aTeX equivalent. Fixes #9302

### LINKED ISSUE #9302: In all documentation pages, consider replacing l_2 with its LaTeX equivalent notation
For example, in the documentation of the `distance` function [here](https://dealii.org/developer/doxygen/deal.II/classPoint.html#a3df8e6ab311dab9337c8d7b039c7b815), it says 

>Return the Euclidean distance of this point to the point p, i.e. the l_2 norm of the difference between the vectors representing the two points.

This would look much more readable if a mathjax-based LaTeX rendering is used. Can we replace all intances of `l_2` in the project with `$L_2$`?

krishnakumarg1984:
Fixes #9302

### DIFF
--- a/include/deal.II/base/point.h
+++ b/include/deal.II/base/point.h
@@ -16,7 +16,6 @@
 #ifndef dealii_point_h
 #define dealii_point_h
 
-
 #include <deal.II/base/config.h>
 
 #include <deal.II/base/exceptions.h>
@@ -107,8 +106,7 @@ DEAL_II_NAMESPACE_OPEN
  * @author Wolfgang Bangerth, 1997
  */
 template <int dim, typename Number = double>
-class Point : public Tensor<1, dim, Number>
-{
+class Point : public Tensor<1, dim, Number> {
 public:
   /**
    * Standard constructor. Creates an object that corresponds to the origin,
@@ -122,8 +120,7 @@ public:
   /**
    * Convert a tensor to a point.
    */
-  explicit DEAL_II_CUDA_HOST_DEV
-  Point(const Tensor<1, dim, Number> &);
+  explicit DEAL_II_CUDA_HOST_DEV Point(const Tensor<1, dim, Number> &);
 
   /**
    * Constructor for one dimensional points. This function is only implemented
@@ -133,8 +130,7 @@ public:
    *
    * @note This function can also be used in CUDA device code.
    */
-  explicit DEAL_II_CUDA_HOST_DEV
-  Point(const Number x);
+  explicit DEAL_II_CUDA_HOST_DEV Point(const Number x);
 
   /**
    * Constructor for two dimensional points. This function is only implemented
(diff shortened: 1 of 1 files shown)


include/deal.II/base/point.h
```diff
@@ -307,7 +307,7 @@ class Point : public Tensor<1, dim, Number>
 
   /**
    * Return the Euclidean distance of <tt>this</tt> point to the point
-   * <tt>p</tt>, i.e. the <tt>l_2</tt> norm of the difference between the
+   * <tt>p</tt>, i.e. the $\ell_2$ norm of the difference between the
```
bangerth: No, that's yet another space: The space of infinite dimensional vectors: https://en.wikipedia.org/wiki/Hilbert_space#Sequence_spaces. It needs to be `$l_2$`.

include/deal.II/base/point.h
```diff
@@ -185,7 +185,7 @@ class Point : public Tensor<1, dim, Number>
    * @note This function can also be used in CUDA device code.
    */
   DEAL_II_CUDA_HOST_DEV Number
-                        operator()(const unsigned int index) const;
+  operator()(const unsigned int index) const;
```
bangerth: I think that the indentation script wants it just the way it was.

krishnakumarg1984: This is really hard since my editor has been configured to indent on save following the .clang-format file in the project's root. I had to rename the .clang-format, work on my edit, but it wouldn't rebase until I had restored .clang-format.   

Why are the rules in the .clang-format file format provided in the project  not adhered to in the header files?

Rombur: I don't know what you mean. We apply clang-format everywhere. Which version of clang-format are you using? It needs to be 6.0.

krishnakumarg1984: The clang-format version compatibility can be specified in the `.clang-format` file. My binary is v9.0

bangerth: I think if you wanted to propose a patch for that, that would be fine -- but I have no trust in that actually working. I would be very surprised if the 9.0 binary with the 6.0 setting produces the same indentation as 6.0.

include/deal.II/base/point.h
```diff
@@ -107,8 +106,7 @@ DEAL_II_NAMESPACE_OPEN
  * @author Wolfgang Bangerth, 1997
  */
 template <int dim, typename Number = double>
-class Point : public Tensor<1, dim, Number>
-{
+class Point : public Tensor<1, dim, Number> {
```
bangerth: There are now a large number of unrelated indentations. Please run the indentation script. (If you use the `./contrib/utilities/download_clang_format` script, it will download the right version for you. You don't have to install or compile it yourself.)

krishnakumarg1984: I shall close this PR, work on a new branch and raise a new PR.

PR #9309, opened 2020-01-13
FESeries: Added features to precalculate all matrices and serialize them.

marcfehling:
Generating transformation matrices for Fourier and Legendre series expansions takes up a lot of time in general. We add the feature to precalculate and serialize these matrices in this PR.

I added unit tests for both Fourier and Legendre expansions, but haven't figured out an easy way how to verify the de-serialized matrices in these tests, since the transformation matrices are a private attribute.

Do you have any suggestion on how to do this without making large changes to the interface? I thought about a simple getter function for the matrix container, but I don't want to introduce this just for the purpose of testing the functionality when there is no other reason for it to exist...


marcfehling:
@davydden @bangerth -- FYI

include/deal.II/fe/fe_series.h
```diff
@@ -115,6 +115,24 @@ namespace FESeries
               const unsigned int            cell_active_fe_index,
               Table<dim, CoefficientType> & fourier_coefficients);
 
+    /**
+     * Calculate all transformation matrices to transfer the finite element
+     * solution to the series expansion representation.
+     *
+     * These matrices will be generated on demand by calling calculate().
+     * However with this function, all matrices will be calculated in advance.
```
bangerth: Can you say what the point of the function is? If `calculate()` calls it anyway, why expose it to the public interface?

include/deal.II/fe/fe_series.h
```diff
@@ -115,6 +115,24 @@ namespace FESeries
...
+     */
+    void
+    precalculate_all_transformation_matrices();
+
+    /**
+     * Write or read the all transformation matrices of this object to or from a
+     * stream for the purpose of serialization.
```
bangerth: same here -- say what the function can be used for or is typically used for.

include/deal.II/fe/fe_series.h
```diff
@@ -213,6 +231,24 @@ namespace FESeries
...
 
+    /**
+     * Calculate all transformation matrices to transfer the finite element
+     * solution to the series expansion representation.
+     *
+     * These matrices will be generated on demand by calling calculate().
+     * However with this function, all matrices will be calculated in advance.
+     */
+    void
+    precalculate_all_transformation_matrices();
```
bangerth: same comments here

include/deal.II/fe/fe_series.h
```diff
@@ -435,6 +471,28 @@ FESeries::process_coefficients(
...
 
 
+
+template <int dim, int spacedim>
+template <class Archive>
+inline void
+FESeries::Fourier<dim, spacedim>::serialize(Archive &ar,
+                                            const unsigned int /*version*/)
+{
+  ar &fourier_transform_matrices;
```
bangerth: is this the only member function of this class? In general, after deserialization you should have the exact same member function again, and it should compare equal.

source/fe/fe_series_fourier.cc
```diff
@@ -181,6 +181,20 @@ namespace FESeries
 
 
 
+  template <int dim, int spacedim>
+  void
+  Fourier<dim, spacedim>::precalculate_all_transformation_matrices()
+  {
+    for (unsigned int fe = 0; fe < fe_collection->size(); ++fe)
+      ensure_existence(*fe_collection,
```
bangerth: I think you could neatly use the following idiom here:
```
  Threads::TaskGroup<> task_group;
   for (unsigned int fe = 0; fe < fe_collection->size(); ++fe)
      task_group += Threads::new_task ([]() {ensure_existence(*fe_collection, ...}

  task_group.join_all();
```
to compute all of these things in parallel -- assuming that the `ensure_existence` serializes access to all member variables appropriately.

Doesn't have to happen for this patch, but keep it in mind :-)

source/fe/fe_series_legendre.cc
```diff
@@ -201,6 +201,17 @@ namespace FESeries
 
 
 
+  template <int dim, int spacedim>
+  void
+  Legendre<dim, spacedim>::precalculate_all_transformation_matrices()
+  {
+    for (unsigned int fe = 0; fe < fe_collection->size(); ++fe)
+      ensure_existence(
+        *fe_collection, *q_collection, N, fe, legendre_transform_matrices);
```
bangerth: same here

tests/serialization/fe_series_01.cc
```diff
@@ -0,0 +1,82 @@
...
+  oa << fourier_original;
+
+  // recover
+  FESeries::Fourier<dim> fourier_deserialized(max_degree + 1, hp_fe, hp_q);
+
+  std::istringstream            iss(oss.str());
+  boost::archive::text_iarchive ia(iss, boost::archive::no_header);
+  ia >> fourier_deserialized;
+
+  deallog << "OK" << std::endl;
```
bangerth: I think you need to compare for equality.

include/deal.II/fe/fe_series.h
marcfehling: This class stores `SmartPointer` objects to a `FECollection` and `QCollection` objects. I don't know how to deal with them.

There are also a container for all `k_vectors` in the Fourier class and one integer `N` in the Legendre class which will be set upon construction. I don't know to serialize them either. Shall I simply override those created on construction?

bangerth: As discussed, let's just save/restore the matrices and reflect that in the function names.

marcfehling: I'm preparing corresponding functions now. I've found a good basis for them in `grid/tria.h`, i.e. both functions `Triangulation::{save|load}()`.

However, to make this fail-proof, I'll need a `get_name()` function for the `Quadrature` classes as well, similar to the `FiniteElement` ones. I'll provide them in a separate PR. Do you agree, or do you have a different idea?

bangerth: That's one way. Alternatively, you could just output the quadrature points and weights as well, and upon reading stuff back in verify that they're still the same.

bangerth: An alternative would be to compute a has of the points and weights. `std::hash` could be overloaded for `Quadrature<dim>` (https://en.cppreference.com/w/cpp/utility/hash) or you could just implement a `Quadrature::hash()` function.

marcfehling:
`Quadrature` offers serialization, so I went for this option for now. This PR still needs a changelog, and I'll consider the `hash` variant later. But first, I need to grasp some sunlight which became so rare in Germany recently :)

marcfehling:
I've achieved everything I wanted with this PR. I'll stick to the `Quadrature::serialize` variant for now, since it stores all weights and quadrature points and is sufficient for the task. I'll may revisit this later and replace it with the corresponding hash functionality.

Ready for final review/tests/merge.

bangerth:
As a follow-up (let's get this merged as is): I don't think we need the version number for the save/restore functions. BOOST wants them for internal purposes if the functions are called `save` and `load`, but that's not applicable here, and we don't make use of the versioning anyway. So I would say let's just get rid of these arguments -- but like I say, let's merge the current version first.

bangerth:
The CI check failed, but it's not clear why: it only shows that the linker exited with error code 1. Since the other tests all succeeded, I'm going to assume that this was a fluke and merge anyway.

PR #9318, opened 2020-01-14
converts l_2 to LaTeX code

### LINKED ISSUE #9302: In all documentation pages, consider replacing l_2 with its LaTeX equivalent notation

krishnakumarg1984:
Fixes #9302

### DIFF
--- a/include/deal.II/base/point.h
+++ b/include/deal.II/base/point.h
@@ -307,7 +307,7 @@ public:
 
   /**
    * Return the Euclidean distance of <tt>this</tt> point to the point
-   * <tt>p</tt>, i.e. the <tt>l_2</tt> norm of the difference between the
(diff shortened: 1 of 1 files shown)


bangerth:
I don't think that this PR actually needs to be tested, but the mandatory status checks seem to be stuck...

masterleinad:
@bangerth You can just try to restart the `Travis` check manually.

PR #9330, opened 2020-01-15
In glossary for coarse_mesh, use monospace font for refine_global call since doxygen wasn't already hyperlinking it anyway

krishnakumarg1984:


doc/doxygen/headers/glossary.h
```diff
@@ -407,7 +407,7 @@
  *
  *   Triangulation objects store cells in <i>levels</i>: in
  *   particular, all cells of a coarse mesh are on level zero. Their
- *   children (if we executed Triangulation::refine_global(1) on a
+ *   children (if we executed <tt>Triangulation::refine_global(1)>></tt> on a
```
bangerth: It should have been
```suggestion
 *   children (if we executed <tt>Triangulation::refine_global(1)</tt> on a
```
but really, the better markup is to not say the font name, but the kind of thing you have: a code snippet:
```suggestion
 *   children (if we executed <code>Triangulation::refine_global(1)</code> on a
```
This can more easily be marked up via
```suggestion
 *   children (if we executed `Triangulation::refine_global(1)` on a
```

krishnakumarg1984:
Changes have been made

PR #9335, opened 2020-01-15
Simplify description in step-49.

### LINKED ISSUE #9334: [Step-49] Unclear wording on the Transform Operation used
In Step-49, it says

>GridTools::transform takes a triangulation and any kind of object that can be called like a function as arguments. This function-like argument can be the address of a function that takes a point and returns a point, an object that has an __operator() like the code below__, or for example, a std::function<Point<2>(const Point<2>)> object one can get as a lambda function. Here we choose the __latter__ option.

From this paragraph, It is not clear whether they mean using the `operator()` approach or the `lambda function` approach. My C++ knowledge is not enough to understand this from the code given right after this above paragraph. Can the maintainers edit this tutorial suitably, if required?

bangerth:
Fixes #9334.

### DIFF
--- a/examples/step-49/step-49.cc
+++ b/examples/step-49/step-49.cc
@@ -226,12 +226,17 @@ void grid_4()
 // and returns a mapped point. In this case, we transform $(x,y) \mapsto
 // (x,y+\sin(\pi x/5))$.
 //
-// GridTools::transform takes a triangulation and any kind of object that can
-// be called like a function as arguments. This function-like argument can be
-// the address of a function that takes a point and returns a point, an object
-// that has an <code>operator()</code> like the code below, or for example, a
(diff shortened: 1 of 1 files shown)


tjhei:
```

/Users/runner/runners/2.163.1/work/dealii/dealii/examples/step-49/step-49.cc:258:73: error: expected ')'
      return Point<2>(in[0], in[1] + std::sin(in[0] / 5.0 * numbers::PI);
```

bangerth:
oops -- fixed now.

drwells:
It looks like this PR also acquired conflicts. It’s nice that we have so many improvements coming in to the tutorials :)

bangerth:
Updated as well.

PR #9343, opened 2020-01-15
Compute diagonal with MatrixFree

peterrum:
This PR adds a new function to compute the diagonal of a system matrix in matrix-free manner for systems with constraints.

I still have to move the function `MF::compute_diagonal` from the test to the right place and generalize it for vector quantities. Also, I am not that happy with the interface yet...

Ping @kronbichler @lpsaavedra @tjhei @tcclevenger


peterrum:
@kronbichler The code now works also for vector quantities. The test @tcclevenger gave me in summer also passes (see tests/matrix_free/compute_diagonal_02.cc). I still need to get block vectors working (any suggestions for a setup so that I reach this line https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/fe_evaluation.h#L4071 ?) and to extend the tests for parallel runs.

peterrum:
Ready to be reviewed. We still need to decide on an appropriate place for the new method and an appropriate interface. As it is now it is somewhat clumpy... Maybe it might be the right time to introduce a *kernel-container*, which stores a reference to an instance of `MatrixFree` and `FEEvaluation` and has for CG a `cell_integral` method.

peterrum:
@kronbichler: @lpsaavedra has copied the code into her test multigrid code and it seems to work as expected!

peterrum:
>The algorithm looks correct to me. The question is how we want to proceed. I guess there would be value to integrate the content of the compute_diagonal function into the library, e.g. by a free function that takes a local_vmult() operation or similar. 

I would propose to introduce a matrix-free utility namespace. We could collect there `MatixFree`-related functions. The diagonal function could be the first function in that namespace, but I think there are many functions developed in our group, which would fit into that namespace (e.g. a utility function to setup the data structures for vector-version of `MatrixFree::reinit`).

Maybe, we could agree on a namespace name and a file name so that I can move the new function there and the review process could be started. 

Regarding the interface, I am open. Maybe other users of `MatrixFree` users could give some input!

> Another question would be if it makes sense to put some of the stuff you compute inside that function to the MatrixFree::reinit() function - but that of course depends on how expensive these operations really are and how often we anticipate that this function will be called. If the typical user calls it once, it makes little sense.

This strongly depends on the use case. If one wants to recompute the diagonal for the same constraint matrix, it might be useful to *cache* the information.

Anyway, I will make some performance tests in the next day, comparing the *new* way to compute the matrix diagonal and the naive way (which does not work for all constraint types) and then we will see how expensive it is to compute the transpose of the local constraint matrix on the the fly.

peterrum:
Lets postpone this PR after the release. First experiments showed that the code is about 3 times slower than the simple implementation. I would like to improve this!

include/deal.II/matrix_free/tools.h
```diff
@@ -27,8 +30,369 @@ DEAL_II_NAMESPACE_OPEN
...
+  compute_diagonal(
+    const MatrixFree<dim, Number, VectorizedArrayType> &matrix_free,
+    FEEvaluation<dim,
+                 fe_degree,
+                 n_q_points_1d,
+                 n_components,
+                 Number,
+                 VectorizedArrayType> &                 phi,
+    LinearAlgebra::distributed::Vector<Number> &        diagonal_global,
+    const std::function<void()> &                       local_vmult)
```
tcclevenger, line 62: Can we add input argument of an unsigned int here and pass the cell index below? I need to query a coefficient table in my `local_vmult()`. 
```suggestion
    const std::function<void(const unsigned int)>&                       local_vmult)
```

include/deal.II/matrix_free/tools.h
```diff
@@ -27,8 +30,369 @@ DEAL_II_NAMESPACE_OPEN
...
+            for (unsigned int i = 0; i < phi.dofs_per_cell; ++i)
+              {
+                // compute i-th column of element stiffness matrix:
+                // this could be simply performed as done at the moment with
+                // matrix-free operator evaluation applied to a ith-basis vector
+
+                for (unsigned int j = 0; j < phi.dofs_per_cell; ++j)
+                  phi.begin_dof_values()[j] = static_cast<Number>(i == j);
+
+                local_vmult();
```
tcclevenger, line 339:
```suggestion
                local_vmult(cell);
```

tests/matrix_free/compute_diagonal_util.h
```diff
@@ -0,0 +1,150 @@
...
+                   n_components,
+                   Number,
+                   VectorizedArrayType>
+        phi(matrix_free);
+
+      VectorType diagonal_global;
+      MatrixFreeTools::compute_diagonal(matrix_free,
+                                        phi,
+                                        diagonal_global,
+                                        [&]() { this->cell_operation(phi); });
```
tcclevenger, line 86:
```suggestion
                                        [&](const unsigned int /*cell*/) { this->cell_operation(phi); });
```

include/deal.II/matrix_free/tools.h
peterrum: @tcclevenger Sorry for the late reply! Thanks for testing the code in ASPECT. I don't think we need the additional argument. A new function has been introduced recently, s.t. it is possible to get the index from `FEEvaluation`:

https://github.com/dealii/dealii/blob/040c6ad7d46f1291ab5c038b70cf1da4fb22b006/include/deal.II/matrix_free/fe_evaluation.h#L237-L241

That should also work, shouldn't it?

peterrum:
I have extended the documentation. Furthermore, I have modified the interface of the function since it was not thread safe. Now a new FEEvaluation is created within the matrix-free loop.

include/deal.II/matrix_free/tools.h
```diff
@@ -119,8 +146,381 @@ namespace MatrixFreeTools
...
   }
-} // namespace MatrixFreeTools
 
+  namespace internal
+  {
+    template <typename Number>
+    struct LocalCSR
+    {
+      std::vector<unsigned int> row_lid_to_gid;
+      std::vector<unsigned int> row{0};
```
kronbichler, line 156: Can you remove the initialization from the struct declaration and initialize at first use instead? Alternatively use a constructor.

include/deal.II/matrix_free/tools.h
```diff
@@ -119,8 +146,381 @@ namespace MatrixFreeTools
...
+
+    // initialize vector
+    matrix_free.initialize_dof_vector(diagonal_global);
+
+    int dummy;
+
+    matrix_free.template cell_loop<VectorType, int>(
+      [&](const MatrixFree<dim, Number, VectorizedArrayType> &matrix_free,
+          LinearAlgebra::distributed::Vector<Number> &        diagonal_global,
+          const LinearAlgebra::distributed::Vector<Number> &,
```
kronbichler, line 192: shouldn't this argument be
```suggestion
          const int &,
```
to reflect the template argument?

include/deal.II/matrix_free/tools.h
```diff
@@ -119,8 +146,381 @@ namespace MatrixFreeTools
...
+        // local storage: buffer so that we access the global vector once
+        // note: may be larger then dofs_per_cell in the presence of
+        // constraints!
+        std::array<std::vector<Number>, n_lanes> diagonals_local_constrained;
+
+        for (unsigned int cell = range.first; cell < range.second; ++cell)
+          {
+            /*****************************************************************
+             * STEP 1: get relevant information from FEEvaluation
+             ****************************************************************/
```
kronbichler, line 213: Not that I feel strongly about it, but I think we typically do not use this heavy-weight separation within a single function in many other places. I think a simple
```suggestion
            // step 1: get relevant information from FEEvaluation
```
would suffice (similarly below).

include/deal.II/matrix_free/tools.h
```diff
@@ -119,8 +146,381 @@ namespace MatrixFreeTools
...
+                next_index_indicators =
+                  dof_info
+                    .row_starts[(cell * n_lanes + v) * n_fe_components +
+                                first_selected_component + 1]
+                    .second;
+
+                /***************************************************************
+                 * STEP 2a: setup locally-relevant constraint matrix in a
+                 *   coordinate list (COO - see:
+                 *   https://en.wikipedia.org/wiki/Sparse_matrix#Coordinate_list_(COO))
```
kronbichler, line 260: I do not think we need a link to COO here. One either understands coordinate list/COO or does not mess around in this code I believe.

include/deal.II/matrix_free/tools.h
```diff
@@ -119,8 +146,381 @@ namespace MatrixFreeTools
...
+                                           c_pool.col.begin() +
+                                             c_pool.row[j + 1],
+                                           i);
+
+                        // explanation: j-th row of C_e^T is empty (see above)
+                        if (scale_iterator ==
+                            c_pool.col.begin() + c_pool.row[j + 1])
+                          continue;
+
+                        // explanation: C_e^T(j,i) is is zero (see above)
```
kronbichler, line 488:
```suggestion
                        // explanation: C_e^T(j,i) is zero (see above)
```

include/deal.II/matrix_free/tools.h
```diff
@@ -119,8 +146,381 @@ namespace MatrixFreeTools
...
+
+            // loop over all columns of element stiffness matrix
+            for (unsigned int i = 0; i < phi.dofs_per_cell; ++i)
+              {
+                // compute i-th column of element stiffness matrix:
+                // this could be simply performed as done at the moment with
+                // matrix-free operator evaluation applied to a ith-basis vector
+
+                for (unsigned int j = 0; j < phi.dofs_per_cell; ++j)
+                  phi.begin_dof_values()[j] = static_cast<Number>(i == j);
```
kronbichler, line 461: I would be curious how the compiler translates this code. :wink:

kronbichler (approved):
Looks good in general.

kronbichler:
Regarding the old question whether to integrate the transposition in the matrix-free setup and
> Anyway, I will make some performance tests in the next day, comparing the new way to compute the matrix diagonal and the naive way (which does not work for all constraint types) and then we will see how expensive it is to compute the transpose of the local constraint matrix on the the fly.

I think we can postpone that part to see how much it affects run time.

include/deal.II/matrix_free/tools.h
```diff
@@ -119,8 +146,381 @@ namespace MatrixFreeTools
...
+              for (unsigned int v = 0; v < n_lanes_filled; ++v)
+                dof_indices[v] =
+                  dof_info.dof_indices.data() +
+                  dof_info
+                    .row_starts[(cell * n_lanes + v) * n_fe_components +
+                                first_selected_component]
+                    .first;
+
+              for (unsigned int v = n_lanes_filled; v < n_lanes; ++v)
+                dof_indices[v] = nullptr;
```
tjhei, line 233: you can just initialize ``dof_indices`` to 0 above and skip this.

peterrum:
@kronbichler @tjhei Thanks for your feedback! I have addressed the comments.

kronbichler (approved):
Thanks for the adjustments.

peterrum:
Thanks! As a follow-up PR I will have a look at block systems.

PR #9351, opened 2020-01-16
Remove trailing whitespace

masterleinad:
This is for all `CMakeLists.txt`, `*.cc`, `*.h`, `*.html` and `*.dox` files in `include`, `source`, `doc`, `contrib` and `examples`.


masterleinad:
> What command did you use to do this check?

I used 
```
find source include/ examples/ contrib/ doc -type f \( -name "CMakeLists.txt" -o -name "*.cc" -o -name "*.h" -o -name "*.html" -o -name "*.dox" \) -exec sed -i "s/\s\+$//g" {} \;
```

masterleinad:
I found some more after rebasing to `master` and force-pushed.

bangerth:
Independent from this PR, do you want to make that script part of `contrib/utilities/indent`?

tjhei:
> Independent from this PR, do you want to make that script part of `contrib/utilities/indent`?

:+1: That's what I was thinking about as well.

masterleinad:
Alright, I added a script (and forced-pushed again due to more trailing whitespaces after rebasing).

contrib/utilities/indent_common.sh
```diff
@@ -162,6 +162,16 @@ format_file()
 }
 export -f format_file
 
+#
+# Remove trailing whitespace
+#
+
+remove_trailing_whitespace()
+{
+  sed -i 's/\s\+$//g' $1
```
drwells: oops: `sed -i` is not portable!

PR #9359, opened 2020-01-17
Work around a doxygen bug that triggers in step-10.

### LINKED ISSUE #9357: Weird doxygen behaviour for Step-10 documentation direct link to Plain Program in the browser
This is a really weird one.  All the tutorials so far have correct hyperlinking of the doxygen-generated pages rendered in a browser.

However, for Step-10, the direct link to the `Plain Program` does not work in both the

- master version (https://dealii.org/developer/doxygen/deal.II/step_10.html#PlainProg)
- v9.1.1 version (https://dealii.org/current/doxygen/deal.II/step_10.html#PlainProg)

Puzzlingly, the problem is only with step-10. I have tested this with Steps 1 through 9 as well as 11, and they all correctly fly to the Plain Program directly on both the master as well as the latest released branches. I could not figure out why this is happening!

bangerth:
Fixes #9357. Description is in the patch itself.

### DIFF
--- a/doc/doxygen/scripts/make_step.pl
+++ b/doc/doxygen/scripts/make_step.pl
@@ -87,6 +87,18 @@ print
 
 system $^X, "$cmake_source_dir/doc/doxygen/scripts/create_anchors", "$cmake_source_dir/examples/$step/doc/intro.dox";
 
+
+# Start the commented program by writing two empty lines. We have had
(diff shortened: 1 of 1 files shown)


tjhei:
ahh, doxygen. So much to dislike about it.

No harm in adding some newlines, so good to go.

PR #9360, opened 2020-01-17
Host of documentation improvements to step-10

krishnakumarg1984:
In step-10, clarify wording (or slightly better wording) on the constant function of unit magnitude

changes const to constexpr which is technically better C++

Adds exception for unused space dimensions functionality in Step-10 tutorial

fixes monospace font for JxW

In step-49, use a variable dim within the try scope of main() for dimension

use monospace font for active_cell_iterators in step-10

monospace for subscription cells in step-10

monospace for long double documentation in step-10

monospace font for the variable pi

monospace font for the variable area

fixes doc shortcoming about static cast to double, in step-10

Revert "fixes doc shortcoming about static cast to double, in step-10"

This reverts commit 99a207d6e9a60271abef3d73cc3c462d20e85cb8.

monospace font fixes

monospace font fixes

fixes a typo in results section

step-10 results section gnuplot lw monospace not rendered

step-10 results section gnuplot lt monospace not rendered


examples/step-10/step-10.cc
```diff
@@ -54,7 +55,7 @@ namespace Step10
   // computed approximation to $\pi$ being in the range of the number of
   // digits which a double variable can hold, we rather declare the reference
   // value as a <code>long double</code> and give it a number of extra digits:
-  const long double pi = 3.141592653589793238462643L;
+  constexpr long double pi = 3.141592653589793238462643L;
```
masterleinad: Didn't you want to get rid of `long double` here? What is the point of making it `constexpr`? The right-hand-side is a constant already.

examples/step-10/step-10.cc
```diff
@@ -64,14 +65,19 @@ namespace Step10
   // of the library). Rather, we just pack the functionality into separate
   // functions. We make these functions templates on the number of space
   // dimensions to conform to usual practice when using deal.II, although we
-  // will only use them for two space dimensions.
+  // will only use them for two space dimensions and throw an exception when
+  // attempted to use for any other spatial dimension
```
masterleinad:
```suggestion
  // attempted to use for any other spatial dimension.
```

examples/step-10/step-10.cc
```diff
@@ -64,14 +65,19 @@ namespace Step10
...
   // The first of these functions just generates a triangulation of a circle
   // (hyperball) and outputs the $Q_p$ mapping of its cells for different values
   // of <code>p</code>. Then, we refine the grid once and do so again.
   template <int dim>
   void gnuplot_output()
   {
+    Assert(
+      dim == 2,
+      ExcMessage(
+        "This tutorial is currently implemented only for two spatial dimensions "));
```
masterleinad: Doesn't this particular function also make sense for 3d?

examples/step-10/step-10.cc
```diff
@@ -399,10 +405,12 @@ int main()
     {
       std::cout.precision(16);
 
-      Step10::gnuplot_output<2>();
+      unsigned int dim = 2;
```
masterleinad: This must be a compile time constant. In other words,
```suggestion
      constexpr unsigned int dim = 2;
```
would work.

examples/step-10/step-10.cc
```diff
@@ -32,6 +32,7 @@
 #include <deal.II/dofs/dof_accessor.h>
 #include <deal.II/fe/fe_nothing.h>
 #include <deal.II/fe/fe_values.h>
+#include <deal.II/base/exceptions.h>
```
masterleinad: You likely don't need this include file at all. In the end, `step-51` is the only one that currently includes it.

examples/step-10/step-10.cc
krishnakumarg1984: @masterleinad  

1. Yes, I shall get rid of the `long double` and update the results to close the other issue once this PR is merged. There were already too many changes in this PR, and this is not linked to any issue. These are general documentation fixes to step-10.

2. `constexpr` guarantees compile-time allocation/optimisation. This exact issue (using `constexpr` for the value of pi and why `constexpr` is better is answered [here](https://stackoverflow.com/questions/13346879/const-vs-constexpr-on-variables)

examples/step-10/step-10.cc
krishnakumarg1984: Yes, it does. However, this particular tutorial introduction claims 
>We make these functions templates on the number of space dimensions to conform to usual practice when using deal.II, although we will only use them for two space dimensions.

And, all the results are provided only for 2 spatial dimensions (circle). And in fact, this is about calculating the value of pi.

I made this tutorial exactly replicating the line of explanation given in tutorial 5 about how to use Assertions.

> 
> Assert(dim == 2, ExcInternalError());
> 
> ExcInternalError is a globally defined exception, which may be thrown whenever something is terribly wrong. Usually, one would like to use more specific exceptions, and particular in this case one would of course try to do something else if dim is not equal to two, e.g. create a grid using library functions. Aborting a program is usually not a good idea and assertions should really only be used for exceptional cases which should not occur, but might due to stupidity of the programmer, user, or someone else. The situation above is not a very clever use of Assert, but again: this is a tutorial and it might be worth to show what not to do, after all.
> 
> So if we got past the assertion, we know that dim==2, and we can now actually read the grid. It is in UCD (unstructured cell data) format (though the convention is to use the suffix inp for UCD files):
> grid_in.read_ucd(input_file);

examples/step-10/step-10.cc
krishnakumarg1984: Yes, it looks like you are right. Step-5 has not included this. I referred to the `Assert` macro, and it said that it was defined in line `xx` in `exceptions.h`

examples/step-10/step-10.cc
masterleinad: Since we are not using it in any `constexpr` context here, there is essentially no difference at all and I don't see a good reason for changing it. 
If you are going to make it `double`, there is no reason to define the variable at all. You can just use `numbers::PI` instead.

krishnakumarg1984: Ok. Let us just restore what we originally had. I will study `numbers::PI` later on and issue a PR later on.

examples/step-10/step-10.cc
krishnakumarg1984: @masterleinad Are you happy with this? Shall we mark this as resolved?

examples/step-10/step-10.cc
bangerth: I'd just remove the `#include`. It's one of those files that is included in essentially every other header file.

bangerth:
@krishnakumarg1984 -- what's the status with your patch here? I think there were really only a couple open issues, the rest looked good!

drwells:
@krishnakumarg1984 I'd like to see this done. Will you have some time to wrap things up this week?

examples/step-10/step-10.cc
krishnakumarg1984: Since there was no feedback, I will remove this assertion from the commit.

examples/step-10/step-10.cc
krishnakumarg1984: Ok. Got it.

examples/step-10/step-10.cc
```diff
@@ -315,7 +321,7 @@ namespace Step10
     std::cout << "Computation of Pi by the perimeter:" << std::endl
               << "===================================" << std::endl;
```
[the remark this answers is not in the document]
krishnakumarg1984: I removed the assertion owing to lack of feedback (I understand the time pressures of the moderators.  I wasn't active for a few weeks myself and was trying to get through video lectures amid other issues)

krishnakumarg1984:
@drwells  Do let me know if the PR as it stands now is acceptable. Apologies for the long delay.

masterleinad:
Hmm... that looks like you tried merging instead of rebasing. Does
```
git rebase upstream/master
```
help to clean the commit history you are seeing?

drwells:
@krishnakumarg1984 Can you rebase this branch, please?

krishnakumarg1984:
Currently, I am out of time to do this, but intend to return to this sometime in the future. Apologies.

bangerth:
Superseded by #11014.

Issue #9353, opened 2020-01-16
Possibility to account for material_id during p::fd::t creation


agrayver:
I need to create a p::fd::t that, along with the geometrical information, also inherits material_id's for all active cells from the base triangulation. As far as I can see, there is no existing mechanism for that. Would something like that in the user work?

```
  for (auto &cell: tria_pft.active_cell_iterators())
  {
    CellId id = cell->id();
    auto cell_base = id.to_cell(basetria);
    cell->set_material_id(cell_base->material_id());
  }
```
@peterrum I guess you're the best person to comment on this.

peterrum:
@agrayver Could you describe the problem you want to solve more in detail?

Are you using one of the utility functions from the `parallel::fullydistributed::Utilities` namespace? If yes, then the `material_id` should be automatically copied from the base class. See https://www.dealii.org/developer/doxygen/deal.II/structparallel_1_1fullydistributed_1_1CellData.html

Regrading your code. I am not 100% sure if the cell ids are the same in `base_tria` and the  `tria_pft`. The coarse-cell id might change (and as a consequence also the CellId). But with high probability, it does not change, so that your code might work.

peterrum:
I will try out later the `CellId`s and let you  know!

agrayver:
> Are you using one of the utility functions from the parallel::fullydistributed::Utilities namespace? 

Yes, pretty much follow *copy_serial_tria_01.cc*

> If yes, then the material_id should be automatically copied from the base class. See https://www.dealii.org/developer/doxygen/deal.II/structparallel_1_1fullydistributed_1_1CellData.html

What exactly are you referring to here?

peterrum:
Sorry, I was confused and noticed an error in the type:
```cpp
       types::material_id manifold_id;
```
This should be `types::manifold_id manifold_id`. This means that `material_id` is not copied...

peterrum:
@agrayver Your approach should work. I have created the following test program (I will create a separate PR to add this test).

```cpp
#include <deal.II/base/mpi.h>

#include <deal.II/distributed/fully_distributed_tria.h>
#include <deal.II/distributed/fully_distributed_tria_util.h>
#include <deal.II/distributed/shared_tria.h>
#include <deal.II/distributed/tria.h>

#include <deal.II/fe/fe_q.h>

#include <deal.II/grid/grid_generator.h>
#include <deal.II/grid/grid_out.h>
#include <deal.II/grid/grid_tools.h>
[...]

PR #9362, opened 2020-01-17
New p:f:t tests

peterrum:
... tests to ensure that the order of pft and be the base triangulations are the same...

related to PR #9353 

FYI @agrayver


agrayver:
Thanks for checking. I closed the issue.

tests/fullydistributed_grids/copy_distributed_tria_03.with_p4est=true.mpirun=5.output
```diff
@@ -0,0 +1,9 @@
+
+
+
+
```
bangerth: Your test doesn't actually produce any output. Where do these empty lines come from?

tests/fullydistributed_grids/copy_serial_tria_08.mpirun=5.output
```diff
@@ -0,0 +1,9 @@
+
+
+
+
```
bangerth: same question

tests/fullydistributed_grids/copy_distributed_tria_03.with_p4est=true.mpirun=5.output
peterrum: I think each process prints one empty line per default.

bangerth: I see.

kronbichler: If you don't need the output of every process, and it appears you don't, replace `MPILogInitAll all;` by `mpi_initlog();`

peterrum:
I have updated the tests!

PR #9366, opened 2020-01-17
In step-10, fixes explanation surrounding new header inclusion. fixes…

### LINKED ISSUE #9356: [Step-10] Potentially need a documentation line explaining why the header file fe_nothing was introduced
Step-10 is the first tutorial wherein the `FENothing` class is introduced.

In the spirit of all other tutorials, the inclusion of the corresponding header file warrants a line of explanation, and probably cannot be glossed over as currently done here. Also, the statement claiming that `MappingQ` is the only one new is false.

__________________________________________
>This is the only new one: in it, we declare the MappingQ class which we will use for polynomial mappings of arbitrary order:
>`#include <deal.II/fe/mapping_q.h>`

>The first of the following include files are probably well-known by now and need no further explanation.
>
```
#include <deal.II/base/quadrature_lib.h>
...
bunch of include files
...
__#include <deal.II/fe/fe_nothing.h>__
#include <deal.II/fe/fe_values.h>
```

krishnakumarg1984:
Fixes #9356

### DIFF
--- a/examples/step-10/step-10.cc
+++ b/examples/step-10/step-10.cc
@@ -30,10 +30,14 @@
 #include <deal.II/grid/grid_out.h>
 #include <deal.II/dofs/dof_handler.h>
 #include <deal.II/dofs/dof_accessor.h>
-#include <deal.II/fe/fe_nothing.h>
 #include <deal.II/fe/fe_values.h>
 
-// This is the only new one: in it, we declare the MappingQ class
+// This include file is new. Even if we are not solving a PDE in this tutorial,
+// we want to use a dummy finite element with zero degrees of freedoms provided
+// by the FE_Nothing class.
+#include <deal.II/fe/fe_nothing.h>
+
+// The following header file is also new: in it, we declare the MappingQ class
 // which we will use for polynomial mappings of arbitrary order:
 #include <deal.II/fe/mapping_q.h>
 


examples/step-10/step-10.cc
```diff
@@ -30,10 +30,14 @@
 #include <deal.II/grid/grid_out.h>
 #include <deal.II/dofs/dof_handler.h>
 #include <deal.II/dofs/dof_accessor.h>
-#include <deal.II/fe/fe_nothing.h>
 #include <deal.II/fe/fe_values.h>
 
-// This is the only new one: in it, we declare the MappingQ class
+// This one is new. In this tutorial, since we are not solving a PDE, we will
+// use include the below header file which provides the dummy class FE_Nothing
+// that has zero degrees of freedom (this will be explained later)
```
masterleinad: This sounds rather convoluted. What about:
> // This include file is new. Even if we are not solving a PDE in this tutorial, we want to
// use a dummy finite element with zero degrees of freedoms provided by the FE_Nothing class.

krishnakumarg1984: Perfect. I will update the PR by committing these changes.

krishnakumarg1984: Done.

bangerth:
Do you think you can squash these two commits?

krishnakumarg1984:
Done!

PR #9342, opened 2020-01-15
Use the same color edges as for nodes in the tutorial graph.

bangerth:
I find that this makes the graph a bit easier to follow. I had to made the fluid dynamics yellow a bit darker for that, though.

PR #9370, opened 2020-01-17
More work on the tutorial graph

bangerth:
This is a follow-up to #9342. It changes two things:
* It gives greater weight to edges between programs of the same kind. This keeps them closer together and leads to better grouping. It gives heavy weight between "basic" tutorial programs, and they are now all arranged in a nearly vertical fashion.
* It changes the blue color for one category to a lighter blue, as requested by @tjhei in https://github.com/dealii/dealii/pull/9342#pullrequestreview-343512307 This makes the black text much easier to read.

### DIFF
--- a/doc/doxygen/scripts/steps.pl
+++ b/doc/doxygen/scripts/steps.pl
@@ -27,13 +27,14 @@ while (my $line = <TUTORIAL>)
 }
 
 # List of additional node and edge attributes to highlight purpose and state of
-# a tutorial or code gallery program
+# a tutorial or code gallery program. For a list of colors, take a look here:
+#   https://www.graphviz.org/doc/info/colors.html
 my %colors = (
  "basic"          => 'green',
  "techniques"     => 'orange',
  "fluids"         => 'yellow2',
  "solids"         => 'lightblue',
- "time dependent" => 'blue',
+ "time dependent" => 'dodgerblue1',
  "unfinished"     => 'black',
  "code-gallery"   => 'black',
     );
--- a/doc/doxygen/scripts/steps.pl
+++ b/doc/doxygen/scripts/steps.pl
@@ -168,10 +168,23 @@ foreach $step (@ARGV)
         # Determine the style of the arrow that connects
         # the two nodes. If the two nodes are of the same
         # kind, use the same color as the nodes as this makes
-        # reading the flow of the graph a bit easier.
+        # reading the flow of the graph a bit easier. Furthermore,
+        # set the edge weight to 5 (instead of the default of 1)
+        # to try and keep programs of the same kind together. The
+        # exception is the "basic" tutorial programs: these are
+        # going to be connected by edges of weight 100, ensuring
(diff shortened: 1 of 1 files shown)


tjhei:
great! Would you mind going to something like orchid1 or peachpuff instead to avoid the two blues looking too similar?

bangerth:
I played with colors for a good long while, and this looked the best to me in the big picture.

bangerth:
By the way, this is how it looks:
image

bangerth:
I tend to think that this makes following the flow of tutorials much easier to read!

bangerth:
Compare to before:
image

tjhei:
certainly much better, I agree.

Is there a way to make the basic tutorials lined up from left to right at the top? This might help even further.

tjhei:
But I am okay to merge as is. One step at a time. :-)

tjhei:
But I am okay to merge as is. One step at a time. :-)

bangerth:
No, we can't do that -- I tried. The directed graph plots things in layers depending on their dependencies, and it does not accept the `pos` argument -- something I tried because the documentation says that it does, but the internet reports that it doesn't work.

tjhei:
okay, merge yourself if you are happy with this.

bangerth:
Just for completeness, I played some more with graphviz and in particular tried what would happen if one had edges going both `step1 -> step2` and `step2 -> step1` (the latter edge could be made invisible), in the hope that in that case graphviz would rank them at the same level and place them horizontally. But it doesn't -- it just puts that node on top that is declared first...

bangerth:
I did find a way to make some nodes horizontal: `{rank=same; Step1, Step2, Step3, Step4}`. But I don't like the results -- it gets very wide, and the flow isn't as good as I want it to. It looks even worse if one puts all seven basic programs horizontally.

image

krishnakumarg1984:
graphviz/dot has features to accept constraints to make the graph not exceed a certain width whilst trying to satisfy specified position, whilst keeping the overall number of cross-overs minimal

bangerth:
Hm, I didn't find these. But I'm attaching the input for dot if you wanted to play!

[x.dot.txt](https://github.com/dealii/dealii/files/4078967/x.dot.txt)

krishnakumarg1984:
If it is okay with you, can you please leave this as an exercise to me (the width-constraint fitting), and I shall update this next week?

bangerth:
Gladly :-) The command I'm using to convert the text file to a pdf is simply `dot x.dot.txt -Tpdf > x.pdf`.

When you play with the file, remember that it's automatically generated. So whatever changes you want to make, we eventually have to put back into the `doc/doxygen/scripts/steps.pl` file.

krishnakumarg1984:
@bangerth Ah. I think I understood the issue (maybe).

The node position attribute "pos" Only works with neato or fdp layout engines, not dot.

https://stackoverflow.com/questions/5343899/how-to-force-node-position-x-and-y-in-graphviz

PR #9371, opened 2020-01-17
New compress/decompress using gzip.

luca-heltai:


drwells:
I don't remember the details, but I do recall talking about this at the workshop; the consensus was that the boost implementation is several times slower than just calling zlib directly. Does anyone remember the results? @luca-heltai would you be willing to check the performance of this versus direct calls to `compress2` before proceeding?

PR #9372, opened 2020-01-17
Fix removing trailing whitespaces

### LINKED ISSUE #9368: /contrib/utilties/indent script is broken for latest update of GNU sed
`/contrib/utilties/indent` has just broken leading to errors like __invalid character etc.__ with the latest update of GNU sed `v4.8`.

Using GNU-specific flags could be problematic, since different contributors will be on different platforms. Perhaps relying only on POSIX flags is the answer?

masterleinad:
Fixes #9368. It would be good if someone using Mac OSX could check.

### DIFF
--- a/contrib/utilities/indent_common.sh
+++ b/contrib/utilities/indent_common.sh
@@ -163,12 +163,16 @@ format_file()
 export -f format_file
 
 #
-# Remove trailing whitespace
+# Remove trailiing whitespace. Mac OSX requires an extension for a backup file
+# for in-place replacements. So we need to provide something before the regex.
+# Using '-e' avoids creating these files on GNU platforms at least.
+# For Mac OSX, we still need to delete the created file.
 #
 
 remove_trailing_whitespace()
 {
-  sed -i 's/\s\+$//g' $1
+  sed -i -e 's/\s\+$//g' "$1"
+  rm -f "$1-e"
 }
 export -f remove_trailing_whitespace
 


krishnakumarg1984:
Works for me on MacOS catalina.

contrib/utilities/indent_common.sh
```diff
@@ -163,12 +163,16 @@ format_file()
 export -f format_file
 
 #
-# Remove trailing whitespace
+# Remove trailiing whitespace. Mac OSX requires an extension for a backup file
```
tjhei:
```suggestion
# Remove trailing whitespace. Mac OSX requires an extension for a backup file
```

tjhei:
this is somewhat ugly, but as it seems to work, so what. :man_shrugging:

bangerth:
Why not just use perl for this purpose then?

drwells:
I don't think that we should require perl for `make indent`, but otherwise I would agree (its almost always the right replacement for sed and awk).

tjhei:
I am merging so we don't break everyone who tries to indent on OSX right now. If anybody has a better solution, follow it up with a second PR.

krishnakumarg1984:
`perl` is not installed by default on windows. Well, so are `sed` and `awk`, but they are available from literally 10 different places. Whereas perl for windows is a bloated install of active perl or strawberry perl. If we can afford it, it is best to stick to POSIX compatible *nix tools.

bangerth:
I thought we required perl for the build process anyway. But I don't care enough one way or the other :-)

krishnakumarg1984:
@bangerth Ya..But sometimes I push from a laptop or a library machine, where I don't have a full-blown deal.ii installed (leave alone `perl`). 

 `sed` and `awk` are so tiny that we can get them as part of `mobaxterm` or `scoop` for example which don't even require admin priveleges for library machines.  Easy for a documentation-related fix, which doen't involve a deal.Ii build in any way whatsoever

drwells:
@krishnakumarg1984 Virtually none of our utilities work on windows and are never going to work on windows unless someone is willing to rewrite all of the shell scripts. Its just not a priority for the main developers.

@bangerth I am pretty sure we can compile the library without perl; we do need perl for documentation though.

krishnakumarg1984:
@drwells  I am talking about a simple documentation fix followed by an indent and push. It worked for me from our Windows 10 library machine.

drwells:
Sure, but `make indent` ultimately calls a shell script which in turn calls `sed`, `rm`, `tr`, etc. - its not going to run on Windows unless things happen to be set up correctly or one uses WSL.

krishnakumarg1984:
@drwells  Contrary to popular thought, all of these are available __natively__ on Windows through multiple means without any intrusive/core installs such as WSL! I do understand where this thinking comes from (years of Microsoft idiosyncrasies and the fundamental difference of the NT kernel and its userspace), but things are drastically different now.

For example, one such method is to simply use [`scoop.sh`](https://scoop.sh/) which doesn't even require admin privileges. This doesn't even require admin privileges and provides `coreutils`, `binutils`, `findutils` and such as simple, yet `native` executables in the user's `PATH`.

In practice, this is a huge boon.  I need to sometimes work at the library. But because there is no guarantee of getting access to the same computer every tim, I simply have all these utils available on my `home` drive which is automounted when I login on any machine campus-wide. 

In fact, over the last week, I have submitted PRs to this project and have run the indent script on 3 different Windows machines at my library. 

For this project,

krishnakumarg1984:
We'll probably need to revisit this? 

Right now, I get the following error when the indent script is run from the root of the repo:

>sed: 1: "examples/step-7/doc/int ...": invalid command code e

I am on BSD sed released May 10, 2005, which is what the latest macOS ships! See @9393

PR #9375, opened 2020-01-17
Remove some redundant edges from tutorial DAG

krishnakumarg1984:
This fixes a few cases that were probably missed out in #9373

### DIFF
--- a/examples/step-44/doc/builds-on
+++ b/examples/step-44/doc/builds-on
@@ -1,1 +1,1 @@
-step-8 step-18
+step-18


examples/step-54/doc/builds-on
```diff
@@ -1 +1 @@
-step-49 step-53
+step-53
```
tjhei: this is in #9373

examples/step-44/doc/builds-on
```diff
@@ -1 +1 @@
-step-8 step-18
+step-18
```
tjhei: this is correct

examples/step-12/doc/builds-on
```diff
@@ -1 +1 @@
-step-7 step-39
+step-39
```
tjhei: I am not sure about this. I would probably remove 39 instead as the prerequisite for 12 instead. I have plans to introduce a new step in between anyways. Can you remove this for now?

examples/step-54/doc/builds-on
krishnakumarg1984: Oops. Didn't see

examples/step-44/doc/builds-on
krishnakumarg1984: Ok. So, we will retain this.

examples/step-12/doc/builds-on
krishnakumarg1984: ok!

bangerth:
Same here: Please squash/fixup.

krishnakumarg1984:
Done!

PR #9376, opened 2020-01-17
Read and write vtu files

luca-heltai:
This PR allows one to add to the standard VTU format a section that is ignored by VTU readers, and that contains a compressed serialized triangulation that can be restored (including all levels and also local refinement) using GridIn::read_vtu.


luca-heltai:
Blocked by #9371

source/grid/grid_in.cc
```diff
@@ -493,6 +498,29 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+GridIn<dim, spacedim>::read_vtu(std::istream &in)
+{
+  namespace pt = boost::property_tree;
+  pt::ptree tree;
+  pt::read_xml(in, tree);
+  auto section = tree.get_optional<std::string>("VTKFile.dealiiData");
+
+  AssertThrow(section,
+              ExcMessage(
+                "While reading VTU file, failed to find dealiiData section"));
```
tamiko: Can you make the error message a bit more descriptive? In particular, we sould get the point across that we can only read in vtus that were generated by us.

include/deal.II/grid/grid_in.h
```diff
@@ -404,6 +406,22 @@ class GridIn
   void
   read_vtk(std::istream &in);
 
+  /**
+   * Read grid data from a unstructured vtu file, saved by deal.II using
+   * GridOut::write_vtu(), with the flag
+   * GridOutFlags::Vtu::serialize_triangulation set to true.
```
tamiko: What about adding something on the line "The routing does not support reading in arbitrary vtu files."

include/deal.II/base/utilities.h
```diff
@@ -136,6 +136,44 @@ namespace Utilities
   pack_integers(const std::array<std::uint64_t, dim> &index,
                 const int                             bits_per_dim);
 
+  /**
+   * If the library is configured with ZLIB, then this function compresses the
+   * input string and returns a non-zero terminated string containing the
+   * compressed input.
```
tamiko: What about mentioning that the output is a byte string (using the full character range) and should not be outputted on terminal, or a textfile directly? (vtu internally uses a base64 encoding as used later on :-))

On a related note, would it hurt to add a boolean parameter enabling base64 output? That would make this function a bit more useful to use (outside of grid_in)? And it is a simple refactoring :-)

include/deal.II/base/utilities.h
```diff
@@ -136,6 +136,44 @@ namespace Utilities
...
+   * @param[in] compression_level The compression level at which we compress
+   *
+   * @return A compressed version of the input string
+   *
+   * @authors Luca Heltai, Nicola Giuliani, 2020
+   */
+  std::string
+  compress(
+    const std::string &input,
+    const int compression_level = boost::iostreams::gzip::default_compression);
```
tamiko: same comment for the `compress` function.

source/grid/grid_in.cc
```diff
@@ -493,6 +498,29 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+
+
+  const auto string_archive =
+    Utilities::decompress(boost::beast::detail::base64_decode(*section));
+  std::istringstream              in_stream(string_archive);
+  boost::archive::binary_iarchive ia(in_stream);
+  tria->load(ia, 0);
```
bangerth: That's just nice!

include/deal.II/grid/grid_in.h
```diff
@@ -404,6 +406,22 @@ class GridIn
...
+   *
+   * The vtu file contains the triangulation in a xml section which is ignored
+   * by vtu readers. This section is only present if you set
+   * GridOutFlags::Vtu::serialize_triangulation to true, otherwise an
+   * exception is thrown.
```
bangerth:
```suggestion
   * exception is thrown. The content of all other sections of the input file are ignored.
```

bangerth:
@luca-heltai -- is this ready in principle? It doesn't currently build (and isn't indented right), though. The issue seems to be this:
```
/Users/jenkins/jenkins/workspace/dealii-OSX_PR-9376/source/grid/grid_in.cc:27:10: fatal error: 'boost/beast/core/detail/base64.hpp' file not found
#include <boost/beast/core/detail/base64.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
It looks to me like this file might have been in your local boost installation, but isn't in the bundled version...

luca-heltai:
This is ready in principle. I'll address all the comments, and try to replace boost/beast with other solutions relying on stuff that we bundle.

include/deal.II/base/utilities.h
luca-heltai: I've opened PR #9787 for the base64 part.

include/deal.II/grid/grid_out.h
```diff
@@ -921,12 +921,23 @@ namespace GridOutFlags
 
   /**
    * Flags for grid output in Vtu format. These flags are the same as those
-   * declared in DataOutBase::VtkFlags.
+   * declared in DataOutBase::VtkFlags, with the addittion of a flag that
```
masterleinad, line 924:
```suggestion
   * declared in DataOutBase::VtkFlags, with the addition of a flag that
```

source/grid/grid_in.cc
```diff
@@ -495,6 +499,35 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
+
+  AssertThrow(section,
+              ExcMessage(
+                "While reading a VTU file, failed to find dealiiData section. "
+                "Notice that we can only read grid files in .vtu format that "
+                "were created by the deal.II library, using a call to "
+                "GridOut::write_vtu(), where the flag "
+                "GridOutFlags::Vtu::serialize_triangulation is set to true."));
+
+  std::cout << "Encoded: " << *section << std::endl;
```
masterleinad, line 520: There shouldn't be any output to `stdout` by default. :slightly_smiling_face:

tests/grid/grid_in_out_vtu_01.cc
```diff
@@ -0,0 +1,79 @@
...
+    std::stringstream             stream;
+    boost::archive::text_oarchive oa(stream);
+    tria.save(oa, 0);
+    deallog << stream.str() << std::endl;
+  }
+  {
+    std::stringstream             stream;
+    boost::archive::text_oarchive oa(stream);
+    tria2.save(oa, 0);
+    deallog << stream.str() << std::endl;
```
masterleinad, line 76: It might be good to check programmatically that the out for both triangulations is the same. I can't tell by looking at the output file. :slightly_smiling_face:

luca-heltai: I generated the output using only the first mesh. :-)

source/grid/grid_in.cc
luca-heltai: Whoops. 😂

luca-heltai: Whoops. 😂

bangerth:
I still think that this is one of the coolest patches in a good long while! :-)

PR #9380, opened 2020-01-17
Replace !has_children() by is_active().

bangerth:
Follow-up to #9352. I've gone through all places where we ask `has_children() == false` or `!has_children()` and thought about what it is that we're asking here. In those places where it seemed like we're asking whether something is active, I've replaced it by `is_active()`.

/rebuild

### DIFF
--- a/source/dofs/dof_accessor_get.cc
+++ b/source/dofs/dof_accessor_get.cc
@@ -50,8 +50,8 @@ DoFCellAccessor<DoFHandlerType, lda>::get_interpolated_dof_values(
   Vector<number> &   interpolated_values,
   const unsigned int fe_index) const
 {
-  if (!this->has_children())
-    // if this cell has no children: simply return the exact values on this
+  if (this->is_active())
+    // If this cell is active: simply return the exact values on this
     // cell unless the finite element we need to interpolate to is different
     // than the one we have on the current cell
     {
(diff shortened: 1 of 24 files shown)


source/fe/fe_values.cc
```diff
@@ -4976,7 +4976,7 @@ FESubfaceValues<dim, spacedim>::reinit(
          ExcIndexRange(subface_no,
                        0,
                        cell->face(face_no)->number_of_children()));
-  Assert(cell->has_children() == false,
```
masterleinad: I would probably not replace this one.

source/grid/tria_accessor.cc
```diff
@@ -2819,7 +2818,7 @@ CellAccessor<dim, spacedim>::neighbor_child_on_subface(
   const unsigned int face,
   const unsigned int subface) const
 {
-  Assert(!this->has_children(),
```
masterleinad: ...and this as well.

source/dofs/dof_accessor_get.cc
```diff
@@ -50,7 +50,7 @@ DoFCellAccessor<DoFHandlerType, lda>::get_interpolated_dof_values(
   Vector<number> &   interpolated_values,
   const unsigned int fe_index) const
 {
-  if (!this->has_children())
```
masterleinad: I would just keep it here.

bangerth: I adjusted the text, but I think that the condition we want to test here really is whether the cell is active, because that is when data is available on it.

source/fe/fe_values.cc
bangerth: Yes, good point.

source/grid/tria_accessor.cc
bangerth: yes.

PR #9384, opened 2020-01-17
Fixes yet another redundancy in tutorial DAGs. 

krishnakumarg1984:
This one is tremendously helpful to avoid crossovers in overall DAG.

Related to #9373 and #9375

### DIFF
--- a/examples/step-12/doc/builds-on
+++ b/examples/step-12/doc/builds-on
@@ -1,1 +1,1 @@
-step-7 step-39
+step-39


tjhei:
I am confused, isn't this the same one as the one we took out of #9375?

krishnakumarg1984:
Ah. Sigh. true that. I was visually hunting mercilessly for any redundancies. Let's just get rid of this. However, if you wish to clean up the DAG, this one is quite helpful to avoid cross-over edges which will force dot to do a better placement.

PR #9387, opened 2020-01-18
Utilities::MPI::NoncontiguousPartitioner

peterrum:
I would like to propose a new vector class, which fully exploits the capabilities of  `Utilities::MPI::internal::ComputeIndexOwner::ConsensusAlgorithmPayload`. The new class makes only one assumption on the given index set: each entry can be owned by at most by one process. I.e. it allows:
- not-continuous index sets,
- the partitioning of index sets does not depend on the order of the partitions, and
- gaps.

~This PR is totally WIP. I only wanted to know if there is interest in such a vector. I have implemented s.th. like this at least two times in separate project, so I thinks it might be useful also for other developers/users! Any suggestions are welcome!~

For more info, see the test.

ping @kronbichler  @tjhei


peterrum:
Use case: repartitioning of a solution vector.

rezarastak:
We have to think about a good name for this. I like the name noncontiguous vector, as you used this name for the namespace. But in other places, e.g. in the name of this PR and the name of the test file, you use the name noncontinuous vector. I oppose to naming it noncontinuous because a vector representing field values of discrete dofs is never continuous. All vector classes are, in my opinion, discontinuous. Only the analytical `Function` classes and shape functions are continuous. 

The name noncontiguous, which is better than the noncontinuous, still needs more thought. Contiguous in C++ usually means that all the stored data is in one chunk of contiguous memory. By this definition, all distributed vector classes are noncontiguous, so it would be confusing to name this class noncontiguous. If we define contiguous to mean the local data is contiguous in memory, then this class, based on my limited understanding of it, can be categorized as contiguous and cannot be named noncontiguous.

If I correctly understand the goal of this class (correct me if I'm wrong please), it's better if we name it non covering vector, or partial coverage vector, partially covering vector, partial-fill vector or maybe (possibly confusing) incomplete vector. Another possible choice is to call it sub-vector.

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,260 @@
...
+#  include <deal.II/lac/vector_space_vector.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+namespace LinearAlgebra
+{
+  namespace noncontiguous
+  {
+    template <typename Number = double>
+    class Vector
```
Rombur: Every new Vector class should derive from `VectorSpaceVector` and/or `ReadWriteVector`.

Rombur:
> Use case: repartitioning of a solution vector.

Not sure why you need a new Vector class to do that. Can you explain?

include/deal.II/base/mpi_noncontiguous_vector.h
peterrum: I know ;) But before I implement all 20 methods, I wanted to ask!

bangerth:
Do you actually want this to be a vector in the linear algebra sense, or an array or arbitrary objects? I suspect that linear algebra operations are going to be pretty inefficient with arbitrary subdivisions onto operators, whereas if all you care about is some distributed container, then maybe this makes less of a difference.

kronbichler:
I agree with @bangerth here: Are we actually going to intend to run computations with such vectors, or are the merely used for representing some data in particular formats? I think this functionality is definitely useful to have in some form. The alternative question, if we do envision computations one this kind of vectors, is whether we should have this functionality inside the regular `LinearAlgebra::distributed::Vector` class. One thing I would like to avoid is to instantiate all our big functions in `VectorTools` or `FETools` or `FEValues` for yet another vector type - those parts are more than big enough already.

peterrum:
@kronbichler  I am not sure about how to proceed here. I would be fine if the user would provide an input/output array and indexset in some form, i.e. manage the data himself. Theoretically we could make it to a function. But, I am hesitant about this, since that would mean that we have compute the communication pattern during each function call...

kronbichler:
Given my previous comment, my first suggestion would boil down to merely make this a *container* and not call it `Vector` (that implies a particular interface and all kinds of inner products etc.). So for example `LinearAlgebra::distributed::NoncontiguousArray` or something like that. What do others think about that?

kronbichler:
Thinking about it, I think the `Utilities::MPI` namespace would be a better place for such an array class than the `LinearAlgebra` namespace.

peterrum:
@kronbichler I would use the following interface:
```cpp
namespace Utilities
{
  namespace MPI
  {
    template <typename Number = double>
    class NoncontiguousArray
    {
      public:
      // setup communication pattern
      void
      reinit(const IndexSet &indexset_has,
             const IndexSet &indexset_want,
             const MPI_Comm  communicator);

      // perform exchange
      void
      exchange(Vector & dst, const Vector & src) const;

      // start exchange (isend)
      void
      exchange_start(Vector & dst, const Vector & src, const unsigned int tag) const;

      // start exchange (irecv)
      void
      exchange_end(Vector & dst, const Vector & src, const unsigned int tag) const;

      private: // TODO
    };
  }
}
```

I am not completely sure about the class name. I would tend that the source and the destination vectors are managed by the user and passed to the function as argument. The class would only have the information related to the communication pattern and would posses a buffer.

kronbichler:
If we call it an array, this class should possess the destination data and provide some basic functionality such as `begin()`, `end()`, `operator[]`, which we could get from `AlignedVector` for example. Given the fact that we take the index set we want, my intuition would have been to manage the memory here. But it depends on how general usage we envision: In case we expect a workflow that mostly only "access" the with some local index afterwards or passes on the array to a third party library, we should integrate the memory allocation here to save the user from handling both the "exchange" object and the array. If we already workflows where there are more things happening downstream, we could go with your interface - the question being what the type/interface for the destination vector should be.

Rombur:
>  I would be fine if the user would provide an input/output array and indexset in some form, i.e. manage the data himself.

That's basically a `ReadWriteVector`. Then you can use [this function](https://dealii.org/developer/doxygen/deal.II/classLinearAlgebra_1_1distributed_1_1Vector.html#a6c22bdc2712f3ef079ee8cd6db754d79) and save the communication pattern so you don't need to recompute it everytime.

bangerth:
As for name, may I suggest just `Utilities::MPI::DistributedArray`? I *think* that that's how you intend to use it, right?

peterrum:
@Rombur Thanks for the info! I have inherited from `LinearAlgebra::CommunicationPatternBase`.

@bangerth I went with `Utilities::MPI::NoncontiguousPartitioner` for now, since the new class should not contain the actual arrays but function similarly as the normal partitioner.

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,290 @@
...
+        for (std::size_t i = 0; i < send_ranks.size(); i++)
+          {
+            // collect data to be send
+            for (std::size_t j = send_ptr[i], c = 0; j < send_ptr[i + 1]; j++)
+              send_buffers[send_ptr[i] + c++] = src[send_indices[j]];
+
+            // send data
+            MPI_Isend(send_buffers.data() + send_ptr[i],
+                      send_ptr[i + 1] - send_ptr[i],
+                      Utilities::MPI::internal::mpi_type_id(&send_buffers[0]),
```
kronbichler:
```suggestion
                      Utilities::MPI::internal::mpi_type_id(send_buffers.data()),
```

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,290 @@
...
+        }
+      }
+
+    private:
+      /// MPI communicator
+      MPI_Comm communicator;
+
+      /// CRS and MPI data structures for sending
+      std::vector<unsigned int>        send_ranks;
+      std::vector<std::size_t>         send_ptr;
```
kronbichler: Is there a reason to use `std::size_t` instead of `types::global_dof_index`?

tests/base/mpi_noncontiguous_vector_01.cc
```diff
@@ -0,0 +1,76 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// Create a serial triangulation and copy it.
```
kronbichler: Adjust test description.

tests/base/mpi_noncontiguous_vector_02.mpirun=8.output
```diff
@@ -0,0 +1,111 @@
...
+DEAL:6::
+
+
+DEAL:7::
+DEAL:7::132 133 134 135 136 137 138 139 140 141 142 143 
+DEAL:7::
+DEAL:7::
+DEAL:7::11 23 35 47 59 71 83 95 107 119 131 143 
+DEAL:7::
+DEAL:7::
```
kronbichler: Is there additional value in recording the reference values for more than say 1 and 4 ranks? I mean in the sense of a probability that we could have those cases passing but the new infrastructure being broken. I'm asking because we tend to be conservative with tests using more than a handful of MPI ranks.

peterrum: I have reduced the number of processes to 5. This is a configuration that has an empty index set.

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,254 @@
...
+    template <typename Number>
+    template <typename VectorType>
+    void
+    NoncontiguousPartitioner<Number>::update_values(VectorType &      dst,
+                                                    const VectorType &src) const
+    {
+      const auto tag = internal::Tags::noncontiguous_partitioner_update_values;
+
+      this->update_values_start(src, tag);
+      this->update_values_finish(dst, tag);
```
peterrum: @tjhei Am I using the tags correctly?

peterrum:
@Rombur May I ask for your feedback. You seem to be one of the vector experts!

peterrum:
@tjhei Now I see it: the parallel github-CI test is running with 64 bit! Smart ;)

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,256 @@
...
+     *
+     * @author Peter Munch, 2020
+     */
+    template <typename Number = double>
+    class NoncontiguousPartitioner
+      : public LinearAlgebra::CommunicationPatternBase
+    {
+    public:
+      /**
+       * Constructor. Setup point-to-point communication pattern based on the
```
kronbichler:
```suggestion
       * Constructor. Set up point-to-point communication pattern based on the
```

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,256 @@
...
+       * IndexSets @p indexset_has and @p indexset_want for the MPI communicator
```
kronbichler:
```suggestion
       * IndexSet arguments @p indexset_has and @p indexset_want for the MPI communicator
```

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,256 @@
...
+       * @p communicator.
+       */
+      NoncontiguousPartitioner(const IndexSet &indexset_has,
+                               const IndexSet &indexset_want,
+                               const MPI_Comm &communicator);
+
+      /**
+       * Constructor. Same as above but vectors of indices @p indices_has
```
kronbichler:
```suggestion
       * Constructor. Same as above but for vectors of indices @p indices_has
```

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,256 @@
...
+             const IndexSet &indexset_want,
+             const MPI_Comm &communicator) override;
+
+    private:
+      /// MPI communicator
+      MPI_Comm communicator;
+
+      /// CRS and MPI data structures for sending
+      std::vector<unsigned int>            send_ranks;
+      std::vector<types::global_dof_index> send_ptr;
```
kronbichler: Can you please document all variables individually?

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,256 @@
...
+        const std::vector<types::global_dof_index> &indices_has,
+        const std::vector<types::global_dof_index> &indices_want,
+        const MPI_Comm &                            communicator);
+
+      /**
+       * Fill the vector @p dst according to the precomputed communication
+       * pattern with values from @src.
+       *
+       * @pre The vectors only have to provide a method data(), which allows
+       *   to access their raw data.
```
kronbichler: I do not think this is what is used in the function - I see `operator[]` to be used. Maybe change it? Another question is whether we should change the necessary interface to `begin()` and `end()` to also work with the local range of `LA::d::Vector` (not sure if it would make sense directly).

include/deal.II/base/mpi_noncontiguous_vector.h
kronbichler: Looks good.

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,256 @@
...
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_mpi_noncontiguous_vector_h
+#define dealii_mpi_noncontiguous_vector_h
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/base/mpi.h>
+#include <deal.II/base/mpi.templates.h>
```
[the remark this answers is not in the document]
peterrum: Nope. I need `Utilities::MPI::internal::mpi_type_id()`.

include/deal.II/base/mpi_noncontiguous_vector.h
peterrum: You are right. `data()` is needed by the buffer. I like the idea with `begin()` and have added this.

include/deal.II/base/mpi_noncontiguous_vector.h
```diff
@@ -0,0 +1,280 @@
+// ---------------------------------------------------------------------
```
kronbichler: Can you change the filename?

include/deal.II/base/mpi_noncontiguous_vector.h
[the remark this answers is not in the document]
kronbichler: That's a pity - I'm always a bit uncomfortable with including definitions which are meant to be compilation targets in header files because of the duplicate symbols this would lead to. Now here it should work because the functions are all templates so no symbols are generated without instantiations. We try to be conservative with including other `.templates.h` files in headers - it is mostly the various `_accessor.templates.h` and `_iterator.templates.h` that come to mind. Now, this is not the topic of this PR, so we just keep it this way.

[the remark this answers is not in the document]
peterrum: How do you like it this way?

https://github.com/dealii/dealii/blob/3778ed08040b39a4858b6f1a85449e29d70b987f/source/base/mpi_noncontiguous_partitioner.inst.in#L26-L49

This way, I was able to move the include to the .cc-file.

If you are happy, I will squash the commits.

peterrum:
I have rebased with master and have added the asserts.

kronbichler:
gcc-4.9 complains about an illegal `auto`:
https://cdash.43-1.org/testDetails.php?test=35424413&build=4765
Can you have a look @peterrum ?

PR #9388, opened 2020-01-19
Implement GridGenerator::replicate_triangulation().

### LINKED ISSUE #6159: An efficient function to replicate triangulations for geometries with repeated unit structure
For geometries with repeated structure, it is quite helpful to have a function to create a triangulation by replicating the repeating unit. In our [project](https://github.com/SlaybaughLab/BART/tree/restart), we need such a functionality, so we created it. I think it might be useful for other people with similar needs. See the [demo](https://www.dropbox.com/s/hycd8h760mrt3xo/Screenshot%202018-04-04%2013.56.38.png?dl=0).

The design of this function would be for basic unit with hyper rectangle shape. So the intake parameters will be like: 1. diagonal points of the basic unit; 2. repetitions along each axis. I would propose something like: `void GridGenerator::replicate_triangulation(const Triangulation<dim, spacedim>& basic_triangulation, const std::vector<double>& diagonals, const std::vector<unsigned int>& repetitions, Triangulation<dim, spacedim> &result)`

In terms of implementation, it would be 1. copy the basic unit; 2. shift the copy to the desirable position;  3. merge the triangulation. The proposed function, however, may only be used with local triangulation to avoid hanging nodes error as it involves shifting.

drwells:
Followup to #9297. Alternative to #6188. Closes #6159.

I used the test from #6188 but implemented a more general algorithm for replication. I also addressed @masterleinad's review comments :-P

@jsrehak I'm not sure if y'all are still interested in this two years after @victor-dr-z put up the initial PR, but here it is :)


source/grid/grid_generator.cc
```diff
@@ -5845,6 +5845,96 @@ namespace GridGenerator
...
+        GridTools::delete_duplicated_vertices(
+          output_vertices,
+          output_cell_data,
+          output_subcell_data,
+          boundary_vertices,
+          1e-6 * input.begin_active()->diameter());
+        // delete_duplicated_vertices also deletes any unused vertices
+        // deal with any reordering issues created by delete_duplicated_vertices
+        GridReordering<dim>::reorder_cells(output_cell_data, true);
+        partially_replicated_trias[d].create_triangulation(output_vertices,
```
masterleinad: Is there any reason that you save all the intermediate triangulations instead of just keeping one temporary object?

drwells: No - this is a leftover from development. I force pushed a new version without this.

drwells:
I just noticed that the subcell data is messed up - I didn't correctly set it and when I tried to fix it other things broke. Let me mark this as WIP until I resolve the problem.

drwells:
I fixed the subcell data - we also now correctly copy face boundary ids.

source/grid/grid_generator.cc
```diff
@@ -5845,6 +5845,233 @@ namespace GridGenerator
...
+                         "manifold id. Double check that all faces "
+                         "which you expect to be merged together have "
+                         "the same manifold id."));
+              }
+          left = right;
+        }
+
+      const auto duplicates_begin =
+        std::unique(subcell_data.begin(), subcell_data.end());
+      subcell_data.resize(duplicates_begin - subcell_data.begin());
```
masterleinad: I guess
```
subcell_data.erase(std::unique(subcell_data.begin(), subcell_data.end()), subcell_data.end());
```
is more idiomatic but this is also fine with me.

source/grid/grid_generator.cc
```diff
@@ -5845,6 +5845,233 @@ namespace GridGenerator
...
+
+
+
+  template <int dim, int spacedim>
+  void
+  replicate_triangulation(const Triangulation<dim, spacedim> &input,
+                          const std::vector<unsigned int> &   extents,
+                          Triangulation<dim, spacedim> &      result)
+  {
+    AssertDimension(dim, extents.size());
```
masterleinad: Why don't we use `std::array` if the size is fixed?

source/grid/grid_generator.cc
```diff
@@ -5845,6 +5845,233 @@ namespace GridGenerator
...
+            for (const CellData<2> &boundary_quad :
+                 input_subcell_data.boundary_quads)
+              {
+                output_subcell_data.boundary_quads.push_back(boundary_quad);
+                for (unsigned int &vertex :
+                     output_subcell_data.boundary_quads.back().vertices)
+                  vertex += vertex_offset;
+              }
+          }
+        // check all vertices: since the grid is coarse most will be on the
```
masterleinad:
```suggestion
        // check all vertices: since the grid is coarse, most will be on the
```

source/grid/grid_generator.cc
```diff
@@ -5845,6 +5845,233 @@ namespace GridGenerator
...
+
+      // Finally, determine which objects are duplicates. Duplicates are
+      // assumed to be interior objects, so delete all but one and change the
+      // boundary id:
+      auto left = subcell_data.begin();
+      while (left != subcell_data.end())
+        {
+          const auto right =
+            std::upper_bound(left, subcell_data.end(), *left, compare);
+          // if the range has more than one item then there are duplicates -
```
masterleinad:
```suggestion
          // if the range has more than one item, then there are duplicates -
```

source/grid/grid_generator.cc
drwells: I would prefer that but we consistently use `std::vector<unsigned int>` for stuff like this in GridGenerator - I am just using the same convention.

drwells:
I fixed three out of four review comments.

Since no other function (such as `subdivided_parallelepiped`, `hyper_cross`, etc) uses `std::array` to indicate counts I think its best to do the same here.

PR #9389, opened 2020-01-19
Add assert to prevent floating point exception

peterrum:
ping @kronbichler


peterrum:
@kronbichler I tried an early return at that place. But that lead to a deadlock. The alternative would to check at multiple places if the index set is empty.

kronbichler:
Unless someone insists in having well-defined operations for zero-size sets (which are rare I suppose), I suggest we go with the proposed solution.

Issue #9390, opened 2020-01-19
Refactor TriangulationDescription::Description


peterrum:
... or `parallel::fullydistributed::ConstructionData` (as it currently still called on master branch) as described by @luca-heltai in the comments of PR #9195.

peterrum:
This can be assigned to me and @luca-heltai.

PR #8622, opened 2019-08-22
Introduce new parallel::fullydistributed::Triangulation

peterrum:
This PR is part of the effort to introduce the new `parallel::fullydistributed::Triangulation` (see #8558).

In this PR, the actual `parallel::fullydistributed::Triangulation` is added. Furthermore some tests are added and a single basic function is added to create the `parallel::fullydistributed::ConstructionData`.

Ping: @tjhei @bangerth @kronbichler @masterleinad

Issue #310, opened 2014-12-15
Verify in multigrid that Triangulation is sufficiently smooth


luca-heltai:
_From [dr.guido.kanschat](https://code.google.com/u/dr.guido.kanschat/) on August 22, 2013 18:28:58_

Multigrid methods will not complain if limit_level_difference_at_vertices is not set, but they produce wrong results at least with H1-conforming elements.

One of the multigrid classes should check whether this smoothness is needed, and if so, throw an exception if the Triangulation does not conform.

_Original issue: http://code.google.com/p/dealii/issues/detail?id=98_

masterleinad:
In `copy_local_forest_to_triangulation` (`distributed/tria.cc::2516`), we force `limit_level_difference_at_vertices` if `construct_multigrid_hierarchy` is set.
Is there a reason to leave this open?

tjhei:
what about serial computations?

masterleinad:
I thought that problem was restricted to `parallel::distributed::Triangulations`, but I might be mistaken here. I will investigate.

tjhei:
The reason for the code in distributed/tria.cc::2516 is that I need this so I can determine cell neighbors correctly without communication, not to make the multigrid method converge.

bangerth:
> what about serial computations?

In other words, you think that `p::d::Triangulation` is correct, but that `::Triangulation` may not, and also `p::s::Triangulation`.

masterleinad:
We need ` limit_level_difference_at_vertices` to have the MGTransfer being sufficient, i.e. we only need to consider transferring information between two levels.

masterleinad:
I would just assert to have `limit_level_difference_at_vertices` when calling `distribute_mg_dofs`.

tjhei:
> I would just assert to have limit_level_difference_at_vertices when calling distribute_mg_dofs.

Yes, but it is currently allowed to construct a p:d:Triangulation without this flag set (even though it will temporarily set it in copy_local_forest_to_triangulation), so the assert would fail. We could set  limit_level_difference_at_vertices in the constructor if construct_multigrid_hierarchy is set?

masterleinad:
In my opinion, we should be consistent and force `construct_multigrid_hierarchy` to be set if using multigrid no matter what kind of object the triangulation is. From what I can say, all the multigrid tests (i.e. ctest -R "multigrid") for `p::d::Triangulation` pass if I use `limit_level_difference_at_vertices` instead of `none` for `MeshRefinement`.

[...]

PR #9391, opened 2020-01-19
Refactor settings of p:f:t

peterrum:
This PR has a look at the settings of `p:f:t` as motivated by @bangerth in PR #9007 . Currently there are two settings:
```cpp
    enum Settings
    {
      default_setting = 0x0,
      construct_multigrid_hierarchy = 0x1
    };
```
Enabling `Settings::construct_multigrid_hierarchy` leads currently directly in enabling `Triangulation<dim, spacedim>::limit_level_difference_at_vertices`, leading to the issue that `p:f:t:` might have more (artificial) cells than the base triangulation it was constructed from (see issue #9353). Which is quite confusing for users.

I therefore would like to prose that `TriangulationDescription::Description` (see #9195) also contains the `MeshSmoothing` info (and in particular the `Setting` field is removed). @luca-heltai what do you think?

~Also I would like to eliminate the  input argument `construct_multilevel_hierarchy` from the function `TriangulationDescription::Utilities::create_description_from_triangulation()`. When this argument is set to `false`,~ the input triangulation is on the fly flattened and for this flattened triangulation the `Description` is created. This option is only applicable for triangulation created by global refinements and is not general. Since the flattening may lead to the generation new `CellIds` this setting also might lead to confusion and should be therefore removed. If we remove this flags, we also need to refactor less code later on (see issue #9390) ;)

~This PR is building on top of the not yet merged PR #9195 and will be rebased as soon as possible.~

FYI @agrayver


peterrum:
The relevant test is *tests/fullydistributed_grids/copy_serial_tria_09.cc*.

bangerth:
Please rebase.

peterrum:
@bangerth Done!

include/deal.II/grid/tria_description.h
```diff
@@ -398,6 +398,8 @@ namespace TriangulationDescription
      * Properties to be use in the construction of the triangulation.
      */
     Settings settings;
+
+    typename Triangulation<dim, spacedim>::MeshSmoothing smoothing;
```
bangerth: Can you add a comment?

bangerth:
Is this really all you wanted to do, or did anything get lost in the rebase?

tests/fullydistributed_grids/copy_serial_tria_09.cc
```diff
@@ -0,0 +1,107 @@
...
+
+  for (auto &cell : tria_pft.active_cell_iterators())
+    {
+      CellId id        = cell->id();
+      auto   cell_base = id.to_cell(basetria);
+      for (unsigned int d = 0; d < dim; d++)
+        Assert(std::abs(cell->center()[d] - cell_base->center()[d]) < 1e-9,
+               ExcMessage("Cells do not match"));
+    }
+}
```
bangerth: We typically output "OK" at the end of the test function for tests that check everything via assertions, rather than by producing some output.

include/deal.II/grid/tria_description.h
peterrum: @bangerth I have not finished this PR, yet. See the description of this PR. Also, I would be interested if you and @luca-heltai agree on having in `TriangulationDescription::Description` the information regarding to mesh smoothing. `TriangulationDescription::Description` is in some form used by all triangulations now...

bangerth: If you include the settings into the structure then the smoothing flags make sense as well.

peterrum: `Settings` is actually what will be removed from the struct and moved to the constructor of `p:f:t` as you suggested in an other PR.

bangerth: OK, and I still agree with that. Then I would also move the flags along with the settings.

bangerth:
So if you move the settings, then the flags should go as there as well and nothing of the current state of the patch is left...

bangerth:
Ping?

peterrum:
@bangerth Give me a couple of days. I am too occupied with a couple of other projects.

peterrum:
I will work on this PR so that we can finalize the settings of `p:f:t` before the 9.2 release.

peterrum:
@bangerth Sorry for the delay of this PR (there have been always tasks with higher priority). It is now ready to be reviewed. Thanks!

Let me review the purpose and reason of this PR. 

@agrayver noticed that p:f:t, when initialized by a serial mesh, was adding additional cells. This was related to mismatching mesh smoothing settings (`none` vs. `limit_level_difference_at_vertices`). This is not a problem as such (since it does not lead to wrong results), however, is confusing for users when traversing over the cells in the source triangulation and in p:f:t .

This PR improves this circumstance by copying the mesh smoothness setting from the source triangulation and by only amending this setting by `limit_level_difference_at_vertices` if `Settings::construct_multigrid_hierarchy`is requested (similarly as in the case of p:d:t).

Furthermore, this PR introduces a second change. While not setting `Settings::construct_multigrid_hierarchy` used to mean that the triangulation is flattened, this is not done any more. (I don't see the purpose of this flattening any more since there is function in `GridTools` exactly for this purpose.)

In conclusion: Independent of if `Settings::construct_multigrid_hierarchy` is set or not, p:f:t is set up now with all levels. Setting `Settings::construct_multigrid_hierarchy` influences only the mesh smoothing setting.

peterrum:
By the way this is related to @tjhei's comment https://github.com/dealii/dealii/pull/8622#discussion_r316937748 which is did not get at that time.

agrayver:
@peterrum thanks for addressing this. 

I have a more general question. What dictates your decision to enforce `limit_level_difference_at_vertices` other than the compatibility with p::d::f? Is there any fundamental reason to have this condition satisfied when requesting `Settings::construct_multigrid_hierarchy`?

peterrum:
> What dictates your decision to enforce limit_level_difference_at_vertices other than the compatibility with p::d::f? Is there any fundamental reason to have this condition satisfied when requesting Settings::construct_multigrid_hierarchy?

I would have thought it is requited to be able to run multigrid on distributed meshes (so that all ghost cells are available to all levels). But maybe my assumption is not correct and there is an other 
reason (like p4test, ...). 

@tjhei could you help me and shortly clarify why `p:d:t` sets `limit_level_difference_at_vertices` when `construct_multigrid_hierarchy` is enabled. Thanks!

masterleinad:
> @tjhei could you help me and shortly clarify why `p:d:t` sets `limit_level_difference_at_vertices` when `construct_multigrid_hierarchy` is enabled. Thanks!

Have a look at https://github.com/dealii/dealii/issues/310.

tjhei:
> Have a look at #310.

Yes, and this has nothing to do with p4est if I remember correctly.

masterleinad:
Yes, we need to make sure that the multigrid transfer only needs to deal with one level at a time. For continuous elements, this implies also that the level of cells meeting at a vertex can not differ by more than one. We make sure that this is the case for all triangulations when we construct the multigrid hierarchy.

peterrum:
@masterleinad @tjhei Thanks for the useful links! This is what I was also thinking. 

@agrayver did the comments answer your question?

agrayver:
> @agrayver did the comments answer your question?

Yes, thanks. Now, the logic of p::f::t construction fully makes sense to me.

peterrum:
Any opinions on this PR?

tests/fullydistributed_grids/copy_distributed_tria_01.with_p4est=true.mpirun=5.output
```diff
@@ -1,77 +1,77 @@
...
+DEAL:4:2d::n_cells:                   1373
+DEAL:4:2d::n_active_cells:            1030
 DEAL:4:2d::
 DEAL:4:2d::n_dofs:                             16641
 DEAL:4:2d::n_locally_owned_dofs:               3280
 DEAL:4:2d::
-DEAL:4:3d::n_levels:                  1
-DEAL:4:3d::n_cells:                   1108
-DEAL:4:3d::n_active_cells:            1108
+DEAL:4:3d::n_levels:                  5
```
tjhei, line 72: I am not sure I understand the changes here. Did you check that this makes sense?

tests/fullydistributed_grids/copy_serial_tria_09.cc
```diff
@@ -0,0 +1,110 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// Create a serial triangulations with different mesh-smoothing flags levels
```
tjhei, line 17:
```suggestion
// Create a serial triangulations with different mesh-smoothing flags 
```

source/grid/tria_description.cc
```diff
@@ -178,134 +178,21 @@ namespace TriangulationDescription
...
-            }
-
-          // 4) correct vertices of cells (make them local)
-          for (auto &cell : construction_data.coarse_cells)
-            for (const unsigned int v : GeometryInfo<dim>::vertex_indices())
-              cell.vertices[v] = vertices_locally_relevant[cell.vertices[v]];
-        }
-      else
+      // process cells and vertices
+      if (true /*TODO: prevent change of indentation*/)
```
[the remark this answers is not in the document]
peterrum: I left this if statement in so that the code there is not reformatted and shown as modified code in the review. I will remove the if statement now!

tests/fullydistributed_grids/copy_distributed_tria_01.with_p4est=true.mpirun=5.output
peterrum: Yes these changes make sense. There used to be the option to flatten the triangulation. This PR removed this option and keeps the levels from the original mesh. In the diff above, you can see the result: the number of levels has increased from 1 to 5.

peterrum:
@tjhei and @luca-heltai Thanks for the review! I have made the changes!

PR #9398, opened 2020-01-20
Fix 1D-bug in one version of make_flux_sparsity_pattern

### LINKED ISSUE #9322: Incorrect pattern with one version of make_flux_sparsity_pattern
There are several version of the function `make_flux_sparsity_pattern`. At least one of them appears to produce an incorrect pattern.

If I create a 1D-triangulation of 3 elements by refining 1 out of 2 elements:
```
const unsigned int n_elements = 3;
GridGenerator::subdivided_hyper_cube(triangulation, n_elements - 1);
auto cell = triangulation.begin_active();
cell++;
cell->set_refine_flag();
triangulation.execute_coarsening_and_refinement();
```
and then distribute `FE_Q<1>(1)` and create a pattern with the function

```
make_flux_sparsity_pattern(const DoFHandlerType &dof_handler,
                             SparsityPatternType & sparsity_pattern);
```

I get the correct sparsity pattern `[i, j0, j1,...]`:

[0,0,1,2]
[1,0,1,2,3]
[2,0,1,2,3]
[3,1,2,3]

but if I call the function

```
void
make_flux_sparsity_pattern(const DoFHandlerType &           dof,
                           SparsityPatternType &            sparsity,
                           const AffineConstraints<number> &constraints,
                           const bool                keep_constrained_dofs,
                           const Table<2, Coupling> &couplings,
                           const Table<2, Coupling> &face_couplings,
                           const types::subdomain_id subdomain_id);
```

with arguments that should be equivalent, I get the following pattern:

[0,0,1]
[1,0,1,2,3]
[2,1,2,3]
[3,1,2,3]

where some flux-couplings are missing.

This doesn't occur if I create an equivalent 3-element triangulation where all
cells are on the same level, see attached example.

[example.zip](https://github.com/dealii/dealii/files/4060515/example.zip)

simonsticko:
- Add a test to reproduce the bug (descrbed in #9322).

- Instantiate make_flux_sparsity_pattern for hp::DoFHandler in 1D to avoid a linking error.

- Solve the bug by avoiding to iterate over children of the face in 1D.

Closes #9322


source/dofs/dof_tools_sparsity.cc
```diff
@@ -880,6 +880,16 @@ namespace DoFTools
...
 
+
+                      // In 1D, go straight to the cell behind this particular
+                      // cell's most terminal cell. This makes us skip the
+                      // if (neighbor->has_children()) section below. We need to
+                      // do this since we otherwise iterate over the children of
+                      // the face, which are always 0 in 1D.
+                      if (DoFHandlerType::dimension == 1)
+                        while (neighbor->has_children())
+                          neighbor = neighbor->child(face_n == 0 ? 1 : 0);
```
bangerth: Ah, nice find! Thanks for tracking this down and fixing it!

PR #9399, opened 2020-01-20
Fix compiling for C++11

masterleinad:
Fixes the errors in https://cdash.43-1.org/viewBuildError.php?onlydeltap&buildid=3808.


drwells:
I'm a bit surprised that this made it through the CI.

PR #9400, opened 2020-01-20
Utilities::to_string -> std::to_string

### LINKED ISSUE #9361: Step-10 tutorial suddenly introduces Utilities::to_string without explanation (or comparison with std::string)
All the way until Step 6,  whenever the mesh refinement cycle number was appended to filenames,  `std::to_string` was used.

However, in step-10, this suddenly switched to a deal.ii function, `Utilities::to_string`, for eg.

` std::string filename_base = "ball_" + Utilities::to_string(refinement); `

Why is this change? How is the deal.ii version different to the std version?

drwells:
None of these use the second arguments for padding so we might as well use the standard library function.

Fixes #9361.

### DIFF
--- a/source/differentiation/sd/symengine_tensor_operations.cc
+++ b/source/differentiation/sd/symengine_tensor_operations.cc
@@ -46,7 +46,7 @@ namespace Differentiation
         {
           std::string out;
           for (unsigned int i = 0; i < rank; ++i)
-            out += dealii::Utilities::to_string(indices[i]);
+            out += std::to_string(indices[i]);
           return out;
         }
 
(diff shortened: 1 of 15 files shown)


drwells:
This change is pretty dull so I'll start the tester myself.

/rebuild

source/differentiation/sd/symengine_tensor_operations.cc
```diff
@@ -46,7 +46,7 @@ namespace Differentiation
         {
           std::string out;
           for (unsigned int i = 0; i < rank; ++i)
-            out += dealii::Utilities::to_string(indices[i]);
+            out += dealii::std::to_string(indices[i]);
```
drwells: Serves me right for using sed and not having symengine installed. Let me fix this.

PR #9401, opened 2020-01-20
Make channel limits consistent.

drwells:
This class uses the MPI partitioner that is okay with channel numbers up to 200.

### DIFF
--- a/include/deal.II/lac/la_parallel_vector.templates.h
+++ b/include/deal.II/lac/la_parallel_vector.templates.h
@@ -900,8 +900,7 @@ namespace LinearAlgebra
       const unsigned int                communication_channel,
       ::dealii::VectorOperation::values operation)
     {
-      (void)communication_channel;
-      (void)operation;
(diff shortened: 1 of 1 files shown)


tjhei:
I think I did this on purpose, see https://github.com/dealii/dealii/blob/d780fd0cc4fb6388987f2ec0b841783bebc77d1b/include/deal.II/lac/la_parallel_block_vector.templates.h#L262-L267

Can you add an assert that the value is <=100 in the code here, instead?

drwells:
Ah - makes sense. Fixed!

drwells:
I just force pushed another change to use `AssertIndexRange`.

kronbichler:
This causes some test failures like the one here:
https://cdash.43-1.org/testDetails.php?test=28781764&build=3829
We need to increase the limit to 200 because block vectors use the range between 100 and 200 and we cannot know from within the vectors if they are used in separation or as part of a block vector.

drwells:
I'll push a quick fix.

PR #9402, opened 2020-01-20
Rename step-71 to step-47.

drwells:
A completely cosmetic and subjective change: unless @bangerth and @tjhei like this then we should do something else.

Now seems like a good time to plug a hole in the tutorial graph and just completely replace step-47 with one of the new tutorial programs. What do y'all think?


tjhei:
Generally, I am not a big fan in renaming tutorials:
1. makes reading the git history and old PRs more difficult.
2. confusing for users, especially if the tutorial used to exist.
3. makes the example number no longer a unique identifier.

There is also no problem in having gaps (in my opinion). That said, step-47 was never a published tutorial, so I am not 100% opposed. @bangerth ?

tamiko:
I am opposed to this change. In particular because it will be confusing between library versions.

drwells:
Why do you think it will be confusing? step 71 is not in any published release and we have never had a step 47.

tamiko:
Ah, sorry I misunderstood!

bangerth:
I'm all for using the step-47 slot, even though a bunch of people over the years found the step-47 code and used pieces for it.

I've asked the two junior authors of step-71 whether they had already used that program's name in application documents. Let's give them a couple of days to respond before we rename something they may have been referring to in documents.

bangerth:
Both of them got back to me already and said it's ok. So let's do it!

bangerth:
Anyone got opposing opinions?

(Now we just need to replace step-19 as well!)

tjhei:
> Anyone got opposing opinions?

No.



> (Now we just need to replace step-19 as well!)

I would leave step-19 alone as it existed as a finished tutorial. There is no big advantage of removing all gaps in my opinion.

examples/step-47/step-71.cc
```diff
@@ -71,7 +71,7 @@
 #include <cmath>
 
 
-namespace Step71
+namespace Step47
```
tjhei: did you forget renaming the .cc file?

drwells: Yes.

kronbichler:
I'm merging now since everyone seemed to agree on this tutorial given that it is only part of the library for less than a month without extensive history.

PR #9409, opened 2020-01-21
Update the documentation of int_to_string() and to_string().

bangerth:

PR #9410, opened 2020-01-21
Use std::to_string() instead of boost lexical_cast.

bangerth:
See also #9409.

### DIFF
--- a/source/base/utilities.cc
+++ b/source/base/utilities.cc
@@ -437,11 +437,20 @@ namespace Utilities
   std::string
   to_string(const number value, const unsigned int digits)
   {
-    std::string lc_string = std::to_string(value);
+    // For integer data types, use the standard std::to_string()
+    // function. On the other hand, that function is defined in terms
+    // of std::sprintf, which does not use the usual std::iostream
+    // interface and tries to render floating point numbers in awkward
+    // ways (see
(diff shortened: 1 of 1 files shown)


drwells:
I think that this change was my first or second deal.II PR ever :)

masterleinad:
You might want to update the documentation as well:
https://github.com/dealii/dealii/blob/1c31021568346676c9ec72e9d9e5f70297061ca3/include/deal.II/base/utilities.h#L200-L207

drwells:
@masterleinad that is in #9409.

bangerth:
This fails the `base/utilities_11` test because `std::to_string()` writes floating point numbers with a different number of significant digits than `boost::lexical_cast`. Indeed, based on https://www.boost.org/doc/libs/1_65_1/doc/html/boost_lexical_cast/synopsis.html#boost_lexical_cast.synopsis.lexical_cast and https://en.cppreference.com/w/cpp/string/basic_string/to_string, `boost::lexical_cast` uses the standard `iostream` interface for conversion, whereas `std::to_string` uses `sprintf`, and as the example on the latter page shows, does not actually yield what one would typically want for floating point numbers.

I modified this consequently: We really only want to use `std::to_string()` for integral types.

PR #9415, opened 2020-01-22
Introduce GeometryInfo::face_indices()

bangerth:
I looked at some patch or other yesterday that contained a loop bounded by `GeometryInfo<dim>::faces_per_cell` and that, unsurprisingly, was too long to fit into one line. My aesthetic sense was offended and I thought that we can do that better.

We recently introduced `cell->face_iterator()`, which can be used in range-based `for` loops. We can do something similar here: Introduce a function `GeometryInfo::face_indices()` that returns all valid indices for faces, and use that in all of these loops.

There were around 250 loops that seem to be bounded by  `GeometryInfo<dim>::faces_per_cell`. It's hard to convert them all because they are often split across lines, but I did manage to script the conversion of some 70 in the second commit here.

### DIFF
--- a/include/deal.II/base/geometry_info.h
+++ b/include/deal.II/base/geometry_info.h
@@ -24,6 +24,7 @@
 
 #include <boost/range/irange.hpp>
 
+#include <array>
 #include <cstdint>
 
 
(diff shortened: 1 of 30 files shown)


drwells:
Its not as efficient, but in other places we just return a `std::array` and let the compiler optimize it away if necessary (instead of the fancy range object). Could we also do that here for consistency?

drwells:
Actually, I'm not sure in what cases the compiler will optimize the loop away. I'll take a closer look.

peterrum:
> Its not as efficient, but in other places we just return a std::array and let the compiler optimize it away if necessary (instead of the fancy range object). Could we also do that here for consistency?

@drwells I don't think it is a good idea to fill some `std::array` with indices (except maybe if we make it a const expression). What the compiler looses is that the indices are sorted and linearly increasing. 

In the case of the iterator, the compiler can convert the iteration to a simple loop. See
https://godbolt.org/z/EL9LJh and https://godbolt.org/z/HJtE5F

But in the array case, it is a loop over all indices in the array. See https://godbolt.org/z/5h3evE

drwells:
You beat me to it! Yeah, my suggestion is not good :smile:

bangerth:
I'll add that I was worried that the boost stuff is just too impenetrable to the compiler and that we would get worse code. So I started to write a replacement `IndexRange` class with all `constexpr` functions. 

But it looks like the compiler can look through it just fine. For example, in https://godbolt.org/z/-s7Dc7 the compiler sees that in 
```
int main()
{
    volatile char *g = (char*)1234;
    for(auto i : boost::irange(0U, 10u))
      *g = 127;
}
```
we just execute the loop ten times and unrolls it without trouble. I think that this should be good enough.

Issue #9406, opened 2020-01-20
[Step-20] Reorganise prior tutorial contents since the input dependency for step-20 is incorrect/incomplete


krishnakumarg1984:
According to the main landing page for the tutorials, Step-20 depends only on Step-4. This is incorrect. 

### Why this needs to be looked at?
New users of the library may tend to go to step-20 after step-4 or soon afterwards because
- the dependency diagram branches off from step-4 (a fairly easy tutorial).
- Prof Bangerth's video lectures move onto this wherein vector-values problems (with the mixed Laplace equation - the 2nd ever PDE discussed) are introduced.

### The problematic concepts for users coming to step-20 from step-4

In Step-20, a few key concepts are presented casually without a reasonable introduction.

1. Namespaces.  The tutorial says _"The last step is as in all __previous__ programs:_
```
namespace Step20
{
  using namespace dealii;
```
which catches everyone by surprise. Step-7 introduces namespaces.

2. The `FEFaceValues` class
There are boundary integrals over faces arising from the weak form of the Mixed Laplace equation which haven't really been introduced in depth. (Step-7 does this).

3. Computation on successively refined grids (global refinement). It is Step-5 which introduces this.

### Potential resolution

Firstly, I think Step-7 introduces way too many concepts. Despite studying it for 3 days, I could not finish it and abandoned my efforts.  It is perhaps more appropriate if Step-7 sticks to introducing two central ideas
1. the verification of code using the Method of Manufactured Solutions, which already introduces pretty substantial content for an early tutorial. 
2. Producing convergence tables for order of accuracy evaluation.

Secondly, the branch-off point for Step-20 can be changed to __step-5__ by incorporating the following changes to it:
- Introduce the `namespace` concept  (this is trivial to do, and does not affect the output results in any way). Additionally, we can include namespaces in Step-6 as well (rest of the tutorials all use namespaces already). 
- Use a non-homogenous Neumann BC (will change the results). This might appear to be a strong change, but I argue otherwise. Step-4 has already introduced a non-homogenous Dirichelet BC, while Step-5 actually regresses a bit and reverts to a zero BC (rather than building gently on Step-4, the introduction of non-homogenous Neumann BC is deferred to Step-7 which causes an information overload along with so many other new things it introduces).  The basic idea of boundary integrals owing to Neumann BCs is so common and is best to introduce `FEFaceValues` in a simple Laplace problem, rather than in a Mixed Laplace problem as is done currently. 

[...]

PR #9416, opened 2020-01-22
Mention namespaces in step-20.

bangerth:
In reference to #9406.


Rombur:
CI test is failing seems totally unrelated but since @bangerth wanted to run the testers I won't merge until the other testers are done.

PR #9419, opened 2020-01-23
Remove seemingly unused code in fe_poly.cc

GrahamBenHarper:
This is related to #8711, because it involves cleaning up the use of class templating in `FE_Poly`.

I think something needs to be done here because `FE_Poly<PolynomialSpace<dim>, dim, spacedim>` is exactly how `FE_DGP` is defined. Also, when #8711 goes through, these specific instances of `fill_fe_values` in `fe_poly.cc` should no longer exist.

The same may not be exactly true for the `FE_Poly<TensorProductPolynomials<dim>, dim, spacedim>` code also in `fe_poly.cc`.

Either way, I ran a build and some relevant tests on my end and they appear to run fine, so I'd like to see how every test runs on the testers just to make sure. If any tests fail, then the correct solution is probably to move the deleted code over to `FE_DGP` instead.


GrahamBenHarper:
Since it looks like the testers are down, I ran the testsuite on my end overnight.

I'm commenting this so I remember later: It looks like `serialization/fe_series_02` is the only test that fails. I'll check why that happens later today.

GrahamBenHarper:
Actually the error was unrelated to my changes. It said `deal.II has to be configured with GSL in order to use Legendre transformation`. :stuck_out_tongue:

masterleinad:
> Actually the error was unrelated to my changes. It said `deal.II has to be configured with GSL in order to use Legendre transformation`.

We fixed that in #9422.

GrahamBenHarper:
Ah! I forgot to update from yesterday's version :stuck_out_tongue: Thanks for letting me know.

PR #9422, opened 2020-01-23
Restrict fe_series_02

masterleinad:
Fixes https://cdash.43-1.org/testDetails.php?test=29110191&build=3878.


bangerth:
It's a bit of a shame: The function in question is `gsl_sf_legendre_Pl()`, but in C++17, we now have these functions available as well: https://en.cppreference.com/w/cpp/numeric/special_functions/legendre

But that's a question for another day.

PR #9428, opened 2020-01-24
Add python wrappers for Quadrature

agrayver:
Part of #9015

### DIFF
--- a/contrib/python-bindings/include/quadrature_wrapper.h
+++ b/contrib/python-bindings/include/quadrature_wrapper.h
@@ -1,0 +1,99 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2016 - 2017 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_quadrature_wrapper_h
+#define dealii_quadrature_wrapper_h
+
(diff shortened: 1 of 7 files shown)


contrib/python-bindings/include/quadrature_wrapper.h
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Destructor.
+     */
+    ~QuadratureWrapper();
+
+    /**
+     * Create QGauss.
+     */
+    void
+    create_gauss(const unsigned int n);
```
masterleinad: Is there any reason not to templatize this to allow any kind of Quadrature formula? In the end, you only care for the quadrature points and weights anyway, don't you? You could still assert inside that the template type is one that is safe to be used, i.e. that doesn't have extra member variables.

agrayver: Not sure I understand. Are you suggesting to pass the quadrature type as a template argument? How to expose this to the user then? Also, many quadratures are constructed with additional parameters, so there is no universal constructor...

agrayver:
@masterleinad thanks, I addressed your points.

contrib/python-bindings/include/quadrature_wrapper.h
masterleinad: I was thinking about having something similar to what we do in [QuadratureSelector](https://www.dealii.org/developer/doxygen/deal.II/quadrature__selector_8cc_source.html#l00059) here.

agrayver: Personally, I don't see why this is better (there are many quadratures with ad-hoc constructor parameters). However, if you feel strong about that, I can do it.

agrayver:
Can we move forward with this?

contrib/python-bindings/include/quadrature_wrapper.h
masterleinad: No, I don't feel strongly about it. Let's move forward. We can always reevaluate before the next release.

PR #9430, opened 2020-01-24
Fix ParameterHandler::read_xml_recursively for actions

### LINKED ISSUE #9427: ParameterHandler::add_parameter () not working properly for...
... `parse_input_from_json ()` and probably not for `parse_input_from_xml ()`.

My observation is (via calling `print_parameters`) that the parameters are read correctly but the values are not set registered during `add_parameter` are not set with the values from the input file. I guess the created actions are not called...

@luca-heltai :is my observation right?

FYI @lpsaavedra

peterrum:
@luca-heltai this is practically how you described it. ~I just had problems with relative (`get_current_full_path`) and absolute paths which is the reason for the new flag.~

closes #9427

FYI @lpsaavedra

### DIFF
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1574,17 +1574,11 @@ public:
    * Exception for when an entry in an XML parameter file does not match the
    * provided pattern. The arguments are, in order, the entry value, entry
    * name, and a description of the pattern.
+   *
+   * @deprecated Use ExcValueDoesNotMatchPattern instead of ExcInvalidEntryForPatternXML.
    */
-  DeclException3(ExcInvalidEntryForPatternXML,
-                 std::string,
-                 std::string,
-                 std::string,
-                 << "    The entry value \n"
-                 << "        " << arg1 << '\n'
-                 << "    for the entry named\n"
-                 << "        " << arg2 << '\n'
-                 << "    does not match the given pattern:\n"
-                 << "        " << arg3);
+  using ExcInvalidEntryForPatternXML DEAL_II_DEPRECATED =
+    ExcValueDoesNotMatchPattern;
 
   /**
    * Exception for when the file given in an include statement cannot be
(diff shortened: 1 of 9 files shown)


luca-heltai:
I'm unsure about the additional flag. From the perspective of the user, it is unclear how to use this flag, since it exposes the internals of the class (i.e., the fact that you can actually write in arbitrary position in the parameter's path, without entering any subsection). I would rather not expose this, if possible. Couldn't we simply enter the correct subsection, before calling `set`? This should be done externally, before calling the `read_xml_recursively`.  This allows us to change the internals of the class, without any consequences in the future.

peterrum:
@luca-heltai Thanks for the hint! I will try that out!

peterrum:
@luca-heltai All your hints were quite useful! Thanks! I have updated the code and have modified the json test so that it also tests if the values have been set correctly.

peterrum:
@luca-heltai Thanks! Done!

peterrum:
@kronbichler Could you add a `ready to test` here as well. This PR has to be merged first ;)

luca-heltai:
I cannot. I’m from my phone...

peterrum:
@kronbichler Thanks!

peterrum:
@luca-heltai @kronbichler With the changes of this PR the thrown exceptions are modified (e.g. `ExcInvalidEntryForPatternXML` -> `ExcValueDoesNotMatchPattern`; see also the failing tests). Is this fine? Or should I catch `ExcValueDoesNotMatchPattern` in `read_xml_recursively` and convert it to `ExcInvalidEntryForPatternXML`?

kronbichler:
That is indeed a backward compatibility issue - assuming this function/exception was introduced before the last release. So then I would indeed go with your suggestion of converting it. @luca-heltai what do you think?

peterrum:
I cannot convert the exception. `ExcInvalidEntryForPatternXML` is of type `DeclException3` and do not have access to those arguments. I instead introduced:
```cpp
using ExcInvalidEntryForPatternXML DEAL_II_DEPRECATED = ExcValueDoesNotMatchPattern;
```

PR #9431, opened 2020-01-25
Fix particle probability density mpi problem

MFraters:
Same fix as in https://github.com/geodynamics/aspect/pull/3349.

Found a problem where for the probability density function particle generator does not generate the right ammount of particles when there are less particles then mpi ranks:

  1.  setting 1 particle in input, while having 3 mpi processes, no particles are created;
  2.  setting 2 particles in input, while having 3 mpi processes, creates 3 particles.

This should fix the problem and I added some tests (2 of which would have failed without this code). The problem was in how the number of particles per mpi process was computed, which is fixed now with the help of @gassmoeller. I tested all the particle generator tests in aspect locally and they all still pass (also added some new tests there which now show the right results). Did not test it with the dealii tests because I am having some problems getting the dealii tester to work on my laptop (on master).

### DIFF
--- a/source/particles/generators.cc
+++ b/source/particles/generators.cc
@@ -312,7 +312,8 @@ namespace Particles
         // Calcualate number of local particles
         const types::particle_index end_particle_id =
           llround(static_cast<double>(n_particles_to_create) *
-                  local_end_weight / global_weight_integral);
+                  ((local_start_weight + local_weight_integral) /
+                   global_weight_integral));
         n_local_particles = end_particle_id - start_particle_id;
 
         if (random_cell_selection)
--- a/source/particles/generators.cc
+++ b/source/particles/generators.cc
@@ -304,13 +304,16 @@ namespace Particles
           }
 #endif
 
-        // Calculate start id and number of local particles
+        // Calculate start id
         start_particle_id =
           std::llround(static_cast<double>(n_particles_to_create) *
                        local_start_weight / global_weight_integral);
-        n_local_particles =
-          std::llround(static_cast<double>(n_particles_to_create) *
-                       local_weight_integral / global_weight_integral);
+
+        // Calcualate number of local particles
+        const types::particle_index end_particle_id =
+          llround(static_cast<double>(n_particles_to_create) *
+                  local_end_weight / global_weight_integral);
+        n_local_particles = end_particle_id - start_particle_id;
 
         if (random_cell_selection)
           {
--- a/source/particles/generators.cc
+++ b/source/particles/generators.cc
@@ -311,9 +311,9 @@ namespace Particles
 
         // Calcualate number of local particles
         const types::particle_index end_particle_id =
-          llround(static_cast<double>(n_particles_to_create) *
-                  ((local_start_weight + local_weight_integral) /
-                   global_weight_integral));
+          std::llround(static_cast<double>(n_particles_to_create) *
+                       ((local_start_weight + local_weight_integral) /
+                        global_weight_integral));
         n_local_particles = end_particle_id - start_particle_id;
 
         if (random_cell_selection)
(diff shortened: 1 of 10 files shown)


MFraters:
I am not sure why it is complaining again about my username
```
Running indentation test on Pull Request #9431

invalid author 'MFraters' without firstname and lastname
```

When I do `$ git config --global -l`
it returns
```
user.email=menno.fraters@outlook.com
user.name=Menno Fraters
```

and the git log also has my full name:
```
commit 3c3bd8f98deb864c0b1f1f7cd6693d92e46fa044 (HEAD -> fix_particle_probability_density_mpi_problem, menno/fix_particle_probability_density_mpi_problem)
Author: Menno Fraters <menno.fraters@outlook.com>
Date:   Fri Jan 24 18:39:20 2020 -0800

    Fix particle probability density mpi problem
```

drwells:
There might be something wrong with the tester - I get
```
[drwells@auricle dealii-dev]$ words=($(git log --since "2019-01-01" --format="%aN" | head -n 1))                
[drwells@auricle dealii-dev]$ echo "${#words[@]}"
2
```

masterleinad:
I guess it is a similar issue as we had with Krishankumar earlier. I just pushed a mailmap entry.

source/particles/generators.cc
```diff
@@ -304,13 +304,16 @@ namespace Particles
...
           std::llround(static_cast<double>(n_particles_to_create) *
                        local_start_weight / global_weight_integral);
-        n_local_particles =
-          std::llround(static_cast<double>(n_particles_to_create) *
-                       local_weight_integral / global_weight_integral);
+
+        // Calcualate number of local particles
+        const types::particle_index end_particle_id =
+          llround(static_cast<double>(n_particles_to_create) *
+                  local_end_weight / global_weight_integral);
```
masterleinad: Looks like you missed to define `local_end_weight`.

MFraters: ah, sorry. Because I was struggling a bit with the dealii tester I apparently forgot to rebuild when I saved the changes. I rebuild it, and the branch compiles fine on my machine now.

masterleinad:
> This should fix the problem and I added some tests (2 of which would have failed without this code).

Did you forget to push the tests?

MFraters:
> > This should fix the problem and I added some tests (2 of which would have failed without this code).
> 
> Did you forget to push the tests?

I meant the tests in aspect there, sorry for not being clear on that. I have now added four tests which pass on my system (and all the particles/generators tests also still pass).

source/particles/generators.cc
```diff
@@ -304,13 +304,17 @@ namespace Particles
...
         start_particle_id =
           std::llround(static_cast<double>(n_particles_to_create) *
                        local_start_weight / global_weight_integral);
-        n_local_particles =
-          std::llround(static_cast<double>(n_particles_to_create) *
-                       local_weight_integral / global_weight_integral);
+
+        // Calcualate number of local particles
+        const types::particle_index end_particle_id =
+          llround(static_cast<double>(n_particles_to_create) *
```
masterleinad: Any reason to drop `std` here?
```suggestion
          std::llround(static_cast<double>(n_particles_to_create) *
```

masterleinad:
Now, the indentation is off. :wink:

MFraters:
fixed the indentation of the last commit :)

masterleinad:
Thanks!

MFraters:
Thanks for the quick review!

tjhei:
> I guess it is a similar issue as we had with Krishankumar earlier. I just pushed a mailmap entry.

yes and no. The problem is that the merge commit is created by github and the name used is the name set in his profile (https://github.com/MFraters shows "MFraters"). But even if he adjusts that, the tester would still fail (which is the case we had with Krishankumar).

MFraters:
> > I guess it is a similar issue as we had with Krishankumar earlier. I just pushed a mailmap entry.
> 
> yes and no. The problem is that the merge commit is created by github and the name used is the name set in his profile (https://github.com/MFraters shows "MFraters"). But even if he adjusts that, the tester would still fail (which is the case we had with Krishankumar).

I have changed my profile name on github. When I got this error when I had just pushed one commit directly from my laptop, so my github name should not have been in the commit, I think.

PR #9435, opened 2020-01-27
Avoid using deprecated functions

bangerth:
I've deprecated a number of functions in `DoFTools` in recent patches, primarily ones that return their information through output arguments. Avoid their use in the library and in the tutorials.

### DIFF
--- a/examples/doxygen/block_dynamic_sparsity_pattern.cc
+++ b/examples/doxygen/block_dynamic_sparsity_pattern.cc
@@ -53,8 +53,8 @@ int main()
   DoFTools::make_hanging_node_constraints(dof, constraints);
   constraints.close();
 
-  std::vector<unsigned int> dofs_per_block(fe.n_blocks());
-  DoFTools::count_dofs_per_block(dof, dofs_per_block);
+  const std::vector<unsigned int> dofs_per_block =
+    DoFTools::count_dofs_per_fe_block(dof);
 
   BlockDynamicSparsityPattern dsp(fe.n_blocks(), fe.n_blocks());
   for (unsigned int i = 0; i < fe.n_blocks(); ++i)
--- a/include/deal.II/dofs/dof_tools.h
+++ b/include/deal.II/dofs/dof_tools.h
@@ -2303,7 +2303,7 @@ namespace DoFTools
    */
   template <typename DoFHandlerType>
   std::vector<types::global_dof_index>
-  count_dofs_per_component(
+  count_dofs_per_fe_component(
     const DoFHandlerType &           dof_handler,
     const bool                       vector_valued_once = false,
     const std::vector<unsigned int> &target_component   = {});
@@ -2338,9 +2338,9 @@ namespace DoFTools
    */
   template <typename DoFHandlerType>
   std::vector<types::global_dof_index>
-  count_dofs_per_block(const DoFHandlerType &           dof,
-                       const std::vector<unsigned int> &target_block =
-                         std::vector<unsigned int>());
+  count_dofs_per_fe_block(const DoFHandlerType &           dof,
+                          const std::vector<unsigned int> &target_block =
+                            std::vector<unsigned int>());
 
   /**
    * @deprecated A version of the previous function that returns its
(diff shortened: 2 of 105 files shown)


bangerth:
See also #9182, #9272.

drwells:
```

/Users/runner/runners/2.164.0/work/dealii/dealii/source/multigrid/mg_transfer_component.cc:351:19: error: no viable overloaded '='
  component_start = DoFTools::count_dofs_per_block(mg_dof, target_component);
```

masterleinad:
So the problem here is, that the old version is a better match if the last argument is actually provided in case deal.II is configured without 64-bit indices.

masterleinad:
I don't quite see a good resolution hear other than renaming the new function.

bangerth:
I squashed the commits in question.

I did think more about names, but I think I still like the name chosen here (`count_dofs_per_fe_component/block()`) better than any of the other suggestions.

masterleinad:
Thanks!

examples/doxygen/block_dynamic_sparsity_pattern.cc
```diff
@@ -54,7 +54,7 @@ int main()
   constraints.close();
 
   std::vector<unsigned int> dofs_per_block(fe.n_blocks());
-  DoFTools::count_dofs_per_block(dof, dofs_per_block);
+  DoFTools::count_dofs_per_fe_block(dof, dofs_per_block);
```
tjhei: this looks incorrect.

include/deal.II/dofs/dof_tools.h
```diff
@@ -2258,7 +2258,7 @@ namespace DoFTools
    */
   template <typename DoFHandlerType>
   std::vector<types::global_dof_index>
-  count_dofs_per_component(
+  count_dofs_per_fe_component(
```
tjhei: I don't think that we have used it in deal.II so far, but this would be a great candidate for [[nodiscard]] to make the conversion from the old function harder to get wrong.

bangerth: Yes, nice idea. We'd need to write a bit of cmake detection code because the attribute is only C++17: https://en.cppreference.com/w/cpp/language/attributes/nodiscard

(That's a nice and drastic example in that page, by the way...)

bangerth: Could we leave that for a later patch?

bangerth: On second though: There are a great many functions that return something. In essence, every function that is `const` falls in this category, as well as all global functions with only `const` arguments. Unless we have a plan to mark *all* of these functions, I think doing it for this one here isn't making much of a point.

masterleinad: We already check for `[[nodiscard]]` when we check for C++17. I thought about using it everywhere as well and there is a `clang-tidy` check (https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nodiscard.html) for it. However, it results in some many code changes (with annoyingly long macro names) that I refrained from doing it.

examples/doxygen/block_dynamic_sparsity_pattern.cc
tjhei: This bug would have been caught by [[nodiscard]], right? :-)

include/deal.II/dofs/dof_tools.h
tjhei: I am fine with doing it at some other point (or not at all).

examples/doxygen/block_dynamic_sparsity_pattern.cc
masterleinad: Does `clang-tidy` check files in `examples/doxygen`?

bangerth: No, that's the problem -- this piece of code wouldn't have compiled at all. I replaced all occurrences of the function name and then compiled the library and all examples and tests, but this file is simply not compiled.

masterleinad: Fix this.

bangerth:
I fixed that last issue in `examples/doxygen/block_dynamic_sparsity_pattern.cc` now.

PR #9437, opened 2020-01-27
Update uses of compute() to evaluate()

GrahamBenHarper:
There are a few places where polynomial classes refer to `compute()` instead of `evaluate()`.

`PolynomialsAdini` had a method named `compute()` still, and the remaining classes referred to `compute()` instead of `evaluate()` in their documentation. This should be cleaned up now.

Related to #8619, #1973, and probably some others that I can't think of at the moment.

### DIFF
--- a/include/deal.II/base/polynomial_space.h
+++ b/include/deal.II/base/polynomial_space.h
@@ -154,7 +154,7 @@ public:
    * Compute the value of the <tt>i</tt>th polynomial at unit point
    * <tt>p</tt>.
    *
-   * Consider using compute() instead.
+   * Consider using evaluate() instead.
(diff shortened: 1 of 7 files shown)


GrahamBenHarper:
@tjhei I think `tjhei-macbook` died while it was testing.

tjhei:
> @tjhei I think `tjhei-macbook` died while it was testing.

Yes, that is correct. It is safe to merge as is, I would guess...

GrahamBenHarper:
If it's not safe I'm sure I'll quickly see @GrahamBenHarper and a PR with "revert #9437" :stuck_out_tongue:

PR #9445, opened 2020-01-28
Add ParameterHandler::subsection_path_exists

elauksap:
This adds a new `ParameterHandler` method to check whether a subsection path has been declared in the current tree.

I needed such a functionality for my application, if you also find it relevant I can implement a small test.

### DIFF
--- a/source/base/parameter_handler.cc
+++ b/source/base/parameter_handler.cc
@@ -825,6 +825,27 @@ ParameterHandler::leave_subsection()
 
 
 
+bool
+ParameterHandler::subsection_path_exists(
+  const std::vector<std::string> &sub_path) const
+{
+  // Get full path to sub_path (i.e. prepend subsection_path to sub_path).
+  std::vector<std::string> full_path(subsection_path);
+  full_path.insert(full_path.end(), sub_path.begin(), sub_path.end());
+
+  boost::optional<const boost::property_tree::ptree &> subsection(
+    entries->get_child_optional(
+      collate_path_string(path_separator, full_path)));
+
+  // If subsection is boost::null (i.e. it does not exist)
+  // or it exists as a parameter/alias node, return false.
+  // Otherwise (i.e. it exists as a subsection node), return true.
+  return !(!subsection || is_parameter_node(subsection.get()) ||
+           is_alias_node(subsection.get()));
+}
+
+
+
 std::string
 ParameterHandler::get(const std::string &entry_string) const
 {
--- a/tests/parameter_handler/parameter_handler_subsection_exists.cc
+++ b/tests/parameter_handler/parameter_handler_subsection_exists.cc
@@ -1,0 +1,86 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2003 - 2018 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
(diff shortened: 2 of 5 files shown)


doc/news/changes/minor/20200128PasqualeClaudioAfrica
```diff
@@ -1,4 +1,5 @@
 Improved: ParameterHandler class can now optionally print parameters in the same
 order as they are declared, rather than sorted in alphabetical order.
 <br>
+New: added ParameterHandler::subsection_path_exists() method.
```
bangerth: Just put it into a separate file.

source/base/parameter_handler.cc
```diff
@@ -763,6 +763,21 @@ ParameterHandler::leave_subsection()
...
+ParameterHandler::subsection_path_exists(
+  const std::vector<std::string> &sub_path) const
+{
+  std::vector<std::string> sub_path_base(sub_path);
+
+  std::string subsection = sub_path_base.back();
+  sub_path_base.pop_back();
+
+  return static_cast<bool>(entries->get_optional<std::string>(
+    get_current_full_path(sub_path_base, subsection)));
```
bangerth: I *think* that what you are doing here is to find out whether there is an entry called `subsection` in the path `sub_path_base`, but some commentary would help. Is it guaranteed that `subsection` actually corresponds to another section, rather than just a regular entry or an alias?

elauksap:
Thanks @bangerth! Now it checks that the path does not correspond to an entry or an alias but rather to an actual subsection.

If that's ok tomorrow I will implement a test to check for all the possible scenarios.

bangerth:
Yes, looks good. More tests are always better, and you can take the idea of other kinds of nodes as a guide to writing a test :-)

elauksap:
Rebased on top of #9443 and new test added. :wink:

doc/news/changes/minor/20200128PasqualeClaudioAfrica
```diff
@@ -1,4 +1,3 @@
-Improved: ParameterHandler class can now optionally print parameters in the same
-order as they are declared, rather than sorted in alphabetical order.
```
tjhei: It looks like you are mixing up your old changelog. Shouldn't the new change go into 20200129*?

source/base/parameter_handler.cc
```diff
@@ -825,6 +825,27 @@ ParameterHandler::leave_subsection()
 
 
 
+bool
+ParameterHandler::subsection_path_exists(
+  const std::vector<std::string> &sub_path) const
+{
+  // Get full path to sub_path (i.e. prepend subsection_path to sub_path).
+  std::vector<std::string> full_path(subsection_path);
```
tjhei: This sounds like the path ``sub_path`` is relative to the currently selected path. Can you make that obvious in the documentation in the .h please?

tests/parameter_handler/parameter_handler_subsection_exists.cc
```diff
@@ -0,0 +1,118 @@
...
+#include <deal.II/base/parameter_handler.h>
+
+#include <iomanip>
+
+#include "../tests.h"
+
+int
+main()
+{
+  try
```
tjhei: no need to catch exceptions.

elauksap: Many of the functions used therein (`declare_entry`, `print_parameters`) may possibly throw exceptions, do you think it is redundant to keep the `try/catch` block?

elauksap:
Thanks @bangerth and @tjhei for the review! Let me know if there is need to address further changes.

tests/parameter_handler/parameter_handler_subsection_exists.cc
masterleinad: The default exception handler is good enough to give meaningful error messages. We (usually) only catch in tests if we want to check that the correct exception is thrown.

PR #9446, opened 2020-01-29
Add ParameterHandler::declare_entry flag to append pattern description to entry documentation

elauksap:
Another small modification to the `ParameterHandler` class that allows to attach the pattern description to documentation entries through an optional input to the `declare_entry()` method.

This can be useful to print the admissible values along with the parameter documentation.

A small test has also been implemented.

### DIFF
--- a/doc/news/changes/minor/20200129PasqualeClaudioAfrica
+++ b/doc/news/changes/minor/20200129PasqualeClaudioAfrica
@@ -1,0 +1,4 @@
+New: ParameterHandler::declare_entry() can now optionally append
+the pattern description along with the entry documentation.
+<br>
+(Pasquale Claudio Africa, 2020/01/29)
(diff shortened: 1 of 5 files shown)


bangerth:
I think this is the wrong place. It should be part of the `print_parameters()` function. For example, I believe that when outputting the documentation in latex format, the pattern is also printed.

elauksap:
> For example, I believe that when outputting the documentation in latex format, the pattern is also printed.

Yes it is right! But what if the user would like to print the pattern description only for specific entries, *e.g.* only those with a `Patterns::Selection` pattern?

If it was part of the `print_parameters()` method it would be printed either for all parameters, or for none of them.

elauksap:
Edit: I figured out this could not work how I thought of. It's better to close this PR!

PR #9447, opened 2020-01-29
Update PolynomialsAdini to derive from ScalarPolynomialsBase

### LINKED ISSUE #9438: PolynomialsAdini does not derive from ScalarPolynomialsBase
I just saw this while working on another PR. It looks like `PolynomialsAdini` hasn't seen much love in a while.

I assign this to myself, but I'm opening an issue so I don't forget. :stuck_out_tongue:

GrahamBenHarper:
This closes #9438 

This changes `PolynomialsAdini` to fit within the context of the polynomial class polymorphism by implementing a `clone` function, calling the `ScalarPolynomialsBase` class constructor, overriding abstract functions, and adding a test for it as well.

While it may not be optimal for fourth-order problems due to the availability of the C0IP method described in step-47, we can keep it for anybody who may be interested and give it a little love too :)

### DIFF
--- a/include/deal.II/base/polynomials_adini.h
+++ b/include/deal.II/base/polynomials_adini.h
@@ -20,6 +20,7 @@
 #include <deal.II/base/config.h>
 
 #include <deal.II/base/point.h>
+#include <deal.II/base/scalar_polynomials_base.h>
 #include <deal.II/base/table.h>
 #include <deal.II/base/tensor.h>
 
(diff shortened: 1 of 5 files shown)


include/deal.II/base/polynomials_adini.h
```diff
@@ -59,42 +64,51 @@ class PolynomialsAdini
...
    * Consider using evaluate() instead.
    */
+  Tensor<2, dim>
+  compute_grad_grad(const unsigned int i, const Point<dim> &p) const;
+
+  /**
+   * Return the name of the space, which is <tt>PolynomialSpace</tt>.
+   */
+  std::string
+  name() const override;
```
bangerth: Shouldn't this be something involving the word "Adini"?

include/deal.II/base/polynomials_adini.h
```diff
@@ -134,6 +150,15 @@ class PolynomialsAdini
 
 
 
+template <int dim>
+inline std::string
+PolynomialsAdini<dim>::name() const
+{
+  return "PolynomialsAdini";
```
bangerth: ah, indeed it is :-)

include/deal.II/base/polynomials_adini.h
GrahamBenHarper: Ah good catch! You can see where I copied it from :stuck_out_tongue:

PR #9448, opened 2020-01-30
Update GridGenerator::half_hyper_shell documentation for correctness

### LINKED ISSUE #9442: Issue with boundary indicator in GridGenerator::half_hyper_shell
The documentation of half_hyper_shell grid generator mentions that If colorize is set to true, the part of the boundary where x=0 (half of it is above the x-axis and the other half is below the x-axis if we consider the origin as the center), gets indicator 2.

I tried using that boundary indicator 2 to set constraints, but deal.ii seems to apply the constraints only to the half of the part which is above the x-axis, and not on the half which is below the x-axis.

GrahamBenHarper:
This closes #9442

Only in 2d, we additionally use the boundary indicator 3 for the edge on the bottom half of the HHS.
In 2d, the indicators are 0 for the inner circle, 1 for the outer circle, 2 for the vertical edge on the top half, and 3 for the vertical edge on the bottom half.

### DIFF
--- a/include/deal.II/grid/grid_generator.h
+++ b/include/deal.II/grid/grid_generator.h
@@ -1126,7 +1126,9 @@ namespace GridGenerator
    *
    * If colorize is set to <code>true</code>, the inner, outer, and the part
    * of the boundary where $x=0$, get indicator 0, 1, and 2,
-   * respectively. Otherwise all indicators are set to 0 (see
+   * respectively. Additionally, in 2d, the boundary indicator 3 is given to
+   * the vertical edge below the x-axis. Otherwise, if colorize is set to
+   * <code>false</code> all indicators are set to 0 (see
    * @ref GlossColorization "the glossary entry on colorization").
    *
    * All manifold ids are set to zero, and a SphericalManifold is attached
--- a/tests/grid/grid_generator_half_hyper_shell.cc
+++ b/tests/grid/grid_generator_half_hyper_shell.cc
@@ -1,0 +1,59 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2005 - 2018 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Test boundary ids for GridGenerator::half_hyper_shell()
+
+#include <deal.II/base/tensor.h>
+
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/grid_out.h>
+#include <deal.II/grid/tria.h>
+
+#include "../tests.h"
+
+
+template <int dim>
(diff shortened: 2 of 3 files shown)


include/deal.II/grid/grid_generator.h
```diff
@@ -1126,8 +1126,9 @@ namespace GridGenerator
    *
    * If colorize is set to <code>true</code>, the inner, outer, and the part
    * of the boundary where $x=0$, get indicator 0, 1, and 2,
-   * respectively. Otherwise all indicators are set to 0 (see
-   * @ref GlossColorization "the glossary entry on colorization").
+   * respectively. Additionally, in 2d, the boundary indicator 3 is given to
+   * the vertical edge below the x-axis. Otherwise all indicators are set to 0
```
bangerth: are there even any other parts of the boundary?

GrahamBenHarper: Are you referring to the line "Otherwise all indicators are set to 0"? That is when `colorize=false`, but my addition there might actually make it ambiguous.

What about "Otherwise, if colorize is set to <code>false</code> all indicators are set to 0"

bangerth: Ah yes, I did not realize that that is what it refers to. I think your text is better indeed!

bangerth: your PROPOSED text

GrahamBenHarper: Sounds good! The test should be ready in a few minutes too.

tests/grid/grid_generator_half_hyper_shell.cc
```diff
@@ -0,0 +1,83 @@
...
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/grid_out.h>
+#include <deal.II/grid/tria.h>
+
+#include "../tests.h"
+
+
+
+void
+dim_2()
```
masterleinad: Is there any reason not to write this as a template?

GrahamBenHarper: I could change it to a template. There's no particular reason other than I copypasted the guts from a test right next to this one and then slightly changed it.

tjhei:
> and 3 for the vertical edge on the bottom half.

huh, fascinating that this is not documented. I guess it makes sense to not give different ids in 3d as they form a single surface. We do make use of that id 3 in ASPECT: https://github.com/geodynamics/aspect/blob/master/source/geometry_model/spherical_shell.cc#L313

tests/grid/grid_generator_half_hyper_shell.cc
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+}
+
+
+int
+main()
+{
+  initlog();
+  deallog.depth_console(0);
```
tjhei: this line is not needed.

GrahamBenHarper: which one? the `deallog`?

tjhei: yes, sorry for not being more precise.

GrahamBenHarper: No worries! When I read code reviews I sometimes forget whether the top or bottom line is the one being reviewed. I guess it wouldn't make sense to say `main()` wasn't needed here so that's on me :stuck_out_tongue:

PR #9449, opened 2020-01-30
Avoid an internal compiler error in gcc 7.4.0.

bangerth:
Why that happens is not clear to me, but a simple replacement of assertions does the trick to get me going again...


tjhei:
fascinating. Would you mind adding comment above this line? Otherwise @masterleinad might replace it again soon. ;-)

bangerth:
Sure thing.

I also voided a local variable that would otherwise lead to an unused variable warning in release mode.

masterleinad:
> Otherwise @masterleinad might replace it again soon. ;-)

For sure!

PR #9450, opened 2020-01-30
Avoid a few warnings about unused variables.

bangerth:


bangerth:
Follow-up to #9438. (@GrahamBenHarper -- FYI.)

Issue #9426, opened 2020-01-23
Dealing with clamped boundary conditions in step-47


bangerth:
@sophy1029 -- here is another issue that would be nice to address in a "Possibilities for extensions" section at the end of the results: The program at the moment implements the kind of boundary conditions where one prescribes the value and the normal derivative of the solution. In the introduction, this is described as the "clamped case".

It would be nice if we could also discuss how to deal with the other "free" boundary case where
```
  u = g
  Delta u = h              // in the actual "free" case, we have h=0
```
If you could look into the papers we based the implementation on to see how this is handled, that would be much appreciated. We could then convert this into a section in the results.

sophy1029:
Compared to the clamped case implemented in the tutorial, in the simply supported condition, there are only contributions from interior edges instead of all edges on the LHS of the scheme. 
Then on the right hand side, there would be no contributions from symmetrization and stabilization terms, but have Delta u =h as the boundary condition (if h = 0, then on the RHS, there's only integral( f ,v )). Does it correct?
How should we add this case to the result section? Shall we provide derivations of the FEM scheme?

bangerth:
I think it would be nice if we could have a section in the "Possibilities for extensions" at the bottom of the "Results" section that derives how the weak formulation of the problem looks like for the simply supported case with `h != 0`. Do you think you could write something up for this?

sophy1029:
Yes, of course!

masterleinad:
@sophy1029 What's the status here?

sophy1029:
@masterleinad  Please check #9558 
We are checking whether the derivation for the new boundary condition case is correct.

Issue #9317, opened 2020-01-14
Double check the derivation in step-47


drwells:
A followup to the conversation in #9111: we should verify that the derivation in the introduction matches the implementation.

bangerth:
Specifically, this relates to the mismatch between terms implemented for the right hand side and the terms described in the introduction.

tjhei:
The term is correct as far as I can tell, but I will go through and fix the derivation.

bangerth:
I think that the implementation is correct but that the documentation is missing a term.

bangerth:
@sophy1029 -- you asked what there is left to do for this tutorial program. I forgot that there is also this issue here: The documentation and the implementation do not match for the right hand side. If you had a bit of time, we could use help in figuring out which part is correct (I think that the implementation is correct) and updating the other side. You probably just need to compare this with what the papers on the topic say.

bangerth:
Fixed by #9451.

PR #9451, opened 2020-01-30
step-47: switch the jump and average in one term of the scheme.

sophy1029:
Should we switch the jump and average in the second face term of the finite element scheme?

### DIFF
--- a/examples/step-47/doc/intro.dox
+++ b/examples/step-47/doc/intro.dox
@@ -334,8 +334,8 @@ where
 \\
 &
  -\sum_{e \in \mathbb{F}} \int_{e}
-  \jump{ \frac{\partial^2 v_h}{\partial \mathbf n^2}}
-  \average{\frac{\partial u_h}{\partial \mathbf n}} \ ds
+  \jump{ \frac{\partial^2 u_h}{\partial \mathbf n^2}}
+  \average{\frac{\partial v_h}{\partial \mathbf n}} \ ds
  -\sum_{e \in \mathbb{F}} \int_{e}
  \average{\frac{\partial^2 v_h}{\partial \mathbf n^2}}
  \jump{\frac{\partial u_h}{\partial \mathbf n}} \ ds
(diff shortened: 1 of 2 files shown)


drwells:
See also #9317.

examples/step-47/doc/intro.dox
```diff
@@ -334,8 +334,8 @@ where
 \\
 &
  -\sum_{e \in \mathbb{F}} \int_{e}
-  \jump{ \frac{\partial^2 v_h}{\partial \mathbf n^2}}
-  \average{\frac{\partial u_h}{\partial \mathbf n}} \ ds
+  \jump{ \frac{\partial^2 u_h}{\partial \mathbf n^2}}
+  \average{\frac{\partial v_h}{\partial \mathbf n}} \ ds
```
bangerth: Was this just wrong in the documentation before?

We like to state problems with the test function multiplied from the left instead of the right. Would you mind just switching the two terms?

examples/step-47/doc/intro.dox
```diff
@@ -349,6 +349,9 @@ where
 and
 @f{align*}{
 \mathcal{F}(v_h)&:=\sum_{K \in \mathbb{T}}\int_{K} v_h f \ dx
+-
+\sum_{e \in \mathbb{F}, e\subset\partial\Omega}
+\int_e \frac{\partial^2 v_h}{\partial \mathbf n^2}} j \ ds
```
bangerth: I guess this term is in the papers but just not our documentation?

examples/step-47/doc/intro.dox
sophy1029: I found it's different from the scheme in the paper.
Yes, I can put the test function on the left.

examples/step-47/doc/intro.dox
sophy1029: This term is not in the paper since the paper deals with 0 boundary conditions.
I suppose we have this term in the implementation.

bangerth: Have you thought about whether it's correct?

bangerth:
@sophy1029 -- let us know when you're done with the modifications we talked about this morning!

sophy1029:
@bangerth  On the LHS of the scheme, the second term is jump of dv/dn times the average.
I add the extra term from symmetrization to the RHS and use average/jump notation on RHS.

bangerth:
Thanks. Are you still thinking about #9426?

sophy1029:
Oh. Yes. I will do it this week.

PR #9453, opened 2020-01-31
Fix a few misprints in the SolutionTransfer documentation.

simonsticko:


tjhei:
Thanks! Documentation only, so I will just merge.

PR #9457, opened 2020-02-01
Improve some operations in symmetric tensor and tensor

jppelteret:
When working on #9456 I spotted a couple of things in the tensor classes that could be useful improvements. The first is to use our internal number converter instead of calling the tensor `Number` constructors directly. IIRC this was sometimes important when considering operations done with AD numbers, and should have no additional overhead when using arithmetic types or vectorised numbers. @kronbichler, since you added these lines in #8428 I thought it best to get you to look at the suggested changes in the first commit. The second commit just standardises how we create some intermediate tensors in some operations.

### DIFF
--- a/include/deal.II/base/symmetric_tensor.h
+++ b/include/deal.II/base/symmetric_tensor.h
@@ -3190,17 +3190,18 @@ DEAL_II_CONSTEXPR inline DEAL_II_ALWAYS_INLINE SymmetricTensor<2, dim, Number>
   switch (dim)
     {
       case 1:
-        tmp.data[0] = Number(1);
+        tmp.data[0] = internal::NumberType<Number>::value(1);
         break;
       case 2:
-        tmp.data[0] = tmp.data[1] = Number(1);
+        tmp.data[0] = tmp.data[1] = internal::NumberType<Number>::value(1);
         break;
       case 3:
-        tmp.data[0] = tmp.data[1] = tmp.data[2] = Number(1);
+        tmp.data[0] = tmp.data[1] = tmp.data[2] =
+          internal::NumberType<Number>::value(1);
         break;
       default:
         for (unsigned int d = 0; d < dim; ++d)
-          tmp.data[d] = Number(1);
+          tmp.data[d] = internal::NumberType<Number>::value(1);
     }
   return tmp;
 }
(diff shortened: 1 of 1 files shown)


include/deal.II/base/symmetric_tensor.h
```diff
@@ -3247,7 +3248,8 @@ deviator_tensor()
   // fill the elements treating the diagonal
   for (unsigned int i = 0; i < dim; ++i)
     for (unsigned int j = 0; j < dim; ++j)
-      tmp.data[i][j] = Number((i == j ? 1 : 0) - 1. / dim);
+      tmp.data[i][j] =
+        internal::NumberType<Number>::value((i == j ? 1 : 0) - 1. / dim);
```
tjhei: wouldn't it make more sense to write 1. and 0. (here and in other places)?

bangerth: I don't think it matters -- the `NumberType<T>::value` functions generally take a `T` (or related scalar) as argument, so it is automatically converted if you write `1` instead of `1.0`. Here, it is converted to `double` because you subtract a `double` from an `int`.

But I'll let @jppelteret comment on this as well.

jppelteret: I think that what @tjhei is suggesting is that we write the first term of `1-1.0/dim` as a double so that we don't need to do an unnecessary implicit `int`->`double` conversion there. I don't know enough about compilers to understand if it would do this by itself and without a performance loss. I'd be happy to follow though with this if there's something to be gained by doing so.

jppelteret: In the other places, where we do (or want to do) `internal::NumberType<Number>::value(1)`, then I'd say that @bangerth's description of the conversion is correct. But we never use `SymmetricTensor<rank,dim, int>`, right? So we could again just circumvent the need for a conversion in the most common case, that where the `Number` is a `double`, by initialising with a double in the first place.

bangerth: For constants, the conversion will actually be done at compile time. I don't think anyone suggested that the change would make things faster or more correct, it's just a matter of stating explicitly what actually happens implicitly. I think it's not a big deal here.

The only case where I could come up with an issue is if you had a type `FP` and wanted to create a `SymmetricTensor<2,dim,FP>` and `FP` has two constructors of which one takes a `double` and the other `int`, and the latter does something completely different than the former and not what we actually want. That's a bit hard to imagine, and would surely be poor design.

I'm tempted to say that we should just merge, and that if anyone wanted to be more explicit, that we address this in a follow-up. It's for sure unrelated to what this patch does in particular.

PR #9461, opened 2020-02-02
Add a missing @endcode in DoFTools documentation

### LINKED ISSUE #9459: [Step-20] DoFTools::count_dofs_per_component() is deprecated since v9.1.1?
Step-20 says:

>The next thing is that we want to figure out the sizes of these blocks so that we can allocate an appropriate amount of space. To this end, we call the `DoFTools::count_dofs_per_component() `function that counts how many shape functions are non-zero for a particular vector component. We have dim+1 vector components, and `DoFTools::count_dofs_per_component()` will count how many shape functions belong to each of these components.

However, there is no documentation for `DoFTools::count_dofs_per_component()` in the `developer` version of the tutorial. Has this function been deprecated? If yes, what is its replacement?

masterleinad:
Fixes #9459.

### DIFF
--- a/include/deal.II/dofs/dof_tools.h
+++ b/include/deal.II/dofs/dof_tools.h
@@ -814,6 +814,7 @@ namespace DoFTools
    *      const Point<dim> &face_center = cell->face(face_index)->center();
    *      return 0 < face_center[0];
    *    };
+   * @endcode
    */
   template <typename DoFHandlerType,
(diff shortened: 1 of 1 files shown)


drwells:
You beat me to it by only a few minutes :)

masterleinad:
> You beat me to it by only a few minutes :)

Ah, sorry. :slightly_smiling_face:

PR #9465, opened 2020-02-02
Fix types in ConsensusAlgorithmProcess

peterrum:
nothing special `int` -> `unsigned int`...


peterrum:
Could I get another `/rebuild`? The tester has not picked up the last one...

PR #9466, opened 2020-02-02
Fix an overzealous assertion related to Dataout

bangerth:
As reported on the mailing list.

Now off to find some bar in Brussels to watch the Super Bowl :-)


drwells:
Is there any way that we could come up with a smaller test case? If its not much work I think its worth doing.

bangerth:
I'll have to think about this. We need a test where the first cell with material_id=1 is on the boundary, and the first cell with material_id=2 is not. I suppose I could finagle something like this -- though I do like the fact that this test checks a whole lot of other things as well, namely the Abaqus reader and that the `merge()` function actually works as expected.

drwells:
> I do like the fact that this test checks a whole lot of other things as well, namely the Abaqus reader and that the merge() function actually works as expected.

Thats a good reason to keep it as-is.

PR #9467, opened 2020-02-03
Use cell->face_iterators() or GeometryInfo::face_indices() in the tutorials.

### LINKED ISSUE #8563: Use cell->face_iterators() in tutorials

bangerth:
This fixes #8563. I had hoped that we could use the idiom
```
  for (const auto &face : cell->face_iterators())
```
in more places, but it turns out that we actually need the face index in quite a number of places to get at the neighbor of a cell in that direction, or doing any number of other operations. So I took the second best option, which is to replace the explicit loop by
```
  for (unsigned int face_no : GeometryInfo<dim>::face_indices())
```

This all is predicated on the assumption that range-based for loops are the idiomatic way to do things these days. It's definitely shorter than before.

If one of you native speakers could also comment on whether `face_no` would be read as the "face number", that would also be helpful. I know that the abbreviation "no" is read as "number" in at least German and Italian, but we've been using it for so many years in deal.II that I can't really say any more whether that's a deal.II-ism, or whether a typical English speaker would recognize that this variable name corresponds to an index.

### DIFF
--- a/examples/step-14/step-14.cc
+++ b/examples/step-14/step-14.cc
@@ -2273,9 +2273,7 @@ namespace Step14
       // After computing the cell terms, turn to the face terms. For this,
       // loop over all faces of the present cell, and see whether
       // something needs to be computed on it:
-      for (unsigned int face_no = 0;
-           face_no < GeometryInfo<dim>::faces_per_cell;
-           ++face_no)
+      for (unsigned int face_no : GeometryInfo<dim>::face_indices())
         {
           // First, if this face is part of the boundary, then there is
           // nothing to do. However, to make things easier when summing up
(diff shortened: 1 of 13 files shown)


bangerth:
Builds on #9415 .

masterleinad:
> If one of you native speakers could also comment on whether `face_no` would be read as the "face number", that would also be helpful. I know that the abbreviation "no" is read as "number" in at least German and Italian, but we've been using it for so many years in deal.II that I can't really say any more whether that's a deal.II-ism, or whether a typical English speaker would recognize that this variable name corresponds to an index.

https://en.wikipedia.org/wiki/Numero_sign is proof enough for me. Honestly, I would not have expected that in abbreviation in German (but rather Nr.).

Issue #7730, opened 2019-02-14
No doxygen for BoundingBox::NeighborType enumerator


davydden:
[subj](https://www.dealii.org/developer/doxygen/deal.II/classBoundingBox.html#a62544200c2e2498fc07845eb3f2c44fa).

kronbichler:
Uncritical, postponing for the 9.2 release.

bangerth:
As reported by @peterrum over at #9486:

> The following enum classes seem to make also problems: Storage, ComponentExtractor and SIMDComparison...

bangerth:
I've tried to figure out what's going on here several times but can't seem to make progress. There is no error in `doxygen.log`, and no other indication of what's going wrong. I've also tried to replace the `enum class` by `enum` or `class` to see whether doxygen likes that better, but the answer is no for `enum`. For `class`, it seems to work, but that's of course not useful. I'm not sure what we should do.

tjhei:
I changed the settings
```
--- a/doc/doxygen/options.dox.in
+++ b/doc/doxygen/options.dox.in
@@ -39,12 +39,12 @@ BUILTIN_STL_SUPPORT    = YES
 # Build related configuration options
 #---------------------------------------------------------------------------
 
-EXTRACT_ALL            = NO
+EXTRACT_ALL            = YES
 EXTRACT_PRIVATE        = YES
 EXTRACT_STATIC         = YES
 EXTRACT_LOCAL_CLASSES  = YES
-HIDE_UNDOC_MEMBERS     = YES
-HIDE_UNDOC_CLASSES     = YES
+HIDE_UNDOC_MEMBERS     = NO
+HIDE_UNDOC_CLASSES     = NO
 SORT_MEMBER_DOCS       = NO
```
and now it shows up for me. I am not sure which of those settings did it, though. Worth changing the defaults for?

bangerth:
Which of these settings is it?

tjhei:
I played with this some more (including other extract_* flags we don't have in our options, setting ``@file``, settings ``@relates``, etc.): The only thing that makes this enum show up is ``EXTRACT_ALL=YES`` or putting it into a namespace (*).
*: we delete the "namespace dealii{" for doxygen, which makes this enum a global symbol. A workaround would be to move it into a namespace (or class scope).
Or, we could just enable ``EXTRACT_ALL``?

PR #9468, opened 2020-02-03
Update documentation of NeighborType.

bangerth:
Found while looking at #7730 (though not finding what the problem is :-( ).


peterrum:
> Found while looking at #7730 (though not finding what the problem is :-( ).

The following enum classes seem to make also problems: `Storage`, `ComponentExtractor` and `SIMDComparison`...

PR #9470, opened 2020-02-03
Address a question on the mailing list about step-20.

bangerth:


kronbichler:
This looks good, but you will need to delete the trailing whitespace.

examples/step-20/doc/intro.dox
```diff
@@ -358,17 +358,25 @@ program. We will therefore not comment much on it below.
...
-symmetric. In other words: the Conjugate Gradient method is not going to
-work since it is only applicable to problems in which the matrix is
-symmetric and positive definite.
+it. The problem here is that the matrix possesses two undesirable properties: 
+- It is <a href="https://en.wikipedia.org/wiki/Definiteness_of_a_matrix">indefinite</a>,
+  i.e., it has both positive and negative eigenvalues.
+  We don't want to prove this property here, but note that this is true
+  for all matrices of the form 
+  $\left(\begin{array}{cc} M & B \\ B^T & 0 \end{array}\right)$
+  such as the one here where $M$ is positive semi-definite.
```
drwells: Why is `M` only semi-definite? Isn't it a mass matrix?

drwells: It might be easier, rather than worrying about spectra, to just cite the fact that it has a zero block as sufficient to establish indefiniteness for the matrix.

bangerth: I meant to say that it's true for all matrices of this kind where M is *at least* positive semidefinite. But that didn't come out right. I just changed it to positive definite.

I don't think it's that much clearer because I think that the zero block by itself is not the issue -- don't you also have to have that the top right block is the transpose of the bottom left? I mean, I could come up with a system of equations in 3 variables where one just happens to not appear in one of the three equations. Is that automatically an indefinite system? (In your reply elsewhere, you quote the issue of the zero pivot, but that too isn't quite clear to me. That's because, first, "pivot" requires you to think about the solution of the linear system in terms of Gaussian elimination, which is specifically not what we do. Second, it's not clear to me what the relationship between pivot and diagonal entry is. The pivots are typically chosen to be diagonal entries, but their value might only be reached after a few elimination steps and may not be the original value on the diagonal any more.)

bangerth:
I made the one change semidefinite->definite

examples/step-20/doc/intro.dox
drwells: You're right. The proposed wording fixes this.

drwells: I updated my answer in the forum to fix my silly mistake with pivots!

Issue #9452, opened 2020-01-31
Refactor Utilities::MPI::compute_n_point_to_point_communications


peterrum:
The function `Utilities::MPI::compute_n_point_to_point_communications()` allocates for per process a vector of size of number of processes:
https://github.com/dealii/dealii/blob/4cb48def4344b1c351e687c70cfeb073312c2bc6/source/base/mpi.cc#L441

For thousands of processes this is too expensive. In these cases we should call  `Utilities::MPI::compute_point_to_point_communication_pattern()`, which uses an MPI-3.0 implementation,  to compute an array processes and return the size of this.

PR #9471, opened 2020-02-03
Move implementation of ConsenusAlgorithm into mpi.templates.h

peterrum:
This PR does nothing than moving the implementation from `mpi.cc` to `mpi.templates.h` so that users can write their own `ConsensusAlgorithmProcess`. The next PR will actually contain something useful (which needs this changes)!


tjhei:
somewhat related: have you considered moving Consensus* into a separate header/.cc? It has becoming quite a lot of code that most users don't need.

peterrum:
> somewhat related: have you considered moving Consensus* into a separate header/.cc? It has becoming quite a lot of code that most users don't need.

If you think it would be better, I can do that. But will lead again to cyclic dependency of the files...

tjhei:
> If you think it would be better, I can do that. But will lead again to cyclic dependency of the files...

That's annoying. Is this because some implementation inside mpi.cc depends on the consensus algorithms?

peterrum:
@tjhei Yes. The function `compute_point_to_point_communication_pattern` is using it here:

https://github.com/dealii/dealii/blob/27eb58067b3dc87d5532377cbbf90e490e0a5ce9/source/base/mpi.cc#L304-L309

and probably soon `compute_n_point_to_point_communications` (see issue #9452).

tjhei:
> Yes. The function `compute_point_to_point_communication_pattern` is using it here:

But that is in the mpi.cc only, so we should be able to move it.

PR #9472, opened 2020-02-03
Add function to print statistics over MPI for TimerOutput

kronbichler:
For #8319, I want to print some statistics over times spent in various sections with MPI, but without adding the barriers of the existing infrastructure. My idea is to add a `TimerOutput::print_summary_of_wall_time_statistics()` function that collects data with `MPI::MinMaxAvg` and prints that information. Sample output (from step-67) is as follows:
```
+-------------------------------------------+------------+------+------------+------------+------+
| Total wallclock time elapsed              |     13.54s |    9 |     13.54s |     13.54s |    0 |
|                                           |            |  min |            |            |  max |
| Section                       | no. calls |   min time | rank |   avg time |   max time | rank |
+-------------------------------+-----------+------------+------+------------+------------+------+
| abc                           |         1 |  1.63e-06s |    3 | 1.766e-06s | 1.958e-06s |    0 |
| compute errors                |        11 |   0.05191s |    1 |   0.05339s |   0.05415s |   11 |
| compute time step             |      2558 |     8.858s |   11 |     8.933s |     9.032s |    1 |
| compute transport speed       |       513 |    0.1651s |    1 |    0.2625s |    0.3377s |   11 |
| output                        |      2558 |     3.934s |    1 |     3.935s |     3.936s |   11 |
| rk_stage - integrals          |     12790 |     6.787s |   10 |     6.904s |     7.014s |    0 |
| rk_stage - inv mass + vec upd |     12790 |     1.913s |    3 |     1.939s |     1.983s |   10 |
+-------------------------------+-----------+------------+------+------------+------------+------+
```
Any suggestions for a better name than my longish one? If we agree, I will add a test and a changelog entry.


source/base/timer.cc
```diff
@@ -861,6 +861,96 @@ TimerOutput::print_summary() const
...
+
+        // resize the array so that it is always of the same size
+        unsigned int pos_non_space = name_out.find_first_not_of(' ');
+        name_out.erase(0, pos_non_space);
+        name_out.resize(max_width, ' ');
+        out_stream << "| " << name_out;
+        out_stream << "| ";
+        out_stream << std::setw(9);
+        out_stream << i.second.n_calls << " |";
+        data = Utilities::MPI::min_max_avg(i.second.total_wall_time, mpi_comm);
```
peterrum: I plan to extend `Utilities::MPI::min_max_avg` so that it operates on a vector...

kronbichler:
Two things that I would like to put up for discussion. One is the way we present the data, as we include both the minimum and the rank where the minimum is recorded. Rather than printing a vertical line to indicate a separate column, we could instead do like this:
```
+-------------------------------------------+------------------+------------+------------------+
| Total wallclock time elapsed              |     13.54s     9 |     13.54s |     13.54s     0 |
|                                           |                  |            |                  |
| Section                       | no. calls | min time on rank |   avg time | max time on rank |
+-------------------------------+-----------+------------------+------------+------------------+
| abc                           |         1 |  1.63e-06s     3 | 1.766e-06s | 1.958e-06s     0 |
| compute errors                |        11 |   0.05191s     1 |   0.05339s |   0.05415s    11 |
| compute time step             |      2558 |     8.858s    11 |     8.933s |     9.032s     1 |
| compute transport speed       |       513 |    0.1651s     1 |    0.2625s |    0.3377s    11 |
| output                        |      2558 |     3.934s     1 |     3.935s |     3.936s    11 |
| rk_stage - integrals          |     12790 |     6.787s    10 |     6.904s |     7.014s     0 |
| rk_stage - inv mass + vec upd |     12790 |     1.913s     3 |     1.939s |     1.983s    10 |
+-------------------------------+-----------+------------------+------------+------------------+
```
The second question is whether we also would like to print the 10 and 90 percentiles of the statistics (or 25/75 percentiles). This would add two more columns of the min/max type (as I would print the rank which is close to the percentile value), so the table gets a bit wide. Given that this information can be useful, I would make it optional. To compute those quantiles, I would need to gather all times on rank 0, compute the quantiles there, and broadcast the result. This makes it a somewhat more expensive than the min/max/avg (especially on 100k ranks because MPI_Gather is O(n) on the main rank, rather than O(log(n)) for Allreduce. But timers are not printed often, so I cannot imagine it be a problem.

source/base/timer.cc
peterrum: @kronbichler Could you use - once PR #9473 is merged - the new function?

kronbichler: I will use it, even though it is a bit of a micro optimization as we compare say 10-20 Allreduce operations to writing the output which happens very rarely in the program by comparison.

peterrum:
> One is the way we present the data, as we include both the minimum and the rank where the minimum is recorded. Rather than printing a vertical line to indicate a separate column, we could instead do like this:

I am not a fan of either formats. I would prefer s.th. what is simple to parse.

> The second question is whether we also would like to print the 10 and 90 percentiles of the statistics (or 25/75 percentiles). This would add two more columns of the min/max type (as I would print the rank which is close to the percentile value), so the table gets a bit wide. Given that this information can be useful, I would make it optional. 

I like this idea!

> To compute those quantiles, I would need to gather all times on rank 0, compute the quantiles there, and broadcast the result. This makes it a somewhat more expensive than the min/max/avg (especially on 100k ranks because MPI_Gather is O(n) on the main rank, rather than O(log(n)) for Allreduce. But timers are not printed often, so I cannot imagine it be a problem.

Could I suggest binning with `ConsensusAlgorithm`/`Dictionary` as an alternative? Each bin is responsible for a certain range between min and max. To determine the bin that contains a certain percentage we could use (ex)scan. What do you think, @kronbichler? I worst case it becomes the algorithm you describe. If you like the idea, I could implement a draft...

kronbichler:
> I am not a fan of either formats. I would prefer s.th. what is simple to parse.

At this point I would put the human readability first. It's not that this format terribly complicated to parse - if you do not have spaces in the section names, you can go by column (and cut the annoying s), or you use the custom delimiter on `|`. As an alternative we could of course also add a parse-friendly variant that prints a csv format, it's not terribly complicated to do so.

kronbichler:
> Could I suggest binning with `ConsensusAlgorithm/Dictionary` as an alternative? 

Do you want to identify additional statistical quantities via bins? Or is it just for the sake of efficiency to avoid `MPI_Gather` (note it is `Gather`, not `Allgather`)? In the latter case, I'd say there is better use for your time elsewhere rather than to over-engineer this rather unimportant place.

kronbichler:
I have now added the option to print quantiles. The output looks like this (a bit wide, as expected):
```
+-------------------------------------------+------------------+------------------+------------+------------------+------------------+
| Total wallclock time elapsed              |     13.64s     6 |     13.64s     4 |     13.64s |     13.64s     7 |     13.64s     0 |
|                                           |                  |                  |                               |                  |
| Section                       | no. calls |   min time  rank |   0.1-tile  rank |   avg time |   0.9-tile  rank |   max time  rank |
+-------------------------------------------+------------------+------------------+------------+------------------+------------------+
| abc                           |         1 | 1.653e-06s     7 | 1.663e-06s    10 | 1.795e-06s | 1.934e-06s     5 | 2.048e-06s     3 |
| compute errors                |        11 |   0.05155s     1 |   0.05159s     2 |   0.05268s |   0.05322s     3 |   0.05328s     5 |
| compute transport speed       |       513 |    0.1758s     1 |    0.1776s     2 |    0.2636s |    0.3238s     9 |    0.3274s    11 |
| evaluate time step            |      2558 |     8.962s    11 |     8.966s     9 |     9.027s |     9.114s     2 |     9.115s     1 |
| output                        |      2558 |     3.941s     2 |     3.941s     1 |     3.942s |     3.942s     4 |     3.943s     5 |
| rk_stage - integrals          |     12790 |     6.886s    10 |     6.925s    11 |      6.99s |     7.079s     2 |     7.105s     0 |
| rk_stage - inv mass + vec upd |     12790 |     1.914s     3 |      1.92s     0 |      1.95s |     1.983s     1 |     1.992s    10 |
+-------------------------------------------+------------------+------------------+------------+------------------+------------------+
```

include/deal.II/base/timer.h
```diff
@@ -574,6 +574,32 @@ class Timer
...
+ * actually adding the barriers. While this approach is still affected by
+ * imbalances between different MPI processes, its output is not the arbitrary
+ * time of rank 0, but the minimum, average and maximum of the MPI results,
+ * using information from Utilities::MPI::MinMaxAvg. As the data is also
+ * equipped with the rank id where the minimum and maximum are attained, this
+ * approach allows to identify on which ranks certain slowdowns occur. In case
+ * some imbalance between the MPI ranks from one section to the next can be
+ * tolerated, this strategy can hence be advantageous over the barrier variant
+ * as it does not synchronize the program in places where it is not
+ * necessary. The usage of this variant is to initialize the output object
```
tamiko:
```suggestion
 * necessary. In order to use this variant is to initialize the output object
```

include/deal.II/base/timer.h
```diff
@@ -574,6 +574,32 @@ class Timer
...
+ *                      TimerOutput::never,
+ *                      TimerOutput::wall_times);
+ * @endcode
+ * and then call
+ * @code
+ *   timer.print_summary_statistics(MPI_COMM_WORLD);
+ * @endcode
+ * Here, the output is written to the <code>pcout</code> object of type
+ * ConditionalOStream passed to the constructor, making sure the information
+ * is only printed once. See step-67 for an example usage of this variant.
```
tamiko: I will most definitely change step-69 to use the new variant as well :-)

source/base/timer.cc
```diff
@@ -861,6 +861,190 @@ TimerOutput::print_summary() const
...
 
+void
+TimerOutput::print_summary_of_wall_time_statistics(const MPI_Comm mpi_comm,
+                                                   const double quantile) const
+{
+  // we are going to change the precision and width of output below. store the
+  // old values so we can restore it later on
+  const std::istream::fmtflags old_flags = out_stream.get_stream().flags();
+  const std::streamsize old_precision    = out_stream.get_stream().precision();
+  const std::streamsize old_width        = out_stream.get_stream().width();
```
tamiko: What about
```
boost::io::ios_flags_saver ifs(out_stream.get_stream());
```

kronbichler: Looks good - I have used this in the new function here and further above in `print_summary()` to remove 12 lines of code :wink:

kronbichler:
I've addressed the comments and also slightly changed the new name: I think `print_wall_time_statistics()` is better than `print_summary_of_wall_time_statistics()` (assuming that statistics already implies some kind of summary).

include/deal.II/base/timer.h
kronbichler: > I will most definitely change step-69 to use the new variant as well :-)

The second order variant of step-69 was one of the applications where this has bothered me - in addition to many of my large-scale scaling tests where I wanted the imbalance to be propagated between different timer sections and visualize it.

PR #9473, opened 2020-02-03
Add Utilities::MPI::min_max_avg() for vectors

peterrum:
The original (scalar) `Utilities::MPI::min_max_avg()` calls now the new function.


doc/news/changes/minor/20200203PeterMunch
```diff
@@ -0,0 +1,4 @@
+New: The function Utilities::MPI::min_max_avg() can now compute sum, average, minimum, maximum,
```
kronbichler:
```suggestion
New: The function Utilities::MPI::min_max_avg() can now compute the sum, average, minimum, maximum, 
```

doc/news/changes/minor/20200203PeterMunch
```diff
@@ -0,0 +1,4 @@
+New: The function Utilities::MPI::min_max_avg() can now compute sum, average, minimum, maximum, 
+process id of minimum and maximum for each entry of a vector.
```
kronbichler:
```suggestion
and the process id of the minimum and maximum for each entry of a vector.
```

include/deal.II/base/mpi.h
```diff
@@ -719,6 +719,22 @@ namespace Utilities
...
     min_max_avg(const double my_value, const MPI_Comm &mpi_communicator);
 
+    /**
+     * Same as above but returning the sum, average, minimum, maximum,
+     * process id of minimum and maximum as a collective operation of on the
+     * given MPI @ref GlossMPICommunicator "communicator" @p mpi_communicator
+     * for each entry of the vector.
+     *
+     * @note This function performs a single reduction sweep, and does not call
+     *   for each vector entry the above function.
```
kronbichler: Simply say `This function performs a single reduction sweep.` This clearly states that we do not repeatedly call the other function.

source/base/mpi.cc
```diff
@@ -528,30 +528,46 @@ namespace Utilities
...
+          std::vector<MinMaxAvg> result(my_values.size());
+
+          for (unsigned int i = 0; i < my_values.size(); i++)
+            {
+              result[i].sum       = my_values[i];
+              result[i].avg       = my_values[i];
+              result[i].min       = my_values[i];
+              result[i].max       = my_values[i];
+              result[i].min_index = 0;
+              result[i].max_index = 0;
```
kronbichler: Even though I am normally against adding new memory allocations for cheap operations like the non-parallel case here (note that I added an additional switch in #9472 to avoid creating the MPI data structures, which showed up in my profiles due to several memory allocations), due to the low frequency by which we call `min_max_avg` I think it is indeed ok here. Well, the only case I can imagine it would hurt is if we have `Timer::stop()` inside an inner loop (taking less than 1us), and in that case one should not use the standard timer classes anyway.

peterrum:
@kronbichler Thanks for the review. I have made the changes.

include/deal.II/base/mpi.h
```diff
@@ -719,6 +719,21 @@ namespace Utilities
...
+     * process id of minimum and maximum as a collective operation of on the
+     * given MPI @ref GlossMPICommunicator "communicator" @p mpi_communicator
+     * for each entry of the vector.
+     *
+     * @note This function performs a single reduction sweep.
+     *
+     * @req Size of the input vector has to be the same on all processes.
+     */
+    std::vector<MinMaxAvg>
+    min_max_avg(const std::vector<double> my_values,
```
kronbichler: clang-tidy rightfully complains that you pass the vector by value rather than reference.

peterrum:
> I'm not a big fan of this function

@bangerth I hope your comment was many referring to the `std::vector`. I have changed it to `ArrayView`.

include/deal.II/base/mpi.h
```diff
@@ -719,6 +719,22 @@ namespace Utilities
...
+     * given MPI @ref GlossMPICommunicator "communicator" @p mpi_communicator
+     * for each entry of the vector.
+     *
+     * @note This function performs a single reduction sweep.
+     *
+     * @req Size of the input vector has to be the same on all processes.
+     */
+    void
+    min_max_avg(const ArrayView<const double> &my_values,
+                const ArrayView<MinMaxAvg> &   result,
```
bangerth: Oh, I had gotten that wrong because I had (wrongly) had the model of the `max` function in mind where the returned object is really just a single scalar of the same type as the input.

This works, but I would also have been ok if the function returned a `std::vector<MinMaxAvg>`. But this is fine by me.

tests/mpi/minmaxavg_vector_01.cc
```diff
@@ -0,0 +1,67 @@
...
+  else if (myid == 2)
+    values = std::vector<double>{3.0, 2.0, 1.0};
+  else if (myid == 3)
+    values = std::vector<double>{3.0, 3.0, 0.0};
+
+  std::vector<Utilities::MPI::MinMaxAvg> results(values.size());
+
+  Utilities::MPI::min_max_avg(
+    ArrayView<const double>(values.data(), values.size()),
+    ArrayView<Utilities::MPI::MinMaxAvg>(results.data(), results.size()),
```
bangerth: Isn't there an automatic conversion from `std::vector` to `ArrayView`? Or at least the syntax `make_array_view(values)` and `make_array_view(results)`?

peterrum:
> This works, but I would also have been ok if the function returned a std::vector<MinMaxAvg>. But this is fine by me.

Let me create an additional function that works with vectors.

bangerth:
Several of the builds currently fail. Can you take a look?

source/base/mpi.cc
```diff
@@ -649,6 +679,27 @@ namespace Utilities
...
+        {
+          result[i].sum       = my_values[i];
+          result[i].avg       = my_values[i];
+          result[i].min       = my_values[i];
+          result[i].max       = my_values[i];
+          result[i].min_index = 0;
+          result[i].max_index = 0;
+        }
+
+      return result;
```
masterleinad:
```suggestion
```

include/deal.II/base/mpi.h
kronbichler: Given that we have been switching some other functions to return a vector in case we do some simple computation (like #9272), I would suggest we provide also this variant. I don't think it hurts to have yet another interface, as the implementation only redirects to this function.

include/deal.II/base/array_view.h
```diff
@@ -98,6 +98,8 @@ class ArrayView
    */
   ArrayView();
 
+  ArrayView(value_type &element);
+
```
peterrum: @bangerth I have tried out you suggestion regarding automatic conversion. It works for `std::vector()` but not for `double` at the moment. Does anything speaks against to introduce a new constructor in `ArrayView` which takes a reference to a `value_type`?  Then the code below could look like `min_max_avg(my_value, result, mpi_communicator);`.

If nothing speaks against that, I would create a separate PR.

bangerth: I see no reason why we shouldn't have such a constructor. I like the idea!

include/deal.II/base/mpi.h
```diff
@@ -719,6 +719,38 @@ namespace Utilities
     MinMaxAvg
     min_max_avg(const double my_value, const MPI_Comm &mpi_communicator);
 
+    /**
+     * Same as above but returning the sum, average, minimum, maximum,
+     * process id of minimum and maximum as a collective operation of on the
```
bangerth:
```suggestion
     * process id of minimum and maximum as a collective operation on the
```

include/deal.II/base/mpi.h
```diff
@@ -719,6 +719,38 @@ namespace Utilities
...
+     * @req Size of the input vector has to be the same on all processes.
+     */
+    std::vector<MinMaxAvg>
+    min_max_avg(const std::vector<double> my_value,
+                const MPI_Comm &          mpi_communicator);
+
+
+    /**
+     * Same as above but returning the sum, average, minimum, maximum,
+     * process id of minimum and maximum as a collective operation of on the
```
bangerth:
```suggestion
     * process id of minimum and maximum as a collective operation on the
```

include/deal.II/base/mpi.h
```diff
@@ -719,6 +719,38 @@ namespace Utilities
...
+     * given MPI @ref GlossMPICommunicator "communicator" @p mpi_communicator
+     * for each entry of the ArrayView.
+     *
+     * @note This function performs a single reduction sweep.
+     *
+     * @req Size of the input ArrayView has to be the same on all processes
```
bangerth: Is `@req` a thing? We use `@pre` elsewhere

peterrum:
This is ready from my side!

Issue #8206, opened 2019-05-16
create GridGenerator images


tjhei:
I think it would be helpful to have images for all the functions in the GridGenerator namespace. Ideally, we would show boundary/manifold ids as well.

PR #9474, opened 2020-02-03
Add boundary id functionality to write_svg

GrahamBenHarper:
I know this will break a few tests since it's technically an incompatibility, but I figured I'd start getting feedback first. The goal is to have `cell->face(f)->boundary_id()` also an option for output for `GridOut::write_svg()`.
**Why?** Because troubleshooting boundary conditions and related issues is a pain, and this can be used to generate pretty pictures for the documentation and #8206.

grid_out_svg_03
Here's a sample of how the output would look for the new test `grid_out_svg_03` (the font size is not currently equal to what my code outputs). Any suggestions for how I may make this more readable?

(also it looks like I might need to add a legend entry for the boundary ids)


peterrum:
As a comment, you can we different quantities in Paraview if you write with GridOut VTK files. In might be useful to only show edged, faces, or cells (see https://www.dealii.org/developer/doxygen/deal.II/structGridOutFlags_1_1Vtk.html).

peterrum:
To your figure. Maybe you could

- make the font of the boundary ids smaller,
- center the labels and
- maybe draw a circle around the label with a background color...

GrahamBenHarper:
> As a comment, you can we different quantities in Paraview if you write with GridOut VTK files. In might be useful to only show edged, faces, or cells (see https://www.dealii.org/developer/doxygen/deal.II/structGridOutFlags_1_1Vtk.html).

@peterrum that does sound like it would be a nice feature to add for GridOut SVG files. The way it currently draws the mesh is edge-by-edge, so it shouldn't be too hard as an additional PR.


> To your figure. Maybe you could
> 
>   * make the font of the boundary ids smaller,
> 
>   * center the labels and
> 
>   * maybe draw a circle around the label with a background color...

That's a good idea! The text does have a default style of `text-anchor:middle`, but the placement does not look centered. I will have to investigate that more. What size should I make the boundary ids? Half of the size of the cell indices?

GrahamBenHarper:
Interestingly enough, vertical alignment of text seems to be rather nuanced. Here's how it looks with the code now, using different image viewers.

The left is displaying the image through firefox, and the right is displaying the image through the default ubuntu image viewer.
boundary_id
It's obviously undesirable to have tags that aren't fully supported across all viewers, but is the left image closer to what you had in mind? The circle size could still be tweaked.

bangerth:
I like this! This would make nice pictures for the documentation in `GridGenerator`!

tamiko:
@GrahamBenHarper You need to update the output of the following tests:
```
The following tests FAILED:
	626 - grid/grid_out_svg_02.debug (Failed)
	756 - grid/grid_out_svg_03.debug (Failed)
	769 - grid/grid_out_svg_01.debug (Failed)
```

GrahamBenHarper:
@tamiko I'm already on it! I actually just updated `grid_out_svg_02` to include the boundary ids and I have to say it looks really neat.

The only thing I want to do before I force-push is figure out how to center the text better and how to fix the legend entry. Take a look :smiley: 
boundary_id2
Currently I'm using SVG tags that look like this for the boundary ids
```
    <circle cx="880" cy="860" r="32" />
    <text x="880" y="860" style="font-size:32px" dominant-baseline="central">1</text>
```
The only problem is that `dominant-baseline` and `alignment-baseline` tags don't always show up properly. For example, the boundary ids in that image look correct if I view the SVG in Firefox (not shown) instead of Ubuntu's image viewer (shown above). Does anybody know a good fix for that without manually moving the x and y coordinates of the text?

GrahamBenHarper:
@bangerth what do you think?

GrahamBenHarper:
Changing the baseline tags to align to "middle" does not affect the bad-looking output in the Ubuntu image viewer, but it looks absolutely perfect (regardless of zoom level) in Firefox.
boundary_id3
The tough part is this gets into compatibilities with tags and different browser and viewer versions etc etc. Perhaps I should just leave a note about this in the documentation and call it good?

include/deal.II/grid/grid_out.h
```diff
@@ -823,6 +828,7 @@ namespace GridOutFlags
         const bool         label_cell_index               = false,
         const bool         label_material_id              = false,
         const bool         label_subdomain_id             = false,
+        const bool         label_boundary_id              = false,
         const bool         draw_colorbar                  = false,
         const bool         draw_legend                    = false);
```
bangerth: The placement of the new function argument makes the change incompatible to what you had before. That's because if someone provided all of the boolean arguments before, the call is still correct but what used to be interpreted as `draw_colorbar` is now the `label_boundary_id` argument.

It's unfortunate, but the only way to do this in a backward compatible way is to put the new argument at the end.

(If we had thought this through some more, we'd have made the `label` kind of arguments an `enum` bitfield where one can select any number of bits, and just has to pass a single argument. But that's for a separate patch.

source/grid/grid_out.cc
```diff
@@ -436,6 +436,7 @@ namespace GridOutFlags
            const bool         label_cell_index,
            const bool         label_material_id,
            const bool         label_subdomain_id,
+           const bool         label_boundary_id,
            const bool         draw_colorbar,
            const bool         draw_legend)
```
bangerth: same here of course

source/grid/grid_out.cc
```diff
@@ -2461,13 +2464,68 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
                                  (height -
                                   (height / 100.) * 2. * margin_in_percent))
                           << "\"/>" << '\n';
+
+
+                      if (svg_flags.label_boundary_id)
+                        {
+                          float distance_to_camera = std::sqrt(
```
bangerth: just make it a `double`. That's what `std::sqrt` returns, after all.

Also, I think you can make it `const`.

source/grid/grid_out.cc
```diff
@@ -2461,13 +2464,68 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
...
+                            std::pow(point[0] - camera_position[0], 2.) +
+                            std::pow(point[1] - camera_position[1], 2.) +
+                            std::pow(point[2] - camera_position[2], 2.));
+                          float distance_factor =
```
bangerth: same here

source/grid/grid_out.cc
```diff
@@ -2461,13 +2464,68 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
...
+                            distance_to_camera /
+                            (2. * std::max(x_dimension, y_dimension));
+
+                          const auto font_size_this_edge =
+                            static_cast<unsigned int>(
```
bangerth: and you know the type for this variable: `unsigned int`

source/grid/grid_out.cc
```diff
@@ -2461,13 +2464,68 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
...
+                          point[1] = cell->face(faceIndex)->center()[1];
+                          point[2] = 0;
+                          projection_decomposition =
+                            svg_project_point(point,
+                                              camera_position,
+                                              camera_direction,
+                                              camera_horizontal,
+                                              camera_focus);
+
+                          unsigned int xc = static_cast<unsigned int>(
```
bangerth: const

source/grid/grid_out.cc
```diff
@@ -2461,13 +2464,68 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
...
+                            .5 +
+                            ((projection_decomposition[0] - x_min_perspective) /
+                             x_dimension_perspective) *
+                              (width -
+                               (width / 100.) * 2. * margin_in_percent) +
+                            ((width / 100.) * margin_in_percent));
+                          unsigned int yc = static_cast<unsigned int>(
```
bangerth: const

bangerth:
By the way, why do existing tests fail?

include/deal.II/grid/grid_out.h
GrahamBenHarper: Ooo good point. Even with the incompatibility changelog it could cause unexplained changes in behavior on the user's end. I'll put it at the bottom.

Would that make this no longer an incompatibility?

source/grid/grid_out.cc
GrahamBenHarper: Got it!

GrahamBenHarper:
@bangerth the existing tests only fail because I updated the style at the top of the SVG output to contain the information `circle{fill:white; stroke:black; stroke-width:2}`.

I've updated them on my end and I'll push them as soon as I have things looking perfect. :)

source/grid/grid_out.cc
GrahamBenHarper: I copied it from line 2293 here
https://github.com/dealii/dealii/blob/27f330ab0685777c0be4049263e33dc9d9f2e275/source/grid/grid_out.cc#L2293
but I can change both to `const double`s while I'm at it.

source/grid/grid_out.cc
GrahamBenHarper: Got it!

include/deal.II/grid/grid_out.h
bangerth: No: Existing code only provided at most N arguments, so the N+1st argument is left at its default, which is `false`.

source/grid/grid_out.cc
bangerth: Ah, thanks. Yes, that would be nice.

bangerth:
Ah, good point with the update to the style! 

Let me know when you're ready!

include/deal.II/grid/grid_out.h
GrahamBenHarper: Cool! I'll update the changelog then.

GrahamBenHarper:
@bangerth I think it's done now. I touched up the legend so that if every flag is enabled simultaneously there won't be any overlap of labels.

I worry that the `.2` and `.avx512` output files will be affected by this.

bangerth:
On 2/6/20 4:14 PM, Graham Harper wrote:
> I worry that the |.2| and |.avx512| output files will be affected by this.

Yes, certainly. Could you just make the exact same change by hand as you 
had made to the .output files you touched? It should just require 
copying a line into the right place in each of these files.

GrahamBenHarper:
Okay I updated it once more but I realized that those output files haven't been updated since before #8451 (they still referred to `level_number`) so they were likely not working as intended before this PR. Sorry @kronbichler!

PR #9476, opened 2020-02-04
Convert more places to use GeometryInfo::face_indices().

bangerth:
More as a follow-up to #9415 .

I'm living under the assumption that we consider range-based for loops idiomatic these days.

### DIFF
--- a/include/deal.II/base/geometry_info.h
+++ b/include/deal.II/base/geometry_info.h
@@ -1211,7 +1211,7 @@ struct GeometryInfo<0>
    * using range-based for loops of the following kind:
    * @code
    *   for (auto &cell : triangulation.active_cell_iterators())
-   *     for (auto face_index : GeometryInfo<dim>::face_indices)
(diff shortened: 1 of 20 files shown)


include/deal.II/base/geometry_info.h
```diff
@@ -1211,7 +1211,7 @@ struct GeometryInfo<0>
    * using range-based for loops of the following kind:
    * @code
    *   for (auto &cell : triangulation.active_cell_iterators())
-   *     for (auto face_index : GeometryInfo<dim>::face_indices)
+   *     for (auto face_index : GeometryInfo<dim>::face_indices())
```
bangerth: Turns out that I forgot the parentheses in my first attempt at this patch (compiler didn't like that) and had to do a global search-replace -- which uncovered that I had made the same mistake writing the documentation in #9415...

bangerth:
@GrahamBenHarper : There are actually *many many* more places that are wrapped onto multiple lines that I haven't touched because they're not easily replaceable by scripts. We'll fix them over the coming years :-)

PR #6830, opened 2018-06-23
Restore backward-compatibility for #6816

masterleinad:
Fixes #6828.

Issue #6828, opened 2018-06-23
Backward compatibility for #6816


masterleinad:
#6816 is not backward-compatible as the necessary changes in the tests and the [recently failing HDF5 test](https://cdash.kyomu.43-1.org/testDetails.php?test=17947376&build=2835) show. 
Examples are the `DataOutBase::write_*` functions.
We should keep the old signatures, deprecate them and forward to the new interface.

davydden:
I know that those functions are in public namespace, but I would not bother deprecating it and just update the remaining tests. The rationale is that they are, IMO, in nature implementations details that users are unlikely to use. They are probably ‘public’ only to make testing possible.

davydden:
Ps. High-level write_vtu and alike are unchanged

masterleinad:
I am much happier to follow the policy of not breaking backward-compatibility in the public interface if avoidable. Let me write a PR.

PR #6818, opened 2018-06-21
Reflow some code samples.

drwells:
The content of @code/@endcode blocks must be less than 80 characters per line to prevent clang-format from rewrapping it in strange ways.

I went ahead and fixed three files: things look better now.

Part of #6814.

PR #9481, opened 2020-02-06
Revert introduction of deprecated functions.

### LINKED ISSUE #6832: Post 9.1: Revert #6828
#6830 is a lengthy patch that introduces a bunch of backward compatibility fixes. Post-9.1, we should probably remove these functions again, and the easiest way to do this is to just revert #6830.

bangerth:
In #6830, @masterleinad introduced a whole functions that retained
backward compatibility for functions that @davydden had changed in #6818.
This was before the 9.1 release. It is time to remove these backward
compatibility functions again.

Relates to #6828. Fixes #6832.

Passes the test suite for me.
/rebuild


bangerth:
Good question. I think I've left other changelog entries about removed things in `minor/` as well.

How about we merge this one as is, and I'll go rummage around `minor/` in a separate PR to figure out what needs to be moved?

PR #9482, opened 2020-02-06
Fix subscript in step-12 intro

rumajo:
I believe the subscript on \phi^upwind in the weak form should be j, rather than i. This matches the code and the corresponding comments.

PR #9483, opened 2020-02-06
Ignore merge commits in author check

masterleinad:
This should work around issues like in https://github.com/dealii/dealii/pull/9482 that we see lately where the author information is set correctly but the merge commit GitHub is creating contains invalid author information.
Of course, this implies that all other merge commits are ignored as well. Alternatively, we could always ignore the last commit since that is a merge commit for pull requests.


tjhei:
Thanks, Daniel. I think this makes sense considering how much pain we had with this so far.

PR #9487, opened 2020-02-06
Fix rhs typo in introduction of step-40

lpsaavedra:


tjhei:
You are correct of course. Thank you for fixing this! Documentation only, so I will merge.

bangerth:
Outstanding, we love to see new faces provide patches! I hope there are many more to come, @lpsaavedra !

PR #9490, opened 2020-02-07
Add header quadrature_generator.h

simonsticko:
I have an implementation of an algorithm that can generate immersed quadrature rules for the different regions of a cell intersected by the zero contour of a level set function. I would like to add a QuadratureGenerator class for this, but it's going to be too much code for one pull request. To split this into smaller parts, I would like to first add a header non_matching/quadrature_generator.h and in several pull requests add some needed tools to a namespace internal::QuadratureGeneratorImplementation. In this pull request I would like to add a Hypercube class to this namespace.

### DIFF
--- a/include/deal.II/non_matching/quadrature_generator.h
+++ b/include/deal.II/non_matching/quadrature_generator.h
@@ -1,0 +1,231 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 1998 - 2018 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_non_matching_quadrature_generator
+#define dealii_non_matching_quadrature_generator
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/base/point.h>
+
+#include <type_traits>
+
+DEAL_II_NAMESPACE_OPEN
+
+namespace NonMatching
+{
+  namespace internal
+  {
+    namespace QuadratureGeneratorImplementation
+    {
+      /**
+       * Returns the center $(0.5, ..., 0.5)$ of the unit hypercube:
+       * $[0,1]^{dim}$.
+       */
+      template <int dim>
+      Point<dim>
+      unit_hypercube_center();
+
+
+      /**
+       * Describes a dim-dimensional hypercube aligned with the coordinate
+       * axes. That is, a region
+       *
+       * @f[
+       * [x_0-L/2, x_0+L/2] \times ... \times [x_{dim-1}-L/2, x_{dim-1}+L/2]
+       * @f]
+       *
+       * where L is the side length and $(x_0, ..., x_{dim-1})$ are the
+       * coordinates of the center of the hypercube.
+       *
+       * The purpose of this class is to help us keep track of where we are on
+       * the reference cell when generating immersed quadrature rules. There
+       * are two main things we want to be able to do.
+       *
+       * 1. Split a hypercube into its $2^{dim}$ children.
+       * 2. Take a cross section orthogonal to a given direction and get this
+       * as a Hypercube<dim-1>.
+       *
+       * These are needed because the algorithm is allowed to recurse both
+       * over children and over dimensions.
+       *
+       * In 3D, the 2 coordinates of the cross section of Hypercube<3> can be
+       * ordered in 2 ways. That is, if we take the cross section orthogonal
+       * to the y direction we could either order a 3D-coordinate into a
+       * 2D-coordinate as $(x,z)$ or as $(z,x)$. This class uses the second
+       * convention, corresponding to the coordinates being ordered cyclicly
+       * $x \rightarrow y \rightarrow z \rightarrow x \rightarrow ... $
+       * To be precise, if we take a cross section:
+       *
+       * | Orthogonal to | Cross section coordinates ordered as |
+       * |:-------------:|:------------------------------------:|
+       * |      x        |               (y, z)                 |
+       * |      y        |               (z, x)                 |
+       * |      z        |               (x, y)                 |
+       *
+       * This is according to the convention set by the function
+       * <code>coordinate_to_one_dim_higher</code>.
+       */
+      template <int dim>
+      class Hypercube
+      {
+      public:
+        /**
+         * Constructor, takes the center of the hypercube and its side length.
+         * Passing no inarguments creates a hypercube between 0 and 1 in all
+         * directions: $[0, 1]^{dim}$.
+         */
+        Hypercube(
+          const Point<dim> &center_coordinate = unit_hypercube_center<dim>(),
+          const double      length_of_side    = 1);
+
+        /**
+         * Return the lower bound of the hypercube in @p direction.
+         */
+        double
+        lower_bound(const unsigned int direction) const;
+
+        /**
+         * Return the upper bound of the hypercube in @p direction.
+         */
+        double
+        upper_bound(const unsigned int direction) const;
+
+        /**
+         * Return the bounds of the hypercube in @p direction, as a one-dimensional
+         * hypercube.
+         */
+        Hypercube<1>
+        bounds(const unsigned int direction) const;
+
+        /**
+         * Returns the cross section of the hypercube orthogonal to @p direction.
+         * This is a hypercube in one dimension lower.
+         */
+        Hypercube<dim - 1>
+        cross_section(const unsigned int direction) const;
+
+        /**
+         * Returns the point in the center of the hypercube.
+         */
+        const Point<dim> &
+        center() const;
+
+        /**
+         * Returns the length of the side of the hypercube.
+         */
+        double
+        side_length() const;
+
+        /**
+         * Returns the volume of the hypercube.
+         */
+        double
+        volume() const;
+
+        /**
+         * Returns the indexth vertex of the hypercube.
+         */
+        Point<dim>
+        vertex(const unsigned int index) const;
+
+        /**
+         * Returns the indexth child of the hypercube. Child is meant in the
+         * same way as for a cell.
+         */
+        Hypercube<dim>
+        child(const unsigned int index) const;
+
+      private:
+        /**
+         * Coordinate in the center of the hypercube.
+         */
+        const Point<dim> center_coordinate;
+
+        /**
+         * Side length of the hypercube.
+         */
+        const double length_of_side;
+      };
+
+
+      /**
+       * This function defines a convention for how coordinates in dim
+       * dimensions should translate to the coordinates in dim + 1 dimensions,
+       * when one of the coordinates in dim + 1 dimensions is locked to a given
+       * value.
+       *
+       * The convention is the following: Starting from the locked coordinate we
+       * store the lower dimensional coordinates consecutively and wrapping
+       * around when going over the dimension. This relationship can be
+       * described by the following tables:
+       *
+       *                        2D
+       * ------------------------------------------------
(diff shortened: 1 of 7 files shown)


drwells:
I'm sorry that this patch has been lingering for so long without attention from the principal developers. I'll take a look at this in the next few days.

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+{
+  namespace internal
+  {
+    namespace QuadratureGeneratorImplementation
+    {
+      /**
+       * Describes a dim-dimensional hypercube aligned with the coordinate
+       * axes. That is, a region
+       *
+       * [x_i - L/2, x_i + L/2]^dim
```
drwells: Could you put this in a doxygen function block?

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+       *
+       * where L is the side length and (x_0, ..., x_{dim-1}) are the
```
drwells:
```suggestion
       * where L is the side length and $(x_0, ..., x_{dim-1})$ are the
```

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+       * coordinates of the center of the hypercube.
+       *
+       * The purpose of this class is to help us keep track of where we are on
+       * the reference cell when generating immersed quadrature rules. There are
+       * two main things we want to be able to do.
+       *
+       * 1) Split a hypercube into its 2^dim children.
```
drwells:
```suggestion
       * 1) Split a hypercube into its $2^{dim}$ children.
```

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+       *
+       * These are needed because the algorithm is allowed to recurse both over
+       * children and over dimensions.
+       *
+       * In 3D, the 2 coordinates of the cross section of HyperCube<3> can be
+       * ordered in 2 ways. That is, if we take the cross section orthogonal to
+       * the y direction we could either order a 3D-coordinate into a
+       * 2D-coordinate as (x,z) or as (z,x). This class uses the second
+       * convention, corresponding to the coordinates being ordered cyclicly
+       * x -> y -> z -> x -> ... To be precise, if we take a cross section:
```
drwells: There is some weird doxygen stuff going on with the arrows and tables:
Screenshot_2020-03-14 The deal II Library NonMatching internal QuadratureGeneratorImplementation Hypercube dim Class Templa

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+       * two main things we want to be able to do.
+       *
+       * 1) Split a hypercube into its 2^dim children.
+       * 2) Take a cross section orthogonal to a given direction and get this as
+       *    a Hypercube<dim-1>.
+       *
+       * These are needed because the algorithm is allowed to recurse both over
+       * children and over dimensions.
+       *
+       * In 3D, the 2 coordinates of the cross section of HyperCube<3> can be
```
drwells:
```suggestion
       * In 3D, the 2 coordinates of the cross section of Hypercube<3> can be
```

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+      class Hypercube
+      {
+      public:
+        /**
+         * Type defining a cross section of a hypercube. The purpose of the
+         * conditional is to avoid that a 1-dimensional hypercube has a cross
+         * section. In 1D, we let the cross section be void.
+         */
+        using CrossSection =
+          typename std::conditional<(1 < dim), Hypercube<dim - 1>, void>::type;
```
drwells: Its more typical in the library that we have a 0-dim base case instead of doing something like that. Would you mind changing this to do that?

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+        volume() const;
+
+        /**
+         * Returns the index:th vertex of the hypercube.
+         */
+        Point<dim>
+        vertex(const unsigned int index) const;
+
+        /**
+         * Returns the index:th child of the hypercube. Child is meant in the
```
drwells:
```suggestion
         * Returns the indexth child of the hypercube. Child is meant in the
```

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+        const double length_of_side;
+      };
+
+
+      /**
+       * Returns a hypercube between 0 and 1 in all directions: [0, 1]^dim.
+       */
+      template <int dim>
+      Hypercube<dim>
+      create_unit_hypercube();
```
drwells: It would be easier to just make these the default arguments to the constructor - what do you think?

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,211 @@
...
+       * dimension to a coordinate index in dim + 1 dimensions.
+       *
+       * @param locked_coordinate should be in the range [0, dim+1).
+       * @param coordiante_in_dim should be in the range [0, dim).
+       * @return A coordinate index in the range [0, dim+1)
+       */
+      template <int dim>
+      unsigned int
+      coordinate_to_one_dim_higher(const unsigned int locked_coordinate,
+                                   const unsigned int coordiante_in_dim);
```
drwells: We should make this inline in the header since its just one statement and two assertions.

include/deal.II/non_matching/quadrature_generator.h
simonsticko: Sure, but I don't understand what I need to do to terminate the recursion here. :thinking: I looked at 
`Quadratrue<dim>::SubQuadrature`, 
which is essentially equivalent to this case:
`using SubQuadrature = Quadrature<dim - 1>;`
I would expect that `Quadrature<0>` creates a `Quadrature<-1>`, which wouldn't compile since it requires a `Point<-1>`. But clearly this doesn't happen. Why?

bangerth: This compiles:
```
template <int dim>
class X
{
public:
  using sub = X<dim-1>;

  sub get();
};


int main()
{
  X<2> x;
  x.get().get().get();
}
```
This is because calling `get()` instantiates `X<dim-1>` as a class, but unless you create an object of type `sub`, there is no need for the compiler to create instantiate `X<dim-2>`. In other words, the call to `.get().get().get()` does create `X<0>`, but just because that type has a `using/typedef` for `X<-1>` does not imply that `X<-1>` is actually instantiated.

include/deal.II/non_matching/quadrature_generator.h
simonsticko: I made this the default of the constructor. I had to add a function to generate the default value of the `Point<dim>`. I couldn't think of another way of doing it.

drwells:
It's completely cosmetic but would you please replace `CrossSection` with `Hypercube<dim - 1>` everywhere?

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,236 @@
...
+      typename Hypercube<0>::CrossSection
+      Hypercube<0>::cross_section(const unsigned int) const;
+
+      template <>
+      Point<0>
+      Hypercube<0>::vertex(const unsigned int) const;
+
+      template <>
+      Hypercube<0>
+      Hypercube<0>::child(const unsigned int) const;
```
drwells: I don't think these specializations are in the patch any more. If they aren't then would you please remove them?

simonsticko: I think I need these to make it compile. It's analogous as the declarations for `Quadrature<0>` at the end of quadrature.h. Now I have added
```
template class Hypercube<0>;
```
in the .cc-file.

`Hypercube<dim>::cross_section(..)` creates an object of type `Hypercube<dim-1>`  
so I need to specialise this function for `dim==0` to avoid that the compiler creates 
`Hypercube<-1>`. 

Both `Hypercube<dim>::vertex(..)` and `Hypercube<dim>::child(..)`  
uses `GeometryInfo<dim>::unit_cell_vertex` which doesn't exist for `dim==0`.
If I remove it I get
`error: ‘unit_cell_vertex’ is not a member of ‘dealii::GeometryInfo<0>’` 

I guess the alternative is to not explicitely template `Hypercube<0>`. I'm not sure what the best option is. :thinking:

simonsticko: But yes, there is no implementation for these functions in the .cc-file.

drwells: I think the lack of an implementation was what confused me. I'll take a closer look later today.

masterleinad: > But yes, there is no implementation for these functions in the .cc-file.

Only providing a forward declaration without any implementation is asking for problems... I would at least expect some undefined references if you really need it.

I would just implement them with an `ExcNotImplemented` assert.

simonsticko: > I would just implement them with an `ExcNotImplemented` assert.

I can do that for two of the functions, but `Hypercube<0>::cross_section` should return a
`Hypercube<-1>`. If I add an implementation the compiler instantiates `Hypercube<-1>` which fails to compile.


> Only providing a forward declaration without any implementation is asking for problems...

Perhaps, but this workaround was based on how it's done in the Quadrature class. This constructor is declared in quadrature.h but does not have an implementation:
```
template <>
Quadrature<0>::Quadrature(const Quadrature<-1> &, const Quadrature<1> &);
```

drwells: I think this strategy is fine - it works for Quadrature and it works here too.

drwells:
I think the OSX tester got stuck previously. Lets try this again:

/rebuild

drwells:
@simonsticko I think we need to rebase and force push to get the tester unstuck. Could you do this?

simonsticko:
> I think we need to rebase and force push to get the tester unstuck. Could you do this?

Alright, rebased and passed now.

drwells:
Thanks!

include/deal.II/non_matching/quadrature_generator.h
```diff
@@ -0,0 +1,231 @@
...
+       * |:-------------:|:------------------------------------:|
+       * |      x        |               (y, z)                 |
+       * |      y        |               (z, x)                 |
+       * |      z        |               (x, y)                 |
+       *
+       * This is according to the convention set by the function
+       * <code>coordinate_to_one_dim_higher</code>.
+       */
+      template <int dim>
+      class Hypercube
```
bangerth: I should have looked at this earlier, but isn't it the case that this class is in essence the same as `BoundingBox`? Couldn't one have extended `BoundingBox` by whatever additional functionality you need here?

simonsticko: > Couldn't one have extended `BoundingBox` by whatever additional functionality you need here?

Yes, probably.  The member functions would be well defined for `BoundingBox` aswell.

I might be interested in the space-time case where the hypercube is 4-dimensional. I guess that would work with a `BoundingBox` too, but I'm not sure if `BoundingBox::get_neighbor_type` makes sense.

bangerth:
On 4/1/20 12:51 PM, Simon Sticko wrote:
> Yes, probably. The member functions would be well defined for |BoundingBox|
>  aswell.
> 
> I might be interested in the space-time case where the hypercube is 
> 4-dimensional. I guess that would work with a |BoundingBox| too, but I'm
> not sure if |BoundingBox::get_neighbor_type| makes sense.

There is nothing wrong with putting an
   Assert (dim<=3, ExcNotImplemented());
into that function.

We're generally pretty loath to have two classes with essentially the same 
functionality. I wished I had seen this before. Would you be willing to 
follow-up with a patch that removes the new class again and instead uses 
BoundingBox (where necessary adding functions to that class)? If you don't 
have the time right now, please open a github issue so that we don't forget 
before the next release. Thanks in advance!

PR #9492, opened 2020-02-07
Add new constructor to ArrayView

peterrum:
as discussed in #9473.

ping @bangerth


include/deal.II/base/array_view.h
```diff
@@ -131,6 +131,15 @@ class ArrayView
...
                             MemorySpaceType> &view);
 
+  /**
+   * A constructor that automatically creates a view from a value_type object.
+   *
+   * This implicit conversion constructor is particularly useful when calling
+   * a function that takes an ArrayView object as argument, and passing in
+   * a value_type.
+   */
+  ArrayView(value_type &element);
```
tjhei: Would you be opposed to making this explicit? I am worried that things become confusing otherwise.

peterrum: Done.

bangerth: I had thought about that as well, but the other single-argument constructors aren't `explicit` either.  But you're probably right :-)

tests/base/array_view_15.cc
```diff
@@ -0,0 +1,54 @@
...
+
+
+void
+test()
+{
+  int v = 9;
+
+  ArrayView<int> a1(v); // writable view
+
+  Assert(a1.size() == 1, ExcInternalError());
```
rezarastak: I suggest we use `AssertThrow` to ensure correctness even in the release mode.

tamiko: @rezarastak `AssertThrow` is reserved for conditions that you would normally handle with the C++ exception mechanism. Typically, when an error condition depends on user input or configuration (for example throwing `SolverControl::NoConvergence`). Here, however, an invalid array view is an internal error that should be handled by `Assert`. It is our choice that we remove `Assert`s in release mode.

rezarastak: Thank you for explaining @tamiko. I think I understand the overall idea and the difference between `Assert` and `AssertThrow` in the code of the library. I want to know what is the policy of dealii developers regarding assert statements in the tests. My opinion is that a test should assert all reasonable truth statements (to ensure correctness) in both debug and release modes. We are not concerned with the cpu efficiency of the code written for test and we are not worried about the bloated code size of the test binary either. So we should make sure that during compilation of tests, the `Assert` statements are not removed in the release mode. Maybe the build system already does this and I am not aware. Alternatively, we should replace all `Assert` s with `AssertThrow` in the test codes. I already know many dealii tests that use `AssertThrow` frequently.

bangerth: I don't think we have a policy for tests, but I seem to remember that we converted some `Assert` statements in tests to `AssertThrow` for exactly the reason you mention.

I would be curious to see what happens if one converted all of them. Want to give that a try? But let's leave this for a separate patch :-)

rezarastak: Sure! I'll provide a patch regarding the use of `Assert` in tests.

PR #9493, opened 2020-02-07
Convert step-30 from using eps/gnuplot to svg/vtu.

bangerth:
The figures as currently used are actually quite ugly. We can do
substantially better. While there, also clean up the
function that creates output a bit.

I've already put the figures on the website.

In reference to #5982.

### DIFF
--- a/examples/step-30/doc/results.dox
+++ b/examples/step-30/doc/results.dox
@@ -1,125 +1,186 @@
 <h1>Results</h1>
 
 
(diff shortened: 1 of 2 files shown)


bangerth:
Ah, good check -- these file names should all have had "9.2" instead of "9.1" in the names. I've fixed the patch. Thanks!

bangerth:
Ping?

PR #9494, opened 2020-02-07
Matrixfree hold all faces to owned cells fix

### LINKED ISSUE #9485: matrix_free/face_setup_01 segfaults
This happens on two of my platforms.

peterrum:
closes #9485

> The problem could have been avoided if any of us had run ctest -R matrix_free locally before the merge, since the CI tests cannot cover all the library and this is a well-defined context, namely matrix-free functionality.

@kronbichler  Done!

>  Is it documented what the five pointers are and that some may be null?

@bangerth  Done!

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.templates.h
+++ b/include/deal.II/matrix_free/matrix_free.templates.h
@@ -1934,6 +1934,15 @@ MatrixFree<dim, Number, VectorizedArrayType>::initialize_indices(
                                 di.vector_partitioner_face_variants[4],
                                 loop_over_all_faces);
             }
+          else
+            {
+              di.vector_partitioner_face_variants[3].reset(
+                new Utilities::MPI::Partitioner(part.locally_owned_range(),
+                                                part.get_mpi_communicator()));
+              di.vector_partitioner_face_variants[4].reset(
+                new Utilities::MPI::Partitioner(part.locally_owned_range(),
+                                                part.get_mpi_communicator()));
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


kronbichler:
Thanks!

peterrum:
> Thanks!

It was my fault ;)

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -1934,6 +1934,15 @@ MatrixFree<dim, Number, VectorizedArrayType>::initialize_indices(
                                 di.vector_partitioner_face_variants[4],
                                 loop_over_all_faces);
             }
+          else
+            {
+              di.vector_partitioner_face_variants[3].reset(
+                new Utilities::MPI::Partitioner(part.locally_owned_range(),
+                                                part.get_mpi_communicator()));
```
bangerth: We've tried to avoid bare `new` operations recently. It is easier to do
```
di.vector_partitioner_face_variants[3]
  = std::make_shared<Utilities::MPI::Partitioner>(part.locally_owned_range(),
                              ￼                                                part.get_mpi_communicator());
```
and the same for the next line.

peterrum: @bangerth Should I replace also the other `new` in that file?

peterrum: @bangerth I would suggest, we keep it this way and I replace the `new`s in the `matrixfree` folder in a follow-up PR. Is that a deal?

bangerth: Yes, I hadn't realized that there were others in the same file. So yes, let's merge this once the tester is through and I'll let you do the news in a separate PR if you're willing to do that.

PR #9495, opened 2020-02-07
Convert step-14 from using eps/gnuplot to svg/vtu.

bangerth:
The figures as currently used are actually quite ugly. We can do
substantially better.

I've already put the figures on the website.

In reference to #5982.

### DIFF
--- a/examples/step-14/doc/results.dox
+++ b/examples/step-14/doc/results.dox
@@ -47,41 +47,46 @@ Refinement cycle: 8
 @endcode
 
 
-First let's look what the program actually computed. On the fifth
-grid, primal and dual numerical solutions look like this:
+First let's look what the program actually computed. On the seventh
+grid, primal and dual numerical solutions look like this (using a
+color scheme intended to evoke the snow-capped mountains of
+Colorado that the original author of this program now calls
+home):
 <table align="center">
   <tr>
     <td width="50%">
-      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-5.png" alt="">
+      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-7.9.2.png" alt="">
     </td>
     <td width="50%">
-      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-5-dual.png" alt="">
+      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-7-dual.9.2.png" alt="">
     </td>
   </tr>
 </table>
-Obviously, the region at the bottom left is so unimportant for the
+Apparently, the region at the bottom left is so unimportant for the
 point value evaluation at the top right that the grid is left entirely
-unrefined there, even though the solution has singularities there! Due
+unrefined there, even though the solution has singularities at the inner
+corner of that cell! Due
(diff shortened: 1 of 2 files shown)


examples/step-14/doc/results.dox
```diff
@@ -47,56 +47,62 @@ Refinement cycle: 8
...
+home):
 <table align="center">
   <tr>
     <td width="50%">
-      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-5.png" alt="">
+      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-7.9.2.png" alt="">
     </td>
     <td width="50%">
-      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-5-dual.png" alt="">
+      <img src="https://www.dealii.org/images/steps/developer/step-14.point-value.solution-7-dual.9.2.png" alt="">
```
drwells: its not immediately related to this issue, but should we version images in this way? We are currently pretty inconsistent.

bangerth:
On 2/7/20 1:49 PM, David Wells wrote:
> its not immediately related to this issue, but should we version images 
> in this way? We are currently pretty inconsistent.

That's what I've been using for other pictures in this directory. I 
think it's not terribly important, but I'm willing to listen to other 
suggestions (for a follow-up PR if you want).

examples/step-14/doc/results.dox
```diff
@@ -47,56 +47,62 @@ Refinement cycle: 8
...
 refinement criterion involving the dual solution chose to refine them
-differently.
+differently -- because we said that we really only care about a single
+function value somewhere at the top right.
 
 
 
 Looking at the grids that are produced in the course of subsequent
-refinement, here are some of them:
+refinement, here are some of them (in refinement cycles 0, 2, 4 in
```
GrahamBenHarper: I think the original sentence actually doesn't sound right when I read it. What about "Here are some of the grids that are produced in the course of subsequent refinement"

examples/step-14/doc/results.dox
```diff
@@ -227,28 +233,33 @@ evaluation shows this:
...
+    <td><img src="https://www.dealii.org/images/steps/developer/step-14.point-derivative.grid-7.9.2.png" alt="" width="100%"></td>
+    <td><img src="https://www.dealii.org/images/steps/developer/step-14.point-derivative.grid-8.9.2.png" alt="" width="100%"></td>
+    <td><img src="https://www.dealii.org/images/steps/developer/step-14.point-derivative.grid-9.9.2.png" alt="" width="100%"></td>
   </tr>
 </table>
 
 Note the asymmetry of the grids compared with those we obtained for
-the point evaluation, which is due to the directionality of the
-x-derivative for which we tailored the refinement criterion.
+the point evaluation. This is due to face that the domain and the primal
```
GrahamBenHarper: Should this be "due to the fact"?

bangerth:
Thanks, @GrahamBenHarper! I've fixed the places you mentioned.

bangerth:
Ping?

bangerth:
Oh, so much better: https://dealii.org/developer/doxygen/deal.II/step_14.html#Results

PR #9499, opened 2020-02-08
Remove most new in folder include/deal.II/matrix_free/

peterrum:
... as requested by @bangerth and promised in #9494.

@kronbichler I guess we cannot eliminate these:
https://github.com/dealii/dealii/blob/80e9b427d67b198dec3e314632ce43cfbabbc7b1/include/deal.II/matrix_free/fe_evaluation.h#L3289-L3299

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -3233,12 +3233,12 @@ inline FEEvaluationBase<dim,
   // Create deep copy of mapped geometry for use in parallel...
   if (other.mapped_geometry.get() != nullptr)
     {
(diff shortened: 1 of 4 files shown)


kronbichler:
> @kronbichler I guess we cannot eliminate [...]

Not easily - it would require turning the data into a shared pointer with different semantics in the two cases, one freeing the memory, the other simply taking an address to somewhere else. Alternatively, introducing another variable that does the memory management through a unique pointer and then let the plain pointer refer to either the unique pointer or the externally given addressed. It could be done, but I would not consider this high priority.

peterrum:
> I would not consider this high priority.

I agree!

bangerth:
Thanks!

PR #9500, opened 2020-02-09
Remove eps step 6

drwells:
Part of #5982.


drwells:
I made some other trivial whitespace fixes while I was looking at this example - see

https://github.com/dealii/dealii/pull/9500/files?w=1

to get rid of them.

PR #9502, opened 2020-02-09
replace Assert with AssertThrow in tests/base

rezarastak:
The first attempt to address [this comment](https://github.com/dealii/dealii/pull/9492#discussion_r376660548).

I only modified the tests in `tests/base/` except `tests/base/assertion_macros.cc`. I added a paragraph to the documentation for developers specifying that we prefer `AssertThrow` to `Assert` in the test codes. Since the Jenkins CI doesn't check the tests in the release build, I ran the test suite on my machine and the tests passed as expected. I will apply this to the rest of the test suite soon.

### DIFF
--- a/doc/developers/testsuite.html
+++ b/doc/developers/testsuite.html
@@ -658,6 +658,15 @@ int main ()
     </p>
 
     <p>
+    In addition to creating meaningful test output using <code>deallog</code>,
+    you can check post-conditions and software guarantees
+    by using the <code>AssertThrow</code> macro within the test code. Note
+    that we make use of the <code>AssertThrow</code> macro instead of
+    <code>Assert</code> because we like to assert post-conditions in both
+    debug and release builds.
+    </p>
+
+    <p>
     There are a number of directories where you can put a new test.
     Extensive tests of individual classes or groups of classes
     have traditionally been into the <code>base/</code>,
--- a/tests/base/array_view_01.cc
+++ b/tests/base/array_view_01.cc
@@ -29,14 +29,14 @@ test()
   ArrayView<int> a(&v[4], 3); // writable view
   a[2] = 42;
 
-  Assert(a[2] == 42, ExcInternalError());
-  Assert(v[6] == 42, ExcInternalError());
+  AssertThrow(a[2] == 42, ExcInternalError());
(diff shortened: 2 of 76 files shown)


rezarastak:
@peterrum That's correct. We won't be able to change those. But changing `Assert` to `AssertThrow` is an improvement that we can make.

tamiko:
A much more pragmatic way that I recommend to pursue would be to override the Assert preprocessor macro in the testsuite for release mode - that way you get the desired behavior (i.e. `Assert` in tests also trigger in release mode) without changing a single line of code in the tests.

tjhei:
> A much more pragmatic way that I recommend to pursue would be to override the Assert preprocessor macro in the testsuite for release mode

I was thinking the same. 

Option 1:
Modify our assert() macro with a
```
#ifdef ASSERT_ALSO_IN_RELEASE_MODE
 #define Assert bla
```
Do we have already have a macro defined for all source when running the test suite? If not, this could be useful to add.

Option 2:
We could simply override the Assert macro in tests/tests.h. This assumes tests.h is included last.

bangerth:
On 2/10/20 7:14 AM, Timo Heister wrote:
> 
> We could simply override the Assert macro in tests/tests.h. This assumes 
> tests.h is included last.

Or first -- in which case you also convert the occurrences of Assert in the 
deal.II header files. You'd then have to make sure that tests/tests.h includes 
base/exceptions.h so that a later inclusion of base/exceptions.h doesn't 
overwrite your definition of Assert any more.

rezarastak:
> We could simply override the Assert macro in tests/tests.h. This assumes tests.h is included last.

If we go with this approach, I think we should have a check in the indentation script that ensures "../tests.h" is included last in all tests. Also, this requirement might be slightly confusing to the new contributors.

masterleinad:
`clang-format` can take care of that.

bangerth:
I'm ok with this, but because it's generated a bit of discussion here and in other PRs, I think it would be good if a second person approves and merges.

masterleinad:
For me, the discussion reads like we rather want to override the Assert preprocessor macro in the testsuite.

masterleinad:
Superseded by #9539.

PR #9503, opened 2020-02-10
Create separate artifacts in GitHub workflows

masterleinad:
Currently, the workflow CI run only creates one artifact which I guess is due to them having the same name. Let's see if choosing different names help.


tjhei:
Let's see if this works.

tjhei:
it does. :-) Thanks!

Issue #9460, opened 2020-02-02
[Step-20] Doxygen linking mechanism is inadequate when linking to different sections having the same heading name?


krishnakumarg1984:
In the table of contents for Step-20, there are two sections with the same heading name

>1. Introduction
    - Linear solvers and preconditioners
2. The commented program
    - Linear solvers and preconditioners

Having read up to the `MixedLaplaceProblem::assemble_system` section in the commented code, I took a break and wanted to resume from the `Linear solvers and preconditioners` in the commented code. However, clicking this link in the TOC makes the page in the browser fly to the corresponding section name in the Introduction section! 

Assuming this is a bug in Doxygen, what would be the workaround in this kind of scenarios to help the reader with the navigation?

masterleinad:
Choosing a slightly different section name?

bangerth:
The script that generates the table of contents is not prepared to deal with sections of the same name. We should just use a different section name -- feel free to propose a patch!

PR #9504, opened 2020-02-10
Fix header names in step-61

GrahamBenHarper:
As pointed out in #9460, doxygen assigns anchor tags based on the header name, meaning that there are actually a handful of tutorials that have similar problems. This fixes just one doxygen linking issue in step-61.

I am also working on upgrading the `checkdoxygen.py` script, but I now realize that I have to cross-reference every `<h*>` tag and `@sect*{` tag between the files `step-**/intro.dox`, `step-**/results.dox`, and `step-**.cc` in order to develop a comprehensive way to tackle the problem across all tutorials.


tjhei:
> This fixes just one doxygen linking issue in step-61.

great! That is a step in the right direction.

bangerth:
You could also force us to use either `@sect` or `<h*` instead of allowing the mixture.

PR #9505, opened 2020-02-10
Upgrade checkdoxygen.py to check for multiple-defined section headers

GrahamBenHarper:
This is related to #9460. `checkdoxygen.py` will now additionally do the following:
1. If it is reading an `intro.dox` file, concatenate the corresponding `step-**.cc` file and the `results.dox` file into one list.
2. Iterate through the list. Every time a `<h` or a `@sect` is found, its contents (with spaces and nonalphanumerics removed, to match the anchors) are added to a dictionary if it is a new title
3. If this header was already found, throw an error with a helpful message to check the file

Running the script on my end yields the following:
```
$ find doc examples include \( -name "*.h" -o -name "*.dox" \) -print | xargs -n 1 contrib/utilities/checkdoxygen.py
Error: repeated header title 'Privatememberfunctions' in file group 'examples/step-28/doc/intro.dox'
Error: repeated header title 'Settingupperiodicityconstraintsondistributedtriangulations' in file group 'examples/step-45/doc/intro.dox'
Error: repeated header title 'ThePointValueEvaluationclass' in file group 'examples/step-14/doc/intro.dox'
Error: repeated header title 'LaplaceProblemLaplaceProblem' in file group 'examples/step-27/doc/intro.dox'
Error: repeated header title 'HelmholtzProblemHelmholtzProblem' in file group 'examples/step-7/doc/intro.dox'
Error: repeated header title 'Streamlinediffusion' in file group 'examples/step-63/doc/intro.dox'
Error: repeated header title 'Linearsolversandpreconditioners' in file group 'examples/step-20/doc/intro.dox'
Error: repeated header title 'TheRightHandSideclass' in file group 'examples/step-62/doc/intro.dox'
Error: repeated header title 'ElasticProblemElasticProblem' in file group 'examples/step-8/doc/intro.dox'
```

I'm working to verify that these are true statements and if those doxygen links are indeed broken, but I thought I'd share nontheless :)

### DIFF
--- a/contrib/utilities/checkdoxygen.py
+++ b/contrib/utilities/checkdoxygen.py
@@ -49,6 +49,45 @@ def check_empty_lines_in_tables(lines):
 
     return
 
+# have more than one header with the same name in a tutorial?
+def check_multiple_defined_headers(lines):
+    headers = []
+    for l in lines:
+        # this may break if a header splits across a line
+        if "<h" in l:
+            # convert thisheader to the doxygen header name equivalent
+            thisheader = l                             # take this line
+            thisheader = thisheader.replace(' ', '')   # remove all whitespace
+            thisheader = thisheader.split('<h')[1]     # remove header tag '<h'
+            if thisheader[0] in ['1', '2', '3', '4', '5']:  # make sure the next character is 1-5
+                thisheader = thisheader[2:]            # remove '*>'
+                if len(thisheader.split('</h'))==2:    # check for close header on the same line
+                    thisheader = thisheader.split('</h')[0]    # remove close header tag
+                    thisheader = ''.join(ch for ch in thisheader if ch.isalnum()) # remove nonalphanumeric
+                    if thisheader in headers:
+                        sys.exit("Error: repeated header title '%s' in file group '%s'"%(thisheader,args[0]))
+                    else:
+                        headers.append(thisheader)
+                else:
+                    sys.exit("Error: mismatched html header flags in file group '%s'"%(args[0]))
+            # do not check else here because it may match with template arguments like '<hsize_t>'
+        
+        # detect @sect[1-5]{ ... }
+        if "@sect" in l:
+            thisheader = l                             # take this line
+            thisheader = thisheader.replace(' ', '')   # remove all whitespace
+            thisheader = thisheader.split('@sect')[1]  # remove '@sect'
+            if thisheader[0] in ['1', '2', '3', '4', '5']:  # make sure the next character is 1-5
+                thisheader = thisheader[2:]          # remove the number
+                thisheader = ''.join(ch for ch in thisheader if ch.isalnum()) # remove nonalphanumeric
+
+                if thisheader in headers:
+                    sys.exit("Error: repeated header title '%s' in file group '%s'"%(thisheader,args[0]))
+                else:
+                    headers.append(thisheader)
+
+    return
+
 
 args = sys.argv
 args.pop(0)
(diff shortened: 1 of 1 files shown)


GrahamBenHarper:
I just double-checked and it's true, those tutorials indeed all have broken section references for various reasons.

The reasons range from the following:
 -  `Problem::~Problem` has the same anchor name as `Problem::Problem`
 -  The same section appears twice in a row for some reason
 -  A section in the commented code matches a section in `intro.dox` or `results.dox`
I can clean them up in an additional PR.

bangerth:
Yes, it would be nice if you had the time to fix these issues. I would suggest we wait with merging this PR until everything is clean, because we can then also hook this script into the CI infrastructure to make sure we don't get ourselves into the same situation again!

GrahamBenHarper:
That sounds great!

It looks like this script gives a couple false positives for trying to count mismatched header tags, so I'm going to remove the part where it tries to match every `<h` with a `</h`. (it's picking up `<hsize_t>` templates in `.cc` files)

contrib/utilities/checkdoxygen.py
```diff
@@ -60,4 +91,26 @@ def check_empty_lines_in_tables(lines):
...
+if 'intro.dox' in args[0]:
+    # get stepname.cc
+    stepname = args[0].split('/doc',1)[0]
+    stepname = stepname + '/' + stepname.split('/',1)[1] + '.cc'
+
+    # get results.dox
+    resultsname = args[0].split('intro.dox',1)[0] + 'results.dox'
+    
+    # check the .cu file for step-64
+    if '64' in stepname:
```
tjhei: this is problematic, as this will break if we ever have other tutorials with cu. How about you unconditionally grab the .cc and the .cu (if they exist).

contrib/utilities/checkdoxygen.py
```diff
@@ -49,6 +49,37 @@ def check_empty_lines_in_tables(lines):
 
     return
 
+# have more than one header with the same name in a tutorial?
+def check_multiple_defined_headers(lines):
+    headers = []
+    for l in lines:
+        # this may break if a header splits across a line
```
tjhei: Can you change this to fail with an error if the line does not contain ``<h?> ... </h?>``?

contrib/utilities/checkdoxygen.py
```diff
@@ -49,6 +49,37 @@ def check_empty_lines_in_tables(lines):
...
+            thisheader = thisheader.split('</h')[0]    # remove close header tag
+            thisheader = ''.join(ch for ch in thisheader if ch.isalnum()) # remove nonalphanumeric
+
+            if thisheader in headers:
+                sys.exit("Error: repeated header title '%s' in file group '%s'"%(thisheader,args[0]))
+            else:
+                headers.append(thisheader)
+        
+        if "@sect" in l:
+            thisheader = l                             # take this line
```
tjhei: please add a comment ``detect @sect[1-4]{...}``

tjhei:
excellent, thanks for making the effort to put this into a script.

contrib/utilities/checkdoxygen.py
GrahamBenHarper: I have a `try` `except` here if it can't find the `.cc`. The other option was to import another package that checks if files exist, but I imagine it just does a `try` `except` under the hood anyway.

contrib/utilities/checkdoxygen.py
GrahamBenHarper: Got it! It checks for a closing header tag on line 72 now.

contrib/utilities/checkdoxygen.py
GrahamBenHarper: Got it!

PR #9507, opened 2020-02-11
Minor edit to step-6's results section.

bangerth:
I was too slow commenting on #9500.

### DIFF
--- a/examples/step-6/doc/results.dox
+++ b/examples/step-6/doc/results.dox
@@ -247,8 +247,10 @@ the Triangulation what to do when a cell is refined: where should the new
 vertices at the edge midpoints and the cell midpoint be located so that the
 child cells better represent the desired geometry than the parent cell.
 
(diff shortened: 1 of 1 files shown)


drwells:
This is a very pedantic complaint so take it for what it is worth. I added a longer blurb in the actual body of the program which explains the line segment process and then refers to step-10 for more information. I would rather add more explanatory text to either step-10 or to the ‘.cc’ file rather than to this section.

bangerth:
I don't feel particularly adamant about the issue. I just thought that the sentence "we have to output more lines on interior line segments as well as boundary line segments so that the interior lines look
curved" just doesn't read very well. It says "lines" four times, and I felt that it is logically challenged as well ;-)

drwells:
Good point: my original sentence needed an upgrade. Pedantry aside, this change improves the example so let’s merge.

PR #9509, opened 2020-02-11
Explicitly call p4est's repartition functions instead of p::d::Tria::repartition().

marcfehling:
While saving/loading a `p::d::Triangulation`, I encountered issues when serializing `active_fe_indices` when the number of processors changed.

Internally, when the number of processors changes, we repartition the triangulation during the call of `load()`. However, when an `hp::DoFHandler` is assigned to that triangulation previously, another transfer of `active_fe_indices` is called via the repartitioning signals. This messes up the order of pack/unpack handles internally.

Thus, I replaced the `repartition()` call in `load()` with the corresponding `p4est` equivalent.

New tests for serialization have been added in addition.

### DIFF
--- a/source/distributed/tria.cc
+++ b/source/distributed/tria.cc
@@ -2826,12 +2826,43 @@ namespace parallel
         &connectivity);
 
       if (numcpus != Utilities::MPI::n_mpi_processes(this->mpi_communicator))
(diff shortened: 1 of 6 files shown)


kronbichler:
@bangerth any last comments on this PR?

PR #9510, opened 2020-02-11
Add more python bindings

agrayver:
Add missing binds for step-49 

Part of #9015

### DIFF
--- a/contrib/python-bindings/include/triangulation_wrapper.h
+++ b/contrib/python-bindings/include/triangulation_wrapper.h
@@ -59,6 +59,14 @@ namespace python
     ~TriangulationWrapper();
 
     /**
+     * Create a triangulation from a list of vertices and a list of indices,
+     * each of the latter being a list of 1<<dim vertex indices.
+     */
+    void
+    create_triangulation(const boost::python::list &vertices,
+                         const boost::python::list &cells_vertices);
+
+    /**
      * Return the number of active cells.
      */
     unsigned int
@@ -136,6 +144,16 @@ namespace python
       PointWrapper &       p,
       boost::python::list &material_id,
       const bool           colorize = false);
+
+    /*! @copydoc GridGenerator::hyper_cube_with_cylindrical_hole
+     */
+    void
+    generate_hyper_cube_with_cylindrical_hole(
+      const double       inner_radius = .25,
+      const double       outer_radius = .5,
+      const double       L            = .5,
+      const unsigned int repetitions  = 1,
+      const bool         colorize     = false);
 
     /**
      * Rectangular domain with rectangular pattern of holes. The domain itself
@@ -288,6 +306,13 @@ namespace python
      */
     void
     shift(boost::python::list &shift_list);
+
+    /**
+     * Scale the entire triangulation by the given factor. To preserve
+     * the orientation of the triangulation, the factor must be positive.
+     */
+    void
+    scale(const double scaling_factor);
 
     /**
      * Given two triangulations, create the triangulation that contains the
--- a/contrib/python-bindings/source/export_triangulation.cc
+++ b/contrib/python-bindings/source/export_triangulation.cc
@@ -77,6 +77,11 @@ namespace python
                                          generate_enclosed_hyper_cube,
                                          0,
                                          4)
+  BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(
+    generate_hyper_cube_with_cylindrical_hole_overloads,
+    generate_hyper_cube_with_cylindrical_hole,
+    0,
+    5)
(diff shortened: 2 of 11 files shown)


contrib/python-bindings/source/triangulation_wrapper.cc
```diff
@@ -35,6 +35,43 @@ namespace python
...
+      std::vector<Point<spacedim>> vertices(n_vertices);
+      for (size_t i = 0; i < n_vertices; ++i)
+        {
+          boost::python::list vertex =
+            boost::python::extract<boost::python::list>(vertices_list[i]);
+          for (int d = 0; d < spacedim; ++d)
+            vertices[i][d] = boost::python::extract<double>(vertex[d]);
+        }
+
+      const size_t               n_cells = boost::python::len(cells_vertices);
```
Rombur: Do you know why there is so many blank spaces? This looks strange.

contrib/python-bindings/tests/triangulation_wrapper.py
```diff
@@ -140,6 +156,14 @@ def test_subdivided_material_hyper_rectangle(self):
...
 
+    def test_hyper_cube_with_cylindrical_hole(self):
+        for dim in self.restricted_dim:
+            triangulation = Triangulation(dim[0])
+            triangulation.generate_hyper_cube_with_cylindrical_hole(inner_radius = .25,
+				outer_radius = .5, L = .5, repetitions = 1, colorize = False)
+            n_cells = triangulation.n_active_cells()
+            self.assertEqual(n_cells, 8)
+
     def test_generate_chees(self):
```
Rombur: Can you fix the typo here `test_generate_cheese`

contrib/python-bindings/include/triangulation_wrapper.h
```diff
@@ -137,6 +145,20 @@ namespace python
       boost::python::list &material_id,
       const bool           colorize = false);
 
+    /**
+     * This function produces a square in the xy-plane with a cylindrical
+     * hole in the middle. The square and the circle are centered at the
+     * origin. In 3d, this geometry is extruded in z direction to the
+     * interval [0,L].
```
bangerth: Is there a way to use `@copydoc` to just copy the documentation of the function I assume you're calling internally?

contrib/python-bindings/source/export_triangulation.cc
```diff
@@ -291,11 +302,24 @@ namespace python
 
 
 
+  const char generate_hyper_cube_with_cylindrical_hole_docstring[] =
+    "This function produces a square in the xy-plane with a cylindrical     \n"
+    "The square and the hole in the middle. circle are centered at the      \n"
```
bangerth: There's a grammar problem here :-)

contrib/python-bindings/source/triangulation_wrapper.cc
agrayver: No idea. `make indent` insists on this...

contrib/python-bindings/include/triangulation_wrapper.h
agrayver: Like that?

agrayver: If it works here, should I replace it everywhere (in a separate PR)? It is rather annoying to copy-paste...

agrayver:
Thanks both, I addressed your comments.

contrib/python-bindings/include/triangulation_wrapper.h
bangerth: Yes, I think that would be great -- this way things don't go out of date.

Have you checked that it actually works?

(I wished we could do the same for the `_docstring` text...

contrib/python-bindings/source/triangulation_wrapper.cc
masterleinad: `clang-format` (with our settings) aligns subsequent variable declarations both at the type name and the variable name.

agrayver:
I have added a couple more small binds here -- with them, I can fully reproduce step-49.

contrib/python-bindings/include/triangulation_wrapper.h
agrayver: > Have you checked that it actually works?

Hm, how do I do that? The documentation for the python bindings is not compiled as far as I can see...

bangerth: Oh, I see. That's unfortunate. Let's just assume that it works, then.

Rombur:
This looks good. The failing tester is because there is no memory left. If there are no more comment, I will merge this tomorrow.

PR #9511, opened 2020-02-11
Indent script: Do not use --diff-filter=d

### LINKED ISSUE #9489: make indent fails with old git version
On one machine with an ancient version of git (1.8.3.1), make indent doesn't do anything. I traced it back to https://github.com/dealii/dealii/blob/master/contrib/utilities/indent_common.sh#L310 which does not return any file. I assume this is because git did not support "--diff-filter=d" (the goal is to exclude deleted files).
Any suggestions on how to work around this? Maybe remove the diff-filter and ignore nonexisting files?

tamiko:
This hopefully fixes some compatibility issues with ancient git versions
not supporting the --diff-filter option.

Closes #9489

### DIFF
--- a/contrib/utilities/indent_common.sh
+++ b/contrib/utilities/indent_common.sh
@@ -307,8 +307,9 @@ process_changed()
   esac
 
   ( git ls-files --others --exclude-standard -- ${1};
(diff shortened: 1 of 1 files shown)


tamiko:
@tjhei Can you give this PR a spin with your ancient git version please?

tjhei:
works, thank you!

Issue #9513, opened 2020-02-11
Linking for debug library fails on power8 arch


GrahamBenHarper:
~~When I attempt to build deal.II with gcc 7.2 on a power8 ppc64le architecture, every class that includes `vectorization.h` gives the following warning:~~
```
/ascldap/users/gbharpe/programming/Deal.II/dealii-source/include/deal.II/base/vectorization.h: In member function ‘dealii::VectorizedArray<double, 2>& dealii::VectorizedArray<double, 2>::operator=(double)’:
/ascldap/users/gbharpe/programming/Deal.II/dealii-source/include/deal.II/base/vectorization.h:3324:26: warning: parameter ‘x’ set but not used [-Wunused-but-set-parameter]
   operator=(const double x)
                          ^
/ascldap/users/gbharpe/programming/Deal.II/dealii-source/include/deal.II/base/vectorization.h: In member function ‘dealii::VectorizedArray<float, 4>& dealii::VectorizedArray<float, 4>::operator=(float)’:
/ascldap/users/gbharpe/programming/Deal.II/dealii-source/include/deal.II/base/vectorization.h:3558:25: warning: parameter ‘x’ set but not used [-Wunused-but-set-parameter]
   operator=(const float x)
                         ^
...
```
Then once it starts linking `lib/libdeal_II.g.so.9.2.0-pre`, the linker fails and yields many messages similar to the following:
```
/ascldap/users/gbharpe/programming/Deal.II/dealii-source/source/numerics/solution_transfer.cc:101:(.text._ZN6dealii16SolutionTransferILi2ENS_13LinearAlgebra11distributed11BlockVectorISt7complexIfEEENS_2hp10DoFHandlerILi2ELi2EEEE27prepare_for_pure_refinementEv[_ZN6dealii16SolutionTransferILi2ENS_13LinearAlgebra11distributed11BlockVectorISt7complexIfEEENS_2hp10DoFHandlerILi2ELi2EEEE27prepare_for_pure_refinementEv]+0x1dc): relocation truncated to fit: R_PPC64_REL24 (stub) against symbol `std::allocator<std::vector<unsigned int, std::allocator<unsigned int> > >::allocator()' defined in .text._ZNSaISt6vectorIjSaIjEEEC2Ev[_ZNSaISt6vectorIjSaIjEEEC5Ev] section in source/numerics/CMakeFiles/obj_numerics_debug.dir/data_out.cc.o
[...]

PR #9515, opened 2020-02-11
Remove some unused variable warnings in vectorization.h

GrahamBenHarper:
As mentioned [here](https://github.com/dealii/dealii/issues/9513#issuecomment-584791532).


kronbichler:
Not that I believe any of the test machines really get there, but let us play safe anyway.

PR #9516, opened 2020-02-11
parallel::CellWeights: Capture values, not references.

marcfehling:
Since passed parameters may only live on a shorter scope, we should definitely capture their values and not their references.

I don't have a particular failing test at hand right now, but it fixes my benchmark. I'll provide a test if you think this is necessary.


marcfehling:
Everything lives in the same scope in our tests... I encountered some nasty segfaults when applying this on my benchmark :-(

PR #9518, opened 2020-02-12
Use value semantics for const parameters with trivial small types

rezarastak:


rezarastak:
> I hope you used a script for that :-)

The mighty `sed` :-)

I think the CI failed because of an unrelated problem (timeout on a multigrid test)

tjhei:
> I think the CI failed because of an unrelated problem (timeout on a multigrid test)

agreed.

PR #9519, opened 2020-02-12
Avoid calling into MPI functions with MinMaxAvg in serial

kronbichler:
This leads to a series of test failures, e.g. here:
https://cdash.43-1.org/viewTest.php?onlydelta&buildid=4309
The failure was introduced by a change in #9473.


kronbichler:
/rebuild

(The tester seems to be stuck.)

kronbichler:
Test failure is unrelated - let us merge this to reduce the test fallout.

PR #9521, opened 2020-02-12
Remove one newline at beginning of timer statistics

kronbichler:
The timer statistics feature added in #9472 previously printed two newlines before the table, which is a bit much in my opinion. Let us reduce this to one newline (leading to an empty line before the table in the usual case) and let the user decide if she wants more.


kronbichler:
Let us see if someone else has an opinion on this - I am happy to remove the other newline as well. (I could see the use for one newline, which is why I kept it in.)

PR #9524, opened 2020-02-12
FEEvaluation: Place exception about no block vector correctly

### LINKED ISSUE #9312: FEEvaluation segfaults for invalid numbers of components
As noted in [the mailing list](https://groups.google.com/forum/#!topic/dealii/L5WpeR5Rv6Y), `FEEvaluation` provides a cryptic segmentation fault if the underlying element is a scalar one, but the template parameters of FEEvaluation indicate vector-valued quantities. This is because we supprt the case that the user gives us an `std::vector` of vectors that are assigned the separate components during evaluation, so we operate on some pointer variables internally. We should add a check early in the function evaluate/integrate functions to catch this case.

kronbichler:
This fixes #9312.

It turns out that we simply had the relevant check too far down in the code, when we had already accessed invalid pointers.


kronbichler:
Test failure is unrelated - let's merge.

PR #9528, opened 2020-02-14
Use types::global_dof_index rather than unsigned int in MGTransfer

### LINKED ISSUE #8937: Hang with 4+ billion DoFs
I am experiencing a deterministic hang for large problem sizes. This is likely related to #8929 but only appears to happen with 64bit indices and >2^32 DoFs. This happens on a branch that contains a bunch of barriers that seem to work around the crashes in #8929.
The offending function in the user code is
```
dealii::LinearAlgebra::distributed::BlockVector<double> u(2);
matrix_free_operator.initialize_dof_vector(u);
u = 0; // <-- this completes, at least on rank 0
constraints.distribute(u); // <- does not finish on rank 0
```

I haven't been able to find the exact place inside deal.II but I realized that the call stack could be something like
```
Partitioner::Partitioner()
Vector<Number, MemorySpaceType>::reinit(
      const IndexSet &locally_owned_indices,
      const IndexSet &ghost_indices,
      const MPI_Comm  communicator)
internal::import_vector_with_ghost_elements()
AffineConstraints<number>::distribute(VectorType &vec) const
```
so this could also be inside ConsensusAlgorithm again.

@peterrum @kronbichler have you done any runs with >4 billion DoFs recently? You do some trickery with 32 bit indices in the Paritioner. Any suggestions on where to look? I tried switching to PEX but that doesn't help. #8936 also doesn't make a difference.

kronbichler:
As discussed in #8937, we used `unsigned int` for some indices that represent global dof indices in the setup of the indices along refinement edges in geometric multigrid. These should be global indices.

This might address #8937 or at least the reduced test case that @tjhei and @tcclevenger constructed, but my job asserting this is still in the queue of the machine. Given the large size (>200 nodes), I also do not see how we could test this feature in the test suite.

### DIFF
--- a/source/multigrid/mg_transfer_internal.cc
+++ b/source/multigrid/mg_transfer_internal.cc
@@ -263,8 +263,8 @@ namespace internal
               const IndexSet &is_local =
                 dof_handler.locally_owned_mg_dofs(level);
 
-              std::vector<unsigned int> level_dof_indices;
-              std::vector<unsigned int> global_dof_indices;
+              std::vector<types::global_dof_index> level_dof_indices;
+              std::vector<types::global_dof_index> global_dof_indices;
               for (const auto &dofpair : send_data_temp)
                 if (dofpair.level == level)
                   {


kronbichler:
I can confirm that this PR does make the reduced test case pass on 6.6 billion dofs with adaptivity and 9600 MPI ranks. I will not set the magic "fix #8937" word in the post above because I will let @tjhei and @tcclevenger verify it independently.

kronbichler:
I have run the Stokes problem one refinement larger, 11.3 billion DoFs, and again pass the test (30 GMRES iterations), solver time is 44 seconds on 18432 ranks:
```
   Number of degrees of freedom: 11391100382 (10932508407+458591975)
DEAL:GMRES::Convergence step 30 value 5.58896e-11
vmult
   FGMRES Solve timings:           35.5764  (30 iterations)
   Workload Imbalance:             4.50048
   VM Peak:                        5641760

+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |        69s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| compute_A_block_diagonals       |         1 |      1.51s |       2.2% |
| correct_stokes_rhs              |         1 |     0.588s |      0.85% |
| evaluate_viscosity              |         1 |      3.27s |       4.7% |
| refine                          |         1 |      4.72s |       6.8% |
| setup                           |         1 |      14.5s |        21% |
| solve                           |         1 |      44.4s |        64% |
+---------------------------------+-----------+------------+------------+
```
This means this PR looks good.

kronbichler:
CI failure is unrelated.

masterleinad:
What about https://github.com/dealii/dealii/blob/7a878ecee8b04a7195515e366620e820d99a9358/source/multigrid/mg_transfer_internal.cc#L687?

peterrum:
@masterleinad This looks like a vector with local indices:

https://github.com/dealii/dealii/blob/7a878ecee8b04a7195515e366620e820d99a9358/source/multigrid/mg_transfer_internal.cc#L1033-L1035

kronbichler:
Indeed, that vector is `unsigned int` intentionally because it describes the local index space.

tcclevenger:
Thanks @kronbichler! We had been searching for this issue for a while now. I'll let @tjhei verify on Frontera, but the output you see is exactly what should be expected.

tjhei:
Thank you @kronbichler, I will test it now but this is certainly a fix worth merging.

bangerth:
That's exactly the kind of situation @tjhei and I were debating this morning before leaving for the workshop :-)

PR #9530, opened 2020-02-14
Avoid correcting higher derivatives if mapping is Cartesian

simonsticko:
I'm using higher derivatives on a perfectly Cartesian mesh. I noticed some bottlenecks in my code, so I did the following test:

On a 10x10x10 mesh, using `FE_Q<3>(3)` elements and `MappingCartesian`, I called `FEFaceValues::reinit` on all faces in the mesh with different update flags. This gave me the following times spend in `FEFaceValues::reinit`:

``` 
UpdateFlags                time/s
update_values              0.25
update_gradients           0.35
update_hessians            3.25
update_3rd_derivatives     71.33
```

Clearly computing 3rd derivatives is very expensive. However, almost all of the time is spent in `FE_Poly::correct_third_derivatives`, which corrects the computation with the derivatives of the Jacobian. When using a `MappingCartesian`, this correction is just 0.

This pull-request changes so that we first check if the mapping is Cartesian before we compute the correction terms for the Hessian and the third derivatives. For the same example as above, this gives the following times:

``` 
UpdateFlags                time/s
update_values              0.24
update_gradients           0.35
update_hessians            1.38
update_3rd_derivatives     5.06
```

### DIFF
--- a/include/deal.II/fe/fe_poly.h
+++ b/include/deal.II/fe/fe_poly.h
@@ -455,6 +455,30 @@ protected:
   };
 
   /**
+   * Correct the shape Hessians by subtracting the terms corresponding to the
+   * Jacobian pushed forward gradient.
+   *
+   * Before the correction, the Hessians would be given by
+   * @f[
+   * D_{ijk} = \frac{d^2\phi_i}{d \hat x_J d \hat x_K} (J_{jJ})^{-1}
+   * (J_{kK})^{-1},
+   * @f]
+   * where $J_{iI}=\frac{d x_i}{d \hat x_I}$. After the correction, the
+   * correct Hessians would be given by
+   * @f[
+   * \frac{d^2 \phi_i}{d x_j d x_k} = D_{ijk} - H_{mjk} \frac{d \phi_i}{d x_m},
+   * @f]
+   * where $H_{ijk}$ is the Jacobian pushed-forward derivative.
+   */
+  void
+  correct_hessians(
+    internal::FEValuesImplementation::FiniteElementRelatedData<dim, spacedim>
+      &output_data,
+    const internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
+      &                mapping_data,
+    const unsigned int n_q_points) const;
+
+  /**
    * Correct the shape third derivatives by subtracting the terms
    * corresponding to the Jacobian pushed forward gradient and second
    * derivative.
@@ -484,6 +508,7 @@ protected:
       &                mapping_data,
     const unsigned int n_q_points,
     const unsigned int dof) const;
+
 
   /**
    * The polynomial space. Its type is given by the template parameter
--- a/include/deal.II/fe/fe_poly.h
+++ b/include/deal.II/fe/fe_poly.h
@@ -506,8 +506,7 @@ protected:
       &output_data,
     const internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
       &                mapping_data,
-    const unsigned int n_q_points,
-    const unsigned int dof) const;
+    const unsigned int n_q_points) const;
 
 
   /**
--- a/include/deal.II/fe/fe_poly.templates.h
+++ b/include/deal.II/fe/fe_poly.templates.h
@@ -27,6 +27,7 @@
 
 #include <deal.II/fe/fe_poly.h>
 #include <deal.II/fe/fe_values.h>
+#include <deal.II/fe/mapping_cartesian.h>
 
 
 DEAL_II_NAMESPACE_OPEN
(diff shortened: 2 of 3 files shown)


include/deal.II/fe/fe_poly.templates.h
```diff
@@ -472,43 +457,75 @@ FE_Poly<PolynomialType, dim, spacedim>::fill_fe_subface_values(
...
+    &output_data,
+  const internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
+    &                           mapping_data,
+  const unsigned int            n_q_points,
+  const Mapping<dim, spacedim> &mapping) const
+{
+  // If we have a Cartesian mapping we don't need to correct the derivatives
+  // because the second and higher derivatives of the mapping are zero.
+  const bool need_to_correct_hessians =
+    nullptr == dynamic_cast<const MappingCartesian<dim> *>(&mapping);
```
kronbichler: I think we typically place the constant on the right for comparison operations, i.e.,
```suggestion
    dynamic_cast<const MappingCartesian<dim> *>(&mapping) == nullptr;
```
However, I think we sometimes simply write
```c++
if (dynamic_cast<const MappingCartesian<dim> *>(&mapping))
  for ...
```

include/deal.II/fe/fe_poly.templates.h
```diff
@@ -472,43 +457,75 @@ FE_Poly<PolynomialType, dim, spacedim>::fill_fe_subface_values(
...
+
+  if (need_to_correct_hessians)
```
bangerth: I think that a more general condition for the need for correction would actually be that `mapping_data.jacobian_pushed_forward_grads` is the zero tensor. Of course, you can shortcut this test if the mapping is Cartesian, as you use, but it might be useful to also use the more general test in a second step. That would also capture the case where someone uses a `MappingQ` (as we almost always do!) on a mesh with Cartesian cells (as we *also* almost always do).

include/deal.II/fe/fe_poly.h
```diff
@@ -454,6 +454,31 @@ class FE_Poly : public FiniteElement<dim, spacedim>
...
+   * Correct the shape Hessians by subtracting the terms corresponding to the
+   * Jacobian pushed forward gradient.
+   *
+   * Before the correction, the Hessians would be given by
+   * @f[
+   * D_{ijk} = \frac{d^2\phi_i}{d \hat x_J d \hat x_K} (J_{jJ})^{-1}
+   * (J_{kK})^{-1},
+   * @f]
+   * where $J_{iI}=\frac{d x_i}{d \hat x_I}$. After the correction, the
+   * correct third derivative would be given by
```
masterleinad:
```suggestion
   * correct Hessians would be given by
```

include/deal.II/fe/fe_poly.templates.h
```diff
@@ -223,6 +224,56 @@ FE_Poly<PolynomialType, dim, spacedim>::requires_update_flags(
...
+  // value, we know that the higher derivatives need to be corrected.
+  for (unsigned int q = 0; q < n_q_points; ++q)
+    for (unsigned int i = 0; i < spacedim; ++i)
+      for (unsigned int j = 0; j < spacedim; ++j)
+        for (unsigned int k = 0; k < spacedim; ++k)
+          {
+            const double jacobian_pushed_forward_grad =
+              mapping_data.jacobian_pushed_forward_grads[q][i][j][k];
+
+            if (std::abs(jacobian_pushed_forward_grad) > 0)
```
simonsticko: Is this what you had in mind @bangerth ?

I guess the `jacobian_pushed_forward_grads` will not be identically zero if they are computed, right?
So I should probably check
`if (std::abs(jacobian_pushed_forward_grad) > tol)` 
here, but I don't really know what a good value of `tol` would be. :thinking:

bangerth: For some mappings, they will be exactly zero -- for example for the `MappingCartesian`. For others that's maybe not the case, but you can think about what these third derivatives represent: derivatives of second derivatives. So you can say that the third derivative 
```
  f'''(x) = lim_{eps->0}  [f''(x+eps)-f''(x)]/eps
```
is essentially zero if
```
  |f'''(x)|  << |f''(x)]| / L
```
where `L` is some sort of length scale -- for example the mesh size. In other words, here you would want to test something along the lines of
```
  if (mapping.jacobian_pushed_forward_grad[q].norm() 
      <=
      1e-12 * mapping.jacobian_pushed_forward[q].norm() / cell->diameter())
    ... term is essentially zero ...
```
A safer bet would probably to to take the max over all `mapping.jacobian_pushed_forward[q]` on the right hand side, just to make sure that you capture the case where by chance the `mapping.jacobian_pushed_forward[q]` at one point is small.

Makes sense?

kronbichler: How is the state of this discussion? In case there is not an immediate solution in terms of the tolerances, I suggest to merge this PR as is (given that it represents already a good improvement for the cases @simonsticko is interested in) and leave the rest to an issue.

bangerth: That would be fine with me.

@simonsticko ?

simonsticko: Well, I'm mainly interested in the case where the mapping in Cartesian. So should I just remove the added check on the values of `jacobian_pushed_forward_grads`?

masterleinad: We would be fine with merging this and maybe improve later if you agree.

simonsticko: I'm okey with merging this. However, if this:
`std::abs(jacobian_pushed_forward_grad) > 0`
is always `true` in practice (due to floating-point-computations) I suspect that these changes might slightly slow down the general case (when MappingCartesian isn't used).

masterleinad: I see. I think just commenting the code above and returning `true` at this point is OK for now.

simonsticko: Okey, I have added a comment and added issue #9682.

PR #9536, opened 2020-02-15
Avoid nonlocal control flow since it's so hard to read.

bangerth:
Seen while trying to understand the implications of #9534.

/rebuild


bangerth:
Best viewed via https://github.com/dealii/dealii/pull/9536/files?w=1

PR #9539, opened 2020-02-16
Redefine Assert as AssertThrow in tests

masterleinad:
Implementing option 2 in https://github.com/dealii/dealii/pull/9502#issuecomment-584141754. Our clang-format options already make sure that `"../tests.h"` is always the last header file to include.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -127,6 +127,8 @@ IncludeCategories:
 # should not be caught here
   - Regex: "<[a-z_]+>"
     Priority: 100000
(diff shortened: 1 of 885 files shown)


drwells:
There are a few other headers in the test suite - could we also convert those while we are making this change?

tjhei:
Can you please leave a comment in tests.h before the define to explain what this does and why we do this?

masterleinad:
So what do we want to do with this? Are there any objections?

drwells:
Unless someone objects I will merge later today.

PR #9543, opened 2020-02-17
Fix use of array concatenation in checkdoxygen.py

GrahamBenHarper:
I incorrectly used `array.append()` in #9505 instead of the += operator, but this actually fixes it. Now `checkdoxygen.py` will also correctly search the `results.dox` files for repeated headers as well.


GrahamBenHarper:
@tjhei this is a minor change but you might want to rebase your changes on it. Sorry for the hassle!

PR #9552, opened 2020-02-19
Update FE_BernardiRaugel to use physical normal vectors

### LINKED ISSUE #8776: Update `FE_BernardiRaugel` to use physical normal vectors and not reference normal vectors
I've been meaning to do this before version 9.2 is released, but I didn't find time yet. We even wrote in the 9.1 paper "this functionality is planned for the next release," so I'm documenting this now to hold myself accountable and ask for feedback on the approach.

Currently `FE_BernardiRaugel` calls the base implementation of `fill_fe_values()`, but it's only correct for uniform rectangular/cuboidal meshes where the cell normals line up with the x,y,z axes.

This is because currently `PolynomialsBernardiRaugel` assigns the bubble function on each face (functions 8,9,10,11 in 2D and functions 24,25,26,27,28,29 in 3D (*counting from 0*)) as scalar edge-based bubble functions times vectors of the form `[1,0]`, `[0,1]`, or `[1,0,0]`, `[0,1,0]`, `[0,0,1]` depending on `dim`. The definition of the finite element uses the physical normal vector instead.

The correct implementation should involve something like overriding `fill_fe_values()` in `FE_BernardiRaugel`, take the normal vector (not always outward because the DoFs would cancel when assembling a finite element problem) from the face `cell->face(f)` and multiply it by component `f/2` (integer division) of the shape function that `PolynomialsBernardiRaugel` returns.

For derivatives, I believe we would still take component `f/2` of the derivative tensor (which is, itself, a tensor) but then do an outer product with the normal vector on face `f`.

This may also need to be done for other function calls like `fill_fe_face_values`, but I haven't thought about the rest of this implementation as carefully.

I last committed to [my personal branch](https://github.com/GrahamBenHarper/dealii/tree/graham_dev_br) on August 8, but all I did was fill it with a bunch of print statements so that I could better understand what the data structures looked like for the class. Any feedback on this is much appreciated!

GrahamBenHarper:
I'm opening the PR now because I want a little bit of feedback before I continue. I have a couple things I want to accomplish and I'd like to ask for feedback:
 1. I'd like to access the physical (not necessarily outward) normal vector information by using `Triangulation<dim, dim>::cell_iterator &cell`. Since it's an iterator, I need to dereference it to get an accessor to access a face and then its normal, correct? What would that look like? My current implementation isn't right (I currently use `Tensor<1,dim> normal = (*cell)->face(f)->normal_vector(k);`).
 2. I will likely need to implement a sign-flip convention similar to `FE_RaviartThomas` or `FE_Nedelec`. The easiest way to do this is implicitly by having each normal oriented with the edge/face's natural normal direction instead of being an outward normal.
 3. Add a test

(I'll also make sure to remove those print statements when I'm done :stuck_out_tongue:)

This closes #8776.

### DIFF
--- a/include/deal.II/fe/fe_bernardi_raugel.h
+++ b/include/deal.II/fe/fe_bernardi_raugel.h
@@ -130,6 +130,24 @@ private:
    */
   void
   initialize_support_points();
+
+  /**
+   * Fill the values Bernardi-Raugel polynomials, but replace the normal
+   * vector from the PolynomialsBernardiRaugel class with the physical
+   * normal vector from @p cell.
+   */
+  virtual void
+  fill_fe_values(
+    const typename Triangulation<dim, dim>::cell_iterator &cell,
+    const CellSimilarity::Similarity                       cell_similarity,
+    const Quadrature<dim> &                                quadrature,
+    const Mapping<dim, dim> &                              mapping,
+    const typename Mapping<dim, dim>::InternalDataBase &   mapping_internal,
+    const dealii::internal::FEValuesImplementation::MappingRelatedData<dim, dim>
+      &                                                       mapping_data,
+    const typename FiniteElement<dim, dim>::InternalDataBase &fe_internal,
+    dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim, dim>
+      &output_data) const override;
 };
 
 DEAL_II_NAMESPACE_CLOSE
--- a/include/deal.II/fe/fe_bernardi_raugel.h
+++ b/include/deal.II/fe/fe_bernardi_raugel.h
@@ -133,8 +133,8 @@ private:
 
   /**
    * Fill the values Bernardi-Raugel polynomials, but replace the normal
-   * vector from the PolynomialsBernardiRaugel class with the physical
-   * normal vector from @p cell.
+   * vector on the reference cell from the PolynomialsBernardiRaugel class
+   * with the normal vector on the physical cell.
    */
   virtual void
   fill_fe_values(
--- a/source/fe/fe_bernardi_raugel.cc
+++ b/source/fe/fe_bernardi_raugel.cc
@@ -206,7 +206,7 @@ FE_BernardiRaugel<dim>::fill_fe_values(
      mapping_data,
      fe_internal,
      output_data);
-  std::cout << "Length of mapping_data.normal_vectors" << normal_vectors.size() << std::endl;
+  std::cout << "Length of mapping_data.normal_vectors" << mapping_data.normal_vectors.size() << std::endl;
 
   // Convert to the correct internal data class for this FE class.
   // Assert(dynamic_cast<const InternalData *>(&fe_internal) != nullptr,
--- a/source/fe/fe_bernardi_raugel.cc
+++ b/source/fe/fe_bernardi_raugel.cc
@@ -206,8 +206,8 @@ FE_BernardiRaugel<dim>::fill_fe_values(
   std::cout << "Length of mapping_data.normal_vectors: " << mapping_data.normal_vectors.size() << std::endl;
 
   // Convert to the correct internal data class for this FE class.
-  Assert(dynamic_cast<const InternalData *>(&fe_internal) != nullptr,
-      ExcInternalError());
+  //Assert(dynamic_cast<const InternalData *>(&fe_internal) != nullptr,
+  //    ExcInternalError());
   const typename FE_PolyTensor<dim>::InternalData &fe_data =
     static_cast<const typename FE_PolyTensor<dim>::InternalData &>(fe_internal);
 
--- a/source/fe/fe_bernardi_raugel.cc
+++ b/source/fe/fe_bernardi_raugel.cc
@@ -206,7 +206,7 @@ FE_BernardiRaugel<dim>::fill_fe_values(
      mapping_data,
      fe_internal,
      output_data);
-  std::cout << "Length of mapping_data.normal_vectors" << mapping_data.normal_vectors.size() << std::endl;
+  std::cout << "Length of mapping_data.normal_vectors: " << mapping_data.normal_vectors.size() << std::endl;
 
   // Convert to the correct internal data class for this FE class.
   // Assert(dynamic_cast<const InternalData *>(&fe_internal) != nullptr,
@@ -232,7 +232,7 @@ FE_BernardiRaugel<dim>::fill_fe_values(
     {
       double psi = 0;
       unsigned int f = i - dim*GeometryInfo<dim>::vertices_per_cell;
-      Tensor<1,dim> normal = (*cell)->face(f)->normal_vector(k);
+      //Tensor<1,dim> normal = (*cell)->face(f)->normal_vector(k);
 
       for (unsigned int d = 0; d < dim; ++d)
         psi += fe_data.shape_values[i][k][d]*fe_data.shape_values[i][k][d];
@@ -241,8 +241,8 @@ FE_BernardiRaugel<dim>::fill_fe_values(
 
       for (unsigned int d = 0; d < dim; ++d)
       {
-        output_data.shape_values(first + d, k) =
-          normal[d]*psi;
+//        output_data.shape_values(first + d, k) =
+//          normal[d]*psi;
       }
 
     }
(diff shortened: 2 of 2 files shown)


GrahamBenHarper:
I'll wait to indent until I'm ready for to run some build tests :)

GrahamBenHarper:
@bangerth Ping?

How do I extract the cell's normal vectors if I have a `Triangulation<dim, dim>::cell_iterator &cell`?

source/fe/fe_bernardi_raugel.cc
```diff
@@ -43,7 +43,7 @@ FE_BernardiRaugel<dim>::FE_BernardiRaugel(const unsigned int p)
       FiniteElementData<dim>(get_dpo_vector(),
                              dim,
                              2,
-                             FiniteElementData<dim>::Hdiv),
+                             FiniteElementData<dim>::H1),
```
bangerth: Is this correct? Is the element in fact continuous?

include/deal.II/fe/fe_bernardi_raugel.h
```diff
@@ -131,6 +131,24 @@ class FE_BernardiRaugel : public FE_PolyTensor<dim>
    */
   void
   initialize_support_points();
+
+  /**
+   * Fill the values Bernardi-Raugel polynomials, but replace the normal
+   * vector from the PolynomialsBernardiRaugel class with the physical
+   * normal vector from @p cell.
```
bangerth: Can you explain why that is correct? I think what you want to say instead of the first sentence is that this function calls the one in the base class but then overwrites the normal vectors. So the comment should explain why the normals filled by the base class aren't correct.

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,6 +177,80 @@ FE_BernardiRaugel<dim>::initialize_support_points()
...
+  const Mapping<dim, dim> &                              mapping,
+  const typename Mapping<dim, dim>::InternalDataBase &   mapping_internal,
+  const dealii::internal::FEValuesImplementation::MappingRelatedData<dim, dim>
+    &                                                       mapping_data,
+  const typename FiniteElement<dim, dim>::InternalDataBase &fe_internal,
+  dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim, dim>
+    &output_data) const
+{
+  std::cout << "fill_fe_values called!" << std::endl;
+  std::cout << "Calling FE_PolyTensor<dim, dim>::fill_fe_values" << std::endl;
```
bangerth: This will have to go in the final version :-)

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,6 +177,80 @@ FE_BernardiRaugel<dim>::initialize_support_points()
...
+     mapping,
+     mapping_internal,
+     mapping_data,
+     fe_internal,
+     output_data);
+  std::cout << "Length of mapping_data.normal_vectors: " << mapping_data.normal_vectors.size() << std::endl;
+
+  // Convert to the correct internal data class for this FE class.
+  // Assert(dynamic_cast<const InternalData *>(&fe_internal) != nullptr,
+  //    ExcInternalError());
```
bangerth: you can put that assertion in again.

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,6 +177,80 @@ FE_BernardiRaugel<dim>::initialize_support_points()
...
+  {
+    const unsigned int first =
+            output_data.shape_function_to_row_table[i * this->n_components() +
+                                                    this->get_nonzero_components(i)
+                                                      .first_selected_component()];
+
+    for (unsigned int k = 0; k < n_q_points; ++k)
+    {
+      double psi = 0;
+      unsigned int f = i - dim*GeometryInfo<dim>::vertices_per_cell;
```
bangerth: make this `const`

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,6 +177,80 @@ FE_BernardiRaugel<dim>::initialize_support_points()
...
+  for (unsigned int i = dim*GeometryInfo<dim>::vertices_per_cell; i < this->dofs_per_cell; ++i)
+  {
+    const unsigned int first =
+            output_data.shape_function_to_row_table[i * this->n_components() +
+                                                    this->get_nonzero_components(i)
+                                                      .first_selected_component()];
+
+    for (unsigned int k = 0; k < n_q_points; ++k)
+    {
+      double psi = 0;
```
bangerth: move right above the loop where you fill it

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,6 +177,80 @@ FE_BernardiRaugel<dim>::initialize_support_points()
...
+      unsigned int f = i - dim*GeometryInfo<dim>::vertices_per_cell;
+      //Tensor<1,dim> normal = (*cell)->face(f)->normal_vector(k);
+
+      for (unsigned int d = 0; d < dim; ++d)
+        psi += fe_data.shape_values[i][k][d]*fe_data.shape_values[i][k][d];
```
bangerth: It's possible that you can just write this whole loop as
```
  const double psi = fe_data.shape_values[i][k] * fe_data.shape_values[i][k];
```

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,6 +177,80 @@ FE_BernardiRaugel<dim>::initialize_support_points()
...
+
+      psi = sqrt(psi);
```
bangerth:
```suggestion
      psi = std::sqrt(psi);
```
If the above works, then you can also move the `std::sqrt` around the dot product

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,6 +177,80 @@ FE_BernardiRaugel<dim>::initialize_support_points()
...
+
+      for (unsigned int d = 0; d < dim; ++d)
+      {
+//        output_data.shape_values(first + d, k) =
+//          normal[d]*psi;
```
bangerth: Nothing happening here so far :-)

bangerth: Nothing happening here so far :-)

GrahamBenHarper:
@bangerth good points. I'm back to working on addressing some of them now.

On my own, I've used the mixture of `mapping_none` and `mapping_piola`, which works for rectangular cells without a complicated mapping thanks to #8352.

If the cell deviates away from a rectangle, there are a host of issues with that approach. The Piola transform causes the vector field to not point in the normal direction (see image of the distorted normal from the Piola transform below). The Piola transform additionally applies a strange scaling factor of something like h^(1-d) to the vector field. Aside from causing shape functions to take unpleasant values, it could cause issues with DOFs on adjacent cells agreeing.

Screenshot from 2020-06-17 16-19-29

Ideally, I would use the exact normal vector that comes from the mapping, but this information has not been filled in the mapping yet, as my code was showing `mapping_data.normal_vectors` is empty. I know now the right approach is to try to coax the normals from the mapping, but I'm not sure how egregious it would be to try to force the mapping to compute the normals inside this code.

source/fe/fe_bernardi_raugel.cc
GrahamBenHarper: I tried putting it back in, but it threw an error
```
/home/graham/Programming/cpp/Deal.II/dealii-source/source/fe/fe_bernardi_raugel.cc:209:29: error: ISO C++ forbids declaration of ‘type name’ with no type [-fpermissive]
  209 |   Assert(dynamic_cast<const InternalData *>(&fe_internal) != nullptr,
```
I think it already checks this when I call `FE_PolyTensor<dim, dim>::fill_fe_values`, so I think it should be safe to remove.

source/fe/fe_bernardi_raugel.cc
GrahamBenHarper: Perfect! I like that.

source/fe/fe_bernardi_raugel.cc
GrahamBenHarper: Depending on how the DOFs on the edges are handled, yes it could be continuously differentiable on the entire mesh.

include/deal.II/fe/fe_bernardi_raugel.h
GrahamBenHarper: Got it. I added a little more detail in there.

GrahamBenHarper:
I don't exactly see the information I'm looking for in the [interplay documentation](https://www.dealii.org/developer/doxygen/deal.II/group__UpdateFlags.html). Is there a way to force an `update_normal_vectors` flag somewhere inside of that routine perhaps?

tamiko:
Let's close this pull request for the time being and track progress on the corresponding issue.

PR #9556, opened 2020-02-20
Add new vectorized_load_and_transpose and vectorized_transpose_and_store

peterrum:
... taking `std::array<Number*, width>`.

I still have to add the specialized versions...

What is really annoying that I cannot write:
```cpp
vectorized_transpose_and_store(true, n_numbers, arr, other_and_offset);
```
but have to write 
```cpp
vectorized_transpose_and_store<Number, width>(true, n_numbers, arr, other_and_offset);
```
The reason is that `VectorizedArray` takes a size of type `int` and `std::array` of `std::size_t` ...

ping @kronbichler


peterrum:
Was it a good decision that `width` is of type `int`? Would it not be better to be of type `std::size_t`?

kronbichler:
> Was it a good decision that `width` is of type `int`? Would it not be better to be of type `std::size_t`?

We have not released a version with this yet, so in principle it should be possible to change without too much hassle. Do you want to give it a try?

peterrum:
What seem to be swallowed by the compiler:
```cpp
template <typename Number, int width, std::size_t width_>
inline DEAL_II_ALWAYS_INLINE void
vectorized_load_and_transpose(const unsigned int                 n_entries,
                              const std::array<Number *, width_> &in,
                              VectorizedArray<Number, width> *   out);
```

I would like to first finish this PR. We should have a look at the width type in a follow-up PR. In particular, to be compatible with future C++ SIMD features.

kronbichler:
And then you can put a `static_assert` to check that both are indeed the same.

peterrum:
> And then you can put a static_assert to check that both are indeed the same.

Yes that would be the idea. But, the test does not like this... But the specialization for the different SIMD types need that additional template argument...

Could you have a look at the code at this stage. I have tackled `vectorized_load_and_transpose` and `gather`. The remaining 50% is still missing...

peterrum:
BTW: I think the loop over lanes here is not necessary:

https://github.com/dealii/dealii/blob/3ff8e288d2e34b238a2d14a2231baaa0ad3d092d/include/deal.II/base/vectorization.h#L1588-L1590

include/deal.II/base/vectorization.h
```diff
@@ -1592,6 +1698,68 @@ vectorized_load_and_transpose(const unsigned int          n_entries,
...
+  // bit index. This reduces the code size and register pressure because we do
+  // shuffles on 4 numbers rather than 16.
+  const unsigned int n_chunks = n_entries / 4;
+
+  // To avoid warnings about uninitialized variables, need to initialize one
+  // variable to a pre-exisiting value in out, which will never get used in
+  // the end. Keep the initialization outside the loop because of a bug in
+  // gcc-9 which generates a "vmovapd" instruction instead of "vmovupd" in
+  // case t3 is initialized to zero (inside/outside of loop), see
+  // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90991
```
kronbichler: I would remove these comments here and refer to the other code. I would instead add a comment (maybe in the central templated variant) why we have two different functions and why we cannot simply let one function call the other - we either have to increment only the base pointer in case of a single offset, but we need to increment all pointers like in the new case. And we want the compiler to be able to perform one or the other. To be honest, sometimes the compiler transforms the case with a single base with n offsets into n separate pointers that are all incremented (especially when we only do 4 updates), but for wider SIMD the increment of the base is better.

include/deal.II/base/vectorization.h
```diff
@@ -1592,6 +1698,68 @@ vectorized_load_and_transpose(const unsigned int          n_entries,
...
+  // Similar to the double case, we perform the work on smaller entities. In
+  // this case, we start from 128 bit arrays and insert them into a full 512
+  // bit index. This reduces the code size and register pressure because we do
+  // shuffles on 4 numbers rather than 16.
+  const unsigned int n_chunks = n_entries / 4;
+
+  // To avoid warnings about uninitialized variables, need to initialize one
+  // variable to a pre-exisiting value in out, which will never get used in
+  // the end. Keep the initialization outside the loop because of a bug in
+  // gcc-9 which generates a "vmovapd" instruction instead of "vmovupd" in
```
kronbichler: And given that I read this one here, maybe limit the statement to "gcc-9.1", because this was fixed in gcc very quickly after my bug report and is part of gcc-9.2 that is much more common nowadays.

kronbichler:
This approach looks good to me, please go ahead.

kronbichler:
> > And then you can put a static_assert to check that both are indeed the same.
>
> Yes that would be the idea. But, the test does not like this... But the specialization for the different SIMD types need that additional template argument...

I do not think it is a particularly big problem to have the second argument of `width; this would call for a short comment why they are needed at the place this interface is defined, but otherwise we can live with them. (Apart from maybe change the type in the future.)

peterrum:
@kronbichler  ready!

include/deal.II/base/vectorization.h
```diff
@@ -716,6 +724,33 @@ vectorized_load_and_transpose(const unsigned int              n_entries,
 }
 
 
+/**
+ * The same as above with the difference that an array of width pointers are
```
```suggestion
 * The same as above with the difference that an array of pointers are
```
I think `width` is clear from the argument.

include/deal.II/base/vectorization.h
```diff
@@ -716,6 +724,33 @@ vectorized_load_and_transpose(const unsigned int              n_entries,
...
+ * passed in as input argument @p in.
+ *
+ * In analogy to the function above, one can consider that
+ * `in+offset[v]' is precomputed and passed as input argument.
```
```suggestion
 * `in+offset[v]` is precomputed and passed as input argument.
```

include/deal.II/base/vectorization.h
```diff
@@ -716,6 +724,33 @@ vectorized_load_and_transpose(const unsigned int              n_entries,
...
+ *
+ * However, this function can also be used if some function returns width number
```
```suggestion
 * However, this function can also be used if some function returns an array of
```

include/deal.II/base/vectorization.h
```diff
@@ -716,6 +724,33 @@ vectorized_load_and_transpose(const unsigned int              n_entries,
...
+ * of pointers and no assumption can be made that they belong to the same array,
+ * i.e., there is a common start pointer.
```
```suggestion
 * i.e., they can have their origin in different memory allocations.
```

include/deal.II/base/vectorization.h
```diff
@@ -778,6 +813,41 @@ vectorized_transpose_and_store(const bool                            add_into,
...
+/**
+ * The same as above with the difference that an array of width pointers are
+ * passed in as input argument @p out.
+ *
+ * In analogy to the function above, one can consider that
+ * `out+offset[v]' is precomputed and passed as input argument.
+ *
+ * However, this function can also be used if some function returns width number
+ * of pointers and no assumption can be made that they belong to the same array,
+ * i.e., there is a common start pointer.
```
kronbichler: Make similar adjustments as above.

include/deal.II/base/vectorization.h
```diff
@@ -778,6 +813,41 @@ vectorized_transpose_and_store(const bool                            add_into,
...
+      for (unsigned int v = 0;
+           v < VectorizedArray<Number, width>::n_array_elements;
+           ++v)
+        out[v][i] += in[i][v];
+  else
+    for (unsigned int i = 0; i < n_entries; ++i)
+      for (unsigned int v = 0;
+           v < VectorizedArray<Number, width>::n_array_elements;
+           ++v)
+        out[v][+i] = in[i][v];
```
```suggestion
        out[v][i] = in[i][v];
```

include/deal.II/base/vectorization.h
```diff
@@ -2533,6 +3060,90 @@ vectorized_transpose_and_store(const bool                       add_into,
...
+      __m128 res1 = _mm256_extractf128_ps(u1, 0);
+      __m128 res5 = _mm256_extractf128_ps(u1, 1);
+      __m128 res2 = _mm256_extractf128_ps(u2, 0);
+      __m128 res6 = _mm256_extractf128_ps(u2, 1);
+      __m128 res3 = _mm256_extractf128_ps(u3, 0);
+      __m128 res7 = _mm256_extractf128_ps(u3, 1);
+
+      // Cannot use the same store instructions in both paths of the 'if'
+      // because the compiler cannot know that there is no aliasing between
+      // pointers
```
kronbichler: Is there a reason why you removed this comment above but added it here?

include/deal.II/base/vectorization.h
```diff
@@ -1592,6 +1796,59 @@ vectorized_load_and_transpose(const unsigned int          n_entries,
...
+      __m512 v3 = _mm512_shuffle_ps(t2, t3, 0xee);
+
+      out[4 * i + 0].data = _mm512_shuffle_ps(v0, v2, 0x88);
+      out[4 * i + 1].data = _mm512_shuffle_ps(v0, v2, 0xdd);
+      out[4 * i + 2].data = _mm512_shuffle_ps(v1, v3, 0x88);
+      out[4 * i + 3].data = _mm512_shuffle_ps(v1, v3, 0xdd);
+    }
+
+  for (unsigned int i = 4 * n_chunks; i < n_entries; ++i)
+    for (unsigned int v = 0; v < 8; ++v)
```
```suggestion
```
Didn't you want to remove this loop?

include/deal.II/base/vectorization.h
```diff
@@ -1704,6 +1961,117 @@ vectorized_transpose_and_store(const bool                        add_into,
...
+    }
+
+  if (add_into)
+    for (unsigned int i = 4 * n_chunks; i < n_entries; ++i)
+      for (unsigned int v = 0; v < 16; ++v)
+        out[v][i] += in[i][v];
+  else
+    for (unsigned int i = 4 * n_chunks; i < n_entries; ++i)
+      for (unsigned int v = 0; v < 16; ++v)
+        out[v][i] = in[i][v];
```
kronbichler: A question: Do we want to use the `scatter` (and `gather + scatter`) methods of `VectorizedArray` here and in all other places of similar code? It makes both the code and machine code a bit shorter (machine code only on AVX-512 targets).

include/deal.II/base/vectorization.h
peterrum: That was the wrong function...

include/deal.II/base/vectorization.h
peterrum: I have also removed the other one.

include/deal.II/base/vectorization.h
peterrum: Let's merge this first. But, I was also thinking about same. But this is an issue not only related to the new functions but also to the old ones...

peterrum:
@kronbichler Thanks for the review. I have applied your suggestion!

include/deal.II/base/vectorization.h
```diff
@@ -526,6 +527,13 @@ class VectorizedArray
     data = base_ptr[offsets[0]];
   }
 
+  DEAL_II_ALWAYS_INLINE
```
tjhei: missing documentation?

include/deal.II/base/vectorization.h
```diff
@@ -526,6 +527,13 @@ class VectorizedArray
...
+  void
+  gather(const std::array<Number *, width> &base_ptr, const unsigned int i)
```
tjhei: base_ptr is not a pointer, though. Rename? Why ``i`` and not ``offset``?

include/deal.II/base/vectorization.h
```diff
@@ -526,6 +527,13 @@ class VectorizedArray
...
+  {
+    data = base_ptr[i][0];
```
tjhei: why do you not need a loop over ``width`` here?

include/deal.II/base/vectorization.h
```diff
@@ -976,6 +1046,14 @@ class VectorizedArray<double, 8>
     data                = _mm512_i32gather_pd(index, base_ptr, 8);
   }
 
+  DEAL_II_ALWAYS_INLINE
+  void
+  gather(const std::array<double *, 8> &base_ptr, const unsigned int offset)
+  {
+    for (unsigned int v = 0; v < 8; v++)
```
tjhei: why do you need all these specializations of ``gather`` even though their implementation is identical?

include/deal.II/base/vectorization.h
peterrum: You are right. I have forgotten it. Now its there.

include/deal.II/base/vectorization.h
peterrum: I changed the names.

include/deal.II/base/vectorization.h
peterrum: This is the specialization for `static const unsigned int n_array_elements = 1;`.

include/deal.II/base/vectorization.h
peterrum: The problem is that not the function is specialized but the whole class...

kronbichler: I guess the question should rather be: Do we really need gather to be a member function in case we cannot go to machine-specific code (and I guess there will not be this kind of intrinsics on the foreseeable architectures), or should we aim for providing a free-standing function `gather` (maybe within some namespace) that can use a single implementation?

include/deal.II/base/vectorization.h
```diff
@@ -676,6 +677,31 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArrayType
...
+inline DEAL_II_ALWAYS_INLINE void
+gather(VectorizedArray<Number, width> &    out,
+       const std::array<Number *, width_> &ptrs,
+       const unsigned int                  offset)
+{
+  static_assert(width == width_, "Length of input vector do not match!");
+
+  for (unsigned int v = 0; v < width_; v++)
+    out.data[v] = ptrs[v][offset];
+}
```
peterrum: @kronbichler I have moved it into a free function. How do you like it this way?

kronbichler: Nice!

PR #9557, opened 2020-02-21
Fix elastic equation of step-8

dangars:
The equation \partial_j (c_{ijkl} \partial_k u_l) is only valid if c_{ijkl} = c_{ijlk}

The rest of the equations are correct because the material is isotropic


bangerth:
Thanks a lot, @dangars! I think I'll take that as an opportunity to expand this part of the introduction a bit more, but this is a good start!

dangars:
Happy it helped!

PR #9558, opened 2020-02-23
step-47: Add derivation for the simply supported plates to result section.

### LINKED ISSUE #9426: Dealing with clamped boundary conditions in step-47

sophy1029:
Fixes #9426.
The results of this case is similar to the "clamped case".
The only difference is the set of faces we integrate. In this case, the sums take place over interior faces. I have a question of how to deal with values boundary faces.
I suppose " u = g " is the essential boundary and " \Delta u = h " is the natural boundary.
So " \Delta u = h " will be calculated on the RHS. It comes from the 2nd term on the LHS.
In the 2nd term, faces of the mesh are separated to interior faces and boundary faces.
Then the boundary faces are calculated on the RHS, and " \frac{\partial^2 u_h}{\partial^2 \mathbf{n}} = h ". This is the part I'm not sure. Is it correct? Does it relate to anything about the tangential component of gradient of continuous function?

### DIFF
--- a/examples/step-47/doc/intro.dox
+++ b/examples/step-47/doc/intro.dox
@@ -230,7 +230,10 @@ As mentioned, this method relies on the use of $C^0$ Lagrange finite
 elements where the $C^1$ continuity requirement is relaxed and has
 been replaced with interior penalty techniques. To derive this method,
 we consider a $C^0$ shape function $v_h$ which vanishes on
-$\partial\Omega$. Since the higher order derivatives of $v_h$ have two
+$\partial\Omega$. We introduce notation $ \mathbb{F} $ as the set of
+all faces of $\mathbb{T}$, $ \mathbb{F}^b $ as the set of boundary faces,
+and $ \mathbb{F}^i $ as the set of interior faces for use further down below.
+Since the higher order derivatives of $v_h$ have two
 values on each interface $e\in \mathbb{F}$ (shared by the two cells
 $K_{+},K_{-} \in \mathbb{T}$), we cope with this discontinuity by
 defining the following single-valued functions on $e$:
--- a/examples/step-47/doc/results.dox
+++ b/examples/step-47/doc/results.dox
@@ -213,10 +213,10 @@ make sense:
   same spirit as we used for the assembly of the linear system.
 
 
-  <h3> Derivation for the simply supported plates </h3>
+  <h4> Derivation for the simply supported plates </h3>
 
-  Similar to the “clamped” boundary condition addressed in the implementation,
-  we will derive the $C^0$ IP finite element scheme for the simply supported plates:
+  Similar to the "clamped" boundary condition addressed in the implementation,
+  we will derive the $C^0$ IP finite element scheme for simply supported plates:
   @f{align*}{
     \Delta^2 u(\mathbf x) &= f(\mathbf x)
     \qquad \qquad &&\forall \mathbf x \in \Omega,
@@ -225,7 +225,7 @@ make sense:
     \Delta u(\mathbf x) &= h(\mathbf x) \qquad \qquad
     &&\forall \mathbf x \in \partial\Omega.
   @f}
-  We multiply the biharmonic equation by the test function $v_h$ and integrate over $\Omega$ and get:
+  We multiply the biharmonic equation by the test function $v_h$ and integrate over $ K $ and get:
   @f{align*}{
     \int_K v_h (\Delta^2 u_h)
      &= \int_K (D^2 v_h) : (D^2 u_h)
@@ -233,11 +233,12 @@ make sense:
        -\int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}).
   @f}
 
-  Summing up over all cells $K \in  \mathbb{T}$,
+  Summing up over all cells $K \in  \mathbb{T}$,since normal directions of $\Delta u_h$ are pointing at
+  opposite directions on each interior edge shared by two cells and $v_h = 0$ on $\partial \Omega$,
   @f{align*}{
   \sum_{K \in \mathbb{T}} \int_{\partial K} v_h \frac{\partial (\Delta u_h)}{\partial \mathbf{n}} = 0,
   @f}
-  and by the definition of jump over cells,
+  and by the definition of jump over cell interfaces,
   @f{align*}{
   -\sum_{K \in \mathbb{T}} \int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}) = -\sum_{e \in \mathbb{F}} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} (\frac{\partial^2 u_h}{\partial \mathbf{n^2}}).
   @f}
@@ -246,7 +247,7 @@ make sense:
   -\sum_{K \in \mathbb{T}} \int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}) = -\sum_{e \in \mathbb{F}^i} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} (\frac{\partial^2 u_h}{\partial \mathbf{n^2}})
   - \sum_{e \in \partial \Omega} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} h,
   @f}
-  Where $\mathbb{F}^i$ is the set of interior faces.
+  where $\mathbb{F}^i$ is the set of interior faces.
   This leads us to
   @f{align*}{
   \sum_{K \in \mathbb{T}} \int_K (D^2 v_h) : (D^2 u_h) \ dx - \sum_{e \in \mathbb{F}^i} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} (\frac{\partial^2 u_h}{\partial \mathbf{n^2}}) \ ds
@@ -285,7 +286,10 @@ make sense:
   and
   @f{align*}{
   \mathcal{F}(v_h)&:=\sum_{K \in \mathbb{T}}\int_{K} v_h f \ dx
-  -
+  +
   \sum_{e\subset\partial\Omega}
-  \int_e \jump{\frac{\partial v_h}{\partial \mathbf n^2}} h \ ds.
+  \int_e \jump{\frac{\partial v_h}{\partial \mathbf n}} h \ ds.
   @f}
+  The implementation of this boundary case is similar to "clamped" version
+  except for `boundary_worker` is no longer needed for system assembling
+  and the right hand side is changed according to the formulation.
--- a/examples/step-47/doc/results.dox
+++ b/examples/step-47/doc/results.dox
@@ -211,3 +211,81 @@ make sense:
   addition should not be overly difficult using, for example, the
   FEInterfaceValues class combined with MeshWorker::mesh_loop() in the
   same spirit as we used for the assembly of the linear system.
+
+
+  <h3> Derivation for the simply supported plates </h3>
+
+  Similar to the “clamped” boundary condition addressed in the implementation,
+  we will derive the $C^0$ IP finite element scheme for the simply supported plates:
+  @f{align*}{
+    \Delta^2 u(\mathbf x) &= f(\mathbf x)
+    \qquad \qquad &&\forall \mathbf x \in \Omega,
+    u(\mathbf x) &= g(\mathbf x) \qquad \qquad
+    &&\forall \mathbf x \in \partial\Omega, \\
+    \Delta u(\mathbf x) &= h(\mathbf x) \qquad \qquad
+    &&\forall \mathbf x \in \partial\Omega.
+  @f}
+  We multiply the biharmonic equation by the test function $v_h$ and integrate over $\Omega$ and get:
+  @f{align*}{
+    \int_K v_h (\Delta^2 u_h)
+     &= \int_K (D^2 v_h) : (D^2 u_h)
+       + \int_{\partial K} v_h \frac{\partial (\Delta u_h)}{\partial \mathbf{n}}
+       -\int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}).
+  @f}
+
+  Summing up over all cells $K \in  \mathbb{T}$,
+  @f{align*}{
+  \sum_{K \in \mathbb{T}} \int_{\partial K} v_h \frac{\partial (\Delta u_h)}{\partial \mathbf{n}} = 0,
+  @f}
+  and by the definition of jump over cells,
+  @f{align*}{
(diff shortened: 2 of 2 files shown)


examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
   addition should not be overly difficult using, for example, the
   FEInterfaceValues class combined with MeshWorker::mesh_loop() in the
   same spirit as we used for the assembly of the linear system.
+
+
+  <h3> Derivation for the simply supported plates </h3>
```
bangerth: Make that into an `h4` so that it shows up as a sub-section under "Possibilities for extensions".

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
...
+
+  Similar to the “clamped” boundary condition addressed in the implementation,
```
bangerth: Let's try to avoid unicode characters :-)
```suggestion
  Similar to the "clamped" boundary condition addressed in the implementation,
```

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
...
+  we will derive the $C^0$ IP finite element scheme for the simply supported plates:
```
bangerth:
```suggestion
  we will derive the $C^0$ IP finite element scheme for simply supported plates:
```

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
...
+  @f{align*}{
+    \Delta^2 u(\mathbf x) &= f(\mathbf x)
+    \qquad \qquad &&\forall \mathbf x \in \Omega,
+    u(\mathbf x) &= g(\mathbf x) \qquad \qquad
+    &&\forall \mathbf x \in \partial\Omega, \\
+    \Delta u(\mathbf x) &= h(\mathbf x) \qquad \qquad
+    &&\forall \mathbf x \in \partial\Omega.
+  @f}
+  We multiply the biharmonic equation by the test function $v_h$ and integrate over $\Omega$ and get:
```
bangerth: Do you mean "integrate over $K$" here?

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
...
+  @f{align*}{
+    \int_K v_h (\Delta^2 u_h)
+     &= \int_K (D^2 v_h) : (D^2 u_h)
+       + \int_{\partial K} v_h \frac{\partial (\Delta u_h)}{\partial \mathbf{n}}
+       -\int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}).
+  @f}
+
+  Summing up over all cells $K \in  \mathbb{T}$,
+  @f{align*}{
+  \sum_{K \in \mathbb{T}} \int_{\partial K} v_h \frac{\partial (\Delta u_h)}{\partial \mathbf{n}} = 0,
```
bangerth: Is it easy to see that this is so? It's not obvious to me right away, but maybe I'm just dense. In any case, maybe explain this in one sentence.

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
...
+  @f}
+  and by the definition of jump over cells,
```
bangerth:
```suggestion
  and by the definition of jump over cell interfaces,
```

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
...
+  @f{align*}{
+  -\sum_{K \in \mathbb{T}} \int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}) = -\sum_{e \in \mathbb{F}} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} (\frac{\partial^2 u_h}{\partial \mathbf{n^2}}).
+  @f}
+  We separate interior faces and boundary faces of the domain,
+  @f{align*}{
+  -\sum_{K \in \mathbb{T}} \int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}) = -\sum_{e \in \mathbb{F}^i} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} (\frac{\partial^2 u_h}{\partial \mathbf{n^2}})
+  - \sum_{e \in \partial \Omega} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} h,
+  @f}
+  Where $\mathbb{F}^i$ is the set of interior faces.
```
bangerth:
```suggestion
  where $\mathbb{F}^i$ is the set of interior faces.
```

bangerth: Is this the symbol we had previously used?

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,81 @@ make sense:
...
+   \jump{\frac{\partial v_h}{\partial \mathbf n}}
+   \jump{\frac{\partial u_h}{\partial \mathbf n}} \ ds,
+  @f}
+  and
+  @f{align*}{
+  \mathcal{F}(v_h)&:=\sum_{K \in \mathbb{T}}\int_{K} v_h f \ dx
+  -
+  \sum_{e\subset\partial\Omega}
+  \int_e \jump{\frac{\partial v_h}{\partial \mathbf n^2}} h \ ds.
+  @f}
```
bangerth: I think it might be worth stating what is different in this formulation from the "clamped" version considered in the rest of the program.

examples/step-47/doc/results.dox
sophy1029: Is that because on the boundary of domain, v = 0, so no contribution, on the interior faces, normal directions of \Delta u are pointing at opposite directions, so contributions from 2 cells canceled?

bangerth:
@sophy1029 -- any news?

examples/step-47/doc/results.dox
sophy1029: @bangerth I'm not sure whether my understanding explained above is correct. Could you please help me with it? 
I will continue editing other parts during the break.

bangerth: I'll give it another read when you've addressed all of the other comments here. Just let me know when you're at that point!

sophy1029: Ok! Thanks!

examples/step-47/doc/results.dox
sophy1029: I suppose in the introduction section, we has boundary interfaces $\mathbb{F}^b$ but no interior interfaces. So I set interior faces to be $\mathbb{F}^i$.

examples/step-47/doc/results.dox
```diff
@@ -285,7 +286,10 @@ make sense:
   and
   @f{align*}{
   \mathcal{F}(v_h)&:=\sum_{K \in \mathbb{T}}\int_{K} v_h f \ dx
-  -
+  +
   \sum_{e\subset\partial\Omega}
-  \int_e \jump{\frac{\partial v_h}{\partial \mathbf n^2}} h \ ds.
+  \int_e \jump{\frac{\partial v_h}{\partial \mathbf n}} h \ ds.
```
sophy1029: I suppose it's "+" instead of "-".

examples/step-47/doc/results.dox
```diff
@@ -246,7 +247,7 @@ make sense:
   -\sum_{K \in \mathbb{T}} \int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}) = -\sum_{e \in \mathbb{F}^i} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} (\frac{\partial^2 u_h}{\partial \mathbf{n^2}})
   - \sum_{e \in \partial \Omega} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} h,
   @f}
-  Where $\mathbb{F}^i$ is the set of interior faces.
+  where $\mathbb{F}^i$ is the set of interior faces.
```
sophy1029: We use $\mathbb{F}^b$ in the introduction section as boundary interfaces but don't mention interior interfaces. So I set interior faces to be $\mathbb{F}^i$.

examples/step-47/doc/results.dox
bangerth: That's fine, but it should be explained somewhere. (If $F^b$ is not explained yet, would you mind adding that too where it is appropriate?)

sophy1029: Shall we explain both of these two in the introduction section where the first place we use notation $\mathbb{F}$?

bangerth: Yes, I think that would be best. (If you don't use a symbol right away but want to introduce it there, you can always say "We will also denote by X the set of ... for use further down below.")

sophy1029: Ok, I see. Thanks!

examples/step-47/doc/intro.dox
```diff
@@ -230,7 +230,10 @@ As mentioned, this method relies on the use of $C^0$ Lagrange finite
 elements where the $C^1$ continuity requirement is relaxed and has
 been replaced with interior penalty techniques. To derive this method,
 we consider a $C^0$ shape function $v_h$ which vanishes on
-$\partial\Omega$. Since the higher order derivatives of $v_h$ have two
+$\partial\Omega$. We introduce notation $ \mathbb{F} $ as the set of
```
kronbichler, line 233:
```suggestion
$\partial\Omega$. We denote $ \mathbb{F} $ as the set of
```

examples/step-47/doc/intro.dox
```diff
@@ -230,7 +230,10 @@ As mentioned, this method relies on the use of $C^0$ Lagrange finite
...
+all faces of $\mathbb{T}$, $ \mathbb{F}^b $ as the set of boundary faces,
+and $ \mathbb{F}^i $ as the set of interior faces for use further down below.
```
kronbichler, line 235:
```suggestion
and $ \mathbb{F}^i $ as the set of interior faces for use further down.
```

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,85 @@ make sense:
...
+  @f}
+  where $\mathbb{F}^i$ is the set of interior faces.
+  This leads us to
+  @f{align*}{
+  \sum_{K \in \mathbb{T}} \int_K (D^2 v_h) : (D^2 u_h) \ dx - \sum_{e \in \mathbb{F}^i} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} (\frac{\partial^2 u_h}{\partial \mathbf{n^2}}) \ ds
+  = \sum_{K \in \mathbb{T}}\int_{K} v_h f  \ dx + \sum_{e\subset\partial\Omega} \int_{e} \jump{\frac{\partial v_h}{\partial \mathbf{n}} h \ ds.
+  @f}
+
+  In order to symmetrize and stabilize the discrete problem,
+  we add symmetrization and stabilization term.
```
kronbichler, line 258:
```suggestion
  we add symmetrization and stabilization terms.
```

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,85 @@ make sense:
...
+   \jump{\frac{\partial u_h}{\partial \mathbf n}} \ ds,
+  @f}
+  and
+  @f{align*}{
+  \mathcal{F}(v_h)&:=\sum_{K \in \mathbb{T}}\int_{K} v_h f \ dx
+  +
+  \sum_{e\subset\partial\Omega}
+  \int_e \jump{\frac{\partial v_h}{\partial \mathbf n}} h \ ds.
+  @f}
+  The implementation of this boundary case is similar to "clamped" version
```
kronbichler, line 293:
```suggestion
  The implementation of this boundary case is similar to the "clamped" version
```

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,85 @@ make sense:
...
+  except for `boundary_worker` is no longer needed for system assembling
```
kronbichler, line 294:
```suggestion
  except that `boundary_worker` is no longer needed for system assembling
```

examples/step-47/doc/intro.dox
bangerth: Actually, let's go with 
```suggestion
$\partial\Omega$. We denote by $ \mathbb{F} $ the set of
```

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,85 @@ make sense:
...
+  @f}
+  We multiply the biharmonic equation by the test function $v_h$ and integrate over $ K $ and get:
+  @f{align*}{
+    \int_K v_h (\Delta^2 u_h)
+     &= \int_K (D^2 v_h) : (D^2 u_h)
+       + \int_{\partial K} v_h \frac{\partial (\Delta u_h)}{\partial \mathbf{n}}
+       -\int_{\partial K} (\nabla v_h) \cdot (\frac{\partial \nabla u_h}{\partial \mathbf{n}}).
+  @f}
+
+  Summing up over all cells $K \in  \mathbb{T}$,since normal directions of $\Delta u_h$ are pointing at
```
bangerth, line 236:
```suggestion
  Summing up over all cells $K \in  \mathbb{T}$, since normal directions of $\Delta u_h$ are pointing at
```

sophy1029:
@bangerth
I committed those suggestions.
But I got a run failed warning previously:
> I got a message with "Run failed: github-CI - step-47 ([f251aa4](https://github.com/sophy1029/dealii/commit/f251aa42eb2987991bc0ac5265ddc6b7747e4e27))" for this commit.

Should I do something for it?

kronbichler:
This reports of possible errors. In this case, we have
```
/home/travis/build/dealii/dealii/build/doc/doxygen/tutorial/step-47.h:1870: warning: <h4> ended with </h3>
```
Can you take a look?

bangerth:
I think we fixed that already. Can you do
```
  git checkout master
  git pull upstream master
  git checkout step-47
  git rebase master
```
and if that works do
```
  git push -f origin step-47
```
Let us know if that worked!

sophy1029:
It seems like there still problems?

examples/step-47/doc/results.dox
```diff
@@ -211,3 +211,85 @@ make sense:
   addition should not be overly difficult using, for example, the
   FEInterfaceValues class combined with MeshWorker::mesh_loop() in the
   same spirit as we used for the assembly of the linear system.
+
+
+  <h4> Derivation for the simply supported plates </h3>
```
kronbichler, line 216: What the CI is still complaining about is this line:
```suggestion
  <h4> Derivation for the simply supported plates </h4>
```

sophy1029: Thanks!

PR #9559, opened 2020-02-24
Update documentation of SphericalManifold::normal_vector.

bangerth:
I needed to explain the issue to myself again after trying to understand https://github.com/geodynamics/aspect/pull/3377.

/rebuild

### DIFF
--- a/source/grid/manifold_lib.cc
+++ b/source/grid/manifold_lib.cc
@@ -419,35 +419,49 @@ SphericalManifold<dim, spacedim>::normal_vector(
   const typename Triangulation<dim, spacedim>::face_iterator &face,
   const Point<spacedim> &                                     p) const
 {
-  // if the maximum deviation for the distance from the vertices to the center
-  // is less than 1.e-5 of the minimum distance to the first vertex, assume we
-  // can simply return p-center. otherwise, we compute the normal using
-  // get_normal_vector
+  // Let us first test whether we are on a "horizontal" face
+  // (tangential to the sphere).  In this case, the normal vector is
+  // easy to compute since it is proportional to the vector from the
+  // center to the point 'p'.
+  //
+  // We test whether that is the case by checking that the vertices
+  // all have roughly the same distance from the center: If the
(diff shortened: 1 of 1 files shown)


source/grid/manifold_lib.cc
```diff
@@ -419,10 +419,17 @@ SphericalManifold<dim, spacedim>::normal_vector(
   const typename Triangulation<dim, spacedim>::face_iterator &face,
   const Point<spacedim> &                                     p) const
 {
-  // if the maximum deviation for the distance from the vertices to the center
-  // is less than 1.e-5 of the minimum distance to the first vertex, assume we
-  // can simply return p-center. otherwise, we compute the normal using
-  // get_normal_vector
+  // Let us first test whether we are on a "horizontal" or "vertical"
```
gassmoeller: I would rephrase this. There are other option than "horizontal" or "vertical" and we only test for "horizontal" and treat it separately. Maybe:
```suggestion
  // Let us first test whether we are on a "horizontal" face. If so ...
```

source/grid/manifold_lib.cc
```diff
@@ -419,10 +419,17 @@ SphericalManifold<dim, spacedim>::normal_vector(
...
+  // face. In the former case, the normal vector is easy to compute
+  // since it is proportional to the vector from the center to the
+  // point 'p'.
+  //
+  // We test whether that is the case by checking that the vertices
+  // all have roughly the same distance from the center: If the
+  // maximum deviation for the distances from the vertices to the
+  // center is less than 1.e-10 of the distance between vertices (as
```
gassmoeller: Oh also it is 1.e-5. The 1e-10 is comparing the square of the distances.

bangerth: Ah yes, good catch -- I took the opportunity to also rename the variables to reflect the fact that they represent *squared* distances.

bangerth:
Updated. I also updated the second place where essentially the same code path and text appeared.

PR #9561, opened 2020-02-24
Avoid using std::endl over '\n'.

bangerth:
std::endl writes an endline character into the stream, but then follows
that up with a flush(). That's expensive on remote-mounted file systems,
and most unnecessary as well -- we should really just write the endline
character and put a flush() at the very end of the operation.


drwells:
\rebuild

bangerth:
On 2/24/20 3:41 PM, Daniel Arndt wrote:
> There are many more places that could be changed in a similar fashion...

Correct. Few of those do lots of output, though. I went through 
DataOutBase a while ago already to get rid of those std::endl.

PR #9562, opened 2020-02-25
Replace std::endl in include and source folder [WIP]

peterrum:
References https://github.com/dealii/dealii/pull/9561#issuecomment-590609506.

The problem with this change is that the `deallog` in the test does not like it. Example:
```
DEAL:2d::Testing FE_DGQ<2>(1)
DEAL:2d::Number of degrees of freedom: 256
DEAL:2d:cg::Starting value 16.0000
DEAL:2d:cg::Convergence step 5 value 3.60502e-07
DEAL:2d::Testing FE_DGQ<2>(1)
```
vs.
```
DEAL:2d::Testing FE_DGQ<2>(1)
DEAL:2d::Number of degrees of freedom: 256
DEAL:2d::Starting value 16.0000
Convergence step 5 value 3.60502e-07
Testing FE_DGQ<2>(1)
```

What is the best approach in this case?


masterleinad:
We should still use `std::flush` to flush the stream explicitly once we pushed all the content into it.

tjhei:
I would leave all ``deallog`` lines untouched to force a flush.

include/deal.II/base/conditional_ostream.h
```diff
@@ -123,7 +123,7 @@ class ConditionalOStream
    * the surrounding class.
    *
    * Note that compilers want to see this treated differently from the general
-   * template above since functions like @p std::endl are actually overloaded
+   * template above since functions like @p '\n' are actually overloaded
```
bangerth: This is wrong :-)

peterrum:
> I would prefer a more targeted approach where we look at places individually.

@bangerth Maybe let's close this PR?

bangerth:
You're the author ;-) If you don't like your own patch any more, feel free to close the PR :-)

PR #9564, opened 2020-02-25
Make assertions to track state in FEEvaluation debug-only

### LINKED ISSUE #9563: Assert -> AssertThrow in release tests with FEEvaluation
Due to #9539, a large number of matrix-free tests now fail:
https://cdash.43-1.org/viewTest.php?onlydelta&buildid=4538
This is because `FEEvaluation` holds some flags to check what the last operation was, which are only set in debug mode (for efficiency reasons). However, the `Assert` statements do not have the `#ifdef DEBUG` guard. It should be easy to go through the 5-10 places and add the flag. I won't be able to do it immediately so I open this for now before someone else stumbles upon the same problem.

kronbichler:
This moves the assertions that check the state of the various `dof_values_initialized` fields in `FEEvaluation` that only get updated in debug mode (for performance reasons) are placed within an `#ifdef DEBUG` clause. While the `tests.h` file is placed after all includes of deal.II, it comes before the local files in the test folder, which the matrix-free framework makes extensive use of and where the file `fe_evaluation.h` gets included. The changes are not overwhelming, so this is the best thing to do without spending too much time on the scripts that manipulate the header position.

Fixes #9563.

While going through the file, I tried to only put the relevant assertion in the `debug` section, whereas the others (that get now also triggered in release mode) are outside of it.

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -4551,8 +4551,10 @@ FEEvaluationBase<dim, n_components_, Number, is_face, VectorizedArrayType>::
     const unsigned int                                        first_index,
     const std::bitset<VectorizedArrayType::n_array_elements> &mask) const
 {
+#  ifdef DEBUG
   Assert(dof_values_initialized == true,
          internal::ExcAccessToUninitializedField());
+#  endif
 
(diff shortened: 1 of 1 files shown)


tjhei:
It is a bit weird having to wrap Assert inside these ifdef's, but it is what it is. :-)

PR #9565, opened 2020-02-25
Enable refreshing of MatrixFree geometry data for changed mapping

kronbichler:
In certain applications, one wants to impose a grid motion (e.g. through an Euler vector with `MappingFEField`) during the course of a simulation, and still keep alive all other data structures in `MatrixFree`, like shared pointers and the like. This commit adds a new function `MatrixFree::update_mapping(const Mapping<dim>&)` that can perform this step. To make the update stable in `MatrixFreeFunctions::MappingInfo`, I reworked the initialization routines a bit to separate the constant data (like quadrature formula) from the geometry-dependent one and let me separately compute the latter again.

This passes all tests `ctest -R matrix_free` (when combined with #9564).

@nfehn FYI.

### DIFF
--- a/include/deal.II/matrix_free/mapping_info.templates.h
+++ b/include/deal.II/matrix_free/mapping_info.templates.h
@@ -63,8 +63,9 @@ namespace internal
       Assert(structdim + 1 <= spacedim ||
                update_flags_inner_faces == update_default,
              ExcMessage("Volume cells do not allow for setting inner faces"));
-      quadrature = Quadrature<structdim>(quadrature_1d);
-      n_q_points = quadrature.size();
+      this->quadrature_1d = quadrature_1d;
+      quadrature          = Quadrature<structdim>(quadrature_1d);
+      n_q_points          = quadrature.size();
       quadrature_weights.resize(n_q_points);
       for (unsigned int i = 0; i < n_q_points; ++i)
         quadrature_weights[i] = quadrature.weight(i);
@@ -129,6 +130,29 @@ namespace internal
               int spacedim,
               typename Number,
               typename VectorizedArrayType>
+    void
+    MappingInfoStorage<structdim, spacedim, Number, VectorizedArrayType>::
+      clear_data_fields()
+    {
+      data_index_offsets.clear();
+      JxW_values.clear();
+      normal_vectors.clear();
+      for (unsigned int i = 0; i < 2; ++i)
+        {
+          jacobians[i].clear();
+          jacobian_gradients[i].clear();
+          normals_times_jacobians[i].clear();
+        }
+      quadrature_point_offsets.clear();
+      quadrature_points.clear();
+    }
+
+
+
+    template <int structdim,
+              int spacedim,
+              typename Number,
+              typename VectorizedArrayType>
     std::size_t
     MappingInfoStorage<structdim, spacedim, Number, VectorizedArrayType>::
       memory_consumption() const
@@ -156,7 +180,7 @@ namespace internal
     template <typename StreamType>
     void
     MappingInfoStorage<structdim, spacedim, Number, VectorizedArrayType>::
-      print_memory_consumption(StreamType &out, const SizeInfo &task_info) const
+      print_memory_consumption(StreamType &out, const TaskInfo &task_info) const
     {
       // print_memory_statistics involves global communication, so we can
       // disable the check here only if no processor has any such data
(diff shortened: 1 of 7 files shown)


include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -287,19 +311,79 @@ namespace internal
...
+    {
+      AssertDimension(cells.size() / VectorizedArrayType::n_array_elements,
+                      cell_type.size());
+
+      for (auto &data : cell_data)
+        data.clear_data_fields();
+      for (auto &data : face_data)
+        data.clear_data_fields();
+      for (auto &data : face_data_by_cells)
+        data.clear_data_fields();
```
peterrum: I this necessary?

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -684,6 +684,18 @@ class MatrixFree : public Subscriptor
...
+   * Refreshes the geometry data stored in the MappingInfo fields when the
+   * underlying geometry has changed (e.g. by a mapping that can deform
+   * through a change in the spatial configuration like MappingFEField)
+   * whereas the topology of the mesh and unknowns have remained the
+   * same. Compared to reinit(), this operation only has to re-generate the
+   * geometry arrays and can thus be significantly cheaper (depending on the
+   * cost to evaluate the geometry).
+   */
+  void
+  update_mapping(const Mapping<dim> &mapping);
```
peterrum: I am wondering if you could add a version without `mapping` input argument. In this case, the old mapping (which is stored in `MappingInfo`) could be used. 

But maybe, I am just not sure how mapping  is used in this context. Is a new mapping create after each let's say timestep or is an existing mapping updated by some delta xs. In the latter case, it would be odd if the mapping has to handed in again...

kronbichler: It is true that one could update the underlying data stored in `mapping` and let `MappingInfo` figure out what should happen, given that it holds a pointer to `mapping`. This would be the typical case when we have a `MappingFEField` or `MappingQCache` where the underlying vector/points change. However, we are not bound to this and we could in principle pass in another `Mapping` object altogether.

My motivation was to explicitly state where the updated information comes from, so by passing it in again from where the object lives I force the user to think about it when calling `MatrixFree::update_mapping()` - after all, there must have been an update to it recently to force the update. Or do you see a case where I would call this update functionality from a different place to where the `Mapping` lives?

include/deal.II/matrix_free/mapping_info.templates.h
kronbichler: In general yes I would say - think of a deforming mesh case where we start with a Cartesian mesh but due to deformation end up with a general type. This will necessitate different array sizes, so I thought it is best to refresh the arrays. I make some assumptions about NUMA in there (even though the threaded case is not well-tuned) that would be affected.

peterrum: > we start with a Cartesian mesh but due to deformation end up with a general type

That is true. I did not think of this case!

include/deal.II/matrix_free/matrix_free.h
peterrum: > Or do you see a case where I would call this update functionality from a different place to where the Mapping lives?

You are probably right. This function is probably called somewhere (same class but different method) where the user also has called `MatrixFree::reinit()`...

> My motivation was to explicitly state where the updated information comes from, so by passing it in again from where the object lives I force the user to think about it

This is a good point! Normally I create a mapping object directly before `MatrixFree::reinit`. If I would call `MatrixFree::update_mapping()` without argument there might be the case that the mapping object is already out of scope...

peterrum:
You need to rebase this PR.

kronbichler:
I rebased. It turned out that I myself had added a `matrix_free/matrix_vector_faces_27` test in #9602, so the name here needed to be switched to `matrix_vector_faces_28`.

kronbichler:
I have a follow-up work that I would like to base on this one, so I would appreciate if someone could have a look soon.

tests/matrix_free/matrix_vector_faces_28.cc
```diff
@@ -0,0 +1,174 @@
...
+
+#include <deal.II/grid/grid_tools.h>
+
+#include <deal.II/numerics/vector_tools.h>
+
+#include "../tests.h"
+
+#include "create_mesh.h"
+
+std::ofstream logfile("output");
```
masterleinad: I really don't understand why we are continuing to do this.

kronbichler: Copy from another test?

kronbichler: I can run a script to remove the old appearances for that.

masterleinad: Not important for this pull request but I thought we wanted to get rid of these global variables in the long run.

kronbichler: I couldn't agree more, it's just that I concentrated on the actual content I wanted to test and did not pay attention to it. That's why we have code review :smile:

PR #9566, opened 2020-02-25
Deprecate another function in DoFTools.

bangerth:
It's again a function that returns by-argument instead of by-value. Replace it by a function that returns by-value.

Also replace the usage of this function in the examples.

### DIFF
--- a/doc/news/changes/incompatibilities/20200225Bangerth
+++ b/doc/news/changes/incompatibilities/20200225Bangerth
@@ -1,0 +1,6 @@
+Deprecated: The DoFTools::extract_locally_relevant_dofs() function
+returned an IndexSet object through a reference argument. This version
+of the function has now been deprecated and replaced by one that
+returns the IndexSet object as a regular function return value.
+<br>
+(Wolfgang Bangerth, 2020/02/25)
--- a/examples/step-18/step-18.cc
+++ b/examples/step-18/step-18.cc
@@ -816,7 +816,8 @@ namespace Step18
   {
     dof_handler.distribute_dofs(fe);
     locally_owned_dofs = dof_handler.locally_owned_dofs();
-    DoFTools::extract_locally_relevant_dofs(dof_handler, locally_relevant_dofs);
+    locally_relevant_dofs =
+      DoFTools::extract_locally_relevant_dofs(dof_handler);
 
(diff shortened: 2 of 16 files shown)


tjhei:
I like the change to return the value, but I have one related question:
```
    locally_owned_dofs = dof_handler.locally_owned_dofs();
    locally_relevant_dofs = DoFTools::extract_locally_relevant_dofs(dof_handler);
```
It is kind of odd to me, that one is a member of the DoFHandler, while the other one is not. If it was part of DoFHandler, we would be able to cache the result as well?

tamiko:
@tjhei That's a good point. Maybe while we're at it (and already deprecate a function signature). What about relocation the function in the DoFHandler as well? After all, the locally relevant set as we define it is a canonical choice (only depending on the DoFHandler).

bangerth:
Well, that's a bigger change -- we currently have three functions in `DoFTools` for locally owned/active/relevant, but only one in `DoFHandler` (for the first). There are also corresponding "level" functions that currently also all reside in `DoFTools`.

I am agnostic as to what to do. If we decided to merge this change, it will not be difficult later on to make a further change by script if the functions moved into `DoFHandler`.

(I will note that they are currently implemented only once in a generic way, whereas we would have to duplicate them for `DoFHandler` and `hp::DoFHandler` -- though the code duplication would be minimal.)

tjhei:
> If we decided to merge this change, it will not be difficult later on to make a further change by script if the functions moved into `DoFHandler`.

I am not insisting on my suggestion, but if we make the additional change later, I would propose not to merge this PR. This just causes additional PRs in downstream projects.

bangerth:
Well, so what should we do. Anyone else got any opinions?

kronbichler:
I'd also like to have them in the `DoFHandler` - the question is whether we really need to have the locally active ones in the DoFHandler, because I have almost never used them, but at this point I think the symmetry in interfaces make the case.

I do however not feel that the addition to `DoFHandler` must be done within this PR (and support this one as is). As long as we manage to put it there before the next release, I think the headache for external projects will be manageable (especially since we would likely still keep the functions that compute the data in `DoFTools`).

bangerth:
We didn't come to a consensus here, and #10563 also points the way for the future.

PR #9572, opened 2020-02-28
improve documentation of MGTransferMatrixFree::interpolate_to_mg()

nfehn:
Current documentation is misleading as it is not clear which of the vectors of the parameter ``dst`` the comment in the documentation refers to.

### DIFF
--- a/include/deal.II/multigrid/mg_transfer_matrix_free.h
+++ b/include/deal.II/multigrid/mg_transfer_matrix_free.h
@@ -156,8 +156,11 @@ public:
    * Restrict fine-mesh field @p src to each multigrid level in @p dof_handler and
    * store the result in @p dst.
    *
-   * If @p dst is empty or has incorrect locally owned size, it will be
-   * resized to locally relevant vectors on each level.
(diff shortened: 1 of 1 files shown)


nfehn:
In my opinion, the documentation should also highlight that the member function ``build()`` is a prerequisite to call ``interpolate_to_mg()``. If this is desired, this should be done for all member functions and by someone who has written the code. Further, meaningful asserts should be inserted into the code that give a hint that one has forgotten to call ``build()`` beforehand.

I could open an issue.

peterrum:
> Further, meaningful asserts should be inserted into the code that give a hint that one has forgotten to call build() beforehand.

Do you get a message in debug mode?

nfehn:
yes, but it is hard to see where this comes from.

tjhei:
> I could open an issue.

Yes, please. Checking if the internally stored data is valid might not be trivial, because you could modify the mesh/DoFHandler after you call build(). But a simple check if build() has been called beforehand is better than the current situation. :-)

PR #9574, opened 2020-02-28
Add AnonymousConsensusAlgorithmProcess

peterrum:
I got tired of creating a new class that implements `ConsensusAlgorithmProcess` every time even if I only use at one place. This PR adds a generic implementation of `ConsensusAlgorithmProcess` which takes in the constructor a set of lambdas, which are executed during `compute_targets`, `create_request`, ...

This is the closest I can get to an anonymous class (a feature I love in Java).

FYI @kkormann @kronbichler 

@tjhei Once this is merged, I will move the `ConsensusAlgorithm`-related stuff to separate files.

### DIFF
--- a/include/deal.II/base/mpi.h
+++ b/include/deal.II/base/mpi.h
@@ -1069,6 +1069,8 @@ namespace Utilities
                   const std::vector<T2> &recv_buffer);
     };
 
+
+
     /**
      * A base class for algorithms that implement the task of coming up with
      * communication patterns to retrieve data from other processes in a
--- a/include/deal.II/base/mpi_consensus_algorithm.h
+++ b/include/deal.II/base/mpi_consensus_algorithm.h
@@ -1,0 +1,202 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2020 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_mpi_consensus_algorithm_h
+#define dealii_mpi_consensus_algorithm_h
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/base/mpi.h>
+#include <deal.II/base/mpi.templates.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+
+namespace Utilities
+{
+  namespace MPI
+  {
+    /**
+     * This class implements Utilities::MPI::ConsensusAlgorithmProcess,
+     * using user-provided function wrappers.
+     * The advantage of this class is that users do not have to write their
+     * own implementation but can register lambda functions directly.
+     */
+    template <typename T1, typename T2>
+    class AnonymousConsensusAlgorithmProcess
+      : public ConsensusAlgorithmProcess<T1, T2>
+    {
+    public:
+      /**
+       * Register functions that should be called for implementing the interface
+       * of ConsensusAlgorithmProcess.
+       *
+       * @param function_compute_targets called during `compute_targets`.
+       * @param function_create_request called during `create_request`.
+       * @param function_answer_request called during `answer_request`.
+       * @param function_prepare_buffer_for_answer called during
+       *   `prepare_buffer_for_answer`.
+       * @param function_read_answer called during `read_answer`.
+       */
+      AnonymousConsensusAlgorithmProcess(
+        const std::function<std::vector<unsigned int>()>
+          &function_compute_targets,
+        const std::function<void(const unsigned int, std::vector<T1> &)>
+          &function_create_request =
+            [](const unsigned int, std::vector<T1> &) {},
+        const std::function<void(const unsigned int,
+                                 const std::vector<T1> &,
+                                 std::vector<T2> &)> &function_answer_request =
+          [](const unsigned int, const std::vector<T1> &, std::vector<T2> &) {},
+        const std::function<void(const unsigned int, std::vector<T2> &)>
+          &function_prepare_buffer_for_answer =
+            [](const unsigned int, std::vector<T2> &) {},
+        const std::function<void(const unsigned int, const std::vector<T2> &)>
+          &function_read_answer =
+            [](const unsigned int, const std::vector<T2> &) {});
+
+      /**
+       * @copydoc ConsensusAlgorithmProcess::compute_targets()
+       */
+      std::vector<unsigned int>
+      compute_targets() override;
+
+      /**
+       * @copydoc ConsensusAlgorithmProcess::create_request()
+       */
+      void
+      create_request(const unsigned int other_rank,
+                     std::vector<T1> &  send_buffer) override;
+
+      /**
+       * @copydoc ConsensusAlgorithmProcess::answer_request()
+       */
+      void
+      answer_request(const unsigned int     other_rank,
+                     const std::vector<T1> &buffer_recv,
+                     std::vector<T2> &      request_buffer) override;
+
+      /**
+       * @copydoc ConsensusAlgorithmProcess::prepare_buffer_for_answer()
+       */
+      void
+      prepare_buffer_for_answer(const unsigned int other_rank,
+                                std::vector<T2> &  recv_buffer) override;
+
+      /**
+       * @copydoc ConsensusAlgorithmProcess::read_answer()
+       */
+      void
+      read_answer(const unsigned int     other_rank,
+                  const std::vector<T2> &recv_buffer) override;
+
+    private:
+      const std::function<std::vector<unsigned int>()> function_compute_targets;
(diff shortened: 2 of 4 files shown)


include/deal.II/base/mpi.h
```diff
@@ -1069,6 +1069,94 @@ namespace Utilities
...
+       * @param function_compute_targets called during `compute_targets`.
+       * @param function_create_request called during `create_request`.
+       * @param function_answer_request called during `answer_request`.
+       * @param function_prepare_buffer_for_answer called during
+       *   `prepare_buffer_for_answer`.
+       * @param function_read_answer called during `read_answer`.
+       */
+      AnonymousConsensusAlgorithmProcess(
+        const std::function<std::vector<unsigned int>()>
+          function_compute_targets,
```
tjhei: I think we typically pass in std::function as a ``const &``.

include/deal.II/base/mpi.templates.h
```diff
@@ -457,6 +457,83 @@ namespace Utilities
 
 
 
+    template <typename T1, typename T2>
```
tjhei: this means I have to include mpi.templates.h to be able to use this class, right? Is that useful? I would move the implementation into a normal header. (and not mpi.h)

include/deal.II/base/mpi.h
```diff
@@ -1069,6 +1069,94 @@ namespace Utilities
                   const std::vector<T2> &recv_buffer);
     };
 
+
+    /**
+     * Anonymous version of the Utilities::MPI::ConsensusAlgorithmProcess.
+     * Advantage of this class is that users do not have to write their
+     * own implementation but can register directly lambdas.
```
tjhei:
```suggestion
     * own implementation but can register lambda functions directly.
```

peterrum:
@tjhei Thanks for the prompt review! I have moved the implementation into a new file. The rest is going into that file in next PR.

Regarding the const reference. This is not possible since I have to store these functions in the object. I have tried out to save the reference, however, get segmentation fault in that case....

masterleinad:
> Regarding the const reference. This is not possible since I have to store these functions in the object. I have tried out to save the reference, however, get segmentation fault in that case....

What are the error messages you get when passing by reference (and storing by value)?

peterrum:
@masterleinad That would work. But I do not like that because the user would get a misleading impression...

masterleinad:
This is what we do elsewhere as well. In particular, for `std::function` there is `LinearOperator`.

peterrum:
@masterleinad  I have made the change.

peterrum:
Is this `ready to test`?

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,201 @@
...
+       * @param function_answer_request called during `answer_request`.
+       * @param function_prepare_buffer_for_answer called during
+       *   `prepare_buffer_for_answer`.
+       * @param function_read_answer called during `read_answer`.
+       */
+      AnonymousConsensusAlgorithmProcess(
+        const std::function<std::vector<unsigned int>()>
+          &function_compute_targets,
+        const std::function<void(const unsigned int, std::vector<T1> &)>
+          &function_create_request = [](const auto, auto &) {},
```
masterleinad:
```suggestion
          &function_create_request = [](const unsigned int, std::vector<T1> &) {},
```

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,201 @@
...
+        const std::function<
+          void(const unsigned int, const std::vector<T1> &, std::vector<T2> &)>
+          &function_answer_request = [](const auto, const auto &, auto &) {},
```
masterleinad:
```suggestion
          &function_answer_request = [](const unsigned int, const std::vector<T1> &, std::vector<T2> &) {},
```

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,201 @@
...
+        const std::function<void(const unsigned int, std::vector<T2> &)>
+          &function_prepare_buffer_for_answer = [](const auto, auto &) {},
```
masterleinad:
```suggestion
          &function_prepare_buffer_for_answer = [](const unsigned int, std::vector<T2> &) {},
```

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,201 @@
...
+        const std::function<void(const unsigned int, const std::vector<T2> &)>
+          &function_read_answer = [](const auto, const auto &) {});
```
masterleinad:
```suggestion
          &function_read_answer = [](const unsigned int, const std::vector<T2> &) {});
```

tests/base/consensus_algorithm_01.cc
```diff
@@ -0,0 +1,71 @@
...
+
+  dealii::Utilities::MPI::AnonymousConsensusAlgorithmProcess<
+    dealii::types::global_dof_index,
+    unsigned int>
+    process(
+      [&]() {
+        std::vector<unsigned int> result{(my_rank + 1) % n_rank};
+        return result;
+      },
+      [&](const auto other_rank, auto &send_buffer) {
```
masterleinad:
```suggestion
      [&](const unsigned int other_rank, std::vector<dealii::types::global_dof_index> &send_buffer) {
```
and similarly below.

peterrum:
@masterleinad done!

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,205 @@
...
+
+DEAL_II_NAMESPACE_OPEN
+
+
+namespace Utilities
+{
+  namespace MPI
+  {
+    /**
+     * Anonymous version of the Utilities::MPI::ConsensusAlgorithmProcess.
```
masterleinad:
```suggestion
     * An anonymous version of the Utilities::MPI::ConsensusAlgorithmProcess.
```
Also, for me, it's not obvious what "anonymous" means in this context (without you explaining in the pull request discussion). I would just say that this class implements `Utilities::MPI::ConsensusAlgorithmProcess` using user-provided function wrappers. Of course, lambdas are a good example.

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,205 @@
...
+     * Advantage of this class is that users do not have to write their
```
masterleinad:
```suggestion
     * The advantage of this class is that users do not have to write their
```

peterrum:
@masterleinad Done!

include/deal.II/base/mpi.h
```diff
@@ -1069,6 +1069,7 @@ namespace Utilities
                   const std::vector<T2> &recv_buffer);
     };
 
+
```
kronbichler: Can you add another empty line as we typically use three between different classes?

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,206 @@
...
+     * The advantage of this class is that users do not have to write their
+     * own implementation but can register lambda functions directly.
+     */
+    template <typename T1, typename T2>
+    class AnonymousConsensusAlgorithmProcess
+      : public ConsensusAlgorithmProcess<T1, T2>
+    {
+    public:
+      /**
+       * Register functions that should be called when calling the methods of
```
kronbichler: This seems a bit confusing. How about:
```suggestion
       * Register functions that should be called for implementing the interface of
```

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,206 @@
...
+       * @param function_prepare_buffer_for_answer called during
+       *   `prepare_buffer_for_answer`.
+       * @param function_read_answer called during `read_answer`.
+       */
+      AnonymousConsensusAlgorithmProcess(
+        const std::function<std::vector<unsigned int>()>
+          &function_compute_targets,
+        const std::function<void(const unsigned int, std::vector<T1> &)>
+          &function_create_request =
+            [](const unsigned int, std::vector<T1> &) {},
```
kronbichler: I guess you prefer not to check for valid non-empty functions in the implementation below - otherwise you could have simply written `{}` as default argument. I am fine with the current setup, as it plays the ball back to the user in case they get a bad function exception.

include/deal.II/base/mpi_consensus_algorithm.h
```diff
@@ -0,0 +1,206 @@
...
+                                 std::vector<T2> &)> &function_answer_request =
+          [](const unsigned int, const std::vector<T1> &, std::vector<T2> &) {},
+        const std::function<void(const unsigned int, std::vector<T2> &)>
+          &function_prepare_buffer_for_answer =
+            [](const unsigned int, std::vector<T2> &) {},
+        const std::function<void(const unsigned int, const std::vector<T2> &)>
+          &function_read_answer =
+            [](const unsigned int, const std::vector<T2> &) {});
+
+
```
kronbichler: While I'm nit-picking about empty lines, please remove one here and in all places below - we tend to insert one between the different methods of a class.

peterrum:
@kronbichler Done!

> (including the promised follow-up).

Good that you remember me ;)

Issue #9573, opened 2020-02-28
Improve documentation/asserts of MGTransferMatrixFree build() usage


nfehn:
When using the member function `` interpolate_to_mg()``, a prerequisite is to call the member function ``build()`` beforehand. This should be highlighted in the documentation, and should be checked also for all the other member functions. Further, meaningful asserts should be inserted into the code that give a hint that one has forgotten to call ``build()`` beforehand.

peterrum:
I guess we could set in debug mode an internal bool `build_has_been_called` here:

https://github.com/dealii/dealii/blob/622eb76bc3a4f7b28ef7200390f4a4407247b5bf/source/multigrid/mg_transfer_matrix_free.cc#L95-L102

and assert it at different places (like `interpolate_to_mg()`) similarily as we do in `FE(Face)Evaluation`.

@kronbichler What do you think?

nfehn:
I am not so familiar with the details of the implementation, but I think a ``bool`` should be avoided at any price as this is very often redundant information and therefore very error prone. It should at least not be done for reasons of computational efficiency in my opinion. I think the aim should be a member function ``build_has_been_called()`` that can extract the information from any other data structure of this class?

kronbichler:
A simple check would be
```c++
  AssertDimension(sizes.size(), dof_handler.get_triangulation.n_levels());
```

peterrum:
@kronbichler You mean:

```cpp
  Assert(sizes.size() == dof_handler.get_triangulation.n_levels(), 
         ExcMessage ("MGLevelGlobalTransfer::build() has not been called!"));
```

else the user still wouldn't know the actual reason, would he?

kronbichler:
Yes, that would be a better message.

PR #9576, opened 2020-02-28
Add assert to MGLevelGlobalTransfer

peterrum:
references #9573

### DIFF
--- a/include/deal.II/multigrid/mg_transfer.templates.h
+++ b/include/deal.II/multigrid/mg_transfer.templates.h
@@ -210,6 +210,9 @@ MGLevelGlobalTransfer<VectorType>::copy_to_mg(
   MGLevelObject<VectorType> &      dst,
   const InVector &                 src) const
 {
+  Assert(copy_indices.size() ==
+           dof_handler.get_triangulation().n_global_levels(),
+         ExcMessage("MGTransfer::build() has not been called!"));
   AssertIndexRange(dst.max_level(),
                    dof_handler.get_triangulation().n_global_levels());
   AssertIndexRange(dst.min_level(), dst.max_level() + 1);
@@ -280,6 +283,9 @@ MGLevelGlobalTransfer<VectorType>::copy_from_mg(
   const MGLevelObject<VectorType> &src) const
 {
   (void)dof_handler;
+  Assert(copy_indices.size() ==
+           dof_handler.get_triangulation().n_global_levels(),
+         ExcMessage("MGTransfer::build() has not been called!"));
   AssertIndexRange(src.max_level(),
                    dof_handler.get_triangulation().n_global_levels());
   AssertIndexRange(src.min_level(), src.max_level() + 1);
@@ -349,10 +355,14 @@ template <typename VectorType>
 template <int dim, class OutVector, int spacedim>
 void
 MGLevelGlobalTransfer<VectorType>::copy_from_mg_add(
-  const DoFHandler<dim, spacedim> & /*dof_handler*/,
+  const DoFHandler<dim, spacedim> &dof_handler,
   OutVector &                      dst,
   const MGLevelObject<VectorType> &src) const
 {
+  (void)dof_handler;
+  Assert(copy_indices.size() ==
+           dof_handler.get_triangulation().n_global_levels(),
+         ExcMessage("MGTransfer::build() has not been called!"));
   // For non-DG: degrees of freedom in the refinement face may need special
   // attention, since they belong to the coarse level, but have fine level
   // basis functions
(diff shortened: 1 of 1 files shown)


include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -279,7 +281,8 @@ MGLevelGlobalTransfer<VectorType>::copy_from_mg(
   OutVector &                      dst,
   const MGLevelObject<VectorType> &src) const
 {
-  (void)dof_handler;
+  Assert(sizes.size() == dof_handler.get_triangulation().n_levels(),
```
bangerth: You still need the `(void)` line because otherwise you'll get warnings in release mode about the unused variable.

include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -349,10 +352,12 @@ template <typename VectorType>
 template <int dim, class OutVector, int spacedim>
 void
 MGLevelGlobalTransfer<VectorType>::copy_from_mg_add(
-  const DoFHandler<dim, spacedim> & /*dof_handler*/,
+  const DoFHandler<dim, spacedim> &dof_handler,
```
bangerth: I'm pretty sure here as well, now that you actually give the variable a name.

include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -520,7 +529,8 @@ MGLevelGlobalTransfer<LinearAlgebra::distributed::Vector<Number>>::copy_from_mg(
   LinearAlgebra::distributed::Vector<Number2> &                    dst,
   const MGLevelObject<LinearAlgebra::distributed::Vector<Number>> &src) const
 {
-  (void)dof_handler;
```
bangerth: and here

include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -591,10 +601,12 @@ template <int dim, typename Number2, int spacedim>
 void
 MGLevelGlobalTransfer<LinearAlgebra::distributed::Vector<Number>>::
   copy_from_mg_add(
-    const DoFHandler<dim, spacedim> & /*dof_handler*/,
-    LinearAlgebra::distributed::Vector<Number2> &                    dst,
+    const DoFHandler<dim, spacedim> &            dof_handler,
+    LinearAlgebra::distributed::Vector<Number2> &dst,
     const MGLevelObject<LinearAlgebra::distributed::Vector<Number>> &src) const
 {
+  Assert(sizes.size() == dof_handler.get_triangulation().n_levels(),
```
bangerth: and here

include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -210,6 +210,8 @@ MGLevelGlobalTransfer<VectorType>::copy_to_mg(
   MGLevelObject<VectorType> &      dst,
   const InVector &                 src) const
 {
+  Assert(sizes.size() == dof_handler.get_triangulation().n_levels(),
+         ExcMessage("MGLevelGlobalTransfer::build() has not been called!"));
```
kronbichler: I think we should write 
```suggestion
         ExcMessage("MGTransfer::build() has not been called!"));
```
The problem is that `build` sits in derived classes like `MGTransferMatrixFree` or `MGTransferPrebuilt`; now we cannot give the right name, so I suggest a simple placeholder that might give an indication of what happens already. Or to extend this to say `The derived MG transfer class did not call build()`.

peterrum:
@bangerth @kronbichler I have made the changes!

include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -210,6 +210,8 @@ MGLevelGlobalTransfer<VectorType>::copy_to_mg(
   MGLevelObject<VectorType> &      dst,
   const InVector &                 src) const
 {
+  Assert(sizes.size() == dof_handler.get_triangulation().n_levels(),
```
kronbichler: I found another bug in here (and in all other cases below):
```suggestion
  Assert(sizes.size() == dof_handler.get_triangulation().n_global_levels(),
```

peterrum: @kronbichler This does not seem to work either (see CI).

peterrum:
CI seems to have issues. I have run locally `ctest -j20 -R multigrid`:
```bash
100% tests passed, 0 tests failed out of 170
```

kronbichler:
Test failure is unrelated.

nfehn:
I would have preferred a private member function according to the "don't repeat yourself" paradigm instead of writing the same code multiple times, see issue https://github.com/dealii/dealii/issues/9573.

masterleinad:
> I would have preferred a private member function according to the "don't repeat yourself" paradigm instead of writing the same code multiple times, see issue #9573.

Just propose a pull request yourself next time if you have a better idea.

nfehn:
We discussed this in the related issue and thought that we agreed on the private member function version. Sorry that I did not have a look at it during the weekend.Nevertheless, I opened a new PR.

PR #9579, opened 2020-02-28
Add GeometryInfo::unit_tangential_vectors

kronbichler:
We have several places where we construct unit tangential vectors (and I am working on an additional one). Thus, I suggest we bring this functionality into the main `GeometryInfo` class as a `constexpr` array. Without the 105 lines of the test I added, this actually reduces the number of lines of code. :-)


include/deal.II/base/geometry_info.h
```diff
@@ -243,6 +269,35 @@ namespace internal
                  Tensor<1, 4>{{0, 0, 0, 1}}}};
       }
 
+      static constexpr std::array<std::array<Tensor<1, 4>, 3>, 8>
```
peterrum: 4D?

kronbichler: We need certain 4D instantiations of this class for dimension-independent programming up to 3D - and this includes the static member arrays.

peterrum: I am curious. Can you give an example?

rezarastak:
We have `dim - 1` unit tangential vectors. Maybe I missed it in the documentation, but do they have any defined ordering? Are we using a right-hand-rule to define which tangent vector is first and which is second for each face in 3D?

kronbichler:
> Are we using a right-hand-rule to define which tangent vector is first and which is second for each face in 3D?

Yes, these vectors follow the right-hand rule. I will augment the documentation.

include/deal.II/base/geometry_info.h
kronbichler: We used to need `GeometryInfo<4>::line_to_cell_vertices`, `child_cell_on_face`, `vertex_to_face` and a few more when we had those in the `geometry_info.cc` file (you can browse the history of `geometry_info.cc` for that). More than that, I do not know so much for the current state because we seem to have become better at avoiding some of those 4D cases. For example, I can link examples if I comment out `template class GeometryInfo<4>`. This indicates that we might be able to get rid of some things if someone wanted to. On the other hand, there is probably better use for time than to find the minimal set of things that are needed, as long as the amount of code we need to maintain is not overwhelming. On the other hand, it's not that one could easily extend the 3D concepts to 4D in the full generality, as you know.

bangerth:
If anyone wanted to try, just place `static_assert(dim<4, "");` into the class and see what happens.

PR #9580, opened 2020-02-28
Add rank 2 tensor orthogonalization function

MFraters:
I am needing this function for a aspect project and I thought it might be useful for deal.ii as well. It orthogonalizes a rank 2 tensor by returning the nearest orthogonal tensor through computing a SVD decomposition of the tensor and putting the parts together as `((U*I)*V^T)^T`. I be happy to learn how to improve this implementation.

I included a test to show that is works. It is related to #8739.

### DIFF
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -2631,7 +2631,7 @@ cofactor(const Tensor<2, dim, Number> &t)
  * @relatesalso Tensor
  */
 template <int dim, typename Number>
-constexpr Tensor<2, dim, Number>
+Tensor<2, dim, Number>
 project_onto_orthogonal_tensors(const Tensor<2, dim, Number> &tensor,
                                 const double                  tolerance)
 {
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -2618,31 +2618,29 @@ cofactor(const Tensor<2, dim, Number> &t)
 
 
 /**
- * Return the nearest orthogonal matrix using a SVD if the
- * deteriminant is more than a tolerance away from one.
- * The orthogonalization is done by combining the products
- * of the SVD decomposition: $((U*I)*V^T)^T$, where I is the
- * idententy matrix and $U$ and $V$ are computed from the SVD
- * decomposition: $\mathbf U \cdot \mathbf S \cdot \mathbf V^T$
+ * Return the nearest orthogonal matrix using a SVD if the determinant is
+ * more than a tolerance away from one. The orthogonalization is done by
+ * combining the products of the SVD decomposition: $U V^T$, where
+ * $U$ and $V$ are computed from the SVD decomposition: $\mathbf U  \mathbf S
+ * \mathbf V^T$, effectively replacing $\mathbf S$ with the identity matrix.
+ * @param tensor The tensor which to find the closest orthogonal
+ * tensor to.
+ * @param tolerance If the $\text{determinant} - 1$ is smaller than
+ * this value, it will just return the current tensor.
+ * Otherwise it will return the nearest orthogonal tensor.
  * @relatesalso Tensor
  */
 template <int dim, typename Number>
 constexpr Tensor<2, dim, Number>
-orthogonalize(const Tensor<2, dim, Number> &tensor, const double tolerance)
+project_onto_orthogonal_tensors(const Tensor<2, dim, Number> &tensor,
+                                const double                  tolerance)
 {
   if (std::abs(determinant(tensor) - 1.0) > tolerance)
     {
-      LAPACKFullMatrix<Number> identity_matrix(dim);
-      for (size_t i = 0; i < dim; i++)
-        {
-          identity_matrix.set(i, i, 1.);
-        }
-
       Tensor<2, dim, Number>   output_tensor;
       FullMatrix<Number>       matrix(dim);
       LAPACKFullMatrix<Number> lapack_matrix(dim);
       LAPACKFullMatrix<Number> result(dim);
-      LAPACKFullMatrix<Number> result2(dim);
 
       // todo: find or add dealii functionallity to copy in one step.
       matrix.copy_from(tensor);
@@ -2651,12 +2649,11 @@ orthogonalize(const Tensor<2, dim, Number> &tensor, const double tolerance)
       // now compute the svd of the matrices
       lapack_matrix.compute_svd();
 
-      // Use the SVD results to orthogonalize: ((U*I)*V^T)^T
-      lapack_matrix.get_svd_u().mmult(result, identity_matrix);
-      result.mmult(result2, (lapack_matrix.get_svd_vt()));
+      // Use the SVD results to orthogonalize: $U V^T$
+      lapack_matrix.get_svd_u().mmult(result, lapack_matrix.get_svd_vt());
 
       // todo: find or add dealii functionallity to copy in one step.
-      matrix = result2;
+      matrix = result;
       matrix.copy_to(output_tensor);
       return output_tensor;
     }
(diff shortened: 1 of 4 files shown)


include/deal.II/base/tensor.h
```diff
@@ -2613,6 +2617,53 @@ cofactor(const Tensor<2, dim, Number> &t)
 }
 
 
+/**
+ * Return the nearest orthogonal matrix using a SVD if the
+ * deteriminant is more than a tolerance away from one.
+ * The orthogonalization is done by combining the products
+ * of the SVD decomposition: $((U*I)*V^T)^T$, where I is the
+ * idententy matrix and $U$ and $V$ are computed from the SVD
```
masterleinad:
```suggestion
 * identity matrix and $U$ and $V$ are computed from the SVD
```

include/deal.II/base/tensor.h
```diff
@@ -2613,6 +2617,53 @@ cofactor(const Tensor<2, dim, Number> &t)
...
+ * decomposition: $\mathbf U \cdot \mathbf S \cdot \mathbf V^T$
+ * @relatesalso Tensor
+ */
+template <int dim, typename Number>
+constexpr Tensor<2, dim, Number>
+orthogonalize(const Tensor<2, dim, Number> &tensor, const double tolerance)
```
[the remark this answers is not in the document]
bangerth: Separately, I don't think that `orthogonalize` is the right name for the function. I think what you're doing is compute some sort of projection onto the set of orthogonal matrices. Maybe `project_onto_orthogonal_tensors()`?

include/deal.II/base/tensor.h
```diff
@@ -2613,6 +2617,53 @@ cofactor(const Tensor<2, dim, Number> &t)
 }
 
 
+/**
+ * Return the nearest orthogonal matrix using a SVD if the
+ * deteriminant is more than a tolerance away from one.
+ * The orthogonalization is done by combining the products
+ * of the SVD decomposition: $((U*I)*V^T)^T$, where I is the
```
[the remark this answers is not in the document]
MFraters: Thanks for the review. Yes, you are right. I can (and will) remove the identity matrix. 

What the method does is is basically replacing Sigma with a identity matrix when reassembling it, making the resulting matrix orthogonal. This works because if the input matrix was orthogonal, the decomposition would have been `UIV*`. There is some more info [here](https://en.wikipedia.org/wiki/Singular_value_decomposition#Nearest_orthogonal_matrix). I think this only works well if the matrix is already nearly orthogonal (which is the case for me) and am not sure how well it works if the matrix is far from it, although the question is what it then means to orthogonality. 

Looking at (this page)[https://en.wikipedia.org/wiki/Orthogonalization], I noticed that this is not the recommended way by wikipedia, so I could add a parameter so that users can choose a method to orthogonalize.

My application is that I have a rotation matrix which is changed when computing Latice Preferred Orientation (LPO). After that computation, due to rounding errors, the matrix might not be fully orthogonal anymore and over time these errors accumulate. I use this to keep the rotation matrix orthogonal. I use the SVD method because some other code to compute this, also used this trick to keep the matrix orthogonal.

include/deal.II/base/tensor.h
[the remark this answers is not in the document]
MFraters: I am fine with changing the name. `project_onto_orthogonal_tensors()` is indeed more clear.

MFraters:
I had to remove the constant expression statement to get one of the testers happy (CI -> serial). 

I am happy enough with the current implementation. Like one of the todo's states, it would be nice to condense those copy lines into one line without having to go through FullMatrix, but I think that would be a different pull request to create that functionality and then apply it.

include/deal.II/base/tensor.h
```diff
@@ -2613,6 +2617,50 @@ cofactor(const Tensor<2, dim, Number> &t)
 }
 
 
+/**
+ * Return the nearest orthogonal matrix using a SVD if the determinant is
+ * more than a tolerance away from one. The orthogonalization is done by
+ * combining the products of the SVD decomposition: $((U V^T)^T$, where
```
bangerth: I think I haven't quite understand the purpose of the last `^T` yet. Apologies for not picking up on that earlier.

Here's what I think you want: Let `A` be a matrix. if `det A \approx 1`, then the matrix is already orthogonal and you just return `A`. If it isn't already orthogonal, then you do `A = U S V^T` and you want to return `B = U V^T`. (This is a projection because if you apply the operation twice, you're going to get `B` again.) But the documentation says that you're not going to return `B = U V^T`, but instead `B^T = (U V^T)^T`. I don't understand the transposition at the end -- to make that compatible with the case where the determinant is one, you would have to return `A^T` in the first case as well, but you don't.

bangerth:
```suggestion
 * combining the products of the SVD decomposition: $(U V^T)^T = V U^T$, where
```

MFraters: Hey Wolfgang,

You are right, the transposed in the documentation is incorrect. The correct equation is I believe to remove the outer transposed from it, and I have done so.

It might be good explain how it ended up there in the first place. While making this function I had to deal with the fact that the LapackFullMatrix class stores the matrix in a transposed way. So I had to re transpose the answer. To remind myself of that I had written it down the way I did. 

Later I changed to code to use the copy functions, which automatically account for that (see https://dealii.org/developer/doxygen/deal.II/classLAPACKFullMatrix.html#a0a7678e72a99b1adbfcf72c76bbc66e8), but when expanding the documentation, I did not update that part of the equation. I think that it is correct now.

MFraters:
I am not sure I understand the errors for the failing `adolc/tensor_functions_01_1.debug` test in the CI tester. 

Especially the errors `error: static assertion failed: The FullMatrix class does not support auto-differentiable numbers.` and `error: ‘const class dealii::Tensor<2, 3, adouble>’ has no member named ‘m’
   this->reinit(M.m(), M.n());` are a bit of a mystery to me why they are causing problems. Could I get some help with this?

tests/ad_common_tests/tensor_functions_01.h
```diff
@@ -55,14 +55,16 @@ test_tensor()
...
-  const ADNumberType                 A_l1_norm   = l1_norm(A);
-  const ADNumberType                 A_linf_norm = linfty_norm(A);
+  const ADNumberType                 det_A = determinant(A);
+  const ADNumberType                 tr_A  = trace(A);
+  const Tensor<2, dim, ADNumberType> A_inv = invert(A);
+  const Tensor<2, dim, ADNumberType> A_T   = transpose(A);
+  const Tensor<2, dim, ADNumberType> A_adj = adjugate(A);
+  const Tensor<2, dim, ADNumberType> A_cof = cofactor(A);
+  const Tensor<2, dim, ADNumberType> A_orth =
+    project_onto_orthogonal_tensors(A, 1e-8);
```
bangerth: You are trying to test your new function here with auto-differentiation types. But that just doesn't work. Just remove this line again and write a separate test (or just go with the existing modification of the other test).

MFraters: Oke, thanks. Since I think the other test already covers the code I will just remove this line. That would mean that this function is only tested for doubles, which is fine with me. Is that what you meant?

bangerth:
On 3/4/20 6:31 PM, Menno Fraters wrote:
> Oke, thanks. Since I think the other test already covers the code I will 
> just remove this line. That would mean that this function is only tested 
> for doubles, which is fine with me. Is that what you meant?

Yes, I think that is fine. Or you could write a separate test for just 
this function that checks it for both double and float. I think that 
there are several such tests already in |tests/tensors/|.

MFraters:
Ah, ok. I will change the test to also test float. Need to recompile for that, so it may take a while :)

MFraters:
I tried to make a float version, but the FullMatrix copy_to() function does not seem support floats

masterleinad:
> I tried to make a float version, but the FullMatrix copy_to() function does not seem support floats

Yes, it might be a mistake not to have a template paramter for the number type of the `Tensor` object.

Issue #9550, opened 2020-02-19
discuss 64bit typedef in places outside of DoFs


tjhei:
(prompted by a comment in #9548)

We currently have: ``types::coarse_cell_id`` and ``types::global_dof_index``, which are both 64 bit (unsigned long long int) if DEAL_II_WITH_64BIT_INDICES is set.

There is several other places in the library where 32 bits are not enough, for example the global number of cells (n_global_active_cells() etc). We currently use the global_dof_index as the return value for that, which seems somewhat awkward.

Should we change that? If yes, a few options come to mind:
1. Rename ``types::global_dof_index`` to ``types::global_index`` and use everywhere. This breaks a lot of code.
2. Unconditionally use size_t (or a typedef) for things where performance doesn't matter (for the global number of cells for example). I kind of like that this simplify things and reduces bugs (we no longer have two different code paths).
3. Leave things as is (use global_dof_index everyhwere)
4. Add a new typedef that is an alias for global_dof_index.

drwells:
2. seems like the right idea - any time we return a size I think we should use `std::size_t`.

bangerth:
I'm a big fan of `typedef`s (i.e., option 4) because that provides some kind of implicit documentation as well. If a function returns a `types::global_dof_index`, you know that it returns something that is related to one particular DoF, or their number. Similarly, if a function takes a vector of `global_dof_index`, you know that it isn't a vector of numbers of cells, for example. So there is an implicit usefulness to using `typedef`s because they provide context to what a function or argument is that would otherwise have to be expressed in words in the documentation.

So I would prefer option 4.

kronbichler:
One problem with too many `typedef`s is that it increases the load of notation a new user needs to go through - even if we find words that are perfectly intuitive to us developers, a newcomer might have a completely different view and feel the threshold to get used with the terms of deal.II to be overly high. Therefore I am agreeing with @drwells that we should try `std::size_t` (or the STL concept `size_type`) as much as possible for sizes because it reduces the mental load; we cannot use them for all numbers in arrays because 32 bit integers take half the memory of 64 bit integers and there are many memory-bound situations, so there needs to be a compromise somewhat (and this is the reason why we do still have lots of `unsigned int`s around).

tjhei:
It seems that in places where we are not concerned about performance/memory, we can unconditionally use a 64 bit type (like n_global_active_cells). But we also don't gain much by doing so.
[...]

PR #9581, opened 2020-03-01
Use uint64_t instead of 'unsigned long long int'.

bangerth:
I wanted to do #9550, but before I get there, I realized that there is a worthwhile cleanup: We have historically used `unsigned long long int` for `global_dof_index` in 64-bit mode. That is what worked in the past if one wanted to get a 64-bit integer, but it is absolutely not easy to understand that that is what it actually is. It's also not documented in the standard -- like `int`, a compiler might choose something different in the same way as a `long int` on 32-bit platforms typically is the same as `int`. This leads to endless confusion.

But, C++11 introduces `uint64_t`, and we should absolutely use it because it documents exactly what we want. Even more interesting to me was that MPI version 2.2 (and maybe earlier) have a matching `MPI_UINT64_T` that we can use in that case.

Do so.

/rebuild

### DIFF
--- a/include/deal.II/base/types.h
+++ b/include/deal.II/base/types.h
@@ -20,6 +20,7 @@
 #include <deal.II/base/config.h>
 
 #include <cstddef>
+#include <cstdint>
 
 
 DEAL_II_NAMESPACE_OPEN
@@ -45,13 +46,13 @@ namespace types
   /**
    * The type used for global indices of vertices.
    */
-  using global_vertex_index = unsigned long long int;
+  using global_vertex_index = uint64_t;
 
   /**
    * An identifier that denotes the MPI type associated with
    * types::global_vertex_index.
    */
-#define DEAL_II_VERTEX_INDEX_MPI_TYPE MPI_UNSIGNED_LONG_LONG
+#define DEAL_II_VERTEX_INDEX_MPI_TYPE MPI_UINT64_T
 
 #ifdef DEAL_II_WITH_64BIT_INDICES
   /**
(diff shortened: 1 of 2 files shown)


masterleinad:
The only concern I have here is that implementing `uint64_t` is optional. I think it is fine to only support platforms and compilers that actually do. What about `uint_fast64_t` or `uint_least64_t` otherwise?

bangerth:
I can't really see compilers on platforms we care about not supporting `uint64_t`. Since using `uint_*64_t` would entail also thinking about what to do with the MPI type tags, I"d say we shouldn't worry unless we have concrete evidence that we *ought* to worry.

masterleinad:
`MPI_UINT64_T` only appears in the MPI standard documents from MPI version 2.2 onward but we claim to require at least version 2.0. The first `OpenMPI` release to support MPI-2.2 was 1.7.3 released in October 2013. Given that this was 6.5 years ago, I think it is OK to lift the requirement. That would also mean we could clean up some code in `mpi.cc`.

tjhei:
@bangerth How about we add a static assert that ``sizeof`` is 64bit?

bangerth:
What specifically do you want to test? That `sizeof(uint64_t)=8`? That seems a bit pointless :-)

Or that `sizeof(unsigned long long)=8`? I had toyed with that idea when thinking about how to make sure that the MPI type tag `MPI_UNSIGNED_LONG_LONG` matches `uint64_t`, but that became obsolete when I found `MPI_UINT64_T`.

tjhei:
makes sense, sorry.

masterleinad:
So what do we do about the MPI version? Provide a workaround for MPI version less than 2.2 or raise the requirement?
Are we were aware of any platforms we care about that only come with a MPI 2.0 implementation?

bangerth:
I'd just try and give it a shot. We've got a few months.

tjhei:
> I'd just try and give it a shot. We've got a few months.

You will need to update doc/readme.html and we might want to add a news entry.

bangerth:
Since I've got a follow-up that both you and I are waiting for, let's merge this one and I'll do the changelog and readme in a separate PR.

tjhei:
I am okay with that. @masterleinad ?

PR #9582, opened 2020-03-01
Improve indentation of Jenkinsfile

masterleinad:
While looking into the failing CI for #9576, I noticed some missing indentation and tabs being used in Jenkinsfile. Also, we were having `set -x` for the MPI build, but not for the serial build.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -116,36 +116,36 @@ pipeline
             failure {
               githubNotify context: 'CI', description: 'serial build failed',  status: 'FAILURE'
             }
-      }
+          }
 
           steps
           {
-          timeout(time: 6, unit: 'HOURS')
-          {
-            sh "echo \"building on node ${env.NODE_NAME}\""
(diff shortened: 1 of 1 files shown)


masterleinad:
On a related note, it seems that the pull request branch did not include `check_doxygen.sh` but the CI was using the updated script (and failed because of that). We might consider testing a pull request after merging with `master` to avoid these problems. `GitHub Actions` and `Travis` do this anyway.

tjhei:
> We might consider testing a pull request after merging with `master` to avoid these problems.

There is an option to do this in jenkins, but I did not like the fact that a change of the SHA1 on master would cause all pull requests to be rebuilt. We can try it again, but this might cause some long queues.

tjhei:
> On a related note, it seems that the pull request branch did not include `check_doxygen.sh` but the CI was using the updated script

ouch, that is not good. As we run this inside travis, I will remove it for now.

PR #9586, opened 2020-03-02
Introduce types::global_cell_index.

### LINKED ISSUE #9550: discuss 64bit typedef in places outside of DoFs

bangerth:
While there, also clean up a bit of the documentation around that point
in the file.

Fixes #9550. Follow-up to #9581.

/rebuild

### DIFF
--- a/include/deal.II/base/types.h
+++ b/include/deal.II/base/types.h
@@ -56,13 +56,17 @@ namespace types
 
 #ifdef DEAL_II_WITH_64BIT_INDICES
   /**
-   * The type used for global indices of degrees of freedom. While in
-   * sequential computations the 4 billion indices of 32-bit unsigned integers
-   * is plenty, parallel computations using the
-   * parallel::distributed::Triangulation class can overflow this number and
-   * we need a bigger index space.
+   * The type used to denote the global index of degrees of freedom. This
+   * type is then also used for querying the global *number* of degrees
+   * of freedom, since the number is simply the largest index plus one.
    *
-   * The data type always indicates an unsigned integer type.
+   * While in sequential computations the 4 billion indices of 32-bit unsigned
+   * integers is plenty, parallel computations using (for example) the
+   * parallel::distributed::Triangulation class can overflow this number and
+   * consequently, deal.II chooses a larger integer type when
+   * configured to use 64-bit indices.
+   *
+   * The data type always corresponds to an unsigned integer type.
    *
    * See the
    * @ref GlobalDoFIndex
@@ -78,13 +82,21 @@ namespace types
 
 #else
   /**
-   * The type used for global indices of degrees of freedom. While in
-   * sequential computations the 4 billion indices of 32-bit unsigned integers
-   * is plenty, parallel computations using the
-   * parallel::distributed::Triangulation class can overflow this number and
-   * we need a bigger index space.
+   * The type used to denote the global index of degrees of freedom. This
+   * type is then also used for querying the global *number* of degrees
+   * of freedom, since the number is simply the largest index plus one.
    *
-   * The data type always indicates an unsigned integer type.
+   * While in sequential computations the 4 billion indices of 32-bit unsigned
+   * integers is plenty, parallel computations using (for example) the
+   * parallel::distributed::Triangulation class can overflow this number and
+   * consequently, deal.II chooses a larger integer type when
+   * configured to use 64-bit indices.
+   *
+   * The data type always corresponds to an unsigned integer type.
+   *
+   * See the
+   * @ref GlobalDoFIndex
+   * page for guidance on when this type should or should not be used.
    */
   using global_dof_index = unsigned int;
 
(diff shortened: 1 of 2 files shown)


include/deal.II/base/types.h
```diff
@@ -56,13 +56,17 @@ namespace types
 
 #ifdef DEAL_II_WITH_64BIT_INDICES
   /**
-   * The type used for global indices of degrees of freedom. While in
-   * sequential computations the 4 billion indices of 32-bit unsigned integers
-   * is plenty, parallel computations using the
+   * The type used to denote the global index of degree of freedom. This
+   * type is then also used for querying the global *number* of degrees
+   * of freedom, since the number is simply the largest inex plus one.
```
masterleinad:
```suggestion
   * of freedom, since the number is simply the largest index plus one.
```

include/deal.II/base/types.h
```diff
@@ -78,13 +82,21 @@ namespace types
 
 #else
   /**
-   * The type used for global indices of degrees of freedom. While in
-   * sequential computations the 4 billion indices of 32-bit unsigned integers
-   * is plenty, parallel computations using the
+   * The type used to denote the global index of degree of freedom. This
+   * type is then also used for querying the global *number* of degrees
+   * of freedom, since the number is simply the largest inex plus one.
```
masterleinad:
```suggestion
   * of freedom, since the number is simply the largest index plus one.
```

include/deal.II/base/types.h
```diff
@@ -56,13 +56,17 @@ namespace types
 
 #ifdef DEAL_II_WITH_64BIT_INDICES
   /**
-   * The type used for global indices of degrees of freedom. While in
-   * sequential computations the 4 billion indices of 32-bit unsigned integers
-   * is plenty, parallel computations using the
+   * The type used to denote the global index of degree of freedom. This
```
kronbichler:
```suggestion
   * The type used to denote the global index of degrees of freedom. This
```

include/deal.II/base/types.h
```diff
@@ -78,13 +82,21 @@ namespace types
 
 #else
   /**
-   * The type used for global indices of degrees of freedom. While in
-   * sequential computations the 4 billion indices of 32-bit unsigned integers
-   * is plenty, parallel computations using the
+   * The type used to denote the global index of degree of freedom. This
```
kronbichler:
```suggestion
   * The type used to denote the global index of degrees of freedom. This
```

include/deal.II/base/types.h
```diff
@@ -78,13 +82,21 @@ namespace types
...
-   * we need a bigger index space.
+   * consequently, deal.II chooses a larger integer type when
+   * configured to use 64-bit indices.
+   *
+   * The data type always corresponds to an unsigned integer type.
    *
-   * The data type always indicates an unsigned integer type.
+   * See the
+   * @ref GlobalDoFIndex
+   * page for guidance on when this type should or should not be used.
```
kronbichler: While at it, do we really need to duplicate the documentation or can we simply move the `#ifdef` around the variable only?

include/deal.II/base/types.h
```diff
@@ -56,13 +56,17 @@ namespace types
...
-   * The type used for global indices of degrees of freedom. While in
-   * sequential computations the 4 billion indices of 32-bit unsigned integers
-   * is plenty, parallel computations using the
+   * The type used to denote the global index of degree of freedom. This
+   * type is then also used for querying the global *number* of degrees
+   * of freedom, since the number is simply the largest index plus one.
+   *
+   * While in sequential computations the 4 billion indices of 32-bit unsigned
+   * integers is plenty, parallel computations using the
    * parallel::distributed::Triangulation class can overflow this number and
```
kronbichler: Since `parallel::distributed::Triangulation` is not the only class that can do this, should we phrase this more generally by `using, e.g., the parallel::distributed::Triangulation` class (here and below)?

bangerth:
@kronbichler -- I've addressed your comments. In particular, I avoided the repetition of the documentation. I was worried that in
```
  /**
   * ...
   */
#if abc
  using type = int;
#else
  using type = long;
#endif
```
doxygen would get confused. But it doesn't, it all works fine. So that's the style I chose.

bangerth:
This seems to compile!

tjhei:
Well then, let's see how this works out.

bangerth:
I'm hoping for better luck than last time :-))

tjhei:
> I'm hoping for better luck than last time :-))

actually, I will compile this PR on my machine first now.

tjhei:
> actually, I will compile this PR on my machine first now. Not trusting your statement. ;-)
```
/home/heister/deal-git/source/dofs/dof_renumbering.cc:433:25: error: no matching function for call to 'reverse_permutation'
          new_indices = Utilities::reverse_permutation(new_indices);
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/heister/deal-git/include/deal.II/base/utilities.h:566:3: note: candidate function not viable: no known conversion from 'vector<types::global_dof_index>' to 'const vector<unsigned int>' for 1st argument
  reverse_permutation(const std::vector<unsigned int> &permutation);
  ^
```

Did you compile this on your machine?

bangerth:
I was certain that there could be no change because all I did was add a `typedef`. But now I realize that I also changed the type for `coarse_cell_id` -- and that might cause an issue.

I've got something else big running on my machines right now, but will check this afternoon.

bangerth:
It compiles on my laptop in 64-bit mode (but without any external dependencies). Let's see what happens here.

PR #9587, opened 2020-03-02
Follow up to #9581

bangerth:
This is the promised changelog entry for #9581. I didn't think anything really needed to be updated in the readme file (nothing actively seemed wrong with the old text), but I always find something anyway.

### DIFF
--- a/doc/readme.html
+++ b/doc/readme.html
@@ -336,10 +336,16 @@
 
         <li>
             <p>
-                <i>64bit indices</i>: By default, deal.II use unsigned int (32bit) indices for degrees of freedom, using the <code>types::global_dof_index</code> type. This limits the number of unknowns to approximately four billions. If larger problem
-                must be solved, pass the
+                <i>64bit indices</i>: By default, deal.II uses
+                unsigned int (32bit) indices for degrees of freedom,
+                using the <code>types::global_dof_index</code> type.
+                (The same is true for a few other index types such as the
+                global number of cells.) This limits the number of
+                cells or unknowns to approximately four billions. If
+                larger problem must be solved, pass the
                 <code>-DDEAL_II_WITH_64BIT_INDICES=ON</code> argument to
-                <code>cmake</code>. You will not be able to solve problems of this size on a single machine, but only when using clusters of computers and linear algebra packages PETSc or Trilinos. To use this option with PETSc, PETSc must be compiled
+                <code>cmake</code>. To use this option with
+                PETSc, PETSc must be compiled
                 with the option <code>--with-64-bit-indices</code>.
             </p>
         </li>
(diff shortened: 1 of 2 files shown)


doc/news/changes/minor/20200301Bangerth
```diff
@@ -0,0 +1,8 @@
+New: The underlying type for types::global_dof_index used to be
+`unsigned long long int`. On all reasonable systems, this is actually
+a 64-bit unsigned integer type, but the name wouldn't tell you that
+unless you already knew. This is now fixed: Instead of the mouthfull
```
kronbichler:
```suggestion
unless you already knew. This is now fixed: Instead of the mouthful
```

doc/readme.html
```diff
@@ -336,10 +336,19 @@ <h4>Selecting optional library features</h4>
...
+                unsigned int (32bit) indices for degrees of freedom,
+                using the <code>types::global_dof_index</code> type.
+                (The same is true for a few other index types such as the
+                global number of cells.) This limits the number of
+                cells or unknowns to approximately four billions. If
+                larger problem must be solved, pass the
                 <code>-DDEAL_II_WITH_64BIT_INDICES=ON</code> argument to
-                <code>cmake</code>. You will not be able to solve problems of this size on a single machine, but only when using clusters of computers and linear algebra packages PETSc or Trilinos. To use this option with PETSc, PETSc must be compiled
+                <code>cmake</code>. You will not be able to solve
+                problems of this size on a single machine, but only
```
kronbichler: I don't think that is true any more. If you have a system with 512 GB of RAM, which is a bit more than the regular workstation/cluster but not that unusual either, you can solve systems of 6-8 billion dofs with explicit time integrators like the one from step-67 (#8319). I suggest to insert `typically`.

doc/readme.html
```diff
@@ -336,10 +336,19 @@ <h4>Selecting optional library features</h4>
...
+                when using clusters of computers and linear algebra
+                packages PETSc or Trilinos. To use this option with
```
kronbichler: It also works with deal.II's own parallel vectors.

doc/readme.html
bangerth: Good point. I was going to add "unless you have exceptional amounts of memory and patience", but I'm sure that even that will become moot over time :-)

doc/readme.html
bangerth: I just removed that whole reference to PETSc and Trilinos.

bangerth:
@kronbichler -- I've made the changes, and also:
* Moved the changelog entry to `incompatibilities`
* Added a paragraph that explained that strictly speaking, things are incompatible but that few places will notice (I sure hope that the latter is true).

PR #9589, opened 2020-03-02
Fix compile error of GeometryInfo::unit_tangential_vectors on clang-4/5

kronbichler:
This is from my previous commit #9579 - clang-4 and clang-5 complain about the constructor of the inner array argument. Now I add two braces `{{}}` and hope that all other CI architectures like MSVC are also happy.
Here is one of the associated failures:
https://cdash.43-1.org/viewBuildError.php?buildid=4633

### DIFF
--- a/include/deal.II/base/geometry_info.h
+++ b/include/deal.II/base/geometry_info.h
@@ -71,7 +71,7 @@ namespace internal
       static constexpr std::array<std::array<Tensor<1, 1>, 0>, 2>
       unit_tangential_vectors()
       {
-        return {{{}, {}}};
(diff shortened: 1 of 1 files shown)


include/deal.II/base/geometry_info.h
```diff
@@ -71,7 +71,7 @@ namespace internal
       static constexpr std::array<std::array<Tensor<1, 1>, 0>, 2>
       unit_tangential_vectors()
       {
-        return {{{}, {}}};
+        return {{{{}}, {{}}}};
```
bangerth: :-))

Issue #9588, opened 2020-03-02
Compilation failes for DEAL_II_WITH_64BIT_INDICES="ON"


peterrum:
The master branch does not compile after the recent changes for 64 bit indices. E.g.: 
```bash
/home/munch/sw-index/dealii/source/dofs/dof_renumbering.cc: In function ‘void dealii::DoFRenumbering::compute_Cuthill_McKee(std::vector<long unsigned int>&, const DoFHandlerType&, bool, bool, const std::vector<long unsigned int>&)’:
/home/munch/sw-index/dealii/source/dofs/dof_renumbering.cc:433:67: error: no matching function for call to ‘reverse_permutation(std::vector<long unsigned int>&)’
  433 |           new_indices = Utilities::reverse_permutation(new_indices);
      |                                                                   ^
In file included from /home/munch/sw-index/dealii/include/deal.II/base/tensor.h:27,
                 from /home/munch/sw-index/dealii/include/deal.II/base/point.h:23,
                 from /home/munch/sw-index/dealii/include/deal.II/base/quadrature.h:22,
                 from /home/munch/sw-index/dealii/include/deal.II/base/quadrature_lib.h:22,
                 from /home/munch/sw-index/dealii/source/dofs/dof_renumbering.cc:16:
/home/munch/sw-index/dealii/include/deal.II/base/utilities.h:566:3: note: candidate: ‘std::vector<unsigned int> dealii::Utilities::reverse_permutation(const std::vector<unsigned int>&)’
  566 |   reverse_permutation(const std::vector<unsigned int> &permutation);
      |   ^~~~~~~~~~~~~~~~~~~
/home/munch/sw-index/dealii/include/deal.II/base/utilities.h:566:56: note:   no known conversion for argument 1 from ‘std::vector<long unsigned int>’ to ‘const std::vector<unsigned int>&’
  566 |   reverse_permutation(const std::vector<unsigned int> &permutation);
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
/home/munch/sw-index/dealii/include/deal.II/base/utilities.h:582:3: note: candidate: ‘std::vector<long long unsigned int> dealii::Utilities::reverse_permutation(const std::vector<long long unsigned int>&)’
  582 |   reverse_permutation(const std::vector<unsigned long long int> &permutation);
      |   ^~~~~~~~~~~~~~~~~~~
[...]

PR #9590, opened 2020-03-02
Eliminate two places with unsigned long long int

kronbichler:
Now that we switched to `std::uint64_t`, we should also address two users of those indices in `TableHandler` and `Utilities`. This addresses the compile error reported in #9588 but not yet the warnings.

Now the only place left where I found `unsigned long long int` by grepping is `include/deal.II/lac/cuda_atomic.h` where we must use `ull` because the interface with CUDA requires it (but then again the assumption is to alias a double with a 64-bit integer - but that is up to CUDA to decide, not us).

### DIFF
--- a/include/deal.II/base/utilities.h
+++ b/include/deal.II/base/utilities.h
@@ -578,16 +578,16 @@ namespace Utilities
    * $p_i\in [0,N)$ and $p_i\neq p_j$ for $i\neq j$), produce the reverse
    * permutation $q_i=N-1-p_i$.
    */
-  std::vector<unsigned long long int>
-  reverse_permutation(const std::vector<unsigned long long int> &permutation);
+  std::vector<std::uint64_t>
+  reverse_permutation(const std::vector<std::uint64_t> &permutation);
 
   /**
    * Given a permutation vector (i.e. a vector $p_0\ldots p_{N-1}$ where each
    * $p_i\in [0,N)$ and $p_i\neq p_j$ for $i\neq j$), produce the inverse
    * permutation $q_0\ldots q_{N-1}$ so that $q_{p_i}=p_{q_i}=i$.
    */
-  std::vector<unsigned long long int>
-  invert_permutation(const std::vector<unsigned long long int> &permutation);
+  std::vector<std::uint64_t>
+  invert_permutation(const std::vector<std::uint64_t> &permutation);
 
   /**
    * Given an arbitrary object of type T, use boost::serialization utilities
--- a/source/base/utilities.cc
+++ b/source/base/utilities.cc
@@ -868,79 +868,6 @@ namespace Utilities
 
 
 
-  std::vector<unsigned int>
-  reverse_permutation(const std::vector<unsigned int> &permutation)
-  {
-    const unsigned int n = permutation.size();
-
-    std::vector<unsigned int> out(n);
-    for (unsigned int i = 0; i < n; ++i)
-      out[i] = n - 1 - permutation[i];
-
-    return out;
-  }
-
-
-
-  std::vector<unsigned int>
-  invert_permutation(const std::vector<unsigned int> &permutation)
-  {
-    const unsigned int n = permutation.size();
-
-    std::vector<unsigned int> out(n, numbers::invalid_unsigned_int);
-
-    for (unsigned int i = 0; i < n; ++i)
-      {
-        AssertIndexRange(permutation[i], n);
-        out[permutation[i]] = i;
-      }
-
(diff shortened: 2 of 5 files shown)


include/deal.II/base/utilities.h
```diff
@@ -578,16 +578,16 @@ namespace Utilities
...
+  std::vector<std::uint64_t>
+  reverse_permutation(const std::vector<std::uint64_t> &permutation);
 
   /**
    * Given a permutation vector (i.e. a vector $p_0\ldots p_{N-1}$ where each
    * $p_i\in [0,N)$ and $p_i\neq p_j$ for $i\neq j$), produce the inverse
    * permutation $q_0\ldots q_{N-1}$ so that $q_{p_i}=p_{q_i}=i$.
    */
-  std::vector<unsigned long long int>
-  invert_permutation(const std::vector<unsigned long long int> &permutation);
```
masterleinad: We could also just templatize these two functions instead of repeating the definition for the two number types.

kronbichler: Well, we do have this one: https://github.com/dealii/dealii/blob/c8233d7c7984e0cc1ddb1f19abb4295366cce289/source/base/utilities.cc#L944-L980
It seems we should simply move this above the other two functions - let me try that out in a minute; regarding the declaration I like to state the supported types given that it is only two, i.e., I prefer a compile error over a link error.

kronbichler: I have now added a second commit that removes the code duplication in favor of the templated function we had in this file already.

kronbichler:
It turns out that I forgot the convergence table test with the unsigned long long int.

source/base/utilities.cc
```diff
@@ -904,79 +906,36 @@ namespace Utilities
...
 
-    std::vector<unsigned long long int> out(n);
-    for (unsigned long long int i = 0; i < n; ++i)
-      out[i] = n - 1 - permutation[i];
 
-    return out;
+  std::vector<unsigned int>
+  reverse_permutation(const std::vector<unsigned int> &permutation)
+  {
+    return reverse_permutation<unsigned int>(permutation);
```
bangerth: do we even still need this function then?

source/base/utilities.cc
```diff
@@ -904,79 +906,36 @@ namespace Utilities
...
-      }
-
-    // check that we have actually reached
-    // all indices
-    for (unsigned long long int i = 0; i < n; ++i)
-      Assert(out[i] != numbers::invalid_unsigned_int,
-             ExcMessage("The given input permutation had duplicate entries!"));
-
-    return out;
+    return invert_permutation<unsigned int>(permutation);
```
bangerth: same here?

source/base/utilities.cc
```diff
@@ -904,79 +906,36 @@ namespace Utilities
...
-
-    std::vector<Integer> out(n);
-    for (unsigned int i = 0; i < n; ++i)
-      out[i] = n - 1 - permutation[i];
 
-    return out;
+  std::vector<std::uint64_t>
+  reverse_permutation(const std::vector<std::uint64_t> &permutation)
+  {
+    return reverse_permutation<std::uint64_t>(permutation);
```
bangerth: and here

source/base/utilities.cc
```diff
@@ -904,79 +906,36 @@ namespace Utilities
...
-      }
-
-    // check that we have actually reached
-    // all indices
-    for (unsigned int i = 0; i < n; ++i)
-      Assert(out[i] != numbers::invalid_unsigned_int,
-             ExcMessage("The given input permutation had duplicate entries!"));
-
-    return out;
+    return invert_permutation<std::uint64_t>(permutation);
```
bangerth: and here

source/base/utilities.cc
masterleinad: @kronbichler wanted to avoid a link-time error this way. As I suggested before, I would be fine with moving the declaration and the definition of the templated version to the header file and remove these two overloads. I don't feel strongly about it, though.

bangerth:
On 3/2/20 11:03 AM, Daniel Arndt wrote:
> @kronbichler <https://github.com/kronbichler> wanted to avoid a 
> link-time error this way. As I suggested before, I would be fine with 
> moving the declaration and the definition of the templated version to 
> the header file and remove these two overloads. I don't feel strongly 
> about it, though.

That was my idea as well. But let's get the current situation under 
control again and then revisit the issue.

tjhei:
I just merged. The cleanup is unrelated to the compiler errors...

PR #9591, opened 2020-03-02
Fix two 64 bit warnings by converting to long long unsigned

### LINKED ISSUE #9588: Compilation failes for DEAL_II_WITH_64BIT_INDICES="ON"

kronbichler:
Together with #9590 this fixes #9588.

As discussed in #9588, we should explicitly convert the variables to `unsigned long long int` before we use the c function `sprintf`. I do this now also for the 32 bit case as it makes code easier to read (and added a comment why). I checked `tests/lac/la_vector_output` which seems the only test to use the block write function.

### DIFF
--- a/include/deal.II/lac/la_vector.templates.h
+++ b/include/deal.II/lac/la_vector.templates.h
@@ -592,15 +592,14 @@ namespace LinearAlgebra
 
     // Other version of the following
     //  out << size() << std::endl << '[';
-    // Reason: operator<< seems to use some resources  that lead to problems in
-    // a multithreaded environment.
-    const size_type sz = this->size();
-    char            buf[16];
-#ifdef DEAL_II_WITH_64BIT_INDICES
+    // Reason: operator<< seems to use some resources that lead to problems in
+    // a multithreaded environment.  We convert the size index to
+    // unsigned long long int that is at least 64 bits to be able to output it
+    // on all platforms, since std::uint64_t is not in C.
+    const unsigned long long int sz = this->size();
+    char                         buf[16];
+
     std::sprintf(buf, "%llu", sz);
-#else
-    std::sprintf(buf, "%u", sz);
-#endif
     std::strcat(buf, "\n[");
 
     out.write(buf, std::strlen(buf));
--- a/include/deal.II/lac/vector.templates.h
+++ b/include/deal.II/lac/vector.templates.h
@@ -920,18 +920,15 @@ Vector<Number>::block_write(std::ostream &out) const
 
   // other version of the following
   //  out << size() << std::endl << '[';
-  // reason: operator<< seems to use
-  // some resources that lead to
-  // problems in a multithreaded
-  // environment
-  const size_type sz = size();
-  char            buf[16];
 
-#ifdef DEAL_II_WITH_64BIT_INDICES
+  // reason: operator<< seems to use some resources that lead to problems in a
+  // multithreaded environment. We convert the size index to
+  // unsigned long long int that is at least 64 bits to be able to output it on
+  // all platforms, since std::uint64_t is not in C.
+  const unsigned long long int sz = size();
+  char                         buf[16];
+
   std::sprintf(buf, "%llu", sz);
-#else
-  std::sprintf(buf, "%u", sz);
-#endif
   std::strcat(buf, "\n[");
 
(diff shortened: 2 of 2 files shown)


kronbichler:
I obviously confirmed that this compiles without warning on my 64 bit machine (Linux with gcc 9.2).

include/deal.II/lac/la_vector.templates.h
```diff
@@ -592,15 +592,14 @@ namespace LinearAlgebra
...
     // Other version of the following
     //  out << size() << std::endl << '[';
-    // Reason: operator<< seems to use some resources  that lead to problems in
-    // a multithreaded environment.
-    const size_type sz = this->size();
-    char            buf[16];
-#ifdef DEAL_II_WITH_64BIT_INDICES
+    // Reason: operator<< seems to use some resources that lead to problems in
+    // a multithreaded environment. We convert the size index to unsigned long
+    // long int that is at least 64 bits to be able to output it on all
```
masterleinad: Nit-picking: It's probably easier to read if we don't break the line for `unsigned long long int`. Same below.

kronbichler: Fixed!

masterleinad:
> So is this suggesting that the compiler treats `unsigned long long int` _differently_ than `uint64_t`? That is my underlying cognitive disconnect: I had assumed that they are the same type, and as a consequence it had not even occurred to me that there might be a problem with compiling because I believed that my patch is a no-op...

See https://stackoverflow.com/questions/32198368/unsigned-long-long-conflict-with-uint64-t

bangerth:
@masterleinad -- thanks for the link. I guess it's just really very awkward to see that `sizeof(long int)==sizeof(long long int)` on some systems, but at the same time `std::is_same<long int,long long int>::value == false` :-( That would never have occurred to me...

PR #9593, opened 2020-03-02
remove redundancy from MGTransfer

### LINKED ISSUE #9573: Improve documentation/asserts of MGTransferMatrixFree build() usage

nfehn:
Edit [MK]: Fixes #9573.

### DIFF
--- a/include/deal.II/multigrid/mg_transfer.h
+++ b/include/deal.II/multigrid/mg_transfer.h
@@ -417,6 +417,14 @@ protected:
    */
   SmartPointer<const MGConstrainedDoFs, MGLevelGlobalTransfer<VectorType>>
     mg_constrained_dofs;
+
+private:
+  /**
+   * This function is called to make sure that build() has been invoked.
+   */
+  template <int dim, int spacedim>
+  void
+  assert_built(const DoFHandler<dim, spacedim> &dof_handler) const;
 };
 
 
@@ -634,6 +642,14 @@ protected:
    */
   mutable MGLevelObject<LinearAlgebra::distributed::Vector<Number>>
     solution_ghosted_level_vector;
+
+private:
+  /**
+   * This function is called to make sure that build() has been invoked.
+   */
+  template <int dim, int spacedim>
+  void
+  assert_built(const DoFHandler<dim, spacedim> &dof_handler) const;
 };
 
 
--- a/include/deal.II/multigrid/mg_transfer.templates.h
+++ b/include/deal.II/multigrid/mg_transfer.templates.h
@@ -210,9 +210,7 @@ MGLevelGlobalTransfer<VectorType>::copy_to_mg(
   MGLevelObject<VectorType> &      dst,
   const InVector &                 src) const
 {
-  Assert(copy_indices.size() ==
-           dof_handler.get_triangulation().n_global_levels(),
-         ExcMessage("MGTransfer::build() has not been called!"));
+  assert_built(dof_handler);
   AssertIndexRange(dst.max_level(),
                    dof_handler.get_triangulation().n_global_levels());
   AssertIndexRange(dst.min_level(), dst.max_level() + 1);
@@ -282,10 +280,7 @@ MGLevelGlobalTransfer<VectorType>::copy_from_mg(
   OutVector &                      dst,
   const MGLevelObject<VectorType> &src) const
 {
-  (void)dof_handler;
-  Assert(copy_indices.size() ==
-           dof_handler.get_triangulation().n_global_levels(),
-         ExcMessage("MGTransfer::build() has not been called!"));
+  assert_built(dof_handler);
   AssertIndexRange(src.max_level(),
                    dof_handler.get_triangulation().n_global_levels());
   AssertIndexRange(src.min_level(), src.max_level() + 1);
@@ -359,10 +354,7 @@ MGLevelGlobalTransfer<VectorType>::copy_from_mg_add(
   OutVector &                      dst,
   const MGLevelObject<VectorType> &src) const
 {
-  (void)dof_handler;
-  Assert(copy_indices.size() ==
-           dof_handler.get_triangulation().n_global_levels(),
-         ExcMessage("MGTransfer::build() has not been called!"));
+  assert_built(dof_handler);
   // For non-DG: degrees of freedom in the refinement face may need special
   // attention, since they belong to the coarse level, but have fine level
   // basis functions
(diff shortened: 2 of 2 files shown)


nfehn:
Yes, I was a bit confused where to put the implementation? It seems as if I have to implement this function twice?

bangerth:
No, like any other function you can declare it inside the class declaration and implement it outside as an `inline` function.

nfehn:
I think it would help me if you could tell me in which file or where exactly this has to be implemented.

nfehn:
Don't see that this is a constructive comment

nfehn:
@bangerth Can you have a look at https://github.com/dealii/dealii/pull/9593/commits/5b7273f0098c9fd5f2f8816d2b7d2e6d3deda47c

I think I just had syntax problems with the many template parameters and the specialization, so that I decided to first implement in declaration of class.

include/deal.II/multigrid/mg_transfer.h
```diff
@@ -417,6 +417,14 @@ class MGLevelGlobalTransfer : public MGTransferBase<VectorType>
...
   SmartPointer<const MGConstrainedDoFs, MGLevelGlobalTransfer<VectorType>>
     mg_constrained_dofs;
+
+private:
+  /**
+   * This function is called to make sure that build() has been invoked.
+   */
+  template <int dim, int spacedim>
+  inline DEAL_II_ALWAYS_INLINE void
+  assert_built(const DoFHandler<dim, spacedim> &dof_handler) const;
```
bangerth: Yes, this looks correct to me. Except the `inline DEAL_II_ALWAYS_INLINE` is probably not necessary here because the compiler doesn't know at this point what to inline anyway.

include/deal.II/multigrid/mg_transfer.h
```diff
@@ -634,6 +642,14 @@ class MGLevelGlobalTransfer<LinearAlgebra::distributed::Vector<Number>>
...
   mutable MGLevelObject<LinearAlgebra::distributed::Vector<Number>>
     solution_ghosted_level_vector;
+
+private:
+  /**
+   * This function is called to make sure that build() has been invoked.
+   */
+  template <int dim, int spacedim>
+  inline DEAL_II_ALWAYS_INLINE void
+  assert_built(const DoFHandler<dim, spacedim> &dof_handler) const;
```
bangerth: same here

include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -399,6 +391,17 @@ MGLevelGlobalTransfer<VectorType>::set_component_to_block_map(
   component_to_block_map = map;
 }
 
+template <typename VectorType>
+template <int dim, int spacedim>
+void
+MGLevelGlobalTransfer<VectorType>::assert_built(
+  const DoFHandler<dim, spacedim> &dof_handler) const
```
bangerth: And this also looks correct to me.

include/deal.II/multigrid/mg_transfer.templates.h
```diff
@@ -663,6 +654,18 @@ MGLevelGlobalTransfer<LinearAlgebra::distributed::Vector<Number>>::
   component_to_block_map = map;
 }
 
+template <typename Number>
+template <int dim, int spacedim>
+void
+MGLevelGlobalTransfer<LinearAlgebra::distributed::Vector<Number>>::assert_built(
+  const DoFHandler<dim, spacedim> &dof_handler) const
```
bangerth: and so does this.

bangerth:
Also, everyone chill out. Patch review is a way for all of us to become better programmers -- that's true for both the writer and the reviewer. It's way for all of us to learn and to teach -- things we point out are supposed to not be an indication of a mistake, but to help the other person grow.

bangerth:
Feel free to squash your three commits into one.

/rebuild

include/deal.II/multigrid/mg_transfer.templates.h
bangerth: Actually, I think that this is where you want to have the `inline DEAL_II_ALWAYS_INLINE`.

masterleinad: I don't think that we have to force the compiler to inline this and would just remove `DEAL_II_ALWAYS_INLINE`.

masterleinad: If the function should be declared `inline` at all is a bit a question of philosophy. Templated functions don't require to be declared `inline` at all. On the other hand, if `*templates.h` are thought to be used the same way as `*.h` files and logically it makes sense to have them `inline` to be consistent with non-templated classes.

In this file, no other function is declared `inline` so I would rather omit it for consistency but I don't feel strongly about this.

bangerth: That's all fair enough :-) I don't feel adamant about it one way or the other -- I suspect that the compiler will just do the right thing, and if it doesn't, then the call to the function is still simply a jump to a `ret` instruction ;-)

masterleinad: Of course. It's just a question of style. :slightly_smiling_face:

include/deal.II/multigrid/mg_transfer.h
kronbichler: I would also remove the always inline statement as we primarily use it as a code optimization vehicle in deal.II.

nfehn:
@bangerth yes, you are right. Sorry if the tone became crude from my side. As a constructive criticism I would like to suggest that it would be a good thing if we could agree on not using "..." at the end of a sentence or comment. It just leads to misunderstandings and can be understood as a provocation. It has probably been my mistake to not mention this earlier.

nfehn:
Just because i have seen the ready-to-merge label. Please wait until tomorrow, I want to remove the inline and squash the commits.

bangerth:
No harm done. We all learn. The issue with "..." is reasonable, though that's really one of the problems of written language: it lacks a lot of the subtlety of the spoken word.

In any case, we're here to all make each other better.

nfehn:
Issue #9573 can be closed once this PR has been merged

PR #9600, opened 2020-03-03
Move ConsensusAlgorithm related functions to mpi_consensus_algorithm.h

peterrum:
As promised, this PR moves all `ConsensusAlgorithm` related functionalities to the file `mpi_consensus_algorithm.h`.

This PR did not make any further modifications.

@kronbichler @tjhei


tests/base/consensus_algorithm_01.cc
```diff
@@ -16,7 +16,7 @@
 
 // Test AnonymousConsensusAlgorithmProcess.
 
-#include <deal.II/base/mpi_consensus_algorithm.h>
+#include <deal.II/base/mpi_consensus_algorithm.templates.h>
```
kronbichler: Do you need the `.templates.h` here?

peterrum: Since #9604 has been merged not; but generally if one want to use `AnonymousConsensusAlgorithmProcess` on arbitrary types this file also has to be included.

peterrum: I have rebased this branch, instantiate the classes for `unsiged int` here:

https://github.com/dealii/dealii/blob/b831c52623a4c2c5d5061211b29500566e74b661/source/base/mpi_consensus_algorithm.cc#L25-L27

and have reverted the include here.

source/base/mpi_consensus_algorithm.cc
```diff
@@ -0,0 +1,41 @@
...
+{
+  namespace MPI
+  {
+    template class ConsensusAlgorithmProcess<unsigned int, unsigned int>;
+
+    template class ConsensusAlgorithmSelector<unsigned int, unsigned int>;
+
+
+    template class ConsensusAlgorithmProcess<
+      std::pair<types::global_dof_index, types::global_dof_index>,
```
kronbichler: I guess it's best to use 
```suggestion
      std::pair<std::uint64_t, std::uint64_t>,
```
to make 32-bit compilations happy and avoid an additional `#ifdef` here.

tjhei: I am not sure I understand, shouldn't global_dof_index be the right thing?

peterrum: My guess the problems are these lines (in an other file):

https://github.com/dealii/dealii/blob/e47f572be1a8b6c46db5ef33dcdc4e1c2f526ed1/source/base/mpi.cc#L1048-L1050

which I have removed in the last push.

kronbichler: The problem is that `types::global_dof_index` is `unsigned int` on non-64-bit compilations, so we have duplicate instantiations with the one a few lines above. But I agree that we should be explicit on `types::global_dof_index` and rather put an `#ifdef DEAL_II_WITH_64BIT_INDICES` around the `unsinged int` instantiations.

kronbichler: Ah sorry, that's right. We have a pair here anyway so there is no clash with the instantiation above.

bangerth:
If you want to use the class also for generic data types, just put it all into the `.h` file and let's nor worry about the instantiations. You already have it in a separate header file, and I suspect that that header file will only ever be `#include`d into `.cc` files -- i.e., it doesn't add to the kind of stuff that we seem to recursively `#include` into everything.

peterrum:
CI seems to have failed but I don't see any report...

tjhei:
yes, server hiccup.

/rebuild

peterrum:
@kronbichler Thanks. Then I can start renaming the classes!

PR #9601, opened 2020-03-03
Augment the introduction of step-8.

bangerth:
We should make the connection to the Laplace equation a bit clearer, and also
introduce some of the notation we use there.

Follow-up to #9557. 

@dangars -- FYI.

### DIFF
--- a/examples/step-8/doc/intro.dox
+++ b/examples/step-8/doc/intro.dox
@@ -26,16 +26,16 @@ as
 @f[
   -
   \text{div}\,
-  (C \nabla \mathbf{u})
+  ({\mathbf C} \nabla \mathbf{u})
   =
   \mathbf f,
 @f]
 where $\mathbf u$ is the vector-valued displacement at each point,
-$\mathbf f$ the force, and $C$ is a rank-4 tensor (i.e., it has four
+$\mathbf f$ the force, and ${\mathbf C}$ is a rank-4 tensor (i.e., it has four
 indices) that encodes the stress-strain relationship -- in essence,
 it represents the 
 <a href="https://en.wikipedia.org/wiki/Hooke%27s_law">"spring constant"</a> in
-Hookes law that relates the displacement to the forces. $C$ will, in many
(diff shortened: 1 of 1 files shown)


examples/step-8/doc/intro.dox
```diff
@@ -12,26 +12,69 @@ understood as well.
...
+indices) that encodes the stress-strain relationship -- in essence,
+it represents the 
+<a href="https://en.wikipedia.org/wiki/Hooke%27s_law">"spring constant"</a> in
+Hookes law that relates the displacement to the forces. $C$ will, in many
+cases, depend on $\mathbf x$ if the body whose deformation we want to
+simulate is composed of different materials.
+
+While the form of the equations above is correct, it is not the way
+they are usually derived. In truth, the gradient of the displacement
+$\nabla\mathbf u$ (a matrix) has no physical reality whereas its
```
tjhei:
```suggestion
$\nabla\mathbf u$ (a matrix) has no physical meaning whereas its
```
maybe?

examples/step-8/doc/intro.dox
```diff
@@ -12,26 +12,69 @@ understood as well.
...
+@f[
+  -
+  \text{div}\,
+  (C \varepsilon(\mathbf u))
+  =
+  \mathbf f,
+@f]
+which you can think of as the more natural generalization of the Laplace
+equation to vector-valued problems. (The form shown first is equivalent to
+this form because the tensor $C$ has certain symmetries, namely that
```
dangars:
```suggestion
this form because the tensor $\mathsf{C}$ has certain symmetries, namely that
```
I'm not sure what is the right convention for fourth rank tensors when you don't write the indices. Wikipedia uses $\mathsf{C}$
https://en.wikipedia.org/wiki/Linear_elasticity
I've also seen in some papers $\mathbf{C}$
You can apply this modification in the rest of the patch.

bangerth:
@tjhei and @dangars -- thanks for the suggestions. I've made these changes.

PR #9602, opened 2020-03-03
Fix bug in FEEvaluation for constraints in face integrals

kronbichler:
The bug was "just" mixing a `||` with a `&&`, but we ran into trouble nonetheless.

FYI @nfehn

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -3899,7 +3899,7 @@ FEEvaluationBase<dim, n_components_, Number, is_face, VectorizedArrayType>::
           Assert(cells[v] < dof_info->row_starts.size() - 1,
                  ExcInternalError());
           has_constraints =
-            has_constraints &&
+            has_constraints ||
             dof_info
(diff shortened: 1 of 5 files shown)


include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -3899,7 +3899,7 @@ FEEvaluationBase<dim, n_components_, Number, is_face, VectorizedArrayType>::
           Assert(cells[v] < dof_info->row_starts.size() - 1,
                  ExcInternalError());
           has_constraints =
-            has_constraints &&
+            has_constraints ||
```
tjhei: This is in a middle of a large block of code without a single comment. Would you mind at least explaining what this check does? Maybe it is worth extracting ``dof_info->...[...]`` as a new variable as it is used twice? This would also help with understanding.

kronbichler: This is a good point. The variable is used 4 times, so this saves some code - my last commit reduces the code by 20 lines despite bringing in some comments of what happens here.

PR #9603, opened 2020-03-03
Convert explicit loops over faces to range-based loops.

bangerth:
This is a follow-up to #9415, #9467, #9476, and 9477. 

I did this by:
* Telling clang-format to use a line length limit of 800 
* Converting these lines using a perl script
* Converting everything back to line lengths of 80
* Going around a few more times because I'm making mistakes in the process.
It might be the first thing I found clang-format to actually be useful for ;-)

/rebuild


bangerth:
By the way, this passed all of my local tests (which are more than the regression tester checks, I believe).

PR #9610, opened 2020-03-04
Add chang-log entry for ConsensusAlgorithm

peterrum:
I think `Utilities::MPI::ConsensusAlgorithm` deserves a change-log entry of its own. On the one hand it is "only" general-purpose (MPI) infrastructure and not limited to FEM (and with high probability user most users will never use it directly), but it has become a key ingredient for large scale simulations. 

ping @kronbichler

### DIFF
--- a/doc/news/changes/major/20190603PeterMunch
+++ b/doc/news/changes/major/20190603PeterMunch
@@ -1,0 +1,5 @@
+New: The namespace Utilities::MPI::ConsensusAlgorithms has been added. It 
+provides efficient implementations for communication patterns (PEX and NBX) to 
+retrieve data from other processes in a dynamic-sparse way.
+<br>
+(Peter Munch, Martin Kronbichler, 2019/06/03)


doc/news/changes/major/20190603PeterMunch
```diff
@@ -0,0 +1,7 @@
+New: The interface Utilities::MPI::ConsensusAlgorithm has been added. It is 
+implemented by Utilities::MPI::ConsensusAlgorithm_NBX and
```
kronbichler: I realize now that the combination of Pascal/camel case and underscore - I feel that I would prefer `ConsensusAlgorithmNBX`. Apart from the finite element classes like `FE_DGQ` where it makes sense because FE is already an abbreviation that we want to separate and some associated polynomials, the combination of underscores with capital letters is not very common in deal.II. As we have not released that class yet, we could still change it - any opinions on this?

peterrum: I am fine with that. It is mainly an internal change. I am normally using them directly, but use `ConsensusAlgorithmSelector` (which does already have the right name format).

peterrum: I will do the renaming. But lets merge #9600 first.

masterleinad:
Should this be in `major`?

kronbichler:
> Should this be in major?

I think others should decide here because I was involved in some parts of it. That said, the algorithm is a substantial progress for setup routines in many large-scale problems and the product of many days (or weeks) of work.

bangerth:
If we still can, let's remove the underscore.

bangerth:
And yes, `major` seems like the right category.

bangerth:
By the way, if you wanted to think about different names anyway, how about putting the various implementations into a namespace `ConsensusAlgorithms`? Then you'd only have to call them `NBX`, ...

peterrum:
> And yes, major seems like the right category.

Well it created major headaches!

kronbichler:
Ask @tjhei for headaches and integer overflows. :grin:

tjhei:
Sigh. I spent way too much time over the last 4 months fixing hard to find bugs. :-(

doc/news/changes/major/20190603PeterMunch
kronbichler: Please rebase from master and update the name now that #9600 is done.

peterrum:
Together with #9623, this PR can be merged.

PR #9611, opened 2020-03-04
[URGENT] Fix indentation in step-8

peterrum:
One file was not formatted correctly in  d773343dbb85ea2ee796f2628fc7863315b4d68c, which prevents that other PRs that are based on master pass the test.

This issue raises the question if merging PRs directly, even if they are only extension of the documentation, are allowed...

FYI @bangerth

### DIFF
--- a/examples/step-8/doc/intro.dox
+++ b/examples/step-8/doc/intro.dox
@@ -12,13 +12,13 @@ understood as well.
 
 @dealiiVideoLecture{19}
 
-In this tutorial program we will want to solve the 
+In this tutorial program we will want to solve the
 <a href="https://en.wikipedia.org/wiki/Linear_elasticity">elastic equations</a>.
 They are an extension to Laplace's equation with a vector-valued solution that
 describes the displacement in each space direction of a rigid body
 which is subject to a force. Of course, the force is also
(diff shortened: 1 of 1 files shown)


kronbichler:
> This issue raises the question if merging PRs directly, even if they are only extension of the documentation, are allowed...

We should be able to leave the decision whether to fire up all expensive tests (that block resources from other threads) to the discretion of the reviewer - no need to over-formalize the process just because mistakes happen sometimes. However, I definitely agree that we should make the script that checks the indentation/trailing whitespace that caused this failure obligatory and run without the "rebuild/ready-to-test" signals. @tjhei can you comment?

kronbichler:
Let me merge this now because it is useful to have a clean master.

peterrum:
Thanks!

tjhei:
Sorry for merging something that broke master and thanks for the quick fix, @peterrum !

> However, I definitely agree that we should make the script that checks the indentation/trailing whitespace that caused this failure obligatory

I am confused why this was not detected by travis: https://travis-ci.org/dealii/dealii/builds/658019033?utm_source=github_status&utm_medium=notification

We need to address that for sure.

bangerth:
Ah, apologies :-(

PR #9616, opened 2020-03-04
More active_cell_iterators().

bangerth:


kronbichler:
The failing tester that seems stuck will not run the test files changed here, so I will merge.

PR #9624, opened 2020-03-05
Introduce and use new ExcNeedsMPI

peterrum:
Related to PR #9387.

I would like to assert here that deal.II was compiled with or without MPI:

https://github.com/dealii/dealii/blob/e78560b2099e1900a9cec668491c6ce8535f4873/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h#L236-L239

https://github.com/dealii/dealii/blob/e78560b2099e1900a9cec668491c6ce8535f4873/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h#L290-L292

However, I did not find an appropriate exception. For example, in `tria_base.cc` a long text is written two times. But I guess there are lot of places where it is checked if deal.II is compiled with or without MPI.

Now we can write `Assert(false, ExcNeedsMPI());`.

### DIFF
--- a/include/deal.II/base/exceptions.h
+++ b/include/deal.II/base/exceptions.h
@@ -1087,6 +1087,14 @@ namespace StandardExceptions
     "find a valid LAPACK library.");
 
   /**
+   * This function requires support for the MPI library.
+   */
+  DeclExceptionMsg(
+    ExcNeedsMPI,
+    "You are attempting to use functionality that is only available "
+    "if deal.II was configured to use MPI.");
+
+  /**
    * This function requires support for the NetCDF library.
    *
    * @deprecated Support for NetCDF in deal.II is deprecated.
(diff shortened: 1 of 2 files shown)


peterrum:
@kronbichler Could we wait with merging #9387 until this PR is merged?

include/deal.II/base/exceptions.h
```diff
@@ -1086,6 +1086,15 @@ namespace StandardExceptions
...
     "find a valid LAPACK library.");
 
+  /**
+   * This function requires support for the MPI library.
+   */
+  DeclExceptionMsg(
+    ExcNeedsMPI,
+    "You are attempting to use functionality that is only available "
+    "if deal.II was configured to use MPI, but cmake did not "
+    "find a valid MPI library.");
```
tjhei: The user needs to explicitly enable DEAL_II_WITH_MPI, so the second part of this sentence doesn't make much sense to me.

peterrum: OK. I have remove that part of the sentence.

PR #9629, opened 2020-03-05
Fix a typo.

bangerth:
Follow-up to #9580.


Rombur:
you need to run the indentation script

bangerth:
Reindented. It didn't occur to me that removing a single character would require that... :-)

PR #9631, opened 2020-03-06
new free function LinearAlgebra::set_zero_mean_value()

nfehn:
Setting the mean value of a vector to zero is a standard operation, e.g., in incompressible flow solvers. This PR provides a short-cut for this type of operation.

### DIFF
--- a/doc/news/changes/minor/20200306NiklasFehn
+++ b/doc/news/changes/minor/20200306NiklasFehn
@@ -1,0 +1,7 @@
+New: LinearAlgebra::set_zero_mean_value()
+Add free function that allows to shift a vector by a constant value in a
+way that its mean value becomes zero. This function only makes use of the
+pure virtual functions add() and mean_value() of the abstract class
+VectorSpaceVector.
+<br>
+(Niklas Fehn, 2020/03/06)
--- a/include/deal.II/lac/vector_space_vector.h
+++ b/include/deal.II/lac/vector_space_vector.h
@@ -273,6 +273,22 @@ namespace LinearAlgebra
   /*@}*/
 } // namespace LinearAlgebra
 
+// ---------------------------- Free functions --------------------------
+
+namespace LinearAlgebra
+{
+  /**
+   * Shift all entries of the vector by a constant factor so that the mean
+   * value of the vector becomes zero.
+   */
+  template <typename Number>
+  void
+  set_zero_mean_value(VectorSpaceVector<Number> &vector)
+  {
+    vector.add(-vector.mean_value());
+  }
+} // namespace LinearAlgebra
+
 DEAL_II_NAMESPACE_CLOSE
 
 #endif


kronbichler:
I would wait for @Rombur's opinion on this, but I want to mention that we tried to keep the interface in the `VectorSpaceVector` small, only adding essential operations without too much redundancy. Given that the implementation here is merely `vec.add(-vec.mean_value());` I see little need to extend the interface. Other vector packages like [the Tpetra MultiVector](https://docs.trilinos.org/dev/packages/tpetra/doc/html/classTpetra_1_1MultiVector.html) also do not add this to the vector interface, despite having a pretty rich set functions.

At the same time, I would see a lot of value in providing a collection of common linear algebra operations (non-member functions) where this would be one of them.

For pure Neumann problems (like incompressible flows), one should also have in mind that this is an algebraic mean value and one of many arbitrary ways to fix the constant. For residuals in finite element methods with shape functions that form a partition of unity (like FE_Q, FE_DGQ, ...), this is also an integral mean value, but already for the solution it is not. For elements like FE_DGP (with a modal basis) it should work to constrain the constant this way, but the result would in general not resemble a zero-mean function.

nfehn:
I see the point. However, looking at the interface class from this perspective I wonder why it is not even leaner. At first sight I am tempted to say that the interface does not seem to provide a minimal set of operations, but I have to admit that I did not go through all the implementations.

There is another aspect that I want to mention and that should be taken into account. @kronbichler You are arguing with the fact that the functionality can be implemented as a one-liner. Nevertheless, writing ``vec.add(-vec.mean_value())`` already requires some form of documentation or explanation in my opinion, while ``set_zero_mean_value()`` is self-documenting.

bangerth:
Isn't there already a function in `VectorTools` that does this, using the "correct" way of computing the mean as an integral rather than the algebraic mean of the vector components?

nfehn:
I think this would be quite expensive, I only want to apply these functions once in the setup but not in every time step. The application with incompressible flow should not be the argument for or against, maybe I should not have mentioned this at all.

nfehn:
To give some background, the origin in incompressible flows is not only postprocessing the pressure field or "an arbitrary way to fix the constant" as written above by @kronbichler, but is also of "algebraic" nature; the right-hand side of the pressure Poisson equation has to be compatible with the nullspace of the discrete pressure Poisson operator (of course with the restrictions on the chosen function spaces as explained by @kronbichler, i.e. does a LinearAlgebra vector with constant entries span the nullspace? -> if yes, you want to remove the nullspace and set the mean value to zero).

Rombur:
> I see the point. However, looking at the interface class from this perspective I wonder why it is not even leaner. 

So what I did is I went through the common interface between all of our vectors and then I kept what was being used inside the library. These are the functions we need if we don't want to rewrite the algorithms in deal.II.
I agree that your function is a useful addition but I would prefer not having this in the interface of `VectorSpaceVector`. I like @kronbichler's idea of having a collection of LinearAlgebra (non-member) functions.

doc/news/changes/minor/20200306NiklasFehn
```diff
@@ -0,0 +1,7 @@
+New: LinearAlgebra::set_zero_mean_value()
+Add free function that allows to shift a vector by a constant value in a
+way that its mean value becomes zero. This function only makes use of the
+pure virtual functions add() and mean_value() of the interface class
```
peterrum:
```cpp
pure virtual functions add() and mean_value() of the abstract class
```

nfehn:
One more question came up when I looked at the code. I have seen that the free function ``swap()`` is implemented several times with the only difference being the vector type. I think it could make sense to implement it only once, even if it is a one-liner. Is there a reason why ``swap()`` is not a member function of ``VectorSpaceVector``?

doc/news/changes/minor/20200306NiklasFehn
```diff
@@ -1,7 +1,7 @@
 New: LinearAlgebra::set_zero_mean_value()
 Add free function that allows to shift a vector by a constant value in a
 way that its mean value becomes zero. This function only makes use of the
-pure virtual functions add() and mean_value() of the interface class
+pure virtual functions add() and mean_value() of the astract class
```
masterleinad:
```suggestion
pure virtual functions add() and mean_value() of the abstract class
```

nfehn:
I squashed the commits into a single one.

doc/news/changes/minor/20200306NiklasFehn
nfehn: has been fixed

kronbichler:
Any additional comments?

PR #9633, opened 2020-03-07
Add default constructor to NoncontiguousPartitioner

peterrum:
When integrating the new `NoncontiguousPartitioner` into an application code, I was missing a default constructor.

### DIFF
--- a/include/deal.II/base/mpi_noncontiguous_partitioner.h
+++ b/include/deal.II/base/mpi_noncontiguous_partitioner.h
@@ -44,6 +44,12 @@ namespace Utilities
     {
     public:
       /**
+       * Default constructor. Requires calling one of the reinit() functions
+       * to create a valid object.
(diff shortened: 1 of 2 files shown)


include/deal.II/base/mpi_noncontiguous_partitioner.h
```diff
@@ -43,6 +43,12 @@ namespace Utilities
       : public LinearAlgebra::CommunicationPatternBase
     {
     public:
+      /**
+       * Default constructor.Please call one of the reinit() functions
```
masterleinad:
```suggestion
       * Default constructor. Requires calling one of the reinit() functions to create a valid object. 
```

peterrum: Thanks! Sounds better!

PR #9636, opened 2020-03-07
Step-69: Introduce asynchronous IO mechanism

tamiko:
This PR introduces a simple asynchronous IO mechanism to step-69:

On every MPI rank a (task group independent) background thread is spawned
for handling IO and writing out the solution and other files to disk.

The main advantage of this approach is that the normal computation can
continue and doesn't have to wait for disk IO. As an example, on a recently
computed 2B gridpoint computation the total time for writeback to the
distributed filesystem was about 20% of the total time of the computation
itself.

### DIFF
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -2740,8 +2740,8 @@ namespace Step69
 
   // @sect5{Output and checkpointing}
   //
-  // Writing out the final vtk files is a quite IO intensive task that can
-  // stall the main loop quite a bit. In order to avoid this we use an <a
+  // Writing out the final vtk files is quite an IO intensive task that can
+  // stall the main loop for a while. In order to avoid this we use an <a
   // href="https://en.wikipedia.org/wiki/Asynchronous_I/O">asynchronous
   // IO</a> strategy by creating a background thread that will perform IO
   // while the main loop is allowed to continue. In order for this to work
@@ -2766,14 +2766,9 @@ namespace Step69
     // If the asynchronous writeback option is set we launch a background
     // thread performing all the slow IO to disc. In that case we have to
     // make sure that the background thread actually finished running. If
-    // not, we have to wait to for it to finish because we would otherwise
-    // overwrite <code>output_vector</code>, rerun the
-    // <code>schlieren_postprocessor</code> and
-    // DataOut<dim>::build_patches() prematurely before the output of the
-    // previous output cycle has been fully written out to disk.
-    //
-    // We launch said background thread with <a
-    // href="https://en.cppreference.com/w/cpp/thread/async"><code>std::async</code></a>
+    // not, we have to wait to for it to finish. We launch said background
+    // thread with <a
+    // href="https://en.cppreference.com/w/cpp/thread/async"><code>std::async()</code></a>
     // that returns a <a
     // href="https://en.cppreference.com/w/cpp/thread/future"><code>std::future</code></a>
     // object. This <code>std::future</code> object contains the return
@@ -2827,7 +2822,6 @@ namespace Step69
     // the <code>this</code> pointer as well as most of the arguments of
     // the output function by value so that we have access to them inside
     // the lambda function.
-
     const auto output_worker = [this, name, t, cycle, checkpoint, data_out]() {
       if (checkpoint)
         {
@@ -2869,7 +2863,6 @@ namespace Step69
     // At this point we can return from the <code>output()</code> function
     // and resume with the time stepping in the main loop - the thread will
     // run in the background.
-
     if (!asynchronous_writeback)
       {
         background_thread_state = std::async(std::launch::async, output_worker);
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -495,7 +495,8 @@ namespace Step69
     std::string base_name;
     double      t_final;
     double      output_granularity;
-    bool        enable_compute_error;
+
+    bool asynchronous_writeback;
 
     bool resume;
 
@@ -2506,6 +2507,11 @@ namespace Step69
                   output_granularity,
                   "time interval for output");
 
+    asynchronous_writeback = true;
+    add_parameter("asynchronous writeback",
+                  asynchronous_writeback,
+                  "Write out solution in a background thread performing IO");
+
     resume = false;
     add_parameter("resume", resume, "Resume an interrupted computation.");
   }
@@ -2847,7 +2853,14 @@ namespace Step69
     // time stepping in the main loop - the thread will run in the
     // background.
 
-    output_thread = std::move(std::thread(output_worker));
+    if (!asynchronous_writeback)
+      {
+        output_thread = std::move(std::thread(output_worker));
+      }
+    else
+      {
+        output_worker();
+      }
   }
 
 } // namespace Step69
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -82,9 +82,10 @@
 #include <boost/range/irange.hpp>
 #include <boost/range/iterator_range.hpp>
 
-// For std::isnan, std::isinf, and std::ifstream
+// For std::isnan, std::isinf, std::ifstream, std::async, and std::future
 #include <cmath>
 #include <fstream>
+#include <future>
 
 // @sect3{Class template declarations}
 //
@@ -505,9 +506,11 @@ namespace Step69
     InitialValues<dim>          initial_values;
     TimeStepping<dim>           time_stepping;
     SchlierenPostprocessor<dim> schlieren_postprocessor;
+    DataOut<dim>                data_out;
 
-    std::thread output_thread;
     vector_type output_vector;
+
+    std::future<void> background_thread_state;
   };
 
   // @sect3{Implementation}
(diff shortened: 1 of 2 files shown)


kronbichler:
> The main advantage of this approach is that the normal computation can
continue and doesn't have to wait for disk IO. As an example, on a recently
computed 2B gridpoint computation the total time for writeback to the
distributed filesystem was about 20% of the total time of the computation
itself.

Without having carefully timed things, I would not unconditionally subscribe to this being an advantage. On some of my systems, the background thread did slow down the overall computation when I had computations on all CPU threads - the point is that you oversubscribe the system. Even though you would assume that the IO thread would not need to get awake often, you still swap in and out the worker thread on a regular basis, destroying caches, prefetching, and all of the other sugar in a CPU.
I am not saying that asynchronous IO could not have improved throughput on the 1.8B computation (especially since I used hyperthreading to improve the compute part by 20-25%, so all IO thread is oversubscription of the cores), but I just want to raise a warning flag that we have not enough measurements to back things up. My initial experiments indicate that it is likely not a feature to enable unconditionally if you care about the best possible performance.

tamiko:
> On some of my systems, the background thread did slow down the overall computation when I had computations on all CPU threads - the point is that you oversubscribe the system.

Did you test only the computational part, or did you also have simultaneous IO?

I agree and revise my comment. I had a closer look at the output logs.

Completely ignoring IO for a minute, I guess we are actually doing quite a bit of computing (including gzip compression) in the `DataOut<dim>::build_patches` and `DataOut<dim>::write_vtu_*`. I thinkg the big difference is that I did typically ran with about 4, 8, 16 worker threads per MPI rank whereas you did restrict your runs to no-threading. I saw quite a bit of improvement in that case (but I will benchmark again).

Now looking at IO: I do run my computations on a very slow NFS. Most of the background threads are in uninterruptible sleep state "D" because they are waiting for IO. I think at tleast this part should be handled by a background thread. But maybe we want to have another approach for that? (I.e. running `build_patches` synchronously and only handing IO over to the background thread)

bangerth:
It's ok with me, but I'll let the discussion above play out.

In the review of step-69, I had suggested to use a background task, rather than thread. @kronbichler -- would that alleviate your concerns? It would not lead to competition between threads, but instead use otherwise unused CPU resources.

kronbichler:
> Did you test only the computational part, or did you also have simultaneous IO?

I also had the IO part - but it is hard to say how much each of those contributed to.

> I think at least this part should be handled by a background thread. But maybe we want to have another approach for that? (I.e. running build_patches synchronously and only handing IO over to the background thread)

I agree, we do indeed a fair bit of computation and that could create some of the noise (leading to a >5x slowdown as long as computation goes on from both threads). I also agree that things are more acute in my setup where I use separate MPI ranks for each core or even for each hyperthread. Those give best performance in the compute part, but things might obviously be a bit different once we have output threads. So yes, running everything except for the actual write synchronously and only the real IO would look like the ideal solution to test. The problem is that `DataOut::write_vtu_in_parallel` or `write_vtu` still run the gzip compression and a few other format-specific things, before finally writing out things to the file system. Not sure how to best approach it. In any case, my comment was not that this will never be worth it, but that I see cases where it is likely not.

> In the review of step-69, I had suggested to use a background task, rather than thread. @kronbichler -- would that alleviate your concerns? It would not lead to competition between threads, but instead use otherwise unused CPU resources.

By background task you mean a handle to TBB task that we keep around and try to join once it is ready? I could see some value in it for the case with many threads because TBB would allocate it to one of them and not oversubscribe. However, I think this then partly defeats the goal of the slow NFS system observed by @tamiko because you really want a thread that can be suspended by the OS if there is no progress due to IO - a task cannot be suspended in the same way. Of course, all other x-1 threads would continue to make progress and simply use more of the CPU resources if available (like shared caches or memory buses).

I guess we should let @tamiko decide what to use - I am definitely not opposed to the setup proposed right now because it solves some problems, as long as we do not sell it as "always use a background thread for IO" to our users because I have seen cases where that degrades performance.

tjhei:
> as long as we do not sell it as "always use a background thread for IO" to our users because I have seen cases where that degrades performance.

I am also skeptical that this is the right choice without measurements. A good parallel filesystem can take tens of GB/s. If you slow down even a single MPI rank by having an extra thread kick in and interrupt an algorithm that contains MPI communication, latency goes up quite a bit slowing everything down.

bangerth:
I'll add that `build_patches()` is already parallel and scales well to large numbers of processor cores. So there is no value in putting that onto a background thread, unlike the actual I/O. I would suspect that the gzip operation is actually *extremely* fast compared to `build_patches()` and the write operations -- I would not worry about it.

drwells:
It sounds to me like we should implement async io as an option in the input file - users could learn a lot by reading this discussion and trying it out themselves.

bangerth:
So what do we do here? It may be worthwhile remembering that tutorials are a place to *teach*. We can state that we show this here because it is a technique of general usefulness -- but then also state that the approach has not been timed and that for "real" HPC, one would have to establish that it's useful. Some version of the discussion above could very well be placed into a section in the introduction.

tamiko:
@bangerth @drwells I will modify this PR to make the asynchronous scheduling optional (that's really easys - instead of creating a thread you simply call the function. We actually have that in our research code already :-)).

I will probably also modify the mechanism to do the `build_patches` synchronously and only do the IO write-out in a background thread.

All of that should be a minor modification. I'll ping you guys again once I am done (hopefully in the next week).

bangerth:
Sounds good -- if you want to switch, it sounds like you want to use `std::packaged_task` :-)

tamiko:
@dealii/dealii All, please have a look at this PR again. I have made the following changes:
 - introduced a configuration option to toggle background IO
 - moved the `build_patches` call out of the background thread
 - changed the thread handling to `std::async` and `std::future::valid/wait`.

examples/step-69/step-69.cc
```diff
@@ -2562,28 +2574,43 @@ namespace Step69
...
+        ProblemDescription<dim>::problem_dimension;
+      const auto &component_names = ProblemDescription<dim>::component_names;
 
-    print_head(pcout, "set up time step");
-    time_stepping.prepare();
-    schlieren_postprocessor.prepare();
+      for (unsigned int i = 0; i < problem_dimension; ++i)
+        {
+          output_vector[i].reinit(offline_data.partitioner);
+          data_out.add_data_vector(output_vector[i], component_names[i]);
```
bangerth, line 2608: I read the first line as just resizing `output_vector[i]`. I guess it also copies, but that wasn't clear to me. Can we make that more obvious? (If necessary through a comment.)

examples/step-69/step-69.cc
```diff
@@ -2675,6 +2702,13 @@ namespace Step69
           }
       }
 
+    // We wait for any remaining background output thread to finish before
+    // printing a summary and exiting.
+    if (background_thread_state.valid())
+      {
+        background_thread_state.wait();
+      }
```
bangerth, line 2710: just drop the curly braces, for consistency in other places

examples/step-69/step-69.cc
```diff
@@ -2723,6 +2757,19 @@ namespace Step69
   }
 
   // @sect5{Output and checkpointing}
+  //
+  // Writing out the final vtk files is a quite IO intensive task that can
```
bangerth, line 2761:
```suggestion
  // Writing out the final vtk files is quite an IO intensive task that can
```

examples/step-69/step-69.cc
```diff
@@ -2723,6 +2757,19 @@ namespace Step69
...
+  // stall the main loop quite a bit. In order to avoid this we use an <a
```
bangerth, line 2762:
```suggestion
  // stall the main loop for a while. In order to avoid this we use an <a
```

examples/step-69/step-69.cc
bangerth: Wait, I must have misunderstood this. Are you saying that you're pro-actively attach these vectors to the `DataOut` object, and only later fill the vector? That might work, but I'm not sure that was ever the intention of the class to work that way. I've always thought that it could also copy these vectors when they are handed to the class.

examples/step-69/step-69.cc
```diff
@@ -2734,57 +2781,101 @@ namespace Step69
...
           << ", checkpoint = " << checkpoint << ")" << std::endl;
 
-    TimerOutput::Scope scope(computing_timer, "main_loop - output");
-
-    if (checkpoint)
+    // If the asynchronous writeback option is set we launch a background
+    // thread performing all the slow IO to disc. In that case we have to
+    // make sure that the background thread actually finished running. If
+    // not, we have to wait to for it to finish because we would otherwise
+    // overwrite <code>output_vector</code>, rerun the
```
bangerth, line 2788: just drop "because we would otherwise...". It disrupts the flow of the rest of the sentence as it's no longer clear what ", rerun" corresponds to: what we need to do, or what would otherwise happen.

(It's also the case that if the IO hasn't finished yet and you start another IO, you might eventually end up with *a lot* of threads still doing IO.)

examples/step-69/step-69.cc
```diff
@@ -2734,57 +2781,101 @@ namespace Step69
...
+    // <code>schlieren_postprocessor</code> and
+    // DataOut<dim>::build_patches() prematurely before the output of the
+    // previous output cycle has been fully written out to disk.
+    //
+    // We launch said background thread with <a
+    // href="https://en.cppreference.com/w/cpp/thread/async"><code>std::async</code></a>
```
bangerth, line 2794:
```suggestion
    // href="https://en.cppreference.com/w/cpp/thread/async"><code>std::async()</code></a>
```

examples/step-69/step-69.cc
```diff
@@ -2734,57 +2781,101 @@ namespace Step69
...
+        output_vector[i] = U[i];
+        output_vector[i].update_ghost_values();
+      }
 
-    data_out.add_data_vector(schlieren_postprocessor.schlieren,
-                             "schlieren_plot");
+    schlieren_postprocessor.compute_schlieren(output_vector);
 
     data_out.build_patches(discretization.mapping,
                            discretization.finite_element.degree - 1);
```
bangerth, line 2822: Apparently this clears all previously computed patches. That, too, wasn't clear to me. 

My preference would have been if you create a `std::shared_ptr<DataOut>` at this place that you pass to the background thread. Once that thread is done, it automatically deletes the pointer. This also obviates the need for attaching vectors pro-actively.

examples/step-69/step-69.cc
```diff
@@ -2734,57 +2781,101 @@ namespace Step69
...
+    // background thread with the help of
+    // <a
+    // href="https://en.cppreference.com/w/cpp/thread/async"><code>std::async</code></a>
+    // function. The function returns a <a
+    // href="https://en.cppreference.com/w/cpp/thread/future"><code>std::future</code></a>
+    // object that we can use to query the status of the background thread.
+    // At this point we can return from the <code>output()</code> function
+    // and resume with the time stepping in the main loop - the thread will
+    // run in the background.
+
```
bangerth, line 2870:
```suggestion
```

examples/step-69/step-69.cc
```diff
@@ -2734,57 +2781,101 @@ namespace Step69
...
-                                cycle,
-                                true,
-                                DataOutBase::VtkFlags::best_speed);
-    data_out.set_flags(flags);
+    // Next we create a lambda function for the background thread. We <a
+    // href="https://en.cppreference.com/w/cpp/language/lambda">capture</a>
+    // the <code>this</code> pointer as well as most of the arguments of
+    // the output function by value so that we have access to them inside
+    // the lambda function.
+
```
bangerth, line 2829:
```suggestion
```

tamiko:
@bangerth I have switched to a `shared_pointer` and addressed your comments regarding documentation.

masterleinad:
`Travis` hasn't reported back here as well. So we either need to rebase here to retrigger `Travis` or remove `Travis` as required check. :wink:

tamiko:
@kronbichler What about I put it on the todo list for a final round of documentation polishing for step-69?

tamiko:
@masterleinad I rebase a second time. Let's see whether this makes travis happy again. :confused:

PR #9639, opened 2020-03-08
Enable Utilities::MPI::NoncontiguousPartitioner to handle padding

peterrum:
This PR enables `Utilities::MPI::NoncontiguousPartitioner` to handle padding in the source and destination vectors. The user can indicate that the purpose of a vector entry is only for padding by inserting `numbers::invalid_dof_index` into `std::vector<types::global_dof_index> &indices_has` and/or `std::vector<types::global_dof_index> &indices_want`.

I still need to write an appropriate test.
 
FYI @nfehn @kronbichler

### DIFF
--- a/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
+++ b/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
@@ -183,26 +183,49 @@ namespace Utilities
       const std::vector<types::global_dof_index> &indices_want,
       const MPI_Comm &                            communicator)
     {
+      // step 0) clean vectors from numbers::invalid_dof_index (indicating
+      //         padding)
+      std::vector<types::global_dof_index> indices_has_clean;
+      indices_has_clean.reserve(indices_has.size());
+
+      for (const auto &i : indices_has)
+        if (i != numbers::invalid_dof_index)
+          indices_has_clean.push_back(i);
+
+      std::vector<types::global_dof_index> indices_want_clean;
+      indices_want_clean.reserve(indices_has.size());
+
+      for (const auto &i : indices_want)
+        if (i != numbers::invalid_dof_index)
+          indices_want_clean.push_back(i);
+
       // step 0) determine "number of degrees of freedom" needed for IndexSet
-      const types::global_dof_index n_dofs = Utilities::MPI::max(
-        std::max(
-          [&indices_has]() {
-            const auto it = max_element(indices_has.begin(), indices_has.end());
-            return it == indices_has.end() ? 0 : (*it + 1);
-          }(),
-          [&indices_want]() {
-            const auto it =
-              max_element(indices_want.begin(), indices_want.end());
-            return it == indices_want.end() ? 0 : (*it + 1);
-          }()),
-        communicator);
+      const types::global_dof_index local_n_dofs_has =
+        indices_has_clean.empty() ?
+          0 :
+          (*std::max_element(indices_has_clean.begin(),
+                             indices_has_clean.end()) +
+           1);
+
+      const types::global_dof_index local_n_dofs_want =
+        indices_want_clean.empty() ?
+          0 :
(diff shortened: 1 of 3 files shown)


include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -183,26 +183,45 @@ namespace Utilities
...
+
       // step 0) determine "number of degrees of freedom" needed for IndexSet
       const types::global_dof_index n_dofs = Utilities::MPI::max(
         std::max(
-          [&indices_has]() {
-            const auto it = max_element(indices_has.begin(), indices_has.end());
-            return it == indices_has.end() ? 0 : (*it + 1);
+          [&indices_has_clean]() {
+            const auto it =
+              max_element(indices_has_clean.begin(), indices_has_clean.end());
```
bangerth: You'll want to use the following -- the current form may not work with some compilers.
```suggestion
              std::max_element(indices_has_clean.begin(), indices_has_clean.end());
```

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -183,26 +183,45 @@ namespace Utilities
...
+            return it == indices_has_clean.end() ? 0 : (*it + 1);
           }(),
-          [&indices_want]() {
+          [&indices_want_clean]() {
             const auto it =
-              max_element(indices_want.begin(), indices_want.end());
-            return it == indices_want.end() ? 0 : (*it + 1);
+              max_element(indices_want_clean.begin(), indices_want_clean.end());
```
bangerth: same here

peterrum:
I have integrate @bangerth's suggestions, made a bug fix, and added a test. I think this PR is ready to be reviewed.

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -183,26 +183,45 @@ namespace Utilities
...
+            return it == indices_has_clean.end() ? 0 : (*it + 1);
           }(),
-          [&indices_want]() {
-            const auto it =
-              max_element(indices_want.begin(), indices_want.end());
-            return it == indices_want.end() ? 0 : (*it + 1);
+          [&indices_want_clean]() {
+            const auto it = std::max_element(indices_want_clean.begin(),
+                                             indices_want_clean.end());
+            return it == indices_want_clean.end() ? 0 : (*it + 1);
```
kronbichler: This would be easier to read if you avoided the lambdas inside `std::max` in favor of plain code here with temporary variable like `types::global_dof_index max_has = ...` and `max_wants`.

peterrum: I have made the change.

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -183,26 +183,48 @@ namespace Utilities
...
-            const auto it =
-              max_element(indices_want.begin(), indices_want.end());
-            return it == indices_want.end() ? 0 : (*it + 1);
-          }()),
-        communicator);
+      const types::global_dof_index local_n_dofs_has = [&indices_has_clean]() {
+        const auto it =
+          std::max_element(indices_has_clean.begin(), indices_has_clean.end());
+        return it == indices_has_clean.end() ? 0 : (*it + 1);
+      }();
```
kronbichler: We're almost there.  I was thinking about this code and similar for the next block below:
```suggestion
      const types::global_dof_index local_n_dofs_has = indices_has_clean.empty() ? 0 :
          *std::max_element(indices_has_clean.begin(), indices_has_clean.end());
```

peterrum: I have mad the change.

kronbichler:
Excellent, thank you!

PR #9650, opened 2020-03-11
Replace logfile in more tests

masterleinad:
None of these use `logfile` as a global variable.


masterleinad:
Locally all tests are passing but I don't have the more special options (hdf5, adolc, assimp, operncascade, sundials, symengine...) enabled.

drwells:
Lets just merge and see if any new things start to fail. This looks fine at a glance.

PR #9653, opened 2020-03-12
ECL: fix variable names and a std::function version

peterrum:
...extracting stuff from #9181.


peterrum:
Thanks @kronbichler . I will rebase #9181.

PR #9654, opened 2020-03-12
Split instantiations of matrix-free MappingInfo

kronbichler:
Now that we instantiate `MatrixFreeFunctions::MappingInfo` for several kinds of vectorized arrays, the compilation has become pretty big and takes quite some memory. This will get even slightly worse with a PR I am currently working on that should provide a much faster initialization method for high-order mappings.

This PR splits the instantiations up into 3 which gives better compile times on parallel systems (though not really the factor 3 I was hoping for because the compiler does keep track of some redundancies when instantiating multiple objects).


tjhei:
ready to merge as soon as tests pass.

PR #9656, opened 2020-03-13
Implement matrix-free evaluation of geometry for MF::MappingInfo

kronbichler:
For higher order mappings and many quadrature points in 3D, the evaluation of the geometry through `FEFaceValues` can take a lot of memory (up to 1 GB for mapping degrees around 15 and 20 quadrature points per direction). And overall, the initialization of `MappingInfo` is slow once we have deforming domains. However, we have the facilities around to do this more quickly, at least for the case the geometry is described through a `MappingQ` formula: The polynomial interpolation is nothing else than a matrix-free evaluation of the geometry data, starting from the points of the mapping.

An addition that was pretty straight-forward with the new organization of data was the fact that I can now also detect similar geometry on non-affine cells, like for an extruded mesh or if only 1 direction is deformed and the others are not. This should speed up operator evaluation because geometry data can come from cache more often.

This PR implements this functionality. To keep the PR compact, I have decided against already implementing the same tricks for the `face_data_by_cells` field because we would need some additional index extraction. Also, I have not yet parallelized the two main loops with `apply_to_subranges` because that is rather straight-forward and should be kept out of providing the functionality.

I have tested the code and it produces the correct results for all `matrix_free` tests.

TODO items
- [x] Add a test case with a 3D ball, high mapping degree and many quadrature points that would have been very expensive before but is OK now.
- [x] Add a changelog, describing this feature.


peterrum:
@kronbichler You need to indent the code. I will have a look at the code now.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -6945,39 +6945,25 @@ FEEvaluation<dim,
...
+      Point<dim, VectorizedArrayType>           point = quadrature_points[0];
+      const Tensor<2, dim, VectorizedArrayType> jac   = this->jacobian[1];
+      for (unsigned int d = 0; d < dim; ++d)
+        for (unsigned int e = 0; e < dim; ++e)
+          point[d] += jac[d][e] *
+                      static_cast<Number>(
+                        this->mapping_data->descriptor[this->active_quad_index]
+                          .quadrature.point(q)[e]);
+      return point;
     }
```
kronbichler: This change here is a simplification for getting the quadrature point: Rather than implementing a rather ad-hoc stepping along the diagonal I simply compute it from the lower left corner of the cell. This is nicer to compute and nicer to read.

include/deal.II/matrix_free/mapping_info.h
```diff
@@ -573,6 +592,11 @@ namespace internal
                      Tensor<1, VectorizedArrayType::n_array_elements, Number>>
           &t2) const;
 
+      template <int dim>
+      bool
+      operator()(const std::array<Tensor<2, dim, Number>, dim + 1> &t1,
+                 const std::array<Tensor<2, dim, Number>, dim + 1> &t2) const;
```
peterrum: Documentation?

include/deal.II/matrix_free/mapping_info.h
```diff
@@ -441,6 +441,25 @@ namespace internal
...
+       * information. This optimized approach is much faster than going
+       * through FEValues and FEFaceValues, especially when several different
+       * quadrature formulas are involved, and consumes less memory.
+       */
+      void
+      compute_mapping_q(
+        const dealii::Triangulation<dim> &                        tria,
+        const std::vector<std::pair<unsigned int, unsigned int>> &cells,
+        const std::vector<
+          FaceToCellTopology<VectorizedArrayType::n_array_elements>> &faces);
```
peterrum: Documentation of the parameters. I am wondering what the meaning of `cells` is (interior & exterior)?

peterrum: Is there also a function for cells?

peterrum: I guess the vector is for all faces. Any chance to provide a method, which only works on a macro cells so that we could call this function to the mapping on-the-fly class.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -349,11 +351,17 @@ namespace internal
                                                                 update_default);
         }
 
-      // Could call these functions in parallel, but not useful because the
-      // work inside is nicely split up already
-      initialize_cells(tria, cells, active_fe_index, mapping);
-      initialize_faces(tria, cells, face_info.faces, mapping);
-      initialize_faces_by_cells(tria, cells, mapping);
+      if (active_fe_index.empty() && !cells.empty() &&
```
peterrum: What is the meaning of `active_fe_index.empty()`?

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -466,6 +498,69 @@ namespace internal
           }
       }
 
+      // for second derivatives on the real cell, we need the gradient of the
```
peterrum:
```suggestion
      // For second derivatives on the real cell, we need the gradient of the
```

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -885,110 +977,50 @@ namespace internal
...
                   data.first[my_q].quadrature_point_offsets.push_back(old_size);
 
-                  if (mapping_info.get_cell_type(cell) == cartesian)
+                  if (mapping_info.get_cell_type(cell) < general)
                     {
-                      for (unsigned int q = 0;
-                           q < n_q_points_1d[my_q][fe_index];
-                           ++q)
+                      Point<dim, VectorizedArrayType> quad_point;
+                      for (unsigned int j = 0;
```
peterrum:
```suggestion
                      for (unsigned int v = 0;
```

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1100,6 +1134,185 @@ namespace internal
...
+       * This evaluates the mapping information on a range of cells calling
+       * into the tensor product interpolators of the matrix-free framework,
+       * using a polynomial expansion of the cell geometry in terms of
+       * MappingQ.
+       */
+      template <int dim, typename Number, typename VectorizedArrayType>
+      void
+      compute_range_mapping_q(
+        const unsigned int               begin_cell,
+        const unsigned int               end_cell,
```
peterrum: Cell or macro cell?

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1100,6 +1134,185 @@ namespace internal
...
+        const std::vector<GeometryType> &cell_type,
+        const std::vector<bool> &        process_cell,
+        const UpdateFlags                update_flags_cells,
+        const AlignedVector<double> &    plain_quadrature_points,
+        const FE_DGQ<dim> &              fe_geometry,
+        MappingInfoStorage<dim, dim, Number, VectorizedArrayType> &my_data)
+      {
+        // we want to use vectorization for computing the quantities, but must
+        // evaluate the geometry in double precision; thus, for floats we need
+        // to do things in two sweeps and convert the final result
```
peterrum: Why are you doing this? Is mapping in deal.II always double precision?

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1754,6 +1973,247 @@ namespace internal
...
+                             VectorizedArrayType::n_array_elements / 2 :
+                             VectorizedArrayType::n_array_elements)>;
+        constexpr unsigned int n_simd_d = VectorizedDouble::n_array_elements;
+
+        ShapeInfo<VectorizedDouble> shape_info;
+        shape_info.reinit(my_data.descriptor[0].quadrature_1d, fe_geometry);
+
+        const unsigned int n_q_points = my_data.descriptor[0].n_q_points;
+        const unsigned int n_mapping_points =
+          shape_info.dofs_per_component_on_cell;
```
peterrum: I am wondering if this not a limitation. We might want to have mapping `mapping_degree < fe_degree` particularly in the case of high-order FEM.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1957,6 +2418,380 @@ namespace internal
...
+      // MappingQGeneric
+      const MappingQGeneric<dim> *mapping_q =
+        dynamic_cast<const MappingQGeneric<dim> *>(&*this->mapping);
+      Assert(mapping_q != nullptr, ExcInternalError());
+
+      const unsigned int mapping_degree = mapping_q->get_degree();
+      const unsigned int n_mapping_points =
+        Utilities::pow(mapping_degree + 1, dim);
+      AlignedVector<double> plain_quadrature_points(cell_array.size() *
+                                                    n_mapping_points * dim);
```
peterrum: OK. Ignore my last comment.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -6945,39 +6945,25 @@ FEEvaluation<dim,
...
-        }
+      Assert(this->jacobian != nullptr, ExcNotInitialized());
+      Point<dim, VectorizedArrayType>           point = quadrature_points[0];
+      const Tensor<2, dim, VectorizedArrayType> jac   = this->jacobian[1];
+      for (unsigned int d = 0; d < dim; ++d)
+        for (unsigned int e = 0; e < dim; ++e)
+          point[d] += jac[d][e] *
+                      static_cast<Number>(
+                        this->mapping_data->descriptor[this->active_quad_index]
+                          .quadrature.point(q)[e]);
```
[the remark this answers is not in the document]
kronbichler: I was a bit unsure about this one so I did not have it from the start - now I added a second conditional branch with the Cartesian case.

include/deal.II/matrix_free/mapping_info.h
kronbichler: I see what you mean but I think this is wrong place to ask for that - this function does the whole initialization both the cells and the faces and all the other tricks we apply like looking for compressions. What we are looking for for the mapping on the fly is the things I put into the `ExtractCellHelper::compute_range_mapping_q` and `ExtractFaceHelper::compute_range_mapping_q`, or, more precisely, the body inside the loop over cell batches there. I think we should reorganize things a bit for that purpose, but I wanted to keep the generalizations (like adding the `fe_degree` and `n_q_points_1d` templates for the case we know them) out of this PR because it is big enough already. In any case, I think now we have collected the code that is needed, so we can move it around later when we use it for different purposes.

include/deal.II/matrix_free/mapping_info.templates.h
kronbichler: These are cell batches, i.e., macro cells. However, we use the terminology `cell` for both single cells and cell batches throughout that file already, and I think it would only lead to confusion if we changed the name in a single spot here. I think the meaning is clear from the context.

include/deal.II/matrix_free/mapping_info.templates.h
kronbichler: Yes, the geometry is always double precision. I actually started out with doing everything in the native `Number` type requested by the user but found out that a number of tests failed when done in `float`. I cannot say at this point if the accuracy is really important but given that we do this initialization once I think we can keep things more accurate this way. If we wanted to do things on the fly, we would of course compute everything from `float` representations of the mapping support points - that's the whole point in getting things more quickly than with precomputed arrays.

kronbichler:
I addressed the comments except for the things I explained above.

include/deal.II/matrix_free/fe_evaluation.h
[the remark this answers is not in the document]
peterrum: Thanks!

include/deal.II/matrix_free/mapping_info.h
peterrum: I agree. I put the issue onto my TODO list!

kronbichler:
I have made another cleanup to avoid setting up the same `ShapeInfo` objects in the `compute_range_mapping_q` functions several times as it also shows up for very high degrees. This PR is now ready.

kronbichler:
@dealii/dealii can someone have a look so I can start the follow-up work (parallelization by threads, `face_data_by_cells`) soon?

PR #9659, opened 2020-03-14
Replace small numbers in test suite output by zero.

bangerth:
Inspired by the discussion in #9644. Just replace every number in output files that is less than 1e-7 by zero. I could have been a bit more aggressive given that we use 1e-6 as absolute tolerance in numdiff, but I thought this is already a nice set of savings.

Let's see what happens with the tester. 

Question: Do we definitely require numdiff these days? Because this patch would most certainly break every non-numdiff test run.

/rebuild


tamiko:
On Sat, Mar 14, 2020, at 09:52 CDT, Wolfgang Bangerth <notifications@github.com> wrote:

> Question: Do we definitely require numdiff these days? Because this
> patch would most certainly break every non-numdiff test run.

We do we require numdiff these days:

(macro_deal_ii_pickup_tests.cmake)

101   IF( "${NUMDIFF_EXECUTABLE}" MATCHES "NUMDIFF_EXECUTABLE-NOTFOUND")             
102     MESSAGE(FATAL_ERROR                                                          
103       "Could not find numdiff, which is required for running the testsuite.\n"   
104       "Please specify NUMDIFF_DIR to a location containing the binary."          
105       )                                                                          
106   ENDIF()  

Best,
Matthias

tests/base/auto_derivative_function.output
```diff
@@ -22,12 +22,12 @@ f       grad         grads[0]       vgrad[1]      vgrads[1]
 4  1.96e-06 16.03 1.96e-06 16.03 1.96e-06 16.03 1.96e-06 16.03 
 8  1.23e-07 16.01 1.23e-07 16.01 1.23e-07 16.01 1.23e-07 16.01 
 16 7.66e-09 16.00 7.66e-09 16.00 7.66e-09 16.00 7.66e-09 16.00
```
marcfehling: It appears to me that you only replaced those numbers smaller than `1e-10`. Do you want to check your threshold?

tests/base/derivative_forms_01.with_trilinos=true.with_trilinos_with_sacado=true.output
```diff
@@ -1,55 +1,55 @@
...
-DEAL::F[0] : 1.00000 [ 6.12323e-17 ] [ 6.12323e-17 [ -1.00000 ] ]
-DEAL::F[1] : 1.22465e-16 [ -1.00000 ] [ -1.00000 [ -1.22465e-16 ] ]
-DEAL::dF[0] : 6.12323e-17 [ -1.00000 ]
-DEAL::dF[1] : -1.00000 [ -1.22465e-16 ]
+DEAL::F[0] : 1.00000 [ 0 ] [ 0 [ -1.00000 ] ]
+DEAL::F[1] : 0 [ -1.00000 ] [ -1.00000 [ -0 ] ]
+DEAL::dF[0] : 0 [ -1.00000 ]
+DEAL::dF[1] : -1.00000 [ -0 ]
 DEAL::ddF[0] : -1.00000
-DEAL::ddF[1] : -1.22465e-16
```
marcfehling: If you want to be pedantic, you can also remove the minus sign in these cases.

kronbichler:
I think I am partly with @peterrum here, up to the extent that I do like to see numbers below `1e-8` (or `1e-10`) in case they are not yet zero to roundoff accuracy. `numdiff` might not detect those numbers, but in case of test failures during development I often look up files and want a human-readable representation. While I would probably be fine to set all numbers less than say `1e-13` to zero, I am not in favor of a tolerance `1e-10` or `1e-8` because those numbers are not zero for double-precision computations irrespective test success detection by numdiff.

tjhei:
I am also skeptical. I see numdiff primarily as a tool to accommodate for differences between compilers/dependency versions and not to account for small tolerances rounding to 0.

bangerth:
Then let's not worry about this any more :-) But then, the corresponding changes in #9644 should also be reverted, I think :-)

PR #9660, opened 2020-03-15
Enable std::max_element for VectorizedArray

peterrum:
...to be able to write `*std::max_element(vector.begin(), vector.end()` with `vector` being of type `VectorizedArray`.

ping @kronbichler @masterleinad 

FYI @DavidSCN

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -101,9 +101,21 @@ public:
    * @param lane A pointer to the current lane.
    */
   VectorizedArrayIterator(T &data, const unsigned int lane)
-    : data(data)
+    : data(&data)
     , lane(lane)
   {}
+
+  /**
+   * Compare for equality.
+   */
+  bool
+  operator==(const VectorizedArrayIterator<T> &other) const
+  {
+    Assert(this->data == other.data,
+           ExcMessage(
+             "You are trying to compare iterators into different arrays."));
+    return this->lane == other.lane;
+  }
 
   /**
    * Compare for inequality.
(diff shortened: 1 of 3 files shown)


include/deal.II/base/vectorization.h
```diff
@@ -114,13 +123,24 @@ class VectorizedArrayIterator
     return this->lane != other.lane;
   }
 
+  /**
+   * Copy assignment.
+   */
+  VectorizedArrayIterator<T>
+  operator=(const VectorizedArrayIterator<T> &other)
```
bangerth: You might want to consider in a second step to make this just `=default` because that's what your implementation actually does.

include/deal.II/base/vectorization.h
```diff
@@ -101,10 +101,19 @@ class VectorizedArrayIterator
...
     , lane(lane)
   {}
 
+  /**
+   * Compare for equality.
+   */
+  bool
+  operator==(const VectorizedArrayIterator<T> &other) const
+  {
+    return this->lane == other.lane;
```
bangerth: Don't you also have to check for `this->data == other.data`? Otherwise you might end up with two iterators pointing into different objects but the same lane to compare equal.

But, thinking about his more, I think the usual model is to do `Assert(this->lane == other.lane, ...)`. This is in keeping with the spirit that in C++, you are only allowed to compare iterators *into the same array*. (You are allowed to compare *pointers* across objects, but for iterators the usual restriction is that the operation is only well-defined if they point into the same object, if I recall correctly.)

tests/base/vectorization_iterator_02.cc
```diff
@@ -0,0 +1,68 @@
...
+                  VectorizedArray<Number>::n_array_elements - 1);
+}
+
+template <typename Number>
+void
+test()
+{
+  VectorizedArray<Number> vector;
+
+  for (unsigned int v = 0; v < VectorizedArray<Number>::size(); v++)
```
bangerth:
```suggestion
  for (unsigned int v = 0; v < VectorizedArray<Number>::size(); ++v)
```

peterrum:
@bangerth Thanks for the review! I have applied the suggestions.

kronbichler:
The tester reports failures for the clang compiler with libc++:
https://cdash.43-1.org/testDetails.php?test=36835058&build=4982
Can you have a look?

peterrum:
> Can you have a look?

@kronbichler I guess I need to implement this interface https://en.cppreference.com/w/cpp/named_req/ForwardIterator. Thanks for the notification!

PR #4701, opened 2017-08-05
Choose tensorized evaluation template instantiation at run time

masterleinad:
Currently, finding the correct tensor evaluation strategy is implemented inside `FEEvaluation`. 
Since we might (or will) use tensorized evaluation at other places as well it makes sense to move this in a separate wrapper function.
Furthermore, this PR chooses the correct strategy not only if the `fe_degree` and `n_q_points_1d` are given as template parameters but also if they are only run time parameters 

This will immediately be useful after #4664 is merged to replace the suboptimal strategy there.

Passes all tests.

PR #9661, opened 2020-03-15
Use degree-templated FEEvaluationImpl for more quadrature formulas

kronbichler:
Given the use of `SelectEvaluator` in #9656 for the mapping I think it makes sense to somewhat extend the number of quadrature formulas that was introduced back in #4701 and we discussed the range `[degree+1, degree+3)` back in https://github.com/dealii/dealii/pull/4701#pullrequestreview-57157737. What I suggest is `[degree+1, degree+5)`, motivated by the fact that we often use mappings with a degree less than what we have quadrature points. But I am open to discussions - I do not think the compile time impact is too big from testing it locally.

I guess at some point we should have a general discussion of pre-compiled information for `FEEvaluation` (we use some related mechanisms in our application codes which use many degrees and several quadrature formulas), but that is a bigger topic and not for this PR.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_selector.h
+++ b/include/deal.II/matrix_free/evaluation_selector.h
@@ -41,7 +41,7 @@ namespace internal
     //    n_q_points=fe_degree+1.
     // 4. If the current assumption on n_q_points_1d doesn't match the runtime
     //    parameter, increase n_q_points_1d by one and try again.
-    //    If n_q_points_1d==degree+3 use the class Default which serves as a
+    //    If n_q_points_1d==degree+5 use the class Default which serves as a
     //    fallback.
 
     /**
@@ -148,7 +148,7 @@ namespace internal
                    1,
                    degree,
                    n_q_points_1d,
-                   typename std::enable_if<n_q_points_1d == degree + 3>::type>
+                   typename std::enable_if<n_q_points_1d == degree + 5>::type>
       : Default<dim, n_components, Number>
     {};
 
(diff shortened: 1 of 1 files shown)


masterleinad:
I don't really have any objections if this is a case you regularly use. Have you measured any noticeable differences in compile time at all?

kronbichler:
To measure compile times, I looked at the branch of #9656 and measured first without and then with this patch on a parallel build with similar load on the system.
Without:
```
obj_fe_release.dir/mapping_q_generic.cc.o        34.8s
obj_matrix_free_release.dir/evaluation_selector.cc.o   69.4s
obj_matrix_free_release.dir/mapping_info.cc.o          73.0s
obj_matrix_free_release.dir/mapping_info_inst2.cc.o    71.6s
obj_matrix_free_release.dir/mapping_info_inst3.cc.o    73.3s
```
With:
```
obj_fe_release.dir/mapping_q_generic.cc.o        39.6s
obj_matrix_free_release.dir/evaluation_selector.cc.o   95.9s
obj_matrix_free_release.dir/mapping_info.cc.o          94.8s
obj_matrix_free_release.dir/mapping_info_inst2.cc.o    84.1s
obj_matrix_free_release.dir/mapping_info_inst3.cc.o    80.6s
```
So I'd say there is an effect, looking at the numbers it appears to be a bit bigger than my initial "feeling" did reveal (because it is a parallel build with other things going on, so slight increases in some components get hidden). Given this number, I am not super keen on getting this in at this point and without a more comprehensive look at how to pre-compile things with matrix-free; #9664 already helps a fair bit. If you do not feel strongly @masterleinad I suggest we close this one for now.

masterleinad:
It's WIP now.

kronbichler:
Given that this is a small change and I know what to do, I suggest we close this for better visibility.

PR #9662, opened 2020-03-15
Fix two tests after elimination of logfile

kronbichler:
As documented by https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=4942 , we have two test failures as a consequence of #9650. This fixes those tests.

### DIFF
--- a/tests/mpi/torus.cc
+++ b/tests/mpi/torus.cc
@@ -76,11 +76,6 @@ main(int argc, char *argv[])
     argc, argv, testing_max_num_threads());
   MPILogInitAll log;
 
-  if (Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == 0)
-    {
-      static initlog();
-    }
-
   parallel::distributed::Triangulation<2, 3> triangulation(
     MPI_COMM_WORLD,
     typename Triangulation<2, 3>::MeshSmoothing(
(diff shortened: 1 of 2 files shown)


tests/mpi/torus.cc
```diff
@@ -76,11 +76,6 @@ main(int argc, char *argv[])
     argc, argv, testing_max_num_threads());
   MPILogInitAll log;
 
-  if (Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == 0)
-    {
-      static initlog();
```
bangerth: I don't even know what this line would have done -- isn't this a K&R-style declaration of a function?

masterleinad: That's a scoped forward declaration (but can't be `static`). https://godbolt.org/z/L1Na_R

bangerth: So what does the original code do, then? I have to admit being very confused :-)

masterleinad: The original code was
```
static std::ofstream logfile("output");
deallog.attach(logfile);
```
replacing everything except `static` by `initlog()` was appropriate in almost all cases. I just missed toi check if there was anything before `std::ofstream` in those lines.

bangerth: What I meant was: I don't understand what the syntax `static initlog();` does. The `static` suggests to me that it wasn't a function call to `initlog()`, but I also can't figure out what else this may have been :-) What code is actually executed here and why? :-))

PR #9664, opened 2020-03-15
Manually unroll a loop in matrix-free tensor product kernel

kronbichler:
This PR is complementary to #9661. It aims to improve the quality of code for the `EvaluatorTensorProduct` without the `n_rows` template (typically the degree of the finite element). I manually unroll the innermost loop for the common use case of `n_rows=2` (e.g., linear mapping interpolated to many points) and `n_rows=3`. The compiler does not seem to do it properly (neither gcc nor clang) due to the 4 nested loop and the second-innermost loop being the one where we have variable length and larger loop counts.

In the tests I made this improves performance by 1.5x to 2x.

### DIFF
--- a/include/deal.II/matrix_free/tensor_product_kernels.h
+++ b/include/deal.II/matrix_free/tensor_product_kernels.h
@@ -675,47 +675,119 @@ namespace internal
                             Utilities::fixed_power<dim - direction - 1>(n_rows);
     Assert(n_rows <= 128, ExcNotImplemented());
 
-    for (int i2 = 0; i2 < n_blocks2; ++i2)
+    // specialization for n_rows = 2 that manually unrolls the innermost loop
+    // to make the operation perform better (not completely as good as the
+    // templated one, but much better than the generic version down below,
+    // because the loop over col can be more effectively unrolled by the
+    // compiler)
+    if (contract_over_rows && n_rows == 2)
       {
-        for (int i1 = 0; i1 < n_blocks1; ++i1)
+        const Number2 *shape_data_1 = shape_data + n_columns;
+        for (int i2 = 0; i2 < n_blocks2; ++i2)
           {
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -675,47 +675,120 @@ namespace internal
...
+      {
+        const Number2 *shape_data_1 = shape_data + n_columns;
+        const Number2 *shape_data_2 = shape_data + 2 * n_columns;
+        for (int i2 = 0; i2 < n_blocks2; ++i2)
           {
-            in += stride * (mm - 1);
-            out += stride * (nn - 1);
+            for (int i1 = 0; i1 < n_blocks1; ++i1)
+              {
+                Number x0 = in[0], x1 = in[stride], x2 = in[2 * stride];
```
masterleinad:
```suggestion
                const Number x0 = in[0], x1 = in[stride], x2 = in[2 * stride];
```

include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -675,47 +675,120 @@ namespace internal
...
+                for (int col = 0; col < nn; ++col)
+                  {
+                    if (add == false)
+                      out[stride * col] = shape_data[col] * x0 +
+                                          shape_data_1[col] * x1 +
+                                          shape_data_2[col] * x2;
+                    else
+                      out[stride * col] += shape_data[col] * x0 +
```
masterleinad: You could use an auxiliary variable to avoid the duplication but I don't feel strongly about it.

kronbichler: That's a good point - I fixed this and the other suggestions.

PR #9667, opened 2020-03-16
Fix clang 10 warnings

Rombur:


include/deal.II/lac/matrix_block.h
```diff
@@ -132,6 +132,12 @@ class MatrixBlock : public Subscriptor
    */
   MatrixBlock(const MatrixBlock<MatrixType> &M);
 
+  /**
+   * Assignment operator.
+   */
+  MatrixBlock<MatrixType> &
+  operator=(const MatrixBlock<MatrixType> &) = default;
```
masterleinad: I had to look up if the copy constructor does anything special here in order to just if defaulting the copy assignment operator is OK. Both of them are, so I would just also default the other operator while being here.

source/hp/dof_handler.cc
```diff
@@ -2327,8 +2327,8 @@ namespace hp
     levels.clear();
     faces.reset();
 
-    vertex_dofs        = std::move(std::vector<types::global_dof_index>());
-    vertex_dof_offsets = std::move(std::vector<unsigned int>());
+    vertex_dofs        = std::vector<types::global_dof_index>();
+    vertex_dof_offsets = std::vector<unsigned int>();
```
bangerth: It's possible that `= {}` would also have worked. But this is fine as is.

PR #9668, opened 2020-03-16
Use copydoc for Python wrappers

agrayver:
Avoid copy-paste for functions, which are mere wrappers. 

Currently, I did only Tria wrappers. I can augment this PR for the rest of the wrappers if people think it is the right direction.

Part of #9015


agrayver:
I will add more here in the coming days.

agrayver:
This should be ready now.

agrayver:
@kronbichler ping

PR #9669, opened 2020-03-16
MF::copy_from: add missing fields

jodlbauer:
The copy constructor / `copy_from` method from `MatrixFree` did not handle `face_info` and `cell_level_index_end_local`.
The crucial one for me was `face_info`, not sure if `cell_level_index_end_local` is necessary.


tjhei:
Daniel, can you check that your email address and username is set up correctly for your commit? It looks like it is not linking to your github account (grey icon instead of your user above).

jodlbauer:
I edited name/email now, but not sure if that worked out as intended.

kronbichler:
Build failure is unrelated.

PR #9674, opened 2020-03-18
VectorizedArrayIterator enable std::distance and add operator+

peterrum:
Independent of the issue reported in https://github.com/dealii/dealii/pull/9660#issuecomment-599985295. I was working on enabling:
```cpp
VectorizedArray<Number> vec;
MatrixFree<Number> matrix_free;

std::max_element(vec.begin(), vec.begin() + matrix_free.n_components_filled (cell));
```

This PR should enable to above code and also fix the issue.

FYI @DavidSCN

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -169,6 +169,25 @@ public:
     return *this;
   }
 
+  /**
+   * Create new iterator, which is shifted by @p offset.
+   */
+  VectorizedArrayIterator<T>
+  operator+(const unsigned int &offset) const
+  {
+    return VectorizedArrayIterator<T>(*data, lane + offset);
+  }
+
+  /**
+   * Compute distance between this iterator and iterator @p other.
+   */
+  unsigned int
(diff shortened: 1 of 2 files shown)


include/deal.II/base/vectorization.h
```diff
@@ -5439,6 +5448,19 @@ namespace std
...
 
+
+
+  /**
+   * Iterator traits for VectorizedArrayIterator.
+   */
+  template <class T>
+  struct iterator_traits<dealii::VectorizedArrayIterator<T>>
+  {
+    using iterator_category = forward_iterator_tag;
```
bangerth: Not wrong, but don't you in fact have a random access iterator? (That might require you to implement `operator-` and possibly `operator +=/-=`, but that would enable a more efficient implementation of `std::distance`.)

peterrum: Thanks for the hint @bangerth! Let me try that out!

peterrum: Done :)

PR #9675, opened 2020-03-18
VectorizedArrayIterator enable std::advance()

peterrum:
As requested in https://github.com/dealii/dealii/pull/9674#pullrequestreview-376505954.

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -141,6 +141,7 @@ public:
    */
   const typename T::value_type &operator*() const
   {
+    AssertIndexRange(lane, T::size());
     return (*data)[lane];
   }
 
@@ -154,6 +155,7 @@ public:
                           typename T::value_type>::type &
   operator*()
   {
+    AssertIndexRange(lane, T::size());
     return (*data)[lane];
   }
 
@@ -165,7 +167,36 @@ public:
   VectorizedArrayIterator<T> &
   operator++()
   {
+    AssertIndexRange(lane + 1, T::size() + 1);
     lane++;
+    return *this;
+  }
+
+  /**
+   * This operator advances the iterator by @p offet lanes and returns a
+   * reference to <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator+=(const unsigned int offset)
+  {
+    AssertIndexRange(lane + offset, T::size() + 1);
+    lane += offset;
+    return *this;
+  }
+
+  /**
+   * Prefix <tt>--</tt> operator: <tt>--iterator</tt>. This operator advances
+   * the iterator to the previous lane and returns a reference to
+   * <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator--()
+  {
+    Assert(
+      lane > 0,
+      ExcMessage(
+        "You can't decrement an iterator that is already at the beginning of the range."));
+    --lane;
     return *this;
   }
 
@@ -173,8 +204,9 @@ public:
    * Create new iterator, which is shifted by @p offset.
    */
   VectorizedArrayIterator<T>
-  operator+(const unsigned int &offset) const
+  operator+(const unsigned int offset) const
   {
+    AssertIndexRange(lane + offset, T::size() + 1);
     return VectorizedArrayIterator<T>(*data, lane + offset);
   }
 
--- a/tests/base/vectorization_iterator_02.cc
+++ b/tests/base/vectorization_iterator_02.cc
@@ -51,6 +51,10 @@ test_nonconst(VectorizedArray<Number> &vector)
                                   (VectorizedArray<Number>::n_array_elements -
                                    1)),
                   VectorizedArray<Number>::n_array_elements - 1);
+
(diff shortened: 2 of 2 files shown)


include/deal.II/base/vectorization.h
```diff
@@ -169,6 +169,29 @@ class VectorizedArrayIterator
...
   }
 
+  /**
+   * This operator advances the iterator by @p offet lanes and returns a
+   * reference to <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator+=(unsigned int offset)
+  {
+    lane += offset;
```
masterleinad: What do you do if `lane` is larger than `T::n_array_elements`?

peterrum: @masterleinad I don't see a problem that `lane` is outside of the range `[0;T::size()[`. In contrast, we need at least `static_cast<unsignd int>(-1)` () and `T::size()` (in the case of incrementing) as `end()`. The problem is only when we want to dereferene the iterator. This is why I have added an assert to `operator*()`. Do you agree that this enough?

masterleinad: Yes, I meant to say that we should make sure that the element referenced stays valid. I think checking in the accessor is fine.

bangerth: I'm usually of the opinion that we should do the checking at the first possible place because it makes debugging so much simpler. So I would advocate putting an assertion here as well.

include/deal.II/base/vectorization.h
```diff
@@ -169,6 +171,29 @@ class VectorizedArrayIterator
     return *this;
   }
 
+  /**
+   * This operator advances the iterator by @p offet lanes and returns a
+   * reference to <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator+=(unsigned int offset)
```
bangerth:
```suggestion
  operator+=(const unsigned int offset)
```

include/deal.II/base/vectorization.h
```diff
@@ -169,6 +171,29 @@ class VectorizedArrayIterator
...
+  /**
+   * Prefix <tt>--</tt> operator: <tt>--iterator</tt>. This operator advances
+   * the iterator to the previous lane and returns a reference to
+   * <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator--()
+  {
+    lane--;
+    return *this;
```
bangerth: That would suggest also putting an assertion here. (And possibly in `operator++`, which is not part of this patch.)

peterrum:
@masterleinad @bangerth Thanks for the review. I have made the changes.

include/deal.II/base/vectorization.h
```diff
@@ -165,16 +167,41 @@ class VectorizedArrayIterator
   VectorizedArrayIterator<T> &
   operator++()
   {
-    lane++;
+    AssertIndexRange(lane + 1, T::size() + 1) lane++;
```
bangerth: Not wrong but hard to read :-)
```suggestion
    AssertIndexRange(lane + 1, T::size() + 1);
    ++lane;
```

include/deal.II/base/vectorization.h
```diff
@@ -165,16 +167,41 @@ class VectorizedArrayIterator
...
+  }
+
+  /**
+   * This operator advances the iterator by @p offet lanes and returns a
+   * reference to <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator+=(const unsigned int offset)
+  {
+    AssertIndexRange(lane + offset, T::size() + 1) lane += offset;
```
bangerth:
```suggestion
    AssertIndexRange(lane + offset, T::size() + 1);
    lane += offset;
```

include/deal.II/base/vectorization.h
```diff
@@ -165,16 +167,41 @@ class VectorizedArrayIterator
...
+  /**
+   * Prefix <tt>--</tt> operator: <tt>--iterator</tt>. This operator advances
+   * the iterator to the previous lane and returns a reference to
+   * <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator--()
+  {
+    AssertIndexRange(lane, T::size() + 1);
+    lane--;
```
bangerth:
```suggestion
    --lane;
```

include/deal.II/base/vectorization.h
```diff
@@ -165,16 +167,41 @@ class VectorizedArrayIterator
...
+
+  /**
+   * Prefix <tt>--</tt> operator: <tt>--iterator</tt>. This operator advances
+   * the iterator to the previous lane and returns a reference to
+   * <tt>*this</tt>.
+   */
+  VectorizedArrayIterator<T> &
+  operator--()
+  {
+    AssertIndexRange(lane, T::size() + 1);
```
bangerth: This is wrong. It should have been
```suggestion
    Assert(lane>0, ExcMessage ("You can't decrement an iterator that is already at the beginning of the range.");
```

peterrum:
@bangerth Thanks! Done!

PR #9676, opened 2020-03-18
merge_triangulations(): replace std::initializer_list by std::vector

nfehn:
The current interface with ``std::initializer_list`` is impractical as it prevents automatization.

This PR suggests to use ``std::vector`` instead. The function can still be called with an initializer list, as a vector can be initialized with such an object. At the same time, the vector improves flexibility.


drwells:
This seems reasonable to me - one cannot programatically build an initializer list.

Issue #9680, opened 2020-03-19
Make VectorizedArray compatible with algorithms from std


peterrum:
The purpose of this issue is to coordinate my effort to make `dealii::VectorizedArray` similar to the `std` containers. As a consequence different algorithms from `std` would work.

peterrum:
I think this is mostly done. Closing.

PR #9681, opened 2020-03-19
Change type of template argument width of VectorizedArray to std::size_t

peterrum:
As discussed in https://github.com/dealii/dealii/pull/9556#issuecomment-589526815.

This enables me to remove annoying workarounds.

In a follow-up PR, we should deprecate `n_array_elements` and replace its usage by `size()`. At the moment there is still a chaos since both are or can be used and they have different types. Any interest that this is done before the next release?

related to #9680

### DIFF
--- a/include/deal.II/base/vectorization.h
+++ b/include/deal.II/base/vectorization.h
@@ -77,7 +77,7 @@ DEAL_II_NAMESPACE_OPEN
 // Enable the EnableIfScalar type trait for VectorizedArray<Number> such
 // that it can be used as a Number type in Tensor<rank,dim,Number>, etc.
 
-template <typename Number, int width>
+template <typename Number, std::size_t width>
 struct EnableIfScalar<VectorizedArray<Number, width>>
 {
   using type = VectorizedArray<typename EnableIfScalar<Number>::type, width>;
@@ -100,7 +100,7 @@ public:
    * @param data The actual VectorizedArray.
    * @param lane A pointer to the current lane.
    */
-  VectorizedArrayIterator(T &data, const unsigned int lane)
+  VectorizedArrayIterator(T &data, const std::size_t lane)
     : data(&data)
     , lane(lane)
   {}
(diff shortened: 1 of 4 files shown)


kronbichler:
> In a follow-up PR, we should deprecate `n_array_elements` and replace its usage by `size()`. At the moment there is still a chaos since both are or can be used and they have different types.

I think the change of a `static constexpr` variable to a wider integer type creates little conflict, so while marking the variable deprecated we could also change the type if we think it simplifies things.

> Any interest that this is done before the next release?

If we think that using `size()` is the right strategy - and it is, given C++11 support for `constexpr` functions that should allow us to replace all appearances - we should indeed make the change sooner rather than later.

include/deal.II/base/vectorization.h
```diff
@@ -173,15 +173,15 @@ class VectorizedArrayIterator
...
+  operator+(const std::size_t &offset) const
   {
     return VectorizedArrayIterator<T>(*data, lane + offset);
   }
 
   /**
    * Compute distance between this iterator and iterator @p other.
    */
-  unsigned int
+  std::size_t
```
kronbichler: I don't know the exact semantics of the standard headers out of my head, but shouldn't this be `std::ptrdiff_t` as the difference can get negative?

bangerth:
On 3/19/20 8:18 AM, Martin Kronbichler wrote:
> If we think that using |size()| is the right strategy - and it is, given C++11 
> support for |constexpr| functions that should allow us to replace all 
> appearances - we should indeed make the change sooner rather than later.

I think that's generically true in many cases where we use static const member 
variables -- these could be replaced by a constexpr size() function. 
Standardizing to this would definitely be the right step.

include/deal.II/base/vectorization.h
bangerth: Yes, indeed. This was wrong before already (I didn't pay attention apparently).

peterrum: I have made the change. I was not aware that the value can be also negative. I think the type is not specified and only has to be consistent with `iterator_traits<dealii::VectorizedArrayIterator<T>>::difference_type`

peterrum:
I have fixed the merge conflict with master.

PR #9685, opened 2020-03-19
Disallow use of Tensor<-1> and Tensor<-2>.

bangerth:
The discussion in #9490 made me look into `tensor.h` where I realize that we have specializations for `Tensor<-1>` and `Tensor<-2>`. The comments say that we use them in some template instantiations -- though I suspect that what that comment really means is that we only *reference* these types instead of actually creating objects of these types.

In any case, creating objects of these types would likely be a bug. So let's avoid that by making the default constructors `=delete`d and not offering any other constructor either.

/rebuild

PR #9686, opened 2020-03-19
base/tensor.h - small cleanups

tamiko:
- Remove redundant overload for schur_product

 - Revert a C++98/03 overload

 - Add static_asserts to ensure that dealii::Tensor is only ever used for
   nonnegative dim and rank

### DIFF
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -58,18 +58,6 @@ namespace Differentiation
 } // namespace Differentiation
 #endif
 
-#ifndef DOXYGEN
-// Overload invalid tensor types of negative rank that come up during
-// overload resolution of operator* and related contraction variants.
-template <int dim, typename Number>
-class Tensor<-2, dim, Number>
-{};
-
-template <int dim, typename Number>
-class Tensor<-1, dim, Number>
-{};
-#endif /* DOXYGEN */
-
 
 /**
  * This class is a specialized version of the <tt>Tensor<rank,dim,Number></tt>
@@ -2010,7 +1998,8 @@ template <int rank_1,
           int rank_2,
           int dim,
           typename Number,
-          typename OtherNumber>
+          typename OtherNumber,
+          typename = typename std::enable_if<rank_1 >= 1 && rank_2 >= 1>::type>
 DEAL_II_CONSTEXPR inline DEAL_II_ALWAYS_INLINE
   typename Tensor<rank_1 + rank_2 - 2,
                   dim,
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -1946,25 +1946,6 @@ inline DEAL_II_CONSTEXPR DEAL_II_ALWAYS_INLINE
 }
 
 /**
- * Entrywise multiplication of two tensor objects of rank 1.
- *
- * @relatesalso Tensor
- */
-template <int dim, typename Number, typename OtherNumber>
-inline DEAL_II_CONSTEXPR DEAL_II_ALWAYS_INLINE
-                         Tensor<1, dim, typename ProductType<Number, OtherNumber>::type>
-                         schur_product(const Tensor<1, dim, Number> &     src1,
-                                       const Tensor<1, dim, OtherNumber> &src2)
-{
-  Tensor<1, dim, typename ProductType<Number, OtherNumber>::type> tmp(src1);
-
-  for (unsigned int i = 0; i < dim; ++i)
-    tmp[i] *= src2[i];
-
-  return tmp;
-}
-
-/**
  * Entrywise multiplication of two tensor objects of general rank.
  *
  * This multiplication is also called "Hadamard-product" (c.f.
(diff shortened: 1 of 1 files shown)


tamiko:
@bangerth Regarding #9685 - this is what I have in mind. Seems to compile so far but might require some further fixes.

include/deal.II/base/tensor.h
```diff
@@ -1945,25 +1940,6 @@ inline DEAL_II_CONSTEXPR DEAL_II_ALWAYS_INLINE
...
-                         Tensor<1, dim, typename ProductType<Number, OtherNumber>::type>
-                         schur_product(const Tensor<1, dim, Number> &     src1,
-                                       const Tensor<1, dim, OtherNumber> &src2)
-{
-  Tensor<1, dim, typename ProductType<Number, OtherNumber>::type> tmp(src1);
-
-  for (unsigned int i = 0; i < dim; ++i)
-    tmp[i] *= src2[i];
-
-  return tmp;
```
bangerth: You can't remove this function -- it builds the base case of the recursion used in the following function.

include/deal.II/base/tensor.h
```diff
@@ -2028,7 +2004,9 @@ template <int rank_1,
           int rank_2,
           int dim,
           typename Number,
-          typename OtherNumber>
+          typename OtherNumber,
+          typename = typename std::enable_if<rank_1 != 0>::type,
+          typename = typename std::enable_if<rank_2 != 0>::type>
```
bangerth: You can reduce this to a single `std::enable_if` if you just connected the conditions with `&&` :-) But I think that the right condition is actually `rank_1 + rank_2 >= 2`.

include/deal.II/base/tensor.h
tamiko: This is really unfortunate, as 
 - base case rank = 0
 - base case rank = 1
 - inductive step

is really not appealing (we have a one step recursion, and not a two step recursion). This redundancy was one of the reasons why I cleaned up the Tensor class a while ago. I will think about how to fix the underlying type issue for a minute.

tamiko: I have changed the second function variant (by explicitly forcing a `Tensor<rank-1, dim, Number>` intermediate. With this minor change `tensors/tensor_10` passes.

bangerth: Oh, I hadn't seen that there was another base case with rank=0...

include/deal.II/base/tensor.h
```diff
@@ -1986,10 +1962,11 @@ inline DEAL_II_CONSTEXPR DEAL_II_ALWAYS_INLINE
                          schur_product(const Tensor<rank, dim, Number> &     src1,
                                        const Tensor<rank, dim, OtherNumber> &src2)
 {
-  Tensor<rank, dim, typename ProductType<Number, OtherNumber>::type> tmp(src1);
+  Tensor<rank, dim, typename ProductType<Number, OtherNumber>::type> tmp;
 
   for (unsigned int i = 0; i < dim; ++i)
-    tmp[i] = schur_product(src1[i], src2[i]);
+    tmp[i] = schur_product(Tensor<rank - 1, dim, Number>(src1[i]),
+                           Tensor<rank - 1, dim, OtherNumber>(src2[i]));
```
bangerth: Why is the cast necessary? How is this not already the correct type?

include/deal.II/base/tensor.h
```diff
@@ -2028,7 +2005,8 @@ template <int rank_1,
           int rank_2,
           int dim,
           typename Number,
-          typename OtherNumber>
+          typename OtherNumber,
+          typename = typename std::enable_if<rank_1 >= 0 && rank_2 >= 0>::type>
```
bangerth: Ugh...
```suggestion
          typename = typename std::enable_if<(rank_1 >= 0 && rank_2)>= 0>::type>
```
But is this enough? You're returning a `Tensor<rank_1+rank_2-2, dim>`. I think the condition should actually be
```suggestion
          typename = typename std::enable_if<(rank_1 + rank_2 >= 2)>::type>
```

tamiko: You are right. The condition actually has to be `rank_1 >= 1 && rank_2 >= 1` to ensure that the ranks of the arguments and the result are all nonnegative. I have updated the statement accordingly.

tamiko: `rank_1 + rank_2 >= 2` is also incomplete because 0 + 2 = 2, or -1 + 3 = 2 :smile:

include/deal.II/base/tensor.h
tamiko: The problem is that `Tensor<1, dim, Number>::operator[]` returns a plain `double` instead of a `Tensor<0, dim, Number>::operator[]`.

In the original version the compiler was unable to deduce that it should use the `rank=0` template specialization in the `for` loop in  `schur_product` because it was called with two doubles once it reached `rank=1` in the recursion. By explicitly "creating" a temporary `Tensor<rank -1, dim, Number>` we avoid that problem. I actually checked that the assembly produced by the compiler in debug mode is unaffected by that change.

This probably also epxlains why we had a `rank=1` overload in the first place (because that fixes the recursion problem as well).

include/deal.II/base/tensor.h
tamiko: Well, you rightfully pointed out that this change (without the fix I commited later) breaks the test :smile:

include/deal.II/base/tensor.h
```diff
@@ -1986,10 +1962,11 @@ inline DEAL_II_CONSTEXPR DEAL_II_ALWAYS_INLINE
                          schur_product(const Tensor<rank, dim, Number> &     src1,
                                        const Tensor<rank, dim, OtherNumber> &src2)
 {
-  Tensor<rank, dim, typename ProductType<Number, OtherNumber>::type> tmp(src1);
+  Tensor<rank, dim, typename ProductType<Number, OtherNumber>::type> tmp;
```
bangerth: Nice catch, by the way :-)

include/deal.II/base/tensor.h
bangerth: Yes, `Tensor` is so lean that the compiler can disassociate/scalarize the whole thing. Think about the mess this would be at the assembly level if `dim` was a run-time variable...

Issue #9538, opened 2020-02-16
Copy the graph of modules from the new deal.II paper to the documentation


bangerth:
The current graph at https://dealii.org/developer/doxygen/deal.II/ is still from the 2007 paper. I think that the graph in the new paper is nicer and better, and we should copy it to the doc front page.

peterrum:
@bangerth Any chance to also add `MatrixFree` to the graph ;) !?

bangerth:
Sure! :-)

bangerth:
@peterrum -- can you say where in the graph #9687 the matrix-free stuff would go?

peterrum:
Closed by #9687.

PR #9687, opened 2020-03-20
Replace the connections graph with the one from the recent paper.

bangerth:
I've done my best with node placement, but graphviz in digraph mode is not nearly as good at allowing one to pick node placement as tikz is in latex. This is what I get:
image

This addresses most of #9538.

### DIFF
--- a/doc/doxygen/headers/main.h
+++ b/doc/doxygen/headers/main.h
@@ -27,37 +27,81 @@
  * the menu at the top of this page). These modules form around the building
  * blocks of any finite element program. An outline of how the primary groups
  * of classes in deal.II interact is given by the following clickable graph,
- * with a more detailed description below:
+ * with a more detailed description below (gray boxes denote a subset of the
+ * optional external libraries, gray ovals a subset of the optional external
+ * applications with which deal.II can interact):
  *
  * @dot
  digraph G
 {
   graph[rankdir="TB",bgcolor="transparent"];
 
-  edge [fontname="FreeSans",fontsize=15,labelfontname="FreeSans",labelfontsize=10];
   node [fontname="FreeSans",fontsize=15,
         shape=record,height=0.2,width=0.4,
         color="black", fillcolor="white", style="filled"];
+  edge [color="black", weight=10];
 
-  FE [label="FiniteElement",URL="\ref feall"];
-  Tria [label="Triangulation",URL="\ref grid"];
-  DoFHandler [label="DoFHandler",URL="\ref dofs"];
-  Quadrature [label="Quadrature",URL="\ref Quadrature"];
-  Mapping [label="Mapping",URL="\ref mapping"];
-  FEValues [label="FEValues",URL="\ref feaccess"];
-  Linear [label="Linear Systems",URL="\ref LAC"];
-  LinearSolver [label="Linear Solvers",URL="\ref Solvers"];
-  Output [label="Output",URL="\ref output"];
+  tria       [label="Triangulation",    URL="\ref grid", ];
+  fe         [label="Finite elements",    URL="\ref feall"];
+  mapping    [label="Mapping",          URL="\ref mapping"];
+  quadrature [label="Quadrature",       URL="\ref Quadrature"];
+  dh         [label="DoFHandler",       URL="\ref dofs"];
+  fevalues   [label="FEValues",         URL="\ref feaccess"];
+  systems    [label="Linear systems",   URL="\ref LAC"];
+  solvers    [label="Linear solvers",   URL="\ref Solvers"];
+  output     [label="Graphical output", URL="\ref output"];
+  manifold   [label="Manifold",         URL="\ref manifold"];
 
-  Tria -> DoFHandler [color="black",fontsize=10,style="solid",fontname="FreeSans"];
-  FE -> DoFHandler [color="black",fontsize=10,style="solid",fontname="FreeSans"];
-  FE -> FEValues [color="black",fontsize=10,style="solid",fontname="FreeSans"];
-  Mapping -> FEValues [color="black",fontsize=10,style="solid",fontname="FreeSans"];
-  Quadrature -> FEValues [color="black",fontsize=10,style="solid",fontname="FreeSans"];
-  FEValues -> Linear [color="black",fontsize=10,style="solid",fontname="FreeSans"];
-  DoFHandler -> Linear [color="black",fontsize=10,style="solid",fontname="FreeSans"];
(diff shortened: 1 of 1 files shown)


bangerth:
For the record, this is the current graph:
image

tjhei:
Would you mind changing the color of the external packages to something that is easier to read? The light gray on white is hard to see. Looks good otherwise!

tjhei:
Can we add a box for matrix free solvers? Maybe as an alternative to "linear systems"?

bangerth:
On 3/19/20 6:53 PM, Timo Heister wrote:
> Can we add a box for matrix free solvers? Maybe as an alternative to "linear 
> systems"?

Well, the linear systems is a doxygen module. I think maybe we should put the 
matrix-free into its own box. The question (see #9538) is where and with which 
connections. Matrix-free doesn't use FEValues, so next to "Linear systems" 
would make sense...

bangerth:
How's this for color scheme? I've also added both MF and MG nodes:
image

bangerth:
I'll add that the gray renders better on screen than in github (click on the picture if you want to see why -- the embedded picture above is downrendered).

masterleinad:
Technically, the `MatrixFree` framework of course also uses mappings and quadrature rules and also has (direct) `CUDA` support.

For me, it looks a bit odd to list `Multigrid` separately from `Linear solvers`.

peterrum:
Personally, I find the second picture too bloated. I prefer the first picture since you can see the old graph in it, which is the core of deal.II. To be able to add nevertheless `MatrixFree`, I would suggest the following picture:

path

It shows `MatrixFree` with a different style than the external packages and does not go into (implementation) details of it.

As a side remark, one could create such a graph also for the `MatrixFree` infrastructure (but that should be not in such a prominent place). Depending on the degree of detail, it would replace simply `FEValues` -> `FEEvalution` and `Linear solvers` -> `Matrix-free linear solver` and position `MatrixFree` below `DoFHandler`.

kronbichler:
I like @bangerth's picture overall because it more accurately represents the fact that MatrixFree bypasses the `FEValues` and `LinearSystems` boxes because it directly constructs a linear operator from the `DoFHandler` (plus `Quadrature` and `Mapping` ingredients). That's one of the key points that most people, even those who never worked with matrix-free methods, agree upon: You create a data structure that allows you to merge assembly and solver stages without a sparse matrix data structure.

One suggestion to reduce the clutter in @bangerth's last graph is to simply put "Multigrid" as a box that can feed "Linear solvers" to the right/left, without being fed by something on its own. That simplifies the `MatrixFree` part to its essence, of course omitting many of the details. As suggested by @peterrum we could then of course also have a matrix-free pathway where all details are given in terms of those data structures, for example for the https://www.dealii.org/developer/doxygen/deal.II/group__matrixfree.html page.

bangerth:
On 3/20/20 2:29 AM, Peter Munch wrote:
> I prefer the first picture since you can see the old graph in it, which is the 
> core of deal.II.

Maybe, but I do think of multigrid and matrixfree also as core parts. And I 
think that the goal should be to create a graph that accurately represents the 
*now* and future; we shouldn't be bound by historical precedent in our 
documentation :-)

bangerth:
I've tried for ten minutes now, but I'm not getting anywhere with this graph. The problem is that if I take away some of the MG connections, then graphviz tries to put PETSc into the middle of the graph. That's because both PETSc and Trilinos have connections to linear systems and solvers, and so to avoid cutting edges, graphviz wants to place them to the left and right of FEValues -- but then one or the other ends up in the middle of the graph.

Having spent more time than is appropriate with this, I'm afraid the last graph above is about as good as it gets :-)

peterrum:
> I like @bangerth's picture overall because it more accurately represents the fact that MatrixFree bypasses the FEValues and LinearSystems boxes because it directly constructs a linear operator from the DoFHandler (plus Quadrature and Mapping ingredients).

I agree on this. What I wanted to show that there are matrix-free alternatives (without going into detail). As @masterleinad already stated, the matrix-free infrastructure also uses `Mapping` and `Quadrature`, which does not become clear from the picture. Adding additional lines would not help. A completely separate picture (with `FEEvaluation`) would help more. Maybe I can prepare a picture for the `MatrixFree` path.

bangerth:
On 3/20/20 9:52 AM, Peter Munch wrote:
> I agree on this. What I wanted to show that there are matrix-free alternatives 
> (without going into detail). As @masterleinad 
> <https://github.com/masterleinad> already stated, the matrix-free 
> infrastructure also uses |Mapping| and |Quadrature|, which does not become 
> clear from the picture. Adding additional lines would not help. A completely 
> separate picture (with |FEEvaluation|) would help more. Maybe I can prepare a 
> picture for the |MatrixFree| path.

Yes -- I think that having the appropriate sub-graphs for each node in the 
respective doxygen modules would make great sense!

peterrum:
My suggestion for the matrix-free path would be:

path3

It is similar to the matrix-based path (see https://github.com/dealii/dealii/pull/9687#issue-391296222 and https://github.com/dealii/dealii/pull/9687#issuecomment-601584796). `FEValues` is replaced by `FEEvaluation`, `MatrixFree loops` is added and the `Linear Systems` is replaced by 
(linear/non-linear) `Operators`.

If you put both pictures next to each other I think the similarities and differences might become clear in particular for new users. In my opinion new users should first see and completely understand the matrix-based path. Once they have understood those concepts the can endeavor with more advanced features of `deal.II` (like `MatrixFree`), which is why I would mention `MatrixFree` in the matrix-based path but without focus (visualized by dashed lines - with a link to the other picture). 

I have added `Geometric Multigrid` into a separate block, since it very well fits into the context of `MatrixFree`. `Geometric Multigrid` could also get a separate picture.

Let me know what you think. If you like it, I can create a PR or post the dot-code here.

bangerth:
Don't overthink this -- it looks great :-) So yes, go ahead and put that somewhere at the top of the relevant module.

bangerth:
@dealii/dealii -- ping? Anyone want to make a call and decide one way or the other?

bangerth:
Ping?

kronbichler:
Given that we have a separate matrix-free picture, I would prefer a lean variant - either the original one without MatrixFree at the top post or a variant as laid out by @peterrum.

masterleinad:
I would also go with the first variant. In any way, you need to remove the trailing whitespace to make the indentation check happy.

bangerth:
I've gone back to the first version then, as per your suggestions.

kronbichler:
I did not hear objects, so I will merge this now - but we can always pick it up again if someone is eager in doing more.

PR #9691, opened 2020-03-20
Modularize step-20 better.

### LINKED ISSUE #9455: [Step-20] Repetition of code for assignment of problem-specific coefficients is tedious and error-prone for keeping in sync
In Step-20, we have:

```
 template <int dim>  double  PressureBoundaryValues<dim>::value(const Point<dim> &p,
                                     const unsigned int /*component*/) const
  {
    const double alpha = 0.3;
    const double beta  = 1;
    return -(alpha * p[0] * p[1] * p[1] / 2 + beta * p[0] -  alpha * p[0] * p[0] * p[0] / 6);
  }
  template <int dim>
  void ExactSolution<dim>::vector_value(const Point<dim> &p,
                                        Vector<double> &  values) const
  {
    const double alpha = 0.3;
    const double beta  = 1;
    values(0) = alpha * p[1] * p[1] / 2 + beta - alpha * p[0] * p[0] / 2;
    ......... 
  }
```

This is problematic if we want to investigate the effect of varying $\alpha$ and $\beta$, since we have to change the code in two different places. It is so easy to change in one place, and forget to change it in the 2nd function, and the compiler will happily compile, leaving the user wondering about unexpected effects in runtime/results.

What is the cleaner alternative to this? Can we declare them as global constants outside of `main()` like below?

```
constexpr double alpha = 0.3;
constexpr double beta  = 1;
```

bangerth:
Fixes #9455 by putting the exact solution, right hand side, coefficient into a namespace, and unifying the occurrence of the coefficients in one place.

Best viewed by excluding whitespace changes.

/rebuild

### DIFF
--- a/examples/step-20/step-20.cc
+++ b/examples/step-20/step-20.cc
@@ -125,154 +125,158 @@ namespace Step20
   // class. For the exact solution, we only declare the function that actually
   // returns the entire solution vector (i.e. all components of it) at
   // once. Here are the respective declarations:
-  template <int dim>
-  class RightHandSide : public Function<dim>
+  namespace PrescribedSolution
   {
-  public:
-    RightHandSide()
-      : Function<dim>(1)
-    {}
-
-    virtual double value(const Point<dim> & p,
-                         const unsigned int component = 0) const override;
-  };
+    constexpr double alpha = 0.3;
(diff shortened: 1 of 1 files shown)


kronbichler:
Failing test is unrelated.

PR #9695, opened 2020-03-20
Fix a few typos in step-69

kronbichler:


drwells:
We only change comments so lets merge without running the test suite once the CI is happy.

PR #9707, opened 2020-03-22
Print a reduced parameter tree for XML and JSON

### LINKED ISSUE #9703: ParameterHandler: add new print_parameters()
I would like to implement a new `print_parameters()` for `.json` and `.xml`, which prints a minimal tree, i.e. does not contain an additional level to print the values and does not print the action, description, ...

peterrum:
In a first attempt, this PR introduces a new parameter `print_documentation` (default value `true`) to the function `ParameterHandler::print_parameters` to be able print a reduced parameter tree in the case of XML or JSON. 

In the case of `Text`, there are the enums `Description` and `ShortText`. These enums have similar effect for `Text`. How can we make the cases consistent? I don't like the option `ShortJSON` and `ShortXML` very much...

closes #9703


source/base/parameter_handler.cc
```diff
@@ -612,6 +612,27 @@ namespace
           }
       }
   }
+
+  void
+  simplify_xml_recursively(boost::property_tree::ptree &source)
```
bangerth: This function doesn't actually have anything to do with XML -- it just works on the ptree. Maybe adjust the name of the function?

source/base/parameter_handler.cc
```diff
@@ -612,6 +612,27 @@ namespace
...
+          {
+            auto temp = p.second.get<std::string>("value");
+            p.second.clear();
+            p.second.put_value<std::string>(temp);
+          }
+        else if (p.second.get_optional<std::string>("alias"))
+          {}
+        else
+          {
+            // it must be a subsection
```
bangerth: Is there a way to assert this? I thought that there are also other fields (documentation, the pattern, ...?) and in any case it would be nice to have an assert here so that if we ever add other fields, we know that we need to update this place.

source/base/parameter_handler.cc
```diff
@@ -1140,10 +1161,16 @@ ParameterHandler::set(const std::string &entry_string, const bool new_value)
...
                                    const OutputStyle style,
-                                   const bool        sort_alphabetical) const
+                                   const bool        sort_alphabetical,
+                                   const bool        print_documentation) const
 {
   AssertThrow(out, ExcIO());
 
+  Assert((style == Text || style == LaTeX || style == Description ||
+          style == ShortText) ^
+           !print_documentation,
```
bangerth: Ugh, xor and negation! Nobody will be able to read this :-) Just write it as `(A && !B) || (!A && B)`.

source/base/parameter_handler.cc
```diff
@@ -1176,6 +1203,9 @@ ParameterHandler::print_parameters(std::ostream &    out,
   // first
   if (style == XML)
     {
+      if (print_documentation == false)
+        simplify_xml_recursively(*current_entries);
```
bangerth: Don't you need to call this on `sorted_entries`? I may be misreading the code here, but it seems to me like you're permanently changing the state of the `entries` object...

tests/parameter_handler/parameter_handler_write_json.cc
```diff
@@ -59,8 +59,18 @@ main()
...
+  prm.print_parameters(deallog.get_file_stream(),
+                       ParameterHandler::JSON,
+                       true,
+                       true);
   deallog.get_file_stream() << std::endl;
 
+  prm.print_parameters(deallog.get_file_stream(),
+                       ParameterHandler::JSON,
+                       true,
+                       false);
```
bangerth: You can test my hypothesis that you're changing the state of the object by adding another `..., true)` output here and checking that the first and third outputs are the same :-)

peterrum:
@bangerth Thanks for the review!  I went with `ShortXML` and `ShortJSON` without the binary operations. Latter might be useful if we intend to pass more parameters in such a way (as we do in `FEValues`): e.g. also `sort_alphabetical`. I would not mind that (before the next release we can still do it).

You were right that I was modifying the actual tree and not a copy. I am now using the `sorted_entries` copy.

I also have added some comments.

source/base/parameter_handler.cc
```diff
@@ -1174,7 +1219,22 @@ ParameterHandler::print_parameters(std::ostream &    out,
...
-  if (style == XML)
+
+  // explicity eliminate the documentation from the tree if requested
+  if (style == ShortXML || style == ShortJSON)
+    {
+      // create a copy of the tree if it has not been done yet
+      if (sort_alphabetical == false)
+        {
+          sorted_or_simplified_entries = *entries;
+          current_entries              = &sorted_or_simplified_entries;
```
bangerth: This repeated copying and taking the address of here and above is just very difficult to follow.

Just copy the whole damn structure once at the top unconditionally into a new variable. Then you don't need to follow two variables with a pointer pointing to one of the two, and the whole sorting and shrinking code gets easier.

peterrum: I have made the change. 

Regarding your first comment: the function has the same structure as `read_xml_recursively()`; that is why I added the following comment:

```cpp
  // This function is strongly based on read_xml_recursively().
```

I guess the way it is done is safe, since it is done also there.

peterrum:
> Is there a way to assert this? I thought that there are also other fields (documentation, the pattern, ...?) 

The format of the internal tree can be expressed in form of a regular grammar with three node types: alias-nodes (A), value-nodes (V), and section nodes (S) like this:

```
A -> "alias"
V -> "value"(,"documentation", "default", "pattern", "action")?
S -> (A|S|V)*
```
Checking for "A"- and "V"-nodes is simple by checking for the strings "alias" or "values". I use the "?" to express that these attributes are not relevant for creating the reduced tree.

> and in any case it would be nice to have an assert here so that if we ever add other fields, we know that we need to update this place.

Checking for "S"-nodes is not that easy; we might have different (corner) cases:
* `S->epsilon` (empty section)
* `S->S...S` (arbitrary number of subsections)

How to assert these cases, I don't know.

Adding additional fields to the "V"-nodes is possible. These will be simply ignored by the algorithm. Adding new node types, will be simply treated as a section.

bangerth:
Thanks for the explanation! This looks good to me then!

/rebuild

PR #9708, opened 2020-03-22
Fix compilation with Tpetra vector

kronbichler:
This attempts to fix the build failure on configurations with Tpetra like most of our testers:
https://cdash.43-1.org/viewBuildError.php?buildid=5046
I do not have the Tpetra compatible on my system so it would be good if someone who has it could check, but I believe that this is the right thing to do.

This is due to #9100.


tjhei:
Thanks for fixing my PR! :+1:

PR #8298, opened 2019-06-02
Replace MPI_Allgather by MPI_Exscan

kronbichler:
This PR, prepared together with @peterrum, replaces the use of `MPI_Allgather` by the much more efficient `MPI_Exscan` in those cases where we wanted to get a parallel prefix sum (MPI calls the version of the prefix sum where the local element is excluded `MPI_Exscan`).

The second component of this PR is to not store `locally_owned_dofs_per_processor` and `n_locally_dofs_per_processor` in the `NumberCache` of `DoFHandler` according to the discussion in #8067. A clean solution to this change necessitates an incompatible change, namely that we have to make the return type of those vectors by value rather than const reference. I think the use cases are limited. We use those vectors in a few dozens of tests - on the one hand, we utilize them for `distribute_sparsity_pattern` that we should replace by a better implementation in a subsequent pull request, and then for some sanity checks. One thing that does not work any more is that we cannot call `dof_handler.locally_owned_dofs_per_processor()` inside loops whose number of execution is different on different ranks because each call now implies global communication. The alternative would had been to populate the variable on the first call to those functions and then return it. It would have required a lock instead to make only one thread fill it. It would have avoided touching all those tests, but I really think we should discourage those functions because they intrinsically do not scale.

Closes #8067.

PR #9710, opened 2020-03-22
SparsityTools::distribute_sparsity_pattern with compute_index_owner

### LINKED ISSUE #8306: Replace n_..._dofs_per_processor in distribute_sparsity_pattern
In the discussion of #8298 we observed that `SparsityTools::distribute_sparsity_pattern` still uses the `n_locally_owned_dofs_per_processor` information that we want to discourage. This issue is to remind ourselves to eventually implement a better-performing alternative that avoids the MPI all-to-all communication, see also #8293.

kronbichler:
This PR adds alternative functions for `SparsityTools::distribute_sparsity_pattern()` and `SparsityTools::gather_sparsity_pattern()` that only take the locally owned index set of the present MPI process, rather than that of all processes. This allows us to simplify the tutorials and a few dozen of test cases, and in particular avoids the expensive call to `MPI_Allgather` discussed in #8298 and #8293, expensive meaning that with more than a few thousands of rank, the new implementation should be considerably faster. I also deprecated two of the three old functions which do not make sense any more; I left the one with the number of dofs per processor because it might be of use in some corner cases.

One of these corner cases are `dealii::Triangulation` cases where the linear system gets split among processes (but not the mesh), which I left as they were - they are mostly based off some variant of step-36.

Fixes #8306 .

This PR also significantly reduces the number of calls to `DoFHandler::compute_{n_}locally_owned_dofs_per_processor()` in the test suite. We can further eliminate a considerable number of them by initializing PETSc sparse matrices from an `IndexSet` of locally owned data rather than *all* sizes - I do not think they belong here. If we believe they are important to switch (personally I am indifferent), we can open an issue to fix those later.

### DIFF
--- a/source/lac/sparsity_tools.cc
+++ b/source/lac/sparsity_tools.cc
@@ -915,52 +915,49 @@ namespace SparsityTools
 
   void
   gather_sparsity_pattern(DynamicSparsityPattern &     dsp,
-                          const std::vector<IndexSet> &owned,
+                          const std::vector<IndexSet> &owned_rows_per_processor,
                           const MPI_Comm &             mpi_comm,
                           const IndexSet &             ghost_range)
   {
     const unsigned int myid = Utilities::MPI::this_mpi_process(mpi_comm);
+    gather_sparsity_pattern(dsp,
+                            owned_rows_per_processor[myid],
+                            mpi_comm,
+                            ghost_range);
+  }
 
-    AssertDimension(owned.size(), Utilities::MPI::n_mpi_processes(mpi_comm));
-#  ifdef DEBUG
-    for (const auto &set : owned)
-      Assert(set.is_contiguous(), ExcNotImplemented());
-#  endif
 
-    Assert(owned[myid].is_ascending_and_one_to_one(mpi_comm),
-           ExcNotImplemented());
 
-    std::vector<DynamicSparsityPattern::size_type> start_index(owned.size() +
-                                                               1);
-    start_index[0] = 0;
-    for (DynamicSparsityPattern::size_type i = 0; i < owned.size(); ++i)
-      start_index[i + 1] = start_index[i] + owned[i].n_elements();
-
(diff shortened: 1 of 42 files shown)


kronbichler:
Could someone take a look here? While I'm confident that all things work as expected and more efficiently than before, I'd feel safer if we have a month with this feature in deal.II before the release so some users get to pick it up.

source/lac/sparsity_tools.cc
```diff
@@ -1040,42 +1032,51 @@ namespace SparsityTools
...
+    distribute_sparsity_pattern(dsp, owned, mpi_comm, myrange);
+  }
+
+
+
+  void
+  distribute_sparsity_pattern(DynamicSparsityPattern &dsp,
+                              const IndexSet &        owned,
+                              const MPI_Comm &        mpi_comm,
+                              const IndexSet &        myrange)
```
drwells, line 1044: Would you mind changing these variable names to match the ones in the declaration? It makes the code a bit easier to read.

source/lac/sparsity_tools.cc
```diff
@@ -1099,66 +1100,67 @@ namespace SparsityTools
...
+                                myrange);
+  }
+
+
 
+  void
+  distribute_sparsity_pattern(BlockDynamicSparsityPattern &dsp,
+                              const IndexSet &             owned,
+                              const MPI_Comm &             mpi_comm,
+                              const IndexSet &             myrange)
```
drwells, line 1124: same

source/lac/sparsity_tools.cc
kronbichler: Definitely, I will do it now!

kronbichler: Done!

kronbichler:
Any further comments?

drwells:
Nope, lets merge!

PR #9712, opened 2020-03-22
Fix compilation with Tpetra vector, 2nd attempt

kronbichler:
Unfortunately my attempt #9708 failed (in the absence of being able to easily test it without a complete Trilinos rebuild):
https://cdash.43-1.org/viewBuildError.php?buildid=5157
Here is my second attempt. I can get a proper setup to check myself but that won't be before tomorrow, so if anyone else is quicker to verify please go ahead.


tamiko:
directly testing on bob now.

PR #9718, opened 2020-03-23
improve robustness and usability of ParameterHandler

nfehn:
The ``ParameterHandler`` allows partial parsing of input files, which is very useful. However, it seems as if it does not assert that all entries added to the ``ParameterHandler`` object are indeed found in the input file (at least I could not find such a functionality).

For example you might write in the, e.g. .json, input file

``
{
    "General": {
        "Precison": "double"
     }
}
``

but the parameter will be ignored because of the typo in **Precision**. Not being a big issue here as it might "only" affect the run time, this behavior can lead to subtle issues that are difficult to find, and more importantly, one might not even recognize that the simulation has been run for the wrong set of parameters.

This PR suggest a fix to this problem by adding a new function to the ``ParameterHandler`` that makes sure that all entries have been found. It does not change the behavior of the current code but could be incorporated in follow-up PRs, e.g., by extending the current functions by an additional parameter with appropriate default value.

To write application code that does not prescribe the type of input file provided (prm, xml, json), it is useful to have a ``parse_input()`` function in the library that allows filetype-independent programming.

### DIFF
--- a/doc/news/changes/minor/20200323NiklasFehn
+++ b/doc/news/changes/minor/20200323NiklasFehn
@@ -1,0 +1,7 @@
+New: ParameterHandler: add functionality ensuring that parameters are 
+indeed set in order to avoid errors that remain unrecognized otherwise.
+By extending the interfaces of declare_entry() and add_parameter(),
+parameters can be declared as mandatory parameters that must be set
+when parsing from an input file or by a set() function.
+<br>
+(Niklas Fehn, 2020/03/23)
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1056,17 +1056,23 @@ public:
    * Declare a new entry with name <tt>entry</tt>, default and for which any
    * input has to match the <tt>pattern</tt> (default: any pattern).
    *
-   * The last parameter defaulting to an empty string is used to add a
-   * documenting text to each entry which will be printed as a comment when
-   * this class is asked to write out all declarations to a stream using the
-   * print_parameters() function.
-   *
    * The function generates an exception of type ExcValueDoesNotMatchPattern
    * if the default value doesn't match the given pattern, using the C++ throw
    * mechanism. However, this exception is only generated <i>after</i> the
    * entry has been created; if you have code where no sensible default value
    * for a parameter is possible, you can then catch and ignore this
    * exception.
+   *
+   * The parameter @p documentation defaulting to an empty string is used
+   * to add a documenting text to each entry which will be printed as a comment
+   * when this class is asked to write out all declarations to a stream using
+   * the print_parameters() function.
+   *
+   * The parameter @p has_to_be_set can be used in order to declare this
+   * parameter as a parameter whose default value has to be overwritten by
+   * one of the methods provided by this class. Whether a parameter has been set
+   * succesfully can be queried by the functions get_entries_wrongly_not_set()
+   * and assert_that_entries_have_been_set().
    *
    * @note An entry can be declared more than once without generating an
    * error, for example to override an earlier default value.
@@ -1075,7 +1081,8 @@ public:
   declare_entry(const std::string &          entry,
                 const std::string &          default_value,
                 const Patterns::PatternBase &pattern = Patterns::Anything(),
-                const std::string &          documentation = "");
+                const std::string &          documentation = "",
+                const bool                   has_to_be_set = false);
 
   /**
    * Attach an action to the parameter with name @p entry in the current
@@ -1132,6 +1139,12 @@ public:
    *
    * By default, the pattern to use is obtained by calling the function
    * Patterns::Tools::Convert<T>::to_pattern(), but a custom one can be used.
+   *
+   * The parameter @p has_to_be_set can be used in order to declare this
+   * parameter as a parameter whose default value has to be overwritten by
+   * one of the methods provided by this class. Whether a parameter has been set
+   * succesfully can be queried by the functions get_entries_wrongly_not_set()
+   * and assert_that_entries_have_been_set().
    */
   template <class ParameterType>
   void
@@ -1139,7 +1152,8 @@ public:
                 ParameterType &              parameter,
                 const std::string &          documentation = "",
                 const Patterns::PatternBase &pattern =
-                  *Patterns::Tools::Convert<ParameterType>::to_pattern());
+                  *Patterns::Tools::Convert<ParameterType>::to_pattern(),
+                const bool has_to_be_set = false);
 
   /**
    * Create an alias for an existing entry. This provides a way to refer to a
@@ -1489,6 +1503,25 @@ public:
   operator==(const ParameterHandler &prm2) const;
 
   /**
+   * Return a set of parameter names (including subsection names) corresponding
+   * to those entries of the parameter handler that have not been set by one of
+   * the functions parsing parameters from an input file or by an explicit call
+   * to one of the set() functions, but that have been declared as mandatory
+   * parameters that must be set (through the last argument of the
+   * declare_entry() function or add_parameter() function).
+   */
+  std::set<std::string>
+  get_entries_wrongly_not_set() const;
+
+  /**
+   * Asserts that those entries of the parameter handler with flag
+   * `has_to_be_set = true` have been set. An exception is invoked
+   * if at least one of these parameters has not been set.
+   */
+  void
+  assert_that_entries_have_been_set() const;
+
+  /**
    * @addtogroup Exceptions
    * @{
    */
@@ -1638,6 +1671,16 @@ private:
    * works around a problem with gcc 4.5.
    */
   std::unique_ptr<boost::property_tree::ptree> entries;
+
+  /**
+   * A map that stores a pair of boolean variables for each entry
+   * added to the parameter handler. The first bool describes whether the
+   * parameter has to be set according to the last argument of the functions
+   * declare_entry() or add_parameter(), and the second bool contains the
+   * information whether the parameter has been set by any of the functions
+   * parsing input parameters or by a set function of this class.
+   */
+  std::map<std::string, std::pair<bool, bool>> entries_set_status;
 
   /**
    * A list of patterns that are used to describe the parameters of this
@@ -2192,7 +2235,8 @@ void
 ParameterHandler::add_parameter(const std::string &          entry,
                                 ParameterType &              parameter,
                                 const std::string &          documentation,
-                                const Patterns::PatternBase &pattern)
+                                const Patterns::PatternBase &pattern,
+                                const bool                   has_to_be_set)
 {
   static_assert(std::is_const<ParameterType>::value == false,
                 "You tried to add a parameter using a type "
@@ -2202,7 +2246,8 @@ ParameterHandler::add_parameter(const std::string &          entry,
                 Patterns::Tools::Convert<ParameterType>::to_string(
                   parameter, pattern.clone()),
                 pattern,
-                documentation);
+                documentation,
+                has_to_be_set);
 
   std::string        path = get_current_full_path(entry);
   const unsigned int pattern_index =
--- a/source/base/parameter_handler.cc
+++ b/source/base/parameter_handler.cc
@@ -473,7 +473,7 @@ ParameterHandler::parse_input(std::istream &     input,
     }
 
   // While it does not make much sense for anyone to actually do this, allow
-  // the last line to end in a backslash. to do do, we need to parse
+  // the last line to end in a backslash. To do so, we need to parse
   // whatever was left in the stash of concatenated lines
   if (is_concatenated)
     scan_line_or_cleanup(fully_concatenated_line, filename, current_line_n);
@@ -740,6 +740,7 @@ void
 ParameterHandler::clear()
 {
   entries = std_cxx14::make_unique<boost::property_tree::ptree>();
+  entries_set_status.clear();
 }
 
 
@@ -748,7 +749,8 @@ void
 ParameterHandler::declare_entry(const std::string &          entry,
                                 const std::string &          default_value,
                                 const Patterns::PatternBase &pattern,
-                                const std::string &          documentation)
+                                const std::string &          documentation,
+                                const bool                   has_to_be_set)
 {
   entries->put(get_current_full_path(entry) + path_separator + "value",
                default_value);
@@ -756,6 +758,13 @@ ParameterHandler::declare_entry(const std::string &          entry,
                default_value);
   entries->put(get_current_full_path(entry) + path_separator + "documentation",
                documentation);
+
+  // initialize with false
+  const std::pair<bool, bool> set_status =
+    std::pair<bool, bool>(has_to_be_set, false);
+  entries_set_status.insert(
+    std::pair<std::string, std::pair<bool, bool>>(get_current_full_path(entry),
+                                                  set_status));
 
   patterns.reserve(patterns.size() + 1);
   patterns.emplace_back(pattern.clone());
@@ -1131,6 +1140,14 @@ ParameterHandler::set(const std::string &entry_string,
 
       // finally write the new value into the database
       entries->put(path + path_separator + "value", new_value);
+
+      auto map_iter = entries_set_status.find(path);
+      if (map_iter != entries_set_status.end())
+        map_iter->second = std::pair<bool, bool>(map_iter->second.first, true);
+      else
+        AssertThrow(false,
+                    ExcMessage("Could not find parameter " + path +
+                               " in map entries_set_status."));
     }
   else
     AssertThrow(false, ExcEntryUndeclared(entry_string));
@@ -1963,6 +1980,47 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
   write_json(o1, *entries);
   write_json(o2, *prm2.entries);
   return (o1.str() == o2.str());
+}
+
+
+
+std::set<std::string>
+ParameterHandler::get_entries_wrongly_not_set() const
+{
+  std::set<std::string> entries_wrongly_not_set;
+
+  for (auto it : entries_set_status)
+    if (it.second.first == true && it.second.second == false)
+      entries_wrongly_not_set.insert(it.first);
+
+  return entries_wrongly_not_set;
+}
+
+
+
+void
+ParameterHandler::assert_that_entries_have_been_set() const
+{
+  const std::set<std::string> entries_wrongly_not_set =
+    this->get_entries_wrongly_not_set();
+
+  if (entries_wrongly_not_set.size() > 0)
+    {
+      std::string list_of_missing_parameters = "\n\n";
+      for (const auto &it : entries_wrongly_not_set)
+        list_of_missing_parameters += "  " + it + "\n";
+      list_of_missing_parameters += "\n";
+
+      AssertThrow(
+        entries_wrongly_not_set.size() == 0,
+        ExcMessage(
+          "Not all entries of the parameter handler that were declared with "
+          "`has_to_be_set = true` have been set. The following parameters " +
+          list_of_missing_parameters +
+          " have not been set. "
+          "A possible reason might be that you did not add these parameter to "
+          "the input file or that their spelling is not correct."));
+    }
 }
 
 
--- a/tests/parameter_handler/parameter_handler_25.cc
+++ b/tests/parameter_handler/parameter_handler_25.cc
@@ -1,0 +1,126 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#include <deal.II/base/parameter_handler.h>
+
+#include <map>
+
+#include "../tests.h"
+
+void
+success()
+{
+  unsigned int dim       = 2;
+  std::string  precision = "double";
+
+  ParameterHandler prm;
+  prm.enter_subsection("General");
+  // this one does not have to be set
+  prm.add_parameter("dim",
+                    dim,
+                    "Number of space dimensions",
+                    Patterns::Integer(2, 3));
+  // this one has to be set
+  prm.declare_entry("Precision",
+                    precision,
+                    Patterns::Selection("float|double"),
+                    "Floating point precision",
+                    true);
+  prm.leave_subsection();
+
+  // this try-catch simulates parsing from an incomplete/incorrect input file
+  try
+    {
+      prm.enter_subsection("General");
+      prm.set("Precision", "float");
+      prm.leave_subsection();
+    }
+  catch (std::exception &exc)
+    {
+      deallog << exc.what() << std::endl;
+    }
+
+  // check set status
+  try
+    {
+      prm.assert_that_entries_have_been_set();
+    }
+  catch (std::exception &exc)
+    {
+      deallog << exc.what() << std::endl;
+    }
+
+  deallog << std::endl << "successful" << std::endl;
+}
+
+void
+fail()
+{
+  unsigned int dim       = 2;
+  std::string  precision = "double";
+
+  ParameterHandler prm;
+  prm.enter_subsection("General");
+  // both parameters have to be set
+  prm.add_parameter(
+    "dim", dim, "Number of space dimensions", Patterns::Integer(2, 3), true);
+  prm.add_parameter("Precision",
+                    precision,
+                    "Floating point precision",
+                    Patterns::Selection("float|double"),
+                    true);
+  prm.leave_subsection();
+
+  // this try-catch simulates parsing from an incomplete/incorrect input file
+  try
+    {
+      prm.enter_subsection("General");
+      prm.set("Precison", "float"); // here is a typo!
+      // dim is not set!
+      prm.leave_subsection();
+    }
+  catch (std::exception &exc)
+    {
+      deallog << exc.what() << std::endl;
+    }
+
+  // check set status
+  try
+    {
+      prm.assert_that_entries_have_been_set();
+    }
+  catch (std::exception &exc)
+    {
+      deallog << exc.what() << std::endl;
+    }
+}
+
+
+int
+main()
+{
+  initlog();
+  deallog.get_file_stream().precision(3);
+
+  try
+    {
+      success();
+      fail();
+    }
+  catch (std::exception &exc)
+    {
+      deallog << exc.what() << std::endl;
+    }
+}
--- a/tests/parameter_handler/parameter_handler_25.output
+++ b/tests/parameter_handler/parameter_handler_25.output
@@ -1,0 +1,28 @@
+
+DEAL::
+DEAL::successful
+DEAL::
+--------------------------------------------------------
+An error occurred in file <parameter_handler.cc> in function
+    void dealii::ParameterHandler::set(const string&, const string&)
+The violated condition was:
+    false
+Additional information:
+    You can't ask for entry <Precison> you have not yet declared.
+--------------------------------------------------------
+
+DEAL::
+--------------------------------------------------------
+An error occurred in file <parameter_handler.cc> in function
+    void dealii::ParameterHandler::assert_that_entries_have_been_set() const
+The violated condition was:
+    entries_wrongly_not_set.size() == 0
+Additional information:
+    Not all entries of the parameter handler that were declared with `has_to_be_set = true` have been set. The following parameters
+
+  General.Precision
+  General.dim
+
+ have not been set. A possible reason might be that you did not add these parameter to the input file or that their spelling is not correct.
+--------------------------------------------------------
+


doc/news/changes/minor/20200323NiklasFehn
```diff
@@ -0,0 +1,8 @@
+New: ParameterHandler::assert_all_entries_have_been_found(), free function
+parse_input()
+The first function asserts that parameters added to the parameter handler
+are indeed found in the input file and fixes robustness problems of the code.
+The second function allows to parse from an input file independently of its 
+file type (prm, xml, json) and used the first function.
```
bangerth: uses?

include/deal.II/base/parameter_handler.h
```diff
@@ -1503,6 +1503,14 @@ class ParameterHandler : public Subscriptor
   bool
   operator==(const ParameterHandler &prm2) const;
 
+  /**
+   * Asserts that all entries of the parameter handler have been found in
+   * the input file. Hence, it only makes sense to call this function after
+   * the parameters have been parsed from an input file.
+   */
+  void
+  assert_all_entries_have_been_found() const;
```
bangerth: Functions that call `Assert` are awkward to use because you can only call them in exactly one context: when you actually do want an assertion. But you have no control over what assertion that is. A better design, I think, would be if you let the function return a `bool` and then the user can decide what they want to do with the result. So maybe call it `all_entries_have_been_set()`? You'd have to update the documentation.

You might want to update the documentation to specify what "have been found" actually means. I think that that means that each parameter has been set, through one of the previous calls to any of the `parse_input()` functions (they can be called multiple times, with multiple input files) or an explicit call to `ParameterHandler::set()` (do you actually handle that as well?).

include/deal.II/base/parameter_handler.h
```diff
@@ -1654,6 +1662,12 @@ class ParameterHandler : public Subscriptor
    */
   std::unique_ptr<boost::property_tree::ptree> entries;
 
+  /**
+   * Map that stores whether an entry added to the parameter handler has
+   * been found in the input file.
+   */
+  std::map<std::string, bool> entries_found;
```
bangerth: maybe call it `entries_set` or `entries_written_to`?

include/deal.II/base/parameter_handler.h
```diff
@@ -1654,6 +1662,12 @@ class ParameterHandler : public Subscriptor
    */
   std::unique_ptr<boost::property_tree::ptree> entries;
 
+  /**
+   * Map that stores whether an entry added to the parameter handler has
```
bangerth:
```suggestion
   * A map that stores whether an entry added to the parameter handler has
```

include/deal.II/base/parameter_handler.h
nfehn: This should be handled as well if all the code runs over the same set() function, which should be the case

include/deal.II/base/parameter_handler.h
nfehn: yes, I prefer your first option

done in https://github.com/dealii/dealii/pull/9718/commits/394b9d581ff172d6d979fa04cc1b5c908f59e3e3

I also improved the docu

include/deal.II/base/parameter_handler.h
nfehn: done in https://github.com/dealii/dealii/pull/9718/commits/394b9d581ff172d6d979fa04cc1b5c908f59e3e3

doc/news/changes/minor/20200323NiklasFehn
nfehn: yes, has been updated

include/deal.II/base/parameter_handler.h
nfehn: First of all I am happy that the design with the map is acceptable, I was a bit afraid that it is a bit of a hack, given that I do not have a good understanding of the ``ParameterHandler`` class. I am not sure how to resolve your first point because I want to know which parameter causes the problems so that the user immediately knows where to look at. Returning more than a bool might probably complicate code readability and documentation. In some sense I have chosen your design, because in the ``parse_input()`` function (second function of this PR) I tell the code through an additional parameter when I want to allow an assertion, and one could use a default value of ``false`` for example.

bangerth: Then return that subset of the map that corresponds to variables not yet set (e.g., returning a `std::set` that contains their keys). The user can then do with that whatever they want: if the set is empty, everything is fine. If the set is not empty, they can create an error that only contains the first unset entry (as you currently do), or outputting all unset entries.

nfehn: Would it be ok if this is done in another function so that the function with the assert can remain and the user can decide which option to call?

nfehn: I made the requested changes here https://github.com/dealii/dealii/pull/9718/commits/f7560419e84c73fa7f53ecce39a1decc7f3c67ea

nfehn:
> I've left you a comment below about the design. I think you probably want to clear the map in clear() as well.

I changed this here https://github.com/dealii/dealii/pull/9718/commits/48910856e99821f6f58647787d30c258ec1e0782

I think i have addressed all the requested changes. Thanks a lot for your input.

The reason why I have chosen a free function is that this function only makes use of the public interface of ``ParameterHandler``, which is why I thought it would be the better design this way, but I do not insist on that and can change this as well.

nfehn:
@bangerth would you also prefer the option by @peterrum or would you stick to the map-design?

peterrum:
I have gone through your discussionn above. I have a couple of questions and comments from a user perspective (independent from the actual internal implementation - as I suggested above). I think the feature proposed in this PR is quite useful and a great addition!

Would it not be an option to add to the `parse_inpute_*`-functions an additional parameter after `skip_undefined`. E.g.:
```cpp
virtual void parse_input (std::istream &input, 
                          const std::string &filename="input file", 
                          const std::string &last_line="", 
                          const bool skip_undefined=false, 
                          const bool assert_all_parameters_have_been_set=false);
```
At the end of the function call, it is optionally asserted that all parameters have been set.

For asserting, a function (`all_entries_have_been_set()`) could be created that returns a bool as @bangerth suggested. This function could be called at the end of the parse functions, via:
```
if(assert_all_parameters_have_been_set == true)
  AssertThrow(all_entries_have_been_set(),  ExcMessage (get_first_unset_entry()));
```

The function `all_entries_have_been_set()` should be accessible from outside since a user might process multiple input files with the same `ParameterHandler`. The user check would look exactly like the code snippet above (so that no dedicated function for asserting would be needed).

Regarding returning a set: is that really needed? Isn't it enough to return only a single entry? When would you need a set? A single entry would be sufficient to create an assert message, that this entry has not been set yet. Or would you want to output all unset parameters?

nfehn:
Regarding the first part, I would also suggest this as further improvements. Regarding the second part, I think one will in the end have to write the ``Assert()`` line with the two function calls multiple times due to the various ``parse_input()`` functions (DRY - don't repeat yourself - paradigm), i.e., there is clearly a need for a function that asserts in my opinion. If I understood the discussion yesterday correctly, the idea is that the user in the application code writes ``get_entries_not_set().size() == 0`` and ``get_entries_not_set().begin()`` to get the action of the two suggested functions in the Assert, but the user has additional flexibility, i.e., the user can also print the whole list of parameters that are not set.

nfehn:
@peterrum I thought once again about your comments. Having slept on it for a night, I think what we want is even greater flexibility. Doing it in a black-or-white fashion in the sense that we enforce all parameters to be set or none is too rigid. There are parameters for which reasonable default values exist such as my example in the beginning, ``{ "Precision": "Double" }``, and for which we do not want to enforce the user to set them, e.g., in order to keep the input file compact. On the other hand, there are important parameters that have to be set by the user and in which cases the user has to be made aware of the fact that the parameter has not been set. So I think we actually want to be able to define in a parameter-wise fashion which individual parameters have to be set through an additional parameter ``has_to_be_set = false`` of the functions ``add_parameter()`` and ``declare_entry()``.

These changes can be found here https://github.com/dealii/dealii/pull/9718/commits/524ac4354cb4c35f7174fe128a82032565775873 .

source/base/parameter_handler.cc
```diff
@@ -703,14 +704,29 @@ void
...
   entries->put(get_current_full_path(entry) + path_separator + "value",
                default_value);
   entries->put(get_current_full_path(entry) + path_separator + "default_value",
                default_value);
   entries->put(get_current_full_path(entry) + path_separator + "documentation",
                documentation);
+  entries->put(get_current_full_path(entry) + path_separator + "has_to_be_set",
+               has_to_be_set);
+  entries->put(get_current_full_path(entry) + path_separator + "has_been_set",
+               false);
```
peterrum: I like the suggested flexibility! Being able to set it via `declare_entry()` and `add_parameter()` is a good idea!

However, I don't think we don't need to add two parameters to the tree. I would suggest:
```suggestion
  entries->put(get_current_full_path(entry) + path_separator + "has_seen_set",
               !has_to_be_set);
```
or alternately:
```suggestion
  if(has_to_be_set)
    entries->put(get_current_full_path(entry) + path_separator + "has_seen_set",
                 false);
```


Also, I would suggest in this case to eliminate the map (because it only contains duplicated information).

source/base/parameter_handler.cc
```diff
@@ -1086,6 +1102,13 @@ ParameterHandler::set(const std::string &entry_string,
 
       // finally write the new value into the database
       entries->put(path + path_separator + "value", new_value);
+      entries->put(path + path_separator + "has_been_set", true);
+
+      entries_set_status.find(path)->second = std::pair<bool, bool>(
+        entries->get<std::string>(path + path_separator + "has_to_be_set") ==
+          "true",
+        entries->get<std::string>(path + path_separator + "has_been_set") ==
+          "true");
```
peterrum: Then the following code could be simplified to:
```suggestion
      entries->put(path + path_separator + "has_been_set", true);
```

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2402,37 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+ParameterHandler::get_entries_wrongly_not_set() const
+{
+  std::set<std::string> entries_wrongly_not_set;
+
+  for (auto it = entries_set_status.begin(); it != entries_set_status.end();
+       ++it)
+    if (it->second.first == true && it->second.second == false)
+      entries_wrongly_not_set.insert(it->first);
+
+  return entries_wrongly_not_set;
```
peterrum: This could be replaced by a iteration over the tree.

include/deal.II/base/parameter_handler.h
```diff
@@ -1063,7 +1063,8 @@ class ParameterHandler : public Subscriptor
   declare_entry(const std::string &          entry,
                 const std::string &          default_value,
                 const Patterns::PatternBase &pattern = Patterns::Anything(),
-                const std::string &          documentation = "");
+                const std::string &          documentation = "",
+                const bool &                 has_to_be_set = false);
```
bangerth: no need for the reference

include/deal.II/base/parameter_handler.h
```diff
@@ -1127,7 +1128,8 @@ class ParameterHandler : public Subscriptor
                 ParameterType &              parameter,
                 const std::string &          documentation = "",
                 const Patterns::PatternBase &pattern =
-                  *Patterns::Tools::Convert<ParameterType>::to_pattern());
+                  *Patterns::Tools::Convert<ParameterType>::to_pattern(),
+                const bool &has_to_be_set = false);
```
bangerth: same here

include/deal.II/base/parameter_handler.h
```diff
@@ -1503,6 +1505,23 @@ class ParameterHandler : public Subscriptor
   bool
   operator==(const ParameterHandler &prm2) const;
 
+  /**
+   * Returns a set of names corresponding to those entries of the parameter
```
bangerth:
```suggestion
   * Return a set of parameter names corresponding to those entries of the parameter
```
Do these names also include subsection names?

include/deal.II/base/parameter_handler.h
```diff
@@ -1503,6 +1505,23 @@ class ParameterHandler : public Subscriptor
...
+   * handler that have not been set by one of the functions parsing parameters
+   * from an input file or by a set function, but that have been defined as
```
bangerth:
```suggestion
   * from an input file, or by an explicit call to one of the set() functions, but that have been declared as
```

include/deal.II/base/parameter_handler.h
```diff
@@ -1503,6 +1505,23 @@ class ParameterHandler : public Subscriptor
...
+   * mandatory parameters that must be set.
```
bangerth:
```suggestion
   * mandatory parameters that must be set through the last argument of the declare_parameter() function.
```

include/deal.II/base/parameter_handler.h
```diff
@@ -1503,6 +1505,23 @@ class ParameterHandler : public Subscriptor
...
+   */
+  std::set<std::string>
+  get_entries_wrongly_not_set() const;
+
+  /**
+   * Asserts that those entries of the parameter handler with flag
+   * has_to_be_set = true have been set. An exception is invoked
```
bangerth:
```suggestion
   * `has_to_be_set = true` have been set. An exception is invoked
```

include/deal.II/base/parameter_handler.h
```diff
@@ -1654,6 +1673,13 @@ class ParameterHandler : public Subscriptor
    */
   std::unique_ptr<boost::property_tree::ptree> entries;
 
+  /**
+   * A map that stores whether an entry added to the parameter handler has
+   * been set by any of the functions parsing input parameters or by a set
+   * function of this class.
+   */
+  std::map<std::string, std::pair<bool, bool>> entries_set_status;
```
bangerth: can you explain what the second part of the map is?

include/deal.II/base/parameter_handler.h
```diff
@@ -2230,6 +2258,20 @@ ParameterHandler::add_parameter(const std::string &          entry,
...
+ * The user has to specify whether parameters in the input file not added to the
+ * parameter handler will be skipped (enables partial parsing), and whether the
+ * code will assert that all parameters of the parameter handler are indeed
+ * found in the input file.
+ */
+void
+parse_input(std::string const parameter_file,
+            ParameterHandler &prm,
+            bool const        skip_undefined               = true,
+            bool const        assert_all_entries_are_found = true);
```
bangerth: Like I said before, let's move this into a separate PR and focus on one aspect at a time. I'm still of the opinion that this should be a member function :-)

(Why? Among other reasons because global functions are hard to find in the documentation. Another reason is that changes in the internal implementation of the `ParameterHandler` class's data structures and public interface can require adjustments in functions; for member functions, these are easy to make, but for global functions they are not.)

source/base/parameter_handler.cc
bangerth: If you do things here in the tree, then you won't need the `entries_set_status` object any more, right? You're just storing the same information twice -- that's a prescription for getting data structures that are out of sync.

source/base/parameter_handler.cc
```diff
@@ -1086,6 +1102,13 @@ ParameterHandler::set(const std::string &entry_string,
 
       // finally write the new value into the database
       entries->put(path + path_separator + "value", new_value);
+      entries->put(path + path_separator + "has_been_set", true);
+
+      entries_set_status.find(path)->second = std::pair<bool, bool>(
+        entries->get<std::string>(path + path_separator + "has_to_be_set") ==
+          "true",
```
bangerth: This is where the duplication of information happens.

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2402,37 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+}
+
+void
+ParameterHandler::assert_all_entries_have_been_set() const
+{
+  std::set<std::string> entries_wrongly_not_set =
+    this->get_entries_wrongly_not_set();
+
+  AssertThrow(
+    entries_wrongly_not_set.size() == 0,
```
bangerth: The fact that you can write the assertion in what is essentially a statement function suggests that it isn't very useful. I still think that we shouldn't have this function. One reason is that right now, you really only print one of the missing parameters, but that will lead to a long sequence of trial and error in practice where every time you call a program, you get told about yet another parameter that is missing. But you never get to see all at once. I don't think that's a useful design.

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2402,37 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
 }
 
 
+std::set<std::string>
+ParameterHandler::get_entries_wrongly_not_set() const
+{
+  std::set<std::string> entries_wrongly_not_set;
+
+  for (auto it = entries_set_status.begin(); it != entries_set_status.end();
+       ++it)
```
bangerth: You can do this more easily with a range-based for loop.

source/base/parameter_handler.cc
```diff
@@ -2622,4 +2676,35 @@ MultipleParameterLoop::Entry::memory_consumption() const
           sizeof(type));
 }
 
+void
+parse_input(std::string const parameter_file,
```
bangerth: separate patch, as mentioned above

source/base/parameter_handler.cc
nfehn: The first option means that ``has_been_set`` is set to true if ``has_to_be_set==false``, the second option that only some entries get the property ``has_been_set``?? In both  cases, the logic is not clear to me.

source/base/parameter_handler.cc
nfehn: Principally, I share the concern regarding the redundancy of data. However, implementing this iteration did not look so trivial to me when I had a look into the code. For me it is actually not worth implementing it but maybe it is much simpler than I thought. How many lines of code would that be? Question: Can you implement this without a copy and past of other code of the parameter handler and without investing a lot of effort into refactoring the ParameterHandler?

source/base/parameter_handler.cc
nfehn: Maybe we mix up different aspects here. Is the problem that it is a public function or would the problem be the same in case of a private one? The code provides the flexibility to do what you want. One can call the parse_input() function with assert_all_entries_have_been_set = false (=default), and subsequently call get_entries_wrongly_not_set() where the user can then print the whole list. For me it is ok if we print the whole list by default. The sentence with "trial and error" sounds a bit negative to me given that I want to improve the ParameterHandler in a way to avoid **errors**.

bangerth:
On 3/25/20 8:49 AM, Niklas Fehn wrote:
> Principally, I share the concern regarding the redundancy of data. However, 
> implementing this iteration did not look so trivial to me when I had a look 
> into the code. For me it is actually not worth implementing it but maybe it is 
> much simpler than I thought. How many lines of code would that be? Question: 
> Can you implement this without a copy and past of other code of the parameter 
> handler and without investing a lot of effort into refactoring the 
> ParameterHandler?

I don't actually know. I think all of the other places that walk the tree do 
so in a recursive way. That might be awkward.

If that's not easily feasible, just keep the global map and drop the local 
flags in the tree that check whether something has been set. Just don't 
duplicate :-)

source/base/parameter_handler.cc
nfehn: I would like to have the discussion here because it is related to the design questions raised above, i.e., this function calls the ``assert_()`` function. We could then still merge as a separate PR?

source/base/parameter_handler.cc
bangerth: The way I approach this is like this:

* Every addition of a function has a cost. It has to be maintained, its documentation has to be kept up to date, it expands the number of functions users have to look through to find what they are looking for, it tempts users to use this function even in cases where it doesn't quite do what they need and then prompts them to write work-arounds, etc. So adding a function is not free.

* This has to be balanced with the question of whether the function (or a variation of it) is *useful*. 

I claim that that the `assert` function is *not actually useful enough* to justify the cost. That's because what it does is trigger an assertion that only shows one parameter that isn't set. But if multiple parameters aren't set in an input file, you still only get to see one of them in the error message. So if you call your application with an input file that misses multiple parameters, you get to see one missing parameter; you fix it; you call the application again; you get another error; you fix that; call the application again; etc. 

A better error message would probably list all critical-but-missing parameters at once. I suppose that I could accept such a function -- it's useful, and I understand your use case and why you want to have it. It's mainly that we don't have these kinds of functions in the library very much, but maybe one can argue that such a function would really not be all that different from having assertions that the input to a `ParameterHandler` must be a well-formed file that follows a certain syntax -- in your case, the syntax just happens to include a clause that certain elements need to be present.

nfehn: For example, @peterrum wants to have the Assert in the other ``parse_input_...()`` functions as well (which is more than one), so doesn't this already imply that we need an assert-function. Since the plan is to make the ``parse_input()`` function of this PR a member function, the assert-function could then become a private member function. Would this be a compromise?

bangerth: I disagree with @peterrum on this :-) I don't think the assertion should be in there because there are use cases where people parse several `.prm` files one after the other. One of them isn't going to fill all parameters, and so the assertion would be misplaced there.

nfehn: I understand your point. At the same time it seems to me that this is not how the ParameterHandler is currently designed. For example, if a parameter does not match the predefined pattern, we currently also have to "debug" through the code in a "last man standing" style. I think the code throws after the first parameter has been found that does not match its pattern. 

A second point is that, for my understanding, the cost of a function is always justified if it avoids code duplication. I would say that in an application code you need around 10-20 lines to implement a routine that gets the set from the ParameterHandler, then outputs the missing parameters, and finally throws an exception. Every application code will have to do that, in reality some application codes will probably implement it more than once. My point of view is probably more application-oriented, but wouldn't it make sense that this functionality is already part of deal.II.

peterrum: @bangerth What I suggested yesterday is to introduce to the `parse_input_*`-functions an additional optional parameter (maybe with the name) `assert_all_parameters_have_been_set`. The reason why I suggested it is that I have the feeling that in many cases - especially if you work with a single parameter file - both undefined entries and parameters not set are unwanted behavior and might be as such asserted.

The usage could be in the case that only one input file is used:
```cpp
prm.parse_input (file, ..., true);
```
and in the case of multiple input files:
```cpp
prm.parse_input (file1, ..., false);
prm.parse_input (file2, ..., false);
prm.parse_input (file3, ..., false);
prm.assert_that_parameters_have_been_set ();
```

A function whose only purpose is to assert is an option. An other option would be to introduce a new exception type:
```cpp
Assert(prm.are_all_parameters_set(),
        Exception::NotSetParameter(prm.get_any_not_set_parameter()) );
```

It is some copy and paste, but not that much. What do you think @nfehn?

nfehn: Yes, this is what I suggested (the last option is not my favorite due to DRY, see above). I mean, no one is forced to run into the assert. We can set the default value to false so that nothing happens. Unfortunately, we need a default value because the ``parse_input()`` functions already have something like ``skip_undefined`` with default value and we now can not change this backwardly. Theoretically, the best design would certainly be to give the user all the freedom and, at the same time, not fool the user by default values that lead to erroneous results because the user has not been aware of them.

nfehn:
@bangerth All missing entries are now printed, parse_input() function has been removed from this PR, see

https://github.com/dealii/dealii/pull/9718/commits/60819472ba74905f6176f40307942b6eaf544fb2

https://github.com/dealii/dealii/pull/9718/commits/bb60477e1bc1946764a74e25fe9667a56a1fb2a0

include/deal.II/base/parameter_handler.h
nfehn: Sorry, I have not seen this question so far. I adjusted the comment.

include/deal.II/base/parameter_handler.h
nfehn: has been fixed

nfehn:
I had again a look at the implementation and simplified it. @bangerth Maybe you could re-review all these changes?

include/deal.II/base/parameter_handler.h
```diff
@@ -1063,7 +1063,8 @@ class ParameterHandler : public Subscriptor
   declare_entry(const std::string &          entry,
                 const std::string &          default_value,
                 const Patterns::PatternBase &pattern = Patterns::Anything(),
-                const std::string &          documentation = "");
+                const std::string &          documentation = "",
+                const bool                   has_to_be_set = false);
```
bangerth, line 1067: You'll have to add documentation for this parameter. (There is also text above that talks about "The last parameter..." which is now no longer correct.)

include/deal.II/base/parameter_handler.h
```diff
@@ -1127,7 +1128,8 @@ class ParameterHandler : public Subscriptor
                 ParameterType &              parameter,
                 const std::string &          documentation = "",
                 const Patterns::PatternBase &pattern =
-                  *Patterns::Tools::Convert<ParameterType>::to_pattern());
+                  *Patterns::Tools::Convert<ParameterType>::to_pattern(),
+                const bool has_to_be_set = false);
```
bangerth, line 1132: same here

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2396,44 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+  std::set<std::string> entries_wrongly_not_set;
+
+  for (auto it : entries_set_status)
+    if (it.second.first == true && it.second.second == false)
+      entries_wrongly_not_set.insert(it.first);
+
+  return entries_wrongly_not_set;
+}
+
+void
```
bangerth, line 2411:
```suggestion


void
```

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2396,44 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+ParameterHandler::assert_that_entries_have_been_set() const
+{
+  std::set<std::string> entries_wrongly_not_set =
```
bangerth, line 2414:
```suggestion
  const std::set<std::string> entries_wrongly_not_set =
```

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2396,44 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+    this->get_entries_wrongly_not_set();
+
+  if (entries_wrongly_not_set.size() > 0)
+    {
+      std::string list_of_missing_parameters = "\n\n";
+      for (auto it : entries_wrongly_not_set)
```
bangerth, line 2420:
```suggestion
      for (const auto &it : entries_wrongly_not_set)
```

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2396,44 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+        list_of_missing_parameters += it + "\n";
+      list_of_missing_parameters += "\n";
+
+      AssertThrow(
+        entries_wrongly_not_set.size() == 0,
+        ExcMessage(
+          "Not all entries of the parameter handler with flag has_to_be_set = true "
```
bangerth, line 2427:
```suggestion
          "Not all entries of the parameter handler that were declared with 'has_to_be_set = true' "
```

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2396,44 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+          " have been set. The following parameters " +
+          list_of_missing_parameters +
+          " have not been set. "
+          "A possible reason might be that you did not add this parameter to the "
```
bangerth, line 2431:
```suggestion
          "A possible reason might be that you did not add these parameters to the "
```

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2396,44 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+          "input file or that the spelling of this parameter is not correct."));
```
bangerth, line 2432:
```suggestion
          "input file or that their spelling is not correct."));
```

source/base/parameter_handler.cc
```diff
@@ -2379,6 +2396,44 @@ ParameterHandler::operator==(const ParameterHandler &prm2) const
...
+ParameterHandler::assert_that_entries_have_been_set() const
+{
+  std::set<std::string> entries_wrongly_not_set =
+    this->get_entries_wrongly_not_set();
+
+  if (entries_wrongly_not_set.size() > 0)
+    {
+      std::string list_of_missing_parameters = "\n\n";
+      for (auto it : entries_wrongly_not_set)
+        list_of_missing_parameters += it + "\n";
```
bangerth, line 2421: Let's put a couple of spaces in front to offset them from the surrounding text.
```suggestion
        list_of_missing_parameters += "   " + it + "\n";
```

nfehn:
A test is a good idea. I added one. I also addressed the minor comments.

nfehn:
I guess this https://github.com/dealii/dealii/pull/9718/commits/365f339843c120310d330fd1ac88353d460eaf90 is what you want to see

peterrum:
> I guess this 365f339 is what you want to see

Have you run the tests? Are all passing?

nfehn:
100% tests passed, 0 tests failed out of 225

bangerth:
This passes all testers except for the "travis/ci" one. I would have merged, except that's a "required" check. How do we deal with travis being unreliable these days?

masterleinad:
> This passes all testers except for the "travis/ci" one. I would have merged, except that's a "required" check. How do we deal with travis being unreliable these days?

I would vote for not making `Travis` required anymore since the `workflow` runs also check indentation and are always running. Otherwise, we need to rebase until it finally reports its status correctly.

bangerth:
@nfehn -- can you try to rebase and force push this PR so that we can finally push it across the finish line?

nfehn:
I don't know what you mean exactly. I thought I have done this already?

peterrum:
> I don't know what you mean exactly. I thought I have done this already?

For some reason the testing system had some problems the last few days. But it seem to back to normal again. You only need to rebase your branch with the current master, then all tests are restarted again, and then this PR should good to go.

nfehn:
Can you please write the commit that you want to have for rebasing or what you mean exactly?

kronbichler:
The exact commit does not matter - it just needs a notice of the difference.
What I always do would correspond to the following steps (there are also several alternatives):
```
git checkout master
git pull
git checkout parameter_handler
git rebase master
git pull nfehn --force
```
The important part is the `git rebase master` command - here you need to look out for possible merge conflicts, but unless someone else modified the same lines as you do here in the meantime (between the time you branched off from master and the one you pulled in), it should not happen.

nfehn:
In the end you mean ``git push nfehn parameter_handler --force`` I guess?

kronbichler:
Yes, that is the complete line, but depending on how you set up git and your branch it might automatically select that remote branch.

nfehn:
I meant the ``push`` in the end but it seems like something happened now ;)

bangerth:
One test only is still not finished, and it's a different one than the one that was previously hanging -- so I'll assume that everything is well with this PR and merge.

PR #9717, opened 2020-03-23
Add a graph to matrix-free module

peterrum:
Preview of the graph:

Screenshot from 2020-03-23 08-40-23

related to #9687.

PR #9719, opened 2020-03-23
Fix MatrixFree graph

peterrum:
fixes https://github.com/dealii/dealii/pull/9717#issuecomment-602664114

@bangerth Could you verify that this patch is working (on my machine I don't have any problems). Thanks!

### DIFF
--- a/doc/doxygen/headers/matrixfree.h
+++ b/doc/doxygen/headers/matrixfree.h
@@ -40,7 +40,7 @@ digraph G
   dh         [label="DoFHandler",       URL="\ref dofs"];
   subgraph linalglibs {
     rank="same";
-    simd     [label="SIMD",     URL="\ref CUDAWrappers", fontname="FreeSans",fontsize=12,
+    simd     [label="SIMD", fontname="FreeSans",fontsize=12,
               shape=record,height=0.2,width=0.4,
               color="gray", fontcolor="gray", fillcolor="white", style="filled"];
     fevalues [label="FEEvaluation"];
@@ -81,12 +81,7 @@ digraph G
   subgraph misclibs {
   systems    [label="Operators"];
   }
-  simd        -> fevalues [dir="none", color="transparent"];
-  fevalues    -> mf       [dir="none", color="transparent"];
-  mf          -> cuda     [dir="none", color="transparent"];
-  cuda        -> tbb      [dir="none", color="transparent"];
   opencascade -> manifold [dir="none"];
-  gmg         -> solvers  [dir="none", style="dashed", color="black"];
 
   node [fontname="FreeSans",fontsize=12,
         shape=ellipse,height=0.2,width=0.4,


bangerth:
No, that doesn't seem to work. I think you should google the error message:
https://www.google.com/search?q=flat+edge+between+adjacent+nodes+one+of+which+has+a+record+shape&oq=flat+edge+between+adjacent+nodes+one+of+which+has+a+record+shape&aqs=chrome..69i57j0.537j0j7&sourceid=chrome&ie=UTF-8
I don't have an idea what is going wrong with the graph, but I have confirmed that if I copy the input between `@dot` and `@enddot` into a file `x.dot` and run `dot x.dot -T png > x.png` that I get the same error. This provides for a simpler way to reproduce the issue.

bangerth:
By the way, here is the minimal testcase:
```
digraph G
{
  subgraph linalglibs {
    rank="same";
    simd     [shape=record];
    fevalues;
  }
  simd -> fevalues        [dir="none", color="transparent"];
}
```
with error
```
1/build-doc> dot x.dot -T png > x.png
Warning: flat edge between adjacent nodes one of which has a record shape - replace records with HTML-like labels
  Edge simd -> fevalues
Error: lost simd fevalues edge
```
I think the issue is that you say that the two nodes must by at the same level. Maybe just let the SIMD node float up by one level?

peterrum:
@bangerth Thank for testing and sending me a minimal test case. However, I do not got any problems on my computer. Could you try out the latest patch?

This is equivalent to:
```
digraph G
{
  subgraph linalglibs {
    rank="same";
    simd     [shape=record];
    fevalues;
    simd -> fevalues        [dir="none", color="transparent"];
  }
}
```

peterrum:
And I have also fixed https://github.com/dealii/dealii/pull/9717#discussion_r396591692.

kronbichler:
I think I can reproduce this on my Ubuntu system - the new variant does not solve it. What I need to do is to remove all edges between "record boxes" and the regular ones. I will mark up what needs to be removed. I think the end result is pretty nice anyway (we can live with the fact that TBB sits on the same side as SIMD), so I'd not pour more time into this now:
x

doc/doxygen/headers/matrixfree.h
```diff
@@ -51,6 +51,10 @@ digraph G
     tbb      [label="TBB",     URL="\ref CUDAWrappers", fontname="FreeSans",fontsize=12,
               shape=record,height=0.2,width=0.4,
               color="gray", fontcolor="gray", fillcolor="white", style="filled"];
+  simd -> fevalues        [dir="none", color="transparent"];
+  fevalues -> mf          [dir="none", color="transparent"];
+  mf -> cuda              [dir="none", color="transparent"];
+  cuda -> tbb             [dir="none", color="transparent"];
```
kronbichler: Remove these

doc/doxygen/headers/matrixfree.h
```diff
@@ -81,12 +85,8 @@ digraph G
   subgraph misclibs {
   systems    [label="Operators"];
   }
-  simd        -> fevalues [dir="none", color="transparent"];
-  fevalues    -> mf       [dir="none", color="transparent"];
-  mf          -> cuda     [dir="none", color="transparent"];
-  cuda        -> tbb      [dir="none", color="transparent"];
   opencascade -> manifold [dir="none"];
-  gmg         -> solvers  [dir="none", style="dashed", color="black"];
+  gmg -> solvers          [dir="none", style="dashed", color="black"];
```
kronbichler: Remove this line

peterrum: Thanks for the fix and the offline tests! I am not completely happy with the graph since the nodes `Triangulation`, `FiniteElement`, `Mapping`, and `Quadrature` are mirrored in comparison to the matrix-based path. But I guess this is fine for now to be merged. 

I will install a different version of dot and try to find a solution which works. My version is:
```
> dot -V
dot - graphviz version 2.30.1 (20180420.1509)
```

bangerth:
On 3/24/20 3:34 AM, Peter Munch wrote:
> |> dot -V dot - graphviz version 2.30.1 (20180420.1509) |

I have 2.40 on my system.

kronbichler:
> I have 2.40 on my system.

Same here.

peterrum:
Now it is visible :) https://www.dealii.org/developer/doxygen/deal.II/group__matrixfree.html

Thanks @bangerth abnd @kronbichler !

bangerth:
Great, thanks for working through the details!

bangerth:
If you wanted to add one more tweak: Highlight the boxes in some way that are relevant for the particular module in question here. Maybe with blue text and box, or with a non-white background?

bangerth:
This would make it clearer where in the graph the current module is actually located.

PR #9727, opened 2020-03-24
Add improvements to the mf path graph

peterrum:
Next attempt. I have installed dot version 2.40.

path

This PR:
* replaces `shape=record` by `shape=box` so that I can add back the transparent lines again and
* adds some color (a long list of possible colors cab be found here https://www.graphviz.org/doc/info/colors.html)


kronbichler:
@peterrum we need a rebase here to get Travis CI alive again before we can merge.

PR #9731, opened 2020-03-25
Introduce new namespace MatrixFreeTools

peterrum:
As proposed in https://github.com/dealii/dealii/pull/9343#issuecomment-600338945. Once we have agreed on the name of the namespace and the file location and this PR has been merged, I will move the content of #9343 into the new namespace.

### DIFF
--- a/include/deal.II/matrix_free/utilities.h
+++ b/include/deal.II/matrix_free/utilities.h
@@ -1,0 +1,28 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2020 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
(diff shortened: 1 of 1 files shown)


tjhei:
The name ``MatrixFreeTools`` would nicely mirror ``GridTools``, ``DoFTools``, etc. .

peterrum:
`MatrixFreeTools` sounds good! But I'll wait another couple of hours before updating. Maybe there some other ideas!

nfehn:
In case of the ``Utilities`` naming, something like ``Utilities::MatrixFree`` might be better? By the way, this is not a vote against ``MatrixFreeTools``.

peterrum:
There have been 3 votes for `MatrixFreeTools`! I have changed the name of the namespace. Can we merge it like this?

PR #9734, opened 2020-03-25
Remove hp::DoFHandler::get_fe()

peterrum:
This PR removes the deprecated method `hp::DoFHandler::get_fe()`. I guess we can do this since the function has been deprecated more than two years ago.

Do you agree @bangerth and @marcfehling?

### DIFF
--- a/doc/news/changes/incompatibilities/20200326PeterMunch
+++ b/doc/news/changes/incompatibilities/20200326PeterMunch
@@ -1,0 +1,4 @@
+Removed: The deprecated method hp::DoFHandler::get_fe() has been removed. Please use
+hp::DoFHandler::get_fe_collection() instead.
+<br>
+(Peter Munch, 2020/03/26)
--- a/include/deal.II/numerics/vector_tools.templates.h
+++ b/include/deal.II/numerics/vector_tools.templates.h
@@ -286,7 +286,7 @@ namespace VectorTools
                 const ComponentMask &                component_mask)
     {
       Assert(component_mask.represents_n_components(
-               dof_handler.get_fe().n_components()),
+               dof_handler.get_fe_collection().n_components()),
              ExcMessage(
                "The number of components in the mask has to be either "
                "zero or equal to the number of components in the finite "
@@ -296,7 +296,7 @@ namespace VectorTools
              ExcDimensionMismatch(vec.size(), dof_handler.n_dofs()));
 
       Assert(component_mask.n_selected_components(
-               dof_handler.get_fe().n_components()) > 0,
+               dof_handler.get_fe_collection().n_components()) > 0,
              ComponentMask::ExcNoComponentSelected());
 
       //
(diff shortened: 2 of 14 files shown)


include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -277,16 +277,21 @@ namespace VectorTools
               int spacedim,
               typename VectorType,
               template <int, int> class DoFHandlerType,
-              typename T>
+              typename T,
+              typename F>
```
bangerth: can you give this argument a more descriptive template type name? Maybe `FEorFECollectionType`?

bangerth: On the other hand, why do you need this? Regular `DoFHandler` has a function `get_fe_collection()` so that's a function you can generically call...

peterrum: @bangerth Right! I was happy that after pages of compiler errors it compiled ;)

peterrum:
@bangerth @masterleinad Thanks for the review! I have made the changes.

masterleinad:
Did you actaully remove the member function?

doc/news/changes/incompatibilities/20200326PeterMunch
```diff
@@ -0,0 +1,4 @@
+Removed: The depricated method hp::DoFHandler::get_fe() has been removed. Please use
```
marcfehling:
```suggestion
Removed: The deprecated method hp::DoFHandler::get_fe() has been removed. Please use
```

include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -553,21 +607,7 @@ namespace VectorTools
     VectorType &                                               vec,
     const ComponentMask &                                      component_mask)
   {
-    Assert(dof_handler.get_fe().n_components() == function.n_components,
```
marcfehling: Do you actually need to introduce your new two `internal::interpolate(...)` functions? Both `DoFHandler` and `hp::DoFHandler` have the member function `get_fe_collection()`.

I guess if you just replace `dof_handler.get_fe()` with `dof_handler.get_fe_collection()` in this assertion, you should be fine!

include/deal.II/numerics/vector_tools.templates.h
```diff
@@ -553,21 +607,7 @@ namespace VectorTools
...
-           ExcDimensionMismatch(dof_handler.get_fe().n_components(),
```
bangerth: I don't quite understand why you need to duplicate this function into one for the hp and one for the non-hp case. Can't you just replace this call to `get_fe()` by `get_fe_collection()`? That works for both `::DoFHandler` and `hp::DoFHandler`. If you did that, then you wouldn't have to duplicate the body of the function...

peterrum:
> Did you actaully remove the member function?

Oops! Not it is removed. I removed the function on an other branch, where I am working of `DoFHandler` and `hp::DoFHandler`.

include/deal.II/numerics/vector_tools.templates.h
peterrum: Done!

masterleinad:
@bangerth still needs to approve.

masterleinad:
Seems like you missed one test:
```
/jenkins/workspace/dealii_PR-9734/tests/dofs/dof_tools_12.cc:31:68: error: no matching function for call to ‘dealii::hp::DoFHandler<1, 1>::get_fe() const’
   std::vector<bool> mask(dof_handler.get_fe().n_components(), false);
```

PR #9735, opened 2020-03-25
Change return type of DoFHandler::n_locally_owned_dofs()

peterrum:
I would like to make the return type of this function consistent with the return type of `hp::DoFHandler::n_locally_owned_dofs()`.


include/deal.II/dofs/dof_handler.h
```diff
@@ -1023,7 +1023,7 @@ class DoFHandler : public Subscriptor
    * cells owned by other processors may be theirs, and degrees of freedom on
    * ghost cells are also not necessarily included.
    */
-  unsigned int
+  types::global_dof_index
```
bangerth: Ouch -- probably not wrong, but definitely an oversight...

PR #9739, opened 2020-03-26
GridTools/GridGenerator: add some functions to support dim-independent programming

nfehn:
The functions ``rotate()`` and ``extrude_triangulation()`` are currently only implemented for dim=2 or 3. This PR allows to write code that can be compiled for dim=2,3.

### DIFF
--- a/include/deal.II/grid/grid_generator.h
+++ b/include/deal.II/grid/grid_generator.h
@@ -1756,6 +1756,22 @@ namespace GridGenerator
     const std::vector<types::manifold_id> &manifold_priorities = {});
 
   /**
+   * Overload of extrude_triangulation() to allow dimension independent
+   * code to compile. This function throws an error when called, as
+   * extrude_triangulation() is only implemented to extrude a dim=2 to a dim=3
+   * Triangulation.
+   */
+  void
+  extrude_triangulation(
+    const Triangulation<2, 2> &            input,
+    const unsigned int                     n_slices,
+    const double                           height,
+    Triangulation<2, 2> &                  result,
+    const bool                             copy_manifold_ids   = false,
+    const std::vector<types::manifold_id> &manifold_priorities = {});
+
+
+  /**
    * Overload of the previous function. Take a 2d Triangulation that is being
    * extruded. Differing from the previous function taking height and number of
    * slices for uniform extrusion, this function takes z-axis values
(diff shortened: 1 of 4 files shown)


drwells:
The changes to rotate make sense but I'm not so sure about extrusion. What case did you find where we need an additional extrusion overload?

nfehn:
I get this error message

error: no matching function for call to extrude_triangulation(dealii::Triangulation<2, 2>&, unsigned int, const double&, dealii::Triangulation<2, 2>&)

drwells:
Sure, but what does the call site look like?

nfehn:
something like this
``
Triangulation<2> tria_2d;
Triangulation<dim> tria_dim;
GridGenerator::extrude_triangulation(tria_2d, foo, bar, tria_dim);
``

nfehn:
@drwells Did I answer your questions? Please let me know if I still misunderstood something.

include/deal.II/grid/grid_tools.h
```diff
@@ -454,8 +454,9 @@ namespace GridTools
    * the transform() function above, so the requirements on the triangulation
    * stated there hold for this function as well.
    */
+  template <int dim>
```
tjhei: can you please leave a ``@note`` here that this function is only supported for dim==2?

nfehn: I changed this. I have written only one ``=`` for consistency with the function below where it is written similarly.

drwells:
Sorry, I meant to get back to you on this. I think this is fine but we need to add documentation explaining the extrude is only available for 2D to 3D.

tjhei:
> I think this is fine

I also think that it is a bit weird but it does help make user code much simpler without having to rely on ``constexpr if`` (my code that uses extrude() always contains specialization based on dimension because of this).

include/deal.II/grid/grid_generator.h
```diff
@@ -1755,6 +1755,16 @@ namespace GridGenerator
     const bool                             copy_manifold_ids   = false,
     const std::vector<types::manifold_id> &manifold_priorities = {});
 
+  void
+  extrude_triangulation(
```
tjhei: Can you add a doxygen comment here please with something like:

Overload of extrude_triangulation() to allow dimension independent code to compile. This function throws an error when called as extrude_triangulation() is only implemented to extrude a dim=2 to a dim=3 Triangulation.

nfehn: I added some documentation. I thought `specialization` might be more appropriate than overload.

drwells: These are technically overloads since they do not have any template parameters :)

nfehn: Technically I agree, but I would say the way it is used is rather in a sense of template specialization. Should I change it to overload then?

drwells: To be clear on the resolution of this thread: the current text is overload so we're good :+1:

PR #9754, opened 2020-03-28
Also output the current time in .vtu files in step-52.

### LINKED ISSUE #9722: Time values of write_vtu seems to be incorrect?
I am raising this issue based on my experience with step-52 tutorial which solves a time-dependent PDE (diffusion equation) using a direct solver on a 2D square mesh with zero Dirichelet BCs (at the left and right edges), and zero Neumann BCs (at the top and bottom edges). 

I think there is something wrong with how the time values of the solution are written by .vtu files.
Just to highlight the issue, for now let us focus on files produced by the backward-euler  and dopri solvers. For the backward Euler run, the tutorial produces __20__ files, named `solution-backward_euler-000.vtu` through `solution-backward_euler-200.vtu`.  For the Dopri run, the tutorial produces __2790__ files, named `solution-dopri-000.vtu` through `solution-dopri-2790.vtu`.   

However, the end time of the simulation (t_final) is actually __10 seconds__.   After loading the backward Euler results into Paraview, the animation control shows the time as 0-20 seconds, while the time-range for the Dopri results are shows as 0-279 seconds. They both are incorrect. Naturally, this incorrectness also happens when using a "time annotation" on the view port.  Furthermore, the non-integer continuous time is not shown at all. This renders the communication of the results over the actual simulation time impossible. 

The filename is clearly appended with the time-step number (at every 10th time-step). Since, it shows the time-step as 1/10th of the value of its corresponding filename, how is Paraview getting this `mod 10` calculation? i.e. how is the time-step encapsulated into the solution vector?  

I thought the 'solution' variable is just a standard Vector type which cannot really hold 'time' anywhere. But the real-requirement is that we must indeed inform 'time' to the .vtu file somehow.  Also, interestingly step-26 which is a very similar problem (using an iterative solver) does not suffer from this issue.

Is this a bug in __write_vtu()__?

bangerth:
Fixes #9722.

/rebuild

### DIFF
--- a/doc/news/changes/minor/20200327Bangerth
+++ b/doc/news/changes/minor/20200327Bangerth
@@ -1,0 +1,4 @@
+Fixed: step-52 now also outputs the time (and not just the time step)
+in the VTU output files.
+<br>
+(Wolfgang Bangerth, 2020/03/27)
--- a/examples/step-52/step-52.cc
+++ b/examples/step-52/step-52.cc
@@ -94,7 +94,8 @@ namespace Step52
                                           const double          tau,
                                           const Vector<double> &y);
 
-    void output_results(const unsigned int               time_step,
+    void output_results(const double                     time,
+                        const unsigned int               time_step,
                         TimeStepping::runge_kutta_method method) const;
 
     // The next three functions are the drivers for the explicit methods, the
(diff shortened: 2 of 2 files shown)


krishnakumarg1984:
@bangerth this probably needs to be reopened again since the fix does not seem to work for the .vtu files produced by the embedded Runge Kutta methods.

PR #9755, opened 2020-03-28
Copy pieces of the module graph into some of the modules.

bangerth:
For some modules, this patch copies a subset of the big module graph into the
module documentation, with the current module highlighted. This makes it
easier to see the connections between modules not just from the documentation
front page, but also from the individual modules.

This is analogous to #9717 and a follow-up to #9687.

### DIFF
--- a/doc/doxygen/headers/dofs.h
+++ b/doc/doxygen/headers/dofs.h
@@ -60,4 +60,60 @@
  * such a way that the bandwidth of associated matrices is minimized. Finally,
  * the DoFTools namespace offers a variety of algorithms around handling
  * degrees of freedom.
+ *
+ * In the grand scheme of things, the pieces of this module interact
+ * with a variety of other parts of the library:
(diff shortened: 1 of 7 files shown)


bangerth:
On 3/28/20 5:06 AM, Peter Munch wrote:
> I far as I see the only difference between the pictures are the highlighted box.
Yes -- these are the core components of the library, so they all got the same 
graph, just different highlighting. I was going to follow up for the remaining 
modules with more customized graphs.

peterrum:
@bangerth Would you mind if I fix the order of `Mapping`, `Quadrature`, `Finite elements`, and `Triangulation` so that it is consistent with the main page? Also why did you keep `OpenCASCADE` and `gmsh` in pictures? I would suggest to keep the first one only in `Manifold` and the second one in `Triangulation` module page.

bangerth:
@peterrum -- yes to all!

PR #6431, opened 2018-05-03
LinearOperator: Implement a mean value filter operator

tamiko:
Closes #6332

PR #5352, opened 2017-10-28
Deprecate BlockTrianglePrecondition and BlockMatrixArray.

drwells:
Closing #3733 reminded me that we still need to deprecate these two classes.

I made some small improvements to the documentation along the way: in particular, I found it hard to figure out how to set up a `BlockLinearOperator` with multiple matrix types so I added a little example up front.

PR #9761, opened 2020-03-28
Delete BlockMatrixArray and MeanValueFilter

masterleinad:
Deprecated in #5352 and #6431.


examples/doxygen/block_matrix_array.cc
```diff
@@ -1,111 +0,0 @@
-// ---------------------------------------------------------------------
```
bangerth: Oh nice -- we should just empty that whole directory!

PR #9764, opened 2020-03-28
Template Utilities::pow()

### LINKED ISSUE #9762: Create a templated version of `Utilities::pow()`
```cpp
   template<typename T>
   constexpr T
   pow(const T base, const int iexp);
```

peterrum:
closes #9762

### DIFF
--- a/include/deal.II/base/utilities.h
+++ b/include/deal.II/base/utilities.h
@@ -480,11 +480,12 @@ namespace Utilities
 
   /**
    * A replacement for <code>std::pow</code> that allows compile-time
-   * calculations for constant expression arguments. The exponent @p iexp
-   * must not be negative.
+   * calculations for constant expression arguments. The @p base must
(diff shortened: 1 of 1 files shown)


include/deal.II/base/utilities.h
```diff
@@ -483,8 +483,9 @@ namespace Utilities
    * calculations for constant expression arguments. The exponent @p iexp
    * must not be negative.
    */
-  constexpr unsigned int
-  pow(const unsigned int base, const int iexp)
+  template <typename T>
+  constexpr T
+  pow(const T base, const int iexp)
```
kronbichler: Please add a comment that `T` must be an integral type (otherwise the initialization `1` might not work) and the case `pow <= 0 -> 1` is not true; it would be best if you add `static_assert(std::is_integral<T>::value, "Only integral types supported")` here.

peterrum: Thank! Done.

PR #9767, opened 2020-03-29
Add output variants for new MF::MappingInfo test

kronbichler:
In #9656 I extended a test for the compression of the mapping. While the output file is correct for AVX and SSE2 (2 cells in horizontal direction trigger the compression), it is not on AVX-512 and without vectorization. In the former, we have 4 cells together in horizontal direction for 2D, so only half as much output. Without vectorization and the full mapping, we need twice as many variants as there is only a single cell.

### DIFF
--- a/tests/matrix_free/compress_mapping.output.avx512
+++ b/tests/matrix_free/compress_mapping.output.avx512
@@ -1,0 +1,21 @@
+
+DEAL:2d::General mesh
+DEAL:2d::OK
+DEAL:2d::Hyper cube
+DEAL:2d::OK
+DEAL:2d::Parallelogram
+DEAL:2d::OK
+DEAL:2d::Deformed hyper cube
+DEAL:2d::Number of different Jacobians: 4
+DEAL:2d::Number of different Jacobians: 36
+DEAL:2d::OK
+DEAL:3d::General mesh
+DEAL:3d::OK
+DEAL:3d::Hyper cube
+DEAL:3d::OK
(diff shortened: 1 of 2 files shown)


tests/matrix_free/compress_mapping.output.avx512
```diff
@@ -0,0 +1,21 @@
+
+DEAL:2d::General mesh
+DEAL:2d::OK
+DEAL:2d::Hyper cube
+DEAL:2d::OK
+DEAL:2d::Parallelogram
+DEAL:2d::OK
+DEAL:2d::Deformed hyper cube
+DEAL:2d::Number of different Jacobians: 4
+DEAL:2d::Number of different Jacobians: 36
```
kronbichler: These two lines differ from the already present output.

tests/matrix_free/compress_mapping.output.novec
```diff
@@ -0,0 +1,21 @@
+
+DEAL:2d::General mesh
+DEAL:2d::OK
+DEAL:2d::Hyper cube
+DEAL:2d::OK
+DEAL:2d::Parallelogram
+DEAL:2d::OK
+DEAL:2d::Deformed hyper cube
+DEAL:2d::Number of different Jacobians: 8
+DEAL:2d::Number of different Jacobians: 144
```
kronbichler: This line and line 20 differ from the output already present.

kronbichler:
Can we merge this, given it is only two output variants?

PR #9771, opened 2020-03-29
Fix broken documentation.

bangerth:
This was apparently mangled when we ran clang-indent the first time.
/rebuild


kronbichler:
Documentation only, so merge.

PR #9772, opened 2020-03-29
Further try and disentangle the tutorial graph a bit.

bangerth:
Description in the long comment in the .pl file.

This is how this looks with this patch (compare with https://dealii.org/developer/doxygen/deal.II/Tutorial.html):
image


tjhei:
Looks good, I like it! :+1:

PR #9773, opened 2020-03-29
Fix order of module graphs

peterrum:
As discussed in https://github.com/dealii/dealii/pull/9755#issuecomment-605658821.

### DIFF
--- a/doc/doxygen/headers/dofs.h
+++ b/doc/doxygen/headers/dofs.h
@@ -69,7 +69,7 @@
   graph[rankdir="TB",bgcolor="transparent"];
 
   node [fontname="FreeSans",fontsize=15,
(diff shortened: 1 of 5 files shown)


doc/doxygen/headers/dofs.h
```diff
@@ -84,6 +84,12 @@
   output     [label="Graphical output", URL="\ref output"];
   manifold   [label="Manifold",         URL="\ref manifold"];
 
+{rank=same
+  mapping -> quadrature        [dir="none", color="transparent"];
```
bangerth: Pretty unimportant, but we're all type A here: Can you indent these sub-graphs and put the `rank=same` on a line of its own?

peterrum: Better?

PR #9779, opened 2020-03-30
Remove Multigrid::set_debug()

### LINKED ISSUE #9768: Multigrid test failures
After #9750 (?), we have some test failures in `mpi/multigrid_uniform` and `mpi/multigrid_adaptive`:
https://cdash.43-1.org/viewTest.php?onlydelta&buildid=5290
@masterleinad can you have a look?

masterleinad:
Fixes #9768. The function was deprecated in https://github.com/masterleinad/dealii/commit/bfc5899bf19593d688056ba309211ab974f311b0.
In the `cycles` test, the F-cycle was previously printing `V-cycle` for the last iteration but is now printing `F-cycle` instead.


include/deal.II/multigrid/multigrid.templates.h
```diff
@@ -215,72 +168,40 @@ template <typename VectorType>
...
-            << std::endl;
-
   if (level == minlevel)
     {
-      if (debug > 0)
-        deallog << cychar << "-cycle coarse level    " << level << std::endl;
-
+      this->signals.coarse_solve(true, level);
       (*coarse)(level, solution[level], defect2[level]);
+      this->signals.coarse_solve(false, level);
```
bangerth: Had we just forgotten to call all of these signals? They appear to have existed, after all...

tjhei: yes, I think I only tested it for the V cycle before.

masterleinad: Yes, we treat the V-cycle in `level_v_step` separately and only called the signals there.

PR #9781, opened 2020-03-30
Fix a bug in mg ghost layer with periodicity and n_levels=1

kronbichler:
This PR fixes a subtle bug in the setup of the level subdomain ids in the ghost layer of the multigrid levels: If we have a mesh with one single level (i.e., not refined, and a case where multigrid may not make much sense), we forgot to identify cells behind periodic boundaries as level ghost cells; this makes downstream multigrid data structures to break down.

The fix is to move the code that sets the periodic data structures before we call
https://github.com/dealii/dealii/blob/a0a79b2d3df7bfd3bb0530dccf557fea1516c680/source/distributed/tria.cc#L4727
Otherwise the code inside this call, namely the one that identifies the appropriate cells and vertices here:
https://github.com/dealii/dealii/blob/a0a79b2d3df7bfd3bb0530dccf557fea1516c680/source/distributed/tria.cc#L3713-L3739
does not include the periodic neighbors for a triangulation immediately after `add_periodicity`.

This was not a problem when we refined the mesh, as then the periodicity has already been set.

FYI @nfehn

### DIFF
--- a/doc/news/changes/minor/20200330MartinKronbichler-1
+++ b/doc/news/changes/minor/20200330MartinKronbichler-1
@@ -1,0 +1,3 @@
+Fixed: A bug in the setup of CellAccessor::level_subdomain_id() for multigrid
+levels that manifested itself only on non-refined meshes has been fixed.
+<br> (Martin Kronbichler, 2020/03/30)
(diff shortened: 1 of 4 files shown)


kronbichler:
Yes, deadlock for DG. For continuous elements we got an error in `MGConstrainedDoFs` that we were trying to get dofs on artificial cells.

tjhei:
> Yes, deadlock for DG. For continuous elements we got an error in `MGConstrainedDoFs` that we were trying to get dofs on artificial cells.

I guess that is good enough. :-)

PR #9783, opened 2020-03-31
Pass ExcNoMatch the description directly

peterrum:
@nfehn has reported that the following program segfaults:
```cpp
#include <deal.II/base/parameter_handler.h>


using namespace dealii;

int
main(int argc, char ** argv)
{
  try
  {
     unsigned int i = numbers::invalid_unsigned_int;

     ParameterHandler params;
     params.add_parameter("dummy", i);

  }
  catch(std::exception & exc)
  {
    std::cout << exc.what() << std::endl;
  }

}
```

The function call `add_parameter()` fails because the variable does not match the pattern (this is wanted in this example). An exception (`ExcNoMatch`) is thrown, whose content (`PatternBase`) does not live in the catch-clause:

https://github.com/dealii/dealii/blob/ec03a88e9f92cdd26f591ebf55983e307a4c973c/include/deal.II/base/patterns.h#L1412-L1416

I think this is an unwanted behavior, isn't it? With replacing the pattern directly by the string, this problem could be *fixed*.

@luca-heltai What do you think?

### DIFF
--- a/include/deal.II/base/patterns.h
+++ b/include/deal.II/base/patterns.h
@@ -1411,9 +1411,9 @@ namespace Patterns
      */
     DeclException2(ExcNoMatch,
                    std::string,
-                   const Patterns::PatternBase *,
+                   std::string,
                    << "The string " << arg1 << " does not match the pattern \""
-                   << arg2->description() << "\"");
+                   << arg2 << "\"");
     //@}
   } // namespace Tools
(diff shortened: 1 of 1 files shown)


luca-heltai:
Yes, I agree. Passing  `const Patterns::PatternBase *` should be avoided. Passing directly the description of the pattern would be the correct option.

PR #9784, opened 2020-03-31
Implement default initializer for EigenvalueInformation

masterleinad:
Fixes most of the warnings in https://cdash.43-1.org/viewBuildError.php?type=1&buildid=5357.

### DIFF
--- a/include/deal.II/lac/precondition.h
+++ b/include/deal.II/lac/precondition.h
@@ -1187,6 +1187,15 @@ public:
      * AdditionalData::degree or estimated as described there).
      */
     unsigned int degree;
+    /**
+     * Constructor initializing with invalid values.
+     */
+    EigenvalueInformation()
+      : min_eigenvalue_estimate{std::numeric_limits<double>::max()}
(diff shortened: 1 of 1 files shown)


masterleinad:
> Would it be simpler to just do
> `EigenvalueInformation info = {};` instead in [/include/deal.II/lac/precondition.h@`master`#L2284](https://github.com/dealii/dealii/blob/master/include/deal.II/lac/precondition.h#L2284) [](/dealii/dealii/blob/HEAD@{2020-03-31T18:13:04Z}/include/deal.II/lac/precondition.h#L2284)

I don't think that will work (see https://godbolt.org/z/C6dkVm) and I also prefer to be explicit about the values the struct is initialized with.

tjhei:
> I don't think that will work (see https://godbolt.org/z/C6dkVm) and I also prefer to be explicit about the values the struct is initialized with.

huh, I guess gcc 4.9 has a different idea about this. ;-)

bangerth:
We could also go with the principle of least surprise: @masterleinad 's way is how we do it everywhere else.

So I'm in favor of the current approach.

PR #4040, opened 2017-03-05
Deprecate SwappableVector.

drwells:
This class is not used anywhere in the library nor are there any tests for it. Since more modern machines have tons of memory I think that the use case for this is not as strong as it used to be and we can retire this. Thoughts?

I noticed this while writing #4039: I had not seen this class before that.

PR #9785, opened 2020-03-31
Remove SwappableVector

masterleinad:
Deprecated in https://github.com/dealii/dealii/pull/4040.


bangerth:
This is a class I wrote as part of my Diploma thesis work, in 1998. I imagine it predates nearly everything else in the library :-)

PR #9787, opened 2020-04-01
Encode and decode Base64 format.

luca-heltai:
Base64 is a group of binary-to-text encoding schemes that represent binary
data in an ASCII string format by translating it into a radix-64          
representation. Base64 is designed to carry data stored in binary formats 
across channels that only reliably support text content. It is used also  
to store binary formats in a machine independent way. 

This PR provides encode_base64 and decode_base64, scorporating part of what was in PR #9376, and making it independent on boost/beast.


tests/base/utilities_encode64_decode64_01.output
```diff
@@ -0,0 +1,10 @@
+
+DEAL::Encoding: Hello
+DEAL::Encoded: SGVsbG8=
+DEAL::Decoded: H e l l o
+DEAL::Encoding: deal.II rocks!
+DEAL::Encoded: ZGVhbC5JSSByb2NrcyE=
+DEAL::Decoded: d e a l . I I   r o c k s !
```
bangerth, line 7: :-)

luca-heltai:
> We already have the ability to do this -- take a look at source/base/data_out_base.cc around lines 88-226. I don't have objections to make this more generic and moving functions into the global namespace, but we shouldn't keep two copies of the same functions :-)

I removed the old implementation, and kept the new one in place. With roughly the same lines of codes, we now also have a decoder. :)

PR #9788, opened 2020-04-01
Fix portability issue

Rombur:
Make step 1, 2, and 3 ISO 646 compliant. There will be a new requirement by xSDK that all the codes need to be ISO 646 compliant. This is because non-compliant code cannot be run on some of the hardware used by the DOE.

### DIFF
--- a/examples/step-1/step-1.cc
+++ b/examples/step-1/step-1.cc
@@ -15,24 +15,25 @@
 
  */
 
+// clang-format off
 // @sect3{Include files}
 
 // The most fundamental class in the library is the Triangulation class, which
 // is declared here:
-#include <deal.II/grid/tria.h>
+%:include <deal.II/grid/tria.h>
 // We need the following two includes for loops over cells and/or faces:
-#include <deal.II/grid/tria_accessor.h>
-#include <deal.II/grid/tria_iterator.h>
+%:include <deal.II/grid/tria_accessor.h>
+%:include <deal.II/grid/tria_iterator.h>
 // Here are some functions to generate standard grids:
-#include <deal.II/grid/grid_generator.h>
+%:include <deal.II/grid/grid_generator.h>
 // Output of grids in various graphics formats:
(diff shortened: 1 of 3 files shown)


examples/step-1/step-1.cc
```diff
@@ -15,24 +15,25 @@
 
  */
 
+// clang-format off
 // @sect3{Include files}
 
 // The most fundamental class in the library is the Triangulation class, which
 // is declared here:
-#include <deal.II/grid/tria.h>
+%:include <deal.II/grid/tria.h>
```
drwells, line 23: It looks like GitHub's syntax highlighting is not standards-compliant :(

masterleinad: ... and for some reason, `clang-format` messes everything up pretty badly.

tjhei: Well, I vote to remove clang-format and to go back to astyle then.

tamiko: That's a shame. :-(

I thought clang-format would support modern C++ and now it turns out it cannot even deal with digraphs.

Alternatively, I suggest we write a small replacement lexer for clang that we can feed into our own clang-format executable that internally translates between standard syntax and digraphs.

masterleinad: Please open a separate pull request for that.

tamiko: on it.

bangerth:
On 4/1/20 11:46 AM, Matthias Maier wrote:
> Alternatively, I suggest we write a small replacement lexer for clang that we 
> can feed into our own clang-format executable that internally translates 
> between standard syntax and digraphs.

We could use the same kind of hack we use to format the .inst.in files!

Rombur:
It looks like xSDK decided not to go with this requirement after all. So closing this.

tamiko:
@Rombur What a shame. And I was just about to upload and open a pull request for a full bundled llvm/clang with custom lexer.

PR #9791, opened 2020-04-01
Simplify DoFCellAccessor::distribute_local_to_global()

peterrum:
This PR removes 4 very old TODOs (at least 11 years old).


peterrum:
@bangerth Done! I have also used the opportunity to refactor the code a bit. With the modification from yesterday, the internal functions for `DoFHandler` and `hp::DoFHandler` were (nearly) identical. I have removed the internal functions and moved the content of one of them into the actual function. Result: -300 locs :)

peterrum:
I have run the test suite with this configuration https://github.com/dealii/dealii/pull/9760#issuecomment-605505157 locally. It passes.

PR #9792, opened 2020-04-01
Allow vector_adaptor to work on ghosted vectors.

dougshidong:
Previous implementation did not work on ghosted vectors since it used
compress().

The test on ghosted vectors did not truly test for ghosted vectors
since it used the copy constructor of LA::distributed::Vector,
which does not fill the ghosted entries.

### DIFF
--- a/include/deal.II/lac/vector.h
+++ b/include/deal.II/lac/vector.h
@@ -1007,6 +1007,14 @@ public:
    */
   std::size_t
   memory_consumption() const;
+
+  /**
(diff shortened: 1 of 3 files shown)


dougshidong:
Moved the header description to appropriate section "Information about the object".

bangerth:
@luca-heltai -- I think you wrote the ROL adapters, right? Can you take a look?

luca-heltai:
Actually, I did not. However, I looked at the changes, and they seem ok to me (and very self contained).

PR #9793, opened 2020-04-02
MappingQGeneric: Copy mapping support points to quadrature points

kronbichler:
This is a small optimization for the tensor-product path of `MappingQGeneric`: For the case we use the tensor product path (mapping degree > 1, tensor product quadrature), we only request the quadrature point locations, and the quadrature point locations coincide with the mapping support points, we can simply copy the mapping support points to the quadrature points without further computation. This case is detected by the `ShapeInfo` object in the mapping.

Since I wanted to convince myself that we really check this new code path, I added a new test that goes into the new path with `QGaussLobatto(mapping_degree+1)`, the old path of tensor product interpolation with `QGaussLobatto(mapping_degree+2)`, and the stand path without tensor products with a `Quadrature(QGaussLobatto(n_pts).get_points())`.

FYI @nfehn


kronbichler:
I also verified that this passes the test suite locally (except for the usual failures on my platform).

PR #9795, opened 2020-04-02
Fix GridIn::read_vtk

peterrum:
The following test fail on the node29 tests:
```
	2498 - grid/grid_out_in_vtk_01.debug (Failed)
	2507 - grid/grid_out_vtk_03.debug (Failed)
```

This PR should fix them.


peterrum:
Currently, `GridIn::read_vtk()`:

https://github.com/dealii/dealii/blob/92a04057bf32ed3739e00064037cd7cd0bce5fac/source/grid/grid_in.cc#L412

and `GridOut::write_vtk()`:

https://github.com/dealii/dealii/blob/92a04057bf32ed3739e00064037cd7cd0bce5fac/source/grid/grid_out.cc#L3290-L3291

treat material-id (and manifold-id) with different types: one with `double` and the other with `int`. This leads the failing tests as listed at the top of this PR.

I have made the types consistent (`int`). Please note that this change required me to update the input file of one of the tests. Is this fine. Or should we support both `int` and `doubles` for reading? Any opinions?

tests/grid/grid_in_vtk_3d_03.vtk
```diff
@@ -34,9 +34,9 @@ CELL_TYPES 4
 12
 12
 CELL_DATA 4
-SCALARS MaterialID double 1
+SCALARS MaterialID int 1
```
bangerth, line 37: I would be more comfortable if there were some kind of assertion that checks that `int` is indeed what this table uses. If you ignore the type specified here, and some input file uses `double`, then your parsing isn't going to succeed.

bangerth: We ignore the data type in current line 391 of the .cc file.

source/grid/grid_in.cc
```diff
@@ -384,11 +384,19 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
...
                       // (the last number is optional)
-                      in.ignore(256, '\n');
+                      std::string line;
+                      std::getline(in, line);
+                      AssertThrow(
+                        line.substr(1,
+                                    std::min(static_cast<std::size_t>(3),
+                                             line.size() - 1)) == "int",
+                        ExcMessage(
+                          "While reading VTK file, material- and manifold IDs can only have type 'int'."));
```
peterrum, line 399: @bangerth Something like this?

bangerth: Yes!

masterleinad:
The same here. Please rebase to get `Travis` unstuck.

masterleinad:
Please try rebasing again.

Issue #8198, opened 2019-05-16
GridGenerator::hyper_ball: Do we want alternative variant?


kronbichler:
The meshes used by `GridGenerator::hyper_ball` use the minimal numbers of cells possible, 5 in 2D and 7 in 3D. An alternative mesh is to use 12 cells for the base mesh in 2D (3 cells per quarter) and 32 in 3D. The 2D mesh looks like this (one can play a bit with where to put the additional point):

![mesh_circle](https://user-images.githubusercontent.com/8276200/57861233-7fb8fa00-77f6-11e9-9cab-0435aad50458.png)

I think it would be good to provide this as alternative to `GridGenerator::hyper_ball` (to be selected via an optional argument `n_cells`?) because this variant has somewhat better mesh quality and more uniform cell sizes between the circumferential and radial directions. In essence, this mesh corresponds to the hyperball mesh refined once and removing one of the outer mesh layers. We also have this form for the quarter ball. What do others think?

tjhei:
> I think it would be good to provide this as alternative to `GridGenerator::hyper_ball` (to be selected via an optional argument `n_cells`?)

I think this would be useful to have. I am not a big fan of optional arguments but it would mirror what we do for hyper_shell...

bangerth:
I like this variant. As mentioned, it nicely mirrors what we already do for other geometries.

kronbichler:
I am looking at this right now. One complication I see is the current signature:
```
  hyper_ball(Triangulation<dim> &tria,
             const Point<dim> &  center = Point<dim>(),
             const double        radius = 1.,
             const bool attach_spherical_manifold_on_boundary_cells = false);
```
If I add an optional argument, I need to add it after `attach_spherical_manifold_on_boundary_cells`, which is a bit different from the `hyper_cell` functions that do this before their `colorize` argument. This sort of confirms @tjhei's scepticism about default integer/bool arguments. Since we have had that argument since the 9.0 release, we probably don't want to change that, even though I assume it is not often used? Opinions?

tjhei:
> Opinions?

I don't like trying to fit in a new function with optional arguments. I would just create a new function with a different name (and maybe deprecate the old one). Maybe something like ``hyper_ball_ex``, ``hyper_ball_version2`` or something?

bangerth:
If you create a different function, you could re-use the same name and make the argument you add the second one, between `tria` and `center`. Then the existing one could just forward to the new one and be deprecated.

[...]

PR #9796, opened 2020-04-02
Improve mesh quality of quarter hyper ball

kronbichler:
This PR is a preparation for #8198 .

I observed that the mesh quality of the `quarter_hyper_ball` functions are not as good as they could be. Thus, I adjusted the relevant parameters a bit, producing the picture on the right (left is the old variant):
  

The 3D pictures changed like this:
  

The idea of choosing the point is to equilibrate the minimal singular value of the Jacobian in the crossing point between the 3 cells. If we agree on this, I will add the new pictures to the documentation of the class.


drwells:
I think we should add a changelog entry since this is technically an incompatibility.

tamiko:
@drwells But a very minor one. We had more severe changes in our mesh generators in the past.
But yes, maybe a minor changes changelog entry?

kronbichler:
I added a changelog entry and the two new pictures for the manual.

kronbichler:
I have adapted the output of the two affected tests as well.

PR #9797, opened 2020-04-02
Remove unused functions

peterrum:
I guess we can remove these function. They are nowhere used.


masterleinad:
Can you please rebase? It seems like `Travis` doesn't report back and we can't merge without it.

peterrum:
@masterleinad I have rebased this branch and the one in PR #9795 in the morning. Travis seems still to be stuck...

kronbichler:
I think Travis CI works better now, so we could give this another spin.

PR #9799, opened 2020-04-02
Minor cleanups.

bangerth:
Found while looking at #9795.
/rebuild

### DIFF
--- a/source/grid/grid_in.cc
+++ b/source/grid/grid_in.cc
@@ -371,17 +371,14 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
                       // Now see if we know about this type of data set,
                       // if not, just ignore everything till the next SCALARS
                       // keyword
-                      std::string set = "";
-                      in >> keyword;
-                      for (const auto &set_cmp : data_sets)
-                        if (keyword == set_cmp)
-                          {
-                            set = keyword;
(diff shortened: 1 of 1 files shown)


source/grid/grid_in.cc
```diff
@@ -411,10 +407,10 @@ GridIn<dim, spacedim>::read_vtk(std::istream &in)
                         {
                           double id;
                           in >> id;
-                          if (set == "MaterialID")
+                          if (keyword == "MaterialID")
```
masterleinad, line 410: AFAICT, these have different values. `set` is the next known data set name after `SCALARS` while we are still updating `keyword`.

bangerth: No, look at line 374-379 (old) above: We do `set = keyword` if we find something (and `continue` if we don't). I think this really is correct.

masterleinad: Yes, but we change `keyword` in lines 389 and 395. I believe this is responsible for the failing tests.

bangerth:
@masterleinad -- You're totally right, of course. I've just given the field name a more descriptive variable name -- the root cause for me missing this is because we keep reusing the variable `keyword` for things that are not actually keywords.

This now succeeds with all relevant tests on my machine.

PR #9800, opened 2020-04-02
Minimize boost/geometry.h includes

masterleinad:
Alternative/Related to #9798. These are the minimal includes and guards I need to build warning-free passing all `boost` tests.

### DIFF
--- a/include/deal.II/boost_adaptors/segment.h
+++ b/include/deal.II/boost_adaptors/segment.h
@@ -22,7 +22,7 @@
 
 #include <deal.II/boost_adaptors/point.h>
 
-#include <boost/geometry.hpp>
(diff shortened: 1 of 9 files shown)


include/deal.II/boost_adaptors/segment.h
```diff
@@ -22,7 +22,7 @@
 
 #include <deal.II/boost_adaptors/point.h>
 
-#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/segment.hpp>
```
tjhei, line 26: This should be guarded as well, I think.

masterleinad: I can do that but I didn't need it to compile warning-free with `gcc-9.3`.

PR #9801, opened 2020-04-02
Remove PointerMatrix*

masterleinad:
Deprecated in https://github.com/dealii/dealii/pull/4607.


drwells:
This class served an important role at one point, but now that we have LinearOperator we can do things in a nicer way.

PR #9802, opened 2020-04-02
Add alternative GridGenerator::hyper_ball with better balance of cells

### LINKED ISSUE #8198: GridGenerator::hyper_ball: Do we want alternative variant?

kronbichler:
Closes #8198.

As discussed in #8198, this PR adds a variant of the hyper ball (disk in 2D, ball in 3D) with more base cells and a better balance. I chose the name `hyper_ball_balanced` because it tries to make cells approximately the same size, but I am open to other name suggestions as well.


kronbichler:

PR #9803, opened 2020-04-02
Fix oversight in GridGenerator::hyper_ball in 2d

kronbichler:
Found while working on #9802.

This is simply an oversight, as the 3D case does the check the other way around (apart from a different cell order):
https://github.com/dealii/dealii/blob/a2518c4284c372151fba97a5ca630151a1a04234/source/grid/grid_generator.cc#L4978

### DIFF
--- a/source/grid/grid_generator.cc
+++ b/source/grid/grid_generator.cc
@@ -4220,7 +4220,7 @@ namespace GridGenerator
         for (unsigned int j = 0; j < 4; ++j)
           cells[i].vertices[j] = cell_vertices[i][j];
         cells[i].material_id = 0;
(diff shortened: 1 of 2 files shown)


kronbichler:
I now also adapted the one test case where the serial CI complained - the switch in manifold ids is expected. I copied the generated file there which includes some trailing whitespace - I think this is technically the more correct solution as it matches what we get out. Or do we have a policy for the `.output` files?

drwells:
Unfortunately I think we do need white space (at least we need multiple blank lines at the ends of files) to make numdiff happy.

drwells:
We presently have 947 output files containing trailing whitespace so I think our policy is 'this isn't worth worrying about'.

PR #9807, opened 2020-04-03
Add additional cell options for 3D hyper shell

kronbichler:
One thing I often found annoying with the 3D hyper shell is that one cannot control the ratio of cells in the azimuthal versus radial direction. The current interface allows for a nice solution: We can add variants that refine the 6-cell/12-cell mesh once but only keep one mesh layer in the radial direction, so we end up with 24/48 cells in the coarse mesh. This is good for thin shells.

I could extend this further to make the mesh accept all variants with 2^m * 6 cells, m integer, by creating m/2 refinements and a single layer in radial direction. This would more closely reflect the 2d case. We could even make the "0" cell case choose the most appropriate variant in terms of mesh quality like in 2D. However, we have the odd one out with the old 96 cells, which does not fit in here.

How do we want to proceed:
- Use the proposed patch with 24 and 48 cells only.
- Generalize to 2^m * 6 cells except for the 96 cell case that would be the odd one out.
- Make the incompatible change for 96 cells and use 2^m * 6 cells for all given sizes. This is the nicest option, but it does affect users who use this case.

I can add a test and adjust the code to what we like most.

### DIFF
--- a/include/deal.II/grid/grid_generator.h
+++ b/include/deal.II/grid/grid_generator.h
@@ -1050,25 +1050,32 @@ namespace GridGenerator
    * is zero (as is the default), then it is computed adaptively such that the
    * resulting elements have the least aspect ratio.
    *
-   * In 3d, only certain numbers are allowed
+   * In 3d, only certain numbers are allowed:
    * <ul>
    * <li> 6 (or the default 0) for a surface based on a hexahedron (i.e. 6
    *      panels on the inner sphere extruded in radial direction to form 6
    *      cells),
    * <li> 12 for the rhombic dodecahedron,
-   * <li> 24 for the hexahedron-based surface refined once in the spherical
+   * <li> 24 for the hexahedron-based surface refined once in the azimuthal
    *      directions but not in the radial direction,
-   * <li> 48 for the rhombic dodecahedron refined once in the spherical
+   * <li> 48 for the rhombic dodecahedron refined once in the azimuthal
    *      directions but not in the radial direction,
    * <li> 96 for the rhombic dodecahedron refined once. This choice dates from
    *      an older version of deal.II before the Manifold classes were
    *      implemented: today this choce is equivalent to the rhombic
    *      dodecahedron after performing one global refinement.
+   * <li> Numbers of the kind $192\times 2^m$ with $m\geq 0$ integer. This
+   *      choice is similar to the 24 and 48 cell cases, but provides
+   *      additional refinements in azimuthal direction combined with a single
+   *      layer in radial direction. The base mesh is either the 6 or 12 cell
+   *      version, depending on whether $m$ in the power is odd or even,
+   *      respectively.
    * </ul>
-   * The versions with 24 and 48 cells are useful if the shell is thin and the
-   * radial lengths should be made more similar to the circumferential lengths.
+   * The versions with 24, 48, and $2^m 192$ cells are useful if the shell is
+   * thin and the radial lengths should be made more similar to the
+   * circumferential lengths.
    *
-   * The grids with 12 and 96 cells are plotted below:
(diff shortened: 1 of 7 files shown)


agrayver:
Personally, I would vote for a general 2^m * 6 solution for this. Thin shells occur very often in Earth sciences (think of ocean, atmosphere, etc). I had to implement this way of refinement in my own code.

include/deal.II/grid/grid_generator.h
```diff
@@ -1039,12 +1039,23 @@ namespace GridGenerator
...
-   * Manifold classes were implemented: today all three choices are roughly
-   * equivalent (after performing global refinement, of course).
+   * In 3d, only certain numbers are allowed
+   * <ul>
+   * <li> 6 (or the default 0) for a surface based on a hexahedron (i.e. 6
+   *      panels on the inner sphere extruded in radial direction to form 6
+   *      cells),
+   * <li> 12 for the rhombic dodecahedron,
+   * <li> 24 for the hexahedron-based surface refined once in the spherical
+   *      directions but not in the radial direction,
```
agrayver, line 1049: "spherical direction" -- at least in my field, this would be called "azimuthal" or "tangential" direction.

kronbichler: You know much more about these things than I, so I'm happy to make that change once we have decided.

bangerth:
ASPECT uses the 96-cell shell. I would really like this to remain compatible to what it currently is.

kronbichler:
@bangerth would you be ok if we go with my option 2, i.e., let the user specify 2^m * 6 cells for integer `m`, creating refinement in azimuthal direction but a single layer of cells in radial direction, but let the 96 cell case be the odd case with 2 refinements in radial direction? I will then augment the documentation and the code to this case.

bangerth:
Yes, as long as you stay backward compatible for 96 cells I'm fine with anything that sounds reasonable to you!

kronbichler:
@bangerth this is now ready for re-view, using option 2 with 2^m * 6 cells of a single radial refinement except the 96 cell case.

source/grid/grid_generator.cc
```diff
@@ -5360,113 +5375,168 @@ namespace GridGenerator
...
+                // We remove all cells which do not have exactly four vertices
+                // inside the outer radius. Due to roundoff, we do not test
+                // for "square < outer_radius * outer_radius" but something
+                // more stable.
+                for (const auto &cell : tmp.active_cell_iterators())
+                  {
+                    unsigned int n_vertices_inside = 0;
+                    for (const auto v : GeometryInfo<3>::vertex_indices())
+                      if ((cell->vertex(v) - p).norm_square() <
+                          inner_radius * outer_radius)
```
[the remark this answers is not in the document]
kronbichler: I test whether each cell has four vertices inside `outer_radius`. Actually, I look for something close to `inner_radius`, like `inner_radius * (1+1e-12)`. Rather than writing it like this, I used a number that I know for sure is less than `outer_radius` and a bit more than `inner_radius`. Since `inner_radius` and `outer_radius` must differ enough to give a viable mesh, I thought this is the least amount of code to write. But I admit it is harder to follow than necessary, and thus will write `inner_radius * inner_radius * (1 + 1e-12)` to be more descriptive of what I want.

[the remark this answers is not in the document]
bangerth: I don't particularly care which way you do the test, as long as I can follow the comment :-)

kronbichler:

include/deal.II/grid/grid_generator.h
tamiko: I'll postpone the documentation change to after the merge.

PR #9809, opened 2020-04-03
new function Triangulation::has_boundary_id()

nfehn:
Aim: ask triangulation if any of the boundary faces has a given boundary id. The function ``get_boundary_ids()`` does something similar. However, the function ``Triangulation::get_boundary_ids()`` has ``std::vector`` as return type, so that it is not possible to simply write ``tria->get_boundary_ids().find(bid)``. It is probably unrealistic to change the return type to ``std::set``.

### DIFF
--- a/include/deal.II/grid/tria.h
+++ b/include/deal.II/grid/tria.h
@@ -1740,6 +1740,19 @@ public:
   get_boundary_ids() const;
 
   /**
+   * Return whether a specified boundary indicator @p bid is assigned
+   * to this triangulation.
+   *
+   * @ingroup boundary
+   *
+   * @see
+   * @ref GlossBoundaryIndicator "Glossary entry on boundary indicators"
+   */
+  bool
+  has_boundary_id(const types::boundary_id bid) const;
+
+
+  /**
    * Return a vector containing all manifold indicators assigned to the
    * objects of this Triangulation. Note, that each manifold indicator is
    * reported only once. The size of the return vector will represent the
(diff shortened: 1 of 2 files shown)


peterrum:
@nfehn Could you shortly describe in which context you need to query if a boundary id is used. I am asking because - if I am reading the implementation correctly - this function returns only the boundaries of the local partition of the triangulation. Would this be a problem in your workflow?

nfehn:
Yes, this would then probably be a problem because I am interested in global information. I did not find this in the documentation of get_boundary_ids(). I think this would be essential information?

nfehn:
I actually think that the ParameterHandler thing is very important and I don‘t see why opening this again after we agreed on that? I also think people will appreciate that functionality in the future. For reasons of fairness, I could mention the write_vtu_with_pvtu_record() thing, where approx. 10 tutorials copied the same twenty lines of code, and certainly a multitude of user codes. A user code that at many places again and again reimplements „the two lines“ of has_boundary_id() is a code I would not like to work with, if this is the developers attitude to „don’t repeat yourself“, and this is actually what professionals told me in C++ courses, not my own fantasy. Back to the current issue: What is the reason for the return type std::vector, which has also been part of my motivation? I am on your side in the sense that I would not have opened this issue in that case, and - at least I think so - I am open to being convinced.

bangerth:
I didn't mean to question the resolution on `ParameterHandler` in #9718 again (I approved it yesterday and the only reason it isn't merged yet is because the tester hadn't approved it yet.

I do think of myself as a professional -- I've been programming in C++ since 1991 and written my fair share of both good and bad code. But let's not make this personal and just do what one does in cases of disagreement: We need a third opinion to break the deadlock here :-)


(I'll add that I wished that the `get_boundary_id()` function returned a `std::set` so that one could call `triangulation.get_boundary_id().contains(some_id)` in C++20.)

PR #9812, opened 2020-04-03
Fix get_boundary_ids() and get_manifold_ids() for parallel triangulations

peterrum:
Related #9809.

The test below shows how to construct a case where `get_boundary_ids()` does not provide the wanted result. (The key aspect is not assigned the boundary ids on the coarse grid but just on the active cells, which is - as far as I know - allowed).

The setup is as follows: left face get the boundary id 1, the right face 2, the bottom face 3, and the top face 4. I considering a subdivided hypercube with 4 subdivisions and two partitions. Process with rank 0 gets the lower part (i.e. boundary ids:1, 2, 3) and process with rank 1 gets the top part of the domain (i.e. boundary ids:1, 2, 4).

The program prints the result of the function call of `get_boundary_id()` for each process:

```
DEAL:0::0 1 2 3 
DEAL:1::0 1 2 4 
```

The zeroes are a problem and is caused by the fact that the function is looping over all active cells (but I only have set the bid of locally-owned cells).

With the changes below, I get the following output:

```
DEAL:0::1 2 3 
DEAL:1::1 2 4 
```

This is better (these are the local bids) but still not what we want. So we need to build the union of the sets (for that I will open a separate PR introducing a new function in `Utilities::MPI`). My suggestion would be to make `get_boundary_ids()` virtual and override it in `parallel::TriangulationBase`.


nfehn:
> My suggestion would be to make get_boundary_ids() virtual and override it in parallel::TriangulationBase.

As we are not happy with the design of ``get_boundary_ids()``, see #9809, an option would be to introduce a new function and ``Assert(false, ...)`` the current function ``get_boundary_ids()`` in the parallel case and maybe deprecate it as well. In my opinion this would be justified as we consider the current behavior a bug. Then we also do not need ``has_boundary_id()``. We could kill two birds with one stone.

peterrum:
This PR is not about deprecating the function `get_boundary_ids()`. I never said that I am not happy with the current design (it is just not working properly in parallel).

I do not see a reason to change the return type from `std::vector` to `std::set` just because latter provides a `find` method. Users can convert the vector to a set by `std::set<boundary_id>(vec.begin(), vec.end())`. This is a C++ issue.

It should be also noted that the function `get_boundary_ids()` is not free since over all faces has to be looped (in particular if global communication has to be performed in parallel). Its use should be limited to a minimum; i.e. if possible only call the function once (and reuse the set/vector).

A problem with a function like `has_boundary_id()` (as proposed in #9809) is that it would call the expensive `get_boundary_ids()` every time. To circumvent this problem, one could cache the boundary ids (as done in many places of the library) but this would require to update the cache regularly. However to determine a time when to update the cache is not trivial (and might require some internal variables which would make the triangulation more complex) since the boundary ids can be randomly changed via the cell iterators (in contrast to our applications where boundary ids do not change over time).

nfehn:
"My suggestion would be ..." and "This PR is not about ..." are somehow in conflict. If you do not want to hear other opinions, just mention that in the very beginning. There are two opinions in PR #9809, both say ``std::set`` would be more appropriate. Please share your concerns about the design of ``get_boundary_ids()`` in PR #9809.

peterrum:
Ready to be reviewed.

include/deal.II/distributed/tria_base.h
```diff
@@ -278,6 +278,24 @@ namespace parallel
...
     void
     fill_level_ghost_owners();
+
+    /**
+     * @copydoc dealii::Triangulation::get_boundary_ids()
+     *
+     * @note This function involves a global communication gathering all current
+     *   IDs from all processes.
+     */
+    std::vector<types::boundary_id>
```
masterleinad, line 288:
```suggestion
    virtual std::vector<types::boundary_id>
```

include/deal.II/distributed/tria_base.h
```diff
@@ -278,6 +278,24 @@ namespace parallel
...
+    get_boundary_ids() const override;
+
+    /**
+     * @copydoc dealii::Triangulation::get_manifold_ids()
+     *
+     * @note This function involves a global communication gathering all current
+     *   IDs from all processes.
+     */
+    std::vector<types::manifold_id>
```
masterleinad, line 297:
```suggestion
    virtual std::vector<types::manifold_id>
```

tests/distributed_grids/get_boundary_ids_01.cc
```diff
@@ -0,0 +1,58 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2008 - 2018 by the deal.II authors
```
masterleinad, line 3:
```suggestion
// Copyright (C) 2020 by the deal.II authors
```

peterrum:
> Looks like you forgot to push the implementation. Also the output file is empty.

Thanks for the review! The implementation is part of an open PR.

masterleinad:
> Thanks for the review! The implementation is part of an open PR.

In other words, this is work in progress and is not ready.

peterrum:
This should work now. Test should pass once #9849 has been merged.

bangerth:
/rebuild

@masterleinad -- you currently own a block on this PR.

tests/distributed_grids/get_boundary_ids_01.cc
```diff
@@ -0,0 +1,58 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// create a tria mesh and copy it
```
masterleinad, line 17: I don't think that is the proper description for this test. :wink:

tests/distributed_grids/get_boundary_ids_01.cc
```diff
@@ -0,0 +1,58 @@
...
+#include "../tests.h"
+
+template <int dim>
+void
+test()
+{
+  parallel::distributed::Triangulation<dim> tria(MPI_COMM_WORLD);
+  GridGenerator::subdivided_hyper_cube(tria, 4);
+
+  for (auto cell : tria.active_cell_iterators())
```
masterleinad, line 32: Not that important, but we usually use
```suggestion
  for (const auto &cell : tria.active_cell_iterators())
```
everywhere (even if you are modifying the underlying object). The same above in `source/grid/tria.cc`.

bangerth:
There are, maybe surprisingly, three failing tests:
```
	3414 - grid/grid_transform.debug (Failed)
	3439 - grid/grid_tools_regularize_01.debug (Failed)
	3502 - grid/grid_tools_regularize_02.debug (Failed)
```
Can you double check whether they fail for good reasons or just need to be updated?

bangerth:
The failure looks like this:
```

An error occurred in line <3597> of file </jenkins/workspace/dealii_PR-9812/include/deal.II/grid/tria_accessor.templates.h> in function

    bool dealii::CellAccessor<dim, spacedim>::is_locally_owned() const [with int dim = 2; int spacedim = 2]

The violated condition was: 

    this->is_active()

Additional information: 

    is_locally_owned() can only be called on active cells!

```

source/grid/tria.cc
```diff
@@ -10397,11 +10397,11 @@ Triangulation<dim, spacedim>::get_boundary_ids() const
   else
     {
       std::set<types::boundary_id> b_ids;
-      active_cell_iterator         cell = begin_active();
-      for (; cell != end(); ++cell)
-        for (const unsigned int face : GeometryInfo<dim>::face_indices())
-          if (cell->at_boundary(face))
-            b_ids.insert(cell->face(face)->boundary_id());
+      for (auto cell : cell_iterators())
```
bangerth, line 10400: This needed to be `active_cell_iterators()` I suppose. We all missed it.

source/grid/tria.cc
```diff
@@ -10414,15 +10414,15 @@ std::vector<types::manifold_id>
...
-  active_cell_iterator         cell = begin_active();
-  for (; cell != end(); ++cell)
-    {
-      m_ids.insert(cell->manifold_id());
-      if (dim > 1)
-        for (const unsigned int face : GeometryInfo<dim>::face_indices())
-          if (cell->at_boundary(face))
-            m_ids.insert(cell->face(face)->manifold_id());
-    }
+  for (auto cell : cell_iterators())
```
bangerth, line 10417: same here

source/grid/tria.cc
peterrum: Yes you are right! Thanks for noticing!

tests/distributed_grids/get_boundary_ids_01.cc
masterleinad: Please fix the description here and I am happy with it.

masterleinad: I'll just do it myself in a follow-up pull request. Then we don't need to run CI again.

PR #9815, opened 2020-04-04
Allow outputting complex-valued vector and tensor fields

### LINKED ISSUE #8625: DataPostprocessor fails with std::complex Vector

bangerth:
Friday evening enjoyment by actually doing some programming :-)

This is going to be a bit of a bear to review and I think I need to lean on those whose long patches I've recently reviewed (hi @tamiko, @kronbichler :-) ). But everyone else is of course welcome to review it too!

As mentioned in #8625 (and related to #1894), outputting complex-valued vector fields currently doesn't yield anything useful. That's because we interleave real and imaginary parts of the solution field, and then neither the real nor the imaginary part of the vector field can easily be visualized any more without further user contortions in the visualization software. The way this needs to be solved is by not interleaving real and imaginary parts. That turns out to be slightly complicated if one wants to do it right, as now discussed in this comment in `data_out.cc`:
```
                    // The situation is more complicated if the input vector is
                    // complex-valued. The easiest approach would have been to
                    // just have all real and then all imaginary components.
                    // This would have been conceptually easy, but it has the
                    // annoying downside that if you have a vector-valued
                    // problem (say, [u v]) then the output order would have
                    // been [u_re, v_re, u_im, v_re]. That's tolerable, but not
                    // quite so nice because one typically thinks of real and
                    // imaginary parts as belonging together. We would really
                    // like the output order to be [u_re, u_im, v_re, v_im].
                    // That, too, would have been easy to implement because one
                    // just has to interleave real and imaginary parts.
                    //
                    // But that's also not what we want. That's because if one
                    // were, for example, to solve a complex-valued Stokes
                    // problem (e.g., computing eigenfunctions of the Stokes
                    // operator), then one has solution components
                    // [[u v] p] and the proper output order is
                    //     [[u_re v_re] [u_im v_im] p_re p_im].
                    // In other words, the order in which we want to output
                    // data depends on the *interpretation* of components.
```
This results in some convoluted code, and the whole thing is not made prettier by a certain formatting tool that has a rather strict preference for 80-columns even if one is in a deeply nested `if`-`for`-`if`-`switch` piece of code.

Regardless: It had been on my plate to fix for a long time already, and I've promised @dangars to do it. So here it is. It was an enjoyable Friday afternoon/evening.

Fixes #8625.

(The first commit is the one in #9813, which I will drop once that is merged.)
/rebuild


bangerth:
I dropped the first commit that was separately merged. I also fixed a small bug that led to a couple of test failures -- but everything that's got `data_out` in its name now succeeds.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1709,32 +1676,107 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::get_dataset_names()
...
+  // consecutive components.
+  //
+  // This sort of logic is also explained in some detail in
+  //   DataOut::build_one_patch().
+  for (const auto &input_data : dof_data)
+    if (input_data->is_complex_valued() == false ||
+        (input_data->postprocessor != nullptr))
+      {
+        for (unsigned int i = 0; i < input_data->names.size(); ++i)
+          names.push_back(input_data->names[i]);
```
kronbichler, line 1701:
```suggestion
        for (const auto &name : input_data->names)
          names.push_back(name);
```

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1709,32 +1676,107 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::get_dataset_names()
...
+                    for (unsigned int tensor_comp = 0; tensor_comp < size;
+                         ++tensor_comp)
+                      names.push_back(input_data->names[i + tensor_comp] +
+                                      "_re");
+                    for (unsigned int tensor_comp = 0; tensor_comp < size;
+                         ++tensor_comp)
+                      names.push_back(input_data->names[i + tensor_comp] +
+                                      "_im");
+
+                    // Move forward by dim components
```
kronbichler, line 1755:
```suggestion
                    // Move forward by dim*dim components
```
alternatively you could also just write `size`.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1709,32 +1676,107 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::get_dataset_names()
...
+
+                default:
+                  Assert(false, ExcInternalError());
+              }
+          }
+      }
+
+  // Do the same as above for cell-type data. This is simpler because it
+  // is always scalar, and so we don't have to worry about whether some
+  // components together form vectors tensors.
```
kronbichler, line 1769:
```suggestion
  // components together form vectors or tensors.
```

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1760,27 +1802,25 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::
...
-      switch ((*d)->data_component_interpretation[i])
+      switch (input_data->data_component_interpretation[i])
         {
           case DataComponentInterpretation::component_is_scalar:
             {
-              // Just move one component forward by one (or two if the component
-              // happens to be complex-valued and we don't use a postprocessor
+              // Just move one component forward by one (or two if the
+              // component happens to be complex-valued and we don't use a
+              // postprocessor
```
kronbichler, line 1817:
```suggestion
              // postprocessor)
```

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1760,27 +1802,25 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::
...
       // see what kind of data we have here. note that for the purpose of the
       // current function all we care about is vector data
-      switch ((*d)->data_component_interpretation[i])
+      switch (input_data->data_component_interpretation[i])
         {
           case DataComponentInterpretation::component_is_scalar:
             {
-              // Just move one component forward by one (or two if the component
-              // happens to be complex-valued and we don't use a postprocessor
+              // Just move one component forward by one (or two if the
```
kronbichler, line 1815:
```suggestion
              // Just move component forward by one (or two if the
```
? In any case, I was a bit confused.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1833,40 +1903,68 @@ DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>::
...
                     DataComponentInterpretation::component_is_part_of_tensor,
                   Exceptions::DataOutImplementation::
-                    ExcInvalidTensorDeclaration(i, (*d)->names[i]));
+                    ExcInvalidTensorDeclaration(i, input_data->names[i]));
 
-              // all seems alright, so figure out whether there is a common name
-              // to these components. if not, leave the name empty and let the
-              // output format writer decide what to do here
-              std::string name = (*d)->names[i];
+              // all seems alright, so figure out whether there is a common
```
kronbichler, line 1916:
```suggestion
              // all seems right, so figure out whether there is a common
```
If I remember correctly, you made that change in the case above.

source/numerics/data_out.cc
```diff
@@ -704,70 +711,268 @@ DataOut<dim, DoFHandlerType>::build_one_patch(
...
+                      offset += dataset->n_output_variables;
+                    }
+                  else
+                    // The situation is more complicated if the input vector is
+                    // complex-valued. The easiest approach would have been to
+                    // just have all real and then all imaginary components.
+                    // This would have been conceptually easy, but it has the
+                    // annoying downside that if you have a vector-valued
+                    // problem (say, [u v]) then the output order would have
+                    // been [u_re, v_re, u_im, v_re]. That's tolerable, but not
```
kronbichler, line 752:
```suggestion
                    // been [u_re, v_re, u_im, v_im]. That's tolerable, but not
```

dangars:
@bangerth Thanks a lot! I'm going to test it tomrrow :)

dangars:
@bangerth I tested this PR and it works, Thanks a lot, this is very useful! I updated step-62 to use this function see PR #9842. 

Is it possible to use DataPostprocessorVector with complex valued vectors? I have not tested this PR with a complex DataPostprocessorVector.

bangerth:
Thanks, @kronbichler, for all of the comments! I've merged them into this branch -- would request merging on occasion!

@dangars: Yes, postprocessors already work. The `DataPostprocessor` class has a longish comment about complex-valued uses.

PR #9817, opened 2020-04-04
Require gcc 4.9 and clang 4.0.

### LINKED ISSUE #8184: Revisit compiler compatibility
After the 9.1 release it is time to revisit our compiler compatibility again - purely for economical reasons.

I currently test on bob:

clang: 4.0, 5.0, 6.0, 7.0, 8.0, 8.0 (with lld)

gcc: 4.9, 6, 7, 8, 9

bangerth:
This fixes #8184. Strictly speaking, it doesn't actually *require* these compiler versions, but it warns (as it did before) if they are not satisfied, and it documents them as "required".

The cmake file has a bunch of checks further below where we set flags for Clang versions <3.6. If we really *required* Clang 4.0, then these could go away, but I left them in place because we're still only warning about earlier compiler warnings, not error out.

### DIFF
--- a/cmake/setup_compiler_flags_gnu.cmake
+++ b/cmake/setup_compiler_flags_gnu.cmake
@@ -21,10 +21,18 @@
 #
 
 IF( CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.8" )
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9" )
   MESSAGE(WARNING "\n"
     "deal.II requires support for features of C++11 that are not present in\n"
-    "versions of GCC prior to 4.8."
+    "versions of GCC prior to 4.9."
+    )
+ENDIF()
+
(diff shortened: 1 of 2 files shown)


bangerth:
I am now doubting whether that's actually a good idea -- so let's have a discussion. (My doubts: The workstations I work on have gcc 4.8.5 as the default compiler. I'm not actually using that compiler, but if that's what's available on my machines, it's likely also what's available on others'...)

/rebuild

rezarastak:
We can try to support the linux distributions that are still supported by their maintainers:

The oldest still-supported Ubuntu is 16.04 LTS which has gcc 5.3.1 and clang 3.8.
The oldest still-supported Debian is Debian Jessie LTS which supports gcc 4.9.2 and clang 3.5.

So, in theory, the users of these linux distribution versions can use gcc version 4.9 higher to compile dealii.

tamiko:
I think it is perfectly reasonable to update to gcc-4.9 and clang-4.0. After all it is just a warning. (We cover gcc-4.9 in one of the tests on bob; we have clang-4.0 covered otherwise).

PR #9819, opened 2020-04-04
Add Utilities::MPI::compute_set_union

peterrum:
Needed by #9812.


include/deal.II/base/mpi.h
```diff
@@ -1024,6 +1024,17 @@ namespace Utilities
...
+    /**
+     * Compute the union of the input vectors @p vec of all processes in the
+     *   MPI communicator @p comm.
+     *
+     * @note This is a collective operation. The result will available on all
+     *   processes.
+     */
+    template <typename T>
+    std::vector<T>
+    compute_union(const std::vector<T> vec, const MPI_Comm &comm);
```
bangerth, line 1036: You'll want to take by reference.

include/deal.II/base/mpi.templates.h
```diff
@@ -413,6 +414,78 @@ namespace Utilities
...
+                                           amount,
+                                           MPI_INT,
+                                           status.MPI_SOURCE,
+                                           status.MPI_TAG,
+                                           comm,
+                                           MPI_STATUS_IGNORE);
+              AssertThrowMPI(ierr_3);
+
+              for (const auto i : buffer)
+                result.insert(i);
```
kronbichler, line 462: More a remark than a comment: I am pretty sure that accumulating the result in the vector directly would be faster than the set (unless compilers changed the implementation of `std::set`), which is why we use `vector` in the `DynamicSparsityPattern` and other places where performance matters. Here the `set` is the least amount of code, and this operation is so much cheaper than the MPI communication, so this is definitely the correct solution.

include/deal.II/base/mpi.templates.h
```diff
@@ -413,6 +414,76 @@ namespace Utilities
...
+                                           MPI_INT,
+                                           status.MPI_SOURCE,
+                                           status.MPI_TAG,
+                                           comm,
+                                           MPI_STATUS_IGNORE);
+              AssertThrowMPI(ierr_3);
+
+              std::sort(result.begin(), result.end());
+              result.erase(std::unique(result.begin(), result.end()),
+                           result.end());
```
peterrum, line 459: @kronbichler What do you think about this implementation?

kronbichler: I did not intend to imply that we should use an alternative implementation given that `set` was so compact, but yes, this should perform better, and it's not even more code in this particular context where MPI wants us to send arrays. :+1:

peterrum:
@bangerth Let's merge this PR so that I can rebase the PR which you approved ;)

include/deal.II/base/mpi.h
```diff
@@ -1024,6 +1024,17 @@ namespace Utilities
...
+    /**
+     * Compute the union of the input vectors @p vec of all processes in the
+     *   MPI communicator @p comm.
+     *
+     * @note This is a collective operation. The result will available on all
+     *   processes.
+     */
+    template <typename T>
+    std::vector<T>
+    compute_union(const std::vector<T> &vec, const MPI_Comm &comm);
```
masterleinad, line 1036: Intuitively, I would expect that an `std::set` is returned. Is there any other reason than #9812 to choose `std::vector` instead?

masterleinad: ...or `std::unordered_set`.

peterrum: @masterleinad If you want, I can introduce second version of the function, which takes a set and returns a set. I would keep the vector version since the internal buffer in the vector is also a vector. Would that work with you?

masterleinad: A lot of functions take in this namespace take an `ArrayView`. Maybe, it makes sense to do the same here (instead).

For me, returning a `std::vector` just feels a bit odd but is more compatible with `get_boundary_ids` (where we might have wanted to use an `std::set` in the first place). I don't feel that strongly about it, though.

bangerth: I agree -- the word "union" already somehow implies that the function operates on sets. Maybe change (or overload) it to take a `std::set` as argument and let it return that? I would also rename the function to `compute_set_union()` or simply `set_union()` to make it clear that that's what it is, rather than some kind of geometric union, for example.

peterrum:
I have made the modifications. I went with `compute_set_union()` and have introduced a version for `std::set` and `std::vector`. The problem with `ArrayView` and MPI datatypes is that we do not know the length of the result at the beginning.

Communicating only only rank 0 is definitely not scaling. But should be fine for now. I can create in a follow-up PR a communication-pattern function which takes a lambda (for arbitrary (all) reduction) and traverses a tree consisting the processes in log(n) steps.

include/deal.II/base/mpi.templates.h
```diff
@@ -413,6 +414,93 @@ namespace Utilities
...
+                              internal::mpi_type_id(vec.data()),
+                              &amount);
+              AssertThrowMPI(ierr_2);
+
+              const unsigned int old_size = result.size();
+              result.resize(old_size + amount);
+
+              const auto ierr_3 = MPI_Recv(result.data() + old_size,
+                                           amount,
+                                           MPI_INT,
```
kronbichler, line 450:
```suggestion
                              internal::mpi_type_id(vec.data()),
```

include/deal.II/base/mpi.templates.h
```diff
@@ -413,6 +414,93 @@ namespace Utilities
...
+
+
+
+    template <typename T>
+    std::set<T>
+    compute_set_union(const std::set<T> &set_in, const MPI_Comm &comm)
+    {
+      // convert vector to set
+      std::vector<T> vector_in;
+      vector_in.assign(set_in.begin(), set_in.end());
```
kronbichler, line 495:
```suggestion
      std::vector<T> vector_in(set_in.begin(), set_in.end());
```
Or is there a reason why the respective constructor is not working?

peterrum:
@kronbichler  I have made the changes! Thanks!

include/deal.II/base/mpi.templates.h
```diff
@@ -413,6 +414,92 @@ namespace Utilities
...
+    compute_set_union(const std::set<T> &set_in, const MPI_Comm &comm)
+    {
+      // convert vector to set
+      std::vector<T> vector_in(set_in.begin(), set_in.end());
+
+      // perform operation to vector
+      const std::vector<T> vector_out = compute_set_union(vector_in, comm);
+
+      // convert vector to set
+      return std::set<T>(vector_out.begin(), vector_out.end());
```
masterleinad, line 500: So you are paying the price of sorting and finding unique elements twice with this implementation. 
I would still prefer to have an `ArrayView` as input and `std::unordered_set` as output and let functions using it deal with the conversion from `std::unordered_set` to `std::vector` (or whatever the function needs to return). In particular, we expect the number of manifold ids and boundary ids to be small anyway such that the overhead doesn't really matter.

masterleinad: > The problem with ArrayView and MPI datatypes is that we do not know the length of the result at the beginning.

The only thing I would change is to create a `std::set` from the intermediate `result` `std::vector` instead of sorting and finding unique elements in-place.
I won't request changes. If others prefer the current state, that's fine with me.

kronbichler: It of course depends on the number of hits and the length of the list, but I would assume the vector variant with sorting to be faster than the set in case we have many insertions and few hits, and set maybe faster in case we have many hits and a lot of redundancy over MPI ranks. `set` is suprisingly slow compared to keeping a vector of integers sorted for small to medium sizes.

masterleinad: I am really only arguing about the interface here and not about the implementation.

bangerth: I am pretty sure that this is the most efficient implementation. We've found in implementing `IndexSet` that the most efficient ways of finding a unique set of entries is to put them into a vector, sort, uniquefy, and then turn into a `std::set`.

For integers, there is really no particular reason to use `std::unordered_set` over `std::set`. The former uses a hash function -- but integers are fairly good hashes for themselves already.

bangerth:
@peterrum -- let's not worry about the follow-up unless we have concrete evidence that the function is actually slow. But if you did want to think about it: You wouldn't have to (and shouldn't) do the tree traversal yourself. MPI has functionality where you can write a custom reduction operation that MPI calls on whatever processes it feels is necessary. It does all of the tree traversal itself and for you.

bangerth:
The failing check, like in another PR, has no recorded output indicating why it failed -- so I'd assume that it's spurious.

Issue #9820, opened 2020-04-05
Warning in release mode with no MPI


rezarastak:
Related to #9613. I am observing these warnings while compiling in release mode with no MPI with gcc 7.5.0.
```
$ ninja obj_base_release 
[50/93] Building CXX object source/base/CMakeFiles/obj_base_release.dir/mpi_noncontiguous_partitioner.cc.o
In file included from ../include/deal.II/lac/vector.h:24:0,
                 from ../include/deal.II/lac/vector_memory.h:26,
                 from ../include/deal.II/lac/lapack_full_matrix.h:27,
                 from ../include/deal.II/base/tensor.h:29,
                 from ../include/deal.II/base/symmetric_tensor.h:25,
                 from ../include/deal.II/base/array_view.h:23,
                 from ../source/base/mpi_noncontiguous_partitioner.cc:16:
../include/deal.II/base/index_set.h: In member function ‘void dealii::Utilities::MPI::NoncontiguousPartitioner<Number>::reinit(const std::vector<unsigned int>&, const std::vector<unsigned int>&, const MPI_Comm&) [with Number = double]’:
../include/deal.II/base/index_set.h:1477:48: warning: ‘return_value’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   , largest_range(numbers::invalid_unsigned_int)
                                                ^
In file included from ../include/deal.II/base/mpi_consensus_algorithms.h:22:0,
                 from ../include/deal.II/base/mpi_compute_index_owner_internal.h:22,
                 from ../include/deal.II/base/mpi_noncontiguous_partitioner.h:22,
                 from ../source/base/mpi_noncontiguous_partitioner.cc:17:
../include/deal.II/base/mpi.templates.h:338:9: note: ‘return_value’ was declared here
       T return_value;
         ^~~~~~~~~~~~
In file included from ../include/deal.II/lac/vector.h:24:0,
                 from ../include/deal.II/lac/vector_memory.h:26,
                 from ../include/deal.II/lac/lapack_full_matrix.h:27,
                 from ../include/deal.II/base/tensor.h:29,
                 from ../include/deal.II/base/symmetric_tensor.h:25,
                 from ../include/deal.II/base/array_view.h:23,
                 from ../source/base/mpi_noncontiguous_partitioner.cc:16:
[...]

PR #9825, opened 2020-04-05
Fix warning in Utilities::MPI for serial release build

peterrum:
Fix warning reported by @rezarastak in #9820.

### DIFF
--- a/include/deal.II/base/mpi.templates.h
+++ b/include/deal.II/base/mpi.templates.h
@@ -230,7 +230,7 @@ namespace Utilities
     T
     sum(const T &t, const MPI_Comm &mpi_communicator)
     {
-      T return_value;
+      T return_value{};
       internal::all_reduce(MPI_SUM,
                            ArrayView<const T>(&t, 1),
                            mpi_communicator,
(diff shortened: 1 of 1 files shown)


masterleinad:
For me, letting `all_reduce` return its result (relying on NRVO) sounds like the better option to avoid the copy.

peterrum:
@masterleinad What do you mean? Will that also work for the other functions, like:

https://github.com/dealii/dealii/blob/c87481cc7e6f18f4dd81726262b7144898343964/include/deal.II/base/mpi.templates.h#L245-L256

https://github.com/dealii/dealii/blob/c87481cc7e6f18f4dd81726262b7144898343964/include/deal.II/base/mpi.templates.h#L260-L267

These functions do not have return values.

masterleinad:
It would work for the other functions better than for the one in question since you need to construct an `ArrayView` in the first place but it's not-owning.
I'd like to understand why we only the warning in this particular case.

peterrum:
@masterleinad I have created a simplified program https://godbolt.org/z/9KN9DF. But unfortunately I don't get any warnings. Could you have a look at compiler flags?

masterleinad:
Here is a reproducer: https://godbolt.org/z/xoZKBF
It also compares default initializing with copy initializing (for an expensive type that would not be usable in the MPI call, but still). I would be fine with default initializing here.
Both copy initializing and default initializing are completely optimized away for `int`, though.

peterrum:
Thanks, @masterleinad ! I have made the modification.

PR #9826, opened 2020-04-05
Do not use FE_Q in MappingQGeneric

kronbichler:
This PR avoids the use of `FE_Q` within `MappingQGeneric`: I noticed that `FE_Q` appears in my profiles for measuring performance, so I thought it is best to remove it. The member variable `fe_q` was not used before due to some previous restructuring, whereas the setup of `FE_Q` for the tensor product `ShapeInfo` can be avoided by initializing from the much cheaper `FE_DGQ<1>` and adding the `dim`-dimensional lexicographic renumbering. As the last change, I also renamed the SIMD length into `n_lanes` from the old `vec_size`. This passes all tests with `mapping` in the name locally.

One thing we could do now is remove `#include<deal.II/fe/fe_q.h>` from `mapping_q_generic.h`. I started with that, but realized that things get messy because we rely on the implicit header inclusion. I started changing for these files here:
```
	modified:   examples/step-10/step-10.cc
	modified:   examples/step-38/step-38.cc
	modified:   examples/step-69/step-69.cc
	modified:   source/meshworker/scratch_data.cc
	modified:   tests/mappings/mapping_q_manifold_01.cc
```
but then I decided that I should not do those in this PR. I can bite the bullet and go through whatever is needed, but we must be clear that this can affect user code. (I still think this should be done.)


kronbichler:
I need to fix a few more tests, so I set `WIP` for now.

kronbichler:
I split off some work done in this PR to #9829 (which includes the first two commits here), so the code to look at is the last commit only.

kronbichler:
This seems to run into severe problems on the tester which I cannot reproduce locally (I have all tests passing apart from the 10 or so that fail regularly on my machine). I will look into them once we have #9829 merged and I have rebased.

include/deal.II/fe/fe_tools.h
```diff
@@ -909,12 +909,23 @@ namespace FETools
    * correct size, which is equal to the number of degrees of freedom in the
    * finite element.
    */
-
   template <int dim>
   void
   hierarchic_to_lexicographic_numbering(unsigned int               degree,
                                         std::vector<unsigned int> &h2l);
```
drwells, line 915: Should we deprecate these?

source/fe/mapping_q_generic.cc
```diff
@@ -817,18 +796,13 @@ MappingQGeneric<dim, spacedim>::InternalData::initialize_face(
 
   if (dim > 1 && tensor_product_quadrature)
     {
-      const unsigned int  facedim = dim > 1 ? dim - 1 : 1;
-      const FE_Q<facedim> fe(polynomial_degree);
+      constexpr unsigned int facedim = dim > 1 ? dim - 1 : 1;
```
drwells, line 799: Since we already assume `dim > 1` in the preceding if statement, could we get rid of this conditional while we are here?

kronbichler:
I have rebased this now - let us see what the CI tester says now. I have checked some of the tests that failed before in valgrind's memory checker and they show no error, so I'm curious about the outcome.

source/fe/mapping_q_generic.cc
kronbichler: I removed this - thanks to the changes in #9829 where I also created instantiations for the `dim=0` case.

source/fe/mapping_q_generic.cc
```diff
@@ -1495,8 +1469,8 @@ namespace internal
 
         const unsigned int n_shape_values = data.n_shape_functions;
         const unsigned int n_q_points     = data.shape_info.n_q_points;
-        const unsigned int vec_length     = VectorizedArray<double>::size();
-        const unsigned int n_comp         = 1 + (spacedim - 1) / vec_length;
+        const unsigned int n_lanes        = VectorizedArray<double>::size();
+        const unsigned int n_comp         = 1 + (spacedim - 1) / n_lanes;
         const unsigned int n_hessians     = (dim * (dim + 1)) / 2;
```
drwells, line 1474: Could we mark the last three as constexpr?

source/fe/mapping_q_generic.cc
```diff
@@ -1537,6 +1511,9 @@ namespace internal
             data.values_quad.resize(n_comp * n_q_points);
             data.gradients_quad.resize(n_comp * n_q_points * dim);
 
+            data.scratch.resize((dim - 1) *
+                                std::max(n_q_points, n_shape_values));
```
drwells, line 1515: This is the only part I don't understand. In `FEEvaluationImpl::evaluate` (which I think this can resolve to since the other evaluation kernels don't use the scratch array at all) we have the following:
```cpp
        temp1 = scratch_data;
        temp2 = temp1 +
                std::max(Utilities::fixed_power<dim>(shape_info.fe_degree + 1),
                         Utilities::fixed_power<dim>(shape_info.n_q_points_1d));
```
which implies that this array needs to be longer. Should it be longer?

source/fe/mapping_q_generic.cc
kronbichler: Yes, of course!

source/fe/mapping_q_generic.cc
kronbichler: Technically the code should have been correct because we only need `temp2` in the 3D case. But this is hard to see and does not matter in terms of memory allocation, so I changed to `2 * std::max(...)` which fits both cases.

masterleinad:
Hmm... `data_out_base_vtu_04` is failing.

kronbichler:
> Hmm... `data_out_base_vtu_04` is failing.

Ah, one of my favorite tests - super sensitive because we write out binary data in compressed form, so each roundoff error surely propagates to the output and causes a failure. Let me get to a machine with the right compile flags for the default build and generate a new blessed file.

kronbichler:
As to the question is why my change caused roundoff differences: I replaced a 2D/3D FE_Q by a 1D FE_DGQ; the multiplication by the "1" (or not exactly one) in other components of `TensorProductPolynomials` can turn the bits one way or the other.

drwells:
That test isn't testing accuracy, just that we can write higher-order output. Could we fix it more permanently by doing something like
```cpp
vec += 1.0; // avoid roundoff for values near zero
Vector<float> vec2 = vec;
vec = vec2;
vec -= 1.0;
```
which works since we interpolate a nonnegative function, or some other strategy for rounding off the last 10 bits of accuracy?

kronbichler:
@drwells you are right that we should reduce the possibility for roundoff. However, that is - as usual - exceedingly difficult. The sensitive quantities here are the grid point positions more than the actual function we write. I tried with `MappingQCache` and truncating all the positions to a fixed number of digits, but the problem is that we do an interpolation within `DataOut` again that likely destroys things again (I have not tried, but there is definitely an interpolation involved via `FEValues`). So what I did instead in the hope of being more robust is to avoid "nice" human readable numbers, as those are likely to contain sequences with sensitive transition from `double` -> `float` that happens within the data out VTU output (where float is written). So I just used some arbitrary numbers I typed in and at least on my machine both the FMA and standard `-O0` output are the same. Unless someone has better ideas, I think this is worth a try and allows us to remove two old specializations.

drwells:
That's a better solution - I forgot that we have to use the mapping again to generate patches. Lets see what the tester does.

drwells:
The only way to know if we need more output files is to commit this so lets merge.

kronbichler:
@drwells Perhaps not surprisingly - my machines seem to be happy with this change:
https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=5670
But also tester does not report failures from what I can see:
https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=5664
(this reminds me that we should reduce the number of failures on `tester` for the release, let me post that in the release issue).

PR #9827, opened 2020-04-05
Fix bug in VectorTools::get_position_vector for high order mappings

### LINKED ISSUE #8850: Disagreement between MappingQ and MappingFEField using find_all_active_cells_around_point
I am trying to use the GridTools function find_all_active_cells_around_point with a MappingFEField in [this post](https://groups.google.com/d/msg/dealii/8ta3DmMCwsE/EJhfX71qBAAJ) and there is unfortunately some error I can't catch using gdb.

I then wanted to confirm that MappingFEField could be used with that function and created the following test

[find_all_active_cells_around_point_mapping_fe_field_02.zip](https://github.com/dealii/dealii/files/3649948/find_all_active_cells_around_point_mapping_fe_field_02.zip)

This is basically a copy of tests/grid/find_all_active_cells_around_point_02.cc, except that it uses MappingFEField instead of MappingQGeneric. According to the [documentation](https://www.dealii.org/developer/doxygen/deal.II/classMappingFEField.html#a92a567d3e25d60d28bfa8a00ae9e13da), I should get the same results since I am using FE_Q and get_vector_position, so I copied 

output grid/find_all_active_cells_around_point_02.output
to 
grid/find_all_active_cells_around_point_mapping_fe_field_02.output. 

However, I do end up with a few different results.

kronbichler:
This fixes #8850 - see https://github.com/dealii/dealii/issues/8850#issuecomment-609401601 for a description of what happens. We really should get rid of the strange default behavior of `MappingQ`.


kronbichler:
The tester failed for unrelated reasons, so I guess we can merge?

doc/news/changes/minor/20200405MartinKronbichler
```diff
@@ -0,0 +1,3 @@
+Fixed: VectorTools::get_position_vector would return wrong positions for
+curved cells in the interior of the domain. This is now fixed.
+<br> (Martin Kronbichler, Doug Shi-Dong, 2020/04/05)
```
bangerth, line 3: Can you break the parentheses onto the next line? I think that the author script requires that.

PR #9828, opened 2020-04-05
Wrapper for SuperLU-MT linear solver

KParas:
This PR aims at developing a wrapper class for the SuperLU-MT SMP linear solver.
Presently, the class has only been written to deal with SparseMatrix and Vector objects. Addition for BlockSparseMatrix along with tests will follow.


KParas:
The primary issue right now is linking  SuperLU-MT library with deal.II. The newly added code compiles but I get linker errors like:
```
../lib/libdeal_II.so.9.2.0-pre: error: undefined reference to 'dCreate_CompCol_Matrix'
../lib/libdeal_II.so.9.2.0-pre: error: undefined reference to 'dCreate_Dense_Matrix'
../lib/libdeal_II.so.9.2.0-pre: error: undefined reference to 'get_perm_c'
../lib/libdeal_II.so.9.2.0-pre: error: undefined reference to 'pdgssv'
```
I tried to write `cmake_config` and `cmake_find` files but it did not work out. I managed to link SuperLU-MT using cmake in a separate example code but the approach involves absolute paths and only works for linking with static library.

Rombur:
@KParas you need deal.II to link against superlu-mt and to do that you need to do something like https://github.com/dealii/dealii/blob/master/cmake/configure/configure_gsl.cmake

KParas:
@Rombur Adding the config file leads to the following error
```
-- Include /calculate/Paras_Work/PhD_Work/DEALII-DEVELOP/dealii/cmake/configure/configure_superlu_mt.cmake
CMake Error at cmake/macros/macro_configure_feature.cmake:164 (MESSAGE):
    Internal build system error: The configuration of DEAL_II_WITH_SUPERLU-MT
  depends on DEAL_II_WITH_BLAS, but CONFIGURE_FEATURE(SUPERLU-MT) was called
  before CONFIGURE_FEATURE(BLAS).
Call Stack (most recent call first):
  cmake/configure/configure_superlu_mt.cmake:22 (CONFIGURE_FEATURE)
  cmake/macros/macro_verbose_include.cmake:19 (INCLUDE)
  CMakeLists.txt:124 (VERBOSE_INCLUDE)
```
SuperLU-MT depends upon `BLAS`
If we comment out the dependency on BLAS, set `-DDEAL_II_WITH_SUPERLU-MT=ON` and set `-DSUPERLU-MT_DIR` manually, we get this error:
```
-- DEAL_II_WITH_SUPERLU-MT has unmet external dependencies.
CMake Error at cmake/macros/macro_configure_feature.cmake:112 (MESSAGE):
  

  Could not find the superlu-mt library!

  Please ensure that a suitable superlu-mt library is installed on your
  computer.

  If the library is not at a default location, either provide some hints for
  autodetection,

      $ SUPERLU-MT_DIR="..." cmake <...>
      $ cmake -DSUPERLU-MT_DIR="..." <...>

  or set the relevant variables by hand in ccmake.
```

masterleinad:
IIRC, the configure files are parsed in alphabetical order so you need to rename it to something like `configure_1_superlu_mt.cmake`.

KParas:
> IIRC, the configure files are parsed in alphabetical order so you need to rename it to something like `configure_1_superlu_mt.cmake`.

The errors still persist even after renaming. I also tried with `configure_2_superlu_mt.cmake` and `configure_3_superlu_mt.cmake`.

bangerth:
Unrelated to the concrete patch here -- doesn't PETSc also wrap SuperLU and could use that solver?

KParas:
> Unrelated to the concrete patch here -- doesn't PETSc also wrap SuperLU and could use that solver?

According to the PETSc [documentation](https://www.mcs.anl.gov/petsc/miscellaneous/external.html), it provides interfaces to SUPERLU and SUPERLU-DIST, but not SUPERLU-MT.

bangerth:
On 4/7/20 12:56 AM, Paras Kumar wrote:
>     Unrelated to the concrete patch here -- doesn't PETSc also wrap SuperLU
>     and could use that solver?
> 
> According to the PETSc documentation 
> <https://www.mcs.anl.gov/petsc/miscellaneous/external.html>, it provides 
> interfaces to SUPERLU and SUPERLU-DIST, but not SUPERLU-MT.

Ah, that's a shame -- I had not realized that they only support two of the 
three :-(

KParas:
@masterleinad Any clues on how to proceed further?

cmake/configure/configure_superlu_mt.cmake
```diff
@@ -0,0 +1,22 @@
...
+## The full text of the license can be found in the file LICENSE.md at
+## the top level directory of deal.II.
+##
+## ---------------------------------------------------------------------
+
+#
+# Configuration for the SuperLU-MT library:
+#
+
+SET(FEATURE_SUPERLU-MT_DEPENDS BLAS)
```
masterleinad, line 20:
```suggestion
SET(FEATURE_SUPERLU-MT_DEPENDS LAPACK)
```
`BLAS`  is configured together with `LAPACK` and only `DEAL_II_WITH_LAPACK` exists as option.

KParas: I still get the same error as before even after setting the path to SuperLU-MT in the cmake prefix
```
-- Include /calculate/Paras_Work/PhD_Work/DEALII-DEVELOP/dealii/cmake/configure/configure_superlu_mt.cmake
CMake Warning at cmake/macros/macro_find_package.cmake:27 (_FIND_PACKAGE):
  By not providing "FindSUPERLU-MT.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "SUPERLU-MT", but CMake did not find one.

  Could not find a package configuration file provided by "SUPERLU-MT" with
  any of the following names:

    SUPERLU-MTConfig.cmake
    superlu-mt-config.cmake

  Add the installation prefix of "SUPERLU-MT" to CMAKE_PREFIX_PATH or set
  "SUPERLU-MT_DIR" to a directory containing one of the above files.  If
  "SUPERLU-MT" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  cmake/macros/macro_configure_feature.cmake:127 (FIND_PACKAGE)
  cmake/macros/macro_configure_feature.cmake:239 (FEATURE_FIND_EXTERNAL)
  cmake/configure/configure_superlu_mt.cmake:22 (CONFIGURE_FEATURE)
  cmake/macros/macro_verbose_include.cmake:19 (INCLUDE)
  CMakeLists.txt:124 (VERBOSE_INCLUDE)


-- DEAL_II_WITH_SUPERLU-MT has unmet external dependencies.
CMake Error at cmake/macros/macro_configure_feature.cmake:112 (MESSAGE):
  

  Could not find the superlu-mt library!

  Please ensure that a suitable superlu-mt library is installed on your
  computer.

  If the library is not at a default location, either provide some hints for
  autodetection,

      $ SUPERLU-MT_DIR="..." cmake <...>
      $ cmake -DSUPERLU-MT_DIR="..." <...>

  or set the relevant variables by hand in ccmake.
```
I fail to understand why the library is not found even after providing the absolute path.

masterleinad: Does `SUPERLU` provide a `CMake` `Config` file? Then, it should work as is. Otherwise, you need to write a `FindSUPERLU-MT.cmake` find similar to the ones in `cmake/modules`. `cmake/modules/FindASSIMP.cmake` might be a good point to start if you don't need to detect more fancy configuration options.

jppelteret:
> Unrelated to the concrete patch here -- doesn't PETSc also wrap SuperLU and could use that solver?

Trilinos has a wrapper for this ([Amesos2 docs](https://trilinos.github.io/amesos2.html)), so I'm sure that extending our list of [supported solver types](https://dealii.org/developer/doxygen/deal.II/structTrilinosWrappers_1_1SolverDirect_1_1AdditionalData.html#a22a275c6fa139cd8bf86cea54ff8f045) should be relatively trivial.

cmake/configure/configure_superlu_mt.cmake
KParas: After adding the following `FindSUPERLU-MT.cmake` file,
```
SET(SUPERLU-MT_DIR "" CACHE PATH "An optional hint to a SuperLU-MT installation")
SET_IF_EMPTY(SUPERLU-MT_DIR "$ENV{SUPERLU-MT_DIR}")

DEAL_II_FIND_LIBRARY(SUPERLU-MT_LIB NAMES libsuperlu_mt_PTHREAD.a
  HINTS ${SUPERLU-MT_DIR}/lib
  PATH_PREFIXES lib${LIB_PREFIX} lib64 lib
  )

DEAL_II_FIND_PATH(SUPERLU-MT_INC slu_mt_ddefs.h
  HINTS ${SUPERLU-MT_DIR}
  PATH_SUFFIXES include
  )

DEAL_II_PACKAGE_HANDLE(SUPERLU-MT
  LIBRARIES REQUIRED SUPERLU-MT_LIB
  INCLUDE_DIRS REQUIRED SUPERLU-MT_INC
  USER_INCLUDE_DIRS REQUIRED SUPERLU-MT_INC
  CLEAR SUPERLU-MT_DIR_LIB SUPERLU-MT_INC
  )
```
We get this type of linker errors:
```
/usr/bin/ld.gold: error: /calculate/iwtm99_lokal/Paras_HomeOffice/spack/opt/spack/linux-ubuntu18.04-x86_64/gcc-7.5.0/superlu-mt-3.1-d5qq466ouorvdkeox332jcpnxb5fg4tt/lib/../lib/libsuperlu_mt_PTHREAD.a(get_perm_c.o): requires dynamic R_X86_64_PC32 reloc against 'getata' which may overflow at runtime; recompile with -fPIC
```
Any clues on how to fix this?

jppelteret: Maybe patch for the SuperLU-MT spack package is what you need:
[superlu-mt: add pic and int64 variants](https://github.com/spack/spack/pull/11361)

If it helps, then perhaps try to raise awareness there and push to get that PR merged.

KParas:
> > Unrelated to the concrete patch here -- doesn't PETSc also wrap SuperLU and could use that solver?
> 
> Trilinos has a wrapper for this ([Amesos2 docs](https://trilinos.github.io/amesos2.html)), so I'm sure that extending our list of [supported solver types](https://dealii.org/developer/doxygen/deal.II/structTrilinosWrappers_1_1SolverDirect_1_1AdditionalData.html#a22a275c6fa139cd8bf86cea54ff8f045) should be relatively trivial.

I tried to move in this direction first as mentioned  [here](https://groups.google.com/d/msg/dealii/xxPjKocjxr8/LHzW_-cNCgAJ) but there seem to be issues while configuring Trilinos with SuperLU-MT using Spack (which is the tool we use for installations) and we are still struggling with that.

cmake/configure/configure_superlu_mt.cmake
bangerth: Yes, that's essentially a bug in the SuperLU configuration. Or, more correctly, it is dangerous to link static (`.a`) libraries together with shared (`.so`) libraries. Does SuperLU not allow to compile itself into a shared library?

KParas: > Maybe patch for the SuperLU-MT spack package is what you need:
> [superlu-mt: add pic and int64 variants](https://github.com/spack/spack/pull/11361)
> 
> If it helps, then perhaps try to raise awareness there and push to get that PR merged.

Thank you so much for the hint. I did these changes on my local spack installation and it works now.

KParas:
Could someone please tell how can I run tests corresponding to a specific `test.cc` file?

bangerth:
`ctest -R regex` where `regex` is a regular expression that matches your test. In particular, the base name of the test (i.e., without the `.cc`) will do as regular expression.

KParas:
> `ctest -R regex` where `regex` is a regular expression that matches your test. In particular, the base name of the test (i.e., without the `.cc`) will do as regular expression.

I tried this already, also for an existing test, but get the following error:
```
Test project /calculate/Paras_Work/PhD_Work/DEALII-DEVELOP/dealii/_build
No tests were found!!!

```

bangerth:
Did you run `make setup_tests`?

masterleinad:
Also, remember that you need a `*.output` file for a new test to be picked up.

KParas:
> Did you run `make setup_tests`?

I was missing this step. The test runs now. Will add some more and push them soon.

KParas:
For the first test, we get errors like the following, if the number of unknowns goes beyond a certain threshold.
```
DEAL::2d
DEAL::Number of dofs = 460289

superlu_mt/superlu_mt_01.debug: RUN failed. ------ Additional output on stdout/stderr:

Storage for L subscripts exceeded; Current column   254917; Need at least 124032458;
You may set it by the 8-th parameter in routine sp_ienv().
Memory allocation failed at line 222 in file pmemory.c

Factor time  =     0.03
Factor flops = 5.445100e+04	Mflops =     1.94
Solve time   =     0.01
Solve flops = 0.000000e+00	Mflops =     0.00
Factor time  =     0.00
Factor flops = 5.308300e+04	Mflops =    18.48
Solve time   =     0.00
Solve flops = 0.000000e+00	Mflops =     0.00
Factor time  =     0.00
Factor flops = 5.308300e+04	Mflops =    20.20
Solve time   =     0.00
Solve flops = 0.000000e+00	Mflops =     0.00
TBB assertion:
Assertion is_set(s) failed on line 214 of file /calculate/Paras_Work/PhD_Work/DEALII-DEVELOP/dealii/bundled/tbb-2018_U2/src/tbb/governor.cpp
Detailed description: Attempt to terminate non-local scheduler instance
terminate called after throwing an instance of 'dealii::StandardExceptions::ExcMessage'
  what():  
An error occurred in line <684> of file </calculate/Paras_Work/PhD_Work/DEALII-DEVELOP/dealii/tests/superlu_mt/../tests.h> in function
    void new_tbb_assertion_handler(const char*, int, const char*, const char*)
The violated condition was: 
    false
Additional information: 
    TBB Exception, see above

Aborted (core dumped)

```

I tried looking into the source code of SuperLU-MT, but could not yet figure out how to remove this error.

bangerth:
Can you get a backtrace for where this error comes from?

KParas:
@bangerth  As per the error message above, the error occurs due to an assertion failure in `Glu_alloc()`. I could trace the call graph upto the top-level `pdgssv()` routine, but could not find where exactly is the root cause, since `Glu_alloc()` is called by several functions (and in some of these functions, several times in a loop) as can be seen  in the call graph below.

Screenshot at 2020-05-31 20-06-20


Also, there is no detailed documentation to aid in understanding what exactly happens within a function.

Andiiiiiii:
I am as well highly interested in this solver. Unfortunately I am not experienced enough to help you with your issue @KParas . The communiy's demand for such a solver should be quite high... This solver would be a jewel of deal.ii :)

Rombur:
@KParas We have changed a lot of things in the way we do multithreading in deal.II. Could you rebase this PR on master and see if that fixes the problem? Thanks

KParas:
@Rombur I got the following error once I try to do `cmake` after rebasing the local feature branch with the upstream master. This file is not present in the  `source/lac` directory. 

```
-- Configuring done
CMake Error at cmake/macros/macro_deal_ii_add_library.cmake:35 (ADD_LIBRARY):
  Cannot find source file:

    swappable_vector.cc

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
Call Stack (most recent call first):
  source/lac/CMakeLists.txt:195 (DEAL_II_ADD_LIBRARY)
```

Rombur:
Did you call cmake in a clean directory?

KParas:
Yes. I did `rm -rf *` in the build directory before calling cmake, but still get the error.

source/lac/CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ SET(_unity_include_src
   sparse_vanka.cc
   sparsity_pattern.cc
   sparsity_tools.cc
+  superlu_mt.cc
+  swappable_vector.cc
```
masterleinad, line 49:
```suggestion
```

masterleinad (changes requested):

KParas:
@masterleinad the code compiles after doing the changes you suggested, but the test still fails with the same error as before.

Rombur:
Is it possible that there are just too many unknowns? Do you know what's the largest problem other people have solved with SuperLU-MT? 
It seems like SuperLU-MT is not on github but SuperLU-dist is and there is support for OpenMP (see https://github.com/xiaoyeli/superlu_dist#15-multicore-and-gpu-optional) I have no clue how it compares to SuperLU-MT but this does look more alive, so it might be worth asking.

KParas:
For `dim=1`, we do not get any errors even for `nDofs=2097153` (obtained by 21 global refinements), but for `dim=2`, we get the error for `nDofs=1050625` (obtained by 10 global refinements).

Rombur:
But how dense is your matrix? Won't the 1D matrix be more sparse and so internally superlu-mt will need to allocate less memory?

bangerth:
A direct solver solving a 2d problem with more than 1M unknowns will need a *very* large amount of memory (I'd estimate several GB to several 10 GB). How much do you have?

Try to solve smaller problems and work your way up to where it fails. Observe your program on those smaller problems with 'top' or some other memory monitoring tool to get a feel for how much memory your program typically needs for different matrix sizes. You can also use the valgrind memory monitor for this purpose.

masterleinad:
Did you actually add your `FindSUPERLU-MT.cmake` file?

KParas:
> Did you actually add your `FindSUPERLU-MT.cmake` file?

Yes, forgot to push it before.

KParas:
We also get the error with `Number of dofs = 263169` for 2D.

I tried solving problems with more than 2M  DOFs on the same computer using the `dealii:: SparseDirectUmfpack` solver and it works fine.  

The matrix is created using `MatrixCreator::create_mass_matrix() on a uniformly refined grid with first order FE_Q ansatz.

masterleinad:
It seems you also still need to remove `swappable_vector.cc` from `source/lac/CMakeLists.txt` for this to compile. :slightly_smiling_face:

KParas:
> It seems you also still need to remove `swappable_vector.cc` from `source/lac/CMakeLists.txt` for this to compile. slightly_smiling_face

again forgot to push before.

bangerth:
It's possible that SuperLU just uses more memory than UMFPACK. I think the only way to really find out is to run a sequence of models with more and more DoFs, see where things start to break, and monitor memory consumption. Have you tried to use the memory monitor tool that's part of valgrind?

Rombur:
Interesting thread https://github.com/xiaoyeli/superlu_dist/issues/64

bangerth:
@KParas What's the status with this patch? I'm thinking that it would be nice to have this in 9.3, for which we'd have ~2 weeks to get the patch in shape. What is missing from your side, and what can we provide/do to move this along?

KParas:
@bangerth I am still stuck at the memory error mentione dabove. I also tired to use the `massif` tool of valgrind for memory usage study, but could not get any useful information from it. It would be great if someone could help me find the root cause of this error, or even re-confirm if this an error at all. May be it is the result that one should expect for dim>1? For dim=1, there are no errors even for 4 million DOFs.

peterrum:
> What's the status with this patch? I'm thinking that it would be nice to have this in 9.3, for which we'd have ~2 weeks to get the patch in shape. What is missing from your side, and what can we provide/do to move this along?

I think it is too ambitious to get it into the 9.3 release. But maybe after that someone could help @KParas!?

include/deal.II/lac/superlu_mt.h
```diff
@@ -0,0 +1,190 @@
...
+      options_.fact   = static_cast<fact_t>(factorize);
+      if (printSolverStats)
+        options_.PrintStat = YES;
+      else
+        options_.PrintStat = NO;
+    }
+    /**
+     * Object to store additional options for the solver
+     */
+    superlumt_options_t options_;
```
bangerth, line 98: Since you reference SuperLU data types and functions in this file, you will have to put most of it into `#ifdef DEAL_II_WITH_SUPERLU`.

include/deal.II/lac/superlu_mt.h
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * Construcotr.
+   */
+  SuperLU_MT(const AdditionalData &data = AdditionalData());
+
+  /**
+   * Desctructor
+   */
+  ~SuperLU_MT() override;
```
bangerth, line 114: Since your class is not derived from anything, is `override` even allowed?

include/deal.II/lac/superlu_mt.h
```diff
@@ -0,0 +1,190 @@
...
+  using size_type = types::global_dof_index;
+
+
+  // \todo remove this issue below
+  /**
+   * in order to avoid the error with
+   * <c> invalid conversion from ‘unsigned int*’ to ‘int_t* {aka int*}’ </c>
+   * we create a temporary type. Ideally this should be same as \c size_type
+   */
+  using my_int = int_t;
```
bangerth, line 58: what is `int_t`? Is this a thing that SuperLU defines?

bangerth:
I've been reading through this again, and the longer I think, the more I believe that using the Amesos2 interfaces is the right way to go. That's in part because we wouldn't have to write the data structure conversion functionality, but also because we will get not only SuperLU, but also parallel versions as well as the much newer STRUMPACK interface for basically free. I think that would make a lot of things possible.

@KParas You mentioned that you had problems with the Amesos interfaces. Can you explain what those were, and maybe whether they still exist 18 months after this thread started?

KParas:
> @KParas You mentioned that you had problems with the Amesos interfaces. Can you explain what those were, and maybe whether they still exist 18 months after this thread started?

@bangerth since it has been a while now, I don't remember the exact error. IIRC, there were two main issues:
-  Firstly, having a serial code based on `dealii::Vector`, I was facing troubles while copying it into a `TrilinosWrappers::MPI:Vector` which would be necessary for using the Trilinos interface.
- Secondly, I was also facing some issues linking Superlu-mt with Amesos and then use it in dealii. I used spack for installing the libraries.

bangerth:
So how do we want to move forward?

peterrum:
@KParas Did you make some progress here?

KParas:
> @KParas Did you make some progress here?

@peterrum Unfortunately not. The current stand is as I mentioned above.

bangerth:
Needs to wait for the next release.

bangerth:
@KParas Same question as last year?

KParas:
> @KParas Same question as last year?

@bangerth Sorry, no further progress. As mentioned last time, I would require some "guidance from more experienced people" to proceed further.

bangerth:
Let's close this PR. We can always re-open it if anyone has the time and interest in working on it, but nobody has found the time to do so in the past several years.

Part of my thoughts about this are that
* SuperLU_MT is no longer actively developed. The team has moved on to SuperLU_dist instead
* If we go with using a direct solver via its interfaces to other software, then we could instead access SuperLU via PETSc or Amesos.
* But, perhaps more interestingly, Trilinos has a package ShyLU that has thread-parallel sparse direct solvers built on Kokkos that are actively developed. See https://trilinos.github.io/shylu.html and https://arxiv.org/pdf/1601.05725.pdf . We should go with these more modern variants instead.

PR #9829, opened 2020-04-05
FETools: Add additional lexicographic_to_hierarchic_numbering functions

kronbichler:
While working on #9826, I realized that `FETools::lexicographic_to_hierarchic_numbering` is a pain to use because we need to provide the result array as argument. This PR adds the functions that return the respective arrays instead, as we have done for a number of other functions in the hierarchic to lexicographic context. Given this new function, I can now also simplify `fe_q_base` and `mapping_q_eulerian` which both used to construct a more complicated data structure to then return things.

I do not yet apply them to `MappingQGeneric` because that is the topic of #9826, but I add that together these functions save around 55 lines of code in already big implementations (`fe_q_base`, `mapping_q_generic`, `mapping_q_eulerian`) by augmenting some tools functions that are well-separated, so I consider this a clear win.

### DIFF
--- a/include/deal.II/fe/fe_tools.h
+++ b/include/deal.II/fe/fe_tools.h
@@ -895,75 +895,72 @@ namespace FETools
    * in y-direction and finally in z-direction. Discontinuous elements of
    * class FE_DGQ() are numbered in this way, for example.
    *
-   * This function constructs a table which lexicographic index each degree of
-   * freedom in the hierarchic numbering would have. It operates on the
-   * continuous finite element given as first argument, and outputs the
-   * lexicographic indices in the second.
-   *
-   * Note that since this function uses specifics of the continuous finite
-   * elements, it can only operate on FiniteElementData<dim> objects inherent
-   * in FE_Q(). However, this function does not take a FE_Q object as it is
-   * also invoked by the FE_Q() constructor.
-   *
-   * It is assumed that the size of the output argument already matches the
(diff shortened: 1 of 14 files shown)


include/deal.II/fe/fe_tools.h
```diff
@@ -909,12 +909,23 @@ namespace FETools
...
   hierarchic_to_lexicographic_numbering(unsigned int               degree,
                                         std::vector<unsigned int> &h2l);
 
+  /**
+   * Like the previous function but instead of returning its result through
+   * the last argument return it as a value.
+   */
+  template <int dim>
+  std::vector<unsigned int>
+  hierarchic_to_lexicographic_numbering(unsigned int degree);
```
drwells, line 923: Could we deprecate the corresponding older functions?

kronbichler: Yes, that is a good idea; in my opinion we should deprecate all functions but the two using the `degree` argument and returning a vector, so also the ones taking `FiniteElementData`. These functions are mostly used within the library and one can always easily get the degree from the finite element data, so I would just keep the simplest form possible. Let me add that as a commit to this PR.

kronbichler:
@drwells I now deprecated the functions and transitioned the whole library including tests (but excluding `mapping_q_generic` which is in the other PR) to the new scheme. Ready for re-review.

kronbichler:
@drwells anything more you want me to address here?

drwells:
Nope, we're set!

PR #9831, opened 2020-04-05
Implement functionality from Hypercube on BoundingBox

simonsticko:
The `Hypercube` class added in #9490 is similar to the already existing class `BoundaryBox`. To avoid duplication, implement the non-existing functionality on `BoundingBox` and remove the `Hypercube` class.


include/deal.II/base/bounding_box.h
```diff
@@ -69,24 +69,27 @@ enum class NeighborType
 };
 
 /**
- * A class that represents a bounding box in a space with arbitrary dimension
- * <tt>spacedim</tt>.
+ * A class that represents a box of arbitrary dimension <tt>spacedim</tt> and
+ * with sides parallel to the coordinate axes. That is, a region
```
kronbichler, line 73:
```suggestion
 * with sides parallel to the coordinate axes, that is, a region
```

include/deal.II/base/bounding_box.h
```diff
@@ -193,6 +267,61 @@ class BoundingBox
   std::pair<Point<spacedim, Number>, Point<spacedim, Number>> boundary_points;
 };
 
+
+/**
+ * Returns the unit box [0,1]^dim.
```
bangerth, line 272:
```suggestion
 * Returns the unit box $[0,1]^\text{dim}$.
```

include/deal.II/base/bounding_box.h
```diff
@@ -193,6 +267,61 @@ class BoundingBox
...
+ */
+template <int dim, typename Number = double>
+BoundingBox<dim, Number>
+create_unit_box();
```
bangerth, line 276: May I ask to just rename this function to `create_unit_bounding_box()` to make its name conform to the classname? You might always want to put `@relates BoundingBox` into the documentation so that it shows up on the relevant doxygen page.

include/deal.II/base/bounding_box.h
```diff
@@ -193,6 +267,61 @@ class BoundingBox
...
+ * dimension to a coordinate index in dim + 1 dimensions.
+ *
+ * @param locked_coordinate should be in the range [0, dim+1).
+ * @param coordiante_in_dim should be in the range [0, dim).
+ * @return A coordinate index in the range [0, dim+1)
+ */
+template <int dim>
+inline unsigned int
+coordinate_to_one_dim_higher(const unsigned int locked_coordinate,
+                             const unsigned int coordiante_in_dim)
```
bangerth, line 317: Is this a function of general interest? Otherwise, it might just go into the `.cc` file or an anonymous namespace.

Same with `@relates` as above.

simonsticko: No, I can't see that it would be useful externally. But for the functionality I'm working on, I will need it internally in more places than in this class, so I would like to keep it in the header. I have moved it to an internal namespace in #9852 .

PR #9834, opened 2020-04-06
Step-63: Fix two typos

tamiko:


drwells:
It's just a documentation change so lets merge.

PR #9835, opened 2020-04-06
Parallelize the main loops of MappingInfo::compute_mapping_q

kronbichler:
This is a follow-up to #9656 and thread-parallelizes the initialization routines for the new path using MappingQ. The change consists of two commits, the first merely re-organizing some code and putting things into separate functions (to make another loop parallelizable), and the second one to actually parallelize things. So the interesting part is the second commit. This passes all tests including `matrix_free` in their name locally.

### DIFF
--- a/include/deal.II/matrix_free/mapping_info.templates.h
+++ b/include/deal.II/matrix_free/mapping_info.templates.h
@@ -1100,6 +1100,154 @@ namespace internal
 
 
       /**
+       * This invokes the FEValues part of the initialization of MappingQ,
+       * storing the resulting quadrature points and an initial representation
+       * of Jacobians in two arrays.
+       */
+      template <int dim>
+      void
+      mapping_q_query_fe_values(
+        const unsigned int                                        begin_cell,
+        const unsigned int                                        end_cell,
+        const MappingQGeneric<dim> &                              mapping_q,
+        const dealii::Triangulation<dim> &                        tria,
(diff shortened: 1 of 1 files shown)


kronbichler:
*ping*

kronbichler:

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -2366,114 +2517,44 @@ namespace internal
...
+        // Create as many chunks of cells as we have threads and spawn the
+        // work
+        unsigned int work_per_chunk =
+          std::max(std::size_t(1),
+                   (cell_array.size() + MultithreadInfo::n_threads() - 1) /
+                     MultithreadInfo::n_threads());
+
+        // we manually use tasks here rather than parallel::apply_to_subranges
+        // because we want exactly as many loops as we have threads - the
+        // initialization of the loops with FEValues is expensive
```
kronbichler, line 2535: This should answer your question: The reason is that `apply_to_subranges` does dynamic loop scheduling with increasingly smaller subsets. For this first loop, it is simply not worth it because the initialization of `FEValues` is pretty expensive. The other loops are cheaper to initialize so we can tolerate it. Of course, one might still wonder whether the dynamic scheduling is helpful for performance or not (I am sceptical), but I think that is the most easily understandable form.

peterrum: OK! Makes sense! Thanks for the explanation!

PR #9836, opened 2020-04-06
Make TpetraWrappers appear in Doxygen

peterrum:
I had to also disable the following line:

https://github.com/dealii/dealii/blob/fec0887042e095a5c3dfeae829a5bd626e5f5b6d/include/deal.II/lac/trilinos_tpetra_vector.h#L22

to make it visible, but I guess `DEAL_II_TRILINOS_WITH_TPETRA` is defined for the online documentation?

### DIFF
--- a/include/deal.II/lac/trilinos_tpetra_vector.h
+++ b/include/deal.II/lac/trilinos_tpetra_vector.h
@@ -48,6 +48,13 @@ namespace LinearAlgebra
   class ReadWriteVector;
 #  endif
 
+  /**
+   * A namespace for classes that provide wrappers for Trilinos' Tpetra vectors.
+   *
+   * This namespace provides wrappers for the Tpetra::Vector class from the
+   * Tpetra package (https://trilinos.github.io/tpetra.html) that is part of
+   * Trilinos.
+   */
   namespace TpetraWrappers
   {
     /**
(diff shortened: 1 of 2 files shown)


jppelteret:
> but I guess DEAL_II_TRILINOS_WITH_TPETRA is defined for the online documentation

Looks to me like this is not the case. The predefined macros are set in [doc/doxygen/options.dox.in](https://github.com/dealii/dealii/blob/master/doc/doxygen/options.dox.in#L171-L225).

peterrum:
@jppelteret Thanks for the hint! I will update the PR!

include/deal.II/lac/trilinos_tpetra_vector.h
```diff
@@ -48,6 +48,13 @@ namespace LinearAlgebra
   class ReadWriteVector;
 #  endif
 
+  /**
+   * A namespace for classes that provide wrappers for Trilinos' Tpetra vectors.
+   *
+   * This namespace provides wrappers for the Epetra_FEVector class from the
```
jppelteret, line 54: Just checking: Is 
> the Epetra_FEVector class

a copy-paste error? I'll also check the link and source files, but maybe you'll beat me to the answer in a reply.

jppelteret: Yeah, I think that this should be `Tpetra::Vector`.

peterrum: This was a copy-and-past error. I have changed it. Your hint with the doxygen configuration file also worked!

jppelteret: Great, thanks a lot for fixing this!

Issue #9839, opened 2020-04-06
Merge DoFHandler and hp::DoFHandler


peterrum:
The purpose of this issue is to collect the PRs related to merging `DoFHandler` and `hp::DoFHanlder` (as explained in https://github.com/dealii/dealii/pull/9760#issue-395116540). The first phase includes the introduction of an `(internal::)DoFHandlerBase`.

peterrum:
I have closed the two PRs related to this issue for now (#9840 and #9760). I will open a new PR targeting the merge of `DoFHandler` with `hp::DoFHandler` after the release of 9.2.

A first implementation shows that merging `DoFHandler` and `hp::DoFHandler` could be possible without any major compatibility issues if the internal data structures are changed and merged. The current code can be found at https://github.com/peterrum/dealii/pull/3. All tests are passing. However, I still have to go through the whole code at least once again.

bangerth:
I think that's the right approach -- let's wait until after 9.2.

If you have smaller pieces that only move code around in ways not visible to users (e.g., does not move public member functions), then I'd be happy to consider them separately now already.

PR #9840, opened 2020-04-06
Move implementation of DoFHandler and hp::DoFHandler [WIP]

peterrum:
This PR is related to #9839.

It moves:
* the inline functions from `dof_handler.h` to the new `dof_handler_base.h` file (the inline functions in `hp::dof_handler.h` are practically the same functions and can be removed after a `DoFHandlerBase` has been introduced).
* the implementations form the `dof_handler.cc`-s to the new `dof_handler_base.templates.h` file

The order of the files in `dof_handler_base.templates.h` are more or less the same as it used to be, however, I positioned the `DoFHandler`- implementations and the `hp::DoFHandler`-implementation next to each other, since the next step will be merge these (and a consequence the diff will be more easily to read).


kronbichler:
This is split out from #9760.

PR #9842, opened 2020-04-06
Output step-62 results as a complex-valued vector

dangars:
This is a follow-up to #9815. Now the results of step-62 can be viewed  as a vector in Paraview.

The PR #9815 has to be merged before this PR is merged.

### DIFF
--- a/examples/step-62/step-62.cc
+++ b/examples/step-62/step-62.cc
@@ -1105,17 +1105,10 @@ namespace step62
     // been described in step-40.
     if (parameters.save_vtu_files)
       {
-        std::vector<std::string> solution_names(1, "displacement_x");
-        if (dim >= 2)
-          {
-            solution_names.emplace_back("displacement_y");
-          }
-        if (dim == 3)
-          {
-            solution_names.emplace_back("displacement_z");
-          }
+        std::vector<std::string> solution_names(dim, "displacement");
         std::vector<DataComponentInterpretation::DataComponentInterpretation>
-          interpretation(dim, DataComponentInterpretation::component_is_scalar);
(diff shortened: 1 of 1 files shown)


dangars:
I wonder if the results "Force" and "PML" could be ouput as vectors. I do not think that is possible, but let me know if I'm wrong. See lines 1135 to 1138
```c++
        for (unsigned int dim_idx = 0; dim_idx < dim; ++dim_idx)
          {
            data_out.add_data_vector(force[dim_idx],
                                     "force_" + std::to_string(dim_idx));
            data_out.add_data_vector(pml[dim_idx],
                                     "pml_" + std::to_string(dim_idx));
          }
```

bangerth:
About the force and PML vectors: The way `DataOut` treats vectors with a number of elements equal to the number of cells is as scalars. If you wanted to output these as vectors, you somehow have to define a `DoFHandler` with a vector-valued element that represents this information. This could, for example, be a `FESystem(FE_DGQ(0), dim)` of piecewise constants, but you do need a `DoFHandler` for anything to be interpreted as a vector or tensor.

dangars:
> About the force and PML vectors: The way `DataOut` treats vectors with a number of elements equal to the number of cells is as scalars. If you wanted to output these as vectors, you somehow have to define a `DoFHandler` with a vector-valued element that represents this information. This could, for example, be a `FESystem(FE_DGQ(0), dim)` of piecewise constants, but you do need a `DoFHandler` for anything to be interpreted as a vector or tensor.

@bangerth Thanks for for suggestion, I'm going to try it and then I'll update step-62

PR #9843, opened 2020-04-06
Do not verify MPI_HAVE_MPI_SEEK_SET during cross compiling

dangars:
This is a follow-up to #9778. I find it convenient to cross-compile deal.II in my machine and then upload it to the cluster. It works quite well, but I have to remove the requirement `MPI_HAVE_MPI_SEEK_SET`.

The property `MPI_HAVE_MPI_SEEK_SET` can not be verified during cross-compiling. I guess that CMake has to compile a program and then execute it to verify `MPI_HAVE_MPI_SEEK_SET`. During cross-compiling that is not always possible.

### DIFF
--- a/cmake/configure/configure_1_mpi.cmake
+++ b/cmake/configure/configure_1_mpi.cmake
@@ -30,7 +30,8 @@ MACRO(FEATURE_MPI_FIND_EXTERNAL var)
   IF(MPI_FOUND)
     SET(${var} TRUE)
 
-    IF(NOT MPI_HAVE_MPI_SEEK_SET)
+    # MPI_HAVE_MPI_SEEK_SET can not be verified during cross compiling
+    IF((NOT MPI_HAVE_MPI_SEEK_SET) AND (NOT CMAKE_CROSSCOMPILING))
       MESSAGE(STATUS
         "Could not find a sufficient MPI version: "
         "Your MPI implementation must define MPI_SEEK_SET.")


cmake/configure/configure_1_mpi.cmake
```diff
@@ -30,7 +30,8 @@ MACRO(FEATURE_MPI_FIND_EXTERNAL var)
   IF(MPI_FOUND)
     SET(${var} TRUE)
 
-    IF(NOT MPI_HAVE_MPI_SEEK_SET)
+    # MPI_HAVE_MPI_SEEK_SET can not be verified during cross compiling
+    IF((NOT MPI_HAVE_MPI_SEEK_SET) AND (NOT CMAKE_CROSSCOMPILING))
```
bangerth, line 34: Is `CMAKE_CROSSCOMPILING` an actual cmake thing, or is this something that you set locally in your build scripts?

drwells: It's a boolean set by cmake: see also

https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html

dangars:
> This seems reasonable. Are there any other features that are problematic when we cross compile?

Everything works fine. The only issue that I found is that the [rpath](https://en.wikipedia.org/wiki/Rpath) of the examples is set relative to the host and not the target. This is a minor issue and I do not think that it is worthy to fix it given that cross-compiling is a narrow topic. It is very easy to fix this afterwards using `LD_LIBRARY_PATH` or `chrpath`. I'm preparing a PR where I update the cross-compiling documentation to explain all this.

cmake/configure/configure_1_mpi.cmake
dangars: The variable `CMAKE_CROSSCOMPILING` will be set to true automatically when you define the CMake cross compiling toolchain using the command line option `-DCMAKE_TOOLCHAIN_FILE=file.cmake`

dangars:
The updated documentation for cross compilation can be found in #9854.

tamiko:
Before we merge this I would like to get this more in line on how the `DEAL_II_PLATFORM_INTROSPECTION` toggle works.

tamiko:
@dangars You are saying that in a cross-compilation scenario the following lines in `FindMPI.cmake` fail?
```
 58 #                                                                                
 59 # Older versions of MPI may not have MPI_SEEK_SET, which we                      
 60 # require. Strangely, unlike MPICH, OpenMPI needs the correct link libraries     
 61 # for this to compile, not *just* the correct include directories.               
 62 #                                                                                
 63                                                                                  
 64 CLEAR_CMAKE_REQUIRED()                                                           
 65 SET(CMAKE_REQUIRED_FLAGS ${DEAL_II_CXX_FLAGS_SAVED} ${MPI_CXX_COMPILE_FLAGS} ${MPI_CXX_LINK_FLAGS})
 66 SET(CMAKE_REQUIRED_INCLUDES ${MPI_CXX_INCLUDE_PATH})                             
 67 SET(CMAKE_REQUIRED_LIBRARIES ${DEAL_II_LINKER_FLAGS_SAVED} ${MPI_LIBRARIES})     
 68 CHECK_CXX_SOURCE_COMPILES(                                                       
 69   "                                                                              
 70   #include <mpi.h>                                                               
 71   #ifndef MPI_SEEK_SET                                                           
 72   #  error                                                                       
 73   #endif                                                                         
 74   int main() {}                                                                  
 75   "                                                                              
 76   MPI_HAVE_MPI_SEEK_SET                                                          
 77   )                                                                              
 78 RESET_CMAKE_REQUIRED() 
```

Would you mind sharing the error message in `CMakeError.log` that you produce?

dangars:
@tamiko You are right!
I didn't asses the problem correctly. There was a library that was not linked properly.
In the cluster the library is found with `LD_LIBRARY_PATH`. To solve the problem I set `LD_LIBRARY_PATH` before running CMake.
Surprisingly this was not a problem for the compilation of deal.II.
I'll add a phrase regarding this potential pitfall in the cross-compilation documentation in #9854

tamiko:
@dangars Ah! That explains a lot :smile:

PR #9844, opened 2020-04-06
Build debug and release with Actions

### LINKED ISSUE #9613: CI: add a check that compiles in release mode
Currently all testers run in debug mode, which means we don't catch warnings like #9609.  I need to think about if we want to add another configuration or change one of the existing ones.

masterleinad:
Fixes #9613.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
         cmake --version
     - name: configure
       run: |
-        cmake -D CMAKE_BUILD_TYPE=Debug -D DEAL_II_CXX_FLAGS='-Werror' .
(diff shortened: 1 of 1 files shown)


drwells:
I verified that the CI output contains debug and release builds so this works as expected.

tjhei:
This means that we have to wait at least 2 hours for each PR. I am not sure that is worth it?

drwells:
That's a good point. What if we made one debug and one release?

tjhei:
> That's a good point. What if we made one debug and one release?

There are stricter limits for OSX machines [1] and I think they are also less reliable and slower to start up. I would use a linux build for release build testing. 

[1]: https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions

tjhei:
See https://github.com/tjhei/dealii/runs/567870188 (I will create a PR after it passes)

PR #9845, opened 2020-04-07
Fix some Doxygen warnings

rezarastak:
I tried to reduce the number of warnings in `build/doxygen.log` by:
 * making sure the function definitions in source files perfectly match their signature in header files. This way Doxygen can put a link in the documentation of that function to the specific line in the source file (`.cc`).
 * putting `#ifndef DOXYGEN` around some of the template specialization instantiations.
 * fixing some typos.

With this patch the number of warnings in `doxygen.log` decreased from 364 to 327, and the total number of lines in `doxygen.log` decreased from 2431 to 2041.


bangerth:
The OS X build fails, but the CI output doesn't list any messages. I'm assuming that it's a spurious failure and merge anyway.

PR #9847, opened 2020-04-07
Rename NoncontiguousPartitioner::update_values()

peterrum:
Togehter with @kronbichler, we are indending to add a `compression()` function to the class `NoncontiguousPartitioner`. To make this class more consistent with `Partitioner`, this PR renames `NoncontiguousPartitioner::update_values()` to `NoncontiguousPartitioner::update_ghost_values()`.

### DIFF
--- a/include/deal.II/base/mpi_noncontiguous_partitioner.h
+++ b/include/deal.II/base/mpi_noncontiguous_partitioner.h
@@ -38,7 +38,6 @@ namespace Utilities
      *
      * @author Peter Munch, 2020
      */
-    template <typename Number = double>
     class NoncontiguousPartitioner
       : public dealii::LinearAlgebra::CommunicationPatternBase
     {
@@ -51,16 +50,16 @@ namespace Utilities
 
       /**
        * Constructor. Set up point-to-point communication pattern based on the
-       * IndexSets arguments @p indexset_has and @p indexset_want for the MPI
+       * IndexSets arguments @p indexset_locally_owned and @p indexset_ghost for the MPI
        * communicator @p communicator.
        */
-      NoncontiguousPartitioner(const IndexSet &indexset_has,
-                               const IndexSet &indexset_want,
+      NoncontiguousPartitioner(const IndexSet &indexset_locally_owned,
+                               const IndexSet &indexset_ghost,
                                const MPI_Comm &communicator);
 
       /**
-       * Constructor. Same as above but for vectors of indices @p indices_has
-       * and @p indices_want. This allows the indices to not be sorted and the
+       * Constructor. Same as above but for vectors of indices @p indices_locally_owned
+       * and @p indices_ghost. This allows the indices to not be sorted and the
        * values are read and written automatically at the right position of
        * the vector during update_values(), update_values_start(), and
        * update_values_finish(). It is allowed to include entries with the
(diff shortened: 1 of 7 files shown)


kronbichler:
We currently do not have the names `update_ghost_values` and `compress` in `Partitioner` (but `export_to_ghosted_array` and `import_from_ghosted_array`) - do we want to unify names? The problem with `compress` is that we have previously associated the operation with the object being compressed, whereas the here the compression goes to one of the arguments we are passing in.

peterrum:
@kronbichler You are right! I mixed the names with those of `Vector`. Let me fix this.

peterrum:
@kronbichler I have made the name change. However, now I am wondering if I should't also adjust the parameters. In particular, the `Partitioner` class does not own the MPI requests and the temporal arrays. Should I change `NoncontiguousPartitioner` so that it resembles `Partitioner` in that aspect also?

masterleinad:
Should this still be WIP?

peterrum:
Yes I am waiting for feedback from @kronbichler. We might have plans to use the class in a different context than it was actually designed for...

masterleinad:
We need to decide if it should be tagged for the `9.2` release or not then.

peterrum:
Yes it should. It is just a simple renaming.

kronbichler:
> In particular, the `Partitioner` class does not own the MPI requests and the temporal arrays. Should I change `NoncontiguousPartitioner` so that it resembles Partitioner in that aspect also?

That question is not easy to answer - from the point of view of responsibilities, they indeed do not belong to the partitioner object, but to whoever is calling them. This is reflected in the basic `Partitioner` class, allowing several vectors of the same partitioner to do an exchange at the same time. In this class, we cannot do it when the temporary data structures are internal. On the other hand, this class is more user-visible than the standard partitioner, and its only intent is to perform this  kind of exchange. At least at the current point in time.
So my suggestion would be as follows:
- If we do not envision other uses of this class, we could keep the arguments currently proposed, even though they differ from the basic partitioner.
- To increase compatibility, and ff the arguments we need for the vector exchange can be indeed the same as in the `Partitioner` exchange model, we could add a second set of  `export_to_ghosted_array` methods with the `Partitioner` interface, and let the implementation of the compact `export_to_ghosted_array` functions go to this function with the internal data structures.

The thing that is not as nice is that we cannot create a free function to do the nice interface, because we need the internal data like the MPI requests to survive between two calls.

peterrum:
@kronbichler I like the suggestion with the second set of methods. Let me implement it today.

peterrum:
@kronbichler I made the change. Now there 4 methods:

```cpp
      template <typename Number>
      void
      export_to_ghosted_array(
        const ArrayView<const Number> &locally_owned_array,
        const ArrayView<Number> &      ghost_array) const;

      template <typename Number>
      void
      export_to_ghosted_array(
        const unsigned int             communication_channel,
        const ArrayView<const Number> &locally_owned_array,
        const ArrayView<Number> &      temporary_storage,
        const ArrayView<Number> &      ghost_array,
        std::vector<MPI_Request> &     requests) const;

      template <typename Number>
      void
      export_to_ghosted_array_start(
        const unsigned int             communication_channel,
        const ArrayView<const Number> &locally_owned_array,
        const ArrayView<Number> &      temporary_storage,
        std::vector<MPI_Request> &     requests) const;

      template <typename Number>
      void
      export_to_ghosted_array_finish(
        const ArrayView<const Number> &temporary_storage,
        const ArrayView<Number> &      ghost_array,
        std::vector<MPI_Request> &     requests) const;
```

What do you think?

include/deal.II/base/mpi_noncontiguous_partitioner.h
```diff
@@ -197,14 +221,22 @@ namespace Utilities
       std::vector<types::global_dof_index> recv_indices;
 
       /**
-       * Buffer containing the values sorted by rank.
+       * Buffer containing the values sorted by rank for sending and receiving.
+       *
+       * @note Only allocated if not provided externally by user.
+       *
+       * @note At this place we do not know the type of the data to be sent. So
+       *   we use an arbitrary type of size 1 byte. The type is cased to the
```
kronbichler, line 229:
```suggestion
       *   we use an arbitrary type of size 1 byte. The type is cast to the
```

include/deal.II/base/mpi_noncontiguous_partitioner.h
```diff
@@ -84,26 +83,61 @@ namespace Utilities
        *   functions separately and hereby overlap communication and
        *   computation.
        */
-      template <typename VectorType>
+      template <typename Number>
       void
-      update_values(VectorType &dst, const VectorType &src) const;
+      export_to_ghosted_array(
+        const ArrayView<const Number> &locally_owned_array,
+        const ArrayView<Number> &      ghost_array) const;
```
kronbichler, line 90: Can you add a comment to this function (and maybe some of the functions below) how large the array is supposed to be? It should correspond to `indices_has.size()` and `indices_want.size()` if I remember correctly.

include/deal.II/base/mpi_noncontiguous_partitioner.h
```diff
@@ -84,26 +83,61 @@ namespace Utilities
...
+       * Utilities::MPI::Partitioner::export_to_ghosted_array_finish. In this
+       * function, the user can provide the temporary data structures to be
+       * used.
+       */
+      template <typename Number>
+      void
+      export_to_ghosted_array(
+        const unsigned int             communication_channel,
+        const ArrayView<const Number> &locally_owned_array,
+        const ArrayView<Number> &      temporary_storage,
```
kronbichler, line 104: How large do we expect `temporary_storage` to be?

include/deal.II/base/mpi_noncontiguous_partitioner.h
```diff
@@ -84,26 +83,61 @@ namespace Utilities
...
+        std::vector<MPI_Request> &     requests) const;
 
       /**
        * Start update: Data is packed, non-blocking send and receives
        * are started.
+       *
+       * @note In contrast to the function
+       *   Utilities::MPI::Partitioner::export_to_ghosted_array_start, the user
+       *   does not pass a reference to the destination vector, since the data
+       *   is received into a designated part of the buffer.
```
kronbichler, line 115:
```suggestion
       *   is received into a designated part of `temporary_storage`. This setup
       *   allows for padding and other post-processing of data.
```

include/deal.II/base/mpi_noncontiguous_partitioner.h
```diff
@@ -84,26 +83,61 @@ namespace Utilities
...
+        std::vector<MPI_Request> &     requests) const;
 
       /**
        * Finish update. The method waits until all data has been sent and
        * received. Once data from any process is received it is processed and
        * placed at the right position of the vector @p dst.
+       *
+       * @note In contrast to the function
+       *   Utilities::MPI::Partitioner::export_to_ghosted_array_finish, the user
+       *   also has to pass a reference to the buffer, since the data has been
```
kronbichler, line 132:
```suggestion
       *   also has to pass a reference to the buffer of `temporary_storage`, since the data has been
```

peterrum:
@kronbichler Thanks for the comments! The sizes were indeed missing. I have updated the comments.

kronbichler:
Thank you.

peterrum:
@kronbichler  I will add a compress function after release. Let's hope we can use this class as planed!

PR #9850, opened 2020-04-07
Document the mental model of DiscreteTime.

bangerth:
@rezarastak -- FYI. I would like us to change the function names to `set_desired_next_time_step()` and `get_actual_next_time_step()` to make the distinction between the two clearer. What do you think of this?

(This is just the documentation part and can be merged independently. I'll wait with the implementation until after #9748 has been merged.)

### DIFF
--- a/include/deal.II/base/discrete_time.h
+++ b/include/deal.II/base/discrete_time.h
@@ -21,13 +21,21 @@
 DEAL_II_NAMESPACE_OPEN
 
 /**
- * Provides a means to keep track of the simulation time in a time-dependent
- * simulation. It manages stepping forward from an initial time to a final
- * time. It also allows adjusting the time step size during the simulation.
- * It is guaranteed that at all times the current simulation time is in the
- * closed interval between the start time and the end time.
+ * This class provides a means to keep track of the simulation time in a
+ * time-dependent simulation. It manages stepping forward from a start time
+ * $T_{\text{start}}$ to an end time $T_{\text{end}}$. It also allows adjusting
+ * the time step size during the simulation. It is guaranteed that at all times
+ * the current simulation time is in the closed interval between the start time
+ * and the end time.
  *
- * You can loop over all of the time steps by using a for loop
+ * The model this class follows is that one sets a *desired* time step length
+ * either through the constructor or using set_desired_next_step_size()
+ * function. This step size will then be used in all following calls to the
+ * advance_time() function, but may be adjusted slightly towards the end of the
+ * simulation to ensure that the simulation time hits the end time exactly. The
+ * adjustment is useful for the following reasons:
+ *
+ * Let's say that you loop over all of the time steps by using a for loop
  * @code
  *   for (DiscreteTime time(0., 1., 0.3);
  *        time.get_current_time() != time.get_end_time();
@@ -36,13 +44,39 @@ DEAL_II_NAMESPACE_OPEN
  *     // Insert simulation code here
  *   }
  * @endcode
+ * In the above example the time starts at $T_{\text{start}} = 0$ until
+ * $T_{\text{end}}=1$. Assuming the time step $dt = 0.3$ is not modified inside
+ * the loop, the time is advanced from $t = 0$ to $t = 0.3$, $t = 0.6$, $t =
+ * 0.9$ and finally it reaches the end time at $t = 1.0$. Here, the final step
+ * size needs to be reduced from its desired value of 0.3 to $dt = 0.1$ in order
+ * to ensure that we finish the simulation exactly at the specified end time. In
+ * fact, you should assume that not only the last time step length may be
+ * adjusted, but also previously ones -- for example, this class may take the
+ * liberty to spread the decrease in time step size out over several time steps
+ * and increment time from $t=0$, to $0.3$, $0.6$, $0.8$, and finally
+ * $t=T_{\text{end}}=1$ to avoid too large a change in time step size from one
+ * step to another.
  *
- * In the above example the time starts at $t = 0$. Assuming the time step
- * $dt = 0.3$ is not modified inside the loop, the time is advanced to
- * $t = 0.3$, $t = 0.6$, $t = 0.9$ and finally it reaches the end time at
- * $t = 1.0$. Note that the final step size is automatically reduced to
- * $dt = 0.1$ in order to ensure that we finish the simulation exactly at
- * the specified end time.
+ * The other situation in which the time step needs to be adjusted (this time to
+ * slightly larger values) is if a time increment falls just short of the final
+ * time. Imagine, for example, a similar situation as above, but with different
+ * end time:
+ * @code
+ *   for (DiscreteTime time(0., 1.21, 0.3);
+ *        time.get_current_time() != time.get_end_time();
+ *        time.advance_time())
+ *   {
+ *     // Insert simulation code here
+ *   }
+ * @endcode
+ * Here, the time step from $t=0.9$ to $t=1.2$ falls just short of the final
+ * time $T_{\text{end}}=1.21$. Instead of following up with a very small step of
+ * length $dt=0.01$, the class stretches the last time step (or last time steps)
+ * slightly to reach the desired end time.
+ *
+ * The examples above make clear that the time step size given to this class is
+ * only a *desired* step size. You can query the actual time step size using the
+ * get_next_step_size() function.
  *
  * @author Reza Rastak, 2019
  */
@@ -79,7 +113,11 @@ public:
   get_end_time() const;
 
   /**
-   * Return the size of the step from current time step to the next.
+   * Return the size of the step from current time step to the
+   * next. As discussed in the introduction to the class, this is the
+   * *actual* time step, and may differ from the *desired* time step
+   * set in the constructor or through the
+   * set_desired_next_step_size() function.
    */
   double
   get_next_step_size() const;


rezarastak:
Thank you helping to clarify the documentation of `DiscreteTime`. I like the name `set_desired_next_time_step()`, or alternatively `suggest_next_step_size()` or `propose_next_step_size()`. I think we can keep the name `get_next_time_step()` as is. Every function is assumed to provide real and actual values unless denoted otherwise. So the prefix `actual` is not needed.

include/deal.II/base/discrete_time.h
```diff
@@ -21,13 +21,20 @@
...
 
 /**
- * Provides a means to keep track of the simulation time in a time-dependent
- * simulation. It manages stepping forward from an initial time to a final
- * time. It also allows adjusting the time step size during the simulation.
- * It is guaranteed that at all times the current simulation time is in the
- * closed interval between the start time and the end time.
+ * This class provides a means to keep track of the simulation time in a
+ * time-dependent simulation. It manages stepping forward from an initial time
+ * to a final time. It also allows adjusting the time step size during the
```
rezarastak, line 26:
```suggestion
 * time-dependent simulation. It manages stepping forward from a start time $T_{\text{start}}$
 * to an end time $T_{\text{end}}$. It also allows adjusting the time step size during the
```
We can use the start-time/end-time idiom to be consistent with our function names. What do you think?

include/deal.II/base/discrete_time.h
```diff
@@ -36,13 +43,38 @@ DEAL_II_NAMESPACE_OPEN
  *     // Insert simulation code here
  *   }
  * @endcode
- *
- * In the above example the time starts at $t = 0$. Assuming the time step
- * $dt = 0.3$ is not modified inside the loop, the time is advanced to
+ * In the above example the time starts at $t = 0$ until $T=1$. Assuming the
+ * time step $dt = 0.3$ is not modified inside the loop, the time is advanced to
```
rezarastak, line 47:
```suggestion
 * time step $dt = 0.3$ is not modified inside the loop, the time is advanced from $t = 0$ to
```

include/deal.II/base/discrete_time.h
```diff
@@ -36,13 +43,38 @@ DEAL_II_NAMESPACE_OPEN
...
- * $t = 1.0$. Note that the final step size is automatically reduced to
- * $dt = 0.1$ in order to ensure that we finish the simulation exactly at
- * the specified end time.
+ * $t = 1.0$. Here, the final step size needs to be reduced from its desired
+ * value of 0.3 to $dt = 0.1$ in order to ensure that we finish the simulation
+ * exactly at the specified end time. In fact, you should assume that not only
+ * the last time step length may be adjusted, but also previously ones -- for
+ * example, this class may take the liberty to spread the decrease in time step
+ * size out over several time steps and increment time from $t=0$, to $0.3$,
+ * $0.6$, $0.8$, and finally $t=T=1$ to avoid too large a change in time step
```
rezarastak, line 55:
```suggestion
 * $0.6$, $0.8$, and finally $t=T_{\text{end}}=1$ to avoid too large a change in time step
```

include/deal.II/base/discrete_time.h
```diff
@@ -36,13 +43,38 @@ DEAL_II_NAMESPACE_OPEN
...
+ * @code
+ *   for (DiscreteTime time(0., 1.21, 0.3);
+ *        time.get_current_time() != time.get_end_time();
+ *        time.advance_time())
+ *   {
+ *     // Insert simulation code here
+ *   }
+ * @endcode
+ * Here, the time step from $t=0.9$ to $t=1.2$ falls just short of the final
+ * time $T=1.21$. Instead of following up with a very small step of length
```
rezarastak, line 71:
```suggestion
 * time $T_{\text{end}}=1.21$. Instead of following up with a very small step of length
```

include/deal.II/base/discrete_time.h
```diff
@@ -36,13 +43,38 @@ DEAL_II_NAMESPACE_OPEN
...
+ * $dt=0.01$, the class stretches the last time step (or last time steps)
+ * slightly to reach the desired end time.
+ *
+ * The examples above make clear that the time step size given to this class is
+ * only a *desired* step size. You can query the actual time step size using the
+ * get_actual_next_step_size() function.
```
rezarastak, line 77:
```suggestion
 * get_next_step_size() function.
```

jppelteret:
< get_actual_next_time_step() ... get_next_time_step()

Perfect timing. I have a question that I've been thinking about all day, since I saw it in use in @rezarastak's other PR. Why is this function called `get_next_time_step()` and not `get_current_time_step()`? It is the timestep (size) that is being used at the current iterate in order to move onto the next one, right?

bangerth:
@rezarastak -- great suggestions. All incorporated.

I'll let Reza answer @jppelteret 's question. I suspect that it comes down to the question of when you *actually* increment the time via `advance_time()`. Do you do this at the beginning of the loop over all time steps before you do the work, or at the end of the loop body when all of the work has already been done. Depending on when you do it, it's either the current or the next time step :-)

jppelteret:
> Depending on when you do it, it's either the current or the next time step

Fair enough, this makes sense. But there should then be a sister function that gives you the current timestep size (i.e. that last one that the time was incremented by) so that you don't have to write a term (lets say, in a constitutive law or weak form) like `(A_{t} - A_{t-1})/dt` as `(A(get_current_time()) - A*(get_current_time() - get_next_step_size()))/get_next_step_size()`. That seems a bit strange to me.

jppelteret:
Ok, it seems like my comments here were a bit misplaced. What I've written about has more-or-less been taken care of in #9748. Sorry for the noise.

include/deal.II/base/discrete_time.h
```diff
@@ -79,7 +113,11 @@ class DiscreteTime
   get_end_time() const;
 
   /**
-   * Return the size of the step from current time step to the next.
+   * Return the size of the step from current time step to the
+   * next. As discussed in the introduction to the class, this is the
+   * *actual* time step, and may differ from the *desired* time step
+   * set in the constructor or through the
+   * set_desired_next_time_step() function.
```
rezarastak, line 120:
```suggestion
   * set_next_step_size() function.
```

include/deal.II/base/discrete_time.h
```diff
@@ -21,13 +21,21 @@
...
+ * This class provides a means to keep track of the simulation time in a
+ * time-dependent simulation. It manages stepping forward from a start time
+ * $T_{\text{start}}$ to an end time $T_{\text{end}}$. It also allows adjusting
+ * the time step size during the simulation. It is guaranteed that at all times
+ * the current simulation time is in the closed interval between the start time
+ * and the end time.
  *
- * You can loop over all of the time steps by using a for loop
+ * The model this class follows is that one sets a *desired* time step length
+ * either through the constructor or using set_desired_next_step_size()
```
rezarastak, line 32:
```suggestion
 * either through the constructor or using set_next_step_size()
```

include/deal.II/base/discrete_time.h
```diff
@@ -36,13 +44,39 @@ DEAL_II_NAMESPACE_OPEN
...
  * @endcode
+ * In the above example the time starts at $T_{\text{start}} = 0$ until
+ * $T_{\text{end}}=1$. Assuming the time step $dt = 0.3$ is not modified inside
+ * the loop, the time is advanced from $t = 0$ to $t = 0.3$, $t = 0.6$, $t =
+ * 0.9$ and finally it reaches the end time at $t = 1.0$. Here, the final step
+ * size needs to be reduced from its desired value of 0.3 to $dt = 0.1$ in order
+ * to ensure that we finish the simulation exactly at the specified end time. In
+ * fact, you should assume that not only the last time step length may be
+ * adjusted, but also previously ones -- for example, this class may take the
+ * liberty to spread the decrease in time step size out over several time steps
```
rezarastak, line 55: We don't currently do this, but I understand you are writing this to reserver the freedom to change in internal implementation and to weaken the post-conditions of some of the methods. It is fine with me.

bangerth:
@rezarastak -- I was going to rename the member function itself in a later patch and so had the new name in the documentation already. Let's leave it as is here. I'll take care of renaming the actual function as a follow-up once #9748 is in. Would that be ok with you?

@dealii/dealii -- anyone else want to review?

rezarastak:
@bangerth Yes it is okay with me to change the name of that function in a separate PR and use the new name here (even if we merge this first). Still there is a slight inconsistency since in the documentation you call the function with both `set_desired_next_step_size()` and `set_desired_next_time_step()`. I think the latter is incorrect.

jppelteret:
@bangerth Feel free to merge this yourself once the mandatory CI is complete.

bangerth:
@rezarastak -- oh, I see now what you mean.

I fixed the naming of the functions.

rezarastak:
> Why is this function called get_next_time_step() and not get_current_time_step()? It is the timestep (size) that is being used at the current iterate in order to move onto the next one, right?

@jppelteret Sorry for replying late to your question. The real and philosophical meaning of `current` as opposed to `next` and/or `previous` in this concept of discrete time incrementation may have different interpretations for each researcher. I have the answer to your question but I want to think through it more and write a proper explanation. As you suggested we can discuss it further in #9748

jppelteret:
> Sorry for replying late to your question.

Thanks, and no worries at all @rezarastak. I should have gone through the other PR before posting the question here, let's all work through this together over there.

PR #9854, opened 2020-04-07
Update the cross compilation documentation

dangars:
This is a follow-up on #9843. I describe how to cross compile deal.II using clang. I describe as well the common pitfalls.


jppelteret:
Cool, thanks for adding this. I'm not the right person to review this (so I won't comment on the content of the PR), but I was curious to know if you've ever looked at using [spack to cross-compile](https://spack.io/about/#custom-versions--configurations)? According to the documentation, it should be somewhat straight-forward to cross-compile the entire dependency chain, but as I've never done it myself I've got no idea if its more challenging than the documentation lets on.

dangars:
@jppelteret I didn't know Spack, it is very cool! I see that it is very active and has many contributors. Is this what supercomputers administrators use to build their packages?
I looks like it might be easier to cross compile using Spack, I'll add a link to Spack as an alternative approach.

jppelteret:
> Is this what supercomputers administrators use to build their packages?

Yes, exactly. Its managed by the LLNL, and we also contribute to it and try to help maintain the deal.II ecosystem of packages. Here's how they describe its utility [in their documentation](https://spack.readthedocs.io/en/latest/):

> Spack is a package management tool designed to support multiple versions and configurations of software on a wide variety of platforms and environments. It was designed for large supercomputing centers, where many users and application teams share common installations of software on clusters with exotic architectures, using libraries that do not have a standard ABI. Spack is non-destructive: installing a new version does not break existing installations, so many configurations can coexist on the same system.

doc/developers/porting.html
```diff
@@ -118,40 +118,101 @@ <h2>Porting to a new platform</h2>
...
-      You have to set up a native deal.II build directory first and run
-      <code>make expand_instantiations_exe</code> in it. The executable is
-      needed for the build system (and obviously the cross compiled version
-      cannot be used). Locate the <code>expand_instantiations</code>
-      executable (it usually resides under
-      <code>${CMAKE_BINARY_DIR}/bin</code> and export its location with the
-      <code>PATH</code> environment variable.
+      You can use any compiler for cross compilation, although we find that
+      <a href="https://clang.llvm.org/">LLVM/clang</a> is the most versatile
+      because it is natively a cross-compiler. Below you can find an example
```
tamiko, line 137: What about slightly rephrasing to "[...], although we find that LLVM/clang is more versatile because it supports multiple architecture targets in a single executable natively"?
I would like to avoid the "natively a cross-compiler" expression. It might also be nice to link to
https://clang.llvm.org/docs/CrossCompilation.html as well.

doc/developers/porting.html
```diff
@@ -118,40 +118,101 @@ <h2>Porting to a new platform</h2>
...
+
+# You may have to include some c++ headers of the target
+include_directories(SYSTEM ${target_root}/path/to/c++/include/files)
+</pre>
+      A great advantage of LLVM/clang is that you can use the toolchain of the
+      target with the option <code>CMAKE_CXX_COMPILER_EXTERNAL_TOOLCHAIN</code>;
+      this means that you do not have to install a specific clang compiler for
+      cross compiling. On the other hand if you use gcc and the version of the
+      host glibc is higher than the version of the target glibc you may run into
+      undefined reference errors. You should place all the relevant libraries of
```
tamiko, line 168: This statement is not correct: The forward compatibility issue with glibc is something you will face with clang as well - the C library is independent of the compiler. Generally, as long as you are not using forward compatible libraries (such as for example musl as C library, or LLVM's libc++ as C++ standard template library) you have to make sure that the glibc and libstdc++ versions used for cross compilation are the same or older than the target.

doc/developers/porting.html
dangars: I see your point. We shouldn't choose any compiler. I took your phrase and replace "we find" by "might". I also rephrased other parts of the PR to keep it more neutral.

tamiko:
This reads very well now, @dangars ready to merge?

dangars:
Yes!

PR #9856, opened 2020-04-08
Added windows build to github-ci

luca-heltai:
With this PR, we could remove appveyor from our CI checks, and keep all builds on the same spot.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,3 +95,42 @@ jobs:
         ./contrib/utilities/check_indentation.sh
         make -j 2
         make -j 2 test #quicktests
+
+  windows-serial:
+    # Serial build on Windows
+    name: Windows Serial
+    runs-on: [windows-latest]
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: info
+      run: |
+        cmake --version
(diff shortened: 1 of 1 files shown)


masterleinad:
`AppVeyor` has proven to be pretty reliable and doesn't seem to suffer from the occasional problems pulling the commits. What are the advantages of building with `Actions` instead?

luca-heltai:
I am agnostic on the two platforms. However, the time limit in Appveyor would not allow to build both debug and release versions. Here the time limit is much less stringent, and would allow (in principle) to build a complete "Windows package", similar to what can be built on mac machines (and actually, we can do that also on the mac builds right now).
We could have all binary packages built in one place, and this is a big plus, in my opinion.

With one action file, we can provide release binaries, and nightly (weekly?) builds on all supported platforms for both Debug and Release. Moreover, it is trivial to have self-hosted runners with github actions (in case we wanted to use some of our servers to do the builds, for example). Appveyor is a windows only solution, while actions are working quite reliably for linux, mac, and windows using the same syntax. This is a big advantage, in my opinion...

masterleinad:
Let's see how this works and decide later for one or the other.

PR #9862, opened 2020-04-09
refactoring of ParameterAcceptor::initialize()

nfehn:
This PR builds on PR #9741 and shows how ``ParameterAcceptor`` can be simplified by implementing base functionality in ``ParameterHandler`` so that it can be used by application classes. According to single responsibility, switching between different code paths according to the type of input file should be done in the base class ``ParameterHandler`` that cares about this. One will otherwise end up in several implementations, one supporting .prm, another one .prm and .xml, another one .json, etc. with a lot of duplicated code. Making chances to the code will become increasingly complex otherwise since several classes have to be kept up-to-date.

### DIFF
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1081,6 +1081,15 @@ public:
   parse_input_from_json(std::istream &input, const bool skip_undefined = false);
 
   /**
+   * Create a default input file by calling print_parameters() depending on the
+   * type of input file specified (prm, xml, json).
+   */
+  void
+  create_default_input_file(const std::string filename,
+                            const OutputStyle style) const;
+
+
+  /**
    * Clear all contents.
    */
   void
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1085,14 +1085,8 @@ public:
    * type of input file specified (prm, xml, json).
    */
   void
-<<<<<<< 20b4164047452179eab3f55314f2a886428e31a0
   create_default_input_file(const std::string filename,
                             const OutputStyle style) const;
-=======
-  create_default_input_file(const std::string &filename,
-                            const bool         sort_alphabetical = true,
-                            const bool         short_version     = true) const;
->>>>>>> use a reference instead of copy
 
   /**
    * Clear all contents.
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1086,7 +1086,8 @@ public:
    */
   void
   create_default_input_file(const std::string &filename,
-                            const OutputStyle  style) const;
+                            const bool         short_version  = true,
+                            const bool keep_declaration_order = true) const;
 
   /**
    * Clear all contents.
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1088,13 +1088,11 @@ public:
   create_default_input_file(const std::string filename,
                             const OutputStyle style) const;
 
-
   /**
    * Clear all contents.
    */
   void
   clear();
-
 
   /**
    * Declare a new entry with name <tt>entry</tt>, default and for which any
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1085,8 +1085,14 @@ public:
    * type of input file specified (prm, xml, json).
    */
   void
+<<<<<<< 20b4164047452179eab3f55314f2a886428e31a0
   create_default_input_file(const std::string filename,
                             const OutputStyle style) const;
+=======
+  create_default_input_file(const std::string &filename,
+                            const bool         sort_alphabetical = true,
+                            const bool         short_version     = true) const;
+>>>>>>> use a reference instead of copy
 
   /**
    * Clear all contents.
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -1085,8 +1085,8 @@ public:
    * type of input file specified (prm, xml, json).
    */
   void
-  create_default_input_file(const std::string filename,
-                            const OutputStyle style) const;
+  create_default_input_file(const std::string &filename,
+                            const OutputStyle  style) const;
 
   /**
    * Clear all contents.
--- a/source/base/parameter_handler.cc
+++ b/source/base/parameter_handler.cc
@@ -802,7 +802,7 @@ ParameterHandler::create_default_input_file(const std::string filename,
                   ExcMessage("OutputStyle does not fit to filename ending."));
       print_parameters(out, style);
     }
- else
+  else
     {
       AssertThrow(
         false,
--- a/source/base/parameter_handler.cc
+++ b/source/base/parameter_handler.cc
@@ -775,6 +775,47 @@ ParameterHandler::parse_input_from_json(std::istream &in,
 
 
 void
+ParameterHandler::create_default_input_file(const std::string filename,
+                                            const OutputStyle style) const
+{
+  std::ofstream out(filename);
+  Assert(out, ExcIO());
+
+  std::string file_ending = filename.substr(filename.find_last_of(".") + 1);
+  boost::algorithm::to_lower(file_ending);
+
+  if (file_ending == "prm")
+    {
+      AssertThrow(style & PRM,
+                  ExcMessage("OutputStyle does not fit to filename ending."));
+      print_parameters(out, style);
+    }
+  else if (file_ending == "xml")
+    {
+      AssertThrow(style & XML,
+                  ExcMessage("OutputStyle does not fit to filename ending."));
+      print_parameters(out, style);
+    }
+  else if (file_ending == "json")
+    {
+      AssertThrow(style & JSON,
+                  ExcMessage("OutputStyle does not fit to filename ending."));
+      print_parameters(out, style);
+    }
+  else
+    {
+      AssertThrow(
+        false,
+        ExcMessage(
+          "Unknown input file. Supported types are .prm, .xml, and .json."));
+    }
+
+  out.close();
+}
+
+
+
+void
 ParameterHandler::clear()
 {
   entries = std_cxx14::make_unique<boost::property_tree::ptree>();
--- a/source/base/parameter_handler.cc
+++ b/source/base/parameter_handler.cc
@@ -775,32 +775,33 @@ ParameterHandler::parse_input_from_json(std::istream &in,
 
 
 void
-ParameterHandler::create_default_input_file(const std::string &filename,
-                                            const OutputStyle  style) const
+ParameterHandler::create_default_input_file(
+  const std::string &filename,
+  const bool         short_version,
+  const bool         keep_declaration_order) const
 {
   std::ofstream out(filename);
   Assert(out, ExcIO());
 
+  const OutputStyle style_short =
+    (short_version ? Short : static_cast<OutputStyle>(0));
+  const OutputStyle style_order =
+    (keep_declaration_order ? KeepDeclarationOrder :
+                              static_cast<OutputStyle>(0));
+
   std::string file_ending = filename.substr(filename.find_last_of('.') + 1);
   boost::algorithm::to_lower(file_ending);
-
   if (file_ending == "prm")
     {
-      AssertThrow(style & PRM,
-                  ExcMessage("OutputStyle does not fit to filename ending."));
-      print_parameters(out, style);
+      print_parameters(out, PRM | style_short | style_order);
     }
   else if (file_ending == "xml")
     {
-      AssertThrow(style & XML,
-                  ExcMessage("OutputStyle does not fit to filename ending."));
-      print_parameters(out, style);
+      print_parameters(out, XML | style_short | style_order);
     }
   else if (file_ending == "json")
     {
-      AssertThrow(style & JSON,
-                  ExcMessage("OutputStyle does not fit to filename ending."));
-      print_parameters(out, style);
+      print_parameters(out, JSON | style_short | style_order);
     }
   else
     {
--- a/source/base/parameter_handler.cc
+++ b/source/base/parameter_handler.cc
@@ -775,13 +775,13 @@ ParameterHandler::parse_input_from_json(std::istream &in,
 
 
 void
-ParameterHandler::create_default_input_file(const std::string filename,
-                                            const OutputStyle style) const
+ParameterHandler::create_default_input_file(const std::string &filename,
+                                            const OutputStyle  style) const
 {
   std::ofstream out(filename);
   Assert(out, ExcIO());
 
-  std::string file_ending = filename.substr(filename.find_last_of(".") + 1);
+  std::string file_ending = filename.substr(filename.find_last_of('.') + 1);
   boost::algorithm::to_lower(file_ending);
 
   if (file_ending == "prm")
--- a/source/base/parameter_handler.cc
+++ b/source/base/parameter_handler.cc
@@ -802,7 +802,7 @@ ParameterHandler::create_default_input_file(const std::string filename,
                   ExcMessage("OutputStyle does not fit to filename ending."));
       print_parameters(out, style);
     }
-  else
+ else
     {
       AssertThrow(
         false,
--- a/include/deal.II/base/parameter_acceptor.h
+++ b/include/deal.II/base/parameter_acceptor.h
@@ -373,22 +373,18 @@ public:
    * default values, and don't want to read external files to use a class
    * derived from ParameterAcceptor.
    *
-   * If outfilename is not the empty string, then write the content that was
-   * read in to the outfilename. The format of both input and output files are
-   * selected using the extensions of the files themselves. This can be either
-   * `prm` or `xml` for input, and `prm`, `xml`, or `tex/latex` for output. If
-   * the output format is `prm`, then `output_style_for_prm_format` is used to
-   * decide whether we write the full documentation as well, or only the
-   * parameters.
+   * If `output_filename` is not the empty string, then write the content that
+   * was read in to `output_filename`. The format of both input and output files
+   * are selected using the extensions of the files themselves.
    *
    * If the input file does not exist, a default one with the same name is
-   * created for you, and an exception is thrown.
+   * created, and an exception is thrown.
    *
    * @param filename Input file name
    * @param output_filename Output file name
-   * @param output_style_for_prm_format How to write the output file if format
-   * is `prm`
+   * @param output_style_for_output_filename Defines style of output file
    * @param prm The ParameterHandler to use
+   * @param output_style_for_filename Defines style of default input file (if created)
    */
   static void
   initialize(
--- a/include/deal.II/base/parameter_acceptor.h
+++ b/include/deal.II/base/parameter_acceptor.h
@@ -391,11 +391,14 @@ public:
    * @param prm The ParameterHandler to use
    */
   static void
-  initialize(const std::string &                 filename        = "",
-             const std::string &                 output_filename = "",
-             const ParameterHandler::OutputStyle output_style_for_prm_format =
-               ParameterHandler::ShortText,
-             ParameterHandler &prm = ParameterAcceptor::prm);
+  initialize(
+    const std::string &                 filename        = "",
+    const std::string &                 output_filename = "",
+    const ParameterHandler::OutputStyle output_style_for_output_filename =
+      ParameterHandler::DefaultStyle,
+    ParameterHandler &                  prm = ParameterAcceptor::prm,
+    const ParameterHandler::OutputStyle output_style_for_filename =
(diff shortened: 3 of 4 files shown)


source/base/parameter_handler.cc
```diff
@@ -736,6 +736,32 @@ ParameterHandler::parse_input_from_json(std::istream &in,
 
 
 
+void
+ParameterHandler::create_default_input_file(const std::string filename,
```
peterrum, line 740: How do you intend to handle the [other output styles](https://www.dealii.org/developer/doxygen/deal.II/classParameterHandler.html#a8364dda711b93753c6809eefe2a8e827)? In particular, `ShortText`, `ShortXML`, `ShortJSON`.

nfehn: Good question. Currently I ignored that because ``ParameterAcceptor`` did also not do that and I first wanted to stay close to the original implementation. Of course, one should think about this. My opinion is that in terms of design principles the current design with all the ``Short...`` versions is a violation of the open-closed principle, so I would vote for a boolean ``short_version`` and entities of ``OutputStyle`` should be "unique" identifiers.

nfehn: I included the Short versions now. 

Independently from this PR, the design would be improved in my opinion by passing ``short_version`` also to ``print_parameters()`` and removing ``OutputStyle::ShortText/JSON/XML``. I briefly looked at the implementation and things would really become simpler by this change.

peterrum: We had this discussion a couple of weeks ago in https://github.com/dealii/dealii/pull/9707#pullrequestreview-379073429. I like the suggestion made by @bangerth there because it resembles the approach also used in `FEValues` to specify the quantities to be updated. But I did not find any time to implement it then.

source/base/parameter_handler.cc
```diff
@@ -736,6 +736,42 @@ ParameterHandler::parse_input_from_json(std::istream &in,
...
+                                            const bool        sort_alphabetical,
+                                            const bool short_version) const
+{
+  std::ofstream out(filename);
+  Assert(out, ExcIO());
+
+  std::string file_ending = filename.substr(filename.find_last_of(".") + 1);
```
peterrum, line 747: As noted in #9741, the ending might be all cap.

source/base/parameter_handler.cc
```diff
@@ -736,6 +736,42 @@ ParameterHandler::parse_input_from_json(std::istream &in,
...
+
+  if (file_ending == "prm")
+    if (short_version)
+      print_parameters(out, ParameterHandler::ShortText, sort_alphabetical);
+    else
+      print_parameters(out, ParameterHandler::Text, sort_alphabetical);
```
peterrum, line 753: If we would make the change described above, we could write:
```suggestion
    print_parameters(out, ParameterHandler::Text | short_version, sort_alphabetical);
```

(Of coarse, in this case `short_version` is not a bool anymore.)

source/base/parameter_handler.cc
nfehn: At many places in the code we have if-statements like ``if(output_style == OutputStyle::Text || output_style == OutputStyle::ShortText)``.  For me, the design problem is that ``ShortText`` etc. is an entity of ``OutputStyle``.

peterrum: I am afraid that the enum entries `Text`, `LaTeX`, `Description`, `XML`, `JSON`, and `ShortText` have stay part of `ParameterHandler::OutputStyle` for compatibility reasons (I have not listed `ShortXML` and `ShortJSON` since they have been added only recently since the last release). The only way out what I see (as suggested by @bangerth) would be introduce new entries in the enum and redefine existing entries, like:

```cpp
  enum OutputStyle {
    with_documentation = 0,                    //  0
    without_documentation = 1,                 //  1

   Text = 2,                                   //  2
   ShortText = Text | without_documentation,   //  3

  LaTex = 4,                                   //  4

  Description = 8,                             //  8

  JSON = 16,                                   // 16
  ShortXML = XML | without_documentation,      // 17

  JSON = 32,                                   // 32
  ShortJSON = JSON | without_documentation,    // 33
  }
```

source/base/parameter_handler.cc
nfehn: I think for this PR it is important to agree on the interface of the function, the implementation is something that can be changed later given that `OutputStyle` is work in progress. For the interface, I believe `bool` is the correct type. If another type is needed to communicate with `OutputStyle`, I think the type should be transferred within this function.

source/base/parameter_handler.cc
nfehn: It's ok. I can live with it if the short versions will remain part of OutputStyle. We should just be aware of that and not force new functions and interfaces to adhere to that redundancy in design.

source/base/parameter_handler.cc
luca-heltai: I would also not use a Boolean for `sort_alfabetical`. If we go towards the route of the enum, also the alphabetical sorting is part of the output style, and could be specified by a bit in the enum.

nfehn:
Can we proceed here?

include/deal.II/base/parameter_handler.h
```diff
@@ -1024,13 +1024,21 @@ class ParameterHandler : public Subscriptor
   virtual void
   parse_input_from_json(std::istream &input, const bool skip_undefined = false);
 
+  /**
+   * Create a default input file by calling print_parameters() depending on the
+   * type of input file specified (prm, xml, json).
+   */
+  void
+  create_default_input_file(const std::string &filename,
```
luca-heltai, line 1032: This should take an `OuptutStyle` instead of bool (see #9888, and #9919)

source/base/parameter_handler.cc
```diff
@@ -762,6 +762,43 @@ ParameterHandler::parse_input_from_json(std::istream &in,
 
 
 
+void
+ParameterHandler::create_default_input_file(const std::string &filename,
```
luca-heltai, line 766: Same here: this should take an `OutputStyle`

include/deal.II/base/parameter_handler.h
nfehn: clearly not. `OutputStyle` contains more information than the info of these function parameters. You are now actually blowing up the interface, against your promises in #9888.

luca-heltai: Why do you think that replacing a function that takes three parameters (two of which are boolean), with one that takes only two is "blowing up the interface"?
The user interface is going to be simpler, and will allow for any other options we may implement in the future without changes in the interface, but only in the (internal) implementation.

nfehn: Ok, I am willing to give more detailed explanations, but let's please stop this black-or-white type argumentations and all these rhetorical questions. 

It is not just counting parameters. A good interface reduces dependencies to a minimum (and also avoids redundancy). It's easy to put everything into a larger data structure and "reduce the interface" in your understanding by just plugging in this bigger data structure. When you make changes to the bigger data structure, you will also have to change many of the functions relying on this data structure. You claim to know exactly where changes will be made in the future and where not? That is part of the problem.

Example: Consider the function `Algorithms::Newton::declare_parameters()/parse_parameters()`. 

https://www.dealii.org/developer/doxygen/deal.II/classAlgorithms_1_1Newton.html

According to your definition, the interface is maximally lean and well-designed. I claim the opposite. I think you will agree that your favorite methods for parsing parameters differ from those used in ``Algorithms::Newton`` (namely ``add_parameter()`` vs. ``declare_entry()``). Furthermore, the interfaces of these function have been extended in ``ParameterHandler`` recently. It does not make sense that we have do update ``Algorithms::Newton`` when making changes to ``ParameterHandler``. In fact, we **do not have to** change ``Algorithms::Newton``, because we always try to be backwards compatible. But these are dependencies that lead to the fact that other modules will become outdated and making changes becomes more and more difficult, but these modules are actually completely independent.  Next, you might argue that you do not want to change the Newton solver whenever you add a new parameter for the Newton solver, sure. For this reason, I would introduce a data structure ``NewtonSolverData`` that is going to be extended whenever the functionality of the Newton solver is updated/changed. Imagine that whenever possible classes in deal.ii would use ``ParameterHandler`` as initialization for their data structures? Reducing dependencies is very important.

luca-heltai:
> Ok, I am willing to give more detailed explanations, but let's please stop this black-or-white type argumentations and all these rhetorical questions.

The question was not rhetorical. It came out already in the issue discussion, and a consensus was reached. The question is honest.

> It is not just counting parameters. 

This is not entirely true: the addition of booleans to a function argument list is bad in many ways, and I could name at least 8 of the main developers who think similarly in this respect. If the arguments we are adding are booleans, then any number > 1 (and for some of us, > 0) is bad. It is *also* about counting parameters.

> A good interface reduces dependencies to a minimum (and also avoids redundancy).

There is no added dependency here. Everything is in `ParameterHandler::OutputStyle` (where it belongs), since `ParameterAcceptor::initialize` is meant to initialize an object of type `ParameterHandler`. What dependency are you discussing about? 

How other classes decide to reduce the interface is not under discussion here. 

> It's easy to put everything into a larger data structure and "reduce the interface" in your understanding by just plugging in this bigger data structure. 

Actually, the data structure is **one** integer, which was already present in the `ParameterHandler` class. No addition need to be done, and no change in the interface w.r.t. to 9.1 is needed. Actually I cannot think of a smaller data structure. I saw no PR with your implementation, so I cannot compare, but *any* data structure will be larger than one integer.

> When you make changes to the bigger data structure, you will also have to change many of the functions relying on this data structure. You claim to know exactly where changes will be made in the future and where not? That is part of the problem.

I don't understand this sentence. What bigger data structure are you talking about? And it is not a rhetorical question. Seriously. 

> Example: Consider the function Algorithms::Newton::declare_parameters()/parse_parameters().

What does this have to do with outputting parameter files? Again, this is not a rhetorical questions. I'm dead serious.

> According to your definition, the interface is maximally lean and well-designed. 

Not what we said. 

The interface that several of us proposed is backward compatible with v. 9.1, it undoes bad programming style (the addition of two bool parameters to several functions and methods), and offers the possibility to be extended *without changing* the user interface. These are the reasons we are saying it is a *good* interface. 

No one claimed it is maximally lean and well-designed. It does what it should, responding to the three points above. The alternative does not (at least not in my opinion). This is good enough for me to lean towards one design strategy w.r.t. the other.

> I claim the opposite. I think you will agree that your favorite methods for parsing parameters differ from those used in Algorithms::Newton (namely add_parameter() vs. declare_entry()).

Actually, they are compatible, thanks to the `ParameterAcceptorProxy`, which makes it behave in my favorite way, without  changing a single line of code.

However, I have no idea what the example above has to do with `ParameterHandler::OutputStyle style = Short | AlphabeticalOrder` VS `bool alphabetical = true; bool short = true`.

> Furthermore, the interfaces of these function have been extended in ParameterHandler recently. 
> It does not make sense that we have do update Algorithms::Newton when making changes to ParameterHandler. 

In what way should this affect Algorithms::Newton (of v9.1)? 

> In fact, we do not have to change Algorithms::Newton, because we always try to be backwards compatible. 

In fact, we don't. v9.1 of `Algorithms::Newton` has no reference to OutputStyle (and should not: why would you want to specify in what format a parameter file should be printed in `Algorithms::Newton`?). On the other hand,  ParameterAcceptor::initialize **has** to follow the capabilities of `ParameterHandler`, since it was designed to initialize, manipulate, and make life easier for the handling of the `ParameterHandler` class itself. 

If you add functionality to the style with which `ParameterHandler` can output files, this functionality should be at least enabled in `ParameterAcceptor`, or people will stop using `ParameterAcceptor` because it lacks the necessary functionality. And, by the way, this should be true **also** with your approach, even if you did not implement it here (i.e., you did not add the two booleans to the `ParameterAcceptor::initialize` function, claiming they were not there, and just arbitrarily decided how to call the new function that creates a default file). 

> But these are dependencies that lead to the fact that other modules will become outdated and making changes becomes more and more difficult, but these modules are actually completely independent. 

Not `ParameterAcceptor` (the **only** subject of discussion of this PR) and `ParameterHandler`. They are dependent on one another (believe me, I know: I wrote one of the two), and they should be kept up to date. `ParameterAcceptor` should be able to initialize parameter files just as you should be able to do through `ParameterHandler::print` itself.

The reason why this is not there, is because the recent changes to `ParameterHandler::print_*` were not enabled by the contributors of those facilities in `ParameterAcceptor`. This does not mean that they were right, nor that it was ok not to insert those capabilities inside `ParameterAcceptor::initialize`. My guess is simply that the authors did not realize that those changes should have gone also in `ParameterAcceptor::initialize`. As soon as someone would have found the need for those, they would have been added (similarly to what you are doing).

> Next, you might argue that you do not want to change the Newton solver whenever you add a new parameter for the Newton solver, sure. For this reason, I would introduce a data structure NewtonSolverData that is going to be extended whenever the functionality of the Newton solver is updated/changed. Imagine that whenever possible classes in deal.ii would use ParameterHandler as initialization for their data structures? Reducing dependencies is very important.

I totally agree. What we should be doing here goes **exactly** in the direction of reducing dependencies, since we **remove** two parameters, in favor of some more values in an already existing enum. If this is not reducing dependencies, I don't know what is.

Ps: This is the last comment I'll write on this topic and on the issue.

With the many hours I spent in one issue and two PR, I could have reviewed at least 4 other PRs, and written half the documentation of step-70. I'm not willing to waste any more time with discussions. 

If you think your solution is better, implement it, push it to a PR, and let the rest of the community decide whether it should or it should not go into the library.

I honestly think that this PR goes in the correct direction. I simply suggested that we wait till #9919  is merged, so that we can remove all booleans from this PR as well, and we can go back to a version of `ParameterHandler` and `ParameterAcceptor` that are backward compatible with 9.1, and that offer all the desired functionality.

Notice also that `ParametereAcceptor` is still broken (see #9889), and this PR would fix it. I'm really looking forward to merging this. If you really want to get this merged as is, we can go ahead, and I'll open another PR later to undo what seems so difficult to fix here, and let other decide about its destiny.

nfehn:
I don't know why you are again so angry and why you want to construct contradictions from my explanations where possible and rephrase them in a way that they appear ridiculous. Anyway, we agreed on a design in issue #9888 and all I was asking today is whether we could complete the present PR first since (i) this PR has been the first one of the currently running PRs and issues on ParameterHandler/Acceptor, and since (ii) it did not change interfaces, but relied on the current version of the deal.II library. (iii) I also think this PR did not break ParameterAcceptor?  I have difficulties to accept your language that this is "bad design" and that one has to "undo" this, I accept of course that you are of a different opinion in certain questions. First and foremost, the present PR is an advancement in my opinion. Of course, feel free to further improve for the future. You are using a language of "we (8 main developers)" against "you". Please don't try to divide the group or exclude newcomers. Let's please try to find a good end to these discussions in a way that we can stay colleagues, ok?

bangerth:
The weather outside is nice and my plan is to spend the next couple of hours working in my garden. So I'm not interested in lengthy philosophical discussions, but it seems like an outside voice is required here.

@nfehn: @luca-heltai has been working on and with deal.II for nearly 20 years. Between the principal developers of deal.II, there are likely more than 100 years of working on large and complex software systems. I suspect that they will all tell you that there are no ideal solutions: in some cases, `bool` arguments may be right, in others they are not. In some cases, `enum`s are the right approach. Statements of the form "minimal interface" or "A good interface reduces dependencies to a minimum (and also avoids redundancy)" or "It's easy to put everything into a larger data structure and "reduce the interface"" are absolute statements, but absolutes don't work well in practice -- every experienced programmer will likely say that good software design is more an art than a science and that "experience" is a very important component in seeing what is likely going to work and what doesn't. (That's also why some of the early parts of deal.II are poorly designed -- simply because at the time we didn't have the same experience we have today.)

What I would like to ask is that if @luca-heltai suggests that something should be done differently, then take that at face value. He's a mild-mannered guy, and he really knows his stuff. If he asks for a different approach, don't argue but trust his experience and learn from it. We are, by and large, a friendly community in which people get along because they appreciate the technical expertise people bring to the table, as well as the opportunity to learn from each other. If @luca-heltai gets annoyed by a discussion, it's likely that it's you not him who's doing something wrong. (I'll readily admit that he's gotten annoyed with me as well in the past and, in hindsight, generally for good reasons.)

We do want your expertise and skills here. You're clearly a good programmer, but from the comments here and elsewhere, you might have the Italian temperament that @luca-heltai doesn't actually have ( ;-) ). Take a step back and think about what's important to you: To get things done the way *you* are convinced is right, or to have the opportunity to learn from people who have many decades of experience writing large software systems and who are willing to explain (and teach!) why they think that a different way is more promising?

luca-heltai:
@nfehn, would you mind rebasing on master, and do the following changes? 

- [ ] change signature of `create_default_input_file` to `create_default_input_file(const std::string &filename, ParameterHandler::OutputStyle style = ParameterHandler::DefaultStyle)`.
- [ ] add to the signature of `ParameterAcceptor::initialize` an additional `OuptuStyle` for the default parameter file
~~~C++
ParameterAcceptor::initialize(
  const std::string &                 filename,
  const std::string &                 output_filename,
  const ParameterHandler::OutputStyle output_style_for_output_filename = ParameterHandler::DefaultStyle,
  ParameterHandler &                  prm = ParameterAcceptor::prm,
  const ParameterHandler::OutputStyle output_style_for_filename = ParameterHandler::DefaultStyle
);
~~~
This allows us to keep backward compatibility, and adds the possibility to also specify a different format for the `filename`, which was not possible before. 

The ordering of the arguments is not ideal, but keeps backward compatibility, while at the same time adding the possibiilty to specify a different format for the default parameter file.

nfehn:
I made the changes here https://github.com/dealii/dealii/pull/9862/commits/fc1b1edd5af404ab201dd5671a569ac61b9617eb

>  change signature of create_default_input_file to create_default_input_file(const std::string &filename, ParameterHandler::OutputStyle style = ParameterHandler::DefaultStyle).

Three additional asserts are necessary since there is a redundancy in the interface (with OutputStyle containing more information than needed by this function), see discussion above. For this reason, I think a default parameter does not really make sense for this function.

source/base/parameter_handler.cc
nfehn: When using OutputStyle as parameter, it is difficult to use this function. The aim of this function is to be able to create a default input file in case none is existing so far (meaning that you can also not configure OutputStyle from parsed parameters). All one has is the filename, and the aim of this function is that the filename ending decides in which format to write the parameter file. However, ParameterHandler has to check the correctness of OutputStyle, which means that the user has to configure OutputStyle correctly before calling this function. Now, to avoid if-else-branches in the application code (which is the purpose of this function), I believe it is necessary and practical to change back to a minimal interface.

nfehn: In my opinion, the second half of ``ParameterAcceptor::initialize()`` (which I did not touch so far) suffers also from this design problem. The "prm" branch currently has an Assert

https://www.dealii.org/developer/doxygen/deal.II/parameter__acceptor_8cc_source.html

 but similar Asserts would be necessary for the other branches. I add these Asserts for this PR but my recommendation is to avoid the redundancy (output format is contained both in ending of filename and in ``OutputStyle``) and change this in a follow-up PR. For example, the "json" version is also not supported by the second half of  ``ParameterAcceptor::initialize()``.

luca-heltai:
In #9950 I've sketched the way I'd have implemented also the `write_default_input_file`. In fact, it is unclear why we should have a `write_default_input_file` instead of simply `print_parameters` with a string, as we did for the corresponding `read_input` functions.

With that implementation done, the `initialize` of this PR should simply check if the strings are empty, and if they are not, it should delegates to the print_function, passing the two styles (in one case, only if the file does not exist, in the second case always).  The documentation clearly state how the `style` parameter is treated in this case.

Of course that's just a suggestion, but I think it allows for a simple -- and uniform -- treatment of the `OutputStyle` enum that we just introduced.

nfehn:
> In #9950 I've sketched the way I'd have implemented also the write_default_input_file. In fact, it is unclear why we should have a write_default_input_file instead of simply print_parameters with a string, as we did for the corresponding read_input functions.

I can also live with this version, since my primary intention of the present PR was actually such a function, no matter how it is called in the end, so print_parameters() is of course also fine. Regarding the design, we have discussed the pros and cons a lot, and I do not want to reiterate again for the sake of a good weather forecast. Since #9950 renders the present PR superfluous (or at least my effort put into it), I would then suggest that I do not waste further time in this PR and you do the refactoring of ``ParameterAcceptor.:initialize()`` yourself, given that you are the ParameterAcceptor expert and that you will fix or adapt the failing tests much quicker. Feel free to close the present PR.

luca-heltai:
> I can also live with this version, since my primary intention of the present PR was actually such a function, no matter how it is called in the end, so print_parameters() is of course also fine

I still think that the function you wrote should exist. However, from the way it was implemented, it was not printing the default values to the input file, but whatever values were stored in the `ParameterHandler` (independently on the bool VS enum questions), so it was actually a variant of the `print_parameters`, and not a `create_default_input_file`.

luca-heltai:
superseeded by  #9950

PR #9863, opened 2020-04-09
Use #ifdef to guard std::async in step-69

dangars:
@tamiko I think that you need `DEAL_II_WITH_THREADS`  for step-69. In my computer I explicitly disable `DEAL_II_WITH_THREADS` and step-69 does not compile because it needs threads.

### DIFF
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -2509,7 +2509,11 @@ namespace Step69
                   output_granularity,
                   "time interval for output");
 
+#ifdef DEAL_II_WITH_THREADS
     asynchronous_writeback = true;
+#else
+    asynchronous_writeback = false;
+#endif
     add_parameter("asynchronous writeback",
                   asynchronous_writeback,
                   "Write out solution in a background thread performing IO");
@@ -2865,7 +2869,15 @@ namespace Step69
     // run in the background.
     if (!asynchronous_writeback)
       {
+#ifdef DEAL_II_WITH_THREADS
         background_thread_state = std::async(std::launch::async, output_worker);
+#else
+        AssertThrow(
+          false,
+          ExcMessage(
+            "\"asynchronous_writeback\" was set to true but deal.II was built "
+            "without thread support (\"DEAL_II_WITH_THREADS=false\")."));
+#endif
       }
     else
       {
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -2867,7 +2867,7 @@ namespace Step69
     // At this point we can return from the <code>output()</code> function
     // and resume with the time stepping in the main loop - the thread will
     // run in the background.
-    if (!asynchronous_writeback)
+    if (asynchronous_writeback)
       {
 #ifdef DEAL_II_WITH_THREADS
         background_thread_state = std::async(std::launch::async, output_worker);


drwells:
I am curious: what part of the program fails to build?

dangars:
Just step-69. I'm going to rerun the build in order to post the error

dangars:
This is the error I get. In my build `DEAL_II_WITH_THREADS=OFF`
```CMakeFiles/step-69.release.dir/step-69/step-69.cc.o:step-69.cc:function std::__future_base::_Async_state_impl<std::_Bind_simple<Step69::MainLoop<2>::output(std::array<dealii::LinearAlgebra::distributed::Vector<dou
ble, dealii::MemorySpace::Host>, 4ul> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, double, unsigned int, bool)::{lambda()#1} ()>, void>::_Async_state_impl(Step69:
:MainLoop<2>::output(std::array<dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host>, 4ul> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, d
ouble, unsigned int, bool)::{lambda()#1} (&&)()): error: undefined reference to 'pthread_create'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[ 87%] Built target step-9.release
examples/CMakeFiles/step-69.release.dir/build.make:189: recipe for target 'bin/step-69.release' failed
make[2]: *** [bin/step-69.release] Error 1
CMakeFiles/Makefile2:6543: recipe for target 'examples/CMakeFiles/step-69.release.dir/all' failed
make[1]: *** [examples/CMakeFiles/step-69.release.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

drwells:
Oh, I see - we can't use `std::thread` without `-lpthread`.

drwells:
I think I would prefer to replace this patch with some `#ifdef`s that guard the call to `std::async`: since we optionally use threads in exactly one place we can pretty easily lift that restriction.

```cpp
    if (!asynchronous_writeback)
      {
        background_thread_state = std::async(std::launch::async, output_worker);
      }
    else
      {
        output_worker();
      }
```
Am I missing something obvious or is the `!` in the wrong place? Why would we create  a `std::future` if `asynchronous_writeback = false`?

dangars:
> I think I would prefer to replace this patch with some `#ifdef`s that guard the call to `std::async`: since we optionally use threads in exactly one place we can pretty easily lift that restriction.
> 
> ```c++
>     if (!asynchronous_writeback)
>       {
>         background_thread_state = std::async(std::launch::async, output_worker);
>       }
>     else
>       {
>         output_worker();
>       }
> ```
> 
> Am I missing something obvious or is the `!` in the wrong place? Why would we create a `std::future` if `asynchronous_writeback = false`?

Thanks @drwells, This works! I'm going to modify this PR and resubmit it

dangars:
I updated the PR and modified the title

drwells:
Thanks. We do need that guard but we also need to fail early on if the user requests asyncio and we do not have thread support. Could you add a second check for that?

tamiko:
@drwells The negation is in the wrong place, @dangars Would you mind to fix that as well, please?

dangars:
Done!

examples/step-69/step-69.cc
```diff
@@ -2863,9 +2867,11 @@ namespace Step69
     // At this point we can return from the <code>output()</code> function
     // and resume with the time stepping in the main loop - the thread will
     // run in the background.
-    if (!asynchronous_writeback)
+    if (asynchronous_writeback)
       {
+#ifdef DEAL_II_WITH_THREADS
         background_thread_state = std::async(std::launch::async, output_worker);
+#endif
```
tamiko, line 2874: would you mind adding something like
```
...
#else
  AssertThrow(false, ExcMessage("\"asynchronous_writeback\" was set to true but deal.II was built without thread support (\"DEAL_II_WITH_TREADS=false\")."));
#endif
```

tamiko: (and improve the error message a bit :smile: )

dangars: I added the assert, I didn't change your message, I think that it is good ;)

dangars: The line of the assert can not be break to 80 columns, I think that is because it is a macro

examples/step-69/step-69.cc
```diff
@@ -2863,9 +2867,16 @@ namespace Step69
...
-    if (!asynchronous_writeback)
+    if (asynchronous_writeback)
       {
+#ifdef DEAL_II_WITH_THREADS
         background_thread_state = std::async(std::launch::async, output_worker);
+#else
+        AssertThrow(
+          false,
+          ExcMessage(
+            "\"asynchronous_writeback\" was set to true but deal.II was built without thread support (\"DEAL_II_WITH_THREADS=false\")."));
```
kronbichler, line 2878: Doesn't something like this work?
```suggestion
            "\"asynchronous_writeback\" was set to true but deal.II was built "
            "without thread support (\"DEAL_II_WITH_THREADS=false\")."));
```
We do it in many other places.

dangars: This works!

drwells:
Are you happy with this @tamiko? If so, lets merge.

PR #9865, opened 2020-04-09
Improve CMake documentation for google search

dangars:
I often use google to access the deal.II documentation, but the results are quite confusing for the CMake documenation.

If I search "CMake deal.II", I obtain:
1. [deal.II CMake documentation](https://www.dealii.org/9.1.1/users/cmake.html)
2. [CMake documentation - deal.II](https://www.dealii.org/9.1.1/users/cmakelists.html)
3. [Build system internals - deal.II](https://www.dealii.org/9.1.1/developers/cmake-internals.html)

The third result (Build system internals) is useful, the other two are not quite useful. They have the same name but point to different pages.

Google can change the titles (see [google titles](https://support.google.com/webmasters/answer/35624?hl=en)) if they are not short.

I would change the title of the "User CMake page" to something shorter (any ideas):
- User CMake configuration - deal.II
- CMake configuration for your deal.II projects

In addition I would rename `cmakelists.html` to something more meaningful such as `cmake-user.html`. I think that would also help google. Let me know if I should rename `cmakelists.html`


tamiko:
@dangars Sounds good. Please also add the other renames to this PR.

dangars:
I renamed `cmake.html` to `cmake_dealii.html` and `cmakelists.html` to `cmake_user.html`.

I updated as well many links.

doc/users/config.sample
```diff
@@ -1,7 +1,7 @@
 ##                                                                       ##
 #                      Example configuration file                         #
 #                                                                         #
-#        See doc/readme.html and doc/users/cmake.html for further         #
+#        See doc/readme.html and doc/users/cmake_dealii.html for further         #
```
jppelteret, line 4: @tamiko must have been in a rush when he reviewed this. His eyes would normally begin to twitch when there is some misalignment of characters floating about the screen. Would you mind please centring this on the line and make the hash symbols align?

dangars: good catch :) I modified it!

tamiko: Almost centered! :stuck_out_tongue_closed_eyes:

tamiko: We'll merge it with the block not fully centered so that everyone with OCD reading the file will scream in horrow.

dangars: Ups! I just centered in #9879!

PR #9868, opened 2020-04-10
Add documentation on using CCache

jppelteret:


drwells:
What is the advantage of this over providing CXX_COMPILER_LAUNCHER as a command line argument?

jppelteret:
I must admit that I didn't know about `CXX_COMPILER_LAUNCHER`. But according to [this comment](https://stackoverflow.com/a/37828605) it's a feature of `CMake >= 3.4`, while what I've added here is supported by our CMake `2.8.12` minimum.

dangars:
Very interesting, I didn't know ccache.

I think that it would be better to merge this PR after #9865 because I have renamed `cmake.html` to `cmake_dealii.html`

kronbichler:
I have used ccache in the past (via the compiler wrapper `mpicxx`) when I only had six cores but went away from it on the bigger machines with more cores. In my particular workflow, I did not have enough reuse of the cache, which probably means that I do not switch back to the same branch (modulo documentation) often enough compared to the things I do in between; even after increasing the limit to 10 GB I still had too much divergence - I think somewhere around half of the testsuite exhausts that limit. I think it is a nice addition, though, so :+1:

jppelteret:
Thanks, I'm glad that the feedback is positive so far. I've only got a 7 year old laptop to work with, and the build times are killing me after each branch switch. There's no huge rush to merge this, as I can continue to evaluate it using the flag that @drwells mentioned. I'm hoping that it'll be of some benefit to me at least, so if you want some more concrete analysis of the benefit to building through CCache on a crusty machine, I can report back in a few days.

> I think that it would be better to merge this PR after #9865 because I have renamed cmake.html to cmake_dealii.html

Sure, thanks for the heads up. That sounds like a good plan.

drwells:
My ccache hit rate (albeit for more projects than deal.II) is currently 36.7%. I haven't had any problems with it but it is occasionally quite handy when I switch back and forth between two branches.

tjhei:
> and the build times are killing me after each branch switch.

Unrelated to this PR, but I gave up on switching branches during development. I use ``git worktree`` to create new directories for all my active branches and have separate build directories.

jppelteret:
> I use git worktree to create new directories for all my active branches and have separate build directories.

I would do that if I didn't have only 15Gb of free HDD space... I desperately need to format and reinstall this computer!

bangerth:
I'm grateful for my 12-core laptop (ok, really just 6x2) and the 40x2 core workstations I have :-) I look forward to the two new 96x2 core machines that should be coming any day now -- `make -j192` is going to look really good ;-))

cmake/checks/check_02_compiler_features.cmake
```diff
@@ -501,3 +501,44 @@ IF(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
...
+#
+SET(BUILD_USING_CCACHE FALSE CACHE BOOL "Enable ccache support")
+
+MESSAGE(STATUS "")
+MESSAGE(STATUS "Looking for CCache.")
+IF(BUILD_USING_CCACHE)
+  #
+  # Built-in CCache support requires CMake version 2.8 or newer
+  #
+  IF(CMAKE_VERSION VERSION_LESS 2.8)
```
bangerth, line 521: We already require 2.8.12 in the top level directory, so this is not necessary

cmake/checks/check_02_compiler_features.cmake
```diff
@@ -501,3 +501,44 @@ IF(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
...
+      MESSAGE(STATUS "deal.II will be compiled using CCache.")
+
+      IF ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" # GNU is GNU GCC
+          OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+        # without this compiler messages in `make` backend would be uncolored
+        SET(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fdiagnostics-color=auto")
+      ENDIF()
+
+      SET_PROPERTY(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "ccache")
+      #SET_PROPERTY(GLOBAL PROPERTY RULE_LAUNCH_LINK "ccache")
```
bangerth, line 539: just remove if you don't need it

cmake/checks/check_02_compiler_features.cmake
tamiko: Please remove the if conditional on the cmake version as @bangerth pointed out

cmake/checks/check_02_compiler_features.cmake
```diff
@@ -501,3 +501,44 @@ IF(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
...
+  #
+  IF(CMAKE_VERSION VERSION_LESS 2.8)
+    SET(BUILD_USING_CCACHE FALSE)
+
+    MESSAGE(STATUS "deal.II requires CMake version 2.8, or newer for CCache support.\n"
+                   "Reconfigure with a sufficient cmake version.")
+  ELSE()
+    FIND_PROGRAM(CCACHE_FOUND "ccache")
+
+    IF (CCACHE_FOUND)
```
tamiko, line 529: Also, I am not a huge fan of searching for a ccache executable (that never gets used).

cmake/checks/check_02_compiler_features.cmake
```diff
@@ -501,3 +501,44 @@ IF(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
...
+
+
+#
+# CCache support is only really useful for developers. 
+# We therefore disable the feature by default.
+#
+# See https://invent.kde.org/kde/konsole/-/merge_requests/26/diffs
+# and https://stackoverflow.com/a/24305849
+#
+SET(BUILD_USING_CCACHE FALSE CACHE BOOL "Enable ccache support")
```
tamiko, line 513: All other cached variables are recorded in `setup_cached_variables.cmake`. So it should be refactored in there.

cmake/checks/check_02_compiler_features.cmake
```diff
@@ -501,3 +501,44 @@ IF(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
...
+  ELSE()
+    FIND_PROGRAM(CCACHE_FOUND "ccache")
+
+    IF (CCACHE_FOUND)
+      MESSAGE(STATUS "deal.II will be compiled using CCache.")
+
+      IF ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" # GNU is GNU GCC
+          OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+        # without this compiler messages in `make` backend would be uncolored
+        SET(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fdiagnostics-color=auto")
```
tamiko, line 535: You have to guard the compiler flag with `ENABLE_IF_SUPPORTED`.

jppelteret:
> If you insist that this is the right way forward

I do not :-)

> I think I am not really sold on the configuration changes for that one.

Ok, no worries then. @drwells's suggestion gets me what I want, so I'm happy either way. Let's leave this then. Thanks everyone!

tamiko:
@jppelteret I suggest we keep the documentation changes :smile_cat:

jppelteret:
> (or we add a small ccache paragraph to the documentation)

Ok, I'll do that once #9865, but in another PR. Until then, we can keep this open as a reminder.

drwells:
@jppelteret Now that #9865 is in, would you please rebase?

jppelteret:
@tamiko @drwells I've now added the documentation for using CCache. Would you mind please taking a look?

jppelteret:
BTW. As of right now, the few little hacking sessions that I've been able to do have left me with a cache hit rate of 61.2%. So it's been a worthwhile exercise for me, especially since I think that a lot of the big object files have been loaded from the cache. Much time is still spent linking the library though.

doc/users/cmake_dealii.html
```diff
@@ -894,6 +895,44 @@ <h3> Selecting a compiler </h3>
...
+
+    <p>
+      When using a version of CMake that is greater or equal to 3.4, then using
+      CCache is a little easier. The above may be compressed into a single
+      configure-time argument
+<pre class="cmake">
+  cmake -DCXX_COMPILER_LAUNCHER="ccache" &lt;...&gt;
+</pre>
+      which ensures that CCache is invoked whenever C++ code is compiled.
+    </p>
```
drwells, line 933: I think we should switch the order of these two paragraphs so that users see the official version first. I would also clarify with `CC="ccache gcc"` that that approach is only necessary with older versions of CMake.

doc/users/cmake_dealii.html
```diff
@@ -894,6 +895,44 @@ <h3> Selecting a compiler </h3>
     </p>
 
 
+    <a name="compileccache"></a>
+    <h3> Compiling using CCache </h3>
+    <p>
+      It is possible for the library to be built using the
+      <a href="https://ccache.dev/">CCache</a> ("compiler cache") tool.
```
drwells, line 902: I think it would be better to start by describing what ccache is and then quickly mention that we support using it so that users get the gist of it by reading the first sentence.

jppelteret: Good suggestion, thanks. I'll swap these two paragraphs around.

doc/users/cmake_dealii.html
jppelteret: Good suggestion, thanks. I'll swap these two paragraphs around.

jppelteret:
Thanks for the suggestions @drwells. Although the first and second paragraphs read quite similarly, I think that it flows better now.

doc/users/cmake_dealii.html
```diff
@@ -894,6 +895,52 @@ <h3> Selecting a compiler </h3>
...
+      the same then rebuilding them can be avoided as the results will be taken
+      from directly the cache. Anecdotally, Ccache is not well suited to
+      everybody's workflow, so it may take some experimentation to determine if
+      it's effective and helpful, or offers no particular gain to you.
+    </p>
+
+    <p>
+      Building with Ccache can be enabled directly via the command line. 
+      When using a version of CMake that is greater or equal to 3.4, then using
+      Ccache is a little easier. The above may be compressed into a single
```
drwells, line 921: 'The above' is no longer above :)

doc/users/cmake_dealii.html
```diff
@@ -894,6 +895,52 @@ <h3> Selecting a compiler </h3>
...
+      configure-time argument
+<pre class="cmake">
+  cmake -DCXX_COMPILER_LAUNCHER="ccache" &lt;...&gt;
```
drwells, line 924: Users sometimes get confused when we do things like `./path/to/library` (they enter it literally). Could we just have the flag alone here to make things slightly less ambiguous?

doc/users/cmake_dealii.html
```diff
@@ -894,6 +895,52 @@ <h3> Selecting a compiler </h3>
...
+      which ensures that Ccache is invoked whenever C++ code is compiled.
+    </p>
+
+    <p>
+      For older versions of CMake, Ccache may be invoked by prefixing the
+      compiler commands passed to CMake by environment or configure-time
+      variables. For example, if building using GCC then
+<pre class="cmake">
+  CC="ccache gcc" CXX="ccache g++"
+
```
drwells, line 935: We don't need these two lines, right?

drwells:
@jppelteret *ping* :)

doc/users/cmake_dealii.html
jppelteret: Ok, I've done now this. But note that the way that I wrote it was more or less consistent with the rest of the documentation on this page.

doc/users/cmake_dealii.html
jppelteret: Removed.

jppelteret:
@drwells Sorry for the delay. I've made the changes that you suggested. Hopefully I've got it right this time!

drwells:
@tamiko I believe all of your comments have been addressed. Can we merge?

drwells:
@tamiko This is done and I would like to merge it - would you please approve? :)

drwells:
:tada:

PR #9869, opened 2020-04-10
Fix a doxygen table.

drwells:
Doxygen has its own syntax for tables. I found this while looking at #9855.


kronbichler:
Documentation only, so I'm merging.

PR #9874, opened 2020-04-11
Merge MatrixFree::internal_reinit for DoFHandler and hp::DoFHandler

peterrum:
Related to #9839.

PR #9875, opened 2020-04-11
Remove MatrixFree::DoFHandlers::level

peterrum:
While working on #9874, I noticed that the information is stored two times (also in `MatrixFree::mg_level`).


kronbichler:
Can you please have a look at the error in serial (unused parameter `additional_data`)?

PR #1356, opened 2015-08-13
Also run doxygen on source files.

bangerth:
For each function recognized in a .cc file, this yields a nice
'Definition in line 1234 of file abcd.cc' mark with the documentation
of the function, both line and file being clickable so that people can
follow the link and actually see the source code.

This fixes #1303 and follows a suggestion by Jason Sheldon.

PR #9876, opened 2020-04-11
Don't parse *.cc files with doxygen

masterleinad:
This cuts down the number of `doxygen` warnings to around 250 lines for me. Admittedly, this, of course, implies that no links these files are generated anymore. On the other hand, most of the warnings come from these files implying that that doesn't work properly anyway.

### DIFF
--- a/doc/doxygen/options.dox.in
+++ b/doc/doxygen/options.dox.in
@@ -70,7 +70,7 @@ WARN_IF_DOC_ERROR      = YES
 
 INPUT                  =
 RECURSIVE              = YES
-EXCLUDE_PATTERNS       = *.templates.h
+EXCLUDE_PATTERNS       = *.templates.h @CMAKE_SOURCE_DIR@/source/*.cc
 EXAMPLE_PATH           = @CMAKE_BINARY_DIR@/doc/doxygen/tutorial \
                          @CMAKE_SOURCE_DIR@/examples/doxygen
 EXAMPLE_RECURSIVE      = NO


drwells:
I'm not sure that we should disable this - we explicitly enabled it in #1356.

Are we getting a lot more doxygen warnings now than we did at the last release?

masterleinad:
I am up for discussion here. My idea was that we could find a compromise that allows as to get rid of all the warnings one way or the other such that we could also check for that in pull requests.

bangerth:
I don't particularly like this patch because I think that parsing the .cc files results in a useful service: doxygen provides a link under ever function documentation where that function is actually implemented.

Take https://dealii.org/developer/doxygen/deal.II/classChartManifold.html#a1e9207f06d5bd518383353ebee3f046c as just one example: There is a link to line 987 of `manifold.cc` that one can follow to see the implementation of a function.

We haven't used that as much as we should have in answers on mailing list questions, but I think we should have. We often say "Take a look at the implementation of that function to see how you would have to implement the case you have in mind." The code linked to from the documentation is fully doxygenated -- it's not just a link to a code file and line, but you can click on everything there to follow classes and functions in the same way as in the tutorials. I think that's actually a great service, but it relies on parsing all of the `.cc` files as well.

tjhei:
> There is a link to line 987 of `manifold.cc` that one can follow to see the implementation of a function.

Just one data point: I use this feature sometimes to quickly check what a function does if I don't have an IDE with deal.II open.

masterleinad:
My experience is that it sometimes works and sometimes it doesn't. That's likely my only point here. If we are to fix all the `doxygen` warnings, we will continue to disable parsing more functions in the `*.cc` files so the coverage will (likely) get even scarcer. But maybe this hacky approach is the way to go.

bangerth:
If it works 75% of the time, then that's still useful in 75% of cases :-)

I get that the doxygen warnings are annoying, but I think disabling parsing the `.cc` files is throwing the baby out with the bath water.

masterleinad:
OK. Let's close it then and deal with the warnings individually.

PR #9877, opened 2020-04-12
update included std headers in number.h and types.h

rezarastak:
Changes:
 * Removed `#include <cstddef>` from `types.h` because we don't use `std::size_t` there anymore.
 * Replaced `<cstdlib>` with `<cstddef>` in `numbers.h` because we use `std::size_t` in `numbers.h` but we don't use other things from `<cstdlib>`.
 * Added `<type_traites>` because we use many structs from it in `numbers.h`. The code already compiles without this header (possibly because it is included in another standard header), but I added it for consistency. 

I'm not sure if these changes are necessarily needed, but I just wanted to share what I found while analyzing these include statements.

### DIFF
--- a/include/deal.II/base/numbers.h
+++ b/include/deal.II/base/numbers.h
@@ -27,7 +27,8 @@
 
 #include <cmath>
 #include <complex>
-#include <cstdlib>
+#include <cstddef>
+#include <type_traits>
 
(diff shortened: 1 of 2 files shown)


masterleinad:
@bangerth Seems that my observation regarding the order of `ready to test` and /rebuild is correct.

PR #9885, opened 2020-04-14
Add incompatibility entry regrading ParameterHandler::parse_input()

peterrum:
In reference to https://github.com/dealii/dealii/pull/9741#issuecomment-613351524.

I am not sure if this is the way to issue a warning.

FYI @nfehn @luca-heltai

### DIFF
--- a/doc/news/changes/incompatibilities/20200326NiklasFehn
+++ b/doc/news/changes/incompatibilities/20200326NiklasFehn
@@ -1,0 +1,5 @@
+Changed: The function ParameterHandler::parse_input() can now also parse xml- and json-files,
+depending on the file suffix. From now on, only files ending with .prm are treated as standard 
+parameter files.
+<br>
+(Niklas Fehn, Peter Munch, 2020/03/26)
--- a/doc/news/changes/minor/20200326NiklasFehn
+++ b/doc/news/changes/minor/20200326NiklasFehn
@@ -1,6 +1,5 @@
-Extension: The function ParameterHandler::parse_input() has been extended so that
-the same parse_input function can be called independently of the type of input
-file (prm, xml, json). An additional parameter assert_mandatory_entries_are_found
-is introduced to avoid errors, e.g., due to typos in the input file.
+Extension: An optional additional parameter assert_mandatory_entries_are_found
+is introduced to the function ParameterHandler::parse_input() to avoid errors, 
+e.g., due to typos in the input file.
 <br>
 (Niklas Fehn, 2020/03/26)


source/base/parameter_handler.cc
```diff
@@ -519,17 +519,25 @@ ParameterHandler::parse_input(const std::string &filename,
       std::istream is(&fb);
       std::string file_ending = filename.substr(filename.find_last_of('.') + 1);
       boost::algorithm::to_lower(file_ending);
-      if (file_ending == "prm")
```
nfehn, line 522: I would like to have this if-statement here, because we will later reintroduce it anyway but it now helps understanding the logic of this function. The original intention of this function otherwise gets blurred or we would have to add documentation

source/base/parameter_handler.cc
```diff
@@ -519,17 +519,25 @@ ParameterHandler::parse_input(const std::string &filename,
...
-        AssertThrow(
-          false,
-          ExcMessage(
-            "Unknown input file. Supported types are .prm, .xml, and .json."));
+        {
+          // See: discussion in https://github.com/dealii/dealii/pull/9741
+          // This PR contains the code which we should use from release
+          // 9.3 onwards.
+#ifdef DEBUG
+          if (file_ending != "prm")
```
nfehn, line 533: I think this if can then be removed.

source/base/parameter_handler.cc
```diff
@@ -519,17 +519,25 @@ ParameterHandler::parse_input(const std::string &filename,
...
         parse_input_from_json(is, skip_undefined);
       else
-        AssertThrow(
-          false,
-          ExcMessage(
-            "Unknown input file. Supported types are .prm, .xml, and .json."));
+        {
+          // See: discussion in https://github.com/dealii/dealii/pull/9741
+          // This PR contains the code which we should use from release
+          // 9.3 onwards.
```
nfehn, line 531: I would like to add a comment explaining what we are doing here: "File endings other than prm, xml, json are currently interpreted as files of prm-type for reasons of backward compatibility."

doc/news/changes/incompatibilities/20200326NiklasFehn
```diff
@@ -0,0 +1,7 @@
+Changed: The function ParameterHandler::parse_input() can now also parse xml- and json-files,
+depending on the file suffix. We treat implicitly all files not ending with .xml or
+.json as standard parameter files: however, in the case that the file does not end with .ptr a warning
```
nfehn, line 3: ptr? Here, I would formulate as if the function already contains if-statements for all current types, prm, xml, json, and everything else is a backward compatibility issue

peterrum:
@nfehn Thanks for the review. I have applied the suggestions.

source/base/parameter_handler.cc
```diff
@@ -519,17 +519,36 @@ ParameterHandler::parse_input(const std::string &filename,
...
+        {
+          // See: discussion in https://github.com/dealii/dealii/pull/9741
+          // This PR contains the code which we should use from release
+          // 9.3 onwards. File endings other than prm, xml, json are currently
+          // interpreted as files of prm-type for reasons of backward
+          // compatibility.
+#ifdef DEBUG
+          std::cout
+            << "WARNING: From release 9.3 onwards, this function will only accept files ending with .prm, .xml, or .json as parameter files."
+            << std::endl;
```
kronbichler, line 539: Are we fine with printing this on all MPI ranks? It probably makes little sense to test for MPI ranks (as we do not know the communicator), so I guess a user cannot complain when getting thousands of those...

peterrum: > I guess a user cannot complain when getting thousands of those...

I hope the users don't run the code in debug mode with thousands of processes ;) But I am happy to change it if there are any suggestion!

masterleinad: I would appreciate if there is a way to silence the warning at least.
Also, I would not mention any specific version at all. Let's just say it's deprecated.

peterrum: > I would appreciate if there is a way to silence the warning at least.

In the case of gcc, we could write `#ifdef __DEPRECATED` (https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html). But I am afraid that other compilers have other macros.

> Also, I would not mention any specific version at all. Let's just say it's deprecated.

I will do that!

peterrum: Would `#ifdef DEAL_II_DEPRECATED` work?

masterleinad: No, we use that already as a replacement for `[[deprecated]]`. :slightly_smiling_face: `DEAL_II_ALLOW_DEPRECATED_BEHAVIOR` or something similar would work for me but others might have better suggestions.

luca-heltai:
I am in favour of deprecating any other extension for the `prm` format, but maybe others think differently... in this case I would simply remove the warning and document the bahaviour.

bangerth:
I'm not particularly fond of this patch for the reasons outlined with creating massive amounts of output. I would much rather us just make an incompatible change -- it's not difficult for people to adjust to this change. We also try very hard to not ever actually write anything to the console (in fact, xSDK guidelines explicitly prohibit doing this).

peterrum:
@bangerth OK. Then let's go only with a incompatibility entry (the function has already been documented in #9741).

PR #2615, opened 2016-05-18
Make parallel vector derive from VectorSpaceVector

kronbichler:
This is the first step to move `parallel::distributed::Vector` to `VectorSpaceVector`. Relates to #2535 and #745.

I now moved all code of the parallel vector to a class `LinearAlgebra::parallel::Vector`. @Rombur is this the right scheme to name things? I noticed that the inner `parallel` namespace is a bit confusing and I had to qualify some things, like the member of `ReadWriteVector`. Should we rather make it `LinearAlgebra::distributed::Vector` or `LinearAlgebra::MPI::Vector`?

Regarding the methods, I removed a few of the more obscure `operator==`, `equ` and `sadd` methods that we had in the class, whereas I put some others that are necessary when using this vector throughout the library still in the vector.

Another thing is that the vector still has access functionality as it was before (to keep the class synchronized with the rest of the library) but at some point at least parts of it could go away.

Finally, I did not put the import function in `ReadWriteVector` yet but that is easy to add.

The directly related tests are passing, the full test suite will run later today.

PR #9886, opened 2020-04-14
Remove deprecated LinearAlgebra::distributed::Vector member functions

masterleinad:
Deprecated in https://github.com/dealii/dealii/pull/2615.
This pull request also undeprecates `LinearAlgebra::distributed::Vector::in_local_range()` since all the other vector classes have it, we use it internally and I don't see a common alternative.


kronbichler:
This change seems to have caused a test failure for `petsc/copy_to_dealvec_block` - at first sight I am not entirely sure if we want to remove that test like the `copy_to_dealvec` ones or that rather indicates something we still wanted to support. Could you please have a look @masterleinad ?

masterleinad:
We didn't remove the overloads for the block version so that test should stay. I will take care of it.

Issue #8482, opened 2019-08-07
step-50 TODO list


bangerth:
- [ ] Don't use vector_t, matrix_t
@tcclevenger: Seeing this in your talk just now -- we generally use `CamelCase` for types, so seeing `vector_t` and `matrix_t` in step-50 looks a bit odd. Maybe want to put that on your list of things to change.

tcclevenger:
Yes, I will definitely change this. Thanks.

tjhei:
- [ ] more interesting problem
- [ ] test with PETSc and Trilinos
- [ ] "some" scaling results (I can probably run on a big machine if we are happy with things)

tcclevenger:
Is there any objection to renaming `mg_dof_handler` to `dof_handler`? I think the name is confusing since `mg_dof_handler` doesn't only contain MG information.

tjhei:
> Is there any objection to renaming `mg_dof_handler` to `dof_handler`?

good idea.

bangerth:
The name `mg_dof_handler` is historical -- we used to have a class of that name. There is no reason to still use this name.

dangars:
@tjhei I would be interested in running step-50 and geometric multigrid with PETSc. I would be happy to contribute. Could you let me know what is missing?

I identified the following items:
- You use `exchange_data = true` in the function `reinit()` when you initialize the system_matrix. The interface of PETSc does not have this option. This option is not used in step-40. It is unclear to me if this is important.
- You use Jacobi as smother. In `mg_smoother.h (line 1169)`  you have `smoothers[level].Tvmult(*d, *r);`. The function `Tvmult` is not defined for `PETScWrappers::PreconditionJacobi `. But this is straightforward, since this is a symmetric preconditioner, this function is the same as vmult(). Happy to submit a PR.
- You set the relaxation parameter of the smoother `omega = 0.5`, but `PETScWrappers::PreconditionJacobi ` does not have this option. I checked the PETSc documentation and I couldn't find a way to change the relaxation parameter of Jacobi. Is this strictly necessary?

Am I missing something? Let me know if there is anything else I can do to make geometric multigrid work with PETSc

tjhei:
Hi Daniel,

@tcclevenger and I will open a PR for the updated step-50 version in the next couple of days (based on the version for a publication). Many things have changed compared to the current version on master.

> * You use `exchange_data = true` in the function `reinit()` when you initialize the system_matrix.

I will fix this (the call to reinit() is just slightly different).

[...]

PR #9887, opened 2020-04-14
Add Tvmult to Petsc PreconditionJacobi

dangars:
This adds the operation Tvmult to `PETScWrappers::PreconditionJacobi` as discussed in https://github.com/dealii/dealii/issues/8482#issuecomment-584555393


dangars:
Should I add Tvmult to `PETScWrappers::PreconditionerBase` as a virtual function and then overload the function in `PreconditionJacobi` or I just add the function Tvmult to `PETScWrappers::PreconditionJacobi`?

tjhei:
Can you implement https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/PC/PCApplyTranspose.html#PCApplyTranspose in the base class instead? I would hope that that would work just as well.

tests/petsc/preconditioner_tvmult_01.cc
```diff
@@ -0,0 +1,104 @@
...
+int
+main(int argc, char **argv)
+{
+  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+  MPILogInitAll                    log;
+
+  test<PETScWrappers::PreconditionJacobi>();
+  test<PETScWrappers::PreconditionBlockJacobi>();
+  test<PETScWrappers::PreconditionBoomerAMG>();
+  test<PETScWrappers::PreconditionNone>();
```
dangars, line 103: I added several preconditioners to the tests. Not all the PETSc preconditioners support `PCApplyTranspose`. For example `Eisenstat` and `ParaSails` don't support it. If I try to run `Tvmult` with `Eisenstat` or `ParaSails` I obtain:
```
deal.II encountered an error while calling a PETSc function.
The description of the error provided by PETSc is "No support for this operation for this object type".
The numerical value of the original error code is 56.
```
I think that the message is quite clear. I think that there is no need for additional Asserts.

tjhei: I agree.

tjhei:
perfect, this looks good!

PR #9890, opened 2020-04-14
Fix set particle positions from Function<spacedim>

luca-heltai:
The code path taking a `Function<spacedim>` would not be taken for derived classes, that would call the templated version with `VectorType` equal to `Function<spacedim>`, leading to several compiler erros.

### DIFF
--- a/include/deal.II/particles/particle_handler.h
+++ b/include/deal.II/particles/particle_handler.h
@@ -334,7 +334,8 @@ namespace Particles
      * @authors Luca Heltai, Bruno Blais, 2019.
      */
     template <class VectorType>
-    void
+    typename boost::disable_if<
+      std::is_convertible<VectorType *, Function<spacedim> *>>::type
     set_particle_positions(const VectorType &input_vector,
                            const bool        displace_particles = true);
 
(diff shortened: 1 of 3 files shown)


luca-heltai:
@blaisb, this was a leftover from one of our PRs.

include/deal.II/particles/particle_handler.h
```diff
@@ -334,7 +334,8 @@ namespace Particles
      * @authors Luca Heltai, Bruno Blais, 2019.
      */
     template <class VectorType>
-    void
+    typename boost::disable_if<
```
masterleinad, line 337: Why not just
```suggestion
    typename std::enable_if<!
```
?

luca-heltai: I thought `disable_if` was more self documenting (I often miss the `!`). Changed.

luca-heltai:
Mmmmh. Apparently some compilers don't like `std::enable_if<!expression>` for reasons that I do not understand. I'm reverting to the old version.

masterleinad:
> What an interesting issue. (And I hadn't even known of `std::disable_if`.) It's definitely time for concepts!

It's `boost::disable_if` which is why i preferred `std::enable_if<!`.

bangerth:
Oh, I see. I'll give `std::enable_if` a try later on unless someone wants to beat me to it.

PR #9892, opened 2020-04-15
Step 58: The nonlinear Schroedinger equation

bangerth:
This took long enough to finish, but I think it turned out to be a nice tutorial program. It is, in the grand scheme of things, one of the shorter ones (only marginally longer than step-6). Looking forward to reviews!

/rebuild


masterleinad:
```
Error: repeated header title 'Linearsolvers' in file group 'examples/step-58/doc/intro.dox'
```

bangerth:
Should be fixed now.

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,205 @@
...
+
+<p align="center">
+  <img src="https://www.dealii.org/images/steps/developer/step-58.phase-cyclic.png"
+         alt="Phase of the solution at t=0.242, with a cyclic color map"
+         width="400">
+</p>
+
+Finally, we can generate a movie out of this. (To be precise, the video
+uses two more global refinement cycles and a time step half the size
+of what is used in the program above.) I made the movie with VisIt,
```
peterrum, line 100:
```suggestion
of what is used in the program above.) We made the movie with VisIt,
```

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,205 @@
...
+because that's what I'm more familiar with, and using a hacked color map
+that is also cyclic -- though my color map lacks all of the skill employed by
```
peterrum, line 102:
```suggestion
that is also cyclic -- though our color map lacks all of the skill employed by
```

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,205 @@
...
+The phase picture shown here clearly has some flaws:
+- First, phase is a "cyclic quantity", but the color scale uses a
+  fundamentally different color for values close to $-\pi$ than
+  for values close to $+\pi$. This is a nuisance -- what we need
+  is a "cyclic color map" that uses the same colors for the two
+  extremes of the range of the phase. Such color maps exist,
+  see <a href="https://nicoguaro.github.io/posts/cyclic_colormaps/">this
+  blog post of Nicolás Guarín-Zapata</a> or
+  <a href="https://stackoverflow.com/questions/23712207/cyclic-colormap-without-visual-distortions-for-use-in-phase-angle-plots">this
+  StackExchange post</a>, for example. The problem is that my favorite
```
peterrum, line 66:
```suggestion
  StackExchange post</a>, for example. The problem is that our favorite
```

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,840 @@
...
+first phase rotation half-step; the output of the current step will
+be denoted by $\psi^{(n,2)}$. $k_{n+1}=t_{n+1}-t_n$ is
+the length of the time step. (One could argue whether $\psi^{(n,1)}$
+and $\psi^{(n,1)}$ live at time step $n$ or $n+1$ and what their upper
+indices should be. This is a philosophical discussion without practical
+impact, and one might think of $\psi^{(n,1)}$ as something like
+$\psi^{(n+\tfrac 13)}$, and $\psi^{(n,2)}$ as
+$\psi^{(n+\tfrac 23)}$ if that helps clarify things -- though, again
+$n+\frac 13$ is not to be understood as "one third time step after
+$t_n$" but more like "we've already done one third of the work necessary
```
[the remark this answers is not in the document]
bangerth: No, I think this is right.

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,840 @@
...
+for time step $n+1$".)
```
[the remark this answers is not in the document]
bangerth: As is this.

include/deal.II/base/function.h
```diff
@@ -57,13 +57,13 @@ class TensorFunction;
  * @code
  * // access to one component at one point
  * double
- * value(const Point<dim> & p,
+ * value(const Point<dim>  &p,
```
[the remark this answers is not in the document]
bangerth: Because that's how I think clang-format would have formatted it, if it was regular code. But that's a slip -- it shouldn't have been included in this patch -- I'll take that out.

bangerth:
Thanks, @peterrum, for the suggestions -- I've incorporated all but two or three (which I commented on).

bangerth:
As discussed the other week, I'm accumulating patches here rather than fixing-up or squashing so that others can check out the branch. So whenever someone says that it's ready to merge -- don't merge, I need to tidy up the history first.

examples/step-58/step-58.cc
```diff
@@ -0,0 +1,660 @@
...
+      for (unsigned int q = 0; q < computed_quantities.size(); ++q)
+        {
+          Assert(computed_quantities[q].size() == 1,
+                 ExcDimensionMismatch(computed_quantities[q].size(), 1));
+          Assert(inputs.solution_values[q].size() == 2,
+                 ExcDimensionMismatch(inputs.solution_values[q].size(), 2));
+
+          computed_quantities[q](0) =
+            std::abs(std::complex<double>(inputs.solution_values[q](0),
+                                          inputs.solution_values[q](1)));
```
dangars, line 483: Great example of `DataProcessor` for complex valued equations!

Maybe you can mention in the text that the number of inputs is equal to two times the number of components. In this case you have one component, therefore you have two inputs. I guess that if you would have 3 components. You would have 6 inputs. Right?

In the case of 3 components (vector field). Which would be the order of the inputs:
- x_re, y_re, z_re, x_im, y_im, z_im
- x_re, x_im, y_re, y_im, z_re, z_im

I think that it would be good to document this.

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,207 @@
...
+consist of a number of variables: The solution (split into its real and imaginary
+parts), the magnitude, and the phase. If we visualize these four fields, we get
+images like the following after a few time steps (at time $t=0.242$, to be
+precise:
+
+<div class="twocolumn" style="width: 80%">
+  <div>
+    <img src="https://www.dealii.org/images/steps/developer/step-58.re.png"
+         alt="Real part of the solution at t=0.242"
+         width="400">
```
dangars, line 30: The colors are great, but when I compile the tutorial to view the doxygen version I do not see the legend, therefore I do not know what you are plotting.

Maybe you can increase the font size.

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,207 @@
...
+  </div>
+  <div>
+    <img src="https://www.dealii.org/images/steps/developer/step-58.magnitude.png"
+         alt="Magnitude of the solution at t=0.242"
+         width="400">
+  </div>
+  <div>
+    <img src="https://www.dealii.org/images/steps/developer/step-58.phase.png"
+         alt="Phase of the solution at t=0.242"
```
dangars, line 44: Is there a reason to have a grid in this plot? I would remove the grid

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,207 @@
...
+         width="400">
+  </div>
+</div>
+
+While the real and imaginary parts of the solution shown above are not
+particularly interesting (because, from a physical perspective, the
+global offset of the phase and therefore the balance between real and
+imaginary components, is meaningless), it is much more interesting to
+visualize the magnitude $|\psi(\mathbf x,t)|$ and phase
```
dangars, line 53: Wouldn't it make sense to plot  the square of the amplitude $|\psi(\mathbf x,t)|^2$ instead of the magnitude. The square of the amplitude is the probability distribution of finding a particle.

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,840 @@
...
+The <a
+href="https://en.wikipedia.org/wiki/Nonlinear_Schr%C3%B6dinger_equation">Nonlinear
+Schr&ouml;dinger Equation (NLSE)</a> for a function $\psi=\psi(\mathbf
+x,t)$ and a potential $V=V(\mathbf x)$ is a model often used in
+quantum mechanics and nonlinear optics. It reads as follows:
+@f{align*}{
+  - i \frac{\partial \psi}{\partial t}
+  - \frac 12 \Delta \psi
+  + V \psi
+  + \kappa |\psi|^2 \psi
```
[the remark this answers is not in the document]
bangerth: Fair point. I just took the form that's on wikipedia, but have now mentioned that one needs to measure in "appropriate units".

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,840 @@
...
+$\kappa|\psi|^2\psi$. However, these are of lower order in the spatial
+derivatives, and while they are obviously important, they do not
+change the character of the equation.
+
+In any case, the purpose of this discussion is to figure out
+what time stepping scheme might be appropriate for the equation. The
+conclusions is that, as a hyperbolic-kind of equation, we need to
+choose a time step that satisfies a CFL-type condition. If we were to
+use an explicit (which we will not), we would have to investigate
+the eigenvalues of the spatial operator. If you followed the discussions
```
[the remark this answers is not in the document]
bangerth: It's really about the eigenvalues of the matrix that corresponds to the spatial operator. I fixed this. But in general, the approach is the same for operators: You try to find, for example, those values `lambda` so that
```
  -Delta u(x) = lambda u(x)
```
The same thing can be done for any differential operator.

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,840 @@
...
+Whether you take the original equation ($s=1,t=2$) or the reformulation
+for only the real or imaginary part, the outcome is that we would need to
+choose $k \propto h^2$ if we were to use an explicit time stepping
+method. This is not feasible for the same reasons as in step-26 for
+the heat equation: It would yield impractically small time steps
+for even only modestly refined meshes. Rather, we have to use an
+implicit time stepping method and can then choose a more balanced
+$k \propto h$. Indeed, we will use the implicit Crank-Nicolson
+method as we have already done in step-25 before for the regular
+wave equation.
```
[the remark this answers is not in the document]
bangerth: Yes, I don't know either. I fixed the reference to step-23, but would like to avoid speculating about whether the C-N scheme conserves energy.

(I think that it might. The phase rotation part of the Strang splitting only, well, rotates the phase and so doesn't affect the energy. One would have to think about what happens with the the spatial half step. Separately: what should be preserved is actually the square of the wave function integrated over the domain. That's the particle number in the current case. It has the same form as the "energy" in other contexts.)

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,840 @@
...
+
+What we will use here, primarily because it makes for good graphics,
+is the following:
+@f[
+  \psi_0(\mathbf x) = \sum_{k=1}^4 \alpha_k e^{-\frac{r_k^2}{R^2}},
+@f]
+where $r_k = |\mathbf x-\mathbf x_k|$ is the distance from the (fixed)
+locations $\mathbf x_k$, and
+$\alpha_k$ are factors denoting the height of these functions.
+We choose $\alpha_k=1$ for all, and $R=0.1$. We choose the
```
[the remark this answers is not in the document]
bangerth: Ah, that's interesting. I had not thought about what the proper normalization is here.

I don't think it matters much for the current program and since all of the figures are already created, would you mind too much if I punt on running everything again if I make a note that that's what the normalization should be?

examples/step-58/step-58.cc
bangerth: That's actually discussed at length in `DataPostprocessor`. The order is `x_re, x_im, y_re, y_im, ...`.

examples/step-58/doc/results.dox
bangerth: Hm, that shows up just fine for me. I'm a bit reluctant to enlarge this part of the figure because I think that it's actually the least important one -- nobody should be interested in the absolute value of a probability density, it's about the pattern that I want to show and that I'd like to focus on.

examples/step-58/doc/results.dox
bangerth: Yes, that's actually an important part because I want to show that the phase changes from +close to pi to -close to pi over the course of just one cell. It refers to the discussion below.

examples/step-58/doc/results.dox
bangerth: Yes, that's a fair point. I guess I have to redo all of my figures again :-(

examples/step-58/doc/intro.dox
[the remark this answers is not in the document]
dangars: For the units you can mention [the Gross–Pitaevskii equation](https://en.wikipedia.org/wiki/Gross%E2%80%93Pitaevskii_equation)

examples/step-58/doc/intro.dox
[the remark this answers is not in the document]
dangars: I agree, for the current program it does not really matter. you can leave it that way. No need to change it. You can mention that the normalization would matter from a physics point of view. I just realized that I made a mistake in my calculation, because you have to use the square of the amplitude for the normalization. You can find more information [here](https://en.wikipedia.org/wiki/Gross%E2%80%93Pitaevskii_equation)

examples/step-58/doc/results.dox
dangars: It is not big deal if you plot the amplitude instead of the square of the amplitude :)

bangerth:
I changed my mind. It may be inconvenient to redo all of the figures, but it's the right thing to do. I'll have to think about the normalization constant as well while there. The problem is that I can easily integrate a Gaussian (I would pick the square root of a Gaussian as initial condition), but not the sum of square roots....

....ah, scratch that. I just need to take the square root of the sum!

Anyone else want to look over this?

bangerth:
@dangars -- if you want to take a look at the last couple of commits, you'll see the definition and implementation of the initial conditions.

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,207 @@
...
+global offset of the phase and therefore the balance between real and
+imaginary components, is meaningless), it is much more interesting to
+visualize the amplitude $|\psi(\mathbf x,t)|^2$ and phase
+$\text{arg}(\psi(\mathbf x,t))$ of the solution and, in particular,
+their evolution. This leads to pictures like the following:
+
+The phase picture shown here clearly has some flaws:
+- First, phase is a "cyclic quantity", but the color scale uses a
+  fundamentally different color for values close to $-\pi$ than
+  for values close to $+\pi$. This is a nuisance -- what we need
```
dangars, line 60: Very interesting and useful discussion.

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,875 @@
...
+  below.
+
+
+
+<h3>A note about the character of the equations</h3>
+
+At first glance, the equations appear to be parabolic and similar to
+the heat equation (see step-26) as there is only a single time
+derivative and two spatial derivatives. But this is misleading and
+more easily seen if we assume for a moment that the potential $V=0$
```
jppelteret, line 79: >  But this is misleading and more easily seen

This reads a bit strangely to me. I can't help but interpret what is "more easily seen" as being the parabolic nature of the equations, which is clearly what you're trying to indicate is not correct. Maybe something like "But this is misleading and their true nature can be more easily seen" is better?

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,875 @@
...
+  dt,
+@f}
+can be solved exactly: the equation is solved by
+@f{align*}{
+  \psi^{(3)}(t) = e^{-i\kappa|\psi(t_n)|^2 (t-t_{n})} \psi(t_n).
+@f}
+This is easy to see if (i) you plug this solution into the
+differential equation, and (ii) realize that the magnitude
+$|\psi^{(3)}|$ is constant, i.e., the term $|\psi(t_n)|^2$ in the
+exponent  is in fact equal to $|\psi^{(3)}(t)|^2$. In other words, the
```
jppelteret, line 367: `exponent  is` -> `exponent is`

(double space, although I'm sure it'd render just fine)

examples/step-58/doc/intro.dox
```diff
@@ -0,0 +1,875 @@
...
+for the $\psi^{(k)}$ start all at $\psi(t_n)$, but instead let the
+trajectory for $\psi^{(2)}$ start at the <i>end point</i> of the
+trajectory for $\psi^{(1)}$, namely $\psi^{(1)}(t_{n+1})$; similarly,
+we will start the trajectory for $\psi^{(3)}$ start at the end point
+of the trajectory for $\psi^{(2)}$, namely $\psi^{(2)}(t_{n+1})$. This
+method is then called "Lie splitting" and has the same order of error
+as the method above, i.e., the splitting error is ${\cal O}(\Delta
+t)$.
+
+In formulas, this variation of operator splitting can be written as
```
jppelteret, line 408: > In formulas, ...

Is the start of this sentence perhaps missing something?

jppelteret: Or maybe its easier to write something like
"This variation of operator splitting can be formulated as follows..."

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,207 @@
...
+of the video.
+
+
+<a name="extensions"></a>
+<h3>Possibilities for extensions</h3>
+
+<h4> Better linear solvers </h4>
+
+The solver chosen here is just too simple. It is also not efficient.
+What we do is just punt the matrix to a sparse direct solver in every
```
jppelteret, line 157: > punt the matrix

Hmm... I don't know if some of our second/third/... language English readers are going to understand this expression.

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,207 @@
...
+  not be true in actual applications. But at least in cases where this
+  does happen to be the case, it would make sense to only factorize
+  the matrix once (i.e., compute $L$ and $U$ factors once) and then
+  use these factors for all following time steps until the matrix
+  $C$ changes and requires a new factorization. The interface of the
+  SparseDirectUMFPACK class allows for this.
+
+- Ultimately, however, sparse direct solvers are only efficient for
+  relatively small problems, say up to a few 100,000 unknowns. Beyond
+  this, one needs iterative solvers such as Conjugate Gradients (for
```
jppelteret, line 174: `Conjugate Gradients` -> `Conjugate Gradient` / `the family of Conjugate Gradient approaches`

I believe that you are implying that CG/CGS/FlexCG are suitable, but I wonder if the link between `Conjugate Gradient` and the canonical CG method is just too strong for the plural form that you chose to be easily understood.

examples/step-58/doc/results.dox
```diff
@@ -0,0 +1,207 @@
...
+
+In order to be usable for actual, realistic problems, solvers for the
+nonlinear Schr&ouml;dinger equation need to utilize boundary conditions
+that make sense for the problem at hand. We have here restricted ourselves
+to simple Neumann boundary conditions -- but these do not actually make
+sense for the problem. Indeed, the equations are generally posed on an
+infinite domain. But, since we can't compute on infinite domains, we need
+to truncate it somewhere and instead pose boundary conditions that make
+sense for this artificially small domain. The approach widely used is to
+use the Perfectly Matches Layer method that corresponds to a particular
```
jppelteret, line 195: `Perfectly Matches Layer` -> `Perfectly Matched Layer` ([wiki link for reference)](https://en.wikipedia.org/wiki/Perfectly_matched_layer)?

jppelteret:
I've looked at the text in all of the docs (I won't pretend that I spent much time trying to absorb the maths at this hour), but haven't reviewed the source code. It's looking pretty good!

examples/step-58/doc/intro.dox
jppelteret: I see that you've introduced some formulas several times in the same manner, so with that in mind I guess that you need not consider changing this if you don't feel like doing so.

examples/step-58/step-58.cc
```diff
@@ -0,0 +1,673 @@
...
+  //  12\Delta t} \; \psi^{(n,0)}$. Here, $\psi^{(n,0)}=\psi^{(n)}$ and
+  //  $\psi^{(n,1)}$
+  // are functions of space and correspond to the output of the previous
+  // complete time step and the result of the first of the three part steps,
+  // respectively. A corresponding solution must be computed for the third
+  // of the part steps, i.e.
+  // $\psi^{(n,3)} = e^{-i\kappa|\psi^{(n,2)}|^2 \tfrac
+  //  12\Delta t} \; \psi^{(n,2)}$, where $\psi^{(n,3)}=\psi^{(n+1)}$ is
+  // the result of the time step as a whole, and its input $\psi^{(n,2)}$ is
+  // the result of the spatial step of the Strange splitting.
```
jppelteret, line 349: `Strange splitting` -> `Strang splitting`

examples/step-58/step-58.cc
```diff
@@ -0,0 +1,673 @@
...
+  template <int dim>
+  void NonlinearSchroedingerEquation<dim>::do_half_phase_step()
+  {
+    for (auto &value : solution)
+      {
+        const std::complex<double> i         = {0, 1};
+        const double               magnitude = std::abs(value);
+
+        value = std::exp(-i * kappa * magnitude * magnitude * (time_step / 2)) *
+                value;
```
jppelteret, line 385: Couldn't you write this as `value *= std::exp(...);`? Then its really clear that it represents an update.

examples/step-58/step-58.cc
```diff
@@ -0,0 +1,673 @@
...
+                                                     dofs_per_cell);
+
+    std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
+    std::vector<double>                  potential_values(n_q_points);
+    const Potential<dim>                 potential;
+
+    for (const auto &cell : dof_handler.active_cell_iterators())
+      {
+        cell_matrix_lhs = std::complex<double>(0.);
+        cell_matrix_rhs = std::complex<double>(0.);
```
jppelteret, line 281: I'm kind of surprised that we don't do type promotion for the RHS. In the docs I believe (although I haven't double checked) that we write something about `cell_matrix = 0` being *the* way to reset a matrix. Maybe we need to check/update the docs, or add some `operator=(int)` to `FullMatrix` that does this internally.

examples/step-58/step-58.cc
```diff
@@ -0,0 +1,673 @@
...
+
+      virtual void evaluate_vector_field(
+        const DataPostprocessorInputs::Vector<dim> &inputs,
+        std::vector<Vector<double>> &computed_quantities) const override;
+    };
+
+
+    template <int dim>
+    ComplexAmplitude<dim>::ComplexAmplitude()
+      : DataPostprocessorScalar<dim>("Magnitude", update_values)
```
jppelteret, line 474: Out of curiosity, is there a reason as to why the class is named `ComplexAmplitude` but the title that you embed in the output is `Magnitude`? I mean, it doesn't really matter and both seem fine, but the inconsistency is a bit unexpected.

examples/step-58/step-58.cc
kronbichler: @jppelteret you should not have made that suggestion - I would have loved to learn about that other type of splitting, the name sounded promising :rofl:

jppelteret: @kronbichler Indeed, it was a missed opportunity to have some fun. But there's still time for you to claim ownership of it ;-) How would you choose to define Strange splitting?

examples/step-58/doc/results.dox
bangerth: Good point -- it's an American Football reference where a "punt" is when you voluntarily give up the ball because you don't know anything better to do with it.

examples/step-58/step-58.cc
bangerth: I could, but I'd like to keep it this way because it matches the formulas: `psi^{n+1} = exp(...) psi^n`. Using `*=` obscures the fact that we actually move from one stage to another.

examples/step-58/step-58.cc
bangerth: You totally all missed out on having a field day with my mistake!

examples/step-58/step-58.cc
bangerth: Nope, it's a mistake. I used to output the *magnitude* and forgot to change this one occurrence.

examples/step-58/step-58.cc
bangerth: Yes, adding `operator=(int)` would resolve this issue. The problem is that there is no automatic conversion of either `double` to `std::complex<double>`, not `int` to `std::complex<double>`. I think that's actually a mistake in the standard, but that's outside my realm of action.

bangerth:
I've made all suggested changes. Anyone else?

If not, if someone were willing to approve the patch, then I'll rebase and squash into a self-contained set of commits.

jppelteret:
@bangerth Your changes look good. Since nobody else has commented, please squash the commits and feel free to merge in the new tutorial :-) Thanks a lot for writing it. It's really interesting, and a cool application! And its nice to have an example where we solve a complex-valued problem without the real-imaginary split.

Thanks as well to @dangars for the expert review. My final perusal of the tutorial was relatively relaxed after his and @peterrum's thorough looks through it.

bangerth:
Thanks indeed to everyone who reviewed! I've squashed this to a small set of self-contained commits. Ready to merge from my side!

bangerth:
Yay!

PR #9897, opened 2020-04-15
Use the same variable name in all output files in step-52.

bangerth:
#9833 appended the method name to the name of the solution variable, but this is awkward because one can then no longer load a different set of file names (the program generates output for a variety of different time integrators): Visit will complain that the variable that was plotted before is no longer in the newly loaded input file and throw away all plots one had previously created.

@krishnakumarg1984 -- FYI.

/rebuild

### DIFF
--- a/examples/step-52/step-52.cc
+++ b/examples/step-52/step-52.cc
@@ -445,19 +445,19 @@ namespace Step52
             break;
           }
       }
-    const std::string solution_name = "solution_" + method_name;
 
     DataOut<2> data_out;
 
     data_out.attach_dof_handler(dof_handler);
-    data_out.add_data_vector(solution, solution_name);
+    data_out.add_data_vector(solution, "solution");
 
     data_out.build_patches();
 
     data_out.set_flags(DataOutBase::VtkFlags(time, time_step));
 
-    const std::string filename =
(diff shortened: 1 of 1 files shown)


masterleinad:
Setting `Reviewed and ready to merge` without someone reviewing seems a little weird.

bangerth:
Oops, not what I meant... I meant to set "Ready for review..."

krishnakumarg1984:

bangerth:
@krishnakumarg1984: That's a separate issue. It may also explain the convergence issues we've documented in another PR. It's unrelated to this one -- but feel free to open a separate issue.

krishnakumarg1984:
>That's a separate issue. 

Sorry Prof Bangerth, but I don't quite understand how is it a separate issue. The time-steps don't really end at the desired time.

masterleinad:
This pull request was aiming at fixing the name of the output files. Hence, if the time-stepping methods reach the final time, is a separate issue.

krishnakumarg1984:
Yes. I understand. I am just unable to express myself of how the adaptive time-steppers fail to reach and stop at exactly the desired end time. I am hoping one of the senior developers here shall help to look into this issue with a simple 1D diffusion equation maybe.

Issue #9684, opened 2020-03-19
[Step-52] Dopri error in the results incorrect?


krishnakumarg1984:
In the [9.1](https://dealii.org/current/doxygen/deal.II/step_52.html#Results) as well as the [dev](https://dealii.org/developer/doxygen/deal.II/step_52.html#Results) version of step-52 tutorial, the dopri error is reported as __0.0165485__  taking __119__ steps to achieve the desired tolerance.

However, in my simulations, I get dopri error to be __5.59e-5__ . This can't just be written off due to machine-specific rounding error since this is three orders of magnitude lower than that reported in the tutorial, with roughly the same no. of steps (121 in my case).

So, the Dormand-Prince simulator seems to be the most accurate one. Can someone else independently verify this?

bangerth:
What happens if you try this with previous versions of deal.II?

krishnakumarg1984:
@bangerth Did you mean v8.5.0 or v9.0.0?   I tried it with v9.1.1 .

Do we expect a difference due to some improvement in the ODE solver between those releases? I am asking because it is non-trivial effort to compile and run an older version of deal.ii (that tutorial also requires the additional effort of configuring the UMFPACK interface.)

bangerth:
On 3/19/20 4:44 PM, Krishnakumar Gopalakrishnan wrote:
> @bangerth <https://github.com/bangerth> Did you mean v8.5.0 or v9.0.0? I tried 
> it with v9.1.1 .

Either would be interesting to try. It is not uncommon that there are 
improvements in library code that have an effect on a tutorial program -- but 
if nobody notices at the time, then the tutorial is never updated. It would be 
interesting to see when that change happened.


> Do we expect a difference due to some improvement in the ODE solver between 
> those releases? I am asking because it is non-trivial effort to compile and 
> run an older version of deal.ii (that tutorial also requires the additional 
> effort of configuring the UMFPACK interface.)

The latter should always be available, shouldn't it? I think that at least for 
this tutorial, you don't need any of the external packages.

krishnakumarg1984:
Looks like the corresponding [source file for that tutorial](https://github.com/dealii/dealii/blob/master/examples/step-52/doc/results.dox) hasn't been updated for the last 4 years.  (Sep 16 , 2016).  

Looks like there was a deal.II version 8.4.2 that released a few days prior to that (Sep 3, 2016). Maybe I can try with that. 

[...]

PR #9899, opened 2020-04-15
Make the output of step-52 a bit easier to read.

bangerth:
Just indenting things make it easier. With #9899, we now get this:
```
Explicit methods:
   Forward Euler:            error=1.00883
   Third order Runge-Kutta:  error=0.000227982
   Fourth order Runge-Kutta: error=1.90541e-06

Implicit methods:
   Backward Euler:           error=1.03428
   Implicit Midpoint:        error=0.00862702
   Crank-Nicolson:           error=0.00862675
   SDIRK:                    error=0.0042349

Embedded explicit methods:
   Heun-Euler:               error=0.0073012
                steps performed=284
   Bogacki-Shampine:         error=0.000403281
                steps performed=181
   Dopri:                    error=5.07155e-05
                steps performed=121
   Fehlberg:                 error=0.00401035
                steps performed=106
   Cash-Karp:                error=0.00156445
                steps performed=107
```
That's better than before, though the "steps performed=" part is still distracting. Will get to that some other time.

Related to #9833 .
@krishnakumarg1984 -- FYI.

/rebuild

### DIFF
--- a/examples/step-52/step-52.cc
+++ b/examples/step-52/step-52.cc
@@ -692,21 +692,21 @@ namespace Step52
                     n_time_steps,
                     initial_time,
                     final_time);
-    std::cout << "Forward Euler:            error=" << solution.l2_norm()
+    std::cout << "   Forward Euler:            error=" << solution.l2_norm()
               << std::endl;
 
     explicit_method(TimeStepping::RK_THIRD_ORDER,
                     n_time_steps,
                     initial_time,
                     final_time);
-    std::cout << "Third order Runge-Kutta:  error=" << solution.l2_norm()
+    std::cout << "   Third order Runge-Kutta:  error=" << solution.l2_norm()
               << std::endl;
 
     explicit_method(TimeStepping::RK_CLASSIC_FOURTH_ORDER,
                     n_time_steps,
                     initial_time,
                     final_time);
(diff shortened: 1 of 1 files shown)


krishnakumarg1984:

drwells:
We do need to make sure that we print out output at the end of time iteration but lets merge this since it improves something else.

krishnakumarg1984:
The reason why I am commenting here is to bring to attention that these solvers don't seem to stop at the expected time-step. They mostly undershoot.

Rombur:
@krishnakumarg1984 As explained [here](https://github.com/dealii/dealii/issues/9684#issuecomment-648950654), the solvers do not undershoot. It looks like they stop early in paraview because we don't write an output file at every single time step.

jppelteret:
> As explained here, the solvers do not undershoot. It looks like they stop early in paraview because we don't write an output file at every single time step.

Is this worth documenting in the tutorial? (I haven't looked at the tut myself, so apologies if it's already explained in there.)

krishnakumarg1984:
>Is this worth documenting in the tutorial? (I haven't looked at the tut myself, so apologies if it's already explained in there.)

Yes! I support this idea.

Actually, taking it one step further, I believe that the right thing to do is to develop some logic to write the output at exactly the finish time.

PR #9905, opened 2020-04-16
Remove some unneeded static casts.

drwells:
These types are all printable now (see 7a3b4065d4d) so we don't need these static casts.

### DIFF
--- a/source/grid/grid_in.cc
+++ b/source/grid/grid_in.cc
@@ -867,8 +867,7 @@ GridIn<dim, spacedim>::read_ucd(std::istream &in,
           if (apply_all_indicators_to_manifolds)
             cells.back().manifold_id =
               static_cast<types::manifold_id>(material_id);
-          cells.back().material_id =
-            static_cast<types::material_id>(material_id);
+          cells.back().material_id = material_id;
 
           // transform from ucd to
           // consecutive numbering
@@ -1876,8 +1875,7 @@ GridIn<dim, spacedim>::read_msh(std::istream &in)
                 // numbers::invalid_material_id-1
                 AssertIndexRange(material_id, numbers::invalid_material_id);
 
-                cells.back().material_id =
-                  static_cast<types::material_id>(material_id);
+                cells.back().material_id = material_id;
 
                 // transform from ucd to
                 // consecutive numbering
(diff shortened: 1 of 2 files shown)


bangerth:
I get the cases where you create output, but the places where you assign `manifold_id` to `material_id` or the other way around, at least conceptually, use different types. They don't *need* the cast because the underlying type happens to be the same, but the cast helps read more easily that something fishy is going on here where objects of different semantics are assigned to each other.

drwells:
Good point - we should keep those.

drwells:
Fixed!

source/grid/grid_in.cc
```diff
@@ -917,8 +916,7 @@ GridIn<dim, spacedim>::read_ucd(std::istream &in,
             }
           else
             {
-              subcelldata.boundary_lines.back().boundary_id =
-                static_cast<types::boundary_id>(material_id);
+              subcelldata.boundary_lines.back().boundary_id = material_id;
```
bangerth, line 919: X

source/grid/grid_in.cc
```diff
@@ -968,8 +966,7 @@ GridIn<dim, spacedim>::read_ucd(std::istream &in,
             }
           else
             {
-              subcelldata.boundary_quads.back().boundary_id =
-                static_cast<types::boundary_id>(material_id);
+              subcelldata.boundary_quads.back().boundary_id = material_id;
```
bangerth, line 969: X

source/grid/grid_in.cc
```diff
@@ -1914,8 +1910,7 @@ GridIn<dim, spacedim>::read_msh(std::istream &in)
                 AssertIndexRange(material_id,
                                  numbers::internal_face_boundary_id);
 
-                subcelldata.boundary_lines.back().boundary_id =
-                  static_cast<types::boundary_id>(material_id);
+                subcelldata.boundary_lines.back().boundary_id = material_id;
```
bangerth, line 1913: X

source/grid/grid_in.cc
```diff
@@ -1954,8 +1949,7 @@ GridIn<dim, spacedim>::read_msh(std::istream &in)
                 AssertIndexRange(material_id,
                                  numbers::internal_face_boundary_id);
 
-                subcelldata.boundary_quads.back().boundary_id =
-                  static_cast<types::boundary_id>(material_id);
+                subcelldata.boundary_quads.back().boundary_id = material_id;
```
bangerth, line 1952: X

drwells:
I agree - I meant to fix those when I fixed the manifold ids. I'll push a fix in a minute.

drwells:
Fixed!

PR #1014, opened 2015-06-15
Deprecate more vector functions

Rombur:
Deprecate the following functions: sadd with 7 arguments, equ with 6 arguments, and print with 3 arguments.

PR #991, opened 2015-06-03
[RFC] Clean distributed vectors

Rombur:
I have deprecated a few functions for distributed vectors. I tried to get ride of functions that exist only for one kind of vector. I had to move some functions in trilinos_parallel_block_vector.cc otherwise including trilinos_parallel_block_vector.h would trigger a warning. I haven't deprecated as many functions as I could because we use them in the tutorials (for example equ with six arguments). Let me know if you think that we should deprecate more.

PR #9906, opened 2020-04-16
Remove deprecated Vector::print overloads

masterleinad:
Deprecated in #991 and #1014.


tests/base/function_parser_05.cc
```diff
@@ -85,7 +85,9 @@ test2()
 
   // Output the evaluated function
   deallog << "Function '" << expressions[0] << "," << expressions[1] << "'"
-          << " @ " << point << " is " << result << std::endl;
+          << " @ " << point << " is " << std::flush;
+  result.print(deallog.get_file_stream(), 4, false);
```
bangerth, line 89: Is this a mistake? I thought you were *removing* the `print` function.

masterleinad: We are removing the overload taking `LogStream` but we still allow `std::ostream`.

masterleinad: Or are you saying that we shouldn't remove the `operator<<(LogStream,...)` function that is implemented in terms of `print` as well?

bangerth: No, I was just confused why we are moving from `operator<<` to `results.print(...)`. I hadn't realized that you were also removing `operator<<(LogStream)`.

But if it were up to me, we'd completely remove `LogStream` to begin with...

PR #6304, opened 2018-04-23
Update gridgenerator examples 2

drwells:
Part of #6274.

I went ahead and regenerated the pictures for step-10. These are now SVGs and look much nicer.

PR #9907, opened 2020-04-16
Do not include fe_q.h in mapping_q_generic.h

kronbichler:
Given that #9826 removed the `FE_Q` from `MappingQGeneric`, I think it is appropriate to not include `deal.II/fe/fe_q.h` in that file. Since the header `mapping_q_generic.h` is included in a central place (due to the ubiquitous use of the linear variant), this means that some places that used to work before now need to include `fe_q.h` explicitly, see the list of 51 files in the library and testsuite in this patch.

I guess this would deserve a changelog in the *incompatibilities section*. I have not added it yet because I wanted to wait if there is objections here. This is what I would write:
```
Changed: The header file `deal.II/fe/fe_q.h` is not included implicitly via
`deal.II/fe/mapping_q_generic.h` any more. Add the header `#include <deal.II/fe/fe_q.h>`
if necessary.
```

### DIFF
--- a/examples/step-10/step-10.cc
+++ b/examples/step-10/step-10.cc
@@ -333,8 +333,8 @@ namespace Step10
         Triangulation<dim> triangulation;
         GridGenerator::hyper_ball(triangulation);
 
-        const MappingQ<dim> mapping(degree);
-        const FE_Q<dim>     fe(1);
+        const MappingQ<dim>   mapping(degree);
+        const FE_Nothing<dim> fe;
 
         DoFHandler<dim> dof_handler(triangulation);
 
--- a/doc/news/changes/incompatibilities/20200416MartinKronbichler
+++ b/doc/news/changes/incompatibilities/20200416MartinKronbichler
@@ -1,0 +1,5 @@
+Changed: The header file `deal.II/fe/fe_q.h` is not included implicitly via
+`deal.II/fe/mapping_q_generic.h` any more. Add the header
+`#include <deal.II/fe/fe_q.h>` if necessary.
+<br>
+(Martin Kronbichler, 2020/04/16)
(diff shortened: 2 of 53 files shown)


examples/step-10/step-10.cc
```diff
@@ -333,8 +333,8 @@ namespace Step10
         Triangulation<dim> triangulation;
         GridGenerator::hyper_ball(triangulation);
 
-        const MappingQ<dim> mapping(degree);
-        const FE_Q<dim>     fe(1);
+        const MappingQ<dim>   mapping(degree);
+        const FE_Nothing<dim> fe;
```
bangerth, line 337: Hm, we haven't discussed that element up until that point, whereas everyone is familiar with `FE_Q`. The actual choice of element doesn't make a difference here, so I'd just stick with `FE_Q` -- the cost is unimportant here anyway.

kronbichler: I started doing as you say before, but then I found this in the piece for the area 120 lines above (the part you see here is for the perimeter):
https://github.com/dealii/dealii/blob/31ab106f152ff9a14708f08291f9937bc477bbae/examples/step-10/step-10.cc#L209-L216
I can of course revert to `FE_Q`, but I think that goes back to #6304.

bangerth: Ah, who knew? If so, then of course leaving it as is is the right way to go! Thanks for finding this!

bangerth:
Please add the changelog entry you mentioned and merge yourself (all checks succeed right now).

examples/step-10/step-10.cc
kronbichler: That tutorial was one of the first build errors I got, so my concentration level was still high :smile:

PR #9913, opened 2020-04-17
Add a second constructor to DiscreteTime.

bangerth:
@rezarastak -- FYI.

/rebuild


rezarastak:
I think this PR can be closed because its changes became unnecessary after merging #9748

PR #9916, opened 2020-04-17
Improve step 8

drwells:
I have a student who is going to do a little bit of work on solid mechanics and I pointed him at this step. I went ahead and improved a few small things:
1. We now set up constraints in the same way we do in step-6.
2. I removed `<code></code>` so that we get more doxygen links.
3. We don't need a destructor.

### DIFF
--- a/examples/step-8/step-8.cc
+++ b/examples/step-8/step-8.cc
@@ -70,11 +70,10 @@ namespace Step8
   // the step-6 example.
   //
   // The only change is the use of a different class for the <code>fe</code>
-  // variable: Instead of a concrete finite element class such as
-  // <code>FE_Q</code>, we now use a more generic one,
-  // <code>FESystem</code>. In fact, <code>FESystem</code> is not really a
+  // variable: Instead of a concrete finite element class such as FE_Q, we now
+  // use a more generic one, FESystem. In fact, FESystem is not really a
   // finite element itself in that it does not implement shape functions of
-  // its own.  Rather, it is a class that can be used to stack several other
+  // its own. Rather, it is a class that can be used to stack several other
   // elements together to form one vector-valued finite element. In our case,
   // we will compose the vector-valued element of <code>FE_Q(1)</code>
   // objects, as shown below in the constructor of this class.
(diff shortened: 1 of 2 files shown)


examples/step-8/step-8.cc
```diff
@@ -598,13 +559,12 @@ namespace Step8
...
-  // <code>triangulation</code> object will not flag any cells for refinement
-  // (why should it if the indicated error is zero for each cell?). The grid
-  // in the next iteration will therefore consist of four cells only as well,
-  // and the same problem occurs again.
+  // the error indicators computed by the KellyErrorEstimator class are zero
+  // for each cell as well, and the call to
+  // Triangulation::refine_and_coarsen_fixed_number() on the will not flag any
+  // cells for refinement (why should it if the indicated error is zero for
+  // each cell?). The grid in the next iteration will therefore consist of
+  // four cells only as well, and the same problem occurs again.
```
tjhei, line 567: Can you please fix the "twice" in the paragraph below as well?

examples/step-8/step-8.cc
```diff
@@ -598,13 +559,12 @@ namespace Step8
...
-  // the error indicators computed by the <code>KellyErrorEstimator</code>
-  // class are zero for each cell as well, and the call to
-  // <code>refine_and_coarsen_fixed_number</code> on the
-  // <code>triangulation</code> object will not flag any cells for refinement
-  // (why should it if the indicated error is zero for each cell?). The grid
-  // in the next iteration will therefore consist of four cells only as well,
-  // and the same problem occurs again.
+  // the error indicators computed by the KellyErrorEstimator class are zero
+  // for each cell as well, and the call to
+  // Triangulation::refine_and_coarsen_fixed_number() on the will not flag any
```
tjhei, line 564: On the will? 😀

drwells: whoops - I only got partway through 'on the triangulation'. Let me fix that.

drwells: Fixed.

examples/step-8/step-8.cc
drwells: It's been a long week. Where is the issue here?

drwells: oh, we refine four times.

drwells: Also fixed!

Issue #9903, opened 2020-04-15
Check the usage of OutputStyle::XML and ::JSON in ParameterHandler


peterrum:
As noted in https://github.com/dealii/dealii/issues/9888#issuecomment-613850600, the following lines:

https://github.com/dealii/dealii/blob/5637bcf9db09af419ea5bbb3c92e5b262fc603aa/source/base/parameter_handler.cc#L1331-L1332

should be:
```cpp
  // this function should not be necessary for XML or JSON output...
  Assert((style != XML) && (style != ShortXML) && 
         (style != JSON) && (style != ShortJSON), ExcInternalError());
```
or 
```cpp
  // this function should not be necessary for XML or JSON output...
  Assert(!((style & XML) || (style & JSON)), ExcInternalError());
```

nfehn:
i guess this issue can also be closed now?

PR #9919, opened 2020-04-17
Extend OutputStyle

### LINKED ISSUE #9888: Configuration of I/O of ParameterHandler
Recently there have been a couple of additions to the `ParameterHandler` class, in particular to its I/O functionalities. 

In the case of output, the additions are:
* `ShortXML` and `ShortJSON`
* the optional `sort_alphabetical` flag

To handle all the new cases and the old cases and to be backwards compatible, we could modify the `OutputStyle` the following way:
```cpp
  enum OutputStyle {
    with_documentation = 0,                      //  0
    without_documentation = 1,                   //  1

    sort_alphabetically = 2                      //  2

    Text = 4,                                    //  4
    ShortText = Text | without_documentation,    //  5

    LaTex = 8,                                   //  8

    Description = 16,                            // 16

    JSON = 32,                                   // 32
    ShortXML = XML | without_documentation,      // 33

    JSON = 64,                                   // 64
    ShortJSON = JSON | without_documentation,    // 65
  }
```

Any options on such a way to configure the output style (motivated by update flags of `FEValues`)?

In the case of input, the additions are:
* `skip_undefined` for XML and JSON files
* the flag `assert_mandatory_entries_are_found`

Wouldn't it be a good time to introduce a struct for these so that the list of defaulted parameters in function call doesn't become too long?

@elauksap @nfehn This might be also relevant for you.

peterrum:
A first draft. An open question is how to treat `Description` and if we need `DefaultStyle`.

The many changed loc are related to the fact that some switch-cases had to be replaced by if-else constructions.

Closes #9888 and #9903.

### DIFF
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -846,57 +846,113 @@ class ParameterHandler : public Subscriptor
 {
 public:
   /**
-   * List of possible output formats used for
-   * ParameterHandler::print_parameters().
+   * List of possible output formats used for functions like
+   * ParameterHandler::print_parameters(). The options can be categorized into
+   * two groups:
+   * - format options: PRM, LaTeX, Description, XML, JSON
+   * - stylistic options: Short, KeepDeclarationOrder
+   *
+   * A valid combination of options is given in the following table:
+   *
+   * Stylistic options    | Default | PRM | LaTeX | Description | XML | JSON
+   * -------------------- | ------- | --- | ----- | ----------- | --- | ----
+   * Default              |    X    |  X  |   X   |      X      |  X  |  X
+   * Short                |    X    |  X  |   -   |      X      |  X  |  X
+   * KeepDeclarationOrder |    X    |  X  |   X   |      X      |  X  |  X
+   *
+   * Valid combinations are checked at the beginning of the functions that
+   * accept an OutputStyle as an input argument.
+   *
+   * Furthermore, a number of shortcuts of commonly used option combinations are
+   * provided. E.g., ShortPRM prints the parameters in the PRM
+   * format, while skipping the documentation.
    */
   enum OutputStyle
   {
     /**
-     * Write human readable output suitable to be read by ParameterHandler
-     * again.
+     * Default stylistic style: print documentation and sort all parameters
+     * alphabetically.
      */
-    Text = 1,
-
-    /**
-     * Write parameters as a LaTeX table.
-     */
-    LaTeX = 2,
-    /**
-     * Write out declared parameters with description and possible values.
-     */
-    Description = 3,
-
-    /**
-     * Write out everything as an <a
-     * href="http://en.wikipedia.org/wiki/XML">XML</a> file.
-     *
-     * See the general documentation of this class for an example of output.
-     */
-    XML = 4,
-
-    /**
-     * Write out everything as a <a
-     * href="http://en.wikipedia.org/wiki/JSON">JSON</a> file.
-     */
-    JSON = 5,
+    DefaultStyle = 0x0000,
 
     /**
      * Write input for ParameterHandler without comments or changed default
      * values.
      */
-    ShortText = 193,
+    Short = 0x0001,
+
+    /**
+     * Keep the order of the parameters as they have been declared.
+     */
+    KeepDeclarationOrder = 0x0002,
+
+    /**
+     * Write human readable output suitable to be read by
+     * ParameterHandler::parse_input() again.
+     */
+    PRM = 0x0010,
+
+    /**
+     * Write human readable output suitable to be read by
+     * ParameterHandler::parse_input() again.
+     *
+     * @deprecated Use `PRM` instead of `Text`.
+     */
+    Text = PRM,
+
+    /**
+     * Write parameters as a LaTeX table.
+     */
+    LaTeX = 0x0020,
+
+    /**
+     * Write out declared parameters with description and possible values.
+     *
+     * @note This format is not suitable to be read back again.
+     */
+    Description = 0x0040,
+
+    /**
+     * Write out everything as an <a
+     * href="http://en.wikipedia.org/wiki/XML">XML</a> file suitable to be read
+     * by ParameterHandler::parse_input_from_xml() again.
+     *
+     * See the general documentation of this class for an example of output.
+     */
+    XML = 0x0080,
+
+    /**
+     * Write out everything as a <a
+     * href="http://en.wikipedia.org/wiki/JSON">JSON</a> file suitable to be
+     * read by ParameterHandler::parse_input_from_json() again.
+     */
+    JSON = 0x0100,
+
+    /**
+     * Write input for ParameterHandler without comments or changed default
+     * values.
+     */
+    ShortPRM = PRM | Short,
+
+    /**
+     * Write input for ParameterHandler without comments or changed default
+     * values.
+     *
+     * @deprecated Use `ShortPRM` instead of `ShortText`.
+     */
+    ShortText = ShortPRM,
 
     /**
      * Write input for ParameterHandler without comments or changed default
      * values as a XML file.
      */
-    ShortXML = 194,
+    ShortXML = XML | Short,
 
     /**
      * Write input for ParameterHandler without comments or changed default
      * values as a JSON file.
      */
-    ShortJSON = 195
+    ShortJSON = JSON | Short,
   };
 
 
@@ -1345,19 +1401,16 @@ public:
   set(const std::string &entry_name, const bool new_value);
 
   /**
-   * Print all parameters with the given style to @p out.
+   * Print all parameters with the given @p style to @p out.
    *
    * Before printing, all current parameters and subsections are sorted
    * alphabetically by default.
-   * This behavior can be disabled setting the last parameter @p sort_alphabetical
-   * to <tt>false</tt>: in this case entries are printed in the same order
-   * as they have been declared.
+   * This behavior can be disabled setting the optional parameter @p style
+   * to <tt>KeepDeclarationOrder</tt>: in this case entries are printed in the
+   * same order as they have been declared.
    *
-   * In the case of <tt>XML</tt> or <tt>JSON</tt>, a reduced tree, only
-   * containing the values and skipping the documentation, can be
-   * printed by setting @p print_documentation to <tt>false</tt>.
-   *
-   * In <tt>Text</tt> format, the output is formatted in such a way that it is
+   * In <tt>PRM</tt>, <tt>XML</tt>, and <tt>JSON</tt> format, the output is
+   * formatted in such a way that it is
    * possible to use it for later input again. This is most useful to record
    * the parameters for a specific run, since if you output the parameters
    * using this function into a log file, you can always recover the results
@@ -1367,6 +1420,12 @@ public:
    * default value of entries if it is different from the actual value, as
    * well as the documenting string given to the declare_entry() function if
    * available.
+   *
+   * By using the flag <tt>Short</tt> in combination with <tt>PRM</tt>,
+   * <tt>XML</tt>, or <tt>JSON</tt> (or by using the shortcuts
+   * <tt>ShortPRM</tt>, <tt>ShortXML</tt>, or <tt>ShortJSON</tt>), a reduced
+   * output can be generated, only containing the values and skipping the
+   * documentation.
    *
    * In <tt>XML</tt> format, the output starts with one root element
    * <tt>ParameterHandler</tt> in order to get a valid XML document and all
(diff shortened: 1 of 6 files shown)


source/base/parameter_handler.cc
```diff
@@ -1329,285 +1328,269 @@ ParameterHandler::recursively_print_parameters(
   AssertThrow(out, ExcIO());
 
   // this function should not be necessary for XML or JSON output...
-  Assert((style != XML) && (style != JSON), ExcInternalError());
+  Assert(((style & XML) != XML) && ((style & JSON) != JSON),
```
nfehn, line 1331: XML has to be written twice, JSON has to be written twice

source/base/parameter_handler.cc
```diff
@@ -1276,35 +1275,35 @@ ParameterHandler::print_parameters(std::ostream &    out,
...
-  if (style == JSON || style == ShortJSON)
+  if (style & JSON)
     {
       write_json(out, current_entries);
       return out;
     }
 
   // for all of the other formats, print a preamble:
-  switch (style)
+  if ((style & (Text | Short)) == (Text | Short))
```
nfehn, line 1285: `Text | Short` has to be written twice

source/base/parameter_handler.cc
```diff
@@ -1738,7 +1717,7 @@ ParameterHandler::log_parameters_section(LogStream &out,
   boost::property_tree::ptree *current_entries = entries.get();
 
   // Sort parameters alphabetically, if needed.
-  if (sort_alphabetical)
+  if (!(style & Unsorted))
```
nfehn, line 1720: double negation

nfehn:
I have another general question: Where is the correctness check done that excludes invalid parameter combinations?

source/base/parameter_handler.cc
bangerth: You could use `count_set_bits(style & (XML | JSON | Text)) == 1` if you had a function that counts the number of bits set in an integer. This way you make sure that exactly one bit is set.

An implementation of such a function is here: https://www.geeksforgeeks.org/count-set-bits-in-an-integer/

bangerth: Better choice is here: https://stackoverflow.com/questions/109023/how-to-count-the-number-of-set-bits-in-a-32-bit-integer

bangerth: Or, with C++20: https://en.cppreference.com/w/cpp/numeric/popcount

include/deal.II/base/parameter_handler.h
```diff
@@ -851,52 +851,59 @@ class ParameterHandler : public Subscriptor
...
+    Text = 0x0010,
 
     /**
      * Write parameters as a LaTeX table.
      */
-    LaTeX = 2,
+    LaTeX = 0x0020,
+
     /**
      * Write out declared parameters with description and possible values.
```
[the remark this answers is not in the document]
nfehn: sure, but don't you see the difficulties in documentation? Since we don't have a struct design, which would more intuitively show to which part this parameter belongs to, we should document it. My point is that this parameter can be (mis-)understood as the boolean `Short` (or its opposite).

luca-heltai:
@peterrum , as a followup to @nfehn comments, could you please include a better documentation of the `Description` format, i.e., specifying that it is actually a format, and not an additional style (like documentation/alphabetical, etc.)?

I think these need to be addressed, and we can proceeed with merging this PR, also according to the votes in #9888:

- [ ] better documentation of the `Description` format
- [ ] adding some `AssertThrow` at the beginning of the print and log functions, that checks that only one output format is used at a given time
- [ ] Making a `DefaultStyle` that does NOT specify a format, but simply the current default stylistic options (i.e., alphabetical, documented)
- [ ] add `operator|` for `OutputStyle`

nfehn:
I have only two minor comments:

1.

> adding some AssertThrow at the beginning of the print and log functions, that checks that only one output format is used at a given time

It would be good to have these checks at only once place in the code. Why not in the ``operator|`` function? Otherwise, we would actually have to go back to discussion #9888 and rethink certain concepts.

2. Could we please merge #9862 first?

luca-heltai:
I think the interface of #9862 should change back to be compatible with 9.1. I think it is better to do this after we have fixed this one.

luca-heltai:
> Why not in the operator| function?

This is also a good idea, but I would keep the Assert in the code nonetheless.

peterrum:
@luca-heltai I'll tackle this PR this evening. I need to finish one or two things.

nfehn:
> I think the interface of #9862 should change back to be compatible with 9.1. I think it is better to do this after we have fixed this one.

Why? I did not change an interface in #9862. I relied on the interface of deal.II at that time. This pull request here and the related discussions in #9888 emerged later.

luca-heltai:
>  Why? I did not change an interface in #9862. I relied on the interface of deal.II at that time. This pull request here and the related discussions in #9888 emerged later.

which we are going to revert here.

nfehn:
@luca-heltai 

> but I would keep the Assert in the code nonetheless.

Why? Because more asserts help more? Don't you think #9888 would have been the place to discuss this and communicate the downsides of your favorite version? So, with every new function you want to add new asserts to the code, where it is unclear which part of the code is actually responsible for that? There were suggestions in #9888 that do not need this kind of asserts here and there in the code.

tamiko:
Dear all, I would really like to read through the `ParameterHandler` changes without such an aggressive tone everywhere :disappointed:

In particular, we are discussing indredibly minor changes here. So the core principles should be (a) not to break our interfaces from previous releases, and (b) provide a reasonably extensible and intuitive interface.

That said I think it makes a lot of sense to (a) avoid bool, (b) retain and mildly abuse `OutputStyle` as well.

I have only one suggestion, though. I see that we need a number of rather finicky bit comparisons. I suggest to give them a name:
```
// extract information from supplied `OutputStyle`
const bool sort_alphabetically = ...;
```

peterrum:
@luca-heltai May I ask you for a second round of review. In particular, I have updated the documentation and added some asserts at the front of the functions. I think it is fine not to have asserts in `operator|` since during construction of the final configuration it might be valid that a non-valid configuration is passed.

source/base/parameter_handler.cc
```diff
@@ -1219,18 +1219,28 @@ ParameterHandler::set(const std::string &entry_string, const bool new_value)
...
 ParameterHandler::print_parameters(std::ostream &    out,
-                                   const OutputStyle style,
-                                   const bool        sort_alphabetical) const
+                                   const OutputStyle style) const
 {
   AssertThrow(out, ExcIO());
 
+  Assert(
+    (((style & XML) > 0) + ((style & JSON) > 0) + ((style & PRM) > 0) +
+     ((style & Description) > 0) + ((style & LaTeX) > 0)),
```
luca-heltai, line 1228: Isn't a `==1` missing?
```suggestion
    (((style & XML) > 0) + ((style & JSON) > 0) + ((style & PRM) > 0) +
     ((style & Description) > 0) + ((style & LaTeX) > 0)) == 1,
```

source/base/parameter_handler.cc
```diff
@@ -1219,18 +1219,28 @@ ParameterHandler::set(const std::string &entry_string, const bool new_value)
 
 std::ostream &
 ParameterHandler::print_parameters(std::ostream &    out,
-                                   const OutputStyle style,
-                                   const bool        sort_alphabetical) const
+                                   const OutputStyle style) const
 {
   AssertThrow(out, ExcIO());
 
+  Assert(
```
luca-heltai, line 1226: This and the others should be `AssertThrow`. In principle, the output style could be decided at run time, by a parsed string. We want to make sure we assert also in Release mode.

source/base/parameter_handler.cc
```diff
@@ -1329,285 +1339,270 @@ ParameterHandler::recursively_print_parameters(
...
+                      << description << '\n';
+              }
+            else if (description_str.empty() == false)
+              out << "  " << description_str[0] << '\n';
+            else
+              out << '\n';
 
-      default:
-        Assert(false, ExcNotImplemented());
+            // if there is a documenting string, print it as well
```
luca-heltai, line 1595: Can you surround this in a `if(!(style & Short))`? This way we'll have also the `ShortDescription` implemented.

Also add this to the possible combinations, and fix the table at the beginning.

luca-heltai:
Doxygen is complaining for some reasons, but I could not find exactly where:

~~~
$ cat doxygen.log && ! [ -s doxygen.log ]
/home/travis/build/dealii/dealii/include/deal.II/base/parameter_handler.h:954: warning: reached end of file while inside a code block!
The command that should end the block seems to be missing!
The command "cat doxygen.log && ! [ -s doxygen.log ]" exited with 1.
~~~

luca-heltai:
I think it is referring to the constructs `<code>PRM<\code> instead of <code>Text<\code>.`

Can you replace those by their mardkdown counterpart (using backticks)?

source/base/parameter_handler.cc
bangerth: Funny question that I'm not sure about: Say, `OutputStyle` uses `short unsigned int` as base type. The `style & XML` is promoted to `int`. Is the result guaranteed to be `>0` is one bit is set, or could this result in the sign bit being set and the result is negative? I think what you *want* to test is that at least one bit is set, so maybe using `!= 0` is the better check over `>0`?

source/base/parameter_handler.cc
```diff
@@ -1219,18 +1219,28 @@ ParameterHandler::set(const std::string &entry_string, const bool new_value)
...
+  Assert(
+    (((style & XML) > 0) + ((style & JSON) > 0) + ((style & PRM) > 0) +
+     ((style & Description) > 0) + ((style & LaTeX) > 0)),
+    ExcMessage(
+      "You have chosen either no or multiple style formats. You can choose between: PRM, Description, LaTeX, XML, JSON."));
+
+  Assert(((style & LaTeX) && (style & Short)) == false,
+         ExcMessage("It is not possible to request short LaTex."));
+  Assert(((style & Description) && (style & Short)) == false,
+         ExcMessage("It is not possible to request short Description."));
```
nfehn, line 1235: Personally, I would prefer to have all these asserts in one private member function that ensures consistency of the output style, as a means to satisfy our design goals from #9888 and to highlight that this check is related to the data structure OutputStyle and not to a specific function? If I understood it correctly, more functions might need these checks, so a copy-paste of this code is forseeable otherwise.

peterrum:
@luca-heltai  That should work. Locally it looks like this:

OutputStyle

I have applied all other suggestions.

source/base/parameter_handler.cc
luca-heltai: That’s actually a very nice suggestion. @peterrum, could you wrap up the checks in a function returning a Boolean, and assert on the returned value?

luca-heltai:
@peterrum, would you mind implementing the other suggestion by @nfehn, i.e., wrap up the check of the consistency in a private function that returns a boolean, and assert on the return value instead of explicitly writing the  checks in the function? This will be useful also elsewhere.

bangerth:
Or a free function in an anonymous namespace in the .cc file if it doesn't require access to member variables.

source/base/parameter_handler.cc
peterrum: I went with @bangerth' suggestion with the free function in an anonymous namespace.

include/deal.II/base/parameter_handler.h
```diff
@@ -846,57 +846,113 @@ class ParameterHandler : public Subscriptor
...
    */
   enum OutputStyle
   {
     /**
-     * Write human readable output suitable to be read by ParameterHandler
-     * again.
+     * Default stylistic style: print documentation and sort all parameters
+     * alphabetically.
+     */
+    DefaultStyle = 0x0000,
```
bangerth, line 876: I regret that we hadn't settled on `use_underscore` instead of `CamelCase` for member variables and enum values back when this class was written :-(

But it's too late now. It makes more sense to keep the members of this class consistent with each other than to have some kind of mixture.

peterrum: @bangerth We can still deprecate all old names. In that case, the question would be how to call `Short` since `short` is already taken.

luca-heltai: Let's discuss this in another PR...

bangerth: Yes. I don't think there is much of an upside to making such a change. It is what it is.

nfehn:
Did we forget to update the Assert in the second half of ``ParameterAcceptor::initialize()`` (if-branch ``if (!output_filename.empty()){...}`` in this PR?

https://www.dealii.org/developer/doxygen/deal.II/parameter__acceptor_8cc_source.html

PR #9920, opened 2020-04-17
Remove doc/doxygen/tutorial

masterleinad:
In one of my later pull requests removing deprecated functions, @bangerth requested to remove all the files in this directory and I agree. The code in there doesn't really explain a lot and is not representative for the library as a whole.


masterleinad:
Oh, yes I see. They are referenced in `include/deal.II/algorithms/theta_timestepping.h` and `include/deal.II/lac/block_sparsity_pattern.h` respectively with `doxygen` pages https://www.dealii.org/developer/doxygen/deal.II/classAlgorithms_1_1ThetaTimestepping.html and https://www.dealii.org/developer/doxygen/deal.II/classBlockDynamicSparsityPattern.html (this one not mentioning the corresponding file, though). You learn something new about the library every day!

I guess it's useful to have them as compilable programs somewhere then even if we don't use this style anywhere else.

bangerth:
We have lots of code snippets in the documentation, both in the library itself as well as in the introductions and results sections of the tutorial program. For all of these, it would *in principle* be nice if we could verify that they compiled correctly, but for 95% of them we don't and it doesn't seem to have been a large problem in the past. So my preference would be to just take the snippets referenced in the documentation that come out of `examples/doxygen` and inline them into the places where they are currently used, then nuke the directory.

PR #9922, opened 2020-04-18
Build and check documentation in CI

masterleinad:
This pull request makes the `Travis` CI runner also build the documentation and checks that there are no warnings. After #9880, I get https://travis-ci.org/github/masterleinad/dealii/builds/676604234.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,24 @@ cache:
   directories:
     - programs
 
+addons:
+  apt:
+    packages:
+      - doxygen
+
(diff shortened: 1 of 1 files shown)


masterleinad:
I rebased and it seems I could keep my promise.

masterleinad:
If we want to go this way, I would appreciate moving forward here so we don't accumulate more warnings in the meantime that would break `Travis` once this is merged effectively stalling all other pull requests.

PR #9925, opened 2020-04-18
Instantiate AffineConstraints::distribute_local_to_global

peterrum:
for `TrilinosWrappers::SparseMatrix` and `LinearAlgebra::distributed::Vector`.


drwells:
This is probably fine but could we modify a trilinos test to also use our native vector class?

peterrum:
@drwells I have updated a test so that it uses the function.

PR #9926, opened 2020-04-18
Remove most <> from tutorials

peterrum:
Or is there an educational purpose in keeping these?

Currently, there are lot of places where s.th. like this is written:
```cpp
SolverRichardson<> solver(solver_control);
```
For the tutorials, I feel it would have more benefit to be explicit about the type:
```cpp
SolverRichardson<Vector<double>> solver(solver_control);
```
or simply write (as proposed by this PR):
```cpp
SolverRichardson solver(solver_control);
```

### DIFF
--- a/examples/step-11/step-11.cc
+++ b/examples/step-11/step-11.cc
@@ -388,10 +388,10 @@ namespace Step11
   template <int dim>
   void LaplaceProblem<dim>::solve()
   {
-    SolverControl solver_control(1000, 1e-12);
-    SolverCG<>    cg(solver_control);
+    SolverControl            solver_control(1000, 1e-12);
+    SolverCG<Vector<double>> cg(solver_control);
 
-    PreconditionSSOR<> preconditioner;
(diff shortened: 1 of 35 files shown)


masterleinad:
Users often start off from one of the example steps and I think they might face another unexpected problem if the default template argument is used in the tutorial but they changed the vector type.
Hence, I would prefer
```
SolverRichardson<Vector<double>> solver(solver_control);
```
over
```
SolverRichardson solver(solver_control);
```
(admitting that the current syntax looks more like the latter version).

bangerth:
I think that it is this C++17 extension: https://en.cppreference.com/w/cpp/language/class_template_argument_deduction

masterleinad:
Yes, https://godbolt.org/z/uwiZos.

masterleinad:
Admittedly, I wasn't so sure if `CTAD` applied here after seeing that all of the `github` CI and `appveyor` succeded. It seems that they all support C++17.

peterrum:
Thanks for the info! I have made the change!

PR #9929, opened 2020-04-21
Split VectorTools up

masterleinad:
Related to #9790. This also allows to only include smaller headers than all of `VectorTools` but I didn't make that transition for the examples or the tests.

The approach taken here, was to first split up the `template` file using the instantiations files and then splitting `vector_tools.h` matching the template files. If we don't want to split up `vector_tools.h` it's easy enough to revert that.

The actual changes are pretty small, it's mostly only moving code around.
Some things that changed:
- using more forward declarations
- some more explicit instantiations for `dim<spacedim` for the boundary functions. We were previously generating them implicitly through the `project` functions.

I verified manually that the documentation didn't change at first sight.


masterleinad:
Yes, the commit history is not super clean. I can try to get every commit compiling but that's quite some more work. Otherwise, I would just squash into a few commits.

tamiko:
Squashing at the end and having the full CI pass is all we need :-)

bangerth:
Have you measured what this means for the overall CPU time? Many of us have many-core machines where this kind of change makes sense if you do `-jmany`. But what happens if you do `-j1` or just `make`? Does this lead to appreciably longer compile times? Should we care about this case?

masterleinad:
I didn't change the targets.

bangerth:
Oh, I see -- I hadn't paid enough attention to the fact that you're not splitting up which functions are compiled in which .cc file.

masterleinad:
Squashed.

PR #9934, opened 2020-04-21
Explicitly =delete copy operators for TaskDescriptor.

bangerth:
These operators previously already threw exceptions unconditionally. We might as well disallow using them outright.

/rebuild

### DIFF
--- a/include/deal.II/base/thread_management.h
+++ b/include/deal.II/base/thread_management.h
@@ -1294,10 +1294,10 @@ namespace Threads
       TaskDescriptor();
 
       /**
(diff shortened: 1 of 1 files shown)


bangerth:
The failing OSX serial check has a download problem and is unrelated.

masterleinad:
Restarted... Even if `GitHub` is not working fully again, we should have at least some of the CI running on this successfully.

masterleinad:
Can you please rebase? That seems to be the most reliable way to get the `jenkins` testers runinng again.

bangerth:
Rebased and works now!

PR #9935, opened 2020-04-21
Remove some pthread vestiges that were deprecated a while ago already

bangerth:
/rebuild


masterleinad:
Didn't we just deprecate that class for `9.1.0`? Is it too early to remove it before the next release? So far, I only removed stuff from `9.0`.

bangerth:
Oh, I didn't pay attention to be honest. I was in that file anyway and fixing stuff so I thought I'd remove it. You're right that this was deprecated in #7244, 19 months ago. Let's postpone till post 9.2 then.

masterleinad:
```
/home/runner/work/dealii/dealii/source/base/thread_management.cc:155:3: error: ‘PosixThreadBarrier’ does not name a type
```

masterleinad:
Can you please rebase?

bangerth:
Rebased.

PR #9936, opened 2020-04-22
Improve calculating boolean values

rezarastak:
I attempted to improve the way some of the Boolean variables are assigned in the code. The goal is to:
* Simplify code flow.
* Make bool variables const.
* Make use of standard algorithms.
* Prevent mistakes.
* Improve performance (in some limited cases).
* Improve readability, although this is subjective.

Some of the standard algorithms, even though they are very expressive and efficient, might become too long because we can't use `auto` parameters in lambdas (with c++11), and we have to write a lot of `begin()` and `end()`. I still preferred them in many cases over `for` loops with `break` statements.

### DIFF
--- a/include/deal.II/hp/fe_collection.h
+++ b/include/deal.II/hp/fe_collection.h
@@ -1006,13 +1006,12 @@ namespace hp
   FECollection<dim, spacedim>::hp_constraints_are_implemented() const
   {
     Assert(finite_elements.size() > 0, ExcNoFiniteElements());
-
-    bool hp_constraints = true;
-    for (unsigned int i = 0; i < finite_elements.size(); ++i)
-      hp_constraints =
-        hp_constraints && finite_elements[i]->hp_constraints_are_implemented();
-
-    return hp_constraints;
+    return std::all_of(
+      finite_elements.cbegin(),
+      finite_elements.cend(),
+      [](const std::shared_ptr<const FiniteElement<dim, spacedim>> &fe) {
+        return fe->hp_constraints_are_implemented();
+      });
   }
 
 
--- a/include/deal.II/lac/affine_constraints.templates.h
+++ b/include/deal.II/lac/affine_constraints.templates.h
@@ -4013,23 +4013,21 @@ AffineConstraints<number>::add_entries_local_to_global(
   Assert(sparsity_pattern.n_rows() == sparsity_pattern.n_cols(),
          ExcNotQuadratic());
 
-  const size_type n_local_dofs       = local_dof_indices.size();
-  bool            dof_mask_is_active = false;
-  if (dof_mask.n_rows() == n_local_dofs)
-    {
-      dof_mask_is_active = true;
-      AssertDimension(dof_mask.n_cols(), n_local_dofs);
-    }
-
+  const size_type n_local_dofs = local_dof_indices.size();
   typename internals::AffineConstraintsData<number>::ScratchDataAccessor
     scratch_data;
 
-  // if the dof mask is not active, all we have to do is to add some indices
-  // in a matrix format. To do this, we first create an array of all the
-  // indices that are to be added. these indices are the local dof indices
-  // plus some indices that come from constraints.
-  if (dof_mask_is_active == false)
+  const bool dof_mask_is_active = (dof_mask.n_rows() == n_local_dofs);
+  if (dof_mask_is_active == true)
     {
+      AssertDimension(dof_mask.n_cols(), n_local_dofs);
+    }
+  else
+    {
+      // if the dof mask is not active, all we have to do is to add some indices
+      // in a matrix format. To do this, we first create an array of all the
+      // indices that are to be added. these indices are the local dof indices
+      // plus some indices that come from constraints.
       std::vector<size_type> &actual_dof_indices = scratch_data->columns;
       actual_dof_indices.resize(n_local_dofs);
       make_sorted_row_list(local_dof_indices, actual_dof_indices);
@@ -4107,11 +4105,8 @@ AffineConstraints<number>::add_entries_local_to_global(
   const bool                    keep_constrained_entries,
   const Table<2, bool> &        dof_mask) const
 {
-  const size_type n_local_rows       = row_indices.size();
-  const size_type n_local_cols       = col_indices.size();
-  bool            dof_mask_is_active = false;
-  if (dof_mask.n_rows() == n_local_rows && dof_mask.n_cols() == n_local_cols)
-    dof_mask_is_active = true;
+  const size_type n_local_rows = row_indices.size();
+  const size_type n_local_cols = col_indices.size();
 
   // if constrained entries should be kept, need to add rows and columns of
   // those to the sparsity pattern
@@ -4131,6 +4126,8 @@ AffineConstraints<number>::add_entries_local_to_global(
   // in a matrix format. To do this, we first create an array of all the
   // indices that are to be added. these indices are the local dof indices
   // plus some indices that come from constraints.
+  const bool dof_mask_is_active =
+    dof_mask.n_rows() == n_local_rows && dof_mask.n_cols() == n_local_cols;
   if (dof_mask_is_active == false)
     {
       std::vector<size_type> actual_row_indices(n_local_rows);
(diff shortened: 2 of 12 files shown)


masterleinad:
```
D:\a\dealii\dealii\source\base\quadrature_lib.cc(616,1): error C3493: 'eps' cannot be implicitly captured because no default capture mode has been specified
```

include/deal.II/hp/fe_collection.h
```diff
@@ -1006,13 +1006,11 @@ namespace hp
...
-
-    bool hp_constraints = true;
-    for (unsigned int i = 0; i < finite_elements.size(); ++i)
-      hp_constraints =
-        hp_constraints && finite_elements[i]->hp_constraints_are_implemented();
-
-    return hp_constraints;
+    return std::all_of(finite_elements.cbegin(),
+                       finite_elements.cend(),
+                       [](decltype(*finite_elements.cbegin()) fe) {
```
bangerth, line 1011: we need C++14 template lambdas! :-)

include/deal.II/hp/fe_collection.h
```diff
@@ -1006,13 +1006,11 @@ namespace hp
...
   {
     Assert(finite_elements.size() > 0, ExcNoFiniteElements());
-
-    bool hp_constraints = true;
-    for (unsigned int i = 0; i < finite_elements.size(); ++i)
-      hp_constraints =
-        hp_constraints && finite_elements[i]->hp_constraints_are_implemented();
-
-    return hp_constraints;
+    return std::all_of(finite_elements.cbegin(),
```
bangerth, line 1009: nice job -- this also leads to the early exit if one FE fails the test, which the programmer of that code missed!

include/deal.II/hp/fe_collection.h
bangerth: But you can make it more readable this way, because I'm pretty sure that the iterators should be convertible:
```suggestion
                       [](const std::shared_ptr<FiniteElement<dim,spacedim>> &fe) {
```

include/deal.II/lac/affine_constraints.templates.h
```diff
@@ -4013,23 +4013,21 @@ AffineConstraints<number>::add_entries_local_to_global(
...
+  const size_type n_local_dofs = local_dof_indices.size();
   typename internals::AffineConstraintsData<number>::ScratchDataAccessor
     scratch_data;
 
-  // if the dof mask is not active, all we have to do is to add some indices
-  // in a matrix format. To do this, we first create an array of all the
-  // indices that are to be added. these indices are the local dof indices
-  // plus some indices that come from constraints.
-  if (dof_mask_is_active == false)
+  const bool dof_mask_is_active = dof_mask.n_rows() == n_local_dofs;
```
bangerth, line 4020:
```suggestion
  const bool dof_mask_is_active = (dof_mask.n_rows() == n_local_dofs);
```

include/deal.II/lac/affine_constraints.templates.h
```diff
@@ -4176,14 +4173,12 @@ AffineConstraints<number>::add_entries_local_to_global(
   typename internals::AffineConstraintsData<number>::ScratchDataAccessor
     scratch_data;
 
-  bool dof_mask_is_active = false;
-  if (dof_mask.n_rows() == n_local_dofs)
+  const bool dof_mask_is_active = dof_mask.n_rows() == n_local_dofs;
```
bangerth, line 4176:
```suggestion
  const bool dof_mask_is_active = (dof_mask.n_rows() == n_local_dofs);
```

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -227,15 +227,13 @@ namespace internal
...
-              {
-                duplicate = true;
-                break;
-              }
-
-          if (duplicate == false)
+          const bool is_duplicate =
+            std::any_of(finite_elements.cbegin(),
+                        finite_elements.cbegin() + dataset,
+                        [&](decltype(*finite_elements.cbegin()) fe) {
```
bangerth, line 233:
```suggestion
                        [&](const std::shared_ptr<dealii::hp::FECollection<dim, spacedim>> &fe) {
```

source/base/quadrature_lib.cc
```diff
@@ -608,21 +608,22 @@ template <>
...
-      on_edge = true;
-  if (on_edge &&
-      (std::abs((singularity - Point<2>(.5, .5)).norm_square() - .5) < eps))
-    on_vertex = true;
+  const double eps = 1e-8;
+  const bool   on_edge =
+    std::any_of(singularity.begin_raw(),
+                singularity.end_raw(),
+                [eps](double coord) {
+                  return std::abs(coord) < eps || std::abs(coord) < eps;
```
bangerth, line 616: I read this several times now -- but I think the condition is `X || X`, right? I think you need this:
```suggestion
                  return std::abs(coord) < eps || std::abs(coord-1) < eps;
```

source/dofs/dof_tools.cc
```diff
@@ -1733,19 +1733,13 @@ namespace DoFTools
...
-            break;
-          }
-      Assert(found == true,
+      const bool cell_found = std::any_of(
+        dof_handler.begin_active(),
+        typename DoFHandlerType::active_cell_iterator{dof_handler.end()},
+        [subdomain](const typename DoFHandlerType::cell_accessor &cell) {
+          return cell.subdomain_id() == subdomain;
+        });
+      Assert(cell_found == true,
```
bangerth, line 1742: Just inline the `std::any_of` into the assertion here, so that it doesn't have to run in release mode.

source/multigrid/mg_transfer_internal.cc
```diff
@@ -764,16 +764,15 @@ namespace internal
               if (!cell->has_children())
                 continue;
 
-              bool consider_cell = false;
-              if (tria.locally_owned_subdomain() ==
-                    numbers::invalid_subdomain_id ||
-                  cell->level_subdomain_id() == tria.locally_owned_subdomain())
-                consider_cell = true;
+              bool consider_cell =
+                tria.locally_owned_subdomain() ==
```
bangerth, line 768:
```suggestion
                (tria.locally_owned_subdomain() ==
```

source/multigrid/mg_transfer_internal.cc
```diff
@@ -764,16 +764,15 @@ namespace internal
...
+                  numbers::invalid_subdomain_id ||
+                cell->level_subdomain_id() == tria.locally_owned_subdomain();
```
bangerth, line 770:
```suggestion
                cell->level_subdomain_id() == tria.locally_owned_subdomain());
```

bangerth:
Nice work. How did you find all of these places?

include/deal.II/hp/fe_collection.h
masterleinad: > we need C++14 template lambdas! :-)

You mean we should move to C++14 soonish (like after the release)? :wink:

rezarastak: `const auto &fe` would've been great for here, but with `decltype` I tried to write it in the most generic way. My thinking was that if we change the exact type signature of `finite_elements` (for example from `vector<shared_ptr>` to `vector<unique_ptr>`) the lambda should work properly.

I understand that using the full type name in the argument list is probably the most readable form. So I will apply @bangerth's suggestion.

rezarastak:
> How did you find all of these places?

I searched lines similar to
```c++
bool var_name = false;
```
or
```c++
bool var_name = true;
```
But most of them had a very complicated logic and I left them as is. Only a small percentage of places could be improved.

include/deal.II/hp/fe_collection.h
bangerth: > You mean we should move to C++14 soonish (like after the release)?

Well well, now, let's not get crazy! ;-))

bangerth: Thanks -- we've generally tried to be explicit about types where possible. It's just one way to make it easier for readers to reason about the correctness of code. (Within limits where the type is obvious -- e.g., in `for (const auto &cell : ...)`.)

include/deal.II/numerics/data_out_dof_data.templates.h
rezarastak: Is there a specific reason for typing `dealii::hp::FECollection` instead of `hp::FECollection`?

source/base/quadrature_lib.cc
rezarastak: Thank you for catching this.

source/dofs/dof_tools.cc
rezarastak: Good idea! then we can remove the `#ifdef DEBUG` guard here.

bangerth: Nice -- I missed that!

include/deal.II/numerics/data_out_dof_data.templates.h
bangerth: No reason -- just a copy-paste issue from when I looked up the declaration of the `finite_elements` variable. Go with the shorter version.

rezarastak: Compiler is telling me that `dealii::` is mandatory because of name collision with `dealii::internal::hp::FECollection`.

rezarastak:
I applied all of the fixes that were suggested. It is ready to be reviewed again.

PR #9938, opened 2020-04-22
const is removed from private member variables of DiscreteTime

rezarastak:
I had made a small mistake when I introduced `DiscreteTime`. Making some private member variables `const` prevented the implicit `operator=()` created by the compiler.

I removed the `const` now. These three variables still cannot be modified with the public interface of the class.

### DIFF
--- a/include/deal.II/base/discrete_time.h
+++ b/include/deal.II/base/discrete_time.h
@@ -365,12 +365,12 @@ private:
   /**
    * The beginning of the time interval.
    */
-  const double start_time;
+  double start_time;
 
   /**
    *The end of the time interval.
    */
-  const double end_time;
+  double end_time;
 
   /**
    * The current time.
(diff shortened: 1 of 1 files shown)


masterleinad:
Does it make sense for this class to be copied?

rezarastak:
Currently I don't have a use-case for copying it, but the copy operation isn't expensive because it is trivially copiable with 5 doubles. I don't feel we should restrict it from being copied.

If in the future we want to configure this object with a `ParameterHandler` class by defining a `parse_parameters(prm)` method, these member variables must be non-const. This is another reason for proposing this PR.

masterleinad:
The question for me was not about the expense of copying an object of this type but rather if it makes sense to copy it and if the default copy constructor does what we want.
Of course, this is slightly outside of the scope of this pull request but it would be good to think about it and then either express intent that the default copy constructor is good enough by defaulting it explicitly, deleting it or implementing it differently.

rezarastak:
Hopefully as I incorporate this class into some of the implementations and tutorials in dealii, we will better learn if copying it makes sense or we should disallow it. I will think more about it.

Thanks

PR #9944, opened 2020-04-23
Create GridTools::compute_vertices_with_ghost_neighbors()

peterrum:
This PR is a follow-up to #8783.

This PR:
1) removes the unused function `internal::p4est::compute_vertices_with_ghost_neighbors()`
2) depreciates `parallel::TriangulationBase::compute_vertices_with_ghost_neighbors()` and
3)  adds a new function `GridTools::compute_vertices_with_ghost_neighbors()`.

The reason for moving `parallel::TriangulationBase::compute_vertices_with_ghost_neighbors()` to `GridTools::compute_vertices_with_ghost_neighbors()` is that the function is general (independent of the triangulation type) and only needs public methods of the triangulation (like the iterators).

If this change is fine, I would modify my old change-log entry accordingly.

### DIFF
--- a/include/deal.II/distributed/p4est_wrappers.h
+++ b/include/deal.II/distributed/p4est_wrappers.h
@@ -542,18 +542,6 @@ namespace internal
     tree_exists_locally(const typename types<dim>::forest *parallel_forest,
                         const typename types<dim>::topidx  coarse_grid_cell);
 
-
-
-    /**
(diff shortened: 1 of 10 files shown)


kronbichler:
Is the testsuite fine with this PR?

peterrum:
@kronbichler All tests have passed.

kronbichler:
Great! But let's wait a day or two if someone else has a comment on this.

PR #9945, opened 2020-04-23
Remove parallel::TriangulationBase::compute_n_locally_owned_active_cellls_per_processor

peterrum:
In 5eb3561664a032f9f9723d96fad06711e8f955fc, we have replaced the method `n_locally_owned_active_cells_per_processor()` by `compute_n_locally_owned_active_cells_per_processor()` since we did not want to store the information in the `NumberCache` anymore (since it is too expensive for large simulations). I think it would be an option to remove the new function since the same effect can be reached in the user code by the following one liner:
```cpp
Utilities::MPI::all_gather(tr.get_communicator(), tr.n_locally_owned_active_cells());
```

I am not sure related backwards compatibility. But, I guess the change in 5eb3561664a032f9f9723d96fad06711e8f955fc was not backwards compatible either.

### DIFF
--- a/examples/step-55/step-55.cc
+++ b/examples/step-55/step-55.cc
@@ -456,7 +456,8 @@ namespace Step55
         dof_handler, coupling, dsp, constraints, false);
       SparsityTools::distribute_sparsity_pattern(
         dsp,
-        dof_handler.compute_locally_owned_dofs_per_processor(),
+        Utilities::MPI::all_gather(mpi_communicator,
+                                   dof_handler.locally_owned_dofs()),
         mpi_communicator,
         locally_relevant_dofs);
       preconditioner_matrix.reinit(owned_partitioning,
--- a/include/deal.II/dofs/dof_handler.h
+++ b/include/deal.II/dofs/dof_handler.h
@@ -1015,77 +1015,16 @@ public:
   locally_owned_mg_dofs(const unsigned int level) const;
 
   /**
-   * Compute a vector with the locally owned DoFs of each processor.
-   *
-   * This function involves global communication via the @p MPI_Allgather
-   * function, so it must be called on all processors participating in the MPI
-   * communicator underlying the triangulation.
-   *
-   * If you are only interested in the number of elements each processor owns
-   * then compute_n_locally_owned_dofs_per_processor() is a better choice.
-   *
-   * If this is a sequential DoFHandler, then the vector has a single element
-   * that equals the IndexSet representing the entire range [0,n_dofs()]. (Here,
-   * "sequential" means that either the whole program does not use MPI, or that
-   * it uses MPI but only uses a single MPI process, or that there are multiple
-   * MPI processes but the Triangulation on which this DoFHandler builds works
-   * only on one MPI process.)
-   */
-  std::vector<IndexSet>
-  compute_locally_owned_dofs_per_processor() const;
-
-  /**
-   * Compute a vector with the number of degrees of freedom each
-   * processor that participates in this triangulation owns locally. The sum
-   * of all these numbers equals the number of degrees of freedom that exist
-   * globally, i.e. what n_dofs() returns.
-   *
-   * This function involves global communication via the @p MPI_Allgather
-   * function, so it must be called on all processors participating in the MPI
-   * communicator underlying the triangulation.
-   *
-   * Each element of the vector returned by this function equals the number of
-   * elements of the corresponding sets returned by
-   * compute_locally_owned_dofs_per_processor().
-   *
-   * If this is a sequential DoFHandler, then the vector has a single element
-   * equal to n_dofs(). (Here, "sequential" means that either the whole program
-   * does not use MPI, or that it uses MPI but only uses a single MPI process,
-   * or that there are multiple MPI processes but the Triangulation on which
-   * this DoFHandler builds works only on one MPI process.)
-   */
-  std::vector<types::global_dof_index>
-  compute_n_locally_owned_dofs_per_processor() const;
-
-  /**
-   * Compute a vector with the locally owned DoFs of each processor on
-   * the given level @p level for geometric multigrid.
-   *
-   * This function involves global communication via the @p MPI_Allgather
-   * function, so it must be called on all processors participating in the MPI
-   * communicator underlying the triangulation.
-   *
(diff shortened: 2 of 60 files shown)


kronbichler:
This touches the bigger question of the same function in the DoFHandler:
https://github.com/dealii/dealii/blob/2ba7b4de3bbe51c865d5dd28da019c5d65508f99/include/deal.II/dofs/dof_handler.h#L1057-L1058
This was introduced before in #8298, so we have not released that part either and we could replace it as well. The only nuisance is the other function
https://github.com/dealii/dealii/blob/2ba7b4de3bbe51c865d5dd28da019c5d65508f99/include/deal.II/dofs/dof_handler.h#L1034-L1035
where we get an `IndexSet` and I am not sure whether `all_gather` works. In case it does, we should remove this lengthy implementation
https://github.com/dealii/dealii/blob/2ba7b4de3bbe51c865d5dd28da019c5d65508f99/source/dofs/number_cache.cc#L121-L229
and just kick out the functions altogether.

peterrum:
@kronbichler After a quick look at the implementation of `Utilities::MPI::all_gather()`, I am quite sure it should  work also for `IndexSet`: it is working via Boost's serialization and can handle variable size quantities. 

See: https://github.com/dealii/dealii/blob/fc54b91f5b3260c9204b8815f41785a7c22056c4/include/deal.II/base/mpi.h#L1199-L1253

peterrum:
Should I include `compute_n_locally_owned_dofs_per_processor()` and `compute_locally_owned_dofs_per_processor()` in this PR?

kronbichler:
They are all of the same kind, so I would say we either remove all or we remove none.

peterrum:
@kronbichler Do you think like this c41b6352f3da058d75f527932da07d8b3f2f7847? Currently 57 tests fail (these use once of the `compute_`-functions). Before I tackle these and modify the `hp::DoFHandler` in the same way, I would like to know if this is what you had in mind?

include/deal.II/dofs/dof_handler.h
```diff
@@ -1508,10 +1447,20 @@ DoFHandler<dim, spacedim>::n_locally_owned_dofs_per_processor() const
...
+      else
+        comm = MPI_COMM_SELF;
+
       const_cast<dealii::internal::DoFHandlerImplementation::NumberCache &>(
         number_cache)
         .n_locally_owned_dofs_per_processor =
-        compute_n_locally_owned_dofs_per_processor();
+        number_cache.get_n_locally_owned_dofs_per_processor(comm);
     }
   return number_cache.n_locally_owned_dofs_per_processor;
```
kronbichler, line 1465: For this implementation, can't we bypass this code and simply return `Utilities::MPI::all_gather(...)`?

examples/step-55/step-55.cc
```diff
@@ -456,7 +456,8 @@ namespace Step55
         dof_handler, coupling, dsp, constraints, false);
       SparsityTools::distribute_sparsity_pattern(
         dsp,
-        dof_handler.compute_locally_owned_dofs_per_processor(),
+        Utilities::MPI::all_gather(mpi_communicator,
+                                   dof_handler.locally_owned_dofs()),
```
kronbichler, line 460: This is unrelated to this patch, but I believe we should replace this argument by simply `dof_handler.locally_owned_dofs()` - I must have forgotten those in #9710.

include/deal.II/dofs/dof_handler.h
```diff
@@ -1525,10 +1474,20 @@ DoFHandler<dim, spacedim>::locally_owned_dofs_per_processor() const
...
+      if (tr != nullptr)
+        comm = tr->get_communicator();
+      else
+        comm = MPI_COMM_SELF;
+
       const_cast<dealii::internal::DoFHandlerImplementation::NumberCache &>(
         number_cache)
         .locally_owned_dofs_per_processor =
-        compute_locally_owned_dofs_per_processor();
+        number_cache.get_locally_owned_dofs_per_processor(comm);
```
kronbichler, line 1490: Same here.

include/deal.II/dofs/dof_handler.h
```diff
@@ -1550,73 +1509,26 @@ DoFHandler<dim, spacedim>::locally_owned_mg_dofs_per_processor(
...
+      else
+        comm = MPI_COMM_SELF;
+
       const_cast<dealii::internal::DoFHandlerImplementation::NumberCache &>(
         mg_number_cache[level])
         .locally_owned_dofs_per_processor =
-        compute_locally_owned_mg_dofs_per_processor(level);
+        mg_number_cache[level].get_locally_owned_dofs_per_processor(comm);
     }
   return mg_number_cache[level].locally_owned_dofs_per_processor;
```
kronbichler, line 1527: and here

source/dofs/number_cache.cc
```diff
@@ -136,95 +125,8 @@ namespace internal
...
-                    std::istringstream              in(decompressed_buffer);
-                    boost::archive::binary_iarchive archive(in);
-
-                    archive >> locally_owned_dofs_per_processor[i];
-                  }
-            }
-#endif
-          return locally_owned_dofs_per_processor;
+          return Utilities::MPI::all_gather(mpi_communicator,
+                                            locally_owned_dofs);
```
kronbichler, line 129: here you use this function, but I think we can remove the function altogether - this class is mostly internal anyway, I think.

peterrum:
@kronbichler Thanks for the fast feedback. 

> For this implementation, can't we bypass this code and simply return Utilities::MPI::all_gather(...)?

Remember back why this functions caused us headaches half a year ago: reason 1) memory consumption scales with the number processes and reason 2) the function does not return a vector but a reference to a vector. Latter was the reason why we fill the vector only when needed.

I would remove the vector once the deprecated functions are deleted. I.e in two weeks?

kronbichler:
Sure, the vector can only be removed once we have eliminated the deprecated function. But the number cache can be cleaned up already in terms of the unnecessary function - but I agree on a second sight that it is probably no big difference since the vector is in `NumberCache`.

masterleinad:
```
/home/runner/work/dealii/dealii/source/dofs/number_cache.cc:90:26: error: unused variable ‘n_procs’ [-Werror=unused-variable]
       const unsigned int n_procs =
                          ^~~~~~~
```

PR #9948, opened 2020-04-24
Make coarse_cell_*_to_coarse_cell_* private

peterrum:
Currently, the functions `coarse_cell_id_to_coarse_cell_index` and `coarse_cell_index_to_coarse_cell_id` are private in `p:d:T` and public `Tria` and `p:f:T`. I would propose to make these functions private (since I don't think the users need to use them) and make `CellId` a friend of `Tria`.

Any concerns with this change? Am I missing where these functions might be useful for users? These functions are new.

### DIFF
--- a/include/deal.II/distributed/fully_distributed_tria.h
+++ b/include/deal.II/distributed/fully_distributed_tria.h
@@ -233,6 +233,7 @@ namespace parallel
       virtual bool
       is_multilevel_hierarchy_constructed() const override;
 
+    private:
       virtual unsigned int
       coarse_cell_id_to_coarse_cell_index(
         const types::coarse_cell_id coarse_cell_id) const override;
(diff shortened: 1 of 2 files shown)


tjhei:
> Any concerns with this change?

I am not exactly concerned about it but I don't see any harm in keeping them public. Somebody might want to do query cells without bothering to create/edit ``CellId``s?

peterrum:
@tjhei I would be fine to make the functions public once someone actually needs it. Until then, I would propose to make the functions private so that we could make (potential) modifications without any headaches. 

Independent of the decision, the access to the functions should have the same accessibility in `parallel::Triangulation` and in the derived classes.

tjhei:
> Independent of the decision, the access to the functions should have the same accessibility in `parallel::Triangulation` and in the derived classes.

I agree.

PR #9949, opened 2020-04-24
Make parallel::shared::Triangulation::update_number_cache private

peterrum:
This function is private in `p:d:t` and `p:f:t`, however, it is protected in `p:s:t`. I would propose to make all private so that they are consistent.


bangerth:
I'm pretty sure this is an oversight...

PR #9950, opened 2020-04-24
Add ParameterHandler::print_parameters with std::string as input.

luca-heltai:
Following the ideas of #9862, we add the possibility to write parameters on files based on their extension.


include/deal.II/base/parameter_acceptor.h
```diff
@@ -373,29 +373,40 @@ class ParameterAcceptor : public Subscriptor
...
+   * supported formats for the @p output_filename.
    *
    * If the input file does not exist, a default one with the same name is
-   * created for you, and an exception is thrown.
+   * created for you following the style specified in
+   * @p output_style_for_filename, and an exception is thrown.
+   *
+   * By default, the file format used to write the files is deduced from
+   * the extension of the file names, ignoring the corresponding
+   * ParameterHandler::OutputStyle specification. The corresponding
```
bangerth, line 389: That's confusing. Why not:
* If `output_style` doesn't have any of the file-type bits set, use the output format that corresponds to the extension only 
* Else, if the file extension is recognized, assert that it matches the bit in `output_style`
* Else, use the bit from the `output_style`
?

luca-heltai:
I'm not sure why, but the web page does not show the correct diff. This is now the documentation:

~~~
   * By default, the file format used to write the files is deduced from
   * the extension of the file names. If the corresponding
   * ParameterHandler::OutputStyle specifies a format specification, this must
   * be compatible with the file extension, or an exception will be thrown.
   *
   * If the extension is not recognized, and you do not specify a format in the
   * corresponding ParameterHandler::OutputStyle, an assertion is thrown.
~~~

luca-heltai:
In a discussion with @peterrum , we concluded that the current master has a bug with the handling of `Short` formats that are not `XML` or `JSON`. 

The last commit of this PR fixes that bug, and some related output files on my PC.

luca-heltai:
Now all possible combinations of `Short` and other formats are supported, so I simply removed the table with the possible combinations, since all of them are now supported.

luca-heltai:
For some reasons, my exception in `parameter_handler_25` looks different here. I added a variant of the output.

luca-heltai:
@peterrum anything I should address?

luca-heltai:
I'll let you do the honors. :)

nfehn:
I wondered why this file

tests/parameter_handler/parameter_handler_25.output.std

is listed under "Files changed". Has this file been added intentionally?

luca-heltai:
Yes, this is intentional.

On my mac (I think it this is due to clang), the exception is thrown with `(const std::string)` instead of `(const string)`, and therefore the test would fail. The testsuite knows about situations where sometimes compilers behave differently, and therefore allows you to specify a different variant of the output file (adding an additional extension after `.output`, as in this case: `.output.std`). The extension itself has no meaning (I could have used just as well `.output.1`), but it is there as a hint that this output version contains `std::string` instead of `string` in the exception message.

PR #9951, opened 2020-04-24
Allow LinOp<LA::dist::Vector>(TrilinosMatrix).

dougshidong:
This follows the discussion from:

https://groups.google.com/forum/#!topic/dealii/MZS13T-UE6E

Basically, the `LinearOperatorImplementation` within the `LinearAlgebra::distributed::Vector` class assumes that a `MatrixFree` object is being used. It calls `initialize_dof_vector()` which only exists for `MatrixFree` and assumes a square matrix corresponding to the system to be solved.

As a result, a `TrilinosWrappers::SparseMatrix` wrapped within an operator will not be able to be applied onto a `LinearAlgebra::distributed::Vector` since it does not have the corresponding function.

The proposed fix checks at compile time whether `initialize_dof_vector()` or `get_mpi_communicator` exists and makes the correct call using a constexpr if. This is very 'hacky', but it's the quickest and least intrusive fix I could come up with without having to understand MatrixFree. 

A 'real' fix would be to have MatrixFree provide the the functions `locally_owned_range_indices()` and `get_mpi_communicator()`.

A test has been added.


peterrum:
@dougshidong  Sorry for not answering you in the forum yet. I am quite busy at the moment and wanted to have a look at your code later today.

I am wondering if it wouldn't be easier from user perspective to write a wrapper around `TrilinosWrappers::SparseMatrix`, which supports both `vmult()` and `initialize_dof_vector()`. What do you think?

kronbichler:
It is definitely a good idea to get things supported in both cases - thanks for the effort. However, we need to be a bit careful because you are using `if constexpr` which is C++17, whereas we need to also support compilation with the C++11 standard. What you want to achieve can be achieved with some additional writing by moving each block ("matrixfree" vs "SparseMatrix") to separate functions, selected with `std::enable_if`.

> A 'real' fix would be to have MatrixFree provide the functions `locally_owned_range_indices()` and `get_mpi_communicator()`.

The problem is somewhat more involved because `initialize_dof_vector` does more than just set the locally owned range and the MPI communicator in the matrix-free case: it sets the right format for the ghost indices as well, which is needed inside the mat-vec. Thus, this makes it a bigger question for how we want to handle vectors matching the matrix in deal.II so that the `LinearOperator` implementation is less of a pain. My preference would be that a `Matrix` object has something like `initialize_domain_vector()` and `initialize_range_vector()` which are compatible with `vmult()`. We need two different names because the matrix may be non-square with different partitions for rows and columns.

dougshidong:
> It is definitely a good idea to get things supported in both cases - thanks for the effort. However, we need to be a bit careful because you are using `if constexpr` which is C++17, whereas we need to also support compilation with the C++11 standard. What you want to achieve can be achieved with some additional writing by moving each block ("matrixfree" vs "SparseMatrix") to separate functions, selected with `std::enable_if`.

That is fine with me. I can make those changes. I guess the question is, would this 'hacky' fix be accepted? I can make the changes such that this works right now and open an issue for the 'real' fix. I am not exactly sure what the 'real' fix would be though.

> My preference would be that a `Matrix` object has something like `initialize_domain_vector()` and `initialize_range_vector()` which are compatible with `vmult()`.

It seems that those functions are currently part of the `Vector` instead of `Matrix` is because of the different ways `Vector`s are initialized. Once you're inside `Matrix`, it has no idea how to correctly initialize `Vector`.

> We need two different names because the matrix may be non-square with different partitions for rows and columns.

I'm assuming the reason why MatrixFree only has one name is because it always applies to square systems of equations?

kronbichler:
> I guess the question is, would this 'hacky' fix be accepted?

Yes, that would be perfectly fine - we have several dozen places like this in the library.

I also agree that you should open an issue describing how a more elegant fix would look like, so we don't forget about it.

dougshidong:
Took a look at some `std::enable_if` being used in deal.II and used those to re-write the fix.

Also opened issue #9961 regarding the more elegant solution.

peterrum:
@dougshidong Thanks for the contribution. BTW, #9925 has been merged. Let us know if you encounter additional issues when using `parallel::distributed::Vector`!

PR #9953, opened 2020-04-24
Include ComponentMask in vector_tools_interpolate.h

kronbichler:
clang complains rightfully that `ComponentMask` is only forward declared, despite using the default constructor:
https://cdash.43-1.org/testDetails.php?test=41845020&build=5839
Thus, we should include the proper header. Follow-up to #9929.


masterleinad:
Hmmm.. I wonder why this haven't shown up for me locally or in CI.

kronbichler:
I think the CI test that does `all-headers` is gcc, and gcc does not show the error.

Issue #8647, opened 2019-08-26
Discrepancies in refinement between p4est and Triangulation


marcfehling:
PR #8637 introduced a scenario in which the refinement and coarsening flags left after calling `prepare_coarsening_and_refinement()` do not represent how refinement will be performed by `p4est` during `execute_coarsening_and_refinement()`.

The fix introduced in this PR will ultimately rely on `p4est`'s refinement behavior. However as @bangerth stated, we are supposed to be predicting during `prepare_coarsening_and_refinement()` how `p4est` will perform refinement. This does not seem to be the case right now.

marcfehling:
See also #8993 and #9036.

tjhei:
Hey Marc, do you know what causes the difference between prepare and what p4est decides to do?

Something related has been on my todo list for a long time and I have (had?) a working implementation at some point. The idea was basically:
1. Turn off any smoothing/ 2-to-1 handling inside p4est.
2. Only use refinement flags on locally owned cells and perform "smoothing" (based on flags) correctly in parallel.
3. Exchange refinement flags in parallel and repeat 2&3 until we converge to a solution.

This might fix the issue at hand. If we hand a valid refinement request to p4est, p4est should be able to perform exactly that.

marcfehling:
Thank you for pointing out your idea! I am still not sure how this happens, but I ran into this problem again just today: `p4est` refined cells that did not have a `refine_flag` set from the `deal.II` side.

For now, I will continue to tie critical operations to the `p::d::Triangulation` callbacks, since the oracle specifies which cells will be refined in the end. I need to postpone tackling this issue for now - finishing my thesis becomes acute...

PR #9954, opened 2020-04-24
Reworked error prediction algorithm.

### LINKED ISSUE #9597: Extend error prediction algorithm for hp refinement.
The current implementation of `hp::Refinement::predict_error()` is insufficient and has limited use. I modified most of it in my PhD project, and would like to adjust it in the near future.

We should definitely make this part of the next release to avoid deprecating its current version. I'm sure that I'll provide my changes in time, but I'd like to keep this as a reminder to myself for now.

Please flag this issue for Release 9.2.

Part of #7515.

marcfehling:
Closes #9597.

Changed:
- Generalized `CoarseningStrategies` to `AdaptationStrategies`.
Upon adaptation, we need to decide how to process data on cells to be refined and to be coarsened. We only took care of the coarsening case so far. With this PR, we can now take care of the refinement case individually as well. The strategy-functions now also require an iterator to the parent cell: This will be of more general use in individual applications.
- Fixed error prediction algorithm.
The `hp::Refinement::predict_error` function is now also applicable on cells flagged for both h- and p-adaptation, and no longer just for either one. During h-adaptation, we make sure that the `l_2`-norm is preserved when distributing values from the parent to the children cells and vice versa. Further, it returns a vector whose `l_2`-norm now corresponds to the global predicted error. This way, it is possible to estimate how the global error will change quantitatively before adaptation actually happened.

I'll create the doxygen documentation on my workstation next to see if all LaTeX formulas are displayed correctly. But code and documentation in this PR are complete.

Part of #7515.

---

The hp-decision strategy based on error prediction would be applicable as is for the `parallel::distributed::Triangulation` as well when used with `parallel::distributed::CellDataTransfer`, if it wasn't for #8647: Since adaptation indicators might differ on the `p4est` mesh, the pre-calculated prediction of errors is no longer valid.

I have a solution for this problem sitting on a separate branch, which features a class `parallel::distributed::ErrorPrediction`. I would suggest to provide it to the library for the 9.2 release in a separate PR, and deprecate/remove the class later once we figured out #8647.

What do you think? (See #10013)


marcfehling:
The corresponding tables for look as follows:
- For each individual cell before performing adaptation. This is what's written in the output vector.
formulas-1
- For each individual cell after performing adaptation.
formulas-2

---

Ready for review!

marcfehling:
I just replaced the `0.5^dim` by `n_children` in the hp-refinement formula, since this is what we're actually doing in the code. This acts as a generalization from isotropic refinement.

kronbichler:
@marcfehling can you please rebase this to master?

include/deal.II/hp/refinement.h
```diff
@@ -388,69 +374,106 @@ namespace hp
...
      * Triangulation::n_active_cells(). The errors are interpreted to be
      * measured in the energy norm; this assumption enters the rate of
-     * convergence that is used in the prediction. The `predicted_errors` output
-     * argument has one entry per <i>current</i> cell, with the $2^d$ values for
-     * each cell that will be coarsened away equal, and with the value stored on
-     * a cell to be refined interpreted as applying to each of the future
-     * children.
+     * convergence that is used in the prediction. The $l_2$-norm of the output
+     * argument @p predicted_errors corresponds to the predicted global error
+     * after adaptation will be applied.
```
kronbichler, line 379:
```suggestion
     * after adaptation.
```

include/deal.II/hp/refinement.h
```diff
@@ -388,69 +374,106 @@ namespace hp
...
      * For h-adaptation, we expect the local error $\eta_K$ on cell $K$ to be
      * proportional to $(h_K)^{p_K}$ in the energy norm, where $h_K$ denotes the
      * cell diameter and $p_K$ the polynomial degree of the currently assigned
-     * finite element on cell $K$. Here, we assume that the finite element will
-     * not change in the adaptation process so that $p_K = \text{const}$.
-     * However during coarsening, the finite elements on siblings may be
+     * finite element on cell $K$.
+     *
+     * If both h- and p-adaptation are applied simultaneously, we need to
+     * determine the order at which which type of adaptation happensis
```
kronbichler, line 392: can you rephrase this? I do not get it.

include/deal.II/hp/refinement.h
```diff
@@ -388,69 +374,106 @@ namespace hp
...
+     * sure to configure your CellDataTransfer object with
+     * AdaptationStrategies::Refinement::l2_norm() as a refinement strategy and
+     * AdaptationStrategies::Coarsening::l2_norm() as a coarsening strategy.
+     * This ensures that the $l_2$-norm of the predict errors is preserved on
+     * both meshes.
+     *
+     * In the context, we assume that the local error on a cell that will be
+     * h-refined, will be divided equally on all $n_{K_c}$ children, whereas
+     * local errors on siblings will be summed up on the parent cell in case of
+     * h-coarsening. This assumption is often not satisfied in practice: For
```
kronbichler, line 446: can you rephrase this? This is a bit convoluted.

include/deal.II/numerics/adaptation_strategies.h
```diff
@@ -0,0 +1,355 @@
...
+     * Return sum of data on all children.
+     *
+     * @f[
+     *   d_{K_p} = \sum d_{K_c}
+     *   \qquad
+     *   \forall K_c \text{ children of } K_p
+     * @f]
+     *
+     * This strategy preserves the $l_1$-norm of the corresponding global data
+     * Vector before and after adaptation.
```
kronbichler, line 148:
```suggestion
     * vector before and after adaptation.
```
I see no need to capitalize this one.

include/deal.II/numerics/adaptation_strategies.h
```diff
@@ -0,0 +1,355 @@
...
+     * Return $ l_2 $-norm of data on all children.
+     *
+     * @f[
+     *   d_{K_p}^2 = \sum d_{K_c}^2
+     *   \qquad
+     *   \forall K_c \text{ children of } K_p
+     * @f]
+     *
+     * This strategy preserves the $l_2$-norm of the corresponding global data
+     * Vector before and after adaptation.
```
kronbichler, line 166:
```suggestion
     * vector before and after adaptation.
```

PR #9962, opened 2020-04-26
Simplify parallel::TriangulationBase::update_number_cache

peterrum:
A simplifications enabled by #9915.

This PR extends `parallel::TriangulationBase::update_number_cache()` with multigrid functionalities, which have been added till now in the derived classes.

FYI @tjhei


peterrum:
The complete test suite has passed locally.

PR #9964, opened 2020-04-27
Fix assertion in CUDA

Rombur:
Working on #8882 (PR coming for that one in the next few days), I realized that we didn't assert important kernels and there was a bug in the way we assert most of the kernel launchs. We used `cudaGetLastError`  but it resets the error code to success after being called. Because the function is called multiple times in `AssertCuda`, the error message is wrong.

### DIFF
--- a/include/deal.II/base/exceptions.h
+++ b/include/deal.II/base/exceptions.h
@@ -1770,6 +1770,37 @@ namespace internal
 #  endif
 
 /**
+ * An assertion that checks that the kernel was launched and executed
+ * successfully.
+ *
+ * @note This and similar macro names are examples of preprocessor definitions
+ * in the deal.II library that are not prefixed by a string that likely makes
+ * them unique to deal.II. As a consequence, it is possible that other
+ * libraries your code interfaces with define the same name, and the result
+ * will be name collisions (see
+ * https://en.wikipedia.org/wiki/Name_collision). One can <code>\#undef</code>
+ * this macro, as well as all other macros defined by deal.II that are not
+ * prefixed with either <code>DEAL</code> or <code>deal</code>, by including
+ * the header <code>deal.II/base/undefine_macros.h</code> after all other
+ * deal.II headers have been included.
+ *
+ * @ingroup Exceptions
+ * @author Bruno Turcksin, 2020
+ */
+#  ifdef DEBUG
+#    define AssertCudaKernel()                                \
+      {                                                       \
+        cudaError_t local_error_code = cudaPeekAtLastError(); \
+        AssertCuda(local_error_code);                         \
+        local_error_code = cudaDeviceSynchronize();           \
+        AssertCuda(local_error_code)                          \
+      }
+#  else
+#    define AssertCudaKernel() \
+      {}
+#  endif
+
+/**
  * An assertion that checks that the error code produced by calling a cuSPARSE
  * routine is equal to CUSPARSE_STATUS_SUCCESS.
  *
--- a/include/deal.II/lac/affine_constraints.templates.h
+++ b/include/deal.II/lac/affine_constraints.templates.h
@@ -1924,12 +1924,7 @@ namespace internal
       const int n_blocks = 1 + (n_constraints - 1) / CUDAWrappers::block_size;
       set_zero_kernel<<<n_blocks, CUDAWrappers::block_size>>>(
         constrained_local_dofs_device, n_constraints, vec.get_values());
-#  ifdef DEBUG
-      // Check that the kernel was launched correctly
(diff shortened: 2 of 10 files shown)


include/deal.II/base/exceptions.h
```diff
@@ -1769,6 +1769,35 @@ namespace internal
...
+ *
+ * @note This and similar macro names are examples of preprocessor definitions
+ * in the deal.II library that are not prefixed by a string that likely makes
+ * them unique to deal.II. As a consequence, it is possible that other
+ * libraries your code interfaces with define the same name, and the result
+ * will be name collisions (see
+ * https://en.wikipedia.org/wiki/Name_collision). One can <code>\#undef</code>
+ * this macro, as well as all other macros defined by deal.II that are not
+ * prefixed with either <code>DEAL</code> or <code>deal</code>, by including
+ * the header <code>deal.II/base/undefine_macros.h</code> after all other
```
masterleinad, line 1784: You also need to list the new macro in that file.

include/deal.II/lac/affine_constraints.templates.h
```diff
@@ -1924,12 +1924,7 @@ namespace internal
       const int n_blocks = 1 + (n_constraints - 1) / CUDAWrappers::block_size;
       set_zero_kernel<<<n_blocks, CUDAWrappers::block_size>>>(
         constrained_local_dofs_device, n_constraints, vec.get_values());
-#  ifdef DEBUG
-      // Check that the kernel was launched correctly
-      AssertCuda(cudaGetLastError());
-      // Check that there was no problem during the execution of the kernel
-      AssertCuda(cudaDeviceSynchronize());
-#  endif
+      AssertCudaKernel;
```
masterleinad, line 1927: Hmm... I would prefer this to look like a function call to be similar to the other asserts.
```suggestion
      AssertCudaKernel();
```

include/deal.II/base/exceptions.h
```diff
@@ -1769,6 +1769,35 @@ namespace internal
...
+ * deal.II headers have been included.
+ *
+ * @ingroup Exceptions
+ * @author Bruno Turcksin, 2020
+ */
+#  ifdef DEBUG
+#    define AssertCudaKernel                          \
```
masterleinad, line 1791:
```suggestion
#    define AssertCudaKernel()                        \
```

include/deal.II/base/exceptions.h
```diff
@@ -1769,6 +1769,35 @@ namespace internal
...
+      cudaError_t error_code = cudaPeekAtLastError(); \
+      AssertCuda(error_code);                         \
+      error_code = cudaDeviceSynchronize();           \
+      AssertCuda(error_code)
+#  else
+#    define AssertCudaKernel \
```
masterleinad, line 1797:
```suggestion
#    define AssertCudaKernel() \
```

include/deal.II/base/exceptions.h
```diff
@@ -1769,6 +1769,35 @@ namespace internal
...
+ * the header <code>deal.II/base/undefine_macros.h</code> after all other
+ * deal.II headers have been included.
+ *
+ * @ingroup Exceptions
+ * @author Bruno Turcksin, 2020
+ */
+#  ifdef DEBUG
+#    define AssertCudaKernel                          \
+      cudaError_t error_code = cudaPeekAtLastError(); \
+      AssertCuda(error_code);                         \
```
masterleinad, line 1793: Any reason to not just do
```suggestion
      AssertCuda(cudaPeekAtLastError());                         \
```
as we did in most places before?

bangerth: You'll have to wrap this piece of code in `{...}`. That's because you're creating a local variable, and if you don't create a scope for it, you can't have two `AssertCudaKernel` statements in the same scope.

include/deal.II/lac/affine_constraints.templates.h
bangerth: Yes, I strongly favor this as well.

include/deal.II/base/exceptions.h
Rombur: done

include/deal.II/lac/affine_constraints.templates.h
Rombur: done

include/deal.II/base/exceptions.h
Rombur: @masterleinad since I need to create an `error_code` if I don't want to call `cudaDeviceSynchronize` twice, I decided to use the `error_code` earlier so that I do not call `cudaPeekAtLastError()` twice
@bangerth done

PR #9969, opened 2020-04-27
Revisited binary search for parallel::distributed::GridRefinement.

marcfehling:
This patch allows negative indicators for refinement in the `parallel::distributed` context again.

It basically reverts a 6-year old commit https://github.com/dealii/dealii/commit/283a86b27f1a9a06ceb6ce90ac2142cdf47b97ac and I'm pretty sure this is the correct way to deal with the problem its author encountered back then.

I guess the cause of their issue was that the difference between both values in `interesting_range` was too large. Thus the lower value got decreased too much, resulting in a wrong value for the `test_threshold` during the binary search and ultimately yielding the wrong number of cells flagged for refinement.

Right now, I'll run the testsuite on my workstation to provide modified test results. But I doubt that there will be any, since in the current state of the code only positive indicators are possible.

@bangerth @tjhei -- What do you think?

### DIFF
--- a/tests/mpi/refine_and_coarsen_fixed_number_08.mpirun=1.with_p4est=true.output
+++ b/tests/mpi/refine_and_coarsen_fixed_number_08.mpirun=1.with_p4est=true.output
@@ -1,0 +1,16 @@
+
+DEAL::n_cells:100
+DEAL::criteria:[b>0,e>0]
+DEAL::  refinement_fraction:0.1 coarsening_fraction:0.1
+DEAL::    n_refine_flags:10 n_coarsen_flags:10
+DEAL::  refinement_fraction:0.1 coarsening_fraction:0.
+DEAL::    n_refine_flags:10 n_coarsen_flags:0
+DEAL::  refinement_fraction:0. coarsening_fraction:0.1
+DEAL::    n_refine_flags:0 n_coarsen_flags:10
+DEAL::criteria:[b=0,e>0]
+DEAL::  refinement_fraction:0.1 coarsening_fraction:0.1
+DEAL::    n_refine_flags:10 n_coarsen_flags:10
+DEAL::  refinement_fraction:0.1 coarsening_fraction:0.
+DEAL::    n_refine_flags:10 n_coarsen_flags:0
+DEAL::  refinement_fraction:0. coarsening_fraction:0.1
+DEAL::    n_refine_flags:0 n_coarsen_flags:10
--- a/source/distributed/grid_refinement.cc
+++ b/source/distributed/grid_refinement.cc
@@ -134,18 +134,30 @@ namespace
   {
     Assert(interesting_range[0] <= interesting_range[1], ExcInternalError());
 
-    Assert(interesting_range[0] >= 0, ExcInternalError());
-
-    // adjust the lower bound only if the end point is not equal to zero,
-    // otherwise it could happen, that the result becomes negative
     if (interesting_range[0] > 0)
-      interesting_range[0] *= 0.99;
-
-    if (interesting_range[1] > 0)
-      interesting_range[1] *= 1.01;
+      {
+        // In this case, we calculate the first interval split point `m` in the
+        // `compute_threshold` functions in the optimized way: We exploit that
+        // the logarithms of all criteria are more uniformly distributed than
+        // their actual values, i.e. m=sqrt(b*e).
+        //
+        // Both factors will modify the split point only slightly by a factor of
+        // sqrt(1.01*0.99) = sqrt(0.9999) ~ 0.9950.
+        interesting_range[0] *= 0.99;
+        interesting_range[1] *= 1.01;
+      }
     else
-      interesting_range[1] +=
-        0.01 * (interesting_range[1] - interesting_range[0]);
+      {
+        // In all other cases, we begin with an the arithmetic mean as the
+        // standard interval split point, i.e. m=(b+e)/2.
+        //
+        // Both increments will add up to zero when calculating the initial
+        // split point in the `compute_threshold` functions.
+        const double difference =
+          std::abs(interesting_range[1] - interesting_range[0]);
+        interesting_range[0] -= 0.01 * difference;
+        interesting_range[1] += 0.01 * difference;
+      }
   }
 
 
@@ -468,8 +480,8 @@ namespace parallel
                                                   tria.n_global_active_cells()),
             mpi_communicator);
 
-        // compute bottom threshold only if necessary. otherwise use a threshold
-        // lower than the smallest value we have locally
+        // compute bottom threshold only if necessary. otherwise use the lowest
+        // threshold possible
         if (adjusted_fractions.second > 0)
           bottom_threshold = dealii::internal::parallel::distributed::
             GridRefinement::RefineAndCoarsenFixedNumber::compute_threshold(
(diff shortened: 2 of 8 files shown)


marcfehling:
Hmm, I'm still thinking about the case if one or both values of `interesting_range` are zero.

We need to slightly increase the interval in this case as well according to @tjhei's comment in the original code. So what was the intention of the patch, and why did the patch work? I'm puzzled

marcfehling:
I concluded that it makes the most sense to adjust the range similar to the way we calculate the initial interval split points, and noted this in the comments. Not sure how the original patch was supposed to work, but right now this is my best solution.

tests/mpi/refine_and_coarsen_fixed_number_08.mpirun=1.with_p4est=true.output
```diff
@@ -0,0 +1,37 @@
...
+DEAL::    n_refine_flags:10 n_coarsen_flags:10
+DEAL::  refinement_fraction:0.1 coarsening_fraction:0.
+DEAL::    n_refine_flags:10 n_coarsen_flags:0
+DEAL::  refinement_fraction:0. coarsening_fraction:0.1
+DEAL::    n_refine_flags:0 n_coarsen_flags:10
+DEAL::criteria:[b=0,e>0]
+DEAL::  refinement_fraction:0.1 coarsening_fraction:0.1
+DEAL::    n_refine_flags:10 n_coarsen_flags:10
+DEAL::  refinement_fraction:0.1 coarsening_fraction:0.
+DEAL::    n_refine_flags:10 n_coarsen_flags:1
```
marcfehling, line 14: There was a bit more to modify to allow negative criteria, which only make sense with the `fixed_number` strategy by the way.

I provided corresponding changes including a corresponding test which looks good. However, I'm not happy with this single cell being coarsened. I'll investigate.

marcfehling: The cause of this were these lines. https://github.com/dealii/dealii/blob/c4c3077c421f758bf76f8eca539a7bdda80415fb/source/distributed/grid_refinement.cc#L549-L551
If no cells will be coarsened and the minimum of all criteria is zero, then the bottom_threshold is also zero and not lower than the minimum. Setting bottom_threshold to std::numeric_limits<Number>::lowest() did the trick

marcfehling:
Ready for review!

marcfehling:
I interpret refinement criteria to have in principle any value and they do not necessarily have to be connected to error indicators at all.

In case of the `fixed_number` strategy, thresholds from criteria are deduced from each cell's criterion individually: `X%` of cells with the largest indicator will be flagged for refinement, and `Y%` with the lowest indicator will be flagged for coarsening. So in this context, a negative value would mean that the criteria is smaller than on other cells and would indicate coarsening. The user could go over all vector entries and `std::abs` all values if they want to work with absolute values.

For the `fixed_fraction` strategy however, in which the sum of criteria on multiple cells are used as a measure to determine thresholds, I see your point and agree with it: It is not clear whether to perform the sum on the actual values or absolute ones. I refrained from changing the `fixed_fraction` strategy and only changed the former one.

---

I stumbled over using negative criteria not in the context of `GridRefinement`, but in my studies with `hp::Refinement`: With the refinement history strategy, I calculate predicted errors `eta_pred` from the error indicators `eta`. As a criteria to decide between h- and p-adaptation, I chose to use `eta_pred - eta`, at which a large difference would indicate p-adaptation, while a low difference would indicate h-adaptation. These criteria may be negative.

I applied these differences as criteria for the `hp::Refinement::p_adaptivity_fixed_number` function, which reuses the binary search algorithm from the `p::d::GridRefinement` framework if a `p::d::Triangulation` is used. This is the point where I initially ran into the assertion in the `adjust_interesting_range` function which strictly forbids negative criteria and was the initial motivation to write this patch. My initial hotfix to make it work anyways was to add the lowest value of the criteria vector to all of its elements, so that the lowest element is zero. This modification does not yield different results in the context of the `fixed_number` logic: The same largest and lowest values are still located on the same cells.

So I thought it would make sense to also generalize the `(p::d::)GridRefinement::refine_and_coarsen_fixed_number` pendants as well. In principle users could come up with anything, as demonstrated in the provided tests of this PR. However, I didn't change any of the `(p::d::)GridRefinement::refine_and_coarsen_fixed_fraction` equivalents: The corresponding assertions `criteria.is_non_negative()` are still part of them.

---

I hope I could convince you for this change. If you're still against it, we could drop major parts of the second commit. However, I must insist to make the binary search work for negative values :-)

(If we stick with negative criteria, we should probably rename the term `smallest indicator` with `lowest indicator` to be more clear.)

bangerth:
Hm, I'm afraid I'm not convinced. I think that in the community, we think of refinement indicators as reflecting some kind of error norm, or at least an estimate for an error in some kind of quantity or a deviation from a desired state -- which only makes sense if one measures the error using a distance function `d(u,u_h)`. That is, we have a metric that measures *distance*, which is a non-negative quantity. 

My preference would be to keep with the expectation of non-negative values (which, if it isn't already, should be asserted). If you have a criterion that "estimates" the error by some kind of extrapolation and this extrapolation creates negative values, then just set these values to zero (because the predicted value clearly can't be negative) or shift as you already do. But I think that should happen in user code (or in the code that does the extrapolation, if that happens in the library). But that seems like the right place for it to me.

marcfehling:
I agree with you when we are talking about this particular metric.

I'm now thinking about applying the same restriction on the `hp::Refinement` algorithms as well, but in this context we do not necessarily think of indicators being tied to some kind of norm. But we might think about this in a different pull request.

I would revert the aspects about negative criteria in the second commit, but keep the test since it revealed a bug. Do you agree to the changes made on the first commit?

bangerth:
Yes, if I interpret this right, then the changes in the first commit are just a small optimization to the algorithm with no change to the semantics of the functions that use the algorithm. Is this right?

marcfehling:
Correct, these changes have no effect on the semantics or the interface. The actual changes to the library are now just about a dozen lines. The remaining lines are tests.

I made the corresponding adjustments. Ready for a final review.

bangerth:
Do you want this in 9.2?

marcfehling:
Since this fixes some minor bugs, it would be favorable to have it in 9.2.

I'll prepare a pull request for the 9.2 branch.

marcfehling:
```
An error occurred in line <205> of file </jenkins/workspace/dealii_PR-9969/source/grid/grid_refinement.cc> in function
    void dealii::GridRefinement::refine_and_coarsen_fixed_number(dealii::Triangulation<dim, spacedim>&, const dealii::Vector<Number>&, double, double, unsigned int) [with int dim = 2; Number = float; int spacedim = 2]
The violated condition was: 
    criteria.is_non_negative()
Additional information: 
    (none)
```

Since negative criteria are no longer allowed, I should also adjust the new test...

kronbichler:
This PR seems to cause 6 new test failures:
https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=6397
`mpi/mpi_ghost_owners_01, mpi/mpi_ghost_owners_02`
I think this is simply a matter of adjusting the output files. I saw this on the 9.2 release branch, so this last-minute change should have deserved on run of the full test suite.

kronbichler:
I'm merging anyway because it is merged on the release branch and we need find a fix for both master and the release branch.

PR #9973, opened 2020-04-28
make ARKode ODE solver use DiscreteTime internally

rezarastak:
I modified the `ARKode` class to use `DiscreteTime`. It helped simplify some aspects of time-stepping and stopping at the proper end time.

I had to change two lines of test output, so this is technicality an incompatible change. However, the new implementation fixes a bug in the old code where it used to unnecessarily run the last step with a tiny `dt` and print two lines of identical output.

note: the call to `ARKodeGetLastStep()` is not needed anymore because we instead can call `time.get_previous_step_size()`.


rezarastak:
I rebased this PR and it can be reviewed.

source/sundials/arkode.cc
```diff
@@ -302,33 +294,43 @@ namespace SUNDIALS
...
 
-    double next_time = data.initial_time;
+    DiscreteTime time(data.initial_time,
+                      data.final_time,
+                      data.initial_step_size);
+    reset(time.get_current_time(), time.get_next_step_size(), solution);
 
     if (output_step)
-      output_step(0, solution, 0);
+      output_step(time.get_current_time(), solution, time.get_step_number());
```
bangerth, line 304: Ah, nice -- that was a real bug I guess :-)

source/sundials/arkode.cc
```diff
@@ -302,33 +294,43 @@ namespace SUNDIALS
...
 
-    while (t < data.final_time)
+    while (time.is_at_end() == false)
       {
-        next_time += data.output_period;
+        time.set_desired_next_step_size(data.output_period);
```
bangerth, line 308: This means that `data.output_period` is really only a suggested period. Can you document this in the documentation of the class of which `data` is an object?

source/sundials/arkode.cc
```diff
@@ -302,33 +294,43 @@ namespace SUNDIALS
...
-        status = SundialsARKode(arkode_mem, next_time, yy, &t, ARK_NORMAL);
+        {
+          double     t;
+          const auto status = SundialsARKode(
+            arkode_mem, time.get_next_time(), yy, &t, ARK_NORMAL);
+          (void)status;
+          AssertARKode(status);
+          time.set_desired_next_step_size(t - time.get_current_time());
+          Assert(std::abs(time.get_next_time() - t) < 1.e-11,
+                 ExcInternalError());
```
bangerth, line 318: I don't think I understand these 3 lines. Can you explain what they do?

bangerth: Specifically, what you're doing here makes sure that the next time around `time.advance_time()` (in your line 321) will move the time to the end time. Or do I misunderstand something?

source/sundials/arkode.cc
```diff
@@ -302,33 +294,43 @@ namespace SUNDIALS
...
+        }
 
-        AssertARKode(status);
-
-        status = ARKodeGetLastStep(arkode_mem, &h);
-        AssertARKode(status);
+        time.advance_time();
 
         copy(solution, yy);
```
bangerth, line 323: Could we move this call right after the call to `SundialsARKode` (i.e., right after the error check for that call)? I think that's where it should conceptually be, not after we advance time.

source/sundials/arkode.cc
rezarastak: Let me try to explain this better. ARKode may choose to change the time-step. So the second argument provided to `SundialsARKode` is the desired next time, and the 4th argument of the function is the actual next time, set by the function. So we use the value of `t` to guide the `time` object, but we expect the `DiscreteTime` class not to modify the step size further. Hence I added the `Assert` statement.

I understand this part of the code is probably not very readable, so I will rewrite this part.

Moreover, I think I should add a new member function called `DiscreteTime::advance_time_to(double new_time)` which guarantees to jump to `new_time` exactly (without any possibility of modifying `new_time`). This method asserts that current_time < new_time <= end_time. The introduction of this member function will simplify the code of this PR as I can just write `time.advance_time_to(t)`.

rezarastak: > I think I should add a new member function called DiscreteTime::advance_time_to(double new_time)

Instead of `advance_time_to()` I can introduce the member function `DiscreteTime::set_next_step_size(double step_size)` which is similar to `set_desired_next_step_size()` but does not modify the step size argument. We should use `set_next_step_size()` here because ARKode automatically adjusts the time steps and takes care of stopping at the end_time. Here we don't DiscreteTime to act smart.

source/sundials/arkode.cc
rezarastak: done

source/sundials/arkode.cc
rezarastak: done

rezarastak:
Using the member function introduced in #10621, I simplified this PR and incorporated all of @bangerth's comments. It is ready to be reviewed. Thank you!

PR #9974, opened 2020-04-28
Restrict mpi/petsc_step-27

masterleinad:
This test requires Hypre.


drwells:
It looks like `trilinos_step-27.cc` and `interpolate_to_different_mesh_02.cc` (which also uses trilinos) in the same directory might suffer from the same problem. Could you fix those too while we are there?

masterleinad:
@drwells The test you are mentioning are using `TrilinosWrappers::PreconditionAMG` which is based on `ML` and that is a required module.

drwells:
Ah: nevermind!

PR #9980, opened 2020-04-28
Improve error message in SymmetricTensor constructor.

simonsticko:


jppelteret:
Thanks for the fix @simonsticko!

PR #9968, opened 2020-04-27
Precalculate hp::FEValues.

marcfehling:
For timing purposes, it makes sense to precalculate all `FEValues` objects in the `hp::FEValues` collection. This PR provides such a feature.

Kind of part of #7515.

@bangerth -- FYI

PR #9986, opened 2020-04-28
Fix copying hp::FEValues objects.

bangerth:
This falls in the category of "How the heck did we never notice this!?!?". While looking at @marcfehling's #9968, I realized that `hp::FEValuesBase` uses `std::shared_ptr` objects to underlying `FEValues` objects. It had no explicit copy constructor or operator, and so source and destination in the automatically generated copy constructor keep pointers to shared space.

Sure enough, it's not very difficult to generate a test where you do this:
```
  const auto        cell1 = dof_handler.begin_active();
  hp::FEValues<dim> hp_fe_values1(fe_collection,
                                  quadrature_collection,
                                  update_quadrature_points);
  hp_fe_values1.reinit(cell1);

  // Now make a copy of the object and initialize it for the second cell
  const auto        cell2 = ++dof_handler.begin_active();
  hp::FEValues<dim> hp_fe_values2(hp_fe_values1);
  hp_fe_values2.reinit(cell2);
```
and now `hp_fe_values1` has the data that corresponds to `cell2`. Not what we wanted. Worse: This is the kind of thing we do all the time in `WorkStream` (where the copies are used on separate threads).

It's unclear to me how we never noticed this issue! Either way, it's easy to fix: Just use `std::unique_ptr` and write an explicit copy constructor.

/rebuild

### DIFF
--- a/source/hp/fe_values.cc
+++ b/source/hp/fe_values.cc
@@ -13,6 +13,7 @@
 //
 // ---------------------------------------------------------------------
 
+#include <deal.II/base/std_cxx14/memory.h>
 #include <deal.II/base/thread_management.h>
 
 #include <deal.II/fe/mapping_q1.h>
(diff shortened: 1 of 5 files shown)


source/hp/fe_values.cc
```diff
@@ -121,7 +160,7 @@ namespace hp
         task_group +=
           Threads::new_task([&, fe_index, mapping_index, q_index]() {
             fe_values_table(TableIndices<3>(fe_index, mapping_index, q_index)) =
-              std::make_shared<FEValuesType>(
+              std::make_unique<FEValuesType>(
```
masterleinad, line 163:
```
/jenkins/workspace/dealii_PR-9986/source/hp/fe_values.cc:163:15: error: ‘make_unique’ is not a member of ‘std’
```

bangerth:
Fixed now, and tests correctly!

PR #9987, opened 2020-04-28
Add range based loop for quadrature points to FEValuesBase

### LINKED ISSUE #9972: Add quadrature_point_indices() to FeValuesBase
Mimicking `GeometryInfo<dim>::face_indices()`, we should have something like `FeValuesBase::quadrature_point_indices()` that allows us to write `for (const unsigned int q_point : fe_values.quadrature_point_indices())`.

jppelteret:
Fixes #9972

### DIFF
--- a/doc/news/changes/minor/20200429Jean-PaulPelteret
+++ b/doc/news/changes/minor/20200429Jean-PaulPelteret
@@ -1,0 +1,5 @@
+New: There is a new function FEValuesBase::quadrature_point_indices()
+that makes writing range-based for loops over all quadrature points of a 
+cell or face much simpler.
+<br>
+(Jean-Paul Pelteret, 2020/04/29)
--- a/include/deal.II/fe/fe_values.h
+++ b/include/deal.II/fe/fe_values.h
@@ -39,6 +39,8 @@
 
 #include <deal.II/hp/dof_handler.h>
 
+#include <boost/range/irange.hpp>
+
 #include <algorithm>
 #include <memory>
 #include <type_traits>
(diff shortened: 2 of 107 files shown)


include/deal.II/fe/fe_values.h
```diff
@@ -2912,6 +2914,25 @@ class FEValuesBase : public Subscriptor
...
+  /**
+   * Return an object that can be thought of as an array containing all
+   * indices from zero to `n_quadrature_points`. This allows to write code
+   * using range-based for loops of the following kind:
+   * @code
+   *   FEValues<dim> fe_values (...);
+   *
+   *   for (auto &cell : dof_handler.active_cell_iterators())
+   *     fe_values.reinit(cell);
+   *     for (auto q_point : fe_values.quadrature_point_indices())
```
bangerth, line 2926: too much python writing -- you need curly braces

jppelteret: Drats. I snuck this in before sitting down to dinner. Syntax goes out the window when you're hungry!

jppelteret:
> Do you plan on adding a test, or just converting a couple of tests?

I think that I'll convert some tests, if that's ok with you. Did we do a sweeping replacement when we added range based loops for cell iterators? If so, then maybe I'll do a search-replace in the library too. I've marked this as WIP for now, since there's still a few days to complete this.

bangerth:
I think if you convert a handful of tests, then that will be good enough.

It would be nice to convert the tutorials as well, but that's clearly for a later patch.

tests/fe/abf_01.cc
```diff
@@ -99,7 +99,7 @@ void EvaluateDerivative(DoFHandler<2> *dof_handler, Vector<double> &solution)
         n_q_points, std::vector<Tensor<1, 2>>(n_components));
       fe_values.get_function_gradients(solution, grads_here);
 
-      for (unsigned int q_point = 0; q_point < n_q_points; ++q_point)
+      for (auto q_point : fe_values.quadrature_point_indices())
```
kronbichler, line 102: I think we more often use the variant with `const` qualifier
```suggestion
      for (const auto q_point : fe_values.quadrature_point_indices())
```
How do others think about this?

jppelteret: Thanks, I came to the same realisation last night when putting together #9990. I didn't make the same mistake there. I've fixed all of the tests and documentation, so now all of these are `const`.

doc/news/changes/minor/20200429Jean-PaulPelteret
```diff
@@ -0,0 +1,5 @@
+New: There is a new function FEValuesBase::quadrature_point_indices()
+that make writing range-based for loops over all quadrature points of a
```
bangerth, line 2: makes

jppelteret: Thanks for the review and for pointing that out. Fixed now!

bangerth:
The OSX parallel 64bit check refuses to run but everything else seems ok, so I'm merging!

PR #7838, opened 2019-03-19
add Utilities::MPI::create_ascending_partitioning()

davydden:

PR #9991, opened 2020-04-30
Add Utilities::MPI::create_evenly_distributed_partitioning()

dougshidong:
Very very similar to pull request #7838, except we provide a global size rather than a local size.


source/base/mpi.cc
```diff
@@ -243,6 +243,22 @@ namespace Utilities
       return res;
     }
 
+    std::vector<IndexSet>
+    create_evenly_distributed_partitioning(
+      const MPI_Comm &           comm,
+      const IndexSet::size_type &total_size)
```
peterrum, line 249: I am not sure about the interface of this function. Passing a communicator seems not necessary. It is enough to pass `n_partitions`:
```suggestion
    std::vector<IndexSet>
    create_evenly_distributed_partitioning(
      const unsigned int         n_partitions,
      const IndexSet::size_type &total_size)
```

dougshidong: For simplicity, I made it call create_ascending_partitioning(), which is a collective operation and requires a communicator. However, it's true that I can make it completely independent of MPI calls.

I think it still makes sense to put it in the Utilities::MPI, next to create_ascending_partitioning() since this partitioning will only be needed for parallel computations.

dougshidong: Come to think of it, whatever function is calling this one will have to call Utilities::MPI::n_mpi_processes(MPI_Comm) beforehand to get n_partitions. Might as well put it within this function by passing the communicator

I can still make the function non-collective by removing its dependence on create_ascending_partition() if that's preferable.

masterleinad: Yes, I think passing `MPI_Comm` is fine even if just for getting the local id and the number of total partitions.
I would prefer to not call `create_ascending_partitioning()`, though.

peterrum: I would prefer to have two functions (name?): one is taking a communicator, calls `Utilities::MPI::n_mpi_processes(comm)` and calls the other function.

Sounds good:

> I can still make the function non-collective by removing its dependence on create_ascending_partition() if that's preferable.

dougshidong: Thanks. Made the changes as suggested and force-pushed them.

include/deal.II/base/mpi.h
```diff
@@ -431,6 +431,32 @@ namespace Utilities
...
+     */
+    std::vector<IndexSet>
+    create_evenly_distributed_partitioning(
+      const unsigned int &       n_partitions,
+      const IndexSet::size_type &total_size);
+
+    /**
+     * Given the total number of elements @p total_size,
+     * create an evenly distributed 1:1 partitioning of the of elements across the MPI communicator @p comm.
+     * Uses @p comm to determine number of partitions and calls @p create_evenly_distributed_partitioning() function above.
```
peterrum, line 453: Could you break the lines.

source/base/mpi.cc
```diff
@@ -93,6 +93,33 @@ namespace Utilities
...
+            ++sizes[i_proc];
+        }
+
+      std::vector<IndexSet> res(n_partitions, IndexSet(total_size));
+      IndexSet::size_type   begin = 0;
+      for (unsigned int i_proc = 0; i_proc < n_partitions; ++i_proc)
+        {
+          res[i_proc].add_range(begin, begin + sizes[i_proc]);
+          begin = begin + sizes[i_proc];
+        }
```
peterrum, line 119: I would merge this to a single for loop.

dougshidong:
Made the changes and moved it out of the MPI namespace. However, I did keep it within the mpi.cc and mpi.h files instead of moving the to the utilities.cc/h files. Let me know if you'd like me to move them.

masterleinad:
Oh, I expected this function to only return one `IndexSet` and not a `std::vector` but realizing now that `create_ascending_partitioning` also returns a `std::vector`. Can we briefly discuss if that's useful?
Is there a case where we need to know about the `IndexSet` for the other processes?

dougshidong:
> Oh, I expected this function to only return one `IndexSet` and not a `std::vector` but realizing now that `create_ascending_partitioning` also returns a `std::vector`. Can we briefly discuss if that's useful?
> Is there a case where we need to know about the `IndexSet` for the other processes?

I don't see why we would need `IndexSet` for the other processes. My personal use of this function basically gets the vector, and query my owned `IndexSet`. However, I stumbled upon the `create_ascending_partitioning()` and basically "copied" their pull request to expedite the process since I assumed it was previously vetted.

Personally, I would have had a function like `complete_index_set()` within the index_set files. At least, that's where I was looking for my contributed function.

masterleinad:
We are currently not using `create_ascending_partitioning` internally at all but I could think about a number of places where we could use the new function but only need the information for the current process.
Hence, I would be in favor of just returning one `IndexSet`. Since this depends on the current process id, it makes sense to me to have it in `Utilities::MPI` (in particular because we also have `Utilities::MPI::Partitioner`).
I don't have a strong opinion on the name, though.

dougshidong:
Alright, made it return a single IndexSet.

include/deal.II/base/mpi.h
```diff
@@ -104,6 +104,23 @@ class IndexSet;
 
 namespace Utilities
 {
+  /**
+   * Given the total number of elements @p total_size, create an evenly
+   * distributed 1:1 partitioning of the of elements for across @p n_partitions.
+   * The local sizes will be equal to the @p total_size divided by the number
+   * of partitions plus the remainder being divided amongst the first
+   * processors. Each process will store contiguous subset of indices, and the
```
masterleinad, line 112:
```suggestion
   * processors. Each process will store a contiguous subset of indices, and the
```

source/base/mpi.cc
```diff
@@ -68,6 +68,28 @@ DEAL_II_NAMESPACE_OPEN
...
+  IndexSet
+  create_evenly_distributed_partitioning(const unsigned int &my_partition_id,
+                                         const unsigned int &n_partitions,
+                                         const IndexSet::size_type &total_size)
+  {
+    const unsigned int remain = total_size % n_partitions;
+
+    unsigned int min_size = total_size / n_partitions;
+
+    std::vector<IndexSet> res(n_partitions, IndexSet(total_size));
```
masterleinad, line 80: I don't think you need to create a `std::vector`. You could just do
```
begin=min_size*my_partition_id+std::min(my_partition_id, remain)
end=min_size*(my_partition_id+i)+std::min(my_partition_id+1, remain)
```

dougshidong:
Thanks for the suggestions. Implemented those and forced pushed.

source/base/mpi.cc
```diff
@@ -68,6 +68,24 @@ DEAL_II_NAMESPACE_OPEN
 
 namespace Utilities
 {
+  IndexSet
+  create_evenly_distributed_partitioning(const unsigned int &my_partition_id,
```
masterleinad, line 72: Let's not pass `unsigned int` as reference.
```suggestion
  create_evenly_distributed_partitioning(const unsigned int my_partition_id,
```

source/base/mpi.cc
```diff
@@ -68,6 +68,24 @@ DEAL_II_NAMESPACE_OPEN
...
+                                         const unsigned int &n_partitions,
```
masterleinad, line 73:
```suggestion
                                         const unsigned int n_partitions,
```

source/base/mpi.cc
```diff
@@ -243,6 +261,19 @@ namespace Utilities
       return res;
     }
 
+    IndexSet
+    create_evenly_distributed_partitioning(
+      const MPI_Comm &           comm,
+      const IndexSet::size_type &total_size)
```
masterleinad, line 267:
```suggestion
      const IndexSet::size_type total_size)
```

source/base/mpi.cc
```diff
@@ -68,6 +68,24 @@ DEAL_II_NAMESPACE_OPEN
 
 namespace Utilities
 {
+  IndexSet
+  create_evenly_distributed_partitioning(const unsigned int &my_partition_id,
+                                         const unsigned int &n_partitions,
+                                         const IndexSet::size_type &total_size)
```
masterleinad, line 74:
```suggestion
                                         const IndexSet::size_type total_size)
```

source/base/mpi.cc
```diff
@@ -664,6 +695,14 @@ namespace Utilities
       return std::vector<IndexSet>(1, complete_index_set(local_size));
     }
 
+    IndexSet
+    create_evenly_distributed_partitioning(
+      const MPI_Comm & /*comm*/,
+      const IndexSet::size_type &total_size)
```
masterleinad, line 701:
```suggestion
      const IndexSet::size_type total_size)
```

doc/news/changes/minor/20200429DougShiDong
```diff
@@ -0,0 +1,4 @@
+New: Add Utilities::MPI::create_evenly_distributed_partitioning() to create a one-to-one
+evenly distributed ascending partitioning from total size.
```
masterleinad, line 2: You should also add the other function here.

tests/base/mpi_evenly_distributed_partitioning_01.output
```diff
@@ -0,0 +1,4 @@
+
+DEAL:0::Total size=17
```
masterleinad, line 2: The file name needs to be `tests/base/mpi_evenly_distributed_partitioning_01.mpirun=1.output` to avoid trouble when running the tests simultaneously.

tests/base/mpi_evenly_distributed_partitioning_01.cc
```diff
@@ -0,0 +1,100 @@
...
+      vec[i_proc] =
+        Utilities::create_evenly_distributed_partitioning(i_proc,
+                                                          n_proc,
+                                                          total_size);
+    }
+  // correct number:
+  AssertThrow(index_set == vec[my_proc], ExcInternalError());
+
+  // correct number:
+  AssertThrow(vec.size() == n_proc, ExcInternalError());
```
masterleinad, line 47: This check is pretty pointless. :wink:

dougshidong:
Thanks for the review! Actually, the reference to `unsigned int` was from copying the `create_ascending_partitioning()` function. Same goes for the output filename needing `mpirun=1`. 

Want me to fix those as well while I'm at it?

Made the changes and force-pushed again.

masterleinad:
Yes, please do in a separate commit. Alternatively, you could also name the file `tests/base/mpi_evenly_distributed_partitioning_01.with_mpi=false.output`

include/deal.II/base/mpi.h
```diff
@@ -104,6 +104,23 @@ class IndexSet;
 
 namespace Utilities
 {
+  /**
+   * Given the total number of elements @p total_size, create an evenly
+   * distributed 1:1 partitioning of the of elements for across @p n_partitions.
```
kronbichler, line 109:
```suggestion
   * distributed 1:1 partitioning of the elements for across @p n_partitions.
```

include/deal.II/base/mpi.h
```diff
@@ -104,6 +104,23 @@ class IndexSet;
...
+   * The local sizes will be equal to the @p total_size divided by the number
+   * of partitions plus the remainder being divided amongst the first
+   * processors. Each process will store a contiguous subset of indices, and the
```
kronbichler, line 112:
```suggestion
   * processes. Each process will store a contiguous subset of indices, and the
```

include/deal.II/base/mpi.h
```diff
@@ -428,8 +445,20 @@ namespace Utilities
...
      */
     std::vector<IndexSet>
-    create_ascending_partitioning(const MPI_Comm &           comm,
-                                  const IndexSet::size_type &local_size);
+    create_ascending_partitioning(const MPI_Comm &          comm,
+                                  const IndexSet::size_type local_size);
+
+    /**
+     * Given the total number of elements @p total_size, create an evenly
+     * distributed 1:1 partitioning of the of elements across the
```
kronbichler, line 453:
```suggestion
     * distributed 1:1 partitioning of the elements across the
```

dougshidong:
Fixed the typos! Thank you!

PR #9993, opened 2020-04-30
Fix SFINAE detection of LinOp for matrix-free operators

kronbichler:
Follow-up to #9951. This fixes the test
https://cdash.43-1.org/testDetails.php?test=42401078&build=5914
The problem was that `initialize_dof_vector` takes a non-const reference of a `LA::dist::Vector`, which means we cannot construct an exemplar in the function, but we must use a variable we can create a reference to.

While there, also fix the oversight that the outer class takes `Vector<Number>`, not just `Vector<double>`.


bangerth:
The "CI - mpi build" failure is spurious.

PR #9996, opened 2020-04-30
CMake: Clear all cached variables in FindGINKGO.cmake

tamiko:


jppelteret:
Is there much point in running this though the testers? Do any of them have Ginkgo installed?

masterleinad:
> Is there much point in running this though the testers? Do any of them have Ginkgo installed?

No, let's just merge.

PR #9997, opened 2020-04-30
Remove MeshWorker::CopyData::operator=(double)

### LINKED ISSUE #9971: Mesh loop assemble_boundary_faces uses empty data structures
I've attached a test where I'm just trying to compute the volume of a triangulation and the area of its boundary using `mesh_loop()`; something easily verifiable on a 1x1 domain. Volume integration seems to work fine. But when I do the boundary integral alone by specifying to `mesh_loop` that it must only `MeshWorker::assemble_boundary_faces`, then I do not get the expected value for the boundary integral (no, the total area of the boundary is not 3 in 2d or 3.5 in 3d!). I've checked that local integral is being computed in the `boundary_worker`, but where the problem seems to manifest itself is the `copier`. On some faces, the `CopyData` sent to the `copier` seems to be reset, so that sometimes `0` is accumulated into the destination. 

@luca-heltai, @tjhei Do you have any idea what might be going wrong here? I'm not 100% sure if this is a bug in the library as opposed to me misusing `mesh_loop`, so I'd be happy to be informed if there's an error in my own code.

[mesh_loop_integrals_01.cc.txt](https://github.com/dealii/dealii/files/4541968/mesh_loop_integrals_01.cc.txt)
[mesh_loop_integrals_01.output.txt](https://github.com/dealii/dealii/files/4541969/mesh_loop_integrals_01.output.txt)

jppelteret:
Also add an example of what not to do with copiers inside of a boundary worker.

Fixes #9971

### DIFF
--- a/doc/news/changes/incompatibilities/20200430Jean-PaulPelteret
+++ b/doc/news/changes/incompatibilities/20200430Jean-PaulPelteret
@@ -1,0 +1,6 @@
+Removed: The function `MeshWorker::CopyData::operator=(const double)` has been
+removed, as resetting the state of the `copier` from within a `boundary_worker`
+or `face_worker` is not permitted and can lead to hard to trace bugs within a
+user's code.
+<br>
+(Jean-Paul Pelteret, 2020/04/30)
--- a/include/deal.II/meshworker/copy_data.h
+++ b/include/deal.II/meshworker/copy_data.h
@@ -72,19 +72,6 @@ namespace MeshWorker
       default;
 
     /**
-     * Allow resetting of all elements of the struct to zero, by simply
-     * calling `(*this) = 0;`
-     *
-     * Notice that the only allowed number here is really `0`. Calling this
-     * function with any other number will trigger an assertion.
-     *
-     * The elements of the arrays of local degrees of freedom indices are
-     * all set to numbers::invalid_dof_index.
-     */
-    void
-    operator=(const double number);
-
-    /**
      * An array of local matrices.
      */
     std::array<FullMatrix<double>, n_matrices> matrices;
@@ -134,25 +121,6 @@ namespace MeshWorker
 
     for (unsigned int i = 0; i < n_dof_indices; ++i)
       local_dof_indices[i].resize(dof_indices_sizes[i++]);
-  }
-
-
-
-  template <int n_matrices, int n_vectors, int n_dof_indices>
-  void
-  CopyData<n_matrices, n_vectors, n_dof_indices>::operator=(const double number)
-  {
-    Assert(number == 0.0,
-           ExcMessage("You should only call this method with "
-                      "argument 0.0"));
-
-    for (auto &m : matrices)
-      m = number;
-    for (auto &v : vectors)
-      v = number;
-    for (auto &d : local_dof_indices)
-      for (auto &val : d)
-        val = numbers::invalid_dof_index;
   }
 
 #endif // DOXYGEN
--- a/include/deal.II/meshworker/copy_data.h
+++ b/include/deal.II/meshworker/copy_data.h
@@ -55,12 +55,12 @@ namespace MeshWorker
      * Initialize everything with the same @p size. This is usually the number
      * of local degrees of freedom.
      */
-    CopyData(const unsigned int size);
+    explicit CopyData(const unsigned int size);
 
     /**
      * For every object, specify the size they should have.
      */
-    CopyData(
+    explicit CopyData(
       const std::array<std::array<unsigned int, 2>, n_matrices> &matrix_sizes,
       const std::array<unsigned int, n_vectors> &                vector_sizes,
       const std::array<unsigned int, n_dof_indices> &dof_indices_sizes);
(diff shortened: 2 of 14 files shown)


jppelteret:
I'd like to get @luca-heltai's opinion on this before we get rid of something that he actually uses!

luca-heltai:
I do use it, but I can survive without. I agree that this construction is complicated, but just as @bangerth stated, I do not have a different option.

jppelteret:
Drats, I see that some tests failed. I'll check them out later today.

bangerth:
This doesn't actually work, see https://jenkins.tjhei.info/blue/rest/organizations/jenkins/pipelines/dealii/branches/PR-9997/runs/3/nodes/52/steps/71/log/?start=0:
```
Test meshworker/scratch_data_02.debug: RUN
===============================   OUTPUT BEGIN  ===============================
[ 83%] Generating scratch_data_02.debug/interrupt_guard.cc
Scanning dependencies of target scratch_data_02.debug
[300%] Building CXX object CMakeFiles/scratch_data_02.debug.dir/scratch_data_02.cc.o
Building CXX object CMakeFiles/scratch_data_02.debug.dir/scratch_data_02.debug/interrupt_guard.cc.o
Linking CXX executable scratch_data_02.debug/scratch_data_02.debug
Built target scratch_data_02.debug
Scanning dependencies of target scratch_data_02.debug.diff
Generating scratch_data_02.debug/output
meshworker/scratch_data_02.debug: BUILD successful.
meshworker/scratch_data_02.debug: RUN failed. ------ Return code 134
meshworker/scratch_data_02.debug: RUN failed. ------ Result: /home/dealii/build/tests/meshworker/scratch_data_02.debug/failing_output
meshworker/scratch_data_02.debug: RUN failed. ------ Partial output:

meshworker/scratch_data_02.debug: RUN failed. ------ Additional output on stdout/stderr:


An error occurred in line <2728> of file </jenkins/workspace/dealii/PR-9997/include/deal.II/base/table.h> in function
    typename dealii::AlignedVector<T>::reference dealii::Table<2, T>::operator()(dealii::Table<2, T>::size_type, dealii::Table<2, T>::size_type) [with T = double; typename dealii::AlignedVector<T>::reference = double&; dealii::Table<2, T>::size_type = long unsigned int]
The violated condition was: 
    static_cast<typename ::dealii::internal::argument_type<void( typename std::common_type<decltype(i), decltype(this->table_size[0])>::type)>:: type>(i) < static_cast<typename ::dealii::internal::argument_type<void( typename std::common_type<decltype(i), decltype(this->table_size[0])>::type)>:: type>(this->table_size[0])
Additional information: 
    Index 0 is not in the half-open range [0,0). In the current case, this half-open range is in fact empty, suggesting that you are accessing an element of an empty collection such as a vector that has not been set to the correct size.

Aborted (core dumped)
```
It fails a whole bunch of tests:
```
The following tests FAILED:
	  1 - meshworker/scratch_data_02.debug (Failed)
	  3 - meshworker/scratch_data_01.debug (Failed)
	  5 - meshworker/mesh_loop_07.debug (Failed)
	  6 - meshworker/scratch_data_03.debug (Failed)
	  9 - meshworker/mesh_loop_06.debug (Failed)
	 12 - meshworker/mesh_loop_04.debug (Failed)
Errors while running CTest
```

bangerth:
The fact that this compiles means that some zero in a `copy_data = 0;` statement is probably implicitly converted to a constructor argument that is then handed to the copy `operator=`. Here's the code in that test:
```
  auto cell_worker =
    [&rhs_function](const Iterator &cell, ScratchData &s, CopyData &c) {
      const auto &fev = s.reinit(cell);
      const auto &JxW = s.get_JxW_values();
      const auto &p   = s.get_quadrature_points();

      c = 0;
```

`CopyData` is declared like this here:
```
  template <int n_matrices    = 1,
            int n_vectors     = n_matrices,
            int n_dof_indices = n_matrices>
  struct CopyData
  {
    /**
     * Initialize everything with the same @p size. This is usually the number
     * of local degrees of freedom.
     */
    CopyData(const unsigned int size);
```

So I think that if you want to outlaw the `operator=(double)`, you probably want to make the conversion constructor `explicit` to avoid the accidental conversion.

jppelteret:
Thanks for the assessment @bangerth. Making the constructor explicit sounds like a good idea -- I'll give it a go! Do you know offhand if this is permitted?
```
template<typename T, std::enable_if<std::is_arithmetic<T>::value>::type>
operator=(const T) = delete;
```
I think that we'd want to disallow any permutation of `= 0`, =`0.0`, `=0.0f`, ...

bangerth:
I think you can do that, but why would you? If you make the conversion constructor explicit, no conversion will happen of the right hand side to a `CopyData` object, and since you removed the conversion copy operator, the only copy operator that's left is the one that takes a `CopyData` object. The only reason why you might want to do what you're doing here is if you wanted to leave the conversion constructor non-`explicit`, and want to ensure that arithmetic types aren't converted by non-arithmetic types may be -- i.e., you want to short-circuit the conversion for certain types.

So you're thinking of something like this: https://godbolt.org/z/NsXmwy But I don't think that's a generally useful idiom. It's much better to just *always* make the conversion constructor `explicit`.

jppelteret:
Thanks @bangerth. Let's see if I've managed to get it right this time.

PR #9999, opened 2020-05-01
Convert steps 4,5,6 to FEValues::quadrature_point_indices().

bangerth:


bangerth:
Follow-up to #9987.

Y'all are sitting on your hands getting all of the good PR numbers. I will be content with 9999 and let someone else have 10,000 :-)

/rebuild

PR #10001, opened 2020-05-01
Remove quartic complexity from create_coupling_mass_matrix

### LINKED ISSUE #9989: create_coupling_mass_matrix too slow
The test `non_matching/coupling_08.debug` regularly times out on some machines, like this one:
https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=5877
The test runs a ridiculously small problem with 125 and 729 dofs, so I think we should be able to reduce the work. What I don't like in the code is this part:
https://github.com/dealii/dealii/blob/ea67b44dc1891da7aab1648f6c6f71143f7c7803/source/non_matching/coupling.cc#L850-L870
That's four nested loops. I have not spent time on thinking through the dependencies of the loop, but my feeling suggests this should be doable with three nested loops. Anyone with an idea? The other thing that comes to mind to calm down the work in the 3D/3D case is to reduce the quadrature formula to only 2 points per direction rather than 3 in that particular test - but again, I have not looked into whether that is possible/necessary.

kronbichler:
Fixes #9989.

This replaces the four nested loops by only three nested loops by pulling out some factors that are constant in one loop to a separate one.

I tested all non-matching tests locally and got no error, so I believe this is correct - nonetheless, I would like someone with knowledge of the algorithms to look into it.

### DIFF
--- a/source/non_matching/coupling.cc
+++ b/source/non_matching/coupling.cc
@@ -851,27 +851,29 @@ namespace NonMatching
         {
           kernel.set_center(fev0.quadrature_point(q0));
           kernel.value_list(fev1.get_quadrature_points(), kernel_values);
-          for (unsigned int q1 = 0; q1 < quadrature1.size(); ++q1)
-            if (kernel_values[q1] != 0.0)
-              {
-                for (unsigned int i = 0; i < fe0.dofs_per_cell; ++i)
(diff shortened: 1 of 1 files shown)


tjhei:
excellent, thank you!

Issue #8905, opened 2019-10-14
Matrix-Free DG with hanging nodes over periodic faces


kronbichler:
As reported on the mailing list at https://groups.google.com/forum/#!topic/dealii/okNWVIHwcTA , there is a bug in the setup the matrix free face pair in case we have hanging nodes over periodic faces, i.e., a level difference of one over periodic faces. To reproduce, change the initialization from `step-59` into
```
            Point<dim> upper_right;
            upper_right[0] = 2.5;
            for (unsigned int d = 1; d < dim; ++d)
              upper_right[d] = 2.8;
            GridGenerator::hyper_rectangle(triangulation,
                                           Point<dim>(),
                                           upper_right);
            triangulation.begin_active()->face(0)->set_boundary_id(10);
            triangulation.begin_active()->face(1)->set_boundary_id(11);
            triangulation.begin_active()->face(2)->set_boundary_id(0);
            for (unsigned int f = 3; f < GeometryInfo<dim>::faces_per_cell; ++f)
              triangulation.begin_active()->face(f)->set_boundary_id(1);

            std::vector<GridTools::PeriodicFacePair<
              typename Triangulation<dim>::cell_iterator>>
              periodic_faces;
            GridTools::collect_periodic_faces(
              triangulation, 10, 11, 0, periodic_faces);
            triangulation.add_periodicity(periodic_faces);

            triangulation.refine_global(1);
            triangulation.begin_active()->set_refine_flag();
            triangulation.execute_coarsening_and_refinement();
          
        setup_system();
```
(slightly reduced from the post on the mailing list.)

mwichro:
This is not related periodic faces, I just forgotten to remove some code from step-59. The following code also reproduces the problem:
```
            Point<dim> upper_right;
            upper_right[0] = 2.5;
            for (unsigned int d = 1; d < dim; ++d)
              upper_right[d] = 2.8;
            GridGenerator::hyper_rectangle(triangulation,
                                           Point<dim>(),
                                           upper_right);
            triangulation.begin_active()->face(0)->set_boundary_id(10);
            triangulation.begin_active()->face(1)->set_boundary_id(11);
            triangulation.begin_active()->face(2)->set_boundary_id(0);
            for (unsigned int f = 3; f < GeometryInfo<dim>::faces_per_cell; ++f)
              triangulation.begin_active()->face(f)->set_boundary_id(1);

            triangulation.refine_global(1);

[...]

PR #8983, opened 2019-10-31
Fix Matrixfree for periodic faces and hanging nodes

peterrum:
ping @kronbichler @lpsaavedra

PR #10004, opened 2020-05-01
Better document missing refinement edge data in MF for DG

kronbichler:
This PR slightly improves the message from #8910: We do support multigrid on adaptive meshes if face integrals are requested in case the user only intends to use face integrals on the boundary and not interior faces, so I moved the assertion to bail out to the appropriate place.

Related to #8905.


tjhei:
Out of curiosity and because I am not sure I fully understand this (or #8905): this is unrelated to periodic boundaries and we run into this for any DG multigrid with adaptivity?

peterrum:
> Out of curiosity and because I am not sure I fully understand this (or #8905): this is unrelated to periodic boundaries and we run into this for any DG multigrid with adaptivity?

Indeed, periodicity with hanging nodes is a separate issue and has been fixed in #8983.

PR #10007, opened 2020-05-01
Remove some unneeded backslashes.

drwells:
MathJax doesn't need to escape underscores in text mode: you can see this at

https://www.dealii.org/current/doxygen/deal.II/group__LAOperators.html#gad5b7eee8927ffd58475199d70e445b16

### DIFF
--- a/include/deal.II/lac/linear_operator.h
+++ b/include/deal.II/lac/linear_operator.h
@@ -370,8 +370,8 @@ public:
  * @relatesalso LinearOperator
  *
  * Addition of two linear operators @p first_op and @p second_op given by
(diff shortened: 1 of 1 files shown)


bangerth:
But what about those who compile the documentation without mathjax (as is the default)?

bangerth:
For reference, on my local copy of the documentation -- using latex to compile formulas --, the backslash doesn't show up.

drwells:
Good question - let me take a look.

drwells:
This is really, really dumb but for whatever reason I can no longer compile the documentation without mathjax - I just get pictures of TeX formulas. I'm still trying to answer this...

tjhei:
Grab _formulas.tex in one of the subdirectories, remove the batchmode, and compile manually. You will find problems and errors easily.

PR #10012, opened 2020-05-02
Group cells by parents in matrix-free setup

### LINKED ISSUE #9689: Group cells for data locality in matrix-free setup
While making benchmarks on #8319 in combination with #9656 I realized that our current scheme of assigning some cells with the need for data exchange to different groups is pretty bad for data locality. Besides the simplification from #9683, putting all children of a cell nearby and making sure they end up in the same SIMD array (if wide enough) should be done at the very least. It is especially profitable for the patch #9656 that now detects that extruded meshes have similar Jacobians; this only really works when the SIMD lanes and cells are kept together.

kronbichler:
Fixes #9689.

This PR aims to group the cells with the same parent together in the matrix-free loops. Unfortunately, the code turned out to be more complicated than I had hoped. The problem is that the categorization of cells (that allows to batch together certain cells by the user to e.g. have the same boundary on all cells of a batch) interferes with this purely performance-oriented setup.

My solution has been to add an additional layer that, within each "category" (or all cells if there is no category), fills data from two "pipes": One consists of cells that have the same parent, and one that contains all other cells.

I have given this quite some thought, but I am not happy with the code because it is rather complicated. @peterrum you often have ideas for code simplification - can you have a look here? One thing that one could still do (but I did not bother to address, because it would have made the code even more complex with the current strategy) is the fact that we could interleave batches of different category as long as they are filled up.

I still need to address the tests that depend on the order of how the cells are passed through, which, on my machine, include these:
```
	6851 - matrix_free/cell_categorization.mpirun=1.debug (Failed)
	6852 - matrix_free/cell_categorization.mpirun=1.release (Failed)
	6853 - matrix_free/cell_categorization.mpirun=7.debug (Failed)
	6854 - matrix_free/cell_categorization.mpirun=7.release (Failed)
	6861 - matrix_free/cell_level_and_index.debug (Failed)
	6862 - matrix_free/cell_level_and_index.release (Failed)
	6881 - matrix_free/dg_pbc_02.mpirun=7.debug (Failed)
	6882 - matrix_free/dg_pbc_02.mpirun=7.release (Failed)
	6883 - matrix_free/dof_info_01.mpirun=1.debug (Failed)
	6884 - matrix_free/dof_info_01.mpirun=1.release (Failed)
	6885 - matrix_free/dof_info_02.mpirun=1.debug (Failed)
	6886 - matrix_free/dof_info_02.mpirun=1.release (Failed)
```
But as we might still touch the algorithms again, I wait until we have converged there.


peterrum:
@kronbichler Yes, I'll do. However, I would like to finish something: I am in the middle of debugging :(

kronbichler:
> BTW: Can you quantify the benefit of this change in regard of the performance?

It is difficult to say in general. If we take for example step-67 in 3D with 40 cores, I get these timings on master as
```
| rk time stepping total        |     17723 | 1.561e+04s     0 | 1.634e+04s | 1.679e+04s    33 |
| rk_stage - integrals L_h      |     88615 | 1.112e+04s    32 | 1.201e+04s | 1.289e+04s     9 |
| rk_stage - inv mass + vec upd |     88615 |      3059s    11 |      4331s |      5663s    32 |
```
and with this PR:
```
| rk time stepping total        |     17723 | 1.539e+04s     0 | 1.597e+04s | 1.627e+04s    33 |
| rk_stage - integrals L_h      |     88615 | 1.029e+04s    32 | 1.141e+04s | 1.273e+04s    11 |
| rk_stage - inv mass + vec upd |     88615 |      2859s    11 |      4563s |      5971s    32 |
```
So here we talk about 3-5%. This test case might exaggerate things a bit, though, because it mixes the data locality in the access into the neighbors with the compression of the geometry data. Recall that the 3D case is an extruded mesh - around the cylinder we have a repeating pattern in the z direction and having children cells sitting on the same cell batch increases the chances that we do get repeating data. I am not sure whether the compression makes the effect stronger.

To get a more precise data point, one should run another case, maybe step-37 or step-59, or some of our benchmarks.

peterrum:
@kronbichler Thanks! 

> This test case might exaggerate things a bit, though,

This is good enough!

kronbichler:
I decided that the current implementation was too convoluted. I now tried to make the decision upside down, i.e., first collect the different batches in a local way, and then determine if they belong to the part where we overlap communication and communication. Also, I changed the decision to take data from two queues into a simple "sort" algorithms, which is more appropriate as the code is not performance-critical. I hope this is more understandable - it still adds quite some new lines, but it is ~80 lines less than the intermediate version. I kept the new code as a separate commit as to make it more visible before I squash it. If we agree, I will proceed with adapting the tests.

kronbichler:
I updated the test cases now and triggered the CI. Should be ready for review now.

peterrum:
OK. I'll have a look in the morning.

source/matrix_free/task_info.cc
```diff
@@ -775,88 +768,42 @@ namespace internal
...
-          Assert(marker != 0, ExcInternalError());
-        }
-
+      // Give the compiler a chance to detect that vectorization_length is a
+      // power of two, which allows it to replace integer divisions by shifts
+      unsigned int vectorization_length_bits = 0;
+      unsigned int my_length                 = vectorization_length;
+      while (my_length >>= 1)
+        ++vectorization_length_bits;
+      const unsigned int n_lanes = 1 << vectorization_length_bits;
```
peterrum, line 786: I guess you are doing this here since the class is not templated with `VectorizedArrayType`. Wouldn't it be an option to make these lines of codes to some kind of utility function? It might be used in other classes where the length is not at compile time.

source/matrix_free/task_info.cc
```diff
@@ -866,93 +813,222 @@ namespace internal
...
+          // Sort the remaining cells and append them as well
+          std::sort(other_cells.begin(), other_cells.end());
+          temporary_numbering.insert(temporary_numbering.end(),
+                                     other_cells.begin(),
+                                     other_cells.end());
+
+          while (temporary_numbering.size() % n_lanes != 0)
+            temporary_numbering.push_back(numbers::invalid_unsigned_int);
+
+          category_size.push_back(temporary_numbering.size());
```
peterrum, line 905: If I read this code correctly, this code enumerates cells differently than in the version before. But this is fixed with the sort below?

source/matrix_free/task_info.cc
kronbichler: Yes, the reason is that we do not have a template and do not want it just for that parameter. We could make it a utility function (that is always inlined). I do not have time now to look for other uses - do you want to open an issue?
If someone else reads this: an integer division takes  32 clock cycles of latency and 8 cycles, whereas a right shift 1-3 clock cycles. At least smart compilers can profit from this.

source/matrix_free/task_info.cc
kronbichler: Yes, we enumerate cells differently now (which is why the tests sensitive to those have changed).
Here we first place the "structured" part and the "unstructured" one. The sort below allows to set priorities on how to arrange cell batches when combining them. Right now we have two priorities:
- In the hp case, cells with the same degree are placed together (because the tests depend on that for now; it could be relaxed, though)
- We keep the cell batches sorted by the index of the last cell. This makes sure that we make a good decision how the two queues are combined.
Do you think we should add something like that as an overall comment?

peterrum: > Do you think we should add something like that as an overall comment?

Comments are always good! It took me a couple of seconds to understand why `std::array<..., 3>`. I like this approach it (and the padding with `numbers::invalid_unsigned_int`) should more or less have the same effect as the stat machine from the weekend.

> Yes, we enumerate cells differently now

My question was referring to the version on the weekend. But anyhow, I am fine with the changes.

kronbichler:
Test failure is unrelated. Any more comments? If not, I will merge later today.

PR #10013, opened 2020-05-02
Introduced p::d::ErrorPredictor.

marcfehling:
Predicts error after adaptation based on future_fe_indices and CellStatus and transfers it to the adapted mesh.

Initially the combination of `hp::Refinement::predict_error()` and `parallel::distributed::CellDataTransfer` should suffice. As discussed in #9954, we have an issue with the p4est oracle, since h-adaptation flags between the deal.II triangulation and the p4est mesh might differ (see #8647). This class acts as an intermediate solution for this problem once we figured out #8647.

### DIFF
--- a/include/deal.II/distributed/error_predictor.h
+++ b/include/deal.II/distributed/error_predictor.h
@@ -1,0 +1,254 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 - 2020 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_distributed_error_predictor_h
+#define dealii_distributed_error_predictor_h
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/base/smartpointer.h>
+
+#include <boost/range/iterator_range.hpp>
+
+#include <vector>
+
+
+DEAL_II_NAMESPACE_OPEN
+
+
+// forward declarations
+#ifndef DOXYGEN
+template <typename Number>
+class Vector;
+
+namespace parallel
+{
+  namespace distributed
+  {
+    template <int dim, int spacedim>
+    class Triangulation;
+  }
+} // namespace parallel
+
+namespace hp
+{
+  template <int dim, int spacedim>
+  class DoFHandler;
+}
+#endif
+
+
+namespace parallel
+{
+  namespace distributed
+  {
+    /**
+     * Predict how current error indicators will change after refinement and
+     * coarsening were to happen on the provided hp::DoFHandler in context of a
+     * parallel::distributed::Triangulation.
+     *
+     * This algorithm follows the same logic as the error prediction algorithm
+     * presented by @cite melenk2001hp and has been implemented in deal.II via
+     * the function hp::Refinement::predict_error(). See the documentation of
+     * this particular function for details on the algorithm and all featured
+     * formulas.
+     *
+     * In theory, the combination of hp::Refinement::predict_error() to predict
+     * how the error will change with  parallel::distributed::CellDataTransfer
+     * to transfer data across meshes forms the intended way to apply the
+     * prediction algorithm. However, p4est determines the details of grid
+     * refinement in the parallel distributed case; consequently, it yields more
+     * reliable and trustworthy results when we determine the predicted errors
+     * during the adaptation process. This is achieved with this class, similar
+     * to the parallel::distributed::SolutionTransfer and
+     * parallel::distributed::CellDataTransfer classes.
+     *
+     * Before adaptation happens, one or multiple vectors containing error
+     * estimates for every cell have to be registered with this class using the
+     * prepare_for_coarsening_and_refinement() functions. Performing adaptation
+     * on the triangulation initiates the calculation of the error prediction
+     * based on how p4est performed grid adaptation. After the refinement
+     * process, the predicted errors need to be unpacked on the updated mesh via
+     * one of the unpack() functions.
+     *
+     * The following code snippet demonstrates how to impose hp-adaptivity based
+     * on refinement history in a parallel distributed application:
+     * @code
+     * // [initialization...]
+     * Vector<float> predicted_error_per_cell(triangulation.n_active_cells());
+     * for(unsigned int i = 0; i < triangulation.n_active_cells(); ++i)
+     *   predicted_error_per_cell[i] = std::numeric_limits<float>::infinity();
+     *
+     * // [during each refinement step...]
+     * // set h-adaptivity flags
+     * Vector<float> estimated_error_per_cell(triangulation.n_active_cells());
+     * KellyErrorEstimator<dim>::estimate(...);
+     * parallel::distributed::GridRefinement::
+     *   refine_and_coarsen_fixed_{number|fraction}(...);
+     *
+     * // set p-adaptivity flags
+     * hp::Refinement::p_adaptivity_from_reference(
+     *   hp_dof_handler,
+     *   estimated_error_per_cell,
+     *   predicted_error_per_cell
+     *   std::less<float>(),
+     *   std::less<float>());
+     * hp::Refinement::{choose|force}_p_over_h(hp_dof_handler);
+     *
+     * // perform adaptation and predict error for the subsequent adaptation
+     * parallel::distributed::ErrorPredictor<dim> predictor(hp_dof_handler);
+     * predictor.prepare_for_coarsening_and_refinement(estimated_error_per_cell);
+     * triangulation.execute_coarsening_and_refinement();
+     *
+     * // unpack predicted errors
+     * predicted_error_per_cell.reinit(triangulation.n_active_cells());
+     * predictor.unpack(predicted_error_per_cell);
+     * @endcode
+     *
+     * For serialization of predicted errors, use the
+     * parallel::distributed::CellDataTransfer class.
+     *
+     * @ingroup distributed
+     * @author Marc Fehling, 2019 - 2020
+     */
+    template <int dim, int spacedim = dim>
+    class ErrorPredictor
+    {
+    public:
+      /**
+       * Constructor.
+       *
+       * @param[in] dof The hp::DoFHandler on which all operations will
+       *   happen. At the time when this constructor is called, the
+       *   DoFHandler still points to the triangulation before the
+       *   refinement in question happens.
+       */
+      ErrorPredictor(const hp::DoFHandler<dim, spacedim> &dof);
+
+      /**
+       * Prepare the current object for coarsening and refinement.
+       * @p all_in includes all vectors that are to be interpolated onto the
+       * new (refined and/or coarsened) grid.
+       *
+       * As specified in the error prediction algorithm in
+       * hp::Refinement::predict_error(), the control parameters @p gamma_p,
+       * @p gamma_h, @p gamma_n can be used to influence the predicted errors in
+       * case of p-adaptation, h-adaptation, and no adapation, respectively.
+       * Their default values comply to the studies of @cite melenk2001hp.
+       */
+      void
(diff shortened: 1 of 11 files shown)


kronbichler:
@bangerth do you have resources to review this? The CI seems fine, but I have limited time to have a closer look. @tamiko should we make this a candidate for 9.2-rc2?

bangerth:
Will do tonight after dinner!

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+}
+#endif
+
+
+namespace parallel
+{
+  namespace distributed
+  {
+    /**
+     * Predict how current error indicators will adapt after refinement and
```
bangerth, line 58: Let's use the word "change" to avoid the connotation with mesh adaptation:
```suggestion
     * Predict how current error indicators will change after refinement and
```

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     * coarsening would have happened on the provideded hp::DoFHandler in
```
bangerth, line 59:
```suggestion
     * coarsening would happen on the provided hp::DoFHandler in
```

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     * context of a parallel::distributed::Triangulation.
+     *
+     * This algorithm follows the same logic as hp::Refinement::predict_error().
+     * See its documentation for details on the algorithm.
+     *
+     * In theory, the combination of hp::Refinement::predict_error() to predict
+     * how the error will change with  parallel::distributed::CellDataTransfer
+     * to transfer data across meshes forms the intended way to apply the
+     * prediction algorithm. However, since p4est specifies grid refinement in
```
bangerth, line 68:
```suggestion
     * prediction algorithm. However, p4est determines the details of grid refinement in
```

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     * the parallel distributed case, it yields more reliable and trustworthy
```
bangerth, line 69:
```suggestion
     * the parallel distributed case; consequently, it yields more reliable and trustworthy
```

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     * results when we determine the predicted errors during adaptation process.
```
bangerth, line 70:
```suggestion
     * results when we determine the predicted errors during the adaptation process.
```

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     * This is achieved with this this class, similar to
```
bangerth, line 71:
```suggestion
     * This is achieved with this class, similar to
```

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     * parallel::distributed::SolutionTransfer and the aforementioned
+     * parallel::distributed::CellDataTransfer classes.
```
bangerth, line 73: well, actually it wasn't mentioned afore :-)

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     *
+     * The following code snippet demonstrates how to impose hp-adaptivity based
+     * on refinement history in a parallel distributed application:
+     * @code
+     * // [initialisation...]
```
bangerth, line 78:
```suggestion
     * // [initialization...]
```

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+     * hp::Refinement::{choose|force}_p_over_h(hp_dof_handler);
+     *
+     * // perform adaptation and predict error for the subsequent adaptation
+     * parallel::distributed::ErrorPredictor<dim> predictor(hp_dof_handler);
+     * predictor.prepare_for_coarsening_and_refinement(estimated_error_per_cell);
+     * triangulation.execute_coarsening_and_refinement();
+     *
+     * // unpack predicted errors
+     * predicted_error_per_cell.reinit(triangulation.n_active_cells());
+     * predictor.unpack(predicted_error_per_cell);
```
bangerth, line 106: So if I read this right, the output is a vector of predicted errors on the new mesh (presumably so that you can compare it with estimated errors on the new mesh).

I think this conclusion of what the *output* is wasn't quite clear to me from the description before. Can you add this in a sentence or two either after or before the code snippet.

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+       *   happen. At the time when this constructor is called, the
+       *   DoFHandler still points to the triangulation before the
+       *   refinement in question happens.
+       */
+      ErrorPredictor(const hp::DoFHandler<dim, spacedim> &dof);
+
+      /**
+       * Destructor.
+       */
+      ~ErrorPredictor() = default;
```
bangerth, line 132: Ah, but that's just five extra lines of code :-) Unless a class has virtual functions, there is no point in explicitly declaring a destructor that is `=default`.

include/deal.II/distributed/error_predictor.h
```diff
@@ -0,0 +1,242 @@
...
+       * Prepare the current object for coarsening and refinement.
+       * @p all_in includes all vectors that are to be interpolated onto the
+       * new (refined and/or coarsened) grid.
+       */
+      void
+      prepare_for_coarsening_and_refinement(
+        const std::vector<const Vector<float> *> &all_in,
+        const double                              gamma_p = std::sqrt(0.4),
+        const double                              gamma_h = 2.,
+        const double                              gamma_n = 1.);
```
bangerth, line 144: Can you document what these three parameters are, and where their default values come from?

include/deal.II/hp/refinement.h
```diff
@@ -382,7 +382,7 @@ namespace hp
 
     /**
      * Predict how the current @p error_indicators will adapt after refinement
-     * and coarsening has happened on the provided @p dof_handler, and write its
+     * and coarsening would have happened on the provided @p dof_handler, and write its
```
bangerth, line 385: maybe say "if refinement and coarsening were to happen". "would have happened" is past tense and considers something that might have happened in the past, but didn't. But you want to consider future tense: something that could happen in the future. "were to happen" is exactly the term you are looking for. (If you want to be extra explicit about the tense, you could say "were to happen in the future", though that means exactly the same.)

source/distributed/error_predictor.cc
```diff
@@ -0,0 +1,363 @@
...
+      parallel::distributed::Triangulation<dim, spacedim> *tria =
+        (dynamic_cast<parallel::distributed::Triangulation<dim, spacedim> *>(
+          const_cast<dealii::Triangulation<dim, spacedim> *>(
+            &dof_handler->get_triangulation())));
+      Assert(tria != nullptr, ExcInternalError());
+
+      tria->notify_ready_to_unpack(
+        handle,
+        [this, &all_out](
+          const typename Triangulation<dim, spacedim>::cell_iterator &cell_,
```
bangerth, line 140: any reason not to call the argument `cell` without the trailing underscore? (Same above and below, I think.)

source/distributed/error_predictor.cc
```diff
@@ -0,0 +1,363 @@
...
+          const typename Triangulation<dim, spacedim>::CellStatus     status,
+          const boost::iterator_range<std::vector<char>::const_iterator>
+            &data_range) {
+          this->unpack_callback(cell_, status, data_range, all_out);
+        });
+
+      for (typename std::vector<Vector<float> *>::iterator it = all_out.begin();
+           it != all_out.end();
+           ++it)
```
bangerth, line 149: I think you could do this more easily with a range-based for loop.

source/distributed/error_predictor.cc
```diff
@@ -0,0 +1,363 @@
...
+      std::vector<Vector<float> *> all_out(1, &out);
+      unpack(all_out);
+    }
+
+
+
+    template <int dim, int spacedim>
+    std::vector<char>
+    ErrorPredictor<dim, spacedim>::pack_callback(
+      const typename Triangulation<dim, spacedim>::cell_iterator &cell_,
```
bangerth, line 170: same here:
```suggestion
      const typename Triangulation<dim, spacedim>::cell_iterator &cell,
```

source/distributed/error_predictor.cc
```diff
@@ -0,0 +1,363 @@
...
+      const typename Triangulation<dim, spacedim>::CellStatus     status)
+    {
+      typename hp::DoFHandler<dim, spacedim>::cell_iterator cell(*cell_,
+                                                                 dof_handler);
+
+      // create buffer for each individual object
```
bangerth, line 176:
```suggestion
      // create buffer for each individual input vector
```

source/distributed/error_predictor.cc
```diff
@@ -0,0 +1,363 @@
...
+                // Then determine the actually contirbution to the predicted
+                // error of every single cells that is about to be coarsened.
+                float sqrsum_of_predicted_errors = 0.;
+                float predicted_error            = 0.;
+                int   degree_difference          = 0;
+                for (unsigned int child_index = 0;
+                     child_index < cell->n_children();
+                     ++child_index)
+                  {
+                    const auto child = cell->child(child_index);
```
bangerth, line 268: I guess we should introduce `cell->child_iterators()` :-) See #10206.

source/distributed/error_predictor.cc
```diff
@@ -0,0 +1,363 @@
...
+                      (gamma_h * std::pow(.5, future_fe_degree));
+
+                    degree_difference =
+                      future_fe_degree - child->get_fe().degree;
+
+                    if (degree_difference != 0)
+                      predicted_error *= std::pow(gamma_p, degree_difference);
+
+                    sqrsum_of_predicted_errors +=
+                      predicted_error * predicted_error;
```
bangerth, line 281: This and all of the similar formulas should be documented somewhere. Is this in the non-`p::d` class? Could we refer to these formulas in the place where the various gamma factors are passed in as argument? Or say something like "The exact details of how the error is predicted are described in the documentation of class X::Y."

source/distributed/error_predictor.cc
marcfehling: Most parts of the code have been taken over from the SolutionTransfer class. My guess is that `cell_` should denote `Triangulation::cell_iterators`, while `cell` stand for `DoFHandler::cell_iterators`. But you're right, the underscore is of no use here.

source/distributed/error_predictor.cc
marcfehling: We convert the `Triangulation::cell_iterator` to a `DoFHandler::cell_iterator` below, which will be assigned to that name. I would leave it like this here.

source/distributed/error_predictor.cc
marcfehling: https://github.com/dealii/dealii/blob/3bc6103128d0cf3ad01023273f827bd9615e2e8c/include/deal.II/distributed/error_predictor.h#L62-L63

I hoped that these lines would suffice. I'll be more verbose on that.

marcfehling:
Applied suggestions.

source/distributed/error_predictor.cc
```diff
@@ -0,0 +1,360 @@
...
+        [this, &all_out](
+          const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+          const typename Triangulation<dim, spacedim>::CellStatus     status,
+          const boost::iterator_range<std::vector<char>::const_iterator>
+            &data_range) {
+          this->unpack_callback(cell, status, data_range, all_out);
+        });
+
+      for (const auto &out : all_out)
+        out->compress(::dealii::VectorOperation::insert);
```
marcfehling, line 147: Is the `const` flag correct here?
`all_out` is of type `std::vector<Vector<float> *>`, so I guess a const reference to an immutable pointer should be okay.

bangerth: If it compiles, it is ok :-)

bangerth:
Thanks!

bangerth:
@tamiko : I don't have a 9.2 branch set up right now. Can you pull this onto 9.2?

PR #10018, opened 2020-05-02
Introduce std_cxx20::ranges::iota_view

### LINKED ISSUE #10000: Create alias for std::ranges::views::iota_view
We have started to use `boost::irange` in a number of places. For example, #9990 and #9987 add members to `FEValues`. `GeometryInfo` gained some uses in #9415 and #9622.

But I think of `boost::irange` as only a temporary solution -- C++20 has `std::ranges::views::iota_view`. I wonder whether we should do as we've done in other cases: Put a header file into `include/deal.II/base/std_cxx20` that either imports `std::ranges::views::iota_view` or defines it in terms of `boost::irange`. This way, we would get to use the new class as we've done with `std::shared_ptr` and `std::unique_ptr` before.

Opinions?

bangerth:
This would fix #10000. It introduces an alias for `std::ranges::iota_view` (see https://en.cppreference.com/w/cpp/ranges/iota_view) using the `boost::integer_range` class that does something sufficiently similar. 

Strictly speaking, `std::ranges::iota_view` is more general since it doesn't just allow integer ranges but in fact ranges over anything that has an `operator++` and this, in particular, includes, iterator ranges. But as a first version, this is good enough for everything we want to do and allows us to use a name that will only require marginal adaptation once we actually use C++20.

I don't have a compiler that already supports C++20, and so haven't added any standard alias as we do in the other `std_cxx??` headers. We can add that at a later time -- for now, we just unconditionally refer to the corresponding boost class.

/rebuild

### DIFF
--- a/include/deal.II/base/std_cxx20/iota_view.h
+++ b/include/deal.II/base/std_cxx20/iota_view.h
@@ -1,0 +1,48 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2020 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+#ifndef dealii_cxx20_iota_view_h
+#define dealii_cxx20_iota_view_h
+
+#include <deal.II/base/config.h>
+
+#include <boost/range/irange.hpp>
+
+
+DEAL_II_NAMESPACE_OPEN
+
+namespace std_cxx20
+{
+  namespace ranges
+  {
+    /**
+     * A poor-man's implementation of std::ranges::iota_view using
+     * boost's integer_range class. The two classes are not completely
+     * identical, but close enough so that we can use the boost
+     * class until std::ranges::iota_range becomes available in
+     * compilers.
+     *
+     * The std::ranges::iota_view class represents a range of values
+     * or iterators that one can loop over. A documentation of this
+     * class can be found at
+     * https://en.cppreference.com/w/cpp/ranges/iota_view .
+     */
+    template <typename IncrementableType>
+    using iota_view = boost::integer_range<IncrementableType>;
+  } // namespace ranges
+} // namespace std_cxx20
+
+DEAL_II_NAMESPACE_CLOSE
+
+#endif
--- a/include/deal.II/base/std_cxx20/iota_view.h
+++ b/include/deal.II/base/std_cxx20/iota_view.h
@@ -17,8 +17,11 @@
 
 #include <deal.II/base/config.h>
 
-#include <boost/range/irange.hpp>
-
+#if defined __cpp_lib_ranges && __cpp_lib_ranges >= 201911
+#  include <ranges>
+#else
+#  include <boost/range/irange.hpp>
+#endif
 
 DEAL_II_NAMESPACE_OPEN
 
@@ -26,6 +29,10 @@ namespace std_cxx20
 {
   namespace ranges
   {
+#if defined __cpp_lib_ranges && __cpp_lib_ranges >= 201911
+    template <typename IncrementableType, typename BoundType>
+    using iota_view = std::ranges::iota_view<IncrementableType, BoundType>;
+#else
     /**
      * A poor-man's implementation of std::ranges::iota_view using
      * boost's integer_range class. The two classes are not completely
@@ -38,8 +45,9 @@ namespace std_cxx20
      * class can be found at
      * https://en.cppreference.com/w/cpp/ranges/iota_view .
      */
-    template <typename IncrementableType>
+    template <typename IncrementableType, typename /*BoundType*/>
     using iota_view = boost::integer_range<IncrementableType>;
+#endif
   } // namespace ranges
 } // namespace std_cxx20
 
--- a/include/deal.II/base/std_cxx20/iota_view.h
+++ b/include/deal.II/base/std_cxx20/iota_view.h
@@ -30,8 +30,7 @@ namespace std_cxx20
   namespace ranges
   {
 #if defined __cpp_lib_ranges && __cpp_lib_ranges >= 201911
-    template <typename IncrementableType, typename BoundType>
-    using iota_view = std::ranges::iota_view<IncrementableType, BoundType>;
+    using std::ranges::iota_view;
 #else
     /**
      * A poor-man's implementation of std::ranges::iota_view using
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -1250,16 +1250,6 @@ namespace Step69
     // to carry out (computation and storage of normals, and normalization
     // of the $\mathbf{c}_{ij}$ of entries) threads cannot conflict
     // attempting to write the same entry (we do not need a scheduler).
-    //
-    // We have one more difficulty to overcome: In order to implement the
-    // <code>on_subranges</code> lambda we need to name the iterator type
-    // of the object returned by <code>std_cxx20::ranges::iota_view%<unsigned
-    // int%>()</code>. This is unfortunately a very convoluted name exposing
-    // implementation details about <code>std_cxx20::ranges::iota_view</code>.
-    // For this reason we resort to the <a
-    // href="https://en.cppreference.com/w/cpp/language/decltype"><code>decltype</code></a>
-    // specifier, a C++11 feature that returns the type of an entity, or
-    // expression.
     {
       TimerOutput::Scope scope(computing_timer,
                                "offline_data - compute |c_ij|, and n_ij");
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -2102,8 +2102,8 @@ namespace Step69
                                "time_stepping - 3 perform update");
 
       const auto on_subranges =
-        [&](typename decltype(indices_owned)::iterator       i1,
-            const typename decltype(indices_owned)::iterator i2) {
+        [&](std_cxx20::ranges::iota_view<unsigned int>::iterator       i1,
+            const std_cxx20::ranges::iota_view<unsigned int>::iterator i2) {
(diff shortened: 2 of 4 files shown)


bangerth:
@tamiko and Ignacio: I should add that your step-69 uses this class and so I've edited that as well. While there, I realized that we can avoid some unnecessary `decltype` uses that we can easily replace by their explicit types.

include/deal.II/base/std_cxx20/iota_view.h
```diff
@@ -0,0 +1,48 @@
...
+#define dealii_cxx20_iota_view_h
+
+#include <deal.II/base/config.h>
+
+#include <boost/range/irange.hpp>
+
+
+DEAL_II_NAMESPACE_OPEN
+
+namespace std_cxx20
```
tjhei, line 25: can you please add a (short) doxygen documentation to the two namespaces so that they appear in the documentation?

bangerth:
Let me do that in a follow-up. I wanted to write more about the various `std_cxx*` namespaces anyway. So if you're happy with what's here, I'd appreciate merging as is.

masterleinad:
I still don't understand why we need to rush this into the next release before we have proper implementations for it. I would rather wait for after release and face possible issues then when we compare to compiler implementations.

Anyway, just trying with https://godbolt.org/z/3AFTWW shows that providing just one template parameter doesn't work as expected. In this case, the constructor expects one parameter to create an unbounded range.
Hence, I would at least introduce the same two template parameters `std::ranges::iota_view` has for forward-compatibility (of course without the concept restrictions). We could, of course, enforce them to be the same.
I would also advocate for including
```
#if defined __cpp_lib_ranges && __cpp_lib_ranges>=201911
  using iota_view = std::ranges::iota_view<IncrementableType, BoundType>;
#endif
```
for a pull request introducing this wrapper.

masterleinad:
As a side comment, I need
```diff
diff --git a/examples/step-69/step-69.cc b/examples/step-69/step-69.cc
index 633efff4e3..7d7d2a6ffe 100644
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -1258,8 +1258,8 @@ namespace Step69
         0, n_locally_relevant);
 
       const auto on_subranges = //
-        [&](std_cxx20::ranges::iota_view<unsigned int>::iterator       i1,
-            const std_cxx20::ranges::iota_view<unsigned int>::iterator i2) {
+        [&](std_cxx20::ranges::iota_view<unsigned int>::_Iterator       i1,
+            const std_cxx20::ranges::iota_view<unsigned int>::_Iterator i2) {
           for (const auto row_index :
                std_cxx20::ranges::iota_view<unsigned int>(*i1, *i2))
             {
@@ -1920,8 +1920,8 @@ namespace Step69
                                "time_stepping - 1 compute d_ij");
 
       const auto on_subranges = //
-        [&](std_cxx20::ranges::iota_view<unsigned int>::iterator       i1,
-            const std_cxx20::ranges::iota_view<unsigned int>::iterator i2) {
+        [&](std_cxx20::ranges::iota_view<unsigned int>::_Iterator       i1,
+            const std_cxx20::ranges::iota_view<unsigned int>::_Iterator i2) {
           for (const auto i :
                std_cxx20::ranges::iota_view<unsigned int>(*i1, *i2))
             {
@@ -2016,8 +2016,8 @@ namespace Step69
       // locally.
 
       const auto on_subranges = //
-        [&](std_cxx20::ranges::iota_view<unsigned int>::iterator       i1,
-            const std_cxx20::ranges::iota_view<unsigned int>::iterator i2) {
+        [&](std_cxx20::ranges::iota_view<unsigned int>::_Iterator       i1,
+            const std_cxx20::ranges::iota_view<unsigned int>::_Iterator i2) {
           double tau_max_on_subrange = std::numeric_limits<double>::infinity();
 
           for (const auto i :
@@ -2102,8 +2102,8 @@ namespace Step69
                                "time_stepping - 3 perform update");
 
       const auto on_subranges =
-        [&](std_cxx20::ranges::iota_view<unsigned int>::iterator       i1,
-            const std_cxx20::ranges::iota_view<unsigned int>::iterator i2) {
+        [&](std_cxx20::ranges::iota_view<unsigned int>::_Iterator       i1,
+            const std_cxx20::ranges::iota_view<unsigned int>::_Iterator i2) {
           for (const auto i : boost::make_iterator_range(i1, i2))
             {
               Assert(i < n_locally_owned, ExcInternalError());
@@ -2352,8 +2352,8 @@ namespace Step69
     // global maxima and minima of the gradients.
     {
       const auto on_subranges = //
-        [&](std_cxx20::ranges::iota_view<unsigned int>::iterator       i1,
-            const std_cxx20::ranges::iota_view<unsigned int>::iterator i2) {
+        [&](std_cxx20::ranges::iota_view<unsigned int>::_Iterator       i1,
+            const std_cxx20::ranges::iota_view<unsigned int>::_Iterator i2) {
           double r_i_max_on_subrange = 0.;
           double r_i_min_on_subrange = std::numeric_limits<double>::infinity();
 
@@ -2437,8 +2437,8 @@ namespace Step69
 
     {
       const auto on_subranges = //
-        [&](std_cxx20::ranges::iota_view<unsigned int>::iterator       i1,
-            const std_cxx20::ranges::iota_view<unsigned int>::iterator i2) {
+        [&](std_cxx20::ranges::iota_view<unsigned int>::_Iterator       i1,
+            const std_cxx20::ranges::iota_view<unsigned int>::_Iterator i2) {
           for (const auto i : boost::make_iterator_range(i1, i2))
             {
               Assert(i < n_locally_owned, ExcInternalError());
```
to be able to compile with a pre-relesase gcc-10 version to be able to compile with `std::ranges::iota_view`. It's clearly not quite complete, yet. `iterator` should be defined.

bangerth:
Ugh.

First for the rationale: I was hoping that by introducing the `typedef` now, we won't have to change the return type again later. There is also a benefit to documentation because people can look up what `std::ranges::iota_view` is. Introducing the second template argument is a simple change I'd be happy to make of course.

The question of course is what others think about this?


Second: As for the changes from `iterator` to `_Iterator` -- my patch doesn't actually use `std::ranges::iota_view`, of course. It uses the boost class unconditionally, until I actually have a way to test, and that might be a couple of years away. For you to see that you needed `_Iterator`, you must have also changes the `.h` file, right? My proposal would be to only replace the boost class by the `std` class (in that hypothetical, future patch) when we can ascertain that the compiler implementation is actually correct. Unless that is the case, we'll just stick with the boost class.

tjhei:
> My proposal would be to only replace the boost class by the `std` class (in that hypothetical, future patch) when we can ascertain that the compiler implementation is actually correct. Unless that is the case, we'll just stick with the boost class.

But I agree with @masterleinad that we should check if this change can be made in the future (and make necessary changes like the second template parameter now). If we keep using the boost version for now, we can ignore the missing ``iterator`` template in the gcc trunk of course, because it will be present in the future.

masterleinad:
Thanks for explaining your rationale!

Yes, I changed the header to use the C++20 classes and include headers.
My only concern here is that people rely on this wrapper to work the same as `std::ranges::iota_view` but we will discover later that it doesn't and we would need to change it in way that is not backward-compatible.

After trying with that `gcc` version, I am reasonably confident that it will work with the actual class, though. I am happy to push a corresponding commit.

masterleinad:
Apart from the `_Iterator` issue the commit I pushed works with trunk `gcc`.

include/deal.II/base/std_cxx20/iota_view.h
```diff
@@ -17,15 +17,22 @@
...
 
 DEAL_II_NAMESPACE_OPEN
 
 namespace std_cxx20
 {
   namespace ranges
   {
+#if defined __cpp_lib_ranges && __cpp_lib_ranges >= 201911
+    template <typename IncrementableType, typename BoundType>
+    using iota_view = std::ranges::iota_view<IncrementableType, BoundType>;
```
bangerth, line 34: Wouldn't it be possible to just do
```
  using std::ranges::iota_view;
```
(i.e., a traditional `using` declaration rather than a template-alias)?

bangerth:
@masterleinad -- thanks for the additional commit. I had intended to get back to it today but got bogged down with meetings and trying to write a Gutachten for @marcfehling 's PhD thesis :-)

include/deal.II/base/std_cxx20/iota_view.h
masterleinad: No, that doesn't work. https://godbolt.org/z/rKH2ur

bangerth: Right, but that's not what I meant. You're using the form
```
  using A = B;
```
which acts like a typedef and requires that `B` is a fully qualified name (unless you put `template` in front of it). But I want the name injection:
```
namespace X{
template <typename T>
struct Foo{};
}
using X::Foo;

int main()
{
  Foo<int> foo;
}
```
which actually compiles: https://godbolt.org/z/4A5yKf

masterleinad: Yes, that works.

bangerth:
Since the two people who discussed this PR are now also both authors, someone else will have to review/approve/merge...

bangerth:
@dealii/dealii -- ping?

include/deal.II/base/std_cxx20/iota_view.h
```diff
@@ -0,0 +1,55 @@
...
+namespace std_cxx20
+{
+  namespace ranges
+  {
+#if defined __cpp_lib_ranges && __cpp_lib_ranges >= 201911
+    using std::ranges::iota_view;
+#else
+    /**
+     * A poor-man's implementation of std::ranges::iota_view using
+     * boost's integer_range class. The two classes are not complete
```
kronbichler, line 37:
```suggestion
     * boost's integer_range class. The two classes are not completely
```

include/deal.II/base/std_cxx20/iota_view.h
masterleinad: Documentation for that namespace is part of https://github.com/dealii/dealii/pull/10047.

bangerth:
@kronbichler Change made!

masterleinad:
You need to rebase.

bangerth:
Done.

kronbichler:
@tamiko?

masterleinad:
The test failures are spurious.

bangerth:
Looks like the two failures are spurious. At least there's no output with the two failed checks...

bangerth:
yay :-)

PR #10021, opened 2020-05-03
New extract_rtree_level function, and RTree visitor.

luca-heltai:
Created a new function that extracts all bonding boxes at a certain level of a boost::rtree.

The following image shows the first level of an Rtree containing all active cells in a triangulation with ~50000 elements.

image

This is a more robust replacement of the function `GridTools::compute_mesh_predicate_bounding_box`, based on the extraction of information from a boost::rtree.

@blaisb, this should also fix the problems that `compute_mesh_predicate_bounding_box` has on co-dimension one grids.


luca-heltai:
If we want to run step-70 with codimension one immersd domains, this PR will be necessary.

kronbichler:
One thing you might consider is to add the picture you posted above to the documentation of the new function.

luca-heltai:
Something like this?

image

image

image

kronbichler:
Yes, that would be some very nice pictures!

PR #10022, opened 2020-05-03
Test: adjust iteration count in ginkgo test

tamiko:


kronbichler:
This won't hit our testers, so merging as is.

PR #10024, opened 2020-05-03
Fix some clang-10 warnings

masterleinad:
Among others, also fixes the warnings in https://cdash.43-1.org/viewBuildError.php?type=1&buildid=5988.

### DIFF
--- a/bundled/umfpack/UMFPACK/Source/umf_grow_front.cc
+++ b/bundled/umfpack/UMFPACK/Source/umf_grow_front.cc
@@ -112,7 +112,7 @@ GLOBAL Int UMF_grow_front
 	/* :: frontal matrix size int overflow :: */
 	/* the desired front size is bigger than the integer maximum */
 	/* compute a such that a*a*s < Int_MAX / sizeof (Entry) */
-	double a = 0.9 * sqrt ((Int_MAX / sizeof (Entry)) / s) ;
+	double a = 0.9 * sqrt (((double) Int_MAX / sizeof (Entry)) / s) ;
 	fnr2 = MAX (fnr_min, a * fnr2) ;
 	fnc2 = MAX (fnc_min, a * fnc2) ;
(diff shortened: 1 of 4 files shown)


bundled/umfpack/UMFPACK/Source/umf_grow_front.cc
```diff
@@ -112,7 +112,7 @@ GLOBAL Int UMF_grow_front
 	/* :: frontal matrix size int overflow :: */
 	/* the desired front size is bigger than the integer maximum */
 	/* compute a such that a*a*s < Int_MAX / sizeof (Entry) */
-	double a = 0.9 * sqrt ((Int_MAX / sizeof (Entry)) / s) ;
+	double a = 0.9 * sqrt (((double) Int_MAX / sizeof (Entry)) / s) ;
```
peterrum, line 115: Is it intended that you have changed also UMFPACK?

masterleinad: Yes, that is an extra commit. The warning is along the lines that the implicit cast to `double` changes the value of `Int_MAX`.

rezarastak, line 115:
```suggestion
	double a = 0.9 * sqrt ((double(Int_MAX) / sizeof (Entry)) / s) ;
```
Doesn't clang-tidy warn also about c-style casts?

tjhei: This is clang-10, not clang-tidy-10, but I assume that we disable c-style cast warnings in umfpack.

masterleinad: This is also the style used in the rest of the file and I decided to rather follow it than to go with what we would do in the (main) library.

PR #10025, opened 2020-05-03
Implement cell-wise coefficient in MF::LaplaceOperator

kronbichler:
As discussed in https://github.com/dealii/dealii/pull/9883#discussion_r418564041, it makes sense to support cell-wise constant coefficients in `MatrixFreeOperators::LaplaceOperator`.

Given the already present interface, in particular `get_coefficient()`, I think that the best way to do things is to simply pass in a table with a single column for the value at all quadrature points. Providing a second set of a coefficient, taking only a vector, would be more complicated I believe. What do other think?

I also added a test and, while there, simplified some other tests that used a very complicated way of creating the logfile.


tjhei:
excellent, thank you!

Issue #10009, opened 2020-05-01
Some improvements for project_onto_orthogonal_tensors


rezarastak:
I looked into the function [project_onto_orthogonal_tensors](https://dealii.org/developer/doxygen/deal.II/classTensor.html#af4119f350cc384d64433dd9d325dc412), introduced in #9580, and I like the functionality it provides. It is a great addition to the new release of dealii.

However, I found areas that can be improved:
 - [x] It is tested within `tests/tensors/constexpr_tensor.cc` but this function is unrelated to constexpr things. *solution: move test to a new test file*
 - [x] It assumes `lapack` is available, which leads to test failure if there is no `lapack`. see [this](https://cdash.43-1.org/testDetails.php?test=41922973&build=5871) and [this](https://cdash.43-1.org/testDetails.php?test=42405948&build=5920). *solution: add `with_lapack=true`*
 - [x] It creates a dependency where `tensor.h` has to include  both `lapack_full_matrix.h` and `full_matrix.h`. This might increase compile times for all source files that need `tensor.h`. *solution: write it without `LAPACKFullMatrix`*.
 - [x] `full_matrix.h` is not included in `tensor.h` creating a broken dependency if we instantiate this function.
 - [x] The function does not work with `float`, it only works with `double`. *solution: write it in a more generic way.*
 - [x] The function assumes a tensor is already orthogonal if its determinant is 1. However, many non-orthogonal tensors have unit determinants. *solution: remove the if statement at the beginning of the function and remove the tolerance parameter altogether.*
 - [x] Copying the underlying tensor data into a matrix and back requires extra copying and dynamic allocation. *solution: improve algorithm*
 - [ ] The name could be improved too. In particular, the `s` at the end can be removed, because we are working with one tensor and not multiple tensors.

I am working on a number of PRs to fix these items one by one. I created this issue here so that we can keep track of the progress. Hopefully we can fix these before release.

For more context, here is the source code for this function.
https://github.com/dealii/dealii/blob/6309a80b94301577bc57e9ee12e17b611abdb803/include/deal.II/base/tensor.h#L2648-L2676

bangerth:
Let's bring @MFraters into the picture :-)

MFraters:
@bangerth Thanks for involving me :)

@rezarastak, these are great proposals and I think you are correct that it is best done incrementally. This code can certainly use improvement and I will be following this with interest!

Point 5: Yes, that was something we left open.

[...]

PR #10027, opened 2020-05-04
Use lapack gesvd() for orthogonal_project

rezarastak:
Related to #10009 

I removed the dependency on `FullMatrix` and `LAPACKMatrix` by calling the lapack svd function directory. It now works with `float` in addition to `double`. Moreover, it avoids any heap allocation and removes some unnecessary copying. Unfortunately it looks very unreadable to anyone not familiar with lapack.


rezarastak:
@MFraters If you have time, please also take a look at this. Thanks!

rezarastak:
Documentation for the lapack svd function can be found in this link
http://www.netlib.org/lapack/explore-html/d1/d7e/group__double_g_esing_ga84fdf22a62b12ff364621e4713ce02f2.html

`dgesvd` uses `double` numbers and `sgesvd` uses `float`s. In dealii we have a templated function `gesvd` which works for both `float` and `double`.

bangerth:
Would it be worthwhile keeping both the old and the new call? deal.II doesn't have to be configured with LAPACK, and it would be nice to have a fall-back if LAPACK wasn't available. Or is there a substantial difference in functionality between the old code and the LAPACK call?

bangerth:
Ah, forget this -- the old code already relied on LAPACK...

rezarastak:
> I also do have a quick question, can you explain why one of the test results are changed?

I initially overlooked this, but I have to think about it more. It seems that I made a poor choice for the input tensor A. By mistake, I chose a tensor with zero determinant, and it seems that the result of orthogonalization is not unique anymore.

rezarastak:
In my latest commit, I made these changes based on the comments I received.
* Added more comments for complicated arguments that I prepared for the lapack call.
* Changed the test file so that it doesn't test a singular (det = 0) input tensor anymore. Now the orthogonalization of this new tensor is going to be unique.

Additionally I added an assert statement to check the output of the lapack call.

MFraters:
> Changed the test file so that it doesn't test a singular (det = 0) input tensor anymore. Now the orthogonalization of this new tensor is going to be unique.

Hmm, I am not sure whether it is the correct way to go to just delete the test for that case. Users can input singular matrices, so or we need a way to prevent that or we need a way to deal with that in some way and test that. I am fine with that just being for example an Assert, so that it is only checked in debug mode. Or does does the Lapack assert cover this case?

rezarastak:
The current implementation of the function works even if the input is singular and provides all of the guaranteed postconditions of the function. The result is always orthogonal. The result is also the closest orthogonal tensor. It means that no other orthogonal tensors can be found with smaller distance from A. However, the function currently does not guarantee a unique solution. In fact, if a user provides a singular tensor, the there are infinite possible results (all orthogonal) for this function.

We can prove this fact mathematically. If A is singular, the last diagonal element of S will be 0. Therefore, the last column of U and the last column of V are not unique. The last column of U can become any arbitrary vector and similarly any arbitrary vector can be provided as the last column of V. So Q = U * V^T is not unique.

In summary, the results provided by by the current implementation is always correct, even if it is non-unique (implementation-dependent) whenever `det A == 0`. Now two questions arise that we should answer:
1. Should we ban using singular matrices in this function? We can rephrase it as: Is there any legitimate use case for calling this function with a singular matrix? I don't know the answer to this. I would never give this function a singular tensor, but I don't know about other people's applications. If there is no legitimate reason to use this function with a singular input, then we should `Assert(det(A) != 0)` to prevent users from doing this. What is everyone's opinion on this?
2. Should we test the behavior of this function in our test-suite with an input tensor that is singular? I believe the answer is no because it would make our test brittle. In the future we may change the implementation details of this function to improve its performance or add more features and it might change the output of the function when the singular test tensor is provided. This will flag a test failure even though the new implementation is correct (false positive) and would confuse the implementers (which it did for me).

> does the Lapack assert cover this case?

Providing a singular tensor as an input does not trigger a lapack assert failure.

MFraters:
> Should we ban using singular matrices in this function? We can rephrase it as: Is there any legitimate use case for calling this function with a singular matrix? I don't know the answer to this. I would never give this function a singular tensor, but I don't know about other people's applications. If there is no legitimate reason to use this function with a singular input, then we should Assert(det(A) != 0) to prevent users from doing this. What is everyone's opinion on this?

I do not think there is a legitimate reason, but it might happen. I do not have a strong opinion on it whether this should be checked in this function.

> Should we test the behavior of this function in our test-suite with an input tensor that is singular? I believe the answer is no because it would make our test brittle. In the future we may change the implementation details of this function to improve its performance or add more features and it might change the output of the function when the singular test tensor is provided. This will flag a test failure even though the new implementation is correct (false positive) and would confuse the implementers (which it did for me).

Thanks for the explanation. Depending on whether we test in the function or not, I agree with your conclusion, but I do not fully understand the argument you are making for it. 

If we decide to test for it, we can just test whether the assert is triggered when we pass it a singular matrix. I do not remember the exact code, but I think there is something we can add to the text so that it expects failure.

If we decide that we do not want to test for it, then I do not see why it would not be possible to make stable tests (basically testing whether the matrix is still orthogonal), but I do think that is will probably be a lot harder, and that would probably be more work than I would think it to be worth it (unless someone can come up with an elegant solution). So if we do not decide to test in the function, getting rid of the test might be the best option (like the code is now).

> Providing a singular tensor as an input does not trigger a lapack assert failure.

Ah, thanks, that is great to know!

rezarastak:
It seems that no one needs to pass a singular tensor to this function in any application. So I can put an assert that prevents it altogether. The good thing it that, we can revise this decision later without breaking any code that depends on it i.e. we can design a narrow contract for this function now which we can widen at some point in the future.

It is not necessary to check if the assert is triggered as part of a test. We don't check that for a ton of assert statements in the library.

masterleinad:
> It seems that no one needs to pass a singular tensor to this function in any application. So I can put an assert that prevents it altogether. The good thing it that, we can revise this decision later without breaking any code that depends on it i.e. we can design a narrow contract for this function now which we can widen at some point in the future.

I don't quite agree. Why doesn't it make sense to pass a singular matrix? In my opinion, it's enough to put a remark in the comments that the result might not be unique in this case. We can still check this case in the test suite by making sure that the distance in the Frobenius norm is correct.
I'd rather narrow the contract later (which would technically not be backward-compatible) than disallowing a valid use case.

bangerth:
On 5/7/20 12:20 PM, Daniel Arndt wrote:
>     The good thing it that, we can revise this decision later without breaking
>     any code that depends on it i.e. we can design a narrow contract for this
>     function now which we can widen at some point in the future.

I like how you're thinking in terms of contract programming! It's a shame that 
contracts didn't make it into C++2- :-(


> I don't quite agree. Why doesn't it make sense to pass a singular matrix? In 
> my opinion, it's enough to put a remark in the comments that the result might 
> not be unique in this case. We can still check this case in the test suite by 
> making sure that the distance in the Frobenius norm is correct.
> I'd rather narrow the contract later (which would technically not be 
> backward-compatible) than disallowing a valid use case.

I think the question is this: What's a sensible thing a user would expect to 
get back when passing in a singular matrix? In some sense, it's a bit like 
having a function
   void Vector::normalize() {
     *this /= this->norm();
   }
and expect it to work even for the zero vector. This is the equivalent 
operation for projecting a vector onto the unit sphere.

Yes, we can say that that's allowed and just choose *any* norm 1 vector to 
return. But is this what anyone wants? Likely not. Rather, it probably 
represents a programming bug to pass in such a vector, and we should catch this.

rezarastak:
I put the `Assert(is_singular(A))` in the code. It uses the ratio `S_33 / S_11` (which we already have from the svd) to determinant whether the matrix is singular.

rezarastak:
This PR, in addition to being an improvement to the algorithm, fixes a compilation bug. The old implementation does not include the `full_matrix.h` inside `tensor.h` but uses `FullMatrix` in the implementation of `project_onto_orthogonal_tensors()`. Therefore, a user that instantiates this function sees a compile error. Also, even if the user includes `full_matrix.h` in addition to `tensor.h`, she/he cannot compile this with `Number = float`. (it only works with `double`)

Since this is a bugfix, please consider merging it into the 9.2 release. I don't think we should postpone it. I addressed all of the comments and I think it is ready to be merged.

drwells:
I totally missed this when we did the release. I'm not too familiar with the problem it solves but it sounds like this isn't yet totally resolved.

One quick note, unrelated to the behavior of the function - we are trying to avoid `#include`ing `lapack_templates.h` in public headers to prevent clashes with other packages' possible wrapping headers. We will need to move either the function into a `.templates.h` file or a source file (and, in either case, add correct instantiations).

rezarastak:
@drwells Thank you for the information about lapack_templates. I incorporated your suggestion and created a source file for that.

This PR is ready for final review.

PR #10029, opened 2020-05-04
Fix C++20 warnings and errors

masterleinad:
- Avoid allocator::pointer removed in C++20 in Boost
- Avoid volatile assignment in TBB
- Avoid deprecated implicit this capture

### DIFF
--- a/bundled/boost-1.70.0/include/boost/signals2/detail/auto_buffer.hpp
+++ b/bundled/boost-1.70.0/include/boost/signals2/detail/auto_buffer.hpp
@@ -142,7 +142,11 @@ namespace detail
         typedef typename Allocator::size_type            size_type;
         typedef typename Allocator::difference_type      difference_type;
         typedef T*                                       pointer;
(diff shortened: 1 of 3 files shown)


bundled/boost-1.70.0/include/boost/signals2/detail/auto_buffer.hpp
```diff
@@ -142,7 +142,7 @@ namespace detail
         typedef typename Allocator::size_type            size_type;
         typedef typename Allocator::difference_type      difference_type;
         typedef T*                                       pointer;
-        typedef typename Allocator::pointer              allocator_pointer;
+        typedef typename Allocator::value_type*          allocator_pointer;
```
bangerth, line 145: Does this match recent upstream changes in boost?

masterleinad: I should have looked at that first. They do it slightly differently and I changed it accordingly.

PR #10030, opened 2020-05-04
Add a reference to the code gallery.

bangerth:
deal.II recently acquired a parallel N-S solver.

### DIFF
--- a/examples/step-57/doc/results.dox
+++ b/examples/step-57/doc/results.dox
@@ -401,4 +401,6 @@ multigrid is needed. See below.
 For larger computations, especially in 3d, it is necessary to implement MPI
 parallel solvers and preconditioners. A good starting point would be step-55,
 which uses algebraic multigrid for the velocity block for the Stokes
-equations.
+equations. Another option would be to take a look at the list of codes
+in the <a href="https://www.dealii.org/code-gallery.html">deal.II code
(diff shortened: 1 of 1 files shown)


examples/step-57/doc/results.dox
```diff
@@ -401,4 +401,6 @@ multigrid is needed. See below.
 For larger computations, especially in 3d, it is necessary to implement MPI
 parallel solvers and preconditioners. A good starting point would be step-55,
 which uses algebraic multigrid for the velocity block for the Stokes
-equations.
+equations. Another option would be to take a look at the list of codes
+in the <a href="https://www.dealii.org/code-gallery.html">deal.II code
+gallery</a>, which already contains parallel Navier-Stokes solvers.
```
tjhei, line 406:
```suggestion
gallery</a>, which already contains a parallel Navier-Stokes solver.
```

bangerth: We actually have two, so plural seemed appropriate :-)

jppelteret: Yip, there are indeed two that advertise themselves as NS solvers. So I'll consider this answered :-)

jppelteret:
This is just documentation, so we can merge it as is.

PR #10033, opened 2020-05-05
The tolerance parameter is removed from project_tensor

rezarastak:
Related to #10009 

The if statement `if det(A) == 1` must be removed because it is not a check for tensor orthogonality. Many non-orthogonal tensors have `det == 1`, and many orthogonal tensors have `det == -1 != 1`.

The correct check for orthogonality could be to check whether `A * A^T` is equal to the identity tensor. This test is relatively expensive, so it is better to not have a check at all. With this PR, we project any arbitrary tensor onto an orthogonal tensor with the minimum difference from the input tensor (measured in terms of the Frobenius norm). We perform this projection regardless of whether the tensor is already orthogonal.

Therefore, the `tolerance` parameter is not needed and must be removed. This is a breaking change, but since we did not have this function in the previous release, it may be ok to change its api.


bangerth:
Let's bring @MFraters into the picture :-)

MFraters:
Hey Reza Rastak,

Thanks for looking at this function! It can definitely be improved and you have some great suggestions in #10009. But I don't think your proposal is correct in this case.

> many orthogonal tensors have det == -1 != 1

Yes, this is true, that is why there is a absolute statement around the determinant. 

> Many non-orthogonal tensors have det == 1

That may be true for general tensors, but this function was written for 2nd order tensors for which I don't think that is the case. I am not a mathematician, but here are some sources: 


https://en.wikipedia.org/wiki/Orthogonal_matrix (I know Wikipedia is not always reliable, but I could not leave it out)

And here is a proof:
https://math.stackexchange.com/questions/1172802/show-that-any-orthogonal-matrix-has-determinant-1-or-1

So unless you want to generalize the function to a N dimensional tensor, I think it is a useful optimization. And even then I think it would be a nice special case. If the sources are incorrect or I am misinterpreting them, please let me know!

rezarastak:
Thank you for the your feedback @MFraters and providing explanations with appropriate links.

>>many orthogonal tensors have det == -1 != 1

>Yes, this is true, that is why there is a absolute statement around the determinant.

The original code compares `abs(det(A) - 1)` with `tol` which essentially checks whether `det(A) == 1)`. If we want to allow `det(A) == -1`, we need to compare `abs(det(A)) - 1` with `tol`.


>>Many non-orthogonal tensors have det == 1

>That may be true for general tensors, but this function was written for 2nd order tensors for which I don't think that is the case.

In a new test statement that I added to `tests/tensors/project_orthogonal.cc` I show that the 3x3 tensor `A = {{1, 2, 0}, {0, 1, 0}, {0, 0, 1}}` satisfies `det(A) == 1` but it is not orthogonal i.e. `A * A^T != I`

MFraters:
> The original code compares abs(det(A) - 1) with tol which essentially checks whether det(A) == 1). If we want to allow det(A) == -1, we need to compare abs(det(A)) - 1 with tol.

Hmm, you are right, that is then incorrect.

> In a new test statement that I added to tests/tensors/project_orthogonal.cc I show that the 3x3 tensor A = {{1, 2, 0}, {0, 1, 0}, {0, 0, 1}} satisfies det(A) == 1 but it is not orthogonal i.e. A * A^T != I

hmm, I should have though more carefully about the  `Many non-orthogonal tensors have det == 1`, and the link I showed only showed the first point. 

So in short, you are correct, this is not a good test. The case I am using it for, I will be testing a lot of tensors which are going to be very close to orthogonal all the time, so I am afraid that orthogonalizing the matrix every time will be expensive. Having said that, I do not have a good intuition if in that case testing `norm(A*A^T)-1<eps` is going to be cheaper than computing the svd. That sound like something which could be tested though callgrind, but I am not sure it is worth the effort.

The more I think about it, the more I think it is probably very problem dependent and it might therefore not be the responsibility of this function to check that, but of the code which calls this function to decide whether it wants to spend the time to check with a profiler whether testing this makes it faster for their use case. It would make this function more simple and readable.

@bangerth, what is your opinion on this?

rezarastak:
> The case I am using it for, I will be testing a lot of tensors which are going to be very close to orthogonal all the time.

Thank you for this information. Knowing your use case helps us better design this function. As you pointed out, in this case, it might be better if you test whether your matrix is almost orthogonal before calling project function. This way we can keep `project_onto_orthogonal_tensors` more general.
```c++
if (std::abs(determinant(A)) - 1. < tol)
  A = project_onto_orthogonal_tensors(A);
```
It should be noted that computing the determinant of a 3x3 tensor is not that cheap, so I am not sure how much efficiency you are saving by checking it.

One of correct methods of checking whether a tensor is orthogonal is as follows. We have to compute the dot product `A_i . A_j` for every two rows of the tensor `A` and compare it with the Kronecker `delta_ij`.
```c++
bool is_orthogonal = true;
  for (unsigned int j = i; j < dim; ++j)
    if (std::abs(A[i] * A[j] - static_cast<Number>(i == j)) > tol)
      is_orthogonal = false;
```

> That sound like something which could be tested though callgrind

Profiling the code within your specific program on your specific machine is always the best guide for you. However, even without benchmarks, we can reason about some the efficiency of some of our approaches. All three operations (determinant, svd, and the dot products A_i * A_j) have complexity in the order of `O(dim^2)` floating-point operations. Of course `svd` is more expensive but I believe I improved its performance considerably in #10027.

On a separate note, having an additional `if` statement also reduces the performance on modern cpus.
https://stackoverflow.com/questions/315306/is-if-expensive

masterleinad:
> On a separate note, having an additional `if` statement also reduces the performance on modern cpus.
> [stackoverflow.com/questions/315306/is-if-expensive](https://stackoverflow.com/questions/315306/is-if-expensive)

Well, that really depends on what you are checking at all and what you are guarding. Both of them are a lot more expensive than a couple instructions here. I am pretty sure that we don't need to worry about that here.

> Profiling the code within your specific program on your specific machine is always the best guide for you. However, even without benchmarks, we can reason about some the efficiency of some of our approaches. All three operations (determinant, svd, and the dot products A_i * A_j) have complexity in the order of `O(dim^2)` floating-point operations. Of course `svd` is more expensive but I believe I improved its performance considerably in #10027.

They should all need `O(dim^3)` multiplications for the square case (https://en.wikipedia.org/wiki/Computational_complexity_of_mathematical_operations) with suitable implementations but the current implementation uses Laplace expansion with complexity `O(n!)`.

rezarastak:
> Am I correct that you can add a early return in the double loop if it sets is_orthorgonal to false?

Yes you can break early out of the loops. A single `break` statement is not enough because you have to get out of two nested `for` loops.

PR #10034, opened 2020-05-05
Add codim 1 instantiations of make_hanging_node_constraints()

starki0815:
As discussed with @drwells in the mailing list, this is the pull request for added codim 1 instantiations of make_hanging_node_constraints()


source/dofs/dof_tools_constraints.cc
```diff
@@ -546,91 +546,56 @@ namespace DoFTools
...
     make_hp_hanging_node_constraints(
-      const dealii::hp::DoFHandler<1> & /*dof_handler*/,
+      const dealii::hp::DoFHandler<1, spacedim> & /*dof_handler*/,
       AffineConstraints<number> & /*constraints*/)
     {
       // we may have to compute constraints for vertices. gotta think about that
       // a bit more
 
       // TODO[WB]: think about what to do here...
+      Assert(false, ExcNotImplemented());
```
kronbichler, line 560: Why did you add these?

source/dofs/dof_tools_constraints.cc
```diff
@@ -546,91 +546,56 @@ namespace DoFTools
...
-      const dealii::hp::DoFHandler<1> & /*dof_handler*/,
+      const dealii::hp::DoFHandler<1, spacedim> & /*dof_handler*/,
       AffineConstraints<number> & /*constraints*/,
       std::integral_constant<int, 1>)
     {
       // we may have to compute constraints for vertices. gotta think about that
       // a bit more
 
       // TODO[WB]: think about what to do here...
+      Assert(false, ExcNotImplemented());
```
kronbichler, line 575: Same here.

source/dofs/dof_tools_constraints.cc
starki0815: I added these, because it is not obvious to me that there are generally no constraints for the hp case in 1d stemming from p-adaptivity (see also the comment in the code of @bangerth). If this is really not the case, the assertion can be eliminated (and also some template specializations joined because then there is no difference between DoFHandler and hp::DoFHandler for 1d anymore).

bangerth: That comment is ~13 years old. I haven't given it any dedicated thought, but I believe that I would have come across a case where we need these constraints since then. So I think both the comment and the assertion can go.

starki0815: All the better, then these assertions are superfluous and the PR should also not break tests anymore :)

starki0815:
@kronbichler As discussed, I have removed the assertions so that also no tests are broken by the PR anymore.

masterleinad:
Would you mind to rebase such that we get rid of these (hopefully) spurious `clang-tidy` CI errors?

PR #10036, opened 2020-05-05
Allow hp::DoFHandler with parallel::DistributedTriangulationBase

starki0815:
This allows for the distribution of dofs for an hp::DoFHandler combined with a triangulation class derived from parallel::DistributedTriangulationBase other than parallel::distributed::Triangulation.

I would be happy if this can be pulled so that I do not have to work with a custom version of deal.II (I have an application, where I do manage the partitioning of a triangulation manually, i.e., without p4est).

### DIFF
--- a/source/hp/dof_handler.cc
+++ b/source/hp/dof_handler.cc
@@ -968,7 +968,7 @@ namespace internal
          * For parallel::shared::Triangulation objects,
          * this information is distributed on both ghost and artificial cells.
          *
-         * In case a parallel::distributed::Triangulation is used,
+         * In case a parallel::DistributedTriangulationBase is used,
          * indices are communicated only to ghost cells.
          */
         template <int dim, int spacedim>
@@ -1014,11 +1014,12 @@ namespace internal
                     cell->index(),
                     active_fe_indices[cell->active_cell_index()]);
             }
-          else if (const dealii::parallel::distributed::Triangulation<dim,
-                                                                      spacedim>
-                     *tr = dynamic_cast<const dealii::parallel::distributed::
-                                          Triangulation<dim, spacedim> *>(
-                       &dof_handler.get_triangulation()))
+          else if (const dealii::parallel::
+                     DistributedTriangulationBase<dim, spacedim> *tr =
+                       dynamic_cast<
+                         const dealii::parallel::
+                           DistributedTriangulationBase<dim, spacedim> *>(
+                         &dof_handler.get_triangulation()))
             {
               // For completely distributed meshes, use the function that is
               // able to move data from locally owned cells on one processor to
@@ -1217,7 +1218,7 @@ namespace internal
 
         /**
          * Coarsening strategy for the CellDataTransfer object responsible for
-         * tranferring the active_fe_index of each cell on
+         * transferring the active_fe_index of each cell on
          * parallel::distributed::Triangulation objects that have been refined.
          *
          * A finite element index needs to be determined for the (not yet
@@ -1716,7 +1717,7 @@ namespace hp
     // decide whether we need a sequential or a parallel shared/distributed
     // policy and attach corresponding callback functions dealing with the
     // transfer of active_fe_indices
-    if (dynamic_cast<const parallel::distributed::Triangulation<dim, spacedim>
+    if (dynamic_cast<const parallel::DistributedTriangulationBase<dim, spacedim>
                        *>(&this->get_triangulation()))
       {
         policy = std_cxx14::make_unique<
@@ -2024,8 +2025,18 @@ namespace hp
   DoFHandler<dim, spacedim>::pre_distributed_active_fe_index_transfer()
   {
 #ifndef DEAL_II_WITH_P4EST
-    Assert(false, ExcInternalError());
+    Assert(false,
+           ExcMessage(
+             "You are attempting to use a functionality that is only available "
+             "if deal.II was configured to use p4est, but cmake did not find a "
+             "valid p4est library."));
 #else
+    // the implementation below requires a p:d:T currently
+    Assert((dynamic_cast<
+              const parallel::distributed::Triangulation<dim, spacedim> *>(
+              &this->get_triangulation()) != nullptr),
+           ExcNotImplemented());
+
     // Finite elements need to be assigned to each cell by calling
     // distribute_dofs() first to make this functionality available.
     if (fe_collection.size() > 0)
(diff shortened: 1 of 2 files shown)


peterrum:
@marcfehling Could you also have a look at this PR.

starki0815:
> For the `DoFHandler` these changes have been already done. But there has been no one who needed this combination till now. Could you also add 1-2 tests.

I have to think about what a meaningful test would be. But, yes, I can do that.

> Out of curiosity, do you have you own `Triangulation` class?

Yes. I have a domain mesh and a codim 1 mesh for interfaces in the domain, where interface related dofs live. The main problem is that integrals on the interfaces involve dofs related to the domain and dofs related to the interface. So, the partitioning of the interface mesh must be consistent with the partitioning of the domain mesh - otherwise one would get quite complex communication between processors. As I can't see an easy way to manually partition with p4est, I use a custom distributed triangulation class for the interface mesh. That is quite straightforward because the partitioning of the interface is prescribed by the partitioning of the domain.

peterrum:
@starki0815 I see. Have you tried out `parallel::fullydistributed::Triangulation`? If yes, what is missing for your use case? Maybe we could add the missing feature after release 9.2.

starki0815:
Well, I just learned about `parallel::fullydistributed::Triangulation` :) Most of my code was written at a time when `parallel::fullydistributed::Triangulation` didn't exist yet. Looking at the documentation, it could potentially work for me. But on the other hand my custom class is quite slim. In fact, the only function which is a bit more complicated is the partitioning - and that is what I would have to implement for `parallel::fullydistributed::Triangulation` as well. So right now, there is no big difference. That would change of course if `parallel::fullydistributed::Triangulation` would work with refinement (at the moment I have to rebuild the interface mesh after every refinement; and solution transfer can only happen through the underlying domain cells, which is a bit awkward).

peterrum:
@starki0815 One of the ideas of `p:f:t` is that you can provide your own partitioned.

BTW: PR #8834 is extending `p:f:t` with coarsening and refinement capabilities. But I never found the time to finish it. If you are interested, we can finish it together.

source/hp/dof_handler.cc
```diff
@@ -1716,39 +1717,55 @@ namespace hp
...
+        // parallel::distributed::Triangulation
+        // TODO: If possible, change interfaces so that the whole hp::DoFHandler
+        // works for any class derived from
+        // parallel::DistributedTriangulationBase as long as the triangulation
+        // class provides with appropriate functionality (the dof handler should
+        // not depend on how exactly things like solution transfer are
+        // implemented by the triangulation class)
+        if (dynamic_cast<
+              const parallel::distributed::Triangulation<dim, spacedim> *>(
+              &this->get_triangulation()))
```
bangerth, line 1737: I've been staring at this for a long time now, but can't figure out whether we actually need the `if` here. We want to attach ourselves to signals so that we can ensure certain things, or do certain actions, whenever something happens. I think that is true whether the triangulation in use is a `p::d::T`, or any other distributed triangulation.

In other words, *that* we want to hear about certain events is independent of the triangulation in question. The only place where one might think about doing something different is *in the implementation* of the functions being called by these signals.

So my suggestion would be to remove the `if` here, which implies that the only thing you change is the top-level `if`on line 1719.

starki0815: I've thought about the `if` for quite a while as well. First of all, I agree that it shouldn't be there in a clean design (which is, in my opinion, not the case right now for historical reasons).

My reasoning for putting the `if` there for now was the following: The implementation of the functions attaching to the signals is currently specific to `p::d::T`. So, if these functions are called for anything else than `p::d::T`, one runs into an assertion. Now, if one derives an own distributed triangulation class from `DistributedTriangulationBase` and triggers those signals during refinement/coarsening, and the `hp::DoFHandler` is attached to those, this will abort the program. Of course, if the the `hp::DoFHandler` does not attach to the signals, this then means that one does have to take care of proper assignment of active fe indices, etc. by oneself.

So, as indicated by my TODO comment above the `if` in question, I think the proper approach would be to make the `hp::DoFHandler` completely independent of how exactly things like the transfer of cell data are implemented. But this will require more changes to the triangulation interfaces, which is something I didn't have the time for right now, but which should definitely be done in the future.

But that's just my opinion. If you say, the `if` should go, that's fine for me. In my own application I can just make sure that the signals are never triggered by the triangulation and then I'm good.

starki0815: What I'm realizing just now: What I can of course do, is to remove the assertions in the functions attaching to the signals and just let them do nothing in case it's not a `p::d::T`. That should be consistent with what you suggest. Do you agree?

starki0815:
@peterrum 

> One of the ideas of p:f:t is that you can provide your own partitioned.

Yes, I understand. But in my case it's essentially as simple as to copy over subdomain ids from the underlying domain cells to the interface cells. From what I understand, that's exactly what I'd have to implement in the partitioner for `p:f:t`. Or am I missing something?

> BTW: PR #8834 is extending `p:f:t` with coarsening and refinement capabilities. But I never found the time to finish it. If you are interested, we can finish it together.

I'm certainly interested, but am certainly pretty busy for the next two or three months. So this would have to wait a while. A question: Do you also plan to implement re-partitioning together with support for transfer of cell-related data? I bet it's extremely hard to make it work. But to make the class really useful in the context of my application that would be necessary.

source/hp/dof_handler.cc
bangerth: Right. That seems like the better approach. Or, if you want to be on the safe side, remove the `if` here and in the functions that are called do something like
```
  Assert (dynamic_cast<p::d::T*>(...) != nullptr, ExcNotImplemented());
```
so that if the signal is triggered, one gets an error in the place that will have to be checked for any new triangulation class, rather than in an unrelated location.

This also means that a program is fine as long as none of these functions are ever triggered. For example, if you run your program with your own triangulation, then just don't refine (or just don't trigger the signal) and everything will be fine.

starki0815: Ok, I agree and have changed the code accordingly.

starki0815:
I accidentally removed the preprocessor checks whether deal.II is compiled with p4est, which of course leads to compiler errors in case deal.II is not compiled with p4est as the implementation of hp::DoFHandler is specific to `p:d:T`. Now, things are hopefully fine.

source/hp/dof_handler.cc
```diff
@@ -2024,8 +2025,18 @@ namespace hp
   DoFHandler<dim, spacedim>::pre_distributed_active_fe_index_transfer()
   {
 #ifndef DEAL_II_WITH_P4EST
-    Assert(false, ExcInternalError());
+    Assert(false,
+           ExcMessage(
+             "You are attempting to use a functionality that is only available "
+             "if deal.II was configured to use p4est, but cmake did not find a "
+             "valid p4est library."));
```
bangerth, line 2032: Thanks for taking the time to improve this error message! :+1:

PR #10038, opened 2020-05-05
Remove some non-ascii characters.

drwells:
Followup to #10007.

My copy of pdflatex does not like (i.e., I cannot compile the doxygen-generated file _formulas.tex) math statements with unicode MINUS SIGN instead of the old ASCII HYPHEN-MINUS. While here I also switched out RIGHT SINGLE QUOTATION MARK for (also ASCII) APOSTROPHE and made a few other small changes.

We should be, in general, able to use unicode in our documentation outside of math mode so I didn't change those usages unless they seemed gramatically incorrect to me (like using the right quotation mark instead of an apostrophe).

### DIFF
--- a/include/deal.II/base/tensor_product_polynomials.h
+++ b/include/deal.II/base/tensor_product_polynomials.h
@@ -409,7 +409,7 @@ private:
   const std::vector<std::vector<Polynomials::Polynomial<double>>> polynomials;
 
   /**
-   * Each tensor product polynomial @þ{i} is a product of one-dimensional
+   * Each tensor product polynomial $p_i$ is a product of one-dimensional
    * polynomials in each space direction. Compute the indices of these one-
    * dimensional polynomials for each space direction, given the index
    * <tt>i</tt>.
--- a/include/deal.II/lac/utilities.h
+++ b/include/deal.II/lac/utilities.h
@@ -445,11 +445,11 @@ namespace Utilities
       // algorithm verbatim:
       //
       // [Y]=chebyshev_filter_scaled(X, m, a, b, aL).
-      // e=(b−a)/2; c=(a+b)/2; σ=e/(c−aL); τ=2/σ;
-      // Y=(H∗X−c∗X)∗(σ/e);
+      // e=(b-a)/2; c=(a+b)/2; σ=e/(c-aL); τ=2/σ;
+      // Y=(H∗X-c∗X)∗(σ/e);
       // for i=2 to m do
-      //   σnew =1/(τ −σ);
-      //   Yt =(H∗Y−c∗Y)∗(2∗σnew/e)−(σ∗σnew)∗X;
+      //   σnew =1/(τ - σ);
+      //   Yt =(H∗Y - c∗Y)∗(2∗σnew/e)-(σ∗σnew)∗X;
       //   X =Y; Y =Yt; σ =σnew;
 
       const double e     = (b - a) / 2.;
--- a/include/deal.II/numerics/vector_tools_boundary.h
+++ b/include/deal.II/numerics/vector_tools_boundary.h
@@ -609,7 +609,7 @@ namespace VectorTools
    * L2-projection method used has been taken from the paper "Electromagnetic
    * scattering simulation using an H (curl) conforming hp finite element
    * method in three dimensions" by PD Ledger, K Morgan and O Hassan ( Int. J.
-   * Num. Meth. Fluids, Volume 53, Issue 8, pages 1267–1296).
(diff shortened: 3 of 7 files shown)


include/deal.II/base/tensor_product_polynomials.h
```diff
@@ -409,7 +409,7 @@ class AnisotropicPolynomials : public ScalarPolynomialsBase<dim>
   const std::vector<std::vector<Polynomials::Polynomial<double>>> polynomials;
 
   /**
-   * Each tensor product polynomial @þ{i} is a product of one-dimensional
```
drwells, line 412: This one is particularly interesting - its not every day we see thorn :)

bangerth:
On 5/5/20 10:18 AM, David Wells wrote:
> This one is particularly interesting - its not every day we see thorn :)

Huh, very cool. How did that get in there!?

drwells:
I was hoping you could tell me since you added it in 015d1d31567aa3dbbf2c436d1f34886051939468. Surely you remember what you were thinking that day 17 years ago?

bangerth:
How bizarre. I'm pretty sure I wrote that code as part of implementing the RT element. But I wouldn't even know how to create that symbol on a keyboard, so it's a mystery to me...

What a fun find!

include/deal.II/lac/utilities.h
```diff
@@ -445,11 +445,11 @@ namespace Utilities
...
-      // e=(b−a)/2; c=(a+b)/2; σ=e/(c−aL); τ=2/σ;
-      // Y=(H∗X−c∗X)∗(σ/e);
+      // e=(b-a)/2; c=(a+b)/2; σ=e/(c-aL); τ=2/σ;
+      // Y=(H∗X-c∗X)∗(σ/e);
       // for i=2 to m do
-      //   σnew =1/(τ −σ);
-      //   Yt =(H∗Y−c∗Y)∗(2∗σnew/e)−(σ∗σnew)∗X;
+      //   σnew =1/(τ - σ);
+      //   Yt =(H∗Y - c∗Y)∗(2∗σnew/e)-(σ∗σnew)∗X;
       //   X =Y; Y =Yt; σ =σnew;
```
jppelteret, line 453: Are these other symbols `τ`, `σ`, etc. actually ascii?

include/deal.II/lac/utilities.h
```diff
@@ -445,11 +445,11 @@ namespace Utilities
       // algorithm verbatim:
       //
       // [Y]=chebyshev_filter_scaled(X, m, a, b, aL).
-      // e=(b−a)/2; c=(a+b)/2; σ=e/(c−aL); τ=2/σ;
-      // Y=(H∗X−c∗X)∗(σ/e);
+      // e=(b-a)/2; c=(a+b)/2; σ=e/(c-aL); τ=2/σ;
```
bangerth, line 448: Does the sigma character actually work?

include/deal.II/lac/utilities.h
drwells: No, but since its not in math mode I left it alone. I suppose I am being slightly inconsistent in my removal of non-ascii characters, but I wanted to consistently remove the minus sign.

include/deal.II/lac/utilities.h
```diff
@@ -445,11 +445,11 @@ namespace Utilities
...
+      // Y=(H∗X-c∗X)∗(σ/e);
       // for i=2 to m do
-      //   σnew =1/(τ −σ);
-      //   Yt =(H∗Y−c∗Y)∗(2∗σnew/e)−(σ∗σnew)∗X;
+      //   σnew =1/(τ - σ);
```
bangerth, line 451: Same for the tau here? Or is this just not critical because we're not in a formula?

drwells: This comment block is wrapped in an `#ifndef DOXYGEN` block so its never rendered anyway.

include/deal.II/lac/utilities.h
drwells: See my answer below.

jppelteret:
Gosh, what a pain. This is the sort of stuff the I cleaned up in our publication list. I remember spending quite some time trying to work out what LaTeX was complaining about!

include/deal.II/lac/utilities.h
jppelteret: Didn't think so :-D But its certainly not important to clean this up, so we won't worry about this now.

jppelteret:
Since this is just documentation, we can merge as is.

PR #10039, opened 2020-05-05
Incompatible: Fix operator<< with Vector

### LINKED ISSUE #10017: operator << (std::ostream&, const Vector &) does not respect stream settings
In the process of a difficult to trace test failure, I came across the following oddity:
```
/**
 * Output operator writing a vector to a stream.
 */
template <typename number>
inline std::ostream &
operator<<(std::ostream &os, const Vector<number> &v)
{
  v.print(os);
  return os;
}
```
which calls this function:
```
template <typename Number>
void
Vector<Number>::print(std::ostream &     out,
                      const unsigned int precision,
                      const bool         scientific,
                      const bool         across) const
{
  Assert(size() != 0, ExcEmptyObject());
  AssertThrow(out, ExcIO());

  std::ios::fmtflags old_flags     = out.flags();
  unsigned int       old_precision = out.precision(precision);

  out.precision(precision);
  if (scientific)
    out.setf(std::ios::scientific, std::ios::floatfield);
  else
    out.setf(std::ios::fixed, std::ios::floatfield);

  if (across)
    for (size_type i = 0; i < size(); ++i)
      out << values[i] << ' ';
  else
    for (size_type i = 0; i < size(); ++i)
      out << values[i] << std::endl;
  out << std::endl;

  AssertThrow(out, ExcIO());
  // reset output format
  out.flags(old_flags);
  out.precision(old_precision);
}
```

In other words, calling
```
  std::cout << vec;
```
does not respect the precision and other settings of the stream. This took me a rather long time to figure out because I needed to increase the precision with which one test outputs its results and I just wasn't getting any results.

Proposal:
* Get rid of `operator<<` calling `print()` and instead output using the flags set by the stream.
* Don't output a newline at the end of the vector -- that ought to be the user's job, and is how other classes do it (e.g., `Point<dim>`).

Opinions?

bangerth:
This addresses the issue mentioned in #10017 by replacing the current implementation of `operator<<(ostream, Vector)` by one that respects the `ostream` flags. The patch itself is relatively small (but incompatible), but required the adjustment of quite a number of tests. The new test output is more systematic, though.

Comments?

Fixes #10017.
/rebuild


masterleinad:
I am not quite sure if it's a good idea to define this function in the first place since none of the other vector types does it. I would be happy to deprecate in favor of `print`.
If we decide to keep it, I am happy with this change, though.

bangerth:
I think the function is not completely pointless -- many of our other data types also define `operator<<` and there is certainly great value in it for debugging purposes. Moreover, it is *the* default way to output data in C++: Everyone would certainly try `std::cout << v;` before looking up the documentation to find an alternative way to output a vector.

I think that if there's a mistake, then it's that the other vector types don't implement the operator.

PR #10040, opened 2020-05-05
Fix some more doxygen issues.

drwells:
Another followup to #10007.

1. I now get a warning about paper sizes with doxygen 1.8.18:

warning: argument 'a4wide' for option PAPER_TYPE is not a valid enum value
Using the default: a4!

   Get around this by just using a4.

2. My LaTeX distribution already defines vcentcolon, so rename our own version
   to dealvcentcolon to preserve compatibility with older tex distributions.

3. Use mathrm instead of text to get underscores to work correctly with
   backslashes with both MathJax and LaTeX. This is the solution we use
   elsewhere in the library.

### DIFF
--- a/doc/doxygen/extra.sty
+++ b/doc/doxygen/extra.sty
@@ -18,7 +18,7 @@
 
 % Note: If you add an entry here, also put it into
 % ./doc/doxygen/scripts/mod_header.pl.in
-\newcommand{\vcentcolon}{\mathrel{\mathop{:}}}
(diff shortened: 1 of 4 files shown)


bangerth:
I don't think this is going to trigger anything we automatically test, so let me merge.

masterleinad:
Well, `Travis` checks for warnings and errors when generating the documentation using `MathJax` but that one already succeeded.

PR #10047, opened 2020-05-06
Better document C++11/14/17/20.

bangerth:
Also related to #10000 and #10018. I moved the file from `c++11.h` to `c++.h`, but only the last paragraph of the C++11 section is new, plus the documentation of the `std_cxx14/17/20` namespaces that were previously not documented.

I've also left notes in the documentation of various functions that put their use in context of the C++11 range-based for loops.


drwells:
Should we also rename the group? e.g., `CPPStandard` or something else that indicates we cover multiple language versions?

rezarastak:
We can also add a short description about `constexpr` and `DEAL_II_CONSTEXPR` to the namespaces `std_cxx11` and `std_cxx14`, respectively. The description can be added in a separate PR.

bangerth:
@drwells Good question. I didn't put the new namespaces into the group. The group as it is currently documented is specifically about C++11, so I think it's right to leave it as is. It *would*, however, make sense to create a new group that this group is part of, along with all of the namespaces. If you would like that, I can do that as a follow-up.

@rezarastak Good point, too. I think you know that stuff better than I do. I think `constexpr` shouldn't be documented as part of that namespace, but it could very well be put into that larger group mentioned above. Would you be interested in writing something as a follow-up to my follow-up?

rezarastak:
> I think constexpr shouldn't be documented as part of that namespace, but it could very well be put into that larger group mentioned above. Would you be interested in writing something as a follow-up to my follow-up?

Sure! I will write a short description about `constexpr` and `DEAL_II_CONSTEXPR` inside the group `CPP11`

PR #10048, opened 2020-05-06
base/utilities_pack_unpack_07: make test more robust

tamiko:


tests/base/utilities_pack_unpack_07.cc
```diff
@@ -53,17 +53,16 @@ check(const double (&array)[N])
...
 
-  // computing_timer.print_summary();
-
   // check if compression has been invoked by comparing sizes
   deallog << "unpacked array: " << sizeof(array) << std::endl;
 
   deallog << "packed array without compression: " << array_uncompressed.size()
           << std::endl;
 
+  // set the lowest digit to 0:
```
bangerth, line 62: ok if you add a sentence explaining *why* you want to do that.

PR #10051, opened 2020-05-07
fe/memory_consumption_01: remove test

tamiko:
When compiled against libc++ we have a slightly different memory
consumption profile than compared to libstdc++. Simply add a variant


drwells:
That's odd - I thought we removed all of the memory consumption tests in #4727. Shouldn't we also just remove this test?

tamiko:
On Wed, May  6, 2020, at 21:10 CDT, David Wells <notifications@github.com> wrote:

> That's odd - I thought we removed all of the memory consumption tests
> in #4727. Shouldn't we also just remove this test?

Ah, that explains why I have so few memory consumption test failures...


deleting.

tamiko:
@drwells Updated.

PR #10052, opened 2020-05-07
mpi/petsc_step-27*: add a test variant for petsc-3.13 and newer

tamiko:


drwells:
I think these iterations are off by 1 from the current output files.

Could we just hard-code the current number of iterations and then use `check_solver_within_range`? I can put together a patch that does this if you want to see it.

tamiko:
> I think these iterations are off by 1 from the current output files.
>
> Could we just hard-code the current number of iterations and then use
> `check_solver_within_range`? I can put together a patch that does this
> if you want to see it.

The total number of degrees of freedom are also different - i.e. we will
need an output variant. Because of that I opted to not change to the
solver_within_range macro but use an output variant instead. If the test
evers fails again (due to a different number of iterations, we can
update to the macro I suggest).

drwells:
Ah, my mistake - I didn't notice the dof change.

PR #10061, opened 2020-05-07
Parallel direct solver block

mwichro:
The initial application was solving coarse Stokes problem for matrix-free mg solver.
The initialization assembles the matrix using N vmults, than initializes Trilinos direct solver.
The internal DoF renumbering is implemented, so the internal matrix is standar SparseMatrix (non-block).
It could be extended to be initialized from BlockSparseMatrix by reordering rows (reordering is already implemented)

This is related to PR #9723

### DIFF
--- a/include/deal.II/lac/block_direct_solver.h
+++ b/include/deal.II/lac/block_direct_solver.h
@@ -1,0 +1,407 @@
+/*
+ * block_direct_solver.h
+ *
+ *  Created on: Mar 18, 2020
+ *      Author: mwichro
+ */
+
+#ifndef INCLUDE_BLOCK_DIRECT_SOLVER_H_
+#define INCLUDE_BLOCK_DIRECT_SOLVER_H_
+
+#include <deal.II/base/index_set.h>
+#include <deal.II/base/logstream.h>
+#include <deal.II/base/utilities.h>
+
+#include <deal.II/lac/trilinos_parallel_block_vector.h>
+#include <deal.II/lac/trilinos_precondition.h>
+#include <deal.II/lac/trilinos_solver.h>
+#include <deal.II/lac/trilinos_sparse_matrix.h>
+
+#include <deal.II/multigrid/mg_coarse.h>
+
+#include <deal.II/numerics/vector_tools.h>
+
+using namespace dealii;
+
+template <typename LevelNumber>
+class BlockDirectSolver
+  : public MGCoarseGridBase<
+      LinearAlgebra::distributed::BlockVector<LevelNumber>>
+{
+public:
+  friend class BlockCoarseTest;
+  typedef LevelNumber                                          number;
+  typedef LinearAlgebra::distributed::BlockVector<LevelNumber> BlockVectorType;
+
+
+  // Constructor. Requires owned_partitioning[block][MPI process],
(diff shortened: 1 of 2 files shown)


luca-heltai:
This is similar to what #9723 does. Can you explain what the differences are?

include/deal.II/lac/block_direct_solver.h
```diff
@@ -0,0 +1,407 @@
...
+                {
+                  if (dst.block(k)(*index_iter) != 0)
+                    matrix->set(ext2int(*index_iter, k),
+                                ext2int(i, b),
+                                dst.block(k)(*index_iter));
+                }
+            }
+        }
+    }
+  matrix->compress(VectorOperation::unknown);
```
peterrum, line 293: @mwichro Would it possible to extract this into separate that is designated to compute from an operator a block matrix?

include/deal.II/lac/block_direct_solver.h
```diff
@@ -0,0 +1,407 @@
...
+template <typename LevelNumber>
+template <class Operator>
+void
+BlockDirectSolver<LevelNumber>::initialize(const Operator &    vmult_operator,
+                                           const unsigned int &lvl)
+{
+  matrix = std::make_unique<TrilinosWrappers::SparseMatrix>(shifted_owned,
+                                                            shifted_owned,
+                                                            mpi_comm,
+                                                            100);
```
peterrum, line 246: Magic number?

mwichro: Yup. Magic.
The truth is, this was implemented to handle Stokes 2D coarse problem and since   compress in  trilinos matrix removes unneeded elements I've putted a very rough approximate. I should change it  and  move it to  AdditionalData structure.

include/deal.II/lac/block_direct_solver.h
mwichro: The point is I am not computing the BlockMatrix but standard SparseMatrix. The internal DoF renumbering is the opposite what DoFRenumbering::component_wise does. The vmult operation is defined as PA^-1 P^-1, where A is internally stored matrix, and P is permutation matrix that ensures the locally owned index set are contiguous. 

For efficiently computing BlockMatrix from operator and solving linear system I would go other way around: first compute BlockMatrix using a sparsity pattern ( PR #9723 ), and than initialize this solver with BlockMatrix (functionality to be added here). Since for some time you will be storing both original and permutated matrix this will be not memmory efficient, but the main purpose of this solver is solving coarse problem (that usually signifficantly smaller that original and thus we don't have to care about memmory so much) and checking if your matrix/operator is properly assembled (looking at the solution of linear system may help localize the problem). In second case the user would most probably be debugging the code -- if you are using block matrices you are probably going to use iterative solver.

mwichro:
> This is similar to what #9723 does. Can you explain what the differences are?

The main functionality of this class is renumbering DoF so that the direct solver can be applied to parallel block system. Here, Trilinos direct solver is used. The matrix is assembled internally by doing N vmults, that is very inefficient,  I'm going to add initialization from the BlockMatrix (that will copy the matrix with renumbering rows and columns). 

 The main purpose are solving coarse system for MG and help users debugging. 

The code commited is Work in Progress.

PR #10064, opened 2020-05-07
Tests: symengine/cse*: Disable unstable output

tamiko:
The internal optimizer state is unfortunately not stable. I get
different results for every combination of compiler, standard library
and, optimization flags I use.

Thus, simply disable this portion of the output. (Note that the result
still gets checked for correctness with a barrage of Asserts.)


tamiko:
@jppelteret *ping*

tests/symengine/sd_common_tests/cse_03.h
```diff
@@ -433,17 +433,12 @@ evaluate_SD_SD_stored_symbols_optimisation(const Tensor<2, dim> &t,
...
-      const bool print_optimiser = false;
-#endif
-      if (print_optimiser == true)
-        {
-          deallog << "Optimizer" << std::endl;
-          optimizer.print(deallog.get_file_stream());
-        }
+      // Unfortunately, we cannot print out the optimizer state. The result
+      // is not stable and depends on standard library, optimization
+      // parameters and compiler being used. MM '20
```
jppelteret, line 438: This is really weird. I wonder what's going on in their library that makes this the case. I don't believe that I've done anything to make the output nondeterministic. Maybe its how the hashes for each expression are computed.

PR #10065, opened 2020-05-07
Silence gcc-10 warnings about invalid array accesses

masterleinad:


include/deal.II/base/tensor.h
```diff
@@ -2391,9 +2391,14 @@ DEAL_II_CONSTEXPR inline DEAL_II_ALWAYS_INLINE
 
   Tensor<1, dim, typename ProductType<Number1, Number2>::type> result;
 
-  result[0] = src1[1] * src2[2] - src1[2] * src2[1];
-  result[1] = src1[2] * src2[0] - src1[0] * src2[2];
-  result[2] = src1[0] * src2[1] - src1[1] * src2[0];
+  // avoid compiler warnings
+  constexpr int s0 = 0 % dim;
+  constexpr int s1 = 1 % dim;
+  constexpr int s2 = 2 % dim;
```
bangerth, line 2397: Ugh, not pretty. Alternatively, you could guard this block with a `if constexpr (dim == 3)` and make the `constexpr` conditional on C++17. Or maybe we just don't care and just go with your modification.

source/fe/fe_bernardi_raugel.cc
```diff
@@ -177,7 +177,6 @@ FE_BernardiRaugel<dim>::initialize_support_points()
     }
 }
 
-template class FE_BernardiRaugel<1>;
```
bangerth, line 180: What happens?

masterleinad: It's the same. We unconditionally access array element 1. AFAICT, it doesn't really make sense to use this finite element for 1D anyway.

bangerth: I guess that makes sense. @GrahamBenHarper -- FYI.

GrahamBenHarper: That's fine. There is no 1D Bernardi-Raugel element. :)

masterleinad:
The CI failure is spurious again.

PR #10066, opened 2020-05-07
Disable a fallback option in the BatchOptimizer

jppelteret:
This is what is promised in the documentation, and I forgot to change the behaviour everywhere.


jppelteret:
@tamiko FYI.

jppelteret:
@Rombur I told you that I removed this fallback option, but it looks like I only did it in one out of two places. Oops!

tamiko:
I guess this won't be seen by the CI at all (the header has a `DEAL_II_WITH_SYMENGINE` header guard.

tamiko:
Indenting successful and serial compilation worked as well.

PR #10068, opened 2020-05-07
symengine/batch_optimizer: Add output variants

tamiko:
Depending on architecture these tests produce slightly different results
with a relative error of just above 1e-6. Fix this by providing output
variants for clang-10 with libc++


jppelteret:
I'll merge this as is, since I believe that these tests won't be run on the CI testers.

Issue #10072, opened 2020-05-08
deal.II Release 9.2.0 (released 2020/05/20)


tamiko:
<b>Release [9.2.0](https://github.com/dealii/dealii/releases/tag/v9.2.0)</b>
===
 - [generated documentation](https://dealii.43-1.org/documentation/9.2.0)

Check list for [release-tasks](https://github.com/dealii/release-papers/blob/master/release-tasks)
===

Release preparation steps:
--

- [x] <b>-3</b>  Send out notice to the mailinglist if anyone has outstanding issues or patches. (@drwells)
- [x] <b>-2</b>  Check all issues and pull-request on github. Create a list of things to fix before the release. (@kronbichler, #9729)
- [x] <b>-1</b>  Update the list of publications. (Send standard e-mail)
- [x] <b>-1a</b> Start release paper process with email to all current (principal) developers
- [x] <b>-1b</b> Update the authors list on the homepage authors.html (@tamiko, mailmap: #10140)
- [x] <b>-1c</b> Update the list of contributors in the release paper (@tamiko)
- [x] <b>0</b>   Do some basic testing as always (@tamiko, #10073)
- [x] <b>0a</b>  Rerun cppcheck across the source files (@drwells #10107)
- [x] <b>0b</b>  Check for doxygen formatting problems @tjhei 
- [x] <b>0c</b>  Fix formatting bugs reported by various command line queries @tjhei #10094
- [x] <b>0d</b>  Check that doxygen produces no errors when generating offline docs. @tjhei 
- [x] <b>0e</b>  Run the copyright script @tjhei  #10097
- [x] <b>0f</b>  Update Trilinos and PETSc to the latest release on the tester and update documentation of supported version. (@tamiko, #10141)
- [x] <b>0g</b>  Run codespell (@drwells, #10075)

Prerelease/Release steps:
--
- [x] <b>*</b>   Nominate a "release officer" that is in charge of (almost all) of the following steps
- [x] <b>1</b>   Create new changes file
- [x] <b>1a</b>  Adjust doc/users/doxygen.html to link to the to-be-created TAG file.
- [x] <b>4</b>   Create a release branch
- [x] <b>5</b>   Update the VERSION from 9.2.0-pre to 9.3.0-pre on master (see the discussion below)
- [x] <b>6</b>   Update the official tester to test on the release branch dealii-9.2
- [x] <b>6a</b>  On mainline, update deal.II dependencies of examples
[...]

PR #10075, opened 2020-05-08
Run codespell.

drwells:
Part of #10072.

### DIFF
--- a/source/multigrid/mg_transfer_component.cc
+++ b/source/multigrid/mg_transfer_component.cc
@@ -531,16 +531,16 @@ MGTransferComponentBase::build(const DoFHandler<dim, spacedim> &mg_dof)
                   for (types::global_dof_index i = 0; i < n_dofs; ++i)
                     {
                       SparseMatrix<double>::iterator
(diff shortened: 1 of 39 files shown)


source/multigrid/mg_transfer_component.cc
```diff
@@ -531,16 +531,16 @@ MGTransferComponentBase::build(const DoFHandler<dim, spacedim> &mg_dof)
                   for (types::global_dof_index i = 0; i < n_dofs; ++i)
                     {
                       SparseMatrix<double>::iterator
-                        anfang = prolongation_matrices[level]
```
masterleinad, line 534: I personally think, we should use more german!

drwells: I could be convinced to leave it :)

bangerth: Ha, that's a good one :-)

bangerth: I bet that dates back to 1999 when @guidokanschat wrote these lines!

PR #10076, opened 2020-05-08
contrib/utilities/query_testsuite: modernize

tamiko:
Update the script to the latest CDash version and the newly introduced
"api/v1" interface. This avoids a lot of ugly html file parsing.


tjhei:
nothing can be tested here, so let's merge.

Issue #8914, opened 2019-10-16
Can one keep manifolds when merging triangulations?


nfehn:
``GridGenerator::merge_triangulations()`` forgets about the manifolds, see

https://www.dealii.org/current/doxygen/deal.II/namespaceGridGenerator.html#a7cd88e7eacd46697dee80ad2b8438d54

> Unlike most GridGenerator functions, this function does not attach any manifolds to result, nor does it set any manifold ids.

Is there a way to keep manifolds or simplify usability as it can become very complicated to reset manifolds after ``merge_triangulation()``?

What happens with manifolds when applying functions like ``GridTools::shift()``, ``GridTools::rotate()``?

masterleinad:
That's not quite true. The documentation also says
> The function copies the material ids of the cells of the two input triangulations into the output triangulation. If copy_manifold_ids is set to true, manifold ids will be copied. [...]

The manifolds assigned to objects stay the same if the position in space is altered.

masterleinad:
As long as the `Manifold` objects are invariant under the transformation, you should not have any problems after moving vertices.

nfehn:
Thanks for the explanations. I want to give further information on the second point mentioned above (moving the grid).

Let's consider the ``GridTools::rotate()`` function by the example of ``GridGenerator::torus<3,3>()``:
The code is the following
```
  double const r = 0.5, R = 1.5;
  GridGenerator::torus(*triangulation, R, r);
  GridTools::rotate(0.5*numbers::PI, 1 /* rotate around y */, *triangulation);
```
This works fine and ``rotate()`` does not pose problems as the torus geometry is rotationally symmetric w.r.t. the y-axis. The result looks like this
![rotate_y_coarse](https://user-images.githubusercontent.com/26224507/66982723-1005ae80-f0b7-11e9-88b1-c013f0b5b572.png)

Now, consider rotation around the z-axis
```
  double const r = 0.5, R = 1.5;
  GridGenerator::torus(*triangulation, R, r);
  GridTools::rotate(0.5*numbers::PI, 2 /* rotate around z */, *triangulation);
```
The code throws an exception in ``TransfiniteInterpolationManifold``, probably because the manifolds are no longer valid after the geometry has been rotated.

[...]

PR #10079, opened 2020-05-08
correct docu of GridTools::transform() and GridGenerator::cylinder()

### LINKED ISSUE #8920: Documentation of GridGenerator::cylinder() incomplete and incorrect.
1. The documentation of ``GridGenerator::cylinder()`` (https://www.dealii.org/current/doxygen/deal.II/namespaceGridGenerator.html#a5cdda7b4a76d509af7d1a8dc1320ddb0)
does not give information on how many elements are used in axial direction. It would make sense to extend the implementation so that the number of axial cells becomes a parameter of this function. No information is given how the 2D mesh looks like.

2. The documentation further says
> If you want the cylinder to revolve around a different axis than the $x$-axis, then simply rotate the mesh generated by this function using the GridTools::transform() function using a rotation operator as argument.

> The manifold id for the hull of the cylinder is set to zero, and a CylindricalManifold is attached to it.

This point is closely related to issue #8914, but is written here separately as things seem to become complicated.

Let's test the manifold first:
```
  double const r = 0.5, L = 1.5;
  GridGenerator::cylinder(*triangulation, r, L);
  triangulation->refine_global(2);
```
cylinder_refine2

Let's now apply a simple transformation on top
```
  double const r = 0.5, L = 1.5;
  GridGenerator::cylinder(*triangulation, r, L);
  Transformation<dim> transformation;
  GridTools::transform(transformation, *triangulation);
  triangulation->refine_global(2);
```
with transformation
```
template<int dim>
class Transformation
{
public:

  Point<dim> operator()(Point<dim> &x_in) const
  {
    Point<dim> x_out;

    for(unsigned int d=0; d<dim; ++d)
      x_out[d] = x_in[d] + 1.0;

    return x_out;
  }
};
```
resulting in
cylinder_affine_transform_refine2

In terms of documentation only, the main problem is that the documentation insinuates that `CylinderManifold` and `GridTools::transform()` are compatible. The correct behavior of functions of this type are discussed elsewhere in #8914, but here I want to point out that the documentation should not promise more than the software can offer. 

Playing around with the offset of the transformation (1.0 in the example above), you can obtain a nice heart-like geometry ;)
cylinder_affine_transform_2_refine2

One might now argue with some experience that one first needs to refine before applying ``GridTools::transform()``, i.e.,
```
  double const r = 0.5, L = 1.5;
  GridGenerator::cylinder(*triangulation, r, L);
  triangulation->refine_global(2);
  Transformation<dim> transformation;
  GridTools::transform(transformation, *triangulation);
```
resulting in (offset = 0.5)
cylinder_affine_transform_3_refine2

So one needs to explicitly deactivate the manifold in addition (see issue #8914)
```
  double const r = 0.5, L = 1.5;
  GridGenerator::cylinder(*triangulation, r, L);
  triangulation->refine_global(2);
  triangulation->reset_all_manifolds();
  Transformation<dim> transformation;
  GridTools::transform(transformation, *triangulation);
```
so that we get at least a cylinder with linear approximation of the boundary

cylinder_affine_transform_4_refine2

Unfortunately, this is not what the documentation suggests, since the documentation of ``GridTools::transform()`` (https://www.dealii.org/current/doxygen/deal.II/namespaceGridTools.html#a212e99cf0d923cebfa04f1d23fa60b04) says

> A safe approach is to use this function prior to any refinement in parallel, if that is possible, but not after you refine the mesh.

nfehn:
Closes #8920. 

Since no changes to the implemented will be made soon, the docu should correctly describe the current behavior.


include/deal.II/grid/grid_tools.h
```diff
@@ -425,9 +425,23 @@ namespace GridTools
...
    * some ghost cells may not. This means that computations like
-   * KellyErrorEstimator may give wrong answers. A safe approach is to use
-   * this function prior to any refinement in parallel, if that is possible,
-   * but not after you refine the mesh.
+   * KellyErrorEstimator may give wrong answers.
+   *
+   * @note This function is in general not compatible with manifolds attached
+   * to the triangulation. If you want to perform refinements according to a
+   * manifold description, you should first do the refinements, subsequently
+   * deactivate all manifolds, and finally call the transform() function. The
```
kronbichler, line 433: I don't think this statement on refinements is true in this generality - it is perfectly valid to refine the mesh after the transformation if one has a manifold for the *transformed* geometry at hand. So I would state this as
```suggestion
   * to the triangulation. If you want to perform refinements according to a
   * manifold description, you either need to attach the appropriate manifold
   * for the transformed geometry and clear possibly old manifolds, or
   * do the refinements before the transformation, deactivate all manifolds,
   * and finally call the transform() function. The
```

nfehn: I can incorporate this idea. However, I would reformulate it a bit because the sentence becomes quite complicated now.

nfehn:
Regarding your question, I agree principally, but I think we have issue #8914 to discuss these basic decisions and separate PRs to work on solutions. I think here we should focus on the documentation.

PR #10080, opened 2020-05-08
Undo accidental change in test result file with LAPACK

kronbichler:
This was accidentally changed in #10050 when marking this test as depending on LAPACK. Fixes the test failure
https://cdash.43-1.org/testDetails.php?test=43666160&build=6160


rezarastak:
Oops. I think when I rebased #10050 it changed that line and I missed it.

PR #10081, opened 2020-05-08
merge_triangulations(): reformulate documentation

nfehn:
Reformulate documentation in a way that keepping the manifolds is the expected default behavior.

This is a minimal solution to issue #8914 for release 9.2.


kronbichler:
Documentation only, so I'm merging.

PR #10084, opened 2020-05-08
Fix matrix-free pre/post loops with threads 

### LINKED ISSUE #9405: Post-function in matrix-free cell_loop() is not executed correctly when not using AdditionalData::none
When choosing to use a task scheme other than `AdditionalData::none` in the matrix-free environment, at least the post-loop-function is not executed correctly at all. I could not test the pre-loop-function yet.

When comparing the resulting vector after the application of the matrix-free `apply()`-function, according to the output there is no visible influence of any post-loop function. This can be tested using the attached code (had to rename it to .txt, github does not allow .cpp-files). When setting `use_scale_function` to `false` and `use_threads` to `true`, the resulting output vector (which can be printed using the commented function print_vector()) does not change, regardless of the applied function. It also is different compared to the output vector resulting from either `use_scale_function = true` (value for `use_threads` is not important here) or the output from `use_scale_function = false` and `use_threads = false`, which both are identical. 

After working on that problem for several days, would it be possible to at least add a warning, to spare others the debugging?
[main.txt](https://github.com/dealii/dealii/files/4087083/main.txt)

kronbichler:
Fixes #9405.

Given that we currently cannot track those dependencies in a thread-parallel loop, as we also promise for the [function interface already](https://dealii.org/developer/doxygen/deal.II/classMatrixFree.html#a6a53b93080a15c49d5513f09fc6bbb5d), we need to do the whole work before and after the loop over the cells.

This fixes behavior introduced after the last release, so no need for a changelog.

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -4568,13 +4568,26 @@ namespace internal
         {
           const internal::MatrixFreeFunctions::DoFInfo &dof_info =
             matrix_free.get_dof_info(dof_handler_index_pre_post);
-          AssertIndexRange(range_index,
-                           dof_info.cell_loop_pre_list_index.size() - 1);
-          for (unsigned int id = dof_info.cell_loop_pre_list_index[range_index];
-               id != dof_info.cell_loop_pre_list_index[range_index + 1];
-               ++id)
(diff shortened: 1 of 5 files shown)


kronbichler:
Yes, I was somehow faster to state the intended behavior in the documentation than actually make sure it actually is implemented like that :wink:

PR #10087, opened 2020-05-08
Disable hold_all_faces_to_owned_cells for FE_Q

peterrum:
This PR targets the setup of `MatrixFree` with a vector of `DoFHandler`s, which have been setup with `FE_Q` and `FE_DGQ` and enabling of the setting `hold_all_faces_to_owned_cells`.  This setup was not possible till now since `FE_Q` is not compatible with `hold_all_faces_to_owned_cells` and some partitioner cannot be setup. This PR skips the setup of these PRs only for the `FE_Q`-`DoFHandlers`.

### DIFF
--- a/tests/matrix_free/partitioner_02.cc
+++ b/tests/matrix_free/partitioner_02.cc
@@ -1,0 +1,130 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+
+// tests matrix-free partitioners for update_ghost_values and compress(add)
+
+
+#include <deal.II/base/mpi.h>
+#include <deal.II/base/quadrature_lib.h>
+
+#include <deal.II/distributed/tria.h>
+
+#include <deal.II/dofs/dof_handler.h>
+
+#include <deal.II/fe/fe_dgq.h>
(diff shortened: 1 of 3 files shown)


kronbichler:
Can you elaborate a bit what fails? I agree with the code and would like to have it in 9.2, but it is good to know if we can fix something more globally after the release.

peterrum:
If `hold_all_faces_to_owned_cells` is enabled, additional partitioners are set up here:

https://github.com/dealii/dealii/blob/5ea1283e4000ffc72a23e3a832ae3721b783af52/include/deal.II/matrix_free/matrix_free.templates.h#L1758-L1769

In the called functions, the degrees of freedom are collected of remote faces. However, for CG degrees of freedoms are found with index `-1`, which leads to the partitioner complaining here:

https://github.com/dealii/dealii/blob/5ea1283e4000ffc72a23e3a832ae3721b783af52/include/deal.II/matrix_free/matrix_free.templates.h#L1584-L1587

kronbichler:
Ah, I see now. The problem is that the faces on all cells call the `fu` function with the `flag=true` unconditionally, so the check that makes sure the `dof_indices_contiguous` only valid for DG elements is bypassed in this setup. In the `FE_Q` case, the regular faces would then fill `has_noncontiguous_cell=true` and simply use the basic partitioner.

tests/matrix_free/partitioner_02.cc
```diff
@@ -0,0 +1,132 @@
...
+}
+
+
+
+int
+main(int argc, char **argv)
+{
+  Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv, 1);
+
+  MPILogInitAll all;
```
kronbichler, line 106: Do you really need `MPILogInitAll` or would `initlog()` be enough? It seems you do not run with MPI, so one could skip that - on the other hand, now I wonder whether the code already triggered in the non-MPI case?

kronbichler:
@peterrum could you please rebase? CI seems to be stuck, but I would like it to run over this.

peterrum:
> CI seems to be stuck

Good old Java exceptions <3 Sigh...

PR #10093, opened 2020-05-08
Fix if-statements with check_for_distorted_cells.

dougshidong:
If `check_for_distorted_cells`, we should check for negative measures. Previously it was `!check_for_distorted_cells`.

Also replaced some `if (check_for_distorted_cells == true)`, to if `(check_for_distorted_cells)`. I don't see the advantage in readability (correct me if I'm wrong).

### DIFF
--- a/source/grid/tria.cc
+++ b/source/grid/tria.cc
@@ -2197,7 +2197,7 @@ namespace internal
               // to exist. If a cell has negative measure, then it must be
               // distorted (the converse is not necessarily true); hence
               // throw an exception if no such cells should exist.
-              if (!tria.check_for_distorted_cells)
+              if (tria.check_for_distorted_cells)
                 {
                   const double cell_measure =
                     GridTools::cell_measure<dim>(vertices,
@@ -4453,7 +4453,7 @@ namespace internal
                                   next_unused_cell,
                                   cell);
 
-                  if ((check_for_distorted_cells == true) &&
+                  if (check_for_distorted_cells &&
                       has_distorted_children(
                         cell,
                         std::integral_constant<int, dim>(),
@@ -8820,7 +8820,7 @@ namespace internal
 
                   // now see if we have created cells that are
                   // distorted and if so add them to our list
-                  if ((check_for_distorted_cells == true) &&
+                  if (check_for_distorted_cells &&
                       has_distorted_children(
                         hex,
                         std::integral_constant<int, dim>(),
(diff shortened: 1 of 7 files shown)


tjhei:
> Also replaced some `if (check_for_distorted_cells == true)`, to if `(check_for_distorted_cells)`. I don't see the advantage in readability (correct me if I'm wrong).

I agree.

dougshidong:
> That's a bug that's at least 10 years old...

Haha, I guess not many people "need" to create a grid with distorted cells, so always checking didn't hurt.

tjhei:
Can you please rebase to a current master?

dougshidong:
Done.

tjhei:
looks like we need to update some tests: https://jenkins.tjhei.info/blue/organizations/jenkins/dealii/detail/PR-10093/6/tests

dougshidong:
> looks like we need to update some tests: https://jenkins.tjhei.info/blue/organizations/jenkins/dealii/detail/PR-10093/6/tests

One option would be to simply delete the check on each cell's measure before building the Triangulation. Looks like no one really depended on it previously.

Any distorted cell the previous check would have caught will be caught later on by the more rigorous function `collect_distorted_coarse_cells()`.

Otherwise, this `cell->measure() > 0` check will always occur before the call to `collect_distorted_coarse_cells()`. If we want to keep `cell->measure() > 0`, the current tests would stress `cell->measure() > 0`, and we would need new tests that stress `collect_distorted_coarse_cells()` only.

tjhei:
@bangerth You are more familiar with these tests and the distorted cell logic. What do you propose? And can this be part of 9.2?

bangerth:
Too late in the game. Let's postpone to 9.3. It's a change that makes me nervous so close to the release.

dougshidong:
PR #10641 fixed this by simply removing the problematic statements.

peterrum:
Let's reopen this PR. I did not intend to delete these checks. I'll fix this later today!

dougshidong:
Sounds good. The new commit updated the tests. Those tests used to catch another exception thrown later on, but I have changed the comparison output file to the exception caught earlier on when it happens.

dougshidong:
Alright, fixed after they have been reintroduced by PR #10817.

peterrum:
Thanks @dougshidong!

dougshidong:
My pleasure. Weirdly enough, I needed to create an invalid grid and this prevented me from running it in debug.

PR #10100, opened 2020-05-08
Properly initalization of velocity vector

resetius:


examples/step-35/step-35.cc
```diff
@@ -796,11 +796,11 @@ namespace Step35
...
         vel_exact.set_component(d);
         VectorTools::interpolate(dof_handler_velocity,
-                                 Functions::ZeroFunction<dim>(),
+                                 vel_exact,
                                  u_n_minus_1[d]);
         vel_exact.advance_time(dt);
         VectorTools::interpolate(dof_handler_velocity,
-                                 Functions::ZeroFunction<dim>(),
+                                 vel_exact,
                                  u_n[d]);
```
kronbichler, line 804: The continuous integration tests complain that you should move the arguments for this second call all to the same line. Have you used the automatic indenting of deal.II before? If you're interested, you can look that up here:
https://www.dealii.org/developer/doxygen/deal.II/CodingConventions.html

resetius: Thank you. I formated the code and updated the PR.

kronbichler:
This is a good bugfix. I checked the output and they are essentially the same - the incompressible flow is mostly driven by the inflow condition and initial pressure, so the flow did develop similarly. Thus, I'm adding it to the imminent 9.2 release.

resetius:
Thanks. I'm working on the similar incompressible flow problem and found out this strange initialization behaviour :)

bangerth:
If this was a bug before, would you be interested in writing an entry in `doc/news/changes/minor` along the lines of the other entries in this directory, and propose it as a separate pull request?

resetius:
sure

PR #10102, opened 2020-05-09
Qualify a call to parallel::apply_to_subrange

kronbichler:
`clang-tidy` reports in #9883 an error in a function introduced in #10084 - there it did not trigger, interestingly. To avoid ambiguity, it is good to fully qualify the call.


peterrum:
CI is making troubles again. I would suggest to merge this PR nevertheless.

kronbichler:
We need to get things going again, so I'm merging.

PR #10103, opened 2020-05-09
Mark LaplaceOperator::apply_add as override

kronbichler:
In #9883 for `step-50` we got another error about a missing `override` statement. I went through the class in question again and I hope we now got them all.


kronbichler:
This needs #10102 to pass the `tidy` test.

PR #10108, opened 2020-05-10
Add mapping argument to find_closest_vertex_of_cell(). 

### LINKED ISSUE #7922: GridTools::find_closest_vertex_of_cell ignores mapping
GridTools::find_closest_vertex_of_cell does not consider the fact that vertices may be in a different position if a Mapping is present that transforms the cells.

luca-heltai:
Closes #7922.

### DIFF
--- a/tests/grid/grid_tools_find_closest_vertex_02.cc
+++ b/tests/grid/grid_tools_find_closest_vertex_02.cc
@@ -60,7 +60,7 @@ test(const Point<spacedim> &p, double displacement)
       const auto &v = mapping.get_vertices(cell);
 
       deallog << "Closest vertex of cell " << cell << " to " << p
-              << " is cell->vertex(" << i << ") :" << v[i] << std::endl;
+              << " is cell->vertex(" << i << ") : " << v[i] << std::endl;
     }
 };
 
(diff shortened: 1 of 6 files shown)


tests/grid/grid_tools_find_closest_vertex_02.cc
```diff
@@ -0,0 +1,78 @@
...
+  MappingQEulerian<dim, Vector<double>, spacedim> mapping(1, dh, disp);
+
+  for (const auto cell : tria.active_cell_iterators())
+    {
+      const auto i =
+        GridTools::find_closest_vertex_of_cell<dim, spacedim>(cell, p, mapping);
+      const auto &v = mapping.get_vertices(cell);
+
+      deallog << "Closest vertex of cell " << cell << " to " << p
+              << " is cell->vertex(" << i << ") :" << v[i] << std::endl;
```
kronbichler, line 63: Can you add a space here?
```suggestion
              << " is cell->vertex(" << i << ") : " << v[i] << std::endl;
```

luca-heltai:
> Then, we have other functions like find_closest_vertex, find_active_cell_around_point that have the mapping first. Does anyone else have a different opinion? We won't clean up those interfaces for the release, so the task is mostly to find what fits best. @luca-heltai what is your opinion?

I think the only advantage of having `mapping` as the first argument is when there are several default arguments, before the actual mapping. We could make a cutoff: if there are more than 3 arguments (besides the mapping), we provide two versions (one with mapping in front and one without, coherently with the rest of the library). Otherwise we keep it at the last argument. But I'm not really happy with this solution either...

Issue #10073, opened 2020-05-08
deal.II Release 9.2.0 - Testsuite status


tamiko:
Jenkins CI:
---
Linux: [![Build Status](https://jenkins.tjhei.info/job/dealii/job/dealii-9.2/badge/icon)](https://jenkins.tjhei.info/job/dealii/job/dealii-9.2/)   OSX: [![Build Status](https://jenkins.tjhei.info/job/dealii-OSX/job/dealii-9.2/badge/icon)](https://jenkins.tjhei.info/job/dealii-OSX/job/dealii-9.2/)   tidy: [![Build Status](https://jenkins.tjhei.info/job/dealii-tidy/job/dealii-9.2/badge/icon)](https://jenkins.tjhei.info/job/dealii-tidy/job/dealii-9.2/)

Regression tester:
---

tamiko:
(comment reserved)

PR #10126, opened 2020-05-10
cmake: restrict p4est tests

tamiko:
Only run these tests if p4est is set up to output ascii


tamiko:
In reference to #10073

PR #10127, opened 2020-05-10
python_bindings: Also update order of compute_aspect_ratio

tamiko:
In reference to #10124
See also: https://cdash.43-1.org/viewBuildError.php?buildid=6258


tamiko:
I do not think that the CI enable the python-wrapper component.

PR #10134, opened 2020-05-11
Make new GridGenerator images appear in the online documentation

kronbichler:
Some figures introduced in #9822 and #9802 do not appear in the online documentation:
https://dealii.org/developer/doxygen/deal.II/namespaceGridGenerator.html#a712f95340c7002afbd5d6fb755e12a61
I think we need to add some specific html/doxygen command as elsewhere in the file. I hope this does the trick - it works locally, but it used to work locally also before.

### DIFF
--- a/include/deal.II/grid/grid_generator.h
+++ b/include/deal.II/grid/grid_generator.h
@@ -749,10 +749,16 @@ namespace GridGenerator
    * <table align="center" class="doxtable">
    *   <tr>
    *     <td>
-   *       <img src="hyper_ball_balanced_2d.png" alt="" width="40%">
+   *       \htmlonly <style>div.image
+   *         img[src="hyper_ball_balanced_2d.png"]{width:40%}</style>
+   *       \endhtmlonly
+   *       @image html hyper_ball_balanced_2d.png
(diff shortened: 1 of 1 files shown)


jppelteret:
I think that there might be a general problem with the online documentation. In all of my browsers I get a warning about an expired security certificate, and if I bypass the warning then most of the images in the tutorials are not rendered for both the "current" and the developer page.

include/deal.II/grid/grid_generator.h
```diff
@@ -749,10 +749,16 @@ namespace GridGenerator
    * <table align="center" class="doxtable">
    *   <tr>
    *     <td>
-   *       <img src="hyper_ball_balanced_2d.png" alt="" width="40%">
+   *       \htmlonly <style>div.image
+   *         img[src="hyper_ball_balanced_2d.png"]{width:40%}</style>
+   *       \endhtmlonly
+   *       @image html hyper_ball_balanced_2d.png
```
tjhei, line 755: I assume this is because an ```` tag won't make doxygen export the .png and make it available online.

PR #10139, opened 2020-05-11
Simplify internal::DoFHandlerImplementation::Policy::ParallelDistributed::distribute_dofs_on_level()

peterrum:
Similar to #9644 but for multigrid levels.

### DIFF
--- a/source/dofs/dof_handler_policy.cc
+++ b/source/dofs/dof_handler_policy.cc
@@ -1659,17 +1659,15 @@ namespace internal
           // Step 1: distribute dofs on all cells, but definitely
           // exclude artificial cells
           types::global_dof_index next_free_dof = 0;
-          typename DoFHandlerType::active_cell_iterator
-            cell = dof_handler.begin_active(),
-            endc = dof_handler.end();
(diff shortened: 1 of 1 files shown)


tjhei:
can this wait to after 9.2?

peterrum:
@tjhei Yes.

peterrum:
What I have noticed that the dofs indices on the levels are not cached. Is there a reason for that?

peterrum:
@kronbichler I would be also in favor of only storing the dofs indices in a caches. However, we also have think about how we should query the dof indices of the other entities: this is currently possible.

To access the dofs on the n-th face, line, vertex of a cell is no problem if we have the appropriate indices of the *reference element*.

The problem is rather how to access the n-th face, line, vertex globally if only the cached information is available. Furthermore, currently, we only have the adjacency information from the triangulation given: *cell->face->line->vertex*. So we do not have a possibility to determine, e.g. that line X is the x-th line in cell Y (this would enable to use the cache of cell Y as described above)... There are many ways to solve this problem...

PR #10140, opened 2020-05-11
add a mailmap entry

tamiko:


drwells:
Both of these authors need slight adjustments to their name formatting so :+1:

PR #10142, opened 2020-05-11
Changelog

resetius:
https://github.com/dealii/dealii/pull/10100


drwells:
Thanks again for the fix!

bangerth:
Thanks very much, and welcome to the list of authors of deal.II!

resetius:
> Thanks very much, and welcome to the list of authors of deal.II!

Oh, I've already been here since 2017 :) But now I'm here twice :) https://www.dealii.org/authors.html
Alexey Ozeritskiy
Alexey Ozeritsky
This is one person recorded according to different transliteration rules :) You can leave any one record.

bangerth:
Oh, interesting :-) In that case, good to see you back!

What version of the name do you use professionally? That's the one we should leave.

resetius:
Alexey Ozeritskiy

bangerth:
So it happened! Thanks for the feedback.

PR #10143, opened 2020-05-11
Tests opencascade/*: Add three output variants

### LINKED ISSUE #10086: opencascade/stl_* failing
The opencascade/stl_* tests are all failing on all configurations in the CI, but locally it works for me with oce 6.9.1. The CI is using `OPENCASCADE_VERSION: 6.8.0`. We might want to restrict the version for that feature.

tamiko:
All tests produce a visually correct output. The test fails due to

 - a different mesh generator used in opencascade 7.4.0 placing vertices
   slightly differently

 - a slightly different refinement behavior when linking against libc++
   versus libstdc++

For now, just be pragmatic and add an output variant for each. We might
have to find a smarter solution in the future

In reference to #10073
Closes #10086
See also: https://cdash.43-1.org/testSummary.php?project=1&name=opencascade%2Fstl_creation_single_patch.debug&date=2020-05-11
See also: https://cdash.43-1.org/testSummary.php?project=1&name=opencascade%2Fstl_creation_watertight.debug&date=2020-05-11
See also: https://cdash.43-1.org/testSummary.php?project=1&name=opencascade%2Fstl_refinement.debug&date=2020-05-11


tamiko:
@nicola-giuliani The test works properly - the output is unfortunately not stable. :neutral_face:

Issue #10146, opened 2020-05-11
Compilation failure with gcc-4.9.4


tamiko:
@kronbichler *ping*, @bangerth *ping*

With the latest changes on the matrix-free code we are now failing to compile with gcc-4.9.4 with an internal compiler error.
https://cdash.43-1.org/viewBuildError.php?buildid=6296

I will try to work around this issue in a moment - I think it is valuable to maintain gcc-4.9.4 compatibility, at least for this release.

Might be related to our old friend: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54844
Alternative candidate: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65186

bangerth:
Do we know which patch caused this?

kronbichler:
Would it help if we create a separate variable for the template argument,
```
constexpr n_lanes_double = (std::is_same<Number, float>::value && n_lanes > 1) ?
                           n_lanes / 2 : n_lanes
```
The code in question comes from #9656, but that has been merged for a while. Where we failing since March 28? Or was it only the combination with the recently merged #9835 that triggered the ICE?

PR #10149, opened 2020-05-11
Fix compilation with gcc-4.9.4

tamiko:
In reference to #10073
Closes: #10146


tjhei:
maybe add a comment that this is so awkwardly written because of gcc 4.9.4?

PR #10150, opened 2020-05-11
Test petsc/reinit_preconditioner_0?: disable for 64bit indices

tamiko:


kronbichler:
Do you want to do the same for `reinit_preconditioner_02`:
https://cdash.43-1.org/testDetails.php?test=44073965&build=6288

tamiko:
@kronbichler Good point. Done.

PR #10155, opened 2020-05-12
Instantiate OpenCASCADE

### LINKED ISSUE #10006: Instantiate all OpenCASCADE classes also for dim=2, spacedim=2
Some of these are still missing. It should be relatively straight forward to do.

masterleinad:
Fixes #10006.


tamiko:
We don't. @masterleinad I will trust you that this doesn't blow up™ :smile:

PR #10157, opened 2020-05-12
Release step 5: update VERSION

tamiko:

PR #10158, opened 2020-05-12
Release step 6a: Require the current version of deal.II

tamiko:


drwells:
This should be merged after #10157.

PR #10161, opened 2020-05-12
Removed two variables which were not used in particles/utilities.h

blaisb:
There are two variables that are declared in particles/utilities.h but that are not used.
As soon as this function will be used, this will lead to two warnings during compilation. This is my fault, since I wrote that function. Merely correcting my mistakes...

### DIFF
--- a/include/deal.II/particles/utilities.h
+++ b/include/deal.II/particles/utilities.h
@@ -206,8 +206,6 @@ namespace Particles
       const auto &tria     = field_dh.get_triangulation();
       const auto &fe       = field_dh.get_fe();
       auto        particle = particle_handler.begin();
-      const auto  max_particles_per_cell =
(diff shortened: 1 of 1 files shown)


tjhei:
This should be cherry-picked to 9.2 as soon as the CI passes.

bangerth:
We'll need a place to collect which PRs these are that we want to cherry-pick...

masterleinad:
> /rebuild

@bangerth wrong order. :slightly_smiling_face:

bangerth:
This system makes no sense.

@tjhei: Why is it that way? Why can't we just test whether two conditions are satisfied, rather than whether one tag was set before a comment was made, or the other way around?

tamiko:
@bangerth Have a look at #10072 - we have everything in one place now.

PR #10163, opened 2020-05-12
added support for quad9 and hex27 elements

pauknerd:
This PR introduces the support to read quadratic elements (9-noded quadrangles and 27-noded hexahedra) from Gmsh files through modification of the GridIn::read_msh function. The additional points, or support points, are stored and can be further used by the user.
Additionally, a new mapping class called MappingQ2 is introduced which can process the previously described vector containing the support points to construct a quadratic mapping. To this end, the method compute_mapping_support_points is modified to return the support points read from the Gmsh input file. Note that so far, mesh refinement is not supported but is work in progress.


peterrum:
The workflow looks like this:

```cpp
  Triangulation<dim, spacedim> tria;

  GridIn<dim, spacedim> gi;
  gi.attach_triangulation(tria);

  std::vector<std::vector<Point<spacedim>>> support_points; // vector for additional 
                                                            // support points

  std::filebuf fb;
  if (fb.open(file_name, std::ios::in))
    {
      std::istream is(&fb);
      gi.read_msh(is, support_points); // read triangulation and store additional 
                                       // support points into vector
    }

  MappingQ2<dim, spacedim> mapping_2(support_points); // new mapping class
```

bangerth:
Let's postpone until after we're through with the release. I have some ideas about the interface following a recent patch where reading auxiliary data from VTK files (#8575) was added. We should come up with a common scheme.

masterleinad:
```
Running indentation test on Pull Request #
invalid author 'DanielPaukner' without firstname and lastname

```

include/deal.II/fe/mapping_q2.h
```diff
@@ -0,0 +1,91 @@
...
+ * quadrilateral/hexahedron given 9-noded quadrangles or 27-noded hexahedra.
+ *
+ * The mapping implemented by this class is similar to creating a mapping of
+ * the type MappingQ(2) or MappingQGeneric(2). However, the difference is that
+ * this mapping gets the necessary support points from the input file read
+ * with read_msh() instead of computing them or getting them from a manifold.
+ *
+ * @note This class does not support mesh refinement yet. Also, only Gmsh
+ * input files can currently be read and processed to give the necessary
+ * vector of support points.
```
kronbichler, line 42: Wouldn't it make sense to integrate this into `MappingQCache` - that class already does exactly this kind of data structure?

peterrum: Would you suggest to derive from `MappingQCache` or add a new `MappingQCache::initialize()` function, which takes the additional support poins?

kronbichler: I would prefer to simply add a constructor - it is useful for any order of the mapping to be able to provide a set of points. We have the lambda already, but there is no reason against adding another one already holding the array. They can come from a mesh optimization algorithm, for example. One thing we should teach this class - it goes a bit into the direction suggested by @DanielPaukner regarding refinement - is to insert the interpolation facilities between different mesh levels to this class. This way, we would also enable to only fill the active cell positions and interpolate them down to the coarser level for MG adaptivity.

kronbichler:
Regarding refinement, we need to teach the `Triangulation` class to refine according to a `Mapping`. The standard interface is to create a `Manifold` from the `Mapping` and redefine `get_new_point`; not sure how efficient that would be because we need to search among all cells to some given points, so this likely needs some `GridTools` candy via `Cache` to avoid quadratic complexity in the number of cells.

include/deal.II/fe/mapping_q2.h
kronbichler: Of course I meant to say `initialize()` function rather the constructor.

DanielPaukner:
Thank you all for the comments and especially the hints regarding mesh refinement @kronbichler ! I will have a look at that and see what I can do!

blaisb:
> Thank you all for the comments and especially the hints regarding mesh refinement @kronbichler ! I will have a look at that and see what I can do!

By the way, this is a feature I have been dying for. This will be an amazingly useful contribution :)!

masterleinad:
Don't forget to fix your `git` author information at some point. :slightly_smiling_face:

DanielPaukner:
Hi all,
I made some changes to prepare removing the MappingQ2 class I introduced in the first commit and use the MappingQCache class instead. Currently, both are present. I added a new initialize() function which takes the vector of support points read from GridIn::read_msh() and fills the cache using a lambda function. However, currently only the first level is filled since I am not sure how to include mesh refinement...

There are a few things that are unclear to me and I would appreciate your input/comments/ideas:
- @kronbichler  you suggested to create a Manifold from a Mapping and redefine get_new_point. One thing that I am wondering though, is how the new vertices should be placed... Should they be placed "in the middle" of the original two vertices defining an edge? And if not placed "in the middle", how could we determine where to place them? The same questions apply to the support points. 
Initially, I though maybe the support points on level 'n' could become the new vertices on level 'n+1' but I am not sure if that is too naive and introduces unnecessary restrictions...

- Another thing that is a bit unclear to me is the workflow of the code.
How I understood it from the comments: We add functions to the MappingQCache class that compute the new vertices and support points if the mesh is refined and then store them in the cache. And to give the triangulation the opportunity to get these points we create a Manifold from the MappingQCache class and redefine get_new_point() to return a point from the cache of MappingQCache. Did I understand that correctly or did I mix something up?

Related to that, how does the mapping know which cells to refine? Do we just compute the refinement of every cell on every level but just not use it if the cell is not flagged for refinement?

- @bangerth in your comment above you mentioned that something related was implemented and we should try to use a common scheme. What ideas did you have in mind?

Thank you all for your help!

peterrum (commented):
@DanielPaukner Could you remove the old implementation and rebase with the current version of deal.II. 

From my perspective this PR looks good. Since I was partly involved in it, I would be happy if someone else could review it. 

I would be happy if we would soon find a good flexible solution here since I would like to use the solution found here as a blue print for high-order mappings for TETs.

drwells:
I like this functionality a lot and its something I want to see in the library.

I don't have time to implement what I believe is the right way to do things so I will not block merging this patch in its current form but I want to at least voice my concerns. Ultimately we might want to implement similar functionality for other element types. Similarly, it's not ideal that this only works with GMSH input and uses a data format (huge array of support points) that isn't like anything else in the library.

I would very much prefer if we took the MSH input code and implemented a new class that knows how to read in a finite element field and populate both a `Vector` (or other vector class) and a `Triangulation`. From there we could use `MappingFEField` in the 'normal' way. We could get good performance back by getting `MappingFEField` to work with `MappingQCache`. Implementing it this way would get us automatic support for all other elements as long as we can set up a finite element solution vector.

DanielPaukner:
@peterrum the old implementation has been removed.

@drwells Thank you for your hints! The way you suggest sounds more flexible and I will discuss the topic further with Peter. For now, this could be a temporary solution which could/should definitely be changed in the future.

dougshidong:
Might be going off-toping but I'd also like to chime in this popular feature after seeing this PR being referred in the deal.II forums https://groups.google.com/g/dealii/c/qYRHgKgZ0E4. 

I have a basic/spaghetti implementation of what @drwells mentionned with MappingFEField and I can attest that it works very well for multiple reasons. Of course it doesn't preclude this PR, but I think it's worth thinking before we start manually implementing quad16, quad25, etc.

- It allows an arbitrarily high geometric order. 
- Since the grid is basically a whole other set of DoFs, we can apply some elasticity equations to deform the grid (I'm thinking shape optimization or fluid-structure interaction).
- Refinement is easy and guaranteed to work (if your initial polynomial grid is a valid curvilinear grid) by using the SolutionTransfer of the vector within the MappingFEField. The Triangulation's associated Manifold won't matter since the Triangulation's sole purpose is to host cell connectivity information; for all we care, all the Triangulation's vertices could be at the origin.
- If we can get MappingFEField to work with hp::DoFHandler, we can end up with the ideal cost scenario of curved elements near the surface and Q1 elements in the domain.

The author of the deal.II post is actually is part of our lab, so I can ask him to do a PR if we have some code to read an arbitrarily high order GMSH grid into a MappingFEField vector.

peterrum:
> The author of the deal.II post is actually is part of our lab, so I can ask him to do a PR if we have some code to read an arbitrarily high order GMSH grid into a MappingFEField vector.

@dougshidong Sorry for the late rely. Feel free to propose an alternative implementation!

drwells:
We should try to get this done soon but it's not going to happen in time for the release.

bangerth:
What's the status of this patch? It has merge conflicts now -- @DanielPaukner, would you be willing to bring it up to date so we can move it across the finish line?

peterrum:
@bangerth The problem here is that we did not find a consensus on dealing with such kind of meshes. Generally, we don't have good ways to deal with additional informations from mesh files.

drwells:
Yes. Setting things up as a Manifold would be an alternative to `MappingFEField`, though I prefer the later.

Moving the gmsh reader code into a `DataIn` class would be great because then we can start thinking about how we want to load finite element data in from external projects or file formats. That's a bigger discussion than this PR, though.

bangerth:
We didn't find a consensus here on how to implement this feature. I suspect that we also want something similar in the VTK interfaces (@luca-heltai). How do we want to proceed?

PR #10166, opened 2020-05-12
examples/step-69: fix texttt formulas

tamiko:


tamiko:
I just noticed this while looking at https://www.dealii.org/developer/doxygen/deal.II/step_69.html

tjhei:
documentation only, so we can merge. Are you going to cherry-pick to 9.2?

PR #10170, opened 2020-05-13
[9.2] remove test mpi/p4est_bug

tamiko:
This test will actually make it past the expect=run stage if p4est is
not compiled with debug asserts. This is unfortunately independent of
whether deal.II is in release or debug mode.
 
Supersedes #10164
In reference to #8603


tjhei:
:-(, I guess that makes sense for 9.2. No need to run any tests.

PR #10174, opened 2020-05-13
Fix find active cell around point.

luca-heltai:
Working on step-70, we hit this bug which makes running particle based applications impossible on large clusters.

I'm not sure yet hat is wrong. For the moment, this is a failing test. The (correct) output is generated using line 46.

@gassmoeller @blaisb 

This is one of the *two* bugs we uncovered with step-70.

My educated guess is that the version that uses a `Cache` object also does things differently, like passing a tree of locally owned vertices, and therefore it fails to find the points when they are on artificial cells.

I'm going to bed... tomorrow I'll try to see if I can solve this.


tests/grid/find_active_cell_around_point_04.cc
```diff
@@ -0,0 +1,49 @@
...
+      Triangulation<2>::smoothing_on_coarsening));
+
+  GridGenerator::hyper_cube(triangulation, -1, 1);
+  triangulation.refine_global(5);
+  GridTools::Cache<2> cache(triangulation);
+
+  Point<2> p(0.239367, 0.341747);
+
+  // auto res = GridTools::find_active_cell_around_point(triangulation, p);
+  auto res = GridTools::find_active_cell_around_point(cache, p).first;
```
bangerth, line 47: Make everyone's life easier and put these two statements into different variables and then
```
  Assert (res1 == res2, ExcInternalError());
```
This way it's clear in the test and in the test result what is correct and what is not.

luca-heltai:
This test currently fails. I'm trying to figure out a fix.

luca-heltai:
The fix is dirty: I did not understand what is wrong with the actual algorithm. This algorithm was developed by @gassmoeller and myself during a deal.II developer workshop, and it seems to work *most* of the times, but sometimes it fails. I have not understood why, so in the fix I revert to the other algorithm, which is less smart and much slower, but seems to do the job correctly. At least the test now is a tautology, since the second algorithm is what is checked for in the first place...

bangerth:
Hm, but then we have a regression-of-sorts. That's disappointing as well. We should really try to find out what the actual problem is...

luca-heltai:
> Hm, but then we have a regression-of-sorts. That's disappointing as well. We should really try to find out what the actual problem is...

I agree. But there are two bugs now in the way of step-70. I'd like to try and fix the other one as well, and see if we manage to get going. At least now the behavior is correct. We should keep the issue open, even if we decide to merge this.

bangerth:
In order to assess how to move forward: What's the other bug you run into?

luca-heltai:
@bangerth : #10211

tamiko:
@bangerth Please re-review and accept if appropriate. I would like to get this into rc2.

kronbichler:
@bangerth let us move forward here. We have time later to understand why the slow path is needed.

bangerth:
@luca-heltai -- can you open a PR that reminds us that we need to investigate this?

kronbichler:
@bangerth we have #10174 open. Let me put the description of what we should improve to that PR and adjust the target milestone to 9.3.

PR #10178, opened 2020-05-13
doxygen: Strip full @CMAKE_SOURCE_DIR@ and @CMAKE_BINARY_DIR@

tamiko:
The current configuration fails to strip the repository path from source
directories, e.g., on the homepage we have for example:

"""
The documentation for this struct was generated from the following files:

    deal.II/lac/petsc_precondition.h
    /home/dealii/dealii/source/lac/petsc_precondition.cc
"""

With this change this now reads:

"""
The documentation for this struct was generated from the following files:

    include/deal.II/lac/petsc_precondition.h
    source/lac/petsc_precondition.cc
"""


bangerth:
Nice improvement!

PR #10182, opened 2020-05-13
clarify documentation of FEEvaluation::set_dof_values()

nfehn:
set_dof_values() might touch non-local degrees of freedom, which is not intuitive to me and should therefore be highlighted in the documentation.

The order of paragraphs in the documentation has been changed according to the order of function arguments. The same should be done in my opinion for the other FEEvaluation functions with similar documentation, and using @note for the "if called without a MatrixFree object" explanation.

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -165,7 +165,14 @@ public:
    * temporary solution should always have homogeneous constraints and this
    * method is the correct one.
    *
-   * If this class was constructed without a MatrixFree object and the
+   * If the given vector template class is a block vector (determined through
+   * the template function 'IsBlockVector<VectorType>::value', which checks
+   * for vectors derived from dealii::BlockVectorBase) or an
+   * std::vector<VectorType> or std::vector<VectorType *>, this function reads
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -264,7 +264,19 @@ class FEEvaluationBase
    * by the current cell are overwritten. Thus, if a degree of freedom is
    * associated to more than one cell (as usual in continuous finite
    * elements), the values will be overwritten and only the value written last
-   * is retained. The mask can be used to suppress the write access for some
+   * is retained. Please note that in a parallel context this function might
+   * also touch non-local degrees of freedom, so that a subsequent update or
```
kronbichler, line 268:
```suggestion
   * also touch degrees of freedom owned by other MPI processes, so that a subsequent update or
```

nfehn: What about the other functions, do it here or elsewhere?

kronbichler: Let us do it here. I agree with that suggestion.

kronbichler:
can you squash the commits into one? Then we're ready to merge.

PR #10184, opened 2020-05-13
[9.2] Take over #10183: Fix missing closing braces.

masterleinad:


tjhei:
Works on master, so we can just merge.

PR #10185, opened 2020-05-13
[9.2] Take over #10179: contrib/utilities/makeofflinedoc.sh: Also install jpg

tamiko:


drwells:
The equivalent is already merged on main and I am 99.9% sure that the CI doesn't build/check the offline documentation so lets just merge.

PR #10189, opened 2020-05-13
Documentation: Treat code gallery projects as examples

tamiko:
Installing the code gallery projects as part of the html documentation
is not ideal. Let us install the code gallery projects as a sub
directory under the EXAMPLES_RELDIR instead:

  examples/code-gallery/*
  examples/step-*

[ I have updated the release tasks to also bundle all examples again, i.e.
we will end up with a tarball for the release containing the code-gallery
examples ]


bangerth:
Does this affect installing the code gallery programs, or the documentation that results from the code gallery programs?

tamiko:
@bangerth @masterleinad This only affects the (optional) installation, not the generation of the documentation. For reference, it avoids creating this directory: https://www.dealii.org/developer/doxygen/code-gallery/

bangerth:
I had no idea that we actually install the code gallery :-) But I found it now, and you're totally correct that it has no business existing -- in the BUILD directory, it should only contain the temp files that doxygen eats.

bangerth:
I don't think this affects any tests, so let's merge.

PR #10190, opened 2020-05-13
Run Travis on all branches

masterleinad:
Following https://github.com/dealii/dealii/pull/10173, I don't see a reason to not run `Travis` on all branches.


tjhei:
how does this work, does this need to be cherry-picked to 9.2?

PR #10191, opened 2020-05-13
Documentation: fix make_gallery.pl

tamiko:
Otherwise newer doxygen does not parse the generated *.h files...
See #8972


tjhei:
cherry-pick?

tamiko:
@tjhei Yup, see #10192

PR #10194, opened 2020-05-13
Avoid using std::cout in a test.

drwells:
This is problematic on misconfigured MPI implementations that print warnings for one reason or another every time they are run (e.g., the default MPI implementation on Ubuntu 16.04).

It is easy enough in this case to do the normal thing and print everything to deallog.

I think that the small number of failing MPI tests on this platform all fail for this reason.

### DIFF
--- a/tests/mpi/petsc_step-27.cc
+++ b/tests/mpi/petsc_step-27.cc
@@ -171,7 +171,7 @@ namespace Step27
     , triangulation(mpi_communicator)
     , dof_handler(triangulation)
     , max_degree(dim <= 2 ? 7 : 5)
-    , pcout(std::cout,
+    , pcout(deallog.get_file_stream(),
             (Utilities::MPI::this_mpi_process(mpi_communicator) == 0))
   {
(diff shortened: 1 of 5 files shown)


tests/mpi/petsc_step-27.cc
```diff
@@ -329,24 +329,24 @@ namespace Step27
...
     preconditioner.initialize(system_matrix, data);
 
-    check_solver_within_range(cg.solve(system_matrix,
-                                       completely_distributed_solution,
-                                       system_rhs,
-                                       preconditioner),
-                              solver_control.last_step(),
-                              5,
-                              40);
+    cg.solve(system_matrix,
```
tjhei, line 346: why this change?

drwells: Since the test copies stdout/stderr to `output` (i.e., things printed to deallog were ignored) that check never made it to the stored `.output` files. I deleted it so that the test had the same output (aside from blank lines) as it did before.

PR #10195, opened 2020-05-13
Doxygen: Fix include strip

tamiko:
This fixes a bug introduced in #10187

I verified that this now yields the intended behavior: inlude statements
are written as `#include<deal.II/...>` and files are referenced as
`include/...` and `source/...`.


tjhei:
let's wait for travis before merging.

PR #10198, opened 2020-05-13
Updates to step-50

bangerth:
This is a follow-up to #9883 with quite a large number of text additions and edits, plus a whole bunch of updates to the code.

The updates to the code are mostly to variable names. The most significant change (also the largest because it affects whitespace) is to replace
```
  if (settings.solver == Settings::gmg_mf)
   ...
  else if (settings.solver == Settings::gmg_mb)
    ...
```
by a `switch` statement. I broke all of the code changes into individual commits that are self-contained to make reviewing simpler if you want to see individual changes (but also because I needed to bisect where I had introduced a bug :-( ).

I would really like to still get this into 9.2 if possible.

### DIFF
--- a/examples/step-50/doc/intro.dox
+++ b/examples/step-50/doc/intro.dox
@@ -1,7 +1,7 @@
 <br>
 
 <i>
-This program was contributed by Thomas C Clevenger and Timo Heister.
+This program was contributed by Thomas C. Clevenger and Timo Heister.
 <br>
 This material is based upon work partly supported by the National
 Science Foundation Award DMS-2028346, OAC-2015848, EAR-1925575, by the Computational
@@ -19,11 +19,32 @@ libraries is described in the <a href="../../readme.html" target="body">README</
 <h1>Introduction</h1>
 
 
-This example shows the usage of the multilevel functions in deal.II on distributed
+This example shows the usage of the multilevel functions in deal.II on
+parallel, distributed
 meshes and gives a comparison between geometric and algebraic multigrid methods.
 The algebraic multigrid (AMG) preconditioner is the same used in step-40. Two geometric
 multigrid (GMG) preconditioners are considered: a matrix-based version similar to that
-in step-16 (but for parallel computations) and a matrix-free version discussed in step-37.
+in step-16 (but for parallel computations) and a matrix-free version
+discussed in step-37. The goal is to find out which approach leads to
+the best solver for large parallel computations.
+
+Algebraic multigrid methods are obviously the easiest to implement
+with deal.II since classes such as TrilinosWrappers::PreconditionAMG
+and PETScWrappers::PreconditionBoomerAMG are, in essence, black box
+preconditioners that require only a couple of lines to set up even for
+parallel computations. On the other hand, geometric multigrid methods
+require changes throughout a code base -- not very many, but one has
+to know what one is doing.
+
+What the results of this program will show
+is that algebraic and geometric multigrid methods are roughly
+comparable in performance <i>when using matrix-based formulations</i>,
+and that matrix-free geometric multigrid methods are vastly better for
+the problem under consideration here. A secondary conclusion will be
+that matrix-based geometric multigrid methods really don't scale well
+strongly when the number of unknowns per processor becomes smaller than
+20,000 or so.
+
 
 <h3>The testcase</h3>
 
@@ -33,14 +54,22 @@ We consider the variable-coefficient Laplacian weak formulation
 @f}
 on the domain $\Omega = [-1,1]^\text{dim} \setminus [0,1]^\text{dim}$ (an L-shaped domain
 for 2D and a Fichera corner for 3D) with $\epsilon = 1$ if $\min(x,y,z)>-\frac{1}{2}$ and
-$\epsilon = 100$ otherwise. The boundary conditions are $u=0$ on the whole boundary and
-the right-hand side is $f=1$. We use continuous Q2 elements to discretize $V_h$ and use a
+$\epsilon = 100$ otherwise. In other words, $\epsilon$ is small along the edges
+or faces of the domain that run into the reentrant corner, as will be visible
+in the figure below.
+
+The boundary conditions are $u=0$ on the whole boundary and
+the right-hand side is $f=1$. We use continuous $Q_2$ elements for the
+discrete finite element space $V_h$, and use a
 residual-based, cell-wise a posteriori error estimator
 $e(K) = e_{\text{cell}}(K) + e_{\text{face}}(K)$ from @cite karakashian2003posteriori with
 @f{align*}
- e_{\text{cell}}(K) = h^2 \| f + \epsilon \triangle u \|_K^2, \qquad
- e_{\text{face}}(K) = \sum_F h_F \| [ \epsilon \nabla u \cdot n ] \|_F^2.
+ e_{\text{cell}}(K) &= h^2 \| f + \epsilon \triangle u \|_K^2, \\
+ e_{\text{face}}(K) &= \sum_F h_F \| \jump{ \epsilon \nabla u \cdot n } \|_F^2,
 @f}
+to adaptively refine the mesh. (This is a generalization of the Kelly
+error estimator used in the KellyErrorEstimator class that drives mesh
+refinement in most of the other tutorial programs.)
 The following figure visualizes the solution and refinement for 2D:
 <img width="400px" src="https://www.dealii.org/images/steps/developer/step-50-2d-solution.png" alt="">
 In 3D, the solution looks similar (see below). On the left you can see the solution and on the right we show a slice for $x$ close to the
@@ -56,9 +85,25 @@ center of the domain showing the adaptively refined mesh.
   </tr>
 </table>
 Both in 2D and 3D you can see the adaptive refinement picking up the corner singularity and the inner singularity where the viscosity jumps, while the interface along the line that separates the two viscosities is (correctly) not refined as it is resolved adequately.
+This is because the kink in the solution that results from the jump
+in the coefficient is aligned with cell interfaces.
 
 
-<h3>Workload imbalance</h3>
+<h3>Workload imbalance for geometric multigrid methods</h3>
+
+As mentioned above, the purpose of this program is to demonstrate the
+use of algebraic and geometric multigrid methods for this problem, and
+to do so for parallel computations. An important component of making
+algorithms scale to large parallel machines is ensuring that every
+processor has the same amount of work to do. (More precisely, what
+matters is that there are no small fraction of processors that have
+substantially more work than the rest since, if that were so, a large
+fraction of processors will sit idle waiting for the small fraction to
+finish. Conversely, a small fraction of processors having
+substantially <i>less</i> work is not a problem because the majority
+of processors continues to be productive and only the small fraction
+sits idle once finished with their work.)
+
 For the active mesh, we use the parallel::distributed::Triangulation class as done
 in step-40 which uses functionality in the external library
 <a href="http://www.p4est.org/">p4est</a> for the distribution of the active cells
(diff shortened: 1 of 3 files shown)


tjhei:
@tcclevenger can you please read through the changes and test that all methods still work?

tjhei:
@bangerth :
```
/home/travis/build/dealii/dealii/build/doc/doxygen/tutorial/step-50.h:362: warning: explicit link request to 'define' could not be resolved
```

examples/step-50/doc/intro.dox
```diff
@@ -19,11 +19,32 @@ libraries is described in the <a href="../../readme.html" target="body">README</
...
+require changes throughout a code base -- not very many, but one has
+to know what one is doing.
+
+What the results of this program will show
+is that algebraic and geometric multigrid methods are roughly
+comparable in performance <i>when using matrix-based formulations</i>,
+and that matrix-free geometric multigrid methods are vastly better for
+the problem under consideration here. A secondary conclusion will be
+that matrix-based geometric multigrid methods really don't scale well
+strongly when the number of unknowns per processor becomes smaler than
```
tjhei, line 45: Typo: smaller

examples/step-50/doc/intro.dox
```diff
@@ -56,9 +85,25 @@ center of the domain showing the adaptively refined mesh.
   </tr>
 </table>
 Both in 2D and 3D you can see the adaptive refinement picking up the corner singularity and the inner singularity where the viscosity jumps, while the interface along the line that separates the two viscosities is (correctly) not refined as it is resolved adequately.
+(This is because the kink in the solution that results from the jump
```
tjhei, line 88: Please remove parentheses.

examples/step-50/doc/results.dox
```diff
@@ -218,7 +247,21 @@ you can compare error rates.
...
+A more interesting example would involve a more complicated coarse mesh (see
+step-49 for inspiration). The issue in that case is that the coarsest
+level of the mesh hierarchy is actually quite large, and one would
+have to think about ways to solve the coarse level problem
+efficiently. (This is not an issue for algebraic multigrid methods
+because they would just continue to build coarser and coarser levels
+of the matrix, regardless of their geometric origin.)
+
+In the program here, we simply solve the coarse level problem with a
+Conjugate Gradient method without any precondition. That is acceptable
```
tjhei, line 259: preconditioner

examples/step-50/step-50.cc
```diff
@@ -52,8 +52,8 @@
 
 #include <deal.II/lac/generic_linear_algebra.h>
 
-// Comment the following \#define if you have PETSc and Trilinos installed and
-// you prefer using PETSc in this example:
+// Comment the following `\#define` in or out if you have PETSc and
```
tjhei: This way to escape seems to be incorrect.

bangerth:
On 5/13/20 4:43 PM, Timo Heister wrote:
> |/home/travis/build/dealii/dealii/build/doc/doxygen/tutorial/step-50.h:362: 
> warning: explicit link request to 'define' could not be resolved |

Should be fixed now!

examples/step-50/step-50.cc
```diff
@@ -1318,7 +1407,7 @@ void LaplaceProblem<dim, degree>::refine_grid()
 
   const double refinement_fraction = 1. / (std::pow(2.0, dim) - 1.);
   parallel::distributed::GridRefinement::refine_and_coarsen_fixed_number(
-    triangulation, estimate_vector, refinement_fraction, 0.0);
+    triangulation, error_estimator, refinement_fraction, 0.0);
```
tjhei, line 1412: I find the name "estimator" confusing for a vector. Do you mind some other name?

examples/step-50/step-50.cc
bangerth: Yep, already fixed.

examples/step-50/step-50.cc
bangerth: Hm, not sure here. We don't name the solution vector `solution_vector`, but just `solution`, and we use `system_rhs`. What I wanted to reflect is that we're estimating the *error* in the solution.

In step-6, we use `estimated_error_per_cell`. Would that be better?

bangerth:
All other comments @tjhei are addressed.

tjhei:
Thanks @bangerth for your dose of German verbosity. :-) Appreciated! 

@tcclevenger I let you go through it all as well.

examples/step-50/doc/intro.dox
```diff
@@ -67,27 +112,34 @@ implements what we will refer to as the "first-child rule" where, for each cell
...
-<img width="400px" src="https://www.dealii.org/images/steps/developer/step-50-workload-example.png" alt="">
+<img width="800px" src="https://www.dealii.org/images/steps/developer/step-50-workload-example.png" alt="">
 
 Included among the output to screen in this example is a value "Partition efficiency"
 given by one over MGTools::workload_imbalance(). This value, which will be denoted
 by $\mathbb{E}$,  quantifies the overhead produced by not having a perfect work balance
-on each level of the multigrid hierarchy (as is evident from the example above).
+on each level of the multigrid hierarchy. This imbalance is evident from the
+example above: while level $\ell=2$ is about as well balanced as possible
+among three processors as is possible with four cells, the coarse
```
tcclevenger, line 128:
```suggestion
example above: while level $\ell=2$ is about as well balanced as is possible
with four cells among three processors, the coarse
```

examples/step-50/doc/results.dox
```diff
@@ -70,15 +70,24 @@ allowing for vectorization over 8 doubles (vectorization used only in the matrix
...
-and 7168 processors. Here, $\mathbb{E}$ is the partition efficiency from the
+We can then gather a variety of information by calling the program
+with the input files that are provided in the directory in which
+step-50 is located. Using these, and adjusting the number of mesh
+refinement steps, we can produce information about how well the
+program scales.
+
+The following table gives weak scaling timings for this program on up to 256M DoFs
+and 7,168 processors. (Recall that weak scaling keeps the number of
+degrees per processor constant while increasing the number of
```
tcclevenger, line 81:
```suggestion
degrees of freedom per processor constant while increasing the number of
```

This change may be unnecessary.

examples/step-50/doc/results.dox
```diff
@@ -137,74 +146,94 @@ for GMG.
...
+On the other hand, the algebraic multigrid in the last set of columns
+is relatively unaffected by the increasing imbalance of the mesh
+hierarchy (because it doesn't use the mesh hierarchy) and the growth
+in time is rather driven by other factors that are well documented in
+the literature (most notably that the algorithmic complexity of
+some parts of algebraic multigrid methods appears to be ${\cal O}(N
+\log N)$ instead of ${\cal O}(N)$ for geometric multigrid).
+
+The upshort of the table above is that the matrix-free geometric multigrid
+method appear to be the fastest approach to solving this equation if
```
tcclevenger, line 218:
```suggestion
method appears to be the fastest approach to solving this equation if
```

examples/step-50/doc/results.dox
```diff
@@ -137,74 +146,94 @@ for GMG.
...
+we solve the same problem on more and more processors. Specifically,
+we consider the problems after 16 mesh refinement cycles
+(32M DoFs) and 19 cycles (256M DoFs), on between 56 to 28,672 processors:
 
-<img width="400px" src="https://www.dealii.org/images/steps/developer/step-50-strong-scaling.png" alt="">
+<img width="600px" src="https://www.dealii.org/images/steps/developer/step-50-strong-scaling.png" alt="">
+
+While the matrix-based GMG solver and AMG scale similarly and have a
+similar time to solution at least as long as there is a substantial
+number of unknowns per processor (say, several 10,000), the
```
tcclevenger, line 231:
```suggestion
similar time to solution (at least as long as there is a substantial
number of unknowns per processor, say, several 10,000), the
```

bangerth:
I made the suggestions. Thanks for the careful review, @tcclevenger !

Triptych: Yes, I was a bit proud of that. I did double check on wiktionary that the word meant what I thought: "A work of art in three pieces" :-)

examples/step-50/step-50.cc
```diff
@@ -233,10 +243,10 @@ number Coefficient<dim>::average_value(
 template <int dim>
 template <typename number>
 std::shared_ptr<Table<2, VectorizedArray<number>>>
-Coefficient<dim>::create_coefficient_table(
+Coefficient<dim>::make_coefficient_table(
   const MatrixFree<dim, number, VectorizedArray<number>> &mf_storage) const
 {
-  auto coefficient_table =
+  std::shared_ptr<Table<2, VectorizedArray<number>>> coefficient_table =
```
tjhei, line 249: I prefer the ``auto`` here, as the type is redundant in this expression.

examples/step-50/step-50.cc
```diff
@@ -411,15 +431,15 @@ class LaplaceProblem
...
+  IndexSet                  locally_owned_dofs;
+  IndexSet                  locally_relevant_dofs;
   AffineConstraints<double> constraints;
 
   MatrixType             system_matrix;
   MatrixFreeActiveMatrix mf_system_matrix;
   VectorType             solution;
   VectorType             right_hand_side;
-  Vector<double>         estimate_vector;
+  Vector<double>         error_estimator;
```
tjhei, line 442: I am not happy with the name "estimator" as this is a vector not a thing computing an estimate. How about ``error_estimate``?

examples/step-50/step-50.cc
```diff
@@ -798,8 +856,7 @@ void LaplaceProblem<dim, degree>::assemble_multigrid()
             for (unsigned int j = 0; j < dofs_per_cell; ++j)
               cell_matrix(i, j) +=
                 (coefficient_value * fe_values.shape_grad(i, q_point) *
-                 fe_values.shape_grad(j, q_point)) *
-                fe_values.JxW(q_point);
+                 fe_values.shape_grad(j, q_point) * fe_values.JxW(q_point));
```
tjhei, line 859: then we should maybe remove the outer ()?

examples/step-50/step-50.cc
tjhei: ``error_estimate`` or ``estimated_error_per_cell`` sounds good

examples/step-50/step-50.cc
bangerth: But only if you know what `std::make_shared` returns. But who cares, I put the `auto` back in.

examples/step-50/doc/intro.dox
```diff
@@ -67,27 +112,34 @@ implements what we will refer to as the "first-child rule" where, for each cell
...
-<img width="400px" src="https://www.dealii.org/images/steps/developer/step-50-workload-example.png" alt="">
+<img width="800px" src="https://www.dealii.org/images/steps/developer/step-50-workload-example.png" alt="">
 
 Included among the output to screen in this example is a value "Partition efficiency"
 given by one over MGTools::workload_imbalance(). This value, which will be denoted
 by $\mathbb{E}$,  quantifies the overhead produced by not having a perfect work balance
-on each level of the multigrid hierarchy (as is evident from the example above).
+on each level of the multigrid hierarchy. This imbalance is evident from the
+example above: while level $\ell=2$ is about as well balanced as is possible
+with four cells among three processors, the coarse
```
kronbichler, line 128: Should we maybe add that this balance is *by chance*?

examples/step-50/doc/intro.dox
```diff
@@ -134,3 +191,64 @@ due to the asynchronous work "covering up" the imbalance (which assumes synchron
 However, for most realistic adaptive meshes the expectation is that this asynchronous work will
 only cover up a very small portion of the imbalance and the efficiency model will describe the
 slowdown very well.
+
+
+<h3>Workload imbalance for algebraic multigrid methods</h3>
+
+The considerations above show that one has to expect certain limits on
+the scalability of geometric multigrid methods because even in cases
```
kronbichler, line 199:
```suggestion
the scalability of the geometric multigrid algorithm as it is implemented in deal.II because even in cases
```
The imperfection is not a given and could be fixed by a more clever strategy for distributing the level cells.

examples/step-50/doc/intro.dox
```diff
@@ -134,3 +191,64 @@ due to the asynchronous work "covering up" the imbalance (which assumes synchron
...
+and PETScWrappers::PreconditionBoomerAMG classes. In some sense, these
+algorithms are simpler than for geometric multigrid methods because
+they only deal with the matrix itself, rather than all of the
+connotations of meshes, neighbors, parents, and other geometric
+entities. At the same time, much work has also been put into making
+algebraic multigrid methods scale to very large problems, including
+questions such as reducing the number of processors that work on a
+given level of the hierarchy to a subset of all processors, if
+otherwise processors would spend less time on computations than on
+communication.
```
kronbichler, line 223: This kind of explains the point of the previous comment - all of this can also be done for the geometric multigrid method, it is just a matter of finding a volunteer. :wink:

examples/step-50/doc/intro.dox
```diff
@@ -134,3 +191,64 @@ due to the asynchronous work "covering up" the imbalance (which assumes synchron
...
+file (here, `gmg_mb_2d.prm`).
+
+The program is intended to be run in parallel, and you can achieve
+this using a command such as
+@code
+  mpirun -np 4 ./step-50 gmg_mb_2d.prm
+@endcode
+if you want to, for example, run on four processors. (That said, the
+program is also ready to run with, say, `-np 28672` if that's how many
+processors you have available.)
```
kronbichler, line 254: I don't think people typically state `-n 28672` explicitly and instead inherit it from the variables exported by the jobfile, for example to the MPI environment itself. I most often just run `mpirun ./step-50 gmg_mf_2d.prm`. But I agree that the number is impressive, so I would not change anything here.

examples/step-50/doc/results.dox
```diff
@@ -218,7 +247,21 @@ you can compare error rates.
...
+level of the mesh hierarchy is actually quite large, and one would
+have to think about ways to solve the coarse level problem
+efficiently. (This is not an issue for algebraic multigrid methods
+because they would just continue to build coarser and coarser levels
+of the matrix, regardless of their geometric origin.)
+
+In the program here, we simply solve the coarse level problem with a
+Conjugate Gradient method without any preconditioner. That is acceptable
+if the coarse problem is really small -- here it is a $9\times 9$
+matrix in 2d, and a $27\times 27$ matrix in 3d. But if the coarse mesh
```
kronbichler, line 261: These are the numbers for a single cell, right? Shouldn't it be 117 in 2d and something around a few hundreds in 3d since we have an L-shaped domain? @tcclevenger can you please check?

tcclevenger: Correct, it's 3 cells in 2d and 7 cells in 3d. The number of DoFs are 21 in 2d and 117 in 3d.

examples/step-50/doc/intro.dox
bangerth: Well, actually no. The finest level is *always* about as well balanced as one can.

examples/step-50/doc/intro.dox
bangerth: Yep. I've added such a comment to this paragraph.

examples/step-50/doc/intro.dox
kronbichler: The *active* level is as balanced as it can be. But that does not imply that the cells on the finest level are balanced. Imagine what would happen if the first of the four level-1 cells in the figure is refined rather than the third. In that case, we have 3 cells on proc 0, 1 cell on proc 1 and none on proc 2. Or did I misread what you wanted to say?

examples/step-50/doc/results.dox
bangerth: Ah, fair point. I'll update.

bangerth:
All suggested changes made.

examples/step-50/doc/intro.dox
bangerth: Oh, I see what you mean. 

I think to explain it at this level of detail would require a paragraph. It may not be necessary in the current context to make the point we're trying to make here. But I will if you insist!

kronbichler: No, I don't insist, let's keep it this way, it is very good and it gets clear from the context if one really wants to know it.

PR #10202, opened 2020-05-13
Update an entry in the changelog.

bangerth:
Needs to go into 9.2.


tamiko:
@bangerth Simply mention #10073 then (or add the PR number to the list in the top comment of #10073) :smiley:

PR #10203, opened 2020-05-13
[9.2.] Take over #10201

tamiko:
#10201 CMake: relax arpack and parpack library directory requirements
See #10072


tamiko:
@bangerth I am very sure that this actually works. What makes you concerned that this is wrong? :smile: 

I tested this on my local machine (usual `/usr/lib64/lib{p,}arpack.so*` setup), and then switched to a setup where `/usr/lib64/libparpack.so*` is located in a separate directory. Both are getting picked up correctly.

The addition in question was introduced by #1146 in 2015, and I bet it was tested with a local spack installation (that doesn't install arpack/parpack in different directories). The constraint in question was probably introduced to work around some local peculiarities. But for starters, we cannot always safeguard against misconfiguration on user side and, more importantly, there is a legitimate reason to allow a more relaxed layout (as it stands you cannot configure parpack on centos/redhat without some workarounds).

bangerth:
That's all I needed to know. I've just become wary of these kinds of configuration changes late in the game -- someone always has a configuration out there that now doesn't work any more, and we won't have time to find out about it before the release.

But if you're certain, then that's ok with me!

PR #10207, opened 2020-05-14
Fix extract_rtree_level for small problem sizes

peterrum:
This PR fixes the function `extract_rtree_level`. For very small problem sizes (approx. less than 12 entries in the rtree), this algorithms did not return any bounding box.


luca-heltai:
Thanks!

PR #10215, opened 2020-05-14
Bugfix: Cells may not receive any variable size data.

### LINKED ISSUE #10211: Bug in ParticleHandler.
It looks like we inherited this bug: 

https://github.com/geodynamics/aspect/issues/3354

marcfehling:
Closes #10211.

Cells will always receive fixed size data (i.e., CellStatus information), but not necessarily variable size data (e.g., with a ParticleHandler a cell might not contain any particle at all). Thus it is sufficient to check if fixed size data has been received.

~~Do not merge yet -- I'd like to provide a failing test first.~~

### DIFF
--- a/source/distributed/tria.cc
+++ b/source/distributed/tria.cc
@@ -1611,16 +1611,16 @@ namespace parallel
       const bool         callback_variable_transfer = (handle % 2 == 0);
       const unsigned int callback_index             = handle / 2;
 
+      // Cells will always receive fixed size data (i.e., CellStatus
+      // information), but not necessarily variable size data (e.g., with a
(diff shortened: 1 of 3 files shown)


luca-heltai:
You'll need to reindent.

marcfehling:
The provided test fails without the patch for `tria.cc`.

This is a very basic test, though, but I think that there shouldn't be any problems in more complex scenarios. You might give this a try for `step-70`! :-)

marcfehling:
Ready for review/tests/merge.

tamiko:
I took the opportunity to run the CI on the release branch in #10216

PR #10218, opened 2020-05-15
[9.2.] Take over #10174

tamiko:
#10174: Fix find active cell around point
See also #10072

This is now a non-failing test with a fix.


tamiko:
I suggest we merge this workaround together with step 70 and tag rc2. After that I would declare the release candidate feature complete. We then have a couple of days to come up with a proper fix.

PR #10225, opened 2020-05-15
Fix test output forgotten in #9969

kronbichler:
This fixes the test error recorded at
https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=6397
due to the adjustment of the parallel grid refinement algorithm.
The CI won't run over this, otherwise we would not have merged #9969, so it can be merged quickly.


marcfehling:
Thank you! I was aware that these outputs had to be altered, but failed to update them before merge. Sorry for the inconvenience!

PR #10227, opened 2020-05-15
Fix warning in step-70

kronbichler:
This should fix the warning displayed on clang compilations:
https://cdash.43-1.org/viewBuildError.php?type=1&buildid=6404
Candidate for the 9.2 release branch.


kronbichler:
There is another set of warnings in line 1402,
https://cdash.43-1.org/viewBuildError.php?type=1&buildid=6412
I don't get to immediately fix it now but that should also be part of fixing the compile warning fixes for step-70.

PR #10229, opened 2020-05-15
[9.2] Take over #10227

tamiko:
#10227: Fix warning in step-70
See also #10072


tjhei:
this ran through on master, so we should be good to go.

PR #10232, opened 2020-05-15
Updated doi.

luca-heltai:
Resered a doi on zenodo.


bangerth:
We also need this on 9.2.

PR #10235, opened 2020-05-15
Prevent instantiation of BoundingBox<0>.

### LINKED ISSUE #10221: boundingbox_6 and dim==0 errors
as discussed on slack with @luca-heltai and @drwells a few days ago but I don't think the fix has landed yet. @drwells did you want to look at it?

I think we need to:
1. static_ssert dim>=1 in BoundingBox.
2. Adapt the test (and anything else that breaks afterwards)

Also see:
https://cdash.43-1.org/testDetails.php?test=44333311&build=6387

drwells:
A zero-dimensional bounding box doesn't make sense and causes weird problems
when the compiler tries to instantiate BoundingBox::cross_section().

Fixes #10221.

### DIFF
--- a/include/deal.II/base/bounding_box.h
+++ b/include/deal.II/base/bounding_box.h
@@ -253,6 +253,9 @@ public:
   /**
    * Returns the cross section of the box orthogonal to @p direction.
    * This is a box in one dimension lower.
+   *
+   * @note Calling this method in 1D will result in an exception since
+   * <code>BoundingBox&lt;0&gt;</code> is not implemented.
    */
   BoundingBox<spacedim - 1, Number>
   cross_section(const unsigned int direction) const;
@@ -266,6 +269,32 @@ public:
 
 private:
   std::pair<Point<spacedim, Number>, Point<spacedim, Number>> boundary_points;
+};
+
+/**
+ * Specialization of BoundingBox for spacedim 0. This class exists to enable
+ * dimension-independent programming but unconditionally throws an exception
+ * in its constructor.
+ */
+template <typename Number>
+class BoundingBox<0, Number>
+{
+public:
+  /**
+   * Default constructor. Throws an exception.
+   */
+  BoundingBox();
+
+  /**
+   * Equivalent two-point constructor. Throws an exception.
+   */
+  BoundingBox(const std::pair<Point<0, Number>, Point<0, Number>> &);
+
+  /**
+   * Equivalent container constructor. Throws an exception.
+   */
+  template <class Container>
+  BoundingBox(const Container &);
 };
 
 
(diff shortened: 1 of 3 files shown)


drwells:
I think I have a workaround - its interesting that intel and MSVC both dislike this function for very different reasons :)

tjhei:
candidate for 9.2?

drwells:
I should be clearer - this isn't going to work with MSVC. I am working on a workaround for the workaround now :)

drwells:
That fix is a little uglier, but simpler - MSVC should accept it.

drwells:
@tjhei I completely rewrote this pr after you added 'reviewed and ready to merge' - are you still happy with it?

tjhei:
can you fix indentation? Yes, still looks good!

include/deal.II/base/bounding_box.h
```diff
@@ -250,11 +252,22 @@ class BoundingBox
...
+  /**
+   * The dimensionality of the box returned by cross_section is constrained to
+   * always be at least one to prevent the instantiation of
+   * <code>BoundingBox&lt;0&gt;</code>.
+   *
+   * @note This has to be explicitly calculated outside the method declaration
+   * in a way that works across multiple compilers (e.g., this check cannot be
+   * put directly in the signature of cross_section when using MSVC).
+   */
+  static constexpr int cross_section_dim = spacedim != 1 ? spacedim - 1 : 1;
```
bangerth, line 264: Totally not important (and maybe not necessary to do before the release), but I prefer things on the right hand side of `=` to be parenthesized if they contain `?:` or comparison operators. I think it's hard to break down for the eye what the left and right hand side of the central operator is without parentheses. (Central here: `=`.)

bangerth: Separately: You explain in the note why this has to be done outside the declaration in question, but not why we need the `?:` to begin with (or, more precisely, why we need the `else` case).

drwells: This is so easy to fix lets just do it now. Fixed.

drwells: My web browser didn't load your second comment for some reason - let me do that too.

drwells: > You explain in the note why this has to be done outside the declaration in question, but not why we need the ?: to begin with (or, more precisely, why we need the else case

I think this is addressed by the first sentence, where I stated that we need to prevent instantiation of `BoundingBox<0>`. Is that not sufficient?

bangerth: Oh, I didn't read closely enough. Yes.

drwells:
MSVC doesn't like the current version of this patch either:
```

2020-05-15T20:01:12.1975125Z D:\a\dealii\dealii\source\base\bounding_box.cc(296): error C2244: 'dealii::BoundingBox<spacedim,Number>::cross_section': unable to match function definition to an existing declaration [D:\a\dealii\dealii\build\source\base\obj_base_debug.vcxproj]
2020-05-15T20:01:12.5417594Z   D:\a\dealii\dealii\source\base\bounding_box.cc(295): note: see declaration of 'dealii::BoundingBox<spacedim,Number>::cross_section'
2020-05-15T20:01:12.9937405Z   D:\a\dealii\dealii\source\base\bounding_box.cc(296): note: definition
2020-05-15T20:01:13.4177337Z   D:\a\dealii\dealii\source\base\bounding_box.cc(296): note: 'dealii::BoundingBox<dealii::BoundingBox<spacedim,Number>::cross_section_dim,Number> dealii::BoundingBox<spacedim,Number>::cross_section(const unsigned int) const'
2020-05-15T20:01:13.8398229Z   D:\a\dealii\dealii\source\base\bounding_box.cc(296): note: existing declarations
2020-05-15T20:01:14.1202125Z   D:\a\dealii\dealii\source\base\bounding_box.cc(296): note: 'dealii::BoundingBox<dealii::BoundingBox<spacedim,Number>::cross_section_dim,Number> dealii::BoundingBox<spacedim,Number>::cross_section(const unsigned int) const'
```
Let me try something else.

drwells:
This is the third rewrite of the PR so someone should approve again before we merge.

masterleinad:
Let's just merge.

PR #10236, opened 2020-05-15
Use std_cxx14::make_unique in step-70.

drwells:


drwells:
We should put this in 9.2.

bangerth:
This needs to go into 9.2 as well.

PR #10237, opened 2020-05-15
Improve spelling in step-70.

drwells:
I was in the neighborhood :)


examples/step-70/doc/intro.dox
```diff
@@ -322,7 +322,7 @@ implementation, thanks to the dimension independent capabilities of deal.II.
 
 In this tutorial, both the embedded grid $\Gamma$ and the embedding
 grid are described using a parallel::distributed::Triangulation. These two
-triangulations can be built from deal.II GridGenerators namespace or by reading
+triangulations can be built from deal.II GridGenerator namespace or by reading
```
jppelteret, line 325: from the deal.II... ?

drwells: good catch. I wrote "from the GridGenerator namespace" since we don't really need to specify that this feature is part of the library.

bangerth:
Documentation only, so will merge.

Issue #10240, opened 2020-05-15
Issues with copyright script.


marcfehling:
I just stumbled over a line that the copyright script misinterpreted in #10097 and found a few other candidates:
```
[marc@ias7154 dealii]$ grep -r . -e "Copyright.*-2020"
./source/differentiation/sd/symengine_utilities.cc:// Copyright (C) 2019-2020 by the deal.II authors
./include/deal.II/differentiation/sd.h:// Copyright (C) 2019 - 2020-2020 by the deal.II authors
./include/deal.II/differentiation/sd/symengine_utilities.h:// Copyright (C) 2019-2020 by the deal.II authors
./cmake/configure/configure_symengine.cmake:## Copyright (C) 2019 - 2020-2020 by the deal.II authors
```

However I also noted that nested directories in the `include` and `source` folder over more than two levels are entirely ignored by the script:

https://github.com/dealii/dealii/blob/b2ab70f362646eefc964d308e2f42de3c6bf5cf6/contrib/utilities/update-copyright#L28-L38

---

EDIT: There's more! Adjusted my search credentials...
```
[marc@ias7154 dealii]$ grep -r . -e "Copyright.*-20.*deal.II"
./source/differentiation/CMakeLists.txt:## Copyright (C) 2017 - 2019-2019 by the deal.II authors
./include/deal.II/lac/read_write_vector.templates.h:// Copyright (C) 2015 - 2020-2016 by the deal.II authors
./tests/matrix_free/thread_correctness_hp.cc:// Copyright (C) 2013 - 2020-2014 by the deal.II authors
./tests/matrix_free/get_functions_faces.cc:// Copyright (C) 2014 - 2020-2014 by the deal.II authors
./tests/lapack/full_matrix_26.cc:// Copyright (C) 2014 - 2018-2015 by the deal.II authors
./tests/lapack/full_matrix_00.cc:// Copyright (C) 2014 - 2018-2015 by the deal.II authors
./tests/grid/build_global_description_tree.cc:// Copyright (C) 2018 - 2020-2019 by the deal.II authors
./tests/grid/distributed_compute_point_locations_02.cc:// Copyright (C) 2017 - 2020-2018 by the deal.II authors
./tests/grid/distributed_compute_point_locations_01.cc:// Copyright (C) 2017 - 2020-2018 by the deal.II authors
./tests/dofs/get_dofs_on_patch.cc:// Copyright (C) 2014 - 2018-2015 by the deal.II authors
[...]

PR #10242, opened 2020-05-15
Fixed a few copyright years manually.

marcfehling:
Part of #10240.


marcfehling:
Wait! There's more!

I'll adjust a few more lines and let you know once I think I'm done.

marcfehling:
I think I got all the anomalies.

I have updated the copyright years to `2020` because we are changing these files right now.

tjhei:
> I have updated the copyright years to `2020` because we are changing these files right now.

The update script only bumps the year if a change that was not a copyright change is detected. I think if I run the script, it should bump the year.
Ideally, your commit should have the title "update copyright" by the way (for the script to ignore your commit).

PR #10247, opened 2020-05-16
[9.2] Take over #10235: Specialize BoundingBox<0> to prevent instantiating BoundingBox<-1>.

masterleinad:
A zero-dimensional bounding box doesn't make sense and causes weird problems
with intel's compiler when the compiler tries to instantiate
BoundingBox::cross_section().

We have to use this workaround and not something like

    BoundingBox<spacedim != 1 ? spacedim - 1 : 1, Number>
    cross_section(const unsigned int direction) const;

to make MSVC happy.

### DIFF
--- a/include/deal.II/base/bounding_box.h
+++ b/include/deal.II/base/bounding_box.h
@@ -253,6 +253,9 @@ public:
   /**
    * Returns the cross section of the box orthogonal to @p direction.
    * This is a box in one dimension lower.
+   *
+   * @note Calling this method in 1D will result in an exception since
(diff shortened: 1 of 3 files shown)


bangerth:
Apparently approving the same PR 3 times in a row doesn't trigger the limit ;-) Someone else will have to merge then :-)

drwells:
I can approve my own PR :grin:

drwells:
This already passed the CI checks so I'm going to merge without running the rest of the test suite.

Issue #10246, opened 2020-05-16
complex number product type deduction is wrong


tamiko:
I have some code that due to some design decisions does a rather interesting tensor product:

```
#include <deal.II/base/tensor.h>
int main()
{
  typedef std::complex<double> rank0_type;
  typedef dealii::Tensor<1, 1, rank0_type> curl_type;
  typedef dealii::Tensor<0, 2, rank0_type> permeability_type;
  permeability_type mu_inv;
  curl_type u;
  rank0_type result = u * mu_inv * u;
  return 0;
}
```

This code fails now (and compiled in December 2019):
```
/home/tamiko/workspace/step/step.cc:13:14: error: no viable conversion from 'complex<std::complex<double>>' to 'complex<double>'
  rank0_type result = u * mu_inv * u;
             ^        ~~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/include/g++-v9/complex:1221:12: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'typename Tensor<1 + 1 - 2, 1, typename ProductType<complex<complex<double> >, complex<double> >::type>::tensor_type' (aka 'std::complex<std::complex<double> >') to 'const std::complex<double> &' for 1st argument
    struct complex<double>
           ^
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/include/g++-v9/complex:1226:26: note: candidate constructor not viable: no known conversion from 'typename Tensor<1 + 1 - 2, 1, typename ProductType<complex<complex<double> >, complex<double> >::type>::tensor_type' (aka 'std::complex<std::complex<double> >') to 'std::complex<double>::_ComplexT' (aka '_Complex double') for 1st argument
      _GLIBCXX_CONSTEXPR complex(_ComplexT __z) : _M_value(__z) { }
                         ^
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/include/g++-v9/complex:1228:26: note: candidate constructor not viable: no known conversion from 'typename Tensor<1 + 1 - 2, 1, typename ProductType<complex<complex<double> >, complex<double> >::type>::tensor_type' (aka 'std::complex<std::complex<double> >') to 'double' for 1st argument
      _GLIBCXX_CONSTEXPR complex(double __r = 0.0, double __i = 0.0)
                         ^
[...]

PR #10249, opened 2020-05-16
Workaround: Add additional complex ProductTypeImpl specializations

tamiko:
The workaround introduced in 32ecab6 needs an additional variant to
deduce the product type of Tensor<..., complex<Number>> and
complex<Number> correctly.

Closes: #10246
See also: #10072


tamiko:
This doesn't have the `#if` in place.

drwells:
I think this is fine as long as we open a TODO or issue to remind us to go back and delete this when we upgrade our minimum compiler requirements.

drwells:
Could we also convert the example code you had in the other issue into a test?

tamiko:
@drwells Yes, I will add a test for unusual tensor products in a minute.

tamiko:
@drwells Done. This is a bit more than just the example. I figured we could just add a `tensors/tensor_31` test, aka don't make Matthias unhappy ^^

PR #10253, opened 2020-05-16
Fix doxygen underscores.

drwells:
We need to use `$\mathrm{A\_B}$` and not `$\texttt{A\_B}$` to correctly print an underscore and not the backslash.

### DIFF
--- a/include/deal.II/physics/elasticity/standard_tensors.h
+++ b/include/deal.II/physics/elasticity/standard_tensors.h
@@ -153,12 +153,12 @@ namespace Physics
        *  \dealcoloneq \{ \bullet \} - \frac{1}{\textrm{dim}}
        *  \left[ \{ \bullet \} : \mathbf{I} \right]\mathbf{I}
        *  = \mathcal{P}^{T} : \{ \bullet \}
-       *  = \texttt{dev\_P} \left( \{ \bullet \} \right)
+       *  = \mathtt{dev\_P} \left( \{ \bullet \} \right)
        * @f]
        * and, therefore,
        * @f[
-       * \texttt{dev\_P} \left( \{ \bullet \} \right) : \mathbf{I}
-       *   = \textrm{trace}(\texttt{dev\_P} \left( \{ \bullet \} \right)) = 0 \,
+       * \mathtt{dev\_P} \left( \{ \bullet \} \right) : \mathbf{I}
+       *   = \mathrm{trace}(\mathtt{dev\_P} \left( \{ \bullet \} \right)) = 0 \,
        * .
        * @f]
        *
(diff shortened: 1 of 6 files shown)


tjhei:
Can't you leave things as is and only replace ``\_`` by ``_`` or do I misunderstand?

drwells:
I think that fails in plain tex. Let me try and see what happens.

drwells:
`\text{n_shells}` is fine with MathJax but doesn't work with latex so I believe this patch is the right approach.

tjhei:
> `\text{n_shells}` is fine with MathJax but doesn't work with latex so I believe this patch is the right approach.

@tamiko didn't you submit a PR the other day that used this style?

tamiko:
>> `\text{n_shells}` is fine with MathJax but doesn't work with latex so I believe this patch is the right approach.
>
> @tamiko didn't you submit a PR the other day that used this style?

Yes, for step-69. (used `\texttt{foo_bar}` in three places).

Best,
Matthias

drwells:
This is weird - I cannot render the step-67 and step-69 MathJax equations. Let me try to debug this to see if its related to this PR.

drwells:
I get unrelated doxygen compilation problems that I suspect are the result of me using a very new version of MathJax (3.0.5). I'm putting together some fixes for step-69 but since this problem only seems to effect me its not that important - I recommend we proceed with this PR since it is simple.

include/deal.II/physics/elasticity/standard_tensors.h
```diff
@@ -201,12 +201,12 @@ namespace Physics
        *  \{ \bullet \} : \hat{\mathcal{P}}
        *    \dealcoloneq J^{-2/\textrm{dim}} \left[ \{ \bullet \} -
        * \frac{1}{\textrm{dim}}\left[\mathbf{C} : \{ \bullet \}\right]
-       * \mathbf{C}^{-1} \right] = \texttt{Dev\_P} \left( \{ \bullet \} \right)
+       * \mathbf{C}^{-1} \right] = \mathrm{Dev\_P} \left( \{ \bullet \} \right)
```
tjhei, line 204: should these be ``\mathtt`` ?

drwells: I wrote this PR before I learned about `\mathtt`

tjhei: would you mind changing them?

tamiko:
@drwells @tjhei Let's get this trivial PR merged at some point :-)

drwells:
Sorry! I'll get to this at some point today.

include/deal.II/physics/elasticity/standard_tensors.h
drwells: Fixed!

drwells:
Done!

drwells:
:eyes:

PR #10257, opened 2020-05-18
sacado/helper_scalar_coupled_3_components_01_4: avoid race condition

tamiko:


tamiko:
@drwells If you create a plain `.output` and an mpi variant `.mpirun=X.output` then unfortunately two targets for the same executable are created. If they run just at the right time one of the tests will either have no executable or an empty object file because it gets overwritten by the other one.

The mpi variants take care of that by only creating one rule for the executable and being prohibited from being scheduled at the same time.

tamiko:
Maybe I should just give up and support that use case...

drwells:
Ah. I think I understand.

Would you please make the equivalent fix to `helper_scalar_coupled_3_components_01_3.output`?

Issue #10243, opened 2020-05-15
step-70: Missing viz file


bangerth:
The results section references https://www.dealii.org/images/steps/developer/step-70.2d_tracing.gif, but this file does not exist.

bangerth:
That's also true for the images further down in the results section -- see https://dealii.org/developer/doxygen/deal.II/step_70.html#Results

bangerth:
We need to address this before the release. @luca-heltai @blaisb -- what's your plan?

luca-heltai:
I have no longer access to the svn server. These are all available in the gist that @blaisb published at the end of the step-70 PR, and someone with svn access should upload them on the webserver.

blaisb:
> We need to address this before the release. @luca-heltai @blaisb -- what's your plan?

All files are under the right name in the gist :  https://gist.github.com/blaisb/a211c601067bb853a23c5beae3af0d9c
I do not have access to the svn to upload the files. Ping me if there is anything you need.

bangerth:
@blaisb The results section also references a 3d gif file that's not in the gist. Do you happen to still have the vtk files for that? I could do the same as for the 2d video.

tamiko:
All files referenced by step-70 are now in place.

PR #10259, opened 2020-05-18
Edits for step-70.cc

bangerth:
I've got to do a couple of other things for the next hour, but thought I'd already make this available to review. There is nothing here that changes functionality. It's really just edits, plus re-sorting the member variables of the class so that things that work together follow each other.

@luca-heltai @blaisb @gassmoeller: It's been my experience that it reads to rather difficult to read tutorial programs if one slices the code too thinly: Documenting individual lines and member variables makes it difficult to follow the flow of the program because one loses track of the bigger picture. So I've often combined blocks of documentation and code into larger chunks to make it easier on the eye. But there are no functional changes to all of this.

More to come later today.

/rebuild

### DIFF
--- a/examples/step-70/step-70.cc
+++ b/examples/step-70/step-70.cc
@@ -386,46 +386,51 @@ namespace Step70
     // passive tracers.
     enter_my_subsection(this->prm);
     this->prm.enter_subsection("Grid generation");
-    this->prm.add_parameter("Fluid grid generator", name_of_fluid_grid);
-    this->prm.add_parameter("Fluid grid generator arguments",
-                            arguments_for_fluid_grid);
+    {
+      this->prm.add_parameter("Fluid grid generator", name_of_fluid_grid);
+      this->prm.add_parameter("Fluid grid generator arguments",
+                              arguments_for_fluid_grid);
 
-    this->prm.add_parameter("Solid grid generator", name_of_solid_grid);
-    this->prm.add_parameter("Solid grid generator arguments",
-                            arguments_for_solid_grid);
+      this->prm.add_parameter("Solid grid generator", name_of_solid_grid);
+      this->prm.add_parameter("Solid grid generator arguments",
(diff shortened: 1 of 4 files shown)


examples/step-70/step-70.cc
```diff
@@ -532,68 +567,86 @@ namespace Step70
...
+    // particles that ended up in my subdomain.
     //
     // While this is not the most efficient strategy, we keep it this way to
-    // illustrate how things would work in a real FSI problem. If a particle
-    // is linked to a specific solid degree of freedom, we are not free to
-    // choose who owns it, and we have to communicate this information around.
-    // We illustrate this here, and show that the communication pattern is
-    // point-to-point, and negligible in terms of total cost of the algorithm.
+    // illustrate how things would work in a real fluid-structure
+    // intraction(FSI_ problem. If a particle is linked to a specific solid
```
luca-heltai, line 599:
```suggestion
    // intraction (FSI) problem. If a particle is linked to a specific solid
```

luca-heltai: Only this minor typo. For the rest: thanks  @bangerth for taking the time to do this!

:)

kronbichler: Wait, shouldn't this be "interaction" rather than "intraction"?

luca-heltai: Yes!

bangerth: Yes, I had to laugh as well :-)

bangerth:
I pushed that one line of change. Of course, this destroys the status of all 11 checks succeeding. If someone wants to be bold, wait for the mandatory checks and merge before it all runs through again :-)

bangerth:
I've gone through the rest of the code. I kept actual code changes separate from documentation changes in the commits to make it easier to understand what I'm changing.

I think this is in great shape. We just have to address #10243 (in a separate PR).

PR #10268, opened 2020-05-19
Simplify get_dof_indices() and set_dof_indices()

peterrum:
This PR makes the functions `get_dof_indices()` and `get_dof_indices()` similar. 

The functions differ now at a handful places as shown by the following diff:
```diff
@@ -1,10 +1,10 @@
       template <typename DoFHandlerType, bool level_dof_access, int structdim>
       static void
-      get_dof_indices(
+      set_dof_indices(
         const dealii::DoFAccessor<structdim, DoFHandlerType, level_dof_access>
-          &                                   accessor,
-        std::vector<types::global_dof_index> &dof_indices,
-        const unsigned int                    fe_index)
+          &                                         accessor,
+        const std::vector<types::global_dof_index> &dof_indices,
+        const unsigned int                          fe_index)
       {
         const unsigned int                                             //
           dofs_per_vertex = accessor.get_fe(fe_index).dofs_per_vertex, //
@@ -22,9 +22,10 @@
         for (const unsigned int vertex :
              GeometryInfo<structdim>::vertex_indices())
           for (unsigned int d = 0; d < dofs_per_vertex; ++d, ++index)
-            dof_indices[index] = accessor.vertex_dof_index(vertex, 
-                                                           d, 
-                                                           fe_index);
+            accessor.set_vertex_dof_index(vertex,
+                                          d,
+                                          dof_indices[index],
+                                          fe_index);
 
         // 2) LINE dofs (2D/3D - see notes in get_dof_indices)
         if (structdim == 2 || structdim == 3)
@@ -32,13 +33,14 @@
                line < GeometryInfo<structdim>::lines_per_cell;
                ++line)
             for (unsigned int d = 0; d < dofs_per_line; ++d, ++index)
-              dof_indices[index] = accessor.line(line)->dof_index(
+              accessor.line(line)->set_dof_index(
                 structdim == 2 ? accessor.get_fe(fe_index)
                                    .adjust_line_dof_index_for_line_orientation(
                                      d, accessor.line_orientation(line)) :
                                  accessor.get_fe(fe_index)
                                    .adjust_line_dof_index_for_line_orientation(
                                      d, accessor.line_orientation(line)),
+                dof_indices[index],
                 fe_index);
 
         // 3) FACE dofs (3D - see notes in get_dof_indices)
@@ -47,18 +49,19 @@
                quad < GeometryInfo<structdim>::quads_per_cell;
                ++quad)
             for (unsigned int d = 0; d < dofs_per_quad; ++d, ++index)
-              dof_indices[index] = accessor.quad(quad)->dof_index(
+              accessor.quad(quad)->set_dof_index(
                 accessor.get_fe(fe_index)
                   .adjust_quad_dof_index_for_face_orientation(
                     d,
                     accessor.face_orientation(quad),
                     accessor.face_flip(quad),
                     accessor.face_rotation(quad)),
+                dof_indices[index],
                 fe_index);
 
         // 4) INNER dofs
         for (unsigned int d = 0; d < inner_dofs; ++d, ++index)
-          dof_indices[index] = accessor.dof_index(d, fe_index);
+          accessor.set_dof_index(d, dof_indices[index], fe_index);
 
         AssertDimension(dof_indices.size(), index);
       }
```

### DIFF
--- a/include/deal.II/dofs/dof_accessor.h
+++ b/include/deal.II/dofs/dof_accessor.h
@@ -1289,6 +1289,16 @@ public:
   DoFInvalidAccessor(const OtherAccessor &);
 
   /**
+   * Return the index of the <i>i</i>th degree of freedom of this object to
+   * @p index. Since the current object doesn't point to anything useful, like
+   * all other functions in this class this function only throws an exception.
+   */
+  types::global_dof_index
+  dof_index(const unsigned int i,
+            const unsigned int fe_index =
+              DoFHandler<dim, spacedim>::default_fe_index) const;
+
+  /**
    * Set the index of the <i>i</i>th degree of freedom of this object to @p
    * index. Since the current object doesn't point to anything useful, like
    * all other functions in this class this function only throws an exception.
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -1588,6 +1588,252 @@ namespace internal
           *local_values_begin = read_write_vector[sorted_indices_pos[i]];
       }
 #endif
+
+
+
+      /**
+       * Loop over all degrees of freedom of the object described by the
+       * provided @p accessor and @p fe_index and perform the static functions
+       * provided by DoFOperation (set/get) on these.
+       */
+      template <typename DoFHandlerType,
+                bool level_dof_access,
+                int  structdim,
+                typename DoFIndicesType,
+                typename DoFOperation>
+      static void
+      process_dof_indices(
+        const dealii::DoFAccessor<structdim, DoFHandlerType, level_dof_access>
+          &                accessor,
+        DoFIndicesType &   dof_indices,
+        const unsigned int fe_index,
+        const DoFOperation &)
+      {
+        const unsigned int                                             //
+          dofs_per_vertex = accessor.get_fe(fe_index).dofs_per_vertex, //
+          dofs_per_line   = accessor.get_fe(fe_index).dofs_per_line,   //
+          dofs_per_quad   = accessor.get_fe(fe_index).dofs_per_quad,   //
+          dofs_per_hex    = accessor.get_fe(fe_index).dofs_per_hex;    //
+
+        const unsigned int inner_dofs =
+          structdim == 1 ? dofs_per_line :
+                           (structdim == 2 ? dofs_per_quad : dofs_per_hex);
+
+        unsigned int index = 0;
+
+        // 1) VERTEX dofs
+        for (const unsigned int vertex :
+             GeometryInfo<structdim>::vertex_indices())
+          for (unsigned int d = 0; d < dofs_per_vertex; ++d, ++index)
+            DoFOperation::process_vertex_dof(
+              accessor, vertex, d, dof_indices[index], fe_index);
+
+        // 2) copy dof numbers from the LINE. for lines with the wrong
+        // orientation (which might occur in 3d), we have already made sure that
+        // we're ok by picking the correct vertices (this happens automatically
+        // in the vertex() function). however, if the line is in wrong
+        // orientation, we look at it in flipped orientation and we will have to
+        // adjust the shape function indices that we see to correspond to the
+        // correct (face/cell-local) ordering.
+        if (structdim == 2 || structdim == 3)
+          for (unsigned int line = 0;
+               line < GeometryInfo<structdim>::lines_per_cell;
+               ++line)
+            for (unsigned int d = 0; d < dofs_per_line; ++d, ++index)
+              DoFOperation::process_dof(
+                *accessor.line(line),
+                accessor.get_fe(fe_index)
+                  .adjust_line_dof_index_for_line_orientation(
+                    d, accessor.line_orientation(line)),
+                dof_indices[index],
+                fe_index);
+
+        // 3) copy dof numbers from the FACE. for faces with the wrong
+        // orientation, we have already made sure that we're ok by picking the
+        // correct lines and vertices (this happens automatically in the line()
+        // and vertex() functions). however, if the face is in wrong
+        // orientation, we look at it in flipped orientation and we will have to
+        // adjust the shape function indices that we see to correspond to the
+        // correct (cell-local) ordering. The same applies, if the face_rotation
+        // or face_orientation is non-standard
+        if (structdim == 3)
+          for (unsigned int quad = 0;
+               quad < GeometryInfo<structdim>::quads_per_cell;
+               ++quad)
+            for (unsigned int d = 0; d < dofs_per_quad; ++d, ++index)
+              DoFOperation::process_dof(
+                *accessor.quad(quad),
+                accessor.get_fe(fe_index)
+                  .adjust_quad_dof_index_for_face_orientation(
+                    d,
+                    accessor.face_orientation(quad),
+                    accessor.face_flip(quad),
+                    accessor.face_rotation(quad)),
+                dof_indices[index],
+                fe_index);
+
+        // 4) INNER dofs
+        for (unsigned int d = 0; d < inner_dofs; ++d, ++index)
+          DoFOperation::process_dof(accessor, d, dof_indices[index], fe_index);
+
+        AssertDimension(dof_indices.size(), index);
+      }
+
+
+
+      /**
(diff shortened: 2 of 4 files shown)


peterrum:
In my second commit, I moved the logic of looping over vertex/line/quad/inner dofs to a single place and have introduced new setter/getter helper classes (just like it is the case in the `MatrixFree` infrastructure - see also `include/deal.II/matrix_free/vector_access_internal.h`).

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1588,6 +1588,223 @@ namespace internal
...
+              Operation::process_dof(
+                *accessor.quad(quad),
+                accessor.get_fe(fe_index)
+                  .adjust_quad_dof_index_for_face_orientation(
+                    d,
+                    accessor.face_orientation(quad),
+                    accessor.face_flip(quad),
+                    accessor.face_rotation(quad)),
+                dof_indices[index],
+                fe_index);
```
peterrum, line 1668: A side comments regarding hp (unrelated the changes in this PR):  here and above the global `fe_index` is passed, which might lead to the fact that - during the call of this function (inside of the loop) - the local line/quad `fe_index`/offset has to be recompute. I have my doubt that the compiler can optimize it away...

I think it might be an option to precompute the local `fe_index`/offset before the loop.

include/deal.II/dofs/dof_accessor.h
```diff
@@ -1288,6 +1288,11 @@ class DoFInvalidAccessor : public InvalidAccessor<structdim, dim, spacedim>
   template <typename OtherAccessor>
   DoFInvalidAccessor(const OtherAccessor &);
 
+  types::global_dof_index
```
kronbichler, line 1291: Can you add a comment for this function?

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1588,6 +1588,223 @@ namespace internal
           *local_values_begin = read_write_vector[sorted_indices_pos[i]];
       }
 #endif
+
+
+
+      template <typename DoFHandlerType,
```
kronbichler, line 1594: Can you add a short comment here, especially regarding the `Operation` type. While it is used elsewhere, it takes a little to grasp it, so let's be explicit.

include/deal.II/dofs/dof_accessor.templates.h
kronbichler: How heavy are the `quad(quad)` or `get_fe(fe_index)` functions that could be moved out. I agree that the compiler won't easily optimize it because we're potentially modifying the object it operates on. In any case, I would not bother too much unless it shows up in profilers.

kronbichler: So maybe it is best to simply leave a comment in case someone comes back to this function?

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1588,6 +1588,223 @@ namespace internal
...
+      template <typename DoFHandlerType, bool level_dof_access, int structdim>
+      static void
+      set_dof_indices(
+        const dealii::DoFAccessor<structdim, DoFHandlerType, level_dof_access>
+          &                                         accessor,
+        const std::vector<types::global_dof_index> &dof_indices,
+        const unsigned int                          fe_index)
+      {
+        // Note: this function is as general as `get_dof_indices()`. This
+        // assert is placed here only since it is currently only used by the
```
kronbichler, line 1793:
```suggestion
        // assert is placed here since it is currently only used by the
```

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1588,6 +1588,223 @@ namespace internal
...
+          const unsigned int)
+        {
+          Assert(false, ExcInternalError());
+        }
+      };
+
+
+
+      template <typename DoFHandlerType, bool level_dof_access, int structdim>
+      class DoFIndexSetter
```
kronbichler, line 1724: Even though it is internal, I would still suggest to add a comment for this struct - and let it be a struct rather than a class because there is no access restriction.

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1588,6 +1588,223 @@ namespace internal
...
+        for (unsigned int d = 0; d < inner_dofs; ++d, ++index)
+          Operation::process_dof(accessor, d, dof_indices[index], fe_index);
+
+        AssertDimension(dof_indices.size(), index);
+      }
+
+
+
+      template <typename DoFHandlerType, bool level_dof_access, int structdim>
+      class DoFIndexGetter
```
kronbichler, line 1680: Even though it is internal, I would still suggest to add a comment for this struct - and let it be a struct rather than a class because there is no access restriction.

include/deal.II/dofs/dof_accessor.templates.h
peterrum: Let's rather open an issue. I see `quad()`/`line()` as a potential problem since these have to go through multiple indirections due to re-orientations. I bigger problem is the `fe_index` because it might to have to perform a search at a lot of points. But let's investigate once the merged `DoFHandler` is merged!

peterrum:
@kronbichler Thanks for the review. I have added some comments!

include/deal.II/dofs/dof_accessor.h
```diff
@@ -1288,6 +1288,16 @@ class DoFInvalidAccessor : public InvalidAccessor<structdim, dim, spacedim>
   template <typename OtherAccessor>
   DoFInvalidAccessor(const OtherAccessor &);
 
+  /**
+   * Return the index of the <i>i</i>th degree of freedom of this object to @p
```
kronbichler, line 1292: Avoid line break after @p. You could use markup style ` instead.

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1588,6 +1588,252 @@ namespace internal
           *local_values_begin = read_write_vector[sorted_indices_pos[i]];
       }
 #endif
+
+
+
+      /**
+       * Loop over all degrees of freedom of the object described by the
+       * provided @p accessor and @p fe_index and perform the static functions,
+       * which is provided by DoFOperation, on these.
```
kronbichler, line 1597:
```suggestion
       * provided @p accessor and @p fe_index and perform the static functions
       * provided by DoFOperation (set/get) on these.
```

peterrum:
Do we need to restart the tests? Merging seems not to be possible...

kronbichler:
I would wait for at least one more approval anyway. In the meantime you might want to rebase to master to get it running. For Travis CI there is a button somewhere to re-start it, but I have not found it yet. :wink:

masterleinad:
> For Travis CI there is a button somewhere to re-start it, but I have not found it yet. wink

It's on https://travis-ci.org/github/dealii/dealii/builds/689782597. I restarted.

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1588,6 +1588,252 @@ namespace internal
...
+        const dealii::DoFAccessor<structdim, DoFHandlerType, level_dof_access>
+          &                                         accessor,
+        const std::vector<types::global_dof_index> &dof_indices,
+        const unsigned int                          fe_index)
+      {
+        // Note: this function is as general as `get_dof_indices()`. This
+        // assert is placed here since it is currently only used by the
+        // function DoFCellAccessor::set_dof_indices(), which is called by
+        // internal::DoFHandlerImplementation::Policy::Implementation::distribute_dofs().
+        // In the case of new use cases, this assert can be removed.
```
Rombur, line 1825: Then what's the point of this assert, if the function always works?

peterrum: > Then what's the point of this assert, if the function always works?

We setting dofs till now only cell-wise. As long as we don't need s.th. more sophisticated I would like to guarantee that this is done also in the (near) future. My intention with this PR was mainly to merge the logic of 4 functions and simplify it and not add functions, which can be accessed by the user.

> This needs a test.

What kind of tests are you thinking about? These are internal implementations. The functions are already tested by all tests, which use the `DoFHandler`.

PR #10274, opened 2020-05-19
Fixed parameters for step-70.

### LINKED ISSUE #10273: step-70: included prms are broken
The included prms ``parameters.prm`` and ``parameters_23.prm`` are not descriptive (what is the difference?) and don't work. Annoyingly, they crash with the indescriptive error
```
Running StokesImmersedProblem<2,3> using Trilinos.
---------------------------------------------------------
TimerOutput objects finalize timed values printed to the
screen by communicating over MPI in their destructors.
Since an exception is currently uncaught, this
synchronization (and subsequent output) will be skipped
to avoid a possible deadlock.
---------------------------------------------------------


----------------------------------------------------
Exception on processing: 

An error occurred in line <1360> of file </ssd/deal-git/source/base/parameter_handler.cc> in function
    void dealii::ParameterHandler::print_parameters(const std::string &, const ParameterHandler::OutputStyle) const
The violated condition was: 
    out
Additional information: 
    An input/output error has occurred. There are a number of reasons why this may be happening, both for reading and writing operations.
```

@luca-heltai ?

luca-heltai:
Closes #10273

### DIFF
--- a/examples/step-70/parameters.prm
+++ b/examples/step-70/parameters.prm
@@ -1,3 +1,5 @@
+# This is a parameter file that can be used to reproduce the two dimensional
+# results of the tutorial program step-70.
 subsection Stokes Immersed Problem
   set Final time                         = 1
   set Homogeneous Dirichlet boundary ids = 0
--- a/examples/step-70/parameters.prm
+++ b/examples/step-70/parameters.prm
@@ -8,7 +8,7 @@ subsection Stokes Immersed Problem
   set Number of time steps               = 501
   set Velocity degree                    = 2
   set Viscosity                          = 1
-  set Output directory                   = results
+  set Output directory                   = .
   subsection Angular velocity
     set Function constants  = 
     set Function expression = t < .500001 ? 6.283185 : -6.283185 # default: 0
--- a/examples/step-70/parameters_23.prm
+++ b/examples/step-70/parameters_23.prm
@@ -1,3 +1,12 @@
+# This is a parameter file that can be used to reproduce the three dimensional
+# results of the tutorial program step-70. The name of the parameter file
+# indicates that the program will expect to work with a fluid grid in three
+# dimensions, and a solid grid representing a surface embedded in three
+# dimensions. 
+#
+# With this parameter file, we generate a hyper rectangular grid sligthly
+# larger w.r.t. to a impeller grid that was generated from the CAD files
+# downloaded from https://grabcad.com/library/lungstors-blower-1
 subsection Stokes Immersed Problem
   set Final time                            = 1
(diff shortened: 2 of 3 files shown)


luca-heltai:
Let me add a comment in the files...

examples/step-70/parameters.prm
```diff
@@ -8,7 +10,7 @@ subsection Stokes Immersed Problem
   set Number of time steps               = 501
   set Velocity degree                    = 2
   set Viscosity                          = 1
-  set Output directory                   = results
+  set Output directory                   = .
```
bangerth, line 13: It would probably be good practice to check whether the directory exists and, if not, create it. (Or just unconditionally create it with `mkdir -p`.)

examples/step-70/parameters_23.prm
```diff
@@ -1,3 +1,12 @@
+# This is a parameter file that can be used to reproduce the three dimensional
+# results of the tutorial program step-70. The name of the parameter file
+# indicates that the program will expect to work with a fluid grid in three
+# dimensions, and a solid grid representing a surface embedded in three
+# dimensions. 
+#
+# With this parameter file, we generate a hyper rectangular grid sligthly
```
bangerth, line 7:
```suggestion
# With this parameter file, we generate a hyper rectangular grid slightly
```

examples/step-70/parameters.prm
luca-heltai: Is it easy to do in a cross OS compatible way? Does `mkdir -p` work on windows?

bangerth: We use this in ASPECT, which I think should be portable:
```
    int
    mkdirp(std::string pathname,const mode_t mode)
    {
      // force trailing / so we can handle everything in loop
      if (pathname[pathname.size()-1] != '/')
        {
          pathname += '/';
        }

      size_t pre = 0;
      size_t pos;

      while ((pos = pathname.find_first_of('/',pre)) != std::string::npos)
        {
          const std::string subdir = pathname.substr(0,pos++);
          pre = pos;

          // if leading '/', first string is 0 length
          if (subdir.size() == 0)
            continue;

          int mkdir_return_value;
          if ((mkdir_return_value = mkdir(subdir.c_str(),mode)) && (errno != EEXIST))
            return mkdir_return_value;

        }

      return 0;
    }
```

I don't think this has to happen for 9.2, but it would probably be useful in the long run.

tjhei: let's do this after 9.2 please.

bangerth:
Oh, did we not merge this in time for the release? Do the input files in the release actually work?

PR #10280, opened 2020-05-19
Link in step-70 pictures and videos.

bangerth:
Addresses the broken links mentioned in #10243 at least for the 2d case. I think that the video is actually a better visualization than the still images in https://gist.github.com/blaisb/a211c601067bb853a23c5beae3af0d9c The youtube video also has more features than embedding an animated gif, so I didn't take that one.

What's still missing are the 3d results, though.

### DIFF
--- a/examples/step-70/doc/results.dox
+++ b/examples/step-70/doc/results.dox
@@ -327,28 +327,27 @@ assembling the Stokes part. This depends highly on the number of Gauss points
 In the present case, a relatively low number of tracer particles are used.
 Consequently, tracking their motion is relatively cheap.
 
-The following images present the initial and the final configuration of the
-simulation domain:
+The following movie shows the evolution of the solution over time:
 
-
+@htmlonly
 <p align="center">
-   <div class="img" align="center">
-      <img src="https://www.dealii.org/images/steps/developer/step-70.initial_configuration.png"
-           alt = ""
-           width="500">
-    </div>
-</p>
+  <iframe width="560" height="315" src="https://www.youtube.com/embed/y4Gypj2jpXw"
+   frameborder="0"
+   allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
+   allowfullscreen></iframe>
+ </p>
+@endhtmlonly
 
+The movie shows the rotating obstacle in gray (actually a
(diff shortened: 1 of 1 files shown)


luca-heltai:
https://youtu.be/Srwq7zyR9mg

luca-heltai:
Ps: if you want to play with the 3d results, here you'll find (as soon as they finish uploading), the vtu files I used to generate the movie above. 

https://gist.github.com/luca-heltai/b459c57000a70c4b8e7c0e72bba5ca07

blaisb:
> Addresses the broken links mentioned in #10243 at least for the 2d case. I think that the video is actually a better visualization than the still images in https://gist.github.com/blaisb/a211c601067bb853a23c5beae3af0d9c The youtube video also has more features than embedding an animated gif, so I didn't take that one.
> 
> What's still missing are the 3d results, though.

Great, love the youtube animation :)!

bangerth:
I also linked in the 3d video. Thanks!

bangerth:
@luca-heltai -- I've played a bit with the 3d files and there is something funny going on with them. The velocity of the fluid drastically changes from one time step to the next and seems to oscillate somehow between about 200 and 1100 as max. Is this expected? I would have assumed that from one time step to the next, the velocity remains roughly the same....

luca-heltai:
> @luca-heltai -- I've played a bit with the 3d files and there is something funny going on with them. The velocity of the fluid drastically changes from one time step to the next and seems to oscillate somehow between about 200 and 1100 as max. Is this expected? I would have assumed that from one time step to the next, the velocity remains roughly the same....

It is usually expected that with high penalty parameters, the L infty norm of the velocity may vary a lot according to where the solid points end up being in the fluid. The L2 norm is still under control, but you get spikes in the velocity here and there.

This happens especially when there are too many quadrature points for the solid part w.r.t. to the fluid part (i.e., when the fluid is not resolved enough). The simulation was admittedly a small one, just to get some quick results out also in the 3d case. Refining the fluid a couple of times more sould fix the issue.

bangerth:
@luca-heltai But it isn't just the L-infinity norm. Have you looked at the velocity field? It *globally* oscillates in magnitude. The *pattern* stays the same from time step to time step, but the magnitude changes drastically.

PR #10282, opened 2020-05-20
BoundingBoxDataOut

luca-heltai:
While preparing my slides for the workshop, I figured out that I needed some ways to output bounding boxes of rtrees -- and possibly their levels -- in a nicer way w.r.t. to gnuplot...

image

image


include/deal.II/numerics/rtree.h
```diff
@@ -452,15 +475,67 @@ extract_rtree_level(const Rtree &tree, const unsigned int level)
...
+      for (unsigned int d = 0; d < GeometryInfo<dim>::vertices_per_cell; ++d)
+        out << " " << i++;
+      out << std::endl;
+    }
+  const unsigned int vtk_cell_type = (dim == 1 ? 3 : dim == 2 ? 8 : 11);
+  out << std::endl << "CELL_TYPES " << N;
+  for (unsigned int b = 0; b < N; ++b)
+    out << " " << vtk_cell_type;
+  out << std::endl;
+};
```
masterleinad, line 529:
```
/home/runner/work/dealii/dealii/include/deal.II/numerics/rtree.h:529:2: error: extra ‘;’ [-Werror=pedantic]
```

luca-heltai:
> With relatively little effort, it would be possible to make this work with any of the graphical output formats we support. In essence, all you have to do is convert your collection of bounding boxes into a `std::vector<Patch>` in the same way as the `MatrixOut` class does. Then this would also work with gnuplot output, vtu, etc.

You are right, I didn't think about this. :) On it.

luca-heltai:
@bangerth, is this what you had in mind?

include/deal.II/base/bounding_box_data_out.h
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * @author Luca Heltai, 2020.
+ */
+template <int dim>
+class BoundingBoxDataOut : public DataOutInterface<dim, dim>
+{
+public:
+  BoundingBoxDataOut() = default;
+
+  ~BoundingBoxDataOut() = default;
```
bangerth, line 50: Documentation missing for these two functions, even if only one word :-)

include/deal.II/base/bounding_box_data_out.h
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * Generate patches from a range of objects that can be converted by boost to
+   * a collection of BoundingBox objects.
+   */
+  template <class ConvertibleToBoundingBoxIterator>
+  void
+  build_patches(const ConvertibleToBoundingBoxIterator &begin,
```
bangerth, line 58: What are examples of these function arguments? Just iterators over collections of `BoundingBox` type, or are there other things that are convertible to `BoundingBox` objects?

include/deal.II/base/bounding_box_data_out.h
```diff
@@ -0,0 +1,159 @@
...
+private:
+  /**
+   * The actual boxes.
+   */
+  std::vector<DataOutBase::Patch<dim, dim>> patches;
+
+  /**
+   * Names of datasets.
+   */
+  std::vector<std::string> dataset_names;
```
bangerth, line 87: I think that in the code, you only ever clear this variable, but never set it. Regardless, since you don't determine the name from any argument, and since `get_names()` returns by value, I don't think you need this variable at all.

include/deal.II/base/bounding_box_data_out.h
```diff
@@ -0,0 +1,159 @@
...
+  return patches;
+}
+
+
+
+template <int dim>
+std::vector<std::string>
+BoundingBoxDataOut<dim>::get_dataset_names() const
+{
+  return dataset_names;
```
bangerth, line 152: Might want to return something useful here.

luca-heltai:
As far as naming is concerned, we have `Particles::DataOut` (not ParticlesOut, as far as I could see). Since `BoundingBox` does not have a namespace, I thought this would be ok. Just let me know what names suits you, and I'll change that.

PR #10285, opened 2020-05-20
ThreadLocalStorage: Reimplement using STL C++14 functionality

### LINKED ISSUE #10284: Bug in AffineConstraintsData
Here's a bug that takes a good long time to explain -- story time: 

`AffineConstraints::distribute_local_to_global()` in its various incarnations uses an object of type `internals::AffineConstraintsData<number>` as a scratch object to work on. That class is, in essence, a singleton:
```
  template <typename number>
  class AffineConstraintsData
  {
  public:
    struct ScratchData
    { ... };

   ...

  private:
    /**
     * The actual data object that contains a scratch data for each thread.
     */
    static Threads::ThreadLocalStorage<ScratchData> scratch_data;
  };
```
All accesses are then through the `static` member variable, which is thread-local and consequently doesn't get in the way as long as each thread on which `AffineConstraints::distribute_local_to_global()` doesn't get interrupted by another call to the same function while it works. (But there is a check that would abort the program with an error if that were to happen.)

The bug now happens in the following context: If one configures deal.II with `DEAL_II_WITH_THREADS=OFF`, then the thread-local variable expands to just one global variable (it's `static` after all). That's fine if we're really just running single-threaded, but maybe we want to use `std::async` anyway -- in other words, we don't want to use the TBB facilities but just the C++11 ones. Why would one want to do that? Because we don't support the TBB on Windows -- so `DEAL_II_WITH_THREADS=OFF`, but maybe we want to run a few things in parallel anyway. For example, if one wanted to solve the PDE with different parameter values: an embarrassingly parallel  problem.

Well, back to the issue at hand: `scratch_data` in that case exists only once, but multiple threads all want to call `AffineConstraints::distribute_local_to_global()` on their respective data structures, and that is going to create trouble.

So how do we address this kind of thing?

tamiko:
This obviously doesn't compile at the moment.

@bangerth This is what I suggest: Simply reimplement the functionality with C++11 STL functionality.

Closes #10284

### DIFF
--- a/include/deal.II/base/thread_local_storage.h
+++ b/include/deal.II/base/thread_local_storage.h
@@ -19,6 +19,9 @@
 
 #  include <deal.II/base/config.h>
 
+#  include <deal.II/base/exceptions.h>
+
+#  include <map>
 #  include <shared_mutex>
 #  include <thread>
 
@@ -62,8 +65,8 @@ namespace Threads
   class ThreadLocalStorage
   {
     static_assert(
-      std::is_trivially_copyable<T>::value ||
-        std::is_copy_constructible<T>::value,
+      std::is_copy_constructible<T>::value ||
+        std::is_default_constructible<T>::value,
       "The stored type must be either copyable, or default constructible");
 
   public:
@@ -80,10 +83,20 @@ namespace Threads
     explicit ThreadLocalStorage(const T &t);
 
     /**
+     * FIXME
+     */
+    explicit ThreadLocalStorage(T &&t);
+
+    /**
      * Copy constructor. Initialize each thread local object with the
      * corresponding object of the given object.
      */
     ThreadLocalStorage(const ThreadLocalStorage<T> &t) = default;
+
+    /**
+     * Move constructor. FIXME
+     */
+    ThreadLocalStorage(ThreadLocalStorage<T> &&t) = default;
 
     /**
      * Return a reference to the data stored by this object for the current
@@ -130,6 +143,13 @@ namespace Threads
      */
     ThreadLocalStorage<T> &
     operator=(const T &t);
+
+
+    /**
+     * FIXME
+     */
+    ThreadLocalStorage<T> &
+    operator=(T &&t);
 
     /**
      * Remove the thread-local objects stored for all threads that have
@@ -196,6 +216,13 @@ namespace Threads
 
 
   template <typename T>
+  inline ThreadLocalStorage<T>::ThreadLocalStorage(T &&t)
+  {
+    exemplar = std::make_shared<T>(std::forward<T>(t));
+  }
+
+
+  template <typename T>
   inline T &
   ThreadLocalStorage<T>::get(bool &exists)
   {
@@ -212,7 +239,8 @@ namespace Threads
       // Take a shared ("reader") lock for lookup and record the fact
       // whether we could find an entry in the boolean exists.
       std::shared_lock<decltype(insertion_mutex)> lock(insertion_mutex);
-      const auto                                  it = data.find(my_id);
+
+      const auto it = data.find(my_id);
       if (it != data.end())
         {
           exists = true;
@@ -230,19 +258,19 @@ namespace Threads
       //
       std::unique_lock<decltype(insertion_mutex)> lock(insertion_mutex);
 
-      if constexpr (std::is_trivially_copyable<T>::value)
+      if constexpr (std::is_copy_constructible<T>::value)
         if (exemplar)
           {
-            const auto it =
-              data.emplace(std::make_pair(my_id, *exemplar)).first;
+            const auto it = data.emplace(my_id, *exemplar).first;
             return it->second;
           }
 
       if constexpr (std::is_default_constructible<T>::value)
         {
-          const auto it = data.emplace(std::make_pair(my_id, T())).first;
-          return it->second;
+          return data[my_id]; // default construct
         }
+
+      Assert(false, ExcInternalError());
     }
   }
 
@@ -268,6 +296,15 @@ namespace Threads
   ThreadLocalStorage<T>::operator=(const T &t)
   {
     get() = t;
+    return *this;
+  }
+
+
+  template <typename T>
+  inline ThreadLocalStorage<T> &
+  ThreadLocalStorage<T>::operator=(T &&t)
+  {
+    get() = std::forward<T>(t);
     return *this;
   }
 
--- a/include/deal.II/base/thread_local_storage.h
+++ b/include/deal.II/base/thread_local_storage.h
@@ -34,6 +34,30 @@ DEAL_II_NAMESPACE_OPEN
 
 namespace Threads
 {
+#  ifndef DOXYGEN
+  namespace
+  {
+    /*
+     * Workaround: The standard unfortunately has a flaw in the type traits
+     * when it comes to std::is_copy_constructible and a std::vector<T>
+     * containing non-copyable objects T. The type trait is true even
+     * though any attempted invocation leads to a compilation error. Work
+     * around this issue by removing the "std::vector" container from the
+     * type trait.
+     */
+    template <typename T>
+    struct unpack_vector
+    {
+      using type = T;
+    };
+    template <typename T, typename A>
+    struct unpack_vector<std::vector<T, A>>
+    {
+      using type = T;
+    };
+  } // namespace
+#  endif
+
   /**
    * @brief A class that provides a separate storage location on each thread
    * that accesses the object.
@@ -65,7 +89,7 @@ namespace Threads
   class ThreadLocalStorage
   {
     static_assert(
-      std::is_copy_constructible<T>::value ||
+      std::is_copy_constructible<typename unpack_vector<T>::type>::value ||
         std::is_default_constructible<T>::value,
       "The stored type must be either copyable, or default constructible");
 
@@ -191,7 +215,7 @@ namespace Threads
     /**
      * An exemplar for creating a new (thread specific) copy.
      */
-    std::shared_ptr<T> exemplar;
+    std::shared_ptr<const T> exemplar;
   };
 } // namespace Threads
 /**
@@ -210,16 +234,14 @@ namespace Threads
 
   template <typename T>
   inline ThreadLocalStorage<T>::ThreadLocalStorage(const T &t)
-  {
-    exemplar = std::make_shared<T>(t);
-  }
+    : exemplar(std::make_shared<const T>(t))
+  {}
 
 
   template <typename T>
   inline ThreadLocalStorage<T>::ThreadLocalStorage(T &&t)
-  {
-    exemplar = std::make_shared<T>(std::forward<T>(t));
-  }
+    : exemplar(std::make_shared<T>(std::forward<T>(t)))
+  {}
 
 
   template <typename T>
@@ -258,7 +280,8 @@ namespace Threads
       //
       std::unique_lock<decltype(insertion_mutex)> lock(insertion_mutex);
 
-      if constexpr (std::is_copy_constructible<T>::value)
+      if constexpr (std::is_copy_constructible<
+                      typename unpack_vector<T>::type>::value)
         if (exemplar)
           {
             const auto it = data.emplace(my_id, *exemplar).first;
--- a/include/deal.II/base/thread_local_storage.h
+++ b/include/deal.II/base/thread_local_storage.h
@@ -104,13 +104,16 @@ namespace Threads
     ThreadLocalStorage();
 
     /**
-     * A kind of copy constructor. Initialize each thread local object by
-     * copying the given object.
+     * A kind of copy constructor. Initializes an internal exemplar by the
+     * given object. The exemplar is in turn used to initialize each thread
+     * local object instead of invoking the default constructor.
      */
     explicit ThreadLocalStorage(const T &t);
 
     /**
-     * FIXME
+     * A kind of move constructor. Moves the given object into an internal
+     * exemplar. The exemplar is in turn used to initialize each thread
+     * local object instead of invoking the default constructor.
      */
     explicit ThreadLocalStorage(T &&t);
 
@@ -121,7 +124,8 @@ namespace Threads
     ThreadLocalStorage(const ThreadLocalStorage<T> &t) = default;
 
     /**
-     * Move constructor. FIXME
+     * Move constructor. Copies the internal state over from the given
+     * object.
      */
     ThreadLocalStorage(ThreadLocalStorage<T> &&t) = default;
 
@@ -173,7 +177,17 @@ namespace Threads
 
 
     /**
-     * FIXME
+     * Move the given argument into the storage space used to represent the
+     * current thread. Calling this function as <code>tls_data =
+     * object</code> is equivalent to calling <code>tls_data.get() =
+     * object</code>. The intent of this operator is to make the
+     * ThreadLocalStorage object look more like the object it represents on
+     * the current thread. Move assignment operator.
+     *
+     * @param t The object to be copied into the storage space used for the
+     * current thread.
+     *
+     * @return The current object, after the changes have been made
      */
     ThreadLocalStorage<T> &
     operator=(T &&t);
@@ -282,7 +296,6 @@ namespace Threads
     {
       // Take a unique ("writer") lock for manipulating the std::map. This
       // lock ensures that no other threat does a lookup at the same time.
-      //
       std::unique_lock<decltype(insertion_mutex)> lock(insertion_mutex);
 
       if constexpr (std::is_copy_constructible<
@@ -295,7 +308,7 @@ namespace Threads
 
       if constexpr (std::is_default_constructible<T>::value)
         {
-          return data[my_id]; // default construct
+          return data[my_id]; // invokes default constructor
         }
 
       Assert(false, ExcInternalError());
--- a/include/deal.II/base/thread_local_storage.h
+++ b/include/deal.II/base/thread_local_storage.h
@@ -19,12 +19,8 @@
 
 #  include <deal.II/base/config.h>
 
-DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
-#  ifdef DEAL_II_WITH_THREADS
-#    include <tbb/enumerable_thread_specific.h>
-#  endif
-DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
-
+#  include <shared_mutex>
+#  include <thread>
 
 
 DEAL_II_NAMESPACE_OPEN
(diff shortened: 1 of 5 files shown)


include/deal.II/base/thread_local_storage.h
```diff
@@ -158,29 +147,21 @@ namespace Threads
...
-     * implementation. This function is really only useful if deal.II has been
-     * configured with multithreading and has no useful purpose otherwise.
+     * The data element we store.
      */
-#  ifdef DEAL_II_WITH_THREADS
-    tbb::enumerable_thread_specific<T> &
-#  else
-    T &
-#  endif
-    get_implementation();
```
tamiko, line 171: This is strictly speaking an incompatible change. What about we deprecate the `get_implementation` method quickly in the 9.2 release?

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +179,59 @@ namespace Threads
...
+        return it->second;
+      }
+
+    exists = false;
+
+    std::lock_guard lock(insertion_mutex);
+
+    // FIXME: Port to C++11
+
+    if constexpr (std::is_trivially_copyable<T>::value)
```
masterleinad, line 212: Like it! :smile:

include/deal.II/base/thread_local_storage.h
tjhei: I think we are done with 9.2.

bangerth: Let's not do anything with 9.2 any more. But I wouldn't worry about it -- it's surely a class where we can afford to remove an obscure function without a deprecation period.

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +179,59 @@ namespace Threads
...
 
 
   template <typename T>
-  inline ThreadLocalStorage<T>::ThreadLocalStorage(
-    const ThreadLocalStorage<T> &t)
-    : data(t)
-  {}
+  inline ThreadLocalStorage<T>::ThreadLocalStorage(const T &t)
+  {
+    exemplar = std::make_shared<T>(t);
```
bangerth, line 189: Move it into the member initializer list :-)

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +179,59 @@ namespace Threads
...
   {
-#    ifdef DEAL_II_WITH_THREADS
-    return data.local();
-#    else
-    return data;
-#    endif
+    const std::thread::id my_id = std::this_thread::get_id();
+
+    auto it = data.find(my_id);
+    if (it != data.end())
```
bangerth, line 200: You're accessing `data` here. You need to move the lock guard above this point.

bangerth: Also, this is most easily written as
```
  if (const auto it = data.find(...))
```
and then you don't need to declare the local variable here.

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +179,59 @@ namespace Threads
...
+    exists = false;
+
+    std::lock_guard lock(insertion_mutex);
+
+    // FIXME: Port to C++11
+
+    if constexpr (std::is_trivially_copyable<T>::value)
+      if (exemplar)
+        {
+          it = data.insert(std::make_pair(my_id, *exemplar)).first;
```
bangerth, line 215: I think you can use `emplace` and then all you need is that the object is copy constructible.

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +179,59 @@ namespace Threads
...
+          return it->second;
+        }
+
+    if constexpr (std::is_default_constructible<T>::value)
+      {
+        it = data.insert(std::make_pair(my_id, T())).first;
+        return it->second;
+      }
+
+    __builtin_unreachable();
```
bangerth, line 225: is there now a `[[attribute]]` for this? Either way, we have a style for this: `Assert (false, ExcInternalError());`

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +179,59 @@ namespace Threads
...
+      if (exemplar)
+        {
+          it = data.insert(std::make_pair(my_id, *exemplar)).first;
+          return it->second;
+        }
+
+    if constexpr (std::is_default_constructible<T>::value)
+      {
+        it = data.insert(std::make_pair(my_id, T())).first;
+        return it->second;
```
bangerth, line 222: Question: Does `std::map` guarantee that references to values remain valid? In other words, that `std::map` never moves objects around? That's because you're returning a reference to the map's value here and above, so it better stay at that memory location.

A safer design would be to make the map's value type a `std::unique_ptr` to a `T` and return the deferenced pointer. We know that that remains at that memory location.

tamiko:
@bangerth Thanks for the quick feedback. This is really a very quick first prototype. I didn't even compile the whole library.

include/deal.II/base/thread_local_storage.h
tamiko: The STL guarantees that `insert` and `emplace` do not invalidate references and iterators. I would prefer to use a `std::map<..., T>` over `std::map<..., std::unique_ptr<T>>` because the latter doesn't have a copy constructor (that we need, because our `ThreadLocalStorage` offers a copy constructor).

include/deal.II/base/thread_local_storage.h
tamiko: This is actually the perfect example of a reader-writer lock. Shall we bump to C++14 in order to use `std::shared_lock` and `std::unique_lock`?

include/deal.II/base/thread_local_storage.h
bangerth: Well, but the requirement is that the *iterator* remains stable, not that what you get by derefencing it remains stable. The iterator could contain a pointer to the map and an integer for an array element in which the element is stored, which continues to work if the map decides to re-allocate its array somewhere else. But the dereferenced address may not work any more after that.

Copy operator: We might simply have to do the copying by hand :-)

include/deal.II/base/thread_local_storage.h
bangerth: Right, there's probably a design pattern for that, but the code as is is still wrong ;-)

include/deal.II/base/thread_local_storage.h
tamiko: The requirement is precisely that if you hold a reference to a stored object that this reference remains stable.

tamiko: (in particular [res.on.data.races] and [container.requirements.dataraces])

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +184,75 @@ namespace Threads
...
+        }
+      else
+        {
+          exists = false;
+        }
+    }
+
+    {
+      // Take a unique ("writer") lock for manipulating the std::map. This
+      // lock ensures that no other threat does a lookup at the same time.
```
tjhei, line 229:
```suggestion
      // lock ensures that no other thread does a lookup at the same time.
```

include/deal.II/base/thread_local_storage.h
```diff
@@ -246,29 +272,12 @@ namespace Threads
...
 
 
-  template <typename T>
-  inline
-#    ifdef DEAL_II_WITH_THREADS
-    tbb::enumerable_thread_specific<T> &
-#    else
-    T &
-#    endif
-    ThreadLocalStorage<T>::get_implementation()
```
tjhei, line 256: do you want to mention this as an incompatibility in the changelog?

tamiko: Yes. Done. :smile:

include/deal.II/base/thread_local_storage.h
```diff
@@ -158,29 +153,25 @@ namespace Threads
...
-     * be of a type that provides a separate object for each thread.
-     * Otherwise, it is simply a single object of type T.
+     * A mutex to guard insertion into the data object.
      */
-    tbb::enumerable_thread_specific<T> data;
+#  ifdef DEAL_II_HAVE_CXX17
+    std::shared_mutex insertion_mutex;
 #  else
-    T data;
+    std::shared_timed_mutex insertion_mutex;
```
bangerth, line 168: Nice idea. I had not known that C++14 had introduced these read/write locks :-) 

I think it would be useful to mention in the documentation why we use this kind of mutex here.

include/deal.II/base/thread_local_storage.h
```diff
@@ -158,29 +153,25 @@ namespace Threads
...
 #  endif
+
+    /**
+     * An exemplar for creating a new (thread specific) copy.
+     */
+    std::shared_ptr<T> exemplar;
```
bangerth, line 174: Could we make this `std::shared_ptr<const T>` to make reasoning about why a shared pointer is the right approach here easier? Intuitively, I wanted to make this a `unique_ptr`, but then realized that `shared_ptr` is entirely sufficient -- but really only because no thread will ever change the exemplar.

include/deal.II/base/thread_local_storage.h
bangerth: I think this is still relevant.

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +184,75 @@ namespace Threads
...
-#    ifdef DEAL_II_WITH_THREADS
-    return data.local();
-#    else
-    return data;
-#    endif
+    const std::thread::id my_id = std::this_thread::get_id();
+
+    // Note that std::map<..>::emplace guarantees that no iterators or
+    // references to stored objects are invalidated. We thus only have to
+    // ensure that we do not performa a lookup while writing, and that we
```
bangerth, line 206:
```suggestion
    // ensure that we do not perform a lookup while writing, and that we
```

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +184,75 @@ namespace Threads
...
+    // do not write concurrently. This is precisely the "reader-writer
+    // lock" paradigm supported by C++14 by means of the std::shared_lock
+    // and th std::unique_lock
```
bangerth, line 209:
```suggestion
    // and the std::unique_lock
```

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +184,75 @@ namespace Threads
...
+      // Take a unique ("writer") lock for manipulating the std::map. This
+      // lock ensures that no other thread does a lookup at the same time.
+      //
+      std::unique_lock<decltype(insertion_mutex)> lock(insertion_mutex);
+
+      if constexpr (std::is_trivially_copyable<T>::value)
+        if (exemplar)
+          {
+            const auto it =
+              data.emplace(std::make_pair(my_id, *exemplar)).first;
```
bangerth, line 237: You can save one move operation using
```suggestion
              data.emplace(my_id, *exemplar).first;
```

include/deal.II/base/thread_local_storage.h
```diff
@@ -193,40 +184,75 @@ namespace Threads
...
+            return it->second;
+          }
+
+      if constexpr (std::is_default_constructible<T>::value)
+        {
+          const auto it = data.emplace(std::make_pair(my_id, T())).first;
```
bangerth, line 243: same here

source/base/logstream.cc
```diff
@@ -385,6 +385,7 @@ LogStream::get_prefixes() const
   bool                     exists         = false;
   std::stack<std::string> &local_prefixes = prefixes.get(exists);
 
+#  if 0 // FIXME
```
bangerth, line 388: What had you planned to do here?

tamiko:
@bangerth Let me finish up the pull request before you review :smile:

source/base/logstream.cc
tamiko: @bangerth I will make logstream a friend and grab the corresponding information directly out of the map.

include/deal.II/base/thread_local_storage.h
tamiko: Already fixed :-)

bangerth:
I didn't think there was much to finish :-) What is the issue with `LogStream`? What information does it need that's not available through the public interface?

tamiko:
@bangerth LogStream traditionally copied the current state from the first thread. I.e. if the main thread had prefix `FOO::BAR` then the thread inherited `FOO::BAR` on first use of deallog. This is different from the `exemplar` that would always default to `DEAL::`.

tamiko:
Fun: https://stackoverflow.com/questions/18404108/false-positive-with-is-copy-constructible-on-vectorunique-ptr

bangerth:
On 5/26/20 4:02 PM, Matthias Maier wrote:
> @bangerth <https://github.com/bangerth> LogStream traditionally copied the 
> current state from the first thread. I.e. if the main thread had prefix 
> |FOO::BAR| then the thread inherited |FOO::BAR| on first use of deallog. This 
> is different from the |exemplar| that would always default to |DEAL::|.

Let's not fret over this -- the class is braindead anyway, let's feel free to 
change this kind of thing in incompatible ways.

bangerth:
I think it is legitimate to assume that `T` can be copied. What was the requirement in the old implementation?

tamiko:
@bangerth Less - TBB had very little requirements. We are using `ThreadLocalStorage<std::vector<unique_ptr<...>>>` at least for the function parser.
I have a fix for this in a moment and ping you when I am done.

include/deal.II/base/thread_local_storage.h
tamiko: Funky new C++14 feature, isn't it? :-)

bangerth:
So how did we do it in the past then? Did we just not copy an exemplar and instead default-constructed the object?

tamiko:
@bangerth We used tbb::enumerable_thread_specific that does all kinds of crazy stuff in the background.

tamiko:
@bangerth So - this is what I had in mind. Let's see what the CI has to say.

source/base/logstream.cc
```diff
@@ -389,18 +390,9 @@ LogStream::get_prefixes() const
...
-      const tbb::enumerable_thread_specific<
-        std::stack<std::string>>::const_iterator first_elem = impl.begin();
-
-      if (first_elem != impl.end())
-        {
-          local_prefixes = *first_elem;
-        }
+      auto it = prefixes.data.find(master_thread);
+      if (it != prefixes.data.end())
+        local_prefixes = it->second;
```
tamiko, line 395: @bangerth Actually, the fix isn't too horrible. It just requires an ugly `friend` declaration.

include/deal.II/base/work_stream.h
```diff
@@ -211,6 +211,13 @@ namespace WorkStream
               , currently_in_use(in_use)
             {}
 
+            // Provide a copy constructor that actually doesn't copy the
+            // internal state. This makes handling ScratchAndCopyDataObjects
+            // easier to handle with STL containers.
+            ScratchDataObject(const ScratchDataObject &)
+              : currently_in_use(false)
+            {}
```
tamiko, line 219: I am not happy about this change (i.e., providing a "wrong" copy constructor that doesn't copy). But I grew tired of blowing up the `is_default_constructible` workaround any further.

include/deal.II/base/work_stream.h
```diff
@@ -710,19 +717,11 @@ namespace WorkStream
...
-        //      non-C++11 mode
-        ScratchAndCopyDataObjects(const ScratchAndCopyDataObjects &o)
-          : scratch_data(o.scratch_data)
-          , copy_data(o.copy_data)
-          , currently_in_use(o.currently_in_use)
+        // Provide a copy constructor that actually doesn't copy the
+        // internal state. This makes handling ScratchAndCopyDataObjects
+        // easier to handle with STL containers.
+        ScratchAndCopyDataObjects(const ScratchAndCopyDataObjects &)
+          : currently_in_use(false)
```
tamiko, line 724: This here must have been some leftover from refactoring. `scratch_data(o.scratch_data)` cannot compile - it is a `std::unique_ptr`.

tamiko:
@dealii/dealii I'd like to get this in soon now :-) I want to make some follow-up cleanups.

include/deal.II/base/thread_local_storage.h
```diff
@@ -19,22 +19,56 @@
...
+    {
+      using type = T;
+    };
+
+    template <typename T, typename A>
+    struct unpack_container<std::list<T, A>>
+    {
+      using type = T;
+    };
+  } // namespace
```
drwells, line 69: Could we put this in `Threads::internal`? anonymous namespaces in headers are problematic.

drwells: Good point - I still have a reflexive aversion to using anonymous namespaces in headers but I suppose this is fine.

tamiko: @drwells I did put the `construct_element` overloads into an internal namespace because these will be potentially instantiated.

tamiko:
@bangerth Please review and approve.

include/deal.II/base/thread_local_storage.h
bangerth: If it makes @drwells uncomfortable, the namespace should probably be named. It also avoids mistakes at a later point if someone, for whatever reason, decides to put something in there without thinking this through.

tamiko: Fixed.

include/deal.II/base/work_stream.h
drwells: I think this is fine for scratch data where the state is thought of as transitory.

include/deal.II/base/thread_local_storage.h
```diff
@@ -19,22 +19,56 @@
...
-#  endif
-DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
-
+#  include <deal.II/base/exceptions.h>
 
+#  include <list>
+#  include <map>
+#  include <shared_mutex>
+#  include <thread>
+#  include <vector>
```
bangerth, line 28: list, map, AND vector?

include/deal.II/base/thread_local_storage.h
```diff
@@ -158,29 +222,32 @@ namespace Threads
...
+    std::shared_mutex insertion_mutex;
 #  else
-    T data;
+    std::shared_timed_mutex insertion_mutex;
 #  endif
+
+    /**
+     * An exemplar for creating a new (thread specific) copy.
+     */
+    std::shared_ptr<const T> exemplar;
```
bangerth, line 248: thanks for making it `const`

source/base/logstream.cc
bangerth: It's not thread-safe to do this, but neither was the original implementation...

tamiko: Yeah. But I won't fix that. :smiling_imp:

include/deal.II/base/thread_local_storage.h
tamiko: Jup. We need the map because we use it and we need to import vector and list for the workaround found below (it is not possible to forward declare these data types in a portable way). On the bright side - we already include list and vector in all of our compilation units, except for the ThreadLocalStorage tests.

Those also only fail for gcc-5 only, all other compiler seem to already include vector and list.

tamiko:
This branch already compiled successfully on github-CI / OSC serial/parallel and MSVC appveyor three times. :-)

PR #10294, opened 2020-05-21
Random doc improvements.

bangerth:
Scenes from my frequent random walks through the library.


drwells:
As always, thanks for taking the time to fix these things :)

PR #10301, opened 2020-05-22
step-19: A simple particles tutorial.

bangerth:
@gassmoeller @blaisb @dangars  -- FYI!

Here's a picture of how the solution currently looks like:
image
The cathode is the wedge on the left. It is negatively charged. Part of the right boundary is positively charged. So the electric field points from left to right (streamlines shown in oranges) and is largest at the tip. I'm creating particles on the cathode if the electric field exceeds a certain threshold (shown in green).

What I need help with:
- [ ] Moving particles in each time step according to mx''(t)=eE
- [ ] Nothing is currently physically correct. I'm not choosing reasonable voltages, the Laplace equation should be stated with the correct coefficient (the dielectric constant of vaccum), and the threshold beyond which electrons can leave a metal is chosen arbitrarily. 
- [ ] We probably need to set the reference position of each particle
- [ ] We don't current output the particle properties (namely, the velocity).

@gassmoeller @blaisb @dangars -- want to help with any of that?

The current version steals tutorial program number 66 from #8229 , but that remains to be negotiated.


peterrum:
Would something like this work/help? I don't know how to get/set the velocity of particles.
```cpp
  // @sect4{CathodeRaySimulator::move_particles}

  template <int dim>
  void CathodeRaySimulator<dim>::move_particles()
  {
    // advance all particles
    const double dt = time.get_next_step_size();

    for (const auto cell : dof_handler.active_cell_iterators())
      {
        std::vector<Point<dim>> ref_points;

        for (const auto &particle : particle_handler.particles_in_cell(cell))
          ref_points.push_back(particle.get_reference_location());

        if (ref_points.size() == 0)
          continue;

        Quadrature<dim> quad(ref_points);

        FEValues<dim> fe_values(mapping, fe, quad, update_gradients);
        fe_values.reinit(cell);

        std::vector<Tensor<1, dim>> solution_gradients(quad.size());
        fe_values.get_function_gradients(solution, solution_gradients);

        unsigned int i = 0;
        for (auto &particle : particle_handler.particles_in_cell(cell))
          particle.set_location(particle.get_location() +
                                dt * (/*particle->velociy()*/ +dt *
                                      solution_gradients[i])); // TODO?
      }
  }


  // @sect4{CathodeRaySimulator::update_timestep_size}

  template <int dim>
  void CathodeRaySimulator<dim>::update_timestep_size()
  {
    // We need to respect a CFL condition whereby particles can not move further
    // than one cell. Need to compute their speed here and divide the cell size
    // by that speed for all particles, then take the minimum.

    const double cfl = 0.15; // TODO

    double dt = std::numeric_limits<double>::max();

    for (auto cell : triangulation.active_cell_iterators())
      {
        const double minimum_vertex_distance = cell->minimum_vertex_distance();

        double max_velocity = 0.0;

        for (const auto &particle : particle_handler.particles_in_cell(cell))
          max_velocity =
            std::max(max_velocity,
                     1.0 /*particle.get_velocity().norm()*/); // TODO?

        if (max_velocity > 0.0)
          dt = std::min(dt, cfl * minimum_vertex_distance / max_velocity);
      }

    time.set_desired_next_step_size(dt);
  }
```

dangars:
@bangerth I would be happy to help :) although my knowledge of particles is limited. I should have time later today to dig into the code!

bangerth:
@peterrum -- yes, I think that's going to get us 90% there.
@gassmoeller -- are you working on this, or should I continue?

gassmoeller:
I would work on #10308 first to allow @blaisb to continue with that. 

I will look into this PR in the afternoon.

bangerth:
I put a variation and worked out version of @peterrum 's `move_particles()` function into the patch.

bangerth:
Now also with `update_timestep_size()`. Though it doesn't actually work right now.

@gassmoeller -- want to take a look and see what one can do about the TODOs? In particular:
* we need to set/update reference locations
* we need to tell the particle handler to update cell ownership

gassmoeller:
Now this is awkward, did we both implement the same things over the afternoon? My changes are here: bangerth/dealii#17. I though you had asked me to work on the open functions and I said I would look into it in the afternoon?

gassmoeller:
Maybe we should write a comment both when we start working on something and once it is done? I am not planning to do anything else on this PR today.

bangerth:
@gassmoeller I thought you were working on the other program. Be that as it may, I spent an hour (while on my 6th hour of zoom meetings of the day) teaching myself how this is likely going to work, dabbled a bit with @peterrum 's patch, and put something together that doesn't actually run. You spent probably the same amount of time writing code that actually does what is needed. I just have to figure out how to merge your patch now :-)

bangerth:
So, @gassmoeller , thank you -- your work is much appreciated!

bangerth:
@gassmoeller -- now slightly updated. Do you happen to know what is going on with that first wave of particles? Is their trajectory broken because of the unrealistically large initial time step?

bangerth:
Yes, that's what is necessary. I've got a good scheme for the first time step.

bangerth:
We agreed to use the step-19 name on the call this morning.

gassmoeller:
I missed the call this morning, but this looks much better, thanks for the clarification. The branch currently does not compile for me, because it relies on newer changes in master. Do you want me to look into something?

bangerth:
Current state: https://youtu.be/zGNjgLdteck

bangerth:
@gassmoeller -- let me rebase to master. I think it's at a good point right now. If you wanted to look over it, feel free to!

gassmoeller:
Taking a look into the text right now.

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,352 @@
...
+(here, the current location, its reference cell location, and its id)
+and call `insert_particle`. The only thing that may be surprising is
+the reference location: In order to evaluate things such as 
+$\nabla V(\mathbf x_i)$, it is necessary to evaluate finite element
+fields at locations $\mathbf x_i$. But this requires evaluating the
+finite element shape functions at points on the refence cell 
+$\hat{\mathbf x}_i$. To make this efficient, every particle doesn't
+just store its location and the cell it is on, but also what location
+that point corresponds to reference coordinates.
+
```
gassmoeller, line 328: Maybe you could add a sentence here that describes that there are convenience functions for creating more complex particle patterns available in the `Particles::Generators` namespace, which will be illustrated in the other particle tutorials?

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,728 @@
...
+
+        if (particle_handler.n_particles_in_cell(cell) > 0)
+          {
+            std::vector<Point<dim>>     particle_positions;
+
+            const typename Particles::ParticleHandler<
+            dim>::particle_iterator_range particles_in_cell =
+                particle_handler.particles_in_cell(cell);
+
+            const unsigned int n_particles_in_cell =
```
gassmoeller, line 291: See my comment further below in the `move_particles` function. Usually I would not worry about optimizations for a simple tutorial, but since this code will likely be copied into other codes we can optimize this the same way I mentioned below. This would also allow to create `particle_positions` immediately with the correct size and save another line).

bangerth:
If you make changes, stash them, blow away your branch, and pull again -- I just force pushed a rebase version of the branch that you probably just pulled...

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,728 @@
...
+                    // ParticleHandler
+
+                    Particles::Particle<dim> new_particle;
+                    new_particle.set_location(location);
+                    new_particle.set_reference_location(
+                      mapping.transform_real_to_unit_cell(cell, location));
+                    new_particle.set_id(n_current_particles);
+                    particle_handler.insert_particle(new_particle, cell);
+
+                    ++n_current_particles;
```
gassmoeller, line 414: This variable is currently unused. Is it useful?

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,728 @@
...
+    for (const auto &cell : dof_handler.active_cell_iterators())
+      if (particle_handler.n_particles_in_cell(cell) > 0)
+        {
+          const typename Particles::ParticleHandler<
+            dim>::particle_iterator_range particles_in_cell =
+            particle_handler.particles_in_cell(cell);
+
+          const unsigned int n_particles_in_cell =
+            particle_handler.n_particles_in_cell(cell);
+
```
gassmoeller, line 444: I do not think we should make too many optimizations for a simple step, but since this code will likely be copied into many programs as the canonical way to do things, can we move this call outside of the `if(n_particles_in_cell > 0)`? It is of `O(NlogN)` complexity which may matter for millions of particles. In fact the optimal way of writing these lines would probably be:
```suggestion
    for (const auto &cell : dof_handler.active_cell_iterators())
    {
     const typename Particles::ParticleHandler<
            dim>::particle_iterator_range particles_in_cell =
            particle_handler.particles_in_cell(cell);

          const unsigned int n_particles_in_cell =
            std::distance(particles_in_cell.begin(),particles_in_cell.end());
      if (n_particles_in_cell > 0)
        {
         

```

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,728 @@
...
+                    std::max(max_particle_velocity, velocity.norm());
+                }
+
+              if (max_particle_velocity > 0)
+                min_cell_size_over_velocity =
+                  std::min(min_cell_size_over_velocity,
+                           cell_size / max_particle_velocity);
+            }
+
+        constexpr double c_saftey = 0.5;
```
gassmoeller, line 540:
```suggestion
        constexpr double c_safety = 0.5;
```

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,728 @@
...
+        time.set_desired_next_step_size(c_saftey * 0.5 *
```
gassmoeller, line 541:
```suggestion
        time.set_desired_next_step_size(c_safety * 0.5 *
```

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,728 @@
...
+                                        min_cell_size_over_velocity);
+      }
+    else
+      {
+        QTrapez<dim>  vertex_quadrature;
+        FEValues<dim> fe_values(fe, vertex_quadrature, update_gradients);
```
gassmoeller, line 547: Does it matter to add the mapping here? Just to prevent possible accidents if this is copied into other codes.

gassmoeller:
> If you make changes, stash them, blow away your branch, and pull again -- I just force pushed a rebase version of the branch that you probably just pulled...

Luckily I did a review and did not work on the branch :-)

examples/step-19/step-19.cc
bangerth: I use this to assign ids three lines above. But the logic is wrong once particles get lost. Will fix.

examples/step-19/step-19.cc
bangerth: I'm happy to do this, but shouldn't the underlying call to `equal_range` just be `O(2*log(N))`? That's one `log N` for the lower-bound for the key, and another one for the upper-bound?

Or do you mean that we do this `M` times (for `M` = number of cells) that that's how you get to your complexity?

examples/step-19/step-19.cc
bangerth: Get it wrong once and autocompletion makes it wrong every time :-)

examples/step-19/step-19.cc
bangerth: I don't have a mapping in this program. There are other places where we would have to do the same -- e.g., interpolate_boundary_values. Let's keep it simple here.

examples/step-19/step-19.cc
gassmoeller: I meant that at the moment there are 3 calls that are executed `n_cells` times with `O(log(N))` each (2 calls to `n_particles_in_cell` and one to `particles_in_cell`), while in practice only one call is necessary, namely the one to `particles_in_cell`. `n_particles_in_cell` can then be computed with `std::distance` (linear in distance). Changing to my suggested code would not change the overall complexity (`O(NlogN)`), but would reduce the number of calls with that complexity.

dangars:
I reviewed the physics part and the introduction of the tutorial. It reads very well!

It is a very nice tutorial.

I do not have any comment. I will review it again when it is close to be finished.

examples/step-19/step-19.cc
bangerth: Good point. I've taken this as a teaching opportunity and instead just doing as you say, pointed out that it's inefficient and wrote a section in the "Possibilities for extensions" about it, and in particular how to reduce the effort to O(N).

bangerth:
This is now ready for review. Experience from previous tutorial programs is that it is easiest if reviewers just make the changes directly in the source files unless it's just a relatively small number of comments.

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,449 @@
...
+much in the same spirit as we do in step-21, step-31, and step-32,
+to name just a few, and can all be understood in the context of
+the operator splitting methods discussed in step-58.
+
+So, if we denote by an upper index $n$ the time step, and if we
+use a simple time discretization for the ODE, then this means
+that we have to solve the following set of equations in each time
+step:
+@f{align*}{
+  -\epsilon_0 \Delta V^(n) &= \sum_i e\delta(\mathbf x-\mathbf x_i^{(n-1)})
```
peterrum, line 178:
```suggestion
  -\epsilon_0 \Delta V^{(n)} &= \sum_i e\delta(\mathbf x-\mathbf x_i^{(n-1)})
```

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,449 @@
...
+Here, $h_i$ is the length of the shortest edge of the cell on which particle
+$i$ is located -- in essence, a measure of the size of a cell.
+
+On the other hand, a particle might already be at the boundary of one cell
+and the neighboring cell might be once further refined. So then the time to
+cross that *neighboring* cell would actually be half the amount above,
+suggesting
+@f[
+  \Delta t \le \min_i \frac{\tfrac 12 h_i}{\|\mathbf v_i\|}.
+@f]
```
peterrum, line 219: I would mention CFL here (because later you state that we already have discussed CFL).

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,449 @@
...
+experience is along these field lines. Finally, the picture shows the
+mesh used in the computation, illustrating that there are
+singularities at the tip of the re-rentrant corner as well
+as at all places where the boundary conditions change; these
+singularities are visible because the mesh is refined in these
+locations.
+
+Of practical interest is to figure out which fraction of the
+electrons emitted from the cathode actually make it through the
+hole in the anode -- electrons that just bounce into the enode
```
peterrum, line 435:
```suggestion
hole in the anode -- electrons that just bounce into the anode
```

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,1112 @@
...
+
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/numerics/data_out.h>
+#include <deal.II/numerics/vector_tools.h>
+#include <deal.II/numerics/error_estimator.h>
+
+
+// The ones that are new are only the following three: The first declared the
```
peterrum, line 51:
```suggestion
// The ones that are new are only the following three: The first declares the
```

examples/step-19/doc/intro.dox
bangerth: It's actually not a CFL condition, just a time step restriction. I opted to simply change the place where I mentioned "CFL" before, rather than doing it here.

bangerth:
I have incorporated @peterrum 's comments (thanks, by the way!).

Anyone else?

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,449 @@
...
+
+The following picture shows the geometry that we're going to use:
+
+<p align="center">
+  <img src="https://www.dealii.org/images/steps/developer/step-19.geometry.png"
+       alt="The geometry used in this program"
+       width="600">
+</p>
+
+In this picture, that parts of the boundary marked in red and blue are the
```
gassmoeller, line 400:
```suggestion
In this picture, the parts of the boundary marked in red and blue are the
```

gassmoeller: or: 
```suggestion
In this picture, those parts of the boundary marked in red and blue are the
```

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,449 @@
...
+cathode, held at an electric potential $V=-V_0$. The part of the cathode shown
+in red is that part that is heated, leading to electrons leaving the metal
```
gassmoeller, line 402:
```suggestion
in red is the part that is heated, leading to electrons leaving the metal
```

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,449 @@
...
+and then being accelerated by the electric field (a few electric
+field lines are also shown). The green part of the boundary is the anode,
+held at $V=+V_0$. The rest of the boundary satisfies a Neumann boundary
+condition.
+
+This set up mimicks real devices. The re-entrant corner results in an
```
gassmoeller, line 408: Maybe:
```suggestion
This setup mimicks real devices. The re-entrant corner results in an
```

examples/step-19/doc/intro.dox
bangerth: I'm putting this into a comment at some point below. Thanks for the suggestion.

bangerth:
Thanks, @gassmoeller ! I've now also incorporated all of his comments.

peterrum:
We need reviewer(s) here (step-19) and in PR #10308 (step-68). What about if @blaisb reviews this tutorial and @bangerth takes a look at PR #10308 and then we should be fine to merge both!

blaisb:
> We need reviewer(s) here (step-19) and in PR #10308 (step-68). What about if @blaisb reviews this tutorial and @bangerth takes a look at PR #10308 and then we should be fine to merge both!

I would be glad to help. I am going for a one week vacation but I can gladly do it in the week of july the 27th. :)! Obviously, Wolfgang's comment on step-68 would be more than welcomed and highly useful.

bangerth:
I'll do my best but it will have to wait till the week after next (I'm going to be mostly offline next week).

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,459 @@
...
+<h1>Introduction</h1>
+
+The finite element method in general, and deal.II in particular, were invented
+to solve partial differential equations -- in other words, to solve
+[continuum mechanics](https://en.wikipedia.org/wiki/Continuum_mechanics) problems.
+On the other hand, sometimes one wants to solve problems in which it is useful
+to track individual objects ("particles") and how their positions evolve. If
+this simply leads to a set of ordinary differential equations, for example
+if you want to track the positions of the planets in the solar system over
+time, then deal.II is clearly not your right tool. On the other hand, if
```
blaisb, line 26: This would really depend on the type of "particle algorithm" you want to implement. A lot of these particle methods (SPH, etc.) use a background grid for neighbour detection (either a tree structure or a flat structure). Consequently, they fall quite nicely in the deal.II framework (especially with the notion of ghost particles). Anyway, just ranting, but deal.II really has nice support even for pure particles method. We use it for our DEM code for example which is purely Lagrangian particle tracking and neighbour detection.

examples/step-19/doc/intro.dox
```diff
@@ -0,0 +1,459 @@
...
+The deal.II class we will use to deal with particles, Particles::ParticleHandler,
+stores particles in a way so that the position $\mathbf x_i$ is part of the
+Particles::ParticleHandler data structures. (It stores particles sorted
+by cell they are in, and consequently needs to know where each particle is.)
+The velocity $\mathbf v_i$, on the other hand, is of no concern to
+Particles::ParticleHandler and consequently we will store it as a
+"property" of each particle that we will update in each time step. Properties
+can also be used to store any other quantity we might care about each particle:
+its charge, or if they were larger than just an electron, its color, mass,
+attitude in space, chemical composition, etc.
```
blaisb, line 87:
```suggestion
altitude in space, chemical composition, etc.
```

examples/step-19/doc/kind
```diff
@@ -0,0 +1 @@
+techniques
```
blaisb, line 1: Do you think it might be a good idea to consider adding a "particles" kind?

examples/step-19/doc/results.dox
```diff
@@ -0,0 +1,290 @@
...
+the electric field changes in response to their motion:
+
+@htmlonly
+<p align="center">
+  <iframe width="560" height="315" src="https://www.youtube.com/embed/HwUtE7xuteE"
+   frameborder="0"
+   allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
+   allowfullscreen></iframe>
+ </p>
+@endhtmlonly
```
blaisb, line 130: Would it be possible to add the time to the movie using the annotate time filter of Paraview? If you output a pvd file + use this filter it could help the discussion I believe (just my opinion though).

examples/step-19/doc/results.dox
```diff
@@ -0,0 +1,290 @@
...
+are
+@f{align*}{
+  \frac{{\mathbf v}_i^{(n)}-{\mathbf v}_i^{(n-1)}}{\Delta t} &= \frac{e\nabla V^{(n)}}{m}
+  \\
+  \frac{{\mathbf x}_i^{(n)}-{\mathbf x}_i^{(n-1)}}{\Delta t} &= {\mathbf v}_i^{(n)}.
+@f}
+This corresponds to a simple forward-Euler time discretization -- a method of
+first order accuracy in the time step size $\Delta t$ that we know we should
+avoid because we can do better. Rather, one might want to consider a scheme such
+as the leapfrog scheme, or even better a Runge-Kutta integrator.
```
blaisb, line 273: Tiny comments. For this type of problem (where the force/acceleration is only a function of the position), classical Verlet integrator (such as velocity-verlet or leapfrog)  are symplectic (https://en.wikipedia.org/wiki/Symplectic_integrator). This gives amazing properties, notably in terms of preserving the energy of the particles.  Consequently, they are quite desirable even if they are of lower order (hence their extensive use in molecular dynamics simulations). I'm not sure I would prefer classical RK over a velocity-verlet scheme here.

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,1112 @@
...
+// for the various parts of the geometry, as well as the physical properties
+// of electrons and other specifics of the setup we use here.
+namespace Step19
+{
+  namespace BoundaryIds
+  {
+    constexpr types::boundary_id open          = 101;
+    constexpr types::boundary_id cathode       = 102;
+    constexpr types::boundary_id focus_element = 103;
+    constexpr types::boundary_id anode         = 104;
```
blaisb, line 84: Just out of curiosity, why 101, 102, 103 and 103 and not 1,2,3,4?

examples/step-19/step-19.cc
```diff
@@ -0,0 +1,1112 @@
...
+  // but the particle handler stores the reference and will use it once
+  // particles are added -- which happens after the triangulation is built).
+  // The other piece of information it gets is how many "properties"
+  // each particle needs to store. Here, all we need each particle to
+  // remember is its current velocity, i.e., a vector with `dim`
+  // components. There are, however, other intrinsic properties that
+  // each particle has and that the Particles::ParticleHandler class
+  // automatically and always makes sure are available; in particular,
+  // these are the current location of a particle, the cell it is on,
+  // and a particle's ID.
```
blaisb, line 175:
```suggestion
  // automatically and always makes sure are available; in particular,
  // these are the current location of a particle, the cell it is on, it's reference location within that cell
  // and a particle's ID.
```

peterrum (approved):
@bangerth I think, once you have addressed @blaisb's comments, this PR is good to go!

examples/step-19/doc/intro.dox
bangerth: Good point -- I added a link.

examples/step-19/doc/intro.dox
bangerth: No, I actually did mean "attitude: A vehicle's orientation about its center of gravity, known as its attitude in flight dynamics" :-)

examples/step-19/doc/kind
bangerth: Eventually maybe. The problem is that you end up with too many colors in the tutorial graph, and that things are no longer distinguishable. I'd vote for leaving it as is for now.

examples/step-19/doc/results.dox
bangerth: I played with this. It just looks so bulky. From an artistic perspective, I found that really unappealing :-(

examples/step-19/doc/results.dox
bangerth: Most excellent idea!

examples/step-19/step-19.cc
bangerth: I think the principle is to use numbers that are uncommon. If there are pre-defined boundary indicators previously generated by the `GridGenerator` functions, they will likely be small integers starting from zero, but not in this rather randomly chosen range. It just avoids the possibility for conflicts, and also reduces the temptation to just spell these numbers out (because you will probably never remember which is which, whereas you might have been tempted if they had started at 0).

It's a good enough question that I copied something like the text above into the program itself.

bangerth:
I've addressed the comments by @blaisb (thanks a lot!). Based on his and @peterrum 's comments, I believe that this is ready to merge.

bangerth:
Anyone want to pull the trigger and merge?

bangerth:
Yay!

PR #10306, opened 2020-05-22
Rename namespace internals:: to internal::AffineConstraints::.

bangerth:
The AffineConstraints class and friends put things into a namespace 'internals'. That's
an uncommon name: We typically use 'internal::'. But a lot of the things in that namespace
also have rather generic names, so do as we often do in such cases: further qualify
the name to indicate what class the internals correspond to.

Almost all of the changes you see here are whitespace changes, so look at the patch with `?w=1`. There are a couple of places where we now need to qualify an `AffineConstraints` class name when used from inside another `internal::` namespace.

/rebuild

### DIFF
--- a/include/deal.II/lac/affine_constraints.templates.h
+++ b/include/deal.II/lac/affine_constraints.templates.h
@@ -2325,1147 +2325,1161 @@ AffineConstraints<number>::distribute(VectorType &vec) const
 }
 
 // Some helper definitions for the local_to_global functions.
-namespace internals
+namespace internal
 {
-  using size_type = types::global_dof_index;
-
-  // this struct contains all the information we need to store about each of
-  // the global entries (global_row): are they obtained directly by some local
-  // entry (local_row) or some constraints (constraint_position). This is not
-  // directly used in the user code, but accessed via the GlobalRowsFromLocal.
-  //
-  // The actions performed here correspond to reshaping the constraint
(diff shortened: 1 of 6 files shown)


include/deal.II/lac/affine_constraints.templates.h
```diff
@@ -2325,1147 +2325,1161 @@ AffineConstraints<number>::distribute(VectorType &vec) const
...
+            Distributing(block_object.get_row_indices().block_start(i)));
+          block_starts[i] = first_block - global_rows.total_row_indices.begin();
+          block_indices   = first_block;
+        }
+      block_starts[num_blocks] = n_active_rows;
+
+      // transform row indices to block-local index space
+      for (size_type i = block_starts[1]; i < n_active_rows; ++i)
+        global_rows.global_row(i) =
+          block_object.get_row_indices()
```
drwells, line 2964: Are we not indenting these files with clang-format? The old version looks wrong.

drwells: Oh, I see - GitHub renders the rest weirdly when we hide whitespace changes. Nevermind!

bangerth:
We do this all the time of course. But now, they don't, I think, because their `userproject::internal` namespace is not part of `dealii::`.

kronbichler:
I guess we just go ahead and see.

bangerth:
Part of #10284.

PR #10309, opened 2020-05-22
Bugfix: hp::Refinement::choose_p_over_h() in parallel

marcfehling:
In this function we ask for coarsening flags of siblings if a cell is flagged for coarsening. Siblings might not live on the same processor though, which I have always assumed naively.

There might be more elegant ways to solve this other than using `GridTools::exchange_cell_data_to_ghosts`. Since this function is not critical for performance I didn't bother to spend more time on it.

Part of `Hackathon 2020`.

### DIFF
--- a/source/hp/refinement.cc
+++ b/source/hp/refinement.cc
@@ -23,6 +23,7 @@
 #include <deal.II/distributed/tria_base.h>
 
 #include <deal.II/grid/grid_refinement.h>
+#include <deal.II/grid/grid_tools.h>
 
 #include <deal.II/hp/dof_handler.h>
 #include <deal.II/hp/refinement.h>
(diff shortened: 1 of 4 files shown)


drwells:
Interesting - two tests failed with the serial configuration:
	2785 - hp/step-27.debug (Failed)
	2970 - hp/hp_coarsening.debug (Failed)

source/hp/refinement.cc
```diff
@@ -635,6 +636,28 @@ namespace hp
...
+      auto unpack =
+        [&ghost_buffer](const typename dealii::hp::DoFHandler<dim, spacedim>::
+                          active_cell_iterator &    cell,
+                        const std::pair<bool, bool> pair) -> void {
+        ghost_buffer.emplace(cell->id(), pair);
+      };
+
+      GridTools::exchange_cell_data_to_ghosts<
+        std::pair<bool, bool>,
+        dealii::hp::DoFHandler<dim, spacedim>>(dof_handler, pack, unpack);
```
bangerth, line 658: This is just so elegant!

bangerth:
There are two failing tests:
```
	2785 - hp/step-27.debug (Failed)
	2970 - hp/hp_coarsening.debug (Failed)
```
Can you investigate?

marcfehling:
> Interesting - two tests failed with the serial configuration:
> 2785 - hp/step-27.debug (Failed)
> 2970 - hp/hp_coarsening.debug (Failed)

> There are two failing tests:
> 
> ```
> 	2785 - hp/step-27.debug (Failed)
> 	2970 - hp/hp_coarsening.debug (Failed)
> ```
> 
> Can you investigate?

Guess is just need to check whether it's a `parallel::Triangulation` we work with...
```
Additional information: 
2806:     The function exchange_cell_data_to_ghosts() only works with parallel triangulations.
```

source/hp/refinement.cc
marcfehling: I agree. This way to communicate data is indeed elegant.

However, in this particular scenario, we certainly communicate more than necessary. But since this is not critical for performance, I'll leave it like this. Maybe my use of the expression 'elegant' in the pull request message was not the proper one - I should've used 'efficient' in this context. :)

marcfehling:
Made the necessary changes. Let's wait for the CI to finish.

PR #10313, opened 2020-05-22
Move a class out of a surrounding class.

bangerth:
Part of #10284. I'll have to get that `ScratchData` object into the `AffineConstraints` class. This is step 1, moving it out of the surrounding class.

/rebuild


bangerth:
Looks like the one failing test is unrelated.

PR #10319, opened 2020-05-22
Add a version of ParticleAccessor::set_properties().

bangerth:
I'm going to use this in #10301.

/rebuild


bangerth:
The underlying problem here is that `Particles::Particle` already has this function, but we typically access these objects via `Particles::ParticleAccesor` which could in principle be just a pass-through entity, but is not for all functions offered by the former class.

@gassmoeller -- FYI.

PR #10320, opened 2020-05-22
Update ParticleHandler::n_particles_in_cell().

bangerth:
I moved that function next to the one where we ask for the particles in the current cell. I had missed that that function actually exists because it wasn't where I had hoped it was.

While there also slightly update a couple of other places.

/rebuild

### DIFF
--- a/source/particles/particle_handler.cc
+++ b/source/particles/particle_handler.cc
@@ -297,6 +297,30 @@ namespace Particles
 
 
   template <int dim, int spacedim>
+  unsigned int
+  ParticleHandler<dim, spacedim>::n_particles_in_cell(
+    const typename Triangulation<dim, spacedim>::active_cell_iterator &cell)
+    const
+  {
+    const internal::LevelInd found_cell =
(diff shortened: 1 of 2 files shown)


source/particles/particle_handler.cc
```diff
@@ -323,11 +347,20 @@ namespace Particles
...
+        const auto particles_in_cell = particles.equal_range(level_index);
+        return boost::make_iterator_range(
+          particle_iterator(particles, particles_in_cell.first),
+          particle_iterator(particles, particles_in_cell.second));
+      }
+    else
+      AssertThrow(false,
+                  ExcMessage("You can't ask for the particles on an artificial "
+                             "cell since we don't know what exists on these "
+                             "kinds of cells."));
```
kronbichler, line 361: This makes the test `particles/particle_handler_06` fail,
https://cdash.43-1.org/testDetails.php?test=45322622&build=6728
@bangerth can you have a look?

bangerth: Yes, this is already #10351 .

PR #10329, opened 2020-05-23
Remove template arguments from internal data structures of Triangulation

peterrum:
This PR removes the template arguments from the internal data structures of Triangulation. This allows to remove all the template class specializations, which made the code very hard to read and to understand since Doxygen also had its problem with them.

The proposed structure looks more or less like this (`TriaObject` is only used during setup, but the indices save now in a big vector):

uml-0

Maybe we could add - at some time - this UML diagram somewhere in the documentation. For me it really helped to have all (relevant) triangulation-related data structures in one picture to understand and demystify the relations! As a side-product, the PR solves some very old TODOs.

While it used to look like this:
- **1D (template specialization!)**: 

uml-1

- **2D**:

uml-2

- **3D**:

uml-3

**Note**: These changes are only internal and is not visible to user code. Also performance should not be influenced since the classed accessing these data structures are still templated.


include/deal.II/grid/tria_accessor.templates.h
```diff
@@ -366,15 +355,14 @@ namespace internal
...
+inline dealii::internal::TriangulationImplementation::TriaObjects &
 TriaAccessorBase<structdim, dim, spacedim>::objects() const
 {
   if (structdim != dim)
     // get sub-objects. note that the
     // current class is only used for
     // objects that are *not* cells
-    return *dealii::internal::TriaAccessorBaseImplementation::get_objects(
+    return *dealii::internal::TriaAccessorBaseImplementation::get_objects<dim>(
       this->tria->faces.get(), std::integral_constant<int, structdim>());
```
kronbichler, line 366: Is this complicated function call setup still needed? Given that we do not need to distinguish types via their template argument any more, can't we just switch to faces/lines as appropriate?

include/deal.II/grid/tria_faces.h
```diff
@@ -29,7 +29,7 @@ namespace internal
   namespace TriangulationImplementation
   {
     /**
-     * General template for information belonging to the faces of a
+     * This calls contains information belonging to the faces of a
```
kronbichler, line 32:
```suggestion
     * This class contains information belonging to the faces of a
```

include/deal.II/grid/tria_levels.h
```diff
@@ -54,10 +54,21 @@ namespace internal
...
+        : dim(dim)
+        , cells(dim)
+      {}
+
+      TriaLevel()
+        : dim(-1)
+        , cells(-1)
+      {}
+
+      unsigned int dim;
```
kronbichler, line 70: Can you document this member variable and the two constructors above?

include/deal.II/grid/tria_levels.h
```diff
@@ -159,8 +170,47 @@ namespace internal
...
+       * cases, and we have code that makes sure that if a mesh allows
+       * this to happen, that we have this convention. We also know
+       * that it is always possible to have opposite faces have
+       * parallel normal vectors. (For both things, see the paper by
+       * Agelek, Anderson, Bangerth, Barth in the ACM Transactions on
+       * Mathematical Software mentioned in the documentation of the
+       * GridReordering class.)
+       *
+       * The problem is that we originally had another condition, namely that
+       * faces 0, 2 and 6 have normals that point into the cell, while the
```
kronbichler, line 187:
```suggestion
       * faces 0, 2 and 4 have normals that point into the cell, while the
```
?

include/deal.II/grid/tria_object.h
```diff
@@ -104,14 +105,47 @@ namespace internal
...
        * Global indices of the face iterators bounding this cell if dim@>1,
        * and the two vertex indices in 1d.
        */
       int faces[GeometryInfo<structdim>::faces_per_cell];
     };
 
+    class TriaObjectView
+    {
+    public:
+      TriaObjectView(const ArrayView<int> faces)
```
kronbichler, line 119:
```suggestion
      TriaObjectView(const ArrayView<int> &faces)
```

include/deal.II/grid/tria_objects.h
```diff
@@ -74,11 +62,56 @@ namespace internal
        */
       TriaObjects();
 
+      /**
+       * Constructor resetting some data.
```
kronbichler, line 66:
```suggestion
       * Constructor for a specific dimension.
```
Please also adjust the comment to the line above.

include/deal.II/grid/tria_accessor.templates.h
peterrum: Indeed:

```cpp
template <int structdim, int dim, int spacedim>
inline dealii::internal::TriangulationImplementation::TriaObjects &
TriaAccessorBase<structdim, dim, spacedim>::objects() const
{
  if (structdim == dim)
    return this->tria->levels[this->present_level]->cells;


  if (structdim == 1 && dim > 1)
    return this->tria->faces->lines;


  if (structdim == 2 && dim > 2)
    return this->tria->faces->quads;


  Assert(false, ExcInternalError());


  return *static_cast<
    dealii::internal::TriangulationImplementation::TriaObjects *>(nullptr);
}
```

Now, the code is readable!

peterrum:
@kronbichler I have addressed your comments!

peterrum:
I have resolved the merge conflict. However, I get the following run-timer errer:
```
4937: terminate called after throwing an instance of 'boost::archive::archive_exception'
4937:   what():  unregistered class
```

I guess it is somehow related `TriaLevel`.  But, I don't know how to resolve this! I am also wondering why `tidy` finds issues in places, which I have not touched...

include/deal.II/grid/tria.h
```diff
@@ -78,12 +79,10 @@ namespace internal
 {
   namespace TriangulationImplementation
   {
-    template <int dim>
     class TriaLevel;
```
bangerth, line 82: Since you `#include` the header file above, you don't need these forward declarations any more.

include/deal.II/grid/tria_accessor.templates.h
```diff
@@ -833,12 +737,8 @@ namespace internal
       line_orientation(const TriaAccessor<2, 3, spacedim> &accessor,
                        const unsigned int                  line)
       {
-        // quads as part of 3d hexes can have non-standard orientation
-
-        // TODO: why is this face_orientation, not line_orientation as in the
-        // setter function?
-        return accessor.tria->faces->quads.face_orientation(
-          accessor.present_index, line);
+        return accessor.tria->faces->quads_line_orientations
```
bangerth, line 740: Did you figure out the TODO?

include/deal.II/grid/tria_faces.h
```diff
@@ -40,100 +40,57 @@ namespace internal
...
     public:
       /**
-       * Constructor. This constructor is deleted to prevent the use of this
-       * template, as only the specializations should be used
+       * Constructor.
        */
-      TriaFaces() = delete;
-    };
-
+      TriaFaces(unsigned int dim);
```
bangerth, line 52:
```suggestion
      TriaFaces(const unsigned int dim);
```

include/deal.II/grid/tria_faces.h
```diff
@@ -40,100 +40,57 @@ namespace internal
...
-     */
-    template <>
-    class TriaFaces<2>
-    {
-    public:
-      /**
-       * The TriaObject containing the data of lines.
-       */
-      TriaObjects<TriaObject<1>> lines;
+      unsigned int dim;
```
bangerth, line 62: I think you should be able to make this `const`, no?

include/deal.II/grid/tria_faces.h
```diff
@@ -40,100 +40,57 @@ namespace internal
...
       /**
-       * The TriaObject containing the data of lines.
+       * Reserve space for line_orientations.
+       *
+       * @note Used only for dim=3.
        */
-      TriaObjects<TriaObject<1>> lines;
+      void
+      reserve_space(const unsigned int new_quads_in_pairs,
+                    const unsigned int new_quads_single = 0);
```
bangerth, line 92: Can you document what the two arguments do?

include/deal.II/grid/tria_levels.h
```diff
@@ -54,10 +54,32 @@ namespace internal
...
+       */
+      TriaLevel()
+        : dim(numbers::invalid_unsigned_int)
+        , cells(numbers::invalid_unsigned_int)
+      {}
+
+      /**
+       * Dimension of the Triangulation.
+       */
+      unsigned int dim;
```
bangerth, line 81: Is it possible to make this variable `const`?

include/deal.II/grid/tria_object.h
```diff
@@ -104,14 +105,47 @@ namespace internal
...
-    protected:
+    public:
       /**
        * Global indices of the face iterators bounding this cell if dim@>1,
        * and the two vertex indices in 1d.
        */
       int faces[GeometryInfo<structdim>::faces_per_cell];
     };
 
+    class TriaObjectView
```
bangerth, line 116: Please document this class and its intent, as well as its members.

include/deal.II/grid/tria_objects.h
```diff
@@ -74,11 +62,55 @@ namespace internal
        */
       TriaObjects();
 
+      /**
+       * Constructor for a specific dimension.
+       */
+      TriaObjects(unsigned int structdim);
+
+    private:
+      unsigned int structdim;
```
bangerth, line 71: can you make it `const`?

include/deal.II/grid/tria_objects.h
```diff
@@ -74,11 +62,55 @@ namespace internal
...
       /**
        * Vector of the objects belonging to this level. The index of the
        * object equals the index in this container.
        */
-      std::vector<G> cells;
+      std::vector<int> cells;
+
+    public:
+      unsigned int
+      n_objects() const
```
bangerth, line 81: can you document this and the following functions?

include/deal.II/grid/tria.h
peterrum: I will don. I still have problems serializing/deserialzing `TriaLevel` (annoying Boost exception) so I did add it back here.

include/deal.II/grid/tria_faces.h
peterrum: No, it't not possible due to serialization/deserialization of Boost.

include/deal.II/grid/tria_faces.h
peterrum: I would if I knew what they  mean in `internal::TriangulationImplementation::TriaObjects::reserve_space()` ;)

peterrum:
@bangerth I'll add the missing documentation later!

include/deal.II/grid/tria_faces.h
peterrum: My suggestion is to postpone this issue. I will dig into `Triangulation` and the internal data structures more deeply in the next weeks. The internal functions might change and/or I might understand the comments then!

peterrum:
@bangerth I have added some comments!

include/deal.II/grid/tria_faces.h
bangerth: In the context of adding children, we used to store all `2^d` children of one cell consecutively and each cell only stored a pointer to the first of its children. But then along came anisotropic refinement where we may have only 2 children, so we now store them in blocks of 2 and store `2^(d-1)` pointers to blocks of 2 children.

I would not be surprised if this is a similar thing: Where we typically store two new faces consecutively, but then there are occasions where we need a single one. That's sort of my educated guess from the context and the names, but I really have no idea whether that's true.

bangerth:
There are 3 failing tests:
```
	569 - bits/refine_and_coarsen_for_active_cell_index_05.debug (Failed)
	3091 - serialization/triangulation_02.debug (Failed)
	3094 - serialization/triangulation_01.debug (Failed)
```
The latter two make sense. Do you know what's happening with the first one?

peterrum:
@bangerth All three are boost related. With the same error message.

peterrum:
All tests passed locally!

PR #10334, opened 2020-05-23
Split up MatrixFree::initialize_indices to speed up compilation

kronbichler:
This PR breaks out two big parts of the humongous function `MatrixFree::initialize_indices()` into separate functions. As the individual parts do not need the vectorized array template argument, they only need to be compiled once (for index manipulations of `DoFInfo`, which really belongs there) or `dim` times for `DoFHandler`-related parts. I admit that moving around 1000 lines of code with different indentation levels creates some noise. On the other hand, the two new functions document much better what happens and they happen at the right spot where they belong to.

I checked all matrix-free tests and they are passing, so I am confident this PR is correct.

This PR is in merge conflict with #10328, so I suggest to merge the other PR first so I can adjust here, unless that PR takes too long to get into. The upside of #10328 is that it will eliminate around 20-30 lines in this PR, because I do not need to distinguish between the `hp_dof_handler` and `dof_handler` any more.

Related to #9790.


kronbichler:
I checked compilation of the now single `matrix_free.cc` file and it is around 60 seconds - similar to many other files. `mapping_info.cc` takes almost 80 seconds for the release build, so this seems to be fine.

include/deal.II/matrix_free/face_setup_internal.h
```diff
@@ -164,15 +164,14 @@ namespace internal
...
 
     template <int dim>
-    template <typename MFAddData>
     void
     FaceSetup<dim>::initialize(
-      const dealii::Triangulation<dim> &                  triangulation,
-      const MFAddData &                                   additional_data,
+      const dealii::Triangulation<dim> &triangulation,
+      const unsigned int                mg_level,
+      const bool                        hold_all_faces_to_owned_cells,
```
peterrum, line 171: That's good :) That `AdditionalData` is part of `MatrixFree` class is kind of annoying due to the template arguments...

kronbichler:
I rebased and cleaned up some of the hp infrastructure. We could also remove the template on the outer `reinit` functions, but that is only a minor thing and I would postpone it for now.

kronbichler:
With #10328 in place, this PR now cuts down code by 250 lines, which is a nice cleanup.

PR #10335, opened 2020-05-23
consistently use local_size().

drwells:
I first thought that this change would be uncontroversial but in the process of writing a test I learned that `TrilinosWrappers::MPI::local_size()` behaves differently from the other two ghosted vectors - it returns the total number of entries (both ghosted and locally owned) while the other two return just the number of locally owned elements.

The documentation of this class clearly states that this behavior is intentional. My proposal: lets deprecate `local_size()` everywhere and implement `locally_owned_size()` in all vector classes. Thoughts?

### DIFF
--- a/include/deal.II/lac/la_parallel_vector.h
+++ b/include/deal.II/lac/la_parallel_vector.h
@@ -893,7 +893,7 @@ namespace LinearAlgebra
        * the C++ standard library by returning iterators to the start and end
        * of the <i>locally owned</i> elements of this vector.
        *
-       * It holds that end() - begin() == local_size().
+       * It holds that end() - begin() == locally_owned_size().
        *
        * @note For the CUDA memory space, the iterator points to memory on the
        * device.
@@ -975,8 +975,8 @@ namespace LinearAlgebra
       /**
        * Read access to the data field specified by @p local_index. Locally
        * owned indices can be accessed with indices
-       * <code>[0,local_size)</code>, and ghost indices with indices
-       * <code>[local_size,local_size+ n_ghost_entries]</code>.
+       * <code>[0,locally_owned_size)</code>, and ghost indices with indices
+       * <code>[locally_owned_size,locally_owned_size+ n_ghost_entries]</code>.
        *
        * Performance: Direct array access (fast).
        */
@@ -986,8 +986,8 @@ namespace LinearAlgebra
       /**
        * Read and write access to the data field specified by @p local_index.
        * Locally owned indices can be accessed with indices
-       * <code>[0,local_size)</code>, and ghost indices with indices
-       * <code>[local_size,local_size+n_ghosts]</code>.
+       * <code>[0,locally_owned_size())</code>, and ghost indices with indices
+       * <code>[locally_owned_size(), locally_owned_size()+n_ghosts]</code>.
        *
        * Performance: Direct array access (fast).
        */
--- a/include/deal.II/lac/la_parallel_vector.h
+++ b/include/deal.II/lac/la_parallel_vector.h
@@ -1602,7 +1602,7 @@ namespace LinearAlgebra
                        partitioner->locally_owned_size() +
                          partitioner->n_ghost_indices());
       // do not allow reading a vector which is not in ghost mode
-      Assert(local_index < local_size() || vector_is_ghosted == true,
+      Assert(local_index < locally_owned_size() || vector_is_ghosted == true,
              ExcMessage("You tried to read a ghost element of this vector, "
                         "but it has not imported its ghost values."));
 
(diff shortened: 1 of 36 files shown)


drwells:
`Partitioner` also uses `local_size`.

drwells:
This PR turned out a little larger than I originally intended, but we now consistently use `locally_owned_size()` throughout the library.

include/deal.II/matrix_free/operators.h
```diff
@@ -1542,7 +1545,7 @@ namespace MatrixFreeOperators
               .local_element(edge_constrained_indices[j][i]) =
               edge_constrained_values[j][i].first;
           }
-        for (; c < BlockHelper::subblock(dst, j).local_size(); ++c)
```
drwells, line 1545: I am not sure if anyone tried to do it but this would not work with a `TrilinosWrappers::Vector` in the past.

include/deal.II/lac/la_parallel_vector.h
```diff
@@ -866,10 +867,20 @@ namespace LinearAlgebra
       /**
        * Return the local size of the vector, i.e., the number of indices
        * owned locally.
+       *
+       * @deprecated use locally_owned_size() instead.
```
bangerth, line 871:
```suggestion
       * @deprecated Use locally_owned_size() instead.
```

include/deal.II/lac/petsc_vector_base.h
```diff
@@ -347,17 +347,31 @@ namespace PETScWrappers
...
     size_type
     local_size() const;
 
+    /**
+     * Return the local dimension of the vector, i.e. the number of elements
+     * stored on the present MPI process. For sequential vectors, this number
+     * is the same as size(), but for parallel vectors it may be smaller.
+     *
+     * To figure out which elements exactly are stored locally, use
+     * local_range().
```
bangerth, line 363: That's not wrong, but I think it's a limitation to the PETSc vectors that they only store contiguous ranges and that what we should recommend is to query `locally_owned_elements()` instead. Maybe list both?

include/deal.II/matrix_free/operators.h
bangerth: Out of curiosity, why not?

bangerth:
You're 98% there. Good work!

include/deal.II/matrix_free/operators.h
drwells: the matrix free code assumes that local_size refers to the number of locally owned dofs, not the number of locally owned dofs + number of ghosts (which is what our trilinos vector assumes).

drwells:
Thanks! Its getting a little late - I'll deal with the remaining problems tomorrow.

include/deal.II/matrix_free/operators.h
kronbichler: At this place of the code, the vectors would need to be non-ghosted, so in principle the ranges would need to match. However, the Trilinos vector does not have a `local_element()` access function because for performance reasons we cannot afford translating MPI-global to MPI-local indices, so the code would not work with Trilinos vectors anyway. One could fix it by adding an index translation for those vectors.

kronbichler:
Like in #10369 I do not particularly like to immediately deprecate the old interfaces because this, one day to the other and without a chance to anticipate it, can generate hundreds of warnings in a medium-sized application code. Let us discuss this at the hackathon.

tjhei:
> Let us discuss this at the hackathon.

We can talk about adding a more sophisticated deprecation macro that a code can opt out of. Something like ``#define OPT_OUT_EARLY_DEPRECATION`` or something?

masterleinad:
> We can talk about adding a more sophisticated deprecation macro that a code can opt out of. Something like `#define OPT_OUT_EARLY_DEPRECATION` or something?

You can always disable deprecations via `-Wno-deprecated-declarations` or similar. Certainly, we must be allowed to introduce deprecations in `deal.II` to move forward and, of course, these deprecations happen from one day to the other but a warning is not an error (unless a user decides to treat it a such and then it's in their responsibility to deal with it).

drwells:
I'm not in love with this approach (which is why I tagged it as discussion).
 
It would be pretty easy to define `DEAL_II_EARLY_DEPRECATED` which is only conditionally expanded and promote them later.

kronbichler:
I agree that we need a way to move forward and so it happens that we change things (I have a longer post on this on slack). But every change has a price, especially when deal.II sits somewhere in the middle of a software stack and you are forced to support multiple versions of deal.II. For small interface changes that happen a few times in a code base, you can easily adjust. But a change that is scattered over many places is a pain for the application writer. Again, we want to be able to adjust and remove some things, but I do not like if the result is deprecation warnings on those *central* things the same day as there is a better variant. So I would indeed love something like `DEAL_II_EARLY_DEPRECATED` that gives the user one more release cycle without warnings (unless the user wants them) before we warn, and then another release cycle before we delete it.

I would not use that macro for all deprecations we are doing, but there are "central" interfaces where I as a developer feel obliged to first provide the alternative for some time and let users read up before I print hundreds of deprecation warnings. Of course, at some point those warnings still arrive.

drwells:
I let this sit since the workshop but ultimately I am convinced that creating an incredible number of deprecation warnings is not worth the improvement in semantics. Lets close this until we get a better solution.

PR #10341, opened 2020-05-24
CMake: Bugfix: Run symbol check (using C compiler)  without CXXFLAGS

tamiko:


drwells:
I don't understand why this is necessary. Why would `CXXFLAGS` ever get used when we call `CHECK_C_SOURCE_COMPILES`?

tamiko:
@drwells Because the `CHECK_C_SOURCE_COMPILES` macro i an ancient beast that simply uses whatever we put into `CMAKE_REQUIRED_FLAGS`.

PR #10342, opened 2020-05-24
Fix compilation with -std=c++20

tamiko:
This is a funny one:

C++20 introduces convenient new operator overloading rules. For example,
one only has to implement

  bool operator==(const Foo&, const Bar&);

in order to be able to compare (with Foo foo; Bar bar;):

  foo == bar; bar == foo; foo != bar; bar != foo;

This change comes at a price, though:

  warning: ISO C++20 considers use of overloaded operator '!=' (with operand types
  'SparsityPattern::iterator' (aka 'dealii::SparsityPatternIterators::Iterator') and
  'SparsityPatternBase::iterator' (aka 'dealii::SparsityPatternIterators::Iterator'))
  to be ambiguous despite there being a unique best viable function
  [-Wambiguous-reversed-operator]

Even worse, due to how we implement operator== and operator!= in
LinearIndexIterator we now produce an error:

  error: use of overloaded operator '!=' is ambiguous (with operand
  types 'SparsityPattern::const_iterator' (aka
  'dealii::SparsityPatternIterators::Iterator') and
  'SparsityPattern::const_iterator')

### DIFF
--- a/source/particles/particle_handler.cc
+++ b/source/particles/particle_handler.cc
@@ -295,7 +295,7 @@ namespace Particles
 
 
   template <int dim, int spacedim>
-  unsigned int
+  types::particle_index
   ParticleHandler<dim, spacedim>::n_particles_in_cell(
     const typename Triangulation<dim, spacedim>::active_cell_iterator &cell)
     const


tamiko:
@drwells *ping* Comments?

marked as WIP for now until I have converted the remaining comparison operators.

drwells:
Does this mean that this class doesn't compile in C++20? That's surprising.

I'll take a closer look a little later today.

bangerth:
Got a link to the ISO paper that introduced this change in C++20?

tamiko:
@drwells Yes. Because the template deduction for choices between `operator==` and `operator!=` has now become ambiguous.

tamiko:
@bangerth https://timsong-cpp.github.io/cppwp/n4861/draft.pdf p.318 [over.match.oper] paragraph 3 onwards (and paragraph 9).

bangerth:
Ugh. That's a rather fundamental shift. Oh well, I guess the C++ gods decided that that's the way it's going to be going forward.

The reference I was looking for (though maybe not so important any more given your find) was the working paper that introduced that change. Those are the nXXXX papers that standardization process relies on.

tamiko:
I am closing this for now and open an issue.

PR #10345, opened 2020-05-24
Fix 64-bit compilation.

bangerth:
This fixes a mistake I introduced in #10320 whereby I changed the return type
of a function in the .h file but not the .cc file. This didn't matter in 32-bit
compilations, but does in 64-bit mode.

I am surprised that this passed the testers. Didn't we run at least one check
in 64-bit mode?

### DIFF
--- a/source/particles/particle_handler.cc
+++ b/source/particles/particle_handler.cc
@@ -295,7 +295,7 @@ namespace Particles
 
 
   template <int dim, int spacedim>
-  unsigned int
+  types::particle_index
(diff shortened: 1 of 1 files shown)


kronbichler:
> Didn't we run at least one check in 64-bit mode?

My bad - we did run and it did actually trigger, https://github.com/dealii/dealii/runs/700903694 , but I was confused by the other failures related to when github was down, so I did not look closely enough.

PR #10348, opened 2020-05-24
Applied user constraints before prolongation in MGTransferPrebuilt

mathmerizing:
Currently only no_normal_flux and zero_boundary_constraints can be used with MGConstrainedDoFs in Multigrid. Since recently, users can also add custom constraints to MGConstrainedDoFs.

In this pull request, I simply apply these user_constraints after the prolongation or restriction in the Multigrid method.

User group post: https://groups.google.com/forum/#!topic/dealii/17V-XvANXGM

### DIFF
--- a/source/multigrid/mg_transfer_prebuilt.cc
+++ b/source/multigrid/mg_transfer_prebuilt.cc
@@ -86,8 +86,9 @@ MGTransferPrebuilt<VectorType>::prolongate(const unsigned int to_level,
          ExcIndexRange(to_level, 1, prolongation_matrices.size() + 1));
 
   prolongation_matrices[to_level - 1]->vmult(dst, src);
-  this->mg_constrained_dofs->get_user_constraint_matrix(to_level).distribute(
-    dst);
+  if (this->mg_constrained_dofs != nullptr)
+    this->mg_constrained_dofs->get_user_constraint_matrix(to_level).distribute(
+      dst);
 }
 
 
@@ -103,8 +104,9 @@ MGTransferPrebuilt<VectorType>::restrict_and_add(const unsigned int from_level,
   (void)from_level;
 
   prolongation_matrices[from_level - 1]->Tvmult_add(dst, src);
-  this->mg_constrained_dofs->get_user_constraint_matrix(from_level - 1)
-    .distribute(dst);
+  if (this->mg_constrained_dofs != nullptr)
+    this->mg_constrained_dofs->get_user_constraint_matrix(from_level - 1)
+      .distribute(dst);
 }
 
 
--- a/source/multigrid/mg_transfer_prebuilt.cc
+++ b/source/multigrid/mg_transfer_prebuilt.cc
@@ -85,14 +85,20 @@ MGTransferPrebuilt<VectorType>::prolongate(const unsigned int to_level,
   Assert((to_level >= 1) && (to_level <= prolongation_matrices.size()),
          ExcIndexRange(to_level, 1, prolongation_matrices.size() + 1));
 
-  VectorType copy_src(src);
   if (this->mg_constrained_dofs != nullptr &&
       this->mg_constrained_dofs->get_user_constraint_matrix(to_level - 1)
           .get_local_lines()
           .size() > 0)
-    this->mg_constrained_dofs->get_user_constraint_matrix(to_level - 1)
-      .distribute(copy_src);
-  prolongation_matrices[to_level - 1]->vmult(dst, copy_src);
+    {
+      VectorType copy_src(src);
+      this->mg_constrained_dofs->get_user_constraint_matrix(to_level - 1)
+        .distribute(copy_src);
+      prolongation_matrices[to_level - 1]->vmult(dst, copy_src);
+    }
+  else
+    {
+      prolongation_matrices[to_level - 1]->vmult(dst, src);
+    }
 }
 
 
--- a/source/multigrid/mg_transfer_prebuilt.cc
+++ b/source/multigrid/mg_transfer_prebuilt.cc
@@ -85,10 +85,11 @@ MGTransferPrebuilt<VectorType>::prolongate(const unsigned int to_level,
   Assert((to_level >= 1) && (to_level <= prolongation_matrices.size()),
          ExcIndexRange(to_level, 1, prolongation_matrices.size() + 1));
 
-  prolongation_matrices[to_level - 1]->vmult(dst, src);
+  VectorType copy_src(src);
   if (this->mg_constrained_dofs != nullptr)
-    this->mg_constrained_dofs->get_user_constraint_matrix(to_level).distribute(
-      dst);
+    this->mg_constrained_dofs->get_user_constraint_matrix(to_level - 1)
+      .distribute(copy_src);
+  prolongation_matrices[to_level - 1]->vmult(dst, copy_src);
 }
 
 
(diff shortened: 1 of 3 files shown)


peterrum:
Why have you closed the other PR?

mathmerizing:
I basically redid my old PR with this Pull Request. At first I haven't seen the naming convention for the commits, where you need to specify first and last name for each commit. I thought it would be faster for me to just make a clean new PR instead of going back and fixing the author name in all the commits

mathmerizing:
I think that now the Travis CI check shouldn't fail

mathmerizing:
> I think that now the Travis CI check shouldn't fail

Nevermind, I have no clue what I am doing wrong and why the Travis CI build is failing...

tjhei:
Thank you for your first contribution! We require that you set a first name and last name in your git configuration. See
For more info. You do not need to close this PR but you can update it after setting your user information.

tjhei:
@tcclevenger can you take a look at this PR? Does this seems reasonable?

To make things easier for @mathmerizing , I would propose we merge this PR with a changelog entry and follow up with a test. @tcclevenger can I ask you to do that?

mathmerizing:
@tjhei Should I still push the nullptr check proposed by @masterleinad together with a changelog entry?

tjhei:
Yes, please!

source/multigrid/mg_transfer_prebuilt.cc
```diff
@@ -85,15 +85,9 @@ MGTransferPrebuilt<VectorType>::prolongate(const unsigned int to_level,
...
 
-#ifdef DEBUG
-  if (this->mg_constrained_dofs != nullptr)
-    Assert(this->mg_constrained_dofs->get_user_constraint_matrix(to_level - 1)
-               .get_local_lines()
-               .size() == 0,
-           ExcNotImplemented());
-#endif
-
   prolongation_matrices[to_level - 1]->vmult(dst, src);
```
tcclevenger, line 88: We found when implementing for MGTransferMatrixFree (in pull request https://github.com/dealii/dealii/pull/8730) that we need to distribute the user constraints before prolongation (i.e. moving `this->mg_constrained_dofs->get_user_constraint_matrix(to_level).distribute()` to before this `->vmult()` call), and then do nothing for the restriction.

source/multigrid/mg_transfer_prebuilt.cc
```diff
@@ -109,6 +103,8 @@ MGTransferPrebuilt<VectorType>::restrict_and_add(const unsigned int from_level,
   (void)from_level;
 
   prolongation_matrices[from_level - 1]->Tvmult_add(dst, src);
```
tcclevenger, line 106: It should be unnecessary to call distribute here..

source/multigrid/mg_transfer_prebuilt.cc
mathmerizing: I tried out your proposed changes. I am testing these changes on my multigrid for the Eddy Current problem. I am testing with 2,3,4,5 global refinements and with a multiplicative Schwarz smoother. 
NOTATION: {n_global_refinements} -> {n_iteration_steps}

1. No distribute() in restrict, distribute() **before** prolongation: 2 -> 21, 3 -> 25, 4 -> 26, 5 -> KILLED

2. No distribute() in restrict, distribute() **after** prolongation: 2 -> 5, 3 -> 6, 4 -> 6, 5 -> 6

3. without your proposed changes: 2 -> 5, 3 -> 6, 4 -> 5, 5 -> 5, 6 -> 5, 7 -> 5

Runs 1 - 3 only differ by the changes to the deal.ii library.

tcclevenger: This is very curious to me since applying constraints after prolongation should not have an effect. Maybe I am missing a difference in MGTransferPrebuilt that is not in MGTransferMatrixFree. 

This may be silly, but could you do the run but without distributing constraints at all during the transfer? (i.e. just commenting out the Assert that doesn't allow for user constraints and not calling distribute).

mathmerizing: I created this pull request, because beforehand I didn't distribute any constraints at all during the transfer. My results were: 2 -> 21, 3 -> 25, 4 -> 26, 5 -> 30. Obviously the issue then was the lack of the constraints in the transfer

tcclevenger: Hmm, yeah, this confuses me as to why this is the case. So distributing only before prolongation iterations are the exact same as if using no constraints. @tjhei thoughts?

Clearly your iteration counts are low and constant, so this appears to be correct for this problem. Let me do some tests with the MatrixFreeTransfer to see if I learn anything new since the two are going to need to be consistent.

mathmerizing: > So distributing only before prolongation iterations are the exact same as if using no constraints. 

Exactly. And with no constraints the preconditioned system matrix had many eigenvalues close to 0 which indicates that it was singular.

> Let me do some tests with the MatrixFreeTransfer to see if I learn anything new since the two are going to need to be consistent.

Sounds good!

tjhei:
It could be that there are two different situations: 1) data after smoothing needs to be corrected (for example with normal flux boundaries) and 2) transfer needs to be corrected. Could that explain the problem?

tcclevenger:
@mathmerizing It seems the failing tests happen because no user constraints are stored but they are begin distributed in the transfer. See lines 389-411 in mg_transfer_matrix_free.cc for how we took care of this issue. Something similar could be done here I imagine.

mathmerizing:
Is there anything else I should do for this PR? Why is the OSX serial build failing?

tjhei:
@tcclevenger you communicated about this PR offline, right? Can I ask you to summarize the conclusions?

mathmerizing:
> @tcclevenger you communicated about this PR offline, right? Can I ask you to summarize the conclusions?

I computed the curl boundary constraints via VectorTools::project_boundary_values_curl_conforming_l2 which is currently not setup to work on different level meshes. After a workaround, I now should have the correct constraints on each level. Now it doesn't make a difference, whether the user_constraints are being applied before the prolongation, like @tcclevenger suggested, or after the grid transfer operations. However, in @tcclevenger 's code it became clear that the constraints only need to be applied before prolongation. Thus now this PR should be consistent with MGMatrixFree

tcclevenger:
Yes, I tested using this branch on the ASPECT code that prompted the addition of user constraints in MGTransferMatrixFree and it works as expected (same output for prebuilt and matrix-free), so I am satisfied that it is correct. 

After this pull request is merged I’ll create a separate pull request with a test (same test as for matrix-free transfer).

tjhei:
> After this pull request is merged I’ll create a separate pull request with a test (same test as for matrix-free transfer).

awesome, thank you.

source/multigrid/mg_transfer_prebuilt.cc
```diff
@@ -85,15 +85,14 @@ MGTransferPrebuilt<VectorType>::prolongate(const unsigned int to_level,
...
-#ifdef DEBUG
-  if (this->mg_constrained_dofs != nullptr)
-    Assert(this->mg_constrained_dofs->get_user_constraint_matrix(to_level - 1)
-               .get_local_lines()
-               .size() == 0,
-           ExcNotImplemented());
-#endif
-
-  prolongation_matrices[to_level - 1]->vmult(dst, src);
+  VectorType copy_src(src);
```
tjhei, line 88: would you mind changing this to only make the copy if this the following ``if`` is indeed true? You will likely need an else branch for the vmult() without the vector copy.

tjhei:
@masterleinad can you take another look?

tcclevenger:
@masterleinad I've created a pull request (https://github.com/dealii/dealii/pull/10481) ready for after this is merged.

PR #10325, opened 2020-05-23
Generalize vertex/line/quad_index

peterrum:
Rely in `internal::TriaAccessorImplementation::Implementation::quad_index()`, `internal::TriaAccessorImplementation::Implementation::line_index()`, and `internal::TriaAccessorImplementation::Implementation::vertex_index()` completely on `GeometryInfo`. 

**Note:** This is solely an internal change.

**internal::TriaAccessorImplementation::Implementation::quad_index()**:
```cpp
      template <int dim, int spacedim>
      inline static unsigned int
      quad_index(const TriaAccessor<3, dim, spacedim> &accessor,
                 const unsigned int                    i)
      {
        return accessor.tria->levels[accessor.present_level]
          ->cells.cells[accessor.present_index]
          .face(i);
      }
```

**internal::TriaAccessorImplementation::Implementation::line_index()**:
```cpp
      template <int dim, int spacedim>
      inline static unsigned int
      line_index(const TriaAccessor<2, dim, spacedim> &accessor,
                 const unsigned int                    i)
      {
        return accessor.objects().cells[accessor.present_index].face(i);
      }


      template <int dim, int spacedim>
      inline static unsigned int
      line_index(const TriaAccessor<3, dim, spacedim> &accessor,
                 const unsigned int                    i)
      {
        const auto pair = GeometryInfo<3>::standard_line_to_quad_line_index(i);
        const auto quad_index = pair[0];
        const auto line_index = GeometryInfo<3>::standard_to_real_face_line(
          pair[1],
          accessor.face_orientation(quad_index),
          accessor.face_flip(quad_index),
          accessor.face_rotation(quad_index));

        return accessor.quad(quad_index)->line_index(line_index);
      }
```

**internal::TriaAccessorImplementation::Implementation::vertex_index()**:
```cpp

      template <int dim, int spacedim>
      inline static unsigned int
      vertex_index(const TriaAccessor<1, dim, spacedim> &accessor,
                   const unsigned int                    corner)
      {
        return accessor.objects().cells[accessor.present_index].face(corner);
      }


      template <int dim, int spacedim>
      inline static unsigned int
      vertex_index(const TriaAccessor<2, dim, spacedim> &accessor,
                   const unsigned int                    corner)
      {
        const auto pair =
          GeometryInfo<2>::standard_corner_to_line_vertex_index(corner);
        const auto line_index   = pair[0];
        const auto vertex_index = GeometryInfo<2>::standard_to_real_line_vertex(
[...]

PR #10357, opened 2020-05-25
Add missing docu to GeometryInfo

peterrum:
Follow up to #10325

FYI @bangerth

### DIFF
--- a/include/deal.II/base/geometry_info.h
+++ b/include/deal.II/base/geometry_info.h
@@ -2312,18 +2312,44 @@ struct GeometryInfo
                              const bool         face_flip        = false,
                              const bool         face_rotation    = false);
 
+  /**
+   * Map the vertex index @p vertex of a line in standard orientation to one of a
+   * face with arbitrary @p line_orientation. The value of this flag default to
+   * <tt>true</tt>.
+   */
   static unsigned int
-  standard_to_real_line_vertex(const unsigned int line,
-                               const bool         face_orientation = true);
+  standard_to_real_line_vertex(const unsigned int vertex,
+                               const bool         line_orientation = true);
 
+  /**
+   * Decompose the vertex index in a quad into a pair of a line index and a
+   * vertex index within this line.
+   *
+   * @note Which line is selected is not of importance (and not exposed on
+   *   purpose).
+   */
   static std::array<unsigned int, 2>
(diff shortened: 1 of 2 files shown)


include/deal.II/base/geometry_info.h
```diff
@@ -2312,18 +2312,35 @@ struct GeometryInfo
...
+  standard_to_real_line_vertex(const unsigned int vertex,
+                               const bool         line_orientation = true);
 
+  /**
+   * Decompose the vertex index in a quad into a pair of a line index and a
+   * vertex index within this line.
+   */
   static std::array<unsigned int, 2>
-  standard_corner_to_line_vertex_index(const unsigned int corner);
+  standard_quad_vertex_to_line_vertex_index(const unsigned int vertex);
```
bangerth, line 2329: Isn't this non-unique? That is, aren't there multiple lines that contain the vertex?

include/deal.II/base/geometry_info.h
```diff
@@ -2312,18 +2312,35 @@ struct GeometryInfo
...
-  standard_corner_to_quad_vertex_index(const unsigned int corner);
+  standard_hex_vertex_to_quad_vertex_index(const unsigned int vertex);
 
+  /**
+   * Decompose the line index in a hex into a pair of a quad index and a line
+   * index within this quad.
+   */
   static std::array<unsigned int, 2>
-  standard_line_to_quad_line_index(const unsigned int line);
+  standard_hex_line_to_quad_line_index(const unsigned int line);
```
bangerth, line 2343: same here

include/deal.II/base/geometry_info.h
peterrum: Indeed. The function selects one possible one as it used to be done in `CellAccessor`. This is generalizable to TETs (see: https://github.com/peterrum/dealii/pull/7#issuecomment-630276686).

bangerth: Can you update the documentation to say that the function chooses one, and maybe which one (or say that it chooses "one" and that it is purposefully unspecified which one that is)?

peterrum: I have more or less copied the comment from `CellId`. Is that fine this way?

PR #10359, opened 2020-05-25
Fix warning in set_bit

peterrum:
`MSVC - appveyor` is giving warnings:

```
C:\projects\source\include\deal.II\grid\tria_accessor.templates.h(751,1): warning C4804: '-': unsafe use of type 'bool' in operation [C:\projects\source\build\source\matrix_free\obj_matrix_free_debug.vcxproj]
```


include/deal.II/grid/tria_accessor.templates.h
```diff
@@ -748,7 +748,7 @@ namespace internal
         // source:
         // https://stackoverflow.com/questions/47981/how-do-you-set-clear-and-toggle-a-single-bit
         // "Changing the nth bit to x"
-        number ^= (-x ^ number) & (1UL << n);
+        number ^= (-static_cast<char>(x) ^ number) & (1UL << n);
```
bangerth, line 751: I find that exceedingly difficult to read but probably correct. Among the questions I have are:
* Does this make the assumption that `char` is `signed`? The standard doesn't actually guarantee that, but I think that's what you assume here.
* You don't check that `n` is between zero and 7. That would probably be worth an assertion. 
* Since you're only working on one byte, you don't need `1UL`. You could probably get away with `signed int` or even `signed short`.

drwells:
I'm late to the party, but could we clean this up by using `std::bitset` instead of doing our own shifts and ands?

peterrum:
@drwells Thanks for the suggestion. I am not sure if it really helps to simplify the code. The input/output has to be (unsigned) char to be enough general for TETs. Only in the case of HEX, we can interpret the individual bits.

That would mean that if I want to set a bit, I would have convert the `char` to a `bitset`, set the bit, convert the biset to a long via `to_ulong`, which then has to be cast to `char`.

peterrum:
@bangerth I have made the changes. Or what is your opinion about the `bitset` idea?

bangerth:
I never used `std::bitset`. It has the advantage that it has `operator[]` and `set()`. I don't know whether it packs to a single byte if the number of bits is less than 8. 

For the current purpose, you're thinking of using everything in internal interfaces. In particular, we ought to fix the warning in the function regardless of what data type we use elsewhere.

PR #10360, opened 2020-05-25
Fix compiling with clang-10 with C++20 support

### LINKED ISSUE #10349: Fix compilation with -std=c++20 and gcc-10
It is currently not possible to compile deal.II with `-std=c++20` due to new rules regarding operator overloading as evidenced here: https://cdash.43-1.org/viewBuildError.php?buildid=6688

As a temporary workaround it is enough to remove `LinearIndexIterator::operator!=()` which allows the compiler to resolve the overload in some compatibility mode (that simply emits a warning instead).

A proper fix for this issue might be to remove all `operator@@()` variants (in case `DEAL_II_HAVE_CXX20` is true) and replace them by a single `operator<=>()`.

masterleinad:
Fixes #10349. With this patch I can at least compile without warnings.

### DIFF
--- a/include/deal.II/base/linear_index_iterator.h
+++ b/include/deal.II/base/linear_index_iterator.h
@@ -249,8 +249,16 @@ public:
    * Comparison operator. Returns <code>true</code> if both iterators point to
    * the same entry in the same container.
    */
-  bool
-  operator==(const LinearIndexIterator &) const;
+  template <typename OtherIterator>
+  friend typename std::enable_if<
+    std::is_convertible<OtherIterator, DerivedIterator>::value,
+    bool>::type
+  operator==(const LinearIndexIterator &left, const OtherIterator &right)
+  {
+    const auto &right_2 = static_cast<const DerivedIterator &>(right);
+    return left.accessor.container == right_2.accessor.container &&
+           left.accessor.linear_index == right_2.accessor.linear_index;
+  }
 
   /**
    * Inverse of operator==().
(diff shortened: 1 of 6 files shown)


drwells:
I suspect we will need more for `LinearIndexIterator` - does the test `transpose_table_iterators_0` pass with this patch?

include/deal.II/base/linear_index_iterator.h
```diff
@@ -249,8 +249,16 @@ class LinearIndexIterator
...
    */
-  bool
-  operator==(const LinearIndexIterator &) const;
+  template <typename OtherIterator>
+  friend typename std::enable_if<
+    std::is_convertible<OtherIterator, DerivedIterator>::value,
+    bool>::type
+  operator==(const LinearIndexIterator &left, const OtherIterator &right)
+  {
+    const auto &right_2 = static_cast<const DerivedIterator &>(right);
```
drwells, line 258: :+1:

masterleinad:
OK. This seems to work now.

include/deal.II/base/linear_index_iterator.h
```diff
@@ -249,8 +249,15 @@ class LinearIndexIterator
    * Comparison operator. Returns <code>true</code> if both iterators point to
    * the same entry in the same container.
    */
-  bool
-  operator==(const DerivedIterator &) const;
+  template <typename OtherIterator>
+  friend typename std::enable_if<
```
tamiko, line 253: What about `std::enable_if_t`? :smile:

masterleinad: Oh, sure we can replace that everywhere now if anyone cares.

PR #10363, opened 2020-05-26
GridTools::Cache::get_locally_owned_cell_bounding_boxes_rtree()

luca-heltai:
The GridTools::Cache::get_cell_bounding_boxes_rtree() returns all cells, including artificial ones. There is now a new version that returns a tree of the locally owned cells. This is then used together with the `extract_rtree_level` function to produce an efficient covering of the domain.


doc/news/changes/minor/20200526LucaHeltai
```diff
@@ -0,0 +1,6 @@
+New: GridTools::Cache::get_locally_owned_cell_bounding_boxes_rtree() extracts
+a tree of bounding boxes covering the locally owned cells of a triangulation.
+This can be used in conjunction with GridTools::Cache::get_covering_rtree() to
+make approximate geometrical queries on who owns what spacial region.
```
bangerth, line 4:
```suggestion
make approximate geometrical queries on who owns what spatial region.
```

include/deal.II/grid/grid_tools_cache.h
```diff
@@ -138,6 +138,16 @@ namespace GridTools
                 typename Triangulation<dim, spacedim>::active_cell_iterator>> &
     get_cell_bounding_boxes_rtree() const;
 
+    /**
+     * Return the cached RTree object of bounging boxes containing locally owned
```
bangerth, line 142:
```suggestion
     * Return the cached RTree object of bounding boxes containing locally owned
```

bangerth: The same typo exists in the previous function.

include/deal.II/grid/grid_tools_cache.h
```diff
@@ -138,6 +138,16 @@ namespace GridTools
...
+     * active cells, constructed using the active cell iterators of the stored
+     * triangulation.
```
bangerth, line 144: I think "locally owned" implies "active". I had to read through the description of this and the previous function several times before I really understood the difference. Would you mind adding a sentence to both functions that contrasts what they do? E.g., "In contrast to the previous function, this function builds the RTree using only the locally owned cells, but does not include ghost cells."

include/deal.II/grid/grid_tools_cache.h
```diff
@@ -170,12 +182,15 @@ namespace GridTools
...
+     * The local bounding boxes are constructed by calling the function
+     * extract_rtree_level() applied to the object returned by
+     * get_locally_owned_cell_bounding_boxes_rtree(), with the specified level
+     * as input.
+     *
+     * @author Giovanni Alzetta, 2018, Luca Heltai, 2020
      */
     const RTree<std::pair<BoundingBox<spacedim>, unsigned int>> &
-    get_covering_rtree() const;
+    get_covering_rtree(const unsigned int level = 0) const;
```
bangerth, line 193: I don't think I understand what the level corresponds to. Maybe I need to look this up with `get_locally_owned_cell_bounding_boxes_rtree()` -- but actually, having done so, that function doesn't take a level.

include/deal.II/grid/grid_tools_cache.h
```diff
@@ -219,9 +234,15 @@ namespace GridTools
...
+     * A collection of rtree objects covering the whole mesh.
+     *
+     * Each entry of the map is constructed from the function
+     * extract_rtree_level() applied to the rtree returned by the function
+     * get_locally_owned_cell_bounding_boxes_rtree(), with the specified level
+     * in input.
      */
-    mutable RTree<std::pair<BoundingBox<spacedim>, unsigned int>>
+    mutable std::map<unsigned int,
+                     RTree<std::pair<BoundingBox<spacedim>, unsigned int>>>
```
bangerth, line 245: same issue here -- what does `level` correspond to?

source/grid/grid_tools_cache.cc
```diff
@@ -160,42 +188,31 @@ namespace GridTools
 
   template <int dim, int spacedim>
   const RTree<std::pair<BoundingBox<spacedim>, unsigned int>> &
-  Cache<dim, spacedim>::get_covering_rtree() const
+  Cache<dim, spacedim>::get_covering_rtree(const unsigned int level) const
   {
-    if (update_flags & update_covering_rtree)
+    if (update_flags & update_covering_rtree ||
+        covering_rtree.find(level) == covering_rtree.end())
```
bangerth, line 194: Hm, I think I'm starting to realize that "level" has nothing to do with the level of a triangulation, but has some other meaning in rtrees. I think it would be useful to clarify this in pretty much all places of the documentation...

include/deal.II/grid/grid_tools_cache.h
luca-heltai: it is the `extract_rtree_level()` that takes a level as input...

include/deal.II/grid/grid_tools_cache.h
luca-heltai: to the argument passed to `extract_rtree_level()`

source/grid/grid_tools_cache.cc
luca-heltai: the function we used is called `extract_rtree_level`. Should we change the name of that function? Level refers to that. I'll try to reformulate the documentation accordingly.

include/deal.II/grid/grid_tools_cache.h
bangerth: I see. Can you point that out in the documentation that that is what `level` refers to?

bangerth: I'm afraid that, after reading through the documentation of that function, I still don't know what the level of an rtree is :-(

include/deal.II/grid/grid_tools_cache.h
luca-heltai: What we really want is "non artificial" and "locally owned" but not ghost. How can we call these?

include/deal.II/grid/grid_tools_cache.h
luca-heltai: but that is well defined for an `rtree` object: see https://en.wikipedia.org/wiki/R-tree. And we do point to that page in the `RTree` documentation. :)

include/deal.II/grid/grid_tools_cache.h
luca-heltai: I've updated the documentation.

include/deal.II/grid/grid_tools_cache.h
luca-heltai: I added a phrase to specify what we mean by level (and added a pointer to the wikipedia page).

luca-heltai:
ping @bangerth ?

PR #10364, opened 2020-05-26
Move template definitions from hdf5.cc to hdf5.h

dangars:
The code of the HDF5 interface of deal.II is very templated. I think that it is better to place it in hdf5.templates.h

I didn't instantiated all the possible combinations/instiantiations. If we move the code to hdf5.templates.h, then the user can do the instantiations.


source/base/hdf5.cc
```diff
@@ -1501,110 +461,6 @@ namespace HDF5
   template void
   HDF5Object::set_attribute<unsigned int>(const std::string &attr_name,
                                           unsigned int       value);
-
-  template std::vector<int>
-  DataSet::read<std::vector<int>>();
-  template std::vector<unsigned int>
-  DataSet::read<std::vector<unsigned int>>();
```
dangars, line 1508: I removed many instantiations that are not very useful

tests/base/hdf5_03.cc
```diff
@@ -803,14 +803,6 @@ main(int argc, char **argv)
...
 #endif
 
-        write_test<std::vector, int>(data_file,
-                                     dataset_dimensions,
-                                     mpi_communicator,
-                                     pcout);
-        write_test<std::vector, unsigned int>(data_file,
-                                              dataset_dimensions,
-                                              mpi_communicator,
-                                              pcout);
```
dangars, line 813: Because I removed many instantiations, I had to modify the tests hdf5_03 and and hdf5_04

tests/base/hdf5_04.cc
```diff
@@ -743,8 +743,6 @@ main()
                                                       dataset_dimensions);
 #endif
 
-        write_test<std::vector, int>(data_file, dataset_dimensions);
-        write_test<std::vector, unsigned int>(data_file, dataset_dimensions);
```
tamiko, line 747: I agree that these tests don't look terribly useful

dangars:
I realized that I updated the output for the complex test hdf5_03 and hdf5_04, but not for the real tests, therefore I think that the tests hdf5_03 and hdf5_04 should fail. I'm going to fix that.

bangerth:
Let me remove the "Ready to merge" tag. Let us know when you think you're done!

dangars:
I have the feeling that the tests for hdf5 are not running. Where can I see the tests that usually run?
I added a commit to purposely make the tests fail and it seems that the tests pass. Below you can find the tests that should fail
fail:tests/base/hdf5_03.with_hdf5=true.with_complex_values=on.mpirun=1.output
tests/base/hdf5_03.with_hdf5=true.with_complex_values=off.mpirun=1.output
tests/base/hdf5_03.with_hdf5=true.with_complex_values=on.mpirun=4.output
tests/base/hdf5_03.with_hdf5=true.with_complex_values=off.mpirun=4.output
tests/base/hdf5_04.with_hdf5=true.with_complex_values=off.output
tests/base/hdf5_04.with_hdf5=true.with_complex_values=on.output

dangars:
I wonder if it would make sense to remove all the instantiations of HDF5 and make the HDF5 interface a template only interface. Most of the instantiations are never going to be used. But I would like to test them in case somebody use them.

Below you can find the time to compile/run the test hdf5_03 with the instantiations in deal.II
```
1/4
Test #1414: base/hdf5_03.mpirun=1.debug ......   Passed   30.27 sec

Size of deal.II1157696 lib/libdeal_II.g.so.9.3.0-pre
167068  lib/libdeal_II.so.9.3.0-pre
```

Below you can find the time to compile/run the test hdf5_03 without the instantiations in deal.II therefore hdf5.templates.h has to be used in hdf5_03. Only 5 sec of difference
```
1/4 Test #1414: base/hdf5_03.mpirun=1.debug ......   Passed   35.97 sec
1157060 lib/libdeal_II.g.so.9.3.0-pre
166936  lib/libdeal_II.so.9.3.0-pre
```

masterleinad:
Currently, `HDF5` is not enabled on the serial CI runner but only on the MPI one that only runs a subset of tests.

dangars:
@masterleinad Thanks! Now I understand :)

dangars:
I updated the PR, following up on the discussion with @tamiko and @bangerth I simplified this PR.
- I removed all the instantiations
- I moved the template definitions from hdf5.cc to hdf5.h
- I din't have to modify the tests. The interface is the same, it shouldn't break any code.

dangars:
This is ready for review. I separated in logical commits, that way it is easier to review.

dangars:
Hi, I rebased the PR. Is there something else I should do :) ?

drwells:
Sorry for the long delay in reviews.

Moving these all into the header is a good thing but its currently not laid out in our typical style: even with template functions we usually have declarations at the top of the file and definitions at the bottom (instead of just providing definitions). Would you please split things up in this way?

dangars:
@drwells thanks for reviewing the code!

I don't have access to a computer this week, I'll address your comments next week!

dangars:
I added the declarations of the internal functions as you suggested. I also added some documentation about the internal functions that now appears in Doxygen.

The modifications are in the commit fd1f05ed8175dd98758b98197dd60e1fa716cf89

dangars:
Great!

PR #10367, opened 2020-05-26
 Compile MSVC testers with warnings as errors

masterleinad:
Depends resolving the issue addressed in #10359.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,6 +8,7 @@ jobs:
     name: Windows Serial
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [windows-2019, windows-2016]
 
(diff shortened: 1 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -8,6 +8,7 @@ jobs:
     name: Windows Serial
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
tjhei, line 11: This will run both to completion, right? Is this useful to keep in?

masterleinad: Yes, this avoids that other runs are aborted if one of them failed. I always found that confusing and normally I want to see all the (possible) errors at once.

masterleinad:
Rebased.

.github/workflows/windows.yml
tjhei: makes sense.

PR #10374, opened 2020-05-26
Better document Threads::internal::return_value.

bangerth:
I would have liked to replace this class by `std::future`, but we can't because `std::future::get()` can only be called once and we have semantics that allow us to read the returned value of a thread or task as many times as we damn well please.


masterleinad:
What about `std::shared_future`?

bangerth:
Long story coming later today. The problem with `std::shared_future` is that its `get()` function returns a value or a `const` reference. Neither is useful if the type in question is `std::future`.

I'd like to leave it as is right here, and get it merged as-is. I will show how the final code later today.

bangerth:
Ping?

drwells:
It's a nice documentation improvement so lets merge!

PR #10376, opened 2020-05-27
Krishna fix redundant edge step50

krishnakumarg1984:


examples/step-67/doc/builds-on
```diff
@@ -1 +1 @@
-step-33 step-48 step-59 step-40
+step-33 step-48 step-59
```
bangerth, line 1: You'll have to take this change out of the patch -- it's already tracked by #10375 .

krishnakumarg1984:
Thanks for merging

PR #10377, opened 2020-05-27
Improve documentation of Threads::Task.

bangerth:


include/deal.II/base/thread_management.h
```diff
@@ -1204,13 +1204,26 @@ namespace Threads
...
+   * value that was returned by the function that was run on that background
+   * task.
    *
-   * @author Wolfgang Bangerth, 2009
+   * This class is conceptually similar to the
+   * [`std::future`](https://en.cppreference.com/w/cpp/thread/future) class that
+   * is returned by
+   * [`std::async`](https://en.cppreference.com/w/cpp/thread/async) (which is
+   * itself similar to what Threads::new_task() does). The principal conceptual
+   * difference is that one can only call `std::future::get()` once, where one
```
rezarastak, line 1223:
```suggestion
   * difference is that one can only call `std::future::get()` once, whereas one
```

bangerth:
I fixed the mistake pointed out by @rezarastak .

PR #10386, opened 2020-05-27
Remove our deprecated nanoflann bindings

drwells:
We deprecated these in #8081 - since they showed up in two releases I think its time to let them go.


luca-heltai:
We need to update this in spack as well. @jppelteret ?

luca-heltai:
FYI: 

https://github.com/spack/spack/pull/16839

PR #7392, opened 2018-10-27
Deprecate the NetCDF bindings.

drwells:
Fixes #5570: we are only compatible with the sparsely-supported `netcdf-cxx-legacy` package and no one seems interested enough in this format to fix that, so I have deprecated the bindings.

PR #10387, opened 2020-05-27
Remove our deprecated netcdf bindings

drwells:
Note that we don't really support netcdf - we support the netcdf C++ legacy bindings. These were deprecated in #7392.


tjhei:
Wait, what? I was going to use netcdf for a project.

drwells:
I don't understand what you are asking - what parts of our netcdf bindings would you like to use?

drwells:
> So, can we keep netcdf as a dependency

Its important to note that we depend on a package usually named `netcdf-cxx-legacy` and not the current version of netcdf because no one has ported our older C++ wrappers to their newer API.

tamiko:
@tjhei What about you simply adopt the `FindNETCDF.cmake` file into your own project (replacing the `DEAL_II_PACKAGE_HANDLE` call with an export target. I am happy to write that for you.

tjhei:
I would like to introduce functionality to use netcdf I/O, but I can use the C library and reintroduce the cmake functionality if you insist on removing it now. Otherwise, I would suggest we leave the cmake handling in place and just remove the deprecated Grid functionality right now.

drwells:
As far as I can tell we don't currently support linking with the netcdf C library - we only link against `libnetcdf_c++`, so removing this shouldn't impact your use case.

Perhaps your netcdf package is installed in a different way than mine is. Can you compile and link, with the current deal.II netcdf configuration, with/against the netcdf c library you want to use?

kronbichler:
Can you please rebase?

tjhei:
> Can you compile and link, with the current deal.II netcdf configuration, with/against the netcdf c library you want to use?

no, I have netcdf and netcdf-cxx installed.

drwells:
I rebased.

As far as I can tell this is unrelated to @tjhei's use case so lets go ahead with removing.

PR #10389, opened 2020-05-27
Rewrite the Threads::Task class using C++11 features.

bangerth:
This saves us 170 lines of code and makes the whole thing a lot simpler to read, I believe :-)

The principal challenge here is the fact that we want `Task` objects that one can repeatedly query using the `return_value()` function. `std::future` does not allow that: it invalidates the return value after the first read. `std::shared_future` does allow that, but it can't be used with objects such as `std::unique_ptr` that can only be moved. It's all very awkward and I think that our class `Task` is actually better suited to do something like
```
class X {
  Task<int> member;

  X()
    : member ([]() {...; return 42;})     // computes 'member' in the background
  {}

  int get_member() {
    return member.return_value();   // waits if necessary
  }
};
```

/rebuid


bangerth:
I pushed a second commit that also verifies that exception propagation works as intended, and documents the feature.

bangerth:
I've got follow-ups, but this is ready to be reviewed by itself. I would appreciate if we could get this merged and I will do the rest as independent PRs.

Things to come:
* Don't actually create a task if `n_threads == 1`.
* Create the tasks via TaskFlow.

I know how to do both, but it's going to be conceptually easiest to do this in separate PRs.

bangerth:
Part of #10389.

PR #10393, opened 2020-05-27
Step-72: The minimal surface equation using automatic differentiation (step-15 revisited)

### LINKED ISSUE #10366: Better visibility for the AD feature from recent versions of deal.ii (beginning from 9.1)
Automatic differentiation example is currently included in Step-33. This is slightly harder to access because

1.  Euler laws of gas dynamics are generally a more specialised problem to CFD (and not in the solid mechanics community) and thus a lot of people will miss out, and

2. Lots of beginners will never reach Step-33 (or will take a long time to reach there)

It would lend better visibility to the AD feature if Step-15 can incorporate it showing the Newton method. I claim better visibility because solving non-linear PDEs is such a common task and lots of new users will anyway come to Step-15. It is also very error prone to manually formulate the Newton-update equation. So, AD will be the perfect candidate to make life easier here. 

Can we at least consider this for the Step-15 extension section?

jppelteret:
Fixes #10366

TODO:
- [x] ~~Use constraints.distribute_local_to_global and remove BC setters (i.e. simplify the logic there)~~ Leaving this for now to minimise the changes between step-15 and this tutorial.
- [x] Apply patch from #12129 
- [x] Re-apply these changes: https://github.com/dealii/dealii/pull/10392#issuecomment-832510538


jppelteret:
Some preliminary timings (residual tolerance is `1e-2`(:

Full hand linearisation
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      32.9s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble                        |        50 |      1.54s |       4.7% |
| Solve                           |        50 |      30.1s |        91% |
+---------------------------------+-----------+------------+------------+
```

Residual linearisation using AD 
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      43.3s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble                        |        50 |      8.72s |        20% |
| Solve                           |        50 |      33.1s |        76% |
+---------------------------------+-----------+------------+------------+
```

Variational formulation (AD used to compute residual and linearisation)
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      51.6s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble                        |        50 |        19s |        37% |
| Solve                           |        50 |      31.2s |        61% |
+---------------------------------+-----------+------------+------------+
```

bangerth:
Is the last one where you basically just describe an energy?

jppelteret:
> Is the last one where you basically just describe an energy?

Yes, exactly.

examples/step-72/doc/results.dox
```diff
@@ -0,0 +1,132 @@
...
+  Residual: 0.00831928
+  Residual: 0.0074859
+  Residual: 0.0067363
+  Residual: 0.00606197
+  Residual: 0.00545529
+```
+
+So what is interesting for us to compare is how long the assembly process takes
+for the three different implementations, and to put that into some greater context.
+Below is the output for the hand linearization (as computed on a circa 2012
```
bangerth, line 33: :-)

bangerth:
I think this is going to be a nice complement to step-77 (#11953) in that the two address the two awkward points in step-15.

As for the results shown: Are these run-times in release or in debug mode?

jppelteret:
> I think this is going to be a nice complement to step-77 (#11953) in that the two address the two awkward points in step-15.

I agree. Thank you for adding the note that links the two tutorials.

> As for the results shown: Are these run-times in release or in debug mode?

These are in release mode. My computer is really starting to show its age :-D

jppelteret:
@bangerth I need just one more day to finish the intro paragraph to `assemble_system_using_energy_functional()`. Once that's done, then I think that I would have sketched out the concepts that I want to transmit. Perhaps after then we could figure out how to best integrate that with what you added 2 days ago. I think that we might need to write the weak forms in full so that I can use a notation with DoF indices "I", "J". I think that this is the best way to explain what derivatives are actually being computed. I'm pretty sure that what I've written today in the paragraph above `assemble_system_with_residual_linearization()` isn't mathematically sound, but at least for now you can get the gist of what I'm trying to say.

bangerth:
Last patch at https://github.com/jppelteret/dealii/pull/7

jppelteret:
I've uploaded the PDF rendered documentation to the step-72 tutorial, which is now ready for review. Due to the doxygen links to step-71 and step-77, it depends on #10392 and #11953.

[2021-05-03-step-72-draft.pdf](https://github.com/dealii/dealii/files/6417342/2021-05-03-step-72-draft.pdf)

bangerth:
I'm ok with merging as is, but will be happy to let others have a look as well! In the interest of making progress with open PRs, though, maybe I can propose that this be merged tomorrow? Any follow-up requests could then be accomplished via separate PRs...

bangerth:
You do need to indent, though, @jppelteret !

bangerth:
Oh, actually, the indent check fails because of the references to step-71 and step-77, which probably have their own circular references to step-72. We might have to merge these at more or less the same time...

bangerth:
@jppelteret Can you apply the same change as in #12129 also to your program here?

jppelteret:
OK @bangerth, will do!

peterrum:
> I would suggest that we merge both this program and step-72 provided that (i) you rebase and fix the merge conflict, and (ii) squash the commits. I can then look through it all again locally and make whatever fixes I think are right/open issues if I could use help from you.

I will try to take a look at this tutorial tonight. So let's wait one more day.

examples/step-72/step-72.cc
```diff
@@ -0,0 +1,1147 @@
...
+    // vector -- the MeshWorker::CopyData therefore has `1` for all three
+    // of its template arguments.
+    using ScratchData = MeshWorker::ScratchData<dim>;
+    using CopyData    = MeshWorker::CopyData<1, 1, 1>;
+
+    // We also need to know what type of iterator we'll be working with
+    // during assembly. For simplicity, we just ask the compiler to work
+    // this out for us using the decltype() specifier, knowing that we'll
+    // be iterating over active cells owned by the @p dof_handler.
+    using CellIteratorType = decltype(dof_handler.begin_active());
```
peterrum, line 291: Why not write below simply:
```cpp
    auto cell_worker = [this](const auto &cell,
                              auto &      scratch_data,
                              auto &      copy_data) {
```
?

examples/step-72/step-72.cc
```diff
@@ -0,0 +1,1147 @@
...
+        copy_data.local_dof_indices[0];
+
+      for (unsigned int i = 0; i < dofs_per_cell; ++i)
+        {
+          for (unsigned int j = 0; j < dofs_per_cell; ++j)
+            system_matrix.add(local_dof_indices[i],
+                              local_dof_indices[j],
+                              cell_matrix(i, j));
+
+          system_rhs(local_dof_indices[i]) += cell_rhs(i);
```
peterrum, line 397: In my opinion, this is something we should only show once (step 3) and then always use `AffineConstraints`.

peterrum (approved):
Looks very good! I am fine with merging once step-71 is merged.

I have open a PR with some minor suggestions.

examples/step-72/step-72.cc
bangerth: I'm not a huge fan of `auto` for lambda functions unless the type is clear to a reader -- which I don't think it is. I'd much rather be explicit:
```suggestion
    using CellIteratorType = typename DoFHandler<dim>::active_cell_iterator;
```

bangerth: But then I'm not the author of the program and am happy to defer.

jppelteret: The way that I wrote it follows exactly the pattern that we outline in the documentation for `mesh_loop`. See [here](https://dealii.org/developer/doxygen/deal.II/group__MeshWorker.html#ga76ec61fbd188fb320fe8ca166a79b322). I'd therefore prefer to keep it as it is, if you all don't mind.

examples/step-72/step-72.cc
jppelteret: I agree, and I even thought about making the change (see my [TODO list](https://github.com/dealii/dealii/pull/10393#issue-424080398)). We'd basically need to make the change consistently across step-15, step-72 and step-77, so I ultimately decided to leave this to another time. For the sake of consistency, I'd thus propose that we leave it how it is and make the improvement after the release.

bangerth: I agree -- consistency is useful!

jppelteret:
Now that step-71 has been merged, I've rebased this test. I've also attended to all of the TODOs the I listed in https://github.com/dealii/dealii/pull/10393#issue-424080398.

bangerth:
Then let's get it merged once the testers are through!

PR #10395, opened 2020-05-27
Deprecate vanka threads

drwells:
Clean up a few things in our Vanka smoother:
- deprecate the `conserve_mem` and `n_threads` arguments
- use tasks instead of threads

The task implementation still uses its own implicit 'break the range of dofs into blocks' scheme as well as its own implicit scratch data object (in this case just a `std::vector`) - I would prefer to handle those in a followup PR.

### DIFF
--- a/include/deal.II/lac/sparse_vanka.h
+++ b/include/deal.II/lac/sparse_vanka.h
@@ -154,6 +154,18 @@ public:
   SparseVanka();
 
   /**
+   * Constructor which also takes two deprecated inputs.
+   *
+   * @deprecated The use of the last two parameters is deprecated. They are
+   * currently ignored.
+   */
+  DEAL_II_DEPRECATED
+  SparseVanka(const SparseMatrix<number> &M,
+              const std::vector<bool> &   selected,
+              const bool                  conserve_memory,
+              const unsigned int n_threads = MultithreadInfo::n_threads());
+
+  /**
    * Constructor. Gets the matrix for preconditioning and a bit vector with
    * entries @p true for all rows to be updated. A reference to this vector
    * will be stored, so it must persist longer than the Vanka object. The same
@@ -164,24 +176,8 @@ public:
    * conceivable that the preconditioner is build up for one matrix once, but
    * is used for subsequent steps in a nonlinear process as well, where the
    * matrix changes in each step slightly.
-   *
-   * If @p conserve_mem is @p false, then the inverses of the local systems
-   * are computed now; if the flag is @p true, then they are computed every
-   * time the preconditioner is applied. This saves some memory, but makes
-   * preconditioning very slow. Note also, that if the flag is @p false, then
-   * the contents of the matrix @p M at the time of calling this constructor
-   * are used, while if the flag is @p true, then the values in @p M at the
-   * time of preconditioning are used. This may lead to different results,
-   * obviously, of @p M changes.
-   *
-   * The parameter @p n_threads determines how many threads shall be used in
-   * parallel when building the inverses of the diagonal blocks. This
-   * parameter is ignored if not in multithreaded mode.
    */
-  SparseVanka(const SparseMatrix<number> &M,
-              const std::vector<bool> &   selected,
-              const bool                  conserve_memory = false,
-              const unsigned int n_threads = MultithreadInfo::n_threads());
+  SparseVanka(const SparseMatrix<number> &M, const std::vector<bool> &selected);
 
   /**
    * Destructor. Delete all allocated matrices.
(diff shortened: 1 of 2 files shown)


drwells:
We only seem to have one test, at the moment, that actually uses this class (linear_operator_08). I checked that we pass.

include/deal.II/lac/sparse_vanka.h
```diff
@@ -197,25 +193,23 @@ class SparseVanka
     /**
      * Constructor. For the parameters' description, see below.
      */
+    AdditionalData(const std::vector<bool> &selected);
+
+    /**
+     * Constructor. For the parameters' description, see below.
+     *
+     * @deprecated The use of this constructor is deprecated - the second two
```
bangerth, line 201:
```suggestion
     * @deprecated The use of this constructor is deprecated - the second and third
```

include/deal.II/lac/sparse_vanka.templates.h
```diff
@@ -133,39 +133,34 @@ SparseVanka<number>::compute_inverses()
...
-  std::vector<std::pair<size_type, unsigned int>> blocking(n_threads);
+  std::vector<std::pair<size_type, unsigned int>> blocking(n_tasks);
 
   unsigned int c      = 0;
-  unsigned int thread = 0;
+  unsigned int task_n = 0;
   blocking[0].first   = 0;
 
-  for (size_type i = 0; (i < matrix->m()) && (thread + 1 < n_threads); ++i)
+  for (size_type i = 0; (i < matrix->m()) && (task_n + 1 < n_tasks); ++i)
```
bangerth, line 142: Is this loop doing something that is expressible through the `Threads::split_range()` function?

include/deal.II/lac/sparse_vanka.templates.h
```diff
@@ -113,13 +111,15 @@ SparseVanka<number>::compute_inverses()
 #else
   const size_type n_inverses =
     std::count(selected->begin(), selected->end(), true);
-
-  const size_type n_inverses_per_thread =
-    std::max(n_inverses / n_threads, static_cast<size_type>(1U));
+  // somewhat arbitrarily set up an equal number of tasks as we have threads
+  const std::size_t n_tasks = MultithreadInfo::n_threads();
```
bangerth, line 115: This would work substantially better if you had something like 4x or 8x the number of tasks.

include/deal.II/lac/sparse_vanka.h
```diff
@@ -197,25 +193,23 @@ class SparseVanka
     /**
      * Constructor. For the parameters' description, see below.
      */
+    AdditionalData(const std::vector<bool> &selected);
```
tjhei, line 196: please mark ``explicit``

include/deal.II/lac/sparse_vanka.templates.h
```diff
@@ -113,13 +111,15 @@ SparseVanka<number>::compute_inverses()
 #else
   const size_type n_inverses =
     std::count(selected->begin(), selected->end(), true);
-
-  const size_type n_inverses_per_thread =
-    std::max(n_inverses / n_threads, static_cast<size_type>(1U));
+  // somewhat arbitrarily set up an equal number of tasks as we have threads
```
tjhei, line 114: you don't have to fix this, but I would guess having more tasks makes sense.

include/deal.II/lac/sparse_vanka.templates.h
```diff
@@ -614,38 +600,11 @@ SparseBlockVanka<number>::vmult(Vector<number2> &      dst,
...
-                              src,
-                              &dof_masks[block]);
-      threads.join_all();
-#else
-      for (unsigned int block = 0; block < n_blocks; ++block)
-        this->apply_preconditioner(dst, src, &dof_masks[block]);
-#endif
-    }
+  Threads::TaskGroup<> tasks;
+  for (unsigned int block = 0; block < n_blocks; ++block)
```
tjhei, line 604: yay, I like it!

include/deal.II/lac/sparse_vanka.templates.h
drwells: I agree, but I am leaving that for a followup PR.

include/deal.II/lac/sparse_vanka.templates.h
drwells: I agree completely but I would like to leave that for a followup PR. Is that okay?

bangerth: Yes.

include/deal.II/lac/sparse_vanka.templates.h
drwells: I believe the answer is no, since we don't apply the smoother to all dofs - the range is chopped up so that we have an even number of smoothed dofs per range.

bangerth: OK.

PR #10397, opened 2020-05-27
[Post-9.4] Use C++14 lambda capture in step-69.

bangerth:
As discussed at great length when we reviewed step-69.

/rebuild

### DIFF
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -2758,14 +2758,12 @@ namespace Step69
 
     schlieren_postprocessor.compute_schlieren(output_vector);
 
-    auto data_out = std::make_shared<DataOut<dim>>();
-
+    std::unique_ptr<DataOut<dim>> data_out = std::make_unique<DataOut<dim>>();
     data_out->attach_dof_handler(offline_data.dof_handler);
 
-    const auto &component_names = ProblemDescription<dim>::component_names;
-
     for (unsigned int i = 0; i < n_solution_variables; ++i)
-      data_out->add_data_vector(output_vector[i], component_names[i]);
+      data_out->add_data_vector(output_vector[i],
+                                ProblemDescription<dim>::component_names[i]);
 
     data_out->add_data_vector(schlieren_postprocessor.schlieren,
                               "schlieren_plot");
@@ -2778,16 +2776,31 @@ namespace Step69
     // the <code>this</code> pointer as well as most of the arguments of
     // the output function by value so that we have access to them inside
     // the lambda function.
-    const auto output_worker = [this, name, t, cycle, data_out]() {
-      DataOutBase::VtkFlags flags(t,
-                                  cycle,
-                                  true,
-                                  DataOutBase::CompressionLevel::best_speed);
-      data_out->set_flags(flags);
+    //
+    // The first capture argument of the lambda function, `data_out_copy`
+    // in essence creates a local variable inside the lambda function into
+    // which we "move" the `data_out` variable from above. The way this works
+    // is that we create a `std::unique_ptr` above that points to the DataOut
+    // object. But we have no use for it any more after this point: We really
+    // just want to move ownership from the current function to the lambda
+    // function implemented in the following few lines. We could have done
+    // this by using a `std::shared_ptr` above and giving the lambda function
+    // a copy of that shared pointer; once the current function returns (but
+    // maybe while the lambda function is still running), our local shared
+    // pointer would go out of scope and stop pointing at the actual
+    // object, at which point the lambda function simply becomes the sole
+    // owner. But using the `std::unique_ptr` is conceptually cleaner as it
+    // makes it clear that the current function's `data_out` variable isn't
+    // even pointing to the object any more.
+    auto output_worker =
+      [data_out_copy = std::move(data_out), this, name, t, cycle]() {
+        const DataOutBase::VtkFlags flags(
+          t, cycle, true, DataOutBase::CompressionLevel::best_speed);
+        data_out_copy->set_flags(flags);
 
-      data_out->write_vtu_with_pvtu_record(
-        "", name + "-solution", cycle, mpi_communicator, 6);
-    };
+        data_out_copy->write_vtu_with_pvtu_record(
+          "", name + "-solution", cycle, mpi_communicator, 6);
+      };
 
     // If the asynchronous writeback option is set we launch a new
     // background thread with the help of
@@ -2801,7 +2814,8 @@ namespace Step69
     // run in the background.
     if (asynchronous_writeback)
       {
-        background_thread_state = std::async(std::launch::async, output_worker);
+        background_thread_state =
+          std::async(std::launch::async, std::move(output_worker));
       }
     else
       {


examples/step-69/step-69.cc
```diff
@@ -2822,14 +2822,12 @@ namespace Step69
 
     schlieren_postprocessor.compute_schlieren(output_vector);
 
-    auto data_out = std::make_shared<DataOut<dim>>();
-
+    std::unique_ptr<DataOut<dim>> data_out = std::make_shared<DataOut<dim>>();
```
kronbichler, line 2825: Am I missing something or why aren't you using 
```suggestion
    std::unique_ptr<DataOut<dim>> data_out = std::make_unique<DataOut<dim>>();
```

bangerth: You didn't miss anything. It took too long to compile on the laptop and so I decided to let the CI infrastructure do it.

bangerth: In any case, fixed now.

examples/step-69/step-69.cc
```diff
@@ -2842,7 +2840,28 @@ namespace Step69
...
+    // object, at which point the lambda function simply becomes the sole
+    // owner. But using the `std::unique_ptr` is conceptually cleaner as it
+    // makes it clear that the current function's `data_out` variable isn't
+    // even pointing to the object any more.
+    const auto output_worker = [this,
+                                name,
+                                t,
+                                cycle,
+                                checkpoint,
+                                data_out_copy = std::move(data_out)]() {
```
tamiko, line 2864: It seems that gcc-5.4.0 is tripping over the move semantics.
```
/jenkins/workspace/dealii/PR-10397/examples/step-69/step-69.cc:2864:68: note: ‘Step69::MainLoop<dim>::output(const vector_type&, const string&, double, unsigned int, bool) [with int dim = 2; Step69::MainLoop<dim>::vector_type = std::array<dealii::LinearAlgebra::distributed::Vector<double>, 4ul>; std::__cxx11::string = std::__cxx11::basic_string<char>]::<lambda()>::<lambda>(const Step69::MainLoop<dim>::output(const vector_type&, const string&, double, unsigned int, bool) [with int dim = 2; Step69::MainLoop<dim>::vector_type = std::array<dealii::LinearAlgebra::distributed::Vector<double>, 4ul>; std::__cxx11::string = std::__cxx11::basic_string<char>]::<lambda()>&)’ is implicitly deleted because the default definition would be ill-formed:
                                 data_out_copy = std::move(data_out)]() {
                                                                    ^
/jenkins/workspace/dealii/PR-10397/examples/step-69/step-69.cc:2864:68: error: use of deleted function ‘std::unique_ptr<_Tp, _Dp>::unique_ptr(const std::unique_ptr<_Tp, _Dp>&) [with _Tp = dealii::DataOut<2, dealii::DoFHandler<2, 2> >; _Dp = std::default_delete<dealii::DataOut<2, dealii::DoFHandler<2, 2> > >]’
In file included from /usr/include/c++/5/bits/locale_conv.h:41:0,
                 from /usr/include/c++/5/locale:43,
                 from /usr/include/c++/5/iomanip:43,
                 from /home/dealii/libs/trilinos-release-12-10-1/include/Teuchos_ConfigDefs.hpp:89,
                 from /home/dealii/libs/trilinos-release-12-10-1/include/Teuchos_PromotionTraits.hpp:45,
                 from /home/dealii/libs/trilinos-release-12-10-1/include/Sacado_Fad_DFadTraits.hpp:142,
                 from /home/dealii/libs/trilinos-release-12-10-1/include/Sacado.hpp:51,
                 from /jenkins/workspace/dealii/PR-10397/include/deal.II/base/numbers.h:151,
                 from include/deal.II/base/config.h:453,
                 from /jenkins/workspace/dealii/PR-10397/include/deal.II/base/conditional_ostream.h:19,
                 from /jenkins/workspace/dealii/PR-10397/examples/step-69/step-69.cc:41:
/usr/include/c++/5/bits/unique_ptr.h:356:7: note: declared here
       unique_ptr(const unique_ptr&) = delete;
       ^
```

bangerth: Ugh. I'm pretty sure my code is right, though.

Also, how did they get that wrong? That was specifically one of the two points I can think of where it makes sense to use these sorts of initializers in lambda captures (capturing member variables of larger objects being the other).

bangerth: I mean my code is straight out of the book: https://stackoverflow.com/questions/25408190/lambda-captures-c14

tamiko: The code is absolutely fine - gcc-5.4.0 is broken. I have no good idea how to proceed. (I very much prefer the explicit move semantics of the unique_ptr as done in your pull request.)

tamiko: (I actually use precisely what you have in your pull request in our higher-order code together with an explicit `.reset()` in order to reliably free up the memory after writeout.)

bangerth:
I tried moving that capture to another place in the capture list, but with no effect. 

What do you all think? Should we just close as "not worth our time to find a solution with gcc 5.4"?

bangerth:
Let's just revisit this after 9.3 when we may have a different minimum GCC version.

bangerth:
I rebased this patch, but it continues to fail with gcc 5.4 and with the cuda-10 platform. I'm going to defer this to post-10.0 in hopes that we can maybe remove support for these old compiler versions.

tamiko:
It's time to get this one in :-)

masterleinad (approved):

masterleinad:
```
usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/future:1729:10: error: no matching constructor for initialization of 'std::tuple<(lambda at /home/runner/work/dealii/dealii/examples/step-69/step-69.cc:2788:7)>' [clang-diagnostic-error]
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/future:1662:10: error: no matching constructor for initialization of 'std::tuple<(lambda at /home/runner/work/dealii/dealii/examples/step-69/step-69.cc:2788:7)>' [clang-diagnostic-error]
```

bangerth:
I don't know how you noticed that I had just rebased this patch, but clearly you saw. In any case, I must have gotten things wrong while rebasing through the merge commit the patch had acquired. 

In any case, though, I cannot figure out what is going wrong with the patch. It may be that it's just too late, or that I have COVID brain, or that it's genuinely something strange the compiler is complaining about. Any help finding the issue would be appreciated.

Here's what we have:
```
    const auto output_worker =
      [data_out_copy = std::move(data_out), this, name, t, cycle]() {
        const DataOutBase::VtkFlags flags(
          t, cycle, true, DataOutBase::CompressionLevel::best_speed);
        data_out_copy->set_flags(flags);

        data_out_copy->write_vtu_with_pvtu_record(
          "", name + "-solution", cycle, mpi_communicator, 6);
      };

    // If the asynchronous writeback option is set we launch a new
    // background thread with the help of
    // <a
    // href="https://en.cppreference.com/w/cpp/thread/async"><code>std::async</code></a>
    // function. The function returns a <a
    // href="https://en.cppreference.com/w/cpp/thread/future"><code>std::future</code></a>
    // object that we can use to query the status of the background thread.
    // At this point we can return from the <code>output()</code> function
    // and resume with the time stepping in the main loop - the thread will
    // run in the background.
    if (asynchronous_writeback)
      {
        background_thread_state = std::async(std::launch::async, output_worker);      // **** error is triggered by this line
      }
```
The error happens deep in std:: headers and reads like this:
```
cons/home/fac/g/bangerth/p/deal.II/1/dealii/examples/step-69/step-69.cc:2835:20:   required from here
/scratch/local/gcc-11.2.0/include/c++/11.2.0/future:1717:11: error: could not convert ‘{std::forward<Step69::MainLoop<2>::output(const vector_type&, const string&, double, unsigned int)::<lambda()>&>((* & __args#0))}’ from ‘<brace-enclosed initializer list>’ to ‘std::tuple<Step69::MainLoop<2>::output(const vector_type&, const string&, double, unsigned int)::<lambda()> >’
 1717 |           _M_fn{{std::forward<_Args>(__args)...}}
      |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |           |
      |           <brace-enclosed initializer list>
```

Anyone see what I'm doing wrong?

sebproell:
> Anyone see what I'm doing wrong?

@bangerth The async call takes the callable as a universal reference. In your case, you pass a const reference, thus an internal copy is made at some point. Unfortunately, your callable is not copyable, since it has a `unique_ptr` member (the pointer to DataOut).

Possible solutions:
- Either make the captured data_out a shared_ptr to make it copyable, or
- Move the callable into the async call.

I prefer the latter, but the first solution is sometimes unavoidable in my experience (e.g. if the called code does not do perfect forwarding and thus moving does not help you). 

PS: I struggled with this way too often myself. It's unfortunate that the compiler error is so unhelpful in this case :(

bangerth:
@sebproell Very nice! You'd think that I've seen everything C++ related at this point, but this one stumped me! The idea of moving the lambda worked.

Of course, it's not entirely clear any more whether that is simple enough to actually do in the tutorial. It is *conceptually* the most elegant way to capture a copy of the pointer, but that doesn't always mean that we *should* do it. On the other hand, this also isn't step-4 any more, so maybe we can be more lenient about it. Thoughts appreciated.

PR #10399, opened 2020-05-28
Remove references to hp::DoFHandler in DoFAccessor.

drwells:
Obviously this won't work until we merge #10328 but I fixed these while looking at another part of that PR.


peterrum:
Thanks!

bangerth:
The documentation components of it, at least. The rest of course doesn't compile right now.

drwells:
That's probably worthwhile just so that the PR doesn't get stale. I will excise the non-documentation changes.

drwells:
Done!

drwells:
our documentation is so up-to-date that it now refers to future implementations :)

Issue #10401, opened 2020-05-28
Implement FEEvaluation::submit_hessian()


kronbichler:
We currently support `FEEvaluation::evaluate(EvaluationFlags::hessians)` and the access `FEEvaluation::get_hessian()` but not `FEEvaluation::integrate(EvaluationFlags::hessians)` and `FEEvaluation::submit_hessian()`. This should be addressed for the next release. One thing that makes things slightly more complicated is the fact that `submit_hessian()` will accumulate contributions both for the second and the first derivatives on the reference cell, the latter only for deformed non-affine element shapes. Thus, we need an additional internal data field similar to `gradients_quad` to store the queued part. As an alternative, we also could write `FEEvaluation::submit_gradient_and_hessian()` and collect the gradient contributions from both slots, but the interface would not be as nice, despite being a bit faster because we avoid the redundant internal field.

peterrum:
This would be nice. But I guess we can postpone it for now.

PR #10402, opened 2020-05-28
Remove pointers of pointers in FEEvaluation

kronbichler:
While thinking about the cleaning of internal interfaces I want to do once #10369 is finished and in preparation for #10401, I decided that I do not like the pointers-of-pointers approach in `FEEvaluation` for the temporary arrays to hold results at quadrature points in the multi-component case with values, gradients, and Hessians. While this does simplify the access to certain arrays in the `get_gradient()` functions and friends, we are not consistently using them. For example, the data of various components is assumed to be contiguous in memory in the internal interfaces, see e.g.
https://github.com/dealii/dealii/blob/a73d35cdfeae3809863fef144410ae3ef1445102/include/deal.II/matrix_free/fe_evaluation.h#L6998-L7007
or what we return to the user here
https://github.com/dealii/dealii/blob/a73d35cdfeae3809863fef144410ae3ef1445102/include/deal.II/matrix_free/fe_evaluation.h#L4654-L4663
Thus, for consistency I prefer to use the layout of addressing with `values_quad[component * n_q_points + q_index]` even though it is a bit more to write than `values_quad[component][q_index]`. The compilers tend to at least not work worse with the new code - my intuition would have said that addressing into a single base pointer with 3 different offsets rather than three base pointers with a single offset should be slightly better, but complex addressing modes in x86-64/AVX make it very similar. It is at best 1-2 fewer instructions (out of around 30) for the q-point loop in step-37.

Note that these are performance-critical code paths and memory comes from an external scratch data provider, so I cannot realistically set up more user-friendly concepts like a `Table<2,...>` or `Table<3,...>` which would be the most appropriate data structure if we could choose freely.

While there, I also tried to consistently use the argument name `q_point` for the quadrature point index, which was not only the most common name before but it is also what we use in the `FEValuesView` concept which is most closely related to the ones here:
https://github.com/dealii/dealii/blob/a73d35cdfeae3809863fef144410ae3ef1445102/include/deal.II/fe/fe_values.h#L285-L286

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -683,7 +683,7 @@ public:
    * times the quadrature weight.
    */
   VectorizedArrayType
-  JxW(const unsigned int q_index) const;
+  JxW(const unsigned int q_point) const;
 
   /**
    * Return the inverse and transposed version of Jacobian of the mapping
@@ -692,7 +692,7 @@ public:
    * the unit cell gradients to gradients on the real cell.
    */
   Tensor<2, dim, VectorizedArrayType>
-  inverse_jacobian(const unsigned int q_index) const;
+  inverse_jacobian(const unsigned int q_point) const;
 
   /**
    * Return the unit normal vector on a face. Note that both sides of a face
@@ -1042,7 +1042,7 @@ protected:
    * memory on the stack, this approach allows for very large polynomial
    * degrees.
    */
-  VectorizedArrayType *values_quad[n_components];
+  VectorizedArrayType *values_quad;
 
   /**
    * This field stores the gradients of the finite element function on
@@ -1057,7 +1057,7 @@ protected:
    * memory on the stack, this approach allows for very large polynomial
    * degrees.
    */
-  VectorizedArrayType *gradients_quad[n_components][dim];
+  VectorizedArrayType *gradients_quad;
 
   /**
    * This field stores the Hessians of the finite element function on
@@ -1070,7 +1070,7 @@ protected:
    * memory on the stack, this approach allows for very large polynomial
    * degrees.
    */
-  VectorizedArrayType *hessians_quad[n_components][(dim * (dim + 1)) / 2];
+  VectorizedArrayType *hessians_quad;
 
   /**
    * Stores the number of the quadrature formula of the present cell.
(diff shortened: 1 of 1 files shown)


kronbichler:
There is a similar cleanup to be done for the `values_dofs`, but I want to first wait if there are objections with this one before I go to the `read_write_operation` functions.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -5845,16 +5848,16 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArrayType
   AssertIndexRange(q_point, this->n_quadrature_points);
   Assert(this->jacobian != nullptr, ExcNotInitialized());
 
-  VectorizedArrayType divergence;
+  VectorizedArrayType        divergence;
+  const VectorizedArrayType *gradients = this->gradients_quad + q_point;
```
peterrum, line 5852: I must admit I don't like this optimization: it is hard to read. I would prefer below `gradients[(d * dim + e) * nqp + q_point]`. The compiler should be able to reuse the computed indices.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -5700,41 +5702,42 @@ template <int dim, typename Number, bool is_face, typename VectorizedArrayType>
...
-          this->J_value[0] * this->quadrature_weights[q_index];
+          this->J_value[q_point] :
+          this->J_value[0] * this->quadrature_weights[q_point];
       for (unsigned int d = 0; d < dim; ++d)
         {
           VectorizedArrayType new_val = jac[0][d] * grad_in[0];
           for (unsigned int e = 1; e < dim; ++e)
             new_val += jac[e][d] * grad_in[e];
-          this->gradients_quad[0][d][q_index] = new_val * JxW;
+          gradients[d * nqp] = new_val * JxW;
```
peterrum, line 5740: The same here.

include/deal.II/matrix_free/fe_evaluation.h
kronbichler: I actually started with what you suggest before but it turns out the compiler can't make the optimization because it is not equivalent as it is written now. The compiler can't know that `q_point + (d * dim + e) * nqp` does not overflow and wrap around the 32 bit integer limit, so it has to evaluate them inside the loop. This would have been avoided if we used signed integers instead of unsigned ones because there overflow is undefined, so the optimization should have been valid. I think the issue is avoided if I instead declare `npg` as `std::size_t`. Then, the expression `q_point + (d * dim + e) * nqp` will all be evaluated in 64 bit integer type and the compiler is back into business, because `ptr + offset` cannot overflow. I don't have time to try it now, but I will do so.

masterleinad:
Can you please rebase?

include/deal.II/matrix_free/fe_evaluation.h
peterrum: @kronbichler Do you still want to investigate this? Or do you want to rebase and merge?

kronbichler: I will investigate it soon I hope, but not this week.

kronbichler: This is now ready for re-review. I replaced the offsets by `size_t` and at least in the scalar case (step-37, step-48, step-59) the compiler generates good code. I'm investigating step-67 but I think we can always improve something later on given the current state.

kronbichler:
Any additional comments on this PR?

PR #10415, opened 2020-05-28
Fix compiling with gcc with C++20

masterleinad:
Fixes https://cdash.43-1.org/viewBuildError.php?buildid=6866.

### DIFF
--- a/include/deal.II/base/linear_index_iterator.h
+++ b/include/deal.II/base/linear_index_iterator.h
@@ -260,10 +260,16 @@ public:
   }
 
   /**
-   * Inverse of operator==().
+   * Opposite of operator==().
    */
-  bool
-  operator!=(const DerivedIterator &) const;
+  template <typename OtherIterator>
+  friend typename std::enable_if<
+    std::is_convertible<OtherIterator, DerivedIterator>::value,
+    bool>::type
+  operator!=(const LinearIndexIterator &left, const OtherIterator &right)
+  {
+    return !(left == right);
+  }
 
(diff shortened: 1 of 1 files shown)


include/deal.II/base/linear_index_iterator.h
```diff
@@ -262,8 +262,14 @@ class LinearIndexIterator
   /**
    * Inverse of operator==().
```
tamiko, line 263: While at it. Would you mind replacing "Inverse" by "Opposite"? :smile:

include/deal.II/base/linear_index_iterator.h
```diff
@@ -262,8 +262,14 @@ class LinearIndexIterator
...
    */
-  bool
-  operator!=(const DerivedIterator &) const;
+  template <typename OtherIterator>
+  friend typename std::enable_if<
+    std::is_convertible<OtherIterator, DerivedIterator>::value,
+    bool>::type
+  operator!=(const LinearIndexIterator &left, const OtherIterator &right)
+  {
+    return left != right;
```
tamiko, line 271: I think this cannot compile in non C++20 mode because there is no rewrite overload for `operator!=`. Alaternatively, what about we just remove the `operator!=` altogether if `DEAL_II_HAVE_CXX20` is true?

masterleinad: Yes, let's do that then.

masterleinad:
Let's see if this works.

PR #10416, opened 2020-05-29
Remove author information from source files

masterleinad:
In the hackathon, I remember the question coming up if we should get rid of all the author information. One of the reasons would be that they become outdated quickly anyway and don't provide a lot of useful information.
This pull request removes the respective information for `include` and `source` but is merely thought as a platform for discussion for what we want to do about it (if anything).


peterrum:
I am not in favor if this change. I like to connect names with functions (even if they are outdated) because it gives me an indication in which context they have been implemented (and why - as a consequence - they have some limitations).

tjhei:
I am in favor of making this change. The existence can make it seem that we are territorial. They are incomplete and as such not very useful in the first place.

Are there advantages of having them?

Rombur:
> I like to connect names with functions (even if they are outdated) because it gives me an indication in which context they have been implemented (and why - as a consequence - they have some limitations)

You can use `blame` for that and you even get the commit message.

gassmoeller:
Just to provide a data point: I just recently extended Particles::DataOut (in #10390), which was only added 5 months ago (in #9072), yet I was unsure if I should add myself to the list of authors. It felt weird to add my own name to a list of authors whom I did not ask (Luca and Bruno would have been fine with the change of course, but nevertheless, its a conceptual problem). In the end I did not add myself, so the list is already outdated.

I agree with Timo that adding `@author` comments into the code feels like a suboptimal approach to provide credit, and adding changelog entries that are published in release notes, and lists of contributors are a better form.

bangerth:
A principle that, I think, has served us rather well is to think of deal.II as the product of a collective, not of individuals. It allows someone who is, say, spending a lot of work on the CI infrastructure to get equal credit to someone who spends time on writing code and to someone who spends time on reviewing code.

I think that removing `@author` tags makes sense from this perspective.

One could think about what that means for tutorials, which often really are just developed by one person (or small numbers of people). I would favor taking the author information off from steps 1-16, for example, given that these truly are collective works where a large number of people have written and rewritten parts. I have no real opinion on the rest.

tamiko:
I agree with almost everything that was said here regarding the @author annotation of classes and functions within the library (i.e., code under `include`, and `source`). They tend to become outdated very fast and are often incorrect. (Some of these functions and classes for example have been altered to such an extend that one could legally argue that the original author doesn't hold any copyright any more).

I think that the example steps are a different story, but I agree with @bangerth that we can take a critical look at our example steps every now and then and decide which of the steps fall into the category of being a community effort and which do not.

blaisb:
I think this is a great idea. I share Rene feelings in the sense that I am always a bit uneased at modifying something that bears someone else's name...

drwells:
I think 8 is a quorum :)

PR #10421, opened 2020-05-29
Bug fix: Multiple valid CUDAWrappers::MatrixFree objects is not allowed

### LINKED ISSUE #10214: Multiple valid CUDAWrappers::MatrixFree objects is not allowed
For performance reasons, we store [shape_values](https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/cuda_matrix_free.templates.h#L57-L60) and [shape_gradients](https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/cuda_matrix_free.templates.h#L80-L83) in constant memory, i.e. in a global variable. The problem arises if you want to have two `CUDAWrappers::MatrixFree` objects that use different shape functions because we can only store one type of shape functions in the global variable. We need to think how we want to deal with this problem, right now we silently use the latest `MatrixFree` object initialized. One way to fix the problem would be to allocate larger arrays and use a static variable to split them when we create a `MatrixFree` object. The difficulty is that the variable is initialized in `MatrixFree` but used in `EvaluateTensorProduct` which is called from `FEEvaluation`. So we will need to change the interface in `FEEvaluation` to explicitly associate an `FEEvaluation` object to a `MatrixFree` object.

Rombur:
The new interface is incompatible in purpose. We could have an interface that's backward compatible but that would lead to bugs hard to debug

Fixes #10214

### DIFF
--- a/include/deal.II/base/cuda_size.h
+++ b/include/deal.II/base/cuda_size.h
@@ -38,6 +38,19 @@ namespace CUDAWrappers
    * Define the number of threads in a warp.
    */
   constexpr int warp_size = 32;
+
+  /**
+   * Define the largest finite element degree that can be solved using
+   * CUDAWrappers::MatrixFree. Changing this number will affect the amount of
+   * constant memory being used.
+   */
+  constexpr unsigned int mf_max_elem_degree = 10;
+
+  /**
+   * Define the maximum number of valid CUDAWrappers::MatrixFree object.
+   * Changing this number will affect the amount of constant memory being used.
+   */
+  constexpr unsigned int mf_n_concurrent_objects = 5;
 } // namespace CUDAWrappers
 
 DEAL_II_NAMESPACE_CLOSE
--- a/include/deal.II/matrix_free/cuda_matrix_free.templates.h
+++ b/include/deal.II/matrix_free/cuda_matrix_free.templates.h
@@ -48,80 +48,85 @@ namespace CUDAWrappers
 {
   namespace internal
   {
-    // These variables are stored in the device constant memory.
-    constexpr unsigned int max_elem_degree = 10;
+    constexpr unsigned int data_array_size =
+      (mf_max_elem_degree + 1) * (mf_max_elem_degree + 1);
+
+    // Default initialized to false
+    std::array<std::atomic_bool, mf_n_concurrent_objects> used_objects;
 
     template <typename NumberType>
-    using DataArray = NumberType[(max_elem_degree + 1) * (max_elem_degree + 1)];
+    using DataArray = NumberType[data_array_size];
 
-    __constant__ double
-      global_shape_values_d[(max_elem_degree + 1) * (max_elem_degree + 1)];
-    __constant__ float
-      global_shape_values_f[(max_elem_degree + 1) * (max_elem_degree + 1)];
+    // These variables are stored in the device constant memory.
+    // Shape values
+    __constant__ double global_shape_values_d[mf_n_concurrent_objects]
+                                             [data_array_size];
+    __constant__ float global_shape_values_f[mf_n_concurrent_objects]
+                                            [data_array_size];
+    // Shape gradients
+    __constant__ double global_shape_gradients_d[mf_n_concurrent_objects]
+                                                [data_array_size];
+    __constant__ float global_shape_gradients_f[mf_n_concurrent_objects]
+                                               [data_array_size];
+    // for collocation methods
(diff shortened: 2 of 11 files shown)


peterrum:
I'll take a look at this PR tomorrow. Out of curiosity, do you observe an performance impacts by the introduced indirection?

Rombur:
I don't know, I didn't compare the performance.

include/deal.II/matrix_free/cuda_tensor_product_kernels.h
```diff
@@ -147,6 +149,8 @@ namespace CUDAWrappers
        */
       __device__ void
       integrate_value_and_gradient(Number *u, Number *grad_u[dim]);
+
+      int mf_object_id;
```
masterleinad, line 153: Can this be `const`?

include/deal.II/matrix_free/cuda_matrix_free.templates.h
```diff
@@ -882,29 +891,40 @@ namespace CUDAWrappers
     unsigned int size_co_shape_values =
       n_q_points_1d * n_q_points_1d * sizeof(Number);
 
+    // Check if we already a part of the constant memory allocated to us. If
+    // not, we try to get a block of memory.
+    bool found_id = false;
+    while (!found_id)
+      {
+        ++my_id;
```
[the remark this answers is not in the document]
Rombur: `my_id` is initialized at -1 and when we call free we reset `my_id` to -1 so we always start from the beginning. The current code only do one pass and I don't think it is necessary to change this. `MatrixFree` objects live for a long time so there is almost not chance that we can wait for an object to get freed.

Rombur:
@peterrum Very good point, it's a much nicer interface. I'll change it.

Rombur:
I have updated the code

include/deal.II/matrix_free/cuda_matrix_free.templates.h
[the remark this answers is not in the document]
tjhei: isn't this missing an ``Assert(my_id<mf_n_concurrent_objects, ...)``?

include/deal.II/matrix_free/cuda_matrix_free.templates.h
```diff
@@ -882,29 +900,40 @@ namespace CUDAWrappers
     unsigned int size_co_shape_values =
       n_q_points_1d * n_q_points_1d * sizeof(Number);
 
+    // Check if we already a part of the constant memory allocated to us. If
+    // not, we try to get a block of memory.
+    bool found_id = false;
```
tjhei, line 905: can you set my_id to -1 here (or at least assert it)? Can internal_reinit() never happen twice?

include/deal.II/base/cuda_size.h
```diff
@@ -38,6 +38,19 @@ namespace CUDAWrappers
    * Define the number of threads in a warp.
    */
   constexpr int warp_size = 32;
+
+  /**
+   * Define the largest finite element degree that can be solved using
+   * CUDAWrappers::MatrixFree. Changing this number will affect the amount of
+   * constant memory being used.
+   */
+  constexpr unsigned int mf_max_elem_degree = 10;
```
tjhei, line 47: Is there an assert somewhere if somebody goes above this? I couldn't see it in this PR.

include/deal.II/matrix_free/cuda_matrix_free.templates.h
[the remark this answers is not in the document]
masterleinad: That is an alternative to what I wrote above.

include/deal.II/matrix_free/cuda_matrix_free.templates.h
Rombur: We set my_id to -1 earlier in the function by calling free.

include/deal.II/matrix_free/cuda_matrix_free.templates.h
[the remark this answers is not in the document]
Rombur: I'll add an assert but the code will always crash because of out of bound access on the std::array

include/deal.II/base/cuda_size.h
Rombur: The `cudaMemcpyToSymbol` will return an error code that we assert

Rombur:
I have an assert

PR #10426, opened 2020-05-31
Fix compiling step-69 with gcc-10 and C++20

masterleinad:
Fixes https://cdash.43-1.org/viewBuildError.php?onlydeltap&buildid=6942 by using generic lambdas.

### DIFF
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -1258,10 +1258,7 @@ namespace Step69
         0, n_locally_relevant);
 
       const auto on_subranges = //
-        [&](
-          std_cxx20::ranges::iota_view<unsigned int, unsigned int>::iterator i1,
-          const std_cxx20::ranges::iota_view<unsigned int,
-                                             unsigned int>::iterator i2) {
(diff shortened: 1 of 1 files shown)


tjhei:
Would it make sense to modify #10018 to unconditionally use the boost version? Paging @bangerth as he introduced this.

tamiko:
I agree that we have to fix up the situation with `iota_view`. This seems to be a bit too experimental at the moment…

Independent of the issue, I very much prefer the generic lambda variant used here. It makes the code much more readible (because it doesn't export an unnecessary implementational detail to the reader).

bangerth:
So does this fix the issue? I have no problem disabling the use of the `std::` facilities in `namespace std_cxx20` if they are not ready. We should of course check for readiness using a cmake macro.

tjhei:
> So does this fix the issue?

it avoids using ``iterator``, so this patch makes step-69 compile again. That doesn't fix the fact that it is broken in 9.2 and that ``std_cxx20`` still behaves differently if you have gcc-10. 
I think it is best to just disable for now. See #10432.

PR #10429, opened 2020-05-31
Make CellData::vertices to std::vector

peterrum:


source/grid/tria.cc
```diff
@@ -10494,8 +10500,23 @@ Triangulation<dim, spacedim>::create_triangulation(
...
+      const bool arbitray_mesh_provided =
+        std::any_of(cells.begin(), cells.end(), [](const auto &cell) {
+          return cell.vertices.size() != GeometryInfo<dim>::vertices_per_cell;
+        });
+
+      if (arbitray_mesh_provided == false)
+        {
+          internal::TriangulationImplementation::Implementation::
+            create_triangulation(v, cells, subcelldata, *this);
+        }
```
peterrum, line 10512: This is the place where it is checked if a hex-only mesh has been provided.

source/grid/tria.cc
```diff
@@ -1747,9 +1747,11 @@ namespace internal
...
                 if (!triangulation.check_for_distorted_cells)
                   {
-                    const double cell_measure =
-                      GridTools::cell_measure<1>(triangulation.vertices,
-                                                 cells[cell_no].vertices);
+                    const double cell_measure = GridTools::cell_measure<1>(
+                      triangulation.vertices,
+                      reinterpret_cast<const unsigned int(
+                          &)[GeometryInfo<1>::vertices_per_cell]>(
+                        cells[cell_no].vertices.front()));
```
bangerth, line 1754: Let's not do this. Just copy the vertices into an array of the right size, or introduce an overload of `GridTools::cell_measure`. 

Same in the other places.

bangerth: For example, you could introduce an overload that takes an `ArrayView` as argument. (Or replace the existing function -- that should work in the best of all cases without further changes.)

source/grid/tria.cc
```diff
@@ -10494,8 +10500,23 @@ Triangulation<dim, spacedim>::create_triangulation(
...
+
+      if (arbitray_mesh_provided == false)
+        {
+          internal::TriangulationImplementation::Implementation::
+            create_triangulation(v, cells, subcelldata, *this);
+        }
+      else
+        {
+          Assert(
+            false,
```
bangerth, line 10516: I think it would be easier to read if you did
```
  AssertThrow (std::any_of (...) == false, ExcMessage(...));
```
This way you don't need to create the variable.

It should be `AssertThrow` so that things read from input files are also captured.

source/grid/tria.cc
peterrum: I went with copying the values into a temporal array.

source/grid/tria.cc
peterrum: I would like to keep it this way. In a follow-up PR, I might save the bool locally and in the else statement I might replace the `Assert(Throw)`.

PR #10430, opened 2020-05-31
Work on TriaObject

peterrum:
Two minor changes (although the number of changed lines might give a different impression):

- construct via `std::initializer_list`
- remove template argument `dim`

### DIFF
--- a/include/deal.II/grid/tria_object.h
+++ b/include/deal.II/grid/tria_object.h
@@ -51,45 +51,18 @@ namespace internal
      *
      * @author Guido Kanschat, 2007
      */
-    template <int structdim>
     class TriaObject
     {
     public:
-      static const unsigned int dimension = structdim;
+      /**
+       * Constructor.
+       */
+      TriaObject(const std::initializer_list<int> &faces);
 
       /**
-       * Default constructor, setting all face indices to invalid values.
+       * Constructor.
        */
-      TriaObject();
-
-      /**
-       * Constructor for a line object with the numbers of its two end points.
-       *
-       * Throws an exception if dimension is not one.
-       */
-      TriaObject(const int i0, const int i1);
-
-      /**
-       * Constructor for a quadrilateral object with the numbers of its four
-       * lines.
-       *
-       * Throws an exception if dimension is not two.
-       */
-      TriaObject(const int i0, const int i1, const int i2, const int i3);
-
-      /**
-       * Constructor for a hexahedron object with the numbers of its six
-       * quadrilaterals.
-       *
-       * Throws an exception if dimension is not two.
-       */
-      TriaObject(const int i0,
-                 const int i1,
-                 const int i2,
-                 const int i3,
-                 const int i4,
-                 const int i5);
-
+      TriaObject(const std::initializer_list<unsigned int> &faces);
 
       /**
        * Return the index of the ith face object.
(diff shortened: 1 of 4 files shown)


include/deal.II/grid/tria_object.h
```diff
@@ -51,45 +51,18 @@ namespace internal
...
-
-      /**
-       * Constructor for a quadrilateral object with the numbers of its four
-       * lines.
-       *
-       * Throws an exception if dimension is not two.
+       * Constructor.
        */
-      TriaObject(const int i0, const int i1, const int i2, const int i3);
+      TriaObject(std::initializer_list<int> faces);
```
bangerth, line 60: I think you want to take the argument as a const reference

include/deal.II/grid/tria_object.h
```diff
@@ -186,87 +160,51 @@ namespace internal
...
 
-
-    template <int structdim>
-    inline TriaObject<structdim>::TriaObject(const int i0,
-                                             const int i1,
-                                             const int i2,
-                                             const int i3,
-                                             const int i4,
-                                             const int i5)
+    inline TriaObject::TriaObject(std::initializer_list<unsigned int> faces)
```
bangerth, line 171: same here

include/deal.II/grid/tria_object.h
```diff
@@ -186,87 +160,51 @@ namespace internal
...
-      Assert(structdim == 3, ExcImpossibleInDim(structdim));
-      faces[0] = i0;
-      faces[1] = i1;
-      faces[2] = i2;
-      faces[3] = i3;
-      faces[4] = i4;
-      faces[5] = i5;
+      this->faces.reserve(faces.size());
+      for (const auto face : faces)
+        this->faces.push_back(face);
```
bangerth, line 175: This doesn't do what you want. It assumes that the object is empty beforehand. I think what you want is
```
  this->faces = std::vector<...>(faces);
```
Extra points if you name the argument differently than the member variable :-)

source/grid/tria.cc
```diff
@@ -2054,8 +2054,8 @@ namespace internal
           for (i = needed_lines.begin(); line != triangulation.end_line();
                ++line, ++i)
             {
-              line->set(internal::TriangulationImplementation::TriaObject<1>(
-                i->first.first, i->first.second));
+              line->set(internal::TriangulationImplementation::TriaObject(
+                {i->first.first, i->first.second}));
```
bangerth, line 2058: If you get rid of `TriaObject`, you could let the `TriaAccessor::set()` function just take the `std::initializer_list` directly.

bangerth:
My preference: Address the first 3 of my comments and let's merge. Then let's address the other points separately.

peterrum:
@bangerth Your description is right. The function `TriaAccessor::set()` takes `TriaObject` and could probably a vector/initializer_list directly.

My goal would to eliminate `TriaObject` altogether and so that we can rename `TriaObjectView` to `TriaObject`.

I'll address your comments right away!

include/deal.II/grid/tria_object.h
peterrum: Actually, it does. It's the constructor; so it is empty at that stage. The code you are suggesting does not work since the types are different (`int` vs. `unsigned int`). The inconsistent usage of `int` and `unsigned int` in `Triangulation`  is a larger issue (see #10431).

bangerth:
I've got a patch for removing `TriaObject` nearly ready. But it's taking a bit longer and I'm not going to finish it today. Nor tomorrow: I have to drive down to Denver to the German Consulate.

You can either leave it to me for later, or start with the situation here:
https://github.com/bangerth/dealii/tree/triaobject

PR #10433, opened 2020-05-31
Move methods of TriaLevel, TriaFaces, and TriaObjects

peterrum:
This PR simply moves some methods of `TriaLevel`, `TriaFaces`, and `TriaObjects` to `tria.cc`. The goal is to make these classes to plain containers (with two methods `memory_consumption()` and `serialize()`).

`TriaObjects` still contains some additional functions like `next_free_single_object()`, `next_free_pair_object()`, `next_free_hex()`. I would keep them at their original place as long as they don't interfere with the generalization.


peterrum:
These functions set up the data structures as `dealii::internal::TriangulationImplementation::Implementation` needs it. A different `dealii::internal::TriangulationImplementation::Implementation` implementation will require a modified setup.

bangerth:
So you'll eventually move them into the `Implementation` classes (i.e., the policies)?

bangerth:
(Merging because I think I get where you're going, just want to understand what your goal is so I have that in mind for the next few patches...)

PR #10434, opened 2020-06-01
Replace TriaObject

peterrum:
This PR removes the old `TriaObject`, replaces its usage by the use of `std::initializer_list`, and renames `TriaObjectView` to `TriaObject`.

@bangerth Thanks for working on this! You were already at 95%!

### DIFF
--- a/include/deal.II/grid/tria_accessor.h
+++ b/include/deal.II/grid/tria_accessor.h
@@ -1629,6 +1629,12 @@ private:
   set(const std::initializer_list<int> &o) const;
 
   /**
+   * The same as above but for `unsigned int`.
+   */
+  void
+  set(const std::initializer_list<unsigned int> &o) const;
+
+  /**
    * Set the flag indicating, what <code>line_orientation()</code> will
    * return.
    *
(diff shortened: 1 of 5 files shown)


include/deal.II/grid/tria_accessor.h
```diff
@@ -1627,7 +1626,13 @@ class TriaAccessor : public TriaAccessorBase<structdim, dim, spacedim>
...
    */
   void
-  set(const dealii::internal::TriangulationImplementation::TriaObject &o) const;
+  set(const std::initializer_list<int> &o) const;
+
+  /**
+   * The same as above but for `unsigned int`.
+   */
+  void
+  set(const std::initializer_list<unsigned int> &o) const;
```
bangerth, line 1635: Ah, nice trick. I really struggled with all of the narrowing warnings and what to do about it :-)

bangerth:
Thanks for finishing this -- I just ran out of time last night :-)

include/deal.II/grid/tria_accessor.h
peterrum: I really hate this (and the previous) hack. But, this was the approach to get as few narrowing warnings as possible.

bangerth:
We can address this in a follow-up at some point where we just replace every signed integer by an unsigned one.

PR #10435, opened 2020-06-01
Improve the efficiency of the step62 assembly

dangars:
Hi @jppelteret, your suggestion improves the assembly speed of step62 by a factor of 7.

That is quite impressive. Let me know if this is what you had in mind, or there is a better way to do it.

### DIFF
--- a/doc/news/changes/minor/20200531GarciaSanchez
+++ b/doc/news/changes/minor/20200531GarciaSanchez
@@ -1,0 +1,4 @@
+Improved: The efficiency of the assembly of step-62 has been improved and now
+it is 7 times faster.
+<br>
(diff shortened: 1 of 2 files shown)


dangars:
> Yes, that's definitely a huge improvement and encapsulates what I suggested. You could probably improve this further by precomputing `alpha[m][n][k][l] * grad_phi_j[l][k]` and `beta[m][n][k][l] * grad_phi_j[l][k]`, because they are independent of the index `i` which forms the outer loop. (or grad_phi_i[m][n] * alpha[m][n][k][l]`etc., which are independent of index`j`.) However, if you do this then I think that it would be at the detriment of the code's readability. So my opinion is that what you have here forms a good compromise between performance and ease of understanding.

Thanks! I'll try it :)

PR #10436, opened 2020-06-01
Use range-based iterators in tria.cc

peterrum:


peterrum:
@masterleinad I have added the consts and successfully run the test suite locally. 

I am bit confused how to interpret the consts in this context. In many instances, the cells have been modified. Is the `TriaIterator` const but the underlying `TriaAccessor` is not const?

masterleinad:
> I am bit confused how to interpret the consts in this context. In many instances, the cells have been modified. Is the `TriaIterator` const but the underlying `TriaAccessor` is not const?

Note that, we use `cell->...` and not `cell.`, i.e. the dereferenced iterator is an iterator itself. Hence, the `const` only means that we don't modify the latter iterator but not that we can't modify the object it points to.

peterrum:
@masterleinad Thanks!

PR #10437, opened 2020-06-01
Revert "disable using c++ 20 iota-view and always use boost"

masterleinad:
As I commented in https://github.com/dealii/dealii/pull/10432#issuecomment-636555712, I believe it is correct that `std::ranges::iota_view` is fully implemented in `gcc-10` and in particular that `std::ranges::iota_view::iterator` is not required to exist (and we don't use it anymore).
To make sure that the wrapper is indeed forward-compatible we should try to use the C++20 feature if we can.

### DIFF
--- a/include/deal.II/base/std_cxx20/iota_view.h
+++ b/include/deal.II/base/std_cxx20/iota_view.h
@@ -17,9 +17,11 @@
 
 #include <deal.II/base/config.h>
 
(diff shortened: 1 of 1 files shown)


bangerth:
Well, it should test alright, but let's err on the safe side and wait for the CI (and others who might want to chime in).

/rebuild

PR #10439, opened 2020-06-01
Introduce internal::TriangulationImplementation::Policy

peterrum:
This PR adds a `Policy` to `Triangulation`. The policy encapsulates the dimension- and triangulation type-dependent way to create a triangulation, to refine it and to coarsen it. 

This change enables to switch the policy during run-time. See: 

https://github.com/dealii/dealii/blob/688377740bff862a0de9e380c33b52f706ff5bd0/source/grid/tria.cc#L11261-L11281

The reason why this PR is so long is that it is not possible to make templated methods virtual.

What exactly will be part of the policy is open and will be probably extended/refined during the next weeks. This should not be a problem since the policy is not visible to the user.

Locally, the tests pass.

### DIFF
--- a/source/grid/tria.cc
+++ b/source/grid/tria.cc
@@ -1777,6 +1777,127 @@ namespace internal
         }
     }
 
+
+
+    /**
+     * An interface for algorithms that implement Triangulation-specific tasks
+     * related to creation, refinement, and coarsening.
+     */
+    template <int dim, int spacedim>
+    class Policy
+    {
+    public:
+      /**
+       * Destructor.
+       */
+      virtual ~Policy() = default;
+
+      /**
+       * Delete children of given cell.
+       */
+      virtual void
+      delete_children(
+        Triangulation<dim, spacedim> &                        triangulation,
+        typename Triangulation<dim, spacedim>::cell_iterator &cell,
+        std::vector<unsigned int> &                           line_cell_count,
+        std::vector<unsigned int> &quad_cell_count) = 0;
+
+      /**
+       * Execute refinement.
+       */
+      virtual typename Triangulation<dim, spacedim>::DistortedCellList
+      execute_refinement(Triangulation<dim, spacedim> &triangulation,
+                         const bool check_for_distorted_cells) = 0;
+
+      /**
+       * Prevent distorted boundary cells.
+       */
+      virtual void
+      prevent_distorted_boundary_cells(
+        Triangulation<dim, spacedim> &triangulation) = 0;
+
+      /**
+       * Prepare refinement.
+       */
+      virtual void
+      prepare_refinement_dim_dependent(
+        Triangulation<dim, spacedim> &triangulation) = 0;
+
+      /**
+       * Check if coarsening is allowed for the given cell.
+       */
+      virtual bool
+      coarsening_allowed(
+        const typename Triangulation<dim, spacedim>::cell_iterator &cell) = 0;
+
+      /**
+       * A sort of virtual copy constructor, this function returns a copy of
+       * the policy object. Derived classes need to override the function here
+       * in this base class and return an object of the same type as the derived
+       * class.
+       */
+      virtual std::unique_ptr<Policy<dim, spacedim>>
+      clone() = 0;
+    };
+
+
+
+    /**
+     * A simple implementation of the interface Policy. It simply delegates the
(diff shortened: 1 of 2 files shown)


peterrum:
Any opinions on this PR? In the meantime, I have created a new policy (for TETs). I would like to open the PR in the next days.

For details, see also: https://github.com/peterrum/dealii/blob/c6dc588e7b49e291bf3ab0ac9017e02d051ca6bb/source/grid/tria.cc#L11501-L11717

source/grid/tria.cc
```diff
@@ -1760,6 +1760,448 @@ namespace internal
...
 
+
+
+    /**
+     * An interface for algorithms that implement Triangulation-specific tasks
+     * related to creation, refinement, and coarsening.
+     *
+     * @note It is necessary to write out the functions for all possible
+     *   dim/spacedim combinations here to be able to make the functions
+     *   virtual.
```
kronbichler, line 1771: Can you explain why we need to write all the functions and not just use the `dim, spacedim` parameters from the class in the `create_triangulation`? At least in the implementations right now and https://github.com/peterrum/dealii/blob/c6dc588e7b49e291bf3ab0ac9017e02d051ca6bb/source/grid/tria.cc I do not see what would not work with that approach. But there are lots of convolutions in that file, so I could easily miss it.

peterrum: I am not completely sure if I understand your question correctly. My goal is to make the function virtual so that I can switch policies (for creating the coarse grid and/or the refinement strategy). To be able make the functions vitrual, the functions cannot be templated. One way to do that would be to write out all possible dim-spacedim combinations as I did. Alternatively, I could use the template arguments of the Policy:

https://github.com/dealii/dealii/blob/fbadd26068f8e95c6100a2f7b23ae538be7e409c/source/grid/tria.cc#L1773-L1774

i.e.: write:
```cpp
     void
     Policy<dim, spacedim>::create_triangulation(const std::vector<Point<spacedim>> &   v,
                           const std::vector<CellData<dim>> &cells,
                           const SubCellData &             subcelldata,
                           Triangulation<dim, spacedim> &           triangulation);
```

But to be able to do that we would need to merge the different template specializations (for 1D, 2D, 3D).

The problem in merging the `create_triangulation()` functions (as they are now) to a single huge function is that we get `TriaAccessors`, which might be invalid (e.g. `structdim=2,dim=2,spacedim=1`) and might not have certain needed functions like `set_used_flag`, `at_boundary`, and many more.... 

The other function have problems due to `RefinementCase`, which is templated, and some other functions that do not exist for certain dimensions.

kronbichler: I was thinking of your second way with the `dim` and `spacedim` parameters in `create_triangulation`. My idea would be to create specializations of `PolicyWrapper` (for the quad/hex case) for `1,spacedim`, `2,spacedim` and `3,spacedim`. Since this class only implements the parts that are actually different between `dim=1,2,3` and not the other part of `Triangulation` (which prevent the use of partial specializations in that class), I would have thought that one can use such a setup. As to refinement case, I think one could find similar solutions, but that's why I'm asking: I don't know how much of the code you would like to move into these functions and where things start to collide. All I was pointing out was that using the `dim,spacedim` parameter on a class and then provide functions which belong to a different dim/spacedim felt unnatural. In the worst case, we could have free functions to specialize.

peterrum: > In the worst case, we could have free functions to specialize.

As far as I remember free functions will not work because the static methods access private fields of `Triangulation` (enables by the `friend` key word).

peterrum:
With PR #10631, this PR is partly outdated: the method `create_triangulation` is generalized so has not to be part of the policy. However, refinement/coarsening has to be part of a policy.

kronbichler:
I suggest we wait for #10631 to be finished and revisit what is needed here. I still think that everything should fit into a class with maybe some specializations, without having to duplicate the functions for different dimensions, but let's have a look at things once we have removed the other code so we can focus on what gets necessary.

source/grid/tria.cc
peterrum: @kronbichler I was able to significantly simplify the code. Let's merge this. I have 1-2 follow-up PRs in the pipeline.

kronbichler (approved):

PR #10448, opened 2020-06-02
Get rid of TriaObject.

bangerth:
TriaObject was a class that was a barely disguised ArrayView, but the class
name wouldn't have told you that. It's not actually very difficult to replace
it by ArrayView, which is what this patch does.

In the process, rename the TriaObjects::get_object() function to
TriaObjects::get_bounding_object_indices() which is both a more
descriptive name of what it returns, and matches the recently
introduced set_bounding_object_indices() function.

/rebuild


source/grid/tria_accessor.cc
```diff
@@ -1489,19 +1489,39 @@ const unsigned int
...
-  this->objects().get_object(this->present_index) = object;
+  const ArrayView<int> bounding_object_index_ref =
+    this->objects().get_bounding_object_indices(this->present_index);
+
+  AssertDimension(bounding_object_index_ref.size(), new_indices.size());
+
+  unsigned int i = 0;
+  for (const auto &new_index : new_indices)
+    {
+      bounding_object_index_ref[i] = new_index;
```
masterleinad, line 1502: You could just combine this with the next line to
```suggestion
      bounding_object_index_ref[i++] = new_index;
```
:wink:

bangerth: We've generally tried to use indexing with side effects. It's hard to read this kind of code unless you're a hardcore C programmer. Most people also don't know/remember the difference between `i++` and `++i`, and so have to think harder about the correctness of the code than is necessary if the index is incremented in a separate statement.

masterleinad: Yes, I know. I was not really serious about suggesting this as a change.

peterrum:
@bangerth @masterleinad Do we really need a  incompatible change entry for this? The file only contains internal data structures!?

bangerth:
I've added a changelog entry.

PR #10453, opened 2020-06-03
Fix distribution of levels in pft

### LINKED ISSUE #10449: Memory consumption of p::f::t
Here is what I get by calling `tria.memory_consumption` for p::f::t:

```
  #proc	Tria	DoFHandler
  0	71594184	114275880
  1	7068584		11769688
  2	7762078		12664680
  3	7548090		12059048
```

I wonder why is there such an imbalance for rank = 0?

For reference, here is p::s::t result:

```
  #proc	Tria	DoFHandler
  0	67860024	114275888
  1	67860024	114275888
  2	67860024	114275888
  3	67860024	114275888
```

peterrum:
closes #10449.


tests/fullydistributed_grids/memory_consumption_01.mpirun=5.output
```diff
@@ -0,0 +1,14 @@
+
+DEAL:0:2d::93274
+
+DEAL:1:2d::121820
```
kronbichler, line 4: We have made the experience that the output of the memory consumption as a plain number is a pain to maintain because the result depends on the number type (32 vs 64 bit integers), changes somewhere in the code that lead to different allocated sizes, or also different implementations of the STL library in terms of how to dynamically grow vectors. Can we find a way to approximately control this number, say, by checking that it is in the right range (and assert that this is the case at least for 32 and 64 bit builds)?

tests/fullydistributed_grids/memory_consumption_01.cc
```diff
@@ -0,0 +1,82 @@
...
+      basetria, comm);
+
+  // actually create triangulation
+  tria_pft.create_triangulation(construction_data);
+
+  auto min_max_avg =
+    Utilities::MPI::min_max_avg(tria_pft.memory_consumption(), comm);
+
+  Assert((min_max_avg.min / min_max_avg.max >= 0.75),
+         ExcMessage("Memory consumption difference is too big!"));
```
peterrum, line 64: @kronbichler Good point! What about this solution?

tests/fullydistributed_grids/memory_consumption_01.cc
```diff
@@ -0,0 +1,82 @@
...
+
+  deallog << "OK!" << std::endl;
+}
+
+int
+main(int argc, char *argv[])
+{
+  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+  MPILogInitAll                    all;
```
kronbichler, line 73: Since we do not need to the specific output any more, how about:
```suggestion
  mpi_initlog();
```

peterrum: Done!

agrayver:
Can this be merged?

peterrum:
I don't want to merge my own PR! Anyone?

PR #10454, opened 2020-06-04
Move some class declarations into the .h file.

### LINKED ISSUE #10284: Bug in AffineConstraintsData

bangerth:
Next step in my quest to fix #10284. I need to create one of these cache objects in the `AffineConstraints` class, so the cache class (and everything that is referenced in it) needs to be in the `.h` file.

No functional changes, just moving stuff from one file to another.

/rebuild


bangerth:
The one test that has not completed appears spurious. All the other tests succeed.

masterleinad:
Just for clarification:
You are saying that you need to create `Cache` objects directly in `affine_constraints.h` and not just in `affine_constraints.templates.h` or `affine_constraints.cc`?

bangerth:
@drwells Correct. We currently have one global such object, declared and defined in the `.templates.h` file. I want to make this object a member variable of `AffineConstraints`, which -- being declared in the `.h` file -- requires me to move these classes to the `.h` file.

drwells:
I think you meant to @masterleinad there.

PR #10455, opened 2020-06-04
Fix clang-10 bug in grid/tria.cc

masterleinad:
Fixes the test failures in https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=7081.
Using `stdL::initializer_list` here results in `stack-use-after-scope` (as detected by `AddressSanitizer`).
Also, beautify the documentation a bit.

### DIFF
--- a/source/grid/tria.cc
+++ b/source/grid/tria.cc
@@ -2701,7 +2701,7 @@ namespace internal
 
       /**
        * Invent an object which compares two
-       * std::initializer_list objects against each
+       * std::vector objects against each
        * other. This comparison is needed in order to establish a map of
        * vertex index tuples to iterators in the
        * Triangulation<3,3>::create_triangulation function.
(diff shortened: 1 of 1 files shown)


source/grid/tria.cc
```diff
@@ -3014,80 +3013,38 @@ namespace internal
...
-                // exchanging lines. we do
-                // so here, just to check
-                // that that flipped quad
-                // isn't already in the
-                // triangulation. if it is,
-                // then don't insert the
-                // new one and instead
-                // later set the
-                // face_orientation flag
-                const std::initializer_list<int> test_quad_1(
```
tamiko, line 3055: Wow. The use after scope issue is very hard to spot...

bangerth: I didn't actually understand what my original bug was. Can you explain where the code was wrong?

tjhei: The temporary return value of the functions goes out of scope and now they are copied into the vector?

bangerth: Oh, you mean because `quad` is `const`, its `operator[]` returns by value?

Yes, tricky. I guess the lesson here is that we can really only use `std::initializer_list` as reference arguments bound to temporaries, rather than as local variables. It's a lesson I guess I will have to learn.

Issue #10175, opened 2020-05-13
Bug on find_active_cell_around_point for p::d::t.


luca-heltai:
PR #10174 reminded us that there is a problem in the fast algorithm, so we had to resort to the slow function
```
          return find_active_cell_around_point(mapping,
                                               mesh,
                                               p,
                                               marked_vertices);
```
in case the initial search did not succeed. This should be investigated and fixed.

kronbichler:
Adjusting to the 9.3 milestone now that we have the temporary solution.

luca-heltai:
So, this has been around for a long time: #7576. I think it is time we think of a permanent solution... (see also #10459, and #10506)

nfehn:
> So, this has been around for a long time: #7576. I think it is time we think of a permanent solution... (see also #10459, and #10506)

@luca-heltai what is the relation to #7930, which you did not mention above?

nfehn:
The problem mentioned above will be solved by PR #10540

bangerth:
@nfehn You mention that this should have been fixed by a patch a couple of years ago. Can you confirm that this is so?

kronbichler:
I think we made several improvements during the last release. However, in https://github.com/exadg/exadg/issues/438 we observed another problematic case. I am not sure if we can fix this entirely for the upcoming 9.5 release, but I think we also need to look over the interfaces, because there are too many functions doing similar things in `GridTools`. Let us keep this on the milestone for now.

kronbichler:
I'm closing this issue, given the progress in #15233. We might have additional bugs or optimizations, but we should make those based on up-to-date test cases and/or code restructuring.

Issue #7930, opened 2019-04-17
Bug in find_active_cell_around_point


luca-heltai:
GridTools::find_active_cell_around_point<dim>(tria,p,marked_vertices) fails in a case like this

![](https://gitlab.com/MDLSoft/ginestra-fem/uploads/209de5a254859cb08d9a5413c8707b20/image.png)

because if marked_vertices is not empty the algorithm does the following:

- searches the closes vertex (V5 in this case)
- searches only between the cells adjacent to the vertex, not finding the point

If marked_vertices is empty the algorithm searches also the other cells.

bangerth:
If I see this right, if we don't find a cell we simply return an empty iterator object, right? If so, should the function simply call itself without marked vertices if that is about to happen?

kronbichler:
I think this is what was eventually fixed in #15233 by @richardschu with help of @jh66637. So I feel this has deserved to be closed.

Issue #7576, opened 2019-01-09
Behaviour of find active cell around point with artificial cells


GivAlz:
Hi everyone,
  I am sorry to have noticed this late!

== SHORT VERSION ==

If GridTools::find_active_cell_around_point is called with a distributed triangulation, on a point lying inside an artificial cell, it now throws an exception. In the past it would return the artificial cell. Did anyone decide this, or is it a side effect of something else?  ( I personally believe it happened after introducing the search for the cell using rtrees ).

I believe this is a bug as some functions now have problems because of this behaviour (e.g. https://github.com/dealii/dealii/issues/7532 is caused by this).

== ADDITIONAL PART ==

As I discussed in the mailing list, in the past "find active cell around point" would also return artificial cells, now it doesn't. Unfortunately some functions relied on the possibility of discarding points lying on artificial cells: this is why with certain configurations NonMatching::coupling_matrix_functions do not work, and it's why the GridTools::distribtued_compute_point_locations went haywire (sorry @masterleinad : I see it was you who found a way to fix the problem, while it should have really been me...then maybe I'd found out about this sooner).

I was going to propose a PR to modify GridTools::compute_point_locations so that ignores completely points which are not found, documenting this change and leaving the "responsibility" to the user (currently what happens if a point "is not found" is not documented), but @luca-heltai  pointed out it would be nice to separate between "point inside mesh but on artificial cell" and "point not found. Anywhere.".

All I in all I believe something about GridTools::find_active_cell_around_point and/or GridTools::compute_point_locations has to be done.

What do you think? Shall we make it so the rtree returns also coarse cells?

Best,
Giovanni

P.S. Here is a simple test which currently crashes because of this exception (call it with -np 4 , for example), but would behave well with deal.ii version 9.0 .

```
#include <deal.II/base/quadrature_lib.h>

#include <deal.II/dofs/dof_accessor.h>
#include <deal.II/dofs/dof_handler.h>
#include <deal.II/dofs/dof_tools.h>

#include <deal.II/fe/fe_q.h>
#include <deal.II/fe/fe_values.h>
#include <deal.II/fe/mapping_q.h>

#include <deal.II/grid/grid_generator.h>
#include <deal.II/grid/grid_tools.h>
#include <deal.II/grid/grid_tools_cache.h>
#include <deal.II/grid/tria.h>

#include <deal.II/distributed/grid_refinement.h>
#include <deal.II/distributed/tria.h>

[...]

PR #10456, opened 2020-06-04
Fix bug in find_active_cell_around_point_tolerance()

nfehn:
In the current code version, the code did not check whether the cell is locally owned. This leads to the problem that `Mapping::transform_real_to_unit_cell()` fails (probably because wrong geometry data is used to generate a triangulation which then leads to invalid cell measure).

### DIFF
--- a/source/grid/grid_tools_dof_handlers.cc
+++ b/source/grid/grid_tools_dof_handlers.cc
@@ -465,45 +465,47 @@ namespace GridTools
             endc = adjacent_cells.end();
           for (; cell != endc; ++cell)
             {
-              try
+              if ((*cell)->is_locally_owned())
                 {
-                  const Point<dim> p_cell =
-                    mapping.transform_real_to_unit_cell(*cell, p);
-
-                  // calculate the infinity norm of
-                  // the distance vector to the unit cell.
-                  const double dist =
-                    GeometryInfo<dim>::distance_to_unit_cell(p_cell);
-
-                  // We compare if the point is inside the
-                  // unit cell (or at least not too far
-                  // outside). If it is, it is also checked
-                  // that the cell has a more refined state
-                  if ((dist < best_distance) ||
-                      ((dist == best_distance) &&
-                       ((*cell)->level() > best_level)))
+                  try
                     {
-                      found         = true;
-                      best_distance = dist;
-                      best_level    = (*cell)->level();
-                      best_cell     = std::make_pair(*cell, p_cell);
+                      const Point<dim> p_cell =
+                        mapping.transform_real_to_unit_cell(*cell, p);
+
+                      // calculate the infinity norm of
+                      // the distance vector to the unit cell.
+                      const double dist =
+                        GeometryInfo<dim>::distance_to_unit_cell(p_cell);
+
+                      // We compare if the point is inside the
+                      // unit cell (or at least not too far
+                      // outside). If it is, it is also checked
+                      // that the cell has a more refined state
+                      if ((dist < best_distance) ||
+                          ((dist == best_distance) &&
+                           ((*cell)->level() > best_level)))
+                        {
+                          found         = true;
+                          best_distance = dist;
+                          best_level    = (*cell)->level();
+                          best_cell     = std::make_pair(*cell, p_cell);
+                        }
                     }
-                }
-              catch (
-                typename MappingQGeneric<dim, spacedim>::ExcTransformationFailed
-                  &)
-                {
-                  // ok, the transformation
-                  // failed presumably
-                  // because the point we
-                  // are looking for lies
-                  // outside the current
-                  // cell. this means that
-                  // the current cell can't
-                  // be the cell around the
-                  // point, so just ignore
-                  // this cell and move on
-                  // to the next
+                  catch (typename MappingQGeneric<dim, spacedim>::
+                           ExcTransformationFailed &)
+                    {
+                      // ok, the transformation
+                      // failed presumably
+                      // because the point we
+                      // are looking for lies
+                      // outside the current
+                      // cell. this means that
+                      // the current cell can't
+                      // be the cell around the
+                      // point, so just ignore
+                      // this cell and move on
+                      // to the next
+                    }
                 }
             }
 
--- a/source/grid/grid_tools_dof_handlers.cc
+++ b/source/grid/grid_tools_dof_handlers.cc
@@ -465,7 +465,7 @@ namespace GridTools
             endc = adjacent_cells.end();
           for (; cell != endc; ++cell)
             {
-              if ((*cell)->is_locally_owned())
+              if ((*cell)->is_artificial() == false)
                 {
                   try
                     {


source/grid/grid_tools_dof_handlers.cc
```diff
@@ -465,45 +465,47 @@ namespace GridTools
             endc = adjacent_cells.end();
           for (; cell != endc; ++cell)
             {
-              try
+              if ((*cell)->is_locally_owned())
```
masterleinad, line 468: I am not quite sure why we should exclude ghost cells.

nfehn: I would say those points are found by other processors

kronbichler: The first question I could not answer in a short search is whether we actually exclude artificial cells. That should be the bare minimum to request here because all mappings that read data from solution vectors would fail. Then the point is whether to include ghosts or not - depends on the context, there certainly are cases where one would like to exclude ghosts, but I am not sure if that is unconditionally true.

kronbichler:
@nfehn can you explain to me what exactly you mean by this:

>  This leads to the problem that Mapping::transform_real_to_unit_cell() fails (probably because wrong geometry data is used to generate a triangulation which then leads to invalid cell measure).

The code is written that it should catch invalid transformations, so I would like to understand what exactly fails. Is it an exception that is not the usual `ExcTransformationFailed`?

source/grid/grid_tools_dof_handlers.cc
peterrum: Indeed the problem are the artificial cells. So, I would suggest:

```suggestion
              if((*cell)->is_artificial() == false)
````

Or, alternatively: we could amend `find_cells_adjacent_to_vertex` so that it does not return artificial cells.  Any preferences?

Comment (nfehn) [reply source/grid/grid_tools_dof_handlers.cc:468]: The documentation specifies that the function may return ghosted cells, so we should probably adhere to this specification and use ``is_artificial() == false``-

Comment (tjhei) [reply source/grid/grid_tools_dof_handlers.cc:468]: > there certainly are cases where one would like to exclude ghosts, but I am not sure if that is unconditionally true.

One problem with ignoring ghost cells is that due to tolerances, two processors at a cell boundary where you query might decide that the point is outside the own cell. If you ignore ghost cells, you might have no way to locate a cell.

**Comment by nfehn** [comment]
I used ``MappingQCache`` and the problem occurred within a rather complex application, so it would probably cost days to setup a unit test. I do not have enough experience to know exactly which data structures i have to touch in order to trigger this when starting from a simple setup.

**Comment by nfehn** [comment]
@kronbichler  Could you make a suggestion how the documentation should look like regarding artificial cells, it currently says "the cell returned may in fact be a ghost or artificial cell (see GlossArtificialCell and GlossGhostCell)", see issue #10459. I think we should update the documentation according to the changes made here and maybe explicitly say a word about artificial cells.

**Comment by kronbichler** [comment]
I guess we can skip the test for now - querying artificial cells can go very wrong if we have mappings that depend on the solution, so this feels like a safe change.
Regarding the documentation and #10459, I suggest to remove the two parts that talk about artificial cells in the return value and also remove the part "decide what to do". Instead you could say that both locally owned or ghost cells can be returned by this function for parallel triangulations.

**Comment by nfehn** [comment]
I would suggest to fix the docu in a separate PR to not get conflicts with the other PRs that also change the docu of this function (we have a separate issue for this #10459).

**Comment by kronbichler** [comment]
Sure, we can do the documentation elsewhere. I just wanted to comment on the general intent.

**Comment by kronbichler** [comment]
Let us merge this now. We might have an eye on the regression testers for the (unlikely) case that something depended on artificial cells that is not covered by the CI testers.

**Comment by nfehn** [comment]
> We need to clean up a couple of MPI tests that use find_active_cells_around_point() and are now failing at least for clang-10 with libc++, see https://cdash.43-1.org/viewTest.php?onlydelta&buildid=7263`

I think this is the relevant PR, or are there failing tests that use `find_all_active_...()`?

**Comment by peterrum** [comment]
Just run all tests, which contain the words `find_all_active_cells_around_point` or `find_active_cell_around_point` to be on the safe side. The relevant tests should be:

```
bits/find_cell_10.cc
bits/find_cell_1.cc
bits/find_cell_11.cc
bits/find_cell_12.cc
bits/find_cell_13.cc
bits/find_cell_2.cc
bits/find_cell_3.cc
bits/find_cell_4.cc
bits/find_cell_5.cc
bits/find_cell_6.cc
bits/find_cell_7.cc
bits/find_cell_8.cc
bits/find_cell_9.cc
bits/find_cell_alt_1.cc
bits/find_cell_alt_2.cc
bits/find_cell_alt_3.cc
bits/find_cell_alt_4.cc
bits/find_cell_alt_5.cc
bits/find_cell_alt_6.cc
bits/find_cell_alt_7.cc
fe/fe_nedelec_singularity_01.cc
fe/fe_nedelec_singularity_02.cc
grid/compute_point_locations_02.cc
grid/distributed_compute_point_locations_02.cc
grid/find_active_cell_around_point_01.cc
grid/find_active_cell_around_point_02.cc
grid/find_active_cell_around_point_03.cc
grid/grid_transform_02.cc
grid/grid_transform_03.cc
grid/find_active_cell_around_point_04.cc
grid/find_all_active_cells_around_point_01_b.cc
mpi/muelu_periodicity.cc
mpi/fe_field_function_02.cc
mpi/periodicity_03.cc
mpi/periodicity_02.cc
mpi/find_active_cell_around_point_01.cc
mpi/periodicity_01.cc
particles/particle_handler_08.cc
particles/particle_handler_serial_01.cc
particles/particle_handler_serial_02.cc
particles/data_out_02.cc
particles/particle_handler_01.cc
particles/particle_handler_05.cc
particles/particle_handler_02.cc
particles/data_out_03.cc
particles/data_out_04.cc
serialization/particle_handler_01.cc
serialization/particle_handler_02.cc
grid/find_all_active_cells_around_point_01.cc
grid/find_all_active_cells_around_point_02.cc
grid/find_all_active_cells_around_point_03.cc
grid/find_all_active_cells_around_point_01_b.cc
mappings/mapping_fe_field_08.cc
```

nfehn:
My point is those that are failing do not contain `find_all_active_...()`, so do you have contradictory results?

nfehn:
I realized that most of these tests do not contain a `try`-block around the function call. The question is, however, whether this behavior should be changed anyway. In my opinion it is not a good design that the function throws if no point is found (because the try-blocks are not used in practice). I would prefer a return type of `std::vector`, and an empty vector indicates that no point has been found, as do the `find_all_...()`-type functions introduced later.

There are other points that are not clear to me: There are several issues on the topic of fixing bugs for `find_active_cell_around_point()`, issues #10175, #7930, #7576. And other issues deal with the unclear aspect of how to deal with artificial cells for `GridTools` functions as well. Especially #7576 seems to be closely related to the recent issue #10459. The documentation of `find_active_cells_around_point()` is highly contradictory, here is a summary of some statements

>  If, however, the function throws an exception, then that would imply that the point lies outside locally owned active cells.

> If the point requested does not lie in any of the cells of the mesh given, then this function throws an exception

> the cell returned may in fact be a ghost or artificial cell

> One of them will then return a locally owned cell and the other one a ghost cell. 

I suggest to use issue #10459 to clarify these points and agree on a concept. For me, it currently does not make sense to try to fix tests given that I cannot know what the expected behavior of this function should be.

PR #10457, opened 2020-06-04
add tolerance to BoundingBox::point_inside()

nfehn:
Provide a user defined tolerance to make the code robust w.r.t. numerical rounding errors related to the point position.

### DIFF
--- a/include/deal.II/base/bounding_box.h
+++ b/include/deal.II/base/bounding_box.h
@@ -182,7 +182,10 @@ public:
   merge_with(const BoundingBox<spacedim, Number> &other_bbox);
 
   /**
-   * Return true if the point is inside the Bounding Box, false otherwise.
+   * Return true if the point is inside the Bounding Box, false otherwise. The
+   * parameter @p tolerance is a factor by which the bounding box is enlarged
+   * relative to the dimensions of the bounding box in order to determine in a
+   * numerically robust way whether the point is inside.
    */
   bool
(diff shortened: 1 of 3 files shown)


drwells:
Nice. Could you add a test?

include/deal.II/base/bounding_box.h
```diff
@@ -185,7 +185,9 @@ class BoundingBox
    * Return true if the point is inside the Bounding Box, false otherwise.
    */
   bool
-  point_inside(const Point<spacedim, Number> &p) const;
+  point_inside(
+    const Point<spacedim, Number> &p,
+    const double tolerance = std::numeric_limits<Number>::epsilon()) const;
```
bangerth, line 190: Can you document what the tolerance is, and maybe also link to the corresponding parameter in `GeometryInfo::is_inside_unit_cell()`.

nfehn: I added docu and improved the implementation. In the previous implementation, the reference length scale multiplied by machine eps might be zero. The reference length scale should instead always be larger than zero and of the order of the dimensions of the bounding box, i.e.,  abs(right-left). Since this is irrelevant for the unit cell with left=0, right=1, I did not refer to `is_inside_unit_cell()`.

nfehn:
I also added tests

bangerth:
The one failure seems spurious.

PR #10460, opened 2020-06-05
Next step in fixing AffineConstraints.

bangerth:
Next step in #10284. The only thing that's happening here is that I'm not initializing the accessor object implicitly using a global variable, but by explicitly providing that global variable.

The next step will be to make that variable a member variable of the `AffineConstraints` class.

/rebuild


bangerth:
Superseded by #10473.

PR #10463, opened 2020-06-05
Fix Trilinos::SparseMatrix::clear_row 2nd arg.

dougshidong:
It was previously never setting the diagonal to 'new_diag_value' unless
the original matrix had a diagonal entry of 0.0.

Not exactly sure why those two additional checks are there.


drwells:
This looks like a bug. Thanks a lot for the fix!

/rebuild

dougshidong:
Quick question. I see that there is already a test for this second argument (which passes even with the bug present).

https://github.com/dealii/dealii/blob/master/tests/trilinos/69.cc

It looks like I can run it with
ctest -R trilinos/69.debug -V

However, I don't see the test being run with the TravisCI. Note that it now fails since the previous output
https://github.com/dealii/dealii/blob/master/tests/trilinos/69.output#L4
was wrong.

Why is the test not run with TravisCI?

drwells:
The whole test site has about 6000 tests so we don’t run the whole thing for each PR.

drwells:
Since you can run the test locally, would you be willing to update the output and add it to this PR?

dougshidong:
> The whole test site has about 6000 tests so we don’t run the whole thing for each PR.

For some reason, I thought that deal.II's testsuite was programmed to only tests affected by the PR's code change through some kind of header detection. Would be nice though.

> Since you can run the test locally, would you be willing to update the output and add it to this PR?

With pleasure. I added an assert statement that should be more rigorous since it checks the diagonal value.

drwells:
> For some reason, I thought that deal.II's testsuite was programmed to only tests affected by the PR's code change through some kind of header detection. Would be nice though.

That would be an outstanding feature to have but we don't have it yet. Besides, changes to fundamental headers (like `Tensor`) would require rerunning the whole test suite, which is, at this point, just too expensive to have.

tests/trilinos/69.cc
```diff
@@ -84,10 +84,14 @@ test(TrilinosWrappers::SparseMatrix &m)
       const double s = m.el(N / 3, i);
       norm_sqr -= s * s;
     }
-  norm_sqr += 2 * rnd * rnd;
+  norm_sqr += 2 * diag * diag;
 
-  const types::global_dof_index rows[2] = {N / 3, N / 2};
-  m.clear_rows(std::vector<types::global_dof_index>(&rows[0], &rows[2]), rnd);
+  const std::vector<types::global_dof_index> rows = {N / 3, N / 2};
+  m.clear_rows(std::vector<types::global_dof_index>(&rows[0], &rows[2]), diag);
```
drwells, line 90:
```suggestion
  m.clear_rows(rows, diag);
```

dougshidong:
Woops, I thought my last force-push included the changes. Must've amended the commit without adding the changes.

Should be OK now!

bangerth:
Thanks!

PR #10465, opened 2020-06-05
Remove unneeded template argument

peterrum:
... since `structdim=3` also implies that `dim=3`.

I have skipped one instance, since it is modified in #10464.


drwells:
I am surprised to see that quite a few tests now fail - can you take a look?

peterrum:
> I am surprised to see that quite a few tests now fail - can you take a look?

@drwells I forgot to eliminate a `dim` at one place... But, that should be fixed now!

include/deal.II/grid/tria_accessor.templates.h
```diff
@@ -662,10 +657,8 @@ namespace internal
 
 
 
-      template <int dim, int spacedim>
       inline static bool
-      face_flip(const TriaAccessor<3, dim, spacedim> &accessor,
-                const unsigned int                    face)
+      face_flip(const TriaAccessor<3, 3, 3> &accessor, const unsigned int face)
```
drwells, line 661: :+1:

bangerth:
The existing infrastructure would have worked in 4d as well. But I suspect that getting 4d implemented will take another 20 years (that's how long the triangle stuff seems to be taking :-) ) and so I don't think we should worry about it :-)

peterrum:
> But I suspect that getting 4d implemented will take another 20 years

I was thinking about comparing the necessary work for extending deal.II to higher dimensions and to TET elements in my presentation at the workshop last week. My conclusion would have been that the latter is significantly more easy ;)

bangerth:
But that's because in hyper.deal you chose a very particular extension to higher dimensions, and because you haven't implemented the FEs for triangles yet :-)

PR #10469, opened 2020-06-05
Fixed default parameters for step-70.

luca-heltai:
In case we make a point release 9.2.x


luca-heltai:
Notice that I'm opening this against dealii-9.2. Should we make a new 9.2.1 branch, just in case?

bangerth:
Releases are just tagged commits on branches. So I *think* that this is the right approach.

tamiko:
> Notice that I'm opening this against dealii-9.2. Should we make a new
> 9.2.1 branch, just in case?

No, there is no necessity for another branch. Let's just push this on
the dealii-9.2 branch (as we have done in the last 5 years). :-)

PR #10470, opened 2020-06-05
Fix more issues found by AddressSanitizer

masterleinad:


masterleinad:
Some more information:
- We need to clear the vector of pointers in the move assignment operator and the move constructor. Otherwise, the `SmartPointer` will not unsubscribe when it is destroyed and the destructor still tries to signalize that the subscribed `SmartPointer`s are no longer valid (even if they don't exist anymore).
- In `IndexSet::is_ascending_and_one_to_one`, we try to find the first index that is not equal to `types::invalid_dof_index`. If the `IndexSet` is empty, this results in accessing memory out-of-bounds.
- In `GridTools::compute_local_to_global_vertex_index_map`, we call `MPI_Exscan` with `DEAL_II_VERTEX_INDEX_MPI_TYPE` which is always a 64bit type. In order to not to copy into memory not allocated, we need to make sure that the receive buffer type is also a 64bit type.

source/grid/grid_tools.cc
```diff
@@ -2266,7 +2266,7 @@ namespace GridTools
 
     // Make indices global by getting the number of vertices owned by each
     // processors and shifting the indices accordingly
-    types::global_dof_index shift = 0;
+    types::global_vertex_index shift = 0;
```
bangerth, line 2269: Good catch!

PR #10477, opened 2020-06-07
refactoring of GridTools::find_all_active_cells_around_point()

nfehn:
There are several functions ``find_active_cell_around_point()`` with different parameters. The function ``find_all_active_cells_around_point()`` is currently implemented in a way that it finds a first cell by calling one particular function ``find_active_cell_around_point()``, and subsequently adds all other neighboring cells. 

This PR enables that ``find_all_active_cells_around_point()`` can be used in combination with all other ``find_active_cell_around_point()`` functions (not just one particular function).


source/grid/grid_tools_dof_handlers.cc
```diff
@@ -609,116 +606,152 @@ namespace GridTools
...
-        // below.
-        else if (n_dirs_at_threshold == 2)
+          cells = find_cells_adjacent_to_vertex(
+            mesh, my_cell->vertex_index(local_vertex_index));
+        for (const auto &cell : cells)
+          if (cell != my_cell)
+            cells_to_add.push_back(cell);
+      }
+    // point on line in 3D: We cannot simply take the intersection between
+    // the two vertices of cels because of hanging nodes. So instead we
```
kronbichler, line 689:
```suggestion
    // the two vertices of cells because of hanging nodes. So instead we
```

source/grid/grid_tools_dof_handlers.cc
```diff
@@ -592,15 +592,12 @@ namespace GridTools
...
-    // of cells) and later add a list of more cells as appropriate.
     std::vector<
       std::pair<typename MeshType<dim, spacedim>::active_cell_iterator,
                 Point<dim>>>
       cells_and_points;
+
+    // find a first cell and point
     try
       {
         cells_and_points.push_back(find_active_cell_around_point_tolerance(
```
kronbichler, line 603: How about rewriting this part slightly, i.e., remove the `cells_and_points` vector above and inside the try clause do
```c++
const auto cell_and_point = find_active_cell_around_point_tolerance(...);
return find_all_active_cells_around_point(
          mapping, mesh, p, tolerance, cell_and_point);
```
and as a default case (search did not succeed) return `{}`? I think that would make the code easier to read.

nfehn: I addressed these points. I am not sure about the ifdef in the declaration because for the new function `MeshType` is now also a parameter of the function. Do I also have to add something to a .inst.in file?

nfehn:
I added an example to the docu.

nfehn:
I added a test and rebased

tamiko:
We need to clean up a couple of MPI tests that use `find_active_cells_around_point()` and are now failing at least for clang-10 with libc++, see https://cdash.43-1.org/viewTest.php?onlydelta&buildid=7263

PR #10483, opened 2020-06-09
Fix compilation with c++14: add missing instantiation

tamiko:
This hopefully fixes https://cdash.43-1.org/viewBuildError.php?buildid=7240

In reference to #10348


bangerth:
Currently fails like this:
```
/home/travis/build/dealii/dealii/source/lac/affine_constraints.cc:131: warning: no uniquely matching class member found for 
  template void::AffineConstraints::distribute<::LinearAlgebra::TpetraWrappers::Vector< float > >(::LinearAlgebra::TpetraWrappers::Vector< float > &) const
```

PR #10162, opened 2020-05-12
Refactor some SD tests

jppelteret:
Some of the SD tests are unstable, and seemingly depend on the standard library, number type, optimization parameters and compiler being used. Hopefully these changes fix that problem.

PR #10486, opened 2020-06-09
tests/symengine/batch_optimizer: add output variants

tamiko:
Add output variants for the symengine/batch_optimizer that fails due to
changes in the libstdc++ library between version 9 and 10.

Let's simply add an output variant for the time being. The proper fix
will be to change these tests to not produce output depending on the
standard library or compiler that is being used.


jppelteret:
I'd like to get some clarification on the potential overlap between this and https://github.com/dealii/dealii/pull/10162 before proceeding.

tamiko:
@jppelteret Wow, I didn't see #10162 at all. I remember looking at the description but I thought it were an issue not a pull request. Testing the proper fix as we speak.

jppelteret:
No worries. Thanks for testing it. I hope it sorts this all out.

tamiko:
Obsolete :smile_cat:

PR #10488, opened 2020-06-09
Insert particles

shahabgol:
A minor bug in calculation of IDs of new particles in multiple insertions is fixed.
The IDs of new particles started from 0 during each insertion. It is now changed to the number of particles in the system.


bangerth:
@gassmoeller -- what do you say?

blaisb:
> @gassmoeller -- what do you say?

I am not @gassmoeller, but I have a quick comment. Would it be better to use the next_free_particle_index() of the particle_handler class instead? The underlying assumption though is still that there is no "hole" in the particle numbering.

source/particles/particle_handler.cc
```diff
@@ -521,7 +521,8 @@ namespace Particles
     // Calculate all starting points locally
     std::vector<unsigned int> particle_start_indices(n_mpi_processes);
```
luca-heltai, line 522:
```suggestion
    std::vector<types::particle_index> particle_start_indices(n_mpi_processes);
```

luca-heltai:
Superseeded. Thanks for spotting this!

bangerth:
@luca-heltai -- could you add what it is superseded by? Also, would it be worth adding the test anyway?

luca-heltai:
Superseeded by #10510. These two were developed at the same time. The test is equivalent to one of the tests added in #10510.

PR #10490, opened 2020-06-10
Give each cell a globally unique (level) ID

peterrum:
... and cache the infomration in `parallel::TriangulationBase::NumberCache`.

~~At the moment, the IDs on the levels are still missing. For that, I am currently preparing a PR, which simplifies the `DoFHandlerPolicy` for multigrid (ETA: soon).~~


peterrum:
This PR is ready for review! I think the best starting point are the tests [grid/global_ids_01](tests/grid/global_ids_01.cc) and  [grid/global_ids_03](tests/grid/global_ids_03.cc).

tjhei:
This is a useful thing to have, but I have a few questions that should probably be addressed in the documentation or fixed:
- why do you call it ``_id`` and not ``_index``? They are contiguous from 0 to count-1, so the name index might be more appropriate?
- Is the ``global id`` for active cells or for all cells?
- What is the default ordering of them, is this ordered by owner? This is worth spelling out.
- What happens with ids when refining/coarsening/repartitioning?
- Are these available for reading (not writing?) for ghost cells?
- When changing an id, what needs to be done to stay consistent? I would think that you would at least be able to violate that these are numbers between 0 and count-1?

peterrum:
In a nutshell, this PR adds a `DoFHandler<dim>` with `FE_DGQ<dim>(0)` to the `Triangulation` (with some optimizations). This used to be the way to enumerate cells contiguously :(

> why do you call it _id and not _index? They are contiguous from 0 to count-1, so the name index might be more appropriate?

You are right. I have changed them to `global_index()` and `global_level_index()`.

> Is the global id for active cells or for all cells? if for active, this should be reflected in the name.

It is for the active ones. The name is chosen as in the case of `subdomain_id()` vs. `level_subdomain_id()`. Normally, we are not explicit about active, are we?

> What is the default ordering of them, is this ordered by owner? This is worth spelling out.

I don't think this is relevant (it is like in the `DoFHandler` case). The user shouldn't make any particular assumptions. It is suitable to set up a partitioner/vector.

> What happens with ids when refining/coarsening/repartitioning?

The indices are set up again.

> Are these available for reading (not writing?) on ghost cells?
When changing an id, what needs to be done to stay consistent? I would think that you would at least be able to violate that these are numbers between 0 and count-1? What about ghost values?

I have disabled writing for now. The indices can be accessed on locally-owned and ghost cells.

> Can you document the difference between cell->index(), cell->id, and this one?

Yes, I can. Probably the `Module` page is the right place? In short:
* `index()`: local index on the level
* `global_index()`: global index on active level
* `global_level_index()`: global index on refinement level
* `id()`: unique *number* for all possible levels

> This is a useful thing to have,

I agree; we already have a couple of use cases in mind!

tjhei:
> Yes, I can. Probably the `Module` page is the right place? In short:

There is also ``active_cell_index`` by the way, which is similar to this. But unfortunately counts artificial and ghost cells. It might be worth trying to get rid of it?

peterrum:
> There is also active_cell_index by the way, which is similar to this. But unfortunately counts artificial and ghost cells. It might be worth trying to get rid of it?

Good point, @tjhei! I wouldn't go so far that I would suggest to get rid of it. But rather to make it consistent with the new one! As far as I can see there is no big use of `active_cell_index()` but to access a local vector. 

This PR enables the access into a global vector (with the help of global indices and a partitioner). The new data structures could be applied the following way to access a local vector:

```cpp
unsigned int
CellAccessor<dim, spacedim>::active_cell_index() const
{
  const auto i = this->tria->levels[this->present_level]
                   ->global_cell_indices[this->present_index];

  const auto tria_parallel =
    dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
      &*this->tria);

  if (tria_parallel)
    return (tria_parallel->global_cell_index_partitioner().global_to_local(i));
  else
    return i;
}
```

(Note: there are still a few tests that don't like this change. But in principal, this or something similar could work.)

However, let's postpone this to follow-up PR.

peterrum:
> Yes, I can. Probably the Module page is the right place?

Could we postpone this to a follow-up PR? I would like to extend `GridOut::write_svg` so that it can also output the global index. A picture is probably the best way to explain the different indices/ids!?

jodlbauer:
Does the same cell return the same ID independent of the number of MPI ranks used?

peterrum:
> Does the same cell return the same ID independent of the number of MPI ranks used?

It depends how you partition your mesh. If you partition a space-filing curve: yes. If you use some graph-partitions: with high probability no.

include/deal.II/distributed/tria_base.h
```diff
@@ -201,6 +202,18 @@ namespace parallel
     const std::set<types::subdomain_id> &
     level_ghost_owners() const;
 
+    /**
+     * Return partitioner for the global cell indices.
+     */
+    const Utilities::MPI::Partitioner &
+    global_cell_index_partitioner() const;
```
bangerth, line 209: I don't think I understand what the function is about. Can you be more explicit in the documentation? Same for the next one.

include/deal.II/grid/tria_accessor.h
```diff
@@ -3337,6 +3343,24 @@ class CellAccessor : public TriaAccessor<dim, dim, spacedim>
    * @}
    */
 
+  /**
+   * Return global cell id for an active cell.
+   */
+  types::global_cell_index
+  global_active_cell_index() const;
```
bangerth, line 3350: Can you add that (assuming this is true) for sequential triangulations, this function returns the same as `active_cell_index()` but that for parallel triangulations, this is not true (and a brief description of what it really is in that case).

include/deal.II/numerics/vector_tools_integrate_difference.templates.h
```diff
@@ -882,6 +882,8 @@ namespace VectorTools
                                                 q,
                                                 update_flags);
 
+      difference = 0;
+
```
bangerth, line 886: changes like to this file, or converting for loops to range-based are independent of this patch -- you'll make everyone's lives easier if you extract these to separate patches because then the one you're really after becomes smaller and easier to review.

include/deal.II/numerics/vector_tools_integrate_difference.templates.h
```diff
@@ -1252,17 +1248,19 @@ namespace VectorTools
            ExcMessage("input vector cell_error has invalid size!"));
 #ifdef DEBUG
     {
+      InVector cellwise_error_copy = cellwise_error;
+
+      for (const auto &cell : tria.active_cell_iterators())
+        if (cell->is_locally_owned())
+          cellwise_error_copy[cell->active_cell_index()] = 0.0;
```
bangerth, line 1255: I don't get this change here. The previous code did not need the copy of the vector -- why was that change necessary?

source/distributed/fully_distributed_tria.cc
```diff
@@ -391,6 +391,18 @@ namespace parallel
...
 
+    template <int dim, int spacedim>
+    void
+    Triangulation<dim, spacedim>::add_periodicity(
+      const std::vector<dealii::GridTools::PeriodicFacePair<cell_iterator>>
+        &periodicity_vector)
+    {
+      dealii::Triangulation<dim, spacedim>::add_periodicity(periodicity_vector);
+      this->update_number_cache();
+    }
```
bangerth, line 402: I think this had already been part of a previous patch?

source/distributed/tria_base.cc
```diff
@@ -164,9 +165,14 @@ namespace parallel
...
-      for (const auto &cell : this->active_cell_iterators())
-        if (cell->is_ghost())
-          number_cache.ghost_owners.insert(cell->subdomain_id());
+      {
+        const auto vertices_with_ghost_neighbors =
+          GridTools::compute_vertices_with_ghost_neighbors(*this);
+        for (const auto &vertex_with_ghost_neighbors :
+             vertices_with_ghost_neighbors)
+          for (const auto &ghost_neigbor : vertex_with_ghost_neighbors.second)
+            number_cache.ghost_owners.insert(ghost_neigbor);
```
bangerth, line 174: and this as well?

source/distributed/tria_base.cc
```diff
@@ -342,6 +351,196 @@ namespace parallel
...
+  TriangulationBase<dim, spacedim>::reset_global_cell_indices()
+  {
+#ifndef DEAL_II_WITH_MPI
+    Assert(false, ExcNeedsMPI());
+#else
+
+    // currently only implemented for distributed triangulations
+    if (dynamic_cast<const parallel::DistributedTriangulationBase<dim, spacedim>
+                       *>(this) == nullptr)
+      return;
```
bangerth, line 365: That looks dangerous -- what happens for other triangulation? Do we get wrong information in those cases?

source/distributed/tria_base.cc
```diff
@@ -342,6 +351,196 @@ namespace parallel
...
+               MPI_SUM,
+               this->mpi_communicator);
+
+    // 3) give global indices to locally-owned cells and mark all other cells as
+    //    invalid
+    for (const auto &cell : this->active_cell_iterators())
+      if (cell->is_locally_owned())
+        cell->set_active_cell_index(cell_index++);
+      else
+        cell->set_active_cell_index(numbers::invalid_dof_index);
```
bangerth, line 387: don't we have an `invalid_cell_index`? This has the wrong type...

bangerth: also, don't you want to set the *global* indices here and below? This looks wrong.

peterrum:
I have updated this PR since I would like to use as discussed in https://github.com/dealii/dealii/pull/11092#discussion_r511081226. The `active_cell_index` is not touched in any way.

I have introduced:
- `CellAccessor::global_active_cell_index` as globally unique index of active cells (and `global_active_cell_index_local` as the local index into a partitioner)
- `CellAccessor::global_level_cell_index` as globally unique index of level cells (and `global_level_cell_index_local` as the local index into a partitioner)

The old version, which redefines `active_cell_index` (which is the way to go in the long term), can be found here https://github.com/peterrum/dealii/tree/cell_global_id_old.

doc/news/changes/minor/20200612Munch
```diff
@@ -0,0 +1,10 @@
+New: Each cell is assigned a globally unique active cell index and (if requested) 
+a level cell index. These indices are integers and originate from a contiguous index set.
```
kronbichler, line 2:
```suggestion
a level cell index. These indices are integers enumerated contiguously within
each subdomain of the mesh. 
```
Is this what you try to say here?

doc/news/changes/minor/20200612Munch
```diff
@@ -0,0 +1,10 @@
+New: Each cell is assigned a globally unique active cell index and (if requested) 
+a level cell index. These indices are integers and originate from a contiguous index set. 
+Users can query locally-owned and ghost cells for their indices via CellAccessor::global_active_cell_index()
+or CellAccessor::global_level_cell_index().
+The value is managed automatically by the Triangulation classes.
+Furthermore, triangulations deriving from parallel::TriangulationBase provide partitioners
+for these indices. These partitioners can be used to set up ghosted vectors with one entry per
```
kronbichler, line 7:
```suggestion
for these indices, which can be used to set up ghosted vectors with one entry per
```

include/deal.II/distributed/tria_base.h
```diff
@@ -201,6 +202,18 @@ namespace parallel
...
     level_ghost_owners() const;
 
+    /**
+     * Return partitioner for the global active cell indices.
+     */
+    const Utilities::MPI::Partitioner &
+    global_active_cell_index_partitioner() const;
+
+    /**
+     * Return partitioner for the global level cell indices of given @p level.
```
kronbichler, line 212:
```suggestion
     * Return partitioner for the global indices of the cells on the given @p level of the triangulation.
```

include/deal.II/grid/tria_accessor.h
```diff
@@ -3443,6 +3449,35 @@ class CellAccessor : public TriaAccessor<dim, dim, spacedim>
    * @}
    */
 
+  /**
+   * Return global active cell id for an active cell.
```
kronbichler, line 3453:
```suggestion
   * Return global active cell index for an active cell.
```

include/deal.II/grid/tria_accessor.h
```diff
@@ -3443,6 +3449,35 @@ class CellAccessor : public TriaAccessor<dim, dim, spacedim>
...
+
+  /**
+   * Return local index inside of the partitioner returned by
+   * global_active_cell_index_partitioner().
+   *
+   * @note For a serial Triangulation, the returned value is the same as
+   *   returned by active_cell_index().
+   */
+  types::global_cell_index
+  global_active_cell_index_local() const;
```
kronbichler, line 3466: I find this name a bit confusing by having `global` and `local` in the name. For what purposes do we anticipate using this index, given that it can be computed reasonably quickly by `tria.get_active_cell_index_partitioner().global_to_local(cell->global_active_cell_index())`? Of course, this would imply that we actually store the global index in the mesh (rather than the MPI-local one), and skip all the index translations in the implementation below.

include/deal.II/grid/tria_levels.h
```diff
@@ -156,6 +156,16 @@ namespace internal
...
+      /**
+       * Global cell index of each active cell.
+       */
+      std::vector<types::global_cell_index> global_active_cell_indices;
+
+      /**
+       * Global cell index of each cell on the given level.
+       */
+      std::vector<types::global_cell_index> global_level_cell_indices;
+
```
kronbichler, line 168: Can you move these vectors close to the `active_cell_indices` field in line 103 or that part down here?

source/distributed/tria_base.cc
```diff
@@ -342,6 +346,201 @@ namespace parallel
...
+        std::vector<types::global_cell_index> n_cells_level(
+          this->n_global_levels(), 0);
+
+        MPI_Allreduce(n_locally_owned_cells.data(),
+                      n_cells_level.data(),
+                      this->n_global_levels(),
+                      Utilities::MPI::internal::mpi_type_id(
+                        n_locally_owned_cells.data()),
+                      MPI_SUM,
+                      this->mpi_communicator);
```
kronbichler, line 456: I'm sure the additional efficiency is minuscule compare to , but I ask nonetheless: What would you think of doing an `MPI_Bcast` the result from `ExScan` on the last rank, augmented by the sum of cells on all levels on that rank?

source/distributed/tria_base.cc
```diff
@@ -342,6 +346,201 @@ namespace parallel
...
+                      this->locally_owned_subdomain())
+                    is_local.add_index(index);
+                  else
+                    is_ghost.add_index(index);
+                }
+
+            number_cache.level_cell_index_partitioners[l] =
+              Utilities::MPI::Partitioner(is_local,
+                                          is_ghost,
+                                          this->mpi_communicator);
```
kronbichler, line 506: Given that we didn't build that information before at all, this is a pretty compact code in relation to the content of information contained here.

source/distributed/tria_base.cc
```diff
@@ -342,6 +346,201 @@ namespace parallel
...
+                        .global_to_local(
+                          cell->global_level_cell_index_local()));
+                }
+          }
+      }
+
+#endif
+  }
+
+
```
kronbichler, line 525: use three empty lines between the different functions, here and below.

kronbichler (commented):
It's good to get this to the finish line. I collected a few questions and remarks below.

include/deal.II/grid/tria_accessor.h
peterrum: I am actually translating during setup the global indices to local indices. This function returns simply that value.

source/distributed/tria_base.cc
peterrum: It has been prepared by a couple of PRs half a year ago ;)

peterrum:
@kronbichler I have incorporated your suggestions!

include/deal.II/grid/tria_accessor.h
kronbichler: I know, that's what I meant by "skip all the index translations" because we would just fill the global index.

peterrum: I am not sure what you are suggesting. Should we save the actual global index?

In my opinion, it is more useful to save the MPI-local indices since in normal case we would access some vectors. But I can change this.

kronbichler: Yes, my intuition would have been to save the global index and only provide an access function to that variable since I see the confusion of two "MPI-local" indices, the `active_cell_index()` we already have and this one here. I admit it is not perfect because one needs to go through the index translations and access vectors with `operator()`. Anyway, this is up to discussion and I wanted to
- hear what others think and
- in case we like the more direct access to the MPI-local indices, have others participate in finding a good name for the access function.

include/deal.II/distributed/tria_base.h
```diff
@@ -201,6 +202,20 @@ namespace parallel
...
+    const Utilities::MPI::Partitioner &
+    global_active_cell_index_partitioner() const;
+
+    /**
+     * Return partitioner for the global indices of the cells on the given @p
+     * level of the triangulation.
+     */
+    const Utilities::MPI::Partitioner &
+    global_level_cell_index_partitioner(const unsigned int level) const;
+
```
kronbichler, line 218: A question on the place for this function: If we envision these functions to be often, I would prefer to have the partitioners in the base class `Triangulation` already, to avoid downcasting in many places. After all, a partitioner is also valid (albeit boring) on a serial triangulation.

kronbichler: Wait, we also provide the MPI communicator only in this function, so we better stick to this class.

kronbichler (approved):

include/deal.II/grid/tria_accessor.h
peterrum: As discussed offline, I am now storing the global indices and have removed the function that are ending with `_local` for now.

kronbichler (approved):

include/deal.II/grid/tria_accessor.h
kronbichler: @peterrum do you still want to augment the documentation somewhat as suggested by @bangerth?

kronbichler:
Let us address the comment in a follow-up PR and merge here.

PR #10492, opened 2020-06-10
Document particle_lost in particle handler.

luca-heltai:
In sort_particles_into_subdomains_and_cells we do not document the possibility to attach to a signal.


luca-heltai:
The failing build is totally unrelated. This is only a doc change.

PR #10496, opened 2020-06-10
Add missing instantiation for PETSc with complex values

masterleinad:
Fixes https://cdash.43-1.org/viewBuildError.php?buildid=7282.

### DIFF
--- a/source/lac/affine_constraints.cc
+++ b/source/lac/affine_constraints.cc
@@ -109,6 +109,14 @@ INSTANTIATE_DLTG_BLOCK_VECTORMATRIX(PETScWrappers::MPI::BlockSparseMatrix,
 INSTANTIATE_DLTG_MATRIX(PETScWrappers::SparseMatrix);
 INSTANTIATE_DLTG_MATRIX(PETScWrappers::MPI::SparseMatrix);
 INSTANTIATE_DLTG_MATRIX(PETScWrappers::MPI::BlockSparseMatrix);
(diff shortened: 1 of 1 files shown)


source/lac/affine_constraints.cc
```diff
@@ -109,6 +109,14 @@ INSTANTIATE_DLTG_BLOCK_VECTORMATRIX(PETScWrappers::MPI::BlockSparseMatrix,
 INSTANTIATE_DLTG_MATRIX(PETScWrappers::SparseMatrix);
 INSTANTIATE_DLTG_MATRIX(PETScWrappers::MPI::SparseMatrix);
 INSTANTIATE_DLTG_MATRIX(PETScWrappers::MPI::BlockSparseMatrix);
+#  ifndef DOXYGEN
+#    ifdef DEAL_II_PETSC_WITH_COMPLEX
+template void
+dealii::AffineConstraints<double>::distribute<
```
bangerth, line 115: Don't you want this to be `PetscScalar`? Or do I misunderstand why you're guarding this with `DEAL_II_PETSC_WITH_COMPLEX`? (Maybe a comment would help :-) )

masterleinad: https://cdash.43-1.org/viewBuildError.php?buildid=7282 says that we need this combination. I didn't investigate more where that comes from.

bangerth: I see. Never mind then!

PR #10503, opened 2020-06-10
Add GridTools::exchange_cell_data_to_level_ghosts()

peterrum:
... similar to `GridTools::exchange_cell_data_to_ghosts()`.

I also used the opportunity to use it and to clean up the setup of the multigrid levels in the `DoFHandler` policy, which now looks like this:

```cpp
template <int dim, int spacedim, class DoFHandlerType>
void
communicate_mg_ghost_cells(
  const typename dealii::parallel::
    DistributedTriangulationBase<dim, spacedim> &tria,
  DoFHandlerType &                               dof_handler)
{
  const auto pack = [&](const auto &cell) {
    // why would somebody request a cell that is not ours?
    Assert(cell->level_subdomain_id() == tria.locally_owned_subdomain(),
           ExcInternalError());

    std::vector<dealii::types::global_dof_index> data(
      cell->get_fe().dofs_per_cell);
    cell->get_mg_dof_indices(data);

    return data;
  };

  const auto unpack = [](const auto &cell, const auto &dofs) {
    Assert(cell->get_fe().dofs_per_cell == dofs.size(),
           ExcInternalError());

    Assert(cell->level_subdomain_id() !=
             dealii::numbers::artificial_subdomain_id,
           ExcInternalError());

    std::vector<dealii::types::global_dof_index> dof_indices(
      cell->get_fe().dofs_per_cell);
    cell->get_mg_dof_indices(dof_indices);

    bool complete = true;
    for (unsigned int i = 0; i < dof_indices.size(); ++i)
      if (dofs[i] != numbers::invalid_dof_index)
        {
          Assert((dof_indices[i] == (numbers::invalid_dof_index)) ||
                   (dof_indices[i] == dofs[i]),
                 ExcInternalError());
          dof_indices[i] = dofs[i];
        }
      else
        complete = false;

    if (!complete)
      const_cast<typename DoFHandlerType::level_cell_iterator &>(cell)
        ->set_user_flag();
    else
      const_cast<typename DoFHandlerType::level_cell_iterator &>(cell)
        ->clear_user_flag();

    const_cast<typename DoFHandlerType::level_cell_iterator &>(cell)
      ->set_mg_dof_indices(dof_indices);
  };

  const auto filter = [](const auto &cell) {
    return cell->user_flag_set();
  };

  GridTools::exchange_cell_data_to_level_ghosts<
    DoFHandlerType>(dof_handler, pack, unpack, filter);
}
```

### DIFF
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -2853,6 +2853,27 @@ namespace GridTools
     const std::function<void(const typename MeshType::active_cell_iterator &,
                              const DataType &)> &        unpack);
 
+  /**
+   * Exchange arbitrary data of type @p DataType provided by the function
+   * objects from locally owned level cells to ghost level cells on other
+   * processes.
+   *
+   * In addition to the parameters of exchange_cell_data_to_ghosts(), this
+   * function allows to provide a @p filter function, which can be used to only
+   * communicate marked cells. In the default case, all relevant cells are
+   * communicated.
+   */
+  template <typename DataType, typename MeshType>
+  void
+  exchange_cell_data_to_level_ghosts(
+    const MeshType &                                    mesh,
+    const std::function<std_cxx17::optional<DataType>(
+      const typename MeshType::level_cell_iterator &)> &pack,
+    const std::function<void(const typename MeshType::level_cell_iterator &,
+                             const DataType &)> &       unpack,
+    const std::function<bool(const typename MeshType::level_cell_iterator &)>
+      &filter = [](const auto &) { return true; });
+
   /* Exchange with all processors of the MPI communicator @p mpi_communicator the vector of bounding
    * boxes @p local_bboxes.
    *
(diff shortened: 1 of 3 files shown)


include/deal.II/grid/grid_tools.h
```diff
@@ -2853,6 +2853,27 @@ namespace GridTools
...
                              const DataType &)> &        unpack);
 
+  /**
+   * Exchange arbitrary data of type @p DataType provided by the function
+   * objects from locally owned level cells to ghost level cells on other
+   * processes.
+   *
+   * In addition to the parameters of exchange_cell_data_to_ghosts(), this
+   * function allows to provide a @p filter function, which can be used to only
+   * communicate marked cells.In the default case, all relevant cells are
```
kronbichler, line 2863:
```suggestion
   * communicate marked cells. In the default case, all relevant cells are
```

include/deal.II/grid/grid_tools.h
```diff
@@ -4035,6 +4056,233 @@ namespace GridTools
...
+    std::map<dealii::types::subdomain_id,
+             std::vector<typename CellId::binary_type>>
+      neighbor_cell_list;
+
+    for (const auto level_ghost_owner : level_ghost_owners)
+      neighbor_cell_list[level_ghost_owner] = {};
+
+    {
+      typename MeshType::level_cell_iterator cell, endc = mesh.end();
+      for (cell = mesh.begin(); cell != endc; ++cell)
```
kronbichler, line 4106: Can you do this with a range-based for loop?
```suggestion
      for (const auto &cell : mesh.cell_iterators())
```
or are there some types which do not support these loops?

include/deal.II/grid/grid_tools.h
```diff
@@ -4035,6 +4056,233 @@ namespace GridTools
...
+
+
+    // Before sending & receiving, make sure we protect this section with
+    // a mutex:
+    static Utilities::MPI::CollectiveMutex      mutex;
+    Utilities::MPI::CollectiveMutex::ScopedLock lock(mutex,
+                                                     tria->get_communicator());
+
+    const int mpi_tag =
+      Utilities::MPI::internal::Tags::dofhandler_communicate_mg_ghost_cells;
```
kronbichler, line 4126: Are you planning to update the mpi tag later to a similar name we have for the exchange function on the active cells or is this simply an oversight?

include/deal.II/grid/grid_tools.h
```diff
@@ -4035,6 +4056,233 @@ namespace GridTools
...
+        ierr = MPI_Recv(cell_data_to_send[idx].data(),
+                        len,
+                        MPI_BYTE,
+                        status.MPI_SOURCE,
+                        status.MPI_TAG,
+                        tria->get_communicator(),
+                        &status);
+        AssertThrowMPI(ierr);
+
+        // store the dof indices for each cell
```
kronbichler, line 4187: Outdated comment?

peterrum:
@kronbichler Thanks for the quick review. I have addressed the small issues. I have not worked on the merge of the two functions yet. A general issue here is that the two functions differ in what to send (not just data is sent from different levels): `
- `exchange_cell_data_to_ghosts()` all cells that neighbor ghost cells are communicated (optionally without any payload); this allows a single communication step
- in `exchange_cell_data_to_level_ghosts`, only needed cells are communicated (which is determined with an additional communication step)

Personally, I find the two-step-approach a bit cleaner. because it does not need the intermediate step via `GridTools::compute_vertices_with_ghost_neighbors()` - see also https://github.com/dealii/dealii/blob/8c5a90392887cc6138d2f9e254321a660efaaaf0/include/deal.II/grid/grid_tools.h#L3897-L3921

What is your preference?

kronbichler:
I agree with the two-step process; it acts on a purely algebraic level without building the dual structure with the vertices. The communication of the first stage is typically sufficiently cheap so we are likely not limited there anyway. I guess this also works easily with periodic boundary conditions, which was a place where we had a bit of trouble when doing it with the vertices in the past.

peterrum:
> I agree with the two-step process;

Then let's merge this PR, as it is - if there are no major requests. And I will move the merged logic of the two functions into an internal function in a follow-up PR. Furthermore, I will introduce new filters in the DoFHandler policy also for active cells.

> I guess this also works easily with periodic boundary conditions, which was a place where we had a bit of trouble when doing it with the vertices in the past.

I know. I also went through that when working on  p:f:t :(

PR #10510, opened 2020-06-11
Add global particles

luca-heltai:
The function Particles::ParticleHandler::add_global_particles() now takes 
another optional argument, that allows one to set ids arbitrarily. Moreover,
now the numbering of the ids is correct also if we call the method more than
one time. Newly added particles, if ids are not specified, now correctly get
the first available ids. 
Added a new version of Particles::ParticleHandler::add_global_particles() that
takes a vector of Particles::Particle objects instead of just their positions.
This can be used in conjunction with the signal
Particles::ParticleHandler::Signals::particle_lost() to reinsert
Particles::Particle objects that went out of the locally owned and ghost cells.

This is why we may need this PR:

```C++
std::vector<Particles::Particle<dim,spacedim>> lost_particles;

auto collect_lost_particles = [&lost_particles] 
	(const typename Particles::ParticleIterator<dim, spacedim> &particle,
	const typename Triangulation<dim, spacedim>::active_cell_iterator &) {
	lost_particles.push_back(particle);
};

particle_handler.signals.particle_lost.connect(lost_particles);
particle_handler.sort_particles_into_subdomains_and_cells();
particle_handler.insert_global_particles(lost_particles);
lost_particles.clear();

```


luca-heltai:
@blaisb and @gassmoeller : this fixes also the problem that @blaisb observed in insert_global_particles.

blaisb:
From my understanding this negates the need for  #10488  :). Sorry @shahabgol , it was  a very nice first contribution :)!.

luca-heltai:
Yes, sorry, I had to incorporate those changes here as well. Buf you merge that one first, I'll rebase, and we can give him credit for this. :)

blaisb:
@luca-heltai  That is not for me to judge :). In all cases I think what matters is that the code is working like it should :)! It is a team effort after all.

include/deal.II/particles/particle_handler.h
```diff
@@ -290,6 +290,10 @@ namespace Particles
      * process that will own each of the particles, and it may therefore be
      * communication intensive.
      *
+     * @param[in] ids (Optional) A vector of ids to associate to each particle.
+     * If the vector is empty, the ids are computed automatically, by assigning
+     * them as a continuous range starting from first available index.
```
bangerth, line 295: Can you clarify what that means in a parallel context? I suspect what you mean to say is that "assigning them" refers to the collection of particles that are being added across all processors. (I'm making the implicit assumption that this is actually a collective function.)

include/deal.II/particles/particle_handler.h
```diff
@@ -299,8 +303,45 @@ namespace Particles
...
+     * particles to the corresponding process.
+     *
+     * In order to keep track of what mpi process received what particles, a map
+     * from mpi process to IndexSet is returned by the function. This IndexSet
+     * contains the local indices of the particles that were passed to this
+     * function on the calling mpi process, and that falls within the part of
+     * the triangulation owned by this mpi process.
+     *
+     * @param[in] particles A vector of particles that do not need to be on the
+     * local processor
```
bangerth, line 326:
```suggestion
     * local processor.
```

include/deal.II/particles/particle_handler.h
```diff
@@ -299,8 +303,45 @@ namespace Particles
...
+     * can be obtained from GridTools::compute_mesh_predicate_bounding_box(),
+     * and the global one can be obtained by passing the local ones to
+     * Utilities::MPI::all_gather().
+     *
+     * @return A map from owner to IndexSet, that contains the local indices
+     * of the points that were passed to this function on the calling mpi
+     * process, and that falls within the part of triangulation owned by this
+     * mpi process.
+     */
+    std::map<unsigned int, IndexSet>
```
bangerth, line 340: Is `IndexSet` the right data structure here? Or could it just be a `std::set`? The difference being that the former is optimized for cases where we have a few contiguous ranges, the latter being a completely unstructured set.

source/particles/particle_handler.cc
```diff
@@ -92,7 +92,7 @@ namespace Particles
     , global_number_of_particles(0)
     , global_max_particles_per_cell(0)
     , next_free_particle_index(0)
-    , property_pool(new PropertyPool(0))
+    , property_pool(std::make_unique<PropertyPool>(0))
```
bangerth, line 95: :-)

source/particles/particle_handler.cc
```diff
@@ -595,35 +597,71 @@ namespace Particles
     std::map<unsigned int, std::vector<std::vector<double>>>
       locally_owned_properties_from_other_processes;
 
-    if (n_global_properties > 0)
+    // A map from mpi process to ids, ordered as in the IndexSet.
+    // Notice that this ordering maybe different from the ordering in the
```
bangerth, line 601:
```suggestion
    // Notice that this ordering may be different from the ordering in the
```

source/particles/particle_handler.cc
```diff
@@ -595,35 +597,71 @@ namespace Particles
...
+    // vectors above, since no local ordering is guaranteed by the
+    // distribute_compute_point_locations() call.
+    // This is only filled if ids.size() is > 0
```
bangerth, line 604: I might have already forgotten, but did you say that in the documentation?

source/particles/particle_handler.cc
```diff
@@ -681,6 +724,43 @@ namespace Particles
...
+    positions.resize(particles.size());
+    ids.resize(particles.size());
+    if (n_properties_per_particle() > 0)
+      properties.resize(particles.size(),
+                        std::vector<double>(n_properties_per_particle()));
+
+    unsigned int i = 0;
+    for (const auto &p : particles)
+      {
+        positions[i] = p.get_location();
```
bangerth, line 748: If you only call `.reserve()` instead of `.resize()` above, then you can here do
```suggestion
        positions.emplace_back (p.get_location());
```
and similarly for the next couple of lines, and you don't need the counter `i` at all.

include/deal.II/particles/particle_handler.h
luca-heltai: If particles are contiguous (and they often are, when calling this function), it is convenient to return an IndexSet.

source/particles/particle_handler.cc
luca-heltai: yes. :)

source/particles/particle_handler.cc
luca-heltai: But then I cannot reserve the size of the properties all at once, because `reserve` does not allow me to call it with an initializer for the entries.

luca-heltai:
Ping @bangerth ?

PR #10516, opened 2020-06-11
Merge implementations of exchange_cell_data_to_(level)_ghosts 

peterrum:
... and update `communicate_dof_indices_on_marked_cells` in `DoFHandler` policy such that it uses filters.

Follow-up PR to #10503.

### DIFF
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -2801,7 +2801,9 @@ namespace GridTools
    *   on the sending side returned a non-empty std_cxx17::optional object.
    *   The @p unpack function is then called with the data sent by the
    *   processor that owns that cell.
-   *
+   * @param cell_filter Only cells are communicated where this filter function returns
+   *   the value `true`. In the default case, the function returns true on all
+   * cells and thus, all relevant cells are communicated.
    *
    * <h4> An example </h4>
    *
@@ -2852,7 +2854,10 @@ namespace GridTools
     const std::function<std_cxx17::optional<DataType>(
       const typename MeshType::active_cell_iterator &)> &pack,
     const std::function<void(const typename MeshType::active_cell_iterator &,
-                             const DataType &)> &        unpack);
+                             const DataType &)> &        unpack,
+    const std::function<bool(const typename MeshType::active_cell_iterator &)>
+      &cell_filter =
+        [](const typename MeshType::active_cell_iterator &) { return true; });
 
   /**
    * Exchange arbitrary data of type @p DataType provided by the function
@@ -2860,7 +2865,7 @@ namespace GridTools
    * processes.
    *
    * In addition to the parameters of exchange_cell_data_to_ghosts(), this
-   * function allows to provide a @p filter function, which can be used to only
+   * function allows to provide a @p cell_filter function, which can be used to only
    * communicate marked cells. In the default case, all relevant cells are
    * communicated.
    */
@@ -2873,7 +2878,7 @@ namespace GridTools
     const std::function<void(const typename MeshType::level_cell_iterator &,
                              const DataType &)> &       unpack,
     const std::function<bool(const typename MeshType::level_cell_iterator &)>
-      &filter =
+      &cell_filter =
         [](const typename MeshType::level_cell_iterator &) { return true; });
 
   /* Exchange with all processors of the MPI communicator @p mpi_communicator the vector of bounding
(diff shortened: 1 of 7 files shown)


peterrum:
The code of `communicate_dof_indices_on_marked_cells()` is short and innocent:

```cpp
const auto pack = [](const auto &cell) {
  Assert(cell->is_locally_owned(), ExcInternalError());

  std::vector<dealii::types::global_dof_index> data(
    cell->get_fe().dofs_per_cell);
  cell->get_dof_indices(data);

  return data;
};

const auto unpack = [](const auto &cell, const auto &dofs) {
  Assert(cell->get_fe().dofs_per_cell == dofs.size(),
         ExcInternalError());

  Assert(cell->is_ghost(), ExcInternalError());

  std::vector<dealii::types::global_dof_index> dof_indices(
    cell->get_fe().dofs_per_cell);
  cell->update_cell_dof_indices_cache();
  cell->get_dof_indices(dof_indices);

  bool complete = true;
  for (unsigned int i = 0; i < dof_indices.size(); ++i)
    if (dofs[i] != numbers::invalid_dof_index)
      {
        Assert((dof_indices[i] == (numbers::invalid_dof_index)) ||
                 (dof_indices[i] == dofs[i]),
               ExcInternalError());
        dof_indices[i] = dofs[i];
      }
    else
      complete = false;

  if (!complete)
    const_cast<typename DoFHandlerType::active_cell_iterator &>(cell)
      ->set_user_flag();
  else
    const_cast<typename DoFHandlerType::active_cell_iterator &>(cell)
      ->clear_user_flag();

  const_cast<typename DoFHandlerType::active_cell_iterator &>(cell)
    ->set_dof_indices(dof_indices);
};

const auto filter = [](const auto &cell) {
  return cell->user_flag_set();
};

GridTools::exchange_cell_data_to_ghosts<
  DoFHandlerType>(dof_handler, pack, unpack, filter);
```

bangerth:
I suspect that @marcfehling touched this last. Let's see if he can give some feedback!

include/deal.II/grid/grid_tools.h
```diff
@@ -3881,182 +3884,280 @@ namespace GridTools
...
+    exchange_cell_data(
+      const MeshType &mesh,
+      const std::function<
+        std_cxx17::optional<DataType>(const MeshCellIteratorType &)> &pack,
+      const std::function<void(const MeshCellIteratorType &, const DataType &)>
+        &                                                        unpack,
+      const std::function<bool(const MeshCellIteratorType &)> &  filter,
+      const std::function<void(
+        const std::function<void(const MeshCellIteratorType &,
+                                 const types::subdomain_id)> &)> process_cells,
```
bangerth, line 3902: I think you want to mark this argument as a reference as well.

include/deal.II/grid/grid_tools.h
```diff
@@ -3881,182 +3884,280 @@ namespace GridTools
...
+              typename MeshType,
+              typename MeshCellIteratorType>
+    void
+    exchange_cell_data(
+      const MeshType &mesh,
+      const std::function<
+        std_cxx17::optional<DataType>(const MeshCellIteratorType &)> &pack,
+      const std::function<void(const MeshCellIteratorType &, const DataType &)>
+        &                                                        unpack,
+      const std::function<bool(const MeshCellIteratorType &)> &  filter,
```
bangerth, line 3899: Maybe call it `cell_filter` to be clearer what it filters?

marcfehling:
@bangerth I have to disappoint you: I did not work on the implementation on these functions, but used them frequently.

I like this functionality! This would be handy for the bug resolved in #10309. I think about adjusting it once this PR has been merged.

Browsing through your source code, I don't have any suggestions to improve it. Looking forward to using it :)

marcfehling:
I'm a bit worried about the test results. As far as I understood, you changed the test in a way that the output should not be affected, but it has been altered. Do you have any idea why?

peterrum:
> I'm a bit worried about the test results. As far as I understood, you changed the test in a way that the output should not be affected, but it has been altered. Do you have any idea why?

@marcfehling I have updated the test such that the output does not have to be altered. However, I must admit the tests are utterly inappropriate for testing the functionally of the function. On the one hand it checks that the right data is sent, but it also it checks in which sequence. Latter is just an implementation aspect and is not relevant for the correctness.

marcfehling:
I'm okay if implementation changes alter test outputs. I just didn't see immediately that this was intended in this particular case. Well obviously the remaining part of the testsuite didn't complain so I shouldn't be too worried about this one in particular...

If you prefer to keep your original version of your adapted test, feel free to recover it. Otherwise leave it as it is

include/deal.II/grid/grid_tools.h
```diff
@@ -2801,7 +2801,8 @@ namespace GridTools
    *   on the sending side returned a non-empty std_cxx17::optional object.
    *   The @p unpack function is then called with the data sent by the
    *   processor that owns that cell.
-   *
+   * @param cell_filter Function, which can be used to only communicate marked cells.
+   *   In the default case, all relevant cells are communicated.
```
kronbichler, line 2805:
```suggestion
   * @param cell_filter Only cells are communicated where this filter function returns
   *   the value `true`. In the default case, the function returns true on all cells
   *   and thus, all relevant cells are communicated.
```

include/deal.II/grid/grid_tools.h
```diff
@@ -3883,182 +3887,282 @@ namespace GridTools
...
-
-  template <typename DataType, typename MeshType>
-  void
-  exchange_cell_data_to_ghosts(
-    const MeshType &                                     mesh,
-    const std::function<std_cxx17::optional<DataType>(
-      const typename MeshType::active_cell_iterator &)> &pack,
-    const std::function<void(const typename MeshType::active_cell_iterator &,
-                             const DataType &)> &        unpack)
+  namespace
```
kronbichler, line 3890: We generally avoid anonymous namespaces in header files. Can you wrap this into an `internal` namespace`?

peterrum:
@kronbichler I have made the changes!

PR #10520, opened 2020-06-12
Clarify docu of GridTools::find_active_cell_around_point()

### LINKED ISSUE #10459: Clarify expected behavior of GridTools::find_active_cell_around_point()
The documentation of ``GridTools::find_active_cell_around_point()`` says that the cell returned may be an artificial cell, but it is unclear whether this makes sense or whether this is possible, see PR #10456. 

This issue should clarify the expected behavior.

nfehn:
I tried to clarify the behavior of this function regarding locally-owned, ghost, and artificial cells. I also shortened the docu because many aspects were mentioned twice.

Further, I changed the order of function declarations so tthat references to "the above function" are more clear.

Closes #10459.

@luca-heltai I would like to also change the order of the two functions with a-priori knowledge (the one with the cache object and the other one with the rtree object). Further, I would like to add a code snipped explaining the usage of these functions (given the complexity of the arguments). Do you agree?

### DIFF
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -1099,27 +1099,6 @@ namespace GridTools
   find_cells_adjacent_to_vertex(const MeshType<dim, spacedim> &container,
                                 const unsigned int             vertex_index);
 
-
-  /**
-   * Find and return an iterator to the active cell that surrounds a given
-   * point. This function simply calls the following one with a
-   * MappingQ1 for the mapping argument. See the following function for
-   * a more thorough discussion.
-   *
-   * @return An iterator into the mesh that points to the surrounding cell.
-   */
-  template <int dim, template <int, int> class MeshType, int spacedim>
-#  ifndef _MSC_VER
-  typename MeshType<dim, spacedim>::active_cell_iterator
-#  else
-  typename dealii::internal::
-    ActiveCellIterator<dim, spacedim, MeshType<dim, spacedim>>::type
-#  endif
-  find_active_cell_around_point(const MeshType<dim, spacedim> &mesh,
-                                const Point<spacedim> &        p,
-                                const std::vector<bool> &marked_vertices = {},
-                                const double             tolerance = 1.e-10);
-
   /**
    * Find an active cell that surrounds a given point @p p. The return type
    * is a pair of an iterator to the active cell along with the unit cell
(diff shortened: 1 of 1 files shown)


include/deal.II/grid/grid_tools.h
```diff
@@ -1218,6 +1175,39 @@ namespace GridTools
                                 const std::vector<bool> &marked_vertices = {},
                                 const double             tolerance = 1.e-10);
 
+  /**
+   * A predecessor of the above function that internally calls the above
+   * function using MappingQ1 for the mapping argument.
```
peterrum, line 1180:
```suggestion
   * A version of the above function that assumes straight boundaries and as a 
   * consequence uses MappingQ1 as mapping.
```

nfehn:
@luca-heltai I will open a separate issue for design questions of find_active_cell_around_point(), @kronbichler and @tamiko have mentioned opinions in other chats, so it's getting too complicated if we do not discuss this at a central place. This PR here is for documentation only.

nfehn:
I have now done the second part of this PR, could you please have a look at it again?

include/deal.II/grid/grid_tools.h
```diff
@@ -1266,8 +1210,46 @@ namespace GridTools
     const double                                tolerance = 1.e-10);
 
   /**
-   * A version of the previous function that exploits an already existing
-   * GridTools::Cache<dim,spacedim> object.
+   * Finding an active cell around a point can be very expensive in terms
+   * of computational costs. This function aims at providing a fast version
+   * of the above functions by using a space-tree to speed up the geometry
+   * search. This information is provided via the parameter @param cache.
```
kronbichler, line 1216:
```suggestion
   * search. This information is provided via the parameter @param cache .
```
It seems this caused the doxygen check to fail:
```
/home/runner/work/dealii/dealii/include/deal.II/grid/grid_tools.h:1216: warning: argument 'cache.' of command @param is not found in the argument list of GridTools::find_active_cell_around_point(const Cache< dim, spacedim > &cache, const Point< spacedim > &p, const typename Triangulation< dim, spacedim >::active_cell_iterator &cell_hint=typename Triangulation< dim, spacedim >::active_cell_iterator(), const std::vector< bool > &marked_vertices={}, const double tolerance=1.e-10)
/home/runner/work/dealii/dealii/include/deal.II/grid/grid_tools.h:1260: warning: The following parameters of GridTools::find_active_cell_around_point(const Cache< dim, spacedim > &cache, const Point< spacedim > &p, const typename Triangulation< dim, spacedim >::active_cell_iterator &cell_hint=typename Triangulation< dim, spacedim >::active_cell_iterator(), const std::vector< bool > &marked_vertices={}, const double tolerance=1.e-10) are not documented:
  parameter 'cache'
  parameter 'p'
```
Also, since you documented all other parameters, you should add @param p in the first line. However, I'm not sure whether doxygen renders these inline definitions of parameters in an appropriate way, i.e., whether it handles the inline text this is embedded into. Did you check the generated documentation? To be on the safe side, I suggest to list the parameters one by one in a list as in one of the other `find_active_point_around_cells` functions, to make it look approximately like this:
https://www.dealii.org/developer/doxygen/deal.II/namespaceGridTools.html#a2e10aeb1c8e76110a84b6945eac3aaf0

nfehn: Oh, I did not want the parameters to be listed. The documentation is indeed nonsense, sorry. Should be fixed now.

nfehn:
I stashed the commits to complete this

PR #10523, opened 2020-06-12
Thread/Task: Fix Schmidt's double checking

tamiko:
We have to use a std::atomic_bool instead of a bool to ensure proper
memory fencing, see [1] for an in-detail discussion.

[1] https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

### DIFF
--- a/include/deal.II/base/thread_management.h
+++ b/include/deal.II/base/thread_management.h
@@ -24,6 +24,7 @@
 #  include <deal.II/base/std_cxx17/tuple.h>
 #  include <deal.II/base/template_constraints.h>
 
+#  include <atomic>
 #  include <condition_variable>
 #  include <functional>
 #  include <future>
@@ -429,9 +430,9 @@ namespace Threads
       std::shared_ptr<return_value<RT>> ret_val;
 
       /**
-       * A bool variable that is initially false, is set to true when a new
-       * thread is started, and is set back to false once join() has been
-       * called.
+       * An atomic  bool variable that is initially false, is set to true
+       * when a new thread is started, and is set back to false once join()
+       * has been called.
        *
        * We use this variable to make sure we can call join() twice on the
        * same thread. For some reason, the C++ standard library throws a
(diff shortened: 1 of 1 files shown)


tamiko:
@bangerth *ping*

tamiko:
@tjhei You're right, we should use `std::atomic<bool>` instead of the typedefs that were introduced in the C++ standard for compatibility with a proposal that didn't make it into the C standard.

bangerth:
Looks good. How the heck did you find that? And are there more places that need this kind of fix? (Maybe search for "double checking" or "double-checking".)

I suspect that in practice, the functions that acquire a lock are written in a way that prevents the movement of loads and stores across the call. In other words, that the call to acquire the lock acts as a memory fence. But that doesn't mean that we shouldn't be doing the right thing.

tamiko:
@bangerth 
> Looks good. How the heck did you find that?

I happen to spot it when reading the file during the thread-cleanup discussion at the hackathon.

> And are there more places that need this kind of fix? 

I don't think so - I did a short search and couldn't find anything. We also have well encapsulated code (i.e., all of the threading internals are in two files).

> I suspect that in practice, the functions that acquire a lock are written in a way that prevents the movement of loads and stores across the call. 

Yes, they actually have to (for synchronizing the mutex itself) if I am not mistaken. I also think the synchronization of the boolean is not a problem. The atomic boolean is however necessary to fence whatever memory side effect the `join()` operation has, and also in order to ensure that no reordering of the if-statements or similar nasty aggressive optimization takes place. That's at least my partial understanding :-D

bangerth:
Yes, I can't claim more than a partial understanding either. It's not trivial to reason about what precisely acquire/release semantics mean for loads and stores, and what precisely is necessary in any given context. 

All that is certainly complicated by the fact that I don't really understand how processors implement cache coherency strategies. Even if I have an atomic read/write, there is a time delay between writing on one processor and when another processor hears about that. Atomics make it sound like things are, well, atomic, but that's really not true and the semantics are confusing because what one really only cares about are memory fences. I can see that having an atomic cmpxchng operation makes sense, but it makes no real sense to have an atomic read/write on a one-byte data structure -- of course it is atomic, it's not like we're writing 1000 bytes in small increments.

PR #10525, opened 2020-06-12
Enable non-default VectorizedArrayType in read_cell_data()

DavidSCN:
Trying to compile `read_cell_data()` with a non-default `VectorizedArray<Number>` type in my code fails and the changed template argument fixed it for me.

The same issue might be here (but I have not tested/changed it):

https://github.com/dealii/dealii/blob/5fc5452285efbc2e879e7ec943bce0c65d928b37/include/deal.II/matrix_free/fe_evaluation.h#L6059

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -3839,8 +3839,7 @@ FEEvaluationBase<dim, n_components_, Number, is_face, VectorizedArrayType>::
   const auto cells = this->get_cell_ids();
 
   // 2) actually gather values
-  VectorizedArrayType out =
-    make_vectorized_array<VectorizedArrayType>(Number(1.));
+  VectorizedArrayType out = Number(1.);
   for (unsigned int i = 0; i < VectorizedArrayType::size(); ++i)
     if (cells[i] != numbers::invalid_unsigned_int)
       out[i] = array[cells[i] / VectorizedArrayType::size()]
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -3839,7 +3839,8 @@ FEEvaluationBase<dim, n_components_, Number, is_face, VectorizedArrayType>::
   const auto cells = this->get_cell_ids();
 
   // 2) actually gather values
-  VectorizedArrayType out = make_vectorized_array<Number>(Number(1.));
+  VectorizedArrayType out =
+    make_vectorized_array<VectorizedArrayType>(Number(1.));
```
peterrum, line 3843: I am wondering if can write `VectorizedArrayType out = Number(1.);`.

DavidSCN: The compiler doesn't complain, so it seems to be possible. Do you want me to change it?

peterrum: Yes. The copy assignment is a rather new feature.

peterrum: May I ask you to also fix this line:

https://github.com/dealii/dealii/blob/0c059d559e7334a1719068e6c08d5bf1e6bfc583/include/deal.II/matrix_free/fe_evaluation.h#L6060

Thanks!

jppelteret:
In many places internal to the library, we've used `internal::NumberType<Number>::value(0.0)` etc. to do similar conversions (thereby ensuring that the correct constructor is called). So I believe that `VectorizedArrayType half = internal::NumberType<VectorizedArrayType>(0.5)` should work.

PR #10526, opened 2020-06-13
Fix hp_choose_p_over_h test

peterrum:
FYI @marcfehling 

references #10530


marcfehling:
Thank you for letting me know! Did this particular test fail for you at some point?

I wanted to repartition the mesh after global refinement happened, so that I have some kind of checkerboard partitioning to stress-test the algorithm. This wouldn't be the case after your change, am I right?

peterrum:
I'll close this PR. I think it will be fixed by #10529 anyway.

PR #10529, opened 2020-06-13
Fix number_cache.ghost_owners

peterrum:
The following code to set up `parallel::TriangulationBase::number_cache::ghost_owners` only works if there is a single layer of ghost cells (which is the normal case):

https://github.com/dealii/dealii/blob/116256d3e3ac6993e6e5b30da6ee39843a6d298f/source/distributed/tria_base.cc#L167-L169

However, in the case of `p:s:t` this might not be the case. Example (run with 2 processes):
```cpp
using namespace dealii;

int
main(int argc, char **argv)
{

  const int dim = 2;
  
  parallel::shared::Triangulation<dim> tria(
    MPI_COMM_WORLD,
    ::Triangulation<dim>::none,
    false /*all cells keep the subdomain id!!!*/,
    parallel::shared::Triangulation<dim>::partition_metis);
  
  // this function actually uses ghost_owners!!!
  GridTools::exchange_cell_data_to_ghosts<types::global_cell_index>(
      tria,
      [](const auto &cell) {(void) cell; return 0; },
      [](const auto &cell, const auto &id) {(void) cell; (void) id;});
}
```

This PR forces that only a single layer of ghost cells is processed for determining `ghost_owners`!

references #10530

### DIFF
--- a/source/distributed/fully_distributed_tria.cc
+++ b/source/distributed/fully_distributed_tria.cc
@@ -391,6 +391,18 @@ namespace parallel
 
 
 
+    template <int dim, int spacedim>
+    void
+    Triangulation<dim, spacedim>::add_periodicity(
+      const std::vector<dealii::GridTools::PeriodicFacePair<cell_iterator>>
+        &periodicity_vector)
+    {
+      dealii::Triangulation<dim, spacedim>::add_periodicity(periodicity_vector);
+      this->update_number_cache();
+    }
+
+
+
   } // namespace fullydistributed
 } // namespace parallel
 
--- a/source/distributed/tria_base.cc
+++ b/source/distributed/tria_base.cc
@@ -164,9 +164,14 @@ namespace parallel
 
     {
       // find ghost owners
-      for (const auto &cell : this->active_cell_iterators())
-        if (cell->is_ghost())
-          number_cache.ghost_owners.insert(cell->subdomain_id());
+      {
+        const auto vertices_with_ghost_neighbors =
+          GridTools::compute_vertices_with_ghost_neighbors(*this);
+        for (const auto &vertex_with_ghost_neighbors :
+             vertices_with_ghost_neighbors)
+          for (const auto &ghost_neigbor : vertex_with_ghost_neighbors.second)
+            number_cache.ghost_owners.insert(ghost_neigbor);
+      }
 
       Assert(number_cache.ghost_owners.size() <
                Utilities::MPI::n_mpi_processes(mpi_communicator),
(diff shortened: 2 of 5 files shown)


source/distributed/fully_distributed_tria.cc
```diff
@@ -391,6 +391,18 @@ namespace parallel
...
 
 
+    template <int dim, int spacedim>
+    void
+    Triangulation<dim, spacedim>::add_periodicity(
+      const std::vector<dealii::GridTools::PeriodicFacePair<cell_iterator>>
+        &periodicity_vector)
+    {
+      dealii::Triangulation<dim, spacedim>::add_periodicity(periodicity_vector);
+      this->update_number_cache();
```
bangerth, line 401: I think I get why these changes are necessary.

source/distributed/tria_base.cc
```diff
@@ -164,9 +164,14 @@ namespace parallel
...
-      for (const auto &cell : this->active_cell_iterators())
-        if (cell->is_ghost())
-          number_cache.ghost_owners.insert(cell->subdomain_id());
+      {
+        const auto vertices_with_ghost_neighbors =
+          GridTools::compute_vertices_with_ghost_neighbors(*this);
+        for (const auto &vertex_with_ghost_neighbors :
+             vertices_with_ghost_neighbors)
+          for (const auto &ghost_neigbor : vertex_with_ghost_neighbors.second)
+            number_cache.ghost_owners.insert(ghost_neigbor);
```
bangerth, line 173: But I don't understand why you need this. If you had a `p::s::T` that corresponds to a long string of cells, subdivided into three partitions, then on processor 0 the only immediately neighboring ghost cells belong to processor 1, but you also have ghost cells from processor 2 -- they just don't share a vertex with the locally owned cells. Why do you not want to consider them here? These are still ghost cells, so they should be reflected in the ghost owners, right?

If the issue is that there are *users* of the `number_cache.ghost_owners` set that only want to send stuff to immediate neighbors, then that's a separate issue -- in that case, we need a separate field -- but this seems like the wrong place to address this.

peterrum: I agree, we can keep `ghost_owners` as it is and introduce a new vector (maybe: `direct_ghost_owners`).

One question: normally, both are identical; exception: p:s:t without artificial cells. What is the right place to fill `direct_ghost_owners` for latter. Here, in `parallel::TriangulationBase` or in `p:s:t`?

bangerth: Another place where you'd get a difference is if we ever have a triangulation that allows us to have more than one layer of ghost cells.

If you can provide the functionality in a base class for all derived classes, then that's worth it. It might cost a bit more time for derived classes where this would be cheap to compute (p::d::T right now), but this is computed at refinement time once. The cost is almost certainly completely negligible to everything else we do after mesh refinement.

peterrum:
I am closing this one in favor of #11868. I find it a bit odd that the cache would only have information regarding a single ghost halo layer even in the case of `p:s:T` without artificial cells.

PR #10532, opened 2020-06-15
Emit warning when AppleClang < 9.0 is detected

### LINKED ISSUE #10531: shared_timed_mutex not available on older versions of MacOS
```
../include/deal.II/base/thread_local_storage.h:253:10: error: 'shared_timed_mutex' is unavailable: introduced in macOS 10.12
    std::shared_timed_mutex insertion_mutex;
         ^
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/shared_mutex:202:58: note: 'shared_timed_mutex' has been explicitly marked unavailable here
class _LIBCPP_TYPE_VIS _LIBCPP_AVAILABILITY_SHARED_MUTEX shared_timed_mutex
```

I just rolled back to version 9.2, which is OK for now. @tamiko I think that you introduced this change a short while ago. Is there an easy(ish) workaround for this? I know that my OS is old (circa Sept 2015), so if the answer is "no" then I recognise that I might just need to upgrade everything. I wonder, though, if there are users that might hit the same issue.

jppelteret:
Closes  #10531

### DIFF
--- a/cmake/setup_compiler_flags_gnu.cmake
+++ b/cmake/setup_compiler_flags_gnu.cmake
@@ -36,6 +36,18 @@ IF( CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND
     )
 ENDIF()
 
+# Correspondence between AppleClang version and upstream Clang version:
+# https://en.wikipedia.org/wiki/Xcode#Xcode_7.0_-_11.x_(since_Free_On-Device_Development)
+IF (POLICY CMP0025)
+  IF( CMAKE_CXX_COMPILER_ID MATCHES "AppleClang" AND
+      CMAKE_CXX_COMPILER_VERSION VERSION_LESS "9.0" )
+    MESSAGE(WARNING "\n"
+      "deal.II requires support for features of C++14 that are not present in\n"
+      "versions of AppleClang prior to 9.0."
+      )
+  ENDIF()
+ENDIF()
+
 
 ########################
 #                      #
--- a/cmake/setup_compiler_flags_gnu.cmake
+++ b/cmake/setup_compiler_flags_gnu.cmake
@@ -21,10 +21,10 @@
 #
 
 IF( CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND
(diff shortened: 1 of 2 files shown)


jppelteret:
Is there a specific reason as to why this only issues a warning, rather than an error message? I'd have thought it kind of pointless continuing with the build if its destined to fail? This warning message gets buried quite quickly by all of the other CMake output.

masterleinad:
I would be fine with turning it into an error.

jppelteret:
OK thanks, I've now made all three checks return errors instead of warnings. Let's see if there's any disagreement with that change :-)

jppelteret:
It might also be necessary to enable `POLICY CMP0025` for this to work unconditionally when CMake < 3.0. This seems to be what's implied in [this discussion](https://gitlab.kitware.com/cmake/cmake/-/issues/19067).

cmake/setup_compiler_flags_gnu.cmake
```diff
@@ -22,20 +22,30 @@
 
 IF( CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND
     CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9" )
-  MESSAGE(WARNING "\n"
+  MESSAGE(FATAL_ERROR "\n"
     "deal.II requires support for features of C++14 that are not present in\n"
     "versions of GCC prior to 4.9."
```
tamiko, line 27: You can upgrade this to "5.0".

jppelteret: Just checking: Did we come to this agreement during the hackathon? If so, then this part of the documentation is outdated and I'll add a commit that updates it as well.

jppelteret: I went ahead and pushed a change for this, and amended what I had previously with some other documentation changes along with the addition of the CMake policy check. We have a similar check in place a little further down in `setup_compiler_flags_gnu.cmake`

masterleinad:
Looks like you introduced some whitespace the indentation checker doesn't like.

PR #10536, opened 2020-06-15
Particles as indexable.

luca-heltai:
Allow Particles::Particle and Particles::ParticleAccessor to be used directly in RTree objects.


luca-heltai:
> Presumably this works because you have previously defined Point to be indexable? Is there a problem if you happen to have two particles at exactly the same location?

> Alternatively, you could make the particle ID indexable.

`indexable` in boost terms means "something that has the traits of a geometrical object, i.e., point, box, segment, polygon, polyline, etc.". Any of these can be a leaf in a RTree. An index is not enough. The RTree is a spatial tree, that requires the notion of a location (any class that satisfies the geometry concept of boost will do: our Point, BoundingBox, and Segment are examples of this).

No problems happen if you have two particles at the same location.

Issue #10333, opened 2020-05-23
Merge DoFHandler and hp::DoFHandler - follow up tasks


peterrum:
Following tasks need to done after PR #10328 has been merged (and before release 9.4):
- [x] remove template parameter `DoFHandlerType` (EDIT @marcfehling #13808)
- [x] do not instantiate `hp::DoFHandler` in `templates-arguments.in`
- [x] go through documentation
- [x] update tests so that they don't use `hp::DoFHandler`
- [x] deprecated `hp::DoFHandler`
- [ ] compress indices
- [x] how to determine normal or hp mode: construction time with enum/bool or dynamic?

peterrum:
Let's collect here places, where it is not trivial to remove `DoFHandlerType`. Maybe someone has a good workaround :)

https://github.com/dealii/dealii/blob/787848f09ac650e91530684647ea4cf98903acd4/include/deal.II/distributed/solution_transfer.h#L231-L234

https://github.com/dealii/dealii/blob/787848f09ac650e91530684647ea4cf98903acd4/include/deal.II/dofs/dof_accessor.h#L206-L207

peterrum:
I think we made good progress here. The open points are, as far as I see, relevant for release 10.0.

peterrum:
I think we can close this one after the release after the deprecated classes have been removed!

marcfehling:
> I think we can close this one after the release after the deprecated classes have been removed!

Indeed! However, there is still "compress indices" left in the original list of things to do.

PR #10538, opened 2020-06-16
Remove DoFHandlerType from MatrixFree::reinit()

peterrum:
References https://github.com/dealii/dealii/pull/10334#issuecomment-644603560 and #10333.

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -556,10 +556,10 @@ public:
    * not allowed. In that case, use the initialization function with
    * several DoFHandler arguments.
    */
-  template <typename DoFHandlerType, typename QuadratureType, typename number2>
+  template <typename QuadratureType, typename number2>
   void
   reinit(const Mapping<dim> &              mapping,
-         const DoFHandlerType &            dof_handler,
+         const DoFHandler<dim> &           dof_handler,
          const AffineConstraints<number2> &constraint,
          const QuadratureType &            quad,
          const AdditionalData &            additional_data = AdditionalData());
@@ -568,9 +568,9 @@ public:
    * Initializes the data structures. Same as above, but using a $Q_1$
    * mapping.
    */
-  template <typename DoFHandlerType, typename QuadratureType, typename number2>
+  template <typename QuadratureType, typename number2>
   void
-  reinit(const DoFHandlerType &            dof_handler,
+  reinit(const DoFHandler<dim> &           dof_handler,
          const AffineConstraints<number2> &constraint,
          const QuadratureType &            quad,
          const AdditionalData &            additional_data = AdditionalData());
@@ -582,10 +582,10 @@ public:
    * more. Use the reinit function without index set argument to choose the
    * one provided by DoFHandler::locally_owned_dofs().
    */
-  template <typename DoFHandlerType, typename QuadratureType, typename number2>
+  template <typename QuadratureType, typename number2>
   DEAL_II_DEPRECATED void
   reinit(const Mapping<dim> &              mapping,
-         const DoFHandlerType &            dof_handler,
+         const DoFHandler<dim> &           dof_handler,
          const AffineConstraints<number2> &constraint,
          const IndexSet &                  locally_owned_dofs,
          const QuadratureType &            quad,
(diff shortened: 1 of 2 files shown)


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -612,10 +612,21 @@ class MatrixFree : public Subscriptor
...
   reinit(const Mapping<dim> &                                   mapping,
-         const std::vector<const DoFHandlerType *> &            dof_handler,
+         const std::vector<const DoFHandler<dim> *> &           dof_handler,
+         const std::vector<const AffineConstraints<number2> *> &constraint,
+         const std::vector<QuadratureType> &                    quad,
+         const AdditionalData &additional_data = AdditionalData());
+
+  /**
+   * Initializes the data structures. Same as above, but  using hp::DoFHandlers.
+   */
```
masterleinad, line 625:
```suggestion
     @deprecated Use the overload taking a DoFHandler object instead.
   */
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -624,9 +635,19 @@ class MatrixFree : public Subscriptor
...
+  template <typename QuadratureType, typename number2>
+  void
+  reinit(const std::vector<const DoFHandler<dim> *> &           dof_handler,
+         const std::vector<const AffineConstraints<number2> *> &constraint,
+         const std::vector<QuadratureType> &                    quad,
+         const AdditionalData &additional_data = AdditionalData());
+
+  /**
+   * Initializes the data structures. Same as above, but  using hp::DoFHandlers.
+   */
```
masterleinad, line 647:
```suggestion
     @deprecated Use the overload taking a DoFHandler object instead.
   */
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -638,10 +659,22 @@ class MatrixFree : public Subscriptor
...
-         const std::vector<const DoFHandlerType *> &            dof_handler,
+         const std::vector<const DoFHandler<dim> *> &           dof_handler,
+         const std::vector<const AffineConstraints<number2> *> &constraint,
+         const std::vector<IndexSet> &      locally_owned_set,
+         const std::vector<QuadratureType> &quad,
+         const AdditionalData &             additional_data = AdditionalData());
+
+  /**
+   * Initializes the data structures. Same as above, but  using hp::DoFHandlers.
+   */
```
masterleinad, line 673:
```suggestion
     @deprecated Use the overload taking a DoFHandler object instead.
   */
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -654,10 +687,21 @@ class MatrixFree : public Subscriptor
...
   reinit(const Mapping<dim> &                                   mapping,
-         const std::vector<const DoFHandlerType *> &            dof_handler,
+         const std::vector<const DoFHandler<dim> *> &           dof_handler,
+         const std::vector<const AffineConstraints<number2> *> &constraint,
+         const QuadratureType &                                 quad,
+         const AdditionalData &additional_data = AdditionalData());
+
+  /**
+   * Initializes the data structures. Same as above, but  using hp::DoFHandlers.
+   */
```
masterleinad, line 700:
```suggestion
     @deprecated Use the overload taking a DoFHandler object instead.
   */
```

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -666,9 +710,19 @@ class MatrixFree : public Subscriptor
...
   void
-  reinit(const std::vector<const DoFHandlerType *> &            dof_handler,
+  reinit(const std::vector<const DoFHandler<dim> *> &           dof_handler,
+         const std::vector<const AffineConstraints<number2> *> &constraint,
+         const QuadratureType &                                 quad,
+         const AdditionalData &additional_data = AdditionalData());
+
+  /**
+   * Initializes the data structures. Same as above, but  using hp::DoFHandlers.
+   */
```
masterleinad, line 722:
```suggestion
     @deprecated Use the overload taking a DoFHandler object instead.
   */
```

include/deal.II/matrix_free/matrix_free.h
peterrum: You are right! I have added the comments!

peterrum:
The two places with `DoFHandlerType` in `matrix_free.h` remaining:

https://github.com/dealii/dealii/blob/57525d52611bae4e1f1517c6f32201722229995e/include/deal.II/matrix_free/matrix_free.h#L1708-L1718

https://github.com/dealii/dealii/blob/57525d52611bae4e1f1517c6f32201722229995e/include/deal.II/matrix_free/matrix_free.h#L2822-L2858

My guess is that the second one can go. What about the first one?

kronbichler:
They can both go away; the first is however an incompatibility in terms of the request by the user, so the question is whether we simply say we ignore that or do deprecation and then have another interface instead. I think the feature is rather new and the non-default dof handler is not common, so it would probably not hurt too much. On the other hand, does it hurt to keep it? Not very much I would say.

peterrum:
> On the other hand, does it hurt to keep it? Not very much I would say.

I agree. But, nevertheless, the goal till the next release should be to try to eliminate all possible `DoFHandlerTypes` as long as they not cause any major incompatibility (see also #10333).

PR #10540, opened 2020-06-17
fix bug in fast version of GridTools::find_active_cell_around_point()

nfehn:
Same as PR #10456, now for the fast version of the function ...

### DIFF
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -4623,7 +4623,10 @@ namespace GridTools
               break;
             }
           catch (...)
-            {}
+            {
+              if (counter == n_points)
+                return cell_qpoint_map;
+            }
 
         auto last_cell = cell_qpoint_map.emplace(std::make_pair(
           point_and_reference_location.first,
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -1682,6 +1682,7 @@ namespace GridTools
               {
                 auto cell = vertex_to_cells[closest_vertex_index].begin();
                 std::advance(cell, neighbor_permutation[i]);
+
                 if (!(*cell)->is_artificial())
                   {
                     const Point<dim> p_unit =
@@ -4605,14 +4606,30 @@ namespace GridTools
           return cell_qpoint_map;
 
         // We begin by finding the cell/transform of the first point
-        auto point_and_reference_location =
-          GridTools::find_active_cell_around_point(cache, points[0]);
+        std::pair<typename Triangulation<dim, spacedim>::active_cell_iterator,
+                  Point<dim>>
+          point_and_reference_location;
+
+        unsigned int counter = 0;
+
+        while (counter < n_points)
+          try
+            {
+              unsigned int i = counter;
+              ++counter;
+
+              point_and_reference_location =
+                GridTools::find_active_cell_around_point(cache, points[i]);
+              break;
+            }
+          catch (...)
+            {}
 
         auto last_cell = cell_qpoint_map.emplace(std::make_pair(
           point_and_reference_location.first,
           std::make_pair(
             std::vector<Point<dim>>{point_and_reference_location.second},
-            std::vector<unsigned int>{0})));
+            std::vector<unsigned int>{counter - 1})));
 
         // Now the second easy case.
         if (n_points == 1)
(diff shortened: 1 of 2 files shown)


peterrum:
I have run locally the test suite. Could you have a look at the following failing tests:
```
	2335 - grid/distributed_compute_point_locations_02.mpirun=3.debug (Failed)
	2337 - grid/distributed_compute_point_locations_03.mpirun=3.debug (Failed)
	3782 - mpi/fe_field_function_03.mpirun=3.debug (Failed)
	4726 - particles/generators_07.mpirun=3.debug (Failed)
```
Thanks!

nfehn:
It looks like `distributed_compute_point_locations()` causes the problems, which internally calls `find_active_cell_around_point()`. Now that we have the if-statement in `find_active_cell_around_point()`, the function `distributed_compute_point_locations()` seems to return fewer cells than before?! The docu of this function says it only computes locally owned points, so I don't know why the change of this PR should affect this function?

peterrum:
I got it running with the following modifications:
```diff
diff --git a/source/grid/grid_tools.cc b/source/grid/grid_tools.cc
index 535e1d4..183f621 100644
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -1682,6 +1682,10 @@ namespace GridTools
               {
                 auto cell = vertex_to_cells[closest_vertex_index].begin();
                 std::advance(cell, neighbor_permutation[i]);
+
+                if ((*cell)->is_artificial())
+                  continue;
+
                 const Point<dim> p_unit =
                   mapping.transform_real_to_unit_cell(*cell, p);
                 if (GeometryInfo<dim>::is_inside_unit_cell(p_unit, tolerance))
@@ -4600,14 +4604,30 @@ namespace GridTools
           return cell_qpoint_map;
 
         // We begin by finding the cell/transform of the first point
-        auto point_and_reference_location =
-          GridTools::find_active_cell_around_point(cache, points[0]);
+        std::pair<typename Triangulation<dim, spacedim>::active_cell_iterator,
+                  Point<dim>>
+          point_and_reference_location;
+
+        unsigned int counter = 0;
+
+        while (counter < n_points)
+          try
+            {
+              unsigned int i = counter;
+              ++counter;
+
+              point_and_reference_location =
+                GridTools::find_active_cell_around_point(cache, points[i]);
+              break;
+            }
+          catch (...)
+            {}
 
         auto last_cell = cell_qpoint_map.emplace(std::make_pair(
           point_and_reference_location.first,
           std::make_pair(
             std::vector<Point<dim>>{point_and_reference_location.second},
-            std::vector<unsigned int>{0})));
+            std::vector<unsigned int>{counter - 1})));
 
         // Now the second easy case.
         if (n_points == 1)
@@ -4619,17 +4639,31 @@ namespace GridTools
                                (0.5 + std::numeric_limits<double>::epsilon());
 
         // Cycle over all points left
-        for (unsigned int p = 1; p < n_points; ++p)
+        for (unsigned int p = counter; p < n_points; ++p)
           {
             // Checking if the point is close to the cell center, in which
             // case calling find active cell with a cell hint
             if (cell_center.distance(points[p]) < cell_diameter)
-              point_and_reference_location =
-                GridTools::find_active_cell_around_point(
-                  cache, points[p], last_cell.first->first);
+              try
+                {
+                  point_and_reference_location =
+                    GridTools::find_active_cell_around_point(
+                      cache, points[p], last_cell.first->first);
+                }
+              catch (...)
+                {
+                  continue;
+                }
             else
-              point_and_reference_location =
-                GridTools::find_active_cell_around_point(cache, points[p]);
+              try
+                {
+                  point_and_reference_location =
+                    GridTools::find_active_cell_around_point(cache, points[p]);
+                }
+              catch (...)
+                {
+                  continue;
+                }
 
             if (last_cell.first->first == point_and_reference_location.first)
               {
@@ -4679,8 +4713,8 @@ namespace GridTools
                                         map_entry.second.first.size()));
             inserted_points += map_entry.second.second.size();
           }
-        Assert(inserted_points == n_points,
-               ExcDimensionMismatch(inserted_points, n_points));
+          // Assert(inserted_points == n_points,
+          //       ExcDimensionMismatch(inserted_points, n_points));
 #endif
         return cell_qpoint_map;
       }
```

@gassmoeller May I ask your opinion!? In particular, I am not sure why I had to disable the assert...

gassmoeller:
I am not sure about the inner workings of this function and in particular the relations between this function and the other search functions. But is it possible that you need to comment the Assert because previously this function was assumed to find cells for all the input points, but with the change the function will skip points in artificial cells (therefore the returned number of points is not equal to the number of input points)?

What I do not understand is why we are changing this function at all (and the one in #10456 as well). The documentation of both of the functions states: 
`If the point requested does not lie in a locally-owned or ghost cell, then this function throws an exception of type GridTools::ExcPointNotFound.` and `If a point is not found inside the mesh, or is lying inside an artificial cell of a parallel::TriangulationBase, an exception is thrown.` So throwing an exception is exactly what the function should do when encountering artificial cells. Or is the problem that the function throws exceptions even when the point does not lie in an artificial cell?

nfehn:
I think the problematic sentence in the documentation mentioned above is

> is lying inside an artificial cell

This insinuates that for an artificial cell, one is able to check whether a given point lies inside this cell. However, in general one does not have the information to check whether a given point lies inside a given artificial cell (the information needed to check this is only stored for locally owned and ghost cells if I understood it correctly). So I would say the formulation in the documentation is incorrect. The documentation of which function is this exactly?

peterrum:
The following tests are still failing :(

```
        3782 - mpi/fe_field_function_03.mpirun=3.debug (Failed)
        4726 - particles/generators_07.mpirun=3.debug (Failed)
```

peterrum:
This should help:
```diff
diff --git a/source/grid/grid_tools.cc b/source/grid/grid_tools.cc
index fdf7bb6..5b005fe 100644
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -4623,7 +4623,10 @@ namespace GridTools
               break;
             }
           catch (...)
-            {}
+            {
+              if (counter == n_points)
+                return cell_qpoint_map; 
+            }
 
         auto last_cell = cell_qpoint_map.emplace(std::make_pair(
           point_and_reference_location.first,
```

peterrum:
@nfehn I'll rerun the complete test suite locally again!

peterrum:
All tests have passed locally. 

/rebuild

PR #10547, opened 2020-06-18
Avoid accessing artificial cells in ParticleHandler

masterleinad:
Fixes https://cdash.43-1.org/testDetails.php?test=47854924&build=7575.


masterleinad:
> Have you run all particle test as well?

Yes

PR #10549, opened 2020-06-18
Use neutral terminology where possible

jppelteret:
As a follow up to #10539, let's replace all instances of oppressive language with a more neutral counterpart.


drwells:
I thought about this some more and I think its a good opportunity to not just avoid problematic language but also use clearer terms. Could we update `slave` to `dependent` in the two source files you changed? That makes it clear that the DoFs are, well, dependent.

drwells:
It looks like you force-pushed an alternative fix as I wrote that comment - nevermind!

jppelteret:
> I think its a good opportunity to not just avoid problematic language but also use clearer terms

Agreed. Let's consider what I've done here a rough first attempt at achieving this.

> Could we update slave to dependent in the two source files you changed?

Good suggestion! I think that I chose `primary/secondary` (pushed at the same time that you commented). But I'd be happy to change to dependent. I'm going to wait to see if there are some other opinions, so that I could do the modifications (hopefully) just once.

masterleinad:
Indentation... :wink:

jppelteret:
> Indentation

Thanks! Although earlier I had to go before I could do this, I think I'll hold off until the end before indenting. It might make initial perusal of the PR easier.

include/deal.II/base/data_out_base.h
```diff
@@ -2708,9 +2708,9 @@ class DataOutInterface
    * parallelize visualization. In that case, you need a
    * <code>.pvtu</code> file that describes which VTU files (written, for
    * example, through the DataOutInterface::write_vtu() function) form a group.
-   * The current function can generate such a master record.
+   * The current function can generate such a centralized record.
    *
-   * The master record file generated by this function
+   * The main record file generated by this function
```
tjhei, line 2713: You seem to replace master seemingly randomly by "parent", "centralized", and "main" here (where it relates to the pvtu files). I find that confusing.

source/base/data_out_base.cc
```diff
@@ -7079,11 +7079,11 @@ DataOutInterface<dim, spacedim>::write_vtu_with_pvtu_record(
           filename_vector.emplace_back(filename);
         }
 
-      std::ofstream master_output((directory + filename_master).c_str());
-      this->write_pvtu_record(master_output, filename_vector);
+      std::ofstream main_output((directory + filename_parent).c_str());
```
tjhei, line 7082: again, you are mixing "main" and "parent" here. ``filename_master`` and ``master_output`` refer to the same thing.

source/distributed/grid_refinement.cc
```diff
@@ -241,15 +241,15 @@ namespace internal
                                            global_min_and_max.second};
             adjust_interesting_range(interesting_range);
 
-            const unsigned int master_mpi_rank = 0;
+            const unsigned int parent_mpi_rank = 0;
```
tjhei, line 244: calling rank 0 in MPI the "parent" is confusing to me. In the MPI world people often use "root" instead.

doc/users/gdb.html
```diff
@@ -59,7 +59,7 @@ <h2>Setting up GDB for deal.II and other sets of printers</h2>
       word <code>end</code>) of the
       file <code>contrib/utilities/dotgdbinit.py</code> into the directory from
       the first step. Rename this file to <code>deal.py</code> so that we can
-      import it into the master GDB file in the next step.
+      import it into the primary GDB file in the next step.
```
bangerth, line 62:
```suggestion
      import it into the principal GDB configuration file in the next step.
```

examples/step-67/step-67.cc
```diff
@@ -2092,7 +2092,7 @@ namespace Euler_DG
   // than the standard library's `std::ofstream` variants used in most other
   // tutorial programs. A particularly nice feature of the
   // `write_vtu_in_parallel()` function is the fact that it can combine output
-  // from all MPI ranks into a single file, obviating a VTU master file (the
+  // from all MPI ranks into a single file, obviating a VTU parent file (the
```
bangerth, line 2095:
```suggestion
  // from all MPI ranks into a single file, making it unnecessary to have a central record of all such files (namely, the
```

include/deal.II/base/data_out_base.h
```diff
@@ -2196,7 +2196,7 @@ namespace DataOutBase
    * piece_names[2].emplace_back("subdomain_01.time_step_2.vtk");
    * piece_names[2].emplace_back("subdomain_02.time_step_2.vtk");
    *
-   * std::ofstream visit_output ("master_file.visit");
+   * std::ofstream visit_output ("parent_file.visit");
```
bangerth, line 2199: Here and below, let's just call this file `solution.visit`. That's consistent with what we call it in most of the tutorial programs.

source/distributed/grid_refinement.cc
bangerth: Yes, I think "root" is better.

source/distributed/grid_refinement.cc
```diff
@@ -324,15 +324,15 @@ namespace internal
...
```
bangerth, line 327: same here

source/distributed/grid_refinement.cc
```diff
@@ -277,10 +277,10 @@ namespace internal
                 // now adjust the range. if we have too many cells, we take the
                 // upper half of the previous range, otherwise the lower half.
                 // if we have hit the right number, then set the range to the
-                // exact value. slave nodes also update their own
+                // exact value. child nodes also update their own
                 // interesting_range, however their results are not significant
                 // since the values will be overwritten by MPI_Bcast from the
-                // master node in next loop.
+                // parent node in next loop.
```
bangerth, line 283: I'd call this the "root node", following the discussion above. Instead of "child nodes", that would call for using the word "non-root nodes"

source/distributed/grid_refinement.cc
```diff
@@ -376,17 +376,17 @@ namespace internal
...
 
                 // now adjust the range. if we have too many cells, we take the
                 // upper half of the previous range, otherwise the lower half.
                 // if we have hit the right number, then set the range to the
-                // exact value. slave nodes also update their own
+                // exact value. child nodes also update their own
                 // interesting_range, however their results are not significant
                 // since the values will be overwritten by MPI_Bcast from the
-                // master node in next loop.
+                // parent node in next loop.
```
bangerth, line 389: same here

include/deal.II/base/data_out_base.h
jppelteret: I changed this to "central record", since that now coincides with @bangerth's suggested change below.

source/base/data_out_base.cc
jppelteret: Much later on I changed the tutorial programs to use `pvtu_output` and `pvtu_filename`. I've now done that here to be consistent.

source/distributed/grid_refinement.cc
jppelteret: Thanks for the suggestion. I've now used "root" everywhere when referring to the MPI rank.

jppelteret:
Thanks @tjhei and @bangerth for your comments. I've left the fixup commit as a stand-alone one for now, and will squash all of the commits into one once I'm able to run the indentation script on this PR.

tests/fe/nedelec_crash_hp.cc
```diff
@@ -18,7 +18,7 @@
 // interpolation matrices from lower to higher order elements. This
 // consequently led to a situation where
 // DoFTools::make_hanging_node_constraints got into trouble because it could
-// not find a master DoF for one particular slave DoF.
+// not find a primary DoF for one particular secondary DoF.
```
tjhei, line 21:
```suggestion
// not find a primary DoF for one particular dependent DoF.
```
?

PR #10550, opened 2020-06-18
Fix detection of muParser 2.3.2.

### LINKED ISSUE #10548: muParser module doesn't work any more
I just got muParser 2.3.2 and this version no longer has

```cpp
#define MUP_VERSION _T("2.2.4");
```
but instead has (in namespace mu)
```cpp
static const string_type ParserVersion = string_type(_T("2.3.2"));
```
as a result we cannot configure with this version.

drwells:
Fixes #10548.

### DIFF
--- a/cmake/modules/FindMUPARSER.cmake
+++ b/cmake/modules/FindMUPARSER.cmake
@@ -42,9 +42,22 @@ DEAL_II_FIND_PATH(MUPARSER_INCLUDE_DIR muParserDef.h
 
 IF(EXISTS ${MUPARSER_INCLUDE_DIR}/muParserDef.h)
   FILE(STRINGS "${MUPARSER_INCLUDE_DIR}/muParserDef.h" MUPARSER_VERSION_STRING_LINE
+    # Try to match the line
+    #
+    #     #define MUP_VERSION _T("2.2.4")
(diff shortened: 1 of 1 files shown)


cmake/modules/FindMUPARSER.cmake
```diff
@@ -44,7 +44,14 @@ IF(EXISTS ${MUPARSER_INCLUDE_DIR}/muParserDef.h)
   FILE(STRINGS "${MUPARSER_INCLUDE_DIR}/muParserDef.h" MUPARSER_VERSION_STRING_LINE
     REGEX "#define MUP_VERSION _T"
     )
- 
+
+  IF("${MUPARSER_VERSION_STRING_LINE}" STREQUAL "")
+    # try again with the newer version format (starting in at least 2.3.2)
+    FILE(STRINGS "${MUPARSER_INCLUDE_DIR}/muParserDef.h" MUPARSER_VERSION_STRING_LINE
+      REGEX "string_type ParserVersion = string_type"
```
bangerth, line 51: Would you mind showing in the comment here (and for the old version above) what the expected string looks like?

drwells: Good idea. Fixed!

PR #10553, opened 2020-06-19
Update output of tests/grid/grid_tools_cache_04

peterrum:


peterrum:
> I agree with this. Can you point me to which PR this change comes from?

It was #10363. My best guess is that different recent PRs had an influence on each other.

luca-heltai:
But  #10363  introduced new functionality. It should not have changed existing one. I added a new function and a new test. I'm not sure what's happened here...

bangerth:
Hm, if we don't know what patch caused the divergence, has anyone verified that the new output is actually correct?

PR #10555, opened 2020-06-19
Remove DoFHandlerType from dof_handler_policy.h/.cc

peterrum:
... part of #10333.

### DIFF
--- a/source/dofs/dof_handler_policy.cc
+++ b/source/dofs/dof_handler_policy.cc
@@ -68,22 +68,15 @@ namespace internal
          * Update the cache used for cell dof indices on all (non-artificial)
          * active cells of the given DoFHandler.
          */
-        template <class DoFHandlerType>
+        template <int dim, int spacedim>
         void
         update_all_active_cell_dof_indices_caches(
-          const DoFHandlerType &dof_handler)
+          const DoFHandler<dim, spacedim> &dof_handler)
         {
-          typename DoFHandlerType::active_cell_iterator
-            beginc = dof_handler.begin_active(),
-            endc   = dof_handler.end();
-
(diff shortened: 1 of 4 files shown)


source/dofs/dof_handler_policy.cc
```diff
@@ -2990,9 +2988,12 @@ namespace internal
...
-          typename DoFHandlerType::cell_iterator cell =
-                                                   dof_handler.begin(level),
-                                                 endc = dof_handler.end(level);
+          typename DoFHandler<dim, spacedim>::cell_iterator cell =
+                                                              dof_handler.begin(
+                                                                level),
+                                                            endc =
+                                                              dof_handler.end(
+                                                                level);
           for (; cell != endc; ++cell)
```
kronbichler, line 2997: Can we use this:
```suggestion
          for (const auto &cell : dof_handler.cell_iterators_on_level(level))
```

peterrum:
@kronbichler Done!

source/dofs/dof_handler_policy.cc
```diff
@@ -4064,7 +4041,7 @@ namespace internal
               {
                 std::vector<dealii::types::global_dof_index> local_dof_indices;
 
-                typename DoFHandlerType::level_cell_iterator
+                typename DoFHandler<dim, spacedim>::level_cell_iterator
                   cell = dof_handler->begin(level),
                   endc = dof_handler->end(level);
```
kronbichler, line 4046: I the second I pressed `approve` I saw these objects. Can we convert the place here and a few more below to range-based for loops as well?

peterrum: Done!

PR #10557, opened 2020-06-19
Remove DoFHandlerType from FEValues templates.

drwells:
Part of #10333.


drwells:
I had the same thought but we will have to be careful to make sure that its backwards compatible.

bangerth:
I see no particular reason why we need to make sure that `DoF(Cell)Accessor` need to stay backward compatible. They should never be used in user code. (And indeed they aren't in either the `examples/` directory or the ASPECT code base.)

PR #10559, opened 2020-06-19
Remove DoFHandlerType from hp::FEValues templates.

drwells:
Same as #10557 but for `hp::FEValues`. Part of #10333.


drwells:
Its nice to write patches where we lower the total number of instantiated templates :)

PR #10560, opened 2020-06-19
Introduce n_vertices(), n_lines(), n_face() in TriaAccessor 

peterrum:
.. and use it within `TriaAccessor`.

This PR introduces the methods:
- `n_vertices()`, `vertex_indices()`
- `n_lines()`, `line_indices()`
- `n_faces()`, `face_indices()`

in `TriaAccessor`. At the moment, these methods delegate the call to `GeometryInfo`. However, this will change to be able to support different geometric entity types. 

Once this PR is merged, we should revisit the usage of `GeometryInfo` throughout the library and should look out for opportunities to replace the usage of `GeometryInfo` by the usage of the new methods.

### DIFF
--- a/include/deal.II/grid/tria_accessor.h
+++ b/include/deal.II/grid/tria_accessor.h
@@ -1606,6 +1606,49 @@ public:
     const TriaIterator<TriaAccessor<structdim, dim, spacedim>> &o) const;
 
   /**
+   * Number of vertices.
+   */
+  inline unsigned int
+  n_vertices() const;
+
+  /**
+   * Number of lines.
+   */
+  inline unsigned int
+  n_lines() const;
+
+  /**
+   * Number of faces.
+   *
+   * @note Only implemented for cells (dim==spacedim).
+   */
+  inline unsigned int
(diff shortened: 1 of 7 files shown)


include/deal.II/grid/tria_accessor.h
```diff
@@ -2766,8 +2835,7 @@ class CellAccessor : public TriaAccessor<dim, dim, spacedim>
   /**
    * Return an array of iterators to all faces of this cell.
    */
-  std::array<TriaIterator<TriaAccessor<dim - 1, dim, spacedim>>,
-             GeometryInfo<dim>::faces_per_cell>
+  std::vector<TriaIterator<TriaAccessor<dim - 1, dim, spacedim>>>
   face_iterators() const;
```
bangerth, line 2839: That's unfortunate since it requires dynamic allocation. What if you used `boost::container::small_vector` instead with a fixed size of `2^dim`?

peterrum: I like this solution! Thanks!

PR #6583, opened 2018-05-12
Rename the opencascade Manifold classes.

drwells:
A quick follow-up to #6582: we are now completely rid of `Boundary`, in both name and inheritance!

Manifolds have names ending in 'Manifold': now that Boundary is gone these names do not make sense.

For backwards compatibility, the old classes now simply inherit from the correctly-named Manifold versions. This requires marking some things that were private now as protected.

PR #10567, opened 2020-06-22
Remove deprecated OpenCASCADE classes

masterleinad:
Deprecated in https://github.com/dealii/dealii/pull/6583.


drwells:
Could we add the header back in with a stub indicating its deprecated? I am a little worried that we are going to create a bunch of errors in user projects like we did when we removed `tria_boundary_lib.h`.

PR #10572, opened 2020-06-22
Check for deprecated functions in tutorials

### LINKED ISSUE #10566: Warn about deprecated stuff in the tutorials
#10565 shows that we don't get compiler warnings when we use deprecated functions in the tutorial programs. That seems to be an oversight: We really shouldn't be using deprecated things in the tutorial programs.

What would it take to change the compiler flags so that tutorial programs aren't treated like internal source files?

masterleinad:
Fixes #10566.
For some reason, `clang-format` messes up the indentation in `include/deal.II/matrix_free/matrix_free.h` so I had to disable it.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -49,6 +49,11 @@ IF(DEAL_II_COMPONENT_EXAMPLES)
 
   IF(DEAL_II_COMPILE_EXAMPLES)
     #
+    # Make sure that there are no deprecated functions used in the tutorials.
+    #
+    STRIP_FLAG(DEAL_II_CXX_FLAGS "-Wno-deprecated-declarations")
+
+    #
     # Set up all executables:
     #
     FILE(GLOB _steps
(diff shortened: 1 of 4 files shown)


tjhei:
> For some reason, `clang-format` messes up the indentation in `include/deal.II/matrix_free/matrix_free.h` so I had to disable it.

can you try DEAL_II_DISABLE_EXTRA_DIAGNOSTICS instead?

masterleinad:
> can you try DEAL_II_DISABLE_EXTRA_DIAGNOSTICS instead?

I expect that to work but oit's a little bit of a sledgehammer approach. :slightly_smiling_face: I don't feel that strongly about it and pushed a corresponding commit.

tjhei:
> a little bit of a sledgehammer approach.

agreed, but better than ``clang-format off``, I feel.

examples/CMakeLists.txt
```diff
@@ -48,6 +48,11 @@ IF(DEAL_II_COMPONENT_EXAMPLES)
     )
 
   IF(DEAL_II_COMPILE_EXAMPLES)
+    #
+    # Make sure that there are no deprecated functions used in the tutorials.
+    #
+    STRIP_FLAG(DEAL_II_CXX_FLAGS "-Wno-deprecated-declarations")
```
bangerth, line 54: Nice -- thanks!

PR #10588, opened 2020-06-23
Add FEInterfaceValues to ScratchData.

luca-heltai:
While I was there, I fixed FEInterfaceValues for co-dim one and two.


bangerth:
The ` meshworker/scratch_data_08.debug` test fails, but other than that, this is good to merge. The error is this:
```
tests/meshworker/scratch_data_08.cc:95:28: error: ‘update_face_normal_vectors’ was not declared in this scope
                            update_face_normal_vectors | update_JxW_values;
                            ^
```

tjhei:
great, thanks!

PR #10595, opened 2020-06-24
Augment documentation of project_onto_orthogonal_tensors().

bangerth:
Follow-up to #10027. @rezarastak -- FYI.

### DIFF
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -2640,16 +2640,66 @@ cofactor(const Tensor<2, dim, Number> &t)
 
 
 /**
- * Return the nearest orthogonal matrix by
- * combining the products of the SVD decomposition: $\mathbf U \mathbf{V}^T$,
- * where $\mathbf U$ and $\mathbf V$ are computed from the SVD decomposition:
- * $\mathbf U  \mathbf S \mathbf V^T$,
- * effectively replacing $\mathbf S$ with the identity matrix.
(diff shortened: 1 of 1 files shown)


include/deal.II/base/tensor.h
```diff
@@ -2640,16 +2640,66 @@ cofactor(const Tensor<2, dim, Number> &t)
...
+ * Return the nearest orthogonal matrix
+ * $\hat {\mathbf A}=\mathbf U \mathbf{V}^T$ by
+ * combining the products of the singular value decomposition (SVD)
+ * ${\mathbf A}=\mathbf U  \mathbf S \mathbf V^T$ for a given input
+ * ${\mathbf A}$, effectively replacing $\mathbf S$ with the identity matrix.
+ *
+ * This is a (nonlinear)
+ * [projection
+ * operation](https://en.wikipedia.org/wiki/Projection_(mathematics)) since when
+ * applied twice, we have $\hat{\hat A}=\hat A$ as is easy to see. (That is
```
rezarastak, line 2652:
```suggestion
 * applied twice, we have $\hat{\hat{\mathbf A}}=\hat{\mathbf A}$ as is easy to see. (That is
```

bangerth: Ah yes, good find!

bangerth:
I've incorporated @rezarastak 's suggestion. Should be good to go now.

PR #10599, opened 2020-06-25
Remove DoFHandlerType from multigrid/mg_transfer.h

peterrum:
references #10333

### DIFF
--- a/include/deal.II/multigrid/mg_transfer.h
+++ b/include/deal.II/multigrid/mg_transfer.h
@@ -53,13 +53,13 @@ namespace internal
 
     static const bool requires_distributed_sparsity_pattern = false;
 
-    template <typename SparsityPatternType, typename DoFHandlerType>
+    template <typename SparsityPatternType, int dim, int spacedim>
     static void
     reinit(Matrix &                   matrix,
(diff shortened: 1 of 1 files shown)


include/deal.II/multigrid/mg_transfer.h
```diff
@@ -76,19 +76,18 @@ namespace internal
...
 
-    template <typename SparsityPatternType, typename DoFHandlerType>
+    template <typename SparsityPatternType, int dim, int spacedim>
     static void
     reinit(Matrix &matrix,
            Sparsity &,
            int                        level,
            const SparsityPatternType &sp,
-           DoFHandlerType &           dh)
+           DoFHandler<dim, spacedim> &dh)
```
kronbichler, line 85: I think the tester complains because you should do
```suggestion
           const DoFHandler<dim, spacedim> &dh)
```

include/deal.II/multigrid/mg_transfer.h
```diff
@@ -108,19 +107,18 @@ namespace internal
...
```
kronbichler, line 116: same here and in lines 148 and 179 below

PR #10607, opened 2020-06-26
Fix a reference entry.

bangerth:
For reasons unclear to me, some of the references in https://dealii.org/developer/doxygen/deal.II/citelist.html are not formatted right. I can't figure out the system, but I know that this patch fixes [31] at least. (And there is probably little loss in not printing the month a journal issue came out.)


bangerth:
I have not been able to see a pattern between entries 31 (now fixed) and 21/27 (still wrong). In the current case, I took a stab at something that happened to work, but I don't know what would work for the other cases...

tjhei:
> 21/27 (still wrong)

doxygen gets confused by the doi, it looks like. You can fix 21 by deleting it.

PR #10609, opened 2020-06-26
Remove DoFHandlerType from numerics/vector_tools

peterrum:
references #10333

### DIFF
--- a/include/deal.II/numerics/vector_tools_boundary.templates.h
+++ b/include/deal.II/numerics/vector_tools_boundary.templates.h
@@ -52,12 +52,11 @@ namespace VectorTools
     template <int dim,
               int spacedim,
               typename number,
-              template <int, int> class DoFHandlerType,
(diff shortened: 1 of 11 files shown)


include/deal.II/numerics/vector_tools_boundary.templates.h
```diff
@@ -219,7 +218,7 @@ namespace VectorTools
             q_collection,
             update_quadrature_points);
 
-          typename DoFHandlerType<dim, spacedim>::active_cell_iterator
+          typename DoFHandler<dim, spacedim>::active_cell_iterator
             cell = dof.begin_active(),
             endc = dof.end();
           for (; cell != endc; ++cell)
```
kronbichler, line 224: Can we have a range-based for loop here?

peterrum: Done!

PR #10621, opened 2020-06-27
DiscreteTime::set_next_step_size() added

rezarastak:
Sometimes (such as #9973) we know what the `dt` should be. We don't want the `DiscreteTime` class to potentially adjust it.

Here we introduce the member function `set_next_step_size()` as a companion to `set_desired_next_step_size()`. It is more strict in the input arguments. For example, it checks that we are not jumping over the end time. It guarantees to not change the provided `dt`.

### DIFF
--- a/source/base/discrete_time.cc
+++ b/source/base/discrete_time.cc
@@ -67,6 +67,20 @@ DiscreteTime::set_desired_next_step_size(const double next_step_size)
 
 
 void
+DiscreteTime::set_next_step_size(const double next_step_size)
+{
+  Assert(next_step_size > 0,
+         ExcMessage("Only positive time step size is allowed."));
+  next_time = current_time + next_step_size;
+  Assert(
+    next_time <= end_time,
+    ExcMessage(
+      "Time step size is too large. The next time cannot exceed the end time."));
+}
+
+
+
+void
 DiscreteTime::advance_time()
 {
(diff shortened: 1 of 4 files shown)


tests/base/discrete_time_1.cc
```diff
@@ -69,7 +69,7 @@ test_adjust_time_step_size()
   print_time(time);
   time.advance_time();
   print_time(time);
-  time.set_desired_next_step_size(0.36);
```
drwells, line 72: Note that like 77 still has `set_desired_next_step_size()` so we do not lose test coverage.

source/base/discrete_time.cc
```diff
@@ -66,6 +66,16 @@ DiscreteTime::set_desired_next_step_size(const double next_step_size)
 
 
 
+void
+DiscreteTime::set_next_step_size(const double next_step_size)
+{
+  Assert(next_step_size > 0, ExcInternalError());
+  next_time = current_time + next_step_size;
+  Assert(next_time <= end_time, ExcInternalError());
```
drwells, line 74: I didn't spot this at first - my mistake. This isn't an internal error - its an error created by incorrect user input. Would you please switch these exceptions to use `ExcMessage`?

rezarastak:
Sorry for the delay. I added better error messages. Ready for final review.

PR #10627, opened 2020-06-29
Remove DoFHandlerType from internal::DoFHandlerImplementation::Iterators

peterrum:
references #10333

### DIFF
--- a/include/deal.II/dofs/dof_iterator_selector.h
+++ b/include/deal.II/dofs/dof_iterator_selector.h
@@ -31,6 +31,9 @@ class DoFAccessor;
 template <int dim, int spacedim, bool lda>
 class DoFCellAccessor;
 
+template <int dim, int spacedim>
+class DoFHandler;
+
 template <typename Accessor>
 class TriaRawIterator;
 template <typename Accessor>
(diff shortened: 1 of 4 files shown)


include/deal.II/dofs/dof_iterator_selector.h
```diff
@@ -56,10 +59,10 @@ namespace internal
      * @ref Iterators
      * module for more information.
      */
-    template <template <int, int> class DoFHandlerType, int spacedim, bool lda>
-    struct Iterators<DoFHandlerType<1, spacedim>, lda>
+    template <int spacedim, bool lda>
+    struct Iterators<1, spacedim, lda>
     {
-      using DoFHandler_type = DoFHandlerType<1, spacedim>;
+      using DoFHandler_type = DoFHandler<1, spacedim>;
```
drwells, line 65: Would you mind making these each `DoFHandlerType`?

drwells: On second thought - could we just deprecate these? They aren't used anywhere in the library AFAICT.

peterrum: Good catch!

drwells:
I didn't realize they were in `namespace internal` - removing them is even better :+1:

peterrum:
> I didn't realize they were in namespace internal - removing them is even better +1

Internal namespace is the friend of us as developers :P

PR #10635, opened 2020-06-30
Generalize line_orientation() for 2D general meshes

peterrum:
... for triangles `TriaLevel::face_orientations` will be filled and used here.

Currently, no test can be added.


drwells:
Interesting. Why exactly do triangles need this?

PR #10631, opened 2020-06-29
Refactor Triangulation::create_triangulation()

peterrum:
This PR refactors and generalizes the function `Triangulation::create_triangulation()`. The setup routines are ready for simplex meshes as well as for mixed meshes.

The refactored routines clearly separate the setup of a (reduced) connectivity graph and the filling of internal data structures. Furthermore, they do not use any maps and use instead vectors, which are sorted if needed. Also the routines do not use any `TriaIterator`s/`TriaAccessor`s but operates directly on plain arrays within `TriaLevel`, `TriaFace`, and `TriaObjects`.

The following graph shows the performance of the new routines. There are two lines *new* and *new (compatibility)*. Both work; however for the latter, I have introduces an additional sorting step to enforce the same enumeration of entities as it is currently the case in deal.II (so that all test - even the test that depend on the order of entities - pass). I have observed for a 3D benchmark a speedup of up to 4 (in compatibility mode) and up to 5 (without additional sorting).

performance

The performance test has been run with the following simple benchmark:

```cpp
#include <deal.II/base/convergence_table.h>
#include <deal.II/distributed/tria.h>
#include <deal.II/fe/fe_q.h>
#include <deal.II/grid/grid_generator.h>

using namespace dealii;

template <int dim>
void
test(unsigned int max_size)
{
  ConvergenceTable table;

  for (unsigned int i = 1; Utilities::pow(i, dim) <= max_size; i *= 2)
    {
      unsigned int n_cells       = 0;
      unsigned int n_repetitions = 5;

      auto start = std::chrono::steady_clock::now();
      for (unsigned int j = 0; j < n_repetitions; ++j)
        {
          Triangulation<dim> basetria;
          GridGenerator::subdivided_hyper_cube(basetria, i);
          n_cells = basetria.n_cells();
        }
      auto end = std::chrono::steady_clock::now();

      std::chrono::duration<double> elapsed_seconds = end - start;

      table.add_value("time", elapsed_seconds.count() / n_repetitions);
      table.set_scientific("time", true);
      table.add_value("cells", n_cells);
    }

  table.write_text(std::cout);
}

int
main(int argc, char *argv[])
{
  const int dim      = argc >= 2 ? atoi(argv[1]) : 2;
  const int max_size = argc >= 3 ? atoi(argv[2]) : 2;

  if (dim == 2)
    test<2>(max_size);
  else if (dim == 3)
    test<3>(max_size);
}
```

[...]

PR #10637, opened 2020-06-30
Introduce ReferenceCell::Type

peterrum:
... which can be queried via ~`TriaAccessor::geometric_entity_type()`~ `TriaAccessor::reference_cell_type()`.  

This info will be used to decide which number of vertices, faces, lines, ... is returned.


include/deal.II/base/types.h
```diff
@@ -150,6 +150,11 @@ namespace types
    * @ref GlossMaterialId "Glossary entry on material indicators"
    */
   using material_id = unsigned int;
+
+  /**
+   * The type used to denote geometric entity types.
+   */
+  using geometric_entity_type = std::uint8_t;
```
bangerth, line 157: In reference to the other PR: How about `reference_cell_kind` or `reference_cell_type`?

include/deal.II/grid/geometric_entity.h
```diff
@@ -0,0 +1,46 @@
...
+
+DEAL_II_NAMESPACE_OPEN
+
+
+/**
+ * Supported geometric entity types.
+ */
+enum class GeometricEntityType : types::geometric_entity_type
+{
+  VERTEX  = 0,
```
bangerth, line 32: let's use lower-case letters. no need to shout :-)

include/deal.II/grid/tria_accessor.h
```diff
@@ -1605,6 +1606,12 @@ class TriaAccessor : public TriaAccessorBase<structdim, dim, spacedim>
   is_translation_of(
     const TriaIterator<TriaAccessor<structdim, dim, spacedim>> &o) const;
 
+  /**
+   * Geometric entity type of the current object.
+   */
+  inline GeometricEntityType
+  geometric_entity_type() const;
```
bangerth, line 1613: No need to put `inline` in front of the declaration. It's only necessary with the definition.

include/deal.II/grid/tria_accessor.h
```diff
@@ -2671,6 +2678,12 @@ class TriaAccessor<0, 1, spacedim>
   bool
   used() const;
 
+  /**
+   * Geometric entity type of the current object.
+   */
+  inline GeometricEntityType
+  geometric_entity_type() const;
```
bangerth, line 2685: same here

include/deal.II/base/types.h
bangerth: Separately, though: Where do you actually need this? I believe that the only place where you use it is in the definition of the `enum class`. Nowhere else should we actually be using this. In that case, let's either make this an internal namespace type or, better, just omit giving it a name altogether: just put it into the declaration of the `enum class` directly.

include/deal.II/grid/tria_faces.h
```diff
@@ -71,6 +72,13 @@ namespace internal
        */
       std::vector<unsigned char> quads_line_orientations;
 
+      /**
+       * Geometric entity type of each quad.
+       *
+       * @note Used only for dim=3.
+       */
+      std::vector<GeometricEntityType> quad_entity_type;
```
bangerth, line 80: Why would these not always have the same value, namely `QUAD`?

include/deal.II/base/types.h
peterrum: I am using it to cast the enum to a number in #10631 so that I can access an entry in an array. But I guess implement it there also differently!

include/deal.II/grid/tria_faces.h
peterrum: Good point: the name is chosen a bit confusing. I guess we should rename all `quad_*` at some time because they can be also triangles. Any suggestions how to abstract quadrilateral and triangles?

include/deal.II/grid/geometric_entity.h
```diff
@@ -0,0 +1,46 @@
...
+  LINE    = 1,
+  TRI     = 2,
+  QUAD    = 3,
+  TET     = 4,
+  PYRAMID = 5,
+  WEDGE   = 6,
+  HEX     = 7,
+  INVALID = static_cast<types::geometric_entity_type>(-1)
+};
```
jppelteret, line 41: Do we need to consider prisms as well? If so, I guess that it should be added between `WEDGE` and `HEX`.

peterrum: A wedge is a prism, isn't?

include/deal.II/grid/geometric_entity.h
peterrum: I am not shouting anymore ;)

include/deal.II/base/types.h
peterrum: OK, I have remove `geometric_entity_type` and rename the enum to `RefernceCell::Type`!

include/deal.II/grid/tria_faces.h
peterrum: Let's keep it like this for now. PR #10631 will fill this array also with `ReferenceCell::Types::Tri` for tets/wedges/pyramids.

Making the naming of 2 entities (quad/tri) consistent is larger clean up effort (see e.g.: `GeometryInfo::quads_per_cell`).

include/deal.II/grid/tria_accessor.h
```diff
@@ -1605,6 +1606,12 @@ class TriaAccessor : public TriaAccessorBase<structdim, dim, spacedim>
   is_translation_of(
     const TriaIterator<TriaAccessor<structdim, dim, spacedim>> &o) const;
 
+  /**
+   * Geometric entity type of the current object.
```
drwells, line 1610:
```suggestion
   * Reference cell type of the current object.
```

include/deal.II/grid/tria_accessor.h
```diff
@@ -2671,6 +2678,12 @@ class TriaAccessor<0, 1, spacedim>
   bool
   used() const;
 
+  /**
+   * Geometric entity type of the current object.
```
drwells, line 2682:
```suggestion
   * Reference cell type of the current object.
```

include/deal.II/grid/tria_levels.h
```diff
@@ -206,6 +207,13 @@ namespace internal
        */
       std::vector<unsigned char> face_orientations;
 
+      /**
+       * Geometric entity type of each cell.
```
drwells, line 211:
```suggestion
   * Reference cell type of each cell.
```

include/deal.II/grid/tria_levels.h
```diff
@@ -206,6 +207,13 @@ namespace internal
...
+       *
+       * @note Used only for dim=2 and dim=3.
+       */
+      std::vector<ReferenceCell::Type> entity_type;
```
drwells, line 215: I would prefer to rename this to something like `reference_cell_types` so that we are using clearer terminology than 'entity'.

peterrum:
@drwells Done!

PR #10639, opened 2020-06-30
Remove GeometryInfro from a few files

peterrum:


include/deal.II/dofs/dof_accessor.h
```diff
@@ -1480,8 +1480,9 @@ class DoFCellAccessor : public DoFAccessor<dimension_,
   /**
    * Return an array of iterators to all faces of this cell.
    */
-  inline std::array<face_iterator, GeometryInfo<dimension_>::faces_per_cell>
-  face_iterators() const;
+  inline boost::container::
+    small_vector<face_iterator, GeometryInfo<dimension_>::faces_per_cell>
+    face_iterators() const;
```
drwells, line 1485: I'm not sure where we should put it, but at some point we should define a `GeometryInfo` like object that defines the maximum number of geometric entities per cell. It might be enough to alias this to `GeometryInfo` itself.

drwells: It looks like this file is missing the `#include <boost/container/small_vector.hpp>`.

peterrum: > It looks like this file is missing the #include <boost/container/small_vector.hpp>.

Done!

> we should define a GeometryInfo like object that defines the maximum number of geometric entities per cell.

This is how I treat it at many places at the moment. It might interesting at some time to investigate how dynamic we can become without loosing performance!?

peterrum:
> Have we done any performance comparisons between the dynamic approach and using GeometryInfo?


No, I have not. However, currently there should be no changes in performance because nothing is actually dynamic. All the `TriaAccessor`-methods used here can be actually evaluated at compile time!

bangerth:
Part of #11879.

PR #10641, opened 2020-07-01
Intoduce ReferenceCellInfo

peterrum:
To be used internally of `TriaAccessor` like this:
```cpp
  inline const internal::GeometricEntityInfo &
  TriaAccessor::geometric_entity_info() const
  {
    static internal::GeometricEntityInfo        gei_invalid;
    static internal::GeometricEntityInfoVertex  gei_vertex;
    static internal::GeometricEntityInfoLine    gei_line;
    static internal::GeometricEntityInfoTri     gei_tri;
    static internal::GeometricEntityInfoQuad    gei_quad;
    static internal::GeometricEntityInfoTet     gei_tet;
    static internal::GeometricEntityInfoPyramid gei_pyramid;
    static internal::GeometricEntityInfoWedge   gei_wedge;
    static internal::GeometricEntityInfoHex     gei_hex;

    switch (this->geometric_entity_type()) // see: PR #10637
      {
        case GeometricEntityType::VERTEX:
          return gei_vertex;
        case GeometricEntityType::LINE:
          return gei_line;
        case GeometricEntityType::TRI:
          return gei_tri;
        case GeometricEntityType::QUAD:
          return gei_quad;
        case GeometricEntityType::PYRAMID:
          return gei_pyramid;
        case GeometricEntityType::WEDGE:
          return gei_wedge;
        case GeometricEntityType::HEX:
          return gei_hex;
        default:
          Assert(false, StandardExceptions::ExcNotImplemented());
          return gei_invalid;
      }
  }
```

### DIFF
--- a/include/deal.II/grid/tria_accessor.templates.h
+++ b/include/deal.II/grid/tria_accessor.templates.h
@@ -536,13 +536,12 @@ namespace internal
       line_index(const TriaAccessor<3, 3, 3> &accessor, const unsigned int i)
       {
         const auto pair =
-          GeometryInfo<3>::standard_hex_line_to_quad_line_index(i);
+          accessor.reference_cell_info().standard_line_to_face_and_line_index(
+            i);
         const auto quad_index = pair[0];
-        const auto line_index = GeometryInfo<3>::standard_to_real_face_line(
-          pair[1],
-          accessor.face_orientation(quad_index),
-          accessor.face_flip(quad_index),
-          accessor.face_rotation(quad_index));
+        const auto line_index =
+          accessor.reference_cell_info().standard_to_real_face_line(
+            pair[1], pair[0], face_orientation_raw(accessor, quad_index));
 
         return accessor.quad(quad_index)->line_index(line_index);
       }
@@ -605,12 +604,21 @@ namespace internal
       face_orientation(const TriaAccessor<3, 3, 3> &accessor,
                        const unsigned int           face)
       {
-        return get_bit(
+        return ReferenceCell::internal::get_bit(
           accessor.tria->levels[accessor.present_level]->face_orientations
             [accessor.present_index * GeometryInfo<3>::faces_per_cell + face],
           0 /*=orientation_bit*/);
       }
 
+
+      inline static unsigned int
+      face_orientation_raw(const TriaAccessor<3, 3, 3> &accessor,
+                           const unsigned int           face)
+      {
+        AssertIndexRange(face, accessor.n_faces());
+        return accessor.tria->levels[accessor.present_level]->face_orientations
+          [accessor.present_index * GeometryInfo<3>::faces_per_cell + face];
+      }
 
 
       /**
@@ -636,37 +644,6 @@ namespace internal
       }
 
 
-      /**
-       * Check if the bit at position @p n in @p number is set.
-       */
-      inline static bool
-      get_bit(const unsigned char number, const unsigned int n)
-      {
-        AssertIndexRange(n, 8);
-
-        // source:
-        // https://stackoverflow.com/questions/47981/how-do-you-set-clear-and-toggle-a-single-bit
-        // "Checking a bit"
-        return (number >> n) & 1U;
-      }
-
-
-
-      /**
-       * Set the bit at position @p n in @p number to value @p x.
-       */
-      inline static void
-      set_bit(unsigned char &number, const unsigned int n, const bool x)
-      {
-        AssertIndexRange(n, 8);
-
-        // source:
-        // https://stackoverflow.com/questions/47981/how-do-you-set-clear-and-toggle-a-single-bit
-        // "Changing the nth bit to x"
-        number ^= (-static_cast<unsigned char>(x) ^ number) & (1U << n);
-      }
-
-
 
       inline static bool
       face_flip(const TriaAccessor<3, 3, 3> &accessor, const unsigned int face)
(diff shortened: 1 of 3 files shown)


bangerth:
For approximately 23 years, I've regretted the choice `GeometryInfo`. A better name would have been `ReferenceCellInfo`, because that's what it really is -- it's not about the *geometry*, but about properties of the *reference cell*.

Note sure how others feel about this, but it's been something I haven't been happy with for a long time...

include/deal.II/grid/geometric_entity_info.h
```diff
@@ -0,0 +1,359 @@
...
+      return GeometryInfo<dim>::faces_per_cell;
+    }
+  };
+
+
+
+  /*
+   * VERTEX
+   */
+  struct GeometricEntityInfoVertex : public GeometricEntityInfoTensorProduct<0>
```
bangerth, line 237: instead of long names, how about putting these all into a namespace `GeometricEntities` (or corresponding to whatever the class name is going to be)?

peterrum: I would be fine with a namespace `ReferenceCellInfo`. How would you call `GeometricEntityType` in PR #10637?

peterrum: I have introduced the namespace!

drwells:
I would prefer that we merge #10637 and then use that enum as a template parameter instead of having `Quad`, `Hex`, `Tri`, etc. classes. In the long term it would be easier to have, for example,

```
template <ReferenceCellInfo::Type cell_type>
struct Info
{
};

template <ReferenceCellInfo::Type::Hex>
struct Info : public TensorProductInfo<3>
{
// etc
};
```

include/deal.II/grid/reference_cell_info.h
```diff
@@ -0,0 +1,369 @@
...
+    // source:
+    // https://stackoverflow.com/questions/47981/how-do-you-set-clear-and-toggle-a-single-bit
+    // "Changing the nth bit to x"
+    number ^= (-static_cast<unsigned char>(x) ^ number) & (1U << n);
+  }
+
+  /**
+   * A namespace for geometric information on reference cells.
+   */
+  namespace ReferenceCellInfo
```
drwells, line 62:
```suggestion
  namespace ReferenceCell
```
Lets put everything related to the reference cell in a single namespace.

drwells:
One problem that I have when using libMesh is that it is really hard to look up information based on the element type, so I have to write a bunch of annoying switch statements that really should be provided for me.. See, for example,

https://github.com/IBAMR/IBAMR/blob/master/ibtk/include/ibtk/libmesh_utilities.h#L102

include/deal.II/grid/tria_accessor.templates.h
```diff
@@ -2182,6 +2124,34 @@ TriaAccessor<structdim, dim, spacedim>::face_indices() const
 }
 
 
+
+template <int structdim, int dim, int spacedim>
+inline const internal::ReferenceCellInfo::Base &
+TriaAccessor<structdim, dim, spacedim>::reference_cell_info() const
```
drwells, line 2130: We should move this function into `reference_cell_info.h`: #10645 needs this functionality too.

peterrum: I'll work on it in 30 minutes!

drwells:
I just had a quick call with @peterrum and he convinced me that we should keep the current names and not try to add template arguments. My original reason for adding templates wouldn't help much anyway since `n_vertices()` et al are all virtual.

The things left to do now are:
1. Move the utility function to get the correct reference cell info object into the reference cell info header
2. Move these classes into `internal::ReferenceCell` or `ReferenceCell::internal`. We need to figure out what members of these classes really need to be public (e.g., users do not need to know about line orientations), but for now lets just make everything internal.

include/deal.II/grid/reference_cell_info.h
peterrum: Done!

include/deal.II/grid/tria_accessor.templates.h
peterrum: Let's keep it like this for now. I would like to introduce the option that the right `ReferenceCell::internal::Info` is used based on the `ReferenceCell::Type` in follow-up PR since I consider it a (internal) function change.

kronbichler:
I have a question regarding the namespaces: Do we intend to have public information in the `ReferenceCell` namespace at some point (sorry, I lost a bit track whether we had that reasoning elsewhere)? If not, our general policy is to wrap the namespace into `internal::ReferenceCell`.

Furthermore I would drop the changelog for now because we tend to not document changes in internal namespaces (as `internal` namespaces is implementation-reserved), unless there is a clear benefit / performance change from that. We will have plenty of positive things to report on later.

kronbichler:
After a look at #10637 I have my first question answer, but I wonder whether we could combine the information here in the `reference_cell.h` file to keep the things going to that namespace in one place (unless we plan for many more things to come).

peterrum:
> Furthermore I would drop the changelog for now because we tend to not document changes in internal namespaces (as internal namespaces is implementation-reserved), unless there is a clear benefit / performance change from that. We will have plenty of positive things to report on later.

I have dropped it.

>  but I wonder whether we could combine the information here in the reference_cell.h

I have merged both files.

kronbichler:
Thanks!

PR #10642, opened 2020-07-02
Add Simplex::ScalarPolynomial

peterrum:


include/deal.II/simplex/polynomials.h
```diff
@@ -0,0 +1,167 @@
...
+namespace Simplex
+{
+  /**
+   * Polynomials defined on dim-dimensional simplex entities. This class is
+   * basis of Simplex::FE_P.
+   */
+  template <int dim>
+  class ScalarPolynomial : public ScalarPolynomialsBase<dim>
+  {
+    static_assert(dim == 2 || dim == 3, "Dimension not supported!");
```
luca-heltai, line 39: Why is dim == 1 not supported?

peterrum:
> In what is this different from PolynomialsP?

PolynomialsP limits the polynomial order. E.g.: `PolynomialsP<2>()` has following shape functions `1`, `x`, `y` but not `xy`.

My class puts the support points into the vertices, middle of lines/faces.

source/simplex/polynomials.cc
```diff
@@ -0,0 +1,423 @@
...
+
+      return 0;
+    }
+  } // namespace
+
+  template <int dim>
+  ScalarPolynomial<dim>::ScalarPolynomial(const unsigned int degree)
+    : ScalarPolynomialsBase<dim>(degree, compute_n_polynomials(dim, degree))
+  {}
+  template <int dim>
```
drwells, line 53:
```suggestion



  template <int dim>
```

source/simplex/polynomials.cc
```diff
@@ -0,0 +1,423 @@
...
+            return 10;
+        }
+
+      Assert(false, ExcNotImplemented());
+
+      return 0;
+    }
+  } // namespace
+
+  template <int dim>
```
drwells, line 49:
```suggestion


  template <int dim>
```

drwells: We need three blank lines between function definitions - please check the rest.

peterrum:
@drwells Done!

PR #10645, opened 2020-07-02
FiniteElementData: select number of objects based on cell type

peterrum:
First step in generalizing the `FiniteElement` classes.

Number of vertices, lines, and faces are decoupled from `GeometryInfo` and depended on the cell type (default Line/Quad/Hex).

I have introduced for this purpose new internal functions `vertices_per_cell()`, ... . This is probably not the best approach, but nevertheless let's not over-engineer this at the moment. I would like to collect needed reference-cell type dependent functionalities (in the `FiniteElement` classes) during the next weeks (when we extend the number of application codes) and settle then on an appropriate data structure.

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -229,6 +229,18 @@ namespace ReferenceCell
 
           return true;
         }
+
+        /**
+         * Return reference-cell type of face @p face_no.
+         */
+        virtual ReferenceCell::Type
+        face_reference_cell_type(const unsigned int face_no) const
+        {
+          Assert(false, ExcNotImplemented());
+          (void)face_no;
+
+          return ReferenceCell::Type::Invalid;
+        }
       };
 
 
@@ -263,7 +275,14 @@ namespace ReferenceCell
        * Vertex.
        */
       struct Vertex : public TensorProductBase<0>
-      {};
+      {
+        ReferenceCell::Type
+        face_reference_cell_type(const unsigned int face_no) const override
+        {
+          (void)face_no;
+          return ReferenceCell::Type::Invalid;
+        }
+      };
 
 
 
@@ -271,7 +290,14 @@ namespace ReferenceCell
        * Line.
        */
       struct Line : public TensorProductBase<1>
-      {};
+      {
+        ReferenceCell::Type
+        face_reference_cell_type(const unsigned int face_no) const override
+        {
+          (void)face_no;
+          return ReferenceCell::Type::Vertex;
+        }
+      };
 
 
 
@@ -298,6 +324,13 @@ namespace ReferenceCell
 
           return GeometryInfo<2>::standard_to_real_line_vertex(
             vertex, line_orientation);
+        }
+
+        ReferenceCell::Type
+        face_reference_cell_type(const unsigned int face_no) const override
+        {
+          (void)face_no;
+          return ReferenceCell::Type::Line;
         }
       };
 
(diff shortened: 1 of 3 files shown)


include/deal.II/grid/reference_cell.h
```diff
@@ -0,0 +1,50 @@
...
+    Wedge   = 6,
+    Hex     = 7,
+    Invalid = static_cast<std::uint8_t>(-1)
+  };
+} // namespace RefernceCell
+
+
+DEAL_II_NAMESPACE_CLOSE
+
+#endif
```
peterrum, line 50: Part of PR #10637.

drwells:
This PR could be simplified significantly by first merging #10637 and then #10641, so I will hold off on reviewing it until those are in place.

peterrum:
The comment in https://github.com/dealii/dealii/pull/10641#discussion_r451002918 

> We should move this function into reference_cell_info.h: #10645 needs this functionality too.

We need in this class different information. In particular, we need information about the sub-entities (e.g., `vertices_per_face`).  I will think about accessing sub-entities and their `ReferenceCell::internal::Info` (and/or `Type`). However, the practical problem is that the information depend on the face we are considering. Consider for example a wedge, where the face is depending on the considered face a quad or triangle.

peterrum:
I have rebased this branch. This PR is necessary for #10646.

source/fe/fe_data.cc
```diff
@@ -19,9 +19,193 @@
...
+        case ReferenceCell::Type::Quad:
+          return GeometryInfo<2>::vertices_per_cell;
+        case ReferenceCell::Type::Tet:
+          return 4;
+        case ReferenceCell::Type::Hex:
+          return GeometryInfo<3>::vertices_per_cell;
+        default:
+          Assert(false, ExcNotImplemented());
+          return 0;
+      }
```
drwells, line 44: Would it be possible to use `internal::ReferenceCell::Base()` here instead of writing these out?

source/fe/fe_data.cc
```diff
@@ -19,9 +19,193 @@
...
+          return GeometryInfo<2>::quads_per_face;
+        case ReferenceCell::Type::Tet:
+          return 1;
+        case ReferenceCell::Type::Hex:
+          return GeometryInfo<3>::quads_per_face;
+        default:
+          Assert(false, ExcNotImplemented());
+          return 0;
+      }
+  }
```
peterrum, line 183: @drwells The problem are these functions. These access information of the faces, which is not (and I don't want it to be come) part of the interface of `ReferenceCell::Info` (because it is not extendable to mixed meshes). Maybe I could add to the interface a function `get_face_reference_type(unsigned int face)`, which would return the right `ReferenceCell::Type` for which we could query here the number of (sub)entities. What do you think?

This would also for mixed meshes.

include/deal.II/grid/reference_cell.h
```diff
@@ -381,8 +414,53 @@ namespace ReferenceCell
...
+
+      /**
+       * Return for a given reference-cell type @p and face number @p face_no the
+       * right Info of the @p face_no-th face.
+       */
+      inline const ReferenceCell::internal::Info::Base &
+      get_face(const ReferenceCell::Type &type, const unsigned int face_no)
+      {
+        return get_cell(get_cell(type).face_reference_cell_type(face_no));
+      }
```
peterrum, line 462: @drwells How do you like this way? I have introduced two utility functions.

drwells: :+1:

PR #10647, opened 2020-07-02
Fix default mapping parameter in VectorTools boundary functions

masterleinad:
Fixes https://cdash.43-1.org/testDetails.php?test=49754466&build=8201.


masterleinad:
This change makes the previous one at least backward-compatible since the non-hp versions are implemented in terms of the `hp` version.
I would prefer to have a more thorough look through the library separately if we indeed want to remove default mappings and provide an incompatibility change log entry then.

PR #10651, opened 2020-07-03
Bugfix: Unify dof indices on ghost interfaces if no finite element dominates.

### LINKED ISSUE #10650: Bug: hp DoF unification fails on ghost interfaces if elements do not dominate each other
Consider the case of two neighboring `(Q1xQ2)`  and `(Q2xQ1)` elements on different processes. None of them dominates the other. 

During DoF distribution, the current algorithm designates the first active finite element on the current object as the dominating finite element. However, this fails in parallel, and I don't know why.

I have a failing test at hand in this [branch](https://github.com/marcfehling/dealii/tree/unification-without-domination) and I'm working on a fix.

marcfehling:
By design, there will be always one dominating element for vertices.

The `merge_dof_identity()` functions should do their work similar to the `compute_dof_identity()` functions. However for vertices, they did not. Guess I introduced that bug somewhen in summer 2018.

Lines 1110-1122 (new) are taken from `compute_vertex_dof_identities()`, which makes the conditional in line 1116 (old) obsolete. The other changes are caused by the indentation change.

Fixes #10650.

### DIFF
--- a/source/dofs/dof_handler_policy.cc
+++ b/source/dofs/dof_handler_policy.cc
@@ -1102,98 +1102,99 @@ namespace internal
 
                     // find out which is the most dominating finite
                     // element of the ones that are used on this vertex
-                    const unsigned int most_dominating_fe_index =
+                    unsigned int most_dominating_fe_index =
                       dof_handler.get_fe_collection().find_dominating_fe(
                         fe_indices,
                         /*codim=*/dim);
 
-                    // if we found the most dominating element, then use
-                    // this to eliminate some of the degrees of freedom
-                    // by identification. otherwise, the code that
-                    // computes hanging node constraints will have to
-                    // deal with it by computing appropriate constraints
-                    // along this face/edge
-                    if (most_dominating_fe_index !=
+                    // if we haven't found a dominating finite element,
+                    // choose the very first one to be dominant similar
+                    // to compute_vertex_dof_identities()
+                    if (most_dominating_fe_index ==
                         numbers::invalid_unsigned_int)
-                      {
-                        // loop over the indices of all the finite
-                        // elements that are not dominating, and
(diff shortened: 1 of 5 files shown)


source/dofs/dof_handler_policy.cc
```diff
@@ -1102,98 +1102,99 @@ namespace internal
...
-                        for (const auto &other_fe_index : fe_indices)
-                          if (other_fe_index != most_dominating_fe_index)
+                      most_dominating_fe_index =
+                        dealii::internal::DoFAccessorImplementation::
+                          Implementation::nth_active_fe_index(
+                            dof_handler,
+                            0,
+                            vertex_index,
+                            0,
+                            std::integral_constant<int, 0>());
```
peterrum, line 1122: I am a bit confused. Shouldn't it be like this?
```cpp
                    if (most_dominating_fe_index ==
                        numbers::invalid_unsigned_int)
                      most_dominating_fe_index =
                        dealii::internal::DoFAccessorImplementation::
                          Implementation::nth_active_fe_index(
                            dof_handler,
                            0,
                            vertex_index,
                            0,
                            std::integral_constant<int, 0>());
```

As far as I see, if a valid `most_dominating_fe_index` is found by the `dof_handler.get_fe_collection().find_dominating_fe()` function call, nothing happens!?

marcfehling: If a valid dominating one has been found, dofs are identified to the dominating one. If none were found, they will be identified to the first active finite element, that this conditional deals with, which lacks any curly brackets.
https://github.com/dealii/dealii/blob/904369c78ae5e461d9168d8f2f6435c54a27f5f8/source/dofs/dof_handler_policy.cc#L1113-L1122

PR #10653, opened 2020-07-05
Remove unused tests

dangars:
These tests have never been used. I realized that they have never been executed because the default parameters of `project_boundary_values_curl_conforming_l2 ` have changed (#10328) and these tests use the old interface, therefore these tests should fail. I took a look to CDash and I believe that they have never been executed.

They have never been executed because they require the complex version of MUMPS.

I do not think that these tests are necessary. I wrote the tests because I misunderstood the meaning of the polynomial degree for the Nedelec elements, see #8720 and #8709

These tests are also quite expensive to run, they are full 3D simulations. Now I'm more careful and I write tiny inexpensive tests :)


drwells:
We have about 20 other tests for this function so I think its fine to remove these more expensive variants.

drwells:
>  these tests use the old interface, therefore these tests should fail.

It looks to me like the first singularity test does not require complex mumps - it should still be run if mumps is installed. Let me see if this test still works for me.

drwells:
It looks like none of the CI machines have mumps installed.

These tests take a really long time to run: they both time out on my desktop - they need to be modified but I would prefer to keep them since they are the only tests we have for the hp variant of the boundary value curl-conforming projection. Would you be willing to keep these tests but modify them so that they run in under a minute in debug mode? We don't necessarily need to show convergence - just calculating a small error that we know is right would suffice.

dangars:
I can modify the tests to make them more simple and not use MUMPS.

I didn't know that I was using the `hp` variant. Where do I use `hp`? Is it because the default parameter of `project_boundary_values_curl_conforming_l2` is a hp mapping?

I didn't use `hp` in purpose, but I'm happy to write a test for the `hp` variant of `project_boundary_values_curl_conforming_l2`

drwells:
The function signature for this function makes the hp version the default:
```cpp  template <int dim, typename number>
  void
  project_boundary_values_curl_conforming_l2(
    const DoFHandler<dim, dim> &           dof_handler,
    const unsigned int                     first_vector_component,
    const Function<dim, number> &          boundary_function,
    const types::boundary_id               boundary_component,
    AffineConstraints<number> &            constraints,
    const hp::MappingCollection<dim, dim> &mapping_collection);
```
which is a result of #10328: i.e., since the mapping collection argument is last and optional this is now the default version of the function.

dangars:
Thanks!
I'm going to modify the tests

tests/numerics/project_bv_curl_conf_05.cc
```diff
@@ -0,0 +1,113 @@
...
+#include <deal.II/fe/fe_nedelec.h>
+#include <deal.II/fe/fe_nedelec_sz.h>
+
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/tria.h>
+
+#include <deal.II/lac/affine_constraints.h>
+
+#include <deal.II/numerics/vector_tools.h>
+#include <deal.II/numerics/vector_tools_boundary.templates.h>
```
dangars, line 31: The hp variant of `project_boundary_values_curl_conforming_l2()` has not been instantiated, therefore I have to add this header.

include/deal.II/numerics/vector_tools_boundary.h
```diff
@@ -687,6 +687,9 @@ namespace VectorTools
    * hp-namespace version of project_boundary_values_curl_conforming_l2
    * (above).
    *
+   * This function has not been instantiated, you may need the templates header
+   * to use it.
```
dangars, line 691: I don't think that it is worthy to instantiate this function, not many people are going to use it. I would just add a comment. What do you think?

dangars:
I rebased to the last version of deal.II. What do you think about the test that I added for the hp variant of `project_boundary_values_curl_conforming_l2()`?

drwells:
Sorry about the long delay in reviewing this.

The tests look fine but missing the instantiations was a bug in the library that has since been fixed: see #10667.

tests/numerics/project_bv_curl_conf_05.cc
drwells, line 31:
```suggestion
#include <deal.II/numerics/vector_tools_boundary.h>
```

include/deal.II/numerics/vector_tools_boundary.h
drwells: We can get rid of this now that #10667 is in.

dangars:
I updated the the PR and took into account your comments :)

PR #10672, opened 2020-07-08
Fixed mesh_loop in one dimension.

luca-heltai:
I remember discussing this a couple of years ago with @tjhei. I really thought we had fixed it, but the test I'm posting here shows otherwise.

### DIFF
--- a/doc/news/changes/minor/20200708LucaHeltai
+++ b/doc/news/changes/minor/20200708LucaHeltai
@@ -1,0 +1,3 @@
+Fixed: MeshWorker::mesh_loop() did not work on 1d with refined grids. This is now fixed.
+<br>
+(Luca Heltai, 2020/07/08)
--- a/include/deal.II/meshworker/mesh_loop.h
+++ b/include/deal.II/meshworker/mesh_loop.h
@@ -435,6 +435,42 @@ namespace MeshWorker
                                     copy);
                       }
                   }
+                else if (CellIteratorType::AccessorType::dimension == 1 &&
+                         cell->level() > neighbor->level())
+                  {
+                    // In one dimension, there is no other check to do
+                    const unsigned int neighbor_face_no =
+                      periodic_neighbor ?
+                        cell->periodic_neighbor_face_no(face_no) :
+                        cell->neighbor_face_no(face_no);
+                    Assert(periodic_neighbor ||
+                             neighbor->face(neighbor_face_no) ==
+                               cell->face(face_no),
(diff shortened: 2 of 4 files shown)


tjhei:
Why is it that the original logic using ``cell->neighbor_is_coarser`` doesn't work? Would this be worth fixing? If not, we should only use it for dim!=1.

I think it might make sense to reshuffle the logic to be
```
if (dim==1 && neighbor is coarser)
{
// your new code
}
else if (neighbor is coarser)
{
old code
}
```

luca-heltai:
The various calls that come afterwards (that try to get information about subfaces) can only be called in dim>1, and the call that checks if the neighbor is coarser returns always false in 1d. The scope of those calls is different in two and three dimensions, since they were meant to deal with anisotropy. I Ciao amore mio reshuffle the order if you prefer.

tjhei:
> the call that checks if the neighbor is coarser returns always false in 1

That is clearly a bug then. When we fix/change that, mesh_loop will be broken, so:

> I Can reshuffle the order if you prefer.

I would think it is useful to make it obvious that the original code is never run for dim==1. Maybe add a ``dim>1 && ..`` in there? Then you can leave the order as is.

tjhei:
by the way, several tests failed: https://jenkins.tjhei.info/blue/organizations/jenkins/dealii/detail/PR-10672/2/tests
```

/jenkins/workspace/dealii/PR-10672/include/deal.II/meshworker/mesh_loop.h:438:68: error: no type named ‘AccessorType’ in ‘class dealii::IteratorOverIterators<dealii::TriaActiveIterator<dealii::DoFCellAccessor<2, 2, false> > >’
                 else if (CellIteratorType::AccessorType::dimension == 1 &&
                                                                    ^
```

tjhei:
can you use ``cell->get_triangulation()::dimension``?

luca-heltai:
> > the call that checks if the neighbor is coarser returns always false in 1
> 
> That is clearly a bug then. When we fix/change that, mesh_loop will be broken

I agree. However this would require a much deeper change. For example, at the moment we cannot call  `neighbor_of_coarser_neighbor` or `periodic_neighbor_of_coarser_periodic_neighbor` in dimension one.
 
> I would think it is useful to make it obvious that the original code is never run for dim==1. Maybe add a `dim>1 && ..` in there? Then you can leave the order as is.

done.

tjhei:
awesome, thank you!

PR #10673, opened 2020-07-08
Use ReferenceCell::Type to get Info

peterrum:
Now that PR #10637 (`ReferenceCell::Type`) and PR #10641 (`ReferenceCell::internal::Info`) are merged, the type can be used to dynamically select the right info.


peterrum:
I have rebased this branch.

PR #10674, opened 2020-07-08
Replace FiniteElementData::dofs_per_* with FiniteElementData::n_dofs_per_*

peterrum:
... and introduce `FiniteElementData::get_first_*`.

They key aspect is that internally the methods are used (and not the fields) such that asserts can be introduced depending on the reference cell type.


luca-heltai:
I like this change. If no-one objects, I'll merge tonight.

PR #10680, opened 2020-07-09
Add check to step-12

peterrum:
I had problems with collection methods.


tjhei:
Interesting. Do you have a quadrature point in the origin?

peterrum:
> Interesting. Do you have a quadrature point in the origin?

More or less: https://github.com/dealii/dealii/blob/b742d2cc8e32e5c9f7b1e58dfba85a868bc2a8b0/source/simplex/quadrature_lib.cc#L114-L119

PR #10687, opened 2020-07-10
Add Simplex::DataOut

peterrum:
A simple output class I am using to visualize the result obtained on a simplex mesh. This class is not very sophisticated but can be used until the regular `DataOut` class has been generalized.

FYI @DanielPaukner


peterrum:
Since I already have a generalized version of `DataOut` on a local branch (see https://github.com/peterrum/dealii/pull/18/commits/2e287318ad368bacf3f8d5053d3efc973d03bd33) this is not needed anymore. I will open a new PR for the generalized version of `DataOut` later today once I have cleaned it up.

PR #10696, opened 2020-07-11
Remove some if-statements

peterrum:


include/deal.II/fe/fe_tools.templates.h
```diff
@@ -247,73 +247,70 @@ namespace FETools
         }
 
       // 2. Lines
-      if (GeometryInfo<dim>::lines_per_cell > 0)
```
peterrum, line 250: These if-statements. Most of the changes lines are only due to the changed indentation.

PR #10704, opened 2020-07-14
Fix cast in communicate_active_fe_indices()

peterrum:


drwells:
Is this the right choice for `dealii::parallel::shared::Triangulation`?

peterrum:
> Is this the right choice for dealii::parallel::shared::Triangulation?

You are right. I meant `parallel::DistributedTriangulationBase`. Technically, it was like this because the the p:s:t check was before that!

drwells:
Oh, I see - the original version would have worked. This change makes things a bit clearer though so lets stick with it.

PR #10705, opened 2020-07-15
Add new hp::FEFaceValues::reinit() functions

peterrum:
...which accepts also face iterators. Just like in the case of `FEFaceValues`.


tests/hp/fe_face_values_reinit_01.cc
```diff
@@ -0,0 +1,131 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2009 - 2020 by the deal.II authors
```
marcfehling, line 3:
```suggestion
// Copyright (C) 2020 by the deal.II authors
```

peterrum: Done!

PR #10707, opened 2020-07-15
Add new hp::MappingCollection constructor

peterrum:
... just like in the case of `hp::FECollection` and `hp::QCollection`.


peterrum:
@masterleinad Done!

PR #10711, opened 2020-07-16
Avoid the use of raw pointers in WorkStream.

bangerth:
We've tried to not use raw pointers throughout the library, but WorkStream had
remaining places. This is not necessary -- it just requires moving a couple of
lines of code around.

/rebuild


tjhei:
> I am a little surprised that these structs are not in an `internal` or `detailed` namespace.

They are in an ``internal`` namespace.

PR #10714, opened 2020-07-17
Add missing CRS entry and some asserts

peterrum:
Missed in #10328. Currently, it is not noticeable that the last entry is missing due to the way the CRS used at the moment, however, I intend to make a change where I need this last entry.


Rombur:
I am confused how did the tests passed before? You change the reference files but this PR only adds asserts.

source/dofs/dof_handler.cc
```diff
@@ -1649,6 +1653,17 @@ namespace internal
                             numbers::invalid_dof_index);
                       }
                 }
+
+            dof_handler.object_dof_ptr[l][d].push_back(
+              dof_handler.object_dof_indices[l][d].size());
```
peterrum, line 1658: @Rombur These lines have been add in addition to the asserts. It was not noticed that the last entry of the CRS pointer was missing since currently only the pointer to the beginning of the *rows* is used.

PR #10720, opened 2020-07-18
Pass ReferenceCell::Type to QProjection

peterrum:
... and move the implementation of `QProjection` to a new file `qprojection.cc`


include/deal.II/fe/fe_base.h
```diff
@@ -219,6 +219,13 @@ class FiniteElementData
    */
   static const unsigned int dimension = dim;
 
+private:
+  /**
+   * Reference cell type.
+   */
+  const ReferenceCell::Type cell_type;
+
+public:
```
peterrum, line 228: From #10717.

peterrum:
This PR does not change any functionality. It only adds an additional parameter (reference cell type) to all methods of `QProjector`. In a follow-up PR (on which I am working), I will use this parameter to project quadrature rules onto faces of triangles and tetrahedrons.

include/deal.II/grid/reference_cell.h
```diff
@@ -45,6 +45,52 @@ namespace ReferenceCell
...
+          return Type::Line;
+        case 2:
+          return Type::Quad;
+        case 3:
+          return Type::Hex;
+        default:
+          Assert(false, ExcNotImplemented());
+          return Type::Invalid;
+      }
+  }
```
peterrum, line 92: To be able to program dimension independently I have introduced the following two functions.

include/deal.II/base/qprojector.h
```diff
@@ -87,19 +89,40 @@ class QProjector
    * is used on face <tt>face_no</tt>. For further details, see the general
    * doc for this class.
    */
-  static void
+  DEAL_II_DEPRECATED static void
```
kronbichler, line 92: Can you add a statement in the function documentation suggesting that the function of the same name with a `ReferenceCell::Type` argument should be used?

include/deal.II/base/qprojector.h
```diff
@@ -87,19 +89,40 @@ class QProjector
...
+                  std::vector<Point<dim>> & q_points);
+
   /**
    * Compute the cell quadrature formula corresponding to using
    * <tt>quadrature</tt> on face <tt>face_no</tt>. For further details, see
    * the general doc for this class.
    */
-  static Quadrature<dim>
+  DEAL_II_DEPRECATED static Quadrature<dim>
   project_to_face(const SubQuadrature &quadrature, const unsigned int face_no);
```
kronbichler, line 114: Same here and in all occurrences below.

include/deal.II/base/qprojector.h
peterrum: Done!

PR #10727, opened 2020-07-20
GridIn::read_msh() now supports triangular and tetrahedral elements

DanielPaukner:
This PR introduces the support to read triangular and tetrahedral elements from Gmsh files through the GridIn::read_msh() function. Note that so far, only Gmsh files are supported but the plan is to add the same functionality to other mesh files such as .ucd and .unv.


source/grid/grid_in.cc
```diff
@@ -1926,14 +1980,30 @@ GridIn<dim, spacedim>::read_msh(std::istream &in)
                       vertex = numbers::invalid_unsigned_int;
                     }
               }
-            else if ((cell_type == 3) && (dim == 3))
+            else if ((cell_type == 2 || cell_type == 3) && (dim == 3))
```
luca-heltai, line 1983: I know it is not your fault, but shouldn't this be triggered also for `cell_type == 1`, i.e., edges at the boundary?

luca-heltai: (This may be addressed in another PR though).

source/grid/grid_in.cc
```diff
@@ -1731,6 +1731,8 @@ GridIn<dim, spacedim>::read_msh(std::istream &in)
   std::vector<CellData<dim>>                 cells;
   SubCellData                                subcelldata;
   std::map<unsigned int, types::boundary_id> boundary_ids_1d;
+  bool                                       is_hex_mesh = false;
+  bool                                       is_tet_mesh = false;
```
luca-heltai, line 1735: Would you mind renaming `is_hex_mesh` to `is_quad_or_hex_mesh` and `is_tet_mesh` with `is_tria_or_tet_mesh`?

luca-heltai:
Thanks for your effort! Except the minor comment on the name of the two boolean variables, this is good to go.

tests/simplex/grid_in_msh.cc
```diff
@@ -0,0 +1,73 @@
...
+#include <deal.II/grid/grid_out.h>
+#include <deal.II/grid/tria.h>
+
+#include <fstream>
+
+#include "../tests.h"
+
+template <int dim, int spacedim>
+void
+check_file(const std::string &file_name, const unsigned int vertices_per_cell)
```
peterrum, line 30: Could you remove the unused parameter. Thanks.

PR #10729, opened 2020-07-20
Fix a warning about virtual destructors and virtual functions.

drwells:


masterleinad:
Looks like the indentation is off. :slightly_smiling_face:

drwells:
That's odd - the indentation check is looking at an older revision that has an extra blank line. Let me try force pushing again.

PR #10736, opened 2020-07-21
Avoid hang by AssertThrow

dangars:
If I write the following code, I get the expected behaviour. The program terminates and I get the stacktrace
```
Assert(Utilities::MPI::this_mpi_process(mpi_communicator) == 0, ExcInternalError());
```
On the other hand, If I write the following code, the program hangs and does not provide any message.
```
AssertThrow(Utilities::MPI::this_mpi_process(mpi_communicator) == 0, ExcInternalError());
```
This behaviour makes some bugs difficult to debug. I wrote a solution in this PR, but I'm happy to hear your suggestions.

This behaviour might be related to my machine, therefore it might be useful if other developers check that the code above hangs deal.II programs.

### DIFF
--- a/include/deal.II/base/exceptions.h
+++ b/include/deal.II/base/exceptions.h
@@ -1514,7 +1514,8 @@ namespace deal_II_exceptions
     {                                                                  \
       if (__builtin_expect(!(cond), false))                            \
         ::dealii::deal_II_exceptions::internals::issue_error_noreturn( \
-          ::dealii::deal_II_exceptions::internals::throw_on_exception, \
+          ::dealii::deal_II_exceptions::internals::                    \
+            abort_or_throw_on_exception,                               \
           __FILE__,                                                    \
           __LINE__,                                                    \
           __PRETTY_FUNCTION__,                                         \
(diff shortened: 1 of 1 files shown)


include/deal.II/base/exceptions.h
```diff
@@ -1512,9 +1512,10 @@ namespace deal_II_exceptions
 #ifdef DEAL_II_HAVE_BUILTIN_EXPECT
 #  define AssertThrow(cond, exc)                                       \
     {                                                                  \
-      if (__builtin_expect(!(cond), false))                            \
+      if (!(cond))                                                     \
```
masterleinad, line 1515: Why did you change this?

include/deal.II/base/exceptions.h
```diff
@@ -1512,9 +1512,10 @@ namespace deal_II_exceptions
...
         ::dealii::deal_II_exceptions::internals::issue_error_noreturn( \
-          ::dealii::deal_II_exceptions::internals::throw_on_exception, \
+          ::dealii::deal_II_exceptions::internals::                    \
+            abort_or_throw_on_exception,                               \
```
masterleinad, line 1518: This turns `AssertThrow` into the exact same thing as `Assert` if I see correctly (except that it is always enabled in `Release` mode). I don't think that we want this. Blocking.

include/deal.II/base/exceptions.h
dangars: This is not necessary. I copied the text from Assert. I removed it.

masterleinad:
Just to be clear, some of our control flow relies on throwing exceptions instead of aborting, so this pull request will not work. It might be interesting to understand why `AssertThrow` is hanging for you, though.

dangars:
Should I open an issue instead? I run into this problem in #10737. It took me very long time to debug because it only happens in parallel and there was no error message.

dangars:
I close this PR, I'll write a test that isolates the problem.

PR #10741, opened 2020-07-22
Generalize process_dof_indices()

peterrum:
This PR generalizes `process_dof_indices()`, s.t. it does not process dof by dof but processes all dofs of an object (referencing the comments in https://github.com/dealii/dealii/pull/10268#discussion_r428651722). The number of dofs are determined with the help of the CRS tables in `DoFHandler`.

The code could be much simpler if the multigrid internal data structures of `DoFHandler` would be updated.

### DIFF
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -223,19 +223,24 @@ namespace internal
 
 
       /**
-       * Set the @p local_index-th degree of freedom corresponding to the
+       * Process the @p local_index-th degree of freedom corresponding to the
        * finite element specified by @p fe_index on the vertex with global
        * number @p vertex_index to @p global_index.
        */
-      template <int dim, int spacedim, int d>
+      template <int dim,
+                int spacedim,
+                int d,
+                typename GlobalIndexType,
+                typename DoFPProcessor>
       static void
-      set_dof_index(const DoFHandler<dim, spacedim> &dof_handler,
-                    const unsigned int               obj_level,
-                    const unsigned int               obj_index,
-                    const unsigned int               fe_index,
-                    const unsigned int               local_index,
-                    const std::integral_constant<int, d> &,
-                    const types::global_dof_index global_index)
+      process_dof_index(const DoFHandler<dim, spacedim> &dof_handler,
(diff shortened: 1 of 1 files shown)


peterrum:
@luca-heltai From your point of view, the function should have hardly changed:

https://github.com/dealii/dealii/blob/e88d8cac1c940b47b5d35228247314712bbe2d13/include/deal.II/dofs/dof_accessor.templates.h#L912-L987

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -312,111 +321,224 @@ namespace internal
...
-                    const unsigned int               obj_index,
-                    const unsigned int               fe_index,
-                    const unsigned int               local_index,
-                    const std::integral_constant<int, d> &)
+      DEAL_II_ALWAYS_INLINE static std::pair<unsigned int, unsigned int>
+      process_object_range(const DoFHandler<dim, spacedim> &dof_handler,
+                           const unsigned int               obj_level,
+                           const unsigned int               obj_index,
+                           const unsigned int               fe_index,
+                           const std::integral_constant<int, d> &)
```
kronbichler, line 345: Can you explain why you need `always_inline` here? Did the compiler make bad decisions and are these functions performance-critical? I would have guess that at least for the simple cases (without the `find` or `distance`), the compiler should be able to apply early return and inline those parts, whereas the more expensive ones do not matter.

peterrum: I didn't have a look at the compiler output. My intention was that the early returns are inlined (because the return values there can be quite simply determined without an extra function call). But I can remove the `DEAL_II_ALWAYS_INLINE`.

kronbichler: I guess that a reasonable optimizing compiler should inline the early returns; without measuring I would not use the `DEAL_II_ALWAYS_INLINE`. I say this as someone who has used it pretty excessively (after measuring).

peterrum: OK. I'll remove it. Are there anything else I should modify?

peterrum: @kronbichler  I have removed them!

PR #10743, opened 2020-07-22
Add -Wmissing-braces to linux test

peterrum:
These types of warnings, we currently only see on `CDash`. The tests will hopefully fail; this should be fixed once https://github.com/dealii/dealii/pull/10742 is merged.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -20,7 +20,7 @@ jobs:
         cmake --version
     - name: configure
       run: |
-        cmake -D CMAKE_BUILD_TYPE=Release -D DEAL_II_CXX_FLAGS='-Werror' .
+        cmake -D CMAKE_BUILD_TYPE=Release -D DEAL_II_CXX_FLAGS='-Werror -Wmissing-braces' .
     - name: archive
       uses: actions/upload-artifact@v1
       with:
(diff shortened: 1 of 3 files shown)


tjhei:
Interesting, we should consider turning this on by default. @tamiko ?

tjhei:
It seems to fail as expected:
```
/home/runner/work/dealii/dealii/include/deal.II/grid/reference_cell.h:339:37: error: missing braces around initializer for ‘std::__array_traits<unsigned int, 2>::_Type {aka unsigned int [2]}’ [-Werror=missing-braces]
             {{0, 0}, {0, 1}, {1, 1}}};
```
Can you rebase to  #10742? (Or I guess we can wait until the tests pass)

peterrum:
Done!

masterleinad:
> Interesting, we should consider turning this on by default. @tamiko ?

Yes, I would just add it since we care.

include/deal.II/grid/connectivity.h
```diff
@@ -166,7 +166,7 @@ namespace internal
       unsigned int
       n_entities(const unsigned int d) const override
       {
-        static std::array<unsigned int, 3> table = {2, 1};
+        static std::array<unsigned int, 3> table = {{2, 1}};
```
masterleinad, line 169: Unrelated to this pull request, I find omitting the last zero here (and in similar places) rather confusing.

peterrum: This was a mistake. The vector should actually have the length 2. I have fixed it in #10742.

Issue #10737, opened 2020-07-21
HDF5 and step-62 are not compatible with find_active_cell_around_point()


dangars:
Step-62 hangs. This related to two issues.
- find_active_cell_around_point() fails in parallel. Even if a point is inside the domain some of the MPI ranks will fail. I think that it is related to #10478 but I'm not sure.
- The `AssertThrow` of find_active_cell_around_point() fails and hangs the program. This is related to #10736

I'm going to prepare a small test to isolate the problem.

dangars:
@masterleinad I have been able to isolate the problem with `AssertThrow` in step-62

It is related to two issues.
- In parallel `find_active_cell_around_point()` fails to find some points in the coarse cells. See test #10747
- `Assert` and `AssertThrow` hang with HDF5. See test #10746

dangars:
The new version of `find_active_cell_around_point()` uses exceptions and that is not compatible with HDF5. See #11050 and #10535

Exceptions combined with parallel HDF5 can freeze deal.II. If one MPI process throws an exception and not the other ones, then deal.II freezes. The reason is that the following happens:
1. On the processor that calls the exception we try to close the parallel MPI file (H5F_try_close) which ultimately results in a barrier (MPI_barrier). This is triggered by the destructor of an HDF5Object.
2. On the processor that does not call the exception we call H5VL_dataset_write() and end up in MPI_Allreduce().

This could happen if an error is found in all the processes except one. For example, the following code could potentially freeze deal.II:
```
HDF5::File data_file(filename,
                          HDF5::File::FileAccessMode::create,
                           mpi_communicator);
{
  auto  dataset_a =
         data_file.create_dataset<double>(dataset_name_a, dataset_dimensions_a);
   dataset_a.write_selection(data_a, coordinates_a);
 }

 AssertThrow(Utilities::MPI::this_mpi_process(mpi_communicator) == 0,
                   ExcInternalError());

{
   auto  dataset_b =
         data_file.create_dataset<double>(dataset_name_b, dataset_dimensions_b);
   dataset_b.write_selection(data_b, coordinates_b);
 }
```

masterleinad:
Isn't this only a problem if the exceptions are not caught and cause the program to abort?

dangars:
I haven't tried to catch the exception of `find_active_cell_around_point()`. I think that using exceptions to communicate that the point is not in an active cell is not very efficient, for this reason I prefer to use an old version of deal.II.

[...]

PR #10746, opened 2020-07-23
Add test for parallel hdf5 and AssertThrow

dangars:
`Assert` and `AssertThrow` hang with parallel HDF5. This test shows the behaviour.

This is related to #10737

### DIFF
--- a/tests/base/hdf5_07_freezes.cc
+++ b/tests/base/hdf5_07_freezes.cc
@@ -1,0 +1,198 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2015 - 2019 by the deal.II Authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE at
+// the top level of the deal.II distribution.
+//
+// ---------------------------------------------------------------------
+
+// Test AssertThrow with hdf5
+
+#include <deal.II/base/conditional_ostream.h>
+#include <deal.II/base/hdf5.h>
+
+#include <deal.II/lac/full_matrix.h>
+#include <deal.II/lac/vector.h>
+
+#include <boost/type_traits.hpp>
+
+#include <numeric>
+
+#include "../tests.h"
+
+
+// This function tests parallel write and gets the group by reference
+template <template <class...> class Container, typename Number>
+void
+write_test(HDF5::Group &              root_group,
+           const std::vector<hsize_t> dataset_dimensions,
+           MPI_Comm                   mpi_communicator,
+           ConditionalOStream         pcout)
+{
+  AssertDimension(dataset_dimensions.size(), 2);
+
+  std::string container_name;
+  std::string type_name;
+
+  if (std::is_same<Container<Number>, std::vector<Number>>::value)
+    {
+      container_name = std::string("std::vector");
+    }
+  else if (std::is_same<Container<Number>, FullMatrix<Number>>::value)
+    {
+      container_name = std::string("FullMatrix");
+    }
+
+  if (std::is_same<Number, double>::value)
+    {
+      type_name = std::string("double");
+    }
+
+  deallog << "Write tests for " << container_name << "<" << type_name << ">"
+          << " datasets" << std::endl;
+
+  auto group = root_group.create_group(container_name + "<" + type_name + ">");
+
+  {
+    std::string dataset_name("dataset_2");
+    auto        dataset =
+      group.create_dataset<Number>(dataset_name, dataset_dimensions);
+    dataset.set_query_io_mode(true);
+    deallog << "Dimensions " + dataset_name << " " << container_name << "<"
+            << type_name << ">"
+            << " (Write): " << dataset.get_dimensions() << std::endl;
+    deallog << "Size " + dataset_name << " " << container_name << "<"
+            << type_name << ">"
+            << " (Write): " << dataset.get_size() << std::endl;
+    deallog << "Rank " + dataset_name << " " << container_name << "<"
+            << type_name << ">"
+            << " (Write): " << dataset.get_rank() << std::endl;
+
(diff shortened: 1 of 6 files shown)


tests/base/hdf5_07.cc
```diff
@@ -0,0 +1,556 @@
...
+                           HDF5::File::FileAccessMode::create,
+                           mpi_communicator);
+
+      write_test<std::vector, double>(data_file,
+                                      dataset_dimensions,
+                                      mpi_communicator,
+                                      pcout);
+
+      AssertThrow(Utilities::MPI::this_mpi_process(mpi_communicator) == 0,
+                  ExcInternalError());
```
dangars, line 523: If this is replaced by Assert, the program will hang as well.

bangerth: This code is equivalent to the following:
```
{
  const bool b = (Utilities::MPI::this_mpi_process(mpi_communicator) == 0);
  AssertThrow (b, ExcInternalError());
}
```
Can you find out whether it also hangs if you replace `AssertThrow` by `Assert` here? In other words, whether the problem is in the call to `Utilities::MPI::this_mpi_process` or in the `AssertThrow`?

Separately, do you actually need all of the code above? Or could `write_test()` be drastically reduced to just one piece?

dangars: I don't have access to a computer this week.

I'll give you a thorough answer next week :)

tests/base/hdf5_07.cc
```diff
@@ -0,0 +1,198 @@
...
+
+      write_test<FullMatrix, double>(data_file,
+                                     dataset_dimensions,
+                                     mpi_communicator,
+                                     pcout);
```
dangars, line 170: Interestingly the test hangs with the following code
```
write_test<FullMatrix, double>(data_file, dataset_dimensions, mpi_communicator, pcout);
```
But the test does not hang with the following code
```
write_test<std::vector, double>(data_file, dataset_dimensions, mpi_communicator, pcout);
```

dangars:
How does `Assert` and `AssertThrow` work in parallel/MPI?

Let's say that only one process does not fulfill an `Assert` condition. Does this process notify the rest of the processes that they have to raise an exception?

drwells:
Good question. If there is an uncaught exception and you initialized things with `MPI_InitFinalize` then we detect it in that classes destructor:
```cpp
          if (std::uncaught_exceptions() > 0)
            {
              // do not try to call MPI_Finalize to avoid a deadlock.
            }
          else
            {
              const int ierr = MPI_Finalize();
              (void)ierr;
              AssertNothrow(ierr == MPI_SUCCESS, dealii::ExcMPI(ierr));
            }
```
At that point `mpirun` knows that a process exited abnormally and it can deal with it (usually by killing the other processes).

On the other hand if we have an assertion failure instead of calling `abort` we call `MPI_Abort`.

dangars:
Thanks @drwells !
Your explanation will help me to find the origin of the problem

dangars:
I debug the program and I think that the problem is an interaction between exceptions and HDF5.

To debug the program I run the test using 3 MPI processes and attached each MPI process to an independent gdb.

I debug two versions of the test. The first version is the test that you can find in this PR.

In the second version I removed the HDF5 code.

**First version**
When I arrive to the lines below. Process 0 will continue normally and processes 1 and 2 execute `ExcInternalError()`

```
AssertThrow(Utilities::MPI::this_mpi_process(mpi_communicator) == 0,
ExcInternalError());
```

Then processes 1 and 2 arrive to the line 1322 of exceptions.h
```
case throw_on_exception:
  throw e;
```
When processes 1 and 2 throw the exception, they freeze instead of exiting the process. Therefore the whole program freezes.

**Second version**
Now I remove the HDF5 code and leave the AssertThrow.

Process 0 continues normally and processes 1 and 2 arrive to the line 1322 of exceptions.h
```
case throw_on_exception:
  throw e;
```
As expected, after the exception is thrown by processes 1 and 2, all the processes exit and the stacktrace is shown.


**Solution**
I see several solutions
- We can give a CMake compile option in which instead throwing an exception deal.II aborts the program and generates a core dump. A core dump can then be used to debug the program and get the stacktrace.
- We try to dig more in order to understand the problem better, although I'm not sure which direction to take.
- We document  the behavior in the HDF5 documentation of deal.II. Although this does not help to debug a program.

drwells:
If I have understood this correctly this means that exceptions behave differently in HDF5 code than in non-HDF5 code. I think its worth understanding what's going on here before we move on to possible fixes. I'll take a look at this in depth at some point over the next few days.

dangars:
Thanks @drwells !

Yes, exceptions interact in a weird way with parallel HDF5. I think that the problem is with HDF5, not with deal.II. I will contact the HDF5 developers once understand the origin of the problem.

I think that the problem arises when a HDF5 handler is opened and one process throws an exception and the other processes do not throw the exception.

I added a couple of tests. You can do a diff to see the differences between the tests.

If you debug the tests, it is helpful to add a breakpoint at the line 1322 of exceptions.h (function `issue_error_noreturn()`)

**hdf5_07**
- A parallel HDF5 handler is opened
- An exception is thrown **by two processes of the three processes**
- The program freezes

**hdf5_08**
- An exception is thrown by two processes of the three processes
- The program exits as expected

**hdf5_09**
- A parallel HDF5 handler is opened
- An exception is thrown **by all the processes**
- The program exits as expected

drwells:
Sorry for the long delay - I finally looked into this and I have an answer.

If we do MPI IO through HDF5 then the following happens:

1. On the processor that calls the exception we try to close the parallel MPI file (`H5F_try_close`) which ultimately results in a barrier (`PMPI_barrier`).
2. On the processor that does not call the exception we call `H5VL_dataset_write` and end up in `PMPI_allreduce`.

Hence I think that exceptions can't work here - we have to abort.

dangars:
@drwells Thanks a lot for the exhaustive analysis! That's very impressive! It is definitively not easy to debug MPI programs.

I can document this in the HDF5 interface of deal.II.

drwells:
You're welcome. The trick is to let things hang in the debugger and then hit ^C to get stack traces - its a good way to debug hangs like this.

drwells:
@dangars what do you want to do with this PR?

dangars:
I think that we should close the PR. You found the problem!

I updated the documentation in #11050

PR #10478, opened 2020-06-07
Equip GridTools::find_active_cell_around_point() with tolerance

nfehn:
The tolerance has been hard-coded at some places, but should generally be a parameter set by the user. Otherwise, different modules using different hard-coded tolerances lead to conflicts.

Issue #10535, opened 2020-06-15
Improve design of GridTools::find_active_cell_around_point()


nfehn:
This issue discusses design changes of `GridTools::find_active_cell_around_point()`, but not bug fixes:

Let's collect the opinions mentioned so far in other chats:

@nfehn: throwing exceptions

> I realized that most of these tests do not contain a try-block around the function call. The question is, however, whether this behavior should be changed anyway. In my opinion it is not a good design that the function throws if no point is found (because the try-blocks are not used in practice). I would prefer a return type of std::vector, and an empty vector indicates that no point has been found, as do the find_all_...()-type functions introduced later.

@kronbichler: Default parameters

> One thing that is a bit annoying is the fact that the function that already had the tolerance, find_all_active_cells_around_point, uses the arguments in a different order to what is done here. I guess there is little we can do there, but in general we should try to get the interface in an intuitive way. Once we have merged #10477, we should also include that function. That said, I would like to have many opinions on how we want to go ahead with default parameters. 

@tamiko: Backward compatibility

> I just want to add one point to the discussion:

> If possible, we should try to maintain a backward compatible behavior as much as possible. Simply because changing functionality (such as returning a different type, throwing/not throwing an exception) will inevitably break user code.

> As a good metric, we have used our testsuite for precisely that purpose in the past: our tests are a good indicator what semantics the original author had in mind, or at least, what another developer thought is an intended use case. So I feel a bit uneasy about the implied statement that the tests in question are faulty - breaking those might imply that we break a number of (formerly valid) user programs in subtle ways.

@luca-heltai: throwing exceptions

> We should also provide a function that does not throw when points are not found. Suggestions?

kronbichler:
Let me comment on this one:
> I realized that most of these tests do not contain a try-block around the function call. The question is, however, whether this behavior should be changed anyway. In my opinion it is not a good design that the function throws if no point is found (because the try-blocks are not used in practice). I would prefer a return type of std::vector, and an empty vector indicates that no point has been found, as do the find_all_...()-type functions introduced later.

I think the most intuitive C++ way is to return `MeshType::end()` in case we do not find a point, rather than throw an exception. We have seen that things get messy with lots of try/catch blocks. Returning the invalid `end()` iterator would be in line with the typical C++ `find` functions.

[...]

PR #10747, opened 2020-07-23
Add parallel test for find_active_cell_around_point

dangars:
The following test succeeds in serial (mpi = 1), but fails in parallel (mpi = 7). I think that it is related to #10478, but I'm not sure.

I get the following error with `find_active_cell_around_point()`:
```
The point <0 0> could not be found inside any of the subcells of a coarse grid cell.
```
But the point <0 0> is in the coarse grid.

This is a follow-up from #10737


peterrum:
The function `find_active_cell_around_point` only succeeds on locally owned and ghost cells (i.e. active cells). My guess is that the problematic point is in neither type of cells on one of the processes. The fact that it part of the coarse grid is not relevant. @dangars Could you check this?

FYI @nfehn

dangars:
@peterrum That is right. The point is not in any of the active cells. But I do not understand why this worked before. In step-62 I used this function to find out if a point is in the locally owned cells.

According to the documentation, this function will raise an exception only if the point does not lie inside any of the cells of the mesh.
```
If the point requested does not lie in any of the cells of the mesh given,
then this function throws an exception of type GridTools::ExcPointNotFound.
```

I'm going to try to find out why the old version didn't raise an exception when the point does not lie in an active cell.

@peterrum Is `find_active_cell_around_point()` the right function to find out if a point is in a locally owned cell?

peterrum:
We have changed recently the behavior of the function(s) because it turned out that working internally with non-active cells was not a good idea.

There is still a discussion how to deal with this changed behavior (see https://github.com/dealii/dealii/issues/10535). The documentation will be adjusted once we found a common ground.

> Is find_active_cell_around_point() the right function to find out if a point is in a locally owned cell?

You can do that. But what you need to do is to catch the exception to be on the safe side.

dangars:
> > Is find_active_cell_around_point() the right function to find out if a point is in a locally owned cell?
> 
> You can do that. But what you need to do is to catch the exception to be on the safe side.

Thanks @peterrum, I understand now :) I think that exceptions are quite inefficient for this. I will leave step-62 in a broken state until it has been decided what to do with this function. Then I will fix it.

bangerth:
What's the status here?

dangars:
I'll update this once #11276 is in master :)

peterrum:
@dangars Could you rebase this PR!? Thanks!

dangars:
I rebased the test. I didn't have the time update the test. I'll try to find some time tomorrow to do it.

bangerth:
@dangars Today would be the day to get it in before the release :-)

dangars:
Hi @bangerth, I didn't have time to finish the test, I guess that this will make it in deal.II 9.4

I pushed my last version of the test, something is not working. It shouldn't be difficult to find the problem, but I need a little bit more time.

step-62 shows that `find_active_cell_around_point()` works well in parallel with MPI.

I'll let you know when it is ready for review!

FYI @luca-heltai

marcfehling:
> I pushed my last version of the test, something is not working. It shouldn't be difficult to find the problem, but I need a little bit more time.

Can you be more specific?

In the meantime, I will check your version against the current master branch. If it runs smoothly, I will merge the test as is.

marcfehling:
I get errors of the following type:
```
37: /dealii/tests/bits/point_value_03.cc: In instantiation of ‘void check() [with int dim = 2]’:
2437: /dealii/tests/bits/point_value_03.cc:208:12:   required from here
2437: /dealii/tests/bits/point_value_03.cc:160:34: error: ‘dealii::GridTools::Cache<2, 2> cache’ has incomplete type
2437:    GridTools::Cache<dim, dim>     cache(tria, mapping);
2437:                                   ^~~~~
```

marcfehling:
You missed to include a header. I will push directly to your branch.

The serial version of the tests passes, but the parallel run with 7 processes still fails:
```
2439: --------------------------------------------------------
2439: An error occurred in line <108> of file <dealii/include/deal.II/numerics/vector_tools_point_value.templates.h> in function
2439:     void dealii::VectorTools::point_value(const dealii::Mapping<dim, spacedim>&, const dealii::DoFHandler<dim, spacedim>&, const VectorType&, const dealii::Point<spacedim>&, dealii::Vector<typename VectorType::value_type>&) [with int dim = 2; VectorType = dealii::PETScWrappers::MPI::Vector; int spacedim = 2; typename VectorType::value_type = double]
2439: The violated condition was: 
2439:     cell_point.first.state() == IteratorState::valid && cell_point.first->is_locally_owned()
2439: Additional information: 
2439:     The given point is inside a cell of a
2439:     parallel::distributed::Triangulation that is not locally owned by this
2439:     processor.
2439: --------------------------------------------------------
```

Can you have a look?

masterleinad:
I pushed a working version.

kronbichler (approved):

PR #10748, opened 2020-07-23
GridIn::read_vtk() for simplex meshes

peterrum:
similar to #10727.

~I'll add tomorrow a test.~


peterrum:
I have added the test :)

luca-heltai:
If no-one objects, I'll merge tonight.

PR #10750, opened 2020-07-24
CMake: Fix static executable build for cmake 3.9 and newer

### LINKED ISSUE #9184: Build of static library possibly broken
According to the documentation (dealii.org/current/users/cmake.html) it should be possible to build a static library by setting DEAL_II_STATIC_EXECUTABLE=ON.

But calling

`cmake -DDEAL_II_STATIC_EXECUTABLE=ON -DDEAL_II_FORCE_BUNDLED_THREADS=ON ..`

gives me the following error:

```
CMake Error at cmake/setup_finalize.cmake:137 (MESSAGE):
  

Configuration error: Cannot compile a test program with the final set of
compiler and linker flags:
  CXX flags (DEBUG): -pedantic -fPIC -Wall -Wextra -Woverloaded-virtual -Wpointer-arith -Wsign-compare -Wsuggest-override -Wswitch -Wsynth -Wwrite-strings -Wno-placement-new -Wno-deprecated-declarations -Wno-literal-suffix -Wno-psabi -fopenmp-simd -std=c++17 -Wno-parentheses -Wno-unused-local-typedefs -O0 -ggdb -Wa,--compress-debug-sections
  LD flags  (DEBUG): -Wl,--as-needed -static -pthread  -pthread -ggdb
  LIBRARIES (DEBUG): /usr/lib/x86_64-linux-gnu/liblapack.a;/usr/lib/x86_64-linux-gnu/libblas.a;gfortran;quadmath;m;c;gcc_s;gcc;dl;rt;
```

As far as I understand, the problem is that LIBRARIES contain gcc_s. This is shared library, but when we link statically the linker looks for a gcc_s.a, which doesn't exist. If I add 
`-Dgcc_s_LIBRARY=""`
in the cmake call the library compiles without any problems. So it seems like gcc_s should be removed from LIBRARIES when DEAL_II_STATIC_EXECUTABLE=ON.

Complete cmake output:
[cmake_output.txt](https://github.com/dealii/dealii/files/3991435/cmake_output.txt)
[CMakeError.log](https://github.com/dealii/dealii/files/3991436/CMakeError.log)
[CMakeOutput.log](https://github.com/dealii/dealii/files/3991437/CMakeOutput.log)

tamiko:
We have to filter the shared low-level runtime library "gcc_s" from the
link interface to work around a change in the CMake configuration of
Fortran.

Closes #9184


cmake/modules/FindLAPACK.cmake
```diff
@@ -82,12 +82,23 @@ ENDIF()
...
+  SET(_fortran_libs ${CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES})
+  #
+  # Since CMake 3.9 the gcc runtime libraries libgcc.a and libgcc_s.so.1
+  # have been added to the CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES variable.
+  # We thus have to remove the shared low-level runtime library
+  # libgcc_s.so.1 from the link interface; otherwise completely static
+  # linkage is broken.
+  #
+  LIST(REMOVE_ITEM _fortran_libs gcc_s)
+  SET_IF_EMPTY(_fortran_libs gfortran quadmath m)
```
masterleinad, line 95: What happened to `c`?

PR #10752, opened 2020-07-25
Improve memory behavior of the BiCGStab solver.

bangerth:
See the updated commentary for more details.

/rebuild


bangerth:
Yes, the design is stupid. The other solvers keep all of the solver loop functionality in one function and so don't need member variables.

But good point about release memory at the end. How about this?

tjhei:
> How about this?

I like it. Also less explaining necessary. :-)

PR #10753, opened 2020-07-27
Fix memory reservation of hp::DoFHandler

peterrum:


peterrum:
I guess it was only a copy-paste error from my side.

PR #10755, opened 2020-07-27
Fix compiling with CUDA 11

masterleinad:
Simply error out if we try to use functionality removed in CUDA 11 for now.


Rombur:
I am planning on fixing this tonight.

PR #10764, opened 2020-07-29
Assign faces different quadrature rules

peterrum:
This PR allows to assign different quadrature rules to each face of a cell. This is needed for wedges and pyramids, where one want to use `FE_Q` on quadrilateral faces and `FE_P` on triangular faces; however the implementation is general so that can also be applied to quads, triangles, tets, and hexes.

This is the first PR in a series of PRs enabling mixed meshes.

This PR required modifications in `QProjector`, `FEValues`, `hp::FEValues`, `FiniteElement`, and `Mapping`.

I still need to add some comments and implement hexes, triangles, and tets. Nevertheless, I would appreciate a first feedback.


include/deal.II/fe/fe_values.h
```diff
@@ -3617,6 +3624,14 @@ class FEValues : public FEValuesBase<dim, spacedim>
            const Quadrature<dim> &             quadrature,
            const UpdateFlags                   update_flags);
 
+  /**
+   * TODO
+   */
+  FEValues(const Mapping<dim, spacedim> &      mapping,
+           const FiniteElement<dim, spacedim> &fe,
+           const hp::QCollection<dim> &        quadrature,
+           const UpdateFlags                   update_flags);
```
peterrum, line 3633: This is a new constructor for non-hp `FEValues` taking a `hp::QCollection` containing a quadrature rule for each face.

include/deal.II/hp/fe_values.h
```diff
@@ -80,6 +80,16 @@ namespace hp
...
 
+    /**
+     * TODO.
+     */
+    FEValuesBase(
+      const MappingCollection<dim, FEValuesType::space_dimension>
+        &mapping_collection,
+      const FECollection<dim, FEValuesType::space_dimension> &fe_collection,
+      const std::vector<QCollection<q_dim>> &                 q_collection,
+      const UpdateFlags                                       update_flags);
```
peterrum, line 91: This is the equivalent in the case of hp. The outer vector groups the quadrature indices; the `QCollection` contains the face quadrature rules of once quadrature index.

peterrum:
Code like this can be simplified:

https://github.com/dealii/dealii/blob/a6e0cd02792b729fee27a88d16bf4870a407530c/include/deal.II/numerics/vector_tools_boundary.templates.h#L2867-L2882

https://github.com/dealii/dealii/blob/a6e0cd02792b729fee27a88d16bf4870a407530c/include/deal.II/numerics/vector_tools_boundary.templates.h#L1109-L1110

tamiko (changes requested):
This pull request is fine with me.

One thing to very quickly discuss is the fact that this pull requests changes the signature of all protected `fill_fe_face_values`. I think this should be noted under incompatibilities.

Also, does the automatic conversion between `Quadrature` and `hp::QCollection` work when supplied as argument? If yes, then we would remain API compatible (to some degree), which would be preferrable.

peterrum:
@tamiko Sorry for the late response. Now that pyramids and wedges are part of master this PR is needed to evaluate surface integrals with `FEFaceValues`.

Let me shortly explain why I think the change of the interface is needed. Within `FEFaceValues` face quadrature rules are projected onto every face and the result is collected within one big quadrature rule. To access the quadrature points of i-th face happens via `QProjector<dim>::DataSetDescriptor::face()`:

https://github.com/dealii/dealii/blob/90882ad117a82cae25e79231f953c686fb0aaee8/include/deal.II/fe/fe_poly.templates.h#L362-L368

which return the correct offset (by multiplying the number of quadrature points per face with i). However this multiplication is not possible in the case of pyramids and wedges, since here each face might have different number of quadrature points. As a result we need to pass in this case either the number of quadrature points of each face to `QProjector<dim>::DataSetDescriptor::face()` or - as I propose here - a quadrature collection (which contains exactly the information of the number of quadrature points of each face).

If you agree, I will fix the merge conflict, rebase, squash the commits, and add an incompatibility log entry.

include/deal.II/fe/fe_values.h
```diff
@@ -2175,9 +2177,24 @@ class FEValuesBase : public Subscriptor
...
   /**
-   * Number of quadrature points.
+   * Number of quadrature points of the current object. Its value is
+   * initialized by the value of max_quadrature_points and is updated,
+   * e.g., if FEFaceValues::reinit() is called for a new cell/face.
+   *
+   * @note The default value equals to the value of max_quadrature_points.
    */
-  const unsigned int n_quadrature_points;
+  unsigned int n_quadrature_points;
```
kronbichler, line 2186: This is a potentially dangerous change, given that it is a public member and non-const. Now I know we won't have a clean solution here, but I would prefer (if possible) to keep this variable `const` and use `const_cast` at the single position this variable can change. What do others think?

include/deal.II/fe/fe_values.h
```diff
@@ -2175,9 +2177,24 @@ class FEValuesBase : public Subscriptor
...
+  /**
+   * Maximum number of quadrature points. This value might be different from
+   * n_quadrature_points, e.g., if a QCollection with different face quadrature
+   * rules has been passed to initialize FEFaceValues.
+   *
+   * This is mostly useful to initialize arrays to allocate the maximum amount
+   * of memory that may be used when re-sizing later on to a the current
+   * number of quadrature points given by n_quadrature_points.
+   */
+  const unsigned int max_quadrature_points;
```
kronbichler, line 2197: I think we more often use the pattern `max_n_quadrature_points` for this kind of names.

include/deal.II/fe/fe_values.h
```diff
@@ -3745,6 +3762,17 @@ class FEValues : public FEValuesBase<dim, spacedim>
            const Quadrature<dim> &             quadrature,
            const UpdateFlags                   update_flags);
 
+  /**
+   * Like the function above, but taking a collection of quadrature rules.
+   *
+   * @note We require, in contrast to FEFaceValues that the number of quadrature
```
kronbichler, line 3768:
```suggestion
   * @note We require, in contrast to FEFaceValues, that the number of quadrature
```

include/deal.II/fe/fe_values.h
```diff
@@ -3754,6 +3782,16 @@ class FEValues : public FEValuesBase<dim, spacedim>
...
```
kronbichler, line 3788:
```suggestion
   * @note We require, in contrast to FEFaceValues, that the number of quadrature
```

include/deal.II/hp/fe_values.h
```diff
@@ -80,6 +80,18 @@ namespace hp
       const QCollection<q_dim> &                              q_collection,
       const UpdateFlags                                       update_flags);
 
+    /**
+     * Like the above function but taking a vector quadrature collections.
```
kronbichler, line 84:
```suggestion
     * Like the above function but taking a vector of quadrature collections.
```

source/fe/fe_values.cc
```diff
@@ -444,10 +444,8 @@ namespace FEValuesViews
         &shape_function_data,
       std::vector<typename ProductType<Number, double>::type> &values)
     {
-      const unsigned int dofs_per_cell = dof_values.size();
-      const unsigned int n_quadrature_points =
-        dofs_per_cell > 0 ? shape_values.n_cols() : values.size();
-      AssertDimension(values.size(), n_quadrature_points);
+      const unsigned int dofs_per_cell       = dof_values.size();
+      const unsigned int n_quadrature_points = values.size();
```
kronbichler, line 448: By the way, this also seems like a good simplification on the way.

kronbichler (commented):
This looks like a very good step forward. I have some minor comments here.

include/deal.II/fe/fe_values.h
peterrum: We should make such variables private^^

peterrum: But that variables are public is an inherent problem in deal.II. See also `FiniteElementData`: `dofs_per_face` vs. `n_dofs_per_face()` with the latter only working for wedges/pyramids.

kronbichler:
A bigger question to me is the API compatibility mentioned by @tamiko : I think this is a serious problem because every user implementing his/her own `FiniteElement` would immediately run into an error. I am thinking whether we should add a compatibility function with the old interface. Unless overridden by a user function, the new function in `FiniteElement` would, for the time being, forward to the old function that a user can provide. One potential problem with this setup would be that neither the old nor the new function would be `pure virtual` with that setup, so a user would then get a run-time error rather than a compile error. Yet, it would define the most gentle way forward I see right now.

include/deal.II/fe/fe_values.h
kronbichler: I agree. The other possibility would be to always return the value 0 and have a getter function for the proper content.

peterrum:
@kronbichler What about (before I start working for an hour):

```cpp
  virtual void
  fill_fe_face_values(
    const unsigned int                                          face_no,
    const Quadrature<dim - 1> &                                 quadrature,
    const Mapping<dim, spacedim> &                              mapping,
    const typename Mapping<dim, spacedim>::InternalDataBase &mapping_internal,
    const dealii::internal::FEValuesImplementation::MappingRelatedData<dim,
                                                                       spacedim>
      &                     mapping_data,
    const InternalDataBase &fe_internal,
    dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim,
                                                                       spacedim>
      &output_data) const = 0;

  virtual void
  fill_fe_face_values_hp(
    const unsigned int                                          face_no,
    const hp::QCollection<dim - 1> &                            quadrature,
    const Mapping<dim, spacedim> &                              mapping,
    const typename Mapping<dim, spacedim>::InternalDataBase &mapping_internal,
    const dealii::internal::FEValuesImplementation::MappingRelatedData<dim,
                                                                       spacedim>
      &                     mapping_data,
    const InternalDataBase &fe_internal,
    dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim,
                                                                       spacedim>
      &output_data) const
  {
    AssertDimension(quadrature.size(), 1);
    fill_fe_face_values(face_no, quadrature[0], mapping, mapping_internal, 
                mapping_data, fe_internal, output_data);
  }
```
with `fill_fe_face_values_hp()` being called by `FEFaceValues` and overridden by certain FEs.

include/deal.II/fe/fe_values.h
peterrum: I have made the change (`const_cast`). But this really hurts and I makes me doubt how the compiler can make any assumptions about variables.

peterrum:
@kronbichler I guess the same discussion is also true for `FiniteElement::get_face_data()`, `Mapping::get_face_data()`, `Mapping::fill_fe_face_values()`?

include/deal.II/fe/fe_values.h
kronbichler: I think it is OK and defined behavior because the `this` pointer is not constant in this context. It would be different if we were messing with a `const` object or even `constexpr` because then the compiler might never create a storage location for that variable.

kronbichler:
The interface for `fill_fe_face_values` depends on what we want to achieve and what our future interface should look like. With [your suggestion](https://github.com/dealii/dealii/pull/10764#issuecomment-744025504) we say that the variable quadrature case is exceptional in the sense that we add another function with a different name. From what I see, this would require a complete implementation of the `fill_fe_face_values_hp` function in derived class for which we want to enable the new flexible setup, whereas we would not touch any of the existing classes derived from `FiniteElement` because they would still go through the old interface. Since we are doing this for pyramid elements and the like only, I think that might be a good strategy because it minimizes the code changes. The downside is that we have two interfaces in the base class, that we fill depending on the type of element, and the two would never go away.

The alternative I had originally been thinking about would be more radical in the sense that it makes the old interface deprecated (in some sense) to make for a smoother transition. So this would look like
```c++
  virtual void
  fill_fe_face_values(
    const unsigned int                                          face_no,
    const Quadrature<dim - 1> &                                 quadrature,
    const Mapping<dim, spacedim> &                              mapping,
    const typename Mapping<dim, spacedim>::InternalDataBase &mapping_internal,
    const dealii::internal::FEValuesImplementation::MappingRelatedData<dim,
                                                                       spacedim>
      &                     mapping_data,
    const InternalDataBase &fe_internal,
    dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim,
                                                                       spacedim>
      &output_data) const
  {
    Assert(false, ExcMessage("Use of a deprecated interface, please implement "
                             "fill_fe_face_values taking a hp::QCollection argument"));
  }

  virtual void
  fill_fe_face_values(
    const unsigned int                                          face_no,
    const hp::QCollection<dim - 1> &                            quadrature,
    const Mapping<dim, spacedim> &                              mapping,
    const typename Mapping<dim, spacedim>::InternalDataBase &mapping_internal,
    const dealii::internal::FEValuesImplementation::MappingRelatedData<dim,
                                                                       spacedim>
      &                     mapping_data,
    const InternalDataBase &fe_internal,
    dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim,
                                                                       spacedim>
      &output_data) const
  {
    // base class version, implement overriden function in derived classes
    AssertDimension(quadrature.size(), 1);
    fill_fe_face_values(face_no, quadrature[0], mapping, mapping_internal, 
                mapping_data, fe_internal, output_data);
  }
```
And then we implement and use only the `QCollection` case in the library like you are doing right now, whereas user codes implementing classes derived from `FiniteElement` might overload the old version with a `Quadrature` object. This makes it slightly easier to forget thinking about an appropriate interface because it is now only a run-time error.

I don't know which approach is better, but I wanted to raise a possible alternative here.

tamiko (approved):

masterleinad:
```
/home/runner/work/dealii/dealii/include/deal.II/fe/fe.h:3346: warning: no uniquely matching class member found for 
  template < dim, spacedim >
  void FiniteElement< dim, spacedim >::fill_fe_face_values(const typename Triangulation< dim, spacedim >::cell_iterator &cell, unsigned int face_no, const hp::QCollection< dim - 1 > &quadrature, const Mapping< dim, spacedim > &mapping, const typename Mapping< dim, spacedim >::FiniteElement::InternalDataBase &mapping_internal, const ::internal::FEValuesImplementation::MappingRelatedData< dim, spacedim > &mapping_data, const typename FiniteElement< dim, spacedim >::FiniteElement::InternalDataBase &fe_internal, ::internal::FEValuesImplementation::FiniteElementRelatedData< dim, spacedim > &output_data) const
Possible candidates:
  virtual void FiniteElement< dim, spacedim >::fill_fe_face_values(const typename Triangulation< dim, spacedim >::cell_iterator &cell, const unsigned int face_no, const hp::QCollection< dim - 1 > &quadrature, const Mapping< dim, spacedim > &mapping, const typename Mapping< dim, spacedim >::InternalDataBase &mapping_internal, const ::internal::FEValuesImplementation::MappingRelatedData< dim, spacedim > &mapping_data, const InternalDataBase &fe_internal, ::internal::FEValuesImplementation::FiniteElementRelatedData< dim, spacedim > &output_data) const' at line 2964 of file /home/runner/work/dealii/dealii/include/deal.II/fe/fe.h
  virtual void FiniteElement< dim, spacedim >::fill_fe_face_values(const typename Triangulation< dim, spacedim >::cell_iterator &cell, const unsigned int face_no, const Quadrature< dim - 1 > &quadrature, const Mapping< dim, spacedim > &mapping, const typename Mapping< dim, spacedim >::InternalDataBase &mapping_internal, const ::internal::FEValuesImplementation::MappingRelatedData< dim, spacedim > &mapping_data, const InternalDataBase &fe_internal, ::internal::FEValuesImplementation::FiniteElementRelatedData< dim, spacedim > &output_data) const' at line 2982 of file /home/runner/work/dealii/dealii/include/deal.II/fe/fe.h

/home/runner/work/dealii/dealii/include/deal.II/fe/fe.h:3376: warning: no uniquely matching class member found for 
  template < dim, spacedim >
  void FiniteElement< dim, spacedim >::fill_fe_face_values(const typename Triangulation< dim, spacedim >::cell_iterator &cell, unsigned int face_no, const Quadrature< dim - 1 > &quadrature, const Mapping< dim, spacedim > &mapping, const typename Mapping< dim, spacedim >::FiniteElement::InternalDataBase &mapping_internal, const ::internal::FEValuesImplementation::MappingRelatedData< dim, spacedim > &mapping_data, const typename FiniteElement< dim, spacedim >::FiniteElement::InternalDataBase &fe_internal, ::internal::FEValuesImplementation::FiniteElementRelatedData< dim, spacedim > &output_data) const
Possible candidates:
  virtual void FiniteElement< dim, spacedim >::fill_fe_face_values(const typename Triangulation< dim, spacedim >::cell_iterator &cell, const unsigned int face_no, const hp::QCollection< dim - 1 > &quadrature, const Mapping< dim, spacedim > &mapping, const typename Mapping< dim, spacedim >::InternalDataBase &mapping_internal, const ::internal::FEValuesImplementation::MappingRelatedData< dim, spacedim > &mapping_data, const InternalDataBase &fe_internal, ::internal::FEValuesImplementation::FiniteElementRelatedData< dim, spacedim > &output_data) const' at line 2964 of file /home/runner/work/dealii/dealii/include/deal.II/fe/fe.h
  virtual void FiniteElement< dim, spacedim >::fill_fe_face_values(const typename Triangulation< dim, spacedim >::cell_iterator &cell, const unsigned int face_no, const Quadrature< dim - 1 > &quadrature, const Mapping< dim, spacedim > &mapping, const typename Mapping< dim, spacedim >::InternalDataBase &mapping_internal, const ::internal::FEValuesImplementation::MappingRelatedData< dim, spacedim > &mapping_data, const InternalDataBase &fe_internal, ::internal::FEValuesImplementation::FiniteElementRelatedData< dim, spacedim > &output_data) const' at line 2982 of file /home/runner/work/dealii/dealii/include/deal.II/fe/fe.h

deprecated:156: warning: explicit link request to 'internal::FEValuesImplementation::MappingRelatedData< dim, spacedim >' could not be resolved
deprecated:156: warning: explicit link request to 'internal::FEValuesImplementation::FiniteElementRelatedData< dim, spacedim >' could not be resolved
deprecated:169: warning: explicit link request to 'internal::FEValuesImplementation::MappingRelatedData< dim, spacedim >' could not be resolved
```

kronbichler:
The new variant looks good to me, if we get the error message reported by @masterleinad fixed. The question is whether we need the backward compatible version. In my opinion yes because otherwise we will make all custom `FiniteElement` implementations overleading this `fill_fe_face_values` function dysfunctional from day one, but I see nobody else confirming this more gentle attitude towards backward compatibility, so I will certainly not insist on it here.

masterleinad:
This should fix the doxygen problems:
```diff
diff --git a/include/deal.II/fe/fe.h b/include/deal.II/fe/fe.h
index ffb353e23a..f9816e8e96 100644
--- a/include/deal.II/fe/fe.h
+++ b/include/deal.II/fe/fe.h
@@ -2996,12 +2996,10 @@ protected:
     const Quadrature<dim - 1> &                                 quadrature,
     const Mapping<dim, spacedim> &                              mapping,
     const typename Mapping<dim, spacedim>::InternalDataBase &mapping_internal,
-    const dealii::internal::FEValuesImplementation::MappingRelatedData<dim,
-                                                                       spacedim>
+    const internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
       &                     mapping_data,
     const InternalDataBase &fe_internal,
-    dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim,
-                                                                       spacedim>
+    internal::FEValuesImplementation::FiniteElementRelatedData<dim, spacedim>
       &output_data) const;
 
   /**
diff --git a/include/deal.II/fe/mapping.h b/include/deal.II/fe/mapping.h
index 4ecc49fa10..8a0bce664c 100644
--- a/include/deal.II/fe/mapping.h
+++ b/include/deal.II/fe/mapping.h
@@ -937,7 +937,7 @@ protected:
     const unsigned int                                          face_no,
     const Quadrature<dim - 1> &                                 quadrature,
     const typename Mapping<dim, spacedim>::InternalDataBase &   internal_data,
-    dealii::internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
+    internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
       &output_data) const;
 
   /**
```

peterrum:
@masterleinad Thanks! This worked!

include/deal.II/fe/fe.h
```diff
@@ -2767,6 +2767,16 @@ class FiniteElement : public Subscriptor, public FiniteElementData<dim>
    * it is no longer necessary.
    */
   virtual std::unique_ptr<InternalDataBase>
+  get_face_data(const UpdateFlags               update_flags,
+                const Mapping<dim, spacedim> &  mapping,
+                const hp::QCollection<dim - 1> &quadrature,
+                dealii::internal::FEValuesImplementation::
+                  FiniteElementRelatedData<dim, spacedim> &output_data) const;
```
peterrum, line 2774: @kronbichler I have adjusted the other functions also that there are two version. However, noticed an issue with `FiniteElement::get_face_data`, since it already has an implementation in the base class so that the version taking `Quadrature` will never be called. Any suggestions!?

doc/news/changes/incompatibilities/20201212Munch
```diff
@@ -0,0 +1,6 @@
+Changed: The interface of FE::fill_fe_face_values() now accepts instead
+of a Quadrature instance a hp::QCollection instance, enabling the evaluation of
+shape functions for different quadrature points at each face, given that the
+FiniteElement supports this feature (e.g. FE_Q, Simplex::FE_P).
```
kronbichler, line 4: I am not 100% sure what you mean here. Is it this:
```suggestion
shape functions for different quadrature formulas on different faces for
FiniteElement classes with supports for this feature (e.g. FE_Q, Simplex::FE_P). 
```

include/deal.II/base/qprojector.h
```diff
@@ -251,8 +253,16 @@ class QProjector
...
    */
   static Quadrature<dim>
-  project_to_all_faces(const ReferenceCell::Type reference_cell_type,
-                       const SubQuadrature &     quadrature);
+  project_to_all_faces(const ReferenceCell::Type       reference_cell_type,
+                       const hp::QCollection<dim - 1> &quadrature);
+
+  /**
+   * Like the above function but taking only a single face quadrature
+   * formula.
```
kronbichler, line 261:
```suggestion
   * Like the above function, applying the same face quadrature
   * formula on all faces.
```

include/deal.II/base/qprojector.h
```diff
@@ -444,6 +454,17 @@ class QProjector
          const bool                face_rotation,
          const unsigned int        n_quadrature_points);
 
+    /**
+     * TODO.
```
kronbichler, line 458: Do you still expect changes here in terms of interface that motivates to postpone the documentation? If not, I suggest to fill it now because it gets forgotten otherwise. It's mostly copy-paste from the previous function anyway.

include/deal.II/fe/fe.h
```diff
@@ -3323,6 +3350,69 @@ FiniteElement<dim, spacedim>::get_associated_geometry_primitive(
...
+                                                                     spacedim>
+    &                                                            mapping_data,
+  const typename FiniteElement<dim, spacedim>::InternalDataBase &fe_internal,
+  dealii::internal::FEValuesImplementation::FiniteElementRelatedData<dim,
+                                                                     spacedim>
+    &output_data) const
+{
+  // base class version, implement overriden function in derived classes
+  AssertDimension(quadrature.size(), 1);
+  fill_fe_face_values(cell,
```
kronbichler, line 3372: Even though this function has no big body, shouldn't we move it to the `.cc` file anyway because it calls a heavy function where the function call overhead is not decisive compared to a slightly leaner header file? Same for the next one.

include/deal.II/fe/mapping.h
```diff
@@ -1257,6 +1278,69 @@ class Mapping : public Subscriptor
...
+void
+Mapping<dim, spacedim>::fill_fe_face_values(
+  const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+  const unsigned int                                          face_no,
+  const hp::QCollection<dim - 1> &                            quadrature,
+  const typename Mapping<dim, spacedim>::InternalDataBase &   internal_data,
+  dealii::internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
+    &output_data) const
+{
+  // base class version, implement overriden function in derived classes
```
kronbichler, line 1292: I would move this function and the ones below to the `.cc` file as well.

source/fe/fe_enriched.cc
```diff
@@ -584,7 +586,7 @@ FE_Enriched<dim, spacedim>::fill_fe_face_values(
 
   if (is_enriched)
     multiply_by_enrichment(
-      quadrature, fe_data, mapping_data, cell, output_data);
+      quadrature[0 /*TODO*/], fe_data, mapping_data, cell, output_data);
```
kronbichler, line 589: Since we use the index `0` here, can you add the `Assert` like in the other functions?

source/fe/fe_values.cc
```diff
@@ -2191,9 +2162,9 @@ namespace FEValuesViews
     Assert(fe_values->update_flags & update_hessians,
            (typename FEValuesBase<dim, spacedim>::ExcAccessToUninitializedField(
              "update_hessians")));
-    Assert(laplacians.size() == fe_values->n_quadrature_points,
+    Assert(laplacians.size() == fe_values->n_quadrature_points /*TODO ?*/,
```
kronbichler, line 2165: Any plans on this todo and the ones below it? I assume this is because we resize the arrays to `max_n_quadrature_points`?

kronbichler (approved):
I would be happy with this state. I collected a few minor remarks as you will see below, none of which is critical. I think it is good with a deprecation mark in the documentation. After a release or two when user codes have had the opportunity to catch up, we can then remove the old function, slimming down the interface again.

include/deal.II/fe/fe_dgp_nonparametric.h
```diff
@@ -534,11 +534,13 @@ class FE_DGPNonparametric : public FiniteElement<dim, spacedim>
                                                                        spacedim>
       &output_data) const override;
 
+  using FiniteElement<dim, spacedim>::fill_fe_face_values;
```
kronbichler, line 537: Just out of curiosity because I always forget the C++ rules on that topic: Does the code give a compile error without this using statement, or is it just that derived classes could not see the other interface any more to override it? In some sense, I would not feel too unhappy if none of the classes exposes the old interface anymore except the base class, as long as a user can *derive* from it.

peterrum: The problem is that the are two virtual functions with the same name in the base class. If you only override one of them you get a warning. This is the "beauty" of C++...

source/fe/fe_values.cc
peterrum: I also think it was related to `max_n_quadrature_points` but I don't think I had any issues till now. So I remove these TODOs.

peterrum:
@kronbichler Thanks for the review. I have made the changes. Ready to go!

kronbichler (approved):
I think this is ready. Given that we have two approvals, I suggest to wait for another day if no further comments arise, and then merge this PR.

PR #7762, opened 2019-02-27
WIP: Non local dofs discussion

luca-heltai:
@davydden

PR #10766, opened 2020-07-30
Non local dofs (inside FE)

luca-heltai:
This is in constrast with #7762. I would like to discuss the interface here. 

In this PR, non local dofs are handled through the FiniteElement. In #7762 there is an additional class that takes care of this. 

What do people think?

@zhaweiliu @peterrum


include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -1682,11 +1701,51 @@ DoFAccessor<structdim, dim, spacedim, level_dof_access>::get_dof_indices(
...
+               this->dof_handler->get_fe(fe_index).n_dofs_per_quad() +
+             this->dof_handler->get_fe(fe_index).n_dofs_per_hex()) +
+              this->dof_handler->get_fe(fe_index).n_non_local_dofs_per_cell(),
+          ExcVectorDoesNotMatch());
+        break;
+      default:
+        Assert(false, ExcNotImplemented());
+    }
+
+
```
luca-heltai, line 1743: I'm not sure this is needed anymore, after @peterrum  changes.

luca-heltai:
No longer working in the current interface.

PR #10769, opened 2020-07-30
Fix conflict with Simplex::GridGenerator

### LINKED ISSUE #10765: doxygen is confused by simplex namespace
subdivided_hyper_rectangle now also contains the documentation for Simplex::GridGenerator::subdivided_hyper_rectangle at the bottom:
https://www.dealii.org/developer/doxygen/deal.II/namespaceGridGenerator.html#ac76417d7404b75cf53c732f456e6e971
(grep for "tetrahedral")
I am not sure why that happens. Maybe we can change use a different namespace for simplex stuff?

Also, shouldn't simplex/grid_generator.h also be guarded by the simplex define?

peterrum:
closes #10765.

### DIFF
--- a/include/deal.II/simplex/grid_generator.h
+++ b/include/deal.II/simplex/grid_generator.h
@@ -27,223 +27,57 @@
 
 DEAL_II_NAMESPACE_OPEN
 
-namespace Simplex
+#ifdef DEAL_II_WITH_SIMPLEX_SUPPORT
+
+namespace GridGenerator
 {
   /**
-   * This namespace provides a collection of functions to generate simplex
-   * triangulations for some basic geometries.
+   * Create a coordinate-parallel brick from the two diagonally opposite
+   * corner points @p p1 and @p p2. The number of vertices in coordinate
+   * direction @p i is given by <tt>repetitions[i]+1</tt>.
    *
-   *  @ingroup simplex
+   * @note This function connects internally 4/8 vertices to
+   *   quadrilateral/hexahedral cells and subdivides these into 2/5
+   * triangular/tetrahedral cells.
+   *
+   * @note Currently, this function only works for `dim==spacedim`.
+   *
+   * @ingroup simplex
    */
-  namespace GridGenerator
-  {
-    /**
-     * Create a coordinate-parallel brick from the two diagonally opposite
-     * corner points @p p1 and @p p2. The number of vertices in coordinate
-     * direction @p i is given by <tt>repetitions[i]+1</tt>.
-     *
-     * @note This function connects internally 4/8 vertices to quadrilateral/
-     *   hexahedral cells and subdivides these into 2/5 triangular/
-     *   tetrahedral cells.
-     *
-     * @note Currently, this function only works for `dim==spacedim`.
-     */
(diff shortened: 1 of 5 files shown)


include/deal.II/simplex/grid_generator.h
```diff
@@ -27,221 +27,217 @@
 
 DEAL_II_NAMESPACE_OPEN
 
-namespace Simplex
+namespace GridGenerator
```
tjhei, line 30: should there be a ``#ifdef DEAL_II_WITH_SIMPLEX_SUPPORT`` here?

tjhei:
I prefer this over the old namespacing. :+1:

include/deal.II/simplex/grid_generator.h
peterrum: Do you like this solution?

tjhei:
I am not sure I understand. I thought that we don't want simplex related functions to show up in doxygen right now. As a consequence, you will need to ``#ifdef`` based on simplex support. Or am I missing something?

peterrum:
I don't really mind if it appears in doxygen as long as there are enough asserts which prevent the usage.

bangerth:
But things that we don't intend to enable yet shouldn't show up in the documentation either. That would just be confusing.

include/deal.II/simplex/grid_generator.h
```diff
@@ -27,221 +27,225 @@
...
-   * This namespace provides a collection of functions to generate simplex
-   * triangulations for some basic geometries.
+   * Create a coordinate-parallel brick from the two diagonally opposite
+   * corner points @p p1 and @p p2. The number of vertices in coordinate
+   * direction @p i is given by <tt>repetitions[i]+1</tt>.
    *
-   *  @ingroup simplex
+   * @note This function connects internally 4/8 vertices to quadrilateral/
+   *   hexahedral cells and subdivides these into 2/5 triangular/
+   *   tetrahedral cells.
```
bangerth, line 39: This is going to create no space before but one space after the `/` when rendered. Just put the first word of lines 38 and 39 at the end of the previous line and then let indentation break where necessary.

include/deal.II/simplex/grid_generator.h
```diff
@@ -27,221 +27,225 @@
...
-  namespace GridGenerator
+  template <int dim, int spacedim>
+  void
+  subdivided_hyper_rectangle_with_simplices(
+    Triangulation<dim, spacedim> &   tria,
+    const std::vector<unsigned int> &repetitions,
+    const Point<dim> &               p1,
+    const Point<dim> &               p2,
+    const bool                       colorize = false)
   {
```
bangerth, line 53: What's the reason to have the implementation of this function in the .h file, by the way?

peterrum:
I have moved the implementation in a separate file, added the ifdefs, and fixed the doxygen comment. Thanks for the comments!

drwells:
The new name is verbose but clear and avoids doxygen problems.

drwells:
@tjhei @bangerth are you two happy with this change? If so lets merge!

PR #10776, opened 2020-07-31
Refactor DoFHandlerImplementation::reserve_space for 3D

peterrum:


source/dofs/dof_handler.cc
```diff
@@ -410,17 +412,53 @@ namespace internal
...
+            {
+              dof_handler.object_dof_ptr[0][2].assign(
+                dof_handler.tria->n_raw_quads() + 1, -1);
+              // determine for each face the number of dofs
+              for (const auto &cell : dof_handler.tria->cell_iterators())
+                for (const auto face_index : cell->face_indices())
+                  {
+                    const auto &face = cell->face(face_index);
+                    const auto  n_dofs_per_quad =
+                      dof_handler.get_fe().n_dofs_per_quad(/*face_index*/);
```
peterrum, line 424: Goal is to make this number variable.

PR #10777, opened 2020-07-31
Introduce FiniteElementData::max_dofs_per_quad() 

peterrum:
... and FiniteElementData::max_dofs_per_face().

Currently, the returned value is identical to `n_dofs_per_quad()`  and `n_dofs_per_face()`. However, I intend to introduce to the functions `n_dofs_per_quad()`  and `n_dofs_per_face()` an input argument `quad_no`/`face_no` to be query the number of each quad/face separately (since quads/faces might have different number of dofs e.g in the case of triangle/quadrilateral faces).

### DIFF
--- a/include/deal.II/hp/fe_collection.h
+++ b/include/deal.II/hp/fe_collection.h
@@ -884,8 +884,7 @@ namespace hp
 
     unsigned int max = 0;
     for (unsigned int i = 0; i < finite_elements.size(); ++i)
-      if (finite_elements[i]->n_dofs_per_vertex() > max)
-        max = finite_elements[i]->n_dofs_per_vertex();
+      max = std::max(max, finite_elements[i]->n_dofs_per_vertex());
 
     return max;
   }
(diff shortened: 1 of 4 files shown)


include/deal.II/hp/fe_collection.h
```diff
@@ -948,8 +948,8 @@ namespace hp
 
     unsigned int max = 0;
     for (unsigned int i = 0; i < finite_elements.size(); ++i)
-      if (finite_elements[i]->n_dofs_per_face() > max)
-        max = finite_elements[i]->n_dofs_per_face();
+      if (finite_elements[i]->max_dofs_per_face() > max)
+        max = finite_elements[i]->max_dofs_per_face();
```
drwells, line 952:
```suggestion
      max = std::max(max, finite_elements[i]->max_dofs_per_face());
```

include/deal.II/hp/fe_collection.h
```diff
@@ -916,8 +916,8 @@ namespace hp
 
     unsigned int max = 0;
     for (unsigned int i = 0; i < finite_elements.size(); ++i)
-      if (finite_elements[i]->n_dofs_per_quad() > max)
-        max = finite_elements[i]->n_dofs_per_quad();
+      if (finite_elements[i]->max_dofs_per_quad() > max)
+        max = finite_elements[i]->max_dofs_per_quad();
```
drwells, line 920: same

include/deal.II/hp/fe_collection.h
peterrum: I have applied your suggestion to all the other functions in the file too.

I also saw these `if(val > max) max = val;` and didn't like it either. I wanted to be smart and use map-reduce, but it turned out that std algorithms don't have that... That would be one line of code :(

PR #10778, opened 2020-07-31
Generalize FiniteElementData s.t. faces can different nr. of dofs

peterrum:
This PR generalizes the notion of **dofs per objects**/**dpo** so that faces in 3D (objects) can have different number of dofs. This is in particular useful for wedges and pyramids. These are geometric entities whose faces consists of triangles and quadrilaterals. In the case of a quadratic elements, triangles have 6 dofs (1 per vertex and line, 0 inside) and quadrilaterals have 9 dofs (1 per vertex and line, and 1 inside).

Elements needing this kind of flexibility can fill a `PrecomputedFiniteElementData` and pass this to the constructor of `FiniteElementData`. As a consequence, all methods related to quad/face get an additional parameter: `quad_no`/`face_no`. Follow up PRs will explicitly pass a number to these functions.

PR #10779, opened 2020-07-31
Add face_no to methods of FiniteElements

peterrum:
Similar to #10778, this PR adds the parameter `face_no` to face-related methods of `FiniteElement`.

These parameters will be used in follow-up PRs.

### DIFF
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -968,6 +968,7 @@ namespace internal
               [&](const auto d) {
                 return fe.adjust_quad_dof_index_for_face_orientation(
                   d,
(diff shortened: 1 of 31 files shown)


drwells:
I would prefer if we added a commented-out `face_no` to the definitions where we don't use it so that its clear what the argument should be (we do that with some of the other arguments). Are you willing to do this? Otherwise :+1:

peterrum:
I will use `face_no` in most places in a follow-up PR. So adding `/*face_no*/` and then replacing it in most cases by `face_no`  does not really help, does it. I would suggest, I add in the follow-up PR the `/*face_no*/`s.

drwells:
Ah. Lets go ahead with the current version.

PR #10780, opened 2020-07-31
Convert all face-related data structures of FiniteElement to vectors

peterrum:
... related to #10779. To be able to assign each `face_no` different internal data structures.


peterrum:
All tests have passed locally.

PR #10782, opened 2020-08-01
Fix step-18.

adamqc:
Step-18 fails when using more than 1 MPI processes. The error messages are
```
An error occurred in line <95> of file </.../source/distributed/shared_tria.cc> in function
    void dealii::parallel::shared::Triangulation<3, 3>::partition() [dim = 3, spacedim = 3]
The violated condition was: 
    max_active_cells == this->n_active_cells()
Additional information: 
    A parallel::shared::Triangulation needs to be refined in the same way on all processors, but the participating processors don't agree on the number of active cells.
```

This is due to that the refinement/coarsening flags are not the same on all participating processors. The bad commit is [739debc](https://github.com/dealii/dealii/commit/739debc31e70d202b4a6ffe3ac4f335fe03c2d44), so I just revert it.

@peterrum

### DIFF
--- a/source/grid/grid_refinement.cc
+++ b/source/grid/grid_refinement.cc
@@ -16,6 +16,8 @@
 
 #include <deal.II/base/template_constraints.h>
 
+#include <deal.II/distributed/tria_base.h>
+
 #include <deal.II/grid/grid_refinement.h>
 #include <deal.II/grid/tria.h>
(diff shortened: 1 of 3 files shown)


adamqc:
I have added a test case, please have a look.

adamqc:
Rebased to master to fix the test failure.

PR #10784, opened 2020-08-01
Use face_no in FE

peterrum:
~~Only the last commit is relevant.~~

### DIFF
--- a/include/deal.II/fe/fe.h
+++ b/include/deal.II/fe/fe.h
@@ -3129,9 +3129,12 @@ template <int dim, int spacedim>
 inline std::pair<unsigned int, unsigned int>
 FiniteElement<dim, spacedim>::face_system_to_component_index(
   const unsigned int index,
-  const unsigned int) const
+  const unsigned int face_no) const
 {
-  AssertIndexRange(index, face_system_to_component_table[0].size());
+  AssertIndexRange(
+    index,
+    face_system_to_component_table[this->n_unique_faces() == 1 ? 0 : face_no]
+      .size());
 
   // in debug mode, check whether the
   // function is primitive, since
@@ -3146,11 +3149,13 @@ FiniteElement<dim, spacedim>::face_system_to_component_index(
   //
   // in 1d, the face index is equal
   // to the cell index
-  Assert(is_primitive(this->face_to_cell_index(index, 0)),
+  Assert(is_primitive(this->face_to_cell_index(index, face_no)),
          (typename FiniteElement<dim, spacedim>::ExcShapeFunctionNotPrimitive(
            index)));
 
-  return face_system_to_component_table[0][index];
+  return face_system_to_component_table[this->n_unique_faces() == 1 ?
+                                          0 :
+                                          face_no][index];
 }
 
 
@@ -3170,10 +3175,14 @@ template <int dim, int spacedim>
 inline std::pair<std::pair<unsigned int, unsigned int>, unsigned int>
 FiniteElement<dim, spacedim>::face_system_to_base_index(
   const unsigned int index,
-  const unsigned int) const
+  const unsigned int face_no) const
 {
-  AssertIndexRange(index, face_system_to_base_table[0].size());
-  return face_system_to_base_table[0][index];
+  AssertIndexRange(
+    index,
+    face_system_to_base_table[this->n_unique_faces() == 1 ? 0 : face_no]
+      .size());
+  return face_system_to_base_table[this->n_unique_faces() == 1 ? 0 : face_no]
+                                  [index];
 }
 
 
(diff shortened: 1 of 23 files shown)


source/fe/fe_trace.cc
```diff
@@ -53,7 +53,10 @@ FE_TraceQ<dim, spacedim>::FE_TraceQ(const unsigned int degree)
     FETools::hierarchic_to_lexicographic_numbering<dim - 1>(degree));
 
   // Initialize face support points
-  this->unit_face_support_points[0] = fe_q.get_unit_face_support_points();
+  AssertDimension(this->n_unique_faces(), fe_q.n_unique_faces());
+  for (unsigned int face_no; face_no < this->n_unique_faces(); ++face_no)
```
masterleinad, line 57:
```suggestion
  for (unsigned int face_no = 0; face_no < this->n_unique_faces(); ++face_no)
```

source/fe/fe_nedelec.cc
```diff
@@ -2453,20 +2470,20 @@ void
 FE_Nedelec<dim>::get_face_interpolation_matrix(
   const FiniteElement<dim> &source,
   FullMatrix<double> &      interpolation_matrix,
-  const unsigned int) const
+  const unsigned int        face_no) const
```
masterleinad, line 2473: This is only used in `Debug` mode.

peterrum: Should be fine now!

include/deal.II/fe/fe_base.h
```diff
@@ -592,7 +592,8 @@ class FiniteElementData
    */
   template <int structdim>
   unsigned int
-  n_dofs_per_object(const unsigned int i = 0) const;
+  n_dofs_per_object(
+    const unsigned int i = 0 /*numbers::invalid_unsigned_int*/) const;
```
masterleinad, line 596: What is this about?

include/deal.II/fe/fe_base.h
```diff
@@ -840,6 +841,10 @@ FiniteElementData<dim>::n_dofs_per_object(const unsigned int i) const
       case 1:
         return n_dofs_per_line();
       case 2:
+        // Assert(!(structdim == 2 && dim == 3) || i !=
+        // numbers::invalid_unsigned_int,
+        //         ExcInternalError());
```
masterleinad, line 846: ???

source/fe/fe_system.cc
```diff
@@ -1702,43 +1719,52 @@ FESystem<dim, spacedim>::initialize(
...
+          // degrees of freedom on the face at all, then we don't care whether
+          // it has support points or not. this is, for example, the case for
+          // the stable Stokes element Q(p)^dim \times DGP(p-1).
+          for (unsigned int base_el = 0; base_el < this->n_base_elements();
+               ++base_el)
+            if (!base_element(base_el).has_support_points() &&
+                (base_element(base_el).n_dofs_per_face(face_no) > 0))
+              {
+                this->unit_face_support_points[face_no].resize(0);
+                goto face_loop;
```
masterleinad, line 1741: really?! Let's not do this. Maybe define another boolean instead and `continue` after the inner loop in the outer loop.

peterrum: Done!

include/deal.II/fe/fe_base.h
peterrum: This and the next was for "debugging". I have remove them.

include/deal.II/fe/fe.h
```diff
@@ -1469,7 +1469,7 @@ class FiniteElement : public Subscriptor, public FiniteElementData<dim>
    */
   std::pair<unsigned int, unsigned int>
   face_system_to_component_index(const unsigned int index,
-                                 const unsigned int face_no = 0) const;
+                                 const unsigned int face = 0) const;
```
kronbichler, line 1472: Is this change intentional? It seems to contract the other `face_no` argument names?

include/deal.II/fe/fe.h
```diff
@@ -1481,7 +1481,7 @@ class FiniteElement : public Subscriptor, public FiniteElementData<dim>
    */
   unsigned int
   adjust_quad_dof_index_for_face_orientation(const unsigned int index,
-                                             const unsigned int face_no,
+                                             const unsigned int face,
```
kronbichler, line 1484: Same here.

include/deal.II/fe/fe.h
```diff
@@ -1766,7 +1766,7 @@ class FiniteElement : public Subscriptor, public FiniteElementData<dim>
    */
   std::pair<std::pair<unsigned int, unsigned int>, unsigned int>
   face_system_to_base_index(const unsigned int index,
-                            const unsigned int face_no = 0) const;
+                            const unsigned int face_no) const;
```
kronbichler, line 1769: Do we need to document this change somewhere as this can have user-visible effects?

include/deal.II/fe/fe.h
peterrum: I have reverted this change.

include/deal.II/fe/fe.h
peterrum: This was not intended. I have removed the default value to be able to identify all instances where these functions are used. Reverted!

kronbichler: > This was not intended.

That's why we have code review :wink:

PR #10734, opened 2020-07-21
Check that MPIEXEC_EXECUTABLE is available in the test suite.

drwells:
Fixes #10331.

PR #10785, opened 2020-08-01
Fix MPIEXEC detection again.

drwells:
Fixes problems with #10734.

We should check the value deal.II finally stores - other MPI variables are not necessarily set at this point.


drwells:
We skipped the CI last time and I think that might have caught this - lets be a bit more paranoid now.

/rebuild

tjhei:
> I think that might have caught this

yep, I think so too.

bangerth:
Ah, sorry about that -- I had imagines that this is a fix for a broken system, not something we would see on one of the testers...

PR #10793, opened 2020-08-03
Fix CUDA 11 build

Rombur:
This PR does the following: 
	1) use the new interface for cuSPARSE
	2) fix a problem with array view. CUDA 11 introduces a new type of host memory which made our asserts failed

Supersede #10755  
closes #10749

PR #10799, opened 2020-08-05
Drop cuda 9

Rombur:
Has requested in #10793, officially drop support for CUDA 9.


Rombur:
I have updated the PR to make clear that we also dropped support for cuda 10.0 and cuda 10.1

masterleinad:
Since I didn't hear any objections, let's just merge.

PR #10800, opened 2020-08-06
Simplify some string handling in an error message.

bangerth:
This was triggered by the fact that the original string lacked a space after 'assign'.


peterrum:
All tests have passed. Jenkins has problem to finalize. Merging.

PR #10803, opened 2020-08-06
Fix bug in DataPostprocessorTensor.

bangerth:
Specifically, the class currently states that the components it outputs are scalars,
when the whole intent of the class is of course to output its result as a tensor
field.

### DIFF
--- a/tests/data_out/data_out_postprocessor_tensor_02.output
+++ b/tests/data_out/data_out_postprocessor_tensor_02.output
@@ -1,0 +1,40 @@
+
+DEAL::Cycle 0:
+DEAL::   Number of active cells:       16
+DEAL::   Number of degrees of freedom: 50
+DEAL:cg::Starting value 0.279006
+DEAL:cg::Convergence step 11 value 3.55007e-13
+<?xml version="1.0" ?> 
+<!-- 
+# vtk DataFile Version 3.0
+#This file was generated 
+-->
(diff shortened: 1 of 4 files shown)


tests/data_out/data_out_postprocessor_tensor_02.output
```diff
@@ -0,0 +1,40 @@
...
+    <DataArray type="Float32" Name="strain_xx__strain_xy__strain_yx__strain_yy" NumberOfComponents="9" format="binary">
```
tjhei, line 35: Is this the name that shows up in ParaView. This can't be what you want. Also why is this 9 components?

tjhei:
I would prefer if we don't add a test that produces 300+kb of output. It should be easy to do this on a 2x2 mesh or not?

tests/data_out/data_out_postprocessor_tensor_02.output
bangerth: Right. That's a separate bug and I already have the fix for that ready and waiting for this PR here to be merged :-)

bangerth:
Yes, fair point. I just copied the test from an existing one. But I've now shrunk the output from 350kB to 2.5kB :-)

bangerth:
Will shrink the original one next.

tests/data_out/data_out_postprocessor_tensor_02.output
bangerth: See #10806.

PR #10806, opened 2020-08-07
Fix another bug with DataPostprocessorTensor

bangerth:
This is the promised follow-up to #10803: We spelled out the components of the tensor, which `DataOutBase` then helpfully collates into a very long string. What we should be doing is name all components the same because they're part of one whole. Do so here.

(Side note: This matches what we do in `DataPostprocessorVector`.)


tjhei:
One disadvantage with tensor output is that they are now "strain" with component "0, ...., 8" in ParaView, but this is certainly better than the old naming.

PR #10809, opened 2020-08-09
ECL: Merge gather and adjust_for_face_orientation

peterrum:
This PR enables, within `FEFaceEvaluation::gather_evaluate()`, re-orientation during reading for certain cases: no subfaces and no gradients in the case of Hermite basis.

In the next step I would like to make `face_no`, `face_orientation`, and `side` to `std::array<unsigned int, N>` to allow faces to have different numbers/orientations.

The fall-back case (`read_dof_values`/`evaluate`) is still not working.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -2355,7 +2355,9 @@ namespace internal
     {
       (void)subface_index;
 
-      if (integrate && (face_orientation > 0))
+      if (integrate &&
+          (face_orientation > 0 &&
+           subface_index < GeometryInfo<dim>::max_children_per_cell))
         adjust_for_face_orientation(face_orientation,
                                     orientation_map,
                                     true,
@@ -2396,6 +2398,18 @@ namespace internal
       if (integrate)
         function_0(temp1, dofs_per_face);
 
+      const unsigned int  dummy = 0;
+      const unsigned int *orientation =
+        (data.data.front().nodal_at_cell_boundaries == true) ?
+          &data.face_orientations[face_orientation][0] :
+          &dummy;
+      const auto reorientate = [&](const unsigned int i) {
+        return (dim < 3 || face_orientation == 0 ||
+                subface_index < GeometryInfo<dim>::max_children_per_cell) ?
+                 i :
+                 orientation[i];
+      };
+
       // case 1: contiguous and interleaved indices
       if (((do_gradients == false &&
             data.data.front().nodal_at_cell_boundaries == true) ||
@@ -2434,10 +2448,11 @@ namespace internal
                   const unsigned int ind2 = index_array[2 * i + 1];
                   AssertIndexRange(ind1, data.dofs_per_component_on_cell);
                   AssertIndexRange(ind2, data.dofs_per_component_on_cell);
+                  const unsigned int i_ = reorientate(i);
                   for (unsigned int comp = 0; comp < n_components; ++comp)
                     function_1a(
-                      temp1[i + 2 * comp * dofs_per_face],
-                      temp1[i + dofs_per_face + 2 * comp * dofs_per_face],
+                      temp1[i_ + 2 * comp * dofs_per_face],
+                      temp1[i_ + dofs_per_face + 2 * comp * dofs_per_face],
                       global_vector_ptr + dof_index +
                         (ind1 + comp * static_dofs_per_component) *
                           VectorizedArrayType::size(),
@@ -2455,9 +2470,10 @@ namespace internal
                 &data.face_to_cell_index_nodal(face_no, 0);
               for (unsigned int i = 0; i < dofs_per_face; ++i)
                 {
+                  const unsigned int i_  = reorientate(i);
                   const unsigned int ind = index_array[i];
                   for (unsigned int comp = 0; comp < n_components; ++comp)
-                    function_1b(temp1[i + 2 * comp * dofs_per_face],
+                    function_1b(temp1[i_ + 2 * comp * dofs_per_face],
                                 global_vector_ptr + dof_index +
                                   (ind + comp * static_dofs_per_component) *
                                     VectorizedArrayType::size());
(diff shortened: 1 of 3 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2100,6 +2100,10 @@ namespace internal
       else
         temp1 = scratch_data;
 
+      // separate re-orientation is need if it is not performed during
```
kronbichler, line 2103:
```suggestion
      // separate re-orientation is need if it is not performed while
```

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2163,16 +2167,24 @@ namespace internal
...
+                  do_vectorized_read(
+                    src_ptr + dof_index +
+                      (index_array[i] + comp * static_dofs_per_component) *
+                        VectorizedArrayType::size(),
+                    temp1[((dim < 3 || face_orientation == 0 ||
+                            subface_index <
+                              GeometryInfo<dim>::max_children_per_cell) ?
+                             i :
+                             orientation[i]) +
+                          2 * comp * dofs_per_face]);
```
kronbichler, line 2184: This is the hot path of the read operation. Have you checked that the compiler can move the checks out of the `i` loop? I would lean towards duplicating the loop because it isn't probably even that much more code to read given the difficult clause in the access.

include/deal.II/matrix_free/shape_info.h
```diff
@@ -467,6 +467,14 @@ namespace internal
        */
       dealii::Table<2, unsigned int> face_to_cell_index_hermite;
 
+      /**
+       * For degrees on faces, the basis functions are not
+       * in the correct order if a face is not in the standard orientation
+       * to a given element. This data structure is used to re-order the
+       * basis functions to represent the correct order.
+       */
+      dealii::Table<2, unsigned int> face_orientations;
```
kronbichler, line 476: I guess we should remove the array from the mapping info class?

peterrum: It is still used in `integrate_scatter`, `scatter`, and `evalute`

include/deal.II/matrix_free/evaluation_kernels.h
peterrum: I agree that the if-statement is a mess. But duplicating the loop would practically lead to the duplication of the whole function. I think we need a different solution: template parameter?

peterrum:
>I have a few comments regarding the code in addition to the comments below:

> I guess we want to move all code to adjust the face orientation to the place where we read from the vectors, and remove the shuffling of data at the quadrature points. Doing it here should generally be more efficient, at least if the neighbor access is narrow.

Done (partly). The annoying thing is that we read at (at least) two places.

> If we move to the actual ECL loops, faces on different lanes can have different orientations, so the question is whether we should do the same for the FCL infrastructure we already have. The reasonable thing to do is to use the present loops in case all lanes have the same decision, and use three nested loops (nested as lane - dof index - component) in the difficult else case.

Let's see what leads to the least code duplication.

> While the Hermite case is essentially the same as this one just in a few other places, the default case is a bit tough because we need to intersect with the default read operation (as you have observed). Given the large number of places we would need to touch, I wonder if it wouldn't be easier if we do a (possibly per-lane) shuffling at the top of evaluate_in_face or bottom of integrate_in_face, e.g. here:

In the meantime I have implemented also the Hermite case.
  
> This is certainly doing extra work because we do another pass through the data lanes but I wonder if the alternative wouldn't also have downsides, like blowing up the instruction cache and having many more places the branch predictor would have to track. I'm not saying we must go that road, just that we should consider both alternatives.

That would definitely an option. However, this would only work to fix the orientation. But `face_no`  might be also different (and for that this place is too late).

> In any case, if we renumber at the level of polynomials (which I believe is the more sane thing to do), we need to assert for non-nodal bases (think about Legendre) because those cannot be simply switched on the dofs, only on the quad points.

I am not sure, but we are checking in all cases `data.data.front().nodal_at_cell_boundaries == true` so shouldn't we be safe?

kronbichler:
I guess a templated choice could be an option, so we could skip some parts. In any case, let us discuss in person how to implement the indirect access option.

peterrum:
@kronbichler I have rebased. Let's merge even this solution is not perfect. As we discussed in person, there will be some follow-up PRs...

include/deal.II/matrix_free/shape_info.h
kronbichler: I was more thinking about the future when we have everything. Right now it's clear we need it this way.

kronbichler:
I believe this has broken an application test case it seems. I have to see if I can extract a file / test case so we can reproduce and fix it here.

peterrum:
@kronbichler 

@nfehn has given me a failing setup of our application code. The bug is fixed if I insert following lines:

```diff
diff --git a/include/deal.II/matrix_free/evaluation_kernels.h b/include/deal.II/matrix_free/evaluation_kernels.h
index 0cadb08..bc25b47 100644
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -2355,9 +2355,51 @@ namespace internal
     {
       (void)subface_index;
 
+      bool skip = false;
+      if (((do_gradients == false &&
+            data.data.front().nodal_at_cell_boundaries == true) ||
+           (data.element_type ==
+              MatrixFreeFunctions::tensor_symmetric_hermite &&
+            fe_degree > 1)) &&
+          dof_info.index_storage_variants[dof_access_index][cell] ==
+            MatrixFreeFunctions::DoFInfo::IndexStorageVariants::
+              interleaved_contiguous)
+        skip = true;
+      else if (((do_gradients == false &&
+                 data.data.front().nodal_at_cell_boundaries == true) ||
+                (data.element_type ==
+                   MatrixFreeFunctions::tensor_symmetric_hermite &&
+                 fe_degree > 1)) &&
+               dof_info.index_storage_variants[dof_access_index][cell] ==
+                 MatrixFreeFunctions::DoFInfo::IndexStorageVariants::
+                   interleaved_contiguous_strided)
+        skip = true;
+      else if (((do_gradients == false &&
+                 data.data.front().nodal_at_cell_boundaries == true) ||
+                (data.element_type ==
+                   MatrixFreeFunctions::tensor_symmetric_hermite &&
+                 fe_degree > 1)) &&
+               dof_info.index_storage_variants[dof_access_index][cell] ==
+                 MatrixFreeFunctions::DoFInfo::IndexStorageVariants::
+                   interleaved_contiguous_mixed_strides)
+        skip = true;
+      else if (((do_gradients == false &&
+                 data.data.front().nodal_at_cell_boundaries == true) ||
+                (data.element_type ==
+                   MatrixFreeFunctions::tensor_symmetric_hermite &&
+                 fe_degree > 1)) &&
+               dof_info.index_storage_variants[dof_access_index][cell] ==
+                 MatrixFreeFunctions::DoFInfo::IndexStorageVariants::
+                   contiguous &&
+               dof_info.n_vectorization_lanes_filled[dof_access_index][cell] ==
+                 VectorizedArrayType::size())
+        skip = true;
+is
+
       if (integrate &&
           (face_orientation > 0 &&
-           subface_index < GeometryInfo<dim>::max_children_per_cell))
+           (subface_index < GeometryInfo<dim>::max_children_per_cell ||
+            skip == false)))
         adjust_for_face_orientation(face_orientation,
                                     orientation_map,
                                     true,
```

The confusion was the in the case of `gather_evaluate()`, the default case (i.e., `evaluate`) is calling `adjust_for_face_orientation()`; but in the case of `integrate_scatter()`, the function `adjust_for_face_orientation()` is called unconditionally here. This asymmetrical behavior is quite confusing...

Is there way simplify the check? I simply have copied the if-statements from below.

PR #10810, opened 2020-08-09
Work on MatrixFreeFunctions::FaceToCellTopology::face_orientation

peterrum:
This is embarrassing: after two years I have finally understood these lines...


kronbichler:
It's good to have someone else add to the description, so :+1:

PR #10811, opened 2020-08-09
Merge implementation of gather_evaluate and integrate_scatter

peterrum:
Independent of the discussion in PR #10809, it would be good if the loop logic of `gather_evaluate` and `integrate_scatter` would extracted into a new function.


peterrum:
@kronbichler Thanks!

> even though I somewhat dislike the long argument list

I dislike functions/classes which do not fit on a single page, let alone files with multiple thousand of files ;) This is life...

PR #10813, opened 2020-08-11
Fix installation of files of step-49

### LINKED ISSUE #10798: step-49 currently fails
As reported on the mailing list (and I can confirm), step-49 currently doesn't work:
```
> ./step-49.debug 


----------------------------------------------------
Exception on processing: 

An error occurred in line <1425> of file </home/bangerth/p/deal.II/1/dealii/source/grid/grid_in.cc> in function
    void dealii::GridIn<dim, spacedim>::read_msh(std::istream&) [with int dim = 2; int spacedim = 2; std::istream = std::basic_istream<char>]
The violated condition was: 
    in
Additional information: 
    An input/output error has occurred. There are a number of reasons why this may be happening, both for reading and writing operations.

If this happens during an operation that tries to read data: First, you may be trying to read from a file that doesn't exist or that is not readable given its file permissions. Second, deal.II uses this error at times if it tries to read information from a file but where the information in the file does not correspond to the expected format. An example would be a truncated file, or a mesh file that contains not only sections that describe the vertices and cells, but also sections for additional data that deal.II does not understand.

If this happens during an operation that tries to write data: you may be trying to write to a file to which file or directory permissions do not allow you to write. A typical example is where you specify an output file in a directory that does not exist.

Aborting!
----------------------------------------------------
```

peterrum:
closes #10798

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -40,8 +40,8 @@ IF(DEAL_II_COMPONENT_EXAMPLES)
     PATTERN "output.reference.dat"              # step-39
     PATTERN "postprocess.pl"                    # step-39
     PATTERN "obstacle.pbm"                      # step-42
-    PATTERN "untitled.geo"                      # step-49
-    PATTERN "untitled.msh"                      # step-49
+    PATTERN "example.geo"                       # step-49
+    PATTERN "example.msh"                       # step-49
     PATTERN "topography.txt.gz"                 # step-53
     PATTERN "input/initial_mesh_3d.vtk"         # step-54
     PATTERN "input/DTMB-5415_bulbous_bow.iges"  # step-54


examples/CMakeLists.txt
```diff
@@ -40,8 +40,8 @@ IF(DEAL_II_COMPONENT_EXAMPLES)
     PATTERN "output.reference.dat"              # step-39
     PATTERN "postprocess.pl"                    # step-39
     PATTERN "obstacle.pbm"                      # step-42
-    PATTERN "untitled.geo"                      # step-49
-    PATTERN "untitled.msh"                      # step-49
+    PATTERN "example.geo"                      # step-49
+    PATTERN "example.msh"                      # step-49
```
tjhei, line 44: Can you indent the ``#`` please?

peterrum: Done!

PR #10814, opened 2020-08-11
Allow AffineConstraints argument to MGTools::make_sparsity_pattern

kronbichler:
This allows to use `constraints[level].distribute_local_to_global` e.g. on cases with periodic boundary conditions which were not possible before. The arguments taken by `MGTools::make_sparsity_pattern` are similar to what we do for `DoFTools::make_sparsity_pattern` so the users should have an intuitive use for these functions.

### DIFF
--- a/include/deal.II/multigrid/mg_tools.h
+++ b/include/deal.II/multigrid/mg_tools.h
@@ -75,16 +75,25 @@ namespace MGTools
    * Write the sparsity structure of the matrix belonging to the specified @p
    * level. The sparsity pattern is not compressed, so before creating the
    * actual matrix you have to compress the matrix yourself, using
-   * <tt>SparseMatrixStruct::compress()</tt>.
+   * <tt>SparsityPatternType::compress()</tt>.
    *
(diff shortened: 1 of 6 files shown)


include/deal.II/multigrid/mg_tools.h
```diff
@@ -75,16 +75,25 @@ namespace MGTools
...
   void
-  make_sparsity_pattern(const DoFHandler<dim, spacedim> &dof_handler,
-                        SparsityPatternType &            sparsity,
-                        const unsigned int               level);
+  make_sparsity_pattern(
+    const DoFHandler<dim, spacedim> &dof_handler,
+    SparsityPatternType &            sparsity,
+    const unsigned int               level,
+    const AffineConstraints<number> &constraints = AffineConstraints<number>(),
+    const bool                       keep_constrained_dofs = true);
```
tjhei, line 96: I am not a big fan of bool arguments, especially with default arguments, but, in this case, it is too late for that and it is better to be consistent.

PR #10816, opened 2020-08-11
ECL: Generalize process_and_io

peterrum:
... follow-up to #10810


peterrum:
@kronbichler This is probably the most boring and error prone task I can think of.

peterrum:
> these changes don't make the already present code much worse

You nailed it ;)

PR #10818, opened 2020-08-12
Inform memory leak of sacado_rad_fad in doc.

dougshidong:
As per 

https://groups.google.com/g/dealii/c/BqtkaahE1NQ/m/1kOPQPXdBgAJ

and

https://github.com/trilinos/Trilinos/issues/7741

update the documentation about memory leak when using sacado_rad_fad.


jppelteret:
Since this is just a documentation change, I'll merge it as is.

Issue #9794, opened 2020-04-02
pre-compiled polynomial degrees for FEEvaluation


kronbichler:
Looking at #9790, I see a lot of expensive functions in `EvaluatorTensorProduct::apply`. This gets called in by the templated class `FEEvaluation<dim,fe_degree,n_q_points,n_components,Number>`. While this function is the main worker for the sum factorization algorithms for matrix-free code, its prominent case suggests we should look into this. The main problem is the template on the polynomial degree and the number of quadrature point per direction, which causes too many compilation calls. At the same time, the template is well motivated because it speeds up execution by 3-5 times, given the large number of short loops in the evaluators.

The situation can also become unbearable when developing bigger application codes with many matrixfree evaluation settings (think of nonlinear and linearized variants, homogeneous operators + inhomogeneous/rhs operators, diagonals + forward operator applications) and support for multiple degrees from parameter files.

This issue is to collect ideas about mechanisms for pre-compiled variants. 

**Status and wish list**
- `SelectEvaluator` in `include/deal.II/matrix_free/evaluation_selector.h` does not cover enough cases. [Side remark: I do not fully understand if/how it uses the pre-compiled code from `evaluation_selector.cc` because compile times of code using `SelectEvaluator` get longer when I expand the use to more templated degrees, which should not be the case.] Thus, we should look over and expand it to better fit needs and to better control compile times.
- We should have a similar setup for the face evaluators.
- We would like to let the user specify which degrees and numbers of quadrature points should be precompiled before we go to the slow path with run-time degrees (that is hit when we say `FEEvaluation<dim,-1,0,n_components,Number`).
- We want users to keep selecting the templated variants if they have non-standard cases and do not want to pay the price of a "jump table" in the pre-compiled code.
- If we at some point decide to aggregate the cell and face terms in DG in a single interface (to re-use some pre-computed terms and better data locality) for some of the most common cases, we could reduce everything to a single jump table.

One thing @nfehn, @peterrum and I have been experimenting with in our application development is a specialization of `FEEvaluation` and `FEFaceEvaluation` that expand the internal functions appropriately and work well. However, the user interface is not fully clear - should we integrate those in the existing `-1,0` templates or choose separate class names? Also, the selection of pre-compilation is a bit ad-hoc.

PR #10820, opened 2020-08-12
Remove n_components template argument from FEEvaluationImpl

kronbichler:
This PR does two things:
- It cleans up some of the internal interfaces of `FEEvaluationImpl` as a follow-up to #10369 to only take an `EvaluationFlags::EvaluationFlags` argument rather than a list of bools
- It removes the `n_components` template argument from the internal `FEEvaluationImpl` classes. The components are only a variable to loop over with a lot of work done inside, so I see little value to have this a compile-time constant. This will help to keep the number of instantiations in a possible resolution of #9794 down because we do not need to compile by the number of components, but it should also generally help the compiler reduce the amount of code we generate.

I have not yet touched the implementation for face integrals because @peterrum is currently working on those parts and because the work is orthogonal to this anyway. In that part, we will also need to switch the order of some loops in `gather_evaluate` and `integrate_scatter`.

One thing we could think about is whether we need a changelog because the `SelectEvaluator` class sits in the general namespace, as opposed to the internal namespace for the `FEEvaluationImpl` class itself.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -23,6 +23,7 @@
 #include <deal.II/base/vectorization.h>
 
 #include <deal.II/matrix_free/dof_info.h>
+#include <deal.II/matrix_free/evaluation_flags.h>
 #include <deal.II/matrix_free/shape_info.h>
 #include <deal.II/matrix_free/tensor_product_kernels.h>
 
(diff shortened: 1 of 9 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -551,7 +547,8 @@ namespace internal
         case 3:
           for (unsigned int c = 0; c < n_components; c++)
             {
-              if (integrate_values == true && integrate_gradients == false)
+              if (integration_flag & EvaluationFlags::values &&
```
drwells, line 550:
```suggestion
              if ((integration_flag & EvaluationFlags::values) &&
```
I don't think this is necessary but I can never remember the order of precedence for `&` and `&&`.

source/fe/mapping_q_generic.cc
```diff
@@ -1509,14 +1518,14 @@ namespace internal
           }
 
         // prepare arrays
-        if (evaluate_values || evaluate_gradients || evaluate_hessians)
+        if (evaluation_flag != EvaluationFlags::nothing)
```
drwells, line 1521: :heavy_check_mark:

include/deal.II/matrix_free/operators.h
```diff
@@ -1041,15 +1039,13 @@ namespace MatrixFreeOperators
...
-                                            n_components,
-                                            VectorizedArrayType>::
-      apply(fe_eval.get_shape_info().data.front().inverse_shape_values_eo,
-            inverse_coefficients,
-            n_actual_components,
-            in_array,
-            out_array);
+    internal::
+      CellwiseInverseMassMatrixImpl<dim, fe_degree, VectorizedArrayType>::apply(
+        fe_eval.get_shape_info().data.front().inverse_shape_values_eo,
```
drwells, line 1044: Does it make sense to make `n_components` the first argument here too, for consistency?

kronbichler: Yes, I will do that.

kronbichler:
I improved the code according to the comments.

drwells:
> One thing we could think about is whether we need a changelog because the SelectEvaluator class sits in the general namespace, as opposed to the internal namespace for the FEEvaluationImpl class itself.

I don't feel strongly either way about this - it might be in the general namespace but none of the tests or examples seem to use the old interface.

We should add a changelog entry, though, for #10369.

kronbichler:
> We should add a changelog entry, though, for #10369.

I agree on this reasoning. Let me open a PR now.

drwells:
We let this get stale - would you please fix the merge conflict so that we can merge?

kronbichler:
I rebased.

peterrum (approved):

Issue #5437, opened 2017-11-10
Make TriaAccessor::vertex_index faster


kronbichler:
In lots of my profiles I see `TriaAccessor<structdim, dim, spacedim>::vertex_index` to take an appreciable amount of time. (E.g. in the optimizations done in #5410.) It is used in a lot of functions, in particular all calls to `TriaAccessor::vertex`. The problem is that the implementation is somewhat nested. To get the vertex index of a cell, we ask all the faces for their vertex index (which in 3D then further goes to the lines). I'd wish to have a single index lookup at least for the cell case (`structdim==dim`). This small increase in storage would be definitely worth it in my opinion. I'm not sure where exactly we'd want to apply it. We have a few complications because this is effectively a number cache that needs to be filled and kept up to date. Not sure if this could be attached to the number cache already present... Any opinions?

bangerth:
Ah, glad you asked before you implemented! :-)

The right way to do this is the same we do for the cell dof indices. Take a look at the `DoFLevel::cell_dof_indices_cache` member variable. It is accessed by `DoFCellAccessor::get_dof_indices()` function:
```
template <typename DoFHandlerType, bool level_dof_access>
inline
void
DoFCellAccessor<DoFHandlerType,level_dof_access>::get_dof_indices
(std::vector<types::global_dof_index> &dof_indices) const
{
  Assert (this->active(), ExcMessage ("get_dof_indices() only works on active cells."));
  Assert (this->is_artificial() == false,
          ExcMessage ("Can't ask for DoF indices on artificial cells."));
  AssertDimension (dof_indices.size(), this->get_fe().dofs_per_cell);

  const types::global_dof_index *cache
    = this->dof_handler->levels[this->present_level]
      ->get_cell_cache_start (this->present_index, this->get_fe().dofs_per_cell);
  for (unsigned int i=0; i<this->get_fe().dofs_per_cell; ++i, ++cache)
    dof_indices[i] = *cache;
}
```

After enumerating DoFs, we call `DoFCellAccessor::update_cell_dof_indices_cache()` which has an internal implementation in `dof_accessor.templates.h`, around line 2657 and following, where we go over all of the sub-objects and get DoF indices that we then store in the cache. 

[...]

PR #10828, opened 2020-08-13
Improve performance of ReferenceCell::Info::get_cell

kronbichler:
I performance checks for #10402 and #10820 I realized that `ReferenceCell::internal::Info::get_cell` is a very hot function, see also the remark in #5437. When we use a static array, things should be much faster. However, it seems that at least my gcc-7.3 compiler does not inline this call in release mode. I must admit I don't really understand because it is like 3 or 4 instructions, and the prologue and epilogue of the function call are at least as much code, so not even code size would suggest this I believe. But I don't want to put a `DEAL_II_ALWAYS_INLINE` here for now, as it already improves and smart compilers should be able to use it.

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -1000,43 +1000,23 @@ namespace ReferenceCell
       };
 
       /**
-       * Return for a given reference-cell type @p the right Info.
+       * Return for a given reference-cell type the right Info.
        */
       inline const ReferenceCell::internal::Info::Base &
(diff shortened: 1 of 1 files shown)


kronbichler:
> Normally the compiler should convert the switch to a jump table

It did not in my case (which is why I looked into it in the first place), but then the compiler seems to be making decisions I cannot really follow anyway (code size minimization by not inlining it? a lot of `if/else` statements because the processor can speculate over branches but not indirect jumps?) - so I guess we have 20 lines of removed code plus, at least on my target, faster execution.

PR #10827, opened 2020-08-13
MF: Fix PBC in the case of non-standard orientation

peterrum:
~I am working now on a test.~

PR #10821, opened 2020-08-12
ECL: enable contiguous data access

peterrum:

PR #10830, opened 2020-08-13
Step-67: add ECL

peterrum:
Once #10821 and #10827 is merged, we can run step-67 with ECL (last commit):

step-67-ecl

At the moment, I don't have time to update the documentation and to exploit all advantages of ECL by merging many operations. However, we should try make this part of the next release!

FYI @kkormann, @DavidSCN, @nfehn

### DIFF
--- a/examples/step-67/step-67.cc
+++ b/examples/step-67/step-67.cc
@@ -1748,24 +1748,18 @@ namespace Euler_DG
                   const Number ai = factor_ai;
                   const Number bi = factor_solution;
 
-                  phi_temp.reinit(cell);
-                  phi_temp.read_dof_values(solution);
-
                   if (ai == Number())
                     {
                       for (unsigned int q = 0; q < phi.static_dofs_per_cell;
                            ++q)
-                        {
-                          phi_temp.begin_dof_values()[q] +=
-                            bi * phi.begin_dof_values()[q];
-                          // phi.begin_dof_values()[q] =
-                          //  phi_temp.begin_dof_values()[q];
-                        }
-                      // phi.set_dof_values(dst);
-                      phi_temp.set_dof_values(solution);
+                        phi.begin_dof_values()[q] =
+                          bi * phi.begin_dof_values()[q];
+                      phi.distribute_local_to_global(solution);
                     }
                   else
                     {
+                      phi_temp.reinit(cell);
+                      phi_temp.read_dof_values(solution);
                       for (unsigned int q = 0; q < phi.static_dofs_per_cell;
                            ++q)
                         {
@@ -1774,10 +1768,10 @@ namespace Euler_DG
                           phi.begin_dof_values()[q] =
                             phi_temp.begin_dof_values()[q] + (ai * K_i);
 
-                          phi_temp.begin_dof_values()[q] += bi * K_i;
+                          phi_temp.begin_dof_values()[q] = bi * K_i;
                         }
                       phi.set_dof_values(dst);
-                      phi_temp.set_dof_values(solution);
+                      phi_temp.distribute_local_to_global(solution);
                     }
                 }
               }
--- a/examples/step-67/step-67.cc
+++ b/examples/step-67/step-67.cc
@@ -80,6 +80,8 @@ namespace Euler_DG
   constexpr unsigned int fe_degree            = 5;
   constexpr unsigned int n_q_points_1d        = fe_degree + 2;
 
+  constexpr bool use_ecl = true;
+
   using Number = double;
 
   constexpr double gamma       = 1.4;
@@ -380,8 +382,7 @@ namespace Euler_DG
     {
       AssertDimension(ai.size() + 1, bi.size());
 
-#ifdef USE_ECL
-      if (true)
+      if (use_ecl)
         {
           double sum_previous_bi = 0;
           for (unsigned int stage = 0; stage < bi.size(); ++stage)
@@ -405,8 +406,7 @@ namespace Euler_DG
           // Switch for odd stages
           sw = sw == 0;
         }
-#else
-      if (true)
+      else
         {
           pde_operator.perform_stage(current_time,
                                      bi[0] * time_step,
@@ -431,7 +431,6 @@ namespace Euler_DG
               sum_previous_bi += bi[stage - 1];
             }
         }
-#endif
     }
 
   private:
@@ -898,10 +897,9 @@ namespace Euler_DG
     additional_data.tasks_parallel_scheme =
       MatrixFree<dim, Number>::AdditionalData::none;
 
-#ifdef USE_ECL
-    MatrixFreeTools::categorize_by_boundary_ids(dof_handler.get_triangulation(),
-                                                additional_data);
-#endif
+    if (use_ecl)
+      MatrixFreeTools::categorize_by_boundary_ids(
+        dof_handler.get_triangulation(), additional_data);
 
     data.reinit(
       mapping, dof_handlers, constraints, quadratures, additional_data);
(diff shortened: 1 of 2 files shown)


peterrum:
I have rebased this PR. Let's use this example to discuss about a nice interface of ECL!

include/deal.II/matrix_free/tools.h
```diff
@@ -27,7 +29,82 @@ DEAL_II_NAMESPACE_OPEN
...
+
+    // ... finalize setup of matrix_free
+    additional_data.hold_all_faces_to_owned_cells        = true;
+    additional_data.cell_vectorization_categories_strict = true;
+    additional_data.mapping_update_flags_faces_by_cells =
+      additional_data.mapping_update_flags_inner_faces |
+      additional_data.mapping_update_flags_boundary_faces;
+  }
+
+} // namespace MatrixFreeTools
```
peterrum, line 107: Moved to PR #10871.

include/deal.II/matrix_free/shape_info.h
```diff
@@ -229,6 +229,18 @@ namespace internal
...
+       * support of the quadrature rule) evaluated at the point 0 and 1
+       * (the vertices) in one data structure.
+       *
+       * This data structure can be used to interpolate from the cell to the
+       * face quadrature points.
+       *
+       * @note In contrast to shape_data_on_face, only the vales are evaluated.
+       */
+      AlignedVector<Number> quadrature_data_on_face[2];
+
```
peterrum, line 243: Moved to #10871.

include/deal.II/matrix_free/shape_info.templates.h
```diff
@@ -112,8 +112,10 @@ namespace internal
         univariate_shape_data.shape_gradients_collocation;
       auto &shape_hessians_collocation =
         univariate_shape_data.shape_hessians_collocation;
-      auto &inverse_shape_values  = univariate_shape_data.inverse_shape_values;
-      auto &shape_data_on_face    = univariate_shape_data.shape_data_on_face;
+      auto &inverse_shape_values = univariate_shape_data.inverse_shape_values;
+      auto &shape_data_on_face   = univariate_shape_data.shape_data_on_face;
+      auto &quadrature_data_on_face =
+        univariate_shape_data.quadrature_data_on_face;
```
peterrum, line 118: Moved to #10871.

include/deal.II/matrix_free/shape_info.templates.h
```diff
@@ -298,6 +300,20 @@ namespace internal
...
+
+      for (unsigned int i = 0; i < quad.size(); ++i)
+        {
+          Point<1> q_point;
+          q_point[0]                    = 0;
+          quadrature_data_on_face[0][i] = fe_quad.shape_value(i, q_point);
+          q_point[0]                    = 1;
+          quadrature_data_on_face[1][i] = fe_quad.shape_value(i, q_point);
+        }
+
```
peterrum, line 316: Moved to #10871.

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -1754,6 +1854,71 @@ namespace internal
...
+                  input, output);
+            }
+          input += in_stride;
+          output += out_stride;
+        }
+    }
+  };
+
+
+
```
peterrum, line 1921: Moved to #10871.

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -896,6 +896,106 @@ namespace internal
...
+                  values_out +
+                    q * dealii::Utilities::fixed_power<next_dim>(np_1),
+                  basis_size_1_variable,
+                  basis_size_2_variable);
+
+          values_in += dealii::Utilities::fixed_power<dim>(np_2);
+          values_out += dealii::Utilities::fixed_power<dim>(np_1);
+        }
+    }
+
```
peterrum, line 998: Moved to PR #10869.

examples/step-67/step-67.cc
```diff
@@ -1465,6 +1507,316 @@ namespace Euler_DG
...
+                degree + 1,
+                n_points_1d,
+                dim + 2,
+                VectorizedArray<Number>,
+                VectorizedArray<Number>>::
+                do_backward_hessians(
+                  data.get_shape_info().data[0].inverse_shape_values_eo,
+                  false,
+                  phi.begin_values(),
+                  phi.begin_dof_values());
```
peterrum, line 1747: I am still not 100% happy. I had to use different internal helper classes to make the algorithm working. @kronbichler Do you have any suggestions how to make code cleaner?

examples/step-67/step-67.cc
```diff
@@ -1465,6 +1501,288 @@ namespace Euler_DG
...
+                  if (dim >= 2)
+                    eval.template gradients<1, false, true>(
+                      grdient_ptr + phi.static_n_q_points * 1, values_ptr);
+                  if (dim >= 3)
+                    eval.template gradients<2, false, true>(
+                      grdient_ptr + phi.static_n_q_points * 2, values_ptr);
+
+                  values_ptr += phi.static_n_q_points;
+                  grdient_ptr += phi.static_n_q_points * dim;
+                }
```
kronbichler, line 1605: This is one of the places we would like to abstract away. We have https://github.com/dealii/dealii/blob/e9461deb2e453f353d6bcab7d571aa2d4715ca47/include/deal.II/matrix_free/evaluation_kernels.h#L1108-L1118 which does exactly this when called with an `EvaluationFlags::gradient` argument. In the end, I guess we would need something like `integrate_to_collocated_points`.

examples/step-67/step-67.cc
```diff
@@ -1465,6 +1501,288 @@ namespace Euler_DG
...
+              if (body_force.get() == nullptr)
+                {
+                  for (unsigned int i = 0;
+                       i < phi.static_n_q_points * (dim + 2);
+                       ++i)
+                    phi.begin_values()[i] = 0.0;
+                }
+
+              VectorizedArray<Number> *values_ptr  = phi.begin_values();
+              VectorizedArray<Number> *grdient_ptr = phi.begin_gradients();
```
kronbichler, line 1589: By the way, there is a typo in the variable name.

examples/step-67/step-67.cc
```diff
@@ -1465,6 +1501,288 @@ namespace Euler_DG
...
+                n_points_1d,
+                VectorizedArray<Number>,
+                VectorizedArray<Number>>::
+                do_backward_hessians(
+                  dim + 2,
+                  data.get_shape_info().data[0].inverse_shape_values_eo,
+                  false,
+                  phi.begin_values(),
+                  phi.begin_dof_values());
+
```
kronbichler, line 1744: Isn't this https://github.com/dealii/dealii/blob/e9461deb2e453f353d6bcab7d571aa2d4715ca47/include/deal.II/matrix_free/evaluation_kernels.h#L3238-L3272

examples/step-67/step-67.cc
```diff
@@ -374,38 +375,70 @@ namespace Euler_DG
                            const double    current_time,
                            const double    time_step,
                            VectorType &    solution,
-                           VectorType &    vec_ri,
-                           VectorType &    vec_ki) const
+                           VectorType &    vec_Ti,
+                           VectorType &    vec_Ki) const
```
kronbichler, line 379: Can you explain why you change the variable names?

examples/step-67/step-67.cc
```diff
@@ -1465,6 +1501,288 @@ namespace Euler_DG
...
+                                 v < VectorizedArray<Number>::size();
+                                 ++v)
+                              {
+                                if (rho_u_dot_n[v] < -1e-12)
+                                  for (unsigned int d = 0; d < dim; ++d)
+                                    flux[d + 1][v] = 0.;
+                              }
+
+                          phi_m.submit_value(-flux, q);
+                        }
```
kronbichler, line 1706: For all the code above, we should have an `always_inline` function that does the operation at a quadrature point. This should be exactly the same as in the respective face function.

examples/step-67/step-67.cc
```diff
@@ -2243,6 +2561,10 @@ namespace Euler_DG
 
     euler_operator.project(ExactSolution<dim>(time), solution);
 
+#ifdef USE_ECL
+    rk_register_2 = solution;
+#endif
+
```
kronbichler, line 2567: Can you explain why this is needed? In other words, why do we assume that we have another copy of `solution`? Furthermore, I assume what you actually want here is `rk_register_2.copy_locally_owned_data_from(solution);`

kronbichler (commented):
I have a few comments on variable levels. I will have a closer look once we are through with the separate functions.

examples/step-67/step-67.cc
DavidSCN: Maybe I can explain: For cell-centric loops, there is only one source rk_register in each RK sweep and two destination vectors, namely  the other rk_register and the solution vector. Both are updated in the `loop_cell_centric` function due to the merging of the inverse mass matrix operation with the RK integral evaluation. If we read and write in the same vector, we overwrite potentially needed data for face integrals/ evaluation of neighbor data. So, we need to have the projected initial data as source data available and the solution vector as destination vector to collect intermediate RK results. 

For face-centric loops, the solution vector is used as source vector during the RK sweep and updated separately during the element-wise inverse mass-matrix operation.

Alternatively, we could, use the rk_register directly in the `project` function, instead of the solution vector. I'm not sure which of both strategies is more intuitive.

kronbichler: It is clear that the result of an RK stage cannot be written to the vector where we read from, which is the solution in the first stage. My point was that this copy is unexpected mathematically, where the assumption is that all that matters is the `solution` vector containing the data from the old time step. So I would much prefer if this could be reflected in the code to avoid surprises; I guess the question is whether to achieve that for some smart swapping of entries or an extra step in the first RK stage.

DavidSCN: Ok, I see your point. In the current implementation, one rk register holds the solution vector between time steps since we copy the data of the final solution vector in the last RK stage.

https://github.com/peterrum/dealii/blob/6c0333bdd584ca4b922a126f01cb29a0cbb06261/examples/step-67/step-67.cc#L1757-L1759

We cannot fully preserve the concept of three vectors, where the `solution` vector always holds the solution data: at some point, we need to use the `solution` as source and write in another destination vector. But we can switch the solution copy into the first RK stage instead of the last RK stage so that the `solution` vector contains the relevant data between different time steps. It results , however, in another read operation of the whole solution vector because we need to finish the first stage before copying the data. The current implementation avoid this additional read since the solution vector is already loaded in the last RK stage.

examples/step-67/step-67.cc
```diff
@@ -54,8 +54,9 @@
 // that we will use for the mass matrix inversion, the only new include
 // file for this tutorial program:
 #include <deal.II/matrix_free/operators.h>
+#include <deal.II/matrix_free/tools.h>
 
-
+#define USE_ECL
```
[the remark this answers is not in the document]
peterrum: No, I won't: it will be gone soon,

examples/step-67/step-67.cc
peterrum: @DavidSCN Let's discuss this tomorrow in person.

peterrum: @DavidSCN Thanks for the discussion today! I have updated the code accordingly.

@kronbichler What do you think about this version?

peterrum:
With the following configuration:

```diff
diff --git a/examples/step-67/step-67.cc b/examples/step-67/step-67.cc
index e3d2e83..9935fd6 100644
--- a/examples/step-67/step-67.cc
+++ b/examples/step-67/step-67.cc
@@ -75,8 +75,8 @@ namespace Euler_DG
   // `output_tick` that specifies in which intervals we want to write output
   // (assuming that the tick is larger than the time step size).
   constexpr unsigned int testcase             = 1;
-  constexpr unsigned int dimension            = 2;
-  constexpr unsigned int n_global_refinements = 3;
+  constexpr unsigned int dimension            = 3;
+  constexpr unsigned int n_global_refinements = 2;
   constexpr unsigned int fe_degree            = 5;
   constexpr unsigned int n_q_points_1d        = fe_degree + 2;
 
@@ -2622,7 +2622,7 @@ namespace Euler_DG
     // mostly done by the TimerOutput::print_wall_time_statistics() function.
     unsigned int timestep_number = 0;
 
-    while (time < final_time - 1e-12)
+    while (time < final_time - 1e-12 && timestep_number < 100)
       {
         ++timestep_number;
         if (timestep_number % 5 == 0)
```

I get the following results our gold nodes (40 cores):
```
FCL:
+-------------------------------------------+------------------+------------+------------------+
| Total wallclock time elapsed              |                  |            |                  |
|                                           |                  |                               |
| Section                       | no. calls |   min time  rank |   avg time |   max time  rank |
+-------------------------------------------+------------------+------------+------------------+
| rk time stepping total        |       100 |     11.39s     0 |     11.54s |     11.64s    34 |
| rk_stage - integrals L_h      |       500 |     8.954s    14 |     9.143s |      9.44s     0 |
| rk_stage - inv mass + vec upd |       500 |     1.935s     0 |     2.387s |      2.54s    36 |
+-------------------------------------------+------------------+------------+------------------+

ECL:
+--------------------------------------+------------------+------------+------------------+
| Total wallclock time elapsed         |                  |            |                  |
|                                      |                  |                               |
| Section                  | no. calls |   min time  rank |   avg time |   max time  rank |
+--------------------------------------+------------------+------------+------------------+
| rk time stepping total   |       100 |     10.04s     0 |     10.25s |      10.4s    34 |
| rk_stage - integrals L_h |       500 |     10.03s     0 |     10.24s |     10.39s    34 |
+--------------------------------------+------------------+------------+------------------+
```

which corresponds to a speed-up of about 13%. I guess I'll create a LIKWID cache analysis to get more insights.

examples/step-67/step-67.cc
```diff
@@ -73,12 +74,14 @@ namespace Euler_DG
   // the final time up to which we run the simulation, and a variable
   // `output_tick` that specifies in which intervals we want to write output
   // (assuming that the tick is larger than the time step size).
-  constexpr unsigned int testcase             = 0;
+  constexpr unsigned int testcase             = 1;
```
kronbichler, line 77: please undo this change.

examples/step-67/step-67.cc
```diff
@@ -73,12 +74,14 @@ namespace Euler_DG
...
   constexpr unsigned int dimension            = 2;
   constexpr unsigned int n_global_refinements = 3;
   constexpr unsigned int fe_degree            = 5;
   constexpr unsigned int n_q_points_1d        = fe_degree + 2;
 
+  constexpr bool use_ecl = true;
```
kronbichler, line 83: Please document this above, maybe also in combination with a more expressive variable name.

examples/step-67/step-67.cc
```diff
@@ -1465,60 +1505,362 @@ namespace Euler_DG
...
+        AlignedVector<VectorizedArray<Number>> buffer(phi.static_n_q_points *
+                                                      phi.n_components);
+
+        data.template loop_cell_centric<
+          LinearAlgebra::distributed::Vector<Number>,
+          LinearAlgebra::distributed::Vector<Number>>(
+          [&](const auto &, auto &dst, const auto &src, const auto cell_range) {
+            for (unsigned int cell = cell_range.first; cell < cell_range.second;
+                 ++cell)
+              {
```
kronbichler, line 1552: I know you like lambdas, but for the sake of education let us move this large function into a separate named function that can be easily identified in all tools (debuggers, profilers, ...) and is consistent with the rest of the program.

examples/step-67/step-67.cc
```diff
@@ -1465,60 +1505,362 @@ namespace Euler_DG
...
+                  {
+                    if (dim >= 1)
+                      eval.template gradients<0, false, true>(
+                        grdient_ptr + phi.static_n_q_points * 0, values_ptr);
+                    if (dim >= 2)
+                      eval.template gradients<1, false, true>(
+                        grdient_ptr + phi.static_n_q_points * 1, values_ptr);
+                    if (dim >= 3)
+                      eval.template gradients<2, false, true>(
+                        grdient_ptr + phi.static_n_q_points * 2, values_ptr);
```
kronbichler, line 1622: Should we look into possibilities to hide some of these functions behind library functions?

kronbichler (commented):
@peterrum what is your time plan on this? I think we should try to get it done timely before the next release, so I left a few more comments.

peterrum:
> what is your time plan on this? I think we should try to get it done timely before the next release, so I left a few more comments.

It would be nice to get this merged soon. I will rebase the code and go through the changes again later this week. I will need to remember the issues I had (my last commit here is more than 2 months ago).

peterrum:
Closing in favor of #11158.

PR #10834, opened 2020-08-17
Run test suite also for json files

peterrum:
This is already working for `.prm`-files.

FYI @nfehn


drwells:
This looks reasonable but, just to check, how does this work with prm files? Is this documented somewhere?

peterrum:
@drwells Here is the relevant information of prm-files: https://www.dealii.org/developer/users/testsuite.html. I have added now an entry for json-files.

PR #10835, opened 2020-08-17
Optional argument for DEAL_II_PICKUP_TESTS

peterrum:
At the moment test are listed as *current_directory_name/file_name (without .cc)* during ctest. However, we have in an in-house-code multiple test folders with the name `tests` which would lead to the circumstance that all the test would start with `tests/` and there would be no grouping according to topic.

This PR enables to add to the folder with the test a `CMakeLists.txt` file with the following content:
```cmake
SET (TEST_LIBRARIES hyperdeal) # this is not the actual application ;) 
DEAL_II_PICKUP_TESTS(poisson)
```
All tests in that folder in that folder will start with `poisson/`

FYI @nfehn


peterrum:
@drwells Done!

cmake/macros/macro_deal_ii_pickup_tests.cmake
```diff
@@ -183,7 +183,15 @@ MACRO(DEAL_II_PICKUP_TESTS)
   ENABLE_TESTING()
 
   SET_IF_EMPTY(TEST_PICKUP_REGEX "$ENV{TEST_PICKUP_REGEX}")
-  GET_FILENAME_COMPONENT(_category ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+
+  SET (extra_macro_args ${ARGN})
```
drwells, line 187:
```suggestion
  SET(extra_macro_args ${ARGN})
```

cmake/macros/macro_deal_ii_pickup_tests.cmake
```diff
@@ -183,7 +183,15 @@ MACRO(DEAL_II_PICKUP_TESTS)
...
+  LIST(LENGTH extra_macro_args num_extra_args)
+  if (${num_extra_args} GREATER 0)
```
drwells, line 189:
```suggestion
  IF(${num_extra_args} GREATER 0)
```

cmake/macros/macro_deal_ii_pickup_tests.cmake
```diff
@@ -183,7 +183,15 @@ MACRO(DEAL_II_PICKUP_TESTS)
...
+    LIST(GET extra_macro_args 0 optional_arg)
+    SET(_category ${optional_arg})
+  else()
```
drwells, line 192:
```suggestion
  ELSE()
```

cmake/macros/macro_deal_ii_pickup_tests.cmake
```diff
@@ -183,7 +183,15 @@ MACRO(DEAL_II_PICKUP_TESTS)
...
+    GET_FILENAME_COMPONENT(_category ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+  endif ()
```
drwells, line 194:
```suggestion
  ENDIF()
```

cmake/macros/macro_deal_ii_pickup_tests.cmake
```diff
@@ -183,7 +183,15 @@ MACRO(DEAL_II_PICKUP_TESTS)
...
-  GET_FILENAME_COMPONENT(_category ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+
+  SET(extra_macro_args ${ARGN})
+  LIST(LENGTH extra_macro_args num_extra_args)
+  IF (${num_extra_args} GREATER 0)
+    LIST(GET extra_macro_args 0 optional_arg)
+    SET(_category ${optional_arg})
+  ELSE()
+    GET_FILENAME_COMPONENT(_category ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+  ENDIF()
```
tamiko, line 194: Would you mind shrinking this down to five lines as follows (untested):
```
IF("${ARGN}" STREQUAL "")
  GET_FILENAME_COMPONENT(_category ${CMAKE_CURRENT_SOURCE_DIR} NAME)
ELSE()
  SET(_category "${ARGN}")
ENDIF()
```

tamiko: I like the change - sorry for the nitpick

peterrum: I'll try it in a minute.

peterrum: Yup works!

tamiko: Thanks!

PR #10836, opened 2020-08-17
DataOut::write_vtu_with_pvtu_record(): prints the path name if file cannot be opened

peterrum:
FYI @nfehn

### DIFF
--- a/tests/data_out/write_vtu_with_pvtu_record_no_folder_01.mpirun=1.output
+++ b/tests/data_out/write_vtu_with_pvtu_record_no_folder_01.mpirun=1.output
@@ -1,0 +1,4 @@
+
+DEAL:0::    Could not open file ./folder/file_0.0.vtu.
+DEAL:0::    Could not open file ./folder/file_0.0.vtu.
+DEAL:0::    Could not open file ./folder/file_0.0.vtu.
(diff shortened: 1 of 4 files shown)


tests/data_out/write_vtu_with_pvtu_record_no_folder_01.mpirun=1.output
```diff
@@ -0,0 +1,31 @@
+
+DEAL:0::
+--------------------------------------------------------
+An error occurred in file <data_out_base.cc> in function
+    std::string dealii::DataOutInterface<dim, spacedim>::write_vtu_with_pvtu_record(const string&, const string&, unsigned int, ompi_communicator_t* const&, unsigned int, unsigned int) const [with int dim = 2; int spacedim = 2; std::string = std::__cxx11::basic_string<char>; MPI_Comm = ompi_communicator_t*]
```
tjhei, line 5: > std::string = std::__cxx11::basic_string<char>; MPI_Comm = ompi_communicator_t*

This is problematic as compiler and MPI details are part of the error message, so this will be fragile. If you remove the output, the tests becomes somewhat useless. :man_shrugging:

peterrum: You are right. What about this approach?

PR #10838, opened 2020-08-18
Destroy tmp matrix in mmult.

richardschu:
When using a vector in the mmult, a tmp matrix with scaled rows is not destroyed, which leads to a memory leak.
This can be fixed using PETScWrappers::destroy_matrix(tmp).
-So, please verify!
Regards,
Richard

### DIFF
--- a/source/lac/petsc_matrix_base.cc
+++ b/source/lac/petsc_matrix_base.cc
@@ -556,6 +556,8 @@ namespace PETScWrappers
                             PETSC_DEFAULT,
                             &result.petsc_matrix());
           AssertThrow(ierr == 0, ExcPETScError(ierr));
+          ierr = PETScWrappers::destroy_matrix(tmp);
+			    AssertThrow(ierr == 0, ExcPETScError(ierr));
         }
     }
   } // namespace internals
--- a/source/lac/petsc_matrix_base.cc
+++ b/source/lac/petsc_matrix_base.cc
@@ -557,7 +557,7 @@ namespace PETScWrappers
                             &result.petsc_matrix());
           AssertThrow(ierr == 0, ExcPETScError(ierr));
           ierr = PETScWrappers::destroy_matrix(tmp);
-			    AssertThrow(ierr == 0, ExcPETScError(ierr));
+          AssertThrow(ierr == 0, ExcPETScError(ierr));
         }
     }
   } // namespace internals


source/lac/petsc_matrix_base.cc
```diff
@@ -556,6 +556,8 @@ namespace PETScWrappers
                             PETSC_DEFAULT,
                             &result.petsc_matrix());
           AssertThrow(ierr == 0, ExcPETScError(ierr));
+          ierr = PETScWrappers::destroy_matrix(tmp);
```
tjhei, line 559: This looks correct to me, we need to delete the temporary. What is ``MatMatMult`` with ``MAT_INITIAL_MATRIX`` doing to the existing ``result`` matrix? This could also be a leak or not?

tjhei:
Thanks for your contribution, can you check your git authorship information, please?
```
invalid author 'richardschu' without firstname and lastname

```

source/lac/petsc_matrix_base.cc
richardschu: In the google group we discussed on a little dummy example and I shared my code there:
https://groups.google.com/g/dealii/c/ad6-kfYP2z0
-Running on my machine that looked fine (doing the mmult with a vector 1000 times with a large-enough matrix).

richardschu:
Ok so I added my full name to the git profile, does that do the trick?
-Sorry for that, I am completely new to git and was unfortunately not aware of the dealii-rules when it comes to commit.

masterleinad:
No, you should have a look at https://git-tower.com/learn/git/faq/change-author-name-email for changing the author's information for past commits.

source/lac/petsc_matrix_base.cc
bangerth: @tjhei: The documentation doesn't say: https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MatMatMult.html

But the argument is simply a `Mat *`, not a `Mat *&`, and so the function clearly expects a matrix object that has been created via `MatCreate` or similar before because it can't return a matrix that it creates itself. 

The implementation is here: https://www.mcs.anl.gov/petsc/petsc-current/src/mat/interface/matrix.c.html#MatMatMult and calls line 9261 from where we get to this place: https://www.mcs.anl.gov/petsc/petsc-current/src/mat/interface/matproduct.c.html#MatProductCreate from where I'm losing track of where it goes. But I suspect that it keeps the object and replaces the content.

tjhei: I guess if a test confirms that we are not leaking PETSc objects when repeatedly calling this function, then we are good.

richardschu:
> No, you should have a look at https://git-tower.com/learn/git/faq/change-author-name-email for changing the author's information for past commits.

Thanks for the hint, I had to do it with git rebase and amend.
-Hope, that works out now?

tjhei:
You did not push your changes yet. If you rewrote your commit, you will need to ``git push -f``

richardschu:
So, the problem was that I did the changes in my browser only - which is why I hadn't any files locally saved.
I checked my commit out and changed the author of the last two commits in this branch using rebase, then I pushed to master.
I checked downloading that commit again and I see the correct author info right away.
- I guess I got it right finally!?
Thanks for commenting everyone!

peterrum:
>  then I pushed to master.

I think you want to push onto the branch `patch-1`.

richardschu:
Hi again,
thanks for the hint, I just checked & there is actually only one branch called 'master' of 'richardschu::dealii'
- I would assume that this is correct, since I cannot push anywhere else
...if that is not the case I can offer to close and commit again with a correct author info.
Let me know!

peterrum:
> I would assume that this is correct, since I cannot push anywhere else
...if that is not the case I can offer to close and commit again with a correct author info.

You need to force push. Let's assume the name of the branch locally is called X and remote Y, then you need write:
```
git push -f Y X:patch-1
```

richardschu:
> You need to force push. Let's assume the name of the branch locally is called X and remote Y, then you need write:
> 
> ```
> git push -f Y X:patch-1
> ```

I did 'git push origin master' and it worked nicely (?), since
'git remote -v' gave me:
origin	https://github.com/richardschu/dealii.git (fetch)
origin	https://github.com/richardschu/dealii.git (push)
And it told me 'Everything up-to-date' after entering my git name & password.

tjhei:
> I did 'git push origin master' and it worked nicely (?),

I think you misunderstood. When creating a pull request, it uses a specific branch that you specify. In this case, it is "patch-1" (you can see this if you scroll to the top of this page). That means that you need to update https://github.com/richardschu/dealii/tree/patch-1 and not master.

masterleinad:
In case your local `master` branch contains the commits you want in this pull request, you would do
```
git push -f origin master:patch-1
```

peterrum (approved):

tjhei:
Thanks again, @richardschu . It looks like it was more difficult than it should be, but your PR is finally merged. The next pull request will be a lot easier. :-)

richardschu:
Ok, well thanks for the guidance & patience - I hope the hrs/line ratio is better in the future! 😉

PR #10840, opened 2020-08-21
Use DiscreteTime in step-52

rezarastak:
Possibly a fix for #9684

### DIFF
--- a/examples/step-52/doc/results.dox
+++ b/examples/step-52/doc/results.dox
@@ -10,28 +10,27 @@ The console output contains both errors and, for some of the methods, the
 number of steps they performed:
 @code
 Explicit methods:
-Forward Euler:            error=1.00883
-Third order Runge-Kutta:  error=0.000227982
-Fourth order Runge-Kutta: error=1.90541e-06
+   Forward Euler:            error=1.00883
+   Third order Runge-Kutta:  error=0.000227982
+   Fourth order Runge-Kutta: error=1.90541e-06
 
 Implicit methods:
-Backward Euler:           error=1.03428
-Implicit Midpoint:        error=0.00862702
-Crank-Nicolson:           error=0.00862675
-SDIRK:                    error=0.0042349
+   Backward Euler:           error=1.03428
+   Implicit Midpoint:        error=0.00862702
+   Crank-Nicolson:           error=0.00862675
+   SDIRK:                    error=0.0042349
 
-Embedded %explicit methods:
-Heun-Euler:               error=0.0073012
-                steps performed=284
-Bogacki-Shampine:         error=0.000403281
-                steps performed=181
-Dopri:                    error=0.0165485
-                steps performed=119
-Fehlberg:                 error=0.00104926
-                steps performed=106
-Cash-Karp:                error=8.59366e-07
-                steps performed=107
-
+Embedded explicit methods:
+   Heun-Euler:               error=0.0073012
+                   steps performed=284
+   Bogacki-Shampine:         error=0.000408407
+                   steps performed=181
+   Dopri:                    error=0.000836695
+                   steps performed=120
+   Fehlberg:                 error=0.00248922
+                   steps performed=106
+   Cash-Karp:                error=0.0787735
+                   steps performed=106
 @endcode
 
 As expected the higher order methods give (much) more accurate solutions. We
--- a/examples/step-52/step-52.cc
+++ b/examples/step-52/step-52.cc
@@ -21,6 +21,7 @@
 
 // The first task as usual is to include the functionality of these well-known
 // deal.II library files and some C++ header files.
+#include <deal.II/base/discrete_time.h>
 #include <deal.II/base/function.h>
 #include <deal.II/base/quadrature_lib.h>
 
@@ -485,6 +486,7 @@ namespace Step52
   // constraints are respected; of course, with the mesh we use here,
   // hanging node constraints are not in fact an issue). It then calls
   // <code>evolve_one_time_step</code> which performs one time step.
+  // Time is stored and incremented through a DiscreteTime object.
   //
   // For explicit methods, <code>evolve_one_time_step</code> needs to
   // evaluate $M^{-1}(f(t,y))$, i.e, it needs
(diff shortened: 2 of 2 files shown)


examples/step-52/step-52.cc
```diff
@@ -616,34 +631,33 @@ namespace Step52
...
-          solution);
+        const double new_time =
+          embedded_explicit_runge_kutta.evolve_one_time_step(
+            [this](const double time, const Vector<double> &y) {
+              return this->evaluate_diffusion(time, y);
+            },
+            time.get_current_time(),
+            time.get_next_step_size(),
+            solution);
+        time.set_next_step_size(new_time - time.get_current_time());
```
tjhei, line 646: I am not sure I understand why you do this: isn't the RK always going to return ``get_current_time+get_next_step_size``? I think this line is not necessary.

examples/step-52/step-52.cc
```diff
@@ -616,34 +631,33 @@ namespace Step52
...
-
-    // Now for the time loop. The last time step is chosen such that the final
-    // time is exactly reached.
-    unsigned int n_steps = 0;
-    while (time < final_time)
+    output_results(initial_time, 0, method);
+    DiscreteTime time(initial_time, final_time, time_step);
+    while (time.is_at_end() == false)
       {
-        if (time + time_step > final_time)
```
tjhei, line 626: so much nicer! :+1:

examples/step-52/doc/results.dox
```diff
@@ -10,28 +10,27 @@ The console output contains both errors and, for some of the methods, the
...
+Embedded explicit methods:
+   Heun-Euler:               error=0.0073012
+                   steps performed=284
+   Bogacki-Shampine:         error=0.000408407
+                   steps performed=181
+   Dopri:                    error=0.000836695
+                   steps performed=120
+   Fehlberg:                 error=0.00248922
+                   steps performed=106
+   Cash-Karp:                error=0.0787735
```
tjhei, line 32: why is this error so much bigger?

examples/step-52/step-52.cc
rezarastak: This line is absolutely necessary since the function `evolve_one_time_step()` may reduce the current dt (in addition to the next dt) internally.
https://github.com/dealii/dealii/blob/ff89e12c96f2186d84e14dc3b7d798e2461710ce/include/deal.II/base/time_stepping.templates.h#L809-L815

examples/step-52/doc/results.dox
rezarastak: In this case, my hypothesis is that the previous code was using a really small time-step at the end which caused the error to go way down. `DiscreteTime` objects can decide to extend the last time-step just a little bit to avoid a tiny final dt.
https://github.com/dealii/dealii/blob/ff89e12c96f2186d84e14dc3b7d798e2461710ce/source/base/discrete_time.cc#L27-L40

Unfortunately, I have not put print statements in either the old and the new code to confirm my hypothesis yet, but the hypothesis agrees with the following observations: 
* The error values don't change at all for most algorithms.
* The result for Cash-Karp has now one fewer time-steps and the error value which used to be unbelievably small (even below the coarsening threshold) is now reasonable (based on the adaptive time-stepping algorithm).

rezarastak: Even before changing the code, the error values I got with my machine was different than those presented in the tutorial (similar to what is being discussed in #9684). What follows is the summary of discrepancies.
Results provided in the online tutorial:
```
Bogacki-Shampine:         error=0.000403281
                steps performed=181
Dopri:                    error=0.0165485
                steps performed=119
Fehlberg:                 error=0.00104926
                steps performed=106
Cash-Karp:                error=8.59366e-07
                steps performed=107
```
Results of the master branch on my machine (identical for both debug and release)
```
   Bogacki-Shampine:         error=0.000403281
                steps performed=181
   Dopri:                    error=5.07155e-05
                steps performed=121
   Fehlberg:                 error=0.00401035
                steps performed=106
   Cash-Karp:                error=0.00156445
                steps performed=107
```
Results after making my changes with `DiscreteTime` (identical both debug and release)
```
   Bogacki-Shampine:         error=0.000408407
                   steps performed=181
   Dopri:                    error=0.000836695
                   steps performed=120
   Fehlberg:                 error=0.00248922
                   steps performed=106
   Cash-Karp:                error=0.0787735
                   steps performed=106
```
This also supports the hypothesis that I mentioned in my previous comment. After making the changes, both Dopri and Cash-Karp get one fewer time step but higher level of error (still below the refinement threshold of 0.1).

Rombur: Yes, that's correct we already discussed this in https://github.com/dealii/dealii/issues/9684

tjhei: Sorry, I didn't follow that conversation very closely.

I guess the test problem is a bit unfortunate if the size of the final time step determines the total error of the solution.

examples/step-52/step-52.cc
tjhei: Ah, yes. It is a bit unfortunate that the 3 loops looks different now or what do you think?

rezarastak: I have already added these comments before the aforementioned code to explain it better.

```c++
   // embedded method. In summary, time step size is potentially modified in
   // three ways:
   //   - Reducing or increasing time step size within
   //     TimeStepping::EmbeddedExplicitRungeKutta::evolve_one_time_step().
   //   - Using the calculated <code>delta_t_guess</code>.
   //   - Automatically adjusting the step size of the last time step to ensure
   //     simulation ends precisely at <code>final_time</code>. This adjustment
   //     is handled inside the DiscreteTime instance.
```

examples/step-52/doc/results.dox
Rombur: That's to be expected with the embedded methods https://github.com/dealii/dealii/issues/9684#issuecomment-627066689

rezarastak:
I would be happy to hear more feedback on this. Otherwise, it is perhaps ready for final review/merging.

tjhei (approved):
Sorry, @rezarastak . Looks good to me.

tamiko (approved):

PR #10846, opened 2020-08-26
Update GridTools::transform for use with simplex meshes

elauksap:
This modification makes `GridTools::transform()` working with **non-refined** simplex meshes.

As far as I know simplex mesh refinement is not supported yet: once available, the loop over faces to adjust the position of hanging nodes can be adapted accordingly to follow the same ordering and numbering convention of children.

### DIFF
--- a/doc/news/changes/minor/20200826PasqualeAfrica
+++ b/doc/news/changes/minor/20200826PasqualeAfrica
@@ -1,0 +1,3 @@
+Updated: GridTools::transform() now works with simplex meshes.
+<br>
+(Pasquale Claudio Africa, 2020/08/26)
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -3198,7 +3198,7 @@ namespace GridTools
       cell = triangulation.begin_active(),
       endc = triangulation.end();
     for (; cell != endc; ++cell)
-      for (const unsigned int v : GeometryInfo<dim>::vertex_indices())
+      for (const unsigned int v : cell->vertex_indices())
         if (treated_vertices[cell->vertex_index(v)] == false)
           {
             // transform this vertex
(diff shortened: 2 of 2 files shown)


peterrum (approved):
Thanks a lot!

/rebuild

tjhei:
> working with non-refined simplex meshes.

What happens when the mesh is does not satisfy that? Would it be possible to add an ``Assert`` at the beginning of this function to check this and provide the user with an error message?

doc/news/changes/minor/20200826PasqualeAfrica
```diff
@@ -0,0 +1,3 @@
+Updated: GridTools::transform now works with simplex meshes.
```
luca-heltai, line 1:
```suggestion
Updated: GridTools::transform() now works with simplex meshes.
```

elauksap: Fixed, thank you!

elauksap:
> 
> 
> > working with non-refined simplex meshes.
> 
> What happens when the mesh is does not satisfy that? Would it be possible to add an `Assert` at the beginning of this function to check this and provide the user with an error message?

Thanks for your suggestion!

In principle the function modified by this PR could even work, depending on the ordering and numbering convention chosen for creating children cells. In my opinion such an assertion would better be added in the `Triangulation::execute_coarsening()` and `Triangulation::execute_refinement()` methods, because they are those contain code that crashes on a Simplex triangulation.

@peterrum what do you think?

peterrum:
> In my opinion such an assertion would better be added in the Triangulation::execute_coarsening() and Triangulation::execute_refinement() methods, because they are those contain code that crashes on a Simplex triangulation.

For this there is an open PR. See PR #10439 (part of that is outdated).

Nevertheless, I agree with @tjhei that an assert would be good (although I must admit that although inserting some assert - I probably have missed a lot of them). You could add an assert like this:

```
Assert(reference_cell_type == ReferenceCell::get_hypercube(dim), ExcNotImplemented());
```

If we mark already parts of the code which hard coded for QUAD/HEX this will enable us later to find these sections and to generalize them.

elauksap:
@tjhei @peterrum I've added a couple of assertions, do you think it is ok?

tjhei:
> @tjhei @peterrum I've added a couple of assertions, do you think it is ok?

Looks great, thank you! This is very helpful for the next person who expects the function to work. It is good to be defensive when programming.

peterrum (approved):

PR #10847, opened 2020-08-26
Try to fix docker.yml.

luca-heltai:
I'd merge right away, since this does not change any code/documentation (only github actions, triggered on merge on master, and I have no way to test this except with multiple PR/merges).


peterrum (approved):
Let's give it a try!

PR #10852, opened 2020-08-27
Support to HDF5 export of data on simplex meshes

elauksap:
I've implemented a few modifications to `DataOut` that enable exporting FE fields on simplex meshes to HDF5 file format.

I've tested it locally in serial, parallel and both on scalar and vector problems.
If you approve the changes I can provide an automatic test 😉

@peterrum @luca-heltai

### DIFF
--- a/include/deal.II/base/data_out_base.h
+++ b/include/deal.II/base/data_out_base.h
@@ -1496,12 +1496,9 @@ namespace DataOutBase
     unsigned int node_dim;
 
     /**
-     * The number of vertices per cell. Equal to
-     * GeometryInfo<node_dim>::vertices_per_cell. We need to store
-     * it as a run-time variable here because the dimension
-     * node_dim is also a run-time variable.
+     * The number of cells stored in @ref filtered_cells.
      */
-    unsigned int vertices_per_cell;
+    unsigned int num_cells;
 
     /**
      * Map of points to an internal index.
@@ -3331,9 +3328,20 @@ public:
   /**
    * Get the XDMF content associated with this entry.
    * If the entry is not valid, this returns an empty string.
+   *
+   * @deprecated Use @ref get_xdmf_content(const unsigned int, const ReferenceCell::Type &) instead.
    */
+  DEAL_II_DEPRECATED
   std::string
   get_xdmf_content(const unsigned int indent_level) const;
+
+  /**
+   * Get the XDMF content associated with this entry.
+   * If the entry is not valid, this returns an empty string.
+   */
+  std::string
+  get_xdmf_content(const unsigned int         indent_level,
+                   const ReferenceCell::Type &reference_cell_type) const;
 
 private:
   /**
--- a/source/base/data_out_base.cc
+++ b/source/base/data_out_base.cc
@@ -315,7 +315,7 @@ namespace DataOutBase
   DataOutFilter::DataOutFilter()
     : flags(false, true)
     , node_dim(numbers::invalid_unsigned_int)
-    , vertices_per_cell(numbers::invalid_unsigned_int)
+    , num_cells(numbers::invalid_unsigned_int)
   {}
 
 
@@ -323,7 +323,7 @@ namespace DataOutBase
   DataOutFilter::DataOutFilter(const DataOutBase::DataOutFilterFlags &flags)
     : flags(flags)
     , node_dim(numbers::invalid_unsigned_int)
-    , vertices_per_cell(numbers::invalid_unsigned_int)
+    , num_cells(numbers::invalid_unsigned_int)
   {}
 
 
@@ -363,6 +363,10 @@ namespace DataOutBase
                                    const unsigned int pt_index)
   {
     filtered_cells[cell_index] = filtered_points[pt_index];
+
+    // (Re)-initialize counter at any first call to this method.
+    if (cell_index == 0)
+      num_cells = 1;
   }
 
 
@@ -432,7 +436,7 @@ namespace DataOutBase
   unsigned int
   DataOutFilter::n_cells() const
   {
-    return filtered_cells.size() / vertices_per_cell;
+    return num_cells;
   }
 
 
@@ -465,9 +469,11 @@ namespace DataOutBase
                             const unsigned int d2,
                             const unsigned int d3)
   {
+    ++num_cells;
+
     const unsigned int base_entry =
       index * GeometryInfo<dim>::vertices_per_cell;
-    vertices_per_cell = GeometryInfo<dim>::vertices_per_cell;
+
     internal_add_cell(base_entry + 0, start);
     if (dim >= 1)
       {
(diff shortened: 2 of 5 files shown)


include/deal.II/base/data_out_base.h
```diff
@@ -1431,6 +1431,12 @@ namespace DataOutBase
     unsigned int
     n_cells() const;
 
+    /**
+     * Return the number of vertices per cell in this DataOutFilter.
+     */
+    unsigned int
+    n_vertices_per_cell() const;
```
peterrum, line 1438: I wouldn't add this function - because it will work now but not for mixed meshes. Cannot you use the `Patch::reference_cell_type`?

https://github.com/dealii/dealii/blob/b22a991ea61976e141de37eca96f13e5a2a9f372/include/deal.II/base/data_out_base.h#L327

include/deal.II/base/data_out_base.h
```diff
@@ -1496,8 +1502,7 @@ namespace DataOutBase
     unsigned int node_dim;
 
     /**
-     * The number of vertices per cell. Equal to
-     * GeometryInfo<node_dim>::vertices_per_cell. We need to store
+     * The number of vertices per cell. We need to store
```
peterrum, line 1505: I guess we should remove this variable and determine its value on-the-fly.

include/deal.II/base/data_out_base.h
```diff
@@ -3333,7 +3338,8 @@ class XDMFEntry
    * If the entry is not valid, this returns an empty string.
    */
   std::string
-  get_xdmf_content(const unsigned int indent_level) const;
+  get_xdmf_content(const unsigned int         indent_level,
+                   const ReferenceCell::Type &reference_cell_type) const;
```
peterrum, line 3342: This looks like a public function. I guess we have to keep the original function which takes a single parameter (we can deprecate it) and add one which takes two.

source/base/data_out_base.cc
```diff
@@ -8518,17 +8546,43 @@ XDMFEntry::get_xdmf_content(const unsigned int indent_level) const
            << "\" NumberOfElements=\"" << num_cells
            << "\" NodesPerElement=\"2\">\n";
       else if (dimension == 2)
-        ss << indent(indent_level + 1) << "<Topology TopologyType=\""
-           << "Quadrilateral"
-           << "\" NumberOfElements=\"" << num_cells << "\">\n";
+        {
+          ss << indent(indent_level + 1) << "<Topology TopologyType=\"";
+          if (reference_cell_type == ReferenceCell::get_hypercube(dimension))
```
peterrum, line 8551:
```suggestion
          if (reference_cell_type == ReferenceCell::Type::Quad)
```

source/base/data_out_base.cc
```diff
@@ -8518,17 +8546,43 @@ XDMFEntry::get_xdmf_content(const unsigned int indent_level) const
...
+            {
+              ss << "Quadrilateral"
+                 << "\" NumberOfElements=\"" << num_cells << "\">\n"
+                 << indent(indent_level + 2) << "<DataItem Dimensions=\""
+                 << num_cells << " " << (1 << dimension);
+            }
+          else
```
peterrum, line 8558:
```suggestion
          else if (reference_cell_type == ReferenceCell::Type::Tri)
```

source/base/data_out_base.cc
```diff
@@ -8518,17 +8546,43 @@ XDMFEntry::get_xdmf_content(const unsigned int indent_level) const
...
+            {
+              ss << "Triangle"
+                 << "\" NumberOfElements=\"" << num_cells << "\">\n"
+                 << indent(indent_level + 2) << "<DataItem Dimensions=\""
+                 << num_cells << " " << 3;
+            }
```
peterrum, line 8564: else: assert

source/base/data_out_base.cc
```diff
@@ -8518,17 +8546,43 @@ XDMFEntry::get_xdmf_content(const unsigned int indent_level) const
...
+        }
       else if (dimension == 3)
-        ss << indent(indent_level + 1) << "<Topology TopologyType=\""
-           << "Hexahedron"
-           << "\" NumberOfElements=\"" << num_cells << "\">\n";
+        {
+          ss << indent(indent_level + 1) << "<Topology TopologyType=\"";
+          if (reference_cell_type == ReferenceCell::get_hypercube(dimension))
```
peterrum, line 8569: the same here

peterrum (commented):
Looks good to me. I have added some comments. Thanks a lot for the work!

Regarding the test: you could take https://github.com/dealii/dealii/blob/master/tests/simplex/data_out_write_vtk_01.cc as basis.

include/deal.II/base/data_out_base.h
elauksap: I'm not sure how we can get the number of vertices inside the `n_cells()` method in that case! Any suggestion?

elauksap:
Updated the code according to @peterrum suggestions and added a test!

include/deal.II/base/data_out_base.h
peterrum: This function looks odd^^ Let's keep it as it is right now and think about mixed meshes later. Could you add an assert somewhere that all cells have the same type?

luca-heltai:
There seems to be an unused variable somewhere:

~~~

[ 80%] Building CXX object source/grid/CMakeFiles/obj_grid_debug.dir/cell_id.cc.o
/Users/runner/work/dealii/dealii/source/base/data_out_base.cc:7552:44: error: unused parameter 'patches' [-Werror,-Wunused-parameter]
  const std::vector<Patch<dim, spacedim>> &patches,
                                           ^
1 error generated.
~~~

include/deal.II/base/data_out_base.h
elauksap: Done! ✔️

elauksap:
@luca-heltai it should be solved now!

elauksap:
@peterrum I think I found out a way not to store `vertices_per_cell` as a member of `DataOutFilter`.
We can simply count the number of cells by counting the number of calls to `DataOutFilter::write_cell()` or to `DataOutFilter::write_cell_single()` (see my latest commit), right?

source/base/data_out_base.cc
```diff
@@ -315,15 +315,15 @@ namespace DataOutBase
   DataOutFilter::DataOutFilter()
     : flags(false, true)
     , node_dim(numbers::invalid_unsigned_int)
-    , vertices_per_cell(numbers::invalid_unsigned_int)
+    , num_cells(0)
```
peterrum, line 318:
```suggestion
    , num_cells(numbers::invalid_unsigned_int)
```

source/base/data_out_base.cc
```diff
@@ -315,15 +315,15 @@ namespace DataOutBase
...
   {}
 
 
 
   DataOutFilter::DataOutFilter(const DataOutBase::DataOutFilterFlags &flags)
     : flags(flags)
     , node_dim(numbers::invalid_unsigned_int)
-    , vertices_per_cell(numbers::invalid_unsigned_int)
+    , num_cells(0)
```
peterrum, line 326: Same here.

peterrum (approved):
The changes look good to me. 

I would like to ask someone else to have a look at the files too and to merge. I haven't yet used these classes...

peterrum:
@elauksap Could you resolve the merge conflict.

I'll merge tonight if there are no objections.

elauksap:
@peterrum I think it's done!

PR #10853, opened 2020-08-27
Use contiguous arrays for some polynomial classes.

drwells:
This is nicer than `std::vector<std::vector<std::vector<double>>>`.

For some reason this class is now giving me a bunch of compiler warnings in the dim = 0 case - I fixed this while investigating.


kronbichler (approved):

kronbichler:
Nice!

PR #10854, opened 2020-08-28
Fix some type limit warnings in TensorProductPolynomials.

drwells:
GCC 10.2.0 complains when we instantiate with dim = 0 since a lot of loops then have checks 0 < 0. Get around this by specializing templates in this case.

### DIFF
--- a/include/deal.II/base/tensor_product_polynomials.h
+++ b/include/deal.II/base/tensor_product_polynomials.h
@@ -659,6 +659,20 @@ TensorProductPolynomials<dim, PolynomialType>::compute_derivative(
 
 
 
+template <>
+template <int order>
+Tensor<order, 0>
+TensorProductPolynomials<0, Polynomials::Polynomial<double>>::
+  compute_derivative(const unsigned int, const Point<0> &) const
+{
+  constexpr int dim = 0;
+  AssertThrow(dim > 0, ExcNotImplemented());
+
+  return {};
+}
+
+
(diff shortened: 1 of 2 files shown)


drwells:
We do need them for some face values - for example in `fe_abf.cc` we have
```cpp
  if (dim > 1)
    {
      QGauss<dim - 1>                   face_points(deg + 1);
      TensorProductPolynomials<dim - 1> legendre =
        Polynomials::Legendre::generate_complete_basis(deg);

      boundary_weights.reinit(n_face_points, legendre.n());
```

drwells:
>  see that instantiations only happen for dim=1,2,3 in source/base/tensor_product_polynomials.cc

Reading this again I'm not completely sure I understand your question - I see instantiations for `TensorProductPolynomials<0, Polynomials::Polynomial<double>>` and `AnisotropicPolynomials<0>` at the bottom of `tensor_product_polynomials.cc` - are we looking at the same thing?

tjhei:
> are we looking at the same thing?

I stand corrected (but at least the one with PiecewisePolynomial is not instantiated for 0).

tjhei:
It seems a bit silly to instantiate them for dim=0 even though none of the functions work, but you are not the person who did this. :-)

I guess this is good to merge.

drwells:
They don't work but we do need them to exist to avoid link errors, so I am content with the status quo.

PR #10859, opened 2020-08-28
CUDA: Compute material property on the host

Rombur:
This PR add a few functions that allow to loop over quadrature points on the host in the same "order" than we loop over them on the device. This is useful when you have complicated material properties and for debugging.


include/deal.II/matrix_free/cuda_matrix_free.h
```diff
@@ -616,6 +624,11 @@ namespace CUDAWrappers
      */
     const DoFHandler<dim> *dof_handler;
 
+    /**
+     * Colored graphed of locally owned active cells.
```
masterleinad, line 628:
```suggestion
     * Colored graph of locally owned active cells.
```

peterrum:
I'll take a look later today!

peterrum (approved):
Looks good to me.

A general remark, wouldn't it be more efficient to copy all points to the host in one go? We could store the information on the host as well?

Rombur:
I pick this interface because it was closer to the CUDA code. What do you propose? One function to get all the quadrature points in a single call and another function to get the quadrature point from the array?

peterrum:
> I pick this interface because it was closer to the CUDA code. What do you propose? One function to get all the quadrature points in a single call and another function to get the quadrature point from the array?

I was more thinking about that certain useful information could be always duplicated on host and device (with certain sync functions - if the mesh has moved). But this more a general design decision and is off-topic; I am fine to merge your PR as it is.

Rombur:
Changing this WIP. I am reworking this PR.

Rombur:
superseeded by https://github.com/dealii/dealii/pull/11335

PR #10860, opened 2020-08-29
Wrap some more calls to delete with unique_ptr.

drwells:
Not the most important feature, but we are actually quite close to getting rid of all new/delete pairs.

### DIFF
--- a/source/fe/fe_abf.cc
+++ b/source/fe/fe_abf.cc
@@ -177,7 +177,7 @@ FE_ABF<dim>::initialize_support_points(const unsigned int deg)
 
   // These might be required when the faces contribution is computed
   // Therefore they will be initialized at this point.
-  std::vector<AnisotropicPolynomials<dim> *> polynomials_abf(dim);
+  std::array<std::unique_ptr<AnisotropicPolynomials<dim>>, dim> polynomials_abf;
 
   // Generate x_1^{i} x_2^{r+1} ...
   for (unsigned int dd = 0; dd < dim; ++dd)
@@ -187,7 +187,7 @@ FE_ABF<dim>::initialize_support_points(const unsigned int deg)
         poly[d].push_back(Polynomials::Monomial<double>(deg + 1));
       poly[dd] = Polynomials::Monomial<double>::generate_complete_basis(deg);
 
-      polynomials_abf[dd] = new AnisotropicPolynomials<dim>(poly);
+      polynomials_abf[dd] = std::make_unique<AnisotropicPolynomials<dim>>(poly);
     }
 
   // Number of the point being entered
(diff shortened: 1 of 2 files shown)


drwells:
The CI fails when I don't use `std::make_unique` - I guess I am still used to thinking of C++11 :)

source/fe/fe_abf.cc
```diff
@@ -187,7 +187,7 @@ FE_ABF<dim>::initialize_support_points(const unsigned int deg)
         poly[d].push_back(Polynomials::Monomial<double>(deg + 1));
       poly[dd] = Polynomials::Monomial<double>::generate_complete_basis(deg);
 
-      polynomials_abf[dd] = new AnisotropicPolynomials<dim>(poly);
+      polynomials_abf[dd].reset(new AnisotropicPolynomials<dim>(poly));
```
bangerth, line 190: The usual idiom would be
```suggestion
      polynomials_abf[dd] = std::make_unique<AnisotropicPolynomials<dim>>(poly);
```
so that you can also avoid the explicit call to `new`.

source/fe/fe_abf.cc
```diff
@@ -262,7 +262,7 @@ FE_ABF<dim>::initialize_support_points(const unsigned int deg)
             poly[d] = Polynomials::Legendre::generate_complete_basis(deg);
           poly[dd] = Polynomials::Legendre::generate_complete_basis(deg - 1);
 
-          polynomials[dd] = new AnisotropicPolynomials<dim>(poly);
+          polynomials[dd].reset(new AnisotropicPolynomials<dim>(poly));
```
bangerth, line 265: same here

source/fe/fe_abf.cc
```diff
@@ -421,15 +415,15 @@ FE_ABF<dim>::initialize_restriction()
...
+  std::array<std::unique_ptr<AnisotropicPolynomials<dim>>, dim> polynomials;
   for (unsigned int dd = 0; dd < dim; ++dd)
     {
       std::vector<std::vector<Polynomials::Polynomial<double>>> poly(dim);
       for (unsigned int d = 0; d < dim; ++d)
         poly[d] = Polynomials::Legendre::generate_complete_basis(rt_order);
       poly[dd] = Polynomials::Legendre::generate_complete_basis(rt_order - 1);
 
-      polynomials[dd] = new AnisotropicPolynomials<dim>(poly);
+      polynomials[dd].reset(new AnisotropicPolynomials<dim>(poly));
```
bangerth, line 426: and here

drwells:
Fixed!

peterrum (approved):

peterrum:
Something is wrong with CI. Could you rebase and push again!?

PR #10861, opened 2020-08-31
Attempt to work around a compiler warning.

bangerth:
#10786 shows a problem with a compiler warning (unwarranted). But it would be nice to do what #10786 wants to do, so let's try to work around this issue.

/rebuild


masterleinad:
Do you know if that actually helps with #10786? Otherwise, it would be a good idea to just push this commit to that branch to see.

tjhei:
I cherry-picked it and pushed to the other branch...

masterleinad:
> I cherry-picked it and pushed to the other branch...

Seems to work.

tjhei:
awesome, thanks @bangerth !

bangerth:
OK, thanks for checking!

PR #10864, opened 2020-08-31
Get GridOut::write_eps working with triangles.

drwells:
There are some additional fixes in here that are necessary for VTU but that is not yet finished.


peterrum (approved):
Looks good to me! As far as I can see, it was mainly removing `GeometryInfo` :) Thanks, @drwells!

/rebuild

PR #10865, opened 2020-08-31
Enable BlockVector for FEEvaluation::gather_evalute()

peterrum:
@kronbichler What do you think about this solution?

FYI @mschreter

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -26,6 +26,7 @@
 #include <deal.II/matrix_free/evaluation_flags.h>
 #include <deal.II/matrix_free/shape_info.h>
 #include <deal.II/matrix_free/tensor_product_kernels.h>
+#include <deal.II/matrix_free/type_traits.h>
 
 
 DEAL_II_NAMESPACE_OPEN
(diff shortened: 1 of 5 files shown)


mschreter:
Thank you Peter. The same issue also arises with FEEvaluation::integrate_scatter().

peterrum:
> Thank you Peter. The same issue arises also with FEEvaluation::integrate_scatter().

Right. Let's wait for a review on this function and then I will the same concept for `integrate_scatter()`.

kronbichler (commented):
This looks good to me. Go ahead with the integration. Do we have to adjust things for `FEFaceEvaluation` as well or does it catch that already by other means (I don't recall the last state and have not time to check now)?

peterrum:
> This looks good to me. Go ahead with the integration. Do we have to adjust things for FEFaceEvaluation as well or does it catch that already by other means (I don't recall the last state and have not time to check now)?

Puh. This was more work than expected^^ But, now there is also a test :)

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2165,10 +2204,126 @@ namespace internal
...
+      const bool           integrate_gradients,
+      const unsigned int,
+      const unsigned int,
+      const std::array<unsigned int, n_face_orientations>,
+      const std::array<unsigned int, n_face_orientations> face_nos,
+      const unsigned int                                  subface_index,
+      const MatrixFreeFunctions::DoFInfo::DoFAccessIndex,
+      const std::array<unsigned int, n_face_orientations> face_orientations,
+      const Table<2, unsigned int> &                      orientation_map)
+    {
```
kronbichler, line 2233: Can you explain why you need to add the code inside this function? At first sight I do not see the difference to `FEFaceEvaluationSelector::integrate` (well, except some arguments). At least, this would need some comments on why things get through another path than in the `evaluate` routine where we can bypass the operation completely.

tests/matrix_free/matrix_vector_faces_29.cc
```diff
@@ -0,0 +1,148 @@
...
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+
+// similar to matrix_vector_faces_13 (renumbering of degrees of freedom to
+// better vectorized access, gather_evaluate and integrate_scatter) but on a
+// different mesh that can trigger different element types
```
kronbichler, line 20: Do you want to also add that this uses a `BlockVector`. I further assume that this text came from another test that should be mentioned here as well.

include/deal.II/matrix_free/evaluation_kernels.h
peterrum: I guess I could call `integrate` in this function. The general question is where we want check if a vector class has the right `begin()`-method. I have chosen to specialize for `FEFaceEvaluationSelector::integrate_scatter()` because I found it more easy in this case; for `FEEvaluation::integrate_scatter` it is done differently...

peterrum: I have updated this PR. How do you like it this way?

peterrum: > Excellent!

This PR has lead to some annoying incompatibilities in one of our application codes where we precompile (in one way or the other) these intenal kernels. To add additional template arguments here seems not to be a good idea. We should probably move the check to `FEFaceEvaluation`.

FYI @mschreter

kronbichler: I agree, but I would append that we need frequent changes in code using these internal interfaces during these days anyway.

Issue #7039, opened 2018-08-07
Revise names in internal::EvaluatorTensorProduct


kronbichler:
The tensor product evaluators are still left with names like `values`, `gradients`, `hessians`, which is somewhat of a misnomer now that we have given them a pure linear algebra interface. This should be changed at some point. Luckily, it is in the internal namespace so we do not have to be very careful regarding backward compatibility.

kronbichler:
We're not going to address this for the current release. Adjusting target milestone.

kronbichler:
This could be done in combination with #9544.

kronbichler:
As suggested in https://github.com/dealii/dealii/pull/11034#pullrequestreview-509051438 we should also think about an appropriate folder for this functionality once we have it generalized. Either `lac` or `base` would come to mind

peterrum:
Postponing after the release.

peterrum:
@kronbichler We could keep this issue in mind when adding new `EvaluatorTensorProduct` versions!

PR #10869, opened 2020-09-01
Add FEEvaluationImplBasisChange::do_backward_hessians

peterrum:
Extracted from #10830. Needed for applying the inverse matrix in the ECL context.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -648,10 +648,11 @@ namespace internal
    * This class allows for dimension-independent application of the operation,
    * implemented by template recursion. It has been tested up to 6D.
    */
-  template <EvaluatorVariant variant,
-            int              dim,
-            int              basis_size_1,
-            int              basis_size_2,
+  template <EvaluatorVariant  variant,
+            EvaluatorQuantity quantity,
+            int               dim,
+            int               basis_size_1,
+            int               basis_size_2,
             typename Number,
             typename Number2>
   struct FEEvaluationImplBasisChange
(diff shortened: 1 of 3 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -891,6 +891,107 @@ namespace internal
...
+    do_backward_hessians(
+      const unsigned int                    n_components,
+      const dealii::AlignedVector<Number2> &transformation_matrix,
+      const bool                            add_into_result,
+      Number *                              values_in,
+      Number *                              values_out,
+      const unsigned int                    basis_size_1_variable =
+        dealii::numbers::invalid_unsigned_int,
+      const unsigned int basis_size_2_variable =
+        dealii::numbers::invalid_unsigned_int)
```
kronbichler, line 910: I must admit that I do not like to duplicate the `do_backward` function for the sake of bypassing the assumption of a zero somewhere in the `values` call in that function. We should make the interfaces uniform by just providing an `apply` function in the tensor product evaluators with an `enum` informing us about special cases, see also #7039.

peterrum: I agree that we should try not to introduce an additional specialization for hessian. But let's not remove the `values`/`gradients`/`hessians`-function calls now from all places (I can put it on my TODO list). I have introduced as suggested an enum (name is up to discussion) and use this now in the `FEEvaluationImplBasisChange` class. What do you think, @kronbichler?

kronbichler: Absolutely, there is no need to clean things up now, neither need we remove the values/gradients/hessian calls now. But I was thinking of providing all evaluators with a `static void apply` function and uniform template parameters along the variants. Or, even better, there is a single code to run through the tensor products and just different 1D matrix-vector products.

kronbichler (approved):
I guess this is fine now; it does not immediately help now but it will once we have clearer interfaces elsewhere.

PR #10870, opened 2020-09-01
Work on interpolation from cell to face quadrature points

peterrum:
Extracted from #10830. In the ECL context, values are available at cell quadrature points. The face quadrature points can be simply computed via a 1D interpolation as proposed here.


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -1735,6 +1735,67 @@ namespace internal
...
+  {
+    template <bool do_evaluate, bool add_into_output>
+    static void
+    interpolate(const internal::MatrixFreeFunctions::ShapeInfo<Number> &data,
+                const Number *                                          input,
+                Number *                                                output,
+                const bool         do_gradients,
+                const unsigned int face_no)
+    {
+      AssertDimension(n_points_1d, data.data.front().quadrature.size());
```
kronbichler, line 1753: Since this function is almost the same as the other `FEFaceNormalEvaluationImpl`, I wonder whether we can unify the calls and let both function call an internal function `interpolate_generic`. What I'm thinking about is to let `interpolate` and `interpolate_quadrature` extract the data array (`shape_data_on_face[face_no % 2]` or `quadrature_data_on_face[face_no % 2]`) as well as the `in_stride` and `out_stride` and let the `interpolate_generic` (or whatever name we choose) do the job.

include/deal.II/matrix_free/shape_info.h
```diff
@@ -229,6 +229,18 @@ namespace internal
        */
       AlignedVector<Number> shape_data_on_face[2];
 
+      /**
+       * Collects all data of 1D nodal shape values (that defined by the
+       * support of the quadrature rule) evaluated at the point 0 and 1
```
kronbichler, line 234:
```suggestion
       * Collects all data of 1D nodal shape values (defined by the Lagrange
       * polynomials in the points of the quadrature rule) evaluated at the point 0 and 1
```

include/deal.II/matrix_free/evaluation_kernels.h
peterrum: Good idea. What I was thinking about as a follow-up PR is to make the function dimension independent since I want to use it in 6D. What do you think?

include/deal.II/matrix_free/shape_info.templates.h
```diff
@@ -298,6 +300,20 @@ namespace internal
...
+      dealii::FE_DGQArbitraryNodes<1> fe_quad(quad);
+
+      for (unsigned int i = 0; i < quad.size(); ++i)
+        {
+          Point<1> q_point;
+          q_point[0]                    = 0;
+          quadrature_data_on_face[0][i] = fe_quad.shape_value(i, q_point);
+          q_point[0]                    = 1;
+          quadrature_data_on_face[1][i] = fe_quad.shape_value(i, q_point);
+        }
```
kronbichler, line 315: I think you need to encapsulate this part in some `if` for the case the number of quadrature points is very high, say a few hundreds. In that case, the `FE_DGQArbitraryNodes` polynomials would underflow and result in a division by zero. The collocation is not really helpful in that case anyway.

kronbichler (commented):
A few more comments.

include/deal.II/matrix_free/evaluation_kernels.h
kronbichler: Sounds good.

peterrum:
Done :)

kronbichler (approved):
Excellent.

PR #10875, opened 2020-09-01
Add more python wrappers for GridTools

agrayver:
- Replace existing ```merge_triangulations``` wrapper that accepts only two triangulations with a more generic equivalent.
- Add python wrapper for GridTools::replicate_triangulation

Part of #9015

### DIFF
--- a/contrib/python-bindings/source/triangulation_wrapper.cc
+++ b/contrib/python-bindings/source/triangulation_wrapper.cc
@@ -515,7 +515,9 @@ namespace python
     template <int dim, int spacedim>
     void
     merge_triangulations(boost::python::list &triangulations,
-                         void *               triangulation)
+                         void *               triangulation,
+                         const double         duplicated_vertex_tolerance,
+                         const bool           copy_manifold_ids)
     {
       Triangulation<dim, spacedim> *tria =
         static_cast<Triangulation<dim, spacedim> *>(triangulation);
@@ -531,7 +533,10 @@ namespace python
             tria_wrapper.get_triangulation()));
         }
 
-      GridGenerator::merge_triangulations(tria_ptrs, *tria);
+      GridGenerator::merge_triangulations(tria_ptrs,
+                                          *tria,
+                                          duplicated_vertex_tolerance,
+                                          copy_manifold_ids);
     }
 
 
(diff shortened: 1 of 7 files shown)


contrib/python-bindings/source/triangulation_wrapper.cc
```diff
@@ -514,23 +514,24 @@ namespace python
...
-      Triangulation<dim, spacedim> *tria_1 =
-        static_cast<Triangulation<dim, spacedim> *>(
-          triangulation_1.get_triangulation());
-      Triangulation<dim, spacedim> *tria_2 =
-        static_cast<Triangulation<dim, spacedim> *>(
-          triangulation_2.get_triangulation());
-      GridGenerator::merge_triangulations(*tria_1, *tria_2, *tria);
-      // We need to reassign tria to triangulation because tria was cleared
-      // inside merge_triangulations.
-      triangulation = tria;
```
Rombur, line 533: So this part is not necessary anymore?

doc/news/changes/minor/20200901AlexanderGrayver
```diff
@@ -0,0 +1,5 @@
+Added: python wrapper for GridTools::replicate_triangulation,
+more general version of the GridTools::merge_triangulations is
+implemented
+<br>
+(Alexander Grayver, 2020/09/01)
```
Rombur, line 5: You technically also introduce an incompatibility in `merge_triangulation` So can you add an entry there.

contrib/python-bindings/source/triangulation_wrapper.cc
agrayver: This is what happens inside

https://github.com/dealii/dealii/blob/e9461deb2e453f353d6bcab7d571aa2d4715ca47/source/grid/grid_generator.cc#L5976

Gigen that we pass the argument by reference, I don't see a need for this. Am I missing something?

agrayver:
Thanks, @Rombur! I added an incompatibility note. Also, I added two optional parameters of the original C++ function to the python wrapper.

Rombur (approved):
This looks good but since the tester is failing can you rebase the PR on master to see if it fixes the problem.

agrayver:
@Rombur looks good now, isn't?

PR #10876, opened 2020-09-02
Reorder computations in inversion of 3x3 tensor

kronbichler:
This PR reorders the computation in the inverse of the 3x3 tensor. I observed that the original code does superfluous operations because of the way it computes the determinant. More precisely, the new computations more aggressively re-uses the first minors (cofactors) of the matrix for the determinant. The old code used 12 FMAs (fused multiply-add), 12 multiplications, and 2 subtractions, whereas the new code uses 11 FMAs and 10 multiplications. [Edit: I do not count the final 9 multiplications because they are the same in both cases.] I would also argue that the new code for the inversion looks a bit better than the old one (partly due to how clang indents).

### DIFF
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -2500,6 +2500,38 @@ constexpr DEAL_II_ALWAYS_INLINE Number
   return t[0][0];
 }
 
+/**
+ * Specialization for dim==2.
+ *
+ * @relatesalso Tensor
+ */
+template <typename Number>
+constexpr DEAL_II_ALWAYS_INLINE Number
+                                determinant(const Tensor<2, 2, Number> &t)
+{
+  // hard-coded for efficiency reasons
+  return t[0][0] * t[1][1] - t[1][0] * t[0][1];
+}
+
+/**
+ * Specialization for dim==3.
+ *
+ * @relatesalso Tensor
+ */
+template <typename Number>
+constexpr DEAL_II_ALWAYS_INLINE Number
+                                determinant(const Tensor<2, 3, Number> &t)
+{
+  // hard-coded for efficiency reasons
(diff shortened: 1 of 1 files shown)


include/deal.II/base/tensor.h
```diff
@@ -2500,6 +2500,38 @@ constexpr DEAL_II_ALWAYS_INLINE Number
...
+  // hard-coded for efficiency reasons
+  const Number C0 = internal::NumberType<Number>::value(t[1][1] * t[2][2]) -
+                    internal::NumberType<Number>::value(t[1][2] * t[2][1]);
+  const Number C1 = internal::NumberType<Number>::value(t[1][2] * t[2][0]) -
+                    internal::NumberType<Number>::value(t[1][0] * t[2][2]);
+  const Number C2 = internal::NumberType<Number>::value(t[1][0] * t[2][1]) -
+                    internal::NumberType<Number>::value(t[1][1] * t[2][0]);
+  return t[0][0] * C0 + t[0][1] * C1 + t[0][2] * C2;
+}
+
```
kronbichler, line 2534: This part of the PR is open to discussion. I observed that the generic implementation is quite bad for the 3D case because it does superfluous multiplication by "-1" and some index comparisons I do not like. I prefer when the determinant (that is often used nearby the inversion) uses the same code as the inverse because then the compiler can optimize one away. But this adds 32 lines of code, so it's not entirely clear whether we want to have it or rather should try to improve the original code a few lines above. In case we should improve the original code, how would we like to do it?

peterrum (approved):
Looks good to me.

bangerth (approved):

bangerth:
The length you go to to reduce 12 to 11 fmas continues to amaze me -- but we all benefit from it! :-)

PR #10877, opened 2020-09-02
Fix statement in documentation

peterrum:


drwells:
It's just documentation so lets merge once GitHub's indent finishes (that rebuilds the docs).

PR #10879, opened 2020-09-02
Use a virtual destructor with virtual methods.

drwells:
This fixes a new warning.

### DIFF
--- a/include/deal.II/base/memory_space.h
+++ b/include/deal.II/base/memory_space.h
@@ -117,6 +117,13 @@ namespace MemorySpace
       : values(nullptr, &std::free)
     {}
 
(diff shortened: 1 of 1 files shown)


peterrum (approved):

drwells:
Those are some interesting failures. I'm investigating.

peterrum:
A general remark. Maybe we could enable some compiler warnings in the the GitHub tests so that we already catch such issues already during the PR. I did that for braces a month ago...

tjhei:
what is the precise warning?

drwells:
I have `-Wnon-virtual-dtor` enabled.

Issue #10882, opened 2020-09-03
Make DoFTools::extract_dofs() faster


bangerth:
As reported on the mailing list, the version of `DoFTools::extract_dofs()` that returns an `IndexSet` is pretty slow. I think that's because we have this insertion of indices one-by-one:
```
    // Then loop over all cells and do the work
    IndexSet                             selected_dofs(dof.n_dofs());
    std::vector<types::global_dof_index> local_dof_indices;
    for (const auto &c : dof.active_cell_iterators())
      if (c->is_locally_owned())
        {
          const unsigned int fe_index      = c->active_fe_index();
          const unsigned int dofs_per_cell = c->get_fe().n_dofs_per_cell();
          local_dof_indices.resize(dofs_per_cell);
          c->get_dof_indices(local_dof_indices);
          for (unsigned int i = 0; i < dofs_per_cell; ++i)
            if (dof.locally_owned_dofs().is_element(local_dof_indices[i]) &&
                local_selected_dofs[fe_index][i])
              selected_dofs.add_index(local_dof_indices[i]);
        }
```
Elementwise insertion into `IndexSet` is slow. What we should be doing is to create a `std::set` first, then bulk-convert that into an `IndexSet` -- that's what we do in other places.

kronbichler:
Fixed by #10883.

PR #10883, opened 2020-09-03
Speed up DoFTools::extract_dofs

kronbichler:
As reported on the mainling list, the function got really slow. It turned out that the function called IndexSet::add_index() for every index, which leads to frequent memory allocations and movements of indices that can lead to a quadratic complexity in the number of indices. This function was introduced in #9182. Luckily, we have a good implementation of that part almost there by an internal function, so we only need to go there.

### DIFF
--- a/source/dofs/dof_tools.cc
+++ b/source/dofs/dof_tools.cc
@@ -456,47 +456,22 @@ namespace DoFTools
              dof.get_fe_collection().n_components())
       return dof.locally_owned_dofs();
 
-    // If we have to deal with a partial component mask, then
-    // set up a table for the degrees of freedom on the reference cell (for
-    // each element in the collection) to see whether it belongs to one of the
-    // desired components.
-    const auto &                   fe_collection = dof.get_fe_collection();
-    std::vector<std::vector<bool>> local_selected_dofs;
-    Assert(fe_collection.n_components() < 256, ExcNotImplemented());
-    for (unsigned int f = 0; f < fe_collection.size(); ++f)
-      {
-        // First get the component for each shape function
-        const std::vector<unsigned char> local_component_association =
-          internal::get_local_component_association(fe_collection[f],
-                                                    component_mask);
+    // get the component association of each DoF and then select the ones
+    // that match the given set of components
+    std::vector<unsigned char> dofs_by_component(dof.n_locally_owned_dofs());
+    internal::get_component_association(dof, component_mask, dofs_by_component);
 
-        // Check which dofs were selected
-        std::vector<bool> this_selected_dofs(
-          fe_collection[f].n_dofs_per_cell());
-        for (unsigned int i = 0; i < fe_collection[f].n_dofs_per_cell(); ++i)
-          this_selected_dofs[i] =
-            component_mask[local_component_association[i]];
+    // fill the selected components in a vector
+    std::vector<types::global_dof_index> selected_dofs;
+    selected_dofs.reserve(dof.n_locally_owned_dofs());
+    for (types::global_dof_index i = 0; i < dofs_by_component.size(); ++i)
+      if (component_mask[dofs_by_component[i]] == true)
+        selected_dofs.push_back(dof.locally_owned_dofs().nth_index_in_set(i));
 
-        local_selected_dofs.emplace_back(std::move(this_selected_dofs));
-      }
-
-    // Then loop over all cells and do the work
-    IndexSet                             selected_dofs(dof.n_dofs());
-    std::vector<types::global_dof_index> local_dof_indices;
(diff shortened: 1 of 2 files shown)


source/dofs/dof_tools.cc
```diff
@@ -456,47 +456,22 @@ namespace DoFTools
...
-        for (unsigned int i = 0; i < fe_collection[f].n_dofs_per_cell(); ++i)
-          this_selected_dofs[i] =
-            component_mask[local_component_association[i]];
-
-        local_selected_dofs.emplace_back(std::move(this_selected_dofs));
-      }
+    // get the component association of each DoF and then select the ones
+    // that match the given set of components
+    std::vector<unsigned char> dofs_by_component(dof.n_locally_owned_dofs());
+    internal::get_component_association(dof, component_mask, dofs_by_component);
```
bangerth, line 462: This function has to translate local to global indices in the index set.

source/dofs/dof_tools.cc
```diff
@@ -456,47 +456,22 @@ namespace DoFTools
...
-            if (dof.locally_owned_dofs().is_element(local_dof_indices[i]) &&
-                local_selected_dofs[fe_index][i])
-              selected_dofs.add_index(local_dof_indices[i]);
-        }
+    // fill the selected components in a vector
+    std::vector<types::global_dof_index> selected_dofs;
+    selected_dofs.reserve(dof.n_locally_owned_dofs());
+    for (types::global_dof_index i = 0; i < dofs_by_component.size(); ++i)
+      if (component_mask[dofs_by_component[i]] == true)
+        selected_dofs.push_back(dof.locally_owned_dofs().nth_index_in_set(i));
```
bangerth, line 469: and here is the second time.

bangerth (commented):
Why not build a `std::set` first (see suggestion in #10882) so that you don't have to do the local-to-global translation twice?

kronbichler:
> Why not build a std::set first (see suggestion in #10882) so that you don't have to do the local-to-global translation twice?

I am aware of the translations, but my reasoning is that `std::set` is a slow data structure, so I'd rather avoid it. The alternative would have been to fill a vector of indices on all elements and then sort it. But when I sort things I possibly also get super-linear work (even though only a factor `log`). Let me run it through my timer.

kronbichler:
Here is the alternative implementation:
```c++
    const auto &                   fe_collection = dof.get_fe_collection();
    std::vector<std::vector<bool>> local_selected_dofs;
    Assert(fe_collection.n_components() < 256, ExcNotImplemented());
    for (unsigned int f = 0; f < fe_collection.size(); ++f)
      {
        // First get the component for each shape function
        const std::vector<unsigned char> local_component_association =
          internal::get_local_component_association(fe_collection[f],
                                                    component_mask);

        // Check which dofs were selected
        std::vector<bool> this_selected_dofs(
          fe_collection[f].n_dofs_per_cell());
        for (unsigned int i = 0; i < fe_collection[f].n_dofs_per_cell(); ++i)
          this_selected_dofs[i] =
            component_mask[local_component_association[i]];

        local_selected_dofs.emplace_back(std::move(this_selected_dofs));
      }

    // Then loop over all cells and do the work
    std::vector<types::global_dof_index> selected_dofs;
    selected_dofs.reserve(dof.n_locally_owned_dofs());
    std::vector<types::global_dof_index> local_dof_indices;
    for (const auto &c : dof.active_cell_iterators())
      if (c->is_locally_owned())
        {
          const unsigned int fe_index      = c->active_fe_index();
          const unsigned int dofs_per_cell = c->get_fe().n_dofs_per_cell();
          local_dof_indices.resize(dofs_per_cell);
          c->get_dof_indices(local_dof_indices);
          for (unsigned int i = 0; i < dofs_per_cell; ++i)
            if (local_selected_dofs[fe_index][i] &&
                dof.locally_owned_dofs().is_element(local_dof_indices[i]))
              selected_dofs.push_back(local_dof_indices[i]);
        }
    std::sort(selected_dofs.begin(), selected_dofs.end());
    IndexSet result(dof.n_locally_owned_dofs());
    result.add_indices(selected_dofs.begin(),
                       std::unique(selected_dofs.begin(),
                                   selected_dofs.end()));
    return result;


```

kronbichler:
If we really insisted on making things faster we could speed up the `get_component_association` function by hard-coding the case of a single index (not sure it is worth it) or teach the index set a function that combines `is_element()` and `index_within_set` in a single check/call, with `types::invalid_dof_index` returned for those indices not in the set.

bangerth (approved):

bangerth:
Ah, thanks for actually timing this -- not what I thought would come out of this, but good to learn!

PR #10886, opened 2020-09-03
Test for HDF5 now runs in parallel

elauksap:
Add a simplex test for exporting data in HDF5 file format using a `parallel::fullydistributed::Triangulation`.

Automatic testing in GitHub workflows depends on #10887.


source/base/data_out_base.cc
```diff
@@ -819,10 +819,7 @@ namespace
           }
         else
           {
-            Assert(patch.n_subdivisions == 1, ExcNotImplemented());
-            const auto &info = ReferenceCell::internal::Info::get_cell(
-              patch.reference_cell_type);
-            n_nodes += info.n_vertices();
+            n_nodes += patch.data.n_cols();
```
drwells, line 822: I don't understand why the assertion is no longer required - what happens when we have multiple subdivisions and tetrahedra?

elauksap: I've removed it for compatibility with the hex case. In case of multiple subdivisions on tets the value is simply ignored. Do you think it is better to keep the assertion?

peterrum: This is happens in the case of VTK: See https://github.com/dealii/dealii/pull/10701.

peterrum: Is there something equivalent in the case of HDF5?

drwells: I didn't know about #10701. If we already switch to higher-order elements when `n_subdivisions = 1` then we should permit it here. It makes sense to ignore it for HDF5 but I still think we should error out in the general case since none of our output formats actually let us subdivide simplices more than once.

elauksap: I suppose something analogous happens with HDF5, I'm noticing now that the file size changes for different numbers of subdivisions!

elauksap: Edit: my mistake, sorry! The results with `n_subdivisions > 1` were wrong, the connectivity was not properly exported!

The `write_cells()` function can be generalized for quadratic elements (see [here](http://www.xdmf.org/index.php/XDMF_Model_and_Format#Topology)), but I would stick with the assertion for the moment.

peterrum (approved):

drwells:
It looks like we forgot the label.

/rebuild

PR #10889, opened 2020-09-03
Remove VectorType from FEFaceEvaluationSelector

peterrum:
References https://github.com/dealii/dealii/pull/10865#discussion_r483178060. @kronbichler What do you think? Does it work for you?

These are the times when I doubt my job and wish for a modern language...

FYI @mschreter @nfehn


kronbichler (approved):
Nice!

PR #10895, opened 2020-09-05
Replace C-style arrays in MatrixFree

peterrum:


drwells (approved):
Thanks a lot for doing this. Seeing arrray references instead of pointers in function calls makes the code much easier to understand.

tjhei (commented):
Is any of this user-facing?

peterrum:
> Is any of this user-facing and as such not backwards compatible? Looks good otherwise.

Should be all internal. But let's wait for @kronbichler's opionion!

kronbichler (approved):

kronbichler:
None of those should be directly user-facing and even the use of internal (implementation-dependent) data structures is unlikely to directly access or pass around C-style arrays.

PR #10896, opened 2020-09-05
EvaluatorTensorProduct::apply_face for higher dimensions

peterrum:
References https://github.com/hyperdeal/hyperdeal/pull/47 and https://github.com/hyperdeal/hyperdeal/pull/48.

### DIFF
--- a/include/deal.II/matrix_free/tensor_product_kernels.h
+++ b/include/deal.II/matrix_free/tensor_product_kernels.h
@@ -296,6 +296,11 @@ namespace internal
      *             derivatives, 2 second derivates. Note that all the
      *             derivatives access the data in @p shape_values passed to
      *             the constructor of the class
+     * @tparam lex_faces Sets how the evaluation points on the faces should be
+     *                   sorted: lexicographically or right-hand-system number
+     *                   (special treatment of orientation 1 in 3D). Per default
+     *                   right-hand-system number is enabled, which is only
+     *                   working for dimensions up to 3.
      *
      * @param in address of the input data vector
      * @param out address of the output data vector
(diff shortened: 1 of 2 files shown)


kronbichler (approved):

include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -303,7 +303,8 @@ namespace internal
     template <int  face_direction,
               bool contract_onto_face,
               bool add,
-              int  max_derivative>
+              int  max_derivative,
+              bool lex_faces = false>
```
kronbichler, line 307: Could you please document the new template parameter?

peterrum: Done!

include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -296,14 +296,20 @@ namespace internal
      *             derivatives, 2 second derivates. Note that all the
      *             derivatives access the data in @p shape_values passed to
      *             the constructor of the class
+     * @tparam lex_faces Sets how the evaluation points on the faces should be
+     *                   sorted: lexicographically or deal.II-style (special
+     *                   treatment of orientation 1 in 3D). Per default
+     *                   deal.II-order is enabled. Note: that the deal.II
+     *                   ordering is only working for dimensions up to 3.
```
kronbichler, line 303: I would not call it `deal.Il` orientation but rather `right-hand-system` numbering or something like that (which is a 3D concept). The idea is to let axis direction wrap around in defining the order, rather than going lexicographically from the faster running tensor index to the slower running one. What do you think?

peterrum: > What do you think?

I agree. I just did not know how to call it. I have made the change!

kronbichler (approved):

PR #10898, opened 2020-09-09
Mention Code Complete as a resource.

bangerth:


bangerth:
I'm reading this book for a second time in its entirety with my research group, and I continue to like it a lot. It's worth mentioning.

examples/step-26/step-26.cc
```diff
@@ -644,10 +644,14 @@ namespace Step26
...
-// code duplication and awkward control flow. This may be one of these
-// places.
+// <code>goto</code> statements are a bad idea but be pragmatic and
+// state that there may be occasions where they can help avoid code
+// duplication and awkward control flow. This may be one of these
+// places, and it matches the position Steve McConnell takes in his
+// excellent book "Code Complete" about good programming practices
+// (see the mention of this book in the introduction of step-1) that
+// spends a surprising ten pages on the question of <code>goto</code>
+// in general.
```
drwells, line 654: While we are at it why don't we add a citation for code complete so that we can cite it in programs in the normal way?

drwells (approved):
Looks good to me but I have an optional improvement.

examples/step-26/step-26.cc
bangerth: Good idea!

bangerth:
So done!

examples/step-1/doc/intro.dox
```diff
@@ -185,6 +185,15 @@ are worthwhile browsing through before you start any large-scale programming:
   do I debug scientific computing codes?" or "Can I train myself to write code
   that has fewer bugs?".
 
+- You will benefit from becoming a better programmer. An excellent
+  resource to this end is the book "<a
+  href="https://en.wikipedia.org/wiki/Code_Complete">Code
+  Complete</a>" by Steve McConnell @code CodeComplete . It's already
```
drwells, line 191:
```suggestion
  resource to this end is the book 
  <a href="https://en.wikipedia.org/wiki/Code_Complete">"Code Complete"</a>
  by Steve McConnell @cite CodeComplete. It's already
```
doxygen just prints the link literally instead of including the HTML when the whole link is quoted for me. We definitely need `@cite` and not `code`.

drwells (changes requested):
I double checked the doxygen output and we need to fix some minor things.

examples/step-1/doc/intro.dox
bangerth: Let me do this via markdown. Does this work for you now?

drwells: Almost - you forgot to delete `@code` and add `@cite`.

bangerth:
Fixed now!

drwells (approved):

PR #10900, opened 2020-09-09
Use safe versions of comparison operators.

bangerth:
I've used Steve McConnell's "Code Complete" as my go-to solution for all questions of good programming practice and have found it to be an excellent resource. It agrees with my general philosophy on programming in nearly all aspects.

So I was intrigued by a section on p. 442 that "In C++, consider creating preprocessor macro substitutions for &&, ||, and ==". Reading through the section, I find that that makes a whole lot of sense to me, and that we should do this in deal.II as well. McConnell suggests using `AND`, `OR`, and `EQUALS` for these macros, but that is not properly namespaced. So I chose `DEAL_II_AND`, `DEAL_II_OR`, and `DEAL_II_EQUALS`.

OK to merge?


include/deal.II/algorithms/general_data_storage.h
```diff
@@ -272,7 +272,7 @@ class GeneralDataStorage : public Subscriptor
   template <typename Type, typename Arg, typename... Args>
   Type &
   get_or_add_object_with_name(const std::string &name,
-                              Arg &&             argument,
+                              Arg DEAL_II_AND argument,
```
masterleinad, line 275: This is just not a comparison and for me also the rest is so much less readable.

peterrum:
I don't like this change.

marcfehling:
While working with python which allows comparison operators like `and`, `or`, `not`, `is` from scratch, I must say that I tend to prefer the written-out variants over the symbol ones.

However, I feel like the `DEAL_II_` prefix makes it way harder to read. I would suggest to get rid of them and accept that they're not properly namespaced...

Further, I am not sure if other (newcomer?) C++ programmers will be irritated by these macros, and if we even scare them off.

drwells:
I think we decided long ago that the C operators were preferable even to the C++ `and`, `or`, etc.

Your mass substitution also includes every single move constructor - writing this PR in such a way that we only have logical operators and not ref-refs will be a herculean undertaking.

Please, lets not do this. Lets not make deal.II look like the old bash source with its pseudo-pascal.

masterleinad (changes requested):
Everyone so far is against this.

bangerth:
Oof, y'all are no fun :-) I guess I should have reserved this patch for in 6 months and 21 days ;-)

masterleinad:
> Oof, y'all are no fun :-) I guess I should have reserved this patch for in 6 months and 21 days ;-)

Absolutely! :slightly_smiling_face:

peterrum:
Let's close this for now. If there should be a time when we really want to abandon the beautiful C-operators, then let's rather consider switching the programming language completely.

bangerth:
On 9/10/20 12:20 AM, Peter Munch wrote:
> If there should be a time when we really want to abandon the beautiful 
> C-operators, then let's rather consider switching the programming language 
> completely.

I propose Ada! (Or maybe Haskell? I've grown to like functional programming 
more over time!) :-)

PR #10901, opened 2020-09-10
Split out mapping-related part from FEEvaluationBase into base class

kronbichler:
This PR splits the mapping-dependent part of `FEEvaluationBase` (a base class of `FEEvaluation`) into a separate base class which I called `FEEvaluationMappingRelated`. The advantage is that we skip the number of solution components in this base class, and that we better group the functionality such that `FEEvaluationBase` can concentrate on the functions related to reading vectors and to handle the solution data (values, gradients) at quadrature points.

If somebody has a suggestion for a better name (we cannot easily change the `FEEvaluationBase` name because it has been part of the public name before), I'd be happy to apply that.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -32,6 +32,11 @@
 DEAL_II_NAMESPACE_OPEN
 
 
+// forward declaration
+template <int, typename, bool, typename>
+class FEEvaluationBaseData;
+
+
 
 namespace internal
 {
@@ -3165,12 +3170,14 @@ namespace internal
   template <int dim, int fe_degree, typename Number>
   struct CellwiseInverseMassMatrixImpl
   {
-    template <typename FEEvaluationType>
     static void
-    apply(const unsigned int      n_components,
-          const FEEvaluationType &fe_eval,
-          const Number *          in_array,
-          Number *                out_array)
+    apply(const unsigned int                  n_components,
+          const FEEvaluationBaseData<dim,
+                                     typename Number::value_type,
+                                     false,
+                                     Number> &fe_eval,
+          const Number *                      in_array,
+          Number *                            out_array)
     {
       constexpr unsigned int dofs_per_component =
         Utilities::pow(fe_degree + 1, dim);
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -60,25 +60,21 @@ class FEEvaluation;
 
 
 /**
- * This is the base class for the FEEvaluation classes. This class is a base
- * class and needs usually not be called in user code. It does not have any
- * public constructor. The usage is through the class FEEvaluation instead. It
- * implements a reinit method that is used to set pointers so that operations
- * on quadrature points can be performed quickly, access functions to vectors
- * for the @p read_dof_values, @p set_dof_values, and @p
- * distribute_local_to_global functions, as well as methods to access values
- * and gradients of finite element functions.
+ * This base class of the FEEvaluation and FEFaceEvaluation classes handles
+ * mapping-related information independent of the degrees of freedom and
+ * finite element in use. This class provides access functionality for user
+ * code but is otherwise invisible without any public constructor. The usage
+ * is through the class FEEvaluation instead.
  *
- * This class has three template arguments:
+ * This class has four template arguments:
  *
  * @tparam dim Dimension in which this class is to be used
  *
- * @tparam n_components Number of vector components when solving a system of
- * PDEs. If the same operation is applied to several components of a PDE (e.g.
- * a vector Laplace equation), they can be applied simultaneously with one
- * call (and often more efficiently)
- *
  * @tparam Number Number format, usually @p double or @p float
+ *
+ * @tparam is_face Whether the class is used for a cell integrator (with
+ * quadrature dimension the same as the space dimension) or for a face
+ * integrator (with quadrature dimension one less)
  *
  * @tparam VectorizedArrayType Type of array to be woked on in a vectorized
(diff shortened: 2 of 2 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -3265,12 +3270,14 @@ namespace internal
...
-          const FEEvaluationType &fe_eval,
-          const Number *          in_array,
-          Number *                out_array)
+    apply(const unsigned int                        n_components,
+          const FEEvaluationMappingRelated<dim,
+                                           typename Number::value_type,
+                                           false,
+                                           Number> &fe_eval,
+          const Number *                            in_array,
+          Number *                                  out_array)
```
kronbichler, line 3280: This function is one of the intended side effects of the proposed change: We can now compute the inverse mass matrix operation without a template parameter on the number of components.

kronbichler:
This PR moves quite a lot of code around and qualifies some accesses with `this->...` because the protected member variables have moved to the base class. There should not be any user-visible change apart from the new class.

kronbichler:
Related to #9794.

kronbichler:
Another question is if we want to move the `FEEvaluationMappingRelated` and possibly also `FEEvaluationBase` and `FEEvaluationAccess` into separate files. The file `fe_evaluation.h` is a monster with more than 8k lines of code.

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -3265,12 +3270,14 @@ namespace internal
...
   {
-    template <typename FEEvaluationType>
     static void
-    apply(const unsigned int      n_components,
-          const FEEvaluationType &fe_eval,
-          const Number *          in_array,
-          Number *                out_array)
+    apply(const unsigned int                        n_components,
+          const FEEvaluationMappingRelated<dim,
+                                           typename Number::value_type,
```
peterrum, line 3276: I guess it is guaranteed that Number is of type VectorizedArray?

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -143,10 +129,360 @@ class FEEvaluationBase
...
+  get_internal_dof_numbering() const;
+
+  /**
+   * Return an ArrayView to internal memory for temporary use. Note that some
+   * of this memory is overwritten during evaluate() and integrate() calls so
+   * do not assume it to be stable over those calls. The maximum size you can
+   * write into is 3*dofs_per_cell+2*n_q_points.
+   */
+  ArrayView<VectorizedArrayType>
+  get_scratch_data() const;
```
peterrum, line 231: These are not really *mapping*-related, are they? Maybe the name is not correctly chosen.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -143,10 +129,360 @@ class FEEvaluationBase
...
+ *
+ * @ingroup matrixfree
+ */
+template <int dim,
+          int n_components_,
+          typename Number,
+          bool is_face                 = false,
+          typename VectorizedArrayType = VectorizedArray<Number>>
+class FEEvaluationBase
+  : public FEEvaluationMappingRelated<dim, Number, is_face, VectorizedArrayType>
```
peterrum, line 474: `FEEvaluation` -> `FEEvaluationAccess` -> `FEEvaluationBase` -> `FEEvaluationMappingRelated`: puh, I somewhat lost the the overview that is responsible for what (see also comment above).

peterrum (approved):
I am happy to merge tomorrow despite my comments below so that you can proceed to refactor!

> Another question is if we want to move the FEEvaluationMappingRelated and possibly also FEEvaluationBase and FEEvaluationAccess into separate files. The file fe_evaluation.h is a monster with more than 8k lines of code.

Feel free to do this! You have my support.

include/deal.II/matrix_free/evaluation_kernels.h
kronbichler: Currently yes because `FEEvaluation` assumes that the data type is always a `VectorizedArray<Number,width>`. We could relax that as soon as `FEEvaluation` opens for other types.

include/deal.II/matrix_free/fe_evaluation.h
kronbichler: The problem with the class it that it brings together so many things. My summary would be:
- `FEEvaluation` contains the degree-specific parts of tensor products (and the user-looking class)
- `FEEvalationAccess` provides the access functionality to quantities at quadrature points
- `FEEvaluationBase` implements the vector access plus the generic part about the quadrature point access so that the `...Access` class does not need to re-implement things between different classes
- The new base class holds some generic value arrays as well as mapping-related data not directly tied to the quantities that get evaluated. So that might give us a hint to what the name could be. I could also think of `FEEvaluationDataMetric` to reflect that we have some data and metric term in this class.

include/deal.II/matrix_free/fe_evaluation.h
kronbichler: Indeed, we should choose a different name. The restriction on the name is that `FEEvaluationBase` is fixed and we need something more "base" than that (unless we change the name and make the `...Base` an alias. My current favorite is `FEEvaluationDataMetric` as said in the other comment.

peterrum: As discussed offline let's use `FEEvaluationBaseData`. Once you have changed the name and the tests have passed, I will merge since there is no reason to over-think things right now since some major internal changes are following. @kronbichler Much appreciated your work! Looking forward to the final result!

kronbichler: I adjusted the name and additionally simplified a few (protected) constructors by removing a template argument on the components thanks to the new base class.

peterrum (approved):

PR #10904, opened 2020-09-11
FEEvaluationFaceImpl: Remove n_components template argument

### LINKED ISSUE #10862: Template argument n_components in FEFaceEvaluationImpl
PR #10820 removed the `n_components` template arguments from the evaluation implementation of cell integrals. We should do something similar for the face evaluators https://github.com/dealii/dealii/blob/13dd68ff4b4ed16ccfd7a3b335279bde2b774a95/include/deal.II/matrix_free/evaluation_kernels.h#L1324-L1331 and the implicit dependence of `CellwiseInverseMassMatrixImpl` on the argument via the call to https://github.com/dealii/dealii/blob/13dd68ff4b4ed16ccfd7a3b335279bde2b774a95/include/deal.II/matrix_free/evaluation_kernels.h#L3126-L3132 where the `FEEvaluationBase` has an `n_components` template argument. For the latter, we probably need to restructure `FEEvaluationBase` somewhat into the part that does not depend on the number of components like the functions querying the geometry as used by the inverse mass matrix here
https://github.com/dealii/dealii/blob/13dd68ff4b4ed16ccfd7a3b335279bde2b774a95/include/deal.II/matrix_free/evaluation_kernels.h#L3163-L3168 and the other parts like reading from vectors or returning the `get_value` part that do indeed depend on the components.

Works towards #9794.

kronbichler:
This gets rid of the `n_components` template argument in the face evaluator of `FEEvaluation`. The code has been changed to move the loop over components from the inner position to outside, which has required some slight restructuring of the vector indices. The test suite says I got all the indices correct.

Works towards #9794. Fixes #10862.


kronbichler:
This is probably easier to review when ignoring whitespace changes,
https://github.com/dealii/dealii/pull/10904/files?w=1

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2393,57 +2379,62 @@ namespace internal
...
           AssertDimension(face_nos.size(), 1);
 
-          FEFaceNormalEvaluationImpl<dim,
-                                     fe_degree,
-                                     n_components,
-                                     VectorizedArrayType>::
+          FEFaceNormalEvaluationImpl<dim, fe_degree, VectorizedArrayType>::
             template interpolate<false, false>(
-              data, temp1, values_array, integrate_gradients, face_nos[0]);
+              1,
```
tjhei, line 2390: can you add /* n_components */?

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -1989,30 +1992,11 @@ namespace internal
           static_dofs_per_face :
           Utilities::pow(data.data.front().fe_degree + 1, dim - 1);
 
-      // we allocate small amounts of data on the stack to signal the compiler
```
tjhei, line 1992: why did you decide that this optimization is no longer necessary?

tjhei (commented):
Looks pretty good, but oh my, this index logic... :)

include/deal.II/matrix_free/evaluation_kernels.h
kronbichler: The problem is that the allocation depends on the number of components, which is now a compile time constant. The alternative would have been to put only the data for a single component on the stack and loop in this code here rather than inside the functions (with the `/* n_components */ 1` argument). The reason why I went this way is that it reduces the amount of code slightly, plus that this part is more of the generic part where the optimization will pay off less than in the more optimized `gather_evaluate` which still has it.

include/deal.II/matrix_free/evaluation_kernels.h
kronbichler: Yes, fixed.

peterrum (approved):
Looks good to me. Out of curiosity: why did you put `n_components` as the first argument of the functions? Is there a particular reasoning or was it the most simple approach (I know it is consistent with a previous PR - but we did you do it like this there)?

kronbichler:
> Out of curiosity: why did you put `n_components` as the first argument of the functions? Is there a particular reasoning or was it the most simple approach (I know it is consistent with a previous PR - but we did you do it like this there)?

I didn't find another place along the argument list where it would fit well in the first PR #10820. Alongside the evaluation flags, it felt like one of the more important parameters for the function in terms of what exactly gets computed. We next have some input data (`ShapeInfo`) and then the data pointers. At this point, we are certainly not consistent among those functions, and we definitely should use the `EvaluationFlags` also for the face part and move it to the same place. There are still some various options or parameters spread around in the arguments which we also could unify at some point.

PR #10905, opened 2020-09-11
remove builds-on 39->12

zjiaqi2018:
Timo said one doesn't need step-39 to understand step-12 :)


masterleinad (approved):
Sounds reasonable.

drwells:
In fact step-39 currently lists step-12b as a dependency.

PR #10906, opened 2020-09-11
Fix warning: declaration of dim hides template parameter when using CUDA

Rombur:
Not sure why nvcc is not happy but since we are not supposed to get in these functions anyway, I think it's fine to set the condition to `false`.


tjhei (approved):

peterrum (approved):

drwells:
We forgot the tag. Lets try again:

/rebuild

PR #10909, opened 2020-09-12
Assign boundary ids

rezarastak:
I am introducing a new method to systematically assign boundary ids to faces of a triangulation. This follows more declarative and functional style of programming and requires the use of `std::function` and lambdas. I have modified a number of tutorials to use this new style. Let me know how you feel about it.

In short, when we assign boundary ids based on geometrical considerations, instead of:
```c++
  if (cell->is_locally_owned())
    for (const auto &face : cell->face_iterators())
      if (face->at_boundary())
        {
          const Point<dim> face_center = face->center();

          if (face_center[2] == 0)
            face->set_boundary_id(0);
          else if (face_center[2] == 3)
            face->set_boundary_id(1);
        }
```
we can write:
```c++
GridTools::assign_boundary_ids(triangulation,
      {{0, [](const auto &p) { return p[2] == 0.; }},
       {1, [](const auto &p) { return p[2] == 3.; }}});
```
The user might define helper classes to simplify this even more. For example, we can write:
```c++
GridTools::assign_boundary_ids(triangulation,
      {{0, OnPlane(2, 0.0)},
       {1, OnPlane(2, 3.0)}});
```
if we have pre-defined this helper class:
```c++
class OnPlane{
  int component;
  double location, tol;

  OnPlane(int component, double location, tol=1.-6) 
    : component{component}, location{location}, tol{tol}
  {}

  template <int dim>
  bool operator()(const Point<dim> &p){
    return std::abs(p[component] - location) < tol;
  }
};
```
Advantages:
* Functional and declarative style rather than procedural.
* Prevents users from making common mistakes such as forgetting to check if the face is on the boundary or forgetting to check if the cell is locally owned.
* Sometimes leads to shorter code especially when we have small number of boundary descriptions.
* More readable code (in my subjective opinion).

Disadvantages:
* Because of using std::function, there may be a small performance penalty but this function is not going to be run in performance-critical loops.
* Might seem less readable to some people because of the lambda syntax and heavy brace initializations.

Future improvements:
* User can provide an optional `Mapping` so that instead of `face->center()` we can check the location of the true center of face on a curved manifold.
* User can provide an optional `Quadrature` to ensure that boundary id is assigned only if all integration points match the predicate (not just the face center).

### DIFF
--- a/doc/news/changes/major/20200910Rastak
+++ b/doc/news/changes/major/20200910Rastak
@@ -1,0 +1,5 @@
+New: GridTools::assign_boundary_ids() allows quick assignment of boundary_id
+values to a triangulation using a number of geometrical predicate functions
+that could be provided as lambdas.
+<br>
+(Reza Rastak, 2020/09/10)
--- a/examples/step-18/step-18.cc
+++ b/examples/step-18/step-18.cc
@@ -40,6 +40,7 @@
 #include <deal.II/grid/tria.h>
 #include <deal.II/grid/grid_generator.h>
 #include <deal.II/grid/grid_refinement.h>
+#include <deal.II/grid/grid_tools.h>
 #include <deal.II/grid/tria_accessor.h>
 #include <deal.II/grid/tria_iterator.h>
 #include <deal.II/grid/manifold_lib.h>
@@ -753,28 +754,27 @@ namespace Step18
   // z-coordinates of either 0 (bottom face), an indicator of 1 for z=3 (top
   // face); finally, we use boundary indicator 2 for all faces on the inside
   // of the cylinder shell, and 3 for the outside.
+  //
+  // Even though we can use the set_boundary_id() method to manually assign
+  // boundary_id values to faces of the coarse mesh (as shown in step-7), here
+  // we use the convenient function GridTools::assign_boundary_ids() to
+  // specify the boundary ids for the whole triangulation in a single
+  // statement.
   template <int dim>
   void TopLevel<dim>::create_coarse_grid()
   {
     const double inner_radius = 0.8, outer_radius = 1;
     GridGenerator::cylinder_shell(triangulation, 3, inner_radius, outer_radius);
-    for (const auto &cell : triangulation.active_cell_iterators())
-      for (const auto &face : cell->face_iterators())
-        if (face->at_boundary())
-          {
-            const Point<dim> face_center = face->center();
 
-            if (face_center[2] == 0)
-              face->set_boundary_id(0);
-            else if (face_center[2] == 3)
-              face->set_boundary_id(1);
-            else if (std::sqrt(face_center[0] * face_center[0] +
-                               face_center[1] * face_center[1]) <
-                     (inner_radius + outer_radius) / 2)
-              face->set_boundary_id(2);
-            else
-              face->set_boundary_id(3);
-          }
+    GridTools::assign_boundary_ids(
+      triangulation,
+      {{/*id =*/0, [](const Point<dim> &p) { return p[2] == 0.; }},
+       {/*id =*/1, [](const Point<dim> &p) { return p[2] == 3.; }},
+       {/*id =*/2,
+        [=](const Point<dim> &p) {
+          return p[0] * p[0] + p[1] * p[1] < (inner_radius + outer_radius) / 2.;
+        }},
+       {/*id =*/3, [](const Point<dim> &) { return true; }}});
 
     // Once all this is done, we can refine the mesh once globally:
     triangulation.refine_global(1);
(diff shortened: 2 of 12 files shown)


bangerth:
I'd rather if the function (lambda) got an iterator to a face as argument. It can then figure out other properties such as the center itself, but can also query information not accessible from just a point.

I'd also revert the order of the pairs: first the function (a "predicate"), then the boundary id to be set if the predicate is true.

rezarastak:
> I'd also revert the order of the pairs

I agree with your suggestion. Thank you for the feedback.

> I'd rather if the function (lambda) got an iterator to a face as argument.

I agree, a function in the form
```c++
std::vector<types::global_cell_index>
assign_boundary_ids(Triangulation<dim, spacedim> &tria,
       std::function<types::boundary_id(Triangulation<dim, spacedim>::face_iterator)> &boundary_selector){
  for (const auto &cell : tria.active_cell_iterators())
     if (cell->is_locally_owned())
       for (auto face : cell->face_iterators())
         if (face->at_boundary())
           face->set_boundary_id(boundary_selector(face));
}
```
is very handy and reusable and extremely flexible. However, I think the original function that I proposed has a distinct and specialized role and deserves its own place because:
 1. A function that accepts `Point<spacedim>` as argument is a true geometrical abstraction independent of triangulation and mesh details. We can define a plane with `[](const auto &p) {return p[0] == 0.;}` and circle with `[](const auto &p){return std::abs(p[0] * p[0] + p[1] * p[1] - 1.) < 1.e-6;}` without knowing or caring about whether this is evaluated at the center of the face or at a different location (based on a custom mapping) or whether it is evaluated at multiple points (based on a custom quadrature) on the face. Assigning boundary id values based on the geometry requires a purely geometrical abstraction, and I think the most concise geometrical abstraction in C++ is `std::function<bool(const Point<spacedim> &)>` (or its c++20 concept equivalent).
2. Using `std::function<bool(const Point<spacedim> &)>` is slightly faster because in the code, for each face we compute the center once and then test the face center with multiple predicates. However, if I describe the boundary with `std::function<bool(Triangulation<dim, spacedim>::face_iterator)>`, each predicate must compute the center inside the body of the predicate. Therefore, the method `center()` will be unnecessarily called multiple times for each face.

In summary, I suggest we have two overloads for assign_boundary_ids, one that accepts a list of pairs of `types::boundary_id` `std::function<bool(const Point<spacedim> &)>` and one that accepts `std::function<types::boundary_id(Triangulation<dim, spacedim>::face_iterator)>`. One of them can internally call the other function. Users can choose between the more flexible version based on face iterators and the version which uses functions to describe the geometry of the boundaries..

bangerth:
I think the reason why I don't particularly like the version that just takes a point is that it's not clear what point one might want to call for a given face: The midpoint of the vertices? The point that the manifold returns as the midpoint? The point that a mapping (which one?) would return for the face midpoint? What if the goal of calling this function is to associate a certain manifold with the face, but that that hasn't happened yet? Then the point you'd get passed would be the midpoint in Cartesian coordinates, but after you've associated a manifold with the face, you'd get a different point.

For these kinds of reasons, I think it would be easier if the thing the function receives is simply the face itself, rather than something that one can obtain from the face. Providing a second variant only makes things more complicated precisely because of considerations such as those above.

bangerth:
@rezarastak Have you had time to think about my last comment?

rezarastak:
Hi @bangerth, sorry I didn't continue working on this. However, I have been trying to find the best programming interface for looping over boundary faces and assigning boundary ids (or perhaps do something else with the boundary faces). The goal is to simplify the boundary assignment process by introducing new abstractions, while keeping it readable and flexible. One idea that seems promising is defining `boundary_face_iterators()`.

For example, we can simplify
```c++
 for (auto &cell : triangulation.active_cell_iterators())
  if (cell->is_locally_owned())
    for (auto &face : cell->face_iterators())
      if (face->at_boundary())
        {
          // test face center and assign boundary id
        }
```

into something like this:
```c++
for (auto &face : triangulation.boundary_face_interators())
{
  // test face center and assign boundary id
}
```

What do you think about this abstraction?

peterrum:
> What do you think about this abstraction?

I think the abstraction would be nice and it would be definitely applied in user code. You probably would need to implement a new `TriaIterator` class and/or apply a `FilteredIterator`. Maybe we could even introduce a new filter that could be used like:

```cpp
for (auto &face : apply(triangulation.face_interators(), BoundaryFaceFilter())
  // test face center and assign boundary id
```

drwells:
This has been a draft for three years. Lets close for now and open it up again if people show interest in the topic again.

PR #10910, opened 2020-09-13
Fix categorize_by_boundary_ids for PBC

peterrum:
... and some clean up

### DIFF
--- a/include/deal.II/matrix_free/tools.h
+++ b/include/deal.II/matrix_free/tools.h
@@ -39,11 +39,24 @@ namespace MatrixFreeTools
   categorize_by_boundary_ids(
     const Triangulation<dim> &tria,
     AdditionalData &          additional_data,
-    const unsigned int        level = numbers::invalid_unsigned_int)
-  {
-    bool is_mg = (level != numbers::invalid_unsigned_int);
+    const unsigned int        level = numbers::invalid_unsigned_int);
 
-    // ... create list for the category of each cell
+
+
+  // implementations
+
+
+
+  template <int dim, typename AdditionalData>
+  void
+  categorize_by_boundary_ids(const Triangulation<dim> &tria,
+                             AdditionalData &          additional_data,
+                             const unsigned int        level)
+  {
+    // ... determine if we are on an active or a multigrid level
+    const bool is_mg = (level != numbers::invalid_unsigned_int);
+
+    // ... create empty list for the category of each cell
     if (is_mg)
       additional_data.cell_vectorization_category.resize(
         std::distance(tria.begin(level), tria.end(level)));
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/tools.h
```diff
@@ -39,40 +39,55 @@ namespace MatrixFreeTools
...
-    auto to_category = [&](auto &cell) {
+    const auto to_category = [&](const auto &cell) {
       unsigned int c_num = 0;
-      for (unsigned int i = 0; i < dim * 2; i++)
+      for (unsigned int i = 0; i < GeometryInfo<dim>::faces_per_cell; i++)
         {
           auto &face = *cell->face(i);
-          if (face.at_boundary())
-            c_num += factors[i] * bid_map[face.boundary_id()];
+          if (face.at_boundary() && !cell->has_periodic_neighbor(i))
```
peterrum, line 85: In `MatrixFree` we regard periodic faces as inner faces. Missing this check led to too many categories and utterly destroyed the performance for small problem sizes in 6D in `hyper.deal`.

kronbichler: The "utterly destroyed [...] performance for small problem sizes" is a general property of categorization and not limited to 6D: If you have few cells locally and you e.g. sit in a corner of the domain, it can be hard to fil all SIMD lanes with the same category. The problematic part with cell-based loops is that the decision applies an `and` across all faces of a cell, so the different restrictions end up accumulating over essentially all work that gets done. It would be interesting to describe the "categorization efficiency" as the ratio between the minimal number of cell batches, `(n_local_cells + n_simd_lanes - 1)/n_simd_lanes` and the actual number of cell batches.

In the longer run, I believe we will need to find entry points for the GPU way (or data-parallel) way of doing things, aka running both conditions of a branch and masking.

kronbichler (approved):

include/deal.II/matrix_free/tools.h
peterrum: > "categorization efficiency" as the ratio between the minimal number of cell batches, (n_local_cells + n_simd_lanes - 1)/n_simd_lanes and the actual number of cell batches

I was thinking about the same thing. Such a metric would have helped me to find the reason for the slow down quicker.

> In the longer run, I believe we will need to find entry points for the GPU way (or data-parallel) way of doing things, aka running both conditions of a branch and masking.

I agree. But image how step-67 + ECL would look like in this case...

kronbichler: > But image how step-67 + ECL would look like in this case...

That calls for a library function hiding the complex details behind a nice user interface. :wink:

PR #10912, opened 2020-09-14
Mark two functions 'noexcept'.

bangerth:
With a newer compiler, I got a warning about missing `noexcept` on two functions. I have no idea what the `// NOLINT` comment on these lines referred to, but assume that it is unrelated. Let's see what the testers have to say.

### DIFF
--- a/include/deal.II/lac/affine_constraints.h
+++ b/include/deal.II/lac/affine_constraints.h
@@ -565,7 +565,8 @@ public:
   /**
    * Move constructor
    */
-  AffineConstraints(AffineConstraints &&affine_constraints) = default; // NOLINT
+  AffineConstraints(AffineConstraints &&affine_constraints) noexcept =
+    default; // NOLINT
 
   /**
    * Copy operator. Like for many other large objects, this operator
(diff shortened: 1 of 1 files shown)


masterleinad:
You need to fix the indentation. :slightly_smiling_face:

drwells:
> I have no idea what the // NOLINT comment on these lines referred to, but assume that it is unrelated.

If I recall correctly we had do add argument names (even though the methods are defaulted) to appease GCC 4.8 or GCC 4.9, while since we never use the arguments they do not need to be named (which causes clang-tidy to print a warning).

tjhei:
> which causes clang-tidy to print a warning

clang-tidy seems to be happy though?

drwells:
You're right - we have some move constructors without variable names and some with variable names, but both sets use `// NOLINT`. I must be thinking of something else.

drwells:
(and some move constructors with no variable name and no `// NOLINT`)

masterleinad:
`clang-tidy` will never complain for lines that have `NOLINT`. `clang-tidy` wants move constructors/assignment operators to be explicitly marked `noexcept` even though the compiler can calculate the appropriate specifier itself. IIRC some older compilers complain, though if the implicit specifier doesn't match the explicit (resp. missing) one. Hence, we just silenced `clang-tidy` here.
If this works with all compilers we care about, we can remove the `NOLINT` here.

bangerth:
Indented -- the result is not pretty :-(

masterleinad (approved):

kronbichler (approved):

PR #10920, opened 2020-09-16
Fix multiple-component bug in FEFaceEvaluationImpl::integrate_scatter

kronbichler:
When using multiple components of a finite element basis without optimized reads (e.g. `FE_DGQ` for computing the gradients on faces), `FEFaceEvaluationImpl::integrate_scatter` would return early to indicate that it needed an external function to write into the vector, but we actually were still looping over the components to prepare the data. So we must complete the loop when writing but not when reading. This bug was introduced in #10904; the more serious defect was that this code path was untested before, so the hardly needed test is also added.

FYI @nfehn @peterrum

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -2251,7 +2251,7 @@ namespace internal
                                temp1,
                                values_quad + comp * n_q_points,
                                gradients_quad + comp * dim * n_q_points,
-                               scratch_data + 2 * n_components * dofs_per_face,
+                               scratch_data + 2 * dofs_per_face,
                                evaluate_values,
                                evaluate_gradients,
                                subface_index);
@@ -2266,7 +2266,7 @@ namespace internal
                                temp1,
                                values_quad + comp * n_q_points,
                                gradients_quad + comp * dim * n_q_points,
-                               scratch_data + 2 * n_components * dofs_per_face,
+                               scratch_data + 2 * dofs_per_face,
                                evaluate_values,
                                evaluate_gradients,
                                subface_index);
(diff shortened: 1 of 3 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -3113,7 +3121,7 @@ namespace internal
                                       gradients_quad);
         }
 
-      return true;
+      return writes_into_vector;
```
peterrum, line 3124: Ooops!

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2251,7 +2251,7 @@ namespace internal
                                temp1,
                                values_quad + comp * n_q_points,
                                gradients_quad + comp * dim * n_q_points,
-                               scratch_data + 2 * n_components * dofs_per_face,
+                               scratch_data + 2 * dofs_per_face,
```
peterrum, line 2254: This is just a refactoring step, isn't it?

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2652,6 +2652,8 @@ namespace internal
                  orientation[v][i];
       };
 
+      bool writes_into_vector = true;
```
peterrum, line 2655: May I ask you to add a comment here.

peterrum (approved):
Looks good to me!

include/deal.II/matrix_free/evaluation_kernels.h
kronbichler: Yes, this was forgotten in #10904.

include/deal.II/matrix_free/evaluation_kernels.h
kronbichler: I added a comment here and further down where the variable is set (or the early return for the read path).

peterrum (approved):
Thanks! I will merge once the testers are happy since this is a bug all matrix-free users might affect.

peterrum (approved):
I'll merge this one although the OSX tests have not passed yet. The bug prevents many of the (if not all modern) matrix-free codes to work properly.

PR #10922, opened 2020-09-16
Refactor access to FEEvaluationImpl and friends

kronbichler:
As a preparation for #9794, I want to distinguish the calls to the various implementations with different `struct` names but the same `run()` function (to generically instantiate whatever is needed for the polynomial degrees). This PR makes the necessary refactoring to the cell and face evaluations, mostly just moving around some code.

~~I have set the `do not merge` label for now because I want to work on the downstream code tomorrow to see if there is anything else needed, but I wanted to announce it now to see if there are comments.~~

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -1396,7 +1396,7 @@ namespace internal
   struct FEEvaluationImplEvaluateSelector
   {
     template <int fe_degree, int n_q_points_1d>
-    static void
+    static bool
     run(const unsigned int                                      n_components,
         const EvaluationFlags::EvaluationFlags                  evaluation_flag,
         const internal::MatrixFreeFunctions::ShapeInfo<Number> &shape_info,
@@ -1519,6 +1519,8 @@ namespace internal
         }
       else
         AssertThrow(false, ExcNotImplemented());
+
+      return false;
     }
   };
 
@@ -1543,7 +1545,7 @@ namespace internal
   struct FEEvaluationImplIntegrateSelector
   {
     template <int fe_degree, int n_q_points_1d>
-    static void
+    static bool
     run(const unsigned int                     n_components,
         const EvaluationFlags::EvaluationFlags integration_flag,
         const internal::MatrixFreeFunctions::ShapeInfo<Number> &shape_info,
@@ -1666,6 +1668,8 @@ namespace internal
         }
       else
         AssertThrow(false, ExcNotImplemented());
+
+      return false;
     }
   };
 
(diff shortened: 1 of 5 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -1967,27 +2291,71 @@ namespace internal
...
-  template <int dim,
-            int fe_degree,
-            int n_q_points_1d,
-            typename Number,
-            typename VectorizedArrayType,
-            typename Number2 = Number>
-  struct FEFaceEvaluationSelector
+  template <typename Number>
+  void
+  adjust_for_face_orientation(const unsigned int            dim,
```
peterrum, line 2296: Does it hurt if you make dim a template parameter?

peterrum (commented):
Looks good from my side. I think I know in which direction this is going!

Ping me once you have tried out the downstream code and added the promised tests! Looking forward to the next step!

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -1377,6 +1377,330 @@ namespace internal
...
+    static void
+    run(const unsigned int                                      n_components,
+        const EvaluationFlags::EvaluationFlags                  evaluation_flag,
+        const internal::MatrixFreeFunctions::ShapeInfo<Number> &shape_info,
+        Number *values_dofs_actual,
+        Number *values_quad,
+        Number *gradients_quad,
+        Number *hessians_quad,
+        Number *scratch_data);
+  };
```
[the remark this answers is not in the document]
kronbichler: I am leaning towards using variadic template `args...` because the only way I see those could be generalized is by having a new struct that hides the function-specific part, which is also not pretty.

peterrum (approved):

include/deal.II/matrix_free/evaluation_kernels.h
kronbichler: It's just a bit more code to be generated without any benefit - `dim` is only used for an outer loop.

kronbichler:
I did two adjustments: I also applied the same interface to the inverse mass matrix and I changed the return type for all functions to `bool` because that is needed for the `gather_evaluate` call. It is not super pretty this way but we can tackle that in a later PR.

I consider this PR ready now.

> [...] and added the promised tests! 

That was a misleading comment from my side - I intended to say that I had to fix some of the failing existing tests (which is now concluded). The new interfaces are exercised by all code using `FEEvaluation` or `FEFaceEvaluation`, thus more than 400 tests (counting debug + release), which should be sufficient at this point. In a follow-up PR addressing #9794 I will switch a good bunch of those matrix-free tests to the non-templated version, so we stress even more parts.

peterrum (approved):
Looks good. As discussed offline, let's get #9794 working as soon as possible. We can clean up the internal data structures of `MatrixFree` afterwards. Let's try to distribute the task then fairly! Do you want to create a new issue so that we don't forget it. You can assign us both to it.

peterrum:
Thanks! One step closer to our goal!

PR #10926, opened 2020-09-16
step-69: Remove boundary c_ij fix

tamiko:
We compute the boundary normal by averaging the normal over the
quadrature points weighted with the shape function. It turns out that
this is sufficient to preserve mass and total energy. No change of the
c_ij's are needed.


tamiko:
@itomasSNL @kronbichler *ping*

tamiko:
All, I would like to get a review by @itomasSNL on this one :sweat_smile:

kronbichler (approved):

tamiko:
Let's merge this as is - @itomasSNL promised to follow up with some documentation updates in a separate pull request :-)

itomasSNL:
@kronbichler @tamiko Now the documentation and the source code are consistent.

kronbichler (approved):

PR #10936, opened 2020-09-20
Fix SCALAPACK tests again

masterleinad:
Fixes https://cdash.43-1.org/testDetails.php?test=58278061&build=11074. This time I actually ran the tests...


drwells (approved):

drwells:
The tester didn't catch it last time but we might as well:

/rebuild

peterrum (approved):

PR #10940, opened 2020-09-20
Cleanup quadrature

drwells:
I started by fixing another `-Wtype-limits` warning and then did some other cleaning up.

### DIFF
--- a/source/base/quadrature.cc
+++ b/source/base/quadrature.cc
@@ -136,9 +136,8 @@ Quadrature<dim>::Quadrature(const SubQuadrature &q1, const Quadrature<1> &q2)
   for (unsigned int i2 = 0; i2 < q2.size(); ++i2)
     for (unsigned int i1 = 0; i1 < q1.size(); ++i1)
       {
-        // compose coordinates of
-        // new quadrature point by tensor
-        // product in the last component
+        // compose coordinates of new quadrature point by tensor product in the
+        // last component
         for (unsigned int d = 0; d < dim - 1; ++d)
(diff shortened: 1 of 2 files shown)


source/base/quadrature.cc
```diff
@@ -107,6 +108,17 @@ Quadrature<1>::Quadrature(const Point<1> &point)
 
 
 
+template <>
+Quadrature<0>::Quadrature(const Point<0> &point)
+  : quadrature_points(std::vector<Point<0>>(1, point))
+  , weights(std::vector<double>(1, 1.))
```
tjhei, line 114: I think you can skip all these initializations.

source/base/quadrature.cc
```diff
@@ -449,108 +456,6 @@ namespace internal
   }   // namespace QIteratedImplementation
 } // namespace internal
 
-// template <>
```
tjhei, line 452: huh

drwells: This function does almost the same thing as the constructor which is why we got rid of it. Are you asking for a change?

source/base/quadrature.cc
drwells: Good point. We do need to initialize the boolean though.

bangerth (approved):
I think the only functional change is the addition of the `Quadrature(const Point<0> &)` constructor, right? Everything else is just removing commented-out lines and reflowing comments. I wonder then why you actually need the new constructor?

drwells:
I need the constructor to avoid the loop in the dim = 0 case to avoid a compiler warning.

PR #10942, opened 2020-09-21
[step-19] Correct & simplify some computations

kronbichler:
I read through the new tutorial program. Given that I missed the party in #10301, I try to add my reflections now. I will do a few comments on the code below in a minute so the tutorial authors can comment.

Note that I get substantially different results because the code contained a big bug as I believe. Have a look here:
https://github.com/dealii/dealii/blob/1107d12a13062a4de70cde81525547469ce6d77b/examples/step-19/step-19.cc#L503-L514
You create a new variable `fe_value` (note the missing *s*) but then use the surrounding `fe_values` object. This was certainly wrong and might be the explanation why I saw so many particles disappearing near the inlet (cathode).

### DIFF
--- a/examples/step-19/doc/intro.dox
+++ b/examples/step-19/doc/intro.dox
@@ -27,7 +27,7 @@ time, then deal.II is clearly not your right tool. On the other hand, if
 this evolution is due to the interaction with the solution of partial differential
 equation, or if having a mesh to determine which particles interact
 with others (such as in the
-[smoothed particle hydrodynamics (SMH)](https://en.wikipedia.org/wiki/Smoothed-particle_hydrodynamics)
+[smoothed particle hydrodynamics (SPH)](https://en.wikipedia.org/wiki/Smoothed-particle_hydrodynamics)
 method), then deal.II has support for you.
 
 The case we will consider here is how electrically charged particles move through
@@ -60,7 +60,7 @@ density. This is augmented by boundary conditions that we will choose as follows
    && \text{on}\; \partial\Omega\setminus\Gamma_\text{cathode}\setminus\Gamma_\text{anode}.
 @f}
 In other words, we prescribe voltages $+V_0$ and $-V_0$ at the two electrodes
-and isolating (Neumann) boundary conditions elsewhere. Since the dynamics of the
+and insulating (Neumann) boundary conditions elsewhere. Since the dynamics of the
 particles are purely due to the electric field $\mathbf E=\nabla V$, we could
 as well have prescribed $2V_0$ and $0$ at the two electrodes -- all that matters
 is the voltage difference at the two electrodes.
@@ -426,7 +426,7 @@ electrons are then accelerated towards the (green) anode which has a
 hole in the middle through which the electrons can escape the device and
 fly on to hit the screen, where they excite the "phosphor" to then emit
 the light that we see from these old-style TV screens. The non-heated
-part of the cathode is not heated, and consequently not subject
+part of the cathode is not subject
 to the emission of electrons -- in the code, we will mark this as the
 "focussing element" of the tube, because its negative electric voltage
 repels the electrons and makes sure that they do not just fly
--- a/examples/step-19/step-19.cc
+++ b/examples/step-19/step-19.cc
@@ -299,7 +299,7 @@ namespace Step19
       }
 
     triangulation.create_triangulation(
-      {std::begin(vertices), std::end(vertices)},
+      vertices,
       cells,
       SubCellData()); // No boundary information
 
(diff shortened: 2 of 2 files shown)


examples/step-19/doc/intro.dox
```diff
@@ -60,7 +60,7 @@ density. This is augmented by boundary conditions that we will choose as follows
    && \text{on}\; \partial\Omega\setminus\Gamma_\text{cathode}\setminus\Gamma_\text{anode}.
 @f}
 In other words, we prescribe voltages $+V_0$ and $-V_0$ at the two electrodes
-and isolating (Neumann) boundary conditions elsewhere. Since the dynamics of the
+and insulating (Neumann) boundary conditions elsewhere. Since the dynamics of the
```
kronbichler, line 63: I think it should be *insulating* for electrical comments, but given that my native language does not distinguish between isolate and insulate I let someone else check whether I am right.

examples/step-19/step-19.cc
```diff
@@ -468,52 +467,16 @@ namespace Step19
...
                 cell_rhs(i) +=
-                  (fe_values.shape_value(i, q_index) *   // phi_i(x_p)
-                   (-Constants::electrons_per_particle * // N
-                    Constants::electron_charge));        // e
-          }
-
+                  (fe.shape_value(i, reference_location) * // phi_i(x_p)
+                   (-Constants::electrons_per_particle *   // N
+                    Constants::electron_charge));          // e
+            }
```
kronbichler, line 479: I did not like the old code at all because the constructor of `FEValues` is so horrendously expensive. Here it is really simple (given that we use a `FiniteElement` whose values do not depend on the position), we should simply use the *virtual* `FiniteElement::shape_value()` function, much faster than the alternative. Of course, the really interesting place would be the one below where we need the gradients which depend on the mapping. Given that we have many particle codes, we should think of a better way to compute mapped gradients in those position. The `FEValues` interface is inappropriate in my opinion, even though it is quite powerfule.

examples/step-19/step-19.cc
```diff
@@ -760,9 +713,9 @@ namespace Step19
                 particle->set_properties(make_array_view(new_velocity));
 
                 // With the new velocity, we can then also update the location
-                // of the particle and set tell the particle about it.
+                // of the particle and tell the particle about it.
                 const Point<dim> new_location =
-                  particle->get_location() + dt * old_velocity;
+                  particle->get_location() + dt * new_velocity;
```
kronbichler, line 718: If I understand the formula and the comment above correctly, it should be the new velocity here, shouldn't it?

examples/step-19/doc/intro.dox
bangerth: I must have been speaking too much German that day. You're completely right.

examples/step-19/step-19.cc
```diff
@@ -722,7 +675,7 @@ namespace Step19
           // Then we can ask the FEValues object for the gradients of the
           // solution (i.e., the electric field $\mathbf E$) at these locations
           // and loop over the individual particles:
-          field_gradients.resize(n_particles_in_cell);
+          field_gradients.resize(quadrature_formula.size());
```
bangerth, line 678: then also just move that variable declaration here

examples/step-19/step-19.cc
bangerth: That's the difference between the explicit Euler and some semi-implicit scheme. For the purposes of this program, it probably doesn't matter very much. You're right that what you have here matches the formula above -- what did I write in the introduction? I think I *wanted* to use the explicit Euler scheme.

bangerth (commented):
Thanks for reading through it carefully! What a bug to have in there :-(

examples/step-19/step-19.cc
kronbichler: I will let you go over it after this PR is merged. I note that the introduction states this one here:
https://github.com/dealii/dealii/blob/f0919993dd8325c993eea8af248e233c1a11bd59/examples/step-19/doc/intro.dox#L180-L186
So this corresponds to some operator splitting (Gauss-Seidel) type approach.
Later, the results section goes on to mention the forward Euler case
https://github.com/dealii/dealii/blob/f0919993dd8325c993eea8af248e233c1a11bd59/examples/step-19/doc/results.dox#L265-L272
which is incorrect because the time `n` refers to a backward Euler case. But this is not really a backward Euler in the sense of differential algebraic equations because the electrical potential does not refer to the new values, but rather the splitting. I will open an issue for this.

examples/step-19/step-19.cc
```diff
@@ -697,19 +660,9 @@ namespace Step19
...
-          {
-            typename Particles::ParticleHandler<dim>::particle_iterator
-              particle = particles_in_cell.begin();
-            for (unsigned int particle_index = 0;
-                 particle != particles_in_cell.end();
-                 ++particle, ++particle_index)
-              particle_positions[particle_index] =
-                particle->get_reference_location();
-          }
+          particle_positions.clear();
```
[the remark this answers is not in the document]
kronbichler: Yes, I will.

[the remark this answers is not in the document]
kronbichler: Thinking about it, I think I will also skip the `reserve` because that seems like a micro-optimization at this point, now that I'm initializing an empty variable and using `push_back` into it.

kronbichler:
I addressed the comments, except for the one on time integration for which I opened #10945.

tamiko (approved):

bangerth:
Thanks for addressing this!

PR #10944, opened 2020-09-22
Fix comping with ICC

masterleinad:
Fixes https://cdash.43-1.org/viewBuildError.php?buildid=11166.


kronbichler (approved):

drwells:
This is also a lot easier to read.

PR #10948, opened 2020-09-22
Fix step 68 animations

blaisb:
The three https links in the step-68 were not working correctly because they did not try to embed the movie (at least this is my hypothesis). This PR should fix this issue. @peterrum


masterleinad (approved):
Looks good to me (after downloading the documentation).

PR #10952, opened 2020-09-23
step-68: A few updates to documentation + slight code cleanup

kronbichler:
The code cleanup (which should deliver equivalent results) do the following:
- Consistently use mapping object of the class itself
- Use MappingQGeneric rather than MappingQ because the latter class is somewhat brittle
- Only set up background dofs for the case of interpolated_velocity
- Do not call Utilities::MPI::this_mpi_rank in inner loop because it shows up in profiles

### DIFF
--- a/examples/step-68/doc/intro.dox
+++ b/examples/step-68/doc/intro.dox
@@ -130,11 +130,11 @@ also be different (non-matching) from the triangulation of the background grid,
 which is useful to generate particles in particular shapes (as in this
 example), or to transfer information between two different computational grids
 (as in step-70).  Furthermore, the Particles::ParticleHandler class provides the
-Particles::ParticleHandler::insert_global_particles function() which enables the
+Particles::ParticleHandler::insert_global_particles() function which enables the
 global insertion of particles from a vector of arbitrary points and a global
 vector of bounding boxes. In the present step, we use the
 Particles::Generators::quadrature_points() function on a non-matching triangulation to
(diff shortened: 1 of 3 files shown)


peterrum (approved):

tjhei:
> * Use MappingQGeneric rather than MappingQ because the latter class is somewhat brittle

The example uses a hyper_cube so we don't have a curved boundary (or a curved interior where MappingQGeneric makes more sense). Wouldn't it make sense to use a MappingQ(1) instead?

kronbichler:
> Wouldn't it make sense to use a MappingQ(1) instead?

Definitely from my point of view (even though it of course boils down to a bilinear mapping because the program is ran with degree 1). I guess it depends on what the authors want to show. @blaisb @gassmoeller @peterrum any comments?

peterrum:
I let @blaisb comment on this. It's his baby.

blaisb:
> > Wouldn't it make sense to use a MappingQ(1) instead?
> 
> Definitely from my point of view (even though it of course boils down to a bilinear mapping because the program is ran with degree 1). I guess it depends on what the authors want to show. @blaisb @gassmoeller @peterrum any comments?

To be honest i don't have any preference. I think it would make sense to use a MappingQ(1). I think I would prefer this option.
Otherwise, thank you very much for the modifications / comments! They are all extremely accurate.

kronbichler:
I changed to `MappingQ1`.

peterrum (approved):

PR #10954, opened 2020-09-23
FEFaceEvaluation: Fix fast vector access case of gather_evaluate

kronbichler:
In the `gather_evaluate` and `integrate_scatter` functions, we assume in some places that we know the polynomial degree as a template parameter. However, for the case the user only requested the values on faces and we have nodal basis functions, we missed that check.

While there (and looking in the wrong place), I also converted all loops to compile-time constants when possible and switched some loops and `if` statements to help the compiler create better code.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -2528,7 +2528,8 @@ namespace internal
         (face_orientations[0] > 0 &&
          (subface_index < GeometryInfo<dim>::max_children_per_cell ||
           !(((do_gradients == false &&
-              data.data.front().nodal_at_cell_boundaries == true) ||
+              data.data.front().nodal_at_cell_boundaries == true &&
+              fe_degree > 0) ||
              (data.element_type ==
                 MatrixFreeFunctions::tensor_symmetric_hermite &&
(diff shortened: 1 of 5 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2679,8 +2680,15 @@ namespace internal
...
+
+        // we can only use the fast functions if we know the polynomial degree
+        // as a template parameter (fe_degree != -1), and it only makes sense
+        // to use the functions for at least linear functions for values on
+        // the faces and quadratic functions for gradients on the faces, so
+        // include the switch here
         if ((do_gradients == false &&
-             data.data.front().nodal_at_cell_boundaries == true) ||
+             data.data.front().nodal_at_cell_boundaries == true &&
+             fe_degree > 0) ||
```
kronbichler, line 2691: This is the essential addition - the `do_gradients == false && data.data.front().nodal_at_cell_boundaries == true)` would run here with `fe_degree = -1` which would not work with some compile-time offsets below. The rest of the changes are just the static loop lengths.

kronbichler:
FYI @nfehn @peterrum

peterrum (approved):

PR #10956, opened 2020-09-23
Fix simplex test failure

peterrum:
... let's hope...


peterrum:
This seems to work. @elauksap what are your thoughts?

kronbichler (approved):

elauksap:
That looks fine, thank you @peterrum!

PR #10962, opened 2020-09-24
Add overload of copy_to_host() and copy_to_dev()

Rombur:


include/deal.II/base/cuda.h
```diff
@@ -151,6 +151,34 @@ namespace Utilities
                                                cudaMemcpyHostToDevice);
       AssertCuda(cuda_error_code);
     }
+
+    /**
+     * Copy @p size elements in @p pointer_dev to the host pointer @p pointer_host.
+     */
+    template <typename T>
+    inline void
+    copy_to_host(const T *pointer_dev, unsigned int size, T *pointer_host)
```
peterrum, line 160: I find the sequence of the arguments a bit odd. Why not do it like in the `cudaMemcpy` or `std::memcpy` case?

Rombur: Because in the library, we always do input first and output last.

peterrum (approved):

masterleinad:
What about an `ArrayView` overload instead? This way, we could also replace the `std::vector` overload.

Rombur:
> What about an ArrayView overload instead? This way, we could also replace the std::vector overload.

I don't have a strong opinion on this but using `ArrayView` creates an annoying circular dependency :disappointed:. I can do it though if you think that's the way to go.

masterleinad:
I think it's slightly better to have an ArrayView overload even though that makes the implementation a little more ugly. This patch works for me
```C++
diff --git a/include/deal.II/base/array_view.h b/include/deal.II/base/array_view.h
index 9a271eb63f..7c719e2e1f 100644
--- a/include/deal.II/base/array_view.h
+++ b/include/deal.II/base/array_view.h
@@ -31,6 +31,12 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+#ifndef DOXYGEN
+namespace MemorySpace
+{
+  struct Host;
+}
+#endif
 
 /**
  * A class that represents a window of memory locations of type @p ElementType
diff --git a/include/deal.II/base/cuda.h b/include/deal.II/base/cuda.h
index 0e6eafbb7b..eaf93d1dff 100644
--- a/include/deal.II/base/cuda.h
+++ b/include/deal.II/base/cuda.h
@@ -18,6 +18,7 @@
 
+#include <deal.II/base/array_view.h>
 #include <deal.II/base/exceptions.h>
 
 #ifdef DEAL_II_COMPILER_CUDA_AWARE
@@ -28,6 +29,7 @@
 #  include <vector>
 
 DEAL_II_NAMESPACE_OPEN
+
 namespace Utilities
 {
   /**
@@ -151,6 +153,37 @@ namespace Utilities
                                                cudaMemcpyHostToDevice);
       AssertCuda(cuda_error_code);
     }
+
+    /**
+     * Copy the elements in @p pointer_dev to the host in @p vector_host.
+     */
+    template <typename T>
+    inline void
+    copy_to_host(const T *                        pointer_dev,
+                 ArrayView<T, MemorySpace::Host> &vector_host)
+    {
+      cudaError_t cuda_error_code = cudaMemcpy(vector_host.data(),
+                                               pointer_dev,
+                                               vector_host.size() * sizeof(T),
+                                               cudaMemcpyDeviceToHost);
+      AssertCuda(cuda_error_code);
+    }
+
+    /**
+     * Copy the elements in @p vector_host to the device in @p pointer_dev. The
+     * memory needs to be allocate on the device before this function is called.
+     */
+    template <typename T>
+    inline void
+    copy_to_dev(const ArrayView<T, MemorySpace::Host> &vector_host,
+                T *                                    pointer_dev)
+    {
+      cudaError_t cuda_error_code = cudaMemcpy(pointer_dev,
+                                               vector_host.data(),
+                                               vector_host.size() * sizeof(T),
+                                               cudaMemcpyHostToDevice);
+      AssertCuda(cuda_error_code);
+    }
   } // namespace CUDA
 } // namespace Utilities
 
diff --git a/include/deal.II/base/memory_space.h b/include/deal.II/base/memory_space.h
index 56c6084498..e002fd4a92 100644
--- a/include/deal.II/base/memory_space.h
+++ b/include/deal.II/base/memory_space.h
@@ -19,7 +19,6 @@
 
-#include <deal.II/base/cuda.h>
 #include <deal.II/base/exceptions.h>
 
 #include <functional>
@@ -27,6 +26,20 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+
+#ifndef DOXYGEN
+namespace Utilities
+{
+  namespace CUDA
+  {
+    void
+    delete_device_data(Number *device_ptr) noexcept;
+  }
+} // namespace Utilities
+#endif
+
+
 /**
  */
 namespace MemorySpace
diff --git a/include/deal.II/lac/affine_constraints.templates.h b/include/deal.II/lac/affine_constraints.templates.h
index a9f0bfcc1e..bdb1a978f6 100644
--- a/include/deal.II/lac/affine_constraints.templates.h
+++ b/include/deal.II/lac/affine_constraints.templates.h
@@ -18,6 +18,7 @@
 
+#include <deal.II/base/cuda.h>
 #include <deal.II/base/cuda_size.h>
 #include <deal.II/base/memory_consumption.h>
 #include <deal.II/base/table.h>
```
Of course, we should also have a test for this or use it somewhere in the code base.

Rombur:
I have updated the PR. I broke the circular dependency by moving `MemorySpaceData` to its own file.

include/deal.II/base/cuda.h
```diff
@@ -124,19 +125,53 @@ namespace Utilities
...
+     * Copy the host ArrayView @p in to the device ArrayView @p out.
+     */
+    template <typename T>
+    inline void
+    copy_to_dev(const ArrayView<const T, MemorySpace::Host> &in,
+                ArrayView<T, MemorySpace::CUDA> &            out)
+    {
+      Assert(in.size() == out.size(),
+             ExcMessage(
+               "Input ArrayView and output ArrayView must have the same size"));
```
masterleinad, line 155:
```suggestion
      AssertDimension(in.size(), out.size());      
```

masterleinad (approved):
Looks good to me. Thanks!

PR #10964, opened 2020-09-24
Pass optional arguments to MGLevelObject

peterrum:


tests/multigrid/mg_level_object_args_01.cc
```diff
@@ -0,0 +1,41 @@
...
+using namespace dealii;
+
+int
+main()
+{
+  initlog();
+
+  Triangulation<2> tria;
+
+  MGLevelObject<DoFHandler<2>> dof_handlers(0, 3, tria, true);
```
peterrum, line 36: The use case are classes which do not have `reinit()` function. Like here.

tjhei:
This is quite useful. I vaguely remember some problems with compilers where default arguments and parameter packs interact in a weird way. Does anybody remember?

include/deal.II/base/mg_level_object.h
```diff
@@ -167,11 +176,13 @@ class MGLevelObject : public Subscriptor
...
 template <class Object>
+template <class... Args>
 MGLevelObject<Object>::MGLevelObject(const unsigned int min,
-                                     const unsigned int max)
+                                     const unsigned int max,
+                                     Args &&... args)
   : minlevel(0)
 {
-  resize(min, max);
+  resize(min, max, args...);
```
masterleinad, line 185:
```suggestion
  resize(min, max, std::forward<Args>(args)...);
```

include/deal.II/base/mg_level_object.h
```diff
@@ -207,7 +220,7 @@ MGLevelObject<Object>::resize(const unsigned int new_minlevel,
 
   minlevel = new_minlevel;
   for (unsigned int i = 0; i < new_maxlevel - new_minlevel + 1; ++i)
-    objects.push_back(std::make_shared<Object>());
+    objects.push_back(std::make_shared<Object>(args...));
```
masterleinad, line 223:
```suggestion
    objects.push_back(std::make_shared<Object>(std::forward<Args>(args)...));
```

masterleinad (commented):
It probably doesn't matter much here since the objects are small anyway, but let's do perfect forwarding just for consistency.

include/deal.II/base/mg_level_object.h
peterrum: I have applied the suggestion! Thanks.

masterleinad (approved):

include/deal.II/base/mg_level_object.h
```diff
@@ -207,7 +220,7 @@ MGLevelObject<Object>::resize(const unsigned int new_minlevel,
 
   minlevel = new_minlevel;
   for (unsigned int i = 0; i < new_maxlevel - new_minlevel + 1; ++i)
-    objects.push_back(std::make_shared<Object>());
+    objects.push_back(std::make_shared<Object>(std::forward<Args>(args)...));
```
bangerth, line 223: You could avoid one copy if you used `emplace_back`.

masterleinad: How is that? The `std::shared_ptr` is moved for both `push_back` and `emplace_back` (see https://en.cppreference.com/w/cpp/container/vector/push_back) or are you referring to something else?

peterrum: I don't have any preferences here.

kronbichler (approved):

include/deal.II/base/mg_level_object.h
masterleinad: Of course, we could also use `std::fill_n` with `std::back_inserter` here if we really want to make the intent clear that all of them are the same. But this solution is fine with me.

masterleinad: Ah, nevermind. We want separate `std::shared_ptr`s od course.

peterrum:
@bangerth It's up to you now!

include/deal.II/base/mg_level_object.h
bangerth: I don't matter much to me. We've tried to move away from `push_back` and use `emplace_back` instead, but you're right that it doesn't matter here because there's an overload of `push_back` that takes an rvalue reference.

bangerth (approved):

PR #10965, opened 2020-09-24
Enable tests with same name

peterrum:
In an application code, I had the issue that tests could not have had the same name even though they are separate folders. This PR fixes the issue by adding the prefix category/folder name at some places.


tamiko (approved):

tamiko:
Looks good!

PR #10970, opened 2020-09-25
Remove unnecessary instantiation files

kronbichler:
Due to #10927, we do not need to instantiate `matrix_free/evaluation_selector.cc` any more. Furthermore, since we do not use the degree templates in `VectorTools::project` any more, it is safe to compile `vector_tools_project.cc` with a single instantiation. This saves quite some time in compiling, aggregated over the full set of files. I checked that the memory consumption with gcc-7 is not more than 1.8G with the new single instantiation file for `vector_tools_project.cc` (and it compiles in 29 seconds for debug mode and 46 seconds for release mode).


masterleinad (approved):
Looks good to me.

peterrum (approved):

PR #10973, opened 2020-09-25
Implement BlockSparsityPattern::print_svg().

bangerth:
My Friday afternoon project.


include/deal.II/lac/block_sparsity_pattern.h
```diff
@@ -307,6 +307,14 @@ class BlockSparsityPatternBase : public Subscriptor
   void
   print_gnuplot(std::ostream &out) const;
 
+  /**
+   * Print the sparsity of the matrix in <tt>svg</tt> format. This is the same
+   * functionality implemented for usual sparsity patterns, see
+   * @ref SparsityPattern.
```
tjhei, line 313:
```suggestion
   * @ref SparsityPatternBase::print_svg().
```

tjhei (commented):
Have you considered moving this into a free-standing function in SparsityTools that could be used for any kind of sparsity pattern instead (block, dynamic, static, etc.)?
Also, I would prefer if the tests would produce smaller text output (a 5x5 matrix is big enough to test that things work).

bangerth:
I've reduced the size of the problems by a factor of 4 (for svg) and 16 (for gnuplot). I hope that's good enough.

As for generic function: The function is actually already generic, being in `template <typename SparsityPatternType>
class BlockSparsityPatternBase`. So it works for all block sparsity pattern (static, dynamic, etc). The existing function works for non-block sparsity patterns. But I can't make the function even more generic because block sparsity patterns don't have iterators that go over all entries.

tjhei (approved):

kronbichler (approved):

Issue #10797, opened 2020-08-05
simplex support comments


tjhei:
After @peterrum 's presentation yesterday, I have a few comments worth discussing. I would propose that we have an API round-table via video chat at some point soon. My notes so that I don't forget:

- [x] I would think ``cell->cell_type() `` might be neater than ``cell->reference_cell_type()`` (#11688)
- [x] Should we deprecate static info in GeometryInfo? Con: a lot of changes for users. Pro: consistency, bug finding. 
- [x] Should we consider deprecating the default mapping arguments? This needs a longer conversation, but I think I am in favor.
- [x] Change default mapping to something that is Q1 or P1 depending on the cells used and call is something like ``MappingLinear`` (I know that it is not strictly correct)?
- [x] Detect functions inside the library that rely on quads and either fix or add an ``Assert(tria->is_quads_only())``. This would involve (temporarily) removing GeometryInfo, MappingQ, etc..

bangerth:
Here's another thing I'd like to see: In classes like `QGauss` and `QTrapez`, the `Q` doesn't actually indicate a quad/hex style quadrature, but that it's a *q*uadrature.

So I don't think we should introduce `PGauss` and `PTrapez`. Rather, I'd suggest giving the constructor an argument of type `ReferenceCellType` (or whatever that should be) and by that either let it construct a tensor product rule or use whatever is the appropriate choice for that reference cell kind.

bangerth:
I can't seem to find any evidence of the `PTrapez` name mentioned above and assume that I just never looked it up.

drwells:
I think we have addressed all of these problems in various ways - we have `ReferenceCell`, many functions assert that the mesh is all hypercube elements, #10310, etc.

PR #10980, opened 2020-09-28
Rename Simplex::PGauss to Simplex::QGauss.

bangerth:
I mentioned this in #10797. I recognize that this is an incompatible change, but I'd rather change it now than when more people start to use it. The point is simply that the `Q` in `QGauss` never refered to the difference between "P" and "Q" spaces, but is just short for "Quadrature". 

Opinions welcome!

/rebuild

### DIFF
--- a/include/deal.II/simplex/quadrature_lib.h
+++ b/include/deal.II/simplex/quadrature_lib.h
@@ -38,13 +38,13 @@ namespace Simplex
    * @ingroup simplex
    */
   template <int dim>
-  class PGauss : public QSimplex<dim>
+  class QGauss : public QSimplex<dim>
   {
   public:
     /**
      * Constructor taking the number of quadrature points @p n_points.
      */
-    explicit PGauss(const unsigned int n_points);
+    explicit QGauss(const unsigned int n_points);
   };
 } // namespace Simplex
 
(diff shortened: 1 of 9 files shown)


masterleinad (approved):

peterrum:
@kronbichler If I remember you were not happy with the term *Gauss* in this context.

Furthermore, we need a better logic here:

```cpp
  static unsigned int                               X        = numbers::invalid_unsigned_int;
  static std::array<std::array<unsigned int, 3>, 4> n_points = {{
    {X, X, X}, /* 0-D */ {X, 2, 3}, /* 1-D */ {X, 3, 7}, /* 2-D */ {X, 4, 10} // 3-D }};

  // setup finite element
  Simplex::FE_P<dim, spacedim>    fe1(degree);
  FE_Q<dim, spacedim>             fe2(degree);

  // setup quadrature rules
  Simplex::QGauss<dim> quad(n_points[dim][degree]);
  Simplex::QGauss<dim - 1> face_quad(n_points[dim - 1][degree]);
```
I guess we could pass into the constructor `degree+1` as in the hypercube case?

kronbichler (approved):
I agree with the name here at this point. While I do not think those formulas are genuinely *Gauss* rules (which I intuitively link to the Gauss-Legendre rules if nothing else is set) but rather some generic quadrature rules on the simplex, I do not have better names as we have *quadrature* already in the name by the `Q`. And many people use Gauss rule and quadrature as synonyms anyway, so no real harm here.

peterrum (approved):

PR #10981, opened 2020-09-29
Deprecate QTrapez and rename it to QTrapezoid

bangerth:
This is really just a 23-year old mistake: In germany, it's the "Trapezregel" and one of us three at the time (Guido, Ralf, or myself) translated it to the "trapez rule", which then became `QTrapez`. But it should always have been `QTrapezoid`, I just never had the courage to fix this.

Here's a patch for it. What do you all think? The class isn't widely used in the (common) tutorials, which makes things a bit easier. But it's still an incompatible change.


drwells:
This has been a long time coming. I don't feel that strongly about it but its probably easier to use the full English name instead of an abbreviation.

drwells (approved):

source/base/quadrature_selector.cc
```diff
@@ -45,7 +45,7 @@ QuadratureSelector<dim>::create_quadrature(const std::string &s,
       else if (s == "simpson")
         return QSimpson<dim>();
       else if (s == "trapez")
```
tjhei, line 47: shouldn't we add "trapezoidal" here as well? Also please leave a comment DEAL_II_DEPRECATE so we can grep and remove later.

include/deal.II/fe/fe_tools.templates.h
```diff
@@ -2622,11 +2622,11 @@ namespace FETools
                     // find sub-quadrature
                     position = name.find('(');
                     const std::string subquadrature_name(name, 0, position);
-                    AssertThrow(subquadrature_name.compare("QTrapez") == 0,
```
tjhei, line 2625: I think this needs more work because the name could be QTrapez or QTrapezoid here, right? (assuming somebody has an FE name in a .prm file.

tjhei: Also: is the string "QTrapez" generated somewhere in the code? I could not find it.

tjhei (changes requested):

tjhei:
In favor of this change, but I have some questions.

tjhei:
Can you also update contrib/utilities/dotgdbinit.py ?

include/deal.II/fe/fe_tools.templates.h
bangerth: Yes, it requires a couple more lines. Fixed.

bangerth:
Thanks, @tjhei , for the careful review. I've addressed all of the issues you pointed out.

include/deal.II/base/quadrature_lib.h
```diff
@@ -108,20 +108,33 @@ class QSimpson : public Quadrature<dim>
...
+/**
+ * An alias for QTrapezoid available for historic reasons. This name is
+ * deprecated.
  *
- * The class is poorly named since the proper name of the quadrature formula
+ * The class was originally named QTrapez, a poorly named choice since the
+ * proper name of the quadrature formula
  * is "trapezoidal rule", or sometimes also called the "trapezoid rule". The
- * misnomer results from the fact that its original authors' poor English
+ * misnomer resulted from the fact that its original authors' poor English
```
peterrum, line 132: I don't see the point to keep these lines (although I must admit I really enjoyed reading them a while ago when I first saw them).

doc/news/changes/incompatibilities/20200928Bangerth
```diff
@@ -0,0 +1,5 @@
+Deprecated: The QTrapez class, poorly named because the proper English
+term is "trapezoidal quadrature rule" has been renamed to QTrapezoid,
```
kronbichler, line 2:
```suggestion
term is "trapezoidal quadrature rule", has been renamed to QTrapezoid,
```

doc/news/changes/incompatibilities/20200928Bangerth
```diff
@@ -0,0 +1,5 @@
...
+and the old name has been deprecated.
```
kronbichler, line 3:
```suggestion
and the class with the old name has been deprecated.
```

kronbichler (commented):
I agree, apart from two clarifications I found plus the comment suggested by @peterrum

doc/news/changes/incompatibilities/20200928Bangerth
bangerth: That's not actually quite true: The class with the old name simply doesn't exist any more, I just introduced an alias (a "name") that is now deprecated.

But that's a subtlety probably lost on our readers :-) So I went with your suggestion :-)

bangerth:
I addressed @kronbichler 's suggestions.

kronbichler (approved):

source/base/quadrature_selector.cc
tjhei: can you leave a comment here that contains "DEAL_II_DEPRECATED" so we can grep for this in the future?

bangerth: So happened!

tjhei (approved):

Issue #10969, opened 2020-09-25
Faster alternative to Mapping::transform_real_to_unit_cell


kronbichler:
For particle simulations, but also in other occasions, we need a better way to compute the transformation from real to unit cell, especially in case many points on a cell are requested. A generic Newton iteration is very slow and needs to query the mapping often; the latter is also typically not particularly optimized. Related to #10968, we should try to specialize some of the functionality to evaluate the underlying polynomials and aggregate the mapping support points (in case of `MappingQGeneric`) over operations on multiple points.

This is probably not something that must be part of the `Mapping` base class. I rather think of an additional class that could help out `ParticleHandler` and other expensive things. I have already tested with good success an algorithm that constructs polynomial approximations of the inverse function from real to unit cell on the *bounding box* of a cell. If the cell is not too deformed and with polynomial degrees of four to six, the accuracy can be as good as 1e-6 to 1e-8. If we are still unhappy with that value, we can use a single Newton iteration to get to roundoff accuracy. This can be super fast for `MappingQGeneric` if we only have a really efficient polynomial evaluation, i.e., #10968.

kronbichler:
Given the work in #10983, I think it would be enough to simply have an interface that transforms not just a single point at a time, but many points we believe sit in the cell at the same time. That would amortize the only memory allocation left after #10983 (and expensive operation in its own), the computation of the mapping support points.

Issue #10968, opened 2020-09-25
Interface for fast evaluation of function value at arbitrary point


kronbichler:
Our generic way of evaluating quantities derived from the solution at some arbitrary location in reference coordinates of a cell is to
- set up a `Quadrature` object with the selected points,
- set up an `FEValues` object with the `Mapping` and `FiniteElement` in question as well as the quadrature formula, and use say `update_values` as flag, and
- call `FEvalues::reinit(cell)` and then retrieve the desired information.

While this certainly works, it is inefficient for cases where the points differ from one cell to the other, like with particle simulations, and especially for higher order shape functions or mappings (degree 3 or higher). By inefficient I mean that things take 10-100 times as much time as they would need to. One observation I made in the context of particles is here: https://github.com/dealii/dealii/pull/10942#discussion_r492061699.

This issue is to collect ideas for a better interface for those cases and eventually implement something fast:

- I think one prerequisite is to avoid memory allocations on a cell, i.e., the computation should either need no scratch data or all scratch data should only depend on the shape functions and mapping but not on the number of points, so they could be re-used on several cells.

- The easy case is the computation of values in elements where no information from the mapping is needed, i.e., no Piola transform or the like. What would allow me to write a quick evaluation is a `virtual` function `double FiniteElement::evaluate_scalar_value(const ArrayView<double> &local_dof_values, const unsigned int component_index) const` and possibly `Tensor<1,dim> FiniteElement::evaluate_vector_value(const ArrayView<double> &local_dof_values, const unsigned int component_index) const`. Here, the `local_dof_values` would be a vector of length `fe.dofs_per_cell` that is obtained by reading out from a global vector. This implementation could be specialized by `FE_Poly` and some `FESystem` of `FEPoly` to step into the tensor product polynomials and only do the 1D evaluation the minimal number of times. The combination with the 2D/3D unknown data could then be vectorized and augmented with all kinds of tricks. Note that the currently fastest way exposed to the user is something similar to this:
https://github.com/dealii/dealii/blob/cd8b20072f3da26ac6924477dbf4af8909ff64f0/examples/step-19/step-19.cc#L474-L478
but with a summation over the solution coefficients rather than the testing by test functions. In this code, we need to evaluate the 1D shape functions in a tensor product far too often, and we need to step into virtual functions.

[...]

PR #10983, opened 2020-09-30
Speed up MappingQGeneric::transform_real_to_unit_cell

kronbichler:
`MappingQGeneric::transform_real_to_unit_cell()` has been made much faster by directly working with the tensor product form of the mapping shape functions and avoiding many unnecessary memory allocations. The main cost is now `MappingQGeneric::compute_mapping_support_points()`, which can be made fast with `MappingQCache`, for example.

To achieve this goal, I added some useful additional member variables to `MappingQGeneric`, namely the 1D polynomials and the renumbering between the lexicographic numbers and the ones in the mapping support points. They are very small so it is fine to keep them around.

Related to #10969, using some of the functions proposed in #10968.

### DIFF
--- a/source/fe/mapping_q_generic.cc
+++ b/source/fe/mapping_q_generic.cc
@@ -1165,7 +1165,54 @@ namespace internal
         const std::vector<unsigned int> &                   renumber,
         const Point<dim> &                                  p)
       {
+        static_assert(dim >= 1 && dim <= 3, "Only dim=1,2,3 implemented");
+
         const unsigned int n_shapes = poly.size();
+
+        // shortcut for linear interpolation to speed up evaluation
+        if (n_shapes == 2)
+          {
+            if (dim == 1)
+              {
+                Tensor<2, spacedim> derivative;
+                derivative[0] = points[1] - points[0];
+                return std::make_pair((1. - p[0]) * points[0] +
+                                        p[0] * points[1],
+                                      derivative);
+              }
+            else if (dim == 2)
+              {
+                const double          x0 = 1. - p[0], x1 = p[0];
+                const Point<spacedim> tmp0   = x0 * points[0] + x1 * points[1];
+                const Point<spacedim> tmp1   = x0 * points[2] + x1 * points[3];
+                const Point<spacedim> mapped = (1. - p[1]) * tmp0 + p[1] * tmp1;
+                Tensor<2, spacedim>   derivative;
+                derivative[0] = (1. - p[1]) * (points[1] - points[0]) +
+                                p[1] * (points[3] - points[2]);
+                derivative[1] = tmp1 - tmp0;
+                return std::make_pair(mapped, transpose(derivative));
+              }
+            else if (dim == 3)
+              {
+                const double x0 = 1. - p[0], x1 = p[0], y0 = 1. - p[1],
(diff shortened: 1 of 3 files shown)


source/fe/mapping_q_generic.cc
```diff
@@ -1148,27 +1148,113 @@ namespace internal
...
+       * derivative almost for free (less than 2x the cost of simply the
+       * values), so we always compute it.
+       */
+      template <int dim, int spacedim>
+      std::pair<Point<spacedim>, Tensor<2, spacedim>>
+      compute_mapped_location_of_point(
+        const std::vector<Point<spacedim>> &                points,
+        const std::vector<Polynomials::Polynomial<double>> &poly,
+        const std::vector<unsigned int> &                   renumber,
+        const Point<dim> &                                  p)
```
kronbichler, line 1166: This is the central function that implements the fast evaluation of the `FE_Q` interpolation in a point (including its derivative). This could be generalized for all evaluations of tensor product interpolations if we find an appropriate function interface as to be discussed in #10968, so the implementation could move somewhere else once we found a strategy.

kronbichler:
As a point of reference, the full Newton iteration on a 3D `MappingQGeneric` of degree 3 with the new code (using on average 2 Newton iterations) takes only around 3.5 times the number of instructions of `TriaAccessor::real_to_unit_cell_affine_approximation`. Of course, the affine approximation of the inverse takes maybe 4 times as many instructions as it would need to because we still need to address #5437, but yet I think the code that does evaluates the polynomial basis is super fast now.

source/fe/mapping_q_generic.cc
```diff
@@ -2333,82 +2449,60 @@ MappingQGeneric<dim, spacedim>::transform_real_to_unit_cell_internal(
...
-  auto mdata = Utilities::dynamic_unique_cast<InternalData>(
-    get_data(update_flags, point_quadrature));
-
-  mdata->mapping_support_points = this->compute_mapping_support_points(cell);
-
-  // dispatch to the various specializations for spacedim=dim,
-  // spacedim=dim+1, etc
   return internal::MappingQGenericImplementation::
-    do_transform_real_to_unit_cell_internal<2>(cell, p, initial_p_unit, *mdata);
+    do_transform_real_to_unit_cell_internal<2>(
```
tjhei, line 2482: I like how much simpler this is!

tjhei:
I only had a quick look, but this is great.

source/fe/mapping_q_generic.cc
```diff
@@ -1148,27 +1148,160 @@ namespace internal
...
+                                      y1 * (points[3] - points[2])) +
+                                z1 * (y0 * (points[5] - points[4]) +
+                                      y1 * (points[7] - points[6]));
+                return std::make_pair(mapped, transpose(derivative));
+              }
+          }
+
+        // Put up to 32 shape functions per dimension on stack, else on heap
+        double *            shapes;
+        double              shapes_stack[64 * dim];
```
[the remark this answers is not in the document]
kronbichler: Definitely, that makes sense. I will change that once I have half an hour to look at things closely.

kronbichler:
I addressed @tjhei's comments.

peterrum (approved):

tjhei (approved):

PR #10984, opened 2020-09-30
step-48: Reset ghost values in vector after output

kronbichler:
When running step-48 with MPI but no threads, the smart loop scheduling approach would get confused when the vector to be written already contains ghost values, working with trash data in the constrained unknowns. We need to make sure that the `output_results` function that calls `update_ghost_values` returns the vector in a non-ghosted state.


peterrum (approved):

tjhei:
> would get confused when the vector to be written already contains ghost values, working with trash data in the constrained unknowns.

When you say "working with trash data", are you saying that no assertion is being triggered? If not, can this be fixed?

kronbichler:
In debug mode, we indeed trigger an assertion. The trash is for release mode (that I tried first and thus guided my explanation).

PR #10987, opened 2020-10-01
Make accessor functions of particles inline

kronbichler:
Looking at a profile of `step-68` I realized that all accessor functions of particles (which often are a single line of code) were pre-compiled in the `.cc` files. We typically place the implementations of those in header files in deal.II. This does not only increase the performance of the generated code (in release mode), but for small functions it also reduces code size because it avoids the function call prologue and epilogue.


include/deal.II/particles/particle.h
```diff
@@ -486,6 +488,91 @@ namespace Particles
     if (n_properties > 0)
       ar &boost::serialization::make_array(properties, n_properties);
   }
+
+
+
+  template <int dim, int spacedim>
+  void
+  Particle<dim, spacedim>::set_location(const Point<spacedim> &new_loc)
```
kronbichler, line 496: While not technically necessary due to the template, I wanted to hear opinions whether I should add the `inline` keyword here and in all other places. I think we most often add the keyword, but I am not sure whether we aim to provide it everywhere.

peterrum (approved):

peterrum:
@gassmoeller Do you happen to work currently on these files? If not we can probably without causing too much pain.

blaisb:
> Looking at a profile of `step-68` I realized that all accessor functions of particles (which often are a single line of code) were pre-compiled in the `.cc` files. We typically place the implementations of those in header files in deal.II. This does not only increase the performance of the generated code (in release mode), but for small functions it also reduces code size because it avoids the function call prologue and epilogue.

Yes! This will really help us for our DEM code. There are still some bottlenecks with particles we are trying to find and optimize. This could help slightly at least :).

bangerth:
> While not technically necessary due to the template, I wanted to hear opinions 
> whether I should add the |inline| keyword here and in all other places. I 
> think we most often add the keyword, but I am not sure whether we aim to 
> provide it everywhere.

We generally do, so let's be consistent.

kronbichler:
> We generally do, so let's be consistent.

I added the `inline` keyword.

bangerth (approved):

PR #10990, opened 2020-10-01
Replace TaskFlow 2.5 by TaskFlow 2.7.

### LINKED ISSUE #10989: Import TaskFlow 2.7.0
TaskFlow 2.7.0 is out, and it contains the new `async` call we will want to use in `Threads::new_task`. We should import this new version.

bangerth:
Fixes #10989 . Let's see whether this works :-)

/rebuild


bangerth:
Hm, it fails in quick_tests because of this piece of code:
```
  auto p =
    taskflow.parallel_for(1, 11, 1, [&](int idx) { counter.fetch_add(idx); });
```
Apparently that function no longer exists:
```
/jenkins/workspace/dealii_PR-10990/tests/quick_tests/taskflow.cc:52:14: error: ‘class tf::Taskflow’ has no member named ‘parallel_for’
     taskflow.parallel_for(1, 11, 1, [&](int idx) { counter.fetch_add(idx); });
              ^
```
What do we do?

bangerth:
Let's try this again.

tjhei:
We probably want to adjust the minimum version as well: https://github.com/dealii/dealii/blob/master/cmake/configure/configure_1_taskflow.cmake

bangerth:
I think I'm going to need some help with this one since I can't reproduce the failure in the quicktest that trips up the testers :-(

bangerth:
On 10/2/20 10:41 AM, Timo Heister wrote:
> We probably want to adjust the minimum version as well: 

Good point -- I'm now requiring 2.7. Thanks for remembering that place!

marcfehling:
> Hm, it fails in quick_tests because of this piece of code:
> 
> ```
>   auto p =
>     taskflow.parallel_for(1, 11, 1, [&](int idx) { counter.fetch_add(idx); });
> ```
> 
> Apparently that function no longer exists:
> 
> ```
> /jenkins/workspace/dealii_PR-10990/tests/quick_tests/taskflow.cc:52:14: error: ‘class tf::Taskflow’ has no member named ‘parallel_for’
>      taskflow.parallel_for(1, 11, 1, [&](int idx) { counter.fetch_add(idx); });
>               ^
> ```
> 
> What do we do?

Looks like `parallel_for` has been replaced by `for_each`/`for_each_index`. See https://github.com/taskflow/taskflow/commit/56443435ff1c8167bdfbe2481c4170325d751b6b

marcfehling:
I cannot reproduce the error either. The only differences I could spot right now between my notebook and the testers are:
- number of threads: 4 (notebook) / 12 (testers)
- gcc version: 10.2.0 (notebook) / 5.4.0 (testers)

The [assertion](https://github.com/taskflow/taskflow/blob/c75c05f3d549534e8bd05d6e9afdac82829ea786/taskflow/core/executor.hpp#L976) that will be triggered by the testers has been introduced in https://github.com/taskflow/taskflow/commit/8e1b019f6d06b3d649e315ec4edead0d1cb9500d. Maybe that's a starting point?

bundled/taskflow-2.7.0/.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,9 @@
+---
```
tjhei, line 1: I don't think we need the files in this folder

tests/quick_tests/taskflow.cc
```diff
@@ -49,9 +49,9 @@ test1()
   B.precede(C);
 
   auto p =
-    taskflow.parallel_for(1, 11, 1, [&](int idx) { counter.fetch_add(idx); });
+    taskflow.for_each_index(1, 11, 1, [&](int idx) { counter.fetch_add(idx); });
 
-  C.precede(p.first);
+  C.precede(p);
```
tjhei, line 54: I am not sure why this fails but the dependency is a bit weird. Can you change this to 
```suggestion
  p.precede(C);
```

tjhei:
> Good point -- I'm now requiring 2.7. Thanks for remembering that place!

I don't see this change in your PR.

tests/quick_tests/taskflow.cc
bangerth: The dependencies are as follows (making this clear to myself as much as to you):
* All tasks operate on the `counter` variable, which is initialized to zero.
* Tasks `A`, `B` increment `counter` by one.
* Task `C` resets the value to zero.
* `A` and `B` both precede `C`, so they count `counter` up to one, then two, and `C` resets everything to zero.
* Task `p` adds `idx` to `counter` for each `idx` between 1 and 11
* Task `C` *precedes* `p`, so what `p` does it start its sum computation from a zero state, resulting in 55 as asserted below.

In other words, the statement here looks correct in view of the assertion below. If we were to make `p` *precede* `C`, we would get some something from tasks `A`, `B`, and `p`, that is then reset to zero by task `C`. So I think that the line as stated is correct -- and also matches what you had in the previous version.

bangerth:
I forgot to push that last patch. I've done so now, plus removed the unnecessary files I had not even noticed.

Let's just see what happens if we run the testers again :-)

tests/quick_tests/taskflow.cc
tjhei: You are right. One thing that might be problematic is the dependency to p. If you look at the generated graph
a dia
I am not sure if the individual pfg tasks have to wait to start or not.

Maybe we should simplify this test?

tjhei: I take that back: I am pretty sure that what we are doing is legal.

tjhei:
I don't understand the bug: even running on the same machine as the automated tester in the same docker container with the same compiler, it does not fail for me.

bangerth:
I rebased. Let's just see whether that problem is reproducible if we run the testers again, or whether it was just a cosmic ray hitting your CPU.

tjhei:
It seems to be. I just realized that we compile with -O1 in debug mode. I will retry my experiment...

bangerth:
Hm, it still fails. Would be happy to hear what you find in your experiments!

tjhei:
> Hm, it still fails. Would be happy to hear what you find in your experiments!

I can reproduce it now. A standalone version of the test without deal.II works fine and it is only triggered when using -O1. I think something weird is happening with the static Executor that is allocated inside deal.II and returned by ``MultithreadInfo::get_taskflow_executor``.
Maybe some static initialization is going terribly wrong?

bangerth:
Hm, I pushed a commit that works around what I thought should fix the problem, but now not only the one taskflow test, but also the lapack test fails. This confuses me.

@tjhei -- can you try using 103e5ae for the testcase you have that breaks taskflow?

bangerth:
I can confirm that that commit fixes your testcase on my system: With stock taskflow 2.7, the testcase aborts. With my commit, it succeeds.

bangerth:
See also https://github.com/taskflow/taskflow/issues/245

bangerth:
@tjhei Can you try with the branch as I have it here? You could reproduce the failures from the tester machine, whereas all of these quicktests seem to succeed for me now.

PR #10991, opened 2020-10-02
Let mapping transform multiple points from real to unit cell

### LINKED ISSUE #10969: Faster alternative to Mapping::transform_real_to_unit_cell

kronbichler:
Fixes #10969. As discussed in that issue, we can accelerate the `Mapping::transform_real_to_unit_cell` quite a bit if we compute multiple points at a time. The most important factor is `compute_mapping_support_points()` for higher order mappings. But we even gain when compared to the analytic inversion function for `MappingQ1` in 2d; partly because the collection of cell vertices is so slow, but also because the evaluation implemented in #10983 is so fast (unless we need more than 2 Newton iterations which is unlikely).

I implemented it within the `ParticleHandler::sort_particles_into_subdomains_and_cells()` and can confirm that it runs much faster in the case we have many particles per cell. The transformation indeed runs so fast that the other operations in that functions take much more time than the computation of reference positions. In fact, it can take more time to compute the individual positions of particles that left the cell and gets picked up in the surrounding cells (I tested in step-68 but it of course depends on the time step size and the fraction of particles leaving a cell in a time step). This latter part can surely be optimized as well to collect candidates from the cell.

~~This PR is right now based on #10983 and also some other function in `GridTools` that I want to break out into a separate PR, but I wanted to post this now already for discussion.~~

### DIFF
--- a/include/deal.II/fe/mapping_q_generic.h
+++ b/include/deal.II/fe/mapping_q_generic.h
@@ -181,6 +181,13 @@ public:
     const typename Triangulation<dim, spacedim>::cell_iterator &cell,
     const Point<spacedim> &p) const override;
 
+  // for documentation, see the Mapping base class
+  virtual void
+  transform_points_real_to_unit_cell(
+    const typename Triangulation<dim, spacedim>::cell_iterator &cell,
+    const ArrayView<const Point<spacedim>> &                    real_points,
+    ArrayView<Point<dim>> &unit_points) const override;
+
   /**
    * @}
    */
--- a/source/fe/mapping_q_generic.cc
+++ b/source/fe/mapping_q_generic.cc
@@ -29,6 +29,7 @@
 #include <deal.II/fe/mapping_q1.h>
 #include <deal.II/fe/mapping_q_generic.h>
 
+#include <deal.II/grid/grid_tools.h>
 #include <deal.II/grid/manifold_lib.h>
 #include <deal.II/grid/tria.h>
 #include <deal.II/grid/tria_iterator.h>
@@ -893,17 +894,15 @@ namespace internal
       /**
        * Implementation of transform_real_to_unit_cell for dim==spacedim
        */
-      template <int dim>
+      template <int dim, int spacedim>
       Point<dim>
       do_transform_real_to_unit_cell_internal(
-        const Point<dim> &                                  p,
+        const Point<spacedim> &                             p,
         const Point<dim> &                                  initial_p_unit,
-        const std::vector<Point<dim>> &                     points,
+        const std::vector<Point<spacedim>> &                points,
         const std::vector<Polynomials::Polynomial<double>> &polynomials_1d,
         const std::vector<unsigned int> &                   renumber)
       {
-        const unsigned int spacedim = dim;
-
         AssertDimension(points.size(),
                         Utilities::pow(polynomials_1d.size(), dim));
 
@@ -917,8 +916,8 @@ namespace internal
         // The shape values and derivatives of the mapping at this point are
         // previously computed.
 
-        Point<dim>                            p_unit = initial_p_unit;
-        std::pair<Point<dim>, Tensor<2, dim>> p_real =
+        Point<dim>                                      p_unit = initial_p_unit;
+        std::pair<Point<spacedim>, Tensor<2, spacedim>> p_real =
           compute_mapped_location_of_point(points,
                                            polynomials_1d,
                                            renumber,
@@ -967,6 +966,9 @@ namespace internal
         const double       eps                    = 1.e-11;
         const unsigned int newton_iteration_limit = 20;
 
+        Point<dim> invalid_point;
+        invalid_point[0] = std::numeric_limits<double>::infinity();
+
         unsigned int newton_iteration = 0;
         double       last_f_weighted_norm_square;
         do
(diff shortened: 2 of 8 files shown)


kronbichler:
The interesting commits are the last two, the rest will go away once #10983 is merged (and I rebased) and when I have the other PR done.

kronbichler:
One more comment: The computation of the transformation from real to unit cell with many points could be vectorized across multiple points easily, so if we need another factor of 3-5 (on AVX-512 targets, a bit less on older machines) I know how to get them.

include/deal.II/fe/mapping_q_generic.h
```diff
@@ -180,6 +181,13 @@ class MappingQGeneric : public Mapping<dim, spacedim>
     const typename Triangulation<dim, spacedim>::cell_iterator &cell,
     const Point<spacedim> &p) const override;
 
+  // for documentation, see the Mapping base class
```
peterrum, line 184:
```suggestion
  /**
   * @copydoc Mapping::transform_points_real_to_unit_cell()
   */
```

include/deal.II/grid/grid_tools.h
```diff
@@ -247,6 +247,29 @@ namespace GridTools
...
+   * or for Cartesian and affine (parallelogram) meshes in 2D/3D.
+
+   * This approximation is underlying the function
+   * TriaAccessor::real_to_unit_cell_affine_approximation() function.
+   *
+   * For exact transformations to the unit cell, use
+   * Mapping::transform_real_to_unit_cell().
+   */
+  template <int structdim, int spacedim>
+  std::pair<DerivativeForm<1, structdim, spacedim>, Tensor<1, spacedim>>
```
peterrum, line 270: What is the first and the second component of the pair?

source/fe/mapping_q_generic.cc
```diff
@@ -2577,46 +2668,70 @@ MappingQGeneric<dim, spacedim>::transform_real_to_unit_cell(
...
+  const ArrayView<const Point<spacedim>> &                    real_points,
+  ArrayView<Point<dim>> &                                     unit_points) const
+{
+  AssertDimension(real_points.size(), unit_points.size());
+  const std::vector<Point<spacedim>> support_points =
+    this->compute_mapping_support_points(cell);
+  const std::pair<DerivativeForm<1, dim, spacedim>, Tensor<1, spacedim>>
+    affine_factors = GridTools::affine_cell_approximation<dim>(
+      ArrayView<const Point<spacedim>>(support_points.data(),
+                                       GeometryInfo<dim>::vertices_per_cell));
```
peterrum, line 2714:
```suggestion
      ArrayView<const Point<spacedim>>(support_points));
```

source/fe/mapping_q_generic.cc
```diff
@@ -2577,46 +2668,70 @@ MappingQGeneric<dim, spacedim>::transform_real_to_unit_cell(
...
-      return this->transform_real_to_unit_cell_internal(cell,
-                                                        p,
-                                                        initial_p_unit);
+      try
+        {
+          unit_points[i] = internal::MappingQGenericImplementation::
+            do_transform_real_to_unit_cell_internal<dim, spacedim>(
+              real_points[i],
+              Point<dim>(apply_transformation(
+                A_inv, real_points[i] - affine_factors.second)),
```
peterrum, line 2726: Could you add a comment here?

source/particles/particle_handler.cc
```diff
@@ -995,29 +995,38 @@ namespace Particles
...
-    for (particle_iterator it = begin(); it != end(); ++it)
+    std::vector<Point<spacedim>> real_locations;
+    std::vector<Point<dim>>      reference_locations;
+    for (auto particle = begin(); particle != end();)
       {
-        const typename Triangulation<dim, spacedim>::cell_iterator cell =
-          it->get_surrounding_cell(*triangulation);
-
-        try
+        const auto cell = particle->get_surrounding_cell(*triangulation);
```
peterrum, line 1002: :)

peterrum (commented):
Very nice!

include/deal.II/fe/mapping_q_generic.h
kronbichler: I believe we do not need those because of #10914.

include/deal.II/grid/grid_tools.h
kronbichler: I will fix that in #10992.

source/fe/mapping_q_generic.cc
kronbichler: I see that this needs a comment. I intentionally only pick the first `2^dim` points to construct a linear mapping and ignore the rest. After all, this only serves for the initial guess.

peterrum (approved):

kronbichler:
I added a test case and changelog entry, so this should be ready for merging from my end.

masterleinad (commented):
Looks good apart from the one small question.

source/fe/mapping_q_generic.cc
```diff
@@ -893,17 +894,15 @@ namespace internal
       /**
        * Implementation of transform_real_to_unit_cell for dim==spacedim
```
rezarastak, line 895: I think this function docstring is getting out-dated.

peterrum (approved):
Thanks: all my questions have been answered!

PR #10992, opened 2020-10-02
Allow to compute affine approximation on arbitrary set of points

kronbichler:
This is the function to compute the affine approximation promised in #10991. Before, we had this as part of `TriaAccessor::real_to_unit_cell_affine_approximation` but it really is a generic concept that can be done on any pair of cells. One thing to note is that we might want to generalize this for simplex (or wedge?) elements; but that would require additional work because the transformation matrix needs a different formula.


kronbichler:
This build error on `OSX` is, well, annoying:
```
 /Users/runner/work/dealii/dealii/source/grid/grid_tools.cc:255:41: error: unused variable 'KA' [-Werror,-Wunused-const-variable]
    const double TransformR2UAffine<3>::KA[GeometryInfo<3>::vertices_per_cell]
                                        ^
/Users/runner/work/dealii/dealii/source/grid/grid_tools.cc:270:41: error: unused variable 'Kb' [-Werror,-Wunused-const-variable]
    const double TransformR2UAffine<3>::Kb[GeometryInfo<3>::vertices_per_cell] =
                                        ^
2 errors generated.
```
This error comes from `grid_tools.cc` because, as far as I can see, we split the instantiation files and we happen to not instantiate the selected variables for the `affine_cell_approximation` in `dim=3` in that particular file. Opinions? I tend towards not splitting up that file or, better, split up the implementations into two separate files without the instantiation tricks.

kronbichler:
I now fixed the comment remarked in https://github.com/dealii/dealii/pull/10991#discussion_r498859111. Furthermore, I now removed the instantiation split of `grid_tools.cc` to test my hypothesis about the missing instantiation in the OSX build. If it works, we might either take the code as is or split in a different (smarter) way.

include/deal.II/grid/grid_tools.h
```diff
@@ -247,6 +247,32 @@ namespace GridTools
...
   cell_measure(const T &, ...);
 
+  /**
+   * This function computes an affine approximation of the map from the unit
+   * coordinates to the real coordinates of the form $p_\text{real} = A
+   * p_\text{unit} + b $ by a least squares fit of this affine function to the
+   * $2^\text{dim}$ vertices representing a quadrilateral or hexahedral cell
+   * in `spacedim` dimensions. The result is returned as a pair with the
+   * matrix <i>A</i> as the first argument and the vector <i>b</i> describing
+   * distance of the plane to the origin.
```
peterrum, line 257: Thanks!

include/deal.II/grid/grid_tools.h
```diff
@@ -247,6 +247,32 @@ namespace GridTools
...
+   *
+   * This approximation is underlying the function
+   * TriaAccessor::real_to_unit_cell_affine_approximation() function.
+   *
+   * For exact transformations to the unit cell, use
+   * Mapping::transform_real_to_unit_cell().
+   */
+  template <int dim, int spacedim>
+  std::pair<DerivativeForm<1, dim, spacedim>, Tensor<1, spacedim>>
+  affine_cell_approximation(const ArrayView<const Point<spacedim>> &vertices);
```
peterrum, line 274: This is a new public function. Do you want to add a change-log entry?

peterrum (approved):

include/deal.II/grid/grid_tools.h
kronbichler: Added a changelog.

peterrum (approved):

PR #10994, opened 2020-10-02
Slightly improve evaluation of polynomial

kronbichler:
When looking at something else, I realized that the compiler stores `values[0], ..., values[n_derivatives]` in every inner loop iteration to memory, rather than just the final result. This is because the compiler cannot know that the polynomial coefficients and the Lagrange weight do not alias with the result array `values`. It is easy to work around (without increasing the number of code lines, apart from the braces `{}` to keep variables local), so do it for efficiency reasons.

### DIFF
--- a/source/base/polynomial.cc
+++ b/source/base/polynomial.cc
@@ -119,6 +119,7 @@ namespace Polynomials
         // form (x-x_1)*(x-x_2)*...*(x-x_n), expand the derivatives like
         // automatic differentiation does.
         const unsigned int n_supp = lagrange_support_points.size();
+        const number       weight = lagrange_weight;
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
Looks good to me!

tjhei (approved):

kronbichler:
As a side remark, if we ever have concerns about polynomial evaluation we should consider a new interface for a `PolynomialBasis` evaluating *all* basis functions within a single call, rather than one basis function at a time. Basis like the Legendre basis or the Lagrange basis have `O(n)` evaluation cost for `n` polynomials when working with the full basis, whereas computing each basis function separately is `O(n^2)`.

PR #10997, opened 2020-10-02
Check for SEACAS support.

drwells:
We need it for ExodusII.

@tamiko approved the CMake changes in #10899 so we might as well merge it in.


cmake/configure/configure_2_trilinos.cmake
```diff
@@ -281,6 +281,43 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
...
+      CHECK_CXX_SOURCE_COMPILES(
+        "
+        #include <exodusII.h>
+        int
+        main()
+        {
+          int component_word_size = sizeof(double);
+          int floating_point_word_size = 0;
+          float ex_version = 0;
+          const int ex_id = ex_open(\"test.ex\",
```
bangerth, line 298: I wished Trilinos would pay more attention to namespacing :-(

cmake/configure/configure_2_trilinos.cmake
```diff
@@ -171,7 +171,7 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
       #
       # Check for modules.
       #
-      FOREACH(_optional_module EpetraExt ROL Sacado Tpetra MueLu Zoltan)
+      FOREACH(_optional_module EpetraExt MueLu ROL Sacado SEACAS Tpetra Zoltan)
```
bangerth, line 174: Does this mean that SEACAS is not a *required* component of Trilinos? Do we need to adjust things such as candi and our installation documentation, or is SEACAS a default component of Trilinos?

cmake/configure/configure_2_trilinos.cmake
drwells: To be fair Exodus, AFAICT, very much predates Trilinos (some of the technical documents are from the early 1990s).

drwells:
Let's not split this off from #10899 now that that PR is feature-complete.

PR #10998, opened 2020-10-02
Use the correct English terminology: factorial.

bangerth:
Seen while looking at #10994. The correct word for the term is "factorial", not "faculty" as in German.


drwells (approved):

jppelteret (approved):

kronbichler (approved):
The old version was actually quite funny - and I have to blame myself because I copied the variable name from line 229 (that pre-existed when I started looking into this file in 2010 or 2011) to line 152.

PR #10999, opened 2020-10-03
Avoid quadratic complexity in ParticleHandler::insert_global_particles

kronbichler:
`IndexSet::add_index` is generally of linear complexity for unsorted data, so calling it separately for every index we add makes this part of `ParticleHandler::insert_global_particles` of quadratic complexity.

Observed when I tried to run step-68 with a million particles.


peterrum (approved):

peterrum (approved):
Since this how we do it at different places of the library, I am merging.

PR #8569, opened 2019-08-14
Add cell->face_iterators()

peterrum:
references #8563

ping @bangerth @masterleinad

PR #11000, opened 2020-10-04
Refactor: cell->child_iterators()

### LINKED ISSUE #10206: Introduce cell->child_iterators()
In the same way as we already have `cell->face_iterators()`.

marcfehling:
Closes #10206.

This PR provides `cell->cell_iterators()` functions similar to #8569 with `cell->face_iterators()`.

~~As this change also affects finding the proper finite element for parent cells, I also added a corresponding `cell->dominated_future_fe_on_children()` in this context. This improves the readability of the entire h-coarsening algorithm in the hp-context.~~
EDIT: This will be part of a later PR.

I did not run any tests yet as I currently do not have access to a suitable machine yet. Code compiles though.


include/deal.II/hp/fe_collection.h
```diff
@@ -739,11 +739,27 @@ namespace hp
...
+      {}
+      virtual void
+      print_info(std::ostream &out) const override
+      {
+        out << "    "
+            << "No FiniteElement has been found in your FECollection that is "
+               "dominated by all children of a cell you are trying to coarsen!"
+            << std::endl;
+      }
+    };
```
marcfehling, line 762: I found that this exception should belong to the `DoFCellAccessor` class rather than to the `FECollection` itself. I could not think of a more elegant way to deprecate this exception. Should we delete it right away?

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -2900,6 +2920,28 @@ DoFCellAccessor<dimension_, space_dimension_, level_dof_access>::
...
 
+template <int dimension_, int space_dimension_, bool level_dof_access>
+inline unsigned int
+DoFCellAccessor<dimension_, space_dimension_, level_dof_access>::
+  dominated_future_fe_on_children() const
+{
+  std::set<unsigned int> future_fe_indices_children;
+  for (const auto &child : this->child_iterators())
+    future_fe_indices_children.insert(child->future_fe_index());
+  Assert(!future_fe_indices_children.empty(), ExcInternalError());
```
peterrum, line 2931: > This function can only be called on direct parent cells, i.e., non-active cells whose children are all active.

Aren't here some asserts missing?

peterrum (commented):
May I ask you to split up this PR? The `child_iterators()` stuff is not related to `dominated_future_fe_on_children` and could be probably more quickly be merged.

include/deal.II/dofs/dof_accessor.templates.h
marcfehling: These assertions will be made when calling `cell->future_fe_index()`.

I agree that we should make these assertions here as well. Will comply!

Note: This will be part of a later PR.

marcfehling:
This PR's sole purpose is now to provide `child_iterators`. Did not run the testsuite on this one yet. But feel free to review!

bangerth (approved):

PR #11001, opened 2020-10-05
Disable compression in MPI::some_to_some

kronbichler:
As we found out in #8513 and #8754, compressing stuff before sending over MPI is typically more expensive than just sending packed but uncompressed data. Thus, I change to that version here.

Found out while looking at a profile of step-68 in parallel.

### DIFF
--- a/include/deal.II/base/mpi.h
+++ b/include/deal.II/base/mpi.h
@@ -1160,7 +1160,8 @@ namespace Utilities
           if (rank_obj.first != my_proc)
             {
               const auto &rank   = rank_obj.first;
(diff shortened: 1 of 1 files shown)


tjhei:
I am not surprised. Out of curiosity, did you profile this with data going over infiniband or tcpip, or is just considering the local case? I am in favor of making this change regardless.

kronbichler:
I did profile it locally (`memcpy` case), but given the timings I saw (equivalent of less than 1 Gbit/s when done with 40 cores) and bandwidth it should be 10x faster to send the uncompressed data over infiniband (40-100 Gbit/s nowadays) than to compress.

masterleinad (approved):

PR #11002, opened 2020-10-05
Make some more functions in ParticleHandler inline

kronbichler:
These are all very light-weight functions (well, maybe apart from the cost of the access to `std::multimap`), so it should be cheaper to let the compiler inline things.
Follow-up to #10987.


masterleinad (approved):
Looks good!

PR #11004, opened 2020-10-05
Use min/max rather than if statements

kronbichler:
`if` generally leads to conditional jumps in the code, whereas compilers can translate `std::min`/`std::max` into specific instructions on most architectures. The latter should be faster in essentially all circumstances, except the rare case being that the jump can be predicted well and the processor correctly speculates beyond it, whereas min/max take up execution resources otherwise free for floating point multiplications and additions. But that could also happen inside `min` or `max`, so better leave this to the compiler who knows more about the target hardware. Furthermore, this change would also allow to easily vectorize these two functions.


masterleinad (approved):
Looks good to me!

PR #11008, opened 2020-10-05
Fix type of FineDoFHandlerView

peterrum:
`unsigned int` -> `types::global_dof_index`


kronbichler (approved):
This looks good to me. Was there are particular test case that exposed this error? I do not see testsuite failures (but I guess it is difficult to test case it requires >4b unknowns).

peterrum:
Yes, I had a larger simulation which failed because of this.

bangerth:
I think you actually used the wrong type. Shouldn't it have been `types::global_cell_index`, given that the objects you refer to are all cell counts (if I see this right)?

PR #11009, opened 2020-10-05
TimeStepping: add SSP RK 3 and convergence tests

zjiaqi2018:
1. SSP_THIRD_ORDER is added to the namespace TimeStepping to 
implement the explicit third order Strong Stability Preserving (SSP) Runge-Kutta method.
2. Test convergence orders of explicit and implicit Runge-Kutta methods in TimeStepping.


Rombur (approved):
This is very nice. Thanks for improving the current test.

PR #11011, opened 2020-10-06
Refactor: Clean up hp-coarsening code. 

marcfehling:
~~Blocked by #11000.~~

After introducing `cell->child_iterators()`, I also added a corresponding `cell->dominated_future_fe_on_children()` function in this context. This improves the readability of the entire h-coarsening algorithm in the hp-context.

Did not run the testsuite on this one yet.

### DIFF
--- a/include/deal.II/dofs/dof_accessor.h
+++ b/include/deal.II/dofs/dof_accessor.h
@@ -2039,6 +2039,48 @@ public:
    */
   void
   clear_future_fe_index() const;
+
+  /**
+   * Return the index of the finite element from the entire hp::FECollection
+   * that is dominated by those assigned as future finite elements to the
+   * children of this cell.
+   *
+   * We find the corresponding finite element among the future finite elements
+   * on the children of this cell. If none of them qualify, we extend our search
+   * on the whole hp::FECollection, which is the element that describes the
+   * smallest finite element space that includes all future finite elements
+   * assigned to the children. If the function is not able to find a finite
+   * element at all, an assertion will be triggered.
+   *
+   * In this way, we determine the finite element of the parent cell in case of
+   * h-coarsening in the hp-context.
+   *
+   * @note This function can only be called on direct parent cells, i.e.,
+   * non-active cells whose children are all active.
+   */
+  unsigned int
+  dominated_future_fe_on_children() const;
+  /**
+   * @}
+   */
+
+  /**
+   * @name Exceptions
+   */
+  /**
+   * @{
+   */
+
+  /**
+   * Exception
+   *
+   * @ingroup Exceptions
+   */
+  DeclExceptionMsg(
+    ExcNoDominatedFiniteElementOnChildren,
+    "No FiniteElement has been found in your FECollection that is "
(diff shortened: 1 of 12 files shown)


include/deal.II/hp/fe_collection.h
```diff
@@ -739,11 +739,27 @@ namespace hp
...
+      {}
+      virtual void
+      print_info(std::ostream &out) const override
+      {
+        out << "    "
+            << "No FiniteElement has been found in your FECollection that is "
+               "dominated by all children of a cell you are trying to coarsen!"
+            << std::endl;
+      }
+    };
```
marcfehling, line 762: I found that this exception should belong to the `DoFCellAccessor` class rather than to the `FECollection` itself. I could not think of a more elegant way to deprecate this exception. Should we delete it right away?

bangerth: Just nuke it. There is no need to deprecate exceptions unless you think that there is used code that catches this exception -- which I think is highly unlikely given that there is nothing user code could do if this exception is thrown.

marcfehling: I can't think of any reason in which a user could have used that particular exception. I'll get rid of it.

bangerth:
Please rebase.

marcfehling:
I just rebased the branch. Ready for review/tests.

include/deal.II/dofs/dof_accessor.h
```diff
@@ -2039,6 +2039,46 @@ class DoFCellAccessor : public DoFAccessor<dimension_,
    */
   void
   clear_future_fe_index() const;
+
+  /**
+   * Return the index of the finite element that is dominated by those assigned
+   * as future finite elements to the children of this cell.
```
bangerth, line 2045: "the index" here is the one among the ones you are finding on the child cells, right, not among all possible indices? That's because your second paragraph seems to suggest that you're only considering all other indices in the second step. Maybe clarify that here.

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -2920,6 +2920,40 @@ DoFCellAccessor<dimension_, space_dimension_, level_dof_access>::
 
 
 
+template <int dimension_, int space_dimension_, bool level_dof_access>
+inline unsigned int
+DoFCellAccessor<dimension_, space_dimension_, level_dof_access>::
+  dominated_future_fe_on_children() const
+{
+  Assert(!(this->is_active()),
```
bangerth, line 2928:
```suggestion
  Assert(!this->is_active(),
```

bangerth (commented):
Almost.

include/deal.II/dofs/dof_accessor.h
marcfehling: I'll add another sentence for the second paragraph.

marcfehling:
Applied your suggestions. If you're happy with the documentation, feel free to merge!

bangerth (approved):

PR #11018, opened 2020-10-08
Fix type of FineDoFHandlerView (again)

peterrum:
Referencing https://github.com/dealii/dealii/pull/11008#issuecomment-704620880.

FYI @bangerth (I totally forgot that we have introduced that alias).


drwells (approved):

drwells:
This is definitely the right type.

bangerth:
Thanks!

PR #11023, opened 2020-10-08
VectorNorm for fixed_fraction refinement.

### LINKED ISSUE #10115: New feature: VectorNorm for fixed_fraction refinement.
For fixed fraction refinement, we concatenate refinement criteria of multiple cells to decide which ones will be refined.

Right now, we invoke the `l1`-norm to estimate the total error and the error on subsets of cells. This is sufficient for refinement heuristics, however users may invoke a different norm in this context, especially the `l2`-norm to reconstruct results from the original paper of Kelly et al. (1983) and Melenk and Wohlmuth (2001).

I would suggest to add a new parameter `VectorNorm` to decide which norm will be used. What do you think? I'd provide a patch in the near future.

marcfehling:
Closes #10115.

I basically reused the original `fixed_fraction` algorithms and the realized the `l2_norm` by adjusting the parameters for the original ones. I moved the original algorithms to an anonymous namespace, thus the number of lines changed is way less than git suggests.

While there, I refactored the original algorithm for the non-parallel application. I feel like it's easier to read now.

I find it strange that the number of refined and coarsened cells differ between serial and parallel implementations. I thought I fixed this in #9140 by adjusting the `fixed_number` algorithm which is inherently called. I'll give it a look (but in a separate PR).

### DIFF
--- a/include/deal.II/distributed/grid_refinement.h
+++ b/include/deal.II/distributed/grid_refinement.h
@@ -23,6 +23,8 @@
 
 #include <deal.II/distributed/tria.h>
 
+#include <deal.II/numerics/vector_tools_common.h>
+
 #include <limits>
 #include <vector>
 
@@ -124,6 +126,28 @@ namespace parallel
        * refined at all.
        *
        * The same is true for the fraction of cells that is coarsened.
+       *
+       * @param[in,out] tria The triangulation whose cells this function is
+       * supposed to mark for coarsening and refinement.
+       *
+       * @param[in] criteria The refinement criterion for each mesh cell active
+       * on the current triangulation. Entries may not be negative.
+       *
+       * @param[in] top_fraction_of_cells The fraction of cells to be refined.
+       * If this number is zero, no cells will be refined. If it equals one,
+       * the result will be flagging for global refinement.
+       *
+       * @param[in] bottom_fraction_of_cells The fraction of cells to be
+       * coarsened. If this number is zero, no cells will be coarsened.
+       *
+       * @param[in] max_n_cells This argument can be used to specify a maximal
+       * number of cells. If this number is going to be exceeded upon
+       * refinement, then refinement and coarsening fractions are going to be
+       * adjusted in an attempt to reach the maximum number of cells. Be aware
+       * though that through proliferation of refinement due to
+       * Triangulation::MeshSmoothing, this number is only an indicator. The
+       * default value of this argument is to impose no limit on the number of
+       * cells.
        */
       template <int dim, typename Number, int spacedim>
       void
@@ -157,14 +181,35 @@ namespace parallel
        * processors, no cells are refined at all.
        *
        * The same is true for the fraction of cells that is coarsened.
+       *
+       * @param[in,out] tria The triangulation whose cells this function is
+       * supposed to mark for coarsening and refinement.
+       *
+       * @param[in] criteria The refinement criterion computed on each mesh cell
+       * active on the current triangulation. Entries may not be negative.
+       *
+       * @param[in] top_fraction_of_error The fraction of the total estimate
+       * which should be refined. If this number is zero, no cells will be
+       * refined. If it equals one, the result will be flagging for global
+       * refinement.
+       *
+       * @param[in] bottom_fraction_of_error The fraction of the estimate
+       * coarsened. If this number is zero, no cells will be coarsened.
+       *
+       * @param[in] norm_type To determine thresholds, combined errors on
+       * subsets of cells are calculated as norms of the criteria on these
+       * cells. Different types of norms can be used for this purpose, from
+       * which VectorTools::NormType::L1_norm and
+       * VectorTools::NormType::L2_norm are currently supported.
        */
       template <int dim, typename Number, int spacedim>
       void
       refine_and_coarsen_fixed_fraction(
         parallel::distributed::Triangulation<dim, spacedim> &tria,
         const dealii::Vector<Number> &                       criteria,
-        const double top_fraction_of_error,
-        const double bottom_fraction_of_error);
+        const double                top_fraction_of_error,
+        const double                bottom_fraction_of_error,
+        const VectorTools::NormType norm_type = VectorTools::NormType::L1_norm);
     } // namespace GridRefinement
   }   // namespace distributed
 } // namespace parallel
--- a/source/distributed/grid_refinement.cc
+++ b/source/distributed/grid_refinement.cc
@@ -186,6 +186,67 @@ namespace
           cell->clear_coarsen_flag();
         }
   }
+
+
+
+  /**
+   * Fixed fraction algorithm without a specified vector norm.
+   *
+   * Entries of the criteria vector and fractions are taken as is, so this
+   * function basically evaluates norms on the vector or its subsets as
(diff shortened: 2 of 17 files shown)


source/distributed/grid_refinement.cc
```diff
@@ -519,45 +582,43 @@ namespace parallel
...
+                // we do not want to evaluate norms on subsets as
+                //   sqrt(c_0^2 + c_1^2 + ...) < fraction * l2-norm(c)
+                // square fractions and criteria instead and work with
+                //   c_0^2 + c_1^2 + ... < fraction^2 * l1-norm(c*c)
+                // we adjust all parameters accordingly
+                Vector<Number> criteria_squared(criteria.size());
+                std::transform(criteria.begin(),
+                               criteria.end(),
+                               criteria_squared.begin(),
+                               [](Number c) { return c * c; });
```
marcfehling, line 608: This might be sped up with a `parallel_for` if you like.

source/grid/grid_refinement.cc
```diff
@@ -280,108 +404,41 @@ GridRefinement::refine_and_coarsen_fixed_fraction(
...
+          // we do not want to evaluate norms on subsets as
+          //   sqrt(c_0^2 + c_1^2 + ...) < fraction * l2-norm(c)
+          // square fractions and criteria instead and work with
+          //   c_0^2 + c_1^2 + ... < fraction^2 * l1-norm(c*c)
+          // we adjust all parameters accordingly
+          Vector<Number> criteria_squared(criteria.size());
+          std::transform(criteria.begin(),
+                         criteria.end(),
+                         criteria_squared.begin(),
+                         [](Number c) { return c * c; });
```
marcfehling, line 427: See above.

marcfehling:
I suppose the differences in the refinement behavior are due to the distinct algorithms to determine the corresponding refinement thresholds. While we work on a sorted criteria vector in the serial case, the thresholds are determined with a bisection algorithm in parallel. At some point both algorithms match, sometimes they don't. I do not see a pattern. I'll provide another test.

include/deal.II/distributed/grid_refinement.h
```diff
@@ -163,8 +165,9 @@ namespace parallel
       refine_and_coarsen_fixed_fraction(
         parallel::distributed::Triangulation<dim, spacedim> &tria,
         const dealii::Vector<Number> &                       criteria,
-        const double top_fraction_of_error,
-        const double bottom_fraction_of_error);
+        const double                top_fraction_of_error,
+        const double                bottom_fraction_of_error,
+        const VectorTools::NormType norm_type = VectorTools::NormType::L1_norm);
```
bangerth, line 170: Can you document the new argument?

source/distributed/grid_refinement.cc
```diff
@@ -186,6 +186,67 @@ namespace
...
+  /**
+   * Fixed fraction algorithm without a specified vector norm.
+   *
+   * Entries of the criteria vector and fractions are taken as is, so this
+   * function basically evaluates norms on the vector or its subsets as
+   * l1-norms.
+   */
+  template <int dim, typename Number, int spacedim>
+  void
+  refine_and_coarsen_fixed_fraction_l1_norm(
```
bangerth, line 201: This is an internal function, right? If so, put it into an anonymous namespace.

bangerth: Also, if it's an internal function, just use the "traditional" order of template arguments: dim, spacedim, Number

source/distributed/grid_refinement.cc
```diff
@@ -519,45 +582,43 @@ namespace parallel
...
+                criteria,
+                top_fraction_of_error,
+                bottom_fraction_of_error);
+              break;
+
+            case VectorTools::NormType::L2_norm:
+              {
+                // we do not want to evaluate norms on subsets as
+                //   sqrt(c_0^2 + c_1^2 + ...) < fraction * l2-norm(c)
+                // square fractions and criteria instead and work with
```
bangerth, line 601: I don't think I understand what you want to say with "square fractions and criteria"

source/grid/grid_refinement.cc
```diff
@@ -280,108 +404,41 @@ GridRefinement::refine_and_coarsen_fixed_fraction(
...
+        //   c_0 + c_1 + ... < fraction * l1-norm(c)
+        refine_and_coarsen_fixed_fraction_l1_norm(
+          tria, criteria, top_fraction, bottom_fraction, max_n_cells);
+        break;
+
+      case VectorTools::NormType::L2_norm:
+        {
+          // we do not want to evaluate norms on subsets as
+          //   sqrt(c_0^2 + c_1^2 + ...) < fraction * l2-norm(c)
+          // square fractions and criteria instead and work with
```
bangerth, line 420: same here

source/distributed/grid_refinement.cc
bangerth: Not worth it. We don't do it often, and it's so cheap.

source/distributed/grid_refinement.cc
marcfehling: Actually, it is already in an anonymous namespace.

I took the order of template arguments from the corresponding functions in `GridRefinement` and `parallel::distributed::GridRefinement`. I will follow your suggestion.

include/deal.II/distributed/grid_refinement.h
marcfehling: Documentation is missing for all arguments of this function and the corresponding `refine_and_coarsen_fixed_number` function. Their documentation refers to the non-parallel implementation in `GridRefinement`.

I will copy the `@param` documentation from the `GridRefinement` namespace for both functions.

marcfehling:
I will squash all commits once you approve this PR.

include/deal.II/distributed/grid_refinement.h
bangerth: Thanks!

include/deal.II/distributed/grid_refinement.h
```diff
@@ -157,14 +181,34 @@ namespace parallel
...
+       * refined. If it equals one, the result will be flagging for global
+       * refinement.
+       *
+       * @param[in] bottom_fraction_of_error The fraction of the estimate
+       * coarsened. If this number is zero, no cells will be coarsened.
+       *
+       * @param[in] norm_type Sets the type of norm to which the criteria of
+       * several cells will be concatenated. Currently
+       * VectorTools::NormType::L1_norm and VectorTools::NormType::L2_norm are
+       * supported.
```
bangerth, line 202: Here and in the other place, I don't think what "concatenated" is supposed to refer to. I think it would be useful if you stated what the refinement criterion actually is, and where these exponents come into play.

bangerth (commented):
Other than the one place (and its copy), this looks good.

include/deal.II/distributed/grid_refinement.h
marcfehling: I felt this was appropriate:
To determine thresholds, combined errors on subsets of cells are calculated as norms of the criteria on these cells.

marcfehling:
I addressed your changes and went on to squash the commits.

bangerth (approved):

Issue #11007, opened 2020-10-05
distribute_dofs with subdivided_hyper_rectangle fails in parallel


jodlbauer:
Dear all,

for some reason, the following code segfaults during `distribute_dofs` when run on 2+ cores.
Could someone please confirm that this bug is not only on my side?
It works if any of np = 1, `GridGenerator::hyper_rectangle` or `shared::Triangulation` is used instead.

My deal.II version is slightly behind master, will try to merge tomorrow.

```
#include <deal.II/distributed/tria.h>
#include <deal.II/dofs/dof_handler.h>
#include <deal.II/fe/fe_q.h>
#include <deal.II/grid/grid_generator.h>
#include <iostream>
#include <vector>

int main(int argc, char *argv[])
{
   //   dealii::Utilities::MPI::MPI_InitFinalize mpi(argc, argv, 1);
   MPI_Init(&argc, &argv);

   dealii::parallel::distributed::Triangulation<2> tria(MPI_COMM_WORLD);

   dealii::Point<2>          a{0, 0};
   dealii::Point<2>          b{1, 1};
   std::vector<unsigned int> reps{2, 2};

   dealii::GridGenerator::subdivided_hyper_rectangle(tria, reps, a, b, false);
   // tria.refine_global(1);

   dealii::DoFHandler<2> dofs(tria);
   dealii::FE_Q<2>       fe(1);
   dofs.distribute_dofs(fe);

   std::cout << dofs.n_dofs() << std::endl;

   MPI_Finalize();
}
```

Best regards,

Daniel

jppelteret:
Hi @jodlbauer,

The mailing list is a better place for this question (we try to keep all community questions in one place, and reserve Github issues for development). But I'll try to work with you a little on this:

I'm curious as to why you have commented out the call to `MPI_InitFinalize`? Are you sure that you're doing everything that [this class' constructor](https://dealii.org/developer/doxygen/deal.II/classUtilities_1_1MPI_1_1MPI__InitFinalize.html#a71954ff48c33b1101b8a4d64d9c685db) does? What happens if you uncomment it, and rather comment out your calls to `MPI_Init()` and `MPI_Finalize()`?

jodlbauer:
Well, I was indecisive whether to put it here or on the mailing list, but chose here, as it might by an issue that needs to be resolved anyways.

[...]

PR #11024, opened 2020-10-09
Add a quick check for matching boost versions.

drwells:
This addresses a problem a user had recently with inscrutable crashes caused by
mixed boost versions. Credit goes to @jodlbauer for reporting the issue.

I don't think boost is particularly strict on ABI stability even in patch versions so I went with the full version check.


drwells:
Related to #11007.

masterleinad (approved):
Looks good to me.

bangerth (approved):

PR #1177, opened 2015-07-22
[WIP] Q1 dimension-dependent real-to-unit transforms

drwells:
This is a work-in-progress because it causes a dozen tests (which test which cell an edge point or corner is in) to fail and I am not sure if breaking that behavior is significant. The failing tests are:

```
1358 - bits/find_cell_4.debug (Failed)
1405 - bits/find_cell_alt_4.release (Failed)
1698 - bits/find_cell_4.release (Failed)
1899 - bits/find_cell_alt_4.debug (Failed)
2003 - bits/find_cell_5.release (Failed)
2053 - bits/find_cell_1.debug (Failed)
2054 - bits/find_cell_1.release (Failed)
2208 - bits/find_cell_alt_1.debug (Failed)
2209 - bits/find_cell_alt_1.release (Failed)
2458 - bits/find_cell_alt_5.debug (Failed)
2663 - bits/find_cell_alt_5.release (Failed)
2759 - bits/find_cell_5.debug (Failed)
```

The 2D transform with the exact formula is about 45x faster and the 3D transform is about 8x faster. The accuracy is about the same (I think the new 3D solver is a bit more accurate). My transforms perform badly when the point is not in the cell (I switch back to the original algorithm in those cases) so this does not quite solve #314, #866, or #872.

PR #11027, opened 2020-10-10
Extend GridTools::min/max_cell_diameter for simplex

peterrum:


drwells (changes requested):
I'm fine with breaking this interface (we need to for simplices to work). We also need to add a change log entry under `incompatibilities`.

peterrum:
@drwells I have added the entry!

drwells (approved):

source/fe/mapping.cc
```diff
@@ -24,15 +24,17 @@ DEAL_II_NAMESPACE_OPEN
...
-  for (const unsigned int i : GeometryInfo<dim>::vertex_indices())
-    {
-      vertices[i] = cell->vertex(i);
-    }
+  boost::container::small_vector<Point<spacedim>,
+                                 GeometryInfo<dim>::vertices_per_cell>
+    vertices(cell->n_vertices());
+  for (const unsigned int i : cell->vertex_indices())
+    vertices[i] = cell->vertex(i);
+
```
kronbichler, line 37: I guess it won't make a big difference performance-wise, but I would prefer this code
```suggestion
  boost::container::small_vector<Point<spacedim>,
                                 GeometryInfo<dim>::vertices_per_cell>
    vertices;
  for (const unsigned int i : cell->vertex_indices())
    vertices.push_back(cell->vertex(i));
```

source/fe/mapping_q_eulerian.cc
```diff
@@ -130,8 +131,9 @@ MappingQEulerian<dim, VectorType, spacedim>::get_vertices(
...
       .compute_mapping_support_points(cell);
 
-  std::array<Point<spacedim>, GeometryInfo<dim>::vertices_per_cell>
-    vertex_locations;
+  boost::container::small_vector<Point<spacedim>,
+                                 GeometryInfo<dim>::vertices_per_cell>
+    vertex_locations(GeometryInfo<dim>::vertices_per_cell);
   std::copy(a.begin(),
             a.begin() + GeometryInfo<dim>::vertices_per_cell,
             vertex_locations.begin());
```
kronbichler, line 139: I think this should work:
```suggestion
  boost::container::small_vector<Point<spacedim>,
                                 GeometryInfo<dim>::vertices_per_cell>
    vertex_locations(a.begin(),
                     a.begin() + GeometryInfo<dim>::vertices_per_cell);
```

source/fe/mapping_q_generic.cc
```diff
@@ -2518,8 +2518,13 @@ MappingQGeneric<dim, spacedim>::transform_real_to_unit_cell(
...
       // * In 3D there is no (known to the authors) exact formula, so the Newton
       //   algorithm is used.
-      const std::array<Point<spacedim>, GeometryInfo<dim>::vertices_per_cell>
-        vertices = this->get_vertices(cell);
+      const auto vertices_ = this->get_vertices(cell);
+
+      std::array<Point<spacedim>, GeometryInfo<dim>::vertices_per_cell>
+        vertices;
+      for (unsigned int i = 0; i < vertices.size(); ++i)
+        vertices[i] = vertices_[i];
```
kronbichler, line 2526: Given #10983, I wonder whether we still need that code from a performance point of view; as much as I like the analytical formula to invert a bilinear mapping, I think the new code mostly eliminates the bottleneck, except maybe the memory allocation that is unnecessary. There is nothing to change here for now, it might just be an opportunity to delete 200 lines of code at some point.

kronbichler (commented):
Looks good to me, apart from a few things where I would directly create the elements in the vector with the right entries rather than using `operator[]`.

peterrum:
@kronbichler I just noticed that I have forgotten to address you comments. It is done now.

drwells (approved):

kronbichler (approved):
Looks good. This is in the public interface of the `Mapping` base class, so potentially a number of users could be affected. On the other hand, we seem to have this function "only" for five years due to #1177, so I think that it will not be that many users after all. And it's not that many who really manually work with the data offered by the mapping.

PR #11033, opened 2020-10-11
Explain SUPG in step-9.

bangerth:
One of my students (hi @dklong-csu!) was reading through step-9 and was confused about SUPG. I thought I should just write up what that method is and why it has its name in step-9, so that it's available for everyone.

### DIFF
--- a/examples/step-9/doc/intro.dox
+++ b/examples/step-9/doc/intro.dox
@@ -11,7 +11,7 @@ In this example, our aims are the following:
 </ol>
 While the second aim is difficult to describe in general terms without
 reference to the code, we will discuss the other two aims in the
-following text. The use of multiple threads will then be detailed at the
+following. The use of multiple threads will then be detailed at the
 relevant places within the program. We will, however, follow the
 general discussion of the WorkStream approach detailed in the
 @ref threads "Parallel computing with multiple processors accessing shared memory"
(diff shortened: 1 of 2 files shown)


examples/step-9/doc/intro.dox
```diff
@@ -176,6 +179,117 @@ and then to obtain
...
+  -
+  \delta
+  \frac{\partial^2}{\partial\beta^2} u_h.
+@f]
+In other words, the obscure choice of test function is equivalent to
+the addition of term to the strong form that corresponds to a second
+order (=diffusion) differential operator in the direction of the wind
+field $\beta$, i.e., in "streamline direction". A fuller account would
+also have to explore the effect of the test function on boundary
+values and why it is necessary to also use the funny test function for
```
drwells, line 245:
```suggestion
values and why it is necessary to also use the same test function for
```
It is a funny test function but readers might get confused by the extra name.

examples/step-9/doc/intro.dox
```diff
@@ -66,24 +67,26 @@ for the derivatives in the perpendicular direction, and consequently
...
-
 These discontinuities lead to numerical instabilities that
 make a stable solution by a standard continuous finite element discretization
-impossible. We will thus use the streamline diffusion stabilized
-formulation, in which we test the equation with test functions $v +
-\delta \beta\cdot\nabla v$ instead of $v$, where $\delta$ is a
+impossible.
+
+A standard approach to address this difficulty is the "<i>streamline
+upwind Petrov Galerkin</i>" (SUPG) method, sometimes also called the
```
drwells, line 75:
```suggestion
A standard approach to address this difficulty is the <em>"streamline-upwind
Petrov-Galerkin"</em> (SUPG) method, sometimes also called the
```
Doxygen for whatever reason renders the HTML tags literally when they are in quotes.

drwells (changes requested):
Looks good - I just have some minor comments.

bangerth:
Thanks, @drwells, for the careful reading. All changes incorporated as suggested!

drwells (approved):

PR #11022, opened 2020-10-08
 Make indent CI check abort on Doxygen warnings

masterleinad:
Apparently, this wasn't working correctly.

PR #11034, opened 2020-10-11
Generic evaluation of tensor product interpolation and gradient in a point

kronbichler:
This PR generalizes the function introduced in #10983 towards a generic evaluation of quantities. It allows the user to pass a `Point<dim,Number2>` and a vector of coefficients to be used for the polynomial description. The latter can be scalars or `Point<spacedim>` or some other type. With the help of `ProductType`, we can even make this function work with `VectorizedArray` types, i.e., provide vectorization evaluation of the interpolation via `Point<dim,VectorizedArrayType>`, without code duplication. To enable this nice functionality, I templated one of our polynomial evaluation classes such that it also takes `VectorizedArrayType` arguments.

In a follow-up PR, I plan to vectorize the `MappingQGeneric::transform_points_real_to_unit_cell` function, but I keep it out from this PR to keep reviews manageable. The four tests go into different types of arguments to the new function, i.e., scalar, vector-valued, vector-valued + vectorized and vector-valued + vectorized + different numbering, showing how broadly this function can be used.

Helper function for #10968.


include/deal.II/base/polynomial.h
```diff
@@ -807,6 +817,141 @@ namespace Polynomials
 
 
 
+  template <typename number>
+  template <typename Number2>
+  inline void
+  Polynomial<number>::value(const Number2      x,
```
peterrum, line 823: As far as I can see this method only has been moved into the header file duet to the new template argument.

include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -2310,6 +2311,194 @@ namespace internal
...
+   * @param renumber Optional parameter to specify a renumbering in the
+   * coefficient vector, assuming that `values[renumber[i]]` returns
+   * the lexicographic (tensor product) entry of the coefficients. If the
+   * vector is entry, the values are assumed to be sorted lexicographically.
+   */
+  template <int dim, typename Number, typename Number2>
+  inline std::pair<
+    typename ProductTypeNoPoint<Number, Number2>::type,
+    Tensor<1, dim, typename ProductTypeNoPoint<Number, Number2>::type>>
+  evaluate_tensor_product_value_and_gradient(
```
peterrum, line 2370: This function has been moved from source/fe/mapping_q_generic.cc.

peterrum (approved):
Looks good once the indentation has been fixed. Most of the code has been just moved around to enable re-usage of the functionalities.

include/deal.II/base/polynomial.h
kronbichler: Yes, it was moved from there, I adjusted the types of temporary variables to `Number2` and removed this shortcut https://github.com/dealii/dealii/blob/d949861f2fb94acc0ce0f318983e214f87913c34/source/base/polynomial.cc#L216-L223 The reason for removing it is that (a) this is the unlikely path (almost all polynomials we use are in the Lagrange product form), so I do not think we need to optimize for it, and (b) I do not want to template the `value(const number x)` function (otherwise we lose the implicit conversion from `int` to `double`, for example) and do not see a reason to implement yet another specialization.

include/deal.II/matrix_free/tensor_product_kernels.h
kronbichler: Yes, again with adjusted types and some template magic.

kronbichler:
> Looks good once the indentation has been fixed. 

Here we see #11022 at work (I missed to rename a parameter in the doxygen).

peterrum (approved):

kronbichler:
Any additional comments on this PR? I have a two pending PRs depending on this which I would like to post soon. It would save some time if the interface is fixed.

peterrum (approved):
To me the PR looks good and it is safe to merge. I will wait till tonight and merge then.

*A side remark: at a point we might think about moving the content of `matrix_free/tensor_product_kernels.h` to another place (base?) if we continue to use more functionalities outside of the `matrix_free` folder.*

kronbichler:
> A side remark: at a point we might think about moving the content of matrix_free/tensor_product_kernels.h to another place (base?)

I agree and consider that part of #7039.

masterleinad (approved):

PR #11038, opened 2020-10-12
Enable FE_Nothing in MatrixFree

Rombur:
With this PR I can pass a simple test that uses` FE_Nothing` and `MatrixFree`. Before merging I want to test this more but I want to make sure that what I am doing make sense (@peterrum @kronbichler). The PR does two things: 
 1) when using `FE_Nothing` or a `Quadrature` with zero points, `ShapeInfo<Number>::reinit` does nothing
 2) I think that was a bug: `task_info.create_blocks_serial` requires the number of dofs per cell. We loop over the `dof_info` but in the `dof_info` we don't loop over `dofs_per_cell`, we would just take the first one. When `FE_Nothing` is first, we get `dofs_per_cell=0` and `task_info.create_blocks_serial` crashes.

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.templates.h
+++ b/include/deal.II/matrix_free/matrix_free.templates.h
@@ -975,9 +975,12 @@ namespace internal
       {
         const bool strict_categories =
           cell_vectorization_categories_strict || hp_functionality_enabled;
-        unsigned int dofs_per_cell = 0;
+        // Use max dofs per cell in create_blocks_serial. This is necessary
+        // to use FE_Nothing.
+        unsigned int max_dofs_per_cell = 0;
         for (const auto &info : dof_info)
-          dofs_per_cell = std::max(dofs_per_cell, info.dofs_per_cell[0]);
+          for (const auto &dofs : info.dofs_per_cell)
+            max_dofs_per_cell = std::max(max_dofs_per_cell, dofs);
 
         // Detect cells with the same parent to make sure they get scheduled
         // together in the loop, which increases data locality.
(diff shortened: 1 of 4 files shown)


include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -977,7 +977,8 @@ namespace internal
           cell_vectorization_categories_strict || hp_functionality_enabled;
         unsigned int dofs_per_cell = 0;
         for (const auto &info : dof_info)
-          dofs_per_cell = std::max(dofs_per_cell, info.dofs_per_cell[0]);
+          for (const auto &dofs : info.dofs_per_cell)
+            dofs_per_cell = std::max(dofs_per_cell, dofs);
```
peterrum, line 981: This looks reasonable. We should rename  the variable `dofs_per_cell` to `max_dofs_per_cell` and add an assert in `TaskInfo::create_blocks_serial` ruling out 0. Maybe we could add a comment here why we are taking here the maximum: it seems to be a useful heuristic for a good blocking!?

include/deal.II/matrix_free/shape_info.templates.h
```diff
@@ -95,6 +95,9 @@ namespace internal
       Assert(fe->n_components() == 1,
              ExcMessage("FEEvaluation only works for scalar finite elements."));
 
+      if ((fe->n_dofs_per_cell() == 0) || (quad.size() == 0))
```
peterrum, line 98: Look reasonable. Although I would move this down after the `univariate_shape_data` has been initialized, i.e., after `univariate_shape_data.n_q_points_1d = quad.size();`. I think there are multiple places in the library where we do something like `shape_infe.data[0].quadrature`, aren't there? @kronbichler What is your opinion?

kronbichler: I agree, we can still set up some data structures with empty content.

Rombur:
I have updated the PR

peterrum (approved):

kronbichler (approved):

PR #11040, opened 2020-10-13
Fix MF setup for hp::FECollection::size() neq hp::QCollection::size()

peterrum:
The following setup failed:
```cpp
  hp::FECollection<dim> fe(FE_Q<dim> {fe_degree}, FE_Q<dim> {fe_degree+1});
  hp::QCollection<1> quad{QGauss<1> {fe_degree + 1}};
```
but
```cpp
  hp::FECollection<dim> fe(FE_Q<dim> {fe_degree}, FE_Q<dim> {fe_degree+1});
  hp::QCollection<1> quad{QGauss<1> {fe_degree + 1}, QGauss<1> {fe_degree + 2}};
```
works. 

This PR does the same decision as done in many places in the library. If not enough quadrature rules are provided, simply use the first one (and not the one that is index with `active_fe_index`).

### DIFF
--- a/include/deal.II/matrix_free/mapping_info.templates.h
+++ b/include/deal.II/matrix_free/mapping_info.templates.h
@@ -799,8 +799,15 @@ namespace internal
         // FEValues
         std::vector<std::vector<std::shared_ptr<dealii::FEValues<dim>>>>
           fe_values(mapping_info.cell_data.size());
+
+        const unsigned int n_active_fe_indices =
+          active_fe_index.size() > 0 ?
+            (*std::max_element(active_fe_index.begin(), active_fe_index.end()) +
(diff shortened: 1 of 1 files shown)


kronbichler (approved):
This looks good. Did you check that the setup of the `ShapeInfo` field in https://github.com/dealii/dealii/blob/dee4b550d3c3bc9014403d9b84740ee28f615e34/include/deal.II/matrix_free/matrix_free.templates.h#L303-L307 works as well?

peterrum:
@kronbichler To me these lines look correct. It explicitly distinguishes `n_fe_in_collection` and `n_quad_in_collection`. See also the four lines above the highlighted code.

PR #11045, opened 2020-10-15
Vectorize MappingQGeneric::transform_points_real_to_unit_cell

kronbichler:
Only the last commit is relevant, the rest is #11034. The vectorization doing several points at once is straight-forward and needs two ingredients:
- Template the internal `do_transform_real_to_unit_cell` with an abstract type `Number` that we can feed a `VectorizedArray<double>`
- Change checks of the type `residual_new <= residual_old` into `std::max(residual_new - residual_old, Number(0.)) == Number(0.)`, given that `==` is the only operator defined for `VectorizedArray` (and the other operators like `<` would be ambiguous, so I opted against introducing them).


kronbichler:
Related to #10969.

source/fe/mapping_q_generic.cc
```diff
@@ -865,20 +874,20 @@ namespace internal
...
-                Point<dim> p_unit_trial = p_unit;
+                Point<dim, Number> p_unit_trial = p_unit;
                 for (unsigned int i = 0; i < dim; ++i)
                   p_unit_trial[i] -= step_length * delta[i];
 
-                // shape values and derivatives
-                // at new p_unit point
-                const auto p_real_trial =
+                // shape values and derivatives at new p_unit point
+                auto p_real_trial =
```
peterrum, line 882: Why is this not const anymore?

source/fe/mapping_q_generic.cc
```diff
@@ -2191,28 +2210,75 @@ MappingQGeneric<dim, spacedim>::transform_points_real_to_unit_cell(
   const DerivativeForm<1, spacedim, dim> A_inv =
     affine_factors.first.covariant_form().transpose();
 
-  for (unsigned int i = 0; i < real_points.size(); ++i)
-    {
-      try
```
peterrum, line 2327: What happened with the try-catch?

source/fe/mapping_q_generic.cc
```diff
@@ -2191,28 +2210,75 @@ MappingQGeneric<dim, spacedim>::transform_points_real_to_unit_cell(
...
-                                 real_points[i] - affine_factors.second));
-          unit_points[i] = internal::MappingQGenericImplementation::
+  const unsigned int n_points = real_points.size();
+  const unsigned int n_lanes  = VectorizedArray<double>::size();
+
+  // only use the more heavy VectorizedArray code path if there are at least
+  // two points to compute
+  if (n_points > 1)
+    for (unsigned int i = 0; i < n_points; i += n_lanes)
+      {
```
peterrum, line 2220:
```cpp
  for (unsigned int i = 0; i < n_points; i += n_lanes)
    if (n_points - i > 1) ... // new vectorized code
    else ...                  // old code
```
What about?

peterrum (approved):
Looks good to me.

source/fe/mapping_q_generic.cc
kronbichler: Effect from switching between branches. I will fix.

source/fe/mapping_q_generic.cc
kronbichler: We should not need the `try-catch` any more because the implementation in the class never throws (at least not the `MappingFailed` exception).

masterleinad (approved):
Looks good to me.

PR #11046, opened 2020-10-15
FEPointEvaluation: Evaluating shape functions on arbitrary points of a cell

### LINKED ISSUE #10370: Allow changing the quadrature for a FEValuesBase object?
We have a number of places in the particle steps (and ASPECT) that do the following:
```
for (all cells)
{
    std::vector<Point<dim>> particle_positions = ... some locations;
    const Quadrature<dim> quadrature_formula(particle_positions);
    FEValues<dim>         fe_value(mapping,
                fe,
                quadrature_formula,
                update_values);

  .. do_something
}
```

This could be much simpler (and save some memory allocations) if we could to the following:
```
const Quadrature<dim> quadrature_formula;
FEValues<dim>         fe_value(mapping,
            fe,
            quadrature_formula,
            update_values);
for (all cells)
{
    std::vector<Point<dim>> particle_positions = ... some locations;
    const Quadrature<dim> quadrature_formula(particle_positions);
    fe_value.set_quadrature(quadrature_formula);
    fe_value.reinit(cell);
}
```

The `set_quadrature` would need to invalidate the current cell of course. Do you think it makes sense to introduce such a function? Should it instead be implemented as a `reinit(cell,quadrature)` function? Are there any obvious difficulties that I do not see at the moment?

### LINKED ISSUE #10968: Interface for fast evaluation of function value at arbitrary point

kronbichler:
This PR ~~currently work in progress,~~ implements a class that can evaluate an interpolation described by `FiniteElement` and given solution values on a cell in arbitrary points. Fixes #10968. Fixes #10370.

For moderate polynomial degrees and especially in 3D, the proposed class is similarly fast or slightly faster than the sequence of `FEValues::reinit()` followed by `FEValues::get_function_gradients` on *fixed* points. In other words, computing the information on the fly is not more expensive than the tabulated data of `FEValues`. For higher degrees, the present class is even considerably faster.

The main evaluator function is `FEPointEvaluation::evaluate()`, which is the place to look at. In the test cases one can see how the class is used. The interface is close to `FEValues::get_function_values/get_function_gradients` with the difference that we specify the cell and the reference points where to evaluate in the same call. The class uses some template magic to implement evaluators for a variable number of components, like scalar or vector-valued data.

~~At this point, I would like us to decide on the interface we expose to the user and the name of the class. I still have to fill in the case which is not `MappingQGeneric` and not using a `FiniteElement` compatible with `MatrixFreeFunctions::ShapeInfo`. That would simply use `FEValues` underneath and be slow, but still able to provide the desired answers.~~


kronbichler:
@blaisb @gassmoeller FYI

kronbichler:
A performance update: I tested the evaluator in step-68 with 13 million particles in 3D and polynomial degree 3 for the velocity field. The interpolation step, provided by this class, takes around 0.54 seconds per time step for `interpolated_euler_step`. This can be compared to around 1.2 seconds for incrementing the `particle_iterator` (aka `std::multimap`) during the same function call. Well, I should add that I increment it twice, once for filling the reference points and once for updating the positions. While this also highlights a deficiency in data structures for the particle storage (a map is far from the fastest option to store this kind of things), it nonetheless shows the potential for the present evaluator.

gassmoeller:
Hi Martin, thanks for working on this, this will be useful for me! I never got around to further work on #10370, but your new class would indeed fix that need. I feel like my inisght into the interface of FEValues and FEEvaluation is too superficial to answer the question about which interface to go for, I guess I can use either one. Here are just some high level ideas I had while looking through the PR:
1. Concerning the name of the class: Would it make sense to stick to the naming scheme of FEValues and FEEvaluation, e.g. by calling the new class `FEValuesFlexible`, `FEValuesFlexiblePoints` or `FEPointEvaluation`? Conceptually the class does the same as FEValues, just on flexible quadrature points?
2. Your current interface seems to support only values and gradients, would hessians or other higher order derivates create a significant problem, or did you just not implement them because they are needed less often? I do not need them at the moment, but we should probably keep it in mind and design the interface to allow an extension in the future, I think your interface would support that by adding another optional input argument? No need to add them before going forward with this though from my point of view.

> This can be compared to around 1.2 seconds for incrementing the particle_iterator (aka std::multimap) during the same function call. Well, I should add that I increment it twice, once for filling the reference points and once for updating the positions. While this also highlights a deficiency in data structures for the particle storage (a map is far from the fastest option to store this kind of things)

Yes, I decided using the map mostly for convenience, and because so far we never had a case where the performance of the container was a limiting factor that could not be worked around by clever use of iterators or container functions (like hints for insertion, inserting sorted elements etc.). If incrementing the iterators is starting to become a performance bottleneck though I am absolutely fine thinking about replacements.

kronbichler:
> Would it make sense to stick to the naming scheme of FEValues and FEEvaluation, e.g. by calling the new class FEValuesFlexible, FEValuesFlexiblePoints or FEPointEvaluation? Conceptually the class does the same as FEValues, just on flexible quadrature points?

It does the subset of functions in `FEValues` that is related to evaluating functions, so `FEPointEvaluation` would make sense I guess. We could also add functionality to test by the values or gradients of test functions similarly to what https://github.com/dealii/dealii/blob/aa93d0eb64f9ecc5e2e818c206a82eddae8b6d30/examples/step-19/step-19.cc#L469-L479
does. That would bring the functionality even closer to `FEEvaluation`, so the name analogy would make sense.

> If incrementing the iterators is starting to become a performance bottleneck though I am absolutely fine thinking about replacements.

I anticipate that in some of the applications I have in mind, the performance of particle iterations (or the access in general) could become an issue. I would be happy to have your insight and collaborate with you if/once we come to the point where the internal data structures of the particle handler might become an issue.

kronbichler:
That said, the particle infrastructure is nicely structured and documented so a *newcomer* like me can get productive within a week or two. :+1:

blaisb:
> A performance update: I tested the evaluator in step-68 with 13 million particles in 3D and polynomial degree 3 for the velocity field. The interpolation step, provided by this class, takes around 0.54 seconds per time step for `interpolated_euler_step`. This can be compared to around 1.2 seconds for incrementing the `particle_iterator` (aka `std::multimap`) during the same function call. Well, I should add that I increment it twice, once for filling the reference points and once for updating the positions. While this also highlights a deficiency in data structures for the particle storage (a map is far from the fastest option to store this kind of things), it nonetheless shows the potential for the present evaluator.

I am afraid I cannot give much comment here since I am not sufficiently aware of the inner working of FEValues and etc.
In our case, the particle data structure has started to have a significant impact on performance,  but our application relatively far away from what @gassmoeller had originally envisionned that I feel it more our fault than his :).

If me or my team can contribute anything to improving the particle data structure, I would be more than I happy to help!

peterrum (commented):
Looks good.

kronbichler:
I now implemented the alternative code path with `FEValues` for cases not supported by the tensor product, changed the name of the class to `FEPointEvaluation` and added some additional tests. This version is now ready for the first serious review.

I have one thing I want hear opinions on: Right now I use some `EvaluatorTypeTraits` to be able to implement
- a scalar version with `value_type = double` and `gradient_type = Tensor<1,dim>`
- a vector-valued version with `value_type = Tensor<1,dim>` and `gradient_type = Tensor<2, dim>`
- an n-component version with `value_type = Tensor<1,n_components>` and `gradient_type = Tensor<1, n_components, Tensor<1,dim>>`

The advantage is that this allows for really general arguments and evaluations, somewhat similar to the matrix-free `FEEvaluation`. However, the big downside is that I do not immediately see how to enable a user to use really dimension-independent code also in 1D without some code duplication, as the vector-valued case (using `Tensor<2,dim>` for the gradient) collapses with the scalar case (using `Tensor<1,dim>` for the gradient). The alternative would be to disallow the `n_components` case altogether and only provide scalar and vector-valued evaluators. The user would then need to have multiple evaluators for multiple quantities, at the benefit of non-ambiguous arguments also in 1D. Another advantage of the simple scalar/vector case would be that I can move the implementation into a `.cc` file, so we reduce the size of our header files and increase the pre-compiled code.

kronbichler:
> Your current interface seems to support only values and gradients, would hessians or other higher order derivates create a significant problem, or did you just not implement them because they are needed less often? 

I spent some more time to think about this question and I agree that we should foresee a case where more quantities are evaluated. For example, in non-matching computations (and with the 'integration' interface that I still want to introduce, i.e., multiplication by shape function valuess or shape function gradients and summation over all points), one might want to also have the Jacobian of the mapping (`DerivativeForm`) to build some volume elements. Given that an interface with more than three or four optional arguments soon becomes confusing, I would suggest to create a `struct` for the evaluation data that we can extend later, and let it hold the values or gradients; a user can then use those quantities and do whatever is desired. What do you think about that?

Once we have converged on this topic, I believe this PR can be merged. In a follow-up PR, I plan to update the particle tutorials with this evaluator.

kronbichler:
I would like to soon come to a consensus on this PR to be able to use it in applications. Given the previous posts, I would right now favor to only have the `n_components=1` and `n_components=dim` to be able to pre-compile the code. I would also suggest this interface for the evaluation function:
```c++
void
FEPointEvaluation<n_components,dim,spacedim>::evaluate(
     const ArrayView<const Point<dim>> &               unit_points,
     const ArrayView<const double> &                   solution_values,
     const UpdateFlags                                 update_flags,
     FEPointEvaluationData<n_components,dim,spacedim> &evaluation_data) const
```
and similarly
```c++
void
FEPointEvaluation<n_components,dim,spacedim>::integrate(
     const ArrayView<const Point<dim>> &                     unit_points,
     const FEPointEvaluationData<n_components,dim,spacedim> &evaluation_data,
     const UpdateFlags                                       update_flags,
     const ArrayView<double> &                               solution_values) const
```
where this latter multiplies by all the test functions and sums over all given points in the array. There is some slight redundancy in this approach when both `evaluate` and `integrate` are needed because the information is redundantly computed, but I believe this can be tolerated.

peterrum (approved):
Looks good me! I will merge tomorrow if there are no objections. All additional modifications can be done in follow-up PRs! We still have enough time till the next release: I have the feeling that by using the new class in actual applications codes we will be able to finalize the interface!

peterrum:
@kronbichler I would like to propose to switch the second and third argument of `FEPointEvaluation::evaluate()`. Form my perspective, it would make sense to group all arguments that have the same size (i.e., `unit_points`, `function_values`, and `function_gradients`). What do you think?

kronbichler:
Yes, I agree that we should change this because it also make sense to first specify the solution to be evaluated (`solution_values`) and then *where* to evaluate (interpolate) them. That said, I expect this to become less of an issue if we, as suggested in my comment https://github.com/dealii/dealii/pull/11046#issuecomment-724029046, wrap all evaluated data into a separate `struct` to keep the interface lean as we add more functionality.

peterrum:
@kronbichler Yes, a `FEPointEvaluationData` would make sense! I really found it somewhat counter-intuitive that we need to specify both `function_values` and `function_gradients` even if we only need one of them. What is your recommendation regarding this issue and #11459? I could set up a PR that introduces such a data structure and evaluation flags. A general question is where we would resize the vectors?

kronbichler:
I would resize the vectors inside `evaluate` - if the user gives us the number of points, the rest can be done internally according to the given update flags.

peterrum:
I would suggest a slightly different approach by storing `FEPointEvaluationData` inside of `FEPointEvaluationData` (similar to `FEValues`):
```cpp
void
evaluate(
     const ArrayView<const Point<dim>> &               unit_points,
     const ArrayView<const double> &                   solution_values,
     const UpdateFlags                                 update_flags) const;

const value_type
get_value(const unsigned int q) const;

const value_type
set_value(const unsigned int q, const value_type & value) const;
```
Are there any disadvantages here?

kronbichler:
That seems like a good suggestion and would make things even more consistent with what we do elsewhere. I do not see a disadvantage `get_value` and some other things we might request at this point. A possible disadvantage of `set_value` or `submit_value`, if we go by the name used by `FEEValuation` (that is not to say we need to use that exact wording) is that we do not know how big we need to set the array that receives the entries. Since we use `std::vector` under the hood anyway, we could simply add an 
```c++
if(values_array.size() <= q)
  values_array.resize(q+1);
... // go one with what we actually want to do
```
inside the `submit_value` function without affecting performance or functionality. And then inside `integrate` we can only assert
```c++
Assert(values_array.size() >= unit_points.size(),
         ExcMessage("Fewer values seem to have been queued for integration by "
                    "FEPointEvaluation::submit_value() than points have been passed via the "
                    "unit_points argument."));
```
and not the exact same size, but I think for 98% of user mistakes I can think of right now this would be good enough.

kronbichler:
The reason why I initially opted against it is that it adds a bigger interface to the class than passing one array, but once we have a separate data container we might as well go with the more appropriate get/set functions.

peterrum:
> The reason why I initially opted against it is that it adds a bigger interface to the class than passing one array, but once we have a separate data container we might as well go with the more appropriate get/set functions.

Thanks for the clarification!

PR #11047, opened 2020-10-16
MatrixFree: only use partitioner for communication for L:p:V

peterrum:


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -3135,7 +3135,7 @@ namespace internal
     template <typename VectorType>
     unsigned int
     find_vector_in_mf(const VectorType &vec,
-                      const bool        check_global_compatibility = true) const
+                      const bool check_global_compatibility = false) const
```
peterrum, line 3138: I had to set this flag to false, since the test `matrix_free/parallel_multigrid_adaptive_03`  was failing. To me it looks like that the problem is that different MPI communications are mixed up. This thesis is strengthened by the fact that if I put a barrier in `is_globally_compatible` the code dead looks. I couldn't figure the cause of this issue but it looks to me like process 0 is doing something different then the other processes.

kronbichler: Out of curiosity, are other test cases deadlocking as well or just this single `multigrid_adaptive_03` case?

peterrum: All other tests passed. But I have not run them with a barrier. So I cannot tell for sure.

kronbichler (commented):
Looks good to me apart from the fact that I would like to better understand the failure of that one test.

include/deal.II/matrix_free/matrix_free.h
kronbichler: But the `is_globally_compatible` contains an `Allreduce`, so it is a form of barrier, so I would not worry. Can you open an issue that we should look into this test some more at some point?

kronbichler (approved):

PR #11050, opened 2020-10-16
Add documentation for HDF5 and exceptions

dangars:
In certain corner cases HDF5 combined with exceptions can freeze deal.II. This documents the behavior.

This is a follow up to #10746

@drwells I copied pasted your explanation ;)

### DIFF
--- a/include/deal.II/base/hdf5.h
+++ b/include/deal.II/base/hdf5.h
@@ -326,6 +326,38 @@ DEAL_II_NAMESPACE_OPEN
  * thread-safe HDF5 version serializes the API but does not provide any level of
  * concurrency. To achieve high parallel performance with HDF5, we advice to use
  * HDF5 with MPI.
+ *
+ * # Exceptions and parallel HDF5
+ * In certain corner cases, exceptions combined with parallel HDF5 can freeze
+ * deal.II. If one MPI process throws an exception and not the other ones, then
+ * deal.II freezes. The reason is that the following happens:
+ * 1. On the processor that calls the exception we try to close the parallel
+ *   MPI file (H5F_try_close) which ultimately results in a barrier
+ *   (MPI_barrier). This is triggered by the destructor of an HDF5Object.
+ * 2. On the processor that does not call the exception we call
+ *   H5VL_dataset_write() and end up in MPI_Allreduce().
+ *
+ * This could happen if an error is found in all the processes except one. For
+ * example, the following code could potentially freeze deal.II:
+ * @code
+ * HDF5::File data_file(filename,
+ *                            HDF5::File::FileAccessMode::create,
+ *                            mpi_communicator);
+ * {
+ *  auto  dataset_a =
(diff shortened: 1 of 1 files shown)


include/deal.II/base/hdf5.h
```diff
@@ -326,6 +326,16 @@ DEAL_II_NAMESPACE_OPEN
...
  * concurrency. To achieve high parallel performance with HDF5, we advice to use
  * HDF5 with MPI.
+ *
+ * # Exceptions and parallel HDF5
+ * In certain corner cases, exceptions combined with parallel HDF5 can freeze
+ * deal.II. If one MPI process throws an exception and not the other ones then
+ * deal.II freezes. The reason is that the following happens:
+ * 1. On the processor that calls the exception we try to close the parallel
+ *   MPI file (H5F_try_close) which ultimately results in a barrier
+ *   (PMPI_barrier). This is triggered by the destructor of HDF5Object.
```
jppelteret, line 336: `PMPI_barrier` -> `PMPI_Barrier` ?

Also, can you confirm that the "P" prefix is correct. I would have (naively) thought that this should be `MPI_Barrier`.

jppelteret, line 336: `of HDF5Object` -> `of an HDF5Object`

include/deal.II/base/hdf5.h
```diff
@@ -326,6 +326,16 @@ DEAL_II_NAMESPACE_OPEN
...
+ * 2. On the processor that does not call the exception we call
+ *   H5VL_dataset_write and end up in PMPI_allreduce.
```
jppelteret, line 338: Since they're functions, I think that we should write `H5VL_dataset_write()` and `PMPI_Allreduce`.

jppelteret, line 338: `PMPI_allreduce` -> `PMPI_Allreduce` ?

jppelteret (commented):
Thanks for extending the documentation to explain this! Would it be possible to briefly explain some of the typical circumstances (or what you've found through experience) that lead to an exception being thrown? I'm struggling to conceptualise what might lead to such a problem.

drwells:
The `PMPI_` prefix is for the MPI profiling library - we should just write `MPI_` everywhere since that detail isn't relevant to the deadlock.

dangars:
@jppelteret @drwells apologies for the delay. I modified the PR.

@jppelteret The new version of `find_active_cell_around_point()` is not backwards compatible, see #10535. If the point is not in the active cells an exception is thrown. That means that one process can throw an exception and the other processes may not throw the exception. It has not been decided the final behavior of `find_active_cell_around_point()`, that's why I don't mention `find_active_cell_around_point()` in the HDF5 documentation. I added a generic code that freezes deal.II.

drwells (approved):

drwells:
This is a good discussion of the problem with HDF5 - the issues with `find_active_cell_around_point()` are much harder.

PR #11054, opened 2020-10-19
Improve termination criterion for transform_real_to_unit_point

kronbichler:
Besides the termination criterion, this PR also adds code to avoid projecting the initial point to the unit cell because we might often search outside and have a good initial guess there. For Cartesian cases, it is even the *exact* location, so no need for an extra round. But with that change we need a recovery strategy, which adds some code.

The rest of PR is based on the observation that we often did one Newton iteration too much; given that we already have the next right hand side and a very good guess of the inverse Jacobian in case of convergence (which would be the next delta), we should simply use the `f_weighted_norm_square` from the line search as the termination, rather than the update from the current step (or the previous step's residual).

Furthermore, we can run into roundoff issues if we search outside the unit cell for higher order mappings; in that case, we often trigger the line search but do not make success. I now define a weighted tolerance of 1e-6 (in reference coordinates) as the threshold under which we accept step length 1 in any case.


tjhei:
I assume that it is difficult to turn your experiments into tests directly, but I think it would be great to have a way to check that we are not regressing in other situations. Maybe we can introduce a way to record convergence information in debug mode?

kronbichler:
> Maybe we can introduce a way to record convergence information in debug mode?

I would have loved to have such a test. (That said, the test `mappings/mapping_fe_field_08` turned out to be super useful because it has a point 2.1e-8 outside of a cell but we try `find_all_active_cells_around_point` with tolerance 1e-8 so it tests all kinds of points outside the unit cell, many of those far outside.)
The easiest way to achieve this would be to move the function https://github.com/dealii/dealii/blob/ee8a81293cda97756517d5e7e81a35c2324e5453/source/fe/mapping_q_generic.cc#L759-L780 to a header file, say `mapping_q_internal.h`, add another flag to print information like https://github.com/dealii/dealii/blob/ee8a81293cda97756517d5e7e81a35c2324e5453/source/fe/mapping_q_generic.cc#L892-L898 to deallog, and pick some nice points/cells from `mapping_fe_field_08`. It would be good to move some functions out of `mapping_q_generic.cc` because it is a monster file already.

masterleinad (approved):

tjhei:
> The easiest way to achieve this would be to move the function

That would be amazing! Do you think the tests would be robust enough to run in CI? If not, we can have them as tests without checking this output but we can require anybody to run them before/after when proposing changes to this function.

kronbichler:
I now refactored the internal functions of `mapping_q_generic.cc` into a new file `mapping_q_internal.h`. This allows me to call into the `do_transform_real_to_unit_cell_internal` function and monitor the output. I also added a test with some points to cover the most important cases of the inner working. I should add that I had to play a bit with the position and which points to allow to not trigger roundoff differences cause additional iterations for the case of FMA vs no FMA. But that setup remains sensitive by nature, so we have to keep an eye on this test or add additional blessed output files.

kronbichler:
For reviewing purposes: The only change in functionality is commit 1; commit 2 adjusts some tests that now failed with the new way of computing points (because the previous implementation would do some line search and bail out for cases where it should not), commit 3 does the refactoring (and replaces the `#ifdef` by conditional output), and commit 4 adds the new test case.

Issue #11055, opened 2020-10-19
TransfiniteInterpolationManifold should copy (coarse) triangulation


tamiko:
I am currently having troubles with a `TransfiniteInterpolationManifold` and the following use case (simplified):

```
constexpr unsigned int dim = 2;

Triangulation<dim> coarse_triangulation;
GridGenerator::hyper_cube(coarse_triangulation);

coarse_triangulation.set_all_manifold_ids(9);
dealii::TransfiniteInterpolationManifold<2> transfinite;
transfinite.initialize(coarse_triangulation);
coarse_triangulation.set_manifold(9, transfinite);

Triangulation<dim> triangulation;
triangulation.copy_triangulation(coarse_triangulation);

triangulation.refine_global(1);
```

This (seemingly innocuous code) fails for two reasons:

(a) It will trigger a (somewhat performance critical) check in `TransfiniteInterpolationManifold<dim, spacedim>::get_possible_cells_around_points()`:
```
for (; cell != endc; ++cell)
  // only consider cells where the current manifold is attached
  if (&cell->get_manifold() != this)
    continue;
```
Here, the problem is that the `copy_triangulation` operation creates a bona fide copy of the manifold object and thus `&cell->get_manifold()` and `this` won't be the same object any more.

(b) The manifold stores a reference to the triangulation, so in above code suppose that I would retain the `triangulation` object but `coarse_triangulation` gets destroyed, the manifold can no longer function because it loses access to the coarse triangulation it was originally initialized on.

The underlying core problem in my opinion is that the `TransfiniteInterpolationManifold` actually stores a reference to the triangulation (that it must be attached to). I undestand why we are doing this - it is the a straight forward solution for the problem that the `TransfiniteInterpolationManifold` must be able to access the coarse grid and all attached manifolds.

Does something speak against copying the coarse grid information, i.e., all relevant coarse cells of the triangulation in addition to all relevant manifolds when `initialize` is called? This would solve both problems above in a (for the user) very convenient manner.

(((
Just in case someone is curious why I want to have use case (a): I need to create a coarse grid with some anisotropic pre refinement, a call to`flatten_triangulation`, and finally copying the final result into a (p4est) distributed triangulation. I have a heavily curved mesh and I want to have a transfinite interpolation between the outer curved boundary and an inner curved boundary (that is an obstacle). Recreating a TransfiniteInterpolationManifold for the (final) coarse mesh leads to very suboptimal results because coarse cells don't reach both boundaries any more.
)))

[...]

PR #11058, opened 2020-10-20
TransfiniteInterpolationManifold: copy triangulation

tamiko:
Closes: https://github.com/dealii/dealii/pull/10971
Closes: https://github.com/dealii/dealii/issues/11055


source/grid/manifold_lib.cc
```diff
@@ -2225,29 +2217,21 @@ TransfiniteInterpolationManifold<dim, spacedim>::
...
-                                                           level_coarse);
+  auto       cell = triangulation.begin(level_coarse);
+  const auto endc = triangulation.end(level_coarse);
   boost::container::small_vector<std::pair<double, unsigned int>, 200>
     distances_and_cells;
   for (; cell != endc; ++cell)
     {
-      // only consider cells where the current manifold is attached
-      if (&cell->get_manifold() != this)
-        continue;
```
tamiko, line 2249: @kronbichler At the time of copying the coarse triangulation in `initialize()` we do not have any information about where the manifold is getting attached to. I.e., we actually cannot query for `&cell->get_manifold() != this`.

source/grid/manifold_lib.cc
```diff
@@ -2225,29 +2217,21 @@ TransfiniteInterpolationManifold<dim, spacedim>::
...
   // for the active cells, but we are here looking at some cells at the coarse
   // level.
-  Assert(triangulation != nullptr, ExcNotInitialized());
-  Assert(triangulation->begin_active()->level() >= level_coarse,
-         ExcMessage("The manifold was initialized with level " +
-                    std::to_string(level_coarse) + " but there are now" +
-                    "active cells on a lower level. Coarsening the mesh is " +
-                    "currently not supported"));
+
+  // FIXME IMPROVE
```
tamiko, line 2221: @kronbichler I am not sure whether we can actually speed up the coarse cell loop below in a meaningfully way - we actually loop over all coarse cells in order to construct a full list of candidate cells.

kronbichler: We certainly could by creating bounding boxes (or even an RTree) that allows us to identify a (small list) of cells as candidates for a given set of points. That would help tremendously in case we have many coarse mesh cells.

tamiko:
```
The following tests FAILED:
	546 - grid/merge_triangulations_05.debug (Failed)
	575 - grid/merge_triangulations_04.debug (Failed)
	589 - grid/grid_generator_plate_with_a_hole.debug (Failed)
	622 - grid/grid_generator_plate_with_a_hole_2.debug (Failed)
	713 - grid/grid_generator_eccentric_hyper_shell.debug (Failed)
	750 - grid/grid_generator_plate_with_a_hole_4.debug (Failed)
	819 - grid/grid_generator_plate_with_a_hole_3.debug (Failed)
```

jppelteret:
Question: What would the implications of such a change be for people using many `TransfiniteInterpolationManifolds` at one time? e.g. microstructural studies
Random_microstructure-250um-Magnetic_field-Periodic

tamiko:
@kronbichler *mhm* Most of the remaining tests now fail because a different (and probably incorrect) coarse cell is used for the transformation leading to relative differences in vertex positions in the 0.1% range.

I don't see how we can resolve this without changing the interface of the `TransfiniteInterpolationManifold` to know on which coarse cells it should work on.

tamiko:
@jppelteret Ideally there would be no change.

tamiko:
@jppelteret But - honestly, let me not touch it if you have concerns.

jppelteret:
> But - honestly, let me not touch it if you have concerns.

No, I'm ambivalent to the change. My only concern is for cases where the coarse triangulation is not very coarse due to meshing difficulties, and there are lots of copies of the triangulation that need to be made. If it were possible for all TSI manifolds to somehow share the same copy of the triangulation then that would be good. So, perhaps passing a cache-like data structure that they share?

kronbichler:
> Let's not touch this for the time being. Too much pain in trying to get it work for all use cases again.

> I don't see how we can resolve this without changing the interface of the TransfiniteInterpolationManifold to know on which coarse cells it should work on.

I tend to agree on this assessment; the class is indeed brittle because of the multiple coarse cell state. I would prefer if we could somewhat automatically create a separate manifold (i.e., a different `Manifold` object) attached to each coarse mesh cell in combination with the surrounding manifold descriptions. But it's not entirely clear how to get there, because also faces can have transfinite interpolation that get fed by the surrounding edges, so even faces would need such a number.

On the other hand, I tend to see transfinite interpolation as the *default* manifold that jumps whenever there is not other manifold set by the user, i.e., do the job that `FlatManifold` does now. This is because transfinite interpolation is not a description of surface or volume per se but merely an averaging of things located around a *cell*. This makes it even more ambiguous to find a reasonable averaging method at places where we query manifolds. Well, unless we directly inject the *surrouding cell* to averaging operation we ask the transfinite interpolation to perform.

PR #11059, opened 2020-10-20
Let PropertyPool keep track of the memory it allocates.

bangerth:
This allows us to also error out if memory is allocated but not released.

This also exposed a bug: We destroy the property pool before we destroy the particles that still reference it. The first of the commits addresses this issue.


tamiko:
@bangerth Tester fails due to missing include:
```
 In file included from /home/runner/work/dealii/dealii/include/deal.II/particles/particle.h:25:0,
                 from /home/runner/work/dealii/dealii/source/particles/particle.cc:18:
/home/runner/work/dealii/dealii/include/deal.II/particles/property_pool.h:117:10: error: ‘set’ in namespace ‘std’ does not name a template type
     std::set<Handle> currently_open_handles;
          ^~~
```

bangerth:
Turns out that the patch uncovered a second bug: When we call `Particle::set_property_pool()`, we need to release memory held in an old memory pool if one was associated with the particle. This is now in the second commit of this series.

bangerth:
I also fixed the missing include file.

masterleinad (approved):

bangerth:
Seems to work now.

kronbichler:
Now that our cdash is back online, it seems to trigger on this PR (or one of mine that changed the internal workings of `transform_real_to_unit_cell`, but I can't reproduce those from my changes):
https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=12297
```
  particle_handler_08
  particle_handler_11
  particle_handler_12
  particle_handler_13
  particle_handler_15
  particle_handler_16
  particle_handler_17
```
The other four tests are unrelated, two of those long-time failures on that machine.

PR #11063, opened 2020-10-20
Ensure all communication is finished in matrix-free loops

### LINKED ISSUE #11052: Enable check_global_compatibility in find_vector_in_mf again
As discussed in https://github.com/dealii/dealii/pull/11047#discussion_r506610584,  test `matrix_free/parallel_multigrid_adaptive_03` fails if `Utilities::MPI::Partitioner::is_globally_compatible()` is called. By disabling the function call, the test is running as expected again. We need to figure out the cause for this behaviour.

FYI @kronbichler

kronbichler:
Should correct the issue that #11062 tries to work around. Fixes #11052. @peterrum can you please test this code as an alternative?

### DIFF
--- a/source/matrix_free/task_info.cc
+++ b/source/matrix_free/task_info.cc
@@ -353,7 +353,7 @@ namespace internal
       if (scheme != none)
         {
           funct.zero_dst_vector_range(numbers::invalid_unsigned_int);
-          if (scheme == partition_partition)
+          if (scheme == partition_partition && evens > 0)
             {
               tbb::empty_task *root =
                 new (tbb::task::allocate_root()) tbb::empty_task;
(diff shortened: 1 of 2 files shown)


source/matrix_free/task_info.cc
```diff
@@ -413,6 +413,13 @@ namespace internal
               root->wait_for_all();
               root->destroy(*root);
             }
+          else if(scheme == partition_partition)
+            {
+              // catch the case of empty partition list where we still finish
+              // the exchange of ghost data
+              funct.vector_update_ghosts_finish();
+              funct.vector_compress_start();
+            }
```
kronbichler, line 422: The problem was that in the other code we would not even enter the code that set up some of the communication because it sits inside a loop over cell blocks - without cells, there is no loop iteration that could execute the code.

peterrum (approved):
Yes, this works. May I ask you to copy my cosmetic changes in `matrix_free.h` from #11062. Else I can do that in a follow-up PR.

kronbichler:
I can add that code; one thing I realize now is that #11047 forces the distributed vector to be compatible with what is given in `DoFInfo`; before we could also handle the case that a distributed vector is coming from somewhere else. While I agree that such a vector should not be handled through `MF::loop` whose purpose is to update those vectors, it might force us to write an incompatibility changelog.

peterrum (approved):
Thanks! I'll add the change-log entry in a follow-up PR.

PR #5090, opened 2017-09-14
Be more generous in what types we allow in the FESystem(...) constructor

bangerth:
This is a follow-up to the new variadic constructor of `FESystem` introduced in #5026: We now allow the argument list to be composed of either (i) pairs of a pointer to finite element and multiplicity, or (ii) just finite element. This allows writing things such as
```
  FESystem (FE_Q(2)^dim, FE_Q(1));
```
where the second argument has no multiplicity.

This works by routing each argument to a function `promote_to_pair()` that just forwards the original pair if it really is a pair, or that creates a pair of that element and multiplicity one if it is not a pair.

The last commit extends this a bit by adding another function `promote_to_pair()` that just eats all other possible argument types but has a `static_assert` that explains in detail that you can't do that. For example, if you pass an integer as an argument, you'd get this sort of error message:
```
40: In file included from /home/fac/f/bangerth/p/deal.II/1/dealii/tests/fe/system_04.cc:25:0:
40: /home/fac/f/bangerth/p/deal.II/1/dealii/include/deal.II/fe/fe_system.h: In function ‘std::pair<std::unique_ptr<dealii::FiniteElement<<anonymous>, <anonymous> > >, unsigned int> dealii::{anonymous}::promote_to_fe_pair(const T&)’:
40: /home/fac/f/bangerth/p/deal.II/1/dealii/include/deal.II/fe/fe_system.h:1183:5: error: static assertion failed: The arguments to the FESystem constructor that takes a variable number of arguments all need to either be finite element objects, or pairs of a std::unique pointer to a finite element and an integer multiplicity (i.e., what you get when you write, for example, (FE_Q<dim>(2)^3'). But here, you are calling that constructor with an argument that is neither of these two options.
40:      static_assert (false,
40:      ^
```

There is a new test, `system_04` that is a variation of `system_03` that just omits the `^1` cases. The output is identical to the previous test.

PR #5026, opened 2017-09-04
Variadic constructor for FESystem

masterleinad:
Fixes #4249. Fixes #2299.
This PR allows to initialize `FESystem` objects as suggested in https://github.com/dealii/dealii/issues/2299#issuecomment-197087786, i.e.
```
FESystem<dim> stokes_fe = {FE_Q<dim>(2) ^ dim, FE_Q<dim>(1)^1};
```
At the moment all the calls are just forwarded to the constructor taking a `std::vector< const FiniteElement< dim, spacedim > * >` and a `std::vector< unsigned int >`.
This could probably be written more efficiently rewriting the whole logic but should be fine for now.

PR #11066, opened 2020-10-20
Use operator^ to create FESystem objects in the tutorials.

bangerth:
A student of mine (@juleoc02) showed me some code in which he created an `FESystem` object with a syntax I did not remember. But it is true that we have an `operator^`  (#5026 by @masterleinad plus follow-up #5090) that can be used in this context, and that moves the notation we use closer to what we write in mathematical notation. 

This patch whole-sale converts all of the tutorials to this notation. step-8 is the first example where we use this, so I added a bit of explanation there to what is happening.

### DIFF
--- a/examples/step-46/step-46.cc
+++ b/examples/step-46/step-46.cc
@@ -220,18 +220,12 @@ namespace Step46
     : stokes_degree(stokes_degree)
     , elasticity_degree(elasticity_degree)
     , triangulation(Triangulation<dim>::maximum_smoothing)
-    , stokes_fe(FE_Q<dim>(stokes_degree + 1),
-                dim,
-                FE_Q<dim>(stokes_degree),
-                1,
-                FE_Nothing<dim>(),
-                dim)
-    , elasticity_fe(FE_Nothing<dim>(),
-                    dim,
-                    FE_Nothing<dim>(),
-                    1,
-                    FE_Q<dim>(elasticity_degree),
-                    dim)
+    , stokes_fe(FE_Q<dim>(stokes_degree + 1) ^ dim, // for the fluid velocity
+                FE_Q<dim>(stokes_degree),           // for the fluid pressure
+                FE_Nothing<dim>() ^ dim)          // for the solid displacement
+    , elasticity_fe(FE_Nothing<dim>() ^ dim,      // for the fluid velocity
+                    FE_Nothing<dim>(),            // for the fluid pressure
+                    FE_Q<dim>(elasticity_degree)) // for the solid displacement
     , dof_handler(triangulation)
     , viscosity(2)
     , lambda(1)
--- a/examples/step-8/step-8.cc
+++ b/examples/step-8/step-8.cc
@@ -189,12 +189,26 @@ namespace Step8
   // the solution function has, which is <code>dim</code> since we consider
   // displacement in each space direction. The FESystem class can handle this:
   // we pass it the finite element of which we would like to compose the
-  // system of, and how often it shall be repeated:
-
+  // system of, and how often to repeat it. There are different ways to
+  // tell the FESystem constructor how to do this, but the one that is
+  // closest to mathematical notation is to write out what we want to do
+  // mathematically: We want to construct the finite element space
+  // $Q_1^d$ where the index 1 corresponds to the polynomial degree and
+  // the exponent $d$ to the space dimension -- because the *displacement*
+  // we try to simulate here is a vector with exactly $d$ components. The
+  // FESystem class then lets us create this space by initialization with
+  // `FE_Q<dim>(1)^dim`, emulating the mathematical notation.
+  //
+  // (We could also have written `fe(FE_Q<dim>(1), dim)`, which would simply
+  // have called a different constructor of the FESystem class that first
+  // takes the "base element" and then a "multiplicity", i.e., a number that
+  // indicates how many times the base element is to be repeated. The two
+  // ways of writing things are entirely equivalent; we choose the one that
+  // is closer to mathematical notation.)
   template <int dim>
   ElasticProblem<dim>::ElasticProblem()
     : dof_handler(triangulation)
-    , fe(FE_Q<dim>(1), dim)
+    , fe(FE_Q<dim>(1) ^ dim)
   {}
   // In fact, the FESystem class has several more constructors which can
   // perform more complex operations than just stacking together several
--- a/examples/step-17/step-17.cc
+++ b/examples/step-17/step-17.cc
@@ -225,7 +225,7 @@ namespace Step17
   // The first step in the actual implementation is the constructor of
   // the main class. Apart from initializing the same member variables
   // that we already had in step-8, we here initialize the MPI
-  // communicator variable we shall use with the global MPI
+  // communicator variable we want to use with the global MPI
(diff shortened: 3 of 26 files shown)


masterleinad (approved):

examples/step-8/step-8.cc
```diff
@@ -193,12 +193,26 @@ namespace Step8
...
+  // we try to simulate here is a vector with exactly $d$ components. The
+  // FESystem class then lets us create this space by initialization with
+  // `FE_Q<dim>(1)^dim`, emulating the mathematical notation.
+  //
+  // (We could also have written `fe(FE_Q<dim>(1), dim)`, which would simply
+  // have called a different constructor of the FESystem class that first
+  // takes the "base element" and then a "multiplicity", i.e., a number that
+  // indicates how many times the base element is to be repeated. The two
+  // ways of writing things are entirely equivalent; we choose the one that
+  // is closer to mathematical notation.)
```
peterrum, line 211: Personally, I would switch this. For me the `operator^` is only syntactic sugar (and just an additional layer of indirection). I would show the `operator^` here and keep the *normal* way in all other tutorials. Furthermore, it looks to me that not all compilers are supporting this feature:
https://github.com/dealii/dealii/blob/53084d8c6a5296b727b1d73c5450fe43d4f5d780/include/deal.II/fe/fe_system.h#L499-L502

bangerth: I think the question is what we want the *definitive* way for things to be. My take is to use the `operator^` from here on out. There is not much use in using two different ways to achieve the same thing in the tutorials: I've come to understand that from a teaching perspective, it's useful to pick one way and stick with it, even if there may be other ways to achieve things. We didn't follow this principle in the early years and introduced different ways of doing things just because we could, and I think that's not a good approach and over the years we've tried to unify things on one style.

As for the Intel compiler: What version are they on now? What version do typical supercomputing centers have on their machines?

examples/step-46/step-46.cc
```diff
@@ -224,18 +224,12 @@ namespace Step46
...
-                    FE_Nothing<dim>(),
-                    1,
-                    FE_Q<dim>(elasticity_degree),
-                    dim)
+    , stokes_fe(FE_Q<dim>(stokes_degree + 1) ^ dim, // for the fluid velocity
+                FE_Q<dim>(stokes_degree),           // for the fluid pressure
+                FE_Nothing<dim>() ^ dim)          // for the solid displacement
+    , elasticity_fe(FE_Nothing<dim>() ^ dim,      // for the fluid velocity
+                    FE_Nothing<dim>(),            // for the fluid pressure
+                    FE_Q<dim>(elasticity_degree)) // for the solid displacement
```
jppelteret, line 232: It looks like the last four comments on the RHS could do with a little nudge so that they all line up.

examples/step-8/step-8.cc
jppelteret: I like the change, and I think that if we can use it everywhere then we should (modernisation of the tutorials is not a bad thing, and the alternative approach that calls the other constructor is mentioned here). I hope that the questions surrounding the compiler incompatibility are not a deal-breaker.

jppelteret (approved):

examples/step-46/step-46.cc
bangerth: That would require getting rid of a certain tool that does this automatically in a way so that the last comment in the first group of 3 does not align with the previous two lines :-(

jppelteret: I guess I walked into that one. For some reason (and I don't quite know why) I thought that CT let one manage the indentation of end-line comments one's self.

jppelteret:
I'm going to mark this as "do not merge" for now, while we come to some determination about the compiler compatibility issue.

examples/step-46/step-46.cc
bangerth: :-)

bangerth:
What is the consensus among developers: As @peterrum correctly points out (https://github.com/dealii/dealii/pull/11066#discussion_r508985433), I'm changing tutorial programs to use an `FESystem` constructor that is disabled for Intel icc 19 and earlier. In essence, merging this patch would then require a newer version of this compiler. Alternatively, we could try out whether we can find a syntax that's acceptable to icc.

What icc versions are people using?

tjhei:
> What icc versions are people using?

Intel 19 is the default on machines like TACC Frontera (at least it was earlier this year).

tamiko (approved):

tamiko:
I like this change.

I do not think that we are losing much if our examples do not run on pre icc-19 out of the box. For starters, icc-19 is already two years old. Secondly, icc users are a bit of a niche and can easily change this one line of code in their program.

It just stands that when using icc you're already in for a rough ride - we had crazy miscompilations of avx2/avx512 features with older icc versions, etc. `operator^` not being supported to due an internal compiler error is just the tip of the iceberg there.

And I think there is really something to be gained from improving our syntax.

kronbichler:
> I do not think that we are losing much if our examples do not run on pre icc-19 out of the box.

Would we plan to disable the automatic compilation of tutorials with old icc then?

bangerth:
I wonder whether we can try again to find a syntax that icc actually accepts. I remember that we didn't try very hard at the time because we didn't want to use the new syntax right away. Would someone want to give it another try and post whatever error messages we get?

tjhei:
> Would we plan to disable the automatic compilation of tutorials with old icc then?

We should, otherwise people will be stuck in a situation where "make install" will not succeed.

I have to admit that I don't think the ^ notation adds a lot. I would have a harder time explaining to my students that there is an operator overload that creates a temporary object, etc.. than just saying ", dim mean dim copies". But that is just me. I would even prefer parsing from a string than using this notation.

bangerth:
For the record, current Intel compilers (December 2020) are based on clang and are happy with the code as proposed here. I'll continue to try (see #11766) to get this working with icc 19.x as well.

bangerth:
So I tried this with the Intel ICC 19.1 compiler. This works, and this also makes sense because the relevant constructors are all guarded by
```
#  if !defined(__INTEL_COMPILER) || __INTEL_COMPILER >= 1900
```
so this seems to work starting from ICC 19.0. From https://en.wikipedia.org/wiki/Intel_C%2B%2B_Compiler#Release_history I see that 19.0 was released in April 2018. (For comparison, GCC 8.x was released in May 2018.)

So I think the question comes down to this: Are there pre-19.0 compilers that are still widely used, and how do we feel about supporting them?

Rombur:
> Are there pre-19.0 compilers that are still widely used, and how do we feel about supporting them?

There was a question on the mailing list earlier this week where someone was using ICC 18.0 on stampede2. So people are using older version of ICC.

bangerth:
Let's postpone the issue to post-9.3 then. We can revisit whether we want to use this for the 2022 release when the number of systems using a compiler more than 4 years old will probably be close to zero.

examples/step-46/step-46.cc
masterleinad: You can always use `clang-format off` in places where you want to be explicit (and probably filter it out when running doxygen).

tamiko (approved):

bangerth:
Oh, not even a discussion about whether that's the right direction? :-)

PR #11071, opened 2020-10-21
Partitioner + L:d:V: new constructor

peterrum:
This PR adds new constructors that allow to explicitly specify the number of locally-owned degrees of freedom and the number of ghost degrees of freedom.

### DIFF
--- a/include/deal.II/base/partitioner.h
+++ b/include/deal.II/base/partitioner.h
@@ -205,6 +205,24 @@ namespace Utilities
       Partitioner(const unsigned int size);
 
       /**
+       * Constructor that takes the number of locally-owned degrees of freedom
+       * @p local_size and the number of ghost degrees of freedom @p ghost_size.
+       *
+       * The local index range is translated to global indices in an ascending
+       * and one-to-one fashion, i.e., the indices of process $p$ sit exactly
+       * between the indices of the processes $p-1$ and $p+1$, respectively.
+       *
+       * @note Setting the @p ghost_size variable to an appropriate value
+       *   provides memory space for the ghost data in a vector's memory
+       *   allocation as and allows access to it via local_element(). However,
+       *   the associated global indices must be handled externally in this
+       *   case.
+       */
+      Partitioner(const types::global_dof_index local_size,
+                  const types::global_dof_index ghost_size,
+                  const MPI_Comm                communicator);
+
+      /**
        * Constructor with index set arguments. This constructor creates a
        * distributed layout based on a given communicators, an IndexSet
        * describing the locally owned range and another one for describing
--- a/include/deal.II/lac/la_parallel_vector.h
+++ b/include/deal.II/lac/la_parallel_vector.h
@@ -363,6 +363,24 @@ namespace LinearAlgebra
         const std::shared_ptr<const Utilities::MPI::Partitioner> &partitioner);
 
       /**
+       * Initialize vector with @p local_size locally-owned and @p ghost_size
+       * ghost degrees of freedoms.
+       *
+       * @note In the created underlying partitioner, the local index range is
+       *   translated to global indices in an ascending and one-to-one fashion,
+       *   i.e., the indices of process $p$ sit exactly between the indices of
+       *   the processes $p-1$ and $p+1$, respectively. Setting the
+       *   @p ghost_size variable to an appropriate value provides memory space
+       *   for the ghost data in a vector's memory allocation as and allows
+       *   access to it via local_element(). However, the associated global
+       *   indices must be handled externally in this case.
+       */
+      void
+      reinit(const types::global_dof_index local_size,
+             const types::global_dof_index ghost_size,
+             const MPI_Comm                comm);
+
+      /**
        * Swap the contents of this vector and the other vector @p v. One could
        * do this operation with a temporary variable and copying over the data
        * elements, but this function is significantly more efficient since it
--- a/include/deal.II/lac/la_parallel_vector.templates.h
+++ b/include/deal.II/lac/la_parallel_vector.templates.h
@@ -440,6 +440,32 @@ namespace LinearAlgebra
 
 
     template <typename Number, typename MemorySpaceType>
+    void
+    Vector<Number, MemorySpaceType>::reinit(
+      const types::global_dof_index local_size,
+      const types::global_dof_index ghost_size,
+      const MPI_Comm                comm)
+    {
+      clear_mpi_requests();
+
+      // check whether we need to reallocate
+      resize_val(local_size + ghost_size);
+
+      // delete previous content in import data
+      import_data.values.reset();
+      import_data.values_dev.reset();
+
+      // create partitioner
+      partitioner = std::make_shared<Utilities::MPI::Partitioner>(local_size,
+                                                                  ghost_size,
+                                                                  comm);
+
+      this->operator=(Number());
+    }
+
+
+
+    template <typename Number, typename MemorySpaceType>
     template <typename Number2>
     void
     Vector<Number, MemorySpaceType>::reinit(
--- a/source/base/partitioner.cc
+++ b/source/base/partitioner.cc
@@ -58,6 +58,40 @@ namespace Utilities
 
 
 
+    Partitioner::Partitioner(const types::global_dof_index local_size,
+                             const types::global_dof_index ghost_size,
+                             const MPI_Comm                communicator)
+      : global_size(Utilities::MPI::sum<types::global_dof_index>(local_size,
+                                                                 communicator))
+      , locally_owned_range_data(global_size)
+      , local_range_data{0, local_size}
+      , n_ghost_indices_data(ghost_size)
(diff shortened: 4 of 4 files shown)


include/deal.II/base/partitioner.h
```diff
@@ -204,6 +204,16 @@ namespace Utilities
...
 
+      /**
+       * Constructor that takes the number of locally-owned degrees of freedom
+       * @p and the number of ghost degrees of freedom @p ghost_size.
+       *
+       * @note An interpretation of the ghost indices is not possible.
+       */
+      Partitioner(const unsigned int size,
+                  const unsigned int ghost_size,
+                  const MPI_Comm     communicator);
```
kronbichler, line 215: I realize this is the same type as in the other constructor, but I guess it would make sense to make these `types::global_dof_index`.

include/deal.II/base/partitioner.h
```diff
@@ -204,6 +204,16 @@ namespace Utilities
        */
       Partitioner(const unsigned int size);
 
+      /**
+       * Constructor that takes the number of locally-owned degrees of freedom
+       * @p and the number of ghost degrees of freedom @p ghost_size.
```
kronbichler, line 209:
```suggestion
       * @p size and the number of ghost degrees of freedom @p ghost_size.
       * The local index range is translated to global indices in an ascending and
       * one-to-one fashion, i.e., the indices of process $p$ sit exactly between the
       * indices of the processes $p-1$ and $p+1$, respectively.
```

include/deal.II/base/partitioner.h
```diff
@@ -204,6 +204,16 @@ namespace Utilities
...
+       *
+       * @note An interpretation of the ghost indices is not possible.
```
kronbichler, line 211:
```suggestion
       * @note This way of setting up the partitioner does not allow for an identification of ghost indices on remote processes.
```

include/deal.II/lac/la_parallel_vector.h
```diff
@@ -346,6 +346,20 @@ namespace LinearAlgebra
...
              const MPI_Comm  communicator);
 
+      /**
+       * Initialize vector with @p local_size locally-owned and @p ghost_size
+       * ghost degrees of freedoms.
+       *
+       * @note The created underlying partitioner is set up so that the
+       *   number of locally-owned and ghost degrees of freedom match the given
+       *   parameters. However, an interpretation of the ghost indices is not
+       *   possible.
```
kronbichler, line 356: Update with information on ascending + 1-to-1 distribution of indices similar to the constructor.

include/deal.II/lac/la_parallel_vector.h
```diff
@@ -346,6 +346,20 @@ namespace LinearAlgebra
...
+       */
+      void
+      reinit(const size_type local_size,
+             const size_type ghost_size,
+             const MPI_Comm  comm);
```
kronbichler, line 361: Did you forget to include the definition of this function? I cannot find it below.

source/base/partitioner.cc
```diff
@@ -58,6 +58,28 @@ namespace Utilities
...
 
 
+    Partitioner::Partitioner(const unsigned int size,
+                             const unsigned int ghost_size,
+                             const MPI_Comm     communicator)
+      : global_size(
+          Utilities::MPI::sum<types::global_dof_index>(size, communicator))
+      , locally_owned_range_data(size)
+      , local_range_data(
+          std::pair<types::global_dof_index, types::global_dof_index>(0, size))
```
kronbichler, line 68: This does not really correspond to the 1-to-1 and ascending indices I was expecting and what the line immediately above regarding `global_size` would suggest. Should this include the result of a prefix sum? I assume you do not really use the global indices or `Vector::operator()(types::global_dof_index)` functions because they would not have the right index.

source/base/partitioner.cc
```diff
@@ -58,6 +58,28 @@ namespace Utilities
...
+      , n_ghost_indices_data(ghost_size)
+      , n_import_indices_data(0)
+      , n_ghost_indices_in_larger_set(0)
+      , my_pid(Utilities::MPI::this_mpi_process(communicator))
+      , n_procs(Utilities::MPI::n_mpi_processes(communicator))
+      , communicator(communicator)
+      , have_ghost_indices(true)
+    {
+      locally_owned_range_data.add_range(0, size);
```
kronbichler, line 77: Same here.

include/deal.II/lac/la_parallel_vector.h
peterrum: The implementation is on include/deal.II/lac/la_parallel_vector.templates.h.

peterrum:
@kronbichler Thanks for the review! I have addressed the comments!

source/base/partitioner.cc
```diff
@@ -58,6 +58,39 @@ namespace Utilities
...
 
 
+    Partitioner::Partitioner(const types::global_dof_index local_size,
+                             const types::global_dof_index ghost_size,
+                             const MPI_Comm                communicator)
+      : global_size(Utilities::MPI::sum<types::global_dof_index>(local_size,
+                                                                 communicator))
+      , locally_owned_range_data(global_size)
+      , local_range_data(std::pair<types::global_dof_index,
+                                   types::global_dof_index>(0, local_size))
```
kronbichler, line 68: Do you want to use `[prefix_sum, prefix_sum + local_size)` here as well?

peterrum: It is overridden later.

kronbichler: I would prefer to set `local_range_data` in this function already to avoid an invalid state; all I see here is `locally_owned_range_data` but `local_range_data != locally_owned_range_data` (even though one of the two is redundant). But we could postpone it to another PR in case this function is reworked again.

include/deal.II/base/partitioner.h
```diff
@@ -204,6 +204,21 @@ namespace Utilities
        */
       Partitioner(const unsigned int size);
 
+      /**
+       * Constructor that takes the number of locally-owned degrees of freedom
+       * @p local_size and the number of ghost degrees of freedom @p ghost_size.
+       *
+       * The local index range is translated to global indices in an ascending
+       * and one-to-one fashion, i.e., the indices of process $p$ sit exactly
+       * between the indices of the processes $p-1$ and $p+1$, respectively.
```
tjhei, line 213: This does not explain what the ghost indices are. Can you add that? Maybe just "can be set using set_ghost_indices()"?

source/base/partitioner.cc
peterrum: You are right. I got confused by the similar names. What about the version now?

include/deal.II/lac/la_parallel_vector.templates.h
```diff
@@ -439,6 +439,32 @@ namespace LinearAlgebra
...
+      clear_mpi_requests();
+
+      // check whether we need to reallocate
+      resize_val(local_size + ghost_size);
+
+      // delete previous content in import data
+      import_data.values.reset();
+      import_data.values_dev.reset();
+
+      // set partitioner to serial version
```
tjhei, line 458: outdated comment?

source/base/partitioner.cc
```diff
@@ -58,6 +58,40 @@ namespace Utilities
...
+      types::global_dof_index prefix_sum = 0;
+
+#ifdef DEAL_II_WITH_MPI
+      MPI_Exscan(&local_size,
+                 &prefix_sum,
+                 1,
+                 Utilities::MPI::internal::mpi_type_id(&prefix_sum),
+                 MPI_SUM,
+                 communicator);
+#endif
```
tjhei, line 85: #else ? Do we need an error check?

include/deal.II/lac/la_parallel_vector.templates.h
peterrum: I have fixed this.

source/base/partitioner.cc
peterrum: No. The expression works nevertheless: `prefix_sum` is set to 0 before this section and if the section is not entered the value  stays the same, which is fine in the serial case.

include/deal.II/base/partitioner.h
peterrum: The comment below
```cpp
       * @note: This way of setting up the partitioner does not allow for an
       *   identification of ghost indices on remote processes.
```
already states that the users cannot access the ghost indices (only the ghost values).

kronbichler: You could maybe add: `Setting this variable to an appropriate value provides memory space for the ghost data in a vector's memory allocation as and allows access to it via local_element(). However, the associated global indices must be handled externally in this case.`

kronbichler (approved):

include/deal.II/base/partitioner.h
peterrum: I have updated the comment.

kronbichler (approved):
This looks good to me.

PR #11081, opened 2020-10-22
MatrixFree: restructure fe_face_evaluation_process_and_io

peterrum:
This PR restructures case 4 (contiguous indices without interleaving) as preparation of the integration of usage of the new MPI-3 features of L:p:V.

The code section is structured as following. Either values+grad is evaluated in the Hermite case or the values in the case of the nodal basis. For both following three evaluation options are available:
1. vectorized version (all lanes are filled, all lanes have the same orientation)
2. only some lanes are filled but all filled lanes have the same orientation
3. lanes have different orientations (it is not relevant if all lanes are filled or not)

The cost of the evaluation increases from top to bottom. Option 2 and 3 could be used also in the case that the cells related to some lanes are residing on other processes on the same shared-memory domain if the index access would be replaced by raw pointers. This does this restructuring.

The pointers to the shared values of the other processes still has to be propagated to the given function.

Only the last commit is relevant.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -2994,14 +2994,45 @@ namespace internal
                     .component_dof_indices_offset[active_fe_index]
                                                  [first_selected_component];
 
+                const unsigned int n_filled_lanes =
+                  dof_info.n_vectorization_lanes_filled[dof_access_index][cell];
+
+                const bool vectorization_possible =
+                  (n_face_orientations == 1) &&
+                  (n_filled_lanes == VectorizedArrayType::size());
+
+                std::array<Number2_ *, VectorizedArrayType::size()>
+                  vector_ptrs = {};
+
+                if (vectorization_possible == false)
+                  {
+                    if (n_face_orientations == 1)
+                      {
+                        for (unsigned int v = 0; v < n_filled_lanes; ++v)
+                          vector_ptrs[v] = vector_ptr + indices[v];
+                      }
+                    else if (n_face_orientations == VectorizedArrayType::size())
+                      {
+                        for (unsigned int v = 0;
+                             v < VectorizedArrayType::size();
+                             ++v)
+                          if (cells[v] != numbers::invalid_unsigned_int)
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2993,14 +2994,45 @@ namespace internal
                     .component_dof_indices_offset[active_fe_index]
                                                  [first_selected_component];
 
+                const unsigned int n_filled_lanes =
+                  dof_info.n_vectorization_lanes_filled[dof_access_index][cell];
+
+                const bool vectorization_possible =
+                  (n_face_orientations == 1) &&
+                  (n_filled_lanes == VectorizedArrayType::size());
```
peterrum, line 3002: This check has to be generalized with an additional `&&` checking if all cells are locally owned.

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -2993,14 +2994,45 @@ namespace internal
...
+                            vector_ptrs[v] =
+                              vector_ptr +
+                              dof_info.dof_indices_contiguous[dof_access_index]
+                                                             [cells[v]];
+                      }
+                    else
+                      {
+                        Assert(false, ExcNotImplemented());
+                      }
+                  }
```
peterrum, line 3029: Here I translate the indices to raw pointers. In the SM case, pairs of process id and index will be translated.

peterrum:
BTW: are there any objections if we disable clang-format in this function. The logic in this function is very complicated as it is; the current formatting makes it unnecessarily significantly harder to understand.

kronbichler (approved):

kronbichler:
> BTW: are there any objections if we disable clang-format in this function. The logic in this function is very complicated as it is; the current formatting makes it unnecessarily significantly harder to understand.

I would be fine with disabling it, as long as the overall look is reasonable. There will be more restructuring in this function.

Let me know once you are happy with the result so we can merge this and go on with the PRs building on top of this.

peterrum:
Let's merge. And tackle the clang-format issue later.

Issue #9544, opened 2020-02-18
Teach FiniteElement how to fill MF::ShapeInfo


kronbichler:
The information for the tensor product information of certain `FiniteElement` classes is extracted in the struct `internal::MatrixFreeFunctions::ShapeInfo` from a given finite element object and a 1D quadrature formula. We do this by some dynamic cast magic in https://github.com/dealii/dealii/blob/1f3f4df3cd5e84a140e453b227608d5e83079789/include/deal.II/matrix_free/shape_info.templates.h#L103-L128 but this is not a real solution and does not scale if we want to add more elements (and I want to add some tensor product evaluators for RT and Nedelec with appropriate basis functions).

peterrum:
This is a good idea!

I guess then we can also eliminate this mess: 

https://github.com/dealii/dealii/blob/1f3f4df3cd5e84a140e453b227608d5e83079789/source/multigrid/mg_transfer_internal.cc#L713-L722

and maybe part of this:

https://github.com/dealii/dealii/blob/1f3f4df3cd5e84a140e453b227608d5e83079789/source/multigrid/mg_transfer_internal.cc#L582-L640

peterrum:
> Move the include/deal.II/matrix_free/shape_info.{h,templates.h} files to the fe/ subfolder, extract it from the internal namespace and give it a proper name. I was thinking along the lines FiniteElementTensorProductInformation. I would put a template for the number type on this class but not the dimension as a tensor product element is in principle not restricted to a particular dimension.

I just wanted to suggest the same!

kronbichler:
I hope that we can do; however, this means we need to add another field to the `ShapeInfo` class for the parent/child relation we use in GMG (which is useful information and cheap to obtain anyway).

peterrum:
@kronbichler  Let me know if I can do something or try something out. I have simply copied those two code snippets to my new transfer operators, but I am not happy with that...

I will have a look at my code maybe I find some additional requirements!

kronbichler:
The evaluator side also relates to #7039 that should be generalized/restructured in one go.

peterrum:
This PR will involve some time we don't have at the time. Adjusting the milestone.

kronbichler:
In #14385 we noticed that the cases where we need special treatments, to be addressed by this PR, just keep increasing. I think we should aim for this effort before the next release.

peterrum:
This is a place where this would be useful:

[...]

PR #11086, opened 2020-10-22
Improve MatrixFree::is_supported(FiniteElement)

kronbichler:
Since we merged the DG support for `MatrixFree` in 2018, we actually support `FiniteElement` objects with multiple base elements.

This also moves the function to `ShapeInfo` because the selection is primarily made in that function, so the code sits where it belongs. Related to #9544. Found while working on #11046.


masterleinad (approved):
Looks OK to me.

peterrum:
Looks good except the failing tests.

peterrum (approved):

PR #11088, opened 2020-10-22
Use non-reference type for copied RBL element

### LINKED ISSUE #11087: Build warnings with apple-clang 12.0
```
include/deal.II/base/patterns.h:1796:26: warning: loop variable 'entry' is always a copy because the range of type 'const std::__1::vector<bool, std::__1::allocator<bool> >' does not return a reference [-Wrange-loop-analysis]
        for (const auto &entry : t)
                         ^
<PATH>/include/deal.II/base/patterns.h:2100:44: note: in instantiation of member function 'dealii::Patterns::Tools::Convert<std::__1::vector<bool, std::__1::allocator<bool> >, void>::to_string' requested here
        return Convert<std::vector<bool>>::to_string(mask, pattern);
                                           ^
<PATH>/include/deal.II/base/patterns.h:1796:14: note: use non-reference type 'std::__1::__bit_const_reference<std::__1::vector<bool, std::__1::allocator<bool> > >'
        for (const auto &entry : t)
```
... Jeez, I think that I'm going to leave the rest for now :-D

jppelteret:
Fixes #11087

### DIFF
--- a/include/deal.II/base/patterns.h
+++ b/include/deal.II/base/patterns.h
@@ -1792,9 +1792,10 @@ namespace Patterns
         auto                     base_p = p->get_base_pattern().clone();
         std::vector<std::string> vec(t.size());
 
-        unsigned int i = 0;
-        for (const auto &entry : t)
-          vec[i++] = Convert<typename T::value_type>::to_string(entry, base_p);
+        std::transform(
+          t.cbegin(), t.cend(), vec.begin(), [&base_p](const auto &entry) {
+            return Convert<typename T::value_type>::to_string(entry, base_p);
+          });
 
         std::string s;
         if (vec.size() > 0)


masterleinad (commented):
The warning only makes sense for `T=std::vector<bool>`.
We should be able to use `std::transform` here instead and delegate these problems to the STL.

jppelteret:
Thanks for the tip! I was going to just replace the RBL by an old-fashioned manual iteration, but your suggestion is much better.

masterleinad (approved):
If this helps with the warning, I am fine with it.

jppelteret:
> If this helps with the warning, I am fine with it.

It did, thanks. Unfortunately there are a whole bunch more, but I think I'll tackle them another time.

kronbichler (approved):

PR #11090, opened 2020-10-23
Extend GridOut::write_svg() for simplices

eliasstudiert:
Hi everyone, 
I've adapted the write_svg() function in order to work with simplices as well.

### DIFF
--- a/source/grid/grid_out.cc
+++ b/source/grid/grid_out.cc
@@ -1584,7 +1584,8 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
   // (, and level subdomain id).
   for (const auto &cell : tria.cell_iterators())
     {
-      for (unsigned int vertex_index = 0; vertex_index < 4; vertex_index++)
+      for (unsigned int vertex_index = 0; vertex_index < cell->n_vertices();
+           ++vertex_index)
         {
           if (cell->vertex(vertex_index)[0] < x_min)
             x_min = cell->vertex(vertex_index)[0];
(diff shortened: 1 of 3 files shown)


source/grid/grid_out.cc
```diff
@@ -1834,8 +1835,11 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
       if (y_min_perspective > projection_decomposition[1])
         y_min_perspective = projection_decomposition[1];
 
-      point[0] = cell->vertex(3)[0];
-      point[1] = cell->vertex(3)[1];
+      if (cell->n_vertices() == 4)
```
peterrum, line 1838: Could you add a comment here why the check is necessary?

peterrum (approved):
Looks good to me.

source/grid/grid_out.cc
drwells: It's unfortunate that the same code is essentially copied and pasted four times. I would like to clean that up but we shouldn't burden this PR with that refactoring.

peterrum: > Could we avoid writing the extra point?

That's true. @eliasstudiert could you fix this moving the bottom curly bracket down.

eliasstudiert: I've added the comments and fixed writing of the extra point.

source/grid/grid_out.cc
```diff
@@ -2179,32 +2179,32 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
...
+                  << ' '
+                  << static_cast<unsigned int>(
+                       .5 + height - (height / 100.) * margin_in_percent -
+                       ((projection_decomposition[1] - y_min_perspective) /
+                        y_dimension_perspective) *
+                         (height - (height / 100.) * 2. * margin_in_percent));
 
-          out << " L ";
+              out << " L ";
+            }
```
eliasstudiert, line 2207: I've moved this inside the if query, since that's not required for simplices, right?

source/grid/grid_out.cc
```diff
@@ -1584,7 +1584,8 @@ GridOut::write_svg(const Triangulation<2, 2> &tria, std::ostream &out) const
   // (, and level subdomain id).
   for (const auto &cell : tria.cell_iterators())
     {
-      for (unsigned int vertex_index = 0; vertex_index < 4; vertex_index++)
+      for (unsigned int vertex_index = 0; vertex_index < cell->n_vertices();
+           vertex_index++)
```
drwells, line 1588:
```suggestion
           ++vertex_index)
```
clang-tidy will catch this so we should just fix it now.

drwells (changes requested):
We will need to fix one minor thing to make the CI happy but otherwise this looks good. I verified that we get a nice SVG image when we run the test.

drwells (approved):
One more thing - do you know how to squash commits?

source/grid/grid_out.cc
drwells: The output looks fine without it so I think not.

eliasstudiert:
> One more thing - do you know how to squash commits?

I've tried git rebase --interactive HEAD~5 to squash the last 5 commits into one commit. If I do then git log, I can see that one commit (containing the 5 single ones). My problem is, I don't know how to push it correctly, so that we can see it here.
Any idea?

masterleinad:
Just push to the branch the same way you did before but add `-f` to force the push and overwrite the history on the remote branch.

eliasstudiert:
> Just push to the branch the same way you did before but add `-f` to force the push and overwrite the history on the remote branch.

Thanks, Daniel!

drwells (approved):

drwells:
The windows and OSX failures are unrelated so lets merge once the other two checks conclude successfully.

peterrum (approved):

kronbichler (approved):

kronbichler:
Test failures unrelated and covered elsewhere, merging.

PR #11089, opened 2020-10-23
MatrixFree: introduce get_vector_data()

peterrum:
and move around `internal::get_beginning()` and use `std::array` in more places.

PR #11091, opened 2020-10-23
MatrixFree: pass SM vectors to functions

peterrum:
Follow up to #11089.


peterrum:
I think this PR is ready for review. The code is actually not used yet (there 1-2 PRs still missing) and I guess there will be some (small) changes once actually used.

kronbichler (approved):

kronbichler (approved):

PR #11075, opened 2020-10-21
Wrap partitioners within MatrixFree

peterrum:
Depends on #11074.

PR #11094, opened 2020-10-23
Fix get_shared_vector_data

peterrum:
The combined merge of #11075 and #11091 revealed a bug that prevents compiling master.

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -4859,7 +4859,7 @@ namespace internal
   const std::vector<ArrayView<const typename VectorType::value_type>> *
   get_shared_vector_data(VectorType &vec /*other parameters?*/)
   {
-    return vec.shared_vector_data();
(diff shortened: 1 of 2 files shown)


drwells (approved):

drwells:
I'm surprised that both of those patches passed the CI but putting them together didn't. Did we do something incorrectly? I don't understand how this circumstance is even possible.

peterrum:
```
    return &vec.shared_vector_data();
```
The first one: I did not expect that this function will be called (since the L:p:V does not have the SM capabilities yet).

```
  for (auto &p : matrix_free.get_dof_info().vector_exchanger_face_variants)
```
I am not sure about the second one. I normally run these test a couple of times an hour. Also, I guess this test is not run by CI!?

kronbichler (approved):

bangerth:
All but two of the testers succeeded, the situation is already dire, so I'm taking the liberty of merging even though two tests are still pending.

peterrum:
@bangerth  Thanks!

PR #11095, opened 2020-10-24
Document reference cells in simplex module page

### LINKED ISSUE #11084: Document all our new shapes.
The node numberings for the new elements (in particular prisms and wedges) are currently only available in some of @peterrum's slides. We should move this into the official documentation.

peterrum:
Closes #11084.

The output would look like this:

Screenshot from 2020-10-24 10-12-12

The 3D pictures need to be so large else you cannot recognize the numbers in the small circles.


doc/doxygen/headers/simplex.h
```diff
@@ -17,5 +17,84 @@
...
+ *
+ * @section simplex_reference_cells Reference cells
+ *
+ * In 2D, we provide triangles and quadrilaterals with the following possible
+ * orientations in 3D:
+ *
+ * <div class="twocolumn" style="width: 100%">
+ *   <div class="parent">
+ *     <div class="img" align="center">
+ *       <img src="https://user-images.githubusercontent.com/15707929/97070787-63b8ab00-15db-11eb-9e4b-b254b110bf18.png"
```
drwells, line 30: We need to host our own images for the offline documentation to work correctly.

drwells (changes requested):
We might want to replace these with ASCII later but we need to publish the numberings somewhere now so this looks good to me.

drwells:
Did we ever agree on what kind of images should be put in `doc/doxygen/images` and which should go on the web site? I think it's fine for small things like this to be in the library (we definitely have test output files larger than 15 kb).

drwells:
The largest test output files are something like 5 MB each so I don't think its a big deal if we put images like this in the repository itself. Can we just do that? If anyone objects let's talk about it.

peterrum:
@drwells I have added the pictures!

drwells (approved):

PR #11096, opened 2020-10-24
step-44: Improve how constraints are constructed

### LINKED ISSUE #11035: Improve step-44's make_constraints()
Step-44's way of making the constraints is overly complicated. A simpler way to structure the constraints construction would be something like this:
```
template <int dim>
void Solid<dim>::make_constraints(const int &it_nr)
{
  std::cout << " CST " << std::flush;
  if (it_nr > 1)
    return;

  if (it_nr == 0)
  {
    // Construct both non-homogeneous and homogeneous constraints
    // This means that a user would only change this code block to add inhomogeneous constraints
    // to the problem.
  }
  else
  {
     // Loop over entries in constraints matrix
     //   - If not hanging node constraint, set inhomogenity to zero.
    //      (or simply merge back in the hanging node constraints, making sure that they dominate)
  }
}
```

jppelteret:
Fixes #11035

### DIFF
--- a/examples/step-44/step-44.cc
+++ b/examples/step-44/step-44.cc
@@ -2391,194 +2391,183 @@ namespace Step44
 
   // @sect4{Solid::make_constraints}
   // The constraints for this problem are simple to describe.
-  // However, since we are dealing with an iterative Newton method,
-  // it should be noted that any displacement constraints should only
-  // be specified at the zeroth iteration and subsequently no
-  // additional contributions are to be made since the constraints
-  // are already exactly satisfied.
+  // In this particular example, the boundary values will be calculated for
+  // the two first iterations of Newton's algorithm. In general, one would
+  // build non-homogeneous constraints in the zeroth iteration (that is, when
+  // `apply_dirichlet_bc == true` in the code block that follows) and build
+  // only the corresponding homogeneous constraints in the following step. While
+  // the current example has only homogeneous constraints, previous experiences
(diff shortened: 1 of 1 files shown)


examples/step-44/step-44.cc
```diff
@@ -2391,194 +2391,183 @@ namespace Step44
...
-  // additional contributions are to be made since the constraints
-  // are already exactly satisfied.
+  // In this particular example, the boundary values will be calculated for
+  // the two first iterations of Newton's algorithm. In general, one would
+  // build non-homogeneous constraints in the zeroth iteration (that is, when
+  // `apply_dirichlet_bc == true` in the code block that follows) and build
+  // only the corresponding homogeneous constraints in the following step. While
+  // the current example has only homogeneous constraints, previous experiences
+  // have shown that a common error is forgetting to add the extra condition
+  // when refactoring the code to specific uses. This could lead errors that are
```
Rombur, line 2401: I think there is a typo?
```suggestion
  // when refactoring the code to specific uses. This could lead to errors that are
```

Rombur (approved):
This looks much cleaner. I think there is a typo. Other than that it's ready to be merged.

examples/step-44/step-44.cc
jppelteret: Thanks for the review, and for spotting this mistake! I've corrected it now.

PR #11074, opened 2020-10-21
Introduce MPI-3 shared-memory capabilities in L:d:V

peterrum:
Depends on #11071.

In contrast to #10872, this PR adds the MPI-3 capabilities to `L:d:V`. This feature (using `MPI_Win_allocate_shared` instead of `posix_memalign`) is used only if an additional communicator is provided that is not `MPI_COMM_SELF`. A new function is `shared_vector_data()`: it gives read-only access to the values of the processes within the second communicator.

PR #11098, opened 2020-10-24
Introduce VectorDataExchange::Full

peterrum:
... and introduce 2nd communicator in MatrixFree.

As long as #11074 is not merged and no SM can be exploited, the new `VectorDataExchange` is not really useful but nevertheless gives the right result.

### DIFF
--- a/include/deal.II/matrix_free/dof_info.templates.h
+++ b/include/deal.II/matrix_free/dof_info.templates.h
@@ -382,7 +382,9 @@ namespace internal
 
 
     void
-    DoFInfo::assign_ghosts(const std::vector<unsigned int> &boundary_cells)
+    DoFInfo::assign_ghosts(const std::vector<unsigned int> &boundary_cells,
+                           const MPI_Comm &                 communicator_sm,
+                           const bool use_vector_data_exchanger_full)
     {
       Assert(boundary_cells.size() < row_starts.size(), ExcInternalError());
 
@@ -497,9 +499,14 @@ namespace internal
         const_cast<Utilities::MPI::Partitioner *>(vector_partitioner.get());
       vec_part->set_ghost_indices(ghost_indices);
 
-      vector_exchanger = std::make_shared<
-        internal::MatrixFreeFunctions::VectorDataExchange::PartitionerWrapper>(
-        vector_partitioner);
+      if (use_vector_data_exchanger_full == false)
+        vector_exchanger =
+          std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
+                             PartitionerWrapper>(vector_partitioner);
+      else
+        vector_exchanger = std::make_shared<
+          internal::MatrixFreeFunctions::VectorDataExchange::Full>(
+          vector_partitioner, communicator_sm);
     }
 
 
@@ -1173,7 +1180,9 @@ namespace internal
       const unsigned int                        n_lanes,
       const std::vector<FaceToCellTopology<1>> &inner_faces,
       const std::vector<FaceToCellTopology<1>> &ghosted_faces,
-      const bool                                fill_cell_centric)
+      const bool                                fill_cell_centric,
+      const MPI_Comm &                          communicator_sm,
+      const bool                                use_vector_data_exchanger_full)
     {
       const Utilities::MPI::Partitioner &part = *vector_partitioner;
 
@@ -1226,9 +1235,14 @@ namespace internal
               ->set_ghost_indices(compressed_set, part.ghost_indices());
           }
 
-        vector_exchanger_face_variants[0] =
-          std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
-                             PartitionerWrapper>(temp_0);
+        if (use_vector_data_exchanger_full == false)
+          vector_exchanger_face_variants[0] =
+            std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
+                               PartitionerWrapper>(temp_0);
+        else
+          vector_exchanger_face_variants[0] = std::make_shared<
+            internal::MatrixFreeFunctions::VectorDataExchange::Full>(
+            temp_0, communicator_sm);
       }
 
       // construct a numbering of faces
@@ -1467,18 +1481,36 @@ namespace internal
             part.locally_owned_range(), part.get_mpi_communicator());
         }
 
-      vector_exchanger_face_variants[1] = std::make_shared<
-        internal::MatrixFreeFunctions::VectorDataExchange::PartitionerWrapper>(
-        temp_1);
-      vector_exchanger_face_variants[2] = std::make_shared<
-        internal::MatrixFreeFunctions::VectorDataExchange::PartitionerWrapper>(
-        temp_2);
-      vector_exchanger_face_variants[3] = std::make_shared<
-        internal::MatrixFreeFunctions::VectorDataExchange::PartitionerWrapper>(
-        temp_3);
-      vector_exchanger_face_variants[4] = std::make_shared<
-        internal::MatrixFreeFunctions::VectorDataExchange::PartitionerWrapper>(
-        temp_4);
+      if (use_vector_data_exchanger_full == false)
+        {
+          vector_exchanger_face_variants[1] =
+            std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
+                               PartitionerWrapper>(temp_1);
+          vector_exchanger_face_variants[2] =
+            std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
+                               PartitionerWrapper>(temp_2);
+          vector_exchanger_face_variants[3] =
+            std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
+                               PartitionerWrapper>(temp_3);
+          vector_exchanger_face_variants[4] =
+            std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
+                               PartitionerWrapper>(temp_4);
+        }
+      else
+        {
+          vector_exchanger_face_variants[1] = std::make_shared<
+            internal::MatrixFreeFunctions::VectorDataExchange::Full>(
+            temp_1, communicator_sm);
+          vector_exchanger_face_variants[2] = std::make_shared<
+            internal::MatrixFreeFunctions::VectorDataExchange::Full>(
+            temp_2, communicator_sm);
+          vector_exchanger_face_variants[3] = std::make_shared<
+            internal::MatrixFreeFunctions::VectorDataExchange::Full>(
+            temp_3, communicator_sm);
+          vector_exchanger_face_variants[4] = std::make_shared<
+            internal::MatrixFreeFunctions::VectorDataExchange::Full>(
+            temp_4, communicator_sm);
+        }
     }
 
 
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -244,6 +244,7 @@ public:
       , cell_vectorization_categories_strict(
           cell_vectorization_categories_strict)
       , communicator_sm(MPI_COMM_SELF)
+      , use_vector_data_exchanger_full(false)
     {}
 
     /**
@@ -270,6 +271,7 @@ public:
       , cell_vectorization_categories_strict(
           other.cell_vectorization_categories_strict)
       , communicator_sm(other.communicator_sm)
+      , use_vector_data_exchanger_full(other.use_vector_data_exchanger_full)
     {}
 
     // remove with level_mg_handler
@@ -299,7 +301,8 @@ public:
       cell_vectorization_category   = other.cell_vectorization_category;
       cell_vectorization_categories_strict =
         other.cell_vectorization_categories_strict;
-      communicator_sm = other.communicator_sm;
+      communicator_sm                = other.communicator_sm;
+      use_vector_data_exchanger_full = other.use_vector_data_exchanger_full;
 
       return *this;
     }
@@ -537,6 +540,13 @@ public:
      * Shared-memory MPI communicator. Default: MPI_COMM_SELF.
      */
     MPI_Comm communicator_sm;
+
+    /**
+     * Experimental: flag to switch between
+     * internal::MatrixFreeFunctions::VectorDataExchange::PartitionerWrapper and
+     * internal::MatrixFreeFunctions::VectorDataExchange::Full.
+     */
+    bool use_vector_data_exchanger_full;
   };
 
   /**
@@ -3295,7 +3305,8 @@ namespace internal
 
           const auto &part = get_partitioner(mf_component);
 
-          if (part.n_ghost_indices() == 0 && part.n_import_indices() == 0)
+          if (part.n_ghost_indices() == 0 && part.n_import_indices() == 0 &&
+              part.n_import_sm_procs() == 0)
             return;
 
           tmp_data[component_in_block_vector] =
@@ -3305,7 +3316,7 @@ namespace internal
           AssertDimension(requests.size(), tmp_data.size());
 
           part.export_to_ghosted_array_start(
-            component_in_block_vector + channel_shift,
+            component_in_block_vector * 2 + channel_shift,
             ArrayView<const Number>(vec.begin(), part.local_size()),
             vec.shared_vector_data(),
             ArrayView<Number>(const_cast<Number *>(vec.begin()) +
@@ -3504,7 +3515,7 @@ namespace internal
 
           part.import_from_ghosted_array_start(
             dealii::VectorOperation::add,
-            component_in_block_vector + channel_shift,
+            component_in_block_vector * 2 + channel_shift,
             ArrayView<Number>(vec.begin(), part.local_size()),
             vec.shared_vector_data(),
             ArrayView<Number>(vec.begin() + part.local_size(),
@@ -3601,6 +3612,9 @@ namespace internal
                 tmp_data[component_in_block_vector]);
               tmp_data[component_in_block_vector] = nullptr;
             }
+
+          if (Utilities::MPI::job_supports_mpi())
+            MPI_Barrier(matrix_free.get_task_info().communicator_sm);
 #  endif
         }
     }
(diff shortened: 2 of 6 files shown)


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -2157,7 +2166,8 @@ MatrixFree<dim, Number, VectorizedArrayType>::initialize_dof_vector(
   const unsigned int comp) const
 {
   AssertIndexRange(comp, n_components());
-  vec.reinit(dof_info[comp].vector_partitioner->size());
+  vec.reinit(
+    dof_info[comp].vector_partitioner->size() /*, task_info.communicator_sm*/);
```
peterrum, line 2170: Here the `reinit` function of `L:p:v` with additional parameter will be used.

include/deal.II/matrix_free/dof_info.templates.h
```diff
@@ -497,9 +498,14 @@ namespace internal
         const_cast<Utilities::MPI::Partitioner *>(vector_partitioner.get());
       vec_part->set_ghost_indices(ghost_indices);
 
-      vector_exchanger = std::make_shared<
-        internal::MatrixFreeFunctions::VectorDataExchange::PartitionerWrapper>(
-        vector_partitioner);
+      if (false)
+        vector_exchanger =
+          std::make_shared<internal::MatrixFreeFunctions::VectorDataExchange::
+                             PartitionerWrapper>(vector_partitioner);
```
kronbichler, line 504: Is this intentional that you use this exchange scheme already now? I assumed from the PR description that it is not of much use (but correct) and thus enabled later? Not that I'm opposed to it, but this looks more like a performance setup to be tested rather than the default before the full functionality is available.

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -243,7 +243,9 @@ class MatrixFree : public Subscriptor
       , hold_all_faces_to_owned_cells(hold_all_faces_to_owned_cells)
       , cell_vectorization_categories_strict(
           cell_vectorization_categories_strict)
-    {}
+    {
+      this->communicator_sm = MPI_COMM_SELF;
```
kronbichler, line 247: Why don't you include `communicator_sm(MPI_COMM_SELF)` in the list above for symmetry with the other parameters?

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -337,12 +337,46 @@ MatrixFree<dim, Number, VectorizedArrayType>::internal_reinit(
             Utilities::MPI::this_mpi_process(task_info.communicator);
           task_info.n_procs =
             Utilities::MPI::n_mpi_processes(task_info.communicator);
+
+          Assert(additional_data.communicator_sm == MPI_COMM_SELF,
+                 ExcNotImplemented());
+
+#ifndef DEAL_II_WITH_MPI
+          Assert(false, ExcInternalError());
+#else
```
kronbichler, line 346: Do we really need this assert, given that 15 lines above we do `job_supports_mpi()`?
```suggestion
#ifdef DEAL_II_WITH_MPI
```

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -337,12 +337,46 @@ MatrixFree<dim, Number, VectorizedArrayType>::internal_reinit(
...
+              task_info.communicator_sm = additional_data.communicator_sm;
+            }
+#  if DEAL_II_MPI_VERSION_GTE(3, 0)
+          else if (true)
+            {
+              MPI_Comm comm_sm;
+              MPI_Comm_split(task_info.communicator,
+                             task_info.my_pid == 2,
+                             task_info.my_pid,
+                             &comm_sm);
```
kronbichler, line 358: Now we never go into this part, but I am confused by this statement and `my_pid == 2` in particular. Can we maybe postpone this part until we have a reasonable criterion here? Or is it just me not understanding the `Comm_split` context here?

include/deal.II/matrix_free/vector_data_exchange.h
```diff
@@ -247,6 +247,307 @@ namespace internal
         const std::shared_ptr<const Utilities::MPI::Partitioner> partitioner;
       };
 
+
+
+      /**
+       * Similar to the above but that the provided partitioner is pre-processed
+       * and indices of degrees of freedom that are shared are identified.
```
kronbichler, line 254: I do not really understand this comment. Do you mean something along the lines
```suggestion
       * Similar to the above but using the internal data structures in the
       * partitioner in order to identify indices of degrees of freedom that are
       * in the same shared memory region.
```
And by `shared` you mean shared memory or something else? Please expand or adapt as appropriate.

include/deal.II/matrix_free/vector_data_exchange.h
```diff
@@ -247,6 +247,307 @@ namespace internal
...
+        std::vector<unsigned int> recv_remote_ranks;
+
+        /**
+         * Pointer into the receive buffer.
+         */
+        std::vector<std::pair<types::global_dof_index, types::global_dof_index>>
+          recv_remote_ptr;
+
+        /**
+         * TODO
```
kronbichler, line 436: Could you please fill in this variable, along with the other todos further down in this file?

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,965 @@ namespace internal
             ghost_array[j] = 0.;
       }
 
+
+
+      namespace internal
+      {
+        void
+        compress(std::vector<unsigned int> &recv_sm_ptr,
```
kronbichler, line 326: Can you please adapt the name of this function? I think a better name could be `compress_to_contiguous_ranges`.

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,965 @@ namespace internal
...
+#ifndef DEAL_II_WITH_MPI
+          Assert(false, ExcNeedsMPI());
+          (void)comm;
+          (void)comm_shared;
+#else
+          // extract information from comm
+          int rank_;
+          MPI_Comm_rank(comm, &rank_);
+
+          const unsigned int rank = rank_;
```
kronbichler, line 378: I do not really see why you switch between unsigned int and signed int here and later do it again in the `Allgather` command below in a strange way. Shouldn't the function be fine with taking an `MPI_INT` as the input and give us a `MPI_UNSIGNED` as output?

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,965 @@ namespace internal
...
+
+          // gather ranks
+          ranks_shared.resize(size_shared);
+          MPI_Allgather(&rank,
+                        1,
+                        MPI_UNSIGNED,
+                        ranks_shared.data(),
+                        1,
+                        MPI_INT,
+                        comm_shared);
```
kronbichler, line 392: To expand on the previous comment, for the current arguments the correct way would be
```suggestion
          MPI_Allgather(&rank,
                        1,
                        MPI_UNSIGNED,
                        ranks_shared.data(),
                        1,
                        MPI_UNSIGNED,
                        comm_shared);
```
But I would change the first to `MPI_INT`.

include/deal.II/matrix_free/vector_data_exchange.h
```diff
@@ -247,6 +247,307 @@ namespace internal
...
+         */
+        std::vector<unsigned int> send_sm_ptr_;
+
+        /**
+         * TODO
+         */
+        std::vector<unsigned int> send_sm_len_;
```
kronbichler, line 548: I have not arrived yet at the implementation, but I find the names here that just differ by the final underscore from other variables too close in the name. Can you find more appropriate names (here or for the variables above).

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,965 @@ namespace internal
...
+        this->comm    = communicator;
+        this->comm_sm = communicator_sm;
+
+#ifndef DEAL_II_WITH_MPI
+        Assert(false, ExcNeedsMPI());
+
+        (void)is_locally_owned;
+        (void)is_locally_ghost;
+        (void)ghost_indices_within_larger_ghost_set;
+#else
```
kronbichler, line 426: Can't you avoid this part by moving the declarations of the aliases into the `#ifdef DEAL_II_WITH_MPI`?

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,965 @@ namespace internal
...
+        std::vector<unsigned int> sm_ranks(
+          Utilities::MPI::n_mpi_processes(comm_sm));
+
+        const unsigned int rank = Utilities::MPI::this_mpi_process(comm);
+
+        MPI_Allgather(
+          &rank, 1, MPI_UNSIGNED, sm_ranks.data(), 1, MPI_UNSIGNED, comm_sm);
+
+        std::vector<unsigned int> owning_ranks_of_ghosts(
+          is_locally_ghost.n_elements());
```
kronbichler, line 443: Do you want to initialize this to `numbers::invalid_unsigned_int` to be explicit that this is filled in the consensus algorithm below?

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,965 @@ namespace internal
...
+
+      void
+      Full::export_to_ghosted_array_start(
+        const unsigned int                          communication_channel,
+        const ArrayView<const double> &             locally_owned_array,
+        const std::vector<ArrayView<const double>> &shared_arrays,
+        const ArrayView<double> &                   ghost_array,
+        const ArrayView<double> &                   temporary_storage,
+        std::vector<MPI_Request> &                  requests) const
+      {
```
kronbichler, line 670: I must say that I do not particularly like to have hardcoded functions for both `double` and `float` because it would make extensions to say `std::complex<double>` (if that ever makes sense) cumbersome and double the number of functions. I realize it is the "best" alternative behind a `virtual` function interface for a class that we do not want to be templated by `Number`, a requirement to be able to run `MF::loop` on vectors of different type than the `MatrixFree` type on its own. If in the end we could move all code through a single realization of `VectorDataExchange` then of course we could skip the virtual functions and go to the templated, aka `_impl` path, at the expense of moving everything into the header file. Anyway, if anyone has other suggestions it would be good to hear them.

include/deal.II/matrix_free/matrix_free.templates.h
peterrum: I think I wanted to write `task_info.my_pid / 2`. In the future the communicator should come from outside.

But anyhow, what I wanted to do is to run all tests with pure MPI, pure SM-MPI and hybrid-MPI. Without having to modify all tests this is the place where it can be achieved without too much work.

source/matrix_free/vector_data_exchange.cc
peterrum: > I must say that I do not particularly like to have hardcoded functions for both double and float

But I have no other solution...

include/deal.II/matrix_free/matrix_free.templates.h
kronbichler: I see, I guess we can keep it for now and simply delete it once everything else is in place.

include/deal.II/matrix_free/vector_data_exchange.h
```diff
@@ -247,6 +247,269 @@ namespace internal
...
+                  std::vector<std::pair<unsigned int, unsigned int>>>
+          sm_import_data;
+
+        /**
+         * Indices where to copy data to during
+         * import_from_ghosted_array_finish().
+         */
+        std::pair<std::vector<unsigned int>,
+                  std::vector<std::pair<unsigned int, unsigned int>>>
+          sm_import_data_this;
```
peterrum, line 510: These are the new variable names. I still need to refactor how these are set up.

kronbichler: I like these names.

peterrum: Puh :) Let me clean up the constructor tomorrow and then this PR is ready from my side!

include/deal.II/matrix_free/dof_info.templates.h
peterrum: Would it be possible for the time being to introduce a new parameter in `AdditionalData`. Hopefully, we can remove it sometime!

peterrum:
@kronbichler May I ask you to have a look at `vector_data_exchange.h` and `vector_data_exchange.cc`. I have refactored the setup routines. Once we are happy with the code, I will remove the unrelated changes from the other files.

include/deal.II/matrix_free/dof_info.templates.h
kronbichler: I guess you intend to look into both schemes, so I would be fine with having that parameter for the time being; if you feel it can also be done with the present `if` statements in the code I would also be fine with the current version.

include/deal.II/matrix_free/vector_data_exchange.h
```diff
@@ -247,6 +247,273 @@ namespace internal
...
+                  std::vector<std::pair<unsigned int, unsigned int>>>
+          ghost_indices_subset_data;
+
+        /**
+         * An array that contains information which processors my ghost indices
+         * belong to, at which offset and how many those indices are
+         */
+        std::vector<
+          std::pair<unsigned int, std::pair<unsigned int, unsigned int>>>
+          ghost_targets_data;
```
kronbichler, line 448: I'm thinking about the data structure here. Would be lose a lot if we made this an `std::array<unsigned int, 3>` and state that the first index is the processor rank, the second the offset within the ghost array, and the third the number of indices? To me, accessing `ghost_targets_data[1]` would at least not be worse than `ghost_targets_data.second.first`.

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,1057 @@ namespace internal
...
+
+        std::vector<
+          std::pair<unsigned int, std::pair<unsigned int, unsigned int>>>
+          import_targets_data;
+
+        std::vector<unsigned int> sm_ghost_ranks;
+
+        std::vector<unsigned int> sm_import_ranks;
+
+        // temporal uncompressed data structures for ghost_indices_subset_data
```
kronbichler, line 408:
```suggestion
        // temporary uncompressed data structures for ghost_indices_subset_data
```

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,1057 @@ namespace internal
...
+                        ghost_targets_data.size() + import_targets_data.size());
+
+        int dummy;
+        // receive a signal that relevant sm neighbors are ready
+        for (unsigned int i = 0; i < sm_ghost_ranks.size(); i++)
+          MPI_Irecv(&dummy,
+                    0,
+                    MPI_INT,
+                    sm_ghost_ranks[i],
+                    communication_channel + 2,
```
kronbichler, line 862: Can this number `communication_channel + 2` and `+ 3` further down mix up with any other non-blocking MPI communication elsewhere? In particular, I wonder how the `communication_channel` is incremented in the consumer code of `MatrixFree` and if different components of a block vector could clash. I believe they can't because the numbers are the same for all blocks of a vector.

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,1057 @@ namespace internal
...
+                        MPI_STATUS_IGNORE);
+
+            const auto s = split(i);
+            i            = s.second;
+
+            if (s.first == 0)
+              {
+                const Number *__restrict__ data_others_ptr =
+                  data_others[sm_ghost_ranks[i]].data();
+                Number *__restrict__ data_this_ptr = ghost_array.data();
```
kronbichler, line 972: Can you use 
```suggestion
                const Number *DEAL_II_RESTRICT data_others_ptr =
                  data_others[sm_ghost_ranks[i]].data();
                Number *DEAL_II_RESTRICT data_this_ptr = ghost_array.data();
```
to make sure this works with all compilers?

source/matrix_free/vector_data_exchange.cc
```diff
@@ -316,6 +318,1057 @@ namespace internal
...
+                        MPI_STATUS_IGNORE);
+
+            const auto &s = split(i);
+            i             = s.second;
+
+            if (s.first == 0)
+              {
+                Number *__restrict__ data_others_ptr =
+                  const_cast<Number *>(data_others[sm_import_ranks[i]].data());
+                Number *__restrict__ data_this_ptr = data_this.data();
```
kronbichler, line 1234: Same here

kronbichler (commented):
Looks good to me apart from a few minor comments given below.

source/matrix_free/vector_data_exchange.cc
peterrum: It looks the incrementation is two: 
https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/matrix_free.h#L3300

I need at least two tags. Should we increment then by 2?

include/deal.II/matrix_free/vector_data_exchange.h
peterrum: I have changing this one and import_targets_data.

include/deal.II/matrix_free/dof_info.templates.h
peterrum: > I guess you intend to look into both schemes, so I would be fine with having that parameter for the time being; if you feel it can also be done with the present if statements in the code I would also be fine with the current version.

My final goal is to through out the usage of `Partitioner` inside of `MatrixFree`. But for the time being (as long as we don't know if the new class performs at least as good as the `Partitioner`), I would like to have an easy way to switch between implementations.

include/deal.II/matrix_free/matrix_free.templates.h
peterrum: I will remove this in the final version.

source/matrix_free/vector_data_exchange.cc
kronbichler: Since we are using `+0`, `+1`, `+2`, `+3` in this PR, I suggest to increase the counter there with `4 * component_in_block_vector`.

kronbichler:
Apart from the previous comment, this looks good. Please remove the unnecessary parts and I can approve this.

include/deal.II/matrix_free/dof_info.templates.h
kronbichler: Sounds good, let's go with this temporary solution.

peterrum:
> Apart from the previous comment, this looks good. Please remove the unnecessary parts and I can approve this.

I have create a new branch as backup with the *unnecessary parts*: https://github.com/peterrum/dealii/tree/vde_full_reference

peterrum:

kronbichler (approved):
Nice, thanks for cleaning things up.

PR #11099, opened 2020-10-24
Step 44: Unify assembly loop and introduce an optimisation

jppelteret:
For quite some time I've felt that step-44's way of assembly, namely having one method to assemble the system matrix and another the RHS, was not such a good approach. It seemed the correct thing to do at the time (the code that step-44 was based off of had a very expensive tangent matrix assembly), but in retrospect it makes extending the code to MPI and other scenarios difficult. In the last few years, each time that I've used this tutorial as a base code I've ended up unifying these two functions. So it seems like its worth formally introducing this approach upstream so that no-one has to perform this task again. It also then better aligns with the other tutorials in this respect.

I've also added a nice little optimisation that @davydden pointed out to me some years ago. Its a small change but it nearly halves the assembly time for the default setup, and it looks like its even more impactful as the number of elements increases.
| Global refinements | Old implementation (tangent + RHS) | New implementation (unified assm. + opt.) |
|--------------------|--------------------------------------|----------------------|
|                             2 |                                         2.65s + 0.5s |                           1.25s |
|                             3 |                                         15s + 0.95s |                            5.3s |

For other coupled problems involving rank-[`>=2`] tensors, the same trick gives even more reward. I can't believe that I had never noticed it before, even though I've stared at this code for many hours.

On the correctness of the change, here is the output at the last time step for the current implementation
```
Timestep 10 @ 1.000e+00s
___________________________________________________________________________________________________________________________________________________________
                 SOLVER STEP                   |  LIN_IT   LIN_RES    RES_NORM     RES_U     RES_P      RES_J     NU_NORM      NU_U       NU_P       NU_J
___________________________________________________________________________________________________________________________________________________________
  0  ASM_R  ASM_K  CST  ------  SLV  PP  UQPH  |     456  1.905e-06  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00
  1  ASM_R  ASM_K  CST  ------  SLV  PP  UQPH  |     260  2.967e-04  1.544e-01  1.544e-01  1.154e+13  1.849e+06  6.013e-02  7.397e-02  6.013e-02  6.013e-02
  2  ASM_R  ASM_K  CST  ------  SLV  PP  UQPH  |     279  2.099e-06  2.903e-03  2.903e-03  6.974e+10  2.060e+03  2.732e-03  1.436e-03  2.732e-03  2.733e-03
  3  ASM_R  ASM_K  CST  ------  SLV  PP  UQPH  |     280  2.494e-09  2.086e-06  2.086e-06  4.357e+07  4.103e+00  1.594e-06  6.952e-07  1.594e-06  1.595e-06
  4  ASM_R  CONVERGED!
___________________________________________________________________________________________________________________________________________________________
Relative errors:
Displacement:	6.952e-07
Force: 		9.451e-10
Dilatation:	1.528e-06
v / V_0:	1.000e-09 / 1.000e-09 = 1.000e+00
```
and this is after the change
```
Timestep 10 @ 1.000e+00s
______________________________________________________________________________________________________________________________________________________
               SOLVER STEP                |  LIN_IT   LIN_RES    RES_NORM     RES_U     RES_P      RES_J     NU_NORM      NU_U       NU_P       NU_J
______________________________________________________________________________________________________________________________________________________
  0  CST  ASM_SYS  ------  SLV  PP  UQPH  |     456  1.905e-06  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00  1.000e+00
  1  CST  ASM_SYS  ------  SLV  PP  UQPH  |     260  2.967e-04  1.544e-01  1.544e-01  1.154e+13  1.849e+06  6.013e-02  7.397e-02  6.013e-02  6.013e-02
  2  CST  ASM_SYS  ------  SLV  PP  UQPH  |     279  2.099e-06  2.903e-03  2.903e-03  6.974e+10  2.060e+03  2.732e-03  1.436e-03  2.732e-03  2.733e-03
  3  CST  ASM_SYS  ------  SLV  PP  UQPH  |     280  2.494e-09  2.086e-06  2.086e-06  4.357e+07  4.103e+00  1.594e-06  6.952e-07  1.594e-06  1.595e-06
  4  CST  ASM_SYS  CONVERGED!
______________________________________________________________________________________________________________________________________________________
Relative errors:
Displacement:	6.952e-07
Force: 		9.451e-10
Dilatation:	1.528e-06
v / V_0:	1.000e-09 / 1.000e-09 = 1.000e+00
```
The results look the same to me, and the amended table output also looks OK.

### DIFF
--- a/examples/step-44/step-44.cc
+++ b/examples/step-44/step-44.cc
@@ -821,13 +821,10 @@ namespace Step44
     // object (see the @ref threads module for more information on this).
     //
     // We declare such structures for the computation of tangent (stiffness)
-    // matrix, right hand side, static condensation, and for updating
+    // matrix and right hand side vector, static condensation, and for updating
     // quadrature points:
-    struct PerTaskData_K;
-    struct ScratchData_K;
-
-    struct PerTaskData_RHS;
-    struct ScratchData_RHS;
+    struct PerTaskData_ASM;
+    struct ScratchData_ASM;
 
     struct PerTaskData_SC;
     struct ScratchData_SC;
@@ -844,29 +841,24 @@ namespace Step44
 
     void determine_component_extractors();
 
+    // Apply Dirichlet boundary conditions on the displacement field
+    void make_constraints(const int &it_nr);
+
     // Several functions to assemble the system and right hand side matrices
     // using multithreading. Each of them comes as a wrapper function, one
     // that is executed to do the work in the WorkStream model on one cell,
     // and one that copies the work done on this one cell into the global
     // object that represents it:
-    void assemble_system_tangent();
+    void assemble_system();
 
-    void assemble_system_tangent_one_cell(
+    void assemble_system_one_cell(
       const typename DoFHandler<dim>::active_cell_iterator &cell,
-      ScratchData_K &                                       scratch,
-      PerTaskData_K &                                       data) const;
+      ScratchData_ASM &                                     scratch,
+      PerTaskData_ASM &                                     data) const;
 
-    void copy_local_to_global_K(const PerTaskData_K &data);
+    void copy_local_to_global_system(const PerTaskData_ASM &data);
 
-    void assemble_system_rhs();
-
-    void assemble_system_rhs_one_cell(
-      const typename DoFHandler<dim>::active_cell_iterator &cell,
-      ScratchData_RHS &                                     scratch,
-      PerTaskData_RHS &                                     data) const;
-
-    void copy_local_to_global_rhs(const PerTaskData_RHS &data);
-
+    // And similar to perform global static condensation:
     void assemble_sc();
 
     void assemble_sc_one_cell(
@@ -875,9 +867,6 @@ namespace Step44
       PerTaskData_SC &                                      data);
 
     void copy_local_to_global_sc(const PerTaskData_SC &data);
-
-    // Apply Dirichlet boundary conditions on the displacement field
-    void make_constraints(const int &it_nr);
 
     // Create and update the quadrature points. Here, no data needs to be
     // copied into a global object, so the copy_local_to_global function is
@@ -1159,22 +1148,26 @@ namespace Step44
   // using TBB. Our main tool for this is the WorkStream class (see the @ref
   // threads module for more information).
 
-  // Firstly we deal with the tangent matrix assembly structures.  The
-  // PerTaskData object stores local contributions.
+  // Firstly we deal with the tangent matrix and right-hand side assembly
+  // structures. The PerTaskData object stores local contributions to the global
+  // system.
   template <int dim>
-  struct Solid<dim>::PerTaskData_K
+  struct Solid<dim>::PerTaskData_ASM
   {
     FullMatrix<double>                   cell_matrix;
+    Vector<double>                       cell_rhs;
     std::vector<types::global_dof_index> local_dof_indices;
 
-    PerTaskData_K(const unsigned int dofs_per_cell)
+    PerTaskData_ASM(const unsigned int dofs_per_cell)
       : cell_matrix(dofs_per_cell, dofs_per_cell)
+      , cell_rhs(dofs_per_cell)
       , local_dof_indices(dofs_per_cell)
     {}
 
     void reset()
     {
       cell_matrix = 0.0;
+      cell_rhs    = 0.0;
     }
   };
 
@@ -1184,18 +1177,22 @@ namespace Step44
   // gradient and symmetric gradient vector which we will use during the
   // assembly.
   template <int dim>
-  struct Solid<dim>::ScratchData_K
+  struct Solid<dim>::ScratchData_ASM
   {
-    FEValues<dim> fe_values;
+    FEValues<dim>     fe_values;
+    FEFaceValues<dim> fe_face_values;
 
     std::vector<std::vector<double>>                  Nx;
     std::vector<std::vector<Tensor<2, dim>>>          grad_Nx;
     std::vector<std::vector<SymmetricTensor<2, dim>>> symm_grad_Nx;
 
-    ScratchData_K(const FiniteElement<dim> &fe_cell,
-                  const QGauss<dim> &       qf_cell,
-                  const UpdateFlags         uf_cell)
+    ScratchData_ASM(const FiniteElement<dim> &fe_cell,
+                    const QGauss<dim> &       qf_cell,
+                    const UpdateFlags         uf_cell,
+                    const QGauss<dim - 1> &   qf_face,
+                    const UpdateFlags         uf_face)
       : fe_values(fe_cell, qf_cell, uf_cell)
+      , fe_face_values(fe_cell, qf_face, uf_face)
       , Nx(qf_cell.size(), std::vector<double>(fe_cell.n_dofs_per_cell()))
       , grad_Nx(qf_cell.size(),
                 std::vector<Tensor<2, dim>>(fe_cell.n_dofs_per_cell()))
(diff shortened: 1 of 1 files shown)


examples/step-44/step-44.cc
```diff
@@ -1761,30 +1690,39 @@ namespace Step44
...
-    // latter is an expensive operation and we can potentially avoid an extra
-    // assembly process by not assembling the tangent matrix when convergence
-    // is attained.
+    // forces, and collectively determine whether or not the equilibrium
+    // solution has been attained.
+    //
+    // Although for this particular problem we could potentially construct the
+    // RHS vector before assembling the system matrix, for the sake of
+    // extensibility we choose not to do so. The benefit to assembling the RHS
+    // vector and system matrix seperately is that latter is an expensive
```
tjhei, line 1702:
```suggestion
    // vector and system matrix separately is that latter is an expensive
```

examples/step-44/step-44.cc
```diff
@@ -821,13 +821,10 @@ namespace Step44
...
     // We declare such structures for the computation of tangent (stiffness)
-    // matrix, right hand side, static condensation, and for updating
+    // matrix and right hand side vector, static condensation, and for updating
     // quadrature points:
-    struct PerTaskData_K;
-    struct ScratchData_K;
-
-    struct PerTaskData_RHS;
-    struct ScratchData_RHS;
+    struct PerTaskData_ASM;
```
tjhei, line 826: Maybe we can call this ``PerTaskData`` now?

examples/step-44/step-44.cc
```diff
@@ -844,29 +841,24 @@ namespace Step44
 
     void determine_component_extractors();
 
+    // Apply Dirichlet boundary conditions on the displacement field
```
tjhei, line 844: The comment does not match the name of the function, do you apply constraints to the displacement or do you create constraints?

examples/step-44/step-44.cc
```diff
@@ -844,29 +841,24 @@ namespace Step44
...
+    void make_constraints(const int &it_nr);
```
tjhei, line 845:
```suggestion
    void make_constraints(const int it_nr);
```

tjhei (commented):
Unifying RHS and matrix assembly makes sense to me, but a quick question:
Do you now assemble the matrix when it is not needed or do you always need both anyways? In other situations I have a single ``assemble(bool only_rhs)`` function to be more efficient while still having the code unified.

examples/step-44/step-44.cc
```diff
@@ -2023,40 +1957,45 @@ namespace Step44
...
       [this](const typename DoFHandler<dim>::active_cell_iterator &cell,
-             ScratchData_K &                                       scratch,
-             PerTaskData_K &                                       data) {
-        this->assemble_system_tangent_one_cell(cell, scratch, data);
+             ScratchData_ASM &                                     scratch,
+             PerTaskData_ASM &                                     data) {
+        this->assemble_system_one_cell(cell, scratch, data);
+      },
+      [this](const PerTaskData_ASM &data) {
+        this->copy_local_to_global_system(data);
```
kronbichler, line 1997: Given that this program already uses lambdas, how about simply inlining the single call to `constraints.distribute_local_to_global` here? I am less sure regarding the `assemble_system_one_cell` call which is a long function, so what would you think about this hybrid approach?

examples/step-44/step-44.cc
```diff
@@ -2163,6 +2106,48 @@ namespace Step44
...
+            // tensor.
+            //
+            // At the loss of some readability, this small change will reduce
+            // the assembly time of the symmetrized system by about half when
+            // using the simulation default parameters, and becomes more
+            // significant as the h-refinement level increases.
+            const SymmetricTensor<2, dim> symm_grad_Nx_i_x_Jc =
+              symm_grad_Nx[i] * Jc;
+            const Tensor<1, dim> grad_Nx_i_comp_i_x_tau_ns =
+              grad_Nx[i][component_i] * tau_ns;
```
kronbichler, line 2148: I guess this does not really pertain to this PR, but I wonder why you use separate arrays for `symm_grad_Nx` and `grad_Nx`, where the former is simply `symmetrize(grad_Nx)`? I would find the `symmetrize` approach clearer because I have to keep track of one variable less, but it might be just me.

Note that I am fine with the temporary variable you introduce, it's just the array coming from the outside.

kronbichler (commented):
Looks good to me apart from some comments.

examples/step-44/step-44.cc
jppelteret: You can't see it in the diff, but within the next few lines another two PerTaskData structs (PerTaskData_SC and PerTaskData_UQPH) are defined. So I think that it makes sense to keep the naming as is to better distinguish between them.

examples/step-44/step-44.cc
jppelteret: Thanks for catching this typo!

examples/step-44/step-44.cc
jppelteret: True. It was a lazy cut and paste. I've changed the comment to "Create Dirichlet constraints for the incremental displacement field".

examples/step-44/step-44.cc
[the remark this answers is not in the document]
jppelteret: Yip, that article was missing. Thanks for spotting my mistake!

examples/step-44/step-44.cc
jppelteret: Thanks, that's a good idea! I also think that `assemble_system_one_cell` deserves its own function due to its length/complexity, but the copier can definitely go inline. I guess I'll add another commit that adds lambdas for the other two calls to `Workstream::run()`. I can then use the same approach there.

examples/step-44/step-44.cc
jppelteret: The simple answer is that I wanted to precompute as much as possible, and avoid the repeated call to `symmetrize(grad_Nx)` in the inner loop. I've got no concrete values to say how much quicker it is (if its quicker at all!) but it was just the philosophy I decided to apply right from the start and I stuck with it. Its true that its one more bit of bookkeeping to do, but that's one of several places in this tutorial where I'd gone for practicality or extensibility over the traditional or simple approach. 

Most of what's done here is, I guess, superseded by our fancy [`ScratchData`](https://www.dealii.org/9.2.0/doxygen/deal.II/classMeshWorker_1_1ScratchData.html) class, but in the past in other codes I would actually move the part that fills the cache somewhere else and share it between multiple weak forms.

> I guess this does not really pertain to this PR

I'd say it would have been a prime topic for discussion 10 years ago, when the tutorial was contributed :-D

jppelteret:
> Do you now assemble the matrix when it is not needed or do you always need both anyways? In other situations I have a single assemble(bool only_rhs) function to be more efficient while still having the code unified.

Thanks for your comment and suggestion! The only time that the tangent matrix is not needed is if we attain equilibrium, and then exit the NR loop without solving the linear system. So that's one unnecessary assembly of the matrix per timestep. But, of course, one can only know that once the residual is constructed, and so one could end up assembling the RHS without the tangent matrix, finding out that the equilibrium solution has not been attained, and then constructing it again along with the tangent matrix. 

I agree that what you suggest is not hard to do (adding a flag is an elegant way to do this), but perhaps in this case it's not worth doing. My initial feeling is that it would also be another twist to an already complex tutorial. Also, it would be contradictory to the new documentation that I added which states come cases for which you need to assemble both together to get the correct residual. So then I'd need to explain that...

examples/step-44/step-44.cc
```diff
@@ -2163,6 +2098,49 @@ namespace Step44
...
+            //
+            // At the loss of some readability, this small change will reduce
+            // the assembly time of the symmetrized system by about half when
+            // using the simulation default parameters, and becomes more
+            // significant as the h-refinement level increases.
+            if (i_group == u_dof)
+              {
+                symm_grad_Nx_i_x_Jc    = symm_grad_Nx[i] * Jc;
+                grad_Nx_i_comp_i_x_tau = grad_Nx[i][component_i] * tau_ns;
+              }
```
jppelteret, line 2141: In my original PR I accidentally introduced a bug here: `component_i` is only `< dim` when the condition `i_group == u_dof` is met. So I need this check for `grad_Nx[i][component_i]` not to go out of bounds.

jppelteret:
Many thanks for your comments @tjhei and @kronbichler! I've attended to them, and have also fixed a bug that I just discovered. I'll squash all commits at the end of the review.

tjhei (approved):
Looks good to me.

kronbichler (approved):

PR #11103, opened 2020-10-25
Pass n_q_points_1D to Simplex::QGauss

peterrum:
As discussed a while ago.


include/deal.II/simplex/quadrature_lib.h
```diff
@@ -28,12 +28,13 @@ namespace Simplex
...
-   * Following number of quadrature points are currently supported for 2D and
-   * 3D:
+   * Users specify the number of quadrature points in 1D direction n_points_1D.
+   * The given value is translated for n_points_1D=1,2,3 to following number of
+   * quadrature points for 2D and 3D:
    *   - 2D: 1, 3, 7
    *   - 3D: 1, 4, 10
    *
-   * For 1D, the quadrature rule degenerates to a `QGauss<1>(n_points)`.
+   * For 1D, the quadrature rule degenerates to a `QGauss<1>(n_points_1D)`.
```
kronbichler, line 37: I suggest to add a comment that the `n_points_1D` reflects the highest polynomial degree that can be integrated exactly, namely `2 n_points_1D - 1` (or is it only degree 4 for the 7/10 point rules on a simplex?); the number of points in 1D on its own can be misleading because these formulas are not a tensor product and thus there is no notion what the specified dimension would be. I think the important relation is to the number of 1D points of a tensor product quadrature formula reaching a similar degree of the polynomials.

peterrum: These are the highest polynomial degree that can be integrated exactly for TETs:
- tet_1 (open) -> precision 1
- tet_4 (open) -> precision 2
- tet_10 (closed) -> precision 3

for TRIs:
- tri_1 (open) -> precision 1
- tri_3 (open) -> precision 2
- tri_7 (open) -> precision 5

The values do not match the given formula. We could implement other quadrature rules (with more points - although I am not sure that we will match every precision value). @kronbichler What is your opionion?

kronbichler: Hm, the problem is the 7/10 point formulas which behave differently. As a side note, this means that the 10-point formula on the tetrahedron does not compute quadratic contributions exactly, so one has to be careful with that formula anyway.

From a user perspective, to me the most intuitive way to use a quadrature formula is give it a polynomial degree I want to have integrated exactly, possibly also somewhat larger polynomial degrees. However, that deviates from the setup we have on quadrilaterals where we explicitly talk about the number of points (see e.g. `QGauss` vs `QGaussLobatto`). Thus, the interface is a bit of a mess and I do not immediately see a good solution to it. I tend to think that suggesting that the `n_points_1D` as an *approximate* number of 1D points that is somehow linked to the polynomial degree (please include also the precision in the documentation) is the best option. I would not like to specify that I use "10" points on quadratic shape functions of the tetrahedron because it destroys dimension independent programming. But that's how far my strong opinions go, and I'd be happy to be convinced of something else as well.

peterrum: @kronbichler What do you think about requiring at least `2 n_points_1D - 1`? I can modify the quadrature formula in a follow-up PR!?

peterrum:
How do we want to proceed here? I would be happy if we could merge now and make the changes in follow up PR. I think we all agree that we don't want in user code constructs like these:
```cpp
  Simplex::QGauss<dim> quad(dim == 2 ? (params.degree == 1 ? 3 : 7) :
                                       (params.degree == 1 ? 4 : 10));
  Simplex::QGauss<dim - 1> face_quad(dim == 2 ? (params.degree == 1 ? 2 : 3) :
                                                (params.degree == 1 ? 3 : 7));
```

include/deal.II/simplex/quadrature_lib.h
kronbichler: I suggest we simply keep the proposed name for now, but add to the documentation that 
```c++
* Users specify a number `n_points_1D` as an indication of what polynomial degree
* to be integrated exactly, similarly to the number of points in a QGauss quadrature object,
* even though the present quadrature formula is not a tensor product. The given value
* is translated for n_points_1D=1,2,3 to following number of
* quadrature points for 2D and 3D:
```
to replace the comment in lines 31-33 above.

peterrum: Thanks for the proposed text. I have added it to the documentation.

kronbichler (approved):

Issue #11093, opened 2020-10-23
Provide a mechanism to update ghost_particles location and properties without regenerating them


blaisb:
Dear all, 
I hope you are well. As always, thank you very much for taking the time.
Our recent benchmark on our usage of the particle methods has highlighted some limitations in the code when it runs in parallel. These could not have been planned or identified by @gassmoeller, there is absolutely not an ounce of blame in what follows. I think @gassmoeller work on particles is just awesome.

Let me describe our application case and things will be clearer. We are implementing a DEM method in Lethe using the particles structures (see : https://www.youtube.com/watch?v=__5UGx4fQps&feature=youtu.be for an example of DEM). On a single core, it has exactly the performance we need right now and everything is perfect (it about 10% faster than LIGGGHTS, a state of the art DEM code based on LAMMPS).

What is important to know about DEM is that we do not need to call sort_particles_into_subdomain_and_cells every iteration. We actually do it every X iterations, where X is the frequency at which we want to couple DEM and the background equations being solved on the mesh.

Currently, deal.II does not provide any mechanism to update the ghost particles location and properties without calling exchange_ghost_particles. However, exchange_ghosts is a very expensive operation. Here is the code for exchange_ghost_particles: https://www.dealii.org/developer/doxygen/deal.II/particle__handler_8cc_source.html#l01160

There are two sub issues to this issue.
The first one is that at every exchange_ghost_particles() the following structure is generated:

    
     std::vector<std::set<unsigned int>> vertex_to_neighbor_subdomain(
       triangulation->n_vertices());
     for (const auto &cell : triangulation->active_cell_iterators())
       {
         if (cell->is_ghost())
           for (const unsigned int v : GeometryInfo<dim>::vertex_indices())
             vertex_to_neighbor_subdomain[cell->vertex_index(v)].insert(
               cell->subdomain_id());
       }

This is a relatively expensive operation since we are doing it millions of times. This sub issue can be solved easily using a deal.II cache I believe, like my PR from last year on sort_particles_into_subdomain_and_cells(). No worries here.

The second issue is more profound. At every call, exchange_ghost_particles() clears the entire ghost_particles structure and rebuilds it from scratch, assuming that ghost particles may have changed cells and everything. The underlying assumption is that we are calling sort_particles_into_subdomain_and_cells at every time step. Regretfully, this is not what we are doing. 

[...]

Issue #11078, opened 2020-10-22
Particle test failures


bangerth:
#11059 caused a number of regressions, as speculated by @kronbichler in a comment in #11059:
```
  particle_handler_08
  particle_handler_11
  particle_handler_12
  particle_handler_13
  particle_handler_15
  particle_handler_16
  particle_handler_17
```
I verified that #11059 is really the cause.

I'll have to investigate.

PR #11104, opened 2020-10-25
Exchange ghost particles optimization by not rebuilding vertex_to_neighbor_subdomain 

blaisb:
This is related to Issue #11093
This PR adds the vertex_to_neighbor_subdomain data structure to the grid_tools_cache. This allows the exchange_ghost_particles() function of the particle_handler not to rebuild this structure when it is not needed.
This is still a WIP. Right now a lot of particles tests are failing because of  a recent PR, so I am having issues finding if I am introducing additional bugs or no.

I must admit my ignorance with respect to the update flags of the grid_tools_cache object. I think this part of my PR might have issues within it? Could somebody look at it and correct me if I am wrong?

As always, all comments are more than appreciated. I will do whatever needed to integrate your comments to the PR. :)!

### DIFF
--- a/include/deal.II/grid/grid_tools_cache.h
+++ b/include/deal.II/grid/grid_tools_cache.h
@@ -153,10 +153,11 @@ namespace GridTools
     get_locally_owned_cell_bounding_boxes_rtree() const;
 
 
-    /** Returns the vector of set of integer containing the subdomain id
+    /**
+     * Returns the vector of set of integer containing the subdomain id
      * to which each vertex is connected to. This feature is used extensively
      * in the particle_handler to detect on which processors ghost particles
-     * must be built
+     * must be built.
      */
     const std::vector<std::set<unsigned int>> &
     get_vertex_to_neighbor_subdomain() const;
@@ -281,7 +282,7 @@ namespace GridTools
 
     /**
      * Store an std::vector of std::set of integer containing the id of all
-     * subdomain to which a vertex is connected to
+     * subdomain to which a vertex is connected to.
      */
     mutable std::vector<std::set<unsigned int>> vertex_to_neighbor_subdomain;
 
--- a/include/deal.II/grid/grid_tools_cache.h
+++ b/include/deal.II/grid/grid_tools_cache.h
@@ -152,6 +152,15 @@ namespace GridTools
                 typename Triangulation<dim, spacedim>::active_cell_iterator>> &
     get_locally_owned_cell_bounding_boxes_rtree() const;
 
+
+    /** Returns the vector of set of integer containing the subdomain id
+     * to which each vertex is connected to. This feature is used extensively
+     * in the particle_handler to detect on which processors ghost particles
+     * must be built
+     */
+    const std::vector<std::set<unsigned int>> &
+    get_vertex_to_neighbor_subdomain() const;
+
     /**
      * Return a reference to the stored triangulation.
      */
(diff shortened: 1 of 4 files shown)


include/deal.II/grid/grid_tools_cache.h
```diff
@@ -152,6 +152,15 @@ namespace GridTools
                 typename Triangulation<dim, spacedim>::active_cell_iterator>> &
     get_locally_owned_cell_bounding_boxes_rtree() const;
 
+
+    /** Returns the vector of set of integer containing the subdomain id
```
kronbichler, line 156:
```suggestion
    /**
     * Return the vector integers containing the subdomain id
```

include/deal.II/grid/grid_tools_cache.h
```diff
@@ -152,6 +152,15 @@ namespace GridTools
...
+     * to which each vertex is connected to. This feature is used extensively
+     * in the particle_handler to detect on which processors ghost particles
+     * must be built
```
kronbichler, line 159: In deal.II we aim to follow the convention of ending comments with a period (here and for the other comments).

kronbichler (approved):
This looks very good to me, apart from some comment beauty discussed below. But I'm not an expert on the cache/particle interface, so I would be happy to let someone else have a look as well.

kronbichler:
/rebuild
(I'm running the CI despite the problem with failing tests, as those failing are listed in #11078 do not go through the CI tests.)

blaisb:
> This looks very good to me, apart from some comment beauty discussed below. But I'm not an expert on the cache/particle interface, so I would be happy to let someone else have a look as well.

Thanks! I will fix the aesthetics of the comments today and push again :). Seems like the CI has passed which is good news.

blaisb:
I fixed the formatting of the comments :).

source/grid/grid_tools_cache.cc
```diff
@@ -199,6 +199,25 @@ namespace GridTools
     return covering_rtree[level];
   }
 
+  template <int dim, int spacedim>
+  const std::vector<std::set<unsigned int>> &
+  Cache<dim, spacedim>::get_vertex_to_neighbor_subdomain() const
+  {
+    if (update_flags & update_vertex_to_neighbor_subdomain)
```
peterrum, line 206: I must admit I don't get the logic. Isn't at the end of the function a `update_flags    = update_flags & ~update_vertex_to_neighbor_subdomain;` missing? Else you will clear and update the data structure at every call?

blaisb: Dear Peter,
To be honest, I am really unsure of how those flags work. I just copied their behavior from another place in the file
e.g. : 
`    if (update_flags & update_vertex_to_cell_map)
`

peterrum: It also contains the cancellation of the bit:

https://github.com/dealii/dealii/blob/e8b1d75b5b26cd125bc7f3b840f28e188a3a0537/source/grid/grid_tools_cache.cc#L63-L68

blaisb: You are right, the bit is never cancelled. I will fix that and push again :)!

peterrum (approved):

source/grid/grid_tools_cache.cc
kronbichler: Now I have a follow-up question: When is this flag actually set, so when do we make the fresh start here?

blaisb: That's a very good questions. I can't seem to find where the initial value of the flag is set. I'll dig deeper. I rebased the PR and test particles/particle_handler_06.mpirun=2.debug is failing, so clearly there is something missing in this PR. I think the triangulation has to specify the update of that flag and right now I did not code that.

blaisb: @kronbichler I change the update_flag to 0xFFF so that the new flag is updated. After discussion with @peterrum  we thought it would be better to just update everything and provision for additional flag in the future by setting update_flag to 0XFFF. The latest push should pass all tests (it passes them on my machine). :). Sorry for the confusion.

kronbichler (approved):

PR #11105, opened 2020-10-25
Make ConsensusAlgorithms "work" in serial

peterrum:
Nothing is done but at least I don't get segfaults if I use it a serial program.

### DIFF
--- a/include/deal.II/base/mpi_consensus_algorithms.templates.h
+++ b/include/deal.II/base/mpi_consensus_algorithms.templates.h
@@ -83,8 +83,10 @@ namespace Utilities
                                    const MPI_Comm & comm)
         : process(process)
         , comm(comm)
-        , my_rank(this_mpi_process(comm))
-        , n_procs(n_mpi_processes(comm))
+        , my_rank(Utilities::MPI::job_supports_mpi() ? this_mpi_process(comm) :
+                                                       0)
+        , n_procs(Utilities::MPI::job_supports_mpi() ? n_mpi_processes(comm) :
+                                                       1)
       {}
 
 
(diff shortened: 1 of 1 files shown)


include/deal.II/base/mpi_consensus_algorithms.templates.h
```diff
@@ -557,22 +559,26 @@ namespace Utilities
...
+            // Depending on the number of processes we switch between
+            // implementations. We reduce the threshold for debug mode to be
+            // able to test also the non-blocking implementation. This feature
+            // is tested by:
+            // tests/multigrid/transfer_matrix_free_06.with_mpi=true.with_p4est=true.with_trilinos=true.mpirun=15.output
 #ifdef DEAL_II_WITH_MPI
 #  if DEAL_II_MPI_VERSION_GTE(3, 0)
 #    ifdef DEBUG
-        if (Utilities::MPI::n_mpi_processes(comm) > 14)
+            if (Utilities::MPI::n_mpi_processes(comm) > 14)
```
kronbichler, line 572: Can't you use
```suggestion
            if (this->n_procs > 14)
```

include/deal.II/base/mpi_consensus_algorithms.templates.h
```diff
@@ -557,22 +559,26 @@ namespace Utilities
...
 #    else
-        if (Utilities::MPI::n_mpi_processes(comm) > 99)
+            if (Utilities::MPI::n_mpi_processes(comm) > 99)
 #    endif
-          consensus_algo.reset(new NBX<T1, T2>(process, comm));
-        else
+              consensus_algo.reset(new NBX<T1, T2>(process, comm));
+            else
```
kronbichler, line 577: and then use here
```suggestion
            else if (n_procs > 1)
```

kronbichler: Which would then allow you to limit the `job_supports_mpi` to a single place, the constructor of `Interface`. It would have the side effect that we do not try to set up anything for the single-rank case even if we are inside MPI.

peterrum: Done!

kronbichler (approved):

Issue #11106, opened 2020-10-26
clang-format version


kronbichler:
Given that clang-format 6 that we use for indentation is now around 2.5 years old (March 2018), I wonder at which point we want to upgrade to a newer version.

I have not checked myself how much of a difference the new clang-format 11 (release this month) would make in terms of code. We have a few things that are annoying, like https://github.com/dealii/dealii/pull/11081#issuecomment-714435660 or #8437 or certainly other places I don't remember, and it would be interesting to see if those are also present with the new version. I recall @masterleinad at some point said to have tried a newer version, but I do not remember what the conclusion was.

Rombur:
Is astyle still maintained? The last commit is over a year old.

bangerth:
I wasn't entirely serious. Last I looked, astyle hadn't been updated very much and I don't think it ever learned how to deal with lambda functions, for example. I do continue to think that the fact that clang-format re-arranges things between lines leads to worse outcomes. I am happy with how it indents each line individually.

tamiko:
I think we should find a consensus first on whether we want to update to `clang-format-11` and what options we want to enable.

Honestly, I think the final set of options we chose for clang-format when we switched away from astyle wasn't particularly good. clang-format has some problems with styles that deviate from the glorious one and only `LLVM` and `Google` styles.
`clang-format` actually recently learned the `GNU` style.

So I *strongly* suggest that this time with a possible (!!) switch to clang-format-11 we nuke our entire `.clang-format` file and we start with:
```
#
# The clang-format (Clang 11) style file used by deal.II. 
#
BasedOnStyle: "GNU"
```

and *then* we have a look at the result and identify behavior we don't like. We can then iterate on configuration options addressing this.

@masterleinad Shall we hijack your pull request and simply reindent everything just with this one command line option? :-)

masterleinad:
I believe that it will be quite impossible to review a pull request/style that implies so many changes so I would rather change options gradually based on the currents settings. That being said, I still have no clear preference for whether we should update or not.

tamiko:
@masterleinad Roger - let me start with the current version and adjust the two most crucial parameters then.

bangerth:
I'm ok if we figure out how to do this now, but could I suggest that we really only apply the patch in a quiet period? That could be during the winter break if we make a concerted effort to merge everything else first, or maybe right after the release?

There is no way we can actually *review* such a large change, but I think that in practice if several people just look at random places, we will find what we dislike.

[...]

Issue #8437, opened 2019-07-31
Look over clang-format indentation with many specifiers


kronbichler:
In #8428 we observed that `clang-format` does sometimes struggle with creating a good indentation when we use many specifiers, sometimes hidden behind the keywords. The result can look e.g.
```c++
static constexpr DEAL_II_ALWAYS_INLINE DEAL_II_CUDA_HOST_DEV T
                                                             value(const F &f,
                                                                   typename std::enable_if<
```
This looks ugly and sometimes even breaks the 80 char line length (as here).

Before the next release we should look this through systematically and find out whether
- an update of `clang-format` to a newer version would help
- there are some settings in `clang-format` that arrange items in a better way
- we can manually fix a few of the most offending lines as proposed by @tjhei in https://github.com/dealii/dealii/pull/8428#discussion_r308345193

kronbichler:
This is not critical and likely nobody is really interested in working on this now, so I'm adjusting the target milestone.

masterleinad:
Let's do #11108 after the release.

kronbichler:
#11108 has been done and it overall looks better, closing.

PR #11108, opened 2020-10-26
Indentation clang-format 11

masterleinad:
Related to #11106, this shows the indentation with clang-format 11.
I am not opposed to updating but this will be a little annoying for everyone submitting a pull request at first. If we decide to do this, we also should also provide pre-compiled versions for various platforms which was a little of a struggle last time.

Some of the diffs are
```diff
diff --git a/include/deal.II/base/numbers.h b/include/deal.II/base/numbers.h
index 4e3432c7bc..452628021a 100644
--- a/include/deal.II/base/numbers.h
+++ b/include/deal.II/base/numbers.h
@@ -561,7 +561,7 @@ namespace numbers
 
   template <typename number>
   constexpr DEAL_II_CUDA_HOST_DEV const number &
-                                        NumberTraits<number>::conjugate(const number &x)
+  NumberTraits<number>::conjugate(const number &x)
   {
     return x;
   }
@@ -690,7 +690,7 @@ namespace internal
   struct NumberType
   {
     static constexpr DEAL_II_ALWAYS_INLINE DEAL_II_CUDA_HOST_DEV const T &
-                                                                       value(const T &t)
+    value(const T &t)
     {
       return t;
     }
@@ -705,8 +705,8 @@ namespace internal
     // Type T is constructible from F.
     template <typename F>
     static constexpr DEAL_II_ALWAYS_INLINE DEAL_II_CUDA_HOST_DEV T
-                                                                 value(const F &f,
-                                                                       typename std::enable_if<
+    value(const F &f,
+          typename std::enable_if<
             !std::is_same<typename std::decay<T>::type,
                           typename std::decay<F>::type>::value &&
             std::is_constructible<T, F>::value>::type * = nullptr)
@@ -717,8 +717,8 @@ namespace internal
     // Type T is explicitly convertible (but not constructible) from F.
     template <typename F>
     static constexpr DEAL_II_ALWAYS_INLINE T
-                                           value(const F &f,
-                                                 typename std::enable_if<
+    value(const F &f,
+          typename std::enable_if<
             !std::is_same<typename std::decay<T>::type,
                           typename std::decay<F>::type>::value &&
             !std::is_constructible<T, F>::value &&
```
so there are at least some improvements.

### DIFF
--- a/contrib/utilities/indent_common.sh
+++ b/contrib/utilities/indent_common.sh
@@ -39,7 +39,7 @@ checks() {
 
   # Add the location 'download_clang_format' or 'compile_clang_format'
   # installs clang-format to the local PATH.
-  CLANG_FORMAT_PATH="$(cd "$(dirname "$0")" && pwd)/programs/clang-6/bin"
+  CLANG_FORMAT_PATH="$(cd "$(dirname "$0")" && pwd)/programs/clang-11/bin"
   export PATH="${CLANG_FORMAT_PATH}:${PATH}"
 
   if ! [ -x "$(command -v "${DEAL_II_CLANG_FORMAT}")" ]; then
@@ -51,14 +51,13 @@ checks() {
     exit 1
   fi
 
-  # Make sure to have the right version. We know that clang-6.0.0
-  # and clang-6.0.1 work. Hence, test for clang-6.0.
+  # Make sure to have the right version.
   CLANG_FORMAT_VERSION="$(${DEAL_II_CLANG_FORMAT} --version)"
   CLANG_FORMAT_MAJOR_VERSION=$(echo "${CLANG_FORMAT_VERSION}" | sed 's/^[^0-9]*\([0-9]*\).*$/\1/g')
   CLANG_FORMAT_MINOR_VERSION=$(echo "${CLANG_FORMAT_VERSION}" | sed 's/^[^0-9]*[0-9]*\.\([0-9]*\).*$/\1/g')
 
-  if [ "${CLANG_FORMAT_MAJOR_VERSION}" -ne 11 ] || [ "${CLANG_FORMAT_MINOR_VERSION}" -ne 0 ]; then
-    echo "***   This indent script requires clang-format version 6.0,"
+  if [ "${CLANG_FORMAT_MAJOR_VERSION}" -ne 11 ] || [ "${CLANG_FORMAT_MINOR_VERSION}" -ne 1 ]; then
+    echo "***   This indent script requires clang-format version 11.1,"
     echo "***   but version ${CLANG_FORMAT_MAJOR_VERSION}.${CLANG_FORMAT_MINOR_VERSION} was found instead."
     echo "***"
     echo "***   You can run the './contrib/utilities/download_clang_format'"
--- a/contrib/utilities/indent_common.sh
+++ b/contrib/utilities/indent_common.sh
@@ -57,7 +57,7 @@ checks() {
   CLANG_FORMAT_MAJOR_VERSION=$(echo "${CLANG_FORMAT_VERSION}" | sed 's/^[^0-9]*\([0-9]*\).*$/\1/g')
   CLANG_FORMAT_MINOR_VERSION=$(echo "${CLANG_FORMAT_VERSION}" | sed 's/^[^0-9]*[0-9]*\.\([0-9]*\).*$/\1/g')
 
-  if [ "${CLANG_FORMAT_MAJOR_VERSION}" -ne 6 ] || [ "${CLANG_FORMAT_MINOR_VERSION}" -ne 0 ]; then
+  if [ "${CLANG_FORMAT_MAJOR_VERSION}" -ne 11 ] || [ "${CLANG_FORMAT_MINOR_VERSION}" -ne 0 ]; then
     echo "***   This indent script requires clang-format version 6.0,"
     echo "***   but version ${CLANG_FORMAT_MAJOR_VERSION}.${CLANG_FORMAT_MINOR_VERSION} was found instead."
     echo "***"
--- a/examples/step-21/step-21.cc
+++ b/examples/step-21/step-21.cc
@@ -647,7 +647,7 @@ namespace Step21
     std::vector<Tensor<2, dim>> k_inverse_values(n_q_points);
 
     std::vector<Vector<double>>              old_solution_values(n_q_points,
-                                                                 Vector<double>(dim + 2));
+                                                    Vector<double>(dim + 2));
     std::vector<std::vector<Tensor<1, dim>>> old_solution_grads(
       n_q_points, std::vector<Tensor<1, dim>>(dim + 2));
 
--- a/examples/step-41/step-41.cc
+++ b/examples/step-41/step-41.cc
@@ -82,7 +82,7 @@ namespace Step41
     void setup_system();
     void assemble_system();
     void
-         assemble_mass_matrix_diagonal(TrilinosWrappers::SparseMatrix &mass_matrix);
+    assemble_mass_matrix_diagonal(TrilinosWrappers::SparseMatrix &mass_matrix);
     void update_solution_and_constraints();
     void solve();
     void output_results(const unsigned int iteration) const;
--- a/include/deal.II/base/table_indices.h
+++ b/include/deal.II/base/table_indices.h
@@ -66,12 +66,14 @@ public:
   /**
    * Read-only access the value of the <tt>i</tt>th index.
    */
-  constexpr std::size_t operator[](const unsigned int i) const;
+  constexpr std::size_t
+  operator[](const unsigned int i) const;
 
   /**
    * Write access the value of the <tt>i</tt>th index.
    */
-  constexpr std::size_t &operator[](const unsigned int i);
+  constexpr std::size_t &
+  operator[](const unsigned int i);
 
   /**
    * Compare two index fields for equality.
@@ -125,8 +127,8 @@ constexpr TableIndices<N>::TableIndices(const T... args)
 
 
 template <int N>
-constexpr inline std::size_t TableIndices<N>::
-                             operator[](const unsigned int i) const
+constexpr inline std::size_t
+TableIndices<N>::operator[](const unsigned int i) const
 {
   AssertIndexRange(i, N);
   return indices[i];
@@ -134,7 +136,8 @@ constexpr inline std::size_t TableIndices<N>::
 
 
 template <int N>
-constexpr inline std::size_t &TableIndices<N>::operator[](const unsigned int i)
+constexpr inline std::size_t &
+TableIndices<N>::operator[](const unsigned int i)
 {
   AssertIndexRange(i, N);
   return indices[i];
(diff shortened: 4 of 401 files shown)


examples/step-50/step-50.cc
```diff
@@ -504,8 +504,7 @@ void LaplaceProblem<dim, degree>::setup_system()
 
   switch (settings.solver)
     {
-      case Settings::gmg_mf:
-        {
+        case Settings::gmg_mf: {
```
tjhei, line 507: ugly

examples/step-21/step-21.cc
```diff
@@ -1174,8 +1174,7 @@ namespace Step21
                   << ", dt=" << time.get_previous_step_size() << '.'
                   << std::endl
                   << std::endl;
-      }
-    while (time.is_at_end() == false);
+    } while (time.is_at_end() == false);
```
tjhei, line 1177: I don't like this behavior.

tjhei (commented):
I am not sure if switching makes sense or not, but I dislike some of the changes. I assume some of them can be controlled by options.

tjhei:
> I assume some of them can be controlled by options.

Before you spend more time on this, @masterleinad , we should decide what to do.

Does the switch fix the problems in #11106?

kronbichler:
> Does the switch fix the problems in #11106?

At least the one in #8437 looks much better from what I see in the top post. If that alone is enough the pain with a different version, I don't know. I am nonetheless positive about it, albeit not very strongly and only if we have a majority of maintainers positive to this topic.

examples/step-28/step-28.cc
```diff
@@ -1664,8 +1663,8 @@ namespace Step28
               }
 
             ++iteration;
-          }
-        while ((error > parameters.convergence_tolerance) && (iteration < 500));
+        } while ((error > parameters.convergence_tolerance) &&
```
bangerth, line 1666: This seems wrong -- the closing brace no longer lines up vertically with the opening brace...

examples/step-33/step-33.cc
```diff
@@ -2126,20 +2122,19 @@ namespace Step33
...
-        // iterations), and then create the direct solver object. When
-        // actually doing the solve, note that we don't pass a
-        // preconditioner. That wouldn't make much sense for a direct solver
-        // anyway.  At the end we return the solver control statistics &mdash;
-        // which will tell that no iterations have been performed and that the
-        // final linear residual is zero, absent any better information that
-        // may be provided here:
-        case Parameters::Solver::direct:
-          {
+          // If the parameter file specified that a direct solver shall be used,
```
bangerth, line 2125: I also don't understand this one -- it's indenting the comment four (instead of two) characters relative to the opening brace.

examples/step-41/step-41.cc
```diff
@@ -85,7 +85,7 @@ namespace Step41
     void setup_system();
     void assemble_system();
     void
-         assemble_mass_matrix_diagonal(TrilinosWrappers::SparseMatrix &mass_matrix);
+    assemble_mass_matrix_diagonal(TrilinosWrappers::SparseMatrix &mass_matrix);
```
bangerth, line 88: This, at least, is progress :-)

tamiko:
This will also require to update our download and compile magic for `clang-format`. But first things first, we should find a consensus whether `clang-format-11` is finally a version we want to update to.

tamiko:
@masterleinad Ah, now I understand. Would you mind to update `examples/.clang-format` as well? :smile:

include/deal.II/integrators/patches.h
```diff
@@ -38,9 +38,9 @@ namespace LocalIntegrators
   {
     template <int dim>
     inline void
-      points_and_values(Table<2, double> &                          result,
-                        const FEValuesBase<dim> &                   fe,
-                        const ArrayView<const std::vector<double>> &input)
+    points_and_values(Table<2, double> &                          result,
+                      const FEValuesBase<dim> &                   fe,
+                      const ArrayView<const std::vector<double>> &input)
```
tamiko, line 43: I am a bit dissapointed that clang-format still cannot align the `&` properly :-(

tamiko:
I think that this configuration with clang-format-11 is now a strict improvement over what we had before and doesn't introduce any new unwanted behavior. Let us discuss at some point whether this change is worth the infrastructure update. :smile:

include/deal.II/base/table_indices.h
```diff
@@ -66,12 +66,14 @@ class TableIndices
   /**
    * Read-only access the value of the <tt>i</tt>th index.
    */
-  DEAL_II_CONSTEXPR std::size_t operator[](const unsigned int i) const;
+  DEAL_II_CONSTEXPR std::size_t
+                    operator[](const unsigned int i) const;
```
drwells, line 70: These look weird but now that we have C++14 perhaps we can fix it by just promoting these from `DEAL_II_CONSTEXPR` to `constexpr`.

masterleinad: We know that at least MSVC has problems and the promotion would not work in general. https://github.com/dealii/dealii/blob/master/cmake/checks/check_01_cxx_features.cmake#L572-L621
Of course, we could try in selected places.

drwells: I'll give it a try

bangerth:
Here's another consistent source of poor indentation: rigidly sticking to 80 chars per line. This forces the tool to break lines, even if they are just a bit longer, and there are many places where we get poor line breaks, or where stuff isn't vertically aligned because it doesn't fit. For example (`step-22.cc`):

* Here, we could have nicely aligned everything if we had more space
```
                symgrad_phi_u[k] =
                  fe_values[velocities].symmetric_gradient(k, q);
                div_phi_u[k] = fe_values[velocities].divergence(k, q);
                phi_p[k]     = fe_values[pressure].value(k, q);
```
or here, where the second argument to the constructor is not aligned with the first one:
```
      SolverControl            solver_control(solution.block(1).size(),
                                   1e-6 * schur_rhs.l2_norm());
```
or here, where nothing is really aligned:
```
  Assert(
    this->at_boundary(),
    ExcMessage(
      "You are trying to set the boundary_id of an internal object, which is illegal!"));
```

I think that a lot of these issues would go away if we gave ourselves 90 or 100 characters.  I think that also shouldn't be an issue any more. The 80 character convention originates in how many characters fit onto a text-only screen, and matches what fit onto a punch card. But our screens are wider and fit more text today -- for example, the longest lines you see in the following picture are 77 characters wide:
image
There is certainly a limit to what is easily readable, but I don't see a good reason not to go for 100 characters -- in practice, quite few lines will actually be this long.

kronbichler:
I would be very happy with a 100 character limit of lines (but we really need a quiet time for that because we would change a considerable number of lines).

masterleinad:
I propose to look into this right after the release.

contrib/utilities/indent_common.sh
```diff
@@ -57,7 +57,7 @@ checks() {
   CLANG_FORMAT_MAJOR_VERSION=$(echo "${CLANG_FORMAT_VERSION}" | sed 's/^[^0-9]*\([0-9]*\).*$/\1/g')
   CLANG_FORMAT_MINOR_VERSION=$(echo "${CLANG_FORMAT_VERSION}" | sed 's/^[^0-9]*[0-9]*\.\([0-9]*\).*$/\1/g')
 
-  if [ "${CLANG_FORMAT_MAJOR_VERSION}" -ne 6 ] || [ "${CLANG_FORMAT_MINOR_VERSION}" -ne 0 ]; then
+  if [ "${CLANG_FORMAT_MAJOR_VERSION}" -ne 11 ] || [ "${CLANG_FORMAT_MINOR_VERSION}" -ne 0 ]; then
     echo "***   This indent script requires clang-format version 6.0,"
```
marcfehling, line 61:
```suggestion
    echo "***   This indent script requires clang-format version 11.0,"
```

masterleinad:
We need to upload binaries for `Linux` and `MacOS` but otherwise, this should be good already.

tamiko:
@masterleinad 75% done, just waiting for the mac binary(ies). I suggest once ready, you simply incorporate #12445 into this PR?

masterleinad:
> @masterleinad 75% done, just waiting for the mac binary(ies). I suggest once ready, you simply incorporate #12445 into this PR?

Yes, already done. The indentation check is passing, too.

include/deal.II/base/table_indices.h
drwells: We fixed a lot of these in #11122 but we can't yet get rid of `DEAL_II_CONSTEXPR` - we need it to compile with some older compilers.

masterleinad:
It looks like only updating @tjhei's CI scripts is missing here now. Since these containers don't get updated automatically, this needs to be done at the same time as merging this pull request.

masterleinad:
Updating the clang-format version in the CI is also necessary to run it at all.

tamiko (approved):

kronbichler (approved):

luca-heltai:
> Updating the clang-format version in the CI is also necessary to run it at all.

This is done automatically in the actions:

```
 - name: setup
      run: |
        sudo apt update && sudo apt install doxygen graphviz perl texlive-bibtex-extra
        doxygen --version
```

masterleinad:
@luca-heltai What I meant is that @tjhei's containers don't update automatically (and that's why one of the `indent` steps and the `CI` step is failing).

luca-heltai:
Sorry, I thought we agreed that we should remove the indent step from the CI.

luca-heltai (approved):

kronbichler (approved):

masterleinad:
> Sorry, I thought we agreed that we should remove the indent step from Timo's CI.

I think we agreed on removing the `indent` check but I'm not sure about the `CI` step. I see some value in only trigger these expensive tests when at least the indentation is correct.

masterleinad:
@tjhei Any chance that we can move forward here (disabling the indentation check for the CI on your side or upgrading it at the time as merging this pull request)?

tjhei:
sure, you are free to disable it for now.

masterleinad:
> sure, you are free to disable it for now.

AFAICT, neither `CI` nor `indent` is defined within the repository; I only see Jenkinsfiles for `ready`, `OSX`, and `tidy`. Someone with `owner` permission status can disable `tidy` (since it is redundant due to a corresponding `GitHub` action) but it seems that for `CI` some action by @tjhei is still required.

tjhei:
Did you see ./Jenkinsfile ?

masterleinad:
> Did you see ./Jenkinsfile ?

Seems that I missed that. Thanks!

tamiko (approved):

kronbichler (approved):

tjhei:
> disable tidy (since it is redundant 

Wait, where are we running clang tidy already?

masterleinad:
> Wait, where are we running clang tidy already?

Oh, typo. I meant `format`. As can be seen in the checks `tidy` is still running.

peterrum (approved):

marcfehling (approved):

PR #11111, opened 2020-10-27
Make sure particles also copy their properties when switching property pools.

### LINKED ISSUE #11078: Particle test failures

bangerth:
This fixes a bug introduced in #11059. This used to accidentally work before #11059 because while we switched property pools, we accidentally kept a handle to the old property pool (which in turn worked because the then implementation of property pools didn't keep any state but simply allocated/deallocated memory on the heap). The solution is to copy properties from the old to the new property pool.

Fixes #11078.

### DIFF
--- a/include/deal.II/particles/particle.h
+++ b/include/deal.II/particles/particle.h
@@ -336,9 +336,9 @@ namespace Particles
      * function is after particle transfer to a new process.
      *
      * If a particle already stores properties in a property pool, then
-     * these are released and the particle will not store any properties
-     * after this call (though new properties can be assigned to it using
-     * the set_properties() call).
+     * their values are saved, the memory is released in the previous
+     * property pool, and a copy of the particle's properties will be
+     * allocated in the new property pool.
      */
     void
     set_property_pool(PropertyPool &property_pool);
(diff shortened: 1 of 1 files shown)


blaisb (approved):
I know I have no say in this, but to me this seems like a very good PR. No comments here.

bangerth:
@blaisb -- this is a community project. We value everyone's input, regardless of their formal status in the project. You're certainly a respected voice over here, and your input is valued!

bangerth:
As an aside, this patch passed all of the local `particle_handler.*debug` tests. These are apparently not run by the testers, or we would have found the issue in #11059 already.

kronbichler (approved):

blaisb:
> As an aside, this patch passed all of the local `particle_handler.*debug` tests. These are apparently not run by the testers, or we would have found the issue in #11059 already.

That sentence made me slightly confused :). Do you mean that the testers do not run the particles tests? I thought they did.

blaisb:
> > As an aside, this patch passed all of the local `particle_handler.*debug` tests. These are apparently not run by the testers, or we would have found the issue in #11059 already.
> 
> That sentence made me slightly confused :). Do you mean that the testers do not run the particles tests or just not their debug version?

kronbichler:
>  Do you mean that the testers do not run the particles tests or just not their debug version?

The CI testers (run for every PR) run all debug tests without MPI (around 4300 out of 6200+ tests in debug), as well as a subset of parallel tests (around 600). The parallel particle tests are not part of that selection.

On the other hand, the regression tests listed [here](https://cdash.43-1.org/index.php?project=deal.II) run all or almost all tests, but only around once per day. Here we tolerate some failures (e.g., one of our primary targets https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=12663 has 12 failures right now) that we aim to have fixed before every release at the latest.

PR #11113, opened 2020-10-29
Do not capture by reference in range-based for over IndexSet

kronbichler:
As observed in #11074 and #11104, the OSX CI machines complain about us using a `const auto &i` to an `IndexSet`. We should capture that variable by value as it is simply an `unsigned int` or `unsigned long int`.


peterrum (approved):

masterleinad (approved):

tjhei:
I tried to understand why master broke but I don't see a change that caused this. Is this because the Github OSX machine got updated to a newer version silently?

kronbichler:
Yes, a new compiler is the likely reason in my understanding as well. Not that I'm complaining about the content of the error, the compiler was absolutely right that capture by value was better.

PR #11114, opened 2020-10-29
Switch SymmetricTensor -> Tensor for better performance

kronbichler:
Follow-up to #11085: Some of my profiles revealed that using `SymmetricTensor` in loops which are not unrolled is a bad idea, because the access functionality is slow (like 10x the number of instructions compared to the new version with `Tensor`). Given that I don't know how to improve that I simply set up the full matrix on the stack despite its symmetry; computing it all allows the innermost loop to be vectorized by the compiler, which gains more speed than skipping the upper diagonal.

### DIFF
--- a/include/deal.II/fe/mapping_q_internal.h
+++ b/include/deal.II/fe/mapping_q_internal.h
@@ -912,7 +912,7 @@ namespace internal
             return;
           }
 
-        SymmetricTensor<2, n_functions> matrix;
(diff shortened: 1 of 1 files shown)


drwells:
I'm surprised that computing the off-diagonal part twice is faster due to loop unrolling than computing it once.

drwells (approved):
The CI failure is unrelated and addressed in #11113 so lets merge.

PR #11116, opened 2020-10-29
Fix multiple definitions for CUDAWrappes::internal::used_objects

### LINKED ISSUE #11115: Global Variable In cuda_matrix_free
The cuda_matrix_free header file creates a global variable,

```
include/matrix_free/cuda_matrix_free.templates.h:55      std::array<std::atomic_bool, mf_n_concurrent_objects> used_objects;
```

This restricts my options for using the library, since then I can only
include this header from one source file.

If I include it in multiple places, then multiple conflicting global variables
are present at linking, leading to the errors,

```
CMakeFiles/kernel.dir/helmholtz.cu.o:(.bss+0x0): multiple definition of `dealii::CUDAWrappers::internal::used_objects'
CMakeFiles/kernel.dir/operator.cu.o:(.bss+0x0): first defined here
CMakeFiles/kernel.dir/main.cu.o:(.bss+0x0): multiple definition of `dealii::CUDAWrappers::internal::used_objects'
CMakeFiles/kernel.dir/operator.cu.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
CMakeFiles/kernel.dir/build.make:228: recipe for target 'kernel' failed
make[2]: *** [kernel] Error 1
CMakeFiles/Makefile2:92: recipe for target 'CMakeFiles/kernel.dir/all' failed
make[1]: *** [CMakeFiles/kernel.dir/all] Error 2
Makefile:146: recipe for target 'all' failed
make: *** [all] Error 2
```

Global variables should be moved into the file that initializes dealii.

masterleinad:
Fixes #11115. @frobnitzem Can you check if that works for you?

### DIFF
--- a/source/matrix_free/cuda_matrix_free.cu
+++ b/source/matrix_free/cuda_matrix_free.cu
@@ -19,8 +19,15 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+
+
 namespace CUDAWrappers
 {
+  namespace internal
+  {
+    extern std::array<std::atomic_bool, mf_n_concurrent_objects> used_objects;
+  }
+
   // Do not instantiate for dim = 1
   template class MatrixFree<2, float>;
   template class MatrixFree<2, double>;
--- a/source/matrix_free/cuda_matrix_free.cu
+++ b/source/matrix_free/cuda_matrix_free.cu
@@ -25,7 +25,7 @@ namespace CUDAWrappers
 {
   namespace internal
   {
-    extern std::array<std::atomic_bool, mf_n_concurrent_objects> used_objects;
+    std::array<std::atomic_bool, mf_n_concurrent_objects> used_objects;
   }
 
   // Do not instantiate for dim = 1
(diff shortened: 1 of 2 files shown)


source/matrix_free/cuda_matrix_free.cu
```diff
@@ -19,8 +19,15 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+
+
 namespace CUDAWrappers
 {
+  namespace internal
+  {
+    extern std::array<std::atomic_bool, mf_n_concurrent_objects> used_objects;
```
drwells, line 28:
```suggestion
    std::array<std::atomic_bool, mf_n_concurrent_objects> used_objects;
```
You don't need `extern` here to define the storage of the object.

drwells (changes requested):

drwells:
I think this is fine but the cuda tester failed for unrelated reasons.

/rebuild

drwells (approved):

masterleinad:
> I think this is fine but the cuda tester failed for unrelated reasons.

Yes, it works now. BTW the GitHub workflows need to be restarted by hand. There is no magic comment as we have for the rest of the CI.

drwells:
> Yes, it works now. BTW the GitHub workflows need to be restarted by hand. There is no magic comment as we have for the rest of the CI.

Do we have documentation on how to do this somewhere?

masterleinad:
> Do we have documentation on how to do this somewhere?

It's https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/re-running-a-workflow.

PR #11117, opened 2020-10-29
Fix Wedge::n_faces().

drwells:
Wedges have five faces. I noticed this when writing tests for #10899.

/rebuild


drwells:
I think something just went wrong on GitHub - I only opened one pull request. Let me close this in favor of #11118.

PR #11119, opened 2020-10-29
update to clang-format-11 restart with fresh .clang-format

### LINKED ISSUE #11106: clang-format version

tamiko:
Closes #11106
Closes #11108

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,66 +1,14 @@
 #
-# The clang-format (Clang 6) style file used by deal.II.
+# The clang-format (Clang 11) style file used by deal.II.
 #
+
+BasedOnStyle: "GNU"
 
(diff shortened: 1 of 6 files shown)


tamiko:
Let me push onto #11108 instead :-)

PR #11121, opened 2020-10-30
Implement face_to_cell_vertices for pyramids and wedges.

drwells:
I need this to finish testing boundary id assignment for #10899.

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -783,10 +783,37 @@ namespace ReferenceCell
         {
           AssertIndexRange(face_no, n_faces());
 
-          if (face_no == 1)
+          if (face_no == 0)
             return ReferenceCell::Type::Quad;
           else
             return ReferenceCell::Type::Tri;
+        }
+
+        unsigned int
+        face_to_cell_vertices(
+          const unsigned int  face,
+          const unsigned int  vertex,
+          const unsigned char face_orientation) const override
+        {
+          AssertIndexRange(face, n_faces());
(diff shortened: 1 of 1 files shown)


include/deal.II/grid/reference_cell.h
```diff
@@ -783,11 +783,39 @@ namespace ReferenceCell
         {
           AssertIndexRange(face_no, n_faces());
 
-          if (face_no == 1)
+          if (face_no == 0)
```
peterrum, line 786: Jup. This was a typo.

include/deal.II/grid/reference_cell.h
```diff
@@ -783,11 +783,39 @@ namespace ReferenceCell
...
+#ifdef DEBUG
+          if (face == 0)
+            {
+              AssertIndexRange(vertex, 4);
+            }
+          else
+            {
+              AssertIndexRange(vertex, 3);
+            }
+#endif
```
peterrum, line 807: Why debug mode? `AssertIndexRange` is only evaluated in debug mode, isn't it?

peterrum (approved):
Thanks a lot, @drwells! I have checked: the tables are identical to what is contained in the pictures part of #11095.

include/deal.II/grid/reference_cell.h
drwells: I can't remember if the compiler will complain about empty if-statements or not. If the CI passes then this is probably fine - I'll force push a version without the `#ifdef` check.

peterrum (approved):

PR #11122, opened 2020-10-30
Check MSVC's constexpr support.

drwells:
We use constexpr for operator[] in Tensor - perhaps it will work here too.

### DIFF
--- a/include/deal.II/base/symmetric_tensor.h
+++ b/include/deal.II/base/symmetric_tensor.h
@@ -57,16 +57,16 @@ constexpr DEAL_II_ALWAYS_INLINE SymmetricTensor<4, dim, Number>
                                 invert(const SymmetricTensor<4, dim, Number> &);
 
 template <int dim2, typename Number>
-DEAL_II_CONSTEXPR inline DEAL_II_ALWAYS_INLINE Number
-                                               trace(const SymmetricTensor<2, dim2, Number> &);
+constexpr inline DEAL_II_ALWAYS_INLINE Number
+                                       trace(const SymmetricTensor<2, dim2, Number> &);
 
 template <int dim, typename Number>
-DEAL_II_CONSTEXPR inline DEAL_II_ALWAYS_INLINE SymmetricTensor<2, dim, Number>
-                                               deviator(const SymmetricTensor<2, dim, Number> &);
+constexpr inline DEAL_II_ALWAYS_INLINE SymmetricTensor<2, dim, Number>
+                                       deviator(const SymmetricTensor<2, dim, Number> &);
 
 template <int dim, typename Number>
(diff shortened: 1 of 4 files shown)


masterleinad:
Seems to work. :slightly_smiling_face:

drwells:
If the CI passing here is enough to prove that this works for C++14 + MSVC then I'll go ahead and add some more functions like this.

masterleinad:
> If the CI passing here is enough to prove that this works for C++14 + MSVC then I'll go ahead and add some more functions like this.

Yes, that's good enough for me.

masterleinad:
BTW, https://github.com/masterleinad/dealii/actions/runs/338613981 shows the problems if you were to replace all places using `DEAL_II_CONSTEXPR`.

rezarastak:
@masterleinad I think the problem you see is because of `standard_tensors.cc`. Currently there is `#define` statements which determines whether we want to define the constants in the header file or the source file. Now that we are using the `constexpr` feature directly, we must remove the initialization in the cc file altogether. For example, we shouldn't allow initialization of the constant here by replacing 

https://github.com/dealii/dealii/blob/2d712df2abcc2da14253ed41288146364607311a/source/physics/elasticity/standard_tensors.cc#L24-L30

with

```c++
template <int dim> 
constexpr SymmetricTensor<2, dim> Physics::Elasticity::StandardTensors<dim>::I;
```

That's my guess. The issue might be something else. I haven't tested my guess yet.

drwells:
I'm not sure how to handle those difficult initialization cases but we should go ahead with the easy ones.

bangerth:
@drwells There are a good number of compile failures here: https://jenkins.tjhei.info/blue/organizations/jenkins/dealii/detail/PR-11122/20/pipeline/ Can you take a look?

drwells:
It looks like I converted a few cases that are actually a constexpr bug on a non-MSVC compiler - I need to convert those back and we should be set.

tjhei:
Do you have access to an Intel compiler? It might be worth checking if things are okay with this PR as well.

drwells:
There's one on a cluster here I can access. I'll check that next.

bangerth:
What should we do here? The check that fails is for GCC 5.4 (the "CI" check), which I have battled to a substantial degree as well regarding `constexpr` in the last week.

drwells:
I've had this on the back burner for awhile for various reasons. I'll try to return to this PR soon.

drwells:
In particular, I should be able to finish it by installing my own copy of GCC 5.4 and then partially reverting the second commit.

drwells:
I had to revert a bunch of code to get this to work with older versions of clang and GCC but this is now ready.

bangerth (approved):
Nice work!

Issue #11102, opened 2020-10-25
Hide `hp_capability_enabled` parameter


marcfehling:
The huge effort of merging both DoFHandlers in  #10328 introduced a parameter `hp_capability_enabled` to still distinguish between both functionalities.

We can actually hide this parameter entirely from the interface and make it private: Depending on which kind of finite element is associated with the DoFHandler, we can set the `hp_capability_enabled` variable accordingly. It will be false for `FiniteElement` or `hp::FECollection` of length 1, and `true` for `hp::FECollection` of length > 1.

I will come up with a patch soon and figure out whether it works. We should make this part of the 9.3 release as this parameter has been introduced since the last release.

---

EDIT: I decided to split solving this issue in multiple PRs:
- [x] #11123 Hide `hp_capability_enabled` parameter.
- [x] #11133 Rework interface to `Triangulation` signals.
- [x] #11160 Slim interface for `DoFHandler` initialization.
- [x] #11149 Simplify `DoFHandler::distribute_dofs`.
- [x] #11220 Deprecate `DoFHandlerType` template argument.
- [x] #11153 Avoid using deprecated DoFHandler functions.

marcfehling:
While preparing a patch for this issue, I must admit that I am not happy with the current interface.

The characteristics of the `DoFHandler` depend on the type of the registered `Triangulation` and the length of the registered `FECollection`. In my opinion, we should only allow to register both simultaneously.

Right now, we have a zoo of functions (which evolved over time) to register them separately:
```
DoFHandler::DoFHandler()
DoFHandler::DoFHandler(tria)
DoFHandler::set_fes(fes)
DoFHandler::initialize(tria, fes)
DoFHandler::distribute_dofs(fes)
```

I would like to propose a reduced set of functions:
```
DoFHandler::DoFHandler()
DoFHandler::DoFHandler(tria, fes)   [internally calls set(tria, fes)]
DoFHandler::set(tria, fes)
DoFHandler::enumerate_dofs()
```
This would result in one single possibility to change both `Triangulation` and `FECollection` at once via `set(tria, fes)`.

However, this would be a fundamental change to the interface, and nearly every single deal.II user code would need to be adapted. Before starting to work on this patch and to deprecate the old interface, I would like to hear your honest opinion on this. :-)

marcfehling:
While preparing a patch for this issue, I must admit that I am not happy with the current interface.

[...]

PR #11123, opened 2020-10-31
Hide hp_capability_enabled parameter.

marcfehling:
Closes part of #11102 .

Between registering a `Triangulation` and a `FECollection`, we do not know in which mode (hp or not) our `DoFHandler` will be operating. Since hp-mode also includes single element `FECollections`, I decided to let every `DoFHandler` begin in hp-mode, and then toggle it if necessary.

I cleaned up some tests and tutorials on the way which are affected by this change.


marcfehling:
@bangerth @peterrum -- Please have a look and tell me your opinion on this approach

include/deal.II/dofs/dof_handler.h
```diff
@@ -422,7 +422,7 @@ class DoFHandler : public Subscriptor
    * Boolean indicating whether or not the current DoFHander has hp
    * capabilities.
    */
-  const bool hp_capability_enabled;
+  bool hp_capability_enabled;
```
peterrum, line 425: Let's make this private now and add a getter.

peterrum (approved):
Looks good to me. Thanks a lot for the efforts! I'll run the whole test suite locally later today.

Toggling is probably the only way forward. Storing two additional integers per cell in the beginning should not hurt.

source/dofs/dof_handler.cc
```diff
@@ -2453,13 +2441,40 @@ DoFHandler<dim, spacedim>::set_fe(const hp::FECollection<dim, spacedim> &ff)
...
   if (this->fe_collection != ff)
-    this->fe_collection = hp::FECollection<dim, spacedim>(ff);
+    {
+      this->fe_collection = hp::FECollection<dim, spacedim>(ff);
+
+      const bool contains_multiple_fes = (this->fe_collection.size() > 1);
+
+      // toggle the hp-mode of this object only if the type of the registered fe
+      // collection changed
+      if (hp_capability_enabled != contains_multiple_fes)
```
peterrum, line 2451: One question. These lines suggest that we could toggle back from normal to hp mode. But that does not make sense, does it? We already have cleared the tables so that active fe indices couldn't have been set.

marcfehling: Hmm, initially I thought that switching back from non-hp to hp mode would be possible. But you are correct: setting active and future fe indices is no longer possible once we've been in non-hp mode.

I will change this behavior and only allow the transition from hp to non-hp mode. If multiple fes are connected to non-hp mode, I will trigger an `AssertThrow`, so that users will be noticed also in RELEASE mode.

marcfehling:
Some tests complain. I will figure out why.

marcfehling:
Tests should pass now. CI ran out of disk space.
I'll rebase on the master branch to trigger the CI again.

marcfehling:
Part of #10333 :)

kronbichler (approved):

drwells:
@peterrum did the whole test suite pass?

peterrum:
I am re-running it now.

peterrum:
@drwells `ctest -R "matrix_free|hp|dofs"` have passed. Now,  I am running the rest of the test I guess I let you know about the results in the morning CET.

peterrum (approved):
All tests have passed (except of the know problematic ones). I am merging!

Thanks for taking over this task! It is a big step towards making hp a first class citizen of deal.II.

PR #11125, opened 2020-10-31
Update ghost particles without rebuilding them

blaisb:
This is a work in progress ([WIP])
This PR aims at allowing for a function that updates the ghost particles without rebuilding them from scratch. Many numerical methods don't require that the particle interact with the mesh at every time step (e.g. MD based methods or DEM). This PR aims at addressing #11093 .

- [X] Create an update_ghost_particles function and move the necessary containers to particle_handler class member. In the update_ghost_particles remove the loop to generate the ghost_particles_by_domain
- [x] Add a unit test for ghost updates
- [ ] Modify send_rcv_particles (or write a new function alltogether) to enable updating existing particles without recreating them from scratch. This part i will do with the advices @peterrum (thanks again for helping me with this :).

The first modification already implemented allowed us a speed up of 100% in our DEM code on 2 cores (from 280s to 127s) and a 30% (103s to 72s) at 16 core. So this is already extremely significant. I think it's the last optimization we need in our case. This will also make the particle_handler a very very flexible class!

### DIFF
--- a/include/deal.II/particles/particle_handler.h
+++ b/include/deal.II/particles/particle_handler.h
@@ -709,7 +709,10 @@ namespace Particles
 
     /**
      * Update all particles that live in cells that are ghost cells to
-     * other processes.
+     * other processes. In this context, update means to update the
+     * location and the properties of the ghost particles assuming that
+     * the ghost particles have not changed cells. Consquently, this will
+     * not update the reference location of the particles.
      */
     void
     update_ghost_particles();
--- a/include/deal.II/particles/particle_handler.h
+++ b/include/deal.II/particles/particle_handler.h
@@ -708,6 +708,13 @@ namespace Particles
     exchange_ghost_particles();
 
     /**
+     * Update all particles that live in cells that are ghost cells to
+     * other processes.
+     */
+    void
+    update_ghost_particles();
+
+    /**
      * Callback function that should be called before every refinement
      * and when writing checkpoints. This function is used to
      * register store_particles() with the triangulation. This function
@@ -808,6 +815,14 @@ namespace Particles
      * particles are equivalent to the ghost entries in distributed vectors.
      */
     std::multimap<internal::LevelInd, Particle<dim, spacedim>> ghost_particles;
+
+    /**
+     * Set of particles that currently live in the ghost cells of the local
+     * domain, organized by the subdomain_id. These
+     * particles are equivalent to the ghost entries in distributed vectors.
+     */
+    std::map<types::subdomain_id, std::vector<particle_iterator>>
+      ghost_particles_by_domain;
 
     /**
      * This variable stores how many particles are stored globally. It is
--- a/source/particles/particle_handler.cc
+++ b/source/particles/particle_handler.cc
@@ -1233,14 +1233,12 @@ namespace Particles
     const auto parallel_triangulation =
       dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
         &*triangulation);
-    if (parallel_triangulation != nullptr)
+    if (parallel_triangulation == nullptr ||
+        dealii::Utilities::MPI::n_mpi_processes(
+          parallel_triangulation->get_communicator()) == 1)
       {
-        if (dealii::Utilities::MPI::n_mpi_processes(
-              parallel_triangulation->get_communicator()) == 1)
-          return;
+        return;
       }
-    else
-      return;
 
 #ifdef DEAL_II_WITH_MPI
     // First clear the current ghost_particle information
@@ -1707,8 +1705,8 @@ namespace Particles
 
     switch (status)
       {
-          case parallel::distributed::Triangulation<dim,
-                                                    spacedim>::CELL_PERSIST: {
+        case parallel::distributed::Triangulation<dim, spacedim>::CELL_PERSIST:
+          {
             auto position_hint = particles.end();
             for (const auto &particle : loaded_particles_on_cell)
               {
@@ -1727,8 +1725,8 @@ namespace Particles
           }
           break;
 
-          case parallel::distributed::Triangulation<dim,
-                                                    spacedim>::CELL_COARSEN: {
+        case parallel::distributed::Triangulation<dim, spacedim>::CELL_COARSEN:
+          {
             typename std::multimap<internal::LevelInd,
                                    Particle<dim, spacedim>>::iterator
               position_hint = particles.end();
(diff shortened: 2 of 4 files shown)


peterrum:
> The second issue is more profound. At every call, exchange_ghost_particles() clears the entire ghost_particles structure and rebuilds it from scratch, assuming that ghost particles may have changed cells and everything. The underlying assumption is that we are calling sort_particles_into_subdomain_and_cells at every time step. Regretfully, this is not what we are doing.

@blaisb Could you quickly explain me the reason for why you don't have to call `sort_particles_into_subdomain_and_cells` in the case of DEM every time step. Maybe this is a stupid question and I am not aware about the differences of DEM and PIC.

blaisb:
> > The second issue is more profound. At every call, exchange_ghost_particles() clears the entire ghost_particles structure and rebuilds it from scratch, assuming that ghost particles may have changed cells and everything. The underlying assumption is that we are calling sort_particles_into_subdomain_and_cells at every time step. Regretfully, this is not what we are doing.
> 
> @blaisb Could you quickly explain me the reason for why you don't have to call `sort_particles_into_subdomain_and_cells` in the case of DEM every time step. Maybe this is a stupid question and I am not aware about the differences of DEM and PIC.

In DEM  the particle interact between themselves every timestep (we do particle-particle force calculation through either contact models or non-contact models (e.g. an LJ potential)). However, we want the particles to interact with the mesh every **n** iterations, where **n** is a number between 1 and 1000. Consequently, when we calculate particle-particle interaction, we don't need to have the particles exactly sorted into the subdomain (we don't need their reference location). The only thing we need is their properties (velocity, diameter) and their physical location. This is sufficient to do the particle-particle interaction. Thus, we need a mechanism to update the properties and the physical location of the ghost particles.

Right now we are using exchange_ghost to achieve this. But exchange ghost has a very large overhead. Right now, a single core DEM benchmark simulation in Lethe runs in 120s on one core, but runs in 287s on two cores because of the overhead of exchange_ghost_particles. The present PR in it's current state brings the two-core cost to 120s, removing 160s of overhead just because of the loop that is removed when calling update_ghosts every iteration and exchange ghost every **n** iterations.

I know this is usage is not what was originally intended for particles. However, I think providing this update mechanism does not break the code for anyone. It provides a much more flexible implementation that allow a code to do particle-particle stuff at one time scale and particle-mesh stuff at another time scale. Our ultimate goal is to be able to build the first families of DEM scheme where DEM can be coupled with a CFD solver without requiring that we couple two existing code (e.g. OpenFOAM and LIGGGHTS, Fluent and EDEM, etc.). We think this will provide at least a 10x speed-up over the current state of the art just by using existing deal.II facilities and what we added in Lethe.

I hope that clearly explains it :)!.

peterrum:
Thanks for the explanation @blaisb! If I see it correctly you make the simplified assumption that no new ghost-cell particles are created during the n time steps, which is probably acceptable in your case. We can discuss this later offline.

include/deal.II/particles/particle_handler.h
```diff
@@ -707,6 +707,13 @@ namespace Particles
     void
     exchange_ghost_particles();
 
+    /**
+     * Update all particles that live in cells that are ghost cells to
```
peterrum, line 711: Could you be more explicit what *update* means in this context.

source/particles/particle_handler.cc
```diff
@@ -1222,6 +1225,31 @@ namespace Particles
...
   }
 
+  template <int dim, int spacedim>
+  void
+  ParticleHandler<dim, spacedim>::update_ghost_particles()
+  {
+    // Nothing to do in serial computations
+    const auto parallel_triangulation =
+      dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
+        &*triangulation);
```
peterrum, line 1235:
```suggestion
    const auto parallel_triangulation =
      dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
        triangulation->get());
```

source/particles/particle_handler.cc
```diff
@@ -1222,6 +1225,31 @@ namespace Particles
...
+    if (parallel_triangulation != nullptr)
+      {
+        if (dealii::Utilities::MPI::n_mpi_processes(
+              parallel_triangulation->get_communicator()) == 1)
+          return;
+      }
+    else
+      return;
```
peterrum, line 1243:
```suggestion
    if (parallel_triangulation == nullptr || 
        Utilities::MPI::n_mpi_processes(parallel_triangulation->get_communicator()) == 1)
      return;
```

source/particles/particle_handler.cc
```diff
@@ -1679,8 +1707,8 @@ namespace Particles
 
     switch (status)
       {
-        case parallel::distributed::Triangulation<dim, spacedim>::CELL_PERSIST:
-          {
+          case parallel::distributed::Triangulation<dim,
+                                                    spacedim>::CELL_PERSIST: {
```
peterrum, line 1711: I guess this is a formatting issue.

source/particles/particle_handler.cc
```diff
@@ -1222,6 +1225,31 @@ namespace Particles
...
+    if (parallel_triangulation != nullptr)
+      {
+        if (dealii::Utilities::MPI::n_mpi_processes(
+              parallel_triangulation->get_communicator()) == 1)
+          return;
+      }
+    else
+      return;
+
+#ifdef DEAL_II_WITH_MPI
```
peterrum, line 1245: Is this guard actually needed?

peterrum (commented):
Looks good to me. Once you have added a test, I think this PR is ready to be merged.

The point:
> Modify send_rcv_particles (or write a new function alltogether) to enable updating existing particles without recreating them from scratch.

could be tackled in a follow-up PR. What do you think?

source/particles/particle_handler.cc
blaisb: This does not seem to work. No member get in dealii::Triangulation<1,1> ?

peterrum: Oh, triangulation is *SmartPoint*. Just ignore my comments ;)

blaisb:
This last push finishes task 2 of this PR.
I added a unit test which has a property and a position, which is changed and the ghosts are updated. It is simple, but it really does the job. I also applied @peterrum  comments and fixed indentation.

peterrum (approved):
Thanks!

blaisb:
> Thanks!

No problem. If it is good for you, I suggest we go as you suggested. We could merge this PR and then together work on the third (and last) item. :)!

peterrum:
> third (and last) item. :)!

Third but definitely not the last item ;)

kronbichler (approved):

PR #11128, opened 2020-11-01
ADHelpers: Fix indexing for symm tensor extractor

jppelteret:


luca-heltai (approved):

jppelteret:
Thanks for the review, @luca-heltai!

PR #11131, opened 2020-11-03
Transfinite interpolation: add a test

tamiko:
See also: #11126


tamiko:
@kronbichler *ping*

tests/manifold/transfinite_manifold_11.cc
```diff
@@ -0,0 +1,155 @@
...
+   *   -0.1271 -0.0177875 -> 0.211251 1.01047
+   *   -0.1271 -0.015564 -> 0.214091 1.01181
+   *   Looking at cell 2_0: with vertices:
+   *   -0.2 0.025    2.5 0.416667    1.25 2.16506    2.5 1.25
+   *   Transformation to chart coordinates:
+   *   -0.1271 -0.0177875 -> 0.253794 -0.0323121
+   *   -0.1271 -0.015564 -> 0.251301 -0.0309146
+   *
+   * The following is a simplified version that fails with a sligthly
+   * different error message.
```
tamiko, line 142: @kronbichler This took a while to extract. I have simplified the test from globally refining 8 times to a single `get_new_point` invocation. The manifold then fails with a slightly different errror path. In any case, #11131 fixes the problem.

kronbichler: Yes, this looks exactly as I would have done it, nice test!

kronbichler (approved):
This is good, but it needs to be rebased to take the freshly merged #11126 into account.

tests/manifold/transfinite_manifold_11.cc
```diff
@@ -0,0 +1,155 @@
...
+   */
+
+  std::vector<Point<2>> points{Point<2>{-0.1271, -0.0177875},
+                               Point<2>{-0.1271, -0.015564}};
+  ArrayView<Point<2>>   points_view{points.data(), 2};
+
+  std::vector<double> weights{0.5, 0.5};
+  ArrayView<double>   weights_view{weights.data(), 2};
+
+  transfinite.get_new_point(points_view, weights_view);
```
kronbichler, line 152: An idea on a second look: Do you want to print the new point to the log file, just in case we ever come to the decision of not throwing upon errors in locating the point?

tamiko:
Oops. Garbled up the pull request this depends on.

tests/manifold/transfinite_manifold_11.cc
tamiko: @kronbichler done!

kronbichler (approved):

tests/manifold/transfinite_manifold_11.cc
```diff
@@ -0,0 +1,155 @@
...
+   */
+
+  std::vector<Point<2>> points{Point<2>{-0.1271, -0.0177875},
+                               Point<2>{-0.1271, -0.015564}};
+  ArrayView<Point<2>>   points_view{points.data(), 2};
+
+  std::vector<double> weights{0.5, 0.5};
+  ArrayView<double>   weights_view{weights.data(), 2};
+
+  deallog << transfinite.get_new_point(points_view, weights_view) << std::endl;
```
kronbichler, line 152: I think you need 
```suggestion
  deallog << triangulation.get_manifold(3).get_new_point(points_view, weights_view) << std::endl;
```
to get this past https://github.com/dealii/dealii/blob/ce4b67869c2f44195290a12976326d3a6debd11f/source/grid/manifold_lib.cc#L2257-L2259

tamiko:
@kronbichler You're right. That happens when you test code with a slightly modified deal.II library...

tamiko:
Finally.

PR #11132, opened 2020-11-03
[WIP] Enable element activation distributed::SolutionTransfer

Rombur:
I am trying to extent `distributed::SolutionTransfer` but I am stuck and I would appreciate any pointer (@marcfehling ?). Here is what I want to achieve, I want to be able to activate cell over time, i.e., I want to be able to do `p-refinement` where a cell changes from `FE_Nothing` to `FE_Q`. After activating new cells, we end up with new dofs whose values cannot be determined from the interpolation from the previous mesh. Ideally, I would set the values in the vector before `interpolate` and `interpolate` would not touch these values since it doesn't know how to treat them anyway. Currently if `FE_Nothing` is attached to a cell, the cell is skipped in the `pack_data` and `unpack_data` functions. This doesn't allow for a `FE_Nothing` cell to change to an active cell (non-`FE_Nothing` cell). To fix that I have created a new variable `past_fe_index` similar to `future_fe_index`. This record the `fe_index` before refinement and this allows me to skip cells that were `FE_Nothing` but that have been activated. This works in serial and does what I want. However, a problem appears when using MPI. Let's assume that you have two processors and on one processor all the cells are `FE_Nothing`. This processor won't do anything while the other correctly interpolates the values. Everything is fine until we call `compress(VectorOperation::insert)`. Now the locally owned value and the ghost value don't match and the code crashes. Note that in the case described here, the dofs can migrate from one processor to the other and so the ghost value has the correct value and the locally owned value is wrong. I could try to introduce a new version of `compress` that does the right thing but I was hoping that someone had a better solution.

### DIFF
--- a/include/deal.II/dofs/dof_accessor.h
+++ b/include/deal.II/dofs/dof_accessor.h
@@ -2060,6 +2060,27 @@ public:
    */
   unsigned int
   dominated_future_fe_on_children() const;
+
+  /**
+   * Return the fe_index of the finite element that was assigned to this
+   * cell before the triangulation gets refined and coarsened.
+   */
+  unsigned int
+  past_fe_index() const;
+
+  /**
+   * Set the fe_index of the finite element that was assigned to this
+   * cell next time the triangulation gets refined and coarsened. A previously
+   * assigned past finite element will be overwritten.
+   */
+  void
+  set_past_fe_index(const unsigned int i) const;
+
+  /**
+   * Return whether a past finite element has been set.
+   */
+  bool
+  past_fe_index_set() const;
   /**
    * @}
    */
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -2370,6 +2370,98 @@ namespace internal
                                      [accessor.present_index] =
           invalid_active_fe_index;
       }
+
+
+
+      /**
+       * Do what the past_fe_index function in the parent class is supposed to
+       * do.
+       */
+      template <int dim, int spacedim, bool level_dof_access>
+      static unsigned int
+      past_fe_index(
+        const DoFCellAccessor<dim, spacedim, level_dof_access> &accessor)
+      {
+        if (accessor.dof_handler->hp_capability_enabled == false)
+          return DoFHandler<dim, spacedim>::
+            default_fe_index; // ::DoFHandler only supports
+                              // a single active fe with
+                              // index zero
+
+        Assert(
+          accessor.dof_handler != nullptr,
+          (typename std::decay<decltype(accessor)>::type::ExcInvalidObject()));
+        Assert(static_cast<unsigned int>(accessor.level()) <
+                 accessor.dof_handler->hp_cell_past_fe_indices.size(),
+               ExcMessage("DoFHandler not initialized"));
+
+        if (past_fe_index_set(accessor))
+          return accessor.dof_handler
+            ->hp_cell_past_fe_indices[accessor.level()][accessor.present_index];
+        else
+          return accessor.dof_handler
+            ->hp_cell_active_fe_indices[accessor.level()]
+                                       [accessor.present_index];
+      }
+
+
+
+      /**
+       * Do what the set_past_fe_index function in the parent class is
+       * supposed to do.
+       */
+      template <int dim, int spacedim, bool level_dof_access>
+      static void
+      set_past_fe_index(
+        const DoFCellAccessor<dim, spacedim, level_dof_access> &accessor,
+        const unsigned int                                      i)
+      {
+        if (accessor.dof_handler->hp_capability_enabled == false)
+          {
+            // ::DoFHandler only supports a single active fe with index zero
+            AssertDimension(i, (DoFHandler<dim, spacedim>::default_fe_index));
+            return;
+          }
+
+        Assert(
+          accessor.dof_handler != nullptr,
+          (typename std::decay<decltype(accessor)>::type::ExcInvalidObject()));
+        Assert(static_cast<unsigned int>(accessor.level()) <
+                 accessor.dof_handler->hp_cell_past_fe_indices.size(),
+               ExcMessage("DoFHandler not initialized"));
+
+        accessor.dof_handler
+          ->hp_cell_past_fe_indices[accessor.level()][accessor.present_index] =
+          i;
+      }
+
+
+
+      /**
+       * Do what the past_fe_index_set function in the parent class is
+       * supposed to do.
+       */
+      template <int dim, int spacedim, bool level_dof_access>
+      static bool
+      past_fe_index_set(
+        const DoFCellAccessor<dim, spacedim, level_dof_access> &accessor)
+      {
+        if (accessor.dof_handler->hp_capability_enabled == false)
+          return false; // ::DoFHandler only supports a single active fe with
+                        // index zero
+
+        Assert(
+          accessor.dof_handler != nullptr,
+          (typename std::decay<decltype(accessor)>::type::ExcInvalidObject()));
+        Assert(static_cast<unsigned int>(accessor.level()) <
+                 accessor.dof_handler->hp_cell_past_fe_indices.size(),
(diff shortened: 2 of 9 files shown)


marcfehling:
As far as I know `SolutionTransfer` is designed for transferring dof-related data, and you are looking for a way to transfer general data associated with active cells. We have a class `CellDataTransfer` for both non-parallel and parallel cases which could probably help you out with the problem. Before hp-adaptation, you can associate any kind of data with each active cell independent of the finite element and transfer it across meshes, even if `FE_Nothing` is associated with them.

Similar to `SolutionTransfer`, you could store the dof values on non-`FE_Nothing` elements in `Vectors` with `get_interpolated_dof_values()`, and do on `FE_Nothing` elements whatever you like. And after the transfer you can restore them with `set_interpolated_dof_values()`. You would have more control over the transferred data in this case than `SolutionTransfer` gives you, which would probably get you around using `past_fe_indices`.

---

If you want to stick to using `past_fe_indices` in parallel, you also need to transfer those between all processors.

I had that idea back then as well, but decided to let go of it since this would require to transfer one more integer per cell. The `future_fe_indices` were just introduced to prepare data on the old mesh for the new one, and then to transfer them to the new mesh as `active_fe_indices`. Hence, `future_fe_indices` only live on locally owned cells. Only `active_fe_indices` are communicated to ghost cells and are thus available on locally relevant cells.

---

Right now, you skip newly activated cells (`FE_Nothing` -> `FE_Q`) in `interpolate`, thus no values on the degrees of freedom on this cell will be set. On shared vertices with neighboring cells however, there may be degrees of freedom that belong to an already activated cell and thus have the correct value assigned. Ultimately there are multiple values for one degree of freedom, and my guess is that this is what `compress` complains about.

A similar race condition will probably happen in your serial implementation as well, but here you wouldn't be informed with an error as values will probably just be overwritten. Be careful!

In parallel applications, you would need to intervene somehow between `notify_ready_to_unpack` and `compress` in `p::d::SolutionTransfer::interpolate()`, but you won't achieve that with a stock `SolutionTransfer` class. I would encourage you to try `CellDataTransfer` as it gives you more flexibility.

---

One last note on hp-adaptation and `FENothing`: `DoFHandler::distribute_dofs()` should enumerate all degrees of freedom on locally owned and ghost cells correctly, but we never tested for a case in which the total number of locally owned degrees of freedom on one processor is actually zero. It could be possible that something goes wrong here, but I guess it's the race condition from above.

marcfehling:
You could also perform just h-adaptation as a first step and transfer the solution with `SolutionTransfer`, and then set the `active_fe_indices` manually afterwards. You could update your solution with the `get_interpolated_dof_values` and `set_dof_values_by_interpolation` function. If desired, you can repartition your `Triangulation` afterwards.

You do not necessarily need to use `execute_coarsening_and_refinement()` for p-adaptation.

Rombur:
> You could also perform just h-adaptation as a first step and transfer the solution with `SolutionTransfer`, and then set the `active_fe_indices` manually afterwards. You could update your solution with the `get_interpolated_dof_values` and `set_dof_values_by_interpolation` functions manually. If desired, you can repartition your `Triangulation` afterwards.
> 
> You do not necessarily need to use `execute_coarsening_and_refinement()` for p-adaptation.

I am not sure I understand how it works. I perform the h-adaptation and transfer the solution. Then I use `get_interpolated_dof_values`, update the fe indices, distribute the dofs, and finally I use `set_dof_values_by_interpolation`. I guess if I don't want to have a problem during `compress` then I cannot use a ghosted vector and I need to "filter" the dofs to eliminate the `set_dof_values_by_interpolation`. When I distribute the dofs, isn't it possible that some dofs move from one processor to the other? This isn't like a regular p-refinement. Since we have a "new" cell, won't the dof_handler redistribute the dofs from one processor to another one?

marcfehling:
> I am not sure I understand how it works. I perform the h-adaptation and transfer the solution. Then I use `get_interpolated_dof_values`, update the fe indices, distribute the dofs, and finally I use `set_dof_values_by_interpolation`.

Yes, the workflow you pointed out corresponds with my idea.

> I guess if I don't want to have a problem during `compress` then I cannot use a ghosted vector and I need to "filter" the dofs to eliminate the `set_dof_values_by_interpolation`. When I distribute the dofs, isn't it possible that some dofs move from one processor to the other? This isn't like a regular p-refinement. Since we have a "new" cell, won't the dof_handler redistribute the dofs from one processor to another one?

The function `get_interpolated_dof_values` takes the global solution vector and extracts the dof values of the present cell, while `set_dof_values_by_interpolation` distributes the dof values on the global solution vector. Calling `distribute_dofs` enumerates all dofs globally, and each _locally relevant_ cell knows which dofs belong to it.

So for the manual p-adaptation, you should call `get_interpolated_dof_values` on all _locally owned_ cells of the solution transferred with `SolutionTransfer`, and then after setting the new active fe indices call `set_dof_values_by_interpolation` on all _locally owned_ cells on a non-ghosted vector. Finally call `compress` to distribute all information in the global MPI universe.

If you need access to the data on ghost cells, create a ghosted vector afterwards. As far as I know, you cannot use `compress` on ghosted vectors (see [glossary](https://www.dealii.org/current/doxygen/deal.II/DEALGlossary.html#GlossCompress)).

You would need to provide some kind of data on your newly activated cells though that matches the neighboring cells (if working with continuous elements). You should be able to look those up in the solution transferred with `SolutionTransfer`. If a neighboring cell is a ghost cell, you need to turn it into a ghosted vector first.

This all sounds very complicated, and I'm just giving you hypothetical advice that I didn't try out myself yet. If we come up with a working example, we may want to turn this construct into a whole new class. If I'll find time in the next weeks, I'm going to work on an example. But feel free to continue experimenting :)

marcfehling:
Or use the `CellDataTransfer` class. On already activated cells, you can basically imitate what SolutionTransfer is doing. On cells to be activated, you attach data for to the new cell already on the old mesh before calling `execute_coarsening_and_refinement()`.

You would need to convert the so transferred cell-by-cell data into a global solution. You could do this by calling `set_dof_values_by_interpolation` for every _locally owned_ cell, then call `compress` and so on...

My guess is that this approach is more performant as less communication is involved.

Rombur:
@marcfehling Thanks for the explanation. I'll give it a shot.

marcfehling:
@Rombur Let me know if I can be of further assistance :)

Rombur:
@marcfehling It seems that I got this working. I have used `CellDataTransfer` with `get_interpolated_dof_values` and `set_dof_values_by_interpolation`. The only tricky part is when using MPI and you activate one cell on the boundary between two processors. If you simply use `set_dof_values_by_interpolation` and then `compress(VectorOperation::insert)`, you get the following error:
```
 Called compress(VectorOperation::insert), but the element received from a remote processor, value 1,
 does not match with the value 0 on the owner processor 0
```
You may be able to use `compress(VectorOperation::add)` instead but then you lose the information on which dofs are newly activated and which dofs are simply zero. Instead I initialize the vector that I will use for the interpolation with `std::numeric_limits<double>::infinity()`, then I use `compress(VectorOperation::min)`. This way the new dofs have a value of infinity and all the other dofs have the correct value. Then I can go back and set the value I want for the new dofs.

Thanks a lot for your help!

marcfehling:
> Instead I initialize the vector that I will use for the interpolation with `std::numeric_limits<double>::infinity()`, then I use `compress(VectorOperation::min)`. This way the new dofs have a value of infinity and all the other dofs have the correct value. Then I can go back and set the value I want for the new dofs.

That is actually a very smart workaround for your scenario. I like your approach! If you are sure that the values can be either `infinity` or the actual correct one, you are fine. Glad that this worked for you!

Do you plan on using grid adaptive methods as well? The `CellDataTransfer` is capable of grid adaptation and gives you a general interface for adaptation strategies (e.g. how values will be transferred from a parent cell to child cells in case of refinement). I'm curious if that would work in your context, too!

Rombur:
> Do you plan on using grid adaptive methods as well? 

Yes, I will. I expect the current code to work but I'll update this thread if it doesn't.

PR #11133, opened 2020-11-03
Rework DoFHandler connections to underlying Triangulation.

marcfehling:
~~Blocked by #11123~~ (EDIT: PR has been merged). Part of #11102.

This PR deals with adjustments to the interface to the Triangulation signals and cleans them up.

As we now initialize the fe index tables right after construction and allow users to set fe indices to their liking, they need to be transferred without a registered collection. We check if the fe indices are valid when calling distribute_dofs.

### DIFF
--- a/include/deal.II/dofs/dof_handler.h
+++ b/include/deal.II/dofs/dof_handler.h
@@ -1587,35 +1587,33 @@ private:
   connect_to_triangulation_signals();
 
   /**
-   * Create default tables for the active_fe_indices.
-   * They are initialized with a zero
-   * indicator, meaning that fe[0] is going to be used by default. This
-   * method is called before refinement and while setting the finite elements
-   * via set_fe(). It ensures each cell has a valid active_fe_index.
+   * Create default tables for the active and future fe_indices.
+   *
+   * Active indices are initialized with a zero indicator, meaning that fe[0] is
+   * going to be used by default. Future indices are initialized with an invalid
+   * indicator, meaning that no p-adaptation is scheduled by default.
+   *
+   * This method is called upon construction and whenever the underlying
+   * triangulation gets created. This ensures that each cell has a valid active
+   * and future fe_index.
    */
   void
   create_active_fe_table();
 
   /**
-   * A function that will be triggered through a triangulation
-   * signal just before the triangulation is modified.
+   * Update tables for active and future fe_indices.
    *
-   * The function that stores the active_fe_flags of all cells that will
-   * be refined or coarsened before the refinement happens, so that
-   * they can be set again after refinement.
+   * Whenever the underlying triangulation changes (either by adaptation or
+   * deserialization), active and future fe index tables will be adjusted to the
+   * current structure of the triangulation. Missing values of active and future
+   * indices will be initialized with their defaults (see
+   * create_active_fe_table()).
+   *
+   * This method is called post refinement and post deserialization. This
+   * ensures that each cell has a valid active and future fe_index.
    */
   void
-  pre_refinement_action();
-
-  /**
-   * A function that will be triggered through a triangulation
-   * signal just after the triangulation is modified.
-   *
-   * The function that restores the active_fe_flags of all cells that
-   * were refined.
-   */
-  void
-  post_refinement_action();
+  update_active_fe_table();
 
   /**
    * A function that will be triggered through a triangulation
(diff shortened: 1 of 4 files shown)


marcfehling:
Part of #10333 :)

peterrum (commented):
I like the proposed changes. I have a second look at the code once the `hp_capabilities_enabled`-PR is merged.

peterrum:
@marcfehling  Could you rebase this PR. Then I will take a second look!

tjhei:
I did not know that some of these even existed! I am okay with the change, but I wonder if anybody makes use of the initialize() call.

source/dofs/dof_handler.cc
```diff
@@ -3226,27 +3280,21 @@ DoFHandler<dim, spacedim>::post_refinement_action()
 
 template <int dim, int spacedim>
 void
-DoFHandler<dim, spacedim>::pre_active_fe_index_transfer()
+DoFHandler<dim, spacedim>::pre_transfer_action()
 {
-  // Finite elements need to be assigned to each cell by calling
-  // distribute_dofs() first to make this functionality available.
-  if (this->fe_collection.size() > 0)
-    {
```
peterrum, line 3234: Why can this if-statement go?

peterrum (approved):
To me the changes look fine: some functions have been deprecated, some (internal) ones have been renamed, a couple of if-statements have been removed so that the indentation changed leading to many *changed* lines.

Thanks you for going through all the tests! However, I was wondering if we shouldn't keep one or two which actually uses the hp-DofHandler dummy class + iterator + ... (or add a new test)? This way we at least can be somewhat sure that we a backwards compatible. One reason why #10328 was merged (except we started to get angry unicorns) was that no test had to be changed.

source/dofs/dof_handler.cc
marcfehling: Initially, fe indices should only be transferred when a corresponding `FECollection` has been registered.

As we now initialize the fe index tables right after construction and allow users to set fe indices to their liking, they need to be transferred without a registered collection. We check if the fe indices are valid when calling `distribute_dofs`.

marcfehling:
I can split this PR into two if you like. We can merge the actual changes to the library (which should not require any tests to be changed), and leave a WIP PR open with the intent to avoid deprecated functions. We can extend that one every once in a while and merge it before the next release?

peterrum:
> I did not know that some of these even existed! I am okay with the change, but I wonder if anybody makes use of the initialize() call.

Yes, I know code which uses for example the `initialize()` function. It does not help that the tutorials used to show that approach in the adaptive case ;)

peterrum:
> I can split this PR into two if you like. 

Splitting up the PR in two is a good idea.

> We can merge the actual changes to the library (which should not require any tests to be changed), and leave a WIP PR open with the intent to avoid deprecated functions. We can extend that one every once in a while and merge it before the next release?

But we should get both merged soon! I know how annoying is to adjust test and I don't see the point in waiting with updating the tests (there is no way we go back in time and split up the DoFHandlers again).

peterrum:
... and here is another instance: https://github.com/kronbichler/advection_miniapp/blob/e443c39909d44edee912f1746bb18a2c9af6e011/advection_solver.cc#L1130

marcfehling:
Moved the deprecation part to #11153.

marcfehling:
So are you pro (:+1:) or con (:-1:) keeping the `initialize` function?

marcfehling:
One of the examples used the now deprecated initialize function. I included the changes to the library and to the examples in this PR, so that the test suite completes successfully.

Changes to all tests are part of the follow-up PR.

include/deal.II/dofs/dof_handler.h
```diff
@@ -449,7 +449,10 @@ class DoFHandler : public Subscriptor
    * Standard constructor, not initializing any data. After constructing an
    * object with this constructor, use initialize() to make a valid
    * DoFHandler.
+   *
+   * @deprecated Use the constructor which expects a Triangulation instead.
    */
+  DEAL_II_DEPRECATED
   DoFHandler();
```
kronbichler, line 456: I am fine with this change, but I remark that this can possibly affect a considerable number of user projects. @tamiko I have seen this kind of constructor used often in some of your codes, can you have a look here?

kronbichler (approved):
I would be fine with the proposed changes and the deprecation of `initialize`, but I think we need a broader set of opinions on the latter topic.

peterrum:
@marcfehling May I ask you to remove the deprecation pragmas for now. Let's get this PR merged; be can the pragmas back later.

marcfehling:
I'll move them to the draft PR then...
I will use the draft as a sandbox for changes originating from merging both `DoFHandlers`. I will cherry-pick things out of there that can be merged one by one.

marcfehling:
Deprecation of the interface has been moved to #11160. Should have done that in the first place - sorry for the inconvenience.

Only the general review of the `DoFHandler` class is left in this PR. Ready for merge, as CI has already passed!

kronbichler (approved):

PR #11134, opened 2020-11-03
Expose mesh loop bug for anisotropic grids.

luca-heltai:
`mesh_loop` does not know how to deal with anisotropic grids.

The first commit exposes the bug. We are working on a fix.


nicola-giuliani:
bug_2d

luca-heltai:
(mesh loop assumes that cells on different levels must be refined differently, and does not ask the proper function)

nicola-giuliani:
bug_3d

kronbichler (approved):

PR #11140, opened 2020-11-03
Store reference_cell_types in Triangulation

peterrum:


include/deal.II/grid/tria.h
```diff
@@ -3359,6 +3359,18 @@ class Triangulation : public Subscriptor
...
+   * TODO
+   */
+  const std::vector<ReferenceCell::Type> &
+  get_reference_cell_types() const;
+
+  /**
+   * TODO
+   */
+  bool
+  all_reference_cell_types_are_hyper_cube() const;
```
peterrum, line 3372: @jppelteret What do you think about this function? Is this what you had in mind at our last chat?

source/grid/tria.cc
```diff
@@ -12771,6 +12771,39 @@ Triangulation<dim, spacedim>::update_periodic_face_map()
 }
 
 
+template <int dim, int spacedim>
```
kronbichler, line 12774: Can you use 3 empty lines between the functions here and below?

peterrum: I can do that. But 2 lines are the norm in tria.cc, which is why I did not use 3 lines.

kronbichler: I saw both in the file with maybe a slight tendency to 2 lines, but 3 lines are also quite common; the reason why I mentioned it is that the place where you inserted the code used to have 3 lines of empty space :wink:

peterrum: Done!

kronbichler (approved):

drwells:
I find using a vector here confusing - when I first read this I thought we could index `reference_cell_types` in a meaningful way. Is there an advantage to using a vector over just using a set?

include/deal.II/grid/tria.h
jppelteret: Yes, precisely @peterrum. Thanks for adding it! I would want to use it for much the same purposes as you've done [here](https://github.com/dealii/dealii/pull/11140/files#diff-a3f9b6c7a51a53e5c8dc202a4dc8b8ab87f574023fb5f04dd3d2d87ecc62d68bR9814-R9818).

peterrum: Thanks for the feedback! It took some time until I had time to work in this issue ;)

drwells:
@peterrum pointed out to me that we already store manifold and boundary ids in this way so this approach is best.

PR #11141, opened 2020-11-04
Add internal::TriangulationImplementation::ImplementationMixedMesh

peterrum:
... for now it is empty (and only asserts - what is better then what we right now). The class will be filled step by step.

### DIFF
--- a/source/grid/tria.cc
+++ b/source/grid/tria.cc
@@ -9354,6 +9354,68 @@ namespace internal
     };
 
 
+    /**
+     * Same as above but for mixed meshes (and simplex meshes).
+     */
+    struct ImplementationMixedMesh
+    {
+      template <int dim, int spacedim>
+      static void
+      delete_children(
+        Triangulation<dim, spacedim> &                        triangulation,
+        typename Triangulation<dim, spacedim>::cell_iterator &cell,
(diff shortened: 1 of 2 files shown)


source/grid/tria.cc
```diff
@@ -9823,6 +9879,19 @@ Triangulation<dim, spacedim>::create_triangulation(
       throw;
     }
 
+  if (true)
```
peterrum, line 9882: Here I would like to use `all_reference_cell_types_are_hyper_cube()` from PR #11140.

kronbichler (approved):

source/grid/tria.cc
```diff
@@ -9810,19 +9866,34 @@ Triangulation<dim, spacedim>::create_triangulation(
...
-               "A cell with invalid number of vertices has been provided."));
-#endif
     }
   catch (...)
     {
       clear_despite_subscriptions();
       throw;
     }
 
+  if (this->all_reference_cell_types_are_hyper_cube())
```
peterrum, line 9876: I have rebased. Now `all_reference_cell_types_are_hyper_cube()` is actually in use :)

kronbichler (approved):

drwells (approved):

PR #11145, opened 2020-11-04
Adjusted mg_constrained_dofs for usage of own IndexSets

mathsen:
Hello,
I needed to setup a numerical example, where I want to impose a zero mean value over my domain boundary in parallel.
For this every processor has to setup an IndexSet with the single DoF that shall be constrained and add all the (scaled) locally owned indices, that have support on the boundary.
This setup works as expected, but to get it to work with the MultiGrid classes I had to make small adjustments to the MGConstrainedDoFs class:
The internal level_constraints (and user_constraints) have to be initialized with the new IndexSet on each level (which ist just for each process the locally relevant DoFs + the one DoF which shall be constrained) and I just overloaded the reinit() function to accomplish that.
With this small addition I am able to simply setup the MultiGrid structure similar as to the "normal active cells" setup (thanks to [pull request 10348](https://github.com/dealii/dealii/pull/10348)) and maybe it's worth being merged into master?

### DIFF
--- a/include/deal.II/multigrid/mg_constrained_dofs.h
+++ b/include/deal.II/multigrid/mg_constrained_dofs.h
@@ -65,21 +65,18 @@ public:
    * not support rotation matrices in the periodicity definition, i.e., the
    * respective argument in the GridTools::collect_periodic_faces() may not
    * be different from the identity matrix.
-   */
-  template <int dim, int spacedim>
-  void
-  initialize(const DoFHandler<dim, spacedim> &dof);
-
-  /**
-   * Same as the function above, but here the internal AffineConstraints objects
-   * on each level are initialized with the argument level_relevant_dofs.
-   * This should define a superset of locally relevant DoFs and allows the user
-   * to add additional indices to the set of constrained DoFs.
+   * If no level_relevant_dofs are passed as second argument the function uses
+   * the locally relevant level DoFs, extracted by
+   * DoFTools::extract_locally_relevant_level_dofs(). Otherwise the user
+   * provided IndexSets are used on each level, which should define a superset
+   * of locally relevant DoFs and allows the user to add additional indices to
+   * the set of constrained DoFs.
    */
   template <int dim, int spacedim>
   void
   initialize(const DoFHandler<dim, spacedim> &dof,
-             const MGLevelObject<IndexSet> &  level_relevant_dofs);
+             const MGLevelObject<IndexSet> &  level_relevant_dofs =
+               MGLevelObject<IndexSet>());
 
   /**
    * Fill the internal data structures with information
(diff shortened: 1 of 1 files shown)


masterleinad:
Since we have `add_user_constraints()`, allowing to provide a custom IndexSet sounds sensible to me. However, you don't need to duplicate the implementation. The two `initialize` functions just differ by the IndexSet to be used and the existing one can easily be implemented in terms of the one you are proposing here.

include/deal.II/multigrid/mg_constrained_dofs.h
```diff
@@ -309,6 +321,94 @@ MGConstrainedDoFs::initialize(const DoFHandler<dim, spacedim> &dof)
 }
 
 
+template <int dim, int spacedim>
+inline void
+MGConstrainedDoFs::initialize(
+  const DoFHandler<dim, spacedim> &dof,
+  const MGLevelObject<IndexSet> &  level_relevant_dofs)
```
kronbichler, line 328: Can we make the other function create an `IndexSet` and then let it use the code here to avoid duplicating 85 lines of code?

mathsen: @masterleinad @kronbichler You are both obviously absolutely right, the code duplication is really nasty -the cause for this is that I normally don't like default arguments for reference parameters and didn't want to introduce smart pointers or something like std::optional.
Nevertheless I have changed the code and unified everything to one function with a second default argument, which basically checks by a call to max_level() whether a second parameter was passed by the caller or the default one has been constructed, which avoids code duplication.

include/deal.II/multigrid/mg_constrained_dofs.h
```diff
@@ -64,10 +65,18 @@ class MGConstrainedDoFs : public Subscriptor
    * not support rotation matrices in the periodicity definition, i.e., the
    * respective argument in the GridTools::collect_periodic_faces() may not
    * be different from the identity matrix.
+   * If no level_relevant_dofs are passed as second argument the function uses
+   * the locally relevant level DoFs, extracted by
+   * DoFTools::extract_locally_relevant_level_dofs(). Otherwise the user
+   * provided IndexSets are used on each level, which should define a superset
+   * of locally relevant DoFs and allows the user to add additional indices to
+   * the set of constrained DoFs.
```
masterleinad, line 73:
```suggestion
   * If no level_relevant_dofs are passed as the second argument, the function uses
   * the locally relevant level DoFs, extracted by
   * DoFTools::extract_locally_relevant_level_dofs(). Otherwise, the user-provided 
   * IndexSets, which should define a superset
   * of locally relevant DoFs, are used on each level to allows the user to add 
   * additional indices to the set of constrained DoFs.
```

masterleinad:
Otherwise, this looks good to me.

include/deal.II/multigrid/mg_constrained_dofs.h
kronbichler: You also have added a second function and let the implementation of the old function use the new one, but this is also nice.

include/deal.II/multigrid/mg_constrained_dofs.h
```diff
@@ -309,7 +333,6 @@ MGConstrainedDoFs::initialize(const DoFHandler<dim, spacedim> &dof)
 }
 
 
-
```
kronbichler, line 312: Can you add back this line?

kronbichler (approved):
Looks good to me apart from @masterleinad's comment.

include/deal.II/multigrid/mg_constrained_dofs.h
mathsen: Thanks for the review! I commited your changes ("make indent" just slightly adjusted the line wrapping).

include/deal.II/multigrid/mg_constrained_dofs.h
mathsen: Thanks for the review. I added the line again in the last commit.

tjhei (approved):
Awesome, this looks good. Thank you!

masterleinad (approved):
Fine with me.

kronbichler (approved):

mathsen:
Hello,
I guess the CI failed build has nothing to do with this commit, the relevant log line should be:

`1629: /tmp/ccPhHvMR.s: Fatal error: can't close CMakeFiles/spherical_manifold_02.debug.dir/spherical_manifold_02.cc.o: No space left on device
`
Maybe someone can free some space on the hard disks? :)

PR #11149, opened 2020-11-05
Clean up distribute_dofs().

marcfehling:
~~Blocked by #11133.~~ ~~(EDIT: Now blocked by #11160)~~ (EDIT: Dependencies have been merged.)

Part of #11102 and #10333.

This is an attempt to clean up the `distribute_dofs()` implementation internally. Right now, it contains both hp and non-hp implementations individually, and we distinguish between both with a conditional. As they are conceptually related, I tried to combine those in this PR.

@bangerth -- FYI

### DIFF
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -251,6 +251,31 @@ namespace internal
       }
 
       /**
+       * Do that part of reserving space that pertains to releasing
+       * the previously used memory.
+       */
+      template <int dim, int spacedim>
+      static void
+      reset_to_empty_objects(DoFHandler<dim, spacedim> &dof_handler)
+      {
+        dof_handler.object_dof_indices.clear();
+        dof_handler.object_dof_indices.resize(dof_handler.tria->n_levels());
+        dof_handler.object_dof_indices.shrink_to_fit();
+
+        dof_handler.object_dof_ptr.clear();
+        dof_handler.object_dof_ptr.resize(dof_handler.tria->n_levels());
+        dof_handler.object_dof_ptr.shrink_to_fit();
+
+        dof_handler.cell_dof_cache_indices.clear();
+        dof_handler.cell_dof_cache_indices.resize(dof_handler.tria->n_levels());
+        dof_handler.cell_dof_cache_indices.shrink_to_fit();
+
+        dof_handler.cell_dof_cache_ptr.clear();
+        dof_handler.cell_dof_cache_ptr.resize(dof_handler.tria->n_levels());
+        dof_handler.cell_dof_cache_ptr.shrink_to_fit();
+      }
+
+      /**
        * Reserve enough space in the <tt>levels[]</tt> objects to store the
        * numbers of the degrees of freedom needed for the given element. The
        * given element is that one which was selected when calling
@@ -259,6 +284,8 @@ namespace internal
       template <int spacedim>
       static void reserve_space(DoFHandler<1, spacedim> &dof_handler)
       {
+        reset_to_empty_objects(dof_handler);
+
         dof_handler.object_dof_indices[0][0].resize(
           dof_handler.tria->n_vertices() *
             dof_handler.get_fe().n_dofs_per_vertex(),
@@ -300,6 +327,8 @@ namespace internal
       template <int spacedim>
       static void reserve_space(DoFHandler<2, spacedim> &dof_handler)
       {
+        reset_to_empty_objects(dof_handler);
+
         dof_handler.object_dof_indices[0][0].resize(
           dof_handler.tria->n_vertices() *
             dof_handler.get_fe().n_dofs_per_vertex(),
@@ -360,6 +389,8 @@ namespace internal
       static void reserve_space(DoFHandler<3, spacedim> &dof_handler)
       {
         const unsigned int dim = 3;
+
+        reset_to_empty_objects(dof_handler);
 
         dof_handler.object_dof_indices[0][0].resize(
           dof_handler.tria->n_vertices() *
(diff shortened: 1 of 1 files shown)


source/dofs/dof_handler.cc
```diff
@@ -3246,22 +3193,12 @@ DoFHandler<dim, spacedim>::update_active_fe_table()
...
-    this->hp_cell_active_fe_indices.push_back({});
-
-  while (this->hp_cell_active_fe_indices.size() > this->tria->n_levels())
-    this->hp_cell_active_fe_indices.pop_back();
-
-  while (this->hp_cell_future_fe_indices.size() < this->tria->n_levels())
-    this->hp_cell_future_fe_indices.push_back({});
-
-  while (this->hp_cell_future_fe_indices.size() > this->tria->n_levels())
-    this->hp_cell_future_fe_indices.pop_back();
```
peterrum, line 3205: It's good that these while-loops a gone! I had to fight a lot with these a lot! Does anyone know why it was written like this?

source/dofs/dof_handler.cc
```diff
@@ -999,59 +1027,6 @@ namespace internal
...
-
-            for (unsigned int level = 0; level < dof_handler.tria->n_levels();
-                 ++level)
-              {
-                // recover backups
-                dof_handler.hp_cell_active_fe_indices[level] =
-                  std::move(active_fe_backup[level]);
-                dof_handler.hp_cell_future_fe_indices[level] =
-                  std::move(future_fe_backup[level]);
-              }
```
peterrum, line 1049: Why does the new code not deal with `hp_cell_active_fe_indices` and `hp_cell_future_fe_indices`?

marcfehling: Adjusting sizes of these containers happens in the `{pre|post}_transfer_action` functions.

Previously, all these internal containers have been cleared entirely and initialized from scratch. Thus, these fe indices had to be saved. But this feels out of place: enumerating dofs should not write in these fe index tables at all!

source/dofs/dof_handler.cc
```diff
@@ -250,6 +250,28 @@ namespace internal
...
+      /**
+       * Do that part of reserving space that pertains to releasing
+       * the previously used memory.
+       */
+      template <int dim, int spacedim>
+      static void
+      reserve_space_release_space(DoFHandler<dim, spacedim> &dof_handler)
+      {
+        dof_handler.object_dof_indices.assign(dof_handler.tria->n_levels(), {});
+        dof_handler.object_dof_indices.shrink_to_fit();
```
marcfehling, line 262: I don't which is faster: either replacing all entries of this vector with empty containers, or iterating over all entries and calling `clear` on each container. Anyways, I find this way easier to read!

peterrum:
@marcfehling Now that #11160 has been merged, I guess we can proceed here?

marcfehling:
I rebased this branch on the master one. Let's move on with this one! :)

drwells:
Could you rebase again? We fixed the mac problem in a recent PR.

source/dofs/dof_handler.cc
```diff
@@ -2511,7 +2489,9 @@ DoFHandler<dim, spacedim>::distribute_dofs(
          ExcMessage("The Triangulation you are using is empty!"));
   Assert(ff.size() > 0, ExcMessage("The hp::FECollection given is empty!"));
 
+  //
   // register the new finite element collection
+  //
```
peterrum, line 2494: ?

source/dofs/dof_handler.cc
```diff
@@ -2546,12 +2526,14 @@ DoFHandler<dim, spacedim>::distribute_dofs(
           "object instead."));
     }
 
+  //
   // enumerate all degrees of freedom
+  //
```
peterrum, line 2531: ?

source/dofs/dof_handler.cc
```diff
@@ -2569,104 +2551,72 @@ DoFHandler<dim, spacedim>::distribute_dofs(
...
+  // if(hp_capability_enabled)
+  //   {
+  //     Threads::TaskGroup<> tg;
+  //     for (int level = this->levels_hp.size() - 1; level >= 0; --level)
+  //       tg += Threads::new_task(
+  //         &dealii::internal::hp::DoFLevel::compress_data<dim, spacedim>,
+  //         *this->levels_hp[level],
+  //         this->fe_collection);
+  //     tg.join_all();
+  //   }
```
peterrum, line 2611: This reminds me that might want to compress the indices again. As I see, it is on the TODO list of https://github.com/dealii/dealii/issues/10333#issue-623676096.

peterrum (approved):
Looks good to me.

source/dofs/dof_handler.cc
marcfehling: I wanted to highlight these comments as they describe the whole following code block and not just a few upcoming lines. I felt like this gives a better overview over the structure of this function.

I don't have a problem with those lines being removed. Do you want them to be removed?

marcfehling:
@peterrum -- Ready for merge?

source/dofs/dof_handler.cc
```diff
@@ -3166,11 +3116,8 @@ DoFHandler<dim, spacedim>::create_active_fe_table()
   //  while (this->levels_hp.size() < this->tria->n_levels())
   //    this->levels_hp.emplace_back(new dealii::internal::hp::DoFLevel);
 
-  while (this->hp_cell_active_fe_indices.size() < this->tria->n_levels())
-    this->hp_cell_active_fe_indices.push_back({});
-
-  while (this->hp_cell_future_fe_indices.size() < this->tria->n_levels())
-    this->hp_cell_future_fe_indices.push_back({});
+  this->hp_cell_active_fe_indices.resize(this->tria->n_levels());
+  this->hp_cell_future_fe_indices.resize(this->tria->n_levels());
```
bangerth, line 3120: :-)

bangerth (commented):
Other than the small comments, the rest looks pretty good. This is low-risk stuff since we have a couple of thousand tests that execute these code paths. The worst that can happen is that we let some memory sit unused, but this looks good to me.

source/dofs/dof_handler.cc
marcfehling: Yeah, that does the job, too.

marcfehling: I will leave the `shrink_to_fit` call to also adjust the capacity. Alternatively, we can also `swap` with an empty vector.

bangerth (approved):
I don't think `shrink_to_fit()` after `clear()` and `resize()` does anything, but it doesn't hurt either.

marcfehling:
`clear` and `resize` change the size of the vector, but not necessarily their capacity.

Since these structures are vectors of vectors, I think the memory amount that we save with this call is rather small. So yeah in the end it probably doesn't matter...

bangerth:
It's true that `clear()` isn't *required* to change the capacity, but I'd be surprised if an implementation doesn't do so in practice :-) Regardless, it's fine this way!

PR #11151, opened 2020-11-06
Fix sm-mode L:d:V for zero sized vectors

peterrum:
With current master, I wanted to run all tests with sm-capability enabled in L:d:V by creating actually a sub-communicator by: 

```diff
diff --git a/include/deal.II/matrix_free/matrix_free.templates.h b/include/deal.II/matrix_free/matrix_free.templates.h
index 311f8b8..bc666a7 100644
--- a/include/deal.II/matrix_free/matrix_free.templates.h
+++ b/include/deal.II/matrix_free/matrix_free.templates.h
@@ -342,7 +342,33 @@ MatrixFree<dim, Number, VectorizedArrayType>::internal_reinit(
           Assert(additional_data.communicator_sm == MPI_COMM_SELF,
                  ExcNotImplemented());
 
-          task_info.communicator_sm = additional_data.communicator_sm;
+#ifdef DEAL_II_WITH_MPI
+          if (false)
+            {
+              task_info.communicator_sm = additional_data.communicator_sm;
+            }
+#  if DEAL_II_MPI_VERSION_GTE(3, 0)
+          else if (true)
+            {
+              MPI_Comm comm_sm;
+              MPI_Comm_split(task_info.communicator,
+                             task_info.my_pid / 2,
+                             task_info.my_pid,
+                             &comm_sm);
+              task_info.communicator_sm = comm_sm;
+            }
+          else
+            {
+              MPI_Comm comm_sm;
+              MPI_Comm_split_type(task_info.communicator,
+                                  MPI_COMM_TYPE_SHARED,
+                                  task_info.my_pid,
+                                  MPI_INFO_NULL,
+                                  &comm_sm);
+              task_info.communicator_sm = comm_sm;
+            }
+#  endif
+#endif
         }
       else
         {
```

However, following tests failed (actually timed out):

```
97% tests passed, 8 tests failed out of 278

Total Test time (real) = 1223.27 sec

The following tests FAILED:
	3607 - matrix_free/parallel_multigrid_02.mpirun=3.debug (Timeout)
	3608 - matrix_free/parallel_multigrid_02.mpirun=7.debug (Timeout)
	3616 - matrix_free/parallel_multigrid_adaptive_06.mpirun=3.debug (Timeout)
	3618 - matrix_free/parallel_multigrid_adaptive_07.mpirun=3.debug (Timeout)
	3619 - matrix_free/parallel_multigrid_adaptive_08.mpirun=3.debug (Timeout)
	3622 - matrix_free/parallel_multigrid_mf_02.mpirun=3.debug (Timeout)
	3639 - matrix_free/step-37-inhomogeneous-1.mpirun=4.debug (Timeout)
	3646 - matrix_free/stokes_computation.mpirun=4.debug (Timeout)
Errors while running CTest
```

After lengthy debugging  a couple of weeks ago, I was able to find out the problem was within `SolverCG`, to be precise in the `VectorMemory`. It calls the destructor of L:d:V on each process, however at different times. Since the destructor of  L:d:V is now sm-global operation, this might be not good and might lead to a deadlock (within the destructor^^).

~This PR tries to fix the issue by changing `SoverCG` and tackling the root cause.~

### DIFF
--- a/include/deal.II/lac/la_parallel_vector.templates.h
+++ b/include/deal.II/lac/la_parallel_vector.templates.h
@@ -134,27 +134,15 @@ namespace LinearAlgebra
         {
           if (comm_shared == MPI_COMM_SELF)
             {
-              if (new_alloc_size > allocated_size)
-                {
-                  Assert(((allocated_size > 0 && data.values != nullptr) ||
-                          data.values == nullptr),
-                         ExcInternalError());
+              Number *new_val;
+              Utilities::System::posix_memalign(
+                reinterpret_cast<void **>(&new_val),
+                64,
+                sizeof(Number) * new_alloc_size);
+              data.values = {new_val, [](Number *data) { std::free(data); }};
 
-                  Number *new_val;
-                  Utilities::System::posix_memalign(
-                    reinterpret_cast<void **>(&new_val),
-                    64,
-                    sizeof(Number) * new_alloc_size);
-                  data.values = {new_val,
-                                 [](Number *data) { std::free(data); }};
+              allocated_size = new_alloc_size;
 
-                  allocated_size = new_alloc_size;
-                }
-              else if (new_alloc_size == 0)
-                {
-                  data.values.reset();
-                  allocated_size = 0;
-                }
               data.values_sm = {
                 ArrayView<const Number>(data.values.get(), new_alloc_size)};
             }


tjhei:
> It calls the destructor of L:d:V on each process, however at different times.

You are removing the VectorMemory in this patch. I think it would help to understand why these are executed at different times. Is the order of destruction not deterministic? Or is the destruction happening after MPI_Finalize? Or something else?

kronbichler:
I agree with @tjhei here - I would be careful with saying that something needs to be "fixed" in `SolverCG` as long as we do not understand the root cause.
By the way, this PR would make all constructors taking a memory object like https://github.com/dealii/dealii/blob/10ceee9b3b6a0d21d9459dd4a9a64bc958d0fad8/include/deal.II/lac/solver_cg.h#L114-L116 obsolete, which is passed along to `SolverBase` here: https://github.com/dealii/dealii/blob/10ceee9b3b6a0d21d9459dd4a9a64bc958d0fad8/include/deal.II/lac/solver.h#L515-L519 
Also, one would need to see over all other solvers such as `SolverGMRES` or `SolverBicgstab` to guarantee that they also work.

peterrum:
This should be better!

kronbichler (approved):
Much better :+1:

tjhei:
Do we have to skip the size 0 optimization that you deleted?

kronbichler (approved):
I guess the size-0 case is not all that important for a distributed vector, given that we might have difficulties to control all cases with two different allocators. So this PR will definitely be on the safe side.

PR #11152, opened 2020-11-06
Finish mechanism for the update of ghost particles

### LINKED ISSUE #11093: Provide a mechanism to update ghost_particles location and properties without regenerating them

blaisb:
In this PR, we add a new feature that allow ghost particles to be recreated without recreating the map from scratch.
This is achieved by caching the necessary communication information in a new send_recv_cache_particles function (which at the moment has a lot of code duplication from send_recv_particles, but this was the least invasive way to do this).
Then we have a new function which only takes care of the update of particles themselves without altering the map.
The code could be made even more efficient by replacing the "destruction/deallocation/reconstruction" of the particles by a new Particle function that would "rewrite/update" the particle data. This would be done in  a follow-up PR.

The code is still a WIP, but it works. We have seen crazy performance improvement (from 2x to 6x as fast on 16 cores) on our applications. For us this PR is a game changer. Simulations that took days now take a few hours.

Right now the code is still in an ugly state (and I hate the fact that I had to copy over send_recv_particles). I will try to clean it ASAP, but I welcome any comments on how to make this better.
Again, this really makes me doubt that the multimap was the right structure for particle, but changing the particle container is out of the scope of the present PR. 

This closes #11093

### DIFF
--- a/include/deal.II/base/mpi_tags.h
+++ b/include/deal.II/base/mpi_tags.h
@@ -109,6 +109,8 @@ namespace Utilities
           /// ParticleHandler<dim, spacedim>::send_recv_particles
           particle_handler_send_recv_particles_setup,
           /// ParticleHandler<dim, spacedim>::send_recv_particles
+          particle_handler_send_recv_particles_cache_setup,
+          /// ParticleHandler<dim, spacedim>::send_recv_particles
           particle_handler_send_recv_particles_send,
 
           /// ScaLAPACKMatrix<NumberType>::copy_to
--- a/include/deal.II/particles/particle_handler.h
+++ b/include/deal.II/particles/particle_handler.h
@@ -705,7 +705,7 @@ namespace Particles
      * member variable.
      */
     void
-    exchange_ghost_particles();
+    exchange_ghost_particles(const bool enable_ghost_cache = false);
 
     /**
      * Update all particles that live in cells that are ghost cells to
@@ -820,14 +820,6 @@ namespace Particles
     std::multimap<internal::LevelInd, Particle<dim, spacedim>> ghost_particles;
 
     /**
-     * Set of particles that currently live in the ghost cells of the local
-     * domain, organized by the subdomain_id. These
-     * particles are equivalent to the ghost entries in distributed vectors.
-     */
-    std::map<types::subdomain_id, std::vector<particle_iterator>>
-      ghost_particles_by_domain;
-
-    /**
      * This variable stores how many particles are stored globally. It is
      * calculated by update_cached_numbers().
      */
@@ -927,6 +919,13 @@ namespace Particles
      * particle to be send in which the particle belongs. This parameter
      * is necessary if the cell information of the particle iterator is
      * outdated (e.g. after particle movement).
+     *
+     * @param [in] enable_cache Optional bool that enables updating
+     * the ghost particles without rebuilding them from scratch by
+     * building a cache of type GhostParticlePartitioner which
+     * stores the necessary information to update the ghost particles.
+     * Once this cache is built, the ghost particles can be updated
+     * by a call to send_recv_particles_properties_and_location().
      */
     void
     send_recv_particles(
@@ -941,8 +940,110 @@ namespace Particles
         &new_cells_for_particles = std::map<
           types::subdomain_id,
           std::vector<
-            typename Triangulation<dim, spacedim>::active_cell_iterator>>());
+            typename Triangulation<dim, spacedim>::active_cell_iterator>>(),
+      const bool enable_cache = false);
+
+    /**
+     * Transfer particles position and properties assuming that
+     * the particles have not changed cells. This routine uses the
+     * GhostParticlePartitioner as a caching structure to update the particles.
+     * It inherently assumes that particles cannot have changed cell.
+     * All updated particles will be appended to the
+     * @p received_particles vector.
+     *
+     * @param [in] particles_to_send All particles for which information
+     * should be sent and their new subdomain_ids are in this map.
+     *
+     * @param [in,out] received_particles A map with all received
+     * particles. Note that it is not required nor checked that the list
+     * is empty, received particles are simply attached to the end of
+     * the vector.
+     *
+     */
+    void
+    send_recv_particles_properties_and_location(
+      const std::map<types::subdomain_id, std::vector<particle_iterator>>
+        &particles_to_send,
+      std::multimap<internal::LevelInd, Particle<dim, spacedim>>
+        &received_particles);
+
+
 #endif
+
+
+    /**
+     * Cache structure used to store the elements which are required to
+     * exchange the particle information (location and properties) accross
+     * processors in order to update the ghost particles.
+     *
+     * This structure should only be used when one wishes to carry out work
+     * using the particles without calling
+     * sort_particles_into_subdomain_and_cells at every iteration. This is
+     * useful when particle-particle interaction occur at a different time
+     * scale than particle-FEM interaction.
+     *
+     * This structure is similar to Utilities::MPI::Partitioner::import_targets
+     * when combined with neighbors.
+     */
+    struct GhostParticlePartitioner
+    {
+      /**
+       * Indicates if the cache has been built to prevent updating particles
+       * with an invalid cache.
+       */
+      bool valid = false;
+
+      /**
+       * Vector of the subdomain id of all possible neighbors of the current
+       * subdomain.
+       */
+      std::vector<types::subdomain_id> neighbors;
+
+      /**
+       * Vector of size (neighbors.size()+1) used to store the start and the
+       * end point of the data that must go from the current subdomain to the
+       * neighbors. For neighbor i, send_pointers[i] indicates the beginning
+       * and send_pointers[i+1] indicates the end of the data that must be
+       * sent.
+       */
+      std::vector<unsigned int> send_pointers;
+
+      /**
+       * Set of particles that currently live in the ghost cells of the local
+       * domain, organized by the subdomain_id. These
+       * particles are equivalent to the ghost entries in distributed vectors.
+       */
+      std::map<types::subdomain_id, std::vector<particle_iterator>>
+        ghost_particles_by_domain;
+
+      /**
+       * Vector of size (neighbors.size()+1) used to store the start and the
+       * end point of the data that must be received from neighbor[i] on
+       * the current subdomain. For neighbor i, recv_pointers[i] indicate the
+       * beggining and reicv_pointers[i+1] indicates the end of the data that
+       * must be received.
+       */
+      std::vector<unsigned int> recv_pointers;
+
+      /**
+       * Vector of ghost particles in the order in which they are inserted
+       * in the multimap used to store particles on the triangulation. This
+       * information is used to update the ghost particle information
+       * without clearing the multimap of ghost particles, thus greatly
+       * reducing the cost of exchanging the ghost particles information.
+       */
+      std::vector<typename std::multimap<internal::LevelInd,
+                                         Particle<dim, spacedim>>::iterator>
+        ghost_particles_iterators;
+    };
+
+    /**
+     * Cache structure used to store the elements which are required to
+     * exchange the particle information (location and properties) accross
+     * processors in order to update the ghost particles. This structure
+     * is only used to update the ghost particles.
+     */
+    GhostParticlePartitioner ghost_particles_cache;
 
     /**
      * Called by listener functions from Triangulation for every cell
--- a/source/particles/particle.cc
+++ b/source/particles/particle.cc
@@ -180,6 +180,14 @@ namespace Particles
       property_pool->deallocate_properties_array(properties);
   }
 
+  template <int dim, int spacedim>
+  void
+  Particle<dim, spacedim>::free_properties()
+  {
+    if (property_pool != nullptr && properties != PropertyPool::invalid_handle)
+      property_pool->deallocate_properties_array(properties);
+  }
+
 
 
   template <int dim, int spacedim>
--- a/source/particles/particle_handler.cc
+++ b/source/particles/particle_handler.cc
@@ -90,7 +90,6 @@ namespace Particles
     , property_pool(std::make_unique<PropertyPool>(0))
     , particles()
     , ghost_particles()
-    , ghost_particles_by_domain()
     , global_number_of_particles(0)
     , global_max_particles_per_cell(0)
     , next_free_particle_index(0)
@@ -112,7 +111,6 @@ namespace Particles
     , property_pool(std::make_unique<PropertyPool>(n_properties))
     , particles()
     , ghost_particles()
-    , ghost_particles_by_domain()
     , global_number_of_particles(0)
     , global_max_particles_per_cell(0)
     , next_free_particle_index(0)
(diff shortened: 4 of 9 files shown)


include/deal.II/particles/particle_handler.h
```diff
@@ -942,6 +942,72 @@ namespace Particles
...
+    struct GhostParticlesCache
+    {
+      bool                             valid = false;
+      std::vector<types::subdomain_id> neighbors;
+      std::vector<unsigned int>        send_pointers;
+      std::vector<unsigned int>        recv_pointers;
+      std::vector<typename std::multimap<internal::LevelInd,
+                                         Particle<dim, spacedim>>::iterator>
+        ghost_particles_iterators;
+    };
```
peterrum, line 1007: I am thinking about making this struct a bit more like the `Partitioner` class by moving the new methods here and renaming it to `ParticlePartitioner`.

source/particles/particle.cc
```diff
@@ -180,6 +180,14 @@ namespace Particles
...
   }
 
+  template <int dim, int spacedim>
+  void
+  Particle<dim, spacedim>::free_properties()
+  {
+    if (property_pool != nullptr && properties != PropertyPool::invalid_handle)
+      property_pool->deallocate_properties_array(properties);
+  }
+
```
peterrum, line 190: I am not that familiar with the particle infrastructure: why is this needed? What is the effect?

source/particles/particle_handler.cc
```diff
@@ -1221,7 +1224,13 @@ namespace Particles
           }
       }
 
-    send_recv_particles(ghost_particles_by_domain, ghost_particles);
+    if (enable_cache)
+      {
+        send_recv_cache_particles(ghost_particles_by_domain, ghost_particles);
```
peterrum, line 1229: Maybe call it `send_recv_particles_and_build_cache`.

source/particles/particle_handler.cc
```diff
@@ -1483,7 +1499,440 @@ namespace Particles
   }
 #endif
 
+#ifdef DEAL_II_WITH_MPI
+  template <int dim, int spacedim>
+  void
+  ParticleHandler<dim, spacedim>::send_recv_cache_particles(
```
peterrum, line 1505: I thinks we might be able to merge `send_recv_cache_particles` and `send_recv_cache`.  As far as I see it, you perform in `send_recv_cache_particles` and store some additional data if the construction of the cache is requested.

source/particles/particle_handler.cc
```diff
@@ -1483,7 +1499,440 @@ namespace Particles
...
+    unsigned int particle_iterator_counter = 0;
+
+    while (reinterpret_cast<std::size_t>(recv_data_it) -
+             reinterpret_cast<std::size_t>(recv_data.data()) <
+           recv_pointers.back())
+      {
+        auto recv_particle =
+          ghost_particles_iterators[particle_iterator_counter];
+
+        recv_particle->second.free_properties();
```
peterrum, line 1918: I guess this is indeed needed?

include/deal.II/particles/particle_handler.h
blaisb: Makes sense, if it makes things more uniform :)

source/particles/particle.cc
blaisb: So this took me totally offguard. When you overwrite the particle to which an iterator points to, the destructor of that particle is not called. Consequently, it still has memory allocated in the memory pool and when the destructor of the memory pool is called, it crashes (because it asserts that the particles have de-allocated it's memory). Consequently, I had to add this routine to correctly clean the memory associated with particles before "overwriting them". Took me a while to figure this out. In the next PR I would replace this with a function that overwrites the data in the same memory space. This would prevent needless allocation/deallocation.

include/deal.II/base/mpi_tags.h
```diff
@@ -109,6 +109,8 @@ namespace Utilities
           /// ParticleHandler<dim, spacedim>::send_recv_particles
           particle_handler_send_recv_particles_setup,
           /// ParticleHandler<dim, spacedim>::send_recv_particles
+          particle_handler_send_recv_particles_cache_setup,
```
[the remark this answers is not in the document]
blaisb: The comment is wrong. It would be a comment for send_recv_cache_particles. My bad

source/particles/particle_handler.cc
blaisb: Sure thing, makes sense!

source/particles/particle_handler.cc
blaisb: Yeah. We could definitely merge them. We could pass the boolean "build_cache" to send_receive_particles and it would take care of building the cache if needed. The added cost would be a  few extra ifs, but it would remove all code duplication. I just did not want to do this now because I was developing it and it was already a mess to understand what was happening. But definitely, now that it works and that I have the test, that could be a good idea.

source/particles/particle_handler.cc
blaisb: Yes! Otherwise the property_pool will crash when it's destructor is called because it still has allocated memory.

blaisb:
@peterrum  If you have time (which I know is a rare thing) feel free to  make modifications directly in the original branch. I am absolutely open to any modifications as long as it keeps the feature there :).

peterrum:
@blaisb I have made some simplifications to our code. In particular, now the cache is set up in `send_recv_particles` (if requested).

I was not able to push to this branch; instead I have opened a PR against this branch (see https://github.com/lethe-cfd/dealii/pull/5).

May I ask you take care of the documentation and the failing tests on GitHub (probably due to used variables...). Thx.

blaisb:
> @blaisb I have made some simplifications to our code. In particular, now the cache is set up in `send_recv_particles` (if requested).
> 
> I was not able to push to this branch; instead I have opened a PR against this branch (see [lethe-cfd#5](https://github.com/lethe-cfd/dealii/pull/5)).
> 
> May I ask you take care of the documentation and the failing tests on GitHub (probably due to used variables...). Thx.

Absolutely. I'll take take of this early this week :)!
Thanks !

source/particles/particle_handler.cc
```diff
@@ -1309,16 +1333,26 @@ namespace Particles
...
     // are send, because we might receive particles from other processes
     if (n_send_particles > 0)
       {
         // Allocate space for sending particle data
-        const unsigned int particle_size =
-          begin()->serialized_size_in_bytes() + cellid_size +
-          (size_callback ? size_callback() : 0);
-        send_data.resize(n_send_particles * particle_size);
+        send_data.resize(n_send_particles *
+                         individual_total_particle_data_size);
```
kronbichler, line 1354: Do we ever expect this number to be larger than 4G? `MPI` will not allow more than 2G or 4G per message as used below, so all I would suggest here is `Assert(static_cast<std::size_t>(n_send_particles) * individual_total_particle_data_size == static_cast<std::size_t>(n_send_particles * individual_total_particle_data_size), ExcMessage("Overflow when trying to send particle data));`

source/particles/particle_handler.cc
```diff
@@ -1485,6 +1549,127 @@ namespace Particles
...
+    // Gather ghost particle iterators from the cache
+    auto &ghost_particles_iterators =
+      ghost_particles_cache.ghost_particles_iterators;
+    unsigned int particle_iterator_counter = 0;
+
+    while (reinterpret_cast<std::size_t>(recv_data_it) -
+             reinterpret_cast<std::size_t>(recv_data.data()) <
+           recv_pointers.back())
+      {
+        auto recv_particle =
```
kronbichler, line 1651: This is totally not your fault, but I wonder if these weird casts of pointers are the best approach. Maybe someone with more insight into the particles can explain me why this actually works?

kronbichler (commented):
This looks very nice to me! I have two only tangentially related comments below.

kronbichler:
> Right now the code is still in an ugly state (and I hate the fact that I had to copy over send_recv_particles). I will try to clean it ASAP, but I welcome any comments on how to make this better.

Regarding the code re-use, would it be an option to let the old function go into this function for the exchange part, after the communication has been constructed? At a superficial sight, it seems that one would not lose much by filling the communication pattern in a vector first, but I might be missing some details here.

peterrum:
> Regarding the code re-use,

@kronbichler  The code has changed somewhat since the comment in the description of the PR. The function `send_recv_particles()` sets up the communication pattern (+communicates the particle IDs) and performs one round of exchange of properties and locations (all in one go); the function `send_recv_particles_properties_and_location()` uses the communication pattern set up by the other function. In my opinion, this is a nice compromise between the case that you need to exchange ghost particles during every time step (as originally designed for) and the case that you only want to update the properties every time step (as targeted by this PR). The code duplication is not that much as it used to be. @blaisb Did I interpret you comment correctly?

source/particles/particle_handler.cc
blaisb: I must be honest and say I have no clue how these work... :(

source/particles/particle_handler.cc
blaisb: This is such a great idea. I have never thought about this limitation.

blaisb:
> > Regarding the code re-use,
> 
> @kronbichler The code has changed somewhat since the comment in the description of the PR. The function `send_recv_particles()` sets up the communication pattern (+communicates the particle IDs) and performs one round of exchange of properties and locations (all in one go); the function `send_recv_particles_properties_and_location()` uses the communication pattern set up by the other function. In my opinion, this is a nice compromise between the case that you need to exchange ghost particles during every time step (as originally designed for) and the case that you only want to update the properties every time step (as targeted by this PR). The code duplication is not that much as it used to be. @blaisb Did I interpret you comment correctly?

In the new  updated version by Peter and I code duplication has been completely eliminated :).
I am much more proud of this new version of it.

blaisb:
@peterrum  and @kronbichler I fixed all the documentation and the tests. There might be bits there and there to clean, but I think this version is ready for review (and it works :) )!

source/particles/particle_handler.cc
peterrum: I think we should replace this weird structure and simply iterate over all entries in `ghost_particles_iterators`.

peterrum (approved):
Looks good to me. Do you want to add a change-log entry for you last few changes?

source/particles/particle_handler.cc
```diff
@@ -1309,16 +1333,26 @@ namespace Particles
     std::vector<unsigned int> send_offsets(n_neighbors, 0);
     std::vector<char>         send_data;
 
+    // TODO: is size cachable?
+    const unsigned int individual_particle_data_size =
+      Utilities::MPI::max(n_send_particles > 0 ?
+                            ((begin()->serialized_size_in_bytes() +
+                              (size_callback ? size_callback() : 0))) :
+                            0,
+                          parallel_triangulation->get_communicator());
```
kronbichler, line 1342: I don't think the `MPI_Allreduce` behind `MPI::max` is all that bad compared to the other things that happen in this function. The only thing that would concern me is the synchronization point in case different MPI ranks run very far ahead others. That would suggest to skip the TODO. The other question is under which circumstances the `individual_particle_data_size` can actually change, which indeed would suggest to use caching.

source/particles/particle_handler.cc
kronbichler: To be on the safe side, which should then change the data type to `std::size_t` for `n_send_particles`. (I fear we will not easily be able to test whether the integer overflow is handled correctly because we would need to have huge amounts of data exchange.) But I think it is a good step to guard for the overflow.

kronbichler (approved):
Apart from a few minor comments, this looks good to me.

source/particles/particle_handler.cc
kronbichler: If that can be made to work, I would very much prefer it.

blaisb: @gassmoeller  I would presume this originates from you? It would be nice to have your input here :)

source/particles/particle_handler.cc
blaisb: Those are good questions. The `individual_particle_data_size` can only change if the number of properties changes. I think it is not possible right now to change the number of properties without building a new particle_handler from scratch. I think right now this should not be a significant overhead since anyway it is followed by two sets of communications with a WaitAll so we are at a section where everything will need to synchronize anyway.

kronbichler: Yes, let's keep things simple.

source/particles/particle_handler.cc
blaisb: For the loop in `send_recv_particles_positions_and_properties` I can modify it to use the vector to loop instead. For the regular `send_recv_particles` I don't have an alternative though :).

blaisb:
@kronbichler  I think I addressed all of your suggestions (which were very good). I managed to replace the annoying loop over the recv_data by a simple loop over the ghost particles iterators. I also added a changelog. I supposed it was a major change and added Peter as co-author.

:) Thanks for the help with this! Once this is merged, I'll move onto trying to improve this a bit more. I think right now that recreating the particle from scratch is still a bit of an expensive operation, but I will need to profile more to make sure of that.

blaisb:
The MPI build seems to have failed but I don't have access to the detail. It works on my machine (and it worked on OS X parallel) though... Is it a CI bug?

tjhei:
> Is it a CI bug?

Maybe, I had to restart things and tests are running again.

peterrum (approved):
Would you mind to squash the commits?

blaisb:
> Would you mind to squash the commits?

Done!
My evil plan to flood you with commits and to beat Wolfgang to top 1 contributor has been thwarted  ;)

peterrum (approved):

kronbichler (approved):

PR #11153, opened 2020-11-06
Avoid using deprecated DoFHandler properties in tests.

marcfehling:
~~Blocked by #11133, #11170, #11171, #11190.~~ (EDIT: PRs have been merged.)

Part of #11102 and #10333.

I removed all occurrences of deprecated functions from the `DoFHandler` unification in all tests.


tests/bits/count_dofs_per_component_hp_01.cc
```diff
@@ -50,33 +47,19 @@ test()
 
   FESystem<dim> fe_system(u, 2, p, 1);
 
-  hp::FECollection<dim> fe_collection;
-  fe_collection.push_back(fe_system);
-
-  hp::DoFHandler<dim> hp_dof_handler(triangulation);
-  DoFHandler<dim>     dof_handler(triangulation);
+  DoFHandler<dim> dof_handler(triangulation);
```
bangerth, line 50: Were the changes in this test intended? We previously checked the function both for the hp and the non-hp case, but yo're now only checking the non-hp case.

tests/bits/fe_tools_06d.cc
```diff
@@ -1,63 +0,0 @@
...
-  if (!fe1.constraints_are_implemented() || !fe2.constraints_are_implemented())
-    return;
-
-  Triangulation<dim> tria;
-  GridGenerator::hyper_cube(tria, 0., 1.);
-  tria.refine_global(2);
-
-  hp::FECollection<dim>                hp_fe1(fe1);
-  std::unique_ptr<hp::DoFHandler<dim>> hp_dof1(
-    make_hp_dof_handler(tria, hp_fe1));
```
bangerth, line 54: This test is now no longer run. Have you determined that it is no longer necessary? I'm asking because while the class itself has disappeared, the different functionality between hp and non-hp case continues to exist and should be tested.

tests/bits/fe_tools_common.h
```diff
@@ -122,18 +122,6 @@ make_dof_handler(const Triangulation<dim> &tria, const FiniteElement<dim> &fe)
...
 
 
-template <int dim>
-hp::DoFHandler<dim> *
-make_hp_dof_handler(const Triangulation<dim> &   tria,
-                    const hp::FECollection<dim> &fe)
-{
-  hp::DoFHandler<dim> *dof_handler = new hp::DoFHandler<dim>(tria);
-  dof_handler->distribute_dofs(fe);
-  return dof_handler;
```
bangerth, line 132: haha, did we create a memory leak here? :-)

tests/bits/map_dofs_to_support_points_hp_01.cc
```diff
@@ -55,28 +53,17 @@ test()
   MappingQ<dim>              mapping(2);
   hp::MappingCollection<dim> mapping_collection(mapping);
 
-  hp::FECollection<dim> fe_collection;
-  fe_collection.push_back(fe_system);
-
-  hp::DoFHandler<dim> hp_dof_handler(triangulation);
-  DoFHandler<dim>     dof_handler(triangulation);
+  DoFHandler<dim> dof_handler(triangulation);
```
bangerth, line 56: same question as above

bangerth (commented):
I only checked the first few files, but I'm having questions about some of the code that you remove.

tests/bits/count_dofs_per_component_hp_01.cc
marcfehling: Yes, they were intended. We now decide between hp and non-hp mode depending on the amount of finite elements in the `FECollection`: if it is only one element, non-hp mode is invoked, otherwise we're in hp mode. The `DoFHandler` stores any finite element in a `FECollection` internally.

So in this example, both objects `dof_handler` and `hp_dof_handler` are in fact operating in non-hp mode since `fe_collection` only contains one finite element. As both `DoFHandlers` are identical, I removed one of them.

tests/bits/fe_tools_06d.cc
marcfehling: This test would be identical to `tests/bits/fe_tools_06c.cc` after the change, which is why I removed it entirely.

We only have one finite element in the `FECollection` object. Thus, the `DoFHandler` internally operates in non-hp mode.

bangerth: Ah, I see. Thanks for clarifying!

tests/bits/count_dofs_per_component_hp_01.cc
bangerth: Ah, I had missed that there is only one element in the collection.

bangerth (approved):
Thanks! That must have been a ton of work!

marcfehling:
Ready for merge.

kronbichler (approved):

PR #11154, opened 2020-11-07
Fix compute_shared_memory_contiguous_indices

peterrum:


include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -1567,7 +1568,8 @@ MatrixFree<dim, Number, VectorizedArrayType>::initialize_indices(
                 if (cell < n_cell_batches())
                   {
                     // locally-owned cell
-                    cells_locally_owned.emplace_back(index, global_cell_index);
```
peterrum, line 1570: The `index` was wrong. I did not intend to use the matrix-free cell index but the "dealii" index.

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -1553,8 +1553,9 @@ MatrixFree<dim, Number, VectorizedArrayType>::initialize_indices(
                 const auto cell_iterator = get_cell_iterator(cell, v, di_index);
 
                 // determine global cell index
-                const unsigned int local_dof_index =
-                  di.dof_indices_contiguous[2][index];
+                const unsigned int local_cell_index =
+                  di.dof_indices_contiguous[2][index] /
+                  this->get_dofs_per_cell(di_index);
```
kronbichler, line 1558: Not related to this PR, but I note it now that I realize it: This will only work for the non-hp case when the number of dofs per cell is constant. Was this one of those places you wanted to apply #10490 to?

peterrum: Let's keep this out of this PR. Indeed, the current implementation does not work for hp. I am making the assumption that all cells of a `DoFHandler` have the same number of unknowns. This assumption allows me to set up the indices once and then scale the values by the number of dofs for each `DoFHandler` within `MatrixFree`. I can have a look at a generalization in a follow-up PR.

kronbichler (approved):

PR #11156, opened 2020-11-10
Improve VectorDataExchange::Full::reset_ghost_values

peterrum:


source/matrix_free/vector_data_exchange.cc
```diff
@@ -1344,10 +1344,18 @@ namespace internal
       void
       Full::reset_ghost_values_impl(const ArrayView<Number> &ghost_array) const
       {
-        // TODO
-        std::memset(ghost_array.data(),
-                    0.0,
-                    ghost_array.size() * sizeof(Number));
+        // reset ghost values coming from shared-memory neighbors
+        // TODO: only needed if values are buffered
```
peterrum, line 1348: I am still not completely sure where this logic should be.

source/matrix_free/vector_data_exchange.cc
```diff
@@ -1344,10 +1344,18 @@ namespace internal
...
+        for (const auto &i : sm_export_data_this.second)
+          std::memset(ghost_array.data() + (i.first - n_local_elements),
+                      0.0,
+                      sizeof(Number) * i.second);
+
+        // reset ghost values coming from remote neighbors
+        for (const auto &i : ghost_indices_subset_data.second)
+          std::memset(ghost_array.data() + i.first,
+                      0.0,
+                      sizeof(Number) * i.second);
```
drwells, line 1358: Could we use `std::fill` here?

source/matrix_free/vector_data_exchange.cc
```diff
@@ -1344,10 +1344,18 @@ namespace internal
...
       {
-        // TODO
-        std::memset(ghost_array.data(),
-                    0.0,
-                    ghost_array.size() * sizeof(Number));
+        // reset ghost values coming from shared-memory neighbors
+        // TODO: only needed if values are buffered
+        for (const auto &i : sm_export_data_this.second)
+          std::memset(ghost_array.data() + (i.first - n_local_elements),
+                      0.0,
```
kronbichler, line 1351: `memset` takes `int` arguments.
```suggestion
                      0,
```

peterrum: I have made the change.

source/matrix_free/vector_data_exchange.cc
peterrum: I left the `std::memset` because it is how it is done in `Partitioner`.

kronbichler (approved):

PR #11158, opened 2020-11-10
Add step-76

peterrum:
This PR adds an alternative version of step-67. It focuses and/or will focus on:
- element-centric loops (ECL)
- the new share-memory capabilities within `MatrixFree`


examples/step-67b/step-67b.cc
```diff
@@ -0,0 +1,1576 @@
...
+                                        ai[stage] * time_step),
+                                     (stage % 2 == 0 ? vec_ki : vec_ri),
+                                     (stage % 2 == 0 ? vec_ri : vec_ki),
+                                     solution,
+                                     vec_ri /*dummy*/);
+
+          if (stage > 0)
+            sum_previous_bi += bi[stage - 1];
+        }
+    }
```
peterrum, line 293: The `perform_time_step()` function has changed.

examples/step-67b/step-67b.cc
```diff
@@ -0,0 +1,1576 @@
...
+
+    const MPI_Comm &get_communicator() const
+    {
+      return comm;
+    }
+
+  private:
+    MPI_Comm comm;
+  };
```
peterrum, line 501: This class is new. I am not sure if this is the right place or we want to move something like this into `Utilities::MPI`.

examples/step-67b/step-67b.cc
```diff
@@ -0,0 +1,1576 @@
...
+      (update_JxW_values | update_quadrature_points | update_normal_vectors |
+       update_values);
+    additional_data.tasks_parallel_scheme =
+      MatrixFree<dim, Number>::AdditionalData::none;
+
+    MatrixFreeTools::categorize_by_boundary_ids(dof_handler.get_triangulation(),
+                                                additional_data);
+
+    additional_data.communicator_sm = subcommunicator.get_communicator();
+    additional_data.use_vector_data_exchanger_full = true;
```
peterrum, line 603: `MatrixFree` has to be set up slightly differently.

examples/step-67b/step-67b.cc
```diff
@@ -0,0 +1,1576 @@
...
+                  phi_temp.set_dof_values(solution);
+                }
+            }
+          }
+      },
+      vec_ki,
+      current_ri,
+      true,
+      MatrixFree<dim, Number>::DataAccessOnFaces::values);
+  }
```
peterrum, line 982: This function is new. It merges `local_apply_inverse_mass_matrix()`, `local_apply_cell()`, `local_apply_face()`, and `local_apply_boundary_face()`.

peterrum (commented):
@kronbichler This is a first draft of `step-67b` with ECL and exploiting the MPI-3.0 capabilities of `MatrixFree`/`L:d:V`. I have marked the relevant new parts compared to `step-67`. Let's discuss in person how to proceed.

bangerth:
Is there a good reason to keep both versions around? If they are solving the same problem but just slightly different techniques, what is the intent of keeping both versions? Shouldn't there be just one recommended way of doing things?

peterrum:
> Is there a good reason to keep both versions around? If they are solving the same problem but just slightly different techniques, what is the intent of keeping both versions? Shouldn't there be just one recommended way of doing things?

@bangerth I had this discussion with @kronbichler yesterday in person. Our goal is to show alternative, advanced, new techniques in `MatrixFree`. To do this we think it is the best to use an existing tutorial (step-67) and slightly modify it. In the assumption that the reader has already read the extensive documentation of step-67 we can here really concentrate on techniques, in particular on SM and cell-centric loops. Although I have worked on latter a lot in the last two years (and in hypder.deal it is the standard choice) it is by no mean as mature as the standard face-centric loop (`MatrixFree::loop()`), e.g. the interfaces are still a bit clumsy (and the user has to do actually do some 1D interpolations; which on completely depends on the given PDE) and adaptive meshes are still not supported.

For us it is important that the features become visible and used: maybe while usage we will be able to come up with better interfaces.

jppelteret:
> Our goal is to show alternative, advanced, new techniques in MatrixFree. To do this we think it is the best to use an existing tutorial (step-67) and slightly modify it.

We should align on the naming convention for these types of tutorials (i.e. those taking one tutorial as a base and making focussed modifications while leaving the majority of the program the same). I'm doing exactly the same thing here (https://github.com/dealii/dealii/pull/10393) but I've assigned this tutorial an entirely new number.

kronbichler:
Just to add on the previous answers: I think the important part here is to have something even more advanced than step-67 (which is pretty heavy already, so you don't want to overwhelm with additional options there), but only deviating at a few spots. One could have done it as "possibilities for extensions" there, but it would have felt a big step to do at that point. Maybe things turn out to go this new variant only in the future, so the current step-67 could be deprecated. But we are definitely not there yet, so it feels good to have some alternative.

> We should align on the naming convention for these types of tutorials (i.e. those taking one tutorial as a base and making focussed modifications while leaving the majority of the program the same). I'm doing exactly the same thing here (#10393) but I've assigned this tutorial an entirely new number.

Yes, it would be good to have a clear policy of how to do these "extend tutorial xx with enough novelty to have a separate program but yet pick up the basic methodology without repeating anything". I do not feel strongly either way of doing step-xxb or step-yy because both have their appeal. I would maybe slightly tend towards using a new number.

examples/step-67b/step-67b.cc
```diff
@@ -0,0 +1,1712 @@
...
+    }
+
+    const MPI_Comm &get_communicator() const
+    {
+      return comm;
+    }
+
+  private:
+    MPI_Comm comm;
+  };
```
peterrum, line 540: FYI @tjhei

tjhei: Awesome, I need that. Do you mind if we pull this out into the main library? Do you want to prep the PR or do you want me to do that?

peterrum: Feel free to do so. I would maybe drop the second option (`MPI_Comm_split(comm, rank / group_size, rank, &this->comm);`) because it makes the assumption that ranks in the communicator have not be reordered before that.

bangerth:
I've given this some more thought. I would favor the following approach, though:
* Not call it step-67b but give it a new name. If your goal is to show something new/different, then it isn't a variant, but a *separate* thing.
* Strip the commentary from all of the parts that are the same as before and just say something like "The following parts are verbatim copies of what has already been discussed in step-67". Let people read through the existing step-67 to understand what the program is about, and then focus the new tutorial on what is new. It's going to be impossible to keep documentation up to date otherwise -- no need to discuss the same things twice.
* Mention (if you haven't already) the new program in the "Possibilities for extensions" of step-67 (and/or at the end of the introduction) to point readers to alternative approaches.

peterrum:
> Not call it step-67b but give it a new name. If your goal is to show something new/different, then it isn't a variant, but a separate thing.

I have renamed it to step-76. At least the numbers are the same.

> Strip the commentary from all of the parts that are the same as before and just say something like "The following parts are verbatim copies of what has already been discussed in step-67". Let people read through the existing step-67 to understand what the program is about, and then focus the new tutorial on what is new. It's going to be impossible to keep documentation up to date otherwise -- no need to discuss the same things twice.

This is already this way.

> Mention (if you haven't already) the new program in the "Possibilities for extensions" of step-67 (and/or at the end of the introduction) to point readers to alternative approaches.

I have added a TODO and will add some text in the next days.

bangerth:
On 2/11/21 4:53 AM, Peter Munch wrote:
> I have renamed it to step-76. At least the numbers are the same.

There you go -- we're all better off with minmimalist changes ;-)

drwells:
Lets get back to this tutorial. Could you rebase and clean up the history?

examples/step-67b/step-67b.cc
bangerth: I'm in favor of just putting it into a function in `Utilities::MPI`. We already have `Utilities::MPI::duplicate_communicator()`. I don't think you actually need a whole class for this.

bangerth: Maybe more to the point/reason: What you do here currently is conflating two things -- the creation of subcommunicators, and the management of communicators in a C++ fashion. The first happens in the constructor, the second in the constructor/destructor pair providing a lifetime to these communicators. This sort of thing is better handled by separate entities, say a free function (or maybe a static member function) that returns objects of the this kind of type, and then a class that only manages the lifetime of communicators.

MPI used to have C++ wrapper classes for communicators and abandoned them. BOOST also has them, but they have never really gained wide acceptance. deal.II never had wrapper classes for communicators and does not seem to have a great need for them, though one could easily achieve this with `std::unique/shared_ptr` with a custom deleter that just calls `Utilities::MPI::free_communicator()`.

(Actually, on second look, it turns out that we *do* have a communicator wrapper class: `DuplicatedCommunicator`. Except it seems to be completely unused anywhere in the library, and furthermore has the same kind of design flaw as I'm trying to point out above. The person to fault for that is @bangerth ca. 2007.)

examples/step-76/doc/intro.dox
```diff
@@ -0,0 +1,468 @@
...
+queried by VectorizedArray::size() and its underlying number type by
+VectorizedArray::value_type.
+
+In the default case (<code>VectorizedArray<Number></code>), the vector length is
+set at compile time of the library to
+match the highest value supported by the given processor architecture.
+However, also a second optional template argument can be
+specified as <code>VectorizedArray<Number, size></code>, where <code>size</code> explicitly
+controls the  vector length within the capabilities of a particular instruction
+set. A full list of supported vector lengths is presented in the following table:
```
drwells, line 410: This information is useful but it should really be in a more basic matrix-free tutorial (like step-37) or a documentation module.

drwells (changes requested):
I have some time to work on reviewing this this week. I have a few tutorial-wide comments:
1. We should add the hyper deal paper to the bibliography and cite it normally
2. `step-67` has evolved since this was written and now uses `TimeStepping::LOW_STORAGE_RK_STAGE3_ORDER3` et al: can we use those here instead of reimplementing them?
3. Like Wolfgang pointed out - all the interesting stuff in `SubCommunicatorWrapper` is in the constructor. We should just have a utility function in the library itself instead of writing MPI wrapper classes.
4. Even if they are the same as step-67 it would be nice to include a picture or two of the solution at the end.

At this point I have read through this example several times and I personally don't think there are any other major things to address. The point of the tutorial is to demonstrate shared memory features and the cell centric loop and it does them both just fine.

examples/step-76/doc/intro.dox
```diff
@@ -0,0 +1,468 @@
...
+- On the other hand, since each entry of the solution vector is accessed exactly
+  once, no synchronization between threads is needed while accessing the solution
+  vector in the case of CCL. This absence of race conditions during writing into
+  the destination vector makes CCL particularly suitable for shared-memory
+  parallelization.
+
+One should also note that although fluxes are computed twice in the case of CCL,
+this does not automatically translate into doubling of the computation, since
+values already interpolated to the cell quadrature points can be interpolated
+to a face with a single 1D sweep.
```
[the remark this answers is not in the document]
peterrum: 1D is fine. E.g., the values are interpolated to the x-face via a 1D interpolation in x-direction.

examples/step-76/doc/intro.dox
peterrum: I would like to keep the table here since here it is the first place where we actually mess with the VectorizedArrayType

peterrum:
> Like Wolfgang pointed out - all the interesting stuff in SubCommunicatorWrapper is in the constructor. We should just have a utility function in the library itself instead of writing MPI wrapper classes.

I think I will simply move - tomorrow - the content of the class into the problem class.

peterrum:
> At this point I have read through this example several times and I personally don't think there are any other major things to address.

@drwells Thanks for taking the time and reviewing the tutorial! I have incorporated your suggestions.

examples/step-76/doc/intro.dox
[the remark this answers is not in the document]
drwells: I'm still getting caught up on the word 'sweep' - is this the term we use elsewhere in the library?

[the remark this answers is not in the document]
peterrum: OK. I modified the text to `with a simple 1D interpolation.`. I guess the original text was too much CS.

examples/step-76/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+# or switch altogether to the large project CMakeLists.txt file discussed
+# in the "CMake in user projects" page accessible from the "User info"
+# page of the documentation.
+SET(TARGET_SRC
+  ${TARGET}.cc
+  )
+
+# Usually, you will not need to modify anything beyond this point...
+
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)
```
drwells, line 24:
```suggestion
CMAKE_MINIMUM_REQUIRED(VERSION 3.1.0)
```

examples/step-76/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+##
+#  CMake script for the step-67b tutorial program:
+##
+
+# Set the name of the project and target:
+SET(TARGET "step-75")
```
drwells, line 6:
```suggestion
SET(TARGET "step-76")
```

drwells (changes requested):
I forgot to look at `CMakeLists.txt` last time. We have two small fixes there and then this should be done!

examples/step-76/CMakeLists.txt
peterrum: Done :+1:

examples/step-76/CMakeLists.txt
peterrum: Done :+1:

drwells (approved):

marcfehling (approved):
Nice work! The tutorial is highly polished. I also like the prose around it. Surprisingly little to do to enable the MPI-3.0 features from the user's point of view. You hid all the magic well in the MatrixFree framework :-)

I just found a few `TODO`s left and have only some minor suggestions that I don't feel strongly about.

examples/step-76/step-76.cc
```diff
@@ -0,0 +1,1634 @@
...
+  // and, therefore, setting the <code>size</code> to <code>-1</code> is a
+  // reasonable choice. By setting, the size to <code>1</code> users explicitly
+  // disable the MPI-3.0 shared-memory features of MatrixFree and rely
+  // completely on MPI-2.0 features, like <code>MPI_Isend</code> and
+  // <code>MPI_Irecv</code>.
+  template <int dim, int degree, int n_points_1d>
+  EulerOperator<dim, degree, n_points_1d>::EulerOperator(TimerOutput &timer)
+    : timer(timer)
+  {
+#if defined(DEAL_II_WITH_MPI) && DEAL_II_MPI_VERSION_GTE(3, 0)
```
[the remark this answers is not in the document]
peterrum: Indeed, this is the reason why I extended the macro last year ;)

peterrum:
@marcfehling Thanks for the review. I have made most of the changes!

drwells:
That puts us at two approvals - unless there are further comments I'll merge this tomorrow.

PR #11130, opened 2020-11-02
set number_cache.n_global_dofs to the right size in MultiGrid scenario

mathsen:
This should hopefully fix the problem, that after reordering of MG DoFs the number of level DoFs was set wrongly to the global number of DoFs, discussed [here](https://groups.google.com/forum/#!topic/dealii/VTrZS9xmk_I)

PR #11159, opened 2020-11-10
parallel running version of multigrid Cuthill McKee reordering

mathsen:
Since [this commit](https://github.com/dealii/dealii/pull/11130) the MG Cuthill McKee reordering also works for parallel distributed triangulations, but supports only one process. I slightly adjusted the compute_Cuthill_McKee() function in dof_renumbering and added a new function  extract_locally_active_level_dofs() in dof_tools (although active might be slightly misleading in a MG scenario - I didn't find a better name).
I tested this function locally with a convergence test and the results were as expected - therefore I also updated the renumber_cuthill_mc_kee test to test the changes.


source/dofs/dof_renumbering.cc
```diff
@@ -406,25 +411,50 @@ namespace DoFRenumbering
...
+      }
+    else
+      {
+        Assert(dof_handler.n_dofs(level) != numbers::invalid_dof_index,
+               ExcDoFHandlerNotInitialized());
+        DoFTools::extract_locally_relevant_level_dofs(dof_handler,
+                                                      level,
+                                                      locally_relevant_dofs);
+        ptr_locally_owned_dofs = &dof_handler.locally_owned_mg_dofs(level);
+      }
```
masterleinad, line 429: I would just initialize via an immediately invoked lambda like
```suggestion
    const IndexSet = [&]() {
      if (reorder_level_dofs == false)
        {
          DoFTools::extract_locally_relevant_dofs(dof_handler,
                                                  locally_relevant_dofs);
          return dof_handler.locally_owned_dofs();
        }
      else
        {
          Assert(dof_handler.n_dofs(level) != numbers::invalid_dof_index,
                 ExcDoFHandlerNotInitialized());
          DoFTools::extract_locally_relevant_level_dofs(dof_handler,
                                                        level,
                                                        locally_relevant_dofs);
          return dof_handler.locally_owned_mg_dofs(level);
        }
      }();
```

include/deal.II/dofs/dof_tools.h
```diff
@@ -1657,6 +1657,21 @@ namespace DoFTools
   extract_locally_active_dofs(const DoFHandler<dim, spacedim> &dof_handler,
                               IndexSet &                       dof_set);
 
+  /**
+   * Same function as above, but on a certain level in a multigrid scenarion.
+   * Strictly speaking we don't talk about active DoFs in such a multigrid
+   * scenario. Nevertheless this function returns all DoF indices that live on
+   * all locally owned cells (including on the interface to ghost cells) on a
+   * certain level, as the function above.
```
masterleinad, line 1665:
```suggestion
   * Same function as above but for a certain (multigrid-)level.
   * This function returns all DoF indices that live on
   * all locally owned cells (including on the interface to ghost cells) on the given level.
```

source/dofs/dof_renumbering.cc
```diff
@@ -406,25 +411,50 @@ namespace DoFRenumbering
...
 
     AffineConstraints<double> constraints;
     if (use_constraints)
       {
+        // regarding constraints is not yet implemented on a level basis
```
masterleinad, line 434:
```suggestion
        // reordering with constraints is not yet implemented on a level basis
```
?

source/dofs/dof_tools.cc
```diff
@@ -1133,6 +1133,43 @@ namespace DoFTools
...
+    // add the DoF on the adjacent ghost cells to the IndexSet, cache them
+    // in a set. need to check each dof manually because we can't be sure
+    // that the dof range of locally_owned_dofs is really contiguous.
+    std::vector<types::global_dof_index> dof_indices;
+    std::set<types::global_dof_index>    global_dof_indices;
+
+    typename DoFHandler<dim, spacedim>::level_cell_iterator
+      cell = dof_handler.begin_mg(level),
+      endc = dof_handler.end_mg(level);
+    for (; cell != endc; ++cell)
```
masterleinad, line 1155: Would you mind converting this to a range-based for loop?
```suggestion
    for (const auto& cell: dof_handler.mg_cell_iterators_on_level(level))
```

masterleinad (commented):
Just some stylistic suggsetions. Otherwise, this looks good to me.

source/dofs/dof_tools.cc
mathsen: @masterleinad Thanks for the feedback! Especially the lambda solution is much nicer than the pointer one.
Regarding the range-based for loop I wasn't sure how backwards compatible one should stay but changed it now to a filtered iterator and range based for loop solution.

masterleinad (approved):

mathsen:
I just pulled another change that slightly changed the lambda function part, so that no additional copy of the IndexSet is created and (as before this change) a const reference is used.

include/deal.II/dofs/dof_tools.h
```diff
@@ -1657,6 +1657,20 @@ namespace DoFTools
...
+   * given level.
+   */
+  template <int dim, int spacedim>
+  void
+  extract_locally_active_level_dofs(
+    const DoFHandler<dim, spacedim> &dof_handler,
+    IndexSet &                       dof_set,
+    const unsigned int               level);
+
+
```
kronbichler, line 1673: Nit-picking, but may I ask you to remove the second empty line to be consistent with the whitespace we add between functions?

source/dofs/dof_renumbering.cc
```diff
@@ -405,26 +410,51 @@ namespace DoFRenumbering
...
+      else
+        {
+          Assert(dof_handler.n_dofs(level) != numbers::invalid_dof_index,
+                 ExcDoFHandlerNotInitialized());
+          DoFTools::extract_locally_relevant_level_dofs(dof_handler,
+                                                        level,
+                                                        locally_relevant_dofs);
+          return dof_handler.locally_owned_mg_dofs(level);
+        }
+    }();
```
kronbichler, line 430: While I like the beauty of this function, I wonder whether the complication is actually necessary. Do I miss something or would the code below achieve the same thing (same number of lines, with one additional copy but so cheap that I as an HPC person would not see the need)?
```c++
    IndexSet locally_owned_dofs;
      if (reorder_level_dofs == false)
        {
          DoFTools::extract_locally_relevant_dofs(dof_handler,
                                                  locally_relevant_dofs);
          locally_owned_dofs = dof_handler.locally_owned_dofs();
        }
      else
        {
          Assert(dof_handler.n_dofs(level) != numbers::invalid_dof_index,
                 ExcDoFHandlerNotInitialized());
          DoFTools::extract_locally_relevant_level_dofs(dof_handler,
                                                        level,
                                                        locally_relevant_dofs);
          locally_owned_dofs = dof_handler.locally_owned_mg_dofs(level);
        }
```

kronbichler (commented):
This looks good. I have one additional request: Would you mind adding a test for `DoFTools::extract_locally_active_level_dofs()`. It can be similar to the present test and simply print the index sets.

source/dofs/dof_renumbering.cc
mathsen: @kronbichler Also thank you for the feedback! I added a simple test, similar to "extract_locally_active_dofs" for the level case.
To the current implementation of the lambda function: With the current one we indeed save some memory because no complete copy of the IndexSet "locally_owned_dofs" is created, we just save a const reference to it (that's why I initially just used a pointer to const, I didn't think of creating a case dependent reference with a lambda function). 
This won't probably become the memory critical part of the program but this was indeed the main motivation behind it.
Another minor advantage is that we can store the reference as "const" and can be sure, that we don't do any changes to the IndexSet by accident.
I really can't answer the question, personally I would go the "memory saving path", but of course if the deal.II developers prefer the "simpler code" solution, I will change it to your proposal.

kronbichler (approved):

source/dofs/dof_renumbering.cc
kronbichler: I do not feel very strong about it, so I let others decide.

Regarding the memory, I see your point. However, I would not be worried about the copy because most of the time `locally_owned_dofs` occupies around 100 bytes in memory. In case a single range is used (no renumbering that cross subdomain boundaries), the by far most usual case, we have 24 byte for that range, 24 byte for the std::vector, 16 bytes for the index space size and largest range, 8 bytes for the `is_compressed` boolean (due to alignment issues), and some memory, let's say 24 bytes, for `Threads::Mutex`. Of course one could construct numberings whose memory scales with the number of locally owned dofs, by using many non-contiguous ranges, but I would not optimize for those.

kronbichler (approved):
There are no more comments on this, so I am happy with the current state and will merge. Thanks for the effort of adding two nice tests!

PR #11162, opened 2020-11-11
Add reference to skew-symmetric implementation in step-67

peterrum:
@kronbichler I think the implementation in your code would be a good starting point. Do you agree?


kronbichler (approved):
Great idea!

PR #11163, opened 2020-11-11
Move GhostParticlePartitioner

peterrum:
Follow-up PR to #11152. This PR moves `GhostParticlePartitioner` to a new class. Currently, this class is only a dumb struct. My long term dream is to fill it with the actual communication algorithms just like it is the case in `Utilities::MPI::Partitioner` in the case for vectors. Maybe at some time, we can even eliminate the redundancies in partitioners (at least 3x) by reducing the problems to common algorithm.

FYI @blaisb

### DIFF
--- a/include/deal.II/particles/partitioner.h
+++ b/include/deal.II/particles/partitioner.h
@@ -1,0 +1,107 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2020 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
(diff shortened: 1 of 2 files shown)


include/deal.II/particles/partitioner.h
```diff
@@ -0,0 +1,106 @@
...
+     * processors in order to update the ghost particles.
+     *
+     * This structure should only be used when one wishes to carry out work
+     * using the particles without calling
+     * sort_particles_into_subdomain_and_cells at every iteration. This is
+     * useful when particle-particle interaction occur at a different time
+     * scale than particle-FEM interaction.
+     *
+     * This structure is similar to Utilities::MPI::Partitioner::import_targets
+     * when combined with neighbors.
```
kronbichler, line 41: Can you move this to `recv_pointers` (I had commented on this in the other PR, but I missed that it ended up in the wrong place in the end)?

kronbichler (approved):

include/deal.II/particles/partitioner.h
peterrum: Done.

kronbichler (approved):

Issue #11166, opened 2020-11-11
Make MatrixFree work for simplex meshes


peterrum:
The purpose of this issue is to collect problems while extending `MatrixFree` for pure simplex meshes and to mange the opened PRs.

peterrum:
There are some unresolved problems regarding face integrals for simplex meshes in `MatrixFree`:
- First of all, we need somehow the face quadrature inside of `MatrixFree`. Should we pass it from outside during `MatrixFree::reinit()` or should happen an implicit construction of the face quadrature rule internally?
- `ShapeInfo` needs the evaluation of the (cell) shape functions at each face (for the correct orientation). The normal way in `FEValues` is have a "table" of matrices (of size n_faces x n_orientations). I guess this approach could also work for `ShapeInfo`.
- The functions `FEFaceEvaluationImplEvaluateSelector::run()` and `FEFaceEvaluationImplIntegrateSelector::run()` both call in different order `FEFaceNormalEvaluationImpl::interpolate()`, `FEFaceEvaluationImpl::evaluate_in_face()`/`FEFaceEvaluationImpl::integrate_in_face()`, and `adjust_for_face_orientation()`. As far as I can see, this order is not working in the general case. I would suggest to introduce the functions `FEFaceEvaluationImpl::evaluate()`/`FEFaceEvaluationImpl::integrate()`, which do the evaluation/integration with the help of the above mentioned tables in `ShapeInfo` (accessing the right face_no and orientation - both information available).

peterrum:
This is already working^^

PR #11167, opened 2020-11-11
Add Poisson/Helmholtz test: MatrixFree + simplex mesh

peterrum:
References #11166.

~Depends on #11186 and on #11188.~

### DIFF
--- a/tests/simplex/matrix_free_01.cc
+++ b/tests/simplex/matrix_free_01.cc
@@ -1,0 +1,294 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2020 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// Solve Poisson problem and Helmholtz problem on a simplex mesh with
+// continuous elements and compare results between matrix-free and matrix-based
+// implementations.
+
+#include <deal.II/distributed/tria.h>
+
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/fe/fe_q.h>
+#include <deal.II/fe/mapping_fe.h>
+
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/tria.h>
+
+#include <deal.II/lac/la_parallel_vector.h>
+#include <deal.II/lac/precondition.h>
+#include <deal.II/lac/solver_cg.h>
+#include <deal.II/lac/solver_control.h>
+#include <deal.II/lac/sparse_matrix.h>
+
+#include <deal.II/matrix_free/fe_evaluation.h>
+#include <deal.II/matrix_free/matrix_free.h>
+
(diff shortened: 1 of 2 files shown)


include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -369,12 +410,20 @@ namespace internal
...
+
+              if (flag == false)
                 {
-                  Assert(quad[my_q][hpq].get_tensor_basis()[0] ==
-                           quad[my_q][hpq].get_tensor_basis()[i],
-                         ExcNotImplemented());
+#ifdef DEAL_II_WITH_SIMPLEX_SUPPORT
+                  continue; // TODO: face data is not set up
+#else
+                  Assert(false, ExcNotImplemented())
```
drwells, line 425:
```suggestion
                  Assert(false, ExcNotImplemented());
```

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -338,27 +365,41 @@ namespace internal
...
```
drwells, line 401:
```suggestion
                  Assert(false, ExcNotImplemented());
```

tests/simplex/matrix_free_02.cc
```diff
@@ -0,0 +1,178 @@
...
+  SolverCG<LinearAlgebra::distributed::Vector<double>> solver(
+    reduction_control);
+  solver.solve(poisson_operator, x, b, PreconditionIdentity());
+
+  if (Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == 0)
+    printf("Solved in %d iterations.\n", reduction_control.last_step());
+
+  constraints.distribute(x);
+
+#if false
```
drwells, line 169:
```suggestion
#if 0
```

I suspect (but I don't have evidence on hand) that some toolchains might print a warning with `#if false` so I'd prefer the unambiguously false version.

drwells (dismissed):
I have some minor requests.

Why don't things work with 3D? It's fine if its not in the PR but I don't see why this is specific to 2D.

peterrum:
> Why don't things work with 3D? It's fine if its not in the PR but I don't see why this is specific to 2D.

I can do that. Probably, it would be also good to compare with an actual sparse matrix vector multiplication. I'll check how the tests are structured in the `matrix_free` test folder.

tests/simplex/matrix_free_02.cc
```diff
@@ -0,0 +1,178 @@
...
+    matrix_free.template cell_loop<VectorType, VectorType>(
+      [&](const auto &, auto &dst, const auto &src, const auto cells) {
+        FEEvaluation<dim, -1, 0, 1, double> phi(matrix_free);
+        for (unsigned int cell = cells.first; cell < cells.second; ++cell)
+          {
+            phi.reinit(cell);
+            phi.gather_evaluate(src, false, true);
+
+            for (unsigned int q = 0; q < phi.n_q_points; ++q)
+              phi.submit_gradient(phi.get_gradient(q), q);
```
kronbichler, line 98: Do we need a test that also goes through the `evaluate(EvaluationFlags::values)` case? If I understand things correctly that part is not yet tested (whereas the integrate part is due to the rhs). Nothing wrong with testing the Laplacian, but I just want to explain why the Helmholtz operator is my favorite: https://github.com/dealii/dealii/blob/7b87282453f779f47413a126b448c3fec7e73c2d/tests/matrix_free/matrix_vector_mf.h#L47-L51

peterrum:
@drwells @kronbichler I have extended the test so that it tests a Poisson problem and a Helmholtz problem for 2D and 3D, as well as, for linear and quadratic elements. The results are compared to a matrix-based implementation.

kronbichler (approved):

PR #11168, opened 2020-11-11
Make MatrixFree compile for arbitrary dim quadrature rules

peterrum:
References #11166.

### DIFF
--- a/include/deal.II/hp/q_collection.h
+++ b/include/deal.II/hp/q_collection.h
@@ -53,12 +53,19 @@ namespace hp
     QCollection() = default;
 
     /**
+     * Copy constructor.
+     */
+    template <int dim_in>
+    QCollection(const QCollection<dim_in> &other);
+
+    /**
      * Conversion constructor. This constructor creates a QCollection from a
      * single quadrature rule. More quadrature formulas can be added with
      * push_back(), if desired, though it would probably be clearer to add all
      * mappings the same way.
      */
-    explicit QCollection(const Quadrature<dim> &quadrature);
+    template <int dim_in>
+    explicit QCollection(const Quadrature<dim_in> &quadrature);
 
     /**
      * Constructor. This constructor creates a QCollection from one or
@@ -93,8 +100,9 @@ namespace hp
      * is later destroyed by this object upon destruction of the entire
      * collection.
      */
+    template <int dim_in>
     void
-    push_back(const Quadrature<dim> &new_quadrature);
+    push_back(const Quadrature<dim_in> &new_quadrature);
 
     /**
      * Return a reference to the quadrature rule specified by the argument.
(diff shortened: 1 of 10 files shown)


drwells (approved):

kronbichler (commented):
While I like the direction in which we are going, I wonder whether it would make sense to use `Quadrature<dim>` in all cases and internally extract the 1D quadrature via Quadrature::get_tensor_basis()` in https://github.com/dealii/dealii/blob/0717633534b77706e89a62758837338294ebe02e/include/deal.II/base/quadrature.h#L250-L275  While this won't solve the issue of what interface we expose to the user (where I believe both a 1D and a dim-D quadrature would be nice), at least for the internal data structures we might be able to avoid the additional template argument.

include/deal.II/hp/q_collection.h
```diff
@@ -239,8 +268,9 @@ namespace hp
 
 
   template <int dim>
+  template <int dim_in>
   inline void
-  QCollection<dim>::push_back(const Quadrature<dim> &new_quadrature)
+  QCollection<dim>::push_back(const Quadrature<dim_in> &new_quadrature)
```
peterrum, line 273: Part of #11176.

peterrum:
@kronbichler I hope you like the new version more!

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -337,18 +337,28 @@ namespace internal
           AssertIndexRange(0, n_hp_quads);
           cell_data[my_q].descriptor.resize(n_hp_quads);
           for (unsigned int q = 0; q < n_hp_quads; ++q)
-            cell_data[my_q].descriptor[q].initialize(quad[my_q][q],
-                                                     update_default);
+            {
+              Assert(quad[my_q][q].is_tensor_product(), ExcNotImplemented());
+              cell_data[my_q].descriptor[q].initialize(
+                quad[my_q][q].get_tensor_basis()[0], update_default);
```
kronbichler, line 343: In principle, we would also need a second assert here that checks that all of the 1D quadrature formulas returned by `get_tensor_basis` are the same. While algorithms could be extended, that would be a lot of work and we should catch it before getting unexpected results.

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -1293,7 +1293,7 @@ MatrixFree<dim, Number, VectorizedArrayType>::initialize_indices(
   Table<2, internal::MatrixFreeFunctions::ShapeInfo<double>> shape_info_dummy(
     shape_info.size(0), shape_info.size(2));
   {
-    QGauss<1> quad(1);
+    Quadrature<dim> quad(QGauss<1>(1));
```
kronbichler, line 1296: Why not
```suggestion
    QGauss<dim> quad(1));
```

include/deal.II/matrix_free/shape_info.h
```diff
@@ -313,7 +313,7 @@ namespace internal
        * Constructor that initializes the data fields using the reinit method.
        */
       template <int dim>
-      ShapeInfo(const Quadrature<1> &     quad,
+      ShapeInfo(const Quadrature<dim> &   quad,
```
kronbichler, line 316: While within an internal namespace, this is still an incompatible change of something that might have appeared in user code, so I would only make the change incompatible if there is a clear benefit for doing so. Do you think it would make sense to add a second template argument `q_dim` here? (I am not entirely sure what is the better approach at this point).

kronbichler:
Yes, I like this variant more and think it is getting very nice!

peterrum:
@kronbichler I have updated this PR according to your suggestions. Ready from my side; unfortunately, the PR is blocked by #11176, which is the reason for the `Do not merge` label.

kronbichler (approved):
Looks good to me. Could you please rebase this onto master to get rid of the code from #11176 (or does git handle the merge just fine anyway since we do not conflict between the two)?

peterrum:
I am also confused since I have expected a merge conflict. But in anyway I am currently on it!

peterrum:
@kronbichler No need to push again, git seems not to have any problems with merging with master. I tested all matrix_free tests.

PR #11171, opened 2020-11-12
Deprecate hp::DoFHandler.

marcfehling:
~~Blocked by #11160.~~ (EDIT: With the use of constructor aliases, there is no need to wait on a decision for the DoFHandler interface)

As there is no need for the hp-specialization anymore since #10328, we can deprecate `hp::DoFHandler`. @peterrum already left a changelog in that particular PR.

In this PR, we actually set the corresponding deprecated flag for `hp::DoFHandler`, and adjust the general documentation for the basic `DoFHandler` class. Further, all references to that particular class are removed in the deal.II library (in those entities which are not deprecated).

I'll address removing the `hp::DoFHandler` from the documentation in a separate PR.

Part of #10333.

### DIFF
--- a/include/deal.II/hp/dof_handler.h
+++ b/include/deal.II/hp/dof_handler.h
@@ -119,45 +119,13 @@ namespace hp
    * @ingroup dofs
    * @ingroup hp
    *
-   * @note Task is delegated to the base class dealii::DoFHandler.
+   * @deprecated The basic dealii::DoFHandler is capable of hp-adaptation now.
    */
   template <int dim, int spacedim = dim>
-  class DoFHandler : public dealii::DoFHandler<dim, spacedim>
+  class DEAL_II_DEPRECATED DoFHandler : public dealii::DoFHandler<dim, spacedim>
   {
-  public:
-    /**
-     * Default Constructor.
-     *
-     * @deprecated Use the overload taking a `bool` instead.
-     */
-    DEAL_II_DEPRECATED
-    DoFHandler();
-
-    /**
-     * Constructor. Take @p tria as the triangulation to work on.
-     *
-     * @deprecated Use the overload taking a
-     * `const Triangulation<dim, spacedim>&` and a `bool` instead.
-     */
-    DEAL_II_DEPRECATED
-    DoFHandler(const Triangulation<dim, spacedim> &tria);
-
-    /**
-     * Copy constructor. DoFHandler objects are large and expensive.
-     * They should not be copied, in particular not by accident, but
-     * rather deliberately constructed. As a consequence, this constructor
-     * is explicitly removed from the interface of this class.
-     */
-    DoFHandler(const DoFHandler &) = delete;
-
-    /**
-     * Copy operator. DoFHandler objects are large and expensive.
-     * They should not be copied, in particular not by accident, but
-     * rather deliberately constructed. As a consequence, this operator
-     * is explicitly removed from the interface of this class.
-     */
-    DoFHandler &
-    operator=(const DoFHandler &) = delete;
+    using dealii::DoFHandler<dim, spacedim>::DoFHandler;
+    using dealii::DoFHandler<dim, spacedim>::operator=;
   };
 
 } // namespace hp
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -631,14 +631,14 @@ public:
          const AdditionalData &additional_data = AdditionalData());
 
   /**
-   * Initializes the data structures. Same as above, but  using hp::DoFHandlers.
+   * Initializes the data structures. Same as above, but  using DoFHandlerType.
    *
    * @deprecated Use the overload taking a DoFHandler object instead.
    */
-  template <typename QuadratureType, typename number2>
+  template <typename QuadratureType, typename number2, typename DoFHandlerType>
   DEAL_II_DEPRECATED void
   reinit(const Mapping<dim> &                                   mapping,
-         const std::vector<const hp::DoFHandler<dim> *> &       dof_handler_hp,
+         const std::vector<const DoFHandlerType *> &            dof_handler,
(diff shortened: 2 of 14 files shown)


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -628,19 +628,6 @@ class MatrixFree : public Subscriptor
...
-   *
-   * @deprecated Use the overload taking a DoFHandler object instead.
-   */
-  template <typename QuadratureType, typename number2>
-  DEAL_II_DEPRECATED void
-  reinit(const Mapping<dim> &                                   mapping,
-         const std::vector<const hp::DoFHandler<dim> *> &       dof_handler_hp,
-         const std::vector<const AffineConstraints<number2> *> &constraint,
-         const std::vector<QuadratureType> &                    quad,
-         const AdditionalData &additional_data = AdditionalData());
```
peterrum, line 642: We cannot do this. I only have deprecated these after the last release.

masterleinad: Correct.

marcfehling: Yes I know, but otherwise the CI complains about using the deprecated `hp::DoFHandler` :(

I thought that technically, this wouldn't make a difference. As the base class of `hp::DoFHandler` is actually the basic `DoFHandler`. So when using a `hp::DoFHandler`, we would now automatically invoke the `DoFHandler` equivalent. Code like this compiles: https://godbolt.org/z/do6znG

peterrum: The problem is the case `std::vector<B *> bar{&b, &b};`.

peterrum: I would really love to see 
```
namespace hp
{
  DEAL_II_DEPRECATED using DoFHandler = dealii::DoFHandler<dim, spacedim>;
}
```
After all the changes this might already work?

marcfehling: I see the problem now. I'll try to find a workaround to use the `reinit` function without triggering the deprecation warning.

With your suggestion for the alias I get
```
error: conflicting declaration of template ‘template<int dim, int spacedim> using DoFHandler = dealii::DoFHandler<dim, spacedim>’
```
Even in the hp namespace. Maybe we should leave it as is.

masterleinad: I mean as soon as hp::DoFHandler is deprecated all functions using it are implicitly deprecated. So it would probably be fine to remove the `[[deprecated]]` attribute from all problematic functions if that helps.

marcfehling: CI is particularly complaining about using deprecated arguments in these function, not about the deprecation of the functions themselves.

Affected examples are as far as I can tell: 37, 48, 50, 59, 67

Testers complain with errors like:
```
error: 'DoFHandler<2, 2>' is deprecated
```
Using `gcc 8.1` without the `-Werror` option:
```
warning: ‘template<int dim, int spacedim> class dealii::hp::DoFHandler’ is deprecated
```

marcfehling:
I've found an example in `glib` on how to tell the compiler to ignore deprecation warnings. I gave that a try and suppressed warnings on the `reinit` functions.

marcfehling:
Just as I wrote these lines, I've found the `DEAL_II_DISABLE_EXTRA_DIAGNOSTICS` pragma in `base/config.h` :-)

kronbichler (approved):
This looks good to me.

include/deal.II/hp/dof_handler.h
```diff
@@ -124,36 +124,8 @@ namespace hp
...
-    /**
-     * Copy operator. DoFHandler objects are large and expensive.
-     * They should not be copied, in particular not by accident, but
-     * rather deliberately constructed. As a consequence, this operator
-     * is explicitly removed from the interface of this class.
-     */
-    DoFHandler &
-    operator=(const DoFHandler &) = delete;
+    using dealii::DoFHandler<dim, spacedim>::DoFHandler;
+    using dealii::DoFHandler<dim, spacedim>::operator=;
```
marcfehling, line 128: Not sure whether this line is necessary, but it probably won't hurt.

marcfehling:
With the use of constructor aliases, there is no need to wait on a decision for the `DoFHandler` interface. So we can deal with this independently. Ready for final review.

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -630,6 +630,8 @@ class MatrixFree : public Subscriptor
          const std::vector<QuadratureType> &                    quad,
          const AdditionalData &additional_data = AdditionalData());
 
+  DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
```
peterrum, line 633: I am wondering if you can write:
```cpp
  template <typename QuadratureType, typename number2, typename DoFHandlerType>
  DEAL_II_DEPRECATED void
  reinit(const Mapping<DoFHandlerType::dimension> &                                   mapping,
         const std::vector<const DoFHandlerType *> &       dof_handler_hp,
         const std::vector<const AffineConstraints<number2> *> &constraint,
         const std::vector<QuadratureType> &                    quad,
         const AdditionalData &additional_data = AdditionalData());
```
The other function should be simply a specialization.

peterrum (approved):
Looks good. I have one question.

include/deal.II/matrix_free/matrix_free.h
marcfehling: ...and leave out the extra diagnostics flag? Hmm this could work.

There are a couple of tests using the hp-variant in `tests/matrix_free`. I'll add a commit and we'll see whether the testsuite complains.

marcfehling:
Testsuite is okay with the `DoFHandlerType` argument. So this is ready for merge!

peterrum (approved):

PR #11176, opened 2020-11-12
QCollection<dim> accept Quadrature<1>

peterrum:


include/deal.II/hp/q_collection.h
```diff
@@ -150,21 +158,41 @@ namespace hp
 
   /* --------------- inline functions ------------------- */
 
+  template <int dim>
+  template <int dim_in>
+  QCollection<dim>::QCollection(const QCollection<dim_in> &other)
+  {
+    for (unsigned int i = 0; i < other.size(); ++i)
+      push_back(other[i]);
```
kronbichler, line 166: Do we need an assertion somewhere to check that the dimensions make sense - or is that already in `Quadrature`? The reason I am asking is that we cannot build anything useful when `dim_in=2` and `dim=3`, for example.

kronbichler (approved):
This looks good to me.

include/deal.II/hp/q_collection.h
peterrum: This is handled by `Quadrature<dim>`: it only can accept 1D or dim-D quadrature rules as input.

PR #11177, opened 2020-11-12
Make assert more verbose in FEEvaluation

peterrum:
The assert in `FEEvaluation` seems to be cryptic and not easy to understand. This PR proposes to make the assert messages a bit output.

For example if `quadrature_point()` is called without enabling `UpdateFlags::update_quadrature_points` flag the output might look like this:
```
The violated condition was: 
    this->mapping_data->quadrature_point_offsets.empty() == false
Additional information: 
    Requested field in MatrixFree is not initialized.

This might happen if MatrixFree was set up without UpdateFlags::update_quadrature_points.
```

Once we agree on an appropriate format, I would insert the new exception at more places in the file. One question might be also if we want to distinguish between cell and face update flags.

FYI @mschreter

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -48,7 +48,18 @@ DEAL_II_NAMESPACE_OPEN
 namespace internal
 {
   DeclException0(ExcAccessToUninitializedField);
-}
+
+  DeclException1(
+    ExcMatrixFreeAccessToUninitializedMappingField,
+    std::string,
+    << "You are requesting information from an FEEvaluation/FEFaceEvaluation "
+    << "object for which this kind of information has not been computed. What "
+    << "information these objects compute is determined by the update_* flags you "
+    << "pass to MatrixFree::reinit() via MatrixFree::AdditionalData. Here, "
+    << "the operation you are attempting requires the <" << arg1
+    << "> flag to be set, but it was apparently not specified "
+    << "upon initialization.");
+} // namespace internal
 
 template <int dim,
           int fe_degree,
@@ -3546,7 +3557,9 @@ inline DEAL_II_ALWAYS_INLINE Tensor<1, dim, VectorizedArrayType>
   get_normal_vector(const unsigned int q_point) const
 {
   AssertIndexRange(q_point, n_quadrature_points);
-  Assert(normal_vectors != nullptr, ExcMessage("Did not call reinit()!"));
+  Assert(normal_vectors != nullptr,
+         internal::ExcMatrixFreeAccessToUninitializedMappingField(
+           "update_normal_vectors"));
   if (this->cell_type <= internal::MatrixFreeFunctions::flat_faces)
     return normal_vectors[0];
   else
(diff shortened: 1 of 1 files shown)


kronbichler:
I like this way of displaying the assertion. It would of course be great to be able to distinguish cell and face update flags, but I leave it up to you if you want to spend the effort. It will be a good improvement already without that.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -5445,7 +5457,8 @@ inline DEAL_II_ALWAYS_INLINE Tensor<1, n_components_, VectorizedArrayType>
          internal::ExcAccessToUninitializedField());
 #  endif
 
-  Assert(this->normal_x_jacobian != nullptr, ExcNotInitialized());
+  Assert(this->normal_x_jacobian != nullptr,
+         internal::MatrixFreeNotInitialized("???"));
```
kronbichler, line 5461: `update_gradients` In principle one should also check for the normal vectors, but we force this to be done at the moment here: https://github.com/dealii/dealii/blob/5163ed818e5a008bdcce7ea53317dca7c6e250a0/include/deal.II/matrix_free/mapping_info.templates.h#L325-L330

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -5909,7 +5926,8 @@ FEEvaluationBase<dim, n_components_, Number, is_face, VectorizedArrayType>::
     const unsigned int                                  q_point)
 {
   AssertIndexRange(q_point, this->n_quadrature_points);
-  Assert(this->normal_x_jacobian != nullptr, ExcNotInitialized());
+  Assert(this->normal_x_jacobian != nullptr,
+         internal::MatrixFreeNotInitialized("???"));
```
kronbichler, line 5930: `update_gradients`

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -6468,7 +6490,7 @@ inline DEAL_II_ALWAYS_INLINE VectorizedArrayType
          internal::ExcAccessToUninitializedField());
 #  endif
   AssertIndexRange(q_point, this->n_quadrature_points);
-  Assert(this->jacobian != nullptr, ExcNotInitialized());
+  Assert(this->jacobian != nullptr, internal::MatrixFreeNotInitialized("???"));
```
kronbichler, line 6493: `update_gradients`

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -6624,8 +6646,8 @@ FEEvaluationAccess<dim, dim, Number, is_face, VectorizedArrayType>::
 {
   Assert(this->cell != numbers::invalid_unsigned_int, ExcNotInitialized());
   AssertIndexRange(q_point, this->n_quadrature_points);
-  Assert(this->J_value != nullptr, ExcNotInitialized());
-  Assert(this->jacobian != nullptr, ExcNotInitialized());
+  Assert(this->J_value != nullptr, internal::MatrixFreeNotInitialized("???"));
+  Assert(this->jacobian != nullptr, internal::MatrixFreeNotInitialized("???"));
```
kronbichler, line 6650: `update_gradients` (the former would be fine with `update_values` as well, but let's stick to the relevant one here).

kronbichler (approved):
Here are the missing messages.

peterrum:
Thanks @kronbichler! These were also my guess ;)

bangerth:
That's essentially the approach used in `FEValues`. For example here:
```
inline const std::vector<double> &
FEValuesBase<dim, spacedim>::get_JxW_values() const
{
  Assert(this->update_flags & update_JxW_values,
         ExcAccessToUninitializedField("update_JxW_values"));
  Assert(present_cell.get() != nullptr,
         ExcMessage("FEValues object is not reinit'ed to any cell"));
  return this->mapping_output.JxW_values;
}
```

peterrum:
Thanks @bangerth! I have modified the message so that it is more similar to the message of `FEValues`.

kronbichler (approved):

PR #11178, opened 2020-11-12
Simplex mesh (2D): implement refinement

peterrum:
This PR adds refinement capabilities to 2D simplex meshes. The code is more or less a copy of the quad implementation, throwing out anisotropy,  and modifying the indices.

Simple example:

ref

~TODO: fix line orientations~


peterrum:
My suggestion would be to completely split up the implementations of the normal and the anisotrop case. In the normal case (99% of the use case), the code path would go through new generalized functions; in the anisotrop case, the old implementation would be used. What do people think?

drwells:
That sounds reasonable to me. Given that anisotropic refinement doesn't really apply to simplices, what would the alternative be?

source/grid/tria.cc
```diff
@@ -4047,6 +4043,479 @@ namespace internal
 
 
 
+      template <int dim, int spacedim>
+      static typename Triangulation<dim, spacedim>::DistortedCellList
+      execute_refinement_isotropic(Triangulation<dim, spacedim> &triangulation,
+                                   const bool check_for_distorted_cells)
```
peterrum, line 4049: As discussed above, I would suggest to have an isotropic and an anisotropic code path. This here is the isotropic code path for refinement of triangles and quadraletrals.

bangerth:
One can have anisotropic refinement also for triangles. I'm not advocating for it, just replying to @drwells 's comment :-)

peterrum:
@drwells @bangerth I don't say that we shouldn't support anisotropy also for simplices in the future (in particular it is very natural for simplices - any volunteers?), however, if we do that I would suggest not to mix up implementations - as it right now: the mix-up makes a code that is in principal quite simple unnecessary hard to read.

source/grid/tria_accessor.cc
```diff
@@ -1496,7 +1496,7 @@ TriaAccessor<structdim, dim, spacedim>::set_bounding_object_indices(
   const ArrayView<int> bounding_object_index_ref =
     this->objects().get_bounding_object_indices(this->present_index);
 
-  AssertDimension(bounding_object_index_ref.size(), new_indices.size());
+  // AssertDimension(bounding_object_index_ref.size(), new_indices.size());
```
tjhei, line 1499: why do you need to remove this?

tests/simplex/refinement_01.cc
```diff
@@ -0,0 +1,70 @@
...
+}
+
+int
+main()
+{
+  initlog();
+  test(0); // no refinement
+  test(1); // refinement of a single cell
+  test(2); // global refinement
+  test(3); // global refinement
```
tjhei, line 69: update comment?

source/grid/tria_accessor.cc
peterrum: I have replaced the assert now by an `AssertIndexRange`. The problem here `riaObjects::get_bounding_object_indices()` returns the maximal possible array size (working for simplices, ..., and for hypercubes).

tests/simplex/refinement_01.cc
peterrum: I have updated the comment.

drwells (approved):

drwells:
This looks good to me. @tjhei are you happy with it?

This code will work for mixed 2D meshes too, right?

peterrum:
> This code will work for mixed 2D meshes too, right?

@drwells I think so. But let me try it out later!

tjhei (approved):

peterrum:
@tjhei @drwells Looks good, doesn't it?

Screenshot from 2020-12-04 19-47-01

... startin with 6 triangles and 1 quad.

I have rebased this branch so that I could use the mixed-mesh capabilities of DataOut, which has been introduced recently. Furthermore, I have extended the test to also test a mixed mesh.

drwells (approved):

PR #11181, opened 2020-11-14
Deprecate MatrixFree::get_dof_handler.

marcfehling:
Introduces `get_dof_handler` with a non-template parameter and deprecates the templated equivalent.

I am not confident on how I implemented this so I need to count on your opinion: Removing the explicit instantiation will throw an error if someone writes `matrix_free.template get_dof_handler<DoFHandler<dim>>();` because of ambiguity. However, this will probably never happen as the default template is way more convenient to use.

Reverts #9216. Part of #10333.

### DIFF
--- a/doc/news/changes/incompatibilities/20201113Fehling
+++ b/doc/news/changes/incompatibilities/20201113Fehling
@@ -1,0 +1,5 @@
+Deprecated: The variant of the function MatrixFree::get_dof_handler
+expecting a DoFHandlerType template has been deprecated. Use the 
+template-less variant returning a DoFHandler instead.
+<br>
+(Marc Fehling, 2020/11/13)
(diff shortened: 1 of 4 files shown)


peterrum (approved):

kronbichler (approved):

marcfehling:
> I am not confident on how I implemented this so I need to count on your opinion: Removing the explicit instantiation will throw an error if someone writes `matrix_free.template get_dof_handler<DoFHandler<dim>>();` because of ambiguity. However, this will probably never happen as the default template is way more convenient to use.

I guess you were okay with that change then :)

kronbichler:
> I guess you were okay with that change then :)

Yes, I am. It was introduced not too long ago, and I believe not many codes have really switched to the templated call, so I'd be happy with it.

PR #11184, opened 2020-11-14
Avoid name macro_cell in matrix-free implementation

kronbichler:
Follow-up to #11183: We used the name `macro_cell` in various instances, which I avoid now. I also tried to unify the name to `cell_batch_index` throughout the code. We might also use `cell_batch`, which would be somewhat shorter and just as verbose. What do others think?


tjhei:
That would have been my follow-up for today as well. Now I don't have to do this work. Thanks!

tjhei (approved):

PR #11186, opened 2020-11-15
Make ShapeInfo compatible with Simplex::FE_P and Simplex::QGauss

peterrum:

PR #11188, opened 2020-11-15
Make FEEvaluationImpl work for ElementType::tensor_none

peterrum:
Depends on #11186. 

This implementation works for now only for a single component. Once we have decided that this is the right approach I will generalize the code.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -137,6 +137,40 @@ namespace internal
 
 
 
+  /**
+   * Specialization for MatrixFreeFunctions::tensor_none, which cannot use the
+   * sum-factorization kernels.
+   */
+  template <int dim, int fe_degree, int n_q_points_1d, typename Number>
+  struct FEEvaluationImpl<MatrixFreeFunctions::tensor_none,
+                          dim,
+                          fe_degree,
+                          n_q_points_1d,
+                          Number>
+  {
+    static void
+    evaluate(const unsigned int                            n_components,
+             const EvaluationFlags::EvaluationFlags        evaluation_flag,
+             const MatrixFreeFunctions::ShapeInfo<Number> &shape_info,
+             const Number *                                values_dofs_actual,
+             Number *                                      values_quad,
+             Number *                                      gradients_quad,
+             Number *                                      hessians_quad,
+             Number *                                      scratch_data);
+
+    static void
+    integrate(const unsigned int                            n_components,
+              const EvaluationFlags::EvaluationFlags        integration_flag,
+              const MatrixFreeFunctions::ShapeInfo<Number> &shape_info,
+              Number *                                      values_dofs_actual,
+              Number *                                      values_quad,
+              Number *                                      gradients_quad,
+              Number *                                      scratch_data,
+              const bool add_into_values_array);
+  };
+
+
+
   template <MatrixFreeFunctions::ElementType type,
             int                              dim,
             int                              fe_degree,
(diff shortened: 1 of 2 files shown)


include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -89,6 +89,13 @@ namespace internal
     static const EvaluatorVariant variant = evaluate_evenodd;
   };
 
+  template <bool is_long>
+  struct EvaluatorSelector<MatrixFreeFunctions::tensor_none, is_long>
+  {
+    static const EvaluatorVariant variant =
+      evaluate_evenodd; // make it compile^^
```
kronbichler, line 96: Since it's definitely not even-odd in that context, I suggest to use `evaluate_general` to not imply any particular structure.

kronbichler (commented):
This looks like the appropriate direction, so please go ahead with the multi-component case.

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -89,6 +89,12 @@ namespace internal
     static const EvaluatorVariant variant = evaluate_evenodd;
   };
 
+  template <bool is_long>
+  struct EvaluatorSelector<MatrixFreeFunctions::tensor_none, is_long>
+  {
+    static const EvaluatorVariant variant = evaluate_none;
+  };
```
peterrum, line 96: here

include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -63,7 +63,12 @@ namespace internal
      * coefficient arrays. See the documentation of the EvaluatorTensorProduct
      * specialization for more information.
      */
-    evaluate_symmetric_hierarchical
+    evaluate_symmetric_hierarchical,
+    /**
+     * Dummy implementation for the case that sum factorization can be
+     * exploited.
+     */
+    evaluate_none
```
peterrum, line 71: here

include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -2503,6 +2508,106 @@ namespace internal
...
+              bool contract_onto_face,
+              bool add,
+              int  max_derivative,
+              bool lex_faces = false>
+    void
+    apply_face(const Number *DEAL_II_RESTRICT, Number *DEAL_II_RESTRICT) const
+    {
+      Assert(false, ExcNotImplemented());
+    }
+  };
```
peterrum, line 2608: here

peterrum (commented):
@kronbichler I have implemented multiple components. I also did not use `evaluate_general` but introduced a new `evaluate_none` with a new dummy `EvaluatorTensorProduct` implementation. I think this is safer.

include/deal.II/matrix_free/tensor_product_kernels.h
kronbichler: Do we ever intend to fill this class? If not, I see little value in providing it and would rather like to see an alternative that avoids those 100 lines, and suggest to either use one of the existing alternatives or do some template magic elsewhere to avoid calling the undefined functions.

peterrum: @kronbichler I have pushed an alternative version. It is nor really based on template magic but only on a template specialization ;)

include/deal.II/matrix_free/evaluation_kernels.h
```diff
@@ -641,6 +675,120 @@ namespace internal
...
+        for (unsigned int c = 0; c < n_components; ++c)
+          for (unsigned int d = 0; d < dim; ++d)
+            for (unsigned int q = 0; q < n_q_points; ++q)
+              for (unsigned int i = 0; i < n_dofs; ++i)
+                values_dofs_actual[n_dofs * dim * c + i] +=
+                  shape_gradients[d * n_dofs * n_q_points + i * n_q_points +
+                                  q] *
+                  gradients_quad[n_q_points * dim * c + n_q_points * d + q];
+      }
+  }
```
peterrum, line 788: @kronbichler What I find "magical" about the current version that everything "simplex" related is centralized and once we should decide to work on more efficient interpolation routines it is clear what to touch!

kronbichler: That's indeed very nice!

kronbichler (approved):

PR #11193, opened 2020-11-17
Better document a bunch of functions.

bangerth:
Continuing through my backlog of places my students encounter and that aren't particularly well documented.

/rebuild


bangerth:
This is repetitive -- we have several functions in several classes that all require slight variations of the same doc string.

masterleinad (approved):
Looks good to me.

PR #11194, opened 2020-11-17
Better document some FEValues functions.

bangerth:
While there, mark some input arguments as 'const'.

Like #11193 .

/rebuild


drwells (approved):
That's a much better description of the intent of the function.

PR #11199, opened 2020-11-18
Add particle update data function to overwrite particle information 

blaisb:
This PR adds a small particle_update_data function to overwrite particle information without re-creating the Particle object from scratch when using the update_ghosts() mechanism. This is extremely beneficial in terms of speed (in some of my benchmarks it gave a speed-up between 20% and 40% depending on the number of core). This was to be expected, because this removes de-allocation and re-allocation of memory at every time-step.
This concludes my series of PR on the update ghosts mechanism started with the immense help of @peterrum ! The deal.II particle library can now be used for Molecular Dynamics and DEM with high efficiency straight away!

### DIFF
--- a/include/deal.II/particles/particle.h
+++ b/include/deal.II/particles/particle.h
@@ -235,6 +235,25 @@ namespace Particles
     void
     write_data(void *&data) const;
 
+
+    /**
+     * Update all of the data associated with a particle : id,
+     * location, reference location and, if any, properties by using a
+     * data array. The array is expected to be large enough to take the data,
+     * and the void pointer should point to the first entry of the array to
+     * which the data should be written. This function is meant for
+     * de-serializing the particle data without requiring that a new Particle
+     * class be built. This is used in the ParticleHandler to update the
+     * ghost particles without de-allocating and re-allocating memory.
+     *
+     * @param[in,out] data A pointer to a memory location from which
+     * to read the information that completely describes a particle. This
+     * class then de-serializes its data from this memory location and
+     * advance the pointer accordingly.
+     */
+    void
+    update_particle_data(const void *&data);
+
     /**
      * Set the location of this particle. Note that this does not check
      * whether this is a valid location in the simulation domain.
--- a/source/particles/particle.cc
+++ b/source/particles/particle.cc
@@ -220,6 +220,34 @@ namespace Particles
   }
 
 
+  template <int dim, int spacedim>
+  void
+  Particle<dim, spacedim>::update_particle_data(const void *&data)
+  {
+    const types::particle_index *id_data =
+      static_cast<const types::particle_index *>(data);
+    id                  = *id_data++;
+    const double *pdata = reinterpret_cast<const double *>(id_data);
+
+    for (unsigned int i = 0; i < spacedim; ++i)
+      location(i) = *pdata++;
+
+    for (unsigned int i = 0; i < dim; ++i)
+      reference_location(i) = *pdata++;
+
+    // See if there are properties to load
+    if (has_properties())
+      {
+        const ArrayView<double> particle_properties =
+          property_pool->get_properties(properties);
+        const unsigned int size = particle_properties.size();
+        for (unsigned int i = 0; i < size; ++i)
+          particle_properties[i] = *pdata++;
+      }
+
+    data = static_cast<const void *>(pdata);
+  }
+
 
   template <int dim, int spacedim>
   std::size_t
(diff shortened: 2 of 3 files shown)


source/particles/particle.cc
```diff
@@ -220,6 +220,34 @@ namespace Particles
   }
 
 
+  template <int dim, int spacedim>
+  void
+  Particle<dim, spacedim>::update_particle_data(const void *&data)
+  {
+    const types::particle_index *id_data =
+      static_cast<const types::particle_index *>(data);
+    id                  = *id_data++;
```
peterrum, line 229: I am a bit confused. As far as I remember we are not communicating the IDs. If we do, we shouldn't because this is information which only needs to be exchanged once during setup. From my point of view, the interface here should look like the following:
```cpp
  void
  Particle<dim, spacedim>::update_particle_data(const numbers::particle_index id, const void *&data)
  {
    AssertDimension(id, this->id);
```
... or maybe even skip the assert totally to avoid a unnecessary data access.

peterrum (commented):
Overall the changes look fine. I have one question/concern.

source/particles/particle.cc
blaisb: That is a good question. Right now we are still communicating the entire particle information. Even though we only use it to update the position and the properties, the data vector contains all the info about a single particle. Consequently, I preferred to keep it general (this however assumes that particle could change ID in their lifetime and etc.). However, I could remove the "id overwritten with itself".  I just felt that since we were already re-building the particle from scratch in the previous PR, this PR would just do the same without messing around with memory. This is not exactly optimal. Right now we are indeed communicating extra information that is not necessary (namely particle id and particle refence location).

peterrum: Let's keep it like this and address this in a follow-up PR.

peterrum (approved):

include/deal.II/particles/particle.h
```diff
@@ -235,6 +235,25 @@ namespace Particles
...
 
+
+    /**
+     * Update all of the data associated with a particle : id,
+     * location, reference location and, if any, properties by using a
+     * data array. The array is expected to be large enough to take the data,
+     * and the void pointer should point to the first entry of the array to
+     * which the data should be written. This function is meant for
+     * de-serializing the particle data without requiring that a new Particle
+     * class be built. This is used in the ParticleHandler, to update the
```
kronbichler, line 246:
```suggestion
     * class be built. This is used in the ParticleHandler to update the
```

include/deal.II/particles/particle.h
```diff
@@ -235,6 +235,25 @@ namespace Particles
...
+     * ghost particles without de-allocating and re-allocating memory.
+     *
+     * @param[in] data A pointer to a memory location from which
+     * to read the information that completely describes a particle. This
+     * class then de-serializes its data from this memory location and
+     * advance the pointer accordingly.
+     */
+    void
+    update_particle_data(const void *&data);
```
kronbichler, line 255: At a closer inspection, I realize that you do update the pointer, which is why you pass it by reference. Could you please mark the param as `@param[in/out]` to highlight this? (It is so easy to get confused by `const`ness of pointers)?

source/particles/particle.cc
```diff
@@ -220,6 +220,34 @@ namespace Particles
...
+    const double *pdata = reinterpret_cast<const double *>(id_data);
```
kronbichler, line 230: Oh mighty, what an accumulation of `reinterpret_cast` in a single function - but I know it's not your fault.

kronbichler (approved):

kronbichler:
> This was to be expected, because this removes de-allocation and re-allocation of memory at every time-step.

This shows that one day we should look into the storage type of particles as a whole, given that it is exactly the `multimap` and associated data types that are leading to fragmented memory and all other kinds of problems.

blaisb:
> > This was to be expected, because this removes de-allocation and re-allocation of memory at every time-step.
> 
> This shows that one day we should look into the storage type of particles as a whole, given that it is exactly the `multimap` and associated data types that are leading to fragmented memory and all other kinds of problems.

I fully agree with you.
I think that storing the particles as a multimap of Particle is a bad idea in the end. It leads to a very large fragmentation of memory which makes all of the operations looping over particles overly expensive. I will apply all of your comments and push again :)!

blaisb:
This should address @kronbichler suggestions :).

peterrum (approved):

kronbichler (approved):

PR #11200, opened 2020-11-18
[CI] OSX tester: change to gcc-10.

luca-heltai:
Try to fix osx build.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -50,7 +50,7 @@ jobs:
         ./contrib/utilities/download_clang_format
     - name: info
       run: |
-        g++-9 -v
+        g++-10 -v
         mpicxx -v
         cmake --version
     - name: configure


peterrum (approved):

.github/workflows/osx.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         ./contrib/utilities/download_clang_format
     - name: info
       run: |
-        g++-9 -v
+        g++-10 -v
```
masterleinad, line 53: Why do we print this at all if we don't use `gcc` here?

luca-heltai: I think we do. for 64bits we use mpicxx (based on gcc). I'd be ok to remove it, but I think it is a good idea to keep in the logs what has been used.

luca-heltai:
the other tests are not necessary I guess. If you want, I can remove the info stage (or make it based on mpicxx instead of gcc-XX).

luca-heltai:
I'll merge. If you want me to make more changes, I'll do this in a followup PR.

PR #11201, opened 2020-11-18
Add GridGenerator::convert_hypercube_to_simplex_mesh

eliasstudiert:
Automatic conversion of non-refined hypercube to simplex mesh. Boundary IDs and material IDs are copied. Manifold IDs are a task for a follow-up PR.

Example (colors are the boundary IDs):

hex_vs_tet

FYI @luca-heltai @blaisb I think this is what we have talked a while ago (months?) :)


include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,28 @@ namespace GridGenerator
...
+   * In 2D: conversion of quad elements into simplices.
+   * In 3D: conversion of hex elements into tet elements.
+   * Material ID and boundary IDs will be inherited after conversion.
+   *
+   * @param in_tria The triangulation containing hex elements.
+   * @param out_tria The converted triangulation containing tet elements.
+   */
+  template <int dim, int spacedim>
+  void
+  convert_hex_to_tet_mesh(const Triangulation<dim, spacedim> &in_tria,
```
peterrum, line 1921:
```suggestion
  convert_hypercube_to_simplex_mesh(const Triangulation<dim, spacedim> &in_tria,
```

include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,28 @@ namespace GridGenerator
   flatten_triangulation(const Triangulation<dim, spacedim1> &in_tria,
                         Triangulation<dim, spacedim2> &      out_tria);
 
+  /**
+   * Convert a hex triangulation into a tet triangulation.
+   * In 2D: conversion of quad elements into simplices.
+   * In 3D: conversion of hex elements into tet elements.
```
peterrum, line 1913:
```suggestion
   * In 2D: conversion of quadrilateral cell into 8 triangle cells.
   * In 3D: conversion of hexadron cell into 24 tetrahedron cells.
```

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,272 @@ namespace GridGenerator
...
+            add_cell(dim, {8, 12, 10, 0}, current_material_id);
+          }
+        else
+          {
+            Assert(false, ExcNotImplemented());
+          }
+
+        // Inherit Boundary_ids
+        for (const auto f : cell.face_indices())
+          {
```
peterrum, line 7257: // Set up sub-cell data.
        for (const auto f : cell.face_indices())
          {
            const auto current_b_id = cell.face(f)->boundary_id()
            if (current_b_id == numbers::internal_face_boundary_id) 
              continue;

peterrum (commented):
I have added some comments. The test seems still be missing?

I also have added a specialization for 1D and moved the namespace of the new function. Don't forget to pull @eliasstudiert!

luca-heltai:
Why `hypercube`? Doesn't this work on other meshes?

peterrum:
> Why hypercube? Doesn't this work on other meshes?

The name is up to debate. I should reflect `quad/hex`-mesh -> `tri/tet`-mesh.

luca-heltai:
but a quad/hex is not a `hyper_cube`... :)

luca-heltai:
There is a `tet_hex` app. I think `hex_tet` is appropriate. :)

blaisb:
> Automatic conversion of non-refined hypercube to simplex mesh. Boundary IDs and material IDs are copied. Manifold IDs are a task for a follow-up PR.
> 
> Example (colors are the boundary IDs):
> 
> hex_vs_tet
> 
> FYI @luca-heltai @blaisb I think this is what we have talked a while ago (months?) :)

There are many ways to split an hex into a tet (I think 121 or something like that) is there a reason why this specific strategy was chosen? (Awesome work btw)

eliasstudiert:
> I have added some comments. The test seems still be missing? Could you also create a change-log entry. Thx.
> 
> I also have added a specialization for 1D and moved the namespace of the new function. Don't forget to pull @eliasstudiert!

How should the change-log entry look like? And where/ in which file should it be located?

masterleinad:
> How should the change-log entry look like? And where/ in which file should it be located?

Have a look at the files in `doc/news/changes/minor`.

luca-heltai:
How about `GridGenerator::hex_or_quad_to_simplex_mesh`?

tjhei:
GridGenerator::hex_to_simplex_mesh ?
I would think "hex mesh" is acceptable to use in 2d and 3d

peterrum:
I would be fine with `GridGenerator::hex_to_simplex_mesh` if all Mathematicians agree on this name - in particular @bangerth.

The current name `GridGenerator::convert_hypercube_to_simplex_mesh` would be consistent with the following (new) functions:

https://github.com/dealii/dealii/blob/e4ad46dc23fecadc80f4e015b18db2ab8e2fbb8c/include/deal.II/grid/reference_cell.h#L71-L92

https://github.com/dealii/dealii/blob/e4ad46dc23fecadc80f4e015b18db2ab8e2fbb8c/include/deal.II/grid/tria.h#L3370-L3375

Are there as well objections to the name of these?

bangerth:
I think hypercube is a good enough name. Or `convert_quad_hex_to_simplex_mesh`? Or maybe `convert_to_simplex_mesh`?

I'm not feeling dogmatic about the name. I think the name @eliasstudiert chose is fine and fits the terminology used in other contexts. People will scroll through the documentation of namespace `GridGenerator` and see the pictures (@eliasstudiert : I hope you plan to copy the picture at the top of the PR into the documentation!), and intuitively understand what the function does regardless of the exact name.

@eliasstudiert : Excellent work, by the way! This would be the kind of PR I'd like to read through to learn about this myself -- alas, not this semester :-( Next semester will be so much better!

eliasstudiert:
> > Automatic conversion of non-refined hypercube to simplex mesh. Boundary IDs and material IDs are copied. Manifold IDs are a task for a follow-up PR.
> > Example (colors are the boundary IDs):
> > hex_vs_tet
> > FYI @luca-heltai @blaisb I think this is what we have talked a while ago (months?) :)
> 
> There are many ways to split an hex into a tet (I think 121 or something like that) is there a reason why this specific strategy was chosen? (Awesome work btw)

Well, this is what Peter suggested to me :) This way we definitely take advantage of symmetry I guess.

eliasstudiert:
I've added a ctest now. I'm not sure is it right the way I did. Especially, do we need to test dim=3, spacedim=2 as well?

bangerth:
`dim=3, spacedim=2` can't work :-) But the opposite would be nice!

masterleinad:
Apparently, you need to fix the indentation. :slightly_smiling_face:

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+            Assert(false, ExcNotImplemented());
+          }
+
+        // Set up sub-cell data.
+        for (const auto f : cell.face_indices())
+          {
+            const auto current_b_id = cell.face(f)->boundary_id();
+            if (current_b_id == numbers::internal_face_boundary_id) 
+              continue;
+            if (dim == 2) // 2D boundary_lines
```
peterrum, line 7261:
```suggestion
            const auto current_boundary_id = cell.face(f)->boundary_id();

            if (current_boundary_id == numbers::internal_face_boundary_id) 
              continue;

            if (dim == 2) // 2D boundary_lines
```

include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,28 @@ namespace GridGenerator
...
+   * @param in_tria The triangulation containing hex elements.
+   * @param out_tria The converted triangulation containing tet elements.
+   */
+  template <int dim, int spacedim>
+  void
+  convert_hypercube_to_simplex_mesh(const Triangulation<dim, spacedim> &in_tria,
+                          Triangulation<dim, spacedim> &      out_tria);
+
+  /**
+   * Specialization for 1D: simply copy triangulation.
```
peterrum, line 1925:
```suggestion
   * Specialization of the above function for 1D: simply copy triangulation.
```

peterrum (commented):
Some minor comments from my side. Also the change-log entry is missing. Once these are done this PR is ready from my side.

peterrum:
> There are many ways to split an hex into a tet (I think 121 or something like that) is there a reason why this specific strategy was chosen? (Awesome work btw)

> Well, this is what Peter suggested to me :) This way we definitely take advantage of symmetry I guess.

@blaisb It is possible to subdivide quadralterals/hexahedrons with less triangles/tetrahedrons (2/5), however, these versions might introduce some anisotropy in the solution (I think we discussed this offline when I first showed you the DG results). In the most basic case of a subdivided hyper-rectangle it would be possible switch the order of cells in a checkerboard-like style to prevent this issue. However, this is not possible in the general case. Furthermore in 3D, it is necessary to access neighboring cells and to query their their orientation since not all faces match between neighboring. Latter point would make the algorithm significantly more complicated. The current algorithm really takes at one cell at a time!

In the future, we could implement additional conversion strategies and configure the function via an enum. But for now I think the current conversion routine is enough.

eliasstudiert:
> Apparently, you need to fix the indentation. slightly_smiling_face

Where exactly?

masterleinad:
The `indent` CI check is failing:
https://github.com/dealii/dealii/pull/11201/checks?check_run_id=1444096507

peterrum:
According to the indentation check the following files are off:
- include/deal.II/grid/grid_generator.h
- source/grid/grid_generator.cc
- tests/simplex/conv_hex_to_simplex.cc

peterrum:
@eliasstudiert What is the state here? I would like to use this feature soon!

peterrum:
> Add testcase dim = 2, spacedim = 3 (subdivided_hyper_cube triangulation).

What you want to do is a function specialization:

```cpp
void create_triangulation(Triangulation<dim, spacedim> & triangulation)
{
  GridGenerator::subdivided_hyper_cube(triangulation);
}

template<int dim>
void create_triangulation(Triangulation<dim, dim> & triangulation)
{
  GridGenerator::quarter_hyper_ball(triangulation);
}
```

and then you call it like this:
```cpp
check_file()
{
  Triangulation<dim, spacedim> in_tria, out_tria;
  create_triangulation(in_tria);
  // ...
}
```

tests/simplex/conv_hex_to_simplex.cc
```diff
@@ -39,56 +39,114 @@ using namespace dealii;
...
+
+  grid_out(out_tria, "tri.vtk");               // total mesh
+  grid_out(out_tria, "tri_surface.vtk", true); // only surface mesh
+
+  deallog << "OK!" << std::endl;
+}
+
+template <int dim, int spacedim>
+void
+check_file_dim_smaller_spacedim() // for dim < spacedim
```
peterrum, line 98: Please don't do this. I have posted a way how to do it.

tests/simplex/conv_hex_to_simplex.cc
```diff
@@ -0,0 +1,134 @@
+/* ---------------------------------------------------------------------
+ *
+ * Copyright (C) 1999 - 2019 by the deal.II authors
```
peterrum, line 3:
```suggestion
 * Copyright (C) 2020 by the deal.II authors
```

tests/simplex/conv_hex_to_simplex.cc
```diff
@@ -0,0 +1,134 @@
...
+              b_id++;
+            }
+        }
+    }
+
+  GridGenerator::convert_hypercube_to_simplex_mesh(in_tria, out_tria);
+
+  // write 2 outputs (total mesh and only surface mesh)
+  const auto grid_out = [](const auto &       tria,
+                           const std::string &label,
```
peterrum, line 85: The label is not used. You can remove it.

peterrum (approved):
I have two minor points. Overall, looks good now! 

Is there another volunteer to review the code? I am kinda biased since I am a co-author ;)

include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,31 @@ namespace GridGenerator
   flatten_triangulation(const Triangulation<dim, spacedim1> &in_tria,
                         Triangulation<dim, spacedim2> &      out_tria);
 
+  /**
+   * Convert a triangulation consisting only of hypercube cells
+   * (quadrilaterals, hexahedrons) to a triangulation only consisting of
```
bangerth, line 1912:
```suggestion
   * (quadrilaterals, hexahedra) to a triangulation only consisting of
```

include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,31 @@ namespace GridGenerator
...
+   * simplices (triangles, tetrahedrons).
```
bangerth, line 1913:
```suggestion
   * simplices (triangles, tetrahedra).
```

include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,31 @@ namespace GridGenerator
...
+   * In 2D: conversion of quadrilateral cell into 8 triangle cells.
+   * In 3D: conversion of hexadron cell into 24 tetrahedron cells.
```
bangerth, line 1915: Can you make these two lines into complete sentences?

bangerth (commented):
More tomorrow.

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+                                              numbers::invalid_unsigned_int);
+    std::vector<unsigned int> indices_faces(in_tria.n_faces(),
+                                            numbers::invalid_unsigned_int);
+
+    // temporary array storing the global indices of each cell entity in the
+    // sequence: vertices, edges/faces, cell
+    std::array<unsigned int, dim == 2 ? 9 : 14> indices_local;
+
+    // loop over all cells
+    for (const auto &cell : in_tria)
```
bangerth, line 7118: Here and below, you have a number of comments that just duplicate what the next line does. I can see that you're looping over all cells from the next line, but what I don't know is *why* you're doing that and *what* you are doing this for. A good comment in front of the `for` loop would be something like this:

We first have to create all of the new vertices. To do this, we loop over all cells and on each cell (i) copy the existing vertex locations (and record their new indices in the 'indices_vertics' [replace by new name] array), (ii) create new midpoint vertex locations for each face (and record their new indices in 'indices_faces' [again, replace by new name], (iii) create new midpoint vertex locations for each cell.

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+
+    Assert(dim > 1, ExcNotImplemented());
+
+    std::vector<Point<spacedim>> vertices;
+    std::vector<CellData<dim>>   cells;
+    SubCellData                  subcell_data;
+
+    // store for each vertex and face the assigned index so that we only
+    // assign them a value only once
+    std::vector<unsigned int> indices_vertics(in_tria.n_vertices(),
```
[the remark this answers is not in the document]
bangerth: On second thought, and reading through the code that follows, what you are doing with this array is really a translation from the old to the new vertex indices. This should be reflected in the name of the variable. I think that a good name for this variable would be `old_to_new_vertex_indices`, and `face_to_new_vertex_indices` for the next variable.

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+                                              numbers::invalid_unsigned_int);
+    std::vector<unsigned int> indices_faces(in_tria.n_faces(),
+                                            numbers::invalid_unsigned_int);
+
+    // temporary array storing the global indices of each cell entity in the
+    // sequence: vertices, edges/faces, cell
+    std::array<unsigned int, dim == 2 ? 9 : 14> indices_local;
```
bangerth, line 7115: You don't need to know this: `std::array` objects are allocated on the stack, not the heap. So there is no cost to moving the declaration of this variable into the loop over all cells. Doing so makes it clear that (i) this variable does not carry state from one loop iteration to the next, (ii) it's clear that the variable should not be used after the end of the loop body. Both of these things are quite useful when reading through code at a later point.

Separately, the variable stores *vertex* indices. Maybe name it `local_vertex_indices`?

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+              {
+                CellData<dim> cell_data(index_vertices.size());
+                for (unsigned int i = 0; i < index_vertices.size(); ++i)
+                  {
+                    cell_data.vertices[i] = indices_local[index_vertices[i]];
+                    cell_data.material_id = material_or_boundary_id;
+                  }
+                cells.push_back(cell_data);
+              }
+            else if (dim == 2 && struct_dim == 1)
```
bangerth, line 7171:
```suggestion
            else if (dim == 2 && struct_dim == 1)  // an edge of a triangle
```

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+              {
+                CellData<1> boundary_line(2);
+                boundary_line.boundary_id = material_or_boundary_id;
+                for (unsigned int i = 0; i < index_vertices.size(); i++)
```
bangerth, line 7175:
```suggestion
                for (unsigned int i = 0; i < index_vertices.size(); ++i)
```

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+                  {
+                    boundary_line.vertices[i] =
+                      indices_local[index_vertices[i]];
+                  }
+                subcell_data.boundary_lines.push_back(boundary_line);
+              }
+            else if (dim == 3 && struct_dim == 2)
```
bangerth, line 7182:
```suggestion
            else if (dim == 3 && struct_dim == 2) // a face of a tetrahedron
```

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+              {
+                CellData<2> boundary_quad(3);
+                boundary_quad.boundary_id = material_or_boundary_id;
+                for (unsigned int i = 0; i < index_vertices.size(); i++)
```
bangerth, line 7186:
```suggestion
                for (unsigned int i = 0; i < index_vertices.size(); ++i)
```

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+                      {
+                        add_cell(1, {0, 6}, current_boundary_id);
+                        add_cell(1, {6, 1}, current_boundary_id);
+                      }
+                    else if (f == 3)
+                      {
+                        add_cell(1, {2, 7}, current_boundary_id);
+                        add_cell(1, {7, 3}, current_boundary_id);
+                      }
+                  }
```
bangerth, line 7286: There's an
```
  else
    Assert (false, ExcInternalError());
```
case missing here.

This whole sequence of if-else-if statements could have been written as a switch-case statement as well, and that's the style we generally use. (Same for the 3d case below)

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,273 @@ namespace GridGenerator
...
+          {
+            Assert(false, ExcNotImplemented());
+          }
+
+        // Set up sub-cell data.
+        for (const auto f : cell.face_indices())
+          {
+            const auto current_boundary_id = cell.face(f)->boundary_id();
+            if (current_boundary_id == numbers::internal_face_boundary_id)
+              continue;
```
bangerth, line 7260: It is easier to read if you write this as
```
  for (const auto f : ...)
    if (cell.face(f)->boundary_id() != numbers::internal_face_boundary_id)
      {
         ...
```
This way it is clear at the top of the loop which faces will actually be processed.

source/grid/grid_generator.cc
bangerth: Alternatively, you could have just put the various indices into one big table of the form
```
  const unsigned int face_to_new_vertex_indices[4][2][2] = ...;
```
and then just call
```
  add_cell (1, face_to_new_vertex_indices[f][0], current_boundary_id);
  add_cell (1, face_to_new_vertex_indices[f][1], current_boundary_id);
```

bangerth:
@eliasstudiert -- nice work! I left a number of comments; none of them about the overall algorithm, which is well done and where I don't have any comments, but mostly on style issues regarding naming of variables and similar things: one of the things I've learned (the hard way!) is that we write code not for compilers but for humans who read the code several years later. So things like the naming of variables, where they are declared etc, really matter because they make it easier to understand what the code does. An example is calling an array variable `indices` -- what are they an index to? What is the array indexed by? Maybe better to call it `old_to_new_vertex_indices` -- that's longer to write (autocompletion in your IDE helps with that) but it's now clear that the elements of the array are the new *vertex* indices and that the array is index by the *old* vertex indices. I can now also reason how large that array will be without having to look it up at the place of its declaration: It must have as many entries as there were old vertices. 

These are all things you'll learn over time, and our jobs over here is to help you learn this sort of stuff and mentor you along the way!

include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,31 @@ namespace GridGenerator
...
 
+  /**
+   * Convert a triangulation consisting only of hypercube cells
+   * (quadrilaterals, hexahedrons) to a triangulation only consisting of
+   * simplices (triangles, tetrahedra).
+   * Thereby, in 2D a quadrilateral cell is converted into 8 triangle cells,
+   * whereas in 3D a hexahedron cell is converted into 24 tetrahedra cells.
+   *
+   * Material ID and boundary IDs will be inherited after conversion.
+   *
```
bangerth, line 1918: For the documentation, it would be nice to have the picture you have at the top of the PR in here as well. To do this, add the .png to `doc/doxygen/images` and reference it like this: https://github.com/dealii/dealii/blob/master/include/deal.II/lac/solver.h#L325
which will then show up like this (scroll down to the pictures):
https://dealii.org/developer/doxygen/deal.II/classSolverBase.html

doc/news/changes/minor/20201123DejeneMunch
```diff
@@ -0,0 +1,3 @@
+New: The function GridGenerator::convert_hypercube_to_simplex_mesh allows to convert a given triangulation based on quadrilaterals (2D) or hexaedrons (3D) to a triangulation based on simplices or tetraedrons, respectively. Thereby, material_IDs and boundary_IDs are inherited from the given triangulation.
```
masterleinad, line 1: PLease split this line at around 80 characters.

source/grid/grid_generator.cc
```diff
@@ -7090,6 +7090,283 @@ namespace GridGenerator
...
+        if (dim == 2)
+          {
+            add_cell(dim, {0, 6, 4}, current_material_id);
+            add_cell(dim, {8, 4, 6}, current_material_id);
+            add_cell(dim, {8, 6, 5}, current_material_id);
+            add_cell(dim, {1, 5, 6}, current_material_id);
+            add_cell(dim, {2, 4, 7}, current_material_id);
+            add_cell(dim, {8, 7, 4}, current_material_id);
+            add_cell(dim, {8, 5, 7}, current_material_id);
+            add_cell(dim, {3, 7, 5}, current_material_id);
```
peterrum, line 7223: I like @bangerth' idea: https://github.com/dealii/dealii/pull/11201/commits/de3ee8db347a517de5ef5995d114fa5bc2c45444#r532703430

Could you convert these types of codes to:
```cpp
            std::array<std::array<unsigned int, 8>, 3> table = {{0,6,4},{8,4,6}, ...};
            for(const auto & cell : table)
              add_cell(dim, table, current_material_id);
```
and in similar the face definitions.

include/deal.II/grid/grid_generator.h
bangerth: This is still open.

include/deal.II/grid/grid_generator.h
```diff
@@ -1907,6 +1907,34 @@ namespace GridGenerator
   flatten_triangulation(const Triangulation<dim, spacedim1> &in_tria,
                         Triangulation<dim, spacedim2> &      out_tria);
 
+  /**
+   * Convert a triangulation consisting only of hypercube cells
+   * (quadrilaterals, hexahedrons) to a triangulation only consisting of
+   * simplices (triangles, tetrahedra).
+   *
+   * @image convert_hypercube_to_simplex_mesh_visualization.png
```
bangerth: You'll have to write this as follows. I'm also adding a suggestion for some more words around it:
```suggestion
   * As an example, the following image shows how a set of three hexahedra
   * meshing one eighths of a sphere are subdivided into tetrahedra, and how
   * the curved surface is taken into account. Colors indicate how boundary indicators
   * are inherited:
   * @image html "convert_hypercube_to_simplex_mesh_visualization.png"
```

source/grid/grid_generator.cc
bangerth: I assume @peterrum meant
```
            constexpr std::array<std::array<unsigned int, 8>, 3> table = {{0,6,4},{8,4,6}, ...};
            for(const auto & cell : table)
              add_cell(dim, cell, current_material_id);
```

bangerth (commented):
Very close!

eliasstudiert:
There are plenty of warnings due to the initialization of the arrays regarding braces. Is there a way to get rid of those?

masterleinad:
> There are plenty of warnings due to the initialization of the arrays regarding braces. Is there a way to get rid of those?

If it's `clang` it normally tells you where to put extra ones for it to be happy.

peterrum (commented):
@eliasstudiert I have updated the code so that it does not give any warning, furthermore, I have moved some code around so that it is formatted more nicely. 

Regarding the picture. I don't think @bangerth had this in mind. Rather something like this:

Screenshot from 2020-12-02 11-27-10

I have created the picture with the following code:
```latex
\documentclass[10pt,a4paper,landscape]{article}
\usepackage[utf8]{inputenc}
\usepackage[english]{babel}
\usepackage{amsmath}
\usepackage{amsfonts}
\usepackage{amssymb}
\usepackage{tikz}
\usepackage{transparent}

\begin{document}

%\documentclass[crop,tikz]{standalone}% 'crop' is the default for v1.0, before it was 'preview'
%\usetikzlibrary{...}% tikz package already loaded by 'tikz' option
%\begin{document}
\begin{tikzpicture}[scale=2.5]
\coordinate (0) at (0,0,1);
\coordinate (1) at (1,0,1);
\coordinate (2) at (0,0,0);
\coordinate (3) at (1,0,0);

\coordinate (4) at (0,1,1);
\coordinate (5) at (1,1,1);
\coordinate (6) at (0,1,0);
\coordinate (7) at (1,1,0);


\coordinate (8) at (0,0.5,0.5);
\coordinate (9) at (1,0.5,0.5);
\coordinate (10) at (0.5,0.5,1);
\coordinate (11) at (0.5,0.5,0);
\coordinate (12) at (0.5,0,0.5);
\coordinate (13) at (0.5,1,0.5);

\foreach \i in {0,1,2,3, 8, 9,11,12}
    \fill (\i) circle (1pt) node [below] {\i};

\foreach \i in {4,5,6,7,10,13}
    \fill (\i) circle (1pt) node [above] {\i};

\draw (0) -- (1) -- (3) -- (2) -- (0);
\draw (4) -- (5) -- (7) -- (6) -- (4);
\draw (0) -- (4);
\draw (1) -- (5);
\draw (2) -- (6);
\draw (3) -- (7);
\end{tikzpicture}
\begin{tikzpicture}[scale=2.5]
\coordinate (0) at (0,0,1);
\coordinate (1) at (1,0,1);
\coordinate (2) at (0,0,0);
\coordinate (3) at (1,0,0);

\coordinate (4) at (0,1,1);
\coordinate (5) at (1,1,1);
\coordinate (6) at (0,1,0);
\coordinate (7) at (1,1,0);


\coordinate (8) at (0,0.5,0.5);
\coordinate (9) at (1,0.5,0.5);
\coordinate (10) at (0.5,0.5,1);
\coordinate (11) at (0.5,0.5,0);
\coordinate (12) at (0.5,0,0.5);
\coordinate (13) at (0.5,1,0.5);

\draw[-, fill=red!30, opacity=.5] (0)--(1)--(12)--cycle;
\draw[-, fill=red!30, opacity=.5] (0)--(1)--(10)--cycle;
\draw[-, fill=red!30, opacity=.5] (0)--(12)--(10)--cycle;
\draw[-, fill=red!30, opacity=.5] (1)--(12)--(10)--cycle;

\foreach \i in {0,1,2,3, 8, 9,11,12}
    \fill (\i) circle (1pt) node [below] {};

\foreach \i in {4,5,6,7,10,13}
    \fill (\i) circle (1pt) node [above] {};

\draw (0) -- (1) -- (3) -- (2) -- (0);
\draw (4) -- (5) -- (7) -- (6) -- (4);
\draw (0) -- (4);
\draw (1) -- (5);
\draw (2) -- (6);
\draw (3) -- (7);

\end{tikzpicture}
\begin{tikzpicture}[scale=2.5]
\coordinate (0) at (0,0,1);
\coordinate (1) at (1,0,1);
\coordinate (2) at (0,0,0);
\coordinate (3) at (1,0,0);

\coordinate (4) at (0,1,1);
\coordinate (5) at (1,1,1);
\coordinate (6) at (0,1,0);
\coordinate (7) at (1,1,0);


\coordinate (8) at (0,0.5,0.5);
\coordinate (9) at (1,0.5,0.5);
\coordinate (10) at (0.5,0.5,1);
\coordinate (11) at (0.5,0.5,0);
\coordinate (12) at (0.5,0,0.5);
\coordinate (13) at (0.5,1,0.5);

\draw[-, fill=red!30, opacity=.5] (9)--(12)--(11)--cycle;
\draw[-, fill=red!30, opacity=.5] (9)--(12)--(3)--cycle;
\draw[-, fill=red!30, opacity=.5] (9)--(11)--(3)--cycle;
\draw[-, fill=red!30, opacity=.5] (12)--(11)--(3)--cycle;

\foreach \i in {0,1,2,3, 8, 9,11,12}
    \fill (\i) circle (1pt) node [below] {};

\foreach \i in {4,5,6,7,10,13}
    \fill (\i) circle (1pt) node [above] {};

\draw (0) -- (1) -- (3) -- (2) -- (0);
\draw (4) -- (5) -- (7) -- (6) -- (4);
\draw (0) -- (4);
\draw (1) -- (5);
\draw (2) -- (6);
\draw (3) -- (7);

\end{tikzpicture}


\end{document}

%\end{document}
```
maybe you could complete it for 2D and 3D.

bangerth:
Actually, I did have the picture @eliasstudiert put into the documentation in mind. But more pictures are more better, and since you already have this other picture, I think it would be nice to have that in the documentation as well!

eliasstudiert (commented):
[https://gist.github.com/eliasstudiert/ed0268d5e54c9ba1c84dcf51fd1bda26](url)

tests/simplex/conv_hex_to_simplex.cc
```diff
@@ -37,7 +42,13 @@ void
 check_file()
 {
   Triangulation<dim, spacedim> in_tria, out_tria;
-  GridGenerator::quarter_hyper_ball(in_tria);
+  // if (dim == spacedim) // test dim = spacedim
```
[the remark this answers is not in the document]
peterrum: I don't get your question. As far a I see it all relevant dim/spacedim pairs are checked (2/2, 2/3, 3/3)!?

peterrum (commented):
I think we could move on here. We are at least one follow-up PR (manifolds).

doc/news/changes/minor/20201123DejeneMunch
```diff
@@ -0,0 +1,6 @@
+New: The function GridGenerator::convert_hypercube_to_simplex_mesh allows to
+convert a given triangulation based on quadrilaterals (2D) or hexaedrons (3D)
+to a triangulation based on simplices or tetraedrons, respectively. Thereby,
```
kronbichler, line 3: here is one additional comment similar to the one by @bangerth:
```suggestion
convert a given triangulation based on quadrilaterals (2D) or hexaedra (3D)
to a triangulation based on simplices or tetraedra, respectively. Thereby,
```

kronbichler (approved):
This is nice work, so I suggest to fix the final parts and merge soon.

peterrum:
@eliasstudiert Could you revert the merge. The PR was already more or less ready!

eliasstudiert (commented):
[REDACTED]

kronbichler (approved):

bangerth:
Good work! Here's the new function already: https://dealii.org/developer/doxygen/deal.II/namespaceGridGenerator.html#ac7515d2b17c025dddc0e37286fb8d216

I think you changed the picture at some point, and the text doesn't match any more. But I'll fix this in a follow-up -- for now thanks for a job well done (and having had a lot of patience with all of our requests)!

PR #11202, opened 2020-11-18
Improve MGTransferMatrixFree::interpolate_to_mg

kronbichler:
In particular, we should use the vector partitioners already in the MG transfer class.

In a run with 25k MPI ranks, I noticed that this function takes an unbearable amount of time. It seems that we waste the time in the setup of the partitioner (which might be a recent regression due to the shared memory stuff, but I have not checked more closely; it could have also been my setup). In any case, we should simply pick the partitioners we have already set up for the purpose of intergrid transfer like prolongation or restriction.

### DIFF
--- a/include/deal.II/multigrid/mg_transfer_matrix_free.h
+++ b/include/deal.II/multigrid/mg_transfer_matrix_free.h
@@ -503,28 +503,6 @@ MGTransferMatrixFree<dim, Number>::interpolate_to_mg(
          ExcDimensionMismatch(
            max_level, dof_handler.get_triangulation().n_global_levels() - 1));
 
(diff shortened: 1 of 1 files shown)


tamiko (approved):

peterrum (approved):

peterrum:
> which might be a recent regression due to the shared memory stuff,

I don't think so. This should not take the sm code path.

kronbichler:
>> which might be a recent regression due to the shared memory stuff,
>
> I don't think so. This should not take the sm code path.

Right, that seems unlikely. But it's still intriguing that the old `interpolate_to_mg` took around 4 seconds per call, whereas the new version takes less than 0.01 seconds on 25k ranks and with around 512m unknowns (so a small size for this problem).

PR #11203, opened 2020-11-18
Add simplex elements to the MF permitted list.

drwells:
A brief followup to #11186.


kronbichler (approved):

peterrum (approved):
Yes, indeed. Thanks! @drwells How did you notice this? This function is not called anywhere, is it?

kronbichler:
> This function is not called anywhere, is it?

It is: https://github.com/dealii/dealii/blob/203108ee27ca1bd04eed10bd1be0907bd352e8c0/include/deal.II/numerics/vector_tools_project.templates.h#L707-L711

PR #11207, opened 2020-11-19
Use hp::MappingCollection initialize_cell_range

peterrum:
... depends on #11209.

### DIFF
--- a/include/deal.II/matrix_free/mapping_info.templates.h
+++ b/include/deal.II/matrix_free/mapping_info.templates.h
@@ -846,15 +846,12 @@ namespace internal
         const dealii::Triangulation<dim> &                        tria,
         const std::vector<std::pair<unsigned int, unsigned int>> &cells,
         const std::vector<unsigned int> &              active_fe_index,
-        const dealii::hp::MappingCollection<dim> &     mapping_in,
+        const dealii::hp::MappingCollection<dim> &     mapping,
         MappingInfo<dim, Number, VectorizedArrayType> &mapping_info,
         std::pair<std::vector<
                     MappingInfoStorage<dim, dim, Number, VectorizedArrayType>>,
                   CompressedCellData<dim, Number, VectorizedArrayType>> &data)
       {
-        AssertDimension(mapping_in.size(), 1);
-        const auto &mapping = mapping_in[0];
-
         FE_Nothing<dim> dummy_fe;
 
         // when we make comparisons about the size of Jacobians we need to
@@ -920,13 +917,15 @@ namespace internal
               const unsigned int hp_quad_index =
                 mapping_info.cell_data[my_q].descriptor.size() == 1 ? 0 :
                                                                       fe_index;
+              const unsigned int hp_mapping_index =
+                mapping.size() == 1 ? 0 : fe_index;
               const unsigned int n_q_points = mapping_info.cell_data[my_q]
                                                 .descriptor[hp_quad_index]
                                                 .n_q_points;
               if (fe_values[my_q][fe_index].get() == nullptr)
                 fe_values[my_q][fe_index] =
                   std::make_shared<dealii::FEValues<dim>>(
-                    mapping,
+                    mapping[hp_mapping_index],
                     dummy_fe,
                     mapping_info.cell_data[my_q]
                       .descriptor[hp_quad_index]
@@ -1076,8 +1075,9 @@ namespace internal
                               cells[cell * VectorizedArrayType::size() + v]
                                 .second);
                           const Point<dim> p =
-                            mapping.transform_unit_to_real_cell(cell_it,
-                                                                Point<dim>());
+                            mapping[hp_mapping_index]
+                              .transform_unit_to_real_cell(cell_it,
+                                                           Point<dim>());
                           for (unsigned int d = 0; d < dim; ++d)
                             quad_point[d][v] = p[d];
                         }


source/hp/fe_values.cc
```diff
@@ -279,14 +279,27 @@ namespace hp
...
 
     if (real_mapping_index == numbers::invalid_unsigned_int)
-      real_mapping_index = 0;
+      {
+        if (real_fe_index == numbers::invalid_unsigned_int ||
+            this->mapping_collection->size() == 1)
+          real_mapping_index = 0;
+        else
+          real_mapping_index = fe_index;
+      }
```
peterrum, line 302: I can move this to separate PR.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -900,6 +901,37 @@ namespace internal
...
+        std::vector<std::shared_ptr<dealii::hp::FEValues<dim>>> fe_values(
+          mapping_info.cell_data.size());
+
+        for (unsigned int my_q = 0; my_q < mapping_info.cell_data.size();
+             ++my_q)
+          fe_values[my_q] =
+            std::make_shared<dealii::hp::FEValues<dim>>(mapping_collection,
+                                                        fe_collection,
+                                                        q_collections[my_q],
+                                                        update_flags_feval);
```
peterrum, line 933: I think it would be appropriate to use a `hp::FEValues` here, since in a nutshell we will be using mapping/fe/quadrature collections in here in the near future. `hp::FEValues` contains the right logic how to select the right `mapping` and `quadrature` (0 or `active_fe_index`) so I don't want to repeat the logic here and at other places.

Setting up the collections (code above) should not be necessary if they are passed here or stored like this at appropriate places. But cleaning up this could be part of a follow-up PR.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -670,7 +670,19 @@ namespace internal
           {
             typename dealii::Triangulation<dim>::cell_iterator cell_it(
               &tria, cells[j].first, cells[j].second);
-            fe_val.reinit(cell_it);
+            hp_fe_val.reinit(cell_it,
+                             numbers::invalid_unsigned_int,
+                             numbers::invalid_unsigned_int,
+                             active_fe_index);
```
peterrum, line 676: Unfortunately, we only have access here to the triangulation and not the dofhandler, so we need write it manually here.

peterrum (commented):
Some remarks:

include/deal.II/matrix_free/mapping_info.templates.h
kronbichler: But there is no `DoFHandler` and `FiniteElement` in this evaluation at all, so to me this looks like a bug of `hp::FEValues` when initialized with a tria cell iterator and it should be fixed there. That said, you could probably just pass in 0 for the active fe index because there really should not be any.

kronbichler: To be more specific, shouldn't we use the active fe index for the quadrature formula and mapping object, rather than the "fe" index, see also the argument list here: https://github.com/dealii/dealii/blob/e4ad46dc23fecadc80f4e015b18db2ab8e2fbb8c/include/deal.II/hp/fe_values.h#L367-L378

peterrum: I must admit I would prefer to have the same decision logic in the `reinit()` function that takes a iterator form a triangulation and that takes a iterator from a dofhandler. 

```cpp
  void
  FEValues<dim, spacedim>::reinit(
    const unsigned int                                          q_index,
    const unsigned int                                          mapping_index,
    const unsigned int                                          fe_index)
  {
    // determine which indices we
    // should actually use
    unsigned int real_q_index       = q_index;
    unsigned int real_mapping_index = mapping_index;
    unsigned int real_fe_index      = fe_index;


    if (real_q_index == numbers::invalid_unsigned_int)
      {
        if (real_fe_index == numbers::invalid_unsigned_int ||
            this->q_collection.size() == 1)
          real_q_index = 0;
        else
          real_q_index = fe_index;
      }


    if (real_mapping_index == numbers::invalid_unsigned_int)
      {
        if (real_fe_index == numbers::invalid_unsigned_int ||
            this->mapping_collection->size() == 1)
          real_mapping_index = 0;
        else
          real_mapping_index = fe_index;
      }


    if (real_fe_index == numbers::invalid_unsigned_int)
      real_fe_index = 0;


    // some checks
    AssertIndexRange(real_q_index, this->q_collection.size());
    AssertIndexRange(real_mapping_index, this->mapping_collection->size());
    AssertIndexRange(real_fe_index, this->fe_collection->size());


    // now finally actually get the
    // corresponding object and
    // initialize it
    this->select_fe_values(real_fe_index, real_mapping_index, real_q_index)
      .reinit(cell);
  }
```
Of course we could have this logic also in `MatrixFree` (as it is currently the case for cells for the quadrature rules, where we select either quadrature index `0` or `active_fe_index`). But since we now also add `MappingQCollection` and will need the same logic for faces I would suggest to handle this at a central place.

kronbichler: I am not generally opposed to having this logic, but to me it completely goes against intuition to make automatic choices about some indices of an `FEValues` object that is not associated with a `DoFHandler` and a genuine `FiniteElement`. At the very least, we would need some careful comments in the `reinit` function and at the place of usage in `MatrixFree` because nobody will remember what happens without looking at the source code.

In my opinion, it would be much better to slightly break symmetry with the version taking `DoFHandler:cell_iterators` to make deductions for say the quadrature index instead, because that is an index that actually makes sense. But I admit that even that choice has a chance of getting misunderstood, so this would need comments as well.

Either way, I would be happy to get convinced of another way.

kronbichler:
Another question: What happens in the case if we have multiple `DoFHandler` objects with different `active_fe_index` settings in `MatrixFree`? Right now I think we pick index 0, but I think we should add an assertion that either all `DoFHandler` objects have the same `active_cell_index` setting (to make the right guess about the quadrature formula) or only active_fe_index 0. It would not make sense in my opinion to support all cross combinations.

peterrum:
> In my opinion, it would be much better to slightly break symmetry with the version taking DoFHandler:cell_iterators to make deductions for say the quadrature index instead, because that is an index that actually makes sense.

We could pass one dofhandler to `MappingInfo` and construct based on that the iterator. The difficulty here will probably be that we have to distinguish between mg and non-mg mode (template argument!).

> Another question: What happens in the case if we have multiple DoFHandler objects with different active_fe_index settings in MatrixFree? Right now I think we pick index 0, but I think we should add an assertion that either all DoFHandler objects have the same active_cell_index setting (to make the right guess about the quadrature formula) or only active_fe_index 0. It would not make sense in my opinion to support all cross combinations.

According to 

https://github.com/dealii/dealii/blob/e4ad46dc23fecadc80f4e015b18db2ab8e2fbb8c/include/deal.II/matrix_free/matrix_free.templates.h#L449-L451

The active fe indices of the first dofhandler is used.

peterrum:
@kronbichler  I have pushed an alternative version without `hp::FEValues` (last commit). I think this a most painless approach for now as I only want to get cell integrals running for mixed meshes. I will revisit the issue once I also have a look at face integrals and their setup routines. 

I have created a back up of the old version: https://github.com/peterrum/dealii/tree/matrixfree_initialize_cell_range_hp_backup

peterrum:
I have rebased this PR.

kronbichler (approved):

PR #11208, opened 2020-11-19
Fix tests with MGTransferMatrixFree::interpolate_to_mg

kronbichler:
I forgot to properly check the tests #11202, so do that now. This means that I have to add back some code that I deleted because we must be able to set up the vectors. As a consequence, I had to fix two things that did not work in the tests.


peterrum (approved):
Thanks!

tamiko (approved):

PR #11209, opened 2020-11-19
Introduce MappingCollection in MatrixFree

peterrum:


include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -843,12 +843,15 @@ namespace internal
...
         const std::vector<std::pair<unsigned int, unsigned int>> &cells,
         const std::vector<unsigned int> &              active_fe_index,
-        const Mapping<dim> &                           mapping,
+        const dealii::hp::MappingCollection<dim> &     mapping_in,
         MappingInfo<dim, Number, VectorizedArrayType> &mapping_info,
         std::pair<std::vector<
                     MappingInfoStorage<dim, dim, Number, VectorizedArrayType>>,
                   CompressedCellData<dim, Number, VectorizedArrayType>> &data)
       {
+        AssertDimension(mapping_in.size(), 1);
```
peterrum, line 852: For now, I am asserting that the collection only has one entry.

include/deal.II/matrix_free/mapping_info.h
```diff
@@ -446,7 +447,7 @@ namespace internal
       /**
        * The pointer to the underlying Mapping object.
        */
-      SmartPointer<const Mapping<dim>> mapping;
+      SmartPointer<const dealii::hp::MappingCollection<dim>> mapping;
```
peterrum, line 450: I really don't like this `SmartPointer` here. Would it be possible to eliminate it?

kronbichler: This was introduced back in #8856 and for good reasons, as one should be able to query the information underlying the object. It technically goes through an internal class but we really must expose this information one way or the other. By the way, it is also this way in `FEValues`.

Thus, we should try a little harder than simply eliminate it (I see why you want to, though). My idea would be to utilize `Mapping::clone()` and store an `std::unique_ptr` with the mapping instead. While the type changes, all typical use cases would be equivalent as it is still a pointer type. Since some mappings have a relatively high memory consumption, we should look over the most important ones, say `MappingQGeneric`, and see if we can make some big members like the interpolation tables `std::shared_ptr` to share them via `MappingQGeneric::clone()`.

peterrum: @kronbichler I went with `std::shared_ptr<dealii::hp::MappingCollection<dim>>`.

include/deal.II/matrix_free/mapping_info.h
```diff
@@ -446,7 +447,7 @@ namespace internal
       /**
        * The pointer to the underlying Mapping object.
        */
-      SmartPointer<const Mapping<dim>> mapping;
+      std::shared_ptr<dealii::hp::MappingCollection<dim>> mapping;
```
kronbichler, line 450: There is one minor issue with it - but unrelated to whether we use a `shared_ptr` here or a `unique_ptr` that we would get by cloning: This breaks backward compatibility. While this is an internal class, this is still the way user codes would access the mapping, so I would like to be cautious here. Maybe call the new member `mapping_collection`, to be specific with the name, and initialize the `mapping` as a `SmartPointer` from that? It is a bit redundant but it would be nice if user codes do not break.

peterrum: You mean the smart pointer points to the first entry of the collection? I can do that.

kronbichler: Yes exactly, so whenever we get a new shared pointer of mapping collections in a initialize/update_mapping function of this class, we update the smart pointer.

peterrum: Done!

kronbichler (approved):
Very nice

PR #11213, opened 2020-11-19
Fix bug in constructor of FEEvaluationBaseData

peterrum:
... four instead of five indices have been passed to `MatrixFree::get_shape_info()` with a middle index being skipped.

### DIFF
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -3266,20 +3266,27 @@ inline FEEvaluationBaseData<dim, Number, is_face, VectorizedArrayType>::
                                      .cell_data[quad_no_in]
                                      .quad_index_from_n_q_points(n_q_points)) :
                         0)
-  , n_quadrature_points(fe_degree != numbers::invalid_unsigned_int ?
-                          n_q_points :
-                          (is_face ? data_in
-                                       .get_shape_info(dof_no,
-                                                       quad_no_in,
-                                                       active_fe_index,
-                                                       active_quad_index)
-                                       .n_q_points_face :
-                                     data_in
-                                       .get_shape_info(dof_no,
-                                                       quad_no_in,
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -3266,20 +3266,27 @@ inline FEEvaluationBaseData<dim, Number, is_face, VectorizedArrayType>::
...
-                                       .n_q_points))
+  , n_quadrature_points(
+      fe_degree != numbers::invalid_unsigned_int ?
+        n_q_points :
+        (is_face ? data_in
+                     .get_shape_info(
+                       dof_no,
+                       quad_no_in,
+                       data_in.get_dof_info(dof_no)
+                         .component_to_base_index[first_selected_component],
```
peterrum, line 3277: This was missing.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -3266,20 +3266,27 @@ inline FEEvaluationBaseData<dim, Number, is_face, VectorizedArrayType>::
...
+                       active_fe_index,
+                       active_quad_index)
+                     .n_q_points_face :
+                   data_in
+                     .get_shape_info(
+                       dof_no,
+                       quad_no_in,
+                       data_in.get_dof_info(dof_no)
+                         .component_to_base_index[first_selected_component],
```
peterrum, line 3286: ... as well as this one.

kronbichler (approved):
The reason why this never triggered was that we have not used complicated setups for the hp case. In the general non-hp case which is well-tested, this was not actually a bug returning a wrong value because the number of quadrature points does not depend on the dof handler index, so always using the index 0 was fine.

That said, this is the right thing to do. The alternative would have been to ask the mapping info which should also have this info but not need the dof handler information for the quadrature points.

PR #11215, opened 2020-11-20
Remove hp::DoFHandler instantiation from SolutionTransfer.

marcfehling:
I've noticed that the `hp::DoFHandler` instantiation from the basic `SolutionTransfer` class has been removed in #10328.

So this should also apply for the `p::d::SolutionTransfer` class, right?

### DIFF
--- a/source/numerics/solution_transfer.inst.in
+++ b/source/numerics/solution_transfer.inst.in
@@ -22,9 +22,5 @@ for (VEC : VECTOR_TYPES; deal_II_dimension : DIMENSIONS;
       deal_II_dimension,
       VEC,
       DoFHandler<deal_II_dimension, deal_II_space_dimension>>;
-    template class SolutionTransfer<
-      deal_II_dimension,
-      VEC,
-      hp::DoFHandler<deal_II_dimension, deal_II_space_dimension>>;
 #endif
   }


marcfehling:
Part of #10333.

marcfehling:
Oops, I just realized it's the other way round. The basic class still has the instantiation, while the `p::d` equivalent doesn't.

drwells (approved):

peterrum (commented):
The following test complain about undefined references.
```
The following tests FAILED:
	3135 - hp/solution_transfer_12.debug (Failed)
	3141 - hp/solution_transfer_10.debug (Failed)
	3145 - hp/solution_transfer_07.debug (Failed)
	3152 - hp/do_function_laplacians_01.debug (Failed)
	3161 - hp/solution_transfer_13.debug (Failed)
	3190 - hp/solution_transfer_09.debug (Failed)
	3199 - hp/solution_transfer_16.debug (Failed)
	3257 - hp/solution_transfer_11.debug (Failed)
	3263 - hp/solution_transfer_08.debug (Failed)
	3268 - hp/solution_transfer_06.debug (Failed)
	3279 - hp/solution_transfer_04.debug (Failed)
	3285 - hp/solution_transfer_05.debug (Failed)
	3311 - hp/do_function_hessians_01.debug (Failed)
	3348 - hp/do_function_derivatives_01.debug (Failed)
	3357 - hp/solution_transfer.debug (Failed)
	3360 - hp/solution_transfer_03.debug (Failed)
```

Could we remove the `DoFHandlerType` template argument here and have a dummy (deprecated) specialization, which takes a `hp::DoFHandler`, in the header file?

marcfehling:
Oh wow I didn't see that `DoFHandler` in the `p::d::SolTrans` instantiation refers to the loop variable, not the actual class.
https://github.com/dealii/dealii/blob/46729a41be0ef3a35862323c16db6767772064f2/source/distributed/solution_transfer.inst.in#L18

Well, then it makes sense that we need to keep the `hp::DoFHandler` instantiation some way.

---

These explicit instantiation will be removed in #11221, so I would like to leave this as it is for now.

PR #11214, opened 2020-11-20
Updated documentation to avoid hp::DoFHandler.

marcfehling:
While browsing through the documentation, I've found duplicates of functions in header files that I removed.

So although this PR effectively just changes the documentation, please run the testsuite to be sure nothing breaks.

PR #11216, opened 2020-11-20
Consistency in using hp concatenation symbols.

marcfehling:
~~Blocked by #11214.~~

Part of #10333.

---

In the deal.II hp paper and the the original hp papers by Babuska and Suri, any term regarding hp adaptation was written in the following style with a space in between:
```
hp adaptation, h refinement, ...
```

In more recent papers, you find a hyphen for concatenation. (I also used this stye in my dissertation.)
```
hp-adaptation, h-refinement, ...
```

Right now in the library, we use both versions inconsistently. This PR is an attempt to adapt to the **original** nomenclature.

Additionally, I removed the underscores from the `active_fe_index` terms in the documentation.

---

This is an entirely cosmetic change to the documentation, and there might be other opinions on which version to choose.

Especially for @bangerth: What are your thoughts on that?

### DIFF
--- a/include/deal.II/distributed/solution_transfer.h
+++ b/include/deal.II/distributed/solution_transfer.h
@@ -153,9 +153,9 @@ namespace parallel
      * @endcode
      *
      *
-     * <h3>Note on usage with DoFHandler with hp capabilities</h3>
+     * <h3>Note on usage with DoFHandler with hp-capabilities</h3>
      *
-     * Since data on DoFHandler objects with hp capabilities is associated with
+     * Since data on DoFHandler objects with hp-capabilities is associated with
      * many different FiniteElement objects, each cell's data has to be
      * processed with its corresponding `future_fe_index`. Further, if
      * refinement is involved, data will be packed on the parent cell with its
@@ -166,11 +166,11 @@ namespace parallel
      * hp::FECollection::find_dominated_fe_extended() for more information).
      *
      * Transferring a solution across refinement works exactly like in the
-     * non-hp case. However, when considering serialization, we also have to
+     * non-hp-case. However, when considering serialization, we also have to
      * store the active_fe_indices in an additional step. A code snippet
      * demonstrating serialization with the
      * parallel::distributed::SolutionTransfer class with DoFHandler objects
-     * with hp capabilities is provided in the following. Here VectorType is
+     * with hp-capabilities is provided in the following. Here VectorType is
      * your favorite vector type, e.g. PETScWrappers::MPI::Vector,
      * TrilinosWrappers::MPI::Vector, or corresponding block vectors.
      *
--- a/include/deal.II/distributed/solution_transfer.h
+++ b/include/deal.II/distributed/solution_transfer.h
@@ -167,7 +167,7 @@ namespace parallel
      *
      * Transferring a solution across refinement works exactly like in the
      * non-hp-case. However, when considering serialization, we also have to
-     * store the active_fe_indices in an additional step. A code snippet
+     * store the active FE indices in an additional step. A code snippet
      * demonstrating serialization with the
      * parallel::distributed::SolutionTransfer class with DoFHandler objects
      * with hp-capabilities is provided in the following. Here VectorType is
@@ -197,7 +197,7 @@ namespace parallel
      *
      * DoFHandler<dim,spacedim> hp_dof_handler(triangulation);
      * // We need to introduce our dof_handler to the fe_collection
-     * // before setting all active_fe_indices.
+     * // before setting all active FE indices.
      * hp_dof_handler.deserialize_active_fe_indices();
      * hp_dof_handler.distribute_dofs(fe_collection);
      *
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -605,7 +605,7 @@ namespace internal
 
 
       /**
-       * Return the fe index of the n-th finite element active on a given
+       * Return the FE index of the n-th finite element active on a given
        * vertex.
        */
       template <int dim, int spacedim, int d>
@@ -618,7 +618,7 @@ namespace internal
       {
         Assert(d == dim || obj_level == 0, ExcNotImplemented());
 
-        // for cells only one active fe index available
+        // for cells only one active FE index available
         Assert(((d == dim) &&
                 (local_index != DoFHandler<dim, spacedim>::default_fe_index)) ==
                  false,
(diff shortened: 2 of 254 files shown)


peterrum (commented):
> h adaptivity -> h adaptivity

I am in favor of `h-adaptivity`,  `p-adaptivity`, and `hp-adaptivity` as it is also frequently used in the multigrid context.

> active_fe_indices -> active fe indices

I don't have any particular opinion here. But shouldn't fe than written in all caps?

bangerth:
We probably need a native speaker here. @drwells @jppelteret ?

I *think* that the grammatically proper resolution would probably be to use the following:
* "h-adaptive refinement" (in the same style as "truck-related accidents" -- it's a compound adjective: https://www.gingersoftware.com/content/grammar-rules/adjectives/compound-adjectives/)
* "h adaptivity", where "h" is an adjective to the noun "adaptivity".

Same of course for "p-adaptive", "hp-adaptive", "p adaptivity", "hp adaptivity".

include/deal.II/distributed/solution_transfer.h
```diff
@@ -157,17 +157,17 @@ namespace parallel
      *
      * Since data on DoFHandler objects with hp capabilities is associated with
      * many different FiniteElement objects, each cell's data has to be
-     * processed with its corresponding `future_fe_index`. Further, if
+     * processed with its corresponding `future fe index`. Further, if
```
bangerth, line 160: You're referencing a technical term here, which is used in function and variable names. I think it should be written with underscores if you mark it up with the backticks. If you want spaces, remove the backticks.

bangerth: Same everywhere else

include/deal.II/distributed/solution_transfer.h
```diff
@@ -157,17 +157,17 @@ namespace parallel
...
+     * `future fe index` and unpacked later with the same index on its children.
      * If cells get coarsened into one, data will be packed on the children with
      * the least dominant finite element of their common subspace, and unpacked
      * on the parent with this particular finite element (consult
      * hp::FECollection::find_dominated_fe_extended() for more information).
      *
      * Transferring a solution across refinement works exactly like in the
      * non-hp case. However, when considering serialization, we also have to
-     * store the active_fe_indices in an additional step. A code snippet
+     * store the active fe indices in an additional step. A code snippet
```
bangerth, line 170: this seems correct then.

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -2905,13 +2905,13 @@ DoFCellAccessor<dimension_, space_dimension_, level_dof_access>::
...
          ExcMessage(
-           "You can not ask for the future_fe_index on a cell that has "
+           "You can not ask for the future fe index on a cell that has "
            "children because no degrees of freedom are assigned "
            "to this cell and, consequently, no finite element "
            "is associated with it."));
   Assert((this->dof_handler->hp_capability_enabled == false) ||
            (this->is_locally_owned()),
-         ExcMessage("You can only query future_fe_index information on cells "
+         ExcMessage("You can only query future fe index information on cells "
```
bangerth, line 2914: Here and elsewhere, I wonder whether if you already remove the underscore, we should improve readability by saying "(active|future) FE index" (with fe capitalized). This makes it clear that this part of the expression is supposed to be an abbreviation.

include/deal.II/dofs/dof_handler.h
```diff
@@ -214,7 +214,7 @@ namespace parallel
  * serialization archive.
  *
  *
- * <h3>hp-adaptive finite element methods</h3>
+ * <h3>hp adaptive finite element methods</h3>
```
bangerth, line 217: I think it should stay here

include/deal.II/fe/fe_enriched.h
```diff
@@ -1035,7 +1035,7 @@ namespace ColorEnriched
    *
    * @warning The current implementation relies on assigning each cell a
    * material id, which shall not be modified after the setup
-   * and h-adaptive refinement. For a given cell, the material id is used
+   * and h adaptive refinement. For a given cell, the material id is used
```
bangerth, line 1038: and here

include/deal.II/fe/fe_series.h
```diff
@@ -50,7 +50,7 @@ DEAL_II_NAMESPACE_OPEN
  * This namespace offers functions to calculate expansion series of the
  * solution on the reference element. Coefficients of expansion are often used
  * to estimate local smoothness of the underlying FiniteElement field to decide
- * on h- or p-adaptive refinement strategy.
+ * on h or p adaptive refinement strategy.
```
bangerth, line 53: and here

and probably a few other places

bangerth (changes requested):

drwells:
Both work but I am in favor of `hp-adaptivity` et al.

marcfehling:
After thinking about it, I decided to use the `hp-` variant in any possible case. This also involves the following which look a bit odd, but I prefer them over the alternative.
```
hp-finite element or hp-FEM
hp-DoFHandlers and non-hp-DoFHandlers
hp-classes and non-hp-classes
```

Further, I changed any standalone `fe` in the documentation to `FE`.

marcfehling:
For completion: The search and replace scripts I used to prepare this PR.
```
perl -p -i -e 's/(^\s*(\/\/|\*).*\W)(?<!\@ref )(?<!\@defgroup )(?<!\@ingroup )(?<!namespace )hp( |$)/\1hp-/g' `find ./ -type f`
perl -p -i -e 's/(^\s*(\/\/|\*).*\W)(?<!\.)h( |$)/\1h-/g' `find ./ -type f`
perl -p -i -e 's/(^\s*(\/\/|\*).*\W)(?<!@)p( |$)/\1p-/g' `find ./ -type f`

perl -p -i -e 's/(^\s*(\/\/|\*).*\W)fe( |$)/\1FE\2/g' `find ./ -type f`
```

`GNU sed` does not allow lookahead/lookbehind, so I had to use pearl here.

jppelteret:
> Both work but I am in favor of hp-adaptivity et al.

I'm super late to the party, but I also lean towards hyphenating all of

> h-adaptivity, p-adaptivity, and hp-adaptivity

Irrespective of whether of not its grammatically correct, its certainly commonplace amongst the scientific literature (and [Wikipedia page](https://en.wikipedia.org/wiki/Hp-FEM), if that means anything at all). So I don't think that we'd be "wrong" in adopting this notation. 

> "h adaptivity", where "h" is an adjective to the noun "adaptivity".

@bangerth I can't think of any rebuttal to this because it seems totally correct, but I need some eye floss after reading this :-D I've always found it odd to see some hovering "h" or "p" or "hp" in a sentence, but that is of course indicative of my personal taste.

marcfehling:
Ready for review.

Please also run the testsuite on this PR as some exception messages have also changed, which are caught in a few tests.

bangerth (approved):

PR #11218, opened 2020-11-20
Better initial guess for Chebyshev eigenvalue estimate with BlockVector

kronbichler:
In an application, @tamiko and I observed that `PreconditionChebyshev` can use very bad eigenvalue estimates when working with `LinearAlgebra::distributed::BlockVector`. But that case is easy to fix, we should just let each block use the initial guess we use for non-block vectors.


masterleinad (approved):
Looks reasonable.

peterrum (approved):

PR #11219, opened 2020-11-21
Deprecated DoFHandlerType in DataOutStack.

marcfehling:
Part of #10333.

Similar to ##11190.


drwells (approved):
This is a clever use of partial specialization.

marcfehling:
> This is a clever use of partial specialization.

All credit belongs to @masterleinad ;-)

PR #11224, opened 2020-11-21
Add hp capabilities to Simplex::FE_P

peterrum:


source/simplex/fe_lib.cc
```diff
@@ -239,6 +241,65 @@ namespace Simplex
...
+  FE_P<dim, spacedim>::compare_for_domination(
+    const FiniteElement<dim, spacedim> &fe_other,
+    const unsigned int                  codim) const
+  {
+    (void)fe_other;
+    (void)codim;
+
+    Assert((dynamic_cast<const FE_Q<dim, spacedim> *>(&fe_other)),
+           ExcNotImplemented());
+    AssertDimension(dim, 2);
```
kronbichler, line 256: I think it makes sense to indicate what the potential problem is, namely the impossibility to set up an interpolation between a 3D FE_Q and a 3D FE_P due to the different types of faces. How about this message:
```suggestion
    Assert(dim == 2, ExcNotImplemented("hp comparision not possible for dim=" + std::to_string(dim)));
```

kronbichler (commented):
I think we should be more specific in the comments that only allow `dim=2` as to what happens.

source/simplex/fe_lib.cc
peterrum: Let's keep this assert. It will be remove quite soon anyway (in the next 1-2 weeks when wedges and pyramids are coming).

kronbichler (approved):
Given that this will be overhauled soon, I am fine with the current state.

PR #11226, opened 2020-11-21
Use EvaluatorTensorProduct in Simplex code path

peterrum:
This PR simplifies the simplex matrix-free evaluation kernels by *misusing* `EvaluatorTensorProduct<evaluate_general, 1, 0, 0, Number, Number>` for performing a dense matrix-vector product or a transposed dense matrix-vector product. These kernels should work decently for small degrees and small number of quadrature points - as we need it for now. Once we go to higher degrees, we might  need to use different dense-matrix-vector product kernels: maybe BLAS :P

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -696,35 +696,50 @@ namespace internal
     const unsigned int n_dofs     = shape_info.dofs_per_component_on_cell;
     const unsigned int n_q_points = shape_info.n_q_points;
 
+    using Eval =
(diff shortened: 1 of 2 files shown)


kronbichler (approved):
This makes sense; at some point one could experiment with passing a `ShapeInfo<double>` in when the `Number` type is `VectorizedArray<double>`, i.e., multiplying several right hand sides of vectorized arrays with a scalar interpolation matrix. At some point in the past, I did experiments with such a setup for the tensor product case and it was surprisingly competitive (0-10% slowdown, which is not bad considering that every matrix entry as to be broadcast to all lanes), whereas I could imagine that here it would even be a win.

PR #11230, opened 2020-11-23
Provide a serial implementation of ConsensusAlgorithms::Interface

peterrum:
Fixes the failing tests:

>base/mpi_noncontiguous_partitioner_02.mpirun=1.debug
base/mpi_noncontiguous_partitioner_02.mpirun=1.release
multigrid-global-coarsening/mg_transfer_a_01.mpirun=1.debug
multigrid-global-coarsening/mg_transfer_a_01.mpirun=1.release
multigrid-global-coarsening/mg_transfer_a_02.mpirun=1.debug
multigrid-global-coarsening/mg_transfer_a_02.mpirun=1.release

The check `if (this->n_procs > 1)` introduced in #11105 was not correct because there places in the library where we expect that the consensus algorithm is run even for a single process. This PR introduces a serial fall back.

### DIFF
--- a/include/deal.II/base/mpi_consensus_algorithms.h
+++ b/include/deal.II/base/mpi_consensus_algorithms.h
@@ -191,6 +191,11 @@ namespace Utilities
         const MPI_Comm &comm;
 
         /**
+         * Cache if job supports MPI.
+         */
+        const bool job_supports_mpi;
+
+        /**
          * Rank of this process.
          */
         const unsigned int my_rank;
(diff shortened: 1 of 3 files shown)


include/deal.II/base/mpi_consensus_algorithms.h
```diff
@@ -440,6 +445,34 @@ namespace Utilities
...
+         *
+         * @param process Process to be run during consensus algorithm.
+         * @param comm MPI Communicator (ignored)
+         */
+        Serial(Process<T1, T2> &process, const MPI_Comm &comm);
+
+        /**
+         * Destructor.
+         */
+        virtual ~Serial() = default;
```
kronbichler, line 467: Do we need this constructor? I believe the virtual destructor in the base class `Interface` should be enough.

include/deal.II/base/mpi_consensus_algorithms.h
```diff
@@ -440,6 +445,34 @@ namespace Utilities
         clean_up_and_end_communication();
       };
 
+      /**
+       * A serial fall back for the above classes to allow programming
+       * independently if MPI is used or not.
```
kronbichler, line 450: Even though I am not a native English speaker, I think this should be 
```suggestion
       * independently of whether MPI is used or not.
```

kronbichler (approved):
Looks good, thank you for addressing the tests.

include/deal.II/base/mpi_consensus_algorithms.h
peterrum: I have removed. Let's see if one of the compilers complain...

include/deal.II/base/mpi_consensus_algorithms.h
peterrum: Done!

kronbichler (approved):
Great!

PR #11231, opened 2020-11-23
Make tests/simplex/step-18 more robust

peterrum:
Fixes:

> simplex/step-18.mpirun=1.debug
simplex/step-18.mpirun=1.release

Instead of printing the number of iterations, print the L2 norm of the solution.

### DIFF
--- a/tests/simplex/step-18.mpirun=1.with_petsc=false.with_simplex_support=on.output
+++ b/tests/simplex/step-18.mpirun=1.with_petsc=false.with_simplex_support=on.output
@@ -4,91 +4,61 @@ DEAL::  Cycle 0:
 DEAL::    Number of active cells:       1875 (by partition: 1875)
 DEAL::    Number of degrees of freedom: 1728 (by partition: 1728)
 DEAL::    Assembling system... norm of rhs is 2.45750e+10
-DEAL:cg::Starting value 1.95453e+10
-DEAL:cg::Convergence step 126 value 2.13799e-06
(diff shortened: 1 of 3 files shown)


tests/simplex/step-18.mpirun=1.with_petsc=false.with_simplex_support=on.output
```diff
@@ -5,90 +5,90 @@ DEAL::    Number of active cells:       1875 (by partition: 1875)
 DEAL::    Number of degrees of freedom: 1728 (by partition: 1728)
 DEAL::    Assembling system... norm of rhs is 2.45750e+10
 DEAL:cg::Starting value 1.95453e+10
-DEAL:cg::Convergence step 126 value 2.13799e-06
+DEAL:cg::Convergence step 126 value 2.30962e-06
```
kronbichler, line 8: Is this value below the threshold of `numdiff`? I thought we have an absolute tolerance of 1e-8, so this seems still dangerous. To be safe, I would suggest to disable this logstream output. More importantly, this still seems to print the iteration count.

kronbichler (changes requested):

kronbichler (approved):

tjhei:
You can use the ``check_solver_within_range`` macro in tests as well.

peterrum:
@tjhei Thanks for the info! I will use it the next time. Let's keep it now how it is.

Issue #11234, opened 2020-11-23
Don't use 'import' as a member function


bangerth:
`import` has become a keyword in C++20. Curiously, the authors of the standard have chosen it to be a "context-sensitive keyword"  so that it only acts as something related to modules if at the global scope and in certain situations where it could not have been used as the name of a class or function. One may have philosophical qualms about this, but it is what it is.

We should nevertheless think about whether we want to continue using this name. I suspect that the use of the keyword would confuse IDEs and, possibly, readers. I found ~90 uses of the word in our code base (see below). Some should be relatively easy to address if they are just local variables, but some are also member functions -- specifically in the parallel vector interfaces. Maybe we could name these functions `import_elements()`?

(There is the related question of using `module`, which is also a context-sensitive keyword in C++20. But we don't seem to be using this name outside the documentation.)

List of places where we seem to be using `import`:
```
include/deal.II/lac/trilinos_epetra_communication_pattern.h:      std::unique_ptr<Epetra_Import> import;
include/deal.II/lac/la_parallel_block_vector.templates.h:          rw_vector.import(trilinos_vec.block(i), VectorOperation::insert);
include/deal.II/lac/la_parallel_block_vector.templates.h:          this->block(i).import(rw_vector, VectorOperation::insert);
include/deal.II/lac/la_parallel_block_vector.templates.h:    BlockVector<Number>::import(const LinearAlgebra::ReadWriteVector<Number> &,
include/deal.II/lac/trilinos_vector.h:      import(const LinearAlgebra::ReadWriteVector<double> &rwv,
include/deal.II/lac/cuda_vector.h:      import(
include/deal.II/lac/trilinos_tpetra_vector.templates.h:    Vector<Number>::import(
include/deal.II/lac/la_parallel_vector.h:      import(const Vector<Number, MemorySpace2> &src,
include/deal.II/lac/la_parallel_vector.h:      import(
include/deal.II/lac/vector_operations_internal.h:      import(
include/deal.II/lac/vector_operations_internal.h:      import(const std::shared_ptr<::dealii::parallel::internal::TBBPartitioner>
include/deal.II/lac/vector_operations_internal.h:      import(const std::shared_ptr<::dealii::parallel::internal::TBBPartitioner>
include/deal.II/lac/vector_operations_internal.h:      import(const std::shared_ptr<::dealii::parallel::internal::TBBPartitioner>
[...]

PR #11237, opened 2020-11-24
Avoid using 'import' as function name in internal interfaces.

bangerth:
See #11234 . These are functions in an `internal` namespace, so there should be no problem with renaming them.

/rebuild


masterleinad (approved):
I don't really see a need for this as of now but it also doesn't hurt.

kronbichler (approved):

PR #11238, opened 2020-11-24
Simplify some code with stuff from later C++ standards.

bangerth:
No need to re-invent wheels. It would be nice if we could have used std::is_base_of,
but that doesn't work because we don't know for sure which ones of the
possible BlockVectorBase<T> classes is the right base class, given that we don't
know T.

/rebuild


masterleinad (approved):
Looks OK.

PR #11240, opened 2020-11-24
Multigrid: Only call MGLevelObject::resize() if necessary

kronbichler:
This safes some memory allocations in case the vectors were already the right size before.


tamiko (approved):

masterleinad:
```
The following tests FAILED:
	3813 - multigrid/cycles.debug (Failed)
```

kronbichler:
I fixed the failing test locally.

tjhei:
Awesome. This can make a big difference for computations with a very large number of ranks!

Issue #11222, opened 2020-11-21
Remove `hp::DoFHandler` and `DoFHandlerType`.


marcfehling:
The ultimate goal of #10328 and #10333.

After introducing legacy classes (#11220) in the next 9.3 release, we might think of finally removing all the `hp::DoFHandler` leftovers in the following release.

---

For the 9.4 release, we need to consider the following:
- [x] Update template arguments to avoid `DoFHandlerType` and deprecate classes in `Legacy` namespace
#11221
- [x] Update tests to avoid `DoFHandlerType`
~#11241~, #11221
- [x] Remove `DOFHANDLER_TEMPLATE` from `template-arguments.in`
#11221

---

For the 9.5 release, let's get rid of:
- [ ] Remove all classes and functions with `DoFHandlerType` templates (including those in the `Legacy` namespace)
#13808
- [ ] Remove `hp::DoFHandler`
#13808

PR #11221, opened 2020-11-21
Replaced template argument DoFHandlerType by spacedim for classes ...

marcfehling:
... SolutionTransfer, FEFieldFunction, DataOut, DataOut_DoFData, DataOutFaces, DataOutRotation.

Part of #10333, #11222.

Meant for the next major release 10.0

After the deprecation of the `hp::DoFHandler`, this pull request changes the template arguments of a couple of classes to avoid the legacy `DoFHandlerType` as follows:
```
SolutionTransfer<dim, VectorType, DoFHandlerType>
  -> SolutionTransfer<dim, VectorType, spacedim>

parallel::distributed::SolutionTransfer<dim, VectorType, DoFHandlerType>
  -> parallel::distributed::SolutionTransfer<dim, VectorType, spacedim>

Functions::FEFieldFunction<dim, DoFHandlerType, VectorType>
  -> Functions::FEFieldFunction<dim, VectorType, spacedim>

DataOut<dim, DoFHandlerType>
  -> DataOut<dim, spacedim>

DataOutFaces<dim, DoFHandlerType>
  -> DataOutFaces<dim, spacedim>

DataOutRotation<dim, DoFHandlerType>
  -> DataOutRotation<dim, spacedim>

DataOut_DoFData<DoFHandlerType, patch_dim, patch_space_dim>
  -> DataOut_DoFData<dim, patch_dim, spacedim, patch_spacedim>
```

PR #11241, opened 2020-11-24
Replaced template argument DoFHandlerType in testsuite.

marcfehling:
Part of #10333, #11222. Blocked by #11221.

Meant for the next major release 10.0

This adjusts the testsuite to use the new interface.


marcfehling:
Now part of #11221.

PR #11244, opened 2020-11-25
Correct poor naming of a class.

bangerth:
Specifically, the IsBlockMatrix class was also used to ask whether a class is
a block sparsity pattern. That's poor programming style, but easy to fix by
duplicating the class.

/rebuild

### DIFF
--- a/include/deal.II/lac/affine_constraints.h
+++ b/include/deal.II/lac/affine_constraints.h
@@ -2280,8 +2280,8 @@ template <typename number>
 class BlockSparseMatrixEZ;
 
 /**
- * A class that can be used to determine whether a given type is a block
- * matrix or block sparsity pattern type or not. For example,
+ * A "traits" class that can be used to determine whether a given type is a
+ * block matrix type or not. For example,
  * @code
  *   IsBlockMatrix<SparseMatrix<number> >::value
  * @endcode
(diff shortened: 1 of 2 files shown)


tjhei:
```
lac/is_block_matrix.debug: DIFF failed. ------ First 20 lines of numdiff output:
----------------
##5       #:2   <== 1
##5       #:2   ==> 0
```

include/deal.II/lac/affine_constraints.h
```diff
@@ -2309,14 +2309,6 @@ struct IsBlockMatrix
   static std::true_type
   check(const BlockMatrixBase<T> *);
 
-  /**
-   * Overload returning true if the class is derived from
-   * BlockSparsityPatternBase, which is what block sparsity patterns do.
-   */
-  template <typename T>
-  static std::true_type
-  check(const BlockSparsityPatternBase<T> *);
```
masterleinad, line 2318: You could rather deprecate this.

bangerth: It's a private member function, so there is no point in deprecating it.

I think that this function is actually in the `dealii` namespace by accident. It's only used in 4 repeated places where we dispatch to different functions. We would likely put it into an internal namespace today, and that was my plan for a follow-up patch. It is true that this PR changes the semantics of the class (it was poorly named because it does two things at the same time, and I'm breaking it up into two classes). One could argue that I should deprecate the old class, but like I said above: I think the class was where it is by accident, and I can't imagine that it was actually used anywhere outside that file.

bangerth:
I've also fixed the failing test.

kronbichler (approved):

peterrum (approved):

bangerth:
Follow-up to #11243.

PR #11246, opened 2020-11-25
MatrixFree: assert that in hp-mode only a single DoFHandler is used

peterrum:
... or alternatively check that `DoFInfo::cell_active_fe_index` is same for each `DoFHandler`.

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.templates.h
+++ b/include/deal.II/matrix_free/matrix_free.templates.h
@@ -442,6 +442,26 @@ MatrixFree<dim, Number, VectorizedArrayType>::internal_reinit(
   // determined in @p extract_local_to_global_indices.
   if (additional_data.initialize_mapping == true)
     {
+      if (dof_handler.size() > 1)
+        {
+          // check if all DoHandlers are in the same hp mode; and if hp
+          // capabilities are enabled: check if active_fe_indices of all
+          // DoFHandler are the same.
+          for (unsigned int i = 1; i < dof_handler.size(); ++i)
+            {
+              Assert(dof_handler[0]->has_hp_capabilities() ==
+                       dof_handler[i]->has_hp_capabilities(),
+                     ExcNotImplemented());
+
+              if (dof_handler[0]->has_hp_capabilities())
+                {
+                  Assert(dof_info[0].cell_active_fe_index ==
+                           dof_info[i].cell_active_fe_index,
+                         ExcNotImplemented());
+                }
+            }
+        }
+
       mapping_info.initialize(
         dof_handler[0]->get_triangulation(),
         cell_level_index,


kronbichler:
It makes sense to go the latter route in my opinion, i.e., check that we have the same active fe index on all `DoFHandler` objects with hp capability enabled. I think at some point the infrastructure supported the case that all `DoFHandler` objects had the same active FE index or some `DoFHandler` objects that are not `dof_handler[0]` are non-hp `DoFHandler` types. Not sure if this is still the case because we did never test it, but I think this is the way to go: We want a single selection of active fe indices or the index zero on non-hp `DoFHandler`s.

peterrum:
@kronbichler OK. I have changed the check.

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -442,6 +442,26 @@ MatrixFree<dim, Number, VectorizedArrayType>::internal_reinit(
   // determined in @p extract_local_to_global_indices.
   if (additional_data.initialize_mapping == true)
     {
+      if (dof_handler.size() > 1)
+        {
+          // check if all DoHandlers are in the same hp mode; and if hp
+          // capabilities are enabled: check if active_fe_indices of all
+          // DoFHandler are the same.
+          for (unsigned int i = 0; i < dof_handler.size(); ++i)
```
marcfehling, line 450: You could start the loop at `i = 1`.

marcfehling (approved):

include/deal.II/matrix_free/matrix_free.templates.h
peterrum: You are right. Thanks. I have changed it.

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -442,6 +442,26 @@ MatrixFree<dim, Number, VectorizedArrayType>::internal_reinit(
...
+      if (dof_handler.size() > 1)
+        {
+          // check if all DoHandlers are in the same hp mode; and if hp
+          // capabilities are enabled: check if active_fe_indices of all
+          // DoFHandler are the same.
+          for (unsigned int i = 1; i < dof_handler.size(); ++i)
+            {
+              Assert(dof_handler[0]->has_hp_capabilities() ==
+                       dof_handler[i]->has_hp_capabilities(),
+                     ExcNotImplemented());
```
kronbichler, line 454: Do we plan to also allow the case of combining `hp` with non-`hp` cases? In a multiphysics context it could make sense to have high polynomial degrees in one component but not in the other. But one might also argue that the slight increase in efficiency from allowing those in the same `MatrixFree` is only a step towards allowing completely different meshes and the new evaluator from #11046, where an extension like this is unnecessary.

kronbichler (approved):
Looks good from the current state.

include/deal.II/matrix_free/matrix_free.templates.h
peterrum: I don't consider these asserts as permanent. In particular, would like to lift `dof_info[0].cell_active_fe_index == dof_info[i].cell_active_fe_index`. But these asserts represent the current state within `MatrixFree` and such are useful.

PR #11247, opened 2020-11-25
Make MappingInfo::initialize_faces work for multiple fe indices

peterrum:


include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1801,34 +1805,52 @@ namespace internal
...
-                     ExcNotImplemented());
+              AssertDimension(mapping_in.size(), 1);
+              AssertDimension(mapping_info.face_data[my_q].descriptor.size(),
+                              1);
+#endif
+
+              const unsigned int fe_index =
+                active_fe_index.size() > 0 ?
+                  active_fe_index[faces[face].cells_interior[0]] :
+                  0;
```
peterrum, line 1818: This PR does not enable hp + DG in `MatrixFree` but is targeting the case that `Mappings` and `Quadrature` rules match at faces like in the case of mixed meshes. 

This is why I am selecting here the `active_fe_index` of the inner cell.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1759,31 +1761,33 @@ namespace internal
...
-        AssertDimension(mapping_in.size(), 1);
-        const auto &mapping = mapping_in[0];
-
         FE_Nothing<dim> dummy_fe;
 
+        const unsigned int n_active_fe_indices =
+          active_fe_index.size() > 0 ?
+            (*std::max_element(active_fe_index.begin(), active_fe_index.end()) +
+             1) :
+            1;
```
marcfehling, line 1778: Could you find a more descriptive variable name? This is more like `max_active_fe_index_plus_one` but there may be a better one :)

marcfehling (approved):

include/deal.II/matrix_free/mapping_info.templates.h
marcfehling: Maybe move the `+1` part to the resize operations later, and turn this variable into `max_active_fe_index`.

peterrum: It's more `n_(potentially_locally_relevant_)active_fe_indices` but I don't think we need to be so verbose here.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1962,9 +1985,39 @@ namespace internal
...
+                                mapping,
+                                dummy_fe,
+                                quadrature,
+                                mapping_info.update_flags_boundary_faces);
+
+                          fe_face_values_container[my_q][fe_index]->reinit(
+                            cell_it, faces[face].exterior_face_no);
+
+                          actual_fe_face_values =
+                            fe_face_values_container[my_q][fe_index].get();
```
peterrum, line 2020: I have added these lines since the neighbor might have different active fe, mapping, and quadrature indices.

include/deal.II/matrix_free/mapping_info.templates.h
kronbichler: I am with @marcfehling here because this really seems like `max_active_fe_index`. By `n_active_fe_indices` I would think of filtering out lower numbers which does not happen (at least not here). Not that our collections waste many numbers, but I like the name.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1801,34 +1805,53 @@ namespace internal
...
+              AssertDimension(mapping_in.size(), 1);
+              AssertDimension(mapping_info.face_data[my_q].descriptor.size(),
+                              1);
+#endif
+
+              const unsigned int fe_index =
+                active_fe_index.size() > 0 ?
+                  active_fe_index[faces[face].cells_interior[0] /
+                                  VectorizedArrayType::size()] :
+                  0;
```
kronbichler, line 1819: Can you put a comment here that this assumes that we have the faces sorted by the active fe indices (#11248) to have the same decision here? Just that we do not forget about it.

include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -1962,9 +1985,39 @@ namespace internal
...
                           GeometryInfo<dim>::max_children_per_cell)
                         {
-                          fe_face_values.reinit(cell_it,
-                                                faces[face].exterior_face_no);
-                          actual_fe_face_values = &fe_face_values;
+                          const unsigned int fe_index =
+                            active_fe_index.size() > 0 ?
+                              active_fe_index[faces[face].cells_exterior[0] /
+                                              VectorizedArrayType::size()] :
+                              0;
```
kronbichler, line 1992: Same here.

include/deal.II/matrix_free/mapping_info.templates.h
kronbichler: I do not really understand why we take different indices here. I agree for the `fe_index` and mapping indices possibly, but shouldn't we use the same quadrature formula in a genuinely hp setup? After all, both sides of an integral must contribute with some terms that get matched up at the same location. I might miss something here, but I must admit I am confused by having the index set up twice.

kronbichler (commented):
Looks good overall, except some comments.

peterrum:
> I do not really understand why we take different indices here. I agree for the fe_index and mapping indices possibly, but shouldn't we use the same quadrature formula in a genuinely hp setup? After all, both sides of an integral must contribute with some terms that get matched up at the same location. I might miss something here, but I must admit I am confused by having the index set up twice.

From a mixed mesh perspective, we need the right mapping and the right **face-quadrature rule collection**. At the moment, we have a "single" face-quadrature rule, which is projected to all faces, so that choosing the same quadrature-rule index would work. However, once we go to genuine 3D mixed meshes this will not be the case.  

This implementation assumes that quadrature rules at faces match from both sides (as stated in one of the many PRs). This is not 100% safe. I guess we should add some kind of a check, but I don't know where. Also I don't know a better solution.

peterrum:
@kronbichler I have renamed the variable and added the comments.

kronbichler (approved):

PR #11249, opened 2020-11-25
Add MatrixFree::create_inner_face_subrange_hp_by_index()

peterrum:
... and `MatrixFree::create_boundary_face_subrange_hp_by_index()`. The test shows how I intend it to use it (here for a collection of two `FE_DGQ`s but in the future with a collection of `FE_DGQ` and `Simplex::FE_DGQ`)

Follow-up PR to PR #11248.


peterrum:
I have rebase this PR so that it is ready from my side. I also used the opportunity to replace some `vector_component` with `dof_handler_index`.

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -122,14 +122,137 @@ MatrixFree<dim, Number, VectorizedArrayType>::create_cell_subrange_hp_by_index(
...
+
+
+template <int dim, typename Number, typename VectorizedArrayType>
+std::pair<unsigned int, unsigned int>
+MatrixFree<dim, Number, VectorizedArrayType>::
+  create_inner_face_subrange_hp_by_index(
+    const std::pair<unsigned int, unsigned int> &range,
+    const unsigned int                           fe_index_interior,
+    const unsigned int                           fe_index_exterior,
+    const unsigned int                           dof_handler_index) const
```
kronbichler, line 177: Given how complicated these functions and the respective functions for the cells are, shouldn't we simply let the `TaskInfo` break up the cell and face ranges by this granularity? That way all calls coming in here would be with the same fe index. What do you think?

peterrum: > What do you think?

Sounds reasonable although it as it is took some time to get this work ;) Where do I find the right lines of code within `TaskInfo`?

kronbichler: I agree these things with hp are complicated. But that's life when we need to go from iterator-friendly unstructured access functions to array-like HPC-optimized data structures for matrix-free loop scheduling.

Regarding the question, we currently only do things for the cells. One place where we set things up in a generic way is this one here for non-hp code: https://github.com/dealii/dealii/blob/662eb95cfecb4f045941ffc67e396a5edfe6b2b7/source/matrix_free/task_info.cc#L1027-L1031
For `hp` I believe we need to propagate the numbering from the crs-like data structures in https://github.com/dealii/dealii/blob/662eb95cfecb4f045941ffc67e396a5edfe6b2b7/source/matrix_free/task_info.cc#L886-L891 to the place where we fill the blocks and split things up.

And then we would need to collect things also for the face part which is currently done without the hp granularity here: https://github.com/dealii/dealii/blob/662eb95cfecb4f045941ffc67e396a5edfe6b2b7/include/deal.II/matrix_free/face_setup_internal.h#L1283-L1286

peterrum: Before proceeding on this issue. I would like to clarify the following: 

If we actually split up the cell and faces ranges that would mean that we would not need `MatrixFree::create_cell_subrange_hp()` and `MatrixFree::create_cell_subrange_hp_by_index()` any more. So we should try to deprecate these.

A code (taken from PR #11250) could look like this:

```cpp
    matrix_free.template cell_loop<VectorType, int>(
      [&](const auto &data, auto &dst, const auto &, const auto cell_range) {
            const unsigned int i = matrix_free.get_active_cell_index(cell_range /* ... ?*/); // <---
            FEEvaluation<dim, -1, 0, 1, number> phi(matrix_free, 0, 0, 0, i, i);

            for (unsigned int cell = cell_range.first; cell < cell_range.second; ++cell)
              {
                phi.reinit(cell);
                for (unsigned int q = 0; q < phi.n_q_points; ++q)
                  phi.submit_value(1.0, q);

                phi.integrate_scatter(true, false, dst);
              }
      },
      vec,
      dummy,
      true);
```
without the additional loop (in the case of cell and boundary faces) and loops (in the case of inner faces).

kronbichler: Yes, that's what I would think, we could do it this way. Regarding the `get_active_fe_index()`, why not simply ask for `matrix_free.get_active_cell_index(cell_range.first)`? And then we put an assert into `FEEvaluation` upon `reinit()` that it refers to the correct active fe index.

peterrum: >  why not simply ask for matrix_free.get_active_hp_fe_index(cell_range.first)?

For me this already expects some pre-knowledge by the user: the user creates an `FEEvaluation` for a rage and initializes by the active cell index of the first macro cell of this range!? Sounds a bit odd...

However, if one would come from a "single cell point of view" and create a new `FEEvaluation` (or `reinit(active_fe_index)`) for each macro cel,l this point of view would make more sense. But this is not what we want users to do (due to setup overheads).

 > And then we put an assert into FEEvaluation upon reinit() that it refers to the correct active fe index.

This is definetly something what we should do!

kronbichler: We can also simply add the whole range and do this step internally (alongside with an assertion).

peterrum: I have added the `Do nor merge` label back. I will probably create a new fresh PR for the issue to have this as reference.

PR #11250, opened 2020-11-25
[WIP] Add Poisson test: MatrixFree + DG + mixed mesh

peterrum:
Only the last commits are relevant.

### DIFF
--- a/include/deal.II/matrix_free/evaluation_kernels.h
+++ b/include/deal.II/matrix_free/evaluation_kernels.h
@@ -2546,8 +2546,70 @@ namespace internal
         const unsigned int            face_orientation,
         const Table<2, unsigned int> &orientation_map)
     {
-      Assert(data.element_type != MatrixFreeFunctions::tensor_none,
-             ExcNotImplemented());
+      if (data.element_type == MatrixFreeFunctions::tensor_none)
+        {
+          const unsigned int n_dofs     = data.dofs_per_component_on_cell;
+          const unsigned int n_q_points = data.n_q_points_face;
+          const auto         shape_info = data.data.front();
+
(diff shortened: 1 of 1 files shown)


peterrum:
General questions here are:
- How should we generate face-quadrature rules?
- I have added in `ShapeInfo` the arrays `shape_values_face` and `shape_gradients_face`. I guess we should also introduce some pointers into these arrays, translating `face_no` and `orientation`. Any thing missing.
- The initialize function of `ShapeInfo`  is currently limited to elements with all faces being the same. I guess we can work on this issue later. We will need to touch multiple parts of the internals of `MatrixFree` for that. E.g., `ShapeInfo::n_q_points_face`.

kronbichler:
> How should we generate face-quadrature rules?

I think it would be most convenient to deduce it from the cell quadrature in terms of the degree of polynomials integrated. However, I can see a user wanting to specialize the function to specific needs, which would suggest adding an argument somewhere. (Not sure how, though.)

> I have added in ShapeInfo the arrays shape_values_face and shape_gradients_face. I guess we should also introduce some pointers into these arrays, translating face_no and orientation. Any thing missing.

Those look fine. Regarding the pointers I would agree, those can be part of the class I believe. (As long as we do not provide for refinement with sub-face integration for all possible cases.)

> The initialize function of ShapeInfo is currently limited to elements with all faces being the same. I guess we can work on this issue later. We will need to touch multiple parts of the internals of MatrixFree for that. E.g., ShapeInfo::n_q_points_face.

We need to identify appropriate data structures for that. I would like to avoid adding another indirection (e.g. data sitting in an array for which we need to look up the pointer) for the default case and rather have this bailing out early, e.g. via a different local data.

PR #11254, opened 2020-11-26
Step-75: Tutorial for parallel hp-FEM

marcfehling:
Part of #7515.

This is an early version of the tutorial for parallel hp-FEM with barely any documentation.

What it currently features:
- solve Laplace equation on L-shaped domain
- analytic solution available -- used for imposing Dirichlet boundary conditions and error calculation
- hp decision strategy by the decay of Legendre coefficients

Features that need to be added:
- ~ConvergenceTable for convergence results~
- ~multiple hp decision strategies: Fourier coefficients, refinement history~
- [x] enhanced preconditioner: p-relaxation before AMG
- [x] MatrixFree


peterrum (commented):
May I ask you - before doing anything - to move the tutorial to the `examples/step-75` folder.

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+    virtual void initialize_dof_vector(VectorType &vec) const = 0;
+  };
+
+
+
+  template <int dim, typename number>
+  class LaplaceOperatorMatrixBased : public LaplaceOperator<dim, number>
+  {
+  public:
+    using VectorType = LinearAlgebra::distributed::Vector<number>;
```
bangerth, line 292: Might be cleaner to import the type from the base class -- just one place where one would have to write it out.

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+
+    void reinit(const hp::MappingCollection<dim> &mapping_collection,
+                const DoFHandler<dim> &           dof_handler,
+                const hp::QCollection<dim> &      quadrature_collection,
+                const AffineConstraints<number> & constraints,
+                VectorType &                      system_rhs) override
+    {
+#ifndef DEAL_II_WITH_TRILINOS
+      Assert(false, StandardExceptions::ExcNotImplemented());
+      (void)mapping_collection;
```
bangerth, line 302: So all of this is going to trigger also with PETSc?

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+      vec.reinit(partitioner_dealii);
+    }
+
+    const TrilinosWrappers::SparseMatrix &get_system_matrix() const override
+    {
+      return this->system_matrix;
+    }
+
+
+    mutable TrilinosWrappers::SparseMatrix system_matrix;
```
bangerth, line 391: This and the next function should probably be `private:`

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+
+
+
+  template <int dim, typename number>
+  class LaplaceOperatorMatrixFree : public LaplaceOperator<dim, number>
+  {
+  public:
+    using VectorType = LinearAlgebra::distributed::Vector<number>;
+
+    void reinit(const hp::MappingCollection<dim> &mapping,
```
bangerth, line 404: same here

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+        // move to the right-hand side
+        system_rhs -= b;
+      }
+    }
+
+
+
+    void vmult(VectorType &dst, const VectorType &src) const override
+    {
+      dst = 0.0; // TODO: needed?
```
bangerth, line 481: doesn't hurt and reduces the number of questions :)

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+        {
+          const_cast<LinearAlgebra::distributed::Vector<number> &>(src)
+            .local_element(constrained_indices[i]) =
+            constrained_values[i].first;
+          dst.local_element(constrained_indices[i]) =
+            constrained_values[i].first;
+        }
+    }
+
+    using FECellIntegrator = FEEvaluation<dim, -1, 0, 1, number>;
```
bangerth, line 514: also `private:` from here on out

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+                                            comm);
+
+      DoFTools::make_sparsity_pattern(dof_handler, dsp, this->constraints);
+
+      dsp.compress();
+      system_matrix.reinit(dsp);
+    }
+
+    void
+    calculate_system_matrix(TrilinosWrappers::SparseMatrix &system_matrix) const
```
bangerth, line 591: do you actually need this function?

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+
+      system_matrix.compress(VectorOperation::add);
+
+      const auto p = system_matrix.local_range();
+      for (auto i = p.first; i < p.second; i++)
+        if (system_matrix(i, i) == 0.0 && constraints.is_constrained(i))
+          system_matrix.add(i, i, 1);
+    }
+
+    dealii::MatrixFree<dim, number> matrix_free;
```
bangerth, line 692: all `private:`

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+    void flag_adaptation(const LinearAlgebra::distributed::Vector<double>
+                           &locally_relevant_solution);
+    virtual void decide_hp(const LinearAlgebra::distributed::Vector<double>
+                             &locally_relevant_solution);
+    void         limit_levels();
+    virtual void execute_refinement();
+
+    const ParameterHandler &params;
+    const unsigned int      min_level = 5, max_level = dim <= 2 ? 10 : 8;
+    const unsigned int      min_degree = 2, max_degree = dim <= 2 ? 7 : 5;
```
bangerth, line 855: we tend to not use initializers on const variables, and instead do this in the constructor

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+      triangulation, estimated_error_per_cell, 0.3, 0.03);
+  }
+
+
+
+  template <int dim>
+  void AdaptationStrategy<dim>::decide_hp(
+    const LinearAlgebra::distributed::Vector<double>
+      & /*locally_relevant_solution*/)
+  {}
```
bangerth, line 910: :-)

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+    decide_hp(locally_relevant_solution);
+    limit_levels();
+    execute_refinement();
+  }
+
+
+
+  template <int dim>
+  void AdaptationStrategy<dim>::flag_adaptation(
+    const LinearAlgebra::distributed::Vector<double> &locally_relevant_solution)
```
bangerth, line 882: You use `AdaptationStrategy` both as a base class (=interface) and to implement a specific strategy. That seems wrong. Why not just have one interface base class and put all specific code into derived classes?

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+      }
+
+    std::vector<int> cells_to_remove(dim, 1);
+    cells_to_remove[0] = -1;
+
+    // TODO
+    // expand domain by 1 cell in z direction for 3d case
+
+    GridGenerator::subdivided_hyper_L(
+      triangulation, repetitions, bottom_left, top_right, cells_to_remove);
```
bangerth, line 1316: Is this not what you would get from `hyper_L` and then one global refinement?

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+                                             0,
+                                             Solution<dim>(),
+                                             constraints);
+
+#ifdef DEBUG
+    // We have not dealt with chains of constraints on ghost cells yet.
+    // Thus, we are content with verifying their consistency for now.
+    IndexSet locally_active_dofs;
+    DoFTools::extract_locally_active_dofs(dof_handler, locally_active_dofs);
+    AssertThrow(constraints.is_consistent_in_parallel(
```
bangerth, line 1349: I think that's something we should put into a test, but maybe not a tutorial program

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1575 @@
...
+          << "PETSc"
+#else
+          << "Trilinos"
+#endif
+          << " on " << Utilities::MPI::n_mpi_processes(mpi_communicator)
+          << " MPI rank(s)..." << std::endl;
+
+    std::shared_ptr<LaplaceOperator<dim, double>> laplace_operator;
+
+    if (false)
```
bangerth, line 1558: :-)

bangerth (commented):
Nice work already. I look forward to it being finished!

I think the program is too long at the moment. It has too many switches to play with: matrix-based vs matrix-free, etc. I know that it represents many of the pieces you put together for your PhD, but from the perspective of a teaching tool, it would be useful to simplify it. One way to do that would be to simplify the solver question: Just throw out the matrix-based thing altogether. It's obvious how to build a matrix and solve for it -- step-40 does that and does it in what's probably 98% the same way as here. Just go full matrix-free if you think that it is at least as good, and remove all of the matrix-based vestiges.

One could think about also picking one refinement strategy you think is good, and possibly relegating all other options to the "Results" section.

marcfehling:
> Nice work already. I look forward to it being finished!
> 
> I think the program is too long at the moment. It has too many switches to play with: matrix-based vs matrix-free, etc. I know that it represents many of the pieces you put together for your PhD, but from the perspective of a teaching tool, it would be useful to simplify it. One way to do that would be to simplify the solver question: Just throw out the matrix-based thing altogether. It's obvious how to build a matrix and solve for it -- step-40 does that and does it in what's probably 98% the same way as here. Just go full matrix-free if you think that it is at least as good, and remove all of the matrix-based vestiges.
> 
> One could think about also picking one refinement strategy you think is good, and possibly relegating all other options to the "Results" section.

I already suspected that you would find the tutorial size too big :-)

The advantage of this tutorial in its current shape is that it presents all differences between the different methods and strategies used. However, I agree that the size may be too large for a tutorial step and there is not much benefit by just presenting all the features we have at our disposal.

Anyways thinking ahead for the parallel hp paper and/or next release paper, it would be nice to have some program to benchmark the benefits of each method.

What do you think of making the current program with all features part of the code gallery, and simplifying it to use only MatrixFree methods and the Legendre adaptation strategy as a tutorial?

@peterrum - Let's continue to work on this as agreed. We can turn this into a code gallery program when we think it's ready, and then continue to simplify it to become a tutorial and write all the prose around it.

examples/step-75/step-75.cc
marcfehling: This was just an intermediate construct :-)

I wanted to get these parameters from the `ParameterHandler` object. But I did not implement the parameter file part yet. This will be gone soon.

examples/step-75/step-75.cc
marcfehling: Agreed.

However, there will be functions which are duplicates of each other. Maybe I'll introduce an anonymous or internal namespace instead which will feature these functions, and each class will simply call them instead of inheriting them from the mother class.

examples/step-75/step-75.cc
marcfehling: `hyper_L` removes the I. quadrant, while this implementation removes the IV. quadrant.

One could argue that it doesn't matter which quadrant will be removed. However on this domain, the `Solution` function is easier to comprehend. (Otherwise we would need to introduce an azimuth shift of `pi/2`.)

I would like to keep it like this to demonstrate how to use `GridGenerator::subdivided_hyper_L`.

bangerth:
Yes, I think that's a useful approach with the code gallery/tutorial split.

For the tutorial, remember that it is a "tool to learn deal.II". It doesn't have to teach all possible avenues by which one can solve a problem in one step, but we've found that programs that focus on one or at most two aspects work best because that makes it easier to figure out what parts of a program are relevant to whatever the reader wants to learn about. That's a bit the same reason as step-40 is step-40 and not step-1 :-)

marcfehling:
I rebased on the `master` branch to incorporate #11283.

peterrum:
@marcfehling OK!

peterrum:
@marcfehling I have rebased to be able remove some default mapping arguments.

peterrum:
@marcfehling I have made some changes to the tutorial. In particular, I have kicked out the matrix-based operator implementation. We can construct the matrix simply with the help of the matrix-free operator so we don't loose any functionality. Furthermore, I have moved part of the multgrid algorithm into a new file. I intend to integrate the content of that file step by step in the next days/weeks into dealii/master. With these changes, the length of the tutorial has dropped by 450 lines of code. 

What do you think of the changes? Maybe we could have a chat in the next days so that we can finalize the layout of the tutorial!

marcfehling:
> What do you think of the changes? Maybe we could have a chat in the next days so that we can finalize the layout of the tutorial!

I appreciate the changes. If we focus on what we would like to present in this tutorial (hp-adaptive methods with GMG preconditioning), the scope of this tutorial will be just right :)

I thought about refraining from adding a `ConvergenceTable` and also to get rid of a couple of `AdaptationStrategies` so that only the Legendre one remains. This will also decrease the size of the tutorial by about 300 lines.

I would like to present graphs in the results section made with a more extensive tool that I was preparing for the code-gallery, and refer to that tool in the results section so that users can easily look this up. We could also present possible extension in that way: Tell the user what they need to do to use a different hp-strategy for example, and refer to the 'solution' in the code-gallery program.

I will update the tool in the next couple of days and make it a PR once your extractions have been merged in `dealii/master`. Finalizing the layout in the nexy couple of weeks would be nice, we then still have plenty of time until the release to work on the documentation.

luca-heltai (commented):
Would you mind moving the changes to deal.II to a different PR? These may be merged much earlier (and are useful for a case of mine...) :)

marcfehling:
Rebased on `master` to incorporate #11981.

marcfehling:
Rebased on `master` to incorporate #12041.

marcfehling:
Only the documentation of the multigrid methods are missing.

I'll clean up that part tomorrow and can hopefully finish the documentation by the end of the week.

marcfehling:
The solver and preconditioner sections might need a bit more documentation, but the overall structure of this tutorial for the upcoming release is finished. You can find the current state of the documentation here: [step-75.pdf](https://github.com/dealii/dealii/files/6340440/The.deal.II.Library.The.step-75.tutorial.program.pdf).

The size of the tutorial will be reduced once #11699 is finished.

examples/step-75/doc/intro.dox
```diff
@@ -0,0 +1,220 @@
...
+
+Eibner and Melenk @cite eibner2007hp argued that a function is analytic,
+i.e., representable by a power series, if and only if the absolute
+values of the expansion coefficients decay exponentially with increasing
+index $k$:
+@f[
+\exists C,\sigma > 0 : \quad \forall k \in \mathbb{N}_0 : \quad |c_k|
+\leq C \exp\left( - \sigma k \right) .
+@f]
+The rate of decay $sigma_K$ can be interpreted as a measure for the
```
peterrum, line 140: `$sigma_K$` seems not to be used above.

examples/step-75/doc/intro.dox
```diff
@@ -0,0 +1,220 @@
...
+
+<h3>The test case</h3>
+
+For elliptic equations, each reentrant corner typically invokes a
+singularity @cite brenner2008. We can use this circumstance to put our
+hp-decision algorithms to a test: on all cells to be adapted, we would
+prefer a fine grid near the singularity, and a high polynomial degree
+otherwise.
+
+As the simplest elliptic problem to solve with these constraints, we
```
peterrum, line 189: I don't get the meaning of `constraints` in this context.

examples/step-75/doc/results.dox
```diff
@@ -0,0 +1,101 @@
...
+Cycle 2:
+   Number of active cells:       3651
+     by partition:               958 818 933 942
+   Number of degrees of freedom: 26519
+     by partition:               6651 6822 6563 6483
+   Number of constraints:        3013
+     by partition:               801 701 790 721
+   Frequencies of poly. degrees: 2:2385 3:758 4:508
+   Solved in 6 iterations.
+...
```
peterrum, line 66: Could you also output the info of the last cycle. Is this debug or release mode. For the final results, let's run it with release mode (on modern CPU architecture).

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1502 @@
...
+    };
+
+    SmootherParameters     smoother;
+    CoarseSolverParameters coarse_solver;
+
+    MGTransferGlobalCoarseningTools::PolynomialCoarseningSequenceType
+      p_sequence = MGTransferGlobalCoarseningTools::
+        PolynomialCoarseningSequenceType::decrease_by_one;
+    bool perform_h_transfer = true;
+  };
```
peterrum, line 495: I would suggest to integrate this into the other parameter struct. Generally, I would suggest to move the parameters to the top of the file.

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1502 @@
...
+
+
+
+  // @sect3{The <code>LaplaceProblem</code> class implementation}
+
+  // @sect4{Constructor}
+
+  // The constructor starts with an initializer list that looks similar to the
+  // one of step-40. We again ensure that the Triangulation remains smooth,
```
peterrum, line 890: I think we can remove any smoothing flags. At least from MG side, we don't need it.

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1502 @@
...
+    data_out.attach_dof_handler(dof_handler);
+    data_out.add_data_vector(locally_relevant_solution, "solution");
+    data_out.add_data_vector(fe_degrees, "fe_degree");
+    data_out.add_data_vector(subdomain, "subdomain");
+    data_out.add_data_vector(estimated_error_per_cell, "error");
+    data_out.add_data_vector(hp_decision_indicators, "hp_indicator");
+    data_out.build_patches(mapping_collection);
+
+    data_out.write_vtu_with_pvtu_record(
+      "./", "solution", cycle, mpi_communicator, 2, 8);
```
peterrum, line 1361: Why not dump everything into a single file?

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1502 @@
...
+
+// @sect4{main()}
+
+// The final function is the <code>main</code> function that will ultimately
+// create and run a LaplceOperator instantiation. Its structure is similar to
+// most other tutorial programs. We define all parameters used in the tutorial
+// here, so this is a good opportunity to justify how we picked them.
+//
+// We provide ranges for levels on which the function space is operating with a
+// reasonable resolution. The minimal polynomial degree is set to 2 as it seems
```
peterrum, line 1426: > The minimal polynomial degree is set to 2

This should be checked, since I (and also the p-multigrid sequences) am assuming the lowest degree is 1.

examples/step-75/step-75.cc
```diff
@@ -0,0 +1,1502 @@
...
+        prm.max_p_degree           = 6;
+        prm.max_p_level_difference = 1;
+
+        prm.refine_fraction    = 0.3;
+        prm.coarsen_fraction   = 0.03;
+        prm.p_refine_fraction  = 0.9;
+        prm.p_coarsen_fraction = 0.9;
+
+        prm.weighting_factor   = 1e6;
+        prm.weighting_exponent = 1.;
```
peterrum, line 1468: Why not make these parameters default? Just like in the MG case?

peterrum (commented):
@marcfehling  I have pushed some modification directly. I furthermore have some questions/remarks.

examples/step-75/step-75.cc
marcfehling: I agree on the integration part.

I don't feel strongly on where to put it in the file.

examples/step-75/step-75.cc
marcfehling: We could do that to reduce a few lines.

The reason I put it here was to justify the choice of parameters after all the theoretical details have been covered in the in-source documentation. But we could also move the justification to the places where they are actually used.

examples/step-75/step-75.cc
marcfehling: Since linear elements do not exist in the collection, a new pair with a default value `(1,0)` will be added to the `fe_index_for_degree` map. That's why no error is triggered and the code still works.

The last commit checks if polynomial degrees exist in the `FECollection`. Hence, this tutorial will throw assertions at the moment :(

marcfehling: We could also use a `FECollection` with polynomial degrees [1,max], but restrict the hierachy for p-refinement to [2,max].

This way, both precondition and hp-adaptation methods would be happy. What do you think?

examples/step-75/doc/results.dox
peterrum: Sounds good.

examples/step-75/step-75.cc
peterrum: Looks good!

examples/step-75/doc/results.dox
marcfehling: I've updated the terminal output. With the recent changes, the number of solver iterations fluctuates between 7 and 11 in this application.

marcfehling:
Current state of the documentation: [step-75.pdf](https://github.com/dealii/dealii/files/6417326/step-75.pdf)

drwells:
I read through the whole thing and I have two comments:
1. This tutorial uses `internal::parallel::distributed::TemporarilyMatchRefineFlags` - if this class is used in a tutorial it isn't really internal anymore so we should move it into `p::d`.
2. A lot of the documentation for `LaplaceOperator` isn't in the usual style of the tutorials - in particular, the documentation in the declarations reads like the manual. Can you adjust the style to be more in-line with what we do elsewhere?

examples/step-75/doc/intro.dox
```diff
@@ -0,0 +1,230 @@
...
+c_{k,K} = \frac{2k+1}{2} \int\limits_{\hat K} \hat{u}_\text{hp}(\hat{x})
+\hat{P}_k(\hat{x}) \left|\det \mathbf{J}(\hat{x})\right|
+\mathrm{d}\hat{x} ,
+@f]
+where $\varphi(x) = \varphi(F_K(\hat{x})) = \hat{\varphi}(\hat{x})$ for
+a general function $\varphi$. A non-degenerate mapping, i.e.,
+$\left|\det \mathbf{J}(\hat{x})\right| \approx 1$, allows us to perform
+the transformation on any cell $K$ exclusively on the reference cell
+$\hat{K}$ and pre-calculate the necessary transformation matrices only
+once:
```
drwells, line 124: I'm not sure this is right. If `J` is `1` or nearly `1` then the mapping is a translation or rotation, but this is basically never the case for finite elements. Is this described in one of Melenk's papers or some other citation?

marcfehling: You are right, this is basically never the case for finite elements and poses a weakness of the current `FESeries` classes. It would be too expensive to build such a transformation matrix with the correct mapping on every single cell. The same logic is applied on the [old version of the `step-27` tutorial](https://www.dealii.org/8.3.0/doxygen/deal.II/step_27.html#LaplaceProblemestimate_smoothness) where this generalization stems from.

We should be more clear here and say that this only works correctly with a non-degenerate mapping, and that results on distorted cells are inaccurate.

drwells: Does that mean that the code works correctly only if the Jacobian is constant?

marcfehling: Yes, the Fourier and Legendre transformations are only correct for constant Jacobians at the moment.
We only estimate the smoothness with these transformations, so this only impacts the choice of the refinement type (h or p). I don't consider this to be too critical.

drwells: Ah, okay. We should clarify this text to say that this code is only correct for affine cells (not `J == 1`) then.

To be clear, I'm not worried about the functionality here or in step-27, I just want to make sure I (and anyone else reading this) understand what the condition on the cell geometry (i.e., the properties of the mapping back to the unit cell) is.

peterrum (approved):
Thanks for the efforts, @marcfehling! It looks already very good! I have pushed a couple of minor modifications directly to your branch. 

Once @drwells is happy, let's merge!

drwells:
I would still like to fix the discussion on requirements on the Jacobian - i.e., what I wrote in https://github.com/marcfehling/dealii/pull/2 shows that the Jacobian just needs sto be constant.

I need to give the rest of the documentation a second read-through though - I'll do that this morning!

drwells:
I'll reword it.

drwells:
Nevermind those complaints: I misread the way the p-refinement loop worked and I missed the h-refinement step completely. This matches the text just fine.

drwells (approved):
OK but we need to rebase.

marcfehling:
> OK but we need to rebase.

Done. This was tricky with the merge commits in between. The `--rebase-merges` flag helped.

peterrum (approved):
Thanks @drwells for the review and the suggestion! I let you merge!

drwells:
:shipit:

PR #11257, opened 2020-11-26
Add hp capabilities to Simplex::FE_DGP

peterrum:
Similar to #11224.

### DIFF
--- a/source/simplex/fe_lib.cc
+++ b/source/simplex/fe_lib.cc
@@ -15,6 +15,7 @@
 
 #include <deal.II/base/config.h>
 
+#include <deal.II/fe/fe_dgq.h>
 #include <deal.II/fe/fe_q.h>
 
 #include <deal.II/simplex/fe_lib.h>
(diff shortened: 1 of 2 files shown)


source/simplex/fe_lib.cc
```diff
@@ -327,6 +328,48 @@ namespace Simplex
...
+  {
+    (void)fe_other;
+    (void)codim;
+
+    Assert((dynamic_cast<const FE_DGQ<dim, spacedim> *>(&fe_other)),
+           ExcNotImplemented());
+    AssertDimension(dim, 2);
+    AssertDimension(this->degree, fe_other.tensor_degree());
+
+    return FiniteElementDomination::either_element_can_dominate;
```
kronbichler, line 345: I have not looked up the other DG (L2-conforming) elements, but is this the same as for the `FE_DGQ`? To me this looks similar to the `FE_P` case but I don't recall if DG elements have some different more relaxed requirements regarding hp.

peterrum: > but is this the same as for the FE_DGQ?

Yes, it is:

https://github.com/dealii/dealii/blob/662eb95cfecb4f045941ffc67e396a5edfe6b2b7/source/fe/fe_dgq.cc#L627-L636

https://github.com/dealii/dealii/blob/662eb95cfecb4f045941ffc67e396a5edfe6b2b7/source/fe/fe_dgq.cc#L563-L587

kronbichler (approved):

PR #11260, opened 2020-11-26
MatrixFree: split up ranges according to active_fe_indices

peterrum:
... alternative to #11249.


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -4525,8 +4596,17 @@ namespace internal
...
+          {
+            const auto cell_subrange =
+              matrix_free.create_cell_subrange_hp_by_index(cell_range, i);
+
+            if (cell_subrange.second <= cell_subrange.first)
+              continue;
+
+            (container.*
+             cell_function)(matrix_free, this->dst, this->src, cell_subrange);
+          }
```
peterrum, line 4609: Here and below the new/old/deprecated functions `MatrixFree::create_cell_subrange_hp_by_index()`, `MatrixFree::create_inner_face_subrange_hp_by_index()`,  `MatrixFree::create_boundary_face_subrange_hp_by_index()` are called. This is definitely not optimal. However, shows that we could introduce an additional loop level here obviously in smarter way (by pre-computed ranges).

peterrum (commented):
@kronbichler This is not yet 100% the promised alternative. The implementation already provides to the user the correct interfaces so that he/she does not have to deal with subranges.

However, the internal implementation is built around the function "introduced" in #11249. This is not optimal and should be replaced by alternative strategies. Nevertheless, I think the current implementation gives us the basis to discuss where to perform changes to `MatrixFree`. I have identified a place where to introduce an additional loop level that does not cause too much work. The proposed modifications in https://github.com/dealii/dealii/pull/11249/files#r530940617 do not seem to be trivial and in particular as it is now does not fit here:
https://github.com/dealii/dealii/blob/662eb95cfecb4f045941ffc67e396a5edfe6b2b7/source/matrix_free/task_info.cc#L592-L632

Where always one cell-, inner-face, and boundary-face range are processed after each other (if I read the code correctly) - however we want to process multiple subranges. The shown code is the serial case; furthermore there is a threaded case. Both of them have completely different setup routines so that I got lost and I wouldn't know where to start...

peterrum:
The Poisson/Helmholtz-operator would look like this:
https://github.com/dealii/dealii/blob/cbdc617f8cf26aea410df1d968f8cc9feb1d7623/tests/simplex/matrix_free_02.cc#L204-L231

The code is more or less identical to the non-hp version. I am wondering if we couldn't create a new constructor for `FEEvaluation` and `FEFaceEvaluation` that takes a range as parameter. Then the code could simply look like this:
```cpp
       FECellIntegrator phi(matrix_free, range); 
```

kronbichler:
>  I am wondering if we couldn't create a new constructor for FEEvaluation and FEFaceEvaluation that takes a range as parameter.

Yes, I completely agree, we should use those constructors. It also might help us to avoid the ones that explicitly set a long list of integer arguments.

kronbichler (commented):
Yes, this is a very nice solution, I like it :+1: and we should go this way.
Could you please add comments on the new functions and move the definition of those functions in `MatrixFree` out of the class declaration?

include/deal.II/matrix_free/matrix_free.h
kronbichler: That's something we can do in a follow-up work.

include/deal.II/matrix_free/matrix_free.templates.h
```diff
@@ -85,12 +85,17 @@ std::pair<unsigned int, unsigned int>
...
+    return range;
+
+  AssertIndexRange(fe_index, dof_info[dof_handler_index].max_fe_index);
   const std::vector<unsigned int> &fe_indices =
-    dof_info[vector_component].cell_active_fe_index;
-  if (fe_indices.empty() == true)
+    dof_info[dof_handler_index].cell_active_fe_index;
+
+  if (fe_indices.empty() == true ||
+      dof_handlers[dof_handler_index]->get_fe_collection().size() == 1)
```
peterrum, line 98: Please not that had to make following modifications to successfully run all `tests/matrix_free`: 
1. if indices are set up (needed by matrix_free/no_index_initialize.debug)
2. in the case of cell catergorization `fe_indices.empty() == true` is not equivalent for the check if hp-capability is enabled or not (needed for `matrix_free/cell_categorization.mpirun=1.debug`)

peterrum (commented):
@kronbichler I have updated the code!

include/deal.II/matrix_free/matrix_free.h
```diff
@@ -2471,6 +2516,81 @@ MatrixFree<dim, Number, VectorizedArrayType>::at_irregular_cell(
...
+
+  if (fe_indices.empty() == true)
+    return 0;
+
+  const auto index = fe_indices[range.first];
+
+#  ifdef DEBUG
+  for (unsigned int i = range.first; i < range.second; ++i)
+    AssertDimension(index, fe_indices[i]);
+#  endif
```
kronbichler, line 2542: I don't think you need the #ifdef here because `AssertDimension` already expands to nothing for non-debug build. There are a few more cases further down.

include/deal.II/matrix_free/matrix_free.templates.h
kronbichler: OK.

kronbichler (approved):

include/deal.II/matrix_free/matrix_free.h
peterrum: I have removed it!

kronbichler (approved):

PR #11264, opened 2020-11-27
Fix indentation CI

peterrum:


.github/workflows/indent.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     - name: setup
       run: |
         ./contrib/utilities/download_clang_format
-        sudo apt-get install doxygen graphviz perl texlive-bibtex-extra
+        sudo apt-get update && sudo apt-get install doxygen graphviz perl texlive-bibtex-extra
```
kronbichler, line 19: Not sure what kind of debian/ubuntu runs on that machine, and not sure about the non-ubuntu case at all, but isn't the preferred tool `apt` rather than `apt-get` nowadays?

peterrum: Lets see!

peterrum: @kronbichler OK. It has passed! Ready to be merged!

kronbichler (approved):

PR #11267, opened 2020-11-28
Introduce FE_Wedge

peterrum:
Let's get this finally starting (after laying untouched 3 months around). What is the mathematical appropriate name of this class?


drwells:
I don't think I have ever seen names for the specific FEs aside from WEDGE6 or WEDGE18. `FE_Q` implies that we have Lagrange tensor-product elements and `FE_P` implies that we have Lagrange simplex elements, so the most logical name for this class is `FE_WedgeP` (and we can write `FE_WedgeDGP` later).

drwells:
The rest of this PR is fine so I'll be happy to merge once we agree on a name.

peterrum:
I am fine with the suggested names! Other suggestions?

peterrum:
> The rest of this PR is fine so I'll be happy to merge once we agree on a name.

@drwells  I have changed the name to `FE_WedgeP`. Let's proceed here so that I can continue with other PRs. We can still change the name later on if there are objections!

drwells (approved):

PR #11270, opened 2020-11-28
Enable DataOut for Wedges

peterrum:
Depends on #11267.

### DIFF
--- a/include/deal.II/numerics/data_out_dof_data.templates.h
+++ b/include/deal.II/numerics/data_out_dof_data.templates.h
@@ -82,6 +82,167 @@ namespace internal
     {}
 
 
+
+    /**
+     * Set up vectors of FEValues and FEFaceValues needed inside of
+     * ParallelDataBase and return the maximum number of quadrature points
+     * needed to allocate enough memory for the scratch data.
+     */
+    template <int dim, int spacedim>
+    unsigned int
+    setup_parallel_data_base_internal(
+      const unsigned int                                  n_subdivisions,
+      const dealii::hp::MappingCollection<dim, spacedim> &mapping_collection,
+      const std::vector<
+        std::shared_ptr<dealii::hp::FECollection<dim, spacedim>>>
+        &               finite_elements,
+      const UpdateFlags update_flags,
+      const bool        use_face_values,
+      std::vector<std::shared_ptr<dealii::hp::FEValues<dim, spacedim>>>
+        &x_fe_values,
+      std::vector<std::shared_ptr<dealii::hp::FEFaceValues<dim, spacedim>>>
+        &x_fe_face_values)
+    {
+      unsigned int n_q_points = 0;
+      if (use_face_values == false)
+        {
+          // determine if specific quadrature rules need to set up
+          bool needs_hypercube_setup = false;
+          bool needs_simplex_setup   = false;
+          bool needs_wedge_setup     = false;
+
+          for (const auto &fe : finite_elements)
+            for (unsigned int i = 0; i < fe->size(); ++i)
+              switch ((*fe)[i].reference_cell_type())
+                {
+                  case ReferenceCell::Type::Vertex:
+                  case ReferenceCell::Type::Line:
+                  case ReferenceCell::Type::Quad:
+                  case ReferenceCell::Type::Hex:
+                    needs_hypercube_setup |= true;
+                    break;
+                  case ReferenceCell::Type::Tri:
+                  case ReferenceCell::Type::Tet:
+                    needs_simplex_setup |= true;
+                    break;
+                  case ReferenceCell::Type::Wedge:
+                    needs_wedge_setup |= true;
+                    break;
+                  default:
+                    Assert(false, ExcNotImplemented());
+                }
+
+          std::unique_ptr<dealii::Quadrature<dim>> quadrature_simplex;
+          std::unique_ptr<dealii::Quadrature<dim>> quadrature_hypercube;
+          std::unique_ptr<dealii::Quadrature<dim>> quadrature_wedge;
+
+          if (needs_simplex_setup)
+            {
+              quadrature_simplex = std::make_unique<Quadrature<dim>>(
+                Simplex::FE_P<dim, spacedim>(n_subdivisions)
+                  .get_unit_support_points());
+            }
+
+          if (needs_hypercube_setup)
+            {
+              quadrature_hypercube =
+                std::make_unique<QIterated<dim>>(QTrapezoid<1>(),
+                                                 n_subdivisions);
+            }
+
+          if (needs_wedge_setup)
+            {
+              quadrature_wedge = std::make_unique<Quadrature<dim>>(
+                Simplex::FE_WedgeP<dim, spacedim>(
+                  1 /*note: vtk only supports linear wedges*/)
+                  .get_unit_support_points());
+            }
+
+          n_q_points = std::max(
+            {needs_wedge_setup ? quadrature_wedge->size() : 0,
+             needs_simplex_setup ? quadrature_simplex->size() : 0,
+             needs_hypercube_setup ? quadrature_hypercube->size() : 0});
+
+          x_fe_values.resize(finite_elements.size());
+          for (unsigned int i = 0; i < finite_elements.size(); ++i)
+            {
+              // check if there is a finite element that is equal to the
+              // present one, then we can re-use the FEValues object
+              for (unsigned int j = 0; j < i; ++j)
+                if (finite_elements[i].get() == finite_elements[j].get())
+                  {
+                    x_fe_values[i] = x_fe_values[j];
+                    break;
+                  }
+              if (x_fe_values[i].get() == nullptr)
+                {
+                  dealii::hp::QCollection<dim> quadrature;
+
+                  for (unsigned int j = 0; j < finite_elements[i]->size(); ++j)
+                    switch ((*finite_elements[i])[j].reference_cell_type())
+                      {
+                        case ReferenceCell::Type::Vertex:
+                        case ReferenceCell::Type::Line:
+                        case ReferenceCell::Type::Quad:
+                        case ReferenceCell::Type::Hex:
+                          quadrature.push_back(*quadrature_hypercube);
+                          break;
+                        case ReferenceCell::Type::Tri:
+                        case ReferenceCell::Type::Tet:
+                          quadrature.push_back(*quadrature_simplex);
+                          break;
+                        case ReferenceCell::Type::Wedge:
+                          quadrature.push_back(*quadrature_wedge);
+                          break;
+                        default:
+                          Assert(false, ExcNotImplemented());
+                      }
+
+                  x_fe_values[i] =
+                    std::make_shared<dealii::hp::FEValues<dim, spacedim>>(
+                      mapping_collection,
+                      *finite_elements[i],
+                      quadrature,
+                      // certain reference-cell kinds take by default the
+                      // curved code path in DataOut::build_one_patch, for
+                      // which the quadrature points need to be enabled
+                      (needs_simplex_setup || needs_wedge_setup) ?
+                        (update_flags | update_quadrature_points) :
+                        update_flags);
+                }
+            }
+        }
+      else
+        {
+          dealii::hp::QCollection<dim - 1> quadrature(
+            QIterated<dim - 1>(QTrapezoid<1>(), n_subdivisions));
+          n_q_points = quadrature[0].size();
+          x_fe_face_values.resize(finite_elements.size());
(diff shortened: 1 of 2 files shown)


include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -103,30 +103,43 @@ namespace internal
...
         {
-          const bool do_simplex =
+          const bool do_hex =
             std::any_of(finite_elements.begin(),
                         finite_elements.end(),
                         [](const auto &fe) {
                           for (unsigned int i = 0; i < fe->size(); ++i)
-                            if ((*fe)[i].reference_cell_type() !=
+                            if ((*fe)[i].reference_cell_type() ==
                                 ReferenceCell::get_hypercube(dim))
```
bangerth, line 112: Maybe as a follow-up, a function of the kind 
```
  ... .reference_cell().is_hyper_cube()
```
would be useful, and similarly
```
  ... .reference_cell().is_simplex()
```
I don't know how many places you use this kind of comparison in, but if it's in more than a few, then think about this as a follow-up.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -103,30 +103,43 @@ namespace internal
       unsigned int n_q_points = 0;
       if (use_face_values == false)
         {
-          const bool do_simplex =
+          const bool do_hex =
             std::any_of(finite_elements.begin(),
```
bangerth, line 107: Shouldn't this now be `std::all_of`?

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -103,30 +103,43 @@ namespace internal
...
-                            if ((*fe)[i].reference_cell_type() !=
+                            if ((*fe)[i].reference_cell_type() ==
                                 ReferenceCell::get_hypercube(dim))
                               return true;
                           return false;
                         });
 
-          const bool do_hex =
+          const bool do_simplex =
+            std::any_of(finite_elements.begin(),
```
bangerth, line 118: same here -- `std::all_of`?

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -103,30 +103,43 @@ namespace internal
...
+                        finite_elements.end(),
+                        [do_hex](const auto &fe) {
+                          for (unsigned int i = 0; i < fe->size(); ++i)
+                            if (do_hex == false &&
+                                (*fe)[i].reference_cell_type() !=
+                                  ReferenceCell::get_hypercube(dim))
```
bangerth, line 124: and then here comparison against `ReferenceCell::get_simplex(dim)`?

include/deal.II/numerics/data_out_dof_data.templates.h
bangerth: As a matter of fact, I now realize that you may be using this variable when asking whether you need to set something up. In that case, `std::any_of` is correct, but the variable name is not giving that away. Maybe `needs_hypercube_setup`?

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -143,8 +156,19 @@ namespace internal
                 QIterated<dim>(QTrapezoid<1>(), n_subdivisions)));
             }
 
-          n_q_points = std::max(do_simplex ? quadrature1->size() : 0,
-                                do_hex ? quadrature2->size() : 1);
+          if (do_wedge)
+            {
+              Assert(1 <= n_subdivisions && n_subdivisions <= 2,
+                     ExcNotImplemented());
+              quadrature3.reset(new Quadrature<dim>(
```
bangerth, line 163: This is better done using
```
  quadrature3 = std::make_unique<Quadrature<dim>> (...)
```
Same above.

Separately, maybe name these variables `quadrature_hypercube`, `quadrature_simplex`, `quadrature_wedge`?

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -143,8 +156,19 @@ namespace internal
...
+                Simplex::FE_WedgeP<dim, spacedim>(
+                  1 /*note: vtk only supports linear wedges*/)
+                  .get_unit_support_points()));
+            }
+
+          n_q_points = std::max(do_wedge ? quadrature3->size() : 0,
```
bangerth, line 169: use the `std::max({...,...,...})` form here

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -164,19 +188,24 @@ namespace internal
 
                   for (unsigned int j = 0; j < this->finite_elements[i]->size();
                        ++j)
-                    if ((*this->finite_elements[i])[j].reference_cell_type() !=
+                    if ((*this->finite_elements[i])[j].reference_cell_type() ==
                         ReferenceCell::get_hypercube(dim))
-                      quadrature.push_back(*quadrature1);
-                    else
                       quadrature.push_back(*quadrature2);
```
bangerth, line 193: Here's a place that lends itself to hard-to-find bugs because quadrature1/2/3 doesn't give away what they correspond to.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -164,19 +188,24 @@ namespace internal
...
+                    else if ((*this->finite_elements[i])[j]
+                               .reference_cell_type() ==
+                             ReferenceCell::Type::Wedge)
+                      quadrature.push_back(*quadrature3);
+                    else
+                      quadrature.push_back(*quadrature1);
```
bangerth, line 199: You'll get into trouble here if another cell type appears (namely, once you have pyramids). You'll have to *know* that you need to add something new here. Better to use a `switch(....reference_cell_type())` statement that's explicit about what cases each of these corresponds to, with an explicit `Assert(false,...)` in the `default:` branch.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -164,19 +188,24 @@ namespace internal
...
                   x_fe_values[i] =
                     std::make_shared<dealii::hp::FEValues<dim, spacedim>>(
                       this->mapping_collection,
                       *this->finite_elements[i],
                       quadrature,
-                      do_simplex ? (update_flags | update_quadrature_points) :
-                                   update_flags);
+                      (do_simplex || do_wedge) ?
+                        (update_flags | update_quadrature_points) :
+                        update_flags);
```
bangerth, line 208: Can you add a comment why you need `update_quadrature_points` but only for some of the element types?

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -242,30 +271,43 @@ namespace internal
         {
           Assert(data.x_fe_face_values.empty() == true, ExcInternalError());
 
-          const bool do_simplex =
+          const bool do_hex =
```
bangerth, line 274: same issues as above

bangerth (changes requested):

peterrum:
@bangerth I think I have incorporated all your suggestions (thanks)!

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -164,19 +182,37 @@ namespace internal
 
                   for (unsigned int j = 0; j < this->finite_elements[i]->size();
                        ++j)
-                    if ((*this->finite_elements[i])[j].reference_cell_type() !=
-                        ReferenceCell::get_hypercube(dim))
-                      quadrature.push_back(*quadrature1);
-                    else
-                      quadrature.push_back(*quadrature2);
+                    switch (
+                      (*this->finite_elements[i])[j].reference_cell_type())
```
drwells, line 186: It's cosmetic, but could we get rid of the extra parentheses?

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -242,44 +278,60 @@ namespace internal
...
+                    break;
+                  default:
+                    Assert(false, ExcNotImplemented());
+                }
+
+          std::unique_ptr<dealii::Quadrature<dim>> quadrature_simplex;
+          std::unique_ptr<dealii::Quadrature<dim>> quadrature_hypercube;
+          std::unique_ptr<dealii::Quadrature<dim>> quadrature_wedge;
+
+          if (needs_simplex_setup)
```
drwells, line 311: How hard would it be to generalize this part of the setup? This code is identical to the other one at circa line 130.

peterrum: Done.

drwells:
@bangerth since this PR (and many like it) are needed for further simplex development, would you mind if (in clear cases) I go ahead and merge PRs if all of your changes have been implemented? That could prevent things from getting stuck that Peter needs elsewhere.

drwells:
@bangerth *ping*

drwells:
@bangerth we need to get this merged so that we can move on to other things with wedges.

bangerth:
@drwells My apologies for not seeing this right away. If it's urgent, send me a direct email -- that I look at every day, this github things not always. But you're as qualified to merge as I do. If you see a clear case, go ahead with it!

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -82,6 +82,170 @@ namespace internal
...
+            for (const auto &fe : finite_elements)
+              for (unsigned int i = 0; i < fe->size(); ++i)
+                switch ((*fe)[i].reference_cell_type())
+                  {
+                    case ReferenceCell::Type::Vertex:
+                    case ReferenceCell::Type::Line:
+                    case ReferenceCell::Type::Quad:
+                    case ReferenceCell::Type::Hex:
+                      needs_hypercube_setup |= true;
+                      break;
```
bangerth, line 120: I feel like this all could have been done more easily using a `std::any` call with a nice little lambda function. That could also have been done in the rhs of the initialization of these variables above, and then you can make them `const`

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -82,6 +82,170 @@ namespace internal
...
+
+            std::unique_ptr<dealii::Quadrature<dim>> quadrature_simplex;
+            std::unique_ptr<dealii::Quadrature<dim>> quadrature_hypercube;
+            std::unique_ptr<dealii::Quadrature<dim>> quadrature_wedge;
+
+            if (needs_simplex_setup)
+              {
+                Assert(1 <= n_subdivisions && n_subdivisions <= 2,
+                       ExcNotImplemented());
+                quadrature_simplex = std::make_unique<Quadrature<dim>>(
```
bangerth, line 140: Why do you need the assertion? Because you currently don't have `Simplex::FE_P` implemented for higher polynomial degrees? That's something that `FE_P` should assert, not here

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -82,6 +82,170 @@ namespace internal
...
+              {
+                quadrature_hypercube =
+                  std::make_unique<QIterated<dim>>(QTrapezoid<1>(),
+                                                   n_subdivisions);
+              }
+
+            if (needs_wedge_setup)
+              {
+                Assert(1 <= n_subdivisions && n_subdivisions <= 2,
+                       ExcNotImplemented());
```
bangerth, line 155: same here

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -82,6 +82,170 @@ namespace internal
     {}
 
 
+
+    namespace
+    {
+      template <int dim, int spacedim>
+      unsigned int
+      setup_parallel_data_base_internal(
```
bangerth, line 90: this function could use a comment about what it does and what it returns (directly or through its non-`const` reference arguments).

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -82,6 +82,170 @@ namespace internal
     {}
 
 
+
+    namespace
```
bangerth, line 86: we can't have anonymous namespaces in header files :-(

bangerth (commented):
Sorry, still a couple of issues :-(

peterrum:
@bangerth I have incorporated most of the comments. I have not introduced the `std::any` suggestion since I don't see how it would benefit here. If you strongly feel about `std::any` we can tackle it in a follow-up PR.

bangerth (approved):

bangerth:
The MPI check failed because the compiler had a segfault. I take that as "out of memory". Merging.

PR #11269, opened 2020-11-28
Add a poisson test on wedge grid

peterrum:
Only the last commit is relevant. 

Depends on #11267 and #11268.

PR #11271, opened 2020-11-28
Enable TriaAccessor::diameter() for wedges

peterrum:
Needed by #11269.


bangerth (approved):
I assume that the formula you have here is a representation of taking the longest of the 3x2 diagonals of the quadrilateral faces. If so, this is ok, and assuming that it is, I'll merge because I have a follow-up -- @peterrum please tell me if I was wrong.

bangerth:
@peterrum -- I think I was too quick to merge this. #11285 documents a case where this is not the right implementation. Would you be willing to follow-up with another patch (after #11284 is merged)?

PR #11274, opened 2020-11-30
Move two classes into an internal namespace.

bangerth:
This is a follow-up to #11244. Move these classes into the internal namespace -- I don't think they were ever meant to be in the public namespace to begin with, given how well they were hidden in the bottom part of a file...

/rebuild

### DIFF
--- a/doc/news/changes/incompatibilities/20201129Bangerth
+++ b/doc/news/changes/incompatibilities/20201129Bangerth
@@ -1,0 +1,8 @@
+Removed: The file `affine_constraints.h` had a class called
+`IsBlockMatrix` that, however, also allowed to check whether its
+template argument is in fact a block sparsity pattern. The class was
+likely never intended to be in the public interface, and has been
+moved to internal::AffineConstraints::IsBlockMatrix and now also
+internal::AffineConstraints::IsBlockSparsityPattern.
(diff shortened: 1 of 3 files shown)


doc/news/changes/incompatibilities/20201129Bangerth
```diff
@@ -0,0 +1,8 @@
+Removed: The file `affine_constraints.h` had a class called
+`IsBlockMatrix` that, however, also allowed to check whether its
+template argument is in fact a block sparsity pattern. The class was
+likely never intended to be in the public interface, and has been
+moved to internal::AffineConstraints::IsBlockMatrix and now also
+internal::AffineConstraints::IsBlockSparsityPattern.
```
peterrum, line 6: `now also`? The sentence structure does not seem to be right!?

masterleinad (approved):
Fine with me.

doc/news/changes/incompatibilities/20201129Bangerth
bangerth: Well, it has been "moved to X and also Y" because it was split into two classes in #11244. I think this is grammatically and semantically correct.

kronbichler (approved):

PR #11276, opened 2020-11-30
Remove ExcPointNotFound from GridTools::find_active_cell_around_point()

luca-heltai:
The current state of  GridTools::find_active_cell_around_point() is broken. 

This is my proposal: let's make things right. User codes will have to check that the returned cell is valid. If not, their code will throw later on.

This is the most agreed solution to long lasting problems with GridTools::find_active_cell_around_point() (@peterrum @dangars @nfehn @bangerth )

#10175 #10535 #7576

I've tried to fix some of the tests. I have problems with a ParticleHandler generator on grids of co-dimension one (for which in general find_active_cell_around_point does not return anything usable).

Feel free to help me out/chime in with suggested changes/proposals. I think this is the least intrusive way forward, before we refactor all the various versions of `find_active_cell_*`  methods.

It took me a long time, as I'm mostly off-programming. Please forgive if I won't be too responsive.

### DIFF
--- a/doc/news/changes/incompatibilities/20201130LucaHeltai
+++ b/doc/news/changes/incompatibilities/20201130LucaHeltai
@@ -1,4 +1,5 @@
 Changed: GridTools::find_active_cell_around_point() no longer throws an exception, but returns an invalid iterator.
-User codes should check that the returned cell is valid instead of relying on exceptions. This solves issues with MPI, and HDF5.
+User codes should now check that the returned cell is valid instead of relying on exceptions, instead of trying
+to catch exceptions that this function may have thrown.
 <br>
 (Luca Heltai, 2020/11/30)
--- a/doc/news/changes/incompatibilities/20201130LucaHeltai
+++ b/doc/news/changes/incompatibilities/20201130LucaHeltai
@@ -1,0 +1,4 @@
+Changed: GridTools::find_active_cell_around_point() no longer throws an exception, but returns an invalid iterator.
+User codes should check that the returned cell is valid instead of relying on exceptions. This solves issues with MPI, and HDF5.
+<br>
+(Luca Heltai, 2020/11/30)
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -862,8 +862,8 @@ namespace GridTools
 
   /**
    * Given a Triangulation's @p cache and a list of @p points, call
-   * find_active_cell_around_point() on each element of @p points , and return
-   * @p cells , referece positions @p qpoints , and a mapping @p maps from local
+   * find_active_cell_around_point() on each element of @p points, and return
+   * @p cells, reference positions @p qpoints, and a mapping @p maps from local
    * to global indices into @p points .
    *
    * @param[in] cache The triangulation's GridTools::Cache .
@@ -891,7 +891,7 @@ namespace GridTools
    *
    * @note If a point is not found inside the mesh, or is lying inside an
    * artificial cell of a parallel::TriangulationBase, the point is silently
-   * igored. If you want to infer for which points the search failed, use the
+   * ignored. If you want to infer for which points the search failed, use the
    * function compute_point_locations_try_all() that also returns a vector of
    * indices indicating the points for which the search failed.
    *
@@ -1451,6 +1451,11 @@ namespace GridTools
    *      // do something with cell_and_ref_point
    *      ...
    *   }
+   *  else
+   *    {
+   *       // The function did not find a locally owned or ghost cell in which
+   *       // the point is located. We ought to handle this somehow here.
+   *    }
    *   ...
    * }
    * @endcode
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -861,7 +861,10 @@ namespace GridTools
   /*@{*/
 
   /**
-   * Given a Triangulation's @p cache and a list of @p points create the quadrature rules.
+   * Given a Triangulation's @p cache and a list of @p points, call
+   * find_active_cell_around_point() on each element of @p points , and return
+   * @p cells , referece positions @p qpoints , and a mapping @p maps from local
+   * to global indices into @p points .
    *
    * @param[in] cache The triangulation's GridTools::Cache .
    * @param[in] points The point's vector.
@@ -883,14 +886,14 @@ namespace GridTools
    * Mapping::transform_unit_to_real(qpoints[c][0])
    * returns @p points[a].
    *
-   * The algorithm assumes it's easier to look for a point in the cell that was
-   * used previously. For this reason random points are, computationally
-   * speaking, the worst case scenario while points grouped by the cell to which
-   * they belong are the best case. Pre-sorting points, trying to minimize
-   * distances between them, might make the function extremely faster.
+   * The algorithm builds an rtree of @p points to sort them spatially, before
+   * attempting to call find_active_cell_around_point().
    *
    * @note If a point is not found inside the mesh, or is lying inside an
-   * artificial cell of a parallel::TriangulationBase, an exception is thrown.
+   * artificial cell of a parallel::TriangulationBase, the point is silently
+   * igored. If you want to infer for which points the search failed, use the
+   * function compute_point_locations_try_all() that also returns a vector of
+   * indices indicating the points for which the search failed.
    *
    * @note The actual return type of this function, i.e., the type referenced
    * above as @p return_type, is
@@ -927,7 +930,10 @@ namespace GridTools
 
   /**
    * This function is similar to GridTools::compute_point_locations(),
-   * but it tries to find and transform every point of @p points.
+   * but while compute_point_locations() silently ignores all points for which
+   * find_active_cell_around_point() fails, this function also returns a
+   * vector containing the indices of the points for which
+   * find_active_cell_around_point() failed.
    *
    * @return A tuple containing four elements; the first three
    * are documented in GridTools::compute_point_locations().
@@ -1280,11 +1286,6 @@ namespace GridTools
    * adjacent cells.
    * @return A vector of cells that lie adjacent to the given vertex.
    *
-   * @note If the point requested does not lie in any of the cells of the mesh
-   * given, then this function throws an exception of type
-   * GridTools::ExcPointNotFound. You can catch this exception and decide what
-   * to do in that case.
-   *
    * @note It isn't entirely clear at this time whether the function does the
    * right thing with anisotropically refined meshes. It needs to be checked
    * for this case.
@@ -1318,14 +1319,8 @@ namespace GridTools
    * tries to identify the cell that is of highest refinement level.
    *
    * If the point requested does not lie in a locally-owned or ghost cell,
-   * then this function throws an exception of type GridTools::ExcPointNotFound.
-   * You can catch this exception and decide what to do in that case. Hence,
-   * for programs that work with partitioned (parallel) triangulations, this
-   * function should always be called inside a `try`-block unless it is a
-   * priori clear that the point with which it is called must be inside
-   * a locally owned or ghost cell (and not close enough to the boundary
-   * between ghost and artificial cells so that decision which cell it is
-   * on depends on floating point accuracy).
+   * then this function will return the (invalid) MeshType<dim, spacedim>::end()
+   * iterator.
    *
    * @param mapping The mapping used to determine whether the given point is
    *   inside a given cell.
@@ -1447,18 +1442,15 @@ namespace GridTools
    *
    * for(auto p : points)
    * {
-   *   try
-   *   {
-   *     auto cell_and_ref_point = GridTools::find_active_cell_around_point(
-   *       cache, p, cell_hint, marked_vertices, tolerance);
+   *   auto cell_and_ref_point = GridTools::find_active_cell_around_point(
+   *     cache, p, cell_hint, marked_vertices, tolerance);
    *
-   *     // use current cell as hint for the next point
-   *     cell_hint = cell_and_ref_point.first;
+   *   if(cell_and_ref_point.first != triangulation.end()) {
+   *      // use current cell as hint for the next point
+   *      cell_hint = cell_and_ref_point.first;
+   *      // do something with cell_and_ref_point
+   *      ...
    *   }
-   *   catch(...)
-   *   {
-   *   }
-   *
    *   ...
    * }
    * @endcode
@@ -1525,9 +1517,9 @@ namespace GridTools
    *
    * This function is used as follows
    * @code
-   *   auto first_cell = GridTools::find_active_cell_around_point(...);
+   *   auto first_pair = GridTools::find_active_cell_around_point(...);
    *   auto all_cells  = GridTools::find_all_active_cells_around_point(
-   *   			   mapping, mesh, p, tolerance, first_cell);
+   *   			   mapping, mesh, p, tolerance, first_pair);
    * @endcode
    */
   template <int dim, template <int, int> class MeshType, int spacedim>
@@ -3426,24 +3418,6 @@ namespace GridTools
                  double,
                  << "The scaling factor must be positive, but it is " << arg1
                  << ".");
-  /**
-   * Exception
-   */
-  template <int N>
-  DeclException1(ExcPointNotFoundInCoarseGrid,
-                 Point<N>,
-                 << "The point <" << arg1
-                 << "> could not be found inside any of the "
-                 << "coarse grid cells.");
-  /**
-   * Exception
-   */
-  template <int N>
-  DeclException1(ExcPointNotFound,
-                 Point<N>,
-                 << "The point <" << arg1
-                 << "> could not be found inside any of the "
-                 << "subcells of a coarse grid cell.");
 
   /**
    * Exception
--- a/include/deal.II/grid/grid_tools.h
+++ b/include/deal.II/grid/grid_tools.h
@@ -1320,7 +1320,8 @@ namespace GridTools
    *
    * If the point requested does not lie in a locally-owned or ghost cell,
    * then this function will return the (invalid) MeshType<dim, spacedim>::end()
-   * iterator.
+   * iterator. This case can be handled similarly to the various `std::find()`
+   * and `std::lower_bound()` functions.
    *
    * @param mapping The mapping used to determine whether the given point is
    *   inside a given cell.
@@ -1445,7 +1446,8 @@ namespace GridTools
    *   auto cell_and_ref_point = GridTools::find_active_cell_around_point(
    *     cache, p, cell_hint, marked_vertices, tolerance);
    *
-   *   if(cell_and_ref_point.first != triangulation.end()) {
+   *   if (cell_and_ref_point.first != triangulation.end())
+   *     {
    *      // use current cell as hint for the next point
    *      cell_hint = cell_and_ref_point.first;
    *      // do something with cell_and_ref_point
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -5484,69 +5484,72 @@ namespace GridTools
     const auto &b_tree = cache.get_cell_bounding_boxes_rtree();
 
     // Now make a tree of indices for the points
-    const auto p_tree = pack_rtree_of_indices(points);
+    // [TODO] This would work better with pack_rtree_of_indices, but
+    // windows does not like it. Build a tree with pairs of point and id
+    std::vector<std::pair<Point<spacedim>, unsigned int>> points_and_ids(np);
+    for (unsigned int i = 0; i < np; ++i)
+      points_and_ids[i] = std::make_pair(points[i], i);
+    const auto p_tree = pack_rtree(points_and_ids);
 
     // Keep track of all found points
     std::vector<bool> found_points(points.size(), false);
 
     // Check if a point was found
-    const auto already_found = [&found_points](const unsigned int &id) {
-      AssertIndexRange(id, found_points.size());
-      return found_points[id];
+    const auto already_found = [&found_points](const auto &id) {
+      AssertIndexRange(id.second, found_points.size());
+      return found_points[id.second];
     };
 
-    // check if the cell was already in the vector before. If so, returns its
-    // index, otherwise adds it to the list of cells, return the new index,
-    // and resize also the other two vectors. Start from last entry.
-    const auto get_local_cell_index =
-      [&](const typename Triangulation<dim, spacedim>::active_cell_iterator
-            &cell) {
+    // check if the given cell was already in the vector of cells before. If so,
+    // insert in the corresponding vectors the reference point and the id.
+    // Otherwise append a new entry to all vectors.
+    const auto store_cell_point_and_id =
+      [&](
+        const typename Triangulation<dim, spacedim>::active_cell_iterator &cell,
+        const Point<dim> &  ref_point,
+        const unsigned int &id) {
         const auto it = std::find(cells_out.rbegin(), cells_out.rend(), cell);
         if (it != cells_out.rend())
           {
-            return (cells_out.size() - 1 - (it - cells_out.rbegin()));
+            const auto cell_id =
+              (cells_out.size() - 1 - (it - cells_out.rbegin()));
+            qpoints_out[cell_id].emplace_back(ref_point);
+            maps_out[cell_id].emplace_back(id);
           }
         else
           {
             cells_out.emplace_back(cell);
-            qpoints_out.resize(cells_out.size());
-            maps_out.resize(cells_out.size());
-            return (cells_out.size() - 1);
+            qpoints_out.emplace_back(std::vector<Point<dim>>({ref_point}));
+            maps_out.emplace_back(std::vector<unsigned int>({id}));
           }
       };
 
     // Check all points within a given pair of box and cell
-    const auto check_all_points_within_box =
-      [&](const decltype(*b_tree.begin()) &leaf) {
-        const auto &box       = leaf.first;
-        const auto &cell_hint = leaf.second;
+    const auto check_all_points_within_box = [&](const auto &leaf) {
+      const auto &box       = leaf.first;
+      const auto &cell_hint = leaf.second;
 
-        for (const auto id :
-             p_tree | bgi::adaptors::queried(!bgi::satisfies(already_found) &&
-                                             bgi::intersects(box)))
-          {
-            const auto cell_and_ref =
-              GridTools::find_active_cell_around_point(cache,
-                                                       points[id],
-                                                       cell_hint);
-            const auto &cell      = cell_and_ref.first;
-            const auto &ref_point = cell_and_ref.second;
+      for (const auto &point_and_id :
+           p_tree | bgi::adaptors::queried(!bgi::satisfies(already_found) &&
+                                           bgi::intersects(box)))
+        {
+          const auto id = point_and_id.second;
+          const auto cell_and_ref =
+            GridTools::find_active_cell_around_point(cache,
+                                                     points[id],
+                                                     cell_hint);
+          const auto &cell      = cell_and_ref.first;
+          const auto &ref_point = cell_and_ref.second;
 
-            if (cell.state() == IteratorState::valid)
-              {
-                const auto cell_id = get_local_cell_index(cell);
-                qpoints_out[cell_id].emplace_back(ref_point);
-                maps_out[cell_id].emplace_back(id);
-                found_points[id] = true;
-              }
-            else
-              {
-                missing_points_out.emplace_back(id);
-                // Don't look anymore for this point
-                found_points[id] = true;
-              }
-          }
-      };
+          if (cell.state() == IteratorState::valid)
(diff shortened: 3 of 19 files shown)


tjhei:
I do like the idea of reworking how this works, because most codes I write wrap the call in a try..catch directly. But I don't think making the change this way is a good idea. Can we implement a different function with a different name instead? Otherwise the user will not get deprecation or any other warnings (the code will just stop working).

luca-heltai:
@tjhei ,  #10737 describes why having an Exception is not a good idea: it is a bug for any HDF5/MPI users. 

Is it really worth it to add a all set of new functions with a new name, just to deprecate the old (bugged) ones?

Notice that all cascade of functions (compute_point_locations, distributed_compute_point_locations, find_all_active_cell_around_point), etc will have to be provided with two versions (throw and no-throw). I'm not sure this is such a good idea...

Trying to use an invalid cell iterator will throw anyway (a different exception, I agree). User codes will have to deal differently with their exception handling.

tjhei:
> @tjhei , #10737 describes why having an Exception is not a good idea: it is a bug for any HDF5/MPI users.

I agree that his is problematic and worth addressing, especially because it causes issues for some people.

> Is it really worth it to add a all set of new functions with a new name, just to deprecate the old (bugged) ones?

My thought is that your change will break everyone's code so anyone using this function will need to remember to go into their code and remove the catch/throw. I think this would be unfortunate. Or am I incorrect?


> Notice that all cascade of functions (compute_point_locations, distributed_compute_point_locations, find_all_active_cell_around_point), etc will have to be provided with two versions (throw and no-throw). I'm not sure this is such a good idea...

Of course it is annoying. At least you should be able to implement the existing ones using the new one.

How do you propose functions like VectorTools::point_value to behave in the future?

luca-heltai:
`VectorTools::point_value` is a *user* of `find_active_cell_around_point`, and should check for correctness of the returned cell (throwing an exception, if a point is not found). I think that the `find` function itself should not throw.

tjhei:
> `VectorTools::point_value` is a _user_ of `find_active_cell_around_point`, and should check for correctness of the returned cell (throwing an exception, if a point is not found)

Okay, that is fine with me.

> I think that the `find` function itself should not throw.

I think this is a better design and I fully agree with you. But again: I would like us to think about the update process for users.

dangars:
@luca-heltai Thanks a lot for working on this!

On Thursday I'll have time to test this solution with step-62 and HDF5. I'll keep you updated with the results.

dangars:
Thanks a lot @luca-heltai !

I confirm that step-62 of deal.II 9.2 does not freeze anymore. It throws a nice message in debug mode and that can help the user to identify that the behavior of `find_active_cell_around_point()` has changed.

I submitted the PR #11310 which is based on this PR and fixes #10737

I'll keep #11310 sync with this PR.

tamiko (approved):

doc/news/changes/incompatibilities/20201130LucaHeltai
```diff
@@ -0,0 +1,4 @@
+Changed: GridTools::find_active_cell_around_point() no longer throws an exception, but returns an invalid iterator.
+User codes should check that the returned cell is valid instead of relying on exceptions. This solves issues with MPI, and HDF5.
```
bangerth, line 2: We decided that the problem isn't actually with MPI or HDF5 but with not checking exceptions. Let's just omit this part.
```suggestion
User codes should now check that the returned cell is valid instead of relying on exceptions, instead of trying
to catch exceptions that this function may have thrown.
```

include/deal.II/grid/grid_tools.h
```diff
@@ -812,7 +812,10 @@ namespace GridTools
   /*@{*/
 
   /**
-   * Given a Triangulation's @p cache and a list of @p points create the quadrature rules.
+   * Given a Triangulation's @p cache and a list of @p points, call
+   * find_active_cell_around_point() on each element of @p points , and return
+   * @p cells , referece positions @p qpoints , and a mapping @p maps from local
```
bangerth, line 817:
```suggestion
   * @p cells, reference positions @p qpoints, and a mapping @p maps from local
```

include/deal.II/grid/grid_tools.h
```diff
@@ -812,7 +812,10 @@ namespace GridTools
   /*@{*/
 
   /**
-   * Given a Triangulation's @p cache and a list of @p points create the quadrature rules.
+   * Given a Triangulation's @p cache and a list of @p points, call
+   * find_active_cell_around_point() on each element of @p points , and return
```
bangerth, line 816:
```suggestion
   * find_active_cell_around_point() on each element of @p points, and return
```

include/deal.II/grid/grid_tools.h
```diff
@@ -834,14 +837,14 @@ namespace GridTools
...
-   * speaking, the worst case scenario while points grouped by the cell to which
-   * they belong are the best case. Pre-sorting points, trying to minimize
-   * distances between them, might make the function extremely faster.
+   * The algorithm builds an rtree of @p points to sort them spatially, before
+   * attempting to call find_active_cell_around_point().
    *
    * @note If a point is not found inside the mesh, or is lying inside an
-   * artificial cell of a parallel::TriangulationBase, an exception is thrown.
+   * artificial cell of a parallel::TriangulationBase, the point is silently
+   * igored. If you want to infer for which points the search failed, use the
```
bangerth, line 845:
```suggestion
   * ignored. If you want to infer for which points the search failed, use the
```

include/deal.II/grid/grid_tools.h
```diff
@@ -1296,18 +1291,15 @@ namespace GridTools
...
-   *   {
+   *   auto cell_and_ref_point = GridTools::find_active_cell_around_point(
+   *     cache, p, cell_hint, marked_vertices, tolerance);
+   *
+   *   if(cell_and_ref_point.first != triangulation.end()) {
+   *      // use current cell as hint for the next point
+   *      cell_hint = cell_and_ref_point.first;
+   *      // do something with cell_and_ref_point
+   *      ...
    *   }
```
bangerth, line 1302:
```suggestion
   *   }
   *  else
   *    {
   *       // The function did not find a locally owned or ghost cell in which the point is
   *       // located. We ought to handle this somehow here.
   *    }
```

source/grid/grid_tools.cc
```diff
@@ -4352,284 +4351,102 @@ namespace GridTools
...
+    // Check if a point was found
+    const auto already_found = [&found_points](const unsigned int &id) {
+      AssertIndexRange(id, found_points.size());
+      return found_points[id];
+    };
 
-        // If a suitable cell was found, use it as hint
-        if (cell_candidate_idx != -1)
+    // check if the cell was already in the vector before. If so, returns its
+    // index, otherwise adds it to the list of cells, return the new index,
```
bangerth, line 4367:
```suggestion
    // index, otherwise add it to the list of cells, return the new index,
```

source/grid/grid_tools.cc
```diff
@@ -4352,284 +4351,102 @@ namespace GridTools
...
+    // and resize also the other two vectors. Start from last entry.
```
bangerth, line 4368:
```suggestion
    // and resize also the other two vectors. Start from last entry, based on the assumption
    // that subsequent points are likely to be in the same cell as the immediately
    // preceding ones.
```

source/grid/grid_tools.cc
```diff
@@ -4352,284 +4351,102 @@ namespace GridTools
...
+    const auto get_local_cell_index =
```
bangerth, line 4369: maybe call the function `maybe_emplace_cell_and_get_index`. The function doesn't just return an index.

source/grid/grid_tools.cc
```diff
@@ -4352,284 +4351,102 @@ namespace GridTools
...
-                  use_try = true;
-
-                // If a cell candidate was found we can stop
-                if (cell_candidate_idx != -1)
-                  break;
-              }
-          }
+    // Check all points within a given pair of box and cell
+    const auto check_all_points_within_box =
+      [&](const decltype(*b_tree.begin()) &leaf) {
```
bangerth, line 4388: since you're already using a generic lambda with the second and third argument, why can't you just use
```suggestion
      [&](const auto &leaf) {
```
here?

source/grid/grid_tools.cc
```diff
@@ -4352,284 +4351,102 @@ namespace GridTools
...
-            catch (const GridTools::ExcPointNotFound<dim> &)
-              {
-                missing_points_out.emplace_back(p);
-                continue;
-              }
-          }
-        else
+        for (const auto id :
+             p_tree | bgi::adaptors::queried(!bgi::satisfies(already_found) &&
+                                             bgi::intersects(box)))
```
bangerth, line 4394: I can't wait to use C++20 ranges with these kinds of filters :-)

source/grid/grid_tools.cc
```diff
@@ -4352,284 +4351,102 @@ namespace GridTools
...
-                      cache, points[p], box_cell[cell_candidate_idx].second);
-                  }
-                catch (const GridTools::ExcPointNotFound<dim> &)
-                  {
-                    missing_points_out.push_back(p);
-                    continue;
-                  }
+                const auto cell_id = get_local_cell_index(cell);
+                qpoints_out[cell_id].emplace_back(ref_point);
+                maps_out[cell_id].emplace_back(id);
```
bangerth, line 4407: I think a simpler design would be to all three of these operations in the lambda. You'd have to pass the `ref_point` and `id` to that function as well, but at least you don't do the resizing in one place and the emplacement of values in another.

source/grid/grid_tools.cc
```diff
@@ -4352,284 +4351,102 @@ namespace GridTools
...
+          // Now checks all points that fall within this box
+          if (leaf != b_tree.qend())
+            check_all_points_within_box(*leaf);
+          else
+            {
+              // We should not get here. Throw an error.
+              Assert(false,
+                     ExcInternalError(
+                       "I cannot find the closest cell to a point."
+                       "Something is rotten inside."));
```
bangerth, line 4439: I did not know that you can pass a different than the default text to `ExcInternalError`. I'd try to stick with the default -- you're not providing much more information here

bangerth (commented):
I have to admit that I didn't try to understand the complete logic of the function, but I trust that it is reasonable and I've left a number of comments.

I think the general idea is correct: Instead of throwing an exception, return an invalid iterator. That's what `std::find` and similar functions do, and we might as well do the same. It's also possible to change this in user code in a backward compatible way: Just add the check for an invalid iterator while keeping the try-catch structure.

masterleinad:
```
D:\a\dealii\dealii\bundled\boost-1.70.0\include\boost/geometry/index/rtree.hpp(1723,30): error C2664: 'boost::geometry::index::detail::translator<IndexableGetter,EqualTo>::translator(const boost::geometry::index::indexable<dealii::Point<1,double>> &,const EqualTo &)': cannot convert argument 1 from 'const IndGet' to 'const boost::geometry::index::indexable<dealii::Point<1,double>> &' [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              IndexableGetter=dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>>,
              EqualTo=boost::geometry::index::equal_to<std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type>
          ]
          and
          [
              IndGet=dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>>
          ]
D:\a\dealii\dealii\bundled\boost-1.70.0\include\boost/geometry/index/rtree.hpp(1729,1): message : Reason: cannot convert from 'const IndGet' to 'const boost::geometry::index::indexable<dealii::Point<1,double>>' [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              IndGet=dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>>
          ]
D:\a\dealii\dealii\bundled\boost-1.70.0\include\boost/geometry/index/rtree.hpp(1723,31): message : No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
D:\a\dealii\dealii\bundled\boost-1.70.0\include\boost/geometry/index/detail/translator.hpp(23,5): message : see declaration of 'boost::geometry::index::detail::translator<IndexableGetter,EqualTo>::translator' [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              IndexableGetter=dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>>,
              EqualTo=boost::geometry::index::equal_to<std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type>
          ]
D:\a\dealii\dealii\bundled\boost-1.70.0\include\boost/geometry/index/rtree.hpp(297): message : see reference to function template instantiation 'boost::geometry::index::rtree<std::_Default_allocator_traits<_Alloc>::size_type,boost::geometry::index::linear<16,4>,dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>>,boost::geometry::index::equal_to<Value>,boost::container::new_allocator<Value>>::members_holder::members_holder<dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>>,boost::geometry::index::equal_to<Value>,const boost::container::new_allocator<Value>&>(const IndGet &,const ValEq &,const Parameters &,Alloc)' being compiled [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              _Alloc=std::allocator<dealii::Point<1,double>>,
              Value=std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type,
              IndGet=dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>>,
              ValEq=boost::geometry::index::equal_to<std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type>,
              Parameters=boost::geometry::index::linear<16,4>,
              Alloc=const boost::container::new_allocator<std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type> &
          ]
D:\a\dealii\dealii\bundled\boost-1.70.0\include\boost/geometry/index/rtree.hpp(296): message : see reference to function template instantiation 'boost::geometry::index::rtree<std::_Default_allocator_traits<_Alloc>::size_type,boost::geometry::index::linear<16,4>,dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>>,boost::geometry::index::equal_to<Value>,boost::container::new_allocator<Value>>::members_holder::members_holder<dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>>,boost::geometry::index::equal_to<Value>,const boost::container::new_allocator<Value>&>(const IndGet &,const ValEq &,const Parameters &,Alloc)' being compiled [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              _Alloc=std::allocator<dealii::Point<1,double>>,
              Value=std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type,
              IndGet=dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>>,
              ValEq=boost::geometry::index::equal_to<std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type>,
              Parameters=boost::geometry::index::linear<16,4>,
              Alloc=const boost::container::new_allocator<std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type> &
          ]
D:\a\dealii\dealii\include\deal.II/numerics/rtree.h(472): message : see reference to function template instantiation 'boost::geometry::index::rtree<std::_Default_allocator_traits<_Alloc>::size_type,boost::geometry::index::linear<16,4>,dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>>,boost::geometry::index::equal_to<Value>,boost::container::new_allocator<Value>>::rtree<IteratorT>(Iterator,Iterator,const boost::geometry::index::linear<16,4> &,const dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>> &,const boost::geometry::index::equal_to<Value> &,const boost::container::new_allocator<Value> &)' being compiled [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              _Alloc=std::allocator<dealii::Point<1,double>>,
              Value=std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type,
              IteratorT=boost::range_detail::integer_iterator<unsigned __int64>,
              Iterator=boost::range_detail::integer_iterator<unsigned __int64>
          ]
D:\a\dealii\dealii\include\deal.II/numerics/rtree.h(466): message : see reference to function template instantiation 'boost::geometry::index::rtree<std::_Default_allocator_traits<_Alloc>::size_type,boost::geometry::index::linear<16,4>,dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>>,boost::geometry::index::equal_to<Value>,boost::container::new_allocator<Value>>::rtree<IteratorT>(Iterator,Iterator,const boost::geometry::index::linear<16,4> &,const dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>> &,const boost::geometry::index::equal_to<Value> &,const boost::container::new_allocator<Value> &)' being compiled [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              _Alloc=std::allocator<dealii::Point<1,double>>,
              Value=std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type,
              IteratorT=boost::range_detail::integer_iterator<unsigned __int64>,
              Iterator=boost::range_detail::integer_iterator<unsigned __int64>
          ]
D:\a\dealii\dealii\source\grid\grid_tools.cc(4355): message : see reference to function template instantiation 'boost::geometry::index::rtree<std::_Default_allocator_traits<_Alloc>::size_type,boost::geometry::index::linear<16,4>,dealii::IndexableGetterFromIndices<std::vector<dealii::Point<1,double>,_Alloc>>,boost::geometry::index::equal_to<Value>,boost::container::new_allocator<Value>> dealii::pack_rtree_of_indices<boost::geometry::index::linear<16,4>,std::vector<dealii::Point<1,double>,_Alloc>>(const ContainerType &)' being compiled [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
          with
          [
              _Alloc=std::allocator<dealii::Point<1,double>>,
              Value=std::_Default_allocator_traits<std::allocator<dealii::Point<1,double>>>::size_type,
              ContainerType=std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>
          ]
D:\a\dealii\dealii\build\source\grid\grid_tools.inst(1137): message : see reference to function template instantiation 'std::tuple<std::vector<dealii::TriaActiveIterator<dealii::CellAccessor<1,1>>,std::allocator<dealii::TriaActiveIterator<dealii::CellAccessor<1,1>>>>,std::vector<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>,std::allocator<std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>>>>,std::vector<std::vector<unsigned int,std::allocator<unsigned int>>,std::allocator<std::vector<unsigned int,std::allocator<unsigned int>>>>,std::vector<unsigned int,std::allocator<unsigned int>>> dealii::GridTools::compute_point_locations_try_all<1,1>(const dealii::GridTools::Cache<1,1> &,const std::vector<dealii::Point<1,double>,std::allocator<dealii::Point<1,double>>> &,const dealii::TriaActiveIterator<dealii::CellAccessor<1,1>> &)' being compiled [D:\a\dealii\dealii\build\source\grid\obj_grid_debug.vcxproj]
```

bangerth:
@luca-heltai Any updates?

luca-heltai:
Yes. This PR uncovered a bug with boost that shows up only in MSVSC. I'm having difficulties in fixing it now, so I'll split the PR in the next days in two parts. One only removing the exception, and the other one refactoring the function. 

This is also needed now that @peterrum has implemented a better version of distributed_compute_point_locations.

I'll work on this ASAP.

luca-heltai:
I asked for help for the bug I discovered, related to MSVC:

https://stackoverflow.com/questions/66889783/using-boostgeometryindexindexable-within-an-indexablegetter-in-windows

you can experiment with this (if you have time) here:

https://godbolt.org/z/vTT5r5MWc

where you have to add boost as a library in the options of the compiler. 

I'll address the rest of this PR today or tomorrow.

luca-heltai:
I've fixed almost all failing stuff. I'm not sure about the three remaining tests that fail. As far as I can tell, they create a triangulation, instantiate an FEFieldFunction, and then verify that generalized interpolation is coherent. However this is not robust when used with discontinuous elements, and the three tests that fail all have discontinuous elements, resulting in differences between evaluated and expected functions. I'm unsure how to proceed. What do people suggest here?

luca-heltai:
This is now ready for re-review. I have changed some test outputs, for DG functions evaluated on generalized support points that now fall within different cells w.r.t. the original version. Apart from this, everything else seems to work.

masterleinad:
If these tests are just not expected to work properly for discontinuous elements, I would rather just disable them instead of updating the output. Have you made an attempt to see what is really going on?
In particular, `numerics/generalized_interpolation_02` only uses a constant function that should be exactly representable by all the elements used and I would expect that a sensible interpolation would represent them exactly.

dangars:
I rebased #11310 and I confirm that step-62 it works well with this PR.

source/grid/grid_tools_dof_handlers.cc
```diff
@@ -1384,9 +1377,6 @@ namespace GridTools
           }
       }
 
-    AssertThrow(best_cell.first.state() == IteratorState::valid,
-                ExcPointNotFound<spacedim>(p));
```
kronbichler, line 1388: In case we have never updated `best_cell`, does the resulting (invalid) iterator compare equal to `DoFHandler::end()`? In other words, I'm asking if the default constructed `DoFHandler::active_cell_iterator` is the same as `end()`. If not, we might need to initialize `best_cell.first` to `DoFHandler::end()`.

tests/numerics/generalized_interpolation_02.output
```diff
@@ -4,61 +4,61 @@ DEAL::Function value at (0.0,0.0): 1.00000
...
-DEAL::Check projection property: 0.00000
+DEAL::Check projection property: 1.11022e-16
 DEAL::dim 2 FE_RaviartThomas<2>(1)
 DEAL::Function value at (0.0,0.0): 1.00000 1.00000 
-DEAL::Check projection property: 0.00000
+DEAL::Check projection property: 4.44089e-16
 DEAL::dim 2 FE_RaviartThomas<2>(2)
 DEAL::Function value at (0.0,0.0): 1.00000 1.00000 
-DEAL::Check projection property: 0.00000
+DEAL::Check projection property: 0.169442
```
kronbichler, line 43: Do you have an explanation for this (large) change? The same for a few entries below for RT and Nedelec cases.

tests/numerics/generalized_interpolation_04.output
```diff
@@ -1,3 +1,3 @@
 
 DEAL::dim 2 FESystem<2>[FE_RaviartThomas<2>(1)-FE_Q<2>(1)-FE_Nedelec<2>(2)^2-FESystem<2>[FE_Q<2>(1)^2]]
-DEAL::Check projection property: 0.00000
+DEAL::Check projection property: 0.699901
```
kronbichler, line 3: Do we understand this change?

tests/particles/particle_handler_07.with_p4est=true.mpirun=1.output
```diff
@@ -31,37 +31,7 @@ DEAL:2d/2d::Particle id 8 is in cell 2.15
...
-DEAL:2d/3d::Particle id 5 is in cell 2.12
-DEAL:2d/3d::     at location 0.550000 0.550000 0.550000
-DEAL:2d/3d::     at reference location 0.200000 0.200000
-DEAL:2d/3d::Particle id 9 is in cell 2.15
-DEAL:2d/3d::     at location 0.950000 0.950000 0.950000
-DEAL:2d/3d::     at reference location 0.800000 0.800000
-DEAL:2d/3d::Particle id 8 is in cell 2.15
-DEAL:2d/3d::     at location 0.850000 0.850000 0.850000
-DEAL:2d/3d::     at reference location 0.400000 0.400000
+DEAL:2d/3d::Particle number: 0
```
kronbichler, line 34: Do you know what changed here?

tests/simplex/compute_point_locations_01.output
```diff
@@ -70,87 +70,140 @@ DEAL::
...
-DEAL::        physical position   : (0.00000 0.00000 0.00000)
-DEAL::        FE mapping position : (0.00000 0.00000 0.00000)
-DEAL::    qpoints 1: 
 DEAL::        reference position  : (0.00000 0.00000 0.333333)
 DEAL::        physical position   : (0.00000 0.00000 0.333333)
 DEAL::        FE mapping position : (0.00000 0.00000 0.333333)
+DEAL::    qpoints 1: 
+DEAL::        reference position  : (0.333333 0.00000 0.666667)
+DEAL::        physical position   : (0.333333 0.00000 0.666667)
+DEAL::        FE mapping position : (0.333333 0.00000 0.666667)
```
kronbichler, line 79: So the changes here seem to indicate that we now, by changing the algorithm, happen to get another cell which shares the edge with the cell that we identified before. This seems uncritical to me, but I wonder how stable this test is going to be, and if configurations with slightly different roundoff behavior will still pass the test.

kronbichler (approved):
I like this change, because I think not finding a point is not exceptional but rather a case that happens. I guess there are quite a number of application codes that need to adjust after this change, but I see it vital for the future use of these kinds of functions. And the old behavior was weird and unexpected from a user perspective, going against what the C++ standard library does for search/find operations.

I am approving already now to express my support, even though I have some remarks below that we might want to look at. None of those is critical, though, and we can postpone them if we want more time with the new functionality before the release.

include/deal.II/grid/grid_tools.h
```diff
@@ -3297,24 +3294,6 @@ namespace GridTools
...
-  DeclException1(ExcPointNotFoundInCoarseGrid,
-                 Point<N>,
-                 << "The point <" << arg1
-                 << "> could not be found inside any of the "
-                 << "coarse grid cells.");
-  /**
-   * Exception
-   */
-  template <int N>
-  DeclException1(ExcPointNotFound,
```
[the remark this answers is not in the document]
luca-heltai: I thought about this, and this is precisely why I did it this way. This also forced me to make sure we were not relying on this exception anywhere in the library. I could put it back in there if people think we could have something like `find_active_cell_around_point_compatibility` that actually throws an exception.

tests/numerics/generalized_interpolation_02.output
luca-heltai: @masterleinad seems convinced that these tests should not fail, but I'm not so sure about this. I'll try to understand a bit better, and if they are not supposed to work, I'll disable them.

The explanation I found is that a point that was previously found on a given cell is now on the neighbor cell, and we are testing discontinuous shape functions on generalized support points. Since this can happen, using a FEFieldFunction on DG elements should not be considered robust w.r.t. to a function beeing single valued.

On the other hand, @masterleinad suggested that the function which is tested here *is* actually continuous, so this should not fail. I'll try to investigate a bit more.

tests/particles/particle_handler_07.with_p4est=true.mpirun=1.output
luca-heltai: Yes. The previous algorithm was returning garbage. Particles don't work (yet) properly on manifolds of co-dimension one or two, since in those cases the transformation from real to unit cell is not invertible in general, but just on the manifold, and our algorithms are not robust for this case. The current algorithm won't find particles because it starts off by checking inside a bounding box of the cell first, and all particles here are away from the manifold.

drwells:
I think the majority of developers are behind this change - I know @tjhei is not (as has been discussed above). Does anyone else object? Either way we need to figure this out before the release.

include/deal.II/grid/grid_tools.h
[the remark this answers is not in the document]
kronbichler: For me personally we don't need to preserve the exception (this is going to be a tough interface break anyway).

kronbichler:
I agree, let us move forward with this one as quickly as possible, so I suggest to merge (or will merge in around 10 hours if nobody speaks up until then). @luca-heltai should we merge immediately and look at or adjust the test later, or do we disable it here?

luca-heltai:
Let's merge, and open an issue pointing to these tests.

PR #11277, opened 2020-11-30
Fix MGLevelObject variadic template constructor with default arguments

masterleinad:
Fixes https://cdash.43-1.org/testDetails.php?test=65698929&build=13766.


tjhei (approved):

bangerth (approved):

bangerth:
Yep, that made no sense before :-)

PR #11280, opened 2020-11-30
Deprecate classes in Legacy namespace.

marcfehling:
Part of #10333, #11222. Blocked by #11220.

For the subsequent release, we would like to deprecate the classes in the `Legacy` namespace. This PR stays WIP until after the 9.3 release.


marcfehling:
We change the template parameters in #11221. Actually we do not need this intermediate deprecation in this PR. I'll close it.

PR #11283, opened 2020-12-01
Use boost implementation of legendre polynomials instead of gsl.

marcfehling:
Noticed in #11254.

Blocked by #11286.

Instead of using the boost variant directly, we could also introduce a `legendre` function in the `std_cxx17` namespace which uses the boost implementation. This way we could remind ourselves of using the CXX17 function once we move to the more recent standard. I don't expect the results to be different, as they are also not different between boost and gsl.

What is your preference?

### DIFF
--- a/cmake/checks/check_01_cxx_features.cmake
+++ b/cmake/checks/check_01_cxx_features.cmake
@@ -25,6 +25,7 @@
 #   DEAL_II_HAVE_FP_EXCEPTIONS
 #   DEAL_II_HAVE_COMPLEX_OPERATOR_OVERLOADS
 #   DEAL_II_HAVE_CXX17_BESSEL_FUNCTIONS
+#   DEAL_II_HAVE_CXX17_LEGENDRE_FUNCTIONS
 #   DEAL_II_FALLTHROUGH
 #   DEAL_II_DEPRECATED
 #   DEAL_II_CONSTEXPR
@@ -355,6 +356,7 @@ UNSET_IF_CHANGED(CHECK_CXX_FEATURES_FLAGS_SAVED
   DEAL_II_HAVE_CXX17_ATTRIBUTE_FALLTHROUGH
   DEAL_II_HAVE_ATTRIBUTE_FALLTHROUGH
   DEAL_II_HAVE_CXX17_BESSEL_FUNCTIONS
+  DEAL_II_HAVE_CXX17_LEGENDRE_FUNCTIONS
   DEAL_II_CXX14_CONSTEXPR_BUG_OK
   )
 
@@ -551,7 +553,7 @@ ENDIF()
 
 
 #
-# Check for c++17 bessel function support. Unfortunately libc++ version 10
+# Check for c++17 Bessel function support. Unfortunately libc++ version 10
 # does not have those.
 #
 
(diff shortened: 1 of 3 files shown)


masterleinad:
We already use `std::cyl_bessel_j*`. I don't see a reason not to do the same for the Legendre polynomials.

bangerth:
Ah, yes, the hive mind works! I had forgotten about `std::cyl_bessel_j`.

@marcfehling -- can you add the same kind of approach to `include/deal.II/base/std_cxx17/cmath.h` and use that here?

marcfehling:
The `std` functions use the type `unsigned int` for the index parameter, while the `boost` implementation uses `int`.

I provided them as functions with a `static_cast` instead of function pointers like with the Bessel functions. What do you think?

marcfehling:
Blocked by #11286 as we include `cmath.h` twice now :-)

marcfehling:
Grr, this issue also affects the double definition of the Legendre polynomials.
```
 /usr/bin/ld.gold: error: base/CMakeFiles/obj_base_release.dir/function_lib.cc.o: multiple definition of 'dealii::std_cxx17::legendre(unsigned int, double)'
```

I'll think of a fix.

bangerth:
If you implement these as functions, then `constexpr` is the wrong approach. You want `inline` :-)

(As a matter of fact, we also want `inline` variables in #11286, but that's a C++17 feature...)

marcfehling:
> If you implement these as functions, then `constexpr` is the wrong approach. You want `inline` :-)

Ah, since those are actual function and not just function pointers. I get it now :-)

bangerth (approved):

bangerth:
Can you squash the last few commits?

bangerth:
Blocked by #11286. Otherwise ok to merge once squashed.

marcfehling:
Hmm it looks like the CI doesn't like the `constexpr` flag for the function pointers, as the linker complains particularly about the Bessel functions.
```
/usr/bin/ld.gold: error: base/CMakeFiles/obj_base_release.dir/function_lib.cc.o: multiple definition of 'dealii::std_cxx17::cyl_bessel_jl'
/usr/bin/ld.gold: fe/CMakeFiles/obj_fe_release.dir/fe_series_legendre.cc.o: previous definition here
/usr/bin/ld.gold: error: base/CMakeFiles/obj_base_release.dir/function_lib.cc.o: multiple definition of 'dealii::std_cxx17::cyl_bessel_jf'
/usr/bin/ld.gold: fe/CMakeFiles/obj_fe_release.dir/fe_series_legendre.cc.o: previous definition here
/usr/bin/ld.gold: error: base/CMakeFiles/obj_base_release.dir/function_lib.cc.o: multiple definition of 'dealii::std_cxx17::cyl_bessel_j'
/usr/bin/ld.gold: fe/CMakeFiles/obj_fe_release.dir/fe_series_legendre.cc.o: previous definition here
```

Shall we turn the function pointers to wrapper functions as well? Hmm, or maybe it would help to flag them static.

bangerth:
Hm. Can you rebase to master and see again what happens?

masterleinad (commented):
We should use a common style here. I would be fine with `inline` functions.

masterleinad (changes requested):
We should use a common style here. I would be fine with inline functions.

marcfehling:
The CI complains about the `constexpr` and `inline` flags for the function pointers. It works with a `static` flag or when turning them to wrapper functions.

I comply to @masterleinad's suggestion to use the `inline` flag in combination with wrapper functions.

masterleinad (approved):

bangerth:
You'll need to run the indentation script.

marcfehling:
Even the indentation script got stuck. I need to rebase now anyways. Give me a sec

marcfehling:
I've changed the parameter names to be in accordance with this [`cmath.h` reference](https://en.cppreference.com/w/cpp/header/cmath).

Ready for merge.

bangerth (approved):

PR #11286, opened 2020-12-01
Mark some variables as constexpr.

bangerth:
These are global variables, and as such should not actually have been declared in
a header file without marking them as 'extern' and providing a definition in a .cc file.
If the compiler had not inlined the definitions of these variables, we would have
gotten duplicate symbol linker errors. C++11 provides us with a way around this:
mark these variables as 'constexpr', thereby avoiding the bug we currently have.

Seen while considering #11283 . @marcfehling -- FYI.

/rebuild


bangerth:
Possible reason why it worked: The `cmath.h` file in question here is `#include`d in only one .cc file (`function_lib.cc`).

masterleinad (approved):

PR #11288, opened 2020-12-01
Fix VectorTools::internal::project_matrix_free for arbitrary geometric objects

peterrum:
Fixes #11204. Probably not the best solution...

In one of the few applications that uses the new simplex infrastructure, one test has been failing since a deal.II update although it has been passing before.

FYI @mschreter: thanks for letting me know. The quick and dirty solution would have been to use instead of `VectorTools::project` `VectorTools::interpolate()` what I have been using in the simplex tests ;) This function does not have to invert the mass matrix!

PR #11291, opened 2020-12-01
Fix lexicographic_numbering for vector-valued Simplex::FE_Q and FE_DGQ

peterrum:
As discussed in https://github.com/dealii/dealii/pull/11288#discussion_r533409039.

### DIFF
--- a/include/deal.II/matrix_free/shape_info.templates.h
+++ b/include/deal.II/matrix_free/shape_info.templates.h
@@ -77,6 +77,96 @@ namespace internal
     }
 
 
+    template <int dim>
+    void
+    get_element_type_specific_information(
+      const FiniteElement<dim, dim> &fe_in,
+      const FiniteElement<dim, dim> *fe,
+      const unsigned int             base_element_number,
+      ElementType &                  element_type,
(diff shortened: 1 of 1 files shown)


kronbichler (approved):

include/deal.II/matrix_free/shape_info.templates.h
```diff
@@ -77,6 +77,96 @@ namespace internal
...
 
+    template <int dim>
+    void
+    get_element_type_specific_information(
+      const FiniteElement<dim, dim> &fe_in,
+      const FiniteElement<dim, dim> *fe,
+      const unsigned int             base_element_number,
+      ElementType &                  element_type,
+      std::vector<unsigned int> &    scalar_lexicographic,
+      std::vector<unsigned int> &    lexicographic_numbering)
```
bangerth, line 88: @peterrum  I've been looking at this patch for a little (see #11313). Can you add some documentation to this function? It's not clear from the naming of the arguments what the first two arguments represent -- `fe` and `fe_in` doesn't give away much as to how these two finite elements related to each other.

PR #11287, opened 2020-12-01
Enable TriaAccessor::diameter() for wedges

peterrum:
As @bangerth rightfully noticed in https://github.com/dealii/dealii/pull/11285#issue-529923727.

PR #11292, opened 2020-12-01
Test different measures for Wedges

peterrum:
In reference to https://github.com/dealii/dealii/pull/11287#issuecomment-736605154.

For this I need #11267 (`FE_Wedge`) and I also need to have a look at the other functions.


drwells:
Could we combine this and #11306 so that we get tests and code at the same time? If not, then how would you like to proceed?

peterrum:
I have rebased this PR. Ready from my side!

drwells (approved):

PR #11293, opened 2020-12-01
Use inline functions instead of constexpr variables.

bangerth:
As @marcfehling found out in #11283, my patch in #11286 does not actually work. This patch fixes that and lays out the path for #11283 as well!

/rebuild


drwells (approved):
This is much nicer than the function pointer references.

marcfehling:
I already incorporated this in #11283.

I can rebase on this branch if you want to, or feel free to close this PR.

masterleinad (approved):

bangerth:
Hm, the OS X checks seem stuck. Anyone courageous enough to merge anyway?

PR #11295, opened 2020-12-02
Enable GridTools:min/max_diameter() for wedges

peterrum:
Similar to #11271, #11284, #11287. Needed by #11292.


bangerth (approved):
Should we just move this function to `TriaAccessor`? That's a separate patch, but it seems silly to have this in two places.

peterrum:
> Should we just move this function to TriaAccessor? That's a separate patch, but it seems silly to have this in two places

I'll prepare in a minute a follow-up PR!

PR #11301, opened 2020-12-03
Unify diameter implementations

peterrum:
As requested by: https://github.com/dealii/dealii/pull/11295#pullrequestreview-543352867.

FYI @bangerth: Moving the implementation to `TriaAccessor` did not work, since `Mapping` want a `CellAccessor`.

### DIFF
--- a/include/deal.II/grid/tria_accessor.h
+++ b/include/deal.II/grid/tria_accessor.h
@@ -53,6 +53,9 @@ namespace parallel
 
 template <int dim, int spacedim>
 class Manifold;
+
+template <int dim, int spacedim>
+class Mapping;
 #endif
 
 namespace internal
@@ -124,7 +127,6 @@ namespace internal
     {
       using type = int;
     };
-
   } // namespace TriaAccessorImplementation
 } // namespace internal
 template <int structdim, int dim, int spacedim>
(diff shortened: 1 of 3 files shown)


bangerth:
Thanks for working on this!

The way I had imagined this is to add a second member function `TriaAccessor::diameter(const Mapping<dim,spacedim> &)`. Then you could move the new function (currently `inline`) into the `.templates.h` function into an internal namespace, and the function in `GridTools` could just call the new member function. (I think we should on that occasion also deprecate the `GridTools` function.)

peterrum:
The problem is that `Mapping` does not take `TriaAccessor` but a `CellAccessor`. To have an additional method with different arguments in the derived class - i.e. `CellAccessor` - with the same name is not possible, is it?

bangerth:
You can have the overload in `CellAccessor` if you add
```
  using TriaAccessor::diameter;
```
in `CellAccessor` as well to inject the function from the base class there as well.

bangerth:
What's the reason `Mapping` is only taking a `CellAccessor`, though? Could it also deal with just getting a `TriaAccessor`, or does it really need a cell?

peterrum:
@bangerth  Done!

bangerth (approved):

PR #11302, opened 2020-12-03
MatrixFree: precompute subranges

peterrum:
Follow-up PR to #11260. As discussed in https://github.com/dealii/dealii/pull/11260#discussion_r531150297 and https://github.com/dealii/dealii/pull/11260#discussion_r533412552, subranges should be precomputed in `MatrixFree`.

### DIFF
--- a/include/deal.II/matrix_free/matrix_free.h
+++ b/include/deal.II/matrix_free/matrix_free.h
@@ -1472,32 +1472,7 @@ public:
     const unsigned int                           dof_handler_index = 0) const;
 
   /**
-   * In the hp-adaptive case, a subrange of internal faces as computed during
-   * loop() might contain internal faces with elements of different active
-   * FE indices. Use this function to compute what the subrange for a given pair
-   * of active FE indices is.
-   */
-  std::pair<unsigned int, unsigned int>
-  create_inner_face_subrange_hp_by_index(
-    const std::pair<unsigned int, unsigned int> &range,
-    const unsigned int                           fe_index_interior,
-    const unsigned int                           fe_index_exterior,
-    const unsigned int                           dof_handler_index = 0) const;
-
-  /**
-   * In the hp-adaptive case, a subrange of boundary faces as computed during
-   * loop() might contain boundary faces with elements of different active
-   * FE indices. Use this function to compute what the subrange for a given
-   * active FE indices is.
-   */
-  std::pair<unsigned int, unsigned int>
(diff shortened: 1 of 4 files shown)


include/deal.II/matrix_free/matrix_free.h
```diff
@@ -4652,11 +4652,34 @@ namespace internal
           }
     }
 
-    // Runs the assembler on interior faces. If no function is given, nothing
-    // is done
     virtual void
-    face(const std::pair<unsigned int, unsigned int> &face_range) override
+    cell(const unsigned int range_index) override
```
peterrum, line 4656: @kronbichler  The PR is not ready yet. But I would like to have your opinion on the approach, since it changes the internal interface of `MFWorkerInterface`. I would like to pass here a `range_index`/`partition_index` that allows me later to efficiently to access some new vectors in `TaskInfo` (containing the pre-computed info).

kronbichler: I fully agree with this change, so please go ahead. This is internal and just used for driving the loops, so I see no need to stick with an interface. One thing I wonder is whether the other `cell` function is still needed or if we can route everything in the new function.

peterrum (commented):
@kronbichler I had some time to work on this PR. The subranges are now pre-computed. For this, the functions `create_cell_subrange_hp_by_index()`,  `create_inner_face_subrange_hp_by_index()`, `create_boundary_face_subrange_hp_by_index` are used. Since, these functions should be used only internally, we could deprecate the first one and inline all functions (and by that remove the recently newly introduced functions). What do you think?

include/deal.II/matrix_free/task_info.h
```diff
@@ -496,6 +501,36 @@ namespace internal
        */
       std::vector<unsigned int> ghost_face_partition_data;
 
+      /**
+       * TODO
```
kronbichler, line 505: Can you please document this member (and the others below)?

kronbichler (commented):
I think this is close to being merged (sorry for taking so long to review). Let us add documentation as indicated below.

include/deal.II/matrix_free/task_info.h
peterrum: Done!

kronbichler (approved):

kronbichler (approved):

peterrum:
The failing test is unrelated.

kronbichler (approved):

PR #11303, opened 2020-12-03
AD doc typos

elauksap:
Just a few typos in the `AD` namespace.

- `ScalarFunction`: the variable is named `psi` instead of `psi_CH`.
- `ResidualLinearization`: the example does not use taped numbers, so referencing to `stop_recording_operations()` may be misleading to some users.


bangerth (approved):

bangerth:
The OS X checks fail, but for reasons that have nothing to do with this patch. It's only documentation, so I'll merge.

@elauksap Thanks for helping improve the documentation!

Issue #11272, opened 2020-11-28
ENABLE_IF_SUPPORTED macro is incompatible with cmake-3.19


tamiko:
CMake 3.19 changed semantics and now seems to quote the flags passed to `CheckCXXCompilerFlag`. Unfortunately we prepend every flag we test with `-Werror `.

The resulting option `"-Werror <flag>"` is then interpreted as one argument by clang and configure fails.

PR #11304, opened 2020-12-03
[WIP] OSX: use c++14

peterrum:
Currently, all OSX tests are failing.

### DIFF
--- a/cmake/macros/macro_enable_if_supported.cmake
+++ b/cmake/macros/macro_enable_if_supported.cmake
@@ -28,7 +28,7 @@ MACRO(ENABLE_IF_SUPPORTED _variable _flag)
   # actual errors with the -Werror switch:
   #
   IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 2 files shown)


tjhei:
This is due to cmake 3.19 and #11272

cmake/macros/macro_enable_if_supported.cmake
```diff
@@ -28,7 +28,7 @@ MACRO(ENABLE_IF_SUPPORTED _variable _flag)
   # actual errors with the -Werror switch:
   #
   IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    SET(_werror_string "-Werror ")
+    SET(_werror_string "")
   ELSE()
     SET(_werror_string "")
```
bangerth, line 33: You don't need the if-else if you set `_werror_string` to the same value regardless :-)

peterrum: I was only trying out something ;) #11305 seems to work for now. So I am closing this PR.

PR #11306, opened 2020-12-03
Implement TriaAccessor::measure() for wedges

peterrum:
Thanks to @drwells! Will be tested in #11292.


peterrum:
@drwells Let's merge first this PR and then #11292.

drwells (approved):

PR #11307, opened 2020-12-03
MatrixFreeTools::compute_diagonal: accept class and function pointer

peterrum:
@kronbichler I would like to be able to hide the `-1` and `0` ;)


include/deal.II/matrix_free/tools.h
```diff
@@ -71,6 +71,31 @@ namespace MatrixFreeTools
...
+            int fe_degree,
+            int n_q_points_1d,
+            int n_components,
+            typename Number,
+            typename VectorizedArrayType>
+  void
+  compute_diagonal(
+    const MatrixFree<dim, Number, VectorizedArrayType> &matrix_free,
+    LinearAlgebra::distributed::Vector<Number> &        diagonal_global,
+    void (CLASS::*cell_operation)(FEEvaluation<dim,
```
bangerth, line 88: I would much rather get away from this style where we have the class name as a template argument and pass a `this` pointer to a class. That's old-style programming. Let's pass `std::function<>` objects instead.

peterrum: We support both this style and the `std::function` style both in `MatrixFree` (and this will not change soon). This PR enables to work consistently in both cases. 

BTW: there has been already an implementation of `MatrixFreeTools::compute_diagonal()` that takes `std::function` for months.

kronbichler: > That's old-style programming. Let's pass std::function<> objects instead.

Since I'm the culprit regarding that interface in the main matrix-free class, I would add that as long as the additional maintenance burden is not too high, I see value in that kind of interface. It goes without saying that the `std::function` interface is useful in many situations. Yet, I see value in being able to easily pass in member functions without explicitly wrapping them into an `std::function` as that avoids some repetitive code on the user's side, at least for a function that takes four arguments.

And I guess we do not disagree on the value of having properly named member functions in terms of structuring and organizing bigger functions, so at least I would want to support the user with such a programming style. If you look at the generated code, for example because you are interested in studying the assembly code, `std::function` objects lead to some heavy spaghetti code at least for gcc, much worse than named functions.

kronbichler (approved):

peterrum:
The failing test is unrated:
```
The following tests FAILED:

	3804 - multigrid/transfer_matrix_free_06.mpirun=15.debug (Timeout)
```

kronbichler (approved):

PR #11312, opened 2020-12-03
Make clear that 'throw' is not a function.

bangerth:
I was looking at the fact that C++17 has removed the `throw()` specification one used to be able to attach to function declarations (the old-style equivalent of saying `noexcept`) and wondering whether we actually use that anywhere. We don't, but I found a couple of places where we throw exceptions and use a syntax that might suggest that `throw` is actually a function -- which it isn't, it's a keyword that is followed by an expression.

Make this a bit easier to read.

/rebuild


bangerth:
Reference: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0003r5.html

masterleinad (approved):
LGTM

PR #11313, opened 2020-12-03
Avoid pointers, use references instead

bangerth:
Follow-up to #11291. @peterrum FYI.

/rebuild

### DIFF
--- a/include/deal.II/matrix_free/shape_info.templates.h
+++ b/include/deal.II/matrix_free/shape_info.templates.h
@@ -254,16 +254,16 @@ namespace internal
 
           AssertDimension(dim, dim_q);
 
-          const auto quad            = Quadrature<dim>(quad_in);
-          const auto fe              = &fe_in.base_element(base_element_number);
+          const auto  quad           = Quadrature<dim>(quad_in);
+          const auto &fe             = fe_in.base_element(base_element_number);
           n_dimensions               = dim;
           n_components               = fe_in.n_components();
           n_q_points                 = quad.size();
-          dofs_per_component_on_cell = fe->n_dofs_per_cell();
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/shape_info.templates.h
```diff
@@ -81,20 +81,20 @@ namespace internal
...
+      const FiniteElement<dim, dim> &fe,
       const unsigned int             base_element_number,
       ElementType &                  element_type,
       std::vector<unsigned int> &    scalar_lexicographic,
       std::vector<unsigned int> &    lexicographic_numbering)
     {
       element_type = tensor_general;
 
-      const auto fe_poly = dynamic_cast<const FE_Poly<dim, dim> *>(fe);
+      const auto fe_poly = dynamic_cast<const FE_Poly<dim, dim> *>(&fe);
```
masterleinad, line 92: It should also work to cast as reference here.
```suggestion
      const auto& fe_poly = dynamic_cast<const FE_Poly<dim, dim> &>(fe);
```

drwells: I'm not sure this will work - we later check if `fe_poly` is not `nullptr` so this would result in an exception in a condition where the code currently executes correctly.

masterleinad: OK. I missed that we check for it being `nullptr`.

bangerth: Yes, correct: If the `dynamic_cast` fails for a pointer, you get a `nullptr`. If it fails for a reference, you get an exception -- that's not what we want.

bangerth:
I rebased to make the merge conflict go away.

drwells (approved):

PR #11317, opened 2020-12-04
Add a Poisson test on pyramid grid

peterrum:
~At least it contains a grid right now...~

The output looks like this:
pyramid
... I'll extract the relevant feature into separate PRs so that this PR only contains the test.

### DIFF
--- a/tests/simplex/poisson_01.cc
+++ b/tests/simplex/poisson_01.cc
@@ -49,6 +49,7 @@
 #include <deal.II/lac/trilinos_sparsity_pattern.h>
 
 #include <deal.II/numerics/data_out.h>
+#include <deal.II/numerics/vector_tools.h>
 
 #include <deal.II/simplex/fe_lib.h>
 #include <deal.II/simplex/grid_generator.h>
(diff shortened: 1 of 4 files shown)


tests/simplex/poisson_01.cc
```diff
@@ -250,6 +251,8 @@ test(const Triangulation<dim, spacedim> &tria,
   SolverCG<VectorType> solver(solver_control);
   solver.solve(system_matrix, solution, system_rhs, PreconditionIdentity());
 
+  deallog << solution.l2_norm() << " " << system_rhs.l2_norm() << std::endl;
+
```
kronbichler, line 255: Do we need this?

tests/simplex/poisson_01.cc
```diff
@@ -250,6 +251,8 @@ test(const Triangulation<dim, spacedim> &tria,
...
   deallog << "   with " << solver_control.last_step()
           << " CG iterations needed to obtain convergence" << std::endl;
```
kronbichler, line 257: Same here, technically the output is already via deallog.

tests/simplex/simplex_grids.h
```diff
@@ -106,25 +106,121 @@ namespace dealii
 
     template <int dim, int spacedim>
     void
-    subdivided_hyper_cube_with_wedges(Triangulation<dim, spacedim> &tria,
-                                      const unsigned int            repetitions,
-                                      const double                  p1 = 0.0,
-                                      const double                  p2 = 1.0,
```
kronbichler, line 112: As the failing test in the CI shows, we still need this function in `simplex/matrix_free_01`.

kronbichler (approved):

PR #11318, opened 2020-12-04
Reshuffle code in SolverCG

peterrum:


include/deal.II/lac/solver_cg.h
```diff
@@ -341,61 +341,83 @@ SolverCG<VectorType>::solve(const MatrixType &        A,
...
+              d.sadd(beta, -1., g);
+            }
+        }
+      else
+        {
+          if (std::is_same<PreconditionerType, PreconditionIdentity>::value ==
+              false)
+            {
+              preconditioner.vmult(h, g);
+
```
tjhei, line 406: remove spaces here?

include/deal.II/lac/solver_cg.h
```diff
@@ -341,61 +341,83 @@ SolverCG<VectorType>::solve(const MatrixType &        A,
...
 
-  if (std::is_same<PreconditionerType, PreconditionIdentity>::value == false)
+  while (conv == SolverControl::iterate)
     {
-      preconditioner.vmult(h, g);
+      it++;
+      number old_alpha = alpha;
 
-      d.equ(-1., h);
+      if (it > 1)
```
tjhei, line 379: Is it a useful change to move the it==1 case from before to inside the loop? It looks more complicated to me.

peterrum:
@tjhei Do you remember what we did with the warning?

```
In file included from /jenkins/workspace/dealii_PR-11318/include/deal.II/lac/precondition.h:33:0,

                 from /jenkins/workspace/dealii_PR-11318/include/deal.II/numerics/vector_tools_boundary.templates.h:34,

                 from /jenkins/workspace/dealii_PR-11318/source/numerics/vector_tools_boundary.cc:17:

/jenkins/workspace/dealii_PR-11318/include/deal.II/lac/solver_cg.h: In member function ‘void dealii::SolverCG<VectorType>::solve(const MatrixType&, VectorType&, const VectorType&, const PreconditionerType&) [with MatrixType = dealii::SparseMatrix<std::complex<float> >; PreconditionerType = dealii::PreconditionSSOR<dealii::SparseMatrix<std::complex<float> > >; VectorType = dealii::Vector<std::complex<float> >]’:

/jenkins/workspace/dealii_PR-11318/include/deal.II/lac/solver_cg.h:433:11: error: ‘old_alpha.std::complex<float>::_M_value’ may be used uninitialized in this function [-Werror=maybe-uninitialized]

           this->coefficients_signal(old_alpha, beta);

           ^

/jenkins/workspace/dealii_PR-11318/include/deal.II/lac/solver_cg.h: In member function ‘void dealii::SolverCG<VectorType>::solve(const MatrixType&, VectorType&, const VectorType&, const PreconditionerType&) [with MatrixType = dealii::SparseMatrix<std::complex<double> >; PreconditionerType = dealii::PreconditionSSOR<dealii::SparseMatrix<std::complex<double> > >; VectorType = dealii::Vector<std::complex<double> >]’:

/jenkins/workspace/dealii_PR-11318/include/deal.II/lac/solver_cg.h:433:11: error: ‘old_alpha.std::complex<double>::_M_value’ may be used uninitialized in this function [-Werror=maybe-uninitialized]

cc1plus: all warnings being treated as errors
```

tjhei:
Can you play with the initialization of ``alpha`` or ``old_alpha``?

Something like ``number alpha = number();``

peterrum:
> Something like number alpha = number();

I already had that...

kronbichler:
So now we are successful with a different set of compile flags, `-Og` vs `-O1`. This indicates (as you hypothesized already) that we are looking at a compile error for which I find it questionable to work around at this point. I would not care much if the flags are changed to `-Og` if it works with our current master in order to track the PRs for errors.

That topic aside, there are now a few failures in the serial tests related to this PR, so please have a look there (and maybe look out whether there are other tests not run in the CI serial version which use SolverCG).

tjhei:
> This indicates (as you hypothesized already) that we are looking at a compile error for which I find it questionable to work around at this point. I would not care much if the flags are changed to `-Og` if it works with our current master in order to track the PRs for errors.

I agree that this is annoying having to work around the compiler here. We did switch to O1 for a reason though, see https://github.com/dealii/dealii/pull/10786
I think going back to Og might be acceptable (while O0 is not).

jppelteret:
>  Something like number alpha = number();

> I already had that...

Have you tried [the big stick](https://github.com/dealii/dealii/blob/master/include/deal.II/base/numbers.h#L749), `number alpha = internal::NumberType<number>::value(0.0);`?

kronbichler:
We appear to have consensus on a possible change from `-O1` to `-Og`, so I suggest we rebase this PR on master to see that the new setting still works without giving new errors. If we get errors and stay with `-O1`, we should try the suggestion by @jppelteret .

peterrum:
> We appear to have consensus on a possible change from -O1 to -Og

The change is already part of this branch. But I will be nevertheless rebase to be on a safe side.

kronbichler:
> The change is already part of this branch. 

I know, all I said is we must ensure that master is still fine with this change.

kronbichler (approved):

include/deal.II/lac/solver_cg.h
kronbichler: Yes, let us remove the spaces here to make the code more readable.

peterrum: This is how it "used" to be. Should I nevertheless remove this line?

kronbichler: Yes, I think it is nicer now that it is inlined within the loop.

kronbichler (approved):

PR #11322, opened 2020-12-05
Add assert to extract_vtk_patch_info

peterrum:
to prevent `write_higher_order_cells` with non-hypercube cells.


Rombur (approved):

Rombur:
The cuda failure is not related to this PR (github could download a tarball). So it's fine to merge with this test failing.

PR #11326, opened 2020-12-05
Implement ScalarPyramidPolynomial and FE_PyramidP

peterrum:


source/simplex/fe_lib.cc
```diff
@@ -559,6 +585,158 @@ namespace Simplex
...
 
+  template <int dim, int spacedim>
+  FE_PyramidP<dim, spacedim>::FE_PyramidP(const unsigned int degree)
+    : dealii::FE_Poly<dim, spacedim>(
+        Simplex::ScalarPyramidPolynomial<dim>(degree),
+        FiniteElementData<dim>(get_dpo_vector_fe_pyramid(degree),
+                               ReferenceCell::Type::Pyramid,
+                               1,
+                               degree,
+                               FiniteElementData<dim>::L2),
```
drwells, line 596:
```suggestion
                               FiniteElementData<dim>::H1),
```

source/simplex/polynomials.cc
```diff
@@ -711,12 +726,255 @@ namespace Simplex
...
+    if (i == 0)
+      return Q14 * ((1.0 - r) * (1.0 - s) - t + ration);
+    if (i == 1)
+      return Q14 * ((1.0 + r) * (1.0 - s) - t - ration);
+    if (i == 2)
+      return Q14 * ((1.0 - r) * (1.0 + s) - t - ration);
+    if (i == 3)
+      return Q14 * ((1.0 + r) * (1.0 + s) - t + ration);
+    else
+      return t;
```
drwells, line 770: Ah - this is consistent with #11327. I would still prefer to use `[0, 1]`.

drwells (changes requested):
I found one minor bug.

source/simplex/fe_lib.cc
```diff
@@ -559,6 +585,158 @@ namespace Simplex
...
+        this->unit_support_points.emplace_back(-1.0, -1.0, 0.0);
+        this->unit_support_points.emplace_back(+1.0, -1.0, 0.0);
+        this->unit_support_points.emplace_back(-1.0, +1.0, 0.0);
+        this->unit_support_points.emplace_back(+1.0, +1.0, 0.0);
+        this->unit_support_points.emplace_back(+0.5, +0.5, 1.0);
+
+        // TODO
+        this->unit_face_support_points[0].emplace_back(1.0, 0.0);
+        this->unit_face_support_points[0].emplace_back(0.0, 1.0);
+        this->unit_face_support_points[0].emplace_back(0.0, 0.0);
```
drwells, line 626: What are these doing here?

source/simplex/fe_lib.cc
```diff
@@ -559,6 +585,158 @@ namespace Simplex
...
+    AssertDimension(dim, 3);
+
+
+    if (degree == 1)
+      {
+        this->unit_support_points.emplace_back(-1.0, -1.0, 0.0);
+        this->unit_support_points.emplace_back(+1.0, -1.0, 0.0);
+        this->unit_support_points.emplace_back(-1.0, +1.0, 0.0);
+        this->unit_support_points.emplace_back(+1.0, +1.0, 0.0);
+        this->unit_support_points.emplace_back(+0.5, +0.5, 1.0);
```
drwells, line 621: This point seems wrong - shouldn't the `z = 1` point be at the origin if we use `[-1, 1]`?

source/simplex/fe_lib.cc
peterrum: You are right!

source/simplex/fe_lib.cc
peterrum: Indeed.

source/simplex/fe_lib.cc
peterrum: I have removed this part. We will need some reference points once we are using `FEFaceValues`. But for that we need some more complicated procedure to distinguish the different types of faces.

drwells (approved):

drwells:
Like the others, one windows tester failed for reasons unrelated to this PR.

PR #11327, opened 2020-12-05
Add QGaussPyramid

peterrum:


source/simplex/quadrature_lib.cc
```diff
@@ -168,6 +168,50 @@ namespace Simplex
...
+  {
+    AssertDimension(dim, 3);
+
+    if (n_points_1D == 1)
+      {
+        const double Q14 = 1.0 / 4.0;
+        const double Q43 = 4.0 / 3.0;
+
+        this->quadrature_points.emplace_back(0, 0, Q14);
+        this->weights.emplace_back(Q43);
```
drwells, line 185: The area of a pyramid is `b * h / 3` where b is the area of the base - does this mean that the reference element is defined on `[-1, 1] x [-1, 1]`? The wedge uses `[0, 1]` so this doesn't seem right to me.

drwells: If the pyramid's base is `[0, 1] x [0, 1]` and it is `1` tall then the weight should be `1/3`, right?

drwells:
If I recall correctly we decided to make the wedge use `[0, 1]` so I would prefer to do that for pyramids too.

source/simplex/quadrature_lib.cc
peterrum: Yes, it is defined on the the `[-1, 1] x [-1, 1]` base. I agree that it would be more consistent to have the base defined on `[0, 1] x [0, 1]`, but this definition I have found. I am happy to change the definition of the finite element and the quadrature rule for pyramids later. But for now, I would appreciate if we would have a passing test for pyramids on master.

drwells (approved):

kronbichler (approved):

source/simplex/quadrature_lib.cc
kronbichler: Let us postpone this for later.

PR #11328, opened 2020-12-05
Enable GridOut::write_vtk() for pyramids

peterrum:


source/grid/grid_out.cc
```diff
@@ -3217,11 +3217,27 @@ GridOut::write_vtk(const Triangulation<dim, spacedim> &tria,
...
+                  break;
+                case ReferenceCell::Type::Tri:
+                case ReferenceCell::Type::Tet:
+                case ReferenceCell::Type::Wedge:
+                  out << cell->vertex_index(i);
+                  break;
+                case ReferenceCell::Type::Pyramid:
+                  static std::array<unsigned int, 5> table{{0, 1, 3, 2, 4}};
+                  out << cell->vertex_index(table[i]);
+                  break;
```
drwells, line 3237: MSVC's error message is hard to understand but I think the root cause of the complaint is that we need brackets around this case (we cannot define variables in switch statement cases without brackets). Could you try that?

bangerth: Yes, having local variables in `case` statements is almost always an error because the variable remains live in later `case` labels (remember that `case` is just a label, it doesn't introduce a scope) and jumping to these later `case` labels bypasses initialization of the local variable.

Just add `{...}` to the code block here. While there, maybe do the same for the other block as well. (And also, mark the `std::array` as `const` :-) )

peterrum: Done :)

drwells (approved):

PR #11336, opened 2020-12-07
Fix a data type issue.

bangerth:
The issue is that `PetscInt` can also be 64-bit, which would lead to an invalid `reinterpret_cast`.

/rebuild

### DIFF
--- a/source/lac/slepc_solver.cc
+++ b/source/lac/slepc_solver.cc
@@ -184,9 +184,16 @@ namespace SLEPcWrappers
     ierr = EPSSolve(eps);
     AssertThrow(ierr == 0, ExcSLEPcError(ierr));
 
-    // get number of converged eigenstates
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
LGTM.

tjhei (approved):

tjhei:
Interesting how the reinterpret_cast should have told us that something is wrong here.

bangerth:
It's only wrong if you configure PETSc with 64-bit integers *and* *also* have SLEPc installed.

tjhei:
> It's only wrong if you configure PETSc with 64-bit integers _and_ _also_ have SLEPc installed.

Sure. My point is that somebody inserted a reinterpret_cast, probably to get rid of a warning. At that moment it should have been clear that something might go wrong.

PR #11339, opened 2020-12-08
Avoid MPI calls in the HDF5 destructors when there is an exception

dangars:
This is work in progress.

@tjhei  As we discussed in #10737 I placed the code with `std::uncaught_exception` in the deleters of `hdf5_reference` and `dataset`. But I think that is not enough. The test from this PR still hangs.

The structure of the test is
```
Open HDF5::File, HDF5::Group and HDF5::Dataset

AssertThrow(Utilities::MPI::this_mpi_process(mpi_communicator) == 0,
                     ExcInternalError());

Do some operations with HDF5::Dataset
```

I think that I have to place `std::uncaught_exception` in other places. I have to do some debugging. I'll come back to you when I have more results.

### DIFF
--- a/source/base/hdf5.cc
+++ b/source/base/hdf5.cc
@@ -105,18 +105,80 @@ namespace HDF5
     , global_no_collective_cause(H5D_MPIO_SET_INDEPENDENT)
   {
     hdf5_reference = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
+      // std::uncaught_exception() is deprecated in c++17
+      if (std::uncaught_exceptions() == 0)
+#    else
+      if (std::uncaught_exception() == false)
+#    endif
+        {
+          // Release the HDF5 resource
+          const herr_t ret = H5Dclose(*pointer);
+          AssertNothrow(ret >= 0, ExcInternalError());
+          (void)ret;
+          delete pointer;
+        }
+      else
+        {
+          std::cerr
+            << "---------------------------------------------------------\n"
+            << "HDF5 DataSet objects call H5Dclose to end access to the\n"
+            << "dataset and release resources used by it. This call\n"
+            << "requires MPI synchronization. Since an exception is\n"
+            << "currently uncaught, this synchronization will be skipped\n"
+            << "to avoid a possible deadlock. As a result the HDF5 file\n"
+            << "might be corrupted.\n"
+            << "---------------------------------------------------------"
+            << std::endl;
+
+          MPI_Abort(MPI_COMM_WORLD, 1);
+        }
+#  else
       // Release the HDF5 resource
       const herr_t ret = H5Dclose(*pointer);
       AssertNothrow(ret >= 0, ExcInternalError());
       (void)ret;
       delete pointer;
+#  endif
     });
-    dataspace      = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
+    dataspace = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
+      // std::uncaught_exception() is deprecated in c++17
+      if (std::uncaught_exceptions() == 0)
+#    else
+      if (std::uncaught_exception() == false)
+#    endif
+        {
+          // Release the HDF5 resource
+          const herr_t ret = H5Sclose(*pointer);
+          AssertNothrow(ret >= 0, ExcInternalError());
+          (void)ret;
+          delete pointer;
+        }
+      else
+        {
+          std::cerr
+            << "---------------------------------------------------------\n"
+            << "HDF5 DataSet objects call H5Sclose to end access to the\n"
+            << "dataspace and release resources used by it. This call\n"
+            << "requires MPI synchronization. Since an exception is\n"
+            << "currently uncaught, this synchronization will be skipped\n"
+            << "to avoid a possible deadlock. As a result the HDF5 file\n"
+            << "might be corrupted.\n"
+            << "---------------------------------------------------------"
+            << std::endl;
+
+          MPI_Abort(MPI_COMM_WORLD, 1);
+        }
+#  else
       // Release the HDF5 resource
       const herr_t ret = H5Sclose(*pointer);
       AssertNothrow(ret >= 0, ExcInternalError());
       (void)ret;
       delete pointer;
+#  endif
     });
 
     *dataspace = H5Screate_simple(rank, dimensions.data(), nullptr);
@@ -279,11 +341,42 @@ namespace HDF5
     : HDF5Object(name, mpi)
   {
     hdf5_reference = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
+      // std::uncaught_exception() is deprecated in c++17
+      if (std::uncaught_exceptions() == 0)
+#    else
+      if (std::uncaught_exception() == false)
+#    endif
+        {
+          // Release the HDF5 resource
+          const herr_t ret = H5Gclose(*pointer);
+          AssertNothrow(ret >= 0, ExcInternalError());
+          (void)ret;
+          delete pointer;
+        }
+      else
+        {
+          std::cerr
+            << "---------------------------------------------------------\n"
+            << "HDF5 Group objects call H5Gclose to end access to the\n"
+            << "group and release resources used by it. This call\n"
+            << "requires MPI synchronization. Since an exception is\n"
+            << "currently uncaught, this synchronization will be skipped\n"
+            << "to avoid a possible deadlock. As a result the HDF5 file\n"
+            << "might be corrupted.\n"
+            << "---------------------------------------------------------"
+            << std::endl;
+
+          MPI_Abort(MPI_COMM_WORLD, 1);
+        }
+#  else
       // Release the HDF5 resource
       const herr_t ret = H5Gclose(*pointer);
       AssertNothrow(ret >= 0, ExcInternalError());
       (void)ret;
       delete pointer;
+#  endif
     });
     switch (mode)
       {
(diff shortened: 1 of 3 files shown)


source/base/hdf5.cc
```diff
@@ -105,18 +106,74 @@ namespace HDF5
     , global_no_collective_cause(H5D_MPIO_SET_INDEPENDENT)
   {
     hdf5_reference = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
-      // Release the HDF5 resource
-      const herr_t ret = H5Dclose(*pointer);
-      AssertNothrow(ret >= 0, ExcInternalError());
-      (void)ret;
-      delete pointer;
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
```
bangerth, line 110: We now require C++14, so you can remove this `#ifdef` here and the whole `#else` part. That will make things simpler to read too.

source/base/hdf5.cc
```diff
@@ -105,18 +106,74 @@ namespace HDF5
...
     });
-    dataspace      = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
-      // Release the HDF5 resource
-      const herr_t ret = H5Sclose(*pointer);
-      AssertNothrow(ret >= 0, ExcInternalError());
-      (void)ret;
-      delete pointer;
+    dataspace = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
```
bangerth, line 145: same here

source/base/hdf5.cc
```diff
@@ -279,11 +336,39 @@ namespace HDF5
     : HDF5Object(name, mpi)
   {
     hdf5_reference = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
-      // Release the HDF5 resource
-      const herr_t ret = H5Gclose(*pointer);
-      AssertNothrow(ret >= 0, ExcInternalError());
-      (void)ret;
-      delete pointer;
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
```
bangerth, line 340: and here

source/base/hdf5.cc
```diff
@@ -367,11 +452,39 @@ namespace HDF5
     : Group(name, mpi)
   {
     hdf5_reference = std::shared_ptr<hid_t>(new hid_t, [](hid_t *pointer) {
-      // Release the HDF5 resource
-      const herr_t err = H5Fclose(*pointer);
-      AssertNothrow(err >= 0, ExcInternalError());
-      (void)err;
-      delete pointer;
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
```
bangerth, line 456: and here

tests/base/hdf5_07.cc
```diff
@@ -0,0 +1,169 @@
...
+                           mpi_communicator);
+
+      auto                       group = data_file.create_group("test_group");
+      const std::vector<hsize_t> dataset_dimensions = {3, 2, 4};
+      auto                       dataset =
+        group.create_dataset<Number>(dataset_name, dataset_dimensions);
+
+      // All the processes throw an exception except mpi_communicator == 0
+      AssertThrow(Utilities::MPI::this_mpi_process(mpi_communicator) == 0,
+                  ExcInternalError());
```
bangerth, line 74: nice!

bangerth (changes requested):
Oh, MPI communication in destructors in the presence of exceptions :-( I know that I dealt with this in the past, but your description of how HDF5 doesn't work with `find_active_cell_around_point()` just didn't click and I never made the connection...

dangars:
@drwells helped me to find the problem :) The discussion can be found here #10746

The destructor of the HDF5 interface does MPI synchronization. This PR solves this problem.

But there is an additional problem. The HDF5 write operations are collective. This means that they call `MPI_Allreduce()`. I don't know how to solve this problem. Process 0 can arrive to `H5DWrite/MPI_Allreduce()` before process 1 has thrown the exception.

This PR without solving the HDF5 write communication problem is not very useful.

Should we assume that HDF5 can potentially freeze with exceptions? One option would be to provide a compile time option that replaces Assert by a hard terminate, that could facilitate debugging. Let me know if you have another idea.

The structure/backtrace of the test is:
```
Open HDF5::File, HDF5::Group and HDF5::Dataset

AssertThrow(Utilities::MPI::this_mpi_process(mpi_communicator) == 0,
                     ExcInternalError());

call HDF5::DataSet::write_selection()
   call H5DWrite()
        call MPI_Allreduce() -> The program hangs here
```

dangars:
Edit to my last comment, I just realized that in the debugger process 1 terminates, but process 0 hangs. I don't think that my proposition of replacing `Assert` by a hard terminate could help debugging

tjhei:
> But there is an additional problem.

I don't quite understand. Does the throwing process abort or does it hang anywhere? It should be completely fine if some processes hang inside an MPI function as long as one of the processes aborts.
If this doesn't work, can you put something like this into your main()?
```
try
{ 
  // code
}
catch (...)
{
 MPI_Abort(MPI_COMM_WORLD, 1);
}
```
and check if it makes a difference?

tests/base/hdf5_07.cc
```diff
@@ -0,0 +1,174 @@
...
+                << "----------------------------------------------------"
+                << std::endl;
+      std::cerr << "Unknown exception!" << std::endl
+                << "Aborting!" << std::endl
+                << "----------------------------------------------------"
+                << std::endl;
+
+      // The error code is equal to 0 in order to pass the test
+      MPI_Abort(MPI_COMM_WORLD, 0);
+      return 0;
```
dangars, line 172: Here the return value is equal to zero to pass the test.

dangars (commented):
@tjhei Thanks, if I add `MPI_Abort()` it works.

The throwing process aborts, the other hangs. But if I add `MPI_Abort(MPI_COMM_WORLD, 1)` in the catch section all the processes abort as expected. Now the PR is ready for review. I also modified the HDF5 documentation.

The output name of the test is
```
tests/base/hdf5_07.with_hdf5=true.expect=diff.mpirun=4.output
```
This way the test will pass if it runs. There is no need to check the diff.

bangerth:
There are some fundamental assumptions one must live with when doing parallel programming. One of those is that all processes get to parallel communication points at some point or other. If one can show that they will not, then the only reasonable way is to abort the program. What you found is one of these cases: An object wants to do some communication in the destructor and finds out that it got into this destructor not as part of the usual course of action, but because there is a pending exception and we're currently unwinding the stack. Under the assumption that this is a situation specific to this one processor, and not happening on other processors, the only reasonable action is to abort the program.

A corollary is that the actual bug is that the exception wasn't caught and dealt with before we ended up in this destructor. What you're doing here is really not fixing anything, but to just produce a reasonable error message. There are many ways to trigger a similar bug, most of which are not amenable to giving reasonable error messages. For example, this code has a bug where all but processor zero will hang (and processor zero will eventually hang in its next attempt at communication):
```
  void foo () 
  {
    if (MPI::this_mpi_rank(MPI_COMM_WORLD) == 0)
       throw 42;

    mpi_allreduce (MPI_COMM_WORLD, ...);
  }
```
There just isn't anything we can do in this case.

source/base/hdf5.cc
```diff
@@ -105,18 +106,64 @@ namespace HDF5
...
+              << "---------------------------------------------------------\n"
+              << "HDF5 DataSet objects call H5Dclose to end access to the\n"
+              << "dataset and release resources used by it. This call\n"
+              << "requires MPI synchronization. Since an exception is\n"
+              << "currently uncaught, this synchronization will be skipped\n"
+              << "to avoid a possible deadlock. As a result the HDF5 file\n"
+              << "might be corrupted.\n"
+              << "---------------------------------------------------------"
+              << std::endl;
+        }
```
bangerth, line 125: I would prefer if you just called `MPI_Abort()` here. The reason is that if you don't, then all of the other processes will hang because this one processor did not participate in the communication. This one processor will then simply go on with its business, will eventually come to a point where it wants to communicate, and hang there because none of the other processors have come to that point. You're really not creating a better outcome, but just postponing the misery.

So, I'd just abort. You then also don't know the paragraph about data corruption in the documentation.

bangerth (changes requested):

tjhei:
There is also ``AssertNothrow()`` which calls ``do_issue_error_nothrow``, which calls ``MPI_Abort``:
https://github.com/dealii/dealii/blob/9df0655e6f72b7f524b7fe112aac789cee2acbb9/source/base/exceptions.cc#L439-L470

source/base/hdf5.cc
```diff
@@ -105,18 +106,64 @@ namespace HDF5
...
+      else
+        {
+          // Release the HDF5 resource
+          const herr_t ret = H5Dclose(*pointer);
+          AssertNothrow(ret >= 0, ExcInternalError());
+          (void)ret;
+          delete pointer;
+        }
+#  else
+        AssertThrow(false, ExcMessage("MPI support is disabled."));
```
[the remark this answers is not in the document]
dangars: Good catch! Your are right :)

dangars:
Now all the MPI processes output the error message and they call `MPI_Abort()` in the destructor if there is a uncaught exception.

I removed the paragraph about HDF5 and exceptions. It is no longer needed.

I removed the test, I don't think that there is a need for test. Let me know if I should bring it back.

I copied the error code for `MPI_Abort()` from `AssertNothrow()`. Why is it 255 instead of 1 ?
```
MPI_Abort(MPI_COMM_WORLD,  255);
```

tjhei:
> Why is it 255 instead of 1 ?

It is the return value. I don't think that matters at all.

bangerth (approved):
This looks good to me now. Thanks!

bangerth:
We have the following error:
```
/jenkins/workspace/dealii_PR-11339/source/base/hdf5.cc:109:11: error: ‘uncaught_exceptions’ is not a member of ‘std’

       if (std::uncaught_exceptions() > 0)

           ^

/jenkins/workspace/dealii_PR-11339/source/base/hdf5.cc: In lambda function:

/jenkins/workspace/dealii_PR-11339/source/base/hdf5.cc:142:11: error: ‘uncaught_exceptions’ is not a member of ‘std’

       if (std::uncaught_exceptions() > 0)

           ^

/jenkins/workspace/dealii_PR-11339/source/base/hdf5.cc: In lambda function:

/jenkins/workspace/dealii_PR-11339/source/base/hdf5.cc:335:11: error: ‘uncaught_exceptions’ is not a member of ‘std’

       if (std::uncaught_exceptions() > 0)

           ^

/jenkins/workspace/dealii_PR-11339/source/base/hdf5.cc: In lambda function:

/jenkins/workspace/dealii_PR-11339/source/base/hdf5.cc:449:11: error: ‘uncaught_exceptions’ is not a member of ‘std’

       if (std::uncaught_exceptions() > 0)

           ^

ninja: build stopped: subcommand failed.

script returned exit code 1
```
Are we using the deprecated version, or has the new version of that function only been introduced *after* C++14?

bangerth:
Or are we just not `#include`ing the right header file?

masterleinad:
It's C++17. We are doing
```
#  if __cpp_lib_uncaught_exceptions >= 201411
      // std::uncaught_exception() is deprecated in c++17
      if (std::uncaught_exceptions() == 0)
#  else
      if (std::uncaught_exception() == false)
#  endif
[...]
```
elsewhere already.

bangerth:
Ah, @dangars , my apologies for taking the code you already had correct out :-( Please put it back in then!

dangars:
@bangerth I think now it should be good :)

drwells:
No, we cannot rebuild - @dangars there was a problem with the CI that we needed to fix - can you rebase and force push? That should fix the testers.

masterleinad:
> No, we cannot rebuild - @dangars there was a problem with the CI that we needed to fix - can you rebase and force push? That should fix the testers.

@drwells You can restart the GitHub Action runs yourself when clicking on `Details`.

drwells:
I should probably know the answer to this but when we rerun something does it use the CI configuration present in the PR or the CI configuration present on master? If its the former then we will need to rebase to get around the mac problem.

masterleinad:
> HEAD is now at f31a9dba Merge 37bfb2ca720c87ef33431c2d59d84e9cc709be4f into 691ff4907964605dd21e94f06c880786af2cd612`

Apparently it uses an old commit for merging... So we still need a rebase here.

dangars:
I rebased, now it should be good :)

bangerth:
@dangars -- many thanks for fixing this!

There is one other place in that file (lines 52-65) that I think we should fix similarly. Want to write a patch for that as well?

dangars:
Thanks @bangerth good catch :)

Now it should be ok, I added the missing code in #11440

Issue #11338, opened 2020-12-08
BiCGStab can loop forever


bangerth:
The implementation of BiCGStab looks like this:
```
  do
    {
      if (step != 0)
        deallog << "Restart step " << step << std::endl;
      if (start(A) == SolverControl::success)
        {
          state.state = SolverControl::success;
          break;
        }
      state = iterate(A, preconditioner);
      ++step;
    }
  while (state.breakdown == true);
```
As my student @scottzig has found out the hard way, this can loop forever if the matrix is so that we always get `state.breakdown==true` whenever we iterate. That's because if that happens, we increment the `step` by one and just try again -- but trying again isn't going to yield anything useful if the matrix will just result in another breakdown the next time around.

We really ought to fix this somehow so that one at least gets an error...

PR #11341, opened 2020-12-09
Rework breakdown condition in BiCGSTAB

Rombur:
This PR does two things: 
1. it reworks the breakdown conditions. Right we have an input parameter that is not used. Instead of the current condition, I've the same conditions than the ones used in Trilinos and in matlab.
2. in case of breakdown, the computation just stops. Neither Trilinos or matlab tries to keep going once there is a breakdown. I don't think there is anything we can do in case of breakdown, since it means that we divided by "zero" and thus, the computation doesn't make sense anymore.

closes https://github.com/dealii/dealii/issues/11338

### DIFF
--- a/include/deal.II/lac/solver_bicgstab.h
+++ b/include/deal.II/lac/solver_bicgstab.h
@@ -140,10 +140,13 @@ public:
      * Constructor.
      *
      * The default is to perform an exact residual computation and breakdown
-     * parameter 1e-10.
+     * parameter is the minimum finite value representable by the value_type of
+     * VectorType.
      */
-    explicit AdditionalData(const bool   exact_residual = true,
-                            const double breakdown      = 1.e-10)
+    explicit AdditionalData(
+      const bool   exact_residual = true,
+      const double breakdown =
+        std::numeric_limits<typename VectorType::value_type>::min())
       : exact_residual(exact_residual)
       , breakdown(breakdown)
     {}
@@ -257,13 +260,6 @@ protected:
 
 private:
   /**
-   * Everything before the iteration loop.
-   */
-  template <typename MatrixType>
-  SolverControl::State
-  start(const MatrixType &A);
-
-  /**
    * A structure returned by the iterate() function representing what it found
    * is happening during the iteration.
    */
@@ -349,20 +345,6 @@ SolverBicgstab<VectorType>::criterion(const MatrixType &A,
 
 
 template <typename VectorType>
-template <typename MatrixType>
-SolverControl::State
-SolverBicgstab<VectorType>::start(const MatrixType &A)
-{
-  A.vmult(*Vr, *Vx);
-  Vr->sadd(-1., 1., *Vb);
-  res = Vr->l2_norm();
-
-  return this->iteration_status(step, res, *Vx);
-}
-
-
-
-template <typename VectorType>
 void
 SolverBicgstab<VectorType>::print_vectors(const unsigned int,
                                           const VectorType &,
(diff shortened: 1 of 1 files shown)


Rombur:
I've reverted the new breakdown conditions because some of our tests fail otherwise. Now we are using the old version but we use the tolerance given by the user.

kronbichler (approved):
Looks good to me.

include/deal.II/lac/solver_bicgstab.h
```diff
@@ -418,9 +416,7 @@ SolverBicgstab<VectorType>::iterate(const MatrixType &        A,
 
       alpha = rho / rhobar;
 
-      // TODO:[?] Find better breakdown criterion
-
-      if (std::fabs(alpha) > 1.e10)
+      if (std::fabs(alpha) * additional_data.breakdown > 1)
```
bangerth, line 419: Interesting. I *think* that must have been the intent all along.

bangerth (commented):
I'm no expert in bicgstab, but my understanding is that breakdown is something that can actually happen and can be recovered from. I'm not sure, but I think https://mathworld.wolfram.com/BiconjugateGradientStabilizedMethod.html hints at that.

My understanding that breakdown in bicgstab is when our Krylov sequence has sort of run into a dead end, but that we can then re-start from there. I *think* that that's what the current implementation does, but I found it difficult to read through the class as it's written right now with the externalized `start()` and `iterate` functions -- maybe it would be simpler to inline these into the main function right away. 

Regardless, what I *think* is also currently happening is that we don't properly check for the number of iterations done and terminate after as many iterations as we have done, with each breakdown counting for one iteration.

bangerth:
Take a look here for the breakdown issue: http://www.math.utep.edu/Faculty/xzeng/2017spring_math5330/2017spring_math5330/MATH_5330_Computational_Methods_of_Linear_Algebra_files/ln07.pdf

bangerth:
Maybe we should just implement the algorithm from that PDF...

Rombur:
To be honest with you I have never heard of restarted bicstab

> I'm no expert in bicgstab, but my understanding is that breakdown is something that can actually happen and can be recovered from. I'm not sure, but I think mathworld.wolfram.com/BiconjugateGradientStabilizedMethod.html hints at that.

Where do you see the hint to the restarted method.

> My understanding that breakdown in bicgstab is when our Krylov sequence has sort of run into a dead end, but that we can then re-start from there. I think that that's what the current implementation does, but I found it difficult to read through the class as it's written right now with the externalized start() and iterate functions -- maybe it would be simpler to inline these into the main function right away.

I am not sure how much we restart the method compared to simply "keep going" which you can still do by putting the breakdown tolerance to zero.

> Regardless, what I think is also currently happening is that we don't properly check for the number of iterations done and terminate after as many iterations as we have done, with each breakdown counting for one iteration.

Right we check the number of iterations inside `iterate()` but each restart count as one external iteration which doesn't matter since we never check the number of external iterations.

> Maybe we should just implement the algorithm from that PDF...

The condition in the PDF is the same that the one used by Trilinos and Matlab. Several tests and step 51 don't pass with this condition and no restart. Basically these implementation check that denominator of the fraction is not too small while we check that the ratio is not too large. That matters if the numerator is also small.

Here is what I propose, we either keep the implementation in this PR and if you want a "restart" you put the breakdown tolerance so the algorithm keeps going until you reach the maximum number of iterations or I implement the same breakdown conditions than Trilinos and a proper restarting strategy. What do you think?

bangerth:
Let's do it right. I think that the implementation we have currently is fundamentally broken. Let's implement something like what Trilinos does + restart. Does that seem reasonable?

Rombur:
Sounds good.

Rombur:
This ready for review. Basically, our algorithm is already the same as Belos. The difference is the breakdown conditions. What I didn't realize is how small the breakdown condition is in Trilinos. It's even worse in PETSc where the breakdown condition is that floating point is equal to zero!! I have fixed the restart condition and now the algorithm exits when we reached `max_iter`.

Rombur:
The OSX failure has nothing to do with this PR. The build does not pass configuration because somehow the build system does not recognize that `AppleClang` support C++14

masterleinad:
> The OSX failure has nothing to do with this PR. The build does not pass configuration because somehow the build system does not recognize that `AppleClang` support C++14

That sounds like https://github.com/dealii/dealii/pull/11305 and should be fixed by rebasing.

Rombur:
@bangerth ping

bangerth (approved):

PR #11346, opened 2020-12-09
Check reference-cell type of cell and FE

peterrum:


drwells (approved):

drwells:
The second windows tester is stuck on something else. Since the rest of the CI checks passed I'll merge (adding an assertion is unlikely to trip up MSVC).

PR #11348, opened 2020-12-09
Fix documentation of two SparseDirectMUMPS functions

simonsticko:
Doxygen block comments were used in petsc_solver.cc when c++ block comments should have been used. This messed up the documentation of the functions set_solver_type() and solve(). Fix this.

See:
https://www.dealii.org/current/doxygen/deal.II/classPETScWrappers_1_1SparseDirectMUMPS.html#a08d73453888955bd5a9863be32f3b347

https://www.dealii.org/current/doxygen/deal.II/classPETScWrappers_1_1SparseDirectMUMPS.html#a11450efb60c23f89b1467deb4efd8a61

### DIFF
--- a/source/lac/petsc_solver.cc
+++ b/source/lac/petsc_solver.cc
@@ -643,7 +643,7 @@ namespace PETScWrappers
   void
   SparseDirectMUMPS::set_solver_type(KSP &ksp) const
   {
-    /**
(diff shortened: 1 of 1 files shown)


bangerth (approved):
Good find, thank you!

PR #11355, opened 2020-12-10
Jenkins: use -Og

peterrum:
As discussed in https://github.com/dealii/dealii/pull/11318#issuecomment-742690589.

Also see #10786


tjhei (approved):

tjhei:
The switch from -O0 was necessary to allow test completion in a reasonable time. We had various issues trying to work around compiler warnings with -O1 so I think it is reasonable to go to -Og instead. This is what we did by default in deal.II before 9.2 for a long time anyways.

kronbichler (approved):

kronbichler:
I agree that we cannot go to `-O0` and think this makes sense.

PR #11359, opened 2020-12-11
Fix FlatManifold::normal_vector for triangular faces

peterrum:
... needed in `VectorTools::compute_no_normal_flux_constraints`.

### DIFF
--- a/source/grid/manifold.cc
+++ b/source/grid/manifold.cc
@@ -892,8 +892,19 @@ FlatManifold<dim, spacedim>::normal_vector(
   const unsigned int facedim = dim - 1;
 
   Point<facedim> xi;
-  for (unsigned int i = 0; i < facedim; ++i)
-    xi[i] = 1. / 2;
+
+  const auto face_reference_cell_type = face->reference_cell_type();
+
+  if (face_reference_cell_type == ReferenceCell::get_hypercube(facedim))
+    {
+      for (unsigned int i = 0; i < facedim; ++i)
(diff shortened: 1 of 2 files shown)


source/grid/manifold.cc
```diff
@@ -892,26 +894,45 @@ FlatManifold<dim, spacedim>::normal_vector(
...
   while (true)
     {
       Point<spacedim> F;
-      for (const unsigned int v : GeometryInfo<facedim>::vertex_indices())
-        F += face->vertex(v) *
-             GeometryInfo<facedim>::d_linear_shape_function(xi, v);
+      for (const unsigned int v : face->vertex_indices())
+        F +=
+          face->vertex(v) *
+          (hyper_cube ? GeometryInfo<facedim>::d_linear_shape_function(xi, v) :
```
peterrum, line 923: I am not sure why `GeometryInfo` is used here? Isn't it the same as in `FE_Q`?

kronbichler: Yes, but it is way too expensive to set up an `FE_Q` here just for the purpose of evaluating the shape functions. Note that this is an inner function that is called for many points on each face of the mesh. I think the right approach would be to add a function that can evaluate the linear basis for all reference cell types that we can query similar to `d_linear_shape_function`.

peterrum: @kronbichler What do you think about the updated version?

kronbichler (approved):
Excellent, thank you!

Issue #11206, opened 2020-11-19
Brainstorming - Re-design the data storage structure used in the ParticleHandler


blaisb:
I think it would be interesting to have an issue where we discuss long term plans for redesigning the data structure used to store the particles within the mesh. Right now the multimap has multiples issues, notably related to memory fragmentation and the fact that it makes it expensive to loop over the particles.

This is obviously a long term project, but I would like to gather all of your opinions on this: @kronbichler @peterrum @luca-heltai @gassmoeller .

bangerth:
Some of my recent changes have also made it possible to re-shuffle the memory used for particle properties so that one accesses particle properties in-memory-order when iterating over particles. This would require one additional re-direction, but would probably reduce cache misses.

gassmoeller:
I agree this will be a valuable discussion. To get the discussion started: There are two alternatives that I had considered briefly before settling on the multimap we already had from an earlier implementation (see below). Of course there may be other structures I am not aware of or have not considered so we should not limit this discussion to these options.

(i) An unordered multimap. Smallest implementation change to the current structure. Gets around the logarithmic cost for accessing a particular cell. Iteration over the container is no longer in order of cells of a triangulation, but since access cost is constant that should not be such a problem. insertion/deletion is similar to now. Particle data is still fragmented.

(ii) two separate containers, one for particle storage (a simple array/vector or similar or the storage we currently use for the dynamic properties), one for storing particle-cell associations (e.g. a linked list of cell iterators). This would require periodic sorts (to keep particles in the same cell in similar memory locations) and manual updates if particles are removed or added. To keep things fast we want to avoid frequent memory allocations and sorting of the whole container, i.e. we would need to keep some sort of list of empty storage locations to fill up with arriving particles before reallocating and would need a mechanism to decide when to sort the container.

I would be interested to hear other suggestions/thoughts.

bangerth:
Out of curiosity, do we know for sure that iteration over particles is expensive? In the existing particle tutorials, we access particles as we loop through cells, and the fact that particles are stored in the same order as cells is quite convenient and makes it rather cheap to find the particles associated with the current cell -- just keep iterating starting from the last particle of the previous cell. I think that's actually a good system.

Memory fragmentation: I presume you mean the memory used to store the particles themselves? Do we know that that is an actual issue?

blaisb:
[...]

PR #11360, opened 2020-12-12
Make Particles::Property pool a template <dim,spacedim>.

bangerth:
Follow-up to #11314 and in pursuance of #11206: I'd like to also store particle locations and reference locations in `PropertyPool`, and for that it needs to have template arguments. I'm working on storing the data, but let's get this simple patch out of the way already.


bangerth:
I've got a number of follow-ups already lined up.

peterrum (approved):

Issue #11352, opened 2020-12-10
Simplex/mixed mesh-support follow up tasks


peterrum:
As indicated in some of the PRs, my imminent goal is to have a basic version of simplex/mesh support as soon as possible on master. In this basic version I would like to be able to be provide support matrix-based algorithms (with `FEValues` and `FEFaceValues`) and matrix-free support (`MatrixFree`, `FEEvaluation`, `FEFaceEvaluation`). To be able to reach such a stage in a reasonable time, we might need to postpone the finding of the perfect solution to a later time. Here, we can collect some TODOs, which need to be addressed before the next release:

- implement better shape functions (also higher order!) and better quadrature rules for pyramids
- revisit the selection of evaluation points in `DataOut` : https://github.com/dealii/dealii/blob/6f14f2dc8d33c9237b42065ec19a37cb687e7eab/include/deal.II/numerics/data_out_dof_data.templates.h#L139-L159
is this a good approach that we take the support points of the shape functions?

peterrum:
Done:
- #11424: there are two places where orientation of faces is dealt with. The following function determines the orientation based on given vertices:
https://github.com/dealii/dealii/blob/66efc23e89fbc029e3868ace8779ba37a49e8e28/include/deal.II/grid/connectivity.h#L993-L1086
the following function does the inverse:
https://github.com/dealii/dealii/blob/865c9e620bc5fbac82886f04bdb389211918921b/source/base/qprojector.cc#L646-L694

- #11423: the following function determines the tangentials on faces of reference cells (part of this used to be the task of `GeometryInfo`):
https://github.com/dealii/dealii/blob/865c9e620bc5fbac82886f04bdb389211918921b/source/fe/mapping_fe.cc#L153-L406

peterrum:
I think we made a good progress here. The rest can be done after the release.

kronbichler:
This issue has made no progress in the last two years, I'm removing the target milestone for now as there seems to be low priority.

bangerth:
@drwells and @peterrum What is left here? The text above references a lot of merged PRs. Perhaps we can open separate issues for everything that is not yet addressed so that it is easier to figure out what still needs to be done, and then close this one?

drwells:
[...]

PR #11361, opened 2020-12-12
Enable FEFaceValues for wedges and pyramids

peterrum:
~Only the last commit is relevant. The others are from #10764 and #11317.~

### DIFF
--- a/source/base/qprojector.cc
+++ b/source/base/qprojector.cc
@@ -16,6 +16,7 @@
 #include <deal.II/base/derivative_form.h>
 #include <deal.II/base/geometry_info.h>
 #include <deal.II/base/qprojector.h>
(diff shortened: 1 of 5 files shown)


peterrum:
I have rebased this branch.

drwells (approved):
I think this is fine but I'm concerned about how much geometric information that's pretty basic (like the vertices and areas of reference cell faces) that's encoded into this file. Should we open an issue to track this and possibly move these tables into `reference_cell.h`? Do you think its worth generalizing these?

peterrum:
@drwells I am on it https://github.com/dealii/dealii/issues/11352 point 5 ;)

drwells:
Other more important PRs depend on this so I am willing to merge this as-is.

PR #11363, opened 2020-12-12
Rename some member functions of PropertyPool, also member variables of Particle

bangerth:
The next step towards #11206: If we want to store locations in the `PropertyPool`, it is no longer accurate to name its member functions after "properties". Instead, call them `(de)register_particle()`, and name the corresponding handle in the `Particle` class just "handle".

Strictly speaking, these kinds of changes are incompatible. I suspect that in reality, few people actually use `PropertyPool` in user codes. But I will write a detailed changelog entry when we've settled on the final interface.

### DIFF
--- a/include/deal.II/particles/particle.h
+++ b/include/deal.II/particles/particle.h
@@ -483,7 +483,7 @@ namespace Particles
     /**
      * A handle to all particle properties
      */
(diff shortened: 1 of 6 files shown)


kronbichler (approved):

kronbichler:
Let us wait for another day regarding comments on the new names, then I suggest to merge.

Issue #11362, opened 2020-12-12
CellId::get_cell() should have been a member function of the Triangulation class


bangerth:
It requires access to internal data structures of the triangulations, and I think that just doesn't sound quite right. It is also a place that's easy to forget needs to be updated when adding new triangulation classes.

We should move this function and deprecate it where it currently is.

(@marcfehling FYI)

marcfehling:
On second thought, we could also create a new constructor for the `TriaIterator` class instead which takes `Triangulation` and `CellId` objects. What do you think?

PR #11364, opened 2020-12-13
Added getter function for child_indices of CellId + CellId benchmark testcase

marcfehling:
I noticed that we use `std::array` for the internal structure of CellId. To my knowledge, there is no real performance difference between using `std::array` and `std::vector` if the latter does not change the size after initialization. My local testsuite passes with these changes so I think this is a reasonable change.

This PR also provides a getter function for the `child_indices` member, which will be useful for #11362.

@bangerth -- FYI


include/deal.II/grid/cell_id.h
```diff
@@ -259,11 +252,15 @@ template <class Archive>
 void
 CellId::serialize(Archive &ar, const unsigned int /*version*/)
 {
+  unsigned int n_child_indices = child_indices.size();
```
marcfehling, line 255: I've added this line for compatibility reasons.

marcfehling:
This reverts part of #3289, but leaves the binary serialization (mostly) untouched.

@gassmoeller -- Do you agree to this change? Do you still have your benchmark at hand to verify performance?

tjhei:
What is the motivation to go from a std::array to a vector? It looks to me that this would make the struct no longer a POD type that is easy to copy/serialize/ship. The extra allocation from the vector is not something I would worry too much about, though.

marcfehling:
We currently reserve memory for the `std::array` and only use `n_child_indices` entries of it. In terms of defensive programming, it might easily happen that someone accidentally iterates over the complete `std::array` without stopping after `n_child_indices`.

We are in the process of untangling both the `Triangulation` and `CellId` classes by restricting access to their private members, and for #11362 we need to expose some of `CellId`'s internal data structure with getter functions. I just was not comfortable with returning the full `std::array` and exposing it to the user because of the above reason.

Hmm, now on second thought, a function `CellId::get_child_index_on_level(unsigned int level)` would also do the trick as a getter function.

But you are correct: this change would make the size of objects variable. But would it make to much of a difference since we are now capable of easily transferring and serializing data of variable size among multiple processes? The binary representation of this class still works with fixed-size arrays. One could think of making those variable-sized as well, but I did not want to think about it in this PR.

marcfehling:
Long story short: I would be happy with either keeping `std::array` and using a `CellId::get_child_index_on_level(unsigned int level)` getter function, or switching to `std::vector`.

@tjhei -- If you want to place a veto, I'll rework this PR into the former option :)

include/deal.II/grid/cell_id.h
peterrum: No need for this. We do not guarantee compatibility regarding serialization.

peterrum (approved):

kronbichler:
> To my knowledge, there is no real performance difference between using `std::array` and `std::vector` if the latter does not change the size after initialization.

The main question to me would be how often we construct `CellId` because memory allocation (and de-allocation) is the expensive part of `std::vector` of short sizes. So it would be good to audit or even benchmark some typical uses of this field. I could not do a comprehensive search, but a short grepping for `CellId` and `cell->id()` revealed some dozen of uses. None of those in really critical paths but nonetheless I would only do it if necessary.

If the question is the exposure of invalid entries, then to me the right thing seems to be to return `ArrayView` of the right size or, as you were suggest, the getter function for the level..

bangerth:
I had some additional comments in #11365.

marcfehling:
Thank you for your feedback! I like the idea of returning an `ArrayView` object. But I got curious about the performance of `std::vector` in this regard. I will prepare a small benchmark.

gassmoeller:
In case you want to reuse the benchmark I used in #3289 this is the setup: 
[cell_id_04.cc.txt](https://github.com/dealii/dealii/files/5691464/cell_id_04.cc.txt)

I did not include timers, but instead used callgrind to count cycles. Adding a timer may be useful as well.
The main reason this was performance critical for me was that when particles are exchanged between MPI ranks, each transferred particle is paired with its surrounding cell, which has to be transmitted as CellID.

agrayver:
I use CellId in many contexts where I need to identify cells and attach some unique information to them (e.g. properties, some flags, etc). The reason is because it is the unique indicator of a cell, thus this information can be transferred (imagine you construct a complex model which you need to pass to users and they run it in different settings).

Therefore, constructing arrays with lots of CellId objects is very common in my codes -- and this is fast with std::array.

bangerth:
@agrayver Do you think you could come up with a micro-benchmark that illustrates the usual use case you have in mind? That might help us decide in figuring out what to do.

@marcfehling A compromise would be using `boost::small_vector` with, say, 8 or 10 (or maybe `11-dim`) slots in static storage. That would probably be good enough for almost all use cases.

kronbichler:
@bangerth But if we expose an `ArrayView` to the outside then we can use whatever efficient internal representation we want, can't we? So we could stick with the current array (which, with 32 entries, allows for 2^30 in factor of mesh size refinement compared to the coarse mesh, so I can't think of a case where it would not suffice) without exposing invalid data. By the way, the array should even consume less memory on 64 bit systems than an `std::vector` (24 byte for the pointer + at least 8 byte, usually 16 byte IIRC, for the allocated part, also when the used data is only 4-5 byte).

marcfehling:
I've provided a new testcase `cell_id_08` which saves/loads all cells of a basic `Triangulation` object in every possible way. Buffer sizes will be written into the `deallog` stream, while timing results will be written into `std::cout`. The test consists of the following four subsections and should represent a typical use-case for the serialization of CellId objects.

- create and store a `CellId` object via `cell->id()` in a `std::vector`
- serialize the container into a `std::string` and back using `boost::binary_archives`
- serialize each element of the container successively into a `std::vector<char>` and back using `Utilities::pack` and `Utilities::unpack`
- serialize each element of the container successively into a `std::vector<CellId::binary_type>` and back using `CellId::to_binary` and `CellId(binary_type)`

I've used this testcase as a benchmark with a lot more cells: I've run this on a 3D cube with 8 global refinements yielding about 16 mio cells. For the timing of each subsection, a sufficient amount of memory has been reserved.

Please find the results of this scenario with and without this patch attached. The presented results incorporate patch #11368, otherwise the `Utilities::pack` would have taken >1hr.

category | without patch | with patch | factor
-- | -- | -- | --
**construction** |   |   |  
cell->id() time [s] | 3.97 | 4.35 | 1.10
**boost binary archive** |   |   |  
buffer size [MB] | 772 | 403 | 0.52
save time [s] | 2.85 | 2.46 | 0.86
load time [s] | 2.74 | 3.32 | 1.21
**Utilities pack/unpack** |   |   |  
buffer size [MB] | 671 | 1158 | 1.72
save time [s] | 0.298 | 35.8 | 120
load time [s] | 0.408 | 35.5 | 87.0
**CellId to/from binary** |   |   |  
buffer size [MB] | 268 | 268 | 1.00
save time [s] | 0.241 | 0.258 | 1.07
load time [s] | 0.297 | 0.838 | 2.82


As @tjhei pointed out, using `std::vector` instead of `std::array` makes the `CellId` type no longer trivially copyable, thus in the `Utilities` option we no longer infer `memcpy` but use the more costly boost serialization method which takes lots of time being set up for each cell individually. I have no idea why the buffer sizes differ in that case though. I could add another test to check how `Utilities::pack/unpack` performs when serializing the whole container object, and not each individual element one after another.

The boost binary archive option uses more memory without the patch probably most probably due to the amount of pre-allocated memory when using `std::array`.

I am convinced now that using a plain `std::vector` is the wrong choice seeing that loading `CellId` using their `binary_type` equivalent takes longer which is probably due to the allocation overhead for this container type as @kronbichler pointed out. A proper patch would not increase the construction time in this regard.

So if you want to I can adjust this patch to use `boost::small_vector` as @bangerth pointed out and see how it affects the timing results, or we can leave the internal structure as is, and I will only add a getter function for the `child_indices` member returning an `ArrayView` object.

kronbichler:
I would vote for keeping the array internally and return `ArrayView` because it seems to fulfill all our needs. Exposing the storage class via a `small_vector` seems unnecessary to me and I believe it will take more memory to store say 20 bytes than the array, unless boost uses similar kinds of tricks as the short string optimizations in the libc++ library (unlikely). In any case, I do not feel strongly about it.

bangerth:
I agree withe `ArrayView` idea. In fact @marcfehling now seems to as well, following our weekly call just now :-) He's going to come up with a patch -- I'll mark this one as WIP for now.

marcfehling:
I've changed this PR to only add the getter function and the benchmark testcase.

I still need to come up with a testcase for the getter function.

marcfehling:
Added the missing testcase. Ready for final review!

bangerth (commented):
The patch is ok, but I'm curious about how long your "benchmark" runs. It's really just 256 cells, so likely not very long, but I also don't want to merge a patch that introduces a long-running test :-)

marcfehling:
> The patch is ok, but I'm curious about how long your "benchmark" runs. It's really just 256 cells, so likely not very long, but I also don't want to merge a patch that introduces a long-running test :-)

I've reduced the amount of cells so that it can be part of the testsuite. We now have 4 global refinements in 2D leading to 256 cells, instead of 8 global refinements in 3D with 16mio cells (see the commented out line). The test takes a fraction of a second in debug mode with 256 cells on your machines.

kronbichler (approved):
Very nice, thank you! I appreciate the careful benchmark of variants.

PR #11365, opened 2020-12-13
Added Triangulation::create_cell_iterator().

### LINKED ISSUE #11362: CellId::get_cell() should have been a member function of the Triangulation class

marcfehling:
Closes #11362. ~~Blocked by #11364.~~ (EDIT: merged)

### DIFF
--- a/include/deal.II/grid/cell_id.h
+++ b/include/deal.II/grid/cell_id.h
@@ -148,9 +148,11 @@ public:
 
   /**
    * Return a cell_iterator to the cell represented by this CellId.
+   *
+   * @deprecated Use Triangulation::create_cell_iterator() instead.
    */
   template <int dim, int spacedim>
-  typename Triangulation<dim, spacedim>::cell_iterator
+  DEAL_II_DEPRECATED typename Triangulation<dim, spacedim>::cell_iterator
   to_cell(const Triangulation<dim, spacedim> &tria) const;
 
   /**
--- a/include/deal.II/grid/tria.h
+++ b/include/deal.II/grid/tria.h
@@ -25,6 +25,7 @@
 #include <deal.II/base/smartpointer.h>
 #include <deal.II/base/subscriptor.h>
 
+#include <deal.II/grid/cell_id.h>
 #include <deal.II/grid/tria_description.h>
 #include <deal.II/grid/tria_iterator_selector.h>
 #include <deal.II/grid/tria_levels.h>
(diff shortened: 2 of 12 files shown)


peterrum (approved):

include/deal.II/grid/cell_id.h
```diff
@@ -259,11 +251,15 @@ template <class Archive>
 void
 CellId::serialize(Archive &ar, const unsigned int /*version*/)
 {
+  unsigned int n_child_indices = child_indices.size();
```
bangerth, line 254: You don't actually have to serialize this any more since you already serialize the whole `child_indices` array.

include/deal.II/grid/tria.h
```diff
@@ -2743,6 +2744,13 @@ class Triangulation : public Subscriptor
   active_cell_iterator
   last_active() const;
 
+  /**
+   * Return an iterator to a cell of this Triangulation object constructed from
+   * an independent CellId object.
```
bangerth, line 2749: Maybe add a sentence like this:
```suggestion
   * an independent CellId object. If the given argument corresponds to a valid
   * cell in this triangulation, this operation will always succeed for sequential
   * triangulations where the current processor stores all cells that are part
   * of the triangulation. On the other hand, if this is a parallel triangulation,
   * then the current processor may not actually know about this cell and in this
   * case ...
```
+ add what actually happens. I suspect that the operation is ok if it's a locally owned or ghost cell, but may not work if it's an artificial cell. Maybe say so.

include/deal.II/grid/tria.h
```diff
@@ -2743,6 +2744,13 @@ class Triangulation : public Subscriptor
...
+   */
+  cell_iterator
+  create_cell_iterator(const CellId &) const;
```
bangerth, line 2752:
```suggestion
  create_cell_iterator(const CellId &cell_id) const;
```

source/grid/tria.cc
```diff
@@ -11985,6 +11985,21 @@ Triangulation<dim, spacedim>::last_active() const
...
 
+template <int dim, int spacedim>
+typename Triangulation<dim, spacedim>::cell_iterator
+Triangulation<dim, spacedim>::create_cell_iterator(const CellId &cell_id) const
+{
+  cell_iterator cell(
+    this, 0, coarse_cell_id_to_coarse_cell_index(cell_id.get_coarse_cell_id()));
+
+  for (const auto &child_index : cell_id.get_child_indices())
+    cell = cell->child(static_cast<unsigned int>(child_index));
```
bangerth, line 11996: I think this would be the place where it's worth adding an assertion that `cell->has_children() == true` and if not provide an error message along the lines of what I suggested to put into the documentation.

tjhei (approved):
Thank you for working on cleaning up my badly designed code. :-)

kronbichler (approved):
Looks good to me once we have agreed on a solution for #11364.

include/deal.II/grid/tria.h
marcfehling: I suspect so, too. Artificial cells may not be refined enough to create a cell iterator. I've adjusted the documentation and the error message accordingly.

marcfehling:
Rebased on master. Ready for final review/merge.

bangerth (approved):

PR #11368, opened 2020-12-13
Remove temporary string objects in Utilities::(un)pack.

marcfehling:
While preparing a benchmark for #11364, I've noticed that we use intermediate temporary `string` buffers for (de)serialization before moving the buffer to the correct position.

This appears unnecessary, and I prepared this PR to directly write to/read from the corresponding objects. I'll present you some benchmark results soon.

### DIFF
--- a/include/deal.II/base/utilities.h
+++ b/include/deal.II/base/utilities.h
@@ -45,15 +45,15 @@ DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
 #include <boost/archive/binary_iarchive.hpp>
 #include <boost/archive/binary_oarchive.hpp>
 #include <boost/core/demangle.hpp>
+#include <boost/iostreams/device/array.hpp>
+#include <boost/iostreams/device/back_inserter.hpp>
+#include <boost/iostreams/filtering_streambuf.hpp>
 #include <boost/serialization/array.hpp>
 #include <boost/serialization/complex.hpp>
 #include <boost/serialization/vector.hpp>
 
 #ifdef DEAL_II_WITH_ZLIB
-#  include <boost/iostreams/device/back_inserter.hpp>
 #  include <boost/iostreams/filter/gzip.hpp>
-#  include <boost/iostreams/filtering_stream.hpp>
-#  include <boost/iostreams/stream.hpp>
 #endif
 
 DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
@@ -1183,9 +1183,6 @@ namespace Utilities
        std::vector<char> &dest_buffer,
        const bool         allow_compression)
   {
-    // the data is never compressed when we can't use zlib.
-    (void)allow_compression;
-
     std::size_t size = 0;
 
     // see if the object is small and copyable via memcpy. if so, use
@@ -1217,31 +1214,21 @@ namespace Utilities
         // use buffer as the target of a compressing
         // stream into which we serialize the current object
         const std::size_t previous_size = dest_buffer.size();
+        {
+          boost::iostreams::filtering_ostreambuf fosb;
 #ifdef DEAL_II_WITH_ZLIB
-        if (allow_compression)
-          {
-            boost::iostreams::filtering_ostream out;
-            out.push(
-              boost::iostreams::gzip_compressor(boost::iostreams::gzip_params(
-                boost::iostreams::gzip::default_compression)));
-            out.push(boost::iostreams::back_inserter(dest_buffer));
-
-            boost::archive::binary_oarchive archive(out);
-            archive << object;
-            out.flush();
-          }
-        else
+          if (allow_compression)
+            fosb.push(boost::iostreams::gzip_compressor());
+#else
+          (void)allow_compression;
 #endif
-          {
-            std::ostringstream              out;
-            boost::archive::binary_oarchive archive(out);
-            archive << object;
+          fosb.push(boost::iostreams::back_inserter(dest_buffer));
 
-            const std::string s = out.str();
-            dest_buffer.reserve(dest_buffer.size() + s.size());
-            std::move(s.begin(), s.end(), std::back_inserter(dest_buffer));
-          }
-
+          boost::archive::binary_oarchive boa(fosb);
+          boa << object;
+          // the stream object has to be destroyed before the return statement
+          // to ensure that all data has been written in the buffer
+        }
         size = dest_buffer.size() - previous_size;
       }
 
@@ -1267,9 +1254,6 @@ namespace Utilities
   {
     T object;
 
-    // the data is never compressed when we can't use zlib.
-    (void)allow_compression;
-
     // see if the object is small and copyable via memcpy. if so, use
     // this fast path. otherwise, we have to go through the BOOST
     // serialization machinery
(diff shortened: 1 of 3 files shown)


tjhei (approved):
This looks good and is a nice simplification. We now unconditionally use boost instead stringstreams, which I assume work as fast but it wouldn't hurt checking.
1. Did you run all serialization tests? Some details might have changed.
2. Would you mind running a quick test to see that this is not a regression? Running one of the tests under valgrind (callgrind) before/after is probably enough.

include/deal.II/base/utilities.h
```diff
@@ -1357,29 +1331,17 @@ namespace Utilities
...
-          {
-            decompressed_buffer.assign(cbegin, cend);
-          }
+        in.push(boost::iostreams::array_source(&*cbegin, &*cend));
 
-        // then restore the object from the buffer
-        std::istringstream              in(decompressed_buffer);
         boost::archive::binary_iarchive archive(in);
-
         archive >> unpacked_object;
```
marcfehling, line 1345: All tests accomplish, but not `base/utilities_pack_unpack_(05|06)`: they fail in this particular line complaining about the input stream:
```
terminate called after throwing an instance of 'boost::archive::archive_exception'
  what():  input stream error
```

My guess is that `boost` expects metainformation about the array in the deserialization process, which we probably do not write out in the serialization process. The metainformation may vanish while writing the contents in the legacy `string`. Calling `boost::serialization::make_array` somewhere might do the trick, and I'm still trying to figure out where.

Please be patient with this PR :)

marcfehling: I've found the issue: it's a synchronization problem. The `zlib_compressor` is not flushable, thus we exit the `pack` function before finishing to write into the buffer and we always return that `0` bytes have been written. In that specific test, we rely on the returned number of bytes to recover the serialized object from the buffer, and there we have the issue.

This seems to be an issue that has been around for decades (see [here](https://groups.google.com/g/boost-list/c/0u584AOTIqM)), and was recently proposed as a change to boost https://github.com/boostorg/iostreams/pull/116.

Destroying the stream object before the return statement is supposed to do the trick. I'll give that one a shot. That took a lot of time to figure out...

marcfehling: Two hours of debugging for a pair of braces...

include/deal.II/base/utilities.h
```diff
@@ -1217,31 +1214,19 @@ namespace Utilities
...
-                boost::iostreams::gzip::default_compression)));
-            out.push(boost::iostreams::back_inserter(dest_buffer));
-
-            boost::archive::binary_oarchive archive(out);
-            archive << object;
-            out.flush();
-          }
-        else
+          if (allow_compression)
+            fosb.push(boost::iostreams::zlib_compressor());
```
bangerth, line 1221: Any reason to switch from the `gzip_compressor` used before to the `zlib_compressor` here?

source/base/utilities.cc
```diff
@@ -398,8 +398,7 @@ namespace Utilities
     std::stringstream origin(input);
 
     bio::filtering_streambuf<bio::input> out;
-    out.push(bio::gzip_compressor(
-      bio::gzip_params(boost::iostreams::gzip::default_compression)));
+    out.push(bio::zlib_compressor());
```
bangerth, line 401: same question here

bangerth (commented):
Good detective work :-)

include/deal.II/base/utilities.h
marcfehling: To my knowledge, `gzip` uses the `zlib` algorithms internally and additionally provides tools for file compression. Since we are dealing with data only here, it felt more appropriate to use the `zlib` wrappers here. So in our context, that's analogous to wrapping a package in either green or red paper.

Since we specifically look for `zlib` in the cmake configuration phase and provide a `DEAL_II_WITH_ZLIB` flag, it appears even more appropriate to use the `zlib` wrappers.

marcfehling: Interesting, the sizes of the compressed objects in test `base/utilities_pack_unpack_07` vary depending on the routines used. Objects are negligibly smaller with `zlib` compression. We could discuss about using `zlib` in a separate PR if you like.

marcfehling:
I removed changing the (de)compressor type in this PR.

---

I don't have much experience with callgrind, but I think the results are in favor of this patch. I used `kcachegrind` for visualization. See results for `tests/base/utilities_pack_unpack_07.debug` attached.

### Without this patch
callgrind_before
### With this patch
callgrind_after

marcfehling:
CI complains about a test
```
3815 - multigrid/transfer_matrix_free_06.mpirun=15.debug (Timeout)
```
This test works fine on my local testsuite. I'll force push to restart the CI.

include/deal.II/base/utilities.h
bangerth: "that's analogous to wrapping a package in either green or red paper" -- what a timely Christmas-themed analogy :-)

bangerth: I don't really care much either way. @kronbichler did some tests a while ago that showed that compression can be surprisingly expensive in terms of CPU time.

include/deal.II/base/utilities.h
```diff
@@ -1217,31 +1214,21 @@ namespace Utilities
...
+#else
+          (void)allow_compression;
 #endif
-          {
-            std::ostringstream              out;
-            boost::archive::binary_oarchive archive(out);
-            archive << object;
-
-            const std::string s = out.str();
-            dest_buffer.reserve(dest_buffer.size() + s.size());
```
marcfehling, line 1241: Working on the benchmark test for #11364 also revealed a bug in the current implementation.

I wanted to pack several `CellId` objects into one container and reserved sufficient memory for `dest_buffer` container to do so. However, this line overwrites my carefully laid out memory management and reserves new memory for each of the 16 million `CellId` objects my test wanted to pack. This is not intentional :)

With only 250k `CellId` objects, the `pack` serialization took about `15s` in release mode on my test system, while all other serialization methods took about `0.3s`. I'll publish the results of the benchmark in the discussion of the above mentioned PR soon.

bangerth: Just for my own education, what is happening is that you've nicely reserved a very large buffer, and then this call comes along and reserves a smaller size thereby releasing memory that you need to allocate again later?

bangerth (approved):

bangerth:
Good work, thank you!

include/deal.II/base/utilities.h
marcfehling: Sorry that I did not make this more clear. Yes, that's exactly what's happening. I think for this case I reserved more than 1GB of memory for the buffer before starting the serialization operation. When saving the first `CellId` with `Utilities::pack`, the buffer reservation will be overwritten to hold less than 100 bytes, thus releasing all the carefully reserved memory. Since I wanted to save about 16mio `CellId` objects, we continue to expand the size of the buffer for every single one of these objects, which ultimately infers moving the continuous memory chunk assigned to the container multiple times.

bangerth: I wonder whether it would make sense to use `std::stack` for this instead of `std::vector`. In general, we won't know how large the final result is going to be, and it would be useful to store into an object that doesn't have to copy everything when it finds that it needs more memory. `std::stack` is such a container.

PR #11371, opened 2020-12-14
FEInterfaceValues: pass to the constructor hp::QCollection

peterrum:
Similar to #10764 but for `FEInterfaceValues`. Motivation is to use the class for wedges and pyramids.

~Depends on #10764.~

### DIFF
--- a/include/deal.II/fe/fe_interface_values.h
+++ b/include/deal.II/fe/fe_interface_values.h
@@ -21,6 +21,8 @@
 #include <deal.II/fe/fe_values.h>
 #include <deal.II/fe/mapping_q1.h>
 
+#include <deal.II/hp/q_collection.h>
+
 DEAL_II_NAMESPACE_OPEN
 
 /**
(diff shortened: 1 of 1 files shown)


include/deal.II/fe/fe_interface_values.h
```diff
@@ -551,6 +581,12 @@ FEInterfaceValues<dim, spacedim>::reinit(
       fe_face_values_neighbor = &internal_fe_subface_values_neighbor;
     }
 
+  AssertDimension(fe_face_values->n_quadrature_points,
+                  fe_face_values_neighbor->n_quadrature_points);
+
+  const_cast<unsigned int &>(this->n_quadrature_points) =
```
drwells, line 587: Could we make `FEInterfaceValues::n_quadrature_points` non-const?

peterrum: Different reviewers different opinions: https://github.com/dealii/dealii/pull/10764#discussion_r541937876

drwells: Lets keep it this way for now but we should eventually clean this up.

drwells (approved):

PR #11374, opened 2020-12-14
Add a particle test.

bangerth:
I felt like I wanted another test to check for the various particle improvements (see everything linked to from #11206). This patch adds a slightly modified version of step-19 as a test -- specifically modified to run within just a few seconds.

/rebuild

### DIFF
--- a/tests/particles/step-19.cc
+++ b/tests/particles/step-19.cc
@@ -1,0 +1,1117 @@
+/* ---------------------------------------------------------------------
+ *
+ * Copyright (C) 2020 by the deal.II authors
+ *
+ * This file is part of the deal.II library.
+ *
+ * The deal.II library is free software; you can use it, redistribute
+ * it, and/or modify it under the terms of the GNU Lesser General
+ * Public License as published by the Free Software Foundation; either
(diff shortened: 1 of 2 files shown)


bangerth:
I believe that the CI (serial build) check failure is spurious.

masterleinad:
```
/tmp/cccynDNr.s: Fatal error: can't close CMakeFiles/chart_manifold_04_embedded.debug.dir/chart_manifold_04_embedded.cc.o: No space left on device
```

masterleinad:
Let's retry.

/rebuild

kronbichler (approved):
This looks good to me in principle, so I already approve. However, I would suggest to remove comments unless you see a good reason for keeping them: Sooner or later they will diverge from the tutorial program so I would see them as merely disturbing here.

bangerth:
I think that if there is divergence between the tutorial program and the test, that it would be that much more useful to have the comments left in. These comments comment the state of the test, not whatever is in the tutorial at any given time.

I recognize that the comments are written at a different level than we usually do for tests, but I think that they still provide a small benefit when left in.

PR #11378, opened 2020-12-15
Link in the correct figure.

bangerth:
Follow-up to #11201 where the text no longer matched the figure. I'm fixing this by just linking in the corrected figure. See https://dealii.org/developer/doxygen/deal.II/namespaceGridGenerator.html#ac7515d2b17c025dddc0e37286fb8d216 for the current state.

@eliasstudiert -- FYI.

/rebuild


peterrum (commented):
That was the correct figure  but the text was outdated.

eliasstudiert:
> Follow-up to #11201 where the text no longer matched the figure. I'm fixing this by just linking in the corrected figure. See https://dealii.org/developer/doxygen/deal.II/namespaceGridGenerator.html#ac7515d2b17c025dddc0e37286fb8d216 for the current state.
> 
> @eliasstudiert -- FYI.
> 
> /rebuild

It's true text and figure don't belong together - my bad, I should have adapted the description. I suggest to either use both pictures or only the new one. If we use both, I could provide a nicer version of the image showing the quater_hypershell without the vertical separation line and less space in between.

bangerth:
I think the figure is ok as is (or if you want, we can replace it at a later point). Is there an actual objection to the PR?

bangerth:
Ping?

masterleinad:
@eliasstudiert Is this fine with you?

eliasstudiert (approved):

peterrum (approved):
We will update the picture in a follow-up PR.

Issue #11375, opened 2020-12-14
Compress test output files?


bangerth:
My `tests/` directory currently has ~275 MB. If I bzip2 all of the output files, it only has 90 MB. This would sound like a win, but one has to balance that against the time it takes to decompress every time a test is run and the output is compared. The latter only takes ~3 seconds on my laptop, which is certainly negligible compared to the overall run time of the test suite.

What do people think about compressing the data?

tjhei:
> What do people think about compressing the data?

What would be the advantage except saving <200MB on your disk? There are a lot of complications to deal with when handling compressed data (including PRs, diffing, storage inside the git repo, etc.).

tamiko:
A massive disadvantage of storing files in `tests/` compressed is that we would check them into git as binary blobs that cannot be compressed efficiently and, worse, where a one line change in an output file will lead to a completely different compressed binary blob. This does not play well with the delta compression git is using. For reference, the `.git` directory for our archive on my computer is only 231MB large - storing all of our archive history from the beginning of time.

tamiko:
Actually two more comments:

1/ (unhelpful remark:) That's one of the reasons why I am a huge fan of filesystems with transparent compression (like btrfs or zfs). I use it all the time on my computers/servers.

2/ Actually, there might be a way to achieve this by declaring a checkout/checkin filter for git. This would alleviate the problems I pointed out. It is still up to debate whether the added complexity and increased test time (necessity to compress/decompress) is worth it. But just for educational purposes I will investigate how to hack together such a filter.

tjhei:
> It is still up to debate whether the added complexity and increased test time (necessity to compress/decompress) is worth it.

I would vote no. We are talking saving less than 200MB here. 

I can think of many other useful problems to spend engineer time on. :-)

bangerth:
Well, that was a resounding answer :-)

PR #11383, opened 2020-12-16
Add a filter=output attribute to output files

tamiko:
This change alone does nothing but helps to set up custom smudge/clean
filters for some files in the repository.

For example, one can set up the following filters in `.git/config` (1, 2):

```
[filter "test-output"]
        clean = zstd -d
        smudge = zstd -z
        required

[filter "bundled-files"]
        clean = zstd -d
        smudge = zstd -z
        required
```

git will then compress all files that had been tagged with "test-output"
and "bundled-files" via zstd saving some discspace.

A couple of comments:

1/ For very good reason it is not possible to automatically add this
   configuration to the repository - git tries very hard to prevent remote
   code execution exploits when handling repositories

2/ I don't see any benefit in changing everyones workflow to handle
   compressed (test output) files.

   But adding filter annotations alone doesn't do anything. So why not
   adding the attributes and let everyone be happy? :-)

3/ This obviously requires an additional hook to decompress test output
   files if we want the testsuite to still function. A possible route is to add
   an override with an environment variable to "preprocess" output
   files prior to comparing with numdiff.

(1) It is recommended to make sure that clean and smudge are projections.
    So it might be best to write a small wrapper ensuring this...

(2) These two filters reduce the size of the working directory by 228M
    (369M bundled + tests down to 140M)

In reference to #11375

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,17 +1,37 @@
-/.github/ export-ignore
-/cmake/scripts/run_test.sh text eol=lf
+#
+# do not include the following files in tarballs created via git-export:
+#
+
+.clang-format export-ignore
+.clang-tidy export-ignore
+.codecov.yml export-ignore
 /contrib/ci/ export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+Jenkinsfile export-ignore
+.mailmap export-ignore
+.travis.yml export-ignore
+
+/.github/ export-ignore
+
 /tests/* export-ignore
 /tests/CMakeLists.txt -export-ignore
 /tests/quick_tests -export-ignore
 /tests/run_test.cmake -export-ignore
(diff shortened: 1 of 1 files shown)


bangerth (approved):
This is ok with me, but it would probably be good to have someone with more git experience to look at this as well.

bangerth:
You might want to rebase to see if the OS X failures go away.

masterleinad (approved):
Looks reasonable to me.

tjhei (approved):
I guess this doesn't hurt...

bangerth:
@tamiko please rebase and see whether it then also succeeds on the OS X machines.

tamiko:
@bangerth Wow. I completely forgot about this pull request. Rebased.

PR #11384, opened 2020-12-16
CMake-3.19 fixes

tamiko:
Closes: #11272

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -30,6 +30,20 @@ jobs:
       run: |
         make -j 2
         make -j 2 test # quicktests
+    - name: archive error 1
+      uses: actions/upload-artifact@v1
+      if: always()
+      continue-on-error: true
+      with:
+        name: osx-serial-CMakeOutput.log
+        path: CMakeFiles/CMakeOutput.log
+    - name: archive error 2
+      uses: actions/upload-artifact@v1
+      if: always()
+      continue-on-error: true
+      with:
+        name: osx-serial-CMakeError.log
+        path: CMakeFiles/CMakeError.log
 
   osx-parallel64:
     # MPI build using apple clang and 64 bit indices
@@ -64,4 +78,18 @@ jobs:
       run: |
         make -j 2
         make -j 2 test #quicktests
+    - name: archive error 1
+      uses: actions/upload-artifact@v1
+      if: always()
+      continue-on-error: true
+      with:
+        name: osx-parallel64-CMakeOutput.log
+        path: CMakeFiles/CMakeOutput.log
+    - name: archive error 2
+      uses: actions/upload-artifact@v1
+      if: always()
+      continue-on-error: true
+      with:
+        name: osx-parallel64-CMakeError.log
+        path: CMakeFiles/CMakeError.log
 
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -14,9 +14,6 @@ jobs:
     - name: setup
       run: |
         ./contrib/utilities/download_clang_format
-        # force cmake version:
-        pip install cmake==3.18.4
-        echo "/Library/Frameworks/Python.framework/Versions/2.7/bin" >> $GITHUB_PATH
     - name: info
       run: |
         g++ -v
@@ -44,9 +41,6 @@ jobs:
     - uses: actions/checkout@v2
     - name: setup
       run: |
-        # force cmake version:
-        pip install cmake==3.18.4
-        echo "/Library/Frameworks/Python.framework/Versions/2.7/bin" >> $GITHUB_PATH
         brew install openmpi
         cmake --version
         # uncomment these for a gcc based build
--- a/cmake/macros/macro_enable_if_links.cmake
+++ b/cmake/macros/macro_enable_if_links.cmake
@@ -23,16 +23,17 @@
 #
 
 MACRO(ENABLE_IF_LINKS _variable _flag)
-  STRING(STRIP "${_flag}" _flag_stripped)
+  # keep on top to avoid cluttering the _flag and _flag_stripped variables
+  ENABLE_IF_SUPPORTED(CMAKE_REQUIRED_FLAGS "-Werror")
 
+  STRING(STRIP "${_flag}" _flag_stripped)
   IF(NOT "${_flag_stripped}" STREQUAL "")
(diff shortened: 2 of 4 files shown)


tamiko:
@tjhei Before doing anything else let's try whether it is simply sufficient to remove this old workaround for something clang related that I do not remember any more. My guess is that this is long fixed in clang.

tjhei:
It would be great if you can try fixing our cmake 3.19 support. For now, we can work around it by using 3.18 for CI.

tamiko:
@tjhei This pull request fixes the issue I described in #11272. I will investigate what is going wrong on OSX but for that we have to extract the contents of `CMakeError.log` first.

masterleinad:
Just push what we have for `MSVC` in the GitHub Action setup.
```
- name: archive error 1
      uses: actions/upload-artifact@v1
      if: always()
      continue-on-error: true
      with:
        name: windows-serial-CMakeOutput.log
        path: build/CMakeFiles/CMakeOutput.log
    - name: archive error 2
      uses: actions/upload-artifact@v1
      if: always()
      continue-on-error: true
      with:
        name: windows-serial-CMakeError.log
        path: build/CMakeFiles/CMakeError.log
```

tamiko:
@tjhei cmake-3.19 seems to have been pulled from pypa. I currently cannot reproduce on OSX.

tamiko:
@tjhei Alright. This should do the trick.

.github/workflows/osx.yml
```diff
@@ -33,6 +30,20 @@ jobs:
...
+      with:
+        name: osx-serial-CMakeOutput.log
+        path: CMakeFiles/CMakeOutput.log
+    - name: archive error 2
+      uses: actions/upload-artifact@v1
+      if: always()
+      continue-on-error: true
+      with:
+        name: osx-serial-CMakeError.log
+        path: CMakeFiles/CMakeError.log
```
tamiko, line 46: @masterleinad I suggest we just keep this in?

.github/workflows/osx.yml
```diff
@@ -45,8 +56,6 @@ jobs:
     - name: setup
       run: |
         # force cmake version:
```
tjhei, line 58: remove this line

.github/workflows/osx.yml
tjhei: Agreed.

cmake/macros/macro_enable_if_links.cmake
```diff
@@ -23,21 +23,19 @@
...
-    SET(_backup ${CMAKE_REQUIRED_LIBRARIES})
+    STRING(REGEX REPLACE "\[-+,\]" "_" _flag_name "${_flag_name}")
+
     LIST(APPEND CMAKE_REQUIRED_LIBRARIES "${_flag_stripped}")
-    CHECK_CXX_COMPILER_FLAG(
-      ""
-      DEAL_II_HAVE_FLAG_${_flag_name}
-      )
-    SET(CMAKE_REQUIRED_LIBRARIES ${_backup})
+    CHECK_CXX_SOURCE_COMPILES("int main(){}" DEAL_II_HAVE_LINKER_FLAG_${_flag_name})
```
tamiko, line 35: We can no longer use the `CHECK_CXX_COMPILER_FLAG` macro because cmake no longer invokes the linker.

tjhei (commented):
Looks good to me.

cmake/macros/macro_enable_if_supported.cmake
```diff
@@ -43,21 +31,19 @@ MACRO(ENABLE_IF_SUPPORTED _variable _flag)
...
-    STRING(REPLACE "," "" _flag_name "${_flag_name}")
-    STRING(REPLACE "-" "_" _flag_name "${_flag_name}")
-    STRING(REPLACE "+" "_" _flag_name "${_flag_name}")
-    CHECK_CXX_COMPILER_FLAG(
-      "${_werror_string}${_flag_stripped}"
-      DEAL_II_HAVE_FLAG_${_flag_name}
-      )
+    STRING(REGEX REPLACE "\[-+,\]" "_" _flag_name "${_flag_name}")
+
+    CHECK_CXX_COMPILER_FLAG("${_flag_sanitized}" DEAL_II_HAVE_FLAG_${_flag_name})
```
tamiko, line 43: A related change in the `CHECK_CXX_COMPILER_FLAG` macro now requires a single compiler flag as argument. I.e., our old `-Werror <flag>` check no longer works because the flag is now handed down to the compiler as a single quoted argument.

cmake/macros/macro_find_system_library.cmake
```diff
@@ -26,10 +26,10 @@ MACRO(FIND_SYSTEM_LIBRARY)
   LIST(GET _argn 0 _variable)
   LIST(REMOVE_AT _argn 0 1)
 
-  if("${_variable}" MATCHES "^${_variable}$")
+  IF(NOT DEFINED ${_variable})
     FOREACH(_arg ${_argn})
-      LIST(APPEND CMAKE_REQUIRED_LIBRARIES "-l${_arg}")
-      CHECK_CXX_COMPILER_FLAG("" ${_variable})
+      LIST(APPEND CMAKE_REQUIRED_LIBRARIES "${_arg}")
+      CHECK_CXX_SOURCE_COMPILES("int main(){}" ${_variable})
```
tamiko, line 32: Similarly, we can no longer check whether `-l<system library>` is a valid linker flag via `CHECK_CXX_COMPILER_FLAG` but have to rely on `CHECK_CXX_SOURCE_COMPILES` instead.

.github/workflows/osx.yml
tamiko: Fixed.

tjhei (approved):

nfehn:
I realized that configuring dealii with an old cmake version (2.8) is no longer possible. I think it is related to this line

STRING(REGEX REPLACE "\[-+,\]" "_" _flag_name "${_flag_name}")

in https://github.com/dealii/dealii/pull/11384/files#diff-b1adf0303e6e4e30dbfe11ab916d102b16b910d2a85f071898ec492a4abe9a7a

tjhei:
> I realized that configuring dealii with an old cmake version (2.8) is no longer possible.

Oh. Can you please open a new issue for this? We can decide if we want to drop 2.8 support or make a fix.

PR #11387, opened 2020-12-17
MappingInfo: convert quadrature to face-quadrature rules

peterrum:
Add to `face_data[my_q].descriptor` more quadrature rules (in the case of wedges/pyramids 2). 

Follow up PR-s, will split up face ranges according to the face types and introduce a logic to access the correct face-quadrature rule depending on the face type.

### DIFF
--- a/include/deal.II/matrix_free/mapping_info.templates.h
+++ b/include/deal.II/matrix_free/mapping_info.templates.h
@@ -392,7 +392,7 @@ namespace internal
                   quad[my_q][q].get_tensor_basis()[0], update_default);
             }
 
-          face_data[my_q].descriptor.resize(n_hp_quads);
+          face_data[my_q].descriptor.reserve(n_hp_quads * 2);
           for (unsigned int hpq = 0; hpq < n_hp_quads; ++hpq)
             {
               bool flag = quad[my_q][hpq].is_tensor_product();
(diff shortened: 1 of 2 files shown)


include/deal.II/matrix_free/mapping_info.templates.h
```diff
@@ -392,7 +392,7 @@ namespace internal
                   quad[my_q][q].get_tensor_basis()[0], update_default);
             }
 
-          face_data[my_q].descriptor.resize(n_hp_quads);
+          face_data[my_q].descriptor.reserve(n_hp_quads * 2);
```
peterrum, line 395: @kronbichler I am wondering if we couldn't write here:
```suggestion
          face_data[my_q].descriptor.resize(n_hp_quads * std::max<unsigned int>(1, dim - 1));
```
with only the relevant descriptors filled.

Given a `hp_quad_index` and a `face_type`, we could access the correct face-quadrature in `FEEvaluationBase` via:
`face_data[quad_no].descriptor[std::max<unsigned int>(1, dim - 1) * hp_quad_index + face_type]` without the need of a complicated logic. `face_type` could  be e.g. stored in `FaceToCellTopology`. In 3D, `face_type` could be 0 (tet) or 1 (quad). I am currently trying to propagate the relevant information into the setup routines.

kronbichler: This looks reasonable. Please go ahead with the simplified version you suggest here.

peterrum: I am on it. If everything goes according to plans we should be able to evaluate face integrals for wedges today.

kronbichler (approved):

peterrum:
@kronbichler I am working on a version which matches the description of https://github.com/dealii/dealii/pull/11387#discussion_r545164483. It takes a bit longer than expected. I hope I have a running version later today. That version will contain many other changes that are not related to this PR, however, will enable us to discuss about the topic as a whole.

peterrum:
Closing in favor of #11400.

PR #11389, opened 2020-12-17
GridGenerator::convert_hypercube_to_simplex_mesh(): manifold_id

eliasstudiert:


source/grid/grid_generator.cc
```diff
@@ -7093,8 +7093,9 @@ namespace GridGenerator
 
   template <int dim, int spacedim>
   void
-  convert_hypercube_to_simplex_mesh(const Triangulation<dim, spacedim> &in_tria,
-                                    Triangulation<dim, spacedim> &out_tria)
+  convert_hypercube_to_simplex_mesh_develop(
```
peterrum, line 7096: I don't think you wanted to rename the function, did you?

eliasstudiert: of course not

source/grid/grid_generator.cc
```diff
@@ -7138,6 +7139,18 @@ namespace GridGenerator
          {{{{0, 1, 12}}, {{1, 12, 3}}, {{12, 3, 2}}, {{0, 12, 2}}}},
          {{{{4, 5, 13}}, {{5, 13, 7}}, {{13, 7, 6}}, {{4, 13, 6}}}}}};
 
+    static const std::array<std::array<unsigned int, 2>, 8>
+      table_2D_inner_manifold = {{{{6, 4}},
```
peterrum, line 7143:
```suggestion
      table_2D_inner_faces = {{{{6, 4}},
```

source/grid/grid_generator.cc
```diff
@@ -7260,26 +7294,54 @@ namespace GridGenerator
...
+                             current_face_manifold_id);
+                else if (dim == 3)
+                  {
+                    // nothing to do, since manifold inheritance in 3D not
+                    // implemented yet
+                  }
+                else
+                  Assert(false, ExcNotImplemented());
+              }
+          }
```
peterrum, line 7338: This function does not look correct. It should be more like:
```cpp
        for (const auto f : cell.face_indices())
            {
              const auto bid = cell.face(f)->boundary_id();
              const auto mid = cell.face(f)->manifold_id();
              if(bid==numbers::internal_face_boundary_id && mid==numbers::flat_manifold_id)
                continue;

              if (dim == 2) // 2D boundary_lines
                for (const auto &face : table_2D_boundary_faces[f])
                  add_cell(1, face, bid, mid);
              else if (dim == 3) // 3D boundary_quads
                for (const auto &face : table_3D_boundary_faces[f])
                  add_cell(2, face, bid, mid);
              else
            }
      }

     // process internal lines
     if (dim == 2 && cell.manifold_id()!=numbers::flat_manifold_id)
       for (const auto &face : table_2D_boundary_faces[f])
         add_cell(1, face, numbers::internal_face_boundary_id, current_face_manifold_id);
```

source/grid/grid_generator.cc
```diff
@@ -7260,26 +7294,54 @@ namespace GridGenerator
...
       }
 
     out_tria.create_triangulation(vertices, cells, subcell_data);
   }
 
 
-
```
peterrum, line 7282: Could you add this empty line back. We normally put three empty lines.

peterrum (commented):
Looks good till now. Could you also add a test?

tests/simplex/conv_hex_to_simplex.cc
```diff
@@ -117,17 +136,17 @@ main()
...
+  //   "2D: conversion triangulation with quad elements to tri elements: ");
+  // check_file<2, 3>();
+  // deallog.pop();
+
+  // // TETRAHEDRAL ELEMENTS
+  // // dim = spacedim = 3
+  // deallog.push(
+  //   "3D: conversion triangulation with tet elements to hex elements: ");
+  // check_file<3, 3>();
+  // deallog.pop();
```
peterrum, line 151: ?

tests/simplex/conv_hex_to_simplex.cc
```diff
@@ -99,6 +106,18 @@ check_file() // for dim = spaceim
...
+      // Temporary: See results in paraview
+      std::ofstream out1("Quad.vtk");
+      GridOut grid_out1;
+      grid_out1.write_vtk(in_tria, out1);
+      std::cout << "Grid written to Quad.vtk" << std::endl;
+
+      std::ofstream out2("Tri.vtk");
+      GridOut grid_out2;
+      grid_out2.write_vtk(out_tria, out2);
+      std::cout << "Grid written to Tri.vtk" << std::endl;
```
peterrum, line 118: ?

source/grid/grid_generator.cc
```diff
@@ -7214,13 +7227,16 @@ namespace GridGenerator
...
+                  cell_data.manifold_id = manifold_id; // set cell-manifold id
                 }
               cells.push_back(cell_data);
             }
           else if (dim == 2 && struct_dim == 1) // an edge of a triangle
             {
               CellData<1> boundary_line(2);
               boundary_line.boundary_id = material_or_boundary_id;
+              boundary_line.manifold_id =
+                manifold_id; // set manifold id on faces
```
peterrum, line 7239:
```suggestion
              boundary_line.manifold_id = manifold_id;
```

source/grid/grid_generator.cc
```diff
@@ -7250,8 +7267,24 @@ namespace GridGenerator
...
+            const auto current_manifold_of_cell =
+              cell.manifold_id(); // get cell-manifold id from current quad cell
+            for (const auto &cell : table_2D_cell)
+              add_cell(dim,
+                       cell,
+                       current_material_id,
+                       current_manifold_of_cell); // set cell-manifold id
+
+            // inherit inner manifold (faces)
+            for (const auto &face : table_2D_inner_faces)
```
peterrum, line 7280:
```suggestion
            // inherit inner manifold (faces)
            if(current_manifold_of_cell != numbers::flat_manifold_id)
            for (const auto &face : table_2D_inner_faces)
```

source/grid/grid_generator.cc
```diff
@@ -7250,8 +7267,24 @@ namespace GridGenerator
         // create cells one by one
 
         if (dim == 2)
-          for (const auto &cell : table_2D_cell)
-            add_cell(dim, cell, current_material_id);
+          {
+            const auto current_manifold_of_cell =
```
peterrum, line 7271:
```suggestion
            const auto current_manifold_id_of_cell =
```

source/grid/grid_generator.cc
```diff
@@ -7260,19 +7293,44 @@ namespace GridGenerator
...
+            else
+              Assert(false, ExcNotImplemented());
+
+            // process internal faces
+            if (dim == 2 && cell.manifold_id() != numbers::flat_manifold_id)
+              for (const auto &face : table_2D_boundary_faces[f])
+                add_cell(1,
+                         face,
+                         numbers::internal_face_boundary_id,
+                         current_manifold_id_of_face);
```
peterrum, line 7332:
```suggestion
```
This is not necessary? It is dealt with above.

source/grid/grid_generator.cc
```diff
@@ -7260,19 +7293,44 @@ namespace GridGenerator
...
-                for (const auto &face : table_3D_boundary_faces[f])
-                  add_cell(2, face, current_boundary_id);
-              else
-                Assert(false, ExcNotImplemented());
-            }
+          {
+            const auto current_boundary_id_of_face =
+              cell.face(f)->boundary_id();
+            const auto current_manifold_id_of_face =
+              cell.face(f)->manifold_id();
```
peterrum, line 7300:
```suggestion
            const auto bid = cell.face(f)->boundary_id();
            const auto mid = cell.face(f)->manifold_id();
```

tests/simplex/conv_hex_to_simplex.cc
eliasstudiert: commented out since 3D is not implemented yet.

tests/simplex/conv_hex_to_simplex.cc
eliasstudiert: will be removed!

tests/simplex/conv_hex_to_simplex.cc
peterrum: It is implemented. Just the manifold ids are not copied...

eliasstudiert: ah your right, but I think the refine_global() would cause an error atm.

source/grid/grid_generator.cc
```diff
@@ -7138,6 +7138,103 @@ namespace GridGenerator
          {{{{0, 1, 12}}, {{1, 12, 3}}, {{12, 3, 2}}, {{0, 12, 2}}}},
          {{{{4, 5, 13}}, {{5, 13, 7}}, {{13, 7, 6}}, {{4, 13, 6}}}}}};
 
+    static const std::array<std::array<unsigned int, 2>, 8>
+      table_2D_inner_faces = {{{{6, 4}},
```
bangerth, line 7142: Can you add a couple of lines of comments in front of each of these tables? I have no idea what these variables mean -- presumably, they *map* something *to* something, but it's not clear what it is. I think it would be useful to name these variables better (e.g. `boundary_face_to_cell_map_2d` or similar, assuming that's what this is: in other words, a `X_to_Y_map` name). It would also be good if each had a description before its declaration that illustrates what the indices actually are. This one, for example, is a table with two indices, but what are they? Face numbers? Vertex numbers? Of quads? Of triangles?

source/grid/grid_generator.cc
```diff
@@ -7214,31 +7312,46 @@ namespace GridGenerator
...
             }
+          else if (dim == 3 && struct_dim == 1) // an edge of a tetrahedron
+            {
+              CellData<1> boundary_line(2);
+              // boundary_line.boundary_id = material_or_boundary_id;
+              boundary_line.manifold_id = manifold_id;
+              for (unsigned int i = 0; i < index_vertices.size(); ++i)
+                {
+                  boundary_line.vertices[i] =
+                    local_vertex_indices[index_vertices[i]];
```
bangerth, line 7351: This will be a bug if `index_vertices.size()>2`. That should of course never happen, but it would still be good to put a
```
Assert (index_vertices.size() == 2, ExcInternalError());
```
at the top of this block (line 7345, before the declaration of `boundary_line`).

bangerth: In fact, such an assertion would be useful in all of the other blocks above as well.

source/grid/grid_generator.cc
```diff
@@ -7250,29 +7363,101 @@ namespace GridGenerator
         // create cells one by one
 
         if (dim == 2)
-          for (const auto &cell : table_2D_cell)
-            add_cell(dim, cell, current_material_id);
+          {
+            const auto current_manifold_id_of_cell =
+              cell.manifold_id(); // get cell-manifold id from current quad cell
+            for (const auto &cell : table_2D_cell)
```
bangerth, line 7369: This is pre-existing, but it's confusing nonetheless. In line 7150 (old line number) above, you already have
```
 for (const auto &cell : in_tria)
```
so you have a local variable `cell` that is a `cell_iterator`. Here, you introduce another local variable `cell` of a different data type that is not actually a cell but (I believe) a triplet of numbers. Choose a different variable name to avoid this kind of confusion. Try to reflect in the variable name what the triplet actually represents.

source/grid/grid_generator.cc
```diff
@@ -7250,29 +7363,101 @@ namespace GridGenerator
...
+              add_cell(dim,
+                       cell,
+                       current_material_id,
+                       current_manifold_id_of_cell); // set cell-manifold id
```
bangerth, line 7373: Drop the comment. Since it's after the whole call, it looks like you're commenting on what the function call here does, but I think what you're really doing is only commenting on the last line -- but since the previous lines don't have such comments, this looks out of place.

source/grid/grid_generator.cc
```diff
@@ -7250,29 +7363,101 @@ namespace GridGenerator
...
-          for (const auto &cell : table_3D_cell)
-            add_cell(dim, cell, current_material_id);
+          {
+            const auto current_manifold_id_of_cell = cell.manifold_id();
+            for (const auto &cell : table_3D_cell)
+              add_cell(
+                dim,
+                cell,
+                current_material_id,
+                current_manifold_id_of_cell); // set all cell manifold ids
```
bangerth, line 7394: same here

source/grid/grid_generator.cc
```diff
@@ -7250,29 +7363,101 @@ namespace GridGenerator
...
+              cell.manifold_id(); // get cell-manifold id from current quad cell
+            for (const auto &cell : table_2D_cell)
+              add_cell(dim,
+                       cell,
+                       current_material_id,
+                       current_manifold_id_of_cell); // set cell-manifold id
+
+            // inherit inner manifold (faces)
+            if (current_manifold_id_of_cell != numbers::flat_manifold_id)
+              for (const auto &face : table_2D_inner_faces)
```
bangerth, line 7377: same comment about using a better name for the `face` variable. The same applies in the code pieces below.

source/grid/grid_generator.cc
```diff
@@ -7250,29 +7363,101 @@ namespace GridGenerator
...
+                dim,
+                cell,
+                current_material_id,
+                current_manifold_id_of_cell); // set all cell manifold ids
+
+            if (current_manifold_id_of_cell != numbers::flat_manifold_id)
+              {
+                // set manifold of inner FACES of tets according to hex cell
+                // manifold id
+                for (const auto &face : table_3D_inner_faces)
```
bangerth, line 7400: I *think* that this is the only place where you're actually using this variable. It might be easier to read if you just moved the variable declaration right into this block, before the `for` (or if you're bold, into `for (... : HERE)`). This way it's more clear what the variable represents because it is declared right next to the only place where it's used and we can see from the comment above the `for` loop and the context what the triplets are going to represent.

tests/simplex/conv_hex_to_simplex.cc
```diff
@@ -99,6 +106,18 @@ check_file() // for dim = spaceim
     grid_out.write_vtk(tria, deallog.get_file_stream());
   };
 
+  // Temporary: See results in paraview
+  std::ofstream out1("Quad.vtk");
+  GridOut       grid_out1;
+  grid_out1.write_vtk(in_tria, out1);
```
bangerth, line 112: This doesn't show up in the test output, and so if someone breaks the functionality, we will not learn about it. You'll have to put the VTK output into the `.output` file in some form.

tests/simplex/conv_hex_to_simplex.cc
bangerth: This is awkward. You're removing things from the test that used to be checked whenever someone ran this test. That's a recipe for stuff breaking. You'll have to investigate what happens when you leave the code in.

bangerth (changes requested):
Nice job continuing with this functionality!

source/grid/grid_generator.cc
eliasstudiert: I haven't changed the names of these tables yet, because I personally think `X_to_Y_map` is a little misleading. How about: `vertices_id_boundary_faces_3d`,   `vertices_id_inner_edge_3d` and so on? Because the table contains the IDs of the vertices that form e.g. a face on the boundary or an edge that is inside.

source/grid/grid_generator.cc
eliasstudiert: I totally agree to change those variable names!

source/grid/grid_generator.cc
eliasstudiert: I've changed the name of the variable, but I don't understand which variable declaration is meant.

tests/simplex/conv_hex_to_simplex.cc
eliasstudiert: I was confused due to an error of refine_global() because it was not implemented in 3D yet, I think. So I've commented out the `out_tria.refine_global(2);` in the test-program to avoid that error.

peterrum (approved):

peterrum:
@bangerth Any objections in merging this one?

source/grid/grid_generator.cc
bangerth: How about `vertex_ids_for_boundary_faces` as a name?

bangerth:
I think there's just the one issue with the names of the tables. I believe that you've addressed all of the other issues -- my apologies for not getting back to this right away :-(

source/grid/grid_generator.cc
eliasstudiert: that sounds good to me, assuming we'll additionally add _2d or _3d, respectively.

eliasstudiert:
I merged sth by accident, I hope that I reverted it correctly. Otherwise please let me know what I did wrong.

source/grid/grid_generator.cc
bangerth: Yes, the suffix is fine of course.

bangerth:
I think you want to *drop* the last two commits (the merge and the reversion of the merge). If you have someone who can help you with git, it would be worthwhile to do this step with someone (virtually) looking over your shoulder. The idea is to do a `git rebase -i` and just don't select the two merge/unmerge commits by deleting these rows.

peterrum:
> I think you want to drop the last two commits (the merge and the reversion of the merge).

The problem was more severe (some of the modifications from this PR has been missing). Luckily, I had the current most current version as backup somewhere. I took the liberty to rename the fields there and to force push myself.

@eliasstudiert For the next time: try not to merge. Double check what is online. If you want to get the newest version of master, prefer rebasing. If you have problems locally, don't force push to the PR: you can consider the code online as your backup and if you force push you might not have a chance to get to the backup anymore.

bangerth (approved):

PR #11390, opened 2020-12-17
Let Particle::get_(reference_)location() return by-value rather than reference

bangerth:
More prep work for moving locations from `Particle` to `PropertyPool`: Let some functions return points by value, rather than by reference. These are objects easily scalarized by compilers, so this might even lead to marginally better code.

In reference to #11206.

/rebuild


bangerth:
This does not work because some part of the BOOST geometry library wants to access individual elements of a location for writing.

PR #11393, opened 2020-12-18
Added missing libraries to Ginkgo linking

nabw:
There were missing references to Ginkgo libraries which give "undefined symbol" errors at compile time.


drwells:
Will this still work with older versions of ginkgo?

masterleinad:
Doesn't GINKGO export is configuration properly such that we can just use
```
# temporarily disable ${CMAKE_SOURCE_DIR}/cmake/modules for module lookup
LIST(REMOVE_ITEM CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)
FIND_PACKAGE(GINKGO)
LIST(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)
```
and include a library that includes all of the other backend-specific ones?

nabw:
I will check this, I'm not sure but apparently yes https://github.com/ginkgo-project/ginkgo/issues/678#issuecomment-748023189. Also, I will check with versions 1.1 and 1.2 if compilations works fine as validation.

nabw:
> Doesn't GINKGO export is configuration properly such that we can just use
> 
> ```
> # temporarily disable ${CMAKE_SOURCE_DIR}/cmake/modules for module lookup
> LIST(REMOVE_ITEM CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)
> FIND_PACKAGE(GINKGO)
> LIST(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules/)
> ```
> 
> and include a library that includes all of the other backend-specific ones?

Only a quick update, I tried this without success, but I am not sure about what you mean with including a specific library. I do include the libginkgo library (through a DEAL_II_FIND_LIBRARY), as well as refer to the ginkgo.hpp file (through a DEAL_II_FIND_PATH) from the FindGinkgo.cmake file. Sorry, I don't have much experience with CMake. Could you elaborate?

If anything, removing the extra libraries and using a FIND_PACKAGE to include everything does not seem to work out of the box. What would be nice to refer to is the GinkgoConfig file in the Ginkgo installation (lib) folder (install-folder/lib/cmake/Ginkgo/GinkgoConfig.cmake).

Best!
NB

masterleinad:
Which CMake variables does `GinkgoConfig.cmake` define? Is there one that includes all libraries (`libginkgo_dpcpp`, `libginkgo_hip`, `libginkgo_cuda`, `libginkgo_omp`,...)

pratikvn:
@nabw , thanks for looking into this, I am working on updating these CMake issues so that we don't have to manually add each library here everytime we update Ginkgo with sub-libraries. With Ginkgo, the target `Ginkgo::ginkgo` should automatically link the ginkgo library and all its dependencies (hip, cuda, dpcpp ...) so adding this should be sufficient.

Additionally, with these changes I also wanted to update the wrapper with some of the newer algorithms that we have added. I also need to update the tests as some iteration counts have now changed due to changes in the algorithms.

If it is okay with you, I would prefer to create a PR to add these changes. 

Thanks,
Pratik.

cmake/modules/FindGINKGO.cmake
```diff
@@ -64,9 +80,9 @@ DEAL_II_FIND_PATH(GINKGO_INCLUDE_DIR ginkgo/ginkgo.hpp
 
 DEAL_II_PACKAGE_HANDLE(GINKGO
   LIBRARIES
-    REQUIRED GINKGO_LIBRARY GINKGO_REFERENCE_LIBRARY GINKGO_OMP_LIBRARY GINKGO_CUDA_LIBRARY
+    REQUIRED GINKGO_LIBRARY GINKGO_REFERENCE_LIBRARY GINKGO_OMP_LIBRARY GINKGO_CUDA_LIBRARY GINKGO_HIP_LIBRARY GINKGO_DPCPP_LIBRARY
```
bangerth, line 83: Does gingko *always* build for HIP and DPCPP? I'm asking because if you're on a CPU-only platform, will this actually work?

pratikvn: No, the different libraries (`ginkgo_cuda`, `ginkgo_hip`, ... ) are always *created*. When on a platform that does not support any of the specific compilers/devices needed for the modules then these libraries contain dummy implementations. 

It is safe to assume that these libraries will always be available (with the appropriate version of Ginkgo) and for example if you are on a CPU-only system and try to create a `HipExecutor` to run on AMD GPU, then a runtime exception is thrown.

pratikvn:
@nabw , This [commit](https://github.com/pratikvn/dealii/commit/f06e331950ec31124b24feb5e1c9fec199cf541c) for all versions of Ginkgo.

nabw:
Thank you all for the quick response! @pratikvn That's great, then tell me how you prefer to proceed. Should I close the request so that you can create another one with the commit you mentioned?

pratikvn:
Yes, if that is okay with you. I can create a new PR with these and probably a few more changes. thanks!

nabw:
Not at all, let me know when it is ready. 

Best, 
NB

tamiko:
Obsoleted by #11402, #11413

PR #11399, opened 2020-12-19
Remove compatibility code for GCC versions prior to 5.0.

bangerth:
We no longer support these compilers, so we can just remove the code without replacement.

Seen while looking at #11368.


kronbichler (approved):

kronbichler:
Test failure is unrelated.

tjhei (approved):

PR #11400, opened 2020-12-19
Use QCollection within MappingInfo::initialize

peterrum:
The relevant part of this PR is (needed for `FEFaceEvaluation`):

https://github.com/dealii/dealii/blob/9a6768579ab1bd837885788cc940324c1970fd36/include/deal.II/matrix_free/mapping_info.h#L461-L462

~Only the last commit is relevant. Depends on #10764.~

### DIFF
--- a/include/deal.II/matrix_free/mapping_info.h
+++ b/include/deal.II/matrix_free/mapping_info.h
@@ -27,6 +27,8 @@
 #include <deal.II/fe/fe.h>
 #include <deal.II/fe/mapping.h>
 
+#include <deal.II/grid/reference_cell.h>
+
 #include <deal.II/hp/mapping_collection.h>
 #include <deal.II/hp/q_collection.h>
 
@@ -193,6 +195,14 @@ namespace internal
        * orientation.
        */
       std::vector<QuadratureDescriptor> descriptor;
+
+      /**
+       * Collection of quadrature formulae applied on the given face.
+       *
+       * @note Only filled for faces, since faces might be quadrilateral or
+       *   triangle shaped.
+       */
+      std::vector<dealii::hp::QCollection<structdim>> q_collection;
 
       /**
        * Stores the index offset into the arrays @p jxw_values, @p jacobians,
@@ -453,6 +463,12 @@ namespace internal
        * The pointer to the first entry of mapping_collection.
        */
       SmartPointer<const Mapping<dim>> mapping;
+
+      /**
+       * Reference-cell type related to each quadrature and active quadrature
+       * index.
+       */
+      std::vector<std::vector<dealii::ReferenceCell>> reference_cell_types;
 
       /**
        * Internal function to compute the geometry for the case the mapping is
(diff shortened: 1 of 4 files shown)


peterrum:
In a follow-up PR, I will clean up:

https://github.com/dealii/dealii/blob/9a6768579ab1bd837885788cc940324c1970fd36/include/deal.II/matrix_free/mapping_info.templates.h#L365-L470

The setup as it is now does not match the logic we need. We need:
- the cell quadrature rule (for each quadrature and active quadrature index) -> `cell_data[mq_q].description[q]`
- the face quadrature rules for each face of a cell in form of a `QCollection` (for each quadrature and active quadrature index - as proposed in this PR)
- the unique face quadrature rules on the faces of a cell (for each quadrature and active quadrature index - as discussed in https://github.com/dealii/dealii/pull/11387#discussion_r545726479) -> in 3D `cell_face[mq_q].description[2*q + 0]` for quads and  `cell_face[mq_q].description[2*q + 1]` for triangles

include/deal.II/matrix_free/mapping_info.h
```diff
@@ -454,6 +454,13 @@ namespace internal
        */
       SmartPointer<const Mapping<dim>> mapping;
 
+      /**
+       * Face QCollections related to each quadrature and active quadrature
+       * index.
+       */
+      std::vector<std::vector<dealii::hp::QCollection<dim - 1>>>
+        face_q_collection;
```
kronbichler, line 462: I do not think this is the right place for this. We have tried to collect all data specific to the hp case into the `descriptor` class. Now this is a slightly different use to the wedge/pyramid case, but then it should be in `MappingInfoStorage` and not in the main class.

peterrum: I don't say that it is perfect as it is now but I don't think `MappingInfoStorage` is the right place. The information is directly liked to `MappingInfo::cell_data` and not to `MappingInfo::face_data`/`MappingInfo::face_data_by_cells`. 

Maybe PR #11401, where I have completely refactored the setup routine (and am hopefully close to getting face integrals running for wedges/pyramids), can demonstrate the relations more clearly:

https://github.com/dealii/dealii/blob/e88f1debe12f61e091b3893db6eb88025fd730c9/include/deal.II/matrix_free/mapping_info.templates.h#L365-L445

peterrum: Alternatively, we could move `face_q_collection` to `MappingInfoStorage` and keep it in `MappingInfo::face_data` empty, or don't save and pass the information directly to relevant functions (in the assumption that it will never be used!?).

peterrum: Please not also the need for the reference-cell type, which can be later passed to a `FE_Nothing` object (see also PR #11403).

peterrum:

peterrum:
It seems this PR deepens on #11401.

peterrum:
I have rebased this PR. It is ready from my side!

include/deal.II/matrix_free/mapping_info.h
kronbichler: Yes, I think putting it in `MappingInfoStorage` but leaving things empty for the cases where it makes no sense (say `face_data`).

kronbichler (approved):
Makes sense, apart from the place where to store it (in my opinion).

include/deal.II/matrix_free/mapping_info.h
peterrum: OK. I have changed it.

kronbichler (approved):
Thank you!

PR #11372, opened 2020-12-14
Generalize ShapeInfo::reinit

peterrum:
Depends on #11361.

PR #11401, opened 2020-12-19
MatrixFree: introduce FaceToCellTopology::face_type

peterrum:
This PR introduces a new `FaceToCellTopology::face_type`:
https://github.com/dealii/dealii/blob/5c47e535302b535ef510f1dca2aad2bc946581ee/include/deal.II/matrix_free/face_info.h#L113

Face ranges are partitioned now also according the face type:
https://github.com/dealii/dealii/blob/5c47e535302b535ef510f1dca2aad2bc946581ee/include/deal.II/matrix_free/matrix_free.h#L1490-L1496
https://github.com/dealii/dealii/blob/5c47e535302b535ef510f1dca2aad2bc946581ee/include/deal.II/matrix_free/matrix_free.h#L1504-L1509

The face type is passed to the constructor of `FEEvaluationBaseData` to select the correct `descriptor`:
https://github.com/dealii/dealii/blob/5c47e535302b535ef510f1dca2aad2bc946581ee/include/deal.II/matrix_free/fe_evaluation.h#L3370-L3375
selecting between the quadrature rules of a triangle or quad in the case of 3D.

@kronbichler What do you think? Any concerns? Do you see a way how to simplify this procedure. From a `FE(Face)Evaluation` perspective, they should be ready for wedges/pyramids (once the underlying kernels are extended to use a newer version of `ShapeInfo`). Now the relevant info has to be set up in `MappingInfo` (PR #11400 and a modified version of #11387) and `ShapeInfo` (PR #11372).

depends on #11400

### DIFF
--- a/include/deal.II/matrix_free/face_info.h
+++ b/include/deal.II/matrix_free/face_info.h
@@ -111,6 +111,12 @@ namespace internal
       unsigned char face_orientation;
 
       /**
+       * Reference-cell type of the given face: 0 for line or quadrilateral,
+       * 1 for triangle.
+       */
+      unsigned char face_type;
+
+      /**
        * Return the memory consumption of the present data structure.
        */
       std::size_t
--- a/include/deal.II/matrix_free/fe_evaluation.h
+++ b/include/deal.II/matrix_free/fe_evaluation.h
@@ -298,7 +298,8 @@ protected:
     const unsigned int n_q_points,
     const bool         is_interior_face,
     const unsigned int active_fe_index,
-    const unsigned int active_quad_index);
+    const unsigned int active_quad_index,
+    const unsigned int face_type);
 
   /**
    * Constructor that comes with reduced functionality and works similar as
@@ -1143,7 +1144,8 @@ protected:
     const unsigned int n_q_points,
     const bool         is_interior_face,
     const unsigned int active_fe_index,
-    const unsigned int active_quad_index);
+    const unsigned int active_quad_index,
+    const unsigned int face_type);
 
   /**
    * Constructor that comes with reduced functionality and works similar as
@@ -1428,7 +1430,8 @@ protected:
     const unsigned int n_q_points,
     const bool         is_interior_face  = true,
     const unsigned int active_fe_index   = numbers::invalid_unsigned_int,
-    const unsigned int active_quad_index = numbers::invalid_unsigned_int);
+    const unsigned int active_quad_index = numbers::invalid_unsigned_int,
+    const unsigned int face_type         = numbers::invalid_unsigned_int);
 
   /**
    * Constructor with reduced functionality for similar usage of FEEvaluation
@@ -1565,7 +1568,8 @@ protected:
     const unsigned int n_q_points,
     const bool         is_interior_face  = true,
     const unsigned int active_fe_index   = numbers::invalid_unsigned_int,
-    const unsigned int active_quad_index = numbers::invalid_unsigned_int);
+    const unsigned int active_quad_index = numbers::invalid_unsigned_int,
+    const unsigned int face_type         = numbers::invalid_unsigned_int);
 
   /**
    * Constructor with reduced functionality for similar usage of FEEvaluation
@@ -1726,7 +1730,8 @@ protected:
     const unsigned int n_q_points,
     const bool         is_interior_face  = true,
     const unsigned int active_fe_index   = numbers::invalid_unsigned_int,
-    const unsigned int active_quad_index = numbers::invalid_unsigned_int);
+    const unsigned int active_quad_index = numbers::invalid_unsigned_int,
+    const unsigned int face_type         = numbers::invalid_unsigned_int);
 
   /**
    * Constructor with reduced functionality for similar usage of FEEvaluation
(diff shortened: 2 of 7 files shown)


peterrum:
@kronbichler This PR should should show everything what we need to be able to evaluate face integrals on meshes with wedges and pyramids (I would not be surprised if the code would already work for mixed meshes). Apart from the question on how to encode the type of a face (and how to handle subranges), we need to agree what kind of information we want to store:
- some kind of `face_type` is needed to be able to distinguish between quadrilateral and triangular faces:
https://github.com/dealii/dealii/blob/410e0daccbd2a3b5ab737b4887067cf7040e6db6/include/deal.II/matrix_free/face_info.h#L113-L117
- to be ablt to set up the mapping on the faces (for which we need `FEFaceValues`), we need a `QCollection` collecting the quadrature rules on each face and a reference-cell type (to create a `FENothing`):
https://github.com/dealii/dealii/blob/410e0daccbd2a3b5ab737b4887067cf7040e6db6/include/deal.II/matrix_free/mapping_info.h#L457-L468
However, it turns out that these two information are currently only needed during setup so that we not necessarily have to save them in `MappingInfo` but can pass them directly to the correct functions.
- we need to store the number of quadrature points on each face:
https://github.com/dealii/dealii/blob/410e0daccbd2a3b5ab737b4887067cf7040e6db6/include/deal.II/matrix_free/shape_info.h#L423-L427
- we already are storing the shape functions evaluated in the quadrature points of each face for different orientations:
https://github.com/dealii/dealii/blob/410e0daccbd2a3b5ab737b4887067cf7040e6db6/include/deal.II/matrix_free/shape_info.h#L286-L298

Furthermore, I would like to point out the redefinition of `MappingInfo::face_data`, which now stores for each cell-quadrature rule the minimal number of face-quadrature rules. 

The needed information (quadrature collection and minimal set of face-quadrature rules) could be provided by the following utility functions:
https://github.com/dealii/dealii/blob/410e0daccbd2a3b5ab737b4887067cf7040e6db6/include/deal.II/matrix_free/util.h#L34-L116

include/deal.II/matrix_free/face_info.h
```diff
@@ -110,6 +110,12 @@ namespace internal
        */
       unsigned char face_orientation;
 
+      /**
+       * Reference-cell type of the given face: 0 for line or quadrilateral,
+       * 1 for triangle.
+       */
+      unsigned char face_type = 0;
```
kronbichler, line 117: Can you remove the in-line default definition and check that things are set correctly? One thing I wonder is if we access this field in a typical matrix-free evaluation loop nowadays - I always tried to make metadata like the present `struct` as small as possible to avoid too much superfluous cache line transfer during the mat-vec. We probably do, because we access at least the orientation bit. I would be curious if there is a measurable impact on performance.

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -1410,7 +1412,8 @@ class FEEvaluationAccess : public FEEvaluationBase<dim,
     const unsigned int n_q_points,
     const bool         is_interior_face  = true,
     const unsigned int active_fe_index   = numbers::invalid_unsigned_int,
-    const unsigned int active_quad_index = numbers::invalid_unsigned_int);
+    const unsigned int active_quad_index = numbers::invalid_unsigned_int,
+    const unsigned int face_no           = numbers::invalid_unsigned_int);
```
kronbichler, line 1416: Should this be `face_type`?

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -3051,6 +3057,9 @@ class FEFaceEvaluation : public FEEvaluationAccess<dim,
    * DoFHandler/AffineConstraints pair the given evaluator should be attached
    * to.
    *
+   * @param face_type In the case of a face, its reference-cell type
+   * (0 for line or quadrilateral 1 for triangle).
+   *
```
kronbichler, line 3062:
```suggestion
   * @param face_type In the case of a face, indicate its reference-cell type
   * (0 for line or quadrilateral 1 for triangle).
   *
```

kronbichler:
I think we should get this going now. It seems we have accumulated a few things on the way, so I suggest that we rebase this branch. I agree with the overall concept and currently only have some small comments.

include/deal.II/matrix_free/face_info.h
peterrum: > One thing I wonder is if we access this field in a typical matrix-free evaluation loop nowadays - I always tried to make metadata like the present struct as small as possible to avoid too much superfluous cache line transfer during the mat-vec. We probably do, because we access at least the orientation bit. I would be curious if there is a measurable impact on performance.

We don't need it for each cell once everything is set up. Then, we only need it for setting up `FEFaceEvaluation` (since the value is the same for the whole cell range). I would suggest to move `face_type` to `FaceInfo` in a follow-up PR.

peterrum:

include/deal.II/matrix_free/face_info.h
kronbichler: OK

kronbichler (approved):

PR #11403, opened 2020-12-20
FE_Nothing: pass reference-cell type to constructor

peterrum:
Needed by #11401.

### DIFF
--- a/include/deal.II/fe/fe_nothing.h
+++ b/include/deal.II/fe/fe_nothing.h
@@ -79,15 +79,26 @@ class FE_Nothing : public FiniteElement<dim, spacedim>
 {
 public:
   /**
-   * Constructor. First argument denotes the number of components to give this
+   * Constructor.
+   *
+   * The first argument specifies the reference-cell type.
+   *
+   * The second argument denotes the number of components to give this
    * finite element (default = 1).
    *
-   * Second argument decides whether FE_Nothing will dominate any other FE in
+   * The third argument decides whether FE_Nothing will dominate any other FE in
    * compare_for_domination() (default = false). Therefore at interfaces where,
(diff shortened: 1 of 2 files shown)


include/deal.II/fe/fe_nothing.h
```diff
@@ -91,6 +91,13 @@ class FE_Nothing : public FiniteElement<dim, spacedim>
    */
   FE_Nothing(const unsigned int n_components = 1, const bool dominate = false);
 
+  /**
+   * Same as above but with the possibility to specify a reference-cell type.
```
Rombur, line 95: You don't have the possibility to specify a reference-cell type, you have to specify it. Right?

source/fe/fe_nothing.cc
```diff
@@ -41,6 +43,17 @@ FE_Nothing<dim, spacedim>::FE_Nothing(const unsigned int n_components,
 }
 
 
+
+template <int dim, int spacedim>
+FE_Nothing<dim, spacedim>::FE_Nothing(const unsigned int n_components,
+                                      const bool         dominate)
+  : FE_Nothing<dim, spacedim>(ReferenceCell::get_hypercube(dim),
```
Rombur, line 50: So by default, we assume the cell type is square or hex. Should we document this?

Rombur (changes requested):

source/fe/fe_nothing.cc
peterrum: I have updated the documentation.

Rombur (approved):
Looks good but the CI is failing.

drwells (approved):
This is fine as-is but I can think of some improvements (like updating `get_name()`) - since this is blocking another PR and those are secondary I'll merge this now and write those improvements in a followup.

PR #11402, opened 2020-12-19
Update Ginkgo cmake detection and tests.

pratikvn:
This PR updates and simplifies the CMake detection of Ginkgo with a `find_package(Ginkgo)` so that each of the library modules of Ginkgo do not have to be explicitly found and linked. 

It also updates the iteration counts for the tests as the GMRES algorithm in Ginkgo has been updated.

Things to discuss: Since the previous addition, we have added a few features into Ginkgo, which currently do not have an interface to deal.ii. The important additions relevant to deal.ii are:

1. A HIP Executor for execution on AMD GPUs (v 1.2.0)
2. BiCG solver (v1.2.0)
3. ILU-type preconditioners (v1.2.0)
4. Incomplete Sparse approximate inverse preconditioners (v1.2.0)
5. Permutations and Sparse matrix representations (v1.2.0 and v1.1.0 respectively)

If I add the ability to run on AMD GPUs with the `HipExecutor`, then to make sure that we have backward compatibility with older versions of Ginkgo, I would need to `#if` guard these additions with the Ginkgo version numbers. Is this acceptable for you ? 

This PR supersedes #11393 

Closes #10008

PR #11410, opened 2020-12-22
Revert "Merge pull request #11402 from pratikvn/fix-ginkgo-cmake"

Rombur:
#11402 breaks the CI. We configure deal.II to use C++17 but at compile time we use the default compiler on the system which does not support C++17.


peterrum (approved):

bangerth:
Is this urgent? Does it break for everyone, or just if you have gingko installed?

masterleinad:
> Is this urgent? Does it break for everyone, or just if you have gingko installed?

A fair share of our regression tests is broken (https://cdash.43-1.org/index.php?project=deal.II). If it fixes that, I'd consider it pretty urgent.

bangerth:
Then let's revert once the testers are through.
@pratikvn -- can you think of an alternative way of achieving what you wanted?

/rebuild

bangerth (approved):

pratikvn:
Sorry, I am a bit confused. We require only at most C++14.

masterleinad:
Admittedly, I also don't quite understand how that pull request would have broken anything. It would be good to have a reproducer we can inspect. :slightly_smiling_face:

Rombur:
> Sorry, I am a bit confused. We require only at most C++14.

What I think is happening is the following: the testing machine has a default compiler that does not support C++17 and additional compilers that do support C++17. With this PR, somehow, the configuration step correctly selects a compiler that supports C++17 but during compilation the default compiler, which does not support C++17, is selected. Because of this we try to compile C++17 code with a compiler that does not support it. 

I am not sure that it is what's happening. We need to wait for the testers to check that reverting this PR fixes the problem.

tamiko:
The problem is that the ginkgo configuration exports a number of `GINKGO_CXX_FLAGS` containing a `-std=c++14` that our build system then dutifully adds to our compile flags.

tamiko:
There is a second, much bigger issue why this pull request must be reverted for now, see my second comment on #11402

PR #11413, opened 2020-12-23
Ginkgo Updates - 2nd try

tamiko:


tamiko:
@pratikvn This is your old pull request with three fixes:
 - deal.II now sanitizes all `${FEATURE_*}` variables in `DEAL_II_PACKAGE_HANDLE`. This ensures that we do no longer pick up the stray `GINKGO_CXX_FLAGS` variables,
 - we unset the `GINKGO_CXX_COMPILER` variable for cosmetic reasons;
 - we save and restore `CMAKE_MODULE_PATH` after the call to `find_package`.

cmake/modules/FindGINKGO.cmake
```diff
@@ -19,54 +19,53 @@
...
+    NAMES ${_library}
+    HINTS ${GINKGO_INSTALL_LIBRARY_DIR}
+    NO_DEFAULT_PATH
+    NO_CMAKE_ENVIRONMENT_PATH
+    NO_CMAKE_PATH
+    NO_SYSTEM_ENVIRONMENT_PATH
+    NO_CMAKE_SYSTEM_PATH
+    NO_CMAKE_FIND_ROOT_PATH
+    )
+ENDFOREACH()
```
tamiko, line 64: @pratikvn Ideally, I would like to recover the full library paths from the exported import targets. That's unfortunately a bit tricky. On my Gentoo Linux machine for example I have in `GinkgoTargets.cmake`:
```
# Import target "Ginkgo::ginkgo_cuda" for configuration "Gentoo"
set_property(TARGET Ginkgo::ginkgo_cuda APPEND PROPERTY IMPORTED_CONFIGURATIONS GENTOO)
set_target_properties(Ginkgo::ginkgo_cuda PROPERTIES
  IMPORTED_LOCATION_GENTOO "/var/tmp/portage/sci-libs/ginkgo-1.1.1-r1/work/ginkgo-1.1.1_build/core/device_hooks/libginkgo_cuda.so.1.1.1"
  IMPORTED_SONAME_GENTOO "libginkgo_cuda.so.1.1.1"
  )
```
Now querying for the compile configuration and querying the `IMPORTED_LOCATION_GENTOO` (or `RELEASE`, or `DEBUG`) is ugly but doable. The real issue is that this doesn't seem to be the installed target location but an exported target during compilation, i.e., a bug in the Ginkgo CMake configuration.

So for now, invoking `FIND_LIBRARY` again (and forcing the search path to `${GINKGO_INSTALL_LIBRARY_DIR}`) seems to be the best option - we do the same for Trilinos.

cmake/modules/FindGINKGO.cmake
```diff
@@ -19,54 +19,53 @@
...
-    REQUIRED GINKGO_LIBRARY GINKGO_REFERENCE_LIBRARY GINKGO_OMP_LIBRARY GINKGO_CUDA_LIBRARY
-  INCLUDE_DIRS REQUIRED GINKGO_INCLUDE_DIR
-  USER_INCLUDE_DIRS REQUIRED GINKGO_INCLUDE_DIR
-  CLEAR
-    GINKGO_LIBRARY GINKGO_REFERENCE_LIBRARY GINKGO_OMP_LIBRARY GINKGO_CUDA_LIBRARY GINKGO_INCLUDE_DIR
+  LIBRARIES REQUIRED ${_libraries}
+  INCLUDE_DIRS REQUIRED GINKGO_INSTALL_INCLUDE_DIR
+  USER_INCLUDE_DIRS REQUIRED GINKGO_INSTALL_INCLUDE_DIR
+  CLEAR Ginkgo_DIR ${_libraries}
   )
```
tamiko, line 71: I have fixed a subtle issue with the `DEAL_II_PACKAGE_HANDLE` macro: it now unsets all "feature" variables before they are populated by the macro. That way all the `GINKGO_CXX_FLAGS*` variables don't make it into our configuration any more. :smile:

cmake/modules/FindGINKGO.cmake
pratikvn: I see. It looks like we fixed this in a later version of Ginkgo (https://github.com/ginkgo-project/ginkgo/pull/414). But for backward compatibility, your solution looks good to me!

pratikvn (approved):
Looks good to me. Thanks for adding these changes!

cmake/modules/FindGINKGO.cmake
pratikvn: By the way, we can also enforce that deal.ii only supports Ginkgo v1.2.0 versions and above. We are okay with that and would probably prefer it as well.

masterleinad (approved):
LGTM

tamiko:
@pratikvn I like the idea to require ginkgo v1.2.0 and clean up our configuration. But I suggest we do that in a separate pull request. Let's merge this one first.

masterleinad:
Hmmm... we now get
```
 undefined reference to 'non-virtual thunk to gko::matrix::Dense<float>::convert_to(gko::matrix::Csr<float, long>*) const'
```
on all configurations, see https://cdash.43-1.org/viewBuildError.php?buildid=14850.

cmake/modules/FindGINKGO.cmake
```diff
@@ -19,54 +19,53 @@
...
+#
+unset(GINKGO_CXX_COMPILER)
+
+#
+# We'd like to have the full library names but the Ginkgo package only
+# exports a list with short names. So check again for every lib and store
+# the full path:
+#
+SET(_libraries "")
+FOREACH(_library ${GINKGO_INTERFACE_LINK_LIBRARIES})
```
pratikvn, line 52: `GINKGO_INTERFACE_LINK_LIBRARIES` only contains the list of libraries that `libginkgo.so` depends on so `libginkgo_reference.so`, `libginkgo_cuda.so` and `libginkgo_omp.so`, But we also need to link against `libginkgo.so` and hence the undefined reference errors.

Essentially `GinkgoConfig.cmake` in `install_dir/lib/cmake/Ginkgo` looks like:
```
set(GINKGO_INTERFACE_LINK_LIBRARIES "ginkgo_omp;ginkgo_cuda;ginkgo_reference")
set(GINKGO_INTERFACE_LINK_FLAGS "-L<ginkgo_install_dir>/lib -lginkgo -lginkgo_omp -lginkgo_cuda -lginkgo_reference"
```



That was why I was using the `GINKGO_INTERFACE_LINK_FLAGS` to get the proper [list of flags to link against in DEAL_II_PACKAGE_HANDLE](https://github.com/dealii/dealii/pull/11413/commits/cbd7e9bed5528e43326a17f667b3a1da735ca456#diff-23b767ae3a5cb11c2e3894299d5d20b9559bb0455563936d950a8da702c99e0bR34).

Ideally, the best way would be the cmake way of using the `Ginkgo::ginkgo` target and just simply using `target_link_libraries(${dealii_target} Ginkgo::ginkgo)`, but I could not find a place to do this within deal.ii.

tamiko: Using imported targets such as `Ginkgo::ginkgo` directly unfortunately doesn't solve the problem that we have to export the full transitive link interface in a meaningful way. And we have to do that uniformly for all external dependencies.

We are currently solving this issue with a custom `DEAL_II_PACKAGE_HANDLE` macro. The whole deal.II build system was designed with cmake-2.8.8 and is thus a bit rusty.  We tried to transition to imported targets once but had to roll back due to some unforeseen incompatibilities. Currently, I would very much prefer not to use imported targets (such as `Ginkgo::ginkgo` before we have figured out how to correctly integrate those.

That said, the core problem is that the CMake ecosystem has not figured out a canonical standard format for finding and setting up external dependencies: Some projects have a "project configuration", some have a `FindFEATURE.cmake` macro shipped with CMake and most do not provide anything. So we unfortunately, had to settle on the greatest commond divisor - which is to do all the dependency setup by hand (in our own `FindFEATURE.cmake` files).

Issue #8911, opened 2019-10-15
Wrong solution using Raviart-Thomas element on 3D distorted meshes


pcafrica:
Dear deal.II community,

@sdigre and I encountered an issue when computing a finite element solution to a Darcy equation using Raviart-Thomas(0) elements on 3D distorted (non-Cartesian) meshes.

The minimal steps to reproduce the issue (summed up in the [code attached](https://github.com/dealii/dealii/files/3730119/step20_3D.cpp.txt)) are inspired by the [step-20](https://www.dealii.org/current/doxygen/deal.II/step_20.html) tutorial, extended trivially to 3D geometries.

We first instantiate the 3D template version of the solver `MixedLaplaceProblem<3>`: so far everything works, providing the same 2D solution extruded across the third dimension _z_.

Now if we introduce a slight random mesh distortion, keeping the boundaries fixed, _i.e._
adding `GridTools::distort_random(0.1, triangulation, true)` after the mesh creation, then problems arise: the _L^2_ error on the velocity is about 6 times larger than the one computed on the undistorted mesh, the _z_ component of the velocity is non-zero, whereas the _x_ and _y_ components show an unphysical pattern.

The same behavior happens importing an externally generated 3D cube mesh using any of the the `GridIn` reader formats.

The 2D version of the code runs smoothly for any kind of mesh.

Trying to debug the code, we noticed that the RT0 basis function associated with a face dof of a distorted cell is not normal to the face itself as it is expected to be, which makes us suspect a bug in the mapping of the Raviart-Thomas functions from the reference cell.

Thanks in advance.
Best regards,
Pasquale and Simone

elauksap:
We have been investigating the error and we suspect that it is related to the issue
[#7970](https://github.com/dealii/dealii/issues/7970)
that has been assigned to the milestone "Release 9.2".

In particular I think that it is due to the missing implementation of `get_face_sign_change_rt` in 3D
(see https://github.com/dealii/dealii/blob/master/source/fe/fe_poly_tensor.cc#L111)

We will be thinking of a solution to that function, do you have any ideas or suggestions to start from?

Thanks.

PR #11414, opened 2020-12-23
Raviart-Thomas orientation and permutation issue

konsim83:
So, let's have Christmas a bit earlier this year... 

This is a patch to remedy the issue #7970  with elements derived from `FE_PolyTensor<dim, spacedim>` that arises on meshes that contain cells with faces that are not in standard orientation or rotated against a neighboring face. There shape functions with support in normal direction need to be permuted and re-oriented on faces shared between cells to match the H(div) normal continuity condition.

This patch fixes only Raviart-Thomas elements and leaves the others untouched while offering a way to fix also the remaining elements. 

1. The shape function permutation is implemented in the table `adjust_quad_dof_index_for_face_orientation_table` inherited from `FiniteElement<dim, spacedim>` 
2. An additional table `adjust_quad_dof_sign_for_face_orientation_table` in the class `FE_PolyTensor<dim, spacedim>` needs to be filled with flags indicating whether the orientation must be fixed or not
3. All these tables are implemented in derived classes
4. Using elements derived from will throw an exception in debug mode if they are used on meshes that suffer from the above deficiencies if these elements suffer from the sign/permutation issue and have not been fixed
5. The implemetation can be done in a similar way for elements that I did not fix yet. Note that for `FE_Nedelec<dim>` or `FE_Bernardi Raugel<dim>` one needs to think about dofs located on lines.
6. The implementation does not break existing functionality. It also does not change the interfaces and users should not notice (apart from the exceptions in 4.)
7. The implementation was tested on a new mesh `GridGenerator::orientation_test_mesh`. It consists of two neighboring cubes and depending on a number of flags passed to the constructing function in can re-orient one of the cubes such that one of the common faces needs to flipped or rotated to get a consistent mesh. This class is not overly useful from the practical point of view but it is good for debugging purposes since it offers a way to create a mini mesh with all possible deficiencies causing trouble
8. The method was tested through visual inspection with a modified version of step-20 with both essential and natural boundary conditions using a manufactured solution.
9. Tests have been added to Deal.II to make sure that the permutation/re-orientation works for Raviart-Thomas up do degree 3 (already 108 dofs per cell) on the `GridGenerator::orientation_test_mesh` in all 8 configurations of the flags `(face_orientation | face_flip | face_rotation)`
10. The patch is only for `dim=3` since this issue does not arise in 1D/2D

Hope this helps others, too. I am already using it on hyper_shells with success.

Best,
Konrad


masterleinad:
I haven't looked at this pull request at all but the CI is complaining about some oxygen issues. Would you mind having a look?
```
/home/runner/work/dealii/dealii/include/deal.II/fe/fe_poly_tensor.h:304: warning: end of comment block while expecting command </code>
/home/runner/work/dealii/dealii/include/deal.II/fe/fe_poly_tensor.h:304: warning: end of comment block while expecting command </code>
/home/runner/work/dealii/dealii/include/deal.II/grid/grid_generator.h:772: warning: argument 'triangulation' of command @param is not found in the argument list of GridGenerator::orientation_test_mesh(Triangulation< dim > &tria, const bool face_orientation, const bool face_flip, const bool face_rotation, const bool manipulate_left_cube)
/home/runner/work/dealii/dealii/include/deal.II/grid/grid_generator.h:791: warning: The following parameters of GridGenerator::orientation_test_mesh(Triangulation< dim > &tria, const bool face_orientation, const bool face_flip, const bool face_rotation, const bool manipulate_left_cube) are not documented:
  parameter 'tria'
```

elauksap:
Great job @konsim83!

I've been looking for this patch for a long time!
This probably also fixes the related issue #8911, which I started working on more than one year ago but unfortunately didn't have time to solve myself.

Thanks 😄

konsim83:
Hi @elauksap , I checked the issue with distorted grids that you mention in [issue#8911](https://github.com/dealii/dealii/issues/8911) but that does not seem to be fixed. I honestly do not have a clue where this issue comes from. I have a vague guess:

I recall that @drwells mentioned to me that these elements do only work with cartesian mappings but I do not find a switch in the code. RaviartThomas elements are mapped with mapping_raviart_thomas which actually amounts to mapping_piola (push-forward for 2-forms) which is correct, see the code [line 63](https://github.com/dealii/dealii/blob/master/source/fe/fe_raviart_thomas.cc#L63) in the FE source file and [line 470](https://github.com/dealii/dealii/blob/master/source/fe/fe_poly_tensor.cc#L470) FEPolyTensor's source.

I tried to find where that issue could come from (Is it really the mapping?) but I did not find it. If anyone has an idea where to look for I will go into that if I have enough time. But maybe that is an issue for another PR?

Cheers,
Konrad

drwells:
I don't fully understand the issue in #7970 but I believe the problem with the mapping (which is not necessarily inside the `Mapping` class) is that we do not respect arbitrary face and edge orientations. We do use the Piola transformation with some Q1 or Cartesian mapping, which is fine, but the orientations are what (I think, again, I'm not an expert) cause problems for general 3D meshes.

konsim83:
Hi @drwells that sounds worth checking. What I don't find in the code is the underlying mapping for the cells (MappingQ or MappingCartesian) of the FE. I see that mapping_piola is being applied but if this is done with a Cartesian mapping of the cell this will give wrong results. For randomly distorted meshes the mapping should of course not be Cartesian.

elauksap:
My suspect was that the issue came from the missing implementation of [`get_face_sign_change_rt` in 3D](https://github.com/dealii/dealii/blob/65148f3558038e750ad06f30b5842ebc4a2c296a/source/fe/fe_poly_tensor.cc#L117), which, if I'm not wrong, should be now fixed by this patch.

I'll have a check. Thank you, by the way!

konsim83:
> My suspect was that the issue came from the missing implementation of [`get_face_sign_change_rt` in 3D](https://github.com/dealii/dealii/blob/65148f3558038e750ad06f30b5842ebc4a2c296a/source/fe/fe_poly_tensor.cc#L117), which, if I'm not wrong, should be now fixed by this patch.
> 
> I'll have a check. Thank you, by the way!

Hi @elauksap this implementation was missing and is now fixed (it caused trouble for example on a hyper shell) but your issue (I believe) is different.

source/fe/fe_nedelec.cc
```diff
@@ -207,9 +206,30 @@ FE_Nedelec<dim>::FE_Nedelec(const unsigned int order)
...
     }
+
+  // We need to initialize the dof permuation table and the one for the sign
+  // change.
+  initialize_quad_dof_index_permutation_and_sign_change();
+
+  /*
+   * Throw an exception on cells with non-standard faces
+   */
+  this->throw_exception_on_occurrance_of_nonstandard_faces = true;
```
agrayver, line 217: For this and other elements where the implementation is missing, this option changes the default behaviour (which presently is basically "do nothing"), does not it? 

This may not be a bad thing to do since results with this element and a given non-standard mesh will likely be incorrect, but this will at least break some tests...

agrayver:
@konsim83 -- very interesting work here. Coming to this from the Nedelec element perspective: do you believe it can also be done for this element in a static manner like you did here for RT?

konsim83:
Hi @agrayver , I am not sure so far if this will break some tests. I ran most tests for FEs on my machine and nothing broke so far. Yes, it will change the default behavior of elements inheriting from 'FEPolyTensor' and should only throw an error if there are flipped faces or lines.

About your question: I think one can fix this for Nedelec elements also using the pattern I used. I am thinking about one thing, though: throwing and error on non-standard faces may not be enough for lowest order Nedelec since these only have line dofs. I do think that flipping of lines will also occur when faces are flipped/rotated but can lines be flipped without faces being affected? (For elements with line dofs there is another table that elements inherit that need to be filled with the correct values if faces/lines are not in standard orientation.) I am not familiar enough with details of triangulations here. If you want to go into that I can show you how I found the correct values using the new orientation test mesh in this PR.

@elauksap Sorry for not being more helpful with your issue. Just a quick remark in case you did not find out already. Instead of Using RaviartThomas(k)-DGQ(k) pairings you can also use BDM(k)-DGP(k-1) for k>=1. They are stable and come from a different deRahm-Complex. I tried it (just modify step-20) and they do not exhibit the issue you describe. If you solve problems in H(curl)-H(div), though, then you must use Nedlec elements of second kind (order k) with BDM(k). This version of Nedelec is unfortunately not implemented yet.

Cheers,
Konrad

agrayver:
> I am not sure so far if this will break some tests.

Hm, this is surprising: I thought we have tests for FE_Nedelec which deal with non-standard meshes just to ensure that once the issue is fixed, the test will signal it...

 > If you want to go into that I can show you how I found the correct values using the new orientation test mesh in this PR.

Yes, would be great! Let's do it elsewhere (drop me an email or open an seperate thread if appropriate).

konsim83:
@agrayver Sorry, you were right. Some tests are broken. I need to identify them and fix them as well. I already see that the divergence theorem tests for Nedelec are broken since the test is on a hyper_ball which contains non-standard faces. Dropped you an email for offline discussion of further issues.

source/fe/fe_nedelec.cc
konsim83: Yep, I am currently going over the broken tests and need to find out why exactly they failed.

konsim83:
OK, I went into some tests that failed and what I can see is that tests fail that involve RaviartThomas FEs on complicated meshes. That is expected.

I need to find out why some Nedelec tests fail as well since I did not change anything in their actual implementation besides the `Assert(!this->throw_exception_on_occurrance_of_nonstandard_faces ...)` I may have a clue why: The author of the original code changed the sign of the Nedelec shape function if the line is flipped. I removed that since this is by far not enough. A face may not be flipped but a line may be. Guess I need to catch that as well with another flag.

include/deal.II/grid/grid_generator.h
```diff
@@ -768,6 +768,36 @@ namespace GridGenerator
...
+   * @param[in] face_flip <code>true</code> if the face is rotated by +180
+   * degrees
+   * @param[in] face_rotation <code>true</code> if the face is rotated
+   * (additionally) by +90 degrees
+   * @param[in] manipulate_left_cube <code>true</code> if the left cube is
+   * re-ordered
+   */
+  template <int dim>
+  void
+  orientation_test_mesh(Triangulation<dim> &tria,
```
agrayver, line 795: How about calling it "non_standard_orientation_mesh" or simply "non_standard_mesh"?

agrayver:
> The author of the original code changed the sign of the Nedelec shape function if the line is flipped. I removed that since this is by far not enough.

This can actually help in several cases with the lowest-order elements.

agrayver:
@konsim83 I have a general suggestion here. This PR currently goes far beyond only RT elements, it changes the default behaviour of five elements and fixes issue in one. I think you're mixing too many things here... 

Would it make sense to split these two things. Just work on RT in this PR. Then you can propose a separate one where an exception is added, but I am not convinced this is so good idea, especially in case of FE_Nedelec...

source/fe/fe_poly_tensor.cc
```diff
@@ -446,56 +423,54 @@ FE_PolyTensor<dim, spacedim>::fill_fe_values(
...
+                    make_array_view(fe_data.shape_values, dof_index),
+                    mapping_piola,
+                    mapping_internal,
+                    make_array_view(fe_data.transformed_shape_values));
                   for (unsigned int k = 0; k < n_q_points; ++k)
                     for (unsigned int d = 0; d < dim; ++d)
                       output_data.shape_values(first + d, k) =
-                        fe_data.sign_change[i] *
-                        fe_data.transformed_shape_values[k][d];
+                        dof_sign * fe_data.transformed_shape_values[k][d];
```
agrayver, line 458: I would not touch Nedelec elements here. I think the issue with non-standard meshes should either be completely resolved in a separate PR or we do not touch them at all.

source/fe/fe_poly_tensor.cc
```diff
@@ -446,56 +423,54 @@ FE_PolyTensor<dim, spacedim>::fill_fe_values(
...
+                    mapping_covariant,
+                    mapping_internal,
+                    make_array_view(fe_data.transformed_shape_values));
 
                   for (unsigned int k = 0; k < n_q_points; ++k)
                     for (unsigned int d = 0; d < dim; ++d)
                       output_data.shape_values(first + d, k) =
-                        fe_data.sign_change[i] *
-                        fe_data.transformed_shape_values[k][d];
+                        dof_sign * fe_data.transformed_shape_values[k][d];
```
agrayver, line 473: Same as above.

source/fe/fe_poly_tensor.cc
konsim83: Nedelec is not touched here at all. All dof signes stay at +1 except for elements that re-implement a certain table.

konsim83:
Hi @agrayver Thanks for your comments.

> > The author of the original code changed the sign of the Nedelec shape function if the line is flipped. I removed that since this is by far not enough.
> 
> This can actually help in several cases with the lowest-order elements.

Is this really the case? Implementing a sign flip function for Nedelec for lowest order in such cases makes only sense if the lines are not permuted upon face flips or orientation changes. If this is the case I can do that also for Nedelec of lowest order. This should not be too much work.


> @konsim83 I have a general suggestion here. This PR currently goes far beyond only RT elements, it changes the default behaviour of five elements and fixes issue in one. I think you're mixing too many things here...
> 
> Would it make sense to split these two things. Just work on RT in this PR. Then you can propose a separate one where an exception is added, but I am not convinced this is so good idea, especially in case of FE_Nedelec...

I see. The idea of the PR was to change only RaviartThomas through changing the interface a bit but not to break standard behavior of the other elements which I only did for Nedelec since I removed the sign change upon line flips (this is also what the tests reflect). It would be harder to split that up and eventually a carpet of patches. My suggestion would be that I go into the Nedelec case and fix it for lowest order once I get confirmed that lines do not permute upon face changes and to fix the RaviartThomas tests since these actually confirmed results that were wrong before the PR.

I understand that this is not a minor issue but let's bite us a way through it. Does that sound reasonable? That issue bugged me for some time in three elements (I am mostly simulating systems coupling the buggy spaces)

agrayver:
@konsim83 I agree that the design and interface changes you made are for better. What I think now is that we should indeed discuss this PR and subsequent for externally (check email). With your present expertise and understanding, we may finally have a chance to fix this very long-standing orientation issue in a more general way.

konsim83:
@agrayver I was investigating the Nedelc element a bit and I saw that even if line orientations are not changed there is permutation of line dofs upon non-standard faces (bad news). I will try to find a system behind that. Unfortunately that means that the sign fix that was originally implemented does not work even for lowest order Nedelec.

konsim83:
@agrayver Do you know who is familiar with the mesh classes? I have a small question on how they are built if faces do not match in their orientation etc. There the lines numbering must be consistent. Maybe there is a chance to use that for the line dofs of Nedelec.

agrayver:
@konsim83 your findings align well with what others (including me) have concluded over the years. This was the nature of my original question whether one do it in a static way for Nedelec as well. I was hoping that at least the lowest order case should be possible. 

> Do you know who is familiar with the mesh classes?

I would think that @bangerth could help. He is likely implemented some of this initially, but additionally there is this recent [paper ](https://dl.acm.org/doi/abs/10.1145/3061708)on the topic.

@peterrum was going through this a lot recently, but not sure he has touched the orientation part much...

peterrum:
@konsim83 @agrayver I am happy to help regarding the internal details of triangulation and orientations in general. But let's do it "offline" in "person" so that we get to the detail you are interested in more quickly.

konsim83:
The renumbering of dofs on permuted lines seems to work well for the FE_Q class (the necessary tables are implemented in FE_Q_Base). I am thinking maybe one can proceed for line dofs similarly for all Nedelec elements. The quad dofs are a different issue and can possibly be taken care of similarly to the RaviartThomas quad dofs. I just don't find the places in the library where this is taken care of.

konsim83:
Hi @agrayver and @peterrum, 
Sorry for the delay but I got snowed under by work.Here the modified pull request limiting to the bugfix in RaviartThomas. Some tests will still be broken and I need to figure out which ones to go into. As I said some wrong behavior of RaviartThomas was confiremed in the tests. Another PR will follow soon.
Best,
Konrad

masterleinad:
> Edit: Does anyone remember how I can change the clang-format version on the command line? I did it in the past but I do not find it any more. :-/

You can just do `make indent` instead. Other than that, define DEAL_II_CLANG_FORMAT to the one you want to use.

konsim83:
> You can just do `make indent` instead. Other than that, define DEAL_II_CLANG_FORMAT to the one you want to use.
@masterleinad Thanks, that is what I was looking for. I added a little hint in the Deal.II coding conventions since I was not looking for this the first time.

konsim83:
@agrayver @peterrum Sorry for all the commits. I will squash everthing later.

konsim83:
The mesh and new conformity tests are in PR #11644

agrayver:
What happened with these fe_values_view_03/fe_values_view_04 tests output that you're adding/removing half a million of lines?

konsim83:
> What happened with these fe_values_view_03/fe_values_view_04 tests output that you're adding/removing half a million of lines? 

Some test cases confirmed wrong behavior of RaviartThomas in 3d. The 2d output was ok but 3d needed to be replaced. Honestly, I found these two tests less helpful since the output is not really understandable.

masterleinad:
Can you please rebase on top of `master` to get rid of all commits that are already merged?

konsim83:
> Can you please rebase on top of `master` to get rid of all commits that are already merged?

I rebased and squashed. But investigating the the orientation issue for three months now I must say that this issue is by far not done. In an upcoming pull request I will add more tests to investigate that issue.

source/fe/fe_poly_tensor.cc
```diff
@@ -546,6 +654,7 @@ FE_PolyTensor<dim, spacedim>::fill_fe_values(
                   for (unsigned int k = 0; k < n_q_points; ++k)
                     for (unsigned int d = 0; d < dim; ++d)
                       output_data.shape_gradients[first + d][k] =
+                        fe_data.dof_sign_change[dof_index] * dof_sign *
```
bangerth, line 657: I think it would be easier to read if you initialized `dof_sign` further up with `fe_data.dof_sign_change[dof_index]` so that here you only have to multiply with one of these variables.

source/fe/fe_poly_tensor.cc
```diff
@@ -572,6 +681,7 @@ FE_PolyTensor<dim, spacedim>::fill_fe_values(
...
```
bangerth, line 684: actually, in all of these places :-)

bangerth: By the way, was this here a bug before, since we weren't multiplying with `fe_data.dof_sign_change[dof_index]` as we do in all of the other places where you now multiply with both numbers?

source/fe/fe_poly_tensor.cc
```diff
@@ -981,32 +1094,76 @@ FE_PolyTensor<dim, spacedim>::fill_fe_face_values(
...
+      /*
+       * This assumes that the dofs are ordered by first vertices, lines, quads
+       * and volume dofs. Note that in 2D this always gives false.
+       */
+      const bool is_quad_dof =
+        (dim == 2 ? false :
+                    (first_quad_index <= dof_index) &&
+                      (dof_index < first_quad_index + n_quad_dofs));
+
+      unsigned int i        = dof_index;
```
bangerth, line 1138: Just make the same change as above where you replace `i` by `dof_index` (which helps with readability).

source/fe/fe_poly_tensor.cc
```diff
@@ -981,32 +1094,76 @@ FE_PolyTensor<dim, spacedim>::fill_fe_face_values(
...
+      double       dof_sign = 1.0;
```
bangerth, line 1139: Pretty much all the same comments as I had above for `fill_fe_values()`.

source/fe/fe_poly_tensor.cc
bangerth: On second reading, I think you've already gotten rid of `i`. So just remove the variable altogether.

source/fe/fe_poly_tensor.cc
```diff
@@ -1608,36 +1776,78 @@ FE_PolyTensor<dim, spacedim>::fill_fe_subface_values(
...
```
bangerth, line 1823: and again all of the same comments as above

source/fe/fe_raviart_thomas.cc
```diff
@@ -248,6 +248,173 @@ FE_RaviartThomas<dim>::initialize_support_points(const unsigned int deg)
...
+template <int dim>
+void
+FE_RaviartThomas<dim>::initialize_quad_dof_index_permutation_and_sign_change()
+{
+  // For 1D do nothing.
+  //
+  // TODO: For 2D we simply keep the legacy behavior for now. This should be
+  // changed in the future and can be taken care of by similar means as the 3D
+  // case below. The legacy behavior can be found in fe_poly_tensor.cc in the
+  // function internal::FE_PolyTensor::get_dof_sign_change_h_div(...)
```
bangerth, line 260: Ah, I see. Yes, I agree that this should be done, but it can wait for a later patch.

source/fe/fe_raviart_thomas.cc
```diff
@@ -248,6 +248,173 @@ FE_RaviartThomas<dim>::initialize_support_points(const unsigned int deg)
...
+  if (dim < 3)
+    return;
+
+  // TODO: the implementation makes the assumption that all faces have the
+  // same number of dofs
+  AssertDimension(this->n_unique_faces(), 1);
+  const unsigned int face_no = 0;
```
bangerth, line 267: Just replace it by zero everywhere below :-)

source/fe/fe_raviart_thomas.cc
```diff
@@ -248,6 +248,173 @@ FE_RaviartThomas<dim>::initialize_support_points(const unsigned int deg)
...
+  // |  +  +  -  |
+  // |___________|
+  //
+
+  for (unsigned int local_face_dof = 0;
+       local_face_dof < this->n_dofs_per_quad(face_no);
+       ++local_face_dof)
+    {
+      // Row and column
+      unsigned int i = local_face_dof % n, j = local_face_dof / n;
```
bangerth, line 333:
```suggestion
      const unsigned int i = local_face_dof % n;
      const unsigned int j = local_face_dof / n;
```

source/fe/fe_raviart_thomas.cc
```diff
@@ -248,6 +248,173 @@ FE_RaviartThomas<dim>::initialize_support_points(const unsigned int deg)
...
+
+      // We have 8 cases that are all treated the same way. Note that the
+      // corresponding case to case_no is just its binary representation.
+      // The above example of (false | true | true) would be case_no=3
+      for (unsigned int case_no = 0; case_no < 8; ++case_no)
+        {
+          // Get the binary representation of the case
+          const bool face_orientation = (((case_no / 4) % 2) == 1);
```
bangerth, line 341: `Utilities::get_bit(case_no, 2)` should do this for you :-)

source/fe/fe_raviart_thomas.cc
```diff
@@ -248,6 +248,173 @@ FE_RaviartThomas<dim>::initialize_support_points(const unsigned int deg)
...
+            }
+          else
+            {
+              // Offset is zero
+              this->adjust_quad_dof_index_for_face_orientation_table[face_no](
+                local_face_dof, case_no) = 0;
+            } // if face needs dof permutation
+
+          // Get new local face_dof by adding offset
+          unsigned int new_local_face_dof =
```
bangerth, line 361:
```suggestion
          const unsigned int new_local_face_dof =
```

source/fe/fe_raviart_thomas.cc
```diff
@@ -248,6 +248,173 @@ FE_RaviartThomas<dim>::initialize_support_points(const unsigned int deg)
...
+            local_face_dof +
+            this->adjust_quad_dof_index_for_face_orientation_table[face_no](
+              local_face_dof, case_no);
+          // compute new row and column index
+          i = new_local_face_dof % n;
+          j = new_local_face_dof / n;
```
bangerth, line 367: I think you're overwriting what you already computed before here. If you don't need the values of `i` and `j` previously computed, then just move the variable declarations to here.

bangerth (commented):
I am sorry for having taken such a long time to look at this.

I think you're on the right track with the two tables you are implementing, but I think the design of how they are filled could be improved. In essence, if I read the patch correctly, you currently have a virtual function `initialize_quad_dof_index_permutation_and_sign_change()` that is called at some point and that is meant to fill certain tables. But we always need these tables, and I think an easier to understand the design is to deal with these tables like with many other tables stored in the base class: they are simply initialized in the constructor of some derived class, and how they choose to do that is of no great concern to anyone. I think that the constructor of `FE_Nedelec` is a good example to look at: It fills all sorts of tables, and the one we're talking about here would just be another example.

konsim83:
Dear all, I will be working on the comments the upcoming weekend, sorry for the delay (I am working in two jobs these days).
Konrad

bangerth:
We appreciate all of the work you're doing on this, on whatever timeline you find time and interest to use!

include/deal.II/fe/fe_abf.h
```diff
@@ -239,6 +239,15 @@ class FE_ABF : public FE_PolyTensor<dim>
...
 
+  /**
+   * Initialize the permutation pattern and the pattern of sign change.
+   *
+   * @note This function is not fully filled with the correct implementation
+   * yet. It needs to be consistently implemented in a future release to work
+   * on meshes that contain cells with flipped faces.
+   */
+  virtual void
+  initialize_quad_dof_index_permutation_and_sign_change() override;
```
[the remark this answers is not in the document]
konsim83: Done.

include/deal.II/fe/fe_dg_vector.h
```diff
@@ -157,6 +167,17 @@ class FE_DGNedelec : public FE_DGVector<PolynomialsNedelec<dim>, dim, spacedim>
...
+private:
+  /**
+   * Initialize the permutation pattern and the pattern of sign change.
+   *
+   * @note This function is not fully filled with the correct implementation
+   * yet. It needs to be consistently implemented in a future release to work
+   * on meshes that contain cells with flipped faces.
+   */
+  virtual void
+  initialize_quad_dof_index_permutation_and_sign_change() override;
```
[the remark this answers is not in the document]
konsim83: Yes, the sign change makes no sense there. Done.

include/deal.II/fe/fe_dg_vector.templates.h
```diff
@@ -98,6 +98,23 @@ FE_DGVector<PolynomialType, dim, spacedim>::has_support_on_face(
...
+template <class PolynomialType, int dim, int spacedim>
+void
+FE_DGVector<PolynomialType, dim, spacedim>::
+  initialize_quad_dof_index_permutation_and_sign_change()
+{
+  AssertThrow(
+    false,
+    ExcMessage(
+      "This method must be implemented in any class derived "
+      "form FE_PolyTensor that is final. This method is specific to the (vector) "
```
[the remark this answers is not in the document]
konsim83: Done.

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -224,6 +224,122 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
...
+   * need to fill it with the correct values in case of non-standard, flipped
+   * (rotated by +180 degrees) or rotated (rotated by +90 degrees) faces. These
+   * are given in the form three flags (face_orientation, face_flip,
+   * face_rotation), see the documentation in GeometryInfo<dim> and
+   * this @ref GlossFaceOrientation "glossary entry on face orientation".
+   *
+   * <h3>Example: Raviart-Thomas Elements of order 2 (tensor polynomial degree
+   * 4)</h3>
+   *
+   * The dofs on a face are connected to a <code>n</code>-by-<code>n</code>
```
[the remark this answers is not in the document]
konsim83: Done.

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -224,6 +224,122 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
...
+   * |  0  1  2  |
+   * |___________|
+   *@endverbatim
+   *
+   * We have for a local <code>face_dof_index=i+n*j</code> with index
+   * <code>i</code> in x-direction and index <code>j</code> in y-direction
+   * running from 0 to <code>n-1</code>.  To extract <code>i</code> and
+   * <code>j</code> we can use <code>i = face_dof_index % n</code> and <code>j =
+   * dof_index / n</code> (integer division). The indices <code>i</code> and
+   * <code>j</code> can then be used to compute the offset.
```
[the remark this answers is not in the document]
konsim83: This may have been a bit confusing. The comment underlines how this can be achieved for the RT-element but not how to do that for other elements. For other H(div) elements such as BDM such a technique may not work. The only message here is: "There is a permutation and a sign change. How to find it depends on the element." 

I rephrased the documentation in the base class and now it hopefully is clearer to the reader.

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -224,6 +224,122 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
...
+   * |  +  +  -  |
+   * |           |
+   * |  +  +  -  |
+   * |           |
+   * |  +  +  -  |
+   * |___________|
+   * @endverbatim
+   *
+   * @note This function is not a pure virtual function but still must be implemented
+   * in finite elements classes derived from FE_PolyTensor since the permutation
```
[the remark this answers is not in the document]
konsim83: Done

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -224,6 +224,122 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
   bool
   single_mapping_kind() const;
 
+  /**
+   * The vector of tables
+   * <code>adjust_quad_dof_index_for_face_orientation_table</code> contains
```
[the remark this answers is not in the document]
konsim83: I rephrased it and moved to raviart_thomas.h (see my other comment above)

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -224,6 +224,122 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
...
+                                            const bool         face_flip,
+                                            const bool face_rotation) const;
+
+  /**
+   * For faces with non-standard face_orientation in 3D, the dofs on faces
+   * (quads) have to be permuted in order to be combined with the correct
+   * shape functions. For elements with shape that have an orientation relative
+   * to faces finding the permutations is not enough. We possibly need to change
+   * the sign of the shape function on faces depending on the combination of the
+   * three bool flags face_orientation, face_flip and face_rotation.
```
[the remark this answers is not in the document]
konsim83: I removed the function from the base class. This hint is nonetheless useful and can be found in adjust_quad . An example can be found the documentation in raviart_thomas.h

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -224,6 +224,122 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
...
+   *
+   * The constructor of this class fills this table with ones, i.e.,
```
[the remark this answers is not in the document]
konsim83: Thank you. The values are actually 'false', i.e., no sign change. Corrected

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -224,6 +224,122 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
...
+   * no sign change at all. Derived finite element classes have to
+   * fill this Table with the correct values, see the documentation in
+   * GeometryInfo<dim> and
+   * this @ref GlossFaceOrientation "glossary entry on face orientation".
+   */
+  std::vector<Table<2, bool>> adjust_quad_dof_sign_for_face_orientation_table;
```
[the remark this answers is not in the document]
konsim83: This may be true. I implemented it in the same terms as the table <code>adjust_quad_dof_index_for_face_orientation_table</code> since this way one does not need to rethink the structure of the implementation for <code>adjust_quad_dof_sign_for_face_orientation_table</code> too much. This may also help other people to fix other elements by having a look at the implementation of `RaviartThomas<dim>`. Besides, I understand your concern but is the overhead really significant here? In most codes the finite element is constructed once at the beginning and the code is not called repeatedly. If you agree I would like to leave it as is.

include/deal.II/fe/fe_poly_tensor.h
```diff
@@ -460,7 +576,7 @@ class FE_PolyTensor : public FiniteElement<dim, spacedim>
     /**
      * Scratch arrays for intermediate computations
      */
-    mutable std::vector<double>              sign_change;
+    mutable std::vector<double>              dof_sign_change;
```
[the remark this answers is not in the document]
konsim83: Thanks you, it seemed more intuitive to me.

source/fe/fe_bdm.cc
```diff
@@ -99,9 +103,25 @@ FE_BDM<dim>::FE_BDM(const unsigned int deg)
           this->interface_constraints(target_row, j) = face_embeddings[d](i, j);
         ++target_row;
       }
+
+  // We need to initialize the dof permuation table and the one for the sign
+  // change.
+  initialize_quad_dof_index_permutation_and_sign_change();
```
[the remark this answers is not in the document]
konsim83: This is true and I removed it from the base. Please see my other comments.

source/fe/fe_dg_vector.cc
```diff
@@ -29,7 +29,21 @@ DEAL_II_NAMESPACE_OPEN
 template <int dim, int spacedim>
 FE_DGNedelec<dim, spacedim>::FE_DGNedelec(const unsigned int p)
   : FE_DGVector<PolynomialsNedelec<dim>, dim, spacedim>(p, {mapping_nedelec})
-{}
+{
+  // We need to initialize the dof permuation table and the one for the sign
+  // change.
+  initialize_quad_dof_index_permutation_and_sign_change();
```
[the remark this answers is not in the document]
konsim83: removed.

source/fe/fe_bernardi_raugel.cc
```diff
@@ -77,6 +85,18 @@ FE_BernardiRaugel<dim>::get_name() const
 }
 
 
+template <int dim>
+void
+FE_BernardiRaugel<dim>::initialize_quad_dof_index_permutation_and_sign_change()
+{
+  // for 1D and 2D, do nothing
+  if (dim < 3)
+    return;
```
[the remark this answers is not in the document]
konsim83: Done

source/fe/fe_poly_tensor.cc
```diff
@@ -38,32 +38,25 @@ namespace internal
...
-       * implements an algorithm that determines those DoFs that need this
-       * sign change for a given cell.
-       */
       template <int spacedim>
       void
-      get_face_sign_change_rt(const dealii::Triangulation<1>::cell_iterator &,
-                              const FiniteElement<1, spacedim> &,
-                              const std::vector<MappingKind> &,
-                              std::vector<double> &)
+      get_dof_sign_change_h_div(const dealii::Triangulation<1>::cell_iterator &,
```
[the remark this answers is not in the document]
konsim83: Done

source/fe/fe_poly_tensor.cc
```diff
@@ -83,11 +76,11 @@ namespace internal
                 const unsigned int nn = cell->neighbor_face_no(f);
 
                 if (nn < GeometryInfo<dim>::faces_per_cell / 2)
-                  for (unsigned int j = 0; j < fe.n_dofs_per_face(f); ++j)
+                  for (unsigned int j = 0; j < fe.dofs_per_face; ++j)
```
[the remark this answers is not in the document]
konsim83: Yes, this function does not necessarily work on other element types and I reverted the change.

source/fe/fe_poly_tensor.cc
```diff
@@ -188,6 +179,23 @@ FE_PolyTensor<dim, spacedim>::FE_PolyTensor(
...
+
+      for (unsigned int f = 0; f < this->n_unique_quads(); ++f)
+        {
+          adjust_quad_dof_sign_for_face_orientation_table[f] = Table<2, bool>(
+            this->n_dofs_per_quad(f),
+            internal::ReferenceCell::get_cell(this->reference_cell())
+                  .face_reference_cell(f) == ReferenceCell::Quadrilateral ?
+              8 :
+              6);
+          adjust_quad_dof_sign_for_face_orientation_table[f].fill(false);
```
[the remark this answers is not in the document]
konsim83: True. The documentation is now corrected.

source/fe/fe_poly_tensor.cc
```diff
@@ -188,6 +179,23 @@ FE_PolyTensor<dim, spacedim>::FE_PolyTensor(
...
+        this->n_unique_quads());
+
+      for (unsigned int f = 0; f < this->n_unique_quads(); ++f)
+        {
+          adjust_quad_dof_sign_for_face_orientation_table[f] = Table<2, bool>(
+            this->n_dofs_per_quad(f),
+            internal::ReferenceCell::get_cell(this->reference_cell())
+                  .face_reference_cell(f) == ReferenceCell::Quadrilateral ?
+              8 :
+              6);
```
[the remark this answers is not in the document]
konsim83: For quadrilaterals we have 8 different face orientations for other element types such as tetraeder (I guess) 6. The same logig is implemented in lines 159+ in fe.cc

source/fe/fe_poly_tensor.cc
```diff
@@ -399,40 +463,82 @@ FE_PolyTensor<dim, spacedim>::fill_fe_values(
...
+  internal::FE_PolyTensor::get_dof_sign_change_nedelec(cell,
+                                                       *this,
+                                                       this->mapping_kind,
+                                                       fe_data.dof_sign_change);
+
+  // TODO: This, similarly to the Nedelec case, is just a legacy function in 2D
+  // and affects only face_dofs of H(div) conformal FEs. It does nothing in 1D.
+  // Also nothing in 3D since we take care of it by using the
+  // adjust_quad_dof_sign_for_face_orientation_table.
+  internal::FE_PolyTensor::get_dof_sign_change_h_div(cell,
```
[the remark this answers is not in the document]
konsim83: That would certainly be a good thing to do - in particular with other element types in mind.

source/fe/fe_poly_tensor.cc
```diff
@@ -399,40 +463,82 @@ FE_PolyTensor<dim, spacedim>::fill_fe_values(
...
-                                                        this->mapping_kind,
-                                                        fe_data.sign_change);
+      double dof_sign = 1.0;
+      if (is_quad_dof)
+        {
+          /*
+           * Find the face belonging to this dof_index. This is integer
+           * division.
+           */
+          unsigned int face_index_from_dof_index =
```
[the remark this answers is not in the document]
konsim83: Done.

source/fe/fe_poly_tensor.cc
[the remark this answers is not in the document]
konsim83: Btw, what is the reason for the public attribute `dofs_per_face` in ` FiniteElementData<dim>` if the number of dofs on a face is not the same for all faces?

source/fe/fe_poly_tensor.cc
```diff
@@ -399,40 +463,82 @@ FE_PolyTensor<dim, spacedim>::fill_fe_values(
...
+            (dof_index - first_quad_index) / (n_dofs_per_quad);
+
+          unsigned int local_quad_dof_index = dof_index % n_dofs_per_quad;
```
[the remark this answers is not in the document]
konsim83: done

source/fe/fe_poly_tensor.cc
konsim83: done

source/fe/fe_poly_tensor.cc
konsim83: I think it was a bug. It was only handled in case of mapping_piola and mapping_nedelec but should be handled in all cases in which 2-forms and 1-forms are mapped

source/fe/fe_poly_tensor.cc
konsim83: Oh, I forgot to remove that. Done.

source/fe/fe_poly_tensor.cc
konsim83: done

source/fe/fe_poly_tensor.cc
konsim83: done

source/fe/fe_raviart_thomas.cc
konsim83: Should I open an issue to document the TODOs for later patches?

source/fe/fe_raviart_thomas.cc
konsim83: Should I really remove it? It is a small issue that someone probably kept to have not too much work or at least to identify places to change things once other element types enter the game. It is used also in other places in this file. See for example `initialize_restriction()`

source/fe/fe_raviart_thomas.cc
konsim83: done

source/fe/fe_raviart_thomas.cc
konsim83: done

source/fe/fe_raviart_thomas.cc
konsim83: The variables are overwritten in each iteration and used. See above comment. The style is maybe not good imo and I will think about a better design. But the tests indicate that it works.

konsim83:
@bangerth I apologize for the long time it took to respond to the comments. I was pretty busy and in addition I was chasing another bug that I saw after the I added a test in the spirit of #11644. Now the issue is fixed as indicated by the test: all normal jumps across faces are machine zero.  I would not have seen that without this test.

After having fixed this some other tests that involve RT elements may be broken and I need to see why. Let's see.

Many thanks for your review! :-)

doc/doxygen/headers/coding_conventions.h
```diff
@@ -56,6 +56,14 @@ pre-commit hook. One way to do so, is to copy
...
 <code>\${SOURCE_DIR}/.git/hooks/pre-commit</code> and make sure it is
 executable.
+
+On some systems the command <code>clang-format</code> by default points to a
+newer version than v6.0 which is currently required by Deal.II. A simple fix
+for that issue is to point to the correct version with
+@code
+  make DEAL_II_CLANG_FORMAT=clang-format-6.0 indent
+@endcode
+provided <code>clang-format-6.0</code> is installed.
```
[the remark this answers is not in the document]
konsim83: Will do. Removed it.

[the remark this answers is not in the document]
konsim83: Will do. Removed it.

source/fe/fe_raviart_thomas.cc
konsim83: Thanks, changed that :-)

konsim83:
OK, I need to figure out why the Nedelec tests are broken. Will be able to do so only next weekend, sorry for the delays.

konsim83:
@bangerth I fixed all tests and it seems like the CI is passing. What I don't understand is the other integration tests. Remark: I did not rebase on the current master yet. Could this be the source?

I added an orientation test using the new test mesh. This test confirms that `FE_RaviartThomas<dim>` is now H(div)-conforming. I removed a less informative test that I added (in this MR) before.

masterleinad:
> @bangerth I fixed all tests and it seems like the CI is passing. What I don't understand is the other integration tests. Remark: I did not rebase on the current master yet. Could this be the source?

Yes, that's possible. The CI merges with the current master before testing.

konsim83:
> > @bangerth I fixed all tests and it seems like the CI is passing. What I don't understand is the other integration tests. Remark: I did not rebase on the current master yet. Could this be the source?
> 
> Yes, that's possible. The CI merges with the current master before testing.

Thank you @masterleinad, I rebased and I hope this issue is done now. Let's see.

include/deal.II/fe/fe_poly_tensor.h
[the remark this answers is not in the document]
bangerth: If you think this is the right way to do it, then let's leave it!

source/fe/fe_raviart_thomas.cc
bangerth: As a general rule, small patches are easiest to deal with. You already see how long it took to get this one finished and reviewed, so yes: If you can break a patch into smaller, self-contained pieces, that is always preferrable.

Maybe as a remark on the side: This patch also contains a number of pieces where you rename variables, or just clean up little stuff. The way I operate is that I commit these smaller things separately into the branch I currently work on, and then create a number of branches based on `master` onto which I cherry-pick these clean-up patches and submit as separate pull requests. It is not difficult to get these sorts of things merged relatively quickly, and then the large patch you're working on becomes smaller and easier to get reviewed.

source/fe/fe_raviart_thomas.cc
bangerth: It's not terribly important, but as a general rule, move declarations of variables as far down in the function as you can, into the innermost scope possible. This way it becomes easier to reason about questions like "Where does this variable live?" or "Does it carry state from one loop iteration to the next?"

bangerth (approved):

bangerth:
Most excellent! Thanks a lot for the patience you had with this patch, and the excellent work getting an issue fixed that has long been a problem!

I hope we get to see more patches from you in the future, and hopefully can provide a quicker turn-around with the reviews!

konsim83:
> Most excellent! Thanks a lot for the patience you had with this patch, and the excellent work getting an issue fixed that has long been a problem!
> 
> I hope we get to see more patches from you in the future, and hopefully can provide a quicker turn-around with the reviews!

I thank you all, @bangerth @masterleinad @agrayver @peterrum, for your patience and support. :-)

PR #7371, opened 2018-10-23
Update step 27

drwells:
I happened to look at the introduction discussion and I realized it was somewhat out of date with respect to how we assemble `SparsityPattern`s these days (and also what is done in step-2 and step-6). 

It looks like we never updated the pictures after #2603; the ones that I can see on the web site were last modified in 2013. I generated 19 new plots for this example in higher resolution (and SVG when possible) Here is a heat map of the smoothness:

step-27-smoothness-5

and a plot of the FE degree:

step-27-cell-degree-5

The fixes in #2603 and lowering the linear solver tolerance make the plots symmetric, which is nice :)

While working on this I discovered that our multi-column table replacement is a bit buggy: I added a fix for that which also fixes, e.g., https://dealii.org/9.0.0/doxygen/deal.II/step_10.html#Results; the previous version used to work on older versions of Firefox but no longer does, and I don't think we ever got margins to work correctly in chrome. The details are in a commit message: in summary, I just hardcoded the margins to something reasonable instead of the (current and broken) percentage.

PR #11418, opened 2020-12-24
step-27: Updated to new interface.

marcfehling:
Successor of #7371. Part of #7515.

This PR simplifies the tutorial code in such a way that it uses the functions in the `SmoothnessEstimator` namespace. Further, it adds possible extensions so users get motivated to try different decision strategies as well as to parallelize this tutorial.

The domain is kept untouched in this rework, but the way how it's generated has changed: instead of re-using the function from `step-14`, now a dimension-independent formulation does the work.

Previously, this tutorial only performed p-refinement. In addition, this PR now also enables p-coarsening as well.

Further, the hp-decision strategy is still the decay of Fourier coefficients one, since the current introduction for this tutorial is very extensive for this particular strategy. I feel it might confuse long-term users to find a different one here (e.g. the one using Legendre coefficients). How do you feel about this?

I still need to work on the results section to update the application output and pictures, which is why I kept this as a draft.

### DIFF
--- a/examples/step-27/doc/results.dox
+++ b/examples/step-27/doc/results.dox
@@ -3,7 +3,7 @@
 In this section, we discuss a few results produced from running the
 current tutorial program. More results, in particular the extension to
 3d calculations and determining how much compute time the individual
-components of the program take, are given in the @ref hp_paper .
+components of the program take, are given in the @ref hp_paper "hp-paper".
 
 When run, this is what the program produces:
 
@@ -41,12 +41,12 @@ The first thing we learn from this is that the number of constrained degrees
 of freedom is on the order of 20-25% of the total number of degrees of
 freedom, at least on the later grids when we have elements of relatively
 high order (in 3d, the fraction of constrained degrees of freedom can be up
-to 30%). This is, in fact, on the same order of magnitude as for non-$hp$
-discretizations. For example, in the last step of the step-6
+to 30%). This is, in fact, on the same order of magnitude as for
+non-$hp$-discretizations. For example, in the last step of the step-6
 program, we have 18353 degrees of freedom, 4432 of which are
 constrained. The difference is that in the latter program, each constrained
 hanging node is constrained against only the two adjacent degrees of
-freedom, whereas in the $hp$ case, constrained nodes are constrained against
+freedom, whereas in the $hp$-case, constrained nodes are constrained against
 many more degrees of freedom. Note also that the current program also
 includes nodes subject to Dirichlet boundary conditions in the list of
 constraints. In cycle 0, all the constraints are actually because of
@@ -141,9 +141,9 @@ grey corresponds to degree two and pink corresponds to degree seven:
 While this is certainly not a perfect arrangement, it does make some sense: we
 use low order elements close to boundaries and corners where regularity is
 low. On the other hand, higher order elements are used where (i) the error was
-at one point fairly large, i.e. mainly in the general area around the corner
+at one point fairly large, i.e., mainly in the general area around the corner
 singularities and in the top right corner where the solution is large, and
-(ii) where the solution is smooth, i.e. far away from the boundary.
+(ii) where the solution is smooth, i.e., far away from the boundary.
 
 This arrangement of polynomial degrees of course follows from our smoothness
 estimator. Here is the estimated smoothness of the solution, with darker colors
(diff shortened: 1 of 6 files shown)


examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
-      }
+    Triangulation<dim> cube;
+    GridGenerator::subdivided_hyper_cube(cube, 4, -1., 1.);
+
+    std::set<typename Triangulation<dim>::active_cell_iterator> cells_to_remove;
+    for (const auto &cell : cube.active_cell_iterators())
+      for (unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell; ++v)
+        for (unsigned int d = 0; d < dim; ++d)
+          if (std::abs(cell->vertex(v)[d]) < .5)
+            cells_to_remove.insert(cell);
```
marcfehling, line 479: We might use `std::copy_if` here instead, but I do not know if this would confuse users who are new to C++.

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+adaptation type from which a couple are already implemented in deal.II:
+<ul>
+  <li><i>Fourier coefficient decay:</i> This is the strategy currently
+  implemented in this tutorial. For more information on this strategy, see
+  the general documentation of the SmoothnessEstimator::Fourier namespace.</li>
+
+  <li><i>Legendre coefficient decay:</i> This strategy is quite similar
+  to the current one, but uses Legendre series expansion rather than the
+  Fourier one: instead of sinusoids as basis functions, this strategy uses
+  Legendre polynomials. The transition from the Fourier strategy to
```
bangerth, line 235:
```suggestion
  Legendre polynomials. Of course, since we approximate the solution using a finite-dimensional
  polynomial on each cell, the expansion of the solution in Legendre polynomials is also finite
  and, consequently, when we talk about the "decay" of this expansion, we can only consider the
  finitely many nonzero coefficients of this expansion, rather than think about it in asymptotic terms.
  But, if we have enough of these coefficients, we can certainly think of the decay of these
  coefficients as characteristic of the decay of the coefficients of the exact solution (which is, in
  general, not polynomial and so will have an infinite Legendre expansion), and considering the
  coefficients we have should reveal something about the properties of the exact solution.
  
  The transition from the Fourier strategy to
```

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+  to be part of the proper namespaces FESeries::Legendre and
+  SmoothnessEstimator::Legendre. For the theoretical background of this
+  strategy, consult the general documentation of the
+  SmoothnessEstimator::Legendre namespace, as well as @cite mavriplis1994hp ,
+  @cite eibner2007hp and @cite davydov2017hp.</li>
+
+  <li><i>Refinement history:</i> The last strategy is quite different
+  from the other two. In theory, we know how the error will converge
+  after changing the discretization of the function space. With
+  h-refinement, the solution converges algebraically as already pointed
```
bangerth, line 247:
```suggestion
  $h$-refinement: the solution converges algebraically as already pointed
```
Maybe say what "algebraically" means?

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+  out in step-7. It the solution is sufficiently smooth, though, we
```
bangerth, line 248:
```suggestion
  out in step-7. If the solution is sufficiently smooth, though, we
```

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+  expect that the solution will converge exponentially with increasing
+  polynomial degree of the finite element. We can compare a proper
+  prediction of the error with the actual error in the following step to
+  see verify if our choice of adaptation type was justified. The transition
```
bangerth, line 252:
```suggestion
  see if our choice of adaptation type was justified. The transition
```

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+  to this strategy is a bit more complicated. For this, we need an
+  initialization step with pure h- or p-refinement and we need to transfer
```
bangerth, line 254:
```suggestion
  initialization step with pure $h$- or $p$-refinement and we need to transfer
```

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+  the predicted errors over adapted meshes. The extensive documentation
+  of the hp::Refinement::predict_error() function describes not only the
+  theoretical details of this approach, but also presents a blueprint on
+  how to implement this strategy in your code. Note that this strategy
+  is only suited for time-independent problems or to create an initial
```
bangerth, line 259: I don't think I understand this comment. I *think* that what you're trying to say is that one can't predict the smoothness of the solution at the next time step, but if one wanted to build the mesh for just one time step (e.g., start each step with a coarse mesh, then keep refining until we're happy with the solution, and only then moving on to the next time step), then this approach here could be used as well.

I think that your comment is not actually specific to this strategy but all the others as well.

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+  mesh for time-dependent problems. For more information, see
+  @cite melenk2001hp .</li>
+</ul>
+
+Try implementing one of these strategies into this tutorial and observe the
+subtle changes to the results. You will notice that all strategies are
+capable of identifying the singularities near the reentrant corners and
+will perform h-refinement in these regions, while preferring p-refinement
```
bangerth, line 267:
```suggestion
will perform $h$-refinement in these regions, while preferring $p$-refinement
```

examples/step-27/doc/results.dox
```diff
@@ -209,7 +209,71 @@ include estimating the smoothness not on single cells, but cell assemblies or
...
+in the bulk domain.
```
bangerth, line 268: I think it would be fair to cite your PhD thesis somewhere here. You *did* implement all of these things, and it makes sense to point the reader to a place where they can find the kind of comparison you're suggesting here.

bangerth (commented):
Just the documentation part. I'll get back to the `.cc` at some later time, but now I have to go meet you (@marcfehling) for a bike ride :-)

examples/step-27/doc/results.dox
marcfehling: Yes, we could do this or point to the parallel hp-paper (if we decide to describe them there, too).

examples/step-27/doc/results.dox
marcfehling: You are right, this short comment does not give the issue justice that I wanted to address. I've extended the documentation.

examples/step-27/doc/results.dox
bangerth: For now, just reference your thesis. We can always add other references at a later time.

examples/step-27/step-27.cc
```diff
@@ -49,17 +50,19 @@
 
 // These are the new files we need. The first and second provide the
 // FECollection and the <i>hp</i> version of the FEValues class as described in
-// the introduction of this program. The last one provides Fourier
-// transformation class on the unit cell.
+// the introduction of this program. The next one provides the functionality
+// for automatic hp-adaptation, for which we will use the estimation algorithms
```
bangerth, line 54:
```suggestion
// for automatic $hp$-adaptation, for which we will use the estimation algorithms
```

examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
-          {
-            cell->clear_refine_flag();
-            cell->set_active_fe_index(cell->active_fe_index() + 1);
-          }
+      // relative threshold. In other words, for every cell for which (i) the
+      // refinement flag is set, (ii) the smoothness indicator is larger than
+      // the threshold, and (iii) we still have a finite element with a
+      // polynomial degree higher than the current one in the finite element
+      // collection, we will assign a future FE index that corresponds to a by
+      // one higher polynomial degree. The following function is capable of
```
bangerth, line 426: ...that corresponds to a polynomial with degree one higher than it currently is

examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
+      // doing exactly this. Absent any better strategies, we will set the
+      // threshold as the mean value between minimal and maximal smoothness
+      // indicators on cells flagged for refinement. This is achieved by setting
+      // the corresponding fraction parameter to a value of 0.5. We deal the
+      // same way with cells that are going to be coarsened and decrease their
```
bangerth, line 431: -> In the same way, we deal with cells that are going...

examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
+      // polynomial degree when their smoothness indicator is below the
+      // corresponding threshold determined on cells to be coarsened.
+      hp::Refinement::p_adaptivity_from_relative_threshold(
+        dof_handler, smoothness_indicators, 0.5, 0.5);
+
+      // The above function only determines whether the polynomial degree will
+      // change via future FE indices, but does not manipulate the h-refinement
```
bangerth, line 438:
```suggestion
      // change via future FE indices, but does not manipulate the $h$-refinement
```

examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
+      // flags. So for cells that are flagged for both refinement categories, we
+      // prefer p- over h-refinement. The following function call ensures that
```
bangerth, line 440:
```suggestion
      // prefer $p$- over $h$-refinement. The following function call ensures that
```

examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
+      // only either p- or h-refinement is imposed, and not both at once.
```
bangerth, line 441:
```suggestion
      // only one of $p$- or $h$-refinement is imposed, and not both at once.
```

examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
-  void LaplaceProblem<2>::create_coarse_grid()
+  // The following function is used when creating the initial grid. The grid we
+  // would like to create is actually similar to the one from step-14, i.e., the
+  // square domain with the square hole in the middle. It can be generated by
+  // excatly the same function. However, since its implementation is only a
+  // specialization of the 2d case, we will present a different way of creating
+  // this domain which is dimension independent.
+  //
+  // We first create a hypercube triangulation with enough cells so that it
+  // already holds our desired domain. We then continue to remove those cells in
```
bangerth, line 464:
```suggestion
  // already holds our desired domain $[-1,1]^d$, subdivided into $4^d$ cells. We then remove those cells in
```

examples/step-27/step-27.cc
```diff
@@ -474,103 +418,70 @@ namespace Step27
...
+  // the center of the domain by identifying them with the physical coordinates
```
bangerth, line 465:
```suggestion
  // the center of the domain by testing the coordinate values
```

examples/step-27/step-27.cc
bangerth: I think this is ok.

bangerth (commented):
Nice job removing all of this code!

marcfehling:
Addressed your comments. As we keep the Fourier strategy, I will generate updated output and figures now.

Who do I need to talk to in order to upload updated result images? :)

bangerth (approved):

bangerth:
You'd talk to me :-)

masterleinad:
Is this still WIP?

marcfehling:
> Is this still WIP?

Unfortunately yes :(

While preparing updated results, I realized that there are issues with the Fourier strategy as the generated mesh now looks far different than the old one. My best guess that the configuration of the `FESeries::Fouier` object is not sufficient. I'll figure out the cause of this...

bangerth:
Let us know once you've got this ready. I'd also be happy to just merge this as-is, and then take the remaining adjustments as a follow-up.

marcfehling:
> Let us know once you've got this ready. I'd also be happy to just merge this as-is, and then take the remaining adjustments as a follow-up.

Fair enough! Let's address details about the hp-parameters and the output in a separate PR then. I'll release this PR from its WIP state.

PR #11420, opened 2020-12-25
Test alternative path in QProjector

peterrum:
The first commit is from #11419.

This PR enables in theory that also for quad and hex the new, more general code path in `QProjector` is used. It should give some confidence that the new implementation works. However, it is disabled for now for two reason: 1) projection onto the faces of hypercube-like cells is trivial and 2) in many of the places of the library a certain order of the face quadrature points is needed (since `QProjector::DataSetDescriptor::face` is not used consistently). I would, nevertheless, appreciate if we could merge this PR, since the change enables to test me offline the routine with a larger set of old tests (in particular the matrix-free tests).

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -887,8 +887,14 @@ namespace ReferenceCell
         standard_line_to_face_and_line_index(
           const unsigned int line) const override
         {
-          static const std::array<unsigned int, 2> table[6] = {
-            {{0, 0}}, {{0, 1}}, {{0, 2}}, {{1, 1}}, {{1, 2}}, {{2, 1}}};
+          static const std::array<unsigned int, 2> table[8] = {{{0, 0}},
+                                                               {{0, 1}},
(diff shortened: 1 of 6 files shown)


bangerth:
Can you please rebase?

bangerth:
I'm not a fan of this patch. It adds a code path that's only visible if you `#ifdef` a pp variable that by default is not defined. The consequence is that we end up with code in the library that is not compiled, not tested, and not executed. Such code is subject to bit-rot and makes it hard to read what the rest of the code is doing.

Can you explain what the difference between the `#if` and the `#else` part is? Why do we need the currently disabled code path? And if we need it, couldn't we get it by passing a flag to the appropriate class/function?

bangerth:
@peterrum  Ping?

peterrum:
Let me close this PR. I would love to change the ordering within of `QProjector`. However, this won't happen, since there are too many places in the library that make assumptions regarding the ordering of faces/orientations.

bangerth:
You can always open an issue that tracks this...

PR #11421, opened 2020-12-26
Fix formatting

peterrum:


tjhei (approved):
Wow. What happened to the formatting here?

PR #11422, opened 2020-12-26
Work on compute_embedding_matrices() for simplex

peterrum:
~Currently, this is only a test but we can later on move the implementation to `FETools`. In particular, a transparent way to construct mappings, quadratures, and coarse grids would be beneficial.~

~Depends on #11426.~


peterrum:
I have updated this PR.

include/deal.II/fe/fe_tools.templates.h
```diff
@@ -1921,12 +1936,11 @@ namespace FETools
...
-                             std::vector<std::vector<FullMatrix<number>>
-
-                                         > &                     matrices,
-                             const bool                          isotropic_only,
-                             const double                        threshold)
+  compute_embedding_matrices(
+    const FiniteElement<dim, spacedim> &          fe,
+    std::vector<std::vector<FullMatrix<number>>> &matrices,
+    const bool                                    isotropic_only,
+    const double                                  threshold)
```
drwells, line 1943: I'm not sure what is going on with the indentation here but the new version is way better.

drwells (approved):

PR #11423, opened 2020-12-27
Introduce ReferenceCell::unit_tangential_vectors()

peterrum:
References #11352. FYI @drwells


include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
     return Point<dim>(+0.0, +0.0, +0.0);
   }
 
+  /**
+   * Return i-th unit tangential vector (array of dim-1 elements of Point<dim>)
```
bangerth, line 187:
```suggestion
   * Return i-th unit tangential vector (i.e., an element of an array of `dim-1` elements of type Tensor<1,dim>)
```

include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
...
+   * of a face of the reference cell. The vectors are arranged such that the
+   * cross product between the two vectors return the unit normal vector.
```
bangerth, line 189:
```suggestion
   * cross product between the two vectors returns the unit normal vector of the face.
```

include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
...
+   */
+  template <int dim>
+  inline Tensor<1, dim>
+  unit_tangential_vectors(const Type &       reference_cell,
+                          const unsigned int face_no,
+                          const unsigned int i)
+  {
+    if (reference_cell == get_hypercube(dim))
+      return GeometryInfo<dim>::unit_tangential_vectors[face_no][i];
```
bangerth, line 198: I think your goal is to long-term remove `GeometryInfo`, right? If so, how about reversing the order: putting the existing code here and initializing the `GeometryInfo` field by calling this function?

Not sure whether that's easy, but I thought I'd ask.

include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
...
+    else if (reference_cell == Type::Tri)
+      {
+        AssertDimension(i, 0);
```
bangerth, line 201: The other cases don't seem to have this assertion. I think you could put it at the top of the function if you manage to extract the dimensionality of the `reference_cell` object somehow.

include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
...
+        static const std::array<Tensor<1, dim>, 3> table = {
```
bangerth, line 202: Maybe make it `constexpr`? I wonder if the whole function could actually be `constexpr`.

That's also not important: If it's easy to do, do it. Otherwise leave it for someone to clean up later.

include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
...
+          {Point<dim>(1, 0),
```
bangerth, line 203: You don't need the detour of creating a `Point` object that is then cast to the base `Tensor` class:
```suggestion
          {{1, 0},
```

bangerth: same below

include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
...
+             Point<dim>(0, +1.0 / sqrt(2.0), +1.0 / sqrt(2.0))}},
+           {{Point<dim>(1, 0, 0),
+             Point<dim>(0, +1.0 / sqrt(2.0), -1.0 / sqrt(2.0))}}}};
+
+        return table[face_no][i];
+      }
+
+    Assert(false, ExcNotImplemented());
+
+    return Point<dim>();
```
bangerth, line 254:
```suggestion
    return {};
```

include/deal.II/grid/reference_cell.h
```diff
@@ -183,6 +183,77 @@ namespace ReferenceCell
...
+   * of a face of the reference cell. The vectors are arranged such that the
+   * cross product between the two vectors return the unit normal vector.
+   */
+  template <int dim>
+  inline Tensor<1, dim>
+  unit_tangential_vectors(const Type &       reference_cell,
+                          const unsigned int face_no,
+                          const unsigned int i)
+  {
+    if (reference_cell == get_hypercube(dim))
```
bangerth, line 197: Can you add a check that the given `dim` argument matches the dimensionality of `reference_cell`?

bangerth (commented):
Nearly there.

Do you think it would be possible to have a test that just checks the output of your new function for all possible arguments?

include/deal.II/grid/reference_cell.h
peterrum: I see your point. But I would keep it for now as it is (since this is the approach I have taken throughout the whole file). I have put this issue onto the TODO list here: https://github.com/dealii/dealii/issues/11352#issue-761425796

include/deal.II/grid/reference_cell.h
peterrum: This assert is only needed here since this is the only cell type that is 2D (and a consequence only has a single tangential).

include/deal.II/grid/reference_cell.h
peterrum: This line is exactly doing this implicitly. The function `get_hypercube()` return the reference-cell type of the right dimension so that if that does not matches the comparison will fail.

include/deal.II/grid/reference_cell.h
peterrum: This is not possible since there is an conversion from `Point` to `Tensor`. The reason why `Points` is used here (and I guess at other places from where I had the idea write it in such a way) it that `Point` has appropriate constructors taking 1-3 doubles. The last point rules out that we can simplify the code with `{}` as proposed by you below.

peterrum:
@bangerth Thanks for the feedback. I have addressed your comments and have added a test.

include/deal.II/grid/reference_cell.h
bangerth: What I meant is: Add such an assertion before the first `if`. Right now, if you call this function with `dim==2` and `reference_cell==tet`, you'll run into the `ExcNotImplemented()` at the very bottom, but that's not helpful to the user: it isn't that the necessary code isn't implemented, but that the function just doesn't make sense with this combination of arguments.

include/deal.II/grid/reference_cell.h
bangerth: But what if someone calls the function with `reference_cell==wedge` and `i==14`? In that case, you're causing an out-of-bounds access for `std::array` that isn't caught.

include/deal.II/grid/reference_cell.h
bangerth: But class `Tensor` has this constructor:
```
  using array_type =
    typename Tensor<rank_ - 1, dim, Number>::array_type[(dim != 0) ? dim : 1];

  constexpr DEAL_II_CUDA_HOST_DEV explicit Tensor(
    const array_type &initializer);
```
You might have to use an extra pair of braces, but you can direct-initialize the `Tensor` objects without the detour of `Point`.

tests/simplex/unit_tangential_vectors_01.cc
```diff
@@ -0,0 +1,55 @@
+// ---------------------------------------------------------------------
```
bangerth, line 1: thanks for adding the test!

include/deal.II/grid/reference_cell.h
peterrum: That is an explicit constructor. So it will not work.

peterrum:
@bangerth  I have added some asserts.

include/deal.II/grid/reference_cell.h
```diff
@@ -45,6 +45,31 @@ namespace ReferenceCell
     Invalid = static_cast<std::uint8_t>(-1)
   };
 
+  /**
+   * Return the dimension of the riven reference-cell type @p type.
```
bangerth, line 49:
```suggestion
   * Return the dimension of the given reference-cell type @p type.
```

bangerth (approved):
Thanks for all the changes -- I appreciate your patience!

There is just one typo in the documentation of one of the new functions. But let's move ahead with it -- fix it in a follow-up or one of the other patches in your series.

bangerth:
Ah, I wanted to merge, but there is now a merge conflict :-( Feel free to rebase and merge yourself.

peterrum:
Since @bangerth gave me the go ahead, I am merging.

PR #11425, opened 2020-12-27
Fix output of grid/cell_id_08

peterrum:
FYI @marcfehling

### DIFF
--- a/tests/grid/cell_id_08.cc
+++ b/tests/grid/cell_id_08.cc
@@ -105,8 +105,9 @@ benchmark(const unsigned int n_global_refinements)
     AssertThrow(cell_ids == deserialized,
                 ExcMessage("Serialization with boost binary archives failed."));
     deallog << "Serialization with boost binary archives successful."
-            << std::endl
-            << "  Buffer size in bytes: " << size_in_bytes(buffer) << std::endl;
+            << std::endl;
+    std::cout << "  Buffer size in bytes: " << size_in_bytes(buffer)
+              << std::endl;
   }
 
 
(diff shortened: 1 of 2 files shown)


marcfehling:
Interesting. This test does not fail on my machine. What kind of version of `boost` and the standard library are you using?

I've used the bundled `boost` library and the standard library shipped with `gcc 8.1.0`.

If the results really depend on the version of third party libraries, we should move the output of buffer sizes in this test from `deallog` to `std::cout`.

bangerth:
Do any of you know which patch broke this?

marcfehling:
> Do any of you know which patch broke this?

I've introduced this test in #11364. The testsuite passed on this one back then.

So it must be either a very recent patch, or this issue depends on the `boost` and `std` libraries.

peterrum:
I have removed the problematic output. The test also fails on CDash.

tests/grid/cell_id_08.cc
```diff
@@ -105,8 +105,9 @@ benchmark(const unsigned int n_global_refinements)
     AssertThrow(cell_ids == deserialized,
                 ExcMessage("Serialization with boost binary archives failed."));
     deallog << "Serialization with boost binary archives successful."
-            << std::endl
-            << "  Buffer size in bytes: " << size_in_bytes(buffer) << std::endl;
+            << std::endl;
+    std::cout << "  Buffer size in bytes: " << size_in_bytes(buffer)
+              << std::endl;
```
bangerth, line 110: Did you mean to remove these two lines? The output you produce here is identical to the code before -- you just split it over two statements.

bangerth (changes requested):
I think you made a mistake in your changes, but the idea of just removing the line seems reasonable to me.

bangerth:
For the record, the current version of the test passes on my system. It must be a BOOST version thing, but in the end it is not particularly important how large the compressed buffer is so let's do as you suggest and just remove the size output.

tests/grid/cell_id_08.cc
peterrum: The output is not identical, is it? The size is now written to `std::cout` as suggested in https://github.com/dealii/dealii/pull/11425#issuecomment-751531382.

marcfehling (approved):

tests/grid/cell_id_08.cc
bangerth: Oh, I see. I missed that part. I think it's pointless to write to `std::cout` -- nobody is ever going to see that output -- but let's just go with it for now. I'm going to remove this again in a follow-up. For now, let's just make the test failure go away.

bangerth (approved):

PR #11427, opened 2020-12-27
Fix tests with output files with same name

peterrum:
In an application code, I had problems to use the test suite when test output files had the same file name (even though the category differed).

### DIFF
--- a/cmake/macros/macro_deal_ii_add_test.cmake
+++ b/cmake/macros/macro_deal_ii_add_test.cmake
@@ -166,14 +166,14 @@ MACRO(DEAL_II_ADD_TEST _category _test_name _comparison_file)
 
         SET(_source_file "${_test_name}.cc")
 
-        SET(_target ${_test_name}.${_build_lowercase}) # target name
+        SET(_target ${_category}.${_test_name}.${_build_lowercase}) # target name
         SET(_run_args "$<TARGET_FILE:${_target}>") # the command to issue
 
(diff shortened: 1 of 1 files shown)


cmake/macros/macro_deal_ii_add_test.cmake
```diff
@@ -192,9 +192,9 @@ MACRO(DEAL_II_ADD_TEST _category _test_name _comparison_file)
         ENDIF()
 
         IF(NOT "${TEST_TARGET_${_build}}" STREQUAL "")
-          SET(_target ${TEST_TARGET_${_build}})
+          SET(_target ${_category}.${TEST_TARGET_${_build}})
         ELSEIF(NOT "${TEST_TARGET}" STREQUAL "")
-          SET(_target ${TEST_TARGET})
+          SET(_target ${_category}.${TEST_TARGET})
```
kronbichler, line 197: Do we need to append the error message in line 201 below as well or is it clear nonetheless?

peterrum: I would say `_target` is only function internal, while `${TEST_TARGET}` come from outside which is the motivation for the error message.

kronbichler:
Can you have a look if the two failing tests in the serial CI are related to this change?

peterrum:
@kronbichler I am on it ;)

masterleinad (approved):
Looks reasonable to me.

PR #11431, opened 2020-12-29
CMake: Fix Ginkgo linkage

tamiko:
We also have to the "ginkgo" library itself...

In reference to #11413


masterleinad (approved):
Hopefully, it works this time.

tamiko:
OK. Configured successfully on all testers - none of them has ginkgo installed. I will merge and restart the regression tester.

PR #11432, opened 2020-12-29
MatrixFreeTools::compute_diagonal() make class const

peterrum:


masterleinad (approved):
The more `const` the better (if we can)!

PR #11437, opened 2020-12-29
Clean up FE_Nothing::get_name().

drwells:
At the same time, since `FiniteElement::operator==` uses `get_name()` we don't need to implement a custom one for `FE_Nothing`.

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -71,6 +71,39 @@ namespace ReferenceCell
   }
 
   /**
+   * Convert the given reference cell type to a string.
+   */
+  inline std::string
+  to_string(const Type &type)
+  {
+    switch (type)
+      {
+        case Type::Vertex:
+          return "Vertex";
+        case Type::Line:
+          return "Line";
+        case Type::Tri:
+          return "Tri";
+        case Type::Quad:
+          return "Quad";
+        case Type::Tet:
+          return "Tet";
+        case Type::Pyramid:
+          return "Pyramid";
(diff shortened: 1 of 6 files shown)


peterrum (approved):

masterleinad (approved):
Looks reasonable to me.

include/deal.II/grid/reference_cell.h
```diff
@@ -45,6 +45,38 @@ namespace ReferenceCell
     Invalid = static_cast<std::uint8_t>(-1)
   };
 
+  /**
+   * Convert the given reference cell type to a string.
+   */
+  inline std::string
+  type_to_string(const Type &type)
```
bangerth, line 52: Just name it `to_string` so that it matches `std::to_string`.

include/deal.II/grid/reference_cell.h
```diff
@@ -45,6 +45,38 @@ namespace ReferenceCell
...
+        case Type::Tet:
+          return "Tet";
+        case Type::Pyramid:
+          return "Pyramid";
+        case Type::Wedge:
+          return "Wedge";
+        case Type::Hex:
+          return "Hex";
+        case Type::Invalid:
+        default:
```
bangerth, line 73: Since you fall through from `Invalid` to `default`, you don't actually need the last `case` statement.

drwells: Let me fix this in another way - we might want to convert `Invalid` to a string after all.

bangerth (approved):

PR #11440, opened 2020-12-30
Avoid MPI calls in the HDF5 destructors when there is an exception - DataSet

dangars:
In #11339 I forgot to add `std::uncaught_exception` in one of the constructors of `DataSet`

See comment https://github.com/dealii/dealii/pull/11339#issuecomment-751899182 of @bangerth


drwells:
This solution is fine but could we add some utility function so that we can have the error message and exception check in only one place?

bangerth (approved):
Let's get this merged and do the unification of code as a follow-up. That should be easier to review.

PR #11442, opened 2020-12-31
Make a test more verbose.

bangerth:
I spent about 2 hours debugging a problem in the next patch I'm going to submit in a minute. In the end, what it came down to was to just output more information about what's currently stored in particles in the `particles/step-19` test. That could have been avoided if I had just added more output statements to that test.

This patch does this. It adds ~60kB to the output, but I think that's worth it because this is the only real-world testcase for particles we have in the test suite: if something nontrivial in our code base goes wrong (as it did in my case), this might very well be the only test that fails (as it did in my case). Having more verbose output would have made debugging the root cause substantially easier.

/rebuild

### DIFF
--- a/tests/particles/step-19.cc
+++ b/tests/particles/step-19.cc
@@ -487,6 +487,10 @@ namespace Step19
         if (particle_handler.n_particles_in_cell(cell) > 0)
           for (const auto &particle : particle_handler.particles_in_cell(cell))
             {
+              deallog << "Evaluating particle " << particle.get_id() << " at "
+                      << particle.get_location() << " with reference location "
+                      << particle.get_reference_location() << std::endl;
+
               const Point<dim> reference_location =
                 particle.get_reference_location();
               for (const unsigned int i : fe_values.dof_indices())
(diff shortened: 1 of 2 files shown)


drwells (approved):

blaisb:
On a similar note, do  you think it would be a good idea to derive a test from step-68? Step-68 with few particles/cells is already very lightweight. It could be a test with mpirun=2 or 3...

bangerth:
@drwells More tests more better. Want to give it a shot, or should I?

drwells:
I don't know much about particles so I think it is a better fit for you to do it.

peterrum (approved):

blaisb:
> @drwells More tests more better. Want to give it a shot, or should I?

@bangerth was that intended for me? If you want, I can take  a jab at making a step-68ish test. I know that @kronbichler used it a bit to benchmark particle and it would be pretty much the only full-fledged parallel particle test.

bangerth:
@blaisb I had meant it for @drwells but it would be equally fantastic if you could convert step-68 to a test. As a test, it (i) doesn't have to run very long to be indicative -- just a few cells and a few particles will do, (ii) in fact, it should run relatively quickly, (iii) it should output relevant information into the log file so that inadvertent changes to the library show up in the tests. So, particle locations or properties at the end of a time step are good indicators, for example.

blaisb:
> @blaisb I had meant it for @drwells but it would be equally fantastic if you could convert step-68 to a test. As a test, it (i) doesn't have to run very long to be indicative -- just a few cells and a few particles will do, (ii) in fact, it should run relatively quickly, (iii) it should output relevant information into the log file so that inadvertent changes to the library show up in the tests. So, particle locations or properties at the end of a time step are good indicators, for example.

Seems good! I'll try to give it a shot in a relatively close time frame.

PR #11443, opened 2020-12-31
Store particle locations in the PropertyPool

bangerth:
This is the next step in #11206. It adds the interface to also store locations/reference locations in the property pool (but doesn't use that yet -- coming up soon). In order to do that, though, one always has to have a property pool associated with each particle, even if that particle doesn't actually store any "properties" (we can then call the locations these particles' only "properties"). So I now have a global pool that every newly created particle attaches to, and when a particle is inserted into a `ParticleHandler`, we switch to that `ParticleHandler`'s property pool.

/rebuild

### DIFF
--- a/include/deal.II/particles/particle.h
+++ b/include/deal.II/particles/particle.h
@@ -480,16 +480,6 @@ namespace Particles
     static PropertyPool<dim, spacedim> global_property_pool;
 
     /**
-     * Current particle location.
-     */
-    Point<spacedim> location;
-
-    /**
-     * Current particle location in the reference cell.
-     */
-    Point<dim> reference_location;
-
-    /**
      * Globally unique ID of particle.
      */
     types::particle_index id;
@@ -568,7 +558,7 @@ namespace Particles
   inline void
   Particle<dim, spacedim>::set_location(const Point<spacedim> &new_loc)
   {
-    location = new_loc;
+    property_pool->set_location(property_pool_handle, new_loc);
   }
 
 
@@ -577,7 +567,7 @@ namespace Particles
   inline const Point<spacedim> &
   Particle<dim, spacedim>::get_location() const
   {
-    return location;
+    return property_pool->get_location(property_pool_handle);
   }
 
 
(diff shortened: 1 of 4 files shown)


bangerth:
The one failing CI check fails like so:
```
The following tests FAILED:
	3818 - multigrid/transfer_matrix_free_06.mpirun=15.debug (Timeout)
```
I believe this to be spurious because that test has nothing to do with particles.

A review would be appreciated :-)

bangerth:
I updated this: the `PropertyPool` object now stores locations and reference locations.

bangerth:
Ping?

luca-heltai:
I really like where this is going. I'll have to read through it once more to understand everything, but I have one question already:

Consistently with the current implementation, you are giving for granted that every particle needs both a location and a reference location. Is there an easy way within this PR to remove the second set of properties from the global property pool, and make it optional? This implementation has the same flaw of the previous one: a particle has no idea of what cell it belongs to, yet, it knows its reference position within that cell. Would it be reasonable to implement it in such a way that the reference location pool is initialized only the first time it is called/accessed?

bangerth:
It wouldn't be very difficult to come up with a patch to tell the `PropertyPool` whether or not to store reference locations, now that all of this is in a central place. We'd just keep the corresponding vector empty and fail every access to that property.

But that's for a later patch :-)

luca-heltai (approved):

luca-heltai:
I'd like @gassmoeller to take a look, but it seems ready to me.

include/deal.II/particles/particle.h
```diff
@@ -641,9 +656,10 @@ namespace Particles
   inline const ArrayView<const double>
   Particle<dim, spacedim>::get_properties() const
   {
-    Assert(has_properties(), ExcInternalError());
-
-    return property_pool->get_properties(property_pool_handle);
+    if (has_properties() == 0)
```
gassmoeller, line 659:
```suggestion
    if (has_properties() == false)
```

gassmoeller (commented):
Yes, except for my one small comment this looks great :+1:. Sorry I missed this before.

bangerth:
@luca-heltai -- I fixed the one comment by @gassmoeller . Should be ready to go.

I guess I should probably also store the particle id in the `PropertyPool`? If you agree, I'll write a follow-up patch.

PR #11444, opened 2020-12-31
Refactor the HDF5 abort code.

drwells:
Followup to #11440.

### DIFF
--- a/source/base/hdf5.cc
+++ b/source/base/hdf5.cc
@@ -25,6 +25,50 @@ DEAL_II_NAMESPACE_OPEN
 
 namespace HDF5
 {
+  namespace internal
+  {
+    namespace HDF5ObjectImplementation
+    {
+      /**
+       * Abort, should there be an exception being processed (see the error
+       * message).
+       */
+      void
+      check_exception(const std::string &type, const std::string &name)
+      {
+#  ifdef DEAL_II_WITH_MPI
+#    if __cpp_lib_uncaught_exceptions >= 201411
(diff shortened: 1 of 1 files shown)


source/base/hdf5.cc
```diff
@@ -25,6 +25,49 @@ DEAL_II_NAMESPACE_OPEN
...
+          {
+            std::cerr
+              << "---------------------------------------------------------\n"
+              << "HDF5 " + type + " objects call " + name +
+                   " to end access to\n"
+              << "them and release any resources they acquired. This call\n"
+              << "requires MPI synchronization. Since an exception is\n"
+              << "currently uncaught, this synchronization will be skipped\n"
+              << "to avoid a possible deadlock. As a result the HDF5 file\n"
+              << "might be corrupted.\n"
```
bangerth, line 55: These last two sentences aren't actually correct any more -- I guess we missed that. Maybe "...currently uncaught, the synchronization would likely deadlock because only the current process is trying to destroy the object. As a consequence, the program will be aborted."

bangerth (changes requested):

bangerth:
I would have done it, but you beat me to it :-)

source/base/hdf5.cc
drwells: Fixed!

drwells:
I complained so I should be the one to fix it.

dangars (approved):
@drwells  Thanks a lot! It is definitely better to put the error message and the exception code in one place :)

bangerth (approved):

PR #11446, opened 2020-12-31
Update grid/cell_id_08 output.

marcfehling:
Follow-up to #11425.

Restructures the output of test `grid/cell_id_08.cc` and puts it into better context.

@bangerth -- I don't think we should just get rid of these lines in this test. If someone decides to continue to work on `CellId` or serialization, this test could be used to easily check the results of these changes. It would help to make this test more prominent in this regard. Would it make sense to move this to `dealii/performance-benchmarks`?

### DIFF
--- a/tests/grid/cell_id_08.cc
+++ b/tests/grid/cell_id_08.cc
@@ -74,8 +74,8 @@ benchmark(const unsigned int n_global_refinements)
     for (const auto &cell : tria.active_cell_iterators())
       cell_ids.push_back(cell->id());
   }
-  deallog << "Creation successful." << std::endl
-          << "  Number of CellId objects: " << cell_ids.size() << std::endl;
+  deallog << "Creation successful." << std::endl;
+  std::cout << "Number of CellId objects: " << cell_ids.size() << std::endl;
 
 
   //
@@ -106,7 +106,8 @@ benchmark(const unsigned int n_global_refinements)
                 ExcMessage("Serialization with boost binary archives failed."));
     deallog << "Serialization with boost binary archives successful."
             << std::endl;
-    std::cout << "  Buffer size in bytes: " << size_in_bytes(buffer)
+    std::cout << "Buffer size in bytes with ..."
+              << "  boost binary_archive : " << size_in_bytes(buffer)
               << std::endl;
   }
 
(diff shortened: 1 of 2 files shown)


peterrum (approved):

bangerth:
I won't object to the patch, but maybe someone else should make the decision.

The way we've generally used the test suite is by comparing what it writes to `deallog` with what has previously been stored. The majority of tests don't write anything to `std::cout` and so one doesn't generally look into what's being output on the console. In other words, I don't expect to look there, and even if I do look, it's hard to know what to do with it because I don't know what the correct answer is. In practice output to the console is therefore a bit pointless -- though I'm not opposed to putting some output there if others feel like it would be a useful thing for future-ourselves.

tjhei:
> Would it make sense to move this to `dealii/performance-benchmarks`?

You are welcome to replace/modify https://github.com/dealii/performance-benchmarks/blob/master/cell_id/main.cc

bangerth:
Ping? Anyone?

masterleinad (commented):
I agree with @bangerth that I would know what to do with the output if the test fails so I would omit it if we don't check it.

tjhei:
Let me side with Marc then. :-)
I don't think these statements hurt and we have thousands of them in the test suite. I like having additional information when I have to look at a broken test. That includes commented out ``write_vtu()`` calls, comments, std::cout messages, and more. None of them are important for the test to pass/fail.

masterleinad:
I certainly don't feel strongly about it. :slightly_smiling_face:

marcfehling:
I drew inspiration from [`base/utilities_pack_unpack_07.cc`](https://github.com/dealii/dealii/blob/master/tests/base/utilities_pack_unpack_07.cc) which redirects timing results to `std::cout`, and thought this would be nice additional information. Similarly, I wanted to add the buffer sizes to the output, which are now redirected to `std::cout` as well.

I still think this is useful information whenever someone gets back to this test. And a simple output of `size` does not cost any performance.

marcfehling:
@peterrum Different question: were you using `clang/libc++` in #11425? If so, a separate output log would do the trick, and this discussion would be over :-)

bangerth:
Let's not agonize over it. If some people feel like it's useful, and it's not actively harmful to anyone else, then let's just do it.

PR #11447, opened 2020-12-31
Utility grid generator function `hyper_line` for tests.

marcfehling:
I keep re-using the same grid generation code snippet for many tests.

I thought this would make a nice addition to the utility functions for the testsuite. This suggestion is not versatile enough to be part of the `GridGenerator` namespace, so I thought it would be best suited in `tests/test.h`. What do you think?

EDIT: I have just spotted `tests/test_grids.h` and moved it there. I think this is a good place to keep it there.

### DIFF
--- a/tests/hp/error_prediction_01.cc
+++ b/tests/hp/error_prediction_01.cc
@@ -23,7 +23,6 @@
 
 #include <deal.II/fe/fe_q.h>
 
-#include <deal.II/grid/grid_generator.h>
 #include <deal.II/grid/tria.h>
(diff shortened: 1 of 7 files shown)


peterrum (approved):
Looks good. But for me it is simply a `line` in `spacedim` space.

marcfehling:
> Looks good. But for me it is simply a `line` in `spacedim` space.

I have no objection to simply calling the function `line` :)

drwells (approved):

PR #11430, opened 2020-12-28
Better document FE_Nothing.

bangerth:
/rebuild

Issue #1496, opened 2015-08-31
FE_Nothing ends up dominating when in an FE_System


bangerth:
@davydden posted a testcase as part of #1424 (now hidden by follow-up commits) that compared a Q2xQ1 against a Q1xNothing. We should get hanging node constraints where the Q2 and the Q1 come together, but no additional constraints for the Q1-vs-Nothing. But we do. That's wrong.

The testcase that goes with this (and that currently produces the wrong answer) is `hp/face_domination_02`, as introduced by #1495. Someone will need to debug `make_hanging_node_constraints` to see what is going wrong. I fear that the problem is that we end up with this because the two vector components have to be treated differently but we only have a single matrix to do this.

bangerth:
![2015-08-31-173257](https://cloud.githubusercontent.com/assets/7504421/9592183/6aeaa25c-5006-11e5-99ee-4921f0b236e3.jpg)

As shown, we should get (and do get) the constraint `x_7=1/2(x_13+x_11)`. But we also do get `x_5=0` and `x_2=0`, and this is wrong.

davydden:
>  I fear that the problem is that we end up with this because the two vector components have to be treated differently but we only have a single matrix to do this.

agreed. Supposedly one would need to introduce a special treatment for FESystem (vector FE which can be broken into a separate FEs) but use the current logic for other vector FEs like FE_RaviartThomas.

p.s. here's a link to the original comment https://github.com/dealii/dealii/pull/1424#discussion_r37973319

davydden:
i hit that issue again while trying to implement something. It took a while to recall what's happening and why. A workaround could be to delete those constraints from constraint matrix, but this is dirty...

The issue is indeed related to the fact that we have a single matrix to treat/introduce constraints.
In scenarios depicted above we end up with 
`case FiniteElementDomination::either_element_can_dominate` (around line [1173](https://github.com/dealii/dealii/blob/master/source/dofs/dof_tools_constraints.cc#L1173)).
Consequently we constrain DoFs on child faces to those on the parent face by calling, at the end of the day,
```
ensure_existence_of_subface_matrix
(cell->get_fe(),
subface->get_fe(subface_fe_index),
c,
subface_interpolation_matrices
[cell->active_fe_index()][subface_fe_index][c]);
```
and
```
// Add constraints to global constraint matrix.
 filter_constraints 
(master_dofs,
slave_dofs,
*(subface_interpolation_matrices
[...]

PR #11448, opened 2020-12-31
Added tests for FENothing constraints.

marcfehling:
~~Blocked by #11447.~~ (EDIT: Has been merged)

The updated documention in #11430 states that `FE_Nothing` elements impose continuity constraints on neighboring continuous elements when their `dominate` parameter is set to `true`.

However, this test states that this is not the case: the `dominate` parameter has no effect. I feel like I stumbled over a bug.

@bangerth -- FYI

May be related to #1496.

### DIFF
--- a/tests/hp/fe_nothing_dominates.h
+++ b/tests/hp/fe_nothing_dominates.h
@@ -1,0 +1,92 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2021 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+
+// Check for continuity requirements of two neighboring finite elements
+// and project a given function subject to these constraints.
+
+
+#include <deal.II/base/function.h>
+
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/grid/tria.h>
+
+#include <deal.II/hp/fe_collection.h>
+#include <deal.II/hp/q_collection.h>
+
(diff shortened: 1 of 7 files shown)


tests/hp/fe_nothing_23.cc
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+int
+main()
+{
+  initlog();
+
+  deallog << std::boolalpha;
+  for (const bool fe_nothing_dominates : {false, true})
```
bangerth, line 102: I think it's just cool that C++11 allows for this construct :-)

bangerth (commented):
Yes, this looks like a bug. I guess the `constraints` object is empty in all cases? What happens if you exchange which element is on the two cells (in other words, from which side we end up looking at the interface)?

My suspicion is that the face interpolation matrix returned when looking at that face is `Nx0` and so we end up not enforcing any constraints. I'm confused why we didn't notice that when @davydden added that flag...

marcfehling:
> I guess the `constraints` object is empty in all cases?

Yes, the `constraints` object is empty in all cases.

> What happens if you exchange which element is on the two cells (in other words, from which side we end up looking at the interface)?

The result is the same. The `constraints` object is still empty. And all dofs on the FEQ cell all have the value 1.

marcfehling:
In `DoFTools::make_hp_hanging_node_constraints()`, we always add constraints from the point of view of the dominating finite element. However, we break if there are no face dofs, and thus no constraints will be imposed if `FENothing` elements dominate.
https://github.com/dealii/dealii/blob/b985807f268b6e179482ea06776f0c4ba75e305c/source/dofs/dof_tools_constraints.cc#L1509-L1513

I do not like that the `FENothing` class needs special treatment in so many different parts in the library, but I guess there is no other way here. I would replace the `break` statement and constrain all `dependent_dofs` to the value `0` here. Maybe check with a dynamic cast whether the neighboring FE is actually `FENothing`, as there may be other finite elements without face dofs.

Do you have a different (better) idea here?

marcfehling:
Let us merge this one too according to https://github.com/dealii/dealii/pull/11549#pullrequestreview-572817961.

marcfehling:
Since #11448 #11549 and #11600 basically did the same, I moved them into this PR, and moved the core function into a separate header file. Ready for review and tests.

bangerth (approved):

PR #11454, opened 2021-01-02
Generalize GridGenerator::flatten_triangulation for simplex meshes

peterrum:


bangerth (approved):
That looks like a no-brainer :-)

PR #11456, opened 2021-01-02
Add a version of update_neighbors() for simplex meshes

peterrum:
While working on PR #11455, I have noticed that the setup of neighborship is not done correctly during refinement of simplex meshes. This PR fixes this by specializing `update_neighbors()` for these kind of meshes. (I have tried to generalize the original `update_neighbors()` but it turned out to be impossible (?) due to the special treatment of the ansio case.) Most of the PR only moved stuff around.


source/grid/tria.cc
```diff
@@ -2312,6 +2143,191 @@ namespace internal
...
+       * For a given triangulation: set up the
+       * neighbor information on all cells.
+       */
+      template <int spacedim>
+      static void update_neighbors(Triangulation<1, spacedim> &)
+      {}
+
+
+      template <int dim, int spacedim>
+      void static update_neighbors(Triangulation<dim, spacedim> &triangulation)
```
tjhei, line 2157: static void?

source/grid/tria.cc
```diff
@@ -2312,6 +2143,191 @@ namespace internal
       }
 
 
+
+      /**
+       * For a given triangulation: set up the
```
tjhei, line 2148: remove this comment blob.

source/grid/tria.cc
```diff
@@ -9913,6 +9929,75 @@ namespace internal
...
+      void static update_neighbors(Triangulation<dim, spacedim> &triangulation)
+      {
+        std::vector<std::pair<unsigned int, unsigned int>> adjacent_cells(
+          2 * triangulation.n_raw_faces(),
+          {numbers::invalid_unsigned_int, numbers::invalid_unsigned_int});
+
+        const auto set_entry = [&](const auto &face_index, const auto &cell) {
+          const std::pair<unsigned int, unsigned int> cell_pair = {
+            cell->level(), cell->index()};
+          unsigned int index = 0;
```
tjhei, line 9946: you can skip the initialization of index here.

bangerth (approved):

PR #11459, opened 2021-01-03
Use FEPointEvaluation in step-19

peterrum:
I love this class (`FEPointEvaluation`). 

@kronbichler Maybe you could add some details on the new class in a follow-up PR!?

### DIFF
--- a/examples/step-19/step-19.cc
+++ b/examples/step-19/step-19.cc
@@ -37,8 +37,9 @@
 #include <deal.II/grid/grid_refinement.h>
 
 #include <deal.II/fe/mapping_q.h>
-#include <deal.II/fe/fe_values.h>
+#include <deal.II/fe/fe_point_evaluation.h>
 #include <deal.II/fe/fe_q.h>
+#include <deal.II/fe/fe_values.h>
 
 #include <deal.II/dofs/dof_handler.h>
 #include <deal.II/dofs/dof_tools.h>
@@ -643,13 +644,15 @@ namespace Step19
   // The approach used here is conceptually the same used in the
   // `assemble_system()` function: We loop over all cells, find the particles
   // located there (with the same caveat about the inefficiency of the algorithm
-  // used here to find these particles), and create an FEValues object with
-  // these positions:
+  // used here to find these particles), and use FEPointEvaluation object to
+  // evaluate the gradient at these positions:
   template <int dim>
   void CathodeRaySimulator<dim>::move_particles()
   {
     const double dt = time.get_next_step_size();
 
+    Vector<double>            solution_values(fe.n_dofs_per_cell());
+    FEPointEvaluation<1, dim> evaluator(mapping, fe);
 
     for (const auto &cell : dof_handler.active_cell_iterators())
       if (particle_handler.n_particles_in_cell(cell) > 0)
@@ -662,21 +665,15 @@ namespace Step19
           for (const auto &particle : particles_in_cell)
             particle_positions.push_back(particle.get_reference_location());
 
-          const Quadrature<dim> quadrature_formula(particle_positions);
-          FEValues<dim>         particle_position_fe_values(mapping,
-                                                    fe,
-                                                    quadrature_formula,
-                                                    update_gradients);
+          cell->get_dof_values(solution, solution_values);
 
-          particle_position_fe_values.reinit(cell);
-
-          // Then we can ask the FEValues object for the gradients of the
-          // solution (i.e., the electric field $\mathbf E$) at these locations
-          // and loop over the individual particles:
-          std::vector<Tensor<1, dim>> field_gradients(
-            quadrature_formula.size());
-          particle_position_fe_values.get_function_gradients(solution,
-                                                             field_gradients);
+          // Then we can ask the FEPointEvaluation object for the gradients of
+          // the solution (i.e., the electric field $\mathbf E$) at these
+          // locations and loop over the individual particles:
+          evaluator.evaluate(cell,
+                             particle_positions,
+                             make_array_view(solution_values),
+                             EvaluationFlags::gradients);
 
           {
             typename Particles::ParticleHandler<dim>::particle_iterator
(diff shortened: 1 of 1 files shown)


kronbichler:
Yes, this was one of the intended cases for this class. I can add more description, but not this week. We should also decide if we want to address https://github.com/dealii/dealii/pull/11046#issuecomment-724029046 in terms of interface (which I believe we should).

examples/step-19/step-19.cc
```diff
@@ -662,21 +666,22 @@ namespace Step19
...
-          std::vector<Tensor<1, dim>> field_gradients(
-            quadrature_formula.size());
-          particle_position_fe_values.get_function_gradients(solution,
-                                                             field_gradients);
+          field_values.resize(particle_positions.size());
+          field_gradients.resize(particle_positions.size());
+
+          cell->get_dof_values(solution,
+                               solution_values.begin(),
+                               solution_values.end());
```
bangerth, line 674: You should be able to write this as `get_dof_values (solution, solution_values)` if you use `Vector<double>` for `solution_values`.

examples/step-19/step-19.cc
```diff
@@ -662,21 +666,22 @@ namespace Step19
...
+
+          // Then we can ask the FEPointEvaluation object for the gradients of
+          // the solution (i.e., the electric field $\mathbf E$) at these
+          // locations and loop over the individual particles:
+          FEPointEvaluation<1, dim> evaluator(mapping, fe);
+          evaluator.evaluate(cell,
+                             particle_positions,
+                             solution_values,
```
bangerth, line 682: And that would suggest that what this function should take is a `Vector` as third argument.

examples/step-19/step-19.cc
```diff
@@ -650,6 +651,9 @@ namespace Step19
   {
     const double dt = time.get_next_step_size();
 
+    std::vector<double>         solution_values(fe.n_dofs_per_cell());
+    std::vector<double>         field_values;
```
bangerth, line 655: We don't actually need `field_values` at all. Maybe call it `dummy_field_values` and not resize it at all?

peterrum: @kronbichler and I had a short discussion on how to improve the interface of the new class (see https://github.com/dealii/dealii/pull/11046#issuecomment-753946564); as a consequence these lines will changes. By the time of merging of the PR, we knew that the interface was was not final but we intended to improve it after we made some experiences during usage in actual application codes. I hope I have time tomorrow to work on the redesign of the interface.

kronbichler (approved):

peterrum:
> I hope I have time tomorrow to work on the redesign of the interface.

OK. This PR is now finally ready. Instead of one day it took 3 weeks. Let's call it an academic day!

examples/step-19/step-19.cc
```diff
@@ -650,6 +651,8 @@ namespace Step19
   {
     const double dt = time.get_next_step_size();
 
+    std::vector<double>       solution_values(fe.n_dofs_per_cell());
+    FEPointEvaluation<1, dim> evaluator(mapping, fe);
```
kronbichler, line 655: Can you change this to the new interface we have on master? Then this PR should be good to be merged.

peterrum: I am a bit confused. Isn't this the new interface? It already uses `EvaluationFlags::gradients`.

peterrum: I guess you mean that I should rebase so that the changes to include/deal.II/fe/fe_point_evaluation.h are not visible?

kronbichler: Sorry, yes, I misread. I was thinking about the changes in the interface for the evaluation.

examples/step-19/step-19.cc
bangerth: This is still unaddressed.

examples/step-19/step-19.cc
```diff
@@ -662,21 +665,17 @@ namespace Step19
           for (const auto &particle : particles_in_cell)
             particle_positions.push_back(particle.get_reference_location());
 
-          const Quadrature<dim> quadrature_formula(particle_positions);
-          FEValues<dim>         particle_position_fe_values(mapping,
```
bangerth, line 666: You now no longer use the `FEValues` object, but the documentation of the function continues to describe this approach. Can you update?

bangerth (commented):
The documentation of the `move_particles()` function needs to be adjusted now.

examples/step-19/step-19.cc
peterrum: Changed. But the disadvantage is that now we nee to use `make_array_view`.

examples/step-19/step-19.cc
peterrum: Done!

examples/step-19/step-19.cc
kronbichler: If we are concerned about it we could fix things by adding another constructor to `ArrayView` or, alternatively, let the user pass in a `Vector`. In `DoFAccessor` we are at least consistent with the latter, so it would not feel wrong to do it here as well, even though some uses of `Vector` vs `std::vector` in `FEValues` are confusing in terms of interface consistency.

peterrum: >  adding another constructor to ArrayView

This is not working. Because some static asserts due to:
```
/home/munch/sw-index/dealii/include/deal.II/lac/vector.h:115:5: error: static assertion failed: The Vector class does not support auto-differentiable numbers.
  115 |     !Differentiation::AD::is_ad_number<Number>::value,
```

> , alternatively, let the user pass in a Vector

means that `FEPointEvaluation` would get additional `evaluate`/`integrate` functions!?

peterrum: > If we are concerned about it

And yes, I don't like it ;)

kronbichler: Out of interest, where does the instantiation of `AD` numbers come from that the class would complain about?

bangerth (approved):

Issue #10715, opened 2020-07-18
step-26: possible issue with some missing variable initialization


cpraveen:
My course student Arpit Babbar raised a possible issue in step-26. In this example, we do some grid refinements in the initial time step to get an acceptable initial mesh.

After this line

https://github.com/dealii/dealii/blob/master/examples/step-26/step-26.cc#L458

we should set
```
time = 0;
timestep_number = 0;
```

He says this was done in the video lecture of Wolfgang but are missing in the current code.

In fact the code does the pre refinement only once
```
$ ./step-26

===========================================
Number of active cells: 48
Number of degrees of freedom: 65

Time step 1 at t=0.002
     7 CG iterations.

===========================================
Number of active cells: 60
Number of degrees of freedom: 81


Time step 2 at t=0.004
     7 CG iterations.
Time step 3 at t=0.006
     6 CG iterations.
Time step 4 at t=0.008
     6 CG iterations.
Time step 5 at t=0.01
     6 CG iterations.
```
but it is supposed to do it n_adaptive_pre_refinement_steps = 4 times.

Also these two lines seem redundant since reinit is done after goto also

https://github.com/dealii/dealii/blob/master/examples/step-26/step-26.cc#L572
https://github.com/dealii/dealii/blob/master/examples/step-26/step-26.cc#L573

bangerth:
Yes, definitely true! Would you want to encourage your student to make a pull request out of his observation?

bangerth:
By the way, this was broken in #4083 (@drwells :-) ), nearly 3 years ago.

drwells:
Ooops :)

cpraveen:
Do you want to also remove the goto ? It does not cause much code duplication, see here

https://gist.github.com/cpraveen/2d6fdade5404d3c9d1fb8b0c125bb375

Changes are in lines
317, 397-406

cpraveen:
Fixed in https://github.com/dealii/dealii/pull/11460

PR #11460, opened 2021-01-04
step-26: Reset time and timestep_number during pre refinement steps

cpraveen:
Otherwise, the pre-refinement is performed only once. See https://github.com/dealii/dealii/issues/10715


bangerth (approved):

bangerth:
Nice! Would you want to add an entry to the changelog by putting a file into `doc/news/changes/minor` using the same format as the existing files there?

bangerth (approved):

peterrum (approved):

PR #11461, opened 2021-01-04
Rename zero_out_ghosts() to zero_out_ghost_values()

peterrum:
The asymmetric of `update_ghost_values()` and `zero_out_ghosts()` only leads to confusion. I would suggest to also rename `has_ghost_elements()` to `has_ghost_values()`.

FYI @mschreter

### DIFF
--- a/include/deal.II/lac/la_parallel_block_vector.h
+++ b/include/deal.II/lac/la_parallel_block_vector.h
@@ -348,9 +348,23 @@ namespace LinearAlgebra
        * After calling this method, read access to ghost elements of the
        * vector is forbidden and an exception is thrown. Only write access to
        * ghost elements is allowed in this state.
+       *
+       * @deprecated Use zero_out_ghost_values() instead.
+       */
+      DEAL_II_DEPRECATED_EARLY void
+      zero_out_ghosts() const;
+
+
+      /**
+       * This method zeros the entries on ghost dofs, but does not touch
+       * locally owned DoFs.
+       *
+       * After calling this method, read access to ghost elements of the
+       * vector is forbidden and an exception is thrown. Only write access to
+       * ghost elements is allowed in this state.
        */
       void
-      zero_out_ghosts() const;
+      zero_out_ghost_values() const;
 
       /**
        * Return if this Vector contains ghost elements.
--- a/include/deal.II/lac/la_parallel_vector.h
+++ b/include/deal.II/lac/la_parallel_vector.h
@@ -134,7 +134,7 @@ namespace LinearAlgebra
      * This vector can take two different states with respect to ghost
      * elements:
      * <ul>
-     * <li> After creation and whenever zero_out_ghosts() is called (or
+     * <li> After creation and whenever zero_out_ghost_values() is called (or
      * <code>operator= (0.)</code>), the vector does only allow writing into
      * ghost elements but not reading from ghost elements.
      * <li> After a call to update_ghost_values(), the vector does not allow
@@ -515,7 +515,7 @@ namespace LinearAlgebra
        * operation. However, global reduction operations like norms or the
        * inner product will always ignore ghost elements in order to avoid
        * counting the ghost data more than once. To allow writing to ghost
-       * elements again, call zero_out_ghosts().
+       * elements again, call zero_out_ghost_values().
        *
        * @see
        * @ref GlossGhostedVector "vectors with ghost elements"
@@ -605,9 +605,22 @@ namespace LinearAlgebra
        * After calling this method, read access to ghost elements of the
        * vector is forbidden and an exception is thrown. Only write access to
        * ghost elements is allowed in this state.
+       *
+       * @deprecated Use zero_out_ghost_values() instead.
+       */
+      DEAL_II_DEPRECATED_EARLY void
+      zero_out_ghosts() const;
+
+      /**
+       * This method zeros the entries on ghost dofs, but does not touch
+       * locally owned DoFs.
+       *
+       * After calling this method, read access to ghost elements of the
+       * vector is forbidden and an exception is thrown. Only write access to
+       * ghost elements is allowed in this state.
        */
       void
-      zero_out_ghosts() const;
+      zero_out_ghost_values() const;
 
       /**
        * Return whether the vector currently is in a state where ghost values
--- a/doc/news/changes/incompatibilities/20210104Munch
+++ b/doc/news/changes/incompatibilities/20210104Munch
@@ -1,0 +1,4 @@
+Deprecated: All `LinearAlgebra::distributed::Vector::zero_out_ghosts()` has been deprecated. 
+Use `LinearAlgebra::distributed::Vector::zero_out_ghost_values()` instead.
+<br>
+(Peter Munch, 2021/01/04)
--- a/examples/step-48/step-48.cc
+++ b/examples/step-48/step-48.cc
@@ -508,7 +508,7 @@ namespace Step48
     data_out.write_vtu_with_pvtu_record(
       "./", "solution", timestep_number, MPI_COMM_WORLD, 3);
 
-    solution.zero_out_ghosts();
+    solution.zero_out_ghost_values();
   }
 
 
--- a/examples/step-67/step-67.cc
+++ b/examples/step-67/step-67.cc
@@ -1544,7 +1544,7 @@ namespace Euler_DG
     FEEvaluation<dim, degree, degree + 1, dim + 2, Number> phi(data, 0, 1);
     MatrixFreeOperators::CellwiseInverseMassMatrix<dim, degree, dim + 2, Number>
       inverse(phi);
-    solution.zero_out_ghosts();
+    solution.zero_out_ghost_values();
     for (unsigned int cell = 0; cell < data.n_cell_batches(); ++cell)
       {
         phi.reinit(cell);
--- a/examples/step-68/step-68.cc
+++ b/examples/step-68/step-68.cc
@@ -502,7 +502,7 @@ namespace Step68
   template <int dim>
   void ParticleTracking<dim>::interpolate_function_to_field()
   {
-    velocity_field.zero_out_ghosts();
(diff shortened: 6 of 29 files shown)


include/deal.II/lac/la_parallel_block_vector.h
```diff
@@ -348,10 +348,24 @@ namespace LinearAlgebra
        * After calling this method, read access to ghost elements of the
        * vector is forbidden and an exception is thrown. Only write access to
        * ghost elements is allowed in this state.
+       *
+       * @deprecated Use zero_out_ghost_values() instead.
        */
-      void
+      DEAL_II_DEPRECATED void
```
kronbichler, line 354: I agree with the general PR and with `@deprecated` in the doxygen comment, but I am very unhappy about adding the `DEAL_II_DEPRECATED` marker for now.

We've had this discussion before, but I want to once again point out the pain this causes in downstream projects. This change raises deprecation warnings from day one without the users having any time to use a different interface, so we force all downstream projects without very tight control on the deal.II version to be used `#ifdef` markers. For more commonly used functions like the present one, I suggest to have one release with both interfaces (and the preferred one in tutorials), one release with the deprecation warning, and then finally remove the old interface two releases ahead.

I want to contrast this to what I consider a gentle deprecation process, e.g. the `MatrixFree` class where we now deprecated `n_macro_cells` but we had at least one previous release with what we suggest as alternative. This still forces users to eventually upgrade and gives us a change to eventually fix an interface we do not like, but without being what I consider as explicitly rude against our users.

peterrum: > I agree with the general PR and with @deprecated in the doxygen comment, but I am very unhappy about adding the DEAL_II_DEPRECATED marker for now.

include/deal.II/lac/la_parallel_block_vector.h
```diff
@@ -348,10 +348,24 @@ namespace LinearAlgebra
...
       void
       zero_out_ghosts() const;
 
+
+      /**
+       * This method zeros the entries on ghost dofs, but does not touch
+       * locally owned DoFs.
+       *
+       * After calling this method, read access to ghost elements of the
+       * vector is forbidden and an exception is thrown. Only write access to
```
bangerth, line 363: An exception is thrown *when*? I think what you say is that an exception is thrown if one tries to read from these values, but that's not clear from this one sentence (though the next clarifies it).

It's an interesting concept, though. Do any other functions impose this kind of restriction? When is the restriction lifted?

include/deal.II/lac/la_parallel_block_vector.h
```diff
@@ -348,10 +348,24 @@ namespace LinearAlgebra
        * After calling this method, read access to ghost elements of the
        * vector is forbidden and an exception is thrown. Only write access to
        * ghost elements is allowed in this state.
+       *
+       * @deprecated Use zero_out_ghost_values() instead.
        */
       void
       zero_out_ghosts() const;
```
bangerth, line 355: Can you mark up this function with `DEAL_II_DEPRECATED`?

include/deal.II/lac/la_parallel_vector.h
```diff
@@ -605,10 +605,23 @@ namespace LinearAlgebra
        * After calling this method, read access to ghost elements of the
        * vector is forbidden and an exception is thrown. Only write access to
        * ghost elements is allowed in this state.
+       *
+       * @deprecated Use zero_out_ghost_values() instead.
        */
       void
       zero_out_ghosts() const;
```
bangerth, line 612: same here

include/deal.II/lac/la_parallel_vector.h
```diff
@@ -605,10 +605,23 @@ namespace LinearAlgebra
...
 
+      /**
+       * This method zeros the entries on ghost dofs, but does not touch
+       * locally owned DoFs.
```
bangerth, line 616: same comments about the documentation of this function

include/deal.II/lac/la_parallel_block_vector.h
bangerth: Ah, I'm only now reading @kronbichler 's comment.

How do we track these things? If we don't `DEAL_II_DEPRECATE` these functions right away, then we should at least have an issue somewhere that collects all of those symbols we'd like to deprecate after the next release...

masterleinad: Yes, also `contrib/utilities/print_deprecated.py` always searches for DEAL_II_DEPRECATED.
I don't think we really came to a conclusion regarding our policy to not deprecate immediately...
I still don't see a problem if codes that use `upstream` just always disable deprecation warnings if they don't want to deal with them.

kronbichler: I am probably biased on this topic because I see the cost of taking care of these warnings in maybe half a dozen projects and I work with upstream. But it is not just upstream (how large is the percentage of users on master vs  a release, by the way?). I think many projects gravitate around two subsequent releases of deal.II. For me it is the latest release and master, but it does not really change much from a user perspective if you work with say the 9.2 and 9.1 releases. If you do not want warnings that came up in 9.2 and where 9.1 did not yet provide an alternative, you need all users of a project using deal.II to update at the same time, unless you ifdef by version. For small functions, ifdef is fine, but for more widely used interfaces it is a pain. So to me the important part is that we have one release with the new alternative and without the warning. Note that I am not opposed to eventually deprecate, but a distance of two releases is much easier to handle, as I see no problem to force users not being more than one release behind whatever project XY sees as the most recent deal.II version.

To me, the question is what we think makes best value out of the deprecation warnings. If we are too aggressive, we force people to either #ifdef by version, or to disable the warnings in all their builds and only be told later by an error (in which case we do not need the warnings). I think both alternatives have a good potential to frustrate users. If we are too soft, we have a maintainability problem, and this is not what I want. That's why I am in favor of an intermediate level for *central functionality* that has the potential to raise dozens of warnings (and make users disable the warning, rather than fix it). Of course, what is *central* is subjective and again puts a maintainability burden on us. I see me doing less work on keeping things up in deal.II than in the user projects I have on deal.II, but I realize that you @masterleinad have done much more cleaning work than me, so I admit that I'm not in a neutral position on this point.

As it seems I am the only person arguing this way, I assume that it is my own work load and not a general problem, so I better keep silent on this now.

peterrum: @kronbichler I see your point and agree with you concerns (I might have been too quick in deprecating since I got the impression that that is the normal: `DoFHandler::initialize()` , `VectorizedArray::n_vector_entries`). I think we were considering a version number of the `DEAL_II_DEPRECATED` macro. Maybe it is a good time to refresh the discussion there. I have a (maybe naive) implementation which might work (and could be configured at compile time via `DEAL_II_DEPRECATED_VERBOSITY`):

```cpp
#define DEAL_II_DEPRECATED_VERBOSITY 0
#define DEAL_II_DEPRECATED "[[deprecated]]"

#define CAT(a, ...) PRIMITIVE_CAT(a, __VA_ARGS__)
#define PRIMITIVE_CAT(a, ...) a ## __VA_ARGS__
#define DEAL_II_DEPRECATED_VERSION(major,minor) PRIMITIVE_CAT(DEAL_II_DEPRECATED_VERSION_, major ## _ ## minor)

#if DEAL_II_DEPRECATED_VERBOSE
  #define DEAL_II_DEPRECATED_VERSION_9_3 DEAL_II_DEPRECATED
#else
  #define DEAL_II_DEPRECATED_VERSION_9_3 ""
#endif

#define DEAL_II_DEPRECATED_VERSION_9_2 DEAL_II_DEPRECATED

#include <iostream>

int main()
{
    std::cout << DEAL_II_DEPRECATED_VERSION(9,2) << std::endl; // [[deprecated]]
    std::cout << DEAL_II_DEPRECATED_VERSION(9,3) << std::endl; //
}
```

... this was the best I could come up with the help of https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms. I would have loved if I didn't need to write out all version but could use the logic of:
```
#  define DEAL_II_MPI_VERSION_GTE(major,minor) \
 ((DEAL_II_MPI_VERSION_MAJOR * 100 + DEAL_II_MPI_VERSION_MINOR) >= (major)*100 + (minor))
```

drwells: We need to support projects where people install "deal.II master as of two weeks ago" and "deal.II master as of two months ago" without printing thousands of deprecation warnings so I agree with @kronbichler's general point (its why I closed #10335).

I support doing versioned deprecations. It wouldn't be that hard to have another level of deprecation that can be turned on or off that we promote to `DEAL_II_DEPRECATED` the moment we increment the version number.

masterleinad: I would be fine with defining a `DEAL_II_DEPRECATED_EARLY` macro that we promote to `DEAL_II_DEPRECATED` with the next release.

masterleinad: Maybe something like https://github.com/dealii/dealii/pull/11472?

peterrum:
I have updated this PR so that it used `DEAL_II_DEPRECATED_EARLY`. Thank you @masterleinad!

masterleinad:
```
/jenkins/workspace/dealii_PR-11461/tests/simplex/compute_point_locations_01.cc: In function ‘void test_in_unit_cube(const std::vector<dealii::Point<dim> >&)’:
/jenkins/workspace/dealii_PR-11461/tests/simplex/compute_point_locations_01.cc:37:3: error: ‘GridGenerator’ has not been declared
   GridGenerator::subdivided_hyper_cube_with_simplices(tria, 1);
   ^
```

peterrum:
@masterleinad That has been fixed on master a couple of days ago. Should I rebase this PR nevertheless?

kronbichler (approved):

masterleinad (approved):

PR #11462, opened 2021-01-04
Return points by value.

bangerth:
For the purposes of returned arguments, we frequently return `Point<dim>` objects by-value rather than `const` reference. That's because they are small and easily scalarized by compilers. Do the same for the functions that return locations in the particle interfaces.

See #11206.

/rebuild

### DIFF
--- a/include/deal.II/particles/particle.h
+++ b/include/deal.II/particles/particle.h
@@ -280,7 +280,7 @@ namespace Particles
      *
      * @return The location of this particle.
      */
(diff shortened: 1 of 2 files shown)


peterrum (approved):

masterleinad (approved):

bangerth:
I don't know where my brain is. I had already proposed this as #11390 and concluded that that doesn't work...

PR #10932, opened 2020-09-17
Step 74: Symmetric interior penalty Galerkin (SIPG) method for Poisson's equation

zjiaqi2018:
In this tutorial, we display the usage of the FEInterfaceValues class, which is designed for assembling face terms arising from discontinuous Galerkin (DG) methods.

PR #11463, opened 2021-01-04
Minor changes to step-74

peterrum:


peterrum:
> Actually, never mind -- I'll just include your changes into my follow-up PR.

Too late ;)

bangerth (approved):

bangerth:
Follow-up to #10932

examples/step-74/step-74.cc
```diff
@@ -202,7 +202,7 @@ namespace Step74
                          const double       cell_extend_left,
                          const double       cell_extend_right)
   {
-    const double degree = std::max(1., static_cast<double>(fe_degree));
+    const double degree = std::max<double>(1, fe_degree);
```
tjhei, line 205: I am not a big fan of this change: It makes it less obvious that a conversion int->double is happening and spells out "double" twice.

bangerth: I changed that further in #11467. Let's not worry about it here.

peterrum:
To me it looks like that #11467 integrated most of my comments. So I will close this PR.

PR #11468, opened 2021-01-05
Better explain the purpose of the L-singularity function.

bangerth:


Rombur (approved):

include/deal.II/base/function_lib.h
```diff
@@ -381,12 +381,27 @@ namespace Functions
 
 
   /**
-   * Harmonic singularity on the L-shaped domain in 2D.
+   * A function that solves the Laplace equation (with specific
+   * boundary values but zero right hand side) and that has a
+   * singularity on the L-shaped domain in 2D
```
tjhei, line 386:
```suggestion
   * singularity in the center of the L-shaped domain in 2D.
```

include/deal.II/base/function_lib.h
```diff
@@ -381,12 +381,27 @@ namespace Functions
...
    * \sin(\frac{2}{3} \phi)$ with a singularity at the origin and should be
    * used with GridGenerator::hyper_L().
    *
+   * This function is often used to illustrate that the solutions of the Laplace
+   * equation
+   * @f[
+   *   -\Delta u = 0
+   * @f]
+   * can be singular even if the boundary values are smooth. (Here, if the
+   * domain is the L-shaped domain $(-1,1)^2 \backslash [0,1]\times [-1,0]$, the
```
tjhei, line 398: I think this is not correct. The part [0,1]x[0,1] is missing.

bangerth:
@tjhei I needed to look this up. The reason why I wrote the choice of the L-shaped domain this way is because the description says that the function involves the term `sin(2/3 phi)`, which would suggest that the function is zero along the positive x-axis and the *negative* y-axis. But it turns out that the implementation computes `phi` in *clockwise* direction against the positive x-axis, and so you are right. I now also document the definition of `phi` in the class declaration.

tjhei (approved):

PR #11477, opened 2021-01-05
step-74 fix: energy norm description

zjiaqi2018:
Add the description of the energy norm in step-74.


bangerth (approved):
I assume that you checked that not having `nu` in the jump terms is the correct approach.

zjiaqi2018:
@bangerth  Yes, not having \nu in the jump terms is the correct approach.

PR #11481, opened 2021-01-06
Add VectorRepartitioner

peterrum:
This PR adds a transfer class that transfers data between vectors setup for `Triangulation` classes that only differ in their partitioning. The implementation practically misuses some of the setup routines of the global-coarsening algorithm. I am not sure regarding the names.

Possibilities for extension:
- asynchronous communication
- second `reinit()`-function taking user provided `extended_partitioner` and `indices`
- different communicators

*This class was the result of the last workshop and was only lying around since than...*


drwells:
Having an equivalent class to PETSc's `VecScatter` would be nice - I need this for a tentative project I am working on.

drwells:
I do have my own WIP version of this class, for the sake of record:

https://github.com/drwells/fiddle/blob/master/include/fiddle/transfer/scatter.h

I have an unusual indexing requirement in that I want to essentially scatter (either add or insert) values that are in a huge ghost region back to their actual owners so this might not be the right solution. Nevertheless if we add this class we should have `_start()` and `_finish()` routines taking channel numbers as arguments, like in `LA::d::V`.

include/deal.II/numerics/vector_repartitioner.h
```diff
@@ -0,0 +1,100 @@
...
+#include <deal.II/dofs/dof_handler.h>
+
+#include <deal.II/lac/la_parallel_vector.h>
+
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * Class for repartitioning data living on the same triangulation, which
+ * has been partitioned differently.
```
tamiko, line 33: Would you mind writing out a detailed usage example in the class documentation?

include/deal.II/numerics/vector_repartitioner.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+template <typename VectorType>
+class VectorRepartitioner
```
tamiko, line 36: I agree that we could find some better names for the class and the operations. `VectorRepartitioner` would intuitively imply that the class takes a vector and modifies it. Instead, however, the class provides functionality to translate from one parallel partition to another one...

include/deal.II/numerics/vector_repartitioner.h
```diff
@@ -0,0 +1,100 @@
...
+{
+public:
+  /**
+   * Transfer data.
+   */
+  void
+  update_forwards(VectorType &dst, const VectorType &src) const;
```
tamiko, line 43: I don't like the "forward" and "backward" in the function name because there isn't really a notion of forward/backward. You use source and destination later on in the constructor. Maybe a name consistent with source/destination?

tamiko (changes requested):
Some quick comments on documentation and naming

drwells:
Do we want to get this done by 9.3?

PR #11484, opened 2021-01-06
Tutorial step-01 with simplices support.

eliasstudiert:


peterrum (approved):

drwells:
Do we expect manifolds to work with simplices at this point? The second grid out doesn't look like we used it but the test sets it up.

test

peterrum:
@drwells #11389 ;)

tests/simplex/step-01.cc
```diff
@@ -0,0 +1,105 @@
...
+
+  const Point<2> center(1, 0);
+  const double   inner_radius = 0.5, outer_radius = 1.0;
+  GridGenerator::hyper_shell(tria_temp, center, inner_radius, outer_radius, 10);
+
+  // convert mesh to a simplex mesh and copy manually the attached manifolds
+  GridGenerator::convert_hypercube_to_simplex_mesh(tria_temp, triangulation);
+  for (const auto i : tria_temp.get_manifold_ids())
+    if (i != numbers::flat_manifold_id)
+      triangulation.set_manifold(i, tria_temp.get_manifold(i));
```
drwells, line 67: I did see #11389 but this code in particular makes it look like this should work with manifolds right now.

drwells: Oh, I see - we copy the manifold but not the manifold id. Got it!

drwells:
If we start copying manifold ids then the output of this test will change so lets not merge this until #11389 is in.

tests/simplex/step-01.cc
peterrum: Yeah. This is an issue that I really don't like, since people (in particular me) tend to forget to copy the actual manifolds.

peterrum:
Looks good now!

mesh svg

drwells (approved):

PR #11486, opened 2021-01-06
Fix bug in DoFHandler::distribute_dofs

### LINKED ISSUE #11485: Bug in DoFHandler::distribute_dofs
The following code will cause an assertion failure in debug mode.

```
#include <deal.II/fe/fe_nedelec.h>


int main() {
  dealii::Triangulation<3> tria;
  dealii::GridGenerator::hyper_cube(tria, -1, 1);

  dealii::FE_Nedelec<3> fe_nedelec(0);

  dealii::DoFHandler<3> dh_nedelec;

  dh_nedelec.reinit(tria);
  dh_nedelec.distribute_dofs(fe_nedelec);

  dh_nedelec.reinit(tria); // The error will disappear without this line
  dh_nedelec.distribute_dofs(fe_nedelec);

  for (auto cell = dh_nedelec.begin_active(); cell != dh_nedelec.end(); ++cell) {
    for (unsigned int i = 0; i < dealii::GeometryInfo<3>::lines_per_cell; ++i) {
      auto line_index = cell->line(i)->dof_index(0);
      (void)line_index;
    }
  }

  return 0;
}
```

```
An error occurred in line <296> of file <.../include/deal.II/dofs/dof_accessor.templates.h> in function
    static void dealii::internal::DoFAccessorImplementation::Implementation::process_dof_index(const DoFHandler<dim, spacedim> &, const unsigned int, const unsigned int, const unsigned int, const unsigned int, const std::integral_constant<int, d> &, GlobalIndexType &, const DoFPProcessor &) [dim = 3, spacedim = 3, d = 1, GlobalIndexType = unsigned int, DoFPProcessor = (lambda at /.../include/deal.II/dofs/dof_accessor.templates.h:535:27)]
The violated condition was: 
    ptr != dof_handler.hp_object_fe_indices[d].begin() + dof_handler.hp_object_fe_ptr[d][obj_index + 1]
Additional information: 
    You are trying to use functionality in deal.II that is currently not implemented. In many cases, this indicates that there simply didn't appear much of a need for it, or that the author of the original code did not have the time to implement a particular case. If you hit this exception, it is therefore worth the time to look into the code to find out whether you may be able to implement the missing functionality. If you do, please consider providing a patch to the deal.II development sources (see the deal.II website on how to contribute).

Stacktrace:
-----------
#0  2   debug_dof_index                     0x000000010ec6de35 _ZN6dealii8internal25DoFAccessorImplementation14Implementation17process_dof_indexILi3ELi3ELi1EjZNS2_13get_dof_indexILi3ELi3ELi1EEEjRKNS_10DoFHandlerIXT_EXT0_EEEjjjjRKNSt3__117integral_constantIiXT1_EEEEUlRKT_RT0_E_EEvS8_jjjjSD_RT2_RKT3_ + 1189: 2   debug_dof_index                     0x000000010ec6de35 _ZN6dealii8internal25DoFAccessorImplementation14Implementation17process_dof_indexILi3ELi3ELi1EjZNS2_13get_dof_indexILi3ELi3ELi1EEEjRKNS_10DoFHandlerIXT_EXT0_EEEjjjjRKNSt3__117integral_constantIiXT1_EEEEUlRKT_RT0_E_EEvS8_jjjjSD_RT2_RKT3_ 
#1  3   debug_dof_index                     0x000000010ec6d968 _ZN6dealii8internal25DoFAccessorImplementation14Implementation13get_dof_indexILi3ELi3ELi1EEEjRKNS_10DoFHandlerIXT_EXT0_EEEjjjjRKNSt3__117integral_constantIiXT1_EEE + 24: 3   debug_dof_index                     0x000000010ec6d968 _ZN6dealii8internal25DoFAccessorImplementation14Implementation13get_dof_indexILi3ELi3ELi1EEEjRKNS_10DoFHandlerIXT_EXT0_EEEjjjjRKNSt3__117integral_constantIiXT1_EEE 
#2  4   debug_dof_index                     0x000000010ec597a8 _ZNK6dealii11DoFAccessorILi1ELi3ELi3ELb0EE9dof_indexEjj + 72: 4   debug_dof_index                     0x000000010ec597a8 _ZNK6dealii11DoFAccessorILi1ELi3ELi3ELb0EE9dof_indexEjj 
#3  5   debug_dof_index                     0x000000010ec5964c main + 300: 5   debug_dof_index                     0x000000010ec5964c main 
#4  6   libdyld.dylib                       0x00007fff6fd24cc9 start + 1: 6   libdyld.dylib                       0x00007fff6fd24cc9 start 
```
I guess the problem is related with variable `hp_capability_enabled`. In `reinit`, we set it to `true`. But if we call `distribute_dofs` with the same `fe`, the following code won't get called, result the `DoFHandler` object in an inconsistent state.
https://github.com/dealii/dealii/blob/622107e2928bbe96110393da1c5bd986ba9b707b/source/dofs/dof_handler.cc#L2499-L2530

If I am right I will create a PR to fix it. @bangerth @peterrum @marcfehling

adamqc:
Fixes #11485


marcfehling:
Thank you for reporting this bug. This is indeed not intended behavior: we forgot to reset the finite element collection in `reinit()`.

Adding the following line to `reinit()` should do the trick. I would prefer this as this is the root of the cause. What do you think?
```
this->fe_collection = hp::FECollection<dim, spacedim>();
```

Thank you for taking care of this!

tjhei:
> Adding the following line to `reinit()` should do the trick

If that makes the test pass, then this is the better choice.

tests/dofs/distribute_dofs_01.cc
```diff
@@ -0,0 +1,55 @@
...
+  dh_nedelec.reinit(tria);
+  dh_nedelec.distribute_dofs(fe_nedelec);
+
+  dh_nedelec.reinit(tria);
+  dh_nedelec.distribute_dofs(fe_nedelec);
+
+  try
+    {
+      for (auto cell = dh_nedelec.begin_active(); cell != dh_nedelec.end();
+           ++cell)
```
masterleinad, line 31:
```suggestion
      for (const auto& cell: dh_nedelec.active_cell_iterators())
```

tests/dofs/distribute_dofs_01.cc
```diff
@@ -0,0 +1,55 @@
...
+            {
+              auto line_index = cell->line(i)->dof_index(0);
+              (void)line_index;
+            }
+        }
+
+      deallog << "ok" << std::endl;
+    }
+  catch (...)
+    {}
```
masterleinad, line 44: Why are you ignoring any exception thrown here? I would just remove the whole try-catch block.

adamqc: The intention I use try-catch block is to avoid the test run into assertion failure. If the test passes, it will print `ok`, otherwise nothing will be printed. Does this make sense?

marcfehling: `ctest` handles this automatically for you: if an assertion is triggered, the test will be considered as `failed`. You can safely remove the `try/catch` block here.

adamqc:
@marcfehling @tjhei @masterleinad Thanks for the suggestions. I modified the PR, please have a look.

tests/dofs/distribute_dofs_01.cc
```diff
@@ -0,0 +1,58 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2010 - 2021 by the deal.II authors
```
marcfehling, line 3: It is sufficient to say 2021 here. :)

marcfehling (approved):
Thank you for the change and again for your effort!

masterleinad (approved):

tjhei (approved):

tests/dofs/distribute_dofs_01.cc
tjhei: True, but I think we don't have to be too strict about it here. :-)

tjhei:
Thanks again for the fix!

PR #11490, opened 2021-01-07
[WIP] Enabling global refinement for TET meshes

peterrum:


tjhei:
There is many different ways to do refinement of a triangle: connecting line midpoints, barycentric refinement, red/green refinement, and probably more. I assume you are connecting midpoints here? There are additional options in 3d of course.
How do you want to make this selection? Do we want to extend the ``RefinementCase`` class? This would be worth thinking about before we sell one of the options as "refinement".

peterrum:
Global refinement of TET-meshes should work now :)

source/grid/tria.cc
```diff
@@ -5177,6 +5179,1237 @@ namespace internal
...
+
+        for (typename Triangulation<dim, spacedim>::quad_iterator quad =
+               triangulation.begin_quad();
+             quad != triangulation.end_quad();
+             ++quad)
+          {
+            if (quad->user_flag_set() == false)
+              continue;
+
+            if (quad->reference_cell_type() == ReferenceCell::Type::Quad)
```
drwells, line 5288: we let this sit for too long (this is partly my fault) so it's gotten stale - we will need to rebase and fix the compilation errors (e.g., this is now just `reference_cell`)

source/grid/tria.cc
```diff
@@ -5177,6 +5179,1237 @@ namespace internal
...
+            {
+              if (line->user_flag_set() == false)
+                continue;
+
+              while (triangulation.vertices_used[next_unused_vertex] == true)
+                ++next_unused_vertex;
+              Assert(
+                next_unused_vertex < triangulation.vertices.size(),
+                ExcMessage(
+                  "Internal error: During refinement, the triangulation wants to access an element of the 'vertices' array but it turns out that the array is not large enough."));
```
drwells, line 5358: Could you split this manually, even if clang-format does not?

source/grid/tria.cc
```diff
@@ -5177,6 +5179,1237 @@ namespace internal
...
+
+              line->set_children(0, next_unused_line->index());
+
+              const typename Triangulation<dim, spacedim>::raw_line_iterator
+                children[2] = {next_unused_line, ++next_unused_line};
+
+              Assert(
+                children[0]->used() == false,
+                ExcMessage(
+                  "Internal error: We want to use a cell during refinement that should be unused, but turns out not to be."));
```
drwells, line 5377: It's not a big deal, but would you mind just defining this as an exception class in `tria.h`? That way we don't have to write the same error message repeatedly.

source/grid/tria.cc
```diff
@@ -5177,6 +5179,1237 @@ namespace internal
...
+                                 {{1, 9, 10, 6}},
+                                 {{9, 3, 11, 7}}}};
+
+              static constexpr std::
+                array<std::array<std::array<unsigned int, 2>, 4>, 4>
+                  quad_line_vertices_quad{
+                    {{{{{0, 4}}, {{6, 8}}, {{0, 6}}, {{4, 8}}}},
+                     {{{{6, 8}}, {{1, 5}}, {{6, 1}}, {{8, 5}}}},
+                     {{{{4, 2}}, {{8, 7}}, {{4, 8}}, {{2, 7}}}},
+                     {{{{8, 7}}, {{5, 3}}, {{8, 5}}, {{7, 3}}}}}};
```
drwells, line 5593: Would you mind adding some ascii art here to show where these numbers come from? It's really hard to understand this without some documentation.

drwells (approved):
It's really hard to review these changes to `Triangulation` because that class is so difficult to understand to begin with. I suppose this is fine since we have tests and they pass.

The code you added has index calculations for both hexahedra and tetrahedra - is this supposed to work for mixed meshes? Do you plan on having this function eventually handle that case?

bangerth:
Could you rebase this? Also, do you want to get this patch into 9.3?

source/grid/tria.cc
peterrum: Let's keep this how it is. This and the below issue are not new and could be handled by a follow-up PR.

peterrum:
> Also, do you want to get this patch into 9.3?

Would be nice. But, I don't know if I will have actually the time to create all the ASCII drawings.

tests/simplex/refinement_03.cc
```diff
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 2021 by the deal.II authors
+// Copyright (C) 2020 by the deal.II authors
```
bangerth, line 3: I get the impression that you're overwriting an existing test here. Please just name your test differently than the one that is pre-existing.

source/grid/tria.cc
```diff
@@ -5179,6 +5181,1234 @@ namespace internal
...
+                    else
+                      Assert(false, ExcNotImplemented());
+
+                    boost::container::small_vector<unsigned int, 30>
+                      relevant_line_indices(relevant_lines.size());
+                    for (unsigned int i = 0; i < relevant_line_indices.size();
+                         ++i)
+                      relevant_line_indices[i] = relevant_lines[i]->index();
+
+                    static constexpr std::array<std::array<unsigned int, 4>, 12>
```
bangerth, line 6043: This can now be done easier using `ndarray`.

bangerth (commented):
I think it would be nice to get this patch merged at some point, but I have to admit that it is difficult to review because it adds 1280 lines to tria.cc, and to the best of my knowledge not one of these lines is a comment (I have spotted a TODO comment). It's difficult to understand the flow of the argument without comments. 

I suspect that the structure of your new function roughly matches the one used for hexes. Is there commentary that could simply be copied from the hex to the tet function? Or another way of adding commentary to help the reader?

peterrum:
Adjusting milestone.

drwells (approved):
@peterrum and I talked about this over zoom and I am basically happy with it - I read through the code and everything is in order.

I have a few cleanups that I want to do, but we should just merge this as-is.

source/grid/tria.cc
drwells: These tables are determined based on the way a Triangulation orders the subobjects - e.g., the quads are ordered in a way that is compatible with the way a Triangulation sets up lines. Hence there really is no better way to explain where these come from, so drawing pictures really won't help that much.

source/grid/tria.cc
drwells: I'll handle this in a followup.

PR #11491, opened 2021-01-07
Clean up some in function_lib.cc.

bangerth:
Specifically, add 'const' to a lot of variables. Also rename r2 to r_squared.

Follow-up to #11468.

/rebuild


masterleinad (approved):
Looks reasonable to me.

tjhei (approved):

PR #11501, opened 2021-01-10
Enabling mixed meshes in GridIn::read_vkt() and read_msh()

peterrum:


drwells (approved):
We should add "delete the old (pre 6.0) mesh numbering code" to the simplex workshop todo list so we can get rid of these warts, but this is the right workaround for now.

PR #11502, opened 2021-01-11
Assert type of coarse grid in p:d:T

mschreter:
This PR introduces asserts when simplex grid generators (```subdivided_hyper_rectangle_with_simplices``` and ```subdivided_hyper_cube_with_simplices```) are called with a ```parallel::distributed``` triangulation, which is currently not (yet) supported. This assert might be also relevant for other code locations related to the simplex support.

It is my first PR involving changes in the code, so feel free to point out general remarks, if there are some.

@peterrum: FYI

### DIFF
--- a/source/distributed/tria.cc
+++ b/source/distributed/tria.cc
@@ -2179,6 +2179,12 @@ namespace parallel
           Assert(false, ExcInternalError());
         }
 
+      Assert(
+        this->all_reference_cell_types_are_hyper_cube(),
+        ExcMessage(
+          "The class parallel::distributed::Triangulation only supports meshes "
+          "consisting only of hypercube-like cells."));
+
       // note that now we have some content in the p4est objects and call the
       // functions that do the actual work (which are dimension dependent, so
       // separate)


peterrum:
@mschreter Thank you for tackling this issue! The proposed solution would indeed fix the problem for these two `GridGenerator` functions. However, as far as I see it is not general enough. 

The cause of the issue lies in that that `parallel::distributed::Triangulation()` (and within these function the p4est functions) is called for cells that are not of hypercube type. Since this function can be called from different places in the library, I would suggest to add the assert here. To be precise, you could add after the following lines: https://github.com/dealii/dealii/blob/03470baa22917d5625112202cc9c2259b6aa1646/source/distributed/tria.cc#L2168-L2180

the lines:
```cpp
Assert(this->all_reference_cell_types_are_hyper_cube(), "The class parallel::distributed::Triangulation only supports meshes consisting only of hypercube cells.")
```

peterrum:
Regarding `p:f:T`: it is supporting simplex meshes. The issue here - what might have caused some confusion - is that the workflow there is slightly different (it uses another version of `create_triangulation()` that is accepting a partitioned construction data).

peterrum (approved):
May I ask you to squash the commits. Thx.

peterrum (approved):

bangerth:
@mschreter Nice work, and welcome to the list of deal.II authors!

bangerth:
https://dealii.org/authors.html, number 221 :-)

mschreter:
@bangerth Thank you :-).

PR #11503, opened 2021-01-11
Convert step-31 to a simplex test

peterrum:
step-31


drwells (approved):

drwells:
This looks good but the output file needs to be updated.

bangerth:
@peterrum Can you please rebase and see whether that fixes the problem?

PR #11505, opened 2021-01-12
Add FE_P_Bubbles

drwells:
This is my project for the hackathon. 

The motivation behind this element is that the quadrature rule associated with using TRI6 or TET10 as a basis has weights of zeroes at the vertices - i.e., it isn't possible to use these elements with nodal quadrature. The root cause of this is the fact that Simpson's rule on simplices should include the centroid, but TRI6 and TET10 don't use that as a support point. The new element defined here does use that as a support point - thus it works with the generalization of Simpson's rule and works well with mass lumping.

It works in 2D (TRI7) - it should be easy to extend it to also work in 3D (~TET11~ TET15). There is also a generalization to cubics that I would like to implement (but its a lot harder than the quadratic version).

Some things to do:
- [x] Finish implementing the finite element for P1 and P2. What values should be set for RIAF and other not-as-well-documented FE features?
- [x] ~Implement the cubic version.~ Skip for now - the quadratic one was hard enough in 3D.
- [x] It would be nice if we had a way to get the nodal quadrature formula out of a finite element automatically. Should we add this? Lets skip this for now - we need to add sufficient checks and answer questions about generalized support points to implement this correctly. It's outside the scope of this PR.
- [x] TET11 is a third-order element (the approximation is like TET10) but the basis functions are cubic. What is the best way to store a 'default' quadrature forumula for this element?
- [x] We need some higher-order quadrature rules in the library to get this thing working for consistent formulations.
- [x] The current polynomial implementation is messy. Is there a better way to do this? Should the base class functions all be implemented and throw exceptions? That would save a lot of typing.
- [x] citations and documentation

This is all based on the paper

Higher Order Triangular Finite Elements With Mass Lumping For The Wave Equation

Siam J. Numer. Anal., VOl 38, No. 6, pp 2047 - 2048.


peterrum (commented):
This implementation here already looks quite mature. What about merging this soon and addressing the open points in follow-up PRs? You could open an issue to track them!?

drwells:
Good idea, but we minimally need tests and to get #11516 in first. I should be able to wrap up the minimal feature set tomorrow morning.

drwells:
This also now depends on #11546.

drwells:
This is mostly done except for adding some citations, e.g., to https://arxiv.org/pdf/1803.10065.pdf and Cohen's original 1997 paper on TRI7.

source/simplex/fe_lib.cc
```diff
@@ -1187,7 +1203,274 @@ namespace Simplex
...
+
+  template <int dim, int spacedim>
+  void
+  FE_P_Bubbles<dim, spacedim>::
+    convert_generalized_support_point_values_to_dof_values(
+      const std::vector<Vector<double>> &support_point_values,
+      std::vector<double> &              nodal_values) const
+  {
+    AssertDimension(support_point_values.size(),
+                    this->get_unit_support_points().size());
```
peterrum, line 1454: Is this check actually needed?

include/deal.II/simplex/fe_lib.h
```diff
@@ -401,7 +401,52 @@ namespace Simplex
...
+   * Since this FE space includes bubbles (i.e., extra functions which are
+   * nonzero only on element interiors), the polynomial degrees of the component
+   * basis functions are higher than the actual approximation degree of the
+   * element. For example, with a constructor argument <code>degree = 2</code>
+   * in 3D, the polynomials are in fact cubic (degree 3) but the order of the
+   * approximation is the same as if we were using quadratic (degree 2) finite
+   * elements.
+   */
+  template <int dim, int spacedim = dim>
+  class FE_P_Bubbles : public dealii::FE_Poly<dim, spacedim>
```
peterrum, line 416: I would love to see a visualization of the shape functions! Is there a proven way to do that?

tests/simplex/fe_p_bubbles_02.cc
```diff
@@ -0,0 +1,332 @@
...
+#include <deal.II/simplex/fe_lib.h>
+#include <deal.II/simplex/quadrature_lib.h>
+
+#include "../tests.h"
+
+using namespace dealii;
+
+template <int dim, int spacedim = dim>
+Quadrature<dim>
+compute_nodal_quadrature(const FiniteElement<dim, spacedim> &fe)
```
peterrum, line 42: Here it is ;)

tests/simplex/fe_p_bubbles_02.cc
```diff
@@ -0,0 +1,332 @@
...
+  initlog();
+
+  test_unit_support_points<1>();
+  test_unit_support_points<1, 2>();
+  test_unit_support_points<1, 3>();
+
+  test_unit_support_points<2>();
+  test_unit_support_points<2, 3>();
+
+  test_unit_support_points<3>();
```
peterrum, line 325: This is part of the other test?

peterrum (commented):
Great progress! Only some minor points.

include/deal.II/simplex/fe_lib.h
```diff
@@ -401,7 +401,52 @@ namespace Simplex
     get_name() const override;
   };
 
+  /**
+   * Enriched version of FE_P that can be used with nodal quadrature.
```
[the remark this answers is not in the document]
drwells: It works with consistent quadrature too (it's a superset of P2 for quadratics and just P1 for linears) - is that what you mean?

source/simplex/quadrature_lib.cc
```diff
@@ -82,6 +82,44 @@ namespace Simplex
             this->weights.emplace_back(q12 * 0.132394152789);
             this->weights.emplace_back(q12 * 0.132394152789);
           }
+        else if (n_points_1D == 4)
+          {
+            const double q12 = 0.5;
+
+            // clang-format off
+            this->quadrature_points.emplace_back(0.06493051315916486, 0.06493051315916486);
```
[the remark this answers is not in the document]
drwells: On my TODO list here is to deconstruct https://github.com/nschloe/quadpy/blob/main/quadpy/t3/_witherden_vincent/witherden_vincent_07.json so that we don't have to copy and paste points.

include/deal.II/simplex/fe_lib.h
drwells: That would be pretty easy to do by wrapping `FiniteElement::shape_value` in a `Function` and plotting it.

tests/simplex/fe_p_bubbles_02.cc
drwells: also on the todo list is moving that to FETools, but I wanted to print it to check.

peterrum:
@drwells Once you have fixed the indentation, let's get this in master!

drwells:
I'll try to wrap it up today. I'd still like to move some code I put in the test into `FETools`.

source/simplex/fe_lib.cc
drwells: I think so - `support_point_values` is a function input.

tests/simplex/fe_p_bubbles_02.cc
drwells: Yes - I'll remove it here.

drwells:
I fixed the documentation, indentation, and test - this should now be ready!

peterrum (approved):
Excellent! Thanks, @drwells!

peterrum (approved):
I think this is great addition! I am looking forward to the follow-up PRs!

I will merge now so that people can use this class and give us feedback. Adjustments can be still made later on. Thx @drwells!

drwells:
TET15 might be the most modern element we have, since it was only first published in 2018 :)

PR #11506, opened 2021-01-12
ParameterAcceptor::enter/leave_subsection()

luca-heltai:
First commit adds the feature, second commit shows its use in step-70.

### DIFF
--- a/source/base/parameter_acceptor.cc
+++ b/source/base/parameter_acceptor.cc
@@ -39,10 +39,13 @@ ParameterAcceptor::ParameterAcceptor(const std::string &name)
 }
 
 
+
 ParameterAcceptor::~ParameterAcceptor()
 {
   class_list[acceptor_id] = nullptr;
 }
+
+
 
 std::string
 ParameterAcceptor::get_section_name() const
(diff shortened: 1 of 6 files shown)


source/base/parameter_acceptor.cc
```diff
@@ -182,9 +191,37 @@ ParameterAcceptor::get_section_path() const
...
+
+void
+ParameterAcceptor::enter_subsection(const std::string &subsection)
+{
+  AssertThrow(subsection.find(sep) == std::string::npos,
+              ExcMessage(
+                "A subsection name cannot contain the special character '/'"));
+  // First the easy case.
+  if (subsection == "")
+    return;
```
bangerth, line 209: I think you should throw an exception in this case. If you just return here, then user code will no longer have matching enter/leave subsection calls.

bangerth (changes requested):

source/base/parameter_acceptor.cc
luca-heltai: done.

bangerth (approved):

PR #11507, opened 2021-01-13
Bugfix: DoFs constrained to 0 if FE_Nothing is part of FE_System

### LINKED ISSUE #1496: FE_Nothing ends up dominating when in an FE_System

marcfehling:
Attempt to fix #1496.

If I understood correctly, we add constraints for every row in face interpolation matrices (that do not correspond to identities). Whenever a pure `FE_Nothing` element is involved, these matrices are empty with `m x 0` or `0 x n` entries.

As part of a `FE_System` this is no longer the case: from my understanding, there will simply be rows with zeros that correspond to DoFs of the `FE_Nothing` counterpart on the other side of the face.

This PR adds a simple check whether a row is empty and skips the constraint if so. This also affects constraints with `FE_Enriched` since its implementation depends on `FE_System`.

---

I am not sure whether my approach is too naive, as it affects few test results in a major way. This first patch implements the change (2 lines of code), while the others fix the output for the testsuite in different categories. Please have a look if they are justified.


source/dofs/dof_tools_constraints.cc
```diff
@@ -533,6 +533,14 @@ namespace DoFTools
                   for (unsigned int i = 0; i < n_primary_dofs; ++i)
                     abs_sum += std::abs(face_constraints(row, i));
 
+                  // if all entries of the row are zero, the corresponding dof
+                  // is not supposed to be constrained to any other dof. we can
+                  // thus skip this row
+                  // this happens for example whenever FE_System elements with
+                  // FE_Nothing elements are constrained
+                  if (abs_sum == 0)
+                    continue;
```
bangerth, line 542: As we discussed today, this isn't correct. It would trigger for the case of `FE_Q(1)` next to an `FE_Nothing` regardless of whether the latter dominates or not, I believe. This would probably be easy to verify with a little testcase, but I'm pretty sure this is not correct.

bangerth (changes requested):

bangerth:
What do we do with this? I think it's not right. My suggestion would be to just close this one and get back at the problem with a different approach.

marcfehling:
> What do we do with this? I think it's not right. My suggestion would be to just close this one and get back at the problem with a different approach.

Yes, this is not the final solution to the problem, but may be part of it. I agree on closing this issue, since it does not mean that this suggestions is lost.

PR #11509, opened 2021-01-13
Add an assert to ReferenceCell::compute_orientation()

peterrum:
Adds an assert. Largest part of the PR is only moving functions to the bottom of the file.

Let's get this merged soon, since I have a follow up messing around with the indices.


drwells (approved):
We don't normally but exceptions in internal namespaces, but I don't think there is any reasonable way to catch this anyway, so 👍

PR #11514, opened 2021-01-13
MappingFE::transform_real_to_unit_cell()

luca-heltai:
First patch needed for particles to work on simplices.


peterrum (approved):

peterrum:
Thanks @luca-heltai! Much appretiated!

luca-heltai:
The failing test is unrelated:

~~~


The following tests FAILED:

	3820 - multigrid/transfer_matrix_free_06.mpirun=15.debug (Timeout)

Errors while running CTest

script returned exit code 8
~~~

peterrum:
@luca-heltai Yes it works :) Thx!

Screenshot from 2021-01-13 22-42-43

luca-heltai:
@lwy5515 has confirmed that this works also with `GridTools::compute_point_locations()`, so `ParticleHandler` should work already...

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
   const typename Triangulation<dim, spacedim>::cell_iterator &cell,
   const Point<spacedim> &                                     p) const
 {
-  Assert(false, StandardExceptions::ExcNotImplemented());
+  const std::vector<Point<spacedim>> support_points =
+    this->compute_mapping_support_points(cell);
+
+  Tensor<1, spacedim> DF[dim];
```
bangerth, line 915: Why not declare this as a `DerivativeForm`? I think that that is the correct data type.

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+  // [TODO]
+  // When 2nd derivatives are implemented, we'll uncomment this
+  // Tensor<1, spacedim> D2F[dim][dim];
```
bangerth, line 918: Why not do it right away?

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+      for (unsigned int j = 0; j < dim; ++j)
+        {
+          DF[j].clear();
+          // [TODO]
+          // When 2nd derivatives are implemented, we'll uncomment this
+          // for (unsigned int l = 0; l < dim; ++l)
+          //   D2F[j][l].clear();
+        }
+    }
+  while (f.norm() > eps && loop++ < loop_limit);
```
bangerth, line 978: You only do the loop incrementation if the first part of the condition is `true`. This is really difficult to read/reason about. Please increment `loop` inside the loop.

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+  // [TODO]
+  // When 2nd derivatives are implemented, we'll uncomment this
+  // Tensor<1, spacedim> D2F[dim][dim];
+
+  Point<dim>      p_unit;
+  Point<spacedim> mapped_point;
+  Point<spacedim> p_minus_F;
+
+  Point<dim>     f;
+  Tensor<2, dim> df;
```
bangerth, line 925: As far as I can see, of these five variables, only `f` and `p_unit` live outside the loop. Please just move the declarations of the others to the place where they are assigned in the loop.

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+
+  const double       eps        = 1.e-12 * cell->diameter();
+  const unsigned int loop_limit = 10;
+
+  unsigned int loop = 0;
+  do
+    {
+      auto mapped_point = support_points[0] * 0.0;
```
bangerth, line 933: ?? Why not be explicit:
```suggestion
      Point<spacedim> mapped_point;
```

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+      for (unsigned int i = 0; i < this->fe->n_dofs_per_cell(); ++i)
+        {
+          mapped_point += support_points[i] * this->fe->shape_value(i, p_unit);
+          auto F = this->fe->shape_grad(i, p_unit);
+          // [TODO]
+          // When 2nd derivatives are implemented, we'll uncomment this
+          // auto H = this->fe->shape_grad_grad(i, p_unit);
```
bangerth, line 940: Don't all finite elements actually implement this?

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+  const unsigned int loop_limit = 10;
+
+  unsigned int loop = 0;
+  do
+    {
+      auto mapped_point = support_points[0] * 0.0;
+      for (unsigned int i = 0; i < this->fe->n_dofs_per_cell(); ++i)
+        {
+          mapped_point += support_points[i] * this->fe->shape_value(i, p_unit);
+          auto F = this->fe->shape_grad(i, p_unit);
```
bangerth, line 937:
```suggestion
          const auto F = this->fe->shape_grad(i, p_unit);
```

bangerth: I think it would be easier to read, though, if you called this variable `G_i` or `shape_grad_i` to make it clearer what it corresponds to.

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+              DF[j] += F[j] * support_points[i];
+              // [TODO]
+              // When 2nd derivatives are implemented, we'll uncomment this
+              // for (unsigned int l = 0; l < dim; ++l)
+              //   D2F[j][l] += H[j][l] * support_points[i];
+            }
+        }
+
+      p_minus_F = p;
+      p_minus_F -= mapped_point;
```
bangerth, line 952: Just write `p-mapped_point`

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+
+      for (unsigned int j = 0; j < dim; ++j)
+        {
+          f[j] = DF[j] * p_minus_F;
```
bangerth, line 956: I think what you're doing here is simply a gradient algorithm -- is this right? Maybe add a comment at the top that outlines the algorithm you're implementing?

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+          for (unsigned int l = 0; l < dim; ++l)
+            // [TODO]
+            // When 2nd derivatives are implemented, we'll uncomment this
+            // df[j][l] = -DF[j] * DF[l] + D2F[j][l] * p_minus_F;
```
bangerth, line 960: Wouldn't you have to take the inverse of D2F?

source/fe/mapping_fe.cc
bangerth: No, it's not a steepest descent. I'm confused. I think I need more comments :-)

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,82 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+        }
+      // Solve  [df(x)]d=f(x)
+      const Tensor<1, dim> d =
+        invert(df) * static_cast<const Tensor<1, dim> &>(f);
+      p_unit -= d;
+
+      // Clear DF and D2F
+      for (unsigned int j = 0; j < dim; ++j)
+        {
+          DF[j].clear();
```
bangerth, line 971: You only need this because you re-use the variable in the next loop iteration and you declared it outside the loop. Just move the declaration of the variable into the loop -- then it dies at the `}` and is initialized automatically once you start the next iteration.

bangerth (changes requested):

source/fe/mapping_fe.cc
luca-heltai: 2nd derivatives are not implemented for Simplex::FE_P (yet)

peterrum: I am on it!

source/fe/mapping_fe.cc
luca-heltai: That's true. I was lazy, and copy/pasted existing code that was there before we had DerivativeForm. :)

source/fe/mapping_fe.cc
luca-heltai: done.

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+
+  const double       eps        = 1.e-12 * cell->diameter();
+  const unsigned int loop_limit = 10;
+
+  Point<dim> p_unit;
+
+  unsigned int loop = 0;
+
+  // This loop solves the following problem:
+  // grad_F^T residual + (grad_F^T grad_F + grad_F^T hess_F^T dp) dp = 0
```
bangerth, line 923:
```suggestion
  // grad_F^T residual + (grad_F^T grad_F + grad_F^T hess_F^T) dp = 0
```

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+          for (unsigned int j = 0; j < dim; ++j)
+            {
+              grad_FT[j] += grad_F_i[j] * support_points[i];
+              for (unsigned int l = 0; l < dim; ++l)
+                hess_FT[j][l] += hessian_F_i[j][l] * support_points[i];
+            }
+        }
+
+      // Residual
+      auto residual = p - mapped_point;
```
bangerth, line 952:
```suggestion
      const auto residual = p - mapped_point;
```

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+  // (grad_F^T hess_F dp) is approximated by (-hess_F * residual)
+  // This is basically a second order approximation of Newton method, where the
+  // Jacobian is corrected with a higher order term coming from the hessian.
+  do
+    {
+      Point<spacedim> mapped_point;
+
+      // Transpose of the gradient map
+      DerivativeForm<1, spacedim, dim> grad_FT;
+      Tensor<1, dim>                   grad_FT_residual;
```
bangerth, line 934:
```suggestion
```

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+                hess_FT[j][l] += hessian_F_i[j][l] * support_points[i];
+            }
+        }
+
+      // Residual
+      auto residual = p - mapped_point;
+      // Project the residual on the reference coordinate system
+      // to compute the error, and to filter components orthogonal to the
+      // manifold, and compute a 2nd order correction of the metric tensor
+      grad_FT_residual = apply_transformation(grad_FT, residual);
```
bangerth, line 956:
```suggestion
      const Tensor<1, dim>  grad_FT_residual = apply_transformation(grad_FT, residual);
```

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+  // This is basically a second order approximation of Newton method, where the
+  // Jacobian is corrected with a higher order term coming from the hessian.
+  do
+    {
+      Point<spacedim> mapped_point;
+
+      // Transpose of the gradient map
+      DerivativeForm<1, spacedim, dim> grad_FT;
+      Tensor<1, dim>                   grad_FT_residual;
+      Tensor<2, dim>                   corrected_metric_tensor;
```
bangerth, line 935: Move this after the `// Now compute the...` comment below

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+        break;
+
+      // Now compute the (corrected) metric tensor
+      for (unsigned int j = 0; j < dim; ++j)
+        for (unsigned int l = 0; l < dim; ++l)
+          corrected_metric_tensor[j][l] =
+            -grad_FT[j] * grad_FT[l] + hess_FT[j][l] * residual;
+
+      // And compute the update
+      auto g_inverse = invert(corrected_metric_tensor);
```
bangerth, line 969:
```suggestion
      const auto g_inverse = invert(corrected_metric_tensor);
```

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
+      p_unit -= Point<dim>(g_inverse * grad_FT_residual);
+
+      ++loop;
+    }
+  while (loop < loop_limit);
```
bangerth, line 974: You dropped the early-exit test for a tolerance. Was this on purpose? This is a relatively expensive function, so it might be useful to check on the size of the residual.

source/fe/mapping_fe.cc
```diff
@@ -909,12 +909,78 @@ MappingFE<dim, spacedim>::transform_real_to_unit_cell(
...
 
-  (void)cell;
-  (void)p;
+  // Here we check that in the last execution of while the first
+  // condition was already wrong, meaning the residual was below
+  // eps. Only if the first condition failed, loop will have been
+  // increased and tested, and thus have reached the limit.
+  AssertThrow(loop < loop_limit,
+              (typename Mapping<dim, spacedim>::ExcTransformationFailed()));
```
bangerth, line 981: Isn't this now always going to fail because there is no other exit of the `do...while` loop?

source/fe/mapping_fe.cc
luca-heltai: it is up there, right after I compute the residual, before I compute the rest. If the residual is lower than tolerance, I just exit with a `break` from the loop.

source/fe/mapping_fe.cc
luca-heltai: see above.

source/fe/mapping_fe.cc
luca-heltai: I do.

source/fe/mapping_fe.cc
luca-heltai: No. This was correct. I'm using  `grad_F + hess_F^T dp` as a higher order approximation of `grad_F`.

luca-heltai:
Ping @bangerth ?

bangerth (approved):

bangerth:
Thanks!

PR #11516, opened 2021-01-13
Add a utility function for Simplex::FE_P's support points.

drwells:
Part of #11505.

While we are here, fix a bug where unit face support points were assigned to adiscontinous element, which doesn't make sense (see FiniteElement::get_unit_face_support_points()).

### DIFF
--- a/source/simplex/fe_lib.cc
+++ b/source/simplex/fe_lib.cc
@@ -57,6 +57,107 @@ namespace Simplex
     }
 
     /**
+     * Set up a vector that contains the unit (reference) cell support points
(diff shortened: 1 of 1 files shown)


peterrum (approved):

peterrum:
> fix a bug where unit face support points were assigned to adiscontinous element, which doesn't make sense (see FiniteElement::get_unit_face_support_points()).

If you say so ;)

drwells:
It's not much of a bug, but who am I to go against the documentation of `FiniteElement`?

drwells:
The only failing test is unrelated and due to a timeout, so I recommend we merge.

PR #11512, opened 2021-01-13
Break the declaration of the static mapping into a file of its own.

bangerth:
First step to what we discussed today about changing the static Q1 mapping to something more generic.

/rebuild

PR #11520, opened 2021-01-13
Introduce ReferenceCell::get_default_linear_mapping() with triangulation argument. Use it.

bangerth:
A better approach than #11512. This also changes all uses of `StaticMappingQ1` in `source/` to the new function -- fixing a few potential bugs in the process, and leaving some TODOs that can't be addressed with the current finite element and quadrature interfaces. I'll convert `include/` later, but wanted to put this out here already for discussion.

/rebuild

### DIFF
--- a/include/deal.II/fe/mapping_q1.h
+++ b/include/deal.II/fe/mapping_q1.h
@@ -80,6 +80,10 @@ public:
  * these contexts throughout the library, this class defines a static $Q_1$
  * mapping object. This object can then be used in all of those places where
  * such an object is needed.
+ *
+ * @note The use of this object should be avoided since it is only applicable
+ *   in cases where a mesh consists exclusively of quadrilaterals or hexahedra.
+ *   Use ReferenceCell::get_default_linear_mapping() instead.
  */
 template <int dim, int spacedim = dim>
 struct StaticMappingQ1
--- a/source/grid/reference_cell.cc
+++ b/source/grid/reference_cell.cc
@@ -140,6 +140,26 @@ namespace ReferenceCell
 
 
 
+  template <int dim, int spacedim>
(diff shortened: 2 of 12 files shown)


source/grid/reference_cell.cc
```diff
@@ -140,6 +140,25 @@ namespace ReferenceCell
 
 
 
+  template <int dim, int spacedim>
+  const Mapping<dim, spacedim> &
+  get_default_linear_mapping(const Triangulation<dim, spacedim> &triangulation)
+  {
+    const auto reference_cell_types = triangulation.get_reference_cell_types();
```
peterrum, line 147:
```suggestion
    const auto & reference_cell_types = triangulation.get_reference_cell_types();
```

peterrum (approved):
Looks good to me.

include/deal.II/fe/mapping_q1.h
```diff
@@ -80,6 +80,10 @@ class MappingQ1 : public MappingQGeneric<dim, spacedim>
  * these contexts throughout the library, this class defines a static $Q_1$
  * mapping object. This object can then be used in all of those places where
  * such an object is needed.
+ *
+ * @note The use of this object should be avoided since it is only applicable
+ *   in cases where a mesh consists exclusively of quadrilaterals or hexahedra.
+ *   Use ReferenceCell::get_default_linear_mapping() instead.
```
tjhei, line 86: Is this a candidate for DEPRECATED_EARLY? Or is this too early? :-)

bangerth: Yes, somewhere in this series of patches.

tjhei (approved):

PR #11522, opened 2021-01-13
Fix missing 'spacedim' template argument.

bangerth:
Just a missing `, spacedim` in two places. I guess these functions were simply never instantiated for `dim != spacedim`, so this wasn't wrong, but if we ever plan on instantiating these functions for `dim < spacedim`, then this fix will be necessary one way or the other.

/rebuild


masterleinad (approved):
LGTM.

tjhei (approved):

PR #11527, opened 2021-01-14
simplex: compare_for_domination

marcfehling:
Cleans up the `FiniteElementDomination` functions for simplices.

### DIFF
--- a/source/fe/fe_dgq.cc
+++ b/source/fe/fe_dgq.cc
@@ -30,6 +30,8 @@
 
 #include <deal.II/lac/vector.h>
 
+#include <deal.II/simplex/fe_lib.h>
+
 #include <iostream>
 #include <memory>
 #include <sstream>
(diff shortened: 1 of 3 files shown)


source/fe/fe_dgq.cc
```diff
@@ -694,6 +696,16 @@ FE_DGQ<dim, spacedim>::compare_for_domination(
       else
         return FiniteElementDomination::other_element_dominates;
     }
+  else if (const Simplex::FE_DGP<dim, spacedim> *fe_dgp_other =
+             dynamic_cast<const Simplex::FE_DGP<dim, spacedim> *>(&fe_other))
```
peterrum, line 700: Why is here only `Simplex::FE_DGP` and not the version for pyramids and wedges?

peterrum: May I ask you to add a test.

marcfehling: You've checked only for certain elements in assertions in `compare_for_domination` inside `fe_lib.cc`.

I've only added corresponding conditionals to the counterparts marked in these assertions.

I'm not happy that we need to declare the domination relations in both elements. There should be one large construct that knows about all FE relations. But for now let's stick to this.

marcfehling: Oh, there are currently no functions `FE_WedgeDGP::compare_for_domination()` and `FE_PyramidDGP::compare_for_domination()`. I will add them.

bangerth: Do you still need to update this one place now with the other kinds of elements?

bangerth (commented):
This looks good to me (except for the one places marked) but it would be good to have test(s).

marcfehling:
Do you have a test like [this one](https://github.com/dealii/dealii/blob/master/tests/hp/fe_collection_06.cc) in mind (which is pretty dull)? Or would #11548 suffice (which only demonstrates `FE_P`)?

bangerth:
Both of these would be useful. Let's make progress here -- the patch you have here passes the tests, I'll take you at your word that you'll finish the tests you're talking about :-)

bangerth (approved):

PR #11528, opened 2021-01-14
simplex: vertex_dof_identities for FE_P

marcfehling:
Cleans up `vertex_dof_identities` (only for `FE_P` for now).


drwells:
I'm a bit surprised that this works without simplex support.

This question might be outside the scope of this PR (which is fine) but I want to know - is there any way that we could implement a better API for stating "this element interpolates nodal values" instead of using `dynamic_cast`?

drwells (approved):

peterrum (approved):

bangerth (approved):

PR #11531, opened 2021-01-14
cell bounding box tests for Simplex grids in 2d and 3d

nicola-giuliani:

PR #11536, opened 2021-01-14
Only consider the tests in tests/simplex/ if we actually have simplex support

bangerth:
This matches what we do in other `tests/*/` directories.

/rebuild


bangerth:
This will allow removing the `.with_simplex_support=true` from all of the output files, but I'll wait with this for a follow-up (and after #11531 is merged).

tjhei (approved):
You beat me to it.

bangerth:
Har har!

peterrum (commented):
I am not happy with this change. Not all tests strictly need simplex support.

bangerth:
Ah... But then, why would something that doesn't actually need simplex support be in that directory? I see the following tests:
```
> ll *output | grep -v with_simplex_support                                                                                                                                        
-rw-rw-r-- 1 bangerth bangerth   3645 Jan  1 17:26 fe_lib_01.output                                                                                                                             
-rw-rw-r-- 1 bangerth bangerth  34291 Jan 14 09:01 fe_lib_02.output                                                                                                                             
-rw-rw-r-- 1 bangerth bangerth    683 Dec  3 20:25 matrix_free_range_iteration_01.output                                                                                                        
-rw-rw-r-- 1 bangerth bangerth   3494 Dec  3 20:25 polynomials_01.output                                                                                                                        
-rw-rw-r-- 1 bangerth bangerth  21832 Jan 14 09:01 q_projection_01.output                                                                                                                       
-rw-rw-r-- 1 bangerth bangerth   1357 Dec  3 20:25 quadrature_lib_01.output                                                                                                                     
-rw-rw-r-- 1 bangerth bangerth   1425 Jan  1 17:26 unit_tangential_vectors_01.output                                                                                                            
-rw-rw-r-- 1 bangerth bangerth   3469 Jan  1 17:26 variable_face_quadratures_01.output                                                                                                          
-rw-rw-r-- 1 bangerth bangerth   1142 Jan  1 17:26 variable_face_quadratures_02.output 
```
We could move these to a different directory.

peterrum:
Here are tests related to simplices and the `Simplex` namespace. `Simplex::QGauss` does strictly speaking not need simplex support (of the triangulation class) and can be used like `QDuffy`.

bangerth:
OK. It's maybe a bit awkward this way, but it's not an important issue either. The flag will eventually go away and we can rename all of the files at that point.

Issue #11540, opened 2021-01-14
Fix wrong default mapping for non-quad/hex meshes


bangerth:
#11539 changes a whole bunch of places that we know are only correct for quad/hex meshes. #11520 also leaves a number of TODOs. We need to figure out a way to describe a "generic" linear mapping that could be passed in these places so as to work also for tri/tet/mixed meshes in these places.

bangerth:
As a point of reference, I removed the default arguments in `grid_tools.h` and compiled the library with no problems. In other words, all places in the library itself that call these functions actually provide a mapping explicitly.

peterrum:
@bangerth What is the progress here?

peterrum:
@bangerth What is missing there? The above stated TODOs are fixed in the mean time.

bangerth:
It's places like these in `GridTools`:
```
  template <int dim, int spacedim>
  double
  volume(const Triangulation<dim, spacedim> &tria,
         const Mapping<dim, spacedim> &      mapping =
           (ReferenceCells::get_hypercube<dim>()
              .template get_default_linear_mapping<dim, spacedim>()));
```
The default value for the last argument will only work if the triangulation is in fact a quad/hex mesh. There are numerous cases like this in this file.

PR #11541, opened 2021-01-14
Use the correct mapping for a finite element.

bangerth:
Follow-up to #11520 and #11540. Fixes the TODO places introduced in #11520 since @peterrum told me how to query the reference cell -- I had looked into class `FiniteElement` and not found it there, but it turns out to be in class `FiniteElementData`.

/rebuild


masterleinad (approved):
Looks good to me.

peterrum (approved):

PR #11543, opened 2021-01-14
Make code slightly easier to read.

bangerth:
Reading through #11516 by @drwells , I found myself confused by the fact that some of the `if` statements were `>`, others `==`. I initially thought they were mutually exclusive and could have been written as `else if`, but they are in fact nested.

Reflect this in the code, and move the assertion to a place where we can actually show the condition that's tested in the error message.

/rebuild

### DIFF
--- a/source/simplex/fe_lib.cc
+++ b/source/simplex/fe_lib.cc
@@ -69,60 +69,52 @@ namespace Simplex
         {
           // We don't really have dim = 1 support for simplex elements yet, but
           // its convenient for populating the face array
+          Assert(degree <= 2, ExcNotImplemented());
           if (degree >= 1)
             {
               unit_points.emplace_back(0.0);
               unit_points.emplace_back(1.0);
-            }
-          if (degree == 2)
-            {
-              unit_points.emplace_back(0.5);
-            }
-          if (degree > 2)
-            {
-              Assert(false, ExcNotImplemented());
+
+              if (degree == 2)
+                unit_points.emplace_back(0.5);
             }
         }
       else if (dim == 2)
         {
+          Assert(degree <= 2, ExcNotImplemented());
           if (degree >= 1)
             {
               unit_points.emplace_back(0.0, 0.0);
               unit_points.emplace_back(1.0, 0.0);
               unit_points.emplace_back(0.0, 1.0);
-            }
-          if (degree == 2)
-            {
-              unit_points.emplace_back(0.5, 0.0);
-              unit_points.emplace_back(0.5, 0.5);
-              unit_points.emplace_back(0.0, 0.5);
-            }
-          if (degree > 3)
-            {
-              Assert(false, ExcNotImplemented());
+
+              if (degree == 2)
+                {
+                  unit_points.emplace_back(0.5, 0.0);
+                  unit_points.emplace_back(0.5, 0.5);
+                  unit_points.emplace_back(0.0, 0.5);
+                }
             }
         }
       else if (dim == 3)
         {
+          Assert(degree <= 2, ExcNotImplemented());
           if (degree >= 1)
             {
               unit_points.emplace_back(0.0, 0.0, 0.0);
               unit_points.emplace_back(1.0, 0.0, 0.0);
               unit_points.emplace_back(0.0, 1.0, 0.0);
               unit_points.emplace_back(0.0, 0.0, 1.0);
-            }
-          if (degree == 2)
-            {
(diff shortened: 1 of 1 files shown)


drwells (approved):

source/simplex/fe_lib.cc
```diff
@@ -69,60 +69,52 @@ namespace Simplex
...
-            {
-              Assert(false, ExcNotImplemented());
+
+              if (degree == 2)
+                unit_points.emplace_back(0.5);
             }
         }
       else if (dim == 2)
         {
+          Assert(degree <= 2, ExcNotImplemented());
```
Rombur, line 84: The new condition is different than the previous one. This looks correct to me but I just want to make sure that it was a deliberate change and not simply a copy/paste.

bangerth: I think the code before was wrong. It had
```
          if (degree >= 1)
            {
              unit_points.emplace_back(0.0, 0.0);
              unit_points.emplace_back(1.0, 0.0);
              unit_points.emplace_back(0.0, 1.0);
            }
          if (degree == 2)
            {
              unit_points.emplace_back(0.5, 0.0);
              unit_points.emplace_back(0.5, 0.5);
              unit_points.emplace_back(0.0, 0.5);
            }
          if (degree > 3)
            {
            }
```
This doesn't cover the case `degree==3` at all, but I think that was a mistake. `degree==3` should trigger the exception, which the changed assertion actually does.

To be fair, I had not seen this issue when writing the patch -- good eyes, @Rombur. It's a bit amusing that the existing code treated the exceptional case in three different ways:
```
      if (dim == 1)
        {
          // We don't really have dim = 1 support for simplex elements yet, but
          // its convenient for populating the face array
          if (degree >= 1)
            {
              unit_points.emplace_back(0.0);
              unit_points.emplace_back(1.0);
            }
          if (degree == 2)
            {
              unit_points.emplace_back(0.5);
            }
          if (degree > 2)
            {
            }
        }
      else if (dim == 2)
        {
          if (degree >= 1)
            {
              unit_points.emplace_back(0.0, 0.0);
              unit_points.emplace_back(1.0, 0.0);
              unit_points.emplace_back(0.0, 1.0);
            }
          if (degree == 2)
            {
              unit_points.emplace_back(0.5, 0.0);
              unit_points.emplace_back(0.5, 0.5);
              unit_points.emplace_back(0.0, 0.5);
            }
          if (degree > 3)
            {
            }
        }
      else if (dim == 3)
        {
          if (degree >= 1)
            {
              unit_points.emplace_back(0.0, 0.0, 0.0);
              unit_points.emplace_back(1.0, 0.0, 0.0);
              unit_points.emplace_back(0.0, 1.0, 0.0);
              unit_points.emplace_back(0.0, 0.0, 1.0);
            }
          if (degree == 2)
            {
              unit_points.emplace_back(0.5, 0.0, 0.0);
              unit_points.emplace_back(0.5, 0.5, 0.0);
              unit_points.emplace_back(0.0, 0.5, 0.0);
              unit_points.emplace_back(0.0, 0.0, 0.5);
              unit_points.emplace_back(0.5, 0.0, 0.5);
              unit_points.emplace_back(0.0, 0.5, 0.5);
            }
          if (degree == 3)
            {
            }
```
I believe that two of the three ways were wrong, and are now correct :-)

source/simplex/fe_lib.cc
```diff
@@ -69,60 +69,52 @@ namespace Simplex
         {
           // We don't really have dim = 1 support for simplex elements yet, but
           // its convenient for populating the face array
+          Assert(degree <= 2, ExcNotImplemented());
```
peterrum, line 72: What about using `AssertIndexRange`? I would like the degree to be printed in the error message.

source/simplex/fe_lib.cc
peterrum: I think before (yesterday or so) it was `else {Assert(false, ExcNotImplemented());}` in all cases ;) But back then, the vertex points were duplicated... This way it is much better!

peterrum (commented):
Thanks a lot! I have minor comment.

source/simplex/fe_lib.cc
bangerth: I would rather leave it as it is: `degree` isn't an index, and the error message shouldn't suggest that it is. The error you get with `AssertIndexRange` reads like this:
```
    << "Index " << arg1 << " is not in the half-open range [" << arg2 << ","
    << arg3 << ")."
```
I don't think that would make things easier to understand.

PR #11544, opened 2021-01-14
Convert ReferenceCell::Type to a class.

bangerth:
@peterrum -- FYI.

The `ReferenceCell` namespace has over time accumulated a good number of free functions that all take `ReferenceCell::Type` as a `const` argument. This suggests a design mistake: `ReferenceCell::Type` should have been a class (rather than an `enum`) so that these functions could all be member functions.

This is a first patch addressing this issue. I'm only converting the enum into a class and provide it with the necessary member functions, but nothing else. This turns out to be backward compatible for now with the exception of a single place in the library. 

A follow-up patch will then start to move the many free functions into the class. This is no longer backward compatible, but I think we can still do this at this time.

/rebuild


peterrum (commented):
Before making a decision I would like to wait (for the follow-up PRs) so that I know how this supposed to be used. 

Do you know why the test output has chaned?

bangerth:
The test output changed because I had to modify the code that serializes these objects. Before it was serialized as an `enum` and now as an `int`.

peterrum (approved):

peterrum:
I though @tjhei wanted to merge this! Sorry.

PR #11545, opened 2021-01-14
Convert step-20 to a simplex test

lpsaavedra:
Since Raviart-Thomas elements are still not available for simplex, another finite element spaces were used. All necessary modifications are described at the beginning of the .cc file.


peterrum (approved):
Thanks a lot, @lpsaavedra!

PR #11546, opened 2021-01-15
Barycentric polynomials

drwells:
We can replace Simplex::ScalarPolynomial with a more general class for handling polynomials on simplices. The barycentric polynomial classes can be assembled algebraically (e.g., `M1 * M2 - 4 * M1 * M3`) which is very convenient and will make adding more simplex FEs a lot easier.


drwells:
I should probably add some more tests before we merge but all the simplex tests pass locally for me.

drwells:
I added a test.

include/deal.II/simplex/barycentric_polynomials.h
```diff
@@ -0,0 +1,639 @@
...
+                          const Number                 coefficient);
+
+    /**
+     * Return the specified monomial.
+     */
+    static BarycentricPolynomial<dim, Number>
+    monomial(const unsigned int d);
+
+    void
+    print(std::ostream &out) const;
```
peterrum, line 55: Documentation is missing.

include/deal.II/simplex/barycentric_polynomials.h
```diff
@@ -0,0 +1,639 @@
...
+
+    Table<3, BarycentricPolynomial<dim>> poly_hessians;
+
+    Table<4, BarycentricPolynomial<dim>> poly_third_derivatives;
+
+    Table<5, BarycentricPolynomial<dim>> poly_fourth_derivatives;
+  };
+} // namespace Simplex
+
+// free template functions (in Simplex, not dealii::Simplex)
```
peterrum, line 254: I don't get this comment.

include/deal.II/simplex/barycentric_polynomials.h
```diff
@@ -0,0 +1,639 @@
...
+  return out;
+}
+
+// Template function definitions
+
+namespace internal
+{
+  template <int N>
+  TableIndices<N>
+  index_to_indices(const std::size_t &index, const TableIndices<N> &extent)
```
peterrum, line 307: What does this function do?

source/simplex/CMakeLists.txt
```diff
@@ -40,7 +41,7 @@ SET(_inst
 )
 
 FILE(GLOB _header
-  ${CMAKE_SOURCE_DIR}/include/deal.II/tet/*.h
+  ${CMAKE_SOURCE_DIR}/include/deal.II/simplex/*.h
```
peterrum, line 44: A relic from old times...

source/simplex/barycentric_polynomials.cc
```diff
@@ -0,0 +1,334 @@
...
+    }
+  } // namespace internal
+
+
+  template <int dim>
+  BarycentricPolynomials<dim>
+  BarycentricPolynomials<dim>::get_fe_p_basis(const unsigned int degree)
+  {
+    std::vector<BarycentricPolynomial<dim>> polys;
+    auto                                    M = [](const unsigned int d) {
```
peterrum, line 57:
```suggestion
    std::vector<BarycentricPolynomial<dim>> polys;
    
    const auto M = [](const unsigned int d) {
```

source/simplex/barycentric_polynomials.cc
```diff
@@ -0,0 +1,334 @@
...
+  {
+    /**
+     * Get the highest degree of the barycentric polynomial (in Cartesian
+     * coordinates).
+     */
+    template <int dim>
+    unsigned int
+    get_degree(const std::vector<BarycentricPolynomial<dim>> &polys)
+    {
+      // Since the first variable in a simplex polynomial is, e.g., in 2D, 1 - x
```
peterrum, line 33: *Since the first variable in a simplex polynomial is* is what?

peterrum (commented):
Looks very good, @drwells! May I ask you to add some references to literature and/or some latex formulas. How easy is now an extension to higher order?

peterrum:
@drwells How difficult is to add degree=0? @vovannikov might want to use this (I am hesitant to add this to `FE_P`/`FE_DGP` and its polynomials because it would lead to some annoying merge conflicts).

drwells:
I'll try to get this done asap - adding degree 0 should be easy: its just the sum of the `d + 1` barycentric polynomials.

include/deal.II/simplex/barycentric_polynomials.h
drwells: It's wrong - I put the free functions in `dealii::` instead of `dealii::Simplex`.

include/deal.II/simplex/barycentric_polynomials.h
drwells: It converts an integral index into something a `Table` can use to look up an entry. I'll document it.

peterrum:
@drwells Thanks :)

peterrum:
I hope I don't create too much work for. I just have the feeling that as it now I couldn't extend the code myself!

source/simplex/CMakeLists.txt
drwells: Indeed - I was in the neighborhood.

include/deal.II/simplex/barycentric_polynomials.h
drwells: Overloading works correctly if they are in `dealii::Simplex` so I'll get rid of this.

source/simplex/barycentric_polynomials.cc
drwells: I think I made this clearer - do you see the example now?

drwells:
I'll force push in a minute - let me make sure I fixed MSVC's complaints.

drwells:
> How easy is now an extension to higher order?

Pretty easy - this was the point of writing this class. For example
```cpp
  template <int dim>
  BarycentricPolynomials<dim>
  BarycentricPolynomials<dim>::get_fe_p_basis(const unsigned int degree)
  {
    std::vector<BarycentricPolynomial<dim>> polys;

    auto M = [](const unsigned int d) {
      return BarycentricPolynomial<dim, double>::monomial(d);
    };
    switch (degree)
      {
        case 0:
          polys.push_back(0 * M(0) + 1);
          break;
        case 1:
          {
            for (unsigned int d = 0; d < dim + 1; ++d)
              polys.push_back(M(d));
            break;
          }
        case 2:
          {
            for (unsigned int d = 0; d < dim + 1; ++d)
              polys.push_back(M(d) * (2 * M(d) - 1));
            polys.push_back(4 * M(1) * M(0));
            if (dim >= 2)
              {
                polys.push_back(4 * M(1) * M(2));
                polys.push_back(4 * M(2) * M(0));
              }
            if (dim == 3)
              {
                polys.push_back(4 * M(3) * M(0));
                polys.push_back(4 * M(1) * M(3));
                polys.push_back(4 * M(2) * M(3));
              }
            break;
          }
        default:
      }

    return BarycentricPolynomials<dim>(polys);
  }
```
implements P0, P1, and P2 in 1D, 2D and 3D (compare this to the amount of code we had before, and that didn't include all the derivatives). I believe cubics and hierarchical elements are also defined as barycentric polynomials - since we now have a proper implementation of a ring it should be straightforward to programmatically build those polynomial spaces.

drwells:
I believe I fixed the problem with MSVC and addressed all review comments.

drwells:
In fact, that factory function builds P0 already - we should just be able to more-or-less delete the assertions in `FE_P` once this is merged and constants should work correctly.

drwells:
The windows-2016 failure is spurious (we failed to upload an archive).

drwells:
@peterrum I made the changes you suggested offline.

peterrum (approved):
Thanks!

PR #11547, opened 2021-01-15
simplex: line_dof_identities for FE_P

marcfehling:
Cleans up `line_dof_identities` (only for `FE_P` for now).

Currently, we do not have a pretty way to access support points located on lines and faces on simplices. As they are currently hard-coded up to degree 2, I did the same here while placing defensive assertions.

We might think of a more sophisticated way to access these unit points in the future. But for now to make 2D p-adaptation available, this should suffice.

### DIFF
--- a/source/fe/fe_q_base.cc
+++ b/source/fe/fe_q_base.cc
@@ -815,11 +815,42 @@ FE_Q_Base<PolynomialType, dim, spacedim>::hp_line_dof_identities(
 
       return identities;
     }
+  else if (const Simplex::FE_P<dim, spacedim> *fe_p_other =
+             dynamic_cast<const Simplex::FE_P<dim, spacedim> *>(&fe_other))
+    {
+      // DoFs are located along lines, so two dofs are identical if they are
+      // located at identical positions. If we had only equidistant points, we
+      // could simply check for similarity like (i+1)*q == (j+1)*p, but we
+      // might have other support points (e.g. Gauss-Lobatto
+      // points). Therefore, read the points in unit_support_points for the
+      // first coordinate direction. For FE_Q, we take the lexicographic
+      // ordering of the line support points in the first direction (i.e.,
+      // x-direction), which we access between index 1 and p-1 (index 0 and p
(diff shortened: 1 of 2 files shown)


bangerth (approved):

source/fe/fe_q_base.cc
```diff
@@ -813,13 +813,44 @@ FE_Q_Base<PolynomialType, dim, spacedim>::hp_line_dof_identities(
                                                [0]) < 1e-14)
             identities.emplace_back(i, j);
 
+      return identities;
+    }
+  else if (const Simplex::FE_P<dim, spacedim> *fe_p_other =
+             dynamic_cast<const Simplex::FE_P<dim, spacedim> *>(&fe_other))
+    {
+      // dofs are located along lines, so two dofs are identical if they are
+      // located at identical positions. if we had only equidistant points, we
```
peterrum, line 822:
```suggestion
      // DoFs are located along lines, so two dofs are identical if they are
      // located at identical positions. If we had only equidistant points, we
```

source/simplex/fe_lib.cc
```diff
@@ -498,19 +498,84 @@ namespace Simplex
...
 
-    return result;
+    if (const FE_P<dim, spacedim> *fe_p_other =
+          dynamic_cast<const FE_P<dim, spacedim> *>(&fe_other))
+      {
+        // dofs are located along lines, so two dofs are identical if they are
+        // located at identical positions. if we had only equidistant points, we
+        // could simply check for similarity like (i+1)*q == (j+1)*p, but we
+        // might have other support points (e.g. Gauss-Lobatto
+        // points). Therefore, read the points in unit_support_points for the
```
peterrum, line 511:
```suggestion
        // located at identical positions. 
        // Therefore, read the points in unit_support_points for the
```

peterrum (commented):
Looks good. Two minor comments.

peterrum (approved):
Indentation is not passing. But else ready to merge!

PR #11548, opened 2021-01-15
simplex: hanging node tests

marcfehling:
~~Blocked by #11527 and #11547.~~ (EDIT: Have been merged)

Simple tests demonstrating local h- and p-refinement with `FE_P`.

Hanging node constraints do not work yet as they require interpolation matrices. But enumerating DoFs works so far.

### DIFF
--- a/tests/simplex/hanging_nodes_01.cc
+++ b/tests/simplex/hanging_nodes_01.cc
@@ -1,0 +1,71 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2021 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+
+// verify hanging node constraints on locally h-refined simplex mesh
+
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/grid/grid_out.h>
+#include <deal.II/grid/tria.h>
+
+#include <deal.II/lac/affine_constraints.h>
+
+#include <deal.II/simplex/fe_lib.h>
+#include <deal.II/simplex/grid_generator.h>
+
+#include "../tests.h"
+
+
+template <int dim>
+void
+test()
+{
+  // setup grid
+  Triangulation<dim> tria;
+  GridGenerator::subdivided_hyper_cube_with_simplices(tria, 1);
+
+  tria.begin_active()->set_refine_flag();
+  tria.execute_coarsening_and_refinement();
+
+#if false
+  GridOut grid_out;
+  grid_out.write_vtk(tria, deallog.get_file_stream());
+#endif
+
+  DoFHandler<dim> dofh(tria);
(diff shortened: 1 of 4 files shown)


bangerth (approved):

tests/simplex/hanging_nodes_02.cc
```diff
@@ -0,0 +1,67 @@
...
+  DoFHandler<dim> dofh(tria);
+  dofh.begin_active()->set_active_fe_index(1);
+
+  hp::FECollection<dim> fes(Simplex::FE_P<dim>(1), Simplex::FE_P<dim>(2));
+  dofh.distribute_dofs(fes);
+  deallog << "ndofs: " << dofh.n_dofs() << std::endl;
+
+  // hanging node constraints
+  AffineConstraints<double> constraints;
+  // DoFTools::make_hanging_node_constraints(dofh, constraints);
```
bangerth, line 52: also here

bangerth (changes requested):

tests/simplex/hanging_nodes_02.cc
marcfehling: I need a bit more context :)

bangerth: Don't you specifically want to test hanging node constraints? If so, you'll have to re-add this line.

marcfehling: Yes, that is the ultimate goal of these tests. However, calling `DoFTools::make_hanging_node_constraints` at this stage will fail an assertion as no interpolation matrices have been defined yet.

Anyways, for hanging nodes we need to `distribute_dofs` first, and I thought it makes sense to integrate these tests in the library already at this stage (after the blocking PRs are merged).

tests/simplex/hanging_nodes_02.cc
```diff
@@ -0,0 +1,67 @@
...
+test()
+{
+  // setup grid
+  Triangulation<dim> tria;
+  GridGenerator::subdivided_hyper_cube_with_simplices(tria, 1);
+
+  DoFHandler<dim> dofh(tria);
+  dofh.begin_active()->set_active_fe_index(1);
+
+  hp::FECollection<dim> fes(Simplex::FE_P<dim>(1), Simplex::FE_P<dim>(2));
```
peterrum, line 46: Maybe lets also test the the switched version.

peterrum: @marcfehling Thanks! Learning from a mistake you have pointed out ;)

marcfehling:
Blocking PRs have been merged. These tests will now pass in their current state.

bangerth:
Ah, I see. I thought (from the title of the PR and the name of the test) that you actually wanted to check that hanging node constraints are already working. If that's not the case, then that's fine as well. More tests are always good.

Are you working on the face interpolation matrices per chance?

bangerth (approved):

tests/simplex/hanging_nodes_01.cc
```diff
@@ -0,0 +1,71 @@
...
+  grid_out.write_vtk(tria, deallog.get_file_stream());
+#endif
+
+  DoFHandler<dim> dofh(tria);
+  dofh.distribute_dofs(Simplex::FE_P<dim>(1));
+  deallog << "ndofs: " << dofh.n_dofs() << std::endl;
+
+  // hanging node constraints
+  AffineConstraints<double> constraints;
+  // DoFTools::make_hanging_node_constraints(dofh, constraints);
```
bangerth, line 56: I now also understand why you were confused by my "also here" comment -- I had meant to attach a comment to this line saying that you probably want that line back in :-)

Issue #11521, opened 2021-01-13
FE_Nothing constraints are weird


bangerth:
Consider two cells neighboring each other:
* One cell that has a `Q1 x Q1` vector-valued element
* One cell that has a `FE_Nothing(true) x FE_Nothing(false)` element, i.e., only one of the two fe nothings is dominating.

In 2d, the face interpolation matrix that `make_hanging_node_constraints()` gets to see is of size `4 x 0` (2+2 DoFs on the face for the `Q1 x Q` element, none for the two `FE_Nothing`s), but out of this we somehow want to generate *2* constraints (not 4). Which of the rows of the `4 x 0` matrix are to be interpreted as constraining is information that, I believe, we are not currently propagating anywhere.

@marcfehling Want to write a little test program for this case? 

See also #11448 .

PR #11549, opened 2021-01-15
Added test for FENothingxFENothing and FEQxFEQ constraints.

marcfehling:
Pathologic test case as requested in #11521. Analogous to #11448.

Again, no constraints at all are being imposed. I suspect that the interpolation matrices are again empty, contrary to #1496.


peterrum (approved):

marcfehling:
I am about the refactor our `FE_Nothing dominates`-tests and will merge this PR with #11448.

PR #11551, opened 2021-01-15
Step 17 as a simplex test

michelebucelli:
Step 17 (linear elasticity, MPI parallel) converted to a simplex test.


peterrum (approved):
Thanks a lot!

drwells:
It looks like the parallel output file here doesn't match what the tester generated - can you update the output?

michelebucelli:
Unfortunately the test passes on my local build, producing the same output as in the output file. Looking at the test log it appears that the execution fails altogether when run in the pipeline. Is there a way that I can obtain the output produced by the pipeline, so I can better understand what's going on?

Also, there seems to be the same issue with my other pull request #11552.

peterrum:
> Is there a way that I can obtain the output produced by the pipeline, so I can better understand what's going on?

tjhei:
> I am not sure if this is the most elegant solution. Any ideas?

Not great, but better than nothing. In ASPECT we automatically generate a diff that can be applied directly. That would require some hacking around in the test suite to copy over files (not easy with different output variants...).

tests/simplex/step-17.cc
```diff
@@ -0,0 +1,487 @@
...
+#endif
+    , dof_handler(triangulation)
+  {}
+
+
+  template <int dim>
+  void
+  ElasticProblem<dim>::setup_system()
+  {
+    GridTools::partition_triangulation(n_mpi_processes, triangulation);
```
drwells, line 192: I'd bet that this is the cause of our problems - we default to using METIS here, but different versions of METIS produce significantly different output. Could you switch this to use `GridTools::partition_triangulation_zorder()` instead? That should produce the same output on all machines.

michelebucelli: Sure thing, I'll change it as you suggest. Thanks for pointing this out!

However, I'm not sure this is causing the error. The few lines of diff in the test log suggest that the test crashes altogether, rather than giving an output different from the provided one. Moreover, the serial test seems to be crashing as well... I think I'll try to get my hands on a different machine to try and reproduce the error.

bangerth:
Same here: This works on my local machine. Can you try and rebase to current master?

michelebucelli:
Done. It doesn't seem to have solved the issue

bangerth:
You're merging master into your branch, but that's not the right approach. Right now your branch has 306 commits when it should have only one. Can you try to *rebase* your branch?

michelebucelli:
Sorry, I believe I did rebase the branch onto master. I ran 
```
git checkout step-17-simplex
git rebase master
```
Is this the correct way to go? I'm not sure why it shows 306 commits and so many changes, sorry if I messed something up. I guess that given that the changes are so localized I might as well start from scratch.

bangerth:
It's hard to get rid of merge commits once you've introduced them. I believe that you only want to submit two patches, 2823c29 and eb8daed, right? If so, then try this:
```
  git checkout master
  git checkout -b v2
  git cherry-pick 2823c29 eb8daed
```
Then you can do either of the following things:
* Be cautious and do
```
  git push origin v2
```
and create a new pull request.
* Rename the old branch, name the new branch like the old branch, and push that one to replace what you have here:
```
  git checkout master
  git branch -m step-17-simplex step-17-simplex-backup
  git branch -D step-17-simplex
  git branch -m v2  step-17-simplex
  git push -f origin step-17-simplex  
```

masterleinad:
I would just do `git rebase upstream/master`. That should be good enough (fixing the merge conflict on the go).

michelebucelli:
Alright, I followed @bangerth's advice and opened a new pull request (#11757) with only the two commits.

peterrum:
This looks outdated due to #11795.

PR #11552, opened 2021-01-15
Step 40 as a simplex test

michelebucelli:
Step 40 (Laplace problem using a fully distributed triangulation) as a simplex test.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -85,6 +85,11 @@ jobs:
         make -j 2
     - name: test
       run: |
+        # Remove warning: "A high-performance Open MPI point-to-point
+        # messaging module was unable to find any relevant network
+        # interfaces."
+        export OMPI_MCA_btl_base_warn_component_unused='0'
+        
         cd build
         make -j 2 setup_tests_simplex
         ctest --output-on-failure -j 2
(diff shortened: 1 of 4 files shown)


peterrum (approved):
Excellent! Thanks a lot!

tamiko (approved):

bangerth:
The windows test failure is spurious, but on Linux, the test actually fails with a different output on 4 processors. @michelebucelli -- can you check that the test actually succeeds on your machine?

michelebucelli:
@bangerth I confirm the test passes on my machine. The same thing happens with my other pull request #11551 for step 17. I will be running the test on a different machine shortly to see if I can reproduce the error. I remark that also the test with a single process fails, and both tests appear to print some error message, instead of a valid but different than expected output.

bangerth:
Thanks for checking -- let us know what you find! We've been letting this PR sit for too long, and it would be good to get it merged!

michelebucelli:
I built and ran the test on a different machine and it passes there too. That machine uses the same set of dependency libraries (Boost, PETSc etc.) as my personal one, so if something is wrong due to some library version mismatch that wouldn't be caught.

bangerth:
I tried and it works on my local machine.

Can you try to rebase your commit on current "master"? This should also make the merge commits you currently have go away.

peterrum:
I'll try to wrap this PR up in the next days.

peterrum:
@michelebucelli Sorry for the delay! It has been some busy months/weeks. Your two PRs are very much appreciated!

michelebucelli:
@peterrum Don't worry about it, it's totally understandable! Thank you for the great work in developing dealii!

PR #11553, opened 2021-01-15
Generalize communicate_locally_moved_vertices() 

peterrum:
...for arbitrary distributed meshes. The new implementation is now based on `GridTools::exchange_cell_data_to_ghosts`.

I have been working on this together with @ifumagalli. Could you try the new function out in lifex if it indeed works for you.


ifumagalli:
Thank you very much, @peterrum !
I tested it both on p::distributed::T and p::fullydistributed::T and it works perfectly!

source/distributed/tria_base.cc
```diff
@@ -527,6 +527,56 @@ namespace parallel
...
+          vertices[v] = cell->vertex(v);
+        else
+          vertices[v] = invalid_point;
+
+      return vertices;
+    };
+
+    const auto unpack = [&](const auto &cell, const auto &vertices) {
+      for (const auto v : cell->vertex_indices())
+        if (std::isnan(vertices[v][0]) == false)
```
masterleinad, line 566:
```suggestion
        if (vertices[v] != invalid_point)
```
for symmetry with `pack`.

masterleinad (commented):
Looks good to me. Just a minor suggestion.

source/distributed/tria_base.cc
peterrum: This is not working. I have tried it out.

peterrum:
Has anyone an idea how to tackle the (is already defined) error?

source/distributed/tria_base.cc
masterleinad: What do you mean by "not working"?

masterleinad:
I pushed a fix.

source/distributed/tria_base.cc
peterrum: I get a `Floating point exception`.

peterrum:
@masterleinad Thanks for the fix! Ready from my side!

source/distributed/tria_base.cc
masterleinad: Hmm... Do you have a particular test that fails?

masterleinad (approved):

peterrum:
@masterleinad All `ctest -R mpi/communicate_moved_vertices` tests run parallel.

PR #11563, opened 2021-01-15
Work on higher-order MappingFE

peterrum:


source/fe/mapping_fe.cc
```diff
@@ -2142,17 +2141,21 @@ MappingFE<dim, spacedim>::compute_mapping_support_points(
...
+    fe->get_unit_support_points().size());
+
+  Table<2, double> weights;
+
+  cell->get_manifold().get_new_points(vertices,
+                                      weights,
+                                      mapping_support_points);
 
-  return a;
+  return mapping_support_points;
```
peterrum, line 2158: @luca-heltai I think this matches what you described. But what I am not sure about how to integrate line and quad manifolds here, like it is done here:

https://github.com/dealii/dealii/blob/9f9205129e30e174594e3103136c5d8f39a2c635/source/fe/mapping_q_generic.cc#L1544-L1639

luca-heltai: For the moment, put an assertion that is thrown if the manifold ids of the subobjects are different than the one of the cell. What you wrote works for objects where everybody has the same manifold id. 

The idea would be to loop over all subobjects, querying (if any) all the support points that live on the edges, quads/tri, hex/tet/etc first, and then using the same thing you did here with the vertices of the subobjects and the support points that live in the middle of the subobjects.

peterrum: I have added the code. What would be an appropriate manifold to test with?

luca-heltai: Hyper shell, or cylinder_shell. They would work with a polar manifold, or a cylindrical manifold.

peterrum:
Linear and quadratic mapping:

mapping-fe

FYI @luca-heltai @agrayver

luca-heltai:
So, the high order output is now implemented also for tria/tets? Nice!

:)

source/fe/mapping_fe.cc
```diff
@@ -2204,23 +2232,74 @@ MappingFE<dim, spacedim>::transform(
...
 
+namespace
+{
+  template <int spacedim>
+  bool
+  check_all_manifold_ids_identical(
+    const TriaIterator<CellAccessor<1, spacedim>> &)
+  {
+    return true;
+  }
```
luca-heltai, line 2243: Please add two blank spaces, both here and below.

luca-heltai (approved):
Minor changes.

peterrum:
> So, the high order output is now implemented also for tria/tets? Nice!

Indeed x2!

luca-heltai (approved):

source/grid/reference_cell.cc
```diff
@@ -101,6 +104,42 @@ namespace ReferenceCell
       }
   }
 
+  template <int dim>
```
kronbichler, line 107: Please add two empty lines before this to have 3 as usual.

source/grid/reference_cell.cc
```diff
@@ -101,6 +104,42 @@ namespace ReferenceCell
...
+  const ScalarPolynomialsBase<dim> &
+  get_linear_poly_space(const Type &reference_cell)
+  {
+    AssertDimension(dim, get_dimension(reference_cell));
+
+    static const TensorProductPolynomials<dim> poly(
+      Polynomials::generate_complete_Lagrange_basis(
+        QGaussLobatto<1>(2).get_points()));
```
kronbichler, line 115: You can have it slightly shorter and with fewer indirections in the implementation by
```suggestion
      Polynomials::LagrangeEquidistant::generate_complete_basis(1));
```

kronbichler (commented):
A few additional small comments.

include/deal.II/grid/reference_cell.h
```diff
@@ -507,6 +509,14 @@ namespace ReferenceCell
   make_triangulation(const Type &                  reference_cell,
                      Triangulation<dim, spacedim> &tria);
 
+
+  /**
+   * Return a linear nodal space matching the given reference cell.
+   */
+  template <int dim>
+  const ScalarPolynomialsBase<dim> &
+  get_linear_poly_space(const Type &reference_cell);
```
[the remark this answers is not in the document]
peterrum: You are right. I have removed the new function and extended the existing one.

kronbichler (approved):

PR #11565, opened 2021-01-15
Work on MappingFEField for simplex meshes

peterrum:
~For some reason, I get a *Arithmetic exception.* within `maybe_update_Jacobians()`...~


source/grid/reference_cell.cc
```diff
@@ -181,6 +181,57 @@ namespace ReferenceCell
...
+          quadrature = Quadrature<dim>(points);
+        }
+
+      return quadrature;
+    };
+
+    if (reference_cell == get_hypercube(dim))
+      {
+        static Quadrature<dim> quadrature;
+        return process(quadrature);
```
bangerth, line 210: This is not thread-safe. The initialization of static variables is guaranteed to be thread-safe, but not later accesses. You will want to write this as
```
  static const Quadrature<dim> quadrature = ...;
  return quadrature;
```

bangerth: I think a more obvious design would also pass the desired reference cell to the `process` function (which I might rename `create_quadrature(const ReferenceCell::Type &)` since the function does something different depending on what argument the *outer* function is being called with.

peterrum:
@luca-heltai @bangerth Seems to work :)

source/grid/reference_cell.cc
```diff
@@ -181,6 +181,52 @@ namespace ReferenceCell
...
+    const auto create_quadrature = [](const Type &reference_cell) {
+      Triangulation<dim> tria;
+      make_triangulation(reference_cell, tria);
+
+      std::vector<Point<dim>> points(tria.n_vertices());
+
+      for (unsigned int v = 0; v < points.size(); ++v)
+        points[v] = tria.get_vertices()[v];
+
+      return Quadrature<dim>(points);
```
luca-heltai, line 199: return 
```suggestion
      return Quadrature<dim>(tria.get_vertices());
```

luca-heltai (changes requested):

source/grid/reference_cell.cc
peterrum: You are right! How stupid of me!

luca-heltai: I do this kind of stuff all the time... :D

luca-heltai (approved):

PR #11567, opened 2021-01-16
Fix: single test output dir per test

peterrum:
The side effect of #11427 was that two folders per test has been create: one with the category as prefix, one without it. This PR fixes this.

### DIFF
--- a/cmake/macros/macro_deal_ii_add_test.cmake
+++ b/cmake/macros/macro_deal_ii_add_test.cmake
@@ -166,6 +166,7 @@ MACRO(DEAL_II_ADD_TEST _category _test_name _comparison_file)
 
         SET(_source_file "${_test_name}.cc")
 
+        SET(_target_short ${_test_name}.${_build_lowercase}) # target name
         SET(_target ${_category}.${_test_name}.${_build_lowercase}) # target name
         SET(_run_args "$<TARGET_FILE:${_target}>") # the command to issue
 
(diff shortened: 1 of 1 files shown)


cmake/macros/macro_deal_ii_add_test.cmake
```diff
@@ -273,17 +277,19 @@ MACRO(DEAL_II_ADD_TEST _category _test_name _comparison_file)
...
 
 
         ADD_EXECUTABLE(${_target} EXCLUDE_FROM_ALL
           ${_generated_files}
           ${_source_file}
-          ${CMAKE_CURRENT_BINARY_DIR}/${_target}/interrupt_guard.cc
+          ${CMAKE_CURRENT_BINARY_DIR}/${_target_short}/interrupt_guard.cc
           )
 
+        set_target_properties(${_target} PROPERTIES OUTPUT_NAME ${_target_short})
```
drwells, line 291:
```suggestion
        SET_TARGET_PROPERTIES(${_target} PROPERTIES OUTPUT_NAME ${_target_short})
```

drwells (changes requested):
Looks good aside from the minor style suggestion.

drwells:
Thanks for checking this out!

cmake/macros/macro_deal_ii_add_test.cmake
peterrum: OK. I have changed this!

drwells (approved):

kronbichler (approved):

bangerth:
Ah, this also explains why I recently had trouble finding the executable for tests as they were no longer located in the usual directory. I had to start searching for them. Thanks for fixing this again.

Issue #11535, opened 2021-01-14
set_all_manifold_ids does not work in simplices


lwy5515:
I am testing Step-6 using global refinement on simplices. Here I use `GridGenerator::convert_hypercube_to_simplex_mesh` to generate the triangular mesh. For example,

```
Triangulation<dim> tria, temp;
GridGenerator::hyper_ball(temp);
GridGenerator::convert_hypercube_to_simplex_mesh(temp, tria);
```

Note that `GridGenerator::convert_hypercube_to_simplex_mesh` would not copy the manifold_id to the output grid. So I need assign the manifold id again like

```
tria.set_all_manifold_ids(0);
tria.set_manifold(0, SphericalManifold<dim>());
```
It seems that this will produce exceptions. The problem is at line 1964 at tria_accessor.template.h
```
      case 2:
        // for quads also set manifold_id of bounding lines
        for (unsigned int i = 0; i < 4; ++i)
          this->line(i)->set_manifold_id(manifold_ind);
        break;
```
where the number 4 in the loop should be replaced by the number of bounding lines for the polygon.  Does anybody know how to fix that?

bangerth:
You will want to use `this->n_faces()` instead of the hard-coded `4`. Want to give this a try?

lwy5515:
Sure. Let me give it a try.

peterrum:
This is indeed an issue. @bangerth As a side remark:  we should merge #11389, since this will copy the manifold ids.

lwy5515:
I just tried as @bangerth suggested. Now I do not have any problems in the code but the output grid is not what I expect
<img width="413" alt="Capture" src="https://user-images.githubusercontent.com/13604144/104630608-3f7cdf80-569b-11eb-94a4-6119345ff520.PNG">

peterrum:
Similarly to #11484, you need to do the following:

https://github.com/dealii/dealii/blob/8562cf6f23c0ded6504fcabc4e79b3b77fa9a66b/tests/simplex/step-01.cc#L64-L67

This is not exactly a simplex mesh issue but an issue generally if we make copies of triangulations. Could you have a try by adding these lines?

bangerth:
@lwy5515  How does this look without further refinement? Just the coarse mesh?

lwy5515:
@bangerth  The above grid is obtained from a 4-time-global refinement of the coarse mesh. The red part has the manifold id 0 and the blue is `flat_manifold_id`.

[...]

PR #11569, opened 2021-01-16
A simplex test for step-38

lwy5515:
The results can be found at https://github.com/dealii/dealii/issues/11535#issuecomment-761589169. Note that this test only checks the linear elements along with the linear FE mapping. The test for the quadratic elements requires the quadratic FE mapping.

### DIFF
--- a/tests/simplex/step-38.cc
+++ b/tests/simplex/step-38.cc
@@ -1,0 +1,445 @@
+/* ---------------------------------------------------------------------
+ *
+ * Copyright (C) 2010 - 2020 by the deal.II authors
+ *
+ * This file is part of the deal.II library.
+ *
+ * The deal.II library is free software; you can use it, redistribute
+ * it, and/or modify it under the terms of the GNU Lesser General
+ * Public License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ * The full text of the license can be found in the file LICENSE.md at
+ * the top level directory of deal.II.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+// Test step-38 on a simplex mesh.
+// Note that currently the quadratic mapping for simplices is not implemented.
+
+#include <deal.II/base/function.h>
+#include <deal.II/base/quadrature_lib.h>
+
+#include <deal.II/dofs/dof_accessor.h>
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/fe/fe_q.h>
+#include <deal.II/fe/fe_values.h>
+#include <deal.II/fe/mapping_fe.h> // new header
+#include <deal.II/fe/mapping_q.h>
+
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/manifold_lib.h>
+#include <deal.II/grid/tria.h>
+
+#include <deal.II/lac/dynamic_sparsity_pattern.h>
+#include <deal.II/lac/full_matrix.h>
+#include <deal.II/lac/precondition.h>
+#include <deal.II/lac/solver_cg.h>
+#include <deal.II/lac/solver_control.h>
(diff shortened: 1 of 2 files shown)


tests/simplex/step-38.cc
```diff
@@ -0,0 +1,445 @@
...
+      GridGenerator::extract_boundary_mesh(volume_mesh,
+                                           tria_temp,
+                                           boundary_ids);
+      tria_temp.set_all_manifold_ids(0);
+      tria_temp.set_manifold(0, SphericalManifold<dim, spacedim>());
+      GridGenerator::convert_hypercube_to_simplex_mesh(tria_temp,
+                                                       triangulation);
+      for (const auto i : tria_temp.get_manifold_ids())
+        if (i != numbers::flat_manifold_id)
+          triangulation.set_manifold(i, tria_temp.get_manifold(i));
```
peterrum, line 240: I am wondering if the following also works:
```suggestion
      Triangulation<spacedim> volume_mesh_temp;
      GridGenerator::convert_hypercube_to_simplex_mesh(volume_mesh, volume_mesh_temp);
      GridGenerator::extract_boundary_mesh(volume_mesh,
                                           triangulation,
                                           boundary_ids);
      triangulation.set_all_manifold_ids(0);
      triangulation.set_manifold(0, SphericalManifold<dim, spacedim>());
```

tests/simplex/step-38.cc
```diff
@@ -0,0 +1,445 @@
...
+  }
+
+  template <int spacedim>
+  LaplaceBeltramiProblem<spacedim>::LaplaceBeltramiProblem(
+    const unsigned degree)
+    : dof_handler(triangulation)
+    , fe(degree)
+#ifdef USE_SIMPLEX
+    // Quadratic mapping for simplex is still not implemented.
+    , mapping(Simplex::FE_P<dim, spacedim>(degree))
```
peterrum, line 212: We are working on that! See #11563.

peterrum (approved):
Thanks a lot @lwy5515!

tests/simplex/step-38.cc
lwy5515: This should not work because the routine `GridGenerator::extract_boundary_mesh` currently relies on  'GeometryInfo' and does not work for simplex meshes in 3D. I can think about how to remove `GeometryInfo` from this routine (maybe in another PR).

tests/simplex/step-38.cc
lwy5515: That's nice. I will try to test quadratic when #11563 is ready to get merged.

tests/simplex/step-38.cc
peterrum: >  I can think about how to remove GeometryInfo from this routine (maybe in another PR).

That would be great!

peterrum:
@lwy5515 Your commit is displayed with no avatar. Would you mind to add your email address used locally by git to your GitHub account. If you do that than your commits will also be counted correctly (see: https://github.com/dealii/dealii/pulse). Thx!

tests/simplex/step-38.cc
peterrum: @lwy5515 It's on master :)

lwy5515:
@peterrum I just tested step-38 with quadratic elements. Here is the result:
```
Surface mesh has 10240 cells.
Surface mesh has 20609 degrees of freedom.
H1 error = 0.00932553
Surface mesh has 40960 cells.
Surface mesh has 82177 degrees of freedom.
H1 error = 0.00234109
```
The approximation converges in the second order. I think the code is working. I would like to note that in the original code the order of `QGauss` is set to be `2 * fe.degree`. This is not available in the current version.

peterrum:
@lwy5515 Excellent! Thank you. Would you like to amend the simplex version of step-38 so that it also runs with quadratic mapping?

lwy5515:
@peterrum  Sure.

PR #11570, opened 2021-01-16
Return the box of support points instead of only the vertices.

luca-heltai:


luca-heltai:
Without this patch, the bounding boxes for our beloved hypersphere look like:

Screenshot 2021-01-16 at 18 11 28

With this patch we get for quadratic:

Screenshot 2021-01-16 at 18 52 46

of course, if we use cubic the situation is worse, since the mid point is not a support point:

Screenshot 2021-01-16 at 18 58 19

luca-heltai:
@blaisb , out of curiousity, how did you get the nice pictures of the bounding boxes with paraview? I don't seem to be able to plot them with lines if I make them transparent, and the ones in the background never show up with their boundaries...

peterrum (approved):

drwells (approved):

blaisb:
> @blaisb , out of curiousity, how did you get the nice pictures of the bounding boxes with paraview? I don't seem to be able to plot them with lines if I make them transparent, and the ones in the background never show up with their boundaries...

I realized I missed this question. Sorry Luca. Never made pictures of the bounding box with paraview, I just outputed the simulation subdomain.

PR #11571, opened 2021-01-17
Add new test for MappingFEField for simplex meshes

peterrum:
With the following result with linear and quadratic mapping:

mapping-fe-fields

FYI @agrayver


peterrum:
Failing test is unrelated and is fixed in #11573.

tests/simplex/mapping_fe_fields_02.cc
```diff
@@ -0,0 +1,123 @@
...
+
+  FEValues<dim> fe_values(mapping,
+                          fe,
+                          quadrature_formula,
+                          update_values | update_gradients);
+
+  for (const auto &cell : dof_handler.active_cell_iterators())
+    {
+      fe_values.reinit(cell);
+    }
```
kronbichler, line 95: Is this really needed? `DataOut::build_patches` should also call this function (using `update_values` and `update_quadrature_points` as arguments).

tests/simplex/mapping_fe_fields_02.cc
```diff
@@ -0,0 +1,123 @@
...
+}
+
+int
+main()
+{
+  initlog();
+
+  test(1); // linear mapping
+  test(2); // quadratic mapping
+}
```
kronbichler, line 123: Do we need a newline here?

peterrum:
@kronbichler Done!

kronbichler (approved):

PR #11572, opened 2021-01-17
Generalize VectorTools::get_position_vector for simplex meshes

peterrum:


peterrum:
Failing test is unrelated and is fixed in #11573.

include/deal.II/numerics/vector_tools_interpolate.templates.h
```diff
@@ -653,8 +653,12 @@ namespace VectorTools
       {
         const Quadrature<dim> quad(fe.get_unit_support_points());
 
-        MappingQGeneric<dim, spacedim> map_q(fe.degree);
-        FEValues<dim, spacedim> fe_v(map_q, fe, quad, update_quadrature_points);
+        const auto map_q = ReferenceCell::get_default_mapping<dim, spacedim>(
+          fe.reference_cell_type(), fe.degree);
+        FEValues<dim, spacedim>              fe_v(*map_q,
```
drwells, line 658: I'm not going to hold up the PR for it, but is it just me or is clang-format getting weirder?

drwells (approved):
I think this is fine in isolation but it doesn't match my recollection of what we agreed on for implementing the geometry-agnostic replacement for `MappingQ1` last Friday. Did we write that down somewhere so I can check my memory?

drwells:
On the other hand this is consistent with #11539 so, even if we want to do something else later, we make the implementation more uniform by merging both so lets merge.

PR #11579, opened 2021-01-18
Use both quadratic element and quadrature mapping in simplex/step-38

lwy5515:
According to https://github.com/dealii/dealii/pull/11569#discussion_r559045899, this is a test for #11563. Now the default order of the FE element and mapping is set to be 2 in simplex/step-38.


peterrum (approved):
Thank you!

PR #11581, opened 2021-01-18
Add step-3simplex and step-3mixed

peterrum:
The tutorial graph could like this with the new tutorials:

path


examples/step-3simplex/doc/results.dox
```diff
@@ -0,0 +1,102 @@
...
+In the general case (also known as mixed mesh), the mesh consists of both
+cell types. In 3D, meshes might even consist of more cell types, like
+wedges/prisms and pyramids. We consider such meshes in the tutorial step-3mixed.
+
+
+<h4>Alternative finite elements, quadrature rules, and mapping objects</h4>
+
+In this tutorial, we used the most basic finite-element, quadrature-rule, and
+mapping classes. However, more classes are compatible with simplices. The
+following list gives an overview of these classes:
```
bangerth, line 51: All of these are used in later tutorial programs. If you make it here from step-3, this is all going to be over your head. I would remove this whole section.

examples/step-3simplex/doc/results.dox
```diff
@@ -0,0 +1,102 @@
...
+- finite elements: FE_SimplexP, FE_SimplexDGP, FE_SimplexP_Bubbles
+- quadrature rules: QGaussSimplex, QWitherdenVincentSimplex, QDuffy
+- mapping objects: MappingFE, MappingFEField
+
+It should be also pointed out that FESystems can also handle simplex finite
+elements which is crucial to solve vector-valued problems, as needed, e.g., to
+solve elasticity and fluid problems (see also step-17).
```
bangerth, line 58: Same here -- readers will not be able to understand what all of this is about.

examples/step-3simplex/doc/results.dox
```diff
@@ -0,0 +1,102 @@
...
+point out the following functions:
+ - GridGenerator::subdivided_hyper_cube_with_simplices() and
+   GridGenerator::subdivided_hyper_rectangle_with_simplices(), which fill a
+   hypercube and a hyperrectangle domain with simplices
+ - GridGenerator::convert_hypercube_to_simplex_mesh(), which converts meshes
+   consisting of hypercube cells to simplex meshes by replacing one
+   quadrilateral with 4 triangles and one hexahedron with 24 tetrahedrons
+
+
+<h4>hp-adaptivity</h4>
```
bangerth, line 77: same for this and the parallelization section -- this is step-3, not somewhere in the step-2x range.

bangerth (commented):
Just a few comments on the step-3simplex. I would be much more interested in putting this into the results section of step-3.

In general, my suggestion would be for us to wait with these tutorials until the simplex stuff is on by default. Then:
* Add something to step-1 that shows how to generate simplex or mixed meshes -- step-1 is already about meshes, so that would fit well there. A `third_mesh()` function would not be out of place.
* Add something to step-3's results section that talks about the changes necessary to switch to simplices.
* A new tutorial program (not using the step-3 prefix) that deals with mixed meshes.

I think there is just not enough new stuff here, and too much code duplication, to justify a new tutorial program for simplex meshes.

peterrum:
What are the feelings now related to a simplex-/mixed-mesh tutorial?

blaisb:
I really like the idea, either as a separate tutorial or as an appendix to step-3... Simplex and mixed mesh are much better supported features now.

peterrum:
Closing, since there seems no interest :disappointed: ...

bangerth:
How do we want to proceed here? #17001 and #17003 add discussions of triangles to steps 1 and 3: https://dealii.org/developer/doxygen/deal.II/step_1.html#step_1-Possibilitiesforextensions and https://dealii.org/developer/doxygen/deal.II/step_3.html#step_3-Possibilitiesforextensions . I continue to not want to add tutorial programs that are in all practicality duplicates of existing ones. I think that the discussions in the links above are sufficient.

blaisb:
@bangerth I understand your point. However, many people still do not think that deal.II support triangles and tets meshes. If I may propose an alternative point of view, sometimes we see things as developers as "duplicates" when from a user perspective, they may not be. For example, these examples would be an easy way to understand simplices without requiring an understanding of anything else. Furthermore, they also introduce the key differences with respect to the quadratures and the mapping, which is not trivial for a user to understand.

Maybe my philosophy on this is very different, but I really like very simple or incremental examples that highlight a single feature. I think these two accomplish just that. At the very least, I really think we need simplex examples to showcase to people not only that we can do it, but how easy it is to do it. Maybe the fact that the example is that common with the regular step-3 is just a testament to how flexible the architecture is. One could even add a comment similar to "You will notice this step is highly similar to step-3, this is because it is relatively easy to make a deal.II-derived application that supports all elements without dramatic changes!".

I think it would be helpful to people, at least, I think 5 years-ago me would have liked such a step :). I think my students would like it too.

drwells:
Yes - it was news to many people at the PETSc workshop that we now support simplices and mixed meshes.

For the early tutorials I think it would be better to just solve the same problem with multiple element types - e.g., `Step3::Step3()` could be

```cpp
Step3::Step3(const ReferenceCell &reference_cell)
   : fe(reference_cell.is_simplex() ? std::make_unique<FE_SimplexP<dim>>(1) : std::make_unique<FE_Q<dim>>(1))
{
    // etc
}
```
or something similar. I'd also prefer to create simplex grids directly in step-1.

blaisb:
> Yes - it was news to many people at the PETSc workshop that we now support simplices and mixed meshes.
> 
> For the early tutorials I think it would be better to just solve the same problem with multiple element types - e.g., `Step3::Step3()` could be
> 
> ```c++
> Step3::Step3(const ReferenceCell &reference_cell)
>    : fe(reference_cell.is_simplex() ? std::make_unique<FE_SimplexP<dim>>(1) : std::make_unique<FE_Q<dim>>(1))
> {
>     // etc
> }
> ```
> 
> or something similar. I'd also prefer to create simplex grids directly in step-1.

I agree with this, but for a first time user that first looks at step-3, it will be very confusing. I don't know, I don't think having two other small steps which are derived from step-3 is such a big deal no? I mean if people are not interested, they just don't look, but if they are, they can be like: " Hey that's so cool, I can do what I did but with simplices and with mixed mesh, nice!". It's simple and for a very new person, this will be trivial to understand. I find that simple intuitive idea, even if they come with some duplication, really help new users.

I always go back to step-68 when I talk about particles and a lot of people write to me regarding it. It's a super dumb step Rene and I made. It does nothing but make particles and move them. However, for most people, that's all they wish to learn. It suits their application perfectly and it demonstrates a small feature set in a very accessible manner.

tjhei:
> I agree with this, but for a first time user that first looks at step-3, it will be very confusing.

I agree. I am in favor of adding one or two simplex tutorials and it is fine with me if they basically revisit the same problem.

kronbichler:
I also think that having dedicated simplex tutorials would be helpful as we discussed in #16944. I won't repeat the arguments there, but I guess the `diff` approach we are nowadays for testing tutorials might also help us if we start accumulating more code paths like this one, to avoid code duplication.

bangerth:
You all should probably just have a poll and overrule me. But just to be clear:

What I mean to say is this:
* We *already* talk about triangles. In step-1, the results section shows quadrilaterals for 1/2 of a page, and then spends two pages showing how to do triangles, see https://dealii.org/developer/doxygen/deal.II/step_1.html#step_1-Usetriangles . We also talk about triangles and mixed meshes in the introduction of step-1.
* In step-3, the results section is ~1 page, and then there is a 2-page discussion about what one needs to do to switch to triangles, see https://dealii.org/developer/doxygen/deal.II/step_3.html#step_3-Possibilitiesforextensions

We cannot force those who learned about deal.II 10 years ago to re-read these tutorial programs, but they are also not going to read any new programs we create. The material is there, and everyone who is new cannot avoid seeing it in the same places where they see quadrilaterals.

From an educational perspective, my thinking is that *contrasting* approaches is the right approach, as we already do in steps 1 and 3. It is less helpful to show people a *separate* version of the program because they wouldn't see the differences.

Anyway, I think I said what I wanted to say. I'm not going to object if others think that we should instead merge this patch.

drwells:
We've run out of time to do this for 9.7. @bangerth raises a lot of valid points. I need to think about this one some more.

drwells (approved):
I don't disagree with those comments, but we have zero tutorial programs which, from start-to-finish, actually use mixed meshes or simplices. We should add some.

I'm in favor of making step-39 a full "this is how multiple reference cell types work" but adding this is a step forward.

blaisb:
@peterrum Do you need help wrapping up the tutorial since I think there's a form of "agreement" towards it? I can help you with it if you need help.

tjhei:
I have a suggestion for a compromise:
Some of us would like a dedicated program they can point towards that uses simplex/mixed meshes. The sticking point about this PR seems to be that a) mixed meshes are more complicated and require additional knowledge that goes beyond step-3, b) this tutorial otherwise doesn't provide anything "new" over the text in step-3 in the results section.

If we instead are okay to consider mixed meshes a somewhat more advanced topic, we could change it into a matrix-based but MPI parallel code. It would necessarily depend on step-40 and therefore indirectly on step-6. That means parallelization, hp-adaptivity, etc. would no longer be a new concept. This would also satisfy one of the goals we discussed a while ago: as steps are the often used as a foundation for own codes, more should be MPI parallel.

Let me know if that makes sense.

drwells:
My opinion on what to do here hasn't changed since last year.

Yes, I'd like to use my slot (step-84) to go over all the nice things `ReferenceCell` can do to make this easier (and show off our simplex support). I envision that tutorial as focused on `ReferenceCell` so there is certainly more to say about mixed meshes, `p::fd::T`, and other relevant topics (i.e., lots of room for more tutorial programs). Managing the CI is a higher priority, though, so I won't have time to write that tutorial until after the release is out.

>  This would also satisfy one of the goals we discussed a while ago: as steps are the often used as a foundation for own codes, more should be MPI parallel.

Yes, students find the switch between `Vector` and `PETScWrappers::MPI::Vector`, `LinearAlgebra::distributed::Vector`, etc. hard (a student told me they found `LinearAlgebra::distributed::Vector` intimidating, which is a real bummer). My experience is that students who start with a parallel tutorial as a template tend to fare better, so we should make MPI-something the default for higher-number tutorials to skip the switching step. (I am jealous of PETSc here, where a `Vec` is a `Vec` and a `Mat` is a `Mat` and you can just switch out types and 'everything works': our own class situation is not as nice)

Issue #11585, opened 2021-01-19
Support high-order Lagrange VTK in DataOut


agrayver:
Track the effort of adding support for this feature here. PRs are coming...

peterrum:
I think to get this working for the coming release is not possible anymore. Let's tackle this PR again after the release.

peterrum:
@agrayver Would you be interested to continue here? It would be nice to support for the other cell shapes!

saitoasukakawaii:
> @agrayver Would you be interested to continue here? It would be nice to support for the other cell shapes!

currently, the dealii supprot the high order cell by set the VtkFlags to [write_higher_order_cells](https://www.dealii.org/developer/doxygen/deal.II/structDataOutBase_1_1VtkFlags.html#aa9dd1830c0ff35a2431704c4d45453eb:~:text=Flag%20determining%20whether%20to%20write%20patches%20as%20linear%20cells%20or%20as%20a%20high%2Dorder%20Lagrange%20cell). I want to know whether it only support high order cell of Quadrilateral cell (2d) and Hexahedron cell (3d)?

bangerth:
@saitoasukakawaii It's not clear to me from looking at the code. What happens if you try?

PR #11588, opened 2021-01-19
Enable VtkFlags::write_higher_order_cells for simplices

peterrum:
High-order VTK output is the last open issue from the hackathon. This PR (co-written by @agrayver) adds the capabilities to triangles so that both 2D simplex meshes and 2D mixed meshes can be written in this way.  

References #11585.


include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -83,6 +83,98 @@ namespace internal
...
+    {
+      std::vector<Point<2>> evalution_points;
+
+      generate_simplex_evaluation_points_recursivly(
+        {{Point<2>(0.0, 0.0), Point<2>(1.0, 0.0), Point<2>(0.0, 1.0)}},
+        n_subdivisions,
+        evalution_points);
+
+      return evalution_points;
+    }
```
peterrum, line 174: We need some kind of `QIterated` for arbitrary reference-cell types.  For now, I am position the evaluation points as depicted in the documentation of VTK (https://blog.kitware.com/wp-content/uploads/2018/09/Source_Issue_43.pdf). However, we might want a different ordering.

agrayver:
It seems you need to rebase...

source/base/data_out_base.cc
```diff
@@ -2700,6 +2705,22 @@ namespace DataOutBase
 
     for (const auto &patch : patches)
       {
+        if (patch.reference_cell_type != ReferenceCell::get_hypercube(dim))
+          {
+            connectivity.resize(patch.data.n_cols());
+
+            for (unsigned int i = 0; i < patch.data.n_cols(); ++i)
+              connectivity[i] = i;
```
peterrum, line 2713: If we choose to have a different ordering, we need to adjust the permutation here.

peterrum (commented):
I would like to point out to places of the code worth discussing.

peterrum:
My suggestion would be to proceed here. We can work on `QIterated` and appropriate permutations in follow-up PRs. I have noted the issue in #11352.

agrayver:
LGTM. 3-D part will follow in a separate PR.

bangerth:
Please rebase.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -83,6 +83,98 @@ namespace internal
...
+          return;
+        }
+
+      for (const auto &p : bounding_vertices)
+        evalution_points.push_back(p);
+
+      if (n_subdivisions == 1)
+        return;
+
+      const auto process = [&](const Point<2> &p0, const Point<2> &p1) {
```
bangerth, line 124: Maybe call this function `generate_inbetween_points()`, which I think would make it more obvious what it does.

bangerth: I think it would also be good to write a comment that explains what the function does. It has a bit of a complicated logic: It *both* returns an array, but *also* appends what it returns to the existing `evaluation_point`.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -83,6 +83,98 @@ namespace internal
...
+
+        evalution_points.insert(evalution_points.end(),
+                                line_points.begin(),
+                                line_points.end());
+
+        return line_points;
+      };
+
+      const auto line_points_0 =
+        process(Point<2>(bounding_vertices[0]), Point<2>(bounding_vertices[1]));
```
bangerth, line 138:
```suggestion
        process(bounding_vertices[0], bounding_vertices[1]);
```
Same for the lines below

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -83,6 +83,98 @@ namespace internal
...
+        process(Point<2>(bounding_vertices[1]), Point<2>(bounding_vertices[2]));
+      const auto line_points_2 =
+        process(Point<2>(bounding_vertices[2]), Point<2>(bounding_vertices[0]));
+
+      if (n_subdivisions == 2)
+        return;
+
+      generate_simplex_evaluation_points_recursivly(
+        {{Point<2>(line_points_0[line_points_0.size() - 2][0],
+                   line_points_1[0][1]),
```
bangerth, line 149: can you explain in a comment what this point represents?

include/deal.II/numerics/data_out_dof_data.templates.h
bangerth: What would be the difficulty in just generalizing `QIterated`? We'll need that one day anyway.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -140,9 +232,13 @@ namespace internal
...
-              quadrature_simplex = std::make_unique<Quadrature<dim>>(
-                Simplex::FE_P<dim, spacedim>(n_subdivisions)
-                  .get_unit_support_points());
+              if (dim == 2)
+                quadrature_simplex = std::make_unique<Quadrature<dim>>(
+                  generate_simplex_evaluation_points<dim>(n_subdivisions));
+              else
+                quadrature_simplex = std::make_unique<Quadrature<dim>>(
+                  Simplex::FE_P<dim, spacedim>(n_subdivisions)
+                    .get_unit_support_points());
```
bangerth, line 241: Is it true that you're basically just special casing the 2d case because it's reasonably simple to do and because creating the `FE_Q` is more expensive?

source/base/data_out_base.cc
```diff
@@ -2730,6 +2735,23 @@ namespace DataOutBase
...
+
+            out.template write_high_order_cell<dim>(count++,
+                                                    first_vertex_of_patch,
+                                                    connectivity);
+
+            first_vertex_of_patch += patch.data.n_cols();
+
+            continue;
```
bangerth, line 2752: I find `continue` statements inside `if` difficult to read. Just drop the `continue` here and make the rest of the body of the loop into the `else`. Or better yet, make the guarding `if` condition a positive and put the existing code into `if (cell = hypercube)` and put the new code at the bottom with `else`.

bangerth: If you do this right, the `if`...`else` would only contain setting up the data structures, and the call to `write_high_order_cell` and `first_vertex_of_patch += ...` would remain outside.

tests/simplex/data_out_write_vtk_01.cc
```diff
@@ -103,24 +111,38 @@ main()
...
-    test<dim>(FESystem<dim>(Simplex::FE_P<dim>(2 /*=degree*/), dim), dim);
-    test<dim>(FESystem<dim>(Simplex::FE_P<dim>(2 /*=degree*/),
-                            dim,
-                            Simplex::FE_P<dim>(1 /*=degree*/),
-                            1),
-              dim + 1);
-  }
+  for (unsigned int i = 0; i < 2; ++i)
+    {
+      const bool do_high_order = i == 1;
```
bangerth, line 116:
```suggestion
      const bool do_high_order = (i == 1);
```

tests/simplex/data_out_write_vtk_01.cc
```diff
@@ -103,24 +111,38 @@ main()
...
+                    do_high_order);
+          test<dim>(FESystem<dim>(Simplex::FE_P<dim>(2 /*=degree*/),
+                                  dim,
+                                  Simplex::FE_P<dim>(1 /*=degree*/),
+                                  1),
+                    dim + 1,
+                    do_high_order);
+        }
+
+      if (do_high_order == false /*TODO*/)
```
bangerth, line 133: can you explain what the TODO refers to?

tests/simplex/data_out_write_vtk_01.cc
```diff
@@ -103,24 +111,38 @@ main()
...
-    test<dim>(Simplex::FE_P<dim>(2) /*=degree*/, 1);
-    test<dim>(FESystem<dim>(Simplex::FE_P<dim>(2 /*=degree*/), dim), dim);
-    test<dim>(FESystem<dim>(Simplex::FE_P<dim>(2 /*=degree*/),
-                            dim,
-                            Simplex::FE_P<dim>(1 /*=degree*/),
-                            1),
-              dim + 1);
-  }
-  {
-    const unsigned int dim = 3;
```
bangerth, line 117: If I see this right, then you are dropping the 3d part of the test. Is this on purpose? That would be unfortunate because we then no longer test this functionality...

tests/simplex/data_out_write_vtk_02.cc
```diff
@@ -0,0 +1,163 @@
...
+
+      data_out.attach_dof_handler(dof_handler);
+      data_out.add_data_vector(solution, "solution");
+
+
+      data_out.build_patches(mapping, n_subdivisions);
+
+#if false
+  std::ofstream output("test." + std::to_string(dim) + "." + std::to_string(counter++) + ".vtk");
+  data_out.write_vtk(output);
```
bangerth, line 129: can you indent manually?

bangerth: or just get rid of the whole thing

tests/simplex/data_out_write_vtk_02.cc
```diff
@@ -0,0 +1,163 @@
...
+}
+
+int
+main()
+{
+  initlog();
+
+  for (unsigned int i = 0; i < 2; ++i)
+    {
+      const bool do_high_order = i == 1;
```
bangerth, line 143:
```suggestion
      const bool do_high_order = (i == 1);
```

tests/simplex/data_out_write_vtk_02.cc
```diff
@@ -0,0 +1,163 @@
...
+
+      if (true)
+        {
+          const unsigned int dim = 2;
+          test<dim>(Simplex::FE_P<dim>(2), FE_Q<dim>(2), 1, do_high_order);
+
+          continue; // TODO
```
bangerth, line 150: you're skipping the code below. we'll never get around to fixing this. Just remove the code below -- if you ever want to implement this, just add another test.

bangerth: separately, though, why does the code below not work?

tests/simplex/data_out_write_vtk_01.cc
peterrum: No, I am not. Diff is only confused by the new indentation.

tests/simplex/data_out_write_vtk_02.cc
peterrum: This is on my TODO list. For some reason, I get an error during the evaluation by `FEValues` for systems on mixed meshes. I have a feeling what is the reason but did not have time for this in the last days (vacations).

bangerth:
OK to merge if you open github issues for the things that you think still need to be fixed (such as the TODO in the test).

peterrum:
I'll push in a couple of minutes some changes.

peterrum:
@bangerth I have integrate your suggestions. I agree that we should generalize `QIterated` for arbitrary reference-cell types. For example, that would simplify the code a bit in `setup_parallel_data_base_internal`:

```cpp
Triangulation<dim, spacedim> dummy;
reference_cell.make_triangulation(dummy);
quadrature_XY = std::make_unique<QIterated<dim>>(
    Quadrature<dim>(dummy.get_vertices()), reference_cell, n_subdivisions)
```

But lets keep this for a follow-up PR. This topic and the TODO is part of the list of https://github.com/dealii/dealii/issues/11352#issue-761425796.

bangerth (approved):
Please squash and merge once the testers are happy.

PR #11590, opened 2021-01-19
simplex: face_interpolation for FE_Poly

marcfehling:
Implements `get_face_interpolation_matrix` and `get_subface_interpolation_matrix` for `FE_Poly` in 2D.

With this PR, hanging node constraints and thus adaptive methods will be possible in 2D. I'll work on porting `step-6` for triangles once this gets merged.


marcfehling:
The p-refinement case seems to work, but the constraints for `h-refinement` are wrong.

According to the `GridOut` output in `tests/simplex/hanging_nodes_01`, vertices (and I assume dofs as well) are enumerated as follows:
```
2-------3
|\      |
|  \    |
6---5   |
|\  |\  |
|  \|  \|
0---4---1
```
This should yield a constraint `u_5 = (u_1+u_2)/2`, but actually gives `u_6 = (u_1+u_2)/2`.

@peterrum -- Did I get the subface numbering in `get_subface_interpolation_matrix` right?

peterrum (commented):
Looks OK. I am not sure if the reason are the interpolation matrices. Could you check if the indices are correct:

https://github.com/dealii/dealii/blob/4318ade62b8a8796d2f7170f7e257069cc5d8528/source/dofs/dof_tools_constraints.cc#L1461-L1464

BTW: would you mind moving the p-adaptivity stuff into a new PR so that we can merge that.

marcfehling:
There was nothing wrong with the constraints. I simply counted them wrong on paper. With some extra diagnostics I was able to find the correct enumeration. Ready for review.

Please find ASCII pictures of the dof enumeration in the preamble of each test.

source/base/qprojector.cc
```diff
@@ -285,8 +310,9 @@ QProjector<2>::project_to_subface(const ReferenceCell::Type reference_cell_type,
                                   std::vector<Point<2>> &   q_points,
                                   const RefinementCase<1> &)
 {
-  Assert(reference_cell_type == ReferenceCell::Type::Quad, ExcNotImplemented());
-  (void)reference_cell_type;
+  Assert(reference_cell_type == ReferenceCell::Type::Tri ||
+           reference_cell_type == ReferenceCell::Type::Quad,
+         ExcNotImplemented());
```
marcfehling, line 315: @peterrum -- This is not needed as well I assume.

peterrum: Yes.

source/simplex/fe_lib.cc
```diff
@@ -369,6 +371,156 @@ namespace Simplex
...
+                                x_source_fe.n_dofs_per_face(face_no)));
+
+    if (const FE_Poly<dim, spacedim> *source_fe =
+          dynamic_cast<const FE_Poly<dim, spacedim> *>(&x_source_fe))
+      {
+        const Quadrature<dim - 1> quad_face_support(
+          source_fe->get_unit_face_support_points(face_no));
+
+        double eps = 2e-13 * this->degree * (dim - 1);
+        // TODO: Check applicability to triangles
```
marcfehling, line 392: here

source/simplex/fe_lib.cc
```diff
@@ -369,6 +371,156 @@ namespace Simplex
...
+           ExcDimensionMismatch(interpolation_matrix.m(),
+                                x_source_fe.n_dofs_per_face(face_no)));
+
+    if (const FE_Poly<dim, spacedim> *source_fe =
+          dynamic_cast<const FE_Poly<dim, spacedim> *>(&x_source_fe))
+      {
+        const Quadrature<dim - 1> quad_face_support(
+          source_fe->get_unit_face_support_points(face_no));
+
+        double eps = 2e-13 * this->degree * (dim - 1);
```
marcfehling, line 462: here

marcfehling (commented):
I just stumbled over this TODO that I added in preparing this PR. I do not know if that is a correct measure for triangles, as I simply copied that from the quad implementation. But the tests pass so I think this is okay.

masterleinad:
```
/home/runner/work/dealii/dealii/source/base/qprojector.cc:607:42: error: ‘permute_according_orientation’ is not a member of ‘dealii::ReferenceCell’
         const auto temp = ReferenceCell::permute_according_orientation(
                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

peterrum (approved):
Ready to go once you have fixed the compilation error :+1:  As a next step you could work on the 3D p-case. Local refinements in 3D is still not working, yet :( I need to find some time to wrap my PR up.

marcfehling:
> As a next step you could work on the 3D p-case. Local refinements in 3D is still not working, yet :( I need to find some time to wrap my PR up.

It is on my todo list, but there are currently other things with higher priority right now... Will work on porting `step-6` next to finish the 2D tri-refinement chapter.

marcfehling:
ping

source/simplex/fe_lib.cc
```diff
@@ -370,6 +372,155 @@ namespace Simplex
...
+            for (unsigned int i = 0; i < this->n_dofs_per_face(face_no); ++i)
+              sum += interpolation_matrix(j, i);
+
+            Assert(std::fabs(sum - 1) < eps, ExcInternalError());
+          }
+#endif
+      }
+    else if (dynamic_cast<const FE_Nothing<dim> *>(&x_source_fe) != nullptr)
+      {
+        // nothing to do here, the FE_Nothing has no degrees of freedom anyway
```
bangerth, line 433: This is actually wrong based on our conversation the other day, correct? This should only be the case if one had the non-dominating `FE_Nothing`.

Nothing for you to do here, just asking.

source/simplex/fe_lib.cc
```diff
@@ -370,6 +372,155 @@ namespace Simplex
...
+           ExcDimensionMismatch(interpolation_matrix.m(),
+                                x_source_fe.n_dofs_per_face(face_no)));
+
+    if (const FE_Poly<dim, spacedim> *source_fe =
+          dynamic_cast<const FE_Poly<dim, spacedim> *>(&x_source_fe))
+      {
+        const Quadrature<dim - 1> quad_face_support(
+          source_fe->get_unit_face_support_points(face_no));
+
+        double eps = 2e-13 * this->degree * (dim - 1);
```
bangerth, line 462:
```suggestion
        const double eps = 2e-13 * this->degree * (dim - 1);
```

bangerth (commented):
The case that I'm worried about is where we accidentally get the orientation of an edge wrong. You'd see that with a P3 element, but we don't currently have that yet. So I guess that will have to wait.

Just minor stuff.

source/simplex/fe_lib.cc
marcfehling: Good point! Let's keep this section in mind.

Issue #11583, opened 2021-01-19
Make gmsh format store manifold ids.


luca-heltai:
msh version > 4 allows to store arbitrary tags along with each mesh object. It would be pretty straight forward to allow saving (and loading) manifold ids with msh using those tags.

blaisb:
This would be an amazing feature

bangerth:
I so so wished that the gmsh people had come up with a self-describing file format. The way attributes are attached to objects is just so completely opaque: There is no place to *describe* what an attribute actually means...

luca-heltai:
> I so so wished that the gmsh people had come up with a self-describing file format. The way attributes are attached to objects is just so completely opaque: There is no place to describe what an attribute actually means...

Digging into their documentation, from version 2.0 onwards, one can actually give a name to physical entities. In #11592 I'm using this to read and write Material, Manifold, and Boundary ids. It's a bit of a hack, I admit, but it's the best I could come up with.

luca-heltai:
> This would be an amazing feature

@blaisb : here it is. :)

PR #11592, opened 2021-01-20
Added support for gmsh library - new write_msh()/read_msh() with manifolds

luca-heltai:
<s>No code changes yet.</s> This adds support for gmsh API, and for openlab API.

Instead of getting crazy with gmsh formats (see #11583), I went all in, and played a bit with their c++ API. It seems doable, and I'm experimenting a bit with their mesh formats.

Their new format allows also for STEP and IGES files to be incorporated in geometry files (!!!) through their opencascade geometry kernel. 

It would take our mesh + manifold support to a whole new level, allowing one to store all geometry and mesh information in a single file.


luca-heltai:
@peterrum , @blaisb , this turns out to be much easier than expected, using gmsh API. 

The included test shows how to output using gmsh itself, all grids we currently support.

Next test will output information on manifold and boundary ids.

luca-heltai:
The added test shows how the output of gmsh would look like with this library:

~~~
$MeshFormat
4.1 0 8
$EndMeshFormat
$PhysicalNames
4
1 1 "MaterialOrBoundaryID 0 - ManifoldId 0"
1 2 "MaterialOrBoundaryID -1 - ManifoldId 1"
2 3 "MaterialOrBoundaryID 0 - ManifoldId 1"
2 4 "MaterialOrBoundaryID 0 - ManifoldId -1"
$EndPhysicalNames
$Entities
0 2 2 0
...
~~~
As you see, I output strings where we identify boundary and manifold ids. The current reader would break with this format. The test can output any mesh type (also to vtk format!).

luca-heltai:
2nd test output with gmsh:

image

image

image

image

image

image

image

luca-heltai:
With the last commit, we have both `read_msh()` and `write_msh()` using GMSH API, with support for both boundary and manifold id. Closes #11583.

source/grid/grid_in.cc
```diff
@@ -2178,6 +2183,198 @@ GridIn<dim, spacedim>::read_msh(std::istream &in)
...
+        {
+          const auto &type       = gmsh_to_dealii_type.at(element_types[i]);
+          const auto  n_vertices = gmsh_to_dealii[type].size();
+          const auto &elements   = element_ids[i];
+          const auto &nodes      = element_nodes[i];
+          for (unsigned int j = 0; j < elements.size(); ++j)
+            {
+              if (entity_dim == dim)
+                {
+                  cells.emplace_back(n_vertices);
```
peterrum, line 2321: I completely forgot about this constructor :)

source/grid/grid_in.cc
```diff
@@ -2178,6 +2183,198 @@ GridIn<dim, spacedim>::read_msh(std::istream &in)
...
+                  for (unsigned int j = 0; j < elements.size(); ++j)
+                    boundary_ids_1d[nodes[j] - 1] = boundary_id;
+                }
+            }
+        }
+    }
+
+  Assert(subcelldata.check_consistency(dim), ExcInternalError());
+
+  tria->create_triangulation(vertices, cells, subcelldata);
```
peterrum, line 2364: :+1: Not the comparability version of the function is called!

source/grid/grid_out.cc
```diff
@@ -1470,6 +1474,211 @@ GridOut::write_xfig(const Triangulation<2> &tria,
...
+            id_pair_to_entity_tag[{boundary_or_material_id, manifold_id}];
+          add_element(element, entity_tag);
+          dim_entity_tag.insert({struct_dim, entity_tag});
+        }
+    };
+
+  // Loop recursively over all cells, faces, and possibly lines.
+  for (const auto &cell : tria.active_cell_iterators())
+    {
+      maybe_add_element(cell, cell->material_id());
```
peterrum, line 1607: Where is the recursion? Why aren't level-cells visited? Can gmsh handle hanging nodes?

peterrum (approved):
Looks very good :1st_place_medal: 

If I am confused with the gmsh format, I will simply output a deal.II triangulation and everything will be clear: I like that :)

source/grid/grid_out.cc
luca-heltai: Gmsh can handle hanging nodes in output, but not in input. 

In a previous version I had a recursion, but it was much more complicated. I replaced the recursive version with loops over cells, faces, and lines: the principle is that every element of every structural dimension of the triangulation is visited and output with its material id (if its a cell), boundary id (if it's a face/line) and manifold id.

peterrum (approved):

luca-heltai:
@tamiko , would you mind taking a look at the cmake part?

tamiko (approved):

PR #11593, opened 2021-01-20
Compile subdivided_hyper_rectangle_with_simplices ...

peterrum:
... even without DEAL_II_WITH_SIMPLEX_SUPPORT enabled and add an assert.

This was the only place we use to disable the compilation of simplex-related functions, passing the responsibility to the user, who want to build their user code with and without simplex support,  to also add ifdefs.

FYI @mschreter @JPmoep

### DIFF
--- a/source/simplex/grid_generator.cc
+++ b/source/simplex/grid_generator.cc
@@ -19,8 +19,6 @@
 
 DEAL_II_NAMESPACE_OPEN
 
-#ifdef DEAL_II_WITH_SIMPLEX_SUPPORT
-
 namespace GridGenerator
 {
   template <int dim, int spacedim>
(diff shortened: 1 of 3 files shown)


source/simplex/grid_generator.cc
```diff
@@ -32,7 +30,10 @@ namespace GridGenerator
     const Point<dim> &               p2,
     const bool                       colorize)
   {
-    AssertDimension(dim, spacedim);
+#ifdef DEAL_II_WITH_SIMPLEX_SUPPORT
+    Assert(false, ExcNeedsSimplexSupport())
```
drwells, line 34:
```suggestion
    Assert(false, ExcNeedsSimplexSupport());
```

drwells (approved):
Looks good, but could you add the semicolon?

drwells (approved):

bangerth (approved):

Issue #10474, opened 2020-06-06
DataOut::add_data_vector vector handling


kronbichler:
The current design of `DataOut::add_data_vector` seems to be a bit annoying for several reasons:
- The vectors passed to `add_data_vector` need to survive until "build_patches" - or is it the write_out functions? This makes writing modular code where the `add_data_vector` function is called in a different module than `build_patches()`/`write_...` cumbersome because one cannot use locally created vectors for output. This hit me more than once, even though I know the documentation.
- One needs to make sure that the vector has ghost values updated, forcing the user to write additional code even though we would know inside of the function what kind of data we need.

It would seem that this is the ideal place to also remove some of the vector templates by either equipping `add_data_vector` to copy itself to a `LinearAlgebra::read_write_vector`. Well, after the discussion at the workshop we might also only allow to pass in that kind of vectors in the first place, but that would not address the first point listed above. What do others think?

@nfehn @peterrum FYI

bangerth:
Yes, I think it would be a nice simplification to just copy the vector. (And to answer your question: The vector only needs to live to `build_patches()` where everything is converted into an internal representation. It doesn't have to live till `write_*()` time.)

peterrum:
I think we can close this issue?

kronbichler:
Let us close this issue, I think the main concept is now implemented with #11594 indeed.

PR #11594, opened 2021-01-20
DataOut: update ghost values

peterrum:
This PR proposes to update the ghost values of distributed vectors if they are not in ghosted state (and to retrieve the original state via calling `zero_out_ghost_values()`) during `DataOut::build_patches()`. This approach is very much influenced by the way how `MatrixFree::VectorDataExchange` handles vectors.

Let's see what the opinions are. While the update of the ghost values is not needed in all cases (DG, output of a subset of cells), we think this modifications would significantly ease the usability in the general case.

This patch was written jointly with @mschreter.

### DIFF
--- a/include/deal.II/numerics/data_out_dof_data.templates.h
+++ b/include/deal.II/numerics/data_out_dof_data.templates.h
@@ -28,6 +28,7 @@
 
 #include <deal.II/dofs/dof_accessor.h>
 #include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
 
 #include <deal.II/fe/fe_dgq.h>
 #include <deal.II/fe/fe_pyramid_p.h>
@@ -44,6 +45,8 @@
 #include <deal.II/hp/fe_values.h>
 #include <deal.II/hp/q_collection.h>
 
+#include <deal.II/lac/block_vector_base.h>
+#include <deal.II/lac/read_write_vector.h>
 #include <deal.II/lac/vector.h>
 
 #include <deal.II/numerics/data_out.h>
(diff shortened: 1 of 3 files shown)


bangerth:
I have some comments I could make on the patch, but before we go there: Why is this needed at all? What `DataOut` needs is locally relevant values, but these are also needed in a wide variety of other contexts for most programs. Why can't you just pass a corresponding vector to `DataOut`?

I'll note that I believe that the functionality doesn't actually work with PETSc and Trilinos vectors since they do not have the ability to update/zero ghost elements. I'm not too much in favor of making stuff work only for a specific vector type.

peterrum:
Our experience is that vectors should be in the general case not be ghosted, since it introduces additional computation/communication steps and you quickly loose performance (in matrixfree-context it might be in the range of double digits percentage). There are only a few cases were you actually need ghosted vectors like in the case of `DataOut`. In these cases, we need to write:
```cpp
vector.update_ghost_values();
date_out.build_patches(mapping); // do something with the vector
vector.zero_out_ghost_values();
```
This is the most simple setup but users already tend to make mistakes here: 1) forget to call `update_ghost_values()` (they will get an assert in debug mode - if they use debug mode) and 2) forget to call `vector.zero_out_ghost_values()` (and here you do not get any warnings).

In our application, we are dealing with multiple vectors that are attached to `DataOut` by multiple operators; calling `zero_out_ghost_values()` is quite complicated in this context. Possible, but you need to add an additional clean up logic in the user code.

Since we think that writing output files in parallel with `DataOut` is such a common task, it would be the best if this would be handled by the library and not the user code.

kronbichler:
I am late to comment on this PR. I would add that it definitely makes sense to let `DataOut` take care of the data exchange, rather than requesting the user to do so. However, I think it would be better if we create copies of the vectors, say into a `ReadWriteVector` or a `LA::d::Vector` with ghosts on all locally relevant dofs, inside the `DataOut` class. This way, we would benefit all vector types equally, and even somewhat simplify the access into the vectors later because they all go through the same vector type. One thing we might want to add then is a `Utilities::MPI::Partitioner` to `DoFHandler` so we don't have to make the expensive construction every time. (This would then also simplify `DoFTools::extract_locally_relevant_dofs` and others).

peterrum:
> I am late to comment on this PR. I would add that it definitely makes sense to let DataOut take care of the data exchange, rather than requesting the user to do so. However, I think it would be better if we create copies of the vectors, say into a ReadWriteVector or a LA::d::Vector with ghosts on all locally relevant dofs, inside the DataOut class. This way, we would benefit all vector types equally, and even somewhat simplify the access into the vectors later because they all go through the same vector type. One thing we might want to add then is a Utilities::MPI::Partitioner to DoFHandler so we don't have to make the expensive construction every time. (This would then also simplify DoFTools::extract_locally_relevant_dofs and others).

@kronbichler Sounds reasonable to me. I'll try to come up with a draft within the next days!

bangerth:
I could be ok with creating a copy of a vector right in `DataOut_DoFData::add_data_vector()`. This could, with some follow-up work, save a large number of template instantiations.

I do have some philosophical objections to this kind of approach, though. We have generally avoided doing these sorts of operations (importing ghost elements) behind the user's back. No other function in deal.II does this, as far as I know. For example, pretty much every function in `VectorTools` that needs to read from vector elements requires you to pass in a read-only ghost-element-containing vector. I think that that is right -- it is slightly more work, but it requires writing clean code in which the semantics of what kind of vector one has is clear. And it is ultimately more efficient: in all "real" codes, one doesn't just want to output vector, but there are other postprocessing steps as well: compute fluxes, compute averages, evaluate error estimators, etc. All of these need vectors with ghost elements. It makes no sense if all of these functions take completely distributed vectors, then do the import of ghost elements themselves, and then throw away the ghost elements again. This is not efficient -- let the user do it once in their code, and then pass the right vector in; that's marginally less inconvenient, but it's not difficult to do once at the beginning of postprocessing. Heck, step-40 shows how to do that.

I will also point out that it is reasonable to run some postprocessing steps in parallel on multiple threads, in particular if these steps require I/O and so have latencies. But that workflow will break if we need to do MPI communication for importation unless each of the places that imports data uses a separate MPI tag for their import -- which we neither do at this time, nor is it easy to change.

peterrum:
> So how do we move forward here?

I see your points. But I nevertheless think it would be useful in the form @kronbichler suggested. It is such a commonly used functionality so that I would advocate to make the life of the users a bit easier. Furthermore, I would suggest to store commonly used partitioners inside of `DoFHandler` so that they can be reused, potentially accelerating the setup a bit, since the consensus algorithms does not have to be run again. 

> We have generally avoided doing these sorts of operations (importing ghost elements) behind the user's back. No other function in deal.II does this, as far as I know.

We are doing similar things already at multiple places in the library. The `MatrixFree`-loops updates ghost values and compress automatically. The multigrid algorithms copy vectors into vectors with appropriate ghost regions needed by the transfer operators. PR #11882 will do it because no assumptions can be made regarding to the partitioners.

bangerth:
I get that in the matrix free algorithms, this may be necessary because you are both reading and writing into vectors and need to exchange information across processors. It's also true that someone who passes a vector to the matrix free subsystem will likely have the mindset that the vector is being worked on and that you can't expect to do something else on it at the same time -- in other words, that the vector *will* be modified.

But for `DataOut`, the mindset is different. The vector is what it is, there is no reason to believe that it will be changed and that consequently we can pass the vector to any number of concurrently running `DataOut` or other postprocessing functions. 

I just really don't like the originally proposed change just because it saves someone three lines. I'm ok with @kronbichler 's suggestion for making a copy and, if necessary, importing ghost elements on the copy.

peterrum:
A first new draft. @kronbichler @bangerth is this more to your liking?

peterrum:
Closes #10474.

peterrum:
The following failing tests are related to block vectors:
```
The following tests FAILED:
	 68 - simplex/step-20.debug (Failed)
	 70 - simplex/step-31.debug (Failed)
	 72 - simplex/step-55.mpirun=2.debug (Failed)
```
First, there is no option to import from a block vector to a `LinearAlgebra::distributed::Vector`; second the locally-owned indexset of the block vector might be not contiguous... which is sub-optimal for `LinearAlgebra::distributed::Vector`. Are there any ideas how to deal with these issues?

peterrum:
I would like to get this working during the workshop.

bangerth (changes requested):
Close. Just cosmetic stuff.

Does the documentation of DataOut ever say that we store a reference to the vector passed to add_data_vector? That would have to be updated as well.

peterrum:
@bangerth Thanks for the review! I have made the changes!

bangerth:
There's a failing test:
```
In file included from /jenkins/workspace/dealii_PR-11594/tests/all-headers/test_header.cc:17:0:
/jenkins/workspace/dealii_PR-11594/include/deal.II/numerics/data_out_dof_data.templates.h:795:41: error: ‘IsBlockVector’ was not declared in this scope
                 typename std::enable_if<IsBlockVector<VectorType>::value,
                                         ^
/jenkins/workspace/dealii_PR-11594/include/deal.II/numerics/data_out_dof_data.templates.h:795:65: error: template argument 1 is invalid
                 typename std::enable_if<IsBlockVector<VectorType>::value,
                                                                 ^
/jenkins/workspace/dealii_PR-11594/include/deal.II/numerics/data_out_dof_data.templates.h:795:66: error: expected identifier before ‘::’ token
                 typename std::enable_if<IsBlockVector<VectorType>::value,
                                                                  ^
/jenkins/workspace/dealii_PR-11594/include/deal.II/numerics/data_out_dof_data.templates.h:795:73: error: expected unqualified-id before ‘,’ token
                 typename std::enable_if<IsBlockVector<VectorType>::value,
                                                                         ^
```
You just need to add the appropriate `#include` to the `.templates.h` file.

bangerth (approved):
OK otherwise!

bangerth:
Nice! Now we just need to remove the vector template argument of these classes and only template over the scalar type.

peterrum:
> Nice! Now we just need to remove the vector template argument of these classes and only template over the scalar type. 

I have just started. I am wondering if we want to distinguish between double/float.

bangerth:
Since we store copies of the data, it's ok to standardize on `double` and `std::complex<double>`.

gassmoeller:
I see an error in some of ASPECT's tests that I suspect is caused by this PR:
```
An error occurred in line <844> of file <../include/deal.II/lac/read_write_vector.templates.h> in function
    void dealii::LinearAlgebra::ReadWriteVector<Number>::import(const dealii::TrilinosWrappers::MPI::Vector&, dealii::VectorOperation::values, const std::shared_ptr<const dealii::Utilities::MPI::CommunicationPatternBase>&) [with Number = double]
The violated condition was: 
    !trilinos_vec.has_ghost_elements()
Additional information: 
    Import() from TrilinosWrappers::MPI::Vector with ghost entries is not
    supported!
```

The tests all seem to be the ones that use periodic boundaries and MPI, and the problem happens during the postprocessing (most likely while building patches in DataOut). 
I am not sure what causes this. Is this expected and we need to do something different when preparing our postprocessing (e.g. ensure our vector do or do not have ghost entries), or is this unintended? 
If you say it is unintended I can try to track down the problem further and track down if it is actually this PR causing it (it must be a change of the last days), and where the call happens (dont have a backtrace right now). 
I just thought I'd ask first.

peterrum:
@gassmoeller Thanks for reporting this.

> I am not sure what causes this. Is this expected and we need to do something different when preparing our postprocessing (e.g. ensure our vector do or do not have ghost entries), or is this unintended?
If you say it is unintended I can try to track down the problem further and track down if it is actually this PR causing it (it must be a change of the last days), and where the call happens (dont have a backtrace right now).
I just thought I'd ask first.

This is not intended. Obviously, there is no test for this configuration. I am not sure why `ReadWriteVector` does not support  ghosted `TrilinosWrappers::MPI::Vector`. I'll investigate tomorrow!

gassmoeller:
Thanks! It is not super urgent, I suspect not many ASPECT users work with the latest deal.II, and it only affects a very specific combination of models, but if it happens in ASPECT it might happen elsewhere too. If you want any more information from me, let me know.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -890,22 +1041,31 @@ namespace internal
        * Pointer to the data vector. Note that ownership of the vector pointed
        * to remains with the caller of this class.
        */
-      const VectorType *vector;
+      LinearAlgebra::distributed::BlockVector<typename VectorType::value_type>
```
tjhei, line 1044: The comment above is out of date.

include/deal.II/numerics/data_out_dof_data.templates.h
```diff
@@ -1237,7 +1399,6 @@ namespace internal
     void
     DataEntry<dim, spacedim, VectorType>::clear()
     {
-      vector            = nullptr;
```
tjhei, line 1240: Shouldn't you reset this vector to free memory?

tjhei:
I assume that code and documentation of step-40 needs to be updated as well?

tjhei:
Does this PR also break the trilinos version of step-40?

peterrum:
> I assume that code and documentation of step-40 needs to be updated as well?

I had a quick look at step-40. I don't think the documentation needs to be updated. Could you point out the relevant lines. Thanks!

PR #11596, opened 2021-01-21
Fix ThreadLocalStorage assigment operators

### LINKED ISSUE #11587: Potential bug in ThreadLocalStorage
We are using `dealii::ThreadLocalStorage<T>` in our project and have had some issues with this class recently.

I think there might be a bug in `dealii::ThreadLocalStorage`. The copy assignment https://github.com/dealii/dealii/blob/master/include/deal.II/base/thread_local_storage.h#L330 and  and the move assignment https://github.com/dealii/dealii/blob/master/include/deal.II/base/thread_local_storage.h#L344 both do not copy/move the `exemplar` member, which is used for constructing new copies for other threads. Additionally, the copy and move constructors do copy/move `exemplar`. I would assume that constructor and assign should "do the same thing".

I am missing something or is this really a bug? Thanks for your help!

nmuch:
The copy and move assignment operators must also copy/move the `exemplar` in case the `data` is still empty.

Closes #11587

### DIFF
--- a/tests/multithreading/thread_local_storage_08.cc
+++ b/tests/multithreading/thread_local_storage_08.cc
@@ -1,0 +1,79 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2008 - 2018 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// test if ThreadLocalStorage::operator= (ThreadLocalStorage<T> &&t) moves the
+// shared ptr to the exemplar
+
+#include <deal.II/base/thread_local_storage.h>
+#include <deal.II/base/thread_management.h>
+
+#include "../tests.h"
+
+
+struct X
+{
+  X()
+  {
+    deallog << "Creating" << std::endl;
+  };
+  X(const X &)
+  {
+    deallog << "Copying" << std::endl;
+  };
+  int i;
+};
+
+Threads::ThreadLocalStorage<X> tls_data;
+
+void
+execute(int i)
+{
+  tls_data.get().i = i;
+}
+
+
+void
+test()
+{
+  // create a thread local storage object
+  X                              exemplar;
+  Threads::ThreadLocalStorage<X> temp(exemplar);
+
+  // move assign thread local storage object
+  tls_data = std::move(temp);
+
+  // create 5 threads and wait for their
+  // return. the OS will create 5 individual
+  // thread ids, which means that we will
+  // create 5 individual thread specific
+  // storage locations
+  Threads::ThreadGroup<> tg;
+  for (unsigned int i = 10; i < 15; ++i)
+    tg += Threads::new_thread(execute, i);
+
+  // now make sure the threads all finish
+  tg.join_all();
+}
+
+
+
+int
+main()
+{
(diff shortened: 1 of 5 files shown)


bangerth (commented):
Good work! Thanks for taking this on!

Do you think you could write a short test that checks that this works now? Take a look at some of the `tests/multithreading/thread_local_storage_*` tests.

nmuch:
I added tests that check if the `exemplar` is copied/moved by the respective assignment operators of `ThreadLocalStorage`. Without the exemplar the threads would create new instances of `X` instead of copying them and the tests fail.

peterrum (approved):

tests/multithreading/thread_local_storage_08.cc
```diff
@@ -0,0 +1,79 @@
...
+struct X
+{
+  X()
+  {
+    deallog << "Creating" << std::endl;
+  };
+  X(const X &)
+  {
+    deallog << "Copying" << std::endl;
+  };
```
masterleinad, line 35:
```suggestion
  X(X &&)
  {
    deallog << "Moving" << std::endl;
  };
```

tests/multithreading/thread_local_storage_08.cc
```diff
@@ -0,0 +1,79 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// test if ThreadLocalStorage::operator= (const ThreadLocalStorage<T> &&t) moves
```
masterleinad, line 17:
```suggestion
// test if ThreadLocalStorage::operator= (ThreadLocalStorage<T> &&t) moves
```

tests/multithreading/thread_local_storage_08.output
```diff
@@ -0,0 +1,8 @@
+
+DEAL::Creating
+DEAL::Copying
+DEAL::Copying
+DEAL::Copying
+DEAL::Copying
+DEAL::Copying
+DEAL::Copying
```
masterleinad, line 8:
```suggestion
DEAL::Moving
DEAL::Moving
DEAL::Moving
DEAL::Moving
DEAL::Moving
DEAL::Moving
```

masterleinad (changes requested):
I think you wanted to check the move assignemt operator in the last test.

tests/multithreading/thread_local_storage_08.cc
nmuch: I want to test the move assignment of `ThreadLocalStorage` which does not move the `exemplar`. It moves a shared pointer to the `exemplar` which does not move the instance of `X`.

tests/multithreading/thread_local_storage_08.output
nmuch: The `exemplar` is used to copy its value into the thread-local `data` if `data` is still empty on a thread. If the `exemplar` is empty, new instances of `X` are constructed instead, using the default ctor.

Without the bugfix in this PR the `exemplar` is not copied/moved and the threads in this test would "create" new instances of `X` instead of copy-constructing them.

tests/multithreading/thread_local_storage_08.cc
nmuch: Thank you for the correction. I changed the comment so it says that `ThreadLocalStorage::operator=(ThreadLocalStorage<T> &&t=` moves the shared pointer to the exemplar and not the exemplar directly.

tests/multithreading/thread_local_storage_08.cc
masterleinad: Yes, I didn't look carefully enough.

masterleinad (approved):

PR #11595, opened 2021-01-21
Make the special ReferenceCell objects objects, not enums.

bangerth:
This is the next step in the evolution of `ReferenceCell` started in #11300: It converts the "special" objects such as `Type::Quad` to objects, rather than integers. My goal is to completely remove the fact that the `ReferenceCell::Type` objects correspond to integers: That should be completely immaterial, and so conversions from `ReferenceCell::Type` to some integer type or back should disappear to the degree possible. This patch gets us 90% there.

/rebuild

PR #11597, opened 2021-01-21
Remove now unnecessary things in ReferenceCell::Type. Make some things constexpr.

bangerth:
Follow-up to #11595. Nothing much to it, just continue working in the same direction as started in #11544.

/rebuild

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -64,12 +64,12 @@ namespace ReferenceCell
     /**
      * Default constructor. Initialize this object as an invalid object.
      */
-    Type();
+    constexpr Type();
 
     /**
      * Constructor.
      */
-    Type(const std::uint8_t kind);
+    constexpr Type(const std::uint8_t kind);
 
     /**
      * Return the dimension of the reference cell represented by the current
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
Looks reasonable. It would, of course, be better if you were already using some of the `constexpr` functions either in the library or in unit tests.

peterrum (approved):

bangerth:
The commit with `constexpr` primarily comes from the fact that I wanted to make these objects into `constexpr` objects:
```
  class Type
  {
  public:
    static const Type Vertex;
    static const Type Line;
    static const Type Tri;
    static const Type Quad;
    static const Type Tet;
    static const Type Pyramid;
    static const Type Wedge;
    static const Type Hex;
    static const Type Invalid;
```
But that doesn't work (at least not without using a different approach that I want to use later): when you make these objects `constexpr`, you have to provide an initializer, but you can't do that because the surrounding type is, at that point, still incomplete. I find that quite annoying -- it means that you can't do `constexpr` singletons that are `static` member functions -- quite a nuisance. The current patch is really just a way station on my path, but I thought it doesn't hurt to mark these things as `constexpr` because I already had that ready. The fact that we don't use that anywhere (and I wouldn't even know where that would make sense, to be honest) doesn't mean that we can't mark these functions as suggested.

PR #11600, opened 2021-01-21
Added test for FEQxFENothing and FENothingxFEQ constraints.

marcfehling:
Another test that joins #11448 and #11549 to figure out what's wrong with `FENothing`.

It looks like constraints will be set correctly in this test in 2D and 3D. If both `FENothing` elements do not dominate, no constraints will be imposed which is correct (not explicitely tested here, but can be easily constructed by changing line 60). If one of them dominates, the correct constraints will be set (see output).

In 1D however, no constraints will be imposed at all, which is wrong.

I don't know why `VectorTools::interpolate` sets some of the values to `1.5` although they should be `1.0` to my understanding.

### DIFF
--- a/tests/hp/fe_nothing_25.output
+++ b/tests/hp/fe_nothing_25.output
@@ -1,0 +1,22 @@
+
+DEAL:1d::constraints:
+DEAL:1d::dof values:
+DEAL:1d:: cell 0: 1 1
+DEAL:1d:: cell 1: 1 1
+DEAL:1d::OK
+DEAL:2d::constraints:
+    1 = 0
+    3 = 0
+DEAL:2d::dof values:
+DEAL:2d:: cell 0: 1.5 0 1.5 0
+DEAL:2d:: cell 1: 1 1 1 1
+DEAL:2d::OK
+DEAL:3d::constraints:
+    1 = 0
+    3 = 0
+    5 = 0
+    7 = 0
+DEAL:3d::dof values:
+DEAL:3d:: cell 0: 1.5 0 1.5 0 1.5 0 1.5 0
+DEAL:3d:: cell 1: 1 1 1 1 1 1 1 1
+DEAL:3d::OK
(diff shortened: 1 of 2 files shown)


tests/hp/fe_nothing_25.output
```diff
@@ -0,0 +1,22 @@
+
+DEAL:1d::constraints:
+DEAL:1d::dof values:
```
bangerth, line 3: So this isn't correct, actually. You'd expect that one of the DoFs on one of the cells is actually constrained.

tests/hp/fe_nothing_25.output
```diff
@@ -0,0 +1,22 @@
...
+DEAL:1d:: cell 0: 1 1
+DEAL:1d:: cell 1: 1 1
+DEAL:1d::OK
+DEAL:2d::constraints:
+    1 = 0
+    3 = 0
+DEAL:2d::dof values:
```
bangerth, line 10: Fascinating: It works in 2d!

tests/hp/fe_nothing_25.output
```diff
@@ -0,0 +1,22 @@
...
+DEAL:2d:: cell 0: 1.5 0 1.5 0
+DEAL:2d:: cell 1: 1 1 1 1
```
bangerth, line 12: But what's going on with the `1.5` values?

tests/hp/fe_nothing_25.output
```diff
@@ -0,0 +1,22 @@
...
+DEAL:2d::OK
+DEAL:3d::constraints:
+    1 = 0
+    3 = 0
+    5 = 0
+    7 = 0
+DEAL:3d::dof values:
+DEAL:3d:: cell 0: 1.5 0 1.5 0 1.5 0 1.5 0
```
bangerth, line 20: same here

bangerth (commented):
I'm for merging as-is, even though we know that the results are wrong. But I'm thoroughly confused about what the tests output...

marcfehling:
I see you are similarly confused as I am :-) At least `FE_Nothing` dominates something as ordered.

I am about the refactor our `FE_Nothing dominates`-tests and will merge this PR with #11448.

PR #11603, opened 2021-01-22
Hide the ability to create ReferenceCell objects from users.

bangerth:
Another follow-up to #11544.

The idea here is that users should not know, and should not be able to affect, what the numeric value is that corresponds to certain reference cell objects. For example, it should not matter at all that `Quad` has value `3` internally. In other words, what I want to achieve is that the class acts like an `enum` whose elements are not explicitly assigned values and, more importantly, for which it is not possible to create any object other than the predefined reference cell types.

The way to generally do this is make the constructor of the class `private` so that it is not possible to create these objects with specific numeric values -- the only way you're ever going to get an object of type `ReferenceCell::Type` is to use one of the pre-defined ones.

That leaves the issue of how the predefined objects are actually created. For singleton classes there is a standard way to do this but that doesn't work here because we don't want to use an `instance()` function as is usually done. Rather, I use a function that's hidden in an `internal` namespace and that is marked as `friend`. That achieves the same goal: For all intents and purposes, users will not be able to create these kinds of objects of their own, and will (as is my intent) have to use the pre-defined ones. The fact that I didn't have to change any code in the rest of the library validates this kind of design.

/rebuild

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -42,6 +42,29 @@ class ScalarPolynomialsBase;
  */
 namespace ReferenceCell
 {
+  class Type;
+
+  namespace internal
+  {
+    /**
+     * A helper function to create a ReferenceCell::Type object from an
+     * integer. ReferenceCell::Type objects are "singletons" (actually,
+     * "multitons" -- there are multiple, but they are only a handful and
+     * these are all that can be used). What is then necessary is to
+     * have a way to create these with their internal id to distinguish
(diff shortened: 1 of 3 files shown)


bangerth:
Also adds the range check @drwells had asked for in https://github.com/dealii/dealii/pull/11595/files/50a95d15ed199b8f41c66e08256d4ffcf53bb3d7#diff-b5db3426cd2f178f860362fb2241a1fb591f012cdf71e7b68665363ef64cfc5d

Rombur (approved):

PR #11607, opened 2021-01-22
Move a function that is not performance-critical to the .cc file.

bangerth:
In the same direction as started in #11544.

/rebuild

### DIFF
--- a/source/grid/reference_cell.cc
+++ b/source/grid/reference_cell.cc
@@ -63,9 +63,6 @@ namespace ReferenceCell
 
 
 
-  /**
-   * Convert the given reference cell type to a string.
-   */
   std::string
   Type::to_string() const
   {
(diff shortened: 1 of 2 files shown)


source/grid/reference_cell.cc
```diff
@@ -62,6 +62,39 @@ namespace ReferenceCell
     internal::make_reference_cell_from_int(static_cast<std::uint8_t>(-1));
 
 
+
+  /**
+   * Convert the given reference cell type to a string.
+   */
```
drwells, line 68:
```suggestion
```
Could we get rid of the documentation? It looks out of place in a source file.

drwells (changes requested):
One minor thing.

peterrum (approved):

bangerth:
@drwells -- yes, good idea! So done.

drwells (approved):

PR #11611, opened 2021-01-23
Improve an error message.

bangerth:
We encourage people to play with the dimension in step-1/2/3, but step-1 also uses `GridOut::write_svg()`. Make sure that people get good error messages if they try that.

/rebuild


bangerth:
Inspired by a forum post today.

drwells (approved):

PR #11613, opened 2021-01-23
Fix a variety of small simplex things.

drwells:
1. Our convention is to still mark overridden methods as virtual.
2. convert_generalized_support_point_values_to_dof_values should be public.
3. Fix simplex reference tria generation in codimension 1.
4. Remove some hard-coded numbers.
5. Fix a quadrature weight.


peterrum (approved):
Looks good! Thanks. My guess is that at least one test will fail. If not, the quadrature rules with 1 points are not tested^^

drwells:
I did some convergence testing for #11505 - I might be the only person to try and use piecewise constants for simplices so far.

peterrum:
This line should be updated:

https://github.com/dealii/dealii/blob/4fbb5374f0dffa75722b49040d4951c4123d4db2/tests/simplex/quadrature_lib_01.output#L8

drwells:
I think I fixed it.

source/grid/reference_cell.cc
```diff
@@ -75,8 +75,19 @@ namespace ReferenceCell
       }
     else if (reference_cell == Type::Tri)
       {
-        static const std::vector<Point<spacedim>> vertices = {
-          {{0.0, 0.0}, {1.0, 0.0}, {0.0, 1.0}}};
+        std::vector<Point<spacedim>> vertices;
```
luca-heltai, line 78: I know it's already merged, but wouldn't this be easier with 
```C++
std::vector<Point<spacedim>> vertices(3);
vertices[1][0] = 1.0;
vertices[2][1] = 1.0;
```
?

bangerth: Even simpler: #11617.

PR #11614, opened 2021-01-23
Permit conversion of triangulations with multiple levels to simplices.

drwells:
I thought this would be straightforward, but trying to write this patch exposed a weird bug where it looks like `Triangulation<2, 3.::active_cell_iterators()` returns the entire set of iterators, not just the active ones.

### DIFF
--- a/include/deal.II/grid/grid_generator.h
+++ b/include/deal.II/grid/grid_generator.h
@@ -145,10 +145,8 @@ namespace GridGenerator
    * @param tria The triangulation to create. It needs to be empty upon
    * calling this function.
    *
-   * @param repetitions A vector of @p dim positive values denoting the number
-   * of cells to generate in that direction. For example, the first component of
-   * the vector specifies the number of cells in the x-direction, the second
-   * component specifies that for the y-direction for dim=2.
+   * @param repetitions The number of cells to place in each coordinate
+   * direction.
    *
    * @param left Lower bound for the interval used to create the hyper cube.
    *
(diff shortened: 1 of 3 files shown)


peterrum (commented):
I am not sure about this. Once could flatten the original triangulation and then proceed with the function as usual. I don't think there is an easy way to make this work for locally-refined meshes.

And the bug should be fixed ;)

drwells:
> I don't think there is an easy way to make this work for locally-refined meshes.

I don't think so either, but we should make things work with all the active cells on the finest level.

bangerth (commented):
With the current version of the patch, you're creating a triangle/tet mesh out of the active cells of the existing quad/hex mesh. But if you have hanging nodes, then you will end up with a non-conforming tri/tet mesh in which you have internal boundaries. That's not what you want. It's the same problem as reading non-conforming locally refined meshes from file: We love the connection between cells. 

I think the way the function should work on meshes with more than one level is by using hierarchical refinement of a coarse mesh converted to simplices. This way you also recreate the multilevel structure.

drwells:
I agree with your position and in fact, that's what I tried to do - I added an assertion at circa line 7100 to check that there are only active cells on the finest mesh level.

drwells:
> I think the way the function should work on meshes with more than one level is by using hierarchical refinement of a coarse mesh converted to simplices. This way you also recreate the multilevel structure.

I'd like to do this too but we don't have global or local refinement available for tet meshes yet.

peterrum:
> I'd like to do this too but we don't have global or local refinement available for tet meshes yet.

Not on master ;) But anyway, if it works for 2D it will also work in 3D (some adjustments for boundary-ids, ...).

bangerth:
If we can't do this on master yet, then I'd favor holding off with this for now. We'd just end up with a function that doesn't do what we really want, and we'd have to think about backward compatibility once we can do local refinement. That doesn't mean that the function as-is can't be put into a test or application code, of course.

drwells:
There is a lot more that needs to be done for this to work - lets just close this PR for now.

Issue #11615, opened 2021-01-24
minimum cmake version?


nfehn:
It seems as if PR  #11384 breaks compatibility with e.g. cmake version 2.8. 

Discuss whether this is intended and how to fix this.

tamiko:
Interesting. I originally tested down to CMake-3.1.

I now tested with CMake-3.0, which seems to be the first failing version.
In summary, CMake versions older than 3.1 seem to not support the `\[...\]` regular expression part.

CMake-3.1 was released early 2015. We currently require as minimum compiler version gcc-5.1 (released April 2015), or clang-4.0.0 (which was released July 2017).

So I suggest we simply increase the minimum required version to 3.1. This is long overdue.

nfehn:
Sounds reasonable

bangerth:
@tamiko What do the typical ~2016 distros ship with? I'm having second thoughts whether requiring 3.1 was the right choice or whether we could have required something higher already. I don't know whether that makes much of a difference, but if these distros all already have, say, 3.5, then maybe it would have been good to actually require that.

PR #11616, opened 2021-01-24
CMake: increase minimum required version to 3.1.0

tamiko:
Closes: https://github.com/dealii/dealii/issues/11615


CMakeLists.txt
```diff
@@ -33,17 +33,14 @@
...
 #
 # We support all policy changes up to version 3.1.0. Thus, explicitly set
 # all policies CMP0001 - CMP0054 to new for version 3.1 (and later) to
 # avoid some unnecessary warnings.
 #
-IF( "${CMAKE_VERSION}" VERSION_EQUAL "3.1" OR
-    "${CMAKE_VERSION}" VERSION_GREATER "3.1" )
-  CMAKE_POLICY(VERSION 3.1.0)
-ENDIF()
+CMAKE_POLICY(VERSION 3.1.0)
```
tamiko, line 43: I think we discussed raising to 3.1 in the near future back when we had to introduce this `CMAKE_POLICY` switch.

tamiko:
I would like to get some sign-offs by my fellow developers on this one. Let's wait for some mor than the minimum required approvals.

For reference: 

CMake-3.1 was released early 2015. We currently require as minimum compiler version gcc-5.1 (released April 2015), or clang-4.0.0 (which was released July 2017).

kronbichler (approved):

tjhei:
Is this one location the only problem for cmake 2.8?
I don't mind increasing the minimum to 3.1 if 2.8 gives us headaches and extra work. Most current OSes have moved to 3.x, I would think. Ubuntu 16.04 has cmake 3.5 for example.  Otherwise, it is easy enough to install cmake yourself, even on a cluster.

bangerth (approved):

bangerth (commented):
Maybe also add a changelog entry?

Rombur (approved):

peterrum (approved):

drwells (approved):

tamiko:
@bangerth Added a changelog entry!

@tjhei Yes, we could have written a simple workaround for the regex.

But I think it's overdue that we raise the version requirement :smile:

masterleinad (approved):

tamiko:
6 approvals, I'd say that's a quorum of some sorts :-)

masterleinad:
Would you mind squashing your commits a bit?

tamiko:
@masterleinad Done.

PR #11618, opened 2021-01-25
Try to fix docker images.

luca-heltai:
Try to fix this:

https://github.com/dealii/dealii/runs/1746366699?check_suite_focus=true

I have no idea if it builds. I was not able to configure with python bindings on my system. It looks like the removal of CellKinds had some leftovers #11595

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -12,7 +12,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Build and push Docker image of master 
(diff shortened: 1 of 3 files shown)


luca-heltai:
@peterrum this should fix our docker build problems. Nothing will be compiled until it is merged to master, so let's see if it works..

peterrum (commented):
Let's give it a shot!

peterrum (approved):

PR #11623, opened 2021-01-25
Work on FEPointEvaluation::integrate

peterrum:
The slow path should already work. Do we want already the fast path getting working in this PR. For me it would more important to get the interface finalized: 1) integrate and 2) input arguments to functions evaluate/integrate.


include/deal.II/matrix_free/tensor_product_kernels.h
```diff
@@ -2518,6 +2518,32 @@ namespace internal
   }
 
 
+
+  /**
+   * TODO
+   */
+  template <int dim, typename Number, typename Number2>
+  inline void
+  integrate_tensor_product_value_and_gradient(
```
peterrum, line 2527: @kronbichler This is still missing. Any recommendations how to tackle this function?

include/deal.II/fe/fe_point_evaluation.h
```diff
@@ -530,6 +672,163 @@ FEPointEvaluation<n_components, dim, spacedim>::evaluate(
...
+
+      // let mapping compute the transformation
+      if (gradients.size() > 0)
+        {
+          Assert(mapping_q_generic != nullptr, ExcInternalError());
+
+          // cast constness away (input is overriden)
+          ArrayView<gradient_type> gradients_non_const(
+            const_cast<gradient_type *>(gradients.data()), gradients.size());
+          mapping_q_generic->transform_variable(
```
peterrum, line 703: I think we need here the transposed of this functions, don't we?

kronbichler: Yes, we need to apply the transpose here.

include/deal.II/matrix_free/tensor_product_kernels.h
kronbichler: It should be the transpose of the other function. So you first evaluate the polynomial basis in all points as above, and then switch the left hand side and right hand sides of the assignments. Here is the example of what I believe is needed in 2D (without renumbering and without cleanup code in case the derivative or value is empty; 1D should 3D be analogous):
```c++
  for (unsigned int i1 = 0, i = 0; i1 < (dim > 1 ? n_shapes : 1); ++i1)
    {
      Number test_value_y = value * shapes[2 * n_shapes + 2 * i1] +
                            gradient[1] * shapes[2 * n_shapes + 2 * i1 + 1];
      Number test_grad_y = gradient[1] * shapes[2 * n_shapes + 2 * i1];
      for (unsigned int i0 = 0; i0 < n_shapes; ++i0, ++i)
        values[i] = shapes[2 * i0] * test_value_y + shapes[2 * i0 + 1] * test_grad_y;
  }
```

kronbichler (commented):
I would be fine to proceed here with the fast path postpone and rather augment this by #11630 first. But I don't see why you add it in the tests as you do not print anything of `integrate`. Do you still want to add that?

peterrum:
@kronbichler I am in progress to update this branch based on the changes of the other branch. I will also output something!

peterrum:
@kronbichler I thinks this PR contains everything what we have discussed: `FEPointEvaluation::integrate()`, `FEPointEvaluation::submit_value()`, and `FEPointEvaluation::submit_gradient()`. For now, I see the interface as finalized.

kronbichler (approved):

peterrum:
@kronbichler In the future I would like to discuss how we can combine `FEPointEvaluation` and `FEEvaluation`. For now, I really appreciate the `read_dof_values()` and `distribute_dof_values()` of `FEEvaluation`. Maybe we can find a nice use case .

kronbichler (approved):

include/deal.II/fe/fe_point_evaluation.h
```diff
@@ -272,14 +407,36 @@ class FEPointEvaluation
...
   void
   evaluate(const typename Triangulation<dim, spacedim>::cell_iterator &cell,
            const ArrayView<const Point<dim>> &     unit_points,
            const ArrayView<const double> &         solution_values,
-           const EvaluationFlags::EvaluationFlags &flags);
+           const EvaluationFlags::EvaluationFlags &evaluation_flags);
+
+  /**
+   * This is one of the main functions in this class and the one that does the
+   * heavy lifting.
```
bangerth, line 421: Can you explain what the function actually *does*? :-)

include/deal.II/fe/fe_point_evaluation.h
```diff
@@ -48,14 +48,23 @@ namespace internal
...
         result[component] = vector_entry;
       }
 
+      static void
+      write_value(double &           vector_entry,
+                  const unsigned int component,
+                  const value_type & result)
+      {
+        AssertIndexRange(component, n_components);
+        vector_entry = result[component];
```
bangerth, line 65: I find the names of these functions confusing. The `read_value()` function actually *writes* into `results` argument, whereas the `write_value()` *reads* from the last argument and returns it through the first.

Why not use `set_value()` instead of `read_value()`, and `get_value()` for `write_value()` (possibly returning via a regular `return` rather than through a reference argument)?

bangerth (commented):
Late to the game, but I have two questions.

PR #11624, opened 2021-01-25
Nicely format exception texts.

bangerth:
We have quite a lot of exceptions now that have very long text descriptions. These are often hard to read because the text is not written in any formatted way. But that's easy to work around -- we even have a function already that breaks text as appropriate.

With this patch, we get this kind of output:
```
An error occurred in line <3150> of file </home/bangerth/p/deal.II/1/dealii/source/dofs/dof_handler.cc> in function
    void dealii::DoFHandler<dim, spacedim>::create_active_fe_table() [with int dim = 2; int spacedim = 2]
The violated condition was: 
    this->hp_cell_active_fe_indices[level].size() == this->tria->n_raw_cells(level) && this->hp_cell_future_fe_indices[level].size() == this->tria->n_raw_cells(level)
Additional information: 
    This exception -- which is used in many places in the library --
    usually indicates that some condition which the author of the code
    thought must be satisfied at a certain point in an algorithm, is not
    fulfilled. An example would be that the first part of an algorithm
    sorts elements of an array in ascending order, and a second part of
    the algorithm later encounters an element that is not larger than the
    previous one.
    
    There is usually not very much you can do if you encounter such an
    exception since it indicates an error in deal.II, not in your own
    program. Try to come up with the smallest possible program that still
    demonstrates the error and contact the deal.II mailing lists with it
    to obtain help.

Stacktrace:
-----------
#0  /home/bangerth/p/deal.II/1/install/lib/libdeal_II.g.so.9.3.0-pre: dealii::DoFHandler<2, 2>::create_active_fe_table()
...
```

/rebuild

### DIFF
--- a/doc/news/changes/minor/20210125WolfgangBangerth
+++ b/doc/news/changes/minor/20210125WolfgangBangerth
@@ -1,0 +1,4 @@
+Improved: Exception texts are now formatted and broken to fixed-length
+lines, rather than flowing to arbitrary lengths.
+<br>
+(Wolfgang Bangerth, 2021/01/25)
--- a/source/base/exceptions.cc
+++ b/source/base/exceptions.cc
@@ -321,10 +321,25 @@ ExceptionBase::generate_message() const
                 << "--------------------------------------------------------"
                 << std::endl;
 
-      // print out general data
+      // Print out general data
       print_exc_data(converter);
-      // print out exception specific data
-      print_info(converter);
+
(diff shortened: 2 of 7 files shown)


drwells (approved):
Nice!

I approve, but I suspect this might break some tests that print exception messages.

bangerth:
Turns out it only broke 5 tests :-) All fixed now.

PR #11625, opened 2021-01-25
simplex: implement 'interface_constraints' for FE_Poly.

marcfehling:
Implements the _oldstyle_ constraints for `FE_Poly`.

Not absolutely necessary once #11590 is merged since it provides _hp_ constraints, but it won't hurt to have the other variant as well I guess. Worked on this by accident as I forgot to rebase on the #11590 branch...


source/simplex/fe_lib.cc
```diff
@@ -151,6 +151,78 @@ namespace Simplex
...
+
+      // Now construct relation between destination (child) and source (mother)
+      // dofs.
+
+      const unsigned int n_dofs_constrained = constraint_points.size();
+      unsigned int       n_dofs_per_face    = degree + 1;
+
+      FullMatrix<double> interface_constraints;
+      {
+        TableIndices<2> interface_constraints_size = {n_dofs_constrained,
```
bangerth, line 201:
```suggestion
        const TableIndices<2> interface_constraints_size = {n_dofs_constrained,
```

source/simplex/fe_lib.cc
```diff
@@ -151,6 +151,78 @@ namespace Simplex
...
+                                                      n_dofs_per_face};
+        interface_constraints.TableBase<2, double>::reinit(
+          interface_constraints_size);
+      }
```
bangerth, line 205: What does this block do? Isn't it just setting the size of `interface_constraints`, and that could already be done in the constructor call above the `{...}` block?

peterrum (approved):

bangerth (approved):

PR #11637, opened 2021-01-27
simplex: Adapt CellAccessor::neighbor_child_on_subface for triangles.

marcfehling:
Required by `DoFTools::make_hanging_node_constraints()` in the context of _hp_-adaptation.

@peterrum -- FYI


source/grid/tria_accessor.cc
```diff
@@ -2845,368 +2845,432 @@ CellAccessor<dim, spacedim>::neighbor_child_on_subface(
...
+                        [GeometryInfo<2>::max_children_per_face] = {{0, 1},
+                                                                    {1, 2},
+                                                                    {2, 0}};
+
+              // determine indices for this cell's subface from the perspective
+              // of the neighboring cell
+              const unsigned int neighbor_face =
+                this->neighbor_of_neighbor(face);
+              // face orientation flips on neighboring cells
+              const unsigned int neighbor_subface = 1 - subface;
```
marcfehling, line 2870: There might be more general approaches to this. Open for suggestions!

peterrum: This is not guaranteed to work. You can construct neighboring cells so that the orientation is right for both sides. I guess you need to check if one of the sides are "misorientated" and if yes make the "flip". I think you should be able to use the `line_orientation()` function.

Side note: the term "flip" only exists in 3D ;)

source/grid/tria_accessor.cc
```diff
@@ -2845,368 +2845,432 @@ CellAccessor<dim, spacedim>::neighbor_child_on_subface(
...
+
+              // indices of child cells that are adjacent to a certain face of
+              // the mother cell
+              // basically GeometryInfo<2>::child_cell_on_face() for triangles
+              constexpr unsigned int faces_per_cell = 3;
+              constexpr unsigned int
+                subcells[faces_per_cell]
+                        [GeometryInfo<2>::max_children_per_face] = {{0, 1},
+                                                                    {1, 2},
+                                                                    {2, 0}};
```
peterrum, line 2863: This information should be move to `ReferenceCell::internal::Info` (the new `GeometryInfo`).

source/grid/tria_accessor.cc
marcfehling: `line_orientation()` seems to be only implemented for quads/hexes right now, considering its use of `GeometryInfo`. 

The function refers to a field `face_orientations` in `TriaLevel`. Does that field already store the correct information for simplex meshes?

marcfehling: I see you already set it for triangles here. I'll give it a try
https://github.com/dealii/dealii/blob/87eb8ae23de620632fad3e77c4adcc799024baed/source/grid/tria.cc#L4473-L4483

source/grid/tria_accessor.cc
```diff
@@ -2845,368 +2845,433 @@ CellAccessor<dim, spacedim>::neighbor_child_on_subface(
...
+          if (this->reference_cell_type() == ReferenceCell::Type::Tri)
             {
-              Assert((GeometryInfo<dim>::face_refinement_case(
-                        sub_neighbor->refinement_case(), neighbor_neighbor) ==
-                      RefinementCase<dim>::no_refinement),
-                     ExcInternalError());
-              sub_neighbor =
-                sub_neighbor->child(GeometryInfo<dim>::child_cell_on_face(
-                  sub_neighbor->refinement_case(), neighbor_neighbor, 0));
+              const auto &neighbor_cell = this->neighbor(face);
```
bangerth, line 2850: The function returns by value, so let's take that value rather than keep a reference to a temporary.
```suggestion
              const auto neighbor_cell = this->neighbor(face);
```

bangerth (approved):

bangerth:
Please make the one change and then let's merge.

/rebuild

PR #11638, opened 2021-01-27
simplex: step-6

marcfehling:
~Blocked by #11590 and #11637.~ (EDIT: Have been merged.)

Port of `step-6` for triangles.

@peterrum -- FYI

### DIFF
--- a/tests/simplex/step-06.cc
+++ b/tests/simplex/step-06.cc
@@ -1,0 +1,394 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2021 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Step-06 on a simplex mesh.
+//
+// The following changes had to be made due to incompatibilities:
+//  - Manifolds are disabled
+//    All manifold ids from the triangulation object generated with
+//    GridGenerator::hyper_ball() had to be removed
+//  - Error estimation has to be skipped
+//    KellyErrorEstimator requires QProjector::project_to_all_subfaces()
+//    to work with triangles
+//  - GridOut::write_gnuplot() has been skipped
+
+
+#define USE_SIMPLEX
+// #define OUTPUT_RESULTS
+
+
+#ifdef USE_SIMPLEX
+#  include <deal.II/base/types.h>
+
+#  include <deal.II/simplex/fe_lib.h>
+#  include <deal.II/simplex/quadrature_lib.h>
+#else
+#  include <deal.II/base/quadrature_lib.h>
+
+#  include <deal.II/fe/fe_q.h>
+#endif
+
+
+#ifdef OUTPUT_RESULTS
+#  include <deal.II/numerics/data_out.h>
+
+#  include <fstream>
+#endif
+
+
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/dofs/dof_tools.h>
+
+#include <deal.II/fe/fe_values.h>
+#include <deal.II/fe/mapping_fe.h>
+
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/grid_refinement.h>
+#include <deal.II/grid/tria.h>
+
+#include <deal.II/lac/affine_constraints.h>
+#include <deal.II/lac/dynamic_sparsity_pattern.h>
+#include <deal.II/lac/full_matrix.h>
+#include <deal.II/lac/precondition.h>
+#include <deal.II/lac/solver_cg.h>
+#include <deal.II/lac/sparse_matrix.h>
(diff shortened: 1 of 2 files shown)


marcfehling:
step-6

tests/simplex/step-06.cc
```diff
@@ -0,0 +1,397 @@
...
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Step-06 on a simplex mesh.
+//
+// The following changes had to be made due to incompatibilities:
+//  - Manifolds are disabled
+//    All manifold ids from the triangulation object generated with
+//    GridGenerator::hyper_ball() had to be removed
```
peterrum, line 21: Why? Could you do it like this: https://github.com/dealii/dealii/blob/b59c748f39b00cf111ab9eec96c115d8aa9792a2/tests/simplex/step-01.cc#L65-L67

peterrum (commented):
Excellent!

tests/simplex/step-06.cc
```diff
@@ -0,0 +1,397 @@
...
+//  - Error estimation has to be skipped
+//    KellyErrorEstimator requires QProjector::project_to_all_subfaces()
```
luca-heltai, line 23: but below it looks like you are actually calling it....

luca-heltai:
This seems to fail:
~~~
simplex/step-06.debug: RUN failed. ------ Additional output on stdout/stderr:


An error occurred in line <889> of file </home/runner/work/dealii/dealii/source/fe/fe.cc> in function
    const dealii::FullMatrix<double>& dealii::FiniteElement<<anonymous>, <anonymous> >::constraints(const dealii::internal::SubfaceCase<dim>&) const [with int dim = 2; int spacedim = 2]
The violated condition was: 
    (this->n_dofs_per_face(face_no) == 0) || (interface_constraints.m() != 0)
Additional information: 
    The finite element for which you try to obtain hanging node constraints does not appear to implement them.

Aborted (core dumped)
CMakeFiles/simplex.step-06.debug.diff.dir/build.make:88: recipe for target 'step-06.debug/output' failed
CMakeFiles/Makefile2:709: recipe for target 'CMakeFiles/simplex.step-06.debug.diff.dir/all' failed
CMakeFiles/Makefile2:716: recipe for target 'CMakeFiles/simplex.step-06.debug.diff.dir/rule' failed
Makefile:382: recipe for target 'simplex.step-06.debug.diff' failed
~~~

tests/simplex/step-06.cc
marcfehling: Will give it a try!

marcfehling:
> This seems to fail:
> 
> ```
> [...]
> ```

Will work once blocking PRs are merged.

bangerth:
It fails with this error:
```

simplex/step-06.debug: RUN failed. ------ Additional output on stdout/stderr:


An error occurred in line <2283> of file </home/runner/work/dealii/dealii/source/fe/mapping_fe.cc> in function
    std::vector<dealii::Point<spacedim> > dealii::MappingFE<dim, spacedim>::compute_mapping_support_points(const typename dealii::Triangulation<dim, spacedim>::cell_iterator&) const [with int dim = 2; int spacedim = 2; typename dealii::Triangulation<dim, spacedim>::cell_iterator = dealii::TriaIterator<dealii::CellAccessor<2, 2> >]
The violated condition was: 
    check_all_manifold_ids_identical(cell)
Additional information: 
    All entities of a cell need to have the same boundary id as the cell has.
```
Can you investigate?

luca-heltai:
> It fails with this error:
> 
> ```
> 
> simplex/step-06.debug: RUN failed. ------ Additional output on stdout/stderr:
> 
> 
> --------------------------------------------------------
> An error occurred in line <2283> of file </home/runner/work/dealii/dealii/source/fe/mapping_fe.cc> in function
>     std::vector<dealii::Point<spacedim> > dealii::MappingFE<dim, spacedim>::compute_mapping_support_points(const typename dealii::Triangulation<dim, spacedim>::cell_iterator&) const [with int dim = 2; int spacedim = 2; typename dealii::Triangulation<dim, spacedim>::cell_iterator = dealii::TriaIterator<dealii::CellAccessor<2, 2> >]
> The violated condition was: 
>     check_all_manifold_ids_identical(cell)
> Additional information: 
>     All entities of a cell need to have the same boundary id as the cell has.
> --------------------------------------------------------
> ```
> 
> Can you investigate?

I think this is due to the fact that currently high order mapping only works for uniform manifold ids (i.e., either all flats, or all curved with the same ids). The current implementation would work for hyper shells (and simplices) but not for balls (where flat manifold / transfinite manifold is used on cell at boundaries, and polar manifold is used on outer cells).

Reset all manifold ids to flat. 

We'll address this at some point. BTW, @peterrum , I guess the error message should have been "All entities of a cell need to have the same manifold id as the cell has.", not "boundary".

marcfehling:
> Reset all manifold ids to flat.

@luca-heltai I'll remove the last commit then. (I didn't test it yet)

bangerth:
Do I understand correctly that you want to let the test stand as is?

marcfehling:
> Do I understand correctly that you want to let the test stand as is?

Yes, I would like to leave it like that for now. Like the other simplex tests, we point out in the preamble what is currently not working. We should address these issues in the future (as @luca-heltai also pointed out). This tests represents what we are currently capable of.

bangerth (approved):

PR #11642, opened 2021-01-28
Move make_triangulation into ReferenceCell::Type

peterrum:
FYI @bangerth @nmuch


masterleinad (approved):

bangerth:
My plan had been to move it into `GridGenerator`, but I guess this is fine as well. Any thoughts about this or that place?

peterrum:
I think this way it is consistent with the other functions in this file. But, I think having a function in `GridGenerator` that simply calls wouldn't be bad either.

bangerth:
I was mostly thinking about where a user would find such a function in the documentation. If you don't mind, I'll move it to `GridGenerator`. That's where we have all of the other generator functions.

PR #11646, opened 2021-01-28
Pass 'const double' objects by-value, rather than by-reference.

bangerth:
/rebuild

Minor follow-up to #11623.

### DIFF
--- a/include/deal.II/fe/fe_point_evaluation.h
+++ b/include/deal.II/fe/fe_point_evaluation.h
@@ -48,7 +48,7 @@ namespace internal
       using gradient_type = Tensor<1, n_components, Tensor<1, dim>>;
 
       static void
-      read_value(const double &     vector_entry,
+      read_value(const double       vector_entry,
                  const unsigned int component,
                  value_type &       result)
       {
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

include/deal.II/fe/fe_point_evaluation.h
```diff
@@ -48,7 +48,7 @@ namespace internal
       using gradient_type = Tensor<1, n_components, Tensor<1, dim>>;
 
       static void
-      read_value(const double &     vector_entry,
+      read_value(const double       vector_entry,
```
peterrum, line 51: I don't understand what is the motivation to switch from const-reference to plain const here. Can you explain @bangerth?

masterleinad: For built-in types passing by reference/pointer often is more expensive than passing by copying. We try to be consistent here.

bangerth: Correct. Passing by value makes it easy for the compiler to reason about what is happening to a variable. We do this everywhere where the argument is just a scalar/small object.

PR #11649, opened 2021-01-28
CMake: Update Sundials version check

tamiko:


tamiko:
@luca-heltai  The Ubuntu 18.04 LTS container actually has sundials 2.7 installed. We were simply missing a version check...

This hopefully fixes https://cdash.43-1.org/viewBuildError.php?buildid=15747

bangerth (commented):
Ah, interesting. I ran into this today on one of my machines that had SUNDIALS 2.7 installed, and that didn't compile because a certain header file was missing. But after commenting out that `#include`, it actually compiled. So that would suggest that 2.7 still works.

I'm now also running the test suite and it turns out that all but one test actually work with 2.7. One is again the same missing header file I believe.

The question I would have then is this: Do we actually need to remove support for SUNDIALS 2.7, or could we support all of these versions at the same time?

tamiko:
@bangerth I think @luca-heltai can comment on this one.

I guess we need the include file for later versions, so we could also simply use a version macro to guard the include.

luca-heltai:
I ran into several issues using sundials 2.7, and upgraded to 3.1 as soon as I could. I see no reasons though for 2.7 not to work, and for 3.1 to require those headers. I have the impression that they could simply be removed. @bangerth did you run into troubles when you commented out the headers, or things just worked? The fact is, we use some obscure internal  features of older sundials, and they tend to move around internal things...

luca-heltai:
I think we could just drop sundials 2.7.

luca-heltai (approved):

luca-heltai:
This is now required for the backport compatibility of the `N_Vector` module #11654  (thanks @sebproell !). There were more internal changes between v2 and v3 of SUNDIALS, that are not compatible with that module.

PR #11651, opened 2021-01-29
Third attempt.

luca-heltai:
Apparently the example that was given won't cancel duplicates on master. Let's try again.


peterrum (approved):

masterleinad:
Did you see that it works for pull requests?

luca-heltai:
> Did you see that it works for pull requests?

It seems to work also for PR. If you check the actions log, you'll see that as soon as someone pushes a new commit to a PR, the new build superseeds the old ones.

PR #11652, opened 2021-01-29
Fix DataOut for mixed meshes and vector quantities

peterrum:
Addresses https://github.com/dealii/dealii/pull/11588#discussion_r566127434.

FYI @bangerth


peterrum:
Only 2 lines have changed. The rest is test.

drwells (approved):
I verified that we do get the expected output components for a mixed mesh.

PR #11653, opened 2021-01-29
Fixed minor warnings.

luca-heltai:
Compiling latest master on my mac, I discovered some warnings that were not present with gcc.

### DIFF
--- a/source/grid/grid_in.cc
+++ b/source/grid/grid_in.cc
@@ -2194,7 +2194,7 @@ GridIn<dim, spacedim>::read_msh(const std::string &fname)
     {{15, 0}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {7, 5}, {6, 6}, {5, 7}}};
 
   // Vertex renumbering, by dealii type
-  const std::vector<std::vector<unsigned int>> gmsh_to_dealii = {
+  const std::array<std::vector<unsigned int>, 8> gmsh_to_dealii = {
     {{0},
      {{0, 1}},
      {{0, 1, 2}},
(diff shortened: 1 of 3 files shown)


drwells (approved):
I'm not sure which way we should go on `const auto l : cell->line_indices()` and similar - if the CI is happy, then I'm happy.

source/grid/grid_in.cc
```diff
@@ -2195,7 +2195,7 @@ GridIn<dim, spacedim>::read_msh(const std::string &fname)
 
   // Vertex renumbering, by dealii type
   const std::vector<std::vector<unsigned int>> gmsh_to_dealii = {
-    {{{0}},
+    {{0},
```
bangerth, line 2198: ?? Isn't this creating an array of length 0, instead of an array of length 1 with one zero entry?

I don't know that it would matter, but it looks strange. Separately, the outer `std::vector` might as well be a `std::array<...,8>`.

luca-heltai:
~~~

int
main(void)
{
  initlog();
  const std::vector<std::vector<unsigned int>> gmsh_to_dealii = {
    {{0}, {{0, 1}}, {{0, 1, 2}}, {{0, 1, 3, 2}}}};

  for (const auto &i : gmsh_to_dealii)
    {
      deallog << "Size: " << i.size() << std::endl;
    }
}
~~~
produces
~~~
DEAL::Size: 1
DEAL::Size: 2
DEAL::Size: 3
DEAL::Size: 4
~~~
on my machine...

luca-heltai: Without this, I get the warning:
~~~
warning: braces around scalar initializer
~~~

bangerth: How bizarre... Thanks for double checking!

bangerth (approved):

PR #11659, opened 2021-02-01
Move some BoundingBox access functions to the header.

drwells:
It's useful to be able to inline these.

### DIFF
--- a/include/deal.II/base/bounding_box.h
+++ b/include/deal.II/base/bounding_box.h
@@ -438,6 +438,24 @@ inline BoundingBox<spacedim, Number>::BoundingBox(const Container &points)
 
 
 template <int spacedim, typename Number>
+inline std::pair<Point<spacedim, Number>, Point<spacedim, Number>> &
+BoundingBox<spacedim, Number>::get_boundary_points()
+{
+  return this->boundary_points;
+}
+
+
+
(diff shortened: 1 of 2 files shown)


include/deal.II/base/bounding_box.h
```diff
@@ -437,6 +437,24 @@ inline BoundingBox<spacedim, Number>::BoundingBox(const Container &points)
 
 
 
+template <int spacedim, typename Number>
+inline std::pair<Point<spacedim, Number>, Point<spacedim, Number>> &
+BoundingBox<spacedim, Number>::get_boundary_points()
```
bangerth, line 442: Unrelated question, but why would we give write access to this information?

bangerth (approved):

include/deal.II/base/bounding_box.h
luca-heltai: It is required by boost to have write access to bonding boxes to be able to convert from and to deal.II bounding boxes in its algorithms.

bangerth: I see. That's a nuisance...

Is there no way to just create a new BB object and then use the assignment operator? Not a question for this PR, but if you wanted to think about it (or if you wanted to point me to the place where this is needed).

PR #11669, opened 2021-02-03
Removed ansi flag from intel compiler setup

nabw:
Small PR to remove -ansi flag from intel compiler, which conflicts with c++14 and thus does not allow for configuration.


masterleinad (approved):
Looks good! Also, see https://godbolt.org/z/n87nvd.

tamiko (approved):

nabw:
> Looks good! Also, see https://godbolt.org/z/n87nvd.

fantastic reference, I had no idea about this. Thanks!

bangerth:
Took us only 18 years to realize that C++ is no longer standardized by ANSI but by ISO ;-)

PR #11670, opened 2021-02-03
Rename PETScWrappers::PreconditionerBase to PreconditionBase

elauksap:
Just a class rename, for consistency with [`TrilinosWrappers::PreconditionBase`](https://github.com/dealii/dealii/blob/eebfb45674f2950ebcf8fd24bd07405b5cb6dc41/include/deal.II/lac/trilinos_precondition.h#L80).

### DIFF
--- a/include/deal.II/lac/petsc_precondition.h
+++ b/include/deal.II/lac/petsc_precondition.h
@@ -53,18 +53,18 @@ namespace PETScWrappers
    *
    * @ingroup PETScWrappers
    */
-  class PreconditionerBase
+  class PreconditionBase
   {
   public:
     /**
      * Constructor.
      */
-    PreconditionerBase();
+    PreconditionBase();
 
     /**
      * Destructor.
      */
-    virtual ~PreconditionerBase();
+    virtual ~PreconditionBase();
 
     /**
      * Destroys the preconditioner, leaving an object like just after having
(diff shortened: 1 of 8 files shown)


masterleinad (commented):
This change is not backward-compatible so we at least need a (deprecated) `using` statement that makes the old name an alias for the new name.

include/deal.II/lac/petsc_precondition.h
```diff
@@ -954,6 +954,12 @@ namespace PETScWrappers
      */
     AdditionalData additional_data;
   };
+
+  /**
+   * Alias for backwards-compatibility.
+   * @deprecated Use PETScWrappers::PreconditionBase instead.
+   */
+  using PreconditionerBase DEAL_II_DEPRECATED = PreconditionBase;
```
elauksap, line 962: @masterleinad should be done, thank you!

masterleinad (commented):
We also need a changelog entry and let's replace `DEAL_II_DEPRECATED` with `DEAL_II_DEPRECATED_EARLY` to adapt to the new policy.

masterleinad (approved):
Fine with me.

bangerth (approved):
Thanks for looking out for the little things such as consistency in a code base! We appreciate these patches!

PR #11671, opened 2021-02-03
Make sure we return all manifold ids.

luca-heltai:
The method `Triangulation::get_manifold_ids()` was not returning all manifold ids, but only the ones on cells and on boundary faces (it was copied from get_boundary_ids(), but without taking into account that we should keep track of line manifolds and of interior face manifolds).


peterrum (approved):

drwells (approved):

bangerth (approved):

bangerth:
Oops!

luca-heltai:
> Oops!

It looks like not many people were using the possibility of attaching a manifold to an edge in 3d... :)

PR #11672, opened 2021-02-03
Make a serialization function 'public'.

bangerth:
/rebuild

### DIFF
--- a/include/deal.II/lac/la_vector.h
+++ b/include/deal.II/lac/la_vector.h
@@ -400,14 +400,6 @@ namespace LinearAlgebra
     memory_consumption() const override;
 
     /**
-     * Attempt to perform an operation between two incompatible vector types.
-     *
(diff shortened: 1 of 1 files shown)


peterrum (approved):

Rombur:
@bangerth Why do you want to do that? The Boost documentation advises to make this function private.
```
it is advisable to always make member serialize functions private. The declaration friend
boost::serialization::access will grant to the serialization library access to private member variables 
and functions.
```
See [here](https://www.boost.org/doc/libs/1_70_0/libs/serialization/doc/tutorial.html)

bangerth:
@Rombur We have these functions `public` for all other classes as well. I have opinion whether that's the right thing to do, but we should be consistent.

bangerth:
For the record, I checked and that is the only place where we make `boost::serialization::access` a `friend`.

PR #11676, opened 2021-02-03
simplex: hp-refinement on hybrid meshes.

marcfehling:
Enables hp-adaptation for hybrid meshes.

Two questions arose while working on this patch:
- Is `Simplex::FE_Poly` the correct class for functions `get_face_interpolation_matrix()` and `get_subface_interpolation_matrix()` that are responsible for building the interpolation matrices?
- ~Since `FE_Q_Base` has an additional `PolynomialType` template parameter, I can not dynamic_cast a general `FiniteElement` object to that particular class. As discussed with @bangerth, I could cast to all dependent classes individually. I decided to only do that to `FE_Q` at this stage of development.~ (EDIT: #11677 dealt with this)


marcfehling:
Will rebase on #11677.

source/grid/grid_tools.cc
```diff
@@ -552,7 +552,7 @@ namespace GridTools
     for (const auto &cell : tria.cell_iterators_on_level(0))
       {
         // Save cell data
-        CellData<dim> cell_data;
+        CellData<dim> cell_data(cell->n_vertices());
```
bangerth, line 555: Is this an unrelated change? I don't quite see the connection. Maybe put it into a separate PR?

tests/simplex/hanging_nodes.h
```diff
@@ -0,0 +1,204 @@
...
+
+#include "../tests.h"
+
+
+
+// ----- setup -----
+
+template <int dim>
+void
+refine(const std::vector<unsigned int> &n_refinements, Triangulation<dim> &tria)
```
bangerth, line 44: Can you explain what the function does?

tests/simplex/hanging_nodes.h
```diff
@@ -0,0 +1,204 @@
...
+
+      AssertDimension(cell->level(), n_refinements[coarse_cell_id]);
+    }
+#endif
+}
+
+
+template <int dim>
+void
+set_active_fe_indices(const std::vector<unsigned int> &fe_indices,
```
bangerth, line 78: same here. I think all of these functions could use some commentary :-)

source/grid/grid_tools.cc
marcfehling: `CellData` takes the `GeometryInfo` equivalent of `n_vertices` as a default argument. This is not correct for simplices and will trigger an assertion later. This change was needed to make `merge_triangulations` work with simplices.

I can't remember which one that was, but I think it was this one
https://github.com/dealii/dealii/blob/5f4d7c1e0dd5afeee39d88bf6fd5d35f16a3c194/source/grid/grid_tools.cc#L594-L606

marcfehling: I've used `merge_triangulation` to generate the hybrid meshes for these tests.

bangerth:
Please rebase now!

marcfehling:
Rebased and ready for review/tests.

marcfehling:
With the `cube_and_pyramid` function, hybrid meshes like the following can be created for testing in both 2D and 3D.
image

marcfehling:
@tamiko This is no longer WIP :-D I would like to have this merged as is.

bangerth (approved):

PR #11677, opened 2021-02-04
Get rid of the PolynomialType template argument of FE_Q_Base.

bangerth:
This is similar to what @GrahamBenHarper did in #8597 and #8711. In short, there is no good reason why the `FE_Q_Base` class should have the polynomial space description as a *template argument*, as opposed to just taking an object as a regular argument in the constructor. Indeed, removing this template argument is not overly complicated, and probably also saves us some compile time and disk space.

Related to #1973.

@marcfehling -- FYI, as discussed earlier.

/rebuild

### DIFF
--- a/doc/news/changes/incompatibilities/20210203Bangerth
+++ b/doc/news/changes/incompatibilities/20210203Bangerth
@@ -1,0 +1,7 @@
+Changed: The FE_Q_Base class took a description of the polynomial
+space as a template argument. But that is not necessary: It is
+entirely sufficient to pass this information to the constructor in the
+form of a regular constructor argument. This has been changed, and the
+class has therefore lost its first template argument.
+<br>
+(Wolfgang Bangerth, 2021/02/03)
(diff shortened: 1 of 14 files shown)


marcfehling (approved):
This will make dynamic_casting between different FiniteElement classes a lot easier in the hp-context.

bangerth:
There is a single failure:
```
	4203 - multigrid/transfer_matrix_free_06.mpirun=15.debug (Timeout)
```
I think that that is unrelated.

masterleinad (approved):

bangerth:
Ping? @marcfehling is waiting on this being merged...

masterleinad:
> Ping? @marcfehling is waiting on this being merged...

I think it's OK to merge your own pull request if it has the `Reviewed and ready to merge label` and all tests passed.

bangerth:
I just don't like doing that. I think we live by peer review, and I'd like others to make that decision to actually pull the trigger.

Thanks, though, for doing it! :-)

PR #11678, opened 2021-02-04
Enable AffineConstraints::copy_from for mixed number (double/float)

peterrum:
FYI @bergbauer


include/deal.II/lac/affine_constraints.h
```diff
@@ -1597,6 +1598,25 @@ class AffineConstraints : public Subscriptor
      */
     number inhomogeneity;
 
+    /**
+     * Copy constructor.
+     */
+    template <typename ConstraintLineType>
+    void
+    operator=(const ConstraintLineType &other) const
+    {
```
bangerth, line 1607: Move it out of line.

include/deal.II/lac/affine_constraints.h
```diff
@@ -1597,6 +1598,25 @@ class AffineConstraints : public Subscriptor
...
+      this->index = other.index;
+
+      entries.clear();
+      entries.reserve(other.entries.size());
+
+      for (const auto &i : other.entries)
+        entries.emplace_back(i.first, i.second);
```
bangerth, line 1614: I think you can do this simpler by saying
```
  entries.clear();
  entries.insert (other.entries.begin(), other.entries.end());
```

bangerth (commented):
Looks mostly ok. Might be nice to have a test. 

This can only work if you have types for the scalar type that are actually convertible into each other. It might be nice to check that with a `static_assert` so people get a reasonable error message. You could use `std::is_constructible<To,From>` in that `static_assert`.

peterrum:
@bangerth I have addressed your comments. Ready from my side!

include/deal.II/lac/affine_constraints.h
```diff
@@ -2467,6 +2507,49 @@ AffineConstraints<number>::add_entries_local_to_global(
...
+
+template <typename number>
+template <typename ConstraintLineType>
+inline typename AffineConstraints<number>::ConstraintLine &
+AffineConstraints<number>::ConstraintLine::
+operator=(const ConstraintLineType &other)
+{
+  this->index = other.index;
+
+  entries.insert(entries.begin(), other.entries.begin(), other.entries.end());
```
bangerth, line 2546: I think you need to clear `entries` before the `insert` call.

tests/lac/constrains_copy_from_mixed.cc
```diff
@@ -0,0 +1,63 @@
+// ---------------------------------------------------------------------
```
bangerth, line 1: I suspect that you meant to say "constraints" instead of "constrains" in the name of the file.

peterrum:
@bangerth You are right! Fixed!

bangerth (approved):

Issue #11591, opened 2021-01-19
Bug in exchange_ghost_particles


shahabgol:
I have observed a bug in the exchange_ghost_particles function. It is related to the new architecture of particle properties. I think this function is not updated with the recent changes of particles class.

blaisb:
So @shahabgol  found this bug when using Lethe, but I managed to reproduce it using step-68 (@bangerth  I really need to write a test using this one).

Using step-68 and the following set of parameters:

```
# Listing of Parameters
# ---------------------
subsection Particle Tracking Problem
  # End time of the simulation
  set Final time                    = 4

  # Refinement level of the fluid domain
  set Fluid refinement              = 4
  set Output directory              = ./

  # Iteration frequency at which output results are written
  set Output frequency              = 10

  # Refinement of the volumetric mesh used to insert the particles
  set Particle insertion refinement = 3

  # Iteration frequency at which the mesh is load balanced
  set Repartition frequency         = 1000
  set Time step                     = 0.002
  set Velocity degree               = 1
end
```

Step-68 will crash with just 2 processors. The error I get is the following:
```
--------------------------------------------------------
An error occurred in line <302> of file </home/blaisb/work/dealii_main/dealii/include/deal.II/particles/property_pool.h> in function
    dealii::ArrayView<double> dealii::Particles::PropertyPool<dim, spacedim>::get_properties(dealii::Particles::PropertyPool<dim, spacedim>::Handle) [with int dim = 2; int spacedim = 2; dealii::Particles::PropertyPool<dim, spacedim>::Handle = unsigned int]
The violated condition was: 
    data_index <= properties.size() - n_properties
Additional information: 
    Invalid property handle. This can happen if the handle was duplicated and then one copy was deallocated before trying to access the properties.

Stacktrace:
-----------
#0  /home/blaisb/work/dealii_main/inst/lib/libdeal_II.g.so.9.3.0-pre: dealii::Particles::PropertyPool<2, 2>::get_properties(unsigned int)
#1  /home/blaisb/work/dealii_main/inst/lib/libdeal_II.g.so.9.3.0-pre: dealii::Particles::PropertyPool<2, 2>::register_particle()
#2  /home/blaisb/work/dealii_main/inst/lib/libdeal_II.g.so.9.3.0-pre: dealii::Particles::Particle<2, 2>::Particle(void const*&, dealii::Particles::PropertyPool<2, 2>*)
[...]

PR #11679, opened 2021-02-04
Add step-68 as a test

blaisb:
This PR adds step-68 as a test. This currently work on a single processor. However, because of a bug in particles properties (see #11591), this code does not work on more than one processor with the current set of parameter. This is exactly why I added it in this fashion, in order to have a way to reproduce this bug :).
I have not been able to identify the source of the bug yet, but I thought that by opening this PR this would start the discussion on this issue. Without this, the majority of codes using the particle do not work with the current deal.II master.

The test is very short (less than 15s in debug mode).


blaisb:
@bangerth  the bug appeared when you changed the particle properties to store the particle position. I think crashes occur when a processor has no more particles.

tests/particles/step-68.with_p4est=true.output
```diff
@@ -0,0 +1,64 @@
+
+DEAL::Number of particles inserted: 24
```
masterleinad, line 2: You should rename this file to "tests/particles/step-68.with_p4est=true.mpirun=1.output" to avoid subtle problems when running the testsuite.

blaisb: Done, good catch :)!

peterrum (approved):

PR #11680, opened 2021-02-04
Provide correct mappings to function calls.

bangerth:
The `GridTools` functions being called have default arguments. In the case of the call in `grid_generator.cc`, the default argument is only correct if the mesh is a hypercube mesh, and we would presumably fail otherwise.

In the case of the call in `particle_handler.cc`, I believe that the omission of the mapping is an actual bug. @gassmoeller @luca-heltai 

/rebuild

### DIFF
--- a/source/particles/particle_handler.cc
+++ b/source/particles/particle_handler.cc
@@ -1033,7 +1033,7 @@ namespace Particles
 
           const unsigned int closest_vertex =
             GridTools::find_closest_vertex_of_cell<dim, spacedim>(
-              current_cell, out_particle->get_location());
+              current_cell, out_particle->get_location(), *mapping);
           Tensor<1, spacedim> vertex_to_particle =
             out_particle->get_location() - current_cell->vertex(closest_vertex);
           vertex_to_particle /= vertex_to_particle.norm();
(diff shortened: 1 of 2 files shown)


source/particles/particle_handler.cc
```diff
@@ -1033,7 +1033,7 @@ namespace Particles
 
           const unsigned int closest_vertex =
             GridTools::find_closest_vertex_of_cell<dim, spacedim>(
-              current_cell, out_particle->get_location());
+              current_cell, out_particle->get_location(), *mapping);
```
gassmoeller, line 1036: I agree this looks like a bug, this accidental omission of the mapping parameter is happening quite often :disappointed: . But this instance also confuses me, this is only relevant for mappings that do not preserve vertex locations, right? So at least in the case of ASPECT it is only problematic when we use a MappingQEulerian for mesh deformation.

luca-heltai (approved):

source/particles/particle_handler.cc
bangerth: I think that's correct.

bangerth:
The one failing check is because of the test that perpetually times out...

peterrum (approved):

bangerth:
I would appreciate this getting merged soon since I need it for #11674. I will then rebase #11674 on it.

bangerth:
In reference to #11540.

PR #11683, opened 2021-02-04
CMake: record the C++ language standard detected during configure

tamiko:


bangerth (approved):
We used to output this in the summary at the bottom and I've missed it lately. I like having it back.

bangerth:
/rebuild

The Windows build fails because it couldn't upload data. I think we can ignore this.

masterleinad (approved):

PR #11685, opened 2021-02-04
Add q witherden vincent

### LINKED ISSUE #11636: Implement Simplex::QGauss(4) without copying and pasting
We can generate a lot of simplex quadrature rules by doing permutations with barycentric coordinates. This issue is a reminder that I should investigate this and implement the actual algorithmic description of this family of rules (which is implemented in quadpy as the Witherden Vincent formulas).

In particular - quadpy implements a lot of simplex formulas with JSON as
```json
{
  "name": "Witherden-Vincent 3",
  "domain": "T3",
  "degree": 3,
  "test_tolerance": 2.379e-17,
  "data": {
    "s31": [
      [
        1.362178425370874e-01,
        1.137821574629126e-01
      ],
      [
        3.281633025163817e-01,
        1.080472498984286e-01
      ]
    ]
  }
}

```
where `s31` is a barycentric permutation function. This is a lot better than copying and pasting tables.

drwells:
Fixes #11636. In particular - some quadrature formulae for simplices have points written as permutations of barycentric coordinates. This PR adds one such rule, rather than hard-coding points (though some numerical values for these rules do have to be hard-coded).


source/simplex/quadrature_lib.cc
```diff
@@ -263,6 +159,248 @@ namespace Simplex
...
+  } // namespace
+
+
+
+  template <int dim>
+  QWitherdenVincent<dim>::QWitherdenVincent(const unsigned int n_points_1D)
+    : QSimplex<dim>(Quadrature<dim>())
+  {
+    std::array<double, dim + 1> centroid;
+    std::fill(centroid.begin(), centroid.end(), 1.0 / (dim + 1));
```
peterrum, line 192: Let's move these lines below, since they are not used in 1D.

peterrum (approved):

source/simplex/quadrature_lib.cc
```diff
@@ -84,41 +84,9 @@ namespace Simplex
...
-            this->weights.emplace_back(q12 * 0.069274682079417);
-            this->weights.emplace_back(q12 * 0.07085308369213365);
-            this->weights.emplace_back(q12 * 0.069274682079417);
-            this->weights.emplace_back(q12 * 0.07085308369213365);
-            this->weights.emplace_back(q12 * 0.069274682079417);
-            this->weights.emplace_back(q12 * 0.07085308369213365);
-            this->weights.emplace_back(q12 * 0.069274682079417);
+            QWitherdenVincent<dim> quad(n_points_1D);
+            this->quadrature_points = quad.get_points();
+            this->weights           = quad.get_weights();
```
bangerth, line 89: I think you should be able to just write this as
```
  *this = QWitherdenVincent<dim>(n_points_1D);
```
But I don't quite understand. The documentation for `Simplex::QGauss` says that for n=4, you have 15 points, whereas QW...V... says it's 19 points.

source/simplex/quadrature_lib.cc
```diff
@@ -176,81 +144,9 @@ namespace Simplex
...
-            this->weights.emplace_back(Q16 * 0.008110770829903342);
-            this->weights.emplace_back(Q16 * 0.03720713072833462);
-            this->weights.emplace_back(Q16 * 0.008110770829903342);
-            this->weights.emplace_back(Q16 * 0.03720713072833462);
-            this->weights.emplace_back(Q16 * 0.008110770829903342);
-            this->weights.emplace_back(Q16 * 0.03720713072833462);
-            this->weights.emplace_back(Q16 * 0.008110770829903342);
+            QWitherdenVincent<dim> quad(n_points_1D);
+            this->quadrature_points = quad.get_points();
+            this->weights           = quad.get_weights();
```
bangerth, line 149: same suggestions and questions here

source/simplex/quadrature_lib.cc
```diff
@@ -263,6 +159,248 @@ namespace Simplex
...
+  QWitherdenVincent<dim>::QWitherdenVincent(const unsigned int n_points_1D)
+    : QSimplex<dim>(Quadrature<dim>())
+  {
+    // Just use Gauss in 1D: this is a high-order open rule so this is a
+    // reasonable equivalent for generic programming.
+    if (dim == 1)
+      {
+        const dealii::QGauss<dim> quad(n_points_1D);
+        this->quadrature_points = quad.get_points();
+        this->weights           = quad.get_weights();
```
bangerth, line 197: same here with `*this = ...`

source/simplex/quadrature_lib.cc
```diff
@@ -263,6 +159,248 @@ namespace Simplex
...
+    // by quite a bit by exploiting symmetry - we do essentially the
+    // same thing for each barycentric coordinate, so we can express
+    // our quadrature rule as permutations of barycentric points
+    // instead of writing things out explicitly.
+
+    // Apply a Barycentric permutation where one point is different.
+    auto process_point_1 = [&](const double a, const double w) {
+      const double                b = 1.0 - dim * a;
+      std::array<double, dim + 1> b_point;
+      std::fill(b_point.begin(), b_point.end(), a);
```
bangerth, line 216:
```suggestion
      std::fill(b_point.begin(), b_point.end()-1, a);
```

bangerth: Or, maybe to make the point more clear:
```suggestion
      std::fill(b_point.begin(), b_point.begin()+dim, a);
```

source/simplex/quadrature_lib.cc
```diff
@@ -263,6 +159,248 @@ namespace Simplex
...
+      b_weights.push_back(w);
+      b_point_permutations.push_back(all_permutations(b_point));
+    };
+
+    // Apply a Barycentric permutation where two points (in 3D) are different.
+    auto process_point_2 = [&](const double a, const double w) {
+      Assert(dim == 3, ExcInternalError());
+      const double                b = (1.0 - 2.0 * a) / 2.0;
+      std::array<double, dim + 1> b_point;
+      std::fill(b_point.begin(), b_point.end(), a);
```
bangerth, line 228: same idea here and probably also for the third of these functions

source/simplex/quadrature_lib.cc
```diff
@@ -263,6 +159,248 @@ namespace Simplex
...
+        if (dim == 2)
+          {
+            process_point_1(9.1576213509770743e-02, 1.0995174365532187e-01);
+            process_point_1(4.4594849091596489e-01, 2.2338158967801147e-01);
+          }
+        else if (dim == 3)
+          {
+            process_point_1(3.281633025163817e-01, 1.362178425370874e-01);
+            process_point_1(1.080472498984286e-01, 1.137821574629126e-01);
+          }
```
bangerth, line 266: could you add
```suggestion
          }
       else
         Assert (false, ExcNotImplemented());
```
here and to the corresponding cases below? This way, if someone for whatever reason tried to do `dim==4`, there would at least be an error message.

source/simplex/quadrature_lib.cc
```diff
@@ -263,6 +159,248 @@ namespace Simplex
...
+      }
+
+    Assert(b_point_permutations.size() == b_weights.size(), ExcInternalError());
+    for (unsigned int permutation_n = 0; permutation_n < b_weights.size();
+         ++permutation_n)
+      {
+        for (const std::array<double, dim + 1> &b_point :
+             b_point_permutations[permutation_n])
+          {
+            const double volume = dim == 2 ? 1.0 / 2.0 : 1.0 / 6.0;
```
bangerth, line 393:
```suggestion
            const double volume = (dim == 2 ? 1.0 / 2.0 : 1.0 / 6.0);
```

tests/simplex/q_witherden_vincent_01.cc
```diff
@@ -0,0 +1,82 @@
+#include <deal.II/base/function_lib.h>
```
bangerth, line 1: please add the customary header here

tests/simplex/q_witherden_vincent_01.cc
```diff
@@ -0,0 +1,82 @@
...
+
+#include <deal.II/simplex/quadrature_lib.h>
+
+#include "../tests.h"
+
+template <int dim>
+void
+print(unsigned int n_points_1D)
```
bangerth, line 9:
```suggestion
print(const unsigned int n_points_1D)
```
Same in a couple more places below

include/deal.II/simplex/quadrature_lib.h
```diff
@@ -51,6 +51,35 @@ namespace Simplex
...
+   * of what polynomial degree to be integrated exactly (e.g., for $n$ points,
+   * the rule can integrate polynomials of degree $2 n - 1$ exactly.). The given
+   * value for n_points_1D = 1, 2, 3, 4, 5 results in the following number of
+   * quadrature points in 2D and 3D:
+   * - 2D: 1, 6, 7, 15, 19
+   * - 3D: 1, 8, 14, 35, 59
+   *
+   * For 1D, the quadrature rule degenerates to a `QGauss<1>(n_points_1D)`.
+   *
+   * These rules match the ones listed for Witherden-Vincent in the quadpy
```
bangerth, line 67: maybe make 'quadpy' a link?

source/simplex/quadrature_lib.cc
drwells: Good catch on the documentation. The copy constructor didn't work so I went ahead and did it manually - I think we need to add a copy constructor to `SimplexQuadrature`.

bangerth: Why doesn't it just use the copy constructor of the `Quadrature` base class?

source/simplex/quadrature_lib.cc
drwells: Would you be OK with me just adding this check at the top of the function? I'd rather write it once there than for every single `n_points_1D` value.

bangerth: I guess that's ok :-)

source/simplex/quadrature_lib.cc
drwells: > But I don't quite understand. The documentation for Simplex::QGauss says that for n=4, you have 15 points, whereas QW...V... says it's 19 points.

I think you misread the `n = 5` case for QWV as `n = 4` - we have an extra rule here that's not yet in QGauss.

include/deal.II/simplex/quadrature_lib.h
drwells: Good idea - we can add it to the bibliography.

source/simplex/quadrature_lib.cc
drwells: The copy operator doesn't work this way. `Quadrature<dim>` defines a copy operator. However, each inheriting class defines its own implicit copy operator, which hides the explicitly-defined base class operator. I implemented the fix with `static_cast`.

drwells:
I implemented all the requested fixes with the possible exception of the one involving `operator=`, which is a problem shared by all the other quadrature rules.

source/simplex/quadrature_lib.cc
bangerth: Can't you just to `this->Quadrature<dim>::operator=(rhs);`? You could even omit the `this->`, I just added it for clarity.

drwells: Oh yeah, that's a much better idea.

drwells: I implemented that fix.

doc/doxygen/references.bib
```diff
@@ -1089,3 +1089,22 @@ @article{Geevers_2018
...
+  author={Witherden, Freddie D and Vincent, Peter E},
+  journal={Computers \& Mathematics with Applications},
+  volume={69},
+  number={10},
+  pages={1232--1241},
+  year={2015},
+  publisher={Elsevier}
+}
+
+@online{quadpy,
```
bangerth, line 1104: Is that an actual category that doxygen understands?

bangerth: I think that quadpy has a DOI. You can always convert DOIs to bibtext format at https://www.doi2bib.org

bangerth (approved):

doc/doxygen/references.bib
drwells: It worked with my copy of doxygen.

kronbichler (approved):

peterrum (approved):
Thanks @drwells! It is great to have a Mathematician on board of the simplex project!

PR #11686, opened 2021-02-04
Disable warnings before including boost header

Rombur:
With C++20 we get a bunch of warnings from `boost::small_vector`. This PR disables the warnings before including the header.

### DIFF
--- a/include/deal.II/dofs/dof_accessor.h
+++ b/include/deal.II/dofs/dof_accessor.h
@@ -26,7 +26,9 @@
 
 #include <deal.II/hp/dof_handler.h>
 
+DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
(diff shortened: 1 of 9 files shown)


drwells:
We include this header in a bunch of other places (like `manifold.cc`) - could you apply the same fix there?

bangerth:
Out of curiosity, what are these warnings?

Rombur:
> Out of curiosity, what are these warnings?

```warning: definition of implicit copy assignment operator is deprecated because it has a user-declared copy constructor [-Wdeprecated-copy]```

Rombur:
> We include this header in a bunch of other places (like manifold.cc) - could you apply the same fix there?

Done

drwells (approved):

PR #11689, opened 2021-02-05
Move the get/set_bit() functions into namespace Utilities.

bangerth:
Follow-up to #11544.

/rebuild


peterrum (approved):
This is the third place now ;)

PR #11694, opened 2021-02-06
Remove a couple of unnecessary forward declarations.

bangerth:
Follow-up to #11544.

/rebuild

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -34,13 +34,9 @@ template <int dim, int spacedim>
 class Mapping;
 template <int dim>
 class Quadrature;
-template <int dim>
-class ScalarPolynomialsBase;
 #endif
 
-/**
- * A namespace for reference cells.
- */
+
 class ReferenceCell;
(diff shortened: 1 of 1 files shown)


include/deal.II/grid/reference_cell.h
```diff
@@ -34,13 +34,9 @@ template <int dim, int spacedim>
...
 class Quadrature;
-template <int dim>
-class ScalarPolynomialsBase;
 #endif
 
-/**
- * A namespace for reference cells.
- */
+
 class ReferenceCell;
```
masterleinad, line 40: This should probably also be hidden from `doxygen`.

peterrum (approved):

include/deal.II/grid/reference_cell.h
bangerth: Since it tests ok and I'd like to clear my backlog of patches, would you allow me to do this in a follow-up?

masterleinad: Sure.

PR #11632, opened 2021-01-26
Removed wrapping for reference cell.

luca-heltai:
Closes #11626.

Since the enum has been removed, I just removed the corresponding wrapper. This won't be compiled by any of the testers, except on merge to master, so none of the tests is relevant, except indent. 

@peterrum, @agrayver

PR #11697, opened 2021-02-06
Fix python wrapper.

adamqc:
In #11632, the wrapping for reference cell was removed, but it still has some legacies. This breaks the python wrapper.


peterrum:
@agrayver Could take a look here? Thanks!

bangerth (approved):

masterleinad:
Should fix https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=16341.

PR #11698, opened 2021-02-06
Accept unique_ptr in different MG-related classes

peterrum:
PR #11699 introduces a base class `MGSolverOperatorBase` that would be accepted by the new class `MGSolver`. Users would need to implement `MGSolverOperatorBase` with relevant functionalities as needed by the selected smoother/transfer operator as it is done in step-75 for the matrix-free hp case:

https://github.com/dealii/dealii/blob/aba3dd712467cba2ead2676b7d468f62684dca75/examples/step-75/step-75.cc#L318-L624

I am reluctant to template `MGSolver` with the the "OperatorType" but would like to pass reference (`std::unique_ptr`) to the base class. Since some MG-related classes (e.g., `MGLevelObjects`) cannot deal with virtual and pure virtual classes `T` directly, they need to be of type `std::unique_ptr<T>` (or something similar).

This PR was requested by https://github.com/dealii/dealii/pull/11254#pullrequestreview-583215678.

### DIFF
--- a/include/deal.II/base/utilities.h
+++ b/include/deal.II/base/utilities.h
@@ -763,20 +763,42 @@ namespace Utilities
    */
   template <typename To, typename From>
   std::unique_ptr<To>
-  dynamic_unique_cast(std::unique_ptr<From> &&p)
-  {
-    // Let's see if we can cast from 'From' to 'To'. If so, do the cast,
-    // and then release the pointer from the old
-    // owner
-    if (To *cast = dynamic_cast<To *>(p.get()))
-      {
-        std::unique_ptr<To> result(cast);
-        p.release();
-        return result;
-      }
-    else
-      throw std::bad_cast();
-  }
+  dynamic_unique_cast(std::unique_ptr<From> &&p);
+
+  /**
+   * Return underlying value. Default: return input.
+   */
+  template <typename T>
+  T &
+  get_underlying_value(T &p);
+
+  /**
+   * Return underlying value. Specialization for std::shared_ptr<T>.
+   */
+  template <typename T>
+  T &
+  get_underlying_value(std::shared_ptr<T> &p);
+
+  /**
+   * Return underlying value. Specialization for const std::shared_ptr<T>.
+   */
+  template <typename T>
+  T &
+  get_underlying_value(const std::shared_ptr<T> &p);
+
+  /**
+   * Return underlying value. Specialization for std::unique_ptr<T>.
+   */
+  template <typename T>
+  T &
+  get_underlying_value(std::unique_ptr<T> &p);
+
+  /**
+   * Return underlying value. Specialization for const std::unique_ptr<T>.
+   */
+  template <typename T>
+  T &
+  get_underlying_value(const std::unique_ptr<T> &p);
 
   /**
    * A namespace for utility functions that probe system properties.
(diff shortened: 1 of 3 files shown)


masterleinad:
Why are these overloads necessary? Is there something you can't achieve when dereferencing a given `std::unique_ptr` when calling these interfaces?

luca-heltai:
> Why are these overloads necessary? Is there something you can't achieve when dereferencing a given `std::unique_ptr` when calling these interfaces?

There are cases where you construct these objects on the fly, and you don't want to book keep where they have been created  to ensure that things live longer than the MG classes.

luca-heltai (approved):

peterrum:
I have remove the changes from `MGTransferGlobalCoarsening`, since I have a different plan here (see #11708).

masterleinad:
> There are cases where you construct these objects on the fly, and you don't want to book keep where they have been created to ensure that things live longer than the MG classes.

At least in this pull request, the `std::unique_ptr` are immediately dereferenced so that there is no practical difference to passing references directly. So it doesn't make any difference with respect to life time of these objects.

Also, I don't see how #11669 would do anything really different. Maybe you could elaborate some more (godbolt?). I will not block, though, if y'all are convinced this the proper way to go forward.

peterrum:
@masterleinad Are referencing the right issue/PR?

masterleinad:
> @masterleinad Are referencing the right issue/PR?

Fixed. Can you give me an example for "Since some MG-related classes (e.g., MGLevelObjects) cannot deal with virtual and pure virtual classes T directly, they need to be of type std::unique_ptr<T> (or something similar)." for simple classes?

masterleinad:
I'm just saying that passing a `const std::unique_ptr<T>&` does not convey any more meaning than passing by `const T&` directly. In particular, you can't transfer ownership at all.
For fun, I also tried to replace the `const std::unique_ptr<>&` interface in `Patterns` (which is more or less the only place where we use them) by references and didn't run into any issues.

I am happy to be convinced of a case where we really need to pass a `std::unique_ptr` by `const &`. :slightly_smiling_face:

bangerth:
It is correct that in `pattern.h`, we could replace the `std::unique_ptr` arguments for the `to_string()` and `to_value()` functions to regular const references. I would accept a patch to this end.

peterrum:
This is a failing test:
``` cpp
#include <deal.II/base/mg_level_object.h>

using namespace dealii;

class T
{
public:
    virtual void test() = 0;
};

int main()
{
    MGLevelObject<T> objects;
}
```
Also if `T` would be not pure virtual, but I would like to have `MGLevelObject` of type `T` (under the assumption that it is the base class), I cannot plug in arbitrary derived classes.

include/deal.II/multigrid/mg_matrix.h
```diff
@@ -214,6 +227,22 @@ namespace mg
...
+  inline void
+  Matrix<VectorType>::initialize(
+    const MGLevelObject<std::unique_ptr<MatrixType>> &p)
+  {
+    matrices.resize(p.min_level(), p.max_level());
+    for (unsigned int level = p.min_level(); level <= p.max_level(); ++level)
+      {
+        matrices[level] =
+          linear_operator<VectorType>(LinearOperator<VectorType>(), *p[level]);
+      }
```
kronbichler, line 241: Maybe I misunderstand, but since all the methods are templated on `MatrixType`, can't we pick this up with the existing interface with `MatrixType=std::unique_ptr<SomeActualMatrixType>` and then have an internal helper function that translates the access from a pointer type to a dereference, i.e.,
```c++
       matrices[level] =
          linear_operator<VectorType>(LinearOperator<VectorType>(), maybe_dereference_pointer(p[level]));
```

peterrum: I will try this out.

peterrum: OK. It does work. I would suggest to introduce in the namepace `Utilities` (in base/utilties.h) the following function(s):
```  
  template <typename T>
  T &
  maybe_dereference_pointer(T &p)
  {
      return p;
  }
  
  template <typename T>
  T &
  maybe_dereference_pointer(const std::shared_ptr<T> &p)
  {
      return *p;
  }
  
  template <typename T>
  T &
  maybe_dereference_pointer(const std::unique_ptr<T> &p)
  {
      return *p;
  }
```

Any thoughts on these?

kronbichler: I would like this solution, but I am biased. @masterleinad ?

masterleinad:
> This is a failing test:
[...]

Right. Why do you want to construct a `MGLevelObject<T> object in the first place? What do you want to do with it? Where does `std::unique_ptr` come into play?

include/deal.II/multigrid/mg_matrix.h
masterleinad: I don't like having `maybe` in a function name but something like `get_underlying_value` or similar. I general I prefer this approach over introducing new overloads but I would really like to understand if there is a use case for passing by `const std::unique_ptr<>&` in the first place.

peterrum:
I have updated this PR to use `Utilities::get_underlying_value()`.

kronbichler (approved):

include/deal.II/base/utilities.h
```diff
@@ -763,20 +763,28 @@ namespace Utilities
...
+  template <typename T>
+  T &
+  get_underlying_value(const std::shared_ptr<T> &p);
+
+  /**
+   * Return underlying value. Specialization for std::unique_ptr.
+   */
+  template <typename T>
+  T &
+  get_underlying_value(const std::unique_ptr<T> &p);
```
masterleinad, line 787: I think we should also provide an overload for `std::unique_ptr<T>&`, see https://godbolt.org/z/7aG1xG. Of course, we also have an issue with temporary objects but that's really hard to workaround.

peterrum: Done. Thanks a lot for the investigation @masterleinad!

masterleinad (approved):
I think this is good enough.

kronbichler (approved):

PR #11699, opened 2021-02-06
Introduce MGSolver and MGSolverOperatorBase

peterrum:
As discussed offline a couple of weeks ago, I would like to improve the usability of the deal.II multigrid infrastructure by introducing an off-the-shelf `MGSolver` class. It would have a list of possible smoothers from which users can select by setting the correct parameters. Adding new ones would be also possible due to some virtual functions.

In the optimal case, the user only has to provide the operators for each level (and the finest level) and the transfer operator (and of course the parameters). The users would need to implement the base class `MGSolverOperatorBase` for their operators. The `MGSolverOperatorBase` class would define the interface needed for the most common smoothers and transfer operators (for local-smoothing as well as for global coarsening). Depending on which smoother/transfer operator is selected different methods need to be implemented (in some cases you need the edge effects in other cases not).

This PR is still very much WIP and does not contain everything I have described. The first use case of this class would be step-75 (https://github.com/dealii/dealii/pull/11254) where we intend to concentrate on matrix-free hp and multigrid is not the focus. However, I am involved in different projects where we have implemented similar multigrid classes and we could instead of using those we could switch to an official version.

### DIFF
--- a/include/deal.II/multigrid/mg_solver.h
+++ b/include/deal.II/multigrid/mg_solver.h
@@ -1,0 +1,343 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2021 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_mg_solver_h
+#define dealii_mg_solver_h
+
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/base/mg_level_object.h>
+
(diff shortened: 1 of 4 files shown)


include/deal.II/multigrid/mg_solver.h
```diff
@@ -0,0 +1,343 @@
...
+  // Finally, solve.
+  SolverCG<VectorType>(solver_control)
+    .solve(fine_matrix, dst, src, preconditioner);
+}
+
+#ifndef DOXYGEN
+
+
+
+#endif // DOXYGEN
```
marcfehling, line 339:
```suggestion
```

marcfehling (commented):
For now, I'll only suggest to move your function definitions in `mg_solver.h` into a separate `cc` file :-)
We will otherwise run into linker problems if this fill will be included in two different locations.

peterrum:
@marcfehling This is very much WIP ;)

drwells:
I'd like to use this for preconditioning a Helmholtz problem in #12467. In fact, since the nonlinearity is handled explicitly, getting a good preconditioner is one of the biggest remaining parts. @peterrum do you have time to work on this soon?

bangerth:
Should we bump this to post-9.5?

PR #11701, opened 2021-02-06
Deal with no MPI in sundials/n_vector.templates.h

masterleinad:
Should fix https://cdash.43-1.org/viewBuildError.php?buildid=16342.

### DIFF
--- a/include/deal.II/sundials/n_vector.templates.h
+++ b/include/deal.II/sundials/n_vector.templates.h
@@ -492,7 +492,7 @@ template <typename VectorType,
           std::enable_if_t<is_serial_vector<VectorType>::value, int>>
 MPI_Comm SUNDIALS::internal::NVectorOperations::get_communicator(N_Vector)
 {
-  return {};
+  return MPI_COMM_SELF;
 }
 
 
--- a/include/deal.II/sundials/n_vector.templates.h
+++ b/include/deal.II/sundials/n_vector.templates.h
@@ -279,7 +279,12 @@ namespace SUNDIALS
       inline void *
       get_communicator_as_void_ptr(N_Vector v)
       {
+#  ifndef DEAL_II_WITH_MPI
+        (void)v;
+        return nullptr;
+#  else
         return get_communicator<VectorType>(v);
+#  endif
       }
 
     } // namespace NVectorOperations
@@ -487,7 +492,7 @@ template <typename VectorType,
           std::enable_if_t<is_serial_vector<VectorType>::value, int>>
 MPI_Comm SUNDIALS::internal::NVectorOperations::get_communicator(N_Vector)
 {
-  return nullptr;
+  return {};
 }
 
 


peterrum:
@sebproell?

sebproell:
@masterleinad Thanks for fixing a bug introduced by me :sweat_smile:  I think we also need to change the declaration in line 256. Does it make sense to change the return value for the serial case to `nullptr_t` instead of `void *`?

masterleinad:
@sebproell I needed to change the approach a little after actually trying to compile the code. :slightly_smiling_face: Please have a look now.

include/deal.II/sundials/n_vector.templates.h
```diff
@@ -487,7 +492,7 @@ template <typename VectorType,
           std::enable_if_t<is_serial_vector<VectorType>::value, int>>
 MPI_Comm SUNDIALS::internal::NVectorOperations::get_communicator(N_Vector)
 {
-  return nullptr;
+  return {};
```
sebproell, line 495: So this returns a default initialized value of whatever type MPI_Comm is. It would be an int (?) if MPI is not included but that doesn't matter due to the new #ifndef? Makes sense I guess. Thanks again!

masterleinad: Yes, exactly. We return `nullptr` in `get_communicator_as_void_ptr` if deal.II is built without MPI support or if a serial vector is provided. This seems to match the intent of the code you have written.

bangerth: I think this should simply be `MPI_COMM_SELF` if you have nothing better to return. We define this in `base/mpi.h` even if we don't have MPI:
```
#if !defined(DEAL_II_WITH_MPI) && !defined(DEAL_II_WITH_PETSC)
// without MPI, we would still like to use
// some constructs with MPI data
// types. Therefore, create some dummies
using MPI_Comm     = int;
using MPI_Request  = int;
using MPI_Datatype = int;
using MPI_Op       = int;
#  ifndef MPI_COMM_WORLD
#    define MPI_COMM_WORLD 0
#  endif
#  ifndef MPI_COMM_SELF
#    define MPI_COMM_SELF 0
#  endif
```

sebproell: Perfect! That's what SUNDIALS expects of us;)

peterrum (approved):

bangerth (approved):

include/deal.II/sundials/n_vector.templates.h
```diff
@@ -279,7 +279,12 @@ namespace SUNDIALS
       inline void *
       get_communicator_as_void_ptr(N_Vector v)
       {
+#  ifndef DEAL_II_WITH_MPI
+        (void)v;
+        return nullptr;
+#  else
```
sebproell, line 285: @masterleinad fyi: this isn't compatible with the test suite for versions > 5.0.0 (which can currently only run on my local machine...) and violates a SUNDIALS contract.
From the SUNDIALS guide on `N_VGetCommunicator`:
> Returns a pointer to the MPI_Comm object associated with the vector (if applicable). For MPI-unaware vector
implementations, this should return NULL.

Problem is that the serial vector will return MPI_COMM_SELF, which is not the nullptr (if MPI is enabled). I mentioned that here https://github.com/dealii/dealii/pull/11701#discussion_r571460766 but somehow forgot when @bangerth brought up the idea with MPI_COMM_SELF :blush:

I will provide a fix.

PR #11706, opened 2021-02-06
Add utility functions to create polynomial coarsening sequences

peterrum:

PR #11707, opened 2021-02-06
Add utility functions to create geometric coarsening sequences

peterrum:


include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1620,6 +1619,81 @@ namespace internal
...
+  template <int dim, int spacedim>
+  std::vector<std::shared_ptr<Triangulation<dim, spacedim>>>
+  create_geometric_coarsening_sequence(
+    const Triangulation<dim, spacedim> &fine_triangulation_in)
+  {
+    std::vector<std::shared_ptr<Triangulation<dim, spacedim>>>
+      coarse_grid_triangulations;
+
+#ifndef DEAL_II_WITH_P4EST
+    (void)fine_triangulation_in;
```
marcfehling, line 1634: You may want to throw an assertion here as well since this hasn't been implemented for any other Triangulation besides the `p::d` one.

include/deal.II/multigrid/mg_transfer_global_coarsening.h
```diff
@@ -37,6 +37,27 @@ namespace internal
...
+ * Global coarsening utility functions.
+ */
+namespace MGTransferGlobalCoarseningTools
+{
+  /**
+   * For a given triangulation @p tria, determine the geometric coarsening
+   * sequence by repeated global coarsening of the provided triangulation.
+   *
+   * @note For convenience, a reference to the input triangulation is stored in
+   *   the last entry of the return vector.
```
marcfehling, line 50: Also note that this has only been implemented for `p::d::Triangulation` yet.

marcfehling (changes requested):
All that workaround because we delete the copy constructor :-)

It's a pity that we need to take a detour over the `save/load` functions of `p4est` to store/recover the refinement history. I don't think it currently offers a direct way to save the history.

However, I think we could deepcopy the underlying `p4est` object from the fine triangulation, and then copy it to the coarse triangulations and call `copy_local_forest_to_triangulation()` to recover the refinement history. Maybe we could realize this in an constructor like `p::d::Triangulation(const ::internal::p4est::types< dim >::forest *)`. But let's keep this for now and we can address this in a future PR.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1620,6 +1619,81 @@ namespace internal
...
+    const unsigned int max_level = fine_triangulation->n_global_levels() - 1;
+
+    const auto coarse_mesh_description =
+      GridTools::get_coarse_mesh_description(*fine_triangulation);
+
+    coarse_grid_triangulations.resize(fine_triangulation->n_global_levels());
+
+    coarse_grid_triangulations[max_level].reset(
+      const_cast<Triangulation<dim, spacedim> *>(&fine_triangulation_in),
+      [](auto &) {});
```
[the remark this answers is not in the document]
peterrum: Nope. I am disabling here the deleter, since the destructor  is called somewhere else.

peterrum:
> All that workaround because we delete the copy constructor :-)

Yes. It would be nice if you don't have to serialize to a file.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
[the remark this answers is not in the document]
kronbichler: But you should add a comment what the second argument means.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1620,6 +1622,82 @@ namespace internal
...
+
+    // create coarse meshes
+    for (unsigned int l = max_level - 1; l != static_cast<unsigned int>(-1);
+         --l)
+      {
+        // store finer mesh to file
+        dynamic_cast<
+          const parallel::distributed::Triangulation<dim, spacedim> *>(
+          coarse_grid_triangulations[l + 1].get())
+          ->save("mesh");
```
kronbichler, line 1665: Are you planning to still work on this (sorry if I missed a discussion on this elsewhere)? It feels unsafe to write to the current directory without the user giving a chance to specify the location. Think of a supercomputer file system where your executable might sit in a place where you don't want to write binary output. I would be much happier if we could write to RAM, but I realize it is not possible with the current infrastructure because we need to synchronize between the MPI ranks. So I guess we will need a somewhat adapted functionality here. Also, it seems we leave a new file around, which is surprising as a user never requested one, do we?

peterrum: Yes, we have plans. See https://github.com/dealii/dealii/pull/11707#pullrequestreview-585109577.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1620,6 +1622,82 @@ namespace internal
...
+
+    coarse_grid_triangulations.resize(fine_triangulation->n_global_levels());
+
+    coarse_grid_triangulations[max_level].reset(
+      const_cast<Triangulation<dim, spacedim> *>(&fine_triangulation_in),
+      [](auto &) {});
+
+    // create coarse meshes
+    for (unsigned int l = max_level - 1; l != static_cast<unsigned int>(-1);
+         --l)
```
kronbichler, line 1659: For what it's worth, elsewhere in the library we often use the loops constructed as
```suggestion
    for (unsigned int l = max_level; l > 0; )
      {
        --l;
```
Not sure if you have strong objections against this one; I prefer it because not all progammers are familiar with the wrap-around behavior of unsigned int specified in the C++ standard.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1620,6 +1622,84 @@ namespace internal
...
+
+    coarse_grid_triangulations[max_level].reset(
+      const_cast<Triangulation<dim, spacedim> *>(&fine_triangulation_in),
+      [](auto &) {
+        // empty deleter, since fine_triangulation_in is an external field
+        // and its destructor is called somewhere else
+      });
+
+    // create coarse meshes
+    for (unsigned int l = max_level; l > 0; --l)
```
peterrum, line 1661: @kronbichler I went with this (and have shifted the indices where they are used). Any objections here.

kronbichler (approved):

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
peterrum: @kronbichler Went faster than expected: #11715

kronbichler:
can you rebase this one on master to resolve the conflict from #11706?

peterrum:
Done!

PR #11713, opened 2021-02-08
Also move get_hypercube/simplex() into namespace ReferenceCells.

bangerth:
Next step for #11544.

/rebuild

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -81,24 +81,6 @@ class ReferenceCell
 {
 public:
   /**
-   * Return the correct simplex reference cell type for the given dimension
-   * `dim`. Depending on the template argument `dim`, this function returns a
-   * reference to either Vertex, Triangle, or Tetrahedron.
-   */
-  template <int dim>
-  static constexpr const ReferenceCell &
-  get_simplex();
-
-  /**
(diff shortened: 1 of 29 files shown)


peterrum (approved):

include/deal.II/grid/reference_cell.h
```diff
@@ -297,12 +280,6 @@ class ReferenceCell
 
 
 
-inline constexpr ReferenceCell::ReferenceCell()
-  : ReferenceCell(static_cast<std::uint8_t>(-1))
-{}
```
bangerth, line 302: Easy to miss in the patch: I moved this function...

include/deal.II/grid/reference_cell.h
```diff
@@ -381,10 +358,35 @@ namespace ReferenceCells
...
+  constexpr const ReferenceCell &
+  get_hypercube();
 } // namespace ReferenceCells
 
 
 
+inline DEAL_II_CONSTEXPR
+ReferenceCell::ReferenceCell()
+  : ReferenceCell(ReferenceCells::Invalid)
+{}
```
bangerth, line 386: ...to here, and also changed the initializer: It's now explicitly `Invalid`, rather than duplicating the place where we say that that corresponds to `-1`. This way, we keep to the way that the only places where one can generate specific objects is in namespace `ReferenceCells`.

PR #11714, opened 2021-02-08
Replace passing by const std::unique_ptr<T>& by const T&

masterleinad:
https://github.com/dealii/dealii/pull/11698#issuecomment-774854191:
> It is correct that in pattern.h, we could replace the std::unique_ptr arguments for the to_string() and to_value() functions to regular const references. I would accept a patch to this end.

This is what this pull request does. I am in favor of doing this but it's not backward-compatible. Let's see what people think.

### DIFF
--- a/include/deal.II/base/parameter_handler.h
+++ b/include/deal.II/base/parameter_handler.h
@@ -2345,7 +2345,7 @@ ParameterHandler::add_parameter(const std::string &          entry,
 
   auto action = [&, pattern_index](const std::string &val) {
     parameter = Patterns::Tools::Convert<ParameterType>::to_value(
-      val, patterns[pattern_index]);
+      val, *patterns[pattern_index]);
   };
   add_action(entry, action);
 }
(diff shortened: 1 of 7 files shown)


include/deal.II/base/parameter_handler.h
```diff
@@ -2334,7 +2334,7 @@ ParameterHandler::add_parameter(const std::string &          entry,
 
   declare_entry(entry,
                 Patterns::Tools::Convert<ParameterType>::to_string(
-                  parameter, pattern.clone()),
+                  parameter, *pattern.clone()),
```
bangerth, line 2337: Here and elsewhere, you don't actually need the `clone()` call any more.
```suggestion
                  parameter, pattern),
```

bangerth (commented):
Let's let @luca-heltai make the call.

bangerth:
@luca-heltai Ping?

luca-heltai:
Ah, sorry. I totally missed this one. 

This is technically an incompatible change. 95% of my codes will break, since I often use the feature of passing a custom pattern. However, I'm in favour of this change. I don't remember why I did it this way in the first place. I suspect that something will break, but I cannot foresee what will....

luca-heltai (approved):

masterleinad:
I added a changelog entry. Ready to be merged from my side.

PR #11718, opened 2021-02-09
Move some functions from internal::ReferenceCell::Base to ReferenceCell

bangerth:
The current set-up of `ReferenceCell` is a bit funny. Some of the functionality is implemented directly as member functions and largely consists of `if-else if-else if` chains covering all of the 8 different possible reference cells.

But then there is a parallel structure whereby there is a hierarchy of classes in `internal::ReferenceCell` that cover all of the 8 different cases; one accesses this hierarchy via the `internal::ReferenceCell::get_cell()` and `internal::ReferenceCell::get_face()` functions, and these then return a reference to one of 8 possible objects that one can query for other pieces of information such as the number of vertices per object. This avoids the longish chain of `if-else if` statements, but at the cost of a virtual function dispatch; also, getting the right reference of course *also* requires an `if-else if` chain.

I *suspect* that the intention was that one generates this kind of object reference only once and then the cost is simply the virtual function call. That works, except in most places I looked at, no caching is actually happening and we just do both of the steps every time, which is not more efficient. In practice, I would be very surprised if one could measure the cost of the `if-else if` statements -- but if it turns out that we can, we can always either reorder the different `if-else` branches, or switch to a table lookup. In any case, the split across two places also makes it quite hard to find from the documentation which information is where.

This patch does the first step in re-unifying the two classes. It moves the `n_vertices/lines/faces()` functions from `internal::ReferenceCell::Base` to `ReferenceCell` in the first commit. The second commit also moves the corresponding `vertex/line/face_indices()` functions. The last one does `standard_vertex_to_face_and_vertex_index()` and `standard_line_to_face_and_line_index()`. In all cases, I don't remove the original functions from their current location but just make these functions `private`, which allows me to find all of the places where we used them and convert these places. I will remove the functions in a final step, but some of the functions I haven't moved yet require the ones I'm touching here and so I will leave them `private` for now until I really can remove them.

/rebuild

### DIFF
--- a/include/deal.II/fe/fe_tools.templates.h
+++ b/include/deal.II/fe/fe_tools.templates.h
@@ -282,11 +282,9 @@ namespace FETools
       // 3. Quads
       for (unsigned int quad_number = 0;
            quad_number <
-           (dim == 2 ? 1 :
-                       (dim == 3 ? dealii::internal::ReferenceCell::get_cell(
-                                     fes.front()->reference_cell())
-                                     .n_faces() :
-                                   0));
+           (dim == 2 ?
+              1 :
+              (dim == 3 ? fes.front()->reference_cell().n_faces() : 0));
            ++quad_number)
         {
           for (unsigned int base = 0; base < fes.size(); ++base)
@@ -499,11 +497,9 @@ namespace FETools
       // 3. Quads
       for (unsigned int quad_number = 0;
            quad_number <
-           (dim == 2 ? 1 :
-                       (dim == 3 ? dealii::internal::ReferenceCell::get_cell(
-                                     fes.front()->reference_cell())
-                                     .n_faces() :
-                                   0));
+           (dim == 2 ?
+              1 :
+              (dim == 3 ? fes.front()->reference_cell().n_faces() : 0));
            ++quad_number)
         {
           unsigned int comp_start = 0;
@@ -761,11 +757,7 @@ namespace FETools
       // 3. Quads
       for (unsigned int quad_number = 0;
            quad_number <
-           (dim == 2 ? 1 :
-                       (dim == 3 ? dealii::internal::ReferenceCell::get_cell(
-                                     fe.reference_cell())
-                                     .n_faces() :
-                                   0));
+           (dim == 2 ? 1 : (dim == 3 ? fe.reference_cell().n_faces() : 0));
            ++quad_number)
         {
           unsigned int comp_start = 0;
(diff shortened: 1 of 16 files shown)


bangerth:
Related to #11544.

peterrum (approved):
Looks good. May I ask to rebase, I think there is clash with a PR of mine that has been merged today.

bangerth:
I rebased an hour ago, and github doesn't show me any conflicts.

bangerth:
But let me compile a rebased version just to make sure that things still work.

peterrum (approved):
Let's merge once the testers are happy!

peterrum:
> But let me compile a rebased version just to make sure that things still work.

Oh. PR #11372 has been only approved and not merged yet...

bangerth:
I would merge that one, but I'm not sure whether @kronbichler's question there ("Looks good, if we can have a good answer to the question of general quadrature formulas.") has been answered to his satisfaction.

peterrum:
> "Looks good, if we can have a good answer to the question of general quadrature formulas."

This is an open issue (for months) and has not been introduced in this PR. So I think you are safe to merge ;)

masterleinad:
There are conflicting files.

bangerth:
I rebased and fixed the merge conflict.

kronbichler (approved):

masterleinad:
```
/home/runner/work/dealii/dealii/include/deal.II/matrix_free/shape_info.templates.h:347:44: error: ‘virtual unsigned int dealii::internal::ReferenceCell::Base::n_vertices() const’ is private within this context
                   n_max_vertices = std::max(n_max_vertices,
                                    ~~~~~~~~^~~~~~~~~~~~~~~~
                                             internal::ReferenceCell::get_face(
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                               reference_cell_type, face_no)
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                               .n_vertices());
                                               ~~~~~~~~~~~~~~
In file included from /home/runner/work/dealii/dealii/include/deal.II/base/qprojector.h:25:0,
                 from /home/runner/work/dealii/dealii/include/deal.II/matrix_free/shape_info.templates.h:25,
                 from /home/runner/work/dealii/dealii/source/matrix_free/shape_info.cc:21:
/home/runner/work/dealii/dealii/include/deal.II/grid/reference_cell.h:1264:7: note: declared private here
       n_vertices() const
       ^~~~~~~~~~
```

bangerth:
Let's try this again. I had it all working until that rebase... ;-)

bangerth:
Seems to work now.

peterrum (approved):
Thanks!

PR #11720, opened 2021-02-10
Allow conversion from C-style array to ArrayView.

bangerth:
/rebuild


tamiko (approved):
Looks good.

masterleinad (approved):

include/deal.II/base/array_view.h
```diff
@@ -169,6 +170,23 @@ class ArrayView
...
   ArrayView(std::vector<typename std::remove_cv<value_type>::type> &vector);
 
+  /**
+   * A constructor that automatically creates a view for a given C-style array.
+   * This constructor can be used as follows:
+   * @code
+   *   ArrayView<int>
+   *   get_data_table ()
+   *   {
+   *     const int my_data[7] = { 1, 1, 2, 3, 5, 8, 13 };
```
peterrum, line 180: Shouldn't this be static as well?

bangerth: Yes, and the return type also needs to be a `const int`. Thanks for spotting the mistake! Patch coming.

PR #11722, opened 2021-02-10
Add recv and send data to particle cache

blaisb:
This adds the recv and send data to the particle cache in order
to prevent dynamic reallocation of this vector, whose size is not
changing anyway. This can significantly (10%) increase performance
in some of our applications

### DIFF
--- a/include/deal.II/particles/partitioner.h
+++ b/include/deal.II/particles/partitioner.h
@@ -99,14 +99,18 @@ namespace Particles
         ghost_particles_iterators;
 
       /**
-       * Vector of char that is used to organize the particle information to be
-       * sent to other processors
+       * Temporary storage that holds the data of the particles to be sent
+       * to other processors to update the ghost particles information
+       * in update_ghost_particles()
+       * send_recv_particles_properties_and_location()
        */
       std::vector<char> send_data;
 
       /**
-       * Vector of char that is used to organize the particle information to
-       * received from other processors
+       * Temporary storage that holds the data of the particles to receive
+       * the ghost particles information from other processors in
+       * in update_ghost_particles()
+       * send_recv_particles_properties_and_location()
        */
       std::vector<char> recv_data;
     };
--- a/include/deal.II/particles/partitioner.h
+++ b/include/deal.II/particles/partitioner.h
@@ -97,6 +97,18 @@ namespace Particles
       std::vector<typename std::multimap<internal::LevelInd,
                                          Particle<dim, spacedim>>::iterator>
         ghost_particles_iterators;
+
+      /**
+       * Vector of char that is used to organize the particle information to be
+       * sent to other processors
+       */
+      std::vector<char> send_data;
+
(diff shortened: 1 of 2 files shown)


include/deal.II/particles/partitioner.h
```diff
@@ -97,6 +97,18 @@ namespace Particles
...
+       * Vector of char that is used to organize the particle information to be
+       * sent to other processors
+       */
+      std::vector<char> send_data;
+
+      /**
+       * Vector of char that is used to organize the particle information to
+       * received from other processors
+       */
+      std::vector<char> recv_data;
```
peterrum, line 111: I would suggest to adjust the comment similar to:

https://github.com/dealii/dealii/blob/79b124c53968490d700fa798d0a794756bc5f06a/include/deal.II/lac/la_parallel_vector.h#L1334-L1339

peterrum (commented):
Looks reasonable if we cannot overlap multiple communications of a `ParticleHandler`.

In the case of `DoFHandler` and `LindearAlgebra::distributed::Vector`, the data is clearly separated from the `DoFHandler` so that we can have multiple vectors each able to communicate independently. 

I am not 100% familiar with the internal data structures of `ParticleHandler`, but I guess there is no such separation of concerns (is there?) and it is not planed. If this is the case, this change should be safe.

kronbichler (approved):
This looks good. I guess the initialization is now done once and does not need to be repeated over the lifetime of the cache, right? In other words, whenever the underlying data sizes etc. change, we are going to set up a completely new object again anyway.

blaisb:
> This looks good. I guess the initialization is now done once and does not need to be repeated over the lifetime of the cache, right? In other words, whenever the underlying data sizes etc. change, we are going to set up a completely new object again anyway.

This is exactly it. The initialization is done once per cache initialization (which is once per exchange_ghosts).
Anyway, this vector was built locally in update_ghosts using information from the cache. It just felt redundant to cache 80% of the content, but to keep a dynamic allocation for the data exchanged. This introduces a slight "permanent" memory cost (number of ghost * number of properties), but in turns saves significant allocation. In our case, it reduces the number of time this vector if allocated by a factor 1000 to 10000 in a single simulation. Which, for large data set and in parallel, has a very significant impact since this can be a very big vector.

include/deal.II/particles/partitioner.h
blaisb: Done in latest commit :). Thanks for the heads up !

kronbichler (approved):

peterrum (approved):
Thx!

blaisb:
> Thx!



> Thx!

Seems like the test that breaks is unrelated. Do you want me to rebase or you will go ahead and merge anyway?

peterrum:
Yes it is unrelated. I will wait that all tests finish and then merge.

peterrum:
I am merging since the failing test unrelated.

PR #11731, opened 2021-02-11
Fix GridIn::read_exodusii

peterrum:
introduced by #11718.


bangerth (approved):

bangerth:
The failing test is the one that keeps timing out. Since the CI didn't catch the introduction of the original bug, it's not going to catch anything else. So I'll merge.

PR #11733, opened 2021-02-11
::SolutionTransfer with p::s::Triangulation and artificial cells.

### LINKED ISSUE #11668: SolutionTransfer does not work with p::s::Triangulation when allow_artificial_cells=true
When using `::SolutionTransfer` with `p::s::Triangulation` with `allow_artificial_cells=true`, it will run into assertion failure
```
An error occurred in line <2579> of file </.../include/deal.II/dofs/dof_accessor.templates.h> in function
    void dealii::DoFCellAccessor<3, 3, false>::get_dof_indices(std::vector<types::global_dof_index> &) const [dim = 3, spacedim = 3, lda = false]
The violated condition was: 
    this->is_artificial() == false
Additional information: 
    Can't ask for DoF indices on artificial cells.
```
Is this a bug or known limitation?

Here is a MWE
```

#include <deal.II/distributed/shared_tria.h>

#include <deal.II/lac/vector.h>

#include <deal.II/numerics/solution_transfer.h>

int main(int argc, char** argv) {
  dealii::Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);

  dealii::parallel::shared::Triangulation<3> tria(
      MPI_COMM_WORLD, dealii::Triangulation<3>::none, /* allow_artificial_cells */ true); 
  dealii::GridGenerator::subdivided_hyper_cube(tria, 4);

  dealii::FE_DGQ<3> fe(0);

  dealii::DoFHandler<3> dh(tria);
  dh.distribute_dofs(fe);

  dealii::SolutionTransfer<3> soltrans(dh);

  dealii::Vector<double> sol_old(dh.n_dofs());
  sol_old = 0.0;

  tria.set_all_refine_flags();

  tria.prepare_coarsening_and_refinement();
  soltrans.prepare_for_coarsening_and_refinement(sol_old);

  tria.execute_coarsening_and_refinement();

  dh.distribute_dofs(fe);
  dealii::Vector<double> sol_new(dh.n_dofs());

  soltrans.interpolate(sol_old, sol_new);

  return 0;
}

```

marcfehling:
Closes #11668.

The provided `Vector` needs to contain valid data on all cells (not only the locally relevant ones).

Further, we use the same kind of code to set and recover the true subdomain ids of `p::s::Triangulations` in multiple locations in the library. We might consider to move it to a separate function.

@adamqc Would you verify that this patch will solve the issue in your use case? I would like to have your okay before merging.

### DIFF
--- a/source/numerics/solution_transfer.cc
+++ b/source/numerics/solution_transfer.cc
@@ -15,6 +15,7 @@
 
 #include <deal.II/base/memory_consumption.h>
 
+#include <deal.II/distributed/shared_tria.h>
 #include <deal.II/distributed/tria.h>
 
 #include <deal.II/dofs/dof_accessor.h>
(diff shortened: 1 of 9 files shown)


adamqc:
I can confirm that it works now. Thanks.

source/numerics/solution_transfer.cc
```diff
@@ -92,6 +93,32 @@ SolutionTransfer<dim, VectorType, DoFHandlerType>::prepare_for_pure_refinement()
 
   clear();
 
+  // We need to access dof indices on the entire domain. For
+  // shared::Triangulations, ownership of cells might change. If they allow the
+  // use of artificial cells, we need to restore the true cell owners
+  // temporarily. We save the current set of subdomain ids, set subdomain ids to
+  // the "true" owner of each cell, and later restore these flags.
+  std::vector<types::subdomain_id> saved_subdomain_ids;
```
peterrum, line 101: I must admit, I would prefer a toggle method in `p:d:T`. We already have these lines of codes a couple of times in the library. I now we would add additional four instances.

tests/mpi/solution_transfer_07.cc
```diff
@@ -0,0 +1,89 @@
...
+
+  FE_Q<dim>       fe(1);
+  DoFHandler<dim> dh(tria);
+  dh.distribute_dofs(fe);
+
+  tria.begin_active()->set_refine_flag();
+  tria.prepare_coarsening_and_refinement();
+
+  Vector<double> sol_old(dh.n_dofs());
+  sol_old = 1.;
```
peterrum, line 57: You could enhance the test by having a second `p:s:T` (with no artificial) cells. Then you could interpolate any function on `sol_old`, do the solution transfer on both triangulations and assert the equality of the results.

peterrum (approved):
> I must admit, I would prefer a toggle method in p:d:T. We already have these lines of codes a couple of times in the library. I now we would add additional four instances.

Let's do this in a follow-up PR!

marcfehling:
I've added the requested tests. I'll work on a toggle method next.

Ready for final review.

PR #11734, opened 2021-02-11
Move some more functions from internal::ReferenceCell::* classes to ReferenceCell

bangerth:
More stuff related to #11544.

/rebuild

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -1217,7 +1217,7 @@ ReferenceCell::standard_to_real_face_vertex(
     }
   else if (*this == ReferenceCells::Pyramid)
     {
-      if (face == 0) // Quad
+      if (face == 0) // The quadrilateral face
         {
           return GeometryInfo<3>::standard_to_real_face_vertex(
             vertex,
@@ -1225,7 +1225,7 @@ ReferenceCell::standard_to_real_face_vertex(
             Utilities::get_bit(face_orientation, 2),
             Utilities::get_bit(face_orientation, 1));
         }
-      else // Tri
+      else // One of the triangular faces
         {
           static const unsigned int table[6][3] = {
             {0, 2, 1}, {0, 1, 2}, {2, 1, 0}, {1, 2, 0}, {1, 0, 2}, {2, 0, 1}};
@@ -1235,7 +1235,7 @@ ReferenceCell::standard_to_real_face_vertex(
     }
   else if (*this == ReferenceCells::Wedge)
     {
-      if (face > 1) // QUAD
+      if (face > 1) // One of the quadrilateral faces
         {
           return GeometryInfo<3>::standard_to_real_face_vertex(
             vertex,
@@ -1243,7 +1243,7 @@ ReferenceCell::standard_to_real_face_vertex(
             Utilities::get_bit(face_orientation, 2),
             Utilities::get_bit(face_orientation, 1));
         }
-      else // TRI
+      else // One of the triangular faces
         {
           static const unsigned int table[6][3] = {
             {0, 2, 1}, {0, 1, 2}, {2, 1, 0}, {1, 2, 0}, {1, 0, 2}, {2, 0, 1}};
@@ -1300,7 +1300,7 @@ ReferenceCell::standard_to_real_face_line(
     }
   else if (*this == ReferenceCells::Pyramid)
     {
-      if (face == 0) // QUAD
+      if (face == 0) // The quadrilateral face
         {
           return GeometryInfo<3>::standard_to_real_face_line(
             line,
@@ -1308,7 +1308,7 @@ ReferenceCell::standard_to_real_face_line(
             Utilities::get_bit(face_orientation, 2),
             Utilities::get_bit(face_orientation, 1));
         }
-      else // TRI
+      else // One of the triangular faces
         {
           static const unsigned int table[6][3] = {
             {2, 1, 0}, {0, 1, 2}, {1, 0, 2}, {1, 2, 0}, {0, 2, 1}, {2, 0, 1}};
@@ -1318,7 +1318,7 @@ ReferenceCell::standard_to_real_face_line(
     }
   else if (*this == ReferenceCells::Wedge)
     {
-      if (face > 1) // QUAD
+      if (face > 1) // One of the quadrilateral faces
         {
           return GeometryInfo<3>::standard_to_real_face_line(
             line,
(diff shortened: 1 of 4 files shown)


include/deal.II/grid/reference_cell.h
```diff
@@ -793,6 +861,53 @@ ReferenceCell::face_reference_cell(const unsigned int face_no) const
...
+    {
+      Assert(false, ExcNotImplemented());
+    }
+  else if (*this == ReferenceCells::Line)
+    {
+      Assert(false, ExcNotImplemented());
+    }
+  else if (*this == ReferenceCells::Triangle)
+    {
+      static constexpr unsigned int subcells[3][2] = {{0, 1}, {1, 2}, {2, 0}};
```
peterrum, line 880: Let's wait here ;)

include/deal.II/grid/reference_cell.h
```diff
@@ -793,6 +861,53 @@ ReferenceCell::face_reference_cell(const unsigned int face_no) const
...
+
+      return subcells[face][subface];
+    }
+  else if (*this == ReferenceCells::Quadrilateral)
+    {
+      Assert(false, ExcNotImplemented());
```
peterrum, line 886: Why is this empty? Shouldn't it be possible to call `GeometryInfo`?

include/deal.II/grid/reference_cell.h
```diff
@@ -919,6 +1032,323 @@ ReferenceCell::standard_line_to_face_and_line_index(
...
+        Utilities::get_bit(face_orientation, 0),
+        Utilities::get_bit(face_orientation, 2),
+        Utilities::get_bit(face_orientation, 1));
+    }
+  else if (*this == ReferenceCells::Tetrahedron)
+    {
+      const static unsigned int table[4][3] = {{0, 1, 2},
+                                               {0, 3, 4},
+                                               {2, 5, 3},
+                                               {1, 4, 5}};
```
peterrum, line 1074: Let's wait here until we have a decision on C-style arrays.

include/deal.II/grid/reference_cell.h
```diff
@@ -919,6 +1032,323 @@ ReferenceCell::standard_line_to_face_and_line_index(
...
+                                               {1, 0, 3},
+                                               {0, 2, 3},
+                                               {2, 1, 3}};
+
+      return table[face][standard_to_real_face_vertex(
+        vertex, face, face_orientation)];
+    }
+  else if (*this == ReferenceCells::Pyramid)
+    {
+      constexpr auto            X           = numbers::invalid_unsigned_int;
```
peterrum, line 1151: We could move the x to the top of the function.

include/deal.II/grid/reference_cell.h
```diff
@@ -919,6 +1032,323 @@ ReferenceCell::standard_line_to_face_and_line_index(
...
+  else if (*this == ReferenceCells::Tetrahedron)
+    {
+      static const unsigned int table[6][3] = {
+        {0, 2, 1}, {0, 1, 2}, {2, 1, 0}, {1, 2, 0}, {1, 0, 2}, {2, 0, 1}};
+
+      return table[face_orientation][vertex];
+    }
+  else if (*this == ReferenceCells::Pyramid)
+    {
+      if (face == 0) // Quad
```
peterrum, line 1220: I guess we should not abbreviate these anymore.

include/deal.II/grid/reference_cell.h
bangerth: I just copied together what was in the `internal::ReferenceCell::Base` class hierarchy and this was the only case that was implemented.

In fact, one can have misgivings about the way this function is currently implemented anyway. In `GeometryInfo`, the function takes orientation flags, whereas here it does not. That means that the triangle case above is probably wrong in some cases, and at least that the function will be wrong if it is ever called for quads because it doesn't have the necessary information to call the respective function in `GeometryInfo`. I don't know what to do here, particularly because I don't know whether the orientation flags mean anything for triangles and tets at this point. The function here is only ever called for triangles at the moment.

include/deal.II/grid/reference_cell.h
bangerth: I think it's fine as is -- it makes it easy to read what the table says without having to scroll to the top of the page. (I actually kind of like your approach with the `X` as an invalid placeholder in tables.)

include/deal.II/grid/reference_cell.h
bangerth: Good point. See the last commit now.

peterrum: Thanks!

include/deal.II/grid/reference_cell.h
peterrum: > In GeometryInfo, the function takes orientation flags, whereas here it does not.

This is not good and should be fixed in a follow-up PR! The orientation flags are relevant both for triangles and tets (see all the other functions).

bangerth:
@peterrum  I reverted to the use of `std::array` in case you think that this allows us to move forward.

include/deal.II/grid/reference_cell.h
bangerth: OK, I will tackle this as a follow-up then.

peterrum (approved):
>  reverted to the use of std::array in case you think that this allows us to move forward.

Thanks!

bangerth:
It tests correctly. Is there anything else you'd like me to do?

include/deal.II/grid/reference_cell.h
```diff
@@ -919,6 +1033,321 @@ ReferenceCell::standard_line_to_face_and_line_index(
...
+        Utilities::get_bit(face_orientation, 0),
+        Utilities::get_bit(face_orientation, 2),
+        Utilities::get_bit(face_orientation, 1));
+    }
+  else if (*this == ReferenceCells::Tetrahedron)
+    {
+      static const unsigned int table[4][3] = {{0, 1, 2},
+                                               {1, 0, 3},
+                                               {0, 2, 3},
+                                               {2, 1, 3}};
```
peterrum, line 1143: I would call this also a C-style array. Let's wait until we have a decision how to proceed regarding these.

bangerth:
I converted all C-style arrays in the new functions that I could find to `std::array`, as per the resolution in #11729 .

bangerth:
Anything else anyone wants me to change here? Let's tackle @tamiko 's `ndarray` as a follow-up (see #11743).

PR #11737, opened 2021-02-12
Add some missing header files for sundials_wrapper.h

masterleinad:
We use `std::funtion` and `std::unique_ptr` here.

### DIFF
--- a/include/deal.II/sundials/sunlinsol_wrapper.h
+++ b/include/deal.II/sundials/sunlinsol_wrapper.h
@@ -23,6 +23,9 @@
 
 #    include <sundials/sundials_linearsolver.h>
 
+#    include <functional>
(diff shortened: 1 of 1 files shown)


peterrum (approved):

bangerth (approved):

bangerth:
There is really not very much that can go wrong if you add include files, but the testers have already started, so let's wait with merging till they're finished... :-)

masterleinad:
> There is really not very much that can go wrong if you add include files, but the testers have already started, so let's wait with merging till they're finished... :-)

Also, we don't build with `SUNDIALS` in the CI at all. :slightly_smiling_face:

kronbichler (approved):

PR #11741, opened 2021-02-13
New class TemporarilyRestoreSubdomainIds.

marcfehling:
~Blocked by #11733.~ (Edit: Has been merged.)

We are using true subdomain ids for `p::s::T` in multiple locations of the library. Instead of copying the same code snippet for future uses, this is an attempt to provide a general interface to set true subdomain ids and recover the default ones.

@peterrum @bangerth -- I'm open for naming suggestions.

### DIFF
--- a/include/deal.II/distributed/shared_tria.h
+++ b/include/deal.II/distributed/shared_tria.h
@@ -460,6 +460,73 @@ namespace parallel
 #endif
 } // namespace parallel
 
+
+namespace internal
+{
+  namespace parallel
+  {
+    namespace shared
+    {
+      /**
+       * This class temporarily modifies the subdomain ID of all active cells to
+       * their respective "true" owner.
+       *
+       * The modification only happens on parallel::shared::Triangulation
+       * objects with artificial cells, and persists for the lifetime of an
+       * instantiation of this class.
+       *
+       * The TemporarilyRestoreSubdomainIds class should only be used for
+       * temporary read-only purposes. For example, whenever your implementation
+       * requires to treat artificial cells temporarily as locally relevant to
+       * access their dof indices.
+       *
+       * This class has effect only if artificial cells are allowed. Without
+       * artificial cells, the current subdomain IDs already correspond to the
+       * true subdomain IDs. See the @ref GlossArtificialCell "glossary"
+       * for more information about artificial cells.
+       */
+      template <int dim, int spacedim>
+      class TemporarilyRestoreSubdomainIds : public Subscriptor
+      {
+      public:
+        /**
+         * Constructor.
+         *
+         * Stores the subdomain ID of all active cells if the provided
+         * Triangulation is of type parallel::shared::Triangulation.
+         *
+         * Replaces them by their true subdomain ID equivalent.
+         */
+        TemporarilyRestoreSubdomainIds(
+          const Triangulation<dim, spacedim> &tria);
+
+        /**
+         * Destructor.
+         *
+         * Returns the subdomain ID of all active cells on the
+         * parallel::shared::Triangulation into their previous state.
+         */
+        ~TemporarilyRestoreSubdomainIds();
+
+      private:
+        /**
+         * The modified parallel::shared::Triangulation.
+         */
+        SmartPointer<
+          const dealii::parallel::shared::Triangulation<dim, spacedim>>
+          shared_tria;
+
+        /**
+         * A vector that temporarily stores the subdomain IDs on all active
+         * cells before they have been modified on the
+         * parallel::shared::Triangulation.
+         */
+        std::vector<unsigned int> saved_subdomain_ids;
+      };
+    } // namespace shared
+  }   // namespace parallel
+} // namespace internal
+
 DEAL_II_NAMESPACE_CLOSE
 
 #endif
--- a/source/distributed/shared_tria.cc
+++ b/source/distributed/shared_tria.cc
@@ -468,6 +468,59 @@ namespace parallel
 #endif
 
 
+
+namespace internal
+{
+  namespace parallel
+  {
+    namespace shared
+    {
+      template <int dim, int spacedim>
+      TemporarilyRestoreSubdomainIds<dim, spacedim>::
+        TemporarilyRestoreSubdomainIds(const Triangulation<dim, spacedim> &tria)
+        : shared_tria(
+            dynamic_cast<
+              const dealii::parallel::shared::Triangulation<dim, spacedim> *>(
+              &tria))
+      {
+        if (shared_tria && shared_tria->with_artificial_cells())
+          {
+            // Save the current set of subdomain IDs, and set subdomain IDs
+            // to the "true" owner of each cell.
+            const std::vector<types::subdomain_id> &true_subdomain_ids =
+              shared_tria->get_true_subdomain_ids_of_cells();
+
+            saved_subdomain_ids.resize(shared_tria->n_active_cells());
+            for (const auto &cell : shared_tria->active_cell_iterators())
+              {
+                const unsigned int index   = cell->active_cell_index();
+                saved_subdomain_ids[index] = cell->subdomain_id();
+                cell->set_subdomain_id(true_subdomain_ids[index]);
+              }
+          }
+      }
+
+
+
+      template <int dim, int spacedim>
+      TemporarilyRestoreSubdomainIds<dim, spacedim>::
+        ~TemporarilyRestoreSubdomainIds()
+      {
+        if (shared_tria && shared_tria->with_artificial_cells())
+          {
+            // Undo the subdomain modification.
+            for (const auto &cell : shared_tria->active_cell_iterators())
+              {
+                const unsigned int index = cell->active_cell_index();
+                cell->set_subdomain_id(saved_subdomain_ids[index]);
+              }
+          }
+      }
+    } // namespace shared
+  }   // namespace parallel
+} // namespace internal
+
+
 /*-------------- Explicit Instantiations -------------------------------*/
 #include "shared_tria.inst"
 
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -2559,47 +2559,31 @@ DoFHandler<dim, spacedim>::distribute_dofs(
 #endif
     }
 
-  // We would like to enumerate all dofs for shared::Triangulations. If an
-  // underlying shared::Tria allows artificial cells, we need to restore the
-  // true cell owners temporarily. We save the current set of subdomain ids, set
-  // subdomain ids to the "true" owner of each cell, and later restore these
-  // flags.
-  std::vector<types::subdomain_id>                      saved_subdomain_ids;
-  const parallel::shared::Triangulation<dim, spacedim> *shared_tria =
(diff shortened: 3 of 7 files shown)


peterrum:
@marcfehling Could you rebase this branch.

marcfehling:
> @marcfehling Could you rebase this branch.

Rebased. Ready for review and tests.

source/distributed/shared_tria.cc
```diff
@@ -319,6 +323,50 @@ namespace parallel
 
 
 
+    template <int dim, int spacedim>
+    void
+    Triangulation<dim, spacedim>::set_subdomain_mask(
+      const SubdomainMask new_mask)
+    {
+      if (allow_artificial_cells && new_mask != subdomain_mask)
```
peterrum, line 331: I am not sure here. Do we need `allow_artificial_cells` any longer? I think we might be able to switch dynamically with the new function?

source/dofs/dof_handler.cc
```diff
@@ -2564,24 +2564,13 @@ DoFHandler<dim, spacedim>::distribute_dofs(
...
-          cell->set_subdomain_id(true_subdomain_ids[index]);
-        }
-    }
+  parallel::shared::Triangulation<dim, spacedim> *shared_tria =
+    (dynamic_cast<parallel::shared::Triangulation<dim, spacedim> *>(
+      const_cast<dealii::Triangulation<dim, spacedim> *>(
+        &this->get_triangulation())));
+  if (shared_tria != nullptr)
+    shared_tria->set_subdomain_mask(
+      parallel::shared::Triangulation<dim, spacedim>::SubdomainMask::true_ids);
```
peterrum, line 2573: Personally, I would prefer:
```suggestion
  if (auto shared_tria = dynamic_cast<parallel::shared::Triangulation<dim, spacedim> *>(
      const_cast<dealii::Triangulation<dim, spacedim> *>(&this->get_triangulation()))))
    shared_tria->set_subdomain_mask(
      parallel::shared::Triangulation<dim, spacedim>::SubdomainMask::true_ids);
```

source/dofs/dof_handler.cc
```diff
@@ -2597,9 +2586,10 @@ DoFHandler<dim, spacedim>::distribute_dofs(
     internal::DoFHandlerImplementation::Implementation::reserve_space(*this);
 
   // undo the subdomain modification
-  if (shared_tria != nullptr && shared_tria->with_artificial_cells())
-    for (const auto &cell : shared_tria->active_cell_iterators())
-      cell->set_subdomain_id(saved_subdomain_ids[cell->active_cell_index()]);
+  if (shared_tria != nullptr)
+    shared_tria->set_subdomain_mask(
+      parallel::shared::Triangulation<dim,
+                                      spacedim>::SubdomainMask::default_ids);
```
peterrum, line 2592: I am not sure if it wouldn't be better to have two functions? The input argument is annoyingly long and complicated. This remembers of the the arguments passed to the constructor of the triangulation classes.

include/deal.II/distributed/shared_tria.h
```diff
@@ -224,6 +224,27 @@ namespace parallel
...
+      /**
+       * Flags that determine which type of subdomain ID will be returned when
+       * calling CellAccessor::subdomain_id() on active cells.
+       */
+      enum SubdomainMask
+      {
+        /**
+         * Subdomain IDs are valid on all active cells that are locally
+         * relevant. They are invalid for artificial cells, if allowed for this
+         * Triangulation.
```
bangerth, line 236: Can you link to the corresponding glossary entries for these kinds of cells?

include/deal.II/distributed/shared_tria.h
```diff
@@ -310,6 +331,25 @@ namespace parallel
...
+       * indices.
+       *
+       * Modifications to the subdomain ID of active cells by calling
+       * CellAccessor::set_subdomain_id() should always be made when working
+       * with the default mask, i.e., SubdomainMask::default_ids. Otherwise,
+       * changes made to the subdomain IDs could be lost and lead to undefined
+       * behaviour.
+       */
+      void
+      set_subdomain_mask(const SubdomainMask new_mask);
```
bangerth, line 351: How can you restore the flag if there is no function that allows you to get the current state?

source/distributed/shared_tria.cc
```diff
@@ -319,6 +323,50 @@ namespace parallel
...
+            default:
+              Assert(false, ExcNotImplemented());
+              break;
+          }
+
+      subdomain_mask = new_mask;
+    }
+
+
+
```
bangerth, line 369: What happens if you have set the mask, and then do mesh refinement, and then reset it to `default_ids`. Isn't the array the wrong size then? I think you probably want to reset to default upon mesh refinement, and `clear()` the array.

source/dofs/dof_handler.cc
bangerth: `decltype(*shared_tria)::SubdomainMask::default_ids`

include/deal.II/distributed/shared_tria.h
```diff
@@ -224,6 +224,27 @@ namespace parallel
         construct_multigrid_hierarchy = 0x8,
       };
 
+      /**
+       * Flags that determine which type of subdomain ID will be returned when
+       * calling CellAccessor::subdomain_id() on active cells.
+       */
+      enum SubdomainMask
```
bangerth, line 231: below you always write `SubdomainMask::default_ids`, but you don't actually need the `SubdomainMask::` part unless you want to (which I would approve of) change this to
```suggestion
      enum class SubdomainMask
```

bangerth (changes requested):

bangerth:
Separately: Is it correct that conceptually, you need this functionality to temporarily change the state and the revert to the previous state? If so, there is an idiom for that -- take a look at this page, for example: https://www.boost.org/doc/libs/1_64_0/libs/io/doc/ios_state.html

The key is to save the state in the constructor of a class, change the state (could also happen in the constructor), and then restore the state in the destructor. I wonder whether we could do something along these lines here as well.

marcfehling:
> Separately: Is it correct that conceptually, you need this functionality to temporarily change the state and the revert to the previous state? If so, there is an idiom for that -- take a look at this page, for example: https://www.boost.org/doc/libs/1_64_0/libs/io/doc/ios_state.html
> 
> The key is to save the state in the constructor of a class, change the state (could also happen in the constructor), and then restore the state in the destructor. I wonder whether we could do something along these lines here as well.

I like the idea! Something like a class that owns a pointer to a `p:s:T`, provides the subdomain modification upon construction, and undoes it upon destruction. Since we only use it for internal purposes, we can make it part of the `internal` namespace. No changes needed to the `p:s:T` interface.

marcfehling:
Changed the toggle function to a state-saving helper class as suggested by @bangerth. I like it a lot more this way!

I will rebase once you are okay with the changes.

marcfehling:
I am confused about the `windows-CI` error messages:
```
D:\a\dealii\dealii\include\deal.II/base/aligned_vector.h(637,60): error C2039: 'ParallelForInteger': is not a member of 'dealii::internal::parallel' [D:\a\dealii\dealii\build\source\numerics\obj_numerics_debug.vcxproj]
D:\a\dealii\dealii\include\deal.II/distributed/shared_tria.h(466): message : see declaration of 'dealii::internal::parallel' [D:\a\dealii\dealii\build\source\numerics\obj_numerics_debug.vcxproj]
```

Are those naming conflicts? `ParallelForInteger` is a member of the `parallel::internal` namespace...

Do you have a convention for the order in which the `internal` namespace should occur? From my experience, `internal` always followed the `dealii` namespace immediately...

masterleinad:
You can resolve these issues by specifying the complete namespace, i.e. `dealii::parallel::internal::ParallelForInteger`. I don't think we have a convention we follow strictly for nested namespaces.

include/deal.II/distributed/shared_tria.h
```diff
@@ -460,6 +460,71 @@ namespace parallel
...
+       * The modification only happens on parallel::shared::Triangulation
+       * objects with artificial cells, and persists for the lifetime of an
+       * instantiation of this class.
+       *
+       * The SubdomainModifier class should only be used for temporary read-only
+       * purposes. For example, whenever your implementation requires to treat
+       * artificial cells temporarily as locally relevant to access their dof
+       * indices.
+       *
+       * This class will not do anything when artificial cells are allowed.
```
bangerth, line 483: Did you mean "are not allowed"?

include/deal.II/distributed/shared_tria.h
```diff
@@ -460,6 +460,71 @@ namespace parallel
...
+       * In this case, the current subdomain IDs already correspond to the true
+       * subdomain IDs.
+       */
+      template <int dim, int spacedim>
+      class SubdomainModifier : public Subscriptor
```
bangerth, line 488: Let's find a better name for the class. Maybe `TemporarilyRestoreSubdomainIds`.

source/dofs/dof_handler.cc
```diff
@@ -2559,47 +2559,30 @@ DoFHandler<dim, spacedim>::distribute_dofs(
...
-      cell->set_subdomain_id(saved_subdomain_ids[cell->active_cell_index()]);
+  {
+    // We would like to enumerate all dofs for shared::Triangulations. If an
+    // underlying shared::Tria allows artificial cells, we need to restore the
+    // true cell owners temporarily.
+    // We use the SubdomainModifier class for this purpose: we save the current
+    // set of subdomain ids, set subdomain ids to the "true" owner of each cell
+    // upon construction of the SubdomainModifier object, and later restore
+    // these flags when it is destroyed.
+    internal::parallel::shared::SubdomainModifier<dim, spacedim>
```
bangerth, line 2570: make it `const`

source/dofs/dof_handler_policy.cc
```diff
@@ -3026,25 +3026,14 @@ namespace internal
...
-                cell->set_subdomain_id(true_subdomain_ids[index]);
-              }
-          }
+        // If an underlying shared::Tria allows artificial cells, we need to
+        // restore the true cell owners temporarily.
+        // We use the SubdomainModifier class for this purpose: we save the
+        // current set of subdomain ids, set subdomain ids to the "true" owner
+        // of each cell upon construction of the SubdomainModifier object, and
+        // later restore these flags when it is destroyed.
+        internal::parallel::shared::SubdomainModifier<dim, spacedim>
```
bangerth, line 3035: same here and elsewhere

include/deal.II/distributed/shared_tria.h
marcfehling: Yes, you are right. But I just realized this forms a double negation. I'll go with something like `will only operate when they are allowed`.

peterrum:
>  I like it a lot more this way!

I don't really. In my opinion it would be better if p:s:T could change it's state. But anyhow, let's proceed here if you both like this approach. This way it is much better than it used to be.

bangerth (approved):
Good work. I'd like to point out that this idiom works so nicely because it restores the flags even if you have a `break` or `return` in the middle of the block, or if an exception is thrown. It's easy to forget restoring the flag for long code blocks where you might have multiple exit paths, and this kind of class makes sure that that is not an issue.

PR #11742, opened 2021-02-13
Fix CutOffFunctionTensorProduct::gradient.

luca-heltai:
This has been broken for a long time...

### DIFF
--- a/tests/base/function_cutoff_03.cc
+++ b/tests/base/function_cutoff_03.cc
@@ -1,0 +1,128 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2019 - 2020 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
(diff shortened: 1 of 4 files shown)


bangerth:
Changed your mind?

luca-heltai:
No. Didn't solve the issue. I figured that the test was bogus (always printing OK). I'm trying to figure out where the issue is, but it is less trivial than I thought...

tests/base/function_cutoff_03.cc
```diff
@@ -0,0 +1,85 @@
...
+    {
+      integral += values[i] * quad.weight(i);
+      grad_integral += gradients[i] * gradients[i] * quad.weight(i);
+    }
+  if (std::abs(integral - 1.0) > 1e-10)
+    deallog << "Value NOT OK" << std::endl;
+  else
+    deallog << "Value OK" << std::endl;
+  if (std::abs(grad_integral - exact[dim - 1]) > 1e-10)
+    deallog << "Gradient OK" << std::endl;
```
luca-heltai, line 72: This should say "Gradient NOT OK", and even with the current changes, the gradient is not ok.

PR #11744, opened 2021-02-13
Remove simplex folders

peterrum:
For now there is still the folder `include/deal.II/simplex` (with dummy headers) so that I did not have to adjust the tests; but I will remove the folder and update the tests once we are happy with new file names.


peterrum:
Any opinions on the file names? At the moment, no one is adding new and/or working finite elements/quadrature rules, ...

Furthermore, what about making the following renamings as a follow-up PR:
```
Simplex::QGauss  -> QGaussSimplex
Simplex::FE_P   -> FE_Simplex_P
Simplex::FE_DGP -> FE_Simplex_DGP
```
This is how it is done in the case of wedges and pyramids. Then the Simplex namespace could be actually be removed.

drwells (approved):
This move needs to happen eventually - we might as well do it now. I would still prefer to keep all the stuff that is formally experimental in a single namespace, though (so we should keep using `Simplex`).

drwells:
Those renames make sense.

A user pointed out at the workshop that it would make more sense if all the tensor product stuff was in `Hex::` and all the simplex stuff was in `Simplex::` but we can't really move `FE_Q` at this point - hence we need to have this weird naming convention for non-tensor product elements.

peterrum:
Thanks @drwells ! Since this is a very important topic, let's wait for the opinion of at least two further main developers!

peterrum:
The `simlex` include and source folders are removed and all includes, incl. in the tests, are updated. Let's get this in master soon; after that I will remove the `Simplex` immediately so that users don't have to adjust their code twice.

kronbichler (approved):

bangerth:
The name `FE_P` is not used yet. Do you currently name all non-hypercube elements with `FE_<shape>_P`? Then `FE_Simplex_P` makes sense. Otherwise, `FE_P` might be good enough.

include/deal.II/base/polynomials_pyramid.h
```diff
@@ -0,0 +1,166 @@
...
+ *  @ingroup simplex
+ */
+namespace Simplex
+{
+  /**
+   * Polynomials defined on pyramid entities. This class is basis of
+   * Simplex::FE_PyramidP.
+   */
+  template <int dim>
+  class ScalarPyramidPolynomial : public ScalarPolynomialsBase<dim>
```
bangerth, line 39: This may be a separate question from this PR, but the name of the class suggests that it is a *generic* base class for polynomials on pyramids, whereas the constructor takes a `degree` argument that suggests that it is actually meant to implement generic Lagrange elements.

include/deal.II/base/polynomials_pyramid.h
```diff
@@ -0,0 +1,166 @@
...
+    void
+    evaluate(const Point<dim> &           unit_point,
+             std::vector<double> &        values,
+             std::vector<Tensor<1, dim>> &grads,
+             std::vector<Tensor<2, dim>> &grad_grads,
+             std::vector<Tensor<3, dim>> &third_derivatives,
+             std::vector<Tensor<4, dim>> &fourth_derivatives) const override;
+
+    /**
+     * @copydoc ScalarPolynomialsBase::compute_value()
```
bangerth, line 68: You don't actually need this: doxygen does this automatically if you don't provide any doc string at all.

include/deal.II/base/polynomials_pyramid.h
```diff
@@ -0,0 +1,166 @@
...
+     */
+    double
+    compute_value(const unsigned int i, const Point<dim> &p) const override;
+
+    /**
+     * @copydoc ScalarPolynomialsBase::compute_derivative()
+     *
+     * @note Currently, only implemented for first derivative.
```
bangerth, line 76: but here you want it

include/deal.II/base/polynomials_pyramid.h
```diff
@@ -0,0 +1,166 @@
...
+
+  template <int dim>
+  template <int order>
+  Tensor<order, dim>
+  ScalarPyramidPolynomial<dim>::compute_derivative(const unsigned int i,
+                                                   const Point<dim> & p) const
+  {
+    Tensor<order, dim> der;
+
+    AssertDimension(order, 1);
```
bangerth, line 153: That's the wrong assertion. It has nothing to do with the dimension. You just want `Assert(order==1, ExcMessage("..."));`

include/deal.II/base/quadrature_lib.h
```diff
@@ -792,6 +792,96 @@ class QSplit : public Quadrature<dim>
...
+   *
+   * Users specify a number `n_points_1D` as an indication of what polynomial
+   * degree to be integrated exactly, similarly to the number of points in a
+   * QGauss quadrature object, even though the present quadrature formula is not
+   * a tensor product. The given value is translated for n_points_1D=1,2,3,4 to
+   * following number of quadrature points for 2D and 3D:
+   *   - 2D: 1, 3, 7, 15
+   *   - 3D: 1, 4, 10, 35
+   *
+   * For 1D, the quadrature rule degenerates to a `QGauss<1>(n_points_1D)`.
```
bangerth, line 808:
```suggestion
   * For 1D, the quadrature rule degenerates to a `dealii::QGauss<1>(n_points_1D)`.
```

include/deal.II/fe/fe_p_pyramid.h
```diff
@@ -0,0 +1,141 @@
+// ---------------------------------------------------------------------
```
bangerth, line 1: The class name is `FE_PyramidP`, so maybe match that with the file name.

include/deal.II/fe/fe_p_wedge.h
```diff
@@ -0,0 +1,141 @@
+// ---------------------------------------------------------------------
```
bangerth, line 1: same here

bangerth (commented):
Most of this is uncontroversial, but I do have a few questions.

include/deal.II/base/polynomials_pyramid.h
peterrum: Yes. Let's address this in a follow-up PR. What about `ScalarLagrangePyramidPolynomial`?

peterrum:
@bangerth I have addressed your comments. In preparation of `Simplex::FE_P   -> FE_Simplex_P`, I also renamed these files.

kronbichler (approved):

kronbichler:
Let us go ahead here to get the follow-up PR soon.

include/deal.II/base/polynomials_pyramid.h
bangerth: I think it would be useful to keep the words "LagrangePolynomial" together. Maybe `ScalarLagrangePolynomialPyramid` or `...OnPyramids`?

Issue #11727, opened 2021-02-11
Better document consensus algorithms


bangerth:
#11725 reminds me of the fact that I still do not really understand what a "consensus algorithm" is and what the context is in which it is to be used. The problem is that the documentation doesn't actually say what the problem is this set of classes is trying to solve, and there is no worked example that would illustrate it either. That's a shame because I have the feeling that these classes can be central to a lot of work, but if I don't understand what they do (what, I don't need to know how), I can't use them :-(

It would be nice if we could adapt one of the two cases used in the test suite as a motivation, providing some documentation what each part of the code there is doing.

Separately, there are a number of implementations of the general interface, but it is unclear which implementation to use when. It might be nice to provide some basic guidance on this question.

bangerth:
See also #9363.

bangerth:
I fixed some of this in #13082, and it turns out that I have noticed the same issue a second time in #13148 -- we can close this issue in favor of the latter.

PR #11751, opened 2021-02-15
Move LinearAlgebra::CommunicationBase -> Utilities::MPI::CommunicationBase.

drwells:
Things use this outside of linear algebra classes now so we should move it out of LinearAlgebra.

Partly related to #11727.


peterrum (commented):
I would move it to `Utiltities::MPI`.

drwells:
Good idea. I moved it. A bunch of default arguments got really long as a result - I went ahead and replaced them with `{}`.

peterrum (approved):
We also need a change-log entry. But else fine from my side!

kronbichler (approved):

PR #11754, opened 2021-02-15
Fix ReferenceCell usage in read_exodusii_sidesets()

peterrum:
Introduced in #11734.


kronbichler (approved):

bangerth:
Thanks. We should really enable exodus on the CI :-(

Issue #11746, opened 2021-02-14
p::dist::Tria setup very slow on many processors


kronbichler:
On a setup with 18432 MPI ranks, I observed extremely slow initialization times for `parallel::distributed::Triangulation`. In the release 9.2 paper in Fig 1, we recorded a scaling limit of around 0.4 seconds for a `GridGenerator::subdivided_hyper_rectangle` followed by some `tria.refine_global()` steps (I also construct the MG hierarchy). However, I now cannot get below 5 seconds on this setup on a recent version, and times are also pretty bad for larger sizes. Here an example:
```
(32 x 32 x 32) -> 5.37 s
(64 x 32 x 32) -> 5.28 s
(64 x 64 x 32) -> 9.91 s
(64 x 64 x 64) -> 12.3 s
(128 x 64 x 64) -> 11.1 s
(128 x 128 x 64) -> 15.5 s
(128 x 128 x 128) -> 18.0 s
(256 x 128 x 128) -> 14.9 s
(256 x 256 x 128) -> 20.8 s
(256 x 256 x 256) -> 23.8 s
(512 x 256 x 256) -> 16.2 s
(512 x 512 x 256) -> 24.2 s
(512 x 512 x 512) -> 25.5 s
(1,024 x 512 x 512) -> 21.2 s
(1,024 x 1,024 x 512) -> 31.2 s
```
I don't have time right now to investigate in more detail where this regression comes from, but this is something we should look into for the upcoming release.

kronbichler:
Note that this is only for large processor counts, a test computation with 1000 MPI ranks does not show these exceedingly high times, so it is not a node-level problem as far as I can say now.

peterrum:
> In the release 9.2 paper in Fig 1, we recorded a scaling limit of around 0.4 seconds for a GridGenerator::subdivided_hyper_rectangle followed by some tria.refine_global() steps (I also construct the MG hierarchy). However, I now cannot get below 5 seconds on this setup on a recent version, and times are also pretty bad for larger sizes. 

That's bad.We haven't changed anything in the communication patterns since the last release, did we? There have been a lot of changes in `Triangulation::create_triangulation()` and in the accessors but these cannot explain the observed behavior.

kronbichler:
[...]

PR #11755, opened 2021-02-15
Improve complexity of compute_set_union

peterrum:


kronbichler (approved):
Very good, logarithmic complexity in the number of ranks.

kronbichler:
Relates to #11746

bangerth:
Nice! One wished that MPI had variable-length reduction calls so that the details of communication could be handled by MPI and we would only have to implement the reduction operation...

PR #11758, opened 2021-02-15
Adjust minimum grain size of dictionary for consensus algorithm

### LINKED ISSUE #11746: p::dist::Tria setup very slow on many processors

kronbichler:
Fixes #11746. ~~This could be the final fix needed to get the initialization of the triangulation work quickly again also on many MPI ranks, but my jobs to verify this are still in the queue.~~

Update: I verified that initialization times are now similar to the 9.2 release on 40k ranks.

### DIFF
--- a/include/deal.II/base/mpi_compute_index_owner_internal.h
+++ b/include/deal.II/base/mpi_compute_index_owner_internal.h
@@ -157,9 +157,27 @@ namespace Utilities
         struct Dictionary
         {
           /**
-           * The minimum grain size for the ranges.
+           * The minimum grain size for the intervals.
+           *
+           * We choose to limit the smallest size an interval for the
+           * two-stage lookup can have with the following two conflicting
+           * goals in mind: On the one hand, we do not want intervals in the
+           * dictionary to become too short. For uneven distributions of
+           * unknowns (some ranks with several thousands of unknowns, others
+           * with none), the lookup DoFs -> dictionary then involves sending
+           * from one MPI rank to many other MPI ranks holding dictionary
+           * intervals, leading to an exceedingly high number of messages some
+           * ranks have to send. Also, fewer longer intervals are generally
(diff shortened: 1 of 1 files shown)


peterrum (approved):

Rombur (commented):
I think there is a problem with the comment.

kronbichler:
@Rombur I have not seen an exact place on where there was a problem, but I agree that there were two inaccuracies which I hope I have now fixed.

Rombur (approved):

PR #11759, opened 2021-02-15
Introduce Utilities::MPI::all_reduce()

peterrum:
Encapsulate the logic introduced in #11755, into a dedicated reduction function that takes a lambda like:
```cpp
      return allreduce(vec, comm, [](const auto &a, const auto &b) {
        // merge vectors, sort, and eliminate duplicates
        auto result = a;
        result.insert(result.end(), b.begin(), b.end());
        std::sort(result.begin(), result.end());
        result.erase(std::unique(result.begin(), result.end()), result.end());
        return result;
      });
```
in the context of `compute_set_union()`.

Note: I still need to move the anonymous function to a proper function in the `Utilities::MPI` namespace (but the logic will be the same).


include/deal.II/base/mpi.templates.h
```diff
@@ -460,77 +460,88 @@ namespace Utilities
     std::vector<T>
     compute_set_union(const std::vector<T> &vec, const MPI_Comm &comm)
     {
+      const auto allreduce = [](const auto &vec,
+                                const auto  comm,
+                                const auto &process) {
```
bangerth, line 465: These arguments are not self-descriptive. Can you give them actual, explicit types?

include/deal.II/base/mpi.templates.h
```diff
@@ -460,77 +460,88 @@ namespace Utilities
...
 #else
-      (void)comm;
-      return vec;
+        (void)comm;
+        (void)process;
+        return vec;
 #endif
+      };
+
+      return allreduce(vec, comm, [](const auto &a, const auto &b) {
```
bangerth, line 537: same here

include/deal.II/base/mpi.templates.h
```diff
@@ -457,80 +457,103 @@ namespace Utilities
...
+                                               status.MPI_TAG,
+                                               comm,
+                                               MPI_STATUS_IGNORE);
+                  AssertThrowMPI(ierr_3);
+
+                  result = binary_op(result, Utilities::unpack<T>(temp, false));
+                }
+              else if (rank_send == rank) // process sends data
+                {
+                  auto temp = Utilities::pack(result, false);
```
kronbichler, line 511: Please spell out the type
```suggestion
                  const std::vector<char> temp = Utilities::pack(result, false);
```

include/deal.II/base/mpi.h
```diff
@@ -1008,6 +1008,19 @@ namespace Utilities
            const T &          object_to_send,
            const unsigned int root_process = 0);
 
+    /**
+     * A function that combines values @p local_value from all processes
+     * via a user-specified binary operation @p binary_op and distributes the
+     * result back to all processes. As such this function is similar to
+     * MPI_Allreduce but due to the user-specified binary operation also
```
kronbichler, line 1015:
```suggestion
     * MPI_Allreduce (implemented by a global reduction followed by a broadcast step)
     * but due to the user-specified binary operation also
```

kronbichler (approved):

kronbichler:
Would you mind fixing the one suggestion I had in https://github.com/dealii/dealii/pull/11759#discussion_r576842445 then this should be good to go.

peterrum:
@kronbichler Done!

kronbichler (approved):

include/deal.II/base/mpi.templates.h
```diff
@@ -457,80 +457,100 @@ namespace Utilities
...
 
-      // 2) broadcast result
-      int size = result.size();
-      MPI_Bcast(&size, 1, MPI_INT, 0, comm);
-      result.resize(size);
-      MPI_Bcast(
-        result.data(), size, internal::mpi_type_id(vec.data()), 0, comm);
+          // 2) broadcast result
+          std::vector<char> temp = Utilities::pack(result, false);
+          int               size = temp.size();
```
bangerth, line 525:
```suggestion
          const int               size = temp.size();
```

bangerth: Could this possibly overflow? Maybe use the correct data type for the second argument of `MPI_Bcast`.

bangerth (commented):
Nice work here!

bangerth:
Would it be possible to add a simple test somehow?

include/deal.II/base/mpi.templates.h
peterrum: I have changed the type to `unsigned int`. The variable cannot be made const due to the bcast.

include/deal.II/base/mpi.templates.h
```diff
@@ -457,80 +457,100 @@ namespace Utilities
...
-        }
 
-      // 2) broadcast result
-      int size = result.size();
-      MPI_Bcast(&size, 1, MPI_INT, 0, comm);
-      result.resize(size);
-      MPI_Bcast(
-        result.data(), size, internal::mpi_type_id(vec.data()), 0, comm);
+          // 2) broadcast result
+          std::vector<char> temp = Utilities::pack(result, false);
```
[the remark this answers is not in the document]
peterrum: This one cannot be made const since the vector is resized after the needed size is known.

peterrum:
@bangerth I have addressed your comments!

bangerth (approved):

PR #11763, opened 2021-02-16
Add some assertions to the noncontiguous partitioner.

drwells:
There are essentially two changes in this PR:
1. Avoid converting ArrayViews to pointers - this way we get some index checks in debug mode
2. Add more index checks when we access `std::vector`

Some of these checks are probably redundant but they are all cheap.

### DIFF
--- a/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
+++ b/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
@@ -321,15 +321,13 @@ namespace Utilities
           AssertThrowMPI(ierr);
         }
 
-      auto src_iterator = src.begin();
-
       // post send
       for (types::global_dof_index i = 0, k = 0; i < send_ranks.size(); i++)
         {
           // collect data to be send
           for (types::global_dof_index j = send_ptr[i]; j < send_ptr[i + 1];
                j++)
-            buffers[j] = src_iterator[send_indices[k++]];
+            buffers[j] = src[send_indices[k++]];
 
           // send data
           const auto ierr =
@@ -360,8 +358,6 @@ namespace Utilities
       (void)requests;
       Assert(false, ExcNeedsMPI());
 #else
-      auto dst_iterator = dst.begin();
-
       // receive all data packages and copy data from buffers
       for (types::global_dof_index proc = 0; proc < recv_ranks.size(); proc++)
         {
@@ -376,7 +372,7 @@ namespace Utilities
           for (types::global_dof_index j = recv_ptr[i], c = 0;
                j < recv_ptr[i + 1];
                j++)
-            dst_iterator[recv_indices[j]] = buffers[recv_ptr[i] + c++];
+            dst[recv_indices[j]] = buffers[recv_ptr[i] + c++];
         }
 
       // wait that all data packages have been sent
--- a/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
+++ b/include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
@@ -302,12 +302,14 @@ namespace Utilities
       Assert(false, ExcNeedsMPI());
 #else
       AssertIndexRange(communication_channel, 10);
+      AssertDimension(requests.size(), recv_ranks.size() + send_ranks.size());
 
       const auto tag =
         communication_channel +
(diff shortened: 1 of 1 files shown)


include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -310,6 +310,8 @@ namespace Utilities
       // post recv
       for (types::global_dof_index i = 0; i < recv_ranks.size(); i++)
         {
+          AssertIndexRange(i + 1, recv_ptr.size());
+          AssertIndexRange(i + send_ranks.size(), requests.size());
```
masterleinad, line 314: It should be sufficient to only check for the last index, isn't it?
```suggestion
          AssertIndexRange(recv_ranks.size(), recv_ptr.size());
          AssertIndexRange(recv_ranks.size() + send_ranks.size(), requests.size() + 1);
```
and moving out of the loop; the same for the cases below.

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -321,17 +323,25 @@ namespace Utilities
           AssertThrowMPI(ierr);
         }
 
-      auto src_iterator = src.begin();
-
       // post send
       for (types::global_dof_index i = 0, k = 0; i < send_ranks.size(); i++)
         {
           // collect data to be send
+          AssertIndexRange(i + 1, send_ptr.size());
```
masterleinad, line 330:
```suggestion
          AssertIndexRange(send_ranks.size(), send_ptr.size());
```

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -321,17 +323,25 @@ namespace Utilities
...
           for (types::global_dof_index j = send_ptr[i]; j < send_ptr[i + 1];
                j++)
-            buffers[j] = src_iterator[send_indices[k++]];
+            {
+              AssertIndexRange(k, send_indices.size());
```
masterleinad, line 334:
```suggestion
              AssertIndexRange(send[ptr[i + 1], send_indices.size() + 1);
```

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
```diff
@@ -321,17 +323,25 @@ namespace Utilities
...
+              buffers[j] = src[send_indices[k]];
+              ++k;
+            }
 
           // send data
+          Assert((send_ptr[i] < buffers.size()) ||
+                   (send_ptr[i] == buffers.size() &&
+                    send_ptr[i + 1] == send_ptr[i]),
+                 ExcMessage("The input buffer doesn't contain enough entries"));
+          AssertIndexRange(i + 1, send_ptr.size());
```
masterleinad, line 344:
```suggestion
          AssertIndexRange(send_ranks.size(), send_ptr.size());
```
and this is then a duplicate of the one in l.330.

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
drwells: Yes - good point.

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
drwells: This may be true but I don't see why - in particular, I don't see why `send_ptr[0] = 0`, which is necessary for this substitution to work. I'd rather keep the less sophisticated check here so that we don't rely on nonlocal assumptions.

masterleinad: Yes, I agree. I was thinking you were checking for `j` and not for `k`.

drwells:
I made the requested changes.

include/deal.II/base/mpi_noncontiguous_partitioner.templates.h
masterleinad: What about this one?

drwells: Ooops - fixed!

masterleinad (approved):
Looks good to me.

peterrum (approved):

PR #11770, opened 2021-02-18
Move the exodus conversion functions to RefereneCell.

bangerth:
/rebuild

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -430,6 +430,27 @@ public:
    */
 
   /**
+   * @name Translating between deal.II indexing and formats used by other programs
+   * @{
+   */
+
+  /**
+   * Map an ExodusII vertex number to a deal.II vertex number.
+   */
+  unsigned int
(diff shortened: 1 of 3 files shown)


bangerth:
Follow-up to #11544.

drwells (approved):

bangerth:
Ping?

drwells:
Sorry for the delay - do you have exodus installed locally? If not I should compile this so we don't break the CI yet again.

bangerth:
I don't, but the functions in question seem to have been used in only two places, which I fixed.

bangerth:
But a cross check would certainly be appreciated!

drwells:
I'll recompile and run the exodus tests locally and if everything passes I will merge. I'm pretty sure this PR is fine as-is but we have messed this up enough times recently that I want to check.

bangerth:
Yes, we have :-)

drwells:
It works!

bangerth:
I appreciate the cross check!

PR #11774, opened 2021-02-19
Add option not to set zero conditions on boundary of refinement edges.

luca-heltai:
This is required in an application with @mwichro.


luca-heltai:
(sorry, not the right branch)

PR #11775, opened 2021-02-19
Rename the write/update particle functions.

bangerth:
Give these functions a more systematic name. While also adjust to a more widely
used idiom: A function that takes a pointer doesn't return the updated pointer
location via that argument, but instead as an explicit return value. This idiom
is in fact used just a couple of lines below the places where these functions
are currently used in particle_handler.cc.

This is an update to #11199 (@lethe-cfd @blaisb FYI). It doesn't change any functionality, just systematize the way these functions work in the context of similar functions in the library.

/rebuild

### DIFF
--- a/include/deal.II/particles/particle.h
+++ b/include/deal.II/particles/particle.h
@@ -187,17 +187,18 @@ namespace Particles
      * and later de-serializing the properties by calling the appropriate
      * constructor Particle(void *&data, PropertyPool *property_pool = nullptr);
      *
-     * @param [in,out] data The memory location to write particle data
-     * into. This pointer points to the begin of the memory, in which the
-     * data will be written and it will be advanced by the serialized size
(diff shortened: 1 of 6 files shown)


peterrum (approved):

blaisb:
> Give these functions a more systematic name. While also adjust to a more widely
> used idiom: A function that takes a pointer doesn't return the updated pointer
> location via that argument, but instead as an explicit return value. This idiom
> is in fact used just a couple of lines below the places where these functions
> are currently used in particle_handler.cc.
> 
> This is an update to #11199 (@lethe-cfd @blaisb FYI). It doesn't change any functionality, just systematize the way these functions work in the context of similar functions in the library.
> 
> /rebuild

thanks for the heads up!

Rombur (approved):

PR #11780, opened 2021-02-19
Add a version of CUDAWrappers::MatrixFree::reinit() that takes a predicate for cell_loop

Rombur:
With this PR, `cell_loop` can skip cells. This is useful if you want to evaluate the operator on part of the domain.


masterleinad (approved):
Looks reasonable to me.

peterrum (approved):

Issue #8562, opened 2019-08-13
Excise remaining compatibility renumberings


drwells:
Followup to #8556. I need to write all this down so that we remember how to actually fix it later.

AFAICT, `GridReordering::invert_all_cells_of_negative_grid` contains a bug where it assumes that the input uses the old-style numbering (i.e., pre 6.0). However, we only call this function immediately before calling `Triangulation::create_triangulation_compatibility`, which also assumes that the cells use the old-style numbering: i.e., it is called in the middle of a pipeline where we use the old numbering, so it works (in this context) correctly.

I suggest that we deprecate this function and replace it with a new function named `GridTools::invert_cells_with_negative_measure` that assumes we use the standard numbering. We should fix our `GridIn` code to actually use the new numbering everywhere so that we can remove all internal calls to `create_triangulation_compatibility` and then deprecate that function too.

PR #11784, opened 2021-02-20
Fix vtk output orientation

### LINKED ISSUE #11496: VTK hexahedra have negative volumes
If I plot mesh quality/volume in VisIt I see only negative values here (other cell types all appear to have positive volume). We should fix this orientation.

An example of this problem is `grid_out_vtk_01.output`: the 3D grid at the end of the file produces hexahedra with volumes of -1 and -1/8.

drwells:
Fixes #11496. Part of #8562.

This PR fixes the volume (previously, volumes were off by a factor of -1) of hexahedra written by `GridOut::write_vtk`.

Fixing this uncovered a second bug with the way we read in VTK grids - the vertices were read in as though they were in the UCD format, but VTK uses a different format where the z = 0 vertices are ordered (using deal.II's convention) as 0 - 1 - 3 - 2 and the z = 1 vertices are numbered as 4 - 5 - 7 - 6. This had to be fixed in such a way that GridOut composed with GridIn remains idempotent. Hence, we now correctly renumber each VTK quad and hex to correspond with the lexical (deal.II 6.0 and newer) nodal numbering.

Fixing this uncovered a (minor) third bug (though its one we talked about at the workshop in 2019) where GridReordering::invert_all_cells_of_negative_grid assumes its inputs are still ordered with UCD. To preserve reading of VTK grids where some cells are flipped that function was updated to understand inputs numbered lexically.


drwells:
A lot of output changed in this PR - I looked at plots in gnuplot or VisIt to make sure things at least still look the same. We have a lot of churn in the output because of orientation requirements - we previously read in VTK hexahedra in a very strange way that by coincidence got sorted out by our grid fixing pipeline, but since faces and edges are now present in each cell in a totally different order the ordering computed by `GridReordering::reorder_cells` is now totally different.

include/deal.II/grid/grid_reordering.h
```diff
@@ -668,7 +668,8 @@ class GridReordering
   static void
   invert_all_cells_of_negative_grid(
     const std::vector<Point<spacedim>> &all_vertices,
-    std::vector<CellData<dim>> &        original_cells);
+    std::vector<CellData<dim>> &        original_cells,
+    const bool                          use_new_style_ordering = false);
```
Rombur, line 672: If you want to introduce the new variable, you should document it. In two releases, nobody will now the difference between old and new style. Should we switch the default value to true so that only people who need backward compatibility have to deal with negative volume?

drwells: I'll add the documentation. However, the problem with changing the default argument here is that this function already expects the cells to use the old ordering, so it really does need to be `false`.

I'm hoping to get rid of all this compatibility stuff before the release and move this function into `GridTools` - see also #8562.

Rombur (approved):

agrayver:
This PR broke my code. Now, when reading the mesh the mesh written with GridOut::write_vtk I get:
```

An error occurred in line <1302> of file </home/agrayver/lib/dealii/source/grid/grid_reordering.cc> in function
    static void dealii::GridReordering<dim, spacedim>::invert_all_cells_of_negative_grid(const std::vector<dealii::Point<spacedim> >&, std::vector<dealii::CellData<dim> >&, bool) [with int dim = 3; int spacedim = 3]
The violated condition was: 
    n_negative_cells == 0 || n_negative_cells == cells.size()
```

drwells:
That's odd - we have tests that verify that we can save a grid and load it again. I'm not sure why they would pass when your code fails. I'll take a look.

agrayver:
Wait, it seems I was accidentaly reading a mesh created with an older version (as you suggested). Now it seems to run. Will report back otherwise. Sorry for false alarm.

PR #11785, opened 2021-02-20
Fix fortran mangling

### LINKED ISSUE #5657: Fortran name mangling
In various interfaces (`arpack`, `parpack`, `lapack`, `blas`, `scalapack`) we assume that name mangling is `name_`. We should eventually adopt something like

```
cmake_minimum_required(VERSION 2.8.6)
project(fortranc Fortran C CXX)
include(FortranCInterface)
FortranCInterface_HEADER(FC.h MACRO_NAMESPACE “FC_”)
#This creates a “FC.h” header that defines mangling macros
# FC_GLOBAL(), FC_GLOBAL_(), FC_MODULE(), and FC_MODULE_().
FortranCInterface_VERIFY(CXX)
FortranCInterface_VERIFY()
```

and fall back to 

```
#define “FC_”GLOBAL(name,NAME) name##_
```
in case Fortran compiler was not available during configuration.

Here's one config example https://github.com/elemental/Elemental/blob/master/cmake/detect/Fortran.cmake

drwells:
Fixes #5657. I wrote some code for another project that generates these macros recently so this was pretty easy to fix.

### DIFF
--- a/doc/news/changes/minor/20210220DavidWells
+++ b/doc/news/changes/minor/20210220DavidWells
@@ -1,0 +1,5 @@
+Fixed: deal.II should now be able to use versions of LAPACK compiled with
+Fortran compilers that do not adhere to the usual Fortran mangling
+convention.
+<br>
+(David Wells, 2021/02/20)
--- a/include/deal.II/lac/lapack_templates.h
+++ b/include/deal.II/lac/lapack_templates.h
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 2005 - 2018 by the deal.II authors
+// Copyright (C) 2005 - 2021 by the deal.II authors
 //
 // This file is part of the deal.II library.
 //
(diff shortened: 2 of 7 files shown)


cmake/configure/configure_0_fortran_mangling.cmake
```diff
@@ -0,0 +1,71 @@
...
+      "${DEAL_II_FORTRAN_MANGLE_UNDERSCORE} ## ${FortranCInterface_GLOBAL__SUFFIX}")
+  ENDIF()
+ELSE()
+  # Otherwise just use GNU mangling. This is almost always the right choice
+  # anyway - until 2021 we did not even support other mangling options
+  MESSAGE(STATUS "Could NOT find a valid Fortran compiler - using default mangling")
+
+  SET(DEAL_II_FORTRAN_MANGLE "name ## _")
+  SET(DEAL_II_FORTRAN_MANGLE_UNDERSCORE "name ## _")
+ENDIF()
```
tamiko, line 71: Let's not put this file into the `configure` subdirectory. It does not configure any external dependency.
 What about `checks/check_01_fortran_features.cmake`?

tamiko (changes requested):
Looks good apart from file naming :smile:

cmake/configure/configure_0_fortran_mangling.cmake
drwells: I put it under `02` so that it has the same precedence as compiler features (this seems slightly more logical to me) - is that OK?

drwells:
I noticed another bug with the declarations in the lapack check - I force pushed a fix.

tamiko (approved):

cmake/configure/configure_0_fortran_mangling.cmake
tamiko: Sounds good.

doc/news/changes/minor/20210220DavidWells
```diff
@@ -0,0 +1,5 @@
+Fixed: deal.II should now be able to use versions of LAPACK compiled with
+Fortran compilers that do not adhere to the standard Fortran mangling
+convention.
```
bangerth, line 3: I don't think there actually *is* a standard naming convention. Maybe use "usual" instead?

include/deal.II/lac/lapack_templates.h
```diff
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 2005 - 2018 by the deal.II authors
+// Copyright (C) 2005 - 2021 by the deal.II authors
```
bangerth, line 3: This file is autogenerated, right?

Maybe as a follow-up, it would be useful to not check it into the repository but in fact generate it on the fly in the build directory.

bangerth (approved):
Is there a system on which we can actually test this?

doc/news/changes/minor/20210220DavidWells
drwells: Good idea. I fixed it.

include/deal.II/lac/lapack_templates.h
drwells: It is autogenerated but this is problematic for two reasons:
1. 99% of the data we use to generate it comes from an external copy of LAPACK, where we parse their function declarations to work out types and constness. Hence we can't generate this file unless someone downloads LAPACK and points `generate_lapack_bindings` to that source directory
2. The script we use to generate it is written in Python (though it wouldn't be too hard to convert it to C++)

drwells:
> Is there a system on which we can actually test this?

I don't know. I also tried using an alternative mangling procedure by providing `-DCMAKE_Fortran_FLAGS="-fsecond-underscore"` but CMake seems to ignore that (which is probably a bug on their end). I did verify that we correctly fall back to the most common mangling procedure if we fail to find a fortran compiler.

include/deal.II/lac/lapack_templates.h
bangerth: I think we already use python for some of our scripts. For sure we use PERL to transform the `.inst.in` files.

bangerth:
It's no worse than it was before this patch, and possibly better. So let's go with this -- at least we now have the infrastructure in place if we run into a system that doesn't work.

Thanks for working on this!

include/deal.II/lac/lapack_templates.h
drwells: What? We use `expand_instantiations.cc`. I don't think we have any build dependencies on perl except for the documentation.

bangerth: I stand corrected :-)

drwells: I think we accidentally acquired a python dependency at some point and it was a minor disaster. We need to keep these things optional for building the library itself.

PR #11743, opened 2021-02-13
add ndarray typedef

tamiko:
TODO:
 - [x] add a test
 - [x] finish up documentation

PR #11788, opened 2021-02-21
Use the new ndarray class instead of nested std::array.

bangerth:
Follow-up to #11743.

/rebuild


masterleinad (approved):
Looks good to me.

masterleinad:
```
In file included from /jenkins/workspace/dealii_PR-11788/tests/lac/block_linear_operator_03.cc:18:0:
/jenkins/workspace/dealii_PR-11788/include/deal.II/lac/block_linear_operator.h: In instantiation of ‘dealii::block_operator(const BlockMatrixType&)::<lambda()> [with Range = dealii::BlockVector<double>; Domain = dealii::BlockVector<double>; BlockPayload = dealii::internal::BlockLinearOperatorImplementation::EmptyBlockPayload<>; BlockMatrixType = std::array<std::array<dealii::LinearOperator<dealii::Vector<double>, dealii::Vector<double>, dealii::internal::LinearOperatorImplementation::EmptyPayload>, 2ul>, 2ul>]’:
/jenkins/workspace/dealii_PR-11788/include/deal.II/lac/block_linear_operator.h:593:30:   required from ‘struct dealii::block_operator(const BlockMatrixType&) [with Range = dealii::BlockVector<double>; Domain = dealii::BlockVector<double>; BlockPayload = dealii::internal::BlockLinearOperatorImplementation::EmptyBlockPayload<>; BlockMatrixType = std::array<std::array<dealii::LinearOperator<dealii::Vector<double>, dealii::Vector<double>, dealii::internal::LinearOperatorImplementation::EmptyPayload>, 2ul>, 2ul>]::<lambda()>’
/jenkins/workspace/dealii_PR-11788/include/deal.II/lac/block_linear_operator.h:593:26:   required from ‘dealii::BlockLinearOperator<Range, Domain, BlockPayload> dealii::block_operator(const BlockMatrixType&) [with Range = dealii::BlockVector<double>; Domain = dealii::BlockVector<double>; BlockPayload = dealii::internal::BlockLinearOperatorImplementation::EmptyBlockPayload<>; BlockMatrixType = std::array<std::array<dealii::LinearOperator<dealii::Vector<double>, dealii::Vector<double>, dealii::internal::LinearOperatorImplementation::EmptyPayload>, 2ul>, 2ul>]’
/jenkins/workspace/dealii_PR-11788/tests/lac/block_linear_operator_03.cc:53:36:   required from here
/jenkins/workspace/dealii_PR-11788/include/deal.II/lac/block_linear_operator.h:594:38: error: ‘const struct std::array<std::array<dealii::LinearOperator<dealii::Vector<double>, dealii::Vector<double>, dealii::internal::LinearOperatorImplementation::EmptyPayload>, 2ul>, 2ul>’ has no member named ‘n_block_rows’
     return block_matrix.n_block_rows();
```

bangerth:
Ah, that's interesting now. We get in trouble in essence because we have
```
  template <typename T>
  void f(const T &t);

  template <int n, int m, typename T>
  void f(const ndarray<T,n,m> &t);
```
and then later do
```
  ndarray<double,2,2> x;
  f(x);
```
which used to match the second version of `f` but now is matched to the first one because the compiler can't deduce `n` and `m` in the second version.

@tamiko: Suggestions for how to work around this? Maybe stick with the original nested `std::array` types in the declaration of the functions in `block_linear_operator.h`?

tamiko:
@bangerth Yes, the problem is that `typename internal::ndarray::HelperArray<T, Ns...>::type` is a "non-deduced context". Off the top of my head I don't see any good solution at the moment except not using `ndarray` as a function argument when template argument deduction is intended :neutral_face:

bangerth:
I decided to take the linear operator part out of this patch. We can always come back to this. It does, however, point to a design flaw we may eventually want to rectify. That should be possible by deriving a `ndarray` class from what you already implemented, rather than using a type alias. That can wait, however.

masterleinad:
> That should be possible by deriving a `ndarray` class from what you already implemented, rather than using a type alias.

That's what I looked into earlier today. Unfortunately, that's not quite enough: https://godbolt.org/z/PY8PKE. We can't deduce the template parameters when calling the `ndarray` overload with a `std::array<std::array<>>` argument.

bangerth:
@masterleinad -- nice example. What this shows is that we can't make things backward compatible, but we can at least deal with it if the passed argument is of type `ndarray`.

Either way, let's keep this for another day.

PR #11792, opened 2021-02-22
Add MGTransferBase::prolongate_and_add()

peterrum:


include/deal.II/multigrid/mg_base.h
```diff
@@ -189,7 +189,22 @@ class MGTransferBase : public Subscriptor
   virtual void
   prolongate(const unsigned int to_level,
              VectorType &       dst,
-             const VectorType & src) const = 0;
+             const VectorType & src) const;
```
kronbichler, line 192: One thing I'm wondering is whether we should leave this function pure virtual and let the other function `prolongate_and_add` simply use `prolongate` followed by a vector addition here in the base class. This would allow us to arrange things the other way around, i.e., no changes in all other transfer classes except the `MGTransferMatrixFree` (and possibly the one for global coarsening). I think this is less code overall and slightly simpler to understand.

include/deal.II/multigrid/mg_base.h
```diff
@@ -189,7 +189,22 @@ class MGTransferBase : public Subscriptor
...
+   * @arg src is a vector with as many elements as there are degrees of
+   * freedom on the coarser level involved.
+   *
+   * @arg dst has as many elements as there are degrees of freedom on the
+   * finer level.
+   */
+  virtual void
+  prolongate_and_add(const unsigned int to_level,
+                     VectorType &       dst,
+                     const VectorType & src) const;
```
kronbichler, line 207: The other thing I would like to bring to discussion now is if we want to call this function `prolongate_and_add` (consistent with `restrict_and_add`) or `prolongate_add` (consistent with `vmult_add`). I guess we will go with the consistency within the multigrid transfer classes.

peterrum: I called it this way because of `restrict_and_add` (a function next to it). But, in your original code it was indeed `prolongate_add`. But, I don't have a strong opinion here.

include/deal.II/multigrid/mg_base.h
peterrum: The reason why I went this way is that in the approach you describing we would need to allocated a temporal vector.

kronbichler: The problem is that `prolongate_and_add` is the less common function, and `vmult_add` internally also allocates temporary vectors for some of our matrix classes, e.g. the Trilinos matrices https://github.com/dealii/dealii/blob/e4171ed1729d09c842a6ee77c08bd7e669c5fd1c/source/lac/trilinos_sparse_matrix.cc#L2060-L2070

Therefore I think we should go with the other option, which is also slightly simpler to understand.

include/deal.II/multigrid/mg_base.h
kronbichler: I agree, let us go with `prolongate_and_add`.

peterrum:
@kronbichler I have updated the PR. But I am not sure if this is what you had in mind.

source/multigrid/mg_base.cc
```diff
@@ -41,6 +41,22 @@ MGSmootherBase<VectorType>::apply(const unsigned int level,
...
 
+
+template <typename VectorType>
+void
+MGTransferBase<VectorType>::prolongate_and_add(const unsigned int to_level,
+                                               VectorType &       dst,
+                                               const VectorType & src) const
+{
+  VectorType temp;
+  temp.reinit(dst);
```
kronbichler, line 52:
```suggestion
  temp.reinit(dst, true);
```

kronbichler (commented):
Yes, this is what I had in mind in terms of base class functionality. And then overload the function in `MGTransferMatrixFree` with the code we had before (and whatever else class can implement this call more efficiently).

source/multigrid/mg_transfer_matrix_free.cc
```diff
@@ -188,6 +188,22 @@ MGTransferMatrixFree<dim, Number>::prolongate(
   Assert((to_level >= 1) && (to_level <= level_dof_indices.size()),
          ExcIndexRange(to_level, 1, level_dof_indices.size() + 1));
 
+  dst = 0;
```
kronbichler, line 191: I think you can delete the assertion two lines above because it is inside the `prolongate_and_add` function below.

kronbichler (approved):

peterrum:
@kronbichler Done!

kronbichler (approved):

PR #11793, opened 2021-02-22
Added hp::FECollection::get_hierarchy_sequence().

marcfehling:
We currently have functions to register custom hierarchies for p-adaptation, but no way to actually grasp the entire hierarchy. This PR provides such a function.

In preparation of a follow-up PR.

### DIFF
--- a/include/deal.II/hp/fe_collection.h
+++ b/include/deal.II/hp/fe_collection.h
@@ -506,6 +506,28 @@ namespace hp
     set_default_hierarchy();
 
     /**
+     * Returns a sequence of FE indices that corresponds to the registered
+     * hierarchy in ascending order, i.e., FE indices are sorted from lowest to
+     * highest level.
+     *
+     * Multiple sequences of FE indices are possible with a single custom
+     * hierarchy that can be registered with set_hierarchy(). This function
+     * will return the sequence that contains the user-provided index
+     * @p fe_index which could be located anywhere inside the sequence. The
+     * default hierarchy set via set_default_hierarchy(), which corresponds to
+     * FE indices in ascending order, consists of only one sequence.
+     *
+     * This function can be used, for example, to verify that your provided
+     * hierarchy covers all elements in the desired order.
+     *
+     * Only one sequence of FE indices exists if the size of the returned
+     * container equals the number of elements of this object, i.e.,
+     * FECollection::size().
+     */
+    std::vector<unsigned int>
+    get_hierarchy_sequence(const unsigned int fe_index = 0) const;
+
+    /**
      * Function returning the index of the finite element following the given
      * @p fe_index in hierarchy.
      *
--- a/source/hp/fe_collection.cc
+++ b/source/hp/fe_collection.cc
@@ -315,6 +315,52 @@ namespace hp
 
 
   template <int dim, int spacedim>
+  std::vector<unsigned int>
+  FECollection<dim, spacedim>::get_hierarchy_sequence(
+    const unsigned int fe_index) const
+  {
+    AssertIndexRange(fe_index, size());
+
+    std::deque<unsigned int> sequence = {fe_index};
+
+    // get predecessors
+    {
+      unsigned int front = sequence.front();
+      unsigned int previous;
+      while ((previous = previous_in_hierarchy(front)) != front)
+        {
+          sequence.push_front(previous);
+          front = previous;
+
+          Assert(sequence.size() <= finite_elements.size(),
+                 ExcMessage(
+                   "The registered hierarchy is not terminated: "
+                   "previous_in_hierarchy() does not stop at a final index."));
+        }
+    }
+
+    // get successors
+    {
+      unsigned int back = sequence.back();
+      unsigned int next;
+      while ((next = next_in_hierarchy(back)) != back)
+        {
+          sequence.push_back(next);
+          back = next;
+
+          Assert(sequence.size() <= finite_elements.size(),
+                 ExcMessage(
+                   "The registered hierarchy is not terminated: "
+                   "next_in_hierarchy() does not stop at a final index."));
+        }
+    }
+
+    return {sequence.begin(), sequence.end()};
+  }
+
+
+
+  template <int dim, int spacedim>
   unsigned int
   FECollection<dim, spacedim>::next_in_hierarchy(
     const unsigned int fe_index) const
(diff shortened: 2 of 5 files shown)


Rombur (approved):

peterrum (approved):
I guess we could also encode this hierarchy here?

https://github.com/dealii/dealii/blob/3f9c9daf01b2f674d2cabdb24ae3448cd570d548/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L1810-L1847

include/deal.II/hp/fe_collection.h
```diff
@@ -505,6 +505,26 @@ namespace hp
...
 
+    /**
+     * Returns a sequence of FE indices that corresponds to the registered
+     * hierarchy in ascending order, i.e., FE indices are sorted from lowest to
+     * highest level.
+     *
+     * Multiple sequences of FE indices are possible with a single custom
+     * hierarchy, which can be registered with set_hierarchy(). This function
+     * will return the sequence that contains the user-provided index
+     * @p fe_index. The default hierarchy consists of only one sequence.
```
bangerth, line 516: I suspect that by the "default" case you mean the one that corresponds to the defaulted argument? Doesn't that correspond to the element with index zero?

In any case, I don't quite understand what that default case would correspond to. What would the sequence be if you had a collection of `FE_Q(2)`, `FE_DGQ(0)`, and `FE_DGPMonomial(1)`?

include/deal.II/hp/fe_collection.h
```diff
@@ -505,6 +505,26 @@ namespace hp
...
+     *
+     * This function can be used, for example, to verify that your provided
+     * hierarchy covers all elements in the desired order.
+     *
+     * Only one sequence of FE indices exists if the size of the returned
+     * container equals the number of elements of this object, i.e.,
+     * FECollection::size().
+     */
+    std::deque<unsigned int>
```
bangerth, line 525: That's uncommon :-) We generally use `std::vector` unless there is a good reason because popping is required from both front and back.

source/hp/fe_collection.cc
```diff
@@ -314,6 +314,50 @@ namespace hp
...
+  FECollection<dim, spacedim>::get_hierarchy_sequence(
+    const unsigned int fe_index) const
+  {
+    AssertIndexRange(fe_index, size());
+
+    std::deque<unsigned int> sequence = {fe_index};
+
+    // get predecessors
+    {
+      unsigned int previous, front = sequence.front();
```
bangerth, line 328:
```suggestion
      unsigned int front = sequence.front();
      unsigned int previous;
```

include/deal.II/hp/fe_collection.h
```diff
@@ -505,6 +505,26 @@ namespace hp
     void
     set_default_hierarchy();
 
+    /**
+     * Returns a sequence of FE indices that corresponds to the registered
+     * hierarchy in ascending order, i.e., FE indices are sorted from lowest to
+     * highest level.
```
bangerth, line 511: I had not realized before reading the implementation that the given index can be somewhere in the middle of this sequence. Can you mention this?

source/hp/fe_collection.cc
```diff
@@ -314,6 +314,50 @@ namespace hp
...
+      while ((previous = previous_in_hierarchy(front)) != front)
+        {
+          sequence.push_front(previous);
+          front = previous;
+
+          Assert(sequence.size() <= finite_elements.size(),
+                 ExcMessage(
+                   "The registered hierarchy is not terminated: "
+                   "previous_in_hierarchy() does not stop at a final index."));
```
bangerth, line 337: That would be a funny thing to happen ;-)

source/hp/fe_collection.cc
```diff
@@ -314,6 +314,50 @@ namespace hp
...
+        }
+    }
+
+    // get successors
+    {
+      unsigned int next, back = sequence.back();
```
bangerth, line 343: same as above

source/hp/fe_collection.cc
```diff
@@ -314,6 +314,50 @@ namespace hp
...
+          back = next;
+
+          Assert(sequence.size() <= finite_elements.size(),
+                 ExcMessage(
+                   "The registered hierarchy is not terminated: "
+                   "next_in_hierarchy() does not stop at a final index."));
+        }
+    }
+
+    return sequence;
```
bangerth, line 356: I see why you chose `std::deque` to build this sequence, but that seems like an implementation detail. If you return a `std::vector`, you can do this as follows:
```suggestion
    return {sequence.begin(), sequence.end()};
```

include/deal.II/hp/fe_collection.h
marcfehling: I've been more elaborate on the latest commit.

The default hierarchy corresponds to the FE indices in ascending order and corresponds to the most basic use case, i.e., `FE_Q` elements with degrees in ascending order.

For any wild collection of finite elements, i.e., `FE_Q(2), FE_DGQ(0), FE_DGPMonomial(1)`, the user needs to provide a custom hierarchy. This hierarchy can correspond to different sequences of finite elements, e.g., continuous and discontinuous elements.

For an `FECollection` `FE_Q(1), FE_Q(2), FE_DGQ(1), FE_DGQ(2)`, such a hierarchy might contain two sequences `0,1` for continous elements and `2,3` for discontinous ones. However, we need to know which one of them we want returned, hence the provided parameter of the function.

source/hp/fe_collection.cc
marcfehling: Users could make mistakes with setting up their custom hierarchy correctly, that is why I left this exception here.

marcfehling:
> I guess we could also encode this hierarchy here?
> 
> https://github.com/dealii/dealii/blob/3f9c9daf01b2f674d2cabdb24ae3448cd570d548/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L1810-L1847

That is a good idea, since right now in `step-75` we assume that the degree of an element is its FE index + 1. I'll keep that in mind.

source/hp/fe_collection.cc
marcfehling: I'm just getting very careful whenever I introduce `while` loops ;-)

bangerth: You know how I feel about assertions ;-)

bangerth (approved):

PR #11794, opened 2021-02-23
Made Tensor::values a protected data member

a-shahba:
`Tensor::values` is currently a private data member. Therefore, classes deriving from `dealii::Tensor` class in external projects would need to do `(*this)[ii] = new_value` (which is a bit ugly) to set and get tensor components. By making `Tensor::values` a protected data member, we can write a bit nicer code `this->values[ii] = new_value` in derived classes. Moreover, it means the `dealii::Point` class does not need to be declared as a friend of the Tensor class any longer

### DIFF
--- a/doc/news/changes/minor/20210222AhmadShahba
+++ b/doc/news/changes/minor/20210222AhmadShahba
@@ -1,0 +1,3 @@
+Changed: Madde Tensor::values a protected data member
+<br>
+(Ahmad Shahba, 2021/02/22)
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -747,7 +747,6 @@ private:
   // Allow an arbitrary Tensor to access the underlying values.
   template <int, int, typename>
   friend class Tensor;
-
 };
 
 
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -717,7 +717,7 @@ public:
    */
   using tensor_type = Tensor<rank_, dim, Number>;
 
-private:
+protected:
   /**
    * Array of tensors holding the subelements.
    */
@@ -725,6 +725,7 @@ private:
   // ... avoid a compiler warning in case of dim == 0 and ensure that the
   // array always has positive size.
 
+private:
   /**
    * Internal helper function for unroll.
    */
@@ -747,9 +748,6 @@ private:
   template <int, int, typename>
   friend class Tensor;
 
-  // Point is allowed access to the coordinates. This is supposed to improve
-  // speed.
-  friend class Point<dim, Number>;
 };
 
 


bangerth:
No, this is the wrong approach. If you do this, we will forever be tied to the internal design of this class because others can actually derive from this class.

What is your use case for actually deriving from this class?

jppelteret:
I concur with @bangerth. The Tensor class provides a number of read-only and read-write accessors to its underlying data, so I'm pretty sure that you should be able to achieve what you want to do using those. Here's a summary of the current options:

Read-only ("get"):
- Iterators via [`begin_raw()`](https://dealii.org/current/doxygen/deal.II/classTensor.html#a7aad90204dd22941d557dd847bd991a9) and [`end_raw()`](https://dealii.org/current/doxygen/deal.II/classTensor.html#ad4e1c9b494f6133582a0520f7c6246e1).
- Access operator ([index variant](https://dealii.org/current/doxygen/deal.II/classTensor.html#a4e175587b3c7ec3e69dc2389d2e12d1f)). Useful in conjunction with [`component_to_unrolled_index()`](https://dealii.org/current/doxygen/deal.II/classTensor.html#a691feb72e96831cab4a85fbc28167abd) if accessing only select entries.
- Access operator ([table index variant](https://dealii.org/current/doxygen/deal.II/classTensor.html#a551bc8879285f6fa0c1af8edfe1e97ba)). Useful in conjunction with [`unrolled_to_component_indices()`](https://dealii.org/current/doxygen/deal.II/classTensor.html#a2967bd64add4b290f52801a682d722d2) if you're looping over all of the components in range [0, [`n_independent_components`](https://dealii.org/current/doxygen/deal.II/classTensor.html#aed199dbe319871d4d1ee68ef763cc4ac)]
- [unroll()](https://dealii.org/current/doxygen/deal.II/classTensor.html#a5247561f23bd98ebc99dc8d54b1b19fe) to copy out a flattened representation of the data

Read-write ("set"):
- Iterators via [`begin_raw()`](https://dealii.org/current/doxygen/deal.II/classTensor.html#aaf51824501b35657c109b15b24845ade) and [`end_raw()`](https://dealii.org/current/doxygen/deal.II/classTensor.html#a71e9c49f0f76fa957bff112e5cf6973e)
- Access operator ([index variant](https://dealii.org/current/doxygen/deal.II/classTensor.html#a4e175587b3c7ec3e69dc2389d2e12d1f))
- Access operator ([table index variant](https://dealii.org/current/doxygen/deal.II/classTensor.html#a4ad45d5ef39bb19c1781104cfa7f2719))

a-shahba:
I can work with the current interface of the `Tensor` class and use its public methods (the ones Jean-Paul mentioned) to set/get the tensor values. No question! 
I thought about Wolfgang's rationale and have to say that I agree. I now understand why you would like to keep it `private`

I am deriving from `Tensor` class in my project to represent a surface which is defined by its out-ward normal and an id.

drwells:
> I am deriving from Tensor class in my project to represent a surface which is defined by its out-ward normal and an id.

It's a little off topic, but for things like this in deal.II we use `std::pair<types::boundary_id, Tensor<1, spacedim>>` rather than using inheritance. The advantage to this approach is that copying, assignment, `operator<`, etc. will all work automatically.

bangerth:
I think we've come to a conclusion here, and so I'll close the issue in a minute.

There is a general rule in OO programming that says that you derive a class if the derived class has an "is a" relationship to the base class. For example, the class `Automobile` would reasonably derive from a base class `Vehicle` because a car *is a* vehicle. On the other hand, if you have a "has a" relationship, then you would use member variables. In your case, you have a manifold that *has a* normal vector, so it would make sense if the normal vector would be a member variable. (I don't know whether that correctly matches what you are doing, but you get the idea.)

PR #11757, opened 2021-02-15
Step 17 as a simplex test (clean commit history)

michelebucelli:
This PR follows the discussion on #11551 in an attempt to fix failing tests.

Step 17 (linear elasticity, MPI parallel) converted to a simplex test.

PR #11795, opened 2021-02-23
Michelebucelli step 17 simplex cleanup

bangerth:
I can not figure out what is happening with #11757. That test works without problem on my laptop. So I'm pushing it from my machine to a separate PR -- let's see what happens here.

/rebuild

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -85,6 +85,11 @@ jobs:
         make -j 2
     - name: test
       run: |
+        # Remove warning: "A high-performance Open MPI point-to-point
+        # messaging module was unable to find any relevant network
+        # interfaces."
+        export OMPI_MCA_btl_base_warn_component_unused='0'
+        
         cd build
         make -j 2 setup_tests_simplex
         ctest --output-on-failure -j 2
--- a/tests/simplex/step-17.cc
+++ b/tests/simplex/step-17.cc
@@ -1,0 +1,488 @@
+/* ---------------------------------------------------------------------
+ *
+ * Copyright (C) 2000 - 2019 by the deal.II authors
+ *
+ * This file is part of the deal.II library.
+ *
+ * The deal.II library is free software; you can use it, redistribute
+ * it, and/or modify it under the terms of the GNU Lesser General
+ * Public License as published by the Free Software Foundation; either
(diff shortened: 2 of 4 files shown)


bangerth:
It fails in the same way as in #11757 even though it works on my system. I'm stumped, and I can't even figure out what precisely the output is on the CI check. Anyone have insight?

peterrum:
@bangerth https://github.com/dealii/dealii/pull/11551#issuecomment-764650004

But I didn't have time to work on that till now.

bangerth:
Rebased on top of #11835.

peterrum:
@bangerth You need to update the tests: the simplex folder has been removed in the meantime and the files/namespaces are also renamed.

peterrum:
The log says:
```
--------------------------------------------------------------------------
[[38647,1],0]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:

Module: OpenFabrics (openib)
  Host: fv-az195-260

Another transport will be used instead, although this may result in
lower performance.

NOTE: You can disable this warning by setting the MCA parameter
btl_base_warn_component_unused to 0.
--------------------------------------------------------------------------
   Number of active cells:       200
   Number of degrees of freedom: 242 (by partition: 242)
   Solver converged in 10 iterations.
```

bangerth:
Huh, what does that even mean? And why does this only happen for this one test, and reproducibly so?

Rombur:
I have seen that warnings in multiple codes. It has to do with some permissions with Docker. For some communications, OpenMPI wants to use a protocol that cannot be used by default in Docker. We should just set the variable to ignore this warnings.

bangerth:
Do you remember how to do that? Separately, why would that only happen to this one test? Does it happen to have an awkward communication pattern?

Rombur:
> Do you remember how to do that?

Yes, it's pretty simple we can simply the environment variables in the jenkins configuration file like [here](https://github.com/ORNL-CEES/DataTransferKit/blob/master/.jenkins#L3-L10)

> Separately, why would that only happen to this one test?

Honestly, I have no idea why this happens for some communications but not others.

peterrum:
I'll try to wrap this PR up in the next days.

bangerth:
Thanks, @peterrum .

peterrum (approved):

PR #11797, opened 2021-02-23
Provide compile-time error messages if we don't handle a type.

bangerth:
Based on a question on the mailing list today.

/rebuild


drwells (approved):
A tester failed (compiler was terminated), so lets try again

/rebuild

PR #11798, opened 2021-02-23
Bugfix: DoFHandler connects to Triangulation::Signals only in hp-mode.

### LINKED ISSUE #11768: p::d::SolutionTransfer::deserialize() fails on non-coarse meshes
Hi everybody,

the [attached code](https://github.com/dealii/dealii/files/5997690/serialization.zip) (inspired by [this test](https://github.com/dealii/dealii/blob/master/tests/fullydistributed_grids/solution_transfer_01.cc)) fails if run in parallel.

It shows a simple use of `p::d::SolutionTransfer::{serialize,deserialize}()` for de-serializing a non-coarse `p::d::Triangulation`.

Success or failure depends on the number of MPI cores (on my machine it fails with 6 cores).
The problem arises both in 2D and in 3D, using either deal.II or Trilinos or PETSc distributed vectors and on older deal.II releases.

The code works correctly for `p::f::Triangulation`s or if the serialized triangulation is coarse.

Any clues?

Thanks!

Attachment:
[serialization.zip](https://github.com/dealii/dealii/files/5997690/serialization.zip)

marcfehling:
Closes #11768.

Whenever a `Triangulation` will be cleared, the enumeration of dofs on associated `DoFHandlers` is no longer valid. This PR proposes to also clear the `DoFHandlers` as well.

### DIFF
--- a/include/deal.II/dofs/dof_handler.h
+++ b/include/deal.II/dofs/dof_handler.h
@@ -1667,6 +1667,13 @@ private:
   std::vector<boost::signals2::connection> tria_listeners;
 
   /**
+   * A list of connections with which this object connects to the
+   * triangulation. They get triggered specifially when data needs to be
+   * transferred due to refinement or repartitioning. Only active in hp-mode.
+   */
+  std::vector<boost::signals2::connection> tria_listeners_for_transfer;
+
+  /**
    * Free all memory used for non-multigrid data structures.
    */
   void
@@ -1700,14 +1707,13 @@ private:
                 const types::global_dof_index global_index) const;
 
   /**
-   * Setup DoFHandler policy.
+   * Set up DoFHandler policy.
    */
   void
   setup_policy();
 
   /**
-   * Setup connections to refinement signals of the underlying triangulation.
-   * Necessary for the hp-mode.
+   * Set up connections to signals of the underlying triangulation.
    */
   void
   connect_to_triangulation_signals();
--- a/source/distributed/solution_transfer.cc
+++ b/source/distributed/solution_transfer.cc
@@ -137,6 +137,10 @@ namespace parallel
       prepare_for_coarsening_and_refinement(
         const std::vector<const VectorType *> &all_in)
     {
+      for (unsigned int i = 0; i < all_in.size(); ++i)
+        Assert(all_in[i]->size() == dof_handler->n_dofs(),
+               ExcDimensionMismatch(all_in[i]->size(), dof_handler->n_dofs()));
+
       input_vectors = all_in;
       register_data_attach();
     }
@@ -230,6 +234,9 @@ namespace parallel
     {
       Assert(input_vectors.size() == all_out.size(),
              ExcDimensionMismatch(input_vectors.size(), all_out.size()));
+      for (unsigned int i = 0; i < all_out.size(); ++i)
+        Assert(all_out[i]->size() == dof_handler->n_dofs(),
+               ExcDimensionMismatch(all_out[i]->size(), dof_handler->n_dofs()));
 
       // TODO: casting away constness is bad
       auto *tria = (dynamic_cast<parallel::DistributedTriangulationBase<
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -1999,11 +1999,14 @@ DoFHandler<dim, spacedim>::DoFHandler(const Triangulation<dim, spacedim> &tria)
 template <int dim, int spacedim>
 DoFHandler<dim, spacedim>::~DoFHandler()
 {
-  // unsubscribe all attachments to refinement signals of the underlying
-  // triangulation
+  // unsubscribe all attachments to signals of the underlying triangulation
   for (auto &connection : this->tria_listeners)
     connection.disconnect();
   this->tria_listeners.clear();
+
+  for (auto &connection : this->tria_listeners_for_transfer)
+    connection.disconnect();
+  this->tria_listeners_for_transfer.clear();
 
   // release allocated memory
   // virtual functions called in constructors and destructors never use the
@@ -2051,6 +2054,10 @@ DoFHandler<dim, spacedim>::reinit(const Triangulation<dim, spacedim> &tria)
   for (auto &connection : this->tria_listeners)
     connection.disconnect();
   this->tria_listeners.clear();
+
+  for (auto &connection : this->tria_listeners_for_transfer)
+    connection.disconnect();
+  this->tria_listeners_for_transfer.clear();
 
   // release allocated memory and policy
   DoFHandler<dim, spacedim>::clear();
@@ -2431,10 +2438,11 @@ DoFHandler<dim, spacedim>::set_fe(const hp::FECollection<dim, spacedim> &ff)
         {
           hp_capability_enabled = false;
 
-          // unsubscribe connections to signals
-          for (auto &connection : this->tria_listeners)
+          // unsubscribe connections to signals that are only relevant for
+          // hp-mode, since we only have a single element here
+          for (auto &connection : this->tria_listeners_for_transfer)
             connection.disconnect();
-          this->tria_listeners.clear();
+          this->tria_listeners_for_transfer.clear();
 
           // release active and future finite element tables
           this->hp_cell_active_fe_indices.clear();
@@ -2510,10 +2518,11 @@ DoFHandler<dim, spacedim>::distribute_dofs(
         {
           hp_capability_enabled = false;
 
-          // unsubscribe connections to signals
-          for (auto &connection : this->tria_listeners)
+          // unsubscribe connections to signals that are only relevant for
+          // hp-mode, since we only have a single element here
+          for (auto &connection : this->tria_listeners_for_transfer)
             connection.disconnect();
-          this->tria_listeners.clear();
+          this->tria_listeners_for_transfer.clear();
 
           // release active and future finite element tables
           this->hp_cell_active_fe_indices.clear();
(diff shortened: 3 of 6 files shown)


marcfehling:
This might break [tests/fullydistributed_grids/solution_transfer_01.cc](https://github.com/dealii/dealii/blob/master/tests/fullydistributed_grids/solution_transfer_01.cc). I'll have a look.

marcfehling:
Test [tests/fullydistributed_grids/solution_transfer_01.cc](https://github.com/dealii/dealii/blob/master/tests/fullydistributed_grids/solution_transfer_01.cc) should fail, but it passes, and I don't know why right now.

If I manually add `dof_handler.clear()` after `triangulation.clear()`, the test fails. Without it, it passes.
It seems like the `clear` signal does not get triggered here. I'll investigate.

source/dofs/dof_handler.cc
```diff
@@ -2006,10 +2006,7 @@ DoFHandler<dim, spacedim>::~DoFHandler()
   this->tria_listeners.clear();
 
   // release allocated memory
-  // virtual functions called in constructors and destructors never use the
-  // override in a derived class
-  // for clarity be explicit on which function is called
-  DoFHandler<dim, spacedim>::clear();
+  clear();
```
bangerth, line 2009: I guess you're changing this because we no longer have classes derived from `DoFHandler`?

source/dofs/dof_handler.cc
```diff
@@ -3044,6 +3041,8 @@ DoFHandler<dim, spacedim>::connect_to_triangulation_signals()
   // connect functions to signals of the underlying triangulation
   this->tria_listeners.push_back(this->tria->signals.create.connect(
     [this]() { this->reinit(*(this->tria)); }));
+  this->tria_listeners.push_back(
+    this->tria->signals.clear.connect([this]() { this->clear(); }));
```
bangerth, line 3045: This looks correct to me. Nothing good can happen if the triangulation is cleared and we still have dangling data.

bangerth:
It might be that the `clear()` signal gets triggered *at the beginning of* `Triangulation::clear()` (which is where it should be in my mind).

marcfehling:
> It might be that the `clear()` signal gets triggered _at the beginning of_ `Triangulation::clear()` (which is where it should be in my mind).

I found the issue. It's because we clear all connections to `Triangulation` signals in non-hp-mode. It's a relic from merging both `DoFHandler` classes. I'm about to prepare a fix.

source/dofs/dof_handler.cc
marcfehling: Exactly. We can also get rid of most `this` pointers in the implementation.

However, this does not have to be part of this PR. I tend to move it to a separate one.

bangerth: That's ok, but it's also fine as is.

marcfehling:
Blocked by #11800.

I had to separate the signal connection in a general and a hp-mode collection. Once we leave hp-mode, only the latter collection will be cleared, while the first one will persist.

`tests/fullydistributed_grids/solution_transfer_01.cc` now fails as expected. I adjusted it to resemble a use case with a fresh `Triangulation`. It works with one process, but still fails with more than one. @elauksap  @peterrum -- Do you have an idea?

marcfehling:
Some more information about the error:
- occurs only on more than 1 MPI process
- happens when updating ghost values after deserialization on [line 110](https://github.com/dealii/dealii/blob/667d1c1d14cc01a2299a0d82f36e4798e78442ba/tests/fullydistributed_grids/solution_transfer_01.cc#L110)
- exception thrown:
```
An error occurred in line <635> of file </home/nonfac/f/fehling/raid/dealii/include/deal.II/base/partitioner.templates.h> in function
    void dealii::Utilities::MPI::Partitioner::import_from_ghosted_array_finish(dealii::VectorOperation::values, const dealii::ArrayView<const ElementType, MemorySpaceType>&, const dealii::ArrayView<ElementType, MemorySpace>&, const dealii::ArrayView<ElementType, MemorySpace>&, std::vector<ompi_request_t*, std::allocator<ompi_request_t*> >&) const [with Number = double; MemorySpaceType = dealii::MemorySpace::Host]
The violated condition was: 
    *read_position == Number() || internal::get_abs(locally_owned_array[j] - *read_position) <= internal::get_abs(locally_owned_array[j] + *read_position) * 100000. * std::numeric_limits<typename numbers::NumberTraits< Number>::real_type>::epsilon()
Additional information: 
    Called compress(VectorOperation::insert), but the element received
    from a remote processor, value 0.125, does not match with the value
    0.5153882032022076 on the owner processor 0
```

marcfehling:
Closes #11768.

elauksap:
@marcfehling thanks for your work!

I've found out that adding either of the following modifications (lines marked with a `//`) actually makes the test pass:
```
  // Line 78.
  else if (mode == "load")
    {
      triangulation.load(filename);

      triangulation.clear(); //
      triangulation.load(filename); //
    }
```

```
  // Line 108.
  else if (mode == "load")
    {
      triangulation.clear(); //
      triangulation.load(filename); //
      solution_transfer.deserialize(vector);

      vector.update_ghost_values();
    }
```

I have no idea why, it is very strange. I try to investigate further.

source/dofs/dof_handler.cc
bangerth: I continue to think that this is correct as is. If you want, break this uncontroversial part out into a patch of its own. Let's get the easy pieces out of the way already.

source/dofs/dof_handler.cc
marcfehling: Hmm, we still have the deprecated `hp::DoFHandler` around that will probably need this. I'll revert it.

marcfehling:
Moved the original PR to #11801.

I left this PR at the current state to keep the comments.

source/dofs/dof_handler.cc
masterleinad: It really doesn't make a difference in practice. The code is just clearly telling what's happening here (which some static analyzers demanded).

bangerth:
Can you rebase now?

include/deal.II/dofs/dof_handler.h
```diff
@@ -1700,8 +1707,7 @@ class DoFHandler : public Subscriptor
   setup_policy();
 
   /**
-   * Setup connections to refinement signals of the underlying triangulation.
-   * Necessary for the hp-mode.
+   * Setup connections to signals of the underlying triangulation.
```
bangerth, line 1710: We do this wrongly in many places, but the correct use of the word is "Set up connections...". For sure for verbs, it should be spelled apart. For nouns ("the setup we have here") might be spelled together, though "the set up of these connections" would probably still be apart.

Really not important here, but I've been finding these places all over the library.

source/distributed/solution_transfer.cc
```diff
@@ -137,6 +137,12 @@ namespace parallel
       prepare_for_coarsening_and_refinement(
         const std::vector<const VectorType *> &all_in)
     {
+#  ifdef DEBUG
+      for (unsigned int i = 0; i < all_in.size(); ++i)
+        Assert(all_in[i]->size() == dof_handler->n_dofs(),
```
bangerth, line 142: I think you can remove the `#ifdef`. The `Assert` expands to nothing in release mode and the compiler can hopefully see that the `all_in.size()` call doesn't change any state of `all_in` so that the whole loop becomes a no-op.

source/distributed/solution_transfer.cc
```diff
@@ -228,8 +234,13 @@ namespace parallel
     SolutionTransfer<dim, VectorType, DoFHandlerType>::interpolate(
       std::vector<VectorType *> &all_out)
     {
+#  ifdef DEBUG
```
bangerth, line 237: same here

source/dofs/dof_handler.cc
```diff
@@ -2431,10 +2438,10 @@ DoFHandler<dim, spacedim>::set_fe(const hp::FECollection<dim, spacedim> &ff)
         {
           hp_capability_enabled = false;
 
-          // unsubscribe connections to signals
-          for (auto &connection : this->tria_listeners)
+          // unsubscribe connections to signals that are relevant for hp-mode
+          for (auto &connection : this->tria_listeners_for_transfer)
```
bangerth, line 2442: It took me a little bit to see why this is possible -- because you aren't actually in hp mode any more. Maybe say that in the comment: "...that are only relevant for hp-mode, but we only have a single element here" (emphasis on *only*)

source/dofs/dof_handler.cc
```diff
@@ -2510,10 +2517,10 @@ DoFHandler<dim, spacedim>::distribute_dofs(
         {
           hp_capability_enabled = false;
 
-          // unsubscribe connections to signals
-          for (auto &connection : this->tria_listeners)
+          // unsubscribe connections to signals that are relevant for hp-mode
```
bangerth, line 2520: same here

bangerth (commented):
Nearly there.

source/distributed/solution_transfer.cc
marcfehling: I don't have a preference on the `DEBUG` macros here. I like being more explicit about sections that only happen in `DEBUG` mode, but right here and in the beginning of the function it should be obvious. Will comply and remove them!

marcfehling:
This PR still breaks `tests/fullydistributed_grids/solution_transfer_01.cc`.
Once we figured out how to fix it, I will recover the PR from its draft state.

marcfehling:
On the other hand, we could deal with changing `tests/fullydistributed_grids/solution_transfer_01.cc` to resemble an actual use case on a separate PR and get this merged. As @bangerth reviewed this PR already, feel free to start the testsuite.

tests/fullydistributed_grids/solution_transfer_01.cc
```diff
@@ -89,6 +89,7 @@ test(TriangulationType &triangulation)
 
   {
     triangulation.load(filename);
+    dof_handler.distribute_dofs(FE_Q<dim>(2));
```
bangerth, line 92: Yes, this makes sense. The fact that this worked before is a complete coincidence and it's probably good that you broke this.

bangerth (approved):

PR #11803, opened 2021-02-24
Introduce three new FEEvaluation functions

peterrum:


include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -3610,6 +3628,17 @@ FEEvaluationBaseData<dim, Number, is_face, VectorizedArrayType>::
 
 
 
+template <int dim, typename Number, bool is_face, typename VectorizedArrayType>
+inline const internal::MatrixFreeFunctions::DoFInfo &
+FEEvaluationBaseData<dim, Number, is_face, VectorizedArrayType>::get_dof_info()
+  const
+{
+  Assert(dof_info != nullptr, ExcInternalError());
```
kronbichler, line 3636: I think `ExcInternalError` is the wrong error type here because you can initialize `FEEvaluation` with a `FiniteElement` (unusual initialization case) here https://github.com/dealii/dealii/blob/e4171ed1729d09c842a6ee77c08bd7e669c5fd1c/include/deal.II/matrix_free/fe_evaluation.h#L3382-L3401
which will not have this field. I think you need something like `ExcMessage("FEEvaluation was not initialized with a MatrixFree object")` and the same for the `get_matrix_free()` function below.

peterrum: Done!

kronbichler (approved):

include/deal.II/matrix_free/fe_evaluation.h
```diff
@@ -3610,6 +3628,19 @@ FEEvaluationBaseData<dim, Number, is_face, VectorizedArrayType>::
...
 
 
+template <int dim, typename Number, bool is_face, typename VectorizedArrayType>
+inline const internal::MatrixFreeFunctions::DoFInfo &
+FEEvaluationBaseData<dim, Number, is_face, VectorizedArrayType>::get_dof_info()
+  const
+{
+  Assert(dof_info != nullptr,
+         ExcMessage(
+           "FEEvaluation was not initialized with a MatrixFree object".));
```
kronbichler, line 3638: The CI complains:
```suggestion
           "FEEvaluation was not initialized with a MatrixFree object!"));
```

PR #11804, opened 2021-02-24
Add utility functions to evaluate values of arbitrary (remote) points

peterrum:
The big brother of `FEPointEvaluation`.

The classes/functions have enabled us to 
- interpolate between distributed non-matching grids (ALE, NS+LS, and [between hex and tet meshes](https://github.com/dealii/dealii/pull/11514#issuecomment-759757724))
- evaluate values along lines
- create slice directly in deal.II (instead of outputting 3D results and creating the slices in Paraview)

I will add some tests tomorrow so that the use cases should be more clear.


luca-heltai:
This basically re-implements `distribute_compute_point_locations` and some of the `GridTools::Cache` objects from scratch, but using basically the same algorithm (bounding boxes describing the domain, check who may possibly own the points, and use a consensus algorithm to decide who owns what, then send around data). 

With the existing functionality, this could have been implemented in the following way:

~~~C++
LA::MPI::Vec global_vec;
...
FiniteElement<dim,spacedim> &fe = ...;
...
std::vector<Point<spacedim>> points(n_points);
std::vector<double> values(n_points);

auto global_boxes = cache.get_covering_rtree(tree_level);
auto [cells, local_points, local_indices, global_points, from_cpu] = distributed_compute_point_locations(points, cache, {global_boxes.begin(), global_boxes.end()});

for(unsigned int i=0; i<cells.size(); ++i) {
   const auto &cell = cells[i];
   const auto &qpoints = local_points[i];
   const auto &indices= local_indices[i];
   for(unsigned int j=0; j<indices.size(); ++j)
   values[indices[j]] = fe.value(qpoints[j]);
}
~~~

I think your implementation is likely to be more efficient, but can we at least gather functions that do the same stuff around with just one implementation (i.e., the one that works best), and throw away the others?

kronbichler:
@luca-heltai you are right, we should try to unify the implementations and have a good infrastructure that fits a wide application range. I will look into this more closely tomorrow.

source/base/mpi_remote_point_evaluation.cc
```diff
@@ -0,0 +1,448 @@
...
+            // send requested points
+            for (auto point :
+                 potentially_relevant_points_per_process[other_rank])
+              for (unsigned int i = 0; i < spacedim; ++i)
+                send_buffer.emplace_back(point[i]);
+          },
+          [&](const unsigned int &       other_rank,
+              const std::vector<double> &recv_buffer,
+              std::vector<unsigned int> &request_buffer) {
+            // received points, determine if point is actually possessed,
```
kronbichler, line 237:
```suggestion
            // Once a tentative point is handed it, determine if it is actually possessed,
```

source/base/mpi_remote_point_evaluation.cc
```diff
@@ -0,0 +1,448 @@
...
+
+      std::sort(recv_ranks.begin(), recv_ranks.end());
+
+      std::vector<unsigned int> quadrature_points_count(
+        quadrature_points.size(), 0);
+      std::vector<std::pair<unsigned int, unsigned int>> indices_temp;
+
+      indices_ptr = {0}; // TODO: name
+
+      for (const auto &i : relevant_points_per_process_offset)
```
kronbichler, line 323: Can you comment what this code does.

source/base/mpi_remote_point_evaluation.cc
```diff
@@ -0,0 +1,448 @@
...
+#endif
+    }
+
+
+    template <int dim, int spacedim>
+    const std::vector<unsigned int> &
+    RemotePointEvaluation<dim, spacedim>::get_point_ptrs() const
+    {
+      return quadrature_points_ptr;
+    }
```
kronbichler, line 413: I think this function (and the small ones below) can go to the header file.

kronbichler (commented):
I have some various comments that I would like to discuss.

peterrum:
> This basically re-implements distribute_compute_point_locations and some of the GridTools::Cache objects from scratch,

This statement is not exactly precise. The main use case of `GridTools::distribute_compute_point_locations()` is `ParticleHandler::insert_global_particles()`: points are generated on one mesh and then sent to another (potentially non-matching) mesh. I.e., data (points) are only sent in one direction and on receiver side some quantities are derived (in particular cells and reference points).

In contrast, the the goal of this PR it be enable evaluation/integration on arbitrary points (local and remote; potentially on non-matching meshes). A function added is `VectorTools::evaluate_at_points()` that allows to evaluate the values at arbitrary points on a global solution vector. This requires that data is also sent back after they have been computed at the points.

To be able to perform the communication, additional data structures have to be set up. In particular, data have to be sorted in each stage differently: during evaluation data needs to be sorted according to the owning cell; during communication the data has to be sorted according to the ranks; and at the end the data has to sorted in the order of the provided points. Furthermore, this needs additional communication steps for finalizing the data structures via handshake that can be performed via `ConsensusAlgorithms`.

Since the result of `GridTools::distribute_compute_point_locations()` is only (in one way or an other) a side result of what we need and can be obtained if the hand-shake is not performed (early return), I have created an internal function `GridTools::internal::distribute_compute_point_locations()` that can be configured via a bool. `GridTools::internal::distribute_compute_point_locations(..., false)` is now used in `GridTools::distribute_compute_point_locations()` and `GridTools::internal::distribute_compute_point_locations(..., true)` is used to set up the communication patter within `Utilities::MPI::RemotePointEvaluation`.

`Utilities::MPI::RemotePointEvaluation` is essentially a class that cashes the communication pattern (similarly to `Utilities::MPI::Partitioner`) and is responsible for the permutation of the data on the sender and receiver side.

This class can be passed to `VectorTools::evaluate_at_points()` (as cache) so that the communication pattern does not have to be updated in each function call. However, the use cases are much more diverse...

_Note: I will move some minor changes to separate PRs._

include/deal.II/base/mpi_consensus_algorithms.templates.h
```diff
@@ -579,7 +579,8 @@ namespace Utilities
 
         this->process.create_request(0, send_buffer);
         this->process.prepare_buffer_for_answer(0, recv_buffer);
-        this->process.answer_request(0, buffer_recv, request_buffer);
+        this->process.answer_request(0, send_buffer, request_buffer);
+        recv_buffer = request_buffer;
```
peterrum, line 583: Moved to #11840.

source/base/mpi_consensus_algorithms.cc
```diff
@@ -66,6 +66,15 @@ namespace Utilities
...
 
+      template class Process<char, char>;
+
+      template class NBX<char, char>;
+
+      template class PEX<char, char>;
+
+      template class Serial<char, char>;
+
+      template class Selector<char, char>;
```
peterrum, line 77: Moved to #11839.

source/base/mpi.cc
```diff
@@ -368,9 +368,9 @@ namespace Utilities
         {
           (void)destination;
           AssertIndexRange(destination, n_procs);
-          Assert(destination != myid,
-                 ExcMessage(
-                   "There is no point in communicating with ourselves."));
+          // Assert(destination != myid,
+          //       ExcMessage(
+          //         "There is no point in communicating with ourselves."));
```
peterrum, line 373: Moved to #11839.

source/base/mpi_remote_point_evaluation.cc
peterrum: This is outdated.

source/base/mpi_remote_point_evaluation.cc
peterrum: This one as well.

source/base/mpi_remote_point_evaluation.cc
```diff
@@ -0,0 +1,448 @@
...
+          try
+            {
+              const auto first_cell = GridTools::find_active_cell_around_point(
+                cache, point, cell_hint, marked_vertices, tolerance);
+
+              cell_hint = first_cell.first;
+
+              const auto active_cells_around_point =
+                GridTools::find_all_active_cells_around_point(
+                  mapping, tria, point, tolerance, first_cell);
```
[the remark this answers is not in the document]
peterrum: True.

include/deal.II/numerics/vector_tools_evaluate.h
```diff
@@ -0,0 +1,252 @@
...
+    using value_type =
+      typename FEPointEvaluation<n_components, dim>::value_type;
+
+    Assert(eval.is_ready(),
+           ExcMessage(
+             "Utilties::MPI::RemotePointEvaluation is not ready yet!"));
+
+    // evaluate values at points if possible
+    const auto evaluation_point_results = [&]() {
+      std::vector<std::unique_ptr<FEPointEvaluation<n_components, dim>>>
```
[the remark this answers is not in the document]
peterrum: I know you don't like these kind of scopes. For me, however, it helps to structure code.

include/deal.II/numerics/vector_tools_evaluate.h
```diff
@@ -0,0 +1,252 @@
...
+    };
+  }
+
+  /**
+   * Given a (distributed) solution vector @p vector, evaluate the values at
+   * the (arbitrary and even remote) points specified by @p evaluation_points.
+   */
+  template <int n_components, int dim, int spacedim, typename VectorType>
+  std::vector<typename FEPointEvaluation<n_components, dim>::value_type>
+  evaluate_at_points(
```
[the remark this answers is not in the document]
peterrum: > In terms of technicalities, I think we can return any information FEPointEvaluation can compute, as long as we can pack the data appropriately.

Yes that is right. I would suggest once other quantities are needed, we move the content of this function to an internal (templated) function, which is called by `evaluate_values_at_points()`, `evaluate_gradients_at_points()`, ...

> I think we should also think about the case where a user requests the gradient of a solution field (or the Hessian). How would we structure the case of various return kinds? Set up some additional data structure that is filled upon demand?

The bigger question is how we can deal with derived quantities (e.g. in the case of SIM I would like to communicate already the force computed with `JxW` so that only has to be tested after communication) and with composites (e.g., you might want to pack multiple quantities (one might be a scalar, the other one vectorial) together so that they are sent in one go; e.g. BDF-2 multiple timesteps after AMR step).

The best solution I was able to come up (in adaflo) for this till now is use `RemovePointEvaluation` directly. This class can work with arbitrary types so that I can also pass in vectors of tuples. What is your opinion?

include/deal.II/numerics/vector_tools_evaluate.h
```diff
@@ -0,0 +1,252 @@
...
+   * the points specified by the @p cache which might have been set up by the
+   * above function.
+   *
+   * @note Refinement/coarsening/repartitioning leads to the invalidation of the
+   *   cache so that the above function has to be called again.
+   */
+  template <int n_components, int dim, int spacedim, typename VectorType>
+  std::vector<typename FEPointEvaluation<n_components, dim>::value_type>
+  evaluate_at_points(
+    const DoFHandler<dim, spacedim> &                           dof_handler,
```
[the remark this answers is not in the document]
peterrum: I don't think it is that "surprising". All `VectorTools` functions get a `DoFHandler`object. The `DoFHandler` object is connected "directly" to the vector while the communication pattern is only connected to the grid.

> So in conclusion I do not disagree here, but I think we need to work hard to document this well.

But yes we should!

include/deal.II/numerics/vector_tools_evaluate.h
```diff
@@ -0,0 +1,252 @@
...
+  namespace EvaluationFlags
+  {
+    enum EvaluationFlags
+    {
+      avg    = 0,
+      max    = 1,
+      min    = 2,
+      insert = 3
+    };
+  }
```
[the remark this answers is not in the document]
peterrum: Before making this change, let's agree on how to deal with `gradients`, `hessians`! Is it an option that we have for each a separate function as stated a couple of minutes ago!? Or do we want to encrypt this in these "flags". An issue I seen in the latter is that we don't have a way to determine the return type of the function based on the flag so that the user might need to be explicit regarding the template argument!?

source/base/mpi_remote_point_evaluation.cc
peterrum: Yes. I can do that. What is the criterion when to leave a function in the header file. I would have thought that there are only two reasons: 1) that the function is templated in such a way that explicit instantiation is hard and 2) the calls is performance critical (as is in many places of `FEEvaluation`). But the functions here are definitely not performance critical so that they do not need to be inlined...

peterrum:
PR is ready from my side. I have renamed some of the fields and have added some documentation. The discussions above (particularity regarding usability and extensibility - computation of other quantities) are still open. I will resolve these once we have agreed on a good solution.

peterrum:
Regarding the discussion above. To be able to work with `RemotePointEvaluation`, one needs to define a lambda-function like:
https://github.com/dealii/dealii/blob/57169e3575a79895ac50278936cc835f49597470/include/deal.II/numerics/vector_tools_evaluate.h#L209-L256

The question what prevents users to write their own lamdba functions. Apart from the fact that I am dealing here with the general case (hp), I see two issues 1) a lot of boiler code and 2) the need to work with the CRS data structures directly. 

In the ideal case, the above code could be rewritten in the following way:
```cpp
      const auto fu = [&](const auto &data, auto &values, const auto cell_range) {
        FEPointEvaluation<n_components, dim> evaluator(data, cache.get_mapping(), dof_handler);

        for (unsigned int cell = cell_range.first; cell < cell_range.second; ++cell)
          {
            evaluator.reinit(cell);
            evaluator.read_dof_values(vector);
            evaluator.evaluate(dealii::EvaluationFlags::values);
            evaluator.set_point_values(values);
          }
      };
```
Adding additional evaluators in this context would be easier. But the question is if it is really worth it to move the complexity to `FEPointEvaluation`!?

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,514 @@
...
+     *   result according to the points.
+     */
+    template <int dim, int spacedim = dim>
+    class RemotePointEvaluation
+    {
+    public:
+      /**
+       * Constructor.
+       */
+      RemotePointEvaluation(const double tolerance = 1e-6);
```
kronbichler, line 50: Can you explain what `tolerance` refers to?

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,514 @@
...
+       *   might be the case if a point coincides with a geometric entity (e.g.,
+       *   vertex) that is shared by multiple cells or a point is outside of the
+       *   computational domain.
+       */
+      template <typename T>
+      void
+      evaluate_and_process(
+        std::vector<T> &output,
+        std::vector<T> &buffer,
+        const std::function<void(const ArrayView<T> &, const CellData &)> &fu)
```
kronbichler, line 105: Can you expand the name of `fu` to better describe its intent?

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,514 @@
...
+         * Reference points in the interval [0,1]^dim.
+         */
+        std::vector<Point<dim>> reference_point_values;
+      };
+
+      /**
+       * Evaluate function @p fu in the given  points and triangulation. The
+       * result is stored in @p output.
+       *
+       * @note Is the map of points to cells is not
```
kronbichler, line 93:
```suggestion
       * @note If the map of points to cells is not a
```

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,514 @@
...
+      void
+      evaluate_and_process(
+        std::vector<T> &output,
+        std::vector<T> &buffer,
+        const std::function<void(const ArrayView<T> &, const CellData &)> &fu)
+        const;
+
+      /**
+       * This method is the inverse of the method evaluate_and_process(). It
+       * makes data at the points and provided by @p input available in the
```
kronbichler, line 110:
```suggestion
       * makes the data at the points, provided by @p input, available in the
```

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,514 @@
...
+
+      /**
+       * Return the Mapping object used during reinit().
+       */
+      const Mapping<dim, spacedim> &
+      get_mapping() const;
+
+      /**
+       * Return if the internal data structures have been set up and if yes
+       * if they are still valid (and have not been invalidated due to changes
```
kronbichler, line 151:
```suggestion
       * whether they are still valid (and not invalidated due to changes
```

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,514 @@
...
+
+      /**
+       * Reference to the Mapping object used during reinit().
+       */
+      SmartPointer<const Mapping<dim, spacedim>> mapping;
+
+      /**
+       * MPI communicator of the triangulation.
+       */
+      MPI_Comm comm;
```
kronbichler, line 189: Do we still need this now that it is reasonably simple to query the communication from the grid? Or do we envision cases where we might use another communicator?

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,514 @@
...
+
+      /**
+       * (One-to-one) relation of points and cells.
+       */
+      bool unique_mapping;
+
+      /**
+       * Since for each point multiple or no results can be available, the
+       * pointers in this vector indicate the first and last entries associated
+       * with a point.
```
kronbichler, line 199:
```suggestion
       * with a point in a CRS-like fashion.
```

kronbichler (commented):
I will write more later, but let me push what I already have.

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,518 @@
...
+          MPI_Recv(buffer.data(),
+                   buffer.size(),
+                   MPI_CHAR,
+                   status.MPI_SOURCE,
+                   internal::Tags::remote_point_evaluation,
+                   tria->get_communicator(),
+                   MPI_STATUS_IGNORE);
+
+          temp_recv_map[status.MPI_SOURCE] =
+            Utilities::unpack<std::vector<T>>(buffer);
```
kronbichler, line 337: If we disable compression above, we of course also need to do that here.

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,518 @@
...
+      const std::function<void(const ArrayView<const T> &, const CellData &)>
+        &evaluation_function) const
+    {
+#ifndef DEAL_II_WITH_MPI
+      Assert(false, ExcNeedsMPI());
+      (void)input;
+      (void)buffer;
+      (void)evaluation_function;
+#else
+      // expand
```
kronbichler, line 367: Can you slightly expand the comment here?

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,518 @@
...
+
+      std::vector<MPI_Request> requests;
+      requests.reserve(recv_ranks.size());
+
+      for (const auto recv_rank : recv_ranks)
+        {
+          if (recv_rank == my_rank)
+            continue;
+
+          temp_map[recv_rank] = Utilities::pack(temp_recv_map[recv_rank]);
```
kronbichler, line 423: same here

include/deal.II/numerics/vector_tools_evaluate.h
```diff
@@ -0,0 +1,302 @@
...
+   *
+   * @note Refinement/coarsening/repartitioning leads to the invalidation of the
+   *   cache so that the above function has to be called again.
+   */
+  template <int n_components, int dim, int spacedim, typename VectorType>
+  std::vector<typename FEPointEvaluation<n_components, dim>::value_type>
+  evaluate_at_points(
+    const DoFHandler<dim, spacedim> &                           dof_handler,
+    const VectorType &                                          vector,
+    const Utilities::MPI::RemotePointEvaluation<dim, spacedim> &cache,
```
kronbichler, line 94: I have been thinking about this function again and I think we should have the `cache` variable as the first argument, to avoid confusion with the previous function and the implicit `mapping` argument other functions use in deal.II.

source/base/mpi_remote_point_evaluation.cc
```diff
@@ -0,0 +1,187 @@
...
+{
+  namespace MPI
+  {
+    template <int dim, int spacedim>
+    RemotePointEvaluation<dim, spacedim>::RemotePointEvaluation(
+      const double tolerance)
+      : tolerance(tolerance)
+      , ready_flag(false)
+    {}
+    template <int dim, int spacedim>
```
kronbichler, line 44: Can you add some empty lines her and below as usual in deal.II.

peterrum:
@kronbichler I have incorporated your suggestions!

tests/remove_point_evaluation/vector_tools_evaluate_at_points_01.cc
```diff
@@ -0,0 +1,272 @@
...
+
+      if (n_refinements >= 3)
+        {
+          for (auto cell : tria.active_cell_iterators())
+            if (cell->is_locally_owned())
+              if (0.335 <= cell->center().norm() &&
+                  cell->center().norm() <= 0.39)
+                cell->set_refine_flag();
+          tria.execute_coarsening_and_refinement();
+        }
```
kronbichler, line 92: I suggest to remove these two additional refinements - I think it's enough if you simply add some step to refine that creates sufficiently different parallel partitions in terms of the owned cells not overlapping. This keeps the test shorted (and the compactness is really the key of this class).

kronbichler (approved):
I think this has reached a good state. Again, I have been involved here, so we need someone else looking into it as well.

source/base/mpi_remote_point_evaluation.cc
```diff
@@ -0,0 +1,200 @@
...
+
+          AssertIndexRange(std::get<1>(data.recv_components[i]) + 1,
+                           this->point_ptrs.size());
+          this->point_ptrs[std::get<1>(data.recv_components[i]) + 1]++;
+        }
+
+      unique_mapping = true;
+      for (unsigned int i = 0; i < points.size(); ++i)
+        {
+          unique_mapping &= this->point_ptrs[i + 1] == 1;
```
masterleinad, line 122: I find this pretty hard to read. Why not just
```suggestion
          if (unqiue_mapping && this->point_ptrs[i + 1] != 1)
            unique_mapping = false;
```
?

tests/remote_point_evaluation/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 3.1.0)
```
masterleinad, line 1: Why do you want another directory for the tests here that doesn't match the location of the corresponding source files?

tests/remote_point_evaluation/remote_point_evaluation_03.cc
```diff
@@ -0,0 +1,461 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+// Like remote_point_evaluation_01.cc but normal and curvature vector living on
+// background mesh.
```
masterleinad, line 17: How is this different from `tests/remote_point_evaluation/remote_point_evaluation_02.cc`?

tests/remote_point_evaluation/vector_tools_evaluate_at_points_01.cc
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+  if (false)
+    {
+      print(mapping_1, dof_handler_1, vector_1, 0);
+      print(mapping_2, dof_handler_2, vector_2, 1);
+    }
+
+  vector_1.print(deallog.get_file_stream());
+  vector_2.print(deallog.get_file_stream());
```
masterleinad, line 243: How do you check for correctness here?

tests/remote_point_evaluation/remote_point_evaluation_01.cc
```diff
@@ -0,0 +1,555 @@
...
+                               "force");
+
+      data_out.build_patches(background_mapping, background_fe_degree + 1);
+      data_out.write_vtu_with_pvtu_record("./",
+                                          "data_background",
+                                          0,
+                                          MPI_COMM_WORLD);
+    }
+
+  force_vector_sharp_interface.print(deallog.get_file_stream());
```
masterleinad, line 543: How do you check for correctness?

tests/remote_point_evaluation/remote_point_evaluation_03.cc
peterrum: I have updated the comment. This was just a copy&paste error.

tests/remote_point_evaluation/vector_tools_evaluate_at_points_01.cc
peterrum: I took a look at the result in Paraview. If you want I can replace the printing of the vectors by computing the global error for the following tests: `vector_tools_evaluate_at_points_01`, `vector_tools_evaluate_at_points_03`, `vector_tools_evaluate_at_points_04`, `vector_tools_evaluate_at_points_05`. For the other tests, it is - as far as I know - not possible to come up with some derived quantities.

tests/remote_point_evaluation/remote_point_evaluation_01.cc
peterrum: I took a look at the result in Paraview. Furthermore, I use the vector as the right-hand-vector in a Navier-Stokes solver and the resulting pressure jump (the derived quantity) is correct there.

tests/remote_point_evaluation/CMakeLists.txt
peterrum: I would like to keep it this way to be able to collect the test that use `RemotePointEvaluation` in one place.

include/deal.II/base/mpi_remote_point_evaluation.h
```diff
@@ -0,0 +1,523 @@
...
+
+            for (unsigned int j = 0; j < n_entries; ++j, ++c)
+              buffer_[c] = input[i];
+          }
+
+        // sort data according to the ranks
+        auto it = recv_permutation.begin();
+        for (auto &j : temp_recv_map)
+          for (auto &i : j.second)
+            i = buffer_[*(it++)];
```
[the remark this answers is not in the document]
peterrum: I must admit I much more prefer it like this. The alternative is adding 3 additional lines of code. But I have changed it.

peterrum:
@masterleinad Thanks for the review. I have addressed most of them. I have one question in https://github.com/dealii/dealii/pull/11804#discussion_r595747689.

include/deal.II/base/mpi_remote_point_evaluation.h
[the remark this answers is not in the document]
masterleinad: I personally don't mind too much but others requested me to change similar places so let's just be consistent.

tests/remote_point_evaluation/vector_tools_evaluate_at_points_01.cc
masterleinad: Yes, I would prefer to check for the global error in the tests you mentioned.

masterleinad: After that, it's good from my side.

masterleinad (approved):

tests/remote_point_evaluation/vector_tools_evaluate_at_points_01.cc
peterrum: OK. Done! Thanks again for the review!

PR #11805, opened 2021-02-24
Simplify code of create_geometric_coarsening_sequence()

peterrum:


include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1878,10 +1878,20 @@ namespace MGTransferGlobalCoarseningTools
...
   {
     std::vector<std::shared_ptr<Triangulation<dim, spacedim>>>
-      coarse_grid_triangulations;
+      coarse_grid_triangulations(fine_triangulation_in.n_global_levels());
+
+    coarse_grid_triangulations.back().reset(
+      const_cast<Triangulation<dim, spacedim> *>(&fine_triangulation_in),
+      [](auto &) {
+        // empty deleter, since fine_triangulation_in is an external field
+        // and its destructor is called somewhere else
```
bangerth, line 1887: This is unrelated to the purpose of the patch, but it doesn't make sense to use `shared_ptr` if you don't have a deleter. In that case, it's just a plain C-style pointer and we shouldn't pretend that the last pointer deletes the object by using `shared_ptr`.

peterrum: This is only the finest mesh on basis of which the coarse triangulations are created. It is user controlled. It is put, nevertheless, into the vector together with the new triangulations to improve the usability of the returned vector.

kronbichler (approved):

PR #11806, opened 2021-02-25
examples/step-69: Fix mesh generation

tamiko:
It turns out that this piece of code only ever worked for a precise
ratio of height and disk ratio of 4. If this ratio was violated the
merge_triangulations() call would fail to properly glue the mesh
together due to unmatched vertices.

### DIFF
--- a/examples/step-69/step-69.cc
+++ b/examples/step-69/step-69.cc
@@ -592,7 +592,7 @@ namespace Step69
 
     triangulation.clear();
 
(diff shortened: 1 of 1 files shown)


tamiko:
Thanks to @ejtovar for pointing this out.
@itomasSNL @kronbichler *ping*

peterrum (approved):

kronbichler (approved):

PR #11812, opened 2021-02-26
Added DoFHandler::prepare_for_coarsening_and_refinement().

marcfehling:
Pendant to `Triangulation::prepare_for_coarsening_and_refinement()` for p-adaptation.

This will prevent that neighboring cells have large differences in their levels. For example, situations like `FE_Q(7)` elements facing `FE_Q(1)` elements will be avoided, in which all face dofs would be constrained to the vertex ones.

@bangerth @peterrum -- FYI

### DIFF
--- a/include/deal.II/dofs/dof_handler.h
+++ b/include/deal.II/dofs/dof_handler.h
@@ -702,6 +702,38 @@ public:
   distribute_mg_dofs();
 
   /**
+   * Prepare all cells for p-adaptation in hp-mode. Does nothing in non-hp-mode.
+   *
+   * Essentially does to future FE indices what
+   * Triangulation::prepare_coarsening_and_refinement() does to refinement
+   * flags.
+   *
+   * In detail, this function limits the level difference of neighboring cells
+   * and thus smoothes the overall function space. Future FE indices will be
+   * raised (and never lowered) so that the level difference to neighboring
+   * cells is never larger than @p max_difference.
+   *
+   * Multiple FE hierarchies might have been registered via
+   * hp::FECollection::set_hierarchy(). This function operates on only one
+   * hierarchy, namely the one that contains the FE index @p contains_fe_index.
+   * Cells with future FE indices that are not part of the corresponding
+   * hierarchy will be ignored.
+   *
+   * The function can optionally be called before performing adaptation with
+   * Triangulation::execute_coarsening_and_refinement(). It is not necessary to
+   * call this function, nor will it be automatically invoked in any part of the
+   * library (contrary to its Triangulation counterpart).
+   *
+   * Returns whether any future FE indices have been changed by this function.
+   *
+   * @note Only implemented for serial applications.
+   */
+  bool
+  prepare_coarsening_and_refinement(
+    const unsigned int max_difference    = 1,
+    const unsigned int contains_fe_index = 0) const;
+
+  /**
    * Returns whether this DoFHandler has hp-capabilities.
    */
   bool
--- a/include/deal.II/hp/refinement.h
+++ b/include/deal.II/hp/refinement.h
@@ -147,9 +147,10 @@ namespace hp
      * Each cell flagged for h-refinement will also be flagged for p-refinement.
      * The same applies to coarsening.
      *
-     * @note Triangulation::prepare_coarsening_and_refinement() may change
-     *   refine and coarsen flags. Avoid calling it before this particular
-     *   function.
+     * @note Triangulation::prepare_coarsening_and_refinement() and
+     *   DoFHandler::prepare_coarsening_and_refinement() may change
+     *   refine and coarsen flags as well as future finite element indices.
+     *   Avoid calling them before this particular function.
      */
     template <int dim, int spacedim>
     void
@@ -164,9 +165,10 @@ namespace hp
      * Each entry of the parameter @p p_flags needs to correspond to an active
      * cell.
      *
-     * @note Triangulation::prepare_coarsening_and_refinement() may change
-     *   refine and coarsen flags. Avoid calling it before this particular
-     *   function.
+     * @note Triangulation::prepare_coarsening_and_refinement() and
+     *   DoFHandler::prepare_coarsening_and_refinement() may change
+     *   refine and coarsen flags as well as future finite element indices.
+     *   Avoid calling them before this particular function.
      */
     template <int dim, int spacedim>
     void
@@ -194,9 +196,10 @@ namespace hp
      * Each entry of the parameter @p criteria needs to correspond to an active
      * cell.
      *
-     * @note Triangulation::prepare_coarsening_and_refinement() may change
-     *   refine and coarsen flags. Avoid calling it before this particular
-     *   function.
+     * @note Triangulation::prepare_coarsening_and_refinement() and
+     *   DoFHandler::prepare_coarsening_and_refinement() may change
+     *   refine and coarsen flags as well as future finite element indices.
+     *   Avoid calling them before this particular function.
      */
     template <int dim, typename Number, int spacedim>
     void
@@ -236,9 +239,10 @@ namespace hp
      * cell. Parameters @p p_refine_fraction and @p p_coarsen_fraction need to be
      * in the interval $[0,1]$.
      *
-     * @note Triangulation::prepare_coarsening_and_refinement() may change
-     *   refine and coarsen flags. Avoid calling it before this particular
-     *   function.
+     * @note Triangulation::prepare_coarsening_and_refinement() and
+     *   DoFHandler::prepare_coarsening_and_refinement() may change
+     *   refine and coarsen flags as well as future finite element indices.
+     *   Avoid calling them before this particular function.
      */
     template <int dim, typename Number, int spacedim>
     void
@@ -279,9 +283,10 @@ namespace hp
      * cell. Parameters @p p_refine_fraction and @p p_coarsen_fraction need to be
      * in the interval $[0,1]$.
      *
-     * @note Triangulation::prepare_coarsening_and_refinement() may change
-     *   refine and coarsen flags. Avoid calling it before this particular
-     *   function.
+     * @note Triangulation::prepare_coarsening_and_refinement() and
+     *   DoFHandler::prepare_coarsening_and_refinement() may change
+     *   refine and coarsen flags as well as future finite element indices.
+     *   Avoid calling them before this particular function.
      */
     template <int dim, typename Number, int spacedim>
     void
@@ -316,9 +321,10 @@ namespace hp
      *
      * For more theoretical details see @cite ainsworth1998hp .
      *
-     * @note Triangulation::prepare_coarsening_and_refinement() may change
-     *   refine and coarsen flags. Avoid calling it before this particular
-     *   function.
+     * @note Triangulation::prepare_coarsening_and_refinement() and
+     *   DoFHandler::prepare_coarsening_and_refinement() may change
+     *   refine and coarsen flags as well as future finite element indices.
+     *   Avoid calling them before this particular function.
      */
     template <int dim, typename Number, int spacedim>
     void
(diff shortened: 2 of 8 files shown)


peterrum (commented):
Great! I'll take a look later!

tests/hp/prepare_coarsening_and_refinement_01.cc
```diff
@@ -0,0 +1,114 @@
...
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+// verify restrictions on level differences imposed by
+// DoFHandler::prepare_coarsening_and_refinement()
```
peterrum, line 18: I must admit I don't get test. Wouldn't it be more clear if you would loop over all cells and over all it (face) neighbors and compare/output the level difference between each!?

include/deal.II/dofs/dof_handler.h
```diff
@@ -701,6 +701,31 @@ class DoFHandler : public Subscriptor
...
+  /**
+   * Prepare all cells for p-adaptation. Only available in hp-mode.
+   *
+   * Does to future FE indices what
+   * Triangulation::prepare_coarsening_and_refinement() does to refinement
+   * flags.
+   *
+   * This function operates only on that part of the FE hierarchy
+   * registered with the hp::FECollection that contains the FE index
+   * @p contains_fe_index.
```
peterrum, line 713: I am not sure about the meaning of `contains_fe_index`. Is there other places in the documentation that explains this?

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+
+  if (!hp_capability_enabled)
+    // nothing to do
+    return false;
+
+  // communication of future fe indices on ghost cells necessary.
+  // thus, currently only implemented for serial triangulations.
+  Assert((dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
+            &(*tria)) == nullptr),
+         ExcNotImplemented());
```
peterrum, line 2665: We need this! But let's do it a follow-up PR!

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+
+  //
+  // establish hierarchy
+  //
+  // - create bimap between hierarchy levels and FE indices
+  // - classify cells based on their level
+  const std::vector<unsigned int> fe_on_level =
+    fe_collection.get_hierarchy_sequence(contains_fe_index);
+
+  std::map<unsigned int, unsigned int> level_on_fe;
```
peterrum, line 2675: Interesting name ;)

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+            &(*tria)) == nullptr),
+         ExcNotImplemented());
+
+  //
+  // establish hierarchy
+  //
+  // - create bimap between hierarchy levels and FE indices
+  // - classify cells based on their level
+  const std::vector<unsigned int> fe_on_level =
+    fe_collection.get_hierarchy_sequence(contains_fe_index);
```
peterrum, line 2673: OK. You are running the algorithms only for a single hierarchy.

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+                const unsigned int new_level = level - max_difference;
+                cells_on_level[new_level].push_back(neighbor);
+
+                // update future FE index
+                neighbor->set_future_fe_index(fe_on_level[new_level]);
+                changed_fe_indices = true;
+              }
+          }
+
+  return changed_fe_indices;
```
peterrum, line 2734: I am not sure if it wouldn't be simpler to work on a single `vector` containing the level and having the size of active cells. Actually I think, we could introduce a global vector and use `Triangulation::global_active_cell_index_partitioner()`/`CellAccessor::global_active_cell_index()` (see also https://github.com/dealii/dealii/blob/5c1822a561be1238e0a51f155a6dcf78ca531904/tests/grid/global_ids_01.cc#L43-L48). Working on a vector and updating it is simple. @kronbichler and I wrote these function for such purposes (`MatrixFree` in mind) but we didn't have time to use it yet ourselves (except in one place indirectly: https://github.com/dealii/dealii/blob/b00679e4089197fd1c0c5e7a2210105d21cff9ea/include/deal.II/matrix_free/matrix_free.templates.h#L1748-L1751).

peterrum: I don't feel strongly about what the best approach is. But the update of ghost cells is easy and efficient in this case since the communication pattern has already be setup once (also using a single `GridTools::exchange_cell_data_to_ghosts`, which is called anyway here: https://github.com/dealii/dealii/blob/e694e602704f0c418cd81484972628ca26f33ab2/source/distributed/tria_base.cc#L411-L417).

tests/hp/prepare_coarsening_and_refinement_01.cc
marcfehling: Yes, this would've worked, too. I wanted to provide tests that resemble more a use case, i.e., multiple p-refinements.

It just happens that in this scenario, you could 'measure' the p-level of a cell by the distance to the center. Comparing levels of neighboring cells would have worked, too.

source/dofs/dof_handler.cc
marcfehling: I'll give the global vector a try. But let's tackle this in a follow-up PR!

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+  // establish hierarchy
+  //
+  // - create bimap between hierarchy levels and FE indices
+  // - classify cells based on their level
+  const std::vector<unsigned int> fe_on_level =
+    fe_collection.get_hierarchy_sequence(contains_fe_index);
+
+  std::map<unsigned int, unsigned int> level_on_fe;
+  for (unsigned int i = 0; i < fe_on_level.size(); ++i)
+    level_on_fe[fe_on_level[i]] = i;
```
bangerth, line 2677: I think you're just inverting the permutation here? So maybe `hierarchy_level_for_fe` as a better name?

bangerth: It might be useful to also mention in the (missing :-) ) documentation of the variable that active fe indices not covered in the hierarchy are simply not in the map.

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+    fe_on_level.size());
+  if (fe_on_level.size() == fe_collection.size())
+    {
+      // all FE indices are part of hierarchy
+      for (const auto &cell : active_cell_iterators())
+        cells_on_level[level_on_fe[cell->future_fe_index()]].push_back(cell);
+    }
+  else
+    {
+      // some FE indices are part of hierarchy
```
bangerth, line 2689: only some...

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+      typename decltype(level_on_fe)::iterator iter;
+      for (const auto &cell : active_cell_iterators())
+        if ((iter = level_on_fe.find(cell->future_fe_index())) !=
+            level_on_fe.end())
```
bangerth, line 2693: we need C++17 with init-statement! https://en.cppreference.com/w/cpp/language/if

source/dofs/dof_handler.cc
```diff
@@ -2643,6 +2643,99 @@ DoFHandler<dim, spacedim>::distribute_mg_dofs()
...
+  for (unsigned int level = fe_on_level.size() - 1; level > max_difference;
+       --level)
+    for (const auto &cell : cells_on_level[level])
+      for (unsigned int f = 0; f < cell->n_faces(); ++f)
+        if (cell->face(f)->at_boundary() == false)
+          {
+            const auto neighbor = cell->neighbor(f);
+
+            const unsigned int neighbor_level =
+              level_on_fe[neighbor->future_fe_index()];
```
bangerth, line 2714: what if `level_on_fe` doesn't actually have that future fe index?

bangerth (changes requested):
I don't like the algorithm. It is quadratic in the number of all cells because you loop over all cells and for each one call `std::find` on a `std::list` of cell iterators. You *could* do this cheaper by making the inner `std::list` into a `std::set`, which gets you down to `O(N*log(N))`, but I still don't think that's good enough.

I think that the algorithm should essentially follow the scheme we use for `T::prepare_c_and_r`, which goes like this:

```
    flags_changed = false;
    for (cell : ...)
      for (neighbors of cell)
         if (neighbor->future_fe_index() is too low in the hierarchy 
                  compared to cell->future_fe_index())
           {
             adjust neighbor upward
             flags_changed = true;
           }
```
You might have to iterate this until nothing changes any more, exactly like the function in class `Triangulation`. But this way you get it all to `O(N)` and with `O(1)` memory.

include/deal.II/hp/refinement.h
```diff
@@ -147,9 +147,10 @@ namespace hp
      * Each cell flagged for h-refinement will also be flagged for p-refinement.
      * The same applies to coarsening.
      *
-     * @note Triangulation::prepare_coarsening_and_refinement() may change
-     *   refine and coarsen flags. Avoid calling it before this particular
-     *   function.
+     * @note Triangulation::prepare_coarsening_and_refinement() and
+     *   DoFHandler::prepare_coarsening_and_refinement() may change
+     *   refine and coarsen flags as well as future finite element indices.
+     *   Avoid calling them before this particular function.
```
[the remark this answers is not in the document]
marcfehling: I see it differently: we should decide between h- and p-refinement on the raw indicators, and then call `prepare_c_and_r` to impose smoothing.

source/dofs/dof_handler.cc
marcfehling: I should have thought of this case and didn't. We will ignore those.

include/deal.II/hp/refinement.h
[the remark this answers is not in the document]
marcfehling: ... but I don't feel strongly about this.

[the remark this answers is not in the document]
bangerth: I see. Maybe you can encode this kind of discussion into the comment as you suggest.

peterrum (approved):
Let's proceed here. I am looking forward to the parallel version!

include/deal.II/hp/refinement.h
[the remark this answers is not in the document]
marcfehling: After thinking about this even more, I would like to keep the text as is.

We treat the refinement and coarsening flags as general indicators, independent of the type of adaptation. Applying `T::prepare_c_and_r` _before_ the `hp::Refinement` functions would not fulfill this purpose, as we would silently treat them for h-adaptation. See also general namespace documentation.

bangerth (approved):

bangerth:
Thanks, very nice!

include/deal.II/dofs/dof_handler.h
```diff
@@ -701,6 +701,38 @@ class DoFHandler : public Subscriptor
...
+   * In detail, this function limits the level difference of neighboring cells
+   * and thus smoothes the overall function space. Future FE indices will be
+   * raised (and never lowered) so that the level difference to neighboring
+   * cells is never larger than @p max_difference.
+   *
+   * Multiple FE hierarchies might have been registered via
+   * hp::FECollection::set_hierarchy(). This function operates on only one
+   * hierarchy, namely the one that contains the FE index @p contains_fe_index.
+   * Cells with future FE indices that are not part of the corresponding
+   * hierarchy will be ignored.
```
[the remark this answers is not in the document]
marcfehling: We don't actually have an interface to determine all sequences for a given hierarchy yet. The hierarchy is defined as two functions that determine either the succeeding or preceding FE index to a given one. The `get_hierarchy_sequence` function was meant as a helper function.

To get all sequences in a `FECollection` we could run `get_hierarchy_sequence` on every FE index, and then determine the unique ones among all. I don't think there is currently a need for this.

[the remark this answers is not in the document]
bangerth: That was what I had in mind. In practice, you only need to run the `get_hierarchy_sequence()` for all those indices you haven't seen yet in previous outputs.

PR #11814, opened 2021-02-26
Fix DataAccessOnFaces::values for meshes containing non-hypercube cells

peterrum:
@zjiaqi2018 This fixes your problem (step-67 on simplex mesh) for me. Could you verify it also works for you. Thanks!


kronbichler (approved):

kronbichler:
Can we add step-67 on simplices as a test (or something related), once you have verified that this works now?

zjiaqi2018:
> Can we add step-67 on simplices as a test (or something related), once you have verified that this works now?

Yes, I am working on it now.

peterrum:
> Can we add step-67 on simplices as a test (or something related), once you have verified that this works now?

Let's add this as follow up. @zjiaqi2018 has verified locally that it works.

PR #11815, opened 2021-02-26
Add MGTransferGlobalCoarsening::interpolate_to_mg()

peterrum:


include/deal.II/fe/fe_simplex_p.h
```diff
@@ -60,6 +60,17 @@ class FE_SimplexPoly : public dealii::FE_Poly<dim, spacedim>
...
+  /**
+   * @copydoc dealii::FiniteElement::get_restriction_matrix()
+   *
+   * @note Only implemented for RefinementCase::isotropic_refinement.
+   */
+  virtual const FullMatrix<double> &
+  get_restriction_matrix(
+    const unsigned int         child,
+    const RefinementCase<dim> &refinement_case =
+      RefinementCase<dim>::isotropic_refinement) const override;
```
peterrum, line 72: Moved to #11819.

peterrum:
All dependent PRs have been merged; I have rebased this branch!

include/deal.II/multigrid/mg_transfer_global_coarsening.h
```diff
@@ -122,6 +122,12 @@ class MGTwoLevelTransfer
    */
   void
   restrict_and_add(VectorType &dst, const VectorType &src) const;
+
+  /**
+   * Perform interpolation.
```
kronbichler, line 127: I think we should be more specific to guide the users with less experience regarding the difference.
```suggestion
   * Perform interpolation of a solution vector from the fine level to the coarse level.
   * This function is different from restriction, where a weighted residual is
   * transferred to a coarser level (transposition of prolongation matrix).
```

include/deal.II/multigrid/mg_transfer_global_coarsening.h
```diff
@@ -184,6 +190,13 @@ class MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>
   restrict_and_add(LinearAlgebra::distributed::Vector<Number> &      dst,
                    const LinearAlgebra::distributed::Vector<Number> &src) const;
 
+  /**
+   * Perform interpolation.
```
kronbichler, line 194: Same here.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -394,6 +394,34 @@ namespace internal
       return FETools::get_fe_by_name<1, 1>(fe_name);
     }
 
+    template <int dim, int spacedim>
+    FullMatrix<double>
+    get_restriction_matrix(const FiniteElement<dim, spacedim> &fe,
+                           const unsigned int                  c)
```
kronbichler, line 400: Please expand `c` - is it the `child` index?

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -2192,6 +2294,120 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::
...
+
+              if (scheme.restriction_matrix_1d.size() > 0)
+                cell_transfer.run(cell_restrictor);
+              else
+                cell_restrictor.run_full(n_scalar_dofs_fine,
+                                         n_scalar_dofs_coarse);
+            }
+          // ----------------------------- coarse ----------------------------
+
+
```
kronbichler, line 2389: Can you adjust the whitespace here to make it similar to the one above?

include/deal.II/multigrid/mg_transfer_global_coarsening.h
```diff
@@ -366,6 +389,24 @@ class MGTransferGlobalCoarsening : public dealii::MGTransferBase<VectorType>
                OutVector &                      dst,
                const MGLevelObject<VectorType> &src) const;
 
+  /**
+   * Restrict fine-mesh field @p src to each multigrid level and store the
+   * result in @p dst.
+   *
```
kronbichler, line 395: Can you add some of the text I added to the `interpolate` functions above also here, in particular the part on how this differs from the restriction. Also, I suggest to avoid the term `Restrict` as the leading phrase to avoid confusion with the restriction. I think `interpolate` would be better, but I do not recall the exact wording we use in the solution transfer context where this should also be done.

kronbichler (approved):
Looks very good, apart from comments to explain what happens.

peterrum:
The failing test is unrelated. I will adjust that test later today and reduce the number of MPI processes.

include/deal.II/multigrid/mg_transfer_global_coarsening.h
peterrum: This is consistent with:

https://github.com/dealii/dealii/blob/87a5130c56dd5fb8cf7d57b8f130a45d5a874425/include/deal.II/multigrid/mg_transfer_matrix_free.h#L158-L173

Should I change it there too?

kronbichler: Yes, that would be great!

peterrum:
@kronbichler I have updated the comment at both places!

kronbichler (approved):

PR #11817, opened 2021-02-27
Better document (and name) a ReferenceCell function.

### LINKED ISSUE #11771: Better document ReferenceCell::combine_face_and_line_orientation
We have this function currently:
```
  /**
   * Combine face and line orientation.
   */
  bool
  combine_face_and_line_orientation(const unsigned int  line,
                                    const unsigned char face_orientation,
                                    const unsigned char line_orientation) const;
```
I have to admit that I don't understand what the function does, and I'm also unsure whether its implementation is actually correct for anything other than hexahedra. But for the moment, I'd like to find a better name and documentation for this function: what does the function combine things *to*? It's just a `bool`, but it's not clear from function name or documentation what that `bool` actually *means*.

@peterrum -- do you have any insight?

bangerth:
Related to #11544. Fixes #11771.

I don't have the battery power right now to run this through a compile cycle, so let's just see what the testers have to say about whether the function is used in only this one place, or in other places as well.

/rebuild


include/deal.II/grid/tria_accessor.templates.h
```diff
@@ -832,17 +832,21 @@ namespace internal
...
           accessor.reference_cell().standard_to_real_face_line(
             pair[1], pair[0], face_orientation_raw(accessor, quad_index));
 
+        //
         return accessor.reference_cell().combine_face_and_line_orientation(
           pair[1],
           face_orientation_raw(accessor, quad_index),
-          accessor.quad(quad_index)->line_orientation(line_index));
+          accessor.quad(quad_index)
+            ->standard_vs_true_line_orientation(line_within_face_index));
```
peterrum, line 849: I think you replaced the wrong word.

include/deal.II/grid/tria_accessor.h
```diff
@@ -924,6 +924,8 @@ class TriaAccessor : public TriaAccessorBase<structdim, dim, spacedim>
    *
    * This function is really only for internal use in the library unless you
    * absolutely know what this is all about.
+   *
+   * This function queries ReferenceCell::standard_vs_true_line_orientation().
```
peterrum, line 928: Why do we need this info?

include/deal.II/grid/tria_accessor.templates.h
bangerth: Yes.

include/deal.II/grid/tria_accessor.h
bangerth: I probably don't *need* it, but I thought it would be useful to make the context of that function clearer by pointing to another place that explains what we are trying to do.

include/deal.II/grid/reference_cell.h
```diff
@@ -380,10 +380,17 @@ class ReferenceCell
                              const unsigned char face_orientation) const;
 
   /**
-   * Combine face and line orientation.
+   * Return whether the line with index @p line is oriented in
+   * standard direction within a cell, give the @p face_orientation of
```
kronbichler, line 384:
```suggestion
   * standard direction within a cell, given the @p face_orientation of
```

bangerth:
@kronbichler Fixed

drwells (approved):

bangerth:
Ping?

peterrum (approved):

Issue #10731, opened 2020-07-20
ParticleHandler does not work properly with parallel::shared::Triangulation


luca-heltai:
ParticleHandler checks if a Triangulation is parallel::distributed to decide whether to use MPI communication or not. 

This fails miserably with parallel::shared::Triangulation objects, since the dynamic cast in this case fails, and the Triangulation is treated as serial.

@gassmoeller @fmilicchio

peterrum:
Can you cast it to `parallel::TriangulationBase`?

drwells:
This seems like a good chance to clean up the way we do parallel detection. Why not add `Triangulation::get_communicator()` (which always returns `MPI_COMM_SELF` and make the communication check both for particles and elsewhere "does this object use MPI_COMM_SELF"?

peterrum:
> Why not add Triangulation::get_communicator()

You have my support here. My workaround in many places is like this:
https://github.com/dealii/dealii/blob/3a9d086947af2f028476e3288fcc8d0e9c6a7965/include/deal.II/multigrid/mg_transfer_interface_util.templates.h#L52-L63

drwells:
We should definitely have a solution that doesn't require `dynamic_cast`. I'll put up a PR a little later today.

drwells:
Now that we have a proper fix can someone go back and fix these for particles?

blaisb:
> Now that we have a proper fix can someone go back and fix these for particles?

@drwells Actually particle_handler now work with shared triangulation. There is even a test for it:
https://github.com/dealii/dealii/blob/master/tests/particles/particle_handler_shared_01.cc

I wrote it a few months ago...

drwells:
Oh, even better. Since this is fixed we can close the issue.

PR #11822, opened 2021-02-28
Add Triangulation/DoFHandler::get_communicator()

### LINKED ISSUE #10733: Find a better way to determine what needs to be communicated with Triangulation classes
Spin-off from #10731: I think we should first discuss this before implementing a solution.

There are two categories of problems that keep popping up from time to time related to communication in `Triangulation`s:
1. Its not clear if we need to communicate data if we only have a `dealii::Triangulation &`
2. Its not clear what kind of communication needs to happen at all since the parallel Triangulation classes lie on a spectrum from `shared` to `fullydistributed`

Our usual technique for figuring out which type of `Triangulation` we have is to use `dynamic_cast`. In particular, we usually assume a Triangulation needs to communicate (category 1) if it inherits from `parallel::TriangulationBase` (exceptions include `minimal_cell_diameter`, which arguably does the wrong thing for `parallel::shared::Triangulation` as a result).

`dynamic_cast` is problematic because we now have three parallel Triangulation classes and there are still some problems with things being supported for `p:d:t` and not `p:f:t`, or some things working for both and `dynamic_cast` being overly restrictive (category 2). @peterrum mentioned to me that ideally these problems would only occur in the DoFHandler policy classes: to me, that indicates that we should improve the API elsewhere to get rid of some casts.

We could fix some problems in the first category by adding `Triangulation::get_communicator()`, which would always return `MPI_COMM_SELF`: calling code could then check if communication is necessary by testing `tria.get_communicator() != MPI_COMM_SELF`. One problem with this solution is that it doesn't make sense to add MPI functions to a purely serial class. It also doesn't address the second category at all.

I don't know the answer, but I think its worth discussing some possible fixes because of the number of PRs opened recently related to these two categories of problems.

peterrum:
Fixes #10733 (edit by @drwells to get automatic closure)

### DIFF
--- a/doc/news/changes/major/20210228Munch
+++ b/doc/news/changes/major/20210228Munch
@@ -1,0 +1,6 @@
+New: The communicator of an arbitrary (not just parallel) Triangulation class can now be 
+queried via Triangulation::get_communicator() or DoFHandler::get_communicator(). In
+the case of serial Triangulations and DoFHandler set up with serial Triangulations,
+MPI_COMM_SELF is returned.
+<br>
+(Peter Munch, 2021/02/28)
--- a/include/deal.II/grid/tria.h
+++ b/include/deal.II/grid/tria.h
@@ -1615,6 +1615,13 @@ public:
   clear();
 
   /**
+   * Return MPI communicator used by this triangulation. In the case of
+   * a serial Triangulation object, MPI_COMM_SELF is returned.
+   */
+  virtual MPI_Comm
+  get_communicator() const;
+
+  /**
    * Set the mesh smoothing to @p mesh_smoothing. This overrides the
    * MeshSmoothing given to the constructor. It is allowed to call this
    * function only if the triangulation is empty.
--- a/source/distributed/tria_base.cc
+++ b/source/distributed/tria_base.cc
@@ -90,7 +90,7 @@ namespace parallel
   {
     std::size_t mem =
       this->dealii::Triangulation<dim, spacedim>::memory_consumption() +
-      MemoryConsumption::memory_consumption(mpi_communicator) +
+      MemoryConsumption::memory_consumption(this->mpi_communicator) +
       MemoryConsumption::memory_consumption(my_subdomain) +
       MemoryConsumption::memory_consumption(
         number_cache.n_global_active_cells) +
@@ -135,7 +135,7 @@ namespace parallel
   }
 
   template <int dim, int spacedim>
-  const MPI_Comm &
+  MPI_Comm
   TriangulationBase<dim, spacedim>::get_communicator() const
   {
     return mpi_communicator;
@@ -170,7 +170,7 @@ namespace parallel
           number_cache.ghost_owners.insert(cell->subdomain_id());
 
       Assert(number_cache.ghost_owners.size() <
-               Utilities::MPI::n_mpi_processes(mpi_communicator),
+               Utilities::MPI::n_mpi_processes(this->mpi_communicator),
              ExcInternalError());
     }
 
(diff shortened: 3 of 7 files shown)


include/deal.II/grid/tria.h
```diff
@@ -3513,6 +3527,13 @@ class Triangulation : public Subscriptor
    */
   std::vector<ReferenceCell> reference_cells;
 
+  /**
+   * Constructor to be used only by parallel Triangulation classes.
+   */
+  Triangulation(const MPI_Comm &    mpi_communicator,
```
bangerth, line 3533: I think this is the wrong approach. The class is serial, it should not have a constructor that takes a communicator, nor should it store a communicator.

source/distributed/tria_base.cc
```diff
@@ -90,7 +90,7 @@ namespace parallel
   {
     std::size_t mem =
       this->dealii::Triangulation<dim, spacedim>::memory_consumption() +
-      MemoryConsumption::memory_consumption(mpi_communicator) +
+      MemoryConsumption::memory_consumption(this->mpi_communicator) +
```
bangerth, line 93: This should go into the base class (except I don't think the variable should actually live in the base class).

bangerth (changes requested):
I don't like this patch. The `::Triangulation` class is serial -- it shouldn't store a communicator, and its constructor should not take a communicator. This should remain with the `p::TriangulationBase` class where it belongs.

I'm not opposed to introducing a virtual function in the base class that returns a communicator. The implementation in `::T` would then simply return `MPI_COMM_SELF`, which you can easily do if you return a communicator instead of a const reference. The derived classes can then return whatever actual communicator they store.

source/grid/tria.cc
```diff
@@ -10190,6 +10190,16 @@ Triangulation<dim, spacedim>::clear()
...
 
 
+template <int dim, int spacedim>
+const MPI_Comm &
+Triangulation<dim, spacedim>::get_communicator() const
+{
+  static const MPI_Comm comm = MPI_COMM_SELF;
+
+  return comm;
+}
```
peterrum, line 10200: @bangerth Would this be fine? Or do you prefer that we remove the reference?

bangerth: That works as well. `MPI_Comm` objects are copyable (they must be -- they need to work in C, in which all data types are copyable), but we typically pass them by reference. So this is fine with me.

kronbichler: Given that we do not store a communicator, I would lean towards returning `MPI_Comm` by value. (Background: I think we should keep global static variables low and only do if necessary. The global startup part of deal.II called before main is very big already.)
However, this technically breaks the backwards compatibility of the derived parallel classes, which used to take a reference. Given the typical use, I would still vote for returning by value, but I would be fine either way.

peterrum: I have changed it so that it return by value. BTW: my first implementation (pre-PR version) had a dummy communicator inside of `Triangulation`.

doc/news/changes/major/20210228Munch
```diff
@@ -0,0 +1,5 @@
+New: A serial triangulation owns now per default a MPI_COMM_SELF communicator. The 
+communicator of an arbitrary (not just parallel) Triangulation class can now be
```
bangerth, line 2: You'll have to update this bit

bangerth (approved):
Just the one issue with the changelog entry. Otherwise ok!

doc/news/changes/major/20210228Munch
peterrum: Changed! Thanks for the review!

bangerth (approved):

drwells:
This also fixes #10733 and will help with #10731.

kronbichler (approved):

PR #11828, opened 2021-02-28
Add GridGenerator::MarchingCubeAlgorithm

peterrum:
The marching-square and marching-cube algorithm allow to construct surface meshes based on iso-lines/surfaces (e.g. the zero value of a level-set field). This PR proposes a class that implements that algorithm ~(for now only in 2D)~ and creates a codim-1 mesh.

The nice thing about the 3D version is that the result consist only of triangles (see http://users.polytech.unice.fr/~lingrand/MarchingCubes/algo.html) so that it fits well in our simplex-support efforts!

~*The implementation is still work in progress (vertices are duplicated, since in our use case, we did't need the actual mesh) but we could discuss if the interface matches other immersed-surface applications than sharp-interface methods.*~

A 3D exmpale:

mc-3D

with 1-3 subdivisions per cell.


source/grid/grid_generator.cc
```diff
@@ -8410,6 +8415,643 @@ namespace GridGenerator
...
+    std::vector<Point<dim>>        vertices;
+    std::vector<CellData<dim - 1>> cells;
+    SubCellData                    subcelldata;
+
+    mc.process(background_dof_handler, ls_vector, vertices, cells);
+
+    std::vector<unsigned int> considered_vertices;
+
+    // GridTools::delete_duplicated_vertices (vertices, cells, subcelldata,
+    // considered_vertices);
```
peterrum, line 9050: For some reasons, I have some issue when merging the triangles... But that is something for later.

peterrum (commented):
PR is ready from my side (and also a 3D version).

include/deal.II/grid/grid_generator.h
```diff
@@ -2360,6 +2364,143 @@ namespace GridGenerator
...
+
+
+  /**
+   * Create a codim-1 triangulation based on a iso-line/surface and the
+   * marching-cube algorithm (MarchingCubeAlgorithm).
+   */
+  template <int dim, typename VectorType>
+  void
+  create_triangulation_with_marching_cube_algorithm(
+    const MarchingCubeAlgorithm<dim, VectorType> &mc,
```
bangerth, line 2498: I have not actually looked at the implementation, but am unsure about this choice of interface. The `MCA` class has no virtual functions, so I assume that the first argument given here is simply of type `MCA` and not derived. If that were so, why would one require that the user creates such an object, rather than letting the user pass the necessary arguments to this function and create the object internally? If that were so, that whole class could be moved to the `.cc` file.

peterrum: > If that were so, why would one require that the user creates such an object, rather than letting the user pass the necessary arguments to this function and create the object internally?

I guess I can do that.

> If that were so, that whole class could be moved to the .cc file.

The reason for having the class being accessible by the user is that you might not want to construct a global surface grid but only a surface grid in a cell, do some postprocessing, and throw the grid away.

Personally, I am using the class this way to compute surface tensions in the case of sharp-interface methods in [adaflo](https://github.com/kronbichler/adaflo) (see: https://github.com/kronbichler/adaflo/blob/7b60179059f0cfd3ec30e7960b563595e9b0b186/include/adaflo/sharp_interface_util.h#L968-L1122):

<p align="center">
 
</p>

 and @mschreter is using it to perform some surface integrals in the context of evaporation in [MeltPoolDG](https://github.com/MeltPoolDG/MeltPoolDG) (see: https://github.com/MeltPoolDG/MeltPoolDG/blob/ba9544808e2ad495424efe1317a2649bcea221aa/include/meltpooldg/evaporation/evaporation_operation_marching_cube.hpp#L153-L291).

peterrum:
I would love to get the marching-cube algorithm into the library before the release. Is that possible?

include/deal.II/grid/grid_generator.h
```diff
@@ -2360,6 +2364,143 @@ namespace GridGenerator
...
+    create_qudrature_rule(const unsigned int n_subdivisions);
+
+    /**
+     * Process a sub-cell (2D).
+     *
+     * @note Subcells with saddle points are ignored. Please increase the number
+     *   of subdivisions in this case.
+     */
+    static void
+    process_sub_cell(const std::vector<value_type> & ls_values,
```
Rombur, line 2460: Why is this function static?

include/deal.II/grid/grid_generator.h
```diff
@@ -2360,6 +2364,143 @@ namespace GridGenerator
...
+                 std::vector<Point<dim>> &       vertices,
+                 std::vector<CellData<dim - 1>> &cells) const;
+
+  private:
+    /**
+     * Internal function to create a quadrature rule with n_subdivisions+1
+     * equally-positioned quadrature points.
+     */
+    static Quadrature<dim>
+    create_qudrature_rule(const unsigned int n_subdivisions);
```
Rombur, line 2451:
```suggestion
    create_quadrature_rule(const unsigned int n_subdivisions);
```

include/deal.II/grid/grid_generator.h
Rombur: I don't understand how the code snippets support the current interface of `create_triangulation_with_marching_cube_algorithm` It seems to me that either I want to do something special and I need to use `MCA` but I don't need `create_triangulation_with_marching_cube_algorithm` or I want something more generic and I don't need to know that `MCA` exists

include/deal.II/grid/grid_generator.h
```diff
@@ -2360,6 +2364,143 @@ namespace GridGenerator
...
+                     const std::vector<unsigned int> mask,
+                     const double                    iso_level,
+                     std::vector<Point<2>> &         vertices,
+                     std::vector<CellData<1>> &      cells);
+
+    /**
+     * Process a sub-cell (3D).
+     */
+    static void
+    process_sub_cell(const std::vector<value_type> & ls_values,
```
Rombur, line 2471: Why can't this be templated on `dim`?

Rombur (commented):
I haven't read through the implementation yet. I think the interface still needs some work.

include/deal.II/grid/grid_generator.h
peterrum: @Rombur Thanks for the review till now. I have incorporated the changes. However, I would like to keep this function as it is. It is private and my fear is that (as long as we don't support if constexpr) we will get warnings which would have to "fixed" in an other way.

source/grid/grid_generator.cc
```diff
@@ -8439,6 +8444,674 @@ namespace GridGenerator
...
+       {{0, 2, 11, 8, 0, 11, X, X, X, X, X, X, X, X, X, X}},
+       {{3, 2, 11, X, X, X, X, X, X, X, X, X, X, X, X, X}},
+       {{2, 3, 8, 2, 8, 10, 10, 8, 9, X, X, X, X, X, X, X}},
+       {{9, 10, 2, 0, 9, 2, X, X, X, X, X, X, X, X, X, X}},
+       {{2, 3, 8, 2, 8, 10, 0, 1, 8, 1, 10, 8, X, X, X, X}},
+       {{1, 10, 2, X, X, X, X, X, X, X, X, X, X, X, X, X}},
+       {{1, 3, 8, 9, 1, 8, X, X, X, X, X, X, X, X, X, X}},
+       {{0, 9, 1, X, X, X, X, X, X, X, X, X, X, X, X, X}},
+       {{0, 3, 8, X, X, X, X, X, X, X, X, X, X, X, X, X}},
+       {{X, X, X, X, X, X, X, X, X, X, X, X, X, X, X, X}}}};
```
Rombur, line 9053: I hope that you have good test because if there is a typo in these tables we will never be able to find it.

source/grid/grid_generator.cc
Rombur: Isn't that a problem though? Let's say that two vertices that form an edge are duplicated. Don't you get two edges? What happens if you refine one cell but not the other? Does everything still works at the hanging node?

tests/grid/grid_generator_marching_cube_algorithm_01.cc
```diff
@@ -0,0 +1,210 @@
...
+  GridGenerator::create_triangulation_with_marching_cube_algorithm(
+    background_mapping,
+    background_dof_handler,
+    ls_vector,
+    iso_level,
+    n_subdivisions,
+    tria);
+
+  // write computed vectors to Paraview
+  if (true)
```
Rombur, line 163: Can you give these `true/false` a name. This is very confusing to read.

Rombur (commented):
It's starting to looks better. I am not sure that `MCA` should be in `GridGenerator` maybe it would be better in something like `GridTools`.

source/grid/grid_generator.cc
```diff
@@ -8439,6 +8444,674 @@ namespace GridGenerator
...
+    // set up dimension-dependent sizes and tables
+    static constexpr unsigned int n_vertices     = 4;
+    static constexpr unsigned int n_sub_vertices = 2;
+    static constexpr unsigned int n_lines        = 4;
+    static constexpr unsigned int n_configurations =
+      Utilities::pow(2, n_vertices);
+    static constexpr unsigned int X = static_cast<unsigned int>(-1);
+
+    static constexpr std::array<unsigned int, n_configurations>
+      configuration_table = {{0b0000,
```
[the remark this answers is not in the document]
peterrum: I have renamed it and added a comment.

source/grid/grid_generator.cc
```diff
@@ -8439,6 +8444,674 @@ namespace GridGenerator
...
+      // cell is not cut (nothing to do)
+      if (configuration_table[configuration] == 0)
+        return;
+
+      // helper function to determine where an edge (between index i and j) is
+      // cut - see also: http://paulbourke.net/geometry/polygonise/
+      const auto interpolate = [&](const unsigned int i, const unsigned int j) {
+        const double tolerance = 1e-10;
+
+        if (std::abs(iso_level - ls_values[mask[i]]) < tolerance)
```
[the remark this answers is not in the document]
peterrum: I have made it a parameter.

source/grid/grid_generator.cc
peterrum: I have not tested all configuration. I hope that the source did not do any mistakes^^

tests/grid/grid_generator_marching_cube_algorithm_01.cc
peterrum: OK

include/deal.II/grid/grid_generator.h
```diff
@@ -2400,6 +2404,134 @@ namespace GridGenerator
...
+    /**
+     * Constructor.
+     */
+    MarchingCubeAlgorithm(const Mapping<dim, dim> &      mapping,
+                          const FiniteElement<dim, dim> &fe,
+                          const unsigned int             n_subdivisions = 1);
+
+    /**
+     * Process all locally-owned cells.
+     */
```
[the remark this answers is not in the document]
peterrum: I have amended the documentation to state that the class does not actually create a triangulation but returns the ingredients to do so.

include/deal.II/grid/grid_generator.h
```diff
@@ -2400,6 +2404,134 @@ namespace GridGenerator
...
+    void
+    process_sub_cell(const std::vector<value_type> & ls_values,
+                     const std::vector<Point<3>> &   points,
+                     const std::vector<unsigned int> mask,
+                     const double                    iso_level,
+                     std::vector<Point<3>> &         vertices,
+                     std::vector<CellData<2>> &      cells) const;
+
+    /**
+     * Number of subdivisions defined in the constructor.
```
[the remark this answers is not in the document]
peterrum: Comment added.

peterrum:
> Isn't that a problem though? Let's say that two vertices that form an edge are duplicated. Don't you get two edges? What happens if you refine one cell but not the other? Does everything still works at the hanging node?

Yes, this is a limitation. But I don't think the resulting mesh will be used for computation but rather for in-situ post processing or evaluate some sharp interface contributions that do not require that the mesh is in one piece.

Rombur:
> Yes, this is a limitation. But I don't think the resulting mesh will be used for computation but rather for in-situ post processing or evaluate some sharp interface contributions that do not require that the mesh is in one piece.

What about we postpone `create_triangulation_with_marching_cube_algorithm` to after the release and just merge `MCA`? The codes that you showed didn't use `create_triangulation_with_marching_cube_algorithm` only `MCA`.

peterrum:
> What about we postpone create_triangulation_with_marching_cube_algorithm to after the release and just merge MCA? The codes that you showed didn't use create_triangulation_with_marching_cube_algorithm only MCA.

That sounds me good.

peterrum:
> What about we postpone create_triangulation_with_marching_cube_algorithm to after the release and just merge MCA? The codes that you showed didn't use create_triangulation_with_marching_cube_algorithm only MCA.

@Rombur Done!

doc/news/changes/minor/20210228Munch
```diff
@@ -0,0 +1,5 @@
+New: A new class GridGenerator::MarchingCubeAlgorithm has been added. It
```
Rombur, line 1: Since it doesn't create a new Triangulation, it shouldn't be in `GridGenerator`. Maybe `GridTools` would be better?

include/deal.II/grid/grid_generator.h
```diff
@@ -2400,6 +2404,127 @@ namespace GridGenerator
...
+    /**
+     * Constructor.
+     */
+    MarchingCubeAlgorithm(const Mapping<dim, dim> &      mapping,
+                          const FiniteElement<dim, dim> &fe,
+                          const unsigned int             n_subdivisions = 1,
+                          const double                   tolerance = 1e-10);
+
+    /**
+     * Process all locally-owned cells and fill @p vertices and @p for all cells
```
Rombur, line 2443:
```suggestion
     * Process all locally-owned cells and fill @p vertices and @p cells for all cells
```

include/deal.II/grid/grid_generator.h
```diff
@@ -2400,6 +2404,127 @@ namespace GridGenerator
...
+     */
+    void
+    process(const DoFHandler<dim> &         background_dof_handler,
+            const VectorType &              ls_vector,
+            const double                    iso_level,
+            std::vector<Point<dim>> &       vertices,
+            std::vector<CellData<dim - 1>> &cells) const;
+
+    /**
+     * Process the provided cell and fill @p vertices and @p for all cells
```
Rombur, line 2454:
```suggestion
     * Process the provided cell and fill @p vertices and @p cells for all cells
```

source/grid/grid_generator.cc
```diff
@@ -8439,6 +8444,661 @@ namespace GridGenerator
...
+      const std::vector<Point<dim>> &          points,
+      const std::vector<unsigned int> &        mask,
+      const double                             iso_level,
+      const double                             tolerance,
+      std::vector<Point<dim>> &                vertices,
+      std::vector<CellData<dim - 1>> &         cells)
+    {
+      // inspired by https://graphics.stanford.edu/~mdfisher/MarchingCubes.html
+
+      static constexpr unsigned int X = static_cast<unsigned int>(-1);
```
Rombur, line 8604: Why is this variable `static`?

source/grid/grid_generator.cc
```diff
@@ -8439,6 +8444,661 @@ namespace GridGenerator
...
+  MarchingCubeAlgorithm<dim, VectorType>::process_sub_cell(
+    const std::vector<value_type> & ls_values,
+    const std::vector<Point<2>> &   points,
+    const std::vector<unsigned int> mask,
+    const double                    iso_level,
+    std::vector<Point<2>> &         vertices,
+    std::vector<CellData<1>> &      cells) const
+  {
+    // set up dimension-dependent sizes and tables
+    static constexpr unsigned int n_vertices     = 4;
```
Rombur, line 8687: All the variables are `static` but I don't get why.

Rombur (commented):
I think we are almost there. Thanks for making the changes.

peterrum:
@Rombur I have made the changes.

source/grid/grid_tools.cc
```diff
@@ -6596,6 +6596,658 @@ namespace GridTools
...
+      const std::vector<Point<dim>> &          points,
+      const std::vector<unsigned int> &        mask,
+      const double                             iso_level,
+      const double                             tolerance,
+      std::vector<Point<dim>> &                vertices,
+      std::vector<CellData<dim - 1>> &         cells)
+    {
+      // inspired by https://graphics.stanford.edu/~mdfisher/MarchingCubes.html
+
+      static constexpr unsigned int X = static_cast<unsigned int>(-1);
```
Rombur, line 6756:
```suggestion
      constexpr unsigned int X = static_cast<unsigned int>(-1);
```

Rombur (approved):
Please fix the last comment and then let's merge this.

PR #11830, opened 2021-03-01
Fix FESeries with non-primitive FEs.

### LINKED ISSUE #11791: FESeries::Fourier/Legendre do not work with non-primitive FEs
It seems that `FESeries::Fourier/Legendre` do not work with non-primitive FEs that derived from `FE_PolyTensor`.

Both Fourier and Legendre called `shape_value`,
https://github.com/dealii/dealii/blob/4cfd36de738c221932d741866d5fe62ba114c476/source/fe/fe_series_fourier.cc#L75
https://github.com/dealii/dealii/blob/4cfd36de738c221932d741866d5fe62ba114c476/source/fe/fe_series_legendre.cc#L84
but the `shape_value` in `FE_PolyTensor` is not implemented.
https://github.com/dealii/dealii/blob/4cfd36de738c221932d741866d5fe62ba114c476/source/fe/fe_poly_tensor.cc#L233

The following MWE will run into assertion failure.
```

#include <deal.II/fe/fe_series.h>
#include <deal.II/fe/fe_nedelec.h>

#include <deal.II/hp/fe_collection.h>

#include <deal.II/lac/vector.h>

#include <deal.II/numerics/smoothness_estimator.h>

int main() {
  dealii::hp::FECollection<3> fe_collection;
  fe_collection.push_back(dealii::FE_Nedelec<3>(0));
  fe_collection.push_back(dealii::FE_Nedelec<3>(1));

  dealii::Triangulation<3> tria;
  dealii::GridGenerator::subdivided_hyper_cube(tria, 2);

  dealii::DoFHandler<3> dh(tria);
  dh.distribute_dofs(fe_collection);

  dealii::Vector<double> solution(dh.n_dofs());
  dealii::Vector<float> smoothness(tria.n_active_cells());

  dealii::FESeries::Fourier<3> fourier = dealii::SmoothnessEstimator::Fourier::default_fe_series(fe_collection);
  dealii::SmoothnessEstimator::Fourier::coefficient_decay(fourier, dh, solution, smoothness);

  return 0;
}
```

```
An error occurred in line <233> of file </.../source/fe/fe_poly_tensor.cc> in function
    virtual double dealii::FE_PolyTensor<3, 3>::shape_value(const unsigned int, const Point<dim> &) const [dim = 3, spacedim = 3]
The violated condition was: 
    false
Additional information: 
    (none)

Stacktrace:
-----------
#0  2   libdeal_II.g.9.3.0-pre.dylib        0x00000001156d2a24 _ZNK6dealii13FE_PolyTensorILi3ELi3EE11shape_valueEjRKNS_5PointILi3EdEE + 84: 2   libdeal_II.g.9.3.0-pre.dylib        0x00000001156d2a24 _ZNK6dealii13FE_PolyTensorILi3ELi3EE11shape_valueEjRKNS_5PointILi3EdEE 
#1  3   libdeal_II.g.9.3.0-pre.dylib        0x00000001157c402a _ZN6dealii12_GLOBAL__N_19integrateILi3ELi3EEENSt3__17complexIdEERKNS_13FiniteElementIXT_EXT0_EEERKNS_10QuadratureIXT_EEERKNS_6TensorILi1EXT_EdEEj + 266: 3   libdeal_II.g.9.3.0-pre.dylib        0x00000001157c402a _ZN6dealii12_GLOBAL__N_19integrateILi3ELi3EEENSt3__17complexIdEERKNS_13FiniteElementIXT_EXT0_EEERKNS_10QuadratureIXT_EEERKNS_6TensorILi1EXT_EdEEj 
#2  4   libdeal_II.g.9.3.0-pre.dylib        0x00000001157b9302 _ZN6dealii12_GLOBAL__N_116ensure_existenceILi3EEEvRKNSt3__16vectorIjNS2_9allocatorIjEEEERKNS_2hp12FECollectionILi3EXT_EEERKNS9_11QCollectionILi3EEERKNS_5TableILi3ENS_6TensorILi1ELi3EdEEEEjRNS3_INS_10FullMatrixINS2_7complexIdEEEENS4_ISR_EEEE + 498: 4   libdeal_II.g.9.3.0-pre.dylib        0x00000001157b9302 _ZN6dealii12_GLOBAL__N_116ensure_existenceILi3EEEvRKNSt3__16vectorIjNS2_9allocatorIjEEEERKNS_2hp12FECollectionILi3EXT_EEERKNS9_11QCollectionILi3EEERKNS_5TableILi3ENS_6TensorILi1ELi3EdEEEEjRNS3_INS_10FullMatrixINS2_7complexIdEEEENS4_ISR_EEEE 
#3  5   libdeal_II.g.9.3.0-pre.dylib        0x00000001157b8dfe _ZN6dealii8FESeries7FourierILi3ELi3EE9calculateIdEEvRKNS_6VectorIT_EEjRNS_5TableILi3ENSt3__17complexIdEEEE + 142: 5   libdeal_II.g.9.3.0-pre.dylib        0x00000001157b8dfe _ZN6dealii8FESeries7FourierILi3ELi3EE9calculateIdEEvRKNS_6VectorIT_EEjRNS_5TableILi3ENSt3__17complexIdEEEE 
#4  6   libdeal_II.g.9.3.0-pre.dylib        0x0000000114ac4dae _ZN6dealii19SmoothnessEstimator7Fourier17coefficient_decayILi3ELi3ENS_6VectorIdEEEEvRNS_8FESeries7FourierIXT_EXT0_EEERKNS_10DoFHandlerIXT_EXT0_EEERKT1_RNS3_IfEENS_11VectorTools8NormTypeEdb + 542: 6   libdeal_II.g.9.3.0-pre.dylib        0x0000000114ac4dae _ZN6dealii19SmoothnessEstimator7Fourier17coefficient_decayILi3ELi3ENS_6VectorIdEEEEvRNS_8FESeries7FourierIXT_EXT0_EEERKNS_10DoFHandlerIXT_EXT0_EEERKT1_RNS3_IfEENS_11VectorTools8NormTypeEdb 
#5  7   feseries-test                       0x0000000108a54310 main + 336: 7   feseries-test                       0x0000000108a54310 main 
#6  8   libdyld.dylib                       0x00007fff72da5cc9 start + 1: 8   libdyld.dylib                       0x00007fff72da5cc9 start 
```

Maybe we should call `shape_value_component` instead of `shape_value`?

adamqc:
Closes #11791.

Add a component parameter to FESeries::Fourier/Legendre to
make them working with non-primitive elements.

### DIFF
--- a/include/deal.II/fe/fe_series.h
+++ b/include/deal.II/fe/fe_series.h
@@ -55,9 +55,10 @@ DEAL_II_NAMESPACE_OPEN
 namespace FESeries
 {
   /**
-   * A class to calculate expansion of a scalar FE field into Fourier series
-   * on a reference element. The exponential form of the Fourier series is
-   * based on completeness and Hermitian orthogonality of the set of exponential
+   * A class to calculate expansion of a scalar FE (or a single component
+   * of vector-valued FE) field into Fourier series on a reference element.
+   * The exponential form of the Fourier series is  based on completeness
+   * and Hermitian orthogonality of the set of exponential
    * functions $ \phi_{\bf k}({\bf x}) = \exp(2 \pi i\, {\bf k} \cdot {\bf x})$.
    * For example in 1D the L2-orthogonality condition reads
    * @f[
@@ -96,10 +97,21 @@ namespace FESeries
      * each direction, @p fe_collection is the hp::FECollection for which
      * expansion will be used and @p q_collection is the hp::QCollection used to
      * integrate the expansion for each FiniteElement in @p fe_collection.
+     *
+     * As the Fourier expansion can only be performed on scalar fields, this
+     * class does not operate on vector-valued finite elements and will
+     * therefore throw an assertion. However, each component of a finite element
+     * field can be treated as a scalar field, respectively, on which Fourier
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
+     * The default value of @p component only applies to scalar FEs, in which
+     * case it indicates that the sole component is to be decomposed. For
+     * vector-valued FEs, a non-default value must be explicitly provided.
      */
     Fourier(const std::vector<unsigned int> &      n_coefficients_per_direction,
             const hp::FECollection<dim, spacedim> &fe_collection,
-            const hp::QCollection<dim> &           q_collection);
+            const hp::QCollection<dim> &           q_collection,
+            const unsigned int component = numbers::invalid_unsigned_int);
 
     /**
      * A non-default constructor. The @p n_coefficients_per_direction defines the
@@ -204,13 +216,20 @@ namespace FESeries
      * Auxiliary vector to store unrolled coefficients.
      */
     std::vector<CoefficientType> unrolled_coefficients;
+
+    /**
+     * Which component of FiniteElement should be used to calculate the
+     * expansion.
+     */
+    const unsigned int component;
   };
 
 
 
   /**
-   * A class to calculate expansion of a scalar FE field into series of Legendre
-   * functions on a reference element.
+   * A class to calculate expansion of a scalar FE (or a single component
+   * of vector-valued FE) field into series of Legendre functions on a
+   * reference element.
    *
    * Legendre functions are solutions to Legendre's differential equation
    * @f[
@@ -262,10 +281,21 @@ namespace FESeries
      * each direction, @p fe_collection is the hp::FECollection for which
      * expansion will be used and @p q_collection is the hp::QCollection used to
      * integrate the expansion for each FiniteElement in @p fe_collection.
+     *
+     * As the Legendre expansion can only be performed on scalar fields, this
+     * class does not operate on vector-valued finite elements and will
+     * therefore throw an assertion. However, each component of a finite element
+     * field can be treated as a scalar field, respectively, on which Legendre
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
+     * The default value of @p component only applies to scalar FEs, in which
+     * case it indicates that the sole component is to be decomposed. For
+     * vector-valued FEs, a non-default value must be explicitly provided.
      */
     Legendre(const std::vector<unsigned int> &n_coefficients_per_direction,
              const hp::FECollection<dim, spacedim> &fe_collection,
-             const hp::QCollection<dim> &           q_collection);
+             const hp::QCollection<dim> &           q_collection,
+             const unsigned int component = numbers::invalid_unsigned_int);
 
     /**
      * A non-default constructor. The @p size_in_each_direction defines the number
@@ -364,6 +394,12 @@ namespace FESeries
      * Auxiliary vector to store unrolled coefficients.
      */
     std::vector<CoefficientType> unrolled_coefficients;
+
+    /**
+     * Which component of FiniteElement should be used to calculate the
+     * expansion.
+     */
+    const unsigned int component;
   };
 
 
--- a/include/deal.II/numerics/smoothness_estimator.h
+++ b/include/deal.II/numerics/smoothness_estimator.h
@@ -232,10 +232,22 @@ namespace SmoothnessEstimator
      * polynomial which is just a constant. Further for each element, we use a
      * Gaussian quadrature designed to yield exact results for the highest order
      * Legendre polynomial used.
+     *
+     * As the Legendre expansion can only be performed on scalar fields, this
+     * class does not operate on vector-valued finite elements and will
+     * therefore throw an assertion. However, each component of a finite element
+     * field can be treated as a scalar field, respectively, on which Legendre
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
+     * The default value of @p component only applies to scalar FEs, in which
+     * case it indicates that the sole component is to be decomposed. For
+     * vector-valued FEs, a non-default value must be explicitly provided.
      */
     template <int dim, int spacedim>
     FESeries::Legendre<dim, spacedim>
-    default_fe_series(const hp::FECollection<dim, spacedim> &fe_collection);
+    default_fe_series(
+      const hp::FECollection<dim, spacedim> &fe_collection,
+      const unsigned int component = numbers::invalid_unsigned_int);
   } // namespace Legendre
 
 
@@ -458,10 +470,22 @@ namespace SmoothnessEstimator
      * a 5-point Gaussian quarature iterated in each dimension by the maximal
      * wave number, which is the number of modes decreased by one since we start
      * with $k = 0$.
+     *
+     * As the Fourier expansion can only be performed on scalar fields, this
+     * class does not operate on vector-valued finite elements and will
+     * therefore throw an assertion. However, each component of a finite element
+     * field can be treated as a scalar field, respectively, on which Fourier
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
+     * The default value of @p component only applies to scalar FEs, in which
+     * case it indicates that the sole component is to be decomposed. For
+     * vector-valued FEs, a non-default value must be explicitly provided.
      */
     template <int dim, int spacedim>
     FESeries::Fourier<dim, spacedim>
-    default_fe_series(const hp::FECollection<dim, spacedim> &fe_collection);
+    default_fe_series(
+      const hp::FECollection<dim, spacedim> &fe_collection,
+      const unsigned int component = numbers::invalid_unsigned_int);
   } // namespace Fourier
 } // namespace SmoothnessEstimator
 
--- a/source/fe/fe_series_fourier.cc
+++ b/source/fe/fe_series_fourier.cc
@@ -65,14 +65,16 @@ namespace
   integrate(const FiniteElement<dim, spacedim> &fe,
             const Quadrature<dim> &             quadrature,
             const Tensor<1, dim> &              k_vector,
-            const unsigned int                  j)
+            const unsigned int                  j,
+            const unsigned int                  component)
   {
     std::complex<double> sum = 0;
     for (unsigned int q = 0; q < quadrature.size(); ++q)
       {
         const Point<dim> &x_q = quadrature.point(q);
         sum += std::exp(std::complex<double>(0, 1) * (k_vector * x_q)) *
-               fe.shape_value(j, x_q) * quadrature.weight(q);
+               fe.shape_value_component(j, x_q, component) *
+               quadrature.weight(q);
       }
     return sum;
   }
@@ -91,6 +93,7 @@ namespace
     const hp::QCollection<1> &                     q_collection,
     const Table<1, Tensor<1, 1>> &                 k_vectors,
     const unsigned int                             fe,
+    const unsigned int                             component,
     std::vector<FullMatrix<std::complex<double>>> &fourier_transform_matrices)
   {
     AssertIndexRange(fe, fe_collection.size());
@@ -103,8 +106,8 @@ namespace
 
         for (unsigned int k = 0; k < n_coefficients_per_direction[fe]; ++k)
           for (unsigned int j = 0; j < fe_collection[fe].n_dofs_per_cell(); ++j)
-            fourier_transform_matrices[fe](k, j) =
-              integrate(fe_collection[fe], q_collection[fe], k_vectors(k), j);
+            fourier_transform_matrices[fe](k, j) = integrate(
+              fe_collection[fe], q_collection[fe], k_vectors(k), j, component);
       }
   }
 
@@ -116,6 +119,7 @@ namespace
     const hp::QCollection<2> &                     q_collection,
     const Table<2, Tensor<1, 2>> &                 k_vectors,
     const unsigned int                             fe,
+    const unsigned int                             component,
     std::vector<FullMatrix<std::complex<double>>> &fourier_transform_matrices)
   {
     AssertIndexRange(fe, fe_collection.size());
@@ -132,8 +136,12 @@ namespace
                ++k2, ++k)
             for (unsigned int j = 0; j < fe_collection[fe].n_dofs_per_cell();
                  ++j)
-              fourier_transform_matrices[fe](k, j) = integrate(
-                fe_collection[fe], q_collection[fe], k_vectors(k1, k2), j);
+              fourier_transform_matrices[fe](k, j) =
+                integrate(fe_collection[fe],
+                          q_collection[fe],
+                          k_vectors(k1, k2),
+                          j,
+                          component);
       }
   }
 
@@ -145,6 +153,7 @@ namespace
     const hp::QCollection<3> &                     q_collection,
     const Table<3, Tensor<1, 3>> &                 k_vectors,
     const unsigned int                             fe,
+    const unsigned int                             component,
     std::vector<FullMatrix<std::complex<double>>> &fourier_transform_matrices)
   {
     AssertIndexRange(fe, fe_collection.size());
@@ -166,7 +175,8 @@ namespace
                   integrate(fe_collection[fe],
                             q_collection[fe],
                             k_vectors(k1, k2, k3),
-                            j);
+                            j,
+                            component);
       }
   }
 } // namespace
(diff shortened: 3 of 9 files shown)


include/deal.II/fe/fe_series.h
```diff
@@ -364,6 +381,12 @@ namespace FESeries
      * Auxiliary vector to store unrolled coefficients.
      */
     std::vector<CoefficientType> unrolled_coefficients;
+
+    /**
+     * Which component of FiniteElement should be used to calculate the
+     * expansion.
+     */
+    unsigned int component;
```
marcfehling, line 389: You can make this `const`, too.

source/fe/fe_series_fourier.cc
```diff
@@ -179,16 +189,23 @@ namespace FESeries
...
+    , component(component)
   {
     Assert(n_coefficients_per_direction.size() == fe_collection.size() &&
              n_coefficients_per_direction.size() == q_collection.size(),
            ExcMessage("All parameters are supposed to have the same size."));
 
+#ifdef DEBUG
+    for (unsigned int i = 0; i < fe_collection.size(); ++i)
+      AssertIndexRange(component, fe_collection[i].n_components());
+#endif
```
marcfehling, line 207: Well placed assertion! :-)

include/deal.II/fe/fe_series.h
```diff
@@ -96,24 +96,29 @@ namespace FESeries
...
      * expansion will be used and @p q_collection is the hp::QCollection used to
      * integrate the expansion for each FiniteElement in @p fe_collection.
+     * @p component defines which component of each FiniteElement will be used,
+     * it is mainly used for non-primitive elements.
      */
     Fourier(const std::vector<unsigned int> &      n_coefficients_per_direction,
             const hp::FECollection<dim, spacedim> &fe_collection,
-            const hp::QCollection<dim> &           q_collection);
+            const hp::QCollection<dim> &           q_collection,
+            const unsigned int                     component = 0);
```
marcfehling, line 105: Here and other occasions: I would suggest to use `numbers::invalid_unsigned_int` as a default parameter.

If users provide vector-valued FEs and use this function, this function will work only on the very first component -- unknowing to the user. Instead, with the invalid parameter, you can check in the constructor with an assertion if the provided `fe_collection` consists of scalar or vector-valued finite elements, and then set `component` to zero if it only contains scalar FEs. This way users will be alarmed whether they are doing something unintentionally.

include/deal.II/fe/fe_series.h
```diff
@@ -96,24 +96,29 @@ namespace FESeries
      * each direction, @p fe_collection is the hp::FECollection for which
      * expansion will be used and @p q_collection is the hp::QCollection used to
      * integrate the expansion for each FiniteElement in @p fe_collection.
+     * @p component defines which component of each FiniteElement will be used,
+     * it is mainly used for non-primitive elements.
```
marcfehling, line 100: Here and other occasions:
```suggestion
     *
     * As the Fourier expansion can only be performed on scalar fields, this class does not operate on vector-valued finite elements and will therefore throw an assertion. However, each component of a finite element field can be treated as a scalar field, respectively, on which Fourier expansions are again possible. For this purpose, the optional parameter @p component defines which component of each FiniteElement will be used.
```

marcfehling (changes requested):
Looks good so far!

Could you provide a test that provides the whole smoothness estimation workflow? You can basically take your minimal example of #11791 for this, and initialize your `solution` vector with meaningful values, if necessary.

I somehow doubt that the estimator functions in the `SmoothnessEstimator` namespace will work with non-primitive finite elements out of the box. See here and other occasions: https://github.com/dealii/dealii/blob/c9feb145b3d368600a145c094dc2739731ef1530/source/numerics/smoothness_estimator.cc#L131-L136

For non-primitive finite elements, you will not need all `dof_values` on the expansion, but only those associated with the `component` of the `fe_legendre` object.

adamqc:
> For non-primitive finite elements, you will not need all `dof_values` on the expansion, but only those associated with the `component` of the `fe_legendre` object.

Good point.
I'm not quite familiar with the `FiniteElement` class, how can I get the `dof_values` associate with the `component`?

marcfehling:
> I'm not quite familiar with the `FiniteElement` class, how can I get the `dof_values` associate with the `component`?

I haven't worked much with vector-valued problems yet, so maybe someone else has a better idea than me:

Maybe the module documentation [Handling vector valued problems - Extracting data from solutions](https://www.dealii.org/developer/doxygen/deal.II/group__vector__valued.html#VVExtracting) helps. The approach described there would involve using `FEValues` and `FEValuesExtractor`, but since we only need the correct indices from the finite element solution on each cell, I think there is a better approach than that.

[`step-8`](https://www.dealii.org/developer/doxygen/deal.II/step_8.html#ElasticProblemassemble_system) describes how to use [`FiniteElement::system_to_component_index()`](https://www.dealii.org/developer/doxygen/deal.II/classFiniteElement.html#a86644fe67824373cd51e9ff7fca94f8c). Maybe you can use the opposite function [`FiniteElement::component_to_system_index()`](https://www.dealii.org/developer/doxygen/deal.II/classFiniteElement.html#ae2ea16b60a6fc644a9bc7097703a53e8) to extract those dof values that belong to your component from [`DoFCellAccessor::get_dof_values()`](https://www.dealii.org/developer/doxygen/deal.II/classDoFCellAccessor.html#a1702d1e9c1cb4fb07b70c38ac9165aaa). It would be handy to introduce an overload of the latter function in this context.

include/deal.II/fe/fe_series.h
bangerth: Yes, I agree.

include/deal.II/fe/fe_series.h
```diff
@@ -204,6 +209,12 @@ namespace FESeries
      * Auxiliary vector to store unrolled coefficients.
      */
     std::vector<CoefficientType> unrolled_coefficients;
+
+    /**
+     * Which component of FiniteElement should be used to calculate the
+     * expansion.
+     */
+    unsigned int component;
```
bangerth, line 217:
```suggestion
    const unsigned int component;
```

include/deal.II/fe/fe_series.h
```diff
@@ -96,24 +96,29 @@ namespace FESeries
...
+     * FiniteElement will be used, it is mainly used for non-primitive elements.
      *
      * @deprecated Use a different constructor instead.
      */
     DEAL_II_DEPRECATED
     Fourier(const unsigned int                     n_coefficients_per_direction,
             const hp::FECollection<dim, spacedim> &fe_collection,
-            const hp::QCollection<dim> &           q_collection);
+            const hp::QCollection<dim> &           q_collection,
+            const unsigned int                     component = 0);
```
bangerth, line 121: This constructor is deprecated anyway, you probably don't need to add the `component` argument.

include/deal.II/fe/fe_series.h
bangerth, line 389:
```suggestion
    const unsigned int component;
```

include/deal.II/numerics/smoothness_estimator.h
```diff
@@ -232,10 +232,13 @@ namespace SmoothnessEstimator
...
      * Gaussian quadrature designed to yield exact results for the highest order
      * Legendre polynomial used.
+     * @p component defines which component of each FiniteElement will be used,
+     * it is mainly used for non-primitive elements.
      */
     template <int dim, int spacedim>
     FESeries::Legendre<dim, spacedim>
-    default_fe_series(const hp::FECollection<dim, spacedim> &fe_collection);
+    default_fe_series(const hp::FECollection<dim, spacedim> &fe_collection,
+                      const unsigned int                     component = 0);
```
bangerth, line 241: same here with the `numbers::invalid_unsigned_int`.

source/fe/fe_series_fourier.cc
```diff
@@ -65,14 +65,16 @@ namespace
...
+            const unsigned int                  component)
   {
     std::complex<double> sum = 0;
     for (unsigned int q = 0; q < quadrature.size(); ++q)
       {
         const Point<dim> &x_q = quadrature.point(q);
         sum += std::exp(std::complex<double>(0, 1) * (k_vector * x_q)) *
-               fe.shape_value(j, x_q) * quadrature.weight(q);
+               fe.shape_value_component(j, x_q, component) *
+               quadrature.weight(q);
```
bangerth, line 77: Yes, I think this is really the only necessary change, at least for primitive vector-valued elements

source/fe/fe_series_fourier.cc
```diff
@@ -179,16 +189,23 @@ namespace FESeries
...
     const std::vector<unsigned int> &      n_coefficients_per_direction,
     const hp::FECollection<dim, spacedim> &fe_collection,
-    const hp::QCollection<dim> &           q_collection)
+    const hp::QCollection<dim> &           q_collection,
+    const unsigned int                     component)
     : n_coefficients_per_direction(n_coefficients_per_direction)
     , fe_collection(&fe_collection)
     , q_collection(q_collection)
     , fourier_transform_matrices(fe_collection.size())
+    , component(component)
```
bangerth, line 198: With the invalid value, this would
```suggestion
    , component(component != numbers::invalid_unsigned_int ? component : 0)
```
and then you would add the following assertion below:
```
  if (fe_collection.n_component() > 0)
    Assert (component != numbers::invalid_unsigned_int,
                ExcMessage ("For vector-valued problems, you need to explicitly specify for which vector "
                        "component you will want to do a Fourier decomposition by setting the 'component' "
                        "argument of this constructor.")
```

source/fe/fe_series_fourier.cc
bangerth: But I think you don't actually need it in this generality. `hp::FECollection` enforces that all elements have the same number of vector components, so it is probably enough to just check for `[0]`.

source/fe/fe_series_fourier.cc
marcfehling: It would make sense to give the constructor parameter and the class member different names to avoid ambiguity, e.g., `component_` for the former and `component` for the latter.

adamqc:
I have addressed all your comments except @marcfehling 's comment about `dof_values`. I think that for non-primitive elements such as Nedelec and RT that all components are coupled with all dofs, they can not be separated simply. Am I right?

masterleinad:
> I think that for non-primitive elements such as Nedelec and RT that all components are coupled with all dofs, they can not be separated simply. Am I right?

`FEValuesBase::shape_value_component` also works for non-primitive elements. The question is rather if it makes sense to only compute a Fourier/Legendre expansion corresponding to one component in this case but the implementation would still allow this AFAICT.

bangerth:
That is the same as for `KellyErrorEstimator`, by the way. If that class does it per component, then so should this one here.

tests/numerics/smoothness_estimator_03.output
```diff
@@ -0,0 +1,3 @@
+
+DEAL::Ok
+DEAL::Ok
```
marcfehling, line 3: Judging from the source code, this should be in capitals. Did you run this test yet and checked if it passes?

```suggestion
DEAL::OK
DEAL::OK
```

tests/numerics/smoothness_estimator_03.cc
```diff
@@ -0,0 +1,72 @@
...
+#include <deal.II/hp/fe_collection.h>
+
+#include <deal.II/lac/vector.h>
+
+#include <deal.II/numerics/smoothness_estimator.h>
+
+#include "../tests.h"
+
+// This test is used to make sure FESeries::Fourier/Legendre
+// woring with non-primitive elements
```
marcfehling, line 18: Would you move your brief test description before the `include` statements, and add the deal.II copyright preamble in the front of this file?

You could draw inspiration from the other smoothness_estimator tests.

```suggestion
// This test is used to make sure that FESeries::Fourier/Legendre
// work with non-primitive elements
```

marcfehling (changes requested):
Minor comments and a few very pedantic suggestions.

tests/numerics/smoothness_estimator_03.output
adamqc: Sorry. The source code was modified unintentionally.

include/deal.II/fe/fe_series.h
```diff
@@ -96,10 +97,18 @@ namespace FESeries
...
+     * therefore throw an assertion. However, each component of a finite element
+     * field can be treated as a scalar field, respectively, on which Fourier
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
      */
     Fourier(const std::vector<unsigned int> &      n_coefficients_per_direction,
             const hp::FECollection<dim, spacedim> &fe_collection,
-            const hp::QCollection<dim> &           q_collection);
+            const hp::QCollection<dim> &           q_collection,
+            const unsigned int component = numbers::invalid_unsigned_int);
```
bangerth, line 111: Can you document how the default value is interpreted? As in, the default value can only be used for a scalar finite element, in which case it indicates that the sole component is to be decomposed. For vector-valued elements, a non-default value of that argument has to be explicitly provided.

include/deal.II/fe/fe_series.h
```diff
@@ -262,10 +278,18 @@ namespace FESeries
...
+     * therefore throw an assertion. However, each component of a finite element
+     * field can be treated as a scalar field, respectively, on which Legendre
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
      */
     Legendre(const std::vector<unsigned int> &n_coefficients_per_direction,
              const hp::FECollection<dim, spacedim> &fe_collection,
-             const hp::QCollection<dim> &           q_collection);
+             const hp::QCollection<dim> &           q_collection,
+             const unsigned int component = numbers::invalid_unsigned_int);
```
bangerth, line 292: same here

include/deal.II/numerics/smoothness_estimator.h
```diff
@@ -232,10 +232,19 @@ namespace SmoothnessEstimator
...
+     * field can be treated as a scalar field, respectively, on which Legendre
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
      */
     template <int dim, int spacedim>
     FESeries::Legendre<dim, spacedim>
-    default_fe_series(const hp::FECollection<dim, spacedim> &fe_collection);
+    default_fe_series(
+      const hp::FECollection<dim, spacedim> &fe_collection,
+      const unsigned int component = numbers::invalid_unsigned_int);
```
bangerth, line 247: and here

include/deal.II/numerics/smoothness_estimator.h
```diff
@@ -458,10 +467,19 @@ namespace SmoothnessEstimator
...
+     * field can be treated as a scalar field, respectively, on which Fourier
+     * expansions are again possible. For this purpose, the optional parameter
+     * @p component defines which component of each FiniteElement will be used.
      */
     template <int dim, int spacedim>
     FESeries::Fourier<dim, spacedim>
-    default_fe_series(const hp::FECollection<dim, spacedim> &fe_collection);
+    default_fe_series(
+      const hp::FECollection<dim, spacedim> &fe_collection,
+      const unsigned int component = numbers::invalid_unsigned_int);
```
bangerth, line 482: and here

bangerth (commented):
Excellent! I have just a couple of comments.

But here is one other question. At the end of the `calculate()` function, we have
```
    const FullMatrix<CoefficientType> &matrix =
      fourier_transform_matrices[cell_active_fe_index];

    [...]

    for (unsigned int i = 0; i < unrolled_coefficients.size(); i++)
      for (unsigned int j = 0; j < local_dof_values.size(); j++)
        unrolled_coefficients[i] += matrix[i][j] * local_dof_values[j];

    fourier_coefficients.fill(unrolled_coefficients.begin());
```
That last block computes the expansion by multiplying the matrix that contains the expansion of each shape function (i.e., the component of interest of each shape function) by the DoF values on each cell.

Many of the columns of this matrix will be zero, except for those columns for which a shape function is nonzero in the selected component. I think that means that all DoF values for other vector components simply don't make a difference, as indeed should be the case. So the fact that we built a potentially much larger matrix than necessary doesn't matter, right?

One could, in a follow-up, optimize the integration of these matrices from the current code
```
        for (unsigned int k = 0; k < n_coefficients_per_direction[fe]; ++k)
          for (unsigned int j = 0; j < fe_collection[fe].n_dofs_per_cell(); ++j)
            fourier_transform_matrices[fe](k, j) =
              integrate(fe_collection[fe], q_collection[fe], k_vectors(k), j);
```
to one that switches the order of loops and simply ignores those `j` indices for which we know that the shape function is zero in the selected component. But let's do that in a later patch.

adamqc:
> One could, in a follow-up, optimize the integration of these matrices from the current code
> to one that switches the order of loops and simply ignores those `j` indices for which we know that the shape function is zero in the selected component. But let's do that in a later patch.

I will investigate how to achieve this.

include/deal.II/numerics/smoothness_estimator.h
adamqc: I have addressed your comments about the documentation.

bangerth (approved):
Great, thanks a lot!

PR #11831, opened 2021-03-01
MGTwoLevelTransfer::restrict_and_add(): remove erroneous if-statement

peterrum:
FYI @nfehn


kronbichler (approved):
Looks good to me. This seems to have been a pretty serious bug with non-robust behavior also for moderate numbers of MPI ranks, so it is good to have it fixed.

masterleinad (approved):

PR #11833, opened 2021-03-02
Modify fullydistributed_grids/solution_transfer_01 to resemble use case.

marcfehling:
Follow-up to #11798.

Adjusting `tests/fullydistributed_grids/solution_transfer_01.cc` to resemble an actual use case reveals that the so proposed workflow throws an error in the current implementation. We do not have a solution for this problem yet -- so this PR stays in draft state until we find one.

The actual error is discussed [here](https://github.com/dealii/dealii/pull/11798#issuecomment-784755579) and a possible hotfix [here](https://github.com/dealii/dealii/pull/11798#issuecomment-784857211).

@peterrum, @elauksap -- FYI

### DIFF
--- a/tests/fullydistributed_grids/solution_transfer_01.cc
+++ b/tests/fullydistributed_grids/solution_transfer_01.cc
@@ -34,7 +34,6 @@
 
 #include "../tests.h"
 
-using namespace dealii;
 
 template <int dim>
 class InterpolationFunction : public Function<dim>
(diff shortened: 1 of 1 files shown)


bangerth:
Has this just become obsolete?

marcfehling:
No, this is still important. I just rebased on the master branch.

`parallel::fullydistributed::Triangulation::load()` claims it needs an empty triangulation to work on. However, if provided a freshly created object, it throws an error (see links in PR description).

In the previous state of the test, `parallel::fullydistributed::Triangulation::clear()` was called, but this does not resemble a use case for serialization over the file system. So this is an open problem.

marcfehling:
I see `tests/fullydistributed_grids/save_load_01.cc` follows the same logic and should be adjusted, too.

bangerth:
OK. Let us know when it's ready for review.

PR #11835, opened 2021-03-02
Save output if simplex tests have failed

peterrum:
Not optimal, since all output files in the simplex test folder are saved if any of the test fails. But for now it is OK, since the size of the whole output is less than one MB (once we have significantly more tests I would suggest to abandon the GitHub action and go to Jenkins). The output can be accessed zipped via the GitHub action.

Tried out for #11503.

FYI @bangerth @michelebucelli (sorry for taking this long)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -59,6 +59,8 @@ jobs:
         cmake --version
     - name: configure
       run: |
+        mkdir build
+        cd build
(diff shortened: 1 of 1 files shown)


bangerth (approved):
Let's use this, at least temporarily. I think that most of it is actually useful independently (namely, the part where you replace the in-source build by an out-of-source build). We can remove the last hunk once we know what's going wrong.

PR #11839, opened 2021-03-03
Enable Utilities::MPI::compute_point_to_point_communication_pattern for self communication

peterrum:
In many cases, it allows to writer simpler code if one assume that a process can also communicate with itself, since no extra specialization has to be explicitly written. This might be a bit slower since data has be written and read to/from buffers but the code is simpler.

### DIFF
--- a/source/base/mpi.cc
+++ b/source/base/mpi.cc
@@ -368,9 +368,6 @@ namespace Utilities
         {
           (void)destination;
           AssertIndexRange(destination, n_procs);
-          Assert(destination != myid,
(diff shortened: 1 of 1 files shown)


drwells:
`some_to_some` permits sending messages to oneself so this should work in the same way.

drwells (approved):

kronbichler (approved):

bangerth (approved):

bangerth:
The usual test timed out in the CI check.

PR #11841, opened 2021-03-03
Reduce number of procs from 15 to 10 in test

### LINKED ISSUE #11789: Some tests in debug configuration take too long
The CI fails reproducibly with the following test:
```
The following tests FAILED:
	3329 - multigrid/transfer_matrix_free_06.mpirun=15.debug (Timeout)
```

I think it is time to benchmark the execution time of all tests again and trim down some bad offenders.

peterrum:
As discussed in https://github.com/dealii/dealii/issues/11789#issuecomment-782913327. Might fix #11789.


drwells:
We have probably answered this in the past but I'll ask anyway. Could we just pass some `enum` to this class that has possible values `algorithm1`, `algorithm2`, and `auto` where `auto` is the current (i.e., dependent on the number processors) behavior? That would sidestep this whole strange requirement for switching between algorithms in debug mode so that we can actually execute them both.

peterrum:
> We have probably answered this in the past but I'll ask anyway. Could we just pass some enum to this class that has possible values algorithm1, algorithm2, and auto where auto is the current (i.e., dependent on the number processors) behavior? 

We could introduce an input parameter for selecting the right algorithm. But if you know which algorithm to use, why not select the right class!?

> That would sidestep this whole strange requirement for switching between algorithms in debug mode so that we can actually execute them both.

No it wouldn't. The class is not used in the test directly, but via vector -> partitioner -> ... So we wouldn't have a way to control the behavior via the enum.

drwells:
That's unfortunate but I guess we end up with the right thing in practice at least.

drwells (approved):

PR #11842, opened 2021-03-03
Sort return object of GridTools::distributed_compute_point_locations

peterrum:
... cell-level/cell-index/rank/local point index. This is consistently as done in #11804. The code will be remove once #11804 is merged; but this PR shows why a single test has a different output.

### DIFF
--- a/source/grid/grid_tools.cc
+++ b/source/grid/grid_tools.cc
@@ -5110,7 +5110,22 @@ namespace GridTools
           }
       }
     } // namespace DistributedComputePointLocations
-  }   // namespace internal
+
+
+    template <typename T, typename U>
+    std::vector<T>
+    permute(const std::vector<T> &input, const std::vector<U> &indices)
+    {
+      std::vector<T> ouput;
+      ouput.reserve(input.size());
+
+      for (unsigned int i = 0; i < input.size(); ++i)
+        ouput.push_back(input[std::get<2>(indices[i])]);
+
+      return ouput;
+    }
+
+  } // namespace internal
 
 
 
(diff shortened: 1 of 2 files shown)


source/grid/grid_tools.cc
```diff
@@ -5341,14 +5356,40 @@ namespace GridTools
     out_points.resize(size_output);
     out_ranks.resize(size_output);
 
-    unsigned int c = 0;
+    unsigned int                                    c = 0;
+    std::vector<std::tuple<int, int, unsigned int>> cells_sorted;
```
bangerth, line 5360:
```suggestion
    std::vector<std::tuple<int, int, unsigned int>> cells_sorted;
    cells_sorted.reserve (found_points.size());
```

source/grid/grid_tools.cc
```diff
@@ -5110,7 +5110,22 @@ namespace GridTools
           }
       }
     } // namespace DistributedComputePointLocations
-  }   // namespace internal
+
+
+    template <typename T, typename U>
+    std::vector<T>
+    permute(const std::vector<T> &input, const std::vector<U> &indices)
```
bangerth, line 5117: That's the kind of function that would also have gone well into namespace `Utilities`.

If you wanted to do that, it would be useful to do this:
```suggestion
    permute(const std::vector<T> &input, const std::vector<typename std::vector<T>::size_type> &indices)
```
and drop the template type `U`.

bangerth (approved):

bangerth:
Feel free to merge as-is, to make changes to your test as suggested, or to ignore things for now and come back to it later. No need to address what I wrote above!

/rebuild

source/grid/grid_tools.cc
peterrum: This is not possible since `U` is not `unsigned int` but a tuple `std::tuple<unsigned int, unsigned int, unsigned int>`. As stated above, this function will be gone after the refactoring.

PR #11848, opened 2021-03-04
Fix tangential vector of wedge

peterrum:


bangerth (approved):
I'll assume that you have checked by hand that the new output is correct.

/rebuild

peterrum:
Yes, I did.

PR #11849, opened 2021-03-04
Use get_communicator()

peterrum:
I searched for `MPI_COMM_SELF`. There are probably more instances where it can be used.


bangerth (approved):
Very nice!

/rebuild

PR #11850, opened 2021-03-05
Added Utilities::MPI::logical_or().

marcfehling:
New function for collective _logical or_ operations. Will be used in a follow-up PR.

Can not be used on `std::vector<bool>` since its specialization may not store its elements in a contiguous array.

### DIFF
--- a/doc/news/changes/minor/20210304Fehling
+++ b/doc/news/changes/minor/20210304Fehling
@@ -1,0 +1,3 @@
+New: Utilities::MPI::logical_or() for collective <i>logical or</i> operations.
+<br>
+(Marc Fehling, 2021/03/04)
--- a/include/deal.II/base/mpi.h
+++ b/include/deal.II/base/mpi.h
@@ -53,6 +53,9 @@ using MPI_Op       = int;
 #  ifndef MPI_SUM
 #    define MPI_SUM 0
 #  endif
+#  ifndef MPI_LOR
+#    define MPI_LOR 0
+#  endif
 #endif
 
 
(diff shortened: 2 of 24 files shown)


peterrum (approved):
Why did you not simply use `Utilities::MPI::max()` which is equivalent?

bangerth (changes requested):
What is `MPI_LOR`? A bitwise `or` operation? I'm not familiar with the term "logical or", and if I'm not, then it might be useful to explain it in the documentation. That would also cover the question of what kind of template arguments `T` this function is useful for.

marcfehling:
> Why did you not simply use `Utilities::MPI::max()` which is equivalent?

This would have worked as well...
In retrospective for my use case, I would only need to provide an overload of the `mpi_type_id` function for the `bool` data type.

marcfehling:
> What is `MPI_LOR`? A bitwise `or` operation? I'm not familiar with the term "logical or", and if I'm not, then it might be useful to explain it in the documentation. That would also cover the question of what kind of template arguments `T` this function is useful for.

`MPI_LOR` is the _logical or_ operation `||`. `MPI_BOR` is the _bitwise or_ operation `|`.
I'll be more elaborate, and also on how different data types `T` will be treated.

bangerth (approved):

bangerth:
Right now, you allow any data type `T` to be passed, but what does MPI do if that data type is, for example, `std::complex<double>`? I would suspect that MPI has some sort of restrictions on what that could be, and it would probably be useful to encode those into a `static_assert` in the one function you route everything through (or all of them). Maybe something like
```
  static_assert (std::is_convertible<T,bool>::value, "...");
```
?

marcfehling:
> Right now, you allow any data type `T` to be passed, but what does MPI do if that data type is, for example, `std::complex<double>`? I would suspect that MPI has some sort of restrictions on what that could be, and it would probably be useful to encode those into a `static_assert` in the one function you route everything through (or all of them). Maybe something like
> 
> ```
>   static_assert (std::is_convertible<T,bool>::value, "...");
> ```
> 
> ?

According to the mpi documentation, the `MPI_LOR` operation requires an `integral` type. I will check for that instead.

bangerth (approved):
Much appreciated, in particular the multitude of tests!

PR #11851, opened 2021-03-05
Replace 'typedef' by 'using' in the test suite.

bangerth:
Not sure what's going to happen here, but one of these patches that we probably have to do at one point :-)

Also my apologies in advance to whomever is going to review this. It's all script generated, so can't possibly be wrong.

/rebuild


peterrum (approved):

masterleinad (approved):
I don't think that we **need** to do this but I very much prefer the `using` syntax` over `typedef`. Since we have already done that conversion in the rest of the repository, this is just consistent.

bangerth:
Yes, I meant "need" as in "we try to be consistent".

PR #11853, opened 2021-03-05
Improve docs (indentation targets)

konsim83:
I added a little note in the coding conventions about the indentation command in case the `clang-format` executable points to a more recent version than v.6.0.

### DIFF
--- a/doc/doxygen/headers/coding_conventions.h
+++ b/doc/doxygen/headers/coding_conventions.h
@@ -56,6 +56,14 @@ pre-commit hook. One way to do so, is to copy
 <code>\${SOURCE_DIR}/contrib/git-hooks/pre-commit</code> to
 <code>\${SOURCE_DIR}/.git/hooks/pre-commit</code> and make sure it is
 executable.
+
+If the system you are working on has more than one version of
+<code>clang-format</code> installed (or if it is not in the path)
+you should replace the above <code>make indent</code> command with
+@code
+  make DEAL_II_CLANG_FORMAT=/path/to/clang-6.0/clang-format indent
+@endcode
+to point to the correct executable.
 </p>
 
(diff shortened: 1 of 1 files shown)


doc/doxygen/headers/coding_conventions.h
```diff
@@ -56,6 +56,14 @@ pre-commit hook. One way to do so, is to copy
...
 <code>\${SOURCE_DIR}/.git/hooks/pre-commit</code> and make sure it is
 executable.
+
+On some systems the <code>clang-format</code> executable points to a more
+recent version such as <code>clang-format-10</code>. In this case you should
+replace the above <code>make indent</code> command with
+@code
+  make DEAL_II_CLANG_FORMAT=clang-format-6.0 indent
+@endcode
+provided <code>clang-format-6.0</code> is installed and in your path.
```
bangerth, line 66: There will rarely be a command that is actually named that way, but you probably mean that one can provide a full path to an executable with that specific version. Maybe replace `clang-format-6.0` by `/usr/local/clang-6.0/clang-format` in the two places you have it above, and say something like "...and assuming that you have installed version 6.0 of clang-format in /usr/local/clang-6.0. Of course, if you have installed that version of clang-format in a different directory, you should use the corresponding path."?

bangerth (changes requested):

bangerth:
Also, thanks for helping out with these rather mundane aspects of project maintenance!

doc/doxygen/headers/coding_conventions.h
konsim83: Done.

konsim83:
> Also, thanks for helping out with these rather mundane aspects of project maintenance!

Pleasure.

bangerth (approved):

Issue #11845, opened 2021-03-04
FEFaceEvaluation: make is_interior_face in constructors an enum


peterrum:
As pointed out by multiple users, the specification of interior/exterior face via a bool is odd and error prone. 

https://github.com/dealii/dealii/blob/87a5130c56dd5fb8cf7d57b8f130a45d5a874425/include/deal.II/matrix_free/fe_evaluation.h#L3092-L3098

In particular, if someone has only worked with `FEEvaluation` and simply copies the code from there and only changes the class name, one might forget to add the bool with the consequence that the dof-index (if it is given explicitly) is implicitly treated as bool. 

I would suggest to make the bool to an enum following the changes regarding `EvaluationFlags`. Since, we have introduced a new constructor since the last release anyway, we could start here and (early) deprecate the other constructors. In my opinion, we should "motivate" users to use the new constructors because only these guarantee that `MatrixFree`/`FEEvaluation`/`FEFaceEvaluation` can be used with simplex/mixed meshes.

Any opinions?

FYI @mschreter

PR #11854, opened 2021-03-05
Improve assert in FEFaceEvaluation

mschreter:
This PR makes the assert message, when `FEFaceEvaluation` is configured with `is_interior_face=false` and a boundary face is requested, more verbose. This assert could happen if one is used to `FEEvaluation` and uses `FEFaceEvaluation<dim>` in the same spirit, with the second argument being the `dof_idx`, which will be in case of `dof_idx=0` implicitly converted to `bool is_interior_face=false`. More details on this issue are pointed out in #11845.


kronbichler (approved):

peterrum (approved):
Thanks!

PR #11857, opened 2021-03-06
Parallelize DoFHandler::prepare_coarsening_and_refinement().

marcfehling:
Blocked by #11850. Follow-up to #11812.

This PR consists of 1 commit:

* [ ] ~Use the correct data type for FE indices in `communicate_active_fe_indices()`.
Could also be tackled in a separate PR if you want.~ (EDIT: Extracted to #11863)
* [x] Parallelization via `LA::distributed::Vector`.
The communication pattern needs to be set up only once, but requires O(N) memory.
* [ ] Parallelization via `GridTools::exchange_cell_data_to_ghosts()`.
The communication pattern needs to be set up in every sweep, but no additional memory is required. (EDIT: No longer part of this PR, but stored in a [separate branch](https://github.com/marcfehling/dealii/tree/limit-levels-parallel-gridtools))

---

The `GridTools` variant works, so I was confident enough to open this PR. The `LA::distributed::Vector` variant however does not work currently and triggers the following assertion. Maybe @peterrum knows what to do.
```
4347: --------------------------------------------------------
4347: An error occurred in line <1650> of file </raid/fehling/dealii/include/deal.II/lac/la_parallel_vector.h> in function
4347:     Number& dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::operator()(dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::size_type) [with Number = int; MemorySpace = dealii::MemorySpace::Host; dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::size_type = unsigned int]
4347: The violated condition was: 
4347:     partitioner->in_local_range(global_index) || partitioner->ghost_indices().is_element(global_index)
4347: Additional information: 
4347:     You tried to access element 0 of a distributed vector, but this
4347:     element is not stored on the current processor. Note: The range of
4347:     locally owned elements is [0,4294967295], and there are 0 ghost
4347:     elements that this vector can access.
4347:     
4347:     A common source for this kind of problem is that you are passing a
4347:     'fully distributed' vector into a function that needs read access to
4347:     vector elements that correspond to degrees of freedom on ghost cells
4347:     (or at least to 'locally active' degrees of freedom that are not also
4347:     'locally owned'). You need to pass a vector that has these elements as
4347:     ghost entries.
4347: --------------------------------------------------------
```

### DIFF
--- a/include/deal.II/dofs/dof_handler.h
+++ b/include/deal.II/dofs/dof_handler.h
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 1998 - 2020 by the deal.II authors
+// Copyright (C) 1998 - 2021 by the deal.II authors
 //
 // This file is part of the deal.II library.
 //
@@ -725,8 +725,6 @@ public:
    * library (contrary to its Triangulation counterpart).
    *
    * Returns whether any future FE indices have been changed by this function.
-   *
-   * @note Only implemented for serial applications.
    */
   bool
   prepare_coarsening_and_refinement(
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 1998 - 2020 by the deal.II authors
+// Copyright (C) 1998 - 2021 by the deal.II authors
 //
 // This file is part of the deal.II library.
 //
@@ -31,6 +31,8 @@
 #include <deal.II/grid/tria_accessor.h>
 #include <deal.II/grid/tria_iterator.h>
 #include <deal.II/grid/tria_levels.h>
+
+#include <deal.II/lac/la_parallel_vector.templates.h>
 
 #include <algorithm>
 #include <memory>
@@ -2670,12 +2672,6 @@ DoFHandler<dim, spacedim>::prepare_coarsening_and_refinement(
     // nothing to do
     return false;
 
-  // communication of future fe indices on ghost cells necessary.
-  // thus, currently only implemented for serial triangulations.
-  Assert((dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
-            &(*tria)) == nullptr),
-         ExcNotImplemented());
-
   //
   // establish hierarchy
   //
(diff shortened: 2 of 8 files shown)


peterrum:
@marcfehling May I ask you to drop the last commit. After that I will review the code.

include/deal.II/dofs/dof_handler.h
```diff
@@ -725,13 +725,10 @@ class DoFHandler : public Subscriptor
...
    * Returns whether any future FE indices have been changed by this function.
-   *
-   * @note Only implemented for serial applications.
    */
   bool
-  prepare_coarsening_and_refinement(
-    const unsigned int max_difference    = 1,
-    const unsigned int contains_fe_index = 0) const;
+  prepare_coarsening_and_refinement(const unsigned int max_difference    = 1,
+                                    const unsigned int contains_fe_index = 0);
```
bangerth, line 731: ha, oops :-)

bangerth:
While we wait for the other PRs to get merged, can you break out the commit that fixes the data types into a PR of its own?

marcfehling:
> @marcfehling May I ask you to drop the last commit. After that I will review the code.

Done! I've saved the last commit (and the `GridTools` implementation) in a [separate branch](https://github.com/marcfehling/dealii/tree/limit-levels-parallel-gridtools).

source/dofs/dof_handler.cc
```diff
@@ -2689,66 +2686,135 @@ DoFHandler<dim, spacedim>::prepare_coarsening_and_refinement(
...
 
+  //
+  // parallelization
+  //
+  // - create distributed copy of future fe index vector
+  // - update ghost values after each iteration (see later)
+  // - no need to compress, since the owning processor will have the correct FE
+  //   index
+
+  LinearAlgebra::distributed::Vector<int> future_fe_indices;
```
marcfehling, line 2697: We could also make its data type `active_fe_index_type` I think.

source/dofs/dof_handler.cc
```diff
@@ -2689,66 +2686,135 @@ DoFHandler<dim, spacedim>::prepare_coarsening_and_refinement(
...
+                               partitioner.get_mpi_communicator());
+    }
+  else
+    {
+      future_fe_indices.reinit(tria->n_active_cells());
+    }
+
+  // communicate active FE indices on ghost cells in case the user changed them
+  dealii::internal::hp::DoFHandlerImplementation::Implementation::
+    communicate_active_fe_indices(*this);
```
marcfehling, line 2715: This is not necessary. But in case the user changes the active FE indices somewhere, this will prevent them from nasty surprises.

peterrum: Not necessary. The values will be overridden anyway by `future_fe_indices.update_ghost_values();` below.

source/dofs/dof_handler.cc
```diff
@@ -2689,66 +2686,125 @@ DoFHandler<dim, spacedim>::prepare_coarsening_and_refinement(
...
+      future_fe_indices.update_ghost_values();
+
       for (const auto &cell : active_cell_iterators())
-        {
-          const auto cell_fe_and_level =
-            hierarchy_level_for_fe_index.find(cell->future_fe_index());
+        if (!cell->is_artificial())
+          {
+            const auto cell_fe_and_level = hierarchy_level_for_fe_index.find(
+              future_fe_indices[cell->global_active_cell_index()]);
```
peterrum, line 2743: I wouldn't search here. Bur rather fill the `future_fe_indices` vector already with normalized future-fe indices. This would mean that you have to translate at the beginning (before filling the vector) the future fe indices to the normalized indices and at the end (while setting the actual future-fe index the via `cell->set_future_fe_index()`) you would need to translate the normalized value to the actual value.

source/dofs/dof_handler.cc
```diff
@@ -2689,66 +2686,125 @@ DoFHandler<dim, spacedim>::prepare_coarsening_and_refinement(
...
+                          fe_indices_changed_in_cycle = true;
+                        }
+                    }
+                }
+          }
+
+      fe_indices_changed_in_cycle =
+        Utilities::MPI::logical_or(fe_indices_changed_in_cycle,
+                                   tria->get_communicator());
+      fe_indices_changed = fe_indices_changed || fe_indices_changed_in_cycle;
```
peterrum, line 2798:
```suggestion
      fe_indices_changed |= fe_indices_changed_in_cycle;
```

bangerth:
I get that one *could* go this route, but what is wrong with the previous approach? This function is called exactly once per mesh refinement cycle, and one follows the call to this function with the (almost certainly) vastly more expensive mesh refinement and DoF enumeration operations. I think that the code @marcfehling originally wrote was very clear, explicitly importing whatever data was necessary, and I don't quite see why we should go to this possibly slightly cheaper but not nearly as obvious code...

source/dofs/dof_handler.cc
marcfehling: That looks way nicer!

source/dofs/dof_handler.cc
marcfehling: You are right! We can leave this out.

source/dofs/dof_handler.cc
marcfehling: You mean to have a `future_level_indices` distributed vector instead, and then translate them back at the end of the loop while actually setting `cell->set_future_fe_index()`? I like this.

marcfehling:
> I think that the code @marcfehling originally wrote was very clear, explicitly importing whatever data was necessary, and I don't quite see why we should go to this possibly slightly cheaper but not nearly as obvious code...

With @peterrum's suggestion to work on levels exclusively, and only translate them from/to FE indices in the beginning/end, I think that this patch has become a lot more readable than the original variant using `GridTools::exchange_cell_data_to_ghosts()`.

I did not like the fact that I introduced a function `communicate_future_fe_indices` to store ghost FE indices in the `DoFHandler` member variables. In my opinion, they should only exist there for locally owned cells. Storing them for ghost cells, even for just a brief moment, could be prone to bugs and unwanted behavior of the code. With the current implementation, we ensure to access them only via the provided interface, and no more dirty hacks like this one are necessary:
https://github.com/marcfehling/dealii/blob/e544209e7e1e9984d58750675f2713cb670b267b/source/dofs/dof_handler.cc#L2787-L2798

We still need to figure out some compilation issues, but otherwise I'd prefer to leave this patch at its current state.

marcfehling:
Also, we did not consider the case in which neighbors can be located on subfaces. I've been working on this on a separate branch. Shall we treat this in two separate PRs, or should I merge the separate branch with this one?

bangerth:
One step at a time.

If you're happier with the way you have it here, then by all means let's leave that choice up to you! Let us know when it all works.

include/deal.II/lac/la_parallel_vector.templates.h
```diff
@@ -21,6 +21,7 @@
 
 #include <deal.II/base/cuda.h>
 #include <deal.II/base/cuda_size.h>
+#include <deal.II/base/partitioner.templates.h>
```
peterrum, line 24: ?

peterrum (approved):
I like this; the code is surprisingly short! Thanks for the efforts!

I tend to suggest to ignore the int-problem for now and use float instead (and cast the values if needed). Getting LA:p:V working for ints or other types can be done later (if it is even possible?).

bangerth:
That's awkward, though: The conversion `int -> float -> int` is not in general going to yield the same integer again, though it will in the current case because we're dealing with small ints. So not great, but will probably be ok.

include/deal.II/lac/la_parallel_vector.templates.h
marcfehling: Compilation fails if I use a data type for `LA:d:V` that it was not specialized for.

marcfehling: As this is unrelated to this PR if we use `float`, I'll revert this change.

marcfehling:
Ready for tests!

bangerth (approved):

PR #11858, opened 2021-03-06
Enable set up of active_cell_index_partitioner for p:s:T

peterrum:
~For now only locally-owned cells and a single halo are considered.~

Needed by #11857.

### DIFF
--- a/include/deal.II/grid/tria.h
+++ b/include/deal.II/grid/tria.h
@@ -4234,8 +4234,13 @@ Triangulation<dim, spacedim>::load(Archive &ar, const unsigned int)
   // this here. don't forget to first resize the fields appropriately
   {
     for (auto &level : levels)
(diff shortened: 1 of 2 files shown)


drwells (approved):

tjhei:
Not sure what is going on with CU here: we quicktests with mpi are timing out, even after restart.

peterrum:
OK. The test actually dead-locked because of https://github.com/dealii/dealii/pull/10529 (`p:s:T` without artificial cells). I now have also included a specialization for this case.

kronbichler (approved):

tjhei:
for the record, which of the tests was that?

peterrum:
> for the record, which of the tests was that?

`tests/quick_tests/step-metis.cc` and others that did use p:s:T without artificial cells.

kronbichler:
I guess we are ready here, let us merge now.

PR #11859, opened 2021-03-06
Make active_cell_index_partitioner to std::shared_ptr

peterrum:
While reviewing #11857, I noticed that it would be better to make `active_cell_index_partitioner` and `level_cell_index_partitioners` to `std::shared_ptr`, since this is the way it is needed to set up global vectors (in my opinion the main purpose of the partitioner).

### DIFF
--- a/tests/grid/global_ids_01.cc
+++ b/tests/grid/global_ids_01.cc
@@ -45,7 +45,8 @@ test(int n_refinements, MPI_Comm comm)
       deallog << cell->id() << " -> " << cell->subdomain_id() << " "
               << cell->global_active_cell_index() << std::endl;
 
-  const auto &part = tria.global_active_cell_index_partitioner();
+  const Utilities::MPI::Partitioner &part =
+    *tria.global_active_cell_index_partitioner().lock();
 
   part.locally_owned_range().print(deallog);
   part.ghost_indices().print(deallog);
--- a/tests/grid/global_ids_03.cc
+++ b/tests/grid/global_ids_03.cc
@@ -52,7 +52,8 @@ test(int n_refinements, MPI_Comm comm)
           deallog << cell->id() << " -> " << cell->level_subdomain_id() << " "
                   << cell->global_level_cell_index() << std::endl;
 
-      const auto &part = tria.global_level_cell_index_partitioner(l);
+      const Utilities::MPI::Partitioner &part =
+        *tria.global_level_cell_index_partitioner(l).lock();
 
       part.locally_owned_range().print(deallog);
       part.ghost_indices().print(deallog);
--- a/include/deal.II/distributed/tria_base.h
+++ b/include/deal.II/distributed/tria_base.h
@@ -206,14 +206,14 @@ namespace parallel
      * Return partitioner for the global indices of the cells on the active
      * level of the triangulation.
      */
-    const Utilities::MPI::Partitioner &
+    const std::weak_ptr<const Utilities::MPI::Partitioner>
     global_active_cell_index_partitioner() const;
 
     /**
      * Return partitioner for the global indices of the cells on the given @p
      * level of the triangulation.
      */
-    const Utilities::MPI::Partitioner &
+    const std::weak_ptr<const Utilities::MPI::Partitioner>
     global_level_cell_index_partitioner(const unsigned int level) const;
 
     /**
@@ -356,12 +356,14 @@ namespace parallel
       /**
        * Partitioner for the global active cell indices.
        */
-      Utilities::MPI::Partitioner active_cell_index_partitioner;
+      std::shared_ptr<const Utilities::MPI::Partitioner>
+        active_cell_index_partitioner;
 
       /**
        * Partitioner for the global level cell indices for each level.
        */
-      std::vector<Utilities::MPI::Partitioner> level_cell_index_partitioners;
+      std::vector<std::shared_ptr<const Utilities::MPI::Partitioner>>
+        level_cell_index_partitioners;
 
       NumberCache();
     };
--- a/source/distributed/tria_base.cc
+++ b/source/distributed/tria_base.cc
@@ -443,7 +443,8 @@ namespace parallel
     is_ghost.add_indices(is_ghost_vector.begin(), is_ghost_vector.end());
 
     number_cache.active_cell_index_partitioner =
-      Utilities::MPI::Partitioner(is_local, is_ghost, this->mpi_communicator);
+      std::make_shared<const Utilities::MPI::Partitioner>(
+        is_local, is_ghost, this->mpi_communicator);
 
     // 6) proceed with multigrid levels if requested
     if (this->is_multilevel_hierarchy_constructed() == true)
@@ -536,9 +537,8 @@ namespace parallel
                                  is_ghost_vector.end());
 
             number_cache.level_cell_index_partitioners[l] =
-              Utilities::MPI::Partitioner(is_local,
-                                          is_ghost,
-                                          this->mpi_communicator);
+              std::make_shared<const Utilities::MPI::Partitioner>(
+                is_local, is_ghost, this->mpi_communicator);
           }
       }
 
@@ -598,14 +598,14 @@ namespace parallel
 
 
   template <int dim, int spacedim>
-  const Utilities::MPI::Partitioner &
+  const std::weak_ptr<const Utilities::MPI::Partitioner>
   TriangulationBase<dim, spacedim>::global_active_cell_index_partitioner() const
   {
     return number_cache.active_cell_index_partitioner;
   }
 
   template <int dim, int spacedim>
-  const Utilities::MPI::Partitioner &
+  const std::weak_ptr<const Utilities::MPI::Partitioner>
   TriangulationBase<dim, spacedim>::global_level_cell_index_partitioner(
     const unsigned int level) const
   {


bangerth:
The question in these sorts of design decisions is whether the object should have an unambiguous owner, or whether the object is collectively owned. Here, it would seem to me that the object really only makes sense (has a reason to live) if the triangulation is still alive -- what would be the use for this kind of partitioner if the triangulation has already been destroyed?

In other words, can you foresee use cases where one can legitimately use the partitioner if the triangulation has already disappeared?

peterrum:
This is how it is done in `MatrixFree`:

https://github.com/dealii/dealii/blob/ac854ef67344a07a72617e8def997c8fe9717eed/include/deal.II/matrix_free/matrix_free.h#L1560-L1561

and how it is needed by the constructor of `LinearAlgebra::parallel::Vector` 

https://github.com/dealii/dealii/blob/ac854ef67344a07a72617e8def997c8fe9717eed/include/deal.II/lac/la_parallel_vector.h#L319-L320

without having to copy the whole index sets.

> In other words, can you foresee use cases where one can legitimately use the partitioner if the triangulation has already disappeared?

No. But would nevertheless make the change to make the code consistent.

kronbichler:
In my opinion definitely makes sense to share the `Partitioner` between different vectors - it follows the philosophy "one X, many Y" (one triangulation, many DoFHandler objects), and there is a clear concept why it is needed such as cheaper initialization and to support all operations between vectors of the same partitioner. As opposed to the triangulations or DoFHandler where we keep a `SmartPointer` (subscriptor) in the user class, the ownership is within the vectors also makes sense to me because one rarely needs the partitioner outside the vector context (`MatrixFree` is the obvious exception). 

When it comes to the particular case of the active cell index in the triangulation, one might argue that it is not needed after the triangulation changes. However, all our vectors are defined on the purely algebraic level without keeping track of where they come from. We do not disconnect them once the underlying `DoFHandler` goes out of scope, either. As to the application, one could imagine that some data transfer between two states of a triangulation is implemented in an algebraic way without `SolutionTransfer` facilities. So I see the code only getting simpler and with less surprises.

kronbichler (approved):

bangerth:
The traditional idiom to use, if there is a clear owner and subsidiary objects that refer to the owner is to use a member object and references to it. That clearly has the disadvantage that if the owner dies, the references become stale.

The modern idiom to use is to have the owner use `std::shared_ptr` and the subsidiary listeners receive information about the object using `std::weak_ptr`: https://en.cppreference.com/w/cpp/memory/weak_ptr . Every time you need to work with the object from the listener, you convert it to a `std::shared_ptr`, but in between you keep it as the weak pointer (i.e., the member variable of the listener remains a `std::weak_ptr`, which within member functions is temporarily converted to a `std::shared_ptr`). This way you will learn if the owner of the pointer has died at some point, and you can generate a meaningful error messages if you still have a dangling reference to it: You know that the reference is dangling because the conversion from weak to shared pointer fails.

bangerth:
I wished that we had had that shared/weak pointer pair back in 1997. Then every user program would have a `std::shared_ptr<Triangulation<dim>>` and `DoFHandler` would store a weak pointer to it. We could then also get rid of the `SmartPointer`/`Subscriptor` class which I never particularly loved. But it is what it is, I guess :-)

peterrum:
I have switched to `std::weak_ptr`. The annoying thing about this class is that there is not implicit conversion and you cannot work with it directly. The user is forced to us the `lock()` function.

tests/grid/global_ids_01.cc
```diff
@@ -45,7 +45,7 @@ test(int n_refinements, MPI_Comm comm)
       deallog << cell->id() << " -> " << cell->subdomain_id() << " "
               << cell->global_active_cell_index() << std::endl;
 
-  const auto &part = tria.global_active_cell_index_partitioner();
+  const auto &part = *tria.global_active_cell_index_partitioner().lock();
```
bangerth, line 48: Here and in the other place below, I had to look up what `lock()` returns to be able to tell what the type of `part` should be. I would just be explicit instead of using `auto` to make that easier to read.

bangerth (commented):
Looking good to me, with the possible one exception (in two places).

tests/grid/global_ids_01.cc
peterrum: Done :)

bangerth: You should be off work at this time of day!

bangerth (approved):

tests/grid/global_ids_03.cc
```diff
@@ -52,7 +52,8 @@ test(int n_refinements, MPI_Comm comm)
           deallog << cell->id() << " -> " << cell->level_subdomain_id() << " "
                   << cell->global_level_cell_index() << std::endl;
 
-      const auto &part = tria.global_level_cell_index_partitioner(l);
+      const Utilities::MPI::Partitioner &part =
+        *tria.global_level_cell_index_partitioner(l).lock();
```
kronbichler, line 56: I'm late to the party, but I don't really understand the concept that is used here. Maybe it's only my partial understanding of what `weak_ptr.lock()` really does, so I'd like someone to help me with my concern:

In short, I can't make sense of the value `weak_ptr` brings to this context. This change of return type surely helps the user who writes `const auto partitioner = tria.global_active_cell_index_partitioner()`, but that's about it. All other situations I image, most importantly the construction and reinitialization of `LinearAlgebra::distributed::Vector` through an `std::shared_ptr<const Partitioner>`, only get more tedious to write by saying `...._index_partitioner().lock()`. But what do we gain then from the weak pointer? We convert it to a shared pointer immediately, which is never a problem because at the place the triangulation must is still be alive because to allow access to its member. But then inside the vector the shared pointer keeps the partitioner alive, irrespective of the life time of the triangulation.

In my opinion, the weak pointer would make more sense as a member variable somewhere, not as a return value. Apart from `auto` capture of course and possibly educating the user to keep an eye on the lifetime of some objects. However, I think it is clear that we don't want to change the vector interface as a linear algebra object whose lifetime does not depend on where the numbering came from, so where is the value?

Not that I'm completely opposed to returning a weak pointer instead of a shared pointer, but at this specific place it seems to mostly provide theoretical value at the cost of some nuisance in practice. (Not all users will know how to handle a weak pointer and get the desired shared pointer other interfaces expect out of it, so we need to write additional documentation somewhere to educate about the not so well-known features of STL, and so on.)

peterrum: A `std::weak_ptr` would be useful if we would make also the internal partitioner of `LA:p:V`to a `weak_ptr`; once it is used, e.g. for communication, (it is converted to a `shared_ptr` and) it is checked if the partitioner and the triangulation class are still living. 

But as long as we are not using it this way (maybe we never what to use it this way), we won't have any benefit, as far as see it. In our use case, we immediately convert the weak pointer to a shared pointer:
```cpp
LinearAlgebra::parallel::Vector<double> vector(tria.global_active_cell_index_partitioner().lock());
``` 
so that we actually need to write more code (`.lock()`).

I would prefer to use `std::shared_ptr` because of the reason listed in https://github.com/dealii/dealii/pull/11859#issuecomment-792071475. But I see @bangerth's reasoning: this might be not beneficial right now, but maybe in the future!?

kronbichler:
> A `std::weak_ptr` would be useful if we would make also the internal partitioner of `LA:p:V` to a `weak_ptr`; once it is used, e.g. for communication, (it is converted to a shared_ptr and) it is checked if the partitioner and the triangulation class are still living.

That's also what I was having in my mind as an example of a *member variable somewhere*. But I don't think it's reasonable in the short and medium run for two reasons: (a) All external vector classes we interface to are autonomous in the sense that once initialized, they can never get into an invalid state just because part of information used to set up the vector disappeared. To me, this is a feature, so I would like our vectors to provide the same behavior. While the *feature* of allowing to attach a weak pointer and track the origin might help uncovering some subtle bugs in some user codes, I think users are relatively experienced along those lines, as errors are most often simple in the sense that you will get some size mismatch and an easy-to-follow-up assertion throwing. (b) With `weak_ptr`, we would also need to call `weak_ptr::lock()` inside `LA:d:V::operator()` for the individual member access where we translate from global to local indices. If I remember the implementation correctly getting the shared pointer means incrementing an atomic counter somewhere and decrementing right after use again, so this would in essence mean we shut down all our HPC ambitions with deal.II's own vectors (or transition the library to uniformly use MPI-local indices for access).

bangerth:
I should add here that my argument about using `std::weak_ptr` started from the original patch, in which the return type of the functions giving access to the partitioners was changed from a const reference to a pointer. When you return a reference, ownership is clear: The object whose member function is called owns the object. If the calling site stores the reference, then it is also clear that it does not own the object. If it chooses to store a copy of the object being referenced, it is also clear who owns the copy.

But if you return a `shared_ptr`, these semantics are lost: Now everyone owns it, and I think that while that is often *convenient*, it does not encourage clean design where semantic relationships such as "this partitioner corresponds to the cells of a triangulation" are reflected: The owner of the partitioner should be the same object that describes the triangulation.

What @kronbichler asks, if I understand correctly, was whether the original change from returning a const reference to a pointer makes sense. That I have no idea about :-)

kronbichler:
I agree that we should return a pointer to make the initialization cost small, so it makes sense along that way. So my concern was about the ownership and giving the vector class free hands (which we do, because we turn the weak pointer into a shared pointer). Anyway, I do not feel strongly here, so I think we can keep this return type as is, given that we have a clear ownership-user relation in this particular case, and just try to go with `lock()` here. If it turns out useful, I would happily support to extend this usage.

PR #11862, opened 2021-03-06
GridOut::write_mesh_per_processor_as_vtu() for simplex

peterrum:
While trying to port `GridOut::write_mesh_per_processor_as_vtu()` to simplex meshes, I observed the following two issues:

- `DataOutBase::Patch::points_are_available = false` is not working yet:
https://github.com/dealii/dealii/blob/ac854ef67344a07a72617e8def997c8fe9717eed/source/grid/grid_out.cc#L3673
- for creating the pvtu record `build_patches()` is called without mapping (and without `DoFHandler` attached so that the mapping index cannot be queried from the cell): 
https://github.com/dealii/dealii/blob/ac854ef67344a07a72617e8def997c8fe9717eed/source/grid/grid_out.cc#L3742

### DIFF
--- a/source/base/data_out_base.cc
+++ b/source/base/data_out_base.cc
@@ -697,14 +697,14 @@ namespace
   //----------------------------------------------------------------------//
   // For a given patch, compute the node interpolating the corner nodes linearly
   // at the point (xstep, ystep, zstep)*1./n_subdivisions. If the points are
-  // saved in the patch.data member, return the saved point instead
+  // saved in the patch.data member, return the saved point instead.
   template <int dim, int spacedim>
   inline Point<spacedim>
-  compute_node(const DataOutBase::Patch<dim, spacedim> &patch,
-               const unsigned int                       xstep,
-               const unsigned int                       ystep,
-               const unsigned int                       zstep,
-               const unsigned int                       n_subdivisions)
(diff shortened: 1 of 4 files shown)


peterrum:
`GridOut::write_mesh_per_processor_as_vtu()` works now also for simplex meshes. Ready once #11924 and #11925 are merged.

source/base/data_out_base.cc
```diff
@@ -2631,17 +2631,33 @@ namespace DataOutBase
...
+                  Assert(
+                    patch.reference_cell != ReferenceCells::Pyramid,
+                    ExcMessage(
+                      "Pyramids need different ordering of the vertices, which is not implemented yet here."));
+
+                  node = patch.vertices[point_no];
+                }
+
+              return node;
+            };
```
drwells, line 2656: I'd like to simplify the control flow here - it's hard to follow when we have a lambda named `compute_node` and a top-level function named `compute_node`.

Could we move the lambda into a new top-level function named `compute_simplex_node` and then rename `compute_node` to `compute_hypercube_node` or something similar? That would make this a lot easier to understand.

drwells (changes requested):

source/base/data_out_base.cc
peterrum: Done!

drwells (approved):

PR #11863, opened 2021-03-07
Use correct type for FE indices in DoFH::communicate_active_fe_indices().

marcfehling:
Excerpt from #11857.

For internal storage in `DoFHandler`, we have the type alias `active_fe_index_type` which translates to `unsigned short int`. The `DoFCellAccessor` interface uses `unsigned int` for this purpose.

We should use the correct data type for internal purposes.

### DIFF
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -17,6 +17,7 @@
 
 #include <deal.II/base/geometry_info.h>
 #include <deal.II/base/memory_consumption.h>
+#include <deal.II/base/mpi.templates.h>
 
 #include <deal.II/distributed/cell_data_transfer.templates.h>
 #include <deal.II/distributed/shared_tria.h>
@@ -1715,6 +1716,9 @@ namespace internal
             dof_handler.hp_capability_enabled == true,
             (typename DoFHandler<dim, spacedim>::ExcOnlyAvailableWithHP()));
 
+          using active_fe_index_type =
+            typename dealii::DoFHandler<dim, spacedim>::active_fe_index_type;
+
           if (const dealii::parallel::shared::Triangulation<dim, spacedim> *tr =
                 dynamic_cast<
                   const dealii::parallel::shared::Triangulation<dim, spacedim>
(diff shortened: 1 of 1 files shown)


source/dofs/dof_handler.cc
```diff
@@ -17,6 +17,7 @@
 
 #include <deal.II/base/geometry_info.h>
 #include <deal.II/base/memory_consumption.h>
+#include <deal.II/base/mpi.templates.h>
```
marcfehling, line 20: We need this because of:
https://github.com/dealii/dealii/blob/538ef0012a8489116311fd4302f6ca03c4d842da/source/dofs/dof_handler.cc#L1744-L1746

bangerth (approved):

source/dofs/dof_handler.cc
peterrum: Else you get an undefined reference? If yes, I would suggest to instantiate the function for `unsigned short int`.

source/dofs/dof_handler.cc
```diff
@@ -1766,15 +1770,17 @@ namespace internal
               // to have functions that can pack and unpack the data we want to
               // transport -- namely, the single unsigned int active_fe_index
               // objects
-              auto pack = [](const typename dealii::DoFHandler<dim, spacedim>::
-                               active_cell_iterator &cell) -> unsigned int {
+              auto pack =
+                [](const typename dealii::DoFHandler<dim, spacedim>::
+                     active_cell_iterator &cell) -> active_fe_index_type {
```
peterrum, line 1775:
```suggestion
                     active_cell_iterator &cell) {
```

source/dofs/dof_handler.cc
```diff
@@ -1766,15 +1770,17 @@ namespace internal
...
 
-              auto unpack = [&dof_handler](
-                              const typename dealii::DoFHandler<dim, spacedim>::
-                                active_cell_iterator &cell,
-                              const unsigned int      active_fe_index) -> void {
+              auto unpack =
+                [&dof_handler](
+                  const typename dealii::DoFHandler<dim, spacedim>::
+                    active_cell_iterator &   cell,
+                  const active_fe_index_type active_fe_index) -> void {
```
peterrum, line 1783:
```suggestion
                  const active_fe_index_type active_fe_index) {
```

bangerth:
I think the `#include` makes sense. We don't want to make assumptions about what the `active_fe_index` data type really is.

source/dofs/dof_handler.cc
marcfehling: Let's leave it like this to be independent of the actual `active_fe_index_type`.

source/dofs/dof_handler.cc
marcfehling: I prefer to be explicit. That also applies to return types :)

source/dofs/dof_handler.cc
marcfehling: Same here.

peterrum (approved):

PR #11867, opened 2021-03-08
Add new MappingQCache::initialize() function

peterrum:
... that transform points.

As far as I see the implementation should be generic enough and thread safe. Have I missed anything? I will add a test later.


kronbichler:
Can you explain the motivation to write this function in the current form? I think the direction is useful indeed, but two changes would make the interface even more powerful:
- It might be unexpected to give the user the points of a `MappingQ1` that should be replaced. If I imagine some generic functionality that merely *deforms* points I would have intuitively thought about `MappingQGeneric` point instead. In other words, I think it will be more natural to the user if the given points respect the manifold in the sense of the underlying mapping degree, and do the transformation on top of that.
- I think a number of more complicated codes will want to have both the `cell` and the point `p` as the argument of the lambda to provide their answer regarding deformed points. I think about the case where I want to implement some clever deformation that is attached by `manifold_id` (or any other mechanism that lives in user code). From a single point coming out of the base geometry + `MappingQGeneric`, one would lose all that information. Of course, it requires the user to add a second argument to a function also in cases when there is only a simple analytic function describing the *full* domain, but I think that case that one uses different descriptions for different parts of the mesh are more representative of complex user codes (and two functions make the interface too big).

source/fe/mapping_q_cache.cc
```diff
@@ -125,6 +129,52 @@ MappingQCache<dim, spacedim>::initialize(
...
+        auto &fe_values = fe_values_all.get();
+        if (fe_values.get() == nullptr)
+          fe_values = std::make_unique<FEValues<dim, spacedim>>(
+            mapping_base, fe, quadrature, update_quadrature_points);
+
+        return *fe_values;
+      }();
+
+      fe_values.reinit(cell);
+      auto points = fe_values.get_quadrature_points();
```
kronbichler, line 165: For the case we prefer to use the points of `MappingQGeneric`, I believe we could simplify this quite a bit (I did not check if it runs well or if I overlooked something here):
```suggestion
  this->initialize(
    tria,
    [&](const typename Triangulation<dim, spacedim>::cell_iterator &cell) {
      auto points = this->MappingQGeneric<dim>::compute_mapping_support_points(cell);
```

peterrum:
> Can you explain the motivation to write this function in the current form?

The intention of writing this function was simply to deform a hyper-cube geometry according to an analytical function. For this purpose it worked very well.

Adding a second argument to the user-provided lambda sounds reasonable. 

Regarding the mapping, I am not sure if it should be `MappingQGeneric` or it should be even an input argument by the user. I like the code you have given for MappingQGeneric`, which could be a specialization of the code if we given the users complete flexibility ragarding the mapping. 

Furthermore, I am not sure if we should introduce a flag for adding the computed values to the point positions (absolute vs. relative values). For the given function, this task could be simply delegated to the user provided function, but once one wants to deal with global vector that the describe relative or absolute positions (as in the case of `MappingFEFields`) this is not possible anymore. 

@nfehn What is your opionion?

nfehn:
I don't like the interface so much. On thing is the `MappingQ1`, but the main point is that the interface again contains a lambda, why not `dealii::Function<dim>`? The latter would be way more user friendly in my opinion.

We did a lot of work in ExaDG (https://github.com/exadg/exadg/) in providing user friendly interfaces for `MappingQCache`, which - together with @kronbichler - I planned to integrate into deal.II once the interfaces have proven stable and useful (there is one open PR in ExaDG on this topic, https://github.com/exadg/exadg/pull/37). In ExaDG, there are a lot of applications in terms of moving mesh problems with analytical mesh movement functions, or Poisson-type or elasticity-type (pseudo-solid) mesh movement PDE solvers. For me, a main motivation was to remove the lambda from the interface of `MappingQCache`. Together, we should port the code from ExaDG to deal.II instead of doing these developments again here.

I think a good interface would be

```
  void
  initialize(std::shared_ptr<Mapping<dim> const> mapping,
               Triangulation<dim> const & triangulation,
               std::shared_ptr<Function<dim>> displacement_function);
```
The parameter `mapping` describes the current geometry (undeformed reference state), and `displacement_function` a displacement on top. If `mapping.get() == 0`, this implies that the `displacement_function` describes absolute coordinates. This gives a lot of flexibility to the user, yet using only standard dealii data structures in the interface.

kronbichler:
@nfehn these are good points and I think the interface would make sense, especially the additional mapping argument. So yes, that very well motivates an additional reinit function taking a `dealii::Function` and the mapping, to give the user control what should happen.

However, I see the main use case for the lambda for something that cannot reasonably by represented with `dealii::Function`, though. As described https://github.com/dealii/dealii/pull/11867#issuecomment-793639510 above, you might want to compute some information based on a `cell_iterator` and `Point` argument, because the user code can compute something based on the cell. For example for some auxiliary deformation function, you really do not want the `Function` interface with just a point. Locating the right cell owner from just the point location in real space is a mess (we do that for some coupling algorithms, including the code in #11804) and despite all optimizations we did in the last year or two, it remains an expensive step to do for each point. So when we have the context of the cell right in our hands, we enable the user to use that.

kronbichler:
Regarding the lambda, there are different tastes and some people like them more than others. It's valuable to have your opinion @nfehn on this, because your view is likely shared by a good share of users, so going through this discussion improves the experience for all of us.

peterrum:
I have pushed a revised version. I think the interface fulfills the requirements. After talking to @nfehn, we have dropped the `std::shared_ptr`.

@nfehn and I have 2 general questions:
- Why is the mapping in the second place in the current `intialize()` function? Can we switch the order?
- How expensive is to call `FETools::hierarchic_to_lexicographic_numbering()`?

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -111,6 +112,38 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
                const typename Triangulation<dim, spacedim>::cell_iterator &)>
                &compute_points_on_cell);
 
+  /**
+   * Initialize the data cache by computing the mapping support points for all
+   * cells (on all levels) of the given triangulation and a given @p mapping
+   * and transforming these points via the function @p transformation_function.
```
bangerth, line 118: Out of curiosity: This is a way to move points, but how are you addressing the fact that the additional move would also lead to a different Jacobian (and its determinant)?

source/fe/mapping_q_cache.cc
```diff
@@ -125,6 +129,97 @@ MappingQCache<dim, spacedim>::initialize(
...
+          points = m->compute_mapping_support_points(cell);
+        }
+      else
+        {
+          // get FEValues (thread-safe); in the case that this thread has not
+          // created a an FEValues object yet, this helper-function also
+          // creates one with the right quadrature rule
+          auto &fe_values = [&]() -> FEValues<dim, spacedim> & {
+            auto &fe_values = fe_values_all.get();
+            if (fe_values.get() == nullptr)
```
bangerth, line 164: This seems like a gratuitous use of a lambda function :-) The only thing you do is immediately call it, then assign its result to a reference, and then call two functions on the reference. You might as well duplicate the two calls into the `if` and `else` branches -- the overall number of lines is going to be less than what you have right now, and I bet it will be easier to read :-)

include/deal.II/fe/mapping_q_cache.h
kronbichler: Indeed, the beauty of this class is that you only need to move points; since this is a polynomial mapping, that's all you need, and the normals, Jacobians, etc. are computed from the points (in terms of the manifold paper, we have stored the result of the point computation in section 3.2 with "new point" and don't need anything else).

kronbichler:
> Why is the mapping in the second place in the current `intialize()` function? Can we switch the order?

You mean because we typically put the mapping first in terms of arguments for many functions in deal.II? I guess I did not pay overly good attention to the design - I feel it is most intuitive to first define a triangulation and then the mapping (that somehow manipulates/adds the points on the cells of a triangulation) in terms of dependencies. But if we are mostly consistent in the other way, we should of course change this. Since you @peterrum have gone through many interface you probably can tell best if this is an outlier, I would have to check and grep. In any case, we could deprecate this order and introduce the one with the mapping first.

> How expensive is to call `FETools::hierarchic_to_lexicographic_numbering()`?

Pretty cheap in the sense that you pay 95% of the time for the memory allocation (or maybe it is two because we return the inverse one, I don't recall). I guess you refer to the fact that this is done for each cell? I would not worry at that place because there is at least one more allocation for the support point computation, maybe many more inside the mapping calculations.

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -111,6 +112,38 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
+             const bool is_displacement_function = false);
+
+  /**
+   * The same as above but taking a dealii::Function object.
+   */
+  void
+  initialize(const Triangulation<dim, spacedim> &tria,
+             const Mapping<dim, spacedim> &      mapping,
+             const Function<spacedim> &          transformation_function,
+             const bool is_displacement_function = false);
```
nfehn, line 145: Generally, I would prefer if the default setting is `true` independently of which logic is implemented as default. 

In ALE literature, the nomenclature is "mesh movement/motion", meaning displacements of the mesh as compared to a "undeformed" reference configuration (i.e. the mesh described by mapping). Hence, I would prefer something like `const bool function_describes_relative_displacement = true`. 

If you want to have the default state as is, I would write `const bool function_maps_to_absolute_coordinates = true`.

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -111,6 +112,38 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
                &compute_points_on_cell);
 
+  /**
+   * Initialize the data cache by computing the mapping support points for all
+   * cells (on all levels) of the given triangulation and a given @p mapping
+   * and transforming these points via the function @p transformation_function.
+   *
+   * If @p is_displacement_function is set to true, the return value of the
+   * function is interpreted as relative deformation and the result is added
+   * to the original point.
```
nfehn, line 122: Related to the other comment and independently of which logic is implemented as default, I believe both states (true/false) should be explained explicitly. Currently, the default value is false but the docu explains the true-case.

include/deal.II/fe/mapping_q_cache.h
bangerth: Ah, I see -- thanks for educating me!

include/deal.II/fe/mapping_q_cache.h
kronbichler: I agree with the `function_describes_relative_displacement` name - it is very easy to get confused by whether the value that gets used is an absolute coordinate position solely determined by the function value (argument would be `false`) or a relative displacement against the position determined by the mapping, so being as explicit as possible both in the argument name and in the comment is important.

peterrum:
> Since you @peterrum have gone through many interface you probably can tell best if this is an outlier, I would have to check and grep. In any case, we could deprecate this order and introduce the one with the mapping first.

There are two styles in deal.II. Style 1) put `Mapping` into the first place (`FEValues`, `MatrixFree`, `DoFTools`, a lot of functions in `VectorTools`) and style 2) put `Mapping` as one of the last (optional) parameters (some of the functions in `VectorTools`). As far as I can tell, in our case the mapping belongs more into the first category and I would propose to switch the order.

peterrum:
I switched the order of the parameters and have (early) deprecated the old function. I also used the opportunity to generalize the new version of the old function from `MappingQGeneric` to `Mapping`. Should I move this into a separate PR?

_Note: I am thinking about moving the setup of `FEValues` into an internal function. But let me postpone to do that once we have added the remaining `initialize` functions._

doc/news/changes/incompatibilities/20210310Munch
```diff
@@ -0,0 +1,4 @@
+Changed: The order of parameters has been chaned in MappingQCache::initialize() so that
+Mapping is in the first place.
```
kronbichler, line 2:
```suggestion
Changed: The order of parameters has been switched in MappingQCache::initialize(), with the
Mapping now being the first argument.
```

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -111,6 +125,40 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
                &compute_points_on_cell);
 
+  /**
+   * Initialize the data cache by computing the mapping support points for all
+   * cells (on all levels) of the given triangulation and a given @p mapping
+   * and transforming these points via the function @p transformation_function.
+   *
+   * The bool @p function_describes_relative_displacement is in the default
+   * case false and indicates that the function @p transformation_function
+   * maps to absolute coordinates.If the parameter is set to true, the return
```
kronbichler, line 135:
```suggestion
   * maps to absolute coordinates. If the parameter is set to true, the return
```

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -111,6 +125,40 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
+   * value of the function is interpreted as relative deformation and the result
+   * is added to the original point.
```
kronbichler, line 137:
```suggestion
   * is eventually added to the original point for the definition of the support
   * points of this class.
```

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -111,6 +125,40 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
+   *
+   * This function calls the above function so that the comments there regarding
+   * threading are also true here.
```
kronbichler, line 140:
```suggestion
   * This function calls the previous function so the comments regarding
   * threading listed above apply also here.
```

source/fe/mapping_q_cache.cc
```diff
@@ -70,18 +74,63 @@ MappingQCache<dim, spacedim>::preserves_vertex_locations() const
...
+      if (mapping_q_generic != nullptr &&
+          this->get_degree() == mapping_q_generic->get_degree())
+        {
+          return mapping_q_generic->compute_mapping_support_points(cell);
+        }
+      else
+        {
+          // get FEValues (thread-safe); in the case that this thread has not
+          // created a an FEValues object yet, this helper-function also
+          // creates one with the right quadrature rule
```
kronbichler, line 102: I remember that we had problems with `WorkStream` before regarding threading when only using only a single `ThreadLocalStorage<FEValues>` instance rather than `ThreadLocalStorage<std::list<FEValues>>`. In the general case, it could happen that the code executed by a thread could itself call into threaded code (here, it could possibly be something in the constructor of `FEValues`), and then TBB would decide to suspend the current thread until the result of the spawned workers would complete. Since then the current thread is idle, it might start the work on another cell and try to acquire an `FEValues` again. I think we are lucky enough here, in the sense that we have the `fe_values.get() == nullptr` check here, and this would simply be false when TBB starts to get crazy and mixes inner parallelism with outer one. It feels a bit brittle here, but I guess this is the best we can do. It would of course be even better if we had a function in `Mapping` that could be evaluated on a set of points without all that extreme overhead of `FEValues`, so we could slip the thread local storage altogether.

kronbichler (approved):
This looks good to me, thank you for the updates.

include/deal.II/fe/mapping_q_cache.h
nfehn: I don't understand "definition of the support points"

source/fe/mapping_q_cache.cc
kronbichler: > we could slip the thread local storage altogether.

Note that we have this kind of function in the mapping interface: https://github.com/dealii/dealii/blob/20d2280dadce1e217e9b3b9b72a46637f49d3e2e/include/deal.II/fe/mapping.h#L476-L480
So the question is if we trust that this unstructured evaluation routine is fast enough. Given the optimizations we did for `MappingQGeneric` previously, and since I can't see that we would be using a faster tensor product structure in this case, I would say yes.

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -111,6 +125,40 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
+   * @note The cache is invalidated upon the signal
+   * Triangulation::Signals::any_change of the underlying triangulation.
+   */
+  void
+  initialize(const Mapping<dim, spacedim> &      mapping,
+             const Triangulation<dim, spacedim> &tria,
+             const std::function<Point<spacedim>(
+               const typename Triangulation<dim, spacedim>::cell_iterator &,
+               const Point<spacedim> &)> &       transformation_function,
+             const bool function_describes_relative_displacement = false);
```
nfehn, line 151: I guess you insist to have the default value false since you already changed the name? Then I would suggest to remove the default value. Default values also have the disadvantage that when extending the interface later, additional parameter must have default values as well., and that default values are error prone.

include/deal.II/fe/mapping_q_cache.h
kronbichler: 'definition' is maybe the wrong word, so what about `for the support points eventually used by this class`? What I wanted to express is a reminder to the user about what happens with the result of the computation provided here.

source/fe/mapping_q_cache.cc
peterrum: I am not opposed to the idea. But, I might need `FEValues` in PR #11882 for the general case. So we move the discussion there.

peterrum:
@nfehn @kronbichler I have incorporated the suggestions. Thanks!

kronbichler (approved):

PR #11869, opened 2021-03-08
Add bool type to HDF5

dangars:
This adds the bool type to HDF5.

Currently the bool type is only supported for the function `get_attribute()`.

I still have to do some testing.


include/deal.II/base/hdf5.h
```diff
@@ -1587,29 +1591,6 @@ namespace HDF5
 
 
 
-  template <>
-  inline bool
-  HDF5Object::get_attribute(const std::string &attr_name) const
```
bangerth, line 1592: What;s the reason for getting rid of this specialization?

bangerth (commented):
The change to `get_hdf5_datatype()` clearly makes sense. The other one I don't understand.

I think it would be nice if we had a test for this functionality. Do you remember how to write tests?

include/deal.II/base/hdf5.h
dangars: This specialization is not necessary because if I added the bool support globally in `get_hdf5_datatype()`.

The bool support will be automatically added to `get_attribute()`, `set_attribute()`, `write_dataset()`, etc. That being said, I still have to test it!

Ideally we should get rid of all the specializations. But it is tricky to get rid of the `std_string` specialization because `std_string` is a UTF8 variable string data type. Therefore the HDF5 code is quite different.
```
template <>
inline std::string
  HDF5Object::get_attribute(const std::string &attr_name) const
```
I think that specializations are difficult to understand for somebody who didn't write the code, I think that C++20 `if constexpr` is cleaner.

How about I write two versions one with C++20 `if constexpr` and another one with the specialization? When deal.II will move to C++20 you can get rid of the old code. I'll use macros to separate the two versions.

dangars:
I'll make a test :)

I just saw this google group discussion https://groups.google.com/u/1/g/dealii/c/MVoUIS745dw

And I wanted to quickly write the code that fixes the problem!

include/deal.II/base/hdf5.h
masterleinad: That's more code to maintain so if we need the specializations anyway, I prefer not to also have a separate version using `if constexpr`.

dangars: Sounds good!

dangars:
I wrote a test but it is not working. I think that `ArrayView` does not work with bool. I'm happy to write a patch to make `ArrayView` work with bool, but I'm not sure where to start. I don't really understand why I get this error with bool and I don't get the error with double of int.

Here is the error:
```
1490: In file included from /home/daniel/repositories/dealii/include/deal.II/base/hdf5.h:23:                                                                                                                         
1490: /home/daniel/repositories/dealii/include/deal.II/base/array_view.h:980:39: error: 'this' argument to member function 'data' has type 'const std::vector<bool>', but function is not marked const               
1490:   return ArrayView<const ElementType>(vector.data(), vector.size());                                                                                                                                           
1490:                                       ^~~~~~                                                                                                                                                                   
1490: /home/daniel/repositories/dealii/include/deal.II/base/hdf5.h:2098:20: note: in instantiation of function template specialization 'dealii::make_array_view<bool>' requested here                                
1490:                    make_array_view(data).data());                                                                                                                                                              
1490:                    ^                                                                                                                                                                                           
1490: /home/daniel/repositories/dealii/tests/base/hdf5_07.cc:101:19: note: in instantiation of function template specialization 'dealii::HDF5::DataSet::write_hyperslab<std::vector<bool, std::allocator<bool> > >' $
equested here                                                                                                                                                                                                        
1490:           dataset.write_hyperslab(data, hyperslab_offset, hyperslab_count);                                                                                                                                    
1490:                   ^                                                                                                                                                                                            
1490: /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/stl_bvector.h:946:7: note: 'data' declared here                                                                                          
1490:       data() _GLIBCXX_NOEXCEPT { }                                                                                                                                                                             
1490:       ^                                                                                                                                                                                                        
1490: 1 error generated.   
```

I use `ArrayView` to get a pointer to the data:
https://github.com/dealii/dealii/blob/eb3ad2401ebf6560893f47536f50ea217865ae2e/include/deal.II/base/hdf5.h#L2111-L2116

masterleinad:
Just don't use `std::vector<bool>` (like ever): https://en.cppreference.com/w/cpp/container/vector_bool. You cannot assume that it stores its elements in a contiguous array.

dangars:
> Just don't use `std::vector<bool>` (like ever): https://en.cppreference.com/w/cpp/container/vector_bool. You cannot assume that it stores its elements in a contiguous array.

Thanks! very good point!

Now with this PR, only set_attribute and get_attribute can be used with bool datatypes.

The PR is ready for review

dangars:
I also updated the documentation.

masterleinad (approved):
Looks good to me.

PR #11877, opened 2021-03-10
Support subfaces for DoFHandler::prepare_coarsening_and_refinement().

marcfehling:
Follow-up to #11857.

We forgot to consider the case that a neighboring cell could have children and thus would not be active. This PR fixes this issue.

### DIFF
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -2741,6 +2741,35 @@ DoFHandler<dim, spacedim>::prepare_coarsening_and_refinement(
   // - always raise levels to match criterion, never lower them
   // - exchange level indices on ghost cells
 
+  // Function that updates the level of neighbor to fulfill difference
+  // criterion, and returns whether it was changed.
+  const auto update_neighbor_level =
+    [&future_levels, max_difference](const active_cell_iterator &neighbor,
+                                     const level_type cell_level) -> bool {
+    // We only care about locally owned neighbors. If neighbor is a ghost cell,
+    // its future FE index will be updated on the owning process and
+    // communicated at the next loop iteration.
+    if (neighbor->is_locally_owned())
+      {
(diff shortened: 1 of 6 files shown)


peterrum (approved):

peterrum:
Interesting:
```
/home/runner/work/dealii/dealii/build/include/deal.II/base/config.h:396:33: internal compiler error: Segmentation fault
 #define DEAL_II_NAMESPACE_CLOSE }
                                 ^
/home/runner/work/dealii/dealii/source/dofs/dof_handler.cc:3711:1: note: in expansion of macro ‘DEAL_II_NAMESPACE_CLOSE’
 DEAL_II_NAMESPACE_CLOSE
 ^
Please submit a full bug report,
with preprocessed source if appropriate.
```

/rebuild

marcfehling:
> Interesting:

Rebasing to restart the CI suite again didn't help. I'm kind of puzzled :-)

bangerth (approved):

bangerth:
That's certainly strange. Try to move the lambda into a function in an anonymous namespace that takes all captures as arguments instead. Maybe that will help the compiler.

bangerth:
Separately, here's something you might want to think about (but don't need to address right now): It's going to be relatively easy to extend this to the case of periodic boundaries. In that case, you have to not just see whether a face is at the boundary, but also whether there isn't a periodic neighbor behind it. If you search for `cell->periodic_neighbor()` in the library, you'll find plenty of examples of this kind of code.

But as I said, not for now. Just to plant the seed for future work :-)

marcfehling:
> That's certainly strange. Try to move the lambda into a function in an anonymous namespace that takes all captures as arguments instead. Maybe that will help the compiler.

With the alias, the static variable, and captures involved, this will grow really ugly. I could also just copy and paste it in the actual loop later, which is fine I think since the actual lambda function is short.

bangerth (approved):

PR #11878, opened 2021-03-10
Fix date of change-log entry

peterrum:


bangerth (commented):
It's still not an 8-digit date, though...

peterrum:
@bangerth How stupid of me. I think it is fixed now.

bangerth (approved):

PR #11880, opened 2021-03-10
VectorTools::get_position_vector(): add mapping as argument

peterrum:


include/deal.II/numerics/vector_tools_interpolate.h
```diff
@@ -284,6 +284,16 @@ namespace VectorTools
                       VectorType &                     vector,
                       const ComponentMask &            mask = ComponentMask());
 
+  /**
+   * Like the above function but also taking @p mapping as argument.
```
bangerth, line 288: Not related to your patch, but is the "then a default one is used" wrong in line 274 and supposed to say "then a default zero is used"?

bangerth (approved):

include/deal.II/numerics/vector_tools_interpolate.h
kronbichler, line 288: Can you slightly expand this comment, e.g., to say:
```suggestion
   * Like the above function but also taking @p mapping as argument. This might
   * introduce an approximation between the true geometry specified by the
   * manifold if the degree of the mapping is lower than the degree of the finite
   * finite element in the DoFHandler, but more importantly it allows to fill location
   * vectors for mappings that do not preserve vertex locations (like Eulerian
   * mappings).
```
(But give me a few minutes to check the internal workings of the original `get_position_vector` function again.)

peterrum: Thanks, I have added the comment.

kronbichler (approved):

include/deal.II/numerics/vector_tools_interpolate.templates.h
```diff
@@ -618,11 +618,30 @@ namespace VectorTools
...
   template <int dim, int spacedim, typename VectorType>
   void
   get_position_vector(const DoFHandler<dim, spacedim> &dh,
                       VectorType &                     vector,
                       const ComponentMask &            mask)
+  {
+    const FiniteElement<dim, spacedim> &fe = dh.get_fe();
+    get_position_vector(
+      *fe.reference_cell().template get_default_mapping<dim, spacedim>(
+        fe.degree),
```
peterrum, line 631: > Not related to your patch, but is the "then a default one is used" wrong in line 274 and supposed to say "then a default zero is used"?

@bangerth These are the lines the comment is referring to.

bangerth: Oh, I get it now. "A default one" isn't meant to reads as "A default 1" but as "A default component mask". I'll fix that separately.

PR #11882, opened 2021-03-11
Add further new MappingQCache::initialize() functions

peterrum:
With the interfaces:

https://github.com/dealii/dealii/blob/30692239388add11e7544ee056a92920429c8ce5/include/deal.II/fe/mapping_q_cache.h#L167-L172

https://github.com/dealii/dealii/blob/30692239388add11e7544ee056a92920429c8ce5/include/deal.II/fe/mapping_q_cache.h#L177-L182

Follow-up PR to #11867.


source/fe/mapping_q_cache.cc
```diff
@@ -125,6 +184,319 @@ MappingQCache<dim, spacedim>::initialize(
...
+  import(const LinearAlgebra::EpetraWrappers::Vector &vector,
+         LinearAlgebra::distributed::Vector<
+           typename LinearAlgebra::EpetraWrappers::Vector::value_type>
+           &vector_ghosted)
+  {
+    Assert(false, ExcNotImplemented());
+    (void)vector;
+    (void)vector_ghosted;
+  }
+} // namespace
```
peterrum, line 315: An issue is how to get the input vectors (of arbitrary types) into the right shape. I used here the suggestion made in https://github.com/dealii/dealii/pull/11594#issuecomment-775774068: copy the values into `LA::d::Vector` with the right partitioner. I don't know if we want to discuss this issue here or we want to move it to #11594 since I would like to use the same approach there. I could instantiate the function for now only for `LA::d::Vector`. 

But in general, is there already an infrastructure to convert vectors?

source/fe/mapping_q_cache.cc
```diff
@@ -125,6 +184,319 @@ MappingQCache<dim, spacedim>::initialize(
...
+      else
+        {
+          result.resize(
+            Utilities::pow<unsigned int>(this->get_degree() + 1, dim));
+        }
+
+      // Step 2c) read global vector and adjust points accordingly
+      if ((is_fe_q || is_fe_dgq) && fe.degree == this->get_degree())
+        {
+          AssertDimension(dim, spacedim); // TODO: codim-1
```
peterrum, line 438: Can someone give me a hint how to approach the problem of codim meshes here (`dim!=spacedim` - my application of this function). At this stage and later on in the general case where we will call `FEValues::get_function_values()`, we will have points of size dim and of size spacedim. Obviously, we need to multiply it with some Jacobian, but where do I get this one? Any hint (maybe in the form of a reference to other classes) are welcome!

source/fe/mapping_q_cache.cc
```diff
@@ -125,6 +184,319 @@ MappingQCache<dim, spacedim>::initialize(
...
+                          [id.first] += vector_ghosted(dof_indices[i]);
+                  else
+                    result[lexicographic_to_hierarchic_numbering[id.second]]
+                          [id.first] = vector_ghosted(dof_indices[i]);
+                }
+            }
+        }
+      else
+        {
+          Assert(false, ExcNotImplemented()); // TODO: general case
```
peterrum, line 474: This will be the general case once I have a plan how to tackle codim problems.

source/fe/mapping_q_cache.cc
```diff
@@ -125,6 +184,319 @@ MappingQCache<dim, spacedim>::initialize(
...
+
+
+template <int dim, int spacedim>
+template <typename VectorType>
+void
+MappingQCache<dim, spacedim>::initialize(
+  const Mapping<dim, spacedim> &   mapping,
+  const DoFHandler<dim, spacedim> &dof_handler,
+  const MGLevelObject<VectorType> &vector,
+  const bool                       vector_describes_relative_displacement)
```
peterrum, line 490: Since this is more or less the same as the above function but with multiple vectors being accessed, I tackle this once the above function is ready.

source/fe/mapping_q_cache.cc
peterrum: Actually, I think this should be not a problem since the `FESystem` is set up with dim components.

source/fe/mapping_q_cache.cc
```diff
@@ -125,6 +184,340 @@ MappingQCache<dim, spacedim>::initialize(
...
+            Vector<typename VectorType::value_type>(spacedim));
+
+          fe_values->get_function_values(vector_ghosted, values);
+
+          for (unsigned int q = 0; q < fe_values->n_quadrature_points; ++q)
+            for (unsigned int c = 0; c < spacedim; ++c)
+              if (vector_describes_relative_displacement)
+                result[q][c] += values[q][c];
+              else
+                result[q][c] = values[q][c];
```
peterrum, line 495: @kronbichler Regarding the discussion in https://github.com/dealii/dealii/pull/11867#discussion_r592198417, this would be the draft of the fall-back code that needs `FEValues`. Do you have a better idea here?

peterrum:
This PR is ready from my side. I will rebase as soon as PR #11867 is merged.

peterrum:
There is one issue I have at the moment:

https://github.com/dealii/dealii/blob/9b819f473bcc49991e505c53f400967a25503160/include/deal.II/fe/mapping_q_cache.h#L173-L178

The `VectorType` is a template argument. The annoying thing is that the compiler gets confused if I want to pass to `intialize()` an anonymous function (not a `std::function`!) and doe not know which function to call the right one, which takes a `std::function`, or the templated function. This is odd since both functions take in the second place different types of arguments (`Triangulation` vs. `DoFHandler`). Any ideas what one can do in such a situation?

Is there a way to check if `VectorType` is actually a vector type via traits?

kronbichler:
>  The annoying thing is that the compiler gets confused if I want to pass to `initialize()` an anonymous function (not a std::function!) and doe not know which function to call the right one, which takes a `std::function`, or the templated function. This is odd since both functions take in the second place different types of arguments (`Triangulation` vs. `DoFHandler`). Any ideas what one can do in such a situation?

I guess the problem is that `DoFHandler` can be constructed from `Triangulation`. I wonder why we don't mark the constructor explicit, https://github.com/dealii/dealii/blob/1a8188467c2ea1f32b4d02029bc0849d215583f2/include/deal.II/dofs/dof_handler.h#L549-L552 because I can't imagine something good coming out from a function that expects a `DoFHandler` but gets a triangulation. But before I go on speculating, would that help or would there be side effects?

peterrum:
> But before I go on speculating, would that help or would there be side effects?

I am pretty sure that that would help. I'll try it out in couple of minutes. If it works, I wills submit a separate PR. I don't think there will be any objections: @tjhei has made so many functions explicit in the recent months!

peterrum:
@kronbichler Your suggestion worked! Thanks! I have opened PR #11893 for this purpose.

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -161,6 +167,38 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
              const Function<spacedim> &          transformation_function,
              const bool function_describes_relative_displacement);
 
+  /**
+   * Initialize the data cache of the active cells by a solution (specified by
```
kronbichler, line 171:
```suggestion
   * Initialize the data cache of the active cells by a discrete field (specified by
```

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -161,6 +167,38 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
+   * @note By using this function for reinitialization, this class behaves like
+   *   MappingFEField (vector_describes_relative_displacement == false) or
+   *   MappingQEulerian (vector_describes_relative_displacement == true).
+   */
+  template <typename VectorType>
+  void
+  initialize(const Mapping<dim, spacedim> &   mapping,
+             const DoFHandler<dim, spacedim> &dof_handler,
+             const VectorType &               vector,
+             const bool vector_describes_relative_displacement);
```
kronbichler, line 184: Do we want to add an optional `ComponentMask` argument to this function, in analogy to `MappingFEField`? This would allow to extract data from a `DoFHandler` that contains additional solution components.

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -161,6 +167,38 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
+
+  /**
+   * Initialize the data cache of all non-artificial cells by a solution
+   * (specified by @p dof_handler and a set of @p vectors) that describes the
```
kronbichler, line 188:
```suggestion
   * (specified by @p dof_handler and a set of @p vectors on all levels of the triangulation) that describes the
```

include/deal.II/fe/mapping_q_cache.h
```diff
@@ -161,6 +167,38 @@ class MappingQCache : public MappingQGeneric<dim, spacedim>
...
+   * absolute or relative position of each support point.
+   *
+   * @note By using this function for reinitialization, this class behaves like
+   *   MappingFEField (vector_describes_relative_displacement == false) or
+   *   MappingQEulerian (vector_describes_relative_displacement == true).
```
kronbichler, line 193:
```suggestion
   *   MappingQEulerian (vector_describes_relative_displacement == true), but with much more
   *   efficient operations internally.
```

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,454 @@ MappingQCache<dim, spacedim>::initialize(
...
+MappingQCache<dim, spacedim>::initialize(
+  const Mapping<dim, spacedim> &   mapping,
+  const DoFHandler<dim, spacedim> &dof_handler,
+  const VectorType &               vector,
+  const bool                       vector_describes_relative_displacement)
+{
+  AssertDimension(dof_handler.get_fe_collection().size(), 1);
+  const FiniteElement<dim, spacedim> &fe = dof_handler.get_fe();
+  AssertDimension(fe.n_base_elements(), 1);
+  AssertDimension(fe.element_multiplicity(0), spacedim);
```
kronbichler, line 326: If we allow a component mask, we can relax this assertion.

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,454 @@ MappingQCache<dim, spacedim>::initialize(
...
+            fe_values->reinit(cell_tria);
+        }
+
+      std::vector<Point<spacedim>> result;
+
+      // Step 2b) read of quadrature points in the relative displacement case
+      // note: we also take this path for non-active or artifical cells so that
+      // these cells are filled with some useful data
+      if (vector_describes_relative_displacement ||
+          (is_active_non_artificial_cell == false))
```
kronbichler, line 421:
```suggestion
          is_active_non_artificial_cell == false)
```

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,454 @@ MappingQCache<dim, spacedim>::initialize(
...
+                          [id.first] = vector_ghosted(dof_indices[i]);
+                }
+            }
+        }
+      else
+        {
+          // case 2: general case; interpolation is needed
+          // note: the following code could be optimized for tensor-product
+          // elements via application of sum factorization as is done on
+          // MatrixFree/FEEvaluation
```
kronbichler, line 482: Or we could fix `FEValues` to use sum factorization :wink: (but not for this PR).

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,454 @@ MappingQCache<dim, spacedim>::initialize(
...
+          for (unsigned int q = 0; q < fe_values->n_quadrature_points; ++q)
+            for (unsigned int c = 0; c < spacedim; ++c)
+              if (vector_describes_relative_displacement)
+                result[q][c] += values[q][c];
+              else
+                result[q][c] = values[q][c];
+        }
+
+      return result;
+    });
```
kronbichler, line 500: One think I wonder for this function: Now that we have interpolated the values from all active cells, we have a mismatch in the geometry between the active cells and the level cells. We would then typically go on and call the function with the `MGLevelObject`. I wonder if we should try to fix things here: Since we know the geometry by the position of the fine cells from the given vector, can't we use #11804 to perform an arbitrary interpolation to get the positions in the level vectors? My idea is that the levels will be only around 1/7 of the cells on the active level in 3D or around 1/3 in 2D, so maybe one could take the risk to choose a brute-force evaluation. This would allow us to simply skip the other function, and free the user from having to set up the MG data structures, which might be unexpected if you don't know about it.

And if we do a proper job with optimizing #11804, we would not even notice that we do some very heavy lifting here.

kronbichler (commented):
I have a few more comments. Please check especially the last one where I reason about the usefulness of the multigrid variant.

source/fe/mapping_q_cache.cc
peterrum: Implementing this wouldn't be hard (and the code would be definitely easier). But I have concerns regarding:

https://github.com/dealii/dealii/blob/f03ed9ec8651af89c4106058391707ad2d0769a0/source/base/mpi_remote_point_evaluation.cc#L90-L91

and all the `GridTools::find_all_active_cells_around_point()` calls. 

I guess the best approach is to split up this PR so that we can merge at least one function while waiting for #11804.

kronbichler: Yes, I had been thinking about this was well. There is also another subtle challenge in this function with the multigrid context that I tacitly ignored before: What if sufficiently many MPI ranks request the value on the same point (or points owned by a single MPI rank)? I think we should aim for eventually solving these issues, but it might be something too challenging for the short term, in which case we might go with the proposed variant for the 9.3 release. So I agree on the general plan.

peterrum: > What if sufficiently many MPI ranks request the value on the same point (or points owned by a single MPI rank)?

What kind of local refinement do you have perform to trigger such a behavior?

include/deal.II/fe/mapping_q_cache.h
peterrum: Let's do this in a follow-up PR. It does not look too complicated (component need to be linearized and we need some additional checks and a new test).

source/fe/mapping_q_cache.cc
kronbichler: A uniform mesh suffices - think of a cube where all processes want the mapping information for level 0. Whoever happens to own the unknown in the lower left corner needs to send this information to all ranks. I don't think we have a mechanism (yet) to mediate such requests to the requester via multiple ranks like in a MPI broadcast operations.

peterrum: > think of a cube where all processes want the mapping information for level 0.

I am a bit confused. In out use case it is the opposite around. Coarser levels want information from finer level (actually the finest level); and those should be partitioned well.

kronbichler: As far as I remember we set up the coarse level cells as ghosts on all MPI processes, also in case those processes do not own any of the cells on that level and even if none of cells on the first level are own on that process. This implies every single rank requests the positions of the single cell on level 0. Even though the points on that single cell will be well-distributed, whoever process owns that cell will have the burden to send the position to every single rank (unless there is some smartness in #11804 that I don't recall) That would be an overwhelming number with 100k MPI ranks and I see lots of eager messages here. So I see the problem on the other end of the communication that you are describing.

peterrum: > As far as I remember we set up the coarse level cells as ghosts on all MPI processes, also in case those processes do not own any of the cells on that level and even if none of cells on the subsequent level are owned on that process.

If that is the case, yes (I should remember what the distributed triangulations are doing).

But as far as I can see it, we don't need to update the mapping of those cells, do we? Essentially they are "artificial"; but we do not have the notion of a different kind of artificial cells. `MappingQCache::initialize()` could make the distinction and query the position of points only for cells that are actually relevant. But this is all for the future...

*I will try to fix the compilation error on the Windows system in a minute.*

kronbichler: > But as far as I can see it, we don't need to update the mapping of those cells, do we? Essentially they are "artificial"; but we do not have the notion of a different kind of artificial cells.

I wonder if we can't simply make them `artificial` for the case where none of the immediate descendants are locally owned? That should avoid this problem in the natural way. Let me try to have a look.

kronbichler: > As far as I remember we set up the coarse level cells as ghosts on all MPI processes, also in case those processes do not own any of the cells on that level and even if none of cells on the subsequent level are owned on that process.

It turned out I was wrong - we only ghost those cells where we have an immediate child (as I guessed would be the right thing to do) at least for the parallel distributed triangulation, see https://github.com/dealii/dealii/blob/57e91c245854cc49033bf3e4e13ec8759cffb15c/source/distributed/tria.cc#L2523-L2541
So this seems less critical than I thought for a possible new function with this scheme (we would only want to fill owned or ghosted cells, not the artificial ones).

kronbichler:
>  But I have concerns regarding: https://github.com/dealii/dealii/blob/f03ed9ec8651af89c4106058391707ad2d0769a0/source/base/mpi_remote_point_evaluation.cc#L90-L91 and all the `GridTools::find_all_active_cells_around_point()` calls.

Giving this another thought, I think we need an `all_reduce` kind of function rather than an `all_gather` kind for the particular data structure of r-trees and bounding boxes. If we insist on `all_gather`, this will hit any use case with the function of #11804 on a sufficiently large scale, which is unfortunate given the expressiveness of this function. So it was good after all to think about these new cases to edge out these possible performance limitations.

kronbichler (approved):
I guess we can postpone the open topics to other PRs or issues. One thing I wonder is the size of the `mapping_q_cache_06` test: Do we expect any loss of information if we lower the mapping degree to 2? 2x 8k lines of output seems a bit much for a test, and I would only do that if necessary.

kronbichler:
Also, I would apprectiate if someone else can give this a look as well, since I've been involved in the design and initial drafts.

peterrum:
> One thing I wonder is the size of the mapping_q_cache_06 test: Do we expect any loss of information if we lower the mapping degree to 2? 2x 8k lines of output seems a bit much for a test, and I would only do that if necessary.

The problem is not related to the mapping degree but with the vtk-output. I have switched to high-order output which seem to consume significantly less locs. Regarding the test I think it is already close to minimal: I need at least two multigrid level and on the coarser one, I need artificial cells.

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,465 @@ MappingQCache<dim, spacedim>::initialize(
...
+      vector_ghosted[i] = vector[i];
+  }
+
+  template <typename VectorType,
+            typename std::enable_if<
+              std::is_same<VectorType,
+                           LinearAlgebra::distributed::Vector<
+                             typename VectorType::value_type>>::value,
+              VectorType>::type * = nullptr>
+  void
```
drwells, line 299: I'd prefer to do something like

```cpp
template <typename T>
void
copy_locally_owned_data_from(
  const LA::d::V<T> &vector,
  LA::d::V<T> &v_g)
```
here and

```cpp
void
copy_locally_owned_data_from(
  const VectorType &vector,
  LA::d::V<typename VectorType::value_type> &v_g)
```
above over using SFINAE. I am pretty sure this would work - here's a silly example of this type of deduction:
```cpp
#include <iostream>
#include <list>
#include <vector>

template <typename T>
void
foo(const std::vector<T> &v,
    const std::vector<T> &u)
{
    std::cout << "foo" << std::endl;
}

template <typename Container>
void
foo(const Container &u,
    const std::vector<typename Container::value_type> &v)
{
    std::cout << "foo2" << std::endl;
}

int main()
{
    std::vector<int> u;
    std::vector<int> v;
    std::list<int> w;

    foo(u, v);
    foo(w, v);
}
// output is 
// foo
// foo2
```

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,465 @@ MappingQCache<dim, spacedim>::initialize(
...
+    const LinearAlgebra::EpetraWrappers::Vector &vector,
+    LinearAlgebra::distributed::Vector<
+      typename LinearAlgebra::EpetraWrappers::Vector::value_type>
+      &vector_ghosted)
+  {
+    Assert(false, ExcNotImplemented());
+    (void)vector;
+    (void)vector_ghosted;
+  }
+#endif
```
drwells, line 320: Don't we have generic vector access functions? Why doesn't this work with all the standard vector classes?

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,465 @@ MappingQCache<dim, spacedim>::initialize(
...
+
+      // Step 2a) set up and reinit FEValues (if needed)
+      if (((vector_describes_relative_displacement ||
+            (is_active_non_artificial_cell == false)) &&
+           ((mapping_q_generic != nullptr &&
+             this->get_degree() == mapping_q_generic->get_degree()) ==
+            false)) /*condition 1: points need to be computed via FEValues*/
+          || ((is_active_non_artificial_cell == true) &&
+              (((is_fe_q || is_fe_dgq) && fe.degree == this->get_degree()) ==
+               false)) /*condition 2: interpolation of values is needed*/)
```
drwells, line 394: Can't you use `update_values_might_be_needed` here?

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,465 @@ MappingQCache<dim, spacedim>::initialize(
...
+                        dof_handler.get_communicator());
+  copy_locally_owned_data_from(vector, vector_ghosted);
+  vector_ghosted.update_ghost_values();
+
+  // FE and FEValues in the case they are needed
+  FE_Nothing<dim, spacedim> fe_nothing;
+  Threads::ThreadLocalStorage<std::unique_ptr<FEValues<dim, spacedim>>>
+             fe_values_all;
+  const bool update_values_might_be_needed =
+    ((is_fe_q || is_fe_dgq) && fe.degree == this->get_degree()) == false;
```
drwells, line 366: This variable name could be improved - don't we always need `update_values` if we cannot just read off locations from the solution vectors?

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,465 @@ MappingQCache<dim, spacedim>::initialize(
...
+            return result;
+        }
+      else
+        {
+          result.resize(
+            Utilities::pow<unsigned int>(this->get_degree() + 1, dim));
+        }
+
+      // Step 2c) read global vector and adjust points accordingly
+      if ((is_fe_q || is_fe_dgq) && fe.degree == this->get_degree())
```
drwells, line 454: same (we can use `update_values_might_be_needed`)

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,465 @@ MappingQCache<dim, spacedim>::initialize(
...
+
+      // Step 2a) set up and reinit FEValues (if needed)
+      if (((vector_describes_relative_displacement ||
+            (is_non_artificial_cell == false)) &&
+           ((mapping_q_generic != nullptr &&
+             this->get_degree() == mapping_q_generic->get_degree()) ==
+            false)) /*condition 1: points need to be computed via FEValues*/
+          || ((is_non_artificial_cell == true) &&
+              (((is_fe_q || is_fe_dgq) && fe.degree == this->get_degree()) ==
+               false)) /*condition 2: interpolation of values is needed*/)
```
drwells, line 594: same

source/fe/mapping_q_cache.cc
```diff
@@ -249,17 +259,465 @@ MappingQCache<dim, spacedim>::initialize(
...
+            return result;
+        }
+      else
+        {
+          result.resize(
+            Utilities::pow<unsigned int>(this->get_degree() + 1, dim));
+        }
+
+      // Step 2c) read global vector and adjust points accordingly
+      if ((is_fe_q || is_fe_dgq) && fe.degree == this->get_degree())
```
drwells, line 654: same

drwells (changes requested):
Looks good - I have a few minor suggestions.

I believe this is most of what is required to generically implement #10163 (except we need a way to read that data in).

source/fe/mapping_q_cache.cc
peterrum: I would also prefer this approach. But Windows does't ;) 

My first version looked exactly like your suggestion. But Windows complained that it couldn't decide which one to use for `LA:d:V`...

source/fe/mapping_q_cache.cc
peterrum: I gave another shot at `LinearAlgebra::ReadWriteVector` and have added two additional methods so that it works for all vectors (see #11938). The code would now look like this:
https://github.com/dealii/dealii/blob/3d64721983a0059b9415d658d0228fab5391fe68/source/fe/mapping_q_cache.cc#L277-L288

source/fe/mapping_q_cache.cc
peterrum: I have renamed the variable to `interpolation_of_values_is_needed` and have added a comment. I now also use this variable at more places (this variable was only introduced later^^).

peterrum:
Now, this PR depends on #11938. Adding the `Do not merge` label.

source/fe/mapping_q_cache.cc
drwells: That's unfortunate - such is life with MSVC.

drwells (approved):

peterrum:
With #11938 being merged, this PR is now ready also from my side!

PR #11883, opened 2021-03-11
Improve an error message.

bangerth:
/rebuild

### DIFF
--- a/source/lac/trilinos_sparse_matrix.cc
+++ b/source/lac/trilinos_sparse_matrix.cc
@@ -2215,7 +2215,7 @@ namespace TrilinosWrappers
                                         const_cast<Epetra_CrsMatrix &>(
                                           tmp_result.trilinos_matrix()));
 #  else
-      Assert("false",
+      Assert(false,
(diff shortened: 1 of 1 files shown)


peterrum (approved):

source/lac/trilinos_sparse_matrix.cc
```diff
@@ -2215,7 +2215,12 @@ namespace TrilinosWrappers
                                         const_cast<Epetra_CrsMatrix &>(
                                           tmp_result.trilinos_matrix()));
 #  else
-      Assert("false", ExcMessage("This function requires EpetraExt."));
+      Assert("false",
```
masterleinad, line 2218:
```suggestion
      Assert("false",
```
Why is this a string?

bangerth: History. Interestingly, this makes the condition `true` :-)

bangerth:
I fixed the condition as well -- good eyes, @masterleinad !

masterleinad (approved):
LGTM.

PR #11888, opened 2021-03-11
Refactored order of data transfer in p:d::Tria.

marcfehling:
This PR wraps the data transfer around the `copy_local_forest_to_triangulation()` function.

This way, we could still change the `future_fe_indices` (using a signal) or other characteristics after modifications to the `p4est` objects have been made, without interfering with the data transfer process.

This signal could be placed between `update_cell_relations` and the repartition section. The updated `future_fe_indices` may determine the weight associated with each cell for load balancing. We will not break the data transfer as it will happen after the repartition section with this PR.

### DIFF
--- a/source/distributed/tria.cc
+++ b/source/distributed/tria.cc
@@ -2690,22 +2690,14 @@ namespace parallel
       // has happened, we need to update the quadrant cell relations
       update_cell_relations();
 
-      // before repartitioning the mesh, store the current distribution
-      // of the p4est quadrants and let others attach mesh related info
-      // (such as SolutionTransfer data)
+      // before repartitioning the mesh, save a copy of the current positions of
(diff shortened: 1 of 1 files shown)


source/distributed/tria_base.cc
```diff
@@ -893,6 +893,7 @@ namespace parallel
           const auto &cell_status = std::get<0>(*cell_rel_it);
           const auto &dealii_cell = std::get<1>(*cell_rel_it);
 
+#ifdef DEBUG
```
bangerth, line 896: There is nothing left from the following `switch` statement if you remove the assertions, and so I'd rather not add these `#ifdef DEBUG` markers. The compiler will be able to remove the whole thing.

bangerth (approved):

PR #11898, opened 2021-03-13
Add get_laplacians to MeshWorker::ScratchData

jppelteret:


jppelteret:
I recall a the time when I wrote this patch that I couldn't find a test to add a check for this function too. I'll take another look to see if there are any tests for ScratchData that test the solution value/gradient/... extraction.

bangerth (approved):

kronbichler (approved):

PR #11900, opened 2021-03-13
SD BatchOptimizer: Implement cached evaluation

jppelteret:


jppelteret:
The tests in `tests/symengine` will provide full coverage for this change, as we use this new method internally. I'll double check that all tests pass, since I don't think that this feature is enabled on any of the testers.

jppelteret:
All related tests pass on this branch:
```
$ ctest -R symengine -j4
...
100% tests passed, 0 tests failed out of 164
```

include/deal.II/differentiation/sd/symengine_optimizer.h
```diff
@@ -1124,6 +1124,9 @@ namespace Differentiation
        * @tparam TensorType The type of tensor to be evaluated and returned
        *         (i.e. Tensor or SymmetricTensor).
        * @param[in] symbol_tensor The symbolic tensor that is to be evaluated.
+       * @param[in] cached_evaluation The vector that stores the numerical
```
bangerth, line 1127: Here and below, since it's an input vector, maybe saying
```suggestion
       * @param[in] cached_evaluation A vector that stores the numerical
```
would be more appropriate because it lives in user space (right?). But I wouldn't know what I should be putting into this argument. Could you explain?

include/deal.II/differentiation/sd/symengine_optimizer.h
```diff
@@ -1921,6 +1933,48 @@ namespace Differentiation
       SymmetricTensor<rank, dim, ReturnType>
       evaluate(const SymmetricTensor<rank, dim, Expression> &funcs) const;
 
+
+      /**
+       * Returns the result of a value substitution into the optimized
+       * counterpart of @p func. This function fetches that one value
+       * stored in the @p cached_evaluation vector.
```
bangerth, line 1940: I have to admit that I don't understand what one should pass as that second argument.

jppelteret: Ok, I'll cross reference the `evaluate()` function that returns the `const std::vector<ReturnType> &`. A copy of that vector is really the only thing that you'd ever want as an input to this function.

jppelteret: I've given a little more detail to the documentation here, as well as at the point to which you indicated in your previous comment.

bangerth (approved):

bangerth:
Thanks!

PR #11901, opened 2021-03-13
Fix TriangulationBase::reset_global_cell_indices()

peterrum:
Fixes the failing tests:
```
	4237 - mpi/prepare_coarsening_and_refinement_01.mpirun=4.debug (Failed)
	4239 - mpi/prepare_coarsening_and_refinement_02.mpirun=4.debug (Failed)
```

The following test is still failing:
```
	4241 - mpi/prepare_coarsening_and_refinement_03.mpirun=4.debug (Failed)
```
with:
```
4210: An error occurred in line <90> of file </home/munch/sw-index/dealii2/source/distributed/shared_tria.cc> in function
4210:     void dealii::parallel::shared::Triangulation<dim, spacedim>::partition() [with int dim = 2; int spacedim = 2]
4210: The violated condition was: 
4210:     max_active_cells == this->n_active_cells()
4210: Additional information: 
4210:     A parallel::shared::Triangulation needs to be refined in the same way
4210:     on all processors, but the participating processors don't agree on the
4210:     number of active cells.

```

FYI @marcfehling

### DIFF
--- a/source/distributed/tria_base.cc
+++ b/source/distributed/tria_base.cc
@@ -400,14 +400,17 @@ namespace parallel
           for (unsigned int i = 0; i < n_subdomains; ++i)
             cell_counter[i + 1] += cell_counter[i];
 
+          AssertDimension(cell_counter.back(), this->n_active_cells());
+
           // create partitioners
-          IndexSet is_local(cell_counter.back());
+          IndexSet is_local(this->n_active_cells());
           is_local.add_range(cell_counter[my_subdomain],
                              cell_counter[my_subdomain + 1]);
-          IndexSet is_ghost(cell_counter.back());
           number_cache.active_cell_index_partitioner =
             std::make_shared<const Utilities::MPI::Partitioner>(
-              is_local, is_ghost, this->mpi_communicator);
+              is_local,
+              complete_index_set(this->n_active_cells()),
+              this->mpi_communicator);
 
           // set global active cell indices and increment process-local counters
           for (const auto &cell : this->active_cell_iterators())
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -2711,11 +2711,8 @@ DoFHandler<dim, spacedim>::prepare_coarsening_and_refinement(
         dynamic_cast<const parallel::TriangulationBase<dim, spacedim> *>(
           &(*tria)))
     {
-      const auto &partitioner =
-        *parallel_tria->global_active_cell_index_partitioner().lock();
-      future_levels.reinit(partitioner.locally_owned_range(),
-                           partitioner.ghost_indices(),
-                           partitioner.get_mpi_communicator());
+      future_levels.reinit(
+        parallel_tria->global_active_cell_index_partitioner().lock());
     }
   else
     {


source/distributed/tria_base.cc
```diff
@@ -405,6 +405,7 @@ namespace parallel
           is_local.add_range(cell_counter[my_subdomain],
                              cell_counter[my_subdomain + 1]);
           IndexSet is_ghost(cell_counter.back());
+          is_ghost.add_range(0, cell_counter.back());
```
kronbichler, line 408: While not wrong for the parallel shared triangulation, I would like to add that the data structures of the partitioner and the vectors built on top of them are not at all optimized for this case. The ghost exchange is some form of `MPI_Allgather` (which by itself is not particularly scalable beyond 1k ranks), but doing it by point-to-point communication to *each* MPI rank is going to be crazy slow beyond 100 ranks already.

peterrum: The alternative is to introduce the notion of a halo layer of a single cell also in the case of `p:s:T` without artificial cells which I find a bit counter-intuitive. If someone wants to have a halo layer ( wants to scale beyond a couple of hundred of ranks even with `p:d:T`), he or she would activate artificial cells.

marcfehling:
> The following test is still failing:
> 
> ```
> 	4241 - mpi/prepare_coarsening_and_refinement_03.mpirun=4.debug (Failed)
> ```

I think that test fails because of a mistake I made: For `p::s::T`, refinement flags need to be set on all cells, and I only set them on locally owned cells. I investigate.

source/distributed/tria_base.cc
peterrum: @kronbichler How should we proceed here? In my opinion, the partioner should reflect that nature of the triangulation. In this case it means that all cells except of the locally-owned ones are ghost.

kronbichler: Yes, this is not wrong with the present patch, but rather with the specific triangulation. Having to send information to more than ~100 MPI ranks hurts and would be better done with other communication routines. But I didn't want to imply we should add alternative communication routines, we should be aware that our data structures break down for the shared triangulation (just in case someone does not realize that from the beginning of duplicating the whole grid).

kronbichler (approved):

source/distributed/tria_base.cc
bangerth: These two lines are equivalent to the following, which I think would be easier to read:
```
  const IndexSet is_ghost = complete_index_set (this->n_active_cells());
```

bangerth: Maybe also use `this->n_active_cells()` in the initialization of `is_local` above, and assert that `this->n_active_cells() == cell_counter.back()`.

bangerth: As for the questions you all had above: I think it is fair to say that if there are no active cells, then *all* cells not locally owned are ghosts. That seems like the intended semantics, whether or not that is efficient. Maybe one could make a statement somewhere where the user selects for/against artificial cells that not allowing artificial cells is inefficient.

I am confused, however, why *all* cells are in the `is_ghost` partitioner. I guess the `MPI::Partitioner` removes the overlap with the `is_local` index set? If so, please rename `is_local_or_ghost`.

peterrum: I have made the changes. I will open a separate PR regarding: 

> Maybe one could make a statement somewhere where the user selects for/against artificial cells that not allowing artificial cells is inefficient.

in a moment.

kronbichler (approved):

PR #11905, opened 2021-03-14
Fix refinement flags in mpi/prepare_coarsening_and_refinement_03().

marcfehling:
Spotted in #11901.

While it is sufficient to only set refinement flags on locally owned cells for distributed Triangulation classes, it is mandatory to do that on all cells for shared Triangulations.

This PR together with #11901 will fix the current issues with `tests/mpi/prepare_coarsening_and_refinement_*`.

### DIFF
--- a/tests/mpi/prepare_coarsening_and_refinement_03.cc
+++ b/tests/mpi/prepare_coarsening_and_refinement_03.cc
@@ -80,10 +80,9 @@ test(parallel::TriangulationBase<dim> &tria,
   for (unsigned int i = 0; i < n_refinements; ++i)
     {
       for (const auto &cell : tria.active_cell_iterators())
-        if (cell->is_locally_owned())
-          for (unsigned int v = 0; v < cell->n_vertices(); ++v)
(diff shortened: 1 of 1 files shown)


bangerth (approved):
The patch is fine as is, but it seems like a bug to me that one has to set refinement indicators on *all* cells on all processors. Is this documented anywhere?

kronbichler (approved):

marcfehling:
> The patch is fine as is, but it seems like a bug to me that one has to set refinement indicators on _all_ cells on all processors. Is this documented anywhere?

Since each shared Triangulation represents the whole mesh on each processor, these flags need to be set on all processors. It looks like we don't communicate them for shared Triangulations, so we need to set them everywhere (also on artificial cells). We do that implicitly in `step-18` by calling `GridRefinement::refine_and_coarsen_fixed_number` on a local copy of a global vector. 

For distributed Triangulations, the additional flags would simply be ignored.

Maybe an additional sentence in the general documentation of the shared Triangulation class will help.

Issue #11829, opened 2021-03-01
Make KINSOL interfaces work for newer SUNDIALS versions


bangerth:
I woke up today and wanted to write a SUNDIALS KINSOL tutorial program. So I spent a few hours on a couple of pull requests, but in the end found that the current KINSOL interfaces only work for SUNDIALS 3.x whereas SUNDIALS is already at version 5.7 right now. Problem is that the SUNDIALS interfaces have changed substantially, and to be quite honest I can't for the life of me not figure out how KINSOL is now supposed to be driven.

The code we have right now makes use of KINSOL's feature that we can attach a custom linear solver for the Jacobian matrix. That makes sense in the FEM context since we know how to solve linearized problems. Specifically, our interfaces have a function `t_kinsol_solve_jacobian` and corresponding "setup" function, and we then do this:
```
        auto KIN_mem        = static_cast<KINMem>(kinsol_mem);
        KIN_mem->kin_lsolve = t_kinsol_solve_jacobian<VectorType>;
        if (setup_jacobian)
          KIN_mem->kin_lsetup = t_kinsol_setup_jacobian<VectorType>;
```
But KINSOL 5.7 no longer exposes these function pointers on the lhs of the assignments -- in fact, the whole `KINMem` structure no longer exists. I believe that there are setter functions one needs to use now, but I can't figure out how this is supposed to work. The question is what these setter functions are and what their arguments relate to. In the `kinsol_ls.h` (ls=linear solve) file, the only exposed functions are `KINSetPreconditioner`, but I think that what that does is only set a preconditioner for one of the built-in matrices. More promising would be to call `KINSetLinearSolver` which allows setting an `SUNLinearSolver` object. The corresponding functions are then in `sundials/sundials_linear_solver.h`, but I'm having a very hard time navigating that file and the corresponding section in the manual: The fundamental difficulty I have is that the only documented ways to use these functions is to create a `SUNLinearSolver` object that corresponds to one of the pre-defined solvers (of which there are many, both direct and iterative), but not a custom solver. The examples that come with SUNDIALS also don't appear to cover that case...

Are there examples for building linear solver objects that use a custom solver?

@luca-heltai @sebproell Any insight?

sebproell:
> Problem is that the SUNDIALS interfaces have changed substantially, and to be quite honest I can't for the life of me not figure out how KINSOL is now supposed to be driven.

[...]

PR #11908, opened 2021-03-14
Update SUNDIALS KINSOL interfaces to their current master

bangerth:
Our KINSOL interfaces only worked for SUNDIALS 3.x. This patch makes them work with the current SUNDIALS 5.7.

There are a couple of issues worth pointing out:
* I implemented things so that the current KINSOL interface is used starting at KINSOL 5.0. But I don't actually know whether that's the version of KINSOL where it was introduced -- it could also have been in KINSOL 4.0 or 4.1. Does anyone have a version in the range 4.0, 4.1, or 5.0 sitting around and would be willing to see what happens with this patch?
* For the `solve` function, KINSOL used to provide the current iterate `x_k` and residual vector `F(x_k)`, and we passed that on to the user-space function that implements the solution of the Jacobian system `J dx = -F`. But that is no longer the case: KINSOL now only provides `-F`. The question is what to do about the member variable to which users attach:
```
    std::function<int(const VectorType &ycur,
                      const VectorType &fcur,
                      const VectorType &rhs,
                      VectorType &      dst)>
      solve_jacobian_system;
```
I now simply provide empty vectors for `ycur` and `fcur` and document that that is so. I think we should probably provide a separate `solve_jacobian_system` member that doesn't have these arguments. But that can wait for a follow-up patch.


bangerth:
The patch also provides 4 more tests.

bangerth:
Seems to work now. Anyone want to take a look at it? @sebproell ?

source/sundials/kinsol.cc
```diff
@@ -147,8 +146,88 @@ namespace SUNDIALS
...
+      // Allocate temporary (deal.II-type) vectors into which to copy the
+      // N_vectors
+      GrowingVectorMemory<VectorType>            mem;
+      typename VectorMemory<VectorType>::Pointer ycur(mem);
+      typename VectorMemory<VectorType>::Pointer fcur(mem);
+      solver.reinit_vector(*ycur);
+      solver.reinit_vector(*fcur);
+
+      copy(*ycur, u);
+      copy(*fcur, f);
```
sebproell, line 176: This whole copy business can be replaced with the deal.II NVectorView (#11395 ) (in a future PR).

source/sundials/kinsol.cc
```diff
@@ -270,12 +350,86 @@ namespace SUNDIALS
...
+
+        LS->ops->gettype =
+          [](SUNLinearSolver /*ignored*/) -> SUNLinearSolver_Type {
+          return SUNLINEARSOLVER_MATRIX_ITERATIVE;
+        };
+
+        LS->ops->free = [](SUNLinearSolver LS) -> int {
+          if (LS->content)
+            {
+              LS->content = NULL;
```
sebproell, line 379:
```suggestion
              LS->content = nullptr;
```

include/deal.II/sundials/kinsol.h
```diff
@@ -664,6 +681,14 @@ namespace SUNDIALS
      */
     void *kinsol_mem;
 
+#  if DEAL_II_SUNDIALS_VERSION_GTE(5, 0, 0)
+    /**
+     * An object that describes the linear solver, in case the user decides
+     * to attach one. Otherwise it remains unused.
+     */
+    SUNLinearSolver sun_linear_solver;
```
sebproell, line 689: Seems to be unused.

sebproell (commented):
LGTM, nice tests! Now that I see your implementation I am no longer disagreeing with this SUNDIALS abuse as much as in my various comments in #11829 ;)

sebproell:
> Does anyone have a version in the range 4.0, 4.1, or 5.0 sitting around and would be willing to see what happens with this patch?

I have a deal.II build with SUNDIALS 4.0.0 and will try it.

source/sundials/kinsol.cc
bangerth: Ah, I was unaware we had that. I read the paper that describes all of the new SUNDIALS stuff and had the impression that one can generate `N_Vector` objects that are just wrappers around everyone else's favorite vector type. But we don't use that here, and I thought we didn't have those implemented yet. I guess I have to read through #11395, unless you wanted to make the conversion yourself. In any case, like you say, this is a future problem.

include/deal.II/sundials/kinsol.h
bangerth: Good eyes -- that was a left-over from one of my previous iterations...

bangerth:
Thanks, @sebproell, for looking this over! I pushed an updated version. If you were willing to try with 4.0, that would be fantastic!

sebproell:
@bangerth It builds with SUNDIALS 4.0 but I run into a SUNDIALS internal segfault even for the ARKode tests... I will have a look but this is probably also failing for master.

sebproell:
KINSOL tests are running but not passing the tests. This is the output of kinsol_03.debug:

```

DEAL::Evaluating the solution at u=(0.500000,1.23400)
DEAL::Setting up Jacobian system at u=(0.500000,1.23400)
DEAL::Solving Jacobian system with rhs=(0.162480,-1.79816)
DEAL::Evaluating the solution at u=(0.581240,-0.645395)
DEAL::Solving Jacobian system with rhs=(-1.16042,0.349431)
DEAL::Evaluating the solution at u=(0.00102861,0.284248)
DEAL::Solving Jacobian system with rhs=(0.0383589,-0.289047)
DEAL::Evaluating the solution at u=(0.0202080,-0.0239792)
DEAL::Solving Jacobian system with rhs=(-0.0404090,0.00378553)
DEAL::Evaluating the solution at u=(3.55540e-06,1.08225e-05)
DEAL::Solving Jacobian system with rhs=(-7.11070e-06,-1.43779e-05)
DEAL::Evaluating the solution at u=(5.16813e-11,-5.16814e-11)
5.168e-11 -5.168e-11 
DEAL::Converged in 5 iterations.
```
The output is too short compared to the expected one as some of the evaluates are missing (same for other new tests). Probably the tests are tied to the SUNDIALS implementation too much? Newer versions seem to do one more evaluate every time...

As for the segfault in ARKode with SUNDIALS 4.0: I don't have much time right now (exam period...) to fix that. Is it decided which SUNDIALS versions should be supported in the future? Does it make sense to spend effort on anything under 5.0?

bangerth:
Thanks, @sebproell , for checking! I think it's ok if we don't get the same output with 4.0. As you probably noticed, I'm only running them with SUNDIALS 5.0 and later, so we won't see the 4.x results as a failure -- the test is simply not run. 

It would not greatly surprise me if we get different results with future SUNDIALS versions going forward. We'll have to tackle this then. I may be outputting too detailed information in these tests, but we won't see until we try with future versions. For the moment, what your checks show is that the KINSOL interfaces do work with SUNDIALS 4.0, and they also work with 5.7: You get convergence with your version, as do I. The fact that the test gives slightly different numbers of iterations is, in my mind, secondary.

As for ARKODE: I would be totally ok if we decided to only support SUNDIALS 5.0 and later, for example. (It is not helpful to see that ARKODE has its own version scheme, independent of SUNDIALS...) I can't seem to figure out when 5.0 was released, but we should spend whatever little time we all have on doing things that are useful in the future, and trying to support old versions is not a good use of time. I would much rather us just disable these interfaces with an error message if we find that ARKODE is too old.

For now, focus on your exams -- *that* is what is important!

bangerth:
I found it on https://github.com/LLNL/sundials/tags: SUNDIALS 5.0 was released in October 2019. That's not crazy old, but I stand by the suggestion to not spend too much time on supporting older versions but focus on newer stuff.

source/sundials/kinsol.cc
```diff
@@ -338,24 +492,24 @@ namespace SUNDIALS
...
+#    ifdef DEAL_II_WITH_TRILINOS
   template class KINSOL<TrilinosWrappers::MPI::Vector>;
   template class KINSOL<TrilinosWrappers::MPI::BlockVector>;
-#      endif
+#    endif
 
-#      ifdef DEAL_II_WITH_PETSC
-#        ifndef PETSC_USE_COMPLEX
+#    ifdef DEAL_II_WITH_PETSC
+#      ifndef PETSC_USE_COMPLEX
```
jppelteret, line 503: Can we perhaps instantiate the KINSOL class on all `VECTOR_TYPES` or `REAL_VECTOR_TYPES`?

jppelteret:
@bangerth I'll try to take a look at your PR this weekend :-)

source/sundials/kinsol.cc
bangerth: I thought about that but in the end decided against it. We'd need to write the `copy()` functions for all vector types, and I don't really know how to address parallel computations. Let me defer this to a later time if you don't ind!

sebproell: Or just use `NVectorView` instead of `copy` which works for all vector types;) (although it might not have all operations required by KINSOL?)

bangerth: Yes, that's the long-term goal.

tests/sundials/kinsol_03.cc
```diff
@@ -0,0 +1,140 @@
...
+#include <deal.II/lac/vector.h>
+
+#include <deal.II/sundials/kinsol.h>
+
+#include "../tests.h"
+
+// Solve a nonlinear system but provide only residual function. KINSOL
+// then uses its internal solvers which are based on a
+// finite-difference approximation to the Jacobian and a direct
+// solver.
```
jppelteret, line 28: This description seems to be in contradiction with the following paragraph.

tests/sundials/kinsol_04.cc
```diff
@@ -0,0 +1,147 @@
...
+#include <deal.II/lac/vector.h>
+
+#include <deal.II/sundials/kinsol.h>
+
+#include "../tests.h"
+
+// Solve a nonlinear system but provide only residual function. KINSOL
+// then uses its internal solvers which are based on a
+// finite-difference approximation to the Jacobian and a direct
+// solver.
```
jppelteret, line 28: Again, general description doesn't agree with the following paragraph.

tests/sundials/kinsol_04.cc
```diff
@@ -0,0 +1,147 @@
...
+    F(1) = std::sin(u[0] - u[1]) + 2 * u[1] - u[1];
+    return 0;
+  };
+
+  FullMatrix<double> J_inverse(2, 2);
+
+  kinsol.setup_jacobian = [&J_inverse](const VectorType &u,
+                                       const VectorType &F) -> int {
+    // We don't do any kind of set-up in this program, but we can at least
+    // say that we're here
```
jppelteret, line 112: This comment seems wrong, given that you do compute `inv(J)`

tests/sundials/kinsol_05.cc
```diff
@@ -0,0 +1,145 @@
...
+#include <deal.II/lac/vector.h>
+
+#include <deal.II/sundials/kinsol.h>
+
+#include "../tests.h"
+
+// Solve a nonlinear system but provide only residual function. KINSOL
+// then uses its internal solvers which are based on a
+// finite-difference approximation to the Jacobian and a direct
+// solver.
```
jppelteret, line 28: Same comment about the general problem description

jppelteret (changes requested):
Hi @bangerth, thanks a lot for modernising this! I have a few minor comments (paper cuts, really) that should be quick to address. Otherwise it looks good :-)

source/sundials/kinsol.cc
jppelteret: >  Let me defer this to a later time if you don't mind!

Yip, no problem. Also, if there's a better solution (`NVectorView`) then I'd support us adopting that instead.

include/deal.II/sundials/kinsol.h
```diff
@@ -19,33 +19,34 @@
 #include <deal.II/base/config.h>
 
 #ifdef DEAL_II_WITH_SUNDIALS
-#  if DEAL_II_SUNDIALS_VERSION_LT(4, 0, 0)
```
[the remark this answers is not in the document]
bangerth: Yes, I suspect so. I will check once this is merged.

source/sundials/kinsol.cc
```diff
@@ -270,12 +350,86 @@ namespace SUNDIALS
...
+            {
+              LS->content = nullptr;
+            }
+          if (LS->ops)
+            {
+              free(LS->ops);
+              LS->ops = nullptr;
+            }
+          free(LS);
+          LS = NULL;
```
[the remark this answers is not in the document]
bangerth: No, just a left-over from copy-pasting code examples (which are typically shown as C code) and not catching all places. I've replaced all left-over occurrences in this file.

tests/sundials/kinsol_03.cc
bangerth: Yes, a left-over from copying the bulk of the test from the _02 test. If fixed things here and everywhere else. Thanks for spotting this!

tests/sundials/kinsol_04.cc
bangerth: Yes, also copied from _03 :-(

bangerth:
Thanks, @jppelteret, for the careful read! I fixed all of the places you mention.

source/sundials/kinsol.cc
bangerth: Yes, the `NVectorView` thing is the way to go. It would allow us to keep using the vector we have, rather than copying stuff back and forth. I just need to learn how to use it -- see #11395.

jppelteret (approved):

bangerth:
Related to #11829.

Issue #11589, opened 2021-01-19
Use a more efficient container for p::DistributedTriangulationBase::local_cell_relations 


elauksap:
Related to #11515.

I suggest to use a more efficient container for the `local_cell_relations` member of `parallel::DistributedBase::local_cell_relations`.

Now we have:
```
using cell_relation_t = typename std::tuple<CellStatus, cell_iterator>;
std::vector<cell_relation_t> local_cell_relations;
```

I would suggest to investigate the possibility to use, *e.g.*
```
std::map<cell_iterator, CellStatus> local_cell_relations;
```

~For consistency, the callback methods implemented in the `parallel::distributed::SolutionTransfer` class would better swap some input arguments (from `method(..., CellStatus, cell_iterator, ...)` to `method(..., cell_iterator, CellStatus, ...)`).~

drwells:
This was discussed in #11913 and we decided to keep the arrays. In particular, we serialize and unserialize the array, loop over it to do consistency checks, and directly index it (with indices calculated by p4est) - it's unlikely that a map is going to be a better choice.

PR #11515, opened 2021-01-13
Generalize p::d::SolutionTransfer for p::f::Triangulation

elauksap:
**Warning**: depends on #1580.

Here is a summary of changes:
- `p::f::T::load()/save()` methods have been implemented for (de-)serializing `p::f::Triangulation`s (with @peterrum) and cell-attached data (from) to file.
- The pure virtual `update_cell_relations()` method has been overridden.

Remark: the new abstract interface of `DataTransfer` and `update_cell_relations()` (implemented in #1580) is possibly ready for a future implementation of data transfer across p::f::T refinement/coarsening (through the flags `CELL_PERSIST`, `CELL_REFINE`, `CELL_COARSEN`, `CELL_INVALID`).

PR #11580, opened 2021-01-18
Provide abstract interface to p::d::Triangulation::load()/save()

elauksap:
Related to #11515.

Here is a summary of changes:
- Methods `register_data_attach()`, `notify_ready_to_unpack()`, `load()`, `save()` have been moved to the abstract `parallel::DistributedTriangulationBase` interface.
- The former `local_quadrant_cell_relations` member of `p::d::T` has been re-designed (it doesn't explicitly depend on `p4est` anymore) and moved to the abstract interface.
- `CellAttachedData` and `DataTransfer` are now sub-classes of the abstract interface.
- New `save_attached_data()`, `load_attached_data()` and pure virtual `update_cell_relations()` methods have been implemented/declared in the abstract interface.

Issue #6726, opened 2018-06-07
Adapt the new quadrant_cell_relations feature to the remaining functions that crawl through p4est.


marcfehling:
The <code>quadrant_cell_relations</code> structure introduced with #6601 to <code>distributed::Triangulation</code> contains pointers to <code>p4est</code> quadrants. In fact, the entries in this particular data structure are ordered by the occurrence of quadrants in the underlying <code>sc_array</code> of the <code>p4est</code> object. Thus, we do not need these pointers, since we can reconstruct the quadrant by its index.

I left the quadrant pointer in this structure, because I did not update this new feature on all functions that crawl through the <code>p4est</code> object recursively and some of them may require these pointers. These functions are:
- <code>build_lists</code>
- <code>set_vertices_recursively</code>
- <code>fill_vertices_recursively</code>
- <code>determine_level_subdomain_id_recursively</code>
- <code>match_tree_recursively</code>

If someone wants to review these functions and decides to continue working on it, we could _probably_ gain some performance. And get rid of the quadrant pointers of course.

marcfehling:
The quadrant pointers have been removed from the structure in #11580. The mentioned functions have not been changed.

marcfehling:
#11589 suggests a more efficient data structure.

PR #11913, opened 2021-03-15
Use std::pair for local_cell_relations

elauksap:
This PR proposes to store the `p::DistributedTriangulationBase::local_cell_relations` member as a `std::vector<std::pair<cell_iterator, CellStatus>>`, rather than a `std::vector<std::tuple<CellStatus, cell_iterator>>`.

The `std::tuple` was a leftover from #11515, whereas the swapped sorting `cell_iterator <-> CellStatus` is just for consistency with the `p::d::SolutionTransfer` callback methods (_e.g._ [this one](https://github.com/dealii/dealii/blob/master/include/deal.II/distributed/solution_transfer.h#L345)).

Related to #11589.

### DIFF
--- a/include/deal.II/distributed/tria_base.h
+++ b/include/deal.II/distributed/tria_base.h
@@ -30,7 +30,6 @@
 #include <functional>
 #include <list>
 #include <set>
-#include <tuple>
 #include <utility>
 #include <vector>
 
@@ -689,9 +688,10 @@ namespace parallel
                        const unsigned int n_attached_deserialize_variable);
 
     /**
-     * Go through all cells and store the relations between locally
-     * owned quadrants and cells in the private member
-     * local_cell_relations.
+     * Go through all cells and store the relations between a deal.II cell and
+     * its current CellStatus in the private member local_cell_relations.
+     * For an extensive description of CellStatus, see the documentation
+     * for the member function register_data_attach().
      *
      * The stored vector will be ordered by the occurrence of quadrants.
      */
@@ -704,12 +704,12 @@ namespace parallel
      * description of the latter, see the documentation for the member
      * function register_data_attach().
      */
-    using cell_relation_t = typename std::tuple<CellStatus, cell_iterator>;
+    using cell_relation_t = typename std::pair<cell_iterator, CellStatus>;
 
     /**
-     * Vector of tuples, which each contain a deal.II cell
+     * Vector of pair, each containing a deal.II cell iterator
      * and their relation after refinement. To update its contents, use the
-     * compute_cell_relations member function.
+     * update_cell_relations() member function.
      *
      * The size of this vector is assumed to be equal to the number of locally
      * owned quadrants in the parallel_forest object.
--- a/include/deal.II/distributed/fully_distributed_tria.h
+++ b/include/deal.II/distributed/fully_distributed_tria.h
@@ -266,7 +266,7 @@ namespace parallel
 
       /**
        * Go through all locally owned cells and store the relations between
-       * cells and their status in the private member local_cell_relations.
+       * cells and their CellStatus in the private member local_cell_relations.
        *
        * The stored vector will be ordered by the occurrence of cells.
        */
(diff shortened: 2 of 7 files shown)


peterrum:
I haven't taken a detailed look at the PR yet. But could you shortly describe what operations are performed on the container. The syntax of `std::vector<std::tuple<>>`might not be nice but the possibility that you can pre-allocate contiguous memory makes it normally more efficient. Of course, the question is also if you need the data to be sorted.

kronbichler:
As I am a culprit of many uses of `std::vector` inside deal.II (even though probably not this one), I would also be interested in the kind of operations. My experience with the red-black trees in `std::map` and `std::set` (at least with gcc and clang implementations) has been that they are 3-5x slower than `std::vector` for the kind of operations I tend to do on those arrays, such as integral or double types. It's not to say that the proposed code couldn't be a more efficient choice, but I'd happily learn about what goes wrong currently and what makes the `map` more efficient.

elauksap:
Thanks @peterrum and @kronbichler. A quick performance test has indeed shown that the `std::map` solution is not faster. It was also buggy in the case of coarsening.

I've updated this PR to just a minor style modification.

marcfehling:
I introduced this container back in #6601, and it was one of the first patches I wrote in the library. Let me elaborate on its motivation:

For distributed Triangulations, we go through the whole p4est tree structure recursively to match it with the local copy of the Triangulation. Back then, we walked through the entire tree structure recursively multiple times for different purposes despite only being interested in its leaves. These `quadrant_cell_relations` were meant to describe the relations between p4est quadrant leaves and dealii active cells (hence it's been a tuple of 3 entries: quadrant - relation - cell) and only had to be generated once for all operations within execute_coarsening_and_refinement. Well, not all of them as I didn't adjust all functions yet (see #6726).

All we did with it back then is to iterate over all entries. I'm not sure if you used it in another way in the fullydistributed class. So it is really just a collection of relations, which is why using `std::map` does not make it any faster.

That being said, I don't have a final opinion on the order on how the elements are stored since this will be only used internally. Since the quadrant pointer has been removed in #11580, it makes sense to move it to the first position.

Does using `std::pair` instead of `std::tuple` gain any performance?

elauksap:
Thank you @marcfehling! Actually I don't think that a pair or a tuple is going to exhibit a noticeable difference in terms or performance, but since it is a collection of just two objects (cell iterator and status) per cell a pair seems a more natural choice to me. At least, unless it is planned to add more properties to the structure in the future.

include/deal.II/distributed/tria_base.h
```diff
@@ -704,12 +703,12 @@ namespace parallel
...
 
     /**
-     * Vector of tuples, which each contain a deal.II cell
+     * Vector of pair, each containing a deal.II cell iterator
      * and their relation after refinement. To update its contents, use the
-     * compute_cell_relations member function.
+     * update_cell_relations() member function.
      *
      * The size of this vector is assumed to be equal to the number of locally
      * owned quadrants in the parallel_forest object.
```
marcfehling, line 714: Their relation to what exactly? After removing the p4est quadrant, we need to be more specific.

The deal.II cell belongs to the old mesh before refinement, the CellStatus or relation describes how this old cell will be changed on the new mesh. You might refer to the documentation of the `register_data_attach` function for more information on CellStatus.

Also, we don't have a parallel_forest in TriaBase anymore, correct? I think the size of the vector then corresponds to the number of active cells on the new mesh? Maybe you can elaborate in which order the cells will be inserted.

For distributed Triangulation, each entrie's index corresponds to the position of the p4est quadrant in the underlying p4est sc_array, i.e., `tree.quadrants_offset + idx` where idx is the position of the quadrant in its corresponding tree. Do you have a specific order in which you insert the information in fullydistributed tria?

marcfehling (changes requested):

include/deal.II/distributed/tria_base.h
elauksap: Thanks! I have updated the documentation. The order the information is stored is documented accordingly in each derived class `p::d::T`/`p::f::T`.

marcfehling (approved):

PR #11914, opened 2021-03-15
Enable QGaussSimplex<0>

mschreter:
This PR introduces the special case `QGaussSimplex<0>`.

@peterrum: FYI

### DIFF
--- a/source/base/quadrature_lib.cc
+++ b/source/base/quadrature_lib.cc
@@ -1355,7 +1355,7 @@ QGaussSimplex<dim>::QGaussSimplex(const unsigned int n_points_1D)
   : QSimplex<dim>(Quadrature<dim>())
 {
   // fill quadrature points and quadrature weights
-  if (dim == 1)
+  if (dim == 0 || dim == 1)
     {
       const dealii::QGauss<dim> quad(n_points_1D);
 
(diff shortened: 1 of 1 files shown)


peterrum (commented):
Thanks a lot! May I ask you to update the documentation  of the class.

peterrum (approved):

Rombur:
Why do we need this? There are no other quadrature instantiated for `dim == 0`. The documentation says that `dim == 0` degenerates to `dealii::QGauss<0>(n_points_1D)` but what does that mean?

mschreter:
We work on a code that supports `dim=1` to `dim=3` and both, hex and simplices. In case of `FEFaceValues` one has to provide a face quadrature rule `Quadrature<dim-1>`, which will end up in `Quadrature<0>` for `dim=1`. This special case is handled by `QGauss` in https://github.com/dealii/dealii/blob/3ff4f8939e989fff0e91058b514bccbd8a3e4be0/source/base/quadrature_lib.cc#L35-L44
which we simply reuse in the case of `QGaussSimplex<0>`. This way, any extra treatment for `dim=1` together with `FEFaceValues` can be avoided.

peterrum (approved):
Fine from my side. @Rombur Are you happy with the documentation as it is? Or is `dealii::QGauss<0>(n_points_1D)` too confusing? This is the implementation perspective.

Rombur:
I would remove the documentation. This is an implementation detail that users don't need to know. When I was looking at the other quadratures I couldn't find anything about `dim == 0`.

kronbichler (approved):

mschreter:
@Rombur: Thanks for your comment. I removed the extra documentation for `dim==0`.

PR #11917, opened 2021-03-15
No need to futz with the VTK flags in step-15.

bangerth:
/rebuild

### DIFF
--- a/examples/step-15/step-15.cc
+++ b/examples/step-15/step-15.cc
@@ -676,13 +676,10 @@ namespace Step15
         data_out.add_data_vector(present_solution, "solution");
         data_out.add_data_vector(newton_update, "update");
         data_out.build_patches();
+
         const std::string filename =
           "solution-" + Utilities::int_to_string(refinement, 2) + ".vtk";
-        std::ofstream         output(filename);
-        DataOutBase::VtkFlags vtk_flags;
-        vtk_flags.compression_level =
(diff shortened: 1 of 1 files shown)


drwells:
Why not? I don't see any reason to remove the demonstration of this feature.

bangerth:
There are a million things we *could* do in these tutorials. But just like with every class lecture we teach, you pick a topic and stick with a certain direction you think is most educational. I see no reason why this particular program should be discussing this issue, so I axed it: Focus on one thing, and stick with it.

There isn't even a particular point in doing this here: We *already* use compression in the VTU writer, we just optimize for space where here we select optimizing for time, or the other way around. What's the point in doing that here?

bangerth:
Maybe put differently: Let's assume the program had not done that so far, and I proposed a patch that adds it. Would you think that a good idea? I think my view of these tutorials is that of a minimalist: Everything that isn't really needed should not be there. (I recognize that that's *my* view of the tutorials and am not offended if others feel differently.)

drwells:
Those are good points. My main concern is that setting that flag makes IO a lot faster in some cases (I remember it was like 2x as fast when I added it vs. the space optimization choice) so I think its important that this code show up in the tutorials.

Fortunately it looks like we set up `VtkFlags` in a couple other tutorials so this patch doesn't completely remove documentation of the feature so I am happy with the change (as opposed to deleting the equivalent code in step-9).

drwells (approved):

drwells (approved):

PR #11918, opened 2021-03-15
[WIP] MappingQCache: setup multigrid levels with RemotePointEvaluation

peterrum:
This PR adds an alternative implementation of `MappingQCache` for multgrid levels (see also #11882). It used `RemotePointEvaluation` from #11804 to directly evaluate the coordinate vector without the need to (mis)use the multigrid infrastructure.

It was suggested by @kronbichler in https://github.com/dealii/dealii/pull/11882#discussion_r593913009.

Only the last commit is relevant. This PR should be only merged once some possible bottleneck in #11804 have been eliminated.


peterrum:
Closing, since filling `MappingQCache` this way would be too expensive, since we throw away useful information regarding the connectivity of children and parent cells and instead do a black-box search.

PR #11919, opened 2021-03-15
examples/step-69: use std::copy_n instead of std::copy

tamiko:
A std::copy_n is more more elegant than the std::copy variant that need
to take the address of an element past the momentum.

Thanks to @guermond and @ejtovar for pointing this out.


masterleinad (approved):

bangerth:
Wait, what, @guermond is now on github and pointing out details in deal.II!?!?

Issue #11860, opened 2021-03-06
Generalize global_active_cell_index for p:s:T (without artificial cells)


peterrum:
Once #11858 is merged, the `global_active_cell_index` functions will also work for p:s:T. However, the info is only available for locally-owned cells and for a single helo of cells. This issue is a reminder to extend the algorithms so that in the case of p:s:T without artificial cells all cells get an appropriate index.

marcfehling:
The following tests rely on this feature:
- `mpi/prepare_coarsening_and_refinement_01.mpirun=4` and `mpi/prepare_coarsening_and_refinement_02.mpirun=4` fail because they rely on this feature.
- `mpi/prepare_coarsening_and_refinement_03.mpirun=4.debug` softlocks.
- `mpi/prepare_coarsening_and_refinement_04` has this check for `p:s:T` disabled right now.

If desired, we can disable the `p:s:T` parts of all these tests like in the last one.

peterrum:
Isn't this fixed by #11901?

marcfehling:
> Isn't this fixed by #11901?

Indeed, it fixed it. I didn't rebase on `master` the last time I ran my testsuite. Sorry about that.

PR #11929, opened 2021-03-17
Treat cells to be h-coarsened as their parent in DoFHandler::prepare_coarsening_and_refinement().

marcfehling:
We need to set future FE indices similarly on all siblings of a cell that will be h-coarsened during `DoFHandler::prepare_coarsening_and_refinement()`. Otherwise, the `max_difference` criterion will not be met on the updated mesh for former parent cells.

---

This functionality does not work entirely for `p:s:T`. It relies on the following features:
- #11860 global_active_cell_index for `p:s:T` without artificial cells
- future FE indices need to be accessed on ghost cells.
We currently require that all siblings of a cell need to belong to the same processor.

However, this feature already works with `p:d:T`, so I didn't want to hold it back until it also works for the shared one.

### DIFF
--- a/include/deal.II/dofs/dof_accessor.templates.h
+++ b/include/deal.II/dofs/dof_accessor.templates.h
@@ -2952,6 +2952,11 @@ DoFCellAccessor<dimension_, space_dimension_, level_dof_access>::
         ExcMessage(
           "You ask for information on children of this cell which is only "
           "available for active cells. One of its children is not active."));
+      Assert(child->is_locally_owned(),
+             ExcMessage(
+               "You ask for information on children of this cell which is only "
+               "available for locally owned cells. One of its children is not "
+               "locally owned."));
       future_fe_indices_children.insert(child->future_fe_index());
     }
   Assert(!future_fe_indices_children.empty(), ExcInternalError());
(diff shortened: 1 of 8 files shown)


marcfehling:
Aw man, not again :-(
```
 In file included from /home/runner/work/dealii/dealii/source/dofs/dof_handler.cc:16:0:
/home/runner/work/dealii/dealii/include/deal.II/base/config.h:396:33: internal compiler error: Segmentation fault
 #define DEAL_II_NAMESPACE_CLOSE }
                                 ^
/home/runner/work/dealii/dealii/source/dofs/dof_handler.cc:3792:1: note: in expansion of macro ‘DEAL_II_NAMESPACE_CLOSE’
 DEAL_II_NAMESPACE_CLOSE
 ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-7/README.Bugs> for instructions.
```

bangerth:
We should just go back to using C++98. That would probably solve your problems :-)

marcfehling:
I'll close this PR for now and will come back to it later - after making some other changes to this particular function first.

PR #11933, opened 2021-03-17
Use the correct file extension.

bangerth:
I don't know when this was changed, but step-15 writes VTU file format into files that have `.vtk` extension. My visit version doesn't want to read this. Easy to fix.

/rebuild


peterrum (approved):

bangerth:
Wow, that was pretty quick approval! :-)

Issue #11564, opened 2021-01-15
Generalize (FE|Mapping|Q)Collection for mixed meshes


bangerth:
Following the discussion at the simplex hackathon today, here is what I think we were throwing around as an idea for the various `hp::*Collection` objects:

* Internally, rather than storing a `std::vector<std::shared_ptr<const Mapping<dim, spacedim>>>`, we store a `std::map<ReferenceCell,std::vector<std::shared_ptr<const Mapping<dim, spacedim>>>>`, i.e., a map from reference cell kind to a vector with `active_fe_index` as index.
* These collections are filled by calling `push_back()`. If we can ask a FE/Mapping/Quadrature object for its reference cell, then we can continue to use this interface: You `push_back()` a new element, and it automatically ends up at the end of the vector for the given reference cell kind.
* We currently access elements using `operator[]`. This will have to change since we will have to index not just by `active_fe_index`, but also by reference cell kind. For backward compatibility, we might want to keep that operator and assert that the map has only a single element (i.e., that only objects for the same kind of reference cell type have been added). But in the end, we need a new lookup function that takes both the reference cell type and the active fe index as input. Shouldn't be too hard to add this, and use it inside the library.

Using this kind of scheme, `active_fe_index` becomes completely separate from the reference cell type and can really be used to just designate the polynomial degree, or the physics that are active on that cell. I kind of like this approach. We might have to adjust the tests and programs that are currently already using mixed meshes, though.

@luca-heltai @tjhei @peterrum @marcfehling -- thoughts?

tjhei:
First, the separation between active_fe_index and cell type makes sense to me. Something like ``mappings[cell_type][active_fe_index]`` seems the right approach (with the first indexing operation being a noop if only a simple mapping is specified). How this is stored precisely, is something I am not quite sure about. Maybe it makes sense to turn the logic around to be ``mappings[active_fe_index][cell_type]`` (see below).

From our conversation earlier it seems to me that it would be great to be able to assign something like a ``MappingLinear`` that can take the role of what you write as ``std::map<ReferenceCell, Mapping*>``. 

One suggestion:
1. add a member function ``Mapping& get_mapping_for(ReferenceCell)`` to ``Mapping``. For the existing mappings it can just return ``*this`` with an Assert that the type is correct.
2. Create a new "(bi/tri)linear mapping" class that returns the appropriate mapping in ``get_mapping_for`` and somewhat sensible information for other mapping functions (or throw for nonsensical operations).
[...]

PR #11934, opened 2021-03-17
Introduce hp::Collection

peterrum:
Preparation for #11564 so that we only need to implement the internal logic once.

### DIFF
--- a/include/deal.II/hp/q_collection.h
+++ b/include/deal.II/hp/q_collection.h
@@ -22,6 +22,8 @@
 #include <deal.II/base/quadrature.h>
 #include <deal.II/base/subscriptor.h>
 
+#include <deal.II/hp/collection.h>
+
 #include <memory>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


include/deal.II/hp/q_collection.h
```diff
@@ -235,12 +196,12 @@ namespace hp
...
   QCollection<dim>::operator==(const QCollection<dim> &q_collection) const
   {
-    const unsigned int n_quadratures = size();
+    const unsigned int n_quadratures = this->size();
     if (n_quadratures != q_collection.size())
       return false;
 
     for (unsigned int i = 0; i < n_quadratures; ++i)
-      if (!(*quadratures[i] == q_collection[i]))
+      if ((this->operator[](i) == q_collection[i]) == false)
```
bangerth, line 204:
```suggestion
      if (this->operator[](i) != q_collection[i])
```
:-)

bangerth (approved):
I'll merge to allow you to make progress, but please address the one issue in a follow-up patch.

include/deal.II/hp/q_collection.h
peterrum: I'll do (tomorrow).

PR #11935, opened 2021-03-17
Fix some links in the AD/SD documentation module

jppelteret:
I just qualified the namespace that the AD classes live in, so that links would appear in the documentation. I also added the `%` in front of "Differentiation" so that it wouldn't link to the namespace.


bangerth (approved):

bangerth:
Doc only, so merging.

PR #11927, opened 2021-03-16
Identity function

drwells:
I've needed this a few times recently for setting up a `MappingFEField`.

PR #11936, opened 2021-03-17
Better document IdentityFunction.

drwells:
This came up in review of #11927 but wasn't fixed before it got merged.


peterrum:
I see; the purpose seems to be same as `VectorTools::get_position_vector()`, isn't it?

jppelteret (approved):

drwells:
> I see; the purpose seems to be same as VectorTools::get_position_vector(), isn't it?

Yes, but its nice to be able to handle deformations in a more generic way (e.g., I would like to use `IdentityFunction` as the default value but call `interpolate` in all code paths)

peterrum (approved):

PR #11938, opened 2021-03-18
LinearAlgebra::ReadWriteVector accept serial vectors

peterrum:


include/deal.II/lac/read_write_vector.templates.h
```diff
@@ -409,6 +410,68 @@ namespace LinearAlgebra
...
+    import_serial_vector(const VectorType &       values,
+                         VectorOperation::values  operation,
+                         ReadWriteVector<Number> &rw_vector)
+    {
+      using size_type = types::global_dof_index;
+
+      const IndexSet &stored = rw_vector.get_stored_elements();
+      if (operation == VectorOperation::add)
+        for (size_type i = 0; i < stored.n_elements(); ++i)
+          rw_vector.local_element(i) += values(stored.nth_index_in_set(i));
```
Rombur, line 426: This is wrong. It assumes that the `IndexSet` contains all the elements in `[0, stored.n_elements)` but this is not a requirement of the `ReadWriteVector`.

Rombur (changes requested):

include/deal.II/lac/read_write_vector.templates.h
peterrum: In what sense is this different from 

https://github.com/dealii/dealii/blob/03f2c7adb24ee8d8fd18e278a4e45cb04e224d0d/include/deal.II/lac/read_write_vector.templates.h#L139-L156

?

Rombur: Sorry you are right. This is correct.

PR #11939, opened 2021-03-18
a new class TimerTree for flexible and modular timer measurements

nfehn:
This PR proposes a new implementation for wall-time measurements that can be used in a flexible and modular way in complex codes.

**Design:** 
The underlying design principle are mainly the open-closed principle and the dependency inversion principle.

As far as I understand the existing `dealii::TimerOutput`, it can be used easily in a dealii tutorial where everything is plugged together within one file, or for use within a certain class. However, for complex applications, timer measurements are spread over many software layers from high-level to low-level code. The low-level code, of course, does not know for which purpose it is used. Instead, a module should be able to query the timings from its sub-modules and plug them together in a hierarchical way.

**Example:**
To give a concrete example consider the application of fluid-structure interaction solved in a partitioned way by plugging together different modules for the solution of the single-field problems. In my opinion, the present PR proposes a minimally invasive way of how to implement useful timing measurements for such problems, without touching low-level code (in this case the single-field solvers). One design (not chosen here) would be to pass a string with an ``id`` to a sub-module that indicates for which purpose this module is used, for example "FSI - fluid" handed to the constructor of the fluid class in case of an FSI application, instead of just "fluid" for a pure fluid solver. Such a design is, however, not minimally invasive and requires to touch many interfaces. The present design, instead, is that the FSI solver asks its sub-module (dependency inversion!), the fluid solver, for its timings by a member function of the fluid class, ``Fluid::get_timings()`` (open-closed!), and the FSI class has to take care of how to insert that sub-tree into the outer timer tree for the FSI application.

**Tests and usability:**
A test shows the capabilities of ``TimerTree`` and how to use this class.

Note that a main strength of this class is in my opinion that it always computes the share of time not covered by measurements (called "Other"), so that the developer/user always sees immediately whether there is a portion of the code that did not get attention regarding timings but must actually be quite important in terms of run-time, giving indications that the measurements need to be extended to improve the understanding of where the code spends its time.

This class is not intended for use in micro-benchmarks or inner-most loops (just as the `Timer` class), since the overhead accompanied by this class would certainly be no longer negligible. This effect is, however, negligible when measuring different parts of one solver module (preprocessing, linear solver, postprocessing) or when coupling different modules.

### DIFF
--- a/include/deal.II/base/timer_tree.h
+++ b/include/deal.II/base/timer_tree.h
@@ -56,12 +56,17 @@ public:
   void
   insert(std::vector<std::string> const ids, double const wall_time)
   {
-    AssertThrow(ids.size() > 0, ExcMessage("empty name."));
+    Assert(
+      ids.size() > 0,
+      ExcMessage(
+        "You need to specify a not empty identifier 'ids' in order to insert a wall_time into the tree."));
 
     if (this->id == "") // the tree is currently empty
       {
-        AssertThrow(sub_trees.empty(),
-                    ExcMessage("invalid state found. aborting."));
+        Assert(
+          sub_trees.empty(),
+          ExcMessage(
+            "The tree is currently empty and does not have an 'id'. Sub trees must therefore also be empty."));
 
         this->id = ids[0];
 
@@ -120,9 +125,10 @@ public:
       }
     else // the provided name does not fit to this tree
       {
-        AssertThrow(false,
-                    ExcMessage("The name provided is " + ids[0] +
-                               ", but must be " + id + " instead."));
+        Assert(false,
+               ExcMessage("The name provided is ids[0] = <" + ids[0] +
+                          ">, but the tree has the id = <" + id +
+                          "> instead."));
       }
   }
 
@@ -130,16 +136,29 @@ public:
    * Inserts a whole sub_tree into an existing tree, where
    * the parameter @p ids specifies the place at which to insert the @p sub_tree.
    * This function allows to combine different timer trees in a modular way.
-   * If a non empty string @p new_name is provided, the id of @p sub_tree is
-   * replaced by @p new_name when inserted into the tree.
+   *
+   * @note If a sub-tree with the same id as @p sub_tree already exists in the tree,
+   * the existing sub-tree is replaced by @p sub_tree.
+   *
+   * @note If a non empty string @p new_name is provided, the id of @p sub_tree is
+   * replaced by @p new_name when inserted into the tree. This is necessary if a
+   * program uses the same module multiple times but for different purposes.
    */
   void
   insert(std::vector<std::string> const         ids,
          std::shared_ptr<TimerTree const> const sub_tree,
          std::string const                      new_name = "")
   {
-    AssertThrow(ids.size() > 0, ExcMessage("Empty ID specified."));
-    AssertThrow(id == ids[0], ExcMessage("Invalid ID specified."));
+    Assert(
+      ids.size() > 0,
+      ExcMessage(
+        "You need to specify a not empty identifier 'ids' in order to insert a sub_tree."));
+
+    Assert(id == ids[0],
+           ExcMessage(
+             "You want to insert a sub_tree with ids[0] = <" + ids[0] +
+             ">, but this identifier does not match the tree's id = <" + id +
+             ">."));
 
     std::vector<std::string> remaining_id = erase_first(ids);
 
@@ -158,26 +177,31 @@ public:
 
     if (found == false)
       {
-        AssertThrow(
+        Assert(
           remaining_id.size() == 0,
           ExcMessage(
-            "Subtree can not be inserted since the specified ID does not exist in this tree."));
+            "Subtree can not be inserted since the specified identifier <" +
+            remaining_id[0] + "> does not exist in this tree."));
 
         std::shared_ptr<TimerTree> new_tree(new TimerTree());
         new_tree->copy_from(sub_tree);
+
+        // The sub_tree's id can be replaced by a new name (if specified when
+        // calling this function)
         if (!new_name.empty())
           new_tree->id = new_name;
 
-        // Make sure that the new tree does not already exist
-        for (auto it = sub_trees.begin(); it != sub_trees.end(); ++it)
-          {
-            AssertThrow(
-              new_tree->id != (*it)->id,
-              ExcMessage(
-                "Subtree can not be inserted since the tree already contains a subtree with the same ID."));
-          }
+        // search whether the new tree already exists
+        auto it = sub_trees.begin();
+        for (; it != sub_trees.end(); it++)
+          if ((*it)->id == new_tree->id)
+            break;
 
-        sub_trees.push_back(new_tree);
+        // replace existing sub-tree if new sub-tree already exists
+        if (it != sub_trees.end())
+          *it = new_tree;
+        else // otherwise, add another sub-tree
+          sub_trees.push_back(new_tree);
       }
   }
 
@@ -232,7 +256,10 @@ private:
   std::vector<std::string>
   erase_first(std::vector<std::string> const &in) const
   {
-    AssertThrow(in.size() > 0, ExcMessage("empty name."));
+    Assert(
+      in.size() > 0,
+      ExcMessage(
+        "First entry of vector can not be erased since the provided vector is empty."));
 
     std::vector<std::string> out(in);
     out.erase(out.begin());
--- a/include/deal.II/base/timer_tree.h
+++ b/include/deal.II/base/timer_tree.h
@@ -450,8 +450,14 @@ private:
       other.print_id_and_data(pcout, offset, length, relative, ref_time);
   }
 
+  /**
+   * Identifier of this tree.
+   */
   std::string id;
 
+  /**
+   * Struct describing the data of a tree.
+   */
   struct Data
   {
     Data()
@@ -461,12 +467,25 @@ private:
     double wall_time;
   };
 
+  /**
+   * Data of this tree.
+   */
   std::shared_ptr<Data> data;
 
+  /**
+   * A vector of sub-trees.
+   */
   std::vector<std::shared_ptr<TimerTree>> sub_trees;
 
+  /**
+   * Offset per level for formatted output specified in number of characters.
+   */
   static constexpr unsigned int offset_per_level = 2;
-  static constexpr unsigned int precision        = 2;
+
+  /**
+   * Value passed to std::setprecision() in print functions.
+   */
+  static constexpr unsigned int precision = 2;
 };
 
 
--- a/include/deal.II/base/timer_tree.h
+++ b/include/deal.II/base/timer_tree.h
@@ -24,7 +24,6 @@ DEAL_II_NAMESPACE_OPEN
 /**
  * A class for wall-time measurements enabling a modular coupling of modules in
  * a minimally invasive way.
- *
  */
 class TimerTree
 {
@@ -49,9 +48,9 @@ public:
   }
 
   /**
-   * Inserts a measured wall_time into the tree, by
+   * Inserts a measured @p wall_time into the tree, by
    * either creating a new entry in the tree if this function is called
-   * the first time with this ID, or by adding the wall_time to an
+   * the first time with the ID @p ids, or by adding the @p wall_time to an
    * entry already existing in the tree.
    */
   void
@@ -129,10 +128,10 @@ public:
 
   /**
    * Inserts a whole sub_tree into an existing tree, where
-   * the parameter names specifies the place at which to insert the sub_tree.
+   * the parameter @p ids specifies the place at which to insert the @p sub_tree.
    * This function allows to combine different timer trees in a modular way.
-   * If a non empty string new_name is provided, the id of  sub_tree is
-   * replaced by new_name when inserted into the tree.
+   * If a non empty string @p new_name is provided, the id of @p sub_tree is
+   * replaced by @p new_name when inserted into the tree.
    */
   void
   insert(std::vector<std::string> const         ids,
@@ -203,8 +202,9 @@ public:
    * a wall time has been set for the parent of these children. In this
    * case, an additional item `other` is created in order to give insights
    * to which extent the code has been covered with timers and to which
-   * extend time is spent is other code paths that are currently not
-   * covered by timers.
+   * extend time is spent in other code paths that are currently not
+   * covered by timers. The parameter @p level specifies the level for which
+   * to print results, where a value of 0 corresponds to the high level module.
    */
   void
   print_level(ConditionalOStream const &pcout, unsigned int const level) const
--- a/include/deal.II/base/timer_tree.h
+++ b/include/deal.II/base/timer_tree.h
@@ -1,6 +1,6 @@
 // ---------------------------------------------------------------------
 //
-// Copyright (C) 1998 - 2020 by the deal.II authors
+// Copyright (C) 2021 by the deal.II authors
 //
 // This file is part of the deal.II library.
 //
--- a/include/deal.II/base/timer_tree.h
+++ b/include/deal.II/base/timer_tree.h
@@ -31,9 +31,22 @@ public:
   /**
    * Constructor.
    */
-  TimerTree()
+  TimerTree(MPI_Comm const &comm)
     : id("")
+    , comm(comm)
   {}
+
+  /**
+   * Copy assignment operator.
+   */
+  TimerTree &
+  operator=(TimerTree const &other)
+  {
+    this->id        = other.id;
+    this->data      = other.data;
+    this->sub_trees = other.sub_trees;
+    this->comm      = other.comm;
+  }
 
   /**
    * Clears the content of this tree. Sub trees inserted
@@ -42,9 +55,10 @@ public:
   void
   clear()
   {
-    this->id = "";
+    this->id = std::string();
     data     = nullptr;
     sub_trees.clear();
+    this->comm = MPI_Comm();
   }
 
   /**
@@ -81,8 +95,7 @@ public:
           {
             std::vector<std::string> remaining_id = erase_first(ids);
 
-            std::shared_ptr<TimerTree> new_tree;
-            new_tree.reset(new TimerTree());
+            std::shared_ptr<TimerTree> new_tree(new TimerTree(comm));
             new_tree->insert(remaining_id, wall_time);
             sub_trees.push_back(new_tree);
           }
@@ -116,8 +129,7 @@ public:
 
             if (found == false)
               {
-                std::shared_ptr<TimerTree> new_tree;
-                new_tree.reset(new TimerTree());
+                std::shared_ptr<TimerTree> new_tree(new TimerTree(comm));
                 new_tree->insert(remaining_id, wall_time);
                 sub_trees.push_back(new_tree);
               }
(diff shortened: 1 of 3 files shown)


jppelteret:
Hi @nfehn, this looks like a pretty neat concept. Although I've not reviewed the branch per se, I have a two small remarks that might help improve what you've got before you get some more concrete feedback.

Firstly, we typically split up class method declarations and definitions between the header and source files, so could you please transfer all of the definitions that are currently inline into a source file?

The second thing is that the class introduction (what's documented) is really concise. I don't think that the users would understand what this class is for, or why they'd want to use it rather than the existing`TimerOutput`. But you'd got this really expansive first comment in the PR that describes so nicely what the purpose of the class is. Why doesn't you transfer all of those ideas to the documentation?

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 1998 - 2020 by the deal.II authors
```
jppelteret, line 3: This should only be for 2021.

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+    for (auto it = sub_trees.begin(); it != sub_trees.end(); ++it)
+      {
+        length = std::max(length, (*it)->get_length() + offset_per_level);
+      }
+
+    return length;
+  }
+
+  void
+  do_print_plain(ConditionalOStream const &pcout,
```
jppelteret, line 271: All of these private member functions, as well as the member variables and internal structs, should also be documented.

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+
+    double wall_time;
+  };
+
+  std::shared_ptr<Data> data;
+
+  std::vector<std::shared_ptr<TimerTree>> sub_trees;
+
+  static unsigned int const offset_per_level = 2;
+  static unsigned int const precision        = 2;
```
jppelteret, line 417: Unless made `constexpr`, these `static` will also need to be defined in the source file.

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+ * a minimally invasive way.
+ */
+class TimerTree
+{
+public:
+  /**
+   * Constructor.
+   */
+  TimerTree()
+    : id("")
```
bangerth, line 35: That's the default. You could just write `TimerTree() = default;` to declare the constructor.

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+  {}
+
+  /**
+   * Clears the content of this tree. Sub trees inserted
+   * into this tree via pointers to external trees are not touched.
+   */
+  void
+  clear()
```
bangerth, line 43: That's confusing. The function name `clear()` to me indicates that the object is reset to the same state one would get when calling the default constructor.

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  /**
+   * Inserts a measured @p wall_time into the tree, by
+   * either creating a new entry in the tree if this function is called
+   * the first time with the ID @p ids, or by adding the @p wall_time to an
+   * entry already existing in the tree.
+   */
+  void
+  insert(std::vector<std::string> const ids, double const wall_time)
```
bangerth, line 57:
```suggestion
  insert(const std::vector<std::string> &ids, const double wall_time)
```

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+  {
+    AssertThrow(ids.size() > 0, ExcMessage("empty name."));
+
+    if (this->id == "") // the tree is currently empty
+      {
+        AssertThrow(sub_trees.empty(),
+                    ExcMessage("invalid state found. aborting."));
```
bangerth, line 64: Let's use complete sentences in error messages, starting with a capital letter, subject verb and noun.

The "aborting" part isn't technically true -- you're just throwing an exception here, which the user can do with as they please. I would remove that part.

Separately: Why use `AssertThrow` instead of `Assert` for these assertions?

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+
+        this->id = ids[0];
+
+        if (ids.size() == 1) // leaves of tree reached, insert the data
+          {
+            data.reset(new Data());
```
bangerth, line 70:
```suggestion
            data = std::make_shared<Data>();
```

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+            data->wall_time += wall_time;
+
+            return;
+          }
+        else // go deeper
+          {
+            std::vector<std::string> remaining_id = erase_first(ids);
+
+            std::shared_ptr<TimerTree> new_tree;
+            new_tree.reset(new TimerTree());
```
bangerth, line 80: same here and maybe also in other places; we try to avoid bare 'operator new'  calls.

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+                new_tree.reset(new TimerTree());
+                new_tree->insert(remaining_id, wall_time);
+                sub_trees.push_back(new_tree);
+              }
+          }
+      }
+    else // the provided name does not fit to this tree
+      {
+        AssertThrow(false,
+                    ExcMessage("The name provided is " + ids[0] +
```
bangerth, line 124: Here and in the following line (and maybe elsewhere in the patch): since `ids[0]` can be any sort of text, the resulting error message can be difficult to read if, for example, someone provides `ids[0]="difficult to read"`. A better way is to make clear what is text and what is a quoted variable name:
```suggestion
                    ExcMessage("The name provided is <" + ids[0] + ">" +
```

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+   * Relative wall times are printed for all children of a sub-tree if
+   * a wall time has been set for the parent of these children. In this
+   * case, an additional item `other` is created in order to give insights
+   * to which extent the code has been covered with timers and to which
+   * extend time is spent in other code paths that are currently not
+   * covered by timers. The parameter @p level specifies the level for which
+   * to print results, where a value of 0 corresponds to the high level module.
+   */
+  void
+  print_level(ConditionalOStream const &pcout, unsigned int const level) const
```
bangerth, line 210: You restrict yourself unnecessarily by hard-coding `ConditionalOStream` for the first argument. I would use `template <typename OStreamType>`

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+  copy_from(std::shared_ptr<TimerTree const> const other)
+  {
+    *this = *other;
+  }
+
+  /*
+   * This function erases the first entry of the vector.
+   */
+  std::vector<std::string>
+  erase_first(std::vector<std::string> const &in) const
```
bangerth, line 233:
```suggestion
  erase_first(const std::vector<std::string> &in) const
```

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+  {
+    AssertThrow(in.size() > 0, ExcMessage("empty name."));
+
+    std::vector<std::string> out(in);
+    out.erase(out.begin());
```
bangerth, line 238: If you do this operation frequently, you should look at using `std::deque` instead of `std::vector` since that allows for `pop_front()`.

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  /*
+   * Returns average wall-time over all MPI processes for the root of this tree.
+   */
+  double
+  get_average_wall_time() const
+  {
+    Utilities::MPI::MinMaxAvg time_data =
+      Utilities::MPI::min_max_avg(data->wall_time, MPI_COMM_WORLD);
```
bangerth, line 250: Using `MPI_COMM_WORLD` is forbidden. You need to provide the communicator as an argument  to the function.

bangerth (commented):
Just a couple of comments, but I haven't looked in any kind of detail. Like @jppelteret , I would very much like it if you could extend the class documentation, possibly with an example of us.

bangerth:
As for the design: What you currently have leaves it to the individual modules to figure out timing information. That seems suboptimal, we already have classes that do timing quite well.

Have you considered a design wherein the individual modules use `TimerOutput`, and you come up with a way by which several `TimerOutput` objects can somehow be combined? Or whereby a `TimerOutput` object can have *other* `TimerOutput` objects that feed information up? In other words, each timer output could be considered as a leaf in your tree (when used like we use it right now) or as a node that points to other nodes (if you come up with a way to connect these objects)? That way, the individual module would only have to make its own `TimerOutput` object available to the next higher level for integration into that higher level's `TimerOutput` object.

bangerth:
As for the design: What you currently have leaves it to the individual modules to figure out timing information. That seems suboptimal, we already have classes that do timing quite well.

Have you considered a design wherein the individual modules use `TimerOutput`, and you come up with a way by which several `TimerOutput` objects can somehow be combined? Or whereby a `TimerOutput` object can have *other* `TimerOutput` objects that feed information up? In other words, each timer output could be considered as a leaf in your tree (when used like we use it right now) or as a node that points to other nodes (if you come up with a way to connect these objects)? That way, the individual module would only have to make its own `TimerOutput` object available to the next higher level for integration into that higher level's `TimerOutput` object.

nfehn:
I think I will wait with all the low-level changes whether the general concept is accepted or not. 

In my opinion, `TimerTree` is more than using `TimerOutput` for its leaves, see all the explanations in the very beginning.

There are indeed many reasons why I think `TimerOutput` is not well-designed. (i) Fat public interface, (ii) depends on the class `Timer` (generally not a good design in my opinion, dependency inversion principle, etc.), (iii) you remind me to not use `ConditionalOStream` (a recommendation that I consider technically correct) but `TimerOutput` does it as well (even has a private member variable of this type which is even worse in my opinion in terms of design), (iv) `TimerOutput` is usable for deal.II tutorials but this does not justify your attribute "quite well" in my opinion. 

Yes, one could change `TimerOutput` for the good in my opinion, but this would break compatibility if one wants to have a design that is understandable. Since I did not see chances to get such changes merged, I came up with proposing a new (in my opinion "clean") design.

include/deal.II/base/timer_tree.h
nfehn: Could you please give me a hint in which sense the function `clear()` does not fulfill your specification, i.e., your comment is too implicit for me?

include/deal.II/base/timer_tree.h
nfehn: improved asserts here https://github.com/dealii/dealii/pull/11939/commits/c7dd464584f2cd84d71cd857684b5cfb755daf5a

include/deal.II/base/timer_tree.h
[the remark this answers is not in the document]
nfehn: same as above

include/deal.II/base/timer_tree.h
nfehn: This is how I learned C++ ("`const` is a trailing keyword"). Could you please reference me to some coding guidelines stating how deal.II wants to have these things?

kronbichler: I believe this not written down explicitly, but it is the de facto convention applied inside deal.II. The coding guidelines state all their examples as a leading keyword, https://dealii.org/developer/doxygen/deal.II/CodingConventions.html , but I think we should state that this is the convention used in that document, to avoid confusion like this.

include/deal.II/base/timer_tree.h
nfehn: as above. So I am resolving this one here because changes will be made consistently anyway.

include/deal.II/base/timer_tree.h
nfehn: is irrelevant since mpi comm needs to be passed to constructor according to comment below.

include/deal.II/base/timer_tree.h
nfehn: changed here https://github.com/dealii/dealii/pull/11939/commits/647d0da4b0f431358b92b969ce5b3fcf983abbda

include/deal.II/base/timer_tree.h
nfehn: has been changed

include/deal.II/base/timer_tree.h
nfehn: done here https://github.com/dealii/dealii/pull/11939/commits/78670552974857ec20150fcc0a7b464bfdbbd087 and https://github.com/dealii/dealii/pull/11939/commits/787cf55e96112e43d7d042a6e1cd6f2d01b912d9

include/deal.II/base/timer_tree.h
nfehn: done here https://github.com/dealii/dealii/pull/11939/commits/4ece8448a5f152b3d55845152ab3c9a69d7decfc

include/deal.II/base/timer_tree.h
nfehn: done here https://github.com/dealii/dealii/pull/11939/commits/e7a0fdd8c914cac1eed20326b0a50e94c901bae9

@kronbichler Yes, I suggest that deal.II is more precise about this, e.g. "In deal.II, const is only used as a trailing keyword where necessary. In all other cases that allow to use const either as a trailing or leading keyword, const is used as a leading keyword in deal.II."

include/deal.II/base/timer_tree.h
```diff
@@ -0,0 +1,423 @@
...
+          }
+      }
+    else
+      {
+        // only print name
+        print_name(pcout, offset, length);
+
+        // recursively print sub trees (decreasing the level and incrementing
+        // the offset)
+        for (auto it = sub_trees.begin(); it != sub_trees.end(); ++it)
```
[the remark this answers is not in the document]
nfehn: done here https://github.com/dealii/dealii/pull/11939/commits/faaff622d5c1273a84bb74f2eb76062d3d99c0be

include/deal.II/base/timer_tree.h
nfehn: I think this is something one could also change later.

include/deal.II/base/timer_tree.h
nfehn: done here  https://github.com/dealii/dealii/pull/11939/commits/5581aacf5ec7433717b86a0675a0e679cbe9d40f

nfehn:
For a concrete example and to see the benefits of `TimerTree`, see https://github.com/exadg/exadg/pull/49 . Before using `TimerTree`, a separate class has been used (`MultigridTimings`) and the output of timings has been triggered by the destructor of the class `MultigridAlgorithm` (V-cycle implementation). Now imagine the program contains different multigrid instances, e.g., one for the velocity and one for the pressure in incompressible Navier-Stokes. With the previous implementation, one just gets multiple multigrid timings in the end when the destructors are called and one does not even know which timings belong to the velocity and which to the pressure. The output of `TimerTree` nicely shows the hierarchy and produces easy-to-interpret results.

kronbichler:
In my opinion, this class is a nice addition for some structured output of nested information. The fact that this class does not hardcode which kind of time is chosen but lets the user supply the timing information has a clear merit, because this allows to collect not only wall times, but in fact any characterizing number. When seeing this class, I immediately thought of collecting some division of work into grains (say "points to be updated"), or some other time than the wall time. In addition, the ability to nest the information by returning the inner object's timer is cleaner to implement than if we pass in an outer timer object (say `TimerOutput`) and try to hook into that. So in that regard, I see this class largely orthogonal to `TimerOuput`, because it tries to address a different problem, and does not try to cover too much with a single interface.

> There are indeed many reasons why I think `TimerOutput` is not well-designed. 

I do not really agree with this assessment, at least in terms of the arguments you have given. (As an aside, the "negative" tone is unnecessary and does not do your arguments a favor, because the classes aim at different problems, at least as I see it, so there is no need to talk badly about something else, but what is good here.) While there is maybe some quirks in the `TimerOutput` constructors, the rest of the interface is indeed pretty lean in my opinion because as a user I only need to define the subsections to be used. Also, I would definitely not discount the advantages of of using the `Timer` class inside, because one only needs to work with the `TimerOutput::Scope` with a name for the section, and nothing else. But as I said, I see a value in both ways. Maybe we can add such tree-like output also for `TimerOutput`, and maybe we find a possibility to combine some of these strengths. But I have the fear of over-engineering things when we try to put all into one class, given what is reasonable effort to put into a timer class, which pretty far from what all of us developers gets paid for.

One thing that I would add is one the documentation of the way how to insert sub-trees and the composition possibilites in general, because it does require quite some attention to get right (I had to sit down for a long while to really see how I would do it). In addition, I think users might want more control than just printing the MPI average. For example, have a look at https://github.com/dealii/dealii/blob/a551e1a2594ee65abbadd2e590ef1b1055b43b95/include/deal.II/base/timer.h#L791-L810 that also prints the skew between MPI ranks. This can be very important to locate hotspots in terms of communication. (This is also a reason why classes tend to get fatter over time, because there are often good reason to add this or that functionality, which accumulates over time.)

nfehn:
@kronbichler I appreciate your feedback regarding the design, and the side note. I think (or hope) I did not mention `TimerOutput` in a negative way in my original long comment (but focused on what is good here in my opinion as you suggested actually (?), so I think it might be that you also have kind of a moralizing tone here; which I just want to provide as some form of feedback as you did before). In the very beginning, I emphasized that I want to provide improved functionality (which is probably the case for most PRs in my opinion). Mentioning the weaknesses of a previous design is a way of being precise about the changes/improvements being made, in my opinion. The reason why I termed `TimerOutput` being "not well-designed" is directly related to the comment by @bangerth regarding the design and whether both classes should be combined (I have the impression that the "not well-designed" statement is cited a bit out of the context since I don't see a reference in this context to my original explanations which are also part of my overall assessment of `TimerOutput`), where I wanted to make clear that `TimerTree` has a different purpose, in my opinion. I accept your criticism and apologize if developers of `TimerOutput` felt critized by this statement. I would like to mention, though, that I now wonder why you do not refer to the statement made previously "What **you currently have** .. That seems suboptimal, **we already have** classes that do timing quite well.", that sounds a bit arrogant to me, given the indirectness of the second part of the statement (which does not refer to the arguments that I tried to provide in the very beginning).

nfehn:
Recently, a discussion in our group came up how to do timings and how to get nice output. As a result of this discussion, there is a clear interest to have `TimerTree` in deal.II, since we would otherwise have to copy this code (currently implemented in [ExaDG)](https://github.com/exadg/exadg) to different code projects based on deal.II (which we want to avoid). It was also an outcome of this discussion that the output produced by `dealii::TimerOutput` is difficult to read for problems with several levels.

Regarding the above discussion, it certainly became quite heated, now reading it again after several months. I certainly made mistakes, but I think mistakes happened on both sides ... In any case, I would appreciate if we could discuss whether `TimerTree` has a value for deal.II or not. I had the feeling that @jppelteret and @kronbichler appreciated the design, but I also wondered why I did not get answers for months. It would also be acceptable for me if deal.II conclusively says "no" to `TimerTree`.

PR #11940, opened 2021-03-18
Add bcast function

bergbauer:
Thought this might be useful

### DIFF
--- a/doc/news/changes/minor/20210318MaximilianBergbauer
+++ b/doc/news/changes/minor/20210318MaximilianBergbauer
@@ -1,0 +1,3 @@
+New: Adds the new generalized MPI_Bcast function Utilities::MPI::broadcast() for arbitrary data types T.
+<br>
+(Maximilian Bergbauer, 2021/03/18)
--- a/include/deal.II/base/mpi.h
+++ b/include/deal.II/base/mpi.h
@@ -1072,6 +1072,28 @@ namespace Utilities
            const unsigned int root_process = 0);
 
     /**
+     * Sends an object @p object_to_send from the process @p root_process
+     * to all other processes.
+     *
+     * A generalization of the classic `MPI_Bcast` function that accepts
+     * arbitrary data types `T`, as long as Utilities::pack() (which in turn
+     * uses `boost::serialize`, see in Utilities::pack() for details) accepts
+     * `T` as an argument.
+     *
+     * @param[in] comm MPI communicator.
+     * @param[in] object_to_send An object to send to all processes.
+     * @param[in] root_process The process that sends the object to all
+     * processes. By default the process with rank 0 is the root process.
+     *
+     * @return Every process receives the object sent by the @p root_process.
+     */
+    template <typename T>
+    T
+    broadcast(const MPI_Comm &   comm,
+              const T &          object_to_send,
+              const unsigned int root_process = 0);
+
+    /**
      * A function that combines values @p local_value from all processes
      * via a user-specified binary operation @p combiner and distributes the
      * result back to all processes. As such this function is similar to
@@ -1460,6 +1482,37 @@ namespace Utilities
             }
         }
       return received_objects;
+#  endif
+    }
+
+    template <typename T>
+    T
+    broadcast(const MPI_Comm &   comm,
+              const T &          object_to_send,
+              const unsigned int root_process)
+    {
+#  ifndef DEAL_II_WITH_MPI
+      (void)comm;
+      (void)root_process;
+      return object_to_send;
+#  else
+      const auto n_procs = dealii::Utilities::MPI::n_mpi_processes(comm);
+      AssertIndexRange(root_process, n_procs);
+      (void)n_procs;
+
+      std::vector<char> buffer      = Utilities::pack(object_to_send, false);
+      unsigned int      buffer_size = buffer.size();
+
+      // Exchanging the size of buffer
+      int ierr = MPI_Bcast(&buffer_size, 1, MPI_UNSIGNED, root_process, comm);
+      AssertThrowMPI(ierr);
+
+      buffer.resize(buffer_size);
+      ierr =
+        MPI_Bcast(buffer.data(), buffer_size, MPI_CHAR, root_process, comm);
+      AssertThrowMPI(ierr);
+
+      return Utilities::unpack<T>(buffer, false);
 #  endif
     }
 
--- a/include/deal.II/base/mpi.templates.h
+++ b/include/deal.II/base/mpi.templates.h
@@ -593,13 +593,7 @@ namespace Utilities
             }
 
           // 2) broadcast result
-          std::vector<char> temp = Utilities::pack(result, false);
-          unsigned int      size = temp.size();
-          MPI_Bcast(&size, 1, MPI_UNSIGNED, 0, comm);
-          temp.resize(size);
-          MPI_Bcast(temp.data(), size, MPI_CHAR, 0, comm);
-
-          return Utilities::unpack<T>(temp, false);
+          return Utilities::MPI::broadcast(comm, result);
         }
 #endif
       (void)comm;
--- a/tests/mpi/broadcast_01.cc
+++ b/tests/mpi/broadcast_01.cc
@@ -1,0 +1,83 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2021 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+
+// Test Utilities::MPI::broadcast().
+
+#include <deal.II/base/mpi.h>
+
+#include "../tests.h"
+
+struct TestObject
+{
+  std::vector<int> int_vector;
+  std::string      string;
+
+  template <class Archive>
+  void
+  serialize(Archive &ar, const unsigned int /*version*/)
+  {
+    ar &int_vector;
+    ar &string;
+  }
+
+  friend std::ostream &
+  operator<<(std::ostream &os, const TestObject &test_object);
+};
+
+std::ostream &
+operator<<(std::ostream &os, const TestObject &test_object)
+{
+  for (const auto &i : test_object.int_vector)
+    os << i << " ";
+
+  os << test_object.string;
+
+  return os;
+}
+
+void
+check(const std::vector<int> &int_vector, const std::string &string)
+{
+  const auto my_proc = Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
+
+  TestObject test_object;
+
+  if (my_proc == 0)
+    {
+      test_object.int_vector = int_vector;
+      test_object.string     = string;
+    }
+
+  const auto result = Utilities::MPI::broadcast(MPI_COMM_WORLD, test_object);
+
+  deallog << result << " ";
+  deallog << std::endl;
+}
+
+
+int
+main(int argc, char *argv[])
+{
+  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+  MPILogInitAll                    log;
+
+  // broadcast test
+  check({1, 2, 3}, "test");
+
+  // broadcast test
+  check({-1, 0, 1}, "i love democracy");
+}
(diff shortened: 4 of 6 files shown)


include/deal.II/base/mpi.h
```diff
@@ -1463,6 +1482,45 @@ namespace Utilities
...
+      (void)comm;
+      (void)root_process;
+      return object_to_send;
+#  else
+      const auto n_procs = dealii::Utilities::MPI::n_mpi_processes(comm);
+      const auto my_rank = dealii::Utilities::MPI::this_mpi_process(comm);
+
+      AssertIndexRange(root_process, n_procs);
+
+      std::vector<char> buffer      = Utilities::pack(object_to_send, true);
```
peterrum, line 1501:
```suggestion
      std::vector<char> buffer      = Utilities::pack(object_to_send, false);
```
(often, not compressing is faster)

include/deal.II/base/mpi.h
```diff
@@ -1463,6 +1482,45 @@ namespace Utilities
...
+      AssertThrowMPI(ierr);
+
+      T received_object;
+
+      if (my_rank != root_process)
+        received_object = Utilities::unpack<T>(buffer, true);
+      else
+        received_object = object_to_send;
+
+      return received_object;
```
peterrum, line 1520:
```suggestion
      return Utilities::unpack<T>(buffer, false);
```

include/deal.II/base/mpi.h
```diff
@@ -1463,6 +1482,45 @@ namespace Utilities
...
+      (void)root_process;
+      return object_to_send;
+#  else
+      const auto n_procs = dealii::Utilities::MPI::n_mpi_processes(comm);
+      const auto my_rank = dealii::Utilities::MPI::this_mpi_process(comm);
+
+      AssertIndexRange(root_process, n_procs);
+
+      std::vector<char> buffer      = Utilities::pack(object_to_send, true);
+      int               buffer_size = buffer.size();
```
peterrum, line 1502:
```suggestion
      unsigned int               buffer_size = buffer.size();
```

include/deal.II/base/mpi.h
```diff
@@ -1463,6 +1482,45 @@ namespace Utilities
...
+
+      // Exchanging the size of buffer
+      int ierr = MPI_Bcast(&buffer_size, 1, MPI_INT, root_process, comm);
```
peterrum, line 1505:
```suggestion
      int ierr = MPI_Bcast(&buffer_size, 1, MPI_UNSIGNED, root_process, comm);
```

peterrum (commented):
Some minor comments.

May I also ask you to replace the following locs 
https://github.com/dealii/dealii/blob/f15a5d2cc86670ebec11f9891e5bfface3896d97/include/deal.II/base/mpi.templates.h#L596-L600
with a call to the new function.

Furthermore, could you add in `doc/new/changes/minor` a change-log entry. Thanks!

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
            const T &          object_to_send,
            const unsigned int root_process = 0);
 
+    /**
+     * A generalization of the classic MPI_Bcast function, that accepts
```
bangerth, line 1075:
```suggestion
     * A generalization of the classic `MPI_Bcast` function that accepts
```

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
...
+     * arbitrary data types T, as long as boost::serialize accepts T as an
```
bangerth, line 1076:
```suggestion
     * arbitrary data types `T`, as long as `boost::serialize` accepts `T` as an
```

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
...
+     * argument.
+     *
+     * @param[in] comm MPI communicator.
+     * @param[in] object_to_send an object to send to all processes
```
bangerth, line 1080:
```suggestion
     * @param[in] object_to_send An object to send to all processes.
```

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
...
+     * @param[in] root_process The process, which sends the object to all
```
bangerth, line 1081:
```suggestion
     * @param[in] root_process The process that sends the object to all
```

bergbauer:
I integrated your suggestions!

bangerth (commented):
Nice work!

You require two bcast calls, one for the size and one for the actual object. But I don't think you actually need this. Isn't there a way to wait for an incoming communication with an `MPI_Request` object (for all non-root processors), extract the size of the message, allocate the buffer, and then execute the bcast call on the receiving processes?

include/deal.II/base/mpi.h
```diff
@@ -1463,6 +1482,38 @@ namespace Utilities
...
+          const T &          object_to_send,
+          const unsigned int root_process)
+    {
+#  ifndef DEAL_II_WITH_MPI
+      (void)comm;
+      (void)root_process;
+      return object_to_send;
+#  else
+      const auto n_procs = dealii::Utilities::MPI::n_mpi_processes(comm);
+      const auto my_rank = dealii::Utilities::MPI::this_mpi_process(comm);
```
bangerth, line 1497: I don't think you actually need these variables, do you? But maybe it would be worthwhile adding an assertion that `0<=root_process<n_proces`.

peterrum:
Regarding @bangerth's comment

> You require two bcast calls, one for the size and one for the actual object. But I don't think you actually need this. Isn't there a way to wait for an incoming communication with an MPI_Request object (for all non-root processors), extract the size of the message, allocate the buffer, and then execute the bcast call on the receiving processes?

You can take a look at:
https://github.com/dealii/dealii/blob/e4f954195edd42d2c1d79cf4bf9f0158bdfff790/source/grid/tria_description.cc#L496-L501

bergbauer:
> Nice work!
> 
> You require two bcast calls, one for the size and one for the actual object. But I don't think you actually need this. Isn't there a way to wait for an incoming communication with an `MPI_Request` object (for all non-root processors), extract the size of the message, allocate the buffer, and then execute the bcast call on the receiving processes?

After studying the MPI documentation I don't think this works because there is no `mpi_tag` for `MPI_Bcast`. If you find a solution I would be happy to implement it!

peterrum (approved):

kronbichler:
> After studying the MPI documentation I don't think this works because there is no mpi_tag for MPI_Bcast. If you find a solution I would be happy to implement it!

I agree, I don't think there is a way to intercept the result of a broadcast as it is against the nature of the broadcast implementations: The MPI implementation needs several intermediate hops (a tree-like structure), and it would need to know the size of the arrays on each of these hops. In the grand scheme of things, it also looks like a premature optimization to request that to be done manually. Given how well-optimized the broadcast for a single integer is with very decent times on even a large supercomputer (around 1e-4 seconds), I can't see that being worth it.

kronbichler (approved):
Looks good, thank you @bergbauer !

Two small requests: Would you mind squashing your commits into one? And I think we should add a test here as well (even though there is now implicit coverage through the allreduce function). If you have not done that before, I can show you tomorrow.

doc/news/changes/minor/20210318MaximilianBergbauer
```diff
@@ -0,0 +1,3 @@
+New: Adds a new generalized MPI_Bcast function for arbitrary data types T.
```
marcfehling, line 1: Could you mention the actual name of the function here? This way doxygen will place a hyperlink to the function in the changelog.

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
...
+     * argument.
+     *
+     * @param[in] comm MPI communicator.
+     * @param[in] object_to_send An object to send to all processes.
+     * @param[in] root_process The process that sends the object to all
+     * processes. By default the process with rank 0 is the root process.
+     *
+     * @return Every process receives the object sent by the @p root_process
+     */
+
```
marcfehling, line 1086:
```suggestion
```

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
            const T &          object_to_send,
            const unsigned int root_process = 0);
 
+    /**
```
marcfehling, line 1074:
```suggestion
    /**
     * Sends an object @p object_to_send from the process @p root_process
     * to all other processes.
     *
```
I'd suggest to describe what this function actually does in the very first sentence.

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
...
+     * A generalization of the classic `MPI_Bcast` function that accepts
+     * arbitrary data types `T`, as long as `boost::serialize` accepts `T` as an
```
marcfehling, line 1076:
```suggestion
     * arbitrary data types `T`, as long as `Utilities::pack` accepts `T` as an
```
This is an implementation detail that we can hide

include/deal.II/base/mpi.h
```diff
@@ -1071,6 +1071,25 @@ namespace Utilities
...
+     * @param[in] object_to_send An object to send to all processes.
+     * @param[in] root_process The process that sends the object to all
+     * processes. By default the process with rank 0 is the root process.
+     *
+     * @return Every process receives the object sent by the @p root_process
+     */
+
+    template <typename T>
+    T
+    bcast(const MPI_Comm &   comm,
```
marcfehling, line 1089: I would be in favor to name this function without abbreviations, i.e., `broadcast`.
```suggestion
    broadcast(const MPI_Comm &   comm,
```

marcfehling (commented):
I would be in favor to name this function without abbreviations, i.e., `broadcast`. What do the others think about it?

include/deal.II/base/mpi.h
kronbichler: I don't think this detail should be omitted - for data types `T` not supported by `boost::serialize` (or for which we don't provide the appropriate interface in deal.II), the user needs to be in the position to add a `template<typename Archive> void serialize(Archive &, const unsigned int)` function to the class `T` to be wrapped. So I would suggest
```suggestion
     * arbitrary data types `T`, as long as Utilities::pack() (which in turn uses
     * `boost::serialize`, see in Utilities::pack() for details) accepts `T` as an
```
Or is there a simpler way to use it? If anything, I believe the serialization functionality is under-documented in deal.II with some non-obvious interface (do we discuss it in any of the tutorials, by the way?).

include/deal.II/base/mpi.h
bergbauer: I thought about this name as well so I changed it.

bergbauer: I thought about both options when naming the function, so democracy wins!

marcfehling (approved):

include/deal.II/base/mpi.h
marcfehling: We could also leave the choice to the user if they want compression with another parameter that defaults to false, but I think this would be overkill in this context.

include/deal.II/base/mpi.h
```diff
@@ -1463,6 +1485,36 @@ namespace Utilities
...
+              const T &          object_to_send,
+              const unsigned int root_process)
+    {
+#  ifndef DEAL_II_WITH_MPI
+      (void)comm;
+      (void)root_process;
+      return object_to_send;
+#  else
+      const auto n_procs = dealii::Utilities::MPI::n_mpi_processes(comm);
+      AssertIndexRange(root_process, n_procs);
```
marcfehling, line 1500: You need to either put this in `#ifdef DEBUG` or write `(void)n_procs;`, otherwise a warning about an unused variable gets triggered in `RELEASE` mode.

bergbauer: Thanks for the remark! I added `(void)n_procs;`

doc/news/changes/minor/20210318MaximilianBergbauer
```diff
@@ -0,0 +1,3 @@
+New: Adds the new generalized MPI_Bcast function Utilities::MPI:broadcast() for arbitrary data types T.
```
kronbichler, line 1:
```suggestion
New: Adds the new generalized MPI_Bcast function Utilities::MPI::broadcast() for arbitrary data types T.
```

kronbichler (approved):
Almost there, except for a typo. I like the test you added.

doc/news/changes/minor/20210318MaximilianBergbauer
bergbauer: I fixed the typo

peterrum:
@bangerth Feel free to merge if your comments have been addressed.

bangerth (approved):

bangerth:
Nice work, @bergbauer !

If you feel challenged, there are a number of places in the library where broadcast operations are used. You could replace those by the new call if you wanted -- it would certainly make the code easier to read!

PR #11937, opened 2021-03-18
Move `DoFHandler::prepare_c_and_r()` to `hp::Refinement::limit_p_level_difference()`.

marcfehling:
Since `DoFHandler::prepare_c_and_r()` does not need access to private member functions, and only makes sense in the hp-adaptation context, this PR will move the former member function to become the now free function `hp::Refinement::limit_plevel_difference()`.

PR #11943, opened 2021-03-19
Treat cells to be h-coarsened as their parent in `hp::Refinement::limit_p_level_difference()`.

marcfehling:
Follow-up to #11937.

Re-opening of #11929. See PR description for details.


include/deal.II/hp/refinement.h
```diff
@@ -677,6 +677,13 @@ namespace hp
      * to call this function, nor will it be automatically invoked in any part
      * of the library (contrary to its Triangulation counterpart).
      *
+     * On cells that will be h-coarsened, we enforce the difference criterion as
+     * if it's already a parent cell. That means, we set the level of all
```
peterrum, line 681:
```suggestion
     * if it is already a parent cell. That means, we set the level of all
```

source/hp/refinement.cc
```diff
@@ -843,6 +843,97 @@ namespace hp
...
+      // This function takes a cell and sets all its siblings to the highest
+      // p-level among them. Returns whether any future levels have been
+      // changed.
+      const auto prepare_level_for_parent = [&](const auto &neighbor) -> bool {
+        Assert(neighbor->is_active(), ExcInternalError());
+        if (neighbor->coarsen_flag_set() && neighbor->is_locally_owned())
+          {
+            const auto parent = neighbor->parent();
+
+            std::set<unsigned int> future_levels_children;
```
peterrum, line 897: Why not use here a vector?

peterrum (approved):

source/hp/refinement.cc
marcfehling: I access the last element of the `set` to determine the highest value.
I could also use `vector` and use `max_element` instead. I don't know what's more practical. Any opinions?

peterrum: Vector :)

peterrum (approved):

kronbichler (approved):

PR #11946, opened 2021-03-20
Use manifold iterator range in copy_triangulation().

marcfehling:


drwells (approved):

source/grid/tria.cc
```diff
@@ -10429,9 +10429,9 @@ Triangulation<dim, spacedim>::copy_triangulation(
     faces = std::make_unique<internal::TriangulationImplementation::TriaFaces>(
       *other_tria.faces);
 
-  auto bdry_iterator = other_tria.manifold.begin();
-  for (; bdry_iterator != other_tria.manifold.end(); ++bdry_iterator)
-    manifold[bdry_iterator->first] = bdry_iterator->second->clone();
+  for (const auto i : other_tria.get_manifold_ids())
```
bangerth, line 10432: `get_manifold_ids()` is a surprisingly expensive function. Couldn't you do this instead?
```suggestion
  for (const auto i : other_tria.manifold)
```

peterrum (approved):

PR #11715, opened 2021-02-09
MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence load forest directly

peterrum:
As discussed in https://github.com/dealii/dealii/pull/11707#pullrequestreview-585109577.

FYI and thanks: @marcfehling 

Depends on #11707

PR #11947, opened 2021-03-20
p:d:T:copy_triangulation() for refined meshes.

marcfehling:
Mentioned in #11668, `p:d:T` was only capable to being copied from another as long as it hasn't been refined. However, this restriction should be easily removed with the new function `p:d:T:load` (introduced in #11715) that takes a p4est object. This is an attempt.

@adamqc -- FYI
@peterrum -- We could use this updated function `p:d:T:copy_triangulation` in place of this [code section](https://github.com/dealii/dealii/blob/71ef4abc5da8c1f5cc04e229d022044f3259c755/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L1982-L1996). The `coarse_mesh_description` will be extracted for every triangulation instead just once, though. What do you think?


source/distributed/tria.cc
```diff
@@ -3267,64 +3265,59 @@ namespace parallel
...
+      // release unused vector memory because we will have very different
+      // vectors now
+      if (dynamic_cast<const dealii::parallel::TriangulationBase<dim, spacedim>
+                         *>(&other_tria))
         {
-          // the underlying triangulation should not be checking for distorted
-          // cells
-          Assert(false, ExcInternalError());
+          ::dealii::internal::GrowingVectorMemoryImplementation::
+            release_all_unused_memory();
```
marcfehling, line 3319: This call was present in `parallel::TriangulationBase::copy_triangulation`. I don't know if it's necessary in this context, but it probably won't hurt anyways.

marcfehling:
Maybe it would make things easier to simply extend the `copy_triangulation` function by calling `internal::p4est::functions::copy_forest()` instead of `load()`. I'll have a look tomorrow.

PR #11948, opened 2021-03-20
p:d:T:copy_triangulation() for refined meshes. Second attempt.

marcfehling:
Alternative to #11947.

Instead of creating a coarse triangulation and reconstructing the refinement history, we can simply copy all data structures in `p:d:Triangulation`. This should be faster, especially in `MGTransferGlobalCoarsening`.


peterrum (commented):
Looks good. Thanks! Could you add a test and run locally `ctest -R multigrid-global-coarsening`

marcfehling:
The `multigrid-global-coarsening` tests have failed prior to your suggestion. Now they pass!

tests/mpi/tria_copy_triangulation_02.cc
```diff
@@ -0,0 +1,130 @@
...
+
+template <int dim>
+void
+test()
+{
+  unsigned int myid = Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
+
+  parallel::distributed::Triangulation<dim> tria1(MPI_COMM_WORLD);
+  {
+    GridGenerator::hyper_cube(tria1);
```
peterrum, line 36: Could you use here a geometry with a manifold attached to? Like hyper_ball? And maybe also set the boundary ids?

doc/news/changes/minor/20210319Fehling
```diff
@@ -0,0 +1,5 @@
+Improved: Function parallel::distributed::Triangulation::copy_triangulation()
```
peterrum, line 1:
```suggestion
Improved: The function parallel::distributed::Triangulation::copy_triangulation()
```

peterrum (approved):

tests/mpi/tria_copy_triangulation_02.cc
marcfehling: I used `hyper_shell` with the `colorize` option and added a few assertions. Looks good to me now!

peterrum (approved):
Thank you!

kronbichler (approved):

PR #11951, opened 2021-03-22
MF::check_vector_compatibility(): improve assert message

peterrum:
FYI @nmuch @mschreter

### DIFF
--- a/include/deal.II/matrix_free/vector_access_internal.h
+++ b/include/deal.II/matrix_free/vector_access_internal.h
@@ -201,8 +201,11 @@ namespace internal
            ExcMessage(
              "The parallel layout of the given vector is not "
              "compatible with the parallel partitioning in MatrixFree. "
-             "Use MatrixFree::initialize_dof_vector to get a "
-             "compatible vector."));
+             "A potential reason is that you did not use "
+             "MatrixFree::initialize_dof_vector() to get a "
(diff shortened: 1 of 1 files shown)


include/deal.II/matrix_free/vector_access_internal.h
```diff
@@ -201,8 +201,11 @@ namespace internal
            ExcMessage(
              "The parallel layout of the given vector is not "
              "compatible with the parallel partitioning in MatrixFree. "
-             "Use MatrixFree::initialize_dof_vector to get a "
-             "compatible vector."));
+             "A potential reason is that you did not use "
+             "MatrixFree::initialize_dof_vector() to get a "
+             "compatible vector. If you did, the dof_handler_index "
+             "used there and passed to the constructor of "
```
kronbichler, line 207:
```suggestion
             "used there and the one passed to the constructor of "
```

kronbichler (approved):
If it could help, we might also be able to give the size of the vector and the number of ghosts as hints. But I think the error message is clearly improved already.

include/deal.II/matrix_free/vector_access_internal.h
peterrum: I think this should be enough!

kronbichler (approved):

PR #11957, opened 2021-03-24
LinearAlgebra::ReadWriteVector::import() add fallback function for unsupported vectors

peterrum:
Needed by #11594.

### DIFF
--- a/include/deal.II/lac/read_write_vector.h
+++ b/include/deal.II/lac/read_write_vector.h
@@ -411,6 +411,16 @@ namespace LinearAlgebra
 #endif
 
     /**
+     * Fall-back function for vectors that are not-supported by this class.
+     */
+    template <typename VectorType>
+    void
+    import(const VectorType &      vec,
+           VectorOperation::values operation,
+           const std::shared_ptr<const Utilities::MPI::CommunicationPatternBase>
+             &communication_pattern = {});
+
+    /**
      * The value returned by this function denotes the dimension of the vector
      * spaces that are modeled by objects of this kind. However, objects of
      * the current class do not actually stores all elements of vectors of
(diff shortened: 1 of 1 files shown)


kronbichler (approved):

kronbichler:
```
 /home/runner/work/dealii/dealii/source/lac/read_write_vector.cu:24:15: error: ambiguous template specialization ‘import<>’ for ‘void dealii::LinearAlgebra::ReadWriteVector<float>::import(const dealii::LinearAlgebra::CUDAWrappers::Vector<float>&, dealii::VectorOperation::values, const std::shared_ptr<const dealii::Utilities::MPI::CommunicationPatternBase>&)’
```

peterrum:
Could someone help me to fix the compilation issue for CUDA? I don't have a clue how to address this. Thx!

peterrum:
Let's close this for now - as long as #11594 is not finalized.

peterrum:
The following failing tests are related to block vectors:
```
The following tests FAILED:
	 68 - simplex/step-20.debug (Failed)
	 70 - simplex/step-31.debug (Failed)
	 72 - simplex/step-55.mpirun=2.debug (Failed)
```
First, there is no option to import from a block vector to a `LinearAlgebra::distributed::Vector`; second the locally-owned indexset of the block vector might be not contiguous... which is sub-optimal for `LinearAlgebra::distributed::Vector`. Are there any ideas how to deal with these issues?

bangerth:
By the way, the error results from the fact that after this patch, you have two functions
```
    void
    import(const CUDAWrappers::Vector<Number> &cuda_vec,
           VectorOperation::values             operation,
           const std::shared_ptr<const Utilities::MPI::CommunicationPatternBase>
             &communication_pattern = {});

    void
    import(const VectorType &      vec,
           VectorOperation::values operation,
           const std::shared_ptr<const Utilities::MPI::CommunicationPatternBase>
             &communication_pattern = {});
```
and when you explicitly instantiate these in
```
  template void
  ReadWriteVector<float>::import(
    const CUDAWrappers::Vector<float> &,
    VectorOperation::values,
    const std::shared_ptr<const Utilities::MPI::CommunicationPatternBase> &);
```
the compiler tells you that it doesn't know which of these two functions you mean because they both match the requested signature.

I don't actually know how to work around this. You can select the second (new) function if you wrote
```
  template void
  ReadWriteVector<float>::import<CUDAWrappers::Vector<float>>(       // note explicit statement of member template argument
    const CUDAWrappers::Vector<float> &,
    VectorOperation::values,
    const std::shared_ptr<const Utilities::MPI::CommunicationPatternBase> &);
```
but I don't know how you would select the first of the members.

PR #11958, opened 2021-03-24
MatrixFreeTools::compute_matrix() for PETScWrappers::MPI::SparseMatrix

peterrum:


include/deal.II/matrix_free/tools.h
```diff
@@ -846,14 +846,25 @@ namespace MatrixFreeTools
...
 
+    const auto &locally_owned_dofs =
+      (matrix_free.get_mg_level() == numbers::invalid_unsigned_int) ?
+        matrix_free.get_dof_handler(dof_no).locally_owned_dofs() :
+        matrix_free.get_dof_handler(dof_no).locally_owned_mg_dofs(
+          matrix_free.get_mg_level());
+
+    for (const auto i : locally_owned_dofs)
+      if (constraints.is_constrained(i))
+        matrix.add(i, i, +1.0);
```
kronbichler, line 857: Why is this necessary? There is a version of `AffineConstraints::distribute_local_to_global` taking only a single slot `dof_indices` that sets the diagonal, rather than the version used in line 839 above, https://github.com/dealii/dealii/blob/a551e1a2594ee65abbadd2e590ef1b1055b43b95/include/deal.II/lac/affine_constraints.h#L1364-L1384. For a square matrix, the simpler version would be preferable. At least with the current code, we would not be easily able to also assemble non-square matrices (matrices from different `DoFHandler` objects or at least some sub-set of unknowns). By the way, I think in line 829 we are missing an argument `dof_no` for `get_cell_iterator` for the case not the `DoFHandler` with index 0 is desired.

peterrum: I am made the change(s). I have verified that the change is working in some local multigrid tests.

kronbichler (approved):

PR #11965, opened 2021-03-25
FETools::Compositing::compute_nonzero_components(): add assert

### LINKED ISSUE #11964: FESystem segfaults if no parameter is given to the constructor
The following program fails with a segmentation fault:
```cpp
#include <deal.II/fe/fe_system.h>

using namespace dealii;

int
main()
{
  FESystem<2> fe;
}
```

I will take a look later today where we could add a useful assert.

FYI @JPmoep

peterrum:
Fixes #11964.

### DIFF
--- a/include/deal.II/fe/fe_tools.templates.h
+++ b/include/deal.II/fe/fe_tools.templates.h
@@ -375,7 +375,10 @@ namespace FETools
       const bool do_tensor_product)
     {
       AssertDimension(fes.size(), multiplicities.size());
-
+      Assert(
+        fes.size() > 0,
+        ExcMessage(
+          "This function only makes sense if at least one FiniteElement is provided."));
       // first count the number of dofs and components that will emerge from the
       // given FEs
       unsigned int n_shape_functions = 0;


peterrum:
> Should we allow a default constructor at all?

@masterleinad I like the idea!

include/deal.II/fe/fe_tools.templates.h
```diff
@@ -421,106 +421,110 @@ namespace FETools
...
+                                     .size(),
+                               ExcInternalError());
+                        retval[total_index][comp_start + c] =
+                          fes[base]->get_nonzero_components(index_in_base)[c];
+                      }
+                  }
+          }
 
       // 4. Hexes
       if (dim == 3)
```
marcfehling, line 530: Why don't you have to check for `if (fes.size() > 0)` here as well?

peterrum: Here it is not needed, since `fes.front()` is not accessed.

marcfehling: I see. `fes.size()` limits the loop, so the whole hexes-part will be skipped.

You could also put parts 1-4 into one large conditional statement, which will also reflect that we really skip everything at first glance. But that's just a suggestion - I don't insist on this :-)

marcfehling: Or, since the default constructor is deleted now, place an assertion that `fes.size() > 0` and remove the conditional.

peterrum: The function takes a vector of FEs; so the default constructor of FESystem is somewhat unrelated (and just pointed out the missing check in this function). I am not sure if the function should work for an empty vector. But in the doubt, I would like to keep it in such a way that it can work under such condition.

kronbichler (approved):

marcfehling (approved):

include/deal.II/fe/fe_tools.templates.h
```diff
@@ -421,106 +421,110 @@ namespace FETools
...
-                      Assert(c < fes[base]
-                                   ->get_nonzero_components(index_in_base)
-                                   .size(),
-                             ExcInternalError());
-                      retval[total_index][comp_start + c] =
-                        fes[base]->get_nonzero_components(index_in_base)[c];
-                    }
-                }
-        }
+      if (fes.size() > 0)
```
bangerth, line 424: Does this function even make sense if `fes.size()==0`? If not, I would prefer having an assertion at the top.

include/deal.II/fe/fe_system.h
```diff
@@ -210,12 +210,13 @@ class FE_Enriched;
  *
  * @ingroup febase fe vector_valued
  *
- * Hartmann 2001.
  */
 template <int dim, int spacedim = dim>
 class FESystem : public FiniteElement<dim, spacedim>
 {
 public:
+  FESystem() = delete;
```
bangerth, line 218: Can you document this function?

bangerth: Ping?

bangerth (changes requested):

include/deal.II/fe/fe_tools.templates.h
```diff
@@ -375,7 +375,10 @@ namespace FETools
       const bool do_tensor_product)
     {
       AssertDimension(fes.size(), multiplicities.size());
-
+      Assert(
+        fes.size() > 0,
+        ExcMessage(
+          "This function only has sense if at least one FiniteElement is provided."));
```
marcfehling, line 381:
```suggestion
          "This function only makes sense if at least one FiniteElement is provided."));
```

bangerth: Ping?

peterrum: Done :)

bangerth (approved):

PR #11966, opened 2021-03-26
Moved `dominated_future_fe_on_children` from class member to internal namespace.

### LINKED ISSUE #11902: Test sharedtria/hp_choose_p_over_h.mpirun=2.debug fails
Due to #11868, `sharedtria/hp_choose_p_over_h.mpirun=2.debug` fails. The reason is that in:
https://github.com/dealii/dealii/blob/30e53a956acf8bce3c2a0b70f15e4fae81a148e8/source/hp/refinement.cc#L660-L680
`GridTools::exchange_cell_data_to_ghosts()` is called even for `p:s:T` without artificial cells. However, `GridTools::exchange_cell_data_to_ghosts()` might deadlock for `p:s:T`, which was the reason for introducing the assert, and this test passed only by chance. 

We need a fall-back solution here. I am not sure what the right approach is. But I could image that we switch for more data structures in `DoFHandler` to `LA::d::V` (which would make such "ghost updates" potentially easier).

marcfehling:
Preparations for #11902.

The function `dominated_future_fe_on_children()` is only used in the hp-context and too specific to be part of the `DoFCellAccessor` class. This PR moves it into the `internal::hp::DoFHandlerImplementation` namespace. I think this is the right place for this kind of function, but I am open for any suggestions in this regard.

Since we need this function in other parts of the library (especially in the `parallel::distributed` namespace), it needs to be properly declared in a header file.

Function has been introduced since the last release in #11011.


include/deal.II/dofs/dof_handler.h
```diff
@@ -1805,6 +1845,9 @@ class DoFHandler : public Subscriptor
   // in the functions that set and retrieve vertex dof indices.
   friend struct dealii::internal::DoFHandlerImplementation::Implementation;
   friend struct dealii::internal::hp::DoFHandlerImplementation::Implementation;
+  friend unsigned int
+  dealii::internal::hp::DoFHandlerImplementation::
+    dominated_future_fe_on_children<dim, spacedim>(const cell_iterator &);
```
drwells, line 1850: Would it work if we added this as another static function inside `hp::DoFHandlerImplementation::Implementation` so that we don't need another `friend` declaration?

drwells (commented):
I have one suggestion that might not be possible so I'm marking this as 'Comment'.

Nice work deleting the old changelog.

include/deal.II/dofs/dof_handler.h
marcfehling: We not necessarily need access to private members for the function in the current state, but I most certainly will when preparing the fix I have in mind for #11902.

Hmm. Would it work if I move this particular function into the `Implementation` struct, and then basically call a wrapper that is part of the public interface? I'll give it a try!

drwells (approved):

drwells:
A non-friend non-member function is best!

PR #11967, opened 2021-03-26
Cleanup doc in DoFHandler.

marcfehling:
I bet those are relics from #10328.

### DIFF
--- a/include/deal.II/dofs/dof_handler.h
+++ b/include/deal.II/dofs/dof_handler.h
@@ -535,7 +535,7 @@ public:
 
   /**
    * Invalid active FE index which will be used as a default value to determine
-   *  whether a future FE index has been set or not.
+   * whether a future FE index has been set or not.
    */
   static const active_fe_index_type invalid_active_fe_index =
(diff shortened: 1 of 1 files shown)


source/dofs/dof_handler.cc
```diff
@@ -39,10 +39,6 @@
 
 DEAL_II_NAMESPACE_OPEN
 
-template <int dim, int spacedim>
-const typename DoFHandler<dim, spacedim>::active_fe_index_type
-  DoFHandler<dim, spacedim>::invalid_active_fe_index;
```
bangerth, line 44: You need to keep this. The variable is `static const`, but the compiler still needs to allocate a memory location for it.

bangerth (commented):
Just the one issue. The rest is ok.

source/dofs/dof_handler.cc
bangerth: Ah, and indeed the testers fail for exactly this reason.

bangerth (approved):

PR #11968, opened 2021-03-26
Replace AffineConstraints::distribute() in MGTwoLevelTransfer

peterrum:


peterrum:
FYI @lpsaavedra

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1979,8 +2022,32 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::prolongate(
...
+
+        Number temp = constraint_coarse.is_inhomogeneously_constrained(i) ?
+                        constraint_coarse.get_inhomogeneity(i) :
+                        0.0;
+
+        const auto constraints = constraint_coarse.get_constraint_entries(i);
+
+        if (constraints)
+          for (const auto p : *constraints)
+            temp += vec_coarse_constraints[p.first] * p.second;
```
kronbichler, line 2045: This will be slow in comparison to other operations we do because this involves an index transfer from global to MPI-local indices inside `operator[]`, but the function can be expected to be much more efficient than the old one that had to construct the indices on the fly.

kronbichler (approved):

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
peterrum: I would be happy to pre-compute everything and remote the `AffineConstrains` completely. But at the moment, I am using it also at two other places:

https://github.com/dealii/dealii/blob/eb3ad2401ebf6560893f47536f50ea217865ae2e/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L2143-L2148

https://github.com/dealii/dealii/blob/eb3ad2401ebf6560893f47536f50ea217865ae2e/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L2227-L2230

I am not sure how to eliminate these.

kronbichler: I agree, and would add that this points to the more general inefficiency we have for MPI inside deal.II. In [ryujin](https://github.com/conservation-laws/ryujin) @tamiko and I also had to copy all data structures again into MPI-local numbering, because the translation can sink performance. Now the function here is not as bad, because it only happens for the constrained unknowns, and only once per grid traversal.

What might fit here is to set up an `AffineConstraints` object in the local indices, using the union of the locally relevant rows and the information you use from the columns for the ghosted indices. And then we do all other stuff with a `dealii::Vector` (non-distributed) that should have fast access. We would then need the copy between `LA:d:Vector` and `Vector` instead of between two `LA:d:Vector` classes. I'm not sure how much work it would be.

peterrum: > In ryujin @tamiko and I also had to copy all data structures again into MPI-local numbering,

Yeah. I remember that.

I am not sure if it wouldn't be easier, instead of setting up a new `AffineConstraint` object plus a new vector, to to extract the relevant information from the `AffineConstraint` object here into a couple of arrays. 

But anyway, not relevant for this PR.

kronbichler: Probably yes, we might even want to create a struct and appropriate function in `AffineConstraints` for it.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -948,6 +948,49 @@ namespace internal
...
+
+      for (const auto i : locally_owned_dofs)
+        {
+          if (constraint_coarse.is_constrained(i) == false)
+            continue;
+
+          const auto constraints = constraint_coarse.get_constraint_entries(i);
+
+          if (constraints)
+            for (const auto p : *constraints)
```
kronbichler, line 973:
```suggestion
            for (const auto &p : *constraints)
```

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -1979,8 +2022,32 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::prolongate(
...
+          continue;
+
+        Number temp = constraint_coarse.is_inhomogeneously_constrained(i) ?
+                        constraint_coarse.get_inhomogeneity(i) :
+                        0.0;
+
+        const auto constraints = constraint_coarse.get_constraint_entries(i);
+
+        if (constraints)
+          for (const auto p : *constraints)
```
kronbichler, line 2044:
```suggestion
          for (const auto &p : *constraints)
```

kronbichler (commented):
I believe you need these changes to make the CI happy.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -948,6 +948,49 @@ namespace internal
...
+      const MeshType &                         dof_handler_coarse,
+      const dealii::AffineConstraints<Number> &constraint_coarse,
+      MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>
+        &transfer)
+    {
+      transfer.constraint_coarse.copy_from(constraint_coarse);
+
+      std::vector<types::global_dof_index> dependencies;
+
+      const auto locally_owned_dofs = dof_handler_coarse.locally_owned_dofs();
```
kronbichler, line 963: clang-tidy wants to change this one (unnecessary copy):
```suggestion
      const auto &locally_owned_dofs = dof_handler_coarse.locally_owned_dofs();
```

kronbichler (approved):

PR #11969, opened 2021-03-26
New Triangulation signal `post_p4est_refinement`.

marcfehling:
Follow-up to #11888.

This PR introduces a new signal `post_p4est_refinement` that gets triggered after the oracle has been refined, but before any changes to the Triangulation have been made or the oracle has been repartitioned. This is necessary as p4est might perform refinement slightly different than intended via refine and coarsen flags of the Triangulation, see #8647.

In this PR, we only connect preparations for the transfer of active FE indices to it and move it from the `pre_distributed_refinement` signal. As the `parallel::fullydistributed::Triangulation` does not use any boost signals by now in the context of refinement and partitioning, we no longer connect to these particular signals anymore for this class.

Some hp-functionalities might connect to this signal, i.e., for error prediction and limiting p-level differences across the grid. I will deal with this in follow-up PRs.

### DIFF
--- a/source/dofs/dof_handler.cc
+++ b/source/dofs/dof_handler.cc
@@ -20,6 +20,7 @@
 #include <deal.II/base/mpi.templates.h>
 
 #include <deal.II/distributed/cell_data_transfer.templates.h>
+#include <deal.II/distributed/fully_distributed_tria.h>
 #include <deal.II/distributed/shared_tria.h>
 #include <deal.II/distributed/tria.h>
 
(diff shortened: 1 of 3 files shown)


source/dofs/dof_handler.cc
```diff
@@ -3067,7 +3074,7 @@ DoFHandler<dim, spacedim>::connect_to_triangulation_signals()
 
       // refinement signals
       this->tria_listeners_for_transfer.push_back(
-        this->tria->signals.pre_distributed_refinement.connect(
+        this->tria->signals.post_p4est_refinement.connect(
```
Rombur, line 3077: I don't understand why we need a new signal? It's not clear to me what's the difference between `pre_distributed_refinement` and `post_p4est_refinement`

marcfehling: Long story short: p4est refines differently than Triangulation. We have functions that rely on correctly set refine flags, and the outcome of these functions affect partitioning. So we need this additional signal after refining p4est, but before repartitioning.

Maybe it gets clearer in context of the follow-up PR #11981.

Rombur: Yes, I understand that but you are not adding a signal here, you are simply replacing the signal by a new one and the documentation of these two signals is almost identical. Why can't you use the current `pre_distributed_refinement`?

marcfehling: The connected function `pre_distributed_transfer_action()` stores all future FE indices in a container and prepares the transfer of them across mesh refinement. Right now, we could attach this signal to either `pre_distributed_refinement` or `post_p4est_refinement` and the outcome would be the same.

However, if we want to manipulate these future FE indices based on the way p4est manages mesh refinement (see #11981), we need to attach this function to the `post_p4est_refinement` signal.

I replaced the signal in this PR already to make sure that the testsuite succeeds with this change.

Rombur (approved):

source/dofs/dof_handler.cc
marcfehling: Sorry for the confusion :)

PR #11970, opened 2021-03-26
Fix references.bib

jppelteret:
The period is just on the wrong side of the brace.

### DIFF
--- a/doc/doxygen/references.bib
+++ b/doc/doxygen/references.bib
@@ -873,7 +873,7 @@ year = {2008},
   pages = {337--360},
   author = {B{\'{e}}atrice Rivi{\`{e}}re and Mary F. Wheeler and Vivette Girault},
   journal = {Computational Geosciences},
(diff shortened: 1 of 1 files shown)


drwells (approved):

drwells:
Nice catch. I'm surprised we don't get a doxygen warning from this.

marcfehling (approved):

jppelteret:
The only reason that I spotted this is that this half of this file was rendered in colour by my IDE and the other was grey. Took me a few moments to spot the culprit!

marcfehling:
Do we need that period at all? Doesn't seem to be part of the official title: https://dx.doi.org/10.1023/A:1011591328604

marcfehling:
The extra period will probably manifest as an extra period in the current bibliography, so we would have two periods here https://www.dealii.org/developer/doxygen/deal.II/citelist.html#CITEREF_Riviere1999

I would suggest to get rid of it.

jppelteret:
@marcfehling, thanks for thinking about what the output formatting would look like. You're right that we should get rid of it (it's also not in the official publication title, anyway). I've done this now.

marcfehling (approved):

PR #11971, opened 2021-03-26
Give the option to use step-62 without the jupyter notebook

dangars:
This allows the user to use step-62 without the jupyter notebook.

This PR depends on #11869 because I need `set_attribute<bool>`


dangars:
Now that #11869 has been merged this PR is ready for review :)

drwells:
This seems more complicated than it needs to be. Wouldn't it be easier to just write two files - one that should be passed to the notebook and one that should not?

drwells:
(sorry, I hit the wrong button!)

examples/step-62/step-62.cc
```diff
@@ -1371,10 +1374,100 @@ int main(int argc, char *argv[])
...
+      auto       data = data_file.open_group("data");
+#else
+      // If we don't use the jupyter notebook we have to create the groups and
+      // set the parameters.
+      HDF5::File data_file("results.h5",
+                           HDF5::File::FileAccessMode::create,
+                           MPI_COMM_WORLD);
+      auto       data = data_file.create_group("data");
+
+      std::vector<std::string> group_names = {"displacement", "calibration"};
```
bangerth, line 1390: make it `const`

examples/step-62/step-62.cc
```diff
@@ -1371,10 +1374,100 @@ int main(int argc, char *argv[])
...
+          // Create the group
+          auto group = data.create_group(group_name);
+
+          // Dimensions of the domain
+          // The waveguide length is equal to dimension_x
+          group.set_attribute<double>("dimension_x", 2e-5);
+          // The waveguide width is equal to dimension_y
+          group.set_attribute<double>("dimension_y", 2e-8);
+          // Position of the probe that we use to measure the flux
+          group.set_attribute<double>("probe_pos_x", 8e-6);
```
bangerth, line 1402: This is more of a style suggestion. It turns out that the tutorials become difficult to read if you alternate between commentary and code too often. It's easier to read if you keep more of the code and more of the documentation together. For example, at the top (line 1393) you could do something like this:
```
// For each of these two group names, we now open the group and put attributes into these groups.
// Specifically, these are:
// - The dimensions of the waveguide (in $x$ and $y$ directions)
// - The position of the probe (in $x$ and $y$ directions)
...
```
Then you can just show the whole code block at once.

bangerth (commented):
Just a couple of suggestions. I agree with @drwells that it would be easier to just always write the file. I'm not fond of the `#ifdef` at the top.

dangars:
How about we use the jupyter notebook only for the analysis of the results? I placed the original python code that sets the parameters in the extensions section.

Now the parameters are set in the C++ file, there are no `#ifdefs`, and you can run step-62 without a jupyter notebook.

The problem setting the parameters in the jupyter notebook is that you need a jupyter notebook to run the tutorial and each time you want to run the tutorial you have to restart the jupyter notebook.

drwells (approved):

drwells:
This looks good to me (and it addresses all of Wolfgang's comments).

dangars:
Great :)

bangerth (approved):

PR #11973, opened 2021-03-27
Remove AffineConstraint from MGTwoLevelTransfer

peterrum:
As discussed in https://github.com/dealii/dealii/pull/11968#discussion_r602225073.

### DIFF
--- a/include/deal.II/multigrid/mg_transfer_global_coarsening.h
+++ b/include/deal.II/multigrid/mg_transfer_global_coarsening.h
@@ -310,14 +310,48 @@ private:
   mutable LinearAlgebra::distributed::Vector<Number> vec_coarse;
 
   /**
-   * Constraint matrix on coarse level.
-   */
-  AffineConstraints<Number> constraint_coarse;
-
-  /**
-   * Internal vector that performs constraint_coarse.distribute().
+   * Internal vector for performing manual constraint_coarse.distribute(), which
+   * is needed for acceptable performance.
    */
   mutable LinearAlgebra::distributed::Vector<Number> vec_coarse_constraints;
+
+  /**
+   * Constraint-entry indices for manually performing
+   * constraint_coarse.distribute() in MPI-local indices (for performance
+   * reasons).
+   */
+  std::vector<unsigned int> constraint_coarse_distribute_indices;
+
+  /**
+   * Constraint-entry values for manually performing
+   * constraint_coarse.distribute() in MPI-local indices (for performance
+   * reasons).
+   */
+  std::vector<Number> constraint_coarse_distribute_values;
+
+  /**
+   * Pointers to the constraint entries for performing manual
+   * constraint_coarse.distribute().
+   */
+  std::vector<unsigned int> constraint_coarse_distribute_ptr;
+
+  /**
+   * Constraint-entry indices for performing manual
+   * constraint_coarse.distribute_local_to_global().
+   */
+  std::vector<unsigned int> distribute_local_to_global_indices;
+
+  /**
+   * Constraint-entry values for performing manual
+   * constraint_coarse.distribute_local_to_global().
+   */
+  std::vector<Number> distribute_local_to_global_values;
+
+  /**
+   * Pointers to the constraint entries for performing manual
+   * constraint_coarse.distribute_local_to_global().
+   */
+  std::vector<unsigned int> distribute_local_to_global_ptr;
 
   /**
    * Number of components.
--- a/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
+++ b/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
@@ -956,8 +956,6 @@ namespace internal
       MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>
         &transfer)
     {
-      transfer.constraint_coarse.copy_from(constraint_coarse);
-
       std::vector<types::global_dof_index> dependencies;
 
       const auto &locally_owned_dofs = dof_handler_coarse.locally_owned_dofs();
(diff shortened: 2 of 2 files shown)


include/deal.II/multigrid/mg_transfer_global_coarsening.h
```diff
@@ -310,14 +310,34 @@ class MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>
...
   /**
-   * Constraint matrix on coarse level.
+   * Internal vector for performing manual constraint_coarse.distribute().
    */
-  AffineConstraints<Number> constraint_coarse;
+  mutable LinearAlgebra::distributed::Vector<Number> vec_coarse_constraints;
 
   /**
-   * Internal vector that performs constraint_coarse.distribute().
+   * Constraint entries for performing manual constraint_coarse.distribute().
```
kronbichler, line 318:
```suggestion
   * Constraint entries for manually performing constraint_coarse.distribute() in
   * MPI-local indices (for performance reasons).
```

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -989,6 +987,71 @@ namespace internal
...
+            }
+
+          transfer.constraint_coarse_distribute_ptr.push_back(
+            transfer.constraint_coarse_distribute_entries.size());
+        }
+    }
+
+    template <int dim, typename Number>
+    static void
+    precompute__restriction_constraints(
```
kronbichler, line 1017:
```suggestion
    precompute_restriction_constraints(
```

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -989,6 +987,71 @@ namespace internal
...
+      const std::shared_ptr<const Utilities::MPI::Partitioner> &partitioner,
+      const dealii::AffineConstraints<Number> &constraint_coarse,
+      MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>
+        &transfer)
+    {
+      transfer.distribute_local_to_global_entries.clear();
+      transfer.distribute_local_to_global_ptr = {0};
+
+      const auto fu = [&](const auto &is) {
```
kronbichler, line 1026:
```suggestion
      const auto fu = [&](const auto &indices) {
```

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -2029,23 +2098,24 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::prolongate(
...
-
-        Number temp = constraint_coarse.is_inhomogeneously_constrained(i) ?
-                        constraint_coarse.get_inhomogeneity(i) :
-                        0.0;
-
-        const auto constraints = constraint_coarse.get_constraint_entries(i);
+    for (unsigned int i = 0; i < constraint_coarse_distribute_ptr.size() - 1;
+         ++i)
+      if ((constraint_coarse_distribute_ptr[i + 1] -
+           constraint_coarse_distribute_ptr[i]) == 0)
```
kronbichler, line 2104: I guess the compile will do it automatically, but still I would prefer
```suggestion
      if (constraint_coarse_distribute_ptr[i + 1] ==
           constraint_coarse_distribute_ptr[i])
```

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -2029,23 +2098,24 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::prolongate(
...
+        vec_coarse.local_element(i) = vec_coarse_constraints.local_element(i);
```
kronbichler, line 2105: When you do this, do you still need this line https://github.com/dealii/dealii/blob/11fcb45cb7797a06dc4fb33a40d43296be49f35d/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L2024 (a few lines above)?

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -2191,6 +2261,26 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::
   AlignedVector<VectorizedArrayType> evaluation_data_fine;
   AlignedVector<VectorizedArrayType> evaluation_data_coarse;
 
+  // a helper function similar to AffineConstraints::distribute_local_to_global
+  // but working with local indices
+  const auto distribute_local_to_global =
+    [&](const auto &index, const auto &value, auto &global_vector) {
+      if ((distribute_local_to_global_ptr[index + 1] -
+           distribute_local_to_global_ptr[index]) == 0)
```
kronbichler, line 2269:
```suggestion
      if (distribute_local_to_global_ptr[index + 1] ==
           distribute_local_to_global_ptr[index])
```

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -2191,6 +2261,26 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::
...
+        global_vector.local_element(index) += value;
+      else if (((distribute_local_to_global_ptr[index + 1] -
+                 distribute_local_to_global_ptr[index]) == 1 &&
+                distribute_local_to_global_entries
+                    [distribute_local_to_global_ptr[index]]
+                      .first == numbers::invalid_unsigned_int) == false)
```
kronbichler, line 2275: Can you explain what the `numbers::invalid_unsigned_int` case indicates? I kind of thought I knew it in the code above, but now I'm lost.

kronbichler (commented):
Looks really good. Maybe these CSR-like structures would be something for the regular `AffineConstraints` object at some point in the future as well?

include/deal.II/multigrid/mg_transfer_global_coarsening.h
```diff
@@ -310,14 +310,34 @@ class MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>
...
    */
-  mutable LinearAlgebra::distributed::Vector<Number> vec_coarse_constraints;
+  std::vector<std::pair<unsigned int, Number>>
+    constraint_coarse_distribute_entries;
```
kronbichler, line 321: `AffineConstraints` aside for a moment, I think the canonical way of storing these is to use two separate arrays for the index and the value; at least with `unsigned int` and `double` numbers, you would spend 16 bytes per entry despite only 12 bytes of actual data. However, the most interesting use of this class is probably `float`, so it would fit from that perspective, so I'm leaning towards choosing this version. What do you think?

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
peterrum: You are right!

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
peterrum: I am marking this way homogeneous constraints. I have added some comments.

peterrum:
@kronbichler Thanks for the review! I have addressed the comments and split up the vector of pairs.

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
kronbichler: So in case we would ever support inhomogeneous constraints here, we would use the value entry for the inhomogeneity, right?

peterrum: Yes.

kronbichler (approved):

PR #11977, opened 2021-03-29
Small changes in FineDoFHandlerView

peterrum:
- fix MPI type
- fix typo
- introduce a new tag

PR #11979, opened 2021-03-29
Add some more AssertThrowMPIs.

drwells:
After reading #11977 I went through the library to see what other `AssertThrowMPI`s were missing and added a bunch.


masterleinad (approved):
Looks good to me.

bangerth:
Nice work, @drwells, like you always do when you decide to clean something up!

PR #11983, opened 2021-03-30
Reset policy during Triangulation::load()

peterrum:
The policy has not been set during `Triangulation::load()`.

FYI @agrayver


agrayver:
Thanks! Hopefully this will be merged soon as all my codes are broken on master now :(

marcfehling (approved):

PR #11984, opened 2021-03-30
Avoid quadratic complexity in number of coarse grid cells

kronbichler:
The first commit is the interesting one - I saw very slow work with the grid manipulation operations on a mesh with many cells on the coarse level, and indeed we did and `trianuglation.end(0)` which is of linear complexity, so when used as the termination criterion of a triangulation this ends up being quadratic costs.

While there, I converted a few loops into range-based for loops.


masterleinad (approved):
Looks good to me!

peterrum (approved):

PR #11988, opened 2021-03-31
Remove assertion from `DoFHandler::get_fe_collection()`.

marcfehling:
We check for empty FE collections at `DoFHandler::get_fe_collection()`. This check is not mandatory as `FECollection` objects are allowed to be empty. Futhermore, other functions can not verify whether a `DoFHandler` has already been initialized by checking the size of the `FECollection` object.

This PR removes the assertion, and lets functions in the `hp::Refinement` namespace quickly determine if they need to do anything at all.

### DIFF
--- a/source/hp/refinement.cc
+++ b/source/hp/refinement.cc
@@ -45,6 +45,10 @@ namespace hp
     void
     full_p_adaptivity(const dealii::DoFHandler<dim, spacedim> &dof_handler)
     {
+      if (dof_handler.get_fe_collection().size() == 0)
+        // nothing to do
+        return;
+
       Assert(
         dof_handler.has_hp_capabilities(),
         (typename dealii::DoFHandler<dim, spacedim>::ExcOnlyAvailableWithHP()));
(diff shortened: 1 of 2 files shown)


source/hp/refinement.cc
```diff
@@ -45,6 +45,10 @@ namespace hp
     void
     full_p_adaptivity(const dealii::DoFHandler<dim, spacedim> &dof_handler)
     {
+      if (dof_handler.get_fe_collection().size() == 0)
+        // nothing to do
+        return;
```
bangerth, line 50: Here and below, wouldn't the condition `==0` imply that the `DoFHandler` has never been initialized via `distribute_dofs()`? I'd consider that an error in this function and just assert so, rather than exiting without doing anything.

bangerth (changes requested):

source/hp/refinement.cc
marcfehling: > Here and below, wouldn't the condition ==0 imply that the DoFHandler has never been initialized via distribute_dofs()?

Yes.

> I'd consider that an error in this function and just assert so, rather than exiting without doing anything.

I'm okay with your proposed assertions for most of the `hp::Refinement` functions.

However, my plan is to attach the `predict_error()` and `limit_p_level_difference()` functions to the new `post_p4est_refinement` signal. Whenever calling the `refine_global` function to create an initial mesh, these functions will ultimately be called and at that stage dofs have not been enumerated yet. This was the original motivation of this PR :-)

I don't know how to deal with this other than exiting without doing anything. Do you have a different idea?

marcfehling: > I don't know how to deal with this other than exiting without doing anything. Do you have a different idea?

Hmm, users could simply connect these functions to the signal after generating the initial mesh. But this feels more like a hotfix than a solution.

bangerth (approved):
I see, thanks for the explanation!

PR #11991, opened 2021-03-31
Add settings to BoomerAMG

bergbauer:


source/lac/petsc_precondition.cc
```diff
@@ -435,19 +435,90 @@ namespace PETScWrappers
...
+    , relaxation_type_up(relaxation_type_up)
+    , relaxation_type_down(relaxation_type_down)
+    , relaxation_type_coarse(relaxation_type_coarse)
+    , n_sweeps_coarse(n_sweeps_coarse)
+    , tol(tol)
+    , max_iter(max_iter)
+    , w_cycle(w_cycle)
   {}
 
+  std::string
```
bangerth, line 464:
```suggestion


  std::string
```

include/deal.II/lac/petsc_precondition.h
```diff
@@ -684,21 +684,58 @@ namespace PETScWrappers
...
+        const double       tol             = 0.0,
+        const unsigned int max_iter        = 1,
+        const bool         w_cycle         = false);
+
+      /**
+       * Converts the enums for the different relax types to the respective
+       * strings for PETSc.
+       */
+      std::string
+      enum_to_string(RelaxationType relaxation_type);
```
bangerth, line 733: Move this function to a `private` section of the class.

bangerth: Also, if you think that one might at some point introduce more parameters one of which is another `enum` type, then maybe calling the function `relaxation_type_to_string` would be a better name.

bergbauer:
> Good work! Would you mind also putting an entry into `doc/news/changes/minor`?

I added a changelog entry!

include/deal.II/lac/petsc_precondition.h
bergbauer: As @nfehn mentioned this could be done implicitly through the `enum` type of the function. But I do not feel strongly here so if you think we should be explicit here I can change it.

bergbauer: Also I cannot call the function `enum_to_string` in `PreconditionBoomerAMG::initialize` if it is private. Would you put it in `PreconditionBoomerAMG` instead of `PreconditionBoomerAMG::AdditionalData`?

bangerth: I see. Maybe just call the function `to_string` then; there is already `std::to_string()` and we use that name in a number of places as well.

Regarding where the function should live: I just don't think it should be part of the public interface of the class. You only need it in the `.cc` file, so maybe make it a non-member function in an anonymous namespace in the `.cc` file?

bergbauer: Good idea, now it looks really clean! I renamed it to `to_string` and made it a `private` function of `PreconditionBoomerAMG`.

peterrum: @bergbauer You could take a look at the following code as reference:

https://github.com/dealii/dealii/blob/d002b101fcb94246913c664b737039aae4c2ff32/source/fe/mapping_q_cache.cc#L275-L289

bergbauer: Thanks for the hint @peterrum ! I pushed a new version with an anonymous namespace in the `.cc` file.

bangerth:
Looks good. The Windows-2019 failure is spurious (some upload failure). Let's get this merged!

/rebuild

bangerth:
Separately, if you wanted to clone some of the existing tests for this class and use different settings, that would of course also be a very nice follow-up PR!

peterrum (approved):

kronbichler:
I am late to this PR I admit, but I would suggest a small follow-up change (if we agree on it): Apart from the very few options set in `PreconditionBoomerAMG::AdditionalData` before, a user could supply other arguments in the PETSc style via the command line. Now we have changed that behavior, because the ones we set now take precedence over the command line arguments. While a programmatic way to change this is most often the way to go (when you nest objects of the same type, for example), it could still represent a regression for some use cases. We could relieve that by choosing default values for the parameters that we detect in the code and interpret as "do not set the PETSc/Hypre parameter in that case", so a command-line value would still get through.

nfehn:
In my opinion, the function `to_string()` should be realized as a public member function of `AdditionalData`. I actually do not see so much benefits by a free function in an internal namespace and would like to ask to reconsider this decision. The main reason is that in application programs, I would like to print all parameters to `std::cout` so that all the settings of a simulation are documented (particularly, I am referring to this code https://github.com/exadg/exadg/blob/a92a96756240cfc5866f81bc81916e4d5e42f97a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_input_parameters.h#L173 for the interested reader). For the PETSc smoother of this PR, this currently means printing integers or reimplementing the function `to_string()` in the application code?

PR #11992, opened 2021-03-31
Delete default constructor FESystem

peterrum:
As requested in https://github.com/dealii/dealii/issues/11964#issuecomment-808898044.

### DIFF
--- a/include/deal.II/fe/fe_system.h
+++ b/include/deal.II/fe/fe_system.h
@@ -210,12 +210,17 @@ class FE_Enriched;
  *
  * @ingroup febase fe vector_valued
  *
- * Hartmann 2001.
  */
 template <int dim, int spacedim = dim>
 class FESystem : public FiniteElement<dim, spacedim>
 {
 public:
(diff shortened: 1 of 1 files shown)


masterleinad (commented):
Alternatively, you might just check for `sizeof...(FEPairs)>0` in the `FESystem(FEPairs &&...)` constructor. I think I like that better since it's more concise.

bangerth:
I like the deleted default constructor better. If someone does what happened in the case where this was found, the intention was almost certainly not to create an `FESystem` with zero components -- in other words, there was no intention to call this constructor with a zero-element list, and it would be confusing to see the error message come from that place.

bangerth:
It might make sense to *also* put a `static_assert` into the variadic constructor. It will never trigger but serve as useful documentation. It would certainly make sense to `Assert` that the `std::initializer_list` constructor gets a non-empty list. But that is all separate from the current patch.

marcfehling (approved):

PR #11993, opened 2021-03-31
Add query methods for vtk cell types in ReferenceCell

agrayver:
This one introduces a centrally available interface. A follow-up PR will make use of it in the library.

### DIFF
--- a/include/deal.II/grid/reference_cell.h
+++ b/include/deal.II/grid/reference_cell.h
@@ -58,6 +58,7 @@ namespace internal
      */
     DEAL_II_CONSTEXPR dealii::ReferenceCell
                       make_reference_cell_from_int(const std::uint8_t kind);
+
   } // namespace ReferenceCell
 } // namespace internal
 
@@ -466,6 +467,26 @@ public:
    */
   unsigned int
   exodusii_face_to_deal_face(const unsigned int face_n) const;
+
+  /**
+   * Return a VTK linear shape constant that corresponds to the reference cell.
+   */
+  unsigned int
+  vtk_linear_type() const;
+
+  /**
+   * Return a VTK quadratic shape constant that corresponds to the reference
+   * cell.
+   */
+  unsigned int
+  vtk_quadratic_type() const;
+
+  /**
+   * Return a VTK Lagrange shape constant that corresponds to the reference
+   * cell.
+   */
+  unsigned int
+  vtk_lagrange_type() const;
 
   /**
    * @}
--- a/source/grid/reference_cell.cc
+++ b/source/grid/reference_cell.cc
@@ -34,6 +34,47 @@
 
 DEAL_II_NAMESPACE_OPEN
 
+namespace
+{
+  namespace VTKCellType
+  {
+    // Define VTK constants for linear, quadratic and
+    // high-order Lagrange geometrices
+    enum
+    {
+      VTK_VERTEX = 1,
+      // Linear cells
+      VTK_LINE       = 3,
+      VTK_TRIANGLE   = 5,
+      VTK_QUAD       = 9,
+      VTK_TETRA      = 10,
+      VTK_HEXAHEDRON = 12,
+      VTK_WEDGE      = 13,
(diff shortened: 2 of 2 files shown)


include/deal.II/grid/reference_cell.h
```diff
@@ -58,6 +58,41 @@ namespace internal
...
+      // Lagrange cells
+      VTK_LAGRANGE_CURVE         = 68,
+      VTK_LAGRANGE_TRIANGLE      = 69,
+      VTK_LAGRANGE_QUADRILATERAL = 70,
+      VTK_LAGRANGE_TETRAHEDRON   = 71,
+      VTK_LAGRANGE_HEXAHEDRON    = 72,
+      VTK_LAGRANGE_WEDGE         = 73,
+      VTK_LAGRANGE_PYRAMID       = 74,
+      // Invalid code
+      VTK_INVALID = unsigned(-1)
```
bangerth, line 93:
```suggestion
      VTK_INVALID = static_cast<unsigned int>(-1)
```

include/deal.II/grid/reference_cell.h
```diff
@@ -58,6 +58,41 @@ namespace internal
      */
     DEAL_II_CONSTEXPR dealii::ReferenceCell
                       make_reference_cell_from_int(const std::uint8_t kind);
+
+
+    // Define VTK constants for linear, quadratic and
+    // high-order Lagrange geometrices
+    enum vtk_cell_types
```
bangerth, line 65: I would prefer if you moved the declaration of this enum into the `.cc` file and if these were members of an `enum class VTKCellTypes`.

include/deal.II/grid/reference_cell.h
```diff
@@ -467,6 +502,25 @@ class ReferenceCell
   unsigned int
   exodusii_face_to_deal_face(const unsigned int face_n) const;
 
+  /**
+   * Return a VTK linear shape constant that corresponds to the reference cell
```
bangerth, line 506:
```suggestion
   * Return a VTK linear shape constant that corresponds to the reference cell.
```
Same below

bangerth (changes requested):
Good work!

bangerth:
It also fails to compile with error messages like this:
```
/Users/runner/work/dealii/dealii/source/grid/reference_cell.cc:299:12: error: use of undeclared identifier 'VTK_VERTEX'; did you mean 'internal::ReferenceCell::VTK_VERTEX'?
    return VTK_VERTEX;
           ^~~~~~~~~~
           internal::ReferenceCell::VTK_VERTEX
```
But that will go away once you move the declaration of the enum anyway.

include/deal.II/grid/reference_cell.h
bangerth: More precisely, move it into an anonymous namespace.

agrayver: I did that, but now there is not implicit conversion to unsigned it, so I need to use static_cast<unsigned int>(...). Is that ok?

bangerth: Ah, I see. That's ugly as well...

How about
```
  namespace {
    namespace VTKCellTypes {
      constexpr unsigned int  VTK_abc = 42;
    }
  }
```
?

bangerth: Or
```
  namespace {
    namespace VTKCellTypes {  
      enum {    // no name given
        VTK_abc = 42
     };
  }
}
```

agrayver: I'm fine doing it either way, but I'm curious why not 

```
 namespace {
      enum VTKCellTypes {   
        VTK_abc = 42
     };
  }
```

bangerth: enums are not scoped and neither are anonymous namespaces. Using your example, you inject the name into the global namespace of that file. We try to avoid that and apply some basic "sanitary practices" to the global namespace :-)

agrayver:
I have addressed the comments.

bangerth (approved):

PR #11995, opened 2021-03-31
Deprecate DoFTools::extract_boundary_dofs() with std::vector<bool> argument

bangerth:
We have slowly been replacing these by functions using `IndexSet` instead, and indeed such a function already exists. The function here is not particularly widely used.

/rebuild

PR #11996, opened 2021-03-31
Deal with the other DoFTools::extract_boundary_dofs() function

bangerth:
This is a follow-up to #11995: There are currently two variants of the function `DoFTools::extract_boundary_dofs()`, both of which return information through a reference argument. One does so via `std::vector<bool>` (deprecated in #11995) whereas the other uses `IndexSet`. But it is easiest (and plays nicer with default arguments) if one would just return objects. We didn't do that in the olden days because there were no move constructors, but there are now and so there is no reason not to do that any more -- and indeed, we've done that with a number of `DoFTools` functions in the recent past.

So this function deprecates the function that returns by reference argument and introduces one that just returns the information. I used this to simplify the code in step-11 a bit.

/rebuild

### DIFF
--- a/doc/news/changes/minor/20210330Bangerth-2
+++ b/doc/news/changes/minor/20210330Bangerth-2
@@ -1,0 +1,7 @@
+Deprecated: The version of DoFTools::extract_boundary_dofs() that
+returns its information via an `IndexSet` reference argument has been
+deprecated. Use the version of the function that returns information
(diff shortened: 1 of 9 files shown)


drwells:
We can close this now that all the commits are in #11995, right?

bangerth:
Yes.

PR #11963, opened 2021-03-25
Add nearest predicate to ArborX wrapper

Rombur:
This PR adds a nearest neighbor predicate to ArborX wrappers. While adding the predicate, I realized that the current design was not extensible and that the names were not great. I have created base classes that contains most of the implementation and added `Predicate` to the name of the classes since that's what they are. The classes were introduces a month ago so there is no backward compatibility problem.

PR #11999, opened 2021-03-31
Avoid usage of the ArborX access traits in polymorphic contexts

masterleinad:
A discussed in https://github.com/dealii/dealii/pull/11963#issuecomment-811072031.

@Rombur Please test that this actually works for you.

### DIFF
--- a/include/deal.II/arborx/access_traits.h
+++ b/include/deal.II/arborx/access_traits.h
@@ -62,8 +62,9 @@ namespace ArborXWrappers
    * This class defines a predicate used by ArborXWrappers::BVH to determine
    * for given points which of the bounding boxes used to build the
    * ArborXWrappers::BVH intersect with them.
+   * @note The class is not supposed to be used in polymorphic context.
    */
-  class PointIntersectPredicate : public PointPredicate
+  class PointIntersectPredicate : private PointPredicate
   {
   public:
     /**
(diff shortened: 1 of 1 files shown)


Rombur:
Actually I was thinking about this more, shouldn't we keep the public inheritance and instead have all the functions (including the constructor) of the base predicate protected. That way the base class cannot be used at all.

bangerth:
That's fine with me as well. I think any solution is good with me if maybe we just augment the documentation of the base class. When I read "base class", I think "interface", but that's not how this is done here. It's just a way to provide some "basic functionality" ("basic" as in the adjective for "base"). If you just write the documentation of the base class as stating that, that would be ok as well.

masterleinad:
> Actually I was thinking about this more, shouldn't we keep the public inheritance and instead have all the functions (including the constructor) of the base predicate protected. That way the base class cannot be used at all.

If we want to prevent use in a polymorphic context, i.e.
```
Base* base = new Derived;
```
we need to derive protectedly or privately and we need to import the functions anyway (even if we derive publically and define the whole body of the base class as `protected`) to access the member functions.
Thus, we could additionally have the body of the base classes protected to disallow creating base class instances but not replace the approach here.

masterleinad:
I used https://godbolt.org/z/PrYev6Pn4 for playing with this.

Rombur (approved):
Thanks for the PR @masterleinad

PR #12002, opened 2021-04-01
FindARBORX: Remove library variable

tamiko:
ArborX is a header only library, so we don't need to specify an empty
string here :-)


tamiko:
@Rombur *ping* Quick, minor cleanup :smile_cat:

Rombur (approved):

masterleinad (approved):

PR #12003, opened 2021-04-01
Add a subdivided_cylinder grid generator and test

blaisb:
This very small PR adds a new grid generator : subdivided_cylinder. This generator is only valid for spacedim==3.
It consists in a cylinder in which the number of cell in the x direction can be controlled. In chemical engineering, we often deal with pipes or fluidized bed reactors, which are very narrow cylinders. Hence we need more control on the granularity in the x direction. This PR achieves just that.
I added it as a new grid generator instead of "generalising" the previous one. I know this introduces code duplication so if you prefer that the regular cylinder gains an optional argument at the end, I will gladly do it. I also added a test to test the new feature.

As always, I will be super happy to integrate all of your comments :).


drwells:
Could you add a picture to the documentation?

blaisb:
> Could you add a picture to the documentation?

Sure thing. I'll add one :)

source/grid/grid_generator.cc
```diff
@@ -5560,6 +5579,137 @@ namespace GridGenerator
...
+                           const unsigned int x_subdivisions,
+                           const double       radius,
+                           const double       half_length)
+  {
+    // Copy the base from hyper_ball<3>
+    // and transform it to yz
+    const double d = radius / std::sqrt(2.0);
+    const double a = d / (1 + std::sqrt(2.0));
+
+    // const unsigned int    npts_per_plane = 8;
```
masterleinad, line 5594:
```suggestion
```
?

include/deal.II/grid/grid_generator.h
```diff
@@ -948,6 +948,46 @@ namespace GridGenerator
...
+   *
+   * The manifold id for the hull of the cylinder is set to zero, and a
+   * CylindricalManifold is attached to it.
+   *
+   * @param tria The triangulation to be created. It needs to be empty upon
+   * calling this function.
+   *
+   * @param x_subdivisions A positive integer denoting the number
+   * of cells to generate in the x direction. The default cylinder has
+   * x_repetitions=2.
```
masterleinad, line 977: Can we implement the "default" cylinder in terms of this function?

source/grid/grid_generator.cc
```diff
@@ -5560,6 +5579,137 @@ namespace GridGenerator
...
+    // an edge is purely in the
+    // interior if one of its vertices
+    // is at coordinates '+-a' as set
+    // above
+    Triangulation<3>::cell_iterator cell = tria.begin();
+    Triangulation<3>::cell_iterator end  = tria.end();
+
+    tria.set_all_manifold_ids_on_boundary(0);
+
+    for (; cell != end; ++cell)
```
masterleinad, line 5670: Let's use range-based for loops.
```suggestion
    for (const auto& cell:tria.cell_iterators())
```

include/deal.II/grid/grid_generator.h
blaisb: Yes absolutely I can do this. Regular cylinder could call this function with 2 subdivision. I'll do this.

source/grid/grid_generator.cc
blaisb: Junk line. Removing.

source/grid/grid_generator.cc
blaisb: Done

blaisb:
This should be ok now, I don't understand why two tests failed though?
@masterleinad  I applied all your comments :)

masterleinad (approved):
You should also provide a changelog entry. :slightly_smiling_face: 
Assuming that this passes the test suite, it looks good to me otherwise.

masterleinad:
> This should be ok now, I don't understand why two tests failed though?

Let's see if restarting the checks helped.

blaisb:
> You should also provide a changelog entry. 
> Assuming that this passes the test suite, it looks good to me otherwise.

Done :). Added a changelog entry. Thanks for reminding me. I always forget!

drwells:
Based on the picture - wouldn't it be easier to generate a disk, extrude, and rotate? We have a generalized extrude function that does slices in a very similar way to what you implemented.

blaisb:
> Based on the picture - wouldn't it be easier to generate a disk, extrude, and rotate? We have a generalized extrude function that does slices in a very similar way to what you implemented.

Maybe yes. I used the same process as the original cylinder code that was there in the grid generator. I guess the same comment could be applied to the regular cylinder that was already there. I'm not sure why the original cylinder did not use this process, maybe there is a reason I am incapable of identifying.
I would be glad to look into this, but maybe in a next PR where I could refactor this? I'm not sure I am 100% aware of all the impacts that this would have on the other cylinder tests and etc. right now.

PR #12005, opened 2021-04-02
Some KINSOL documentation updates.

bangerth:


include/deal.II/sundials/kinsol.h
```diff
@@ -572,6 +572,38 @@ namespace SUNDIALS
...
+     * typically because they have different physical units and represent
+     * different things. For example, if one were to solve a nonlinear Stokes
+     * problem, the solution vector has components that correspond to velocities
+     * and other components that correspond to pressures. These have different
+     * physical units and depending on which units one chooses, they may have
+     * roughly comparable numerical sizes or maybe they don't. To give just one
+     * example, in simulations of flow in the Earth's interior, one has
+     * velocities on the order of maybe ten centimeters per year, and pressures
+     * up to around 100 GPa. If one expresses this in SI units, this corresponds
+     * to velocities of around $0.000,000,003=3 \times 10^{-9}$ m/s, and
```
jppelteret, line 587: I've got LibreOffice set to "US English" and punted `0.0000000003` into calc. It prints out `0.0000000003` without any decimals... is that really how you'd format it (just asking)? It looks so weird to me.

jppelteret (approved):
Thanks for this clarification @bangerth. I think that helps a lot in understanding what these functions do.

jppelteret:
Since this is just documentation, I'll merge it as is.

include/deal.II/sundials/kinsol.h
bangerth: @jppelteret  My brain has a parser failure with your sentence above, but I suspect that you are commenting on the use of commas as thousands separators? If so, it is common in AE usage to use commas as thousands separator before the decimal point. It is not frequently used after the decimal point, and I don't know whether that is idiomatic or not. But it enhanced the readability of the number, and so I thought it's worth using here. It's just documentation, so I thought not overly critical.

bangerth:
Related to #11829.

Issue #12008, opened 2021-04-02
Allow sharing of data across MPI processes


bangerth:
@tjhei and I were talking about sharing large data chunks across MPI processes that live on the same machine and, consequently, can use shmem. We were specifically interested in making this possible with `InterpolatedTensorProductFunction` and friends. For ASPECT, we will probably have some quite large data tables for initial conditions that require this kind of thing -- think 100s of MB or even a few GB; small enough that they can fit into memory, but large enough that it will be quite inefficient if the data is replicated 64 times on a single machine.

Here's how I think we should address this:
* Underlying all relevant classes is `AlignedVector`. If we could get that class to share its data with others, that would already be a big step forward. My plan is to give this class a function `share_data_across_communicator(MPI_Comm)` in which one creates an `AlignedVector` object on one processor and then calls this function collectively on a sub-communicator of processes that can participate via shared memory. The data is then moved into a shmem arena that can be accessed by all other processes.
* The next level up is `TableBase` that under the hood uses `AlignedVector`. It will get the same function and just pass it on to its underlying aligned vector plus exchange minimal amount of information for the size of the table.
* `InterpolatedTensorProductGridData` currently copies its `Table` of data from the calling site, but it would actually make sense to give this class a separate constructor that *moves* the data. In most cases, one is happy to get rid of the table with which this class was constructed, because it is used only in this one context. Once the class is able to move the data, it should also be able to deal with data that lives in a shmem arena.

peterrum:
The description sounds reasonable. This implies that the user needs to know that the user knows which process has rank 0 in the subcommunicator. 

Regarding the subcommunicator, you might want to take a look at: https://github.com/dealii/dealii/pull/11158#discussion_r549487431

You might also want to have a look at the implementation in `LA:d:V`: 

https://github.com/dealii/dealii/blob/09b93afd2f490bb6c517fe4e20f1e02108837905/include/deal.II/lac/la_parallel_vector.templates.h#L135-L246

I guess in your use case `alloc_shared_noncontig=true` has to be set to `alloc_shared_noncontig=false`, which might limit the maximum size of the array.

tjhei:
This makes sense to me and is roughly what @bangerth and I talked about before. My thought was to leave the MPI logic outside of AlignedVector and only give a new reinit()/constructor to AlignedVector that points to an existing piece of memory.

We need to be careful when modifying AlignedVector and Table to not introduce performance regressions, as they are used in performance critical areas.

[...]

PR #12009, opened 2021-04-02
Use std::unique_ptr to manage memory in AlignedVector.

bangerth:
Relates to #12008. There is more to do here, but this is a good first step.

/rebuild


peterrum (approved):

tjhei (approved):
Looks good. Have you considered renaming data_begin to data? Maybe for a followup...

include/deal.II/base/aligned_vector.h
```diff
@@ -719,7 +720,7 @@ inline AlignedVector<T>::AlignedVector()
 
 template <class T>
 inline AlignedVector<T>::AlignedVector(const size_type size, const T &init)
-  : data_begin(nullptr)
+  : data_begin(nullptr, [](T *) { Assert(false, ExcInternalError()); })
```
drwells, line 723: This doesn't seem right. Suppose someone calls this with `size = 0` and never resizes the vector. The object will then be in a state where its destructor unconditionally aborts.

A better solution would be to just have an empty function body for the deleter - it doesn't need to do anything.

drwells (changes requested):
I have a comment on the deleters.

include/deal.II/base/aligned_vector.h
masterleinad: ... also destructors should not throw (which might happen if `disable_abort_on_exception` is enabled).

drwells: I should have checked the standard before writing that - it looks like unique_ptr doesn't call the deleter when it has a `nullptr`. The following code works:
```cpp
#include <iostream>
#include <memory>

template <typename T>
struct AlignedVector
{
  std::unique_ptr<T[], void (*)(T *)> data_begin;

    AlignedVector()
        : data_begin(nullptr, [](T *) { std::cout << "BBB\n"; std::abort();})
    {
        std::cout << "AAA\n";
    }
};

int main()
{
    AlignedVector<int> vec;
}
```

bangerth:
The tester seems stuck. Let me rebase...

/rebuild

bangerth:
I've also added a test to check the cases @drwells was concerned about.

bangerth:
On 4/5/21 10:28 AM, Timo Heister wrote:
> Looks good. Have you considered renaming data_begin to data? Maybe for a 
> followup...

Yes, actually that patch is already ready and waiting for this one to be
merged :-) The asymmetry between `data_begin` (pointing to memory it owns) and 
`data_end` (not owning anything) bothered me.

PR #12012, opened 2021-04-04
Performance improvement of MGTwoLevelTransfer

kronbichler:
The idea of this PR is to improve the performance of the two level transfer in a few scenarios:
- Cast from `double` to `float` implied by `1.0`; since the code in question is very short we can simply duplicate the loop
- Avoid zeroing the vector when we can overwrite things
- Do not involve an additional `vec_fine` for the case of discontinuous elements because there is no ghost.

There are some doubts I cannot resolve quickly with the time I have right now, so I wanted to discuss them with @peterrum: The first one is that the current optimization might be invalid in case we do h-transfer between discontinuous elements, and we would need the fine side to still import additional ghosts. Is that right? If yes, we might need to add another check for that. The other one is that we might want to further optimize the access e.g. by vectorizing things. The last one is whether we can skip the parts with constraints in `prolongate` in case we have discontinuous elements.

### DIFF
--- a/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
+++ b/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
@@ -1238,12 +1238,6 @@ namespace internal
         transfer.schemes[1].fine_element_is_continuous =
           fe_fine.dofs_per_vertex > 0;
 
-      Assert(
-        transfer.schemes[0].fine_element_is_continuous ||
-          constraint_fine.n_constraints() == 0,
-        ExcMessage(
-          "For discontinuous elements, no constraints are supported at the moment."));
-
       // count coarse cells for each scheme (0, 1)
       {
         transfer.schemes[0].n_coarse_cells = 0; // reset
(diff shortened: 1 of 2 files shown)


peterrum:
I have added the `Do not merge` label, since the tests are not passing (these are not executed by Jenkins). I guess the reason is:
> The first one is that the current optimization might be invalid in case we do h-transfer between discontinuous elements, and we would need the fine side to still import additional ghosts. Is that right?

I'll need to investigate. 

> The other one is that we might want to further optimize the access e.g. by vectorizing things.

@kronbichler What do you have here in mind?

> The last one is whether we can skip the parts with constraints in prolongate in case we have discontinuous elements.

A simple check it DG or not won't work, since users can provided any type of constraints. However, one could check if any constraints are provided and if not (what is the normal case in DG) one could indeed skip the step.

peterrum:
@kronbichler I have pushed directly onto your branch. I did the following two changes:
- perform copy on the fine level only if `is_globally_compatible()==false` (which is not optimal, but I have not a better heuristic at the moment)
- assert that DG does not have any constraints (for now that should be fine - only we need this combination, we can add an appropriate implementation)

include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h
```diff
@@ -2124,6 +2136,13 @@ MGTwoLevelTransfer<dim, LinearAlgebra::distributed::Vector<Number>>::prolongate(
 
   const unsigned int n_lanes = VectorizedArrayType::size();
 
+  const bool fine_vectors_are_compatible =
+    this->vec_fine.get_partitioner()->is_globally_compatible(
+      *dst.get_partitioner());
```
peterrum, line 2141: I think we should actually rewrite `Paritioner::is_globally_compatible()` so that it is hash based and does only perform communication during setup but not during this function call. This would fix this issue here.

peterrum:
@kronbichler Do you have time in the next days to finish this PR together?

peterrum (approved):

masterleinad (approved):
Looks good to me.

PR #12016, opened 2021-04-05
Improve conformity tests of 2D vector elements 

konsim83:
After @bangerth asked me (when merging a non-standard mesh into the GridGenerator namespace) if I can generate all possible orientation cases I should have though a bit longer... I generated only half of the cases. This PR fixes this issue.

The new conformity tests for vector elements are of course affected by that and it turned out though that (since now two more scenarios are tested) 2D RaviartThomas are broken in these cases (Nedelec elements are even broken in all pathological cases).I did not change element behavior, only tests. This [forum post](https://groups.google.com/g/dealii/c/BJI2uj-iAqQ) is related to that.

These pathological cases in pure 2D are rare but I guess can happen for periodic meshes since these are topologically 3D. I guess it is worth to go into that.

This PR is work in progress and affects only tests (some will be broken).

**Remark:** Funnily the NedelecSZ element in 2D is broken as indicated by the new tests. This element does not query the `cell->line_orientation(line_no)` but rather the `cell->face_orientation(face_no)`. Could it be that this is a bug in the `cell->face_orientation(face_no)` output in 2D? This confuses me. Face orientation is not the same as line orientation.


peterrum:
> This element does not query the cell->line_orientation(line_no) but rather the cell->face_orientation(face_no). Could it be that this is a bug in the cell->face_orientation(face_no) output in 2D? This confuses me. Face orientation is not the same as line orientation.

It would surprise that `line_orientation()` and `face_orientation()` return different values in 2D, since `face_orientation()` should call `line_orientation()` in 2D:

https://github.com/dealii/dealii/blob/2a8b880ac45aed20c14c4b36da8274a33d33a4cf/include/deal.II/grid/tria_accessor.templates.h#L678-L684

and the returned value is `true`, since the mesh has been reorientated:

https://github.com/dealii/dealii/blob/2a8b880ac45aed20c14c4b36da8274a33d33a4cf/include/deal.II/grid/tria_accessor.templates.h#L802-L803

Do you have a simple code where these the return values do not match?

konsim83:
> It would surprise that `line_orientation()` and `face_orientation()` return different values in 2D, since `face_orientation()` should call `line_orientation()` in 2D:
> Do you have a simple code where these the return values do not match?

Oh, sorry, this is not what I meant to say. The output of both functions does match. My question is rather whether they should really do that since lines and faces are two different things. To quote from my forum post:

"""
Supoose I have 2 squares (a left and a right one) sharing one edge. If everything is fine all lines and faces have standard orientation, i.e., line_orientation==true and face_orientation==true for all lines and faces. Note that lines and faces are NOT the same geometric entity here since their orientation is determined differently (faces have normals and lines have tangents).

Now, if I rotate the right square by 90 degrees I get (on the shared edge) non-matching face normals but all line orientations do match.

If I rotate the right square by 180 degrees I get (on the shared edge) non-matching face normals and non-matching line orientations.

If I rotate the right square by 270 degrees I get (on the shared edge) matching face normals and non-matching line orientations.

This is at least my intuition. When I query these information for each rotation case I see that my intuition is confirmed for lines but the face orientations are always true.
"""
Some of these things rarely happen in 2d. One example would be simple periodic meshes which cause line orientations not to match (this case is caught). In more complicated scenarios involving periodicity and rotations of faces it can happen though that lines do match but normals do not.

According to @bangerth it seems worth to have a look at the mesh orienter. And I may give it a try.

peterrum:
@konsim83 I would be happy to get involved in your investigations. I bumped into orientation issues not just once.(see also https://github.com/dealii/dealii/pull/10827).

But let me ask you why should lines and faces be different in 2D?

A simple failing test would be a good starting point.

>  One example would be simple periodic meshes which cause line orientations not to match (this case is caught).

Boundary faces are always correctly orientated. I guess the sign-flip (or other corrections) happen (or should happen) in the context of PBC with the help of the `AffineCostraint` class (`DoFTools::make_periodicity_constraints()`).

source/grid/grid_generator.cc
```diff
@@ -235,7 +235,8 @@ namespace GridGenerator
         /**
          * Create a serial/parallel distributed triangulation.
          */
-        void create_triangulation(
+        void
+        create_triangulation(
```
bangerth, line 239: Is this what the indentation script does? Or did you make these changes by hand?

bangerth: I actually like this style better, and if the indentation script lets you do that, I'm in favor. But let's put this into a separate pull request.

tests/fe/fe_conformity_test.h
```diff
@@ -209,9 +209,11 @@ namespace FEConforimityTest
     const bool rotate_left_square  = (((config_switch / 2) % 2) == 1);
     const bool rotate_right_square = ((config_switch % 2) == 1);
 
+    // alias for better readabilty
+    const unsigned int &n_rotate_right_square = config_switch;
```
bangerth, line 213:
```suggestion
    const unsigned int n_rotate_right_square = config_switch;
```

tests/grid/grid_generator_non_standard_orientation_mesh_2d.cc
```diff
@@ -74,25 +74,22 @@ main(int /*argc*/, char ** /*argv*/)
   deallog.attach(logfile);
 
   /*
-   * 3D test only
+   * 2D test only
    */
   const int          dim = 2;
```
bangerth, line 79: :-)

bangerth (commented):
Here are a couple of comments. Am I correct in reading your original post to say that with these changes, the tests fail?

source/grid/grid_generator.cc
konsim83: The indentation script did it. I do not indent by hand since then the CI usually complains.

tests/fe/fe_conformity_test.h
konsim83: done

konsim83:
> @konsim83 I would be happy to get involved in your investigations. I bumped into orientation issues not just once.(see also #10827).
 
Will be glad to talk.

> But let me ask you why should lines and faces be different in 2D? Maybe this is a stupid question but I would say in 2D faces are lines.

This question is far from stupid. It depends on how you want to describe a geometric object that lives on it. Vector fields are not vector fields (in particular not in higher dimensions), meaning that it really depends what you want to describe. To answer with an analogy from physics: there are quantities that can be measured (integrated) along lines like electric fields and there are quantities that describe fluxes through surfaces such as magnetic fields. Both are vector fields but have somewhat orthogonal meaning. Physicists often mumble something like "some things transforms co-variantly or contra-variantly". These quantities of course appear in PDEs and tend to live in different spaces. In Numerics you see that when you want to discretize: Nedelec elements (defined through line moments) are made for thing like electric fields while Raviart-Thomas elements (defined through face moments) are more suitable for magnetic fields (there is of course some form of geometric duality involved between them). 

Now, in Deal.II (or anywhere else) one needs to carefully look into the definition of elements in terms of the reference element that has prescribed orientations. Imagine a 2d (or 3d) mesh. Under rotation of a physical cell in the interior of a mesh it can happen that a rotation of this cell will create non-matching lines while having matching face orientation or the other way around or even bot non-matching. 

Long story short: Nedelec-type or H(div)-type elements querying the wrong type of information can lead to inconsistencies in the discretization. Also, if these elements query the correct info the discretization can be messed up if there is a little inconsistency in the orientation concept of cells. I guess this is what I see here but I am not 100% confident yet.

> A simple failing test would be a good starting point.
> 
> > One example would be simple periodic meshes which cause line orientations not to match (this case is caught).
> 
> Boundary faces are always correctly orientated. I guess the sign-flip (or other corrections) happen (or should happen) in the context of PBC with the help of the `AffineCostraint` class (`DoFTools::make_periodicity_constraints()`), which connects the distinct entities.

I was wrong here, sorry,  you are right.

tests/grid/grid_generator_non_standard_orientation_mesh_2d.cc
konsim83: luckily it was just a comment ...

source/grid/grid_generator.cc
konsim83: Apparently, something went wrong here. :-/ It should be fixed now.

konsim83:
OK, little update: While killing some time with this issue I found the following:

- I could fix the elements ABF, BDM, RaviartThomas and RaviartThomasNodal.
- I found an inconsistency in FE_Q which is quite worrying but that was probably not noticed since the corresponding mesh config probably rarely arises. Potentially it could on 3D surface meshes.
- Nedelec in 2D is broken but I have a clue why
- NedelecSZ in 2d is broken for degree 1 and larger 
- all tested 3D cases seem fine (note I did not test all elements)

After having a conversation with @peterrum I have a hypothesis why NedelecSZ and FE_Q could be broken (not confirmed). This is also what let me come up with this issue from the beginning: in 2d line and face orientation should not be the same. I will investigate this issue further once I have some more peace and quite.

source/fe/fe_poly_tensor.cc
```diff
@@ -67,17 +69,26 @@ namespace internal
...
+                                                     spacedim>::cell_iterator
+                             neighbor_cell_at_face = cell->neighbor(f);
+                const CellId neigbor_cell_id = neighbor_cell_at_face->id();
+
+                // if (nn < GeometryInfo<dim>::faces_per_cell / 2)
+                // Only fix sign if the orientation is opposite and if it has
+                // not been fixed yet.
+                // NOTE: This assumes that we iterate through cells in
+                // ascending cell_id-order
+                if (((nn + f) % 2 == 0) && this_cell_id < neigbor_cell_id)
```
bangerth, line 91: Do you mean a loop over all cells? In that case, the assumption is wrong. Roughly speaking, the order in which cells are traversed is the order in which the cells were created. So, if you start from a mesh of two cells, and you now have a mesh with 2x4 cells (globally refined) you may first visit the 4 children of the first coarse mesh cell, or the 4 children of the second coarse mesh cell, depending on what steps you took to arrive at the mesh with the 2x4 cells.

source/grid/grid_generator.cc
```diff
@@ -5086,19 +5089,25 @@ namespace GridGenerator
...
+                                              {3, 5, 10, 11}};
 
-    // all 4 true-false combinations of (rotate_left_square | rotate_right_square) to a number 0..3
-    unsigned int this_case = 2 * rotate_left_square + rotate_right_square;
-
-    switch (this_case)
+    switch (n_rotate_middle_square)
       {
-        case /* rotate only right square */ 1:
+        case /* rotate right square */ 1:
```
bangerth, line 5110: did you mean "rotate the central square"?

tests/fe/fe_conformity_test.h
```diff
@@ -209,9 +209,11 @@ namespace FEConforimityTest
     const bool rotate_left_square  = (((config_switch / 2) % 2) == 1);
     const bool rotate_right_square = ((config_switch % 2) == 1);
 
+    // alias for better readabilty
+    const unsigned int n_rotate_right_square = config_switch;
```
bangerth, line 213:
```suggestion
    const unsigned int n_rotate_central_square = config_switch;
```

bangerth (commented):
I'm tempted to merge this as is, if you're ok with it. You mention in your last commit that you've found all sorts of other bugs -- but they are not addressed in this patch?

bangerth:
Might also want to squash commits. Thanks as always for the work!

source/fe/fe_poly_tensor.cc
konsim83: I see. My observation was that (as was probably intended also before my change) that one should only correct a sign once when checking neighbor orientations. If we have a cell whose face dof signs have been corrected one must not do so again. The approach before unfortunately fails on some mesh configurations.

source/grid/grid_generator.cc
konsim83: yes

konsim83:
> Might also want to squash commits. Thanks as always for the work!

You are most welcome. I would not feel good though to merge it yet. Some test will be broken I guess and I need to see a bigger picture. Is it ok to add the ready for test label? What is your policy with that?

source/fe/fe_poly_tensor.cc
konsim83: OK, from the tests I see that (among others) fe/fe_values_view_09.cc is broken. That is likely because of that cell_id_order

konsim83:
> I'm tempted to merge this as is, if you're ok with it. You mention in your last commit that you've found all sorts of other bugs -- but they are not addressed in this patch?

@bangerth  I am trying to address a bug in FE_RaviartThomas<2> but I guess this is a minor issue. Not sure if I should really attack that in this PR. What is the deal.ii policy when changing tests and discovering new bugs? The old test output is then obviously incomplete while the new test output that revealed the bug is obviously wrong. Do you remove the old tests then? Or restrict them?

source/fe/fe_poly_tensor.cc
bangerth: Right. What you need is a "tie breaker" that for a given face decides whether one or the other adjacent cell needs to flip a sign. You can't rely on the iteration order of cells (i.e., the `cell->active_cell_index()`), but the order provided by `cell_id` is stable and also stable for parallel triangulations. So, the approach should be something like this: 
* figure out whether for the current face a sign needs to be flipped
* flip the sign on that cell that has `this_cell_id < neighbor_cell_id`.

I don't understand the first part of the condition, but the second part looks correct. I just can't quite make the connection to the comment above -- we don't iterate over cells in ascending cell_id order, but I also don't think the condition actually relies on this.

bangerth:
@konsim83 First, the "ready to test" label and associated "/rebuild" tag is really only there to avoid the security problems you run into if you always run the test suite for *any* PR that comes in. Someone could propose a PR that changes things so that the test suite would execute malicious code. As a consequence, someone will have to look through a patch to convince themselves that nothing of that sort happens, and then we are happy to set the label and tag. Daniel has already done this.

As for fixes that break tests: Since we can only commit pull requests that pass all tests that the CI infrastructure runs, there are only two options.
* Remove the tests that the patch uncovers have been wrong in the past. We don't like to do that: Most of the tests are actually useful, they just had output that was wrong (but we didn't know that). So removing a test is not usually useful.
* Fix the output. If the test was useful in the past, it will be even more useful with corrected output. So if a patch uncovers a bug, we verify that the new output is correct and then update the output of the test.

The current situation (as I write this comment) is that all but the "CI" checks succeed. That one, though, shows that the following tests fail:
```
	592 - grid/grid_generator_non_standard_orientation_mesh_2d.debug (Failed)
	823 - fe/fe_values_view_14.debug (Failed)
	825 - fe/fe_values_view_10_single_03.debug (Failed)
	834 - fe/fe_conformity_dim_2_fe_bdm.debug (Failed)
	840 - fe/fe_values_view_03.debug (Failed)
	866 - fe/fe_conformity_dim_2_fe_nedelec.debug (Failed)
	871 - fe/fe_conformity_dim_3_fe_rt.debug (Failed)
	876 - fe/fe_values_view_10.debug (Failed)
	965 - fe/fe_values_view_18.debug (Failed)
	1003 - fe/fe_values_view_08.debug (Failed)
	1009 - fe/fe_conformity_dim_2_fe_rt_nodal.debug (Failed)
	1020 - fe/fe_conformity_dim_2_fe_abf.debug (Failed)
	1029 - fe/fe_conformity_dim_3_fe_nedelec_sz.debug (Failed)
	1038 - fe/fe_conformity_dim_2_fe_rt.debug (Failed)
	1064 - fe/fe_values_view_04.debug (Failed)
	1096 - fe/abf_01.debug (Failed)
	1099 - fe/fe_values_view_16.debug (Failed)
	1100 - fe/fe_values_view_15.debug (Failed)
	1109 - fe/fe_conformity_dim_2_fe_nedelec_sz.debug (Failed)
	1148 - fe/fe_values_view_09.debug (Failed)
	1166 - fe/fe_values_view_20.debug (Failed)
	1180 - fe/fe_values_view_10_single_01.debug (Failed)
	1187 - fe/fe_conformity_dim_2_fe_q.debug (Failed)
	2175 - numerics/create_mass_matrix_05.debug (Failed)
	2184 - numerics/vectors_boundary_rhs_02.debug (Failed)
	2906 - hp/create_mass_matrix_05.debug (Failed)
	2930 - hp/vectors_boundary_rhs_02.debug (Failed)
	2994 - hp/vectors_rhs_02.debug (Failed)
```
You probably can reproduce these failures on your local system. You can update the stored outputs for these tests, but you should probably go through each test and convince yourself that the new output is indeed "more correct" than the old output. This looks like a daunting task with 28 affected tests, but you will probably realize quickly that for most of these, the changes are all of the same kind and that the 28 tests only fall into 3 or 4 categories that need to be verified.

I assume that you know how to look at failing output with `ctest -V -R testname`? And that among the output of `ctest -V` is a line that says `Check ...1 ...2` where the first `...1` is the output file just created and the second is the reference test stored in the git repo. I usually copy paste this onto the command line as
```
  diff ...1 ...2 | less
```
to look at the diffs, and if I'm convinced that things are correct now, do
```
  cp ...1 ...2 | less
```

As for the other bugs you know about: Let's please keep these for separate PRs :-) It's hard enough to review a PR that fixes a single issue and updates a bunch of test outputs. It gets worse if a single PR commingles several of these :-) So, one step at a time!

bangerth:
@konsim83 I think it would be nice if we could merge this before the next release, which means in essence this week or next. Do you think you have time to work on the failing tests?

konsim83:
@bangerth Yes, I saw your update :-)  I will work this Friday only on code. That is possible.

bangerth:
Great, thank you!

source/fe/fe_poly_tensor.cc
konsim83: @bangerth The first part of the condition essentially says that when two faces with opposite normals meet then change the sign in the cell with the id that is higher. Does your comment mean that this is consistent? This is how I would understand that.

konsim83:
OK, update:
* I removed the fe_q test in dim=2. This test fails for higher order when the mesh is non-standard. This was not detected since it probably rarely happens but technically it could on embedded meshes like a Moebius strip. Tests for elements that inherit
* I also added tests in 3D. If a vector element that inherits from FE_PolyTensor<dim> is not listed in 3D one can safely assume it is broken on non-standard meshes
* I will squash and rebase on master tomorrow since I need to resolve a merge conflict in grid_generator.cc (nothing serious)

Should I open an issue for broken 2d elements that I discovered? This concerns FE_Q<2> and FE_NedelecSZ<2>

Cheers,
Konrad

konsim83:
@bangerth I added quite some files but mostly tests. Only minor changes in grid_generator.cc/h and fe_poly_tensor.cc make sure that RaviartThomas<2> and Nedelec<2>(0) will work correctly in 2D.

Thank you all for the patience and the code reviews. :-)

source/fe/fe_poly_tensor.cc
konsim83: After thinking about it: The tie breaker is correct :-)

konsim83:
@bangerth Is this still relevant for the new release?

kronbichler (approved):

konsim83:
Hi @kronbichler and @bangerth , sorry for the long delay. I did not notice there was a merge conflict with master. I resolved it and now everything should pass. Well, should....

tests/fe/fe_conformity_dim_2_fe_nedelec.cc
```diff
@@ -42,7 +42,7 @@ main(int, char **)
 
       constexpr int dim = 2;
 
-      for (unsigned int fe_degree = 0; fe_degree < 5; ++fe_degree)
+      for (unsigned int fe_degree = 0; fe_degree < 1; ++fe_degree)
```
drwells, line 45: Could we add a comment like https://github.com/dealii/dealii/pull/12016#issuecomment-817137541 here to remind ourselves that this is an issue?

include/deal.II/grid/grid_generator.h
```diff
@@ -779,29 +779,26 @@ namespace GridGenerator
                       const double        radius = 1.);
 
   /**
-   * Generate a 2D mesh consisting of the unit square joined with a copy shifted
-   * by $s = (1,0)$. Depending on the flags passed either the right or the left
-   * square is rotated by $\pi/2$. This way one can generate a mesh in which one
-   * square possibly contains an edge that has the opposite tangential (and
-   * hence also opposite normal) orientation of the neighboring edge of the
-   * other square.
+   * Generate a 2D mesh consisting of five squares aranged in a plus-shape.
```
kronbichler, line 782:
```suggestion
   * Generate a 2D mesh consisting of five squares arranged in a plus-shape.
```

konsim83:
OK, I installed clang-format-11.1 rebased on master, and squashed my commits. Anything else left?

kronbichler (approved):
Looks good to me.

include/deal.II/grid/grid_generator.h
konsim83: done

bangerth:
Thanks everyone, and my apologies for dropping the ball on this. I'm glad this finally got merged!

PR #12017, opened 2021-04-05
Rename member variables.

bangerth:
Related to #12008 and follow-up to #12009: There is a symmetry in the member variables names `data_begin` and `data_end` that is not actually matched by their names. More precisely, the former is a pointer owning the data pointed to, whereas the latter is just a pointed to the end of the currently used data area (not even to the end of the owned memory range -- that is what `allocated_end` does). This mismatch between member variable name and semantics is undesirable. This patch renames all three member variables to make things more easy to understand.

@tjhei in a comment of #12009 incidentally suggested this very action after I had already written this patch.

I would have liked to use the `data` name for the pointed to area, but that name is already taken by the `data()` member function. So I chose `elements`.

/rebuild

### DIFF
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -349,17 +349,17 @@ private:
   /**
    * Pointer to actual data array.
    */
-  std::unique_ptr<T[], void (*)(T *)> data_begin;
+  std::unique_ptr<T[], void (*)(T *)> elements;
 
   /**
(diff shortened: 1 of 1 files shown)


drwells (approved):

bangerth:
I have some follow-ups already ready, so wouldn't mind someone hitting the "Merge" button if they were so inclined :-)

masterleinad (approved):

PR #12018, opened 2021-04-06
More work on AlignedVector

bangerth:
Next update to #12009.

This PR has a bunch of commits, but really only the first one ("Simplify the logic...") has anything interesting. In the previous iteration, the function first swapped the old and new pointers, and then copied data. This looked funny because it copied from variables called "new" to variables called "old". It is easier to read to move the copying forward. That patch does this.

The remaining patches rename a bunch of variables in a few places and update documentation. They're all self-contained and so I think should remain as they are.

/rebuild

### DIFF
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -121,15 +121,29 @@ public:
   operator=(AlignedVector<T> &&vec) noexcept;
 
   /**
-   * Change the size of the vector. It keeps old elements previously available
-   * but does not initialize the newly allocated memory, leaving it in an
-   * undefined state.
+   * Change the size of the vector. If the new size is larger than the
+   * previous size, then new elements will be added to the end of the
+   * vector; these elements will remain uninitialized (i.e., left in
+   * an undefined state) if `std::is_trivial<T>` is `true`, and
+   * will be default initialized if `std::is_trivial<T>` is `false`.
+   * See [here](https://en.cppreference.com/w/cpp/types/is_trivial) for
+   * a definition of what `std::is_trivial` does.
+   *
+   * If the new size is less than the previous size, then the last few
+   * elements will be destroyed if `std::is_trivial<T>` will be `false`
+   * or will simply be ignored in the future if
+   * `std::is_trivial<T>` is `true`.
+   *
+   * As a consequence of the outline above, the "_fast" suffix of this
+   * function refers to the fact that for "trivial" classes `T`, this
+   * function omits constructor/destructor calls and in particular the
+   * initialization of new elements.
    *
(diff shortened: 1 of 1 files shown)


peterrum (approved):

include/deal.II/base/aligned_vector.h
```diff
@@ -121,15 +121,29 @@ class AlignedVector
   operator=(AlignedVector<T> &&vec) noexcept;
 
   /**
-   * Change the size of the vector. It keeps old elements previously available
-   * but does not initialize the newly allocated memory, leaving it in an
-   * undefined state.
+   * Change the size of the vector. If the new size is larger than the
+   * previous size, then new elements will be added to the end of the
+   * vector; these elements will remain uninitialized (i.e., left in
+   * an undefined state) if `std::is_trivial<T>` is `false`, and
```
kronbichler, line 127:
```suggestion
   * an undefined state) if `std::is_trivial<T>` is `true`, and
```

include/deal.II/base/aligned_vector.h
```diff
@@ -121,15 +121,29 @@ class AlignedVector
...
+   * will be default initialized if `std::is_trivial<T>` is `true`.
```
kronbichler, line 128:
```suggestion
   * will be default initialized if `std::is_trivial<T>` is `false`.
```

include/deal.II/base/aligned_vector.h
```diff
@@ -121,15 +121,29 @@ class AlignedVector
...
+   * See [here](https://en.cppreference.com/w/cpp/types/is_trivial) for
+   * a definition of what `std::is_trivial` does.
+   *
+   * If the new size is less than the previous size, then the last few
+   * elements will be destroyed if `std::is_trivial<T>` will be `true`
```
kronbichler, line 133:
```suggestion
   * elements will be destroyed if `std::is_trivial<T>` will be `false`
```

include/deal.II/base/aligned_vector.h
```diff
@@ -121,15 +121,29 @@ class AlignedVector
...
+   * or will simply be ignored in the future if
+   * `std::is_trivial<T>` is `false`.
```
kronbichler, line 135:
```suggestion
   * `std::is_trivial<T>` is `true`.
```

kronbichler (approved):
Looks good to me, except to what seems to be a mixup in what the `trivial` versus non-trivial data types are supposed to do.

bangerth:
Thanks, @kronbichler ! I got it wrong in the first place and then copy-pasted for all of the other places :-(

bangerth:
Related to #12008.

PR #12019, opened 2021-04-06
Break generation of graphical data into its own function in step-15.

bangerth:
This makes the `run()` function easier to read, and that is a good thing I believe.

/rebuild


masterleinad (approved):
Looks good to me.

Rombur (approved):

PR #12021, opened 2021-04-06
Play with include files

bangerth:
I ran into a circular header file issue when working on #12008 and needed to play with things. The solution is that `array_view.h` does not actually need to `#include` header files for all of the types it can wrap: If someone wants to call an `ArrayView` function with a specific type, then they will have created an object of that type, and will have the complete declaration of that type already. The header file only needs to know a forward declaration for these types if these types have a template argument because then every use of these types in the implementation of functions in `array_view.h` will be deferred due to ADL anyway.

/rebuild

### DIFF
--- a/include/deal.II/base/array_view.h
+++ b/include/deal.II/base/array_view.h
@@ -21,15 +21,19 @@
 #include <deal.II/base/exceptions.h>
 #include <deal.II/base/memory_space.h>
 #include <deal.II/base/symmetric_tensor.h>
-#include <deal.II/base/table.h>
 #include <deal.II/base/tensor.h>
-
-#include <deal.II/lac/lapack_full_matrix.h>
 
 #include <type_traits>
 #include <vector>
 
 DEAL_II_NAMESPACE_OPEN
+
+// Forward declaration
+template <int N, typename T>
+class Table;
+
+template <typename number>
+class LAPACKFullMatrix;
 
 
 /**
--- a/include/deal.II/base/incremental_function.h
+++ b/include/deal.II/base/incremental_function.h
@@ -22,6 +22,9 @@
 #include <deal.II/base/function.h>
 #include <deal.II/base/thread_management.h>
 
+#include <deal.II/lac/vector.h>
+
+
 DEAL_II_NAMESPACE_OPEN
 
 // Forward declaration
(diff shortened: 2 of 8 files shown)


masterleinad (approved):

masterleinad:
```
n file included from /jenkins/workspace/dealii_PR-12021/include/deal.II/lac/la_vector.h:26:0,

                 from /jenkins/workspace/dealii_PR-12021/include/deal.II/lac/la_vector.templates.h:21,

                 from /jenkins/workspace/dealii_PR-12021/source/lac/la_vector.cc:16:

/jenkins/workspace/dealii_PR-12021/include/deal.II/lac/read_write_vector.h:743:29: error: ‘parallel’ is not a member of ‘dealii’

     mutable std::shared_ptr<::dealii::parallel::internal::TBBPartitioner>

                             ^

/jenkins/workspace/dealii_PR-12021/include/deal.II/lac/read_write_vector.h:743:29: error: ‘parallel’ is not a member of ‘dealii’

/jenkins/workspace/dealii_PR-12021/include/deal.II/lac/read_write_vector.h:743:73: error: template argument 1 is invalid

     mutable std::shared_ptr<::dealii::parallel::internal::TBBPartitioner>
```

bangerth:
Yes, I found this too by now, along with a good number of failing tests. Let's see whether this iteration fixes this, but I have a suspicion that this will be a longer piece of work :-(

bangerth:
Ah, what do you. Not so much work after all! :-)

Rombur (approved):

peterrum:
As noted by @mschreter in https://github.com/kronbichler/adaflo/pull/42, this PR lead to some incompatibility. Maybe we should add an appropriate change-log entry?

bangerth:
I can add something, but I don't quite know what. In a sense, you @mschrete just fixed a pre-existing bug of sorts: If you use a certain class, you need to add the corresponding include file. I don't know what specific include chains I broke, and so it's difficult for me to write an entry that explains what will no longer work now...

peterrum:
In our case `#include <deal.II/lac/lapack_full_matrix.h>` was missing. I think @kronbichler normally writes a change-log entry if include files are removed from an include file.

PR #12022, opened 2021-04-07
Add AlignedVector::replicate_across_communicator()

bangerth:
First step towards implementing #12008. This first implementation just copies the data across all processes in a communicator, without using shared memory. I'm interested in feedback about the name and argument list for the moment.

Looking ahead, my plan for the next step is to extend the implementation by figuring out how to put the data into a shmem space. I *think* that the interface will be able to stay the same if I get this right.

/rebuild

### DIFF
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -301,6 +301,30 @@ public:
    * @note The use of shared memory between MPI processes requires
    *   that the detected MPI installation supports the necessary operations.
    *   This is the case for MPI 3.0 and higher.
+   *
+   * @note This function is not cheap. It needs to create sub-communicators
+   *   of the provided @p communicator object, which is generally an expensive
+   *   operation. Likewise, the generation of shared memory spaces is not
+   *   a cheap operation. As a consequence, this function primarily makes
+   *   sense when the goal is to share large read-only data tables among
+   *   processes; examples are data tables that are loaded at start-up
+   *   time and then used over the course of the run time of the program.
+   *   In such cases, the start-up cost of running this function can be
+   *   amortized over time, and the potential memory savings from not having to
+   *   store the table on each process may be substantial on machines with
+   *   large core counts on which many MPI processes run on the same machine.
+   *
+   * @note This function only makes sense if the data type `T` is
+   *   "self-contained", i.e., all if its information is stored in its
+   *   member variables, and if none of the member variables are pointers
+   *   to other parts of the memory. This is because if a type `T` does
+   *   have pointers to other parts of memory, then moving `T` into
+   *   a shared memory space does not result in the other processes having
+   *   access to data that the object points to with its member variable
+   *   pointers: These continue to live only on one process, and are
+   *   typically in memory areas not accessible to the other processes.
+   *   As a consequence, the usual use case for this function is to share
+   *   arrays of simple objects such as `double`s or `int`s.
    */
   void
   replicate_across_communicator(const MPI_Comm &   communicator,
@@ -422,7 +446,7 @@ private:
   /**
    * Pointer to actual data array.
    */
-  std::unique_ptr<T[], void (*)(T *)> elements;
+  std::unique_ptr<T[], std::function<void(T *)>> elements;
 
   /**
    * Pointer to one past the last valid value.
@@ -1094,7 +1118,7 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   // Create communicators for each group of processes that can use
   // shared memory areas. Within each of these groups, we don't care about
   // which rank each of the old processes gets except that we would like to
-  // make sure that the (global) root process will be have rank=0 within
+  // make sure that the (global) root process will have rank=0 within
   // its own sub-communicator. We can do that through the third argument of
   // MPI_Comm_split_type (the "key") which is an integer meant to indicate the
   // order of processes within the split communicators, and we will set it to
@@ -1140,7 +1164,7 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   // There are different ways of creating a "shmem_roots_communicator".
   // The conceptually easiest way is to create an MPI_Group that only
   // includes the shmem roots and then create a communicator from this
-  // by way of MPI_Comm_create or MPI_Comm_create_group. The problem
+  // via MPI_Comm_create or MPI_Comm_create_group. The problem
   // with this is that we would have to exchange among all processes
   // which ones are shmem roots and which are not. This is awkward.
   //
@@ -1151,15 +1175,15 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   // anything among themselves with the communicator so created.
   //
   // Using MPI_Comm_split has the additional benefit that, just as above,
-  // we can choose where each rank will end up in the shmem roots communicator.
+  // we can choose where each rank will end up in shmem_roots_communicator.
   // We would again like to set key=0 for the original root_process -- in other
   // word, we would then know that among the shmem roots, the original root has
   // rank=0. But, just like above, this doesn't appear to be working, so we put
-  // the origin root at the *end* of the shmem roots communicator.
+  // the origin root at the *end* of shmem_roots_communicator.
   //
   // In any case, this makes it easy to next determine which process among the
   // shmem roots is the one who initiates the broad cast operation mentioned
-  // above
+  // above.
   MPI_Comm shmem_roots_communicator;
   {
     const int key =
@@ -1183,35 +1207,273 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
 
   // Now let the original root_process broadcast the current object to all
   // shmem roots. We know that the last rank is the original root process that
-  // has all of the data
+  // has all of the data.
   if (is_shmem_root)
-    Utilities::MPI::broadcast(
-      shmem_roots_communicator,
-      *this,
-      Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1);
+    {
+      if (std::is_trivial<T>::value)
+        {
+          // The data is "trivial", i.e., we can copy things directly without
+          // having to go through the serialization/deserialization machinery of
+          // Utilities::MPI::broadcast.
+          //
+          // In that case, first tell all of the other shmem roots how many
+          // elements we will have to deal with, and let them resize their
+          // (non-shared) arrays.
+          const size_type new_size = Utilities::MPI::broadcast(
+            shmem_roots_communicator,
+            size(),
+            Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1);
+          if (Utilities::MPI::this_mpi_process(shmem_roots_communicator) !=
+              Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1)
+            resize(new_size);
+
+          // Then directly copy from the root process into these buffers
+          int ierr = MPI_Bcast(
+            elements.get(),
+            sizeof(T) * new_size,
+            MPI_CHAR,
+            Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1,
+            shmem_roots_communicator);
+          AssertThrowMPI(ierr);
+        }
+      else
+        {
+          // The objects to be sent around are not "trivial", and so we have
+          // to go through the serialization/deserialization machinery. On all
+          // but the sending process, overwrite the current state with the
+          // vector just broadcast.
+          if (Utilities::MPI::this_mpi_process(shmem_roots_communicator) ==
+              Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1)
+            Utilities::MPI::broadcast(
+              shmem_roots_communicator,
+              *this,
+              Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1);
+          else
+            *this = Utilities::MPI::broadcast(
+              shmem_roots_communicator,
+              *this,
+              Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1);
+        }
+    }
 
   // We no longer need the shmem roots communicator, so get rid of it
   {
     const int ierr = MPI_Comm_free(&shmem_roots_communicator);
     AssertThrowMPI(ierr);
   }
+
+
   // **** Step 3 ****
   // At this point, all shmem groups have one shmem root process that has
-  // a copy of the data. Let each of these shmem roots broadcast the
-  // data to the other processes in their shmem group. As mentioned above,
-  // we know that the shmem roots is the last rank in their respective
+  // a copy of the data. This is the point where each shmem group should
+  // establish a shmem area to put the data into. As mentioned above,
+  // we know that the shmem roots are the last rank in their respective
   // shmem_group_communicator.
-  *this = Utilities::MPI::broadcast(
+  //
+  // The process for all of this works as follows: While all processes in
+  // the shmem group participate in the generation of the shmem memory window,
+  // only the shmem root actually allocates any memory -- the rest just
+  // allocate zero bytes of their own. We allocate space for exactly
+  // size() elements (computed on the shmem_root that already has the data)
+  // and add however many bytes are necessary so that we know that we can align
+  // things to 64-byte boundaries. The worst case happens if the memory system
+  // gives us a pointer to an address one byte past a desired alignment
+  // boundary, and in that case aligning the memory will require us to waste the
+  // first (align_by-1) bytes. So we have to ask for
+  //   size() * sizeof(T) + align_by - 1
+  // bytes.
+  //
+  // Before MPI 4.0, there was no way to specify that we want memory aligned to
+  // a certain number of bytes. This is going to come back to bite us further
+  // down below when we try to get a properly aligned pointer to our memory
+  // region, see the commentary there. Starting with MPI 4.0, one can set a
+  // flag in an MPI_Info structure that requests a desired alignment, so we do
+  // this for forward compatibility; MPI implementations ignore flags they don't
+  // know anything about, and so setting this flag is backward compatible also
+  // to older MPI versions.
+  //
+  // There is one final piece we can already take care of here. At the beginning
+  // of all of this, only the shmem_root knows how many elements there are in
+  // the array. But at the end of it, all processes of course need to know. We
+  // could put this information somewhere into the shmem area, along with the
+  // other data, but that seems clumsy. It turns out that when calling
+  // MPI_Win_allocate_shared, we are asked for the value of a parameter called
+  // 'disp_unit' whose meaning is difficult to determine from the MPI
+  // documentation, and that we do not actually need. So we "abuse" it a bit: On
+  // the shmem root, we put the array size into it. Later on, the remaining
+  // processes can query the shmem root's value of 'disp_unit', and so will be
+  // able to learn about the array size that way.
+  MPI_Win        shmem_window;
+  void *         base_ptr;
+  const MPI_Aint align_by   = 64;
+  const MPI_Aint alloc_size = Utilities::MPI::broadcast(
     shmem_group_communicator,
-    *this,
+    (size() * sizeof(T) + align_by - 1),
     Utilities::MPI::n_mpi_processes(shmem_group_communicator) - 1);
 
-  // We now also no longer need the shmem group communicators, so get rid of
-  // them
   {
-    const int ierr = MPI_Comm_free(&shmem_group_communicator);
+    const int disp_unit = (is_shmem_root ? size() : 1);
+
+    int ierr;
+
+    MPI_Info mpi_info;
+    ierr = MPI_Info_create(&mpi_info);
+    AssertThrowMPI(ierr);
+    ierr = MPI_Info_set(mpi_info,
+                        "mpi_minimum_memory_alignment",
+                        std::to_string(align_by).c_str());
+    AssertThrowMPI(ierr);
+    ierr = MPI_Win_allocate_shared((is_shmem_root ? alloc_size : 0),
+                                   disp_unit,
+                                   mpi_info,
+                                   shmem_group_communicator,
+                                   &base_ptr,
+                                   &shmem_window);
+    AssertThrowMPI(ierr);
+
+    ierr = MPI_Info_free(&mpi_info);
     AssertThrowMPI(ierr);
   }
+
+
+  // **** Step 4 ****
+  // The next step is to teach all non-shmem root processes what the pointer to
+  // the array is that the shmem-root created. MPI has a nifty way for this
+  // given that only a single process actually allocated memory in the window:
+  // When calling MPI_Win_shared_query, the MPI documentation says that
+  // "When rank is MPI_PROC_NULL, the pointer, disp_unit, and size returned are
+  // the pointer, disp_unit, and size of the memory segment belonging the lowest
+  // rank that specified size > 0. If all processes in the group attached to the
+  // window specified size = 0, then the call returns size = 0 and a baseptr as
+  // if MPI_ALLOC_MEM was called with size = 0."
+  //
+  // This will allow us to obtain the pointer to the shmem root's memory area,
+  // which is the only one we care about. (None of the other processes have
+  // even allocated any memory.) But this will also retrieve the shmem root's
+  // disp_unit, which in step 3 above we have abused to pass along the number of
+  // elements in the array.
+  //
+  // We don't need to do this on the shmem root process: This process has
+  // already gotten its base_ptr correctly set above, and we can determine the
+  // array size by just calling size().
+  unsigned int array_size =
+    (is_shmem_root ? size() : numbers::invalid_unsigned_int);
+  if (is_shmem_root == false)
+    {
+      int       disp_unit;
+      MPI_Aint  alloc_size; // not actually used
+      const int ierr = MPI_Win_shared_query(
+        shmem_window, MPI_PROC_NULL, &alloc_size, &disp_unit, &base_ptr);
+      AssertThrowMPI(ierr);
+
+      // Make sure we actually got a pointer, and also unpack the array size as
+      // discussed above.
+      Assert(base_ptr != nullptr, ExcInternalError());
+
+      array_size = disp_unit;
+    }
+
+
+  // **** Step 5 ****
+  // Now that all processes know the address of the space that is visible to
+  // everyone, we need to figure out whether it is properly aligned and if not,
+  // find the next aligned address.
+  //
+  // std::align does that, but it also modifies its last two arguments. The
+  // documentation of that function at
+  // https://en.cppreference.com/w/cpp/memory/align is not entirely clear, but I
+  // *think* that the following should do given that we do not use base_ptr and
+  // available_space any further after the call to std::align.
+  std::size_t available_space       = alloc_size;
+  void *      base_ptr_backup       = base_ptr;
+  T *         aligned_shmem_pointer = static_cast<T *>(
+    std::align(align_by, array_size * sizeof(T), base_ptr, available_space));
+  Assert(aligned_shmem_pointer != nullptr, ExcInternalError());
+
+  // There is one step to guard against. It is *conceivable* that the base_ptr
+  // we have previously obtained from MPI_Win_shared_query is mapped so
+  // awkwardly into the different MPI processes' memory spaces that it is
+  // aligned in one memory space, but not another. In that case, different
+  // processes would align base_ptr differently, and adjust available_space
+  // differently. We can check that by making sure that the max (or min) over
+  // all processes is equal to every process's value. If that's not the case,
+  // then the whole idea of aligning above is wrong and we need to rethink what
+  // it means to align data in a shared memory space.
+  //
+  // One might be tempted to think that this is not how MPI implementations
+  // actually arrange things. Alas, when developing this functionality in 2021,
+  // this is really how at least OpenMPI ends up doing things. (This is with an
+  // OpenMPI implementation of MPI 3.1, so it does not support the flag we set
+  // in the MPI_Info structure above when allocating the memory window.) Indeed,
+  // when running this code on three processes, one ends up with base_ptr values
+  // of
+  //     base_ptr=0x7f0842f02108
+  //     base_ptr=0x7fc0a47881d0
+  //     base_ptr=0x7f64872db108
+  // which, most annoyingly, are aligned to 8 and 16 byte boundaries -- so there
+  // is no common offset std::align could find that leads to a 64-byte
+  // aligned memory address in all three memory spaces. That's a tremendous
+  // nuisance and there is really nothing we can do about this other than just
+  // fall back on the (unaligned) base_ptr in that case.
+  if (Utilities::MPI::min(available_space, shmem_group_communicator) !=
+      Utilities::MPI::max(available_space, shmem_group_communicator))
+    aligned_shmem_pointer = static_cast<T *>(base_ptr_backup);
+
+
+  // **** Step 6 ****
+  // If this is the shmem root process, we need to copy the data into the
+  // shared memory space.
+  if (is_shmem_root)
+    {
+      if (std::is_trivial<T>::value == true)
+        std::memcpy(aligned_shmem_pointer, elements.get(), sizeof(T) * size());
+      else
+        for (std::size_t i = 0; i < size(); ++i)
+          new (&aligned_shmem_pointer[i]) T(std::move(elements[i]));
+    }
+
+  // **** Step 7 ****
+  // Finally, we need to set the pointers of this object to what we just
+  // learned. This also releases all memory that may have been in use
+  // previously.
+  //
+  // The part that is a bit tricky is how to write the deleter of this
+  // shared memory object. When we want to get rid of it, we need to
+  // also release the MPI_Win object along with the shmem_group_communicator
+  // object. That's because as long as we use the shared memory, we still need
+  // to hold on to the MPI_Win object, and the MPI_Win object is based on the
+  // communicator. (The former is definitely true, the latter is not quite clear
+  // from the MPI documentation, but seems reasonable.) So we need to have a
+  // deleter for the pointer that ensures that upon release of the memory, we
+  // not only call the destructor of these memory elements (but only once, on
+  // the shmem root!) but also destroy the MPI_Win and the communicator. All of
+  // that is encapsulated in the following call where the deleter makes copies
+  // of the arguments in the lambda capture.
+  elements =
+    decltype(elements)(aligned_shmem_pointer,
+                       [is_shmem_root,
+                        array_size,
+                        aligned_shmem_pointer,
+                        shmem_group_communicator,
+                        shmem_window](T *) mutable {
+                         if (is_shmem_root)
+                           for (unsigned int i = 0; i < array_size; ++i)
+                             aligned_shmem_pointer[i].~T();
+
+                         int ierr;
+                         ierr = MPI_Win_free(&shmem_window);
+                         AssertThrowMPI(ierr);
+
+                         ierr = MPI_Comm_free(&shmem_group_communicator);
+                         AssertThrowMPI(ierr);
+                       });
+
+  // We then also have to set the other two pointers that define the state of
+  // the current object. Note that the new buffer size is exactly as large as
+  // necessary, i.e., can store size() elements, regardless of the number of
+  // allocated elements in the original objects.
+  used_elements_end      = elements.get() + array_size;
+  allocated_elements_end = used_elements_end;
 
   // **** Consistency check ****
   // At this point, each process should have a copy of the data.
@@ -1356,7 +1618,7 @@ template <class Archive>
 inline void
 AlignedVector<T>::save(Archive &ar, const unsigned int) const
 {
-  size_type vec_size(size());
+  size_type vec_size = size();
   ar &      vec_size;
   if (vec_size > 0)
     ar &boost::serialization::make_array(elements.get(), vec_size);
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -325,6 +325,23 @@ public:
    *   typically in memory areas not accessible to the other processes.
    *   As a consequence, the usual use case for this function is to share
    *   arrays of simple objects such as `double`s or `int`s.
+   *
+   * @note After calling this function, objects on different MPI processes
+   *   share a common state. That means that certain operations become
+   *   "collective", i.e., they must be called on all participating
+   *   processors at the same time. In particular, you can no longer call
+   *   resize(), reserve(), or clear() on one MPI process -- you have to do
+   *   so on all processes at the same time, because they have to communicate
+   *   for these operations. If you do not do so, you will likely get
+   *   a deadlock that may be difficult to debug. By extension, this rule of
+   *   only collectively resizing extends to this function itself: You can
+   *   not call it twice in a row because that implies that first all but the
+   *   `root_process` throw away their data, which is not a collective
+   *   operation. Generally, these restrictions on what can and can not be
+   *   done hint at the correctness of the comments above: You should treat
+   *   an AlignedVector on which the current function has been called as
+   *   `const`, on which no further operations can be performed until
+   *   the destructor is called.
    */
   void
   replicate_across_communicator(const MPI_Comm &   communicator,
@@ -1114,6 +1131,17 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
 {
 #  ifdef DEAL_II_WITH_MPI
 #    if DEAL_II_MPI_VERSION_GTE(3, 0)
+
+  // **** Step 0 ****
+  // All but the root process no longer need their data, so release the memory
+  // used to store the previous elements.
+  if (Utilities::MPI::this_mpi_process(communicator) != root_process)
+    {
+      elements.reset();
+      used_elements_end      = nullptr;
+      allocated_elements_end = nullptr;
+    }
+
   // **** Step 1 ****
   // Create communicators for each group of processes that can use
   // shared memory areas. Within each of these groups, we don't care about
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -265,7 +265,11 @@ public:
    * @p root_process across all processes of the MPI communicator. The current
    * state found on any of the processes other than @p root_process is lost
    * in this process. One can imagine this operation to act like a call to
-   * Utilities::MPI::broadcast() from the root process to all other processes.
+   * Utilities::MPI::broadcast() from the root process to all other processes,
+   * though in practice the function may try to move the data into shared
+   * memory regions on each of the machines that host MPI processes and
+   * let all MPI processes on this machine then access this shared memory
+   * region instead of keeping their own copy.
    *
    * The intent of this function is to quickly exchange large arrays from
    * one process to others, rather than having to compute or create it on
@@ -273,6 +277,12 @@ public:
    * disk -- say, large data tables -- that are more easily dealt with by
    * reading once and then distributing across all processes in an MPI
    * universe, than letting each process read the data from disk itself.
+   * Specifically, the use of shared memory regions allows for replicating
+   * the data only once per multicore machine in the MPI universe, rather
+   * than replicating data once for each MPI process. This results in
+   * large memory savings if the data is large on today's machines that
+   * can easily house several dozen MPI processes per shared memory
+   * space.
    *
    * This function does not imply a model of keeping data on different processes
    * in sync, as parallel::distributed::Vector and other vector classes do where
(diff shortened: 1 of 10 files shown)


peterrum:
> I think that the interface will be able to stay the same if I get this right.

Could you elaborate this comment. I don't see this to be the case.

bangerth:
@peterrum Instead of trying to explain what I'm thinking, I've just pushed another commit that provides the outline of how that function will look like. This function does (I believe) work correctly. I will need to adjust the part in "step 3" so that objects can share their data via shmem, but maybe the outline explains where I'm going.

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1083,162 @@ AlignedVector<T>::fill(const T &value)
...
+    const int key =
+      (Utilities::MPI::this_mpi_process(communicator) == root_process ?
+         Utilities::MPI::n_mpi_processes(communicator) :
+         0);
+    const int ierr = MPI_Comm_split_type(communicator,
+                                         MPI_COMM_TYPE_SHARED,
+                                         key,
+                                         MPI_INFO_NULL,
+                                         &shmem_group_communicator);
+    AssertThrowMPI(ierr);
```
peterrum, line 1121: I have not a good experience with creating sub-communicators internally and throwing away them at the end of the function call. In hyper.deal, we used to do this so that we do not need to pass around the communicator throughput the code, but it turned out that creating it is too expensive for more than 1000 nodes on SuperMUC-NG (1-10 seconds - I don't remember the exact numbers). Now, we create all communicators in one place: https://github.com/hyperdeal/hyperdeal/blob/60091feb59f98789bc082395150c7537f6b3dc07/examples/advection/advection.cc#L140-L181

Maybe, there are some useful functions in https://github.com/hyperdeal/hyperdeal/blob/60091feb59f98789bc082395150c7537f6b3dc07/source/base/mpi.cc#L32-L83 that we could port to deal.II.

peterrum: Also this communicator needs - as far as I know - to survive as long as the data should be accessible in shmem.

bangerth: Yes, I suspect once I implement the actual shmem stuff, this communicator will need to survive until we release the memory.

As for creating communicators:
* First, I read up on `MPI_Comm_create` vs `MPI_Comm_create_group` and that the latter may be expensive. So I chose the former. But in general, this is going to be an expensive function one way or the other. I will document that that is a fact and that the function should not be used in inner loops -- but that of course was never the intent. I also don't think that there is a way around it: You just *have* to create a subcommunicator if you want to use shmem.
* Let's also remember the *intent* of what I'm trying to do here. The goal is not to create a small shared memory representation, but to replicate huge data sets that one just doesn't want to keep on every MPI process -- think gigabyte-sized data. I think for these, an expensive communicator replication is acceptable.

bangerth:
OK, I've completed the implementation now, including shared memory. It takes about 270 lines to do all of this, but 2/3 of it are probably comments. That was fun!

I will point out that this had to change the deleter type of the pointer in `AlignedVector`, as already discussed in #12008. Other than that, there is no change in what the class already does, just a new function.

tjhei:
This looks quite neat already. 

I am surprised that you went with a design that replicates existing data instead of just allocating shared data (and let the user fill in useful information afterwards). This way you have a large chunk in memory twice for some period of time. Any advantage of your approach? This could just be a constructor/reinit() function...

peterrum:
> I am surprised that you went with a design that replicates existing data instead of just allocating shared data (and let the user fill in useful information afterwards). This way you have a large chunk in memory twice for some period of time. Any advantage of your approach? This could just be a constructor/reinit() function...

I agree with @tjhei's comment. Furthermore, let's step one step back and think about how the new feature of `AlignedVector` will be used. If `AlignedVector` is used by itself one would only set a size of the `AlignedVector` on the root process of the shared-memory domain and then one would call `AlignedVector::replicate_across_communicator()`. We still have the issue here that we need twice as much memory. But how does the user know which process is allowed to allocate memory and set the values? Does the user has to create a subcommunicator on his own? How is is guaranteed that `AlignedVector` and the user selects the same root process!?

Now, let's consider that we want to use the new feature indirectly via `Table`. The constructor and the `reinit()`-function of `Table` take the size of the table, which is translated into the length of the `AlignedVector`. When should we call `replicate_across_communicator()`? If we don't call it during the `reinit()`-function, this means that every process will allocate memory, which is definitely too much.

@bangerth How would you tackle these issues. Maybe you have already given thought to them!?


Regarding https://github.com/dealii/dealii/issues/12008#issuecomment-817240454, I don't mind the std::function approach.

bangerth:
The way I envisioned this was as follows: I imagine these replicated arrays/tables to be treated as read-only because the semantics of sharing are so difficult -- users generally don't know which processes participate in shared memory arenas.

So the workflow I had in mind goes like this, in a parallel program:
```
  Table<2,double> data;
  const unsigned int root_process = 0;    // but could be any other process as well
  if (Utilities::MPI::this_mpi_process(communicator) == root_process)
  {
    data.reinit (2000,2000);
    ...read the data from file somehow, or get it from elsewhere...
  }
  data.replicate_across_communicator (communicator, root_process);  // to be implemented; will simply call the current function
```
That last step makes the data available to all processes and, conceptually, turns it into a read-only table. Maybe I should show this kind of example in the documentation of the function as well.

It is true, as has been pointed out above, that there is a point in time where we use twice the necessary amount of data. My design was chosen based on the following considerations:
* It is only twice of what is necessary on one process; on all of the other processes, it is exactly the minimum necessary, and in the grand scheme of things, it is only twice the single-process minimum, whereas on today's machines with N cores, you need N times the minimum right now. That's still a large saving.
* It hides almost all of the complexity from the user. If I had generated a constructor or `reinit()` function that does this, then you would have to (i) return to the user information that indicates which processes have to fill the shared memory areas (one root process per shmem arena) which most people would probably implement by letting these processes read from file rather than reading on only one process and then determining which other processes the data should be sent to; (ii) you lose the nice distinction that semantically once `replicate_across_communicator()`, the data should be treated as read-only.
* There is exactly one function that needs to know anything about MPI, communicators, etc. The rest of the interface remains agnostic about all of this.

masterleinad:
The indentation is off:
```
diff --git a/include/deal.II/base/aligned_vector.h b/include/deal.II/base/aligned_vector.h
index 0340242..f40b795 100644
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -1115,14 +1115,15 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm     communicator,
 #  ifdef DEAL_II_WITH_MPI
 #    if DEAL_II_MPI_VERSION_GTE(3, 0)
 
-    // **** Step 0 ****
-    // All but the root process no longer need their data, so release the memory used to store the previous elements.
-    if (Utilities::MPI::this_mpi_process(communicator) != root_process)
-      {
+  // **** Step 0 ****
+  // All but the root process no longer need their data, so release the memory
+  // used to store the previous elements.
+  if (Utilities::MPI::this_mpi_process(communicator) != root_process)
+    {
       elements.reset();
-      used_elements_end = nullptr;
+      used_elements_end      = nullptr;
       allocated_elements_end = nullptr;
-      }
+    }
 
   // **** Step 1 ****
   // Create communicators for each group of processes that can use
```

bangerth:
Fixed.

Of all of the things currently pending, this is one I'd really like to get into 9.3 because it is functionality we need in ASPECT. I would appreciate if we could come to some sort of conclusion of how to move forward...

tjhei:
> So the workflow I had in mind goes like this, in a parallel program:

I did not get the idea to only set the size on ``root`` from the implementation/documentation. I think this makes sense. Not as flexible as what I was originally thinking, but easy to use.

Let me go through the PR one more time...

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1107,302 @@ AlignedVector<T>::fill(const T &value)
...
+      array_size = disp_unit;
+    }
+
+
+  // **** Step 5 ****
+  // Now that all processes know the address of the space that is visible to
+  // everyone, we need to figure out whether it is properly aligned and if not,
+  // find the next aligned address
+
+  // TODO: Align the base_ptr via std::align
```
tjhei, line 1319: is this still missing?

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1107,302 @@ AlignedVector<T>::fill(const T &value)
...
+  T *aligned_shmem_pointer = base_ptr;
+
+
+  // **** Step 6 ****
+  // If this is the shmem root process, we need to copy the data into the
+  // shared memory space.
+  if (is_shmem_root)
+    {
+      if (std::is_trivial<T>::value == true)
+        std::memcpy(static_cast<void *>(aligned_shmem_pointer),
```
tjhei, line 1329: are these casts necessary?

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1107,302 @@ AlignedVector<T>::fill(const T &value)
...
+      Assert(Utilities::MPI::this_mpi_process(shmem_roots_communicator) ==
+               Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1,
+             ExcInternalError());
+  }
+
+  // Now let the original root_process broadcast the current object to all
+  // shmem roots. We know that the last rank is the original root process that
+  // has all of the data
+  if (is_shmem_root)
+    Utilities::MPI::broadcast(
```
tjhei, line 1223: doesn't this need a ``*this = Utilities::MPI::broadcast(...`` ?

tjhei (commented):
I left a bunch of questions and comments. Overall, you convinced me that the design is good as is. If we need more control later on, we can always add a different constructor/replicate function.
A few final questions/comments:
- Some of the tests seem quite redundant. Minimizing the number of new tests seems sensible to me (compile time).
- I am not sure that the alignment works correctly right now (and if the pointers need to be adjusted)
- Did you test that this whole thing works correctly on >1 node (see the broadcast comment)?

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1107,302 @@ AlignedVector<T>::fill(const T &value)
...
+      shmem_roots_communicator,
+      *this,
```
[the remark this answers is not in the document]
bangerth: Yes, I looked into this and found the same. I would like to do the optimization inside `Utilities::MPI::broadcast` if possible. Right now, `Utilities::MPI::broadcast` calls `Utilities::pack()` on the aligned vector and sends this packed data across the network. That can certainly be improved. Let me think about it, but if you allow me to, I'd like to get the general structure in place first.

As for your second question: For trivial data types, I could in principle do as you suggest and just let MPI unpack everything into the destination buffer. I think I know how to do as you suggest, but as above, let me do this in a follow-up.

include/deal.II/base/aligned_vector.h
bangerth: Yes. I just don't quite know how to do it. I have to admit that I don't quite understand the syntax of `std::align.`

include/deal.II/base/aligned_vector.h
bangerth: I copied this from someone else's code I believe where there was a comment about this being necessary to avoid warnings. I can try to remove this.

include/deal.II/base/aligned_vector.h
bangerth: Oh, you're totally right. This only ever worked because I was on a single shmem domain and the shmem_communicator has only one process! Good find!

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1107,302 @@ AlignedVector<T>::fill(const T &value)
...
+    }
+
+
+  // **** Step 5 ****
+  // Now that all processes know the address of the space that is visible to
+  // everyone, we need to figure out whether it is properly aligned and if not,
+  // find the next aligned address
+
+  // TODO: Align the base_ptr via std::align
+  T *aligned_shmem_pointer = base_ptr;
```
masterleinad, line 1320:
```suggestion
  std::size_t dummy = -1;
  T *aligned_shmem_pointer = std::align(align_by, sizeof(T) * size(), base_ptr, dummy);
```
?

bangerth:
@tjhei: I addressed the majority of your points. Specifically:

* I now avoid the packing/unpacking if the data is "trivial", but I'm not yet putting things right into their final location. I will do that at a later point if you allow me to.
* I've also removed one of the tests that calls the function twice in a row. That turned out to be quite difficult to get right, and I decided to just document that this isn't allowed. One runs into all sorts of funny deadlocks.
* As for the alignment: I'm leaving this as a TODO for now as well. I'll address this as a follow-up, if necessary with an assertion.

As for testing: Funny as that may sound, I don't actually have convenient access to a system that does not have a shared memory space. I have to admit that I don't currently have a log-in to an actual cluster (but need to remind @marcfehling to tell me how to log into the XSEDE machine). Would one of you all be willing to run one of the new tests with a sufficiently large number of MPI ranks?

bangerth:
@masterleinad and @tjhei:
I tried to use `std::align`. One of the issues I can't quite figure out is what would happen if `base_ptr` is already properly aligned in one address space, but not the other. I don't see a solution for this issue.

But more concretely, I tried the following piece of code:
```
diff --git a/include/deal.II/base/aligned_vector.h b/include/deal.II/base/aligned_vector.h
index b93e48bdf4..d900052a98 100644
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -1315,14 +1315,14 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   // processes can query the shmem root's value of 'disp_unit', and so will be
   // able to learn about the array size that way.
   MPI_Win        shmem_window;
-  T *            base_ptr;
-  const MPI_Aint align_by = 64;
+  void *         base_ptr;
+  const MPI_Aint align_by   = 64;
+  const MPI_Aint alloc_size = Utilities::MPI::broadcast(
+      shmem_group_communicator,
+          ((size() * sizeof(T) + align_by - 1) / sizeof(T)) * sizeof(T),
+    Utilities::MPI::n_mpi_processes(shmem_group_communicator) - 1);
 
   {
-    const MPI_Aint alloc_size =
-      (is_shmem_root ?
-         ((size() * sizeof(T) + align_by - 1) / sizeof(T)) * sizeof(T) :
-         0);
     const int disp_unit = (is_shmem_root ? size() : 1);
 
     const int ierr = MPI_Win_allocate_shared(alloc_size,
@@ -1372,10 +1372,37 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   // **** Step 5 ****
   // Now that all processes know the address of the space that is visible to
   // everyone, we need to figure out whether it is properly aligned and if not,
-  // find the next aligned address
-
-  // TODO: Align the base_ptr via std::align
-  T *aligned_shmem_pointer = base_ptr;
+  // find the next aligned address.
+  //
+  // std::align does that, but it also modifies its last two arguments. The
+  // documentation of that function at
+  // https://en.cppreference.com/w/cpp/memory/align is not entirely clear, but I
+  // *think* that the following should do given that we do not use base_ptr and
+  // available_space any further after the call to std::align.
+  std::size_t available_space = alloc_size;
+  std::cout << "align_by=" << align_by << ", alloc_size=" << alloc_size
+            << ", base_ptr=" << base_ptr
+            << ", available_space=" << available_space << std::endl;
+  T *aligned_shmem_pointer = static_cast<T *>(
+    std::align(align_by, alloc_size, base_ptr, available_space));
+  Assert(aligned_shmem_pointer != nullptr,
+         ExcMessage("The shared memory region could not be properly aligned. "
+                    "This should not happen. There is nothing you can "
+                    "do about this."));
+
+  // There is one step to guard against. It is *conceivable*, though probably
+  // not how MPI implementations actually arrange things, that the base_ptr we
+  // have previously obtained from MPI_Win_shared_query is mapped so awkwardly
+  // into the different MPI processes' memory spaces that it is aligned in one
+  // memory space, but not another. In that case, different processes would
+  // align base_ptr differently, and adjust available_space differently. We can
+  // check that by making sure that the max (or min) over all processes is equal
+  // to every process's value. If that's not the case, then the whole idea of
+  // aligning above is wrong and we need to rethink what it means to align data
+  // in a shared memory space.
+  Assert(Utilities::MPI::max(available_space, shmem_group_communicator) ==
+           available_space,
+         ExcInternalError());
 
 
   // **** Step 6 ****
```
The problem is that the first of the two assertions triggers. I get output like this:
```
23: align_by=64, alloc_size=100, base_ptr=0x7f909be17108, available_space=100
23: align_by=64, alloc_size=100, base_ptr=0x7f442f86f1d0, available_space=100
23: align_by=64, alloc_size=100, base_ptr=0x7f3ff389d108, available_space=100
```
So what I see here is that indeed `base_ptr` is 8-byte aligned on one machine and 16-byte aligned on another. The issue is that that means that the offsets into the buffer are different in different process spaces, and I cannot get things 64-byte aligned in all processes. Or am I reading this wrong?

I have to admit that I'm pretty confused by this. My intuition would have been that MPI asks the OS for a whole page to create a shmem window, and that the window's base address is physically maximally aligned (4k these days) and that that extends to the actual base ptr addresses we get -- or at least that they are all a constant offset from the window's base address. Apparently this is not the case. I don't really know what to do here.

tjhei:
Tricky. I found that there is some conversation about memory alignment in MPI implementation if you search for the hint mpi_minimum_memory_alignment. It might require a newer MPI version:
https://github.com/mpi-forum/mpi-issues/issues/121
https://github.com/pmodels/mpich/pull/4901

bangerth:
So what should we do here? Proposal:
* I leave the alignment code in, but if the uniform alignment fails, we fall back to the `base_ptr` -- which in that case just doesn't have the correct alignment.
* I add the necessary flag to the `MPI_Info` structure -- it will just have no effect on current MPI versions and there is no way for me to test any effects.
Does that sound reasonable?

bangerth:
That's what that last commit does. Feedback appreciated!

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1124,405 @@ AlignedVector<T>::fill(const T &value)
...
+  // documentation, and that we do not actually need. So we "abuse" it a bit: On
+  // the shmem root, we put the array size into it. Later on, the remaining
+  // processes can query the shmem root's value of 'disp_unit', and so will be
+  // able to learn about the array size that way.
+  MPI_Win        shmem_window;
+  void *         base_ptr;
+  const MPI_Aint align_by   = 64;
+  const MPI_Aint alloc_size = Utilities::MPI::broadcast(
+    shmem_group_communicator,
+    ((size() * sizeof(T) + align_by - 1) / sizeof(T)) * sizeof(T),
```
masterleinad, line 1333: Can you comment on this choice some more? We need at least `size()*sizeof(T)`. If want a multiple of `align_by` shouldn't this be
```
((size() * sizeof(T) + align_by - 1) / align_by * align_by,
```
?

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1124,405 @@ AlignedVector<T>::fill(const T &value)
...
+  // given that only a single process actually allocated memory in the window:
+  // When calling MPI_Win_shared_query, the MPI documentation says that
+  // "When rank is MPI_PROC_NULL, the pointer, disp_unit, and size returned are
+  // the pointer, disp_unit, and size of the memory segment belonging the lowest
+  // rank that specified size > 0. If all processes in the group attached to the
+  // window specified size = 0, then the call returns size = 0 and a baseptr as
+  // if MPI_ALLOC_MEM was called with size = 0."
+  //
+  // This will allow us to obtain the pointer to the shmem root's memory area,
+  // which is the only we one we care about. (None of the other processes have
```
masterleinad, line 1373:
```suggestion
  // which is the only one we care about. (None of the other processes have
```

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1124,405 @@ AlignedVector<T>::fill(const T &value)
...
+  // std::align does that, but it also modifies its last two arguments. The
+  // documentation of that function at
+  // https://en.cppreference.com/w/cpp/memory/align is not entirely clear, but I
+  // *think* that the following should do given that we do not use base_ptr and
+  // available_space any further after the call to std::align.
+  std::size_t available_space       = alloc_size;
+  void *      base_ptr_backup       = base_ptr;
+  T *         aligned_shmem_pointer = static_cast<T *>(
+    std::align(align_by, array_size * sizeof(T), base_ptr, available_space));
+  Assert(aligned_shmem_pointer != nullptr, ExcInternalError());
```
masterleinad, line 1413: Hmmm... I would think that the allocation is actually shifted so using `base_ptr_backup` would potentially point to something wrong if `std::align` actually does something.

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1124,405 @@ AlignedVector<T>::fill(const T &value)
...
+    aligned_shmem_pointer = static_cast<T *>(base_ptr_backup);
+
+
+  // **** Step 6 ****
+  // If this is the shmem root process, we need to copy the data into the
+  // shared memory space.
+  if (is_shmem_root)
+    {
+      if (std::is_trivial<T>::value == true)
+        std::memcpy(aligned_shmem_pointer, elements.get(), sizeof(T) * size());
```
masterleinad, line 1451: Can't we broadcast into this address instead of using a separate buffer first and then do the memcpy?

tjhei: I asked the same earlier. I think we should let this go and optimize in a follow up. I want this feature in 9 .3 and optimizations can wait.

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1124,405 @@ AlignedVector<T>::fill(const T &value)
...
+  // order of processes within the split communicators, and we will set it to
+  // zero for the root processes and one for all others -- which means that
+  // for all of these other processes, MPI can choose whatever order it
+  // wants because they have the same key (MPI then documents that these ties
+  // will be broken according to these processes' rank in the old group).
+  //
+  // At least that's the theory. In practice, the MPI implementation where
+  // this function was developed on does not seem to do that. But it
+  // is willing to put the root process onto the *last* rank of the
+  // communicator.
```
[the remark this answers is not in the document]
bangerth: The problem is that I don't like what I'm doing. It is common practice to make process zero special, whereas I am making process N-1 special. It is certainly conceivable that someone (myself included) gets back at that code and changes things to zero if I don't explain why we didn't choose zero (or one, or in fact any other choice).

There is also the issue that I'm not 100% sure that what I'm doing actually works in all situations (but I have assertions that check things). So I wanted to document not just *what* the function does, but also *why* it does things the way it does things.

include/deal.II/base/aligned_vector.h
bangerth: Can you elaborate? I *think* what you're saying is that `std::align` actually shifts data around (from `base_ptr` to the aligned address), but I don't think that is what `std::align` does. My reading is that it really just aligns the *pointer*, but doesn't touch data. Regardless, we haven't actually put any data into the memory region, so that is not a concern.

Or am I completely misreading what you're saying?

include/deal.II/base/aligned_vector.h
bangerth: Yes, you are right. I copied this line from @peterrum somewhere without thinking about it.

bangerth: Thanks for pointing it out -- I should really have thought about this more!

bangerth: Well, actually still not. This actually fails right away. What your formula does is find the smallest multiple of `align_by` bytes so that we can fit `size()*sizeof(T)` bytes into it. But that's not what we are after. What we need is this: What is the smallest size of a buffer (however that may be aligned) so that we are guaranteed that we can put `size()*sizeof(T)` bytes onto an aligned address?

The answer is `(size() * sizeof(T) + align_by - 1)`, which happens if the memory system gives us a pointer to an address that is one byte past an alignment boundary. In that case, we need to waste the first `align_by-1` bytes.

bangerth:
@masterleinad: Thanks very much for the careful reading! I've incorporated everything discussed above and reduced the number of commits to what I think is reasonable.

include/deal.II/base/aligned_vector.h
masterleinad: Yes, I agree that it only changes the value of the pointer and you're right that we don't have any data we should worry about at this point.

include/deal.II/base/aligned_vector.h
masterleinad: That sounds reasonable. :slightly_smiling_face:

masterleinad (approved):
Looks OK to me.

bangerth:
@tjhei?

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1124,411 @@ AlignedVector<T>::fill(const T &value)
...
+                                    /*color=*/
+                                    (is_shmem_root ? 0 : 1),
+                                    key,
+                                    &shmem_roots_communicator);
+    AssertThrowMPI(ierr);
+
+    // Again verify the explanation from above
+    if (Utilities::MPI::this_mpi_process(communicator) == root_process)
+      Assert(Utilities::MPI::this_mpi_process(shmem_roots_communicator) ==
+               Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1,
```
tamiko, line 1232: Creating to variables `this_mpi_process` and `n_mpi_processes` at the beginning of the function with one call to `Utilities::MPI::this_mpi_process` and `Utilities::MPI::n_mpi_processes` each would make all of the arguments of the if statements fit in one line. This would make it significantly easier to read. We would also avoid potential repeated calls to these utility functions.

tjhei:
It is awkward that aligned vector might contain unaligned data now. Would that silently produce wrong results in instructions that require it? Otherwise, failing might be a better option.
What routines require aligned data?

Looks good to me otherwise.

include/deal.II/base/aligned_vector.h
bangerth: I disagree. I have three communicators floating around, and would need 6 variables, all with long names :-)

bangerth:
I guess that's a question for @kronbichler . I don't know. I think the only instructions that might require long alignment would be some vectorized load/store instructions -- so if we had `AlignedVector<VectorizedArray<...>>`. I don't know whether we actually ever do this. I don't think that any other memory instructions require more than 8 bytes of alignment -- after all, alignment can only possibly matter if the size of an object to be loaded is at least as large as the alignment, or one could never load individual such objects. So the only instructions that could possibly care about 64-byte alignment must be AVX512 instructions. I don't know whether they just don't work on unaligned data, or simply incur a time penalty.

include/deal.II/base/aligned_vector.h
tamiko: I only meant for `Utilities::MPI::this_mpi_process(shmem_roots_communicator)` and `Utilities::MPI::n_mpi_processes(shmem_roots_communicator)`. The other calls would not need a local variable.

But actually, looking at it a second time: You use the boolean
```
Utilities::MPI::this_mpi_process(shmem_roots_communicator) == Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1
```
over and over again. What about introducing a boolean for that?
```
const bool is_last_shmem_process = (Utilities::MPI::this_mpi_process(shmem_roots_communicator) == Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1);

```
If find
```
if (last_shmem_process) {
  // ...
}
if (! last_shmem_process) {
  // ...
}
significantly easier to read.

Comment (tamiko) [reply include/deal.II/base/aligned_vector.h:1232]: (this is a minor nitpick obviously)

**Comment by bangerth** [comment]
@tamiko I did as you suggested!

File: include/deal.II/base/aligned_vector.h [hunk 13]
```diff
@@ -1036,6 +1124,418 @@ AlignedVector<T>::fill(const T &value)
...
+  // order of processes within the split communicators, and we will set it to
+  // zero for the root processes and one for all others -- which means that
+  // for all of these other processes, MPI can choose whatever order it
+  // wants because they have the same key (MPI then documents that these ties
+  // will be broken according to these processes' rank in the old group).
+  //
+  // At least that's the theory. In practice, the MPI implementation where
+  // this function was developed on does not seem to do that. But it
+  // is willing to put the root process onto the *last* rank of the
+  // communicator.
```
peterrum, line 1161: The function `MPI_Comm_split_type` seems to ignore for some reason the key. What one could do is to call `MPI_Comm_split_type` without any keys and resort the ranks - if needed - afterwards.

```diff
diff --git a/include/deal.II/base/aligned_vector.h b/include/deal.II/base/aligned_vector.h
index 9aab94e..b16be26 100644
--- a/include/deal.II/base/aligned_vector.h
+++ b/include/deal.II/base/aligned_vector.h
@@ -1161,26 +1161,32 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   // communicator.
   MPI_Comm shmem_group_communicator;
   {
+    MPI_Comm shmem_group_communicator_temp;
+    int      ierr =
+      MPI_Comm_split_type(communicator,
+                          MPI_COMM_TYPE_SHARED,
+                          Utilities::MPI::this_mpi_process(communicator),
+                          MPI_INFO_NULL,
+                          &shmem_group_communicator_temp);
+
+    AssertThrowMPI(ierr);
     const int key =
       (Utilities::MPI::this_mpi_process(communicator) == root_process ?
-         Utilities::MPI::n_mpi_processes(communicator) :
-         0);
-    const int ierr = MPI_Comm_split_type(communicator,
-                                         MPI_COMM_TYPE_SHARED,
-                                         key,
-                                         MPI_INFO_NULL,
-                                         &shmem_group_communicator);
+         0 :
+         Utilities::MPI::this_mpi_process(communicator) + 1);
+    ierr = MPI_Comm_split(shmem_group_communicator_temp,
+                          1,
+                          key,
+                          &shmem_group_communicator);
     AssertThrowMPI(ierr);
 
     // Verify the explanation from above
     if (Utilities::MPI::this_mpi_process(communicator) == root_process)
-      Assert(Utilities::MPI::this_mpi_process(shmem_group_communicator) ==
-               Utilities::MPI::n_mpi_processes(shmem_group_communicator) - 1,
+      Assert(Utilities::MPI::this_mpi_process(shmem_group_communicator) == 0,
              ExcInternalError());
   }
   const bool is_shmem_root =
-    Utilities::MPI::this_mpi_process(shmem_group_communicator) ==
-    Utilities::MPI::n_mpi_processes(shmem_group_communicator) - 1;
+    Utilities::MPI::this_mpi_process(shmem_group_communicator) == 0;
 
   // **** Step 2 ****
   // We then have to send the state of the current object from the
@@ -1216,8 +1222,8 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   {
     const int key =
       (Utilities::MPI::this_mpi_process(communicator) == root_process ?
-         Utilities::MPI::n_mpi_processes(communicator) :
-         0);
+         0 :
+         Utilities::MPI::this_mpi_process(communicator) + 1);
 
     const int ierr = MPI_Comm_split(communicator,
                                     /*color=*/
@@ -1228,13 +1234,11 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
 
     // Again verify the explanation from above
     if (Utilities::MPI::this_mpi_process(communicator) == root_process)
-      Assert(Utilities::MPI::this_mpi_process(shmem_roots_communicator) ==
-               Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1,
+      Assert(Utilities::MPI::this_mpi_process(shmem_roots_communicator) == 0,
              ExcInternalError());
   }
-  const unsigned int shmem_roots_root_rank =
-    Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1;
-  const bool is_shmem_roots_root =
+  const unsigned int shmem_roots_root_rank = 0;
+  const bool         is_shmem_roots_root =
     (Utilities::MPI::this_mpi_process(shmem_roots_communicator) ==
      shmem_roots_root_rank);
 
@@ -1260,12 +1264,11 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
             resize(new_size);
 
           // Then directly copy from the root process into these buffers
-          int ierr = MPI_Bcast(
-            elements.get(),
-            sizeof(T) * new_size,
-            MPI_CHAR,
-            Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1,
-            shmem_roots_communicator);
+          int ierr = MPI_Bcast(elements.get(),
+                               sizeof(T) * new_size,
+                               MPI_CHAR,
+                               0,
+                               shmem_roots_communicator);
           AssertThrowMPI(ierr);
         }
       else
@@ -1279,8 +1282,7 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
           // pointer, which would trigger the deleter which would lead to a
           // deadlock. So we just send the result of the broadcast() call to
           // nirvana on the root process and keep our current state.
-          if (Utilities::MPI::this_mpi_process(shmem_roots_communicator) ==
-              Utilities::MPI::n_mpi_processes(shmem_roots_communicator) - 1)
+          if (Utilities::MPI::this_mpi_process(shmem_roots_communicator) == 0)
             Utilities::MPI::broadcast(shmem_roots_communicator,
                                       *this,
                                       shmem_roots_root_rank);
@@ -1340,11 +1342,11 @@ AlignedVector<T>::replicate_across_communicator(const MPI_Comm &   communicator,
   // able to learn about the array size that way.
   MPI_Win        shmem_window;
   void *         base_ptr;
-  const MPI_Aint align_by   = 64;
-  const MPI_Aint alloc_size = Utilities::MPI::broadcast(
-    shmem_group_communicator,
-    (size() * sizeof(T) + align_by - 1),
-    Utilities::MPI::n_mpi_processes(shmem_group_communicator) - 1);
+  const MPI_Aint align_by = 64;
+  const MPI_Aint alloc_size =
+    Utilities::MPI::broadcast(shmem_group_communicator,
+                              (size() * sizeof(T) + align_by - 1),
+                              0);
 
   {
     const int disp_unit = (is_shmem_root ? size() : 1);

````
This seems to work.

File: include/deal.II/base/aligned_vector.h [hunk 14]
```diff
@@ -1036,6 +1124,418 @@ AlignedVector<T>::fill(const T &value)
...
+  // the shmem group participate in the generation of the shmem memory window,
+  // only the shmem root actually allocates any memory -- the rest just
+  // allocate zero bytes of their own. We allocate space for exactly
+  // size() elements (computed on the shmem_root that already has the data)
+  // and add however many bytes are necessary so that we know that we can align
+  // things to 64-byte boundaries. The worst case happens if the memory system
+  // gives us a pointer to an address one byte past a desired alignment
+  // boundary, and in that case aligning the memory will require us to waste the
+  // first (align_by-1) bytes. So we have to ask for
+  //   size() * sizeof(T) + align_by - 1
```
peterrum, line 1318:
```suggestion
  //   `size() * sizeof(T) + (align_by - 1)`
```

include/deal.II/base/aligned_vector.h
```diff
@@ -1036,6 +1124,418 @@ AlignedVector<T>::fill(const T &value)
...
+  // documentation, and that we do not actually need. So we "abuse" it a bit: On
+  // the shmem root, we put the array size into it. Later on, the remaining
+  // processes can query the shmem root's value of 'disp_unit', and so will be
+  // able to learn about the array size that way.
+  MPI_Win        shmem_window;
+  void *         base_ptr;
+  const MPI_Aint align_by   = 64;
+  const MPI_Aint alloc_size = Utilities::MPI::broadcast(
+    shmem_group_communicator,
+    (size() * sizeof(T) + align_by - 1),
```
peterrum, line 1346:
```suggestion
    (size() * sizeof(T) + (align_by - 1)),
```

include/deal.II/base/aligned_vector.h
```diff
@@ -259,6 +260,93 @@ class AlignedVector
...
+   *   for these operations. If you do not do so, you will likely get
+   *   a deadlock that may be difficult to debug. By extension, this rule of
+   *   only collectively resizing extends to this function itself: You can
+   *   not call it twice in a row because that implies that first all but the
+   *   `root_process` throw away their data, which is not a collective
+   *   operation. Generally, these restrictions on what can and can not be
+   *   done hint at the correctness of the comments above: You should treat
+   *   an AlignedVector on which the current function has been called as
+   *   `const`, on which no further operations can be performed until
+   *   the destructor is called.
```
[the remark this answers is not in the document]
bangerth: Maybe nothing. It is true that we could go into much more detail as for one *could* in principle do after calling `replicate_across_communicator()`, but my goal with this description is to not do that: I don't want to support any operations after the call, and got bitten several times in attempts at implementing any other scheme that would provide semantics to what these functions beyond declaring that one should consider the object as `const`. I also see no reason to support anything other: if you want another object, just create a new one.

include/deal.II/base/aligned_vector.h
```diff
@@ -259,6 +260,93 @@ class AlignedVector
...
+   */
+  void
```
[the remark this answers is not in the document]
bangerth: Well, in principle they are all allowed to modify the data, and there is a test that shows this.

Do you have a use case where it would be useful? As I mentioned elsewhere, I would very much like to describe all of this as turning the object into a read-only one.

include/deal.II/base/aligned_vector.h
bangerth: Right, but does this actually improve things? I would (i) still have to explain in the comments why I'm not doing the renumbering in `MPI_Comm_split_type` lest someone come along and make that change, (ii) we're adding another collective communication step.

I don't disagree that one *could* do it, I just don't quite see *why*. But I'm also not opposed if you're adamant about it.

include/deal.II/base/aligned_vector.h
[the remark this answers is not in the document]
peterrum: If this is the way users should use this class maybe we should disable the mentioned functions (except of `clear()`) by adding some asserts that are triggered only if `replicate_across_communicator()` was called in debug mode.

include/deal.II/base/aligned_vector.h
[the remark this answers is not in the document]
peterrum: > Well, in principle they are all allowed to modify the data, and there is a test that shows this.

I know that all can modify the data but then you would have the problem with race conditions (and the easiest way would be to let only the root process modify the data).  

> Do you have a use case where it would be useful?

`AlignedVector` might be used/misused for exchanging simply data. But I guess to make it really powerful, I guess the users might want the flexibility to provide their own subcommunicators. But I guess this could be done later on in a follow-up PR.

include/deal.II/base/aligned_vector.h
peterrum: Not making rank 0 to the root look a bit odd. That the `keys` do not work properly is a bit worrying. I am fine to leave it as it is but have you tried to run the tests also on multiple compute nodes? Is the behavior the same there? The tests in the test suite can only test the case that all processes are on the same shmem domain. This will probably not change as long as not a user-provided communicator can passed the function.

include/deal.II/base/aligned_vector.h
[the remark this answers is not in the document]
bangerth: I thought about that as well, but it would require storing information somewhere in a flag -- and making the object larger by another 4 bytes. That was already the most contentious point :-(

For completeness, I had also considered storing this information elsewhere. For example, we have the two pointers that point at the end of the *used* area and the end of the *allocated* area. I had thought about setting the latter to `nullptr` to indicate that we have a read-only object. That didn't seem like a particularly nice way to do that, but has the advantage that we don't waste memory. If that would be acceptable, then I'd be happy to do this.

bangerth:
On 4/26/21 1:11 PM, Peter Munch wrote:
> Not making rank 0 to the root look a bit odd. That the |keys| do not work 
> properly is a bit worrying. I am fine to leave it as it is but have you
> tried to run the tests also on multiple compute nodes? Is the behavior the
> same there? The tests in the test suite can only test the case that all
> processes are on the same shmem domain. This will probably not change as
> long as not a user-provided communicator can passed the function.

As mentioned above, I don't actually have access to a multinode system :-( I 
would appreciate if someone could test this for me.

It is a fair concern that what happens to work for me here might not actually 
work in general. Maybe I should, after all, use your patch.

tjhei:
I will test this but this might take me a week or two. I also saw that intel MPI does the alignment correctly. Need to test this as well.

bangerth:
@tjhei Thanks for offering to test!

@marcfehling Do you happen to have access to a cluster at the moment? Would you mind trying this out?

tamiko:
@bangerth Happy to run this on our small cluster. What do I have to do? Just run all new tests?

tamiko:
@bangerth I do encounter the following for `base/aligned_vector_replicate_02.release`

`# srun -N 1 -n 3 aligned_vector_replicate_02.release` (one node, three ranks) I get the expected output:
```
JobId a4
DEAL:0::On process 0: 
DEAL:0::42 43 2 44 4 0 6 0 0 0 
JobId a4
DEAL:1::On process 1: 
DEAL:1::42 43 2 44 4 0 6 0 0 0 

JobId a4
DEAL:2::On process 2: 
DEAL:2::42 43 2 44 4 0 6 0 0 0
```

`# srun -N 2 -n 3 aligned_vector_replicate_02.release` (two nodes, three ranks):
```
JobId a4
DEAL:0::On process 0: 
DEAL:0::42 43 2 0 4 0 6 0 0 0 
JobId a4
DEAL:1::On process 1: 
DEAL:1::42 43 2 0 4 0 6 0 0 0 

JobId a5
DEAL:2::On process 2: 
DEAL:2::0 0 2 44 4 0 6 0 0 0 
```

`srun -N 3 -n 3 aligned_vector_replicate_02.release` (three nodes, three ranks):
```
JobId a4
DEAL:0::On process 0: 
DEAL:0::42 0 2 0 4 0 6 0 0 0 
JobId a5
DEAL:1::On process 1: 
DEAL:1::0 43 2 0 4 0 6 0 0 0 

JobId a6
DEAL:2::On process 2: 
DEAL:2::0 0 2 44 4 0 6 0 0 0
```

tests/base/aligned_vector_replicate_02.cc
```diff
@@ -0,0 +1,84 @@
...
+  else if (Utilities::MPI::this_mpi_process(communicator) == 2)
+    avec[3] = 44;
+
+  // Final step, let every process output what it now has. Every
+  // process should be able to see the changed element, even though
+  // only one changed it.
+  deallog << "On process " << Utilities::MPI::this_mpi_process(communicator)
+          << ": " << std::endl;
+  for (const auto i : avec)
+    deallog << i << ' ';
```
peterrum, line 71: This does not look right. One needs to introduce here a barrier else we might get race conditions.

peterrum:
@tamiko To me the output looks alright. That the test "fails" for multiple nodes is expected since processes can only see changes in the shmem domain.

peterrum:
I have created the following mini test:
```cpp
#include <deal.II/base/aligned_vector.h>
#include <deal.II/base/mpi.h>

using namespace dealii;

void
test(const unsigned int root)
{
  if (Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == root)
    std::cout << root << " of "
              << Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD) << ": ";

  AlignedVector<int> avec(1000);
  if (Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == root)
    {
      unsigned int counter;
      for (auto &i : avec)
        i = counter++;
    }

  avec.replicate_across_communicator(MPI_COMM_WORLD, root);

  {
    unsigned int counter = 0;

    for (auto &i : avec)
      AssertDimension(i, counter++);
  }

  MPI_Barrier(MPI_COMM_WORLD);

  if (Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == root)
    std::cout << "OK!" << std::endl;
}



int
main(int argc, char **argv)
{

  for (unsigned int i = 0; i < Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
       ++i)
    test(i);
}
```
I wanted to run it on multiple nodes. But I have issues with a single node when I use two processes and root is rank 0. Maybe I am doing something wrong. Does anyone have similar issues?

tamiko:
@peterrum I figured that the test will give different results for different node configurations. :smile_cat:

tamiko (approved):
I propose we merge quickly now so that we have enough time to run this through some "production" stages including the full testsuite and some hand-picked testing on small clusters.

peterrum (changes requested):
Before merging and creating follow-up PRs, I would like to discuss why the test https://github.com/dealii/dealii/pull/12022#issuecomment-827390738 is failing. Maybe, I am using the new function in the wrong way or there is indeed an issue with the keys.

tests/base/aligned_vector_replicate_02.cc
bangerth: Yes, you're right. I added the barrier!

bangerth:
OK, the news:
* The OpenMPI bug has been fixed with a one-character patch.
* I ended up using @peterrum 's approach using `MPI_Comm_split()` after all. No solution is particularly elegant when working around the OpenMPI bug...
* I put the little testcase @peterrum created into the test suite as the _03 test. 

@peterrum -- I don't know why it failed for you, but the testcase has two little bugs: the first time you create `counter`, you leave it uninitialized. Basically the same mistake as we had on the mailing list a couple of days ago :-) Second, in the place where you use `counter` the second time, you have
```
AssertDimension(i, counter++);
```
which is only going to do what you wanted in debug mode.

@tamiko -- thanks very much for testing! If you wanted to give it another try, run the _03 test on more ranks than you have processes on one machine.

peterrum (approved):
> I don't know why it failed for you,

It failed inside of `AlignedVector::replicate_across_communicator()` at the asserts. I guess I would have noticed the unitialized variable once I would have reached my asserts ;)

peterrum:
I am merging. We can address the const-mode in follow-up PRs.

bangerth:
On 4/27/21 11:52 PM, Peter Munch wrote:
> It failed inside of |AlignedVector::replicate_across_communicator()| at the 
> asserts. I guess I would have noticed the unitialized variable once I would 
> have reached my asserts ;)

But is it fixed now?

peterrum:
Yes it is :+1: Forgot to mentioned it: 7 nodes with each 4 processes.

bangerth:
On 4/28/21 9:08 AM, Peter Munch wrote:
> Forgot to mentioned it: 7 nodes with each 40 processes.

Excellent, thank you for testing!

PR #12024, opened 2021-04-08
Add class FEImmersedSurfaceValues

simonsticko:
That has the same shape function values/derivatives as FEValues,
but has JxW-values corresponding to integration over a surface
immersed in the cell. This is achieved by passing an
ImmersedSurfaceQuadrature instead of a Quadrature to the constructor.

Add a virtual function fill_fe_immersed_surface_values on Mapping to
be able to compute the correct JxW-values and surface normals. Make it
throw ExcNotImplemented by default, but override it with an
implementation on MappingCartesian.

### DIFF
--- a/include/deal.II/non_matching/fe_immersed_values.h
+++ b/include/deal.II/non_matching/fe_immersed_values.h
@@ -1,0 +1,153 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2021 - 2021 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+
+#ifndef dealii_non_matching_fe_immersed_values_h
+#define dealii_non_matching_fe_immersed_values_h
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/dofs/dof_accessor.h>
+
+#include <deal.II/fe/fe.h>
+#include <deal.II/fe/fe_update_flags.h>
+#include <deal.II/fe/fe_values.h>
+#include <deal.II/fe/mapping.h>
+
+#include <deal.II/grid/tria_iterator.h>
+
+#include <deal.II/non_matching/immersed_surface_quadrature.h>
+
+DEAL_II_NAMESPACE_OPEN
+namespace NonMatching
+{
+  /**
+   * Finite element evaluated in the quadrature points of an
+   * ImmersedSurfaceQuadrature of a cell.
+   *
+   * The shape functions values and their derivatives are the same as for an
+   * FEValues-object, but the JxW-values are computed with the transformation
+   * described in the documentation of ImmersedSurfaceQuadrature. Further, the
+   * normal_vector-function returns the normal to the immersed surface.
+   *
+   * The reinit-function of this class exist mostly to be consistent with the
+   * other FEValues-like classes. The immersed quadrature rule will typically
+   * vary between each cell of the triangulation. Thus, an
+   * FEImmersedSurfaceValues object can, typically, not be reused for different
+   * cells.
+   *
+   * See also documentation in FEValuesBase.
+   *
+   * @ingroup feaccess
+   */
+  template <int dim>
+  class FEImmersedSurfaceValues : public FEValuesBase<dim, dim>
+  {
+  public:
+    /**
+     * Constructor. Gets cell-independent data from mapping and finite element
+     * objects, matching the quadrature rule and update flags.
+     *
+     * @note Currently this class is only implemented for MappingCartesian.
+     */
+    FEImmersedSurfaceValues(const Mapping<dim> &                  mapping,
+                            const FiniteElement<dim> &            element,
+                            const ImmersedSurfaceQuadrature<dim> &quadrature,
+                            const UpdateFlags                     update_flags);
+
+    /**
+     * Reinitialize quantities (normals, JxW-values, etc) for the given cell of
+     * type "iterator into a Triangulation object".
+     */
+    void
+    reinit(const typename Triangulation<dim>::cell_iterator &cell);
+
+    /**
+     * Reinitialize quantitites (shape function values, gradients, etc) for the
+     * given cell of type "iterator into a DoFHandler object", and the finite
+     * element associated with this object.
+     */
+    template <bool level_dof_access>
+    void
+    reinit(
+      const TriaIterator<DoFCellAccessor<dim, dim, level_dof_access>> &cell);
+
+    /**
+     * Returns the surface gradient of the shape function with index
+     * @p function_no at the quadrature point with index @p quadrature_point.
+     *
+     * The surface gradient is defined as the projection of the gradient to the
+     * tangent plane of the surface:
+     * $ \nabla u - (n \cdot \nabla u) n $,
+     * where $n$ is the unit normal to the surface.
+     *
+     * @dealiiRequiresUpdateFlags{update_gradients | update_normal_vectors}
+     */
+    Tensor<1, dim>
+    shape_surface_grad(const unsigned int function_no,
+                       const unsigned int quadrature_point) const;
+
+    /**
+     * Return one vector component of the surface gradient of the shape function
+     * at a quadrature point. See the definition of the surface gradient in the
+     * shape_surface_grad function.
+     *
+     * @p function_no Index of the shape function to be evaluated.
+     *
+     * @p point_no Index of the quadrature point at which the function is to be
+     * evaluated.
+     *
+     * @p component Vector component to be evaluated.
+     *
+     * @dealiiRequiresUpdateFlags{update_gradients | update_normal_vectors}
+     */
+    Tensor<1, dim>
+    shape_surface_grad_component(const unsigned int function_no,
+                                 const unsigned int quadrature_point,
+                                 const unsigned int component) const;
+
+    /**
+     * Return a reference to the copy of the quadrature formula stored by this
+     * object.
+     */
+    const NonMatching::ImmersedSurfaceQuadrature<dim> &
+    get_quadrature() const;
+
+  protected:
+    /**
+     * Do work common to the constructors.
+     */
+    void
+    initialize(const UpdateFlags update_flags);
+
+    /**
+     * The reinit() functions do only that part of the work that requires
+     * knowledge of the type of iterator. After setting present_cell(), they
+     * pass on to this function, which does the real work, and which is
+     * independent of the actual type of the cell iterator.
+     */
+    void
+    do_reinit();
+
+    /**
+     * Copy of the quadrature formula that was passed to the constructor.
+     */
+    const ImmersedSurfaceQuadrature<dim> quadrature;
+  };
+
+} // namespace NonMatching
+DEAL_II_NAMESPACE_CLOSE
+
+#endif /* dealii_non_matching_fe_immersed_values_h */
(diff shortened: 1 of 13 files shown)


simonsticko:
> One concern I have is that we have added something to `Mapping` but only implemented it for `MappingCartesian`. Is there a path forward for doing this for all the other mapping classes?

Yes, the only thing that differs is that the `JxW`-values and the normals transform as written down here:

https://www.dealii.org/current/doxygen/deal.II/classNonMatching_1_1ImmersedSurfaceQuadrature.html

which is well-defined regardless of mapping. In practice, one would almost always use a Cartesian mesh, but If you want to, I can implement the function that was added on `Mapping` on `MappingQGeneric` too. Maybe that makes sense since you get `MappingQGeneric` by default for the other `FE*Values`-classes.

source/fe/mapping_cartesian.cc
```diff
@@ -601,6 +601,90 @@ MappingCartesian<dim, spacedim>::fill_fe_subface_values(
...
+  dealii::internal::FEValuesImplementation::MappingRelatedData<dim, spacedim>
+    &output_data) const
+{
+  Assert(dim == spacedim, ExcNotImplemented());
+
+  // Convert data object to internal data for this class. Fails with an
+  // exception if that is not possible.
+  Assert(dynamic_cast<const InternalData *>(&internal_data) != nullptr,
+         ExcInternalError());
+  const InternalData &data = static_cast<const InternalData &>(internal_data);
```
peterrum, line 619:
```suggestion
  const auto data = dynamic_cast<const InternalData *>(&internal_data);
  Assert(data != nullptr, ExcInternalError());
```

source/fe/mapping_cartesian.cc
```diff
@@ -601,6 +601,90 @@ MappingCartesian<dim, spacedim>::fill_fe_subface_values(
...
+          output_data.jacobians[i][d][d] = data.cell_extents[d];
+      }
+
+  if (data.update_each & update_inverse_jacobians)
+    for (unsigned int i = 0; i < output_data.inverse_jacobians.size(); ++i)
+      {
+        output_data.inverse_jacobians[i] = DerivativeForm<1, dim, spacedim>();
+        for (unsigned int d = 0; d < dim; ++d)
+          output_data.inverse_jacobians[i][d][d] = 1. / data.cell_extents[d];
+      }
```
peterrum, line 650: These look like in the case of `fill_fe_face_values` and the two below are the new ones? Am I right?

peterrum (commented):
Overall, it looks to me good. But, I must admit I am not an expert in these classes.

source/fe/mapping_cartesian.cc
kronbichler: The original code is the same as in all other cases, and its goal is to avoid the (expensive) dynamic cast with its lookup in release mode. This might seem nit-picking but benchmarking things around 12 years ago I realized it actually pays off. :smile:

include/deal.II/fe/fe_values.h
```diff
@@ -4588,6 +4590,110 @@ class FESubfaceValues : public FEFaceValuesBase<dim, spacedim>
...
+ * other FEValues-like classes. The immersed quadrature rule will typically vary
+ * between each cell of the triangulation. Thus, an FEImmersedSurfaceValues
+ * object can, typically, not be reused for different cells.
+ *
+ * See also documentation in FEValuesBase.
+ *
+ * @ingroup feaccess
+ */
+template <int dim>
+class FEImmersedSurfaceValues : public FEValuesBase<dim, dim>
```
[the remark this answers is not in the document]
kronbichler: Yes, I think it would make sense to keep things collected in that namespace for compactness. `fe_values.h` is a very large file already, so it just helps with the additional implementation. Since we do not put restrictions on the include files between different subfolders in deal.II, it is also no problem to load the additional header in the mapping classes and set the appropriate forward declaration for the `friends` specification.

source/fe/mapping_cartesian.cc
peterrum: I mean why don't we create a deal.II-macro for this purpose. Dynamic casting is quite common in the library!

kronbichler:
> In practice, one would almost always use a Cartesian mesh, but If you want to, I can implement the function that was added on Mapping on `MappingQGeneric` too. Maybe that makes sense since you get `MappingQGeneric` by default for the other FE*Values-classes.

I would really like to have this for `MappingQ` as well and possible also for `MappingFE` which is used by the simplices. The reason for `MappingQ` is that I imagine to compute, at some point, some overlapping meshes that are not Cartesian. Not sure if all our infrastructure supports it yet, but I think having a path forward is great.

simonsticko:
I think I've addressed everything here. In particular:
- Moved this class to namespace `NonMatching`
- Implemented `fill_fe_immersed_surface_values` on `MappingQ`, so one can use an cut-method even if the background mesh isn't Cartesian.

source/non_matching/fe_immersed_values.cc
```diff
@@ -0,0 +1,184 @@
...
+
+
+  template <int dim>
+  inline void
+  FEImmersedSurfaceValues<dim>::initialize(const UpdateFlags update_flags)
+  {
+    UpdateFlags flags = this->compute_update_flags(update_flags);
+
+    if (flags & (update_JxW_values | update_normal_vectors))
+      flags |= update_covariant_transformation;
```
simonsticko, line 140: Wasn't really sure if this was the right place to set this flag.

include/deal.II/non_matching/fe_immersed_values.h
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Reinitialize the gradients, Jacobi determinants, etc for the incoming @p cell.
+     */
+    template <bool level_dof_access>
+    void
+    reinit(
+      const TriaIterator<DoFCellAccessor<dim, dim, level_dof_access>> &cell);
+
+
```
kronbichler, line 78:
```suggestion
```

source/non_matching/fe_immersed_values.cc
```diff
@@ -0,0 +1,184 @@
...
+    const TriaIterator<DoFCellAccessor<dim, dim, level_dof_access>> &cell)
+  {
+    // Assert that the finite elements passed to the constructor and used by the
+    // DoFHandler used by this cell, are the same
+    Assert(static_cast<const FiniteElementData<dim> &>(*this->fe) ==
+             static_cast<const FiniteElementData<dim> &>(cell->get_fe()),
+           (typename FEValuesBase<dim>::ExcFEDontMatch()));
+
+    this->maybe_invalidate_previous_present_cell(cell);
+    this->present_cell = {cell};
```
kronbichler, line 55: Does this call via the constructor https://github.com/dealii/dealii/blob/2890a38bc0909248658137c26984c6d8e9f276f8/include/deal.II/fe/fe_values.h#L3833-L3835 work for all compilers? If I remember correctly we've had trouble with some code like this for some of the compilers (but I believe the CI should reveal it if problematic).

include/deal.II/non_matching/fe_immersed_values.h
```diff
@@ -0,0 +1,145 @@
...
+                            const UpdateFlags                     update_flags);
+
+    /**
+     * Reinitialize the gradients, Jacobi determinants, etc for the incoming @p cell.
+     */
+    template <bool level_dof_access>
+    void
+    reinit(
+      const TriaIterator<DoFCellAccessor<dim, dim, level_dof_access>> &cell);
+
```
kronbichler, line 77: Do you anticipate that we never need an initialization with `typename Triangulation<dim, spacedim>::cell_iterator &cell`, as we provide in the `FEValues` class besides the `TriaIterator<DoFCellAccessor>`?

kronbichler (approved):
This looks great to me, thank you for your endurance. I have a few small comments and questions, nothing serious any more.

source/non_matching/fe_immersed_values.cc
simonsticko: I don't know. This is the same way `present_cell` is reset in the other `FEValues` class, but this was recently rewritten. What does @peterrum think?

peterrum: Yes. This is a plain copy from the the other `FEValues` classes. I think we use `={}` at other places in the library. I think there are even compilers that complain if you are explicitly write the type twice...

kronbichler: Well, since we've had a round of testers on it since it was merged, I think we are good. I just wanted to check because I've been bitten by these before.

peterrum:
@simonsticko Could you rebase this branch. I am wondering why the undefined reference is still popping up!

peterrum:
> I am wondering why the undefined reference is still popping up!

Looks good now :)

include/deal.II/non_matching/fe_immersed_values.h
simonsticko: I've added another `reinit` function and modified the test slightly to check that it works.

source/non_matching/fe_immersed_values.cc
```diff
@@ -0,0 +1,214 @@
...
+    this->fe_data = std::move(fe_get_data.return_value());
+    if (flags & update_mapping)
+      this->mapping_data = std::move(mapping_get_data.return_value());
+    else
+      this->mapping_data =
+        std::make_unique<typename Mapping<dim>::InternalDataBase>();
+  }
+
+
+#include "fe_immersed_values.inst"
```
peterrum, line 211: The error looks odd. Maybe you could disable this include if doxygen is built?

simonsticko: Yeah, I didn't understand if this was a doxygen bug or something else. I've added your suggestion.

peterrum: Excellent, it worked! I'll have another look at the PR later today!

source/non_matching/fe_immersed_values.cc
```diff
@@ -0,0 +1,216 @@
...
+  {
+    // Check that mapping and reference cell type are compatible:
+    Assert(this->get_mapping().is_compatible_with(cell->reference_cell()),
+           ExcMessage(
+             "You are trying to call FEValues::reinit() with a cell of type " +
+             cell->reference_cell().to_string() +
+             " with a Mapping that is not compatible with it."));
+
+    // No FE in this cell, so no assertion necessary here.
+    this->maybe_invalidate_previous_present_cell(cell);
```
peterrum, line 55: I am wondering if we should be explicit here and prevent users to reuse this class:
```suggestion
    Assert(this->present_cell==false, ExcMessage("This class should not be reused."));
```

source/non_matching/fe_immersed_values.cc
peterrum: Where is `update_covariant_transformation` needed?

peterrum (approved):
I left two minor comments. I am merging; we can make adjustments in follow-up PRs.

peterrum:
> Implemented fill_fe_immersed_surface_values on MappingQ, so one can use an cut-method even if the background mesh isn't Cartesian.

I guess not the quadrature-point generator needs to be extended as well ;)

source/non_matching/fe_immersed_values.cc
simonsticko: Yes. It's probably a mistake if you call `reinit` twice.

source/non_matching/fe_immersed_values.cc
simonsticko: When computing normals and the `JxW`-values in  `MappingQ::fill_fe_immersed_surface_values`. Line 1363 in `mapping_q.cc`

simonsticko:
> I guess not the quadrature-point generator needs to be extended as well ;)

Not if we use a discrete level set function (`DoFHandler` and a `Vector`), because the new class in #12756 transforms such a level set function to reference space in the background. So we will generate the quadrature rules in reference space, then just use these in `FEValues` or `FEImmersedSurfaceValues`.

source/non_matching/fe_immersed_values.cc
kronbichler: Indeed. The covariant/contravariant transformations are the most fundamental flags to compute the Jacobians of the transformation between reference and real cell, and as such the necessary step before one can fill the JxW or normal vectors, which depend on the Jacobian.

PR #12027, opened 2021-04-08
Improve error messages during configuration.

bangerth:
We print an error message in our wrappers to the `FIND_` macros that shows a string representation of arguments to the macro. But cmake has a habit of clobbering these arguments and overwriting with values that are not useful to output. We should output the original set of arguments, so let's save them.

Example
```
-- SUNDIALS_CONFIG_H not found! Call:
--     FIND_FILE(SUNDIALS_CONFIG_H NAMES sundials_config.h HINTS SUNDIALS_INCLUDE_DIR-NOTFOUND/sundials)
```
Note the appearance of the `SUNDIALS_INCLUDE_DIR-NOTFOUND` text.

/rebuild

### DIFF
--- a/cmake/macros/macro_deal_ii_find_path.cmake
+++ b/cmake/macros/macro_deal_ii_find_path.cmake
@@ -18,11 +18,14 @@
 #
 
 MACRO(DEAL_II_FIND_PATH _path_name)
+  # Save a string representation of the arguments before cmake's
+  # FIND_PATH gets its hands on it.
+  TO_STRING(_str ${ARGN})
+
   FIND_PATH(${_path_name} ${ARGN})
 
   IF(${_path_name} MATCHES "-NOTFOUND")
-    MESSAGE(STATUS "${_path_name} not found! Call:")
-    TO_STRING(_str ${ARGN})
+    MESSAGE(STATUS "${_path_name} not found! The call was:")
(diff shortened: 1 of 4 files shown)


cmake/macros/macro_deal_ii_find_path.cmake
```diff
@@ -18,11 +18,14 @@
 #
 
 MACRO(DEAL_II_FIND_PATH _path_name)
+  # Save a string representation of the arguments before cmake's
+  # FIND_FILE gets its hands on it.
```
drwells, line 22:
```suggestion
  # FIND_PATH gets its hands on it.
```

cmake/macros/macro_deal_ii_find_program.cmake
```diff
@@ -18,11 +18,14 @@
 #
 
 MACRO(DEAL_II_FIND_PROGRAM _file_name)
+  # Save a string representation of the arguments before cmake's
+  # FIND_FILE gets its hands on it.
```
drwells, line 22:
```suggestion
  # FIND_PROGRAM gets its hands on it.
```

drwells (changes requested):
I checked that this fixes the output issue - nice work. We need to update two comments, though.

bangerth:
So done, thanks for the review @drwells !

masterleinad (approved):

drwells (approved):

PR #12028, opened 2021-04-08
Make cmake messages easier to read.

bangerth:
/rebuild


cmake/configure/configure_2_trilinos.cmake
```diff
@@ -38,17 +38,17 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
     # Check whether all required modules of trilinos are installed:
     #
     MESSAGE(STATUS
-      "Check whether the found trilinos package contains all required modules:"
+      "Checking whether the found trilinos package contains all required modules:"
```
bangerth, line 41: This isn't supposed to be a command ("check!") but a status update about what cmake is currently doing.

cmake/configure/configure_2_trilinos.cmake
```diff
@@ -38,17 +38,17 @@ MACRO(FEATURE_TRILINOS_FIND_EXTERNAL var)
...
       )
 
     FOREACH(_module
         Amesos Epetra Ifpack AztecOO Teuchos ML
       )
       ITEM_MATCHES(_module_found ${_module} ${Trilinos_PACKAGE_LIST})
       IF(_module_found)
-        MESSAGE(STATUS "Found ${_module}")
+        MESSAGE(STATUS "  Found ${_module}")
```
bangerth, line 49: I'm indenting a long list here relative to the text that describes what the list is about.

Rombur (approved):

Issue #5027, opened 2017-09-05
More SUNDIALS interfaces.


luca-heltai:
Sundials provides several interfaces that should be included in deal.II. 

In particular we should provide interfaces to

- [ ] CVODE
- [ ] CVODEs
- [x] ARKODE
- [x] IDA
- [ ] IDAs
- [x] KINSOL

In https://github.com/mathlab/deal2lkit we cover `IDA`, `ARKODE`, and `KINSOL`, but the general idea is simple enough to extend to all other solvers as well.

luca-heltai:
#4852 Implements IDA

bangerth:
#5044 adds ARKODE

juan42:
Since we had a really pleasant experience with the the library and existing IDA interface and our final goal is time dependent adjoint sensitivities, i'd like to implement the IDAs interface.

Before i really start though, i'd like to know if there is already work going on.
My initial steps would be

1. Mostly copy/paste work from existing IDA wrapper
2. Test IDAS for an IVP (which is exactly what IDA is doing right now)
3. Test IDAS for FSA (forward sensitivity analysis)
4. Test IDAS for ASA (adjoint sensitivity analysis)

I'd appreciate any suggestions on how to tackle this. I'd also like to know if there is an opinion on having IDA and IDAs coexist.

drwells:
We've made a lot of independent progress on SUNDIALS recently that is tracked elsewhere.

PR #12030, opened 2021-04-08
Support different KINSOL versions.

bangerth:
In #11908, I made the assumption that the current KINSOL interface was introduced in SUNDIALS 5.0, but did not check. I now downloaded a whole bunch of SUNDIALS versions and checked that things compile. It's not so easy. The old interface worked until 4.0, they made changes in 4.1, and then again switched to a new interface for 5.0 and later. In practice, that meant that deal.II didn't compile when using SUNDIALS 4.1.

I made the minimal change set to make things compile for 4.0, 4.1, 5.0, and 5.7 with this patch. The interfaces even work correctly for 4.0, 5.0, 5.7 (and presumably everything between 5.1 ... 5.6 as well). I just put an assert in for the case of 4.1. One could in principle spend a bunch of time to get that working as well, but I saw relatively little point spending an hour or two supporting one very specific old version of SUNDIALS when they have already brought out 8 other releases since then that work for us.

Related to #5027.

/rebuild

### DIFF
--- a/source/sundials/kinsol.cc
+++ b/source/sundials/kinsol.cc
@@ -160,7 +160,7 @@ namespace SUNDIALS
 
 
 
-#  if DEAL_II_SUNDIALS_VERSION_LT(5, 0, 0)
+#  if DEAL_II_SUNDIALS_VERSION_LT(4, 1, 0)
     template <typename VectorType>
     int
     setup_jacobian_callback(KINMem kinsol_mem)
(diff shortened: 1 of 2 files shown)


source/sundials/kinsol.cc
```diff
@@ -431,13 +431,27 @@ namespace SUNDIALS
     if (solve_jacobian_system) // user assigned a function object to the solver
                                // slot
       {
-#  if DEAL_II_SUNDIALS_VERSION_LT(5, 0, 0)
+#  if DEAL_II_SUNDIALS_VERSION_LT(4, \
+                                  1, \
+                                  0) /* interface up to and including 4.0 */
```
masterleinad, line 436:
```suggestion
/* interface up to and including 4.0 */
#  if DEAL_II_SUNDIALS_VERSION_LT(4, 1, 0) 
```

source/sundials/kinsol.cc
```diff
@@ -431,13 +431,27 @@ namespace SUNDIALS
...
         auto KIN_mem        = static_cast<KINMem>(kinsol_mem);
         KIN_mem->kin_lsolve = solve_with_jacobian_callback<VectorType>;
         if (setup_jacobian) // user assigned a function object to the Jacobian
           // set-up slot
           KIN_mem->kin_lsetup = setup_jacobian_callback<VectorType>;
-#  else
+
+#  elif DEAL_II_SUNDIALS_VERSION_LT(5, \
+                                    0, \
+                                    0) /* interface up to and including 4.1 */
```
masterleinad, line 445:
```suggestion
/* interface up to and including 4.1 */
#  elif DEAL_II_SUNDIALS_VERSION_LT(5, 0, 0) 
```

masterleinad (approved):
Looks good to me but I would prefer to avoid the line breaks by shifting the comments.

bangerth:
@masterleinad I made the changes suggested. Thanks for the review!

PR #12031, opened 2021-04-08
Update the KINSOL callbacks to what KINSOL current provides.

bangerth:
Related to #5027, follow-up to #11908.

Starting with SUNDIALS 5.0, KINSOL passes the user a different set of inputs to the function that solves a linear system with the Jacobian matrix. Recall that KINSOL solves a problem of the form `F(U)=0` and so in the user function one can set up to solve a linear system with the Jacobian, i.e., `J_k x = b`, old KINSOL also gave us `U_k` and `F(U_k)`, along with whatever right hhand side `b` KINSOL wanted us to solve with. Presumably, they did that because that would allow computing the Jacobian matrix `J_k` right there and then. But that's already what the "setup" callback is there for.

So "new" KINSOL now only passes `b` and a `tolerance`. So I introduce a new callback that has exactly this interface, and deprecate the old one.

/rebuild


bangerth:
Ping?

If possible, I'd like to get this merged because then I can switch step-77 to use this more modern interface.

drwells (approved):

drwells:
This looks good and I appreciate the docs (I can understand this without understanding KINSOL). Do any of the tests actually execute this code?

bangerth:
@drwells Good point -- I meant to do that and forgot. I've cloned the three existing tests and switched them to the new interface.

drwells (approved):

PR #12032, opened 2021-04-09
Do not allocate memory for DoFs on artificial cells

peterrum:
This reduces the memory consumption of `DoFHandler` for a non-refined `p:d:T` from `#cells * #n_dofs_cells` to `#cells + #non_artificial_cells * #n_dofs_cells`. The `#cells` (pointers) is still not optimal; but is caused by the fact that all coarse cells are stored and the way cell information are accessed (cell level and index).


include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -811,9 +791,10 @@ namespace internal
       n_dof_indices(
         const dealii::DoFAccessor<structdim, dim, spacedim, level_dof_access>
           &                accessor,
-        const unsigned int fe_index_)
+        const unsigned int fe_index_,
+        const bool         level_dof_access_)
```
kronbichler, line 795: Do get it right that you change this to be able to use this new parameter independently of the template parameter of `DoFAccessor`? Can you maybe add a more descriptive name, say `count_level_dofs` or something like that? (And the same in the function using this below.)

kronbichler (approved):
I like this a lot, thanks for the work. 

The old approach to allocate the memory was just insane, but we were forced to in the original data structures. Thanks to the big updates that brought the `hp::DoFHandler` into `DoFHandler` (among other things), we have gained the flexibility to finally get rid of this nuisance.

include/deal.II/dofs/dof_accessor.templates.h
peterrum: Done :+1:

kronbichler (approved):

include/deal.II/dofs/dof_accessor.templates.h
```diff
@@ -811,9 +791,13 @@ namespace internal
       n_dof_indices(
         const dealii::DoFAccessor<structdim, dim, spacedim, level_dof_access>
           &                accessor,
-        const unsigned int fe_index_)
+        const unsigned int fe_index_,
+        const bool         count_level_dofs)
       {
-        if (level_dof_access)
+        // note: we cannot rely on the the template parameter level_dof_access
```
tjhei, line 797: Duplicated the

peterrum: Done!

Issue #12025, opened 2021-04-08
Fix MatrixFree::find_vector_in_mf for multiple DoFHandler/AffineConstrains


peterrum:
The following lines might return the wrong partitioner if `MatrixFree` has been set up with multiple `DoFHandler` and `AffineConstraint` object:
 
https://github.com/dealii/dealii/blob/20530a7ef6c7cf81ae5342e2a8cda7c98b90ed35/include/deal.II/matrix_free/matrix_free.h#L3284-L3291

with the consequence that different processes use different partitioners with different communication patterns, possibly leading to MPI errors of the type (see also https://github.com/MeltPoolDG/MeltPoolDG/pull/117#issuecomment-811762113):

```
[] *** An error occurred in MPI_Waitall
[] *** reported by process [2149777409,1]
[] *** on communicator MPI_COMM_WORLD
[] *** MPI_ERR_TRUNCATE: message truncated
[] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[] *** and potentially your MPI job)
```

By checking more internal fields in `Partitioner::is_compatible()` this issue could be solved (without the need to perform a global communication):
```diff
diff --git a/source/base/partitioner.cc b/source/base/partitioner.cc
index b6639aa..0e58858 100644
--- a/source/base/partitioner.cc
+++ b/source/base/partitioner.cc
@@ -487,8 +487,22 @@ namespace Utilities
         }
 #endif
       return (global_size == part.global_size &&
+              locally_owned_range_data == part.locally_owned_range_data &&
               local_range_data == part.local_range_data &&
-              ghost_indices_data == part.ghost_indices_data);
+              ghost_indices_data == part.ghost_indices_data &&
+              n_ghost_indices_data == part.n_ghost_indices_data &&
+              ghost_targets_data == part.ghost_targets_data &&
+              import_indices_data == part.import_indices_data &&
+              n_import_indices_data == part.n_import_indices_data &&
+              import_targets_data == part.import_targets_data &&
+              import_indices_chunks_by_rank_data ==
+                part.import_indices_chunks_by_rank_data &&
+              n_ghost_indices_in_larger_set ==
+                part.n_ghost_indices_in_larger_set &&
+              ghost_indices_subset_chunks_by_rank_data ==
+                part.ghost_indices_subset_chunks_by_rank_data &&
+              ghost_indices_subset_data == part.ghost_indices_subset_data);
     }
```
[...]

PR #12033, opened 2021-04-09
Compare pointers in find_vector_in_mf

peterrum:
Motivated by:

> The simple fix I initially though of would have been to first run through all partitioners in MatrixFree and check if the shared_ptr of the vector points to the same data as one of the partitioners/DoFHandlers in MatrixFree. That would be the one to pick. Such a mechanism would cover all cases where we do MatrixFree::initialize_dof_vector without further copies of the partitioners, which maybe helps in 90+% of the cases.

References #12025.


kronbichler (approved):
That is a good step forward. We might need to eliminate the global communication again if we can, so I think it's good to keep #12025 open for now to see if we can improve the local checks.

PR #12034, opened 2021-04-09
Step 79: A topology optimization problem

juleoc02:
Step 79: A topology optimization program


bangerth:
@juleoc02  is one of my students. I would appreciate if someone else could take a look at the program and provide feedback.

Justin: In one of the two .dox files, there is a place where you end a paragraph with `\\`. I think from looking at the latex document earlier that you do is quite a number of places. Just replace these places with an empty line -- something you should probably also do when you write plain latex. One of the automatic checks currently complains about that. (Once this is fixed, it will likely complain about many more things -- go to this issue on github and see what the red crosses for all of the checks are about by looking at the "Details" link.)

peterrum:
@@juleoc02 Thanks for the contribution. May I ask you post a PDF version of the tutorial. At this stage, I have problems to be build the documentation locally.

juleoc02:
> @@juleoc02 Thanks for the contribution. May I ask you post a PDF version of the tutorial. At this stage, I have problems to be build the documentation locally.

Absolutely! Here it is.

[TopoOptTutorial.pdf](https://github.com/dealii/dealii/files/6290468/TopoOptTutorial.pdf)

peterrum:
Thanks!

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+#include <fstream>
+#include <algorithm>
+
+// Above are fairly normal files to include.  These also include the
+// one for the sparse direct class SparseDirectUMFPACK. This is not
+// the most efficient way to solve large linear problems, but it will
+// do for now.
+//
+// As usual, we put everything into a common namespace. We then start
+// by declaring a number of symbolic names for numbers that will be
```
peterrum, line 64:
```suggestion
// by declaring a number of symbolic names for constants that will be
```

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+// used throughout. Specifically, we have a *lot* of variables in this
```
peterrum, line 65:
```suggestion
// used throughout this tutorial. Specifically, we have a *lot* of variables in this
```

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+//
+// Finally, while we're at it, we introduce symbolic names also for
+// the boundary indicators we will use, in the same spirit as was done
+// in step-19.
+//
+// In all of these cases, we declare these variables as members in a
+// namespace. In the case of the solution components, the concrete
+// values of these variables depend on the space dimension, so we use
+// [template
+// variables](https://en.cppreference.com/w/cpp/language/variable_template)
```
peterrum, line 89: does markdown work?

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+  // well as for the inequality constraints.
+  //
+  // The order in which these elements appear is documented above.
+  template <int dim>
+  SANDTopOpt<dim>::SANDTopOpt()
+    : fe(FE_DGQ<dim>(0) ^ 1,
+         (FESystem<dim>(FE_Q<dim>(1) ^ dim)) ^ 1,
+         FE_DGQ<dim>(0) ^ 1,
+         (FESystem<dim>(FE_Q<dim>(1) ^ dim)) ^ 1,
+         FE_DGQ<dim>(0) ^ 5)
```
peterrum, line 245: Is  `^ 1` needed?

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+                  }
+                else if (std::fabs(center(0) - 0) < 1e-12)
+                  {
+                    face->set_boundary_id(BoundaryIds::no_force);
+                  }
+                else if (std::fabs(center(0) - 6) < 1e-12)
+                  {
+                    face->set_boundary_id(BoundaryIds::no_force);
+                  }
+              }
```
peterrum, line 313: Why don't you use the coloring flag of `subdivided_hyper_rectangle()`?

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+                const auto center = face->center();
+
+                // Check whether the current face is on the bottom
+                // boundary, and if it is whether one of its
+                // vertices might be the bottom left or bottom
+                // right vertex:
+                if (std::fabs(center(1) - 0) < 1e-12)
+                  {
+                    for (unsigned int vertex_number = 0;
+                         vertex_number < GeometryInfo<dim>::vertices_per_cell;
```
peterrum, line 349: Let't not use `GeometryInfo`. here you could use `cell->vertex_indices()`

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+  }
+
+
+  // @sect3{Assembling the Newton matrix}
+
+  // Whereas the previous function built a matrix that is the same as
+  // long as the mesh does not change (which we don't do anyway in
+  // this program), the next function builds the matrix to be solved
+  // in each iteration. This is where the magic happens. The components
+  // of the system of linear equations describing Newtons' method for
```
peterrum, line 874:
```suggestion
  // of the system of linear equations describing Newton's method for
```

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+    const FEValuesExtractors::Scalar density_lower_slack_multipliers(
+      SolutionComponents::density_lower_slack_multiplier<dim>);
+    const FEValuesExtractors::Scalar density_upper_slacks(
+      SolutionComponents::density_upper_slack<dim>);
+    const FEValuesExtractors::Scalar density_upper_slack_multipliers(
+      SolutionComponents::density_upper_slack_multiplier<dim>);
+
+    QGauss<dim>       quadrature_formula(fe.degree + 1);
+    QGauss<dim - 1>   face_quadrature_formula(fe.degree + 1);
+    FEValues<dim>     fe_values(fe,
```
peterrum, line 911: Could we introduce a mapping?

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2756 @@
...
+        fe_values[unfiltered_densities].get_function_values(
+          filtered_unfiltered_density_solution,
+          filtered_unfiltered_density_values);
+        fe_values[unfiltered_density_multipliers].get_function_values(
+          filter_adjoint_unfiltered_density_multiplier_solution,
+          filter_adjoint_unfiltered_density_multiplier_values);
+
+        for (unsigned int q_point = 0; q_point < n_q_points; ++q_point)
+          {
+            for (unsigned int i = 0; i < dofs_per_cell; ++i)
```
peterrum, line 1029: Here, you could use iterators from `FEValues`, like `fe_values.quadrature_point_indices()`.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,397 @@
+<a name="Intro"></a>
+<h1>Introduction</h1>
+
+Topology Optimization of Elastic Media is a technique used to optimize a structure that is bearing some load. Ideally, we would like to minimize the maximum stress placed on a structure by selecting a region $E$ where material is placed.
```
peterrum, line 4: Could you break the lines at 80 characters.

peterrum (commented):
Amazing work! I have added some comments but need to go through the tutorial once more - once you have fixed the formatting and Doxygen is built successfully - to understand all the connections between the theory and the implementation.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,397 @@
...
+@f]
+
+The solution of the optimization problem then needs to satisfy what are known as the Karush-Kuhn-Tucker (KKT) conditions: The derivatives of the Lagrangian with respect to all of its arguments need to be equal to zero, and because we have inequality constraints, we also have "complementarity" conditions. Since we here have an infinite-dimensional problem, these conditions all involve directional derivatives of the Lagrangian with regard to certain test functions -- in other words, all of these conditions have to be stated in weak form as is typically the basis for finite element methods anyway.
+
+The barrier method allows us to initially weaken the ``complementary slackness" as required by the typical KKT conditions. Typically, we would require that $s_i z_i = 0$, but the barrier formulations give KKT conditions where $s_i z_i = \alpha$, where $\alpha$ is our barrier parameter. As part of the barrier method, this parameter must be driven close to 0 to give a good approximation of the original problem.
+
+In the following, let us state all of these conditions where
+$d_{\{\cdot\}}$ is a test function that is naturally paired with the $\{\cdot\}$ function.
```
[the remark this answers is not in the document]
juleoc02: This is because of the large numbers of different test functions I would need - u (my displacement) would pair nicely with v, but to make it consistent across all of my functions, I'm using this notation. It also alludes to the derivatives where they came from

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+    // boundary is given a separate boundary indicator: This is
+    // where we will apply the down force.
+    for (const auto &cell : triangulation.active_cell_iterators())
+      {
+        for (const auto &face : cell->face_iterators())
+          {
+            if (face->at_boundary())
+              {
+                const auto center = face->center();
+                if (std::fabs(center(1) - 0) < 1e-12)
```
tjhei, line 289: here and below for all other coordinates:
```suggestion
                if (std::fabs(center(1) - 0.) < 1e-12)
```

examples/step-79/step-79.cc
tjhei: yes, agreed. That would be much simpler.

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+                         vertex_number < GeometryInfo<dim>::vertices_per_cell;
+                         ++vertex_number)
+                      {
+                        const auto vert = cell->vertex(vertex_number);
+
+                        if (std::fabs(vert(0) - 0) < 1e-12 &&
+                            std::fabs(vert(1) - 0) < 1e-12)
+                          {
+                            const unsigned int x_displacement =
+                              cell->vertex_dof_index(vertex_number, 0);
```
tjhei, line 357: (here and below) the function returns a value of type ``types::global_dof_index``, please use that instead of ``unsigned int``.

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+  template <int dim>
+  void SANDTopOpt<dim>::setup_block_system()
+  {
+    std::vector<unsigned int> block_component(9, 2);
+    block_component[0] = 0;
+    block_component[1] = 1;
+    const std::vector<types::global_dof_index> dofs_per_block =
+      DoFTools::count_dofs_per_fe_block(dof_handler, block_component);
+
+    const unsigned int                                n_p = dofs_per_block[0];
```
tjhei, line 409:
```suggestion
    const types::global_dof_index                                n_p = dofs_per_block[0];
```

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+    const unsigned int                                n_u = dofs_per_block[1];
```
tjhei, line 410: same here

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+    // expressions lengthy and unwieldy: An term such as
+    // `coupling[SolutionComponents::density_upper_slack_multiplier<dim>][SolutionComponents::density<dim>]`
+    // just doesn't read very well, and would either have to be
+    // split over several lines or run off the right edge of
+    // nearly every screen. As a consequence, we open a
+    // curly-brace enclosed code block in which we temporarily
+    // make the names in namespace `SolutionComponents` available
+    // without the namespace qualifier, by saying `using namespace
+    // SolutionComponents`.
+    Table<2, DoFTools::Coupling> coupling(2 * dim + 7, 2 * dim + 7);
```
tjhei, line 446: I think the default is ``none``, so you should be able to remove all lines below that set ``::none``.

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+    const IndexSet      density_dofs =
+      DoFTools::extract_dofs(dof_handler, density_mask);
+
+    const unsigned int last_density_dof =
+      density_dofs.nth_index_in_set(density_dofs.n_elements() - 1);
+    constraints.clear();
+    constraints.add_line(last_density_dof);
+    for (unsigned int i = 1; i < density_dofs.n_elements(); ++i)
+      constraints.add_entry(last_density_dof,
+                            density_dofs.nth_index_in_set(i - 1),
```
tjhei, line 641:
```suggestion
    for (unsigned int i = 0; i < density_dofs.n_elements()-1; ++i)
      constraints.add_entry(last_density_dof,
                            density_dofs.nth_index_in_set(i),
```

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+    // with `using namespace` as above:
+    nonlinear_solution.reinit(block_sizes);
+    system_rhs.reinit(block_sizes);
+
+    {
+      using namespace SolutionBlocks;
+
+      for (unsigned int k = 0; k < n_u; ++k)
+        {
+          nonlinear_solution.block(displacement)[k]            = 0;
```
tjhei, line 751: vectors are initialized to 0 be default, so you can delete this.

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+          nonlinear_solution.block(displacement_multiplier)[k] = 0;
+        }
+      for (unsigned int k = 0; k < n_p; ++k)
+        {
+          nonlinear_solution.block(density)[k]            = density_ratio;
```
tjhei, line 756: I think you can replace the loop by `` nonlinear_solution.block(density).add(density_ratio);``, etc.

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+
+        std::set<unsigned int>                               neighbor_ids;
+        std::set<typename Triangulation<dim>::cell_iterator> cells_to_check;
+
+        neighbor_ids.insert(i);
+        cells_to_check.insert(cell);
+
+        unsigned int n_neighbors = 1;
+        filter_matrix.add(i, i, filter_r);
+        while (true)
```
tjhei, line 810: wouldn't it be easier and faster to iterate over the sparsity pattern instead of figuring out the close cells again?

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+                   size alpha, and gives complementary slackness
+                   from KKT conditions when $\alpha$ goes to 0. */
+                cell_rhs(i) +=
+                  -1 * fe_values.JxW(q_point) *
+                  (lower_slack_phi_i *
+                   (old_lower_slack_multiplier_values[q_point] -
+                    barrier_size / old_lower_slack_values[q_point]));
+
+                /* Equation 8 */
+                cell_rhs(i) +=
```
tjhei, line 1357: wow, that is quite a complicated system!

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+                fe_face_values.reinit(cell, face);
+
+                for (unsigned int face_q_point = 0;
+                     face_q_point < n_face_q_points;
+                     ++face_q_point)
+                  {
+                    for (unsigned int i = 0; i < dofs_per_cell; ++i)
+                      {
+                        Tensor<1, dim> traction;
+                        traction[1] = -1;
```
tjhei, line 1384:
```suggestion
                        traction[1] = -1.;
```

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+
+                        cell_rhs(i) +=
+                          -1 * traction *
```
tjhei, line 1387: I would add parens to multiply the two tensors first.

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2755 @@
...
+  // function above did.
+  template <int dim>
+  BlockVector<double> SANDTopOpt<dim>::calculate_test_rhs(
+    const BlockVector<double> &test_solution) const
+  {
+    BlockVector<double> test_rhs;
+    test_rhs.reinit(
+      system_rhs); /* a zero vector with size and blocking of system_rhs */
+
+    const FEValuesExtractors::Scalar densities(
```
tjhei, line 1574: you are using these extractors more than once. Maybe move them into a namespace?

tjhei (commented):
This is a very cool tutorial. The code is a bit on the long side, but I only see few ways to shorten. I left a couple of comments (but I did not read the whole code, sorry).

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,397 @@
...
+
+The straightforward application of the effect of this "density" on the elasticity of the media would be to simply multiply the stiffness tensor $C_0$ of the medium by the given density, that is, $C = \rho C_0$. However, this approach often gives optimal solutions where density values are far from both 0 and 1. As one wants to find a real-world solution, meaning the material either is present or it is not, a penalty is applied to these in-between values. A simple and effective way to do this is to multiply the stiffness tensor by the density raised to some integer power penalty parameter $p$, so that $C = \rho^p C_0$. This makes density values farther away from 0 or 1 less effective. It has been shown that using $p=3$ is sufficiently high to create 'black-and-white' solutions: that is, one gets optimal solutions in which material is either present or not present at all points.
```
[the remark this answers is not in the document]
juleoc02: I saw a discussion on the step 78 pull request that even though this looks funny in latex, It needs to look this way for doxygen

examples/step-79/step-79.cc
juleoc02: Just found these! That's really nice. :)

examples/step-79/doc/intro.dox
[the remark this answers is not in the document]
bangerth: Correct, these need to stay as is.

examples/step-79/step-79.cc
juleoc02: I think this returns the global vertex index instead of the local one - unless you're saying there's a way to find the number of vertices a cell has with vertex_index()?

examples/step-79/step-79.cc
juleoc02: They are not actually needed, but I like them - they make it look more uniform. 
I can of course change it if that would make it easier to read or if that would be normal practice.

examples/step-79/step-79.cc
juleoc02: I would need a way to get access to the neighboring cell (and specifically its center point) from the neighboring cell's index to do this, which I don't know how to do. This is because I then use the distance between the two cells. Is there a nice way to do that?

examples/step-79/step-79.cc
bangerth: Hm, I learned this syntax from looking at this program the first time around and changed all of the other tutorials to it. But that didn't find universal approval, see #11066. Would you mind switching to the old-style syntax?

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,397 @@
...
+@f]
+The final constraint, the elasticity equation, gives a method for finding $\sigma$ and $\epsilon$ given the density $\rho$.
+
+
+<h3>Elasticity Equation</h3>
+The elasticity equation in the time independent limit reads
+@f[
+  \nabla \cdot \sigma + \textbf{F} = 0.
+@f]
+In the situations we will care about, we will assume that the medium is linear and in that case, we have that
```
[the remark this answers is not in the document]
juleoc02: It looks like most tutorials use "we", so I'll change to that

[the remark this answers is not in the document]
bangerth: Yes. I changed that in a lot of places in your documentation already. We tend to think of tutorials as not work of a single person, but as part of the library -- different from the code gallery programs, which are definitely works of individuals.

examples/step-79/step-79.cc
juleoc02: I don't really understand what you're asking - sorry. I took a brief look at the mapping class documentation, but I'm not sure if that's what you're talking about. What would the mapping be for?

examples/step-79/step-79.cc
juleoc02: Yes! I just double checked on my locally built documentation and it's looking like it works fine. :)

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,397 @@
...
+      \int_\Omega d_{s_1}(s_1z_1 - \alpha) = 0 \;\;\forall d_{s_1} ,\;\;\; \alpha \to 0
+    @f]
+    @f[
+      \int_\Omega d_{s_2}(s_2z_2 - \alpha) = 0  \;\;\forall d_{s_2} ,\;\;\; \alpha \to 0
+    @f]
+
+\item Dual Feasibility:
+    @f[
+      s_{1,i},s_{2,i},z_{1,i},z_{2,i} \geq 0 \;\;\;\; \forall i
+    @f]
```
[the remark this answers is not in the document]
juleoc02: I changed this to clarify that these come from the derivatives of the lagrangian

examples/step-79/step-79.cc
juleoc02: Easy enough!

examples/step-79/step-79.cc
juleoc02: While looking into the coloring flag, I realized I actually only use this information for those parts of the boundary that get assigned the down_force boundary id. Would it still be worth doing the coloring when I shorten this to only assign that part of the boundary?

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2627 @@
...
+            // Specifically, from the final (unsuccessful) state
+            // of the loop above, we seek one more update
+            // direction and take what is called a "stretch
+            // step". If that stretch state satisfies a condition
+            // involving the merit function, then we go there. On
+            // the other hand, if the stretch state is also
+            // unacceptable (as all of the watchdog steps above
+            // were), then we discard all of the watchdog steps
+            // taken above and start over again where we had
+            // started the watchdog iterations -- that places was
```
marcfehling, line 2481:
```suggestion
            // started the watchdog iterations -- that place was
```

examples/step-79/doc/results.dox
```diff
@@ -0,0 +1,72 @@
...
+
+<img src="https://www.dealii.org/images/steps/developer/step-6.filtereddensity.png"
+ alt="Filtered Density Solution">
+
+<img src="https://www.dealii.org/images/steps/developer/step-6.unfiltereddensity.png"
+ alt="Unfiltered Density Solution">
+
+
+These pictures show that what we find here is in accordance with what one
+typically sees in other publications on the topic. Maybe more interestingly, the
```
marcfehling, line 41: Can you cite one or two publications here?

marcfehling (changes requested):
Thank you for the contribution! I like the way you write extensive documentation. I have a couple of suggestions that you will hopefully agree on :-)

examples/step-79/step-79.cc
juleoc02: I cannot get this to work, or find any documentation about cell->vertex_indices() or similar - I'm worried I'm looking in the wrong place

peterrum: https://www.dealii.org/developer/doxygen/deal.II/classTriaAccessor.html#ae40947cf863ecef1ee08ec93f411baaa

I guess you are not on the developer version of the documentation, are you?

juleoc02: That makes sense. Thank you!

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+  \text{s.t.   } \rho = H(\sigma)
+@f]
+@f[
+  \int_\Omega \rho^p \left(\frac{\mu}{2}\left(\epsilon(\textbf{v}):
+  \epsilon(\textbf{u})) \right) + \lambda \left( \nabla \cdot \textbf{u} \nabla
+  \cdot \textbf{v} \right)  \right) = \int_{\partial \Omega} \textbf{v} \cdot
+  \textbf{t}
+@f]
+@f[
+  \int_\Omega \rho = V
```
jppelteret, line 199:
```suggestion
  \int_\Omega \rho dV = V
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+@f]
+@f[
+  0\leq \sigma \leq 1
```
jppelteret, line 202: Which `\sigma` is this?

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+  \min_{\rho,\sigma,\mathbf u} \int_{\partial\Omega} \textbf{u} \cdot \textbf{t}
+@f]
+@f[
+  \text{s.t.   } \rho = H(\sigma)
+@f]
+@f[
+  \int_\Omega \rho^p \left(\frac{\mu}{2}\left(\epsilon(\textbf{v}):
+  \epsilon(\textbf{u})) \right) + \lambda \left( \nabla \cdot \textbf{u} \nabla
+  \cdot \textbf{v} \right)  \right) = \int_{\partial \Omega} \textbf{v} \cdot
+  \textbf{t}
```
jppelteret, line 196:
```suggestion
  \int_\Omega \rho^p \left(\frac{\mu}{2}\left(\boldsymbol{\varepsilon}(\textbf{v}):
  \boldsymbol{\varepsilon}(\textbf{u})) \right) + \lambda \left( \nabla \cdot \textbf{u} \nabla
  \cdot \textbf{v} \right)  \right) dV = \int_{\partial \Omega} \textbf{v} \cdot
  \textbf{t} dA
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+  \int_\Omega \rho = V
+@f]
+@f[
+  0\leq \sigma \leq 1
+@f]
+
+The inequality constraints are dealt with by first introducing slack variables,
+and second using a log barriers to ensure that we obtain an interior-point
+method. The penalty parameter is going to be $\alpha$, and the following slack
+variables
```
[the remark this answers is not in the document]
jppelteret: I agree with Marc. Without this, the sentence appears to be incomplete.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+  \forall d_\rho
+@f]
+@f[
+  \int_\Gamma \mathbf d_\textbf{u} \cdot \textbf{t} + \int_\Omega p\rho^{p}
+  \left[\lambda (\nabla \cdot \mathbf d_\textbf{u})( \nabla \cdot \textbf{y}_1)
+  + \mu \varepsilon(\mathbf d_\textbf{u}):\varepsilon(\textbf{y}_1)\right]=0\;\;
+  \forall \textbf{d}_\textbf{u}
+@f]
+@f[
+  \int_\Omega -d_\sigma z_1 + d_\sigma z_2 + H(d_\sigma)y_2 = 0\;\;\forall
```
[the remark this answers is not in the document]
jppelteret: I think that at this point I'm going to have to abandon making the suggested changes that add the differentials to the integrals. Please add them to the relevant equations that follow :-)

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+  d_\sigma
+@f]
+</li>
+<li> Primal Feasibility:
+    @f[
+      \int_\Omega \rho^{p}\lambda (\nabla \cdot \mathbf d_{\textbf{y}_1})
+      (\nabla \cdot \textbf{u}) +  \rho^{p}\mu  \varepsilon(\mathbf
+      d_{\textbf{y}_1}) : \varepsilon(\textbf{u}) - \int_\Gamma \mathbf
+      d_{\textbf{y}_1} \cdot \textbf{t} =0 \;\;\forall \textbf{d}_{\textbf{y}_1}
```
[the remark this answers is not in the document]
jppelteret: I'm also going to leave the rest of the changes that lift the small strain tensor from the scalar like notation `\varepsilon` to the tensor notation `\boldsymbol{\varepsilon}`. It should be easy enough to do a search-replace for this. (I should probably have checked how many equations you had before I started off on my quest to assist :-D )

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+
+Newton's method applied to the equations above results in the following system.
+Here, variational derivatives with respect to the $\{\cdot\}$ variable are taken
+ in the $c_{\{\cdot\}}$ direction. This gives
+
+<ol>
+<li> Stationarity - these equations ensure we are at a critical point of the
+objective function when constrained
+
+Equation 0
```
jppelteret, line 349: A bit pedantic, but maybe its not worth starting equation labels at 0?

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+\textbf{c}_\textbf{u} +  \mu  \varepsilon(\textbf{c}_\textbf{u})
+\varepsilon(\textbf{d}_{\textbf{y}_1})]
+\\
+ =-\int_\Omega \rho^{p}[\lambda \nabla \cdot \textbf{d}_{\textbf{y}_1} \nabla
+ \cdot \textbf{u} + \mu  \varepsilon(\textbf{u}) \varepsilon
+ (\textbf{d}_{\textbf{y}_1})]  + \int_\Gamma  \textbf{d}_{\textbf{y}_1}
+ \cdot \textbf{t}
+@f}
+
+    Equation 4
```
jppelteret, line 398: Hmm... perhaps you could tidy this up (the line indentation for the "Equation" headers, that is).

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+  \cdot \textbf{d}_\textbf{u} +  \mu  \varepsilon(\textbf{d}_\textbf{u})
+  \varepsilon(\textbf{y})] + \rho^{p} [\lambda \nabla \cdot
+  \textbf{c}_{\textbf{y}_1} \nabla \cdot \textbf{d}_\textbf{u} +  \mu
+  \varepsilon(\textbf{d}_\textbf{u})\varepsilon(\textbf{c}_{\textbf{y}_1})]
+\\
+=-\int_\Gamma \textbf{d}_\textbf{u} \cdot \textbf{t} -\int_\Omega \rho^{p}
+[\lambda \nabla \cdot \textbf{y} \nabla \cdot \textbf{d}_\textbf{u} + \mu
+\varepsilon(d_\textbf{u})\varepsilon(\textbf{y}_1)]
+@f}
+Equation 2
```
jppelteret, line 375:
```suggestion
@f}

Equation 2
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+      (z_1 - \alpha/s_1) ,\;\;\; \alpha \to 0
+    @f]
+
+    Equation 8
+    @f[
+      \int_\Omega d_{s_2} (c_{s_2}z_2/s_2 + c_{z_2} )=-\int_\Omega d_{s_2}
+      (z_2 - \alpha/s_2) ,\;\;\; \alpha \to 0
+    @f]
+
+    Dual Feasibility - Multiplier on slacks and slack variables must be kept
```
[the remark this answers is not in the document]
jppelteret: Agreed.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+
+<h3>Nonlinear Algorithm</h3>
+
+While most of the discussion above follows traditional and well-known approaches
+ to solving nonlinear optimization problems, it turns out that the problem is
+ actually quite difficult to solve in practice. In particular, it is quite
+ nonlinear and an important question is not just to find search directions
+ $c_\bullet$ as discussed above based on a Newton method, but that one needs to
+ spend quite a lot of attention to how far one wants to go in this direction.
+ This is often called ``line search" and comes down to the question of how to
```
jppelteret, line 453:
```suggestion
 This is often called "line search" and comes down to the question of how to
```
Double check this -- I seem to recall someone else mentioning just a few days ago that `doxygen` doesn't like these LaTeX style quotes.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+ to discretize the density, unfiltered density, density slack variables, and
+ multipliers for the slack variables and filter constraint.
+
+<h3>Nonlinear Algorithm</h3>
+
+While most of the discussion above follows traditional and well-known approaches
+ to solving nonlinear optimization problems, it turns out that the problem is
+ actually quite difficult to solve in practice. In particular, it is quite
+ nonlinear and an important question is not just to find search directions
+ $c_\bullet$ as discussed above based on a Newton method, but that one needs to
```
jppelteret, line 451: Earlier you used the notation `$c_{\{\cdot\}}$` for the search direction. Can you please make this uniform?

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+ iterate $x_k$ to the next iterate $x_{k+1}=x_k+\alpha_k x_k$ in as efficient a
+ way as possible. It is well understood that we need to eventually choose
+ $\alpha_k=1$ to realize the Newton's method's quadratic convergence; however,
+ in the early iterations, taking such a long step might actually make things
+ worse, either by leading to a point that has a worse objective function or at
+ which the constraints are satisfied less well than they are at $x_k$.
+
+Very complex algorithms have been proposed to deal with this issue.
+@cite Nocedal2009 @cite Wchter2005 Here, we implement a watchdog-search
+algorithm.@cite Nocedal2006 When discussing this algorithm, we will use the
```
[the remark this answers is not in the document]
jppelteret: I agree with Marc's suggested change here.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+vector $\textbf{x}$ to represent all primal variables - the filtered and
+unfiltered densities, slack variables and displacement - and use the vector
+$\textbf{y}$ to represent all of the dual vectors. The solution to the Newton
+System above will now be referred to as $\Delta \textbf{x}$ and $\Delta
```
jppelteret, line 468:
```suggestion
$\textbf{y}$ to represent all of the dual vectors. The (incremental) solution to the nonlinear
system of equations stated above will now be referred to as $\Delta \textbf{x}$ and $\Delta
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+\textbf{y}$ instead of $c_{\{\cdot\}}$. A merit function is referred to here as
+$\phi(\textbf{x,\textbf{y}})$
+
+The watchdog algorithm applied to a subproblem with a given barrier parameter
```
jppelteret, line 472: Do you have a link (Wikipedia, perhaps) for "watchdog algorithm"? It might be nice to give some "further reading" to those of us that have not come across the concept before (or perhaps know it by another name?).

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+ which the constraints are satisfied less well than they are at $x_k$.
+
+Very complex algorithms have been proposed to deal with this issue.
+@cite Nocedal2009 @cite Wchter2005 Here, we implement a watchdog-search
+algorithm.@cite Nocedal2006 When discussing this algorithm, we will use the
+vector $\textbf{x}$ to represent all primal variables - the filtered and
+unfiltered densities, slack variables and displacement - and use the vector
+$\textbf{y}$ to represent all of the dual vectors. The solution to the Newton
+System above will now be referred to as $\Delta \textbf{x}$ and $\Delta
+\textbf{y}$ instead of $c_{\{\cdot\}}$. A merit function is referred to here as
```
jppelteret, line 469: Excuse my ignorance, but what is a merit function? My guess would be a function that somehow assigns a score to the quality of the proposed solution?

EDIT: I see that you actually introduce the concept a little later on. You might wish to mention that here.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+@f]
+@f[
+  \beta^\textbf{x} = \min\{1,\max \beta \text{ such that }\left(\textbf{s}_{k+i}
+   + \beta^\textbf{s}_{k+i} \Delta \textbf{s}_{k+i}\right)_j \geq \zeta
+   \textbf{s}_{k+i,j} \forall j\}
+@f]
+
+Above, $\zeta$ is the "fraction to boundary" that is allowed on any step.
+Because the derivatives become ill-conditioned near the boundary, this technique
+ stands in for a trust region and is necessary to ensure good approximations in
```
jppelteret, line 505:
```suggestion
 stands in for a [trust region](https://en.wikipedia.org/wiki/Trust_region) and is necessary to ensure good approximations in
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+ $\alpha$ and gradually decrease it as the outer iterations proceed.
+
+In the monotone method implemented here, the barrier parameter is updated
+whenever some level of convergence is achieved at the current barrier parameter.
+ We use the $l_\infty$ norm of the KKT conditions to check for convergence at
+ each barrier size. The requirement is that
+ $\norm{KKT}_{l_\infty} < c \cdot \alpha$ where $c$ is a constant over any
+ barrier size and $\alpha$ is the barrier parameter. This forces better
+ convergence in later iterations, and is the same requirement as is used in
+ IPOPT.
```
jppelteret, line 532:
```suggestion
[IPOPT](https://coin-or.github.io/Ipopt/) (an open source software package for large-scale nonlinear optimization).
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+
+Here, the barrier is reduced linearly at larger values, and superlinearly at
+smaller values. At larger values, it is multiplied by a constant (around .6),
+and at lower values the barrier value is replaced by the barrier value raised
+to some exponent (around 1.2). This method has proven to be effective at keeping
+ the subproblem solvable at large barrier values, while still allowing
+ superlinear convergence at smaller barrier values.
+@f[
+  \alpha_{k+1} = \min\{\alpha_k^{1.2},.6\alpha_k\}
```
jppelteret, line 541: This equation requires some introduction. Its just "floating" below the paragraph above.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+@f]
+
+While taking large steps at reducing the barrier size when convergence is
+reached is widely used, more recent research has shown that it typically faster
+to use algorithms that adaptively update barrier each iteration, i.e., in which
```
jppelteret, line 546:
```suggestion
to use algorithms that adaptively update barrier each iteration, i.e., methods in which
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+factors that are independent of the current solution. That said, such methods
+are also more complicated and we will not do this here.
+
+<h3>Merit Function</h3>
+
+The algorithm outlined above makes use of a "merit function". Merit functions
+are used to determine whether a step from $x_k$ to a proposed point $x_{k+1}$ is
+ beneficial. In unconstrained optimization problems, one can simply check this
+ with the object function we try to minimize, and typically uses conditions such
+  as the Wolfe and Goldstein conditions.
```
jppelteret, line 558:
```suggestion
  as the [Wolfe and Goldstein conditions](https://en.wikipedia.org/wiki/Wolfe_conditions).
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+
+In constrained optimization problems, the question is how to balance reduction
+in the objective function against a possible increase in the violation of
+constraints: A proposed step might make the objective function smaller but be
+further away from the set of points that satisfy the constraints -- or the other
+ way around. This trade-off is typically resolved by using a merit function that
+  combines the two criteria.
+
+Here, we use an exact $l_1$ merit function to test the steps.
```
jppelteret, line 567:
```suggestion
Here, we use an exact $l_1$ merit function to test the steps:
```

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+
+Here, $p$ is a penalty parameter. This merit function being exact means that
+there exists some $p_0$ so that for any $p > p_0$, the merit function has its
+minima at the same location as the original problem. This penalty parameter is
+updated (by recommendation of Nocedal and Wright @cite Benson2002) as follows:
+@f[
+  p > \frac{.5 x^T H x - x^t\nabla f}{\norm{c_i}_{l_\infty}, i \in \mathcal{E}}
+@f]
+Where $H$ is the hessian of the objective function, $x$ is a vector of our
+decision variables, $f$ is the objective function, and $c_i$ is the error on a
```
jppelteret, line 589: You've not used the term "decision variables" before. What does this mean?

jppelteret (commented):
I've read through the introduction thus far, and its mightily impressive. Awesome work :-) Although there are quite a number of comments, you should not be discouraged. I was trying to consistently change some of the equations, but in the end there was just too much to do and it would probably be easier for you to do yourself with a search-replace anyway. But in between I do have a couple of points related to the section on solid mechanics that need addressing. I also have a few questions that arise due to my complete lack of knowledge on constrained optimization (too many years have elapsed between now and my 2nd year calculus class). But I reckon that if I have these questions, then perhaps others that are new to the field would have them too. Perhaps you wouldn't mind spending a few minutes thinking about them?

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+$V_\text{max}$ is the maximal amount of material we are willing to provide to
+build the object. The last of the constraints is the partial differential
+equation that relates stress $\sigma$ and forces $\mathbf F$ and is simply the
+stead-state force balance.
+
+That said, the infinity norm above creates a problem: As a function of location
+of material, this objective function is necessarily not differentiable, making
+prospects of optimization rather bleak. So instead, a common approach in
+topology optimization is to find an approximate solution by optimizing a related
+problem: We would like to minimize the strain energy, or compliance. This is a
```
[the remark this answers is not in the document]
juleoc02: Compliance is sometimes used to talk about strain energy in Topology Optimization stuff. I have not seen it anywhere else, and I agree that it is rather confusing wording. I'll go through and make it all strain energy, as it's much more universal a term.

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+measure of the potential energy stored in an object due to its deformation, but
+also works as a measure of total displacement over the structure.
+
+@f[
+  \text{Minimize  } \int_E \frac{1}{2}\sigma : \epsilon
```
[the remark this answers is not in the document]
juleoc02: Absolutely! I

examples/step-79/doc/intro.dox
juleoc02: It appears it doesn't like the latex quotes

jppelteret:
@juleoc02 I've opened up a PR https://github.com/juleoc02/dealii/pull/1 that suggests some more changes to the equations. Can you please take a look at it, and we can discuss it's contents in that PR. Thanks!

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+mesh is refined further, fractal structures are often formed. As the mesh gains
+resolution, the optimal solution typically gains smaller and smaller structures.
+There are a few competing work-arounds to this issue, but the most popular for
+first order optimization is the sensitivity filter, while second order
+optimization methods tend to prefer use of a density filter.
+
+As the filters affect the gradient and Hessian of the compliance (i.e., the
+objective function), the choice of filter has an effect on the solution of the
+problem. The density filter as part of a second order method works by
+introducing an unfiltered density, which we refer to as $\sigma$, and then
```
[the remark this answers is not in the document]
juleoc02: I agree with this - I was looking for a symbol that would naturally pair with \rho, which is the density. I think that as the stress is now bold they are distinguishable, but it isn't ideal

examples/step-79/doc/intro.dox
juleoc02: The not bold one. :) The cauchy stress only appears in the introductory part of the intro (making it very clear to me which I was using before I shoved it all together and realized I used the same thing twice)

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+  \int_\Omega  - d_\rho y_2 + p\rho^{p-1}d_\rho \left[\lambda
+  (\nabla \cdot \textbf{y}_1) (\nabla \cdot \textbf{u}) +
+  \mu \varepsilon(\textbf{u}):\varepsilon(\textbf{y}_1)\right]=0\;\;
+  \forall d_\rho
+@f]
+@f[
+  \int_\Gamma \mathbf d_\textbf{u} \cdot \textbf{t} + \int_\Omega p\rho^{p}
+  \left[\lambda (\nabla \cdot \mathbf d_\textbf{u})( \nabla \cdot \textbf{y}_1)
+  + \mu \varepsilon(\mathbf d_\textbf{u}):\varepsilon(\textbf{y}_1)\right]=0\;\;
+  \forall \textbf{d}_\textbf{u}
```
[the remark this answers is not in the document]
juleoc02: It represents the Neumann part of the boundary - I just added an explanation further up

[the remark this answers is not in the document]
juleoc02: I am using d\Omega and d\partial\Omega - I'm a little curious about your thoughts on the second one, it feels kind of weird, but seems more appropriate for going between 2d/3d than saying dV and dA

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2627 @@
...
+                      {
+                        const double distance = cell->center().distance(
+                          check_cell->neighbor(n)->center());
+                        if ((distance < filter_r) &&
+                            !(neighbor_ids.count(
+                              check_cell->neighbor(n)->active_cell_index())))
+                          {
+                            cells_to_check_temp.insert(check_cell->neighbor(n));
+                            neighbor_ids.insert(
+                              check_cell->neighbor(n)->active_cell_index());
```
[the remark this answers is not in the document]
juleoc02: I agree. This looks way better

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2627 @@
...
+    // Having so generated the "dynamic" sparsity pattern, we can
+    // finally copy it to the structure that is used to associate
+    // matrices with a sparsity pattern. Because the sparsity
+    // pattern is large and complex, we also output it into a file
+    // of its own for visualization purposes -- in other words,
+    // for "visual debugging".
+    sparsity_pattern.copy_from(dsp);
+
+    std::ofstream out("sparsity.plt");
+    sparsity_pattern.print_gnuplot(out);
```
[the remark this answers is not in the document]
juleoc02: I put this here just to double check the sparsity pattern. I think that it makes sense here, right after the pattern is made, for the "visual debugging" purpose, but I don't have terribly strong feelings about this and can move it quite easily

examples/step-79/doc/intro.dox
juleoc02: Decision variables is a fairly common term in linear/nonlinear programming for the variables that can be changed in the objective function. Because I'm using log barriers, the slack variables are a part of the barrier function that I treat as the objective function, so it's actually exactly all of my primal variables here - I can make that change pretty easily.

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2627 @@
...
+              }
+
+            if (neighbor_ids.size() == n_neighbors)
+              break;
+            else
+              {
+                cells_to_check = cells_to_check_temp;
+                n_neighbors    = neighbor_ids.size();
+              }
+          }
```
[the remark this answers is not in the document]
juleoc02: I fixed the while statement...
As per the other kinds of elements issue, the cells in these problems are typically viewed as 'voxels', with the goal of forcing each 'voxel' to either 0 or 1, so piecewise constant is really the only thing that makes sense for the density variable. I can still try to change it as having more options could be interesting, but I don't think it's ever really used.

examples/step-79/doc/intro.dox
juleoc02: I can't find an easy link, but I do have the paper that introduced it. I'll cite it here.

examples/step-79/doc/intro.dox
juleoc02: Decision variables is typically a way of talking about the variables in the objective function, so primal not-slack variables. That said, because I am actually using log barriers, I replace the objective function a barrier function that does involve the slacks, so in the case of this problem it is all of the primal variables. Decision variables is more technically correct here as it translates better to other problems, but I could easily replace this with primal variables and have it still be correct.

jppelteret: Then it sounds perfectly reasonable to want to keep this terminology in place. But could you then write it as "decision (primal) variables" or "primal (decision) variables"? Since you introduce it as "our decision variables", one gets the feeling that one should be familiar with the term after reading the preceding paragraphs (which would not be the case for readers not familiar with topology optimisation).

examples/step-79/doc/intro.dox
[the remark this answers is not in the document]
jppelteret: > It represents the Neumann part of the boundary - I just added an explanation further up

Great, thanks a lot!

> I am using d\Omega and d\partial\Omega - I'm a little curious about your thoughts on the second one, it feels kind of weird

It is a _little_ awkward, but I wouldn't say that its incorrect. Looking at some of the other tutorials (step-8, step-18, step-41), they all use bilinear forms which circumvents the issue. But what I do notice is that they use `\Gamma` for the boundaries, with `\Gamma_{D}` and `\Gamma_{N}` respectively denoting the Dirichlet and Neumann boundaries. If you're willing to follow suite, then you could use `d\Gamma` for the boundary terms as well.

> but seems more appropriate for going between 2d/3d than saying dV and dA

I hadn't thought about that. That's a fair point!

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2444 @@
...
+        fe_values[unfiltered_densities<dim>].get_function_values(
+          test_solution, old_unfiltered_density_values);
+        fe_values[unfiltered_density_multipliers<dim>].get_function_values(
+          test_solution, old_unfiltered_density_multiplier_values);
+        fe_values[unfiltered_densities<dim>].get_function_values(
+          filtered_unfiltered_density_solution,
+          filtered_unfiltered_density_values);
+        fe_values[unfiltered_density_multipliers<dim>].get_function_values(
+          filter_adjoint_unfiltered_density_multiplier_solution,
+          filter_adjoint_unfiltered_density_multiplier_values);
```
tyand96, line 1382: This looks similar to what you were doing above in that huge block of code. Would it make sense to put this into a function?

examples/step-79/doc/intro.dox
```diff
@@ -0,0 +1,597 @@
...
+introducing an unfiltered density, which we refer to as $\sigma$, and then
+requiring that the density be a convolution of the unfiltered density:
+@f[
+  \rho = H(\sigma).
+@f]
+Here, $H$ is an operator so that $\rho(\mathbf x)$ is some kind of average of
+the values of $\sigma$ in the area around $\mathbf x$ -- i.e., it is a smoothed
+version of $\sigma$.
+
+This prevents checkerboarding; the radius of the filter allows the user to
```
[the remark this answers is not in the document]
juleoc02: I added one. That's exactly what it is

examples/step-79/step-79.cc
```diff
@@ -0,0 +1,2444 @@
...
+    constexpr unsigned int density_lower_slack = 3 + 2 * dim;
+    template <int dim>
+    constexpr unsigned int density_lower_slack_multiplier = 4 + 2 * dim;
+    template <int dim>
+    constexpr unsigned int density_upper_slack = 5 + 2 * dim;
+    template <int dim>
+    constexpr unsigned int density_upper_slack_multiplier = 6 + 2 * dim;
+  } // namespace SolutionComponents
+
+  namespace SolutionBlocks
```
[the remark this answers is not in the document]
jppelteret: Two suggestions (just putting this out there): 
- For `SolutionComponents`, you could make each of these reference the one above so there's less that's hardcoded. All that each entry needs to know is the entry before it and how many components that entry had. That also self-documents that there's a hierarchy that's being constructed.  It you changed the namespace to a struct, then you could also have the template in a single place:
```c++
  template <int dim>
  struct SolutionComponents
  {
    constexpr unsigned int density = 0;
    constexpr unsigned int displacement = density + 1;
    constexpr unsigned int unfiltered_density = displacement + dim;
    constexpr unsigned int displacement_multiplier = unfiltered_density + 1;
    ...
  } // namespace SolutionComponents
```
- As for the `SolutionBlocks`, those could just be an enum with `density=0` and the rest unnumbered. Then you wouldn't need to explicitly write all of that out.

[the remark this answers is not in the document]
bangerth: And there I was so proud of the template variables in a namespace...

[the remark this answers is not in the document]
juleoc02: The nice thing about the namespace is that, because I typically use a lot of these all at once, I don't have to have SolutionBlocks:: before every one, which might make some of these already hard to read parts even a bit more cumbersome to read ([Like here that Tyler mentioned](https://github.com/dealii/dealii/pull/12034#discussion_r630369908)).

marcfehling:
I've built the tutorial [documentation locally](https://www.dealii.org/developer/readme.html#documentation) and noticed a few things regarding the proper display of the documentation. I've addressed some of them in https://github.com/juleoc02/dealii/pull/3. There are still a few things I would like to ask you to fix:
- `doxygen` (or rather `mathjax`) doesn't know about the LaTeX command `\norm` and the environment `multiline`. Would you please find substitutes?
- Did you create the picture `step-79.mbbgeometry.png` yourself, or did you take it from somewhere? If so, we need to give credit for the creators, or diretly link to the source.

bangerth:
If you want to improve that last commit a bit more, use `\| ... \|` for norms, instead of `|| ... ||`. This makes sure the spacing of the two bars is correct.

marcfehling:
You still have issues with the `multiline` environment.

marcfehling (approved):

marcfehling:
Please squash your commits into one with `git rebase`. Since you merged with the master branch in your commit history, this can get tricky. Let us know if you need help.

marcfehling:
Please squash your commits into one with `git rebase`. Since you merged with the master branch in your commit history, this can get tricky. Let us know if you need help.

marcfehling:
The `indent` worker fails with the following message:
`/home/runner/work/dealii/dealii/build/doc/doxygen/tutorial/step-79.h:3515: warning: Unexpected html tag  found within <a href=...> context`

Apparently this has been fixed in newer versions of `doxygen`. I think we can get rid of the hyperlink here as you placed the literature reference right before the picture.

marcfehling:
I squashed your commit history into one commit. I created a [backup](https://github.com/marcfehling/dealii/tree/juleoc02-step-79-backup) just in case.

marcfehling:
Rebased on master.

PR #12039, opened 2021-04-13
Get rid of create_triangulation_compatibility in the UNV reader.

drwells:
Part of #8562.

This case is interesting since the current code only works by coincidence. Our test files have vertices numbered like

    2 ---- 3
    |      |
    |      |
    1 ---- 0

which we read in in the old-style (UCD) numbering, which is like mirroring across the y-axis. This causes all the cells to have a negative measure (but they aren't pinched). GridReordering then fixes this by swapping vertices 1 and 3 of every single cell - i.e., adding another -1 factor to the measure, which now causes all cells to have positive measure. Something similar happens in 3D.

As far as I can tell there is no widely accepted node numbering for UNV like there is for everything else so I am going by what Salome generates. Since we now use their numbering directly (instead of something that worked by coincidence) the output files are slightly different but all cells do end up looking correct (nothing is twisted when we visualize with VTK).


Rombur (approved):
This looks good. Since there is a visible effect for the user, you should add an entry in the changelog.

drwells:
Good point - I added a changelog entry.

kronbichler (approved):

PR #12042, opened 2021-04-14
Get rid of create_triangulation_compatibility in the UCD reader.

drwells:
Also part of #8562.

### DIFF
--- a/source/grid/grid_in.cc
+++ b/source/grid/grid_in.cc
@@ -946,7 +946,7 @@ GridIn<dim, spacedim>::read_ucd(std::istream &in,
           // allocate and read indices
           cells.emplace_back();
           for (const unsigned int i : GeometryInfo<dim>::vertex_indices())
-            in >> cells.back().vertices[i];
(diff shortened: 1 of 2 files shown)


kronbichler (approved):

peterrum (approved):
Is here the same true as here https://github.com/dealii/dealii/pull/12039#pullrequestreview-635164672?

masterleinad (approved):

drwells:
> Is here the same true as in [#12039 (review)](https://github.com/dealii/dealii/pull/12039#pullrequestreview-635164672)?

In this particular case, no. The old-style ordering is the same as UCD, so this function just does the reordering here instead of later - the end result is identical.

drwells:
Then again some tests fail now so that may not be true for code that calls this code. I'll investigate.

drwells:
The ouputs were slightly different since we used an acceptable (but different) reordering for hexahedra with negative volume - the commit I just pushed should fix the issue but it might mess up some output changed in #11784.

peterrum (approved):

Rombur (approved):

PR #12043, opened 2021-04-15
Remove hp::DoFHandler from tutorial list.

marcfehling:
After merging both DoFHandlers in #10328, we should not refer to the deprecated class anymore.


masterleinad (approved):
Looks good to me.

PR #12044, opened 2021-04-15
CMake: FindKOKKOS: Work around configuration failure

tamiko:
Configuration did fail in the following situation
 - Kokkos installed via Trilinos (MPI enabled)
 - deal.II configured without MPI

In this situation FindTRILINOS.cmake bails out because of an MPI
configuration mismatch, so the manual search for Kokkos in
FindKOKKOS.cmake is triggered. The latter one fails because Kokkos
installed via Trilinos did export a valid project configuration only
containing the Kokkos::kokkos target and not the Kokkos::kokkoscore
target.

Work around this issue by only configuring if Kokkos::kokkoscore has
indeed been imported.


masterleinad (approved):
Looks OK to me. We normally export
```
Kokkos::kokkoscore Kokkos::kokkoscontainers Kokkos::kokkosalgorithms Kokkos::kokkos
```
in a standalone build.

Rombur (approved):

PR #12047, opened 2021-04-15
FEInterfaceValues extractors

zjiaqi2018:
Added FEInterfaceValues extractors, two simple tests and the Stokes test.


tjhei:
Great, thank you! 

@blaisb this is what you were asking about, right?

tjhei:
You forgot to remove two large output files that are no longer in use.

tjhei:
Can you edit/squash your commits to get rid of the addition followed by deletion of the large files? They will be part of the history otherwise.

zjiaqi2018:
> Can you edit/squash your commits to get rid of the addition followed by deletion of the large files? They will be part of the history otherwise.
Yes.

tjhei:
Thank you. I will let somebody else review this code, as it is partly written by me. :-)

tests/feinterface/stokes.cc
```diff
@@ -0,0 +1,2542 @@
+/* ---------------------------------------------------------------------
```
masterleinad, line 1: There is a lot of commented code in this test. What do we actually test here?

tests/feinterface/stokes.output
```diff
@@ -0,0 +1,256 @@
...
+DEAL::	Number of degrees of freedom: 352 (288+64)
+DEAL:FGMRES::Starting value 13.
+DEAL:FGMRES:cg::Starting value 1.2e-33
+DEAL:FGMRES:cg::Convergence step 1 value 1.9e-49
+DEAL:FGMRES:cg::Starting value 1.0
+DEAL:FGMRES:cg::Convergence step 1 value 4.9e-15
+DEAL:FGMRES:cg::Starting value 1.0
+DEAL:FGMRES:cg::Convergence step 1 value 2.0e-16
+DEAL:FGMRES:cg::Starting value 7.5
+DEAL:FGMRES:cg::Convergence step 1 value 3.0e-14
```
masterleinad, line 14: Can we avoid printing these numbers. They tend to differ for different compiler setups and we are probably only interested in errors here and not any solver information.

include/deal.II/fe/fe_interface_values.h
```diff
@@ -497,8 +862,11 @@ FEInterfaceValues<dim, spacedim>::FEInterfaceValues(
   , internal_fe_subface_values_neighbor(mapping, fe, quadrature, update_flags)
   , fe_face_values(nullptr)
   , fe_face_values_neighbor(nullptr)
+  , fe_interface_views_cache(*this)
```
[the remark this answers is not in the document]
tjhei: I don't understand what you mean by this, can you clarify please?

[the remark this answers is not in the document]
masterleinad: I am not quite clear what the purpose of `fe_interface_views_cache` is. Why do we need `fe_interface_views_cache`, `fe_face_values` and `fe_face_values_neighbor`? How are they related?

tests/feinterface/stokes.cc
tjhei: Agreed. That is my fault for handing him a large and incomplete test code. Jiaqi, please remove from thi PR. We will submit additional tests later.

bangerth (commented):
Nice work -- we needed this!

I think there is potential for future optimization (but not necessary for this patch). If you look at the implementation of functions such as
```
  typename Base<dim, spacedim>::value_type
  Scalar<dim, spacedim>::jump(const unsigned int interface_dof_index,
                              const unsigned int q_point) const
  {
    return this->fe_interface->jump(interface_dof_index,
                                    q_point,
                                    extractor.component);
  }
```
then the `FEInterfaceValues::jump()` function needs to go through a bunch of hoops to figure out whether the selected component of the given shape function is actually nonzero. For `FESystem` objects, the answer is most often "the shape function is zero", but we still have to do a good deal of work. On the other hand, the `FEValues` views do a lot of work caching stuff to make this efficient. It might be useful to think about whether we can appropriate that work here by doing something like
```
  typename Base<dim, spacedim>::value_type
  Scalar<dim, spacedim>::jump(const unsigned int interface_dof_index,
                              const unsigned int q_point) const
  {
    return fe_face_values_here[extractor].value(translated interface dof index, q_point)
             -fe_face_values_there[extractor].value(translated interface dof index, q_point);
  }
```
This clearly takes a bit more thought and needs to copy the style of `FEInterfaceValues::jump()` to do the translation of DoF indices and querying what `FEFaceValues` objects actually exist. But the point is that then we can apply the extractor to the underlying `FEFaceValues` objects, and benefit from the caching of data that happens there.

bangerth:
More to the point, this will also substantially accelerate how the vector cases work (because we can avoid the loop over the components), and it should make it almost trivial to implement the tensor cases as well.

include/deal.II/fe/fe_interface_values.h
```diff
@@ -25,6 +25,342 @@
...
+   */
+  template <int dim, int spacedim = dim>
+  class Scalar : public Base<dim, spacedim>
+  {
+  public:
+    /**
+     * This is the type returned for gradients, for example from
+     * average_gradient().
+     */
+    using gradient_type = dealii::Tensor<1, spacedim>;
```
bangerth, line 70: I think it would actually be nice if we could inherit this information from the place where it really comes from:
```suggestion
    using gradient_type = typename FEValuesViews::Scalar<dim,spacedim>::gradient_type;
```
This way we don't have to duplicate the information.

include/deal.II/fe/fe_interface_values.h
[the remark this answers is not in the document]
tjhei: It stores the ``FEInterfaceViews`` objects. 

> Why do we need fe_interface_views_cache

What else should operator[] return? I would think always returning a new temporary is not as efficient or what are you proposing?

> How are they related?

I don't understand your question, I am sorry.

tjhei:
> More to the point, this will also substantially accelerate how the vector cases work

This makes sense. Maybe we can proceed and do this as a follow-up (that might or might not make it into 9.3)?

tjhei:
Updated.

include/deal.II/fe/fe_interface_values.h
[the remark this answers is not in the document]
bangerth: What @masterleinad is pointing out is that in contrast to `FEValues`, the `FEInterfaceValues` views are really cheap: They only store a pointer and a single integer. There is essentially no cost associated with creating these objects. So there doesn't seem to be much of a case for keeping a cache around for these objects because the cache does not actually cache any pre-computed information.

include/deal.II/fe/fe_interface_values.h
```diff
@@ -1037,6 +1418,324 @@ FEInterfaceValues<dim, spacedim>::jump_3rd_derivative(
...
+
+
+
+namespace internal
+{
+  namespace FEInterfaceViews
+  {
+    template <int dim, int spacedim>
+    Cache<dim, spacedim>::Cache(
+      const FEInterfaceValues<dim, spacedim> &fe_interface)
```
bangerth, line 1717: This function should probably go into the `.cc` file if we want to keep the cache at all.

tjhei: @zjiaqi2018 you will need to do the following:
1. create source/fe/fe_interface_values.cc (it doesn't exist yet, you can steal the structure/header from one of the other files)
2. Move the implementation of Cache::Cache() (where this comment is) into the .cc file
3. add a fe_interface_value.is.inst.in (include in the .cc) and instantiate the Cache for all pairs of dim and spacedim, see fe_dgp.inst.in for an example.
Let me know if you need more help with that.

tjhei: Actually, scratch that and ignore what I wrote. Too much effort as both of them asked to remove the Cache class anyways.

Can you try to:
1. change ``operator[]`` to return the views by value instead of by reference.
2. In the implementation of operator[], create the ``FEInterfaceViews::Scalar<dim, spacedim>`` and return it immediately.
3. Get rid of everything about ``Cache``

zjiaqi2018: @tjhei  Updated, please let me know if something needs to be changed.

tjhei:
Great, thanks.

include/deal.II/fe/fe_interface_values.h
```diff
@@ -427,6 +742,24 @@ class FEInterfaceValues
                       const unsigned int q_point,
                       const unsigned int component = 0) const;
 
+  /**
+   * Create a view of the current FEInterfaceValues object that represents a
+   * particular scalar component of the possibly vector-valued finite element.
+   * The concept of views is explained in the documentation of the namespace
+   * FEInterfaceViews.
```
tjhei, line 749:
```suggestion
   * FEValuesViews
```
We actually don't explain anything here...

include/deal.II/fe/fe_interface_values.h
```diff
@@ -427,6 +742,24 @@ class FEInterfaceValues
...
+   */
+  const FEInterfaceViews::Scalar<dim, spacedim>
+  operator[](const FEValuesExtractors::Scalar &scalar) const;
+
+  /**
+   * Create a view of the current FEInterfaceValues object that represents a set
+   * of <code>dim</code> scalar components (i.e. a vector) of the vector-valued
+   * finite element. The concept of views is explained in the documentation of
+   * the namespace FEInterfaceViews
```
tjhei, line 758: same here

bangerth:
OK, looks good. In a last step, would you mind merging the 5 commits into 1 (or, if you want, one for the code changes, one for the changelog, and one for the tests + output)?

Please leave a note here once you are done so I can see that I should take another look! Thanks for these contributions!

zjiaqi2018:
@bangerth I merged the 5 commits into 1.

bangerth (approved):

bangerth:
Nicely done, @zjiaqi2018 ! Thank you for your contribution, and we hope that there will be many more!

zjiaqi2018:
@bangerth  thanks, more will be coming :)

PR #12048, opened 2021-04-16
Get rid of create_triangulation_compatibility in the GMSH reader.

drwells:
Another part of #8562.


masterleinad (approved):
Looks reasonable.

PR #12050, opened 2021-04-17
Fix test matrix_free/categorize_by_boundary_ids_01

peterrum:
see https://cdash.43-1.org/test/297464

### DIFF
--- a/tests/matrix_free/categorize_by_boundary_ids_01.cc
+++ b/tests/matrix_free/categorize_by_boundary_ids_01.cc
@@ -100,8 +100,6 @@ test(const unsigned int n_refinements)
     return result;
   };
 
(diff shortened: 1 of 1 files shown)


Rombur:
Why is the assert not passing anymore?

peterrum:
> Why is the assert not passing anymore?

If vectorization is disabled, all cells in the macro cell (1 cell) are categorized correctly by definition.

drwells (approved):

drwells:
Would it be better to only add that assertion if the vector width is greater than 1?

peterrum:
> Would it be better to only add that assertion if the vector width is greater than 1?

I must admit it is not really relevant if it was sorted or not before. The requirement is that it is sorted after the function call.

PR #12051, opened 2021-04-18
Fix matrix_free/compute_diagonal tests in release mode

peterrum:
See https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=86.


peterrum:
The failing test is unrelated.

drwells (approved):

PR #12054, opened 2021-04-18
Utilities::MPI::mpi_sum_avg: create MPI_Datatype and MPI_Op once

tamiko:
Instead of repeatedly creating and freeing an MPI_Datatype and MPI_Op for
every invocation of Utilities::MPI::mpi_sum_avg let us define these objects
statically and create them only once.

This works around a memory leak issue with OpenMPI 4.1.0 that will leak 
memory in the subsequent MPI_Allreduce call.

Upstream bug report: https://github.com/open-mpi/ompi/issues/8827

### DIFF
--- a/doc/news/changes/minor/20210419Maier
+++ b/doc/news/changes/minor/20210419Maier
@@ -1,0 +1,8 @@
+Added: The MPI_InitFinalize RAII class has gained an
+MPI_InitFinalize::signals::at_mpi_init and an
+MPI_InitFinalize::signals::at_mpi_finalize signal that are triggered
+immediately after initializing the MPI context with <code>MPI_Init</code>
+and immediately before deinitializing the MPI context with
+<code>MPI_Finalize</code>.
+<br>
+(Matthias Maier, 2021/04/19)
--- a/source/base/mpi.cc
+++ b/source/base/mpi.cc
@@ -922,6 +922,9 @@ namespace Utilities
           // finally set this number of threads
           MultithreadInfo::set_thread_limit(n_threads);
         }
+
+      // As a final step call the at_mpi_init() signal handler.
+      signals.at_mpi_init();
     }
 
 
@@ -954,6 +957,9 @@ namespace Utilities
 
     MPI_InitFinalize::~MPI_InitFinalize()
     {
+      // First, call the at_mpi_finalize() signal handler.
+      signals.at_mpi_finalize();
+
       // make memory pool release all PETSc/Trilinos/MPI-based vectors that
       // are no longer used at this point. this is relevant because the static
       // object destructors run for these vectors at the end of the program
--- a/source/base/mpi.cc
+++ b/source/base/mpi.cc
@@ -786,6 +786,9 @@ namespace Utilities
 
 #endif
 
+    /* Force initialization of static struct: */
+    MPI_InitFinalize::Signals MPI_InitFinalize::signals =
+      MPI_InitFinalize::Signals();
 
 
     MPI_InitFinalize::MPI_InitFinalize(int &              argc,
--- a/source/base/mpi.cc
+++ b/source/base/mpi.cc
@@ -623,19 +623,10 @@ namespace Utilities
         }
 
       /*
-       * Create a custom MPI datatype handle describing the memory layout
-       * of the MinMaxAvg struct and a custom MPI op handle for our
-       * max_reduce function.
-       *
-       * This setup will leak some minor allocated memory (in the range of
-       * a few bytes) at the end of the program run because we never call
-       * MPI_Type_free(&type) and MPI_Op_free(&op) to explicitly free up
-       * memory. This is a cosmetic issue but some address sanitizers will
-       * complain about this (such as clang's -fsanitize=address).
+       * A custom MPI datatype handle describing the memory layout of the
+       * MinMaxAvg struct. Initialized on first pass control reaches the
+       * static variable. So hopefully not initialized too early.
        */
-
-      /* Initialized on first pass control reaches the static variable. So
-       * hopefully not initialized too early. */
       static MPI_Datatype type = []() {
         MPI_Datatype type;
 
@@ -654,11 +645,21 @@ namespace Utilities
         ierr = MPI_Type_commit(&type);
         AssertThrowMPI(ierr);
 
+        /* Ensure that we free the allocated datatype again at the end of
+         * the program run just before we call MPI_Finalize():*/
+        MPI_InitFinalize::signals.at_mpi_finalize.connect([type]() mutable {
+          int ierr = MPI_Type_free(&type);
+          AssertThrowMPI(ierr);
+        });
+
         return type;
       }();
 
-      /* Initialized on first pass control reaches the static variable. So
-       * hopefully not initialized too early. */
+      /*
+       * A custom MPI op handle for our max_reduce function.
+       * Initialized on first pass control reaches the static variable. So
+       * hopefully not initialized too early.
+       */
       static MPI_Op op = []() {
         MPI_Op op;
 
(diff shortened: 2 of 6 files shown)


tamiko:
@tjhei *ping*

tjhei (commented):
I am okay with merging, but consider my two suggestions.

source/base/mpi.cc
```diff
@@ -605,6 +605,45 @@ namespace Utilities
                 const ArrayView<MinMaxAvg> &   result,
                 const MPI_Comm &               mpi_communicator)
     {
+      /*
+       * Create a custom MPI datatype handle describing the memory layout
+       * of the MinMaxAvg struct and a custom MPI op handle for our
+       * max_reduce function.
+       *
+       * This setup will leak some minor allocated memory (in the range of
```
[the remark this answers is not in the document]
tamiko: What about we refactor all of this into its own `static struct` in the cc file?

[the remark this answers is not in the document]
tamiko: Actually. I like this approach. Let me refactor.

source/base/mpi.cc
```diff
@@ -605,6 +605,45 @@ namespace Utilities
...
+          MPI_Datatype types[] = {MPI_DOUBLE, MPI_INT, MPI_DOUBLE};
+
+          ierr =
+            MPI_Type_create_struct(3, lengths, displacements, types, &type);
+          AssertThrowMPI(ierr);
+
+          ierr = MPI_Type_commit(&type);
+          AssertThrowMPI(ierr);
+        });
+
```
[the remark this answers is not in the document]
tamiko: Refactored.

source/base/mpi.cc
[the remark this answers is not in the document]
tamiko: @tjhei I know that this is not elegant. The entire MPI context however is already leaking like crazy.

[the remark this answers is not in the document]
bangerth: You can always to a `static std::unique_ptr<...>` local variable. That is initialized on first run-through of the function, and deallocated at the end of the program.

source/base/mpi.cc
```diff
@@ -622,13 +622,49 @@ namespace Utilities
...
+       * max_reduce function.
+       *
+       * This setup will leak some minor allocated memory (in the range of
+       * a few bytes) at the end of the program run because we never call
+       * MPI_Type_free(&type) and MPI_Op_free(&op) to explicitly free up
+       * memory. This is a cosmetic issue but some address sanitizers will
+       * complain about this (such as clang's -fsanitize=address).
+       */
+      static MPI_Datatype type;
+      static MPI_Op       op;
```
bangerth, line 637: You can do this easier than using `std::call_once` if you do 
```suggestion
      static MPI_Op       op = [=]() {...} ();
```
In other words, you just provide an initializer for the static variable.

source/base/mpi.cc
```diff
@@ -622,13 +622,49 @@ namespace Utilities
...
+
+      std::once_flag initialization_flag;
```
bangerth, line 639: This isn't going to work unless you mark the variable as `static`. But I think there is a better option (see above).

bangerth (changes requested):

source/base/mpi.cc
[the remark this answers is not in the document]
tamiko: @bangerth I had that variant before and reverted to this first version. The problem with this is that the deallocation happens after `dealii::MPI_InitFinalize` closed the MPI context and the program will abort with an error.

source/base/mpi.cc
tamiko: I thought about this as well.

I opted for the `std::call_once` variant because I think it is mildly cleaner. (We cannot call to `MPI_Type_create_struct`, `MPI_Type_commit` and `MPI_Op_create` outside of an `MPI_InitFinalize` so all of the static initializer would need a separate `if(job_supports_mpi())`.)

source/base/mpi.cc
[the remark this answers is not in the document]
masterleinad: I would also prefer to use a static variable in `MPI_InitFinalize` (or a callback) that can handle the cleanup in the destructor of `MPI_InitFinalize`.

source/base/mpi.cc
bangerth: No, that's what local `static` variables do: They are initialized upon first go-through of the function. They do exactly what you are doing with the `call_once`.

source/base/mpi.cc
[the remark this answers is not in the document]
bangerth: I see. What we need is a `static` member variable of type `boost::signal` in the `MPI_InitFinalize` class, say called `at_mpi_exit` that one can attach to. That's where you would put the call that de-allocates the memory.

source/base/mpi.cc
tamiko: I will revise.

tamiko: `std::call_once` is now gone.

bangerth (approved):
Looks good to me. We *could* deal with the memory leak, but I'm not actually too concerned about it. My major worry is that having these kinds of leaks makes it more difficult to run tools such as `memcheck` on deal.II-based codes. But let's do this one step at a time.

source/base/mpi.cc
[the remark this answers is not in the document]
tamiko: Implementing.

tamiko:
@bangerth I am now halfway through implementing the signal handlers. Let me add that to the pull request. :smile:

tamiko:
@bangerth @masterleinad I will finish up the singal handler this evening. (Tests and hooking up the deallocator are missing).

source/base/mpi.cc
tjhei: > They are initialized upon first go-through of the function.

Yes, and the compiler even generates code to make it thread-safe.

doc/news/changes/minor/20210419Maier
```diff
@@ -0,0 +1,8 @@
+Added: The MPI_InitFinalize RAII class has gained a
+MPI_InitFinalize::singals::at_mpi_ini and an
```
kronbichler, line 2:
```suggestion
MPI_InitFinalize::singals::at_mpi_init and an
```

kronbichler (approved):

doc/news/changes/minor/20210419Maier
jppelteret: singals -> signals

doc/news/changes/minor/20210419Maier
```diff
@@ -0,0 +1,8 @@
+Added: The MPI_InitFinalize RAII class has gained a
+MPI_InitFinalize::singnals::at_mpi_init and an
```
masterleinad, line 2:
```suggestion
MPI_InitFinalize::signals::at_mpi_init and an
```

masterleinad (commented):
Apart from the typo, this looks good to me.

doc/news/changes/minor/20210419Maier
tamiko: Fixed!

tamiko:
@dealii/dealii Alright. I have added a small test for the signal handler and I think I have addressed all typos. Let's see whether the testsuite likes it.

doc/news/changes/minor/20210419Maier
```diff
@@ -0,0 +1,8 @@
+Added: The MPI_InitFinalize RAII class has gained an
+MPI_InitFinalize::signals::at_mpi_init and an
+MPI_InitFinalize::signals::at_mpi_finalize signal. That are triggered
```
masterleinad, line 3:
```suggestion
MPI_InitFinalize::signals::at_mpi_finalize signal that are triggered
```
nitpick: not a sentence. :slightly_smiling_face:

tests/base/mpi_init_finalize_signals_01.cc
```diff
@@ -0,0 +1,42 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2017 - 2018 by the deal.II authors
```
masterleinad, line 3:
```suggestion
// Copyright (C) 2021 by the deal.II authors
```

masterleinad (approved):

tjhei (approved):

bangerth (approved):

bangerth:
Thanks, @tamiko , for sticking with everyone's escalating requests!

PR #12057, opened 2021-04-19
Provide a move constructor for InterpolatedTensorProductGridData.

bangerth:
Related to #12008.

/rebuild

### DIFF
--- a/tests/base/functions_10_move.cc
+++ b/tests/base/functions_10_move.cc
@@ -1,0 +1,149 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2013 - 2018 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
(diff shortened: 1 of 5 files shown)


tests/base/functions_10_move.cc
```diff
@@ -0,0 +1,149 @@
...
+  Table<dim, double> data = fill(coordinates);
+
+  Functions::InterpolatedTensorProductGridData<dim> f(std::move(coordinates),
+                                                      std::move(data));
+
+  // Verify tha the data has really been moved
+  for (unsigned int d = 0; d < dim; ++d)
+    Assert(data.size()[d] == 0, ExcInternalError());
+  for (unsigned int d = 0; d < dim; ++d)
+    Assert(coordinates[d].size() == 0, ExcInternalError());
```
masterleinad, line 86: Hmmm. Can we really rely on accessing the moved-from variables to yield something predictable?

bangerth: We certainly can for this class, where we know what happens. I might not recommend this in user code, but in the test suite, I think this ok.

The standard says in general that moved-from objects are in an undefined state, but in practice, this state must be well-enough defined that one can run the destructor on these objects. In the current case, I cannot see that a `std::vector` that has been moved from could reasonably yield anything other than zero for `size()`.

bangerth:
Ping?

masterleinad (approved):

PR #12062, opened 2021-04-20
Add function estimating Function bounds over a BoundingBox

simonsticko:
Add a function taylor_estimate_function_bounds, which estimates the
range of a Function over a BoundingBox using a Taylor expansion. Both
the range of the function value and each gradient component is
estimated.

Add another function, lower_bound_on_abs, which converts the lower and
upper bounds to a lower bound on the absolute value of the function,
C \leq |f|.

### DIFF
--- a/doc/news/changes/minor/20210425SimonSticko
+++ b/doc/news/changes/minor/20210425SimonSticko
@@ -1,0 +1,6 @@
+New: Add a function taylor_estimate_function_bounds, which estimates the range
+of the value and gradient components of a Function<dim> over a
+BoundingBox<dim>, by approximating the function by a 2nd order Taylor
+polynomial.
+<br>
+(Simon Sticko, 2021/04/25)
--- a/include/deal.II/base/function_tools.h
+++ b/include/deal.II/base/function_tools.h
@@ -1,0 +1,71 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (C) 2010 - 2019 by the deal.II authors
+//
+// This file is part of the deal.II library.
+//
+// The deal.II library is free software; you can use it, redistribute
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+// ---------------------------------------------------------------------
+
+#ifndef dealii_function_tools_h
+#define dealii_function_tools_h
+
+
+#include <deal.II/base/config.h>
+
+#include <deal.II/base/bounding_box.h>
+#include <deal.II/base/function.h>
+
+DEAL_II_NAMESPACE_OPEN
+
+namespace FunctionTools
+{
+  /**
+   * Estimate bounds on the value and bounds on each gradient component of a
+   * Function, $f$, over a BoundingBox, by approximating it by a 2nd order
+   * Taylor polynomial starting from the box center.
+   *
+   * Each lower and upper bound is returned as a
+   * <code>std::pair<double, double></code>, such that the first entry is the
+   * lower bound, $L$, and the second is the upper bound, $U$, i.e.
+   * $f(x) \in [L, U]$.
+   *
+   * The function value, gradient, and Hessian are computed at the box center.
+   * The bounds on the value of the function are then estimated as
+   *
+   * $f(x) \in [f(x_c) - F, f(x_c) + F]$,
+   * where
+   * $F = \sum_i |\partial_i f(x_c)| h_i
+   *    + 1/2 \sum_i \sum_j |\partial_i \partial_j f(x_c)| h_i h_j$.
+   *
+   * Here, $h_i$ is half the side length of the box in the $i$th coordinate
+   * direction, which is the distance we extrapolate. The bounds on the gradient
+   * components are estimated similarly as
+   *
+   * $\partial_i f \in [\partial_i f(x_c) - G_i, \partial_i f(x_c) + G_i]$,
+   * where
+   * $G_i = \sum_j |\partial_i \partial_j f(x_c)| h_j$.
+   *
+   * If the function has more than 1 component the @p component parameter can
+   * be used to specify which function component the bounds should be computed
+   * for.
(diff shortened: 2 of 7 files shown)


simonsticko:
Need this as part of implementing something else. I'm not really sure if this is too specific to be in deal.II/base or not.

Rombur (commented):
I have a few minor comments. This needs a changelog. TBH I don't really like the name of the functions but I don't have anything better.

include/deal.II/base/function_bounds.h
```diff
@@ -0,0 +1,112 @@
...
+
+DEAL_II_NAMESPACE_OPEN
+
+/**
+ * Struct storing lower and upper bounds, $[L, U]$, on the value of a
+ * function, $f(x)$, such that
+ *
+ * $L \leq f(x) \leq U$.
+ */
+struct FunctionBounds
```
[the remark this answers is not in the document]
simonsticko: Not really. I just think `max`/`min` are much better names than `first`/`second`.
Would you prefer `std::pair`?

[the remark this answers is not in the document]
Rombur: I don't have a strong preference. I am just wondering what other people think.

simonsticko:
Fixed the comments, except for the discussion on `std::pair` vs using a custom struct.
Waiting on more feedback before I change it.

include/deal.II/base/function_bounds.h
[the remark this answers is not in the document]
bangerth: I think a `std::pair` is good enough.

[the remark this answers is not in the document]
bangerth: In C++17, you will be able to write
```
  double [min,max] = taylor_estimate_function_bounds(...);
```
to decompose a pair right away.

include/deal.II/base/function_bounds.h
```diff
@@ -0,0 +1,112 @@
...
+   *
+   * $f(x) \leq U$.
+   */
+  double max;
+};
+
+
+/**
+ * Estimate bounds on the value and each gradient component of a Function, $f$,
+ * over a BoundingBox, by approximating it as a 2nd order Taylor polynomial
```
bangerth, line 60:
```suggestion
 * over a BoundingBox, by approximating it by a 2nd order Taylor polynomial
```

include/deal.II/base/function_bounds.h
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * The function value, gradient and Hessian are computed at the box center.
+ * The upper and lower bounds on the functions value are then estimated as
+ *
+ * $f(x) \in [f(x_c) - F, f(x_c) + F]$,
+ * where
+ * $F = \sum_i |\partial_i f(x_c)| h_i
+ *    + 1/2 \sum_i \sum_j |\partial_i \partial_j f(x_c)| h_i h_j$.
+ *
+ * Here, $h_i$ is half the side length of the box, which is the distance we
```
bangerth, line 71:
```suggestion
 * Here, $h_i$ is half the side length of the box in the $i$th coordinate direction, which is the distance we
```

include/deal.II/base/function_bounds.h
```diff
@@ -0,0 +1,112 @@
...
+ * where
+ * $G_i = \sum_j |\partial_i \partial_j f(x_c)| h_j$.
+ *
+ * If the function has more than 1 component the @p component parameter can
+ * be used to specify which function component the bounds should be computed
+ * for.
+ */
+template <int dim>
+void
+taylor_estimate_function_bounds(
```
bangerth, line 84: would it make sense to put this function into `namespace Functions` or a new namespace `FunctionTools`?

include/deal.II/base/function_bounds.h
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Return a lower bound, $L_a$, on the absolute value of a function, $f(x)$:
+ *
+ * $L_a \leq |f(x)|$,
+ *
+ * by estimating it from the incoming lower and upper bounds:
+ * $L \leq f(x) \leq U$.
+ *
+ * By rewriting the above as
```
bangerth, line 100: what does "the above" refer to? The documentation of the other function? I think you probably want to be explicit here.

include/deal.II/base/function_bounds.h
```diff
@@ -0,0 +1,112 @@
...
+ * $F - C \leq f(x) \leq F + C \LeftRightArrow |f(x) - F| \leq C$,
```
bangerth, line 101: It is not clear to me how this function can determine F and C from just its argument. Can you make that connection?

include/deal.II/base/function_bounds.h
```diff
@@ -0,0 +1,112 @@
...
+ * with $F = (U + L)/2$, $C = (U - L)/2$, and using the inverse triangle
+ * inequality, $|F| - |f(x)| \leq |f(x) - F|$, we get that $L_a = |F| - C$.
+ * If this is negative, $L_a$ is set to 0, since the absolute value can't be
+ * negative.
+ */
+double
+lower_bound_on_abs(const FunctionBounds &function_bounds);
```
bangerth, line 108: On second thought about this function, I think it is too specialized. I can see the point of the function above, but would suggest that you keep this one where you actually need it. I think that would also address the issues with the documentation: The function needs substantially more space for documentation that for the implementation because one has to explain the context, and if you just inlined the computations this function does into the place where you currently use it, the context is easier to understand for a reader.

include/deal.II/base/function_bounds.h
simonsticko: If `FunctionTools` is better, do you think the header should be `function_tools.h` instead of `function_bounds.h`?

bangerth: Yes, probably. We may want to add more functions to the namespace.

simonsticko:
Alright:
 - Changed so that the bounds are returned as a `std::pair<double, double>`
 - Moved the function to a new namespace `FunctionTools`
 - Renamed the file to `function_tools.h`
 - Removed the function `lower_bound_on_abs`.

bangerth (approved):

PR #12067, opened 2021-04-20
Deprecate TriaAccessor::number_of_children.

### LINKED ISSUE #10205: Rename TriaAccessor::number_of_children
The `TriaAccessor` class has a (rarely used) function `number_of_children()`. But it also has `n_children()`. The latter returns the number of immediate children, whereas the former returns the total number of descendants. The similarity in naming is certainly confusing.

We should return `number_of_children()` to `n_descendants()` to make it clearer that it is a recursive relationship.

bangerth:
Use TriaAccessor::n_active_descendants() instead. Fixes #10205.

/rebuild


masterleinad (approved):
Looks good to me!

Rombur (approved):

PR #12068, opened 2021-04-20
Improve the scalar assignment exception message.

drwells:
This is from a question I got from a student today in class - we should use the exception in `Tensor` and be a bit more explicit about why this operation is not allowed.


jppelteret (approved):
Much more expansive explanation. Thanks!

masterleinad (approved):

bangerth (approved):
+1

PR #12074, opened 2021-04-22
Bugfix: `hp::Refinement::predict_error` - cast to int

marcfehling:
Error prediction fails for p-coarsening. Subtracting two `unsigned ints` is dangerous.

Added and adjusted tests to verify p-coarsening.

### DIFF
--- a/source/hp/refinement.cc
+++ b/source/hp/refinement.cc
@@ -634,7 +634,8 @@ namespace hp
             if (cell->future_fe_index_set())
               {
                 predicted_errors[cell->active_cell_index()] *=
-                  std::pow(gamma_p, future_fe_degree - cell->get_fe().degree);
+                  std::pow(gamma_p,
+                           int(future_fe_degree) - int(cell->get_fe().degree));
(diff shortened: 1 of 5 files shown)


source/hp/refinement.cc
```diff
@@ -634,7 +634,9 @@ namespace hp
             if (cell->future_fe_index_set())
               {
                 predicted_errors[cell->active_cell_index()] *=
-                  std::pow(gamma_p, future_fe_degree - cell->get_fe().degree);
+                  std::pow(gamma_p,
+                           static_cast<int>(future_fe_degree -
+                                            cell->get_fe().degree));
```
drwells, line 639:
```suggestion
                           int(future_fe_degree) -
                           int(cell->get_fe().degree));
```
casting to an `int` after the fact relies on some implementation-defined behavior - casting before subtracting is unambiguous.

drwells (changes requested):

drwells (approved):

PR #12080, opened 2021-04-22
[C++17] Check whether the output directory exists.

### LINKED ISSUE #10281: [C++17] step-70: Create output directory if necessary
See this comment: 
https://github.com/dealii/dealii/pull/10274/files#diff-5e19932b9e815b576e4bb4f4bacac9ca

There may be other steps to which this applies as well.

bangerth:
Fixes #10281 . We can only apply this once we support C++17, but I had this ready so I'll post for whenever the time comes.

### DIFF
--- a/examples/step-70/step-70.cc
+++ b/examples/step-70/step-70.cc
@@ -144,6 +144,7 @@ namespace LA
 #endif
 
 #include <cmath>
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <memory>
@@ -769,8 +770,11 @@ namespace Step70
 
   // In the constructor, we create the mpi_communicator as well as
   // the triangulations and dof_handler for both the fluid and the solid.
-  // Using the mpi_communicator, both the ConditionalOStream and TimerOutput
+  // Using the `mpi_communicator`, both the ConditionalOStream and TimerOutput
   // object are constructed.
+  //
+  // In the constructor, we also check whether the output directory
+  // specified in the input file exists and, if not, create it.
   template <int dim, int spacedim>
   StokesImmersedProblem<dim, spacedim>::StokesImmersedProblem(
     const StokesImmersedProblemParameters<dim, spacedim> &par)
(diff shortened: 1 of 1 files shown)


examples/step-70/step-70.cc
```diff
@@ -144,6 +144,7 @@ namespace LA
 #endif
 
 #include <cmath>
+#include <filesystem>
```
tamiko, line 147: That'll be some fun - For `libstdc++` we will have to figure out a sane way to decide whether we have to link against `libstdc++fs` as well or not.

bangerth: Shouldn't `-std=c++17` imply that?

tamiko: @bangerth gcc version 9 onwards should just require `-std=c++17`, gcc version 8 and older require manual linking with `libstdc++fs`. At least if I recall correctly :smile_cat:

marcfehling: I remember that I had similar issues with gcc8. I solved this by substituting them with `boost` libraries:
```
#ifdef __cpp_lib_filesystem
#  include <filesystem>
namespace filesystem
{
  using namespace std::filesystem;
}
#else
#  include <boost/filesystem.hpp>
namespace filesystem
{
  using namespace boost::filesystem;
}
#endif
```
I'm not sure if the `__cpp_lib_filesystem` flag only exists for `gcc`.


We could move this code snippet into the `std_cxx17` namespace. What do you think?

bangerth: The problem is that we don't include the boost filesystem parts in the `bundled/` directory :-( There is also the issue that this is a tutorial and I really didn't want to introduce some complicated workaround in tutorials -- I'd rather just leave things as they are if we can't do it right....

bangerth:
This is now open for merge :-)

tamiko (approved):

tamiko:
The minimal version we test and thus meaningfully support now is gcc-9 so this can go in as is.

kronbichler (approved):

masterleinad:
> The minimal version we test and thus meaningfully support now is gcc-9 so this can go in as is.

tester-ng still uses gcc-7.5, though, see https://cdash.dealii.org/viewSite.php?siteid=3&project=1&currenttime=1688518800.

PR #12082, opened 2021-04-22
Move GridReordering::reorder_cells().

### LINKED ISSUE #12058: Remove all public-facing references to the old-style ordering
Followup to #8562.

Backstory, for those that are new: in 5.2 (circa commit ca020c3a56696bee27f8da93f5992381ec210213) we [1] changed the numbering scheme from UCD to lexical, but there are still a few places in the library that expect the old numbering. The only place we now use the old-style reordering is inside the `GridReordering` class, which needs to be deprecated and have its functions put in `GridTools`.

Now is the time to finally upgrade past these functions since we need to move to support mixed meshes soon, and handling mixed mesh and compatibility with the old numbering is just too much.

[1] not really 'we', this commit came about 10 years before I started working on deal.II :)

drwells:
Fixes #12058. This was easier than I thought it would be - the second `GridReordering` function already used the new-style numbering internally.


bangerth (approved):
I assume that the code is simply copied without actual changes?

drwells:
Yup!

bangerth:
Hm, the testers fail:
```
/home/runner/work/dealii/dealii/include/deal.II/grid/grid_reordering.h:37:32: error: attribute ignored in declaration of ‘class dealii::GridReordering<dim, spacedim>’ [-Werror=attributes]
 DEAL_II_DEPRECATED_EARLY class GridReordering
                                ^~~~~~~~~~~~~~
/home/runner/work/dealii/dealii/include/deal.II/grid/grid_reordering.h:37:32: note: attribute for ‘class dealii::GridReordering<dim, spacedim>’ must follow the ‘class’ keyword
```
So it's a placement thing for the attribute.

drwells:
I can never remember where these things go - it should be fixed now.

PR #12083, opened 2021-04-22
Simplify some type computations in FEValuesViews.

bangerth:
Specifically, instead of using a nested class, just use a template type alias. While there, also fix a few places where we should have used the existing type alias but instead had repeated the same type computation used in the definition of the type alias.

This patch is pretty repetitive. It's unlikely going to be fun to look over, but at least should be very well covered by the test suite. I'll try not to write any such patches any more until the release and focus on the things that *actually* need to be done.

/rebuild

### DIFF
--- a/doc/news/changes/minor/20210425Bangerth
+++ b/doc/news/changes/minor/20210425Bangerth
@@ -1,0 +1,10 @@
+Deprecated: The FEValuesViews::Scalar, FEValuesViews::Vector,
+FEValuesViews::Tensor, and FEValuesViews::SymmetricTensor classes all
+have `OutputType` structures that defined the types of evaluating
+the values and derivatives of finite element fields at quadrature
+points. These structures have been deprecated, and the corresponding
+types are now defined via local `using` declarations in the classes
+mentioned above.
+<br>
+(Wolfgang Bangerth, 2021/04/25)
+
--- a/include/deal.II/fe/fe_values.h
+++ b/include/deal.II/fe/fe_values.h
@@ -224,6 +224,56 @@ namespace FEValuesViews
       typename ProductType<Number, third_derivative_type>::type;
 
     /**
+     * A struct that provides the output type for the product of the value
+     * and derivatives of basis functions of the Scalar view and any @p Number type.
+     *
+     * @deprecated Use the types defined in the surrounding class instead.
+     */
+    template <typename Number>
+    struct DEAL_II_DEPRECATED_EARLY OutputType
+    {
+      /**
+       * An alias for the data type of the product of a @p Number and the
+       * values of the view the Scalar class.
+       */
+      using value_type =
+        typename ProductType<Number,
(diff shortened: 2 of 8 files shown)


bangerth:
I realized all of this when looking at what #12047 does.

drwells (approved):

jppelteret:
This is an incompatible change (at least, I use these types in an external project) so I think that it should be documented as such.

drwells:
That's a good point. Do we need to keep `OutputType` et al around to keep that code working?

jppelteret:
@bangerth Do you perhaps need to change this https://github.com/dealii/dealii/blob/master/include/deal.II/meshworker/scratch_data.h#L45-L47 as well? I also see more instances of `OutputType` in `fe_values.inst.in` and these tests:
- tests/fe/fe_values_views_types.cc
- tests/fe/fe_values_view_get_function_from_local_dof_values_01.cc
- tests/fe/fe_values_view_get_function_from_local_dof_values_01.cc

jppelteret:
> That's a good point. Do we need to keep OutputType et al around to keep that code working?

Its not a public project, so I don't mind fixing the breakage. But I think that it might be worth deprecating this, since I think that it's one of the only reliable mechanisms that we have for deducing an operation result type for a specified scalar type.

masterleinad:
There are some failing tests
```
1674 - fe/fe_values_view_get_function_from_local_dof_values_01.debug (Failed)
1860 - fe/fe_values_views_types.debug (Failed)
3119 - meshworker/scratch_data_09d.debug (Failed)
3125 - meshworker/scratch_data_09b.debug (Failed)
3126 - meshworker/scratch_data_09c.debug (Failed)
3129 - meshworker/scratch_data_04.debug (Failed)
3132 - meshworker/scratch_data_09a.debug (Failed)
```

bangerth:
Huh, who knew that that internal type calculation leaks out of these classes. That certainly was never the intention.

I can fix that tomorrow, and if you all insist, I can even keep the old `OutputType` classes. The backward compatible way -- which I would prefer -- is to document that one can get at that type using `decltype(fe_values[pressure].get_function_values(...))`. Would that be acceptable?

jppelteret:
> if you all insist, I can even keep the old OutputType classes. 

Not necessary, from my point of view. Thanks for offering though!

> The backward compatible way -- which I would prefer -- is to document that one can get at that type using decltype(fe_values[pressure].get_function_values(...)). Would that be acceptable?

I'd be happy with us doing this.

jppelteret (changes requested):
Hi @bangerth, I'm just going to temporarily block this PR from being merged so that we don't forget to document the incompatibility that it introduces.

bangerth:
I brought the types back and deprecated them. I also added a changelog entry.

jppelteret (approved):

jppelteret:
> I brought the types back and deprecated them. I also added a changelog entry.

Thanks a lot for doing that!

PR #12084, opened 2021-04-22
Guard SUNDIALS variables more carefully

masterleinad:
Looking at https://github.com/LLNL/sundials/blame/73c280cd55ca2b42019c8a9aa54af10e41e27b9d/src/sundials/sundials_nvector.c, we need to introduce some more guards.
Should fix https://cdash.43-1.org/test/1218759.


bangerth (approved):
Thanks for figuring this out. They've been making a lot of incompatible changes :-(

PR #12085, opened 2021-04-22
Fix compilation with gcc-5.

drwells:
The compiler generates identical symbols for each static declaration, which leads to a problem in the assembler. Work around it by not declaring the variable as static.

### DIFF
--- a/source/hp/refinement.cc
+++ b/source/hp/refinement.cc
@@ -828,8 +828,7 @@ namespace hp
       // possible
       using level_type =
         typename dealii::DoFHandler<dim, spacedim>::active_fe_index_type;
-      static const level_type invalid_level =
-        dealii::DoFHandler<dim, spacedim>::invalid_active_fe_index;
+      const auto invalid_level = static_cast<level_type>(-1);
 
       // map from FE index to level in hierarchy
       // FE indices that are not covered in the hierarchy are not in the map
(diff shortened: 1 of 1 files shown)


tjhei (approved):

masterleinad (approved):

Rombur:
Windows doesn't like this change.
```
D:\a\dealii\dealii\source\hp\refinement.cc(903): error C3493: 'invalid_level' cannot be implicitly captured because no default capture mode has been specified
```

source/hp/refinement.cc
```diff
@@ -828,7 +828,7 @@ namespace hp
       // possible
       using level_type =
         typename dealii::DoFHandler<dim, spacedim>::active_fe_index_type;
-      static const level_type invalid_level =
+      constexpr level_type invalid_level =
```
masterleinad, line 831:
```suggestion
      const level_type invalid_level =
```
I don't see a good reason why this should be `constexpr` or `static` anyway.

bangerth: True. Just make it a regular `const` variable.

But there is a semantic issue here. Why would we assign a variable with name `invalid_active_fe_index` to a variable called `invalid_level` and type `level_type`? These two things are completely different. I *think* what is called for is something like `static_cast<level_type>(-1)` instead.

marcfehling: Oh, I'm responsible for this. `static_cast<level_type>(-1)` would indeed be a better choice here.

Since `level_type` is essentially the same as `active_fe_index_type`, the result will be the same. It is more clear with the `static_cast` though.

drwells: Fixed!

bangerth (approved):

PR #12087, opened 2021-04-23
Avoid warning about out-of-bound access in Tensor

masterleinad:
This should fix https://cdash.43-1.org/build/287. I'm not quite sure if this is too costly or not, though.

### DIFF
--- a/include/deal.II/base/tensor.h
+++ b/include/deal.II/base/tensor.h
@@ -1293,7 +1293,7 @@ constexpr DEAL_II_ALWAYS_INLINE
   AssertIndexRange(i, dim);
 #  endif
 
-  return values[i];
+  return values[i < dim ? i : 0];
 }
 
 


include/deal.II/base/tensor.h
```diff
@@ -1293,7 +1293,7 @@ constexpr DEAL_II_ALWAYS_INLINE
   AssertIndexRange(i, dim);
 #  endif
 
-  return values[i];
+  return values[i < dim ? i : 0];
```
bangerth, line 1296: This seems like the wrong approach. For reference, the warning is this:
```
/.../dealii/include/deal.II/base/tensor.h:1296:16: warning: array subscript 2 is above array bounds of ‘const dealii::Tensor<0, 2, double> [2]’ [-Warray-bounds]
```
So we have a rank-0 tensor (a scalar), for which the only valid index is `i=0`. As a matter of fact, I don't understand why for rank=0 we use this operator to begin with -- we have a specialization `Tensor<0,dim,Number>` for this.

bangerth (commented):
I think this is the wrong approach :-( We need to dig a bit deeper as to why we get here at all.

PR #12088, opened 2021-04-23
Fix a static_assert to actually make sense.

bangerth:
More specifically, there is a specialization Tensor<0,dim,Number> so we should reallly never get here with rank==0.

There is of course another funny-looking static assertion in the next line that one might investigate (will do next) but that is a separate matter.

/rebuild


drwells (approved):

bangerth:
We can't change the second `static_assert` without breaking a *lot* of things. This must wait for a later time.

PR #12090, opened 2021-04-24
fix all-headers/sundials/sundials_backport.h test

tamiko:
This header file is only included for sundials prior to version 5.0.0.
Nevertheless, we also have to add a version guard in the header itself
so that the header include test doesn't fail.

### DIFF
--- a/include/deal.II/sundials/sundials_backport.h
+++ b/include/deal.II/sundials/sundials_backport.h
@@ -29,8 +29,9 @@
 #include <deal.II/base/config.h>
 
 #ifdef DEAL_II_WITH_SUNDIALS
+#  if DEAL_II_SUNDIALS_VERSION_LT(5, 0, 0)
 
-#  include <sundials/sundials_nvector.h>
+#    include <sundials/sundials_nvector.h>
 
 DEAL_II_NAMESPACE_OPEN
 namespace SUNDIALS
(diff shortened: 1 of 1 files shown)


include/deal.II/sundials/sundials_backport.h
```diff
@@ -29,8 +29,9 @@
 #include <deal.II/base/config.h>
 
 #ifdef DEAL_II_WITH_SUNDIALS
+#  if DEAL_II_SUNDIALS_VERSION_LT(5, 0, 0)
```
tamiko, line 32: The only time `sundials_backport.h` is included in any deal.II header is guarded with `DEAL_II_SUNDIALS_VERSION_LT(5, 0, 0)`.

include/deal.II/sundials/sundials_backport.h
```diff
@@ -53,10 +54,10 @@ namespace SUNDIALS
       ops->nvspace           = nullptr;
       ops->nvgetarraypointer = nullptr;
       ops->nvsetarraypointer = nullptr;
-#  if DEAL_II_SUNDIALS_VERSION_GTE(5, 0, 0)
+#    if DEAL_II_SUNDIALS_VERSION_GTE(5, 0, 0)
```
tamiko, line 57: I have deliberately kept all of these parts in place even though these preprocessor conditionals can never be true.

masterleinad:
Looks like you need to rebase.

tamiko:
@masterleinad I think 32bf924707dfc6c6e11ac0a25da42a26f967ba3d probably already fixes the issue.

tamiko:
This PR is probably obsolete, closing :slightly_smiling_face:

PR #12095, opened 2021-04-26
Fix mpi/extract_boundary_dofs

masterleinad:
Should fix https://cdash.43-1.org/test/2073991. We are actually modifying `boundary_dofs` below.


bangerth (approved):
Oops :-(

PR #12096, opened 2021-04-26
Fix base/unsubscribe_subscriptor for clang

masterleinad:
Should fix https://cdash.43-1.org/test/2044524. I remember us breaking lines error messages as well in some patch and it looks like this is one of the tests we forgot to update.


drwells (approved):

bangerth:
That was #11624.

PR #12099, opened 2021-04-26
Fix the number of entries in the FEValues view cache.

bangerth:
Looking at the implementation of the FEValuesViews::Vector and similar classes, we see that they return 'spacedim' objects at a time, and similar for the tensor views. The number of entries in the cache for vectors was already keyed off 'spacedim', but for the tensors it was based on 'dim' instead. This isn't quite right.

I thought about how I could write a test, but what it boils down is that in the past, we simply had too many elements in these arrays whereas now I restrict it to as many as there really should be. There is no test that previously failed and now succeeds :-)

/rebuild


bangerth:
Found while looking at #12047.

masterleinad (approved):
Looks good to me.

PR #12100, opened 2021-04-27
Fix up tests for ICC 2018

masterleinad:
Related to #12098 and #12096, fixes tests in https://cdash.43-1.org/viewTest.php?onlyfailed&buildid=413. I also needed to provide a user-provided destructor for `DistortedCellList` to avoid linker errors.

### DIFF
--- a/include/deal.II/grid/tria.h
+++ b/include/deal.II/grid/tria.h
@@ -1532,12 +1532,7 @@ public:
      * automatically generated destructor would have a different one due to
      * member objects.
      */
-#ifdef __INTEL_COMPILER
-    virtual ~DistortedCellList() noexcept override
-    {}
-#else
-    virtual ~DistortedCellList() noexcept override = default;
-#endif
+    virtual ~DistortedCellList() noexcept override;
 
     /**
      * A list of those cells among the coarse mesh cells that are deformed or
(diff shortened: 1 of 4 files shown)


include/deal.II/grid/tria.h
```diff
@@ -1532,7 +1532,11 @@ class Triangulation : public Subscriptor
      * automatically generated destructor would have a different one due to
      * member objects.
      */
+#ifdef __INTEL_COMPILER
+    virtual ~DistortedCellList() noexcept override {}
+#else
     virtual ~DistortedCellList() noexcept override = default;
+#endif
```
bangerth, line 1539: Let's just make this change unconditionally. We're going to scratch our heads in a couple of years what we were working around with this. If you wanted to be super clean, you'd just do
```
virtual ~DistortedCellList() noexcept override;
```
in the `.h` file, and 
```
DistortedCellList::~DistortedCellList() noexcept = default;
```
in the `.cc` file.

bangerth (commented):
The rest of the patch is ok.

include/deal.II/grid/tria.h
masterleinad: I followed that idea.

bangerth (approved):
Thanks!

PR #12101, opened 2021-04-27
Add class RefSpaceFEFieldFunction

simonsticko:
Add a new class RefSpaceFEFieldFunction, which similarly to
FEFieldFunction evaluates a function defined by a solution vector and a
DoFHandler, but transformed to reference space. Thus defining the
function

\hat{f}(\hat{x}) = \sum_{j=0}^{n-1} f_j \hat{\phi}_j(\hat{x})

where \hat{\phi}_j(\hat(x)) are the local reference space shape
functions, f_j the solution coefficients, and \hat{x} a point on the
reference cell.

### DIFF
--- a/include/deal.II/numerics/fe_field_function.h
+++ b/include/deal.II/numerics/fe_field_function.h
@@ -486,6 +486,125 @@ namespace Functions
       const typename DoFHandler<dim, spacedim>::active_cell_iterator &cell,
       const Point<dim> &point) const;
   };
+
+
+  /**
+   * This class evaluates a function defined by a solution vector and a
+   * DoFHandler transformed to reference space. To be precise, if we let
+   * $\hat{x}$ be a point on the reference cell, this class implements the
+   * function
+   *
+   * $\hat{f}(\hat{x}) = \sum_{j=0}^{n-1} f_j \hat{\phi}_j(\hat{x})$,
+   *
+   * where $f_j$ are the local solution values and $\hat{\phi}_j(\hat(x))$ are
+   * the local reference space shape functions. The gradient and Hessian of this
+   * function are thus derivatives with respect to the reference space
+   * coordinates, $\hat{x}_0, \hat{x}_1, \ldots$.
+   *
+   * Note that this class is similar to FEFieldFunction, but that
+   * FEFieldFunction implements the following function on a given cell, $K$,
+   *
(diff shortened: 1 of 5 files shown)


simonsticko:
Need this class as part of implementing something else. I'm not really sure if it makes sense to have available or if I should just keep it in an internal namespace.

kronbichler:
This seems to be related to the new class FEPointEvaluation https://dealii.org/developer/doxygen/deal.II/classFEPointEvaluation.html - I suggest we should sit down together at some point and discuss the requirements and goals on a higher level to find a good strategy for unfitted FEM calculations with ghost-penalty-type stabilizations in deal.II.

simonsticko:
I don't think this class has the same purpose as `FEPointEvaluation`. This class gives you derivatives with respect to reference space coordinates.

I use this class to transform a level set function in real space to a level set function in reference space, so that the quadratures I construct integrate over a subset of reference space.

image

kronbichler:
I agree that the purpose is different at this stage; what I'm trying to understand (and happy to learn more about in a personal discussion) is to what extent we can offer them within a single interface to the user, and where to best collect various functionalities with similar purposes.

peterrum (commented):
Is this class used in #12757? I couldn't find it.

> I agree that the purpose is different at this stage; what I'm trying to understand (and happy to learn more about in a personal discussion) is to what extent we can offer them within a single interface to the user, and where to best collect various functionalities with similar purposes.

Did you get to chance to discuss these topics? I agree that FPE is doing somewhat similar things, with the exception that mapping is applied and the user reading the values from the global vector. But I guess - if we decide to go with `RefSpaceFEFieldFunction` -  we could extend FPE and use the information in this class here!?

simonsticko:
> Is this class used in #12757? I couldn't find it.

Not directly, it is used in the implementation of `NonMatching::FEValues` in #12756. Not sure if this would be better as internal function in `source/non_matching/fe_values.cc` or not.

peterrum:
> Not sure if this would be better as internal function in source/non_matching/fe_values.cc or not.

That would be an option. If there is currently no use case, I think I would prefer if this class could be internal.

simonsticko:
Closing this, since we in the meeting decided to make this an internal class of #12756.

PR #12103, opened 2021-04-27
Simplify use of a function in ReferenceCell.

bangerth:
One has to provide the template arguments for this function explicitly. But we can default `spacedim` to `dim`.

/rebuild


masterleinad (approved):
We do that in a lot of other places, too.

peterrum (approved):

Issue #11879, opened 2021-03-10
Completely replace GeometryInfo by ReferenceCell


bangerth:
`ReferenceCell` is our way forward, and we should move the documentation we currently have in `GeometryInfo` to its new home. There are two things that need to happen:
* Move the documentation itself for quads/hexes
* Write corresponding documentation for all of the other element types.
This would, in a sense, be a continuation of #11544.

I have given some thought to whether we should attach the specific numbering conventions to the `ReferenceCell` class itself, or to the special objects in `namespace ReferenceCells`. I have not come to a conclusion in this regard -- I lean towards the former, but would be open to other suggestions.

The next steps will then be:
* Deprecate `GeometryInfo`
* Remove all uses of that class in the library.

peterrum:
We are too close to the release. Let's postpone this after the release.

drwells:
Related: we need something like `CellInfo<dim>::max_faces_per_cell` or equivalent instead of hard-coding in things like #12126.

kronbichler:
As observed in #12135, we would (also) need a non-templated version that takes `dim` as an argument, possibly in a `constexpr` context to ensure the parts known at compile time are handled well.

peterrum:
What is realistic here before the release?

> Deprecate GeometryInfo
Remove all uses of that class in the library.

would be nice and important. But I am afraid many of us don't have time to work on this.

bangerth:
Needs to wait for the next release to be completely finished.

PR #12107, opened 2021-04-27
Replace some more occurrences of GeometryInfo.

bangerth:
`Triangulation::refine_global()` still has a long ways to go for simplex meshes, but here are a couple of the first places where we currently run into assertions. Might as well fix them. I stopped at the first one that I didn't immediately understand :-)

/rebuild

### DIFF
--- a/source/grid/tria.cc
+++ b/source/grid/tria.cc
@@ -448,12 +448,12 @@ namespace
         std::vector<unsigned int> quad_cell_count(triangulation.n_raw_quads(),
                                                   0);
         for (const auto &cell : triangulation.cell_iterators())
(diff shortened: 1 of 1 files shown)


drwells (approved):

peterrum (approved):

peterrum:
A short side-remark. As far as I know meshes with triangles should work for `Triangulation::refine_global()`. Also TET meshes should work once #11490 is merged. But I am not sure when is the right time to merge that (I haven't had to time to create nice (ASCII) pictures yet).

bangerth:
Part of #11879.

PR #12111, opened 2021-04-28
Provide TableBase::replicate_across_communicator().

### LINKED ISSUE #12008: Allow sharing of data across MPI processes

bangerth:
Closes #12008 at long last.

/rebuild

### DIFF
--- a/include/deal.II/base/function_lib.h
+++ b/include/deal.II/base/function_lib.h
@@ -1398,6 +1398,55 @@ namespace Functions
    * @note The use of the related class InterpolatedUniformGridData is
    * discussed in step-53.
    *
+   *
+   * <h3>Dealing with large data sets</h3>
+   *
+   * This class is often used to interpolate data provided by fairly
+   * large data tables that are expensive to read from disk, and that take
+   * a large amount of memory when replicated on every process of parallel
+   * (MPI) programs.
+   *
+   * The Table class can help with amortizing this cost by using
+   * shared memory to store the data only as often as necessary -- see the
+   * documentation of the TableBase class. Once one has obtained such a
+   * Table object that uses shared memory to store the data only as often
+   * as is necessary, one has to avoid that the current class *copies*
+   * the table into its own member variable. Rather, it is necessary to
+   * use the *move* constructor of this class to take over ownership of
+   * the table and its shared memory space. This can be achieved using
+   * the following extension of the code snippet shown in the
+   * documentation of the TableBase class:
+   * @code
+   *    const unsigned int N=..., M=...;     // table sizes, assumed known
+   *    Table<2,double>    data_table;
+   *    const unsigned int root_rank = 0;
+   *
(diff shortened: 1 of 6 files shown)


include/deal.II/base/function_lib.h
```diff
@@ -1398,6 +1398,55 @@ namespace Functions
...
+   *
+   * The Table class can help with amortizing this cost by using
+   * shared memory to store the data only as often as necessary -- see the
+   * documentation of the TableBase class. Once one has obtained such a
+   * Table object that uses shared memory to store the data only as often
+   * as is necessary, one has to avoid that the current class *copies*
+   * the table into its own member variable. Rather, it is necessary to
+   * use the *move* constructor of this class to take over ownership of
+   * the table and its shared memory space. This can be achieved using
+   * the following extension of the following code snippet shown in the
```
masterleinad, line 1417:
```suggestion
   * the following extension of the code snippet shown in the
```

include/deal.II/base/function_lib.h
```diff
@@ -1530,6 +1579,14 @@ namespace Functions
    *
    * @note The use of this class is discussed in step-53.
    *
+   *
+   * <h3>Dealing with large data sets</h3>
+   *
+   * This class supportes the same facilities for dealing with large data sets
```
masterleinad, line 1585:
```suggestion
   * This class supports the same facilities for dealing with large data sets
```

bangerth:
@masterleinad I've incorporated your suggestions.

masterleinad (approved):

bangerth:
Ping?

PR #12113, opened 2021-04-29
DynamicSparsityPattern: Avoid bug with 64 bit integers

kronbichler:
This is a very nasty bug that made some computations run out of memory when using more than 4b DoFs. In order not to forget things, I post this already even though I won't have time to provide a test for a week or so.


peterrum (approved):

masterleinad (changes requested):
Please write a test. :slightly_smiling_face:

bangerth:
Nice find! I'm tagging this for 9.3.

How are you going to write a test for >4B tests that doesn't require an enormous amount of memory?

peterrum:
> How are you going to write a test for >4B tests that doesn't require an enormous amount of memory?

Do we actually need a test here? That the change is correct is a no-brainer, isn't it?

bangerth (approved):

bangerth:
/rebuild

Let's run this PR through the CI machinery. I'm for merging as-is. We can always add another test later if @kronbichler feels like it, but I agree that the patch is pretty much obvious.

bangerth:
@masterleinad What do you think?

masterleinad (approved):
Sure. We can merge as is. I just thought @kronbichler wanted a reminder to provide a test case.

kronbichler:
Let me add the test here now, as I have some time now. I agree with @masterleinad that adding a test and a reminder is great to have :wink:

peterrum (approved):

kronbichler:
Test is added, ready to go from my end.

tests/sparsity/dynamic_sparsity_pattern_23.cc
```diff
@@ -0,0 +1,68 @@
...
+#include <deal.II/lac/dynamic_sparsity_pattern.h>
+
+#include "../tests.h"
+
+
+void
+test()
+{
+  // take a bit more indices than we would have for 32 bit integers
+  IndexSet set((1ULL << 32) + 4);
```
bangerth, line 30: That's actually about twice as many as we have for 32-bit integers :-)

kronbichler: You mean I should have said *unsigned integers*?

bangerth: The largest unsigned int is `1 << 31`, not `1 << 32`. So you actually have 8B+4 here. Not that that matters, it was just a comment on the "a bit more".

PR #12118, opened 2021-04-30
Remove DEAL_II_WITH_SIMPLEX_SUPPORT

peterrum:


marcfehling (approved):

marcfehling:
A few tests fail on the `gcc-serial` runner:
```
The following tests FAILED:
	225 - simplex/step-12c.debug (Failed)
	246 - simplex/matrix_free_02.debug (Failed)
```

tjhei (approved):
Looks good. Did you all agree to do this step before the release? Makes sense to me. No harm ...

peterrum:
> Looks good. Did you all agree to do this step before the release? Makes sense to me. No harm ...

Yes, we did. Independent experience showed that enabling the flag per default did not harm.

